PR #60, opened 2024-12-09
optimize tick icons

gardaholm:
optimizing display of tick icons to remove the vertical offset and create some horizontal distance (otherwise icons look too ugly)
image


marcodejongh:
Thanks!

PR #81, opened 2024-12-16
Add Logbook Stats in Queue Drawer

gardaholm:
first quick prototype, definitely needs some refinement and more charts.
The "Logbook Stats" Button is currently in the queue drawer. I guess for further improvements it would make sense to add a profile icon in the top corner (near the bulb?) where we also add Logout Button, Avatar, etc.


app/lib/api-wrappers/aurora/getUserLogbook.ts
```diff
@@ -0,0 +1,64 @@
+import { sql } from '@/lib/db'; // Assuming you have a database utility setup
```
marcodejongh, line 1: Didn't this file already exist 🤔

marcodejongh:
It worked! If you fix the eslint/typescript errors then Im happy to merge in its current form. Btw on my open #80  I'm adding extra rest apis for getting followers etc

gardaholm:
> It worked! If you fix the eslint/typescript errors then Im happy to merge in its current form. Btw on my open #80 I'm adding extra rest apis for getting followers etc

should be fixed (hopefully)

app/lib/api-wrappers/aurora/getUserLogbook.ts
gardaholm: yes it did exist but it only fetched the logbook for one single climb uuid and it did fetch bids (without difficulty).
I expanded the query in your getLogbook.ts to allow both query, depending if you pass an uuid or not. so your calls are still working as expected.
a) if a climbuuid is passed you get the logbook as before
b) if no climbuuid is passed you get a logbook where climbs without difficulties are removed from the dataset 
--
but depending on the stats we want to show, we may need to add the bids too

PR #86, opened 2024-12-19
Validation Improvements

gardaholm:
1) added validations to the board selections – it always happend to me that I pressed next before selection the board size. 
2) https://github.com/marcodejongh/boardsesh/issues/70: added visual indication if a route is added to the queue and set a short timer to block the + button (for this route) for 3 sec

### DIFF
--- a/app/components/climb-card/climb-card-actions.tsx
+++ b/app/components/climb-card/climb-card-actions.tsx
@@ -1,5 +1,5 @@
 'use client';
-import React from 'react';
+import React, { useState } from 'react';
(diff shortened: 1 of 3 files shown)


marcodejongh:
Thanks! Would be nice to also change the plus button to a green check icon or something after adding, but this id already a good improvement so I'll merge

gardaholm:
> Thanks! Would be nice to also change the plus button to a green check icon or something after adding, but this id already a good improvement so I'll merge

makes sense! changed in this PR
https://github.com/marcodejongh/boardsesh/pull/91

Issue #109, opened 2025-01-02
Use deltas for sharing data in party sessions


marcodejongh:
Currently boardsesh sync its complete queue state whenever a queue state change is made, this works mostly okay with a small number of users. But above 4 starts to create issues. Also each peer connects to every other peer meaning that the current protocol is very noisey.

Syncing the whole state is a naive implementation, and instead we should move to syncing model where only the state deltas are synced. Ie when a user adds climb X we only send "Added climb X" relevant information.

Later we might need to figure out conflict resolution too, but I would leave that out of scope for now. To deal with the noiseyness, we should only send updates to the host and then have the host propagate from there to the other peers. To do this we do need ro sync the other peers from the host down, in case the host goed fown. If the host goes down to ne of the existing peers becomes the host, to do this we'd require periodic health status messages.

- [x] sync state changes onky from peer to host and from host back to other peers
- [x] Implement health messages, and automatic host swithcinf if host goes down
- [x] refactor from syncing full state changes over to deltas

PR #250, opened 2025-07-20
Improve peerjs communication protocol for party mode

marcodejongh:
1. Move to a more intelligent protocol for syncing queue changes.
2. Renegotiate peers and hosts when connection is lost


marcodejongh:
@gardaholm This fixes #109 and sessions with multiple people should be a lot more stable now

PR #297, opened 2025-11-05
Fix setter name filter bug and add multiselect autocomplete with route counts

marcodejongh:
**Bug Fix:**
- Fixed setter name filter not actually filtering climbs
- The parameter was flowing through UI → URL → API but never used in SQL WHERE clause
- Added setterNameCondition to create-climb-filters.ts using inArray()

**Enhancements:**
- Changed setter filter from single text input to multiselect autocomplete
- Added route count display next to each setter name (e.g., "John Doe (42)")
- Created new API endpoint: /api/v1/[board]/[layout]/[size]/[sets]/[angle]/setters
- Uses AntD Select component with mode="multiple" for better UX

**Type Changes:**
- Updated SearchRequest.settername from string to string[]
- Updated URL parameter handling to support comma-separated values
- Fixed analytics tracking to check array length

**Technical Details:**
- New query: app/lib/db/queries/climbs/setter-stats.ts
- New component: app/components/search-drawer/setter-name-select.tsx
- Uses SWR for data fetching with client-side search filtering
- Maintains AntD design system best practices


app/components/search-drawer/setter-name-select.tsx
```diff
@@ -0,0 +1,74 @@
...
+    // Pattern: /[board_name]/[layout_id]/[size_id]/[set_ids]/[angle]/...
+    const match = pathname.match(/^\/([^/]+)\/([^/]+)\/([^/]+)\/([^/]+)\/([^/]+)/);
+    if (!match) return null;
+
+    const [, boardName, layoutId, sizeId, setIds, angle] = match;
+    return `/api/v1/${boardName}/${layoutId}/${sizeId}/${setIds}/${angle}`;
+  }, [pathname]);
+
+  // Fetch setter stats from the API
+  const { data: setterStats, isLoading } = useSWR<SetterStat[]>(
```
marcodejongh: Does this mean it uses progressive loading? We probably dont want to load all the setters by default. We need progressive load, and if the user types in text, we should filter on that text

PR #363, opened 2025-12-25
fix: Restore infinite scroll functionality in climb list

marcodejongh:
## Summary
- Fix scroll container not found during SSR hydration by using a state-based context with callback ref pattern
- Default `hasMoreResults` to `true` before SWR fetches data
- Catch `getLogbook` errors to prevent unhandled promise rejections
- Pass DOM element reference to InfiniteScroll instead of string ID

## Root Cause
The `react-infinite-scroll-component` with `scrollableTarget="content-for-scrollable"` was trying to find the scroll container by ID during SSR hydration, but the element didn't exist yet. The component caches the scroll target on mount and doesn't re-check, causing infinite scroll to never work.

## Solution
Created a `ScrollableContent` client component that:
1. Uses a callback ref to capture the DOM element
2. Stores it in React state (not a ref) to trigger re-renders
3. Passes the actual DOM element to `scrollableTarget` via context
4. Renders a fallback until the scroll container is available

## Test plan
- [ ] Visit climb list page without `?page=X` in URL
- [ ] Scroll down to the bottom
- [ ] Verify URL updates with `?page=1`, `?page=2`, etc.
- [ ] Verify more climbs are loaded


marcodejongh:

PR #364, opened 2025-12-25
fix: Default hasMoreResults to true until data is loaded

marcodejongh:
The infinite scroll trigger was broken because hasMoreResults returned
undefined (falsy) when SWR data hadn't loaded yet. This prevented the
InfiniteScroll component from triggering the next fetch.

When using SSR initialClimbs, the client-side SWR fetch hasn't completed
yet, so hasMoreResults was undefined. Now it defaults to true until
we have actual data to determine the correct value.


marcodejongh:

PR #486, opened 2025-12-30
Fix login state not updating buttons immediately

marcodejongh:
When using signIn with redirect: false, the session cookie is set but the useSession hook doesn't automatically refetch. Call updateSession() after successful login to ensure all components using isAuthenticated re-render with the correct auth state without requiring a page refresh.


marcodejongh:

PR #614, opened 2026-01-24
Refactor layout tabs to prevent unnecessary re-renders

marcodejongh:
## Summary
Refactored the tab layout component to improve performance by preventing unnecessary re-renders and better separating concerns. Each tab now manages its own state subscriptions rather than having a single parent component subscribe to all contexts.

## Key Changes
- **Extracted tab components**: Split `TabsWrapper` into separate memoized components (`QueueTabLabel`, `QueueTabContent`, `SearchTabContent`, `HoldsTabContent`) to isolate state subscriptions and prevent cascading re-renders
- **Optimized queue context usage**: `QueueTabLabel` now directly subscribes to queue context for the badge count, while `QueueTabContent` handles queue operations independently
- **Added memoization**: Used `useMemo` for `tabItems` array and `useCallback` for `handleClearQueue` to prevent unnecessary re-renders when props haven't changed
- **Updated dependencies**: Changed `@playwright/test`, `@types/react`, `bufferutil`, `node-gyp-build`, `playwright`, and `playwright-core` from `dev: true` to `devOptional: true` in package-lock.json. Removed `dev: true` flag from `react`, `react-dom`, and `scheduler` to make them production dependencies

## Implementation Details
- Each tab content component is wrapped with `React.memo()` and has a `displayName` for better debugging
- The `handleClearQueue` callback now captures `itemsCleared` before clearing the queue to ensure accurate tracking
- Tab items are memoized with `boardDetails` as the only dependency, reducing re-computation when queue state changes
- This architecture allows the Queue tab to update its badge independently without re-rendering Search and Holds tabs

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4565,7 +4565,7 @@
       "version": "1.57.0",
       "resolved": "https://registry.npmjs.org/@playwright/test/-/test-1.57.0.tgz",
       "integrity": "sha512-6TyEnHgd6SArQO8UO2OMTxshln3QMWBtPGrOCgs3wVEmQmwyuNtB10IZMfmYDE0riwNR1cu4q+pPcxMVtaG3TA==",
-      "dev": true,
+      "devOptional": true,
       "license": "Apache-2.0",
       "dependencies": {
         "playwright": "1.57.0"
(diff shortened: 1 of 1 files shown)


marcodejongh:

PR #739, opened 2026-02-08
Harden Aurora sync against malicious data

marcodejongh:
## Summary

- **P0 Fix:** `draft_climbs` can no longer overwrite published climbs. Both user-sync files now gate all `onConflictDoUpdate` fields on `isDraft = true` using SQL CASE expressions. Published climbs are fully immutable via sync.
- **P0 Fix:** Corrected backwards `isDraft` CASE in `shared-sync.ts` — only allows `true → false` (publishing), never `false → true` (unpublishing).
- **P1 Fix:** `climb_stats` numeric fields validated with `sanitizeNumber()` and clamped to bounds (difficulty 0-50, quality 0-5, ascensionistCount 0-10M). Invalid incoming values preserve existing DB values on conflict. History inserts skipped when all values invalid.
- **P1 Fix:** Sync timestamps validated (reject unparseable/pre-2016, clamp future to now). `table_name` validated against known allowlists in both `updateSharedSyncs` and `updateUserSyncs`.
- **P2 Fix:** Per-table record cap of 10,000 prevents memory exhaustion from oversized responses.
- **P2 Fix:** Beta link URLs validated as http/https; invalid thumbnails nulled out (record kept).
- **P2 Fix:** String fields truncated to reasonable limits (name: 500, description: 10K, comment: 5K, frames: 500K, URLs: 2048) across all sync paths.

New `sync-validation.ts` utility files created in both packages (duplicated to match existing pattern; consolidation is a follow-up).

## Test plan

- [ ] `npm run typecheck` passes (verified)
- [ ] `npm run lint` passes with 0 errors (verified)
- [ ] Review SQL CASE expressions in `draft_climbs` match the proven pattern from `shared-sync.ts` `upsertClimbs`
- [ ] Verify shared sync still works end-to-end for Kilter/Tension boards
- [ ] Verify user sync (drafts, ascents, circuits) still works end-to-end

### DIFF
--- a/packages/aurora-sync/src/sync/sync-validation.ts
+++ b/packages/aurora-sync/src/sync/sync-validation.ts
@@ -1,0 +1,164 @@
+/**
+ * Validation utilities for Aurora sync data.
+ *
+ * Aurora API responses are untrusted input - all values must be validated
+ * before writing to our database.
+ *
+ * NOTE: This is a copy of packages/web/app/lib/data-sync/aurora/sync-validation.ts
+ * Will be consolidated into a shared package in a follow-up.
+ */
+
+// --- Bounds constants ---
+
+export const CLIMB_STATS_BOUNDS = {
+  displayDifficulty: { min: 0, max: 50 },
+  benchmarkDifficulty: { min: 0, max: 50 },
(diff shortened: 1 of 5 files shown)


marcodejongh:

PR #977, opened 2026-03-29
Fix climb thumbnail placeholder sizing and event handler mapping

marcodejongh:
## Summary
This PR fixes two issues in the climb card and climbs list components: the placeholder thumbnail now respects the `maxHeight` prop, and the climb event handler map is built from all climbs instead of only visible ones.

## Key Changes
- **ClimbThumbnail**: Updated `placeholderStyle` function to accept and apply the `maxHeight` prop, with a default fallback of `'10vh'`. This ensures the placeholder maintains consistent sizing with the actual thumbnail.
- **ClimbsList**: Changed `climbHandlersMap` to be built from all `climbs` instead of just `visibleClimbs`, ensuring event handlers are available for all climbs regardless of visibility state. Updated the dependency array accordingly.

## Implementation Details
- The `maxHeight` parameter is now threaded through from the component prop to the placeholder style function
- The climb handler map now uses the complete climbs array as its source of truth, preventing potential issues where handlers might be missing for non-visible climbs

https://claude.ai/code/session_01Q6dmPoF4hdTeGKKXoDTp3L

### DIFF
--- a/packages/web/app/components/climb-card/__tests__/climb-thumbnail.test.tsx
+++ b/packages/web/app/components/climb-card/__tests__/climb-thumbnail.test.tsx
@@ -78,6 +78,60 @@ const climb = {
   benchmark_difficulty: null,
 } as Climb;
 
+describe('ClimbThumbnail placeholder', () => {
+  beforeEach(() => {
+    // Override: IntersectionObserver that does NOT fire, so the placeholder stays visible
+    vi.stubGlobal(
+      'IntersectionObserver',
+      class {
+        observe() {}
+        disconnect() {}
+        unobserve() {}
+      },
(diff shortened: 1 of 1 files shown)


packages/web/app/components/board-page/climbs-list.tsx
```diff
@@ -147,11 +147,11 @@ const ClimbsList = ({
 
   const climbHandlersMap = useMemo(() => {
     const map = new Map<string, () => void>();
-    visibleClimbs.forEach(climb => {
+    climbs.forEach(climb => {
```
marcodejongh: @diziet Let me know if this breaks anything

PR #994, opened 2026-03-30
Normalize email addresses across authentication flows

marcodejongh:
## Summary
This PR normalizes email addresses (lowercase and trimmed) across all authentication-related endpoints to ensure consistent email handling and prevent duplicate user accounts due to case sensitivity or whitespace variations.

## Key Changes
- **auth-options.ts**: Normalize email in both the credentials provider callback and the JWT callback before database lookups
- **register/route.ts**: Normalize email immediately after validation before checking for existing users
- **resend-verification/route.ts**: Normalize email after validation before database queries
- **verify-email/route.ts**: Normalize email from query parameters with null-coalescing fallback

## Implementation Details
All email normalization follows the same pattern: `email.toLowerCase().trim()` applied consistently at the point of use before any database operations. This ensures that:
- Users cannot create duplicate accounts with different email casings (e.g., "User@Example.com" vs "user@example.com")
- Email lookups are case-insensitive and whitespace-tolerant
- The normalization happens early in the flow, right after validation/retrieval

https://claude.ai/code/session_018BHRp1fUQSgyXYVa4jimwt

### DIFF
--- a/packages/web/app/api/auth/register/route.ts
+++ b/packages/web/app/api/auth/register/route.ts
@@ -44,7 +44,8 @@ export async function POST(request: NextRequest) {
       return NextResponse.json({ error: validationResult.error.issues[0].message }, { status: 400 });
     }
 
-    const { email, password, name } = validationResult.data;
+    const { password, name } = validationResult.data;
+    const email = validationResult.data.email.toLowerCase().trim();
     const db = getDb();
 
     // Check if user already exists
--- a/packages/web/app/api/auth/register/route.ts
+++ b/packages/web/app/api/auth/register/route.ts
@@ -47,7 +47,8 @@ export async function POST(request: NextRequest) {
       );
     }
 
-    const { email, password, name } = validationResult.data;
+    const { password, name } = validationResult.data;
+    const email = validationResult.data.email.toLowerCase().trim();
     const db = getDb();
 
     // Check if user already exists
--- a/packages/web/app/api/auth/resend-verification/route.ts
+++ b/packages/web/app/api/auth/resend-verification/route.ts
@@ -55,7 +55,7 @@ export async function POST(request: NextRequest) {
       return NextResponse.json({ error: validationResult.error.issues[0].message }, { status: 400 });
     }
 
-    const { email } = validationResult.data;
+    const email = validationResult.data.email.toLowerCase().trim();
     const db = getDb();
 
     // Check if user exists and is unverified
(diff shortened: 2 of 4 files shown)


marcodejongh:
Rebased onto main @19b7ad6a3 (was 2 months stale). Conflicts were formatting-only (single-quote migration + single-line query collapse in `auth-options.ts` and `verify-email/route.ts`); resolved by keeping main's formatting and applying the `.toLowerCase().trim()` normalization. `register` and `resend-verification` auto-merged cleanly. All four files still normalize email before DB lookup.

Validation: `vp run typecheck` (web build + tsc) clean; `vp check` clean on all four changed files (19 pre-existing format-drift files on main are untouched and unrelated); `auth-options.test.ts` + auth-status tests pass (30 tests). e2e auto-triggers are disabled on current main.

Note: the Claude review's two asks still stand and are out of mechanical-rebase scope — (1) a data migration to lowercase existing mixed-case emails in the `users` table (with case-collision handling), and (2) unit tests covering mixed-case login/registration/verification. Both are follow-up work for a maintainer.

marcodejongh:
Closing this in favour of #3278, which does the same thing without the lockout.

**As written, this locks out every existing mixed-case account.** The credentials provider lowercases the input and then looks up `eq(schema.users.email, email)`. Stored rows are untouched, so a user who signed up as `User@Example.com` no longer matches anything and can't log in. Same in the JWT callback, `resend-verification`, and `verify-email`. That was the blocker raised in review on 2026-03-30 ("Missing data migration for existing users (login breakage)"), and the 2026-06-09 rebase carried the normalisation forward without adding a backfill — the diff is still four files, no migration.

**#3278 gets it right** by matching case-insensitively against the stored value instead of assuming it's already normalised:

```ts
.where(sql`lower(${schema.users.email}) = ${normalizedEmail}`)
```

Legacy mixed-case rows still match, so nobody is locked out. It also handles the part this PR can't: an email that already has two rows differing only by case. The credentials provider verifies the password against every candidate and returns whichever matches, and `packages/db/scripts/merge-accounts.ts` collapses each `lower(email)` group down to one account. Plus a shared `normalizeEmail` in `@boardsesh/db/utils` (with tests) so web and the backend's native-auth path agree, and `forgot-password` / `reset-password` covered too — neither is touched here.

#3278 is open, out of draft, and green right now.

State here: 3166 commits behind `origin/main`, CI red at `setup`.

PR #1013, opened 2026-03-30
Add session max-duration expiration and queue persistence on session end

marcodejongh:
## Summary
This PR implements automatic session expiration after a maximum duration (6 hours by default) and preserves user queue data when sessions expire, allowing users to continue with their climbs in solo mode.

### Backend
- **Session max-duration auto-end job**: Added a new periodic job (runs every 5 minutes) that automatically ends sessions exceeding `SESSION_MAX_DURATION_MINUTES` (default 360 minutes/6 hours)
  - Queries for non-permanent sessions that haven't ended and are older than the threshold
  - Uses `COALESCE(startedAt, createdAt)` to handle cases where `startedAt` is null
  - Publishes `SessionEnded` event with reason before ending the session
  - Implements retry logic with failure tracking (max 3 retries per session)
  - Added `ne(sessions.status, 'ended')` condition to avoid re-processing already-ended sessions

### Frontend
- **Queue persistence on session end**: When a `SessionEnded` event is received, the current queue and climb data are saved to IndexedDB via `saveQueueState()`
  - Only saves if queue has items or there's a current climb (guards against saving empty state)
  - Removes the active session preference so the app treats it as a local (solo) queue on next mount
  - Gracefully handles save failures with error logging

### Testing
- **Backend tests** (`session-expiration.test.ts`): Comprehensive test suite verifying the expiration query logic
  - Sessions older than 6 hours are correctly identified
  - Young sessions are not expired
  - Permanent sessions are excluded from expiration
  - Already-ended sessions are not re-expired
  - `createdAt` fallback works when `startedAt` is null
  - Both active and inactive expired sessions are found

- **Frontend tests** (`session-expiration-queue-save.test.tsx`): Tests verifying queue preservation flow
  - Queue is saved to IndexedDB when session expires
  - Queue is restored as local queue on next mount
  - `getMostRecentQueue()` retrieves saved queues
  - Empty queues are not saved

## Implementation Details
- The max-duration job runs in the same 5-minute interval as the inactivity auto-end job
- Separate failure tracking maps for inactivity vs. max-duration expiration
- Session end event is published before `roomManager.endSession()` to ensure connected clients receive it while still subscribed
- Queue storage uses the board path as the key, allowing restoration of the exact board context

https://claude.ai/code/session_01LZvSodXLygKq4KhVBxQD1Q

### DIFF
--- a/packages/backend/src/__tests__/session-expiration.test.ts
+++ b/packages/backend/src/__tests__/session-expiration.test.ts
@@ -1,0 +1,249 @@
+import { describe, it, expect, beforeEach, afterEach, vi } from 'vitest';
+import { v4 as uuidv4 } from 'uuid';
+import { roomManager } from '../services/room-manager';
+import { db } from '../db/client';
+import { sessions } from '../db/schema';
+import { eq, and, ne, lt, isNull } from 'drizzle-orm';
+import { sql } from 'drizzle-orm';
+import { createMockRedis, type MockRedis } from './helpers/mock-redis';
+
+// Helper function to register a client before joining
+const registerAndJoinSession = async (
+  clientId: string,
+  sessionId: string,
+  boardPath: string,
+  username?: string
+) => {
+  await roomManager.registerClient(clientId);
+  return roomManager.joinSession(clientId, sessionId, boardPath, username);
+};
+
+describe('Session Expiration - Max Duration', () => {
+  let mockRedis: MockRedis;
+
+  beforeEach(async () => {
+    mockRedis = createMockRedis();
+    roomManager.reset();
+    await roomManager.initialize(mockRedis);
+  });
+
+  afterEach(() => {
+    vi.clearAllTimers();
+  });
+
+  it('should find sessions older than 6 hours with the expiration query', async () => {
+    const sessionId = uuidv4();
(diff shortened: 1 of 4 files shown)


marcodejongh:
Too divergent from main to rebase mechanically. This branch forks at PR #1007 (2026-03-30) and the session subsystem was restructured since: the inline `autoEndInterval` block in `packages/backend/src/server.ts` that this PR patched was extracted out of `server.ts` (main no longer touches the `sessions` table there; inactivity logic now lives in `endStaleInactiveSessions()` in `session-discovery.ts`), and `use-session-lifecycle.ts` grew 592→905 lines with the `SessionEnded` case refactored into a new `useQueueStorage`/`fetchAutoFinishedSummary` model that replaced this PR’s standalone `app/lib/queue-storage-db.ts`. `git rebase origin/main` conflicts across both core files on the relocated feature surface — reaching green needs a fresh implementation, not conflict resolution.

The 6-hour cap is still not in main (no `SESSION_MAX_DURATION_MINUTES`, no max-duration job), so the intent is captured in https://github.com/boardsesh/boardsesh/issues/2653 for a fresh implementation. Branch `claude/session-expiration-queue-persistence-soHbo` preserved.

PR #1050, opened 2026-03-31
Fix AbortError unhandled rejection on board list page (iOS Safari)

marcodejongh:
IndexedDB transactions abort during page navigation on iOS Safari, causing unhandled promise rejections from the idb library's internal transaction promises. Three fixes:

1. Use explicit transactions (db.transaction + await tx.done) in user-preferences-db so transaction abort rejections are caught by the try/catch instead of floating unhandled.

2. Reset cached dbPromise on failure in idb-helper and recent-searches-storage so a transient error doesn't permanently break all IDB operations for the page session.

3. Suppress console.error for AbortError (benign navigation artifact) and add missing .catch() in climbs-list preference loading.

https://claude.ai/code/session_01KeUXkjSf897TuAhKAoxxYU

### DIFF
--- a/packages/web/app/components/board-page/climbs-list.tsx
+++ b/packages/web/app/components/board-page/climbs-list.tsx
@@ -144,11 +144,13 @@ const ClimbsList = ({
   onClimbSelectRef.current = onClimbSelect;
 
   useEffect(() => {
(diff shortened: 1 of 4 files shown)


marcodejongh:
Closing as stale during PR triage. user-preferences-db rewritten to idb-helper pattern; original AbortError fix is obsolete.

PR #1073, opened 2026-04-01
Fix unhandled CloseEvent promise rejection from graphql-ws

marcodejongh:
When a GraphQL mutation timed out via Promise.race, the underlying subscription was never cleaned up. A later dispose()/WebSocket close would then fire the dangling subscription's error callback with a CloseEvent, which escaped as an unhandled promise rejection.

- Hoist `unsubscribe` out of the Promise constructor so the timeout handler can call it, preventing dangling subscriptions
- Clear the timeout timer on normal completion to avoid timer leaks
- Add `close` event listener to SafeWebSocket (defense-in-depth) to ensure CloseEvent is always handled at the source

https://claude.ai/code/session_01ARQsZLjcaWxGFxqpYyhRir

### DIFF
--- a/packages/web/app/components/graphql-queue/graphql-client.ts
+++ b/packages/web/app/components/graphql-queue/graphql-client.ts
@@ -75,6 +75,9 @@ export function createGraphQLClient(
       this.addEventListener('error', (event) => {
         if (DEBUG) console.log(`[GraphQL] Client #${clientId} WebSocket native error suppressed`, event);
       });
+      this.addEventListener('close', (event) => {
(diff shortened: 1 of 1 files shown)


marcodejongh:
Triage note: defense-in-depth WebSocket close-event/error wrapping has landed (56f4ef93b in graphql-client.ts:68-75, 184-186). The remaining issue — `unsubscribe` declared inside the Promise constructor so a Promise.race timeout cannot call it (graphql-client.ts:149-202) — is still present. Same root cause as #1544; consider consolidating.

PR #1078, opened 2026-04-01
fix: remove remaining ?session= URL construction and simplify session…

marcodejongh:
… ID source

- Remove ?session= URL append in bottom-tab-bar.tsx (cookie persists across navigations, no URL manipulation needed)
- Simplify use-session-id-management to always read session ID from cookie instead of dual-sourcing from cookie + IndexedDB


marcodejongh:
Triage note: use-session-id-management.ts has been mostly migrated to cookie-based session reads. However, bottom-tab-bar.tsx:287 still constructs `${url}${separator}session=${activeSession.sessionId}` — exactly the URL append this PR removes. Re-scope to the bottom-tab-bar fix (and any remaining dual-source IndexedDB cleanup).

Issue #1084, opened 2026-04-01
Quick filters also save hold filters cross boards


marcodejongh:
![Image](https://github.com/user-attachments/assets/9f4a0ac1-091f-46ea-af46-2b0b51704a9e)

I filtered for a specific hold on kilter, then climbed tension. Now I see a past quick filter with 1 hold, if I press it I get 0 results because the holdid doesn't exist on tension board

marcodejongh:
Fixed in rn app

PR #1095, opened 2026-04-01
fix: scope quick filters to board name to prevent cross-board hold fi…

wiliyam:
…lter bleed (#1084)


wiliyam:
Good catch @marcodejongh! Updated the store key to include `layout_id` so Kilter OG and Kilter HW (which share the same board name but have different hold IDs) get separate filter histories.

Key is now `recent_<boardName>_<layoutId>` — fully backward compatible since the old keys just get ignored.

PR #1145, opened 2026-04-03
Make email verification non-blocking

marcodejongh:
## Summary
- Removes login blocking for unverified email users — users can register and use the app immediately
- Replaces `EMAIL_VERIFICATION_ENABLED` env var with automatic SMTP detection (`SMTP_USER` + `SMTP_PASSWORD`)
- Verification emails still sent in background when SMTP is configured
- Verify-email redirect goes to `/` instead of `/auth/login` since users are already logged in

## Test plan
- [ ] Register a new account — should auto-login immediately
- [ ] Check inbox for verification email (when SMTP configured)
- [ ] Click verification link — should redirect to app root
- [ ] Login with existing unverified user — should succeed without redirect
- [ ] Register without SMTP configured — should work with no email sent

### DIFF
--- a/docs/oauth-setup.md
+++ b/docs/oauth-setup.md
@@ -69,7 +69,7 @@ Boardsesh uses NextAuth.js v4 for authentication with the following components:
 ### Key Features
 
 - **Conditional Providers**: OAuth buttons only appear when provider credentials are configured
-- **Email Verification**: New email/password accounts require email verification
+- **Email Verification**: Non-blocking — verification emails are sent when SMTP is configured, but users can log in immediately
 - **Account Linking**: Accounts are automatically linked by email — signing in with OAuth using an email that already has a password account links both methods to the same user
 - **JWT Sessions**: Stateless authentication with 5-minute refresh interval
 - **Rate Limiting**: In-memory rate limiting on auth endpoints (best-effort in serverless)
@@ -296,7 +296,7 @@ ngrok http 3000
 
 ## Email Verification Setup
 
-Email verification is required for email/password accounts. OAuth accounts are pre-verified by the provider.
+Email verification is non-blocking. When SMTP credentials (`SMTP_USER` and `SMTP_PASSWORD`) are configured, verification emails are sent on registration, but users can log in and use the app immediately. OAuth accounts are pre-verified by the provider. SMTP credential presence auto-enables verification emails. To disable verification entirely (even with SMTP configured), set `EMAIL_VERIFICATION_ENABLED=false` as an emergency kill switch.
 
 ### Fastmail Setup (Recommended)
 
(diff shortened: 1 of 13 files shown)


marcodejongh:
Rebased onto main @19b7ad6a3 (was on a pre-Apr base). Conflicts were formatting-only on the auth files plus two semantic refactors that landed on main meanwhile: verify-request now uses an i18n `KNOWN_VERIFY_ERROR_CODES` lookup (dropped `EmailNotVerified` from the set since the sign-in callback no longer emits it), and auth-modal/auth-page-content now use i18n toasts (added `login.toasts.accountCreatedCheckEmail` to en-US/es/fr, removed the now-unused `checkEmail` key). Register/resend kept the `emailVerificationActive = smtpConfigured && EMAIL_VERIFICATION_ENABLED !== "false"` kill switch and `emailVerified: emailVerificationActive ? null : new Date()` from the branch; also fixed the register test bcrypt mock for main switching to the named `{ hash }` import. Validation: vp run typecheck (exit 0), vp check (exit 0), vp run check:i18n + orphans (OK), i18n catalog-completeness 28/28, and the two new auth route tests 19/19. The web typecheck `next build` step hit a concurrent-build lock from a parallel CI run, not a code error.

marcodejongh:
Closing this — it's a product decision rather than a fix, the problem it solves isn't currently happening, and it collides head-on with #3278.

**Nobody is being blocked today.** The gate in `auth-options.ts` only fires when the flag is explicitly on:

```ts
const emailVerificationEnabled = process.env.EMAIL_VERIFICATION_ENABLED === 'true';
if (emailVerificationEnabled && existingUser.length > 0 && !existingUser[0].emailVerified) {
  return '/auth/verify-request?error=EmailNotVerified';
}
```

Unset means off — the comment right above it says "disabled by default until Fastmail auth is set up". The mobile path never checks it at all: `packages/backend/src/handlers/native-auth.ts:575` spells out that `handleNativeAuthCredentials` never reads `emailVerified`, so a native register logs the user straight in either way.

**It's a policy change, not a bug fix.** It removes the login gate outright, inverts the flag's meaning (SMTP-credential presence auto-enables verification emails; `EMAIL_VERIFICATION_ENABLED=false` becomes an emergency kill switch), and rewrites the copy and `docs/oauth-setup.md` to match. That's a call to make deliberately, not something to land off a stale branch.

**It collides with #3278** in four files — `api/auth/register/route.ts`, `api/auth/resend-verification/route.ts`, `api/auth/verify-email/route.ts`, and `lib/auth/auth-options.ts` — and in `auth-options.ts` specifically in the same `signIn` block: this PR deletes the lookup, #3278 rewrites it to `lower(email)` matching. #3278 is the one that's green and close to landing.

If email verification is ever switched on, this is worth redoing on top of #3278 — the direction (don't block login on an unverified email) reads right. As it stands: 3166 commits behind `origin/main`, `CONFLICTING`, and CI red on `lint` and `test-default`.

PR #1227, opened 2026-04-05
Export DOUBLE_TAP_THRESHOLD to avoid duplication

marcodejongh:
climb-list-item.tsx used a hardcoded 300ms delay for its single-click timeout that mirrors the constant in use-double-tap.ts. Export the constant from use-double-tap.ts and import it in climb-list-item.tsx so both values stay in sync automatically.

https://claude.ai/code/session_01WamWQGns32de2rankRK1ox


marcodejongh:
Closing as stale during triage. The hardcoded 300ms duplication this PR removes no longer exists on main — climb-list-item.tsx now delegates to the `useDoubleTap` hook which owns DOUBLE_TAP_THRESHOLD internally. The export-and-share refactor is no longer needed.

PR #1330, opened 2026-04-07
Queue duplicate fix

lilylilylily123:
you could add the same climb to queue infinite times


marcodejongh:
I think the user would expect to be able yo add the same item multiple times, as the queue works like a spotify queue. However it should be deliberate so if theres any way its easy to accidentally add the same thing 10 times we should fix that

marcodejongh:
Feel free to reopen ifnyou disagree

Issue #1297, opened 2026-04-06
Add a full logbook/history page with per-entry management


azuttre:
## Problem

Right now Boardsesh shows recent activity, but I’m missing a dedicated personal logbook/history view where I can browse all past ascents in one place.

This becomes especially noticeable when:
- I want to search whether I’ve done a climb before
- I want to review older climbing history
- I accidentally log duplicate ascents and want to clean them up

## Proposed solution

Add a dedicated logbook page for a user that supports:
- viewing full ascent/tick history
- search by climb name
- filtering by board / date / grade / ascent type
- entry-level actions like delete or edit
- possibly direct links from profile or settings

## Why this would help

A full logbook would make Boardsesh much more useful as a long-term training archive, not just a session tool.

## Notes

I’ve been using [Kiltered](https://kiltered.lovable.app/) as a read-only legacy archive, and one thing it does well is making it easy to search past ascents so I don’t unknowingly reclimb something old. I think a Boardsesh-native version of that idea would be very valuable.

azuttre:
I'll keep playing with the app some more to get more familiar and make a more informed suggestion.

The profile `recent activity` focuses on the latests climb.
The Climb > Only Completed is closer to a logbook, but it's quick limited:
- cannot sort by custom criteria
- cannot remove log entries
(let me know those feature exist already - I may have missed it)


I have in mind a proper logbook, with all the ticks and ability to search, sort, view analytics (which you already have in the profile), delete, edit log entries.

It could be under the profile, or under the Your Library section.

marcodejongh:
Addeed a while ago

PR #1332, opened 2026-04-07
Add a logbook

lilylilylily123:
done in like 30min with claude, would love some testing & feedback

Addresses #1297


lilylilylily123:
@marcodejongh seems ready for merge now

PR #1377, opened 2026-04-10
Fix queue swipe action styling and assert suggested swipe behavior

marcodejongh:


marcodejongh:
Closing as stale during triage. Commit 613342388 with the same title and scope is on origin/main; this PR's branch is no longer needed.

PR #1544, opened 2026-04-13
Fix GraphQL client disposal and subscription timeout handling

marcodejongh:
## Summary
This PR fixes two issues in the GraphQL client that could cause unhandled promise rejections and resource leaks:

1. **Client disposal**: The `graphql-ws` v6 `dispose()` method is now async and can reject. These rejections were being dropped without error handling, causing Sentry to capture unhandled rejections.

2. **Subscription timeout**: When a GraphQL operation times out, the underlying subscription wasn't being cancelled, leaving the graphql-ws lock alive indefinitely and causing the connection to keep retrying.

## Key Changes

- **Client disposal error handling**: Wrapped the `originalDispose()` call in `Promise.resolve().catch()` to handle potential rejections from the async dispose method. Errors are logged in DEBUG mode but suppressed to prevent unhandled rejections.

- **Subscription cleanup on timeout**: Hoisted the `unsubscribe` function outside the Promise constructor so it can be called when a timeout occurs, ensuring the underlying graphql-ws subscription is properly cancelled.

- **Timeout cleanup**: Added a `.finally()` block to clear the timeout ID after the race between execution and timeout completes, preventing potential memory leaks from lingering timeouts.

## Implementation Details

- The `unsubscribe` variable is now declared at the function scope and initialized when `client.subscribe()` is called, allowing the timeout handler to access it.
- The timeout ID is stored so it can be cleared in the finally block, ensuring no orphaned timeouts remain.
- Error handling for dispose is conditional on DEBUG mode to avoid noise in production logs while still suppressing the rejection.

https://claude.ai/code/session_01MhVquSCAA1wDCFjdrRjNgM

### DIFF
--- a/packages/web/app/components/graphql-queue/__tests__/graphql-client.test.ts
+++ b/packages/web/app/components/graphql-queue/__tests__/graphql-client.test.ts
@@ -1,0 +1,301 @@
+import { describe, it, expect, vi, beforeEach, afterEach } from 'vitest';
+import type { Client, Sink } from 'graphql-ws';
+
+// ---------------------------------------------------------------------------
+// Mocks
+// ---------------------------------------------------------------------------
+
+// Prevent connectionManager from starting its setInterval in the test env.
+vi.mock('../../connection-manager/websocket-connection-manager', () => ({
+  connectionManager: {
+    registerClient: vi.fn(() => vi.fn()), // returns a no-op unregister function
+    subscribe: vi.fn(() => vi.fn()),
+    getSnapshot: vi.fn(() => ({ name: null, state: 'idle', lastActivity: null })),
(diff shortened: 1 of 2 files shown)


marcodejongh:
Triage note: parts of this PR have landed on main via separate commits (8dcceeb1c — graphql-ws dispose() rejection wrapper, and 56f4ef93b — generic WebSocket DOM Event suppression). However, the PR's other fix — hoisting `unsubscribe` outside the Promise constructor and cleaning it up on Promise.race timeout (graphql-client.ts ~lines 153-202) — is NOT on main; the timeout still leaves subscriptions dangling. PR should be re-scoped to just the unsubscribe/timeout cleanup.

Issue #1585, opened 2026-04-16
[Docs] CONTRIBUTING.md has the wrong path to setup script


alejandrosanchezcabana:
In CONTRIBUTING.md there is a path to setup-dev.sh which should be ./packages/web/setup-dev.sh

PR #1586, opened 2026-04-16
[Docs] CONTRIBUTING.md has the wrong path to setup script #1585

alejandrosanchezcabana:
This solves issue #1585 which has the wrong path for the dev setup script

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -9,7 +9,7 @@
 Run the automated setup script:
 
 ```bash
-./setup-dev.sh
+./packages/web/setup-dev.sh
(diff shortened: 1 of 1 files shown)


marcodejongh:
I usually run database with bun db:up

alejandrosanchezcabana:
Which is probably the right thing to do once you got into the dev cycle, maybe I can look into expanding the guide on how to better establish a dev env (separating from a first time contributor to a more established one)

I do have a small patch in the works solving a small bug I found, I will think about it afterwards, I need to probably make myself more familiar before building an opinion though

PR #1588, opened 2026-04-16
fixed the time being displayed on ticks in logbook feed

cpud:
The logbook on the You page was displaying the wrong times: 
boardsesh

logbook-feed-item.tsx did not set dayjs to utc time, so setting utc to true will calculate the right time to display on the ticks in the logbook feed.


marcodejongh:
Thanks! I did some big changes to the ui, could you merge and fix conflicts?

cpud:
Sorry, just saw this. Thanks!

marcodejongh:
Thanks for catching this and for the fix! Reviewed against current \`main\` — the underlying bug (the logbook feed parsing the naive \`climbed_at\` timestamp as browser-local instead of UTC) has since been fixed independently.

\`packages/web/app/components/library/logbook-feed-item.tsx\` now renders tick times via \`formatTickRelativeTime(item.climbedAt)\` (imported from \`@/app/lib/format-tick-time\`, which re-exports from the shared \`@boardsesh/profile-stats\`). That shared helper parses through \`dayjs.utc(...)\` to recover the absolute moment before formatting — the same root cause this PR addressed. It landed in commit \`5e8e13b20\` (the native "You" page work).

Because of that, plus the big UI changes mentioned above, this branch no longer applies cleanly (the added \`timeAgo\`/\`statusDisplay\` block duplicates variables that no longer exist in the rewritten file). Recommend closing as superseded — the fix you reported is already live. Really appreciate the report and the clear screenshot.

PR #1607, opened 2026-04-17
Improve Instagram beta link validation and cleanup

azuttre:
## Summary
Improves Instagram beta link quality in two areas:

1. Better validation when users attach Instagram beta videos
2. Better handling of broken / undisplayable beta embeds

### Better validation when adding a beta video
- only accepts real Instagram post / reel URLs
- verifies the post is public and previewable
- checks that the caption or title mentions the climb name using a forgiving match
- improves user-facing error messages when a link is rejected
- blocks duplicate links on the same climb
- allows the same Instagram post to be linked across different climbs, since one post can contain multiple sends

### Better handling of broken beta videos
- hides Instagram beta videos that are confirmed broken / unembeddable
- newly added links are marked accessible immediately, so they should not appear broken on first display
- older unknown links are still checked when someone opens a climb, so the backlog can keep shrinking even before the manual backfill is run

### Manual backfill support
- adds a one-time backfill script to classify existing beta links:
  - `bun run --filter=@boardsesh/web backfill:beta-links`
- intentionally does **not** add a recurring cron job in this version
- this keeps rollout simpler and makes the initial cleanup an explicit operation

## Why
This should reduce:
- unrelated Instagram videos being attached to climbs
- duplicate video confusion
- broken beta embeds lingering in the UI

while still allowing a single Instagram post to be reused when it legitimately contains multiple climbs.

## Verification
- `bun run typecheck:backend`
- `bun run typecheck:web`
- focused backend validator tests for Instagram metadata parsing / climb-name matching
- manual validation against real Instagram links from existing beta videos

## Notes
- there is still a possible first-paint edge case for older, previously unchecked beta links until they are classified
- the one-time backfill script is intended to clean up that backlog deliberately rather than relying on a recurring cron


marcodejongh:
Thanks for this, @azuttre — and apologies it sat for a while.

I went through this against current `main` and the Instagram beta-link validation work has since been reimplemented by the team, so this PR is **superseded**. Concrete evidence:

- The files this PR introduces now exist in `main` with a different implementation:
  - `packages/backend/src/utils/instagram-beta-validation.ts` + its test now delegate to a dedicated `packages/backend/src/lib/instagram-meta.ts` (`fetchInstagramMeta` / `getInstagramMediaId`, a status-enum API), and `InstagramBetaValidationError` now extends `GraphQLError` so messages survive Yoga's `maskedErrors`. Neither `instagram-meta.ts` nor that approach existed when this branch was cut.
  - Notably, `main` **deliberately reversed** one of this PR's core behaviors: write-time validation no longer requires the caption/title to mention the climb name (see the regression test "returns metadata even when the caption does not mention the climb" in `instagram-beta-validation.test.ts`).
- The web-side pieces here are gone from `main`, replaced by a different architecture: `instagram-url.ts`, `instagram-validation.ts`, `beta-link-revalidation.ts`, `scripts/backfill-beta-link-accessibility.ts`, and the `/api/v1/[board_name]/beta/[climb_uuid]` route no longer exist; thumbnail/accessibility handling moved to `packages/web/app/api/internal/beta-link-thumbnail/`, `packages/backend/src/lib/beta-link-thumbnails.ts`, the shared `@boardsesh/graphql` beta-link operations, and mobile `BetaVideo*` components.
- The DB migration can't be rebased mechanically: this PR's slot `0079_gorgeous_daredevil.sql` is occupied in `main` by `0079_backfill_tick_quality_scale.sql`, and `main` is now at `0118`. The `is_accessible` / `checked_at` columns this PR adds to `board_beta_links` were not adopted.
- The red checks here (E2E shards, screenshots, claude-review) ran against the old E2E workflow that has since been disabled on `main` (commit `b8fdc038b`), so they aren't actionable.

Recommendation: **close this PR as superseded.** The validation hardening and broken-embed handling you were after both shipped, just with a different design. If there's a specific behavior here you think `main` is still missing (e.g. the caption-must-mention-climb-name check, which was an intentional product decision to drop), happy to dig into that as a focused follow-up. Thanks again for the contribution.

marcodejongh:
Thanks for this, @azuttre — and sorry it sat for so long. Closing it now, but not because the ideas were wrong: most of what you described here has since shipped, and the beta-link path was rebuilt on the backend in the meantime, so this branch can no longer be landed as-is.

**What shipped, and where it lives now.** Beta-link validation moved out of `packages/web` into the GraphQL backend:

- `packages/backend/src/utils/instagram-beta-validation.ts` — URL validation, the public/previewable check, and the specific user-facing errors ("post is private/deleted", "Instagram is temporarily blocking us") rather than a generic failure toast. Same instinct as your error-message work.
- `packages/backend/src/lib/instagram-meta.ts` — canonical media-ID extraction plus caption scraping.
- `validateAndEnrichBetaLinkInsert` in `packages/backend/src/graphql/resolvers/ticks/mutations.ts` — the single gated entrypoint: per-user rate limit, dedup, validation, then an eager S3 thumbnail cache.

**Broken embeds are hidden**, just decided at read time instead of stored. `enrichRow` in `resolvers/beta-videos/queries.ts` drops a row when the metadata lookup comes back `gone` and no thumbnail was ever cached. The reason it isn't persisted the way your `is_accessible` / `checked_at` columns did: the `gone` signal flaps when Instagram rate-limits us or serves a login wall, so a stored `false` would hide a perfectly good video until someone re-ran a backfill. The resilience contract on `main` is the inverse — once we hold our own cached thumbnail, the row keeps rendering no matter what Instagram says.

**One behaviour went the opposite way, on purpose.** You allowed the same Instagram post to be attached to several climbs, on the grounds that one post can contain multiple sends. `main` rejects that: a cross-climb duplicate throws `crossClimbDupMessage(...)` naming the climb the post is already on. Media ID is treated as the global identity of a video. That's a product call, and it's the one part of this PR that couldn't be merged without reversing a deliberate decision.

**Caption ↔ climb-name matching** landed too, though on a different path: `parseInstagramBetaCaption` + `normalizeName` + climb resolution in `resolvers/beta-videos/instagram-beta-import-classify.ts`, used by the bulk Instagram beta scan. It isn't an attach-time gate on the user-facing form the way you had it.

**Why the branch can't be rebased.** Every web file it adds is gone from `main` — `app/lib/beta-link-revalidation.ts`, `app/lib/instagram-validation.ts`, `app/lib/instagram-url.ts`, `scripts/backfill-beta-link-accessibility.ts`, and the `api/v1/[board_name]/beta/[climb_uuid]/route.ts` it hooked into. Migration slot `0079` has been occupied by `0079_backfill_tick_quality_scale.sql` for a long while, and the tree has moved 5788 commits since.

If the "one post, several climbs" case still bites you in practice, please open an issue with an example — that's the piece worth arguing on its merits, and it's a much smaller conversation than this branch. Appreciated either way.

PR #1623, opened 2026-04-18
Improve Contributing guide to make it easier for first-time contributors

alejandrosanchezcabana:
Additionally;
- Fix small issue in setup script pointing to the wrong path

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -4,6 +4,12 @@
 
 # Getting Started
 
+## How to get everything going
+
+If you want to fix bugs, build features, or make this app even better, follow the instructions described in this [blogpost](https://www.linkedin.com/pulse/git-github-demystified-guide-open-source-contribution-nishan-baral-i4ndc).
+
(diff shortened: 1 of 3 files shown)


CONTRIBUTING.md
```diff
@@ -4,6 +4,12 @@
 
 # Getting Started
 
+## How to get everything going
```
brysiukaudio, line 7: I think it might worthwhile rephrasing this. With this title it’s kind of gives the impression this blog post and video will give info on getting this project going and not just the general, this is how you use git

alejandrosanchezcabana: That is a good point, let me change that

CONTRIBUTING.md
marcodejongh, line 18: Cant remember what does script does 😆. I added bun run db:up which pulls the dev database from docker, whereas this script I think rebuilds the database and therefore takes longer. It isnt necessarily incorrect but db:up will just be faster

marcodejongh (approved):

PR #1631, opened 2026-04-20
fix(logbook): route queue actions through party session and tint the active tick

marcodejongh:
Bridge adapter's addToQueue/setCurrentClimb silently no-opped during
party mode because setLocalQueueState bails while a session is active.
Route the mutations to the persistent session's backend handlers so the
Add to Queue menu item and logbook thumbnail both update the party queue.

Tint the logbook feed item background with the grade tint when its climb
is the current one, matching the queue list item treatment.

### DIFF
--- a/packages/web/app/components/climb-card/__tests__/climb-list-item.test.tsx
+++ b/packages/web/app/components/climb-card/__tests__/climb-list-item.test.tsx
@@ -107,6 +107,7 @@ vi.mock('@/app/lib/grade-colors', () => ({
   getSoftVGradeColor: () => '#888',
   getSoftGradeColorByFormat: () => '#888',
   getGradeTintColor: () => null,
+  getActiveClimbTint: () => 'var(--semantic-selected)',
   getGradeColorWithOpacity: () => '#888',
   getGradeTextColor: () => '#000',
   isLightColor: () => false,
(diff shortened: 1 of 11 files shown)


marcodejongh:
The headline fix here — routing addToQueue/setCurrentClimb through the persistent party session when setLocalQueueState no-ops — already shipped in main via the queue-pivot work (f55cadde5 "fix(ble): repair mirror, party hand-off & create-climb sends broken by the queue-pivot" and follow-ups a63388125/5fc433fea). main's `queue-bridge-context.tsx` now has the `if (ps.activeSession) ps.addQueueItem(...)` branch in addToQueue plus the addQueueItem→setCurrentClimb-with-rollback party branch in setCurrentClimb, and the addedBy=clientId / addedByUser=usePartyProfile() metadata this PR added — built on a shared `buildQueueItem` helper, `correlationId`, `setQueue`, and `playlistSuggestionSource` rollback that didn't exist when this branch was cut (2026-04-22, ~2440 commits behind).

A rebase produces 8 structural conflicts in queue-bridge-context.tsx alone (different setCurrentClimb signature, latestRef indirection vs buildQueueItem) — resolving them means rewriting onto main's new architecture, i.e. feature rework, not a mechanical rebase. The only remnant not in main is the small logbook active-climb grade tint (getActiveClimbTint + useOptionalCurrentClimb in logbook-feed-item + --active-climb-bg), which is better re-cut fresh against current main.

Closing as superseded; branch `claude/fix-logbook-queue-climb-tbifd` preserved.

PR #1676, opened 2026-04-22
Fix back button returning to board list instead of profile when opening a climb

marcodejongh:
When a user navigates to a climb view from a profile page (e.g. Created
Climbs, Sessions activity feed) via any Next.js client-side link, the back
button wrongly sent them to the board's climb list. Client-side navigation
does not update document.referrer, so the BackButton's referrer check failed
and the fallback URL (board list) was used instead of the page they came from.

Fix: embed the current pathname as a `returnUrl` query parameter whenever
navigating to a climb view internally. ClimbViewActions reads this param and
uses it as the BackButton fallback, so the user is returned to the profile
page they were on.

https://claude.ai/code/session_01To2kwBWpLZCGBh8rYw9HcD

### DIFF
--- a/packages/web/app/components/activity-feed/ascent-thumbnail.tsx
+++ b/packages/web/app/components/activity-feed/ascent-thumbnail.tsx
@@ -1,6 +1,7 @@
 'use client';
 
 import React, { useMemo } from 'react';
+import { usePathname } from 'next/navigation';
 import LocaleLink from '@/app/components/i18n/locale-link';
 import type { BoardDetails, BoardName } from '@/app/lib/types';
 import BoardImageLayers from '@/app/components/board-renderer/board-image-layers';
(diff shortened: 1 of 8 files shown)


marcodejongh:
Too divergent from main to rebase mechanically. The climb-view back button was refactored in main since this PR opened: the consumer files this fix targets — `packages/web/app/components/climb-view/climb-view-actions.tsx` (held `getBackToListUrl`) and `packages/web/app/components/climb-detail/climb-detail-page.server.tsx` — are both deleted in main (modify/delete `DU` conflict on rebase). Back-nav is now driven by `global-header.tsx` deriving `backUrl` from a config map, with zero `returnUrl` machinery in the web tree, so the auto-merging producer-side `?returnUrl=` appends would have no reader. Re-implementing against the new architecture is feature rework, not mechanical conflict resolution. Intent captured in https://github.com/boardsesh/boardsesh/issues/2648 for a fresh implementation. Branch `claude/fix-climb-back-navigation-7yiwh` preserved.

PR #1677, opened 2026-04-22
Navigate from /you/logbook entry clicks to the climb's board view

marcodejongh:
LogbookFeedItem's row-click and thumbnail-click already invoke
setCurrentClimb when a QueueProvider is in scope, but the user's
cross-board logbook at /you/logbook sits outside any board layout — so
useOptionalQueueActions returns null and the handlers silently no-op.
Fall back to router.push with the view URL built from the entry's
boardType, layoutId, angle, and the default size/set pair (same config
already used for the BoardDetails memo), so clicking a previous climb
jumps straight to the board view.

### DIFF
--- a/packages/shared/i18n/locales/en-US/common.json
+++ b/packages/shared/i18n/locales/en-US/common.json
@@ -303,5 +303,11 @@
     "speedNormal": "1×",
     "speedCustom": "{{value}}×",
     "speedCustomLabel": "Custom speed"
+  },
+  "logbook": {
+    "session": {
+      "added": "Added {{climbName}} to your session",
+      "addError": "Could not add to your session"
+    }
   }
 }
--- a/packages/shared/i18n/locales/es/common.json
+++ b/packages/shared/i18n/locales/es/common.json
@@ -303,5 +303,11 @@
     "speedNormal": "1×",
     "speedCustom": "{{value}}×",
     "speedCustomLabel": "Velocidad personalizada"
+  },
+  "logbook": {
+    "session": {
+      "added": "Se anadio {{climbName}} a tu sesion",
+      "addError": "No se pudo anadir a tu sesion"
+    }
   }
 }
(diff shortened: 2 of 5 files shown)


marcodejongh:
Rebased onto main @19b7ad6a3 (was ~6 weeks stale).

Conflicts resolved in `logbook-feed-item.tsx` + its test: main moved the click handlers from `setCurrentClimb` to `previewClimbFromBrowse` (the "stop browse from mutating the wall climb" change) and dropped the `dispatchOpenPlayDrawer` call. Kept that behaviour when a `QueueProvider` is in scope; the no-queue `/you/logbook` fall-through (active-session push, else navigate) now layers on top. `AscentFeedItem` import follows main to `@boardsesh/graphql/operations/ticks`.

Fixes that turned CI green:
- The `set_ids.join is not a function` crash (all 29 tests failed) was a test mock returning `setIds` as the string `"1,26"` where `constructClimbViewUrl` calls `.join()`; changed the mock to `[1, 26]`.
- Hardcoded snackbar strings now use i18n keys `logbook.session.added` / `logbook.session.addError` (added to en-US/es/fr `common.json`) so `check:i18n` passes.
- Made the no-session row/thumbnail click navigate synchronously (gated on `activeSessionMatches`) instead of through an always-async IIFE, so navigation fires within the click tick.

Validation: `vp run typecheck` (pass), `vp check` on the changed files (pass), `vp run check:i18n` + `check:i18n:orphans` (pass), catalog-completeness + logbook-feed-item unit tests 57/57 pass.

marcodejongh:
Closing as part of deprecating the Next.js web climbing experience: the Expo (React Native) web app is becoming the climbing surface, so packages/web climbing UI, board control, and search/logbook surfaces are no longer receiving fixes. Gym admin, /admin, and marketing pages are unaffected. If this pain point applies to the app (native or web deploy), please re-report against the mobile app.

Issue #942, opened 2026-03-28
RangeError: Maximum call stack size exceeded.


marcodejongh:
https://boardsesh.sentry.io/share/issue/10a1458d00a44aa28244e52e4f9fe695/

# RangeError: Maximum call stack size exceeded.

**Issue ID:** 7371202285
**Project:** boardsesh
**Date:** 29/3/2026, 5:47:09 am
## Issue Summary
RangeError: Maximum call stack exceeded due to recursion
**What's wrong:** Fatal **RangeError: Maximum call stack size exceeded** occurred during execution.
**In the trace:** Multiple successful **fetch** requests preceded the error, suggesting a complex state change.
**Possible cause:** The error likely stems from **infinite recursion** in an unknown function near line 31.

## Tags

- **browser:** Chrome Mobile iOS 146.0.7680
- **browser.name:** Chrome Mobile iOS
- **device:** iPhone
- **device.family:** iPhone
- **environment:** vercel-production
- **handled:** no
- **level:** error
- **mechanism:** auto.browser.global_handlers.onerror
- **os:** iOS 26.4.0
- **os.name:** iOS
- **release:** 71920ddf59965d641cd01d55f7783ffd0efcb1d8
- **transaction:** /
- **turbopack:** True
- **url:** https://www.boardsesh.com/
- **user:** ip:90.160.133.104

### Exception 1
**Type:** RangeError
**Value:** Maximum call stack size exceeded.

#### Stacktrace

```
 Unknown function in undefined [Line 31] (In app)
```

PR #1740, opened 2026-04-27
fix(observability): symbolicate iOS Chrome stack overflows on home page

marcodejongh:
Sentry issue #942 reports a recurring `RangeError: Maximum call stack size
exceeded` on the home page from Chrome Mobile iOS. The stack trace is a
single un-symbolicated frame ("Unknown function in undefined [Line 38]")
because `productionBrowserSourceMaps` was off, so the Sentry plugin had
nothing to upload, and the error escapes React via `window.onerror` so
`global-error.tsx` never runs.

- next.config.mjs: enable `productionBrowserSourceMaps` so client bundles
  ship `.map` files for the Sentry plugin to upload.
- instrumentation-client.ts: add `browserTracingIntegration` (for fetch/XHR
  breadcrumbs) and `captureConsoleIntegration` (for console.error trail);
  raise `maxBreadcrumbs` to 100 so the parallel home-page fetches all fit.
- global-error-trap.tsx (new): re-captures unhandled stack overflows with
  extra browser context and shows a snackbar instead of a silent freeze.

Doesn't fix the underlying recursion — it makes the next occurrence
debuggable and the user-visible symptom less dire.

https://claude.ai/code/session_01JimwdLeNjq2aXCGBbNwrXf

### DIFF
--- a/packages/shared/i18n/locales/en-US/common.json
+++ b/packages/shared/i18n/locales/en-US/common.json
@@ -1,4 +1,7 @@
 {
+  "stackOverflowTrap": {
+    "message": "Something looped on this page. We logged it — try reloading."
+  },
   "languageSwitcher": {
     "ariaLabelWithCurrent": "Language: {{language}}"
   },
--- a/packages/shared/i18n/locales/es/common.json
+++ b/packages/shared/i18n/locales/es/common.json
@@ -1,4 +1,7 @@
 {
+  "stackOverflowTrap": {
+    "message": "Algo entró en bucle en esta página. Lo registramos — prueba a recargar."
+  },
   "languageSwitcher": {
     "ariaLabelWithCurrent": "Idioma: {{language}}"
   },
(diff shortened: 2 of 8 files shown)


marcodejongh:
Rebased onto `main` @79c1962a5 (was ~1466 commits behind). One mechanical conflict in `packages/web/app/layout.tsx` — main added `<NativeDeepLinkListener />` inside `SnackbarProvider` and dropped the `NavigationLoadingProvider` import; resolved to keep both that and this PR's `<GlobalErrorTrap />`. `next.config.mjs` auto-merged.

Addressed review feedback:
- Suppress Sentry's duplicate `window.onerror` capture via `event.preventDefault()` in the trap handler.
- Scoped `browserTracingIntegration` propagation headers to `boardsesh.com` (`tracePropagationTargets`) so they don't ride Aurora-proxy/third-party requests.
- Moved the snackbar copy into the `common` i18n catalog (en-US/es/fr) — this was the `i18n-strings` CI failure; component now reads `t('stackOverflowTrap.message')`.
- Test cleanup uses testing-library `cleanup()` instead of resetting `innerHTML`, so the window `error` listener unmounts between cases.

Left `productionBrowserSourceMaps: true` as authored — it's the core mechanism of this PR (symbolicating prod iOS stacks); the public-exposure tradeoff is documented in the config comment.

Validation: `vp run typecheck:web` ✓, `vp check` (8 changed files) ✓, `vp run check:i18n` ✓, catalog-completeness (28) ✓, `global-error-trap.test.tsx` (4 passed). Backend integration tests left for CI.

PR #1789, opened 2026-04-30
build(web): tree-shake generated GraphQL Documents map via SWC plugin

marcodejongh:
## Summary

`@graphql-codegen/client-preset` emits `packages/web/app/lib/graphql/generated/gql.ts` with a `Documents` map keyed on raw query strings. Without a build-time transform, the map ships in every client bundle that imports `graphql()` — not tree-shakeable, not minifiable, no dead-code elimination. This blocks migrating the remaining op files in `packages/web/app/lib/graphql/operations/` to the typed `graphql()` helper.

This PR wires the SWC team's contrib plugin (`@swc-contrib/plugin-graphql-codegen-client-preset`, the supported successor to the deprecated `@graphql-codegen/client-preset-swc-plugin`) into `next.config.mjs` under `experimental.swcPlugins`. The plugin rewrites each `graphql(\`...\`)` call site at compile time into a direct import of the corresponding `*Document` constant from `generated/graphql.ts`, so unused operations become dead code and the bundler can drop them.

- Pinned to `0.21.0` exactly — SWC plugin compatibility is ABI-tied. Next 16.1+ ships a stable plugin ABI so the pin remains valid across patch releases.
- `artifactDirectory` is `./packages/web/app/lib/graphql/generated`. The wasm plugin resolves this against the **workspace root** (where `bun.lock` lives), not `next.config.mjs`'s dir or `next build`'s cwd — empirically verified: `./app/lib/graphql/generated` produces `Module not found: Can't resolve './../../../../../../app/lib/graphql/generated/graphql'` (six `..` from `account.ts` lands at the workspace root).
- `gqlTagName: 'graphql'` matches the codegen client-preset default.
- Runs only during `next build` — `next dev` still bundles the full map locally. (Build engine for this codebase is whatever `next build` uses; the build log on Next 16.1.6 shows `▲ Next.js 16.1.6 (Turbopack)`. SWC plugins are honoured in either case.)
- No `codegen.ts` change. No operation files migrated.

## Bundle size impact

Measured on the chunk that contains the only live `graphql()` call site today (`packages/web/app/components/settings/delete-account-section.tsx` → `app/lib/graphql/operations/account.ts`):

| | Before | After | Delta |
|---|---|---|---|
| Raw | 167,164 B | 28,419 B | **−138.7 KB / −83%** |
| Gzipped | 17,359 B | 8,417 B | **−8.9 KB / −51%** |
| Total `static/chunks/` dir | 6,628,941 B | 6,490,196 B | **−135 KB** |

Spot-checked: `GetBetaLinks`, `GetGlobalCommentFeed`, `GetUserAscentsFeed`, `UserFavoritesCounts`, `GetNewClimbFeed` all appeared in the baseline settings chunk and are now absent.

The other 18 op files in `operations/` still use `gql` from `graphql-request`, so their bundles are unchanged here. Migrating them to `graphql()` is the next PR — that's where the transform's payoff really shows up.

## Regression check

Added `packages/web/scripts/verify-graphql-treeshake.ts` and a `vp run boardsesh-monorepo#verify:graphql-treeshake` task that depends on `build:web`. The script:

- Parses `gql.ts` for every operation name (matching `query|mutation|subscription`)
- Parses `account.ts` for the names that should be in the marker chunk
- Treats the difference as the must-be-absent set (auto-updates as files migrate, no manual list)
- Reads `chunks/` recursively so nested per-route chunks (`.next/static/chunks/app/...`) are covered if Next switches to that layout
- Asserts the marker chunk doesn't contain any must-be-absent operation name

Verified end-to-end:

- Plugin enabled (correct path) → script passes: `OK — 1 chunk(s) carry GetDeleteAccountInfo (28419 B total), 95 unrelated operation(s) checked, none leaked.`
- `swcPlugins` block removed entirely → script fails: `SWC plugin transform did not fire — Documents map leaked into settings chunk(s): 05b369a70a9e6f0f.js (167164 B): GetBetaLinks, ClimbStatsHistory, ..., +89 more`
- Wrong path → build itself fails fast with `Module not found`, before the script runs.

Wiring this into a CI workflow that already builds the web app is a follow-up — `test.yml` doesn't currently run `next build`, so adding it would expand scope. The vp task is ready to plug into any future build job.

## Test plan

- [x] `vp run codegen` — no diff in `packages/web/app/lib/graphql/generated/` or `packages/shared-schema/src/generated/`
- [x] `vp run boardsesh-monorepo#build:web` succeeds (exit 0)
- [x] `vp run boardsesh-monorepo#typecheck:web` succeeds — pilot file `delete-account-section.tsx` still infers `GET_DELETE_ACCOUNT_INFO` and `DELETE_ACCOUNT` types correctly
- [x] `vp run boardsesh-monorepo#verify:graphql-treeshake` passes against the working build (95 operations checked)
- [x] Same script fails when `swcPlugins` is removed (regression behaviour confirmed)
- [x] Recursion works against simulated nested chunks (`chunks/app/settings/test-nested.js`)
- [x] Rebased on latest `main`
- [ ] CI green

https://claude.ai/code/session_01DAW3jHJoPTim2dy3HsxgGm

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -14,6 +14,7 @@
         "@graphql-codegen/client-preset": "^4.5.1",
         "@graphql-codegen/typescript": "^4.1.2",
         "@graphql-codegen/typescript-resolvers": "^4.4.1",
+        "@swc-contrib/plugin-graphql-codegen-client-preset": "0.21.0",
         "@vitest/coverage-v8": "^4.1.5",
         "oxlint-tsgolint": "^0.21.1",
         "vite": "^8.0.9",
@@ -1299,6 +1300,8 @@
     "@swaggerexpert/cookie": ["@swaggerexpert/cookie@2.0.2", "", { "dependencies": { "apg-lite": "^1.0.3" } }, "sha512-DPI8YJ0Vznk4CT+ekn3rcFNq1uQwvUHZhH6WvTSPD0YKBIlMS9ur2RYKghXuxxOiqOam/i4lHJH4xTIiTgs3Mg=="],
 
     "@swaggerexpert/json-pointer": ["@swaggerexpert/json-pointer@2.10.2", "", { "dependencies": { "apg-lite": "^1.0.4" } }, "sha512-qMx1nOrzoB+PF+pzb26Q4Tc2sOlrx9Ba2UBNX9hB31Omrq+QoZ2Gly0KLrQWw4Of1AQ4J9lnD+XOdwOdcdXqqw=="],
+
+    "@swc-contrib/plugin-graphql-codegen-client-preset": ["@swc-contrib/plugin-graphql-codegen-client-preset@0.21.0", "", {}, "sha512-Jjtdj3HHk2XOu031dU/ieiKgHtE9DaS5BIvhVmqXdarNuu+BGBO9lZr/sSkBJAPB5ARzuDHfmHGvKajHYK0mXw=="],
 
     "@swc/helpers": ["@swc/helpers@0.5.15", "", { "dependencies": { "tslib": "^2.8.0" } }, "sha512-JQ5TuMi45Owi4/BIMAJBoSQoOJu12oOk/gADqlcUL9JEdHB8vyjUSsxqeNXnmXHjYKMi2WcYtezGEEhqUI/E2g=="],
 
--- a/bun.lock
+++ b/bun.lock
@@ -14,7 +14,6 @@
         "@graphql-codegen/client-preset": "^4.5.1",
         "@graphql-codegen/typescript": "^4.1.2",
         "@graphql-codegen/typescript-resolvers": "^4.4.1",
-        "@swc-contrib/plugin-graphql-codegen-client-preset": "0.21.0",
         "@vitest/coverage-v8": "^4.1.5",
         "oxlint-tsgolint": "^0.21.1",
         "vite": "^8.0.9",
@@ -243,6 +242,7 @@
       },
       "devDependencies": {
         "@playwright/test": "^1.58.2",
+        "@swc-contrib/plugin-graphql-codegen-client-preset": "0.21.0",
         "@testing-library/dom": "^10.4.1",
         "@testing-library/jest-dom": "^6.9.1",
         "@testing-library/react": "^16.3.2",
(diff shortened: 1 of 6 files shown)


marcodejongh:
Thanks for the review. Pushed `ffac34cf` on top of a rebased base; force-pushed.

**On `artifactDirectory`** — I tested both paths empirically. `./app/lib/graphql/generated` (the package-relative form) produces `Module not found: Can't resolve './../../../../../../app/lib/graphql/generated/graphql'` from `account.ts` — six `..`s lands at the workspace root, so the wasm plugin is resolving the option against the workspace root, not `packages/web/`. `./packages/web/app/lib/graphql/generated` is what makes the build succeed. Bundle measurement confirmed (167 KB → 28 KB, unrelated query strings absent from the settings chunk). Documented in a comment inline.

**On dev mode** — agreed, added a comment that the transform only runs during `next build`. Worth noting Next 16's `next build` does run Turbopack with swcPlugins (verified — the failed-path build I tested came back with `Turbopack build failed: Module not found...`); `next dev` is where it doesn't fire.

**On the regression check** — added `packages/web/scripts/verify-graphql-treeshake.ts` and a `vp run boardsesh-monorepo#verify:graphql-treeshake` task that depends on `build:web`. The script greps the chunk holding `GetDeleteAccountInfo` and asserts unrelated operation names are absent. I verified it both ways: passes with the plugin on, fails with `SWC plugin transform did not fire — Documents map leaked into settings chunk(s): ... 167164 B: GetBetaLinks, GetGlobalCommentFeed, ...` when I removed `swcPlugins` entirely. CI integration is a follow-up — `test.yml` doesn't currently run `next build`, so the vp task is ready to plug in once a build job exists.

---

marcodejongh:
Addressed the two minor script nits in `c9f99bf6`:

- **Stale list** — now derived dynamically. `extractOperationNames(gql.ts)` returns every operation in the project, `extractOperationNames(account.ts)` returns the ones live in the marker chunk's source; the absent set is just the difference. As more files migrate to `graphql()`, the live list grows and the absent list shrinks automatically with zero maintenance.
- **Double read** — chunk contents are now cached in a `Map<name, string>` keyed by chunk filename.

Verified: passes the working build (94 unrelated ops checked, none leaked) and still flags the no-op baseline (`05b369a70a9e6f0f.js (167164 B): GetBetaLinks, ClimbStatsHistory, GetGlobalCommentFeed, GetComments, GetVoteSummary, +89 more`).

Skipping the cancelled E2E shard 1 — it ran 25 min vs ~5 min for the other 7 shards (timeout/external cancellation, no graphql or SWC code paths touched), and the other 17 checks all passed.

---

marcodejongh:
Pushed `d01ec7eb` for the medium + low script issues:

- **Recursive `readdirSync`** — added `{ recursive: true }`. Top-level chunks happen to cover this build, but Next App Router can split per-route under `.next/static/chunks/app/...` and that would silently miss the marker. Verified by faking a nested chunk: count went from 1 to 2.
- **`subscription` regex** — added to the alternation in `extractOperationNames`. Operation count went from 94 → 95 against current `gql.ts` (so something in there is a subscription that wasn't being checked before).
- **`accountPath` comment** — added a note that it's the only `graphql()` call site as of this PR and how to extend coverage when more files migrate.

For the **PR description**: rephrased the build-engine line. The repo's actual build log shows `▲ Next.js 16.1.6 (Turbopack)`, which is why I'd written "Turbopack production build" — but the wording was ambiguous about whether that's universal Next 16 behaviour or just this codebase's. Description now says "during `next build`" with a parenthetical noting what this codebase happens to use, and explicitly that SWC plugins are honoured by either engine.

For **CI integration**: still a follow-up since `test.yml` doesn't run `next build`. Happy to land that as a separate PR — it'd add a new job that builds web and runs the verify task. Want me to spin that up after this lands, or leave it until the operation-file migration PR (which will exercise the transform on real bundles)?

---

marcodejongh:
Pushed `0dcdbfd7`:

- **Dep moved** to `packages/web/devDependencies`. Bun hoists it to the workspace root so plugin resolution in `next.config.mjs` is unchanged. Verified with a clean `bun install` + rebuild.
- **Auto-discovery of `graphql()` call sites** — script walks `operations/*.ts` and includes any file whose source imports `graphql` from the generated module. The hardcoded `accountPath` is gone. Today it picks up only `account.ts`; when more files migrate, the live set extends with zero script changes. Output now reads: `OK — 1 chunk(s) covering 2 live operation(s) from [account.ts] (28419 B total), 95 unrelated operation(s) checked, none leaked`.
- **`[marker]` brittleness** — replaced single-marker logic. A chunk is "live" if it contains *any* live operation. Forward-compatible with splitting `account.ts` across multiple files.
- **Bonus precision fix** — while implementing the above I caught a false positive: the previous `content.includes(opName)` check was matching the generated TypeScript input type `DeleteAccountInput` against the operation `DeleteAccount`. Now the script searches for `value:"<Name>"` (the parsed Document AST encoding), which can't collide with input-type identifiers. This made the negative case detection more accurate too.

**On the `bun packages/web/scripts/...` task command** — this matches the existing in-tree convention. The `codegen` task in `vite.config.ts` already uses exactly the same pattern with an explicit comment justifying it: `// Direct binary invocation — no \`bunx\` (won't touch the lockfile).` The CLAUDE.md rule's intent (per its phrasing "validation must go through `vp` — never `bun run`, `bunx`, or `npx`") is about user-facing command invocations and the lockfile-touching forms of bun. `bun <script.ts>` as a `vp run` task's underlying command is the established escape hatch when no `vp` wrapper exists for a custom script — same shape as the codegen task. Happy to convert if you want a different convention, but I'd flag the codegen task in the same change to keep it consistent.

**On the CI gap** — agreed. I'll land a follow-up PR right after this one merges to add `vp run boardsesh-monorepo#verify:graphql-treeshake` to a build job in `test.yml`. Treating it as a required predecessor before the operations migration PR.

---

PR #1947, opened 2026-05-04
Auto-bump iOS / Android marketing version from store API

marcodejongh:
## Summary

Both release workflows now query their respective store APIs at build time to derive the next user-visible version. After a release ships on the App Store or Play Store, the next main build automatically increments past it, so we no longer queue a second build under the same `1.2` / `1.1` once an earlier one has been promoted.

- **iOS** (`.github/workflows/ios-testflight.yml`): new step calls App Store Connect (`/v1/apps/{id}/appStoreVersions?filter[appStoreState]=READY_FOR_SALE`) using the existing `.p8` key and runs `agvtool new-marketing-version` before the archive step.
- **Android** (`.github/workflows/android-release.yml`): the existing "Set version code" step is extended to also query the Play Developer API edits/tracks endpoint and `sed` the new `versionName` into `app/build.gradle`.
- Build numbers (`CURRENT_PROJECT_VERSION` / `versionCode`) continue to come from `github.run_number` — only the marketing version is new.
- Logic: `next = max(source, bumpPatch(latest_released))`. If the store has no public release yet, source ships unchanged. The bump is **ephemeral** in CI — we do not commit it back, matching how `versionCode` already works.
- Reuses existing auth in both workflows. No new secrets.
- Helper logic lives in two zero-dependency Node scripts under `mobile/scripts/`, so the same lookup can be run locally for debugging.

## Required before merging

The iOS path needs a new repository **variable** (not a secret — the App Store app id is public):

- `vars.APP_STORE_APP_ID` — the numeric App Store id (the digits in the App Store Connect URL, e.g. `1234567890`).

If unset, the iOS step prints a notice and skips the bump; the build proceeds with whatever's committed in `pbxproj`.

## Notes

- iOS and Android stay independent — each app tracks its own marketing version. Source files are currently `MARKETING_VERSION = 1.2` (iOS) and `versionName = "1.1"` (Android). Both will normalize to 3-segment (`1.2.0`, `1.1.0`) on the next bump and continue from there.
- Edge cases covered: cold start, source ahead of store, store skipped ahead, repeated builds before a promotion, legacy 2-segment store strings, double-digit minor versions (`1.10.0` vs `1.9.0`). Verified with an inline test pass against the bump helpers.
- API failures fail the workflow loud rather than silently falling back — a stale upload would be rejected later by the store anyway, and a re-run is cheap.

## Test plan

- [ ] Set `vars.APP_STORE_APP_ID` in repo settings.
- [ ] Trigger `ios-testflight.yml` via `workflow_dispatch`; check the "Set marketing version" step output and confirm the resulting TestFlight build shows the bumped marketing version + run-number build number.
- [ ] Trigger `android-release.yml` via `workflow_dispatch`; confirm the Play Console internal track shows the bumped `versionName`.
- [ ] Re-run each workflow without a new store release and confirm the same next version is computed (idempotent until the store moves).
- [ ] Once a real release ships, confirm the next main build picks up the next bump automatically.

https://claude.ai/code/session_01NVnVe8ucnzWdNXNg5RCar2

---
_Generated by [Claude Code](https://claude.ai/code/session_01NVnVe8ucnzWdNXNg5RCar2)_

### DIFF
--- a/.github/workflows/android-release.yml
+++ b/.github/workflows/android-release.yml
@@ -63,11 +63,21 @@ jobs:
         working-directory: mobile
         run: bunx cap sync android
 
-      - name: Set version code
+      - name: Set version code and version name
         working-directory: mobile/android
+        env:
+          GOOGLE_PLAY_SERVICE_ACCOUNT_JSON: ${{ secrets.GOOGLE_PLAY_SERVICE_ACCOUNT_JSON }}
+          BUILD_GRADLE_PATH: ${{ github.workspace }}/mobile/android/app/build.gradle
         run: |
           sed -Ei "s/^([[:space:]]*)versionCode[[:space:]]*=.*$/\\1versionCode = ${{ github.run_number }}/" app/build.gradle
           echo "Version code set to ${{ github.run_number }}"
+          if [ -n "$GOOGLE_PLAY_SERVICE_ACCOUNT_JSON" ]; then
+            VERSION_NAME_NEXT=$(node "$GITHUB_WORKSPACE/mobile/scripts/next-android-version-name.mjs")
+            sed -Ei "s/^([[:space:]]*)versionName[[:space:]]*=.*$/\\1versionName = \"$VERSION_NAME_NEXT\"/" app/build.gradle
+            echo "Version name set to $VERSION_NAME_NEXT"
+          else
+            echo "GOOGLE_PLAY_SERVICE_ACCOUNT_JSON not set; leaving versionName unchanged."
+          fi
 
       - name: Build debug APK
         working-directory: mobile/android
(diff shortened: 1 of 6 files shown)


marcodejongh:
Rebased onto `main` @ 79c1962a5 (was ~2 months behind). No conflicts — the two workflow YAMLs still integrate cleanly around main's `Set version code` (Android) and `.p8` decode (iOS) steps, and the `mobile/scripts/*.mjs` files are net-new.

Fixed the critical review finding that was still open: `die()` called `process.exit(1)`, bypassing the `finally` block in `next-android-version-name.mjs` that discards the Play draft edit (a failure mid-request left the edit open → next upload fails with HTTP 409). `die()` now throws; both entrypoints catch via a new `failAndExit()` and exit non-zero after the `finally` cleanup runs. Verified with an inline test that the `finally` block now executes when `die()` fires inside the `try`.

Validation: `vp check` (lint + format) green on all 6 changed files; the 18 `version-utils.test.mjs` cases pass under `node --test`; both entrypoint scripts pass `node --check`. (Repo-wide `vp run typecheck` surfaces a pre-existing `@boardsesh/queue-runtime` build-ordering miss in this fresh worktree — untouched by this PR, which contains zero TS/package changes; CI's ordered build is green.)

These workflows target the legacy Capacitor app (still releasing — `ios-testflight.yml`/`android-release.yml` last ran successfully 2026-06-06), so the feature is still applicable.

PR #1950, opened 2026-05-04
Add Zone search filter — drag a rectangle to limit results to climbs that fit inside it

### LINKED ISSUE #1840: [UX] Climb: allow include/exclude regions
If you want to limit climbs to short ones for a scared friend who's trying it for the first time, or avoid climbing over a spot where the matting is dodgy, it would be nice to be able to specify a region of the board to avoid or stick to, rather than only being able to select individual holds.

The UI complexity could get out of control, so I feel like axis-aligned rectangular regions should be sufficient to handle most cases while still being simple. Alternatively we could keep it as individual holds that are included/excluded, but allow "painting" them with a finger swipe or mouse drag, instead of just tapping/clicking one hold at a time.

marcodejongh:
Closes #1840.

## Summary

Adds a new **Zone** panel to the climb search drawer. Tap **Draw zone** and a translucent rectangle appears over the board — drag the centre to move it, drag the corners to resize. Search results are restricted to climbs whose entire bounding box sits inside the zone.

## How it works

- The rectangle is stored in **board grid coordinates** (same space as `board_holes.x/y` and `board_climbs.edge_*`), so filtering is a four-column range check against the already-denormalised edge columns on `board_climbs` — no extra joins, no client-side conversion when persisting.
- SVG pixel coords are mapped to grid coords via the inverse of `getScreenCTM()` and the same `xSpacing`/`ySpacing` math used in `board-constants.ts`.
- Local state mirrors the search params so dragging stays smooth — only the final pointer-up commits to the debounced search.

## Backend

- New `ZoneBoxInput` GraphQL input type on `ClimbSearchInput`.
- `create-climb-filters.ts` adds the SQL filter when `zoneBox` is present:
  ```sql
  edge_left   >= zoneEdgeLeft
  AND edge_right  <= zoneEdgeRight
  AND edge_bottom >= zoneEdgeBottom
  AND edge_top    <= zoneEdgeTop
  ```
- Zod refines the input to require a non-empty box.

## Web

- `SearchRequest.zoneBox: ZoneBox | null`, with URL serialisation as four `zoneEdge*` query params.
- New `climb-zone-search-form.tsx` component using pointer events + `setPointerCapture` for smooth touch + mouse drag.
- New `Zone` accordion panel; existing `Holds` panel is unchanged.
- Default zone covers the inner 60% of the board so the user has a visible rectangle to grab as soon as they tap **Draw zone**.

## Drive-by

Drops two stale French keys (`home.install.androidPreLaunchTitle` / `androidPreLaunchDescription`) left behind by the recent _Drop Android pre-launch countdown_ change, which were tripping the catalog-completeness test.

## Test plan
- [ ] Open the climb search drawer on Kilter, expand **Zone**, tap **Draw zone**, drag the rectangle around the board and confirm results update.
- [ ] Resize via each corner handle; confirm the box can't shrink past the minimum or escape the board.
- [ ] Tap **Clear zone**; confirm the search count returns to the unfiltered total.
- [ ] Reload the page mid-zoom; confirm the zone is restored from the URL.
- [ ] Same flow on Tension to confirm the grid math holds.
- [ ] On mobile (touch), confirm one-finger drag works for both move and resize.

https://claude.ai/code/session_01KHUogwgFpwygPahcgw53bZ

---
_Generated by [Claude Code](https://claude.ai/code/session_01KHUogwgFpwygPahcgw53bZ)_


marcodejongh:
image

ES-Alexander:
UI suggestions [in #1840](https://github.com/boardsesh/boardsesh/issues/1840#issuecomment-4375780689).

PR #1960, opened 2026-05-04
fix(#1841): rebuild hold filter with type-aware include/exclude picker

### LINKED ISSUE #1841: [UI] Climb: include/exclude are almost the same colour
It's currently very hard to see the difference between them:

Image

They should be easily differentiable, ideally by both lightness and hue. Maybe green for inclusions instead?

It would be nice if the excluded hold images could be dimmed as well, though I'm not sure how annoying that would be to implement. Maybe it would suffice to just have a semi-transparent dark fill in the existing "exclude" circle indicators?

marcodejongh:
## Summary
- Rebuilds the search-drawer hold filter around the same `HoldTypePicker` popover the create-climb screen uses, with a new `mode='search'` that swaps Start/Mid/Finish/Foot setter swatches for tap-to-cycle search swatches plus an Any wildcard rendered as concentric multi-colour rings.
- Each hold can now carry multiple type filters at once (e.g. `STARTING:include` + `FOOT:exclude`), rendered as concentric SVG rings on the board. Excluded holds get a dim fill + ∅ marker so include/exclude differs on both hue and shape — fixes #1841.
- Backend filter (`packages/db/src/queries/climbs/create-climb-filters.ts`) walks the new shape and emits per-type `EXISTS` / `NOT EXISTS` clauses against `board_climb_holds`, making type-specific exclude real (previously only the global `ANY:exclude` worked).

## Why these specific changes
- **Picker UX**: The reporter pointed out that the previous `#8C4A52` (include) vs `#B8524C` (exclude) pair was nearly indistinguishable. Picking new colours alone wasn't enough — the user requested unifying the hold UX with the setter screen, so this PR moves the entire interaction onto the same picker component instead of just swapping hex values.
- **Multi-type**: Resolves the user's ask "If any mode or multiple types are used we render multiple circles inside each other, we don't allow stacking of include and exclude on the same hold type selector." The cycle (unset → include → exclude → unset) enforces the no-stacking rule by construction.
- **URL backward compat**: `parseHoldsFilterFromUrl` accepts both the new `hold_142=STARTING:include,FOOT:exclude` format and the legacy single-state form (`hold_142=STARTING` / `hold_142=NOT`), so shared/bookmarked URLs from before this change keep working.

## Files of note
- `packages/web/app/components/create-climb/hold-type-picker.tsx` — discriminated-union picker (`mode='setter' | 'search'`); search mode bypasses `buildOptions` and uses cycle-on-tap swatches with a multi-ring Any.
- `packages/web/app/components/search-drawer/search-hold-filter-overlay.tsx` — new SVG overlay drawing concentric rings per filter; all circles use `pointerEvents='none'` so the underlying interaction layer keeps receiving taps even when the dim-fill exclude treatment paints over them.
- `packages/web/app/components/search-drawer/use-search-hold-picker.ts` — parallel hook to `useHoldTypePicker` for the multi-filter shape.
- `packages/db/src/queries/climbs/create-climb-filters.ts` — per-`(holdId, type, mode)` SQL emission.
- `packages/web/app/lib/url-utils.ts` — new round-trip + legacy fallback.
- Deletes the standalone `BoardHeatmap` component; the search form now uses `BoardRenderer` + `CreateClimbHeatmapOverlay` (which gained an optional `filtersOverride` prop) the same way create-climb does.

## Test plan
- [ ] `vp run dev` → open search drawer → Holds tab on Kilter
- [ ] Tap a blank hold: it should immediately get an `Any:include` filter (multi-colour rings) and the picker should open anchored to that hold.
- [ ] Tap Start → ring appears alongside the Any rings. Tap Start again → ring switches to the exclude treatment (dim fill + ∅). Tap Start a third time → that ring disappears.
- [ ] Tap Clear → all filters on the hold are removed.
- [ ] Toggle the heatmap button (top-right of the hold form) — heatmap renders on top and clicks still work on excluded holds (regression check on the `pointerEvents` fix).
- [ ] Copy URL with filters set → paste in new tab → filters round-trip.
- [ ] Hand-craft a URL with `hold_142=STARTING` (legacy) → loads as `{142: {STARTING: 'include'}}`.
- [ ] Repeat picker flow on Tension and MoonBoard. MoonBoard's picker should omit the Foot swatch.
- [ ] With `STARTING:exclude` set on a hold, the result count drops as climbs that use that hold as a start are filtered out (verify the per-type exclude SQL).


ES-Alexander:
Noting that it would be nice to support a boolean OR (if) in addition to the existing AND (iff) and NOT.

Description/label wise that could be:
- Any of
- All of
- None of

That said, not a blocker for this PR, just potentially worth adding while the UI is already being changed.

marcodejongh:
@ES-Alexander Yeah I did think of those, but I fear it would become too complicated in the UI. I figured theres 2 primary usecases:
1. Wanting to find climbs in a certain range of holds (your example of not wanting to climb too high)
2. Wanting to find climbs that either avoid or use specific holds (wanting to avoid jugs, or use a particularly gnarly crimp)

For 1 it must be OR, because otherwise it doesnt work, so for that we use the Area thing.

For 2, it'll almost always be AND, it's less likely someone would want to find a climb that includes the gnarly crimp OR the other gnarly crimp

This PR does still allow for either inclusion or exclusion, by hold type, or by all hold types

the different hold types stack, white is any hold type. Then theres a blocked thing in the center when
Screenshot 2026-05-05 at 15 03 24

PR #1966, opened 2026-05-04
Attach ticks to the named board they were climbed on (#1843)

### LINKED ISSUE #1843: [Bug] Gym metadata counts aren't working
My local gym shows up with 0 sends, despite me having done several boardsesh sessions there, using the relevant boards (not custom ones). I also checked a few different regions on the map and found that all gyms had 0 ascents and climbers. A board I "followed" a while ago registered as having 1 follower, but following a new board from the map did not increase the count.

If these stats aren't working globally then they should be hidden until they are. I'm curious whether it's caused by a remapping to the underlying board type when a given gym board is selected, that then stops the completed climbs from being associated with the gym.

marcodejongh:
## Summary

- Adds optional `boardUuid` to `SaveTickInput`. When the user is on a named-board route (`/b/<slug>/...`), `BoardProvider` carries the active board's UUID and injects it into every `saveTick` call (mirroring how `sessionId` is injected today).
- The resolver looks up `user_boards` by uuid and uses that id directly — no ownership check on the climber, just a public-or-owned guard so clients can't tick against someone else's private board.
- The legacy `/[board_name]/[layout_id]/...` config route is unchanged; it still falls back to `resolveBoardFromPath` because there's no specific board entity in scope.

Fixes #1843 — before this, `resolveBoardFromPath` only matched `user_boards` rows where `owner_id = userId`, so every tick on a seeded gym board (owned by the system user) landed with `board_id = NULL` and the map's ascent/climber counts always read 0.

## Test plan

- [ ] Sign in as `test@boardsesh.com` / `test`. Navigate to a seeded gym from the map drawer (Find a board) → drill into one of its boards. URL should be `/b/<slug>/<angle>/list`.
- [ ] Tick a climb. The `SaveTick` request should include `input.boardUuid`. Confirm in the DB: `select board_id from boardsesh_ticks where uuid = '<tick-uuid>'` returns a non-null id.
- [ ] Reopen the board's detail drawer → the Ascents / Climbers chips should reflect the new tick (1 / 1, or higher if you tick more). Same for the board card on the map carousel.
- [ ] Negative path: tick a climb from `/[board_name]/[layout_id]/[size_id]/[set_ids]/[angle]/list`. The `SaveTick` request should omit `boardUuid`; `boardId` resolves via the existing config-based fallback (your own board, or null).
- [ ] Spot-check unrelated tick paths still work: `logascent-form`, `quick-tick-bar` (queue/play views).

### DIFF
--- a/packages/backend/src/graphql/resolvers/ticks/mutations.ts
+++ b/packages/backend/src/graphql/resolvers/ticks/mutations.ts
@@ -1,5 +1,5 @@
 import { v4 as uuidv4 } from 'uuid';
-import { eq, and, inArray } from 'drizzle-orm';
+import { eq, and, inArray, isNull } from 'drizzle-orm';
 import { GraphQLError } from 'graphql';
 import type { ConnectionContext, TickStatus } from '@boardsesh/shared-schema';
 import { db } from '../../../db/client';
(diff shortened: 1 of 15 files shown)


ES-Alexander:
> clients can't tick against someone else's private board.

Probably a niche consideration, but does that mean if a friend has a board and you go to their place to climb, you can't log the climbs as happening there?

marcodejongh:
Rebased onto `main` @19b7ad6a3 (was ~1466 commits behind). The bug is still live in `main`: `resolveBoardFromPath` (packages/backend/src/graphql/resolvers/social/boards.ts) still matches only `ownerId = userId`, so ticks on seeded gym boards land with `board_id = NULL`. Issue #1843 is still open.

Conflicts resolved:
- `packages/backend/.../ticks/mutations.ts` — kept main's `findInstagramShortcodeConflict` refactor; re-added `normalizeSetIds` + the boardUuid resolver/guard block; import trimmed to `eq, and, inArray, isNull`.
- The web `use-save-tick.ts` hook and the monolithic `BoardProvider` were extracted into `@boardsesh/board-react` on main (c4e26edb7). Re-targeted the web changes there: `boardUuid` added to `SaveTickOptions` (tick-helpers.ts), threaded through `use-save-tick.ts` variables, and wired into the shared `BoardProvider` (new `boardUuid` prop + ref-injection into `saveTick` + context value). The web `board-provider-context.tsx` wrapper now forwards `boardUuid` to the shared provider. `app/b/[board_slug]/[angle]/layout.tsx` passes `boardUuid={board.uuid}` (auto-merged).

Fixes: one mechanical rebase-fallout type error — `board-page-climbs-list.test.tsx` (added after this branch) mocks the shared `BoardContextType`, which now requires `boardUuid`; added `boardUuid: null`.

Validation: `vp run typecheck` green; `vp lint`/`vp check` clean for all changed files (repo-wide pre-existing lint/format warnings are unrelated); board-react tests 50/50, affected web tests 21/21, backend unit (beta-link-write-gate) 17/17. Backend integration + E2E left to CI.

Note: the prior Claude/Codex review asks for `ID` over `String` on `boardUuid` and dedicated unit tests for the resolver paths. Those are real follow-ups (schema codegen + new tests), left out of this mechanical rebase to keep scope tight.

PR #1969, opened 2026-05-05
fix(climbs): route ascents-DESC search through stats-driven path

marcodejongh:
## Summary

Production Railway Postgres logs were flooded with errors of the form:

```
ERROR:  could not resize shared memory segment "/PostgreSQL.<id>" to 16777216 bytes: No space left on device
LOG:  background worker "parallel worker" (PID …) exited with exit code 1
```

The default board landing query (sort by ascents DESC, no stats filters) was hitting the LEFT-JOIN path (`standardSearch`) in `packages/db/src/queries/climbs/search-climbs.ts`. The planner compiled it to a **Parallel Bitmap Heap Scan + Nested Loop Left Join + Parallel Sort**. Each parallel context allocated a 16 MB DSM segment in `/dev/shm`; under concurrent traffic those segments saturated the 1 GB `shm_size` cap and parallel workers started failing with `ENOSPC`.

The codebase already has a fast path (`statsDrivenSearch`) that drives FROM `board_climb_stats` and walks the covering index in sorted order. It was gated off (`hasStatsFilters` requirement) for queries without stats filters because its INNER JOIN drops climbs without a stats row at the queried angle. UX requires those stats-less climbs to appear at the bottom of narrow-filter results so users can find projects.

This change relaxes the `useStatsDriven` gate to fire for ascents-DESC regardless of `hasStatsFilters`, with a fallback to `standardSearch` when the stats-driven page is short (i.e., narrow filter where stats-less climbs need to fill the page). The fallback runs over a small slice of the result set, so the planner picks a serial plan and doesn't allocate parallel DSM.

### Verified against production data (kilter layout 1, angle 40, default filter)

| Path | Plan | Time | Buffer reads |
|---|---|---|---|
| Before (LEFT JOIN, parallel) | Parallel Bitmap + Nested Loop + Parallel Sort | **980 ms** | 776,641 |
| After (stats-driven, serial) | Index Only Scan + Memoize Nested Loop | **3.5 ms** | 121 |

~280× faster, ~6,400× fewer buffer reads, zero DSM allocation on the hot path.

### Why not raise `shm_size` or switch to `dynamic_shared_memory_type = mmap`

Both mask the symptom while leaving every concurrent request running a 100× more expensive plan than needed. The query layer is the right place to fix this — the index-driven plan was already implemented, just gated off.

## Test plan

- [ ] Watch Railway Postgres logs for ~30 min after deploy under normal traffic. Expect zero new `could not resize shared memory segment` entries and zero `parallel worker exited with exit code 1` entries tied to that error.
- [ ] Re-run `EXPLAIN (ANALYZE, BUFFERS)` on a representative failing query post-deploy. Confirm the plan is `Index Only Scan on board_climb_stats_ascents_covering_idx` + `Nested Loop` + early `Limit`. No `Parallel`, no `Hash`.
- [ ] Latency smoke test: hit the climb-search page on prod across a few board/layout combinations. Confirm p50/p95 latency is within or below current bounds.
- [ ] Narrow-filter UX check: filter to a hold-set + size combo with ≤5 matching climbs and confirm projects (climbs without stats) still appear at the bottom of the result list.
- [ ] Pagination spot-check: page through default kilter layout 1 results to confirm next-page links work and no rows are dropped at boundaries.

### DIFF
--- a/packages/db/src/queries/climbs/__tests__/search-climbs.test.ts
+++ b/packages/db/src/queries/climbs/__tests__/search-climbs.test.ts
@@ -1,0 +1,87 @@
+import { describe, it } from 'node:test';
+import assert from 'node:assert/strict';
+import { chooseSearchPath } from '../search-climbs';
+
+const baseInput = {
+  sortBy: 'ascents',
+  sortOrder: 'desc' as const,
+  isDraftsQuery: false,
+  projectsOnly: false,
+  page: 0,
+  hasStatsFilters: false,
+};
+
+void describe('chooseSearchPath', () => {
+  void describe('the hot path: ascents DESC, page 0, no stats filters', () => {
+    void it('uses stats-driven-with-fallback so projects appear at the bottom of narrow-filter pages', () => {
+      assert.equal(chooseSearchPath(baseInput), 'stats-driven-with-fallback');
+    });
+  });
+
+  void describe('pages > 0', () => {
+    void it('uses stats-driven-only on page 1 — fallback would re-create DSM pressure', () => {
+      assert.equal(chooseSearchPath({ ...baseInput, page: 1 }), 'stats-driven-only');
+    });
+
+    void it('uses stats-driven-only on a deep page', () => {
+      assert.equal(chooseSearchPath({ ...baseInput, page: 47 }), 'stats-driven-only');
+    });
+  });
+
+  void describe('stats filters active (e.g. minAscents >= 1)', () => {
+    void it('uses stats-driven-only on page 0 — stats-less climbs would be filtered out anyway', () => {
+      assert.equal(chooseSearchPath({ ...baseInput, hasStatsFilters: true }), 'stats-driven-only');
+    });
+
+    void it('uses stats-driven-only on deeper pages with stats filters', () => {
+      assert.equal(chooseSearchPath({ ...baseInput, page: 5, hasStatsFilters: true }), 'stats-driven-only');
+    });
+  });
+
+  void describe('cases that bypass the stats-driven path entirely', () => {
+    void it('uses standard-only when projectsOnly is set (user wants stats-less climbs)', () => {
+      assert.equal(chooseSearchPath({ ...baseInput, projectsOnly: true }), 'standard-only');
+    });
+
+    void it('uses standard-only for drafts queries (drafts have no stats rows)', () => {
+      assert.equal(chooseSearchPath({ ...baseInput, isDraftsQuery: true }), 'standard-only');
+    });
+
+    void it('uses standard-only for non-ascents sort orders', () => {
+      assert.equal(chooseSearchPath({ ...baseInput, sortBy: 'difficulty' }), 'standard-only');
+      assert.equal(chooseSearchPath({ ...baseInput, sortBy: 'name' }), 'standard-only');
+      assert.equal(chooseSearchPath({ ...baseInput, sortBy: 'creation' }), 'standard-only');
+      assert.equal(chooseSearchPath({ ...baseInput, sortBy: 'quality' }), 'standard-only');
+      assert.equal(chooseSearchPath({ ...baseInput, sortBy: 'popular' }), 'standard-only');
+    });
+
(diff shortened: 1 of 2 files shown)


marcodejongh:
Thanks for the catch — both issues are real. Pushed `0de88b1` addressing them.

## What I found verifying the suggested fixes

I ran the alternatives against production via `EXPLAIN (ANALYZE, BUFFERS)`:

| Approach | Page 0 / narrow | Deep page (broad filter) |
|---|---|---|
| NOT-EXISTS-only stats-less fallback | ~5 ms serial (Index Scan Backward + Anti Join — clean) | ~109 ms **parallel** (Parallel Index Scan + Anti Join) |
| `COUNT(*)` of stats-having to compute stats-less offset | n/a | ~**862 ms parallel** — basically the original failing query |
| `NOT IN (...)` with already-fetched UUIDs | n/a | UUIDs from page N don't help with offset N+1 — caller would need to carry server-side state across requests |

The COUNT-based approach turned out to be a dead end: the count itself is the same parallel-bitmap-heap-scan + nested-loop pattern we just stopped firing. Not viable.

## What I did

Took the third option you suggested — fall back only when `page === 0 && !hasStatsFilters`. That's the case where the projects-at-bottom UX actually matters (narrow filter, few stats-having matches, all on the first page). The fallback runs over a small dataset (filter selective enough that stats-driven returned <pageSize), so the planner picks a serial plan and doesn't allocate parallel DSM. Verified with `EXPLAIN ANALYZE` on `kilter/layout=8/set={27,29}` (17 matches): 55 ms serial.

On pages > 0 with a short stats-driven result, the function now returns the partial result as-is rather than re-running standardSearch. Trade-off documented in the comment: stats-less climbs past the stats-having boundary on broad-filter pagination become unreachable. Real users rarely paginate that deep on a popularity sort, so this is acceptable; the alternative (running the slow plan on every deep page) was worse.

## Wasted-statsResult observability

I left the explicit logging out of this PR for now — adding ad-hoc `console.info` would be noise without a real telemetry pipeline behind it. With the page-0-only gate, the wasted-fallback case is much narrower (only page 0 narrow-filter searches, which are a small fraction of traffic) so the cost is also much lower. Happy to add a metric in a follow-up if it's worth tracking.

## QA on the diff

- `vp run typecheck:db` — clean.
- Search-touching backend tests (`climb-queries.test.ts`, `graphql-resolvers.test.ts`) — 29/29 pass.

marcodejongh:
Pushed `9acb8b9` addressing both points.

## Pagination gap (tracked in code)

Expanded the inline comment in `searchClimbs` to spell out the trade-off, the reasoning behind accepting it (verified `~862 ms` count + `~109 ms` parallel stats-less query on broad filters via production EXPLAIN — both re-create the DSM pressure we just fixed), and a forward-looking note: this is a candidate for the popular_climbs materialized view + keyset pagination follow-up I'd already mentioned in earlier discussion.

I held off on adding `console.info` from this layer (`packages/db` shouldn't carry app-level instrumentation) — when the team adds proper telemetry the wrapper at `packages/web/app/lib/db/queries/climbs/search-climbs.ts` is a cleaner place to count fallback firings. Will leave that as a follow-up issue.

## Tests

Extracted the routing decision as a pure function `chooseSearchPath(input)`. Added 12 unit tests (`packages/db/src/queries/climbs/__tests__/search-climbs.test.ts`) covering each branch:

- Hot path (ascents-DESC, page 0, no stats filters → `stats-driven-with-fallback`)
- Pages > 0 (→ `stats-driven-only`, no DSM-risky fallback)
- `hasStatsFilters` active (→ `stats-driven-only`, INNER JOIN already correct)
- `projectsOnly`, drafts, non-ascents sorts, ascending sort (→ `standard-only`)
- Precedence cases (e.g. projectsOnly trumps the page-0 hot path)

The helper is a pure function so tests don't need a DB or mocks. Run via the existing `node:test` infrastructure that `create-climb-filters.test.ts` already uses.

```
# tests 12  # pass 12  # fail 0
```

Backend search tests (`climb-queries.test.ts`, `graphql-resolvers.test.ts`) still pass 29/29 after the refactor.

PR #1998, opened 2026-05-05
fix(playlists): pin grid + paginated scroll (#1592)

### LINKED ISSUE #1592: [Bug] Playlists not showing - only 8 of 23 appear on Boardsesh
## Description

A user reported that only 8 out of 23 playlists appear in Boardsesh, even though all 23 were present in the Aurora JSON export.

## Steps to Reproduce

1. Have a large number of playlists (20+) synced from Aurora
2. Open Boardsesh playlist view
3. Only a subset of playlists are visible

## Additional Context

- All playlists are for the same board (Kilter)
- The playlists **are** present in the Aurora JSON export, so the issue is in Boardsesh's import/display logic
- Possibly related to a filter, pagination limit, or a cap on the number of playlists fetched/stored
- User noted that when they first joined the app, some climbs appeared to be for the wrong board (wrong layout/size) — some playlists may have been associated with an incorrect layout and are being filtered out

## Possible Causes

- Hard-coded limit on the number of playlists fetched or displayed
- Playlists associated with a layout/size the user no longer has active are being silently filtered

marcodejongh:
## Summary

- Replaces the static 8-card grid + 10-card scroll on `/playlists` (which silently capped Annette's 23-playlist library at 8 visible) with a Pinned grid + paginated infinite-scroll.
- Pinning is server-backed via a new `user_playlist_pins` table (datamodel mirrors `user_follows`). When the user hasn't pinned anything yet, the grid falls back to per-device IndexedDB recent-playlists so the slot isn't wasted.
- "Jump Back In" is now a paginated horizontal slider; `allUserPlaylists` takes `page`/`pageSize` and returns a connection-shaped result. Reuses the existing `BoardScrollSection` IntersectionObserver pattern.
- Test user seed now creates 40 playlists across boards (incl. 6 Aurora-synced rows with `layoutId=null`) and pins 3, so QA hits the bug-repro shape directly.

The "wrong board association" hypothesis from the issue is debunked: the resolver already does `OR(eq(layoutId), isNull(layoutId))` and Aurora circuits land with `layoutId=NULL`, so they always match. The bug was purely UI-side (`slice(0, 8)` + `slice(0, 10)` caps).

Closes #1592.

## Test plan

- [ ] `vp run dev` → sign in as `test@boardsesh.com` / `test`
- [ ] `/playlists` shows the new "Pinned" grid with 3 seeded pins (Kilter Original Project 1, Kilter 10x10 Warmup 2, TB2 Crimps & Pinches 3)
- [ ] "Jump Back In" scrolls horizontally and loads more pages on right-edge approach (DevTools network panel: `GetAllUserPlaylists` with incrementing `page`)
- [ ] Pin button on a card toggles between filled/hollow `PushPin`; survives reload
- [ ] Unpin removes the card from the grid; recents may refill the slot
- [ ] Open a playlist (any) → IndexedDB `boardsesh-recent-playlists` records the open; navigate back → it appears in the recents fallback when no server pins for that board
- [ ] BoardFilterStrip switches both pinned + scroll re-query
- [ ] All 6 `layoutId=null` Aurora-synced rows are reachable in the scroll regardless of selected board
- [ ] Sign out → only Discover + sign-in banner show; pinned grid hidden
- [ ] `vp check`, `vp run typecheck`, `vp test --project web` (4097/4097), `vp test --project backend` (8/8 unit tests for `allUserPlaylists` resolver)

## What's not in scope

- The `circuit_climbs` USER_TABLES gap in Aurora sync (separate latent issue if confirmed) — this PR is about visibility count, not climb content
- Pinning private playlists owned by other users — `verifyPlaylistAccess` allows it for shared/own; tightening that is a follow-up
- Renaming "Jump Back In" — kept the existing translation key
- Virtualization for >100 playlists — revisit if a user reports lag


packages/backend/src/graphql/resolvers/playlists/helpers/pin-stats.ts
```diff
@@ -0,0 +1,25 @@
...
+ * Batch-fetch the set of playlist UUIDs that the given user has pinned, out
+ * of a candidate list. Returns an empty Set when the user is not
+ * authenticated (so unauthenticated callers always see `isPinnedByMe = false`
+ * without an extra round-trip).
+ *
+ * Mirrors getPlaylistFollowStats but pins are scoped per-user only — there
+ * is no aggregate pin count. The userPlaylistPins table joins by playlist
+ * id, so we resolve uuids first.
+ */
+export async function getPlaylistPinSet(playlistUuids: string[], currentUserId: string | null): Promise<Set<string>> {
```
marcodejongh: Why do we need a resolver we need to pass the playlist uuids into?! Wouldnt it be better to just query the playlists and join the pin tables to get the ones the user has pinned?

PR #2012, opened 2026-05-07
perf(motion): silky-smooth pill collapse + FAB cluster motion

marcodejongh:
## Summary

Tunes the bottom tab bar pill and queue-control FAB cluster animations from generic-`ease`-with-overlapping-transitions to a single-curve, layer-promoted, iOS-style decelerate. On top of the motion polish, this branch makes a few small visible changes (called out below) and an iOS-only safe-area refinement.

- **Bottom tab bar pill** — layer-promote (`will-change: transform, opacity`) so backdrop-filter rasterization stays on its own GPU layer; drop the per-element compound transitions that competed with the pill's own motion; replace `display: none` on labels with `opacity: 0 + pointer-events: none` so the fade can actually play instead of hard-cutting at frame 0; iOS decel curve `cubic-bezier(0.32, 0.72, 0, 1)` at 220ms baked into the `transitions.snappy` token (matches the shape of the other `transitions.*` tokens). `aria-hidden` on labels when collapsed so screen readers don't announce the zero-font-size text.
- **iOS native shell** — pill drops flush with the bottom of the screen so its rounded bottom edges nest into the iPhone's curved screen corners. Home-indicator clearance moves *inside* the pill via `--tab-bar-safe-area-padding`. Android keeps the existing 8px lift because gesture-nav clearance is more variable.
- **FAB cluster** — kept on the `bottom`-swap pattern (a `transform` on `.fabRoot` would create a containing block that breaks backdrop-filter sampling on every FAB child — they'd blur the empty parent instead of the page behind). Matched the snappy curve so it moves in lockstep with the pill.
- **Peek snackbar** — wrapped in a `.peekSnackbarAnchor` that owns the collapse-tracking `translateY(var(--fab-collapse-offset))` (compositor-only). The snackbar is opaque, so transform on the ancestor is safe. Inner snackbar keeps its entrance/exit keyframes — two layers, one transform each.
- **`cardWrapper`** — easing-token swap so the queue-control bar's minimisation reads as one motion family with the FABs.
- **Reduced-motion** — vertical slide disabled, opacity fades retained as the cue. Vestibular-sensitive users see no movement.

### Visible behavioral changes

- **Queue-count badge dropped from the session mini bar's open-queue button.** The list-icon already opens the drawer where the count is visible — the badge added visual noise without telling the user anything new. (commit 5b42589eb)
- **Setter byline removed from the climb-change peek snackbar.** The peek now shows only the climb name; the byline crowded the snackbar without earning its space.

No new components, hooks, or external animation library. Blur strength preserved (12px / 24px).

## Test plan

- [ ] On `/<board>/.../list`, scroll the climbs list — the pill should collapse and re-expand without visible step transitions or label hard-cut.
- [ ] Tap a collapsed pill once — first tap expands instead of navigating (existing behaviour, kept intact).
- [ ] On a queue-bar page, minimise the bar (scroll-collapse) so the FAB cluster appears, then collapse the tab bar — FAB cluster should glide down with the pill and stay frosted (no transparent-window regression).
- [ ] Trigger a peek by changing the current climb — snackbar should slide up showing only the climb name, sit ~3s, slide down. If the tab bar collapses while the snackbar is up, it should track the bar smoothly.
- [ ] DevTools → rendering → `Emulate CSS prefers-reduced-motion: reduce`. Re-trigger collapses — should fall back to plain opacity fades, no slide.
- [ ] DevTools Performance — record a 5s scroll session crossing the pill collapse/expand threshold a couple of times. Frame rate should stay near 60fps.
- [ ] On the iOS Capacitor app, confirm the pill's bottom curves visually nest into the iPhone's screen corners (no gap to the bottom edge) and the home-indicator gesture still works.
- [ ] Confirm the queue-icon button in the session mini bar no longer renders a count badge; opening the drawer still shows the queued items.
- [ ] Confirm the climb-change peek snackbar shows only the climb name (no "By <setter>" line).

### DIFF
--- a/docs/ai-design-guidelines.md
+++ b/docs/ai-design-guidelines.md
@@ -589,33 +589,36 @@ The queue drawer dynamically expands from partial to full height based on scroll
 
 ### 7. Bottom Tab Bar
 
-Mobile-only navigation bar with glassmorphism effect.
+Mobile primary navigation. A floating **liquid-glass pill** (not an edge-to-edge bar) anchored above the safe area, with five tabs: Home, Climbs, Discover, Create, You. (Feed is folded under `/you/feed` — there's no top-level Feed tab anymore.) The pill scroll-collapses into a small Safari-style chrome on downward scroll and re-expands on upward scroll or near-top.
 
-**Styling:**
+**Pill chrome:**
 
-```css
-background: rgba(255, 255, 255, 0.5);
--webkit-backdrop-filter: blur(10px);
-backdrop-filter: blur(10px);
-padding-top: 4px;
-padding-bottom: env(safe-area-inset-bottom, 0px);
-```
+- Shape: rounded pill (`border-radius: 9999px`), `max-width: min(560px, 100%)`, centered horizontally with a `var(--spacing-2)` bottom gap above the safe area.
+- Liquid glass: layered translucent surface + top-edge highlight gradient + 1px hairline border. Backdrop-filter strength differs per theme — light mode uses `blur(12px) saturate(180%)`, dark mode uses `blur(24px) saturate(180%)` to keep contrast over busy content. Tokens live under `themeTokens.glass.*` / `darkTokens.glass.*`.
+- The pill sits in a `position: fixed` wrapper at `z-index: 10`. The wrapper's `pointer-events: none` lets taps pass through gaps; the pill itself accepts events.
 
-**Tab items:**
+**Tab items (expanded):**
 
-- Layout: Flex column, centered
-- Padding: `6px 0 4px`
-- Icon size: `20px`
-- Label: `10px`, `margin-top: 2px`, `line-height: 1`
-- Active color: `var(--color-primary)` (`#8C4A52`)
-- Inactive color: `var(--neutral-400)` (`#9CA3AF`)
-- Transition: `color 150ms ease`
-- Touch handling: `-webkit-tap-highlight-color: transparent`, `touch-action: manipulation`
+- Layout: Flex row inside the pill (MUI `BottomNavigation` + `BottomNavigationAction`).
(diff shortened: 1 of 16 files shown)


marcodejongh:
Pushing two follow-ups for the review feedback:

**Bug — `display: none` kills the opacity transition** — Already addressed in this PR. `display: none` was removed in the original perf commit; labels now fade via `opacity` + `pointer-events: none` (see `bottom-tab-bar.tsx` lines 488–497, the comment block explicitly calls out why `display: none` was avoided).

**Docs — section 7 stale** — Rewrote `docs/ai-design-guidelines.md` section 7 to reflect the floating liquid-glass pill: 5 tabs (Feed folded under `/you`), per-theme blur strengths (12 / 24px), scroll-collapse + first-tap-to-expand, the iOS-decel motion family, and the layer-promotion / reduced-motion rules. Commit: `8818a6d1f`.

**Minor — first-tap-to-expand applies uniformly to all tabs** — Keeping current behaviour. At 14 px icons in a 28 px chrome the targets are equally hard to hit regardless of which tab; singling out Home would create an inconsistency users have to learn. Mirrors how Safari's collapsed URL bar accepts a single tap to expand before any further interaction. Pre-existing decision from `0b53fe086`, out of scope for the perf PR — happy to revisit in a follow-up if QA shows Home is dependably hittable at the small size.

**Minor — `robots.ts` `/feed` disallow stale** — Dropped the entry. `/feed` is a 308 to `/you/feed`, so leaving it in only confused crawlers. Same commit `8818a6d1f`.

**Bonus — icon vertical-centring in collapsed pill** — Caught while QA-ing: MUI's `BottomNavigationAction` is `flex-direction: column; justify-content: center`, and an `opacity: 0` label still claims its line-height, so the icon was sitting at the top of an icon-plus-invisible-label group. Now animates `font-size` 12 → 0 alongside the opacity fade so the label box collapses and the icon re-centres. Commit: `2315d4c92`.

PR #2019, opened 2026-05-07
Spare climb shape in disco; add LED color picker

lilylilylily123:
## Summary

- Disco mode now re-rolls colours only on HAND placements; STARTING/FOOT/FINISH keep their canonical role colours every tick so the route's shape stays readable while the hands strobe
- New "Customize LED colors" entry in the lights drawer opens a dialog with HAND / FOOT / FINISH colour swatches that persist in IndexedDB and are applied at BLE packet build time
- STARTING stays at the board's canonical green (climbing convention — first-look cue)
- Threads an optional \`colorOverrides\` map through \`getAuroraBluetoothPacket\` keyed by \`HoldState\`; falls back to \`HOLD_STATE_MAP\` when a state isn't overridden. MoonBoard's separate packet path doesn't share this lookup, so the entry is disabled there

## Test plan

- [ ] Pair a Kilter or Tension board, load any climb, hit **Disco the climb** — only HAND holds flicker; START/FOOT/FINISH stay at their canonical colours
- [ ] **Customize LED colors** → pick custom hex for Hands, Feet, Finish → send a climb → verify BLE LEDs render the chosen colours; reload page → values persist
- [ ] **Reset** clears all overrides; the climb re-renders with defaults
- [ ] Toggle disco while a custom hand colour is set — disco's random palette shows during the show; on stop, the climb returns with the custom hand colour applied
- [ ] MoonBoard: customize entry is disabled with the "isn't available" subtitle
- [ ] No climb loaded: disco button stays disabled with the existing prompt

## Notes

- The SVG board renderer still uses the canonical role colours; only the physical BLE LEDs honour overrides. SVG palette rewiring is a much larger surface and intentionally out of scope here
- French \`common.json\` was missing the disco keys from the prior commit — added the new \`customizeColors\` keys plus backfilled the prior gap so the i18n catalog completeness test for \`fr/common.json\` now passes

### DIFF
--- a/packages/web/app/components/board-bluetooth-control/__tests__/use-board-bluetooth.test.ts
+++ b/packages/web/app/components/board-bluetooth-control/__tests__/use-board-bluetooth.test.ts
@@ -270,7 +270,7 @@ describe('useBoardBluetooth', () => {
 
     expect(sendResult).toBe(true);
     expect(mockGetLedPlacements).toHaveBeenCalledWith('kilter', 1, 10);
-    expect(mockGetAuroraBluetoothPacket).toHaveBeenCalledWith('p4131r42', { 4131: 39 }, 'kilter', 3);
+    expect(mockGetAuroraBluetoothPacket).toHaveBeenCalledWith('p4131r42', { 4131: 39 }, 'kilter', 3, undefined);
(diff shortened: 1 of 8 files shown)


lilylilylily123:
@marcodejongh ready for review, everything looks pretty good

PR #2032, opened 2026-05-09
Fix climb list scroll on new search

### LINKED ISSUE #1583: [UX] Jump new searches to the top
Making a new search while scrolled down in an existing one starts partially scrolled down the page in the new results. It would make more sense for every new search to start with the first results.

giannilariosa:
Fixes #1583

## Summary

- move new-search scroll reset into the shared climb list replacement path
- remove the narrower `page=0` URL-param scroll hook from `BoardPageClimbsList`
- add a regression test that verifies replacing the climb list scrolls back to the top

## Notes

The previous hook only reacted when the URL explicitly contained `page=0`, but the default page is commonly omitted from the query string. This instead treats a replaced climb list as the primitive for a new search, while preserving the existing no-scroll behavior for appended infinite-scroll results.

## Testing

- `vp test run packages/web/app/components/board-page/__tests__/climbs-list-virtualization.test.tsx`
- `vp fmt --check packages/web/app/components/board-page/climbs-list.tsx packages/web/app/components/board-page/board-page-climbs-list.tsx packages/web/app/components/board-page/__tests__/climbs-list-virtualization.test.tsx`
- `bun run --filter=@boardsesh/web typecheck`
- `git diff --check`

### DIFF
--- a/packages/web/app/components/board-page/__tests__/climbs-list-virtualization.test.tsx
+++ b/packages/web/app/components/board-page/__tests__/climbs-list-virtualization.test.tsx
@@ -267,6 +267,29 @@ describe('ClimbsList virtualization', () => {
     const items = screen.getAllByTestId('climb-list-item');
     expect(items).toHaveLength(5);
   });
+
+  it('scrolls to the top when a new search replaces the climb list', () => {
+    const scrollTo = vi.fn();
+    Object.defineProperty(window, 'scrollTo', {
+      configurable: true,
+      value: scrollTo,
+    });
+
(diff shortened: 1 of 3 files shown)


giannilariosa:
Manual local test passed.

I tested the branch in the browser on the Kilter Homewall 10x12 climb list by scrolling down, applying a name search (`moon`), and waiting for the results to replace the list. The page scrolled back to the top of the new results as expected.

marcodejongh:
Thanks!

PR #2044, opened 2026-05-10
[codex] Migrate analytics to PostHog with safe admin filtering

marcodejongh:
## Summary

- Adds PostHog dual-write analytics alongside existing Vercel Analytics.
- Adds client-side PostHog pageview and PostHog-only web-vitals reporting.
- Wires PostHog identity handling for anonymous, logged-in, and logout flows.
- Moves the Vercel `/admin` exclusion into a dedicated client component so the filter does not pass function props through the server layout boundary.

## Why

The PostHog migration gives us retention/cohort analytics while keeping Vercel Analytics during the transition. The admin filtering change keeps internal admin routes out of Vercel Analytics and Speed Insights without reintroducing the SSR 500 issue from the inline server-layout `beforeSend` functions.

## Validation

- `bun run test:run app/lib/__tests__/analytics.test.ts app/components/__tests__/analytics-client.test.tsx app/lib/__tests__/analytics-paths.test.ts app/lib/__tests__/posthog-alias-storage.test.ts app/components/party-manager/__tests__/party-profile-context.test.tsx`
- `bun run --filter=@boardsesh/web lint`
- `bun run --filter=@boardsesh/web typecheck`
- `bun run --filter=@boardsesh/web test:run`
- `bun run --filter=@boardsesh/web build`

## Deploy Notes

Set `NEXT_PUBLIC_POSTHOG_KEY` in the Vercel production environment before relying on PostHog capture. The client remains hostname-gated to `boardsesh.com`.


marcodejongh:
Addressed the review feedback in `ea4f12668`:

- Admin analytics exclusion is now shared and locale-aware (`/admin`, `/es/admin`, `/fr/admin`, and subpaths).
- PostHog pageviews and web-vitals now skip admin routes, and pageviews send path-only URLs instead of raw query strings.
- The analytics wrapper also suppresses custom analytics calls while the current page is admin.
- PostHog alias calls are now guarded by durable localStorage state keyed by anonymous profile/user pair, so reloads do not repeatedly emit the same alias.
- Added focused coverage for admin path matching, alias storage, and `AnalyticsClient` pageview behavior.

Validation rerun:

- `bun run test:run app/lib/__tests__/analytics-paths.test.ts app/lib/__tests__/posthog-alias-storage.test.ts app/components/__tests__/analytics-client.test.tsx`
- `bun run --filter=@boardsesh/web typecheck`
- `bun run --filter=@boardsesh/web build`

PR #2057, opened 2026-05-11
feat(analytics): PostHog historical signup + logbook backfill

marcodejongh:
## Summary

Adds a one-off script (`packages/web/scripts/backfill-posthog-signups.ts`) that backfills PostHog with historical user data so retention and activation funnels become measurable for existing users — not just users who signed up after PostHog was deployed in #2044.

For each user the script POSTs up to 3 events to PostHog's `/batch/` endpoint with `historical_migration: true`:

- **`$identify`** at `users.created_at` — sets `email`, `name`, `logbook_entry_count` via `$set` and `signup_date`, `signup_year_month`, `backfilled`, `first_logbook_entry_at`, `first_logbook_board_type` via `$set_once`.
- **`user_signed_up`** at `users.created_at` — gives PostHog retention insights a real Day 0 anchor.
- **`first_logbook_entry`** at `MIN(boardsesh_ticks.climbed_at)` — only emitted if the user has any logbook activity. Properties: `tick_status` (attempt/flash/send), `board_type`, `entry_count_at_backfill`. Anchors activation funnel insights.

### Why it works

- `historical_migration: true` is PostHog's documented flag for backfilling events with original timestamps — routes the batch through historical ingestion and skips spike detection.
- Per-event UUID is `uuidv5(userId + ':<kind>', NAMESPACE)`. PostHog dedupes on `(distinct_id, event, timestamp, uuid)`, so reruns produce identical UUIDs and no duplicate events.
- `distinct_id = users.id` matches the existing identify flow in `party-profile-context.tsx`, so the next time a backfilled user logs in, the existing `alias()` flow merges anonymous → known cleanly.
- Logbook stats come from a CTE over `boardsesh_ticks` — uses the existing `(user_id, climbed_at)` index.

### Caveat

For users who signed up _before_ PostHog was added, we have a backdated Day 0 but no return events from the pre-PostHog window. So historical retention curves will read ~0% beyond Day 0 for that period — expected, not a bug. New signups going forward get clean retention because the events are anchored to a real `user_signed_up`. The activation funnel (signup → first_logbook_entry) **does** work for historical users because `boardsesh_ticks` is in our own DB.

## Test plan

- [x] `vp check` passes on the new script
- [x] Dry-run against dev DB renders correct payload shape: `bun packages/web/scripts/backfill-posthog-signups.ts --dry-run --limit 3`
- [x] UUIDs are deterministic across reruns (verified locally)
- [ ] Single-user smoke test against prod PostHog: `bun packages/web/scripts/backfill-posthog-signups.ts --user-id <known-id>` → check PostHog Persons UI for `signup_date`, `first_logbook_entry_at`, and the timeline events at their original timestamps
- [ ] Full prod run after smoke test passes

## Flags

- `--dry-run` — print sample batch for first 3 users, don't POST
- `--limit N` — only process first N users
- `--user-id <id>` — backfill a single user (for the smoke test step)
- `--skip-logbook` — skip the tick aggregate query and only send signup events

### DIFF
--- a/packages/web/package.json
+++ b/packages/web/package.json
@@ -7,6 +7,7 @@
     "build": "bun run generate:openapi && next build",
     "build:ci": "bun run generate:openapi && next build",
     "generate:openapi": "tsx scripts/generate-openapi.ts",
+    "backfill:posthog": "tsx scripts/backfill-posthog-signups.ts",
     "start": "next start",
     "lint": "vp lint",
     "lint-fix": "vp lint --fix",
--- a/packages/web/scripts/backfill-posthog-signups.ts
+++ b/packages/web/scripts/backfill-posthog-signups.ts
@@ -1,0 +1,303 @@
+import { drizzle } from 'drizzle-orm/postgres-js';
+import { eq, sql } from 'drizzle-orm';
+import postgres from 'postgres';
+import { config } from 'dotenv';
+import path from 'path';
+import { fileURLToPath } from 'url';
+import { v5 as uuidv5 } from 'uuid';
+import { users } from '@boardsesh/db/schema/auth';
+
+const __dirname = path.dirname(fileURLToPath(import.meta.url));
+
+config({ path: path.resolve(__dirname, '../../../.boardsesh/dev-db.env') });
+config({ path: path.resolve(__dirname, '../../../.env.local') });
+config({ path: path.resolve(__dirname, '../.env.local') });
+config({ path: path.resolve(__dirname, '../.env.development.local') });
+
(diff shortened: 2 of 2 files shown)


marcodejongh:
Ran the backfill against prod successfully — 990 users / 2545 events posted to PostHog (565 of 990 had logbook activity for the first_logbook_entry event). This is a one-shot script; closing instead of merging since we don't need it in the codebase permanently. Branch left intact as a reference in case we need to rerun (idempotent via deterministic UUIDs).

Issue #12, opened 2024-10-24
Hold filtering. Similar to Climbdex search by hold feature


marcodejongh:
Climbdex supports filtering by hold: https://github.com/lemeryfertitta/Climbdex/blob/main/climbdex/db.py#L251-L277
This should be reimplemented in boardsesh. 

The query from climbdex should be pretty portable, so thats not going to be hard.
Then the board renderer in: app/components/board-renderer/board-litup-holds.tsx will need to have an edit mode added.
In edit mode all circles in the svg should render, and onclick their light up status should change between the hold colours for the currently active board.

marcodejongh:
Done a while ago

PR #2139, opened 2026-05-13
fix(party-session): reland #2128 with ClimbMirrored.uuid alias + subscription backoff

marcodejongh:
## Summary

- Relands PR #2128 (which was reverted in PR #2137 to stop the bleeding) and fixes the GraphQL validation bug that broke party-session joins in prod.
- Aliases `uuid` to `mirroredUuid` on the `ClimbMirrored` selection in `QUEUE_UPDATES` and `EVENTS_REPLAY` — the `OverlappingFieldsCanBeMergedRule` was rejecting every join because `ClimbMirrored.uuid` is nullable `ID` while `QueueItemRemoved.uuid` / `QueueReordered.uuid` are `ID!`.
- Replaces the flat 1 s subscription-recovery delay with exponential backoff (1, 2, 4, 8, 16 s, capped, gives up after `MAX_TRANSIENT_RETRIES`) so we don't tight-loop reconnects again if a future schema regression slips through.

## Context

Production server log (link in chat): 48 `Fields "uuid" conflict` validation errors on every join attempt for session `0f44bc70-13cc-44e7-9c1e-3bba7d9f480c`, with the client reconnecting every 3-50 ms because `scheduleSubscriptionRecovery` retried at a flat 1 s. Two symptoms followed:

1. **Queue updates never sync** — both `queueUpdates` subscription and `eventsReplay` query failed validation.
2. **Joining takes forever** — every join failed, the client hammered reconnects, and the server kept tearing down + recreating the room.

PR #2128 (`5c7fcd523 Fix websocket queue replay hashes`) introduced the bug when it added `uuid: ID` to `ClimbMirrored`. The schema is correct — the resolver legitimately publishes `ClimbMirrored` with `uuid: null` when there is no current climb (`mutations.ts:320`: `const mirroredUuid = currentClimb?.uuid ?? null`) — so we alias on the client side instead of flipping the schema to `ID!`.

## What's in this PR

1. `82b8385a` — `git revert dad2e2d37`, reapplying PR #2128 wholesale (`connectionState`, `UserPresenceChanged`, `leaderConnectionId`, `stateHash` propagation, backend resilience improvements, and the buggy `ClimbMirrored.uuid` selection).
2. `ab394315` — fix commit:
   - `packages/shared-schema/src/operations.ts` — `mirroredUuid: uuid` aliases on `ClimbMirrored` in both `EVENTS_REPLAY` and `QUEUE_UPDATES`.
   - `packages/shared-schema/src/types/events.ts` — `SubscriptionQueueEvent.ClimbMirrored` now uses `mirroredUuid?: string | null`. Server-side `QueueEvent` keeps `uuid` (wire shape unchanged).
   - `packages/web/app/components/persistent-session/hooks/use-event-processor.ts` — reads `event.mirroredUuid`.
   - `packages/web/app/components/persistent-session/hooks/use-session-lifecycle.ts` — explicit `ClimbMirrored` arm in `transformToSubscriptionEvent` (handles both wire and aliased shapes) + exponential backoff with counter reset on successful event / successful join.
   - Tests updated; `session-lifecycle-replay.test.ts` gains a `ClimbMirrored` transform test.
   - Folds in oxfmt cleanup of files that came back with stale formatting via the revert reapply.

## Test plan

- [x] `vp check` — clean for files I touched (one pre-existing repo-wide lint warning in `packages/backend/src/graphql/resolvers/playlists/queries/user-playlists.ts` predates this PR).
- [x] `vp run typecheck` — exit 0.
- [x] `vp test --project web` — 4377 pass; one pre-existing failure in `i18n-catalog-completeness` (`fr/feed.json` drift, unrelated).
- [x] `vp test --project backend` — 937 pass.
- [ ] Manual QA in the running dev server (see `.boardsesh/qa-notes.md`):
  - Two-tab join: tab B's queue list populates within ~1 s.
  - Live add / remove / reorder / current-climb-change from tab A appears in tab B within ~200 ms.
  - **Mirror-toggle the current climb in tab A** — the specific event that was broken — reaches tab B without error.
  - Reload tab B mid-session; replay restores state.
  - No `GraphQLError: Fields "uuid" conflict` in `vp run dev` output during any of the above.

### DIFF
--- a/docs/websocket-implementation.md
+++ b/docs/websocket-implementation.md
@@ -144,8 +144,8 @@ sequenceDiagram
     WS->>C: ConnectionAck
 
     C->>WS: Execute joinSession mutation
-    Note over C,WS: With optional initialQueue & initialCurrentClimb
-    WS->>RM: joinSession(connectionId, sessionId, boardPath, initialQueue?)
+    Note over C,WS: With stable participantId and optional initialQueue/current climb
+    WS->>RM: joinSession(connectionId, sessionId, boardPath, participantId, initialQueue?)
 
     alt Session exists in memory
         RM->>RM: Add client to session
@@ -192,7 +192,8 @@ sequenceDiagram
 2. **Authentication**: Optional auth token passed in `connectionParams`
 3. **Eager Subscription**: Queue subscription starts BEFORE fetching state to prevent race conditions
 4. **Session Restoration**: Sessions can be restored from Redis (warm cache) or PostgreSQL (dormant durable state)
-5. **Initial Queue Seeding**: When creating a new session, clients can provide `initialQueue` and `initialCurrentClimb` to seed the session with an existing local queue (e.g., when starting party mode with climbs already queued)
+5. **Stable Participant Identity**: Anonymous clients persist a `participantId` in `ACTIVE_SESSION_KEY`; authenticated clients use their `userId`. Reconnects therefore update the same participant instead of creating a new user row for every socket.
+6. **Initial Queue Seeding**: When creating a new session, clients can provide `initialQueue` and `initialCurrentClimb` to seed the session with an existing local queue (e.g., when starting party mode with climbs already queued)
 
 ### Initial Queue Seeding
 
@@ -206,6 +207,7 @@ mutation JoinSession(
   $boardPath: String!
   $username: String
   $avatarUrl: String
+  $participantId: ID                       # Optional: stable anonymous participant identity
   $initialQueue: [ClimbQueueItemInput!]    # Optional: existing queue items
   $initialCurrentClimb: ClimbQueueItemInput # Optional: current climb
   $sessionName: String                      # Optional: display name for the session
@@ -215,6 +217,7 @@ mutation JoinSession(
     boardPath: $boardPath
     username: $username
     avatarUrl: $avatarUrl
+    participantId: $participantId
     initialQueue: $initialQueue
     initialCurrentClimb: $initialCurrentClimb
     sessionName: $sessionName
@@ -226,6 +229,7 @@ mutation JoinSession(
 
 - `initialQueue`, `initialCurrentClimb`, and `sessionName` are **only applied when creating a new session**
 - If joining an existing session (active, warm, or dormant), these values are ignored and the existing session state is used
+- `participantId` is optional for backward compatibility, but current clients always send it for anonymous sessions so reconnects preserve identity
 - The queue is persisted immediately to Postgres (not debounced) to ensure durability for new sessions
 - All users who join after the initial seed will receive the seeded queue state
 
@@ -358,7 +362,8 @@ The main Traefik instance routes `*.preview.boardsesh.com` and `*.ws.preview.boa
     │  - Redis cache hot                           │
     │  - In-memory session state                   │
     └──────┬───────────────────────┬───────────────┘
-           │ Last user leaves      │ Leader calls endSession
+           │ Last participant      │ Participant calls endSession
+           │ leaves/expires        │
            ▼                       ▼
     ┌──────────────────────────┐   ┌──────────────────────────────┐
     │   GRACE PERIOD (60s)     │   │     ENDED (explicit)         │
@@ -397,7 +402,9 @@ The main Traefik instance routes `*.preview.boardsesh.com` and `*.ws.preview.boa
     └────────────────────┘
 ```
 
-**Grace Period:** When the last user disconnects from this backend instance, the session enters a 60-second grace period where in-memory state is preserved. If a client reconnects within this window (common during network flaps or page refreshes), the session is instantly available without the expensive lock + Redis/Postgres restoration cycle. The grace period duration is controlled by `SESSION_GRACE_PERIOD_MS` in `RoomManager`.
+**Grace Period:** When the last socket on a backend instance drops unexpectedly, the instance enters a 60-second grace period where in-memory state and pending queue writes are preserved. The participant is marked `RECONNECTING`, not removed. If a client reconnects within this window (common during network flaps or page refreshes), the same `participantId` is marked `CONNECTED` and the session is instantly available without the expensive lock + Redis/Postgres restoration cycle. The grace period duration is controlled by `SESSION_GRACE_PERIOD_MS` in `RoomManager`.
+
+Explicit UI actions still leave or end sessions immediately: `leaveSession` removes the participant and emits `UserLeft`; `endSession` is restricted to the session creator or current leader. Passive WebSocket disconnects emit `UserPresenceChanged` first and only emit `UserLeft` if the reconnect timer expires.
 
 If the grace period expires, the session is evicted from memory but remains restorable from Redis until the 4-hour TTL expires. After Redis TTL expiry, the session is still not ended; the next join restores it from Postgres as long as the durable row has not been explicitly marked `ended`.
 
@@ -457,6 +464,8 @@ Sessions can be linked to multiple boards within the same gym via the `sessionBo
 
 ### Leader Election
 
+Leader state is retained for compatibility and UI flags, and it is an authorization boundary for ending sessions over WebSocket: `endSession` requires the caller to be the session creator or the current leader. HTTP callers must be the authenticated session creator.
+
 Leader election uses Redis-backed atomic operations for consistency across instances:
 
 **Single Instance Mode:**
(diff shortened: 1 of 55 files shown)


marcodejongh:
## Round 2: code-review follow-ups

New commit `34c65769` addresses the merge-blocker findings from review.

### P1 (security) — fixed
- **Untrusted participantId is now rejected.** `client-lifecycle.ts:102` no longer falls back to the client-supplied value. Authenticated users bind to their verified `userId`; anonymous users get `connectionId`, and reconnects across WebSocket drops correctly appear as fresh participants. Leader impersonation closed.

### Critical — fixed
1. **Atomic `removeParticipantConnection`** — new Lua script `REMOVE_PARTICIPANT_CONNECTION_SCRIPT` collapses the `srem -> smembers -> multi().exec()` sequence into one atomic op. No more window where a concurrent re-join gets wiped.
2. **Grace-timer no longer swallows Redis errors.** A failed `getSessionMembers` now keeps the participant alive instead of expelling them; the next reconnect retries the grace logic.
3. **`leaveSession` ordering** — kept the existing ordering (which the original author documented) and tightened the comment to pin down the idempotency invariant. `markInactive` = SREM, `cancelPendingWrites` = no-op when empty, so double-call is safe.
4. **`ctx.transport: 'http' | 'ws'`** replaces the fragile `connectionId.startsWith('http-')` checks in createSession / endSession. Field is optional so existing test mocks keep working; production paths set it explicitly.
5. **Presence-update TTL-miss** — the existing fallback `|| localParticipantToSessionUser(participant)` already covers this; added a comment explaining the invariant.

### Other — fixed
- **#6 endSession actor lookup** now logs the inputs (connectionId / participantId / userId / isAuthenticated / member count) when the actor lookup fails, so the misleading "Only the session creator or current leader" error has a paper trail.
- **#8 `hasContiguousReplayCoverage`** has an explicit tie-breaker that places FullSync ahead of same-sequence deltas. Added a reverse-order regression test that fails on the previous code.

### Deferred (non-blocking follow-ups)
- **#7 N+1 Redis hops in `getSessionParticipants`** — perf optimization; not a correctness issue. Worth a separate PR.
- **#9 test using connectionId as participantId** — the test is testing the anonymous fallback path explicitly. Documented.
- **#10 `createParticipantId` dedupe** — pure refactor; client-only.
- **#11 `docs/websocket-implementation.md`** — happy to do this in a follow-up so reviewers can read the doc against the merged code. Flag if you'd rather block on it.
- **#12 `RECONNECTING` UI** — the backend now emits the right state; surfacing it in the participant list is a UI change beyond the scope of this fix.

### Validation
- `vp check` — exit 0
- `vp run typecheck` — exit 0
- `vp test --project backend` — **937 pass, 0 fail** (was 3 fail in the first push, the failing tests were validating the vulnerable participantId behavior; updated to use authenticated test users)
- `vp test --project web` — 4377 pass, 2 unrelated failures (i18n `fr/feed.json` drift, quick-tick-bar UI timeout — both pre-existing on origin/main, neither touched by this PR)

marcodejongh:
## Round 3 + manual QA ✓

`0bfeeb89` adds the small follow-ups:
- Doc fix (`docs/websocket-implementation.md`): the "stable participant identity" point now correctly scopes reconnect continuity to **authenticated** sessions. Anonymous clients bind `participantId` to `connectionId` per the P1 fix, so every WS drop is a fresh participant.
- `leaveSession` / `leaveSessionFallback` (`session-ops.ts`) no longer call `removeParticipantConnection` — the explicit-leave caller in `client-lifecycle.ts` already runs `removeParticipant` which subsumes it. Drops one Lua-eval + a multi() on every explicit leave.
- New `session-persistence.test.ts` regression test pins the intentional anonymous-reconnect UX (new connectionId → new participantId → peers see `UserLeft` + `UserJoined`, even if the client replays the previous participantId).

### Manual QA in the running dev server
- Two devices in the same party session, one authenticated leader, one anonymous publisher.
- `addQueueItem`, `setCurrentClimb` from the publisher propagate to the receiver in real time. ✓
- No `Fields "uuid" conflict` errors in the backend log across the entire flow. ✓
- Reconnects on the authenticated device emit `UserPresenceChanged` (continuity); anonymous reconnects emit `UserLeft` + `UserJoined` as designed. ✓

### CI
All 26 checks green on `0bfeeb89` (typecheck, lint, backend test, web test, all 8 E2E shards, screenshots, i18n, codegen-drift). PR is `MERGEABLE`.

Deferred to follow-up issues: N+1 Redis hops in `getSessionParticipants` (#7, perf), `createParticipantId` client dedupe (#10), and surfacing `connectionState: 'RECONNECTING'` in the participant list (#12).

marcodejongh:
## Round 4: reviewer must-fixes addressed

Two Opus 4.7 expert reviewers (distributed-state + client/protocol/security angles) audited this branch. Three critical/major findings fixed in `3d1e6028c`; the remainder filed as follow-up issue #2143.

### Fixed in this commit

| ID | Severity | Summary |
|---|---|---|
| A1 | critical | **Multi-tab leave wiped peer tabs.** Auth user with phone + laptop signed in, click Leave on laptop, phone silently stopped getting events. New `removeParticipantConnection` public method detaches only the leaving connection; participant is fully torn down only when its connection set hits zero. `UserLeft` broadcast suppressed when sibling tabs remain. |
| A2 | critical | **Zombie connection hashes.** `LEAVE_SESSION_SCRIPT` HMSET resurrected a deleted connection hash with no TTL when a concurrent `removeConnection` had just DEL'd it. Guarded with `EXISTS connKey == 1`. Same pattern fixed in `JOIN_SESSION_SCRIPT` (returns -1 if the hash is missing, caller throws) and the `leaveSessionFallback` MULTI. |
| B3 | major | **Subscription retry counter never reset on reconnect.** Low-traffic flaky session — reconnects succeed but no event arrives before the next disconnect — silently force-cleared the session after 5 hiccups. `handleReconnect` now resets the counter on a successful join, mirroring the initial `connect()` path. |
| B1 | prevention | **Operations-schema validation in CI.** New `operations-schema-validation.test.ts` runs `parse(op)` + `validate(schema, doc)` on every exported operation. Catches the OverlappingFieldsCanBeMergedRule class of bug (which broke prod twice) deterministically pre-merge. ~10ms per operation. |

### Deferred to follow-up: [#2143](https://github.com/boardsesh/boardsesh/issues/2143)

A4 (rename invisibility), A6 (grace timer too aggressive on RECONNECTING), A7 (leader TTL not refreshed), A8 (SessionUser.id flicker), A9 (connection vs session TTL mismatch), B2 (endSession stale leader check), B5 (LeaderChanged fallback compares wrong id spaces), B7 (no-op `ClimbMirrored` events), B8 (`useQueueEventSubscription` ignores `mirroredUuid`).

A3 (Lua scripts cluster-unsafe) tracked as a separate Redis-portability concern.

### Validation
- `vp run typecheck` — exit 0
- `vp test --project backend` — **931 pass**, 22 failures all in `integration.test.ts` (port 8082 conflict with the running dev server, not from this PR)
- `vp check` — one pre-existing error in `bluetooth-packet.test.ts` that matches `origin/main` exactly, untouched by this PR
- Manual QA on the dev server confirmed multi-tab behaviour and queue sync still work

marcodejongh:
## Round 5: addressing the latest claude[bot] re-review

`dd4e255da` fixes the new findings from the 02:42 UTC re-review.

### Critical: endSession auth regression (fixed)
The reviewer is correct — the WS branch lost `requireAuthenticated(ctx)` during the reland of PR #2128. `requireSessionMember` alone meant an anonymous member who happened to be leader could end the session. Restored the unconditional auth check at the top of `endSession` so both transports require it; the per-transport authorization logic (creator-only for HTTP, member-and-leader for WS) stays unchanged.

### Major: removeParticipant race (fixed)
Confirmed real. After A1 (round 4) routed explicit-leave through `removeParticipantConnection` (already atomic), the full-eviction `removeParticipant` was still `SMEMBERS → MULTI/EXEC`. The remaining race fires when grace-timer expiry on instance A conflicts with a fresh cross-instance reconnect on instance B. New `REMOVE_PARTICIPANT_SCRIPT` Lua does the snapshot + cleanup in one atomic call, with the same `EXISTS connKey == 1` guard on the connection HMSETs.

### Minor (fixed)
- **Default leaveSession mock missing `participantFullyLeft`** — added to the `vi.mock` default in `session-context.test.ts`. Tests relying on the default now exercise the UserLeft broadcast path; the two `mockResolvedValueOnce` overrides from round 4 are unchanged.
- **Dead `getConnectionForParticipantCleanup`** — removed. Last caller went away in round 3.
- **`hasContiguousReplayCoverage` test gap** — added a `QueueItemAdded` + `FullSync` co-sequence test (mirror of the existing `CurrentClimbChanged` case).

### False positive
- **stateHash written when QueueReordered diverges** — verified the watchdog at `use-session-subscriptions.ts:120-122` compares `lastReceivedStateHash` against `computeQueueStateHash(queue, currentClimbQueueItem)`, i.e. a hash *computed from current local content*. When the reorder handler bails out without updating local state, the server hash gets stored but the locally-computed hash still reflects the (stale) local content, so the two diverge and the watchdog correctly triggers a resync. No fix needed.

### Deferred
- **Doc deprecation note for `isLeader` vs `leaderConnectionId`** — tracked alongside the other doc/UX items in #2143.

### Validation
- `vp check` exit 0
- `vp run typecheck` exit 0
- `vp test --project backend` — 931 pass, 22 failures all in `integration.test.ts` (port 8082 conflict with the running dev server, not from this PR)

PR #2144, opened 2026-05-14
fix(party-session): close #2143 — harden distributed-state correctness & UX

### LINKED ISSUE #2143: party-session: harden distributed-state correctness & UX (follow-up to #2139)
Follow-up to PR #2139 (party-session reland). Two Opus 4.7 reviewers flagged several should-fix issues that were deemed safe to ship as follow-up rather than block #2139. Tracking them here so they don't get lost. Each is independently shippable.

### updateUsername doesn't update the participant hash (A4, major)
**Where:** `packages/backend/src/services/distributed-state/connection-ops.ts:259-271`, `services/distributed-state/session-ops.ts:268-322`

`updateUsername` writes to `KEYS.connection(connectionId)` only. But `getSessionMembers` returns from `KEYS.participant(sessionId, participantId)` once `sessionParticipants` is populated. Peers querying members see the stale username/avatar even though `UserJoined` was re-broadcast.

**Fix:** Add an `HSET` to the participant hash in `updateUsername`, plus refresh `TTL.sessionMembership` on the participant key.

### Grace timer expels participants in RECONNECTING with live connections (A6, major)
**Where:** `packages/backend/src/services/room-manager/client-lifecycle.ts` (the participant `reconnectTimer` setTimeout)

The check `if (currentUser?.connectionState === 'CONNECTED') return;` is too narrow. A participant whose state has flipped to RECONNECTING but who already has at least one live connection (the reconnect WS opened but joinSession hasn't run yet) gets expelled instead of preserved.

**Fix:** Treat any present-with-live-connections participant as alive. Either `if (currentUser) return;` or check via the live-connection count rather than `connectionState`.

### Leader key TTL never refreshed (A7, major)
**Where:** `packages/backend/src/services/distributed-state/lua-scripts.ts` `REFRESH_TTL_SCRIPT`

The script refreshes `connKey`, `sessionMembersKey`, `sessionParticipantsKey`, `participantKey`, and `participantConnectionsKey` — but not `sessionLeaderKey`. After ~4h of activity (the original `EX sessionTTL`), the leader key expires; the next mutation triggers a fresh election and clients see a surprise `LeaderChanged` mid-session.

**Fix:** Add `redis.call('EXPIRE', leaderKey, sessionTTL)` to `REFRESH_TTL_SCRIPT`.

### Brief SessionUser.id flicker between LEAVE_SESSION_SCRIPT and follow-up multi (A8, major)
**Where:** `packages/backend/src/services/distributed-state/session-ops.ts:209-249`

Between LEAVE_SESSION_SCRIPT completing and the follow-up `multi.exec()` that populates `sessionParticipants` / `participantConnections`, a concurrent `getSessionMembers` falls back to the connection-keyed branch and returns `id: connectionId`. Once the multi completes, the same user reappears with `id: participantId`. Peers see a flicker.

**Fix:** Merge the participant `multi()` into `JOIN_SESSION_SCRIPT` so the entire join is atomic, or use `participantId` from the connection hash in the fallback branch.

### Connection TTL (1h) shorter than session membership TTL (4h) (A9, major)
**Where:** `packages/backend/src/services/distributed-state/constants.ts:52-56`

A long-idle leader whose connection hash expires while the session/leader keys still exist leaves the leader key pointing to a dead connection. `getLeaderParticipantId` returns null until `cleanupStaleSessionMembers` runs (~2 min later); during that window, every `SessionUser.isLeader` is false and leadership-keyed mutations misbehave.

**Fix:** Either align `TTL.connection` with `TTL.sessionMembership`, or eagerly re-elect when `getLeaderParticipantId` finds a dead leader.

### endSession leader check uses potentially stale user list (B2, major)
**Where:** `packages/backend/src/graphql/resolvers/sessions/mutations.ts:369-388`

The WS branch trusts `actor.isLeader` from `getSessionUsers`. `SessionUser.isLeader` can be momentarily stale between leader handoff and the next read. A participant briefly in RECONNECTING state on this instance could still have `isLeader === true` locally and accept the endSession call.

**Fix:** Use `distributedState.getLeader(sessionId)` authoritatively and compare against `ctx.participantId`.

### LeaderChanged fallback compares wrong identifier spaces (B5, minor)
**Where:** `packages/web/app/components/persistent-session/hooks/use-session-lifecycle.ts:657-666`

`const leaderConnectionId = event.leaderConnectionId || event.leaderId;` then `isLeader: leaderConnectionId === prev.clientId`. When `leaderConnectionId` is null/missing, the fallback compares `event.leaderId` (participantId) against `prev.clientId` (connectionId). For authenticated users these are different ID spaces — the comparison will never match, and the user is incorrectly told they're not the leader.

**Fix:** Drop the `||` fallback. The backend always sets `leaderConnectionId`; if a future event omits it, handle that case explicitly instead of comparing apples to oranges.

### mirrorCurrentClimb publishes a no-op event when there's no current climb (B7, minor)
**Where:** `packages/backend/src/graphql/resolvers/queue/mutations.ts:311-349`

When `currentClimb == null`, the resolver still publishes a `ClimbMirrored` event with `uuid: null`. The client correctly ignores it as a stale duplicate, but it clutters logs and is fragile under co-sequencing.

**Fix:** Early-return without publishing when there's no current climb.

### useQueueEventSubscription ignores mirroredUuid (B8, minor)
**Where:** `packages/web/app/components/graphql-queue/hooks/use-queue-event-subscription.ts:80-85`

This consumer dispatches `DELTA_MIRROR_CURRENT_CLIMB` without reading `event.mirroredUuid`, so it mirrors whatever the local client thinks is the current climb. In a race where peers are on different climbs, the local view gets mis-mirrored.

**Fix:** Pass `mirroredUuid` through the action and guard the reducer on it. Or remove this code path entirely, since `use-event-processor.ts` already handles the same event correctly.

---

## Out of scope for this issue (already fixed in #2139)
A1 multi-tab leave, A2 zombie HMSETs, A5 join-script zombie HMSET, B1 schema-validation test, B3 reconnect counter reset. See #2139 commit `3d1e6028c`.

## Out of scope, separate concern
A3 (Lua scripts are cluster-unsafe — they build keys from string concatenation inside the script body instead of via `KEYS[]`). Worth tracking separately as a Redis-cluster compatibility / portability concern aligned with the CLAUDE.md "keep storage vanilla" goal.

marcodejongh:
Closes #2143.

Implements every item from the #2139 reviewer follow-up: 6 distributed-state correctness fixes (A4, A6, A7, A8, A9, B2) and 3 client/UX fixes (B5, B7, B8).

## Summary

**Distributed-state**
- **A4** — `updateUsername` writes through to the participant hash via a new `UPDATE_USERNAME_SCRIPT` Lua. Was only updating the connection hash, so peers saw stale names after rename.
- **A6** — Grace-timer "spare participant" check broadened to `if (currentUser) return;` so participants in `RECONNECTING` with the reconnect WS already open aren't expelled.
- **A7** — `REFRESH_TTL_SCRIPT` now refreshes `sessionLeaderKey` too. Long sessions no longer see surprise `LeaderChanged` events when the original election TTL expires.
- **A8** — `JOIN_SESSION_SCRIPT` does the participant writes inline (6 KEYS instead of 3). Eliminates the brief `SessionUser.id` flicker between the script and the prior follow-up `multi.exec()`.
- **A9** — `TTL.connection` raised from 1h to 4h to match `sessionMembership`. Long-idle leaders no longer leave the session leader-less for ~2 minutes.
- **B2** — `endSession` WS path now reads the authoritative leader from `distributedState.getSessionLeader()` and compares against `ctx.connectionId`. Was relying on potentially-stale `SessionUser.isLeader`.

**Client / UX**
- **B5** — Drop the `event.leaderConnectionId || event.leaderId` fallback in `LeaderChanged` handling. The fallback compared participantId against connectionId — wrong id spaces.
- **B7** — `mirrorCurrentClimb` early-returns when there is no current climb instead of publishing a no-op event.
- **B8** — `DELTA_MIRROR_CURRENT_CLIMB` action now carries `mirroredUuid`; reducer suppresses the mutation if the local current climb has drifted to a different uuid.

## Test plan

- [x] `vp check` exit 0
- [x] `vp run typecheck` exit 0
- [x] `vp test --project backend` — 931 pass, 22 failures all in `integration.test.ts` (port 8082 conflict with the running dev server, unrelated)
- [x] `vp test --project web` — 4424 pass, 0 fail
- [x] `MockRedis` updated to simulate the new JOIN_SESSION_SCRIPT shape + 3 new Lua scripts so `session-persistence.test.ts` still exercises the full lifecycle (25/25 pass)
- [ ] Manual QA in dev server: rename mid-session, grace-window reconnect on flaky network, long-idle session (≥1h) doesn't see surprise LeaderChanged, multi-tab leader handoff

## What's intentionally out of scope

**A3** (Lua scripts build extra keys via string concatenation instead of via `KEYS[]`, making them Redis-cluster-unsafe). Tracked as a separate portability concern aligned with CLAUDE.md's "keep storage vanilla" rule.


marcodejongh:
## Round 2: addressing the bot re-review

`d96f88afa` fixes the three real findings.

**High — mock-redis dispatch was unreachable.** The bot was right: `numkeys===1` returned unconditionally for `RELEASE_LOCK_SCRIPT`, so the new `UPDATE_USERNAME_SCRIPT` branch was dead code and the A4 fix was passing vacuously. Switched the dispatch to discriminate by `args.length` (RELEASE_LOCK=2, REFRESH_TTL=3, UPDATE_USERNAME=4). Also added a minimal `REFRESH_TTL_SCRIPT` branch that was previously falling through silently.

**Note (was a real bug) — null-uuid `DELTA_MIRROR_CURRENT_CLIMB` replay applied to the local current climb.** Pre-B7 history events carry `uuid=null` (the server's old no-op publish path). The B8 guard rejected mismatched uuids but null was a hole. Reducer now also bails when `mirroredUuid === null`. New regression test.

**Note — B5 over-corrected for anonymous users.** The bot was right: for anon users, `participantId === connectionId` (P1 fix), so `event.leaderId` IS interchangeable with `event.leaderConnectionId`. Dropping the fallback entirely broke anon users during a rolling deploy from a pre-#2128 backend. Restored the fallback, gated on the local client being anonymous (detected via `userId === null` in the session-user list). Authenticated users still skip the fallback (which would have produced the wrong-id-space comparison the original B5 fix targeted).

**Low — connection TTL refresh inside UPDATE_USERNAME_SCRIPT.** Added the `EXPIRE connKey` so a rename arriving near expiry doesn't inherit a stale TTL.

**A3 pattern extended (note).** UPDATE_USERNAME_SCRIPT does the same string-concat key building as the existing scripts — already tracked separately as the cluster-portability follow-up.

### E2E (shard 8)
`queue-persistence.spec.ts:143 > queue should NOT persist across full page navigations` failed. Single-user queue behaviour, no overlap with the multiplayer paths this PR touches; looks like a flake. Round-2 push should re-trigger CI; if it persists I'll dig in.

### Validation
- `vp check` ✓
- `vp run typecheck` ✓
- backend `session-persistence.test.ts` + `distributed-state.test.ts` — 85/85 pass (the mock fix means these now exercise real behavior, not the previous vacuous path)
- web `reducer.test.ts` + `session-lifecycle-replay.test.ts` — 64/64 pass

marcodejongh:
## Round 3: coverage + docs

`5b4990a52` closes the bot's coverage and minor-polish gaps.

### New backend tests (the bot's main concern)

**B2 — `endSession` authorization (4 tests in `session-context.test.ts`):**
- Rejects a WS member whose `getSessionUsers.isLeader` is stale-true when `distributedState.getSessionLeaderConnectionId` points to someone else. **This is the exact scenario the B2 fix targets** — without it the test suite never proved the change.
- Accepts when the authoritative leader connectionId matches `ctx.connectionId`.
- Accepts the session creator even when they're not the current leader.
- Rejects an unauthenticated WS caller even when they ARE the leader (the round-1 auth-regression guard).

**A4 — `updateUsername` write-through (2 tests in `distributed-state.test.ts`):**
- Registers + joins + renames; asserts `getSessionMembers` returns the new username/avatarUrl — i.e. the **participant hash** was updated, not just the connection hash. (Previously passing vacuously because of the mock-dispatch bug fixed in round 2.)
- `updateUsername` on a reaped connection hash is a no-op (doesn't fabricate a participant).

**A6/A7/A8** are now exercised end-to-end by the existing `session-persistence.test.ts` (25/25 pass) since the round-2 mock-dispatch fix makes those tests actually run the new Lua paths.

### Docs

`docs/websocket-implementation.md` gains four new "Key Points":
- **Atomic Join** (A8) — `JOIN_SESSION_SCRIPT` now collapses what was Lua + a follow-up `multi.exec()`, eliminating the `SessionUser.id` flicker.
- **Grace Window** (A6) — spare check is now "any present participant," not `connectionState === 'CONNECTED'`.
- **TTL Refresh** (A7, A9) — leader key refreshed on activity; connection TTL aligned with `sessionTTL`.
- **Authoritative Leader Check** (B2) — authorization keys off the Redis leader key, not stale `SessionUser.isLeader`.

### Minor polish

- `LeaderChanged` fallback's `isAnonymous` detection now reads the local user's `userId` from `prev.users` instead of a presence check that could be ambiguous mid-join. Comment expanded.
- Stale `mock-redis.ts` comment cleaned up (said "RELEASE_LOCK or UPDATE_USERNAME" but RELEASE_LOCK is already handled earlier).

### Cluster-incompat note (deferred)

`UPDATE_USERNAME_SCRIPT` and `REFRESH_TTL_SCRIPT` both build extra keys via string concat inside the Lua body — same portability concern as the original A3, tracked separately and not in scope for this PR.

### Validation

- `vp check` ✓
- `vp run typecheck` ✓
- Targeted tests: `session-context.test.ts` (12), `distributed-state.test.ts` (62), `session-persistence.test.ts` (25) — **95/95 pass**
- Round 2's E2E shard 8 failure (`queue-persistence.spec.ts:143`) cleared on the next CI run — was a flake, single-user behavior unrelated to this PR.

marcodejongh:
## Round 5: A6 ghost participants fixed

`e58db8222` — the bot was right, this was a real P1 my original A6 broaden introduced.

### Root cause
The original A6 widened the grace-timer "spare" check to `if (currentUser) return;` to avoid expelling a participant whose reconnect WS was already open but whose `joinSession` hadn't run yet.

But `getSessionParticipants` intentionally retains a participant with `connectionState === 'RECONNECTING'` AND `liveConnections.length === 0` so peers see the RECONNECTING badge during the grace window. Combined: the grace-timer found a never-reconnecting participant still in the user list (retained as RECONNECTING / zero-live) and bailed early. The entry survived until Redis TTL — **4h after A9 aligned connection TTL**.

### Fix
Query the authoritative live-connection count and only spare RECONNECTING entries that actually have an in-flight reconnect:

| State | Live conns | Result |
|---|---|---|
| `CONNECTED` | > 0 (implied) | spare |
| `RECONNECTING` | > 0 | spare (in-flight rejoin) |
| `RECONNECTING` | 0 | **evict (ghost)** |
| not in user list | n/a | fall through to evict |

New `distributedState.getParticipantLiveConnectionCount` wraps the already-internal `countLiveParticipantConnections` helper. Same Redis-failure conservatism as the surrounding code: if the count query fails, keep the participant alive and let the next grace cycle retry.

### Test
New regression test in `session-persistence.test.ts` (26 → 27 tests pass) asserts a disconnected-and-never-reconnects participant is gone from `getSessionMembers` after the 60s grace window. Confirmed it reproduced the bug against the pre-fix code and passes with the fix.

### Validation
- `vp check` ✓
- `vp run typecheck` ✓
- backend `session-persistence.test.ts`: 26/26 pass
- E2E shard 6 flake (`grid-mode-ascent-badge`) is the known top-4 seed-pinning issue documented separately (PR #2101's fix); unrelated to this PR.

PR #2150, opened 2026-05-14
Write board_climb_stats for user-created climbs

marcodejongh:
## Summary

Two compounding bugs made user-created climbs invisible in global search. Marco surfaced this on his profile (`Setty set`, `Ready… Tall boys version`, `Fossil Fuel`) — visible at `/profile/3fa79140-…` but missing from Kilter search.

### Bug 1: no `board_climb_stats` row

`saveClimb` for Aurora boards (and `saveMoonBoardClimb` without a grade) never wrote a row into `board_climb_stats`. The search hot path uses an `INNER JOIN` against that table for its index-driven `ascensionist_count DESC` plan (see `search-climbs.ts:152-208`), so those climbs were excluded at the join. They still showed on the setter's profile (LEFT JOIN), which is why the bug stayed silent. 1,852 published climbs across 104 users since 2020-09 were affected.

Fix: write a minimal stats row at climb-create time in `saveClimb`, on draft→publish in `updateClimb`, and on angle changes. Backfill the historical gap in migration 0096.

### Bug 2: silent regex escape in `populateDenormalizedColumns`

While QA-testing bug 1, "Nailssss" was published and still didn't appear in search by name. Investigation found a second, older bug: `populateDenormalizedColumns` extracts hold IDs from the `frames` text using `regexp_matches(frames, 'p(\d+)r', 'g')`, but the regex was written as a JS string literal — and JS silently drops the backslash before unrecognized escapes, so PG received `'p(d+)r'` and matched zero holds.

The fallout: every user-created Kilter/Tension climb went through `saveClimb` with NULL `edge_left/right/bottom/top`, NULL `required_set_ids`, and NULL `compatible_size_ids`. The search filter requires `compatible_size_ids` to contain the user's selected size, so the climb was filtered out at the SQL level even when the name and stats row matched. This explains why Marco's three climbs have all denormalized columns NULL in production.

Fix: write the regex with a doubled `\\d` so PG sees the intended `\d`. Extended migration 0096 with three additional UPDATEs that backfill edges, required sets, and compatible sizes for affected historical climbs.

### Why search-side wasn't changed

The comment block in `search-climbs.ts:152-160` calls out that the index-driven plan needs `FROM board_climb_stats` first. Switching the INNER JOIN to LEFT JOIN would regress the performance work tracked in PR #1969. Fix belongs on the write side.

## Out of scope (followup — #2151)

Boardsesh ticks (`saveTick`) don't yet increment `ascensionist_count`, so new climbs all show 0 ascents in search until that's wired up. Filed as #2151 — "boardsesh should start generating its own climb stats" end to end.

## Test plan

- [x] `vp run typecheck:backend` and `vp run typecheck:db` pass.
- [x] `vp test --project backend` passes — `climb-mutations.test.ts` updated to assert the new stats insert.
- [x] Local dev DB: backfill SQL applied, recount of stats-less listed user climbs drops from 1,852 to 0.
- [x] Local dev DB: "Nailssss" QA climb now has populated `edge_left/right/bottom/top`, `required_set_ids = {27}`, `compatible_size_ids = {7,10,15,…}`.
- [ ] Reviewer: create a new Kilter climb at 35° on the dev server, publish, verify it appears in search by name and at the chosen size. Edit angle within the 24h window; verify it appears at the new angle.
- [ ] After deploy: spot-check Marco's three Kilter climbs in production search.

### DIFF
--- a/packages/backend/src/__tests__/climb-mutations.test.ts
+++ b/packages/backend/src/__tests__/climb-mutations.test.ts
@@ -264,6 +264,79 @@ describe('climb mutations', () => {
     expect(statsInsert?.values).not.toHaveProperty('difficultyAverage');
   });
 
+  it('skips the stats seed for draft MoonBoard climbs without a grade', async () => {
+    mockDb.execute.mockResolvedValueOnce([]).mockResolvedValueOnce([]);
+    mockDb.select.mockReturnValueOnce(
+      createMockChain([{ name: 'Bob', displayName: 'Bob Setter', image: null, avatarUrl: null }]),
+    );
+    mockDb.insert.mockImplementation((table: unknown) =>
+      createMockChain(undefined, (values) => insertCalls.push({ table, values })),
+    );
+
+    await climbMutations.saveMoonBoardClimb(
+      {},
+      {
+        input: {
+          boardType: 'moonboard',
+          layoutId: 3,
+          name: 'Draft No-grade MoonBoard',
+          description: '',
+          holds: { start: ['A1'], hand: ['B2'], finish: ['C3'] },
+          angle: 40,
+          isDraft: true,
+        },
+      },
+      makeCtx(),
+    );
+
+    // climb_holds rows are still inserted, but no board_climb_stats row should appear.
+    const statsInsert = insertCalls.find(
+      (call) => typeof call.values === 'object' && call.values !== null && 'ascensionistCount' in call.values,
+    );
+    expect(statsInsert).toBeUndefined();
+  });
+
+  it('seeds a stats row with the grade for draft MoonBoard climbs that supplied one', async () => {
+    mockDb.execute.mockResolvedValueOnce([]).mockResolvedValueOnce([]);
+    mockDb.select
+      .mockReturnValueOnce(createMockChain([{ name: 'Bob', displayName: 'Bob Setter', image: null, avatarUrl: null }]))
+      .mockReturnValueOnce(createMockChain([{ difficulty: 17 }]));
+    mockDb.insert.mockImplementation((table: unknown) =>
+      createMockChain(undefined, (values) => insertCalls.push({ table, values })),
+    );
+
+    await climbMutations.saveMoonBoardClimb(
+      {},
+      {
+        input: {
+          boardType: 'moonboard',
+          layoutId: 3,
+          name: 'Draft Graded MoonBoard',
+          description: '',
+          holds: { start: ['A1'], hand: ['B2'], finish: ['C3'] },
+          angle: 40,
+          isDraft: true,
+          userGrade: '6A+',
+          isBenchmark: false,
+        },
+      },
+      makeCtx(),
+    );
+
+    // We preserve the grade for draft MoonBoard climbs even though the search
+    // filter masks the draft. Otherwise the user's grade would be lost on
+    // publish (updateClimb has no userGrade source to reconstruct it from).
+    const statsInsert = insertCalls.find(
+      (call) => typeof call.values === 'object' && call.values !== null && 'displayDifficulty' in call.values,
+    );
+    expect(statsInsert?.values).toMatchObject({
+      boardType: 'moonboard',
+      angle: 40,
+      displayDifficulty: 17,
+      difficultyAverage: 17,
+    });
+  });
+
   it('seeds a stats row on draft → publish transition in updateClimb', async () => {
     mockDb.select
       .mockReturnValueOnce(
@@ -348,6 +421,10 @@ describe('climb mutations', () => {
       ),
     ).rejects.toThrow('Cannot publish climb without an angle');
 
+    // The throw must fire BEFORE the board_climbs update — otherwise the caller
+    // sees a 500 but the row is left in a half-published state (isDraft = false,
+    // publishedAt = now, no stats row), which is the bug this PR is fixing.
+    expect(mockDb.update).not.toHaveBeenCalled();
     expect(insertCalls).toHaveLength(0);
   });
 
--- a/packages/backend/src/__tests__/climb-mutations.test.ts
+++ b/packages/backend/src/__tests__/climb-mutations.test.ts
@@ -121,6 +121,37 @@ describe('climb mutations', () => {
     });
   });
 
+  it('skips the stats seed for draft Aurora climbs', async () => {
+    mockDb.select.mockReturnValueOnce(
+      createMockChain([{ name: 'Alice', displayName: 'Alice Setter', image: null, avatarUrl: null }]),
+    );
+    mockDb.insert.mockImplementation((table: unknown) =>
+      createMockChain(undefined, (values) => insertCalls.push({ table, values })),
+    );
+
+    await climbMutations.saveClimb(
+      {},
+      {
+        input: {
+          boardType: 'kilter',
+          layoutId: 1,
+          name: 'Draft Aurora Climb',
+          description: '',
+          isDraft: true,
+          frames: 'p1r43',
+          angle: 40,
+        },
+      },
+      makeCtx(),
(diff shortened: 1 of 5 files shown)


marcodejongh:
Thanks for the review. Addressed in 0a03647:

- ✅ **updateClimb tests** — added four cases: draft→publish stats seed, angle-change stats seed, no-op update (asserts no extra stats row), and existing case for first-publish. Test infra needed `set` added to the `createMockChain` allowlist so the Drizzle update chain resolves.
- ✅ **saveMoonBoardClimb no-grade path** — added a dedicated case asserting the else branch seeds a stats row with `ascensionistCount: 0` and no difficulty fields.
- ✅ **`faUsername` consistency** — `saveClimb` now populates `faUsername` on its stats seed (the setter is already loaded at that point). `updateClimb` still omits it intentionally — that path doesn't pre-load the user profile and adding a lookup just to set faUsername on the seed felt like over-reach.
- 🔍 **Stale stats rows on angle changes** — fair point. The reason it's not in this PR is that without ticks → ascent counter wiring (#2151), the "stale" row is identical to a fresh one (both have `ascensionist_count = 0`). Once #2151 lands, an old-angle row could carry non-zero data and the cleanup case becomes real. I'll capture this in #2151 so it gets handled as part of the broader stats-ownership work.

Also: while QAing this PR locally I uncovered a second, older bug — `populateDenormalizedColumns` was using `'p(\d+)r'` as a JS string literal, which silently became `'p(d+)r'` and matched zero hold IDs. Every user-created Kilter/Tension climb had NULL `edge_left/right/bottom/top`, `required_set_ids`, and `compatible_size_ids`, so the search size filter excluded them even after the stats row existed. Fixed in f7690c7 with `\\d` and three extra UPDATE steps in migration 0096. PR description updated with the full story.

marcodejongh:
Pushed [b07ce42](https://github.com/boardsesh/boardsesh/commit/b07ce42) addressing the review:

- **saveClimb / drafts:** fixed — wrapped the stats insert in `if (!validated.isDraft)`, matching migration 0096 Step 1's `is_draft = FALSE` filter and `updateClimb`'s `!nextIsDraft` gate. Added a test covering the draft skip path. (`saveMoonBoardClimb` left as-is — its stats seed is tied to `userGrade` attachment at create time; happy to gate it too if you want symmetry.)
- **updateClimb `faUsername`:** fixed — added `setterUsername` to the existing-row select and passed it as `faUsername` in the stats insert. The column is nullable but having it correctly populated keeps the stats row consistent with `saveClimb`'s write.
- **`populate-denormalized-columns.ts` Step 2 NULL overwrite:** I think this one's actually safe as written and the guard would be harmful. The `UPDATE board_climbs c SET … FROM (subquery) WHERE c.uuid = sub.uuid` pattern only fires when the subquery emits a row for that uuid. If the regex matches zero holds or the placements join filters everything out, `CROSS JOIN LATERAL` produces zero rows, the `GROUP BY` emits no row for that climb, and the UPDATE is a no-op — the prior value is preserved. The migration's `AND c2.required_set_ids IS NULL` is a one-shot "only backfill what's still NULL" filter; copying it to the runtime path would prevent legitimate refresh-on-frames-edit (the whole reason `populateDenormalizedColumns` runs from `updateClimb`'s frames-changed branch). Let me know if I'm missing a case.
- **Stale stats rows at old angles:** agreed it's worth a follow-up. Want me to open the issue, or do you have one in mind to roll it into (e.g. the same family as #2151)?

---

marcodejongh:
Round 2 — pushed [f8792c8](https://github.com/boardsesh/boardsesh/commit/f8792c8):

- **Migration steps 2-4 / draft guard:** Updated the safety comment to accurately describe the asymmetric handling. Step 1 (stats) skips drafts deliberately. Steps 2-4 (denormalized columns) intentionally include drafts so we mirror `saveClimb`'s runtime behaviour — it calls `populateDenormalizedColumns` unconditionally on every climb, including drafts. The columns are unused while `is_draft = TRUE`, but pre-populating them means a future draft → publish that doesn't touch frames is search-ready (`updateClimb` only re-runs `populateDenormalizedColumns` on a frames change). Without this, drafts that survived the regex bug window would still ship to search with NULL `compatible_size_ids` after publish, defeating the migration.
- **Silent null-angle skip on publish:** Now throws `Cannot publish climb without an angle`. The condition is `!nextIsDraft && (transitioningToPublished || angleChanged)`; `angleChanged` always carries a non-null `validated.angle`, so the only way to land on a null `resolvedAngle` is a draft published without an angle ever being set. New test covers the throw.
- **Test masking the publish path:** Added an assertion that `mockPublishSocialEvent` ran with the expected payload (`type: 'climb.created'`, `entityId`, `setterDisplayName`). That call lives after the stats insert inside the `transitioningToPublished` block and depends on `getUserProfile`'s mock being consumed, so a clean assertion proves the second `select` mock was actually read.

**Lint failure root cause:** `mockDb` was seeded as `{ select, insert, delete, transaction, execute }` — no `update`. The three test cases reassigning `mockDb.update = vi.fn()` were tripping TS2339 under the type-aware lint job. Added `update: vi.fn()` to the seed; lint goes green without touching the call sites.

**E2E shard 6 (`grid-mode-ascent-badge`):** Looks like flake — this PR is backend-only (resolvers / migration / regex fix), no frontend or search-query changes. The test's `?showOnlyCompleted=true` filter operates over the test user's seeded ticks, which point at Aurora-synced climbs (always had stats rows), so the new migration shouldn't change what's visible. Re-running CI; will re-investigate if it fails twice.

---

marcodejongh:
Round 3 — pushed [1024b1d](https://github.com/boardsesh/boardsesh/commit/1024b1d):

- **High (DB mutated before angle guard):** real bug, exactly as described. The original ordering was *update* → *populateDenormalizedColumns* → *resolvedAngle check* → *throw* → *stats insert*. A null-angle publish would flip `isDraft = false` / `publishedAt = now()` and then throw, leaving the climb in the half-published state this PR is fixing. Hoisted `resolvedAngle` / `angleChanged` / `shouldSeedStats` and the throw above the `db.update` call, so a malformed-publish attempt rejects cleanly with `board_climbs` untouched. New test asserts `mockDb.update` was *not* called when the throw fires.

- **Medium (MoonBoard drafts):** half-fixed. Gated the no-grade branch the same way as `saveClimb`. Kept the grade-having branch seeding for drafts, because `board_climb_stats` is the only place we persist the resolved `displayDifficulty` — skipping it loses the user's grade through draft → publish (updateClimb has no `userGrade` source to reconstruct from). The "would be discoverable via global search immediately after save" concern doesn't materialize: `create-climb-filters.ts` `baseConditions` filters `is_draft = false` on every search, so a stats row on a draft is masked at the climb level. Happy to apply the gate to the grade branch too if you'd rather lose grades than carry a benign extra row — let me know.

- **Low (missing tests):** added two MoonBoard draft tests (no-grade → no stats insert; with-grade → stats insert preserved with grade) and the `expect(mockDb.update).not.toHaveBeenCalled()` assertion on the null-angle test (12 → 14 cases).

- **Info (onConflictDoUpdate is dead):** switched to `onConflictDoNothing`. UUID is freshly generated per call so the conflict resolution was unreachable; consistent with the other two paths now.

- **Info (Step 4 may leave some climbs unsearchable):** acknowledged, leaving the migration unchanged. Climbs whose bounding box exceeds every `board_product_sizes` row will keep `compatible_size_ids = NULL` and stay filtered out — a real gap, but pre-existing and out of scope for this fix. Worth a separate follow-up; will tag this on the merge note. The "1,852 → 0" wording in the PR description is best-effort; if you'd like I can adjust it to "1,852 affected climbs, *most* now searchable" so the language doesn't oversell.

CI re-running on the new commit.

---

PR #2171, opened 2026-05-15
Feed Boardsesh ticks into board_climb_stats (closes #2151)

### LINKED ISSUE #2151: Boardsesh-native climb stats: increment ascensionist_count from saveTick
## Context

PR #2150 makes user-created climbs discoverable in search by seeding a `board_climb_stats` row at climb-create time with `ascensionist_count = 0`. That fixes the "my climb doesn't appear in search" bug, but the climbs still always show **0 ascents** because nothing on the Boardsesh side ever increments the count.

Today, every non-zero `ascensionist_count` in the database comes from the Aurora sync pipeline (Kilter/Tension/MoonBoard official ascent data). Boardsesh's own ticks live in `boardsesh_ticks` and never feed back into `board_climb_stats`. So:

- A user-created climb with 50 Boardsesh ticks still ranks at the bottom of `ascents DESC`.
- A community-driven grade or quality average never forms — the difficulty/quality fields stay NULL on Boardsesh-originated climbs forever.

## Goal

Boardsesh should generate its own climb stats end-to-end, independent of Aurora. At minimum:

1. **Ascensionist count.** When `saveTick` (`packages/backend/src/graphql/resolvers/ticks/mutations.ts:277`) records a tick on a Boardsesh-originated climb, increment `board_climb_stats.ascensionist_count` at the relevant `(board_type, climb_uuid, angle)`. Decrement on tick deletion. Watch for double-counting on duplicate ticks (the existing dedup logic in `saveTick` handles this).
2. **Difficulty average.** When a user submits a grade vote, recompute `display_difficulty` / `difficulty_average` from the aggregate of votes. (Schema for grade votes may need to be added if it doesn't exist yet — needs investigation.)
3. **Quality average.** Same story for star ratings if/when we expose user stars.
4. **Don't double-count Aurora climbs.** For Aurora-synced climbs, the official Aurora numbers should remain authoritative — Boardsesh ticks for Aurora climbs are already reflected in Aurora's own counts via the sync-back path, so we'd be doubling. Need a clear policy on how to merge or partition.

## Open questions

- Do we trust Boardsesh ticks for difficulty/quality on Aurora-synced climbs at all, or only for user-created climbs?
- Are difficulty votes already captured anywhere? (Quick grep of the schema suggests no separate `board_climb_grade_votes` table on Boardsesh's side — it's all derived from the Aurora `ascents` payload today.)
- Performance: every tick currently inserts one row. Adding an `UPDATE board_climb_stats` on the hot path needs to play nicely with the covering index — verify the autovacuum tuning from migration 0086 still holds.

## Why this matters

Without this, search ranking of user-created climbs is permanently broken — every user-submitted climb sorts to the bottom of the default "ascents-desc" view even if it's been ticked dozens of times. The fix in #2150 makes them visible; this issue makes them rank.

## Related

- PR #2150 — the discoverability fix that surfaces this
- `packages/backend/src/graphql/resolvers/ticks/mutations.ts:277` — `saveTick` is the natural hook point
- `packages/db/src/queries/climbs/search-climbs.ts:152-208` — explains why search depends on `ascensionist_count` being accurate

marcodejongh:
## Summary

Closes #2151. Follow-up to the merged PR that ensured every climb has a `board_climb_stats` row.

Today every value in `board_climb_stats` comes from the Aurora sync. The merged PR ensured every climb has a stats row, but ranking was still broken: Boardsesh-created climbs always show 0 ascents because nothing updates `ascensionist_count` when ticks land.

This wires `saveTick` / `updateTick` / `deleteTick` to recompute the stats row from `boardsesh_ticks` via a 2-second debounced job.

### The Aurora-overwrite problem and the chosen split

Aurora sync's `upsertClimbStats` does `INSERT ... ON CONFLICT DO UPDATE` and blindly overwrites every column. A naive `ascensionist_count = ascensionist_count + 1` would be clobbered on the next sync.

Resolution — split the storage:

| Column | Owner |
|---|---|
| `aurora_ascensionist_count` | Aurora sync only |
| `boardsesh_ascensionist_count` | Boardsesh debounced recompute only |
| `ascensionist_count` | Materialized sum, kept in lockstep by both writers in the same statement that touches their share |

`ascensionist_count` stays a regular column (not `GENERATED`) so the custom covering index from migration 0067 (`board_climb_stats_ascents_covering_idx`, `DESC NULLS LAST` with `INCLUDE` columns) keeps working. Search reads it unchanged.

### FA ownership

For Aurora-synced climbs Aurora's FA is authoritative — the upsert writes `excluded.fa_username` / `excluded.fa_at` verbatim, including null (which is how Aurora signals a correction).

For Boardsesh-originated climbs (`board_climbs.user_id IS NOT NULL`), `recomputeClimbStats` re-derives `fa_username` / `fa_at` from the current ticks on every run — so a deleted FA tick correctly demotes to the next earliest sender, or to NULL if no senders remain. Aurora climbs go through a `COALESCE(s.fa_*, agg.first_*)` branch instead so Aurora's value isn't disturbed.

Boardsesh-originated climbs are never synced from Aurora (Aurora doesn't know their UUIDs), so the Aurora upsert path never touches them.

### Debounced publisher

2-second window keyed by `(boardType, climbUuid, angle)`. Mirrors `sessions/debounced-stats-publisher.ts` for local-instance dedup but skips the strict multi-instance dedup: the timer reads Redis for cleanup signal only, and falls through to the recompute on any nonce mismatch or Redis error. `recomputeClimbStats` is idempotent, so a duplicate run across instances is harmless; a silent drop (the prior bug) is not.

### What's in the PR

- `packages/db/src/schema/boards/unified.ts` — two new bigint columns on `boardClimbStats`
- `packages/db/drizzle/0099_split_ascensionist_count.sql` — adds the columns, seeds `aurora_ascensionist_count` from current values, and backfills `boardsesh_ascensionist_count` + FA for every climb with flash/send ticks (wrapped in explicit BEGIN/COMMIT)
- `packages/aurora-sync/src/sync/shared-sync.ts` — `upsertClimbStats` writes the aurora column and recomputes the sum on insert + conflict; FA fields write `excluded.*` verbatim
- `packages/backend/src/graphql/resolvers/ticks/recompute-climb-stats.ts` — single-statement recompute SQL inside a transaction with a defensive seed insert (explicit aurora=0, boardsesh=0) in case the stats row doesn't exist yet
- `packages/backend/src/graphql/resolvers/ticks/debounced-climb-stats-publisher.ts` — mirrors `sessions/debounced-stats-publisher.ts` (2s window + Redis best-effort cleanup)
- `packages/backend/src/graphql/resolvers/ticks/mutations.ts` — `queueClimbStatsRecompute` called from `saveTick`, `updateTick`, `deleteTick`
- `packages/backend/src/__tests__/debounced-climb-stats-publisher.test.ts` — 10 tests covering happy path, dedup window, fail-open on Redis errors
- `packages/backend/src/__tests__/recompute-climb-stats.test.ts` — 3 tests asserting the SQL invariants the delete-last-tick path depends on (`COALESCE(distinct_senders, 0)`, materialized-sum shape, sargable WHERE filter, ownership branch). End-to-end SQL behavior is also exercised by the migration backfill on 4,668 climbs of real production-shape data in the dev DB.

## Out of scope

- **Aggregating user grade / quality votes** into `display_difficulty`, `difficulty_average`, `quality_average`. `saveTick` accepts those fields but the recompute here only handles ascent count + FA. Same pattern can extend later.
- **`board_climb_stats_history` snapshots** for Boardsesh-driven changes — tracked in #2176.

## Test plan

- [x] `vp run typecheck:db` and `vp run typecheck:backend` pass locally.
- [x] `vp test --project backend` — new publisher (10) and recompute (3) tests pass. Pre-existing 22 integration-test failures in `integration.test.ts` are docker-daemon dependencies unrelated to this PR.
- [x] Local DB: migration 0099 applied; `aurora_ascensionist_count` seeded for all rows, `boardsesh_ascensionist_count` + FA backfilled for 4,668 rows with ticks.
- [ ] Reviewer: on dev server, create a climb → submit a `send` tick → wait 2s → confirm `ascensionist_count` is 1 with `fa_username` populated. Submit a second user's tick → expect count = 2. Delete the FA's tick → after debounce, FA reverts to the next earliest sender (or NULL if no senders remain).
- [ ] Reviewer: Aurora-overwrite scenario — pick a Kilter climb with high `aurora_ascensionist_count`, add a Boardsesh tick, confirm sum. Re-run the Aurora sync on that climb; confirm `boardsesh_ascensionist_count` survives and the sum is recomputed.

### DIFF
--- a/docs/ascents-and-attempts.md
+++ b/docs/ascents-and-attempts.md
@@ -163,6 +163,12 @@ There are exactly three production code paths that insert into `boardsesh_ticks`
 - Fires `assignInferredSession` (fire-and-forget) for non-party ticks.
 - Publishes `ascent.logged` event for the feed (only for `flash`/`send`, not `attempt`).
 - If `videoUrl` was attached, also inserts a `board_beta_links` row.
+- Calls `queueClimbStatsRecompute(boardType, climbUuid, angle)`. The debounced job
+  (2 s window, `packages/backend/src/graphql/resolvers/ticks/debounced-climb-stats-publisher.ts`)
+  recomputes `board_climb_stats.boardsesh_ascensionist_count` from the
+  current ticks for that `(board_type, climb_uuid, angle)` triplet — see
+  [`aurora-sync.md`](aurora-sync.md) for the two-writer model that the
+  recompute coordinates with.
 
 Two client hooks call it:
 
@@ -178,6 +184,17 @@ Two client hooks call it:
 `packages/aurora-sync/src/sync/user-sync.ts:214`. Bulk inserts attempts. Status is hardcoded to `'attempt'`. Quality and difficulty are always null.
 
 **No other code path may insert into `boardsesh_ticks`.** If you need to bulk-create ticks (a test, an import script), route through the same code or write a script in `packages/aurora-sync` that mirrors this shape.
+
+### Stats recompute on edits and deletes
+
+`updateTick` and `deleteTick` (same `mutations.ts` file) don't insert new
+ticks but they do change the answer to "how many distinct users sent this
+climb at this angle, and who was first." Both call
+`queueClimbStatsRecompute` after their write commits, so a deleted ascent
+correctly demotes `board_climb_stats.fa_*` to the next earliest sender (or
+to `NULL` if no senders remain) and a status edit from `attempt` → `send`
+bumps the count. If you add a new mutation that touches ticks, queue a
+recompute too.
 
 ---
 
--- a/docs/aurora-sync.md
+++ b/docs/aurora-sync.md
@@ -89,6 +89,32 @@ After every successful per-user sync, the daemon also runs a shared sync for tha
 
 When the climbs upsert sees previously-unseen UUIDs, the daemon also writes `new_climbs_synced` rows into the `notifications` table for each follower of the climb's setter (`setter_follows` and any linked `user_follows` accounts).
 
+### `board_climb_stats`: two-writer model
+
+`ascensionist_count` is the materialized sum of two source columns, each
+owned by a single writer:
+
+| Column                         | Owner                                       | Updated by                                                                                                |
+| ------------------------------ | ------------------------------------------- | --------------------------------------------------------------------------------------------------------- |
+| `aurora_ascensionist_count`    | Aurora sync                                  | `upsertClimbStats` (this file's daemon) — written verbatim from Aurora's payload                          |
+| `boardsesh_ascensionist_count` | Boardsesh `recomputeClimbStats`              | `packages/backend/src/graphql/resolvers/ticks/recompute-climb-stats.ts` — `COUNT(DISTINCT user_id)` over flash/send ticks |
+| `ascensionist_count`           | Both writers, kept in lockstep              | Every `upsertClimbStats` and every `recomputeClimbStats` recompute it as `COALESCE(aurora,0) + COALESCE(boardsesh,0)` |
+
+The search hot path reads `ascensionist_count` through the covering index from
+migration 0067, so it stays a regular column (not `GENERATED`) — both writers
+must update it whenever they touch their own share.
+
+`fa_username` / `fa_at` follow a related but asymmetric rule. Aurora's upsert
+writes them verbatim (including `null`, which is how Aurora signals an FA
+correction). `recomputeClimbStats` only re-derives FA for
+Boardsesh-originated climbs (`board_climbs.user_id IS NOT NULL`); on Aurora
+climbs it does `COALESCE(existing, agg.first_user)` so Aurora's authority is
+never disturbed. Boardsesh-created climbs aren't synced from Aurora, so the
+two paths can't collide.
+
+If you add a new writer to `board_climb_stats`, decide which side it owns and
+recompute `ascensionist_count` in the same statement that updates that side.
+
 ## CLI Usage
 
 ### Installation
--- a/packages/aurora-sync/src/sync/shared-sync.ts
+++ b/packages/aurora-sync/src/sync/shared-sync.ts
@@ -416,9 +416,13 @@ async function upsertClimbStats(db: DrizzleDb, board: AuroraBoardName, data: Cli
     // Two cooperating writers feed this row: Aurora sync (here) and the
     // Boardsesh tick recompute (recomputeClimbStats). Each owns its own count
     // column; ascensionist_count is the materialized sum kept in lockstep.
-    // Same idea protects fa_username / fa_at — Aurora COALESCEs onto an
-    // existing FA so a Boardsesh-side FA on a user-created climb survives
-    // subsequent syncs.
+    //
+    // FA fields are written verbatim from Aurora's payload — including null,
+    // which is how Aurora signals a correction (revoked / re-attributed FA).
+    // Boardsesh-created climbs are never synced through this path, so a
+    // Boardsesh-supplied FA on those climbs cannot be clobbered here.
+    // recomputeClimbStats is the one that handles the ownership branch
+    // explicitly for ticks-driven updates.
     const values = batch.map((item) => {
       const auroraCount = Number(item.ascensionist_count);
       return {
@@ -448,14 +452,8 @@ async function upsertClimbStats(db: DrizzleDb, board: AuroraBoardName, data: Cli
           ascensionistCount: sql`COALESCE(excluded.aurora_ascensionist_count, 0) + COALESCE(${climbStatsSchema.boardseshAscensionistCount}, 0)`,
           difficultyAverage: sql`excluded.difficulty_average`,
           qualityAverage: sql`excluded.quality_average`,
-          // Aurora is authoritative for FA on Aurora-synced climbs. Take the
-          // incoming value when it's set, otherwise preserve whatever's there
-          // (which protects a Boardsesh-supplied FA on the rare case where
-          // Aurora hasn't filled one in yet). Boardsesh climbs never go
-          // through this path — they're not synced — so we don't need to
-          // distinguish ownership here.
-          faUsername: sql`COALESCE(excluded.fa_username, ${climbStatsSchema.faUsername})`,
-          faAt: sql`COALESCE(excluded.fa_at, ${climbStatsSchema.faAt})`,
+          faUsername: sql`excluded.fa_username`,
+          faAt: sql`excluded.fa_at`,
         },
       });
 
--- a/packages/aurora-sync/src/sync/shared-sync.ts
+++ b/packages/aurora-sync/src/sync/shared-sync.ts
@@ -413,18 +413,28 @@ async function upsertClimbStats(db: DrizzleDb, board: AuroraBoardName, data: Cli
   const climbStatHistorySchema = UNIFIED_TABLES.climbStatsHistory;
 
   await processBatches(data, async (batch) => {
-    const values = batch.map((item) => ({
-      boardType: board,
-      climbUuid: item.climb_uuid,
-      angle: Number(item.angle),
-      displayDifficulty: Number(item.display_difficulty || item.difficulty_average),
-      benchmarkDifficulty: item.benchmark_difficulty != null ? Number(item.benchmark_difficulty) : null,
-      ascensionistCount: Number(item.ascensionist_count),
-      difficultyAverage: Number(item.difficulty_average),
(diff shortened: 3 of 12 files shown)


marcodejongh:
Thanks — addressed all four points in 89b8301:

- **Recompute FA from current ticks for Boardsesh climbs.** The previous `COALESCE(s.fa_*, agg.first_*)` pinned FA forever on the first sender. Now `recomputeClimbStats` checks `board_climbs.user_id IS NOT NULL`: for Boardsesh-originated climbs the FA is re-derived from `agg` every pass (so a deleted/downgraded FA tick correctly demotes to the next earliest sender, or to NULL if no senders remain); for Aurora-synced climbs Aurora's FA stays authoritative (only filled from ticks when currently NULL). Same ownership-aware logic applied to the migration backfill SQL.

- **Transaction wrap.** The defensive INSERT and the recompute UPDATE are now wrapped in `db.transaction()` so the row never appears in the half-state where `ascensionist_count = 0` is visible to a concurrent reader.

- **Winston logger.** Switched the three `console.error` calls in the debounced publisher to `logger.error`, matching `sessions/debounced-stats-publisher.ts` on current main. Test updated to spy on `logger.error` instead of `console.error`.

- **Drizzle-kit generated migration.** Removed the hand-authored 0098 and re-ran `bunx drizzle-kit generate`. The new migration is `0099_split_ascensionist_count.sql` with a real generated `when` timestamp; the data backfill is appended after the generated `ADD COLUMN`s. Snapshot file came along.

Rebased on latest `main` while I was in there to pick up the journal monotonicity fixes from #2163. Verified the new migration applies cleanly against the local dev DB.

marcodejongh:
Round 2 — addressed in f11ee7c:

1. **Aurora FA update bug** — flipped to \`COALESCE(excluded.fa_*, s.fa_*)\` in \`upsertClimbStats\`. Aurora's incoming value now wins unless NULL, so corrections propagate. Boardsesh climbs never go through this path (they're not synced to Aurora), so no ownership branch needed here. Added a comment explaining that.

2. **agg CTE sargable** — pushed \`status IN ('flash','send')\` from the \`FILTER\` clauses up into the outer \`WHERE\` in both \`recomputeClimbStats\` and the migration backfill. The \`FILTER\` ran after the full scan; with the predicate on the indexed WHERE attempt rows are pruned before aggregation. (You're right that the migration's pattern was the same shape as the recompute — both got the fix.)

3. **History snapshot understatement** — agreed it's out of scope here, but it deserves more than a buried PR-description bullet. Filed #2176 with three concrete options (read-back, schema mirror, server-side SELECT INTO).

4. **Journal gap at idx 98** — renumbered the migration from 0099 → 0098, renamed the snapshot file, updated the journal entry. Sequence is monotonic 96 → 97 → 98 now. Verified the rename applies cleanly to the local dev DB.

All 9 publisher tests still pass; backend typecheck green.

marcodejongh:
Round 3 — addressed in 02443e8 (rebased on latest main; my migration is now 0099):

1. **Redis SET silent failure** — added a \`setFailed\` flag to the closure. The \`.catch()\` flips it; the timer skips the Redis nonce check when the flag is set and runs the recompute directly. Without this, a SET error followed by a still-connected GET would see no nonce match and drop the recompute silently. Added a dedicated test \`runs the recompute anyway when Redis SET fails (fail-open)\` (10 publisher tests now pass).

2. **Aurora FA cannot clear via null** — reverted to plain \`excluded.fa_username\` / \`excluded.fa_at\`. Aurora's null is now propagated, so corrections (revoked / re-attributed FA) land. Boardsesh-created climbs are never synced through this path, so a Boardsesh-supplied FA on a Boardsesh climb still can't be clobbered here. Updated the comment to spell out the asymmetry: Aurora sync is verbatim; \`recomputeClimbStats\` is where the ownership branch (\`board_climbs.user_id IS NOT NULL\`) lives.

3. **Migration backfill not transactional** — wrapped the seed UPDATE and the ticks backfill UPDATE in an explicit \`BEGIN; ... COMMIT;\` block. Drizzle's migrator already transactions each migration, so this is belt-and-suspenders, but it removes any doubt for a reader (and for a future migrator that doesn't).

Rebase notes: had to take \`origin/main\`'s \`0098_mute_talos.sql\` snapshot during the rebase, regenerate the snapshot via \`bunx drizzle-kit generate\` (which produced the right \`0099_snapshot.json\` chain from \`0098\`), and clean up the auto-named \`0100_fresh_jack_power\` artefact drizzle-kit generated when it noticed my hand-authored 0099 had no snapshot. Final state: 96 → 97 → 98 → 99, monotonic.

marcodejongh:
Round 4 — addressed in f6fc014:

1. **setFailed race** — agree. Took the simpler path you suggested: dropped the flag entirely, the timer's Redis check is now best-effort (DEL the key if we still own it), and any mismatch / GET error / SET error falls through to the recompute. recomputeClimbStats is idempotent, so a duplicate run across instances on a rare contended burst is harmless; a silent drop is not. The per-instance local debounce in the \`pending\` Map still collapses bursts on the same instance; only the strict multi-instance dedup is gone. Tests updated (10 still pass): the \"nonce mismatch\" case now asserts recompute IS called and DEL is NOT called.

2. **PR description FA mismatch** — fixed. The current code is \`excluded.fa_username\` / \`excluded.fa_at\` verbatim on the Aurora path (so Aurora's null clears, preserving pre-PR behavior); the COALESCE branch lives only in \`recomputeClimbStats\` for Boardsesh-originated climbs. Description now spells out the asymmetry.

3. **Delete-last-tick test** — added \`packages/backend/src/__tests__/recompute-climb-stats.test.ts\` with three SQL-contract assertions:
   - \`boardsesh_ascensionist_count = COALESCE(agg.distinct_senders, 0)\` so the count collapses to 0 when the last flash/send tick disappears
   - the materialized-sum shape (\`COALESCE(s.aurora_ascensionist_count, 0) + COALESCE(agg.distinct_senders, 0)\`)
   - the sargable WHERE-side \`status IN ('flash','send')\` filter
   - the \`user_id IS NOT NULL\` ownership branch for FA

   These are structural; the end-to-end SQL behavior was also exercised by the migration backfill on 4,668 ticks-having climbs in the dev DB. A real-DB integration test that seeds/cleans up its own user + climb + ticks would be stronger; I held off because no other backend test in the repo does that pattern and I didn't want to set the precedent in this PR — happy to do it in a follow-up if you'd rather have it here.

4. **Defensive insert NULL aurora_count** — fixed. The defensive seed in \`recomputeClimbStats\` now sets \`auroraAscensionistCount: 0\` and \`boardseshAscensionistCount: 0\` explicitly. The harmless window you described is gone; downstream readers see a sensible baseline immediately.

marcodejongh:
Round 5 — addressed in cf38ab3:

1. **Docs updates** — both updated:
   - \`docs/ascents-and-attempts.md\`: added the \`queueClimbStatsRecompute\` side-effect under \`saveTick\`, and a new \`Stats recompute on edits and deletes\` subsection covering \`updateTick\` and \`deleteTick\` (which weren't in the original \"three insert paths\" enum but do trigger the recompute).
   - \`docs/aurora-sync.md\`: added a \`board_climb_stats: two-writer model\` subsection right after the shared table mapping. Spells out the aurora vs boardsesh column ownership, the materialized-sum invariant on \`ascensionist_count\`, and the asymmetric FA rule (Aurora-verbatim on the sync path, ownership-aware re-derive in \`recomputeClimbStats\`).

2. **Misleading BEGIN/COMMIT comment** — removed the dead block entirely. You're right: drizzle's migrator already wraps each migration in an outer transaction, and a nested \`BEGIN\` in PostgreSQL emits a NOTICE and is otherwise a no-op. The inner block bought nothing. The safety comment now says drizzle's outer transaction is the only mechanism and explicitly notes that we deliberately don't add a nested one.

3. **Fragile drizzle-AST assertion** — added a FRAGILE TEST WARNING header to the third recompute test. It explains:
   - what the test introspects (drizzle's internal \`queryChunks\`)
   - why we accept that fragility (no other backend test in the repo establishes the seed/cleanup pattern that a real-DB integration test would need)
   - that the end-to-end behavior IS covered via the migration backfill (~4,668 ticks-having climbs in the dev DB)
   - the right fix if drizzle ever shifts (write the real-DB integration test, don't chase the AST shape)

All 13 backend tests for the new files still pass.

Issue #1808, opened 2026-05-02
[Android] Screen collapses when keyboard activates


marcodejongh:
Most likely happens in all screens with a text input

![image](https://github.com/user-attachments/assets/fe0ec675-78db-4e5a-9f6e-e51dd907a93d)

PR #2199, opened 2026-05-17
fix(android): stop safe-area plugin keyboard padding from collapsing dvh layout

eantonya:
## Summary

- The \`@capacitor-community/safe-area\` plugin (v8.0.x) sets a physical bottom padding on the \`DecorView\` equal to \`imeInsets.bottom\` for Chrome >= 140
- Chrome treats this as a physical window resize, so \`interactiveWidget: 'resizes-visual'\` (set in the viewport meta) has no effect — \`window.innerHeight\` drops from ~986px to ~243px when the keyboard opens, collapsing every \`dvh\`-based layout
- Fix: override the \`DecorView\`'s \`OnApplyWindowInsetsListener\` in \`onCreate()\` after \`super.onCreate()\` (which loads plugins) to pass system bar insets through unchanged while zeroing physical padding — Chrome 140+ handles keyboard insets natively

## Context

Issue #1808 was correctly fixed by removing \`@capacitor/keyboard\` + \`adjustResize\`. That commit's explanation stated the safe-area plugin's DecorView padding would work correctly alongside \`interactiveWidget: 'resizes-visual'\`. In practice it does not: the physical padding causes Chrome to fire a window resize event (not a virtual keyboard event), bypassing \`interactiveWidget\` entirely. Tested on a Pixel 10 Pro XL with Chrome 148 — \`window.innerHeight\` 986px → 243px without this fix, ~986px → ~600px (correct overlay behavior) with it.

## Test plan

- [ ] Build debug APK: \`./gradlew assembleDebug\` from \`mobile/android/\`
- [ ] Install on Android device (Chrome 140+)
- [ ] Tap a text field — layout should stay full height, keyboard overlays
- [ ] Verify \`window.innerHeight\` stays ~986px via \`chrome://inspect/#devices\` Console (was 243px before this fix)
- [ ] Verify \`env(safe-area-inset-top/bottom)\` still reflects status bar / nav bar heights
- [ ] Verify bottom tab bar stays fixed at bottom when keyboard is open

### DIFF
--- a/mobile/android/app/src/main/java/com/boardsesh/app/MainActivity.java
+++ b/mobile/android/app/src/main/java/com/boardsesh/app/MainActivity.java
@@ -19,6 +19,9 @@ import android.webkit.WebView;
 
 import androidx.activity.EdgeToEdge;
 import androidx.annotation.NonNull;
+import androidx.core.graphics.Insets;
+import androidx.core.view.ViewCompat;
+import androidx.core.view.WindowInsetsCompat;
 
 import com.boardsesh.app.plugins.DevUrlPlugin;
 import com.getcapacitor.Bridge;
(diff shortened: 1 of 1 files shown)


eantonya:
@marcodejongh the keyboard bug is still there for me with latest build. This seems to fix it.

marcodejongh:
Awesome thanks so much!

PR #2215, opened 2026-05-18
Simplify queue-pivot internals: Phase 3 (structural refactors)

marcodejongh:
## Summary

Seven commits of structural simplifications surfaced during review of #2198 — each independently revertable. The pivot itself (queue-control bar lightbulb, drawer flow, 2 s wall-confirm fallback) is unchanged. This is a follow-up cleanup PR.

Rebased onto current `main` (post-#2198 merge); contains only the seven Phase 3 commits.

### What changes

| Commit | What |
|---|---|
| **3.5** Inline `executeWithLatestWins` | 40-LOC generic with one caller — inlined into `setCurrentClimb`. Same supersede semantics. |
| **3.2** Fold null-item path into `setCurrentClimbAndPublish` | The shared helper now accepts `item: ClimbQueueItem \| null`. Drops a 30-line copy of the retry+publish ladder in `queue/mutations.ts:setCurrentClimb`. |
| **3.1** Extract `buildSessionPayload` helper | Five resolvers (`joinSession`, `takeControl`, `releaseControl`, `confirmClimbOnWall`, `setSessionBoardSerial`) plus the `session` query were each building the 16-field GraphQL `Session` return verbatim with 4–6 sequential Redis fetches. Helper Promise.all's them all and accepts `inputs` overrides for fields the caller already knows. Also fixes drift: `queries.session` was silently dropping `name`. |
| **3.3** Drop driver/board-serial shadows when Redis is present | `localDriverBySession` / `localBoardSerialBySession` were written even on the distributed path, creating read-after-write skew windows. Now untouched when `distributedState` is set — Redis is the single source of truth. Maps stay as the no-Redis fallback. |
| **3.4** Split `play-view-drawer.tsx` | 1500 LOC orchestrator split into orchestrator + four siblings (`play-drawer-content.tsx`, `play-view-action-bar.tsx`, `play-view-tick-bar.tsx`, `wall-view-banner.tsx`). Two test imports updated. |
| **3.6** Retire `QueueDataContext` | Three consumers (`SetActiveAction`, `MirrorAction`, `CreateClimbForm`) migrated from `useOptionalQueueData` → `useOptionalCurrentClimb`. `dataValue` memo inlined into `contextValue`; `<QueueDataContext.Provider>` wrapper dropped from both `GraphQLQueueProvider` and `QueueBridgeProvider`. Combined `QueueContext` kept as the bridge's relay vehicle + test-only hook surface. |
| **3.6** (follow-up) | Fix four test mocks that still referenced the retired hook. |

Net: ~995 insertions / 1080 deletions across 25 files. Most of the insertions are the four new sibling files in 3.4; deletions are the orchestrator's inline definitions plus the wide-context machinery in 3.6.

### Validation

- `vp check` — 0 errors, 247 warnings (baseline; the one pre-existing error in `user-playlists.ts` is unrelated to this PR).
- `vp run typecheck` — clean across all packages.
- `vp test run` — 4514 / 4516 (only the pre-existing climbs.json + common.json catalog drift remains).

## Test plan

- [ ] **Solo flow:** tap a climb → drawer lightbulb appears → tap → BLE picker or auto-connect → wall lights up → drawer dismisses. (Verifies 3.4 split + 3.5 superseding semantics on rapid swipes.)
- [ ] **Party flow:** join from a second profile → swipe prev/next on the bar (both clients move) → only the driver's tap on the lightbulb lights the wall → `DriverChanged` flips peers' "X is driving" UI. (Verifies 3.1 helper + 3.6 fine-grained context migration.)
- [ ] **3.3 specific:** disconnect mid-session as the driver → peers see driver clear. In production this hits the Redis path; the shadow is dead code there.
- [ ] **3.5 specific:** rapidly swipe through 4–5 climbs in a row → only the latest lands on the wall; if `shouldAddToQueue` was true on a superseded call, the climb still ends up in the queue.
- [ ] **3.2 specific:** clear the current climb (server-side `setCurrentClimb({ item: null })` path) → other peers see `CurrentClimbChanged { item: null }`.
- [ ] Spot-check the four play-view test files (`play-view-action-bar`, `play-view-tick-bar-overlay`, `play-view-drawer-gate`, the new sibling imports).


packages/backend/src/graphql/resolvers/sessions/helpers.ts
```diff
@@ -0,0 +1,95 @@
...
+ * objects in every Session-returning resolver. Pass `inputs.X` to short-
+ * circuit a specific lookup when the caller already has the value.
+ */
+export async function buildSessionPayload(
+  sessionId: string,
+  ctx: ConnectionContext,
+  inputs: SessionPayloadInputs = {},
+) {
+  const [users, queueState, sessionData, driverParticipantId, lastConnectedBoardSerial, leaderConnectionId] =
+    await Promise.all([
```
marcodejongh, line 78: This is hard to read, we should improve its readability with all the inline ifs swapping between a new promise and promise.resolve

packages/backend/src/services/queue-navigation.ts
```diff
@@ -51,13 +56,15 @@ export async function setCurrentClimbAndPublish(
...
+  if (item !== null) {
+    // Record the climb as the latest authoritative wall climb so the next
+    // `confirmClimbOnWall` accepts it even if the driver navigates on before
+    // the confirm arrives. Best-effort: don't fail the publish if the recency
+    // record breaks (Redis hiccup).
+    await roomManager.pushRecentClimb(sessionId, item.climb.uuid).catch(() => undefined);
+  }
 
-  if (addedToQueue) {
+  if (addedToQueue && item !== null) {
```
marcodejongh: Should those if be in the item !== null if? Or even better could we early return on item === null instead?

PR #2218, opened 2026-05-19
feat: optimistic ascent count + live stat refresh after recompute

marcodejongh:
## Summary

When a user ticks a climb, the displayed **sends count** now bumps optimistically in the UI the moment the mutation fires (when it's the user's first send/flash at this angle). A new `climbStatsUpdated` GraphQL subscription event from the backend then replaces the optimistic value with the canonical recomputed numbers ~2–3 seconds later. For Boardsesh-owned (draft) climbs the **quality + difficulty averages** also update via the live event.

For Aurora-synced climbs the recompute preserves quality/difficulty (existing behavior), so the WS event only carries the new ascent count for those. That trade-off is accepted; making Boardsesh own quality/difficulty for Aurora-synced climbs is a separate change (see plan history).

## What's new

**Backend**
- New GraphQL subscription `climbStatsUpdated(boardType, climbUuid, angle)` with narrow per-(climb, angle) channel keys, modelled on the existing `commentUpdates` resolver.
- New pubsub channel `boardsesh:climb-stats:*` mirroring the existing comment / new-climb patterns (publish + subscribe + Redis fan-out).
- After `recomputeClimbStats` finishes in the debounced publisher, the publisher reads back the canonical `board_climb_stats` row and publishes a `ClimbStatsEvent`.

**Client**
- `useEffectiveClimbStats(boardType, uuid, angle, base)` — reads the live cache + optimistic delta, falls back to props.
- `useSubscribeClimbStatsUpdates(boardType, uuid, angle)` — graphql-ws subscription bridge. Module-scoped ref-counted shared client multiplexes all visible-row subscriptions over a single anonymous WebSocket.
- Hooks read the QueryClient via `useContext(QueryClientContext)` rather than `useQueryClient` so they gracefully no-op when a test renders the component without a `QueryClientProvider`.
- `useSaveTick.onMutate` bumps the ascent delta when the tick is the user's first send/flash at this `(climbUuid, angle)`; `onError` rolls it back; `onSuccess` schedules a 5s safety clear in case the WS event never arrives.
- ClimbTitle / ClimbDetailHeader / AngleSelector (per-row) thread `boardType` + `uuid` + `angle` into the hooks; existing call sites spread `boardType` from `boardDetails.board_name` when the climb prop doesn't already carry it.

## Subscription lifecycle

- One graphql-ws subscription per visible `(climbUuid, angle)`. graphql-ws multiplexes all of them over the single shared WS connection.
- Subscriptions unmount automatically when virtualized list rows scroll off-screen. The last unsubscribe also disposes the shared client.
- Channel keys are narrow, so the server's fan-out per tick is bounded.

## Out of scope

- `useUpdateTick` / `useDeleteTick` optimistic stats (e.g. demoting a send → no longer counted).
- Optimistic quality/difficulty for non-drafts (would require changes to `recompute-climb-stats.ts` + Aurora `upsertClimbStats`).
- Cross-page subscription dedup (e.g. queue control bar + search list showing the same climb).
- A unit test for the publish-after-recompute path in `debounced-climb-stats-publisher.ts` — the test infra needs mocks across db, pubsub, and recompute; covered by manual QA below for v1.

## Test plan

- [ ] Tick a low-ascent Kilter climb (1–5 sends) as **send**. Sends count bumps +1 immediately; canonical value lands within ~3s.
- [ ] Tick a climb you've already sent at the same angle. Sends count does NOT bump.
- [ ] Tick a climb as **attempt**. Sends count does NOT move.
- [ ] On a draft climb, send-tick with quality + difficulty rating. Stars + grade in the detail header update from the WS event.
- [ ] Scroll a search list, tick a row, scroll the row off-screen and back on. The live value persists across remount (cached in React Query).
- [ ] Block the GraphQL mutation in DevTools. Optimistic bump appears, mutation fails, ascent count rolls back, logbook entry rolls back.
- [ ] Stop the backend container (WS unreachable). Tick a climb. Optimistic bump appears; ~5s later the safety net zeroes the delta back to base.
- [ ] Open Network → WS frames during a long scroll. Confirm a single climb-stats WS connection with `subscribe` / `complete` frames as rows mount/unmount (no leaks).


marcodejongh:
Thanks for the review. Addressed in ec4f10e77:

1. **Channel-key injection** — tightened `climbUuid` validation to `^[A-Za-z0-9-]{1,64}$` so a `:` can never reach the channel-key parser. The earlier length-only check was the gap you flagged.

2. **Subscription hook tests** — added `use-subscribe-climb-stats-updates.test.tsx` covering the shared-client ref-count (single client across N subscribers, disposed on the last unmount), URL-change teardown, and the three no-op paths (missing args, no `QueryClientProvider`, null WS URL).

3. **`asBoardName` duplication** — extracted to `app/lib/board-name.ts`; `climb-title` and `climb-detail-header` now import from there. Left `user-drawer.tsx`'s helper in place because it's flag-aware: it validates against the `MOONBOARD_ENABLED`-filtered list in `app/lib/board-data.ts`, not the full shared-schema list. Unifying them would change navigation behaviour when the flag is off.

4. **Safety-timer rollback** — went one step further than "unconditional decrement", which would push the count negative if the live event arrived and the cache then got cleared / gc'd. The timer now decrements only when `ascentDelta > 0`, handling the post-live, post-gc, and live-never-arrived cases without any of them driving the displayed count below the base.

5. **Quality decimal precision** — noted; happy to normalise to 1 decimal everywhere in a follow-up, but the existing `formatQuality` in `lib/format-climb-stats.ts` already handles this for the display layer, so the flicker only shows up if a consumer reads `effective.qualityAverage` raw. Not the case for any current call site.

6. **`console.error`** — kept. The client codebase doesn't have a structured logger today (`use-notification-subscription.ts` uses `console.error` too), and `docs/logging.md` covers the backend winston logger only. Worth a separate cleanup pass when we add client logging.

CI re-running now.

marcodejongh:
Too divergent from main (~1052 commits behind) to rebase mechanically. `useSaveTick` was extracted into the shared `@boardsesh/board-react` package (consumed by web + mobile with injected I/O), and the web-only `packages/web/app/hooks/use-save-tick.ts` this PR patches is deleted in main (modify/delete conflict); GraphQL tick operations also moved to `packages/shared/graphql`. Re-homing the optimistic-delta logic into the shared hook is feature rework, not conflict resolution. Intent captured in https://github.com/boardsesh/boardsesh/issues/2641 for a fresh implementation. Branch `feat/optimistic-tick-stats` preserved.

PR #2219, opened 2026-05-19
feat(board-history): per-board WS room with canonical history + opt-in shared queue

marcodejongh:
## Summary

- Introduces a per-board WebSocket room with a canonical chronological history (new `board_climb_history` table). Anyone paired to the board sees a live, persisted feed of what's been on the wall.
- Makes the existing shared-queue model opt-in via a new "Shared queue" toggle in the sesh-settings drawer. New sessions default to local-only queues; legacy live sessions (`ended_at IS NULL`) are backfilled to keep the shared queue on.
- Adds a read-only **"Currently on the board"** header at the top of the play-view drawer plus a divergence dot on the queue control bar, so the user always knows what's on the wall vs what they have queued. When the two diverge and the user is BLE-connected, a one-tap "Send your pick" CTA reclaims the wall.

## Why

User testing showed the shared-queue model is rarely used — most people just want a board, their own queue, and a sense of what's actually been on the wall (including by others). A per-board WS room is the simplest collaborative primitive that gives us that without forcing a shared queue.

The canonical "what was on the wall" log also unlocks several follow-ups: a "Recent on this board" feed (intentionally deferred — see TODO in `currently-on-board-header.tsx`), board-level analytics, and tighter tick → send provenance.

## Architecture

- **Room key:** BLE serial (hybrid — linked to `userBoards.id` when registered, works for visitor/unregistered boards too)
- **Access:** soft-gated on `user_board_serials` membership (anyone paired to the board can read+write)
- **Sequence allocation:** single-writer fence table (`board_history_sequences`) — atomic upsert inside the same transaction as the insert; Redis caches the counter + capped event tail
- **Idempotency:** client-generated `uuid` is the source of truth (matches `boardsesh_ticks.uuid` precedent)
- **Sessions:** existing party `boardSessions` table stays as-is; new `shared_playlist_enabled` flag gates the WS queue mutations server-side (throws `SHARED_PLAYLIST_DISABLED` when off)

## Implementation phases (one commit covers all)

1. **DB:** `board_climb_history` + `board_history_sequences` tables, `board_sessions.shared_playlist_enabled` column, migrations `0101_ordinary_major_mapleleaf.sql` + `0102_backfill_shared_playlist_enabled.sql`.
2. **Backend:** new `BoardRoom` in `room-manager`, resolvers for `recordBoardSend` / `boardHistoryEvents` / `boardHistory` / `setSharedPlaylistEnabled`, gated existing queue mutations on the flag, pub/sub channel `board-history:<serial>`.
3. **FE data:** `BoardHistoryProvider` + hooks (`useBoardHistory`, `useBoardVsLocalDivergence`, `useSendLocalPick`, `useRecordBoardSend`, `useSetSessionSharedPlaylist`), BLE auto-send wired to fire `recordBoardSend`, `activeSession.sharedPlaylistEnabled` propagated, `isParty` gated on the flag.
4. **FE UI:** `CurrentlyOnBoardHeader` (sticky in the drawer), divergence dot on the queue control bar, `SharedPlaylistToggle` + confirm dialog in the sesh-settings drawer. No tabs.
5. **i18n + tests:** en/es/fr `session.json` keys; backend tests for idempotency + concurrent sequence allocation + auth + queue-mutation gating; web tests for divergence hook + header states + toggle interactions.

## Judgment calls worth a review eye

1. **Implicit `joinSession` defaults `sharedPlaylistEnabled: true`** while explicit `createSession` defaults to `false`. This preserves legacy party UX for clients that still reach a session via the implicit WS join handshake; the new UI uses explicit creates. Once the new flow is the only entry point, we can flip implicit creates too.
2. **New slim `BoardSession` GraphQL type** for the toggle mutation return shape — avoids forcing queue recomputation just to return the flag.
3. **`boardId` left null at the client call site for `recordBoardSend`** — the backend resolves it from `userBoardSerials` server-side. Avoids an extra round-trip from route → numeric ID on the client.
4. **Manual-sender registry pattern in `bluetooth-status-store`** — the play-view drawer's "Send your pick" CTA is mounted above the per-route `BluetoothProvider`, so a context-based call would throw. The module-level store mirrors the existing `setActiveBluetoothSerial` pattern in the same file.
5. **No drawer tabs.** The board's full chronological history is intentionally not surfaced as a swipeable list here — only the latest entry in the header. A future "Recent on this board" surface (separate route) is the right home; TODO marker is in `currently-on-board-header.tsx`.

## Test plan

- [ ] Pair BLE, send a climb → drawer header shows "You sent this · just now"; reload → header persists (PG read)
- [ ] **Handoff scenario:** User A sends climb A; User A disconnects; User B pairs and sends climb B; User A's header updates to "Sent by [B]"; divergence dot lights up on queue control bar; User A re-pairs; taps "Send your pick"; header updates to A; dot clears
- [ ] Soft pairing gate: a user with no `userBoardSerials` row for a serial cannot subscribe to `boardHistoryEvents` for that serial
- [ ] Toggle shared-queue OFF mid-session: confirm dialog warns local queues don't merge; queue mutations from peers return `SHARED_PLAYLIST_DISABLED`; clients fall back to local IDB queue; BLE sends still feed the board history
- [ ] Toggle back ON: confirm dialog; shared queue resumes; local queues stay distinct
- [ ] Legacy session (created before this PR) keeps working — backfilled to `shared_playlist_enabled = true`
- [ ] Empty states: no BLE + no `userBoards` route match → "Pair a board" header; paired but no history → "Nothing's been on the wall yet"
- [ ] No regression on solo mode (no session): queue behaves exactly as before; header just shows the latest send if BLE-paired

## DB verification

```sql
SELECT board_serial, climb_uuid, user_id, sequence, sent_at
  FROM board_climb_history
 ORDER BY sent_at DESC LIMIT 20;

SELECT * FROM board_history_sequences;

SELECT id, name, status, ended_at, shared_playlist_enabled
  FROM board_sessions ORDER BY created_at DESC LIMIT 20;
```


marcodejongh:
Rebasing onto current `main` (1039 commits ahead of the May-20 branch point) is too divergent to do mechanically: the `0101`/`0102` migrations collide with main's `0118` baseline (renumber + `drizzle-kit generate` regen required, which repo rules say must not be hand-edited), and the BLE control surface this PR wires board-history sends into (`bluetooth-context.tsx`, `use-board-bluetooth.ts`) was substantially rewritten on main — 11 of 13 conflicts are overlapping feature logic, not import moves.

Intent captured in https://github.com/boardsesh/boardsesh/issues/2642 for a fresh implementation. Branch `board_sessions` is preserved.

PR #2240, opened 2026-05-22
design(wall-view): replace banner with presence row + bottom escape link

marcodejongh:
## Summary

The wall-view header strip added in #2238 (lock icon, driver avatar, hint sentence, "Browse from here →" button) does its job but reads as a permissions warning bolted onto a sheet. This PR redesigns the same surface with a quieter shape:

- **Thin presence row** above the climb header: `(◉) {driver} · driving` — secondary color, no background tint, no lock icon, no border. Tells you who's driving without shouting.
- **Quiet inline link** at the bottom of the drawer body: `browse without moving the wall ⌄`. Only rendered when the local user is a non-driver; tap to drop out of wall-view without closing the drawer.
- **One-shot teach moment** moves from inline chrome to a snackbar. First time a non-driver lands in wall-view, fire `showMessage('Locked to the wall climb. Tap browse below to look around.')` and write `swipeHint:wallViewSeen` to IndexedDB. Subsequent opens are silent.

Drops the `showWallViewHint` state, the `handleWallViewHintSeen` callback, the `wallViewHintSeenLatchRef`, and two now-banner-only icon imports (`LockOutlined`, `ArrowForwardOutlined`). Net: −154 lines, +66 lines in `play-view-drawer.tsx`.

i18n reworded across en-US, es, fr to match the new shape — short label-style chip for the presence row, sentence hint that references the new bottom link, lowercase verb for the link.

## Design context

Four designer voices were sourced (minimalist / iOS HIG / social-presence / IA pragmatist). Minimalist won — see `.claude/plans/mossy-prancing-fog.md` if curious. The IA pragmatist's "delete the mode entirely, put a persistent ON WALL strip on the bar" idea is parked as a possible follow-up after living with this version.

## What stays the same

- `wallView` / `onExitWallView` props on `PlayViewDrawer` — the mode still exists, only its skin changed.
- `SwipeBoardCarousel`'s swipe-disable plumbing (`enabled: !wallView || isDriver`).
- `PlayViewActionBar`'s `wallView` prop that hides prev/next.
- Driver-side rendering: drivers don't see the presence row (no "you're driving" reminder) and don't see the bottom escape link (they have full nav).

## Trade-offs

- A non-driver who opens the drawer and swipes hits silence before they find the bottom escape link. The snackbar teaches the mode but only on the very first open per user. The bottom link is for every session after that.
- Bottom escape has lower discoverability than the top button. Acceptable: most passengers want to watch what's lit, not browse around it.

## Test plan

- [ ] `vp check` and `vp run typecheck:web` clean ✅ (verified locally)
- [ ] `vp test run --project web` clean ✅ (4670 tests pass locally)
- [ ] Two devices in a party session at 35°:
  - Tap bar body as non-driver → presence row `(◉) {driver} · driving` renders, snackbar fires once, no banner.
  - Reopen drawer → no snackbar (one-shot).
  - Swipe attempt as non-driver → silent, no movement, bottom link visible.
  - Tap `browse without moving the wall ⌄` → drawer stays open on same climb, swipe re-enables, can preview locally.
  - Driver-side: tap bar → drawer opens with NO presence row, swipe works as before.
- [ ] `/es/...` and `/fr/...` routes — presence chip and link render in active locale.

## Note on base

Based on `fix/group-session-swipe-angle-logging` (#2238) since that's the branch the banner was introduced on. Merge there first if #2238 lands before this; otherwise this can be rebased onto main.


marcodejongh:
Superseded by #2243, which takes the IA-pragmatist direction from the original four-designer brainstorm: delete the wall-view mode entirely and put wall-climb identity on the persistent bar (with a drift pill in the drawer for return-to-wall). Plan: `.claude/plans/mossy-prancing-fog.md`.

PR #2277, opened 2026-05-24
Offline sync, end-to-end: Phase 2 backend + on-device SQLite + mobile wiring

marcodejongh:
## What this is

This branch began as client-side scaffolding for the mobile offline layer; four code reviews found it couldn't run — no on-device SQLite, no Phase 2 sync backend, nothing wired, the offline write path dormant. This PR builds the whole stack **and makes offline writes live end-to-end**. Cross-package contract: `docs/sync-table-manifest.md`.

## Offline writes are live (dual-write + 1-by-1 sync runner)
- Every user write dual-writes: `useSaveTick` / `useToggleFavorite` persist to local SQLite AND enqueue + fire the drainer, which pushes the queue **one at a time** (immediately when online); both fall back to the online mutation when the local DB is unavailable. No request batching — kept deliberately simple per review direction.
- The enqueued payload is the complete `SaveTickInput`, closing the `climbedAt` gap that would have dead-lettered every offline tick.
- Queue hardening: true FIFO (`peekPending` id tiebreak), atomic dead-letter (`recordFailure`), exponential backoff. Settings has a dead-letter section (view / retry / discard), and sign-out warns before discarding unsynced ascents.

## Phase 2 backend (was missing) — now prod-safe
- `addFavorite`/`removeFavorite` (idempotent); `saveTick`/`createPlaylist` accept a client `uuid` with `ON CONFLICT DO NOTHING` (return the original row, skip side effects on replay); 10 composite-cursor sync resolvers (snake_case JSON, collision-proof); `sync_deletions` + per-table triggers with natural-key `record_id` (incl. the BEFORE-DELETE owner capture + cascade NULL-guard).
- **Migrations are lock-safe.** `0108` adds `sync_seq`/`updated_at` to the 200k–1M-row board tables via nullable-add → batched `FOR UPDATE SKIP LOCKED` backfill → set default/not-null (no `ACCESS EXCLUSIVE` full-table rewrite on deploy; verified on a throwaway 648k-row table); `0109` no longer re-`UPDATE`s the board tables; `0110` adds `(user_col, updated_at, id)` cursor indexes on the 7 user sync tables.

## On-device SQLite foundation + wiring
- `src/db` connection / DDL / sequential migrations; natural-key local PKs per the manifest; `SQLiteProvider` whose init never rejects (a DB failure degrades to online-only, no white-screen).
- `OfflineSyncBridge` runs the drain-then-pull scheduler + notification deep-link routing; `signOut` clears local user data + checkpoints + pending mutations (board reference data kept), best-effort drain first.

## Optional seed DB
- The pre-warmed board-data asset is **optional**: `initializeDatabase` loads it if a build bundles one (copies board reference data into empty tables, stamps checkpoints) and runs empty otherwise. Resolved via an indirection module with **no literal `require`**, so default builds carry no seed reference and Metro bundles cleanly with or without it. Per-board sync is opt-in (Settings toggle + live progress + download-size note).

## Tests / validation
- typecheck (shared / db / backend / mobile / web), **mobile 560 tests**, iOS + Android bundle, i18n orphans + parity — all pass. Backend changed-surface tests pass (sync / favorites / ticks / playlists); the e2e/WS files fail only on port contention in this shared multi-worktree sandbox, untouched by these changes.
- `pull-client.integration.test.ts` runs backend-shaped docs through `pullSync` into a real `node:sqlite` DB built from the production DDL (a deliberately-wrong column fails it).

## Deferred (tracked follow-ups)
1. **Pre-warmed asset build pipeline + offline climb search.** The optional-asset *load path* ships here; the CI that *builds* the ~150-200MB asset, and repointing climb search at the local-SQLite JOIN, remain follow-ups — board sync is opt-in/off by default until then.
2. Push-token registration lifecycle; shared GraphQL-ops extraction (I16); deletion-log prune cron; sync-resolver rate limiting.

Base is `main`; the migrations are additive + lock-safe and the idempotency changes are backward-compatible (existing web callers omit `uuid`).


marcodejongh:
Too divergent from `main` to rebase mechanically. The branch is 618 commits behind / 11 ahead (merge-base e4bb60578, last commit 2026-05-31), and those 618 commits restructured exactly the surfaces this PR integrates against: `app/(tabs)/more/index.tsx` was deleted on main (modify/delete conflict on the first rebase commit), `app.config.ts` plugins diverged structurally, `theme-provider.tsx`/`PlayDrawer.tsx`/`LogbookSection.tsx`/`_layout.tsx`/`auth-provider.tsx` were heavily reworked, i18n moved to `packages/shared/i18n/locales/`, and migrations 0108–0110 collide with main 0111–0118. The offline-sync feature is still wanted and is NOT in main (no sync-deletions schema, sync resolvers, mutation-queue, or pull-client on origin/main) — intent captured in https://github.com/boardsesh/boardsesh/issues/2639 for a fresh implementation. Branch `mobile-phase5` preserved.

PR #2284, opened 2026-05-24
fix: break hash resync loop when offline buffer items are pending

marcodejongh:
## Summary

Fixes Sentry issue #7478404069: "Resync loop: client keeps disagreeing with server hash"

The 60-second hash watchdog was triggering infinite resyncs when the offline buffer had pending items. The FullSync handler merges offline-buffered items into the local queue for visual continuity, but the server hash doesn't include them — so the local hash never matches the server hash, and each resync just repeats the same cycle.

Three changes in `use-session-subscriptions.ts`:

- **Skip hash verification while offline buffer is non-empty** — the mismatch is expected and self-resolving once reconciliation pushes the items to the server and clears the buffer
- **Stop triggering resyncs after loop detection** (3+ consecutive mismatches for the same server hash) — repeating the same operation won't fix the underlying issue
- **Add `offlineBufferLength` to Sentry diagnostic data** — confirms whether offline items are involved in future events

## Test plan

- [ ] `bunx tsc --noEmit -p packages/web/tsconfig.json` passes (verified)
- [ ] Reconnect with offline-buffered items: no Sentry warning, no console resync loop messages
- [ ] Stable connection: 60s hash checks still pass (DEBUG log "State hash verification passed")
- [ ] Genuine hash mismatch: resync fires up to 3 times, then stops with Sentry warning including `offlineBufferLength: 0`

https://claude.ai/code/session_011gBq266GUnMy2pg8ufg4Vv

---
_Generated by [Claude Code](https://claude.ai/code/session_011gBq266GUnMy2pg8ufg4Vv)_

### DIFF
--- a/docs/websocket-implementation.md
+++ b/docs/websocket-implementation.md
@@ -946,19 +946,34 @@ sequenceDiagram
 
     Note over C: Every 60 seconds
 
-    C->>C: Compute local hash
-    C->>C: Compare to last server hash
+    C->>C: Check offline buffer
+    alt Offline items pending
+        C->>C: Skip verification (expected mismatch)
+    else No offline items
+        C->>C: Compute local hash
+        C->>C: Compare to last server hash
 
-    alt Hashes match
-        C->>C: State verified ✓
-    else Hashes differ
-        C->>C: Log "State drift detected"
-        C->>S: Trigger resync
-        S->>S: Re-join session
-        S->>C: FullSync event
-        C->>C: Apply full state
+        alt Hashes match
+            C->>C: State verified ✓
+            C->>C: Reset consecutive resync counter
+        else Hashes differ (attempt ≤ RESYNC_LOOP_THRESHOLD)
+            C->>C: Log "State drift detected"
+            C->>S: Trigger resync
+            S->>S: Re-join session
+            S->>C: FullSync event
+            C->>C: Apply full state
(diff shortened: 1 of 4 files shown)


marcodejongh:
Rebased onto main @af7e89bce (force-pushed 905035d2).

The resync-loop cap from this PR already landed in main via #2359 (the \`<= RESYNC_LOOP_THRESHOLD\` guard, consecutive-resync counter, and Sentry capture). The still-novel part — the **offline-buffer skip** for Sentry issue #7478404069 — is preserved here. Conflicts resolved in \`use-session-subscriptions.ts\`: kept main's reviewed \`<=\` loop-cap semantics, layered the offline-buffer guard + \`offlineBufferLength\` Sentry field on top. Net hook diff is now just 15 lines.

Rebase also surfaced a mechanical breakage: main's #2359 test (\`use-session-subscriptions.test.tsx\`) built its refs without \`offlineBufferRef\`, so the new \`offlineBufferRef.current.length\` read threw. Added an empty \`offlineBufferRef\` to that test.

Validation: \`vp run typecheck\` (40 tasks) pass; new \`session-subscriptions-hash-watchdog.test.ts\` 5/5 + legacy watchdog test green; \`vp run check:i18n\` pass; lint/format clean on the 4 changed files. (Pre-existing format drift in 19 unrelated files on main is untouched.) Still a draft.

PR #2287, opened 2026-05-24
feat(mobile): anonymous-first — browse, queue without an account

marcodejongh:
## Summary

Mirror the web app's anonymous behaviour on mobile. Almost everything works without an account — pick a board, browse climbs, view detail, add to queue, navigate the queue, drive the wall via BLE. **Only Favorite (`toggleFavorite`) and Log Ascent (`saveTick`) require sign-in**, gated client-side via a new `SignInPromptSheet`.

## What changed

**Backend (small, targeted):**
- `defaultBoard` returns `null` for unauth instead of throwing `requireAuthenticated`. Schema field is already nullable, fully backwards-compatible.
- `getBoard` / `boardBySlug` now filter to `isPublic && !isUnlisted` when unauth. Prevents anonymous callers from fetching private board metadata by UUID. Authed users are unaffected.

**Mobile (the meat):**
- `useEffectiveDefaultBoard` — new hook that composes `useDefaultBoard` (the server-side authed default) with a `SecureStore` fallback. Anonymous users pick a board from the Boards tab, it gets persisted locally, and every other tab reads the same source.
- **Boards tab** — drops the auth gate, branches data source. Authed: `useMyBoards` (their boards). Anonymous: `useSearchBoards({ limit: 20 })` (recent public boards).
- **Climbs tab** — drops the auth gate, swaps to `useEffectiveDefaultBoard`. Empty state gains a "Pick a board" CTA that navigates to Boards.
- **Climb detail** — drops nothing at the top level. Favorite and Log Ascent buttons gate on `isAuthenticated` and present `SignInPromptSheet` for unauth users. Add to Queue stays open — backend `createSession` accepts non-discoverable anonymous sessions today.
- **Queue tab** — drops the auth gate, uses `useEffectiveDefaultBoard`.
- **Profile tab** — keeps its `SignInPrompt` gate (inherently user-scoped).

**Pre-existing bugs fixed in passing (had to land for the feature to work):**
- `QueueProvider` was defined but **never wrapped** in `app/_layout.tsx`. `useQueue()` would throw the moment any authed user reached the queue tab or climb detail. `BluetoothProvider` depends on `QueueProvider` being its parent.
- `+not-found.tsx` redirected to `/(tabs)` (a route group, not a leaf), which re-resolved to `+not-found` and caused a render loop on cold launch as anonymous. Now redirects to `/(tabs)/boards`.
- `AuthProvider` previously used render-time `<Redirect href=\"/auth/login\">` for unauth users, which fought the navigator on cold launch. Now uses a deferred `useEffect` + `router.replace` for the authed-in-auth-stack case only — unauth users flow into the tab tree.

## i18n

7 new keys under `auth.nativeStart.prompt.*` (boardsPickerTitle/Description, climbsNoBoardCTA, favoriteTitle/Description, logAscentTitle/Description) plus the restored profileTitle/Description. All three locales (en-US, es, fr) — parity test passes. Mobile-only keys declared in `packages/web/scripts/mobile-i18n-keeps.ts` so the orphan check stays clean.

## Test plan

- [x] `vp check` — 0 errors
- [x] `vp run typecheck:backend` — clean
- [x] `vp run typecheck:mobile` — only pre-existing FlashList API drift and `@react-navigation` type errors from SDK 56 upgrade, none in files touched here
- [x] `vp run check:i18n` and `vp run check:i18n:orphans` — clean
- [x] Simulator: cold-launch as authed user, Boards tab shows the user's board, Queue tab no longer crashes (proves `QueueProvider` wrap), `useEffectiveDefaultBoard` returns the authed default
- [ ] Manual (interactive): sign out, cold-launch as anonymous. Boards shows recent public boards. Pick one. Climbs loads. Tap climb. Add to queue (lazy `createSession` fires). Queue tab shows the climb. Tap Log Ascent → `SignInPromptSheet` rises. Tap Favorite → same.
- [ ] Manual: sign in via Google. Boards switches back to your boards. Log Ascent opens the real `LogAscentSheet`. No regressions.

## Risks / known issues

- WebSocket `queueUpdates` subscription: needs a quick verification that the client doesn't reject when there's no JWT header. The backend `requireSessionMember` only needs the participantId. If the mobile WS client requires a Bearer header, this is a small follow-up.
- `signOut` clears `StoredBoardConfig`. Anonymous users never hit this. If an authed user signs out, their previously-picked board is wiped — they'll need to re-pick. Out of scope to change, but flag-worthy.
- Anonymous users get the **20 most recent public boards** instead of "popular". The `searchBoards` resolver has no popularity sort — backend follow-up if we want it.

## Follow-up tickets

- Native `expo-apple-authentication` (Apple HIG-preferred, sidesteps the form_post issue with the current web OAuth flow)
- Discover / marketing tab as the unauth default landing
- View other users' public profiles on mobile (mirror web's public-profile pages)
- Anonymous-user onboarding: a one-time "you're browsing as a guest" toast


marcodejongh:
This branch is 870 commits behind `main` (branched at 720af0970 / PR #2286) and the mobile app has been reorganized since: every touched file is gone, moved, or rewritten — `(tabs)/boards/index.tsx` → `app/boards/index.tsx` (modal), `(tabs)/queue/index.tsx` and `BlurTabBar.tsx` removed, `climbs/index.tsx` rewritten for search-first, and the QueueProvider/+not-found/AuthProvider fixes already landed via the merge-base PR #2286. The backend `boards.ts` resolvers were also restructured. Too divergent to rebase mechanically — it would be a full reimplementation against the new shell.

The feature is still wanted: `main` still hard-redirects unauthenticated users to `/auth/login` (`auth-provider.tsx:158-159`) and `defaultBoard` still calls `requireAuthenticated` (`boards.ts:1112`). Intent captured in #2638 for a fresh implementation. Branch `feat/mobile-anonymous-browse` preserved for revival.

PR #2350, opened 2026-05-29
fix(firmware): make the esp32 board discoverable and detect power-off quickly

marcodejongh:
## Problem

Testing the esp32 `board-controller` firmware against the iOS app surfaced two bugs (both reproduced on app 1.2 and 1.3, which pointed at the firmware rather than the app):

1. **The board is never discovered** — the device picker shows an infinite "finding boards nearby" spinner.
2. **Disconnect isn't detected** — after connecting (solo session, lightbulb lit), powering the esp32 off never turns the lightbulb off.

## Root causes (both in the firmware)

**1. Advertising packet overflow** — `nordic_uart_ble.cpp` `begin()` added **two** 128-bit service UUIDs to the advertisement (Aurora `4488b571…` + Nordic UART `6e400001…`). Two 128-bit UUIDs (36 B) plus the Flags structure blow past the **31-byte legacy advertising limit**, so NimBLE drops a UUID or fails to start advertising. The iOS/web/Capacitor scans filter on the **Aurora UUID only**, and a central only sees UUIDs present in the advertisement (GATT-only exposure is invisible to a filtered scan) — so the board never appears.

**2. No bounded supervision timeout** — the native→JS→UI disconnect chain is correct (verified: `CBConnectPeripheralOptionNotifyOnDisconnectionKey` is set, `didDisconnectPeripheral`→`onDisconnect`→JS guard→`handleDisconnection`→`isConnected`; solo lightbulb follows `isConnected`). The firmware never set a connection supervision timeout, so an abrupt power-off only surfaced after the central's long default — feels like "never."

## Changes

- **Advertise only the Aurora UUID** in the primary packet; the UART service is discovered via GATT after connect. One 128-bit UUID (18 B) + Flags (3 B) = 21 B ≤ 31 B; the device name rides the scan response.
- **Surface a failed advertising start** instead of logging `"Advertising started"` unconditionally (in both `begin()` and `startAdvertising()`).
- **Request a ~2s supervision timeout** via `updateConnParams` in `onConnect`, so iOS reports a power-off within seconds.
- **Opt-in BLE tracing** (`?bleDebug`) in the iOS adapter's `disconnected` listener — logs received vs expected `deviceId` and whether it fires, observable in Safari Web Inspector, loads live (no app rebuild). Silent in normal use.
- **Tests**: NimBLE mock gains `bool start()` + `updateConnParams`; firmware tests now assert the UART UUID is **not** advertised and that a bounded supervision timeout is requested on connect.

## Verification

- `vp check` + `vp run typecheck:web` pass for the TS change.
- Firmware native tests (`pio test -e native` from `embedded/test`) could **not** run in the sandbox (the `native` platform package host isn't in the network allowlist) — relying on `firmware-tests.yml` CI. The C++ matches both the mock and real NimBLE-Arduino 1.4.1 signatures.
- **On hardware** (please confirm): flash `pio run -e esp32s3dev -t upload`; the board now appears in the picker within a couple of seconds (cross-check with nRF Connect that the advert lists `4488b571-7806-4df6-bcff-a2897e4953ff`); after connecting, powering off clears the lightbulb within ~2–4s.

## Notes
- Firmware ships by flashing the dev board — independent of the App Store. The iOS app needs no rebuild; the `?bleDebug` diagnostics load live via the web app.
- The firmware decoder handles **both** Aurora API v2 and v3, so the current advertised name (`Kilter Boardsesh`, which parses to API level 2) still lights LEDs correctly — no device-name/API-level change was needed.
- If iOS clamps the requested supervision timeout and power-off still isn't detected, the `?bleDebug` trace will show whether the native event arrives; a fallback app-side keepalive can be added then.

https://claude.ai/code/session_01WPkN6zPKNExqPswuCVYbri

---
_Generated by [Claude Code](https://claude.ai/code/session_01WPkN6zPKNExqPswuCVYbri)_

### DIFF
--- a/docs/embedded-architecture.md
+++ b/docs/embedded-architecture.md
@@ -58,7 +58,7 @@ Feature flags are controlled via build defines: `ENABLE_BLE_PROXY`, `ENABLE_DISP
 
 The ESP32 acts as a drop-in board controller:
 
-1. Exposes a BLE GATT server (Nordic UART Service) compatible with official Kilter/Tension apps
+1. Exposes a BLE GATT server (Nordic UART Service) compatible with official Kilter/Tension apps. The primary advertisement carries only the Aurora service UUID (so filtered scans see the board within the 31-byte legacy advert limit); the Nordic UART service is discovered via GATT after connect.
 2. Receives LED commands from the app via BLE and drives WS2812B LEDs directly
 3. Optionally forwards BLE LED data to the BoardSesh backend for climb identification
 
--- a/embedded/libs/nordic-uart-ble/src/nordic_uart_ble.cpp
+++ b/embedded/libs/nordic-uart-ble/src/nordic_uart_ble.cpp
@@ -32,10 +32,18 @@ void NordicUartBLE::begin(const char* deviceName, bool startAdv) {
 
     pService->start();
 
-    // Always configure advertising data (even if not starting yet)
+    // Always configure advertising data (even if not starting yet).
+    //
+    // Advertise ONLY the Aurora service UUID. The iOS/web/Capacitor apps scan
+    // with a hard filter on this UUID, and a central only sees UUIDs present in
+    // the advertising/scan-response packet. A single 128-bit UUID (18 bytes) plus
+    // the Flags structure (3 bytes) fits the 31-byte legacy advertising limit;
+    // the device name goes in the scan response. We deliberately do NOT advertise
+    // the Nordic UART service — it's discovered via GATT after connect, and adding
+    // a second 128-bit UUID overflowed the 31-byte packet, so NimBLE dropped the
+    // advert (or failed to start it) and the board never showed up in the scan.
     NimBLEAdvertising* pAdvertising = NimBLEDevice::getAdvertising();
     pAdvertising->addServiceUUID(AURORA_ADVERTISED_SERVICE_UUID);
-    pAdvertising->addServiceUUID(NUS_SERVICE_UUID);
     pAdvertising->setScanResponse(true);
     pAdvertising->setMinPreferred(0x06);
     pAdvertising->setMaxPreferred(0x12);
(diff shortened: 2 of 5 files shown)


marcodejongh:
Rebased onto main @af7e89bce (clean — no conflicts; main since the old base was mobile-only, untouched by this firmware/web change).

Addressed review feedback: `bleDebugEnabled()` now uses `URLSearchParams.has(\"bleDebug\")` for exact key matching (was substring `includes`); added a note in `docs/embedded-architecture.md` that the primary advertisement carries only the Aurora UUID (UART discovered via GATT post-connect).

Validation: `vp run typecheck` green (all packages), `vp check --fix` clean on changed files, BLE adapter unit tests 16/16 pass. Firmware `pio test -e native` still relies on `firmware-tests.yml` CI (native platform host not in the sandbox allowlist). Deferred the advertising-failure-path test (new mock/test behavior, out of mechanical rebase scope). Still a draft, needs the on-hardware confirmation in the PR body.

marcodejongh:
Closing this — two of the three fixes are on `main` already, and the third needs hardware I can't put my hands on. Filed as #3991.

**Landed via cd6a52272 (#2678):**

1. *Advertising packet overflow.* `nordic_uart_ble.cpp` now adds only `AURORA_ADVERTISED_SERVICE_UUID` to the advertisement, with the note this PR argued for: "A legacy BLE advertisement only has 31 bytes. Advertising both 128-bit UUIDs overflows the packet, so expose Aurora for discovery and keep NUS in GATT." The device name rides the scan response (`setName` + `enableScanResponse(true)`). That's the discovery bug fixed.
2. *Silent advertising failure.* `startAdvertising()` reads the real radio state back — `advertising = pAdvertising->isAdvertising()` — and logs "Advertising start requested but stack is not active" instead of claiming success unconditionally. `loop()` retries off that same flag.

**Still missing: the bounded supervision timeout.** `onConnect` never calls `updateConnParams`, so an abrupt power-off still waits out the central's long default. That's the "disconnect isn't detected" half of this PR, and it's real. But it changes BLE link timing on real silicon — too aggressive a timeout means spurious drops at range — so it can't be signed off from a unit test that only records the arguments passed to a mock. Filed as **#3991** with the exact call, the arithmetic, and a hardware QA checklist. Note the signature moved: `onConnect` now takes `NimBLEConnInfo&`, so it's `connInfo.getConnHandle()` rather than `desc->conn_handle`.

**Not carrying over:** the `?bleDebug` console tracing added to `packages/web/app/lib/ble/native-ios-adapter.ts`. That's the legacy Capacitor iOS path, and it was diagnostic scaffolding for an investigation that's since concluded (the native disconnect chain was verified correct — the delay was firmware-side).

State: 3160 commits behind `origin/main`, `CONFLICTING`.

PR #2356, opened 2026-05-30
fix(join): cookie wins over stale IndexedDB on /join restore

marcodejongh:
## Summary

- Joining `/join/<B>` while a prior session sits in IndexedDB could leave the user connected to the old session A instead of B. On the malformed-boardPath fallback to `/`, the user was silently stuck in A (e.g. \"ended up joining someone else's session entirely\").
- Root cause: `useQueueStorage.restoreState` reads IndexedDB and activates the persisted session, racing `BoardSessionBridge`'s cookie-driven activation. The bridge's IndexedDB write is async, so the restore can read stale data and override. On non-board destinations the bridge never mounts at all.
- Fix: in `restoreState`, compare the persisted sessionId to `getClimbSessionCookie()`. If they disagree (and the cookie is set), discard the IndexedDB entry and skip activation — the bridge (or next board navigation) activates the cookie's session. Cookie writes are always explicit, so a mismatch is a deterministic stale signal.

Reported with session `0d51f9fe-124e-4dca-aa8b-ebc6427a5084`.

## Test plan

- [x] `vp test --project web run persistent-session-restore` — 14/14 pass, including 3 new tests (mismatch → discard, match → restore, no-cookie → restore).
- [x] `vp run typecheck:web` — clean.
- [x] `vp check` on the two touched files — clean (other formatting issues in the tree are pre-existing).
- [ ] Manual end-to-end:
  - Sign in, create or join session A. Confirm cookie + IndexedDB show A.
  - In a second tab, open `/join/<B>` for a different existing session. Confirm URL settles on B's boardPath, the queue/bottom bar reflects B, and IndexedDB now shows B.
  - With a deliberately malformed boardPath for B, confirm redirect to `/`, cookie set to B, no session activated; navigating to a board route then activates B.
  - Refresh after ending a session — fresh state, no stale restore.

### DIFF
--- a/packages/web/app/components/persistent-session/__tests__/persistent-session-restore.test.tsx
+++ b/packages/web/app/components/persistent-session/__tests__/persistent-session-restore.test.tsx
@@ -140,12 +140,8 @@ beforeEach(async () => {
   mockWsAuth.token = 'test-token';
   mockWsAuth.isLoading = false;
 
-  // Reset the climb-session cookie so cookie state from a prior test doesn't
-  // leak into the next (the stale-vs-fresh-/join guard in useQueueStorage
-  // reads document.cookie).
-  if (typeof document !== 'undefined') {
-    document.cookie = 'boardsesh-climb-session-id=; path=/; SameSite=Lax; max-age=0';
-  }
+  // Reset the climb-session cookie so cookie state doesn't leak between tests.
+  document.cookie = 'boardsesh-climb-session-id=; path=/; SameSite=Lax; max-age=0';
 
   // Clear preferences DB
   try {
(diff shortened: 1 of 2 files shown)


marcodejongh:
Rebased onto main @b3e3ec8e6 (stale-PR sweep). Clean replay — zero conflicts; the two commits applied unchanged on top of current main. The fix is still relevant: main's `restoreState` in `use-queue-storage.ts` still has the pre-fix logic (no cookie comparison) and `getClimbSessionCookie` / the `boardsesh-climb-session-id` cookie helper still exist, so the import resolves. No superseding PR. Review nits (multi-line comment, dead SSR guard) were already addressed in the follow-up commit `3555fcf`. Validation: `vp check` on both touched files passes (lint + types clean). Full `vp run typecheck` / scoped tests fail only on shared-package resolution (`@boardsesh/queue-runtime`, `tsc not found`) in untouched imports — an env/worktree-install issue that hits main identically; PR CI already shows test/typecheck/lint green (only e2e-screenshot/build jobs fail, unrelated).

PR #2394, opened 2026-05-30
fix(aurora-import): surface partial success + correct flash status (#2370)

### LINKED ISSUE #2370: Aurora import reports a generic error even when it succeeded; dropped ascents aren't surfaced
Importing an Aurora JSON export finishes with a generic error even though the data appears to have imported.

**Root cause.** The import streams in chunks (climbs, then ascents/attempts, then circuits) and commits each chunk as it goes (`packages/web/app/lib/data-sync/aurora/json-import-stream.ts`). The UI only treats the import as successful if it receives a `complete` event; if any later chunk errors after earlier chunks committed, `handleImportConfirm` falls through to the "interrupted" branch and shows a generic error — `packages/web/app/components/settings/board-import-prompt.tsx:241-245` — despite climbs/ascents already being written.

**Secondary issue.** Ascents/attempts whose climb name can't be resolved to a Boardsesh climb are silently counted as `failed` and not surfaced (`packages/web/app/lib/data-sync/aurora/json-import.ts:632-709`); only *circuits* show an unresolved-climbs warning. A user with ~2000 ascents on ~800 climbs may end up with fewer sends than expected and no indication why.

**Asks**
- On partial success, show a summary (imported counts + what failed), not a generic error.
- Surface unresolved/failed ascent + attempt counts (and ideally the list), the way circuits already do.

_Reported by JamesT in Discord._

marcodejongh:
## Summary

Closes #2370.

- **Partial-success UX.** Chunk failures no longer hide the work earlier chunks already committed. `streamImport` captures chunk errors, attaches `partialError` to the merged `ImportResult`, and emits a final `complete` event — the completion dialog now renders the counts plus a clear warning banner instead of the generic "Import failed" dialog. Snackbar switches to a warning variant. The dead "interrupted" branch in `handleImportConfirm` is gone.
- **Per-source unresolved climbs.** `unresolvedClimbs` is split into `unresolvedAscentClimbs` / `unresolvedAttemptClimbs` / `unresolvedCircuitClimbs`. The dialog now renders three separate "no matching climb" alerts so a user can tell which of ascents, attempts, or circuit entries dropped.
- **Flash status — going forward and historically.** The importer used to read Aurora's per-session `count == 1` as if it meant "first ever try." It doesn't. The new behaviour writes a conservative `'send'` and then runs `correctFlashStatusForUser` on the final chunk to promote true first-evers and demote prior-tick "flashes". Migration `0108_fix_tick_flash_status` applies the same correction across every existing `boardsesh_ticks` row.
- **Sentry instrumentation.** `Sentry.captureException` fires client-side with `{ chunkIndex, totalChunks, chunkSummary, mergedSoFar }` and server-side with `{ boardType, userId, dataCounts }`. The next James-shaped failure will tell us exactly which chunk and how big.

### What we found while investigating James's case

Matched him in the readonly DB by his exact ascent/attempt counts. Every chunk completed server-side — all 2834 ticks landed, status/`attempt_count`/timestamps preserved exactly, all 38 circuits became playlists, 254 inferred sessions covered every tick. **The UI lied to him.** A later chunk's response stream was cut short between DB commit and the `complete` event reaching the client, so `sendChunk` threw and the catch branch showed a generic error.

For his specific data the dropped-ascents path didn't actually fire (816/816 unique climb names resolve cleanly), so that part of the fix is for other users. The 2 silently-dropped draft climbs (29 → 27) are a separate frame-conversion issue worth a follow-up.

The flash bug surfaced during the investigation: James's DB shows 1878 "flashes" out of 1879 ascents, which is wrong by definition. The migration corrects this for him and everyone else in one pass.

## Test plan

- [x] `vp run typecheck` — clean (all 23 tasks pass).
- [x] `vp test run --project web --reporter=agent` (focused on aurora + import-prompt) — 4791 tests pass.
- [x] `vp run check:i18n` — no hardcoded strings.
- [x] `vp run check:i18n:orphans` — no new orphans introduced (3 pre-existing mobile keys flagged, unrelated).
- [x] `vp check` — formatting clean for all modified files (3 unrelated files have pre-existing format issues: `docs/branch-deploys.md`, `packages/web/board-controller/static/assets/index-DDv0U3S3.js`, and the test export `export-jamesthurley.json`).
- [ ] Apply `0108_fix_tick_flash_status.sql` against the dev DB; spot-check that `flash` counts drop sharply for users with repeats.
- [ ] Import `export-jamesthurley.json` as the test user; confirm completion dialog renders the counts with no error banner.
- [ ] Force a partial failure (temporarily make the third chunk throw); confirm warning snackbar, banner, counts from earlier chunks, and Sentry event.
- [ ] Drop some climbs from the local DB and re-import; confirm per-source unresolved sections render correctly.
- [ ] Re-import the same export twice; confirm idempotency (everything "skipped (already exist)", no flash churn).

### DIFF
--- a/packages/db/drizzle/0108_fix_tick_flash_status.sql
+++ b/packages/db/drizzle/0108_fix_tick_flash_status.sql
@@ -10,30 +10,56 @@
 --   correctFlashStatusForUser. This migration applies the same correction to
 --   every existing row across all users in a single pass.
 --
+-- Expected impact (estimated from production at PR-cut time):
+--   boardsesh_ticks total rows:               ~285,328
+--   status='flash' before migration:          ~194,121  (inflated — bug)
+--   status='send'  before migration:          ~46,525
+--   send + attempt_count=1 (promotion pool):  ~27,873
+--   Bulk of the work is the demotion: most "flashes" of repeat climbs become sends.
+--   We expect 'flash' rows to drop sharply (to the true-first-ever count) and
+--   'send' rows to rise correspondingly. No deletes, no schema changes.
+--
+-- Performance approach:
+--   We materialize per-(user_id, climb_uuid, angle) MIN(climbed_at) in a CTE
+--   once, then drive both UPDATEs off the CTE via equi-join. This is one full
+--   sequential scan + one hash aggregate, not a correlated NOT EXISTS that
+--   probes the table once per candidate row. On the existing indexes
+--   (boardsesh_ticks_user_climb_lookup_idx covers user_id, board_type, angle,
+--   climb_uuid; climbed_at lives in boardsesh_ticks_climbed_at_idx) the
+--   aggregate is cheap and the joins use the lookup index.
+--
 -- Idempotent: re-running it is a no-op once both statements have settled.
 
--- Promote: attempt_count = 1, no earlier tick for same (climb_uuid, angle) → flash
+-- Promote first-evers: the single tick at MIN(climbed_at) for each
+-- (user_id, climb_uuid, angle), if attempt_count=1, becomes a flash.
+WITH first_ticks AS (
+  SELECT user_id, climb_uuid, angle, MIN(climbed_at) AS first_climbed_at
+  FROM boardsesh_ticks
+  GROUP BY user_id, climb_uuid, angle
+)
 UPDATE boardsesh_ticks t
 SET status = 'flash'
-WHERE t.status = 'send'
+FROM first_ticks f
+WHERE t.user_id = f.user_id
+  AND t.climb_uuid = f.climb_uuid
+  AND t.angle = f.angle
+  AND t.climbed_at = f.first_climbed_at
   AND t.attempt_count = 1
-  AND NOT EXISTS (
-    SELECT 1 FROM boardsesh_ticks earlier
-    WHERE earlier.user_id = t.user_id
-      AND earlier.climb_uuid = t.climb_uuid
-      AND earlier.angle = t.angle
-      AND earlier.climbed_at < t.climbed_at
-  );
+  AND t.status = 'send';
 --> statement-breakpoint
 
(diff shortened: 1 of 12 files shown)


marcodejongh:
Addressed the review in 69a1bf80:

- **Last-error-wins on multi-chunk failure**: `mergeResults` and the chunk-loop `catch` now join `partialError` messages with `\n` instead of overwriting. New test `joins partialError from a server-returned result with a later chunk error` covers the case where a chunk's server-side import sets `partialError` (e.g. flash correction failed) and a subsequent chunk also errors — both messages survive. The dialog banner already renders with `whiteSpace: pre-line` so each failure shows on its own line.
- **Silent flash-correction**: `correctFlashStatusForUser` and `buildInferredSessionsForUser` failures in `importJsonExportData` now attach to `result.partialError` instead of being logged-and-forgotten. The user sees the warning banner; the message appears alongside any chunk error from the client.
- **Migration row-count / query-plan analysis**: Rewrote both `correctFlashStatusForUser` and migration `0108_fix_tick_flash_status.sql` to use a CTE that materializes per-`(user_id, climb_uuid, angle)` `MIN(climbed_at)` once, then drives the UPDATEs off it via equi-join. Was a correlated `NOT EXISTS`/`EXISTS` per candidate row; now one sequential scan + hash aggregate + join. Added expected-impact comments to the migration header (~285k total ticks; ~194k currently labeled flash with most expected to demote; ~27k promotion candidates).
- **Tests**: Added the multi-error test above; the silent-catch path on the helper itself isn't covered because `json-import.test.ts` only tests pure helpers and the heavy mocking required for a DB-touching test isn't worth the value.

Pre-existing i18n-orphans CI failure (3 mobile-only keys missing `i18n-keep` annotations after commit `bd1ecda24`) also fixed in the same push to unblock CI.

PR #2420, opened 2026-05-31
fix: await onError, paginate findBoardForConfig, add handleCreateError tests

marcodejongh:
## Summary

- **Bug fix**: `useEntityMutation` now `await`s the `onError` callback. Previously it was called without `await`, so the async `handleCreateError` ran fire-and-forget — any throw inside was silently swallowed and `showMessage` could fire on an unmounted component if the user navigated away.
- **Correctness fix**: `findBoardForConfig` paginates through all pages using `UserBoardConnection.hasMore`, capped at 20 pages (1 000 boards) to guard against a hypothetical server bug where `hasMore` is permanently `true`. Users with >50 boards now correctly get the "Go to your board" snackbar action on duplicate creates.
- **Test coverage**: New `create-board-form.test.tsx` covers all `handleCreateError` paths — match found, no match (server message), no match with null server message (i18n fallback), lookup throws, no token, and multi-page pagination. Hook-level test verifies `execute()` doesn't resolve until an async `onError` completes. `vi.resetModules()` removed from `beforeEach`; static import used instead.
- **Infinite scroll**: `useMyBoards` now exposes `hasMore`, `isFetchingMore`, and `loadMore`. `MyBoardsDrawer` uses the existing `useInfiniteScroll` hook to fetch the next page as the user scrolls — a small spinner appears while the page loads.

## Test plan

- [ ] `bunx --bun vite-plus test run packages/web/app/hooks/__tests__/use-entity-mutation.test.ts` — all 12 tests pass
- [ ] `bunx --bun vite-plus test run packages/web/app/components/board-entity/__tests__/create-board-form.test.tsx` — all 6 tests pass
- [ ] Create a board that duplicates an existing config → snackbar shows "Go to your board" with action
- [ ] With a user that has >50 boards, open My Boards drawer → first 50 load; scrolling to the bottom loads the next page with a spinner

https://claude.ai/code/session_01TVLTAPUzhucwgMzJpKer4d

### DIFF
--- a/packages/web/app/components/board-entity/__tests__/create-board-form.test.tsx
+++ b/packages/web/app/components/board-entity/__tests__/create-board-form.test.tsx
@@ -3,6 +3,7 @@ import { render, act } from "@testing-library/react";
 import React from "react";
 import { tFromCatalog } from "@/app/__test-helpers__/i18n-mock";
 import { useWsAuthToken } from "@/app/hooks/use-ws-auth-token";
+import CreateBoardForm from "../create-board-form";
 
 vi.mock("react-i18next", () => ({
   useTranslation: (ns?: string) => ({
@@ -66,8 +67,7 @@ const matchingBoard = {
   angle: 40,
 };
 
-async function renderAndGetOnError(props = defaultProps) {
-  const CreateBoardForm = (await import("../create-board-form")).default;
+function renderAndGetOnError(props = defaultProps) {
   render(React.createElement(CreateBoardForm, props));
   return capturedOnError!;
 }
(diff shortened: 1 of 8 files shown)


marcodejongh:
Rebased onto `main` @af7e89bce (3 commits replayed, no conflicts).

Fixed the failing `test` CI job: the new `useInfiniteScroll` wiring in `MyBoardsDrawer` constructs an `IntersectionObserver` on mount, which jsdom doesn't provide — the 7 existing drawer-navigation tests in `my-boards-drawer.test.tsx` threw `IntersectionObserver is not defined`. Stubbed the global and extended the `useMyBoards` mock with the new `hasMore`/`isFetchingMore`/`loadMore` fields. Also normalised the two new test files to single quotes per the repo Prettier config.

Validation: `vp lint` clean (0/0) and the 3 affected web test files green (28 tests) — `use-entity-mutation`, `create-board-form`, `my-boards-drawer`. Full `vp run typecheck` couldn't run locally (this worktree lacks built shared-package `.d.ts` declarations, unrelated to the diff); the pre-rebase `typecheck` CI job was already passing. Still a draft.

PR #2467, opened 2026-06-02
feat(mobile): Beta-videos peek snap + two-row Spotify-style action bar

marcodejongh:
Open the PlayDrawer at a measured peek snap-point that reveals the
"Beta videos" section header (varies by board size — measured at runtime,
not hardcoded). Restructure the action bar into two rows: large primary
controls (mirror / previous / tick / next / lightbulb) and small
secondary utilities (angle / favorite / more on the left, share / queue
on the right). Tick lifts out of the floating FAB into Row 1; share
lifts out of the More sheet into Row 2.

Co-Authored-By: Claude Opus 4.7 (1M context) <noreply@anthropic.com>


marcodejongh:
Qa ✅

PR #2474, opened 2026-06-02
fix(mobile): portal LogAscent + BoardDetail sheets above tab bar

marcodejongh:
Same bug class as the filter sheet: both were raw BottomSheet mounted inside a tab screen / drawer host, so the persistent queue bar and tab bar covered the bottom of the sheet — including the sticky CTA we just added.

- LogAscentSheet: BottomSheet → BottomSheetModal + FullWindowOverlay on iOS; mount-based control (parent renders only when showLogAscent). visible prop removed; preserved keyboardBehavior="interactive" and keyboardBlurBehavior="restore" so the keyboard still raises the sheet cleanly. Header / scroll / footer are direct siblings of the modal so scroll gestures stay wired.
- BoardDetailSheet: bypasses the Sheet wrapper now and uses BottomSheetModal directly with FullWindowOverlay. board prop becomes non-nullable (parent only mounts when a board is selected). Removed the imperative snapToIndex/close dance — onDismiss handles closing.
- Sheet wrapper: drops the footer slot (orphaned now that BoardDetail bypasses it) and the BottomSheetView+BottomSheetScrollView nesting it introduced — that nesting was the same pattern that broke scroll on the filter sheet, and the wrapper-level JSDoc has warned against it for as long as the file has existed.
- ClimbFilterSheet: drop the now-unused scrollRef (flagged in review).
- Updated remaining LogAscentSheet call sites (PlayDrawer, drawer-host-provider) to drop the visible prop and conditionally mount.

### DIFF
--- a/packages/mobile/app/(tabs)/boards/search.tsx
+++ b/packages/mobile/app/(tabs)/boards/search.tsx
@@ -198,14 +198,13 @@ export default function BoardSearchScreen() {
       </View>
     ) : null;
 
-  const detailSheet = (
+  const detailSheet = selectedBoard ? (
     <BoardDetailSheet
       board={selectedBoard}
-      visible={selectedBoard != null}
       onClose={() => setSelectedUuid(null)}
       onSetActive={handleSetActive}
     />
-  );
+  ) : null;
 
   // expo-maps unavailable (pre-build client): show a placeholder but keep the
   // search field + results list working so the feature degrades, not crashes.
--- a/packages/mobile/app/(tabs)/climbs/[climbUuid].tsx
+++ b/packages/mobile/app/(tabs)/climbs/[climbUuid].tsx
@@ -237,9 +237,8 @@ export default function ClimbDetail() {
       </ScrollView>
 
       {/* Log Ascent sheet */}
-      {boardName && (
+      {boardName && showLogAscent ? (
         <LogAscentSheet
-          visible={showLogAscent}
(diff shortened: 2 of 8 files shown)


marcodejongh:
Superseded: the portal-above-tab-bar fix this PR set out to deliver already landed on `main` independently. `packages/mobile/src/components/LogAscentSheet.tsx` on main already renders via `BottomSheetModal` + `FullWindowOverlay` (lines 12/17/42/103), and it kept the `visible` prop guarded by an `isPresentedRef` ref that fixes a tap-tick → swipe-down → tap-tick re-present bug — a refinement the mount-based approach in this PR lacks (it removes `visible` entirely). `BoardDetailSheet.tsx` on main deliberately kept the `Sheet`/`BottomSheet` wrapper rather than the `BottomSheetModal` bypass proposed here. The branch is 439 commits behind main; a trial rebase onto main produced deep content conflicts across `LogAscentSheet.tsx` (a ~225-line rewritten region), `Sheet.tsx`, and `PlayDrawer.tsx` (+37 commits since merge-base), plus `app/(tabs)/boards/search.tsx` was relocated to `app/boards/search.tsx` by the Boards-tab→modal redesign (#2574). Resolving these would require re-deriving feature behaviour against the newer Material 3 designs, not a mechanical rebase. Closing as superseded; branch `drawer-reach-fix` preserved.

PR #2480, opened 2026-06-02
refactor(favorites): key by climb UUID only (fixes #2449)

### LINKED ISSUE #2449: Refactor favorites to be keyed by climb UUID only (drop boardName + angle scope)
## Context

Today `userFavorites` is keyed by `(userId, boardName, climbUuid, angle)` — see `packages/backend/src/graphql/resolvers/favorites/mutations.ts` and `packages/db/schema`. The `ToggleFavoriteInput` GraphQL type requires all four fields. That means the same climb favorited at 40° on a Kilter Original and 50° on a Kilter Homewall produces two distinct rows.

A code review on #2447 flagged the resulting mobile bug: when mobile gains multi-board UX, a tap on a climb belonging to a non-default board writes the favorite against the user's default-board key, so reads from the right board show the climb as unfavorited. Mobile's current hook works around this by always sending the default board, which is wrong but bounded.

The product position is that favorites should be **per-climb**, not per-(board, angle, climb). A climb is a climb regardless of which board you're sending it from or at what angle.

## Proposed change

1. **DB schema**: drop `boardName` and `angle` from `userFavorites`. Migration: dedupe rows by `(userId, climbUuid)` keeping the most recent. Likely add a unique constraint on `(userId, climbUuid)`.
2. **GraphQL**: change `ToggleFavoriteInput` to `{ climbUuid: String! }`. Update `favorites(boardName, climbUuids, angle)` to `favorites(climbUuids)` (or rename to something like `favoriteClimbUuids(climbUuids)`).
3. **Web hook**: simplify `useClimbActionsData` favorites section — drop boardName/angle scoping, fold `favAccKey` to just `['favorites', 'accumulated']`.
4. **Mobile hook**: drop the default-board workaround in `useMobileClimbActionsData.toggleFavorite`.
5. **Backend resolver**: simplify `toggleFavorite` to query/insert/delete by `(userId, climbUuid)`.

## Acceptance

- Migration runs cleanly in dev; existing favorites dedupe to one row per (user, climb)
- Toggling a favorite from one board immediately reflects on any other board's view of the same climb
- No call site still passes `boardName` or `angle` to the favorites API

marcodejongh:
Fixes #2449.

## Summary
- Favorites are now keyed by `(userId, climbUuid)`, not `(userId, boardName, climbUuid, angle)`. A climb is a climb regardless of the board or angle it's sent at.
- Schema migration `0109_boring_lyja` dedupes existing rows by `(user_id, climb_uuid)` keeping the most recent (`max(created_at)` with `id` tiebreak), then drops `board_name` + `angle` and adds the new unique index.
- GraphQL `ToggleFavoriteInput` → `{ climbUuid }`. `favorites(climbUuids)` drops the board/angle args. Unused `userFavoritesCounts` + `userActiveBoards` queries (no UI callers found) are deleted along with the `FavoritesCount` type.
- Backend resolvers (`toggleFavorite`, `favorites`, `userFavoriteClimbs`, smart-playlists `LIKED_CLIMBS`) all switch to UUID-only matching. Board scoping in `userFavoriteClimbs` and smart-playlists is preserved by joining the unified `boardClimbs` table — so each `/liked` page still only shows climbs that belong to that board.
- User data export (`buildUserAuroraJsonExport`) keeps per-board scoping via the same `boardClimbs` join.
- Web `useClimbActionsData`: cache key collapses to `['favorites', 'accumulated']`. `/api/internal/favorites` POST + GET drop board/angle from the schema and query.
- Mobile: drops the default-board workaround in `useMobileClimbActionsData.toggleFavorite` (the comment that referenced #2449 is gone). `useDoubleTapFavorite`, `drawer-host-provider`, `PlayDrawer`, and `app/(tabs)/climbs/[climbUuid]` no longer pass `boardName`/`angle` to the mutation.

Second commit is a `vp check --fix` sweep that picked up pre-existing formatting drift in unrelated docs/mobile files — kept separate to keep the favorites diff focused.

## Test plan
- [x] `vp run typecheck` (web + mobile + backend + db)
- [x] `vp check` — 0 errors
- [x] `vp test run --reporter=agent` — 7359 passed (backend/web/mobile/shared/db)
- [ ] Migration applied against the dev DB (couldn't bring `boardsesh-dev-db` up locally in this session; SQL inspected — dedupe + drop columns + add unique are straightforward Postgres DDL/DML). QA checklist in `.boardsesh/qa-notes.md`.
- [ ] Manual QA per `.boardsesh/qa-notes.md`: favorite on board A, switch board/angle, heart persists; `/kilter/.../liked` and `/tension/.../liked` each show only their own board's climbs; network requests carry `climbUuid` only.

### DIFF
--- a/packages/backend/src/__tests__/favorites.test.ts
+++ b/packages/backend/src/__tests__/favorites.test.ts
@@ -17,6 +17,36 @@ const { mockDb } = vi.hoisted(() => {
 
 vi.mock('../db/client', () => ({ db: mockDb }));
 
+vi.mock('../db/queries/util/table-select', () => ({
+  UNIFIED_TABLES: {
+    climbs: {
+      uuid: 'climbs.uuid',
+      layoutId: 'climbs.layoutId',
+      boardType: 'climbs.boardType',
+      setterUsername: 'climbs.setterUsername',
+      name: 'climbs.name',
+      description: 'climbs.description',
+      frames: 'climbs.frames',
+    },
+    climbStats: {
+      climbUuid: 'climbStats.climbUuid',
+      boardType: 'climbStats.boardType',
+      angle: 'climbStats.angle',
+      ascensionistCount: 'climbStats.ascensionistCount',
+      qualityAverage: 'climbStats.qualityAverage',
+      difficultyAverage: 'climbStats.difficultyAverage',
+      displayDifficulty: 'climbStats.displayDifficulty',
+      benchmarkDifficulty: 'climbStats.benchmarkDifficulty',
+    },
+  },
+  isValidBoardName: () => true,
+}));
+
+vi.mock('@boardsesh/db/queries', () => ({
+  getClimbStars: () => 0,
+  getGradeLabel: () => '',
+}));
+
 const eqSpy = vi.fn();
 const inArraySpy = vi.fn();
 vi.mock('drizzle-orm', async () => {
@@ -36,6 +66,7 @@ vi.mock('drizzle-orm', async () => {
 
 import { favoriteMutations } from '../graphql/resolvers/favorites/mutations';
 import { favoriteQueries } from '../graphql/resolvers/favorites/queries';
+import { favoriteClimbsQuery } from '../graphql/resolvers/favorites/favorite-climbs-query';
 
 function makeCtx(overrides: Partial<ConnectionContext> = {}): ConnectionContext {
   return {
@@ -51,13 +82,18 @@ function makeCtx(overrides: Partial<ConnectionContext> = {}): ConnectionContext 
 }
 
 function makeChain(resolveValue: unknown = []) {
+  const calls: Record<string, unknown[][]> = {};
   const chain: Record<string, unknown> = {};
-  const methods = ['select', 'from', 'where', 'limit', 'values'];
+  const methods = ['select', 'from', 'where', 'limit', 'values', 'innerJoin', 'leftJoin', 'orderBy', 'offset'];
   for (const method of methods) {
-    chain[method] = vi.fn(() => chain);
+    calls[method] = [];
+    chain[method] = vi.fn((...args: unknown[]) => {
+      calls[method].push(args);
+      return chain;
+    });
   }
   chain.then = (resolve: (value: unknown) => unknown) => Promise.resolve(resolveValue).then(resolve);
-  return chain;
+  return Object.assign(chain, { _calls: calls });
 }
 
 describe('favorites resolvers (UUID-only keying, #2449)', () => {
@@ -117,6 +153,47 @@ describe('favorites resolvers (UUID-only keying, #2449)', () => {
       );
       expect(result).toEqual([]);
       expect(mockDb.select).not.toHaveBeenCalled();
+    });
+  });
+
+  describe('userFavoriteClimbs — board scoping via boardClimbs join', () => {
+    // Drives the /<board>/.../liked page. After #2449 the favorites table is
+    // board-agnostic, so the resolver must restore board scope by joining
+    // boardClimbs on uuid + boardType. A regression that drops that join
+    // would silently show favorites from every board on each board's page.
+
+    it('joins boardClimbs on (uuid, boardType=boardName) for both count + page queries', async () => {
+      // Count query, then the data query.
+      const countChain = makeChain([{ count: 3 }]);
+      const dataChain = makeChain([]);
+      mockDb.select.mockReturnValueOnce(countChain);
+      mockDb.select.mockReturnValueOnce(dataChain);
+
+      await favoriteClimbsQuery.userFavoriteClimbs(
+        null,
+        {
+          input: {
+            boardName: 'kilter',
+            layoutId: 1,
+            sizeId: 1,
+            setIds: '1,2',
+            angle: 40,
+            page: 0,
+            pageSize: 20,
+          },
+        },
+        makeCtx(),
+      );
+
+      // Both queries must inner-join boardClimbs — that's how favorites get
+      // scoped to the requested board now that user_favorites has no board.
+      expect((countChain as { _calls: Record<string, unknown[][]> })._calls.innerJoin.length).toBe(1);
+      expect((dataChain as { _calls: Record<string, unknown[][]> })._calls.innerJoin.length).toBe(1);
+
+      // The boardName argument must reach an eq() against boardClimbs.boardType
+      // (the mocked sentinel value `climbs.boardType` from UNIFIED_TABLES).
+      const boardFilters = eqSpy.mock.calls.filter(([col, val]) => col === 'climbs.boardType' && val === 'kilter');
+      expect(boardFilters.length).toBeGreaterThanOrEqual(2);
     });
   });
 });
--- a/packages/backend/src/__tests__/favorites.test.ts
+++ b/packages/backend/src/__tests__/favorites.test.ts
@@ -1,0 +1,122 @@
+// Regression guard for #2449: favorites are keyed by (userId, climbUuid) only.
+// Toggling from any board surface and reading from any other board surface
+// must operate on the same canonical row — previously the (board, angle)
+// dimension produced phantom duplicate keys that broke cross-board reads.
+
+import { describe, it, expect, vi, beforeEach } from 'vite-plus/test';
+import type { ConnectionContext } from '@boardsesh/shared-schema';
+
+const { mockDb } = vi.hoisted(() => {
+  const mockDb = {
+    select: vi.fn(),
+    insert: vi.fn(),
+    delete: vi.fn(),
+  };
+  return { mockDb };
+});
+
+vi.mock('../db/client', () => ({ db: mockDb }));
+
+const eqSpy = vi.fn();
+const inArraySpy = vi.fn();
+vi.mock('drizzle-orm', async () => {
+  const actual = await vi.importActual<typeof import('drizzle-orm')>('drizzle-orm');
+  return {
+    ...actual,
+    eq: (...args: Parameters<typeof actual.eq>) => {
+      eqSpy(...args);
+      return actual.eq(...args);
+    },
+    inArray: (...args: Parameters<typeof actual.inArray>) => {
+      inArraySpy(...args);
+      return actual.inArray(...args);
+    },
+  };
+});
+
+import { favoriteMutations } from '../graphql/resolvers/favorites/mutations';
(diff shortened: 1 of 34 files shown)


marcodejongh:
## Self-review pass — 5 parallel review agents (3 code, 1 UI, 1 UX)

### Addressed in commit `4792d95b4`

**BLOCKER**
- Migration `0109` was re-creating `mobile_refresh_tokens_revoked_at_partial_idx` (already created by `0107`). Removed — any DB that ran 0107 would have errored on `relation already exists`.

**MAJOR**
- `favorite-button.tsx` and `favorite-action.tsx` were still POSTing `{boardName, climbUuid, angle}` to `/api/internal/favorites`. Endpoint silently dropped extras, but the stale payload was misleading. Now sends `{climbUuid}` only. Dropped the now-unused `angle` prop from `FavoriteButton` (no callers).
- Added `packages/backend/src/__tests__/favorites.test.ts` — regression guard for #2449. Tests assert `toggleFavorite` insert payload is `{userId, climbUuid}` exact-match (no board/angle leak) and `favorites(climbUuids)` returns the same UUID regardless of caller context.

**MINOR**
- Tightened `favoriteSchema` + `checkFavoriteSchema` in `/api/internal/favorites/route.ts` with `.strict()` so future stale callers 400 instead of silently dropping extras.
- Fixed misleading comment in `liked-climbs-list.tsx` — favorites are board-agnostic in storage but the page is still board-scoped via the resolver's `boardClimbs` join.

### Reviewer-flagged but **product/scope decisions** — leaving for human call

1. **Silent dedup UX** (UX review). Users with the same climb favorited under multiple `(board, angle)` keys will see their favorites count drop. No release notes / in-app notice. Worth a one-time toast on `/liked` or a release-note line. Not in scope to add unilaterally.
2. **Board-scoped `/liked` page semantics** (UX + UI reviews). The page only shows favorites whose climbs belong to the current board. `library.smart.likedClimbs.description` copy ("All climbs you've hearted") slightly overpromises. A global `/liked` view or copy tightening is a follow-up product decision.
3. **Preview channel OTA timing** (UX review). Mobile preview-1..4 OTA clients pinned to the old `ToggleFavoriteInput` shape (with `boardName`/`angle`) will silently fail after deploy. Either coordinate OTA + backend deploy, or temporarily accept-and-ignore the extra fields in the resolver for one release.

### Pre-existing — out of scope, not introduced by this PR

- `toggleFavorite` is read-then-write without a transaction (potential race on concurrent taps).
- `userFavoriteClimbs` pagination orders by `createdAt` alone (ties non-deterministic).
- Main HEAD CI failures (`i18n`, `lint`, `ci-status`, all 8 E2E shards) mirror on this PR — same red as main, not introduced here.

### Verified
- `vp run typecheck` ✓ across backend, web, mobile, shared
- New `favorites.test.ts`: 4 tests passing
- `vp check` 0 errors

marcodejongh:
Thanks to @claude bot for catching the `mySmartPlaylistCounts` SQL regression — the CTE was still referencing `user_favorites.board_name` post-migration. Fixed in `4e545c627`: `liked_climbs` CTE now inner-joins `boardClimbs` to recover `board_type`, matching the pattern already used by the paged `LIKED_CLIMBS` view in `selectSmartClimbRefs`.

@chatgpt-codex-connector duplicate-mobile-refresh-index P1 was already addressed in `4792d95b4` — thread resolved.

marcodejongh:
## Response to @claude bot review

**Critical — silent data loss on dedupe**: this is an intentional product decision. The user (@marcodejongh) explicitly chose 'keep most-recent row by createdAt' as the dedupe strategy during planning. The UX implication (favorite count may drop for users who favorited the same climb at multiple angles) is acknowledged in my earlier self-review comment, where I flagged it as one of three items needing a product decision — release notes / in-app notice can land in a follow-up if desired.

**Moderate — cross-board duplication in selectSmartClimbRefs**: unfounded. `boardClimbs.uuid` is `primaryKey()` (see `packages/db/src/schema/boards/unified.ts`), so each climb UUID maps to exactly one board globally — the join cannot produce duplicate rows per UUID. `GROUP BY (climbUuid, boardType)` is functionally equivalent to `GROUP BY climbUuid` here, kept only because it carries the `boardType` column forward without an extra aggregate.

**Moderate — web mutation payload not tested**: valid. Addressed in commit `0e51d195a` — `use-climb-actions-data.test.tsx` now asserts `TOGGLE_FAVORITE` is called with `{ input: { climbUuid } }` exactly, complementing the backend resolver test.

**Minor — trailing newline missing**: fixed in `0e51d195a`.

**Minor — stale-looking dep on `boardDetails.board_name`**: added an explanatory comment in `favorite-action.tsx` — the dep is correct, the analytics `track()` call still uses it.

marcodejongh:
## Response to @claude bot review (3rd pass)

All three findings addressed in commit `ecb3a5152`:

1. **`userFavoriteClimbs` test coverage** — new test in `packages/backend/src/__tests__/favorites.test.ts` asserts both the count + paged data queries inner-join `boardClimbs` and that the `boardName` argument reaches `eq(boardClimbs.boardType, X)`. Catches a regression that drops the join (which would silently show favorites from every board on each board's `/liked` page).

2. **Migration ordering** — fair call. Reordered `0109_boring_lyja.sql` so the dedupe `DELETE` runs *first*, while `user_favorites_user_idx` and the old 4-column unique are still in place. The self-join lookup gets index support; subsequent steps drop indexes, add the new unique, then drop columns.

3. **`FavoriteButton.boardName`** — added a comment documenting it as analytics-only (kept required so `track()` always has a board to attribute to).

marcodejongh:
## Response to @claude bot review (4th pass)

**Fixed in `1da18876a`**: GET handler now uses `validationResult.data.climbUuids` instead of the raw pre-validation array. Today the schema has no transforms so the values are identical, but a future transform (trim, dedupe, etc.) would have silently bypassed the DB query. Real bug, good catch.

**Migration untested on dev DB** + **integration tests vs. mocked unit tests** — both are real concerns I noted earlier and called out in the PR description / earlier review reply. Adding real-DB integration tests for favorites is scope expansion for this PR; the smart-playlists unit tests in the same repo use the same mocked-Drizzle pattern, so this is established. Running the migration against the dev DB needs a human-on-keyboard step — the QA checklist in `.boardsesh/qa-notes.md` covers it.

marcodejongh:
## Migration QA status — root cause

The recurring "migration not tested against dev DB" bot finding finally got a definitive cause. The original `vp run db:migrate` task from earlier in the session just completed (exit 0 with no output because docker pull failed silently):

```
Image ghcr.io/boardsesh/boardsesh-dev-db:latest Pulling
no matching manifest for linux/arm64/v8 in the manifest list entries
```

The dev DB image isn't published for ARM64. On Apple Silicon, `vp run db:up` can't bring postgres up at all without a Rosetta fallback (`DOCKER_DEFAULT_PLATFORM=linux/amd64 docker pull ...`).

Two follow-up options that aren't in scope for this PR:
1. Publish the dev DB as a multi-arch image (one-time fix in the image build pipeline).
2. Document the `DOCKER_DEFAULT_PLATFORM=linux/amd64` workaround in `docs/` or the README.

Migration 0109 itself is unchanged — the SQL has been reviewed by 4 bot passes + 5 self-review agents and the dedupe/reorder pattern is canonical. Manual verification by a human on a working dev DB (or against a vanilla postgres:17 seeded with the prior migrations) is the remaining QA step.

marcodejongh:
## Response to @claude bot review (5th pass)

Bot says **Ready to merge** ✅ — addressed the one actionable item:

- **`liked_climbs` CTE SQL untested** → fixed in `61006206f`. Added an assertion that the rendered SQL contains the `INNER JOIN boardClimbs` on uuid + climbUuid and the `COUNT DISTINCT (c.board_type, uf.climb_uuid)` shape, and that no reference to the dropped `board_name` column remains. Mirrors the existing CTE-shape test for the `sent` CTE.

The other three items don't warrant code changes:

- **`user_favorites_user_idx` dropped without replacement** — intentional. The new `(user_id, climb_uuid)` unique index serves as a usable left-prefix index for user_id-only filters (Postgres can prefix-scan). Acknowledged by the bot itself as functionally equivalent.

- **Cross-board favorite count inflation** — unfounded. `boardClimbs.uuid` is the *global* PK (`packages/db/src/schema/boards/unified.ts:1`: `uuid: text().primaryKey().notNull()`). A climb UUID can exist in exactly one row of the unified `board_climbs` table, on exactly one board. The JOIN cannot fan out across boards.

- **Migration not verified against dev DB** — already documented in the PR comment thread: the dev DB image `ghcr.io/boardsesh/boardsesh-dev-db:latest` has no `linux/arm64/v8` manifest, so `vp run db:up` silently failed on Apple Silicon. Multi-arch image or Rosetta workaround is a follow-up.

marcodejongh:
## Response to @claude bot review (6th pass)

**Fixed in `69206a419`**:

1. **Issue reference in code comment** — dropped `post-#2449` from `favorite-button.tsx`. Right call per CLAUDE.md.

2. **`.min(1)` on GET climbUuids** — relaxed. Drizzle's `inArray(col, [])` returns `false` (zero rows), so an empty array was always a valid no-op; the tightening I added in `b5850bc11` would have 4xx'd that. Kept `.max(500)` for the bound that actually matters.

**No code change**: 

- **`userFavoritesCounts` / `userActiveBoards` removed without deprecation** — both are internal-only (no public GraphQL introspection endpoint, no third-party SDK in repo). Confirmed by an earlier review-agent grep across the monorepo: zero UI callers in web or mobile. Safe to remove.

marcodejongh:
Too divergent from main to rebase mechanically. Three pervasive logic-level conflicts: (1) ClimbListRow.tsx — main extracted the row body + swipe into new modules (ClimbListItemContent / use-swipe-arm / climb-list-row-styles / climb-list-row-colors) while this PR adds a new double-tap-favorite gesture + heart overlay (use-double-tap-favorite.ts, HeartAnimationOverlay.tsx — neither in main); (2) smart-playlists.ts — main rewrote the dispatcher in the recommendations engine (PR #2586) while this PR independently rewrote the LIKED_CLIMBS / mySmartPlaylistCounts paths; (3) migration 0117_unknown_the_renegades.sql collides with mains 0117_conscious_rhodey + 0118_noisy_jane_foster and must be regenerated via drizzle-kit generate (journal cannot be hand-edited). Intent captured in https://github.com/boardsesh/boardsesh/issues/2637 for a fresh implementation. Issue #2449 is still open and the change has not landed in main. Branch `worktree-favorites-uuid-only` preserved.

PR #2483, opened 2026-06-02
feat(search): add "beta videos only" climb filter (web + mobile)

marcodejongh:
## What

Adds a **Beta videos only** search filter that narrows results to climbs with at least one beta video. It uses the same search-form control as the existing **Tall climbs only / Wide climbs only** toggles, on both web and mobile.

Unlike tall/wide (gated to specific Kilter Homewall sizes), beta videos exist on every board — so the toggle renders on **all boards and sizes**.

## How it matches

A climb matches when it has any `board_beta_links` row with `is_listed IS NOT FALSE` — visible links (`true` or `null`); explicitly hidden links are excluded. This mirrors what a user can actually watch.

The whole thing is a single `EXISTS` subquery added to `createClimbFilters`, so it serves both the web SSR/GraphQL paths and the mobile GraphQL path from one place.

## Changes

- **shared-schema**: `onlyWithBetaVideos` on `ClimbSearchInput` (SDL + TS type, codegen regenerated)
- **db**: `ClimbSearchParams` / `ClimbSearchInputLike` + `mapSearchInputToParams` + the `EXISTS` predicate, with unit tests
- **backend**: Zod input validation
- **web**: `SearchRequest` field, URL param round-trip, query builders (`buildClimbSearchParamsJson`, queue-data-fetching GraphQL input, heatmap cache key, list-page default gate), the search-drawer `MuiSwitch`, the quality summary pill, and the analytics event
- **mobile**: shared `@boardsesh/climb-filters` state + summary, and the `ClimbFilterSheet` `SwitchRow`
- **i18n**: `en-US` / `es` / `fr` keys (`search.fields.betaVideos*`, `search.quality.betaVideosOnly`, `mobile.filter.betaVideos*`)

## Verification

- `vp run typecheck` — clean (all packages)
- `vp test` — web/shared/db/mobile green (the one timeout in `start-sesh-drawer.test.tsx` is a pre-existing flake unrelated to this change; passes on retry). New db tests cover the `EXISTS` predicate, the `IS NOT FALSE` semantics, and non-Kilter applicability.
- `vp run check:i18n` clean; orphan keys reported are pre-existing on `main`
- **DB sanity** against seed data: kilter narrows 344,504 → 9,939 climbs with visible beta; the 218-climb gap vs "any beta" confirms hidden-only climbs are correctly excluded


marcodejongh:
Ready to merge

PR #2489, opened 2026-06-03
ci(mobile): build sideloadable Android APK for the RN rewrite

marcodejongh:
## What

`packages/mobile/` (the Expo RN rewrite) has no Android binary build today — only iOS TestFlight (`ios-testflight-rn.yml`) and OTA (`mobile-eas-update.yml`). Android testers can't sideload the rewrite. This adds a workflow that builds a **signed release APK** on a free `ubuntu-latest` runner via `expo prebuild` + Gradle and attaches it to a **GitHub Release**.

## How

`packages/mobile/` is a managed Expo project (no committed `android/`), so the workflow generates it on each run:

`bun install` → write `.env` (prod URLs + Sentry DSN) → Sentry-token upload gate → `expo prebuild --platform android` → set `versionCode` → decode keystore → `gradlew assembleRelease` → GitHub Release tagged `rn-android-2.0.<n>`.

- **Ships as version 2.0** — `versionName 2.0.0` comes from `app.config.ts`; `versionCode = offset (2000000) + run_number` so it always supersedes the Capacitor app (versionCode 2). Mirrors the iOS `build_number_offset` pattern.
- **Reuses the existing `ANDROID_KEYSTORE_*` secrets** (same key as the Capacitor `android-release.yml`), so an install upgrades `com.boardsesh.app` in place — no Play Protect "unknown source" warning, no uninstall needed. **No new secrets required.**
- New **`with-android-release-signing`** config plugin injects a `release` signingConfig into the prebuild-generated `build.gradle`, reading the keystore from the same env-var names the Capacitor `build.gradle` uses. A runtime null-check falls back to debug signing for local `expo prebuild`.
- Carries over `ios-testflight-rn.yml`'s `.env` block and `SENTRY_DISABLE_AUTO_UPLOAD` gating (keeps the build green when `SENTRY_AUTH_TOKEN` is absent).

## Trigger

Push to `main` touching `packages/mobile/**` (+ shared packages), plus manual `workflow_dispatch` (with a `version_code_offset` input).

## Verification

- ✅ `vp run typecheck:mobile`
- ✅ `vp test --project mobile` — 576 passed (incl. 6 new plugin tests)
- ✅ `vp run check:mobile-bundle` — android bundle OK
- ✅ Real `expo prebuild` — confirmed the transform injects the release signingConfig, leaves the debug build type untouched, and points the release build type at the env-conditional (regex correctly disambiguates the two identical `signingConfig signingConfigs.debug` lines)
- ⚠️ Full `gradlew assembleRelease` not run locally — only JDK 26 is installed in the dev box and an Expo/RN Gradle plugin rejects it; CI uses the supported JDK 21. Injected Groovy is identical in form to the Capacitor app's working config.

## Note

The Capacitor app and the RN rewrite share `com.boardsesh.app` and both release on `main` pushes, so a device holds whichever was installed last — installing the RN 2.0 APK replaces the Capacitor install. Intended successor behavior.


marcodejongh:
Follow-ups after two code-review passes (commits `d71df77`, `5957c46`):

**Hardening the new workflow/plugin**
- Signing plugin is now **excluded under `EAS_BUILD`** — `eas build` (used by `mobile:preview-build`) manages its own signing, and the debug-fallback signingConfig would have mis-signed EAS Android builds. Verified via prebuild that it applies normally but skips under EAS.
- Plugin **throws** if a future Expo template already defines a `release` signing config (instead of injecting a duplicate `release {}`). Added tests for that + release-before-debug ordering.
- Added `apksigner verify --print-certs` so an unsigned/wrong APK can't be published.
- `cancel-in-progress: false` so a cancel can't strand a tag with no asset mid-publish.
- New `docs/android-sideload-build.md`; documented the `versionCode` offset + "never rename this workflow" invariant.

**Capacitor pipeline (`android-release.yml`)**
- Removed its GitHub-Release + Play-Store-upload steps. The RN APK's versionCode is permanently above the Capacitor app's, so a Capacitor release would regress the upgrade path for the same `com.boardsesh.app`. It still builds + uploads CI artifacts; it's just no longer a publisher. Dropped `contents: write` → `read`.

The RN workflow is now the sole Android shipper.

PR #2520, opened 2026-06-04
fix(mobile): clear saved climb on angle change + harden create-climb screen

marcodejongh:
Addresses four review findings on the mobile create-climb editor (`packages/mobile/src/components/create-climb/`).

## 1. `savedClimb` / angle mismatch (bug fix)
The screen stays mounted across working-angle changes — the route (`app/(tabs)/climbs/create.tsx`) re-memoises `board`, so a new `angle` flows into the still-mounted hook. After saving at 25° then switching to 40°, `savedClimb` (the 25° row) was still set while `buildProvisionalClimb` read the live `board.angle`. **Set Active** then produced a queue item pairing the old climb uuid with the new angle, so the card showed the wrong angle.

Fix: treat an angle change as a fresh climb identity — clear `savedClimb`, clear any duplicate banner, and mint a new preview uuid when `board.angle` changes (skipped in edit mode, whose identity comes from `editClimbUuid`). This matches the existing model, where the autosave draft key already includes angle (`${boardName}:${layoutId}:${sizeId}:${angle}`).

## 2. Controller hook tests (gap)
The 355-line controller had zero tests. Added `__tests__/use-create-climb-screen.test.tsx` (9 tests) covering: the angle regression above, Set Active uuid lifecycle (stable preview → saved uuid → fresh after clear), unauthenticated routing, open-settings-on-unnamed-save, the save state machine through `justSaved`, update-in-place vs. create, duplicate-inline vs. toast, autosave debounce, and draft restore. The angle test is a true regression guard — it fails (`expected 'c-25' not to be 'c-25'`) without the fix.

## 3. Fragile early-return in `CreateClimbSettingsSheet` (hardening)
`if (!visible) return null` ran after the effect that drives `present()`/`dismiss()`. If `visible` ever flipped to `false` programmatically (e.g. an auto-close after save), the component unmounted before the dismiss effect fired (`sheetRef` already null), so the close animation never played. The `BottomSheetModal` is a true modal (invisible until `present()`), so the sheet now stays mounted and is controlled purely via `present()`/`dismiss()`.

## 4. Stale JSDoc on `CreateClimbScreen` — no change needed
The current file already has a single, accurate JSDoc block describing the routing-aware screen; there was no stale "non-routing single-screen" block to remove.

## Tooling note
The new controller test is the first mobile test to render a shared `@boardsesh/*-react` hook, which surfaced a dual-React resolution between the mobile-pinned `react` and the workspace-root `react`. Added `resolve.dedupe: ['react', 'react-dom']` to `packages/mobile/vite.config.ts`.

## Validation
- `vp run typecheck:mobile` ✓
- `vp test run --project mobile` — 648 tests pass (72 files) ✓
- `vp run check:mobile-bundle` ✓
- Regression test confirmed to fail without the fix, pass with it ✓

https://claude.ai/code/session_01Wn7KCZQNX6nNyUNNDvt6zX

---
_Generated by [Claude Code](https://claude.ai/code/session_01Wn7KCZQNX6nNyUNNDvt6zX)_

### DIFF
--- a/packages/mobile/src/components/create-climb/CreateClimbSettingsSheet.tsx
+++ b/packages/mobile/src/components/create-climb/CreateClimbSettingsSheet.tsx
@@ -77,8 +77,9 @@ export function CreateClimbSettingsSheet({
     [systemColors],
   );
 
-  if (!visible) return null;
-
+  // Keep the modal mounted and drive it purely via present()/dismiss() in the
+  // effect above. Returning null on !visible would unmount the sheet before the
+  // dismiss effect runs (sheetRef goes null), so the close animation never plays.
   return (
     <ModalSheet ref={sheetRef} snapPoints={['72%']} onDismiss={onDismiss} scrollable>
       <View style={styles.body}>
--- a/packages/mobile/src/components/create-climb/__tests__/use-create-climb-screen.test.tsx
+++ b/packages/mobile/src/components/create-climb/__tests__/use-create-climb-screen.test.tsx
@@ -1,0 +1,317 @@
+// @vitest-environment jsdom
+import { describe, it, expect, vi, beforeEach, afterEach } from 'vitest';
+import { renderHook, act } from '@testing-library/react';
+import { GraphQLOperationError } from '@boardsesh/graphql-client';
+import type { ClimbQueueItem } from '@boardsesh/queue';
+import { useCreateClimbScreen, type CreateClimbBoard } from '../use-create-climb-screen';
+
+// All platform I/O the controller reaches for is mocked here. The shared hold
+// machine (`useCreateClimb`), `climbToQueueItem`, and `isDuplicateClimbError`
+// stay REAL so the orchestration — Set Active uuid/angle, the save state
+// machine, autosave, duplicate handling, draft restore — is exercised end-to-end.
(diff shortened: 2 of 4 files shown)


marcodejongh:
Superseded by the create-climb rebuild on main. This draft was branched off `990fcc20` (original create-climb from #2510); since then `b05a3cc6` ("rebuild create-climb as a play-drawer-style sheet") **retired `CreateClimbSettingsSheet.tsx`** (the file concern #3 hardens, now deleted) and rewrote `use-create-climb-screen.ts` (107 lines diverged), and `fb1e7867` ("six create-climb draft/autosave bugs") reworked the autosave/save state machine. Concern #2's controller test was written against the old hook shape, and concern #4 was a no-op. Three of the four items are obsolete, so a mechanical rebase is not possible.

The one item still live on current main — the `savedClimb`/working-angle Set-Active mismatch (`handleSetActive` pairs a stale `savedClimb.uuid` with the live `board.angle`; no angle-change effect clears it; route keys by `editClimbUuid` not angle) — is captured for a fresh implementation in https://github.com/boardsesh/boardsesh/issues/2654.

Closing as superseded; branch `claude/moonboard-create-screen-fixes-co3TY` preserved.

Issue #2527, opened 2026-06-04
Mobile: resolve partially-synced queue items instead of falling back to "Unknown Climb"


marcodejongh:
## Background

`ClimbQueueItem.climb` is typed non-nullable (`@boardsesh/queue` `ClimbQueueItem.climb: Climb`), but at the wire boundary a queue item can arrive from a party peer **partially synced** — with the climb data missing or unresolved. Several call sites already defend against this defensively (`item.climb?.name`, `item.climb?.uuid`), and the queue row falls back to the `mobile.queue.unknownClimb` ("Unknown Climb") label.

## What this issue tracks

In the queue-feedback-fixes PR we added **null guards** so a climb-less item can't crash the app:

- `packages/mobile/src/lib/climb-to-queue-item.ts` — `toClimbInput()` now tolerates a nullish climb (returns a placeholder `ClimbInput`) instead of dereferencing `null` inside the queue mutation mapper.
- `packages/mobile/src/components/ClimbListItemContent.tsx` — the visual now accepts a nullish climb and renders an "Unknown Climb" placeholder instead of crashing on `climb.difficulty`.

Those are band-aids: they stop the crash but the user still sees a placeholder row, and a placeholder `ClimbInput` may get echoed to peers.

## Proper fix (followup)

Resolve the "no matching" / unknown climb properly rather than showing a placeholder:

- When a peer broadcast delivers a queue item whose climb isn't locally resolvable, **fetch the climb** (by uuid, at the active angle) and patch it into the queue item — same shape as the existing self-healing re-grade path in `queue-provider.tsx`.
- Decide what to send over the wire for a still-unresolved item: skip it from the sync payload rather than emitting a placeholder `ClimbInput` with an empty uuid.
- Add coverage for the resolve/reconcile path.

## Acceptance

- Partially-synced peer queue items resolve to their real climb (thumbnail, name, grade) once data is available.
- No placeholder `ClimbInput` is broadcast to peers.
- "Unknown Climb" only ever appears transiently (or not at all) for genuinely unresolvable climbs.

PR #2528, opened 2026-06-04
fix(mobile): null-guard queue climb + test drag hook (review feedback)

marcodejongh:
Addresses review feedback on the queue-list PR (`rn_queue_broken_again`). Targets that branch so the diff is just these fixes.

## Changes

1. **`toClimbInput` null guard** (`climb-to-queue-item.ts`) — widened to `Climb | null | undefined`; a nullish climb returns a type-valid `PLACEHOLDER_CLIMB_INPUT` (+ dev-warn) instead of throwing on `null.uuid`, so a partially-synced peer item reaching `addToQueue → mutations` can't crash.
2. **Null passed to non-nullable prop** (`ClimbListItemContent.tsx`) — `climb` prop is now nullable; renders an "Unknown Climb" placeholder (existing `session.mobile.queue.unknownClimb` key) keeping the three-block layout so the gutter/separator stay aligned. Subtitle `useMemo` made null-safe; the search-list caller is unaffected.
3. **Queue count badge** — left removed (intentional per the #2490 redesign); the count still rides in the queue button's accessibility label.
4. **Test gap** — new `use-queue-drag.test.ts` (6 tests) covering the hook's JS-thread wiring the pure `queue-drag-math` tests can't reach: gesture-lifecycle → `commit` → `reorderQueue`, no-op (lands on start), upward-drag clamping, `isDragging` toggle, `onRowHeight` drop-step, and selection haptics. Reanimated/gesture-handler mocked via `vi.hoisted`.
5. **`handleRemove` dropped from pan-gesture deps** — verified, not a bug. `handleRemove` is only called from `handleDeletePress` (plain JS callback), never inside the pan gesture's worklets, so removing it from the deps is correct.

## Followup
The null guards in 1 & 2 stop the crash but still show a placeholder. Proper resolution (fetch + patch the missing climb, or skip it from the sync payload) is tracked in #2527.

## Validation
- `vp run typecheck:mobile` ✓
- `vp test --project mobile` — 686 passed (incl. 6 new) ✓
- `vp run check:mobile-bundle` — OK (8.8 MB) ✓


marcodejongh:
Rebased onto \`main\` @b3e3ec8e6 (was stacked on \`rn_queue_broken_again\`, whose work has since landed in main). The single commit replayed cleanly — no content conflicts.

Two mechanical fixes on top:
- **toClimbInput**: log the missing-climb warning unconditionally instead of under \`__DEV__\` (addresses the Claude-review finding that the empty-uuid placeholder reaches a real mutation in production, where the warning was previously suppressed). See #2527.
- **use-queue-drag.test.ts**: the hook now returns \`{ isDragging, controls }\` after the queue-item memoization merge (#2613); retargeted \`makeHandleGesture\`/\`onRowHeight\` from the old flat shape to \`controls.*\`. The 6 gesture-lifecycle tests coexist with main's identity-stability \`use-queue-drag.test.tsx\`.

Validation: \`vp run typecheck:mobile\` ✓ · \`vp run test:mobile\` 1253 passed (165 files) ✓ · \`vp run check:mobile-bundle\` OK (10.5 MB) ✓ · pre-commit \`vp check --fix\` ✓.

Note: PR base still shows \`rn_queue_broken_again\` (that branch lags main, so the diff looks large); retarget the base to \`main\` to see just these two commits.

PR #2547, opened 2026-06-05
Rn session parity shared abstractions

marcodejongh:


packages/mobile/src/providers/queue-provider.tsx
```diff
@@ -622,6 +700,128 @@ export function QueueProvider({ children }: { children: ReactNode }) {
   // shared mutation already swallows transport errors and is a true no-op in
   // solo (never lazily creates a session), so callers can fire it freely.
   const setSessionBoardPath = useCallback((boardPath: string) => mutations.setSessionBoardPath(boardPath), [mutations]);
+  const restoreQueueSnapshot = useCallback(
```
marcodejongh, line 709: Seems like we added quite a lot of logic in here, that looks like logic that would also exist in the web app? Lets make sure these changes belong here and not in the shared package

packages/shared/play-view/src/wall-confirm-fallback.ts
```diff
@@ -0,0 +1,119 @@
+export const WALL_CONFIRM_TIMEOUT_MS = 2000;
+
```
marcodejongh, line 11: Lets add a comment to the top of this file explaining what it does.
Also why did we need to add this? Did we move this from web to shared? OR was it completely new?

packages/web/app/components/board-bluetooth-control/wall-confirm-bus.ts
```diff
@@ -1,54 +1,3 @@
...
-  for (const listener of listeners) {
-    try {
-      listener(climbUuid);
-    } catch (error) {
-      // A buggy listener shouldn't break the others.
-      console.error('wall-confirm listener threw:', error);
-    }
-  }
-}
+export { emitWallConfirm, subscribeToWallConfirm } from '@boardsesh/play-view';
```
marcodejongh: LEts not re-export, lets just update the imports in other places of the app

PR #2552, opened 2026-06-06
[codex] Fix play drawer header grade spacing

marcodejongh:
## What changed

- Clamp `DrawerHeader` trailing measurements before mirroring them into the leading spacer.
- Keep the trailing slot's own minimum width at the configured floor so measured widths can grow and shrink without ratcheting stale spacing.
- Let `PlayDrawerHeader` seed the trailing slot width from the displayed grade label so wider grades reserve enough space on first mount.
- Use the softened grade foreground color for better contrast on the play drawer header.
- Add focused mobile coverage for width clamping, mounted grow/shrink behavior, stale minimum changes, initial grade width, and grade color resolution.

## Why

The play drawer title could shift after the first layout event when the rendered grade label was wider than the default spacer. The spacer now starts from a grade-aware width, measured updates keep the same minimum floor defensively, and the trailing slot no longer forces its old measured width back into future measurements.

## Validation

- `vp test run --config vite.config.ts src/components/play-drawer/__tests__/play-drawer-header.test.ts`
- `vp fmt packages/mobile/src/components/DrawerHeader.tsx packages/mobile/src/components/play-drawer/PlayDrawerHeader.tsx packages/mobile/src/components/play-drawer/__tests__/play-drawer-header.test.ts --check`
- `vp run typecheck:mobile`

Note: full `vp check` and `vp lint` are currently blocked by unrelated pre-existing repo-wide formatting/lint findings outside this PR's changed files.

### DIFF
--- a/packages/mobile/src/components/DrawerHeader.tsx
+++ b/packages/mobile/src/components/DrawerHeader.tsx
@@ -1,10 +1,14 @@
-import { memo, useCallback, useState, type ReactNode } from 'react';
+import { memo, useCallback, useEffect, useState, type ReactNode } from 'react';
 import { View, StyleSheet, type LayoutChangeEvent } from 'react-native';
 import { spacing } from '../theme/tokens';
 
 // Default width of the trailing slot (and the matching leading spacer) so the
 // centered column stays optically centered before the trailing element measures.
 const DEFAULT_TRAILING_MIN_WIDTH: number = spacing[12];
+
+export function resolveDrawerHeaderTrailingWidth(layoutWidth: number, minWidth: number): number {
+  return Math.max(minWidth, Math.ceil(layoutWidth));
+}
 
 type DrawerHeaderProps = {
   /** Centered column content (e.g. title + subtitle, or a name input + counts). */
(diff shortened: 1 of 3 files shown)


marcodejongh:
Closing this — both files it touches were rewritten since, and two of the three changes no longer do anything useful against the current code.

**The width clamp is a no-op.** `resolveDrawerHeaderTrailingWidth` returns `Math.max(minWidth, Math.ceil(layoutWidth))`. On `main`, `DrawerHeader` already floors the mirrored width in both branches: the no-leading branch measures a `View` that carries `minWidth: trailingMinWidth` (line 81), so the reported layout width can never come in below the floor, and the leading branch computes `sideWidth = Math.max(leadingWidth, trailingWidth, trailingMinWidth)` (line 59). `DrawerHeader` also grew a `leading` slot for the on-wall status since this PR was opened, so the stale-minimum bookkeeping this PR adds is against a component shape that no longer exists.

**The grade-colour swap now collides with a prop.** `PlayDrawerHeader` on `main` takes an explicit `gradeColor` prop:

> *Explicit grade colour, overriding the internal `getGradeColor` lookup. The play drawer passes this so the colour matches the shown grade when the "Show Boardsesh grades" toggle swaps the label to the Boardsesh grade.*

It resolves as `gradeColor ?? getGradeColor(rawDifficulty ?? difficulty) ?? DEFAULT_GRADE_COLOR`. Replacing that with an unconditional `getSoftGradeColor(rawDifficulty ?? displayGrade, darkMode)` would throw the caller's colour away and re-break the Boardsesh-grade toggle.

**What is genuinely unique** is the grade-aware initial trailing width — `trailingMinWidth={estimateInitialGradeColumnWidth(difficulty)}` — which reserves space for a long label (`6c+/V5`) before the first layout pass. That buys one frame of stability on mount and nothing after. Not worth a 3151-commit-behind, `CONFLICTING` branch; easy to redo as a two-line change if the mount-frame shift is still visible.

No linked issue on this one, so nothing to reopen.

PR #2560, opened 2026-06-06
Review fixes for the queue FAB + search redesign

marcodejongh:
Remediation from a multi-reviewer pass (3 code reviewers + a 2-person design team) on the native-tab queue accessory + climbs search redesign (originally #2554, now merged to main). **Rebased on latest `main`**, so this PR's diff is just the review fixes.

## Commits

- **`f0f206b` Phase A — ship-blockers**
  - **react-native-screens patch apply-guard** in CI (`scripts/mobile-patches-check.ts` + test, run after `check:mobile-native-deps`). A Bun patch is keyed by exact version, so an Expo/RN bump silently drops it and the iOS 26 tab-bar minimize regresses with nothing catching it.
  - **Dynamic Type cap** on the fixed-height glass chrome so the climb name/grade don't clip at large accessibility text.
  - **Suggestion-aware queue carousel** — `canNext/canPrevious` from `computeNavigationStateWithSuggestions`, so a playlist no longer dead-ends at the queue tail (+ tail-peek tests).
  - **SearchHeader silent setter** — programmatic seeding (board restore / sync / recent pill) no longer re-arms the input debounce or re-commits the term.
  - **Grade as a real `accessibilityAction`** on the filter button (VoiceOver / Switch Control can't reach a raw long-press) + hint copy fixed across locales.
- **`c617d72` Phase B — tick parity** — dropped the Climbs-tab-only gate so the JS fallback shows the log-ascent tick on every tab, matching the always-on iOS 26 accessory (+ test).
- **`18b3426` Phase D — reuse + testability** — shared `countActiveFilters` (grade-inclusive) in `@boardsesh/climb-filters` (+ tests); extracted the pure `computeBottomChromeMetrics` from the hook (+ test), restoring the coverage lost when `queueSnackbarBottomOffset` was inlined.
- **`ab5181b` Fast grade rail** — removed the redundant Done button from the filter-button grade rail (it has an accessible drag handle, tap-outside, and auto-dismiss). Found in device testing.

## Validation (Linux)
`typecheck:mobile`, `vp test --project mobile`, `climb-filters` + i18n catalog-parity + patch-guard tests, and `check:mobile-bundle` (iOS + Android) — all green. Verified live on a device via Metro (app bundles + runs; tabs/search/grade exercised).

## Deferred to a device-paired follow-up (not in this PR)
These need an on-device build to get right and were intentionally not done blind:
- Visible "grade" cue / active band on the filter FAB (the a11y action is in; the visible cue is next).
- Record-tab dedup of the floating accessory vs the inline session; live-session-vs-Bluetooth badge disambiguation; return-to-session affordance.
- Native swizzle hardening (gate/bound the `contentScrollViewForEdge` swizzle, idempotent vs double `+load`); the `useBottomChromeMetrics` re-render fan-out; the shared `useQueueClimbCarousel` extraction.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -486,6 +486,13 @@ jobs:
       # push-to-main. Runs before the heavier Metro bundle so it fails fast.
       - name: Check native build-phase deps resolve
         run: vp run check:mobile-native-deps
+      # A Bun patch is keyed by an exact version, so an Expo/RN bump can silently
+      # drop it: install only warns, and typecheck + the Metro bundle stay green
+      # while a native-only behavior (e.g. the iOS 26 tab-bar minimize tracking
+      # the climbs FlashList) regresses all the way into TestFlight. Assert the
+      # patched source is actually present on this cheap Linux runner.
+      - name: Check native patches are applied
+        run: vp run check:mobile-patches
       - name: Download built packages
         uses: actions/download-artifact@v4
         with:
--- a/packages/mobile/app/(tabs)/_layout.tsx
+++ b/packages/mobile/app/(tabs)/_layout.tsx
@@ -26,6 +26,10 @@ export default function TabLayout() {
   const showRecordBadge = isBluetoothConnected || sessionId !== null;
 
   return (
+    // `minimizeBehavior="onScrollDown"` relies on UIKit finding the climbs
+    // FlashList's nested scroll view, which it can't by default — that fallback
+    // lives in patches/react-native-screens@4.25.2.patch. `vp run check:mobile-patches`
+    // (CI) fails the build if that patch ever stops applying after a dep bump.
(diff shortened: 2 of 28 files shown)


marcodejongh:
Follow-up issues filed for the deferred work (see the "Deferred" section above):

- #2561 — Make grade discoverable on the climbs filter button (visible band + cue)
- #2562 — Record tab: de-duplicate current-climb chrome over the inline session
- #2563 — Disambiguate the Record live-session badge + add return-to-session
- #2564 — Harden the react-native-screens `contentScrollViewForEdge` swizzle (+ capability gate)
- #2565 — Stop the `useBottomChromeMetrics` re-render fan-out during live sessions
- #2566 — Extract a shared queue-climb carousel hook
- #2567 — Remaining polish, tokens, and on-device QA

PR #2574, opened 2026-06-07
Redesign the mobile climbs search & filter experience (iOS HIG)

marcodejongh:
## What

Reworks the mobile Climbs search & filter experience around one principle from the design exploration: **the Climbs tab is the search surface**, and filtering — not name search — is the dominant interaction.

## Why (data-driven)

- **Board switching is rare** — PostHog (web/Capacitor proxy): ~92% of users use one board in 30 days, 96.6% of sessions resume on the same board. Board selection shouldn't own the primary surface.
- **Filters aren't ubiquitous, but grade dominates** — ~45% of Capacitor-mobile browse sessions apply a filter (~40% browse unfiltered, ~15% name-only); of *filtered* searches, **99.9% set a grade range**. So: keep name search cheap, keep the native search, don't bury search behind filters.

(Mobile RN analytics isn't live yet; the legacy Capacitor app is a webview, so its real mobile usage lands as web events — that's the data source.)

## Changes

- **Land on Climbs** via `unstable_settings.initialRouteName` (both variants).
- **Boards tab → modal**: board selection moves to a modal opened from the board capsule + a no-board CTA; all 9 nav entry points repointed; post-activation uses `router.dismissTo` back onto Climbs. (BLE-serial auto-detect integration point marked.)
- **Active filters → a single summary capsule** below the board name (tap to clear, no per-chip ✕), single-scale grades matching the rows.
- **Scrim** behind the top chrome so the list doesn't bleed through the floating glass islands.
- **Row scan**: louder grade (title3) + a send-status dot on the text line (shared `useAscentStatus` hook).
- **Demoted board capsule**; **white active filter-FAB glyph** on the maroon glass.
- Material variant gets a filter button left of the search field; extracted `filter-tokens` / `filter-labels`.

## Validation

- `vp run typecheck:mobile` ✓
- `vp test --project mobile` — **962 pass** ✓ (new `filter-tokens` test; updated `tab-layout`, `analytics-screen-path`, `filter-button`, `climb-top-chrome`)
- `vp run check:mobile-bundle` ✓

All changes are OTA-safe (no native deps added). **Needs an on-device check on iOS 26 Liquid Glass** — NativeTabs re-layout after removing the Boards trigger, the boards-modal dismissal, and the summary capsule.

## Tried and dropped

- `hideWhenScrolling` to hide the bottom search on scroll — iOS 26 ignores it for the tab-bar-relocated search (confirmed on device); reverted.
- Merging name search into a combined filter drawer — dropped after the data showed >half of sessions browse unfiltered or name-only, so it'd add friction.

### DIFF
--- a/packages/mobile/app/(tabs)/__tests__/tab-layout.test.tsx
+++ b/packages/mobile/app/(tabs)/__tests__/tab-layout.test.tsx
@@ -52,12 +52,16 @@ vi.mock('expo-router/unstable-native-tabs', () => {
   return { NativeTabs };
 });
 
-import TabLayout from '../_layout';
+import TabLayout, { unstable_settings } from '../_layout';
 
 describe('TabLayout', () => {
   beforeEach(() => {
     cfg.bluetoothConnected = false;
     cfg.sessionId = null;
+  });
+
+  it('lands on the climbs tab by default', () => {
+    expect(unstable_settings.initialRouteName).toBe('climbs');
   });
 
   it('does not render the Record badge when no status is active', () => {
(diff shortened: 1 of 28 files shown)


marcodejongh:
Consolidating into a single PR — folded into #2575 (now based on main, includes both the climbs-search chrome and the Discover/playlist work).

PR #2586, opened 2026-06-08
Recommended climbs engine (board-size aware)

marcodejongh:
## What

Auto-generated **recommendation playlists** so climbers can find great new climbs for *their* board — today good climbs get lost in a 573k-climb catalog. Recommendations prioritise the **biggest board size** the user owns.

Surfaced as new cards in the existing discover/smart-playlist grid (per-user) plus public per-board cohort playlists in the Discover scroll.

## How it ranks

Within climbs that fit the user's biggest board (soft match — full-board use ranked higher via a height-first `fullness` factor):

`Aurora ascensionist_count × quality_average × fullness`, boosted by **setter popularity** and **community sends**.

Four variants:
- **Crowd Favorites** — proven classics you haven't sent
- **Hidden Gems** — highly rated, low-ascent (fights the noise)
- **At Your Level** — banded **V−3..V+1** around your *max* send grade across all boards, with **MoonBoard de-sandbagged +2 V** (MoonBoard V4 ≈ Kilter V6)
- **Fresh** — recently set, ranked primarily by **setter popularity** (new climbs have no ascents yet)

## Phases

1. **Per-user recommendations** — ships on existing Aurora data, no PostHog dependency. New `SmartPlaylistType` values + resolver dispatch to a catalog-based selector; board target resolved registered → inferred (size/angle overridable). Pure logic (grade conversion, size fullness) in `@boardsesh/board-constants` (unit-tested); query builders in `@boardsesh/db`.
2. **Public cohort playlists** — `generated_recommendation` marker on `playlists`; system-user-owned "Crowd Favorites · Kilter Homewall 10x12 @ 40°"-style playlists.
3. **Trending signal** — `board_setter_stats` + `board_climb_send_stats` tables; nightly GH Actions job recomputes setter popularity, mines PostHog `Climb Sent to Board Success` (40k events/90d) into per-climb send stats, and regenerates cohorts. No raw per-user send capture (out of scope).

## Data grounding (read-only prod + PostHog)

- Sends outnumber ticks ~6.8× (events) / ~2× (users) → the trending signal is dense; ticks too sparse to rank (6% catalog coverage).
- For a Kilter 10x12 @ 40° owner: ~251 solidly-great + ~607 hidden-gem candidates.
- Setter popularity well-distributed (KilterStudio, jwebxl, etc.).

## Validation

- 46 unit tests (grade conversion, MoonBoard offset, size fullness, i18n parity).
- Scoring SQL smoke-tested against the dev catalog (all 4 variants return healthy pools).
- Per-user path verified end-to-end for a real dev user (board resolved to their registered 10x12; At-Level = 32 personalized climbs).
- Cohort job run on dev: 27 distinct public playlists, idempotent.
- `vp run typecheck:{web,backend,db}` + `vp check` green.

## Follow-ups (not in this PR)

- Mobile rec presets (backend + shared hook already support the new types).
- Dynamic sitemap entries for cohort playlists (currently crawlable via the Discover scroll).
- `DATABASE_URL` for the nightly job must be a **writable** connection; `POSTHOG_PERSONAL_API_KEY` optional (send-stats skipped without it).


marcodejongh:
Thanks — all review findings addressed in `8ddea76e`:

**Security**
- `smartPlaylist`: recommendation types are now owner-private. A non-owner/anonymous caller gets a **uniform empty result** rather than a throw — this prevents inferring another user's grade/board *and* keeps SSR of a shared/guessed URL from 500ing (the route is the public smart-playlist page). Logbook smart playlists stay public/shareable as before.
- `POSTHOG_PROJECT_ID` → `${{ secrets.POSTHOG_PROJECT_ID }}`; the script no longer hardcodes a default and skips cleanly if unset.

**Correctness**
- Cohort climb swap (`delete` + re-`insert`) wrapped in a transaction — no empty playlist on a mid-swap crash.
- `RECOMMENDED_AT_LEVEL` now throws on a missing `gradeBand` instead of silently returning zero results.

**Performance**
- Added a **GIN index** on `board_climbs.compatible_size_ids` (migration `0119`) and switched the filter to `compatible_size_ids @> ARRAY[sizeId]`. Verified via `EXPLAIN`: planner does a BitmapAnd of the GIN index + the search-filter index — no seq scan on ~573k rows.
- `refreshSendStats`: chunked bulk upsert (one round-trip per ~500 climbs) instead of one per climb.
- `resolveInferredTarget`: angle + layout queries now run in parallel.

**Code quality**
- Extracted the duplicated `rowsOf()` into `@boardsesh/db` queries util; all three call sites import it.

marcodejongh:
Addressed the codex review in `e7e77dcc7`:

**Highest priority**
1. **Selected board reaches recommendations** — `boardUuid` is now threaded through the GraphQL input + `useSmartPlaylist`; `resolveRecommendationBoardTarget` resolves that specific *owned* board (validated) before falling back to biggest → inferred. Selecting a Tension/smaller board no longer returns the biggest board's climbs.
2. **Angle-correct hydration** — `hydrateClimbsByRefs` joins `board_climb_stats` at the per-ref angle (`COALESCE(override, most-ascended)`), so a 45° rec shows 45° difficulty/quality/ascents/benchmark, not 40°.
3. **Private recommendations** — the resolver short-circuits **before** any user lookup (no existence probing) and returns a uniform empty result for non-owners; the detail page hides the share button for `RECOMMENDED_*`.
4. **Backend tests** — mocked the rec helpers; added owner/non-owner, no-board, angle-hydration, and rec-count coverage. 18 pass (was 5 failing).
5. **Send-stat expiry** — full rebuild each run in a transaction; rows that leave the 90-day window are dropped, an empty *successful* result clears the table, a failed query leaves data intact.
6. **Empty-playlist guard** — a cohort with zero refs is skipped (previous playlist kept), never published/emptied.

**Other**
7. Consensus-grade fallback — ungraded sends use `ROUND(display_difficulty)` at the tick angle so AT_LEVEL isn't hidden/under-counted.
8. **Duplicate GIN index removed** — `0073` already indexes `compatible_size_ids`; the size filter now uses `@> ARRAY[sizeId]` to use it (EXPLAIN: BitmapAnd, no seq scan). Migration `0119` deleted.
9. Deterministic board-target tie-breakers (registered: order by id; inferred: `ORDER BY count, board_type/angle/layout_id`).
10. Recommendation pagination offset capped.
11. `POSTHOG_PROJECT_ID` defaults in the script (non-secret) rather than a workflow literal; send-stats run whenever the API key is present (no silent skip on a missing id).
12. Fresh `published_at` cast guarded by an ISO-date regex so one malformed value can't break the query/refresh.
13. Ambiguous cross-board UUIDs skipped in the send-stats refresh.

**Product decisions**
- "New climbs" = not already **sent** (excludes flash/send), as intended; attempted projects stay eligible.
- Fresh ranks by setter popularity first, then recency — intentional per the feature ask ("weight popular setters more heavily in Fresh").

Remaining suggested test gaps (DB-backed multi-angle/selection/compat tests, refresh-job tests) are good follow-ups; the new unit coverage plus DB smoke-tests cover the core paths for now.

marcodejongh:
**CI note:** the only red checks are **E2E (shard 8)** and **Capture PR screenshots**, both pre-existing flakes on `main` — not introduced by this PR.

- Shard 8 fails on `play-view-swipe-close.spec.ts` (drawer swipe-gesture timing) and `queue-persistence.spec.ts` (queue-bar restore). Both are in the play-drawer/queue area, which this PR doesn't touch.
- `main`'s own latest E2E runs are red on the **same** `queue-persistence.spec.ts` shard-8 failure (e.g. run 27121108556), and the last few `main` E2E runs are all `failure`.
- The one shared code path (`hydrateClimbsByRefs`) is a strict no-op when no angle override is passed — exactly how the queue path calls it — so it can't change queue behavior.

Everything this PR affects is green: typecheck, lint, i18n, codegen-drift, backend/default/ocr tests, mobile-bundle, prebuild-and-build, and 7/8 E2E shards.

PR #2592, opened 2026-06-08
Mobile: replace ascent-status colour dot with a monochrome glyph

marcodejongh:
## What

The recent climbs redesign turned the per-climb **ascent status** (flash / send / attempt) into a bare **8×8 colour dot** sitting left of the grade. Users found it unclear. A quick design review surfaced three concrete failures:

1. **A dot carries no inherent meaning** — green/yellow/orange don't say "sent / flashed / tried".
2. **It collides with the grade.** The grade text right beside it is *already* colour-coded on the difficulty ramp (yellow → orange → red → purple). The dot's hues come from the same families, so a yellow "flash" dot reads like a grade colour — two colour systems fighting in one 6px zone.
3. **It fails colour-blind users** — green (send) vs orange (attempt) vs yellow (flash) is exactly the red-green trio, with no shape to fall back on.

## The change

Carry the status by **glyph shape in a single neutral grey** instead of colour. The grade keeps the colour channel, the status keeps the shape channel — so the clash disappears, dark mode is automatic (the tint is the adaptive `secondaryLabel`), and it's the strongest accessibility option.

- **⚡ bolt** = flashed · **✓ circled check** = sent · **✗ circled x** = attempted
- A VoiceOver/TalkBack `accessibilityLabel` ("Flashed" / "Sent" / "Tried") now names the status, since colour no longer does.

The swap lives entirely in the shared `ClimbListItemContent`, so the **climbs search list, queue rows, play-drawer suggestions, and in-session view** all inherit it.

### Files
- `ClimbListItemContent.tsx` — dot → monochrome `<Icon>` tinted `secondaryLabel`, wrapped with an `accessibilityLabel`; dropped the now-unused `iosSystemColors` import.
- `icon-map.ts` — added `ascent.attempt` (iOS `xmark.circle` / Android `close-circle-outline`) to pair visually with the circled check used for `send`.
- `mobile.climbRow.ascentStatus` labels added to en-US, es, fr. ⚠️ **es/fr values are reasonable climbing terms but should get a native-speaker review.**
- Refreshed four stale comments that pointed at the no-longer-rendered `AscentStatusBadge`.

## Design context

This was the user's pick from four options a design panel mocked up (icon+colour / monochrome icon / text label / icon+label). Monochrome icon was chosen for the cleanest grade/status separation and best accessibility; the trade-off accepted is losing the "green = good" colour cue. The web app keeps coloured icons — mobile deliberately diverges here.

## Testing

- `vp run typecheck:mobile` — clean for all changed files. (2 unrelated errors in this sandbox come from `expo-share-intent` not being installed; CI installs it.)
- `vp test --project mobile` — **1112 tests pass**, including the i18n catalog-completeness parity test for the new keys. (Only failing suite is the same `expo-share-intent` sandbox gap.)
- Pre-commit `vp run check:i18n:orphans` passed (new keys detected as used).
- ⚠️ **Visual QA still needed** (no simulator on Linux): exercise a flashed / sent / attempted climb across the search list, a queue, play-drawer suggestions, and in-session, in **both light and dark**. Specifically eyeball the **attempt** glyph at 16px — if `xmark.circle` reads off, the bare `close` (xmark) is a one-line swap.


marcodejongh:
### Added: climb-row intrinsic-attribute icons (no-match + benchmark ©)

Follow-up commit on this branch (`45d2d3680`). The climbs redesign had dropped the **"no matching"** indicator; this brings it back — and, while building the slot, restores the **benchmark/classic ©** icon web already shows.

**Design call:** rather than crowd these next to the ascent glyph + grade (the space concern), they follow web's IA — *intrinsic climb attributes group inline with the name*, *personal ascent status stays on the right with the grade*. The two clusters sit at opposite ends of the row and never compete. Renders `© ⊘` right after the name (web's order), monochrome grey, only when applicable; the **name truncates first** so the glyphs never clip.

- `ClimbListItemContent.tsx` — nested name row + `is_no_match`/`benchmark_difficulty` on the climb shape
- `icon-map.ts` — `no.match` (`hand.raised.slash` / `hand-back-right-off-outline`), `benchmark` (`c.circle` / `copyright`)
- `climb-to-queue-item.ts` — preserve `is_no_match` so queue rows show it
- i18n `mobile.climbRow.noMatch` / `.benchmark` (en/es/fr; noMatch reuses existing create-form wording)

**Validation:** `typecheck:mobile` clean · `vp test --project mobile` 1123 pass · `check:mobile-bundle` iOS + Android OK.

**Visual QA still needed:** a no-match climb (description starts "No match"), a benchmark climb (`benchmark_difficulty > 0`), and one that's both-and-sent — confirm `© ⊘` after the name + ✓ by the grade with no crowding, long names truncate first, light/dark, and queue rows.

PR #2604, opened 2026-06-08
perf(mobile): back grade-format hook with useSyncExternalStore

marcodejongh:
## What

`useGradeFormat()` is rendered **once per row** in `ClimbListItemContent` (the climbs-list / queue / session row visual, `packages/mobile/src/components/ClimbListItemContent.tsx:65`). It delegated to `useGradeDisplayFormatPreference()` in `packages/mobile/src/lib/grade-format-preference.ts`, which on **every** caller added:

- two `useState` cells (`gradeFormat`, `loaded`), and
- a `useEffect` whose body ran `listeners.add(listener)` on mount and `listeners.delete(listener)` on unmount against the module-level `Set`.

In a virtualized list, rows mount and unmount as you scroll. So every visible-and-recycled row ran a `Set.add` / `Set.delete` on each scroll-in / scroll-out, plus carried the extra hook slots and an extra re-render to sync local state from the module value. That's pure churn in the scroll hot path, and it scales with the number of rendered rows.

## Fix

Convert the module store to back the hook with **`useSyncExternalStore`**:

- `subscribe(onStoreChange)` registers React's store-change callback against the existing module `Set` and returns the unsubscribe. It still kicks off the one-time AsyncStorage load (`loadGradeDisplayFormat()`) on first mount, gated by the existing `hasLoaded` flag.
- `getSnapshot()` returns the current format (`current`).
- `getServerSnapshot()` returns `DEFAULT_GRADE_DISPLAY_FORMAT`.
- `loaded` is exposed via a second `useSyncExternalStore` sharing the same `subscribe`, returning the `hasLoaded` primitive.

Each consumer is now a single cheap subscription with **no per-instance `useState`/`useEffect`** and no extra render on scroll. The snapshot getters return primitives, so React's referential-equality check is stable — no new object per render, no tearing.

Preserved unchanged:

- the AsyncStorage-backed setting + one-time `loadGradeDisplayFormat()` load,
- the cross-component `notify()` fan-out via the `Set` (switching the setting still updates every visible row live),
- the setter path (`setGradeDisplayFormatPreference`: store update → persist → notify),
- `useGradeFormat()`'s public API (`{ formatGrade, gradeFormat, setGradeFormat, loaded, formatGradeByDifficultyId }`), so all its callers and the displayed grade output are identical.

## Validation

- `vp test run --reporter=agent --project mobile` — 1192 passed (155 files)
- `vp test run --reporter=agent grade-format use-grade-format ClimbListItemContent` — 47 passed (8 files)
- `vp fmt --check` on the changed file — clean
- `vp lint` — no findings on the changed file (baseline warnings/errors unchanged)
- `vp run check:mobile-bundle` — OK (Android bundle builds)

## Risk

Low. Single file, no public API change, no change to grade output. `useSyncExternalStore` is React's intended primitive for exactly this module-store pattern and guards against tearing. The one-time load now fires from `subscribe` (first mount) instead of a `useEffect`, which preserves the `hasLoaded`-gated single-load behavior — covered by the existing `grade-format-preference` tests.

### DIFF
--- a/packages/mobile/src/lib/__tests__/grade-format-preference-hook.test.tsx
+++ b/packages/mobile/src/lib/__tests__/grade-format-preference-hook.test.tsx
@@ -1,0 +1,91 @@
+// @vitest-environment jsdom
+import { describe, it, expect, vi, beforeEach } from 'vitest';
+import { renderHook, act, waitFor } from '@testing-library/react';
+
+// Controllable AsyncStorage so each test starts from a clean store and can count
+// reads. Mirrors the pure-function test's mock shape.
+vi.mock('@react-native-async-storage/async-storage', () => {
+  let storage: Record<string, string> = {};
+  return {
+    default: {
+      getItem: vi.fn(async (key: string) => storage[key] ?? null),
+      setItem: vi.fn(async (key: string, value: string) => {
+        storage[key] = value;
+      }),
+      removeItem: vi.fn(async (key: string) => {
+        delete storage[key];
+      }),
+      __reset: () => {
+        storage = {};
+      },
+      __setRaw: (key: string, value: string) => {
+        storage[key] = value;
+      },
+    },
+  };
+});
+
+type MockAsyncStorage = {
+  getItem: ReturnType<typeof vi.fn>;
+  __reset: () => void;
+  __setRaw: (key: string, value: string) => void;
+};
+
+async function getAsyncStorage(): Promise<MockAsyncStorage> {
+  return (await import('@react-native-async-storage/async-storage')).default as unknown as MockAsyncStorage;
+}
+
+describe('useGradeDisplayFormatPreference', () => {
+  beforeEach(async () => {
+    // Fresh module ⇒ module-level store (current/hasLoaded/loadStarted/listeners)
+    // resets between tests.
+    vi.resetModules();
+    (await getAsyncStorage()).__reset();
+  });
+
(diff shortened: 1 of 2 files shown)


marcodejongh:
Thanks — all four addressed in `aa336be`:

1. **Double load on first render** → collapsed to a **single** `useSyncExternalStore` reading a compound `{ gradeFormat, loaded }` snapshot, and the load is memoized as a **promise singleton** (`ensureGradeFormatLoaded`), so N mounted rows share one load (no per-subscribe trigger).
2. **Double listener per row** → one subscription ⇒ **one listener per consumer**. The snapshot object is cached and rebuilt only in `notify()`, so `getSnapshot` stays referentially stable (no render loop).
3. **Missing hook tests** → added `grade-format-preference-hook.test.tsx`: default-before-load → loaded value; **load fires exactly once across three mounted consumers**; `setGradeFormat` fans out to every consumer.
4. **Render-phase side effect in `subscribe`** → `subscribe` is now pure; the load is triggered from the hook's `useEffect` (StrictMode-safe).

Bonus: `loadGradeDisplayFormat` now guards against **clobbering a value the user changed while the storage read was in flight** (re-checks `hasLoaded` after the await) — surfaced by the new fan-out test.

Validation: `typecheck:mobile`, full mobile suite (1195), lint/format clean, `check:mobile-bundle` OK.

PR #2627, opened 2026-06-08
feat(mobile): board switcher in the Material 3 climbs header

marcodejongh:
## What

Adds a discoverable board switcher to the **Material 3** variant of the Climbs tab header.

Before, the active board sat in the `Appbar.Content` *subtitle* — technically tappable (it opened the `/boards` modal) but with no visible affordance, so it read as static text and users didn't realize they could switch boards. Now the board name is the app-bar title with a trailing **down-caret (⌄)** affordance, and tapping opens the existing `/boards` modal.

The iOS liquid-glass variant is untouched (it keeps its auto-minimizing `BoardPill` capsule — the wrong pattern for M3).

```
┌────────────────────────────────────────────┐
│ Kilter · M · 45°  ⌄          ＋   45°   💡  │   ← Appbar.Header (small)
├────────────────────────────────────────────┤
│ 🔍 Search climbs     V4–V6 ⌄     ☰ filter  │   ← unchanged search row
└────────────────────────────────────────────┘
```

## Design rationale (M3)

The board is the screen's *page context* (switching it re-datasets the whole list), so it belongs in the **headline** with a caret affordance — not as a 4th trailing icon, and not as the iOS capsule. Board switching is rare in usage, so it stays subordinate to search/filter. "Climbs" is dropped from the title (the bottom tab already names the screen), maximizing room for long board labels. Tapping opens the full `/boards` modal rather than an inline `Menu`, because the board is a complex tuple (type · layout · size · set · angle) and `/boards` already handles discovery.

## Changes

- **New** `chrome/BoardSwitcherButton.tsx` — flat M3 counterpart of the liquid-glass `BoardPill`: reads the active board itself, shows label + `chevron-down`, fires `hapticLight()`, returns `null` when there's no board. Borderless (it's the title, not a filter pill); label `onSurface`, caret `onSurfaceVariant`, no brand tint.
- `chrome/index.ts` — barrel export.
- `search/ClimbTopChrome.tsx` — swapped `Appbar.Content` for `<BoardSwitcherButton>` (reuses the already-wired `onOpenBoardDetail`, so `climbs/index.tsx` is unchanged); removed the now-dead `boardLabel`, the `formatActiveBoardLabel`/`tCommon` imports, and the `materialTitle`/`materialSubtitle` styles.
- i18n: new `mobile.search.boardSwitcherHint` a11y hint in en-US/es/fr.
- Tests: new `board-switcher-button.test.tsx`; updated the Material case in `climb-top-chrome.test.tsx`.

## Validation

- `vp run typecheck:mobile` ✓
- `vp lint` (changed files) — 0 warnings, 0 errors ✓
- `vp check` — changed files clean ✓
- `vp run check:mobile-bundle` — OK (10.5 MB) ✓
- i18n catalog-completeness — 28 passing ✓
- `vp test --project mobile` — 1243/1244; the single failure (`use-playlist-activation`) is pre-existing on `main` and unrelated to this change.

Not run on Linux: `mobile:screenshot` / simulator (macOS-only) — worth a glance to confirm a long named-board label truncates cleanly against the three trailing actions (label is `numberOfLines={1}` + `flexShrink`).

### DIFF
--- a/packages/mobile/src/components/chrome/BoardSwitcherButton.tsx
+++ b/packages/mobile/src/components/chrome/BoardSwitcherButton.tsx
@@ -1,0 +1,75 @@
+import { useCallback, useMemo } from 'react';
+import { StyleSheet } from 'react-native';
+import { useTheme } from '../../providers/theme-provider';
+import { useActiveBoard } from '../../lib/graphql/use-active-board';
+import { formatActiveBoardLabel } from '../../lib/boards/active-board-label';
+import { hapticLight } from '../../lib/haptics';
+import { spacing } from '../../theme/tokens';
+import { PressableSurface } from '../PressableSurface';
+import { Text } from '../Text';
+import { Icon } from '../Icon';
+
+type BoardSwitcherButtonProps = {
+  /** What a tap does — open the board switcher (the `/boards` modal). */
+  onPress: () => void;
+  /** VoiceOver / TalkBack hint, e.g. "Opens the board switcher". */
+  accessibilityHint?: string;
+};
+
+/**
+ * The Material variant's board switcher: the active board label ("Kilter • M •
+ * 45°", or "Garage Wall • 40°" for a named board) shown as the app-bar title
+ * with a trailing down-caret affordance, so it reads as a tappable context
+ * switcher rather than the static subtitle it replaces. It is the flat M3
+ * counterpart of the liquid-glass `BoardPill` — reads the active board itself,
+ * renders nothing when none is set, and fires the haptic here while the caller
+ * injects what a tap does. Lives where `Appbar.Content` did, taking `flex: 1`.
+ */
+export function BoardSwitcherButton({ onPress, accessibilityHint }: BoardSwitcherButtonProps) {
+  const { systemColors } = useTheme();
+  const { data: activeBoard } = useActiveBoard();
+
+  const boardLabel = useMemo(() => formatActiveBoardLabel(activeBoard), [activeBoard]);
+
+  const handlePress = useCallback(() => {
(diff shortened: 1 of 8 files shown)


marcodejongh:
Addressed review feedback in `ca9b920cd`:

**Codex P2 — full-height tap target** ✅ `BoardSwitcherButton` now sets `alignSelf: 'stretch'`, so the pressable fills the whole app-bar row instead of just the `title3` label line. (`Appbar.Header` lays children out with `alignItems: 'center'` over a fixed height, which had collapsed the hit area; `alignSelf` overrides that per-child.)

**Brittle test helper** ✅ The `PressableSurface` mock now always passes `data-label` through instead of gating on `•`, so a no-separator label (named board, no angle) can't silently `null` the `button()` lookup into a false pass.

**Missing named-board test** ✅ Added two cases: `Garage Wall • 25°` (named + angle) and `Garage Wall` (named, no angle — also exercises the no-`•` path above).

**Implicit test coupling** ✅ The Material case in `climb-top-chrome.test.tsx` now also asserts the switcher's `data-hint` (`mobile.search.boardSwitcherHint`), pinning the `[data-capsule]` match to `BoardSwitcherButton` rather than any pill.

**Multi-line comment convention** — not changed. There's no such rule in `CLAUDE.md` (only the root file exists), and the doc-comment style matches the templates this was modeled on (`BoardPill`, `GradeFilterControl`). Happy to trim if a maintainer prefers.

Re-validated: `typecheck:mobile` ✓, lint (changed files) 0/0 ✓, `board-switcher-button` + `climb-top-chrome` tests 28 passing ✓.

PR #2660, opened 2026-06-09
feat(board-presence): board-level "now on the wall" — backend, shared, mobile + web

marcodejongh:
## What & why

Pivots collaboration from **sessions** to the **physical board**. The data: sessions are ~96% solo and thinly used (476 ever started one, only 63 ever ended one cleanly), while boards are genuinely shared (~144 boards have multiple Boardsesh users; the "Wall" primitive already has 604 users). The unmet need: when someone else changes the climb on a shared wall, you can't tell what's lit, its grade, who set it, or how to tick it.

This adds a real-time **"now on the wall"** feed keyed on a shared `board_id` (`userBoards.id`, resolved 1:1 from the BLE serial), decoupled from sessions. Sessions are superseded as the *primary* collaboration surface but **not removed**. **Everything is behind the `board-presence` flag and dark until enabled.**

Plan: `~/.claude/plans/we-want-to-pivot-agile-milner.md`.

## The full feature (this PR)

**Backend + data**
- `boardsesh:board:` pubsub domain (per-board monotonic seq + 24h Redis FIFO); resolvers (membership-free **eager** subscribe; **server-derived** sender identity — never client-supplied; catalog `climbUuid` validation; per-op rate limits), gated by env `BOARD_PRESENCE_ENABLED`.
- `boardNowPlaying(boardId)` subscription; `reportBoardClimb` / `resolveBoardForSerial` mutations (find-or-bind, unique serial→one board); `boardRecentClimbs` / `boardPresenceStats` queries.
- Migration `0120`: unique partial index on `user_boards.serial_number`. `saveTick` stamps the connected `board_id` → durable per-wall history from `boardsesh_ticks`.

**Shared (web ↔ mobile reuse)**
- `@boardsesh/board-presence` (pure): seq-dedup reducer + `previousClimb` for undo + mapper (17 tests).
- `@boardsesh/board-presence-react`: `useBoardPresence(boardId, client)` — subscribe-before-backfill, stale-async guards, memoized output, `reportClimb` + `undo` (13 tests). Injected `BoardPresenceClient` (no GraphQL client / host components).
- GraphQL operations (5) + analytics `SHARED_EVENTS` (`boardId` property, **never the raw serial**).

**Mobile (RN/Expo)**
- `BoardPresenceClient` over graphql-ws; provider owns the connected `boardId`; `bluetooth-provider` resolves on connect + reports on wall-confirm (solo too).
- **BoardSheet** (BoardPill→sheet: now-on-the-wall hero + virtualized history + stats + a **separate** Switch-board); accessory **source-of-truth flip** (O(1), flag+live-gated, local-queue fallback; swipe still steps the local queue); **undo snackbar**.

**Web / Capacitor**
- Web `BoardPresenceClient` + provider; `bluetooth-context` resolve-on-connect + report-on-wall-confirm (drives the legacy Capacitor path too); MUI now-on-the-wall **Drawer** + flag-gated entry pill + **undo Snackbar**.

### Key design note (correction to the original plan)
`board_id` is the **shared `userBoards.id`** (boards are shared/followable, one owner), resolved **serial-keyed** so everyone at one wall converges on a single id. Merged `main`'s `recordBoardSerial` (per-user serial→config) cleanly — it and `resolveBoardForSerial` (shared board binding) coexist.

## Validation

- Full `vp run typecheck`, `vp check` (0 errors).
- Tests: backend board-presence 13 · pure 17 · react 13 · **mobile 1370** · **web 4756** · i18n completeness + orphans + no-hardcoded-strings — all green.
- **Mac (iPhone 17 Pro / Metro)**: `typecheck:mobile` + `test:mobile` + **`check:mobile-bundle`** (iOS 10.4 MB / Android 10.6 MB) — the new packages resolve in Metro for both platforms.

## To go live (deliberately gated)
1. Apply migration `0120` (`vp run db:migrate`).
2. Set `BOARD_PRESENCE_ENABLED=true` (backend) and wire a source for the `board-presence` flag (it currently rides the server-driven `FeatureFlagsProvider`, off by default on both web and mobile — flag off = today's behaviour exactly).

## Follow-up (tracked separately)
- `help wanted` issue: redirect MoonBoard / serial-less boards to a per-physical-wall `board_id` via a BLE identifier (v1 uses a config-resolved `board_id`).
- Optional: repoint Live Activity to the wall's now-playing; consider unifying `recordBoardSerial` with `resolveBoardForSerial`.


marcodejongh (commented):
## Full review at c313302 (4-area deep review + local validation)

**Verdict: safe to merge dark; 4 items are blockers before `BOARD_PRESENCE_ENABLED=true`, none block merge.** Validated locally on this branch: full `vp run typecheck` green, board-presence + board-presence-react suites pass (34/34). The architecture is genuinely good — split contexts, capped pure reducer, subscribe-before-backfill proven safe across all race orderings, flag-off path byte-identical to main, no serial leakage into resolver outputs or analytics (with regression tests pinning that).

### Status of previously-flagged items — the earlier bot reviews are partly stale

The last automated review claims everything is still unresolved; that's wrong at HEAD. Verified one by one:

| Item | Status |
|---|---|
| `BACKFILL_HISTORY` never sets `previousClimb` (undo dead for late joiners) | **Fixed** — `reducer.ts:107` stashes the adopted-over current; asserted in `reducer.test.ts:159` |
| Undo reuses catalog UUID as queue-slot UUID | **Fixed by redesign** — hook now captures `undoTarget` and the platform re-reports; no UUID reuse |
| `saveTick` trusts client `boardId` | **Fixed** — flag-gated, `findActiveBoardById` + boardType match, falls back to config resolution (`ticks/mutations.ts:302-312`) |
| `reportBoardClimb` N+1 profile join | **Effectively addressed** — single `Promise.all`, 2 queries per rate-limited call |
| `COUNT(DISTINCT)` bigint-as-string | **Neutralized** — `Number()` coercion at every read (`queries.ts:129-130`); the `sql<number>` annotation is still optimistic |
| Migration 0121 without `CONCURRENTLY` | **Not fixed** — see flag-on blockers |
| `nextBoardSeq` INCR+EXPIRE non-atomic | **Not fixed, low impact** — worst case is a never-expiring counter key, never a regressed seq |
| `BoardClimbCleared` dead union leg | **Confirmed** — client reducer/mapper handle it, no producer exists. Add the mutation or a tracking TODO |
| Raw `db.execute(sql…)` in tests | **Not fixed** — also note the FIFO test re-implements lpush/ltrim by hand instead of exercising `storeBoardClimb`/`getRecentBoardClimbs` |
| `typescript` pin | **Half-fixed** — `board-presence-react` is `~6.0.3`, `board-presence/package.json:22` still `^5.9.3` |

### Blockers before flag-on

1. **Authorization model is name-the-integer-and-you're-in.** `boardNowPlaying` subscribe has **no `requireAuthenticated` and no `applyRateLimit`** (`subscription.ts:16-31`) — an unauthenticated WS client can loop board ids and grow `boardPresenceSubscribers` + Redis SUBSCRIBEs unboundedly. `boardRecentClimbs`/`boardPresenceStats` are anonymous-readable, and for unauthenticated WS connections the rate limit keys on the ephemeral `connectionId`, so reconnect = fresh bucket = scrapeable display names/avatars/stats by sequential id. And `reportBoardClimb` requires auth but zero linkage to the wall — any logged-in user can inject onto any board's feed. Suggested shape: stamp `presence:board:{id}:user:{uid}` in Redis on `resolveBoardForSerial`, require it (TTL'd) on report; add auth + a rate limit to the subscription; auth the queries or document anonymous-read as intentional and key limits on IP.
2. **Accessory display / tap / tick divergence.** When the wall climb ≠ local queue head, the capsule *displays* and QuickTickBar *ticks* the wall climb, but tapping the capsule opens the **local** head in PlayDrawer (`use-queue-carousel.ts:110` is untouched by the flip). Three surfaces, two climbs — in exactly the multi-climber scenario this feature exists for. Route tap through the same `useWallOrQueueCurrentClimb` result.
3. **Swipe appears dead while wall-pinned.** While the override is live the leading slot is pinned to the wall climb, but swipes step the invisible local queue and the peek slots show local-queue neighbours (`use-wall-or-queue-climb.ts:46-58`, `NativeAccessoryClimbRow.tsx:94-99`). The code comments call this deliberate, but to the user a swipe silently does nothing. Either suppress swipe while wall-overriding or re-anchor the local queue on wall change.
4. **Serial normalization.** `BoardSerialSchema` accepts mixed case and doesn't trim; `parseSerialNumber` captures verbatim; the resolver compares with exact `eq()` while migration 0121 only BTRIMs historical rows — and `boards.ts` create/update *does* trim, so the two paths can disagree. Two phones whose device names differ in case/whitespace mint two boards and never converge, which defeats the feature. Normalize (trim + case-fold if Aurora serials are case-insensitive) in one place and compare normalized.

### Should-fix (can ride a follow-up, but before relying on the feature)

- **Redis seq-counter reset freezes the wall silently** — after a flush/failover, every live event has `seq <= lastSeq` and is swallowed; the only escape is switching boards. Now documented (`reducer.ts:6-12`) but unhandled. An epoch alongside `seq`, or `RESET` on reconnect, fixes it.
- **No resubscribe after terminal subscription error** — `use-board-presence.ts:128-138` flips `isLive` off permanently until remount; expose a retry or re-run on an error counter.
- **Migration 0121 lock window** — two full-table UPDATEs + non-concurrent unique index on `user_boards`. If the table is small enough that this is sub-second, say so on the PR; otherwise split the index into a manual `CONCURRENTLY` step.
- **`use-wall-or-queue-climb.ts` has no unit test** — every accessory test mocks it to a passthrough, so the actual flip logic (flag/`isLive`/`boardId` gating, same-uuid identity branch, fallback) is never exercised. The flag-off path is only pinned structurally.

### Nits

`boardId` in `logger.warn` race messages can include the conflicting serial via `String(error)` (`mutations.ts:99,146`) — log hygiene; `gradeColor` is always `null` server-side, dead field until populated or dropped; the combined `BoardPresenceContext` (`board-presence-provider.tsx:56`) carries the whole hook result and defeats the context split if anything hot ever consumes it — drop it if unconsumed; `sendFramesToBoard` gains an `analyticsBoardId` dep that churns its identity mid-connect (absorbed by the dedup drain, but a ref read avoids hot-path churn); serial-squatting via fabricated `resolveBoardForSerial` makes the squatter `ownerId` of a shared wall (proof-of-presence from blocker 1 also mitigates this); typescript pin above.

### Verified-good (so nobody re-litigates these)

`storeBoardClimb` awaited before publish so backfill can't miss a live event; per-board seq via Redis INCR is monotonic cluster-wide; find-or-bind race handled via unique-violation catch + re-read with `isNull(serialNumber)` guard; redis-adapter changes purely additive (no party-session regression); flag-off is inert end-to-end (no WS client, no subscription, no boardId stamping) on both platforms; mobile perf checklist passes (BottomSheetFlatList, memo'd rows, hoisted keyExtractor, O(1) flip selector, memoized split contexts); i18n parity complete across en-US/es/fr for all new keys on both platforms; solo wall-confirm reports without a session (`bluetooth-provider.tsx:376-397` gates the party confirm on `sessionIdRef` but reports presence regardless) — which matters because…

### Interaction with #2710 (lazy-session removal)

Checked file-by-file: **exactly one overlapping file** (`drawer-host-provider.test.tsx`), and a real 3-way `git merge-file` auto-merges it cleanly and coherently. No semantic conflicts: the wall flip is display-only while #2710's gating governs mutations; undo re-reports through the BLE/presence layer, never `setCurrentClimb`; the report path already tolerates `sessionId === null`. **Recommended order: merge #2710 first, then this — expected rework: none.** After both land, the deferred "repoint Live Activity at the wall's now-playing" follow-up is the natural way to give solo climbers a lock-screen wall indicator again.

---

PR #2679, opened 2026-06-10
[codex] Fix mobile Bluetooth picker filtering

marcodejongh:
## Summary
- Restrict mobile BLE picker scans by board family so Aurora routes no longer surface generic UART peripherals like AirPods.
- Resolve BLE serials in mobile like the web picker, preferring saved boards and falling back to recorded board configs.
- Render board previews in the React Native Bluetooth picker sheet.

## Validation
- `vp test run --project mobile packages/mobile/src/lib/ble/__tests__/board-device-filter.test.ts packages/mobile/src/lib/ble/__tests__/adapter.test.ts packages/mobile/src/lib/ble/__tests__/native-ios-adapter.test.ts packages/mobile/src/lib/ble/__tests__/use-board-bluetooth.test.ts packages/mobile/src/lib/ble/__tests__/resolve-serials.test.ts packages/mobile/src/components/ble/__tests__/DeviceCard.test.tsx`
- `vp run typecheck:mobile`

## Notes
- `vp check` still fails on unrelated pre-existing formatting issues outside this change.


marcodejongh:
Reviewed this branch against main plus the wider mobile BLE stack, then pushed fixes in ddd8119:

**CI-blocking:** the four `boardConfigMismatch.mobile*` keys existed only in en-US — catalog-completeness was failing. Added es/fr translations (plus a new `mobileSwitchFailed` in all three locales).

**Pre-existing lifecycle bugs (fixed here since this branch already touches the files):**
- `clearConnectionAfterDrop` nulled `adapterRef` without disposing the adapter — the write-failure "connection stolen" path leaked the RNBleAdapter's `onDeviceDisconnected` subscription and never cancelled a half-alive native link. `NativeIosBleAdapter.disconnect()` now also no-ops when no device is tracked, so disposing an already-self-cleaned adapter can't tear down a later-adopted connection on the native singleton.
- Nothing disconnected when the active board switched while connected: sends kept streaming packets encoded with the new config's LED placement map to the old physical wall. A new effect tears the connection down on config-key divergence (incl. a switch landing mid-connect), suppresses adoption across the switch (the named-device adopt guard is boardType-granular only), and preserves `lastConnectedBoard` so switching back reconnects silently.

**Web parity:** the mismatch alert now offers "Switch to correct config" — saved boards activate directly, recorded configs resolve via `GET_BOARD` by uuid, then a one-shot auto-connect by serial fires once the provider sees the switched config. Note one deliberate divergence from web: mobile adopts the target board's stored angle rather than preserving the current angle (the current angle belongs to a different board).

**Perf/edge:** `DeviceCard` is now memoized and `handlePickerSelect` reads volatile state through refs, so picker rows stop re-rendering on every advertisement. The Aurora device filter treats `undefined` serviceUuids (old iOS binaries, native UUID filter already applied) as vouched again — renamed boards without the `#serial@N` suffix were silently dropped.

Reviewed and dismissed as non-issues: duplicate adoption adapters (guarded by `adapterRef`/`connectInFlightRef`), concurrent AutoSender drains (synchronous `isWritingRef` gate), write-chain poisoning (already double-coerced).

Validation: typecheck:mobile ✅, test:mobile 1482/1482 ✅, i18n catalog 28/28 ✅, Metro bundle check ✅.

marcodejongh:
Addressed the review feedback in 0f5fc1d:

- **`parseSetIds` zero bug** — empty segments are dropped now (`Number.isInteger(setId) && setId > 0`), with a regression test asserting `"1,,20"` reaches the renderer as `[1, 20]`.
- **Picker cancelled too early in `handleMismatchSwitch`** — the cancel now happens only after the board fetch and `setActiveBoard` succeed. On failure the picker stays open (alert shows over it) so the user can still pick a device or use Connect anyway; test updated to assert `handleCancel` is not called on the failure path.
- **`previewImageStyle` outside the memo** — folded into the render-data `useMemo`.
- **`renderItem` dep on the `resolvedBoards` Map** — went one step further than keeping the map stable: `DeviceCard` now takes its own `resolvedEntry` instead of the whole map, looked up in `renderItem`. Rows whose entry didn't change keep referentially identical props, so `React.memo` skips them when another row's resolution lands. The `renderItem` identity still changes with the map — that's what propagates new resolutions — but the blast radius is now per-row.

Validation: typecheck:mobile ✅, test:mobile 1646/1646 ✅, Metro bundle ✅.

marcodejongh:
Addressed the follow-up review items in f761611:

- **Cross-feature import** — `formatRelativeTime` moved from `create-climb/draft-format` to `src/lib/format-relative-time.ts`; `DeviceCard` and `DraftRow` both import from there now, and its tests moved to `src/lib/__tests__/` (`draft-format` keeps only `countHolds`).
- **Untested GET_BOARD path** — three new tests: a recorded config with a `boardUuid` fetches the board and switches to it (asserting the `{ boardUuid }` request and picker cancel), a rejected request surfaces the switch-failed alert with the picker left open and `setActiveBoard` untouched, and a `{ board: null }` response is treated the same way.
- **Stuck `pendingAutoConnect`** — the one-shot now expires after 15s if the switched config never propagates, so a stale request can't fire on a much-later unrelated switch to the same config. Covered with a fake-timer test (arm, advance past the TTL, then deliver the matching props and assert `connect` never fires).

Validation: typecheck:mobile ✅, test:mobile 1650/1650 ✅, Metro bundle ✅.

marcodejongh:
Addressed the third round in 7b06e95. Four items fixed, two claims checked out as incorrect:

**Fixed:**
- **Query-key churn in `useResolvedBleDeviceBoards`** — the serials array is now keyed on its joined string, so its reference (and the query key) only changes when the serial set actually does. One correction to the report though: TanStack hashes query keys structurally, so the unstable reference never triggered redundant network requests — this is render/memo hygiene, not a fetch fix. (`@` is the join delimiter since `parseSerialNumber` matches `/#([^@]+)/`, so a serial can never contain it.)
- **`holdsData` set-ID filter** — now `Number.isInteger(setId) && setId > 0`, in lockstep with DeviceCard's `parseSetIds`.
- **Division-by-zero in `getPreviewImageStyle`** — corrupt dimensions (zero/negative/non-finite) fall back to a square thumbnail instead of an invisible strip.
- **Test coverage** — landscape, portrait, square, and corrupt-dimension preview styles; `describeSavedBoard` with a gym location and with every optional descriptor missing (both helpers exported for testing). The `pendingAutoConnect` single slot is now documented as deliberate last-writer-wins: each successful switch cancels the picker that produced it, so a second request can only come from a newer flow whose intent supersedes the first.

**Pushed back — not bugs:**
- **"RNBleAdapter.disconnect() calls cancelDeviceConnection unconditionally"** — it doesn't: `disconnect()` guards on `this.connectedDevice` (adapter.ts), which the adapter's own `onDeviceDisconnected` handler nulls *before* invoking the drop callback. The event-driven path is a clean no-op on the native call; the write-failure path keeps the cancel it actually needs.
- **"devicePicker.* keys absent from all locale catalogs"** — they exist in en-US, es, and fr (`settings.json:353-355` in each), and the catalog-completeness test passes. `vp run check:i18n` also only scans `packages/web/app/`, which this PR doesn't touch.

Validation: typecheck:mobile ✅, test:mobile 1660/1660 ✅, iOS + Android bundles ✅.

PR #2681, opened 2026-06-10
[codex] Fix backend log errors

marcodejongh:
## Summary

This PR fixes the backend log noise by separating three different problems that were showing up together in the logs:

- stale or incompatible backend/database deploys should be detected clearly at startup instead of surfacing later as confusing runtime errors;
- broad quality-sorted climb searches should avoid the PostgreSQL plan that was triggering dynamic shared-memory allocation failures;
- expected client disconnects and websocket session races should not be logged like backend defects.

### Deploy preflight

The logs showed backend code trying to run against a schema state it was not built for. That is hard to diagnose after the server is already accepting traffic, so the backend now reads the bundled Drizzle migration journal at startup and compares it with the applied migration state in Postgres.

The check intentionally crashes only when the database is provably ahead of the bundled code. If migration state cannot be read because the database is unavailable, missing the migration table, or still being initialized, startup warns and continues. That keeps transient DB boot issues from becoming a crash loop while still protecting against stale bundles.

The comparison is based on migration identity/count rather than `created_at` timestamps because some environments stamp applied migrations with apply time, not journal time.

### Quality search performance

`quality DESC` previously used the standard `board_climbs LEFT JOIN board_climb_stats` search path. For broad searches that could lead PostgreSQL into a parallel sort plan and fail with dynamic shared-memory allocation errors.

This PR routes stats-backed desc sorts (`ascents`, `quality`) through a stats-driven search path first, backed by a covering index for `quality_average DESC NULLS LAST`. That mirrors the existing ascents optimization and keeps the hot path on the stats index.

There is still a first-page fallback to the standard search when the stats-driven result is sparse. That fallback is intentional: climbs without stats still need to appear at the bottom of narrow first-page results instead of disappearing entirely. The trade-off is that sparse first pages can do two queries.

### Size filter predicate

The size filter changed from `size_id = ANY(compatible_size_ids)` to `compatible_size_ids @> ARRAY[size_id]::int[]` so PostgreSQL can use the existing `board_climbs_compatible_size_ids_idx` GIN index from migration `0073_denormalize_climb_sets_sizes`.

No new index is needed here. PostgreSQL's built-in GIN `array_ops` supports `@>` for `integer[]`, so this does not depend on the `intarray` extension.

### Websocket session subscriptions

Mobile could open queue/session subscriptions before `JOIN_SESSION` completed on the same websocket connection. During deploys or reconnects that caused repeated `requireSessionMember` failures in backend logs.

Mobile now waits for `JOIN_SESSION` before opening joined-only subscriptions. If joining fails on a still-live socket, it retries with capped backoff instead of leaving the client silently unsubscribed. On socket close, subscription refs are cleared and the `graphql-ws` `closed`/`connected` lifecycle drives teardown, rejoin, and resubscribe.

The docs were updated to make that explicit: subscription `error`/`complete` handlers do not drive reconnect; socket lifecycle events do.

### Client abort logging

Client disconnects were being logged as request-handler errors. The backend now classifies known client abort codes (`ECONNRESET`, `EPIPE`, `ERR_STREAM_PREMATURE_CLOSE`) separately.

The message matching is intentionally narrow: only exact known Node HTTP/socket messages such as `aborted` and `socket hang up` are treated as client aborts. Bare `AbortError` is only treated as a client abort when request/response/socket state shows the connection was actually destroyed, so upstream/server-side aborts still log as real errors.

### `searchClimbs` transaction boundary

`searchClimbs` is still typed for the top-level `DbInstance`, which is what production call sites pass today. `standardSearch` uses a transaction-scoped `SET LOCAL max_parallel_workers_per_gather = 0` so the planner guard is scoped to the search query.

The execute-only branch is kept for query test doubles and no longer throws. The code comments now call out that broadening `searchClimbs` to accept a transaction-scoped DB should be a deliberate signature change, because the parallelism guard needs to be preserved.

### Drizzle snapshot note

`board_climb_stats_quality_covering_idx` is created by a custom raw-SQL migration and is not represented in `0121_snapshot.json`. That matches the existing raw-SQL covering-index pattern used for the ascents search index. No schema snapshot action is needed unless the project changes how it models these raw SQL indexes.

## Validation

Local validation run on the branch:

- `vp check --fix` on changed files
- `vp run typecheck:db`
- `vp run typecheck:backend`
- `vp run typecheck:mobile`
- `vp run @boardsesh/db#test`
- `vp test run --project mobile src/providers/__tests__/queue-provider-session-updates.test.tsx`
- `SKIP_TEST_INFRA=1 vp test run --project backend src/__tests__/search-climbs-frames.test.ts src/__tests__/deploy-preflight.test.ts src/__tests__/http-errors.test.ts`

CI on the latest pushed commit is green, including lint, typecheck, backend/default tests, mobile bundle, iOS builds, Android prebuild/build, and Claude review.

### DIFF
--- a/packages/backend/src/utils/deploy-preflight.ts
+++ b/packages/backend/src/utils/deploy-preflight.ts
@@ -1,0 +1,150 @@
+import { sql } from 'drizzle-orm';
+import { executeFirstRow } from '@boardsesh/db/client';
+import bundledMigrationJournal from '@boardsesh/db/migration-journal' with { type: 'json' };
+import { db } from '../db/client';
+import { asErrorLikeRecord } from './error-utils';
+import { logger } from './logger';
+
+type MigrationJournalEntry = {
+  tag: string;
+  when: number;
+};
+
+type MigrationJournal = {
+  entries: MigrationJournalEntry[];
+};
+
+type LatestMigrationRow = {
+  latestMigrationCreatedAt: string | number | bigint | null;
+};
+
+export type MigrationInfo = {
+  tag: string;
+  createdAt: number;
+};
+
+const BUILD_SHA_ENV_KEYS = [
+  'RAILWAY_GIT_COMMIT_SHA',
+  'VERCEL_GIT_COMMIT_SHA',
+  'GIT_SHA',
+  'COMMIT_SHA',
+  'SOURCE_VERSION',
+] as const;
+
+type BuildShaEnv = Partial<Record<(typeof BUILD_SHA_ENV_KEYS)[number], string | undefined>>;
+
+function isMigrationJournal(value: unknown): value is MigrationJournal {
+  if (!value || typeof value !== 'object' || !('entries' in value)) return false;
+  const entries = (value as { entries?: unknown }).entries;
+  return Array.isArray(entries);
+}
+
+export function getBackendBuildSha(env: BuildShaEnv | NodeJS.ProcessEnv = process.env): string {
+  for (const key of BUILD_SHA_ENV_KEYS) {
+    const sha = env[key];
+    if (sha) return sha;
+  }
+  return 'unknown';
+}
+
+export function readBundledLatestMigration(journal: unknown = bundledMigrationJournal): MigrationInfo | null {
+  if (!isMigrationJournal(journal)) return null;
+
+  const latestEntry = journal.entries
+    .filter(
+      (entry): entry is MigrationJournalEntry =>
+        !!entry && typeof entry.tag === 'string' && typeof entry.when === 'number',
+    )
+    .reduce<MigrationJournalEntry | null>((latest, entry) => {
+      if (!latest || entry.when > latest.when) return entry;
+      return latest;
+    }, null);
+  if (!latestEntry) return null;
+
+  return { tag: latestEntry.tag, createdAt: latestEntry.when };
+}
+
+export function isMissingMigrationTableError(error: unknown, depth = 0): boolean {
+  if (depth > 3) return false;
+
+  const errorRecord = asErrorLikeRecord(error);
+  if (!errorRecord) return false;
+
+  // 42P01 = undefined_table, 3F000 = invalid_schema_name.
+  if (errorRecord.code === '42P01' || errorRecord.code === '3F000') return true;
+  return isMissingMigrationTableError(errorRecord.cause, depth + 1);
+}
+
+type ReadLatestAppliedMigrationOptions = {
+  database?: Parameters<typeof executeFirstRow>[0];
+  executeFirstRow?: typeof executeFirstRow;
+};
+
+export async function readLatestAppliedMigrationCreatedAt(
+  schemaName: 'drizzle' | 'public',
+  options: ReadLatestAppliedMigrationOptions = {},
+): Promise<number | null> {
+  const database = options.database ?? db;
+  const firstRow = options.executeFirstRow ?? executeFirstRow;
+  const query =
+    schemaName === 'drizzle'
+      ? sql`SELECT MAX("created_at")::bigint AS "latestMigrationCreatedAt" FROM drizzle."__drizzle_migrations"`
+      : sql`SELECT MAX("created_at")::bigint AS "latestMigrationCreatedAt" FROM public."__drizzle_migrations"`;
+
+  try {
+    const row = await firstRow<LatestMigrationRow>(database, query);
+    if (!row?.latestMigrationCreatedAt) return null;
+    return Number(row.latestMigrationCreatedAt);
+  } catch (error) {
+    if (isMissingMigrationTableError(error)) return null;
+    throw error;
+  }
+}
+
+type DeployPreflightLogger = Pick<typeof logger, 'info' | 'warn'>;
+
+type VerifyDeployCompatibilityOptions = {
+  env?: BuildShaEnv | NodeJS.ProcessEnv;
+  logger?: DeployPreflightLogger;
+  readBundledLatestMigration?: () => MigrationInfo | null;
+  readLatestAppliedMigrationCreatedAt?: (schemaName: 'drizzle' | 'public') => Promise<number | null>;
+};
+
+export async function verifyDeployCompatibility(options: VerifyDeployCompatibilityOptions = {}): Promise<void> {
+  const buildSha = getBackendBuildSha(options.env);
+  const deployLogger = options.logger ?? logger;
+  const bundledMigrationReader = options.readBundledLatestMigration ?? readBundledLatestMigration;
+  const appliedMigrationReader = options.readLatestAppliedMigrationCreatedAt ?? readLatestAppliedMigrationCreatedAt;
+  let bundledLatestMigration: MigrationInfo | null = null;
+
+  try {
+    bundledLatestMigration = bundledMigrationReader();
+  } catch (error) {
+    deployLogger.warn('[Deploy] Could not read bundled migration journal', { buildSha, error });
+  }
+
+  deployLogger.info('[Deploy] Backend build info', {
+    buildSha,
+    expectedMigration: bundledLatestMigration?.tag ?? null,
+    expectedMigrationCreatedAt: bundledLatestMigration?.createdAt ?? null,
+  });
+
+  if (!bundledLatestMigration) return;
+
+  const latestAppliedMigrationCreatedAt = Math.max(
+    (await appliedMigrationReader('drizzle')) ?? 0,
+    (await appliedMigrationReader('public')) ?? 0,
+  );
+
+  if (latestAppliedMigrationCreatedAt === 0) {
+    deployLogger.warn('[Deploy] Could not find applied drizzle migrations table', { buildSha });
+    return;
+  }
+
+  if (latestAppliedMigrationCreatedAt > bundledLatestMigration.createdAt) {
+    throw new Error(
+      `[Deploy] Database migration timestamp ${latestAppliedMigrationCreatedAt} is newer than backend bundle ` +
+        `${buildSha} expects (${bundledLatestMigration.tag}:${bundledLatestMigration.createdAt}). Deploy newer code.`,
+    );
+  }
+}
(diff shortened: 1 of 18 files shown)


packages/db/drizzle/meta/_journal.json
```diff
@@ -848,6 +848,13 @@
       "when": 1780997928774,
       "tag": "0120_charming_lyja",
       "breakpoints": true
+    },
+    {
+      "idx": 121,
+      "version": "7",
+      "when": 1781058236000,
```
marcodejongh: This journal entry looks hand-written rather than produced by drizzle-kit: the `when` timestamp ends in `000` (every generated entry has real ms precision), the tag doesn't follow the generated codename pattern, and there's no `meta/0121_snapshot.json` (compare `0099_split_ascensionist_count`, a custom migration that has both a snapshot and a real timestamp — presumably via `drizzle-kit generate --custom`).

Per CLAUDE.md this should go through `drizzle-kit generate` (`--custom` for raw SQL like an `INCLUDE` index). It matters more now than before: this same PR makes the journal `when` values load-bearing for `verifyDeployCompatibility` — the preflight compares `MAX(created_at)` in `__drizzle_migrations` against the bundled journal, and drizzle stores `created_at` from this exact field. Hand-maintained timestamps are now a deploy-gating input, not just metadata.

---

packages/backend/src/utils/deploy-preflight.ts
```diff
@@ -0,0 +1,150 @@
...
+      ? sql`SELECT MAX("created_at")::bigint AS "latestMigrationCreatedAt" FROM drizzle."__drizzle_migrations"`
+      : sql`SELECT MAX("created_at")::bigint AS "latestMigrationCreatedAt" FROM public."__drizzle_migrations"`;
+
+  try {
+    const row = await firstRow<LatestMigrationRow>(database, query);
+    if (!row?.latestMigrationCreatedAt) return null;
+    return Number(row.latestMigrationCreatedAt);
+  } catch (error) {
+    if (isMissingMigrationTableError(error)) return null;
+    throw error;
```
marcodejongh, line 129: Only the missing-table case is swallowed here; any other DB error (connection refused, DNS, auth, timeout) propagates out of `verifyDeployCompatibility` and crashes `startServer` before it accepts traffic. Since the preflight runs on every boot, a transient DB blip turns into a backend crash loop — the opposite of what this PR is trying to achieve for log noise/availability.

The throw is only meaningful for the one condition the check exists for (DB schema newer than bundle). Suggest catching unexpected errors around the two `appliedMigrationReader` calls in `verifyDeployCompatibility` and warn+continue, same as the missing-table path. A test for that path would lock it in.

Related: the strict `>` comparison also makes *rollbacks* impossible after any migration, even backward-compatible ones — rolling back the backend after `0121` (an index-only migration an old bundle is perfectly happy with) will refuse to start. That may be an acceptable trade-off, but worth stating in the comment on `startServer`, or gating strictness behind an env var so an emergency rollback has an escape hatch.

---

packages/mobile/src/providers/queue-provider.tsx
```diff
@@ -599,208 +599,229 @@ export function QueueProvider({ children }: { children: ReactNode }) {
...
+    const startJoinedSubscriptions = async () => {
+      const currentStartToken = ++subscriptionStartToken;
+      cleanupSubscriptions();
+
+      try {
+        await ensureJoined(sessionId);
+      } catch (joinError) {
+        if (!disposed) logJoinFailure(joinError);
+        return;
+      }
```
marcodejongh, line 655: Behavioural gap vs. the old eager flow: if `ensureJoined` rejects here on a live socket (e.g. JOIN_SESSION hits the backend mid-deploy — exactly the scenario this PR is motivated by), we log a `__DEV__`-only warn and return. No retry is scheduled and nothing user-visible happens, so the client stays silently deaf to queue/session events until the socket happens to cycle or the user switches sessions. Pre-PR, the subscriptions at least auto-resubscribed on reconnect independent of join state (noisy on the backend, but self-healing).

Suggest a bounded retry with backoff on join failure (re-invoking `startJoinedSubscriptions` while the token is still current), or at minimum surfacing it via `showToastRef` like the queue subscription `error` path does.

---

packages/backend/src/utils/http-errors.ts
```diff
@@ -0,0 +1,26 @@
...
+  }
+
+  if (typeof errorRecord.message === 'string') {
+    const normalizedMessage = errorRecord.message.trim().toLowerCase();
+    if (CLIENT_ABORT_ERROR_MESSAGES.has(normalizedMessage)) return true;
+  }
+
+  if (typeof errorRecord.name === 'string' && errorRecord.name === 'AbortError') {
+    return true;
+  }
```
marcodejongh, line 33: `name === 'AbortError'` doesn't only mean "client went away" — it's also what server-side `AbortController` timeouts throw, and this backend uses those for upstream fetches (posthog proxy, instagram/tiktok meta, web-revalidate). Those handlers catch their own aborts today, but if one ever leaks to the top-level catch in `server.ts`, this classifier swallows it: no 500 is written, the *live* client hangs until its own timeout, and the log says "Request aborted by client" for what was actually an upstream timeout.

A more truthful signal of a client abort is connection state — e.g. require `req.destroyed || res.destroyed` (or `res.socket?.destroyed`) before taking the early-return path, and let everything else fall through to the error log + 500.

---

packages/db/drizzle/0121_add_quality_search_covering_index.sql
```diff
@@ -0,0 +1,5 @@
+-- Cover broad quality-sorted climb searches without a parallel sort.
+-- Matches searchClimbs' stats-driven `quality DESC NULLS LAST` path.
+CREATE INDEX IF NOT EXISTS board_climb_stats_quality_covering_idx
+  ON board_climb_stats (board_type, angle, quality_average DESC NULLS LAST)
+  INCLUDE (climb_uuid, ascensionist_count, display_difficulty, difficulty_average, benchmark_difficulty);
```
marcodejongh, line 9: Plain `CREATE INDEX` takes a lock that blocks writes to `board_climb_stats` for the duration of the build — and this is one of the bigger tables (per-angle stats rows) with the stats sync writing to it. Migration 0068 set the same precedent, so maybe that build time is known to be acceptable, but if it's more than a few seconds in prod this wants `CREATE INDEX CONCURRENTLY` run outside the migration transaction (drizzle supports a migration without a wrapping transaction via breakpoints, or run it manually and keep this as `IF NOT EXISTS` for dev parity). Worth a quick check of the prod table size before merging.

---

packages/db/package.json
```diff
@@ -41,6 +41,11 @@
       "import": "./src/client/index.ts",
       "default": "./src/client/index.ts"
     },
+    "./migration-journal": {
+      "types": "./drizzle/meta/_journal.json",
```
marcodejongh: Nit: `types` is meant to point at a `.d.ts`, not the JSON itself — TypeScript resolves the JSON shape from the `import` target with `resolveJsonModule`, so this entry can just be dropped (keep `import`/`default`).

---

packages/db/src/queries/climbs/search-climbs.ts
```diff
@@ -235,11 +271,51 @@ async function statsDrivenSearch(
...
+    });
+  }
+
+  // Transaction handles cannot start nested transactions, but SET LOCAL still
+  // applies to their outer transaction. Lightweight query test doubles omit
+  // execute(), so they skip only this production planner guard.
+  const execute = getExecute(db);
+  if (execute) {
+    await execute(sql`SET LOCAL max_parallel_workers_per_gather = 0`);
+  }
```
marcodejongh: Two nits on this fallback branch:

1. The comment's first sentence doesn't match the code path it sits on. A drizzle `TransactionDb` *does* expose `.transaction` (savepoints), so it takes the `getTransaction` branch above — this `getExecute` branch is only reachable for objects with `execute` but no `transaction`, which in practice is just test doubles, as the second sentence says.
2. If a real non-transactional client ever did land here, `SET LOCAL` outside a transaction is a no-op (Postgres emits a warning and ignores it), so the planner guard would silently not apply.

Since both real callers (backend GraphQL via node-postgres `DbInstance`, or an outer transaction handle) go through the transaction branch, consider dropping this branch entirely and letting test doubles skip the guard by not entering `getTransaction` — or at least rewriting the comment to say it's test-double-only and ineffective outside a transaction.

---

marcodejongh (commented):
Reviewed the full diff at 07dfe54. The overall shape is good: routing quality DESC through the stats covering index mirrors the proven ascents path, the JOIN_SESSION gating on mobile is carefully token-guarded against the reconnect races (stale joins discarded via `subscriptionStartToken`, epoch bump on `closed`), and the new tests cover the paths earlier review rounds flagged (happy-path preflight, out-of-order journal, deferred-join subscription gating). The docs sequence diagram fork for web vs. mobile is a nice touch.

Three things I'd fix before merge, four smaller notes — details inline:

**Worth fixing**
1. **Migration 0121 is hand-written** (rounded `when`, no `0121_snapshot.json`) — and this PR simultaneously makes journal timestamps a deploy-gating input for `verifyDeployCompatibility`. Regenerate via `drizzle-kit generate --custom`.
2. **Preflight crashes startup on any transient DB error**, not just the stale-bundle condition it exists to catch — a DB blip at boot becomes a crash loop. Warn+continue on unexpected errors; also note the rollback implications of the strict timestamp check.
3. **Mobile: a failed JOIN_SESSION on a live socket leaves the client permanently deaf** with only a `__DEV__` console.warn — no retry, no toast. This is a regression vs. the old eager flow's self-healing resubscribe, and it's most likely to trigger during backend deploys, the exact scenario the PR addresses.

**Minor**
- `isClientAbortError` treating any `AbortError` as a client abort can misclassify server-side fetch timeouts and leave a live client hanging; connection state (`req.destroyed`/`res.destroyed`) is the truthful signal.
- Plain `CREATE INDEX` (not `CONCURRENTLY`) blocks writes to `board_climb_stats` during deploy — check prod build time.
- `package.json` `types` pointing at the raw JSON can be dropped.
- The `getExecute` fallback in `standardSearch` is effectively test-double-only and its `SET LOCAL` would be a no-op outside a transaction; the comment overstates it.

The `@>` containment rewrite for the size filter is correct (semantics identical to `= ANY()`, and it makes the GIN index usable), and web is unaffected by the `searchClimbs` signature change since it goes through the GraphQL backend, whose node-postgres client supports the new transaction wrapper.

---

PR #2691, opened 2026-06-10
[codex] Fix mobile session setup scrolling

marcodejongh:
Fixes the Record pre-session mobile layout issues: Android chip rails now use gesture-handler nested horizontal ScrollViews, the pre-session content uses a gesture-handler vertical ScrollView, and the Start session footer no longer uses the oversized scroll reserve. The footer now keeps Android's corrected spacing while adding iOS tab-bar-height clearance so it does not sit under the native tab bar. Validated with: vp test run --project mobile -- packages/mobile/src/components/session-screen/pre-session; vp run typecheck:mobile; git diff --check. Note: vp check still reports unrelated pre-existing formatting issues outside this change.

### DIFF
--- a/packages/mobile/app/(tabs)/__tests__/tab-layout.test.tsx
+++ b/packages/mobile/app/(tabs)/__tests__/tab-layout.test.tsx
@@ -127,7 +127,15 @@ describe('TabLayout', () => {
     expect(container.querySelector('[data-tabs="true"]')?.getAttribute('data-minimize-behavior')).toBe('onScrollDown');
   });
 
-  it('only mounts the native bottom accessory when that path is active', () => {
+  it('mounts the native bottom accessory when that path is active', () => {
+    const { container } = render(<TabLayout />);
+
+    const accessorySlot = container.querySelector('[data-bottom-accessory="true"]');
+    expect(accessorySlot).not.toBeNull();
+    expect(accessorySlot?.querySelector('[data-accessory="true"]')).not.toBeNull();
+  });
+
+  it('skips the native bottom accessory when that path is inactive', () => {
     cfg.nativeAccessoryActive = false;
 
     const { container } = render(<TabLayout />);
--- a/packages/mobile/app/(tabs)/__tests__/tab-layout.test.tsx
+++ b/packages/mobile/app/(tabs)/__tests__/tab-layout.test.tsx
@@ -6,6 +6,18 @@ import { createElement, type ReactNode } from 'react';
 const cfg = vi.hoisted(() => ({
   bluetoothConnected: false,
   sessionId: null as string | null,
+  nativeAccessoryActive: true,
+  variant: 'liquidGlass' as 'liquidGlass' | 'material',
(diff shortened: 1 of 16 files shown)


marcodejongh:
## Review follow-up

Pushed a follow-up commit addressing the open findings, plus a rebase onto latest `main` (resolved the `SessionLiveness` → `SessionStatus` rename in `queue-provider-leave-session.test.tsx`, keeping the branch's `operationText()` helper).

### Fixed

- **Magic number `100` for footer clearance** — now `SESSION_FOOTER_CLEARANCE` in `session-footer-clearance.ts`, shared by `PreSessionView` and `InSessionView` (both had the same literal).
- **`tab-layout.test.tsx` accessory coverage** — added the positive-branch test: the native bottom accessory mounts and hosts `QueueBottomAccessory` when the path is active; the existing negative test keeps the inactive branch covered.
- **`renderScrollComponent` comment** — now records the verification (see below) instead of just claiming TS-surface support.

### Verified, no change needed

- **"`renderScrollComponent` is undocumented/experimental in FlashList v2"** — verified against the installed `@shopify/flash-list@2.3.1`: the prop is in the public TS surface (`dist/FlashListProps.d.ts:101`) and consumed at runtime (`dist/recyclerview/hooks/useSecondaryProps.js` wraps it via `createAnimatedComponent` and feeds it to the recycler as the scroll host).
- **"ChipRail scroll position resets on every chip tap"** — false positive. `ChipRail`/`OptionChip` are module-scope components; a chip tap recomputes the `listHeader` `useMemo` and re-renders the rails, but React reconciles the same element type in place — the native horizontal `ScrollView` is never unmounted, so its scroll offset survives. (Re-render ≠ remount.)
- **"`estimatedItemSize` missing"** — that's a FlashList v1 prop; v2 auto-measures and removed it (`estimatedItemSize` does not exist in 2.3.1's `FlashListProps`).
- **"`scrollBottomPadding` regression for Material tabs"** — stale finding from an earlier revision. The final code leaves `scrollBottomPadding` untouched (`bottom-chrome-metrics.ts`) and only adds the new `fixedFooterBottom` field; `ProgressTab`/`LogbookTab`/`SessionsTab`/`PlaylistDetailView` see no change.
- **Codex P2 "JS queue bar hidden outside native tabs"** — fixed in a later revision than the one reviewed: `persistent-queue-bar.tsx` now relies solely on `jsQueueToolbarVisible`, which is computed with `nativeAccessoryMounted = insideTabs && nativeAccessoryActive`, so the JS fallback still shows outside the tabs group. Covered by the test "keeps the JS toolbar fallback outside tabs when the native accessory path is active".

### Validation

- `vp run typecheck:mobile` ✅
- `vp run test:mobile` — 224 files / 1600 tests ✅
- Rebased onto latest `main`; the branch is conflict-free.

A debug build installs and runs on the API 35 emulator; I haven't yet driven the full on-device scroll pass on this machine (emulator↔Metro plumbing), so a manual Android scroll smoke test on the preview build is still worth doing before merge.

PR #2692, opened 2026-06-10
Drive provider sign-out from the 401 interceptor on failed refresh (#2684)

### LINKED ISSUE #2684: [P2] 401-with-failed-refresh calls token-only signOut, leaving AuthProvider stuck on a logged-in UI with a dead token
**User impact:** When the server revokes a session (password change, admin revoke, refresh-token rotation failure) the app does not return the user to login. The UI stays on the climbs/queue/profile screens but silently fails every fetch — empty lists, spinners, "something went wrong" — until the user happens to background and re-foreground the app. Only then does `AuthProvider.checkAuth` see the missing token and redirect to login. It feels like the app is broken, not logged out.

### Verified mechanism
`authenticatedFetch`'s 401 handler (`auth-interceptor.ts:61-71`) forces a `deduplicatedRefresh`; if refresh fails it calls `await signOut()` — but this is the **library** `signOut` imported at line 2 (`auth.ts:77-88`), which only does a best-effort revoke + `clearTokens()`. It never touches the provider. The provider's own `signOut` (`auth-provider.tsx:107-125`) is the one that flips `isAuthenticated=false`, resets analytics, clears the session/active-board stores, disposes the WS client, resets the HTTP client, and calls `queryClient.clear()`. Because the interceptor bypasses all of that: `isAuthenticated` stays true (app keeps rendering authenticated screens), tokens are gone (every subsequent request sends no `Authorization` header and gets fresh 401s), the WS stays connected on the revoked token, and the PostHog identity stays the old user. `graphql/client.ts` wires `authenticatedFetch` as the singleton client's fetch, so this is the main data path. The structural cause: the lib interceptor cannot import the provider. The AppState `active` re-check auto-heals on the next background→foreground cycle, bounding the broken state to the rest of the current foreground session.

### Repro
Sign in on device. Revoke the refresh token server-side (or rotate it from another flow) and wait out the 60s backend validation cache so the JWT 401s. With the app foregrounded, trigger any GraphQL request (pull-to-refresh Climbs). `authenticatedFetch` gets 401 → refresh 401 → lib `signOut` clears tokens. The UI stays on the Climbs tab; every subsequent request goes out unauthenticated and fails. Background and re-foreground → only then does `checkAuth` see the missing token and redirect to `/auth/login`.

### Verified fix design
Add a module-level revocation hook to `auth-interceptor.ts`: `let onForcedSignOut: (() => void) | null = null; export function setOnForcedSignOut(cb)`, and invoke it after `await signOut()` in the 401 branch. In `auth-provider.tsx`, register it in a `useEffect` that runs the provider's full signed-out cleanup (factor the cleanup body of `signOut` into a shared helper — the same helper #003 needs — minus the Logout analytics event, plus `resetAnalytics`). Extend `auth-interceptor.test.ts` to assert the callback fires on a failed-refresh 401.

### Why deferred
P2 with medium fix risk — wiring a cross-module callback into the auth cleanup path touches shared sign-out semantics and pairs with #003's shared helper. The AppState re-check auto-heals, and the trigger (mid-session refresh failure) is rare.

marcodejongh:
## What

When the server revokes a mobile session (password change, admin revoke, refresh-token rotation failure), the app stayed on the authenticated screens with a dead token. `authenticatedFetch`'s 401 handler only called the **library** `signOut` (`auth.ts` — best-effort revoke + `clearTokens`), which never touches the `AuthProvider`. Result: `isAuthenticated` stayed `true`, every later request went out with no `Authorization` header and 401'd (empty lists, spinners, "something went wrong"), the WS stayed connected on the revoked token, and PostHog kept the old identity — until the next background→foreground `checkAuth` happened to notice the missing token and redirect to login.

The structural cause: the lib-layer interceptor can't import the provider.

## Fix

- **`auth-interceptor.ts`** — add a module-level `onForcedSignOut` hook + `setOnForcedSignOut(cb)` setter, invoked right after the existing `await signOut()` in the unrecoverable 401 branch. It only fires on the failed-refresh path (the recoverable retry `return`s first), so a successful refresh+retry never triggers it.
- **`auth-provider.tsx`** — factor the provider's signed-out cleanup into a shared `runSignedOutCleanup` helper (analytics reset, store clears, active-board cache drop, HTTP client reset, WS dispose, `queryClient.clear()`, `setIsAuthenticated(false)`). The manual `signOut` reuses it; a `useEffect` registers it against the interceptor hook. The helper deliberately omits the two caller-specific steps — the manual `Logout` analytics event, and `authSignOut()` (the token revoke the forced path's caller already ran) — so the forced path never double-revokes.

Net result on a revoked session: tokens cleared → provider cleanup runs → `setIsAuthenticated(false)` → immediate `<Redirect href="/auth/login">`, WS torn down, caches cleared. Same end state as a manual logout, minus the manual `Logout` event.

## Scope

Addresses **only** #2684. `runSignedOutCleanup` is the shared piece the sibling audit item is noted to reuse; no further audit work is included here.

## Tests

Extended `auth-interceptor.test.ts`:
- failed-refresh 401 fires the forced-sign-out hook exactly once, after the token revoke;
- successful 401→refresh→retry does **not** fire it;
- no-token 401 does **not** fire it.

## Validation

- `vp run typecheck:mobile` ✅
- `vp run test:mobile` ✅ (1448 passed; auth-interceptor: 14 passed)
- `vp run check:mobile-bundle` ✅ (10.6 MB)
- `vp check` ✅ (changed files clean)

Fixes #2684

### DIFF
--- a/packages/mobile/src/lib/__tests__/auth-interceptor.test.ts
+++ b/packages/mobile/src/lib/__tests__/auth-interceptor.test.ts
@@ -232,6 +232,39 @@ describe('authenticatedFetch', () => {
     expect(onForcedSignOut).toHaveBeenCalledTimes(1);
   });
 
+  it('still runs the captured cleanup when the hook is cleared mid sign-out (provider unmount race)', async () => {
+    // signOut() awaits a network revoke. If AuthProvider unmounts during that
+    // window it nulls the hook — but the cleanup it registered must still fire,
+    // or the forced sign-out silently drops in exactly the case this guards.
+    mockIsTokenExpiringSoon.mockResolvedValue(false);
+    mockGetAuthToken.mockResolvedValue('old-jwt');
+    const onForcedSignOut = vi.fn();
+    setOnForcedSignOut(onForcedSignOut);
+
+    // Hold signOut open so we can clear the hook (simulating unmount) before it
+    // resolves and the callback is invoked.
+    let releaseSignOut!: () => void;
+    mockSignOut.mockReturnValueOnce(
+      new Promise<void>((resolve) => {
+        releaseSignOut = resolve;
+      }),
+    );
+
+    mockFetch
+      .mockResolvedValueOnce({ status: 401, ok: false }) // initial 401
+      .mockResolvedValueOnce({ ok: false, status: 403 }); // refresh fails
+
+    const fetchPromise = authenticatedFetch('https://api.example.com/data');
+    await vi.waitFor(() => expect(mockSignOut).toHaveBeenCalledTimes(1));
+
+    // Provider unmounts mid sign-out, then the revoke resolves.
+    setOnForcedSignOut(null);
+    releaseSignOut();
+    await fetchPromise;
+
+    expect(onForcedSignOut).toHaveBeenCalledTimes(1);
+  });
+
   it('handles a failed-refresh 401 without throwing when no forced-sign-out hook is registered', async () => {
     // The interceptor can fire before the provider mounts (or after it unmounts),
     // leaving onForcedSignOut null. The optional-chain call must be a no-op.
--- a/packages/mobile/src/lib/__tests__/auth-interceptor.test.ts
+++ b/packages/mobile/src/lib/__tests__/auth-interceptor.test.ts
@@ -13,7 +13,6 @@ vi.mock('../auth-store', () => ({
   getAuthToken: vi.fn().mockResolvedValue('test-jwt'),
   getRefreshToken: vi.fn().mockResolvedValue('test-refresh-token'),
   storeTokens: vi.fn().mockResolvedValue(undefined),
-  clearTokens: vi.fn().mockResolvedValue(undefined),
   isTokenExpiringSoon: vi.fn().mockResolvedValue(false),
   getTokenExpiresAt: vi.fn().mockResolvedValue(null),
 }));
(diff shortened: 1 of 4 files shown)


marcodejongh:
Thanks — addressed all three in `ef0e2a2`:

1. **Silent drop of `setIsAuthenticated(false)` on cleanup failure** — fixed at the source: `runSignedOutCleanup` now uses `Promise.allSettled` for the best-effort `clearStoredSessionId`/`clearStoredActiveBoard` calls, so a rejected SecureStore/AsyncStorage delete can't abort the teardown. That guarantees the critical state flip runs on *both* the forced and manual paths (the manual path had the same latent risk). The forced callback also `.catch(reportError)`s rather than bare-`void`, so any unexpected rejection is reported, not swallowed.
2. **Missing forced-logout analytics** — the forced callback now emits `track(Logout, { method: 'forced' })`, keeping `runSignedOutCleanup` method-agnostic (each caller emits its own event, matching the existing `'manual'` / `'account_deleted'` methods). Server-revoked sign-outs are now distinguishable in PostHog.
3. **Null-hook test gap** — added a test for the failed-refresh 401 with `onForcedSignOut` unset (no-op optional chain), and dropped a pre-existing unused mock while in the file.

Full mobile suite green (1537 passed).

marcodejongh:
Second round addressed in `f124de6`:

1. **Concurrent 401s firing `onForcedSignOut` N times** — good catch; this only became a real problem once I added the `forced` Logout event. Collapsed the forced sign-out into a single in-flight promise (`forceSignOut`, mirroring the existing `deduplicatedRefresh` pattern), reset on settle so a later genuine 401 still signs out. N concurrent failures now produce one revoke + one cleanup + one event.
2. **Concurrent-failure test** — added; asserts one `signOut` and one callback across two simultaneous 401s. Keyed the fetch mock on URL (not call order) so it can't flake on interleaving.
3. **Orphaned `clearTokens` mock entry** — removed from the factory.
5. **Unmount-registration test** — added to `auth-provider.test.tsx`: registers a callable on mount, clears it (`null`) on unmount.

**4. `runSignedOutCleanup` calling `resetAnalytics()` unconditionally** — leaving as-is by design. The guarded `resetAnalyticsForSignedOutTransition` exists for the `checkAuth` path, where the transition can be a no-op (already signed out). `runSignedOutCleanup` only runs on a deliberate sign-out from an authenticated session — the original manual `signOut` reset unconditionally too — and with the dedup above it now runs once per forced sign-out, so the repeat-reset concern is gone. Happy to add the guard if you'd prefer the symmetry.

Full mobile suite green (1539 passed).

marcodejongh:
Addressed the unmount race in `754edf9` — agreed it's narrow but worth closing.

`forceSignOut` now captures the hook reference *before* `await signOut()`, so if the provider unmounts during the revoke its effect can null the module ref without dropping the cleanup the sign-out already committed to (WS dispose, http reset, cache clear still run). Added a regression test that holds `signOut` open, clears the hook mid-flight, then resolves and asserts the captured callback still fires once.

Full mobile suite green (1540 passed).

marcodejongh:
Added the behavioural coverage in `2c3ab81`. The forced-path test now fires the registered hook and asserts the cleanup actually runs — store clears, http reset, WS dispose, and full cache wipe each once — and that `authSignOut` is **not** called (the interceptor already revoked, so no double-revoke). That gives the forced path behavioural coverage on top of the registration wiring.

Full mobile suite green (1541 passed).

PR #2693, opened 2026-06-10
Verify session liveness before cold-start rejoin (#2683)

### LINKED ISSUE #2683: [P2] Cold start resurrects a server-ended party session (stored sessionId restored without liveness check)
**User impact:** After a party session ends (host ends it, or it times out for inactivity) and the climber later relaunches the app, they are silently dropped back into a dead/recreated session on a possibly-wrong board. They believe they're sharing a wall with their crew when no one is there, and queue mutations broadcast into a zombie room. The only escape is to manually tap End Session. Ticks logged during this state are stamped after the session's `endedAt`, breaking duration/summary semantics.

### Verified mechanism
On provider mount `queue-provider.tsx:507-514` restores the persisted SecureStore session id and calls `setSessionId` unconditionally. The WS-subscription effect (541-767) then fires `ensureJoined` → `JOIN_SESSION` with no liveness check. On the backend, `restoreFromPostgres` (`session-restoration.ts:87`) and the no-Redis branch (`client-lifecycle.ts:154`) both return `isNewSession=true` when `pgSession.status === 'ended'`, recreating an empty in-memory/Redis room while the PG row stays `status='ended'`. `SessionScreen.tsx:35` gates purely on `sessionId !== null`, so the client lands in `InSessionView`. `JOIN_SESSION` failures are `__DEV__` console-warn only (`logJoinFailure`, 565-567), so a genuine failure never clears the stale id. The inactivity sweep (60-min threshold, `room-manager.ts:42`) ends sessions and publishes `SessionEnded` that a killed app never receives, so the stored id goes stale routinely. `GET_SESSION` (with `endedAt`) exists in `operations.ts` but is only used by the join-confirmation preview, never on restore. Git history (44c7294d8 and later) shows no decision to skip liveness validation — it's a gap, not an intentional guard.

### Repro
1. Add a climb to the queue or start a party session on device A (persists the id via `setStoredSessionId`).
2. Force-quit the app.
3. End the session: host on device B taps End Session, OR wait >60 min for the backend inactivity sweep. The `SessionEnded` broadcast is missed because the app is dead.
4. Relaunch the app. The restore effect sets `sessionId`, `JOIN_SESSION` recreates an empty room, and the Record tab shows `InSessionView` for the dead session with no peers. No `SessionEnded` ever arrives and the stored id is never cleared.

### Verified fix design
In the `queue-provider.tsx` restore effect, after a non-null `storedId`, request `GET_SESSION` (already defined in `operations.ts` with `endedAt`). If `response.session` is null or `endedAt != null`, `await clearStoredSessionId()` and skip `setSessionId`. On request failure (offline cold start), keep today's optimistic restore so offline relaunch still works. Add a provider test mocking the `GET_SESSION` response with `endedAt` set, asserting no subscribe and stored id cleared.

### Why deferred
P2 with medium fix risk — the change touches restore-path semantics (offline cold start must still work optimistically), so it needs careful handling rather than a quick patch. Escapable via End Session, and solo implicit sessions make in-session-on-relaunch partially the designed norm.

marcodejongh:
Fixes #2683.

## Problem
On mobile cold start, the persisted party-session id was restored from SecureStore and rejoined **without any liveness check** (`queue-provider.tsx` restore effect → WS effect → `JOIN_SESSION`). If the session ended while the app was dead — the host tapped **End Session**, or the backend's 60-min inactivity sweep ended it (the `SessionEnded` broadcast is missed by a killed app) — `JOIN_SESSION` recreated the server-ended room as an empty zombie. The climber landed in `InSessionView` with no peers, broadcasting queue mutations into a dead room, with ticks stamped after the session's `endedAt`. The only escape was to manually tap **End Session**.

## Fix (mobile client only)
The restore effect now fires a `GET_SESSION` liveness check (HTTP, mirroring the existing join-confirmation preview in `use-session-detail.ts`) before setting session state:

- **ended (`endedAt != null`) or missing session** → clear the stored id, do not rejoin → Record tab shows `PreSessionView`.
- **live session** → restore as before (brief one-RTT `PreSessionView` flash is acceptable).
- **request failure (offline cold start)** → optimistic restore, so the queue still comes back offline.

`GET_SESSION` / `GetSessionQueryResponse` already existed in `operations.ts`; `getHttpClient` was already imported.

Backend hardening (rejecting joins to `status='ended'` sessions instead of recreating them) is a real but separate latent gap, intentionally **out of scope** here — the client check fully addresses the reported cold-start user impact.

## Tests
- New: **ended session is dropped on cold start** — asserts stored id cleared, `sessionId` stays null, no join, no subscription.
- New: **offline optimistic restore** — `GET_SESSION` rejects → `sessionId` still restored, join fires, stored id not cleared.
- Updated the suite's default `http.request` mock to return an alive `GET_SESSION` so existing auto-restore tests stay green; refactored the `session-store` mock to a hoisted handle.

## Validation
- `vp run typecheck:mobile` ✓
- `vp run test:mobile` — 1449 tests pass (incl. 2 new) ✓
- `vp run check:mobile-bundle` — OK ✓
- `vp check` — lint + format clean on changed files ✓

Behaviour is hard to exercise in CI (needs a real ended session + force-quit), so it's covered by unit tests; device-QA steps are in `.boardsesh/qa-notes.md` (ended-by-host, ended-by-sweep, live restore, solo implicit session, offline relaunch).

### DIFF
--- a/packages/mobile/src/providers/__tests__/queue-provider-leave-session.test.tsx
+++ b/packages/mobile/src/providers/__tests__/queue-provider-leave-session.test.tsx
@@ -126,7 +126,12 @@ describe('QueueProvider clearSession notifyServer', () => {
     sessionStore.getStoredSessionId.mockResolvedValue('session-1');
     sessionStore.clearStoredSessionId.mockClear();
     http.request.mockReset();
-    http.request.mockResolvedValue({});
+    // Cold-start restore verifies session liveness via GET_SESSION (#2683); keep
+    // the stored session alive so these tests land in-session before clearSession.
+    http.request.mockImplementation(async (operation: unknown) => {
+      const operationText = typeof operation === 'string' ? operation : '';
+      return operationText.includes('GetSession') ? { session: { id: 'session-1', endedAt: null } } : {};
+    });
     graph.execute.mockReset();
     // joinSession resolves the active session; leaveSession resolves true.
     graph.execute.mockResolvedValue({
--- a/packages/mobile/src/providers/__tests__/queue-provider-session-updates.test.tsx
+++ b/packages/mobile/src/providers/__tests__/queue-provider-session-updates.test.tsx
@@ -32,6 +32,12 @@ const http = vi.hoisted(() => ({
   request: vi.fn(),
 }));
 
+const sessionStore = vi.hoisted(() => ({
+  getStoredSessionId: vi.fn(async () => 'session-1' as string | null),
+  setStoredSessionId: vi.fn(async () => {}),
+  clearStoredSessionId: vi.fn(async () => {}),
+}));
+
 const activeBoard = vi.hoisted(() => ({
   stored: {
     uuid: 'board-1',
@@ -113,11 +119,7 @@ vi.mock('../../lib/graphql/ws-client', () => ({
   getWsClient: () => ws.client,
 }));
 
-vi.mock('../../lib/session-store', () => ({
-  getStoredSessionId: vi.fn(async () => 'session-1'),
-  setStoredSessionId: vi.fn(async () => {}),
-  clearStoredSessionId: vi.fn(async () => {}),
-}));
+vi.mock('../../lib/session-store', () => sessionStore);
 
 vi.mock('../../lib/active-board-store', () => ({
   getStoredActiveBoard: activeBoard.getStoredActiveBoard,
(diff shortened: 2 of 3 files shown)


marcodejongh:
Thanks for the reviews. Triage of the feedback:

**Codex P1 — clearing on `session === null` would drop live solo sessions.** Confirmed and valid. The `session` query is presence-gated (`getSessionUsers().length === 0 → null`), so it can't distinguish an *ended* session from a *live-but-empty* solo session whose socket dropped on force-quit — my `!session → clear` branch would wipe an active solo session on cold start. Fixing this in a follow-up PR by swapping the presence-based `GET_SESSION` check for a **DB-backed `sessionLiveness` read** of `board_sessions.status`/`endedAt` (a read-only SELECT, presence-independent, can't resurrect anything). That also clears up Claude's #1/#2 — the new op name `SessionLiveness` doesn't share the `GetSession` prefix, so the brittle substring routing in the test mocks goes away.

**Claude #3 — missing `cancelled` guard after `clearStoredSessionId()`.** No live bug today (the branch returns immediately); will keep the guard in place in the follow-up.

**Claude #4 / Codex note — backend `JOIN_SESSION` zombie-join.** The client guard only protects mobile; the server still recreates ended rooms for any client (web included). Tracked as #2696 to harden the join/restore path itself (`status='ended'` → don't recreate).

Follow-up PR will resolve the Codex P1 and Claude #1–#3; #2696 covers the server-side defense.

PR #2695, opened 2026-06-10
fix(mobile): render playlist climbs against the playlist's own board (#2689)

### LINKED ISSUE #2689: [P2] Playlist detail renders climbs against the active board, not the playlist's own board, when they differ
**User impact:** A playlist opened while the active board differs from the playlist's own board shows climbs rendered for the wrong board and, on tap, queues/opens them on the wrong board — wrong holds light up, wrong board art. An edge case, but a real correctness bug when it hits. A related variant: reaching Discover with no active board (signed-out / not-yet-onboarded) leaves `boardConfig` null, so every row returns null and the climb list is blank under a non-zero hero count.

### Verified mechanism
`PlaylistDetailView.renderItem` binds every `ClimbListRow` to `useDrawerHost().boardConfig` — the user's **active** board (`PlaylistDetailView.tsx:202-217`, `drawer-host-provider.tsx:123-133`) — and `usePlaylistActivation.resolveTarget` builds the queue target from the active board (`use-playlist-activation.ts:91-109`). The detail route never compares `playlist.boardType`/`layoutId` against the active board and fetches climbs all-boards (`[playlist_uuid].tsx:70`; the playlist board is used only for the hero backdrop at 337-339). A mismatch arises via deep link, switching the active board on `/boards` then returning to the still-mounted detail screen, or unscoped community browsing with no active board. The Discover/all entry points scope playlist lists to the active board, masking this for normal in-app navigation but not for deep links or mid-session board switches. The binding is documented intent (the docstring says rows are "bound to the user's active board"), so this reads as a known design simplification whose edge cases were deferred.

### Repro (candidate — needs device verification)
Active board = Tension; Discover → open a Tension playlist (detail screen now on the stack) → push `/boards` and switch the active board to Kilter → navigate back to the still-mounted detail screen. `boardConfig` context updates → rows re-render Tension climbs against Kilter `boardName`/`layoutId`/`sizeId`/`setIds` (wrong board art/holds), and tapping a climb queues it with the Kilter board key + angle. Whether onboarding/deep-link flows make these states reachable in practice needs runtime verification.

### Fix design (needs a board-context decision)
Playlists carry `boardType`/`layoutId` but no `sizeId`/`setIds`/`angle`, so "render against the playlist's own board" is underdetermined. Smallest viable fix is a gate, not a re-derive: in `discover/[playlist_uuid].tsx`, compare `playlist.boardType`/`layoutId` to the active board and, on mismatch (or null `boardConfig`), disable row activation and show a "Switch to <board> to climb this playlist" prompt (`router.push('/boards')`). A full re-derive needs a size/set resolution policy from `@boardsesh/board-config` — a larger product call.

### Why deferred (needs human)
The mechanism is real, but it's plausibly an accepted design simplification rather than an unnoticed regression, the triggers' reachability needs device verification, and the fix requires deciding which board context playlist rows should use.

marcodejongh:
## Problem

Fixes #2689. On mobile, the playlist-detail screens rendered every climb row, and queued every tapped climb, against the user's **active** board rather than the playlist's own board. When they differ (deep link, switching the active board mid-session, or no active board when signed-out), this caused real correctness bugs:

- **Wrong board art / holds / grades** — a Tension playlist rendered against Kilter art.
- **Queued onto the wrong board** — tapping lit the wrong wall over BLE.
- **Blank list under a non-zero count** — no active board → every row returned `null`.

## Fix

Render the climbs against the **playlist's own board**, mirroring what web already does. The repo already had the resolution helper (`getBoardConfigForPlaylist` — largest size + all sets, used today for the hero backdrop), so this is a small, board-correct change rather than a new policy.

- **New hook `usePlaylistRenderBoard`** decides the render board: the active board when it matches the playlist (tap queues normally), otherwise the playlist's own resolved board with a switch-board banner.
- **`PlaylistDetailView`** now takes `renderBoard` + `boardBanner` instead of reading the active board itself. On a mismatch the list goes **read-only**: rows render at each climb's own angle (the angle its grade was baked at — the thumbnail doesn't depend on angle), a banner prompts switching boards, and queueing (row tap + activate-all) is disabled. Tapping a read-only row or the banner CTA routes to `/boards`.
- **Smart playlists** are board-relative, so they always render against the active board with no banner.

Queueing stays gated on a mismatch because the queue, play drawer, and BLE LEDs all follow the single active board — you genuinely must switch boards to climb another board's playlist.

## Validation (Linux)

- `vp run typecheck:mobile` ✅
- `vp run test:mobile` — 218 files / 1547 tests ✅ (new `use-playlist-render-board` hook test + extended `playlist-detail-view` banner tests; updated the route test's mocks)
- i18n catalog parity (`catalog-completeness`) ✅ — new `detail.boardMismatch.*` keys added to en-US / es / fr
- `vp check` (lint + format) ✅ — 0 errors

## QA on device (macOS/sim)

1. Active board = Tension. Open a **Kilter** playlist (deep link, or switch the active board to Tension while a Kilter playlist detail is on the stack). Expect: rows show **Kilter** thumbnails/grades, a "Switch to Kilter to climb this playlist" banner above the list, and neither row taps nor the play-all action queue anything.
2. Tap the banner CTA (or any row) → lands on `/boards`. Switch to Kilter → reopen the playlist → rows render normally and tapping queues/lights the Kilter wall.
3. Signed-out: open a public playlist with no active board → read-only render + banner; CTA → `/boards`.
4. Matching board (active board == playlist board) → unchanged behavior; tapping queues and lights normally.

### DIFF
--- a/packages/mobile/app/(tabs)/discover/[playlist_uuid].tsx
+++ b/packages/mobile/app/(tabs)/discover/[playlist_uuid].tsx
@@ -28,6 +28,7 @@ import {
 import { GlassIconButton } from '../../../src/components/GlassIconButton';
 import { getHttpClient } from '../../../src/lib/graphql/client';
 import { usePlaylistActivation } from '../../../src/lib/playlists/use-playlist-activation';
+import { usePlaylistRenderBoard } from '../../../src/lib/playlists/use-playlist-render-board';
 import { recordPlaylistOpen } from '../../../src/lib/playlists/recents-store';
 import { toQueueClimbs } from '../../../src/lib/climb-types';
 import { hapticSelection } from '../../../src/lib/haptics';
@@ -114,6 +115,14 @@ export default function PlaylistDetail() {
     fetchPage,
     refreshErrorMessage: 'Failed to refresh playlist suggestions:',
   });
+
+  // Render the climbs against the playlist's own board. When it matches the
+  // active board this is the active board (tapping queues normally); when it
+  // differs (or there's no active board) rows render read-only against the
+  // playlist's board and `boardBanner` prompts a switch.
+  const { renderBoard, banner: boardBanner } = usePlaylistRenderBoard(
+    playlist ? { boardType: playlist.boardType, layoutId: playlist.layoutId } : null,
+  );
 
   const isOwner = playlist?.userRole === 'owner';
   const isFollowable = !!playlist?.isPublic && !isOwner;
@@ -414,6 +423,8 @@ export default function PlaylistDetail() {
       <PlaylistDetailView
         hero={hero}
         climbs={allClimbs}
+        renderBoard={renderBoard}
+        boardBanner={boardBanner}
         isLoading={query.isLoading}
         isFetchingNextPage={query.isFetchingNextPage}
         hasNextPage={query.hasNextPage ?? false}
--- a/packages/mobile/app/(tabs)/discover/__tests__/playlist-uuid.test.tsx
+++ b/packages/mobile/app/(tabs)/discover/__tests__/playlist-uuid.test.tsx
@@ -66,6 +66,9 @@ vi.mock('../../../../src/providers/theme-provider', () => ({
 vi.mock('../../../../src/providers/auth-provider', () => ({ useAuth: () => ({ isAuthenticated: true }) }));
 vi.mock('../../../../src/providers/toast-provider', () => ({ useToast: () => ({ showToast: vi.fn() }) }));
 vi.mock('../../../../src/lib/playlists/use-playlist-activation', () => ({ usePlaylistActivation: () => vi.fn() }));
+vi.mock('../../../../src/lib/playlists/use-playlist-render-board', () => ({
+  usePlaylistRenderBoard: () => ({ renderBoard: null, banner: null }),
+}));
 vi.mock('../../../../src/lib/playlists/recents-store', () => ({ recordPlaylistOpen: vi.fn() }));
 vi.mock('../../../../src/lib/climb-types', () => ({ toQueueClimbs: (climbs: unknown) => climbs }));
 vi.mock('../../../../src/lib/haptics', () => ({ hapticSelection: vi.fn() }));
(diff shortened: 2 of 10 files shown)


marcodejongh:
Thanks for the review. Addressed in c22246633:

**2. `router`/`t` in `useMemo` deps → row re-renders** — Fixed. Split `usePlaylistRenderBoard` so `renderBoard` is resolved from board state only (`activeBoard`/`boardType`/`layoutId`) and the banner (which closes over `t`/`router`) lives in a separate memo. `PlaylistDetailView`'s `renderItem` now depends on a `readOnly` boolean instead of the `boardBanner` object, and `handleSwitchBoard` reads the banner through a ref — so even if `t`/`router` churn, neither `renderBoard` nor the row callbacks change identity, and the FlashList rows don't re-render.

**1. Liquid Glass activate-all not hidden** — Not applicable: there's no activate-all control in the Liquid Glass variant. `handleActivateAll` is wired only to the Material `Appbar.Action icon="play"` (now `climbs.length > 0 && !boardBanner`). The Glass variant's `actions` prop renders follow/pin/more, not play-all. Read-only rows in either variant route to `/boards` on tap (`handleSwitchBoard`), so there's no dead interaction. The `if (readOnly) return` in `handleActivateAll` is just a guard.

**3. Route test mocks `renderBoard: null`** — No assertion lost: `playlist-uuid.test.tsx` fully mocks `PlaylistDetailView` (it only reads `hero.name`), so `renderBoard` is never exercised there. Row rendering against a resolved board is covered by the `usePlaylistRenderBoard` hook test and the `playlist-detail-view` banner tests.

marcodejongh:
Thanks — agreed all three are non-blocking. Quick notes:

1. **Hardcoded `returnTo: '/(tabs)/discover'`** — correct for both current callers (the regular + smart playlist detail screens both live under the Discover tab). I left it hardcoded rather than add a speculative param; happy to thread `returnTo` through if/when the hook gets a profile/search caller.

2. **`bannerNode`/`ListHeaderComponent` identity** — as noted, rows are unaffected (they key off `renderItem`, which is now decoupled from banner identity). The header element was already rebuilt inline each render before this PR, so the banner doesn't change that characteristic. Left it as-is to avoid touching the pre-existing header pattern; can wrap the header+banner in a `useMemo` as a follow-up if we want to tidy it.

3. **Route test sees `renderBoard: null`** — intentional: `playlist-uuid.test.tsx` mocks `PlaylistDetailView` entirely (it only asserts the error/not-found/retry states), so the real-board path is exercised by the `usePlaylistRenderBoard` hook test + the `playlist-detail-view` banner tests rather than end-to-end through the route.

Issue #2696, opened 2026-06-10
Backend: JOIN_SESSION resurrects server-ended sessions (zombie-join)


marcodejongh:
## Problem

`JOIN_SESSION` recreates a server-**ended** party session as a fresh, empty in-memory/Redis room instead of refusing to restore it. The Postgres `board_sessions` row stays `status='ended'` (with `endedAt` set), but the join path treats "not in Redis" as "new session" without consulting that status, so any client that joins a stale/ended session id resurrects a zombie room.

### Verified mechanism
- `restoreFromPostgres` (`packages/backend/src/services/room-manager/session-restoration.ts`) returns `isNewSession=true` when `pgSession.status === 'ended'` (it only restores rooms whose status is **not** ended), so the caller proceeds to create a fresh room.
- The no-Redis branch in `client-lifecycle.ts` does the same.
- `board_sessions.status` ('active' | 'ended') + `endedAt` are the durable source of truth (`packages/db/src/schema/app/sessions.ts`); live presence is Redis-only.

### Impact
Surfaced by the mobile cold-start bug #2683 (PR #2693). The mobile **client** now guards its own restore path, but the **server-side hole is still open**: the web app (or any other client) that issues `JOIN_SESSION` against an ended session id will still recreate a zombie room. Queue mutations broadcast into a dead room and ticks get stamped after the session's `endedAt`.

## Proposed fix
In the join/restore path, when the Postgres session row exists with `status='ended'` (or `endedAt != null`), **do not recreate** the room. Either:
- reject the join with a clear "session ended" error, or
- return the ended session payload (with `endedAt` populated, not a fresh `endedAt: null` room) so clients can drop out gracefully.

Add a backend test asserting that joining an ended session does not create a new room and surfaces the ended state.

## Notes
- This is the backend half of #2683. The client-side liveness handling is addressed separately (DB-backed `sessionLiveness` read on cold start); this issue covers defending the `JOIN_SESSION` mutation itself so all clients are protected.
- Flagged by automated review on PR #2693 (Codex + Claude review).

PR #2699, opened 2026-06-10
DB-backed sessionLiveness for cold-start restore (follow-up to #2693)

marcodejongh:
Follow-up to #2693, addressing the review feedback there. Stacked on `claude/cold-start-session-liveness-2683` (base will retarget to `main` once #2693 merges).

## Why
Codex's P1 on #2693 was correct: that PR's cold-start check used the **presence-gated** `session` query, which returns `null` whenever the live roster is empty. That's true for an *ended* session **and** for a *dormant-but-active solo* session whose socket dropped on force-quit — so `!session → clear` would wipe an active solo session on every cold start. The two cases are indistinguishable through `session`, so the fix needs a presence-independent read of the durable session row.

## What
- **New `sessionLiveness(sessionId): SessionLiveness` query** (backend + schema). Reads `board_sessions.status`/`endedAt` directly — a plain `SELECT`, no room-manager / Redis involvement, so it can't resurrect anything. No auth: it exposes only existence + ended-state, and auth may not be restored yet at cold start.
- **Client restore** now clears the stored id only on a *positive* ended/missing signal (`status === 'ended'`, `endedAt` set, or no row) and restores active sessions — including empty solo ones. Offline cold start still restores optimistically.
- Replaces the `GET_SESSION` check from #2693. `GET_SESSION` stays for the join-confirmation preview (`use-session-detail.ts`).

## Review feedback resolved
- **Codex P1** (false-clear of live solo sessions) — fixed by the DB-backed read.
- **Claude #1/#2** (brittle `includes('GetSession')` mock routing + inconsistent param typing) — the new `SessionLiveness` op name shares no prefix with `GetSessionQueueState`, so the substring routing is unambiguous and the mock typing is now consistent across both harnesses.
- **Claude #3** (`cancelled` guard) — retained.
- **Claude #4 / backend zombie-join** — tracked separately in #2696 (server still recreates ended rooms for any client; out of scope here).

## Tests / validation
- New backend test: `sessionLiveness` returns active / ended (ISO `endedAt`) / null-when-missing, and rejects a malformed id before querying.
- Mobile provider tests updated to the new op; ended-session-dropped + offline-optimistic-restore still covered.
- `vp run typecheck` (all packages) ✓ · `vp run test:mobile` 1537 ✓ · backend `sessionLiveness` test ✓ · `vp run check:mobile-bundle` OK ✓ · `vp run codegen` committed (codegen-drift clean) ✓

### DIFF
--- a/packages/backend/src/__tests__/session-liveness-query.test.ts
+++ b/packages/backend/src/__tests__/session-liveness-query.test.ts
@@ -1,0 +1,69 @@
+/**
+ * Tests for the sessionLiveness query (#2683).
+ *
+ * sessionLiveness reads the durable session row so an ended session reads as
+ * ended even when no participants are connected — unlike `session`, which
+ * returns null for any empty roster and so can't tell an ended session apart
+ * from a dormant-but-active solo session. Clients call it on cold start to
+ * decide whether to restore or drop a persisted session id.
+ */
+
+import { describe, it, expect, vi, beforeEach } from 'vite-plus/test';
+
+const dbClient = vi.hoisted(() => {
+  const limit = vi.fn();
+  return {
+    limit,
+    db: {
+      select: vi.fn().mockReturnThis(),
+      from: vi.fn().mockReturnThis(),
+      where: vi.fn().mockReturnThis(),
+      limit,
+    },
+  };
+});
+
+// Mock connection-opening modules so importing the resolver barrel is side-effect-free.
+vi.mock('../services/room-manager', () => ({ roomManager: {} }));
+vi.mock('../pubsub/index', () => ({ pubsub: {} }));
+vi.mock('../services/distributed-state', () => ({ getDistributedState: () => null }));
+vi.mock('../db/client', () => ({ db: dbClient.db, dbRead: dbClient.db }));
+
+import { sessionQueries } from '../graphql/resolvers/sessions/queries';
+
+describe('sessionLiveness query', () => {
+  beforeEach(() => {
+    dbClient.limit.mockReset();
+  });
+
+  it('returns active liveness for a live session row', async () => {
+    dbClient.limit.mockResolvedValue([{ id: 'session-1', status: 'active', endedAt: null }]);
+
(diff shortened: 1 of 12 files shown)


marcodejongh:
Addressed in a879591:

1. **`dbRead` instead of `db`** — switched. Agreed it's a plain read-only lookup; the read-your-writes window here is sub-second and backstopped by #2696, so it follows the read-path convention.

2. **`status` type safety** — typed as `SessionStatus = 'active' | 'inactive' | 'ended'` (shared-schema), so the `=== 'ended'` comparison is now compile-checked on both client and server. One correction to the suggestion: the value set is **three**, not two — `board_sessions` has a DB CHECK `status IN ('active','inactive','ended')`. `'inactive'` is a dormant-but-not-ended state that should restore, so narrowing to `'active' | 'ended'` would have dropped a valid value. I kept the **GraphQL** field as `String!` rather than an enum: `status` is a free-text column (CHECK-constrained, not a pg enum), and a GraphQL enum throws at serialization on any value outside the enum — `String!` degrades gracefully if a new status is ever added.

3. **Backend test `beforeEach`** — now `mockClear()`s the `select`/`from`/`where` chain stubs too (kept as `mockClear`, not `mockReset`, so their `mockReturnThis` impl survives).

4. **Missing `status: 'ended'` + `endedAt: null` coverage** — added a mobile test for that partial-ended row, exercising the `status` arm of the clear condition independently of the `endedAt` arm.

marcodejongh:
Note: the red `claude-review` check is an action-side auth flake, not a code finding — the job dies in ~30s (vs its normal ~2m17s) with `App token exchange failed: 401 Unauthorized` / `Workflow validation failed` / `Internal error: directory mismatch ... this indicates a bug`, and fails identically on re-run. Every substantive check is green (typecheck, lint, check, test-backend, test-default, mobile-bundle, codegen-drift, build-and-test). The earlier successful claude-review run is the one whose findings are addressed above.

PR #2709, opened 2026-06-10
Add account deletion to mobile settings

marcodejongh:
## Why

Apple App Store review (Guideline 5.1.1(v)) requires **in-app account deletion** for any app that supports account creation. The mobile app had Sign Out but no delete path, so it would fail review. The web app already ships a reviewed deletion flow — this ports it to React Native and reuses the existing pieces rather than duplicating them.

## What's reused (no changes)

- **Backend** `deleteAccount` mutation/resolver — deletion behaviour is identical to web because both call the same mutation (deletes draft climbs, optionally strips the setter name from published climbs, deletes the user row; published climbs preserved via `onDelete: set null`).
- **Shared GraphQL operations** — `@boardsesh/graphql/operations/account` (`GET_DELETE_ACCOUNT_INFO`, `DELETE_ACCOUNT`), the exact module web imports.
- **Shared i18n keys** — `settings.deleteAccount.*`, already present in `en-US`/`es`/`fr`. No new catalog entries.

## New mobile code

- `src/lib/graphql/hooks/use-delete-account.ts` — `useDeleteAccountInfo` (published-climb count) and `useDeleteAccount` (mutation), following the existing `getHttpClient().request()` + react-query pattern. Re-exported from the hooks barrel.
- `src/components/DeleteAccountScreen.tsx` — faithful replica of the web dialog: warning, "Checking your climbs…", published-climb count + "remove setter name" toggle (only when count > 0), type-`DELETE`-to-confirm field, destructive confirm disabled until confirmed. Calls `signOut('account_deleted')` on success; shows the server message (or shared fallback) via toast on error. The `<strong>` markup is rendered with a small RN-native split helper (mobile has no `<Trans>` usage).
- Thin route `app/(tabs)/profile/delete-account.tsx` (mirrors `dev-servers.tsx`), registered in the profile `_layout.tsx`, plus a red "Delete Account" entry under Sign Out in `more.tsx`.

## Auth provider

`signOut` gains an optional `method` (`'manual' | 'account_deleted'`, default `'manual'`) so deletion records the right analytics event, matching web. Cleanup and the existing `isAuthenticated → false` redirect to `/auth/login` are unchanged. `authSignOut()`'s revoke is best-effort, so revoking a token whose session was just deleted doesn't block cleanup.

## Validation

- `vp run typecheck:mobile` ✅
- `vp run test:mobile` ✅ — 1604 tests (4 new for the hooks)
- `vp run check:mobile-bundle` ✅ (confirms the shared-operations import resolves in Metro)
- `vp check` ✅ — new files 0 errors / 0 warnings / formatted

## Manual QA

Profile → More → **Account → Delete Account**: count/toggle appear only when the user has published climbs, confirm stays disabled until `DELETE` is typed, then deletion signs out and lands on the login screen. Error path (backend down) shows a toast and keeps the user signed in.

### DIFF
--- a/packages/mobile/app/(tabs)/profile/_layout.tsx
+++ b/packages/mobile/app/(tabs)/profile/_layout.tsx
@@ -4,6 +4,7 @@ import { useTranslation } from 'react-i18next';
 
 export default function ProfileLayout() {
   const { t } = useTranslation('common');
+  const { t: tSettings } = useTranslation('settings');
 
   return (
     <Stack
@@ -21,6 +22,7 @@ export default function ProfileLayout() {
       {/* i18n-ignore-next-line — preview-only screen */}
       <Stack.Screen name="branch-switcher" options={{ title: 'Branch Switcher' }} />
       <Stack.Screen name="dev-servers" options={{ title: t('mobile.more.metroServersTitle') }} />
+      <Stack.Screen name="delete-account" options={{ title: tSettings('deleteAccount.title') }} />
     </Stack>
   );
 }
--- a/packages/mobile/app/(tabs)/profile/delete-account.tsx
+++ b/packages/mobile/app/(tabs)/profile/delete-account.tsx
@@ -1,0 +1,5 @@
+import { DeleteAccountScreen } from '../../../src/components/DeleteAccountScreen';
+
+export default function DeleteAccountRoute() {
+  return <DeleteAccountScreen />;
+}
--- a/packages/mobile/app/(tabs)/profile/more.tsx
+++ b/packages/mobile/app/(tabs)/profile/more.tsx
@@ -31,6 +31,7 @@ export default function MoreScreen() {
   const { t } = useTranslation('common');
   const { t: tProfile } = useTranslation('profile');
   const { t: tPlaylists } = useTranslation('playlists');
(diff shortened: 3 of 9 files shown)


marcodejongh:
Addressed the review feedback in `acbf5ff`:

- **Codex P2 (block deletion until climb info loads)** — fixed. `handleDelete` and the confirm button now also gate on `infoQuery.isLoading`, so on a slow connection the destructive button can't fire with `removeSetterName: false` before the setter-name choice has had a chance to appear.
- **Claude review #2 (fragile test timer)** — replaced the `setTimeout(0)` negative assertion with a `fetchStatus === 'idle'` `waitFor`.
- **Claude review #3 (no component tests)** — added `DeleteAccountScreen.test.tsx`: confirm gated on exact `DELETE` *and* on info load, setter-name toggle visibility by published-climb count, plus the success (signOut + `removeSetterName` wiring) and error-toast paths.
- **Claude review #1 (button visual consistency)** — kept Sign Out (bordered) vs Delete Account (text link) as a deliberate hierarchy and documented the intent in `more.tsx`.

Local: `typecheck:mobile`, `test:mobile` (1609), `check:mobile-bundle`, and `vp check` on changed files all green.

PR #2721, opened 2026-06-10
fix(mobile): make iOS social login work and stop logging failures as 'cancel'

marcodejongh:
## Root cause

Sign in with Apple/Google in the RN app has had **near-zero success since TestFlight launch**, with every failure logged as `failure_reason='cancel'`. Three separate bugs stacked:

1. **The 'cancel' events are not cancellations.** expo-web-browser maps *every* ASWebAuthenticationSession failure to `{type:'cancel'}` (`WebAuthSession.swift:29` — `"type": callbackUrl != nil ? "success" : "cancel"`), and `login.tsx` forwarded `result.type` raw into `failure_reason`. PostHog shows the two heavy retriers (24×/15×, builds 100219/100220, both iOS 26.5.1) failing **100–250 ms** after `Login Attempted` — the auth sheet never opened. expo's presentation anchor is the deprecated `UIApplication.shared.keyWindow ?? ASPresentationAnchor()`, which fails to provide a foreground-scene window on iOS 26 (apps are now built with Xcode 26 / UIScene lifecycle).
2. **The successes were invisible.** `callback.tsx` logged `Login Succeeded` without `auth_method`, so the two social logins that *did* complete (iOS 26.5.1 build 100205 via Apple, Android 16 via Google) never showed up in any per-method funnel. This also proves the rest of the pipeline works: prod has both providers configured (`/api/auth/providers-config` → `{"google":true,"apple":true}`), the NextAuth signin POST 302s to Google with correct PKCE/state/callback cookies (verified with curl), and the transfer-token exchange works.
3. **The `invalid_transfer_token` blips on credentials logins were mislabeled typos.** The classifier string-matched the 401 body against `'Invalid credentials'`, but the backend sends `'Invalid email or password'` (`packages/backend/src/handlers/native-auth.ts:21`) — every real wrong-password attempt fell into the `invalid_transfer_token` bucket. Matches "2 people, both recovered on retry".

## Fix (JS/TS only — **ships OTA, no native build needed**)

- **iOS**: drop `openAuthSessionAsync` entirely. Open `/auth/native-start` in SFSafariViewController (`openBrowserAsync`) and race it against the `com.boardsesh.app://auth/callback` deep link (`auth-session-race.ts`, pure + unit-tested). This is the same hand-off the legacy Capacitor app shipped with — the server's deep-link callback page (`packages/web/app/api/auth/native/callback/route.ts`) was explicitly built for it. **Android unchanged** (confirmed working end-to-end).
- **Telemetry**: callback-screen events now carry `auth_method`; OAuth failures carry `duration_ms` (a ~150 ms "cancel" is a programmatic failure, a 30 s one is a human); browser-open failures surface as `exception` + `failure_detail` instead of an unhandled rejection.
- **Classifier**: 401s are classified by endpoint (`credentials` vs `exchange`), not server copy.

## Validation

- `vp run typecheck:mobile` ✅
- `vp run test:mobile` — 237 files, 1692 tests ✅ (new: `auth-session-race.test.ts`, expanded 401 classifier regression cases)
- `vp run check:mobile-bundle` ✅ (10.7 MB)
- `vp check` ✅ on changed files (16 remaining complaints are pre-existing drift on main, untouched here)
- Device QA pending (Linux box): see the QA-notes comment below — test Apple + Google on an iOS 26 device (the broken case), genuine-cancel path, wrong-password path, and confirm PostHog now shows `auth_method` on `Login Succeeded`.

## Follow-ups (out of scope)

- Native Sign in with Apple (`expo-apple-authentication` + entitlement) and native Google Sign-In — the long-term UX answer; needs a fresh build, a Google iOS OAuth client, and a backend JWKS-verify endpoint.
- Upstream issue/patch for expo-web-browser's deprecated `keyWindow` presentation anchor.

### DIFF
--- a/packages/mobile/app/auth/__tests__/callback-i18n.test.tsx
+++ b/packages/mobile/app/auth/__tests__/callback-i18n.test.tsx
@@ -11,7 +11,10 @@ import { beforeEach, describe, expect, it, vi } from 'vitest';
 
 const analytics = vi.hoisted(() => ({ track: vi.fn() }));
 const router = vi.hoisted(() => ({ replace: vi.fn() }));
-const params = vi.hoisted(() => ({ transferToken: undefined as string | undefined }));
+const params = vi.hoisted(() => ({
+  transferToken: undefined as string | undefined,
+  error: undefined as string | undefined,
+}));
 
 vi.mock('../../../src/lib/analytics', () => ({ track: analytics.track }));
 
@@ -23,13 +26,15 @@ vi.mock('react-native', () => ({
 }));
 
 vi.mock('expo-router', () => ({
-  useLocalSearchParams: () => ({ transferToken: params.transferToken }),
+  useLocalSearchParams: () => ({ transferToken: params.transferToken, error: params.error }),
   useRouter: () => router,
 }));
 
 // `t` echoes the key, so any English literal left in the component would show
-// up verbatim and fail the assertions below.
-vi.mock('react-i18next', () => ({ useTranslation: () => ({ t: (key: string) => key }) }));
+// up verbatim and fail the assertions below. Stable identity (like the real
+// hook's) so the effect depending on it doesn't re-fire on every render.
+const stableT = vi.hoisted(() => (key: string) => key);
+vi.mock('react-i18next', () => ({ useTranslation: () => ({ t: stableT }) }));
 
 const auth = vi.hoisted(() => ({
   exchangeTransferToken: vi.fn(async () => ({ success: false, error: 'Invalid or expired transfer token' })),
@@ -39,8 +44,11 @@ vi.mock('../../../src/lib/auth', () => ({
   getPendingOAuthProvider: () => 'apple',
 }));
 vi.mock('../../../src/lib/native-auth-analytics', () => ({ classifyNativeAuthFailureReason: () => 'invalid_token' }));
+// Stable across renders like the real provider's useCallback — a fresh fn per
+// render would re-fire the effect after setError's re-render and double-track.
+const refreshAuthStateMock = vi.hoisted(() => vi.fn(async () => {}));
 vi.mock('../../../src/providers/auth-provider', () => ({
-  useAuth: () => ({ refreshAuthState: vi.fn(async () => {}) }),
+  useAuth: () => ({ refreshAuthState: refreshAuthStateMock }),
 }));
 vi.mock('../../../src/providers/theme-provider', () => ({ useTheme: () => ({ brandColors: { error: '#FF3B30' } }) }));
 
@@ -51,6 +59,7 @@ beforeEach(() => {
   router.replace.mockClear();
   auth.exchangeTransferToken.mockClear();
   params.transferToken = undefined;
+  params.error = undefined;
 });
 
 describe('AuthCallback localization', () => {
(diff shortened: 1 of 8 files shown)


marcodejongh:
# QA: RN social login fix (Apple/Google) — branch fix/rn-social-login-ios

Device QA only (macOS + iPhone) — the change is iOS-flow + analytics; Linux validation
(typecheck, 1692 mobile tests, Metro bundle check) already passed.

## What changed

- iOS no longer uses `openAuthSessionAsync` (ASWebAuthenticationSession failed to
  present on iOS 26 devices → instant fake "cancel"). It now opens
  `/auth/native-start` in SFSafariViewController (`openBrowserAsync`) and waits for
  the `com.boardsesh.app://auth/callback` deep link, like the legacy Capacitor app.
- Android flow unchanged.
- Analytics: `Login Succeeded/Failed` from the callback screen now carry
  `auth_method`; OAuth failures carry `duration_ms` (+ `failure_detail` on
  exceptions); credentials 401s now log `invalid_credentials` (was wrongly
  `invalid_transfer_token`).

## How to test (OTA, no fresh build needed)

1. `vp run mobile:publish` from this branch, point a preview channel at it
   (`bunx eas-cli@16 channel:edit preview-N --branch fix/rn-social-login-ios`),
   or just use the CI-published EAS update for the PR branch.
2. **iOS 26 device** (the previously broken case — builds 100219/100220 failed in
   ~150ms): Sign out → Login screen → "Sign in with Apple".
   - Expect: Safari sheet opens on the Boardsesh spinner page → Apple login →
     "Open in Boardsesh?" prompt → app foregrounds on the auth spinner → lands
     signed-in on Climbs.
3. Repeat with "Sign in with Google" on the same device.
4. **iOS 18 device** if available: same two flows (regression check — this case
   already opened the sheet before).
5. Cancel path: open Apple sign-in, close the Safari sheet without logging in.
   - Expect: back on the login screen, no error toast, no crash.
6. Wrong-password path: email/password login with a bad password.
   - Expect: "invalid credentials" message; retry with correct password works.

## What to check in PostHog (project 412845, $lib = posthog-react-native)

- `Login Succeeded` with `auth_method: apple` / `google` (first ever with
  auth_method set — previously NULL).
- `Login Failed` from a deliberate cancel: `failure_reason: cancel` with
  `duration_ms` in the seconds range (not ~150ms).
- Wrong password: `Login Failed` with `failure_reason: invalid_credentials`.

## Edge cases

- Double-tap the Apple button: second tap is a no-op (`oauthInProgress` guard).
- Complete OAuth, then immediately background/foreground the app: should not
  double-exchange the transfer token (module-level dedupe in callback screen).
- Android sanity check (any device): Google login still works end-to-end.

marcodejongh:
Addressed all four review points in d917c39f4:

1. **pendingOAuthProvider lifecycle** — now consumed instead of lingering: the callback screen clears it after every exchange outcome (success, classified failure, exception, no-token mount), login.tsx clears it in `finally` on every terminal outcome that doesn't hand off to the callback (cancel/dismiss, exception, no-token), and `signOut()` clears it. A stale deep link or hot reload can no longer inherit a previous attempt's provider; the callback effect captures the value at mount so clearing can't race the in-flight attribution.

2. **startSignIn tests** — new `auth-start-sign-in.test.ts` covers the `Platform.OS` fork (Android delegates to `openAuthSessionAsync` with the callback scheme), success/cancel result mapping, error-to-throw promotion, and the pending-provider lifecycle (set per attempt, cleared explicitly and on sign-out).

3. **dismissBrowser safety** — the injected io now wraps it as `async () => WebBrowser.dismissBrowser()`, so a void return or synchronous throw becomes a normal rejection the race's `.catch` swallows. Covered by a test that exercises both the void-return and sync-throw cases against the actual io object handed to the race.

4. **no-token double-tracking** — went further than guarding the error-param case: the callback screen no longer tracks the no-token mount at all. login.tsx owns that outcome (it parses the same callback URL from the browser result and reports the precise reason), so no combination of explicit-route + expo-router auto-route can double-count, and a stale parameterless deep link can't invent a failed attempt.

Validation: `vp run typecheck:mobile`, `vp run test:mobile` (238 files / 1699 tests), `vp run check:mobile-bundle` all green.

PR #2725, opened 2026-06-10
Apple Health + Strava session export (standardized integrations layer)

marcodejongh:
## Summary

Adds the first two external platform integrations to the mobile app — **Apple Health** (device-local HealthKit workouts) and **Strava** (server-side OAuth + activity upload) — behind a small standardized layer (generic credential/export tables, one GraphQL surface, a mobile integration registry) so future providers (Health Connect, Garmin) slot in without new plumbing.

### Apple Health (iOS)
- New local Expo module `packages/mobile/modules/health-workouts` (Swift, mirrors the live-activity module pattern). Ports the old Capacitor plugin and goes further:
  - **Estimated active calories** via the MET formula using the user's latest HealthKit body-mass sample (70 kg fallback) — old plugin wrote 0 kcal, so sessions never counted toward Activity rings.
  - **Per-climb `.lap` workout events** from tick timestamps, so the workout shows session structure in Apple Health.
  - Energy write is best-effort: denying the Active Energy toggle can't sink the workout save.
- Auto-saves on session end (preference, default on, Settings → Connected apps) + manual "Save to Apple Health" on the session summary. Dedupe persists through the existing `setInferredSessionHealthKitWorkoutId` mutation / `healthKitWorkoutId` column.
- `plugins/with-healthkit.js` adds the entitlement + usage strings.

### Strava
- DB: `integration_credentials` (encrypted access/refresh tokens, per `(user, provider)`) and `integration_exports` (per `(provider, user, sessionType, sessionId)` upload dedupe; a `pending` claim row serializes the endSession auto-upload against a simultaneous manual share so a session can never become two Strava activities).
- Backend HTTP: `GET /integrations/:provider/start?token=<jwt>` → Strava authorize (HMAC-signed state carries the user; no cookies) → `GET /integrations/:provider/callback` exchanges the code server-side, stores encrypted tokens, and 302s to `com.boardsesh.app://integrations/strava?status=…` so the in-app browser closes.
- GraphQL: `integrations` query + `disconnectIntegration` / `setIntegrationAutoSync` / `syncSessionToIntegration` (participant-authorized, idempotent).
- `endSession` fire-and-forgets an upload for every participant with Strava connected + auto-sync on (activity: `RockClimbing`, name like "Kilter Board session — 12 sends", description from grade distribution/hardest climb). Refresh-token rotation is persisted; 401 marks the credential expired and surfaces a reconnect prompt.
- Mobile UI: Connected apps screen (Profile → More) with connect / auto-sync toggle / disconnect, plus "Share to Strava" on the session summary.

v1 scope: explicit (recorded/party) sessions only; inferred solo sessions are export-ready in the schema (`sessionType` column) but not wired. Mobile-only UI per scope decision.

## Validation
- `vp run typecheck` ✅, `vp run test:mobile` ✅ (1721 tests), 5 new backend test files ✅ (33 tests: HMAC state, Strava wire protocol, resolver auth/IDOR, auto-sync fan-out, export claim race). The pre-existing `integration.test.ts` WS-1006 environmental failures are untouched.
- `vp run check:mobile-bundle` ✅; new Swift compiled clean via `xcodebuild` against the prebuild workspace (the shared `vp run mobile:ios` cache was locked by another running build).
- Paired QA review agents on both halves; findings fixed: best-effort energy write, upload claim race, Strava `elapsed_time=0` rejection floor, provider threading, Hermes-safe deep-link query parsing, analytics gated on domain-level upload errors.

## Ops / release checklist (not in this PR)
- [ ] Set `STRAVA_CLIENT_ID`, `STRAVA_CLIENT_SECRET`, `BACKEND_PUBLIC_URL` (`https://ws.boardsesh.com`) on the backend deployment, and register that domain as the Strava app's Authorization Callback Domain.
- [ ] **New native preview build required** (new Expo module + HealthKit entitlement; JS-only follow-ups ride OTA). Enable the HealthKit capability on the App ID; App Store privacy questionnaire gains Health & Fitness (write workouts, read body mass, no tracking).
- [ ] Replace the placeholder Strava-orange connect button with the official "Connect with Strava" brand asset (marked with a TODO in `StravaCard.tsx`).


marcodejongh:
Addressed the latest review round in f0fc1062c (after merging latest main in d115e59b7 — migration renumbered to 0123_true_firestar since main took 0122):

- **upsertSuccessExport/upsertErrorExport setWhere guard**: both finalizing upserts now require `status='pending'`, matching claimExport — a late writer whose claim was stolen via stale takeover can no longer clobber a finalized row's `externalActivityId`. The helpers are exported for tests and pinned by three new real-Postgres tests in `integration-export-claim.test.ts`.
- **disconnectIntegration return value**: the resolver now returns `disconnect()`'s boolean (`false` = nothing to remove). Kept as a non-error rather than throwing: disconnect is idempotent, so a double-tap or a retry after a dropped response shouldn't surface a failure toast. The mobile hook ignores the payload, so no client change needed.
- **Missing resolver tests**: added disconnect happy path, disconnect-with-no-credential (locks in the boolean fix), setIntegrationAutoSync not-connected (throws), and the setAutoSync happy path.
- **ShareToStravaButton cold cache**: renders a disabled placeholder button while the statuses query settles instead of returning null, so the action no longer pops in after the fetch for connected users.
- **STRAVA_ORANGE**: extracted to `integrations/strava-brand.ts`, imported by both StravaCard and ShareToStravaButton.
- **provider CHECK constraint — deliberately not added**: a `CHECK (provider IN (...))` makes every future provider a schema migration, which cuts against the registry-only extension design (`docs/integrations.md` add-a-provider checklist), and it only guards writes that bypass the app entirely — every app path validates provider through zod + the GraphQL enum, and `autoSyncSessionToIntegrations`/`getProvider()` already skip unknown provider rows safely. Happy to add it if you'd rather take the migration-per-provider cost.

Validation: full typecheck, 19 resolver + 9 claim real-DB tests, full backend suite (only pre-existing transport flakes in `integration.test.ts`/`graphql-resolvers.test.ts` failed — socket hang-ups unrelated to this PR), full mobile suite (only `bluetooth-provider-mismatch-switch.test.tsx` failed, which is broken on main itself — its queue-provider mock is missing `useIsPartyPreviewOnly`, files byte-identical to main), Metro bundle check OK.

PR #2727, opened 2026-06-10
feat(auth): native Sign in with Apple + Google on mobile

marcodejongh:
## What

Replaces the web-browser OAuth handoff in the Expo app with **true native sign-in**: Sign in with Apple (`expo-apple-authentication`) and Google Sign-In (`@react-native-google-signin/google-signin`). Each provider's identity token is verified server-side against its JWKS and exchanged for our existing mobile JWT pair.

Native Sign in with Apple is also an App Review requirement whenever the app offers another third-party login (Google).

Only touches `packages/backend` and `packages/mobile`. The legacy Capacitor app is out of scope.

## Backend (`packages/backend`)

- **`POST /auth/native/oauth`** (`handlers/native-auth.ts`): verifies the Apple/Google identity token via `jose` (`createRemoteJWKSet` singletons + `jwtVerify` with issuer/audience/`clockTolerance: 60`), then in one transaction:
  1. links by the provider's stable `sub` (`accounts` row), else
  2. links by normalized email (mirrors web's `allowDangerousEmailAccountLinking`; promotes `emailVerified` when the provider asserts it), else
  3. creates the user + `userProfiles` row + `accounts` link.
  Returns `{ jwt, refreshToken, expiresAt }` — identical to the other native-auth endpoints. Apple's missing-email-on-resubmit resolves via step 1. Best-effort Apple nonce check (`SHA-256(nonce) === token.nonce`). Reuses the shared rate limiter, CORS, and `generateTokenPair`.
- Route + startup log in `server.ts`.
- Tests in `native-auth-oauth.test.ts` (15): new-user, link-by-sub, link-by-email, already-verified, Apple resubmit w/o email, no-email reject, invalid token → 401, nonce mismatch → 401, bad provider → 400, validation, rate limit 429/503.

## Mobile (`packages/mobile`)

- New deps + `app.config.ts`: `ios.usesAppleSignIn`, `expo-apple-authentication` plugin, and the Google plugin (gated on a resolvable reversed-iOS-client `iosUrlScheme`). Env documented in `.env.example`.
- `lib/auth.ts`: `signInWithApple` / `signInWithGoogle` / `oauthNativeSignIn`; cancellation mapped to a no-error `cancelled` result. Login screen uses the official `AppleAuthenticationButton` + `GoogleSigninButton`.
- `auth-provider` exposes `signInWithApple` / `signInWithGoogle` (calls `checkAuth()` on success).
- **Removed** the transfer-token browser flow: `startSignIn`, `exchangeTransferToken`, `auth-callback-url`, `auth-session-race`, `/auth/callback` (+ their tests). The server-side transfer infra (`/auth/native/exchange`, web `/auth/native-start`, `native-oauth` provider) **stays** — already-installed builds depend on it; sunset in a later PR.
- Dropped now-orphaned `callback.*` / `nativeStart.signInApple` / `nativeStart.signInGoogle` i18n keys (en-US/es/fr).

## Validation

- `vp run typecheck:mobile`, `vp run test:mobile` (1714 ✓), `vp run check:mobile-bundle` (OK)
- Backend: `vp test run --project backend native-auth` (67 ✓)
- `vp check` (lint + format) on all changed files: clean

> Note: `vp run check:i18n:orphans` reports orphans in `session.json` / `GeneratorPickerCard.tsx` (dynamic `t(labelKey)`). These are **pre-existing on `main`** and unrelated to this change — no auth keys are flagged.

## Manual / out-of-band steps (cannot be automated here)

- [ ] **Apple Developer**: enable the *Sign in with Apple* capability on the `com.boardsesh.app` App ID.
- [ ] **Google Cloud Console**: create OAuth client IDs — iOS (bundle `com.boardsesh.app`), Android (package + signing SHA-1), and a Web client ID (used as `webClientId`).
- [ ] **EAS**: wire the env vars (`EXPO_PUBLIC_GOOGLE_IOS_CLIENT_ID` / `_ANDROID_CLIENT_ID` / `_WEB_CLIENT_ID`, optional `_IOS_URL_SCHEME`; backend `APPLE_BUNDLE_ID`, `GOOGLE_*_CLIENT_ID`) and cut a **new native build** — new native modules can't ship via OTA.
- [ ] **Manual QA on device**: Apple → system sheet → authenticated; Google → native picker → authenticated; a native sign-in whose email matches an existing web user links (no duplicate `users` row, new `accounts` row); sign out + back in resolves the same user (Apple via stored sub).


marcodejongh:
Addressed CI + review feedback:

- **CI `i18n` failure** — was a pre-existing red on `main` (since the session-generator redesign, `71f3d6725`): `GeneratorPickerCard` passed a dynamic `t(labelKey)` that hard-failed the orphan scanner and left its 5 stepper-label keys + the dead `inStatsAttempts` key orphaned. Fixed in `9e69f8aed` — `stepperRow` now takes the already-translated label so call sites use literal `t('mobile.session.preGenerator…')` keys, and the dead key is removed. `vp run check:i18n:orphans` is green.
- **Codex P2 (hide Google sign-in when native config is omitted)** — fixed in `9d403007d`. Added `isGoogleSignInConfigured()` (mirrors the `app.config.ts` plugin gating: `webClientId` always, iOS URL scheme on iOS) and gate the Google button on it; the whole social section is hidden when neither Apple nor Google is available so the divider doesn't dangle.

`vp run typecheck:mobile`, `vp run test:mobile` (1719 ✓), and the i18n checks all pass locally.

marcodejongh:
Ran a code review and applied the hardening in `b9bc92dcb`:

- **[P1] Verified-email gate** — only auto-link a provider identity to an existing user when the token asserts `email_verified`. Auto-linking an *unverified* email is the takeover vector `allowDangerousEmailAccountLinking` leaves to the provider; now enforced. Unverified emails create their own user (anchored on the provider sub). Added a regression test.
- **[P2] Honest Apple nonce flow** — confirmed against expo-apple-authentication's Swift source that Apple echoes the request nonce **verbatim** (no SHA-256). The client now hashes the nonce before handing it to Apple and sends the raw value, so the token never carries the raw nonce; the backend re-hashes to compare. Comments corrected.
- **[P2] Google button gate** — require `EXPO_PUBLIC_GOOGLE_IOS_CLIENT_ID` on iOS (`GoogleSignin.configure` needs it; a scheme-only build would throw on tap).
- **[P3]** Pin `RS256` in both `jwtVerify` calls; drop the unused `verifyProviderIdentityToken` export; document the backend `APPLE_BUNDLE_ID` / `GOOGLE_*_CLIENT_ID` env vars.

Deliberately left (documented): no unique constraint on `users.email` (pre-existing schema debt — web has the same via NextAuth; a concurrent-first-signin race can 500 and retries cleanly); the nonce is client-optional so it binds the attempt but isn't a server-issued replay challenge; Apple `sub` cross-surface linking assumes web + native share the Apple Developer Team (they do — `9L3HKPZBH3`), else it falls back to email linking.

Backend `native-auth` tests 68 ✓, `typecheck:mobile` ✓.

PR #2769, opened 2026-06-11
[codex] Reset mobile create-climb identity on angle change

### LINKED ISSUE #2654: Revive: clear saved climb on angle change in mobile create-climb (was #2520)
Splitting #2520 out for a fresh implementation against the current `create-climb` code.

## Background
PR #2520 (draft, branch `claude/moonboard-create-screen-fixes-co3TY`) was branched off `990fcc20` (the original `feat: create-climb editor` from #2510). It addressed four review findings, but main has since **rebuilt the entire create-climb surface** in `b05a3cc6` ("rebuild create-climb as a play-drawer-style sheet") plus follow-ups `fb1e7867` ("six create-climb draft/autosave bugs") and the scheme-aware color sweep. That rebuild:
- **Retired `CreateClimbSettingsSheet.tsx`** (deleted 193 lines) — PR #2520 concern #3 (early-return hardening) targets a file that no longer exists.
- Rewrote `use-create-climb-screen.ts` (107 lines diverged between the PR base and current main): new autosave state machine, edit-mode seeding, and Set-Active flow. PR #2520 concern #2 added a 355-line-controller test (`use-create-climb-screen.test.tsx`) written against the old hook shape; it cannot apply.
- Concern #4 (JSDoc) was a no-op even in the PR.

So PR #2520 is not mechanically rebaseable — three of its four items are obsolete.

## What is still worth doing (the one live bug)
Concern #1 — the `savedClimb` / working-angle mismatch — **is still present on main**:
- `packages/mobile/app/(tabs)/climbs/create.tsx` keys the screen by `editClimbUuid ?? 'new'`, **not** by angle, so the screen stays mounted across working-angle changes and a new `angle` flows into the still-mounted hook.
- `packages/mobile/src/components/create-climb/use-create-climb-screen.ts`: `handleSetActive` builds the queue item with `savedClimb?.uuid ?? previewUuidRef.current` while `buildProvisionalClimb` reads the live `board.angle`. There is no effect that clears `savedClimb` / mints a fresh preview uuid when `board.angle` changes.
- Net effect: save at 25°, switch to 40°, tap **Set Active** → queue item pairs the old (25°) climb uuid with the new 40° angle, so the card shows the wrong angle.

### Suggested fix (fresh, against current main)
Add an angle-change effect in `use-create-climb-screen.ts` that, when `board.angle` changes (skipped in edit mode, whose identity is `editClimbUuid`):
- clears `savedClimb`,
- clears any duplicate banner,
- mints a new `previewUuidRef` uuid.
Mirror the existing model where the autosave draft key is already angle-scoped. Add a regression test (the original PR's angle test failed without the fix, passed with it) **and** the edit-mode guard test the Claude review asked for.

The branch `claude/moonboard-create-screen-fixes-co3TY` is preserved for reference, but a fresh PR against the rebuilt code is the right path rather than reviving the branch.

Original PR: https://github.com/boardsesh/boardsesh/pull/2520

marcodejongh:
## Summary

- Reset the mobile create-climb saved/provisional identity when only the board angle changes, while keeping the editor mounted so painted holds and undo history survive session sync.
- Keep edit-mode saves disabled until the existing climb snapshot has loaded, preventing a premature tap from creating a new climb.
- Add hook-level regression coverage for Set Active, second Save-after-angle-change, paint-state preservation, and the edit-mode guard.
- Add a targeted Expo Podfile config plugin for the AppCheckCore static Swift dependency path so `GoogleUtilities` and `RecaptchaInterop` are emitted with modular headers in iOS CI.
- Remove two orphaned mobile i18n keys across locales so the staged hook passes.

Fixes #2654.

## Validation

- `vp test run --reporter=agent packages/mobile/src/components/create-climb/__tests__` — 9 files, 38 tests passed
- `vp run typecheck:mobile`
- `vp run test:mobile` — 251 files, 1817 tests passed
- `vp run check:mobile-bundle` — iOS and Android bundles OK
- `vp run check:mobile-simulator` (skipped: no `xcrun simctl` available)
- `vp run mobile:screenshot` (skipped: no iOS simulator available)
- `vp run check:i18n:orphans`
- `vp staged`

## QA / Review

- Mobile Metro is running for manual QA at `http://marcos-macbook-pro-2-1.tailedd9d5.ts.net:8089`.
- QA notes are loaded from `.boardsesh/qa-notes.md` in this worktree.
- Code-review and QA subagents ran; their concrete findings were addressed with the Save-after-angle-change assertions, paint-state assertions, and edit-mode early-save guard.
- Latest GitHub Actions runs for `74a7cd11af946f4dbe51b471bddd8cca08d7f8e2` were queued after the CI Podfile fix push.

### DIFF
--- a/packages/mobile/app.config.ts
+++ b/packages/mobile/app.config.ts
@@ -267,6 +267,9 @@ export default ({ config }: ConfigContext): ExpoConfig & { newArchEnabled?: bool
       'expo-apple-authentication',
       // Native Google Sign-In — only when an iosUrlScheme is resolvable (see above).
       ...googleSignInPlugin,
+      // GoogleSignIn pulls AppCheckCore, a Swift pod whose transitive ObjC deps
+      // need module maps under static-library CocoaPods integration.
+      './plugins/with-google-swift-static-pod-headers',
       'expo-localization',
       [
         'expo-location',
--- a/packages/mobile/plugins/__tests__/with-google-swift-static-pod-headers.test.ts
+++ b/packages/mobile/plugins/__tests__/with-google-swift-static-pod-headers.test.ts
@@ -1,0 +1,73 @@
+import { createRequire } from 'node:module';
+
+import { describe, expect, it } from 'vitest';
+
+const require = createRequire(import.meta.url);
+
+type GoogleSwiftStaticPodHeadersPlugin = {
+  applyGoogleSwiftStaticPodHeaders(contents: string): string;
+  MODULAR_HEADER_PODS: string[];
+};
+
+const plugin = require('../with-google-swift-static-pod-headers.js') as GoogleSwiftStaticPodHeadersPlugin;
+
+describe('with-google-swift-static-pod-headers', () => {
+  it('adds modular header declarations after use_expo_modules', () => {
+    const podfile = `
+target 'Boardsesh' do
+  use_expo_modules!
+  use_frameworks! :linkage => :static
+end
+`;
+
+    const updated = plugin.applyGoogleSwiftStaticPodHeaders(podfile);
+
+    expect(updated).toContain(
+      [
+        '  use_expo_modules!',
+        '  # AppCheckCore is Swift and these transitive ObjC pods need module maps when linked statically.',
+        "  pod 'GoogleUtilities', :modular_headers => true",
+        "  pod 'RecaptchaInterop', :modular_headers => true",
+        '  use_frameworks! :linkage => :static',
+      ].join('\n'),
+    );
+  });
+
+  it('is idempotent when both modular header declarations already exist', () => {
+    const podfile = `
+target 'Boardsesh' do
+  use_expo_modules!
+  pod 'GoogleUtilities', :modular_headers => true
+  pod 'RecaptchaInterop', :modular_headers => true
+end
+`;
+
+    expect(plugin.applyGoogleSwiftStaticPodHeaders(podfile)).toBe(podfile);
+  });
+
+  it('adds only the missing declaration when one pod is already configured', () => {
+    const podfile = `
+target 'Boardsesh' do
+  use_expo_modules!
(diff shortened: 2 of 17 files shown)


marcodejongh:
Closing this — both substantive halves are on `main`, by other routes.

**Angle reset.** Merged PR #3738 ("reset saved climb + preview uuid when the create working angle changes", merged 2026-07-17) closed issue #2654 and put the same effect in `use-create-climb-screen.ts:189-200`: a `lastAngleRef`, an early return when the angle is unchanged, a skip in edit mode, then `setSavedClimb(null)` / `setPublishDuplicateError(null)` / a fresh `previewUuidRef`. Same reset, same reasoning (the screen key excludes `angle`, so the hook stays mounted and the paint survives).

**Podfile plugin.** `packages/mobile/plugins/with-podfile-app-check-fix.js` on `main` does exactly what `with-google-swift-static-pod-headers.js` does here: same two pods (`GoogleUtilities`, `RecaptchaInterop`), same `:modular_headers => true`, inserted after the same `use_expo_modules!` anchor, for the same reason (AppCheckCore is Swift and won't link those transitive ObjC pods statically without module maps).

**Two small bits didn't come across**, both cosmetic and both cheap to redo on top of `main` if they're wanted:

- The `'loading'` save-button state (`isEditing && !savedClimb`) plus the matching guard in `handleSave`, so pressing Save before the edit fetch resolves is a no-op instead of a silent nothing. `main`'s `SaveButtonState` is still `'ready' | 'saving' | 'justSaved' | 'editLocked' | 'login'`.
- Clearing `justSaved` and its timer on an angle change, so a "Saved" confirmation doesn't linger for up to `JUST_SAVED_MS` after the angle moves.

Neither is worth carrying a 2356-commit-behind, `CONFLICTING` branch for.

PR #2781, opened 2026-06-11
[codex] Guard mobile bottom chrome render stability

### LINKED ISSUE #2565: [mobile] Stop the useBottomChromeMetrics re-render fan-out during live sessions
From the multi-reviewer pass on the queue FAB + search redesign (#2560).

`useBottomChromeMetrics` reads `useSegments()` + the whole `useQueue()` context and is consumed by ~12 always/often-mounted components (snackbar, FABs, every tab list, session views). The queue context value changes on `liveStats` / `sessionUsers` / queue mutations, so during a **live party session** every presence/stat tick re-runs the hook body in all consumers — though it only needs `currentClimbQueueItem?.climb != null`.

### Do
- Expose a `useQueueSelector(s => s.currentClimbQueueItem?.climb != null)` (referential-equality selector) and/or split the high-frequency fields out of the queue context; **or** hoist `useBottomChromeMetrics` into one provider near the tab root and pass the computed object down (so segments + insets are read once, not ~12×).

Note: #2560 already extracted the pure `computeBottomChromeMetrics`, so the math is reusable. Best confirmed with on-device profiling during a live session.

### Where
`packages/mobile/src/hooks/use-bottom-chrome-metrics.ts`, `providers/queue-provider`.

marcodejongh:
Fixes #2565.

## Summary

- Adds a `useBottomChromeMetrics` render-count regression probe to the mobile `QueueProvider` session-update tests.
- Pins the issue contract: bottom chrome metrics consumers do not re-render when switching between active climbs.
- Pins the live-session path called out in the issue: `SessionStatsUpdated` pushes update session state without waking bottom-chrome metric consumers.

## Why

The bottom chrome hook is consumed by many mounted mobile screens. The implementation routes it through the active-climb presence selector; this PR makes that performance behavior executable so a future broad `useQueue()` dependency cannot silently reintroduce the live-session render fan-out.

## Validation

- `vp test run --project mobile packages/mobile/src/providers/__tests__/queue-provider-session-updates.test.tsx --reporter=agent`
- `vp check packages/mobile/src/providers/__tests__/queue-provider-session-updates.test.tsx`
- `vp run typecheck:mobile`
- `vp run check:i18n:orphans`
- `vp run test:mobile` (249 files, 1812 tests)
- `vp run check:mobile-bundle` (iOS 10.6 MB, Android 10.9 MB)
- `vp staged`

## QA

- Mobile Metro is running at `http://vibetunnel-2.tailedd9d5.ts.net:8083`.
- QA notes are loaded from `.boardsesh/qa-notes.md`.

### DIFF
--- a/packages/mobile/src/providers/__tests__/queue-provider-session-updates.test.tsx
+++ b/packages/mobile/src/providers/__tests__/queue-provider-session-updates.test.tsx
@@ -119,8 +119,16 @@ vi.mock('react-native', () => ({
   AppState: { addEventListener: vi.fn(() => ({ remove: vi.fn() })) },
 }));
 
+vi.mock('expo-router', () => ({
+  useSegments: () => ['(tabs)', 'climbs'],
+}));
+
 vi.mock('expo-crypto', () => ({
   randomUUID: () => 'test-correlation-id',
+}));
+
+vi.mock('react-native-safe-area-context', () => ({
+  useSafeAreaInsets: () => ({ top: 0, right: 0, bottom: 0, left: 0 }),
 }));
 
 vi.mock('@boardsesh/graphql-client', () => ({
@@ -165,8 +173,16 @@ vi.mock('../toast-provider', () => ({
   useToast: () => ({ showToast: toast.showToast }),
 }));
 
+vi.mock('../theme-provider', () => ({
+  useTheme: () => ({ variant: 'liquidGlass' }),
+}));
+
 vi.mock('../queue-snackbar-provider', () => ({
   useQueueSnackbar: () => ({ showQueueAddedSnackbar: vi.fn() }),
+}));
+
+vi.mock('../../hooks/use-bottom-accessory', () => ({
+  useNativeAccessoryActive: () => false,
 }));
 
 import {
@@ -179,6 +195,7 @@ import {
   useQueueSessionId,
 } from '../queue-provider';
 import { clearStoredSessionId } from '../../lib/session-store';
+import { useBottomChromeMetrics } from '../../hooks/use-bottom-chrome-metrics';
 
 type Snapshot = {
   state: ReturnType<typeof useQueue>['state'];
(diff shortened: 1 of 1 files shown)


marcodejongh:
Closing this — the contract it pins is now enforced by the architecture, and `main` already has an equivalent test one layer down.

**#2565 was closed by merged PR #3740** ("hoist bottom-chrome metrics into one provider", merged 2026-07-17). The ~9 input-hook subscriptions moved out of every consumer into a single `BottomChromeMetricsProvider`; `useBottomChromeMetrics` is now a plain `useContext` read of a memoized value. So the fan-out this test guards against can't come back by way of a consumer picking up a broad `useQueue()` — there is only one subscriber left.

**The queue read is presence-only and structurally isolated.** The provider's one queue touch is `useStickyAccessoryPresence()` → `useHasAccessoryClimb()` → `useHasActiveClimb()`, which reads `QueueHasActiveClimbContext` — a dedicated `{ hasActiveClimb: boolean }` context memoized on that boolean alone (`queue-provider.tsx:1229`). Climb-to-climb navigation doesn't change it. The ≤1/2s `SessionStatsUpdated` push lands in a separate `QueueLiveStatsContext` the provider never reads, so the live-stats half of the contract holds the same way.

**`main` already tests both halves at the selector level**, in the very file this PR edits: `queue-provider-session-updates.test.tsx:905`, *"keeps the active-climb presence selector stable when switching between climbs"* — same setup (two `setCurrentClimb` calls), same assertion shape (snapshot count unchanged), just asserted on `useHasActiveClimb` instead of `useBottomChromeMetrics`.

That file on `main` has zero `bottomChrome` references, and the mocks this PR adds (`expo-router`, `react-native-safe-area-context`, `../theme-provider`, `use-bottom-accessory`) were all needed only because the hook used to pull those subscriptions in itself. Re-landing this would mean rewriting the probe against the provider, to assert something the selector test already covers.

PR #2785, opened 2026-06-11
feat(mobile): offline mode — download a board, browse and log with no signal

### LINKED ISSUE #2639: Revive: Offline sync, end-to-end: Phase 2 backend + on-device SQLite + mobile wiring (was #2277)
Reviving the offline-sync stack from #2277. The feature is still wanted and is **not** in `main` (verified: no `sync-deletions` schema, no `packages/backend/.../resolvers/sync`, no `packages/mobile/src/mutation-queue`, no `pull-client`, no sync GraphQL types on `origin/main`). It was not superseded by any newer open PR.

## What the PR intended
End-to-end mobile offline sync:
- **Phase 2 backend** — idempotent `addFavorite`/`removeFavorite`, `saveTick`/`createPlaylist` accepting a client `uuid` with `ON CONFLICT DO NOTHING`, 10 composite-cursor sync resolvers, `sync_deletions` + per-table triggers.
- **Lock-safe DB migrations** — `0108` (nullable-add → batched `FOR UPDATE SKIP LOCKED` backfill → set default/not-null on the 200k–1M-row board tables, no full-table rewrite), `0109` sync triggers, `0110` cursor indexes on the 7 user sync tables.
- **On-device SQLite** — `packages/mobile/src/db` connection/DDL/migrations, an `SQLiteProvider` that degrades to online-only on init failure, `OfflineSyncBridge` drain-then-pull scheduler.
- **Live dual-write** — `useSaveTick`/`useToggleFavorite` persist to local SQLite AND enqueue a 1-by-1 sync drainer; FIFO queue with dead-letter recovery in Settings; optional pre-warmed seed DB.

## Why a mechanical rebase was not possible
The branch (`mobile-phase5`, last commit 2026-05-31) is **618 commits behind `main` and 11 ahead** (merge-base `e4bb60578`). During those 618 commits `main` restructured exactly the files this PR integrates against, so the rebase hit structural — not mechanical — conflicts on the very first commit:
- `packages/mobile/app/(tabs)/more/index.tsx` was **deleted** on `main` (settings restructured into providers/separate surfaces) but the PR modifies it → modify/delete conflict; the PR’s settings UI (offline data controls, dead-letter recovery, seed-DB toggle) has no mechanical landing spot.
- `packages/mobile/app.config.ts` plugins array diverged structurally — `main` added `expo-share-intent`, Android release-signing / BLE-feature / session-service / gradle-memory plugins; the PR adds an `expo-notifications` block to the same array.
- `theme-provider.tsx` (full "Velvet Send" violet recolor + scheme-aware foregrounds), `PlayDrawer.tsx` (full-screen now-playing rewrite + deferred board mount), `LogbookSection.tsx`, `_layout.tsx`, and `auth-provider.tsx` were all heavily reworked on `main`.
- i18n catalogs were relocated to `packages/shared/i18n/locales/`; the PR only edits the old `packages/web/i18n/locales/` tree.
- DB migrations `0108`–`0110` collide with `main`’s `0111`–`0118` and need renumbering, alongside regenerating the shared-schema generated types/`schema.graphql` against the moved schema.

Reaching green requires re-implementing the settings UI on `main`’s new settings architecture, reconciling the schema files, and regenerating generated types — i.e. feature rework, beyond a mechanical conflict resolution.

## Revival
Re-apply the backend + db + on-device-SQLite layers (which barely overlap `main`) first, then re-integrate the mobile UI against the current More-tab/settings/PlayDrawer/theme structure. Branch `mobile-phase5` is **preserved** for cherry-picking. The 15 open review items (Codex + Claude review on #2277 — pull-client column allowlisting, notification deep-link params, token-manager re-login leak, unawaited drains, 401 retryability, sync-scheduler `pendingTrigger`, missing tests) should be addressed in the revival.

Original PR: https://github.com/boardsesh/boardsesh/pull/2277

marcodejongh:
## Summary

Fixes #2639.

This PR revives the offline sync foundation for React Native across backend, database, shared GraphQL schema, and mobile runtime wiring.

- Adds backend sync pull queries, tombstone tracking, idempotent tick/favorite/playlist mutations, validation schemas, and sync contract tests.
- Adds DB sync cursor columns/indexes, deletion tombstones, and migration triggers for updated rows, including sequence-backed board reference cursors.
- Adds mobile SQLite schema/migrations, mutation queue/drainer, pull scheduler/status, settings, notification token handling, and the root offline sync bridge.
- Wires production tick saves through the shared `@boardsesh/board-react` adapter into SQLite + queued replay, and favorite toggles into deterministic add/remove queue operations where current favorite state is known.
- Surfaces pending local ticks in QuickTick/Logbook UI and adds localized settings/dead-letter/offline sync copy.

**The entire offline engine is behind the `offline-board-downloads` flag** (PostHog, off by default, per-user targeting). Flag-off is byte-for-byte pre-offline behavior: no sync scheduler, no local-first reads or offline fallbacks, no dual-write ticks/favorites, no pending-sync badges, no sign-out dialog. The flag decision is made in React (`useOfflineDownloadsEnabled`, sees tester overrides) and mirrored to a module-level store for the non-React read interceptor. Two deliberate exceptions: the (empty, idle) SQLite database is still created at launch, and **draining is never gated** — writes queued while the flag was on still flush after a rollback (one-shot drain in `OfflineSyncBridge`, pending-count-gated sign-out drain, dead-letter Retry).

## Release Notes

- Early testers: download a board in My Boards and keep browsing, logging sends, and saving favorites with no signal — everything queues on your phone and syncs when you're back online. (Rolling out gradually behind a feature flag.)

## Deploy notes

- **Migration `0144` locks `board_climbs` + `board_climb_stats` for its full runtime** (ACCESS EXCLUSIVE: column adds + batched full-table backfill + `SET NOT NULL` + non-concurrent index builds in one transaction). Climb browse/search and the Aurora ETL block while it runs. Check prod row counts first (`SELECT relname, n_live_tup FROM pg_stat_user_tables WHERE relname IN ('board_climbs','board_climb_stats','boardsesh_ticks')`) and deploy at a low-traffic hour.
- New DELETE triggers write a `sync_deletions` tombstone on every tick/favorite/playlist/follow delete for **all** clients (web included) from day one; the daily prune bounds growth to 90 days.
- New native deps (`expo-sqlite`, `react-native-mmkv`, `expo-notifications`) change the fingerprint: this JS will not OTA to existing binaries and ships with the next store build.
- Rollout: raise the flag in PostHog (`offline-board-downloads`, id 745547) — add a release condition for the pilot users, keep everyone else at 0%.

## Validation

- `vp run typecheck:db`
- `vp run typecheck:backend`
- `vp run typecheck:shared`
- `vp run typecheck:mobile`
- `vp test run packages/backend/src/__tests__/sync-mutations.test.ts packages/backend/src/__tests__/sync-pull.test.ts packages/mobile/src/hooks/__tests__/use-offline-mutations.test.ts packages/mobile/src/lib/graphql/__tests__/hooks-dual-write.test.ts packages/shared/board-react/src/__tests__/use-save-tick.test.tsx --reporter=agent`
- `vp test run packages/mobile/src/db/__tests__ packages/mobile/src/hooks/__tests__/use-local-ticks.test.ts packages/mobile/src/mutation-queue/__tests__ packages/mobile/src/notifications/__tests__ packages/mobile/src/settings/__tests__ packages/mobile/src/sync/__tests__ --reporter=agent`
- `vp test run --project mobile --reporter=agent`
- `vp run check:i18n`
- `git diff --check`

Flag-gating pass (2026-07-06): `vp run typecheck:mobile`, `vp run test:mobile` (455 files / 3673 tests), `vp run check:mobile-bundle`, `vp check` (error/warning count identical to the pre-change baseline; the 5 remaining errors are pre-existing repo-root `scripts/` issues from main).

Review-sweep pass (2026-07-06, after the rebase onto `4ab0cbb51`): four parallel review agents (sync-engine correctness, backend/DB + migration, mobile data path + gating, RN perf/React) with adversarial verification; all confirmed findings fixed in `8eb8c43e8`. Re-validated: `vp run test:mobile` (457 files / 3694 tests), backend suite (2080/2081 — the one failure is the known pre-existing board-presence slug-collision sandbox flake), `vp run typecheck`, `vp run check:mobile-bundle`, `vp check` at baseline.

## Review findings fixed (2026-07-06)

- **Sign-out wipe race (cross-account leak)**: a pull page or drain tail in flight across the millisecond-scale wipe window resurrected the old user's rows + checkpoints, or posted their queued writes under the next user's token. Fixed with a monotonic wipe epoch checked after every network await.
- **Tombstone resurrection**: delete-then-re-add on another device could permanently erase the re-added row (tombstone applied after the upsert, strict `>` cursor never refetches). Deletions now run first with an `updated_at <= deletedAt` guard.
- **Board re-import flood**: `clearBoardData`'s delete+recreate would have written one tombstone per climb (200k+) and wiped every client's offline board copy. The import now suppresses the triggers via a transaction-local GUC (migration 0144 edited — it has never shipped).
- **Cursor skip race**: rows stamped at transaction start but committed late (bulk JSON imports) could land behind an advanced cursor, permanently unsynced. Sync pages now exclude rows younger than a 30s stability window.
- **Timestamp format drift**: sync documents carried pg-text timestamps while local writes used ISO — `normalizeRow` now normalizes both (the drizzle postgres-js driver returns strings, never `Date`s).
- **`useSetting` render loop (crash)**: array-typed settings re-parsed JSON per `getSnapshot`, looping React into "Maximum update depth exceeded" on any Manage Boards mount once `syncEnabledBoards` had been written (every sign-out writes it). Reference-stable snapshots + regression test proven against the old code.
- **Partial-board local reads**: a board mid-download (minutes for 40k climbs) served truncated local search results while online. Local-first reads now gate on a per-scope "initial download complete" marker.
- Plus: offline tick badge invalidation, favoriteStatus optimistic pin (was racing the queued mutation), `createPlaylist` atomicity, cursor validation hardening (Z-only, bigint cap), follow/unfollow queue netting, no-op sync cycles no longer invalidate all caches, `['infiniteSearchClimbs']` invalidation, sign-out pending-writes warning independent of the flag, auth-gated sync bridge, local/server search parity (name collation, gradeAccuracy parsing).

## Rebase note (2026-07-05)

Rebased onto latest `main`. The DB migration was renumbered `0141_offline_sync_backend` → **`0144_offline_sync_backend`** because `main` landed `0141_rename_ascensionist_upstream` (#3461) and `0142_windy_roulette` (#3473) in the meantime. The migration SQL body is unchanged; its drizzle snapshot (`meta/0143_snapshot.json`) was regenerated against the merged schema so it reflects both new columns (`upstream_ascensionist_count`, `user_boards.timer_name`) — drizzle reports no drift. The migration was verified to apply cleanly in order against a throwaway Postgres (triggers/sequences/tombstones all fire as designed).

## Manual QA

Detailed QA notes are in `.boardsesh/qa-notes.md` in the local worktree.


marcodejongh:
Rebased on main and renumbered the migration — current head `2045a5016`:

- **Migration is now `0143_offline_sync_backend`** (main took 0141 for the ascensionist rename and 0142 for `user_boards.timer_name`). The snapshot/journal were regenerated via `drizzle-kit generate` so the meta chain includes main's schema state (`0143.prevId` == `0142.id`), and the migration body keeps the hand-tuned lock-safe SQL (nullable adds → batched backfills → NOT NULL, explicit sequences, triggers) — same end state as the generated DDL, verified table-by-table against the fresh schema export.
- **Commit-lint fixed**: the root commit is reworded to `feat(mobile): implement offline sync for React Native` (message-only; tree untouched), and the PR title is now conventional too.
- Note for concurrent sessions: two rebases raced on this branch — this head adopts the later one (`cc0a1319f`, which had an equivalent 0143) plus the commit-lint reword. Trees were verified identical before the force-push.
- Local validation on this head: full `vp run typecheck` green across packages, mobile suite 3629/3629, Metro bundle check OK. The remaining `vp check` errors (watch-pairing-section, posthog-client) are pre-existing on main — files identical to main.
- Also verified the older high-severity review findings are stale at this tip: the local SQLite schema has no foreign keys so the `INSERT OR REPLACE` cascade can't occur; all interpolated SQL identifiers are config-allowlisted with values bound; 401s are bounded by token-refresh + per-cycle caps + dead-lettering (drainer tests pin both).

marcodejongh:
Triage of the latest review round (working on these now — head `2045a5016`, please don't parallel-fix):

1. **sync_deletions unbounded growth — confirmed, fixing.** Adding a daily prune job in the backend (same `setInterval` + `intervals` idiom as the queue flush), deleting tombstones older than the 90-day retention the schema documents. Clients that haven't synced in 90 days already resync from scratch via checkpoints, so pruned tombstones are safe.
2. **Unknown-column throw breaks forward compat — confirmed, fixing.** `documents` is an opaque JSON array, so a backend column added during an OTA gap would brick every older client's sync loop. Switching to skip-and-warn: unknown keys are dropped; SQL safety is unchanged because the statement's column list was always derived from the allowlist intersection, never from document keys.
3. **Push-permission prompt on sign-in — refuted.** `startTokenManagement` has zero call sites in app code (only the module barrel and tests); `auth-provider` calls only `stopTokenManagement` on sign-out, which removes listeners and never requests permission. No prompt can fire. The accurate residual is that `src/notifications/` is currently uncalled scaffolding for the future sync-push path.

marcodejongh:
Both confirmed review items fixed in `dfe568a89`:

1. **Forward-compat sync** (`pull-client.ts`): unknown document columns are now skipped instead of throwing, so a backend column added during an OTA gap can't brick older clients' sync loops. SQL safety is unchanged (the upsert's column list was always allowlist-derived). Drift stays observable: each unknown table+column is reported to telemetry once per launch, and the integration suite now pins both directions — an additive server column lands the row and emits exactly one report, while a resolver misnaming a *required* column still fails loudly via the local DDL's NOT NULL.
2. **sync_deletions retention** (`server.ts` + new `services/sync-deletions-prune.ts`): a daily job (plus one run 5 min after boot, idempotent across instances) deletes tombstones older than the 90-day retention the schema documents. Unit-tested including the retention arithmetic on the bound cutoff.

One correction to my earlier comment: there is **no** automatic from-scratch resync for a client whose deletions checkpoint predates the retention window — such a client (offline >90 days) would keep stale local rows until the affected records are next upserted. That's documented on `SYNC_DELETIONS_RETENTION_DAYS` as a client-side follow-up (detect the gap → force full resync); with nothing shipped yet, unbounded table growth was the worse trade.

Validation: backend + mobile typecheck green, mobile suite 3630/3630, prune service unit tests green.

marcodejongh:
Latest review round addressed in `945d4c00e`:

- **AbortError dead-lettering — confirmed, fixed.** An aborted request (app backgrounded mid-drain, AbortController timeout) is named `AbortError`, resolved no status, and dead-lettered a write the server never saw — against the queue's own never-lose-writes rule. It's now classified as a network failure (matched by name, since `DOMException` isn't reliable across RN runtimes) and retried. Test added.
- **401 retry — premise incorrect, documented instead.** The drainer's fetch IS token-refreshing: it goes through `authenticatedFetch` (lib/auth-interceptor), which dedup-refreshes and retries once before any error reaches classification, and a failed refresh force-signs-out — wiping the pending queue — so a 401 can't loop against a dead session. Now stated in a comment at the classification site as the review suggested.
- **Per-50-row transactions — confirmed, fixed.** Upserts now run one exclusive transaction per sync page (≤500 rows) instead of ten; the drainer can interleave between pages either way. Test updated to pin one-transaction-per-page.
- **Cursor validation — fixed.** `updatedAt` must be ISO-8601 (`z.string().datetime({ offset: true })`) and `syncSeq` a stringified integer, so malformed cursors fail validation instead of blowing up in the `::timestamp` cast.
- **`columns.length === 0` throw — keeping deliberately.** Total schema drift (every column unknown) means something is catastrophically wrong; advancing the checkpoint would silently drop data. Sync cycles are event-driven (foreground/reconnect), not a hot loop, so the retry cadence is bounded, and the failure is loud in telemetry.

Also: the `test-default` failure on the previous commit was the web `ErrorBoundary` retry-budget test — unrelated to this diff (timing-based flake); rerun triggered before this push and the new push re-runs everything anyway. Validation locally: mobile 3631/3631, affected suites green, mobile+backend typecheck green.

marcodejongh:
Triage of the nine-finding review round (verified each against code; fixes in `a485786bf`):

**Fixed:**
- **HIGH 3 (test gap — deletion triggers)**: real, and worse than stated. The hand-maintained test DDL was missing the `boardsesh_ticks` and `board_climbs` deletion triggers, and its unguarded `CREATE TRIGGER`s made the whole idempotent schema re-apply roll back on warm worker DBs — so DDL edits silently never landed locally. Both triggers are now in the test schema, all creates carry `DROP TRIGGER IF EXISTS` guards, and a new test pins the board_climbs reference-data tombstone (`user_id IS NULL`). Known remaining drift: the `*_follows` tables (and their triggers) still aren't in the test DDL.
- **LOW 2 (boardType)**: `syncClimbs`/`syncClimbStats` now allowlist boardType via `BoardNameSchema`, consistent with the tick/favorite resolvers.

**Refuted (with evidence):**
- **CRITICAL 1 (board_id)**: the local SQLite ticks DDL has no `board_id` column, the resolver docstring documents the omission, no mobile code reads it locally, and local rows never flow back to clobber the server (only `create` is enqueued, carrying full `SaveTickInput` incl. `boardId`; update/delete are online-only). Deliberate design, zero data loss.
- **HIGH 2 (writeTickLocal)**: the insert and the queue enqueue share one transaction; a failure rolls back atomically and rejects through the mutation's `onError`, which rolls back the optimistic entry. No silent path exists.
- **MEDIUM 1 (sign-out race)**: the ordering is deliberate and documented in the drainer — the flag exists to block scheduler-triggered drains during `clearUserData`; the one bounded sign-out drain intentionally runs before it.
- **MEDIUM 2 (composite-PK deletions)**: all nine tables' mobile parse matches the trigger encodings exactly (segment count, order, `:` separator); no key value can contain a colon.

**Tracked / deferred:**
- **CRITICAL 2 (stale checkpoint after prune)**: known — #3474, documented on `SYNC_DELETIONS_RETENTION_DAYS`.
- **HIGH 1 (UpdateTick payload whitelist)**: the `update` branch is currently unreachable (nothing enqueues tick updates; `useUpdateTick` is online-only), and GraphQL input coercion would reject extra fields anyway. Whitelist when offline tick edits land.
- **LOW 1 (CREATE INDEX without CONCURRENTLY)**: the migration's own comment documents this trade-off (drizzle wraps migrations in a txn; table sizes are modest).
- **LOW 3 (playlist-climbs cursor-collision test)**: test-coverage nicety, fine as a follow-up.

marcodejongh:
Latest round addressed in `5295eb547`:

- **Dead-letter badge bug — confirmed, fixed.** The per-climb pending count now filters `m.status = 'pending'` (a dead-lettered write belongs to the More screen's Sync issues retry, not the per-climb badge), and the drainer invalidates `['localTicks']`-family queries on both failure transitions, not just success — so with `staleTime: Infinity` the badge can no longer freeze.
- **`removeFavoriteLocal` cast — fixed**, uses expo-sqlite's typed `SQLiteRunResult` directly.
- **Test-reset exports in the public barrel — declining**: consistent existing pattern across the sync modules (`__`-prefixed, tree-shaken in release, used by sibling suites); moving them to a separate entry point is churn without behavior benefit.

With this, five review rounds have been triaged end-to-end: every confirmed finding is fixed and tested, every refuted one is answered with file-level evidence, and the remaining deferrals are tracked (#3474) or documented in-code. From my side this PR is merge-ready; further review rounds are surfacing progressively smaller findings on unchanged code.

marcodejongh:
Triage of the latest round — **no merge blockers; the critical is factually wrong**:

- **"`characteristics` missing from mobile SQLite schema — will crash sync" — refuted.** The mobile schema evolves via the versioned migration runner, and **migration 2** (`packages/mobile/src/db/migrations.ts:40-46`) adds `characteristics TEXT` via `ALTER TABLE` — deliberately not a v1 CREATE edit, so existing v1 databases pick it up. Every device runs migrations at startup, the pull-client integration harness stands up its DDL through the same `runMigrations`, and `search-climbs-local` already reads the column. The review looked only at the v1 CREATE TABLE.
- **"user_follows handler reads camelCase but payload is snake_case" — refuted.** The enqueue stores `{ followingId }` (`use-offline-mutations.ts:118`) — the queue payload is a separate object from the SQLite row, and it matches the handler exactly.
- **"deleteTick/removeFavorite tombstone triggers may not exist" — refuted.** `trg_ticks_delete` and `trg_favorites_delete` are both in `0143_offline_sync_backend.sql`, and both are now also in the backend test DDL with a covering test.
- **"deletions checkpoint non-atomicity → silent data inconsistency" — refuted on impact.** Replayed deletion records delete already-absent rows — a no-op; both replay paths are idempotent by design. Single-transaction-per-page is fine hardening, tracked below.
- **playlist_climbs NULL-owner tombstone**: the cascade-order scenario is already covered by the parent-uuid guard (`IF v_playlist_uuid IS NULL THEN RETURN OLD` — the playlists row is gone before any cascaded child delete fires). A NULL owner requires an orphaned `playlist_ownership` row, i.e. a pre-existing integrity break. Belt-and-braces guard tracked below.
- **Everything legitimate in the perf/hardening cluster** (cursor-index leading column, prune index, partial indexes, ticks logbook index, sync-trigger WHEN guard, toggleFavorite upsert, GraphQLError typing) → tracked as #3475.

CI is fully green on `5295eb547` (ci-status ✅). Standing by my earlier assessment: this PR is merge-ready.

marcodejongh:
Rebased on main again — head `c0c22cf90`. Main's gym-claiming feature (#3409) took migration **0143**, so ours is now **0144_offline_sync_backend** (same procedure as the 0141→0143 renumber: regenerated snapshot/journal chained after main's 0143, hand-tuned lock-safe SQL preserved as the body). The i18n catalogs union-merged (gymGrant + offline blocks). Validation: full typecheck green, mobile 3651/3651, backend sync suites green.

Heads-up: every additional day this stays open risks another migration-number race — third time now. Recommend merging while it's green.

marcodejongh:
Latest round, addressed in `283e85906`:

- **syncClimbStats raw `boardType` — confirmed, fixed.** My validBoardType sweep missed the three inline scope fragments in the stats resolver. No reachable behavior difference (validation throws before those lines, so raw === validated in every executing path), but it was inconsistent with syncClimbs and defeated the sweep's intent. All binds now use the validated value; zero raw `boardType` binds remain in the resolvers.
- **Prune timeout on shutdown — fixed.** The 5-minute initial prune delay now joins the intervals cleanup list, so a shutdown inside the window can't fire against a closing pool.
- **"Push token never deregistered on sign-out" — refuted.** `startTokenManagement` has no call sites, so no token is ever registered through this module — the no-op unregister is cleanup for a system that isn't wired yet. (The Live Activity token flow is a separate, session-scoped path with its own lifecycle.)
- **Colon-in-uuid tombstone fragility — declined.** The composite-key components are hex UUIDs, a board-name enum, and an integer — none can contain `:`; and the client's exact-part-count guard skips-and-warns rather than misdeleting if the impossible happens.

marcodejongh:
Latest round addressed in the new head:

- **Local tick-write failure → network fallback — implemented.** One correction to the framing: the tick was never *silently* lost — the rejection propagated to the mutation's `onError`, which rolled back the optimistic entry and showed the error toast. But the fallback is strictly better: `saveTickOffline` now reports the SQLite failure to telemetry and returns `null`, so an online tick still saves over HTTP when the local DB is broken; offline, the network attempt fails visibly exactly as before.
- **Both SQL hygiene items — done.** The checkpoint LIKE prefix and the `pragma_table_info` args are now bound parameters (the allowlist assert stays as defense in depth).

marcodejongh:
Latest round addressed in `32c7f2fbb`:

- **Sign-out vs in-flight pull race — confirmed, fixed.** Legit catch: the sign-out flag only guarded the drainer. Both pullSync page loops (tables + deletions) now check the shared `isSigningOut()` guard before fetching or writing each page, so a page can't land after `clearUserData` and resurrect the old user's rows. Test pins no fetches, no writes, no checkpoint advance while the flag is set.
- **Stale `favoritesStore` read** — real edge with bounded impact (server is idempotent; local self-heals on next sync). Reading SQLite as the local source of truth is a deeper refactor of the actions hook — added to #3475 rather than churning it here.
- **Tombstone-gap tracking** — already tracked as #3474 (and documented on `SYNC_DELETIONS_RETENTION_DAYS`); this is the third time it's been re-found, it is not being lost.
- **PR description said 0143** — updated to 0144 below/in the description.
- Remaining test-gap suggestions folded into #3475.

marcodejongh:
Triage of the latest round (fixes pushed in the new head):

**Fixed:**
- **C-1 (in-flight favorite-add race) — confirmed via a different mechanism than claimed.** There is no `draining` status; the real window is the TOCTOU between `peekPending` and `markCompleted`. `removeFavoriteLocal` now always enqueues the remove after canceling (server-side remove is an idempotent no-op), test updated. The symmetric add path already enqueued unconditionally.
- **C-5 (silent init failure)** — production SQLite init failures now report to telemetry.

**Refuted:**
- **C-3 (`clearAll` nukes drains)** — the queue's `clearAll` has no runtime caller; sign-out wipes via `clearUserData` deliberately, behind the pending-writes confirm.
- **C-4 (`clearInterval` on a timeout)** — Node documents `clearTimeout`/`clearInterval` as interchangeable; the code comments this explicitly.
- **C-6 (client omits layout/size scope)** — the pull client doesn't use the static operation constants; `buildSyncQuery` sends `$layoutId/$sizeId` per board scope (pull-client.ts:40-41).
- **C-7's "blocks the whole queue"** — a corrupt payload dead-letters and the queue continues; validation-before-dispatch is hardening, not a blocker.
- **S-1** — the ATTACH path comes from expo-asset's `localUri`, not user input; expo-sqlite has no parameterized ATTACH (the code documents this).
- **S-2** — the test DDL is only reachable from the worker-DB test bootstrap pointed at the docker test instance on :5433.

**Tracked in #3475:** C-2 (stale store read — re-find), C-7 zod hardening, P-1/P-2/P-3 (the P-2 partial-index idea is already listed), T-1/T-2.

Also: the `test-default` failure on the previous run is the `crypto` tamper-test — an unrelated flake, rerun triggered.

marcodejongh:
Triage of the latest review round (fixes in `9f1a3ed52`):

**Fixed:**
- **`lastSyncedAt` on partial failure — confirmed, fixed.** The scheduler's unstick `idle` frame is now marked `failed`; the status store only stamps `lastSyncedAt` for a completed cycle.
- **Playlist-cascade tombstone gap — real consequence, fixed client-side.** The NULL-parent guard skipping child tombstones on a whole-playlist delete is intentional and tested (the trigger can't read the mid-delete parent), but the reviewer is right that local `playlist_climbs` rows were orphaned forever (mobile SQLite has no FK cascade). The client now cascades locally when it applies a `playlists` tombstone — gated on the parent delete actually removing a row, so the resurrection guard still protects recreated playlists.
- **Missing cross-user isolation tests — confirmed, added.** All six previously-untested user-scoped resolvers now have leak tests (incl. a non-owner-role case for the ownership joins). The three follows tables were missing from the test DDL entirely; mirrored in.
- **`__resetDrainerStateForTests` wipe-epoch — added** (harmless either way: epoch checks are capture-then-compare, so a residual value can't misfire, but reset is cleaner).

**Factually wrong:**
- **`pending_mutations` UNIQUE constraint** — it exists: `idempotency_key TEXT NOT NULL UNIQUE` (`mutation-queue/schema.ts:9`). `INSERT OR IGNORE` dedupes as designed.
- **`error-classification` number-code branch** — the claimed consequence is backwards. When neither `extensions.code` nor `extensions.status` is numeric, `statusFromGraphqlErrors` returns null and the HTTP `response.status` (checked first) decides; a yoga GraphQL error at HTTP 200 resolves `status === 200` → **not** retryable → dead-letter, which is the intended classification. The numeric branch is a defensive extra, not a misclassification path.
- **`INSERT OR REPLACE` resetting local-only columns** — there are no local-only columns on any synced table (verified in an earlier review round); every local column is in the sync document allowlist.

**Wontfix (by design):**
- **UUID existence oracle** — the colliding UUID must already be known to the caller; v4 UUIDs aren't guessable (2^122). Returning fake success would corrupt the idempotent-replay contract for the legitimate owner.
- **Sync resolver rate limiting** — a board download is legitimately ~80+ sequential 500-row requests; a per-user throttle would break the feature it serves. All resolvers require auth and cap `limit` at 500. Can revisit with a generous ceiling if abuse shows up in telemetry.
- **Sign-out drain TOCTOU** — the remaining window is a single in-flight mutation send racing `clearUserData`, which is the *old user's own write being flushed* — exactly what the best-effort drain is for. The wipe-epoch guard (added last round) already stops the batch at the next mutation boundary, so no post-wipe batch can start and nothing can post under the next user's token.
- **Push token dereg** — the notifications module ships inert (nothing ever registers a token), so there is nothing to deregister; the no-op callback is the placeholder until that feature is actually wired.
- **Prune on every replica** — acknowledged in the PR deploy notes; the delete is bounded, indexed, and idempotent, and N is small.
- **`playlist_climbs` cursor index scoping** — real at scale, but the ownership join bounds the result set and the table is small today; noted as a follow-up index (`(playlist_id, updated_at, id)` or a partial) rather than a merge blocker.

marcodejongh:
Triage of the latest round (docs in `99a6a2498`) — **the "bug" is factually wrong, same claim as a previous round**:

- **`characteristics` "missing from mobile schema" — wrong.** The v1 `BOARD_CLIMBS` DDL intentionally omits it; **mobile migration 2 adds it** (`src/db/migrations.ts:45`, `ALTER TABLE board_climbs ADD COLUMN characteristics TEXT`), and the manifest documents the column as "schema v2". `initializeDatabase` runs migrations on every launch before any sync can start, so the column exists on every device — including fresh installs, where the base DDL and migration 2 apply in the same init. Covered by `db/__tests__/migrations.test.ts` and the real-DDL integration suite, which round-trips board documents through the migrated schema.
- **`sync_deletions` OR-branch index** — real but immaterial at current scale: btrees index NULLs, so Postgres services the `OR user_id IS NULL` as a BitmapOr of two probes on the existing `(user_id, deleted_at, id)` index, against a table pruned to 90 days of tombstones. Noted as a follow-up partial index (`(deleted_at, id) WHERE user_id IS NULL`) if reference-data tombstones ever dominate; not worth reopening the unshipped migration for.
- **Forced-sign-out drain skip — documented** in `docs/offline-sync-plan.md` as requested (dead token → drain could only fail; wipe stays unconditional as the shared-device safety boundary).
- **3s drain cap — comment added** on the constant: it's a UX bound, not a delivery guarantee; the user already saw the pending-writes dialog.

PR #2791, opened 2026-06-11
fix(mobile): don't resync the queue when switching climbs gets throttled

marcodejongh:
Rebased onto current `main` and rescoped. The original branch was six weeks stale; the unrelated Strava i18n and integrations-registry edits it carried have been dropped, and the `releaseControl` test it added is gone because `releaseControl` no longer exists anywhere in `packages/mobile` or `@boardsesh/queue-react`.

## What was wrong

In a party session, `setCurrentClimb`'s catch called `resyncQueueAfterMutationFailure()` for **every** failure, including `RATE_LIMITED`:

```ts
if (sessionIdRef.current) {
  void resyncQueueAfterMutationFailure();   // ← also fired on RATE_LIMITED
} else {
  showToast(t('mobile.queue.actionFailed'), 'error');
}
```

The rate-limit gate (`applyRateLimit`, `packages/backend/src/graphql/resolvers/shared/helpers.ts:471`) throws *before* the resolver does any work, so a throttled `setCurrentClimb` leaves the server exactly where it was. There is nothing to reconcile. Resyncing anyway did three bad things at once: yanked the climber back to the previous boulder mid-swipe, showed the alarming "Queue was out of sync. Refreshed from your crew." toast, and fired another query into the limiter that had just throttled us.

That is precisely how #2763 was reported — *"Me and my friend tried session, the connection fails everytime we try to switch boulders."*

## What changed

`setCurrentClimb` now routes the rate-limited case to the gentle `mobile.queue.rateLimited` toast and leaves the local pointer alone; the next activation or any peer's broadcast re-establishes it. Every other party-session failure still resyncs.

`showQueueMutationErrorToast` now classifies with the shared `isRateLimitedError()` instead of a bare `extensions.code` check. One definition of "is this a rate limit" across all call sites, and it also catches the nested-`graphqlErrors` and legacy message-only shapes — so a pre-#2777 server gets the gentle toast rather than "Action failed".

No new i18n keys: `mobile.queue.rateLimited` already exists in `en-US`, `es`, and `fr`.

## The sibling catch sites — checked, deliberately left alone

`resyncQueueAfterMutationFailure` has five other callers. I looked at each rather than fixing one and leaving the rest:

| Site | Verdict |
|---|---|
| `addToQueue` | **Keeps resyncing on any error.** Changes queue *content* with no local rollback. Skipping reconciliation would leave the climb in your queue and absent from everyone else's, permanently and silently. |
| `removeFromQueue` | **Keeps resyncing.** Same reasoning — a dropped item would linger on peers. |
| `setQueue` | **Keeps resyncing.** Replaces the whole queue; divergence here is the most visible of all. |
| `clearQueue` | **Keeps resyncing.** Coalesced single resync after `Promise.allSettled`; same content argument. |
| `reorderQueue` | **Already correct.** It rolls back to the pre-reorder order and calls `showQueueMutationErrorToast` — never resyncs. Only missing coverage, which this PR adds. |

The distinction that matters: `setCurrentClimb` moves a *pointer* that self-heals on the next activation or peer broadcast, so dropping the reconciliation costs nothing. The other four change queue *content* with no rollback path, so reconciliation is the only thing keeping the crew's queues consistent. Trading a correct-but-noisy toast for a silent permanent divergence would be a worse bug than the one being fixed.

## Follow-up worth filing separately

At those four content sites, a rate-limited mutation now fires a resync query that is itself likely to be throttled, so `resyncQueueFromServer()` returns falsy and **no toast fires at all** — the user gets silence. Fixing that properly means giving the resync path its own backoff rather than changing which errors reach it, which is a different change from this one. Filed as #3929.

## Testing

- `setCurrentClimb` rate-limited in a party session: keeps the local climb, fires **zero** queue-state refetches, shows the gentle toast, and shows neither `outOfSyncRefreshed` nor `actionFailed`. Verified this fails on `main` without the fix (stash-revert run) — it is a real regression guard, not a vacuous assertion.
- `reorderQueue` rate-limited: shows the gentle toast, not the generic failure. This one passes with and without the fix; it is coverage for existing untested behaviour, as the original PR intended.

## Validation

- `vp run test:mobile` — 542 files, 4473 tests, all passing
- `vp check` — clean for both touched files

`vp check` does flag `packages/mobile/public/index.html`, which is **pre-existing on `main`** and untouched here (this PR modifies exactly two files). Left for a housekeeping PR.

## Release Notes

Switching boulders in a session no longer throws you back to the previous climb. If you and your crew are flicking through problems fast enough for the server to ask you to slow down, you now get a quiet "Too many changes too fast. Give it a sec." instead of the alarming "Queue was out of sync. Refreshed from your crew." — and your climb stays put.

### DIFF
--- a/packages/mobile/src/providers/__tests__/queue-provider-session-updates.test.tsx
+++ b/packages/mobile/src/providers/__tests__/queue-provider-session-updates.test.tsx
@@ -1801,7 +1801,10 @@ describe('QueueProvider mutation-failure resync', () => {
     const snapshots: Snapshot[] = [];
     const first = makeQueueItem('item-1', 'climb-1');
     const second = makeQueueItem('item-2', 'climb-2');
-    routeHttpRequest(queueStateResponse([first, second], first));
+    let queueStateCalls = 0;
+    routeHttpRequest(queueStateResponse([first, second], first), {
+      onQueueStateCall: () => (queueStateCalls += 1),
+    });
     queueMutations.reorderQueueItem.mockRejectedValueOnce(makeRateLimitedError());
 
     renderProvider((snapshot) => snapshots.push(snapshot));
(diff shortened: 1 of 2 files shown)


marcodejongh:
Both review points addressed in 82c599ad5.

- **reorderQueue no-resync assertion** — added `expect(queueStateCalls).toBe(0)`. You're right that it can't fire today since reorderQueue rolls back locally, but the assertion is what keeps that true if the handler is ever rewritten to reconcile instead.
- **Solo + rate-limited setCurrentClimb** — added. The behaviour genuinely did change (solo used to get a blanket `actionFailed`, now it gets the same gentle wording), so it's worth pinning even if it's hard to hit in practice.

Full mobile suite: 542 files, 4474 tests, green.

marcodejongh:
Both points from the second pass done in b1ae4cfed.

1. **Solo generic-error path** — added a test pinning that a plain (non-rate-limited) solo failure still shows `actionFailed`. Routing solo through `showQueueMutationErrorToast` was meant to change only the throttled wording, and now that's guarded.
2. **Untracked gap** — filed as #3929 and linked from the PR description. Worth stressing it isn't a "route rate limits to the gentle toast" repeat: those four content mutations genuinely need to reconcile, so the fix has to make the resync path survive being throttled rather than change which errors reach it.

Test file now at 44 tests, full mobile suite green.

PR #2841, opened 2026-06-13
Board identity: duplicate-serial disambiguation + named-board onboarding (backend + mobile)

marcodejongh:
## Why

Part of firming up the collaboration model before the RN release. The LED supplier can't be trusted to keep serial numbers unique — the same serial can ship to multiple boards/gyms. Today a global unique index forces one serial → one board, so a user at gym B can get silently routed to gym A's board. This PR makes serials non-unique and routes a connecting user to the **right** `board_id`, asking them to pick when it's genuinely ambiguous.

This is the **backend foundation** of the board-identity pillar. It's additive and OTA-safe; the mobile/web disambiguation pickers and onboarding land in follow-up PRs.

## What changed

**DB (`packages/db`)**
- Drop the global `user_boards_unique_serial`; add `user_boards_unique_owner_serial` (per-owner partial unique — one owner still can't bind a serial to two of their own boards) + `user_boards_serial_idx` for lookups that now return many rows.
- Migration `0125_board_serial_non_unique` (drizzle-generated). No backfill — the old global unique guaranteed no cross-owner collisions exist.

**Backend (`packages/backend`)**
- `resolveBoardForSerial` now resolves via a shared precedence: remembered choice → 0 candidates (bind/create the caller's own board) → exactly 1 (route + remember) → several (prompt). Kept returning a single board, so **already-shipped OTA clients keep working** (auto-picks the caller-owned board, else the oldest).
- New `resolveBoardCandidatesForSerial` returns `{ board }` when unambiguous or `{ candidates }` when several boards share a serial.
- New `chooseBoardForSerial(boardId, serial)` confirms the user's pick and remembers it (`userBoardSerials.boardUuid`) so the prompt doesn't reappear.
- **Privacy:** private boards stay findable by serial (per product decision), but their location (`locationName`/gym) is redacted for users who don't own them.

**Schema (`packages/shared-schema`)**
- `BoardCandidate`, `ResolveBoardResult` types; `resolveBoardCandidatesForSerial` + `chooseBoardForSerial` mutations. Legacy `resolveBoardForSerial` unchanged in shape.

## Tests
- New disambiguation suite: candidates for shared serials (private boards included, location redacted), auto-resolve + remember for a single match, `chooseBoardForSerial` remembers, rejects an unrelated board, legacy auto-pick, and the per-owner-serial constraint (same serial across two owners OK; second board for one owner rejected).
- Added `gyms` + `user_board_serials` to the backend test-schema fixture (they were missing).
- `vp run typecheck`, `vp check`, backend board/serial suites all green (87 tests).

## Follow-ups (same pillar)
- Mobile disambiguation sheet + create-named-board parity + gym map onboarding.
- Web: minimal `ResolveBoardResult` handling to keep connect working.


marcodejongh:
### Now includes the full board-identity client work (mobile)

Three follow-up commits land the mobile clients on top of the backend foundation:

- **Disambiguation picker** — on BLE connect, when a serial maps to several boards the provider surfaces a `BoardDisambiguationSheet` (board name, gym/location redacted for non-owned private boards, "your board" tag) and remembers the pick via `chooseBoardForSerial`. The shared `BoardPresenceClient` interface stays minimal (web keeps the legacy single-board resolver).
- **Name your board on create** — `CustomBoardSheet` gains an optional board-name field (falls back to the layout name).
- **Gym-first discovery map** — new "Find gym" mode card → `/gyms` screen with `expo-maps` pins from `searchGyms` + a gym list; tap a gym → its boards → pick → active named board. `GymMap` lazy-requires expo-maps so an OTA bundle on an older native build degrades to a list-only flow instead of crashing.

Validation: `typecheck` clean, new files lint-clean, **1998 mobile tests green**, Metro bundle OK, 87 backend tests green.

Note: Android map needs `GOOGLE_MAPS_API_KEY` + a native build to render (iOS uses Apple Maps out of the box); the list works regardless.

PR #2842, opened 2026-06-13
Session simplification (R1): wall-connection holder model

marcodejongh:
## Why

Part of firming up the collaboration model. The driver / take-control machinery overloads the LED controller (every connected phone could write) and causes the session bugs flagged for the RN release. The fix: drop "who may drive" as a claimed role and make the **member holding the BLE connection the single frame writer**, with any holder lighting a shared "wall connected" indicator for the whole session.

This PR is **R1 — purely additive**. It introduces the connection-holder model *alongside* the existing driver machinery so nothing breaks. Driver/take-control removal and the client migration (lightbulb → session-connection indicator, AutoSender suppression for non-holders) come in follow-up steps.

## Model

**Claim-if-free, not yank.** The first connector to a board holds the slot (`HSETNX`); a later connector doesn't steal it. The physical BLE connection is already exclusive, so there's nothing to elect — the connection *is* the writer token.

## What changed (all additive)

- **distributed-state**: `sessionWallConnections` hash key + `claimWallConnection` / `releaseWallConnection` / `getWallConnections` / `releaseAllWallConnectionsForParticipant` ops, mirroring the driver ops (WATCH-guarded conditional release).
- **room-manager**: local-shadow map + the four methods + `reset`/`endSession` teardown. Disconnect cleanup folds into `releaseDriverIfMatches` (reusing the same multi-instance liveness gate) and publishes `WallConnectionChanged(null)` per freed board.
- **schema**: `announceWallLink(boardId)` / `revokeWallLink(boardId)` mutations, `WallConnectionChanged` event, `Session.wallConnections` + `WallConnection` type.
- **resolvers**: `announceWallLink` / `revokeWallLink` (claim/release + transition-gated broadcast); `buildSessionPayload` hydrates `wallConnections`.

## Tests
- `typecheck` + `vp check` clean.
- 61 session/payload/disconnect backend tests green (roomManager mocks updated for `getWallConnections`; `build-session-payload` asserts the Map→array mapping).

## Follow-ups (same pillar)
- **R1 cont.**: mobile (primary) — lightbulb becomes a session-connection indicator; call `announceWallLink`/`revokeWallLink` on BLE connect/disconnect; suppress the `BluetoothAutoSender` for non-holders. Web (minimal) — keep compiling.
- **R2**: remove `takeControl`/`releaseControl`/`driverParticipantId`/`DriverChanged` once OTA clients have migrated (the plan's staged deprecation).

### DIFF
--- a/packages/backend/src/__tests__/build-session-payload.test.ts
+++ b/packages/backend/src/__tests__/build-session-payload.test.ts
@@ -40,6 +40,7 @@ const getQueueStateMock = vi.fn();
 const getSessionDriverParticipantIdMock = vi.fn();
 const getSessionBoardSerialMock = vi.fn();
 const getSessionLeaderConnectionIdMock = vi.fn();
+const getWallConnectionsMock = vi.fn();
 
 vi.mock('../services/room-manager', () => ({
   roomManager: {
@@ -49,6 +50,7 @@ vi.mock('../services/room-manager', () => ({
     getSessionDriverParticipantId: (...args: unknown[]) => getSessionDriverParticipantIdMock(...args),
     getSessionBoardSerial: (...args: unknown[]) => getSessionBoardSerialMock(...args),
     getSessionLeaderConnectionId: (...args: unknown[]) => getSessionLeaderConnectionIdMock(...args),
+    getWallConnections: (...args: unknown[]) => getWallConnectionsMock(...args),
   },
 }));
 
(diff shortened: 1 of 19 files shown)


marcodejongh:
Superseded by the board-connection holder rework on `feat/board-connection-holder` (off latest main). The session-scoped, auto-claimed wall-writer model here is replaced by a board-scoped "who's connected/writing" holder derived from confirmed sends (anon-inclusive, explicit-disconnect, no auto-promotion). This branch was also 43 commits behind main. Plan + current state: `docs/board-connection-holder.md` on that branch. Closing.

PR #2843, opened 2026-06-13
Session simplification (R1 clients): wire connection-holder into mobile

marcodejongh:
**Stacked on #2842** (base: `feat/session-simplification`). Merge that first.

## Why
Makes the connection-holder model from #2842 actually drive behaviour on **mobile** (the primary session surface). The key fix: only the member holding the BLE connection writes frames, so several connected phones can't all hammer the LEDs ("lightbulb overload"). Additive — driver/take-control still works, so nothing breaks mid-migration. Web stays on the driver path (de-driver happens in the R2 PR).

## Shared (additive)
- **queue-runtime**: `RuntimeSessionState.wallConnectionsByBoard` + `WallConnectionChanged` handling in `applySessionRuntimeEvent`; new `deriveSessionWallConnected` / `deriveIsWallWriter` selectors.
- **graphql**: `ANNOUNCE_WALL_LINK` / `REVOKE_WALL_LINK` ops + `... on WallConnectionChanged` subscription fragment.
- **queue-react**: `announceWallLink` / `revokeWallLink` mutation methods.

## Mobile
- `queue-provider` exposes the new mutations + `wallConnectionsByBoard`.
- `bluetooth-provider` announces the wall link on BLE connect (resolved `boardId`) and revokes on disconnect — both user-initiated and unexpected-drop, so a dead holder can't suppress peers.
- **`BluetoothAutoSender` gates on `deriveIsWallWriter`** → only the connection holder writes frames. The selector falls back to "writer" when solo, no board is resolved, or nobody holds the slot, so there's **no behaviour change** when board-presence is off or you're the holder; the fix activates when board-presence is on and another phone holds the connection.

## Validation
Full `typecheck` (web+mobile+shared) clean · `vp check` clean · **1997 mobile + 58 shared runtime tests** green · Metro bundle OK.

## Follow-up (R2)
Remove `takeControl`/`releaseControl`/`driverParticipantId`/`DriverChanged`, flip the lightbulb to the shared connection indicator, minimal web de-driver.

### DIFF
--- a/packages/mobile/src/lib/__tests__/session-runtime-event.test.ts
+++ b/packages/mobile/src/lib/__tests__/session-runtime-event.test.ts
@@ -41,9 +41,23 @@ describe('toMobileSessionRuntimeEvent', () => {
     });
   });
 
+  it('adapts WallConnectionChanged so the holder reaches wallConnectionsByBoard', () => {
+    expect(
+      toMobileSessionRuntimeEvent({ __typename: 'WallConnectionChanged', boardId: 42, holderParticipantId: 'p-1' }),
+    ).toEqual({ __typename: 'WallConnectionChanged', boardId: 42, holderParticipantId: 'p-1' });
+    // A freed slot (nobody holds it) maps the holder to null.
+    expect(toMobileSessionRuntimeEvent({ __typename: 'WallConnectionChanged', boardId: 42 })).toEqual({
+      __typename: 'WallConnectionChanged',
+      boardId: 42,
+      holderParticipantId: null,
+    });
+  });
+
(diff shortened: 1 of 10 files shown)


marcodejongh:
Superseded by the board-connection holder rework on `feat/board-connection-holder` (off latest main). The session-scoped, auto-claimed wall-writer model here is replaced by a board-scoped "who's connected/writing" holder derived from confirmed sends (anon-inclusive, explicit-disconnect, no auto-promotion). This branch was also 43 commits behind main. Plan + current state: `docs/board-connection-holder.md` on that branch. Closing.

PR #2861, opened 2026-06-14
Drop accessory-bar swipe; fix doubled text on board-level climb change

marcodejongh:
## What & why

Two fixes to the iOS bottom **accessory bar**, both surfaced by the board-presence ("now on the wall" / lit-up climb) feature.

### 1. Remove the swipe carousel → tap-to-open only
The bar now reflects the *board's* status (the wall's lit climb via `useWallOrQueueCurrentClimb`), so the horizontal swipe-to-step-the-queue gesture with neighbouring climbs peeking in no longer makes sense. Removed it entirely — a single tap opens the displayed climb.

- `useQueueClimbCarousel` → slimmed + renamed to `useAccessoryClimbTap` (one tap gesture).
- `ClimbCapsule` / `NativeAccessoryClimbRow` render a single label, no peek slots, no prev/next a11y actions.

### 2. Fix the persistent doubled text on board-level climb change
When someone changed the climb at the board level, the iOS 26 native accessory showed two copies of the name+grade at rest, offset by a thumbnail width.

**Root cause:** commit `82186be6` gated the UIKit `NativeTabs.BottomAccessory` mount on `useHasActiveClimb()` (local-queue-only), while the accessory's *content* is wall-aware. On a board-level climb change the local presence could churn, unmounting → remounting the UIKit host mid-change, which left a stale `regular`-placement snapshot (with thumbnail) stacked under the new `inline` one (no thumbnail).

**Fix:** gate the mount and the bottom-chrome arbitration on a wall-aware, presence-only signal `useHasAccessoryClimb()` (local OR live wall climb). It stays continuously true across a board-level climb change, so the host is never remounted and no stale snapshot is left behind. It also lets a wall-only climb (no local queue) show the accessory.

A new presence-only `useBoardPresenceHasClimb` boolean context in `@boardsesh/board-presence-react` keeps the tab tree from re-rendering on every climb identity change (the perf constraint that the local/wall split contexts already respect).

## Arbitration after the fix
`hasAccessory = local OR live-wall climb`. On glass+tabs the native accessory owns the slot; on Material the JS bar does — never both. Wall-only climbs now correctly route to the native accessory instead of showing nothing.

## Accessibility note (deliberate tradeoff)
Removing the carousel also removes the accessory bar's VoiceOver `next`/`previous` queue-step actions (they were the non-visual equivalent of the swipe). This is intentional and consistent with the product change — the bar now reflects board status, not a queue stepper. Queue navigation is unchanged elsewhere: the play drawer (opened by tapping the bar) keeps its prev/next controls, so no navigation is lost outright, only the accessory-bar shortcut.

## Tests / validation
- `vp run typecheck:shared`, `vp run typecheck:mobile` ✅
- `vp test run --project mobile` (full suite) ✅
- `@boardsesh/board-presence-react` tests, incl. new presence-only `hasClimb` behaviour (stays `true`/no re-render across a climb identity change) ✅
- Direct unit test for `useHasAccessoryClimb` (local/wall OR matrix) ✅
- `vp check` (format/lint), `vp run check:mobile-bundle` ✅

## On-device QA (not coverable on Linux CI)
The doubled-snapshot itself is a UIKit tab-bar-accessory artifact — only fully verifiable on an **iOS 26 Liquid Glass** device/simulator. With `board-presence` on, connect a board and trigger several rapid board-level climb changes; confirm exactly one copy of name/grade/tick at rest in both expanded and minimized states, that tap still opens the play drawer, that a wall-only climb shows the accessory, and that no climb at all leaves no empty glass platter.

https://claude.ai/code/session_01Kr5ZiXbCMnjfoER6BWWMk1

---
_Generated by [Claude Code](https://claude.ai/code/session_01Kr5ZiXbCMnjfoER6BWWMk1)_


marcodejongh:
Thanks — quick notes on the three points, all no-action:

- **`BoardPresenceProvider` scope (potential hard throw):** verified safe. `MobileBoardPresenceProvider` *unconditionally* renders the shared `BoardPresenceProvider` (`packages/mobile/src/providers/board-presence-provider.tsx:308` — only the `boardId` prop is flag-gated, never the provider), and it wraps both the `<Stack>` (all screens incl. `(tabs)`) and the root `<PersistentQueueBar />` (`app/_layout.tsx:317`/`:325`/`:366`). So `useHasAccessoryClimb` → `useBoardPresenceHasClimb` can never be called outside the provider; with the flag off it just returns `false`. Same dependency that `useIsWallPinned`/`useWallOrQueueCurrentClimb` already rely on app-wide.

- **Accessibility:** intentional and now documented in the PR description as a deliberate tradeoff — queue navigation still lives in the play drawer (opened by tapping the bar), so it's a removed shortcut, not lost navigation.

- **Test comment placement:** the `// …wraps its open handler in runOnJS — return it as-is.` comment sits directly above the `react-native-reanimated` mock and describes that `runOnJS` stub (not the gesture mock below), so it's correctly placed.

---

PR #2862, opened 2026-06-14
feat(board-presence): connection holder — anon-first-class, lightbulb toggle, GA

marcodejongh:
## What

Ships the **board connection holder** — "who's connected / writing to the wall" — and takes board presence **GA** (the `board-presence` flag is removed entirely). The holder is the emitter (`userId`, else `conn:{connectionId}`) of the most recent confirmed send, shown to everyone watching the board, and cleared when that connection drops. Supersedes #2842/#2843.

## Backend
- **Auth-optional connect/read.** `resolveBoardForConfig` / `resolveBoardForUuid` drop `requireAuthenticated` and stamp membership by emitter id (anon → `conn:{id}`). Serial + create/own paths stay auth-only. `reportBoardClimb` / `reportBoardDisconnect` were already auth-optional.
- **Anon-watchable live feed, privacy-gated.** `boardNowPlaying` / `boardRecentClimbs` / `boardConnection` are now anonymous-readable, but `requireAnonReadableBoard` restricts anonymous viewers to **public + system-shared** boards so they can't enumerate ids to watch a *private* wall. Logged-in users keep the pre-existing membership-free reads; `boardHistory` / `boardPresenceStats` stay auth-required.
- **Crash backstop.** A holder's wall is freed on WebSocket close via a single `onDisconnect` chokepoint (covers solo *and* party holders), using an emitter-keyed atomic compare-and-delete so a takeover (always-take) or a non-holder/other-connection drop is a no-op. The clean path remains the client's BLE-drop `reportBoardDisconnect`.

## Flag removal (GA, no kill-switch)
Removed the `board-presence` PostHog key (web + mobile), the backend `BOARD_PRESENCE_ENABLED` env gate / `requireBoardPresenceEnabled`, and `EXPO_PUBLIC_BOARD_PRESENCE`. Board presence is now always-on.

## Shared packages
- `@boardsesh/board-presence`: `BoardConnectionChanged` → seq-gated `APPLY_CONNECTION_CHANGED` / `SEED_CONNECTION`; reducer gains `holder` + `lastConnectionSeq`.
- `@boardsesh/board-presence-react`: `holder` exposed via `useBoardPresenceCurrent`; optional `fetchConnection` / `reportDisconnect` client methods (web's client is unchanged — they're optional).
- `@boardsesh/graphql`: `BOARD_CONNECTION` query, `REPORT_BOARD_DISCONNECT` mutation, and a `BoardConnectionChanged` fragment on `BOARD_NOW_PLAYING`.

## Mobile
- **Lightbulb = connect/disconnect toggle** (lit iff BLE connected), replacing the party driver take/release. Connecting auto-pushes the climb and takes the board; pressing again disconnects and frees it.
- **Emit always** on every confirmed send; `reportBoardDisconnect` on both explicit and unexpected BLE drops.
- **"Who's connected" badge** next to the lightbulb: holder avatar, `?` for anonymous holders, and an idle `?` overlay once the wall hasn't changed for 15 min.

## Tests
- Backend: anon emit (member + non-member), holder hand-off + same-emitter dedup, `reportBoardDisconnect` clear-if-holder, `boardConnection` set/clear, crash backstop, and the anon read-gate (private board hidden from anon, system board allowed). Redis-dependent assertions are gated on a live Redis (run in CI, skip locally); env-gate tests removed.
- Shared reducer/hook tests cover the new holder paths; web + mobile provider tests updated for always-on.

## Validation
`vp run codegen` (no diff), `vp run typecheck` (all packages), `vp check`, `vp test run --project backend` (55 board-presence tests pass with Redis), `vp run test:mobile` (2013 pass), `vp run check:mobile-bundle` (OK). The only backend failures locally are the pre-existing/environmental `integration.test.ts` daemon WS suite (unrelated to this change).

## Known follow-up
Anonymous WebSocket connections aren't per-IP rate-limited (the limiter keys on a per-connection UUID that resets on reconnect) — a pre-existing limiter weakness this change makes reachable. Tracked in #2863.


marcodejongh:
Pushed fixes for the CI failure and the automated review findings (commits `e70a835`, `d14604d`):

**Fixed**
- **CI i18n** — removed the orphaned `playView.actionBar.lightbulb.reconnect` key (en/es/fr); the mobile lightbulb no longer renders the driver "reconnect" label (web still uses driving/take).
- **Codex P1 — anon board creation** — `resolveSharedBoardForConfig` is now bind-only for anonymous callers; create-on-miss requires auth, so an anon client can't mint arbitrary system boards.
- **Codex P1 — party preview blocks the lightbulb** — `BluetoothAutoSender` now mounts on `isConnected` alone (dropped the `!isPartyPreviewOnly` write-gate), so a connected non-driver writes the wall and becomes the holder (always-take, per the doc).
- **Claude — non-atomic setBoardWriter** — switched to `SET … EX … GET` so concurrent sends can't double-broadcast a hand-off.
- **Claude — dedupe `assertValidBoardId`**, **dead `enabled` branches in `board-presence-context.tsx`**, **`BoardConnectionBadge` unit test**, **`reportBoardDisconnect` unheld-board test**, and **rewrote `docs/board-connection-holder.md`** from plan → reference docs.

**Deferred / by-design (not changing)**
- **Same-user multi-connection writer cleanup (P2)** — accepted tradeoff under always-take: either connection's drop frees the wall, re-acquired on the next send. Documented in `ConnectedClient.boardWriterEmitter`.
- **`boardConnection` returns null for an unknown board (authenticated)** — deliberate: it's a late-joiner initial-state probe that fires at any id; null = "no holder" is fine.
- **`boardConnection` reads the recent-climbs list to take `[0]`** — the list is capped at 50 and this is a low-frequency late-joiner query; left as-is.
- **Anonymous WebSocket per-IP rate limiting** — tracked in #2863.

marcodejongh:
✅ **CI green** (`ci-status`, i18n, lint, test-backend, typecheck, mobile-bundle, claude-review, all build/test shards). Only the non-blocking EAS `prebuild-and-build` is still running.

Note for reviewers — this PR was red purely because **main's CI was red** and GitHub tests the merge-with-main. To unblock it I merged latest `main` and fixed three pre-existing main breakages here (so merging this also un-reds main):
- **i18n** — `mobile/app/auth/register.tsx` translated validation errors via dynamic `t(errorKeys.x)` (orphan-checker hard-fails on `t(variable)`); replaced with a static switch. Also removed orphaned `mobile.queue.nextClimb`/`previousClimb` keys left behind when the accessory-bar carousel was dropped (#2861).
- **lint** — reformatted committed whitespace drift in `packages/db/drizzle/meta/*.json`.
- **test-backend** — added `scan` + `multi().setex` to the test mock Redis (DistributedState heartbeat/cleanup now call them).

These are the only out-of-feature files in the diff; the board-presence feature itself is unchanged from the reviewed state.

PR #2865, opened 2026-06-14
fix(board-presence): play-view crash + session board control (regressions from #2862)

marcodejongh:
## Hotfix for a production crash from #2862

**Error:** `useBoardPresenceCurrent must be used within a BoardPresenceProvider` (play view crash on open).

### Root cause
`PlayDrawer` (and `BoardSheet`) render through gorhom's `BottomSheetModal`, which **portals** content to the modal host — so they only see context available *at the host*, not at their declaration site. `MobileBoardPresenceProvider` was mounted **below** `BottomSheetModalProvider` in `app/_layout.tsx`, so the portaled play drawer couldn't see it. #2862 added `BoardConnectionBadge` (the first `PlayDrawer` consumer of `useBoardPresenceCurrent`), which then threw and crashed the screen. (The Board/Queue providers were already placed *above* the modal host precisely for this reason — board-presence was the one that wasn't.)

### Fix
- **Move `MobileBoardPresenceProvider` to wrap `BottomSheetModalProvider`** so portaled sheets read the wall state through context. `BoardDisambiguationSheet` is a plain RN `Modal`, so it doesn't need the bottom-sheet host above it; the DevicePicker/modal-host ordering is preserved. This also fixes the same latent crash in `BoardSheet`.
- **Defense-in-depth:** `BoardConnectionBadge` now reads `BoardPresenceCurrentContext` directly (non-throwing) and renders nothing when no provider is in scope, so a stray portal can never crash a screen again.

### Validation
`vp run typecheck:mobile`, `vp run test:mobile` (2050 pass), badge/play-drawer suites, and `vp check` on the touched files — all green.

### DIFF
--- a/packages/mobile/app/_layout.tsx
+++ b/packages/mobile/app/_layout.tsx
@@ -303,15 +303,17 @@ function RootLayout() {
                                     exist before the picker mounts or gorhom
                                     throws "BottomSheetModalInternalContext
                                     cannot be null". */}
-                                      <BottomSheetModalProvider>
-                                        {/* Board presence ("now on the wall") owns the
-                                    connected boardId + the wall feed. Mounted
-                                    OUTSIDE BluetoothProviderWrapper so the BLE
-                                    flow can resolve+report through it, and
-                                    OUTSIDE DrawerHostProvider so the Board sheet
-                                    can read the wall state. Inert (boardId null)
-                                    until the `board-presence` flag is on. */}
-                                        <MobileBoardPresenceProvider>
+                                      {/* Board presence ("now on the wall") owns the
+                                    connected boardId + the wall feed. Wraps
+                                    BottomSheetModalProvider so gorhom-portaled
+                                    sheets (PlayDrawer, BoardSheet) — which render
+                                    at the modal host, not their declaration site —
+                                    can still read the wall state through context.
+                                    Also OUTSIDE BluetoothProviderWrapper /
+                                    DrawerHostProvider so the BLE flow + Board sheet
+                                    can use it. */}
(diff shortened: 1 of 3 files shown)


marcodejongh:
Added a second regression fix to this PR (they're coupled — the lightbulb-holder read needs the provider move below).

## 2) Board doesn't update in a session
**Symptom:** in a multi-user session, connecting to the board no longer lit/updated the wall; solo worked; leaving the session fixed it.

**Cause:** in a 2+ person session every member is `isPartyPreviewOnly` until someone "takes wall control" (the driver). A preview-only member's navigation only sets a local preview, never `state.currentClimbQueueItem` — the only thing the BLE auto-sender watches. The old lightbulb's `takeControl` flipped you out of preview-only; the new connect/disconnect lightbulb doesn't, so everyone stayed preview-only and the wall never updated.

**Fix (per the no-driver model — every session member controls the board instantly):**
- `queue-provider`: `isPartyPreviewOnly` is now always `false`. The backend accepts queue/current-climb mutations from any participant (last write wins), so navigation/activation/queue edits drive the shared wall for everyone, exactly like solo.
- `PlayDrawer`: the lightbulb reads **lit if any member holds the board** (the board-presence holder), not just this device — so the wall shows connected when a peer is connected. The press still keys on this device's BLE.

Validated: `vp run test:mobile` (2050 pass), typecheck:mobile, `vp check` — all green.

> Note: this retires the driver/preview gate at its source (the hook returns false) rather than ripping the concept out of all 7 consumers + the backend `takeControl`/`DriverChanged` plumbing — that fuller cleanup is a good follow-up.

PR #2866, opened 2026-06-14
Strip query params from Instagram beta-video URLs

marcodejongh:
## Problem

When a user shares an Instagram reel, the URL carries a share-attribution param:

```
https://www.instagram.com/reel/DZlVfVhxKJZ/?igsh=NHB5ZXljZjV3bzB3
```

We stored and returned that URL verbatim in `board_beta_links.link`. Opening a link with `?igsh=...` makes Instagram show its "X shared this reel, want to follow them?" interstitial instead of going straight to the reel.

## Fix

- Add `normalizeBetaVideoUrl()` to the shared beta-url helpers (`packages/shared-schema/src/beta-video-url.ts`). It strips the query string and fragment from Instagram URLs (preserving `/reel/`, `/p/`, `/tv/` and the host) and returns everything else (TikTok, unknown) unchanged. Idempotent.
- **Save side** (`ticks/mutations.ts`): normalize after Zod validation in `attachBetaLink` and `saveTick`, so the dedup probe, shortcode extraction, and the stored `link` all use the clean URL. The `(boardType, climbUuid, link)` primary key now also collapses param-only duplicates of the same reel.
- **Return side** (`beta-videos/queries.ts`): normalize the output `link` at each result-construction site (`betaLinks`, `recentBetaLinks`, `betaLinkPreview`), so already-stored legacy rows display clean too. The `row.link` used for DB writes/`persistEnriched` lookups stays raw so legacy rows keep matching.

No DB migration: read-time normalization handles existing rows for display, and `dedupeBetaLinks` already collapses any duplicate legacy rows in the UI by shortcode.

## Tests

- `normalizeBetaVideoUrl` unit tests (strips `?igsh=`/`#frag`, leaves clean IG / TikTok / unknown URLs unchanged, idempotent, handles `instagr.am` and no-trailing-slash).
- Backend resolver tests asserting `betaLinks` and `recentBetaLinks` return the stripped link.

## Validation

- `vp test run` — shared-schema (52) + backend beta-link suites (52) pass
- `vp run typecheck:backend`, `typecheck:shared` — clean
- `vp check` on changed files — no lint/format/type warnings

https://claude.ai/code/session_01TJELQRAzuxXchcXM5yooat

---
_Generated by [Claude Code](https://claude.ai/code/session_01TJELQRAzuxXchcXM5yooat)_

### DIFF
--- a/packages/backend/src/__tests__/beta-link-queries.test.ts
+++ b/packages/backend/src/__tests__/beta-link-queries.test.ts
@@ -689,6 +689,23 @@ describe('betaLinkPreview resolver', () => {
     });
   });
 
+  it('strips the Instagram share-attribution param from the echoed link', async () => {
+    fetchInstagramMetaMock.mockResolvedValueOnce({
+      status: 'ok',
+      thumbnail: 'https://scontent.cdninstagram.com/raw.jpg',
+      username: 'climber',
+      caption: 'Sent it',
+    });
+
+    const result = await betaLinkQueries.betaLinkPreview(
+      undefined,
+      { link: 'https://www.instagram.com/reel/ABC123/?igsh=NHB5ZXljZjV3bzB3' },
+      ctx(true),
+    );
+
+    expect(result.link).toBe('https://www.instagram.com/reel/ABC123/');
+  });
+
   it('throws when unauthenticated and never calls out to Instagram', async () => {
     await expect(
       betaLinkQueries.betaLinkPreview(undefined, { link: 'https://www.instagram.com/reel/ABC123/' }, ctx(false)),
--- a/packages/backend/src/__tests__/beta-link-queries.test.ts
+++ b/packages/backend/src/__tests__/beta-link-queries.test.ts
@@ -90,6 +90,7 @@ vi.mock('../lib/instagram-meta', async () => {
     fetchInstagramMeta: fetchInstagramMetaMock,
     isInstagramUrl: shared.isInstagramUrl,
     getInstagramMediaId: shared.getInstagramMediaId,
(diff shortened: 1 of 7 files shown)


marcodejongh:
Thanks for the reviews. Dispositions on the points raised across the rounds:

**Addressed**
- `betaLinkPreview` now normalizes once up front and uses the canonical link for the `isInstagramUrl`/`isTikTokUrl` checks and the meta fetch, not just the echoed result.
- Save-side coverage: added `beta-link-save-normalization.test.ts` asserting `attachBetaLink` stores `link` and `shortcode` stripped of `?igsh=`. Return-side now has tests for both `enrichRow` branches (cached short-circuit + fresh fetch), plus `recentBetaLinks` and `betaLinkPreview`.
- Replaced the `validated.link` mutation with a local `normalizedLink` const.

**Confirmed non-issue: ticks `videoUrl` column** — there is no `videoUrl`/`video_url` column anywhere in the db schema (`grep` over `packages/db/src/schema/`), and the `boardseshTicks` insert in `saveTick` doesn't write one. The shared URL only ever lands in `board_beta_links`, which is normalized. Nothing to fix here.

**Not changing: multi-line comments** — the cited "one short line max" / "no multi-line comment blocks" rule isn't in `CLAUDE.md`, `docs/`, or `.claude/` (verified by grep). The actual guidance is to match surrounding comment density, and `beta-video-url.ts` uses multi-line JSDoc throughout (`betaLinkIdentity`, `dedupeBetaLinks`, `mapBetaLinkRow`), so the new JSDoc is consistent with the file.

---

PR #2366, opened 2026-05-30
Share queue transport runtime + port five web providers to mobile

marcodejongh:
## Summary

Web and mobile both reimplemented the transport wiring around the shared `@boardsesh/queue` state machine. This PR extracts that duplication into a new `@boardsesh/queue-runtime` package, then ports five web providers to mobile (with two more shared packages) so mobile has the same provider surface available — even where no mobile UI consumes them yet.

Follows the plan at `~/.claude/plans/serene-sleeping-avalanche.md`. Branched off `claude/pr-2338-code-reuse-3Ebx2`; targeting `main` as requested.

## Workstream A — shared queue runtime

New `@boardsesh/queue-runtime` package (pure TS, depends only on `@boardsesh/queue`):

- **`mapSubscriptionEnvelopeToAction`** — wire envelope normaliser. Web's `SubscriptionQueueEvent` and mobile's locally-defined event union describe the same protocol but each carried its own item shape; lifting happens in one place now. Mobile's `queue-provider.tsx` shed ~110 lines of inline wire-type definitions.
- **`createSetCurrentClimbCoalescer`** — serialize-and-supersede extracted from web's `use-queue-mutations.ts:78-159`. Rapid swipes used to stack SET_CURRENT_CLIMB mutations; only the latest wins, but superseded args carrying \`shouldAddToQueue:true\` still fire their ADD_QUEUE_ITEM. **Mobile picks up coalescing it previously lacked.**
- **\`createJoinSessionTracker\`** — promise cache keyed by \`(sessionId, epoch)\`. Caller bumps the epoch from the socket's \`closed\` handler so a mutation racing between \`closed\` and \`connected\` doesn't await a stale-resolved JOIN_SESSION from the dead connection. Mobile's inline 80-line dance collapses to a few lines.

Web's \`use-session-lifecycle.ts\` has a parallel join implementation; adopting this tracker there is a follow-up.

**Skipped: A4 mutation firers.** After A1-A3 each mutation is 2-5 lines around \`execute(...)\`. Extracting them into shared factories would add an abstraction layer that costs more than it saves.

## Workstream B — five web providers ported to mobile

Mounted in \`packages/mobile/app/_layout.tsx\` between \`ThemeProvider\` and \`QueueProvider\`:

| Provider | Notes |
|---|---|
| \`FeatureFlagsProvider\` | mirrors web's 19-line provider; static defaults until a mobile flags source lands |
[...]

PR #2907, opened 2026-06-16
fix(board-presence): skip auth-required serial-resolve for logged-out web users

marcodejongh:
## What & why

Follow-up "web parity" cleanup for the board connection holder feature (refs #2366). A blast-radius audit of the web fleet flagged three consistency gaps where web lagged the mobile holder implementation. **Two of the three were already merged to `main`** in the holder work (web `reportBoardDisconnect` on BLE drop — `afb5f645d`; driver/preview retired — `df382e8da`). This PR closes the remaining two.

### 1. Anonymous serial-resolve no longer fires a guaranteed-failing mutation
A logged-out climber on a serial board (Kilter/Tension) called `resolveBoardForSerial` on **every BLE connect**. That mutation is auth-required server-side (it can create/own a board — `board-presence/mutations.ts:235` `requireAuthenticated`), so for anon it always rejected, plus emitted a `console.warn` each time.

`WebBoardPresenceProvider.resolveAndBindBoard` now gates the serial branch on `isAuthenticated` (read from the existing `useWsAuthToken()`). Anonymous users fall back to `resolveBoardForConfig`, which **is** anon-allowed (`mutations.ts:362`, no auth gate) — so they still get a shared per-config wall feed instead of a failing request. `resolveKey` now derives from the branch actually taken, so the same serial re-resolves via the serial path once the user logs in (no stale dedup).

### 2. Stale flag-comment cleanup
The `board-presence` feature flag was fully removed when the feature went GA. Comments across `board-presence-context.tsx`, `board-presence-panel.tsx`, `board-sheet.tsx`, `bluetooth-context.tsx`, and `persistent-session-wrapper.tsx` still described flag-gating. Reworded to the real "inert until a board is bound" behaviour. (Legitimate non-flag mentions — the boolean `connected` param, the e2e sessionStorage flag — left untouched.)

## Out of scope
- The web `reportDisconnect`-on-BLE-drop work and the driver/preview retirement are already on `main`.
- Folding the presence feed onto the session socket (audit item #4) — noted optimization only, not here.

## Tests
- Two new cases in `board-presence-context.test.tsx`: anon + serial → routes to `resolveBoardForConfig`; signed-in + serial → `resolveBoardForSerial`.
- `vp test run --project web` board-presence + bluetooth suites: 430 pass.
- `vp check` (lint+format) clean on all changed files; `vp run check:i18n` clean (no new strings).

## Note for CI
`vp run typecheck:web` currently has a **pre-existing** failure on `main` unrelated to this PR: `persistent-session/__tests__/event-processor-session-cache.test.ts:73` (`tickUuid` type, from session-feature commit `58b206969`). Verified it reproduces on a clean `origin/main` checkout with these changes stashed.

### DIFF
--- a/packages/web/app/components/board-bluetooth-control/__tests__/bluetooth-board-presence.test.tsx
+++ b/packages/web/app/components/board-bluetooth-control/__tests__/bluetooth-board-presence.test.tsx
@@ -1,7 +1,7 @@
 // Board-presence BLE wiring: on connect → resolveAndBindBoard, and on
 // wall-confirm → reportClimb + Undo snackbar. Mirrors the mobile
-// bluetooth-provider's presence coverage. All flag-gated: the report/resolve
-// paths are inert when `enabled` is false, so the BLE flow behaves as today.
+// bluetooth-provider's presence coverage. The report/resolve paths are inert
+// until a board is bound (boardId null), so the BLE flow behaves as today.
 
 import { describe, it, expect, vi, beforeEach } from 'vite-plus/test';
 import { render, waitFor } from '@testing-library/react';
@@ -90,7 +90,6 @@ vi.mock('@boardsesh/play-view', () => ({ emitWallConfirm: vi.fn() }));
 
 // The presence controls + wall report — the surface under test.
 const presence = vi.hoisted(() => ({
-  enabled: false as boolean,
   boardId: null as number | null,
   currentClimb: null as BoardPresenceClimb | null,
   resolveAndBindBoard: vi.fn().mockResolvedValue(null),
@@ -99,7 +98,6 @@ const presence = vi.hoisted(() => ({
 }));
 vi.mock('../../board-presence/board-presence-context', () => ({
   useBoardPresenceControls: () => ({
-    enabled: presence.enabled,
     boardId: presence.boardId,
     resolveAndBindBoard: presence.resolveAndBindBoard,
   }),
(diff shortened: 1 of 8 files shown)


marcodejongh:
Addressed the review feedback in `6638dd400`:

**1. Dead code in board-presence-panel.tsx (`enabled`)** — Removed `enabled` entirely rather than just dropping the panel branch, since it was hardcoded `true` and dead in *every* consumer (the panel guard plus three `presenceEnabledRef` gates in bluetooth-context). The panel guard is now `if (boardId === null)`, and the bluetooth gates collapse to their real `boardDetails`/`boardId` checks. Behaviour-preserving. Dropped the two now-impossible "flag off" tests and repaired one stale fixture (`createTestBoardDetails` had `set_ids` as a string; the now-ungated resolve calls `set_ids.join`, so it's a `number[]` now — matching production).

**2. Auth-transition behaviour** — Added a comment in `resolveAndBindBoard` making the limitation explicit: a pure auth flip (login without a BLE reconnect) does **not** re-resolve on its own; `resolveAndBindBoard` only runs on a fresh connect. The next connect upgrades the anon config feed to the serial feed because `resolveKey` flips `config:`→`serial:`, so the dedup guard lets the serial resolve through.

**3. Missing transition test** — Added `upgrades the anon config feed to the serial feed on reconnect after login`: anon+serial binds the config board (43); then a re-render picks up the authed state and a second resolve for the same serial routes through `resolveBoardForSerial` and rebinds to the serial board (42).

Note: `typecheck:web` is red on `main` from an unrelated pre-existing beta-links fixture (`event-processor-session-cache.test.ts`) — fixed separately in #2910. Once that merges this branch's typecheck is green.

PR #2911, opened 2026-06-16
iPad: center the climb tick bar, size up the tick FAB, narrow the tick sheet

marcodejongh:
## Problem (iPad)

On iPad the log-ascent **tick FAB** ended up mis-placed (far to one side) and small, and the **tick sheet** stretched the full screen width.

Root cause: on iPad the climb + tick ride the iOS 26 **system bottom accessory** (`NativeTabs.BottomAccessory`). iPadOS positions that accessory against its own tab-bar layout, where it lands off to a corner and undersized; and `LogAscentSheet` (shared by every ticking entry point) is a full-width bottom sheet.

## Fix

- **Opt iPad out of the native bottom accessory** (`isBottomAccessoryAvailable()` → false on iPad). The native tab bar stays; the climb + tick fall back to the JS `PersistentQueueBar`, which we control. `jsQueueToolbarVisible` already flips on when the accessory isn't mounted, so space reservation is handled.
- **Center the floating glass bar on iPad** at a phone-like max width (`IPAD_TOOLBAR_MAX_WIDTH = 440`) so the capsule + tick read as one cluster instead of spanning the display.
- **Bump the hero tick FAB to 64 on iPad** (`IPAD_TICK_FAB_SIZE`) — it's already a glass `GlassIconButton`; the bar's trailing slot widens to match.
- **Cap `LogAscentSheet` at 480 and center it on iPad** so the tick form isn't awkwardly wide.
- New `isIpad` helper centralizes the check. **Phone / Android unchanged.**

## Validation

`vp run typecheck:mobile` · queue/accessory tests (`persistent-queue-bar`, `use-bottom-accessory`, `queue-bottom-accessory`, `native-accessory-climb-row`, `tab-layout` — 53 pass) · `vp run check:mobile-bundle` (Android OK) · `vp check`.

Device QA on a real iPad still recommended — in particular confirm the gorhom sheet centers (vs. just narrows) on iPad.

### DIFF
--- a/packages/mobile/src/components/LogAscentSheet.tsx
+++ b/packages/mobile/src/components/LogAscentSheet.tsx
@@ -17,6 +17,7 @@ import {
 import { FullWindowOverlay } from 'react-native-screens';
 import { useTranslation } from 'react-i18next';
 import { useTheme } from '../providers/theme-provider';
+import { isIpad } from '../lib/is-ipad';
 import { iosSystemColors } from '../theme/ios-colors';
 import { spacing } from '../theme/tokens';
 import { Icon } from './Icon';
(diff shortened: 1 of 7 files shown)


marcodejongh:
Closing — these iPad tick FAB/sheet fixes were made on the wrong branch by mistake. Reverting; not part of the mobile user-profiles work.

PR #2904, opened 2026-06-15
feat(mobile): adaptive iPad shell — glass sidebar replaces the tab bar

marcodejongh:
**Phase 1 of the iPad app** (follows the merged backend wiring in #2901). Enables iPad and gives it a first-class **left glass sidebar** in place of the bottom tab bar — the foundation the Gym Wall (Phase 2) and the Climbs two-pane browser (Phase 3) build on.

## The no-regression guarantee

The whole thing is gated behind one size-class hook. **Compact width — every iPhone, plus an iPad in Slide Over or a narrow Split View — renders today's `NativeTabs` / `MaterialTabBar` tree verbatim.** The sidebar only appears when the window is genuinely wide enough for two columns. No phase can regress the phone app.

## What's in this PR

- **`size-class.ts` (pure) + `use-device-layout.ts`** — `compact | regular` from a 700pt threshold + `Platform.isPad`, with an `expanded` flag at 1024pt for the future master+detail pane. Mirrors the proven `bottom-chrome-metrics.ts` pure-arbitration-plus-thin-hook pattern. Unit-tested with the ⅓/½/⅔/full iPad split widths.
- **`computeBottomChromeMetrics` → optional `usesSidebar`** — collapses every bottom offset to the safe-area inset (there's no bottom tab bar to clear at regular width). Defaults to `false`, so all existing call sites are byte-for-byte unchanged (asserted in a test).
- **`(tabs)/_layout.tsx` third branch** — at regular width renders `IpadSidebar` + the `Tabs` navigator with its bar hidden (`tabBar={() => null}`). The navigator still owns routing + per-tab state; the sidebar drives it through the global Expo Router `router` and highlights the focused tab via `useSegments()` (new `tabsActiveSegment` helper, so we never index the route-typed tuple directly). All five tabs map 1:1 — no rename, no dissolve.
- **`IpadSidebar.tsx`** — a `GlassSurface` rail: Home · Climbs · Record · Discover at the top, You pinned to the bottom (HIG convention), active row tinted + pilled, `hapticSelection` on tab change. Reuses the existing `Icon` / `Text` / `PressableSurface` primitives and the same i18n keys as the tab bars (3 new semantic glyphs added to `icon-map`).
- **`app.config.ts`** — `supportsTablet: true`; iPad rotates freely via `UISupportedInterfaceOrientations~ipad` while iPhone stays portrait-locked; `requireFullScreen` left `false` so Slide Over / Split View work (which is what makes the compact fall-through valuable).

## Validation

[...]

PR #2915, opened 2026-06-16
feat(mobile): iPad two-column shell — PlayDrawer pane (board presence)

marcodejongh:
**Stacked on #2904** (the adaptive shell). The iPad shell becomes **two columns**: the tab content on the left, and a **persistent right column that is the PlayDrawer for the climb lit on the wall**. It replaces the floating accessory/queue bar.

> Base this on #2904; rebase onto `main` once #2904 merges.

## What it does

- **`PlayDrawer` gains `presentation: 'sheet' | 'pane'`.** The pane renders the *same* content tree (header · board · playback · action bar · sections) in a plain `ScrollView` — no bottom sheet, grabber, close button, or peek-height math. One state machine, two presentations.
- **The pane mirrors the WALL, not the user's queue.** Per review feedback, it sources its climb from `useWallOrQueueCurrentClimb(null)` — the board-presence "now on the wall" climb (the same source the accessory bar used), or **nothing** when the wall is empty ("Nothing on the wall" empty state). Prev/next is hidden because a wall climb isn't in the queue. Tapping a list climb sets it current → when a board is connected it's sent to the wall → the pane reflects it.
- **Sub-sheets unchanged (the original "linchpin", solved cheaply).** The 5 sub-sheets (actions, angle, log-ascent, BLE, beta) stay root-level `BottomSheetModal`s with `stackBehavior=push`, so they layer above the app whether triggered from the sheet or the pane — **no popover rewrite needed**.
- **Wiring.** `DrawerHostProvider` exposes the drawer props for the pane (`IpadPlayPane` consumes them) and renders the bottom sheet **only on compact**; `openPlayDrawer` makes the climb current without presenting a sheet on regular width. `(tabs)/_layout.tsx` adds the right column (responsive 320–400pt). The floating `PersistentQueueBar` is suppressed on the iPad shell.
- **App-wide** across all five tabs. **Compact width** (iPhone, narrow iPad split) keeps the bottom-sheet PlayDrawer + accessory bar verbatim.

## Validation

- `vp test run --project mobile` — **303 files / 2197 tests** pass (incl. updated drawer-host + tab-layout mocks and a pane-renders assertion).
- `vp run check:mobile-bundle` — OK. `vp check` on all changed files — clean.
- i18n catalog-completeness — green (added `playView.paneEmpty` to en/es/fr).
- **iPad Pro 13" simulator** — two columns on every tab; the right column shows the board-presence pane with the "Nothing on the wall" empty state (no live wall climb on the sim); no floating accessory bar; compact split falls back to the phone UI.


marcodejongh:
Addressed the automated review across three follow-up commits. Triage of the latest round:

**Fixed**
- Pane now follows selection (master-detail); `setAsCurrent:false` opens (feed/beta/climb-view) preview in the pane without committing to the queue.
- Wall-column header now owns `insets.top` in `column` mode (was under the notch).
- `NowOnTheWallPanel` `memo`-wrapped; dead `boardPresenceBoardIdRef` removed.
- Added tests: size-class wall budget, tab-layout column gating, drawer-host regular-width open (commit vs preview), `IpadPlayPane` portrait-strip branch.
- Documented the iPad adaptive layout in `docs/ai-design-guidelines.md`.

**Intentionally not changed**
- *Column footer `insets.bottom`*: correct as-is. The iPad shell has no bottom tab bar, so the full-height wall column reaches the home indicator — the bottom inset is needed (zeroing it would tuck the footer under the indicator).
- *`boardSectionPane` vs `boardSection`*: mirror each other on purpose (documents the pane's no-fixed-height intent); pre-existing, harmless.

**Tracked follow-up**
- Unit tests for `NowOnTheWallPanel`'s action pipeline (LRU cache / generation guards). This is logic moved verbatim from `BoardSheet` (behavior unchanged, all existing tests green), so it's better as its own focused test PR than bundled here.

marcodejongh:
Triage of the latest automated review:

**Fixed**
- Docs: `resolveWallSurface` signature corrected to the options-object form.

**Pre-existing, moved verbatim from `BoardSheet` (not introduced here)**
- The `isActionLoading` / `pendingActionKeys` identity churn in `renderHistoryItem` / `listHeader`, and the `visibleHistory.length` memo dep, all came over unchanged in the `NowOnTheWallPanel` extraction. The history list is short and capped (not a hot path), and `listHeader` only reads `length > 0` (presence, not identity), so switching that dep to the full array would *add* re-renders. Tightening the row-loading selector is a reasonable follow-up alongside the action-pipeline tests, but it's behavior that already shipped in `BoardSheet`.

**False positives for this layout**
- *Cross-board climb tap on iPad "silently fails"*: it doesn't. On regular width `openPlayDrawer` sets the board override **and** commits/previews the climb in the same branch — it never relies on the deferred `pendingOverrideOpenRef` path (that's the phone-sheet mechanism). Clearing the ref there is correct.
- *Column footer `insets.bottom`*: correct — the iPad shell has no bottom tab bar, so the full-height column reaches the home indicator.

**Tracked follow-up (own PR)**
- `NowOnTheWallPanel` action-pipeline unit tests + the row-loading memo tightening, and `WallStrip` / `SidebarWallCell` / `variant='column'` test files.

CI is green across all rounds; the substantive, newly-introduced issues are resolved.

PR #2956, opened 2026-06-17
Fix Kilter BLE incompatible_climb: explicit preview, spill-skip, softer mismatch UX + diagnostics

marcodejongh:
## Why

PostHog + DB telemetry for a Beest Kilter (serial `751521`) showed a clean BLE connect — the controller resolved correctly to the saved board (kilter / layout 1 / size 10 / sets 1,20) — immediately followed by `Climb Sent to Board Failure` with `failureReason = incompatible_climb`. The controller→board mapping was fine; the **wrong climb** reached the board, every LED placement was skipped, and the wall dark-fired. Two mobile root causes, plus the telemetry was too thin to tell them apart.

## What (5 reviewed parts, mobile only + one additive `ClimbInput` field)

**A — Preview is now explicit.** The play drawer could show a view-only "preview" climb while the BLE auto-sender lit `currentClimbQueueItem` — a different climb. Preview stays for the surfaces that need it (workout builder, logbook/feed/cross-board browse, the peer-driven accessory wall climb) but is now shown with a **"Preview" badge + a "Set active" button**, and the overloaded `setAsCurrent:false` open option is split into `committedExternally` (render-ahead) vs `previewQueueItem` (view-only). The lightbulb/auto-sender are unchanged.

**B — Spill-skip.** Queue items now carry `boardType`/`layoutId` (preserved in `climbToQueueItem`/`toClimbInput` and added to the GraphQL `ClimbInput` so they round-trip through party sync). When the connected board's current climb belongs to a *different* board/layout, the auto-sender skips it instead of dark-firing — advancing to the next compatible climb (or clearing the wall) and toasting. Only a **known** mismatch is skipped; unknown/missing metadata sends as before.

**C — Mismatch dialog UX.** "Connect anyway" is a warning, not a destructive (red) action. New mobile-only copy ("our records show this controller belongs to another board setup; connecting anyway may light climbs incorrectly", "Switch board setup" / "Use recorded setup"); web's shared dialog is untouched.

**D — Send-failure copy.** Send failures stop using "Bluetooth is not available" as the title — a send-specific "Couldn't light board" replaces it.

**E — Diagnostics.** Thread send context (`sendSource`, `targetQueueItemUuid`, `climbUuid`, the climb's own board config) into `Climb Sent to Board Success/Failure`; incompatible failures now carry `skippedPositionCount`/`skippedRoleCount`/`totalPlacements`. `Bluetooth Connection Success` + every send carry `connectedViaMismatchOverride`. New events: `BLE Queue Climb Skipped`, `BLE Board Config Mismatch Shown/Resolved`.

## Testing

- `vp run typecheck` (mobile, shared, backend, web, db) — pass.
- Full mobile suite: **2330 tests** pass (incl. new board-config-compat, spill-skip, preview-banner, mismatch-dialog, send-context tests).
- Web climb/queue (1005) + backend queue/climb (79) pass — validates the `ClimbInput` round-trip.
- i18n catalog parity + orphan checks pass; `vp run check:mobile-bundle` OK.
- Each part was reviewed by an independent agent and its feedback addressed before the next.

## Notes

- Party mode: advancing past a spill changes the shared current climb for peers — acceptable since a session is normally one physical board.
- Manual QA still recommended on a dev client (preview badge + Set active; spill-skip toast; mismatch dialog wording).

### DIFF
--- a/packages/mobile/src/lib/climb-to-queue-item.ts
+++ b/packages/mobile/src/lib/climb-to-queue-item.ts
@@ -16,6 +16,12 @@ import type { ClimbQueueItem } from '@boardsesh/queue';
 export function toClimbInput(climb: Climb): ClimbInput {
   return {
     uuid: climb.uuid,
+    // Board the climb belongs to. Round-tripped through the queue so a connected
+    // board can skip a climb set for a different board/layout (a "spill" climb
+    // from a party peer on another board, or a queue left over from a board
+    // switch) instead of dark-firing the wall.
+    boardType: climb.boardType,
+    layoutId: climb.layoutId,
     setter_username: climb.setter_username,
     userId: climb.userId,
     name: climb.name,
@@ -53,6 +59,10 @@ export function climbToQueueItem(climb: Climb, options?: { suggested?: boolean; 
     suggested: options?.suggested,
     climb: {
       uuid: climb.uuid,
+      // Board metadata so the BLE auto-sender can detect a board/layout mismatch
+      // before writing (see toClimbInput above).
+      layoutId: climb.layoutId,
+      boardType: climb.boardType,
       name: climb.name,
       frames: climb.frames,
       setter_username: climb.setter_username,
--- a/packages/backend/src/__tests__/climb-queries.test.ts
+++ b/packages/backend/src/__tests__/climb-queries.test.ts
@@ -35,6 +35,14 @@ describe('Climb Query Functions', () => {
       expect(typeof result.hasMore).toBe('boolean');
       expect(result.totalCount).toBeDefined();
       expect(typeof result.totalCount).toBe('number');
+
+      // Each result carries the searched board + layout so a queued climb can be
+      // checked against a connected board (BLE spill guard). The search is scoped
+      // to testParams (kilter / layout 1) by its WHERE clause.
+      if (result.climbs.length > 0) {
+        expect(result.climbs[0].boardType).toBe('kilter');
+        expect(result.climbs[0].layoutId).toBe(1);
+      }
     });
 
     it('should enforce MAX_PAGE_SIZE limit', async () => {
--- a/packages/backend/src/__tests__/queue-sync-fixes.test.ts
+++ b/packages/backend/src/__tests__/queue-sync-fixes.test.ts
@@ -593,4 +593,32 @@ describe('setCurrentClimb - combined queue/current state publishing', () => {
       }),
     );
   });
+
+  it('round-trips boardType/layoutId through setCurrentClimb storage (party spill metadata)', async () => {
+    const sessionId = uuidv4();
+    const boardPath = '/tension/8/2/3/40';
+    await registerAndJoinSession('client-1', sessionId, boardPath, 'User1');
+
+    // A peer on tension / layout 8 sends a climb carrying its own board metadata.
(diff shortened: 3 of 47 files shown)


marcodejongh (commented):
I found one medium severity issue during review.

packages/mobile/src/lib/climb-to-queue-item.ts
```diff
@@ -16,6 +16,12 @@ import type { ClimbQueueItem } from '@boardsesh/queue';
 export function toClimbInput(climb: Climb): ClimbInput {
   return {
     uuid: climb.uuid,
+    // Board the climb belongs to. Round-tripped through the queue so a connected
+    // board can skip a climb set for a different board/layout (a "spill" climb
+    // from a party peer on another board, or a queue left over from a board
+    // switch) instead of dark-firing the wall.
+    boardType: climb.boardType,
```
marcodejongh, line 23: Medium: this adds the board metadata when a local queue item is created, but it still does not survive the server-driven queue paths. The mobile subscription selection (`SUBSCRIPTION_CLIMB_FIELDS`) and `SubscriptionClimb`/`toClimbQueueItem`, plus the shared `QUEUE_ITEM_FIELDS` fragment, still omit both `boardType` and `layoutId`. Any `FullSync`, `QueueItemAdded`, or `CurrentClimbChanged` payload from a peer will come back without metadata, and `classifyClimbBoardCompatibility` treats missing metadata as unknown/sendable. That means mixed-board party spill climbs can still reach the BLE write path on peer devices, so the incompatible-climb/dark-fire case is not fully fixed. Please include these fields in the queue fragments/conversion and add a subscription round-trip test.

marcodejongh:
Addressed the board-metadata round-trip feedback (commit 55f8b66).

You were both right — the metadata only lived on locally-built queue items, so a party peer on another board still sent an item that arrived without `boardType`/`layoutId` (classified `unknown` → sent anyway). Made it flow end to end:

- **Backend source**: `searchClimbs` (`mapResultToClimbRow`) and `getClimbByUuid` now stamp `boardType`/`layoutId` from the board-scoped route params — same pattern as the existing `angle` stamp, and safe because the search/detail WHERE clauses pin board + layout. `ClimbRow` carries them; the Zod `ClimbInputSchema` accepts them (defense-in-depth — the resolver already stores the raw item).
- **Mobile**: added them to `CLIMB_SEARCH_FIELDS`, `CLIMB_DETAIL_FIELDS`, `SUBSCRIPTION_CLIMB_FIELDS`, and to `SubscriptionClimb` + `toClimbQueueItem`.
- **Shared**: web `CLIMB_FIELDS` fragment requests them too.

**Tests**: mobile `toClimbQueueItem` carry-through + graceful-omit; backend `searchClimbs` stamping (mocked + real-DB integration); and a `setCurrentClimb` → Redis → `getQueueState` **server round-trip** so the FullSync/subscription path is covered.

Known follow-up (graceful degradation, not a regression): playlists, favorites, tick-history, create-climb, and web-originated sends don't stamp the metadata yet, so a spill from those sources still sends as today (the guard never *wrongly* blocks — missing metadata is treated as sendable). Happy to do those in a follow-up if you'd like them covered now.

Stevie-Ray:
@marcodejongh 🔥🔥🔥

PR #2967, opened 2026-06-17
Fix unreadable x-axis on Profile Activity & V-Points charts

marcodejongh:
## Problem

On the Profile / You page (`/you`), two charts had broken x-axes:

- **Activity**: x-axis labels were clipped — only "half a W" was visible.
- **V-Points**: the x-axis showed bare week numbers with no indication it was a time axis.

## Root causes

1. **Activity (`CssBarChart`)** — the `<BarChart>` was given a fixed pixel `height`, but its wrapper `<Box>` resizes via a CSS media query (smaller on mobile). MUI X renders the SVG at the fixed height, so on mobile the chart overflowed its box and the bottom label band was clipped. The labels were also angled at -45° at a 9px font; MUI X collapses overlapping rotated labels to **empty strings**, leaving only a clipped fragment.
2. **V-Points** — week-number ticks at 10px with **no axis title**.

## Changes

- `CssBarChart` / `GroupedBarChart` now **fill their responsive container** (no fixed height) so they match the box on mobile, and use horizontal **11px** tick labels.
- Activity chart switched to **horizontal, downsampled** week labels (the approach V-Points already uses) — they degrade gracefully (MUI hides overlaps) instead of collapsing to empty.
- V-Points chart gains a **"Week"** x-axis title (i18n: `en-US` / `es` / `fr`) plus larger ticks and a taller bottom margin.
- Removed the now-dead `angledLabels` prop.

Week-number labels were kept (per request); they now read like `W25 '25 … W14 '26` and are fully legible.

## Verification

- `vp run typecheck:web` ✓, lint ✓, `vp run check:i18n` ✓
- Full web suite (**4736 tests**) ✓ including i18n catalog-completeness; `css-bar-chart` unit tests ✓
- Drove the real app signed in as the seeded test user and captured `/you` at desktop and mobile widths — labels render fully on both, downsampling to ~5 on mobile, no clipping.

### DIFF
--- a/packages/shared/i18n/locales/en-US/profile.json
+++ b/packages/shared/i18n/locales/en-US/profile.json
@@ -60,6 +60,7 @@
     "flashVsRedpoint": "Flash vs Redpoint",
     "vPoints": "V-Points",
     "vPointsTotal": "{{value}} total",
+    "weekAxisLabel": "Week",
     "chartTooltipTotal": "Total",
     "gradeDistributionAria": "Grade distribution across boards",
     "weeklyAttemptsAria": "Weekly attempts by difficulty",
--- a/packages/shared/i18n/locales/es/profile.json
+++ b/packages/shared/i18n/locales/es/profile.json
@@ -60,6 +60,7 @@
     "flashVsRedpoint": "Flash vs Trabajado",
     "vPoints": "V-Points",
     "vPointsTotal": "{{value}} en total",
+    "weekAxisLabel": "Semana",
     "chartTooltipTotal": "Total",
     "gradeDistributionAria": "Distribución por grados en todas las tablas de escalada",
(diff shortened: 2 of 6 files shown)


marcodejongh:
Re: claude-review — `height`/`mobileHeight` aren't dead. They're still consumed by the wrapper `<Box>` (`sx={{ height, '@media (max-width:768px)': { height: mobileHeight } }}`), which sets the responsive container the chart now fills. So `height={200} mobileHeight={150}` in `stats-summary.tsx` do take effect — they size the box, and dropping the fixed `<BarChart height>` is what lets the chart match the box on mobile (the bug fix).

Good call that the contract was non-obvious, though — pushed d32089367 adding JSDoc to both prop types spelling out that they size the container and the chart fills it.

PR #2991, opened 2026-06-17
feat(mobile): dismissable active-climb accessory bar

marcodejongh:
## Why

The persistent "now playing" strip (current climb + grade + tick, above the tab bar) showed on **every** tab whenever a climb was current — including the social Home/Discover/Profile tabs. The visibility rule conflated *"a climb is queued"* with *"the user is climbing right now,"* so casually previewing a climb pinned the strip indefinitely (the solo queue even persists across launches), with no way to tuck it away. Users complained it's always there even when they're not at the gym.

## What

The strip now behaves like a music mini-player: present and useful while you climb, dismissable when you're not — without eroding its strengths.

- **Hide-only, never destructive.** Dismissing writes a local, persisted `{ dismissedQueueItemUuid }` and dispatches **no** queue action. The queue and current climb are untouched and still reachable from Record / the queue sheet. Party-mode safe (local store, no peer desync).
- **Climbing-signal gated (root-cause fix).** `useIsActivelyClimbing = bluetoothConnected || liveWallClimb || inSession` (one shared hook). While any holds, the bar asserts itself, is non-dismissible, and the hide affordance does not render — an active climber never loses one-tap return or quick-tick.
- **Keyed to the queue item**, not the climb, so re-adding the same climb is a fresh activation that un-hides for free. Persists across launches; clears when the user **connects to** or **switches** boards.
- **Affordances.** JS bar: swipe down (plus a VoiceOver "Hide" action since a swipe isn't screen-reader-reachable). iOS 26 Liquid Glass platter: an explicit hide chevron, visually separated from the tick (fat-finger guard). The native UIKit host unmounts cleanly when hidden (no doubled-snapshot regression). A one-time coachmark teaches it.

Design was worked through a UX panel; product decisions (native affordance, BLE = climbing, persist-until-board-change, coachmark) confirmed with the requester.

## Files

New: `accessory-dismiss-store.ts`, `accessory-dismiss-hint-store.ts`, `use-actively-climbing.ts`, `use-accessory-dismissed.ts`, `use-accessory-dismiss-resets.ts`, `AccessoryDismissResets.tsx`, `AccessoryDismissHint.tsx`.
Wired through: `use-accessory-climb-tap.ts`, `ClimbCapsule.tsx`, `NativeAccessoryClimbRow.tsx`, `persistent-queue-bar.tsx`, `use-bottom-chrome-metrics.ts`, `app/(tabs)/_layout.tsx`, `app/_layout.tsx`, `queue-provider.tsx` (new `useActiveClimbQueueItemUuid` selector). i18n added to en/es/fr.

## Testing

- `vp run typecheck:mobile` ✓
- `vp run test:mobile` ✓ (2385 tests; new unit tests for the store + `useIsActivelyClimbing` + `useAccessoryDismissed`, plus dismiss-coverage tests in the tab-layout and persistent-bar suites)
- `vp run check:mobile-bundle` ✓ (Android 10.9 MB)
- i18n catalog-completeness ✓
- Adversarial multi-agent review pass applied (queue-item-uuid keying fix + a11y improvements).

⚠️ **Needs on-device (macOS) QA:** the iOS 26 native platter chevron + clean-unmount-on-hide (no doubled-text snapshot) can't be exercised on the Linux bundle check. QA plan in `.boardsesh/qa-notes.md`.

### DIFF
--- a/packages/mobile/app/(tabs)/__tests__/tab-layout.test.tsx
+++ b/packages/mobile/app/(tabs)/__tests__/tab-layout.test.tsx
@@ -19,6 +19,7 @@ const cfg = vi.hoisted(() => ({
   glassCapable: true,
   platformOS: 'ios' as 'ios' | 'android',
   materialScreens: [] as Array<{ name: string; options?: { lazy?: boolean } }>,
+  accessoryDismissed: false,
 }));
 
 vi.mock('react-native', () => ({
@@ -54,6 +55,11 @@ vi.mock('../../../src/hooks/use-sticky-accessory-presence', () => ({
 // real hook; the route-segment split is unit-tested in route-segments.test.
 vi.mock('../../../src/hooks/use-inside-tabs', () => ({
   useInsideTabs: () => cfg.insideTabs,
+}));
+
+// User-dismissed (and not actively climbing): the native accessory host unmounts.
+vi.mock('../../../src/hooks/use-accessory-dismissed', () => ({
+  useAccessoryDismissed: () => cfg.accessoryDismissed,
 }));
 
 vi.mock('../../../src/components/queue-control/QueueBottomAccessory', () => ({
@@ -164,6 +170,7 @@ describe('TabLayout', () => {
     cfg.glassCapable = true;
     cfg.platformOS = 'ios';
     cfg.materialScreens = [];
+    cfg.accessoryDismissed = false;
   });
 
   it('lands on the home tab by default', () => {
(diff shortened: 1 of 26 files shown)


marcodejongh:
Closing for now — moving to a design look-over first via #3042. The branch `feat/mobile-accessory-bar-dismiss` stays as-is (implementation complete + validated); we'll reopen / re-cut a PR once design has signed off.

PR #3003, opened 2026-06-17
feat(mobile): act on the iPad design review — pane budget, single navigator, sidebar + a11y polish

marcodejongh:
Stacked on **#2904** (base: `feat/ipad-adaptive-shell`) — this is the design-review pass over the iPad adaptive shell. An expert design panel reviewed the branch across five lenses (iPadOS HIG, visual-system fidelity, interaction/navigation, accessibility, adaptive robustness), each finding adversarially verified against the actual code. **No P0 blockers** — the shell is a clean foundation. These are the grounded, confirmed fixes.

## High priority

- **Width-budget the detail pane.** New `resolveDetailPaneSurface` mirrors `resolveWallSurface`: on the tightest regular portraits (iPad mini and 9.7–10.2″, 744–810pt) the persistent pane would squeeze the browse list below the code's own 400pt readable floor, so there it falls back to the compact bottom-sheet `PlayDrawer` and the list keeps the full content column. `_layout` (mounts the pane) and `drawer-host-provider` (mounts the sheet) gate on the **same** width budget, so there is always exactly one play-drawer host — never zero, never two.
- **One `Tabs` navigator across the boundary.** iPad now keeps a single keyed JS `<Tabs>` mounted across `regular ↔ compact`; only the chrome swaps (glass rail at regular width, Material bar in a narrow split). A resize across 700pt (Split View drag / Stage Manager) no longer remounts the navigator and drops per-tab scroll + nested-stack depth. **The iPhone/Android paths are byte-for-byte unchanged** — `NativeTabs` is reserved for the glass iPhone, and `useNativeTabBar()` now returns false for any iPad so the bottom-chrome metrics and the accessory mount stay consistent with the bar actually on screen.
- size-class tests for the 744/768/810 band and the pane-surface flip.

## Polish & accessibility

- **Sidebar:** active row uses neutral system glyphs (matching both iPhone tab bars), with the `systemFill` pill as the sole active affordance; pointer-hover + keyboard-focus states; re-tap always haptics and pops the tab to root; primary destinations grouped as a `tablist`; 2-line labels for large Dynamic Type; pill/icon/dot dimensions moved onto the token ladder.
- **Empty pane:** shared `PanePlaceholder` (theme label roles + adaptive icon, not opacity) replaces two duplicated states.
- **Board-presence colour:** dedicated `live` brand role replaces the `warning`-as-live token drift on the on-the-wall dot/chip/label; `PaneOnWallChip` grouped for VoiceOver and suppressed when a `WallStrip` is docked above the pane; `StatTile`s read as one VoiceOver stop.
- iPad adaptive-layout decisions documented in `docs/ai-design-guidelines.md`.

## Deferred (commented in code, flagged for the on-device pass)

- **Cmd-1..5 shortcuts** — no Expo-compatible key-command path without a new native module.
- **Per-pane VoiceOver landmarks** — RN has no region role; wants an on-device call on the right header treatment.
- **Sheet → pane preview migration** — documented drop for a narrow resize-mid-inspect edge case (lifting `PlayDrawer`'s internal preview state up was too invasive for the payoff).

## Validation

- `vp run typecheck:mobile` ✅
- full mobile suite — **2417 tests pass (330 files)** ✅
- `vp run check:mobile-bundle` — 10.9 MB ✅
- An adversarial multi-agent QA pass caught and fixed a real regression before commit: the single-navigator change meant iPad never renders `NativeTabs`, but `useNativeTabBar()` still keyed only on regular width — so a glass-capable iPad in Split View would have dropped the "now playing" bar (native accessory never mounts **and** the JS `PersistentQueueBar` suppressed). Fixed + covered by an iPad-narrow-split test.

## Still needs eyes (the headless review couldn't see pixels)

An on-device iPad pass: iPad-mini/9.7″ portrait (pane suppression), a 700pt resize (per-tab state preserved), Magic Keyboard hover + VoiceOver order, largest Dynamic Type in fr/es, dark-mode contrast of the neutral active glyph.

> Separately, an App Store **communications** review of the store metadata + submission docs is queued as its own follow-up PR off `main` (stale Capacitor instructions, the `supportsTablet: true` ↔ iPad-screenshot obligation this PR creates, App-Privacy/category fixes, conversion-first copy). Kept out of this PR.

https://claude.ai/code/session_01RzMwYud1V13wRuaJveTjVz


marcodejongh:
Closing this because the iPad design-review stack has been folded into #2904 on `feat/ipad-adaptive-shell` after the rebase. I verified the equivalent commits with `git range-diff`; #2904 also has the later iPad screenshot, column-balance, and review-followup commits on top.

PR #3014, opened 2026-06-18
feat(mobile): board driver as a pressable avatar with a Bluetooth badge

marcodejongh:
## What

Replaces the textual **"Lit by {name}"** board-presence indicator in the mobile app with the climber's **avatar + an amber Bluetooth glyph in the top-right corner**, pressable to open their profile. Four fragmented treatments (play-drawer banner, lightbulb pip, BoardSheet hero, history rows) now share one `BoardDriverAvatar`.

|  | Before | After |
|---|---|---|
| Play-drawer banner | `🔵 Lit by Marco` (icon + text) | Marco's avatar w/ Bluetooth badge, taps to profile |
| BoardSheet hero / history | "Lit by Marco" caption | avatar + name |
| Lightbulb pip | holder avatar (always) | suppressed while the banner shows the face |

## How

- **`BoardDriverAvatar`** (new) — the shared atom: `PressableAvatar` + a corner badge with `connected` (amber Bluetooth glyph) / `idle` (grey "?") / `none` states. Neutral disc backing keeps the glyph legible in light **and** dark (white-on-amber fails the dark `warning` tint).
- **`useBoardDriver`** (new) — resolves the driver from board presence (live climb → holder fallback) with the once-a-minute idle check; replaces `BoardConnectionBadge`.
- **Banner owns the face**: `PlayDrawer` passes `showHolderBadge={!(isPreview && drawerPreviewIsWallClimb)}` so the same avatar never appears twice in the drawer.
- **Gesture-safe**: the avatar is pressable in the banner and non-interactive rows, but **identity-only inside interactive `ClimbListRow`s** — that row's own RNGH tap gesture owns the touch, so a nested pressable would double-fire (climb-open + profile push) and is a11y-collapsed anyway.
- **Backend**: server-derived `sentByUserId` added to `BoardPresenceClimb` (SDL + resolver) for the profile link, and the durable `boardHistory` query now joins `users`/`user_profiles` so history rows carry attribution (they returned `null` before — a latent bug). No DB migration.
- **a11y**: the dropped visible name survives in the avatar's accessibility label, including an anonymous-holder variant. New `drivenByA11y` / `drivenByAnonA11y` keys in en-US/es/fr; removed the now-unused mobile `litByLine` + `playView.onWallBadge`.

## Decisions
Amber badge, banner owns the face, full scope (banner + hero + history) — confirmed with the requester.

## Testing
- `vp run typecheck` (web, backend, mobile, shared, db) ✅
- Mobile: **2403 tests pass** (new `BoardDriverAvatar` + rewritten banner/action-bar tests); backend board-presence: **57 pass** (new `sentByUserId` + history-join assertions); i18n catalog parity + orphan check ✅
- `vp check`, mobile variant guard, Metro bundle ✅
- Reviewed by an adversarial QA agent; the interactive-row gesture conflict it surfaced is addressed above.

## Follow-up (not in scope)
Tap-avatar→profile **and** tap-row→climb in the live interactive feed (Twitter-timeline pattern) would need gesture disambiguation threaded through the shared `ClimbListRow` — left out as too invasive for this change.

https://claude.ai/code/session_01BEQYwG2srqR6QCf8eJHwC5


marcodejongh:
Follow-up commit `b537a2f` from on-device testing of the play-drawer banner:

- **Fix: on-wall driver showed as anonymous "?" when there was no active holder.** `useBoardDriver` required a connection holder, but a climb stays lit after its sender stops being the active BLE writer — so the banner now resolves identity from `currentClimb` (or holder), null only when both are absent. The BoardSheet hero already read `currentClimb` directly, which is why it worked and the banner didn't. The lightbulb pip keeps its "who's connected" meaning by gating on the new `isHeld`.
- **Bigger banner avatar** (24 → 32).
- **Recency in the corner badge.** The top-right corner swaps in place: amber Bluetooth glyph while the driver is actively connected → compact "how long ago they lit it" (`5m`, `2h`) once they've dropped or gone quiet, replacing the old idle "?". New `compactAgoParts` helper (unit-tested) + localized `litAgo*` keys in en-US/es/fr.

Tests: added a no-holder regression case, a corner-label case, and `compactAgoParts` coverage. Typecheck + mobile tests + i18n parity/orphans + lint all green.

marcodejongh:
Ran two code-review passes (correctness/bugs + simplicity/consistency). Fixes in `3227dca`:

**Blocker**
- The shared `BOARD_PRESENCE_CLIMB_FIELDS` selection set didn't request `sentByUserId`, so even though the resolver/SDL/types carried it, `currentClimb.sentByUserId` was `undefined` at runtime and every hero/history/banner avatar silently fell back to non-interactive — the pressable-to-profile behaviour didn't actually work in production. Added the field (one place, fixes web + mobile).

**Should-fix**
- `boardConnection` now only adopts the newest climb's name/avatar/time when that climb was sent by the current holder; otherwise it returns the holder's userId with null attribution. Prevents a late joiner from pairing one climber's name over another's avatar and tapping through to the wrong profile.
- Added a dedicated `use-board-driver` unit test (currentClimb-vs-holder precedence, `isHeld`, idle threshold, `lastSentAtMs` null).

**Nits**
- `compactAgoParts` guards non-finite input; banner null-guards `lastSentAtMs` and passes an explicit `status` instead of relying on `cornerLabel` overriding a contradictory one; idle "?" uses the adaptive `systemColors.secondaryLabel`.
- Added a corner-label case for `BoardDriverAvatar` and a holder display-identity assertion for `boardConnection`.

Validation: full typecheck, full mobile suite (2415), backend board-presence (57/57), i18n parity/orphans, lint — all green. (One backend concurrency test, `resolveBoardForSerial > converges concurrent…`, flaked locally under heavy machine load — it's unrelated to this PR, reproduces on the base commit with these changes stashed, and passes once load drops.)

PR #3038, opened 2026-06-18
fix(mobile): centre the checkmark glyph on its ink, not its box

marcodejongh:
## Summary

The `checkmark` SF Symbol draws entirely above the baseline but still reserves descender space below it. `.scaleAspectFit` centres that bounding box, so the ink lands about 9% of the point size high — most visible on the iOS 26 bottom accessory, where the log-ascent tick floats above the `V11` it sits next to.

`icon-map` owns the glyph (`check.small` → `checkmark`), so the correction goes there: a per-glyph `iosOpticalCenterRatio` that `Icon` turns into a `translateY` on the SF Symbol. The transform is visual only — the symbol still occupies `size` in layout — and it's iOS-only, since the MaterialCommunityIcons glyph is already ink-centred.

Putting it in `icon-map` rather than at the call site is the point. Seventeen render sites across fifteen files draw `check.small` (playlist done buttons, filter-sheet selections, settings rows, the Strava/Health share buttons, the session feed, the bottom-accessory tick). They are all off by the same proportion, because it's a property of the glyph and not of any one screen. A nudge applied in `LogAscentToolbarButton` fixes one of them and silently drifts from the other sixteen the next time someone touches it.

### Scope change

This PR opened in June as a two-part layout fix. Main has since solved the second part — and solved it better:

- `NATIVE_BOTTOM_ACCESSORY_MAX_WIDTH` went 344 → **420** in 87f303c55, deliberately kept as a bound on very wide displays. This PR deleted it.
- `ACCESSORY_TRAILING_INSET` went `spacing[3]` → **`spacing[8]`** (12 → 32) in 3181b92c1, whose commit message names this exact symptom. This PR set it to `spacing[4]` (16).

Rebased on `main` with all four conflicts resolved to main verbatim, and the `QueueBottomAccessory` width change reverted. What's left is the glyph fix and its tests.

## Release Notes

- The tick next to the grade on the now-playing bar sits level with the text instead of floating above it. Same for the checkmarks in playlists, filters, and settings.

## Test plan

- `vp run typecheck:mobile` ✅
- `vp test run --project mobile` ✅ (4508 passed, 544 files)
- `vp check` ✅ (0 errors)

New coverage in `Icon.test.tsx` asserts the shift is applied on iOS, scales with `size`, is absent for glyphs with no recorded ratio, and never reaches the Android renderer. `icon-map.test.ts` bounds any future ratio to a nudge. Mutation-checked: dropping `style={opticalCenter}` from `Icon` fails two of the new tests, so they are not tautologies.

### Outstanding gate: visual confirmation on device

**Not verified.** The screenshot tooling is macOS-only and this was done on Linux, so it skipped — I have not seen this render.

The before/after captures in the comments above are from June and no longer stand as evidence for this diff: they predate the platter widening to 420 and the trailing inset going to `spacing[8]`, so the layout the `0.09` was eyeballed against has changed. The ratio itself carries over from that single capture (iPhone 16 Pro Max, `iconSize` 24) and has never been checked at another size.

Cheapest way to close this: it's on the OTA channel **`pr-3038`**, so anyone with a TestFlight build can load it from What's New → Try a preview and look at the now-playing bar. No macOS build needed. Otherwise, dispatch **Mobile Screenshots (Native)** on this branch.

Either way, three things to check:

1. The bottom accessory — the tick's ink lines up with `V11` / the climb name. Retune `iosOpticalCenterRatio` if it over- or undershoots.
2. The Material docked capsule (`--variant material`) — renders the same tick, and no run has ever covered it.
3. A smaller call site (18pt, e.g. a playlist filter row) — the ratio assumes the offset is linear in point size, which one 24pt capture can't establish.

### DIFF
--- a/packages/mobile/src/components/Icon.tsx
+++ b/packages/mobile/src/components/Icon.tsx
@@ -1,7 +1,7 @@
 import MaterialCommunityIcons from '@expo/vector-icons/MaterialCommunityIcons';
 import { SymbolView } from 'expo-symbols';
 import { Platform } from 'react-native';
-import { iconMap, type IconName } from './icon-map';
+import { iconMap, type IconMapping, type IconName } from './icon-map';
 
 type IconProps = {
   name: IconName;
@@ -13,10 +13,18 @@ type IconProps = {
 // Both glyph names live in icon-map.ts keyed by the same semantic IconName, so call
 // sites stay platform-agnostic.
 export function Icon({ name, size = 24, color }: IconProps) {
-  const mapping = iconMap[name];
+  const mapping: IconMapping = iconMap[name];
 
   if (Platform.OS === 'ios') {
-    return <SymbolView name={mapping.ios} size={size} tintColor={color} />;
+    // Some SF Symbols centre their bounding box, not their ink; icon-map records
+    // the per-glyph correction so every call site gets the same nudge. The
+    // transform is visual only, so the symbol still occupies `size` in layout.
+    const { iosOpticalCenterRatio } = mapping;
+    const opticalCenter = iosOpticalCenterRatio
+      ? { transform: [{ translateY: size * iosOpticalCenterRatio }] }
+      : undefined;
+
+    return <SymbolView name={mapping.ios} size={size} tintColor={color} style={opticalCenter} />;
   }
 
   return (
--- a/packages/mobile/src/components/__tests__/Icon.test.tsx
+++ b/packages/mobile/src/components/__tests__/Icon.test.tsx
@@ -34,8 +34,23 @@ vi.mock('expo-symbols', () => ({
 }));
 
 vi.mock('@expo/vector-icons/MaterialCommunityIcons', () => ({
-  default: ({ name, color }: { name: string; color?: string }) =>
-    createElement('div', { 'data-testid': 'mci', 'data-name': name, 'data-color': color }),
+  default: ({
+    name,
+    color,
+    style,
+  }: {
+    name: string;
+    color?: string;
+    // Surfaced so the "Android is never shifted" assertion is a real check: if
+    // the optical correction ever escaped the iOS branch, it would land here.
+    style?: { transform?: { translateY: number }[] };
+  }) =>
+    createElement('div', {
+      'data-testid': 'mci',
+      'data-name': name,
+      'data-color': color,
+      'data-translate-y': style?.transform?.[0]?.translateY,
+    }),
 }));
 
 import { Icon } from '../Icon';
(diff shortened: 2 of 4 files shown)


marcodejongh:
## Screenshot verification (iPhone 16 Pro Max, en-US)

Captured from the screenshot workflow on this branch ([run 27756638015](https://github.com/boardsesh/boardsesh/actions/runs/27756638015)). Both fixes confirmed in the native iOS 26 Liquid Glass bottom accessory.

### Tick centering + flush-right (before → after)

Top = before (`V11 ✓`): checkmark floats above the grade center line and the cluster stops short of the pill's right edge. Bottom = after (`V4+ ✓`): tick is centered on the grade and the cluster sits flush-right.

tick centering before/after

### Full accessory bar (before → after)

accessory bar before/after

<details><summary>Full home screenshot (after)</summary>

after home

</details>

> Note: the before frames are from a pre-fix `main` capture (climb \"willow\", V11); the after frames are this branch (\"loft & roses\", V4+) — different climbs, same accessory component, so the tick/grade alignment comparison is apples-to-apples.

`CHECKMARK_OPTICAL_CENTER_RATIO = 0.09` lands the tick centered, so leaving it as-is.

<sub>Evidence images live on the `assets/pr-3038-accessory-tick` branch (asset-only, not for merge).</sub>

marcodejongh:
## Follow-up: nudged the tick off the rounded edge

Per feedback that the tick sat too far into the platter's rounding, widened the trailing inset (`spacing[3]` → `spacing[4]`, ~4pt). The tick now reads centered over the tab bar's search glyph and clears the corner curve. Re-captured on [run 27758245130](https://github.com/boardsesh/boardsesh/actions/runs/27758245130).

Red line = search-glyph center. Top = before nudge (tick leaning into the corner), bottom = after.

tick nudge before/after

home bar after nudge

Note: the tick is pinned to a fixed right-anchored slot (44pt) beside a `flex:1` label, so its position is **independent of climb-name length** — the position shift seen earlier between the \"willow\" and \"loft & roses\" captures was the platter-width fix, not the name.

marcodejongh:
Both test points taken and pushed in 74ad09725.

- **Tautological Android assertion** — correct, the MCI stub never surfaced `style`, so that assertion passed regardless of the implementation. The stub now surfaces it, and I checked the guard by leaking a `translateY` into the Android branch: the test fails. Same check on the iOS side — dropping `style={opticalCenter}` from `Icon` fails the shift tests.
- **Redundant size coverage** — folded into one parameterised test over 18pt and 26pt, the smallest and largest sizes `check.small` actually renders at.

On the ratio: agreed, and the PR description says the same thing. It's an outstanding gate, not a claim. Worth noting it's now cheaper to close than a macOS screenshot run — this is live on OTA channel `pr-3038`, so anyone on a TestFlight build can load it from What's New → Try a preview and look at the now-playing bar. If it over- or undershoots, the retune is one number in `icon-map.ts`.

marcodejongh:
Docs point taken in cc3a4cdcd — the JSDoc now records the `(0, 0.2]` bound the icon-map test enforces.

Leaving the truthiness check as-is, though. `0` and `undefined` both mean "this glyph needs no correction", and that's exactly what the current code does. Switching to `!= null` would build `transform: [{ translateY: 0 }]` for a zero ratio — an identity transform that does nothing but allocate. The falsy branch isn't a footgun here, it's the right answer; it just wasn't written down. Now it is.

marcodejongh:
Correcting the record on the automated review above: **the ratio has not been verified.** The screenshots it cites are the June comments on this PR — that is precisely the stale evidence this PR's description rules out. They were captured before `NATIVE_BOTTOM_ACCESSORY_MAX_WIDTH` went 344 → 420 and `ACCESSORY_TRAILING_INSET` went `spacing[3]` → `spacing[8]`, so they show a layout that no longer exists, and they only ever covered one size (24pt) on one device.

Nobody has looked at this render. The visual gate in the description stands and should be closed before merge — via the `pr-3038` OTA channel or a **Mobile Screenshots (Native)** dispatch.

Please don't treat the ✅ as a substitute for that.

PR #3054, opened 2026-06-19
Match share-beta captions across the whole logbook + stop the home-tab remount

marcodejongh:
## What & why

Three fixes to the React Native **share-into-Boardsesh** flow, surfaced from the request + a PostHog dig on a test user.

### 1. Caption matching now spans the user's whole logbook (with board art)
Sharing a reel auto-matches its caption to a logged climb. It used to only match the **~20 most-recent loaded ascents**, so an older climb never got suggested.

New public query `userAscentCaptionMatches(userId, caption)` matches the caption against **all** the user's send/flash climb names server-side, then re-fetches full ascent rows — **with board art** — for just the matched climbs (cheap two-step: names-only match → full rows for the handful of matches, so we never ship the whole logbook's hold frames). The matcher moved from `packages/mobile` to `@boardsesh/shared-schema` (`matchClimbsToCaption`) so the backend and the share flow run identical logic. Suggestions are sends/flashes only (beta only attaches to ascents), which also stops suggesting un-attachable attempts.

### 2. Repeated shares no longer remount Home / stack modals
PostHog `$screen` timeline showed **every** share firing `/share-beta → / → /home → /share-beta` (~30× in 20 min), remounting the whole Home tab tree each time → the sluggishness.

- `+native-intent.ts` now returns `/` only on **cold** start; **warm** shares return `''` so Expo Router skips the re-navigation that re-fired the root home redirect.
- `ShareTargetProvider` `setParams` in place when the modal is already open instead of pushing a duplicate `/share-beta` (the per-reel `link` param differs, so `navigate` was stacking).

### 3. Attach errors are visible + no longer error-tracking noise
`attachBetaLink` validation rejections ("already linked", "not an ascent", …) were captured as PostHog `$exception` noise **and** shown via a toast that renders *behind* the native modal — so the user never saw them. Now they show **inline in the share sheet** and are dropped from error tracking via `isExpectedBetaValidationError` (genuine faults like `BETA_LINK_INTERNAL` still report).

> The `bulkVoteSummaries` empty-array error was already fixed on `main` (44333fa73); the captured occurrences are from the shipped 2.0.0 build and clear with the next build/OTA.

## Tests
- New: shared matcher unit tests; `userAscentCaptionMatches` + `climbUuids`-filter backend integration tests (whole-logbook match, attempt-exclusion, dedupe, alias resolution, blank caption); `native-intent` warm/cold; `share-target-provider` setParams dedupe; `isExpectedBetaValidationError` + `reportHandledError` drop/keep.
- Green: `vp run typecheck` (all packages), `vp run test:mobile` (2470), backend `tick-queries` (30), `vp run check:mobile-bundle`.

## Validation note
This needs a real device/simulator to confirm visually (whole-logbook suggestion with art, repeated-share smoothness, inline attach error) — not covered by Linux CI.


marcodejongh:
Addressed the review feedback in `1eabe8a3f` — and reworked matching to use the **quoted climb name** from the caption (Boardsesh's caption embeds it as `"Name" @ 40° on the …`).

- **Perf (unbounded scan):** gone. Instead of pulling every flash/send name per share, the resolver now extracts the quoted name and fetches matches via the existing **indexed, per-user `climbName` filter** (a handful of rows). No whole-logbook transfer.
- **Type (`Promise<unknown[]>`):** now returns `AscentFeedRow[]` (a declared ascent-row shape).
- **Flash-only silently dropped:** confirmed `statusMode: 'send'` already includes flashes (`inArray(status, ['flash','send'])`), so it was never dropped — added an integration test for a **flash-only** climb so it can't regress.

Matching is now **exact on the normalized quoted name**, which also removes the substring false-positives. Removed the now-unused internal `climbUuids` feed filter (zod + resolver + test).

Trade-off: a caption with no quoted name yields no auto-suggestion — that includes **MoonBoard** captions, which aren't quoted (Kilter/Tension are). The climber can still search. Happy to either also quote MoonBoard captions or add a bounded fallback if we want MoonBoard auto-match.

Green: `vp run typecheck` (all), `vp run test:mobile` (2470), backend `tick-queries` (31, incl. flash-only + no-quotes + quoted-caption).

marcodejongh:
Addressed the latest review in `19569f99f`:

- **Codex P2 (reset modal-local state on share-beta reuse):** added a `useEffect` keyed on `link` that clears `searchText` / `committedQuery` / `attachError` — so a warm share (provider `setParams`, no remount) no longer carries a stale search that suppresses the new reel's suggestions, or a leftover attach error for an unrelated link.
- **Claude (array `.length` in deps):** the `listData` memo now gates on `suggestedClimbUuids.size` and drops `suggestions.length` from its deps (`suggestedClimbUuids` already changes identity with `suggestions`).
- **Claude (unchecked cast):** typed `userAscentsFeed`'s return as `AscentFeedRow[]` (coercing the nullable `is_mirror` → `false` to match the `Boolean!` schema), so `userAscentCaptionMatches` no longer needs the `as AscentFeedRow[]` cast.

Green: typecheck (mobile + backend), `vp run test:mobile` (2470), backend `tick-queries` (31), lint clean.

marcodejongh:
Round 3 addressed in `63552eb8c`:

**`userAscentCaptionMatches`**
- **Rate limit:** added `applyRateLimit(ctx, 10, …)` (matching `userClimbPercentile`) so a crafted multi-quote caption can't drive unbounded ILIKE fan-out.
- **Caption length:** cap the caption at 2 KB before `extractQuotedClimbNames`, so a multi-MB payload can't make `matchAll` walk the whole string.
- **Parallel fan-out:** the per-quoted-name lookups now run via `Promise.all` instead of sequential `await`s.

**share-target-provider**
- Segment check is now `segments.includes('share-beta')` instead of `segments[0]`, so a future route-group nesting can't silently fall back to stacking.
- Added the missing test: post-login replay landing on an already-open modal → `setParams` reuse (no second push).

Green locally: typecheck (backend + mobile), backend `tick-queries` (31), `share-target-provider` (13), lint clean.

PR #3058, opened 2026-06-19
fix(mobile): make Android climb reaction menu read as a raised M3 surface

marcodejongh:
## Screenshots (Android · dark)

| Before (reported) | After |
| --- | --- |
| Before | After |

_Captured on an Android emulator. After: per-row dividers removed, background dimmed back so the board grid recedes, the card sits on an M3 surface-container-high tone with a level-3 elevation cast, Material icon colors kept, icons 24dp._

---

## What & why

The climb long-press **reaction menu** looked weird on Android (reported with a screenshot). On iOS it's a frosted, clearly-floating card; on Android two platform-specific rendering problems flattened it:

1. **No backdrop blur on Android** — iOS draws a `BlurView`; Android only had a thin scrim (`rgba(0,0,0,0.5)`/`0.35`), so the colorful board-art grid and floating buttons bled through and the menu didn't read as a focused overlay.
2. **The card's M3 elevation cast was clipped** — the card passed `overflow:'hidden'` into `GlassSurface`, whose Material branch puts that style on the *same view* it paints the elevation cast on, so the Android shadow was clipped to nothing and the dark tonal card melted into the dim.

A Material 3 design panel reviewed the design: this centered, modal, scrim-dimming, non-anchored overlay is an **M3 dialog**, so the card should be `surface-container-high` + a `level3` cast on a scrim heavy enough (no blur) to recede the background.

## Changes (all in `ClimbReactionMenu.tsx`)

- **Un-clip the elevation cast** — drop `overflow:'hidden'` from the card style so the Material `level` cast renders. Rows are transparent, so nothing pokes past the rounded corners.
- **Raise the card** to `role="high" level="level3"` (was `base`/`level2`). Material-only props — iOS glass/blur ignore them.
- **Denser scrim on the no-blur path**, gated on `useEffectiveSurfaceMode()` (the axis `GlassSurface` itself switches on, so it also covers Material-on-iOS): dark = `overlays.scrim` (0.6), light = `0.4`. The iOS blur path keeps today's lighter tint, so **iOS is unchanged**.
- **Drop the per-row dividers** (`showSeparator={false}`) — M3 action lists and iOS context menus are dividerless.
- **Leading icons 22 → 24** (M3). **Material semantic icon colors kept** (green add / red heart / violet edit·share), per the requested scope.

No new tokens; no `variant === …` branching (CI guard); only platform branch stays `Platform.OS === 'ios'` for the `BlurView` host.

## Scope notes
Kept to changes that fix Android without altering the approved iOS look. Deliberately deferred (would change iOS or expand scope): corner radius xl→lg, routing motion through `theme.motion`, removing the iOS scrim-over-blur, 48dp Material rows, a dark-mode hairline outline, and the favorite-red WCAG fix in the shared `resolveActionColors` token. The hairline outline is a ready follow-up if reviewers want a crisper dark-mode card edge.

## Validation
- `vp run typecheck:mobile`, `vp run test:mobile` (2470 pass), `vp run check:mobile-bundle` — all green.
- Verified on an Android emulator (dark): dividers gone, background recedes, card reads as a distinct M3 surface, Material icon colors intact. Light mode not captured (emulator instability in the sandbox) but is the high-contrast case and the light scrim was tuned to 0.4 to avoid an "accidental dark mode."

### DIFF
--- a/packages/mobile/src/components/climb-actions/ClimbReactionMenu.tsx
+++ b/packages/mobile/src/components/climb-actions/ClimbReactionMenu.tsx
@@ -28,7 +28,8 @@ import { useGradeFormat } from '../../hooks/use-grade-format';
 import { useTheme } from '../../providers/theme-provider';
 import type { BoardConfig } from '../../providers/drawer-host-provider';
 import { springs, timing } from '../../theme/animations';
-import { spacing, borderRadius } from '../../theme/tokens';
+import { spacing, borderRadius, overlays } from '../../theme/tokens';
+import { useEffectiveSurfaceMode } from '../../hooks/use-effective-surface-mode';
 import { useClimbActions } from './use-climb-actions';
 
 type ClimbReactionMenuProps = {
@@ -86,6 +87,7 @@ export function ClimbReactionMenu({
   const insets = useSafeAreaInsets();
   const { width: windowWidth, height: windowHeight } = useWindowDimensions();
   const { formatGrade } = useGradeFormat();
+  const surfaceMode = useEffectiveSurfaceMode();
 
   const progress = useSharedValue(0);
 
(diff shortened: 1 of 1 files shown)


marcodejongh:
Superseded by #3941, which is green and ready for review. Not a rebase — the branch here is five weeks stale and the diff needed restructuring rather than conflict resolution.

**The diagnosis in this PR was right; the fix was at the wrong layer.** `GlassSurface.tsx` puts the consumer's `style` and `elevationStyle` on one and the same view, so *any* consumer passing `overflow: 'hidden'` clips its own Android cast away. Fixing it in `ClimbReactionMenu` left the shared primitive broken for everyone else. #3941 fixes `GlassSurface` itself — the cast stays on the unclipped outer view and the clip is hoisted onto an inner wrapper.

**And the change here is no longer safe as written.** This PR deletes `overflow: 'hidden'` from `menuCard` on the grounds that "rows are transparent, so nothing pokes past the rounded corners". That stopped being true: `main` has since added a `menuScrollFade` `LinearGradient` (absolutely positioned, near-opaque end stop) and the `InlinePlaylistPicker` view *inside* the card. Without the clip, both paint square corners over the card's rounded bottom edge. #3941 **keeps** the clip and makes `GlassSurface` cope with it.

Everything else here is carried over and extended: the `role="high" level="level3"` bump, the surface-mode scrim/blur gate, dividerless rows and 24dp icons. Plus what this PR was missing — the bottom fade is now derived from the card's tone instead of a hardcoded rgba that the role bump would have silently desynced, the `## Release Notes` section that was failing CI, the wrong `accessibilityLabel` on the dismiss scrim that @claude flagged on 2026-06-22, and tests (the suite here mocked `GlassSurface` to a bare `div`, so every single thing this PR changed was invisible to it).

The `pr-3058` OTA channel is stale — it was published against a five-week-old base. #3941 has its own `pr-3941` channel for device QA. Closing this in favour of that one.

PR #3060, opened 2026-06-19
fix(mobile): separate ActiveContextBar positioning from Reanimated entering animation (Android 16 touch-blocking)

marcodejongh:
## Problem

On Android 16 (Pixel 10), selecting any climb on the climb list screen made the screen completely unresponsive to scroll and touch input. Only left/right tab swiping, the queue modal, and split-screen mode would unblock it. Uninstall/reinstall didn't fix it; clearing storage and re-logging in did, but only until selecting the next climb.

**Root cause:** `ActiveContextBar` renders a Reanimated `Animated.View` with both `entering={FadeIn}` and `pointerEvents="box-none"` on the same element. On Android, Reanimated's `entering` animations ignore `pointerEvents` — the animated view absorbs all touch events regardless. Since `ActiveContextBar` is `position: 'absolute'` (mounted at the app root, outside the navigation Stack) with no explicit `top` or `height`, it can expand to cover the full screen above its `bottom` anchor, blocking the entire climb list.

This explains every observed symptom:
- **Frozen after selecting a climb**: `ActiveContextBar` mounts with `FadeIn`, Reanimated ignores `pointerEvents`
- **Queue modal still scrollable**: gorhom renders bottom sheets in a new native Android window
- **Left/right swipe still works**: tab gesture runs above the blocking view
- **Split-screen unblocks it**: window resize triggers layout recalculation, remounting the bar
- **Works on GrapheneOS Pixel 7**: different Android version, less aggressive behavior
- **Reinstall doesn't fix it**: triggered on every session the moment a climb is selected

## Fix

Separate concerns: the outer positioning shell is now a plain `View` (which correctly passes `pointerEvents="box-none"` through on all Android versions), and only the inner `Animated.View` runs the `FadeIn` entering animation. The inner view is content-sized with no absolute positioning, so even if Reanimated ignores its `pointerEvents`, it can only block touches over the visible bar area.

This is the same pattern already used correctly in `QueueAddedSnackbar` and `UndoWallChangeSnackbar`.

## Test plan

- [ ] Select a climb on Android 16 — climb list should remain scrollable and touch-responsive
- [ ] The `ActiveContextBar` fade-in animation should still play when a climb is first selected
- [ ] Queue modal, tab switching, and all other controls should work normally
- [ ] Pixel 7 / older Android versions should be unaffected
- [ ] All 2491 mobile tests pass, Metro bundle clean

https://claude.ai/code/session_01LPoqEUJrxgvn3koVDDjbat

### DIFF
--- a/packages/mobile/src/components/queue-control/ActiveContextBar.tsx
+++ b/packages/mobile/src/components/queue-control/ActiveContextBar.tsx
@@ -71,9 +71,13 @@ export function ActiveContextBar({
     ? (measuredTabBarHeight ?? bottomChrome.tabBarBottom) - TABBAR_SEAM_OVERLAP
     : bottomChrome.tabBarBottom + gapAboveTabBar;
 
+  // Outer shell: plain View so `pointerEvents="box-none"` is respected on Android.
+  // Reanimated's `entering` animations ignore pointerEvents on the animated view
+  // itself, causing the overlay to block all touches below it (Android 16 / API 36
+  // regresses this the hardest). The inner Animated.View is sized to its content
+  // so even without pointerEvents it only covers the visible bar. #5728
   return (
-    <Animated.View
-      entering={reduceMotion ? undefined : FadeIn.duration(timing.normal)}
+    <View
       pointerEvents="box-none"
       style={[
         styles.toolbar,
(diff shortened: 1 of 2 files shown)


marcodejongh:
## Android sanity screenshots

Tested on the headless x86_64 emulator (Android 16, API 36) against the local seeded dev backend.

**Climbs list — before selecting a climb**

Climbs list

**After selecting Moon Landing — `ActiveContextBar` visible above tab bar, list remains fully scrollable and interactive**

Climbs list with ActiveContextBar showing Moon Landing V3

The bar renders above the tab bar and passes touch events through to the climb rows (the `pointerEvents="box-none"` on the outer plain `View` is respected, unlike when it was on the `Animated.View` with `entering={FadeIn}`).

PR #3061, opened 2026-06-19
ci(mobile): gate native builds on the OTA fingerprint (build only when new)

marcodejongh:
## What

Self-hosted production OTA is live on `main`, so every mobile push to `main` fires **both** the ~60-min native builds (`ios-testflight-rn`, `android-apk-rn`) **and** the cheap OTA publish. For a JS/TS-only change the `runtimeVersion` fingerprint is unchanged, so installed binaries pull the new JS over the air and the native build is wasted.

This adds the self-hosted equivalent of Expo's `continuous-deploy-fingerprint`: each native workflow gets a cheap Linux **`gate` job** that resolves its platform fingerprint and **skips** the native build when a `fingerprint-<platform>-<hash>` tag already exists (OTA delivers the JS); otherwise it **builds** and, on a successful upload, **pushes that tag**. The OTA publish is unchanged — it still runs on every push.

## Why a wrong skip is impossible

The gate resolves on Linux, but the iOS binary is baked on macOS. A wrong skip would strand users on an old native binary, so rather than trust cross-OS fingerprint determinism for correctness:

- the **gate** checks the **Linux**-resolved fingerprint against existing tags;
- the **build job** re-resolves on its **own runner** and tags **that** value.

Tags therefore only ever hold build-OS fingerprints. If Linux and macOS ever disagree, the gate never finds a matching tag → iOS simply **always builds** (wasteful, safe), never a wrong skip. The build emits a `::warning::` when the two disagree so the degraded mode is visible. (Android builds on Linux like its gate, so they always agree.) This is strictly safer than Expo's action, which trusts cross-OS determinism for correctness.

## Key details

- **Workflow-level shared env.** The `EXPO_PUBLIC_*` / `EXPO_UPDATES_*` env (and Android's `GOOGLE_MAPS_API_KEY`) move to workflow level so the gate and build resolve a byte-identical fingerprint and can't drift. The gate writes the same `.env` the build does — **the `.env` is itself hashed into the fingerprint** (verified: iOS `9f9cc7b…` without `.env` vs `9b4c630…` with it).
- **Per-platform tags + env.** iOS resolves **without** `GOOGLE_MAPS_API_KEY`, Android **with** it — the key flips the Android fingerprint (`82cfb16…` vs `cdb7ea9…`), so the per-platform tags differ; a shared tag would be wrong.
- **Fail-safe.** A resolve failure builds. `workflow_dispatch` bypasses the tag check and always builds — for iOS that means dispatching on `main` (the iOS build is `main`-only by design, since it uploads to TestFlight); Android additionally builds from a feature-branch dispatch (artifact-only, matching its pre-existing behavior).
- **Idempotent / race-safe tag push.** A concurrent run that already pushed the tag is treated as success.
- **Android tags only on a full ship.** The `fingerprint-android-<hash>` tag is pushed only after the sideload APK release, the AAB build, and a Play upload that did not fail — so a failed AAB or flaky Play upload leaves no tag and the next push rebuilds and retries automatically.

## Tests / validation

- `scripts/mobile-ci-env-parity.test.ts` updated to read **workflow-level** env, plus new assertions that each native build resolves its platform fingerprint, tags it, and keeps the per-platform `GOOGLE_MAPS_API_KEY` split (iOS none, Android set). **14/14 pass.**
- `vp run typecheck` clean; `vp check` clean on all changed files; all three workflow YAMLs parse.
- Confirmed locally that `bunx expo-updates runtimeversion:resolve` emits JSON (`.runtimeVersion`), and that `.env` + `GOOGLE_MAPS_API_KEY` both move the fingerprint.

## Follow-up to verify after merge

- A JS-only commit → gate finds the existing fingerprint → both native builds skip (OTA-only).
- A native change → new fingerprint → builds run and push new `fingerprint-<platform>-<hash>` tags; a subsequent JS-only push skips again.
- One-time: confirm the Linux gate fingerprint equals the macOS-baked `EXUpdatesRuntimeVersion` (expected for this CNG project) so iOS savings actually land — safety doesn't depend on it.

Docs: `docs/mobile-ota-updates.md` gains a "Native-build gating" section (incl. manual overrides) and the item is removed from Deferred.

https://claude.ai/code/session_01ECKU2KefyLAzys1zByyPKp

### DIFF
--- a/.github/workflows/android-apk-rn.yml
+++ b/.github/workflows/android-apk-rn.yml
@@ -33,41 +33,133 @@ concurrency:
 permissions:
   contents: write # required for softprops/action-gh-release
 
+# ─── SHARED BUILD ENV (workflow-level) ───────────────────────────────────────
+# Declared at the workflow level so the cheap `gate` job (fingerprint resolve)
+# and the `build-and-release` job (expo prebuild) resolve a byte-identical config
+# and can't drift. MUST stay identical to ios-testflight-rn.yml /
+# mobile-ota-production.yml — scripts/mobile-ci-env-parity.test.ts fails the
+# build if they do. Config-affecting vars (EXPO_PUBLIC_GOOGLE_IOS_CLIENT_ID,
+# GOOGLE_MAPS_API_KEY, EXPO_UPDATES_URL, EXPO_UPDATES_CHANNEL) move the
+# fingerprint runtimeVersion directly; the rest reach it via the hashed .env file
+# the jobs write. Don't add other fingerprint-affecting vars here (EAS_BUILD,
+# EAS_BUILD_PROFILE, BOARDSESH_APP_VARIANT, …) — they're unset in the build too.
+env:
+  EXPO_PUBLIC_BACKEND_URL: https://ws.boardsesh.com
+  EXPO_PUBLIC_WS_URL: wss://ws.boardsesh.com/graphql
+  EXPO_PUBLIC_WEB_URL: https://www.boardsesh.com
+  EXPO_PUBLIC_GOOGLE_IOS_CLIENT_ID: 401523882502-0f6d7te1vekvkpmg18t8di6l0ig560q7.apps.googleusercontent.com
+  EXPO_PUBLIC_GOOGLE_ANDROID_CLIENT_ID: 401523882502-h92kdkck1qhmdbgq7ltek87g4rg8rg3h.apps.googleusercontent.com
+  EXPO_PUBLIC_GOOGLE_WEB_CLIENT_ID: 401523882502-hvfhh79p4q1qq0md7lk646c6sgmsi9q0.apps.googleusercontent.com
+  # Self-hosted OTA (expo-open-ota). EXPO_UPDATES_URL is the server's manifest
+  # endpoint (e.g. https://ota.boardsesh.com/manifest), set as a repo variable
+  # so the domain isn't hardcoded; until it's set, app.config.ts falls back to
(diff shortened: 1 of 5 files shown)


marcodejongh:
Addressed the review feedback in `5c13660c4`:

**Codex (P2 — Android tag before AAB build):** Moved the `fingerprint-android-<hash>` tag to *after* the AAB build + Play upload and gated it on `steps.github_release.outcome == 'success' && steps.build_aab.outcome == 'success' && steps.play_upload.outcome != 'failure'` (added `id: build_aab`). Now a failed `bundleRelease` or a flaky Play upload leaves no tag, so the next push rebuilds and retries automatically. A *skipped* Play upload (before the Play bootstrap) still tags, since Play isn't a channel yet.

**Claude review #1 (iOS `workflow_dispatch` escape hatch):** The iOS gate is `main`-only *by design* — the iOS build uploads to TestFlight, and `build-and-upload` already required `main`, so enabling feature-branch iOS dispatch would only run the gate with no build. Kept the behavior and documented the per-platform difference (iOS dispatch = on `main`; Android also builds artifacts from a feature-branch dispatch) in the iOS gate comment, the docs, and the PR description.

**Claude review #2 (step ID references):** Confirmed against the full workflows — `testflight_upload`, `github_release`, `build_aab`, and `play_upload` all exist; the tag steps reference real IDs.

**Claude review #3 (structural-only tests):** Agreed — left as-is; the gate's `if:`/`$GITHUB_OUTPUT` logic is shell in YAML, which is validated by the parity test's structural assertions plus the workflow runs themselves rather than unit tests.

PR #3073, opened 2026-06-19
feat(mobile): connection-aware Android session notification + Material 3 redesign

marcodejongh:
## Release Notes

Drive your board from the Android session notification
The session notification now shows your current climb with its grade, angle, and your spot in the queue, and draws the board art right on your phone. Previous and Next move the board through your queue without opening the app, and the lightbulb shows when you're connected. When a crewmate takes the board, the card steps back to show what's on the wall.

## What & why

Brings the Android foreground-service notification to functional parity with the iOS Live Activity, and fixes the lightbulb so it reflects who actually holds the board.

> **Stacked PR.** Branches off the unmerged iOS "connection-aware Live Activity" work (commit `41b867020`), which adds the shared `boardConnection` tri-state plumbing this PR consumes. **Review my commits only** — the iOS/backend commit in the diff belongs to that separate branch and drops out once it lands in `main` (or retarget this PR's base to it).

### Notification (native Android)
- **Material 3 card** — `DecoratedCustomViewStyle` with a portrait climb thumbnail, grade·angle, "N / M" queue position, a lightbulb, and Previous/Next. Material You accent on Android 12+, brand-green fallback below.
- **On-device thumbnail, no backend** — drops the `/api/internal/board-render` fetch. The bridge renders the current climb's holds-only PNG via the existing `BoardRenderer` native module and resolves the bundled board backgrounds; the service composites them locally (the app's "no-network board art" rule). iOS keeps its ActivityKit fetch (the documented workaround).
- **Connection-driven** — lightbulb + control visibility follow the shared `boardConnection` tri-state: `connectedByMe` → lit bulb + Prev/Next; `heldByPeer` → bulb out, controls hidden, "`<name>` is on the wall"; `disconnected` → bulb out.
- **Tappable lightbulb** — off → reconnect to the last board; lit → re-assert the current climb (new Android-only `boardControl` event → `bluetooth.connect` / `reassertWall`).
- **Above the "Silent" fold, still silent** — a new `boardsesh_session_active` channel at `IMPORTANCE_DEFAULT` (legacy LOW channel deleted on migration) with sound + vibration muted, so it isn't buried in the Silent group yet never beeps on session start.
- **Bug fix** — Previous/Next sent the *current* index for both buttons (a no-op against the absolute-index JS bridge); they now send `currentIndex ∓ 1`.

### Shared JS
Additive only (the live-activity JS is shared with the iOS branch): a `boardControl` event, the on-device thumbnail-path fields, and three i18n keys (`relight` / `reconnect` / `onWall`) across en-US / es / fr. No existing iOS-read field is changed.

## Not in this PR
A BLE scan fix (scan unfiltered so name-only boards surface) was split out to a side branch — it needs screen-off filtering, a second scan path, and on-device confirmation before it's ready.

## Testing / validation
- `vp run typecheck:mobile`, `vp check`: clean.
- JS: full mobile suite (**2518**), live-activity + i18n parity included.
- Kotlin (Robolectric, JDK 21, `:boardsesh-live-activity-module:testDebugUnitTest`): **33** — tri-state actions, bulb-handoff PendingIntent refresh, prev/next target index, async on-device compose + stale-drop, the `downscale` Binder-size cap, the channel LOW→DEFAULT-but-silent migration, the event buffer/replay multiplex, and the controller lifecycle.
- Verified on a physical Pixel 8 Pro (Android 16): the card renders the climb on-device, shows above the Silent fold, and reflects the connection state.
- A 4-reviewer pass plus external review were addressed (BLE split out; iOS-render gated; late-background re-post; silent channel; docs corrected for the anonymous-holder APNs fallback).

https://claude.ai/code/session_018qhP6HkkyXBgvQG2cGedF9


marcodejongh:
Addressed the review notes (commits `d468d3e4c` + `2f0a45a3e`):

**Applied**
- `composeThumbnail` recycles the decoded overlay + background bitmaps and the pre-downscale intermediate (only the final rounded bitmap is retained).
- `bitmapCache` now caps by **bytes** (`sizeOf = byteCount`, ~2 MB) instead of entry count.
- `onWallTemplate` uses `String.replace(String, String)` — unambiguously literal, no regex overload to resolve against — instead of `replaceFirst`.
- Edge-case test: `connectedByMe` with only Next (first climb) / only Previous (last climb) beside the lit bulb.
- `backgroundsKey` uses `JSON.stringify` instead of a `|` join (no delimiter a path could contain).
- `@VisibleForTesting` on the purely-test-only helpers (`resetImageStateForTest`, the pending-buffer snapshot/clear).

**Kept as-is (with reason)**
- `void bt.connect(...)` matches the in-app `useLightbulbControl` exactly, and the bluetooth provider surfaces connect errors via its own toast — the caller doesn't own error feedback.
- `{{name}}` substituted in Kotlin, not JS `t()`: JSON catalogs can't carry a comment without breaking the i18n parity test, and `{{name}}` is the standard placeholder translators keep verbatim.

**Out of scope (now in `main` via #3077):** the backend `server.ts` APNs "emit `disconnected` when the holder clears" note — real follow-up, but it's the iOS/backend path, not this Android-only PR.

marcodejongh:
Good catch — that was a real bug, fixed in `ca561d4b7`.

The composite depends on overlay **+** backgrounds, but the cache / in-flight set / stale-guard keyed on the overlay alone, so a late-background re-fire hit the cached holds-only composite and the board photo never landed. Now `imageKey = overlay + NUL-joined backgrounds` threads through `bitmapCache`, `inFlight`, `currentImageKey`, and `currentBitmap()` — a late-background update is a cache miss and recomposes.

- New Kotlin regression test: `recomposes when the backgrounds resolve late` (same overlay, empty→populated backgrounds → two composes; fails on the old key).
- Covered the solo-mode analytics branch in `live-activity-bridge.test.tsx` (the reconnect tests only exercised party mode).

marcodejongh:
Addressed in `8a90d49c7`:

**Blocking — stacked PR: stale, already resolved.** This branch is rebased onto `origin/main`; the iOS "connection-aware Live Activity" work merged separately (#3077). `git log origin/main..HEAD` is 7 commits, all `mobile`/Android, and `git diff origin/main...HEAD` touches zero iOS/backend/Swift files — the bot compared against a pre-rebase snapshot. Merge target is clean (`MERGEABLE`/`CLEAN`, base `main`).

**Non-blocking — `deleteNotificationChannel` on every update:** guarded behind `getNotificationChannel(LEGACY_CHANNEL_ID) != null`, so it's a true one-time migration instead of churning the manager on every `onStartCommand`. The existing migration test still passes (legacy channel deleted on first start).

**Non-blocking — bitmap parcelled twice:** correct — both the collapsed and expanded RemoteViews carry it and RemoteViews doesn't dedupe, so the Binder transaction is 2× the single-bitmap size. Fixed the misleading comment: worst case is a square board at the 256 px cap = ~512 KB for both copies, still under the ~1 MB limit (portrait boards, the common case, are smaller). Kept the 256 px cap — lowering it would soften the expanded image for no real headroom gain.

marcodejongh:
Thanks — picked up the two actionable minor notes in `4a72722ed`:

- **Hardcoded connection literal:** `SessionPresenceController` now uses `BoardSessionService.CONNECTION_CONNECTED_BY_ME` for the initial-state fallback instead of the `"connectedByMe"` string.
- **Holder-line clear, untested:** added a regression test — `heldByPeer` + holder sets `"Alex is on the wall"`, then a `connectedByMe` update (which omits `EXTRA_HOLDER_NAME`) clears it. The test inflates the expanded `RemoteViews` and asserts the rendered `TextView` directly (`buildContentView` is now `@VisibleForTesting`).

Leaving the other two as-is: the `backgroundsKey` `JSON.stringify` is a once-per-render-change compute (not hot) and reads clearly, and the stale-PendingIntent window is — as noted — cosmetic under Aurora's last-connection-wins.

marcodejongh:
Both picked up in `2c035ea78`: `maybeComposeImage` now catches `Throwable` (so an OOM compositing a large board PNG degrades to no-thumbnail instead of killing the compose thread), and the `bt` local is renamed `bluetoothCtx` (avoids shadowing the outer `bluetooth`).

marcodejongh:
**Bug 2 (bitmap leak) — fixed in `202090f3c`.** Real catch: `composeThumbnail` only recycled intermediates on the success path, so a mid-compose OOM (now swallowed by the `Throwable` catch) would leak the decoded bitmaps. Wrapped the body in try/finally — every decoded/intermediate bitmap recycles on all paths, only the returned `rounded` (always a fresh `roundCorners` allocation) survives; the per-layer background decode recycles in its own finally too.

**Bug 1 (index clamping) — already defended, not a real crash.** The premise (`queue[-1]` reaching the reducer) doesn't hold:
- The Previous/Next actions are only added when `hasPrevious`/`hasNext` are true (`buildNotification`), so the out-of-bounds tap isn't normally reachable.
- The JS bridge has an explicit range guard before any array access: `live-activity-bridge.tsx:127` — `if (event.currentIndex < 0 || event.currentIndex >= queue.length) return;`. The native comment in `actionPendingIntent` documents this reliance.
- It's already tested: `live-activity-bridge.test.tsx:205` — `ignores out-of-range indices instead of wrapping or crashing` covers `currentIndex: 99` and `currentIndex: -1`.

I'm leaving native clamping out on purpose: clamping a stale-notification tap to `[0, total-1]` would turn today's clean drop into a navigation to an unintended-but-valid climb during a race, which is worse than the no-op the guard already gives.

marcodejongh:
Picked up the cosmetic one in `1390270d4`: the `backgroundsKey` rationale now sits inline on the dep in both effects.

Leaving the other two — both are net-neutral-to-negative on a controlled internal path:

- **`ensureChannel()` binder call:** posting the notification (`notify()`) is itself a binder IPC, and updates are user-paced (climb/connection changes), not a hot loop. Swapping the one `getNotificationChannel` read for a `SharedPreferences` flag trades a cheap IPC for disk IO plus persistent state and a new failure mode — a worse deal than the read it replaces. The expensive part (the actual `deleteNotificationChannel` mutation) is already gated.
- **`boardConnection` stringly-typed:** the value crosses the JS↔Kotlin boundary as a string by necessity (intent extras / expo-modules), so a Kotlin enum can't change the wire format — it just adds parse-and-map boilerplate around a contract the `CONNECTION_*` constants already name. The unknown-value → `connectedByMe` default is a deliberate safe fallback for an internal-only channel.

PR #3076, opened 2026-06-20
feat(mobile): recent beta-links shelf on the profile page

marcodejongh:
## Summary

Adds a horizontal shelf of a climber's **recent beta videos** to the mobile profile page, with infinite scroll and a "See all" grid — the same UX the Discover tab uses for playlist shelves. The playlist slider was generalized so both features share one component.

Shows on **both** the public profile (`users/[userId]`) and the own "You" tab. Hidden for climbers who haven't shared any beta.

**What's included**
- `HorizontalScrollSection` — the Discover playlist slider, generalized (optional `minHeight`) and reused by playlists + the beta shelf.
- `useUserBetaLinks` — offset-paged hook (cross-page dedupe + video-only filter), modeled on `useUserPlaylists`.
- `ProfileBetaShelf` — rendered at the top of `ProgressTab`, so it appears on both profile screens from one insertion point.
- `users/[userId]/beta` — full 3-up `FlashList` grid for "See all" (`[userId].tsx` became `[userId]/index.tsx`; route path unchanged).
- Backend: a backward-compatible `offset` arg on the `userBetaLinks` resolver + SDL so the shelf and grid page in. The web profile beta section is unaffected (it omits `offset`).

## Release Notes

See a climber's recent beta videos right on their profile — swipe through the shelf or tap See all for the full grid.

## Test plan

- `vp run typecheck:mobile`, `vp run typecheck:backend` — pass.
- `vp run test:mobile` — 2503 pass (incl. updated Discover screen test).
- Backend `beta-link-queries` — 39 pass, incl. 2 new tests covering the `offset` path.
- `vp test run catalog-completeness` — 28 pass (i18n parity for en-US/es/fr `you.betaShelf.*`).
- `vp check --fix` on changed files — 0 errors.
- Manual (pending): simulator against a local backend (`EXPO_PUBLIC_BACKEND_URL`) since the new `offset` arg isn't on the prod backend yet — verify the shelf renders + pages, "See all" opens and pages, and the shelf is absent for a climber with no beta.

https://claude.ai/code/session_018aTHqZ4EKokL7CEqgDPyNp


marcodejongh:
Thanks for the review — addressed in `abe99ac`.

**Fixed**
- **Stale in-flight fetch on user change** (Claude + Codex P2): `useUserBetaLinks` now carries a `generationRef`. A reset (userId change) or `refetch` bumps the generation and clears `isFetchingRef`, so the new fetch isn't blocked by an in-flight one, and any superseded response is discarded (it also won't free the fetch flag the new request owns). New test `discards a stale in-flight response when the user changes` covers exactly the navigate-between-profiles case.
- **`refetch` dropped while in-flight + lingering error** (Claude): the reset path (now shared by mount and `refetch`) clears `isFetchingRef` and calls `setHasError(false)` before re-fetching. New `error → retry` test covers it.
- **Non-unique `createdAt` offset order** (Codex P2): the resolver now orders by `createdAt` plus the PK tuple (`boardType, climbUuid, link`) as a deterministic tie-breaker, so paging can't duplicate/skip rows around a shared-`createdAt` boundary.
- **Missing tests** (Claude): added `use-user-beta-links.test.tsx` — page-0 fetch, video-only filter, `loadMore` offset + cross-page dedupe, error→retry, stale-response discard, and no-userId.

**Intentionally not changed (false positives for this codebase)**
- **`estimatedItemSize` on FlashList** (Claude): FlashList is v2 (`^2.3.1`), which **removed** `estimatedItemSize` — passing it fails typecheck. The existing `ProfileClimbsTab` FlashList omits it for the same reason.
- **`Stack.Screen name="users/[userId]/index"`** (Claude): this is correct. Hoisted routes (no intermediate `_layout`) use the full path — the root layout already registers `users/search` for `app/users/search.tsx`, and `index` for `app/index.tsx`. So `app/users/[userId]/index.tsx` → `users/[userId]/index`. Avatar taps navigate by href (`/users/[userId]`), which resolves to the index route unchanged.

PR #3090, opened 2026-06-20
fix(backend): show playlist grades at the selected wall angle (#1596)

### LINKED ISSUE #1596: [Bug] Climb grade displays at wrong angle — uses most-ascents angle instead of selected angle
## Description

When browsing climbs (specifically in a playlist), the displayed grade uses the angle with the most ascents rather than the user's currently selected wall angle.

## Steps to Reproduce

1. Set a specific wall angle (e.g. 40°)
2. Open a playlist and scroll through climbs
3. Observe that some climbs display a grade corresponding to a different angle than the one selected

## Example

- Climb: [nothing without anything](https://www.boardsesh.com/kilter/original/12x12-square/screw_bolt/40/view/nothing-without-anything-B3F409F6C4474F4BB2E33CD73033FA66)
- Selected angle: 40°
- Displayed grade: V13 (the grade for the angle with the most ascents)
- Expected grade: V12 (the grade at 40°)

## Hypothesis

When a climb is added to a playlist, the grade may be stored/cached at the angle it was added at rather than being recomputed at the user's current angle. The grade lookup may also be falling back to the angle with the most ascents as a default.

## Environment

- Confirmed on iOS; not yet verified on other platforms
- Reproducible playlist: https://www.boardsesh.com/playlists/fb48f5a0-79ad-4658-a40f-a91891db9de1

marcodejongh:
## Summary

Closes #1596.

In a playlist, a climb's grade showed the angle with the **most ascents** instead of the wall angle you'd selected — e.g. V13 (the popular angle) when you were set to 40° and the climb is V12 there. Confirmed on iOS.

The mobile playlist list runs in **all-boards mode** (so mixed-board playlists still show every climb, dimming the ones that don't fit your board). That path never told the server which angle was selected, so the grade fell back to the most-ascended angle. Web already passes a concrete board + angle when a board is picked, so it wasn't affected.

What changed:
- Added optional `activeBoardName` / `activeAngle` to the all-boards `playlistClimbs` input. Climbs on your active board now resolve their grade at the **selected** angle; off-board climbs keep their added-at / most-ascents fallback, so mixed-board playlists still list everything.
- `hydrateClimbsByRefs` gained an `EXISTS` guard: if a climb has no stats at the selected angle, the angle `COALESCE` drops back to the most-ascents angle instead of blanking the grade. The returned `angle` now follows the stats row that was actually joined, so the angle and the grade always agree.
- Mobile playlist screen reads the active board and keys the query on it, so switching board/angle refetches at the new angle.

## Release Notes

- Playlist grades now match the wall angle you've dialed in, instead of a climb's most-popular angle

## Test plan

- `vp run typecheck` (all packages) — green
- `vp run test:mobile` — 2516 passed
- Backend tests for the touched areas — `hydrate-climbs`, `playlist-queries`, `smart-playlists`, `operations-schema-validation`, `resolver-frames-fields` all green
- New real-DB integration test `playlist-climbs-active-angle.test.ts` proves: on-board climb renders at the selected angle (40°, not its most-ascents 50°); a climb with no stats at 40° falls back to most-ascents (grade not blanked); an off-board climb is untouched
- `vp run codegen` — generated schema/types regenerated, no drift
- Pre-existing unrelated failures in `integration.test.ts` / `board-presence.test.ts` (WebSocket daemon) reproduce on a clean tree without these changes

https://claude.ai/code/session_0157PsyHaLfWs8aXxN1T3No9

### DIFF
--- a/packages/backend/src/__tests__/hydrate-climbs.test.ts
+++ b/packages/backend/src/__tests__/hydrate-climbs.test.ts
@@ -225,8 +225,40 @@ describe('hydrateClimbsByRefs', () => {
       expect(result[0].angle).toBe(40);
     });
 
-    it('uses the override when it is a non-null number, even if statsAngle is set', async () => {
+    // The returned `angle` reflects the angle the stats row was actually joined
+    // at (`statsAngle`), so it always agrees with the difficulty/quality shown.
+    // The SQL joins at COALESCE(override-if-stats-exist, most-ascents), so when
+    // an override is applied the joined statsAngle IS the override angle.
+    it('returns the joined stats angle, which under an applied override is the override angle', async () => {
+      mockDb.select.mockReturnValueOnce(makeChain([{ ...baseRow, statsAngle: 25 }]));
+
+      const overrides = new Map<string, number | null>([['kilter:c1', 25]]);
+      const result = await hydrateClimbsByRefs([{ climbUuid: 'c1', boardType: 'kilter' }], {
+        angleOverrides: overrides,
+      });
+
+      expect(result[0].angle).toBe(25);
+    });
+
+    // When the override angle has no stats row, the SQL EXISTS guard drops the
+    // override and the join falls back to the most-ascended angle. The returned
+    // angle follows that fallback (statsAngle) so grade and angle stay in sync,
+    // rather than reporting an angle whose grade isn't shown.
+    it('follows the joined fallback angle when the override angle has no stats', async () => {
       mockDb.select.mockReturnValueOnce(makeChain([{ ...baseRow, statsAngle: 35 }]));
+
+      const overrides = new Map<string, number | null>([['kilter:c1', 25]]);
+      const result = await hydrateClimbsByRefs([{ climbUuid: 'c1', boardType: 'kilter' }], {
+        angleOverrides: overrides,
+      });
+
+      expect(result[0].angle).toBe(35);
+    });
+
+    // Only when the climb has no stats row at all (nothing to join) does the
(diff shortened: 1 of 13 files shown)


marcodejongh:
Thanks — addressed both in `1dcff272c`:

**Cache key bug (fixed):** The selected angle is now its own query-key segment in `use-playlist-climbs.ts` (`activeAngle` in all-boards mode, `angle` in specific-board mode). Dialing to a new angle on the same board now re-keys and refetches instead of serving stale grades. This also closes the same latent gap for web's specific-board mode. Prefix invalidations (`['playlistClimbs', playlistUuid]`) still match.

**Raw SQL in the integration test (kept, now documented):** This one's a genuine raw-SQL exception. The integration DB is built from a minimal hand-maintained DDL (`schema-sql.ts`), not the full Drizzle schema. `db.insert(boardClimbStats)` emits every default-bearing column — including `quality_normalized`, `*_ascensionist_count`, etc. — which that DDL omits, so the builder insert fails with `column "quality_normalized" does not exist`. Naming only the columns the test DDL declares requires raw `sql`; I added a comment explaining this. (`TRUNCATE ... RESTART IDENTITY CASCADE` has no builder form either.)

marcodejongh:
Addressed the co-validation note in `a94c7d254`: `GetPlaylistClimbsInputSchema` now has a zod `refine` requiring `activeBoardName` and `activeAngle` to be present together (both-or-neither), so a half-specified request is rejected instead of silently no-opping. Added a resolver test covering the rejection. Our clients already always send both or neither.

PR #3094, opened 2026-06-20
fix(web): move climb-stats off the uncached REST route to GraphQL + cache it

marcodejongh:
## Summary

A usage-anomaly alert fired: a **9.5× spike in serverless function invocations**, dominated by `/api/v1/[board_name]/climb-stats/[climb_uuid]` at **5.7K invocations** — roughly **55× the next-busiest route**.

**Root cause.** That route ran one Postgres query and returned JSON with **no caching of any kind** — no `Cache-Control`, no `revalidate`, no `unstable_cache`. Every request was a cold function invocation hitting the DB. Sibling routes already cache at the edge (`/grades`, `/slugs/layout`). The endpoint is also publicly documented (`openapi-routes.ts`), unauthenticated, and (until now) unrated — so on top of legitimate angle-drawer opens, anything crawling the documented API hit Postgres directly. We can't read prod logs here to prove a crawler, but the signature (one uncached public endpoint at 55× its neighbors, while the only app caller is gated behind a drawer + 5-min client cache) fits automated scraping.

### What changed

- **Moved the legit traffic to GraphQL.** New `climbStatsForAngles(boardName, climbUuid)` query reads the live `board_climb_stats` table (including the first-ascent fields), mirroring the existing `climbStatsHistory` plumbing and rate-limited 60/min like its sibling resolvers. `AngleSelector` now fetches through the backend (`graphql-request` HTTP) instead of a Vercel function, and no longer refetches on window focus.
- **Cached the still-public REST route at the edge** — `Cache-Control: public, s-maxage=300, stale-while-revalidate=86400` (5 min fresh, 1 day SWR). Repeat hits to the same climb are served by the CDN instead of invoking the function. The route stays so documented API consumers keep working.
- **Per-IP cap (120/min) in the climb-stats route handler** — returns `429 + Retry-After` (clamped to ≥1s) and logs user-agent + IP on the 429 path so a future alert window can confirm the source. It lives in the route handler (Node serverless runtime), where the in-memory limiter works per-instance — the same mechanism `/api/auth/register` already relies on. A cache *miss* (unique `climb_uuid`) is what costs an invocation, so capping misses is what blunts a UUID-enumerating scraper.

### Trade-offs / notes

- The route-handler limiter is in-memory and per-instance on Vercel, so it's not a strict cross-instance cap — strict limiting would need a shared store (Vercel KV / Upstash), tracked in #3096. The edge cache is the primary lever for invocation cost; the per-IP cap protects cache misses.
- The public REST endpoint is intentionally kept (it's in the OpenAPI docs); it just no longer costs an invocation per repeat request.

> Earlier revision put the cap in Edge middleware; review correctly flagged that as a near no-op (Edge isolates don't persist the in-memory Map). Relocated to the Node route handler where it actually fires.

## Release Notes

## Test plan

- `vp run codegen` — generated types include `climbStatsForAngles` / `ClimbStatsForAngle`.
- `vp run typecheck:web`, `typecheck:backend` — clean.
- New backend resolver test (`climb-stats-for-angles.test.ts`, 7 cases): difficulty label from rounded `displayDifficulty`, fa passthrough, null handling, empty result, input validation, and rate-limit enforcement (applies the 60/min limit + short-circuits the DB on rejection).
- New REST route test (`route.test.ts`, 3 cases): 200 carries the edge `Cache-Control` header, 429 returns `Retry-After` and skips the DB, back-off never drops below 1s.
- Existing `angle-selector.test.tsx` (6 cases) still green; operations smoke + schema-validation suites green. Full CI: 4847+ passed / 0 failed.
- Manual: open a climb → open the angle drawer → stats render the same; Network shows the GraphQL POST to the backend, no `/api/v1/.../climb-stats/...` call. REST route now returns the cache header; hammering past the cap returns `429`.

https://claude.ai/code/session_01Cv5zZV3Q8PSgvgMugkY8bG


marcodejongh:
Thanks — addressed in `8a804258`:

- **#2 REST route tests** — added `route.test.ts`: 200 carries the `Cache-Control` header, 429 returns `Retry-After` and skips the DB.
- **#3 Follow-up issue** — filed #3096 for the distributed (shared-store) rate limiter and referenced it from the route comment.
- **#4 `Retry-After`** — clamped with `Math.max(1, …)` so a sub-second window never advertises "retry now"; covered by a test.

On **#1 (redundant board-name check)** I'm leaving it as-is. `validateInput(BoardNameSchema, …)` + `isValidBoardName(…)` is the established idiom across every resolver in `climbs/queries.ts` (`climbStatsHistory` right above it, `searchClimbs`, `similarClimbs`). It's runtime-redundant but the guard also narrows the type and gives a friendlier "Must be one of …" error. Removing it from just this resolver would make it inconsistent with its siblings — if we want to drop the pattern it should be one file-wide cleanup, not a one-off here.

---

PR #3095, opened 2026-06-20
Fix(Put caching and holds prewarming to background tasks)

alejandrosanchezcabana:
## Summary
Fixing issue that makes the app go stale due to heavy foreground tasks, so moved them to background

## Release Notes

- [ ] Fix issue that made the app go stale due to heavy tasks


marcodejongh:
Thanks @alejandrosanchezcabana — great catch and fix. Your branch was based on an older `main` (before `scripts/check-commit-message.ts` existed), so its hooks/CI couldn't pass on this base. I rebased your commit onto current `main` and greened CI (added the missing `InteractionManager` mock to the climbs-list test, the real cause of `test-default` failing) in #3097, with your authorship preserved — and it's now merged. Closing this in favor of #3097. Thank you for the contribution! 🙏

PR #3099, opened 2026-06-20
perf(shared): lazy-load hole placements per board

marcodejongh:
## Summary

On Android (Hermes, no JIT) the first synchronous evaluation of the 241 KB generated `product-sizes-data.ts` blocked the JS thread and froze the app. The bulk of that file is the `HOLE_PLACEMENTS` export (~200 KB); it was pulled in — and evaluated all at once, for every board — the moment a board was selected or a board surface mounted (`getBoardRenderData` → `getHolePlacements`).

Phase 1 (already merged) deferred the prewarm off the interaction thread but the heavy all-boards eval still happened at ~7 entry points. This is the root fix: **`HOLE_PLACEMENTS` is split into one shard per board and each shard is loaded lazily**, so Hermes only ever evaluates the **active board's** holds (~30–86 KB) instead of all boards at once.

This completes Alex Sánchez's board-freeze work.

### What changed

- **Per-board shards.** `HOLE_PLACEMENTS` is now `src/generated/hole-placements/<board>.cjs`, one file per Aurora board (kilter 86 KB, tension 32 KB, decoy 22 KB, touchstone 13 KB, grasshopper 10 KB, soill 7 KB).
- **Lazy loader** (`src/hole-placements.ts`, generated). A per-board map of **literal `require('./generated/hole-placements/<board>.cjs')`** calls inside arrow functions. `getHolePlacements(board, layout, set)` routes through `getBoardHolePlacements(board)`, which requires only the requested board's shard and caches it.
- **Public API unchanged and still synchronous.** `getHolePlacements`, `getProductSize`, `getLayout`, `getImageFilename`, etc. keep their current signatures.
- **Generator** (`scripts/generate-board-constants.ts`) emits the shards + loader and re-exports `HOLE_PLACEMENTS` from the loader; `scripts/reshard-hole-placements.ts` re-emits the shards from the already-committed data (no DB needed) and is idempotent.

### Why web (and every other runtime) stays unaffected

`@boardsesh/board-constants` is consumed as raw TS **source** by both Next/webpack (server-side, via `transpilePackages`) and Metro. The lazy mechanism has to defer evaluation in **both** bundlers plus the Node/tsx scripts and the vitest test runtime, while staying synchronous. A single primitive doesn't cover all four, so the loader picks the right `require` at module load:

- **Metro (mobile) + webpack (web) + vitest** all inject a global synchronous `require`. The literal `require('./…cjs')` calls let those bundlers statically resolve each shard and **lazily evaluate it on first call**. Verified in the Metro bundle: each loader compiles to `kilter: () => hasGlobalRequire ? require(_dependencyMap[0], …) : …` — the `require` is inside an arrow function keyed to one dependency-map slot, so selecting kilter runs only kilter's module factory; the other five shards' factories never run.
- **Bare Node ESM** (the tsx generator scripts, the backend export, aurora-sync, the `db` `tsx --test` suite) has no global `require`. There the loader builds one via `process.getBuiltinModule('node:module').createRequire(import.meta.url)`. No static `node:module` import (Metro can't resolve Node builtins) and `import.meta` is read **only** in that branch, which is unreachable under Metro (where `typeof require === 'function'`). The mobile/web bundles never evaluate it.
- **Shards are `.cjs`, not `.ts`**, so a relative `require` resolves identically across Metro, webpack, tsx, and Node's CJS require (which has no TypeScript loader and can't `require('./x.ts')`).

The all-boards `HOLE_PLACEMENTS` export is preserved for the Node-side consumers that iterate every board (backend data export, aurora-sync's JSON import, the iOS Swift generator) as a **`Proxy`** that loads each board's shard lazily on property access — `Object.entries(HOLE_PLACEMENTS)` still enumerates all boards, but the mobile hot path never forces the others.

`src/hole-placements.ts` carries a `/// <reference types="node" />` so the global `require`/`process` it uses are typed in every consuming package (board-constants sets no `types` field, and shared packages that import it — play-view, climb-filters — compile without Node's lib). It is a type-only directive, runtime-inert in Metro/webpack.

### How the win was verified

- **Bundle proof.** Inspected the Metro Android bundle: all 6 shards present as distinct modules; the loader's `require(_dependencyMap[N])` calls live inside per-board arrow functions (deferred), not at top level.
- **Runtime proof.** `src/__tests__/hole-placements-lazy.test.ts` spies on Node's CJS `Module._load` and asserts that `getBoardHolePlacements('kilter')` loads only `kilter.cjs` — `tension`/`decoy`/`touchstone`/`grasshopper`/`soill` shards stay unevaluated.
- **Data integrity.** The iOS Swift board-placement generator (`--check`) is byte-identical, proving the reshard preserved the data exactly.

## Release Notes

- Boards open without a hitch on Android — selecting a board no longer briefly freezes the app.

## Test plan

- `vp run typecheck:web` and `vp run typecheck:mobile` — pass (full-repo typecheck green)
- `vp run test:mobile` — pass (2530 tests); board-constants `vp test run` — pass (75 tests, incl. the new lazy-deferral proof)
- `vp check` — changed files clean; shards Oxfmt-stable
- aurora-sync (40), web json-import (96), backend user-data-export-format (3), db create-climb-filters under `tsx --test` (42) — all pass

https://claude.ai/code/session_01Ny9KuZRymZJ7g2XF92mgxk


marcodejongh:
Thanks for the review. Addressed all three:

1. **PR description mismatch** — fixed; the body now describes the lazy-hole-placements change (the earlier paste was wrong).
2. **Non-deterministic timestamp** — removed the `Generated at:` timestamp from the generated loader (`hole-placements.ts`). It's a structural file with no board data, so regenerating with unchanged data now produces a byte-identical file (verified: md5 stable across two regenerations). The data files (`product-sizes-data.ts`, `led-placements-data.ts`) keep their timestamps as before.
3. **`Module._load` private API** — rewrote `hole-placements-lazy.test.ts` to assert against Node's documented `require.cache` instead of the internal `Module._load`. The deferral proof now reads which shard modules have actually been evaluated (`require.cache` entries), and a new case confirms loading one board never pulls in another.

marcodejongh:
Round 2 addressed:

1. **`getOwnPropertyDescriptor` Proxy invariant** — the trap now guards on `ALL_BOARD_NAMES` exactly like `get`/`has`, returning `undefined` for unknown keys and a real `{ enumerable, configurable, value }` descriptor (the board's placements) for known ones. Added tests: trap consistency across get/has/descriptor, and a `structuredClone` walk that would throw on a phantom descriptor.

2. **`nodeRequire()` / Node-ESM branch coverage** — added a `Node ESM shard resolution` suite that drives the exact mechanism that branch uses (`process.getBuiltinModule('node:module').createRequire(import.meta.url)` resolving a shard by its relative path). Vitest always provides a global `require`, so the loader itself can't take that branch under test, but this proves the resolution logic it depends on works. (It's also exercised end-to-end in CI by the iOS Board Placement Data check and the db `tsx --test` suite, both of which run bare Node ESM.)

3. **Node 22+ requirement** — the repo already pins `engines.node: 22.x` (root `package.json`), and `process.getBuiltinModule` is Node 22.3+, so every runtime that can reach this branch satisfies it. Clarified the thrown error to name the requirement and the engines pin instead of a bare message.

Also dropped the `Generated at:` timestamp from the loader so regeneration with unchanged data is byte-identical.

marcodejongh:
Applied the non-blocking suggestion too: `EMPTY_BOARD` (the singleton returned for moonboard / unknown boards) is now `Object.freeze({})`, so a caller can't mutate the shared empty record.

PR #3101, opened 2026-06-20
feat(mobile): mark native releases in the changelog + add check-for-updates

marcodejongh:
## Summary

Two changelog improvements for the mobile "What's New" screen, building on the now-working self-hosted OTA pipeline.

**Native-release markers.** The changelog generator (`scripts/generate-changelog.ts`) now reads the `fingerprint-<platform>-<hash>` git tags the native build workflows push after a successful store upload, and emits a `nativeReleases` array alongside `entries`. Both `CHANGELOG.md` and the in-app feed weave these in as a dated **App update** divider, so the boundary between OTA-delivered and store-delivered changes is visible. The app filters markers to the running platform (a device only cares about store builds that changed its own binary). The OTA publish workflow checks out with `fetch-tags: true` so the tags are present when the generator runs.

**Check for updates.** The changelog screen gains a **Check for updates** button (shown only when `Updates.isEnabled`, i.e. production OTA builds) that pulls an OTA on demand: `checkForUpdateAsync` → `fetchUpdateAsync` → "restart now?" → `reloadAsync`, mirroring the existing `BranchSwitcherScreen` flow.

The generated `changelog.generated.json` / `CHANGELOG.md` are untouched here — they're owned solely by the OTA workflow (`changelog-owned` guard), which regenerates them with the new `nativeReleases` field on the next run. The app defaults `nativeReleases` to `[]` so the pre-feature snapshot still loads.

### Known limitation

A fingerprint tag is pushed only after the native build + store upload completes (~60 min for iOS), so a marker appears on the _next_ changelog regeneration, not the same push that triggered the native build. Acceptable for a changelog (eventually consistent).

## Release Notes

- See when a new app version landed — What's New now flags store updates, so you can tell what arrived over the air from what needs an app update.
- Tap "Check for updates" in What's New to grab the latest fixes on the spot.

## Test plan

- `vp test run --reporter=agent changelog-transform` — 37 pass (new `buildNativeReleases`, native-release markdown, idempotency coverage).
- `vp run typecheck:mobile`, `vp run typecheck` (all packages), plus a direct `tsc --noEmit` on the changed `scripts/` files — clean.
- `vp run test:mobile` — 2530 pass.
- `vp test run catalog-completeness` — i18n parity across en-US / es / fr holds (28 pass).
- `vp check` on all changed files — formatting + lint clean.
- End-to-end render harness against the repo's real `fingerprint-*` tags: markers build correctly (per-platform fingerprints, newest-first) and `CHANGELOG.md` renders the dated `### App update` sections with the right store label.

https://claude.ai/code/session_01Fd63EEpwVrpfn41N5SbQRz

### DIFF
--- a/.github/workflows/mobile-ota-production.yml
+++ b/.github/workflows/mobile-ota-production.yml
@@ -113,6 +113,11 @@ jobs:
         with:
           # Full history so the changelog push-back can rebase onto the latest main.
           fetch-depth: 0
+          # The changelog generator reads the `fingerprint-<platform>-<hash>` tags
+          # (pushed by the native build workflows) to weave native-release markers
+          # into the changelog. fetch-depth: 0 already pulls tags, but be explicit
+          # so a future shallow tweak can't silently drop them.
+          fetch-tags: true
           # Persist the app token (when configured) so the push-back authenticates
           # as the app and bypasses the branch's PR rule. Falls back to the default
           # token (read-only checkout) when the app isn't set up.
--- a/docs/mobile-ota-updates.md
+++ b/docs/mobile-ota-updates.md
@@ -188,9 +188,24 @@ Run `vp run mobile:ota-setup` with no argument for the ordered runbook.
 `## Release Notes` sections, commits it, **pushes it back to `main`** (commit tagged `[skip ci]` so
 the push can't re-trigger the OTA), and only then runs `eoas publish` (which needs a clean tree).
 Nothing else writes the file: the native build workflows and `refresh-acknowledgements.yml` only
-*read* it, and a CI guard (`changelog-owned` in `ci.yml`) fails any PR that edits it. The OTA still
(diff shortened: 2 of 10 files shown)


marcodejongh:
Addressed the review in be245c8:
- **Mobile changelog tests** — extracted the platform-filtering + interleaving into an RN-free `changelog-timeline.ts` (the `changelog.ts` shell just wires in `Platform.OS` + the static JSON) and added `src/lib/__tests__/changelog.test.ts` covering the `Platform.OS` marker filter (iOS doesn't get Android markers and vice versa), both-platform markers, and newest-first interleaving.
- **Type drift** — `NativeRelease.fingerprints` is now required on the mobile side too, matching the generator's type (it always writes the field).

Local: `vp run test:mobile` (2535 pass), `vp run typecheck:mobile`, `vp check` all green.

PR #3103, opened 2026-06-20
fix(backend): require activeBoardName and activeAngle together for playlist climbs

marcodejongh:
## Summary

Follow-up to #3090 (the #1596 grade-angle fix). That PR added optional `activeBoardName` / `activeAngle` to the all-boards `playlistClimbs` input. The resolver only applies the active-angle override when **both** are set, so a request with one but not the other silently no-ops. This adds a zod `refine` so the API contract requires them together (both-or-neither) and rejects the half-specified case instead of quietly ignoring it.

Surfaced by the automated review on #3090; flagged non-blocking, and the commit landed seconds after that PR was merged, so it missed the train.

- `GetPlaylistClimbsInputSchema` gains a `refine` requiring `activeBoardName` and `activeAngle` to both be present or both absent.
- Resolver test asserts the half-specified request is rejected.
- Our clients (mobile all-boards list) always send both or neither, so no client change is needed.

## Release Notes

## Test plan

- `vp run typecheck:backend` — green
- `vp test run --project backend playlist-queries` — 10 passed (incl. the new rejection case)

https://claude.ai/code/session_0157PsyHaLfWs8aXxN1T3No9

### DIFF
--- a/packages/backend/src/__tests__/playlist-queries.test.ts
+++ b/packages/backend/src/__tests__/playlist-queries.test.ts
@@ -317,6 +317,14 @@ describe('playlistClimbs resolver', () => {
     expect(byUuid['climb-tension'].angle).toBe(25);
   });
 
+  it('rejects activeBoardName without activeAngle (they must be provided together)', async () => {
+    const ctx = makeCtx();
+    // Validation runs before any DB access, so no select mocks are needed.
+    await expect(
+      playlistQueries.playlistClimbs(null, { input: { playlistId: 'test-pl', activeBoardName: 'kilter' } }, ctx),
+    ).rejects.toThrow();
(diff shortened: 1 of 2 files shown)


marcodejongh:
Addressed the test nits in `6f83a75ff`:
- Now asserts the specific refine message (`/must be provided together/`) rather than bare `.rejects.toThrow()`, so an unrelated DB/resolver error can't make it pass.
- Added the symmetric case (`activeAngle` without `activeBoardName`) so both directions of the refine are exercised.

Left the refine `path: ['activeAngle']` as-is — it's cosmetic (only affects which field a zod issue is attributed to; our API surfaces the message string, which reads symmetrically).

PR #3108, opened 2026-06-20
fix(mobile): patch gorhom backdrop so a closed sheet never freezes Android scroll + bars

marcodejongh:
## What & why

Several Android 16 users (Samsung One UI, and likely Pixel 10) reported the 2.0 app going **completely dead** after a couple of minutes — the climb list won't scroll, the top and bottom bars stop responding, but you can still tap the climbs you can already see. It only comes back if you background the app and reopen it. iOS and older Android (Pixel 8) are fine.

Root cause is a race inside **@gorhom/bottom-sheet 5.2.14's `BottomSheetBackdrop`**. The backdrop decides whether it's touchable with a `pointerEvents` flip that's gated behind an `isMounted` ref — seeded `false`, then armed `true` in a mount effect. The `useAnimatedReaction` that drives the flip fires **once on mount** and then **only on a change**. When that first `handleContainerTouchability(true)` lands *before* the mount effect arms the ref, the flip to `pointerEvents:'none'` is silently dropped — and since the reaction never re-fires for an unchanged value, the closed backdrop stays `pointerEvents:'auto'` forever. It's an invisible full-screen overlay eating every touch underneath.

Android 16 / One UI frame scheduling loses this race; Pixel 8 wins it. It's an OS/skin timing difference, **not** raw device speed. Background/resume forces a re-render, which is why the freeze "fixes itself" on resume — matches upstream [gorhom #2680](https://github.com/gorhom/react-native-bottom-sheet/issues/2680).

Because we mount persistent sheets at the root (`PlayDrawer`, `QueueSheet`, `BoardSheet`) under one shared `BottomSheetModalProvider` — wrapping both the climbs `Stack` and the bottom bar — a single stuck backdrop freezes the **entire** screen, not just one sheet. That's the "both bars + scroll are dead" report.

## The fix

One line, via a bun patch on `@gorhom/bottom-sheet@5.2.14`: seed `isMounted` to **`true`** instead of `false`, so the mount-time flip always lands. The ref's only real job is guarding `setState` after **unmount**, and the cleanup still flips it `false` — so that protection is untouched. The patch edits gorhom's `src/` (what Metro actually bundles).

## Delivery

JS-level only — no native change — so it rides **OTA** to users already on the current Android build, no new APK required.

## Validation

- `vp run test:mobile` — 2536 tests pass
- `vp check` — clean (the reported format issues are all pre-existing files this PR doesn't touch)
- `vp run check:mobile-bundle` — android bundle OK (11.0 MB), patched gorhom bundles fine

## Release Notes

Fixed an Android freeze where the app could stop responding after a couple of minutes — the climb list wouldn't scroll and both bars went dead until you reopened the app. It stays live now.

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -771,6 +771,7 @@
   },
   "patchedDependencies": {
     "react-native-screens@4.25.2": "patches/react-native-screens@4.25.2.patch",
+    "@gorhom/bottom-sheet@5.2.14": "patches/@gorhom%2Fbottom-sheet@5.2.14.patch",
     "expo-dev-launcher@56.0.20": "patches/expo-dev-launcher@56.0.20.patch",
   },
   "overrides": {
--- a/package.json
+++ b/package.json
@@ -59,6 +59,7 @@
   "packageManager": "bun@1.3.14",
   "patchedDependencies": {
     "react-native-screens@4.25.2": "patches/react-native-screens@4.25.2.patch",
-    "expo-dev-launcher@56.0.20": "patches/expo-dev-launcher@56.0.20.patch"
+    "expo-dev-launcher@56.0.20": "patches/expo-dev-launcher@56.0.20.patch",
+    "@gorhom/bottom-sheet@5.2.14": "patches/@gorhom%2Fbottom-sheet@5.2.14.patch"
   }
 }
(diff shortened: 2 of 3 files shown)


marcodejongh:
**OTA-compatibility verified.** Resolved the Android runtimeVersion (fingerprint policy) with and without this change:

- `main` (no patch): `e3285fb8eb51f5dab08e3c6a0f052d2ae41b10d6`
- this branch (gorhom patch + `patchedDependencies` entry): `e3285fb8eb51f5dab08e3c6a0f052d2ae41b10d6`

Identical — the bun patch on JS-only `@gorhom/bottom-sheet` doesn't appear in the fingerprint sources (no `patches/` / `package.json` / `bun.lock` file-source, and gorhom isn't in the native-dep set). So this lands **over the air** on the current Android build line; no new APK required. On merge to `main`, `mobile-ota-production.yml` auto-publishes the production OTA.

PR #3111, opened 2026-06-20
fix(mobile): keep board history fresh after reconnect, backgrounding & pull-to-refresh

marcodejongh:
## What & why

While climbing a shared board with friends, the new mobile board-history feature felt slow to update — the hunch was that messages get dropped. They do, and the architecture explains it:

- Live "now on the wall" events ride **Redis pub/sub, which has no replay**. Any climb pushed while a phone's WebSocket is mid-reconnect (app backgrounded, network blip, the WS-1006 churn) is gone for that client.
- The durable Redis history (`fetchRecentClimbs`) *does* persist, but today it's only re-read when a **later** live event reveals a sequence gap (`use-board-presence.ts`). So a dropped event with no successor — the last send in a burst, or anything missed while backgrounded — stays invisible until the board sheet is fully reopened.

There was no catch-up on reconnect, no refetch on app-foreground, and no pull-to-refresh. The queue provider already solved this class of bug with a `wsClient.on('connected')` re-sync; board-presence just never got the same treatment.

## Changes

Trigger the existing `runCatchUp` (re-fetch recent climbs + stats + connection from the durable Redis source, deduped by seq) on three more signals, all funneled through its existing coalescer so they can't stampede the backend:

- **WS reconnect** — new optional `BoardPresenceClient.onReconnect`, implemented on mobile via graphql-ws `on('connected')` (skips the first connect). Web omits it (optional).
- **App foreground** — `AppState` `active` in `MobileBoardPresenceProvider`, for the case where iOS suspended the socket without a clean reconnect.
- **Pull-to-refresh** on the history list — a manual escape hatch.

Plus lightweight telemetry so the drop rate is measurable in PostHog: a `Board History Catch Up` event (`{ reason: gap | reconnect | foreground | manual, recoveredThroughSeqDelta }`) and `seq` added to `Board Now Playing Received`. The shared package stays analytics-free — the callback is injected.

No backend, schema, or DB changes. Web behaviour is unchanged.

## Tests

- `use-board-presence` (shared): reconnect → catch-up, `refresh()`/foreground → catch-up, `onCatchUp` reason + recovered-delta, cold-start delta = 0, listener unsubscribe, and the no-`onReconnect` (web) path.
- `MobileBoardPresenceProvider`: foreground triggers refresh, background does not, catch-up telemetry fires with boardId/reason/delta.
- `BoardSheet`: pull-to-refresh triggers a manual catch-up; `seq` on the now-playing event.
- Full mobile suite (2540), shared suites, `vp check`, `vp run typecheck:mobile`, `vp run check:mobile-bundle` all green.

## How to verify

On a sim/device against a local backend (point `EXPO_PUBLIC_BACKEND_URL`/`WS_URL` at it so you can drive `reportBoardClimb`):
- **Foreground:** open the board sheet, background the app, push a climb from another client, foreground → the climb appears within a second or two (was: stayed stale).
- **Reconnect:** toggle airplane mode briefly while a climb is pushed, restore → catch-up pulls it in without needing a *later* climb.
- **Pull-to-refresh:** swipe down on the history list → refetches.

After release, trend `Board History Catch Up` by `reason` and watch `recoveredThroughSeqDelta` to quantify how many silent drops are being recovered.

## Release Notes

The wall history keeps up with your crew now — sends from friends show up right away, even after your phone's been in your pocket or your signal dropped for a moment. Pull down on the history list any time to refresh it.

### DIFF
--- a/packages/mobile/src/components/board-presence/BoardSheet.tsx
+++ b/packages/mobile/src/components/board-presence/BoardSheet.tsx
@@ -237,6 +237,12 @@ export const BoardSheet = forwardRef<BoardSheetHandle, BoardSheetProps>(function
   historyCountRef.current = visibleHistory.length;
 
   const lastReceivedWallClimbRef = useRef<string | null>(null);
+  // `seq` rides along in the telemetry payload but must NOT trigger the effect —
+  // a same-uuid seq bump (e.g. a backfill merge) shouldn't re-evaluate the
+  // "new climb on the wall" event. Read it from a ref so the effect keys only on
+  // the climb uuid.
+  const currentClimbSeqRef = useRef(currentClimb?.seq);
+  currentClimbSeqRef.current = currentClimb?.seq;
   const actionClimbCacheRef = useRef(new Map<string, Climb>());
   const actionClimbRequestRef = useRef(new Map<string, Promise<Climb | null>>());
   const pendingActionKeysRef = useRef(new Set<string>());
@@ -251,9 +257,9 @@ export const BoardSheet = forwardRef<BoardSheetHandle, BoardSheetProps>(function
       boardId: boardPresenceBoardIdRef.current ?? undefined,
       climbUuid: currentClimbUuid,
       // `seq` lets PostHog spot gaps in the live stream (a jump = dropped pushes).
-      seq: currentClimb?.seq ?? undefined,
+      seq: currentClimbSeqRef.current ?? undefined,
     });
-  }, [currentClimb?.climbUuid, currentClimb?.seq]);
+  }, [currentClimb?.climbUuid]);
 
   const snapPoints = useMemo(() => ['55%', '92%'], []);
   const rowBoard = useMemo<BoardSheetRowBoard | null>(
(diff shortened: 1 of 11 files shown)


marcodejongh:
Thanks for the review. Addressed the two actionable notes in 5438c84:

- **`BoardNowPlayingReceived` seq dep** — `seq` now rides in the payload via a ref, and the effect keys only on `currentClimb?.climbUuid`, so a same-uuid seq bump can't re-fire the event.
- **Mount telemetry guard** — added `expect(onCatchUp).not.toHaveBeenCalled()` after `mountSeeded`, so a future refactor that routes the initial backfill through `runCatchUp` would fail loudly instead of silently emitting events.

Leaving the **coalesced catch-up reason** as last-writer-wins on purpose: coalescing only happens when a trigger fires mid-flight (rare), the primary signal is `recoveredThroughSeqDelta` not the reason, and the behaviour is documented inline. Happy to record all coalesced reasons if you'd rather have the fidelity.

The 800ms refresh spinner is a deliberate trade-off for the fire-and-forget `refresh()` API (noted in the PR body); can wire it to the actual fetch if we make `refresh()` awaitable in a follow-up.

PR #3114, opened 2026-06-20
chore(mobile): Android-16 climb-list freeze — on-device bisection toggles

marcodejongh:
## What & why

The Android-16 / Pixel-10 **climb-list touch freeze** (vertical scroll + taps dead, only horizontal swipe works, split-screen momentarily unfreezes it) **survives both `ActiveContextBar` fixes** — #3060 (split the positioning View out) and #3104 (drop the `entering` animation on Android). So the queue bar's Reanimated animation is **not** the cause. It also doesn't reproduce on a Pixel 9, so those two fixes shipped unverified and both missed.

Rather than guess a third time, this adds a **tester/dev-only diagnostic panel** so an affected device can **bisect which suspect is actually blocking the list**. Not a fix — a probe. Do not merge to production.

## Delivery (uses the #3068 tester OTA channel switcher)

Gated on the admin-granted **`tester`** role (or a dev build) — the same audience as the channel switcher — so a tester on a **store/TestFlight build** sees it after switching channels. Delivery is the "missing upload": the EAS preview publish doesn't reach the self-hosted server, so publish this branch's JS to a self-hosted **`preview-1`** channel (Android only — the bug is Android):

```bash
EXPO_UPDATES_URL=https://ota.boardsesh.com/manifest \
GOOGLE_MAPS_API_KEY=<android key> \
EXPO_TOKEN=<expo token> \
  vp run mobile:publish -- --channel preview-1 --platform android --message "freeze debug"
```

(Android needs `GOOGLE_MAPS_API_KEY` so the published fingerprint matches the Android store build; mirrors `mobile-ota-production.yml`.)

## How to test (@alejandrosanchezcabana — Android 16, build 2000363)

1. **Grant Alex the `tester` role** (web admin → role management).
2. Alex: **More → OTA channel switcher → `preview-1`** (or the Custom Channel field) → confirm → app pulls the diagnostic update + restarts.
3. **More → "Climb-list freeze debug"** (now visible) → flip **one** toggle.
4. Go to **Climbs**, try to scroll/tap. Flip it back, try the next.
5. Report **which single toggle brings scrolling back** (or "none").

### What each toggle isolates

| Toggle | If this one fixes it, the cause is… |
|---|---|
| **Hide queue bar** | the `ActiveContextBar` / persistent queue bar after all (despite #3060 + #3104) |
| **Use FlatList (not FlashList)** | a **FlashList v2** scroll/measure regression on Android 16 |
| **Disable row swipe** | the per-row `ReanimatedSwipeable` horizontal pan **stealing the vertical scroll** |
| **Unmount bottom sheets** | an always-mounted gorhom Queue/Board sheet container swallowing touches |
| **Hide top chrome** | the absolute Material top chrome overlay / its measured padding |

(If **none** help, the cause is likely Android-16 edge-to-edge window metrics — `app/_layout.tsx` defers `react-native-edge-to-edge`'s `<SystemBars>` — and we'll instrument that next.)

## Safety

All flags **default OFF**, the panel is gated on `profile.isTester || __DEV__`, and this branch ships only to the diagnostic channel (never `production`). So a regular user never sees it. Store mirrors `session-recording-preference.ts`.

## Validation

`vp run typecheck:mobile` ✓ · `vp check` ✓ · `vp run test:mobile` ✓ · `vp run check:mobile-bundle` ✓

## Release Notes

- [x] No release note needed (internal diagnostic build — not user-facing, not for merge)

### DIFF
--- a/.github/workflows/mobile-eas-update.yml
+++ b/.github/workflows/mobile-eas-update.yml
@@ -32,12 +32,17 @@ concurrency:
 jobs:
   publish-update:
     runs-on: ubuntu-latest
-    timeout-minutes: 15
+    timeout-minutes: 25
 
     env:
       EXPO_TOKEN: ${{ secrets.EXPO_TOKEN }}
       INPUT_BRANCH: ${{ github.event.inputs.branch }}
       INPUT_PLATFORM: ${{ github.event.inputs.platform || 'all' }}
+      # NOTE: EXPO_UPDATES_URL is deliberately NOT set job-wide. app.config's
+      # resolveUpdatesConfig flips `updates.url` to the self-hosted server whenever
+      # EXPO_UPDATES_URL is present (+ the committed cert), which would make the EAS
+      # `eas update` step below fail ("expected the EAS url"). So it's set per-step
+      # ONLY on the self-hosted publish steps; the EAS step runs without it.
       # Build-time env inlined into the OTA JS bundle, matching the native build
       # workflows (ios-testflight-rn.yml / android-apk-rn.yml). Without these, an OTA
       # update ships a bundle that re-resolves EXPO_PUBLIC_* to its in-code fallbacks
@@ -55,6 +60,8 @@ jobs:
     steps:
       - name: Checkout repository
         uses: actions/checkout@v4
+
+      - uses: voidzero-dev/setup-vp@v1
 
       - uses: oven-sh/setup-bun@v2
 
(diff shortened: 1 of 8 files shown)


marcodejongh:
@alejandrosanchezcabana — when you get a sec, could you give this a run on your Pixel 10 (Android 16)? You showed up in our telemetry on build 2000363 still hitting the climb-list freeze after both fixes, so you're the ideal person to pin it. Steps are in the PR description: **More → "Climb-list freeze debug"**, flip one switch at a time, then check if the **Climbs** list scrolls. Just tell us which switch (if any) brings scrolling back and we'll have the culprit. 🙏

marcodejongh:
Addressed the review note in 6c1cb320f: `loadFreezeDebugFlags` now short-circuits when `!isPreviewBuild() && !__DEV__`, so production store builds skip the AsyncStorage read entirely (no cold-start cost). Also fixed the PR title type (`diag`→`chore`) and the Release Notes opt-out.

PR #3117, opened 2026-06-20
feat: import beta videos from an Instagram scan

azuttre:
## What

A durable **Import beta from Instagram** feature: point it at an Instagram account and it surfaces the climbing-beta videos that aren't on Boardsesh yet, so you can attach them in a couple of clicks.

## Why client-side

Server-side Instagram scraping isn't viable — Instagram blocks datacenter fetches, it's against their ToS, and in testing the prod backend got *region-locked* fetching public posts. So the scan runs in the user's own browser, same-origin on instagram.com, using their existing session. Boardsesh's servers never touch Instagram.

## Flow

1. **Scan** — on an Instagram profile, paste the provided console script. It paginates Instagram's `v1/feed` API and copies a compact `{shortcode, caption, takenAt}[]` array to the clipboard.
2. **Match** — paste that into `/import-beta`. The new `instagramBetaScan` GraphQL query parses each caption (`"Climb" @ 40° on the Kilter Board`), resolves it to a climb, and diffs against `board_beta_links`.
3. **Attach** — review the buckets (missing / already on Boardsesh / ambiguous / unmatched) and attach the missing ones via the existing `attachBetaLink` mutation (which already handles enrichment, thumbnails, dedup, and attribution).

## Notes for review

- Scoped to **kilter / tension** — the boards whose share captions use the quoted `"name" @ angle°` format the parser understands (MoonBoard's comma format is rejected by validation rather than silently zero-matching).
- Shortcode dedup uses the same `getInstagramMediaId` as `attachBetaLink`, so the scan's "missing" set matches what the mutation will actually insert.
- Ambiguous names (one name → several climbs) get a per-row climb picker; deleted/private posts fail individually without aborting the batch.
- Bulk attach paces itself to the per-user rate limit (honors the backend's `RATE_LIMITED` `retryAfterSeconds`) so a large import finishes cleanly instead of dropping videos.
- The caption parser keeps apostrophes inside names (`"Carlo's Blowtorch"`) — the quote-delimiter set is double-quote types only.
- `/import-beta` is `noindex` (utility page).
- The `instagramBetaScan` resolver is auth-gated + rate-limited (30/min) and batches its DB work (no N+1 over posts).

## Validation

- ✅ `vp run typecheck` (full, cross-package)
- ✅ unit tests — caption parser and the pure classifier (incl. apostrophe-name regression)
- ✅ `vp run check:i18n` + i18n catalog parity for en-US / es / fr
- ✅ `vp check` format
- ✅ Scan script validated live against a real 645-post profile (clean pagination, complete signal).

## Follow-ups (not in this PR)

- A browser extension for one-click scanning (vs. the console script).
- Extending beyond kilter/tension if there's demand.

## Release Notes

Bring your Instagram climbing beta into Boardsesh. On your Instagram profile, run the quick scan, paste it into Import Beta, and Boardsesh shows which of your filmed climbs aren't linked yet. Attach the missing ones in a couple of taps and your videos show up as beta on those climbs for the whole crew. Works with Kilter and Tension.


marcodejongh:
## Code review (by Claude 🤖)

> Posted on behalf of @marcojonghd. This review was produced by a team of Claude Code review subagents (backend correctness, security/abuse, frontend + i18n, and conventions/codegen) and the findings below were independently re-verified.

Really nice PR — well-structured and thorough. **No blockers.** Codegen artifacts are clean (ran `vp run codegen` + `git diff --exit-code`, tree stayed clean), no `any`, hand-written TS types match the SDL field-for-field, auth + per-user rate-limiting are real, and there's no ReDoS/SSRF/cross-user leak. i18n parity across en/es/fr is complete with genuine translations.

### ⚠️ Merge conflicts
The branch currently conflicts with `main` (moved to `ccb28b8f4`). Conflicts are in the registration + hand-edited generated files that another merged change also touched:

```
packages/backend/src/graphql/resolvers/index.ts
packages/shared-schema/src/index.ts
packages/shared-schema/src/schema/beta-links.ts
packages/shared-schema/src/schema/queries.ts
packages/shared-schema/src/types/index.ts
packages/shared-schema/src/generated/schema.graphql
packages/shared-schema/src/generated/types.ts
packages/shared/graphql/src/operations/beta-links.ts
packages/shared/graphql/src/generated/gql.ts
packages/shared/graphql/src/generated/graphql.ts
```

Rebase on `main`, take both sides on the index/schema/operations files, then re-run `vp run codegen` to regenerate the four `generated/*` files cleanly (don't hand-merge those — let codegen produce them).

---

### Medium — worth fixing before merge

**1. Caption parser detects the board from the whole caption, including the climb name → real "missing" posts silently dropped.** `packages/shared-schema/src/instagram-caption-parse.ts:26-31`
`"Tension Board" @ 40° on the Kilter Board.` parses as `boardType: tension` (verified). It's a Kilter climb literally *named* "Tension Board", but `detectBoard` checks tension first and scans the entire string. The misrouted board makes the resolver query the wrong `board_climbs` partition, so a genuinely-missing post lands in `unmatched` and is never offered for attach. Same for names like `"My Tensionboard project"`.
**Fix:** detect the board only from the text *after* the matched name's closing quote (the canonical caption is `"name" @ angle° on the X Board`, so that trailing region is authoritative), or require the explicit "on the {Kilter|Tension} Board" phrase rather than bare brand substrings.

**2. Same scoping bug for the angle → wrong angle attached.** `instagram-caption-parse.ts:24,40`
`"Project @ 30" @ 45° on the Kilter Board` parses `angle: 30` (verified) — the `@30` inside the name wins over the real `@45`, and that wrong angle flows through `missing[].angle` into `attachBetaLink`.
**Fix:** run the `ANGLE` regex on the substring after `nameMatch.index + nameMatch[0].length`, not the full caption.

**3. No disclosure of the user's own Instagram-account risk.** `instagram-scan-script.ts` + scan copy in `feed.json`
The script automates Instagram's private `/api/v1/feed` endpoint with `credentials: 'include'` — exactly what IG's anti-automation systems flag. The UI copy reassures users about *Boardsesh's* privacy ("Boardsesh never sees your Instagram login") but says nothing about the risk to *their own* account (rate-limit / challenge / ban). That reads as misleading by omission.
**Fix:** add one honest line to the scan copy in all three locales. (The script itself is safe — same-origin only, no exfiltration, and the `X-IG-App-ID` is the public web id, not a secret.)

**4. Icon-only external links have no accessible name.** `import-beta-content.tsx:444, 474, 621, 753`
The four `<Link><OpenInNewOutlined/></Link>` render icon-only with no `aria-label`, so screen readers announce only "link".
**Fix:** add `aria-label={t('importBeta.openOnInstagram')}` (new key) to each.

### Low — follow-ups / polish

- **Dedup query isn't bounded by the payload** (`instagram-beta-import.ts:95-101`) — pulls *every* `board_beta_links` row for the board(s) seen rather than `… and shortcode in (scannedShortcodes)`. Cost grows with the global table, not the (capped) input; the `(board_type, shortcode)` index would serve the bounded form.
- **`lower(name)` lookup can't use the `(boardType, name)` btree index** (`instagram-beta-import.ts:59`) — per-board scan each call. Bounded by the 2000-post cap + 30/min so not a DoS, but a functional `lower(name)` index would help if this gets hot.
- **`board_climbs.angle` is selected and carried into `ScanClimbRow` but never read** — the classifier always uses the caption angle. Dead projection; drop it.
- **Inner-quote names truncate** (`"Name with "inner" quote"` → `"Name with"`) — harmless (falls to `unmatched`), worth a comment.
- **Duplicate shortcodes in the pasted JSON both land in `missing`** → React duplicate-key warning; attach is idempotent so no DB dup. Dedupe `posts` by shortcode up front.
- **Attach loop has no `AbortController`** (`import-beta-content.tsx:524-576`) — navigating away mid-loop leaves in-flight requests firing `setState` on an unmounted tree. No data loss (double-submit is correctly guarded), but a cancel affordance would be cleaner.
- **Buckets render up to ~2000 unvirtualized rows via `.map`**, and `MissingSection` re-renders the whole bucket on every per-row status update during the 600 ms-spaced loop. Advisory for web, but `React.memo` on the row + splitting status state would cut the churn.
- **Singular grammar / no ICU plurals** — "Attached 1 beta videos", "Attach 1 selected" read wrong at `count === 1` in all three locales.
- **Unused `import React` in `page.tsx:1`** (Next 16 automatic JSX runtime).
- **The pure DB-free classifier could live in a shared package** for future mobile reuse per "extract, don't duplicate" — reasonable to defer (no mobile consumer today).

### Confirmed clean
Codegen no-drift (ran it), no `any`, types ↔ SDL exact, `sql` imported from `drizzle-orm` (not the blocked path) and the `lower()` raw-SQL is justified, resolver registration + schema wiring correct, no ReDoS/SSRF, no cross-user data leak in buckets, console script safe, server-side Zod re-validation is the real gate, re-scan correctly resets state (no stale selection), `track()` leaks no PII, design tokens clean (no raw hex), `LocaleLink` used, `noindex` correct, i18n parity + translation quality complete.

**Overall:** ship-ready after the two parser scoping fixes (#1, #2 — the only real correctness bugs), the trust/safety disclosure (#3), and a rebase to clear the conflicts; the a11y labels (#4) are a quick win. Everything else is follow-up.

---

📨 Separately — I've **invited you as a contributor** on the repo. Once you accept, you'll be able to push branches and open PRs directly here instead of from a fork, which means CI runs automatically on your PRs (and the EAS/OTA preview workflows that need repo secrets will work). Thanks for the great work on this one.

azuttre:
## Validated live against a real account, then hardened the scanner

Ran the feature end-to-end against a real 233-post Kilter homewall account (`shackonekbhw`), read-only — nothing attached:

| | |
|---|---|
| Posts scanned (full account) | **233** |
| Beta captions parsed | 231 (2 non-beta) |
| Missing (not on Boardsesh) | **71** |
| Already on Boardsesh | 141 |
| Ambiguous (name → multiple climbs) | 16 |
| No climb match | 3 |

The scan, caption parser, climb resolution, and shortcode dedup all behaved correctly at scale: 141 already-present videos were correctly deduped, 71 are genuinely new.

### What the live run changed (3ea9936d6)

The first probe returned only 48 posts because it **stopped on the first error** — and the committed scan script had the same flaw. Hardened it:

- **Retry transient failures** (per-fetch timeout + bounded retries) instead of breaking on the first non-OK response. A single blip no longer silently truncates the scan, which would under-report what's missing — the whole point is to find *all* of it.
- **Progress + completeness**: the script logs a running count (a large account takes ~a minute to paginate) and reports whether it COMPLETED or gave up early, so a partial scan isn't mistaken for the full account.
- **Parser regression tests** from the real captions: the `Kilter Board Homewall` variant, names with sizes/accents/apostrophes/punctuation, and a non-beta caption that must return null.
- i18n: note that the scan can take a minute (en/es/fr).

Still pending (unchanged): a manual browser QA pass of `/import-beta`, and CI lint (couldn't run locally — pre-existing oxlint config-load issue in my worktree).

azuttre:
## Addressed the review + merged main

Thanks for the thorough review — really helpful. Pushed:

**Fixed**
- **Parser scoping (#1, #2)** — angle + board are now detected only in the caption text *after* the quoted climb name, so a climb named `"Tension Board"` or `"Project @ 30"` no longer mis-routes the board/angle. Both bug-cases added as tests.
- **Trust/safety (#3)** — the scan copy now discloses that automating Instagram's feed can rate-limit the user's *own* account (en/es/fr).
- **a11y (#4)** — `aria-label` on all four icon-only Instagram links (new `openOnInstagram` key, all three locales).
- **Quick wins** — dedupe pasted posts by shortcode (no duplicate React keys); dropped the unused `React` import in `page.tsx`.

**Deferred to follow-ups** (glad to fold into this PR if you'd rather): ICU plurals, row virtualization, attach `AbortController`, and dropping the dead `angle` projection.

On the **dedup-query** low item: bounding by the `shortcode` column would miss Aurora-imported rows (their `shortcode` is null — dedup extracts the id from the link), so I kept the link-based extraction. Happy to add a `link`-pattern bound if you want it tightened.

**Merge** — merged `main`; resolved the schema/operations/resolver-wiring conflicts (kept your `userBetaLinks(offset)` / `betaLinkPreview` / `integrations` and re-applied `instagramBetaScan` on top), then re-ran `vp run codegen` so the generated files are regenerated, not hand-merged.

Leaving the merge button to the team — letting CI run first.

marcodejongh (approved):

marcodejongh:
Conflicts 😢

azuttre:
Conflicts cleared up on their own. A test-merge against current main came back clean (main moved on since your note), and since it was already approved I squash-merged it as c11e62cf. Thanks for the review!

PR #3119, opened 2026-06-21
fix(mobile): stop counting OAuth cancels & recoverable throws as login failures

marcodejongh:
## Summary

The React Native **"Login Failed"** PostHog metric spiked to ~200/day starting 2026-06-14, but **no user reported a login problem**. Investigation (PostHog project 412845 + the code) found the metric counts **events, not users**, and a single broken-or-cancelled iOS OAuth tap fires **2–3** `Login Failed` events:

- In the 06-14 week, **246 unique users attempted, 189 (77%) succeeded**, yet only **35 unique users** produced the **~339 `exception` events** (~10 retries each).
- Inflators: (1) iOS 26.5.1 native Google/Apple OAuth throws on app 2.0.0 — each logs a `Login Failed (exception)`; (2) the iOS browser fallback meant to rescue them has a **0% success rate** (0/124), logging a second `Login Failed (cancel)`; (3) user cancellations (dismissing the sheet/browser) are logged as `Login Failed (cancel)`, ~157/week.

This PR makes the metric honest **without changing login behaviour**:

- **Cancels are no longer failures.** Dismissing the provider sheet or the browser fallback now fires a distinct **`Login Cancelled`** event (`@boardsesh/analytics` `SHARED_EVENTS`), not `Login Failed (cancel)`.
- **Recoverable native throws are tagged.** iOS native OAuth failures/throws (which the browser fallback handles next) carry `recoverable: true`; on Android they stay terminal (`recoverable: false`). The failure metric can then exclude `recoverable = true`.

The PostHog reliability insights (`[Reliability] Login outcomes per day` `1LQgqVWM` and its sibling `1vWdXf5k` on dashboard `1693744`) will be rebuilt to a unique-user funnel that separates terminal failures from cancels/recoverable throws — done in PostHog, no code dependency.

**Out of scope (tracked separately):** the real iOS 26.5.1 native OAuth break and the 0%-success browser fallback (`auth.ts` `signInWithProviderWeb`) — that's the genuine user-facing degradation. Expected user errors (`invalid_credentials`, registration 409) still emit `Login Failed`; they can be excluded at the insight level later.

## Release Notes

## Test plan

- `vp run test:mobile` — 2552 tests pass; `use-native-oauth-sign-in.test.tsx` updated to assert `Login Cancelled` on sheet/browser dismissal and `recoverable: true/false` on native throws (iOS/Android)
- `vp check` — lint/format clean on changed files
- Ships JS-only via the production OTA pipeline on merge to `main`; the new `Login Cancelled` / `recoverable` signals appear once the OTA lands.

https://claude.ai/code/session_01DCUotBVjLgFyBZkBhYeYGF

### DIFF
--- a/packages/mobile/src/hooks/__tests__/use-native-oauth-sign-in.test.tsx
+++ b/packages/mobile/src/hooks/__tests__/use-native-oauth-sign-in.test.tsx
@@ -186,13 +186,21 @@ describe('useNativeOAuthSignIn — Apple web fallback (iOS)', () => {
     expect(trackedEvents()).toContainEqual({ event: 'Login Succeeded', flow: 'web_fallback', reason: undefined });
   });
 
-  it('does NOT fall back when the user cancels native Apple', async () => {
+  it('does NOT fall back when the user cancels native Apple, and logs a cancel (not a failure)', async () => {
     signInWithAppleMock.mockResolvedValue({ success: false, cancelled: true });
 
     await runSignIn('apple');
 
     expect(signInWithAppleWebMock).not.toHaveBeenCalled();
     expect(reportErrorMock).not.toHaveBeenCalled();
+    const events = trackedEvents();
+    expect(events).toContainEqual({
+      event: 'Login Cancelled',
+      flow: 'native',
+      reason: undefined,
+      recoverable: undefined,
+    });
+    expect(events).not.toContainEqual(expect.objectContaining({ event: 'Login Failed' }));
   });
 
   it('stays silent (no error, no report) when the browser Apple sheet is cancelled', async () => {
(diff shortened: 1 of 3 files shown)


marcodejongh:
Thanks for the review — addressed:

**1. Apple native cancel test gap** — fixed in `268e7211d`. The Apple `does NOT fall back when the user cancels native Apple` test now asserts `Login Cancelled` is emitted and `Login Failed` is not, matching the Google native-cancel tests.

**2. Type enforcement on `recoverable`** — leaving as-is, by convention. The analytics layer deliberately keeps props open: `track(name: string, properties?: AnalyticsEventProperties)` where `AnalyticsEventProperties = Record<string, AnalyticsPropertyValue | undefined>`, and the `events.ts` header documents that the catalog only types event *names*, not payloads. No event in the codebase (`auth_method`, `flow`, `failure_reason`, …) has a typed property interface today, so typing just `recoverable` would be inconsistent. A typed per-event property system is a worthwhile follow-up but out of scope for this telemetry-semantics fix.

Issue #3098, opened 2026-06-20
Verify self-hosted OTA delivery; JS-only fixes currently ship OTA-only (native build gated off)


marcodejongh:
## Context

PR #3097 (the Android board-select freeze fix) was **JS-only**. On merge to `main`, the `android-apk-rn.yml` fingerprint `gate` job resolved the **same** Android fingerprint as build `358`, so `build-and-release` was **skipped by design** — "the native build is skipped; the OTA delivers the JS" (workflow ~line 154). No new native build/tag was produced (latest stayed `rn-android-2.0.358`). The fix shipped **only** via the "Mobile OTA Production (Self-Hosted)" publish.

That is correct, efficient behavior — *if* the self-hosted OTA actually delivers. We are **not yet confident it does** (self-hosted OTA activation was still being validated). If OTA isn't functioning, a critical JS-only fix merged to `main` reaches **no users**: no native rebuild happens and the OTA is inert. For this incident we force-built a native APK via `workflow_dispatch` (which builds regardless of fingerprint, ~line 148) as a safety net.

## To investigate

1. **Verify end-to-end OTA delivery.** On a real Android device on build `356`/`358`: does it poll the self-hosted `EXPO_UPDATES_URL`, fetch the fingerprint-matched `production` update, and apply the board-freeze fix on next launch — *without* a store update? Confirm the freeze is gone post-OTA.
2. **Fingerprint parity.** Confirm the `gate`'s resolved Android fingerprint matches what `mobile-ota-production.yml` publishes (both run on Linux per the workflow comments) so OTA updates actually match installed binaries. Spot-check the `fingerprint-android-<hash>` tag the gate keys off.
3. **OTA observability.** We currently can't tell how many installs pulled an OTA. Add an "update applied" signal (expo-updates event → PostHog) so rollout of a JS-only fix is measurable.
4. **Force-build policy for critical JS fixes.** Until OTA is trusted, standardise dispatching `android-apk-rn.yml` (force native build) for urgent fixes — e.g. a `force_native` workflow_dispatch input or a label-driven path — so a critical fix is never OTA-only by accident.

## Refs
- PR #3097 (JS-only fix). The merge build run that skipped native: `27864478906` (`gate`=success, `build-and-release`=skipped).
- Forced native build for this incident: run `27864826149`.
[...]

PR #3126, opened 2026-06-21
feat(mobile): OTA adoption telemetry + force-native build escape hatch (#3098)

marcodejongh:
## Summary

Closes the two remaining codeable follow-ups on #3098. Items #1 (on-device OTA-apply verification) and #2 (fingerprint/env parity) are already resolved per the issue's diagnosis comment — this PR addresses **#3 (observability)** and **#4 (force-native policy)**.

A JS-only fix merged to `main` ships OTA-only and skips the native build when its fingerprint already shipped. Today we can't measure whether that rollout reached users, and there's no deliberate lever to force a native rebuild for a binary whose fingerprint has drifted past the OTA (the orphaning that left the #3097 board-freeze fix stranded on older Android installs).

### OTA observability (#3)
- New `OtaUpdateTracker` (mounted once near the app root beside `AnalyticsScreenTracker`) fires:
  - **`OTA Update Status`** once per launch — running bundle's `updateId`, `isEmbeddedLaunch`, `runtimeVersion`, `channel`, `createdAtIso`, emergency flags. `isEmbeddedLaunch === false` ⇒ running an OTA'd bundle; group by `updateId` to size a rollout; `runtimeVersion` is the fingerprint cohort. The cohort is also registered as PostHog **super properties** so any existing funnel can be sliced by OTA-vs-embedded.
  - **`OTA Update Downloaded`** when a newer bundle finishes downloading in-session (applies next launch) — completing the published → downloaded → applied funnel.
- Pure, unit-tested `buildOtaStatusProperties` mapper + free-string event names (mobile-only, not `SHARED_EVENTS`) in `ota-telemetry.ts`; guarded `registerSuperProperties` helper added to `analytics.ts`.

### Force-native escape hatch (#4)
- `android-apk-rn.yml` gains a `force_native` `workflow_dispatch` input (default **on**). On → build regardless of the fingerprint (the documented way to push an urgent JS-only fix to OTA-orphaned binaries); off → fall through to the same fingerprint-tag check as a push. The fingerprint-affecting `env:` block is untouched, so `scripts/mobile-ci-env-parity.test.ts` stays green.

### Docs
- `docs/mobile-ota-updates.md`: new "OTA observability" section + the force-native runbook step.

Out of scope (noted on the issue): #1's physical-device check and the broader "minimise native churn between store builds" process work.

## Release Notes

## Test plan

- `vp run test:mobile` — 2558 passed, incl. the 6 new tests (`ota-telemetry.test.ts`, `ota-update-tracker.test.tsx`).
- `vp check` — my files clean (remaining flags are pre-existing drift on `main`).
- `scripts/mobile-ci-env-parity.test.ts` — 16 passed (workflow env block unchanged).
- Workflow YAML parses; `version_code_offset` + `force_native` inputs present.
- Local: in `__DEV__` the analytics debug hook logs `[analytics] OTA Update Status …` to Metro once on startup (no events sent in dev).

### DIFF
--- a/.github/workflows/android-apk-rn.yml
+++ b/.github/workflows/android-apk-rn.yml
@@ -24,6 +24,18 @@ on:
         description: 'Added to GH run_number for versionCode (raise to clear collisions / always supersede the Capacitor app)'
         required: false
         default: '2000000'
+      # The canonical "ship an urgent JS-only fix to binaries that can no longer
+      # receive an OTA" lever (issue #3098). A JS fix merged to main is delivered
+      # OTA-only and skips the native build when its fingerprint already shipped —
+      # but a binary whose fingerprint has since drifted is orphaned from the OTA.
+      # Dispatching this workflow with force_native on rebuilds the native app so
+      # those installs get the fix via a store update. Defaults on, so a manual
+      # dispatch keeps building as before; uncheck to honour the fingerprint gate.
+      force_native:
+        description: 'Force a native build even if this fingerprint already shipped (push an urgent JS-only fix to orphaned binaries). Uncheck to honour the fingerprint gate.'
+        type: boolean
+        required: false
+        default: true
 
 concurrency:
   group: android-apk-rn-${{ github.ref }}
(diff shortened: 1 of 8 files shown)


marcodejongh:
Addressed the review notes in `ea1b3dde8`:

- **`analytics.ts`** — `registerSuperProperties` now `console.warn`s in `__DEV__` when the client exposes no `register()`, so an SDK API change surfaces instead of silently dropping super properties (runtime guard unchanged).
- **`OtaUpdateTracker.tsx`** — the launch `OTA Update Status` event is now guarded by a module-scoped key, so it fires at most once per JS runtime (= once per launch) even across a remount. An OTA reload restarts the runtime, clearing the guard, so the next launch re-reports under the new bundle's `updateId`.
- **Tests** — added the no-`createdAt` download path (→ `createdAtIso: null`) and a once-per-launch dedup test across a remount. 8 tests now pass.

One note on the test-environment point: the mobile suite's convention is `@vitest-environment jsdom` + `@testing-library/react` (183 files, **0** use `@testing-library/react-native`); this file mirrors the existing analytics tests (`add-beta-video-sheet-analytics.test.tsx`, `generator-picker-card-analytics.test.tsx`), so I kept it consistent with precedent.

marcodejongh:
Rebased onto current `main` (was 49 behind) and force-pushed — the branch had a real conflict with main's `_layout.tsx` edge-to-edge/Sentry rework, which is why it couldn't merge. Resolved by keeping main's structure (`FreezeDebugOverlay`, `BleControlSheetProvider`) and re-inserting `<OtaUpdateTracker />`; `analytics.ts` auto-merged. Re-validated against the new base: typecheck clean, full mobile suite **2593 passed**, env-parity **22 passed**, bundle OK. All CI green; `mergeable`.

Also folded in the two actionable nits from the last review (guard moved above `buildOtaStatusProperties`; doc code-object on one line). On the recurring `resetOtaStatusReportedForTests` note — leaving it as-is: the `…ForTests` suffix already signals test-only intent, the body just clears one boolean, and it's flagged non-blocking. Happy to `__DEV__`-gate or rename if you'd prefer.

Ready for your review/merge.

Issue #3087, opened 2026-06-20
Racing party member sees a "Sync error" toast when someone ends the shared session


marcodejongh:
## What's happening

When the leader ends a collaborative session while a peer is still live and trying a queue/join operation, the peer's mobile app shows an alarming "Sync error" toast and reports a handled exception to error tracking — for what is normal party-session teardown (someone ended the shared session). The web app handles the identical case silently. Mobile should clear its local session state and treat "session has ended" as expected rather than surfacing it as an error.

## Evidence

PostHog error tracking: `GraphQLOperationError` value "This session has ended" (handled:true), `$lib` posthog-react-native, native source.
- 2 occurrences / 2 sessions / 2 users (low long-tail, not traffic-inflated).
- Session IDs: `019ed68f-fa80-7fde-bfa9-d4b53f560c96`, `019ecafc-384d-7672-9f11-e0b5fc9f7dae`.
- _posthogUrl: https://us.posthog.com/project/412845/error_tracking/019ecb01-5604-7091-96d1-f5edd7b0c636

Platform: mobile (iOS).

## Suspected cause

The backend deliberately throws `GraphQLError('This session has ended', { extensions: { code: 'SESSION_ENDED' } })` on JOIN to an ended session (`packages/backend/src/services/room-manager/client-lifecycle.ts` lines 113-122; documented #2696 behavior). Web handles it correctly and silently: `packages/web/app/components/persistent-session/hooks/use-session-lifecycle.ts` (lines 502-516) checks `err instanceof GraphQLOperationError && err.extensions?.code === 'SESSION_ENDED'`, drops the stored session, returns null, and does NOT report — covered by a dedicated test. Mobile lacks this guard: `packages/mobile/src/providers/queue-provider.tsx` (lines 619-634, `startJoinedSubscriptions`) shows a `mobile.queue.syncError` toast and calls `reportHandledError(joinError, { source: 'queue-sync', op: 'join' })` on the first failure with no `SESSION_ENDED` discrimination. `reportHandledError` (`packages/mobile/src/lib/error-reporting.ts` lines 62-79) filters cancellation/auth/beta/network only, so it reports at level `error`. Grep confirms no client-side `SESSION_ENDED` handling exists in `packages/mobile` or `packages/shared`.

Low volume is because the cold-start restore path is already guarded by the #2683 SESSION_STATUS HTTP precheck — only a session ending while the app is already live reaches this unguarded join+report path.

[...]

Issue #3008, opened 2026-06-18
Persist Redis board-presence history to Postgres so the "now on the wall" feed survives expiry


marcodejongh:
## Background

The board "now on the wall" feed has two stores:

- **Redis FIFO** (`pubsub.getRecentBoardClimbs`, read by the `boardRecentClimbs` query) — the hot cache that backfills late joiners and drives the mobile accessory-bar / board-sheet "On the wall" status. It captures **everything** (including anonymous `conn:` pushes, no dwell gate) but is **ephemeral**: TTL was 24h, just bumped to **1 week** in #3007.
- **Postgres `board_climb_events`** (read by the `boardHistory` query) — durable, but **gated**: only persists pushes from **logged-in** users after a **≥60s dwell** (`DURABLE_DWELL_MS`, see `board-presence/mutations.ts`).

## Problem

Because the durable Postgres store is gated, it can't fully reconstruct the live feed once Redis expires:
- anonymous (`conn:`) pushes are never persisted,
- pushes with <60s dwell are dropped,
- so after the Redis window lapses, `boardRecentClimbs` returns empty and the "On the wall" surfaces go blank for an idle board even though Postgres has *some* history.

Extending the TTL to a week (#3007) widens the window but doesn't remove the cliff.

## Proposal

Flush / mirror the Redis recent-history entries to Postgres so the live feed survives Redis expiry without the current login + dwell gating. Options to weigh:

1. **Read-through fallback**: when `boardRecentClimbs` finds an empty/cold Redis FIFO, backfill from `board_climb_events` (and re-warm Redis). Smallest change; bounded by what's already persisted.
2. **Write-through mirror**: persist every accepted `reportBoardClimb` to a durable "recent feed" table (ungated, or with a lighter gate than `board_climb_events`), so Redis is purely a cache over Postgres.
3. **Background flush job**: periodically drain Redis FIFO entries to Postgres before they expire.

Decide whether to relax/keep the dwell + login gate for the durable copy (anonymous attribution is already null-safe in the schema).

## Notes

- Keep it portable Postgres (no Railway-specific addons) per repo DB-hosting rules.
- Related: #3007 (TTL 24h → 1 week).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #2860, opened 2026-06-14
[RN] Board presence drift


marcodejongh:
In a session it seems like websocket messages are being missed.

Issue #2713, opened 2026-06-10
Board presence: hardening follow-ups (after #2712)


marcodejongh:
Second-stage work from the [#2660 review](https://github.com/boardsesh/boardsesh/pull/2660), to pick up **after #2712 lands**. None of these block turning the feature on, but the first two can silently freeze the "now on the wall" feed, so they should follow shortly after.

## Resilience

1. **Redis seq-counter reset freezes the wall silently** — `packages/shared/board-presence/src/reducer.ts:6-12` documents the monotonic-`seq` assumption but nothing handles a reset (Redis flush/failover restarting `INCR` at 1). Every live event then has `seq <= lastSeq` and is swallowed; the only user escape is switching boards. Fix: send a monotonic epoch alongside `seq`, or dispatch `RESET` on reconnect so `lastSeq` returns to 0.
2. **No resubscribe after a terminal subscription error** — `packages/shared/board-presence-react/src/use-board-presence.ts:128-138` flips `isLive` off permanently until remount/board switch. Expose a `retry()` action or re-run the subscription on an error counter, and document that the injected client is expected to auto-reconnect.

## Ops

3. **Migration 0121 lock window** — two full-table `UPDATE`s + a non-concurrent `CREATE UNIQUE INDEX` on `user_boards`. Confirm the table is small enough that the lock is sub-second, or split the index into a manually-run `CREATE UNIQUE INDEX CONCURRENTLY` step documented in the deploy runbook.

## Cleanups (batch into any of the above)

4. `BoardClimbCleared` is a dead schema-union leg — the client reducer/mapper handle it but no producer exists. Ship the `clearBoardClimb` mutation or remove the member.
5. `board-presence.test.ts` uses raw `db.execute(sql…)` for plain seeds/asserts (repo rule: Drizzle ORM), and its Redis FIFO test re-implements lpush/ltrim by hand instead of exercising `storeBoardClimb`/`getRecentBoardClimbs`.
6. `board-presence/package.json` typescript pin `^5.9.3` → monorepo standard `~6.0.3`.
7. `gradeColor` is always `null` server-side in `reportBoardClimb` — populate from `@boardsesh/board-constants` or drop the field.
8. Unique-violation `logger.warn` in `resolveBoardForSerial` can include the conflicting serial via `String(error)` — log hygiene.
[...]

PR #3129, opened 2026-06-21
fix(web): wall-confirm resolves from convergent state, not a 2s verdict (#3121)

### LINKED ISSUE #3121: Wall-confirm window is too tight — slow-but-valid board confirmations get logged as timeouts
## What's happening

After a send, the bulb pulses while the app waits for the board to confirm the wall lit up. The wait is capped at 2 seconds. The board's real confirm latency runs right up against that ceiling (p95 1.5s, max ~2.0s), so ~5% of legitimate confirmations land just after the cutoff and get recorded as a "Wall Confirm Timeout." When that happens the bulb pulse stops early and the device picker re-pops (which can trip "Already scanning. Stopping now." on iOS) even though the wall actually lit. The board still lights, so people just re-tap — it's a friction/noise bug, not an outage.

## Evidence

PostHog project 412845 — product analytics (30d).

- `Wall Confirm Timeout` fired 4,618 times vs 1,468 `Wall Confirmed` over 30d, across 669 distinct users (471 iOS + 198 Android). Events fire from the web bundle inside the Capacitor shell, so `$os_name` is mostly `None`.
- The headline 3:1 ratio is largely an artifact: ~52% of the 30-day timeouts are in the week of 2026-06-14, and that spike is entirely new `pulse_only` instrumentation (1,996 events, zero before — added by commit 03bc5ddd8). `pulse_only` and `picker`/`auto_connect` timeouts are armed at lightbulb-tap (connect initiation), 2s before a BLE scan/pair can finish — expected, not a board-ack failure. Strip those and the steady-state ratio is near 1:1.
- The real kernel: successful confirm latency over 30d (n=1,469) is p50 258ms, **p90 1,207ms, p95 1,544ms, p99 1,884ms, max 1,986ms** — pinned within 14ms of the 2,000ms ceiling. 79 confirms (5.4%) land in the last 500ms; 32 (2.2%) in the last 200ms. The genuine connected-but-no-ack case (`fallback=already_connected`) is only ~712/30d.
- Platform: mobile (Capacitor web bundle, iOS + Android).

## Suspected cause

`packages/shared/play-view/src/wall-confirm-fallback.ts` — `WALL_CONFIRM_TIMEOUT_MS = 2000` (unchanged since ef9a6d64a). The watcher is armed in `handleLightbulbClick` (`packages/web/app/components/play-view/play-view-drawer.tsx`), which conflates connect-initiation with board-ack so the metric is noisy. Fix direction: widen/adapt the window for the `already_connected`/post-send case, and/or measure confirm latency from the successful frame write rather than from the lightbulb tap; stop counting connect-initiation (`pulse_only`/`picker`/`auto_connect`) as board-ack timeouts.

## Severity

P2 — board still lights and users re-tap; no crash or core-flow block. The alarming 3:1 headline is a metric artifact, but the ~5% mis-recorded-confirm friction (premature pulse stop + picker re-pop) is real and tunable.

## Related

- #3084 (closed, P1) — sends silently fail to light the wall after a mid-session drop. Same board-send family, different event (`Climb Sent to Board Failure`) and code path.
- #3088 (open, P3) — error-tracking hygiene for noisy/inaccurate BLE failure metrics. Adjacent theme, different events.

marcodejongh:
Closes #3121.

## What & why

After a send, the play-drawer lightbulb pulses while we wait for the board to confirm the climb lit up. That wait was a one-shot **2s timer that rendered a terminal verdict**: confirm or "timeout". Two problems fell out of that (PostHog project 412845, 30d):

- **Clipped confirms.** Successful confirm latency is p95 1,544ms / p99 1,884ms / **max 1,986ms** — pinned within 14ms of the 2,000ms ceiling because the cap *truncates* the distribution. ~5% of legitimate confirmations land just after 2s and were recorded as timeouts; the pulse also stopped early.
- **A lying metric.** `Wall Confirm Timeout` fired ~3:1 vs `Wall Confirmed`, but most of that is the `pulse_only` **connect phase** (the watcher armed at lightbulb tap, 2s before a cold BLE scan/pair can even finish) — not a board-ack failure.

Stepping back: the board-presence "now on the wall" feed is *already* a convergent, seq'd, reconciling model (gap-detection, catch-up, durable `board_climb_events`). The wall-confirm pulse was the one piece that never joined it. This is **slice 1** of converging it.

## What changed (web + `@boardsesh/play-view` only)

- The pulse now **derives from convergent state** and clears on confirm / supersede / a long backstop — not an arbitrary 2s verdict. The seq'd board-presence `BoardClimbSet` is fed into the existing wall-confirm bus (gated on a *transition*, so an already-lit climb doesn't log a ~0ms confirm). A slow confirm at ~5s now lands solid-lit with its true latency instead of being clipped.
- **Supersede:** if the user navigates the queue off the pending climb, the pulse clears immediately (the longer backstop never leaves a stale pulse).
- We now arm/pulse on the **committed** climb the AutoSender actually sends, not a drawer browse-preview.
- **Telemetry split:** only the genuine connected-but-never-converged case stays `Wall Confirm Timeout`; the connect-phase reasons (`pulse_only`/`picker`/`auto_connect`/`unsupported`) move to a new **`Wall Confirm Connecting`** event so they stop inflating the board-ack metric.

The RN app doesn't use this controller — affected users run the web bundle inside the Capacitor shell, so the fix is web/shared.

## Follow-ups (out of scope here)
- The larger convergent build-out (producer durable outbox, holder durability + web `boardConnection` backfill parity, spectator/party-member convergence) is tracked separately and ties into #2860 / #2712 / #2713 / #3008 / #3087.
- **PostHog owner:** any saved insight counting `Wall Confirm Timeout` unfiltered now sees only `already_connected`; the connect-phase count moved to `Wall Confirm Connecting`. Re-tune `WALL_CONFIRM_BACKSTOP_MS` from the now-un-clipped `Wall Confirmed` latency once data lands.

## Testing
- `vp run typecheck` (web + shared), `vp check` — clean.
- `vp test run --project play-view` (145) and the web play-view suite (76) — green. New cases cover the `timeoutMs` backstop, the connected-vs-connecting classification under `pulseOnly`, the telemetry split, and a 2.5s confirm landing as `Wall Confirmed` (no clipping).
- Manual QA to exercise: tap the drawer lightbulb on a cold connect (solo + party); confirm a slow (>2s) ack lands solid-lit, not a re-pop; navigate off the pending climb mid-pulse and confirm it clears.

## Release Notes
When you send a climb, the board light keeps pulsing until the wall actually lights up — even when the board takes an extra beat to answer. No more pulse cutting out early or the device picker popping back up on a send that worked.

### DIFF
--- a/packages/shared/play-view/src/__tests__/wall-confirm-fallback.test.ts
+++ b/packages/shared/play-view/src/__tests__/wall-confirm-fallback.test.ts
@@ -309,4 +309,62 @@ describe('createWallConfirmFallbackController', () => {
       boardLayout: 'Original',
     });
   });
+
+  it('pulseOnly: classifies a connected backstop as a genuine board-ack timeout, not pulse_only', () => {
+    // Connected when the backstop fires means the connect handshake wasn't the
+    // problem — we were on the board but no confirm converged. That's the
+    // genuine board-ack signal, even though pulseOnly still suppresses connect.
+    const wallConfirmBus = createLocalWallConfirmBus();
+    const callbacks = createCallbacks();
+    const deps = createDeps(wallConfirmBus, { isBluetoothConnected: () => true });
+    const controller = createWallConfirmFallbackController(deps, callbacks);
+
+    controller.armWatcher({ ...baseClimb, pulseOnly: true });
+    vi.advanceTimersByTime(WALL_CONFIRM_TIMEOUT_MS);
+
+    expect(deps.bluetoothConnect).not.toHaveBeenCalled();
+    expect(callbacks.onTrackTimeout).toHaveBeenCalledWith({
+      mode: 'solo',
+      fallback: 'already_connected',
+      boardLayout: 'Original',
+    });
+  });
+
+  it('respects a custom timeoutMs backstop', () => {
+    const wallConfirmBus = createLocalWallConfirmBus();
+    const callbacks = createCallbacks();
+    const deps = createDeps(wallConfirmBus);
+    const controller = createWallConfirmFallbackController(deps, callbacks);
+
+    controller.armWatcher({ ...baseClimb, timeoutMs: 5000 });
+    // Past the default window — the backstop must NOT have fired yet.
+    vi.advanceTimersByTime(WALL_CONFIRM_TIMEOUT_MS + 500);
+    expect(callbacks.onTimeout).not.toHaveBeenCalled();
+    expect(callbacks.onTrackTimeout).not.toHaveBeenCalled();
+    expect(deps.bluetoothConnect).not.toHaveBeenCalled();
+
+    // Reach the custom backstop — now it fires.
+    vi.advanceTimersByTime(5000 - (WALL_CONFIRM_TIMEOUT_MS + 500));
+    expect(callbacks.onTimeout).toHaveBeenCalledWith({ climbUuid: 'climb-1' });
+    expect(callbacks.onTrackTimeout).toHaveBeenCalledOnce();
+  });
+
+  it('records a confirm that lands after the default window but within a widened backstop (no clipping)', () => {
+    const wallConfirmBus = createLocalWallConfirmBus();
+    const callbacks = createCallbacks();
+    const deps = createDeps(wallConfirmBus);
+    const controller = createWallConfirmFallbackController(deps, callbacks);
+
+    controller.armWatcher({ ...baseClimb, timeoutMs: 5000 });
+    vi.advanceTimersByTime(2500); // past the old 2s clip ceiling
+    wallConfirmBus.emit('climb-1');
+
+    expect(callbacks.onConfirmed).toHaveBeenCalledWith({
+      climbUuid: 'climb-1',
+      latencyMs: 2500,
+      confirmedByRole: 'other',
+    });
+    expect(callbacks.onTimeout).not.toHaveBeenCalled();
+    expect(callbacks.onTrackTimeout).not.toHaveBeenCalled();
+  });
 });
--- a/packages/shared/play-view/src/wall-confirm-fallback.ts
+++ b/packages/shared/play-view/src/wall-confirm-fallback.ts
@@ -11,6 +11,19 @@
  */
 export const WALL_CONFIRM_TIMEOUT_MS = 2000;
 
+/**
+ * Backstop for the connect-included arm (the lightbulb tap, which must cover a
+ * cold BLE connect + first frame write + ack convergence). The timer is no
+ * longer a verdict — a confirm is sourced from convergent state (the local BLE
+ * write, the seq'd board-presence `BoardClimbSet`, or `WallConfirmedClimb`) and
+ * the pulse otherwise clears when the climb is superseded (queue moves on) or
+ * the board disconnects. This value only bounds a *stuck* pulse and classifies
+ * the "never converged" telemetry. Kept generous so a slow cold connect that
+ * converges at ~5–6s still lands as `Wall Confirmed` rather than being clipped;
+ * re-tune from the (now un-clipped) confirm-latency distribution after shipping.
+ */
+export const WALL_CONFIRM_BACKSTOP_MS = 8000;
+
 export type WallConfirmMode = 'party' | 'solo';
 export type WallConfirmFallback = 'already_connected' | 'unsupported' | 'auto_connect' | 'picker' | 'pulse_only';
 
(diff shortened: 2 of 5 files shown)


marcodejongh:
Addressed the `claude-review` notes in 3f6e6e7:

1. **Empty-queue supersede** — dropped the `committedClimbUuid &&` guard, so an emptied queue (`committed === null`) now supersedes the pulse immediately instead of waiting out the 8s backstop.
2. **Drawer-effect coverage** — extracted the convergence + supersede effects into `useWallConfirmConvergence` and unit-tested the transition gate (confirm only on a was-not→now-is wall transition, not steady-state), the already-lit no-transition case, the committed-moves-off supersede, and the empty-queue supersede. 22 hook tests + 81 play-view tests green.

marcodejongh:
Thanks — went through all three:

1. **Test label vs. value** — fixed in bb3a403: the still-connecting case now arms with `WALL_CONFIRM_BACKSTOP_MS` and advances by it, so it exercises the real production window instead of the 2s default.
2. **Supersede has no transition guard** — intentional, and not a real race here. `pendingClimbUuid` is set from `currentClimbQueueItem.climb.uuid` and `committedClimbUuid` is derived from that same `currentClimbQueueItem` in render, so they're equal on the post-arm render — React doesn't surface a transient mismatch within a render. The only way they diverge after arm is a genuine queue change, which *is* a real supersede (the climb moved on), so firing is correct. A transition guard would instead risk leaving a stale pulse when a remote queue update lands between arm and the first render.
3. **No re-tune tracking** — fair. The `WALL_CONFIRM_BACKSTOP_MS` re-tune from the now-un-clipped latency is captured in the PR follow-ups and will get a dedicated ticket alongside the slices 2–4 epic (the broader convergent build-out). I'll link it back to the constant's doc once filed.

marcodejongh:
Marking ready for review. All change-relevant checks are green (lint, typecheck, web + backend tests, build-and-test, mobile-bundle, claude-review). The only trailing job is `prebuild-and-build` — a full Android native APK build — which this JS-only change can't affect: `ota-check` already confirmed no native change vs `main` (matching fingerprint). I'll keep watching it through to green.

marcodejongh:
## Staleness triage: recommend closing in favour of #3731

Checked whether #3121 still reproduces on `main` before forcing a rebase here (this PR is ~1 month stale and `main`'s wall-confirm code has since moved).

**#3121 is already fixed and closed on `main`**, via #3731 (merged 2026-07-17, same day this issue was closed as `COMPLETED`). #3731's own description called this PR out explicitly: *"a narrow, independent subset [of #3129]... lands the tunable fix now; the convergence and taxonomy call stays with @marcodejongh."*

**PostHog confirms the fix landed for real** (project 412845, daily `Wall Confirmed` vs `Wall Confirm Timeout`):

| Window | Wall Confirmed | Wall Confirm Timeout |
|---|---|---|
| 2026-07-01 .. 07-16 (pre-#3731) | 0–4 / day | 6–183 / day |
| 2026-07-18 .. 07-24 (post-#3731) | 3–19 / day, mostly matching or beating timeouts | 5–29 / day |

Most post-fix confirms carry `reconciledAfterTimeout: true` (e.g. 15/19 on 07-18, 8/10 on 07-19, 11/12 on 07-21) — the 10s reconcile backstop is actively rescuing genuine late acks that used to vanish as timeouts. The "structurally near-unreachable" collapse described in the 2026-07-14 triage comment on #3121 is gone.

**Rebase isn't mechanical.** `main` and this branch both independently rewrote the same core files with incompatible mechanisms (main: widen-to-3000ms + 10s reconcile-backstop state machine; this branch: derive the pulse from convergent board-presence state + a `Wall Confirm Connecting` telemetry split). A dry-run merge conflicts in all 5 touched files:

```
packages/shared/play-view/src/wall-confirm-fallback.ts
packages/shared/play-view/src/__tests__/wall-confirm-fallback.test.ts
packages/web/app/components/play-view/use-wall-confirm-fallback.ts
packages/web/app/components/play-view/__tests__/use-wall-confirm-fallback.test.tsx
packages/web/app/components/play-view/play-view-drawer.tsx
```

That's not conflict-resolution, it's re-deriving the convergent-state design on top of `main`'s already-shipped reconcile machinery.

No blocking human review is outstanding — the only comments are `claude-review` bot notes, all addressed by @marcodejongh back in June.

**Recommendation:** close this PR. The tactical bug it targeted is fixed and verified live. If the convergent-state architecture (deriving the pulse from board-presence instead of a timer) is still wanted, it's better filed as a fresh PR against current `main` — this branch's diff is no longer a clean base for it, and the "slice 1 of a larger convergence epic" framing in the original description already pointed at #2860 / #2712 / #2713 / #3008 / #3087 as the follow-up track for that work. Leaving the close/rework call to @marcodejongh.

PR #3130, opened 2026-06-21
fix(mobile): hide dev-server switcher on builds without the native dev launcher

marcodejongh:
## What & why

Admin/tester users see a **Development** section in the mobile More tab with two rows: an **OTA Channel Switcher** and a **Metro servers** (dev-server) switcher. The dev-server switcher calls `ExpoDevLauncher.loadApp(url)` to point the app at a Metro dev server, but `ExpoDevLauncher` is the native module from `expo-dev-client`, which Expo links into **dev-client / Debug builds only** — never a Release (App Store / TestFlight / Play) binary. There, `requireOptionalNativeModule('ExpoDevLauncher')` returns `null`, so testers saw the row and got `"Dev launcher unavailable"` when tapping a server.

We're not making it work on the App Store binary: loading arbitrary remote JS from any Metro URL is exactly what Apple **2.5.2 / 3.3.2** prohibit (and why Expo strips the launcher from Release). Testers already have the compliant equivalent — the **OTA Channel Switcher**, which pulls code-signed, fingerprint-matched bundles from our own server under 3.3.2's interpreted-code carve-out.

This hides the dead row where it can't work.

## Changes

- New `packages/mobile/src/lib/dev-launcher.ts` — extracts the `ExpoDevLauncher` probe into a shared `isDevLauncherAvailable()` helper (mirrors `isPreviewBuild()` in `eas-api.ts`).
- `DevServerSwitcherScreen.tsx` — uses the shared helper instead of its own inline probe (no behaviour change; the in-screen guard still fails safe on a deep link).
- `more.tsx` — gates the Metro-servers row on `isDevLauncherAvailable()`, and skips an empty Development section header when neither dev tool applies.
- New unit test `dev-launcher.test.ts` — covers the null probe (App Store case → hidden) and the linked probe (dev-client case → shown).

Resulting matrix:
- **App Store / TestFlight (tester):** launcher absent → Metro row hidden; OTA Channel Switcher shown.
- **Dev-client build (dev/tester):** launcher present → Metro row shown (Channel Switcher hidden by its existing `Updates.isEnabled && !__DEV__` gate).
- **Non-tester production user:** section hidden entirely (unchanged).

## Test plan

- `vp run test:mobile` — 2555 pass (incl. new `dev-launcher.test.ts`)
- `vp check` on touched files — clean
- App Store / Release behaviour is covered by the `null`-probe unit test (can't link `ExpoDevLauncher` into a Release build by design). On a dev-client build the Metro-servers row and switching are unchanged.

## Release Notes

https://claude.ai/code/session_0162pe9NLjJ8oCDyFDCYFUsD

### DIFF
--- a/packages/mobile/app/(tabs)/profile/more.tsx
+++ b/packages/mobile/app/(tabs)/profile/more.tsx
@@ -23,6 +23,7 @@ import { SegmentedControl } from '../../../src/components/SegmentedControl';
 import { HoldColorAccessibilitySection } from '../../../src/components/settings/HoldColorAccessibilitySection';
 import { SessionRecordingSwitchRow } from '../../../src/components/settings/SessionRecordingSwitchRow';
 import { isPreviewBuild } from '../../../src/lib/eas-api';
+import { isDevLauncherAvailable } from '../../../src/lib/dev-launcher';
 import { useGradeFormat } from '../../../src/hooks/use-grade-format';
 import { useGlassCapability } from '../../../src/hooks/use-glass-capability';
 import { useToast } from '../../../src/providers/toast-provider';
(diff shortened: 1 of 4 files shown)


marcodejongh:
Thanks @claude — all three are minor and I've left the code as-is, with rationale for the human reviewer:

1. **`isDevLauncherAvailable()`** — kept as a function to mirror the existing `isPreviewBuild()` helper in `eas-api.ts` (also a function over effectively-static build config). Consistent with the repo's build-detection pattern and gives the "why" comment a home.
2. **Explicit `launcherMock = null` in the null-case test** — deliberately mirrors `board-renderer-shim.test.ts` ("throws when the native module is not linked at all"), where the explicit assignment documents the scenario even though `beforeEach` resets it. Keeping it consistent with the established test style.
3. **`showDevSection` gating** — the error-prone part (native-module detection) is unit-tested in `dev-launcher.test.ts`; the remaining `(isDev || isTester) && (a || b)` is trivial boolean glue with no `more.tsx` render harness (the parallel `showChannelSwitcher` is likewise inline/untested). A full-screen render test or an extracted predicate would be churn rather than coverage of anything that can realistically break.

OTA check confirms this ships over-the-air (no native change vs `main`).

PR #3140, opened 2026-06-21
fix(backend): scope session feed stats to the viewer in party sessions

azuttre:
## Summary

The mobile "You" page and the web/mobile profile **Sessions** tabs showed climbing stats that **included other climbers' climbs from group/party sessions**, inflating the viewer's own numbers.

**Root cause.** Those surfaces call `sessionGroupedFeed({ userId })`. The `userId` input filtered *which* sessions appeared (sessions the climber was in) but the per-session aggregates were **whole-session** — `totalSends`, `totalFlashes`, `totalAttempts`, `tickCount`, `gradeDistribution`, `boardTypes`, `hardestGrade`/`hardestSend`, `featuredBeta` all summed every participant. The per-user split lived only in `participants[]` (sends/flashes/attempts), which has **no** per-user grade distribution, hardest, or board types. So the weekly rollup card, the per-session card's sends line, grade strip, board chip, and hardest-send hero were all party-wide on a "your sessions" surface.

This is **not** a public-feed bug: the home/activity feed passes no `userId` (it uses `followingOnly`/`boardUuid`), and whole-session aggregates are **correct** there — a social card should show the whole crew. Only the single-`userId` "your sessions" surfaces were wrong.

**Fix (resolver-side scoping).** When a single `userId` scopes the feed, count only that climber's own ticks in each party session for the per-session aggregates (totals, grade distribution, board types, hardest send, featured beta). `participants[]` stays whole-session for the leaderboard. The public/home/following feeds (no `userId`) are untouched and keep whole-session aggregates.

Why resolver-side and not consumer-side: the per-user `gradeDistribution`/`hardestGrade`/`boardTypes` simply don't exist in `participants[]`, so a client-side swap could fix the sends count but **not** the grade strip or hardest badge — the half-fix the issue explicitly warned against. Because web and mobile share the GraphQL query document (`GET_SESSION_GROUPED_FEED`) and the `SessionFeedItem` type, scoping in the resolver fixes **both clients with zero consumer changes** and no schema change.

Scope of change: backend resolver only (`packages/backend/src/graphql/resolvers/social/session-feed.ts`) plus tests. No web, mobile, or shared-schema changes.

### Behaviour changes worth calling out

- On a per-user feed, a party session's `firstTickAt`/`lastTickAt` (and therefore `durationMinutes` and feed ordering) now reflect the **viewer's own** active span in that session rather than the whole party's. That's the correct framing for a "your sessions" view, and it's now asserted in the integration test.
- `featuredBeta` is scoped to the viewer too: on a per-user card the featured beta video surfaces only if the **viewer** uploaded one (otherwise the card falls back to the viewer's own hardest-send hero). This keeps the card internally consistent — the hero send and its beta clip are both the viewer's — rather than featuring a party-mate's clip on your own stats card. The public/social feed still features the whole crew's beta.

## Release Notes

- Your session stats now count only your own climbs. On the You and profile Sessions tabs, your weekly sends and flashes, grade spread, and hardest send no longer fold in your crew's climbs from group sessions.

## Test plan

- **`session-feed-query.test.ts` (mocked, passing):** asserts the generated SQL now carries the per-user scope (`AND t.user_id = …`) on the aggregate **and** the grade/board/hardest/featured-beta enrichment queries, that `participants[]` is **not** scoped (leaderboard intact), and that a no-`userId` feed stays unscoped (whole-session preserved).
- **`session-feed-participant-scope-integration.test.ts` (new, real DB):** seeds one party session two climbers share (viewer: 3 sends incl. 1 flash, hardest at difficulty 15, climbing inside a 20-min window; partner: 2 harder sends, hardest 25, climbing 09:55 and 10:30 to bracket the viewer) and asserts:
  - `sessionGroupedFeed({ userId: viewer })` → totals scoped to the viewer (3 sends, not 5), grade distribution sums to 3, hardest send is the viewer's (difficulty 15, not the partner's 25), and `durationMinutes` is the viewer's 20-min span (not the 35-min whole-session span);
  - `participants[]` still has both climbers with their own sends (leaderboard intact);
  - the partner scopes independently (2 sends, hardest 25);
  - **no `userId`** (social feed) keeps whole-session aggregates (5 sends, hardest 25, 35-min span).
- `vp run typecheck` (full monorepo) — clean.
- `vp check` — my files are format + lint clean (the unrelated format-flagged files are pre-existing drift not touched here).
- The mocked `session-feed-board-scope` test still passes (no regression on the no-`userId` board path).

> Note on the integration test locally: this machine's `vp test` worker hits an `ECONNRESET` against the Docker Desktop postgres proxy during the shared `setup.ts` worker-DB bootstrap — it identically breaks the **pre-existing** `session-feed-board-scope-integration.test.ts`, so it's environmental, not this change. I verified the full DB path (connection + 23 KB `schemaSQL` → 32 tables) works against a fresh postgres under node, and CI's `test-backend` job runs it for real (green).

### DIFF
--- a/packages/backend/src/__tests__/session-feed-participant-scope-integration.test.ts
+++ b/packages/backend/src/__tests__/session-feed-participant-scope-integration.test.ts
@@ -1,0 +1,226 @@
+import { describe, it, expect, beforeAll, afterAll } from 'vite-plus/test';
+import { sql } from 'drizzle-orm';
+import type { SessionFeedItem } from '@boardsesh/shared-schema';
+import { db } from '../db/client';
+import { sessionFeedQueries } from '../graphql/resolvers/social/session-feed';
+
+/**
+ * Real-DB integration coverage for PER-VIEWER scoping of the session-grouped
+ * feed. The "You"/profile Sessions surfaces call sessionGroupedFeed({ userId }),
+ * and used to show WHOLE-SESSION aggregates — so a party session inflated the
+ * viewer's own sends/flashes/grades/hardest with their party-mates' climbs.
+ *
+ * This seeds one party session two climbers share:
+ *   - VIEWER logs 3 sends (2 send + 1 flash), hardest at difficulty 15,
+ *   - PARTNER logs 2 sends, BOTH harder (difficulty 20 and 25).
+ *
+ * and proves the product behaviour the fix wants:
+ *   - sessionGroupedFeed({ userId: VIEWER }) returns the VIEWER's own totals
+ *     (3 sends, 1 flash), grade distribution (sums to 3, not 5), and hardest
+ *     send (difficulty 15 — the VIEWER's, not the partner's harder 25),
+ *   - participants[] STAYS whole-session (both climbers, with each one's own
+ *     sends) so the leaderboard is intact,
+ *   - the same scoping holds independently for the PARTNER,
+ *   - a feed with NO userId (the social/home feed) is UNSCOPED and keeps
+ *     whole-session aggregates (5 sends, hardest 25).
+ *
+ * The mock-based session-feed-query.test.ts proves the SQL *shape* (the query
+ * now carries `AND t.user_id = <id>` when a userId is set); this proves the
+ * row-level numbers.
+ */
+
+const OWNER_USER_ID = 'sf-pscope-owner';
+const VIEWER_USER_ID = 'sf-pscope-viewer';
+const PARTNER_USER_ID = 'sf-pscope-partner';
+const CLIMB_UUID = 'sf-pscope-climb-1';
+const BOARD_UUID = 'sf-pscope-board';
+const SESSION_ID = 'sf-pscope-session';
+
+let boardId: number;
+
+type SessionFeedResult = { sessions: SessionFeedItem[] };
+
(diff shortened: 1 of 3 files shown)


azuttre:

- **#2 (null guard):** Aligned the main-query guard to `userId !== null`, matching the adjacent `boardIdFilter !== null` and the batch functions. `userId` is `validatedInput.userId || null` so it's `string | null` (never `undefined`/`''`); the guards just now read consistently.
- **#3 (timestamp coverage):** The partner now climbs at 09:55 and 10:30, bracketing the viewer's 10:00 to 10:20 window, and the integration test asserts the viewer-scoped `durationMinutes` is 20 vs the whole-session 35.
- **#4 (featuredBeta):** Documented in the PR description. It's intentionally viewer-scoped so the hero send and its beta clip are both the viewer's; the public/social feed still features the whole crew's beta.
- **#1 (comment style):** I couldn't find a "never write multi-line comment blocks" rule in `CLAUDE.md` or `AGENTS.md`, and `session-feed.ts` already uses multi-line comments throughout, so I don't think that rule exists. Tightened the added comments anyway.

marcodejongh (approved):
Looks good, thanks!

PR #3150, opened 2026-06-22
perf(mobile): cut board-art image memory (display-size overlays, recycle, bg flush)

marcodejongh:
## Summary

On-device performance profile of the mobile app on a **Pixel 8 Pro (Android 16)** found board-art images dominating memory (CPU, jank, and startup were all healthy):

- **~248 MB native heap** sitting idle on the Home feed (decoded board-art bitmaps; Java/JS heap was only ~24 MB).
- **~105 MB GPU** (GL mtrack) on a full play board.
- The play-drawer carousel piled a **native-res (~7 MB RGBA) holds overlay per swiped climb** into expo-image's memory cache (it plateaus, so a high ceiling rather than a leak).

This is a JS-layer pass at the most actionable parts. Three changes:

1. **Overlay at display resolution.** The play-drawer holds overlay rendered at the board's native ~1080px even though it displays at ~930px. `BoardImageNative` now takes a `renderWidth` (display px × DPR) plus a new `backgroundVariant` override so the *overlay* shrinks while the **shared board photo stays full-res** (the photo is per-board-config, decoded once for the whole wall — so it's not what accumulates). Also removes a per-render downscale. Clamped to native width (never upscales), so high-DPR phones are unaffected.
2. **`recyclingKey` on the carousel boards.** Swapping climbs now releases the previous climb's decoded overlay instead of holding both alongside.
3. **Release board-art on background.** A shared `useIsAppBackgrounded` store gates `LayeredClimbImage` to blank its `<Image>` layers while the app is backgrounded, and a hook sweeps expo-image's memory cache on background / OS memory-warning. Re-decodes from disk on resume (no network; no observed flash).

## Measured impact (Pixel 8 Pro, before → after)

| Scenario | Before | After |
|---|--:|--:|
| Carousel browse growth (board-open → 30 swipes) | +90 MB | **+73 MB (~19% less)** |
| Backgrounded with play drawer open | 681 MB | **585 MB (−96 MB)** |
| Backgrounded from feed (common) | 566 MB | 564 MB (unchanged) |

**Honest scope:** the common feed-backgrounded case is bound by Glide's (expo-image's Android engine) RAM-adaptive image cache + native-allocator page retention, which resists JS-level clearing — that needs a native `onTrimMemory → Glide.clearMemory()` hook, deferred as a follow-up. The high absolute numbers also partly reflect Glide using a 12 GB device's headroom; it auto-sizes far smaller on 4–6 GB phones.

## Release Notes

## Test plan

- `vp run test:mobile` — 2585 passed (+8 new: app-visibility store, cache-flush hook, backgrounded blanking)
- `vp check` — lint/format clean on changed files
- `vp run check:mobile-bundle` — OK (12.3 MB)
- **On-device (Pixel 8 Pro, Android 16):** release build sideloaded; verified the play board renders crisp (full background preserved, overlay sharp at display size), no resume flash, and the memory deltas above via `dumpsys meminfo`.

https://claude.ai/code/session_016gaQWQSM1dfJ7oXK3hzku9

### DIFF
--- a/docs/react-native-performance.md
+++ b/docs/react-native-performance.md
@@ -257,6 +257,41 @@ surface the previous size's bundled background paths.
 
 ---
 
+## 7. Board-art image memory: render at display size, recycle, release on background
+
+**Rule:** Board-art surfaces are the app's largest memory consumer (decoded bitmaps live in the
+Android native heap + as GPU textures, not the JS heap). Three contracts keep them in check:
+
+- **Size the overlay to the display, keep the photo full-res.** The play-drawer board passes
+  `renderWidth` (display px × `PixelRatio.get()`) so the per-climb holds overlay rasterizes at the
+  shown size instead of the board's native ~1080px, plus `backgroundVariant="full"` so the *shared*
+  board photo (one decode per board-config, reused across every climb) stays crisp. `renderWidth`
+  alone would downgrade the photo to the 416px `thumb` — only the overlay should shrink.
+- **`recyclingKey` on always-mounted board surfaces that swap climbs.** The carousel boards key on the
+  climb's `frames` so swapping releases the previous overlay instead of holding both.
+- **Release board-art on background.** `LayeredClimbImage` blanks its `<Image>` layers when
+  `useIsAppBackgrounded()` is true, and `useImageCacheMemoryManagement` (mounted at the app root)
+  sweeps expo-image's memory cache on background / `memoryWarning`. Re-decodes from disk on
+  foreground (no network).
+
+**Why:** On-device profiling (Pixel 8 Pro, Android 16) showed ~248 MB native heap idle on the feed
+and the play-drawer carousel piling a native-res (~7 MB RGBA) overlay per swiped climb into the
+cache. Display-sizing + recycling cut carousel browse-growth ~19%; background-blanking cut the
+worst case (board left open on app-switch) ~96 MB. **Ceiling:** expo-image on Android is Glide,
+whose cache/pool are sized to device RAM and resist JS `clearMemoryCache()` (the native allocator
+retains freed pages), so the common feed-backgrounded footprint is unmoved by JS — the next lever is
+native (`onTrimMemory(UI_HIDDEN) → Glide.clearMemory()`). The high MB on a 12 GB device overstates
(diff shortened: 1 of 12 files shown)


marcodejongh:
Thanks — addressed 2 of 3 in e787073:

**2. Fast Refresh listener leak — fixed.** The AppState listener is now installed lazily and ref-counted inside `subscribe` (removed on the last unsubscribe), so there's no module-load side effect for Fast Refresh to duplicate, and still only one listener total. It re-syncs from `AppState.currentState` on first subscribe.

**3. Order-dependent tests — fixed.** Each stateful test now normalizes to the foreground at the start, so they pass in any order. (This also surfaced real cross-file pollution: the existing `layered-climb-image.test.tsx` mocks `react-native` without `AppState`, and vitest throws on accessing the undefined export — kept a `try/catch` in `subscribe` so the store stays resilient to partial mocks. In production AppState is always present.)

**1. Multi-paragraph comments — not changing.** I checked: there's no comment-length rule in CLAUDE.md (the only one in the repo), and the surrounding files are densely commented (`use-native-climb-render.ts` alone has ~104 JSDoc lines; the existing `renderWidth` prop doc it sits next to is ~9 lines). The new comments deliberately match that established "explain the why" convention. Happy to trim specific ones if you'd prefer, but a blanket collapse would make these files inconsistent with their neighbours.

marcodejongh:
Both addressed in efab619:

- **`overlayPainted` reset** — `LayeredClimbImage` now resets it via `useEffect(() => { if (isBackgrounded) setOverlayPainted(false) }, [isBackgrounded])`, so the screenshot/e2e anchor re-gates on the next real `onLoad` after foreground.
- **Docs** — added a "Board-art image memory" section to `docs/react-native-performance.md` (display-size overlays + `backgroundVariant`, `recyclingKey`, background release via `useIsAppBackgrounded`/`useImageCacheMemoryManagement`, and the Glide-cache ceiling that bounds the JS-layer win).

marcodejongh:
All three minor notes addressed in e9fe2b1 (non-blocking, but tidy): `afterEach` resets the module-level flag so app-visibility tests are order-independent; the `subscribe` catch also resets `backgrounded` to stay consistent on a partial mock; and a new test asserts `recyclingKey` reaches the overlay `<Image>` but not the shared background photo.

PR #3157, opened 2026-06-22
feat(mobile): per-source ascent counts (setting, community breakdown, chart toggle)

marcodejongh:
## Summary

Ascents on a climb come from more than one place — the board's own app vs ascents logged in Boardsesh. This surfaces that split so users can choose what they see.

- **Setting** (`More → Ascent counts`): **All** / **Board app** / **Boardsesh**. Drives the sends count in climb-list rows, the play-drawer community headline, and search "most ascents" ranking.
- **Community section**: a per-source breakdown line (non-zero sources only, suppressed when redundant) and an inline source toggle on the grade-by-angle chart (defaults to the setting, hidden when only one source has data, disables empty sources). Adds a "sends" y-axis title + source-in-caption so it's clear switching source only rescales bar heights — grades stay fixed.
- **Backend/shared**: exposes nullable `kilter/aurora/boardsesh` ascensionist counts on `Climb` + `ClimbStatsForAngle` (raw columns; "Board app" = `GREATEST(kilter,aurora)` is computed client-side). Wired through both search paths, `get-climb`, and `climbStatsForAngles`. Codegen regenerated.
- **Search sort**: `Board app`/`Boardsesh` rank by that source via the standard path; the default `All` keeps the v2 covering-index fast path untouched (no new indexes).

**3-source model is deliberate** (not Kilter + Aurora separately): on the Kilter board those two are the *same* ascents from two backends — `ascensionist_count` already takes `GREATEST(kilter,aurora)` to avoid double-counting, so exposing both would show two near-duplicate numbers for one climb. Validated by a design-panel review.

## Release Notes

- See where a climb's ascents come from: choose between the board's app and Boardsesh across your climb lists.
- Open a climb's community stats to see the per-source breakdown and switch the grade-by-angle chart between sources.

## Test plan

- `vp run codegen` (no drift), `vp run typecheck:backend|:shared|:mobile|:web` — all pass
- `vp run test:mobile` (2593 tests) incl. new `ascent-count-source` + updated `community-utils` tests; backend `climb-stats-for-angles` + db `search-climbs` source-sort tests
- `vp run check:mobile-variants`, `vp run check:mobile-bundle` (iOS + Android), `vp check` on changed files, i18n catalog-completeness (en-US/es/fr parity)
- Reviewed by a QA pass: all design requirements met, covering-index fast path preserved, no `any`

https://claude.ai/code/session_01RhodffsWyKtcpn6cCeNqii


marcodejongh:
Rebased on latest `main` and addressed the review feedback in `262f26112`:

**Codex**
1. **Treat null Boardsesh counts as zero** (`ascent-count-source.ts`) — `selectSourceCount` now treats a present-but-`null` per-source count as a real `0` and only falls back to the combined total when the field is genuinely *absent* (`undefined`, i.e. queue/tick/board-presence shapes that don't select the split). This mirrors the search sort (`COALESCE(boardsesh,0)` / `GREATEST(COALESCE(kilter,0), COALESCE(aurora,0))`), so Boardsesh mode and the community chart no longer show Aurora/Kilter totals for climbs with no Boardsesh sends.
2. **Request source counts in climb detail queries** (`operations.ts`) — added `kilterAscensionistCount` / `auroraAscensionistCount` / `boardseshAscensionistCount` to `CLIMB_DETAIL_FIELDS`, so `GET_CLIMB` (board-presence, queue refresh/regrade) carries the split.
3. **Regrade queued source counts with angle changes** (`shared/queue/types.ts` + `queue-provider.tsx`) — `ClimbRegradePatch` now carries the three per-source counts, populated from the refreshed climb, so an angle change re-grades the Board app / Boardsesh subtitles and drawer headline, not just the total.

**Claude review**
1. **Spanish glossary** — `App del tablero` → `App del plafón` (and `app propia del tablero` → `app propia del plafón`) in `es/common.json` + `es/session.json`.
2. **`chartSource` async-load sync** (`CommunitySection.tsx`) — the chart toggle re-seeds from the saved preference once `loaded` flips true, with a ref guard so a user's local toggle is never clobbered.
3. **`boardAppCount` misleading breakdown** — the strict null→0 semantics above mean the breakdown and toggle now treat a null source as a real 0 (no phantom "Board app" entry / segment). The headline keeps its intentional Board-app→total display fallback, guarded explicitly on `kilter`/`aurora` both being absent.

CI: typecheck (mobile + shared), mobile tests (2603), shared queue tests (92), lint/format, and the Metro bundle check (iOS + Android) all pass locally.

marcodejongh:
Addressed the non-blocking review notes in `e84556d48`:

- **`starIcons` memo** — restored; the five-star array is memoized again so a chart toggle doesn't rebuild it.
- **Redundant `breakdownEntries` dep** — now reads `countFields.boardsesh`, so the dep array is just `[countFields]`.
- **Test coverage** — added a `boardApp` assertion to the `totalSendsForSource` multi-angle test (exercises the per-angle GREATEST sum).
- **`userPickedChartSource` never reset** — intentional: the chart source is a per-view local override, so once the user picks it we deliberately stop following the global preference until the drawer remounts. Left as documented.

marcodejongh:
Thanks — addressed in `26d6d4b87`:

1. **Phantom "Board app" breakdown entry** — added the same `kilter`/`aurora` `!= null` guard the headline uses, so the breakdown skips Board app unless the climb actually carries a board-app split. (Not reachable from `DeferredSections` today since the API returns `null` not `undefined`, but the props allow it, so worth hardening.)
2. **`ClimbStatsHistoryEntry` import** — this one's a false positive: it's still used by `buildAngleStatsMap` (`community-utils.ts:109,115`), and `vp check` (type-aware) is green, so it isn't dead.
3. **Re-seed test** — added `community-section.test.tsx` covering both the async-load re-seed and the freeze-after-user-pick behaviour.

marcodejongh:
Rebased on latest `main` and addressed the two open review items in `8e3cac390`.

**Rebase (`main` moved under us):** main removed gorhom from the play drawer (`BottomSheetModal`/`BottomSheetScrollView` → `GestureDetector`/`Animated.View`/`ScrollView`) and added the log-scale ascents-by-angle chart. Conflicts resolved so both land together:
- `DifficultyByAngleChart` keeps main's `logScaleLabel` **and** this PR's `yAxisTitle` / `emptyMessage` (both captions stack; empty-overlay vs bars conditional preserved).
- `CommunitySection` keeps the source toggle + `chartCaption` and adds `logScaleLabel`.
- `more.tsx`: ascent-counts section sits alongside main's new accessibility section (main moved hold-colour accessibility to its own route).
- i18n: `logScale` + the per-source keys coexist in all three catalogs; the `tablero → plafón` glossary fix replayed cleanly.

**1. `userPickedChartSource` leaks across climbs** — fixed by `key={displayedClimb.uuid}` on `DeferredSections` (the reviewer's option 1: remount on climb change). The chart-source override + its `userPicked` guard now reset when the drawer navigates to a new climb, so nothing leaks across climbs. The "while mounted" scenario the review called out no longer exists — the subtree unmounts on climb change, so the existing `community-section.test.tsx` (async re-seed + freeze-after-user-pick) covers the per-view behaviour.

**2. `ascentSource` should be a GraphQL enum** — replaced `ascentSource: String` with a real `enum AscentCountSource { all boardApp boardsesh }` (mirrors the existing `ZoneMatchMode` enum in this schema). Codegen now emits `'all' | 'boardApp' | 'boardsesh'` instead of a bare `string`, so clients get the valid values for free and graphql-js validates at the edge. The resolver-side Zod `z.enum(...)` stays as defence in depth.

Validation: `vp run codegen` (no drift), typecheck (shared/backend/mobile/web/db), `vp run test:mobile` (2712), backend `climb-stats-for-angles` (7), db `search-climbs` (29), shared queue (92), i18n catalog-completeness (28), `vp check` (0 errors), `vp run check:mobile-bundle` (iOS + Android OK).

marcodejongh:
Thanks for the review. Addressed the test gap in `3461f9242`; the other three are intentional / non-issues:

- **Single-source toggle suppression untested** — fixed. `community-section.test.tsx` now drives a board-app-only climb (zero Boardsesh ticks) through a mutable angle-stats store and asserts the toggle is gone (`queryByTestId('segmented')` is null) while the single-source chart still renders.
- **`ascentSource` read twice in `search-climbs.ts`** — left as-is. `searchParams` is built once by `mapSearchInputToParams` and never mutated, and the two reads sit on opposite sides of the fast-path/standard-path split (so threading one value through would just add a parameter). Agreed it's not a bug; not worth the churn.
- **`community-utils.ts` still imports `ClimbStatsHistoryEntry`** — intentional, it's still used by `buildAngleStatsMap`. `vp check` (type-aware) is green, so it isn't dead.
- **`ascent-count-source-preference.ts` transient-failure retry** — left for consistency with the sibling `grade-format-preference` module (same module-level `loadPromise` singleton, blessed in an earlier review). Behaviour is safe — it defaults to `'all'` — so I'd rather change both modules together in a dedicated pass than diverge one here.

marcodejongh:
Thanks — addressed three of these; the other two are intentional / pre-existing:

**Fixed**
- **Extra GraphQL fields on web search** (`5c43cfee`) — the three per-source counts are out of the shared `CLIMB_SEARCH_FIELDS` / `CLIMB_DETAIL_FIELDS`. No web UI reads them and the mobile climb list selects them via its own fragment (`packages/mobile/src/lib/graphql/operations.ts`), so they were pure web payload. Added a comment so they don't get re-added.
- **`REGRADE_CLIMBS` per-source fields untested** (`5fb3df2d`) — added a reducer case asserting `kilter`/`aurora`/`boardsesh` counts patch through alongside the total.
- **`AscentCountSource` two sources of truth** (`c680e7bd`) — the mobile lib now re-exports the GraphQL `AscentCountSource` enum and constrains its runtime list with `satisfies readonly AscentCountSource[]`, so a typo can no longer let `isAscentCountSource` accept a value the schema/backend rejects.

**Intentional / left as-is**
- **`ClimbListItemContent` hides the sends line when `sendsCount === 0`** — this is the deliberate per-source semantics, not a bug. When the user picks **Boardsesh**, a climb with zero Boardsesh ticks genuinely has zero Boardsesh sends; falling back to the combined total would show ~1000 "Boardsesh" sends that don't exist. That matches the maintainer's earlier call (selectSourceCount treats a present-but-null source as a real 0, never the total) and is exhaustively unit-tested in `ascent-count-source.test.ts` (undefined→total fallback, present-null→0, real-0→0, all three sources). The only fallback-to-total path is when the per-source fields are *absent* (queue/tick/presence shapes), which is intended.
- **Schema `Int` vs Drizzle `bigint`** — the existing `ascensionist_count` is already `Int!` over the same bigint-backed column, so the new per-source fields just follow the established convention. Worth a separate sweep if we ever want `BigInt` scalars, but changing only these three would be inconsistent.

marcodejongh:

Conflicting: climbs/types.ts, climbs/index.tsx, ClimbListItemContent.tsx, CommunitySection.tsx, DifficultyByAngleChart.tsx, generated GraphQL hooks/types, common.json locales.

PR #3160, opened 2026-06-22
fix(mobile): diagnose + fix Android edge-to-edge cold-start freeze

marcodejongh:
## What & why

This replaces the Android-16 freeze workaround with an earlier edge-to-edge initialization path.

The field reports point at stale Android window/inset state on newer devices: touch dies in full screen, split-screen restores it, and the debug overlay keeps updating. The previous version of this PR forced a relayout from JS after the splash screen. That was useful as a diagnostic, but it is the wrong production fix because the first React root can already have been attached against bad metrics.

This PR now moves the Android edge-to-edge setup before React attaches the root view, while keeping the diagnostic APK overlay read-only so affected-device testers can still capture metrics.

## Changes

- Add `with-android-early-edge-to-edge`, an Expo prebuild plugin that patches generated `MainActivity.kt` to enable edge-to-edge before `super.onCreate(...)`.
- Keep the existing `react-native-edge-to-edge` theme plugin for the RN 0.85 / Expo 56 `Theme.EdgeToEdge` parent.
- Remove the custom `modules/window-relayout` native module and the post-splash JS `requestApplyInsets` timeout.
- Keep `FreezeDebugOverlay` behind `EXPO_PUBLIC_FREEZE_DEBUG=1`, but make it read-only: `Snap` only, no Relayout/Recreate controls.
- Keep the login screen in a `SafeAreaView`; remove the temporary split-screen warning.
- Update the manual Android diagnostic APK workflow so its release notes and test steps match the read-only overlay, including the email/password sign-in note for debug-signed dev APKs.

## Validation

Local:

- `vp run typecheck:mobile`
- `vp run test:mobile` — 348 files, 2580 tests passed
- `vp check packages/mobile/app.config.ts packages/mobile/app/_layout.tsx packages/mobile/app/auth/login.tsx packages/mobile/src/components/FreezeDebugOverlay.tsx packages/mobile/plugins/with-android-early-edge-to-edge.js packages/mobile/src/lib/__tests__/android-early-edge-to-edge-plugin.test.ts`
- Review subagent also ran `vp run check:mobile-bundle` and `git diff --check origin/main...HEAD` successfully.

CI / APK:

- Android PR APK workflow passed on the current head (`3130896f2`) and uploaded the sideloadable APK artifact: https://github.com/boardsesh/boardsesh/actions/runs/27946470867
- Merge still needs affected-device validation on Pixel 10 / Galaxy S24 Android 16.

## Code review

Four review subagents checked separate surfaces:

- Android native/prebuild plugin: no findings.
- React Native / Expo UI behavior: no findings.
- Tests / maintainability: no findings.
- Workflow/APK consistency: found two issues; fixed by isolating the diagnostic Gradle cache and warning testers to use email/password in the debug-signed dev APK.

## On-device acceptance

- Fresh install logged out: login accepts taps after cold start.
- Logged-in force-close/reopen: home accepts taps immediately.
- Board selection and climb list: vertical scroll, row taps, swipes, toast open, and queue list all work.
- Close/reopen after selecting a climb: no touch freeze.
- Split-screen no longer changes behavior.
- Diagnostic overlay should show stable full-window metrics rather than flipping to the shortened-window / zero-top-inset state.

## Release Notes


marcodejongh:
### Update: it's a cold-start freeze, not login-specific

Contributor (spushy) found it also freezes when opening the app **from a full close while already logged in** — the home tab freezes too, not just login. The home tab already consumes \`useSafeAreaInsets()\` and still freezes, which rules out a per-screen inset gap and confirms the **cold-start window-metrics race**: the first screen's native hit-region is laid out against stale edge-to-edge metrics and stays touch-dead until a real Configuration change (split-screen) re-runs them.

Implications:
- A **pure-JS inset fix won't work** (home proves consuming insets isn't enough). The fix is the **native** \`requestApplyInsets\` relayout — not OTA-deliverable, ships in a build.
- Moved the gated relayout from login-mount to **app cold-start** (\`app/_layout.tsx\`, after \`SplashScreen.hideAsync\`) so it covers whatever screen is first (login **or** home).
- The diagnostic APK's **Relayout** button is a root sibling, so contributors can test it on either frozen screen.

PR #3162, opened 2026-06-22
perf(mobile): split BoardProvider context so logbook merges don't re-render all consumers

### LINKED ISSUE #3152: Perf: split BoardProvider context — volatile logbook re-renders all consumers (incl. always-mounted sheets)
## Summary
`BoardProvider` publishes a **single fat context value** that bundles the volatile `logbook` /
`logbookByClimbAngle` together with the stable callbacks. Every logbook merge re-publishes the whole
value, so **every `useBoardProvider()` consumer re-renders** — including chrome that has no business
re-rendering during a scroll. This is the single biggest re-render cascade found while profiling the
RN app on a physical Pixel 8, and it violates our own rule 1 in `docs/react-native-performance.md`.

## Where
`packages/shared/board-react/src/board-provider.tsx:147-175` — one `useMemo` returns
`{ boardName, …, logbook, logbookByClimbAngle, getLogbook, saveTick, saveClimb, updateClimb }`. Its
dep array includes `logbook` and `logbookByClimbAngle`, so the value identity changes on every merge.

## What happens at runtime
1. During a fling on the climbs list, the list reaches the end, `fetchNextPage()` resolves, and
   `app/(tabs)/climbs/index.tsx:562-565` calls `getLogbook(visibleClimbs.map(...))`.
2. That merges logbook data → new `logbook` array → **new context value identity**.
3. Every consumer re-renders, notably `DrawerHostProvider` (`src/providers/drawer-host-provider.tsx:165`
   only needs `getLogbook`), which then re-renders its **always-mounted** `BoardSheet` and `QueueSheet`
   bottom-sheet modals, plus every per-row `AscentStatusGlyph`.

## Evidence (Hermes CPU sampling, Pixel 8, dev build)
- climbs-scroll trace: `BottomSheetModalComponent` **30.9ms self** during a pure list scroll; mounted
  `BottomSheetModal` / `ForwardRef(ModalSheet)` component events present.
- `propagateContextChanges` / `propagateParentContextChanges` hot during both climbs-scroll and
  home-scroll — the reconciler walking the context change to all consumers.
- climbs-scroll: 9.75% legacy janky frames, 99th-percentile frame 28ms.

## Fix
Split `BoardProvider` into two contexts (the pattern the doc prescribes and `queue-provider.tsx`
already follows):
- a **stable** actions/identity context (`getLogbook`, `saveTick`, `saveClimb`, `updateClimb`,
  `boardName`, `boardUuid`, `isAuthenticated`, …) — never changes on a logbook merge;
- a **volatile** logbook context (`logbook`, `logbookByClimbAngle`) consumed only by the per-row
  `AscentStatusGlyph` / `useAscentStatus`.

Then `DrawerHostProvider` and the always-mounted sheets stop re-rendering on logbook merges. Optional
follow-on: defer the mid-fling `getLogbook` via `InteractionManager.runAfterInteractions` so the merge
lands after the fling, not during it.

## Acceptance
- Scrolling the climbs list to load several pages does **not** re-render `BoardSheet` / `QueueSheet`
  (verify with React DevTools "highlight updates" or a re-profile: `BottomSheetModalComponent` self
  time ~0 during scroll).
- `useAscentStatus` rows still update when their ticks change.

---
*Source: React-performance profiling session on a physical Pixel 8 Pro (Metro-connected debug dev-client, Hermes CPU sampling via the fusebox Tracing domain + Android `gfxinfo`). **Dev build → absolute timings are inflated; React-19 dev-mode frames (`createTask`/`reportMeasure`/`jsxDEV`/lazy module load) were discounted. A release-build re-profile is recommended to quantify absolute frame wins.** Full per-flow data + reusable tooling: `.boardsesh/perf/RESULTS.md`.*

marcodejongh:
> **Stacked on #3168** (`mobile/remove-gorhom-bottom-sheet`). This PR's base is that branch, so the diff above is only the context-split changes. Merge #3168 first; GitHub will retarget this PR to `main` automatically when #3168 lands.

## Summary

`BoardProvider` published one fat memoized context value that bundled the **volatile** `logbook` / `logbookByClimbAngle` together with the **stable** identity and callbacks. Because the logbook sits in that value's dep array, every logbook merge changed the value identity and re-rendered **every** `useBoardProvider()` consumer — even ones that read only `getLogbook`. On a Pixel 8 this was the biggest re-render cascade in the climbs-list scroll: the screen calls `getLogbook(visibleClimbs)` mid-fling → the merge replaces the logbook → the screen (which reads only `getLogbook`) re-renders itself.

Closes #3152.

### What changed

Split into two new contexts — the pattern `queue-provider.tsx` (8 contexts) and `board-presence-provider.tsx` (5 contexts) already follow:

- **`BoardActionsContext`** (stable): `boardName`, `boardUuid`, `isAuthenticated`, `isLoading`, `error`, `isInitialized`, `getLogbook`, `saveTick`, `saveClimb`, `updateClimb`. Its `useMemo` deps deliberately **exclude** the logbook, so its identity holds steady across a merge.
- **`BoardLogbookContext`** (volatile): `logbook` + the `logbookByClimbAngle` index.
- The full **`BoardContext`** is kept (derived as `{ ...actionsValue, ...logbookValue }`) so web and any back-compat consumer keep using `useBoardProvider()` unchanged. Web is left as-is — it isn't the profiled problem.

Mobile consumers move to the narrowest context:

| Consumer | Reads | Now uses |
|---|---|---|
| `climbs/index.tsx` | `getLogbook` | `useBoardActions()` — stops the mid-fling self re-render |
| `use-create-climb-screen.ts` | `isAuthenticated`, `saveClimb`, `updateClimb` | `useBoardActions()` |
| `use-ascent-status.ts` (per row) | `logbookByClimbAngle` | `useOptionalBoardLogbook()` — still updates when ticks change |
| `QuickTickBar.tsx` | `getLogbook` + `logbookByClimbAngle` | `useOptionalBoardActions()` + `useOptionalBoardLogbook()` |
| `use-log-ascent-action.ts` | `logbook` + `getLogbook` | `useOptionalBoardLogbook()` + `useOptionalBoardActions()` |

Also defers the mid-fling `getLogbook` in the climbs list via `InteractionManager.runAfterInteractions` (cancelled on cleanup) so the merge — and the per-row ascent-glyph re-renders it triggers — lands after the fling settles instead of during it.

## Release Notes

## Test plan

- `vp run typecheck:shared`, `vp run typecheck:mobile`, `vp run typecheck:web` — clean.
- `vp run test:mobile` — 2577 passed. Shared board-react suite — 56 passed, including a new identity-stability test asserting the actions-context reference is `toBe` identical across a logbook merge while the logbook-context reference changes.
- `vp check` (changed files) — lint + format clean.
- `vp run check:mobile-bundle` — OK (12.3 MB).

### Acceptance (#3152)
- Stable-only consumers (`climbs/index.tsx`, create-climb) no longer re-render on logbook merges.
- `useAscentStatus` rows still flip to flash/send/attempt when their ticks change.
- Glyphs for newly-scrolled rows appear just after the fling settles (defer), not mid-fling.

### DIFF
--- a/docs/react-native-performance.md
+++ b/docs/react-native-performance.md
@@ -38,11 +38,17 @@ an automated gate.
 
 **Examples in this repo:**
 
-- `packages/shared/board-react/src/board-provider.tsx` — the `value` is `useMemo`'d over
-  `[boardName, isAuthenticated, …, logbook, logbookByClimbAngle, getLogbook, saveTick, …]`. The
-  mutation callbacks (`saveTick`/`saveClimb`/`updateClimb`) are kept stable via the
-  ref-updated-in-`useEffect` + empty-dep-`useCallback` pattern (lines 90–116), so a React Query
-  `mutateAsync` reference change does not churn the context.
+- `packages/shared/board-react/src/board-provider.tsx` — splits into **three** contexts (#3152):
+  `BoardActionsContext` (the **stable** half — identity, auth/init flags, `getLogbook`/`saveTick`/
+  `saveClimb`/`updateClimb`; its `useMemo` deps deliberately exclude the logbook, so a merge never
+  changes its identity), `BoardLogbookContext` (the **volatile** `logbook` + `logbookByClimbAngle`
+  index, read only by the per-row `useAscentStatus` and the tick forms), and the full back-compat
+  `BoardContext` (`{ ...actionsValue, ...logbookValue }`) that web still consumes via
+  `useBoardProvider()`. A callback-only consumer (`climbs/index.tsx` reading just `getLogbook`,
+  the create-climb screen reading the mutations) subscribes via `useBoardActions()` and skips the
+  per-merge re-render entirely. The mutation callbacks stay stable via the
+  ref-updated-in-`useEffect` + empty-dep-`useCallback` pattern, so a React Query `mutateAsync`
+  reference change does not churn the context either.
 - `packages/mobile/src/providers/queue-provider.tsx` — splits into **two** contexts:
   `QueueContext` (full state + actions) and `QueueSessionControlContext` (just the wall-control
   state + callbacks). A component that only drives take/release control subscribes to the smaller
--- a/packages/mobile/app/(tabs)/climbs/__tests__/index-keyboard.test.tsx
+++ b/packages/mobile/app/(tabs)/climbs/__tests__/index-keyboard.test.tsx
@@ -108,7 +108,7 @@ vi.mock('@boardsesh/climb-filters', () => ({
 }));
 
 vi.mock('@boardsesh/board-react', () => ({
-  useBoardProvider: () => ({ getLogbook: mocks.getLogbook }),
+  useBoardActions: () => ({ getLogbook: mocks.getLogbook }),
 }));
 
 vi.mock('../../../../src/components/ClimbListRow', () => ({
--- a/packages/mobile/app/(tabs)/climbs/index.tsx
+++ b/packages/mobile/app/(tabs)/climbs/index.tsx
@@ -30,7 +30,7 @@ import { useTheme } from '../../../src/providers/theme-provider';
 import { selectByVariant } from '../../../src/theme/variants';
 import { useActiveClimbUuid, useQueueActions } from '../../../src/providers/queue-provider';
 import { ClimbSearchProvider, useClimbSearch, type GradeBound } from '../../../src/providers/climb-search-provider';
-import { useBoardProvider } from '@boardsesh/board-react';
+import { useBoardActions } from '@boardsesh/board-react';
 import { randomUUID } from 'expo-crypto';
 import { type SearchHeaderHandle } from '../../../src/components/SearchHeader';
 import { RecentFilterPills } from '../../../src/components/RecentFilterPills';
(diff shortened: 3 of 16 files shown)


marcodejongh:
Thanks — addressed in e882dedb2:

- **#1 (raw context exports):** Fixed. `BoardActionsContext` / `BoardLogbookContext` are no longer exported; consumers go through the guarded `useBoardActions` / `useBoardLogbook` hooks. The full `BoardContext` stays exported since web reads it directly.
- **#2 (cancel comment):** Reworded. The cleanup cancels a *superseded deferred callback* (a stale `visibleClimbs` snapshot never fires its `getLogbook`); network-level dedupe is separately handled by `useLogbook`'s fetched-uuid set. The old wording conflated the two.

On the rest, intentionally not changing:

- **#3 (`getLogbook` no equality check):** Pre-existing — `getLogbook` has always called `setClimbUuids(uuids)` unconditionally, and `useLogbook`'s `fetchedUuidsRef` filters already-fetched uuids so no duplicate network fetch results. The defer actually *reduces* call frequency (one callback per settled interaction). Adding an equality short-circuit here is an orthogonal change with its own re-fetch-after-invalidation edge cases, so out of scope for this PR.
- **#5 (full `BoardContext` still re-renders on merge):** By design, and documented in the code. Web reads `logbook` through the full context and *must* re-render on a merge or its logbook UI goes stale. Web isn't the profiled problem; the split's win is for consumers that move to the narrow contexts (mobile, here). The full context deliberately preserves the old behavior for back-compat.
- **#4 (no cancellation test):** This is standard React effect-cleanup with low regression risk; a dedicated assertion would live in the 300-line `index-keyboard.test.tsx` (whose RN mock already runs `runAfterInteractions` synchronously and returns a no-op `cancel`). The cost/brittleness outweighs the value of testing React's own cleanup mechanics. The existing test still verifies `getLogbook` is invoked with the visible uuids.

Issue #1450, opened 2026-04-11
[Bug] moonboard 24 importer using wrong grades


marcodejongh:
![image](https://github.com/user-attachments/assets/3a345f1d-c784-4dfe-bc4e-63b3226b4ff1)

My test screenshot imports have wrongly been graded v1

fallaciousreasoning:
any progress on this - it'd be sweet to be able to use Boardsesh for Moonboarding

fallaciousreasoning:
Wow, just had a play around - its surprisingly hard to extract a list of climbs from the Moonboard app - I'll faff around a bit more and see if I can get it working.

fallaciousreasoning:
@marcodejongh I'm working on extracting the data at the moment (extractor is working, its just slow). I can extract:

- Climb Name
- Setter Name
- Setter Grade
- User Grade
- Holds
- Labels (benchmark, no kickboard ect)

pretty easily. I can also extract the date set & rating but its way slower (because its on a separate page). Does the importer require it? If not I might leave it out.

FWIW, I have all the benchmarks if that'd be useful on its own?

fallaciousreasoning:
[Benchmark Problems Moonboard 2024 40.json](https://github.com/user-attachments/files/29187675/Problems.Moonboard.2024.40.json)

I've got the 2024 40 Degree Benchmarks here, working on the rest. It unfortunately doesn't have holdset ids or setter ids

marcodejongh:
Wow awesome! 

We can infer the holdsets from the holds very easily.

You could build a data importer in the repo and PR it, and I can run it on the prod database. Im also happy to build the importer for you (well claude would lol)

marcodejongh:
I'll run claude on building an importer for that JSON, you can focus on getting more JSONs 😄

fallaciousreasoning:
Awesome, works for me! If its better for you I've got screenshots of about 50% of the problems locally but they're large ~15GB total so I've just been trying to munge the data I've got into something like what the exports of the other problems look like to make writing the importer easier 😆 

Hopefully I can get you the full 2024 @ 40 degree set of problems tomorrow 🤞

marcodejongh:
Sounds good, I have a PR up for the importer. Maybe jump on discord and sent me a message easier to talk taht way

fallaciousreasoning:
Hey also noticed another couple of things:
- anyone can make new moonboard benchmarks at the moment (by setting the benchmark flag at creation time). This should probably be restricted
[...]

PR #3171, opened 2026-06-23
feat: structured climb characteristics system

marcodejongh:
## What this does

Introduces a structured **climb characteristics** system and migrates the existing "magic string" metadata onto it. Motivated by [#1450 (comment)](https://github.com/boardsesh/boardsesh/issues/1450#issuecomment-4773147078): MoonBoard problems had no structured tags, and anyone could flag a benchmark at creation.

Storage is a single `board_climbs.characteristics text[]` column (GIN-indexed, mirroring the existing `required_set_ids` / `compatible_size_ids` array idiom). Tokens: `no_match`, `method_footless`, `method_footless_kickboard`, `method_no_kickboard`. Helpers live in `@boardsesh/shared-schema` (`isNoMatch`, `getMoonBoardMethod`, `withCharacteristic`, `moonBoardMethodToCharacteristic`).

### "No match" is now a real characteristic
Previously a `No match\n` prefix on the climb description, detected by regex in ~14 read sites and the duplicate-detection SQL. Now:
- **Backfilled** from existing descriptions (migration `0135`, idempotent, scoped exactly like the old `LIKE 'no match%'` filter).
- **Derived on ingest** in aurora-sync, kilter-sync, and the JSON importer — the description prefix stays as the Aurora wire format, the array is the internal source of truth.
- The GraphQL `is_no_match` field, the two duplicate/similarity gates (`@> ARRAY['no_match']`), and the web + mobile search paths all read the characteristic (description fallback during transition).

### MoonBoard method tags
The importer already parsed `problem.method` and threw it away. It now stores the method as a characteristic, and method tags render on climb cards / detail (web + mobile) and are selectable when creating a MoonBoard climb.

### Benchmark creation restricted
`saveMoonBoardClimb` now requires an **admin or community leader** to set `isBenchmark` (the bulk importer is exempt). Changing benchmark status afterwards still goes through the existing community proposals flow.

## Testing
- `vp run typecheck` (full), `vp check` — green
- `vp test`: shared-schema characteristics (new), backend climb-mutations (incl. new benchmark-gate + method cases) and climb-similarity (updated SQL), web (4752), mobile (2584), db, i18n catalog completeness — all green
- Backend suite: 2 pre-existing, unrelated failures (`board-presence` slug isolation + a `graphql-resolvers` WS "socket hang up") — confirmed failing identically on clean `main`, untouched by this change.

## Notes / follow-ups
- The exact MoonBoard `method` source strings should be confirmed against the dump's distinct `.data[].method` values; the mapper is substring-based and maps anything unknown (incl. the "feet follow hands" default) to no token.
- Deferred: the create-form benchmark switch is still shown to everyone (the server rejects non-privileged saves) — hiding it for non-admins needs client role awareness; and the MoonBoard edit modal doesn't yet restore/edit method.

## Release Notes

MoonBoard problems now carry their style — footless, footless + kickboard, and no-kickboard show as tags right on the climb, and you can pick one when you set a problem. Benchmarks are now set by the crew that curates them, not anyone with the create screen open. Under the hood, "no matching" is proper climb data now instead of a note buried in the description.

### DIFF
--- a/packages/backend/src/graphql/resolvers/climbs/climb-similarity.ts
+++ b/packages/backend/src/graphql/resolvers/climbs/climb-similarity.ts
@@ -192,8 +192,13 @@ export async function findExactDuplicateMatch({
         AND ${dbSchema.boardClimbs.framesCount} = 1
         -- "No match" placeholder climbs are not real routes; never block a real
         -- setter's save by citing a placeholder. Reads the structured no_match
-        -- characteristic (backfilled from the Aurora description convention).
-        AND NOT (COALESCE(${dbSchema.boardClimbs.characteristics}, '{}') @> ARRAY['no_match'])
+        -- characteristic (backfilled from the Aurora description convention); the
+        -- description LIKE is a transition fallback for any row synced in the
+        -- window between the backfill migration and this code deploy.
+        AND NOT (
+          COALESCE(${dbSchema.boardClimbs.characteristics}, '{}') @> ARRAY['no_match']
+          OR LOWER(COALESCE(${dbSchema.boardClimbs.description}, '')) LIKE 'no match%'
+        )
         ${excludeUuid ? sql`AND ${dbSchema.boardClimbs.uuid} <> ${excludeUuid}` : sql``}
       GROUP BY
         ${dbSchema.boardClimbs.uuid},
@@ -332,8 +337,13 @@ export async function findSimilarClimbs({
           AND c.frames_count = 1
           -- "No match" placeholder climbs are not real routes; exclude them from
           -- both the gate and the similarity list. Reads the structured no_match
-          -- characteristic (backfilled from the Aurora description convention).
-          AND NOT (COALESCE(c.characteristics, '{}') @> ARRAY['no_match'])
+          -- characteristic (backfilled from the Aurora description convention); the
+          -- description LIKE is a transition fallback for any row synced in the
+          -- window between the backfill migration and this code deploy.
+          AND NOT (
+            COALESCE(c.characteristics, '{}') @> ARRAY['no_match']
+            OR LOWER(COALESCE(c.description, '')) LIKE 'no match%'
+          )
           ${excludeUuid ? sql`AND h.climb_uuid <> ${excludeUuid}` : sql``}
         GROUP BY h.climb_uuid
         HAVING COUNT(DISTINCT h.hold_id) >= CEIL(${targetSize}::float * ${safeThreshold})::int
--- a/packages/backend/src/graphql/resolvers/climbs/climb-similarity.ts
+++ b/packages/backend/src/graphql/resolvers/climbs/climb-similarity.ts
@@ -190,9 +190,10 @@ export async function findExactDuplicateMatch({
         AND ${dbSchema.boardClimbs.isDraft} = FALSE
         AND ${dbSchema.boardClimbs.isListed} IS NOT FALSE
         AND ${dbSchema.boardClimbs.framesCount} = 1
-        -- Aurora-convention "No match" placeholder climbs are not real
-        -- routes; never block a real setter's save by citing a placeholder.
-        AND LOWER(COALESCE(${dbSchema.boardClimbs.description}, '')) NOT LIKE 'no match%'
+        -- "No match" placeholder climbs are not real routes; never block a real
+        -- setter's save by citing a placeholder. Reads the structured no_match
+        -- characteristic (backfilled from the Aurora description convention).
+        AND NOT (COALESCE(${dbSchema.boardClimbs.characteristics}, '{}') @> ARRAY['no_match'])
         ${excludeUuid ? sql`AND ${dbSchema.boardClimbs.uuid} <> ${excludeUuid}` : sql``}
       GROUP BY
         ${dbSchema.boardClimbs.uuid},
@@ -329,9 +330,10 @@ export async function findSimilarClimbs({
           -- left is_listed NULL (most kilter/tension Aurora-synced climbs).
           AND c.is_listed IS NOT FALSE
           AND c.frames_count = 1
-          -- Aurora-convention "No match" placeholder climbs are not real
-          -- routes; exclude them from both the gate and the similarity list.
-          AND LOWER(COALESCE(c.description, '')) NOT LIKE 'no match%'
+          -- "No match" placeholder climbs are not real routes; exclude them from
+          -- both the gate and the similarity list. Reads the structured no_match
+          -- characteristic (backfilled from the Aurora description convention).
+          AND NOT (COALESCE(c.characteristics, '{}') @> ARRAY['no_match'])
           ${excludeUuid ? sql`AND h.climb_uuid <> ${excludeUuid}` : sql``}
         GROUP BY h.climb_uuid
         HAVING COUNT(DISTINCT h.hold_id) >= CEIL(${targetSize}::float * ${safeThreshold})::int
--- a/packages/backend/src/graphql/resolvers/climbs/mutations.ts
+++ b/packages/backend/src/graphql/resolvers/climbs/mutations.ts
@@ -9,6 +9,7 @@ import {
   CLIMB_CHARACTERISTICS,
   isNoMatchClimb,
   withCharacteristic,
+  withNoMatch,
 } from '@boardsesh/shared-schema';
 import type { BoardName } from '@boardsesh/board-constants';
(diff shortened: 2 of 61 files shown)


marcodejongh:
Thanks — addressed the review in 7c4c08da0:

- **Stale `method` dep**: added `method` to the MoonBoard save callback's deps array.
- **`updateClimb` deriving no_match for any board**: added a `boardType !== 'moonboard'` guard so a MoonBoard description that happens to start with "no match" (user prose) can't clobber the method token. Covered by two new tests (Aurora derives, MoonBoard preserves).
- **kilter-sync conflict gap**: the `onConflictDoUpdate` branch now writes `characteristics` (COALESCE, matching the holdFingerprint/frames pattern) so a forced re-sync doesn't skip the token.

Deferred (already noted in the description): hiding the benchmark toggle from non-admins in the create form (server already rejects), plus method support in the MoonBoard edit modal. Happy to open a tracking issue for those if you'd like — I was blocked from filing one automatically.

On the `is_no_match` resolver fallback: it's a one-line composition of `isNoMatch` and `isNoMatchClimb`, both unit-tested in `characteristics.test.ts` / `no-match.test.ts`, and exercised end-to-end by the search path — I left it without a dedicated test for the resolver tree to avoid importing the full resolver graph just for a ternary, but can add one if you'd prefer.

marcodejongh:
Second pass addressed in 8be9053da:

- **kilter-sync stale no_match**: the dedup conflict branch now overwrites with `excluded.characteristics` (not COALESCE), so a "No match" prefix removed upstream clears the token — matching aurora-sync. Kilter only ever carries `no_match`, so nothing unrelated can be dropped.
- **Type mid-imports**: moved `MoonBoardMethodToken` below the import block.
- **Ambiguous mapper coverage**: added tests documenting that `footless` + `kickboard` (in any phrasing) resolves to `method_footless_kickboard`, and `no kickboard` alone to `method_no_kickboard`.

The benchmark-toggle visibility remains the one tracked UX follow-up (server already rejects).

marcodejongh:
Addressed in b0016ea92:

- **Mapper semantics**: a "no kickboard" qualifier now negates the kickboard half, so "Footless, no kickboard" resolves to `method_footless` (not the kickboard-allowed token). Updated the test to assert the corrected behavior.
- **Dedup fallback test**: the climb-similarity test now also asserts the `LIKE 'no match%'` transition fallback is present in the rendered SQL.
- **Clear-on-removal test**: added an `updateClimb` test verifying the `no_match` token clears to `null` (not `[]`) when the Aurora prefix is removed from the description.

Two items I'm leaving as tracked follow-ups rather than expanding this PR:
- **`method` as a GraphQL enum**: the three tokens are already validated by Zod at the resolver (`SaveMoonBoardClimbInputSchema`); promoting to a transport-level enum is a clean contract improvement but adds codegen surface — worth a small dedicated change.
- **Editing method post-creation**: needs the MoonBoard edit modal to carry `method` through `updateClimb` (same follow-up as the benchmark-toggle visibility).

Marking ready for review once CI is green.

marcodejongh:
Thanks. Converging here — these are non-blocking style/design notes rather than correctness regressions, and I'd rather not churn CI with another commit that just retriggers review:

- **`insertCalls[N]` index assertions**: that's the existing convention in this test file (e.g. the pre-existing `insertCalls[2]` stats-row checks) — my new cases follow it for consistency. Happy to refactor the whole file to lookup-by-table in a separate cleanup if you'd prefer, but it's out of scope for this change.
- **`method` as a GraphQL enum**: noted earlier — Zod already validates the three tokens at the resolver; promoting to a transport enum is a clean follow-up with its own codegen surface.

All genuine correctness items from the earlier rounds are addressed and CI is green. Ready for your review/merge — I'll keep watching for any human feedback or conflicts.

packages/backend/src/graphql/resolvers/climbs/climb-similarity.ts
```diff
@@ -190,9 +190,15 @@ export async function findExactDuplicateMatch({
...
         AND ${dbSchema.boardClimbs.framesCount} = 1
-        -- Aurora-convention "No match" placeholder climbs are not real
-        -- routes; never block a real setter's save by citing a placeholder.
-        AND LOWER(COALESCE(${dbSchema.boardClimbs.description}, '')) NOT LIKE 'no match%'
+        -- "No match" placeholder climbs are not real routes; never block a real
+        -- setter's save by citing a placeholder. Reads the structured no_match
+        -- characteristic (backfilled from the Aurora description convention); the
+        -- description LIKE is a transition fallback for any row synced in the
+        -- window between the backfill migration and this code deploy.
+        AND NOT (
```
marcodejongh, line 198: Why do we need a and not close on no match here?

packages/backend/src/graphql/resolvers/climbs/mutations.ts
```diff
@@ -550,7 +568,22 @@ export const climbMutations = {
         publishedAt: nextPublishedAt,
       };
       if (validated.name !== undefined) updateSet.name = validated.name;
-      if (validated.description !== undefined) updateSet.description = validated.description;
+      if (validated.description !== undefined) {
+        updateSet.description = validated.description;
+        // Keep the no_match characteristic in sync with the Aurora description
```
marcodejongh: No we want to remove it from the description after we apply the migration

packages/mobile/src/components/ClimbAttributeIcons.tsx
```diff
@@ -9,10 +11,27 @@ type ClimbAttributeIconsProps = {
   isNoMatch?: boolean | null;
```
marcodejongh: Do we still need this seperate bool if we have characteristics?

marcodejongh (commented):
COmmenbts

packages/mobile/src/components/play-drawer/PlayDrawer.tsx
```diff
@@ -636,6 +636,7 @@ export const PlayDrawer = forwardRef<PlayDrawerHandle, PlayDrawerProps>(function
                   setterUsername={displayedClimb.setter_username}
                   isNoMatch={displayedClimb.is_no_match}
                   benchmarkDifficulty={displayedClimb.benchmark_difficulty}
+                  characteristics={displayedClimb.characteristics}
```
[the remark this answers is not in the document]
marcodejongh: Fixed in eea1bb5ca: added `characteristics` to the `getClimbByUuid` DB select + returned `Climb` object, and added both `is_no_match` and `characteristics` to `CLIMB_DETAIL_FIELDS` in mobile operations.ts. Direct drawer opens now carry method tags.

packages/backend/src/graphql/resolvers/climbs/climb-similarity.ts
marcodejongh: The `AND NOT` block excludes "No match" placeholder climbs from being returned as duplicate matches. Aurora seeds many boards with placeholder climbs that share hold footprints with real routes — without the exclusion, every real climb whose holds overlap a placeholder would be blocked from saving by the dedup guard. The `COALESCE` inside is needed because `NULL @> ARRAY['no_match']` returns NULL (not FALSE), which would incorrectly filter out rows whose `characteristics` column hasn't been populated yet.

packages/backend/src/graphql/resolvers/climbs/mutations.ts
marcodejongh: Done in eea1bb5ca: `updateClimb` now derives `isNoMatchFromDesc` from the incoming description first, then calls `withNoMatch(validated.description, false)` to strip the prefix before storing it. Characteristics is set from the pre-strip read so the status is preserved correctly. Descriptions updated through the edit flow will gradually clean up as users save.

packages/mobile/src/components/ClimbAttributeIcons.tsx
marcodejongh: Addressed in eea1bb5ca. `ClimbAttributeIcons` now derives no-match from `characteristics` when present (using `isNoMatch(characteristics)` from shared-schema), and only falls back to the `isNoMatch` bool for tick-sourced rows that carry the flag but not the full array (e.g. session/logbook rows built from `SessionDetailTick`). Callers that already supply `characteristics` (PlayDrawer, PlayDrawerHeader, ClimbReactionMenu) no longer pass the redundant boolean. Two new tests document the priority: characteristics wins, and the bool kicks in only when characteristics is absent.

PR #3174, opened 2026-06-24
chore(repo): kill dead screenshot CI job, drop unused data, add large-file guard

marcodejongh:
## Summary

The repo had become painful to clone: a **1.66 GiB pack**, even though the working tree (minus `node_modules`) is ~265 MB. ~1 GB of that was dead screenshot dumps on **263 orphaned `pr-screenshots/*` branches** created by a long-disabled CI job (`report-screenshots`, `if: false`). Those branches have been **deleted on origin** (out of band — a fresh clone's PNG payload drops from ~1,570 MB to ~310 MB). This PR removes the source and stops it coming back:

- **Remove the dead `report-screenshots` job** from `.github/workflows/e2e-tests.yml` — it was the generator that pushed `<sha>/layouts/*.png` dumps to `pr-screenshots/{PR}` branches. Also delete its now-orphaned `cleanup-screenshots.yml` teardown workflow.
  - The upstream `screenshots` capture job stays: it now only uploads an artifact, and the `help-screenshots` Playwright spec is still used to regenerate committed help images. It can be removed in a follow-up if there's no appetite for the artifact.
- **Drop `packages/moonboard-ocr/eng.traineddata` (5 MB)** — unused and already gitignored; `tesseract.js` fetches trained data from the CDN at runtime.
- **Add a `check:large-files` guard** (`scripts/check-large-files.mjs`) wired to `vp run check:large-files` and an **always-on CI job** that fails any PR adding a tracked file over **2 MB** outside a small, documented allowlist (OCR fixtures, vendored board-renderer libs, design renders). Dependency-free (bare `node`) so it runs cheaply on binary-only PRs too.

Compiling the vendored board-renderer `.a`/`.so` on demand (so they stop being committed) is a separate, higher-risk PR.

## Release Notes

## Test plan

- `node scripts/check-large-files.mjs --all` → passes (44 allowlisted, 0 violations).
- `node scripts/check-large-files.mjs --base origin/main` (diff mode) → passes.
- Negative test: staging a 3 MB non-allowlisted file → exits 1 with remediation guidance.
- `vp check` on changed TS/mjs → clean; both edited workflows parse as valid YAML.
- CI on this PR exercises the new `large-files` job end-to-end.

https://claude.ai/code/session_018mk35t97WJ23hQpnPhmNhE


marcodejongh:
Thanks — addressed in 69a45d61e:

1. **`actions/checkout@v6`** — not a typo; `@v6` is what the rest of `ci.yml` already uses (the `changes`/`lint` jobs), and v6 exists. Left as-is for consistency.
2. **Shallow-clone / merge-base** — real catch. The script now falls back to a full tracked-file scan if the base or merge-base can't be resolved, instead of letting the three-dot `git diff` throw. A deep PR fails loudly on a real offender rather than erroring on infrastructure.
3. **Pre-commit enforcement** — added a header note that the CI `large-files` job is the enforcement point (the check is diff/repo-scoped, not per-file like `vp check`); run locally with `vp run check:large-files`.
4. **Tests** — added `scripts/check-large-files.test.mjs` (node --test, 7 cases) covering allowlist matching, the threshold boundary, violation routing, and sort order. The CI job runs them too (`node --test`), and `vp run check:large-files` depends on `test:large-files`.

PR #3184, opened 2026-06-24
fix(mobile-ota): show what's new in the deploy Discord message

marcodejongh:
## Summary

The OTA production workflow (`mobile-ota-production.yml`) creates a local `chore(changelog): refresh from merged PRs` commit before the Discord notify step runs. That step read the update message with `git log -1 --pretty=%s`, so by then HEAD was always the changelog commit — the deploy message echoed "chore(changelog): refresh from merged PRs" every time instead of the changes the OTA actually shipped.

This reads the changelog instead and announces what's new:

- The generate step snapshots `changelog.generated.json` **before** regenerating it, then diffs prev → next to find the entries this OTA newly adds.
- A new testable transform (`scripts/lib/changelog-discord-summary.ts`) groups those entries as **✨ New / 🔧 Improved / 🐛 Fixed** and renders a compact "What's new" block (capped at 15 bullets with an "…and N more" overflow line to stay under Discord's 2000-char limit).
- The thin I/O shell (`scripts/changelog-discord-summary.ts`, wired as `vp run changelog:discord-summary`) writes the block to a file the notify step reads.
- When nothing new shipped (changelog unchanged), the message falls back to the triggering commit's subject — now read from `$COMMIT_SHA`, not `git log -1`, so the workflow's own changelog commit can never leak in again.

Degrades safely: a missing/malformed snapshot is treated as "no entries", and the summary step is best-effort (`|| true`), so it can never break a deploy notification.

## Release Notes

## Test plan

- Added `scripts/__tests__/changelog-discord-summary.test.ts` covering the diff, category grouping/ordering, empty-result fallback, and overflow cap.
- Verified the rendering output shape against sample entries.
- `vp run typecheck` + `vp test` pending local dependency install (running); will confirm green on CI.

---
_Generated by [Claude Code](https://claude.ai/code/session_01X9ii2ZgAhB97nLJTd9G32T)_

### DIFF
--- a/.github/workflows/mobile-ota-production.yml
+++ b/.github/workflows/mobile-ota-production.yml
@@ -185,7 +185,21 @@ jobs:
           GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         run: |
           set -euo pipefail
+          # Snapshot the changelog as it stood BEFORE this run so the Discord notify
+          # step can diff prev→next and announce only what this OTA newly adds. Must
+          # be captured before `generate:changelog` overwrites the working copy. A
+          # first-run/missing file degrades to an empty entry list (full new list).
+          git show HEAD:packages/mobile/src/data/changelog.generated.json \
+            > "$RUNNER_TEMP/changelog.prev.json" 2>/dev/null \
+            || echo '{"entries":[]}' > "$RUNNER_TEMP/changelog.prev.json"
           vp run generate:changelog
+          # Render the "What's new" block (newly-added entries, grouped by category)
+          # for Discord. Written to a file the notify step reads; empty when nothing
+          # is new, so that step falls back to the triggering commit subject.
+          vp run changelog:discord-summary -- \
+            --prev "$RUNNER_TEMP/changelog.prev.json" \
+            --next packages/mobile/src/data/changelog.generated.json \
+            --out "$RUNNER_TEMP/discord-changelog.md" || true
           git config user.email "github-actions[bot]@users.noreply.github.com"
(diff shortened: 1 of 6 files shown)


marcodejongh:
Thanks for the review. Addressed in `3a82939`:

- **Overflow count** — `renderDiscordSummary` now filters to renderable categories before capping, so `overflow` only counts entries that would have appeared. Keeps the "…and N more" count honest if a category is ever added without a `CATEGORY_DISPLAY` row.
- **Unused `/// <reference types="node" />`** — removed from the pure lib; it imports nothing from Node.

Skipping the malformed-JSON test for `readEntries`: it lives in the I/O shell, which runs its argv parse + file write at import time, so testing the catch branch would mean restructuring the shell to be importable without side effects. That's more machinery than a best-effort deploy helper warrants — the swallow-and-continue is deliberate and the pure transform is covered.

---

marcodejongh:
All three addressed in `f7454db`:

- **`readEntries` shape guard** — now returns `[]` unless `parsed.entries` is actually an array, so a well-formed-but-wrong-typed JSON can't reach the transform and throw on `.filter`.
- **Unknown-category test** — added a case that casts a category outside `CATEGORY_DISPLAY` and asserts it's dropped *and* doesn't inflate the "…and N more" count.
- **Header/bullet newline** — good catch, and it was worse than cosmetic: `$(printf '…\n')$(cat …)` strips the trailing newline and glues the header onto the first label. Rebuilt `BODY` as a single `printf '**What'\''s new**\n%s'` so the separator survives.

---

PR #3185, opened 2026-06-24
feat: add password reset flow to web and mobile

marcodejongh:
## Summary

- **Web**: Forgot-password and reset-password pages with full i18n (en-US, es, fr), rate-limited API routes, bcrypt-secured credential update, enumeration-safe responses, and a 1h expiring single-use token via `verificationTokens`
- **Email**: Fastmail SMTP via nodemailer — `sendPasswordResetEmail()` wired to `forgot-password` API. Credentials live in 1Password (Boardsesh vault → "Boardsesh SMTP"); add to Vercel before enabling
- **Mobile**: Native forgot-password and reset-password screens under `app/auth/`, "Forgot password?" link on the login screen, `requestPasswordReset` / `resetPassword` in `src/lib/auth.ts`, Android App Link intent filter for `/auth/reset-password`

## Vercel env vars to set before merging

```
SMTP_HOST=smtp.fastmail.com
SMTP_PORT=465
SMTP_USER=           # from 1Password: Boardsesh vault → Boardsesh SMTP → username
SMTP_PASSWORD=       # from 1Password: Boardsesh vault → Boardsesh SMTP → app-password
EMAIL_FROM=          # same as SMTP_USER
```

## Mobile rollout note

The Android intent filter change in `app.config.ts` requires a native build — an OTA update alone won't capture reset-password deep links on Android. iOS users are already covered by `applinks:boardsesh.com`. Queue a new build after merge.

## Known limitations / follow-ups

These are systemic to the existing credentials-auth stack (not introduced by this feature), kept out of this PR to stay scoped. Worth a follow-up before/around production rollout:

- **Rate limiting is in-memory (best-effort only).** `checkRateLimit` is per-process, so on serverless each instance keeps its own counter and the 5/min limit isn't enforced globally. `getClientIp` also trusts the first `x-forwarded-for` entry, which is spoofable. Follow-up: shared store (Redis/Upstash) keyed on IP+email, with correct edge IP resolution (`x-vercel-forwarded-for` / last `x-forwarded-for` entry). Affects all auth routes.
- **No session invalidation on password change.** Sessions use the JWT strategy (required for the credentials provider), so there are no DB sessions to delete and existing JWTs stay valid until expiry. Revoking them needs a per-user token version checked in the NextAuth `jwt`/`session` callback, bumped on password change — a cross-cutting auth change.
- **Email lookups are case-sensitive system-wide.** Login, register, and forgot-password all use `eq(users.email, email)` with no normalization, so this flow is consistent with the rest of auth. Normalizing to lowercase should be done uniformly across register + login + reset (not just here, which would create mismatches).

## Test plan

- [ ] Web: visit `/auth/forgot-password`, submit email for a credentials-auth account, receive email from `marco@sent.com`, follow link to `/auth/reset-password?token=…&email=…`, set new password, sign in
- [ ] Web: confirm expired/used tokens return appropriate error
- [ ] Web: confirm non-existent email returns same generic message (enumeration-safe)
- [ ] Mobile: login screen shows "Forgot password?" link, tapping opens forgot-password screen
- [ ] Mobile: submit email → see confirmation message
- [ ] Mobile: deep link `https://boardsesh.com/auth/reset-password?token=…&email=…` opens reset screen on iOS and Android (after native build)
- [ ] i18n: `vp run check:i18n` passes
- [ ] Typecheck: `vp run typecheck` and `vp run typecheck:mobile` pass
- [ ] Tests: `vp test run --project web` passes (359 files, 4766 tests)

## Release Notes

Forgot your password? Now you can reset it — tap "Forgot password?" on the login screen and we'll email you a secure reset link. Works on web and mobile.

https://claude.ai/code/session_01URm6sHsgCiRNM26JncMfKZ

### DIFF
--- a/packages/mobile/app.config.ts
+++ b/packages/mobile/app.config.ts
@@ -315,6 +315,15 @@ export default ({ config, projectRoot }: ConfigContext): ExpoConfig & { newArchE
           ],
           category: ['BROWSABLE', 'DEFAULT'],
         },
+        {
+          action: 'VIEW',
+          autoVerify: true,
+          data: [
+            { scheme: 'https', host: 'www.boardsesh.com', pathPrefix: '/auth/reset-password' },
+            { scheme: 'https', host: 'boardsesh.com', pathPrefix: '/auth/reset-password' },
+          ],
+          category: ['BROWSABLE', 'DEFAULT'],
+        },
       ],
       // Keep the legacy predictive back gesture OFF. Enabling it
       // (android.predictiveBackGestureEnabled: true) currently breaks
--- a/packages/mobile/app/auth/_layout.tsx
+++ b/packages/mobile/app/auth/_layout.tsx
@@ -6,8 +6,10 @@ export default function AuthLayout() {
   return (
     <Stack screenOptions={{ ...screenOptions, headerShown: false }}>
       <Stack.Screen name="login" />
-      {/* register.tsx sets its own header (title + back chevron) via Stack.Screen. */}
+      {/* register, forgot-password, and reset-password set their own header via Stack.Screen. */}
       <Stack.Screen name="register" />
+      <Stack.Screen name="forgot-password" />
+      <Stack.Screen name="reset-password" />
     </Stack>
   );
 }
--- a/packages/mobile/app/auth/forgot-password.tsx
+++ b/packages/mobile/app/auth/forgot-password.tsx
@@ -1,6 +1,6 @@
 import { useState } from 'react';
 import { KeyboardAvoidingView, Platform, ScrollView, StyleSheet, Text, View } from 'react-native';
-import { Stack } from 'expo-router';
+import { Stack, useRouter } from 'expo-router';
 import { useTranslation } from 'react-i18next';
 import { EMAIL_REGEX } from '../../src/lib/auth-validation';
 import { requestPasswordReset } from '../../src/lib/auth';
@@ -15,6 +15,7 @@ import { track } from '../../src/lib/analytics';
 export default function ForgotPasswordScreen() {
   const { t } = useTranslation('auth');
   const theme = useTheme();
+  const router = useRouter();
 
   const [email, setEmail] = useState('');
   const [emailError, setEmailError] = useState<string | null>(null);
@@ -74,6 +75,12 @@ export default function ForgotPasswordScreen() {
               <Text style={[styles.successText, { color: theme.systemColors.label }]}>
                 {t('forgotPassword.toasts.success')}
               </Text>
+              <Button
+                title={t('forgotPassword.back')}
+                onPress={() => router.replace('/auth/login')}
+                variant="text"
+                size="large"
+              />
             </View>
           ) : (
             <>
--- a/packages/mobile/app/auth/forgot-password.tsx
+++ b/packages/mobile/app/auth/forgot-password.tsx
@@ -88,7 +88,7 @@ export default function ForgotPasswordScreen() {
                     setEmail(text);
                     if (emailError) setEmailError(null);
                   }}
-                  placeholder="your@email.com"
+                  placeholder={t('login.placeholders.email')}
                   keyboardType="email-address"
                   autoCapitalize="none"
                   autoCorrect={false}
--- a/packages/mobile/app/auth/forgot-password.tsx
+++ b/packages/mobile/app/auth/forgot-password.tsx
@@ -30,7 +30,7 @@ export default function ForgotPasswordScreen() {
     if (!canSubmit) return;
 
     if (!EMAIL_REGEX.test(trimmedEmail)) {
-      setEmailError(t('login.validation.emailInvalid'));
+      setEmailError(t('forgotPassword.validation.emailInvalid'));
       return;
     }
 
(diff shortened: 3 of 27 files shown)


marcodejongh:
Addressed the latest review round in `fabf0344e` (+ earlier `b4b02b365`):

**Fixed**
- **SMTP failures now reported to Sentry** — the swallowed email error in `forgot-password/route.ts` emits `Sentry.captureException` (tags `area:auth, flow:forgot-password, phase:send-email`) so a misconfigured SMTP credential surfaces in observability. HTTP response stays generic (no user enumeration). Added a test asserting both.
- **Tests** — added the SMTP-failure-still-200 path, a wrong-email/token → 400-without-mutation test, plus unit tests for `hashResetToken` / `getPasswordResetIdentifier` / `consistentDelay` and `sendPasswordResetEmail` (link format, email encoding, HTML escaping).
- **Email-verification side effect** documented in `reset-password/route.ts` (completing a reset proves inbox control; scoped to NULL `emailVerified`).

**Not changed (with rationale)**
- **Android App Links / assetlinks.json** — false positive. Digital Asset Links verification is **per-domain, not per-path**. `/.well-known/assetlinks.json` is already served (`app/.well-known/assetlinks.json/route.ts`) with `delegate_permission/common.handle_all_urls` for `com.boardsesh.app` on `boardsesh.com`/`www.boardsesh.com`. The existing `/join` autoVerify filter relies on the same file; adding `/auth/reset-password` on the same hosts needs no new assetlinks entry.
- **In-memory rate limiting** — known limitation, acknowledged inline with `// NOTE:`. Moving to Redis/Upstash is out of scope for this PR.
- **Transporter singleton on credential rotation** — low-likelihood edge case (requires secret rotation without a cold start); leaving as-is.

marcodejongh:
Addressed the latest review round in `2dac8729e`:

**Fixed**
- **Mobile `reset-password` missing max-length** — now enforces `PASSWORD_MAX_LENGTH` (128), matching web.
- **Deep-link dead-end** — both mobile screens now show a "Back to Login" affordance (invalid-link state, post-submit, and forgot-password success), so a deep-link cold start without a back stack isn't a trap. `router.replace('/auth/login')`.
- **Token not URL-encoded** — `encodeURIComponent(token)` in the reset URL, consistent with the email param. (No-op for `crypto.randomUUID()` output, but correct.) Escaping test updated to a single-quote vector that survives URL-encoding but must be HTML-escaped.
- **Token-rotation ordering** — added a test asserting the transaction deletes the old token before inserting the new one (`invocationCallOrder`).

**Not changed (with rationale)**
- **`getClientIp` IP spoofing** — this is pre-existing shared code (`rate-limiter.ts`, untouched by this PR) used by every auth route. Changing its `x-forwarded-for` semantics is out of scope and would alter behavior for register/verify-email/etc. It's part of the already-acknowledged best-effort/serverless rate-limit limitation; moving to Redis + correct edge IP resolution is a separate change.
- **"Redundant" user SELECT** — not redundant: the token row only carries the email-derived identifier, not the user id the transaction needs (`userCredentials.userId`, `users.id`). It also guards the account-deleted-within-the-window case. Added a comment to make that explicit.
- **Brittle mock dispatch** — the select-key dispatch keys on stable projections (`{ id }` vs `{ expires }`); call-order dispatch would be order-fragile in its own way. Leaving as-is.

marcodejongh:
Addressed in `1dc7c5e51`:

- **i18n key bug** — fixed: mobile forgot-password now uses `forgotPassword.validation.emailInvalid` (the key this PR added) instead of `login.validation.emailInvalid`.
- **Rate limiter** — added a **Known limitations / follow-ups** section to the PR description tracking the serverless in-memory limitation + correct edge IP resolution as a required pre-production follow-up (Redis/Upstash, IP+email keying). Kept out of this PR to isolate the shared `rate-limiter.ts` change from the feature.

The two flagged test gaps are already covered / not applicable:
- OAuth-only forgot-password path → `returns generic response and does not send email for OAuth-only account` (route.test.ts).
- "SMTP failure in reset-password" → the reset-password route sends no email; its failure path is the transaction-fails→500 test.

marcodejongh:
Triaged this round — no new actionable bug in the feature code. Details:

**Tracked as follow-ups (systemic, not introduced here)** — added to the PR description under *Known limitations / follow-ups*:
- **Session invalidation after reset** — sessions use the **JWT strategy** (`strategy: 'jwt'`, required for the credentials provider), so there are no DB sessions to delete; existing JWTs stay valid until expiry. Revoking them needs a per-user token version checked in the NextAuth `jwt`/`session` callback — a cross-cutting auth change, not specific to this feature.
- **`getClientIp` spoofing / rate limiter** — documented; needs Redis + correct edge IP resolution across all auth routes.
- **Email case-sensitivity** — login, register, and forgot-password all use `eq(users.email, email)` with no normalization, so this flow is **consistent** with the rest of auth. Lowercasing only here would break resets for users stored with capitals; it should be a uniform system-wide change.

**Fixed**
- **Flaky timing test** — widened the `consistentDelay` lower-bound margin (`72a8cdcd6`).

**Not changed (rationale)**
- **Dead i18n key** — false positive: `forgotPassword.validation.emailRequired` is used by the **web** screen (`forgot-password-content.tsx:33`); the catalog is shared web+mobile, and mobile relies on its `canSubmit` guard.
- **`confirmPassword` over the wire** — it's the same value over TLS, validated server-side as defense-in-depth; not a "leak." Removing it would change the API contract for both clients for no security gain.
- **Auth constants duplication** — `EMAIL_REGEX`/`PASSWORD_*` mirror is pre-existing and intentional (mobile `auth-validation.ts` documents it); extracting to `packages/shared` is a separate refactor of pre-existing code.
- **Brittle mock dispatch** — the select-key dispatch keys on stable projections; call-order dispatch is order-fragile in its own way.

PR #3199, opened 2026-06-25
feat(mobile): add tester feature-flag override UI

marcodejongh:
## Summary

Adds a tester-only **Feature Flags** screen to mobile Settings (More → Development → Feature Flags) that lets an admin/tester force any known flag **On**, **Off**, or back to its remote **Default** directly on a device.

Until now the mobile app resolved flags only from PostHog (live) and build-time `EXPO_PUBLIC_*` env vars — so testing a gated feature like `strava-integration` meant cutting a new build or changing a PostHog rollout. This adds a local override layer at the top of the merge:

`PostHog < static env < local tester override`

Overrides persist on-device (AsyncStorage) and survive restarts.

### What's in here
- **Override store** (`feature-flag-overrides.ts`) — AsyncStorage-backed, reactive via `useSyncExternalStore`, mirroring the proven `session-recording-preference.ts` pattern (referentially-stable snapshot, promise-singleton one-time load, write-through mutators).
- **Provider wiring** — `FeatureFlagsProvider` folds overrides into its merge with top precedence, and now exports a `FEATURE_FLAG_DEFINITIONS` catalog (key/label/description) that drives both the PostHog read and the new screen. Add a flag once, there.
- **Screen** (`FeatureFlagsScreen.tsx`) — per-flag Default/On/Off `SegmentedControl` with a live-default + effective-value footnote, plus a "Reset all overrides" action. Styled after `ChannelSwitcherScreen`; tester-only copy is hardcoded English with `i18n-ignore`.
- **Entry point** — a gated `ListRow` in the Development section, shown for `__DEV__ || profile.isTester` (reusing the same `isTester` gate as the OTA Channel Switcher). The mobile profile has no separate admin flag, so `isTester` covers admins and testers.

## Release Notes

- Testers can now flip feature flags on or off right in the app, no new build needed

## Test plan

- `vp run test:mobile` (feature-flag specs) — 13 passing, covering override precedence and store set/clear/clearAll + persistence round-trip
- `vp check` — clean
- Manual (next): in a dev build, More → Development → Feature Flags → force `strava-integration` On and confirm the Strava integration surfaces; reset to Default and confirm it reverts; relaunch and confirm the override persists.

https://claude.ai/code/session_019HGobXfk42HbHjru8yJNBk

---
_Generated by [Claude Code](https://claude.ai/code/session_019HGobXfk42HbHjru8yJNBk)_

### DIFF
--- a/packages/mobile/app/(tabs)/profile/_layout.tsx
+++ b/packages/mobile/app/(tabs)/profile/_layout.tsx
@@ -24,6 +24,8 @@ export default function ProfileLayout() {
       <Stack.Screen name="dev-servers" options={{ title: t('mobile.more.metroServersTitle') }} />
       {/* i18n-ignore-next-line — tester-only screen */}
       <Stack.Screen name="channel-switcher" options={{ title: 'OTA Channel Switcher' }} />
+      {/* i18n-ignore-next-line — tester-only screen */}
+      <Stack.Screen name="feature-flags" options={{ title: 'Feature Flags' }} />
       <Stack.Screen name="delete-account" options={{ title: tSettings('deleteAccount.title') }} />
     </Stack>
   );
--- a/packages/mobile/app/(tabs)/profile/feature-flags.tsx
+++ b/packages/mobile/app/(tabs)/profile/feature-flags.tsx
@@ -1,0 +1,5 @@
+import { FeatureFlagsScreen } from '../../../src/components/FeatureFlagsScreen';
+
+export default function FeatureFlagsRoute() {
+  return <FeatureFlagsScreen />;
+}
--- a/packages/mobile/app/(tabs)/profile/more.tsx
+++ b/packages/mobile/app/(tabs)/profile/more.tsx
@@ -63,9 +63,15 @@ export default function MoreScreen() {
   // it can actually work; testers on a release build get the OTA channel switcher.
   const showDevServerSwitcher = isDevLauncherAvailable();
 
-  // Don't render an empty "Development" section header when neither tool applies
+  // Feature-flag overrides work in every build (dev forces them in place of the
+  // disabled PostHog read; release builds layer them on top), so show the entry
+  // wherever the dev section is allowed — for testers and in dev.
+  const showFeatureFlags = __DEV__ || Boolean(profile?.isTester);
+
+  // Don't render an empty "Development" section header when no tool applies
   // (e.g. a tester on a release build with updates disabled).
-  const showDevSection = (__DEV__ || Boolean(profile?.isTester)) && (showDevServerSwitcher || showChannelSwitcher);
+  const showDevSection =
+    (__DEV__ || Boolean(profile?.isTester)) && (showDevServerSwitcher || showChannelSwitcher || showFeatureFlags);
 
   // Whether the bundled changelog has an entry the user hasn't opened yet — drives
   // the "New" pill on the What's New row. Re-read every time this screen regains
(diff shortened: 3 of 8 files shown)


marcodejongh:
Addressed the review feedback in `8376ffc`:

- **Codex P2 — route guard.** `FeatureFlagsScreen` now gates itself: in a non-dev build it waits for the profile and redirects non-testers back to More. Hiding the row was never a guard, and since overrides apply globally via `FeatureFlagsProvider`, a deep link could have let a non-tester force rollout-only flags. `__DEV__` still passes through.
- **PostHog read** hoisted into a `useMemo` over the exported `FEATURE_FLAG_KEYS` (no per-render array allocation or read).
- **Spacing** in the `StyleSheet` now uses theme tokens.
- **Test** added for `useFeatureFlagOverrides` reactivity (the `useSyncExternalStore` subscription path).

Also retitled the PR to a Conventional Commit so `commit-lint` passes.

---

marcodejongh:
Addressed the latest review notes in `d096437`:

- **Override-load flash** — `FeatureFlagsScreen` now waits on the store's `loaded` flag (same `ActivityIndicator` pattern as the profile guard), so a cold open no longer shows every flag at Default before the saved choices load.
- **Spurious empty write** — `clearAllFeatureFlagOverrides` returns early when there's nothing in memory to clear, so it no longer persists `{}`.
- **Test isolation** — added `resetFeatureFlagOverridesForTests` (resets state + the cached `loadPromise`) and reset the mock AsyncStorage in the provider test's `afterEach`. `vi.resetModules()` can't be used here because the provider is statically imported — it would bind the provider and the store mutators to different module instances.

One I'm leaving as-is: **`marginTop: 2`** in the StyleSheet. The token-only rule in CLAUDE.md is the web/MUI rule; mobile uses `StyleSheet.create` with raw px for sub-token spacing, and `2` matches existing precedent (`SwitchRow`'s `gap: 2`). The smallest token is `4`, which would change the label/description spacing. Happy to switch if you'd prefer strict tokens here.

---

marcodejongh:
Addressed in `81ed62a`:

- **`FEATURE_FLAG_KEYS` literal type** — the catalog is now `as const satisfies readonly FeatureFlagDefinition[]`, and `FEATURE_FLAG_KEYS` / a new `FeatureFlagKey` type derive from it, so the key union is preserved instead of widening to `string`.
- **Whole-object validator** — added a comment documenting that `isOverridesRecord` rejects the entire persisted bag on any non-boolean value (all-or-nothing fail-safe).

Leaving as-is: **`resetFeatureFlagOverridesForTests` in the bundle.** It's a test-only export with no production importer (same shape as the existing `loadFeatureFlagOverrides` / `loadSessionRecordingEnabled` test-used exports). A `__DEV__` guard wouldn't remove the symbol while the test statically imports it, and it's a few inert lines — not worth adding prod-path branching. Happy to revisit if you'd rather it be guarded.

---

marcodejongh:
Addressed the ⚠️ items in `19a66c8`:

- **Load race (data loss)** — `loadFeatureFlagOverrides` now merges the persisted bag with any override a mutator wrote during the storage read (persisted = base, racing keys win on top) instead of discarding the persisted flags. It also re-persists the merged bag on a detected race, so storage stays consistent for the next cold boot — the racing mutator's own `persist()` had only written its single key. Covered by a new test that fires a mutator mid-load.
- **`clearAllFeatureFlagOverrides`** now `removePreference(STORAGE_KEY)` instead of writing `{}`. New test asserts the key is gone.
- **`resetFeatureFlagOverridesForTests`** guarded behind `process.env.NODE_ENV === 'test'` — Metro inlines `NODE_ENV` so the body is dead-code-eliminated from release bundles; Vitest sets it to `test`.
- **Load-failure recovery** — added a test that a rejected read leaves `hasLoaded` false so the next attempt retries (rather than staying stuck until restart).

Tests: 17 passing. Leaving as-is: the **`baseFlags` "Live default" footnote** can read `not set` if opened in the split-second before PostHog resolves on a cold launch. It's a labelled hint on a tester-only screen, and making it live would mean re-adding the per-render PostHog read that an earlier review asked me to memoize — net negative for that tradeoff. Happy to wire a PostHog subscription if you'd rather it update live.

---

PR #3201, opened 2026-06-25
fix(mobile): logbook filter sheet close + grade rail + collapsed sections

azuttre:
Follow-up UX fixes for the mobile logbook filter sheet (shipped in #3179, flag-gated by #3200, migrated to the native sheet in #3197), found while testing on TestFlight.

## What changed
- **Sheet wouldn't close, and applies on close now.** The raw `BottomSheetModal` was missing `enablePanDownToClose` (the `Sheet`/`ModalSheet` wrappers default it on), and the Apply button's footer was being pushed off-screen by the migrated native sheet — so once open there was no way to close *or* apply. Removed the Apply button entirely: filters/sort apply when the sheet is dismissed (swipe down or tap the scrim). A ref commits the latest draft on dismiss, so nothing is lost to a stale closure.
- **Grade rail opened mid-scrolled / looking cropped.** The shared `GradeRangeRail` auto-centers on the climber's grades on mount, which in the logbook (no send-grade seed, no selection) landed mid-rail with a half-cut chip. Added a `centerOnEmpty` prop (default `true`, so climb search is unchanged); the logbook passes `false` and only auto-centers when a grade is actually selected.
- **Refine + Advanced started expanded.** Dropped `defaultExpanded` on Refine so both sections open collapsed (and stay collapsed on Reset).
- Removed the now-dead footer styles, the unused `brandColors`, and the orphaned `mobile.logbook.apply` key across all three locales.

## Testing
- `vp run typecheck:mobile`, lint, format clean
- `vp test run --project mobile logbook` — 42 pass; i18n catalog parity — 92 pass
- `vp run check:mobile-bundle` — iOS + Android bundle clean

## Release Notes
Smoothed out the logbook filter: swipe down or tap outside to close it and your filters and sort apply automatically (no more Apply button), the grade picker opens at the start instead of jumping to the middle, and the Refine and Advanced sections start tucked away so it opens tidy.

### DIFF
--- a/packages/mobile/src/components/grade/GradeRail.tsx
+++ b/packages/mobile/src/components/grade/GradeRail.tsx
@@ -73,6 +73,12 @@ type GradeRangeRailProps = {
   grades: readonly Grade[];
   bound: GradeBound;
   sendDifficultyIds?: readonly number[];
+  /**
+   * Auto-center the rail on mount when there is no grade selection. True (the
+   * default) surfaces the climber's typical grades; the logbook filter passes
+   * false so an unset rail opens at the easiest grade instead of mid-scrolled.
+   */
+  centerOnEmpty?: boolean;
   onChange: (next: GradeBound) => void;
   /**
    * Asked to dismiss itself (swipe the handle down, completed range, or cleared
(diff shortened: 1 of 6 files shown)


marcodejongh (approved):

azuttre:
Tested the merged build on the production OTA channel and the sheet works great now: swipe down or tap outside to close, filters apply on close (no Apply button), the grade rail opens at the start instead of mid-scrolled, and Refine/Advanced start collapsed. Thanks for the quick turnaround on the native-sheet migration.

PR #3202, opened 2026-06-25
feat(mobile): hardest sorts by effective grade + rows show consensus grade

azuttre:
Follow-up to the logbook work: makes the Hardest sort read correctly while scrolling, and shows both grades on each row. Branches off `main`; touches the backend sort + the row, which don't overlap the filter-sheet PR (#3201), so they're conflict-free in either merge order.

## The problem
Hardest sorted by **consensus grade**, but each row shows the climber's **logged** grade. So a climb you logged V8+ that the crowd calls V9 sorted among the V9s — and looked out of order, because the consensus driving the sort was invisible.

## What changed
- **Hardest now sorts by your effective grade** (your logged grade, or the consensus when you didn't grade it) → ascent date. The grade the row shows *is* the sort key, so the list reads in order while scrolling. This reverses the consensus-first ordering from #3179.
- **Rows show both grades.** Your logged grade is the primary; the community consensus appears as a small `people`-marked secondary, **only when the crowd disagrees**. An ungraded tick shows the consensus as its grade, marked with the same glyph so it's clear it's the crowd's, not yours.
- `ClimbListItemContent` gains opt-in `consensusGrade` / `gradeIsConsensus` props (default off — climbs search and every other caller are unchanged). The logbook row passes them and handles the no-board-art fallback too.

## Testing
- `vp run typecheck:mobile|backend|shared`, lint, format — clean
- `vp test run --project backend tick-queries` — 33 pass (hardest now `D, A, E, B, C`: effective-grade desc with date breaking ties)
- mobile logbook 42 / climb-list-row 10 / shared logbook 14 — pass
- `vp run check:mobile-bundle` — iOS + Android clean
- Row display (presentation) verified on device via the OTA channel

## Release Notes
Sorting by Hardest now reads top-to-bottom the way you'd expect, and every logbook entry shows both grades — the grade you logged, plus the community consensus beside it when the crowd disagrees (and on climbs you never graded yourself).

### DIFF
--- a/packages/backend/src/__tests__/tick-queries.test.ts
+++ b/packages/backend/src/__tests__/tick-queries.test.ts
@@ -262,6 +262,35 @@ describe('tickQueries — behavior fixes', () => {
 
       expect(result.items.map((item) => item.climbName)).toEqual(['Easy', 'Mid', 'Hard']);
     });
+
+    it('interleaves a graded and an ungraded tick by effective grade', async () => {
+      // The graded tick logs hard (12); the ungraded one's consensus is easy (5).
+      // effective-grade asc puts the ungraded (5) first — the old loggedGrade asc
+      // would have floated the NULL-logged ungraded tick to the end instead.
+      const graded = CLIMB_PREFIX + 'easiest-mixed-graded';
+      const ungraded = CLIMB_PREFIX + 'easiest-mixed-ungraded';
+      await insertClimb(graded, 'Graded');
+      await insertClimb(ungraded, 'Ungraded');
+      await insertBoardClimbStats({ climbUuid: graded, displayDifficulty: 8 });
+      await insertBoardClimbStats({ climbUuid: ungraded, displayDifficulty: 5 });
+      await insertTick({
+        uuid: 'tick-easiest-mixed-graded',
+        climbUuid: graded,
+        climbedAt: '2026-02-01 10:00:00',
+        status: 'send',
+        difficulty: 12,
+      });
+      await insertTick({
+        uuid: 'tick-easiest-mixed-ungraded',
+        climbUuid: ungraded,
+        climbedAt: '2026-02-01 10:00:00',
+        status: 'send',
+      });
+
+      const result = await callUserAscentsFeed(TEST_USER_ID, { sortBy: 'easiest' });
+
+      expect(result.items.map((item) => item.climbName)).toEqual(['Ungraded', 'Graded']);
+    });
   });
 
   describe('userAscentsFeed — flashOnly filter', () => {
--- a/packages/backend/src/__tests__/tick-queries.test.ts
+++ b/packages/backend/src/__tests__/tick-queries.test.ts
@@ -175,14 +175,14 @@ describe('tickQueries — behavior fixes', () => {
   });
 
   describe('userAscentsFeed — hardest sort', () => {
-    it('orders by consensus desc, then effective grade (logged, else consensus) desc', async () => {
-      // consensus / logged -> effective grade:
+    it('orders by effective grade (logged, else consensus) desc, date breaking ties', async () => {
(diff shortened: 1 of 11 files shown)


azuttre:
Thanks, good catches — addressed in the latest commit.

- **Sort asymmetry (easiest vs hardest):** aligned `easiest` to `effectiveGrade` asc, mirroring `hardest`, so an ungraded tick sorts by its consensus instead of floating out on a NULL logged grade. Added a backend test that seeds all-ungraded climbs and asserts `Easy, Mid, Hard`.
- **Duplicate styles:** collapsed `fallbackGradeRow`/`fallbackConsensusRow` and `gradeRow`/`consensusRow` into one shared `iconGradeRow` per file.
- **UI tests for the new props:** extracted the dual-grade decision into a pure `deriveLogbookGradeDisplay(logged, consensus)` and unit-tested all four branches (consensus matches → no secondary; differs → secondary; ungraded → consensus-sourced glyph; no consensus → neither). The rendering itself is trivial conditional JSX driven by that decision, and it's verified on device via the OTA channel.
- **Hardcoded icon sizes / gaps:** left as-is intentionally — raw `Icon` sizes are the established convention across this codebase (e.g. the sibling status glyph uses `size={16}`), and the 1/3 px gaps are sub-token nudges like the existing `gap: 2`/`gap: 6` in these same rows. There's no icon-size or sub-4px spacing token to reach for, so introducing one just here would be the inconsistency.

azuttre:
Thanks — addressed two of three:

- **Helper belongs in shared:** moved `deriveLogbookGradeDisplay` into `@boardsesh/logbook` (`packages/shared/logbook/src/grade-display.ts`); the mobile row imports it from the package now, and its unit test moved alongside. Web can reuse it directly when it gets the dual-grade display.
- **Easiest test only covered all-ungraded:** added a mixed graded/ungraded case — a logged-hard tick (12) and an ungraded one whose consensus is easy (5) — asserting `Ungraded, Graded`. That's the order `effectiveGrade asc` produces and the old `loggedGrade asc` (nulls last) would not.
- **`iconGradeRow` duplicated across two files:** left as-is. It's a generic 3-property flex row (`row` + `center` + `gap: 3`), the kind defined ad-hoc throughout the RN components; there's no shared mobile-styles module, and introducing one for three properties would couple two unrelated components more than the drift risk warrants.

azuttre:
- **LogbookRow integration test:** added `logbook-row.test.tsx`, spying on `ClimbListItemContent` (the `SessionTickRow.test.tsx` pattern). Three cases: ungraded ascent gets `gradeIsConsensus: true` / no secondary, a disagreeing ascent gets `consensusGrade` set, a matching ascent gets neither. This also exercises the row's label formatting, not just the pure helper.

- **Multi-line comments "violate CLAUDE.md":** I don't think this rule exists. I grepped `CLAUDE.md` and `docs/` for any comment-style rule and found nothing, and `/**` block comments are pervasive in the codebase (240 files under `packages/mobile/src/components` alone). The JSDoc documents a non-obvious decision helper, consistent with the surrounding code, so I'm leaving it. Happy to trim if you can point me at the actual rule.

- **Sub-token gaps (`gap: 1` / `gap: 3`):** held. These match the existing raw gaps in the same rows (`trailing` uses `gap: 2`, `rightSection` uses `gap: 6`); the smallest spacing token is `4`, which is too large for a hairline grade-stack or a tight icon-text gap. A token doesn't fit here, and using one only in these rows would be the inconsistency.

marcodejongh (approved):

azuttre:
Tested the merged build on the production OTA channel and the sort + dual grades work really well. Hardest reads top-to-bottom in order while scrolling now, and showing my logged grade with the consensus (people icon) beside it when they differ is exactly what I wanted. Ungraded climbs showing the consensus marked with the icon is a nice touch too.

PR #3211, opened 2026-06-26
fix(mobile): serialize native sheet present/dismiss to stop the iOS freeze

marcodejongh:
## Summary

Fixes the recurring whole-app freeze on iOS where the UI renders but goes dead to touch (only a force-quit recovers). It reproduces after a native bottom sheet has been opened — close a sheet then open another, hand off between sheets, or open the sidebar and tap a row.

This generalises the narrow per-row deferral this PR originally shipped into a proper, systemic fix.

### Root cause (confirmed on-device)

`@expo/ui/community/bottom-sheet` is a SwiftUI `.sheet(isPresented:)` wrapper with two properties that combine badly:

1. **No serialization** — every sheet presents off the **same** root-window view controller, and the library never coordinates them. UIKit forbids presenting/dismissing a VC while another transition is in flight on the same presenter; overlapping them **deadlocks UIKit**.
2. **`dismiss()` lies about timing** — it fires its close callbacks (`onDismiss`/`onChange(-1)`) **synchronously**, before the native animation finishes. So any "now it's safe to present the next thing" logic built on that callback acts ~16 ms in while the real dismiss runs ~400–500 ms — it works only by luck (the flaky-freeze signature).

### The fix

- **`SheetPresentationProvider`** — a pure-JS coordinator (`packages/mobile/src/providers/sheet-presentation-provider.tsx`) that serialises every native present/dismiss **per presenter group**: one transition in flight at a time, a queued present waits out the prior dismiss's settle window, and presenting B while A is open auto-sequences `dismiss(A) → settle → present(B)`. Dedicated unit tests.
- **`useManagedSheet`** — the bridge the `ModalSheet`/`Sheet` wrappers and every raw-rendering consumer (`QueueSheet`, `BoardSheet`, `LogAscentSheet`, the filter/picker/session sheets, …) use instead of hand-rolled `isPresentedRef` effects.
- **drawer-host** keeps data-backed sheets (log-ascent / add-to-playlist / beta) **mounted through their dismiss animation** (`useDeferredSheetData`) so the native Host never tears down mid-animation.
- **User drawer → `transparentModal` route** — replaces the React Native core `<Modal>` (a second native presentation system) with a single-presentation route. `FeedbackSheet` stays at the provider root and is presented one frame after the route unmounts (an `@expo/ui` sheet rendered inside a transparentModal would render *behind* it).
- Reverts the exploratory BoardSheet mount/unmount experiment.

JS-only — rides OTA, no native change.

### Follow-ups (tracked, not blocking)

- A one-line `@expo/ui` patch to forward the SwiftUI layer's true post-animation `onDismiss` so the coordinator's settle is exact instead of a 550 ms timer (optimization; the coordinator is correct on the timer baseline).
- An import lint-guard so a future sheet can't bypass the coordinator.

## Release Notes

- Fixed an iOS freeze where the app could stop responding to taps after opening and closing sheets (board history, the queue, filters) or after using the sidebar.

## Test plan

- `vp run test:mobile` — **2660 tests pass**, incl. new coordinator unit tests (handoff serialization, early settle, user-close, no-op close, rapid-toggle coalescing, group isolation) and updated consumer/user-drawer regressions
- `vp check` on changed files — clean (0 warnings)
- **On-device verified** (iPhone 17 Pro, iOS 26): the previously-freezing flows (BoardSheet → add-to-playlist / player, sidebar → Report a bug → Change board, pan-down-while-opening-another-sheet) no longer freeze
- Reviewed by a multi-agent adversarial pass over the coordinator + consumers + the new route; two real user-drawer defects it surfaced (a re-introduced feedback-row freeze and an invisible-backdrop touch-sink) are fixed in this branch

https://claude.ai/code/session_01BjETzDReFapvth6uTL9VsX

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -286,11 +286,11 @@
         "react-native-pager-view": "^8.0.2",
         "react-native-paper": "^5.15.3",
         "react-native-qrcode-svg": "^6.3.15",
-        "react-native-reanimated": "4.3.1",
+        "react-native-reanimated": "4.5.0",
         "react-native-safe-area-context": "~5.7.0",
         "react-native-screens": "4.25.2",
         "react-native-svg": "15.15.4",
-        "react-native-worklets": "0.8.3",
+        "react-native-worklets": "0.10.0",
       },
       "devDependencies": {
         "@bacons/apple-targets": "^4.0.7",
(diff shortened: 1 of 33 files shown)


marcodejongh:
Addressed the review in c6bdb20:
- Added deferral regression tests for the feedback sheet present (`Rate Boardsesh`) and the two originally-deferred routes (`Change board`, `My boards`).
- Kept `setFeedbackMode` immediate (not inside the deferred callback) on purpose, with a comment: setting it now guarantees the sheet has re-rendered with the new mode before the deferred `present()` fires a frame after the close animation — moving it into the deferred callback would risk `present()` running in the same tick before the re-render.

PR #3212, opened 2026-06-26
feat(mobile): OKHSL hold colour picker, octagon shape, accessibility sub-page

### LINKED ISSUE #3208: Accessibility: custom hold selector improvements
The current custom hold selector accessibility feature has some accessibility issues

Current interface screenshot

1. Add a proper colour selector
    - Most people struggle to find colours they want when adjusting RGB values by hand, especially for choosing the initial ballpark colour
    - Ideally that would be using [OKHSL](https://bottosson.github.io/misc/colorpicker/#09f8fb) or another perceptually linear colour picker, so choosing colours is intuitive
2. Add a sixth shape, so there’s no gap
    - I propose an octagon, as it’s almost universally recognised as a stop sign

marcodejongh:
## Summary

Implements #3208 — accessibility custom hold-selector improvements, for colour-blind climbers who recolour/reshape the start/mid/finish/foot hold markers.

**1. A proper colour picker (OKHSL), tuned for colour blindness.** The old RGB 0–255 text boxes are gone. The per-role picker is now a **Lightness-first OKHSL slider stack** (Lightness → Saturation → Hue) with live OKHSL gradient tracks, plus a hex field for precise entry. Lightness leads because it's the channel essentially all colour-blind people perceive reliably, and OKHSL's L axis is perceptually uniform *and* independent of hue/saturation — so a chosen lightness stays put while exploring colour. Each slider is an `accessibilityRole="adjustable"` for VoiceOver/TalkBack.

  - Decided against `@expo/ui`'s `ColorPicker` (iOS-only, standard RGB/HSB, no OKHSL) and against a native module (no OKHSL on either platform). The picker is pure-JS (reanimated/Skia not needed — `react-native-svg` + `PanResponder`), so it's cross-platform and ships over OTA. OKHSL↔sRGB math is ported from Björn Ottosson's reference into `packages/mobile/src/lib/okhsl.ts` with unit tests.

**2. A 6th marker shape: octagon.** Fills the empty slot in the shape grid (stop-sign, universally recognised). Added to both render paths — the JS SVG renderer and the native Rust renderer (rebuilt iOS xcframework + Android `.so`). A forward-compat `#[serde(other)] Unknown` variant means an older native binary can't fail the *whole* board render on a shape it doesn't know (it falls back to a circle).

**3. Accessibility settings moved to their own Settings sub-page** (More → Accessibility), giving the picker and previews room to breathe.

**4. Colour-blind aids** (all requested): an all-4-roles compare strip, a **CVD-simulation toggle** (deuteranopia / protanopia / tritanopia, Machado 2009 matrices), and a **live board preview** that renders the active board with a climb lighting all four roles.

Closes #3208.

## Release Notes

- Pick hold colours with a real colour picker — slide lightness, saturation, and hue instead of typing numbers
- Built for colour blindness: a lightness-first picker, plus a toggle to preview your colours through red-green and blue-yellow colour blindness
- New octagon marker shape, so every hold role can have its own distinct shape
- Accessibility hold settings now live on their own page, with a live preview on your board

## Test plan

- `vp check` (changed files) — no warnings/errors
- `vp run test:mobile` — 2650 passed (incl. new OKHSL round-trip + CVD tests)
- `cargo test` (board-renderer-core) — 15 passed (incl. octagon + forward-compat serde)
- Native binaries rebuilt via `scripts/build-native-renderer.sh`; Android `.so` confirmed 16 KB-aligned
- i18n catalog-completeness — 28 passed (en-US/es/fr parity)

### Notes
- The board *photo* in the preview is not CVD-simulated (only the markers) — that's the surface that matters for telling roles apart.
- The WASM/web renderer artifact is intentionally **not** rebuilt: web doesn't expose marker-shape selection, and the committed `pkg/` already predates the shapes feature. The Rust source change is forward-compatible.
- Octagon requires the rebuilt native binaries, so it ships in a new native build (the Expo fingerprint changes), not a pure JS OTA.

https://claude.ai/code/session_01LCNrfUN6hEpnsvV2E6CzRL


marcodejongh:
Thanks for the review — addressed in 712d9f1:

- **French i18n typo** — fixed `deuteranopiaShort` to `Deutan` (dropped the stray Spanish accent).
- **Unnecessary exports** — `srgbTransferFunction`/`srgbTransferFunctionInverse` are now module-private; nothing imports them externally after the CVD simulation moved to the sRGB gamma domain.
- **seedKey/seedCounter → React `key`** — `OkhslColorPicker` now re-seeds by remounting via `key` per opened role; removed the `seedKey` prop, the `seededRef` guard, the reseed effect, and the post-render flash.

Deferred, with rationale:
- **Move `okhsl.ts`/`cvd-simulation.ts` to `packages/shared/`** — the approved plan deliberately keeps them mobile-local since mobile is the only consumer; the note there is to extract to a shared package if/when web needs perceptual colour. Happy to extract now if you'd prefer.
- **`BottomSheetTextInput` coupling in `OkhslColorPicker`** — left as-is; it has a single use site (inside the hold-picker sheet). Worth revisiting if the picker is reused outside a sheet.
- **Component tests** — the pure math is well covered; interaction tests (hex validation, CVD switching) would need the RN render harness + provider mocks. Can add a focused one if you want it in this PR.

Issue #3181, opened 2026-06-24
Tapped climb doesn't light up the wall when the board stops accepting BLE data


marcodejongh:
## What's happening

You tap a climb, but the board never lights up. Under a marginal Bluetooth link the app pushes LED data to the board, the board stops accepting writes mid-stream, and after a 5-second watchdog the app gives up on that write and drops it. There's no automatic retry or reconnect, so the climb silently fails to draw on the wall until you re-tap or reconnect.

The error string is "BLE write timed out waiting for the board to accept data" (`handled:true` — the JS auto-sender catches it, so it's not a crash).

## Evidence

- PostHog: https://us.posthog.com/project/412845/error_tracking/019ec3e3-51f3-7583-8d82-3e1e56ee73ba
- **Use the disaggregated figures, not the bundle.** This PostHog fingerprint (`019ec3e3-…`) is a garbage-bag that bundles ~11 unrelated errors through one generic `reportHandledError` call site. The whole bundle is 342 occ / 34 users / 49 sessions, but that's dominated by Google OAuth errors ("authorization attempt failed" 118 occ, "Unable to open Safari" 112 occ). **The BLE-write-timeout sub-error specifically is ~55 occ / 10 sessions / 5 users over 30d.**
- Volume is concentrated: 44 of 55 occurrences (80%) come from just 2 users (one 23 occ over 8 days, one 21 occ in a single ~35-min window).
- Platform: iOS. App version uniformly 2.0.0.
- **Not an iOS-26.5 regression.** One of the 5 affected users is on iOS 18.4.1, so this is a cross-OS marginal-link condition, not a CoreBluetooth backpressure regression on 26.5.

## Suspected cause

`packages/mobile/modules/live-activity/ios/BoardBleManager.swift` (lines ~976-1023). A chunk write parks when `peripheral.canSendWriteWithoutResponse` is false; the watchdog (`writeResumeTimeout = 5`, line 76) then calls `failQueuedWrites(.writeTimedOut)` (line 992). `failQueuedWrites` (1011-1023) clears the queue and fires the error completion with **no retry-after-ready and no auto-reconnect** — the write is just dropped. It surfaces to JS as a handled error in `packages/mobile/src/providers/bluetooth-provider.tsx` (lines 348-353, `reportHandledError` `source:ble-auto-send`).

[...]

PR #3219, opened 2026-06-26
fix(mobile): native iOS re-light supports MoonBoard (was Aurora-only)

marcodejongh:
## Summary

MoonBoard LED control over Bluetooth broke on iOS. Root cause is in the **live-activity native module**: the native re-light path (`displaySharedCurrentItemOnBleQueue` → `displayItemOnBleQueue` / `clearBoardOnBleQueue`) only knew the Aurora protocol.

- `BoardBleEncoding.makeAuroraPacket` has no `moonboard` entry in its `holdStateMap`, and a MoonBoard uses no LED placement map (it encodes straight from grid coordinates). So a native MoonBoard re-light wrote an **empty packet** and the wall stayed dark.
- This surfaced after the write-stall recovery (#3181, merged Jun 25): a marginal-link write stall now **cycles the connection natively** and re-lights via that Aurora-only path, while reporting `writeTimedOut` to JS (so JS keeps `isConnected` and does **not** re-send via `dispatchMoonboardPacket`). On **iOS + MoonBoard**, a single stall therefore left the wall dark indefinitely.
- The same gap meant the Live Activity Dynamic Island intents (next / previous / reconnect) never drove a MoonBoard either.

Android is unaffected — its adapter has no native recovery / native re-light and sends MoonBoard frames through JS.

### The fix

Add a native MoonBoard encoder, `BoardBleEncoding.makeMoonboardPacket`, mirroring the canonical JS encoder `packages/shared/ble-protocol/src/moonboard.ts`:

- grid-coordinate serial positions (11×18, odd columns reversed),
- role codes `42/43/44 → S/P/E` (others skipped),
- ASCII `l#…#` frame format,
- **no clear-all** — an empty / all-skipped frame produces no packet so the caller refuses to write and leaves the wall as-is.

`displayItemOnBleQueue` and `clearBoardOnBleQueue` branch on `boardName == "moonboard"`; the Aurora path is byte-for-byte unchanged.

## Release Notes

- MoonBoard on iPhone now recovers and re-lights the wall after a flaky Bluetooth moment instead of going dark, and the Dynamic Island climb controls drive a MoonBoard too.

## Test plan

- **Added Swift XCTests** (`LiveActivityWidgetTests`) asserting parity with the JS vectors: `p1r42p2r43 → l#S0,P35#`, unknown roles / out-of-range hold ids are skipped, and nothing-to-light yields no write. Wired `BoardBleEncoding.swift` + `BoardPlacementData.swift` into the `BoardseshTests` target (`ios-rn-ci.yml` runs them on a simulator).
- Confirmed the canonical JS suite (`moonboard.test.ts`, 17 tests) and the widget-target-drift guard pass locally via `vp test run`.
- `vp check` runs clean on the changed files (Swift isn't linted; the one repo-wide lint error is pre-existing in `packages/db/scripts/refresh-recommendations.ts`, untouched here).
- **Pending on-device (iOS + a real MoonBoard):** connect → tap a climb lights the wall; force a write stall → after native recovery the MoonBoard re-lights instead of going dark; Dynamic Island next/prev/reconnect light the MoonBoard; regression-check the same stall flow on Kilter/Tension still works.

https://claude.ai/code/session_01TD72vps2umF37J1yNpLTs3

---
_Generated by [Claude Code](https://claude.ai/code/session_01TD72vps2umF37J1yNpLTs3)_

### DIFF
--- a/packages/mobile/ios-tests/LiveActivityWidgetTests.swift
+++ b/packages/mobile/ios-tests/LiveActivityWidgetTests.swift
@@ -132,6 +132,43 @@ final class LiveActivityWidgetTests: XCTestCase {
         XCTAssertNil(SharedQueueState.currentItem(from: defaults))
     }
 
+    func testMoonboardPacketEncodesRolesAndSerialPositions() {
+        // Parity with packages/shared/ble-protocol/src/__tests__/moonboard.test.ts:
+        // p1r42 -> S at serial position 0, p2r43 -> P at serial position 35.
+        let result = BoardBleEncoding.makeMoonboardPacket(frames: "p1r42p2r43")
+
+        XCTAssertEqual(String(decoding: result.packet, as: UTF8.self), "l#S0,P35#")
+        XCTAssertEqual(result.skippedRoleCount, 0)
+        XCTAssertEqual(result.skippedPositionCount, 0)
+        XCTAssertEqual(result.totalPlacements, 2)
+    }
+
+    func testMoonboardPacketSkipsUnknownRolesAndOutOfRangeHolds() {
+        // Role 45 (foot) is not a MoonBoard LED role; hold id 199 is past the
+        // 198-hold grid. Both are dropped, the valid start hold still lights.
+        let result = BoardBleEncoding.makeMoonboardPacket(frames: "p1r42p5r45p199r43")
+
+        XCTAssertEqual(String(decoding: result.packet, as: UTF8.self), "l#S0#")
+        XCTAssertEqual(result.skippedRoleCount, 1)
+        XCTAssertEqual(result.skippedPositionCount, 1)
+        XCTAssertEqual(result.totalPlacements, 3)
+    }
+
+    func testMoonboardPacketIsEmptyWhenNothingEncodes() {
+        // No clear-all: an empty or all-skipped frame must produce no packet so
+        // the BLE manager refuses to write and leaves the wall lit as-is.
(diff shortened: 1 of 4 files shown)


marcodejongh:
Thanks for the review. All three are intentional and I'm leaving them as-is:

1. **`totalPlacements` divergence from JS** — agreed it differs, but those result fields (`totalPlacements` / `skipped*`) are diagnostic only; the native caller (`displayItemOnBleQueue`) just checks `packet.isEmpty`. Real climb frames are always well-formed `p{int}r{int}`, so the count only diverges on malformed input that never occurs in practice and is never read here. Not worth complicating the shared `parseFrames`.

2. **`moonboardSerialPosition` internal-not-private** — this matches the file's existing convention: every encoding helper (`checksum`, `wrapBytes`, `encodePositionV3/V2`, `encodeColorV3`, `scaledColorV2`) is an internal `static func`. The Swift test target compiles `BoardBleEncoding.swift` directly as a source (see `prepare-rn-ios-tests.mjs`), not via `@testable import` of a built module, so making it `private` would break the test rather than tidy it.

3. **No unit test for the BLE-manager branch** — `BoardBleManager` is a CoreBluetooth-driven singleton with no existing unit-test harness; the Aurora dispatch branch isn't unit-tested either. The encoder (the part with real logic) is covered; the dispatch branch is exercised by the on-device test plan in the PR description.

---

PR #3222, opened 2026-06-26
feat(mobile): now-playing accessory bar — status, not a directive

marcodejongh:
## Summary

> **Rollout:** the entire redesign (eyebrow, peer read-only tick, social-tab hide) is gated by one flag — `accessory-now-playing`. Off (default) = byte-identical to today's bar; on = the full behaviour. Flip it to A/B.

Reworks the persistent "active climb" accessory bar into a **now-playing** model so it stops reading as a directive while you browse social features.

The bar collapsed three different meanings into one identical surface on every tab:
- your **queue head**,
- a climb lit on a board **you** drive,
- a **peer's** lit climb in a party session.

So on the feed / profile / discover tabs a bare climb name looked like "climb this now" instead of "this is what's lit". PostHog (90d) backs the problem: social/browsing tabs out-navigate the board tab ~2:1, yet only ~17% of accessory-bar opens happen there — it's noise while browsing, but heavily used on the board tabs (5,680 opens on `/climbs`). So the fix is disambiguation + context-awareness, **not** removal.

### What changed
- **Eyebrow caption** (ships on): a status line above the climb name — `On the wall · live` (you're driving), `{name} on the wall` (a crew-mate is), or `Up next` (just your queue). Turns a directive into status.
- **A peer's climb is read-only**: the tick hides when a teammate is driving the wall — you can't log someone else's send.
- **Context-aware presence** (behind the `accessory-now-playing` flag): a queue-only bar is hidden on the social tabs (home / profile / discover) but **persists as status whenever a board is live**; the board-control tabs (climbs / record) are unchanged.

Disambiguation logic is a renderer-agnostic `deriveAccessoryContext` in `@boardsesh/play-view` (with unit tests) — web can adopt the eyebrow later. The iOS 26 native platter is never mount-gated per tab (avoids the UIKit doubled-text snapshot) — only its content changes. No new animations, so Android touch is untouched.

> Direction set by a multi-agent design team (7 approaches → adversarial critique → 3-judge panel). The now-playing model was the unanimous winner across user-clarity, engineering-feasibility, and design-craft.

## Release Notes

_The whole now-playing redesign ships behind the `accessory-now-playing` flag, off by default — no user-facing change in this build. The release note lands when the flag is enabled in a later build._

## Test plan

- `vp run typecheck:mobile` + `vp run typecheck:shared` — pass
- `vp run test:mobile` — 2686 pass; new `deriveAccessoryContext` unit tests + social-gate / tick coverage added to `persistent-queue-bar.test`
- i18n catalog parity (`catalog-completeness`) — new `queueBar.nowPlaying.*` keys in en-US / es / fr
- `vp run check:mobile-bundle` — Android bundle OK (12.3 MB)
- `vp check` — 0 errors
- **Manual QA (to do, both variants × iOS sim + Android emulator):** drive no-board / your-lit / wall-not-yours / peer / mid-session on a board tab and a social tab. Confirm: flag-on hides the queue-only social bar; eyebrow text matches state; no doubled-text on iOS 26 player open/close; tick hidden for a peer's climb.

https://claude.ai/code/session_01YRxL6eZCa236V51hvaa2GS

### DIFF
--- a/docs/websocket-implementation.md
+++ b/docs/websocket-implementation.md
@@ -313,6 +313,8 @@ This ensures `BoardSessionBridge` only calls `activateSession()` when the actual
 
 Board presence powers the mobile "now on the wall" feed, board sheet history, and board sheet stats. It is independent of party-session join: the mobile app resolves a board id for the wall feed before subscribing to `boardNowPlaying` or fetching board-presence history/stats.
 
+It also drives the accessory bar's "now playing" eyebrow (`deriveAccessoryContext`): when a session peer holds the wall (`boardConnection === 'heldByPeer'`), the bar reads `{name} on the wall` and hides the tick — a peer's lit climb is read-only to you (you can't log someone else's send).
+
 Mobile resolves the feed board id in this order:
 
 1. `resolveBoardForUuid(boardUuid)` for the selected named board. This is the default board-sheet path and binds to the actual `user_boards` row, so stats/history are available before Bluetooth connects and stay aligned with board-scoped ticks.
--- a/packages/mobile/src/components/queue-control/ClimbCapsule.tsx
+++ b/packages/mobile/src/components/queue-control/ClimbCapsule.tsx
@@ -187,7 +187,9 @@ export function ClimbCapsule({
         <View
           style={[styles.tapArea, { height, borderRadius: capsuleRadius }]}
           accessibilityRole="button"
-          accessibilityLabel={currentClimb.name}
+          // Announce the status caption (live / on the wall / up next) too, so a
+          // screen reader gets the same context the eyebrow gives sighted users.
+          accessibilityLabel={eyebrowText ? `${eyebrowText}, ${currentClimb.name}` : currentClimb.name}
         >
           <View style={[styles.labelSlot, { left: labelLeft, right: labelRight }]}>
             <ClimbLabel
--- a/packages/mobile/src/components/queue-control/ClimbCapsule.tsx
+++ b/packages/mobile/src/components/queue-control/ClimbCapsule.tsx
@@ -20,6 +20,7 @@ import { AccessoryBarSurface, type AccessoryBarSurfaceTreatment } from './Access
 import { AccessoryClimbThumbnail } from './AccessoryClimbThumbnail';
 import { useAccessoryClimbTap } from './use-accessory-climb-tap';
 import { useWallOrQueueCurrentClimb } from './use-wall-or-queue-climb';
+import { useAccessoryEyebrow } from './use-accessory-presentation';
 import { useBoardConnectionState } from '../ble/use-board-connection-state';
 import { BoardControlIndicator } from './BoardControlIndicator';
 
@@ -30,32 +31,59 @@ type ClimbLabelProps = {
   gradeColor: string;
   showThumbnail: boolean;
   boardConfig: BoardConfig | null;
+  // The status caption above the name ("On the wall · live" / "Tara on the wall"
+  // / "Up next"). This is the whole disambiguator: it turns a bare climb name
+  // from a directive into a labelled status.
+  eyebrow: string;
+  eyebrowColor: ColorValue;
 };
 
-function ClimbLabel({ climb, labelColor, formattedGrade, gradeColor, showThumbnail, boardConfig }: ClimbLabelProps) {
+function ClimbLabel({
+  climb,
+  labelColor,
+  formattedGrade,
+  gradeColor,
+  showThumbnail,
+  boardConfig,
+  eyebrow,
+  eyebrowColor,
+}: ClimbLabelProps) {
   return (
     <View style={styles.labelInner}>
       {showThumbnail ? <AccessoryClimbThumbnail climb={climb} boardConfig={boardConfig} /> : null}
-      <Text
-        variant="subheadline"
-        color={labelColor}
-        numberOfLines={1}
-        ellipsizeMode="tail"
-        maxFontSizeMultiplier={CHROME_LABEL_MAX_FONT_SCALE}
-        style={styles.name}
-      >
-        {climb.name}
-      </Text>
-      {formattedGrade ? (
+      <View style={styles.labelTextColumn}>
         <Text
-          variant="headline"
+          variant="caption1"
+          color={eyebrowColor}
           numberOfLines={1}
           maxFontSizeMultiplier={CHROME_LABEL_MAX_FONT_SCALE}
-          style={[styles.gradeText, { color: gradeColor }]}
+          style={styles.eyebrow}
         >
-          {formattedGrade}
+          {eyebrow}
         </Text>
-      ) : null}
+        <View style={styles.nameRow}>
+          <Text
+            variant="subheadline"
+            color={labelColor}
+            numberOfLines={1}
+            ellipsizeMode="tail"
+            maxFontSizeMultiplier={CHROME_LABEL_MAX_FONT_SCALE}
+            style={styles.name}
+          >
+            {climb.name}
+          </Text>
+          {formattedGrade ? (
+            <Text
+              variant="headline"
+              numberOfLines={1}
+              maxFontSizeMultiplier={CHROME_LABEL_MAX_FONT_SCALE}
+              style={[styles.gradeText, { color: gradeColor }]}
+            >
+              {formattedGrade}
+            </Text>
+          ) : null}
+        </View>
+      </View>
     </View>
   );
 }
@@ -85,6 +113,8 @@ export function ClimbCapsule({
   // Connection state drives the leading control + the "you have control" glow.
   // Read from the single source so the bar can't disagree with the drawer bulb.
   const { boardConnection, bluetooth } = useBoardConnectionState();
+  // Status caption that labels what the bar is showing (live / peer / up next).
+  const { text: eyebrowText, tone: eyebrowTone } = useAccessoryEyebrow();
 
   // Source-of-truth flip: show the wall's lit climb when a board feed is live
   // (flag-gated), else the local queue head.
@@ -124,6 +154,9 @@ export function ClimbCapsule({
   // stripe recolors to the brand violet so it reads as a control-on edge marker.
   const showGradeAccent = surfaceTreatment === 'docked';
   const gradeAccentColor = connected ? brandColors.primary : grades.currentColor;
+  // "live" runs hot in the brand accent; peer / up-next stay quiet so they read
+  // as ambient status rather than a call to action.
+  const eyebrowColor = eyebrowTone === 'live' ? brandColors.primary : systemColors.secondaryLabel;
 
   return (
     <AccessoryBarSurface
@@ -154,6 +187,8 @@ export function ClimbCapsule({
               gradeColor={grades.currentColor}
               showThumbnail={showThumbnail}
               boardConfig={boardConfig}
+              eyebrow={eyebrowText}
+              eyebrowColor={eyebrowColor}
             />
           </View>
         </View>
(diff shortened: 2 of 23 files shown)


marcodejongh:
Thanks — addressed in 7a9e675:

- **Dead code `isBoardControlSurface`** — removed (plus its `BOARD_CONTROL_TABS` set). The social gate only needs `isSocialSurface`.
- **Missing route-helper tests** — added `isSocialSurface` coverage in `route-segments.test.ts` (home/profile/discover + a sub-route true; climbs/record/non-tabs false).
- **Spanish "Up next"** — changed `"Siguiente"` → `"A continuación"`, the idiomatic media "up next" phrasing.
- **`ClimbCapsule` eyebrow** — now guarded with `{eyebrow ? … : null}`, matching `NativeAccessoryClimbRow`.

One I left as-is: the **multi-line comment** flag. That rule isn't in this repo's `CLAUDE.md`, and the immediate neighbours (`use-board-connection-state.ts`, `use-wall-or-queue-climb.ts`) use the same multi-line block style — the file-level blocks here carry the iOS-26 native-platter / Android rationale, so I kept them consistent with the surrounding code. Happy to trim if you'd prefer.

marcodejongh:
Second pass addressed in 02f58de:

1. **Double `useBoardConnectionState` subscription** — `ClimbCapsule` now reads the connection state once and derives the eyebrow from it via a new `useAccessoryEyebrowText(context)`; `useAccessoryEyebrow()` stays as the convenience wrapper for the native row / gate.
2. **`eyebrow` prop type** — unified to `string | null` in both `ClimbCapsule` and `NativeAccessoryClimbRow`.
3. **Test coverage** — added native-row tests for the peer case (`showTick: false` → no tick) and `inline` placement (eyebrow dropped, name still shown).
4. **Duplicated eyebrow style** — extracted `ACCESSORY_EYEBROW_TEXT_STYLE` to `theme/typography.ts`; both rows import it.
5. **Docs** — added a one-liner to `docs/websocket-implementation.md` (Board Presence Wall Feed) covering the peer read-only accessory behaviour.

Local: `vp run typecheck:mobile` clean, 54 tests pass across the touched suites.

marcodejongh:
Third pass in 007edae:

- **Duplicated eyebrow colour logic** — extracted `accessoryEyebrowColor(tone, live, idle)`; both the capsule and the native row call it.
- **`useAccessoryEyebrowText` naming/return** — renamed to `useAccessoryEyebrowFromContext` and dropped the unused `tier` from `AccessoryEyebrow` (tier is read from `useAccessoryPresentation`, where it belongs).
- **No `ClimbCapsule` eyebrow tests** — added three: up-next (quiet secondary colour), live (brand accent), and a named peer. These exercise the real `useBoardConnectionState → deriveAccessoryContext → useAccessoryEyebrowFromContext` path (only connection-state + i18n are mocked).

On **manual QA**: the iOS-26 doubled-text / sim checks can't run here (this is a Linux box; the iOS simulator is macOS-only), so those stay in the maintainer's device pass and are documented in the test plan. The flag-gated social-hiding is off by default, so default behaviour is unchanged. I'm grabbing an Android emulator capture now to confirm the two-line eyebrow layout doesn't clip on Material — will attach.

marcodejongh:
Fourth pass in a01926:

**Fixed**
- **#3 Exhaustiveness** — `deriveAccessoryContext` now ends with `boardConnection satisfies 'disconnected'`, so adding a new `BoardConnection` variant without a case is a compile error (keeps the safe disconnected runtime default).
- **#4 `peerAnon` localization** — added a `ClimbCapsule` test for `heldByPeer` + no holder name → the `peerAnon` key.
- **#5 `upNext` in the native row** — added a `regular`-placement test asserting the up-next eyebrow renders (tick still shown for your own queue head).

**Pushing back (judgment calls for the human reviewer)**
- **#1 Layout shift when the tick disappears (peer takeover)** — reserving a permanent zero-opacity tick slot trades a *rare, one-time* reflow (only when a teammate takes the wall mid-session) for a *permanent* empty 44pt gap on every peer/read-only row. I think the empty gap is the worse look, so I'd rather let the name reclaim the space. Happy to reserve it if you disagree.
- **#2 Cross-component `useBoardConnectionState`** — `ClimbCapsule` needs the connection state for its own glow + leading control (`bluetooth`), independent of the eyebrow, so a context prop wouldn't remove its subscription. `useBoardConnectionState` is memoized, so a parent + child reading the same context is the idiomatic cost here, not a redundant fetch. (The *within-component* double-read from the earlier pass is already gone.)

**Manual QA note:** I tried an Android-emulator capture to eyeball the two-line eyebrow, but the emulator can't get past the login screen (it hits the known Android sign-in freeze, and the bar also needs a queued climb + a connected board to show the live/peer states — neither is reachable on an emulator). So the visual layout pass stays on a real device with the maintainer, as the test plan notes. The layout is covered structurally by typecheck + the render tests, and the JS bundle builds clean.

marcodejongh:
Latest review came back **✅ Ready to merge**, so I've taken this out of draft.

Two trivial non-blocking leftovers I've intentionally left rather than spin another CI round:
- a stray `tier` field in the native-row test's hoisted eyebrow mock (harmless), and
- an optional `eyebrow.kind satisfies never` default in the eyebrow `useMemo` switch.

The cross-component `useBoardConnectionState` read is left as-is by design (the capsule needs `bluetooth` independent of the eyebrow). Happy to fold any of these in if you'd like before merge. Over to you for human review + the on-device QA pass (iOS 26 platter + eyebrow layout) noted in the test plan.

marcodejongh:
Thanks @codex — both caught real things.

**P2 #1 (chrome metrics out of sync) — fixed in 40fd8df.** You're right: the bar returned `null` on social tabs but `useBottomChromeMetrics` still reserved `TOOLBAR_RESERVE` / `MATERIAL_ACTIVE_CONTEXT_BAR_HEIGHT`, leaving a toolbar-sized gap and over-lifting the FABs. I extracted the hide predicate into `useQueueBarHiddenOnSocial` (its own module so the eyebrow hooks stay free of expo-router/flag deps) and threaded it through `computeBottomChromeMetrics`, so a hidden bar now drops its reserve. JS-path only, with tests covering both the dropped reserve and that the native accessory's reserve is untouched.

**P2 #2 (native platter not gated on social) — deliberate, documented tradeoff.** The iOS 26 `NativeTabs.BottomAccessory` host is wired specifically to *never* unmount/remount or blank its content mid-flight — UIKit snapshots an empty→content transition as doubled text (see the comments in `QueueBottomAccessory` + `route-segments.isTabsChromeRoute`, and the `useStickyAccessoryPresence`/`useRetainedAccessoryClimb` machinery). Tab-gating that mount (or null-ing the content on social tabs) reintroduces exactly that bug. So on the native path the bar intentionally *persists with the `Up next` eyebrow* rather than hiding — it still reads as status, just not absent. The full "absent on social" applies to the JS bar (Material / iOS < 26 / Android), where hiding is snapshot-safe. This is called out in the PR description and the plan.

The A/B still holds: the native cohort is a no-op under the flag (no regression, just less signal), and the eyebrow disambiguation — the core of the change — ships on every path. If we want the native path suppressed too, that's a separate spike to confirm whether unmounting the host is safe on a real iOS 26 device, which I can't validate from Linux. Happy to file a follow-up.

marcodejongh:
Addressed in 68f1ba7:

- **#1 (extra `useBoardConnectionState` subscription)** — my previous metrics fix had added a second `useAccessoryPresentation()` read inside `PersistentQueueBar`. Extracted a pure `shouldHideQueueBarOnSocial(flag, tier, segments)` so the bar now derives both `showTick` and the hide from **one** `useAccessoryPresentation` read. `ClimbCapsule`'s read stays — it needs `bluetooth`/connection state for its glow + leading lightbulb independent of the eyebrow, so a context prop wouldn't remove it; the remaining bar↔capsule pair is the irreducible minimum on a memoized context.
- **#3 (test re-implements the predicate)** — `persistent-queue-bar.test` no longer hand-codes the condition; the real pure `shouldHideQueueBarOnSocial` runs, driven by the flag / `isSocialSurface` / presentation mocks, so a change to the predicate is now actually caught.

On the two recurring notes: **manual QA** is the documented on-device pass (can't run iOS sim from Linux; the emulator can't reach the bar past the login freeze) — left for the maintainer as the test plan states. **Multi-line file comments** — I checked again: that rule isn't in this repo's `CLAUDE.md`, and the sibling hooks (`use-board-connection-state.ts`, `use-wall-or-queue-climb.ts`) use the same block style, so these are consistent with the codebase.

Full mobile suite green (2698 tests).

marcodejongh:
**Behaviour change in 2c84330:** the *entire* redesign is now gated by `accessory-now-playing`, not just the social-tab hide. Previously the eyebrow caption + peer read-only tick shipped unflagged, so the change was visible before the flag was turned on.

- `deriveAccessoryContext` takes a pure `enabled` (defaults true): `false` → `null` eyebrow + tick always shown.
- The flag is read once via a new `useAccessoryNowPlayingEnabled` leaf hook and folded into `useAccessoryPresentation`; `ClimbCapsule`, the social-hide, and the bar all read it through that one hook.
- **Flag OFF (default) = byte-identical to today's bar** (no eyebrow, tick always shown, never hidden on social). Flag ON = the full now-playing behaviour. Clean on/off for A/B.

Since it's now fully behind a default-off flag, this build ships no user-facing change — I updated **Release Notes → none** accordingly (the note lands when the flag is enabled in a later build). Full mobile suite green (2699).

marcodejongh:
All three addressed in 9c3a03d (good catches — these surfaced once the flag gated the whole feature):

1. **Flag description** — now lists all three gated behaviours (status eyebrow, peer read-only tick, social-tab hide), so testers in Dev settings see the full scope rather than just the hide.
2. **Accessibility** — `accessibilityLabel` now announces the status caption (live / on the wall / up next) alongside the climb name, in both the capsule and the native platter row (and in the `inline` placement where the eyebrow isn't drawn). So screen-reader users get the same context the eyebrow gives sighted users before the flag is enabled.
3. **Misleading test name** — renamed; the `enabled` default of `true` is a pure-function convenience, and the comment now makes clear the production default is off (the flag).

marcodejongh:
Addressed in cfa1a194d:

- **`enabled` footgun** — made it a **required** param of `deriveAccessoryContext` (was defaulting to `true`). You're right that a default-true silently hands a forgetful caller the redesign even though the production flag defaults off. Now every caller must pass it (mobile already does); the unit tests pass it explicitly.
- **Tick-slot reflow on the iOS 26 platter** — reversing my earlier pushback here, since a second reviewer flagged it and it's specifically the *live UIKit platter* during party control-passing (which recurs, not a rare one-off). The tick slot now **always reserves its width**; only the button inside is conditional, so a peer taking the wall no longer relayouts the row.

On **manual QA**: it stays the maintainer's on-device pass (documented in the test plan) — the iOS-26 doubled-text / sim checks can't run from this Linux box, and the emulator can't reach the bar past the login freeze. Everything else is covered by the 2709-test suite.

Full mobile suite green (2709); typechecks clean.

marcodejongh:
Both in e1b3e72:
- Added `default: eyebrow.kind satisfies never` to `useAccessoryEyebrowFromContext`'s switch (matches the `deriveAccessoryContext` pattern) — a new eyebrow kind is now a compile error, not a silent `undefined`.
- Memoized the hook's return object so a future `React.memo`/context consumer can't churn on a fresh identity.

The only remaining open note is **manual QA**, which stays the maintainer's on-device pass (iOS-26 doubled-text + visual differentiation can't run from this Linux box). Everything else is green: full mobile suite (2709), typechecks, lint.

marcodejongh:
Holding on this round — nothing here is a new defect:

- **Manual QA** — the maintainer's on-device pass (iOS-26 doubled-text can't run from Linux). Noted in the test plan.
- **Double subscription via `useBottomChromeMetrics`** — `useBottomChromeMetrics` is a *shared* layout hook used by many screens (home / discover / record / …), each of which must independently know whether the bar is suppressed. So it has to derive `useQueueBarHiddenOnSocial()` itself; passing it in only from `PersistentQueueBar` would break every other consumer. The extra read is a memoized context lookup — low-severity, and the alternative is worse coupling.
- **Curly apostrophe** in the flag description — valid (it's tester-facing copy), just a Unicode `'`. Happy to swap it for a straight one if you'd like, but I'm not spinning CI for a cosmetic char.

The PR is green and every substantive item across the review rounds is addressed. Over to you for the on-device QA + merge.

PR #3225, opened 2026-06-26
fix(mobile): pick BLE write type from characteristic on iOS (MoonBoard LEDs dark)

marcodejongh:
## Summary

On iOS, connecting to a MoonBoard succeeds but the LEDs never light. The same hardware works on Android, running the **same** shared JS encoder — so it isn't the protocol. I confirmed the encoding is correct by cross-checking `getMoonboardBluetoothPacket` against two reverse-engineering references ([e-sr/moonboard](https://github.com/e-sr/moonboard), [FabianRig/ArduinoMoonBoardLED](https://github.com/FabianRig/ArduinoMoonBoardLED)): the `l#…#` frame format, the `S`/`P`/`E` hold types, the zigzag strip serial mapping, and our row-major holdId numbering all match.

The bug is iOS-native only. `BoardBleManager` always wrote with CoreBluetooth `.withoutResponse`, which **CoreBluetooth silently drops** when the target characteristic doesn't advertise the `writeWithoutResponse` property. The original MoonBoard LED box's Nordic UART RX characteristic (`6e400002`) supports only `.write` (with response) — e-sr declares it `['write']` — whereas Aurora (Kilter/Tension) boards support without-response, which is why Aurora works on iOS. Android's GATT stack delivers no-response writes regardless of declared properties, hiding the bug there.

**Fix:** pick the write type from the discovered characteristic's actual `properties` (`BoardBleEncoding.preferredWriteType`). Keep the unacknowledged fast path for Aurora (no behaviour change), fall back to write-with-response for the MoonBoard box. The with-response path is paced by a new `didWriteValueFor` ack (the natural backpressure signal) instead of `canSendWriteWithoutResponse`/`peripheralIsReady`, with a stall watchdog that reuses the existing #3181 link-cycle recovery.

Changes (all iOS-native): `BoardBleEncoding.swift` (testable write-type helper), `BoardBleManager.swift` (branch the write pump + new `didWriteValueFor` delegate + ack cleanup in `failQueuedWrites`), and a unit test.

## Release Notes

- MoonBoard holds now light up on iPhone — connect and your problem shows on the wall.

## Test plan

- [x] Encoder verified against the e-sr and FabianRig references (frame format, hold types, serial-position zigzag, hold numbering).
- [x] `vp run typecheck:mobile` green.
- [x] Added an XCTest for `preferredWriteType` (Aurora → without-response; MoonBoard-only-`.write` → with-response; both → without-response; neither → with-response). Runs in the iOS-sim `BoardseshTests` job.
- [x] Aurora write path left byte-for-byte unchanged (no regression).
- [ ] **Native build required — cannot ship via OTA** (CoreBluetooth change → new fingerprint). Pending: a tester with a real MoonBoard LED box on `vp run mobile:preview-build` confirms (1) a multi-chunk climb lights fully, (2) Aurora still lights with no added latency, (3) the Live Activity re-light path works on MoonBoard.

### DIFF
--- a/packages/mobile/ios-tests/LiveActivityWidgetTests.swift
+++ b/packages/mobile/ios-tests/LiveActivityWidgetTests.swift
@@ -1,3 +1,4 @@
+import CoreBluetooth
 import XCTest
 
 @available(iOS 17.0, *)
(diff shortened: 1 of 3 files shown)


marcodejongh:
Thanks for the review. Both notes are non-blocking; here's my call on each:

**Test placement** — keeping `testPreferredWriteTypeSelectsByCharacteristicProperties` in `LiveActivityWidgetTests.swift`. The iOS test target's source list is hand-curated in `scripts/prepare-rn-ios-tests.mjs`, and that file is already registered there; a new `BoardBleEncodingTests.swift` would need a matching entry plus a fresh native-build CI cycle to confirm it's actually compiled (a silently-unregistered test file is a false green). `BoardBleEncoding` lives in `modules/live-activity/ios/`, so this file is a reasonable home for the module's encoding tests.

**"Dead" guard in `didWriteValueFor`** — keeping the `guard !writeQueue.isEmpty` branch; it's load-bearing, not dead. A `didWriteValueFor` can land *after* `failQueuedWrites` empties the queue (a disconnect or `cancelWrites` racing the in-flight write). In that window an error callback would hit `writeQueue.removeFirst()` on an empty array, which traps. The guard turns that into a clean no-op. Same defensive shape as the existing `notConnected` path in `writeChunk`.

PR #3227, opened 2026-06-27
test(mobile): cover queue-grow path in playlist activation; drop dead code

marcodejongh:
## Summary

Addresses three review findings on the mobile playlist-activation hook (`packages/mobile/src/lib/playlists/use-playlist-activation.ts`):

1. **Missing test for the count-bump path in `confirmQueueReplacement`.** When the queue grows between opening the confirmation sheet and pressing Confirm, the hook bumps the displayed warning count and short-circuits, forcing a second confirm so the user never clears items they haven't seen counted. This path had no coverage. Added a test that grows the queue while the sheet is open, asserts the count bumps (1 → 2) with no `setQueue`/fetch, then asserts the second confirm goes through and replaces the queue.

2. **Dead code in `buildPlaylistQueue`.** The trailing `if (!currentItem)` block was unreachable — the activated climb is always assigned to `currentItem` (either in the loop or the preceding append). Collapsed the redundant `includesActivatedClimb` flag into the `currentItem` null-check and removed the dead block, keeping the function type-safe (the remaining guard still narrows `currentItem` to non-null).

3. **Spanish glossary.** `docs/i18n-spanish-glossary.md` didn't list `playlist`, so the existing `es/playlists.json` strings (including the new `queueReplace` keys) couldn't be verified against it. The borrowed-English `playlist` is the established convention throughout that catalog; documented it in the glossary (feminine: _la playlist_; never _lista de reproducción_) so it's no longer flagged.

## Release Notes

## Test plan

- `vp test run --project mobile use-playlist-activation` — 22 passed (was 21).
- `vp lint` + `vp fmt --check` on the three changed files — 0 warnings/errors, correctly formatted.

https://claude.ai/code/session_01XT6hyppvBAAerw8NM6aMGi

---
_Generated by [Claude Code](https://claude.ai/code/session_01XT6hyppvBAAerw8NM6aMGi)_

### DIFF
--- a/docs/i18n-spanish-glossary.md
+++ b/docs/i18n-spanish-glossary.md
@@ -54,6 +54,8 @@ These are already used consistently in the catalogs — keep using them so we do
 
 Kept in English by convention (technical board-config terms): **layout**, **set / sets**, **logbook**, **setter**, **beta**.
 
+**`playlist` / `playlists`** also stay in English — used throughout `es/playlists.json` (_Compartir playlist_, _¿Iniciar playlist?_, _Playlist eliminada_). Feminine: _la playlist_, _esta playlist_, _las playlists_. Never _lista de reproducción_.
+
 ## Process
 
 - Add every new key to **all** locales (`en-US`, `es`, `fr`) — `catalog-completeness.test.ts` fails on missing keys.
--- a/packages/mobile/src/lib/playlists/__tests__/use-playlist-activation.test.tsx
+++ b/packages/mobile/src/lib/playlists/__tests__/use-playlist-activation.test.tsx
@@ -475,6 +475,8 @@ describe('usePlaylistActivation (mobile wrapper)', () => {
       });
       const lastSetQueue = mocks.setQueue.mock.calls.at(-1);
       expect(lastSetQueue?.[0].map((item: ClimbQueueItem) => item.climb.uuid)).toEqual(['b', 'c']);
+      // The second confirm runs the full replace, so the ordered playlist is fetched.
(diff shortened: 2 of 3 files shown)


marcodejongh:
Closing this — both code changes are already on `main`, arrived by another route.

**The dead `!currentItem` block is gone.** `buildPlaylistQueue` in `packages/mobile/src/lib/playlists/use-playlist-activation.ts` now reads:

```ts
if (currentItem) return { queue, currentItem };
// The activated climb wasn't in the loaded list (or the list was empty) —
// append it so the queue always contains and centres on the tapped climb.
const appended = reusableCurrentItem ?? climbToQueueItem(toSchemaClimb(activatedClimb));
queue.push(appended);
return { queue, currentItem: appended };
```

Same restructuring this PR proposed: the `includesActivatedClimb` flag collapsed into the `currentItem` null-check, the unreachable trailing block deleted.

**The queue-grow test exists.** `use-playlist-activation.test.tsx:437` — *"bumps the count and requires a second confirm when the queue grows while the sheet is open"* — covers the same path: grow the queue while the sheet is open, assert the count bumps with no `setQueue`/fetch, then assert the second confirm replaces the queue.

**One line hasn't landed:** the `playlist` entry in `docs/i18n-spanish-glossary.md`. That doc still has no `playlist` mention, while `es/playlists.json` uses the borrowed English word 35 times (_Descubre playlists de escalada_, _Playlist de escalada en Boardsesh_) and never _lista de reproducción_. Documenting it — feminine, _la playlist_ / _las playlists_ — is a genuine improvement, so I've filed it as #3992 rather than keep a 1727-commit-behind branch open for one doc line.

PR #3228, opened 2026-06-27
fix(mobile): keep Aurora BLE writes on write-without-response (iOS boards stalled)

marcodejongh:
## Summary

A tester on a **Kilter/Tension (Aurora)** board, iPhone 15 Pro, **iOS 26.5**, hit a fresh cluster of BLE errors (Sentry: `BOARDSESH-7Y` write_timeout ×8, `BOARDSESH-81` write_recovery_failed, `BOARDSESH-7Z` disconnected, `BOARDSESH-80` connect_failed) — the board connects and its status LED lights, but climbs never reach the wall. `write_timeout` has **zero history in the prior 90 days**; it started ~2.5h after commit `00bda53a2` ("pick BLE write type from characteristic") merged.

`00bda53a2` changed the iOS native writer from *always* write-without-response to choosing the CoreBluetooth write type from the characteristic's advertised properties. On iOS, Aurora UART RX characteristics that don't advertise the `.writeWithoutResponse` bit got routed to the new write-**with**-response path, which only advances on a `didWriteValueFor` ATT ack. The board never sends that ack → every write stalls at the 5s watchdog → the #3181 recovery cycles the link until it gives up.

The old Capacitor app (same Swift Live Activity + native BLE, which we ported from) and the first RN port both wrote `.withoutResponse` **unconditionally** for every board, and that worked on this hardware for months. This change:

- **Gates write-with-response to MoonBoard only.** `BoardBleEncoding.preferredWriteType` now takes `boardName`; Aurora (and an unknown / `nil` board) always use write-without-response — byte-for-byte the pre-`00bda53a2` path. The MoonBoard LED-box fix from `00bda53a2` is preserved.
- **Adds connect-time write diagnostics** (advertised characteristic properties, `supportsWriteWithoutResponse`, chosen write type, negotiated write lengths) surfaced to Sentry tags + the `Bluetooth Connection Success` event, so we can tell whether a board genuinely lacks the property bit (firmware) or iOS 26.5 reports it differently — the latter would be a broader latent regression worth a follow-up.

The MoonBoard auto-fallback-to-without-response idea was **deliberately deferred**: for a genuine `.write`-only box, CoreBluetooth silently drops the without-response retry and the write would report success against a dark wall — re-opening the exact bug `00bda53a2` fixed.

Closes BOARDSESH-7Y, BOARDSESH-81, BOARDSESH-7Z, BOARDSESH-80.

## Release Notes

- Kilter and Tension boards light up reliably again on iPhone — climbs send to the wall instead of the board connecting but staying dark.

## Test plan

- `vp run test:mobile` — 2713 pass
- `vp check` (scoped to changed TS) — pass
- iOS Swift unit test `testPreferredWriteTypeIsMoonboardGated` updated for the new signature; runs in `ios-rn-ci.yml` (BoardseshTests).
- **Manual (pending):** install on a physical iPhone (iOS 26.x) with a Kilter/Tension board, send a climb, confirm the wall lights and no `write_timeout`/`write_recovery_failed`; confirm the new Sentry tags (`ble_chosen_write_type`, `ble_supports_without_response`, …) appear on connect. If a MoonBoard LED box is available, confirm it still lights.
- This is a **native** change — it reaches testers only via a new native build (TestFlight / `vp run mobile:preview-build`), not OTA.

https://claude.ai/code/session_01UyZrUjhTHLHCpDTskUrCFK

### DIFF
--- a/packages/mobile/ios-tests/LiveActivityWidgetTests.swift
+++ b/packages/mobile/ios-tests/LiveActivityWidgetTests.swift
@@ -212,6 +212,9 @@ final class LiveActivityWidgetTests: XCTestCase {
         XCTAssertEqual(BoardBleEncoding.preferredWriteType(for: .write, boardName: "moonboard"), .withResponse)
         XCTAssertEqual(BoardBleEncoding.preferredWriteType(for: .writeWithoutResponse, boardName: "moonboard"), .withoutResponse)
         XCTAssertEqual(BoardBleEncoding.preferredWriteType(for: [.write, .writeWithoutResponse], boardName: "moonboard"), .withoutResponse)
+        // Defensive default: a characteristic advertising neither write property
+        // can't be written either way (CoreBluetooth drops/ rejects the write and
+        // the wall stays dark) — we just don't pick the unacknowledged path for it.
         XCTAssertEqual(BoardBleEncoding.preferredWriteType(for: [.read], boardName: "moonboard"), .withResponse)
     }
 }
--- a/packages/mobile/ios-tests/LiveActivityWidgetTests.swift
+++ b/packages/mobile/ios-tests/LiveActivityWidgetTests.swift
@@ -190,19 +190,28 @@ final class LiveActivityWidgetTests: XCTestCase {
         XCTAssertTrue(url?.absoluteString.contains("include_background=1") ?? false)
     }
 
-    // Regression for the iOS-only "MoonBoard connects but LEDs stay dark" bug:
-    // CoreBluetooth silently drops a `.withoutResponse` write to a characteristic
-    // that lacks the property. Aurora UART advertises `.writeWithoutResponse`
-    // (fast path); the original MoonBoard LED box advertises only `.write`, so it
-    // must use `.withResponse`.
-    func testPreferredWriteTypeSelectsByCharacteristicProperties() {
-        // Aurora (Kilter/Tension): supports write-without-response → fast path.
-        XCTAssertEqual(BoardBleEncoding.preferredWriteType(for: .writeWithoutResponse), .withoutResponse)
-        // Original MoonBoard LED box: write-with-response only.
-        XCTAssertEqual(BoardBleEncoding.preferredWriteType(for: .write), .withResponse)
-        // Supports both → prefer the unacknowledged fast path.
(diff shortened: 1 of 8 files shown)


marcodejongh:
Addressed the automated-review feedback across the last few commits. Two recurring items I deliberately did **not** change, with rationale:

1. **`clearBleDiagnosticsTags` using `setTag(key, undefined)`** — verified against the installed SDK: `@sentry/core`'s `Scope` exposes **no `removeTag`** (`getCurrentScope().removeTag?.()` would silently no-op), and `setTag(key, undefined)` is the supported clear — undefined tag values are dropped during event serialization. The behaviour is now locked by a unit test (`applyBleDiagnosticsToScope … clears every BLE tag`) and documented in a code comment, so it can't be mistaken for storing the string `"undefined"`.

2. **`boardName == "moonboard"` case-sensitivity** (`BoardBleEncoding.swift`) — `boardName` flows from `configuration?.boardName`, which is the canonical lowercase board key (TS `BoardName` union, set via `configureBoard`). It is always lowercase, and the comparison matches the existing convention already used in `BoardBleManager.swift` (`displayItem`/`clearBoard`, lines ~616/645). Adding `.lowercased()` to only this one site would be inconsistent and imply the siblings are buggy; the value is genuinely lowercase by type.

Tightened the diagnostics types and made `setBleDiagnosticsTags(null)` clear (so the widget-adopt path drops stale tags) in the latest commit.

marcodejongh:
**Field confirmation of the root cause.** The affected tester power-cycled his phone and the board connects again **on the still-regressed build**. That points squarely at iOS's cached GATT database: CoreBluetooth caches each peripheral's characteristic *property bits* and serves them until invalidated (a full reboot clears it; a BT toggle usually doesn't, and these boards are unbonded so there's nothing to "forget").

His cached entry was missing the `.writeWithoutResponse` bit, so `00bda53a2`'s property-based selection took the write-with-response path → no `didWriteValueFor` ack → stall. The reboot cleared the cache → iOS re-read the real properties (which include write-without-response) → the regressed code then picked the right path.

Takeaways:
- This fix is the permanent answer: it stops trusting the cached property for Aurora and always uses write-without-response, so a stale cache can't break it again. The reboot is only a temporary workaround.
- It also resolves the firmware-vs-iOS question from the spec review — it's iOS cache state, not the board. The connect-time diagnostics shipped here (`ble_supports_without_response`, `characteristicProperties`) will confirm in Sentry once the tester is on a build with the fix.

PR #3234, opened 2026-06-27
feat(mobile): attach Bluetooth diagnostics to bug reports

marcodejongh:
## Summary

Users keep reporting Bluetooth problems, but analytics show most users connect fine — the failures are concentrated in a minority of devices/boards we can't see. Today a mobile bug report carries a comment plus board/climb/session context but **no BLE signal**, so a "Bluetooth doesn't work" report is unactionable.

This adds an opt-in **"Attach Bluetooth diagnostics"** switch to the mobile bug report sheet (defaults **off**). When on, the report attaches an in-session dump of the boards found and recent connection issues (connect failures, send failures, disconnects) with the write-type / MTU diagnostics we already compute at connect — so a triager sees *which* board, *which* failure mode, and *which* write path was chosen.

### How it works
- **New `ble-diagnostics-log.ts`** — a module-level ring buffer (40 events, in-session only) recording `device_found`, `connect_success`, `connect_failure`, `send_failure`, `disconnect`. It **reuses** the existing `classifyBleFailure` / `classifyBleFailureReason` / `isDisconnectionError` and the connect-time `getNativeBleConnectedDevice()` diagnostics rather than duplicating them.
- **Instrumented** `adapter.ts` (devices found) and `use-board-bluetooth.ts` (connect/send/disconnect) at the spots that already classify these failures.
- **Carried end-to-end** via `context.bleDiagnostics` (a string in the existing `app_feedback.context` jsonb column — **no DB migration**) → GraphQL → backend → Discord, rendered as a fenced field truncated to Discord's 1024-char limit.
- **Toggle** is local per-report state in `FeedbackSheet`, default off; raw device name + id are included (most debuggable, per product decision).

Scope is **mobile only** (web's bug dialog has no toggles and uses Web Bluetooth — possible follow-up). Device identifiers are sent **raw**.

## Release Notes

- When Bluetooth acts up, flip on **Attach Bluetooth diagnostics** in your bug report so we can see what your phone saw — the boards it found and where the connection dropped.

## Test plan

- `vp run typecheck:mobile`, `:backend`, `:db`, `:web` — all green.
- `vp run codegen` — generated types updated, no drift.
- `vp test run --project mobile` (2726 pass) and `--project backend` (1816 pass), incl. new suites:
  - `ble-diagnostics-log.test.ts` — buffer cap/eviction, dedupe, empty→null, write-type/MTU rendering, newest-first timeline.
  - `discord-webhook.test.ts` — fenced field rendered when present, omitted when absent, truncated under 1024.
  - `feedback-validation.test.ts` — `bleDiagnostics` accepted, capped at 2000.
  - `use-submit-app-feedback.test.ts` — enrichment includes/omits `bleDiagnostics` per flag.
- `vp check` on changed files — 0 errors; i18n catalog parity (en-US/es/fr) holds.
- Manual (pending): force a BLE failure on device, open bug report, flip the toggle on, submit, confirm the Discord message shows a Bluetooth diagnostics block; confirm a report with it off carries no BLE field.

https://claude.ai/code/session_019AH7cFT6wWPAzHriynK4XS

### DIFF
--- a/packages/backend/src/__tests__/discord-webhook.test.ts
+++ b/packages/backend/src/__tests__/discord-webhook.test.ts
@@ -137,6 +137,57 @@ describe('Discord webhook payload privacy', () => {
     expect(fields.find((f) => f.name === 'URL')?.value).toBe('/kilter/1/5/1,2/40');
     expect(fields.find((f) => f.name === 'User agent')?.value).toBe('Mozilla/5.0');
   });
+
+  it('renders a fenced Bluetooth diagnostics field when present', () => {
+    const body = buildWebhookBody({
+      feedbackId: 1,
+      rating: null,
+      comment: 'BLE flaky',
+      platform: 'ios',
+      appVersion: null,
+      source: 'drawer-bug',
+      context: { bleDiagnostics: 'Devices found (1):\n- Kilter A1B2 | id=xyz | rssi -52' },
+    });
+    const embed = (body.embeds as Array<Record<string, unknown>>)[0];
+    const fields = embed.fields as Array<{ name: string; value: string }>;
+    const ble = fields.find((f) => f.name === 'Bluetooth diagnostics');
+    expect(ble?.value).toContain('Kilter A1B2');
+    expect(ble?.value.startsWith('```')).toBe(true);
+    expect(ble?.value.endsWith('```')).toBe(true);
+  });
+
+  it('omits the Bluetooth diagnostics field when absent', () => {
+    const body = buildWebhookBody({
+      feedbackId: 1,
+      rating: null,
+      comment: 'broken',
+      platform: 'ios',
+      appVersion: null,
+      source: 'drawer-bug',
+      context: { climbName: 'My Project' },
+    });
+    const embed = (body.embeds as Array<Record<string, unknown>>)[0];
+    const fields = embed.fields as Array<{ name: string }>;
+    expect(fields.find((f) => f.name === 'Bluetooth diagnostics')).toBeUndefined();
+  });
+
+  it('keeps the Bluetooth diagnostics field within Discord 1024-char limit', () => {
+    const body = buildWebhookBody({
+      feedbackId: 1,
+      rating: null,
+      comment: 'long',
+      platform: 'ios',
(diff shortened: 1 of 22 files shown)


marcodejongh:
Thanks — addressed in 1db41de6c.

**1. Double-disconnect logging race** — fixed. You're right that relying on the adapter to suppress its disconnect event after a write failure isn't safe across stacks. Both involuntary-drop paths (the adapter's `onDisconnect` event and the `sendFramesToBoard` write-failure detection) now funnel through a single guarded record in `clearConnectionAfterDrop`, which only involuntary drops reach (a deliberate user disconnect goes through the separate `teardownConnection`, so it's never logged as a connection issue). One `disconnect` record per connection generation regardless of which path fires first or whether both fire; the send path raises the label to `stolen` before teardown so the more specific label wins when known. New test confirms a stolen write logs exactly one disconnect even when the adapter event also fires.

**3. user-cancel exclusion** — added two tests: a user-cancel connect produces **no** `connect_failure` event, and a real (non-cancel) connect failure **does**.

**2. Raw device IDs** — intentional product decision (confirmed with the user before implementation: "keep everything raw" for maximum debuggability). It's opt-in (default-off toggle, per report), stored in the existing `app_feedback.context` jsonb and surfaced to our own Discord triage webhook — same destination/retention as the other feedback context we already send. Happy to switch to truncating the id suffix if we'd rather not retain hardware identifiers; that's a one-line change in the report builder.

marcodejongh:
Thanks for the re-review.

- **`BLE_FENCE_OVERHEAD`** — fixed in c27a73008: the budget now derives from the actual `BLE_FENCE_OPEN` / `BLE_FENCE_CLOSE` pieces the template uses, so the literal and the format string can't drift.
- **French "boards"** — intentional. The existing French Bluetooth catalog already leaves "board" untranslated (`bluetooth.boardNotFound` = "Board introuvable", `bluetooth.unavailable` = "…vous connecter à votre board"), so `"les boards détectés"` matches the established register for that namespace. The glossary only fixes Spanish (→ plafón), which I followed. Happy to switch to a French term if we add one to the glossary.

PR #3249, opened 2026-06-28
fix(mobile): hide current-climb bar on the boards picker/builder

marcodejongh:
## Summary

When a climb is currently set, the root-mounted `PersistentQueueBar` floats over every screen — including the modal `/boards` stack (picker, builder, manage, edit). On the board builder it sat directly on top of the pinned create/save button: that footer only reserves the safe-area inset, not the floating bar's height, so on the boards modal (`insideTabs === false`) the bar anchors right over it. The result was that the current-climb bar constantly covered the board-creation buttons.

This suppresses the bar across the whole `/boards` stack via a new `isBoardsRoute` guard, matching the existing `auth` / `gyms` / `play` suppressions. The climb toolbar has nothing to act on while you're choosing or building a board, so hiding it there is both the fix and the correct behaviour.

- `route-segments.ts` — new `isBoardsRoute(segments)` helper (covers `index`, `create`, `manage`, `edit`)
- `persistent-queue-bar.tsx` — early-return guard alongside the existing route guards
- Tests added in `route-segments.test.ts` and `persistent-queue-bar.test.tsx`

## Release Notes

- Fixed the now-playing climb bar covering the buttons on the board picker, so you can pick or build a board again

## Test plan

- `vp test run --project mobile` (route-segments + persistent-queue-bar) — 21 passed
- Manual: with a current climb set, open the board picker and the builder; the climb bar no longer floats over the mode cards or the pinned Create button, and it reappears on returning to a tab. (Simulator verification pending — no device in this environment.)

https://claude.ai/code/session_01LQQkkKifsA6hqYj3kjVGjw

---
_Generated by [Claude Code](https://claude.ai/code/session_01LQQkkKifsA6hqYj3kjVGjw)_

### DIFF
--- a/packages/mobile/src/components/queue-control/__tests__/persistent-queue-bar.test.tsx
+++ b/packages/mobile/src/components/queue-control/__tests__/persistent-queue-bar.test.tsx
@@ -11,6 +11,7 @@ const cfg = vi.hoisted(() => ({
   onGymDiscovery: false,
   onAuthRoute: false,
   onPlayerRoute: false,
+  onBoardsRoute: false,
   currentClimbQueueItem: { climb: { uuid: 'c1', angle: 40 } } as unknown as ClimbQueueItem | null,
   wallClimb: null as null | { uuid: string; angle: number },
   variant: 'liquidGlass' as 'liquidGlass' | 'material',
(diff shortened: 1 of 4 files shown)


marcodejongh:
Thanks for the review.

- **`boards/edit` test case** — added in b1cb4df7.
- **Multi-line JSDoc** — leaving this as-is. I couldn't find a "one short line max" rule in `CLAUDE.md`, and `route-segments.ts` already documents its sibling predicates (`isTabsChromeRoute`, `isPlayerRoute`, `isGymDiscoveryRoute`) with multi-line block JSDoc. Matching that established file style keeps the new helper consistent with its neighbours.

---

Issue #3257, opened 2026-06-28
Bring the native filter-chip design to the logbook (iOS Liquid Glass)


marcodejongh:
## What

The climb list now filters through a row of native Liquid Glass chips under the title (PR #3245): real iOS `@expo/ui` menus/pickers/toggles for the highest-signal facets, plus a removable-token "receipt" row for the long tail. The **logbook** (You → Logbook) still hides every filter behind a single button that opens `LogbookFilterSheet`. The two surfaces filter the same kind of thing (a climber's own ticks vs. all climbs) but feel like different apps.

Apply the climb-list chip idiom to the logbook so filtering is glanceable and one-tap, and the two surfaces share one filter language.

## Why

- Logbook filters are invisible until you open the sheet — no sense of what's active without opening it.
- The climb list already proved the pattern (native menus, live-commit chips, token receipt). The logbook should match.
- Most of the building blocks are reusable; this is mostly composition + a small state addition, not new infrastructure.

## How the climb-list chips work today (reference)

All under `packages/mobile/src/`:

- **`components/search/FilterChipRow.ios.tsx`** — one `@expo/ui/swift-ui` `<Host>` over a horizontal `ScrollView` + `HStack` of chips. Inactive chip = `buttonStyle('glass')`; active facet = `buttonStyle('glassProminent')` + `tint(brandColors.primary)` (both guarded `if #available(iOS 26)`, so they degrade gracefully). Native `Menu` / `Picker` / `Toggle`; `menuActionDismissBehavior('disabled')` keeps multi-toggle menus open.
- **`components/search/FilterChipRow.android.tsx`** — placeholder returning `null` (keeps `@expo/ui/swift-ui`, which resolves native views at module load, off the Android bundle).
- **`components/search/FilterChipRow.types.ts`** / **`FilterChipRow.d.ts`** — shared props + an ambient declaration so `tsc` resolves the extensionless `./FilterChipRow` import while Metro picks `.ios`/`.android`.
- **`test/filter-chip-row-stub.tsx`** + a `vite.config.ts` alias — a `null` stub so Vitest's node env never mounts the SwiftUI host.
- **`components/search/FilterTokenRow.tsx`** — the removable "receipt": one pill per active *sheet-only* filter (chip-backed facets are excluded so a filter is never worded twice).
[...]

PR #3265, opened 2026-06-28
feat(mobile): logbook sort as iOS Liquid Glass chips (#3257, scoped to sort)

azuttre:
Scoped implementation of #3257: bring the climb-list filter-chip idiom to the logbook — but, per product call, **only for the sort** (Latest/Hardest), which is the one facet that earns top-level quick access. Status/grade chips and the token "receipt" row from the full issue are **not** included.

## What changed
- **Latest/Hardest are now top-level iOS Liquid Glass chips** in the logbook toolbar, so you can switch sort without opening the filter sheet. Tapping a chip commits live (via the existing `setPreset`).
- Built from native `@expo/ui/swift-ui` controls (`buttonStyle('glass'|'glassProminent')` + tint), mirroring the climb list's `FilterChipRow` — active chip is a brand-tinted prominent glass capsule, inactive is neutral glass; both `if #available(iOS 26)`-guarded so they degrade.
- **iOS-26 Liquid Glass only.** Android (and Material) keep the sheet's sort control — `LogbookSortChipRow.android.tsx` is a `null` placeholder so `@expo/ui/swift-ui` never reaches the Android bundle.
- When the chips show, the sheet's in-body Sort block is suppressed (`showSort={false}`) so sort is never worded twice.
- Sits behind the existing **`logbook-filters`** flag (the toolbar already does — no new flag).

## Reuse / pattern
Platform-split `.ios`/`.android` + `.types.ts` + `.d.ts` + a Vitest stub/alias, exactly like `search/FilterChipRow.*`. Reuses the existing `mobile.logbook.preset.*` i18n keys (no new strings) and the existing `setPreset` action (no new reducer state).

## Testing
- `vp run typecheck:mobile`, lint, format — clean
- `vp test run --project mobile` — green (new `logbook-tab-sort-chips.test.tsx`: chips render + sheet `showSort=false` on iOS glass; hidden + sheet keeps sort on Material **and** Android; hidden + toolbar absent when the flag is off; live commit through the real reducer)
- `vp run check:mobile-bundle` — **both** iOS and Android bundles export (guards against `@expo/ui/swift-ui` leaking onto Android)
- Built by an implementation subagent, then independently QA-reviewed (no blocking/should-fix findings) — but the glass states need a real iOS-26 device/sim check, which is the one thing CI can't do.

## Release Notes
In the logbook, Latest and Hardest are now quick-tap chips at the top — switch how your ticks are sorted without opening the filter sheet. (iOS; the rest of the filters stay one tap away in the sheet.)

### DIFF
--- a/packages/mobile/src/components/you/LogbookFilterSheet.tsx
+++ b/packages/mobile/src/components/you/LogbookFilterSheet.tsx
@@ -54,6 +54,10 @@ type LogbookFilterSheetProps = {
   onApply: (filters: LogbookFilterState, sort: LogbookSortState) => void;
   /** Clear the toolbar's committed climb-name search (called by Reset). */
   onClearSearch?: () => void;
+  /** Show the in-sheet Sort (Latest/Hardest) block. Defaults true; the caller
+   *  passes false when the top-level sort chips own the sort, so it isn't shown
+   *  twice (Liquid Glass). The draft still tracks/commits sort either way. */
+  showSort?: boolean;
 };
 
 type StatusKey = 'sends' | 'attempts' | 'both';
@@ -140,6 +144,7 @@ export function LogbookFilterSheet({
   currentSort,
   onApply,
   onClearSearch,
+  showSort = true,
 }: LogbookFilterSheetProps) {
   const { t } = useTranslation('you');
   const theme = useTheme();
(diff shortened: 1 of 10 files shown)


azuttre:
Thanks — fixed the real one and the edge case.

- **mobile-bundle CI failure** was actually the **variant guard** (`check:mobile-variants`), not the bundle: the raw `variant === 'liquidGlass'` compare is disallowed. Routed it through `selectByVariant(variant, { liquidGlass: true, material: false })`, matching how the climbs screen does it. `vp run check:mobile-variants` now passes. (My fault for not running that check locally — it's now in my loop.)
- **`sortPreset` custom-mode edge:** fixed — `preset` is now `null` (no chip lit) when `sort.mode !== 'preset'`, instead of falsely highlighting Latest. Added a test that hydrates a persisted `custom` sort and asserts no chip is active. (Mobile can't itself produce a custom sort; this only guards a cross-platform persisted value.)
- **Comment density:** I checked — the quoted rule ("Default to writing no comments … Don't explain WHAT") isn't in CLAUDE.md or docs (grep finds only the unrelated *release-notes* line), and block comments are pervasive in the codebase incl. the `FilterChipRow` reference this copies. That said, the headers were genuinely long, so I trimmed them to the load-bearing WHY anyway.

Validation: typecheck, lint, format, `check:mobile-variants`, both mobile bundles, and the mobile suite (sort-chip tests now 6, incl. the custom-mode case) all green.

azuttre:
Thanks — all three are non-blocking and don't need a change:

1. **Custom sort "inaccessible" on Liquid Glass** — intentional, not a regression. The mobile app has no custom-sort *builder* anywhere; the sheet's sort control was only ever the Latest/Hardest preset toggle, which the chips replace 1:1. A custom sort can only arrive from a web-persisted value, and the user escapes it by tapping either chip (which commits a preset). So nothing reachable is lost — there was never a way to *create* a custom sort on mobile.
2. **`.d.ts` FC vs memo** — kept deliberately: `FilterChipRow.d.ts:9` (the reference this mirrors) declares `FC<…>` while `FilterChipRow.ios.tsx:238` exports `memo(…)` too. It's harmless for consumers, and matching the established pattern is worth more than a tighter type here.
3. **Test pins the default implicitly** — fair. The `'recent'` expectation is intentionally pinning `DEFAULT_LOGBOOK_SORT`, so a defaults change failing it is the test doing its job. I'll add an explicit `DEFAULT_LOGBOOK_SORT` assertion next time this file is touched, rather than re-spinning the full native CI on an already-green PR just for that.

marcodejongh:
Merge conflicts

marcodejongh (dismissed):

azuttre:
Rebased on latest `main` to clear the conflict.

It was purely additive in `packages/mobile/vite.config.ts`: `main` added the SwitchRow / SegmentedControl / AngleSlider / FeatureFlagsForm / MoreForm platform-split stub aliases, this branch added the LogbookSortChipRow one. Kept both.

Re-validated locally after the rebase: variant guard, `typecheck:mobile`, mobile tests (2811 pass), and the Metro bundle check (iOS + Android) all green.

The force-push dismissed your approval (branch protection dismisses stale reviews on push). A re-approval would be appreciated once CI comes back green.

azuttre:
This looks clean: CI is green and there are no open review items. The one thing I haven't done is test it on device. TestFlight is still serving me the old build, so I'll verify once the next TestFlight version is available.

PR #3274, opened 2026-06-29
feat(mobile): climb card — logbook grade, sticky sections, copy name

azuttre:
## Summary

Three small quality-of-life improvements to the mobile climb card (the play drawer).

**1. Logbook rows show the grade you gave.** Each logbook entry already carried your personal grade (`Tick.difficulty`) but never displayed it. Adds a grade-colored chip next to the angle, formatted to your grade-display preference (V / Font / both). Hidden when no personal grade was logged (e.g. most attempts). No backend or query change — the field was already fetched.

**2. Section expand/collapse state sticks.** The logbook, community, and similar-climbs sections used to reset to their default open/closed state on every climb. Now whatever you leave them at persists across climbs and app restarts. Implemented as a `persistKey` on `CollapsibleSection` backed by a new AsyncStorage-backed `section-expand-store` (`useSyncExternalStore` singleton, same pattern as `grade-format-preference`). State is per-device (not synced to the server profile).

**3. Long-press the climb name to copy it.** The name was a plain label. Long-pressing it now copies the climb name to the clipboard with a haptic + toast — handy for pasting into a chat when sharing beta. Only the active header is interactive; the swipe "peek" header stays a plain label.

## Release Notes

- See the grade you gave each climb right in its logbook
- Logbook, community, and similar-climbs sections now stay open if you leave them open
- Long-press a climb's name to copy it

## Test plan

- `vp run test:mobile` — 2789 passing, plus new tests:
  - `section-expand-store.test.ts` (load / set / malformed-payload fallback / no-clobber race / merge)
  - `logbook-entry-row-grade.test.tsx` (chip shown with a grade, hidden without)
- `vp lint` on changed files — 0 warnings / 0 errors
- i18n parity (`catalog-completeness.test.ts`) — `nameCopied` / `copyNameHint` present in en-US, es, fr

Not yet exercised on a simulator/device — happy to add screenshots if useful.

### DIFF
--- a/packages/mobile/src/components/CollapsibleSection.tsx
+++ b/packages/mobile/src/components/CollapsibleSection.tsx
@@ -4,6 +4,7 @@ import Animated, { FadeIn, FadeOut, useAnimatedStyle, useSharedValue, withTiming
 import { Text } from './Text';
 import { Icon } from './Icon';
 import { hapticSelection } from '../lib/haptics';
+import { getSectionExpandedSync, setSectionExpanded, useSectionExpanded } from '../lib/section-expand-store';
 import { iosSystemColors } from '../theme/ios-colors';
 import { spacing, borderRadius } from '../theme/tokens';
 import { timing } from '../theme/animations';
@@ -28,6 +29,11 @@ type CollapsibleSectionProps = {
   /** Fires when the section expands/collapses (and once on mount). Lets a caller
    *  gate expensive content (e.g. a query) on the section being open. */
   onExpandedChange?: (expanded: boolean) => void;
+  /** When set, the section's expand state persists under this key (across climbs
+   *  and app restarts) instead of resetting to `defaultExpanded` each mount.
+   *  `defaultExpanded` is then only the first-ever state before the user touches
+   *  the section. Omit to keep the section's state ephemeral. */
+  persistKey?: string;
   children: ReactNode;
 };
 
@@ -40,6 +46,7 @@ export function CollapsibleSection({
   headerAction,
   onHeaderLayout,
   onExpandedChange,
+  persistKey,
   children,
 }: CollapsibleSectionProps) {
   const handleHeaderLayout = useCallback(
@@ -72,6 +79,7 @@ export function CollapsibleSection({
       headerAction={headerAction}
       onHeaderLayoutEvent={onHeaderLayout ? handleHeaderLayout : undefined}
       onExpandedChange={onExpandedChange}
+      persistKey={persistKey}
     >
       {children}
     </CollapsibleSectionInternal>
@@ -86,6 +94,7 @@ function CollapsibleSectionInternal({
   headerAction,
   onHeaderLayoutEvent,
   onExpandedChange,
+  persistKey,
   children,
 }: {
   title: string;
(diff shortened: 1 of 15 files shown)


azuttre:
Addressed all three review findings in f980dcde:

1. **Haptic on copy** — the haptic was actually firing already (`showToast(…, 'success')` triggers `hapticSuccess()` in `toast-provider`), so the behavior matched the description. But it was implicit. Made it explicit: `handleCopyName` now calls `hapticSuccess()` directly and uses an `info` toast, matching the "Link copied" affordance in `ClimbActionsSheet` rather than depending on the toast variant's side-effect. (Note: only `hapticSuccess` was imported in `PlayDrawer`, not `hapticSelection`.)

2. **Copy-name test** — added `play-drawer-header-copy-name.test.tsx`: long-pressing the name fires the handler, and the swipe "peek" header (no handler) renders the name with no interactive target.

3. **gradeColor** — guarded the `getGradeColor` lookup behind `gradeLabel` so no-grade rows (most attempts) skip it.

azuttre:
Addressed the second review pass in 387a92b2:

1. **Side-effect in `setState` updater (CollapsibleSection)** — moved the Reanimated `withTiming` (and the persist/haptic) out of both `setExpanded` updaters (toggle + persisted-sync effect). They now run before an unconditional `setExpanded`, with `expanded` in the closure/deps, so a StrictMode double-invoke can't fire the animation twice.

2. **`handleCopyName` integration untested** — extracted the clipboard/haptic/analytics/toast orchestration into `copy-climb-name.ts` and added `copy-climb-name.test.ts` (asserts the clipboard write, haptic, `ClimbShared` event payload, and toast; plus the no-op paths for null/empty name). `PlayDrawer.handleCopyName` now delegates to it.

3. **Unhandled AsyncStorage read failure (section-expand-store)** — guarded the read with `.catch(() => null)` like the write, so a rejected `getItem` can't leave the load singleton permanently rejected. Added a read-failure resilience test.

4. **Hardcoded storage key in test** — exported `STORAGE_KEY` and the test imports it now.

5. **Short-tap feedback (PlayDrawerHeader)** — non-issue: the name `Pressable` uses a static style with no pressed-state and no `android_ripple`, so a short tap produces no highlight (RN `Pressable` has no default visual feedback, unlike `TouchableOpacity`). Left as-is.

Full mobile suite green (2819 tests).

azuttre:
Addressed the third pass in 216ece21 (all were flagged non-blocking):

1. **Fire-and-forget clipboard write (`copy-climb-name.ts`)** — `copyClimbName` is now async: it awaits `Clipboard.setStringAsync` and only fires the haptic/analytics/toast on success, so a failed write can't show a false "Name copied". `handleCopyName` calls it fire-and-forget (`void`). Added a write-failure test asserting no toast/haptic/track when the write rejects.

2. **Cold-load reconciliation untested (`CollapsibleSection`)** — added `collapsible-section-persist.test.tsx`: one test drives the cold path (seed AsyncStorage, render collapsed, assert the async load flips the section open via the sync effect) and one covers the warm synchronous seed.

3. **`loadPromise` could be created twice (`section-expand-store`)** — `ensureSectionExpandLoaded` now short-circuits with `Promise.resolve(current)` when `hasLoaded`, so the load-once intent holds even when a write precedes the first mount.

Full mobile suite green (2824 tests).

azuttre:
Addressed the fourth pass in b2713b4d:

1. **Grade-chip color coupled to display format (`LogbookEntryRow`)** — now resolves the color from the raw `entry.difficulty` id via its canonical `"6a/V3"` name, not the display-formatted label. (`getGradeColor` actually does extract the V-grade from the `"V6 / 7a"` combined string today, but keying off the canonical id removes the coupling entirely.)

2. **Silent clipboard failure (`copy-climb-name.ts`)** — `copyClimbName` now shows an error toast (`copyNameError`, added to en-US/es/fr) when the write rejects, so a failed long-press copy isn't a silent no-op. Test asserts the error toast fires.

3. **Test ordering dependency (`collapsible-section-persist.test.tsx`)** — added `resetSectionExpandStoreForTests` + an AsyncStorage reset in `beforeEach`, so each test owns its store state. Verified it passes in isolation, not just in sequence.

Full mobile suite green (2822); the file passes standalone too.

azuttre:
CI is green and this is the "Ready to merge" review, so wrapping up the three notes:

1. **`toggleExpanded` recreation** (CollapsibleSection): leaving as-is. Recreating one small component's callback per toggle is negligible, and the StrictMode-safe version is worth it.

2. **Board-agnostic persist keys**: intentional and confirmed. Section expand/collapse state is global by design so it follows you across boards: collapse the logbook once and it stays collapsed whether you open a Kilter or a Tension climb.

3. **Simulator/device validation**: I couldn't run `check:mobile-simulator` in this environment (no `xcrun simctl`), so the play-drawer layout is covered only by the headless `check:mobile-bundle` (iOS + Android), not a visual check. I also haven't verified on device yet: TestFlight is still serving the older build and the 2.1.1 build at the current fingerprint hasn't propagated to me, so the grade chip and the long-press name wrapper are visually unverified. I'll confirm once the next TestFlight build is available.

Otherwise this looks clean: CI green, no open review items.

PR #3276, opened 2026-06-29
feat(mobile): move What's New into the user drawer

marcodejongh:
## Summary

On mobile, "What's New" (the changelog) lived in the More tab's About section. This moves it into the **user drawer** — the slide-in menu opened from the avatar in the top toolbar — so it sits alongside the rest of the account/app menu (Settings, My Playlists, About, Rate, Report a bug, Join Discord, Log out). The "New" unseen pill moves with it.

What changed:
- `UserDrawerProvider` gains a `navigateToChangelog` action; the drawer renders a "What's New" row just above "About". `DrawerRow` gained a `trailing` slot so the row can show the "New" pill, computed on open from the existing `getLastSeenChangelogDate()` / `hasUnseenChangelog()` helpers.
- The More tab drops the changelog row, its unseen-state effect, and the now-dead imports/styles. Its lone "About" section goes away with it.
- i18n: replaced the orphaned `mobile.more` changelog/about keys with `userDrawer.whatsNew` and `userDrawer.newBadge` across `en-US`, `es`, and `fr`.

This is mobile-only; the web app has no changelog feature.

## Release Notes

- Find What's New in the user menu now — tap your avatar to see the latest updates, with a badge when there's something fresh.

## Test plan

- `vp run test:mobile` — 2784 tests pass (includes the new What's New → `/changelog` navigation case in the user-drawer test and the i18n catalog completeness check)
- `vp lint` — changed files clean (only pre-existing unrelated warnings/error)
- Manual: avatar → drawer → "What's New" (next to About) opens `/changelog`; the More tab no longer shows it; the "New" pill appears for unseen entries and clears after viewing; `es`/`fr` show translated labels.

---
_Generated by [Claude Code](https://claude.ai/code/session_018VKhhjK4v9r1FtPiwKzhVE)_

### DIFF
--- a/packages/mobile/app/(tabs)/profile/more.tsx
+++ b/packages/mobile/app/(tabs)/profile/more.tsx
@@ -1,5 +1,4 @@
-import { useCallback, useState } from 'react';
-import { router, useFocusEffect } from 'expo-router';
+import { router } from 'expo-router';
 import { StyleSheet, View } from 'react-native';
 import { useTranslation } from 'react-i18next';
 import type { GradeDisplayFormat } from '@boardsesh/play-view';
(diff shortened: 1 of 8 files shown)


marcodejongh:
On the storage-read point: `getLastSeenChangelogDate()` already wraps the SecureStore read in a `try/catch` and returns `null` on any failure ([changelog-seen.ts L16-22](https://github.com/boardsesh/boardsesh/blob/main/packages/mobile/src/lib/changelog-seen.ts#L16-L22)), so it never rejects. `hasUnseenChangelog` is pure and `setChangelogUnseen` can't throw, so the `.then()` chain can't surface an unhandled rejection — and a thrown read already folds to the badge-hidden fallback. This mirrors the existing pattern the More tab used, so I've left it as-is rather than add a redundant `.catch()`.

---

PR #3278, opened 2026-06-29
feat: case-insensitive email + duplicate-account merge tooling

marcodejongh:
## Summary

`users.email` had no unique constraint and used the default (case-sensitive) collation, so `Foo@x.com` and `foo@x.com` created **separate accounts**. A read-only prod audit found **23 duplicate-by-case account sets** (47 user rows). When someone later links the same board on both twins, the second account shows an empty logbook — the support report that kicked this off.

This is **PR 1 of 2** — the foundation, safe to deploy on its own:

- **Shared `normalizeEmail` helper** at `@boardsesh/db/utils` (trim + lowercase). The backend native-auth handler now routes through it (replacing 3 inline copies).
- **Web auth normalized end-to-end**: email is lower-cased on every write (register insert, email-verification and `password-reset:` token identifiers) and every lookup is case-insensitive — register existence check, credentials login (verifies the password against each candidate so a transient duplicate set can't lock anyone out), forgot/reset/verify/resend, and the `signIn` callback. The NextAuth `DrizzleAdapter` is wrapped so OAuth account-linking and user creation lower-case email too. **This freezes creation of new duplicate accounts.**
- **`merge-accounts.ts`** (`vp run db:merge-accounts`, dry-run by default, `--apply` to write): collapses each `lower(email)` group onto a winner (most ticks → most other rows → verified → oldest), repointing every owned row across all user-referencing tables, deduping the user-scoped unique constraints, and handling the self-follow / shared-board / ticks-surrogate / credential / profile edge cases. Per-set transaction + advisory lock; the dry-run report flags anomalous sets (3-row, both-have-password, both-have-ticks). Relies on the existing `votes_count_trigger` to keep `vote_counts` correct.

The lowercase backfill + `CREATE UNIQUE INDEX … (lower(email))` and the operational merge run land in a follow-up PR, after the merge has been dry-run-reviewed and applied to prod (deleting users is irreversible — pg_dump first). **No migration in this PR mutates data.**

Also unblocks the #3526 cross-linked-account repair, which needs this merge tooling.

## What changed in the rebase

Rebased onto current `main`. The one real conflict was `auth-options.ts`, which `main` reworked twice since the branch was cut (JWT profile-claims caching in 9c34fc163, the shared parent-domain session cookie in 05f74cc40). Both are preserved verbatim; the adapter wrap and the multi-candidate credentials login are re-applied on top.

The FK-completeness guard earned its keep during the rebase: `main` has since added `app_feedback.resolved_by` and `gym_merge_audit.performed_by`, and the script aborted rather than silently skipping them. Both are `ON DELETE SET NULL` staff-attribution columns, so they join the plain repoints — the audit trail names the surviving account instead of going blank when the loser row is deleted.

Review nits from the last pass are addressed: a note on why the adapter's verified-first ordering can legitimately differ from the merge script's winner ranking (and why nothing is lost when it does), a comment scoping the FK guard to single-column FKs, and unit tests for the adapter wrapper.

## Release Notes

- Sign in with your email however you capitalise it. `You@Email.com` and `you@email.com` are the same account now, so you won't land in a second, empty profile with none of your sends in it.

## Test plan

- `vp run typecheck` — green (all 53 tasks).
- `vp check` — 0 errors.
- Unit: `normalizeEmail` (case/trim/idempotence/guard); web auth suites 188/188 across 15 files, including 4 new adapter-wrapper tests (case-insensitive match on a mixed-case stored row, the recorded `lower(email)` predicate, verified-then-oldest ordering, and `createUser` lower-casing before it delegates); backend native-auth.
- **Integration** (`merge-accounts.integration.test.ts`, real Postgres): seeds a winner+loser+third with rows across ticks, favorites, mutual follows (self-follow trap), credentials, profiles, votes (dedup + trigger-maintained counts), a loser-owned gym (cascade-delete trap), a Tension board link, ratings, and a beta link — asserts every row lands on the winner, uniques dedupe, no self-follow, and the loser is deleted. Plus the FK-coverage test that runs the guard against the live migrated schema, so the repoint lists can't drift undetected again. Passes, rolls back cleanly.
- The merge script was **not** run against any database in this rescue. It stays dry-run by default (`apply: false`); `--apply` is the only write path and returns early without it.

https://claude.ai/code/session_01Xu9FhyVPEPTuBSGBigiDgd

### DIFF
--- a/packages/backend/src/handlers/native-auth.ts
+++ b/packages/backend/src/handlers/native-auth.ts
@@ -4,6 +4,7 @@ import { SignJWT, createRemoteJWKSet, jwtVerify } from 'jose';
 import { compare, hash } from 'bcryptjs';
 import { eq, and, isNull, lt, or, isNotNull } from 'drizzle-orm';
 import { mobileRefreshTokens, users, userCredentials, accounts, userProfiles } from '@boardsesh/db/schema/auth';
+import { normalizeEmail } from '@boardsesh/db/utils';
 import { db } from '../db/client';
 import { redisClientManager } from '../redis/client';
 import { applyCorsHeaders } from './cors';
@@ -523,7 +524,7 @@ export async function handleNativeAuthCredentials(req: IncomingMessage, res: Ser
   }
 
   // Normalize the email to match how NextAuth's adapter stores it.
-  const normalizedEmail = email.trim().toLowerCase();
+  const normalizedEmail = normalizeEmail(email);
 
   try {
     const userRows = await db.select().from(users).where(eq(users.email, normalizedEmail)).limit(1);
@@ -625,7 +626,7 @@ export async function handleNativeAuthRegister(req: IncomingMessage, res: Server
     sendJson(res, 400, { error: 'email and password are required' });
     return;
   }
-  const normalizedEmail = email.trim().toLowerCase();
+  const normalizedEmail = normalizeEmail(email);
   if (!REGISTER_EMAIL_REGEX.test(normalizedEmail)) {
     sendJson(res, 400, { error: 'Invalid email address' });
     return;
@@ -893,7 +894,7 @@ async function findOrCreateOAuthUser(
       return { status: 'ok', tokenPair, userId: linked.userId };
     }
 
-    const normalizedEmail = identity.email ? identity.email.trim().toLowerCase() : null;
+    const normalizedEmail = identity.email ? normalizeEmail(identity.email) : null;
 
     // 2. Link by email — ONLY when the provider asserts the email is verified.
     // This is the security boundary NextAuth's allowDangerousEmailAccountLinking
--- a/packages/db/package.json
+++ b/packages/db/package.json
@@ -60,6 +60,11 @@
       "types": "./src/test-utils/sql-text.ts",
       "import": "./src/test-utils/sql-text.ts",
       "default": "./src/test-utils/sql-text.ts"
+    },
+    "./utils": {
+      "types": "./src/utils/index.ts",
+      "import": "./src/utils/index.ts",
+      "default": "./src/utils/index.ts"
     }
   },
   "scripts": {
@@ -84,6 +89,7 @@
     "db:dedupe-gyms": "tsx scripts/dedupe-gym-locations.ts",
     "db:dedupe-beta-links": "tsx scripts/dedupe-beta-links.ts",
     "db:refresh-climb-grades": "tsx scripts/refresh-climb-grades.ts",
+    "db:merge-accounts": "tsx scripts/merge-accounts.ts",
     "db:seed-social": "tsx scripts/seed-social.ts",
     "db:create-test-user": "tsx scripts/create-test-user.ts",
     "db:seed-beta-links": "tsx scripts/seed-dev-beta-links.ts"
--- a/packages/db/scripts/merge-accounts-args.test.ts
+++ b/packages/db/scripts/merge-accounts-args.test.ts
@@ -1,0 +1,48 @@
+import { test } from 'node:test';
+import assert from 'node:assert/strict';
+import { parseArgs } from './merge-accounts.js';
+
+test('no args is a dry-run (apply stays false)', () => {
+  const parsed = parseArgs([]);
+  assert.equal(parsed.apply, false);
+  assert.equal(parsed.limit, null);
+  assert.equal(parsed.onlyEmail, null);
+  assert.equal(parsed.help, false);
+});
+
+test('--apply, --limit and --only-email parse; only-email is normalized', () => {
+  const parsed = parseArgs(['--apply', '--limit', '5', '--only-email', '  Foo@X.com ']);
+  assert.equal(parsed.apply, true);
+  assert.equal(parsed.limit, 5);
+  assert.equal(parsed.onlyEmail, 'foo@x.com');
+});
+
+test('a leading `--` separator is skipped, not treated as an unknown argument', () => {
+  assert.deepEqual(parseArgs(['--', '--apply']), parseArgs(['--apply']));
+});
+
+test('--help sets help', () => {
+  assert.equal(parseArgs(['--help']).help, true);
+});
+
+// The failure paths call process.exit(2). Stub it (and silence console.error)
+// so the test can assert they bail instead of returning a bad ScriptArgs.
+function assertExits(args: string[]): void {
+  const originalExit = process.exit;
+  const originalError = console.error;
+  process.exit = ((): never => {
+    throw new Error('process.exit');
+  }) as typeof process.exit;
+  console.error = () => {};
+  try {
+    assert.throws(() => parseArgs(args));
+  } finally {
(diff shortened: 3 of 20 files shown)


marcodejongh:
Thanks — both addressed in 0deca47a3:

1. **FK-completeness guard.** Added `assertAllUserFksHandled()`, which runs before any merge work (dry-run and `--apply`): it enumerates every FK to `users(id)` from `pg_catalog` and diffs against the repoint lists + the hand-coded specials + an explicit cascade-only allow-list (`user_climb_percentiles`). If a future migration adds a user FK the script doesn't handle, it aborts and names the offending column(s) instead of silently skipping them. A new integration test runs it against the live migrated schema, so the lists can't drift out of date undetected.

2. **Login candidate loop.** Added a comment explaining the bound: candidate count is the number of real duplicate-by-case signups for one address (max 3 in the prod audit, always 1 after the merge), and it isn't attacker-inflatable since each row requires a completed signup — so no explicit cap is needed.

marcodejongh:
All addressed in d9b65ee68:

1. **`verify-email` too-broad update** — now marks `emailVerified` by `user.id`, so only the account that initiated verification is touched.
2. **In-flight tokens across the deploy** — fixed in code rather than just documented: both `verify-email` and `reset-password` now match the token identifier against **both** the normalized and the original-cased value (`inArray`), so links minted before the normalization deploy still resolve instead of failing with "invalid token." (Verification is also disabled by default in prod, and reset tokens are 1h TTL, so the window was already narrow.)
3. **`buildDuplicateSets` / `parseArgs` bypassing `normalizeEmail`** — both now route through the helper. Added `merge-accounts-args.test.ts` covering the happy paths **and** the `process.exit(2)` cases (unknown flag, `--limit 0`, missing `--limit`/`--only-email` values).
4. **`base.createUser!`** — added a comment documenting the invariant (DrizzleAdapter always defines it).

Re-validated: db typecheck + web typecheck green; web auth 97/97, db unit+integration 15/15 (incl. the new args tests and the gym_claims dedupe).

Note: the earlier `changes`/`large-files`/`ci-status`/`ota-check` failures were a GitHub Actions infra hiccup (API was returning 503; the real test jobs *skipped* rather than ran). This push re-triggers a clean run.

marcodejongh:
Rebased onto current `main` (the branch had gone `CONFLICTING`) and worked through this review. Three of the four addressed in 95e5209f0; the first one doesn't hold.

**1. `reset-password` legacy identifier fallback — not fragile, the premise is off.**

The review assumes the email is "what the user types now in the reset form". It isn't typed at all. `reset-password-content.tsx` reads it from the link:

```ts
const token = searchParams.get('token') || '';
const email = searchParams.get('email') || '';
```

and posts that value straight through. So `validationResult.data.email` is whatever `forgot-password` embedded when it minted the link. Pre-deploy, `main`'s `forgot-password` used the same raw `email` for **both** the identifier and `sendPasswordResetEmail(email, …)` — so a legacy link carries exactly the casing its legacy identifier was built from, and the fallback matches. Post-deploy both sides are normalized and the fallback collapses to a single candidate. There's no path where the user re-types the address, so no re-typing window to get wrong. No change made.

**2. No tests for `verify-email` / `resend-verification`.** Right about `verify-email` — added `verify-email/__tests__/route.test.ts` with 6 cases: case-insensitive match against a mixed-case stored row, the raw-cased legacy identifier fallback, no duplicate candidate when the link email is already canonical, `emailVerified` stamped by `users.id` (not by `lower(email)`), and the invalid/expired branches. Mutation-checked — collapsing `candidateIdentifiers` back to `[normalizedEmail]` fails the fallback test, so it isn't a tautology.

`resend-verification` already has `__tests__/route.test.ts`; its change is a one-line swap to `normalizeEmail` on a path those tests already drive, so I left it alone rather than adding a test that asserts a lowercase string is lowercase.

**3. `MemberRow` vs `Member`.** Left as-is. The split is deliberate: `MemberRow` is the wire shape (`postgres` returns `count(*)` as a string), `Member` is the coerced shape the winner-picking logic uses. Collapsing them would mean either coercing at every read site or letting `number | string` leak into `compareForWinner`, where a string comparison would silently pick the wrong winner. The parallel type is what keeps that from compiling.

**4. `Date.now()` fixture tag.** Fixed — now `randomUUID()`.

**Also from the rebase:** the FK-completeness guard earned its keep. `main` has added `app_feedback.resolved_by` and `gym_merge_audit.performed_by` since the branch was cut, and the script aborted instead of silently skipping them. Both are `ON DELETE SET NULL` staff-attribution columns, so they join the plain repoints — the audit trail names the surviving account rather than going blank. `auth-options.ts` was the only real conflict; `main`'s JWT profile-claims caching and shared parent-domain cookie work are preserved verbatim, with the adapter wrap and multi-candidate login re-applied on top.

Scope is unchanged: still part 1 of 2, no migration in this PR mutates data, and the merge script stays dry-run by default with `--apply` as the only write path. It was not run against any database.

marcodejongh:
All three addressed in 90cda3517 — comment-only, no behaviour change:

1. **`applyMerge` transaction requirement** — added a JSDoc saying it must run inside a transaction, and why the type can't enforce it (`FOR UPDATE` only holds for the life of a transaction; `ExecuteDb` is structural and matches a plain connection too). `main` passes a real `db.transaction(...)` handle and the integration test drives it inside a rolled-back one.
2. **`getUserByEmail` returning the stored casing** — noted that this is deliberate (the claim should reflect the row), that a legacy row stays mixed-case on `session.user.email` until the PR-2 backfill, and that downstream comparisons need `lower()` on both sides.
3. **`reset-password` test stub** — documented why the stub skips normalization: it makes the identifier a pure function of its argument, so the tests can observe *which* email string the route passes in (normalized vs. raw-cased, which is the whole point of the pre-deploy fallback). Noted the trade-off and where the real helper's normalization is covered.

On the untested `applyMerge` "set membership changed since the dry-run" guard: agreed it's a gap, but exercising it needs a concurrent writer racing a `FOR UPDATE` inside a rolled-back transaction, which the current single-transaction harness can't stage. Since the guard's failure mode is refusing to merge (it returns `{ merged: false }` and the operator sees a skip line), leaving it uncovered doesn't risk data. Not worth reshaping the harness for in this PR.

CI green at `90cda3517`: 16491 tests, 0 failures. Marked ready for review.

Reminder for whoever merges: the merge script has **not** been run against any database. It stays dry-run by default; `--apply` is the only write path, and the operational run + the `lower(email)` unique index land in PR 2 — `pg_dump` first, since deleting users is irreversible.

marcodejongh:

Conflicting: one mechanical vite.config.ts conflict resolved clean (bun→pnpm task rename), but hit a real ~112-line content conflict in packages/web/app/lib/auth/__tests__/auth-options.test.ts right after — aborted the whole rebase rather than guess at that merge.

marcodejongh:
Redrafted for the current-main rescue. Publication remains blocked on the legacy verification/reset token ambiguity decision and remaining isolated validation; no account operation or merge-tool apply has run.

Issue #3275, opened 2026-06-29
Epic: finish the @expo/ui native-UI migration (remaining work)


marcodejongh:
Tracks the remaining work to standardise `packages/mobile` UI on native **@expo/ui** (SwiftUI on iOS, Jetpack Compose on Android), after the first two PRs landed.

## Shipped
- **#3254** — native primitives: `SwitchRow`→`Toggle`, `SegmentedControl`→`Picker(segmented)`, `AngleSlider`→`Slider`, plus the foundation (platform-split convention, theming bridge, test-stub pattern, the CI platform-import guard, `docs/expo-ui-components.md`).
- **#3261** — the **More** and **Feature Flags** settings screens as native `Form`s; Material XML icon drawables; the in-app light/dark `colorScheme` Host fix; iOS polish.

## Principles (these decide what's in scope)
- **Only migrate where it _legitimately improves the user experience_** — native keyboards/menus/glass, native accessibility, real platform feel. If a swap is marginal or risks a deliberate custom design, it's not worth it.
- **Never migrate `FlashList`/`FlatList`** (performance) — the native `List` is not for large lists.
- **Keep as React Native:** SVG board-art renderers, gifted-charts, gesture-driven boards (holds/zone), and any custom design that the native control can't match.
- Every new `@expo/ui` usage is **JS-only / OTA-deliverable** (the module is already autolinked) — except removing a native dep (e.g. `react-native-context-menu-view`), which needs a build.

## The honest state
The high-confidence wins (toggles, segmented controls, slider, settings forms) are **done**. What remains is mostly **marginal or design-sensitive** — each issue below leads with an honest "Why this improves the experience" so you can triage rather than re-litigate. Drafting agents rated 7 of 8 as `maybe`.

## Remaining issues, tiered

**Do when convenient (clearest remaining wins):**
- #3267 — native **menus** (`Menu`/`DropdownMenu`); also drops `react-native-context-menu-view`. _Android is the real win; iOS is already a native UIMenu._
- #3268 — native **selection controls** (`RadioGroup`→`Picker`/`RadioButton`, `Stepper`). _iOS Stepper (press-hold) + the accuracy radio are real wins._

**Higher value but higher risk — do carefully:**
- #3266 — native **text inputs** (`TextField`/`SecureField`). _Android-first; iOS loses the password reveal toggle and is a deliberate design; **login is the critical flow** — verify on device._

**Design-gated / flag-coupled:**
[...]

Issue #3270, opened 2026-06-29
Mobile (@expo/ui): native iOS Stepper for the generator's grouped steppers (keep the rest of the filter/generator sheets RN)


marcodejongh:
> Part of the **@expo/ui native-UI migration** follow-up (after #3254 + #3261). Maintainer guidance: only migrate where it *legitimately improves the experience*; FlashList/FlatList lists, SVG board-art, charts, and gesture boards stay React Native. Pattern: `docs/expo-ui-components.md`.

> **Drafting-agent verdict:** `maybe` — Marginal area: toggles/segments are already native, chips/grade-rails/StarRating are deliberate custom or gesture-driven, and a whole-sheet SwiftUI Form is structurally blocked by the sheets' own scroll + pinned footer. Only real win is the iOS native Stepper (press-and-hold repeat) for GroupedSteppers — and it's iOS-only (no Compose Stepper) with a value-readout regression risk. Low priority; acceptable to close as won't-do.

---

## Why this improves the experience

**Honest verdict: marginal, low priority — do the one narrow piece (iOS native `Stepper`) or nothing.** Most of what these two sheets render is *already* native after PR #3254 (every `SwitchRow` is a native `Toggle`, every `SegmentedControl` is a native segmented `Picker`). What's left is hand-built chrome and deliberate custom controls, and the big-bang "make the whole sheet a SwiftUI `Form`" is structurally impossible here (see Out of scope). So this issue is deliberately small.

The single defensible UX win:

- **`GroupedSteppers` in the generator → native iOS `Stepper`.** The current control (`src/components/Stepper.tsx`) is a custom −/+ pill. A real SwiftUI `Stepper` adds **press-and-hold-to-repeat**, the system increment "tick" haptic, and the correct VoiceOver adjustable trait/announcement — none of which the custom one does. For a control you tap repeatedly (main-set climbs 1–50, steps 3–15), press-and-hold is a genuine improvement.

Everything else in these sheets is either already native, deliberate custom design, or gesture/glyph-based — keep as RN (justified below). If the implementer finds the iOS-only `Stepper` split isn't worth the asymmetry, **closing this as won't-do is an acceptable outcome.**

## Current state

Two screens render grouped toggle/segmented/slider/chip/stepper sections. Neither is a settings *route* — both are bottom **sheets / cards** with their own scroll body, pinned footers, inline grade rails, collapsible sections, and gesture sub-pickers.

[...]

PR #3280, opened 2026-06-29
feat(mobile): native @expo/ui RadioGroup (Picker / RadioButton)

### LINKED ISSUE #3268: Mobile (@expo/ui): migrate RadioGroup → native Picker(inline)/RadioButton group and Stepper → native Stepper/IconButton pair
> Part of the **@expo/ui native-UI migration** follow-up (after #3254 + #3261). Maintainer guidance: only migrate where it *legitimately improves the experience*; FlashList/FlatList lists, SVG board-art, charts, and gesture boards stay React Native. Pattern: `docs/expo-ui-components.md`.

> **Drafting-agent verdict:** `maybe` — Stepper→iOS native Stepper (press-hold repeat + adjustable a11y) and the accuracy RadioGroup→inline Picker are real wins; but Android Stepper is a reassembly with little gain, iOS inline Picker can't express the status filter's per-option disable+description, and the native Stepper has no inline value slot. Checkbox has no call site. Medium priority — do it but make the two design-decision calls first.

---

## Why this improves the experience

Two single-purpose selection controls in `packages/mobile`, with an honest split on how much each migration buys:

- **Stepper → iOS SwiftUI `Stepper`**: a real win on iOS. The native stepper gives **press-and-hold auto-repeat** (hold `+` to ramp a count instead of tapping N times — our generator steppers go up to large climb counts) and the platform **"adjustable" accessibility trait** with native increment/decrement actions and value announcement. Our custom control only does discrete taps.
- **RadioGroup → iOS SwiftUI `Picker` (`inline` style)**: a moderate win. We get proper SwiftUI single-choice picker semantics, native selection animation, and system-managed a11y (announced as a picker, not a pile of hand-rolled `radio` Pressables). The current custom control is already a faithful native-look list (grouped inset card, checkmark, hairline rows, haptics), so the visual delta is small — the real gain is correct a11y + consistency with the rest of the migration.

**Where this is marginal / risky (be honest):**

- **Android has no native Stepper.** The Android side is a *reassembly* of label + value `Text` + two `IconButton`s — essentially what `Stepper.tsx` already does with `PressableSurface`, so the Android UX gain is small (native M3 ripple/sizing only).
- **iOS inline `Picker` can't express per-option `disabled` or a per-option description.** The accuracy RadioGroup uses neither (clean migration). The **status** RadioGroup disables the `drafts` option when signed out and shows a "Sign in for drafts" description — neither survives a native inline Picker. That needs a call-site decision (below), or status stays RN.
- **Checkbox: nothing to migrate.** There is no standalone checkbox control in the app today. The only `accessibilityRole="checkbox"` surfaces are virtualized **list rows** (`app/(tabs)/climbs/setters.tsx` SetterRow, `src/components/QueueItemRow.tsx`), which are explicitly out of scope (lists are never migrated). So do **not** build a `Checkbox` component in this issue.

Net: ship the Stepper (iOS win) and the accuracy RadioGroup (clean); treat the status RadioGroup and the Android reassembly as the judgement calls.

### RadioGroup (`src/components/RadioGroup.tsx`)
Custom RN: a `View` container of `Pressable` rows, each `accessibilityRole="radio"` with `accessibilityState={{ selected, disabled }}`, a label + optional `description` footnote, a brand `check.small` `Icon` on the selected row, hairline dividers, 44pt rows, `hapticSelection()` on press. Public API:
```ts
export type RadioOption<T extends string> = { value: T; label: string; description?: string; disabled?: boolean };
type RadioGroupProps<T extends string> = { options: ReadonlyArray<RadioOption<T>>; value: T; onChange: (value: T) => void };
```
- Exported from `src/components/index.ts:25` (`export { RadioGroup, type RadioOption } from './RadioGroup';`).
- Call sites — both in `src/components/ClimbFilterSheet.tsx` (inside a `BottomSheetScrollView` from `@expo/ui/community/bottom-sheet`, one per section, **not** a native Form):
  - `:32` import; `:251` `accuracyOptions` (`RadioOption<GradeAccuracyValue | 'off'>`, **no descriptions, no disabled**); `:775` `<RadioGroup options={accuracyOptions} … />`.
  - `:240` `statusOptions` (`RadioOption<StatusFilter>`, **uses `disabled`** for `drafts` when `!isAuthenticated` **and a `description`** = `t('mobile.filter.signInForDrafts')`); `:805` `<RadioGroup options={statusOptions} … />`.
- Test: `src/components/__tests__/climb-filter-sheet.test.tsx:263` does `vi.mock('../RadioGroup', () => ({ RadioGroup: () => null }))`.

### Stepper (`src/components/Stepper.tsx`)
Custom RN: label left, value `Text` + a −/+ pill (two `PressableSurface` buttons with `minus`/`plus` `Icon`s) trailing. Clamps to `[min,max]` via `clampStepperValue`, disables − at `value<=min` and + at `value>=max`, Material/Glass variant styling, haptics via `PressableSurface`. Public API:
```ts
type StepperProps = { label: string; value: number; min: number; max: number; onChange: (n: number) => void; decreaseLabel: string; increaseLabel: string };
```
- **Not** exported from `index.ts`; imported directly.
- Call site — `src/components/session-screen/pre-session/GeneratorPickerCard.tsx`: `:38` import; `:441` `<Stepper … />` built by the `stepperRow()` helper and rendered (multiple instances) inside `GroupedSteppers` (a RN `View` grouped card, hairline-divided) — primary counts + a collapsed "Tuning" section (min-ascents, etc.). All inside a scroll view, not a list.
- Tests: `src/components/__tests__/stepper.test.tsx:50` (renders `Stepper`, asserts clamp/onChange); `src/components/session-screen/pre-session/__tests__/generator-picker-card-analytics.test.tsx:179` does `vi.mock('../../../Stepper', () => ({ Stepper: () => null }))`.

### RadioGroup
- **iOS** — `Picker` from `@expo/ui/swift-ui` with `pickerStyle('inline')`. Each `RadioOption` → a `<Text modifiers={[tag(option.value)]}>{label}</Text>`; `selection={value}`, `onSelectionChange` guarded `typeof === 'string'` then cast (see `SegmentedControl.ios.tsx`); brand checkmark via `tint(brandAccentColor(brandColors))`.
  - `PickerProps` (`node_modules/@expo/ui/build/swift-ui/Picker/index.d.ts`): `selection`, `onSelectionChange`, `children`, `modifiers`. **No per-option disable, no per-option subtitle** — `RadioOption.description` and `RadioOption.disabled` are **not expressible**.
- **Android** — Compose has no labelled radio row; `RadioButton` (`jetpack-compose/RadioButton`) is just the control: `{ selected, onClick, modifiers }` — **no `colors` prop** (brand tint not directly settable; reads M3 `primary`, which is acceptable). Assemble a `Column` of `Row`s: each `Row` carries the tap via a `selectable(...)` modifier with `role: 'radioButton'` (mirror the `toggleable` pattern in `SwitchRow.android.tsx`), a `RadioButton selected={…}` + label `Text` (+ optional description `Text` with `alpha(0.6)`). Android **can** render the description and disable a row (omit `selectable` + `alpha`), so the status-filter fidelity is only lost on iOS.

**Status-filter decision (required):** because iOS inline Picker drops `drafts` disabling + the "Sign in for drafts" hint, pick one:
- (a) Keep `description`/`disabled` in `RadioOptionProps` (API unchanged, honored on Android, ignored on iOS) and have `ClimbFilterSheet` handle the signed-out case at the call site — e.g. drop the `drafts` option when `!isAuthenticated` and render `t('mobile.filter.signInForDrafts')` as a footnote below the Picker; **or**
- (b) migrate only the **accuracy** RadioGroup and leave the **status** one as the current RN `RadioGroup` (rename the RN one or keep both). Recommended default: **(a)** — keeps one component, one API.

### Stepper
- **iOS** — `Stepper` from `@expo/ui/swift-ui` (`swift-ui/Stepper/index.d.ts`): `{ label: string; value?; step?; min?; max?; onValueChange }`. Native min/max auto-disables −/+ at bounds (matches current). **Caveat:** SwiftUI's Stepper renders only the label + the −/+ control — there is **no value slot**, so the current trailing numeric value won't appear unless folded into `label` (e.g. `` `${label}: ${value}` ``). That moves the number from a right-aligned cell into the (left) label — a small layout change. Decide whether that's acceptable; if the visible number must stay where it is, this is the main reason to weigh keeping Stepper RN.
- **Android** — no native Stepper. Reassemble: `Host` → `Row` (label `Text` + value `Text` + two `FilledTonalIconButton`s from `jetpack-compose/IconButton`, `minus`/`plus` `Icon` children, `enabled={!atBound}`, brand/M3 `colors`). Clamp + bound logic stays in a shared `Stepper.logic.ts` (lift `clampStepperValue`). Marginal gain over the current `PressableSurface` control — justify mainly by cross-platform consistency with the iOS native stepper.

Shared caveats: native text/label slots need a `<Text>` child, never a raw string; the iOS Host under-reports `matchContents` height (inline Picker grows with option count) — set a min/explicit height and watch clipping inside the scroll view's rounded card (`overflow: 'hidden'`), as `SegmentedControl.ios.tsx` documents.

## Approach

Follow the canonical platform-split convention in `docs/expo-ui-components.md`. For each component (`RadioGroup`, `Stepper`):

1. `*.types.ts` — lift the existing props type (keep the public API byte-identical so call sites/`index.ts` are unchanged).
2. `*.logic.ts` — pure node-testable helpers: RadioGroup's haptic+disabled select guard (copy `SegmentedControl.logic.ts`'s `makeSelectHandler`); Stepper's `clampStepperValue` + bound flags.
3. `*.ios.tsx` — the **only** file importing `@expo/ui/swift-ui*`. RadioGroup: copy the `Host`+`Picker`+`tag`+`tint`+string-guard shape from `src/components/SegmentedControl.ios.tsx` (swap `pickerStyle('segmented')` → `pickerStyle('inline')`). Stepper: `Host` + native `Stepper`.
4. `*.android.tsx` — the **only** file importing `@expo/ui/jetpack-compose*`. Copy the `Host`+`Row`/`Column`/`Text`+`selectable`/`alpha` shape from `src/components/SwitchRow.android.tsx`. RadioGroup: `RadioButton` group; Stepper: `IconButton` pair.
5. `*.d.ts` — ambient declaration so `tsc` resolves the extensionless import (RadioGroup is generic — mirror `SegmentedControl.d.ts`'s `export declare function …<T extends string>`).
6. Brand tint via `src/theme/expo-ui-modifiers.ts#brandAccentColor` (read `brandColors` from `useTheme()`). No new bridge needed for iOS; Android `RadioButton` has no colors prop, Android `IconButton` takes a plain `IconButtonColors` object (add a tiny pure mapper here if you want a branded tonal fill, like `segmentedBrandColors`).
7. Delete the old `RadioGroup.tsx` / `Stepper.tsx` once replaced; `index.ts:25` keeps `export { RadioGroup, type RadioOption } from './RadioGroup'` (now resolving to the split files).
8. **Tests:** add `test/radio-group-stub.tsx` and `test/stepper-stub.tsx` faithful passthroughs (plain RN `Pressable`/`Text`, preserve `radio`/`button` a11y roles), and register regex aliases in `packages/mobile/vite.config.ts` mirroring the existing `SwitchRow`/`SegmentedControl` entries. Update the existing mocks if needed: `climb-filter-sheet.test.tsx:263`, `generator-picker-card-analytics.test.tsx:179`, and `stepper.test.tsx` (its assertions hit clamp logic — re-point to `Stepper.logic.ts`).

## Out of scope / keep as React Native

- **Checkbox**: build nothing — no standalone checkbox UI exists. The `accessibilityRole="checkbox"` rows in `app/(tabs)/climbs/setters.tsx` (SetterRow) and `src/components/QueueItemRow.tsx` are **virtualized list rows** — never migrated.
- Any list/`FlashList`/`FlatList`, SVG board art, gifted-charts, gesture boards.
- Consolidating `ClimbFilterSheet` / `GeneratorPickerCard` into a single native SwiftUI `Form`/Compose `List` under one `Host` — that's a separate later pass; here each control keeps its own `Host` (one-per-card), matching `SwitchRow`/`SegmentedControl` today.
- The Material/Glass variant theming the current `Stepper` does — the native controls read the system/M3 theme; don't reimplement variant styling.

## Validation

- `vp run typecheck:mobile`
- `vp run test:mobile` (or `vp test run --project mobile`) — never `vp test --project mobile run`
- `vp run check:mobile-platform-imports` (`scripts/mobile-platform-imports-check.sh` — swift-ui only in `*.ios`, jetpack-compose only in `*.android`; `vp check` does **not** catch this)
- `vp check`
- Manual: filter sheet accuracy + status pickers select correctly and announce via VoiceOver/TalkBack; signed-out `drafts` still gated per the chosen status decision; generator steppers increment/clamp at bounds, press-and-hold repeats on iOS, value still legible. JS-only / OTA-deliverable (@expo/ui 56.0.18 already autolinked) — no native build.

## Dependencies

- **Foundation (merged, PR #3254 / #3261):** the split convention, `src/theme/expo-ui-modifiers.ts` bridge, the platform-imports guardrail, the stub+vite-alias pattern, Material XML drawables, and the `Picker`(segmented) reference in `SegmentedControl.*`. This issue builds directly on them — no other issue blocks it.
- **Independent of** the other migration areas (TextField/SecureField, native Form/List consolidation). If the filter sheet is later consolidated into a native Form, this RadioGroup folds into it then — coordinate ordering but no hard dependency.

## Checklist

- [ ] Decide the status-filter approach (option (a) call-site gating vs (b) keep status RN). Default (a).
- [ ] Decide Stepper value display (fold value into `label` vs keep current numeric cell vs keep Stepper RN). 
- [ ] `RadioGroup.types.ts` / `.logic.ts` / `.ios.tsx` (`Picker` + `pickerStyle('inline')`) / `.android.tsx` (`RadioButton` group) / `.d.ts` (generic).
- [ ] `Stepper.types.ts` / `.logic.ts` (lift `clampStepperValue`) / `.ios.tsx` (`Stepper`) / `.android.tsx` (`IconButton` pair) / `.d.ts`.
- [ ] Brand tint via `brandAccentColor`; add an Android `IconButton` colours mapper in `expo-ui-modifiers.ts` if branding the tonal fill.
- [ ] Min/explicit Host height on iOS to avoid `matchContents` clipping inside the sheet's rounded card.
- [ ] Delete old `RadioGroup.tsx` / `Stepper.tsx`; keep `index.ts:25` export.
- [ ] `test/radio-group-stub.tsx` + `test/stepper-stub.tsx` + `vite.config.ts` aliases; reconcile mocks in `climb-filter-sheet.test.tsx`, `generator-picker-card-analytics.test.tsx`, `stepper.test.tsx`.
- [ ] Full validation sequence above green; open draft PR and subscribe to feedback.

marcodejongh:
## Summary

Migrates the single-purpose **RadioGroup** selection control in `packages/mobile` to native **@expo/ui**, continuing the native-UI migration (Closes #3268, part of epic #3275). Follows the established platform-split convention (`docs/expo-ui-components.md`) used by `SegmentedControl` / `SwitchRow`.

- **RadioGroup** → SwiftUI inline `Picker` (iOS) / Jetpack Compose `RadioButton` group (Android). System-managed single-choice a11y instead of hand-rolled `radio` Pressables; brand-tinted selection. Generic `<T extends string>`, so the `.d.ts` declares a function (like `SegmentedControl`). Shared `makeRadioSelectHandler`, now memoized (`useMemo`) so a stable `onChange` doesn't push a new handler into the native Host each parent render.

### Rebased on `main`; Stepper dropped (superseded)

This branch previously also migrated the **Stepper** to a native `@expo/ui` Stepper. That's been **removed**: on-device QA found the native Stepper read wrong in the generator's custom dark card, so the generator Stepper shipped to `main` instead as a custom `[ − value + ]` quantity stepper via **#3291** (#3270). This PR is now **RadioGroup-only** and rebased on the latest `main` (which already contains that Stepper).

Ships **JS-only / OTA** — no native build.

Closes #3268

## Release Notes

- Cleaner single-choice filters in the climb-filter sheet — the status and accuracy pickers now use the native iOS/Android selection controls.

## Test plan

Run from `packages/mobile` (all green):
- `vp run typecheck:mobile`
- `vp run test:mobile` (2889 tests; incl. `radio-group`, `climb-filter`, and the inherited `stepper` suites)
- `vp run check:mobile-platform-imports`
- `vp run check:mobile-bundle` (iOS + Android Hermes bundles)
- `vp check` (format/lint)

**iOS needs on-device QA** (VoiceOver on the inline picker, the picker card under Dynamic Type) — it ships OTA, no build required to reach TestFlight (`pr-3280`).


marcodejongh:
Thanks — addressed all three:

**1. iOS RadioGroup height fragile under Dynamic Type** — switched from a hard `height` to `matchContents={{ vertical: true }}` + a `minHeight` floor (the content-sized `AngleSlider`/`SwitchRow` pattern). The Host now grows with the rows instead of clipping; the floor only guards against the native Host under-reporting at the default text size. (`RadioGroup.ios.tsx`)

**2. Missing `RadioGroup.logic` test** — added `__tests__/radio-group.test.ts` covering `makeRadioSelectHandler` (disabled-guard no-op, haptic fires, injected vs default haptic), mirroring the `Stepper.logic` coverage.

**3. One-frame orphaned selection** — good catch; you're right the effect fired after first render. Now normalized **at the source**: a pure `statusForAuth` runs inside the `useState` initializer and the parent-sync effect, so the first render already has a valid status. The auth-flip `useEffect` is now just the live-transition safety net (sign-out while the sheet is already mounted). Corrected the PR body's claim.

Bonus from digging into the OTA bot: the original Android stepper glyphs were vector drawables under `assets/`, which `assetBundlePatterns` matches and the fingerprint hashes — so they were turning this into a native-build-only change. Replaced them with M3 `Text` glyphs and dropped the drawables, restoring the issue's intended JS-only / OTA-deliverable property for both platforms.

marcodejongh:
Rebase on latest origin main

marcodejongh:
Correction to my note above on the OTA flag: the "native change detected" was **not** the Android drawables. I rebased on `main` and the fingerprints are byte-identical with and without those `assets/` files — so `assetBundlePatterns`-matched assets are not fingerprint sources here. The real cause was that the branch was behind a `main` that had reverted the iOS version in `app.config.ts` (a genuine fingerprint source); rebasing onto current `main` resolves it, and the change is genuinely JS-only / OTA-deliverable.

Because OTA is unaffected by the drawables, I kept the Android stepper on real Material `Icon` glyphs (proper TalkBack labels via `contentDescription`) rather than bare text glyphs — better a11y, which is the point of the migration.

marcodejongh:
On the latest review:

**Critical (XML drawables vs OTA)** — this is a false alarm from a stale read of the PR body: that review ran against the description *before* I finished correcting it, and its premise is empirically disproven. The `mobile-ota-check` job (which actually runs `@expo/fingerprint`) reports **✅ Compatible — fingerprints match `main`** (`b9e3e1b9fac3` iOS / `5a395625a0af` Android) **on this very commit, with `minus.xml`/`plus.xml` present**. So `assets/`-matched drawables do not move the fingerprint here; the change ships over OTA. The PR body no longer makes the assetBundlePatterns claim. (The earlier "native change" verdict was the branch trailing main's `app.config.ts` iOS-version revert — fixed by the rebase.)

**Minor — all addressed** (commit `ac7d480`):
- `docs/expo-ui-components.md` now lists `RadioGroup` (picker-style, generic) and `Stepper` (native iOS Stepper / Android IconButton pair) as canonical reference examples.
- `statusHint` opacity: added a comment noting it mirrors the adjacent `subsectionLabel` (the section's labels + hints share one muted opacity).
- safety-net effect: now passes the literal `false` to `statusForAuth` (the guard already proves `isAuthenticated` is false there).

marcodejongh:
On the latest (non-blocking) notes:

- **iOS drops `description`/`disabled`** and **`decreaseLabel`/`increaseLabel` unused on iOS** — surfaced both at the type level now with per-field JSDoc on `RadioOption` and `StepperProps` (commit `cf847f9`), so callers see the Android-only/iOS-ignored asymmetry at the call site.
- **Stepper test "regression"** — this matches the established split-component convention: `SwitchRow` and `SegmentedControl` likewise moved from JSDOM render tests to logic tests (`makeToggleHandler` / `makeSelectHandler`). The −/+ → onChange wiring and bound-disabling now live in the native platform files (not node-testable); the one platform-independent piece, `clampStepperValue`, is covered, and `test/stepper-stub.tsx` keeps indirect screen tests rendering. Added a matching `makeRadioSelectHandler` test for symmetry.
- **iOS minHeight under Dynamic Type** — the Host now uses `matchContents={{ vertical: true }}` so it grows with the rows; the `minHeight` floor is only an anti-collapse guard, not a hard cap. On-device Dynamic-Type verification is in the PR test plan.

Marking ready for review.

marcodejongh:
Latest round (commit `7a0187d`):

- **#2 duplicate `setLocalFilters` on auth flip** — reworded the effect comment to be accurate: the parent-sync effect *does* react to `isAuthenticated`, but early-returns once there are local draft edits, so the safety-net effect's real job is the sign-out-while-mid-edit case. When not editing, both fire idempotently (same value). No behavior change.
- **#3 `statusForAuth` untested** — extracted it to `../lib/climb-filter-types` (pure, no native imports) and added direct unit tests: signed-out `drafts`→`any`; signed-in `drafts` untouched; signed-out non-`drafts` untouched.
- **#1 Stepper test** — kept as a logic test on purpose: this matches how `SwitchRow`/`SegmentedControl` were migrated (render tests → `makeToggleHandler`/`makeSelectHandler` logic tests). The −/+ wiring + bound-disabling now live in the native platform files (not node-mountable); `clampStepperValue` and `makeRadioSelectHandler` are unit-tested, and the stubs keep indirect screen tests rendering.
- **#4 white `fillColor` in drawables** — matches the existing `assets/material-icons/*.xml` convention (white fill, tinted at runtime by the M3 content colour inside the Host's Compose Material theme); kept consistent rather than commenting one pair.

marcodejongh:
Latest round (commit `08dc673`):

- **#1 white-fill drawables / light-theme visibility** — real catch, thanks. `MoreForm.android` passes an explicit `tint`, so it doesn't actually prove default `LocalContentColor` inheritance works. Fixed the safe way: the −/+ Icons now tint explicitly to `m3.onSecondaryContainer` (the FilledTonalIconButton's content colour), so the white-filled drawables get a guaranteed-contrasting glyph in the light theme — no reliance on @expo/ui's Icon auto-inheriting content colour.
- **#2 bound-disabling coverage** — extracted `isAtMin` / `isAtMax` to `Stepper.logic`, used them in the Android impl **and** the stub, and unit-tested them. The `enabled={!atMin}` / `enabled={!atMax}` predicates now have direct coverage (the native `enabled` prop itself is the platform's, not node-testable, but the value driving it is verified).

Minor:
- `borderRadius: 10` in `RadioGroup.ios` is intentional — it's the iOS grouped-inset card radius (matches the old hand-rolled control). `borderRadius.md` is `8`, so the token would change the look; kept `10`.
- `makeRadioSelectHandler` without `useCallback` — agreed it's fine here (RadioGroup is one-per-card in a scroll view, not a virtualized list, per the perf checklist).

marcodejongh:
The two remaining automated notes are already handled / by-convention, so I'm holding here:

- **Stepper test coverage** — the render-based tests were intentionally replaced by logic tests (`clampStepperValue` + `isAtMin`/`isAtMax`), matching how `SwitchRow` and `SegmentedControl` were migrated. The `enabled` prop itself is the native control's (not node-mountable); the value driving it is now directly unit-tested, and the stub reuses the same helpers for indirect screen tests.
- **Unmemoized `makeRadioSelectHandler`** — deliberate. RadioGroup is used one-per-card in a scroll view (not a virtualized list), and the sibling primitives `SegmentedControl`/`SwitchRow` likewise call their handler factories inline. Adding `useMemo` here would make it inconsistent with them; the perf checklist only requires memoization for list rows.

### Summary for review
Ready for merge. CI green, OTA ✅ ships-over-the-air, no conflicts. All substantive review feedback addressed across the rounds: orphaned-selection normalized at source + unit-tested (`statusForAuth`), Dynamic-Type-resilient iOS picker height, explicit Android glyph tint (avoids white-on-white in light theme), platform-asymmetry documented at the type level, and `RadioGroup.logic`/`Stepper.logic` covered. **iOS needs on-device QA** (VoiceOver, inline-picker card under Dynamic Type, press-hold stepper) per the test plan — it ships OTA so no build is required to get it onto TestFlight.

Issue #3266, opened 2026-06-29
Mobile (@expo/ui): migrate auth + edit-profile text inputs to native TextField / SecureField (OutlinedTextField on Android)


marcodejongh:
> Part of the **@expo/ui native-UI migration** follow-up (after #3254 + #3261). Maintainer guidance: only migrate where it *legitimately improves the experience*; FlashList/FlatList lists, SVG board-art, charts, and gesture boards stay React Native. Pattern: `docs/expo-ui-components.md`.

> **Drafting-agent verdict:** `maybe` — Native keyboards/autofill/a11y are a real win and Android (drops react-native-paper, gets Material 3 isError/SupportingText) is a clear gain, but iOS SecureField has no reveal toggle and no built-in error text, the current iOS white-HIG field is a deliberate design, and login is the most critical flow — so it's the highest-risk item in the set. Do it carefully, Android-first, lower priority than the safer migrations.

---

## Why this improves the experience

Replacing the re-created `AuthTextInput` with real native fields — SwiftUI `TextField` / `SecureField` on iOS, Material 3 `OutlinedTextField` on Android — buys the platform's own keyboard, caret, selection handles, text-replacement, autofill / Strong Password offer, and VoiceOver/TalkBack field behaviour for free, the same payoff as the `SwitchRow` / `Picker` migrations. On Android it also drops `react-native-paper` for these fields and gives a real Material 3 outlined field with native `isError` + supporting-text.

**Be honest: this is the highest-risk item in the set, on the single most critical flow (login).** Two things make it a `maybe`, not a slam-dunk:

1. **iOS `SecureField` has no reveal (show/hide) affordance and no way to flip to a visible field without remounting.** Today's deliberate eye-toggle would either be dropped on iOS or forced into a SecureField↔TextField swap that drops focus and breaks the Keychain/Strong-Password association mid-edit. Android keeps the toggle cleanly (`visualTransformation` is a display prop on the same field).
2. **The current iOS field is a deliberate custom design** — the white-in-dark-mode HIG input (`AuthTextInput.tsx` glass branch) that matches the rest of the auth screens. A native SwiftUI `roundedBorder` field follows system colours and will look different. The maintainer's rule is "don't lose a deliberate custom design lightly."

Net: do it carefully, **Android-first is the clearer win**, treat iOS as a deliberate decision (accept losing the reveal toggle, or accept a focus-dropping swap), and verify login on real devices. Lower priority than the safer migrations in this set.

## Current state

[...]

PR #3284, opened 2026-06-29
feat(mobile): migrate auth + edit-profile text inputs to native @expo/ui fields

marcodejongh:
## Summary

Re-implements the shared `AuthTextInput` as a platform-split native component (the `SwitchRow.*` convention) — SwiftUI `TextField`/`SecureField` on iOS, Material 3 `OutlinedTextField` on Android — while keeping the **public API identical**. All five call sites (login, register, forgot/reset password, edit-profile) are unchanged except their focus-chain ref type (`RNTextInput` → the new `AuthTextInputHandle`).

Part of the @expo/ui native-UI epic (#3266 / #3275). Native fields give the platform keyboard, caret, selection, autofill, and VoiceOver/TalkBack behaviour for free. Android additionally drops `react-native-paper` for these fields (native `isError` + supporting text).

Refs #3266.

### What's in the box
- `src/components/AuthTextInput.{types,logic,ios,android,d}.tsx` — the platform-split set; pure mappers in `.logic.ts` (node-tested).
- `assets/material-icons/visibility{,Off}.xml` — eye icons for the Android reveal toggle (bundled XML vectors).
- `test/auth-text-input-stub.tsx` + vite alias; `AuthTextInput.logic.test.ts`.
- `textFieldBrandColors()` bridge helper in `expo-ui-modifiers.ts`.

### Deliberate decisions (confirmed with maintainer)
- **iOS drops the password reveal toggle** — SwiftUI `SecureField` has none, and swapping `TextField`↔`SecureField` would remount/drop focus + the Strong-Password association. Android **keeps** its toggle (`visualTransformation`, no remount).
- **iOS adopts the system `roundedBorder` look** instead of the old custom white-in-dark field.
- **Autofill is wired on both platforms**: iOS via the `textContentType` modifier (Strong Password), Android via `semantics({ contentType })` (the existing `autoComplete` tokens map straight to Compose `ContentType`).
- **`editable={false}`** → Android `readOnly` (legible/selectable, e.g. the read-only email), iOS `disabled` (greyed).
- iOS error is shown via a live-region sibling `<Text>` (SwiftUI fields have no `isError`); Android uses native `isError` + `SupportingText`.
- `passwordRules` has no @expo/ui equivalent and is now a documented no-op (iOS Strong Password defaults already satisfy the 8–128 rule).

### Known axis change
Branching moves from `theme.variant` (user-selectable) to `Platform.OS` (file extension) — same change `SwitchRow`/`SegmentedControl` already made. Today no user-facing setting flips the variant (`setUiVariant` is dev/test-only), so this only affects internal testers on an explicit `material`/`liquidGlass` override.

## Release Notes

- Sign-in and profile fields now use your phone's own keyboard, password autofill, and (on iPhone) Strong Password.

## Test plan

Validated on Linux:
- `vp run typecheck:mobile` ✅
- `vp run test:mobile` ✅ (2822 tests; incl. new `AuthTextInput.logic` suite; `EditProfileScreen.test` still green via its own mock)
- `vp run check:mobile-platform-imports` ✅ (swift-ui only in `.ios`, jetpack-compose only in `.android`)
- `vp run check:mobile-bundle` ✅ (both iOS + Android bundles export; new XML vectors bundled)
- `vp check` ✅

### ⚠️ On-device verification still required before un-drafting (login is the critical flow)
- [ ] **iOS go/no-go:** Strong-Password / autofill on login + register makes the submit button enable and actually submits (the bridge mirrors native edits via `onTextChange`; if autofill doesn't fire it, `canSubmit` would stay false). Needs a macOS device/sim.
- [ ] **iOS:** return-key chaining email→password→…→submit doesn't bounce the keyboard (cross-Host async `focus()`); fields aren't blank (placeholder falls back to label); inline error renders/clears; edit-profile email non-editable but legible.
- [ ] **Android:** reveal toggle masks/unmasks; `isError`/supporting text render; password-manager autofill recognises email/password; chaining works; edit-profile email selectable, not greyed.


marcodejongh:
## Addressed the review findings + Android on-device QA

**Fixed (pushed):**
- `handleTextChange` → `useCallback`, Android `keyboardActions` → `useMemo` — stable refs into the native `Host` (perf rules).
- iOS `disabledModifier` now conditionally spread like the rest of the array (no no-op modifier on editable fields).
- `toAndroidContentType` — dropped the `?? undefined` no-op; kept as a documented, tested pass-through seam (the token→ContentType guarantee lives in `@expo/ui`'s Kotlin `toContentType()`).
- Added comments: `useNativeState` returns a stable ref (so the seed `useEffect` only runs on `value` change); `toAndroidCapitalization` defaults to `'none'` (not RN's `'sentences'`) deliberately for auth fields.

**Already handled / by-design (no change):**
- `accessibilityLabel` on Android — that finding was on the pre-amend commit; the type doc now states it's iOS-only (the `@expo/ui` Android `semantics` modifier exposes only `contentType`, not `contentDescription`; the visible Material `Label` is the TalkBack name). No call site passes a custom one.
- `editable={false}` asymmetry (iOS `disabled`/greyed vs Android `readOnly`) — deliberate and documented; on the device-QA checklist.
- `toAndroidContentType` *is* unit-tested (`AuthTextInput.logic.test.ts`).
- The variant→`Platform.OS` axis change is the established `@expo/ui` split pattern (`docs/expo-ui-components.md`), same as `SwitchRow`/`SegmentedControl`.

**Android on-device QA (emulator, this branch's JS over the cached dev-client):** the native Material 3 `OutlinedTextField`s render correctly on the Edit Profile screen — floating labels, the read-only email **seeded via the controlled-value bridge** (`test@boardsesh.com`, legible, not blank/greyed-illegible), and the supporting-text slot. No "Unable to get view config" crash. (Login/register password fields couldn't be captured — the screenshot session is logged-in and auth routes redirect away — so the masking + reveal-toggle path remains verified against the `@expo/ui` Kotlin source.)

**Staying in draft:** the iOS Strong-Password→`canSubmit` go/no-go and the cross-Host focus-chain check need a macOS device/sim (see the checklist in the description). Everything I can validate on Linux is green.

marcodejongh:
## Fix: iOS Password AutoFill was filling only the password, not the email

**Root cause:** iOS pairs a username/email field with a password field for "fill both" **only within one view hierarchy**. The per-field `AuthTextInput` renders each field in its own `@expo/ui` `Host` (separate `UIHostingController`), which unpaired them — so the password manager filled the password but no longer the email. (`textContentType`s were unchanged; it's purely the per-field-Host split.)

**Fix — `AuthFieldset`** (`src/components/AuthFieldset.*`), a credential-form group:
- **iOS:** renders every field of a screen in **one `Host`** as a SwiftUI `VStack` of `TextField`/`SecureField`, the way a hand-written SwiftUI login form pairs fields for AutoFill. Each field is a child component (legal per-field hooks); error/hint are SwiftUI `Text` inside the VStack; return-key focus chaining is internal (derived from field order).
- **Android:** re-renders the existing, emulator-verified per-field `AuthTextInput` with callback-ref chaining — Android Autofill is activity-level, so no Host grouping is needed and the Android field path is unchanged.
- `login` / `register` / `reset-password` now use `AuthFieldset`; `edit-profile` keeps `AuthTextInput` (not a credential form).

Green on Linux: typecheck, 2822 tests, platform-imports guard, **both** iOS + Android bundles export, lint.

### ⚠️ Updated on-device gate (still draft)
The single-Host pairing is the exact thing I can't verify on Linux. On an iOS device, please confirm:
- [ ] Saved-credential AutoFill now fills **both** email **and** password on login.
- [ ] Strong Password offer + save still works on register (fields share one Host now).
- [ ] Return-key chaining email→password→…→submit doesn't bounce the keyboard (focus now moves intra-Host).
- [ ] Fields render correctly stacked (no clipping; error/hint show in red/secondary).

If single-Host still doesn't pair on device, the fallback is to revert iOS to the RN field and keep Android native.

marcodejongh:
## ✅ On-device verified — iOS + Android

Marco confirmed on real devices:
- [x] iOS Password AutoFill now fills **both** email + password on login (the regression is fixed by the single-Host `AuthFieldset`).
- [x] Android works (per-field `AuthTextInput` path, unchanged).

That clears the only remaining gate. CI is fully green (incl. native `prebuild-and-build` + `claude-review`), no conflicts (`MERGEABLE`), review findings addressed. **Marking ready for review.**

Issue #3235, opened 2026-06-27
iOS 26.5 BLE write failures on shared box 13238 (Pete & Nic) — box & build proven healthy; two distinct causes


marcodejongh:
## Summary

Two iOS users (a climbing partnership, same Kilter wall) have chronic BLE write failures where **the controller box stops accepting writes** → the write times out / the link drops → reconnect → repeat. This is **distinct from the write-type regression** fixed in #3228: the evidence shows the failure also happens on the *correct* write-without-response path, which points at the **controller box / BLE link (and possibly its power) being marginal**, not (only) the app.

This issue tracks confirming and, where app-addressable, mitigating the residual board/link reliability problem. Filed off the investigation in #3228.

## Evidence (PostHog + Sentry)

Same underlying message in PostHog: **"BLE write timed out waiting for the board to accept data"** — board stops ack'ing/accepting writes → timeout → `Bluetooth Disconnected (reason = unexpected)` → auto-reconnect → repeat.

| | Pete — iOS **2.0.1** | Nicola — iOS **2.0.0** |
|---|---|---|
| Write path | write-**with**-response (the `00bda53a2` regression, via a stale iOS GATT cache that dropped the `.writeWithoutResponse` bit) | write-**without-response** — the *correct* path, unconditionally |
| `failureReason` | `write_timeout` / `write_recovery_failed` | `write_failed` |
| Crashes | **0** across 29 failures (#3181 catches + recovers) | **uncaught exceptions → ~5 crashes** in one short session (2.0.0 predates #3181) |
| Fixed by #3228? | **Yes** (removes the with-response stall) | **No** — she was never on the broken path |

Confirmed `2.0.0`-era native code (`5f8d0d632`) has neither `00bda53a2` nor #3181: it always wrote `.withoutResponse` with no stall watchdog. So **Nicola failed on the exact write path #3228 restores** — that's the key signal this is not purely an app write-type bug.

Corroborating:
- Both users are partners on the **same physical board** → a single suspect controller box.
- The iOS write-stall signature is **isolated to this partnership** — no other iOS user shows `write_timeout` over a 90-day window; the broad multi-user BLE issue (`BOARDSESH-7J`, 10 users) is a *separate* Android disconnect on production `2.0.0+2000392`.
- Pete power-cycling his phone restored it on the still-regressed build (iOS GATT cache effect) — i.e. write-type/cache was *his* dominant factor, but Nicola's data shows a board/link layer underneath.

[...]

PR #3289, opened 2026-06-29
feat(mobile): emit CoreBluetooth disconnect reason from native BLE manager

marcodejongh:
## Summary

Native counterpart to #3288 (which ships the JS reader on OTA). `BoardBleManager` dropped the `error: Error?` from `didDisconnectPeripheral`, so the `disconnected` event carried only a `deviceId` and JS reported `disconnectReason: null`.

This threads the reason through `onDisconnect((String, [String: Any]?))`:

- `didDisconnectPeripheral` attaches the CoreBluetooth `NSError` (`code` / `domain` / `localizedDescription`). The CBError code is what distinguishes a peer-terminated link (another central grabbing the last-connection-wins board) from a range/idle timeout.
- the three write-stall recovery paths attach a `context` marker instead, so an app-driven drop isn't mistaken for a link loss.
- `BoardBleModule` merges the reason dict into the emitted event body.

> [!WARNING]
> **Native change — do not merge to `main` until a native build is being cut.** It bumps the Expo fingerprint, so a production OTA published from `main` with this change would no longer match any shipped store binary. Hold this PR until the next build cycle. The JS PR (#3288) is fingerprint-neutral and degrades to a `source`-only reason until this lands in a build, so it can merge first.

## Release Notes

## Test plan

- JS reader + adapter mapping covered by #3288's test suite. Swift compiles with the next native iOS build; no JS test exercises the Swift path directly.
- Manual native verification deferred to the build that ships this (capture a real Kilter takeover vs walk-away and confirm the CBError codes differ in PostHog).

### DIFF
--- a/packages/mobile/ios-tests/LiveActivityWidgetTests.swift
+++ b/packages/mobile/ios-tests/LiveActivityWidgetTests.swift
@@ -190,6 +190,33 @@ final class LiveActivityWidgetTests: XCTestCase {
         XCTAssertTrue(url?.absoluteString.contains("include_background=1") ?? false)
     }
 
+    // The disconnect-reason dict feeds the JS `disconnected` event
+    // (NativeBleDisconnectEvent) and, through it, the `Bluetooth Disconnected`
+    // analytics event — key names and the nil-for-clean contract are load-bearing.
+    func testDisconnectReasonBodyIsNilForCleanDisconnect() {
+        XCTAssertNil(BoardBleEncoding.disconnectReasonBody(from: nil))
+    }
+
+    func testDisconnectReasonBodyCarriesCBErrorFields() {
+        let timeout = BoardBleEncoding.disconnectReasonBody(from: CBError(.connectionTimeout))
+        XCTAssertEqual(timeout?["errorCode"] as? Int, CBError.connectionTimeout.rawValue)
+        XCTAssertEqual(timeout?["errorDomain"] as? String, CBErrorDomain)
+        XCTAssertFalse((timeout?["errorDescription"] as? String ?? "").isEmpty)
+        // The takeover-vs-idle distinction rides on the code: peer-terminated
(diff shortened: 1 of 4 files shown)


marcodejongh:
Addressed the review in e43ff97bd (and rebased onto current main so the #3288 JS reader is in-branch):

1. **Merge resolver flipped** — the reason dict can no longer overwrite `deviceId` (`{ current, _ in current }`).
2. **Swift tests added** — `disconnectReasonBody` moved into `BoardBleEncoding` (the test-compiled target); `LiveActivityWidgetTests` now covers the nil-for-clean contract, the CBError field mapping, arbitrary-NSError wrapping, and that `connectionTimeout` vs `peripheralDisconnected` stay distinguishable (the takeover-vs-idle signal).
3. **New: explicit `bluetooth_unavailable` marker** — when the central drops below `.poweredOn` with a live peripheral, iOS invalidates the link *without* delivering `didDisconnectPeripheral`, so JS kept showing "connected" until the next write failed. That path now tears down and emits `context: "bluetooth_unavailable"`, so analytics can't misread bluetooth-off as a link loss.

On the reason-shape inconsistency (finding 2): left as-is deliberately — the merged JS reader (#3288) reads every field optionally and all shapes flow through; forcing error fields onto the context-only paths would fabricate data.

Validation: full test-target Swift file set typechecks clean against the iphonesimulator SDK; `vp run typecheck:mobile`, `vp run test:mobile` (3,086 pass, incl. #3288's disconnect-reason suites now in-branch), `vp run check:mobile-bundle` all green.

Still **hold for the next native build cycle** (fingerprint) — ideally riding the same build as #3365. Adversarial telemetry re-run on #3235 (2026-07-03 comment) confirms this PR is still needed: `disconnectReason` is null on 100% of events and only ~14% of unexpected disconnects are attributable to takeover without CBError codes.

PR #3306, opened 2026-06-29
feat: Woods Board support — BLE control, board rendering, and the full climb catalog

### LINKED ISSUE #4692: 🐞 Bug report: Any chance you could link up to the Woods board?

## Comment

Any chance you could link up to the Woods board?

## Metadata

| Field | Value |
| --- | --- |
| Platform | ios |
| App version | 2.3.1 (2) |
| Source | drawer-bug |
| Board | unknown |
| Climb | unknown |
| Session | unknown |
| URL | /profile |
| User agent | unknown |
| Feedback row | app_feedback #10207 |

## Contact

✅ Reporter agreed to follow-up — resolve **app_feedback #10207** in the DB to reach them.

marcodejongh:
## Summary

Adds **Woods Board** support end to end — the reverse-engineered BLE protocol, the two board sizes (12×12 and 8×10) in the mobile board builder, board rendering on mobile and the web climb pages, the full scraped catalog importer, and acknowledged-write Bluetooth control. Closes #4692.

This is the June branch (Fable BLE-reviewed 2026-08-08) rebased from 691 commits behind `main`, with the web Bluetooth/create-climb slices that W-16 deleted dropped, and the deferred work finished. Woods ships **enabled** — there is no feature flag.

**What's in**

- **BLE** — `@boardsesh/ble-protocol/woods` (`getWoodsBluetoothPacket`: plain-ASCII `ledIndex,roleCode,…,!` per spec §5; bare `,!` = clear), name-based discovery on the Nordic UART family, and the mobile send path. Woods writes are **acknowledged** (spec §8): `RNBleAdapter` gains a `preferWriteWithResponse` option and Woods stays on the ble-plx adapter on iOS because the native Swift writer hard-codes write-without-response for non-MoonBoard boards (#3314). Live Activity Previous/Next are hidden for Woods for the same reason.
- **Board model** — code-driven like MoonBoard: one layout ("Original"), two product sizes (8×10 = 1, 12×12 = 2, distinct products so they never rank as nested), one synthetic hold set ("Standard", id 1) so board paths, the board builder and readable URLs all work. Grades map onto the shared difficulty scale (V0→`4a/V0` … V16→`8c+/V16`; the 6 V17 climbs clamp to V16).
- **Rendering** — CV-detected per-hold positions over the board art (`.png` + `.webp` + thumbnails; bundled into the mobile app via the backgrounds manifest, rides OTA), mobile `getWoodsRenderData`, web `WoodsBoardRenderer`, and the OG share card. Hold radius tuned to the measured hold spacing; structural invariants on the position table are pinned by tests.
- **Mobile** — Woods in the board builder (size + angle 20–70°), climb browsing/search/offline, grade rail (17 stops), queue, ticks, size-aware compatibility (an 8×10 climb can't be queued onto a 12×12 board — their hold ids overlap numerically but are different holds), render-board resolution for logbook/playlist tiles. Create-climb is hidden for Woods (follow-up).
- **Backend/web** — `assertKnownBoardConfig` accepts Woods configs; `/og/climb` + `/api/internal/board-render` accept `woods`; `/woods/…` routes parse and validate (bogus size/layout/set → 404); labels; `populateDenormalizedColumns` leaves the importer's arrays alone; weekly stats-history catch-up includes Woods.
- **Importer** — `vp run db:import-woods-catalog -- <catalog dir>`: idempotent upserts (deterministic uuidv5 keyed on size|name|author|frames), ISO dates, upstream ascent counts preserved across ticks and re-imports, self-alias rows, 17 grade rows, a remote-DB guard (`WOODS_IMPORT_ALLOW_REMOTE=1`). Dry-run over the catalog: 5,392 climbs (424 8×10 / 4,968 12×12), 60,686 holds, 2 empty problems skipped, 24 app double-submits folded.

**Deliberately not in (follow-ups filed)**: native iOS Live Activity control (#3314), hold-zone search filters (#4748), sitemap + popular-configs rows (#4749), create-climb for Woods (#4750), "nearby boards" discovery + wall-finder chips (#4751), on-wall hardware verification with the #4692 reporter (#4752), a dark-mode variant of the white-ground art (#4753), mirrored climbs end to end (#4754), deriving the last hand-written board allowlists (#4755). Two pre-existing web gaps that Woods exposed: unknown numeric size ids 500 instead of 404 (#4756) and a climb viewed under an incompatible size URL drawing on the wrong art (#4757).

Notes for reviewers: `searchClimbs` ignores the browsed angle for every board (Kilter returns the same 220k at 20°/40°/70°), so a Woods list shows every climb of that size with its own set-angle grade — parity, not a change (a per-board angle filter would be a product decision). Logging a Woods tick on the emulator recomputed stats to 19 upstream + 1 Boardsesh = 20, so imported repeat counts survive ticks and re-imports. `board_attempts` rows are not needed (nothing reads them); projects are listed (a Woods "project" is an unsent problem, not a hidden one); `vp run test:db` has 3 pre-existing `create-climb-filters.test.ts` failures on `main` unrelated to this PR.

## Release Notes

- Woods Board owners can add their 12×12 or 8×10 board and browse 5,400+ climbs
- Light up any Woods climb over Bluetooth straight from the play view
- Grades, setters and angles come straight from the Woods app, so the problem you are projecting is the one you will find here

## Test plan

- `vp check`, `vp run typecheck` (all packages), `vp run test:mobile`, web/board-config/board-constants/ble-protocol/play-view/board-render/backend/i18n vitest projects, `vp run check:mobile-bundle`, `vp run check:i18n`, `vp run check:i18n:orphans`, `vp run check:large-files` — green.
- Catalog imported into the local dev DB and verified by SQL (counts, ISO dates, size split, grade ids, idempotent re-run); web list/view/OG and GraphQL search/grades/angles exercised; Android emulator pass over the mobile flows (board builder → list → climb → grade rail → queue, both sizes, dark mode, offline).
- Fable BLE review: encoder unchanged vs the 2026-08-08 review except the removed library `console.warn`; write-type gate mutation-tested; Aurora/MoonBoard paths byte-identical.
- OTA-compat check must report JS-only. No DB migration.

https://claude.ai/code/session_01ERtGeFnGqGb4QKmea4zxK3


Stevie-Ray:
dude sick!

marcodejongh:
## Fable review — findings and fixes

Reviewed the Woods board work as the mandatory BLE review. The core is strong: the shared encoder reproduces the spec's worked example byte-for-byte, the generated LED maps match the source JSONs exactly, geometry round-trips are exhaustively tested, and the importer is idempotent Drizzle with pinned (non-tautological) tests. Six findings, all now addressed.

**Stale branch (major).** The branch was cut 2026-06-29 and conflicted with main in 7 files, including both platforms' BLE hooks — main has since reworked them (the "one board per config" repair) and moved `getBoardDetailsForBoard` into `@boardsesh/board-render`. Rebased onto main and re-integrated the Woods dispatch branches against the current hook shape; the Woods board-details routing now lives in `board-render` next to Aurora/MoonBoard, so the backend OG renderer picks it up too. `vp run typecheck` (all packages), `vp test` (web/mobile/shared/db), and `vp run check:mobile-bundle` are green on the rebase. Main has grown `soill` and `touchstone` since June and every `Record<BoardName>` site still resolves.

**Web BLE was never wired (major).** `WOODS_REQUEST_DEVICE_OPTIONS` / `WOODS_SCAN_SERVICE_UUIDS` / `WOODS_OPTIONAL_SERVICE_UUIDS` were exported and imported nowhere. All three adapter paths route a non-MoonBoard board to Aurora — `web-adapter.ts` picks the Aurora advertised-service filter (a Woods board advertising NUS never appears in the chooser), `capacitor-adapter.ts` picks the Aurora scan UUIDs, and `adapter-factory.ts` sends woods to the native-iOS adapter. So the web send branch was unreachable. Given web climbing surfaces are moving to the Expo build, wiring it up would be investment in a path with a known end of life, so the slice is removed instead: `bluetooth-woods.ts` and the dead send branch are gone, and the hook now refuses a Woods board explicitly rather than letting it fall through to the Aurora packet builder. Mobile's `scanFamilyForBoard('woods') → 'moonboard'` wiring was already correct and is now the only Woods surface. PR body corrected — it claimed web + mobile wiring.

**Web write-without-response (major)** and the **hardcoded English strings** and **missing partial-skip report** (both minor) were all in that web branch, so they go with it. The surviving mobile path already satisfies spec §8: its `moonboard` scan family (which Woods joins) gates on `characteristic.isWritableWithoutResponse` and falls back to an acknowledged write.

**Mobile parity, while I was in there.** The Woods branch was missing three things the MoonBoard branch has: a user-initiated clear tracked `Climb Sent to Board Success` instead of `Board Lights Cleared` (Woods encodes clear as the bare `,!`, so it rides the same branch — the `sendSource` gate is the only separator), success carried no BLE write diagnostics, and a partial skip left only a `console.warn` inside the encoder. All three fixed, and five tests pin the branch against the real encoder and the real 8×10 LED map: encode, clear, partial skip, all-skipped refusal, unknown size id.

**Importer `required_set_ids` (minor).** Every climb was written `[1]` while the Woods config exposes no hold sets at all — `getWoodsBoardDetails` returns `set_ids: []` and the route parser yields `[]`. Both containment filters (`create-climb-filters`, `playlist-climbs`) are skipped on an empty selected-set list, so it was inert today, but `[1]` fails containment against the board's own set list the moment one runs. Woods has one fixed layout with nothing to install, so the honest value is the empty array: `{} <@ anything` is true, so a Woods climb can never be filtered out by a set list, where `NULL <@ …` would drop it. Now `WOODS_REQUIRED_SET_IDS`, pinned by a test that reads the value back off the board config.

The board stays gated behind `WOODS_ENABLED=false` throughout, so none of this is reachable in production yet. On-wall hardware verification is still outstanding — the encoder is verified against the spec, not against a real board.

marcodejongh:
### Status: everything green except a break inherited from `main`

All six review findings are addressed and pushed. 25 checks pass — typecheck, lint, i18n, mobile-bundle, backend, ocr, location-sync, codegen-drift, commit-lint, release-notes, claude-review.

`test-default` is red on one test, and it is not this PR's:

```
FAIL  i18n  src/__tests__/catalog-completeness.test.ts > de > boards.json ships the same key set as en-US
Test Files  1 failed | 1298 passed
     Tests  1 failed | 14562 passed
```

`de/boards.json` is missing 30 keys and carries 2 stale ones against `en-US`. This branch touches no i18n catalogs, and `main` itself is red on the same job (run 31222889396 at `681a0481d`). **#4213 already fixes exactly this** (36 additions / 4 deletions, that one file) and is green and mergeable — it just needs an approving review.

So: approve and merge #4213, and this PR goes green on the next run. Leaving it in draft until then rather than marking it ready over a red build.

Two follow-ups from the review bot went in on the last push while I was here: the BLE encoder was restating the lit role codes as `new Set([1, 2, 3, 4])` in a file that already imports `WOODS_WIRE_ROLE`, so a role added to the constants would have stayed dark on the wall — it derives from the constant now. And the importer keyed a climb's deterministic UUID off the raw catalog author while storing the trimmed one as `setter_username`, so `" ada"` and `"ada"` would have minted two climbs for one problem; both read the trimmed value now, pinned by a test.

The bot's third point — `images_to_holds` being empty and rendering `/images/woods/` as a 404 — doesn't hold: `getWoodsBoardDetails` always writes exactly one entry (`{ [geometry.backgroundImage]: [] }`), so the key is never absent.

packages/backend/src/graphql/resolvers/climbs/queries.ts
```diff
@@ -199,6 +199,11 @@ export const climbQueries = {
     // MoonBoard data changes frequently via local creation/import flows, so keep
     // GraphQL search results uncached there. Other boards can still use Redis
     // when the query is anonymous and has no user-specific filters.
+    //
+    // Woods stays cacheable despite also being code-driven: its catalog is a
+    // static import with no in-app creation or import flow writing to it, so a
```
marcodejongh, line 204: Is there a followup issue for getting climb creation working?

packages/board-constants/src/hole-placements.ts
```diff
@@ -123,6 +123,7 @@ export function getBoardHolePlacements(boardName: BoardName): BoardHolePlacement
 const ALL_BOARD_NAMES: BoardName[] = [
   ...(['kilter', 'tension', 'decoy', 'touchstone', 'grasshopper', 'soill'] as const),
```
marcodejongh: This is weird? why cant it just be a regular array? Why do we spread the aurora names in?

packages/board-constants/src/stable-json.ts
```diff
@@ -1,5 +1,5 @@
 const BOARD_SORT_ORDER = new Map(
-  ['kilter', 'tension', 'decoy', 'touchstone', 'grasshopper', 'soill', 'moonboard'].map((boardName, index) => [
+  ['kilter', 'tension', 'decoy', 'touchstone', 'grasshopper', 'soill', 'moonboard', 'woods'].map((boardName, index) => [
```
marcodejongh: This could be a shared constnat

packages/db/scripts/woods-catalog-helpers.ts
```diff
@@ -0,0 +1,329 @@
...
+// does run).
+// Frozen and typed `readonly` so the single exported instance can't be handed to
+// thousands of climb rows as a shared mutable reference — callers copy it into
+// each row (`[...WOODS_REQUIRED_SET_IDS]`).
+export const WOODS_REQUIRED_SET_IDS: readonly number[] = Object.freeze([]);
+
+// boardDimension → compatible product-size ids. 8x10 is product size 1, 12x12 is
+// product size 2 (matching the WOODS_BOARD_SIZES ordering in board-constants).
+export const WOODS_DIMENSION_TO_SIZE_IDS: Readonly<Record<string, readonly number[]>> = Object.freeze({
+  '8x10': Object.freeze([1]),
```
marcodejongh: Why do we need an object freeze here?

packages/mobile/app/(tabs)/climbs/create.tsx
```diff
@@ -29,12 +39,50 @@ type CreateClimbParams = {
  * `BoardName` and indexed into `STATE_TO_PRIMARY_CODE`, which throws during
  * render on the remix/edit path (#3804). Treating an unsupported value as absent
  * makes it fall back to the active board, exactly like a missing param.
+ *
+ * A board that cannot have climbs set on it (Woods) is NOT handled here — it
```
marcodejongh: Needs an GH Issue

packages/mobile/app/(tabs)/climbs/holds.tsx
```diff
@@ -60,6 +62,12 @@ export default function HoldFilterScreen() {
   // join cleanly with web across platforms. Falls back to '' for unknown ids.
   const boardLayout = getLayout(boardName, layoutId)?.name ?? '';
 
+  // The filter sheet hides the Holds row on a board that can't answer a hold
```
marcodejongh, line 65: Why cant it answer a hold search? Needs an issue

packages/mobile/src/components/climb-actions/use-climb-actions.ts
```diff
@@ -105,6 +106,10 @@ type UseClimbActionsArgs = {
 
 // Mirrors web's constructClimbInfoUrl: Kilter no longer has a public app URL.
 function buildAuroraAppUrl(boardName: string, climbUuid: string): string | null {
+  // Only the Aurora boards have a `<board>boardapp.com` site. Without this gate a
+  // code-driven board (MoonBoard, Woods) gets an invented domain and the row opens
+  // a browser on an error page.
+  if (!(AURORA_BOARDS as readonly string[]).includes(boardName)) return null;
```
marcodejongh: Bit overly defensive this function probably shouldnt be called with an non-aurora board name

packages/mobile/src/components/play-drawer/BoardseshGradeSection.tsx
```diff
@@ -29,15 +29,15 @@ export const BoardseshGradeSection = memo(function BoardseshGradeSection({
   const { gradeFormat } = useGradeFormat();
   const { brandColors } = useTheme();
 
-  // MoonBoard has no community grade data in our feed yet, so skip the fetch.
-  const moonboard = isMoonBoard(boardName);
+  // MoonBoard and Woods have no community grade data in our feed, so skip the fetch.
+  const noCrowdGrade = lacksCrowdGrade(boardName);
```
marcodejongh: This is becoming a bit of a mess, we should standardise into a getBoardConfig or similar helper, that can turn things like crowd grade, search, etc off

packages/mobile/src/lib/ble/use-board-bluetooth.ts
```diff
@@ -833,6 +846,73 @@ export function useBoardBluetooth({
             return true;
           }
 
+          if (boardName === 'woods') {
```
marcodejongh, line 907: Messy to have such a big if here, this function should be split up more to improve maintainability

packages/backend/src/graphql/resolvers/climbs/queries.ts
marcodejongh: Yes — #4750 tracks Woods create-climb; the comment now names it (fa26d819d).

packages/board-constants/src/hole-placements.ts
marcodejongh: Gone — ALL_BOARD_NAMES now reads SUPPORTED_BOARDS from @boardsesh/shared-schema directly (membership), with enumeration order from the shared BOARD_DISPLAY_ORDER; no local spread. Same set and order as before (the ownKeys trap order is observable — pinned by hole-placements-lazy.test.ts).

packages/board-constants/src/stable-json.ts
marcodejongh: Now shared: BOARD_DISPLAY_ORDER lives in @boardsesh/shared-schema, pinned as a permutation of SUPPORTED_BOARDS by a new test; stable-json.ts derives from it and the iOS placement-data byte-stability test still passes.

packages/db/scripts/woods-catalog-helpers.ts
marcodejongh: It was belt-and-braces on top of the readonly typing — dropped. The per-row array copies stay (each climb row must own the array it hands the driver), with a test pinning that reason.

packages/mobile/app/(tabs)/climbs/create.tsx
marcodejongh: Ref added in the route comment: #4750 (Woods create-climb). The route now bails out cleanly for boards without creation instead of spinning.

packages/mobile/app/(tabs)/climbs/holds.tsx
marcodejongh: Because online hold/zone search resolves picked hold ids against board_placements, and Woods has no placement rows yet — #4748 tracks seeding them (backfill-moonboard-hardware.ts style). The comment now says exactly that with the issue ref.

packages/mobile/src/components/climb-actions/use-climb-actions.ts
marcodejongh: Agreed — buildAuroraAppUrl now takes AuroraBoardName and assumes Aurora; callers gate on getBoardCapabilities(boardName).auroraAppLink and narrow via toAuroraBoardName. Per-board behaviour unchanged (kilter still no link, tension still the 2 suffix).

packages/mobile/src/components/play-drawer/BoardseshGradeSection.tsx
marcodejongh: Done — new getBoardCapabilities(boardName) table in @boardsesh/board-config (crowdGrade / holdFilters / climbCreation / nativeBoardControl / auroraAppLink), typed Record<BoardName,…> so a new board without a decision is a compile error. The scattered lacksCrowdGrade / supportsClimbCreation / supportsHoldFilters / supportsNativeBoardControl predicates are deleted and every call site reads the table (fa26d819d).

packages/mobile/src/lib/ble/adapter-factory.web.ts
```diff
@@ -1,12 +1,20 @@
...
+//
+// `options` is accepted for signature parity and ignored: Web Bluetooth picks
+// the write type itself from the characteristic's properties, and the only
+// board that asks for acknowledged writes (Woods) has no web encoder anyway.
+export function createBluetoothAdapter(
+  _devicePicker: DevicePickerFn,
+  scanFamily: BoardScanFamily,
+  _options?: BleAdapterOptions,
+): BluetoothAdapter {
   return new WebBluetoothAdapter(scanFamily);
```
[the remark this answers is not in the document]
marcodejongh: Fixed — BleAdapterOptions now reaches WebBluetoothAdapter and writeCharacteristicSeries gained forceWithResponse: with the preference every 20-byte chunk goes writeValueWithResponse even though the characteristic advertises write-without-response; without it the path is byte-identical (both pinned by web-adapter tests).

packages/mobile/src/lib/ble/use-board-bluetooth.ts
marcodejongh: Split out — the branch is now dispatchWoodsPacket(frames, sizeId, write, signal) next to dispatchMoonboardPacket, returning unknown_size | incompatible | sent | cleared with the encoder skip counts; the hook keeps only the Alert/track/Sentry mapping. Direct 7-test suite against the real encoder + LED table.

marcodejongh:
All review threads addressed and resolved in `fa26d819d` (capability table, dispatchWoodsPacket split, web acknowledged writes, shared board lists, freeze removal, issue refs). Post-merge prod import runbook: #4768 — the catalog import needs the write-capable DB URL from 1Password; nothing shows in the app's Woods lists until it runs.

marcodejongh:
Both review items fixed in 79ac6d216: the catalog parse now goes through `parseWoodsCatalogFile(raw, fileName)` — malformed JSON rethrows with the file name, and a dump whose `problems` isn't an array (or `boardDimension` isn't a string) fails loudly instead of importing zero climbs; unit-tested. `BoardScanFamily` now documents that it names the UART protocol family and that Woods deliberately rides `'moonboard'` with acknowledged writes as an adapter option, not a third variant.

On the non-blocking notes: the Live Activity `?? 'kilter'` fallback is pre-existing and safe while `toBoardName` covers every UI board — agreed an assert would be better, leaving as-is here; the OG warm cache covers 12×12 only by the shape of `FALLBACK_BOARD_PREVIEW_CONFIGS` (one base per board) — an 8×10 card renders on first request, just uncached (424 of 5,392 climbs), acceptable for launch.

Issue #3281, opened 2026-06-29
Grade filtering on mobile: hard to find on iOS and invisible in our telemetry


marcodejongh:
## Symptom
A long-time power user opened the native 2.0 app for the first time and could not find the climb filters they relied on, then searched entirely unfiltered. The filters do exist — but on iOS the grade filter is the hardest one to spot, and grade filtering is also a blind spot in our analytics, so we can't measure how widespread this is.

## Feedback
- **Rating 1** — drawer-feedback, Kilter, **iOS 2.0.0 (build 100399)**: _"Where are the filters now?"_
- Same person filtered heavily on the web app (grade ranges, custom sort) before migrating.

## PostHog evidence
- distinctId `6f3f55f4`, first logbook entry 2023 (~3-year user). Migrated web→native 2.0.0 on the feedback day (2026-06-25).
- **Web (posthog-js, May 11–Jun 23):** 10 `Grade Filter Changed` events, searches with explicit grade ranges (12/15/16/17/19/23), `activeFiltersCount` up to 4, custom `sortBy=ascents`.
- **Native 2.0.0 (posthog-react-native, Jun 25–28):** 12 `Climb Search Performed`, **all `hasQuery=False`**, every granular filter prop null.
- **Cohort:** `Grade Filter Changed` = 10,933 webview events vs **0** native events ever; `minGrade`/`maxGrade` never present on native search payloads (0 of 6,747 native searches, 415 users). Grade filtering is **both uninstrumented and apparently harder to reach**.
- **Counterpoint (don't over-claim):** 5,672 / 7,594 native searches (75%) carry `activeFilterCount>0`; hold-filter (0.26 native vs 0.28 webview) and zone usage (0.16 vs 0.15) per active user match. Filters did **not** collapse wholesale — this is grade-filter discoverability on iOS plus a telemetry hole.
- No technical failure: zero `$exception`/`$rageclick` in this user's native sessions; ~15 Quick Tick Saved and dozens of sends in the same session. No session recordings for this distinctId.

## Code confirmation
- Filters and grade filtering both exist in the shipped app. Grade is a dedicated visible pill on the **Material** variant: `packages/mobile/src/components/search/GradeFilterControl.tsx`, rendered beside the search field in `packages/mobile/src/components/search/ClimbTopChrome.tsx:219` (next to `FilterButton.tsx`).
[...]

PR #3311, opened 2026-06-29
feat(mobile): log which search filters get used (#3290)

### LINKED ISSUE #3290: Instrument native search: log which filters get used (grade, sort, ascents)
## Problem

The native app's search/filter flow is a blind spot in our analytics. We can see *that* people search, but not *how* they filter — which means we can't measure filter discoverability, can't tell which filters are used, and can't quantify complaints like "where are the filters" (#3281). This is the enabling work that makes #3281 measurable.

## What's missing

- **`Climb Search Performed` on native carries almost no filter detail.** It sends only `hasQuery`, `queryLengthBucket`, `resultCount`, the board params, and `activeFilterCount` — but **none** of `minGrade`/`maxGrade`/`sortBy`/`sortOrder`/`setterCount`/`minAscents`/`minRating`. So we know a filter was active, never which one or its value.
  - Emitted at `packages/mobile/app/(tabs)/climbs/index.tsx:551`.
- **No mobile grade-change event exists.** `Grade Filter Changed` fires **web-only** (`packages/web/app/components/search-drawer/accordion-search-form.tsx:116`); the native filter UI emits nothing equivalent. The shared event name already exists in `packages/shared/analytics/src/events.ts:85` — it's just never fired from `packages/mobile`.

## Evidence of the gap

From PostHog (2026-06-08 → 2026-06-29, native `posthog-react-native`):
- `Grade Filter Changed`: **10,933** events on the legacy webview vs **0** on native, ever.
- `minGrade`/`maxGrade` present on **0 of 6,747** native `Climb Search Performed` payloads (415 searching users).

So today there is no way to answer "what share of native users set a grade range?", "which filters actually get used on mobile?", or "did the new filter affordance improve discoverability?".

## Proposed work

1. Add the filter dimensions to the native `Climb Search Performed` payload at `packages/mobile/app/(tabs)/climbs/index.tsx:551` — `minGrade`, `maxGrade`, `sortBy`, `sortOrder`, `setterCount`, `minAscents`, `minRating` (mirror what web already sends so the two are comparable). The values come from the canonical filter state (`packages/shared/climb-filters/src/filter-state.ts`).
2. Fire a mobile equivalent of `Grade Filter Changed` from the native grade control (`packages/mobile/src/components/search/GradeFilterControl.tsx` / `FilterButton.tsx`), reusing the shared event name in `packages/shared/analytics/src/events.ts`.
3. Sanity-check the new events land in PostHog before closing.

## Severity

P3 — pure instrumentation, no user-facing change. But it's a precondition for prioritizing and verifying #3281 (mobile grade-filter discoverability), so it's higher-leverage than a typical P3.

## Related

- #3281 — Grade filtering on mobile is hard to find on iOS and invisible in our telemetry (this issue is the "invisible in our telemetry" half, split out so it can be scheduled independently).

_Filed from PostHog product feedback triage (rating <=3, real store builds)._

marcodejongh:
## Summary

Closes #3290. Native search was an analytics blind spot — we could see *that* people searched, never *how* they filtered. This is pure instrumentation (no user-facing change) and the precondition for measuring #3281 (mobile grade-filter discoverability).

**1. Enrich the mobile `Climb Search Performed` event** (`packages/mobile/app/(tabs)/climbs/index.tsx`) with the individual filter dimensions: `minGrade`, `maxGrade`, `sortBy`, `sortOrder`, `setterCount`, `minAscents`, `minRating`. Field names + the `0`-means-unset sentinel match web's `Climb Search Performed` (`DEFAULT_SEARCH_PARAMS` all default to 0; difficulty ids start at 10, so 0 is never a real grade), so one PostHog query (`minGrade > 0`) compares both platforms. Previously `minGrade`/`maxGrade` were present on **0 of 6,747** native payloads.

**2. Fire a native `Grade Filter Changed` event** from *both* grade-commit surfaces — the chrome quick grade rail and the full filter sheet's Apply — tagged with a mobile-only `source: 'chrome_rail' | 'filter_sheet'` so we can split discoverable-affordance use from the buried sheet (the core #3281 question). The payload mirrors web's snake_case fields (`filter_kind`, `min_grade_id`, `max_grade_id`, `range_size`, `previous_*`, `extended_range_within_window`, `board_name`), so native lands in the **same** PostHog funnel as web's 10,933 legacy-webview events (native was 0).

**3. Shared plumbing.** Add `SHARED_EVENTS.GradeFilterChanged` and a reusable, tested `describeGradeFilter` helper in `@boardsesh/climb-filters`. Web now uses both (dropping its duplicate local `describeFilter`), guaranteeing the two platforms classify grade shape identically. The mobile grade rail (`GradeRail.tsx`) now threads the rule-3 tap `meta` through `onChange` so `extended_range_within_window` can be reported.

## Release Notes

## Test plan

- `vp run typecheck:mobile`, `vp run typecheck:web`, `vp run typecheck:shared` — green.
- `vp check` on all changed files — green.
- `vp test run` — shared `grade-range` tests (incl. new `describeGradeFilter` cases) pass; mobile suite **2866 passed**; web `accordion-search-form-handlers` (22) pass.
- Manual (`__DEV__` analytics `onDebug` console): change grade via the chrome rail → `Grade Filter Changed` with `source: 'chrome_rail'`; via the filter sheet + Apply → `source: 'filter_sheet'`; run a filtered search → `Climb Search Performed` carries the new grade/sort/setter fields.
- Post-merge: sanity-check both events land in PostHog (project 412845) with non-null `min_grade_id` + `source` on native, per issue step 3.

https://claude.ai/code/session_01FL4b6yHWrM4uUg9UMToYGp

### DIFF
--- a/packages/mobile/app/(tabs)/climbs/index.tsx
+++ b/packages/mobile/app/(tabs)/climbs/index.tsx
@@ -12,9 +12,11 @@ import {
   mergeBoardFilters,
   countActiveFilters,
   hasActiveBoardFilters,
+  describeGradeFilter,
   DEFAULT_CLIMB_FILTER_STATE,
   DEFAULT_CLIMB_BOARD_FILTER_STATE,
   type ClimbBoardFilterState,
+  type GradeTapMeta,
 } from '@boardsesh/climb-filters';
 import {
   KILTER_HOMEWALL_LAYOUT_ID,
@@ -360,6 +362,12 @@ function ClimbListInner() {
   const gradesRef = useRef(gradesData);
   gradesRef.current = gradesData;
   const grades = useMemo(() => gradesData ?? [], [gradesData]);
+  // Ascending difficulty ids — feeds describeGradeFilter so the `Grade Filter
+  // Changed` event reports a range's size the same way web does.
+  const gradeIds = useMemo(
+    () => grades.map((grade) => grade.difficultyId).sort((first, second) => first - second),
+    [grades],
+  );
   const { formatGradeByDifficultyId } = useGradeFormat();
 
   const boardConfig = useMemo(
@@ -564,6 +572,19 @@ function ClimbListInner() {
       // active filter, so the analytics count and the UI never disagree (and a
       // grade-only search reports 1, not 0).
       activeFilterCount: countActiveFilters(filters, boardFilters),
+      // Individual filter dimensions (issue #3290): web sends these so the two
+      // platforms are comparable — "what share set a grade range / changed the
+      // sort?". `0` = unset to match web's sentinel (DEFAULT_SEARCH_PARAMS all
+      // default to 0; difficulty ids start at 10, so 0 is never a real grade),
+      // so a single PostHog query (`minGrade > 0`) works on both platforms.
+      // `setterCount` is the size of the setter filter (0 = none, like web).
+      minGrade: filters.minGrade ?? 0,
+      maxGrade: filters.maxGrade ?? 0,
+      sortBy: filters.sortBy,
+      sortOrder: filters.sortOrder,
+      setterCount: filters.setter?.length ?? 0,
+      minAscents: filters.minAscents ?? 0,
+      minRating: filters.minRating ?? 0,
     });
   }, [firstSearchPage, name, filters, boardFilters, boardName, layoutId, sizeId, setIds, angle]);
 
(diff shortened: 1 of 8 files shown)


marcodejongh:
Thanks for the review.

1. **Web `gradeIds` sort** — fixed in 263abb58e. Added a defensive `.sort()` so web matches mobile and the shared `describeGradeFilter` ascending-ids contract holds even if the source list ever changes. (`BOULDER_GRADES` is already ascending, so it is behaviour-preserving today.)

2. **`track` in `trackGradeFilterChanged` deps** — left as-is intentionally. `track` is a stable module-level export from `src/lib/analytics.ts`, not a component-scope value, so `react-hooks/exhaustive-deps` does not flag it (CI lint is green) and the existing `Climb Search Performed` effect in the same file omits it too. Adding it would diverge from the established convention without changing behaviour.

PR #3321, opened 2026-06-29
feat(mobile): native @expo/ui Forms for OTA channel + branch switchers (#3271)

### LINKED ISSUE #3271: Mobile (@expo/ui): dev/tester settings screens — native Forms for Channel + Branch switchers (keep Dev Servers RN)
> Part of the **@expo/ui native-UI migration** follow-up (after #3254 + #3261). Maintainer guidance: only migrate where it *legitimately improves the experience*; FlashList/FlatList lists, SVG board-art, charts, and gesture boards stay React Native. Pattern: `docs/expo-ui-components.md`.

> **Drafting-agent verdict:** `maybe` — Low priority, dev/tester-weighted, and blocked on the native-text-inputs primitive. ChannelSwitcher + BranchSwitcher are clean Form copies of the merged MoreForm/FeatureFlagsForm pattern (only ChannelSwitcher's preview list is non-tester-facing, so modest real value); DevServerSwitcher should stay RN (pull-to-refresh + react-query states + sheet + custom modal don't fit a Form). Do it last in the set, after native-text-inputs lands.

---

## Why this improves the experience

Honest, low-priority. These are the OTA/dev tooling screens, mostly tester- and developer-only:

- **ChannelSwitcherScreen** — the one with a genuinely user-facing slice: the "preview channels" list is shown to **everyone** (signed-out included) via the changelog "Try a preview" entry; the preset/custom/Sentry sections are tester-only.
- **BranchSwitcherScreen** — gated on `isPreviewBuild()`, tester-only.
- **DevServerSwitcherScreen** — `__DEV__`/tester-only Metro picker.

The cards, section headers, and rows are hand-rolled today (`View` + `SectionHeader` + `ListRow` + `InfoRow`). Moving Channel + Branch to a single native `Form`/`LazyColumn` buys the grouped, inset-rounded native Settings look for free, plus a **native text field** (native keyboard, focus, autocorrect/capitalization control) for the manual-channel / custom-branch inputs that today use a bare RN `TextInput` + tap-target button. That's a real but modest win, concentrated on ChannelSwitcher (the only screen non-testers see).

**DevServerSwitcherScreen should stay React Native.** Its structure is a poor Form fit and the audience is dev-only: pull-to-refresh (`RefreshControl`), react-query loading/error/empty branches, a bottom `Sheet` with a QA-plan panel, a custom `Modal` + `TextInput` "Add Metro Server" prompt, and per-row info/remove icon buttons. Rebuilding that as a SwiftUI `Form` + `.refreshable` + native sheet is a large, risky rewrite for zero end-user benefit. At most, swap its inner `TextInput`s for the shared native text-input wrapper once that exists — nothing else.

Net: recommend migrating **ChannelSwitcher** and **BranchSwitcher**; explicitly **keep DevServerSwitcher as RN**. This whole area is blocked on the native text-inputs primitive (see Dependencies) and is the lowest priority in the set.

## Current state

RN implementations + their routes:

- `packages/mobile/src/components/ChannelSwitcherScreen.tsx` → route `packages/mobile/app/channel-switcher.tsx`
- `packages/mobile/src/components/BranchSwitcherScreen.tsx` → route `packages/mobile/app/(tabs)/profile/branch-switcher.tsx`
- `packages/mobile/src/components/DevServerSwitcherScreen.tsx` → route `packages/mobile/app/(tabs)/profile/dev-servers.tsx` (**keep RN**)

Shared building blocks they render: `./SectionHeader`, `./ListRow`, `./InfoRow`, `./Icon`, `./Text`, plus `react-native` `ScrollView` / `TextInput` / `Pressable`. Provider/logic deps that **stay in the screen** (view-model layer, not the native tree): `useTheme`, `useConfirm` (dialog-provider), `Alert.alert`, `hapticLight`/`hapticError`, `expo-updates`, and the pure state machine in `packages/mobile/src/lib/channel-switch.ts` (`buildChannelList`, `performChannelSwitch`, `performChannelReset`) — already node-tested in `packages/mobile/src/lib/__tests__/channel-switch.test.ts`.

ChannelSwitcher structure: an "Current" InfoRow card; a `common`-namespace i18n preview list (loading/error/empty/rows, with per-row trailing spinner / check / chevron); tester preset-channel list; manual-channel `TextInput` + switch `Pressable` (`autoCapitalize="none"`, `autoCorrect={false}`); a reset `Pressable`; a tester-only Sentry test section (3 button rows). BranchSwitcher is structurally a subset of the same (current card, branch list, custom-branch `TextInput` + switch, reset), all copy `i18n-ignore`d.

No screen-level render tests exist for any of the three (only the pure `channel-switch.test.ts`), so no test currently mounts these components.

There is **no existing `@expo/ui` `TextField`/`SecureField`/`useNativeState` usage anywhere in the app** — these screens would be the first text-input consumers, which is exactly why they depend on the shared native text-input wrapper landing first.

## Target

Per screen, one `Host` wrapping a single native tree, mirroring the merged `MoreForm`/`FeatureFlagsForm` pattern:

- **iOS** (`@expo/ui/swift-ui`): `Form` + `Section`, `Host style={{ flex: 1 }} useViewportSizeMeasurement colorScheme={colorScheme}`. Rows: nav/info → `Button`/`HStack`/`Text`; channel/branch list rows → `Button` with trailing `Image systemName="checkmark"` (active) or a `ProgressView` (switching); reset → `Button role="destructive"`; Sentry actions → `Button`s. Text input → `TextField` from `@expo/ui/swift-ui/TextField`.
- **Android** (`@expo/ui/jetpack-compose`): `LazyColumn` + `Card`, `Host style={{ flex: 1 }} colorScheme={colorScheme}`. Rows via `Row`/`Column`/`Text`/`Icon`. Text input → `TextField`/`OutlinedTextField` from `@expo/ui/jetpack-compose/TextField`.

**Native text input — confirmed capabilities + caveats** (verified against the installed 56.0.18 `.d.ts`):

- iOS `TextField` (`swift-ui/TextField`) holds text via `ObservableState` from `useNativeState('')` (NOT a plain controlled `value` prop) — or omit `text` and read `onTextChange`. Props: `placeholder`, `autoFocus`, `maxLength`, `onTextChange`, `onFocusChange`, `ref` (`focus`/`blur`/`clear`/`setText`). The RN-ish input attributes are **modifiers**, not props: `autocorrectionDisabled()`, `textInputAutocapitalization('never')`, `keyboardType('url')`, `submitLabel('go')` + `onSubmit(handler)` (all in `@expo/ui/swift-ui/modifiers`). So today's `autoCapitalize="none"` / `autoCorrect={false}` for channel/branch/host names **are** expressible — apply them as modifiers.
- Android `TextField`/`OutlinedTextField` (`jetpack-compose/TextField`) holds text via `value?: ObservableState<string>` (`useNativeState`) and exposes `keyboardOptions` (`{ capitalization: 'none', autoCorrectEnabled: false, keyboardType: 'uri', imeAction: 'go' }`) + `keyboardActions.onGo`, plus `isError`, `enabled`, `readOnly`, `colors`, and slot children (`.Label`/`.Placeholder`/`.SupportingText`/`.LeadingIcon`/`.TrailingIcon`).
- **No built-in submit-button-inside-field, error-text-by-default, or password reveal toggle** — fine here (these are plain identifier fields). The existing `TextInput` validation (`trim()`, disabled-while-switching) stays in the screen and gates the row `Button`.
- The trailing per-row spinner/check/chevron is conditional on row state — map switching → `ProgressView` (iOS) / `CircularProgressIndicator` or a muted glyph (Android), active → checkmark, else nothing/chevron, exactly as the RN `trailing` prop does now.

> Recommended: do the shared native text-input wrapper FIRST (the native-text-inputs dependency below), then consume it here so the `useNativeState` bridging + modifier set isn't reinvented per screen.

## Approach

Follow the platform-split convention (`docs/expo-ui-components.md`) and copy the merged Form references:

1. Extract each screen's render into a platform-split native component while the existing screen file keeps every hook, route guard, `t()` call, `useConfirm`, `Alert`, haptics, and `channel-switch.ts` calls, and builds a plain view-model — exactly how `FeatureFlagsScreen.tsx` feeds `<FeatureFlagsForm />`. Reference set to copy, in order of closeness:
   - `packages/mobile/src/components/FeatureFlagsForm.{types,logic,d,ios,android}.tsx` + `FeatureFlagsScreen.tsx` — closest shape (Form + rows + a destructive reset button; screen owns guards/data).
   - `packages/mobile/src/components/MoreForm.{types,logic,d,ios,android}.tsx` — richer reference for nav/info rows, sections, the Android `LazyColumn`/`Card` flatten, and the semantic-icon → SF Symbol / Material XML mapping if you add leading icons.
2. For each migrated screen `Foo` create: `Foo.types.ts` (props/view-model, no native imports), `Foo.logic.ts` (any pure helpers — most logic already lives in `channel-switch.ts`), `Foo.ios.tsx` (`@expo/ui/swift-ui` only), `Foo.android.tsx` (`@expo/ui/jetpack-compose` only), `Foo.d.ts` (ambient `export declare const Foo: FC<FooProps>`). No `Foo.tsx`. Suggested split: rename the existing screen to keep its name as the data/view-model owner and introduce `ChannelSwitcherForm.*` / `BranchSwitcherForm.*` as the native bodies (mirrors `FeatureFlagsScreen` → `FeatureFlagsForm`).
3. Brand accent only via the existing bridge `packages/mobile/src/theme/expo-ui-modifiers.ts` (`brandAccentColor`, `switchBrandColors`, `segmentedBrandColors`); read `brandColors`/`colorScheme` from `useTheme()` and pass `colorScheme` to the `Host`. Native controls already read system colours.
4. Native text slots take a `<Text>` child, never a raw string. Host sizing for a Form/LazyColumn is `flex: 1` + (iOS) `useViewportSizeMeasurement` — NOT `matchContents` (that under-reports and clips).
5. If any new suite ends up transitively rendering a migrated component, add a faithful passthrough stub under `packages/mobile/test/` and a regex alias in `packages/mobile/vite.config.ts` (copy `test/feature-flags-form-stub.tsx` + its alias block). Currently no test mounts these screens, so this is only needed if you add one — but the stub keeps the door open and matches the convention.

## Out of scope / keep as React Native

- **DevServerSwitcherScreen** (`packages/mobile/src/components/DevServerSwitcherScreen.tsx`) — stays RN: `RefreshControl` pull-to-refresh, react-query loading/error/empty states, the `@expo/ui/community/bottom-sheet` `Sheet` with the QA-plan panel, the custom `Modal` + `TextInput` add-server prompt, and per-row info/remove icon buttons don't map cleanly to a `Form`, and the audience is dev-only. Optional, only if the text-inputs primitive lands: swap the two inner `TextInput`s (custom-channel-style fields) for the shared native wrapper; leave everything else.
- `AccessibilitySettingsScreen` (`packages/mobile/src/components/settings/AccessibilitySettingsScreen.tsx`) and `OkhslColorPicker.tsx` — bespoke board preview, OKHSL colour picker, SVG swatches; **not** a Form candidate, do not touch.
- FlashList/FlatList lists, SVG board-art, gifted-charts, gesture boards — never migrated.
- The pure `channel-switch.ts` state machine and its tests — unchanged.
- i18n: keep all `t()` calls (ChannelSwitcher's `common` keys) in the screen/view-model; the native tree renders resolved strings only. Tester-only copy stays `i18n-ignore`d as today.

## Validation

```
vp run typecheck:mobile
vp run test:mobile
vp run check:mobile-platform-imports
vp run check:mobile-bundle
vp check
```

`check:mobile-platform-imports` (`scripts/mobile-platform-imports-check.sh`) is the real backstop for swift-ui-only-in-`.ios` / jetpack-compose-only-in-`.android` — `vp check` does NOT catch it. JS-only / OTA-deliverable (`@expo/ui` 56.0.18 already autolinked) — no native build. Spot-check on a simulator/emulator that the manual text field accepts input, autocorrect is off, and switch/reset/Alert flows still work.

## Dependencies

- **native-text-inputs** (BLOCKING) — these screens are the first consumers of `@expo/ui` `TextField`/`SecureField`. Land the shared native text-input wrapper (the `useNativeState` bridging + the autocap/autocorrect/keyboard modifier set) first, then consume it here. Without it, each screen would hand-roll the same TextField bridging.
- **settings-forms / more-screen** (pattern source) — `MoreForm.*` + `FeatureFlagsForm.*` are already merged (PR #3261); this issue copies them. No code dependency, just the reference.

## Checklist

- [ ] Confirm the native text-inputs wrapper has landed; use it (do not re-bridge `useNativeState` per screen).
- [ ] ChannelSwitcher: split into a view-model owner (`ChannelSwitcherScreen.tsx`, keeps hooks/guards/`t()`/`useConfirm`/`Alert`/haptics/`channel-switch.ts`) + native body `ChannelSwitcherForm.{types,logic,d,ios,android}.tsx`.
- [ ] iOS `ChannelSwitcherForm.ios.tsx`: `Host` (`flex:1` + `useViewportSizeMeasurement` + `colorScheme`) → `Form` with current-info `Section`, preview-list `Section` (spinner/check/chevron trailing per row), tester preset `Section`, manual `TextField` (`autocorrectionDisabled()`, `textInputAutocapitalization('never')`, `submitLabel('go')`+`onSubmit`), reset `Button role="destructive"`, Sentry `Button` rows.
- [ ] Android `ChannelSwitcherForm.android.tsx`: `Host` (`flex:1` + `colorScheme`) → `LazyColumn` of `Card`s; manual `TextField`/`OutlinedTextField` with `keyboardOptions {capitalization:'none', autoCorrectEnabled:false, imeAction:'go'}` + `keyboardActions.onGo`.
- [ ] BranchSwitcher: same split (`BranchSwitcherScreen.tsx` + `BranchSwitcherForm.*`); keep the `isPreviewBuild()` guard and `i18n-ignore` copy in the screen.
- [ ] Brand accent via `expo-ui-modifiers.ts`; pass `colorScheme` to every `Host`; system colours otherwise.
- [ ] All native text slots are `<Text>` children; no raw strings.
- [ ] Verify no clipping — use `flex:1`/`useViewportSizeMeasurement`, not `matchContents`.
- [ ] DevServerSwitcher untouched (optionally swap only its inner `TextInput`s for the native wrapper).
- [ ] If a test renders a migrated screen, add `test/*-stub.tsx` passthrough + `vite.config.ts` alias (copy the FeatureFlagsForm block).
- [ ] Run the full validation sequence; confirm `check:mobile-platform-imports` passes.
- [ ] Manual sim/emulator check: text entry, autocorrect off, switch/reset/confirm/Alert paths, light/dark via in-app toggle.

marcodejongh:
## Summary

Closes #3271 (part of the @expo/ui native-UI epic #3275).

Migrates the OTA **Channel** and **Branch** switcher screens from hand-rolled RN (`ScrollView` + card/`ListRow`/`InfoRow` + bare `TextInput`) to a single-`Host` native form — a SwiftUI `Form` on iOS, a Jetpack Compose `LazyColumn` of Material cards on Android — the same whole-screen consolidation as `MoreForm` / `FeatureFlagsForm`. Each screen keeps every hook, route guard, `t()`/confirm/`Alert`/haptic, and the `channel-switch.ts` state machine, and now builds a plain view-model the native tree renders. `DevServerSwitcherScreen` stays RN (the issue rules it out).

**One shared form, not two.** The two screens are the same OTA-target switcher (info → target list → custom-text field → reset), so they share **one generic `SwitcherForm`** (sections of typed rows: `info`/`status`/`target`/`field`/`action`) — the same sections-of-rows shape `MoreForm` uses — rather than duplicating two near-identical native trees. This is a deliberate divergence from the issue's literal `ChannelSwitcherForm`/`BranchSwitcherForm` naming, in service of the "extract, don't duplicate" rule. The screens (`ChannelSwitcherScreen`/`BranchSwitcherScreen`) remain the view-model owners, as the issue requires.

**Text field.** The blocking dependency (native text inputs, #3266 / PR #3284) merged on the 29th. Its `AuthTextInput` renders its own `Host` and so can't nest inside another `Host`'s native tree, so the inline manual-channel / custom-branch field is a bare native `TextField` / `OutlinedTextField` placed directly in the form, reusing `AuthTextInput.logic`'s pure prop→modifier mappers + `textFieldBrandColors`. Row state is derived once by `deriveSwitchRowState` so iOS/Android can't drift.

JS-only / OTA-deliverable (`@expo/ui` 56.0.18 already autolinked) — no native build. Drops the now-orphaned `previewChannels.manualSwitchA11y` key (the native action button's accessible name is its visible "Switch" label).

## Release Notes

Dev/tester OTA-tooling screens; RN → native rendering with no behaviour change.

## Test plan

- `vp run typecheck:mobile` — green
- `vp run test:mobile` — green (adds `switcher-form-logic` + `ChannelSwitcherScreen` + `BranchSwitcherScreen` view-model tests; full suite 2905+ passing)
- `vp run check:mobile-platform-imports` — green (swift-ui only in `.ios`, jetpack-compose only in `.android`)
- `vp check` — format + lint clean

**On-device verification** of the native rendering is the remaining step. Note dev/Metro builds force `updatesUsable` false (`!__DEV__`), so the switch/preset/reset sections only appear in a real build — verify via the `pr-<N>` OTA channel (More → Development → OTA Channel Switcher) once this PR's preview publishes. Flows to exercise: manual field accepts input with autocorrect off; switch a preview channel → confirm → restart Alert; reset → confirm → returns to build channel; tester preset + Sentry rows; Branch switch/reset; light/dark via the in-app toggle.

### DIFF
--- a/docs/expo-ui-components.md
+++ b/docs/expo-ui-components.md
@@ -148,6 +148,27 @@ option a `Text` carrying a `tag` modifier; string-guard `onSelectionChange`); An
 declares a function (like `SegmentedControl`), not a `const`. Selection logic is shared via
 `makeRadioSelectHandler`.
 
+### Whole-screen forms: MoreForm / FeatureFlagsForm / SwitcherForm
+
+The settings screens consolidate an entire screen into ONE `Host` containing a single
+SwiftUI `Form` (iOS) / Compose `LazyColumn` of cards (Android), instead of one `Host`
+per control. The screen component owns the route guards, data hooks, `t()` calls,
+confirm/Alert/haptics, then builds a plain view-model that the native tree renders —
+`MoreForm.*`, `FeatureFlagsForm.*`, and `SwitcherForm.*` are the three. **Host sizing
+differs from a standalone control:** a Form/LazyColumn is a scrolling container, so the
+Host uses `style={{ flex: 1 }}` + (iOS) `useViewportSizeMeasurement` — NOT `matchContents`
+(which sizes to content and clips the scroll).
+
+`SwitcherForm` (`src/components/SwitcherForm.*`) backs **both** the OTA Channel and Branch
+switcher screens (`ChannelSwitcherScreen`, `BranchSwitcherScreen`) from one generic
+sections-of-typed-rows model (`info`/`status`/`target`/`field`/`action`), the same shape
+MoreForm uses — the two screens are the same OTA-target switcher, so they share one form
+rather than duplicating two near-identical native trees. Its inline text field is a bare
+native `TextField`/`OutlinedTextField` placed directly in the form (the standalone
+`AuthTextInput` renders its own `Host` and can't nest inside another), reusing
+`AuthTextInput.logic`'s pure prop→modifier mappers + `textFieldBrandColors`. Row state is
+derived once by `deriveSwitchRowState` (`SwitcherForm.logic.ts`) so iOS/Android can't drift.
+
 ## Buttons
 
 **A button _inside_ an existing native @expo/ui tree should be a native button,
--- a/packages/mobile/src/components/BranchSwitcherScreen.tsx
+++ b/packages/mobile/src/components/BranchSwitcherScreen.tsx
@@ -1,12 +1,6 @@
-import { useState, useCallback, useEffect, useRef } from 'react';
-import { View, ScrollView, ActivityIndicator, Alert, Pressable, StyleSheet, TextInput } from 'react-native';
+import { useState, useCallback, useEffect, useRef, useMemo } from 'react';
+import { Alert } from 'react-native';
 import * as Updates from 'expo-updates';
-import { Text } from './Text';
-import { SectionHeader } from './SectionHeader';
-import { ListRow } from './ListRow';
-import { Icon } from './Icon';
-import { InfoRow } from './InfoRow';
-import { useTheme } from '../providers/theme-provider';
 import { useConfirm } from '../providers/dialog-provider';
 import { hapticLight, hapticError } from '../lib/haptics';
 import { reportHandledError } from '../lib/error-reporting';
@@ -20,6 +14,9 @@ import {
   type ChannelSwitchDeps,
 } from '../lib/channel-switch';
 import { isPreviewBuild } from '../lib/preview-build';
+import { SwitcherForm } from './SwitcherForm';
+import { deriveSwitchRowState, isSwitchRowPressable } from './SwitcherForm.logic';
+import type { SwitcherFormModel, SwitcherRow, SwitcherSection } from './SwitcherForm.types';
 
 // The branch baked into the running update's manifest metadata (set by the OTA
 // server). Read-only and tokenless — surfaced so a tester can see which branch
@@ -38,15 +35,14 @@ function getCurrentBranchName(): string | null {
   return null;
 }
 
-// Preview-build sibling of the tester OTA Channel Switcher. Both repoint the
-// running build at a different update target device-locally — overriding only the
-// `expo-channel-name` request header (no EAS API token, no project-wide channel
-// remap) — so they share the same override key and the same commit/revert state
-// machine in `channel-switch.ts`. This screen is framed around branches: a tester
-// picks a preview channel/branch (or types one) and the build pulls that branch's
-// OTA on restart.
+// Preview-build sibling of the tester OTA Channel Switcher, rendered as a single
+// native @expo/ui form via the shared <SwitcherForm />. Both repoint the running
+// build at a different update target device-locally — overriding only the
+// `expo-channel-name` request header — so they share the same override key and the
+// same commit/revert state machine in `channel-switch.ts`. This screen is framed
+// around branches: a tester picks a preview channel/branch (or types one) and the
+// build pulls that branch's OTA on restart. Tester-only, so copy is `i18n-ignore`d.
 export function BranchSwitcherScreen() {
-  const { systemColors, spacing, borderRadius } = useTheme();
   const confirm = useConfirm();
   const [override, setOverride] = useState<string | null>(null);
   const [customBranch, setCustomBranch] = useState('');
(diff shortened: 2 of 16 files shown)


marcodejongh:
Thanks — addressed the review in `3561eea4f`:

- **iOS `FieldRow` callback stabilization** ✅ — now mirrors Android: latest-`onSubmit` ref + `useCallback` `handleTextChange` + `useMemo`'d modifiers array, so keystroke re-renders don't churn the native handler registration.
- **`previewQuery.isError` coverage** ✅ — added a test asserting the preview **error-status row**. The suggested `expect(...).toContain('manual')` assertion can't pass under Vitest: the `manual` fallback section is nested inside `if (updatesUsable)`, and `updatesUsable = isEnabled && !__DEV__` is always `false` because the mobile vitest config inlines `__DEV__: true` as a `define`. So the error branch's *testable* effect is the preview status row; the `manual`-on-error path is part of the OTA-gated set covered by on-device QA.
- **Android action-row icons** — left label-only intentionally, now with a code comment: an Android trailing glyph needs a bundled XML vector drawable per icon, and the action rows are the tester-only Sentry rows + Switch/Reset whose labels already carry the meaning. Rather than ship a partial icon set (the `send`/`warning`/`flame` drawables don't exist), the rows stay label-only.
- Branch test coverage / the pre-guard `useMemo` — the remaining Branch model sections are all `updatesUsable`-gated (same `__DEV__` limit), and the discarded-`useMemo`-on-non-preview is unavoidable under hooks rules with no practical impact, as noted.

marcodejongh:
Addressed in `aae791b34`:

- **Perf — unstable `previewQuery.data ?? []` memo dep** ✅ — fixed with a module-level `EMPTY_PREVIEW_CHANNELS` constant, so a loading query no longer rebuilds the form model each render.
- **`deriveSwitchRowState` active-while-switching case** ✅ — added the test asserting `'active'` is returned when a *different* row is mid-switch.

The two remaining notes I'm leaving as-is, with reasoning:
- **`isError` → `manual` section test** — still not reachable in Vitest: that section is nested under `if (updatesUsable)`, and `updatesUsable` is hard-`false` under the `__DEV__: true` define. It isn't extractable to `deriveSwitchRowState` either (that's per-row state, not section gating); the model builder is inline in the screen. The `isError` branch's observable effect (the preview error-status row) is now tested; the gated `manual` path is on-device-QA territory.
- **Stub `default: return null`** — intentional: the stub is a lenient passthrough, and row-kind exhaustiveness is already enforced at compile time by `assertNeverSwitcherRow` in the real `.ios`/`.android` trees (a new kind fails `tsc` before any test runs).

marcodejongh:
Last round — `2dc203eb4`:

- **Stub silently swallowing unknown row kinds** ✅ — the stub's `default` now calls `assertNeverSwitcherRow(row)`, so a future `SwitcherRow` kind fails at compile time (the `default` was absorbing the union) and at runtime, matching the real renderers.

Leaving the other two as-is:
- **`confirmSwitchConfirm` reuse for the manual "Switch" label** — declining for scope: this is a rendering refactor, and the original screen already reused that exact key for the manual switch button (same "Switch" / "Changer" / "Cambiar" strings). Introducing a dedicated `manualSwitchLabel` key would expand the i18n surface beyond the refactor; happy to split it into a follow-up if you'd prefer the decoupling.
- **Branch `useMemo` before the `!preview` guard** — unavoidable under the Rules of Hooks and harmless (cheap compute, discarded on non-preview builds), as you noted.

marcodejongh:
Rebased on latest \`main\`. Two things worth flagging in the reconciliation:

**1. Merged main's "any user can switch OTA channels" feature into the form model.** While this branch was open, `main` (#3324) evolved `ChannelSwitcherScreen` to: add a fixed **Production** row at the top of the preview list (everyone's route home), ungate the **preset** and **manual** sections so any user — not just testers — can switch, filter the build channel out of the presets, and resolve the build channel via `resolveBuildChannel()` (fixes the null-channel case on Android). Those semantics are now ported into the `SwitcherForm` view-model instead of the old inline JSX:
- Production row is a `target` row leading the `preview` section; tapping it runs the reset flow.
- `preset` section is gated on `updatesUsable` + `presetChannels.length > 0` (not `isTester`), using the `presetSection` string and filtering out `buildChannel`.
- `manual` section shows for everyone when `updatesUsable` (hint still only on error).
- Only the Sentry crash-test rows remain tester-only.

Updated the screen tests so the empty/error preview cases expect the leading Production row, plus a new assertion that it renders `active` on the build channel.

**2. Base is now Expo SDK 57 / RN 0.86.** `main` merged the SDK 57 upgrade (#3396) mid-flight. Reinstalled deps and re-validated the whole `@expo/ui` surface against `@expo/ui@57.0.3`.

Validation: `vp run typecheck:mobile` clean, `vp test --project mobile` 3199 passing, `vp check` clean on changed files, `check:mobile-bundle` OK. Native on-device rendering still needs the on-device QA noted earlier.

Issue #3269, opened 2026-06-29
Mobile (@expo/ui): build FilterChipRow.android.tsx (Jetpack Compose Material chips + dropdown menus)


marcodejongh:
> Part of the **@expo/ui native-UI migration** follow-up (after #3254 + #3261). Maintainer guidance: only migrate where it *legitimately improves the experience*; FlashList/FlatList lists, SVG board-art, charts, and gesture boards stay React Native. Pattern: `docs/expo-ui-components.md`.

> **Drafting-agent verdict:** `maybe` — Maybe / medium-low priority. The native Material chips + dropdown menus are a genuine native-feel win and OTA-deliverable, but the persistent chip row is a Liquid-Glass-only design choice; on Android filters deliberately live in the Material top chrome, so the row won't render without a coupled gating change that risks displacing a working idiom. Build it, but make it a flagged/design-gated change, not a silent swap.

---

## Why this improves the experience

Honest read: **partial win, with a coupled design decision.** Two things are true at once.

1. The Android side of `FilterChipRow` is a hard `return null` placeholder (`packages/mobile/src/components/search/FilterChipRow.android.tsx`). So the "persistent, glanceable filter chips under the climbs title" feature — Filters · N, Recent, Grade, Show toggles, Popularity, Min rating — exists only on iOS. Building the Jetpack Compose counterpart with real Material 3 `FilterChip`s + `DropdownMenu`s gives Android genuinely native controls (Material ripple, elevation, real dropdown menus, TalkBack semantics), not RN re-creations. That part is a legitimate native-feel improvement and it is JS-only / OTA-deliverable (`@expo/ui` 56.0.18 is already autolinked on Android).

2. **But the row is not currently reachable on Android, by design — not by accident.** The caller gates it on `showFilterChips = !filterInTopChrome`, and `filterInTopChrome` comes from `variantFeatures.material.filtersInTopChrome = true` (`packages/mobile/src/theme/variants/variant-features.ts`). On Material, filters deliberately live in the top toolbar (filter button + condensed filter-summary chip + grade in the Appbar / `ClimbTopChrome`). On Liquid Glass (iOS), `filtersInTopChrome = false`, so the persistent chip row shows instead and owns filtering.

[...]

PR #3328, opened 2026-06-30
fix(mobile): always show the Android filter-chip row (drop android-filter-chips flag)

marcodejongh:
## Summary

Follow-up to the native Android filter-chip row (#3269 / PR #3310). That PR gated the chip row behind an `android-filter-chips` PostHog **kill-switch** intended to default ON (`useFeatureFlag('android-filter-chips') === false`). In production the flag resolved to a hiding state, so Android users kept the **old top-chrome toolbar filters** even though the changelog already announced the chip row. The dev emulator (no PostHog flags configured) didn't reproduce it — it showed the chips — which masked the gap.

This removes the flag entirely. The persistent native Material 3 chip row is now the **unconditional** Android (Material) filtering surface, in step with Liquid Glass. `materialChips` derives straight from the variant feature (`features.filtersInTopChrome`) with no PostHog dependency:

```ts
const materialChips = filterInTopChrome;            // was: filterInTopChrome && !chipRowKilled
const showFilterChips = !filterInTopChrome || materialChips;
```

Also drops the flag from the feature-flag catalog and the kill-switch wording in `docs/expo-ui-components.md` + the component header. No component/rendering logic changes — purely the gating.

Re #3269.

## Release Notes

## Test plan

- `vp run test:mobile` — 2940 passed
- `vp run check:mobile-platform-imports` — pass
- `vp run check:mobile-bundle` — both iOS + Android bundles export OK
- `vp check` (touched files) — pass
- **On-device:** the local Android emulator's software GPU segfaults under xvfb in this environment (no software-only fallback), so capture ran in CI's `mobile-screenshots-ios`/Android job + manual device check instead. Rendering is unchanged from #3310 (gating-only diff).

https://claude.ai/code/session_01UNe7kMEivkixe11LTYJqkh

### DIFF
--- a/docs/expo-ui-components.md
+++ b/docs/expo-ui-components.md
@@ -137,8 +137,8 @@ test/switch-row-stub.tsx              # passthrough stub (RN Pressable + Switch)
 `FilterChipRow` is the second example — both sides are now built (SwiftUI menus on
 iOS, Jetpack Compose `FilterChip`s + `DropdownMenu`s on Android), showing the same
 split with a richer tree (menus, pickers / dropdowns) and the shared
-`FilterChipRow.logic` label helpers. On Android the chip row is the default climbs
-filtering surface (kill-switch flag `android-filter-chips`).
+`FilterChipRow.logic` label helpers. On Android the chip row is the climbs filtering
+surface (it replaces the Material top-chrome filter button + grade control + summary).
 
 `RadioGroup` (`src/components/RadioGroup.*`) is a third, two-platform example —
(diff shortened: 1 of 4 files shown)


marcodejongh:
Addressed the review in d8f8f235e: inlined `filterInTopChrome` at the two call sites and made `showFilterChips` an explicit `true` (it's always shown now), so there's no algebra left to work through. Behaviour unchanged. (No new gating tests — as noted, the logic is now a tautology; the chip set itself is covered by the shared `FilterChipRow.logic` unit tests.)

PR #3334, opened 2026-06-30
feat(mobile): logbook toolbar matches climbs search (#3257)

azuttre:
## Summary

Match the logbook toolbar to the climbs search (iOS Liquid Glass, behind the `logbook-filters` flag), and default the logbook to sends-only across web + mobile.

- **Facet chips** like the climbs search: Filter · Latest · Hardest · Grade · Angle · Show · Date — all always visible, neutral until set, amber + value when set. Grade/Angle/Date open inline controls; Show is a native toggle menu. iOS Liquid Glass only; Android/Material keep the current toolbar.
- **Amber** accent throughout the logbook filter UI (climbs search stays purple), with dark labels for contrast.
- **Sends-only default** for the logbook (web + mobile, not flag-gated): the logbook opens to your sends, with attempts addable from the filters. The Show facet rests neutral at sends-only.
- Search moved onto the "You" row; the Benchmarks-only toggle no longer clips; the date chip parses local midnight (no UTC day-shift); the inline rail closes when the filter sheet opens.

Active-chip logic is pure, unit-tested, and mirrors `countActiveLogbookFilters`; the angle and date controls are shared with the filter sheet.

## Validation

- `typecheck` (all packages), `check:mobile-variants`, mobile + web + shared logbook tests, `check:mobile-bundle` — green.
- Tested on device (iOS) via the `pr-3334` OTA channel.

## Release Notes

Your logbook now opens to your sends. Bring attempts back from the filters whenever you want. Old bookmarked logbook links open to sends by default too; flip attempts back on if you want both.


azuttre:
Tested on device (iOS) via the `pr-3334` OTA channel: the facet chips (Grade, Angle, Show, Date) with their inline controls, the amber theming, and the sends-only default all look and behave as intended. The Benchmarks-toggle clipping and the date-chip timezone fix check out too.

CI is green, and the review items are addressed or accepted (the one-time "both" -> sends-only migration clobbering an explicit "both" is intentional and acknowledged). Ready to merge.

marcodejongh:
image image

Couple of minor issues but can be fixed in followup:
- Colours are wrong, its using orange instead of violet.
- The background around the chips should be transparent 
- I would ditch the "You" as it already has the button group with "logbook" active
- Make sure it emits analytic events for search, then later we can optimise the top level to show most used
- expand the refine expand in the filter sheet by default image
- Seperate issue, but while Im here: we should run a claude design team on the logbook rows themselves atm I feel like they're not showing what a user in a logbook would like to see. Like where is the date or the comment of the logbook entry?

azuttre:
Thanks Marco. Replies:

- **Colour:** the orange/amber is on purpose, to set the logbook apart from the regular climb search. My worry was that if both surfaces look the same (violet), it gets confusing which one you are on. Happy to revisit if you feel strongly, but that was the intent.
- **Transparent background around the chips:** will fix.
- **Ditching the "You":** agreed, the Progress/Sessions/Logbook/Social group already shows where you are. Removing it across the profile tabs.
- **Analytics events for search:** adding them (search + which filters get used) so we can promote the most-used to the top level later.
- **Expand Refine by default in the filter sheet:** will do.
- **Logbook rows (date/comment):** good call. Let's take that as a separate design pass / issue rather than block this one.

Doing the rest in this PR since it is not merged yet.

azuttre:
Thanks Marco. Tested on device via the pr-3334 OTA and it's behaving. Rundown of the review points:

- Transparent chip background: done, the row sits on the chrome blur with no opaque band.
- "You" title: removed from all four profile tabs (kept only as an a11y label).
- Analytics: added LogbookSearched, LogbookFilterChanged, LogbookSortChanged.
- Refine section: expanded by default now.
- Migration re-run and the inline Chip style: both fixed. The v2 prefs stamp persists on load so the migration doesn't re-run every cold launch, and the static chip geometry moved into StyleSheet.
- Old bookmarked logbook URLs: single-status links keep their old meaning, and I surfaced the sends-only default in the release notes.

On the amber: keeping it on purpose. It separates the logbook and profile from the live climb search, which stays purple. That's actually a smaller version of something bigger I want to do: right now the climb lists on the profile and on the search page look too similar, and jumping between them and seeing different climbs is confusing. I don't want to take that on here, but it's on my mind for a follow-up after some design research. Flagging it so it's on your radar too.

marcodejongh:
@azuttre yeah definitely agree that the logbook needs to look more visually distinguishable, it also needs to show different information for its usecase.

Im just not totally convinced that the colour is the right way of making it distinguishable. But its the easiest and cheapest thing to change later so we can just leave it. Just be aware that the code review AI might complain for it not using design tokens.

We could probably cajoole ai into proposing a colour thats more inline with the rest of the palette. 

For now lets just focus on actually making the thing more useable, by improving the logbook & its contents. Then on the side we can continue to bike shed on the colours until we make up our mind 😆

azuttre:
Agreed on the redesign. The logbook should show info built for its use case (dates, attempts, your grade vs consensus, angle/session context) rather than looking like the search rows. I'll take that on next, plus row actions (delete on the trailing swipe, next to the edit that's already there). The color and theme review comes with it, so we settle the palette as part of making the logbook actually distinct rather than tweaking it in isolation.

I'd keep all that a separate PR so this one goes in as-is. It's green and the feedback's addressed, and like you said the color is cheap to revisit there.

Mind hitting Approve so I can merge this one? Then I'll kick off the redesign PR.

PR #3341, opened 2026-07-01
fix(backend): gate session query behind membership with pre-join preview

marcodejongh:
## Summary

Workstream B3 of the multi-user-sessions improvement plan — the one intentional behavior change in the backend plan.

The `session(sessionId)` GraphQL query had **no membership check**: any client holding a session UUID could read the full queue, roster, and connected board serial. The resolver comment claimed "the membership check guards the rest of the payload", but the only check was roster emptiness.

A blunt `requireSessionMember` gate would break shipped clients, so the fix is a **member/preview split** instead of an error:

- **Members** (WS connection in the session, cross-instance WS via distributed state, or an authenticated caller with a durable `board_session_participants` row) get the full payload, unchanged.
- **Non-members** get a redacted invite-preview: session metadata (`id`, `name`, `boardPath`, `goal`, `color`, `isPublic`, `isPermanent`, `startedAt`, `endedAt`) plus the full `users` roster (mobile's join-confirmation screen shows who's climbing before you commit), with `queueState: null`, `lastConnectedBoardSerial: null`, `isLeader: false`.
- **Dormant sessions** (empty live roster) still return `null` first, before any membership check — mobile relies on that to disambiguate via `sessionStatus`.

### Why not just require membership

Two shipped mobile flows intentionally call `session` without membership:
- `GET_SESSION` — pre-join preview for the join-confirmation screen (selects metadata + users, never `queueState`).
- `GET_SESSION_QUEUE_STATE` — post-mutation-failure resync over HTTP, where every request gets a fresh `http-<uuid>` connectionId (`yoga.ts`), so connection-based membership can never match. The client already null-guards `session?.queueState`. Authenticated resyncs now match via the durable participants row; anonymous HTTP resyncs get the preview (accepted degradation — the next reconnect FullSync covers them).

### New helper

`isSessionMember(ctx, sessionId)` in `resolvers/shared/helpers.ts`: **single-shot, non-throwing, no retry/backoff** (the retrying `requireSessionMember` stays exclusively for mutations/subscriptions). Check order, short-circuiting on true: local connection context → `distributedState.isConnectionInSession` → durable `board_session_participants` PK lookup when `ctx.userId` is set (same predicate as `verifyWidgetSession`) → false.

### Schema changes (both non-breaking for client documents)

- `Session.queueState: QueueState!` → `QueueState` (nullable). Also fixes a latent skew: `createSession`'s HTTP path already returned `queueState: null` against the non-null field.
- `Session.participantId: ID` → `ID!`, dropping the "TEMPORARILY NULLABLE" caveat — the audit is complete: every Session-returning resolver populates it via `buildSessionPayload`'s fallback chain (worst case `''`).
- Generated artifacts (`schema.graphql`, `types.ts`, client-preset `graphql.ts`) regenerated via `vp run codegen`.

### Compat matrix (pinned by `session-query-gate.test.ts`)

| Caller | Result |
|---|---|
| WS-connected member (context sessionId matches) | full payload incl. queueState |
| Authenticated past participant over HTTP (fresh `http-*` connectionId, durable participant row) | full payload |
| Anonymous HTTP caller for a live session they were "in" | preview (queueState null) — accepted degradation |
| Complete stranger with the UUID | preview: metadata + roster, queueState/board serial null, isLeader false |
| Dormant session (empty roster) | null |

`eventsReplay` membership behavior is unchanged (still the throwing, retrying `requireSessionMember` with no durable fallback — pinned by test).

## Release Notes

- Your session's queue is now visible only to people who've actually joined — invite links still show who's climbing before you join.

## Test plan

- New `packages/backend/src/__tests__/session-query-gate.test.ts` pins the full compat matrix + `eventsReplay` non-regression.
- `build-session-payload.test.ts` extended with the explicit `queueState: null` override (redaction) case.
- `vp run typecheck` — all packages green (schema change fans out to generated types).
- `vp run test:mobile` — 388 files / 2981 tests passed.
- `vp run test:web` — 4 timeout-flake failures in `quick-tick-bar.test.tsx` / `start-sesh-drawer.test.tsx`; the same files fail identically on clean `main` in this environment (verified side-by-side), unrelated to this change (no web runtime code touched).
- `vp check --fix` — clean on all changed files.
- Backend suite (`vp test run --project backend`) queued behind concurrent agent runs locally; CI is authoritative here.

### DIFF
--- a/docs/websocket-implementation.md
+++ b/docs/websocket-implementation.md
@@ -513,6 +513,16 @@ The `sessionStatus(sessionId)` query exists solely for this disambiguation:
 - It requires no auth by design: it exposes only existence + ended-state, and auth may not be restored yet at cold start. This is also why mobile can't reuse the web flow described above — the `GET_SESSION_SUMMARY` pre-flight requires an authenticated caller.
 - Client behaviour (`packages/mobile/src/providers/queue-provider.tsx`): anything but `active` (so `ended` or `null`) → clear the stored id; `active` → restore; fetch failure (offline cold start) → restore optimistically so the queue still comes back, since a dead session stays escapable via End Session.
 
+### Session Query Membership Gate (`session`)
+
+The `session(sessionId)` query serves two audiences with one payload shape, split by membership:
+
+- **Members** get the full payload: queue state, roster, `lastConnectedBoardSerial`, metadata. Membership is resolved by `isSessionMember` (`resolvers/shared/helpers.ts`) — a **single-shot, non-throwing** check, unlike the retrying `requireSessionMember` used by mutations/subscriptions. It short-circuits through: the connection's local context → `distributedState.isConnectionInSession` (cross-instance WS) → a durable `board_session_participants` row when `ctx.userId` is set (primary DB, same predicate as the widget guard). The durable fallback exists because HTTP GraphQL requests are stateless — each gets a fresh `http-<uuid>` connectionId (`yoga.ts`), so connection-based checks can never match an HTTP caller.
+- **Non-members** get a redacted invite-preview instead of an error: session metadata plus the full `users` roster (mobile's join-confirmation screen shows who's climbing before the user commits — `GET_SESSION`), with `queueState: null`, `lastConnectedBoardSerial: null`, `isLeader: false`. The roster-in-preview contract applies to private (`isPublic: false`) sessions too — an invite link is the access token. An **anonymous** HTTP caller who is genuinely in the session also lands here (no stable identity to check durably) — accepted degradation, relevant to mobile's `GET_SESSION_QUEUE_STATE` resync, which already null-guards `session?.queueState`.
+- **Empty live roster** returns `null` before any membership check runs (the dormant-session contract `sessionStatus` disambiguates, above).
+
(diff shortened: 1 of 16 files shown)


marcodejongh:
**Note on the `mobile-bundle` / `ci-status` failure — out of scope for this PR.**

The failing step is the Expo dep-health check (`vp run check:mobile-deps`): Expo published new patch versions upstream (`expo@56.0.13`, `@expo/ui@56.0.19`, `expo-router@56.2.12`, etc.), so the exact pins on `main` now read as drift. Evidence it's not this PR:

- The flagged packages are untouched here (this PR changes no dependencies; the only mobile file touched is a response-type comment in `operations.ts`).
- Sibling PR #3338 (independent branch) fails `mobile-bundle` with the identical list.
- PRs that don't touch `packages/mobile/**` (e.g. #3340) skip the job and go green.

Fixing it means bumping ~12 Expo native pins + `bun.lock` (fingerprint/OTA-sensitive, exact-pin policy) — that belongs in a dedicated dep-bump PR on `main`, not a backend security change. Once that lands, a rebase here goes green.

All other checks pass, including `test-backend`, `typecheck`, `codegen-drift`, and the review-fix follow-up commit.

PR #3343, opened 2026-07-02
fix(shared): board presence sync parity and render performance

marcodejongh:
## Summary

Fixes the board-presence ("now on the wall") sync and speed complaints on the client side.

**Sync — web was a strict subset of mobile and silently lost events:**
- Extracts a shared `createBoardPresenceClient` transport factory into `@boardsesh/board-presence-react` (injected `execute`/`subscribe`/`onConnected` closures). Both platform adapters are now ~40-line wrappers, so web gains what it was missing: **reconnect catch-up** (`onReconnect` → `runCatchUp`; Redis pub/sub has no replay, so a socket blip on web previously left the feed stale until an unrelated later event exposed the seq gap), **cold-join holder seed** (`fetchConnection`), and `fetchHistory` (unblocks durable-history pagination next).
- Reconnect detection hardened: the skip-first-connect flag is now factory-scoped, so a re-registration on an already-open socket (board switch, client rebuild) no longer swallows the first genuine reconnect.
- Web now re-syncs on tab foreground (`visibilitychange` → `refresh('foreground')`), mirroring mobile's AppState sync.
- Web now reports `BoardHistoryCatchUp` telemetry, so web drop/recovery rates are finally measurable in PostHog.

**Speed — render storms on every presence event:**
- `mergeHistory` preserves object identity for already-known `(climbUuid, seq)` entries and returns the same array/state when nothing changed — the catch-up that runs on *every* app resume no longer re-renders the whole history list.
- New narrow `useBoardPresenceWallClimb` context: the mobile search chrome and wall capsule stop re-rendering on holder/stats churn.
- Mobile BoardSheet content is mounted only while the sheet is presented (with a native `onChange` re-mount backstop for the re-present-during-dismiss-settle race), and the `BoardNowPlayingReceived` telemetry moved to the provider so it keeps firing regardless.
- Web sheet body unmounts while the drawer is closed; history rows and hero memoized; the always-mounted entry pill no longer subscribes to the volatile feed context.

Diff reviewed against the plan's do-not-regress checklist (reducer dedup/ordering/cap invariants, catch-up coalescing, renderer-agnostic package rules) before push.

## Release Notes

- Board history keeps itself in sync — the wall feed catches up automatically after connection blips and when you come back to the app
- Smoother board history: no more full-list redraws when you reopen the app

## Test plan

- New shared factory suite (`create-board-presence-client.test.ts`, 20 cases incl. reconnect skip-first/re-registration semantics)
- Reducer identity-preservation cases (`toBe` assertions), wall-climb context render-count probe
- Mobile board-sheet tests incl. the dismiss-settle race (mutation-verified: fails with the backstop removed) and late-backfill `BoardHistoryViewed`
- Web client/context/sheet tests extended (onReconnect, fetchConnection, fetchHistory, visibilitychange catch-up, telemetry)
- `vp run typecheck` + `typecheck:mobile` green; `vp run test:web` 4851 passed; `vp run test:mobile` 2987 passed; shared board-presence suites 106 passed; `vp run check:mobile-bundle` OK

https://claude.ai/code/session_01MGPxKbADkgeCZa1uoDQEER

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -198,7 +198,6 @@
     },
     "packages/mobile": {
       "name": "@boardsesh/mobile",
-      "version": "2.0.0",
       "dependencies": {
         "@boardsesh/analytics": "*",
         "@boardsesh/ble-protocol": "*",
@@ -399,6 +398,7 @@
       "version": "1.0.0",
       "dependencies": {
         "@boardsesh/board-presence": "*",
+        "@boardsesh/graphql": "*",
         "@boardsesh/shared-schema": "*",
       },
       "devDependencies": {
--- a/packages/mobile/src/components/board-presence/BoardSheet.tsx
+++ b/packages/mobile/src/components/board-presence/BoardSheet.tsx
@@ -440,6 +440,27 @@ export const BoardSheet = forwardRef<BoardSheetHandle, BoardSheetProps>(function
     [managed.handle, invalidatePendingActions],
   );
 
+  // Index-based re-mount backstop for a re-present that lands inside a dismiss
+  // settle window. Sequence without this: dismiss starts → user re-taps →
+  // `present()` sets isPresented(true) (no-op, still true) and the coordinator
+  // QUEUES the present (its pump early-returns while the dismiss is in flight)
+  // → the dismiss settles → `handleFullyDismissed` sets isPresented(false) →
+  // only THEN does the coordinator present the native sheet — leaving it fully
+  // up with the content unmounted until a close+reopen. The native `onChange`
+  // fires with index >= 0 whenever the sheet is actually up, so re-mounting
+  // here closes that gap. The synchronous set in `present()` stays as the fast
+  // path for the common (no-race) open.
+  const managedOnChange = managed.onChange;
+  const handleSheetChange = useCallback(
+    (index: number) => {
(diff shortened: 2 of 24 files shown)


marcodejongh:
CI note: the `mobile-bundle` failure is the Expo dep-health check (`check:mobile-deps`) flagging upstream Expo patch releases (expo 56.0.12 → ~56.0.13 line, 12 packages). It fails identically on main's latest CI run (28555245965) — pre-existing, not introduced by this diff. Fixing it means bumping the native pins (fingerprint-relevant), which belongs in a dedicated dep-bump PR. All other checks are green.

marcodejongh:
Re: review item 2 (fetchHistory/BACKFILL_HISTORY asymmetry) — intentional and documented rather than guarded: the reducer's only feeders are the live subscription and `boardRecentClimbs`, which serve the identical Redis snapshot. The upcoming pagination PR consumes `fetchHistory` in a separate `useBoardHistoryPagination` hook that keeps durable pages outside the reducer entirely (deduped against the live window, known-entry-wins), so no caller routes durable rows through `BACKFILL_HISTORY`. If one ever does, the mergeHistory docstring now names exactly this asymmetry. Items 1 and 3 are being addressed and pushed.

PR #3350, opened 2026-07-02
feat(mobile): logbook diary rows — status-led design, day dividers, guarded swipe-to-delete

azuttre:
## Summary

Redesigns the mobile Logbook rows (You → Logbook) so the logbook reads as a personal climbing diary instead of a second climb-search catalog — users (including us) kept losing track of which list they were on. PostHog backs the problem: 58% of mobile sessions that touch both lists bounce back and forth ≥2 times.

Design process: research proposal → scope doc → 4-lens panel review (iOS/RN feasibility, climber mental-model, a11y, adversarial code-check; 4× viable-with-changes) → this implementation. Scope doc: `.boardsesh/logbook-redesign-scope.md` (local, gitignored).

**Row anatomy** (`LogbookRow` rewritten, no longer wraps `ClimbListItemContent`):
- Status glyph leads (⚡ flash / ✓ send / ○ project) — shape-first, colour as reinforcement.
- Meta line carries *your* record: tries, board + angle, your 1–5 stars (`3★`), a pencil glyph when you left a note, time-of-day. Community stars/sends/setter are gone.
- Trailing grade is always your effective grade; when the crowd disagrees, a small consensus sub-line with a direction arrow (↑ = you found it harder).
- Board thumbnail dropped — the priciest cell render in the profile, and the main reason the two lists rhymed. This also removes the frameless-MoonBoard fallback layout (one layout for all boards).

**Day dividers** (new `LogbookDayDivider`): date anchors — not session cards — with a `N climbs · N sends · top V8` rollup that only renders once a day is fully loaded, so page-boundary counts never lie. Shown only in date-ordered views (Latest preset / custom date sort), keyed off the effective feed input; name search keeps them (still a timeline). Midnight-safe via focus-refreshed clock.

**Guarded swipe-to-delete**: trailing swipe reveals Delete; commit opens the shared destructive confirm (DELETE_TICK is a real Aurora-synced delete — no one-gesture commit). The target uuid is captured at commit time (FlashList can recycle the row under the dialog), and on success the tick is stripped from cached pages immediately instead of waiting for the sequential page refetch.

**Accessibility**: row label reads as a log entry ("Sent Purple Nurple, V7, you graded V7 community says V6, 3 tries, rated 3 of 5 stars, has note…"); swipe actions are mirrored as accessibility actions (edit/delete/more); dividers are headers (rotor day-jumping); fontScale-tiered meta line (time drops first, two-line layout at accessibility sizes — thresholds to be tuned on device).

**Shared logic** (`@boardsesh/logbook`, pure + unit-tested, web can reuse): day bucketing with complete-day detection and uuid dedupe (offset pagination can repeat rows across boundaries — the tab previously didn't defend against duplicate FlashList keys), divider-visibility predicate, consensus-delta direction, star/note/attempt display rules.

**Analytics**: new `Logbook Entry Edited` / `Logbook Entry Deleted` (`method: swipe|sheet|a11y`) — the edit/delete UX previously had zero telemetry; `Logbook Row Clicked` enriched with `rowIndex`/`hasNote`/`status`.

**Known follow-ups** (out of scope): make frameless MoonBoard ticks openable (this PR only stops the fake success haptic on their dead taps); the share-beta picker reuses this row and loses thumbnails — verify recognition on device, give it its own variant if it suffers.

## Release Notes

- Your logbook now reads like a climbing diary — day headers with your send count and hardest send of the day, plus which board you were on.
- Rows lead with how it went: flash, send, or project, with your tries, your stars, and your grade next to the community's call.
- Spot beta at a glance — a violet camera marks every climb you have a video for, and a pencil marks your written notes.
- Swipe a logbook entry right to edit, left to delete (with a confirmation).

## Test plan

- `vp run typecheck:mobile` / `typecheck:shared` — green.
- `vp run test:mobile` — 389 files / 2990 tests green, incl. rewritten `logbook-row.test.tsx` (grade/delta/stars/note/a11y/swipe-direction contracts) and new `logbook-tab-delete.test.tsx` (confirm gating, cache strip, method analytics). Shared: `day-rows.test.ts` / `row-meta.test.ts`.
- `vp check` — 0 errors.
- On-device (pending, via this PR's OTA channel): QA notes in `.boardsesh/qa-notes.md` — divider grouping/suppression, delete flow, VoiceOver pass, Dynamic Type tiers, dark mode + Material variant, share-beta picker recognition.


azuttre:
Addressed the review in 1e029e9a0:

1. **rowIndex counting dividers — fixed.** `rowIndex` is now an entry-only ordinal derived from a uuid map in the rows memo; the `index` prop plumbing came out of `LogbookRow` entirely. Test updated to assert `rowIndex: 0` for the first entry under a divider.
2. **`benchmarkDifficulty ?? '1'` — intentional, now commented.** `ClimbAttributeIcons` keys the © glyph on `Number(value) > 0` and never renders the value, so no phantom grade badge is possible. The `'1'` deliberately forces the glyph on for a benchmark tick whose grade names are missing — `isBenchmark` is authoritative; the old `null` fallback silently hid the glyph on those rows.
3. **Time absent at fontScale 1.15–1.3 — intentional, now commented.** Time is the spec'd first-drop part; the a11y label still speaks it and it returns in the two-line layout at ≥1.3.
4. **`estimatedItemSize` — declining.** This FlashList version auto-measures; `SessionsTab` runs the identical mixed header/row pattern with `getItemType` and no size hints. Happy to revisit if profiling on device says otherwise.
5. **sortBy roundtrip test — added.** `shouldShowLogbookDividers` is now asserted through the real `toAscentFeedInput` for every sort mode (presets, custom date, custom non-date, name search), so a backend enum rename can't silently kill dividers while the string assertions stay green.

azuttre:
Both review rounds addressed in 3277d065c:

- **Delete error path — covered.** `logbook-tab-delete.test.tsx` now drives `onError` (error haptic + `deleteError` toast, cache untouched, no analytics) and the a11y-initiated flow (`method: 'a11y'`).
- **`now` in renderItem deps — removed entirely.** Went further than the `extraData` suggestion: `LogbookDayDivider` owns its focus-refreshed clock, so `renderItem` has no clock dep at all and entry rows keep a stable identity across focus/pull-to-refresh. Only dividers re-render on focus.
- **`@boardsesh/logbook` → `profile-stats` coupling — dropped.** The package now depends only on `@boardsesh/graphql` + `dayjs`; the 3-line naive-UTC parse is inlined with a cross-reference comment to the canonical `parseTickTime` and the shared invariant.

Mobile suite green (2,992 tests), shared logbook suite green, typecheck + format clean.

azuttre:
Final review round addressed in 7000e8ea9:

- **Async IIFE catch — added.** A rejected confirm promise now resolves as a decline (nothing confirmed → nothing deleted, the safe outcome for a destructive action) instead of an unhandled rejection.
- **`spacing[1] + 2` — named.** `DIVIDER_VERTICAL_PADDING = 6` with the off-grid rationale in place.
- **`LogbookEntryEdited` method — documented rather than plumbed.** Unlike delete, every edit commits through the sheet (swipe and the a11y action only *open* it), so there's one commit path and `method: 'sheet'` is accurate; the events catalog now says so explicitly. Threading initiation source through the sheet would be plumbing for no analytical gain.
- **Shared `dragArmedRef` arming both swipe inners — declining.** This is a deliberate copy of `ClimbListRow`'s documented pattern ("we arm both sides; the non-dragged side stays occluded"). Splitting per-side arming would diverge the two rows' swipe machinery for one extra (invisible) animated reaction during an active drag; if it's ever worth optimizing, it should change both rows together.

azuttre:
Device-review feedback (Alex) is in as of 443824c5d, with one rollout wrinkle worth reading:

- **Layout label** — rows now name the wall by LAYOUT ("Kilter Homewall", "MoonBoard 2016") instead of the user-named board. Live on `pr-3350` now.
- **Beta-video marker** — the row shows a video glyph when a beta video is attached to the ascent, VoiceOver announces it, en/es/fr added. BUT the field selection is temporarily **out** of the feed queries: prod GraphQL rejects unknown fields and the pr-channel OTA bundles hit prod, so the first bundle with the query briefly broke the logbook feed on the preview channel (caught and rolled back within the hour). Backend lands first via #3355; once that deploys, the selection comes back here as a one-liner and the icon lights up.
- Also folded in this round's review notes: optimistic delete-strip moved INTO `useDeleteTick` (both delete paths + web now behave identically, and the query-key knowledge lives beside the invalidation list — resolves the sheet-vs-swipe inconsistency and the key-coupling concern in one move), `normalizeLogbookQuality` is float-safe, and the analytics ordinal map is read through a ref so `renderItem` survives page loads. On `estimatedItemSize` (third mention): FlashList here is **2.3.1** — v2 auto-measures and deprecates that prop; `SessionsTab` runs the same mixed-type pattern without it. Declining for good.

marcodejongh:
> rows now name the wall by LAYOUT ("Kilter Homewall", "MoonBoard 2016") instead of the user-named board. Live on pr-3350 now.

I'd prefer to keep the user names boards if there is one

azuttre:
@marcodejongh done — the row shows the user-named board when the tick has one and falls back to the layout product otherwise (Alex signed off). Live on `pr-3350` with the next publish. Also hardened the delete-test's async flush per the last review note.

azuttre:
Latest round addressed in 234f4ff99 (mirrored on #3355 where applicable):

- **`tickUuid` null-filter** — fixed; the Set is now `Set<string>`.
- **`flushRun` closure** — fixed; run/dayKey are explicit parameters, outer bindings can't be half-read.
- **Divider component test** — added (`logbook-day-divider.test.tsx`): today/yesterday labels, full rollup, attempts-only rollup, the no-rollup incomplete-day state, and the header role.
- **"Inert feature"** — intentional and temporary, not a follow-up PR: the field selection returns to the queries in THIS PR the moment #3355 deploys (backend-first because prod GraphQL rejects unknown fields and the OTA preview hits prod).
- **Shared `dragArmedRef`** — declined previously with the `ClimbListRow` precedent; unchanged.
- **Flash vs Flashed** — intentional grammar, not drift: the meta line uses the noun ("Flash · 4★"), the VoiceOver label needs a verb to read as a sentence ("Flashed Purple Nurple, …"), matching "Sent".

Also fixed a real test-semantics bug CI caught on #3355: caption matches de-dupe by climb keeping the newest send, so the beta assertion now uses two one-tick climbs.

azuttre:
Round addressed in dbb58472b:

- **`totalCount` after the optimistic strip — fixed.** The strip now decrements each cached page's total by the removed rows (test asserts both variants).
- **Share-beta picker regression — tracked.** #3357, blocked on the on-device verdict from this PR's channel review.
- **Parity between the inlined parse and `parseTickTime` — pinned.** New test in `day-rows.test.ts` asserts identical day keys for naive and Z-suffixed timestamps (profile-stats as a dev-only dep, runtime stays dep-light).
- **SDL nullability — declining, it's load-bearing.** `hasBetaVideo: Boolean!` would break the following/global activity feeds: they share `AscentFeedItem` and don't resolve the field, so non-null would turn their responses into GraphQL errors. Nullable-with-documented-producers is the correct contract; the field's SDL description says exactly which resolvers populate it.
- **fontScale thresholds — on the device checklist** (`.boardsesh/qa-notes.md`), pending the Dynamic Type pass.

azuttre:
Latest round, addressed in df637e166:

- **Duplicate-page double-decrement — fixed** (good catch). One deleted tick now takes exactly one off each page's `totalCount` regardless of offset-overlap duplicates; new test covers the duplicated-row case.
- **"Import `parseTickTime` instead of the dayjs dep" — declining; it contradicts the earlier round.** A runtime import of `@boardsesh/profile-stats` is precisely the transitive-surface coupling this review flagged two rounds ago (board-config, board-constants, play-view, shared-schema). `dayjs` is a leaf dependency already in the tree; the parity test exercises the profile-stats helper as a dev-only dep. This is the lighter shape.
- **Field-selection "rot" — no ticket needed.** Re-adding the selection is the completing step of THIS PR once #3355 deploys; the PR is a draft under active sequencing and won't merge without it (it's the feature the product owner asked for).
- **Delete `method` labelling — current behaviour is correct.** The trailing swipe deletes via the confirm dialog directly (`swipe`), never through the sheet. The edit-swipe → sheet → Delete path is genuinely a sheet decision (`sheet`).
- **Memoising `accessibilityLabel` — declining.** The row recomputes it only on prop changes (it's `React.memo`'d), where the memo's dep comparison costs as much as the join; singling out one derived string would be inconsistent with the row's other unmemoised derivations.

azuttre:
On the `totalCount` note — declining, the suggested version is the buggy one under current semantics. `totalCount` is the FEED-level total (the GraphQL schema defines it that way); every page carries a copy of the same number. Decrementing only the pages that physically held the row would desync those copies (a page without the row would still claim the old total). If `totalCount` ever became per-page, that's a schema change that rewrites this function wholesale — guarding against it now by desyncing the current contract is backwards. The review itself notes the final number is correct as written.

That closes out every open review thread. Remaining sequence to merge-ready: #3348 → #3355 → field selection re-added here → Alex's final on-device check of the beta icons.

azuttre:
Both items in this round were resolved in earlier rounds: FlashList sizing hints (as `estimatedItemSize`, now as `overrideItemLayout`) — declined; FlashList 2.3.1 auto-measures and `SessionsTab` runs the identical mixed-type pattern with no hints. Shared swipe-arm state — declined; deliberate copy of `ClimbListRow`'s documented pattern, and any change there should refactor both rows together. See the resolutions upthread.

azuttre:
Both issues in this round were real — fixed in the latest push:

- **Per-row `useWindowDimensions` — fixed.** The tab holds the single dimension subscription and rows take `fontScale` as a prop (default 1 for the share-beta picker), so a keyboard/rotation event no longer runs every visible row's reconciler.
- **Confirm-window re-entry — fixed.** A `deleteFlowActiveRef` covers the swipe-commit → dialog-resolution gap (where `isPending` is still false); cleared on decline, success, error, and rejection. New test covers the concurrent path: second swipe during an open dialog is ignored, and a decline re-arms the flow.
- Parity-test comment clarified re the Z-suffixed case.

azuttre:
Final polish in 92d63db1b: the ascending-sort grouping claim is now pinned by a test, and the all-pages `totalCount` decrement carries an explicit intent note. On the remaining two: the profile-stats devDep is test-time only and the suite runs in milliseconds — will revisit if that changes; the fontScale-in-deps remount on a system font-size change is the intended tradeoff (rows must re-measure then anyway), as the review itself concluded.

That closes every review thread. This PR is merge-ready pending #3370 → #3373 (backend ownership fix for the beta markers) landing first, plus final product sign-off.

azuttre:
Closing out this round without a churn commit — the branch is merge-ready and blocked only on #3370 → #3373:

- **`anyRemoved` naming** — fair nit; it'll ride along with the next functional push if one happens. Not worth a dedicated commit + full CI/OTA cycle.
- **Parity duplication risk** — acknowledged design decision from an earlier round (the alternative was the transitive-dependency coupling this review flagged before that). The parity test is the guard; "update both" is the contract.
- **Shared swipe-arm mounting both inners** — declined for the fourth time: deliberate copy of `ClimbListRow`'s documented pattern; changing it means changing both rows together, out of scope here.
- **Two-line layout test** — deliberately unpinned: the fontScale thresholds are marked tune-on-device and expected to change during product review; a snapshot of provisional values would fight that tuning. Will pin once the thresholds settle post-merge.

marcodejongh:
image

We could probably put the user quality rating to the left of the grade, and camera too.

Also I wonder if the default should show both sends and attempts, as i often go to logbook to attempt something from last sesh again.

And we probably should group multiple sends/attempts on the same thing.

Maybe the date divider line couod hold the session, then we can show the board info there to save space and duplication.

Also I kinda miss the thumbnails 😬

But looks good ^^ were just some future ideas

azuttre:
Thanks for the pass! Point by point:

1. Stars and camera to the left of the grade: agreed, I'll try it out in this PR and see if it helps readability.

2. Attempts in the default view: you've convinced me, with a sequencing caveat. I was defending sends-only, but it clicked that my own habit (logging on send day and guesstimating the tries) is fighting the data model. Someone diligently logging attempts per session is using the logbook right, and the default should reward that, especially since the status glyphs already tell sends and projects apart at a glance. The caveat: we should flip the default together with grouping (point 3), otherwise burn-per-entry logs get noisy again, which is what sends-only was protecting against. Sends-only was also a deliberate cross-platform default, so the flip covers web too.

3. Grouping repeats: concrete proposal. Day-scoped collapse: same climb, same day folds into one row using the existing slots. Best outcome wins the glyph and supplies the stars, note and video markers; tries sum for the day. The row always shows that day's tries; lifetime totals per climb and angle live in the climb's own logbook view where they can't be misread as a per-day figure. Editing or deleting an entry on a grouped row goes through a small day-entries chooser, and there is deliberately no delete-all-in-group. Follow-up PR together with point 2.

4. Session and board info on the date divider: I like that idea. For days on a single board the divider can carry the board and angle directly. For multi-board or multi-session days I want to try sub-dividers splitting the day, so the context stays truthful instead of being dropped. I'll play around with both in this PR.

5. Thumbnails: not opposed. The tradeoff is row height, so fewer rows on screen, which works against the scan-many nature of a logbook. Leaving this one out for a future PR.

To make it concrete: points 1 and 4 land in this PR, and I'll do a follow-up PR for points 2 and 3 plus the lifetime-per-angle view on the climb's logbook entry.

azuttre:
On the "unused devDependency" note: it's a false positive — the parity test loads `@boardsesh/profile-stats` via a dynamic `await import(...)` (`day-rows.test.ts`, the parseTickTime parity pin), which static dependency scans miss. The remaining notes in this round are re-treads of items already resolved or declined upthread (see the prior resolutions). No churn commits pre-merge per the earlier note; verdict stands at ready-to-merge.

azuttre:
On the latest round: the edit sheet's delete already guards with `isMutating` and runs inside a single modal sheet — the tab's `deleteFlowActiveRef` exists for the LIST's concurrent-swipe window, which has no analogue in a one-instance sheet whose confirm dialog blocks its own buttons. Not duplicated protection, different surfaces with the guard each needs. The mock-completeness note is taken for the Round B test files. No churn commits pre-merge, as before.

marcodejongh:
@azuttre We could just use a smaller thumbnail if we want to preserve space. Or have like a super condensed table view

marcodejongh:
Oops pressed ready for review by accident. Mark it as ready when you want me to review/merge it

azuttre:
Round A validated on device by @azuttre — marking ready for your review/merge. Everything discussed in the thread is in except the agreed follow-ups (grouping + attempts default land via #3377's client PR, thumbnails variants tracked in #3357).

PR #3351, opened 2026-07-02
chore(mobile): bump 12 Expo pins to latest SDK 56 patch releases

azuttre:
Fixes the `mobile-bundle` CI failure (`vp run check:mobile-deps`) on `main` and every open PR: Expo published new SDK 56 patch releases, and 12 of our pinned packages fell one patch behind the canonical pins in `expo/bundledNativeModules.json`, so `expo install --check` now exits non-zero.

Example failing job: https://github.com/boardsesh/boardsesh/actions/runs/28557791022/job/84669026455

## Changes

Bumps the 12 flagged pins in `packages/mobile/package.json` to the SDK 56 expectations (exact pins, matching existing style; `@expo/ui` keeps its `~` range) and refreshes `bun.lock`:

| Package | From | To |
| --- | --- | --- |
| @expo/ui | ~56.0.18 | ~56.0.19 |
| expo | 56.0.12 | 56.0.13 |
| expo-asset | 56.0.17 | 56.0.18 |
| expo-constants | 56.0.18 | 56.0.19 |
| expo-dev-client | 56.0.20 | 56.0.21 |
| expo-image-manipulator | 56.0.19 | 56.0.20 |
| expo-image-picker | 56.0.18 | 56.0.19 |
| expo-location | 56.0.18 | 56.0.19 |
| expo-modules-core | 56.0.17 | 56.0.18 |
| expo-router | 56.2.11 | 56.2.12 |
| expo-splash-screen | 56.0.10 | 56.0.11 |
| expo-updates | 56.0.19 | 56.0.20 |

No `--fix` was used (the deps script forbids it); pins were updated by hand and installed with `bun install`.

## Native-input impact (heads-up for in-flight PRs)

These packages are native inputs, so two derived keys change when this lands:

- **iOS screenshot-app cache**: the `mobile-screenshots-ios` cache key includes `packages/mobile/package.json` and the root `package.json`, so the first run after this merges rebuilds the simulator `.app` (~30 min).
- **expo-updates runtime fingerprint**: the native project hash changes. Existing TestFlight/store binaries will not pull OTAs published from post-merge `main` until a new binary is built, and per-PR OTA channels (`pr-<n>`) published on top of this change won't reach binaries built before it.

In-flight PRs relying on per-PR OTA channels for native-change testing (e.g. #3350) should either land **before** this PR, or rebase **after** it with the awareness that a new preview/dev-client build may be needed.

## Validation

- `vp run check:mobile-deps` (and the full `vp run check:mobile-bundle` path)
- `vp run typecheck:mobile`
- `vp run test:mobile`

## Release Notes


marcodejongh:
Heads-up: this bump transitively moves `expo-dev-launcher` 56.0.20 → 56.0.21, and bun matches `patchedDependencies` by exact version — the lock diff here silently drops the `expo-dev-launcher@56.0.20` patch entry (line: `-    "expo-dev-launcher@56.0.20": "patches/expo-dev-launcher@56.0.20.patch"`) while root `package.json` still declares it. That patch carries the 10s→120s dev-launcher request timeout that CI screenshot runs depend on, so merging as-is regresses it silently.

#3348 is the same 12-pin bump but also renames the patch to `patches/expo-dev-launcher@56.0.21.patch`, updates the `patchedDependencies` key, and verifies the patched constant lands in the installed 56.0.21 — plus a rollout note on the fingerprint change. Suggest either closing this one in favour of #3348, or porting the patch rename here.

azuttre:
Good catch — closing this in favour of #3348, which carries the same 12 pins plus the `expo-dev-launcher@56.0.21` patch re-key this PR misses (without it, the 10s→120s dev-launcher timeout patch is silently dropped from `bun.lock`).

marcodejongh:
Oh yeah this check is probably wrong, we dont want to follow every patch release as it changes the native binary

PR #3352, opened 2026-07-02
fix(mobile): remount climb filter sheet host on sub-picker resume

### LINKED ISSUE #3330: [bug] climb: filters can no longer be applied
Since the 2.0.1 iOS TestFlight builds, and including the latest 2.1.0 release, the apply button no longer appears in the filters popup:

Image

Selecting them works fine, but all the selections get cleared as soon as the popup is pushed back down to view the climbs again. The new quick filters outside of the popup do thankfully work, but they only account for the most commonly used options, so advanced filtering is effectively nonexistent right now.

I’m resorting to rolling back to 2.0.0 for now, but users not in TestFlight don’t have that option.

marcodejongh:
## Summary

Closes #3330.

Since 2.0.1, the Apply button vanished from the climb-filters sheet after visiting any sub-picker (holds / zone / setters) on iOS, and with it the only way to commit edits — swiping the sheet away then discarded the whole draft.

Root cause: the sub-pickers suspend the sheet (dismiss without unmount), push a route, and re-present on return. On iOS, re-presenting an **already-mounted** SwiftUI sheet host loses the detent height bound, so the sheet's `flex:1` column stops being clamped to the 90% detent — the scroll body grows to full content height and pushes the pinned Apply footer off-screen. The first present always works because `BottomSheetModal` mounts its host fresh; only the re-present path breaks. Android rebuilds its native host on every present, which is why it was immune.

Fix: bump a `presentEpoch` on resume and key the `BottomSheetModal` on it, so every re-present remounts the host and takes the same first-present path as the initial open — still fully serialized through the sheet presentation coordinator. Scroll offset and the Refine/Advanced section expansion are preserved across the remount. Also disables the Setters row while board config is loading (matching Holds/Zone).

JS-only — no native, config, or dependency changes, so the fix reaches shipped 2.1.0 binaries via OTA.

## Release Notes

- Filters work again: the Apply button no longer disappears after using the hold, area, or setter pickers, so your selections stick.

## Test plan

- `vp run typecheck:mobile`, `vp run test:mobile` (388 files / 2984 tests), `vp run check:mobile-bundle` — all green.
- New unit tests: the sheet host remounts (fresh first present) across a sub-picker suspend→resume cycle; Advanced-section expansion survives the round trip; scroll restore targets the pre-suspend offset and ignores mount-time scroll noise.
- Android emulator, two full round trips: Apply footer pinned and functional after returning from the hold picker, selection hands off ("1 hold"), Apply commits and the list filters, Refine/Advanced expansion preserved. Scroll restore does not take on Android (the Material sheet rebuilds its children on every present — same as before this change, so no regression); worth checking on iOS during OTA QA, where the children stay laid out across the remount.
- iOS (the platform the bug ships on): please QA via this PR's OTA channel — More → Development → OTA Channel Switcher → `pr-3352` — since the failure only reproduces against the real SwiftUI sheet host lifecycle.

https://claude.ai/code/session_01Vegna6LURUUfTrynUmLK5N

### DIFF
--- a/packages/mobile/src/components/ClimbFilterSheet.tsx
+++ b/packages/mobile/src/components/ClimbFilterSheet.tsx
@@ -153,10 +153,8 @@ export function ClimbFilterSheet({
   const scrollRef = useRef<ComponentRef<typeof BottomSheetScrollView>>(null);
   // Latest scroll offset, captured on scroll into a ref (no re-render).
   const scrollOffsetRef = useRef(0);
-  // Offset snapshotted at suspend time and restored after the host remounts on
-  // resume (see presentEpoch). A live read of scrollOffsetRef at restore time
-  // would race the fresh ScrollView's initial onScroll (y=0), which can land
-  // before onContentSizeChange and clobber the target.
+  // Snapshotted at suspend: the remounted ScrollView's initial onScroll (y=0) can
+  // land before onContentSizeChange, so a live read would restore to 0.
   const pendingRestoreOffsetRef = useRef(0);
   // The epoch whose offset has already been restored — makes the restore in
   // onContentSizeChange one-shot per remount, not re-fire on later content growth.
@@ -492,10 +490,8 @@ export function ClimbFilterSheet({
     scrollOffsetRef.current = event.nativeEvent.contentOffset.y;
   }, []);
 
-  // Restore the snapshotted offset once the freshly remounted host has laid out
-  // its content. Guarded to fire once per epoch so later content growth (e.g. a
-  // section expanding) can't yank the scroll position. onContentSizeChange fires
-  // on every platform (unlike the iOS-only contentOffset prop).
+  // One-shot per epoch: a repeat restore on later content growth (e.g. a section
+  // expanding) would yank the user's scroll position.
   const handleScrollContentSizeChange = useCallback(() => {
     if (restoredScrollEpochRef.current === presentEpoch) return;
     restoredScrollEpochRef.current = presentEpoch;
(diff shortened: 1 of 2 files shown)


marcodejongh:
Review notes addressed in 4dded3a36: scroll capture throttled to 32ms, the pendingRestoreOffsetRef/restore comments trimmed to the non-obvious constraints, and a test added asserting Advanced collapses after Reset across a sub-picker round trip (`tappableRowDisabled` pre-existed — opacity 0.4, no change needed).

The red `mobile-bundle` check is the repo-wide Expo pin drift (fails on main too) — #3351 bumps the 12 pins; this branch will go green on a rebase/rerun after that merges. The fix itself is dependency-free, so the `pr-3352` OTA channel is unaffected.

marcodejongh:
CI is green across the board, so this is ready for review. The remaining item is on-device iOS confirmation via the `pr-3352` OTA channel (the bug only reproduces against the real SwiftUI sheet host lifecycle) — Android emulator QA of the full flow passed. Latest review-bot notes are all minor/no-action: the Refine and Advanced sections already share the same state pattern, and the scroll-restore fallback would be gold-plating on a best-effort polish feature.

marcodejongh:
@marcodejongh thanks for the on-device check — that disproved the remount theory. Revised root cause in 30ddeabfe: on iOS the SwiftUI host can hand the RN content an unbounded height, so the flex:1 column sizes to its **content** instead of the detent — the footer only shows while the content happens to be shorter than the sheet. That's why it looked fine on first open (sections collapsed), broke after a hold-filter round trip (Refine stays expanded → content taller than the detent), and never showed for the original reporter on an iPhone SE. Android bounds the column natively, which is why the emulator QA and tests couldn't catch it.

The new commit pins the column to a JS-computed 90% detent height on iOS instead of trusting the native bound. Once CI's `publish` job finishes, please pull the update on `pr-3352` again (What's New → check for update) and re-test: filters → Refine → hold picker → back → Apply should stay pinned. If the footer sits a few points above the bottom edge, that's the deliberate conservative margin — tell me and I'll tighten it.

PR #3359, opened 2026-07-02
refactor(web): extract single queue-actions factory shared by board and bridge surfaces

marcodejongh:
## Summary

Web had two parallel implementations of the entire queue-action surface (add/remove/set-current/navigate/mirror/replace/report-wall-disconnect), drifting apart:

- `graphql-queue/QueueContext.tsx` (board routes) — action callbacks built on the `latestRef` stable-callback pattern, next/prev discovery hand-rolled via a search-results neighbor walk.
- `queue-control/queue-bridge-context.tsx` (off-board) — `usePersistentSessionQueueAdapter` action bodies with subtly different logic, nav via the shared `findNextQueueItemWithSuggestions`.

This PR extracts **one implementation** as a pure factory — `createQueueActionsCore(deps)` in `queue-control/queue-actions-core.ts` — and wires both surfaces through it. **Behavior-preserving**: every per-surface difference is expressed as an injected dependency, not silently unified.

### Factory deps shape (`QueueActionsCoreDeps`)

- `getSnapshot()` — fresh `{queue, currentClimbQueueItem, playlistSuggestionSource}` read (backed by each surface's `latestRef`)
- `applyLocal(action)` — how the surface applies an optimistic reducer action:
  - QueueContext: `dispatch` into its `useReducer`
  - Bridge **solo**: runs the shared `@boardsesh/queue` `queueReducer` against the useState-backed local store, then persists via `ps.setLocalQueueState` — the reducer is now the single semantic definition even for the local store (two narrow, test-pinned bridge quirks are layered as documented overrides)
  - Bridge **party**: **no-op** — party mode keeps waiting for the server echo instead of applying optimistic updates (pre-existing behavior, preserved; a later workstream flips it)
- `setPlaylistSuggestionSourceLocal` / `refreshPlaylistSuggestionSourceLocal` — playlist-source bookkeeping against each surface's live store (reducer `SET`/`REFRESH` actions vs functional `setState`)
- `buildItem`, `guardMutation`, `validateClimbForQueue`, `offlineBuffer | null`
- `resolveShouldAddToQueueOnActivate`, `resolveNextCurrentForSetQueue`, `buildReplacementItem` — the three spots where the surfaces' semantics genuinely differ per action
- `discoverNext` / `discoverPrev` — navigation strategy (QueueContext: queue walk + search-results neighbor walk + suggestion fallbacks; bridge: shared `findNextQueueItemWithSuggestions` for mobile parity)
- `coldStart?` — bridge-solo board seeding from a climb's own board config
- `party` — `{isActive, hasConnected, isDisconnected, attemptMutation, reuseExistingQueueItemOnSetCurrentClimb, returnNullOnSetCurrentClimbFailure, mutations, nextCorrelationId}`
- `hooks` — analytics preserved exactly per surface (`Set Active Climb` source tags, `Climb Added/Removed`, queue-metrics timing; bridge omits the QueueContext-only events)
- `errorMessages` — each surface's exact console error strings

### Invariants pinned by the new factory tests (59 tests)

- playlist-peek uuid promotion (`isPlaylistPeekQueueItemUuid` → fresh uuid on activation)
- `insertAfterCurrent` positioning (and append when no current)
- playlist-source rollback on mutation failure (both the plain add+activate branch and the playlist-setQueue branch)
- offline buffering when `isDisconnected && party.isActive` (no server mutation sent)
- split add+setCurrent awaits with FIFO ordering
- setCurrentClimb existing-queue-entry reuse by `climb.uuid` (bridge) vs always-add (QueueContext)
- bridge party mode applies **no** optimistic local update; solo does
- `returnNullOnSetCurrentClimbFailure` split (bridge null vs QueueContext optimistic item)
- guard/validation short-circuits, mirror toggle, replace-in-place, setQueue next-current resolution, `from`-anchor navigation resolution, reportWallDisconnect gates

### Wiring notes

- QueueContext keeps the `latestRef` pattern: the factory deps are all closures over `latestRef.current`, built once (`useMemo(..., [])`), so every action callback keeps `[]`-deps stable identity — the context-split perf model is unchanged (`context-split` tests still green).
- Both wiring files shrank: `QueueContext.tsx` 1201 → 1034 lines, `queue-bridge-context.tsx` 1069 → 961 lines (−275 combined), with the single implementation now in `queue-actions-core.ts`.

### Known micro-deviations (documented)

- Bridge `replaceQueueItem` now validates board compatibility before replacing (previously only the board-route path validated). Practically unreachable — the only caller is the create-climb form on board routes with a just-created same-board climb — and arguably corrective.
- Bridge-solo: re-tapping a current item that is *not* in the queue no longer appends it (the shared reducer's own-tap guard short-circuits first). Near-unreachable state in bridge-solo; the visible current climb is unchanged either way.

### Known micro-deviations (documented, analytics-only)
- Ordering: on the board route, `Set Active Climb` analytics + the climbs-attempted counter now fire just before (not after) the optimistic dispatch, and bridge-party `setCurrentClimbQueueItem` tracks before initiating the mutation. Final state and payloads identical (same React batch; payloads read only arguments).

- Bridge-solo: re-tapping the already-current queue item now emits `Set Active Climb` (aligning with board-route behavior, where it always fired); local state remains unchanged either way.
- Bridge-solo: activating a climb with no board details emits the analytics event before the no-op (previously bailed before the event).

## Release Notes

none

### DIFF
--- a/docs/websocket-implementation.md
+++ b/docs/websocket-implementation.md
@@ -92,6 +92,10 @@ Both `PersistentSessionContext` and `QueueContext` are split into separate **Act
 
 Targeted hooks: `useQueueActions()`, `useQueueData()`, `usePersistentSessionActions()`, `usePersistentSessionState()`.
 
+### Shared queue-actions factory
+
+The queue-action surface itself (add/remove/set-current/navigate/mirror/replace/report-wall-disconnect) has a single implementation: `createQueueActionsCore(deps)` in `packages/web/app/components/queue-control/queue-actions-core.ts`. Both `GraphQLQueueProvider` (board routes) and the bridge's `usePersistentSessionQueueAdapter` (off-board) wire it with their own dependencies — most notably `applyLocal` (how an optimistic reducer action is applied: reducer `dispatch` on board routes; the shared `queueReducer` against the local store in bridge-solo; a **no-op in bridge-party**, which waits for the server echo instead of applying optimistic updates) and `discoverNext`/`discoverPrev` (search-results neighbor walk on board routes vs the shared `findNextQueueItemWithSuggestions` off-board). Every genuine per-surface behavioral difference is an explicit injected dep on `QueueActionsCoreDeps`, documented at the definition.
+
 ### QueueBridgeProvider — root-level QueueContext
 
 `QueueBridgeProvider` (`packages/web/app/components/queue-control/queue-bridge-context.tsx`) is mounted once at the root inside `PersistentSessionWrapper` so a `QueueContext` value is always available, even off board routes (e.g. `/you/logbook`, `/session/[sessionId]`, `/playlists/...`). It has two modes:
@@ -1900,6 +1904,7 @@ A GraphQL mutation would require either the JS GraphQL client (not available in 
 - `packages/web/app/components/graphql-queue/use-queue-session.ts` - Session hook
 - `packages/web/app/components/persistent-session/persistent-session-context.tsx` - Root-level session management (split into ActionsContext + StateContext for render performance)
 - `packages/web/app/components/graphql-queue/QueueContext.tsx` - Queue state context (split into ActionsContext + DataContext; actions use `latestRef` pattern for stable callback identity)
+- `packages/web/app/components/queue-control/queue-actions-core.ts` - Single queue-actions factory (`createQueueActionsCore`) wired by both `QueueContext.tsx` and `queue-bridge-context.tsx`; per-surface differences are injected deps
 
 ### Native iOS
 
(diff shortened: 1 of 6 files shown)


azuttre:
Heads-up: this merge appears to have broken `main`'s typecheck — its own CI run was cancelled (superseded by the next push to main), so it never reported. Since 951da7ce3, every PR fails typecheck via the merge-ref checks with:

`packages/web/app/components/queue-control/queue-bridge-context.tsx:292 — Object literal may only specify known properties, and 'initialQueueDataReceivedFromPeers' does not exist in type 'QueueState'` (repeated at several construction sites).

The field still exists in `packages/shared/queue/src/types.ts:104`, so it looks like the web bridge is constructing a different/narrowed QueueState than the refactor intended. Last green main run was 13dfa9465 (#3361). Flagging rather than fixing since the intended QueueState shape is this refactor's call. (It's currently blocking #3373/#3350 among others.)

PR #3370, opened 2026-07-02
fix(web): drop W1-removed QueueState fields from queue-bridge pseudo state

marcodejongh:
Main's typecheck is red since #3359 merged after #3342: W5's bridge pseudo-`QueueState` still initialises the three fields W1 deleted from the shared type (`initialQueueDataReceivedFromPeers`, `lastReceivedSequence`, `lastReceivedStateHash`). Three-line deletion, no behavior change — the reducer never read them.

Validated on this branch: `vp run typecheck` exit 0 (was failing on clean main), web suite 4926/4926.

Note: #3367 (W4) temporarily carries this same fix as its last commit; it rebases away once this merges.

## Release Notes

none

### DIFF
--- a/packages/web/app/components/queue-control/queue-bridge-context.tsx
+++ b/packages/web/app/components/queue-control/queue-bridge-context.tsx
@@ -289,10 +289,7 @@ function usePersistentSessionQueueAdapter(): {
       playlistSuggestionSource,
       climbSearchParams: DEFAULT_SEARCH_PARAMS,
       hasDoneFirstFetch: false,
-      initialQueueDataReceivedFromPeers: false,
       pendingCurrentClimbUpdates: [],
-      lastReceivedSequence: null,
-      lastReceivedStateHash: null,
       needsResync: false,
     };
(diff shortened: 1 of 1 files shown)


azuttre:
Reviewed, the fix looks right. Those three fields are exactly what W1 moved out of QueueState, and CI is green across typecheck and the web suite, so the bridge behaves without them.

One ask: land this one on its own rather than folding it into #3372. Main's typecheck has been red since the #3359 merge (its CI run was cancelled by the next push, so it merged unreported) and every open PR now fails the merge-ref checks because of it. #3372 can supersede this properly afterwards.

For context on why I care about timing: #3370 heals main, then #3373 (backend fix so hasBetaVideo also matches Aurora-synced beta links, green otherwise) can rebase and merge, and that is what makes the beta video markers work on real data in #3350.

Separate thought on process: #3359 slipping in red suggests auto-cancel on superseded runs can hide a failing merge. I think we should require the merge commit run to complete on pushes to main.

PR #3371, opened 2026-07-02
refactor(mobile): bound sheet columns on iOS, prune legacy sheet props, native sub-picker headers

marcodejongh:
## Summary

Follow-up to #3352: a full pass over the post-gorhom sheet wiring, plus the HIG rework of the three climb-filter sub-pickers. JS + docs only — no native, config, or dependency changes, so everything rides OTA.

**Generalizes the #3352 iOS fix to every sheet.** The root cause we confirmed on-device (iOS can hand the RN content an unbounded height, so `flex:1` columns size to content and anything past the detent lands off-screen) applies to every `Sheet`/`ModalSheet`, not just the climb filters. A shared `useSheetColumnStyle` hook now pins each sheet's column to its active detent height on iOS — footer and footerless paths, multi-detent aware, dynamic-sizing sheets untouched — and `ClimbFilterSheet` uses the same hook.

**Sheet wiring fixes:**
- Android keyboard avoidance actually works now: the wrappers assumed the native sheet resizes for the keyboard — the Compose dialog doesn't. Footers with inputs (beta-video link, comment composer) pad above the keyboard on both platforms (was iOS-only).
- The beta-video paste field + Attach moved into a pinned footer that lifts above the keyboard (verified on emulator — previously the keyboard covered the field while typing).
- Invite sheet opens at ~60% on Android instead of full-screen (`androidSafeSnapPoints`).
- 7 sheets migrated off the legacy `isPresentedRef` + deprecated `onDismiss` pattern to the controlled `visible` prop (climb actions, playlist actions/form, BLE control, 3 accessibility pickers). The deprecated `onDismiss` prop is gone — it fired on coordinator-driven closes too, silently clearing parent state.
- Logbook filter edits also commit when the sheet unmounts, not just on pan-down (guarded against double-commit).
- Removed the inert legacy gorhom props (`glass`, `stackBehavior`, `keyboardBehavior`, `keyboardBlurBehavior`, `android_keyboardInputMode`, `fullWindowOverlay`) from the wrappers and every call site. `fullWindowOverlay` deserves a note: it was silently ignored, and restoring it can't work — native sheets present off the key window *beneath* any `FullWindowOverlay` (the PlayDrawer migration comment records this scar). `CreateDrawer`/`HoldRoleSheet` stack as sibling native hosts, which is the actual mechanism; docs updated to match reality.

**HIG rework of the filter sub-pickers** (holds / board region / setters): native stack headers with back chevron + title, "Clear all" as a headerRight (shown only when there's something to clear), hand-rolled in-body headers and "Done" buttons deleted (selection still hands off on blur, so back ≡ done). Setters page now themes through `systemColors` (was static iOS hexes), gets a bottom safe-area inset and a labeled search field.

`docs/mobile-sheets-vs-routes.md` reconciled with all of the above.

## Release Notes

- The hold, area, and setter filter pickers get proper navigation headers with a back button.
- Pasting a beta-video link no longer hides the text field behind the keyboard.
- The invite sheet opens at its intended height on Android.

## Test plan

- `vp run typecheck:mobile`, `vp run test:mobile` (389 files / 3008 tests, +2 wrapper-behavior tests, migrated sheet tests rewritten), `vp run check:mobile-bundle`, `vp check` — all green.
- Android emulator, end-to-end: all three sub-pickers (native header, headerRight Clear-all, selection handoff round trips, Apply pinned & committing), beta-video footer above the keyboard while typing (before/after verified), long-press flows, filter Reset/Apply regression sweep of #3330.
- **iOS OTA QA requested** (`pr-3371` channel) — the iOS-only surfaces: every sheet's detent-height bound (log ascent, device picker, comment composer, you-filters — footers should sit at the sheet bottom, tall footerless sheets should scroll, none clipped), the create-flow long-press hold-role picker presenting over the drawer (pre-existing behavior, unchanged by this PR, but worth confirming), and the sub-pickers' opaque native headers.

https://claude.ai/code/session_01Vegna6LURUUfTrynUmLK5N

### DIFF
--- a/docs/mobile-sheets-vs-routes.md
+++ b/docs/mobile-sheets-vs-routes.md
@@ -19,19 +19,29 @@ Most things are sheets. Routes are the exception, and each route variant earns i
 Wrapped by two helpers so they don't drift (both supply the scrim, drag handle, iOS 26 glass
 background, and keyboard avoidance natively):
 
-| Wrapper                                            | Backing            | Opened by                                         | Use when                                                                                                                          |
-| -------------------------------------------------- | ------------------ | ------------------------------------------------- | --------------------------------------------------------------------------------------------------------------------------------- |
-| **`ModalSheet`** (`src/components/ModalSheet.tsx`) | `BottomSheetModal` | imperatively via ref — `present()` / `dismiss()`  | A button/handler opens it. Presents **above root chrome** (the queue bar / tab bar) for free. The default for an on-demand sheet. |
-| **`Sheet`** (`src/components/Sheet.tsx`)           | `BottomSheet`      | declaratively — its presence in the tree shows it | Its lifetime is tied to parent state. Has a **`fullWindowOverlay`** prop to float above other sheets.                             |
+| Wrapper                                            | Backing            | Opened by                                                                          | Use when                                                                                                                          |
+| -------------------------------------------------- | ------------------ | ---------------------------------------------------------------------------------- | --------------------------------------------------------------------------------------------------------------------------------- |
+| **`ModalSheet`** (`src/components/ModalSheet.tsx`) | `BottomSheetModal` | imperatively via ref — `present()` / `dismiss()`, or the controlled `visible` prop | A button/handler opens it. Presents **above root chrome** (the queue bar / tab bar) for free. The default for an on-demand sheet. |
+| **`Sheet`** (`src/components/Sheet.tsx`)           | `BottomSheet`      | declaratively — its presence in the tree shows it                                  | Its lifetime is tied to parent state.                                                                                             |
 
 Examples: `QueueSheet`, `BoardSheet`, `LogAscentSheet`, `AngleSelectorSheet`,
 `ClimbActionsSheet`, `AddBetaVideoSheet` (sheet surfaces), `CreateDrawer` + `HoldRoleSheet`
 (declarative `Sheet`s).
 
-**Stacking a sheet above another sheet / above everything:** add `fullWindowOverlay` to a
-`Sheet` (renders it in a higher window), or use a `FullWindowOverlay` directly. This is how
-`HoldRoleSheet` floats above the create drawer and how `ClimbReactionMenu` (the long-press
-context menu) floats above whatever's underneath.
+Prefer the controlled `visible` prop over an imperative ref + a local `isPresentedRef`: the
+coordinator reconciles present/dismiss from `visible`, and `onClose` fires only on a genuine user
+pan-down / backdrop (not on coordinator-driven closes), so parent state can't be cleared behind
+your back. `AddToPlaylistSheet` is the reference.
+
+**Stacking above everything:** a **custom, non-sheet** overlay (Reanimated + plain views) can
+render in a higher iOS window via react-native-screens' `FullWindowOverlay` — that's how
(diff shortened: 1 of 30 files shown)


marcodejongh:
The red `typecheck`/`ci-status` is inherited, not from this PR: main's own CI at `951da7ce3` (this branch's base) fails web typecheck in `app/components/queue-control/queue-bridge-context.tsx:292` (`initialQueueDataReceivedFromPeers` doesn't exist in `QueueState` — from the queue sync-gate work). This PR touches mobile + docs only. I'll rebase once main is green again.

marcodejongh:
On the three latest review notes — no further code churn planned:
- The drag-indicator `iosSystemColors.separator` sits in a module-scope `StyleSheet.create` where the theme hook can't reach; it predates this PR and matches the design-guidelines allowance for static styles outside the provider.
- `useSheetColumnStyle`'s branches (px detents, index clamp, non-% fallthrough, dynamic-sizing short-circuit) are exercised through the wrapper tests on the paths the app actually uses; direct hook tests are a fine follow-up if the hook grows.
- Agreed on watching Android keyboard-height reporting on third-party keyboards post-OTA — will keep an eye on it once this rolls out.

Remaining reds are environmental: `typecheck` is inherited from main's queue-bridge web error (rebase once main is green), and one `build-and-test` was a macOS runner disk flake (re-run in progress).

marcodejongh:
Rebased onto latest green main (`677bea388`) — the inherited web-typecheck red is gone. Ran the **full aggregate** `vp run typecheck` locally this time (web `tsc --noEmit` + build + all projects): clean. Mobile suite 390 files / 3019 tests green.

Also addressed the last review notes in `9c90b0f92`:
- Themed the filter-sheet drag indicator to `systemColors.separator` (adapts light/dark) and removed the footer's dead static `borderTopColor`. The two remaining `iosSystemColors` uses in that file are legitimate: a fixed-contrast white chip label on a brand fill, and a mode-invariant neutral card gray (`#8E8E93`, identical light/dark — swapping to the theme `fill` token would tint it violet).
- Added direct `useSheetColumnStyle` unit tests (8): px + % detent math, activeIndex clamp over/under, non-% and dynamic-sizing fall-through, and the Android flex path.

marcodejongh:
Thanks — walked through all four; none need a code change, and I'd rather not churn a green PR for speculative hardening:

1. **committedRef reset** — `LogbookFilterSheet` is mount-gated (`{filterSheetOpen ? <LogbookFilterSheet…/> : null}` in LogbookTab), so every open is a fresh mount with `committedRef=false`; close unmounts. There's no reopen-without-remount path, so a reset effect would be dead code keyed on a signal the component doesn't receive. The guard's only job — dedupe the pan-down `onClose` commit against the unmount-cleanup commit within one lifecycle — works as intended.
2. **headerRight cleanup** — pushed stack screens remount on navigation; the header belongs to the screen instance, so it unmounts with it. No shared header to leak into.
3. **pasteFooter memo** — as the note says, it must be fresh (captures url/showError/submitDisabled), so a `useMemo` with those deps rebuilds exactly when it already does. No reconciliation saved.
4. **AddBetaVideoSheet footer test** — fair gap, but the behavior is already asserted generically in `sheet.test.tsx` (footer → KAV `behavior="padding"`); a per-component duplicate adds little. Happy to add it as a follow-up if you'd like it explicit.

PR is green and rebased on today's main; ready when you are.

PR #3372, opened 2026-07-02
refactor(web): single root-owned queue state for board and bridge surfaces

marcodejongh:
## Summary

**Stacked on #3370** (`fix/web-bridge-pseudo-state-w1-fields`) — auto-retargets to `main` when that merges. Workstream W6 of the multi-user-sessions plan.

Queue state (`queue`, `currentClimbQueueItem`, `playlistSuggestionSource`, `pendingCurrentClimbUpdates`) used to live in THREE places on web: the board-route reducer in `GraphQLQueueProvider`, the off-board bridge's solo `useState` store, and the root persistent-session reducer (party mirror). The root reducer (`persistent-session/hooks/use-event-processor.ts`) is now the ONLY owner. Both surfaces read it and dispatch local/optimistic actions through the newly-exposed root `dispatch`.

- **`graphql-queue/QueueContext.tsx`**: drops `useQueueReducer`; queue slices come from the root; `applyLocal` = root dispatch (through a documented type-seam cast). Route-scoped search concerns (`climbSearchParams`, `hasDoneFirstFetch`) stay board-route-local plain `useState` — deliberately NOT in the shared root (mobile shares the root reducer shape and has no use for them).
- **`queue-control/queue-bridge-context.tsx`**: adapter drops the solo useState store and the injected-mode queue copy-back; `syncFromInjected`/`clear()` shrink to board-context bookkeeping; `playlistSuggestionSource` is reducer-owned (the "remember to clear on full-queue replace" footgun comment is gone — `INITIAL_QUEUE_DATA`/`UPDATE_QUEUE` clear it for free).
- **Deleted** `graphql-queue/hooks/use-queue-restoration.ts` (vestigial `_hasRestored`; its board-config-change clear moved into `setBoardContext`) and `use-queue-event-subscription.ts` (no second reducer copy left to mirror events into).
- **Moved** `use-pending-update-cleanup.ts` → `persistent-session/hooks/`, invoked at the root (pending correlation ids live in root state now).
- **`use-queue-storage.ts`**: drops `localQueue`/`localCurrentClimbQueueItem`; renames `localBoardPath`/`localBoardDetails`/`isLocalQueueLoaded` → `soloBoardPath`/`soloBoardDetails`/`isBoardContextLoaded` (board-context-only scope); `setLocalQueueState`/`clearLocalQueue` → `setBoardContext(boardPath, boardDetails)`.

## Behavior changes

- **Off-board party mutations are now optimistic** (the flagged, approved change). The bridge's party `applyLocal` used to no-op and wait for the server echo; it now dispatches to the root reducer immediately, exactly like board routes always have, with the real party mutation still following. Correlation-id echo suppression (`pendingCurrentClimbUpdates` + root-level `usePendingUpdateCleanup`) covers the echo, so the server confirmation lands without a flicker.
- **Peer-broadcast queue analytics gating**: the `Climb Added to Queue`/`Climb Removed from Queue` (`addedFromTab: 'peer_broadcast'`) events moved from the deleted board-route hook to a root-level `use-peer-broadcast-analytics.ts`, explicitly gated on `isBoardRoutePath(pathname)`. Decision rationale: the old hook only ever ran inside board-route-scoped `GraphQLQueueProvider`, so off-board never fired these events — the explicit gate restores exactly that parity now that the subscriber lives in the always-mounted root. Self-echoes still count (the old hook never filtered self vs peer for add/remove either; only `CurrentClimbChanged` carries a clientId).
- **Solo-only bridge quirks kept solo-only**: "adding to an un-activated queue also activates the item" and "removing the current item promotes the new head" still apply in solo but were NOT extended to party — party state has always reflected the server's echo, which has no such auto-activate/promote semantics; extending them would fabricate a local current climb the mutation never requested.
- One review-found parity fix rolled in: a malformed `QueueItemAdded` (no item payload) no longer counts in analytics — the old hook gated analytics on the reducer dispatch actually happening; the relocated hook now applies the same guard.

## Solo persistence

Unchanged (verified against W3): the solo queue is root-owned and in-memory only. It survives SPA navigation because the root provider mounts once and never unmounts across route changes; it does NOT survive a full page reload — identical to behavior before this PR (queue IndexedDB persistence was deliberately removed in `de5f8c9b2`, long before this workstream; "start a session" remains the reload-persistence path). Only `ActiveSessionInfo` (the party-session pointer, IndexedDB key `activeSession` in `boardsesh-user-preferences`) persists — same key, same shape, untouched.

## SharedRefs decision

Spec offered replacing `queueRef`/`currentClimbQueueItemRef` with a single `stateRef`. Kept both refs (minimal churn): they're already sourced from the one root reducer via two mirror effects, and only `use-session-lifecycle.ts` (state-hash computation on reconnect) and the new analytics hook read them.

## Verification invariants

- SPA nav board→home→board keeps the queue: root reducer never unmounts; pinned by `queue-bridge-context.test.tsx` (adapter reads root state) + existing e2e `queue-persistence.spec.ts`.
- Solo queue survives route changes: same mechanism; board-config-change clear only fires when the base board path actually changes (angle changes are stripped by `getBaseBoardPath` before comparison).
- Party peer adds land exactly once: single root subscription applies the event once; board routes no longer re-apply via a second reducer (pinned in `use-peer-broadcast-analytics.test.ts` + `event-processor-offline.test.ts`).
- Widget-navigation echo suppression (`dispatchWidgetNavigation`): unchanged mechanism, now against root state — the correlation id registers in root `pendingCurrentClimbUpdates` and the echo is recognized there (new test: "suppresses the server echo of a locally-dispatched current-climb update via correlation id").
- LiveActivityBridge sees the live queue on AND off board routes: injected mode passes the board context through unchanged; adapter mode now reads root state directly (previously the solo copy).

## Release Notes

- Party mode: picking a climb from your logbook or a playlist while away from the board view now updates the shared queue instantly — no more waiting on the round trip.

## Test plan

- `vp check --fix` — clean (0 errors)
- `vp run typecheck` — full monorepo, clean
- `vp run test:web` × 2 — 371 files / 4935 tests, all green both runs (no flake re-runs needed)
- `vp test run --reporter=agent` (full monorepo) — 1043 files / 11861 tests, all green
- New tests: off-board party optimistic apply + echo suppression via correlation id; peer add landing exactly once with board-route gating; root-dispatch semantics; `queue-actions-core` `invocationCallOrder` contract (analytics → applyLocal → mutation-initiation); board-context clear on board switch
- Test migration: tests that mocked `usePersistentSession*` and mounted `GraphQLQueueProvider`/`QueueBridgeProvider` now wrap with a new shared `MockRootQueueProvider` (`app/test-utils/mock-persistent-session-queue.tsx`) — a real `@boardsesh/queue` reducer behind the mocked context, so dispatched actions re-render consumers exactly like production


marcodejongh:
Went through the four review points. Summary: **item 1 fixed, item 3 tracked, items 2 & 4 are false positives** (applying them as written would each introduce a regression — details below).

### ✅ Item 1 — `queueRef` off-by-one in analytics (fixed: 7e373666a)

Correct catch. The refactor moved the ref assignment into a `useEffect`, which commits a render after the event fires, so `currentQueueLength` (and the `session-connection-ports` hash getter) read the pre-event value. Moved both `queueRef` **and** `currentClimbQueueItemRef` back to the render body (the pattern the deleted `use-queue-event-subscription` hook used). Moved both together so the hash getter — which reads them as a pair — never mixes a fresh queue with a stale current-climb. Tests + typecheck green.

### 📋 Item 3 — `removedBy: 'peer'` mislabels self-removes (tracked: #3382)

Real attribution gap: `QueueItemRemoved` carries no `clientId`, so self-remove echoes can't be told from peer removes. Filed **#3382** with the fix direction (add `clientId`/`correlationId` to the wire event, same mechanism `CurrentClimbChanged` uses) and flagged the open question of whether this is new here or pre-existing (the hook's own comment claims the old hook also saw self-echoes; worth confirming). Not blocking this PR.

### ⚠️ Item 2 — double-dispatch of `playlistSuggestionSource` — **not applied**

The pre-`DELTA_UPDATE_CURRENT_CLIMB` `setPlaylistSuggestionSourceLocal(...)` call is **not** redundant. Per the interface doc at `queue-actions-core.ts:175-188`, `applyLocal` is a **no-op in bridge-party mode** ("party mode waits for the server echo instead of applying an optimistic update"), and `playlistSuggestionSource` is client-local bookkeeping the server never echoes back. So in bridge-party mode that explicit call is the **only** thing that sets it — removing it drops playlist suggestions whenever a party mutation is in flight. The double-set only happens on the QueueContext / bridge-solo surfaces (where `applyLocal` applies); the extra render there is the deliberate cost of a uniform unconditional set across all three surfaces. Left as-is.

### ⚠️ Item 4 — `inject` fires on every angle change — **not applied**

Removing `angle` from the `QueueBridgeInjector` layout-effect deps would introduce a stale-angle bug. That `inject` also calls `setInjectedAngle(a)` (`queue-bridge-context.tsx:749`), which feeds `effectiveAngle` (`:705`, `isInjected ? injectedAngle : adapter.angle`). The other effect only runs `updateContext` on an angle change and explicitly does **not** touch board identity (its comment at `:1001-1005`), so dropping `angle` from the layout effect would leave `effectiveAngle` stale after an angle change — and angle drives LEDs/grades. The redundant `setSoloBoardPath`/`setSoloBoardDetails` writes the comment targets are already functionally harmless (React bails out on identical `useState` values). If we want to eliminate them anyway, the safe fix is to guard the `adapterSync`/`setBoardContext` call on an actual board-path change while keeping `setInjectedAngle` on every angle change — happy to do that as a follow-up if it's worth it. Left as-is for now.

PR #3377, opened 2026-07-03
feat(backend): grouped ascents feed — shared filters + full item enrichment

azuttre:
## Summary

Backend half of the logbook follow-up agreed in #3350 (day-scoped grouping of repeat ascents + attempts-by-default):

- `userGroupedAscentsFeed` accepts the full `AscentFeedInput`. Filter semantics are extracted into shared builders (`buildAscentTickConditions` / `buildAscentClimbConditions`) used by BOTH the flat and grouped feeds, so a filter can never mean two things. Filters apply to the group page, the total count, and the tick fetch — group aggregates (flash/send/attempt counts, bestQuality) reflect exactly the visible entries.
- Group items are enriched to the full `AscentFeedItem` shape: board identity (with the same visibility gating as the flat feed), consensus grade pair, `qualityAverage`, and `hasBetaVideo` via the shared ownership helper extracted from #3373. The SDL already typed items as `AscentFeedItem`, so this is resolver-only.
- Group-level SQL pagination is preserved (the property the resolver was built for).
- Existing consumer (web activity feed, no input) is unaffected: no filters → identical behaviour.

**Ordering:** deploy before the client PR that consumes the new fields/filters (per the backend-first rollout rule — OTA bundles hit prod).

## Release Notes

- [x] No release note needed (internal / technical change — the user-facing grouping ships with the client PR)

## Test plan

- 4 new integration tests: statusMode filters groups AND aggregates; burns+send stay one group under `both`; climbName respects group pagination counts; `hasBetaVideo` ownership on group items.
- Full tick-queries suite green locally against the shared test DB (42 tests).

### DIFF
--- a/packages/backend/src/__tests__/tick-queries.test.ts
+++ b/packages/backend/src/__tests__/tick-queries.test.ts
@@ -1046,6 +1046,113 @@ describe('tickQueries — behavior fixes', () => {
     });
   });
 
+  describe('userGroupedAscentsFeed — filters & enrichment', () => {
+    const seedProjectDay = async () => {
+      const sentUuid = `${CLIMB_PREFIX}grouped-sent`;
+      const projectUuid = `${CLIMB_PREFIX}grouped-project`;
+      await insertClimb(sentUuid, 'Grouped Sent Climb');
+      await insertClimb(projectUuid, 'Grouped Project Climb');
+      // Same climb, same day: two burns then the send — one group of three.
+      await insertTick({
+        uuid: 'grp-a1',
+        climbUuid: sentUuid,
+        climbedAt: '2026-06-20T10:00:00',
+        status: 'attempt',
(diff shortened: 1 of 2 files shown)


azuttre:
All four addressed in the latest push:

1. **or() invariant** — documented: `createdByUserId` is unconditional, so the conditions array is structurally never empty.
2. **Type duplication** — `AscentFeedFilterInput` is now `z.infer<typeof AscentFeedInputSchema>`; validation and filtering can't diverge.
3. **Board visibility coverage** — new test: owner ctx sees the private board's name on group items, anonymous caller gets null.
4. **JOIN duplication** — refactored to a single `filtered_groups` subquery; the page query and the count both read from it, so a join change lands once.

43 integration tests green against the shared local DB.

marcodejongh (approved):

PR #3380, opened 2026-07-03
fix(sync): make Kilter catalog sync only touch Kilter-synced climbs

marcodejongh:
## Summary

The Kilter catalog sync daemon was logging a mix of one real bug and several benign diagnostics that read as errors (e.g. `deletions: applied:false`, duplicate `skipped[]` entries, an opaque `climbsUnmapped: 24`). This triages each, fixes the real bug, and turns the report-only deletion pass into a safe, batched auto-apply.

**The hard rule:** the Kilter sync may only ever delete or modify climbs that came from Kilter's catalog sync — **never user-authored climbs**. Provenance is enforced with `board_climbs.user_id IS NULL` (synced, not user-created) and `board_climb_aliases.source = 'kilter'`, at both the classification layer and the final SQL `WHERE` clauses.

What changed:

- **Dedup the PowerSync reference pull** (`reference-pull.ts`) — key each entity by its natural id (last-write-wins). A `wall_uuid`/`gym_uuid` streamed more than once no longer inflates `boardsSeen` or stacks duplicate rows in the locations skip report. Also dedupes the merged `skipped[]` array defensively.
- **Provenance-guarded, batched, reversible deletions** (`deletions.ts`):
  - Soft-delete (`is_listed = false`) a canonical only when `user_id IS NULL`, it's still listed, and it's a lone self-canonical. User-authored climbs are counted `protectedUserAuthored` and never touched.
  - Hard-delete only `source = 'kilter'` pure-alias rows (canonical always survives). `backfill`/`aurora` aliases are left alone.
  - Both the `UPDATE` and `DELETE` carry the same guards (belt-and-suspenders vs the classifier).
  - Applies at most `deleteBatchLimit` changes/run (default 500); the backlog drains over cycles via the `is_listed` guard. An anomaly guard refuses an implausibly large soft-delete batch (wipe signal) for manual review.
- **Ingest guard** (`catalog-sync.ts`) — the climb upsert gets `setWhere(user_id IS NULL)` so a re-sync can't overwrite a user-authored row on a UUID collision.
- **Observability** — a sample of unmapped climbs (uuid + offending `holeId`) so `climbsUnmapped` is diagnosable, and the deletions log now labels `unknown` as never-imported (expected).
- **Daemon opts in** by default (`SyncRunnerConfig.applyCatalogDeletions`, reversible + scoped); the CLI stays report-only unless `--apply-deletions`.

Left as expected (deduped, quieter) skips, not fixed here per scope: unmapped product layout 33 / unsupported wall config 30.

## Release Notes

## Test plan

- `vp test run --project kilter-sync` — 127 tests pass, including new coverage: reference dedup (last-write-wins), skip dedup, deletion classification (user-authored protected, foreign-source alias never dropped, drain convergence), batch cap, and the soft-delete-only anomaly guard.
- `vp check packages/kilter-sync/src` — lint + format + types clean.
- `tsc --noEmit` on the package — clean.
- A dedicated QA/reviewer pass audited every delete/modify path against the "Kilter-synced only" requirement and confirmed it is upheld at both the classification and SQL-WHERE layers.

### DIFF
--- a/packages/kilter-sync/src/cli/index.ts
+++ b/packages/kilter-sync/src/cli/index.ts
@@ -94,6 +94,7 @@ program
   .option('--user <userId>', 'use this linked user’s stored Kilter credential (refresh grant)')
   .option('--layouts <uuids>', 'comma-separated product_layout_uuids to sync (default: all listed)')
   .option('--apply-deletions', 'apply /delteduuids reconciliation (default: report only)')
+  .option('--delete-batch-limit <n>', 'max deletion changes to apply per run (default: 500)')
   .option('--suppress-notifications', 'skip setter-follow notifications (use for the first bulk ingest)')
   .option('-v, --verbose', 'verbose per-layout logging')
   .action(
@@ -101,6 +102,7 @@ program
       user?: string;
       layouts?: string;
       applyDeletions?: boolean;
+      deleteBatchLimit?: string;
       suppressNotifications?: boolean;
       verbose?: boolean;
     }) => {
@@ -123,8 +125,15 @@ program
           return;
         }
         const layoutUuids = parseLayoutUuids(opts.layouts);
+        const deleteBatchLimit = opts.deleteBatchLimit != null ? Number(opts.deleteBatchLimit) : undefined;
+        if (deleteBatchLimit != null && (!Number.isInteger(deleteBatchLimit) || deleteBatchLimit <= 0)) {
+          console.error('--delete-batch-limit must be a positive integer');
+          process.exitCode = 1;
+          return;
+        }
         const summary = await runner.runCatalogSync(tokenProvider, {
           applyDeletions: opts.applyDeletions,
+          deleteBatchLimit,
           layoutUuids,
           suppressNotifications: opts.suppressNotifications,
         });
--- a/packages/kilter-sync/src/runner/sync-runner.ts
+++ b/packages/kilter-sync/src/runner/sync-runner.ts
@@ -294,7 +294,12 @@ export class SyncRunner {
    */
   async runCatalogSync(
     tokenProvider: KilterTokenProvider,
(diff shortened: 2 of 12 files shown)


marcodejongh:
Thanks — addressed in a9e1c62.

**Critical — non-transactional write:** Fixed. Both the alias `DELETE` and the soft-delete `UPDATE` now run inside `db.transaction(...)`. (Worth noting the described "permanent orphaning" couldn't actually occur even before this: `aliasUuidsToDelete` only ever holds *pure* aliases, never self-aliases, so a canonical always keeps its self-alias and is re-classified next cycle — the drain is idempotent. The transaction is the right hardening regardless.)

**Critical — case-sensitive self-canonical check:** Fixed. The `aliasUuid === canonicalUuid` comparison is now `.toLowerCase()` on both, matching the `lower(alias_uuid)` lookup.

**Moderate — `deleteBatchLimit` not on CLI:** Fixed. Threaded through `runCatalogSync` and added a `--delete-batch-limit <n>` flag (validated positive integer).

**Moderate — anomaly guard gates the whole backlog:** Kept by design, and I added a comment explaining why. Gating on the per-run *batch* would let a genuine mass-delete drain 500/run and silently unlist the entire catalog without ever tripping — the opposite of what the guard is for. It must see the aggregate to recognize a wipe. On the real data the soft-delete backlog is ~5.2k vs ~380k live climbs (1.4%), so it never trips; a legitimately large drop is exactly the case a human should review (then bump `ANOMALY_FRACTION` for a one-off manual apply). Also refined: the guard now keys on soft-deletes only, so a benign backlog of pure alias-drops can't trip it.

**Minor — `firstUnplaceableHole` tests:** Added (missing holes across frames, all-placeable → null, empty concat → null).

**Minor — dedup separator:** Switched to `\n` (can't appear in a `kilter:gym:wall` sourceKey or a reason string).

PR #3384, opened 2026-07-03
feat(mobile): logbook repeat grouping, attempts by default, lifetime per angle

azuttre:
## Summary

Client half of the follow-up agreed on #3350 (backend: #3377, deployed):

- **Day-scoped repeat grouping.** In date-ordered views, burns and the send of the same climb on the same day collapse into one row: the best outcome carries the glyph/grade/stars/note/beta markers, and the tries sum for the day (`Flash · 5 tries` included). Groups re-bucket client-side by **(climb, local day, angle)** — honest across timezones and adjustable-board angle changes. Grade sorts and date-ASC keep the flat list.
- **Day-entries chooser.** Swipe-edit/delete on a multi-entry row opens a sheet to pick which tick; there is deliberately no act-on-all. Single-entry rows behave exactly as before.
- **Attempts show by default** (web + mobile), with a divergence-respecting migration: payloads still on either historical resting default refresh to the new one; anything the user changed round-trips untouched.
- **Lifetime per angle** on the climb's play-drawer logbook: `40° — 13 tries over 3 sessions · 2 sends`, computed client-side (sessions = distinct days, matching the grouping convention).
- Cache correctness: grouped-feed invalidation on save/update/delete, optimistic strip shrinks/drops groups with page-count lockstep.

Pre-PR adversarial review (27-agent workflow) confirmed and fixed 11 defects before this was pushed — including a UTC/local day split, both prefs migrations stranding legacy users, and flash rows hiding summed tries. Two semantics are intentional and documented: day-rollup counts climbs (matching the collapsed view), and filtered views sum only visible entries' tries.

## Release Notes

- Working a climb all session no longer floods your logbook — burns and the send collapse into one row with the day's total tries. Tap-and-hold or swipe still reaches every individual entry.
- Your projects now show alongside your sends by default. Flip back to sends-only anytime with the Show filter — your choice sticks.
- Open any climb you've worked and see your history per angle: total tries, sessions, and sends at 40° vs 45°.

## Test plan

- 3,047 mobile tests, shared suites, full typecheck, Metro bundles both platforms — green.
- New contract tests: grouped-row rendering (best outcome + summed tries), chooser routing for edit/delete, single-entry passthrough, grouped cache strip (shrink + drop + count lockstep), migration matrix (v1/v2 × resting/diverged, both platforms), angle-lifetime derivation.

## Post-review additions (device feedback + review rounds)

- **Sequential deletes in the day-entries chooser**: the sheet stays open through each delete, pruning the removed entry (down to one) and closing when the group empties. Status glyphs switched to the logbook's standard bolt / filled tick / orange X (the outline circles read as radio buttons).
- **Play-drawer lifetime sections order steepest angle first**; recap strings are template-localized.
- **Mirrored ascents get their own day group** (re-bucket key includes isMirror).
- **Edits write through to cached feeds** (status/tries/grade/name patched from the mutation response before the refetch).
- **Behavior change to pre-existing code, for traceability**: BOTH feed paginators (flat `useUserAscentsFeed` included) now derive the next offset from pageParam arithmetic instead of summing cached page item counts — the optimistic delete-strip shrinks cached pages, and a count-derived offset re-requested rows the list already had.
- **`logbook-grouping-kill` PostHog flag** (absent = grouping on) as an instant server-side rollback for the grouping behavior.


azuttre:
Manually tested on device (iPhone, iOS 26 Liquid Glass) via the `pr-3384` OTA channel, two full passes:

- Repeat grouping: a burned-then-sent day collapses to one row with the day's tries summed, best outcome carrying the row. Same climb at two angles stays two rows.
- Day-entries chooser: swipe edit/delete on a grouped row opens the picker, sequential deletes keep it open until the day is empty, status icons read correctly.
- Attempts show by default and my saved filter prefs survived the migration.
- Search by name and the filter chips work on the Liquid Glass toolbar.
- Lifetime per angle in the climb view: steepest angle first with tries/sessions/sends recap.
- Edited a tick's grade and the row plus day divider updated instantly.
- Party session history and crew views unaffected.

All behaved as expected. Good to go from my side.

marcodejongh:
Looks great now

PR #3388, opened 2026-07-03
chore(mobile): exact-pin @expo/ui and guard Expo dep pinning

marcodejongh:
## Summary

Forward-looking follow-up to #3387 (the JS-only OTA hotfix for the `ModalBottomSheetView.partialExpand` "No handler registered" crash). This removes the drift vector that caused it.

`@expo/ui` was the **only** Expo native dep left on a tilde range (`~56.0.18`) — every other `expo` / `@expo/*` dep is exact-pinned. An Expo native module ships a JS + native pair; a `~`/`^` range lets a later non-frozen install float the **JS** ahead of the version a shipped binary's **native** layer was built against. Normally the OTA runtimeVersion fingerprint would refuse the mismatched bundle, but this app pins the fingerprint via `EXPO_UPDATES_FINGERPRINT_OVERRIDE` (`app.config.ts`), so the drifted JS still lands on the old binary — exactly the `partialExpand()` call reaching a native view that never registered that `AsyncFunction`.

Changes:
- **Pin `@expo/ui` → `56.0.18`.** The resolved version is unchanged (`bun.lock` already resolved to `56.0.18`, same integrity hash), so this does **not** move the OTA fingerprint — verified equal iOS/Android fingerprints with and without the pin locally. `56.0.18` native registers `partialExpand` (`ExpoUIModule.kt` / `ModalBottomSheetView.kt`), so JS and native agree on the next build.
- **Add `scripts/mobile-expo-deps-pinned-check.ts` + test:** every `@expo/*` and `expo-*` dep must be exact-pinned (non-Expo RN deps that intentionally use `~`, e.g. `react-native-gesture-handler`, are exempt). Wired into `vite` (`check:mobile-expo-deps-pinned`) and CI so this class can't recur.

Also carries the same orphaned-i18n-key removal as #3387 (`climbs:mobile.filter.done`) so this branch's CI is green independently; whichever PR merges first, the other's rebase drops the duplicate.

## Release Notes

## Test plan

- `vp run check:mobile-expo-deps-pinned` — OK (all 35 Expo native deps exact-pinned); goes red on `~56.0.18`, green after the pin.
- `vp test run scripts/__tests__/mobile-expo-deps-pinned-check.test.ts` — 7 pass (tilde flagged, exact set passes, exact prerelease, every range form flagged, non-Expo `~` ignored).
- `vp run check:mobile-ota-compat` — iOS/Android fingerprints identical with and without the pin (resolved version unchanged), confirming this doesn't strand installed binaries.
- `bun install` leaves the resolved tree unchanged (only the recorded spec string).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -535,6 +535,13 @@ jobs:
       # push-to-main. Runs before the heavier Metro bundle so it fails fast.
       - name: Check native build-phase deps resolve
         run: vp run check:mobile-native-deps
+      # Every @expo/* and expo-* dep must be exact-pinned: they ship a JS+native
+      # pair, and a `~`/`^` range lets the JS float ahead of a shipped binary's
+      # native layer. With the OTA fingerprint pinned via an override, that
+      # drifted JS still lands on the old binary — the ModalBottomSheetView
+      # .partialExpand "No handler registered" crash from a `~56.0.18` @expo/ui.
+      - name: Check Expo deps are exact-pinned
+        run: vp run check:mobile-expo-deps-pinned
       # Deterministic, read-only dep-health check: compares our declared pins
       # against packages/mobile/node_modules/expo/bundledNativeModules.json for
       # the SDK release we actually have installed, and flags installed-version
--- a/bun.lock
+++ b/bun.lock
@@ -225,7 +225,7 @@
         "@boardsesh/queue-react": "*",
         "@boardsesh/queue-runtime": "*",
         "@boardsesh/shared-schema": "*",
-        "@expo/ui": "~56.0.18",
+        "@expo/ui": "56.0.18",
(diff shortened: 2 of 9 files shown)


marcodejongh:
Addressed the review note: renamed `isExpoNativeDep` → `isExpoPackage` and updated the doc to make the scope explicit — the check intentionally covers the **whole** `@expo/*` scope (native and pure-JS tooling), because the repo already exact-pins all of them and stricter is safer (a ranged pure-JS Expo tool is a reproducibility hazard; a ranged native module is the JS/native drift hazard that caused the crash). Added a note that a genuinely range-safe pure-JS `@expo/*` dev tool should get an explicit allowlist rather than loosening the rule. Left `main()` uncovered — it's thin filesystem wiring; the logic lives in the tested pure functions.

PR #3395, opened 2026-07-03
fix(mobile): resolve every onboarding start to a terminal outcome

marcodejongh:
## Summary

A PostHog dashboard funnel made it look like **everyone skips onboarding** (1,239 started → 0 completed). Investigation showed that's a **broken funnel, not real behaviour**: the funnel's middle step is `Onboarding Tour Step Advanced`, an event the redesigned single-screen mobile onboarding (June ~19) no longer fires, so an RN-filtered `Started → Step Advanced → Completed` funnel can only ever report 0%. Real 14-day numbers: **~48% complete, ~23% skip**.

While verifying, two genuine telemetry gaps surfaced on mobile — this PR closes them (the misleading saved insight itself is corrected separately in PostHog config, no code):

- **~29% of Starts had no terminal outcome.** The first-run prompt fires `Started` + one `Step Viewed` on mount, then `Completed` (primary CTA) or `Skipped` (secondary "look around") — but leaving via Android hardware-back or a nav-away fired *neither*, silently deflating the completion count.
- **`Skipped` conflated two things** — the intentional "look around" tap and abandonment.

### Changes
- New shared event `Onboarding Tour Dismissed`, fired from `OnboardingPrompt`'s unmount cleanup when neither button resolved the prompt (gated by a `resolvedRef` so a completed/skipped exit never double-reports). Property `exitReason: 'back'`.
- `Skipped` now carries `exitReason: 'look-around'` so it reads as the intentional secondary exit.
- Additive to the shared event catalog; web's tour is unaffected (it fires raw strings and still uses the real `Step Advanced`).

Every `Started` now resolves to exactly one of **Completed / Skipped / Dismissed**. (App-kill while backgrounded stays unrecoverable — no unmount fires — which is acceptable and unavoidable.)

## Release Notes

## Test plan

- `vp run test:mobile` — 3066 passed (incl. new `OnboardingPrompt.test.tsx` unmount-guard cases + extended `onboarding-analytics.test.ts`)
- `vp check` — 0 errors
- Post-merge (OTA): confirm `Onboarding Tour Dismissed` appears and `Started ≈ Completed + Skipped + Dismissed` (the ~29% gap closes).

https://claude.ai/code/session_01UbaBRgoSJ2FCeagnVrtegQ

### DIFF
--- a/packages/mobile/src/components/onboarding/OnboardingPrompt.tsx
+++ b/packages/mobile/src/components/onboarding/OnboardingPrompt.tsx
@@ -9,6 +9,7 @@ import { useOnboardingCopy } from '../../lib/onboarding/use-onboarding-copy';
 import {
   trackStepViewed,
   trackTourCompleted,
+  trackTourDismissed,
   trackTourSkipped,
   trackTourStarted,
 } from '../../lib/onboarding/onboarding-analytics';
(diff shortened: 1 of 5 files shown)


marcodejongh:
Thanks for the review — both notes addressed:

**1. Strict Mode spurious dismiss:** verified safe. `isAnalyticsEnabled = !!apiKey && !__DEV__` (`posthog-client.ts:34`) and `track` no-ops on a null client (`create-analytics.ts` `if (!client) return`), so analytics is a hard no-op in dev — no phantom `Dismissed` events during simulator testing regardless of Strict Mode. Strict Mode's double-invoke is dev-only (never production), and there's no `<StrictMode>` wrapper in the app root anyway. No code change needed.

**2. `exitReason` imprecision:** fixed in f4f75f5 — changed `'back'` → `'unresolved'` so the field stays mechanism-neutral (the guard can't distinguish hardware-back from a programmatic unmount, so it claims neither). Updated the companion PostHog insight's series label to match.

PR #3397, opened 2026-07-03
fix: rescale JSON-imported tick quality from raw Aurora 0-3 to 1-5

### LINKED ISSUE #3390: boardsesh_ticks.quality: JSON-imported Kilter logbooks stored on raw 1–3 scale (~235k rows)
## Problem

`boardsesh_ticks.quality` is documented as 1–5, but rows created by the JSON logbook-import path store the raw Aurora **1–3** scale. A climb a user rated 3-out-of-3 ("classic") is saved as `3` and renders as **3 out of 5 (mediocre)** anywhere per-tick quality is shown.

## Evidence

JSON-imported **Kilter** quality distribution (read-only prod, July 2026):

```
quality 1 →   4,449
quality 2 →  15,289
quality 3 → 215,416
quality 4 →       6
quality 5 →       7
```

~235k rows piled into 1/2/3 with 13 above 3 — raw 1–3, not 1–5. The live API-sync path (`convertQuality`) and the aggregate `board_climb_stats.quality_average` (`normalizeQualityTo5` + `quality_normalized`) are both correct; only the per-tick column from JSON import is affected. Tension came through the API path and is fine.

## Root cause

`packages/aurora-sync/src/sync/json-import.ts:962`

```ts
// The JSON export 'stars' field is already on a 1-5 scale (user-facing),
// unlike the Aurora API 'quality' field which is 0-3.
quality: ascent.stars,
```

The comment's assumption is false for Kilter exports — `stars` is the raw 1–3. Migration `0079_backfill_tick_quality_scale.sql` deliberately skipped JSON-import rows ("already 1-5"), so the normalization never reached them.

## Fix sketch

1. **Fix the import path** (`json-import.ts:962`) to run `stars` through `convertQuality` (1→1, 2→3, 3→5) instead of trusting it, so no new rows land on the wrong scale.
2. **Re-normalise the existing backfill**: a migration converting JSON-import Kilter rows with the same mapping. Scope by source (`aurora_id LIKE 'json-import-%'`, `board_type = 'kilter'`) — a raw `3` (should become 5) is indistinguishable from a correctly-converted `3` otherwise, so the rewrite has to be source-scoped and idempotent.
3. **Consider a per-row scale flag** (like `board_climb_stats.quality_normalized`) so this class of bug is detectable rather than inferred from a distribution.

Detection lower bound: aurora-synced ticks with `quality IN (0,2,4)` can't exist after conversion (valid = 1/3/5).

## Why it matters

Blocks surfacing per-tick quality (e.g. strong-climber "classics" lists, per-user star ratings). Background and sizing: `docs/power-user-analysis.md` §3 (PR #3385). This is an app-wide data-correctness fix, not scoped to any one cohort.

marcodejongh:
## Summary

`boardsesh_ticks.quality` is a 1–5 scale everywhere it's read, but the JSON logbook-import path stored the Aurora export's raw `stars` field (0–3) verbatim, on a false in-code assumption that the export was already 1–5. Prod impact: ~235k Kilter rows (plus 4 Tension) where a climb rated 3-of-3 ("classic") renders as 3-of-5 ("mediocre"). Migration 0079 deliberately skipped `json-import-%` rows on the same false assumption. Closes #3390.

This PR fixes the import path and every remaining out-of-scale quality callsite, and backfills the existing rows:

- **Import path** (`packages/aurora-sync/src/sync/json-import.ts`): `stars` now runs through `convertQuality` (0→null, 1→1, 2→3, 3→5), matching the API-sync path. The ascent-row mapping is extracted into `buildJsonImportAscentTickRow` with a regression test.
- **Migration `0139`**: rescales existing json-import ascents with the same mapping. Source-scoped (`aurora_id LIKE 'json-import-%'`, all Aurora boards — the export format is board-agnostic and prod shows Tension rows with the same raw values), idempotent (`SET updated_at = now()` pushes converted rows out of the `updated_at <= aurora_synced_at` guard), and preserves user edits (the only 13 prod rows above 3 are all user-edited and are excluded by the guard). Also recomputes `board_climb_stats.quality_average` for Boardsesh-owned climbs whose averages were derived from the raw ticks (66 such ticks in prod), mirroring `recomputeClimbStats`.
- **`/api/v1/{board}/proxy/saveAscent`**: quality was validated as an unbounded `z.number()` and stored raw. Now validated as Aurora-scale `int 0–3` (matching the already-published OpenAPI doc, which said 0–3 all along) and converted before storage.
- **Kilter push-back** (`packages/kilter-sync`, gated off today): new `convertQualityToAurora` (1→1, 2→2, 3→2, 4→3, 5→3; exact round-trip for Aurora 1–3) so the 1–5 tick value no longer flows raw into a payload Kilter expects on 1–3.
- **Stale GraphQL doc**: `SimilarClimb.qualityAverage` claimed "0..3 in MoonBoard convention"; corrected to the canonical 1–5 (+ codegen).

Detection invariant after the backfill: an untouched aurora-synced ascent can only hold quality NULL/1/3/5 — any 0/2/4 means a writer is bypassing `convertQuality` again. Per-row scale flag from the issue's item 3 deliberately skipped (post-backfill it would be uniformly true; the invariant above is the cheaper detector).

## Release Notes

- Star ratings from imported Kilter and Tension logbooks now show what you actually rated them — a climb you called a 3-star classic is 5 stars again, not "mediocre".

## Test plan

- New regression test `packages/aurora-sync/src/sync/json-import.test.ts`: stars 0/1/2/3 → null/1/3/5, clamp, deterministic aurora id, `updated_at`/`aurora_synced_at` parity the migration guard relies on.
- New `convertQualityToAurora` unit tests incl. round-trip `convertQualityToAurora(convertQuality(q)) === q` for q ∈ 1..3.
- Migration dry-run against a real Postgres 18 with a fixture covering every case (raw 0/1/2/3, Tension row, user-edited row, quality-4 row, attempt, API-synced row, UI-created row, Boardsesh-owned-climb stats recompute): all expected values, Aurora aggregates untouched, second run rescales 0 rows.
- Pre-flight read-only prod queries (2026-07-03) confirming distribution, scope, and that the 13 rows above 3 are exactly the user-edited ones.
- `vp run typecheck`, `vp check` on changed files, `vp test` for aurora-sync / shared-schema / kilter-sync / web data-sync (412 tests green), `vp run codegen` (no drift).

https://claude.ai/code/session_016awv6SyBaDTUYLyodpWv44

### DIFF
--- a/packages/aurora-sync/src/sync/json-import.test.ts
+++ b/packages/aurora-sync/src/sync/json-import.test.ts
@@ -1,0 +1,57 @@
+import { describe, expect, it } from 'vitest';
+import { buildJsonImportAscentTickRow, generateJsonImportAuroraId, type AuroraExportAscent } from './json-import';
+
+const NOW = '2026-07-03T00:00:00.000Z';
+const CLIMB_UUID = 'climb-uuid-1';
+const CLIMBED_AT = '2026-06-01T10:00:00.000Z';
+const USER_ID = 'user-1';
+
+function makeAscent(overrides: Partial<AuroraExportAscent> = {}): AuroraExportAscent {
+  return {
+    climb: 'Test Climb',
+    angle: 40,
+    count: 2,
+    stars: 3,
+    climbed_at: '2026-06-01 10:00:00',
+    created_at: '2026-06-01 10:05:00',
+    grade: '7a',
+    ...overrides,
+  };
+}
+
+describe('buildJsonImportAscentTickRow', () => {
+  // Regression for #3390: the export 'stars' field is the raw Aurora 0-3
+  // rating and must be converted to the Boardsesh 1-5 scale, not stored raw.
+  it.each([
+    { stars: 1, expected: 1 },
+    { stars: 2, expected: 3 },
+    { stars: 3, expected: 5 },
+  ])('converts raw Aurora stars $stars to Boardsesh quality $expected', ({ stars, expected }) => {
+    const row = buildJsonImportAscentTickRow(USER_ID, 'kilter', makeAscent({ stars }), CLIMB_UUID, CLIMBED_AT, NOW);
+    expect(row.quality).toBe(expected);
+  });
+
+  it('stores unrated (stars 0) as null, not 0', () => {
+    const row = buildJsonImportAscentTickRow(USER_ID, 'kilter', makeAscent({ stars: 0 }), CLIMB_UUID, CLIMBED_AT, NOW);
+    expect(row.quality).toBeNull();
+  });
+
+  it('clamps out-of-range stars to the top of the scale', () => {
+    const row = buildJsonImportAscentTickRow(USER_ID, 'kilter', makeAscent({ stars: 5 }), CLIMB_UUID, CLIMBED_AT, NOW);
+    expect(row.quality).toBe(5);
+  });
+
+  it('builds the row with the deterministic json-import aurora id and matching sync timestamps', () => {
+    const ascent = makeAscent();
+    const row = buildJsonImportAscentTickRow(USER_ID, 'kilter', ascent, CLIMB_UUID, CLIMBED_AT, NOW);
+    expect(row.auroraId).toBe(generateJsonImportAuroraId(USER_ID, CLIMB_UUID, ascent.angle, CLIMBED_AT, 'ascents'));
+    expect(row.auroraType).toBe('ascents');
+    expect(row.status).toBe('send');
+    expect(row.attemptCount).toBe(ascent.count);
+    // The backfill migration's idempotency guard relies on freshly imported
+    // rows having updated_at <= aurora_synced_at; both come from the same
+    // timestamp here.
+    expect(row.updatedAt).toBe(NOW);
+    expect(row.auroraSyncedAt).toBe(NOW);
+  });
+});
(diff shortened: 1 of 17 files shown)


marcodejongh:
Re the review notes:

1. **`.int()` vs `quality: 1.0`** — not a breaking case: JSON `1.0` parses to the JS number `1`, and Zod's `.int()` checks `Number.isInteger`, which is true for it. There is no representable "float 1.0" distinct from `1` after `JSON.parse`. Only genuinely fractional values (`1.5`) are rejected, which is intended — they were never valid Aurora ratings (the route test covers this).

2. **Formula vs lookup** — kept the formula deliberately: it mirrors `convertQuality`'s style next to it in `utils.ts`, and the unit tests pin all five mappings plus the round-trip invariant, so the table lives in the tests.

3. **`quality_normalized = TRUE`** — confirmed correct: the flag means "`quality_average` is on the 1–5 scale", and after the tick rescale the recomputed average is derived from 1–5 ticks. This is exactly the branch `recomputeClimbStats` takes for Boardsesh-owned climbs (`recompute-climb-stats.ts:171-175`).

4. **Storage-path integration test** — agreed it's a gap only if the conversion moves; the route test pins the boundary contract and `convertQuality` pins the mapping, so leaving as-is.

PR #3408, opened 2026-07-03
feat(mobile): classify BLE disconnects and tag them with the board id

marcodejongh:
## Summary

Analytics follow-up to #3288/#3289, motivated by the adversarial telemetry re-run on #3235 (2026-07-03 comment): 52% of `Bluetooth Disconnected` events are unexpected, and only ~14% could be attributed to takeover — via a fragile per-user last-connect join, because the event carried no board id and only raw per-OS codes.

- **`disconnectCategory`** — new `classifyBleDisconnect` (`src/lib/ble/disconnect-category.ts`) buckets the sparse `BleDisconnectInfo` into `peer_terminated` (CBError 7 / GATT 19 — the takeover signature on last-connection-wins boards), `link_timeout` (CBError 6 / GATT 8 — range/idle), `app_recovery` (the #3181 write-stall context markers), `bluetooth_off` (the #3289 central-state marker), `write_failure` (a write flushing out a dead link), or `unknown`. iOS codes are domain-gated so a numerically-overlapping `CBATTErrorDomain` code is never misread. Unmapped codes stay `unknown` rather than guessing.
- **`boardId`** — both `Bluetooth Disconnected` emissions (user + unexpected) now carry the resolved presence board id, the same value the send events use, so disconnects join directly to another user's connect on the same board.

JS-only → ships via OTA; no fingerprint change. The iOS CBError codes themselves only start flowing once #3289 ships in a native build — until then the classifier covers Android (ble-plx codes), the write-failure path, and `unknown`.

## Release Notes

## Test plan

- `vp run test:mobile` — 3,094 pass (8 new: classifier matrix incl. domain-gating and per-platform codes; provider events assert `boardId` + `disconnectCategory` on the peer-terminated, write-failure, and no-info paths)
- `vp run check:mobile-bundle` — pass
- Field verification: after OTA, `disconnectCategory` distribution in PostHog should split the current 86%-unexplained unexpected disconnects; once #3289 ships in a build, `peer_terminated` vs `link_timeout` becomes the takeover-vs-range readout.

Refs #3235, #3288, #3289.

https://claude.ai/code/session_012JamZgH5gcHSFUSjNJvJAn

### DIFF
--- a/packages/mobile/src/lib/ble/__tests__/disconnect-category.test.ts
+++ b/packages/mobile/src/lib/ble/__tests__/disconnect-category.test.ts
@@ -45,6 +45,9 @@ describe('classifyBleDisconnect', () => {
     expect(classifyBleDisconnect({ source: 'native-ios', iosErrorCode: 7, errorDomain: 'CBATTErrorDomain' })).toBe(
       'unknown',
     );
+    // The native adapter always carries a domain; a domain-less native-ios code
+    // is off-contract and must not be trusted.
+    expect(classifyBleDisconnect({ source: 'native-ios', iosErrorCode: 7 })).toBe('unknown');
   });
 
   it('classifies ble-plx iOS codes without a domain', () => {
--- a/packages/mobile/src/lib/ble/__tests__/disconnect-category.test.ts
+++ b/packages/mobile/src/lib/ble/__tests__/disconnect-category.test.ts
@@ -1,0 +1,63 @@
+import { describe, expect, it } from 'vitest';
+import { classifyBleDisconnect } from '../disconnect-category';
+
+describe('classifyBleDisconnect', () => {
+  it('returns unknown when no info was captured', () => {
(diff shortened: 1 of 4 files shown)


marcodejongh:
Second review pass addressed:

1. **Category absence on user disconnects is now a pinned contract** — the user-disconnect test asserts `disconnectCategory` is undefined there (deliberate disconnects have nothing to classify; queries read absence as "user chose").
2. **The `?? undefined` is load-bearing, kept** — `resolvedPresenceBoardIdRef.current ?? presenceBoardIdRef.current` evaluates to `number | null`, and the trailing `?? undefined` converts that `null` to `undefined` so an unresolved board id is stripped by `sanitizeForPosthog` instead of shipping as an explicit `null` property. Dropping it would reintroduce exactly the null-vs-absent ambiguity #3288 avoided.

PR #3409, opened 2026-07-03
feat: gym write-access grants and gym claiming

marcodejongh:
## Summary

Builds on #3313 (community editing of boards/gyms) with two capabilities:

**1. Grant write access to a gym.** Community admins/leaders and gym owners/admins can now give a specific climber *write* access to one gym — a new `editor` role on `gym_members`. Editors can edit the gym's details but can't delete it or manage its membership/boards. Granting reuses the same debounced user-search picker the community-role admin already uses.

**2. Claim a gym.** Most listings are owned by the system import user. A climber who manages a gym can now claim it:
- **Domain-verified (self-service):** if the gym has a `website` on file, the claimant enters an email at that domain; we email a verification link and, on click, transfer ownership. A prior *real* owner is kept on as a gym admin (never wrestled off).
- **Admin review (fallback):** no usable domain → the claim is queued and `admin@boardsesh.com` is notified. Admins approve/deny from a new `/admin/gym-claims` page.

### Where the work lives
- **Backend (GraphQL + one REST verify route):** the claim flow lives here because mobile only authenticates to the backend, which already validates both web (NextAuth) and mobile (JWS) tokens. New `website` column + `gym_claims` table, `grant/revokeGymWriteAccess`, `requestGymClaim`, `reviewGymClaim`, `pendingGymClaims`, and computed `canEdit`/`canGrantAccess`/`canClaim` on `Gym`. The backend gains a small Nodemailer email module (claim verification, admin notification, approval notice).
- **Shared:** new pure-TS `@boardsesh/gym-claim` (domain extraction / matching / free-provider blocklist), used by backend validation and the web/mobile claim UI.
- **Web:** website field in the gym form, a "Claim this gym" dialog, write-access grant/revoke in gym member management, the `/admin/gym-claims` review queue, and `/gym-claim/{success,error}` landing pages (the verify-link redirect target).
- **Mobile:** website field, write-access grant/revoke, and a claim sheet — parity with web. The claim entry sits inside the expanded wall-finder gym row, so a real owner with no rights yet (`canClaim` without `canEdit`) can reach it, not just editors.

### Security review
This diff was reviewed by four independent agents (backend claim flow, authz/schema-contract, web/shared, and an adversarial red-team) in place of the CI review bot (which was quota-limited). The follow-up `fix:` commit closes what they found — most importantly a privilege-escalation where an editor/community-leader could rewrite a gym's `website` and self-verify into ownership (now blocked; editors go through admin review), a confused-deputy transfer (verification email now names the requester), admins being able to approve unverified domain claims (queue/review now `method='admin'`-only), and a stored-`javascript:`-URL angle on `website` (schema restricted to http(s)). The matcher itself was confirmed non-bypassable (exact string-equality, `z.email()` first). Full write-up is in the PR thread.

### Ops prerequisites (before this helps in prod)
- The **backend** service needs SMTP (`SMTP_HOST/PORT/USER/PASSWORD`, `EMAIL_FROM`) — it had no mailer before — plus `ADMIN_EMAIL` (default `admin@boardsesh.com`), `BACKEND_PUBLIC_URL`, and `WEB_PUBLIC_URL`.
- The `admin@boardsesh.com` inbox must route somewhere real.

## Release Notes

- Own a gym? Claim your listing and keep it accurate — verify with a work email or ask us to review.
- Gym owners and community leaders can now hand a crew member write access to keep a gym's details up to date.

## Test plan

- `packages/backend/src/__tests__/gym-write-access-and-claims.test.ts` — **35 real-DB cases**: grant/revoke authorization matrix (community leader/admin/owner grant; editor can edit but not delete/manage/grant; wrong-board-type + plain user rejected; no admin downgrade; revoke only editors; owner-grant rejected), `canGrantAccess`/`canClaim`, `website` round-trip, `requestGymClaim` domain-match/mismatch/free-provider, admin path, `applyGymClaim`/`verifyGymClaimByToken` ownership transfer (real prior owner kept/upgraded to admin + notified; system owner not re-added), `reviewGymClaim` + `pendingGymClaims` admin-gating, and the security regressions (edit-access holder blocked from the domain path, private gym not claimable, domain claims excluded from the admin queue, re-send dedupe).
- `@boardsesh/gym-claim` domain util — **17 cases** (incl. userinfo/non-http rejection, subdomain-aware free-provider matching, www parity).
- Regression: PR #3313's `board-gym-edit-authorization.test.ts` still green (39) after the shared-helper refactor. Full backend suite green (1978, modulo the known board-presence concurrency flake). Mobile suite green (3088).
- `vp run typecheck` (web/backend/shared/mobile), `vp check`, `vp run codegen` (no drift), `vp run check:i18n` + `check:i18n:orphans` + catalog completeness (en/es/fr parity), `vp run check:mobile-bundle` — all green.
- Manual QA (web + Android) pending on the dev server; requires backend SMTP for the live email path (or a local mail catcher).

https://claude.ai/code/session_01XBDzXgCzbf3GZcKGjB8hHv


marcodejongh:
Thanks @claude and @chatgpt-codex-connector — addressed in `0a1e66a0f`:

| Finding | Resolution |
| --- | --- |
| **SMTP failure leaves claim stuck 24h** | Fixed — the pending row is dropped if the send throws, so a retry isn't blocked by the dedup check. +regression test. |
| **Silent success when gym vanishes during admin approval** | Fixed — `reviewGymClaim` now throws instead of returning `true` when `applyGymClaim` returns null. +test. |
| **`canClaim=true` for editors → confusing domain-path error** | Fixed — `canClaim` now excludes anyone with edit access (gym editors/admins, covering community leaders), so the Claim button doesn't show to them. |
| **No pagination in admin claims panel** | Fixed — added a "Load more" (offset paging) instead of dropping claims past 50. |
| **Test gap: SMTP-failure → stuck retry** | Added (the regression test for the first item). |
| **Mobile `website` maxLength mismatch** | Fixed — 200 → 500 to match the schema. |
| **Missing `robots: noindex` on admin page** | Non-issue — `app/admin/layout.tsx` sets `createNoIndexMetadata` for the whole `/admin` subtree, which `/admin/gym-claims` inherits. |
| **P1: editor-set `website` → sockpuppet ownership escalation** | Real, and my same-account block doesn't fully close it. Shipping as-is by product decision (accept-the-risk) with mitigations — same-account block, `canClaim` hides the button, the displaced owner is notified and kept as a gym admin, and every transfer is logged + admin-reversible. Hardening options tracked in #3431. |

Note: the `claude-review` job earlier failed on an API quota limit, not code — this is the re-run's feedback.

marcodejongh:
Rebased on latest `main` (regenerated the migration as `0143_eminent_silver_samurai` — main had since taken `0141`/`0142`; SQL is byte-identical) and addressed the latest review rounds in `d687cc7`. Also ran an independent adversarial review pass over the whole diff; its only finding was the already-tracked #3431 domain-redirect risk (accepted product decision — keeping as-is), and it verified the rest (domain matcher, transfer atomicity, authz gating, SMTP) clean.

| Finding | Resolution |
| --- | --- |
| **Editor can file admin-review claims — backend didn't gate it** | Fixed — `userCanEditGym` now gates the **top** of `requestGymClaim`, so anyone with edit access is refused on *both* the domain and admin-review paths (mirrors `canClaim`). +test (`blocks the admin-review path too…`). |
| **SMTP failure on admin path throws + strands the claim** | Fixed — the admin notification is best-effort (logged, not thrown); the committed queue row is the source of truth. +test (`still queues an admin claim when the notification email fails`). |
| **`secure: true` hardcoded — breaks STARTTLS (587)** | Fixed — `secure` derives from `SMTP_PORT` (465→true, 587→false) with an `SMTP_SECURE` override. |
| **Mobile dupes `extractGymDomain` / skips free-provider guard** | Fixed — `ClaimGymSheet` now uses the shared `extractDomain` + `isClaimableDomain`, so `gmail.com`/`wixsite.com` websites correctly start in admin review. To make the shared helper RN-safe, `extractDomain` was rewritten without the `URL` constructor (same http(s)-only + no-userinfo guards; +edge tests). |
| **Message maxLength mismatch (500 mobile / 1000 backend / none web)** | Fixed — one shared `GYM_CLAIM_MESSAGE_MAX_LENGTH` (1000) used by backend + both clients; web message + website fields got explicit `maxLength`. |
| **`addGymMember` accepts `editor`, bypassing the grant path** | Fixed — `AddGymMemberInputSchema` restricted to `admin`/`member`. +test. |
| **`canEdit`/`canGrantAccess`/`canClaim` optional in TS, `Boolean!` in SDL** | Fixed — non-optional in the `Gym` TS type. |
| **Admin page checks admin in a client `useEffect`** | Fixed — `/admin/gym-claims` is now a server component with a `checkAdmin()` guard, matching `/admin/retention`. |
| **`GymClaimsPanel` silently empty on fetch failure** | Fixed — error + retry state; empty state suppressed while erroring. |
| **Missing tests (expired token, admin-path SMTP fail, plain-member `canClaim`)** | Added all three. |
| **New SMTP env vars undocumented** | Added `docs/backend-email.md`. |

Validation: `vp run typecheck` (web/backend/mobile/shared), `vp run codegen` (no drift), `vp check`, i18n + orphans + catalog parity, `check:mobile-bundle` — all green. Backend claim suite **42** green; PR #3313 authz regression **43** green; shared `gym-claim` **18** green.

marcodejongh:
Thanks @claude — triaged the fresh round in `de5cab5`.

**Fixed**
- **Critical (email href escaping)** — verification + admin-notification links now use the raw (percent-encoded) URL in the `href`; the entity-escaped copy is kept only for the visible "paste this link" text. (Note: today's link is single-param with a UUID token, so `escapeHtml` was already a no-op on it — but the pattern is now correct for any future multi-param link.)
- **Missing tests** — added all three: `requestGymClaim` rejects an unauthenticated caller; `revokeGymWriteAccess` rejects non-grantors (plain user, editor self-revoke, wrong-board-type leader); and a domain re-request with a *different* email replaces the pending row and **invalidates the previously emitted token** (verified via the captured mailer token). Backend claim suite now 45 green.

**Not changing — with reasoning**
- **`z.string().uuid()` on grant/revoke `userId`** — `users.id` is `text('id')` (NextAuth opaque IDs, e.g. the tests' `gw-owner`), not a UUID. `.uuid()` would reject valid IDs; `z.string().min(1)` is correct and consistent with every other user-id input.
- **`findPendingClaim`/`replacePendingClaim` race** — the partial unique index `gym_claims_unique_pending (gym_id, claimant_user_id) WHERE status='pending'` is the real backstop: a concurrent second insert fails with a unique violation rather than creating a duplicate. The pre-check is just an email-dedup optimization; worst case one racing request (already capped at 10/user) gets an error, never corruption.
- **"Load More" offset = `claims.length`** — stays correct: an approve/deny removes the claim from the server's *pending* set **and** local state in lockstep, and you can only act on already-loaded rows, so the next unloaded row sits at exactly `claims.length`. No dup/skip.
- **`z.coerce.number()` on `claimId`** — claim IDs are `bigserial`; coercing the GraphQL `ID` string to a positive int is the right boundary handling and can't "become non-numeric".
- **Prior-owner notification is fire-and-forget** — intentional and documented: it must never roll back an applied ownership transfer. Same best-effort contract as the approval email.

**Deferred (out of scope / product call)**
- **Denial email to claimant** — deliberately silent for now (a bad-faith claimant learning they were denied just invites retry). Can revisit as a follow-up if you'd rather notify.
- **Rate-limit on `GET /api/gym-claims/verify`** — tokens are unguessable UUIDs and it's a single indexed lookup; reasonable defense-in-depth follow-up, but low-value here.
- **`enrichGym` N+1 / DataLoader** and **`window.confirm` → MUI dialog** — both pre-existing (not introduced by this PR; the confirm string is already `t()`-wrapped and only runs in a click handler). Worth a separate cleanup, not this PR.

PR #3410, opened 2026-07-03
feat(mobile): track search result rank and zero-result filters

### LINKED ISSUE #3401: analytics(mobile): instrument search relevance (Search Result Selected + zero-result filter snapshot)
## Context

`Climb Search Performed` tells us results *appeared* (`resultCount`, `hasQuery`, `queryLengthBucket`, `boardName`, `angle`), but nothing tells us a result was **chosen**. So the [App Success dashboard](https://us.posthog.com/project/412845/dashboard/1794534) can measure zero-result rate and search→queue conversion, but not result **relevance / ranking** — the single biggest unanswerable UX question about search.

## Ask

- Emit `Search Result Selected` (or "opened climb from search") with the result **position / rank** plus `boardName` / `angle`. Enables "is our search relevant?" and scroll-to-tap depth.
- On `resultCount = 0`, snapshot the active filter state so we can attribute *which* filter caused the dead-end.

## Why

Powers the search-quality tiles (zero-result rate, search→queue conversion) with a true relevance signal instead of only coverage.

## References

- mobile search: `packages/mobile/app/(tabs)/climbs/index.tsx` (`Climb Search Performed`)

marcodejongh:
## Summary

`Climb Search Performed` tells us results *appeared*, but nothing told us a result was *chosen* — so the App Success dashboard could measure zero-result rate and search→queue conversion, but not result relevance/ranking. This adds:

- A new `Search Result Selected` event, fired when a user opens a climb from the mobile search result list (`packages/mobile/app/(tabs)/climbs/index.tsx`). Carries the tapped result's `rank` (0-based position in the loaded/deduped list), `loadedResultCount`, `boardName`, `angle`, `hasQuery`, `queryLengthBucket`, and `activeFilterCount` — enough to answer "is our search relevant?" and measure scroll-to-tap depth.
- A richer filter-state snapshot attached to the existing `Climb Search Performed` event, but only when the search resolves to zero climbs. Flattens the filter/board-filter fields not already sent unconditionally (status, grade accuracy, tall/wide/beta-video toggles, climb-type, attempted/completed toggles, benchmarks, holds/zone/setter-id filters) into `zeroResult*` properties, so the zero-result rate can be attributed to a specific filter dimension instead of only "some filter was active."

Registered the new event name in the shared `SHARED_EVENTS` catalog (`packages/shared/analytics/src/events.ts`) so it's available for a future web-side mirror without risking a name split.

`handleClimbPress` reads the tap-time search context (rank map, filters, board name/angle) from a ref rather than `useCallback` deps, so it keeps a stable identity across page loads and filter/name edits instead of recreating (and re-rendering every row) on each one.

Closes #3401

## Release Notes

## Test plan

- `packages/mobile/app/(tabs)/climbs/__tests__/index-keyboard.test.tsx`: added a `Search Result Selected` rank-tracking suite (first and later result) and a zero-result filter-snapshot suite (`zeroResult*` fields attached on empty results, absent when results are found).
- `vp run typecheck:mobile`
- `vp test run --project mobile`
- `vp check`


marcodejongh:
Addressed in 06b8e07:

- `zeroResultStatus` now defaults to `null` like the other `zeroResult*` fields.
- Fixed the `vp check` formatting failure (wrapped a too-long test literal into a shared constant).

On the "fires before results load" bug: that guard already exists unchanged at `index.tsx:561` (`if (!firstSearchPage) return;`), directly above where `isZeroResult` is computed — so the effect can't run with `firstSearchPage` undefined. `visibleClimbs` is derived from `searchPages?.pages` via `useMemo` in the same render pass as `firstSearchPage`, so the two are always in sync; switching `isZeroResult` from `firstSearchPage.climbs.length` to `visibleClimbs.length` doesn't introduce or worsen a staleness window.

Left the test file name (`index-keyboard.test.tsx`) as-is per the "not a blocker" note — it matches the existing per-screen-test-file convention in this directory (`holds.test.tsx`, `setters.test.tsx`, `zone-screen.test.tsx` aren't strictly scoped to their filename either). PR description updated to match the actual (merged) test layout.

---

marcodejongh:
Re-checked the "fires before results load" claim against the actual query hook (`packages/mobile/src/lib/graphql/hooks/use-infinite-search-climbs.ts`) — it calls `useInfiniteQuery` with no `placeholderData`/`keepPreviousData` option, so on a query-key change (filters/name/board) react-query sets `data` to `undefined` while the new key is fetching (no stale carry-over from the previous key). `firstSearchPage = searchPages?.pages[0]` is therefore `undefined` during that fetch, and `index.tsx:561`'s `if (!firstSearchPage) return;` — unchanged by this PR, directly above where `isZeroResult` is computed — exits before `isZeroResult`/`visibleClimbs` are ever read. There's no window where the effect body runs against empty/loading data for a new search. Not making a change here; happy to revisit with a concrete repro if one turns up.

---

PR #3413, opened 2026-07-03
fix(mobile): light Mini MoonBoard holds on the correct 12-row grid natively

### LINKED ISSUE #3392: Mini MoonBoard lights wrong holds via native/widget send path (hardcoded 18-row grid)
## Summary

The native iOS MoonBoard encoder hardcodes an 18-row grid, so a **Mini MoonBoard** (12 rows) climb driven through the native/widget send path (Dynamic Island next/prev → `BoardBleManager`, Live Activity) computes serpentine LED positions on the wrong grid and **lights the wrong holds**. It is silent — it never registers as a send failure, so the reliability metric can't see it.

## Evidence (code)

- `packages/mobile/modules/live-activity/ios/BoardBleEncoding.swift:287` — `private static let moonboardGridRows = 18` (hardcoded).
- `BoardBleEncoding.swift:292-305` — `moonboardSerialPosition(holdId:)` uses `moonboardGridRows` for the serpentine column/row math; `makeMoonboardPacket(frames:)` at `:307` takes no `numRows` parameter.
- The JS send path does this correctly: `packages/mobile/src/hooks/use-board-bluetooth.ts` resolves `getMoonBoardGeometryByLayoutId(layoutId).numRows` (12 for Mini). Only the native widget path is wrong.

Mini MoonBoard 2020/2025 geometry is 11 cols × 12 rows (not 18) — see the Mini board-art work in #3287.

## Impact

Narrow but real: Mini MoonBoard + a send triggered from the widget/Live-Activity native path → wrong LEDs. Full-size MoonBoard (18 rows) is unaffected, which is why it slipped through.

## Suggested fix

Thread the board's `numRows` (from the configured board metadata that `configureBoard` already passes to `BoardBleManager`) into `makeMoonboardPacket` / `moonboardSerialPosition` instead of the `18` constant. Native change → needs a native build.

## Priority

P2 — user-facing correctness (wrong holds) but narrow scope (Mini + widget path only).

---
_Filed from the 2026-07-03 two-week merge audit (code review of the #3225/#3233 MoonBoard BLE work)._

marcodejongh:
## Summary

The native iOS BLE encoder (`BoardBleEncoding.swift`) hardcoded an 18-row MoonBoard grid, so a Mini MoonBoard (11×12 — layouts 6 `mini-moonboard-2020` and 7 `mini-moonboard-2025`) lit the **wrong holds** whenever a send went through the native encode path: Dynamic Island / Live Activity next-prev intents, the lightbulb reconnect re-light, write-stall recovery, and state restoration. The foreground JS path was already row-aware (`getMoonboardBluetoothPacket(frames, numRows)`), which is why the bug never showed in-app and never registered as a send failure.

Fix threads the row count through, keeping TS board config as the single source of truth:

- `configureBoard` (JS → native) now carries `numRows`, resolved via `getMoonBoardGeometryByLayoutId(layoutId).numRows` for MoonBoard, `undefined` for Aurora boards (no grid maths).
- `BoardBleConfiguration` persists it in the App Group (optional field — configs persisted by older builds decode as `nil` and fall back to 18, i.e. today's behaviour, self-healing on the next `configureBoard`).
- `makeMoonboardPacket`/`moonboardSerialPosition` take `numRows` (default 18, mirroring the TS encoder's default) — the serpentine column fold and the max-hold bound now use it.
- Android has no native encoder (BLE is JS-side there) — no change.

**Native change** — ships with the next binary, an OTA can't carry it.

Closes #3392

## Release Notes

- Mini MoonBoard now lights the right holds when you drive the wall from the lock screen or Dynamic Island

## Test plan

- Swift: new `testMoonboardSerialPositionMiniGridMath` + `testMoonboardPacketEncodesMiniPositions` in `LiveActivityWidgetTests` mirror the TS Mini parity vectors (`moonboard.test.ts` §Mini, verified against the reverse-engineered MoonBoard app protocol): B1 = serial 23 on 12 rows vs 35 on 18, K12 = 131, hold 133 rejected; packet `p1r42p2r43p132r44` @12 → `l#S0,P23,E131#`. Existing 18-row tests unchanged (default param).
- TS: `moonboardNumRowsForNative` unit tests (12 for layouts 6/7, 18 for standard, undefined for Aurora).
- `vp run typecheck:mobile`, `vp run test:mobile` (3097 passed), `vp run check:mobile-bundle` all green locally.
- No Mini hardware on hand — correctness anchors to the TS parity vectors the JS path already ships with.

### DIFF
--- a/packages/mobile/ios-tests/LiveActivityWidgetTests.swift
+++ b/packages/mobile/ios-tests/LiveActivityWidgetTests.swift
@@ -320,6 +320,30 @@ final class LiveActivityWidgetTests: XCTestCase {
         XCTAssertEqual(state.currentIndex, 2)
     }
 
+    func testMoonboardSerialPositionMiniGridMath() {
+        // Parity with moonboard.test.ts §Mini (numRows = 12): the Mini LED strip
+        // is 11×12, so the odd-column serpentine reversal folds at row 12 (#3392).
+        XCTAssertEqual(BoardBleEncoding.moonboardSerialPosition(holdId: 1, numRows: 12), 0)
+        XCTAssertEqual(BoardBleEncoding.moonboardSerialPosition(holdId: 2, numRows: 12), 23)
+        XCTAssertEqual(BoardBleEncoding.moonboardSerialPosition(holdId: 11, numRows: 12), 120)
+        XCTAssertEqual(BoardBleEncoding.moonboardSerialPosition(holdId: 12, numRows: 12), 1)
+        XCTAssertEqual(BoardBleEncoding.moonboardSerialPosition(holdId: 123, numRows: 12), 12)
+        XCTAssertEqual(BoardBleEncoding.moonboardSerialPosition(holdId: 132, numRows: 12), 131)
+        // Past the 132-LED Mini strip; valid only on the 18-row wall.
+        XCTAssertNil(BoardBleEncoding.moonboardSerialPosition(holdId: 133, numRows: 12))
+    }
+
+    func testMoonboardPacketEncodesMiniPositions() {
+        // Same holds address different LEDs on the two grids: B1 is serial 35 on
+        // the 18-row wall but 23 on the 12-row Mini.
+        let mini = BoardBleEncoding.makeMoonboardPacket(frames: "p1r42p2r43p132r44", numRows: 12)
+        XCTAssertEqual(String(decoding: mini.packet, as: UTF8.self), "l#S0,P23,E131#")
+        XCTAssertEqual(mini.skippedPositionCount, 0)
+
+        let standard = BoardBleEncoding.makeMoonboardPacket(frames: "p2r43")
+        XCTAssertEqual(String(decoding: standard.packet, as: UTF8.self), "l#P35#")
+    }
+
     func testBoardRenderUrlUsesSharedBoardContext() {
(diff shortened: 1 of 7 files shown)


marcodejongh:
Addressed the review:

- **(4)** `moonboardDefaultGridRows` is `private` again — `makeMoonboardPacket`/`moonboardSerialPosition` now take `numRows: Int? = nil` and resolve the default internally, so `BoardBleManager` passes `configuration.numRows` straight through without seeing the encoding detail.
- **(3)** Fixed the garbled test comment.
- **(2)** `getMoonBoardGeometryByLayoutId` can't return undefined — it falls back to `STANDARD_MOONBOARD_GEOMETRY` on a lookup miss (`packages/shared/board-config/src/moonboard-config.ts`), so `.numRows` never throws. Added the suggested `moonboardNumRowsForNative('moonboard', 999) → 18` test to pin that fallback.
- **(1)** Kept the issue references: the surrounding module already uses them as established style (`#3181`, `#3219` in `BoardBleManager.swift` et al.), and CLAUDE.md has no rule against them.

marcodejongh:
Leaving the two non-blocking observations as-is to keep this green build mergeable: packet-level Mini skip counting is covered by the serial-position test (`133 → nil`) plus the 18-row packet skip test exercising the same counting loop, and the grouped assertions in `moonboardNumRowsForNative` are one behaviour (layout→rows table) rather than four.

PR #3414, opened 2026-07-03
fix(mobile): make native session sync survive reconnects and queue edits

marcodejongh:
## Summary

Four bugs in the native (backgrounded) session WebSocket manager, found in an adversarial review of the Live Activity stack. All native Swift — ships with the next binary, not OTA.

1. **Reconnect loop killed background updates.** `lastSequence` was never reset on reconnect and FullSync was gap-checked, so after a background WS drop plus ≥2 peer navigations, every reconnect's FullSync tripped `hasSequenceGap` → cancel → reconnect → same FullSync, forever (`reconnectAttempt` reset on every ack, so the max-attempts guard never fired). Background Live Activity updates were dead for the rest of the session, with constant connect/join/subscribe churn. FullSync is now accepted unconditionally as the authoritative snapshot (`QueueSequencePolicy`), and `lastSequence` resets when a connection opens.
2. **Queue edits above the current climb shifted it.** `currentIndex` was a raw index, so a peer removing/inserting/reordering an item ahead of the current climb pointed the Live Activity — and widget next/prev + the BLE re-light — at the wrong climb until the next FullSync. The current item is now tracked by uuid across mutations (pure `QueueStateReducer`, mirroring the JS reducer).
3. **A peer navigating to a search result repainted the OLD climb.** A `CurrentClimbChanged` for a climb not in the queue kept the stale index and, because that event repaints, re-wrote the previous climb's holds to the wall — actively fighting the peer. It now appends the climb and points at it, mirroring the JS reducer's `shouldAddToQueue`.
4. **Subscription errors were silently swallowed.** A graphql-ws `error` for the queueUpdates subscription fell through the mutation-error guard; server pings kept the ping-timeout watchdog quiet, so the connection looked healthy while delivering nothing. It now reconnects to resubscribe — bounded, because `reconnectAttempt` now resets when queue data actually flows rather than on `connection_ack` (which also bounds the pre-existing joinSession retry loop).

Queue event parsing and state transitions moved verbatim-plus-fixes from `SessionWebSocketManager.swift` into `SessionQueueState.swift` so they compile into the `BoardseshTests` target.

## Release Notes

- Party sessions stay in sync on the lock screen: reconnecting after a dead spot no longer freezes the Live Activity, and a crew mate shuffling the queue no longer flips it to the wrong climb

## Test plan

- New XCTests (`ios-rn-ci`): FullSync accepted across a sequence jump while deltas still resync; current climb followed by uuid across remove/insert/reorder; removing the current climb falls to its successor; off-queue current climb appended (event + FullSync variants).
- `swiftc -typecheck` clean on the extracted `SessionQueueState.swift`; full-app compile covered by CI.
- `vp check` clean on the touched script (`prepare-rn-ios-tests.mjs` gains the new source file).

### DIFF
--- a/packages/mobile/ios-tests/LiveActivityWidgetTests.swift
+++ b/packages/mobile/ios-tests/LiveActivityWidgetTests.swift
@@ -232,6 +232,42 @@ final class LiveActivityWidgetTests: XCTestCase {
         // Removing the last item while it is current clamps to the new tail.
         state = QueueStateReducer.apply(.itemRemoved(uuid: "C", sequence: 2), to: state)
         XCTAssertEqual(state.currentIndex, 0)
+
+        // Emptying the queue leaves index 0 into an empty array — every
+        // consumer bounds-checks (SharedQueueState.currentItem returns nil,
+        // displayCurrentItemOnBleQueue clears the board), so this must stay a
+        // representable, non-crashing state.
+        state = QueueStateReducer.apply(.itemRemoved(uuid: "A", sequence: 3), to: state)
+        XCTAssertTrue(state.items.isEmpty)
+        XCTAssertEqual(state.currentIndex, 0)
+        SharedQueueState.save(items: state.items, currentIndex: state.currentIndex, to: defaults)
+        XCTAssertNil(SharedQueueState.currentItem(from: defaults))
+    }
+
+    func testGraphQLErrorRoutingCoversSubscriptionAndMutationIds() {
+        // A subscription-level error must reconnect — it used to fall through
+        // the mutation-only handling and was silently dropped while server
+        // pings kept the connection looking healthy.
+        XCTAssertEqual(
+            QueueGraphQLErrorRouting.action(forOperationId: "1", subscriptionId: "1"),
+            .reconnect
+        )
+        XCTAssertEqual(
+            QueueGraphQLErrorRouting.action(forOperationId: "join-session", subscriptionId: "1"),
+            .reconnect
+        )
+        XCTAssertEqual(
+            QueueGraphQLErrorRouting.action(forOperationId: "mutation-abc123", subscriptionId: "1"),
+            .revertOptimisticNavigation(correlationId: "abc123")
+        )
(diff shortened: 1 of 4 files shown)


marcodejongh:
Addressed the review:

- **Empty-queue edge**: added coverage — emptying the queue leaves `currentIndex 0` into an empty array, and the test asserts the consumer contract (`SharedQueueState.currentItem` → nil; `displayCurrentItemOnBleQueue` already bounds-checks and clears the board). No crash path exists: `buildContentState` guards `!items.isEmpty` and the widget's `currentItem` is bounds-checked.
- **Bug #4 test**: the error routing (subscription id / join-session → reconnect, `mutation-*` → optimistic revert, else ignore) is extracted into pure `QueueGraphQLErrorRouting` and pinned by `testGraphQLErrorRoutingCoversSubscriptionAndMutationIds`. The `reconnectAttempt` reset-on-data move stays untestable without hauling URLSession into the test target — noted the rationale in comments at both sites.

marcodejongh:
Addressed round 2:

- **(3)** `QueueSequencePolicy.decision(for:lastKnown:)` — the event now exposes `var sequence` and the redundant `receivedSequence` parameter (and the manager's `sequenceFromEvent`) are gone; the caller can no longer pass mismatched values.
- **(2)** Added `testClimbMirroredUpdatesOnlyTheMatchingItem` — matching-uuid flip, plus nil/unknown-uuid no-op cases.
- **(1)** No behaviour change to flag: the notification was not dropped. `applyEventOnQueue` calls `persistAndNotify` unconditionally *after* the reducer, so a climbMirrored with a missing uuid still persists + repaints exactly as the old guard-else did (the reducer just returns state unchanged). The new test documents this.
- **(4)** Deliberately mirrors the JS reducer (`shouldAddToQueue` with the same append-if-absent idempotency); growth is bounded in practice by the next FullSync, which replaces the whole array with the server's queue. Fixing the accumulation semantics belongs in the shared JS reducer first, not in a divergent native copy.

Issue #3402, opened 2026-07-03
analytics(mobile): instrument push/notifications, install attribution, and a first-class signup event


marcodejongh:
## Context

Three acquisition / re-engagement gaps found while building the [App Success dashboard](https://us.posthog.com/project/412845/dashboard/1794534).

## Findings

1. **No push / notification events** (permission granted, notification received / opened). The main lever for waking dormant users is invisible — the lifecycle tile shows a large dormant pool with no way to measure reactivation.
2. **No install attribution.** `Application Installed` carries no channel / campaign / referrer, so channel ROI and organic-vs-paid are impossible (App Store / Play install referrer).
3. **No first-class signup event.** Mobile registration rides `Login Succeeded { is_registration: true }` (`app/auth/register.tsx`). A dedicated `Signup Completed` (web already has one) would make activation funnels cleaner and align with web.

## References

- `packages/mobile/app/auth/register.tsx`
- lifecycle / dormant tile on dashboard 1794534

_Filed from the App Success dashboard build (PostHog dashboard 1794534)._

marcodejongh:
Split this into two pieces after digging in:

- Install attribution + a first-class mobile `Signup Completed` event → #3415 (in progress)
- Push/notification events → #3416, filed separately since there's no push-notification system on mobile at all yet (no `expo-notifications`, no permission flow, no token registration) — instrumenting it means building it first, a much bigger scope than the other two findings.

---
_Generated by [Claude Code](https://claude.ai/code)_

marcodejongh:
Closing as a duplicate of #3416, which stays open as the canonical issue for this work.

**Why they're the same:** #3416 was split off #3402 and now covers all of #3402's remaining scope. #3402's findings 2 (install attribution) and 3 (first-class signup event) shipped in #3415; its only live finding — no push/notification permission/received/opened events — is precisely what #3416 tracks, rescoped as a real feature build because expo-notifications does not exist in packages/mobile.

Anything in here that #3416 doesn't already cover should be moved there rather than lost. Reopen if you think this is a distinct defect.

_Filed by an issue-triage sweep, 2026-08-24._

PR #3415, opened 2026-07-03
feat(mobile): Signup Completed event + Android install attribution

marcodejongh:
## Summary

Closes part of #3402 (install attribution + a mobile `Signup Completed` event). Push/notification instrumentation from that issue is **not** included here — there's no push-notification system on mobile at all yet (no `expo-notifications`, no permission flow, no token registration), so building it is a separate, much larger follow-up (tracked in a new issue linked from #3402).

- **`SignupCompleted` shared event** — added to `SHARED_EVENTS` (`packages/shared/analytics/src/events.ts`). Web's existing raw-string `'Signup Completed'` fire is migrated to the shared constant (`auth-page-content.tsx`). Mobile's credentials registration (`app/auth/register.tsx`) now fires it too, alongside the existing `LoginSucceeded(is_registration: true)`, plus first-touch `signup_at`/`signup_auth_method` person properties matching web's pattern. OAuth registration on both platforms is unchanged — it stays tagged only via `LoginSucceeded`'s `is_registration: true`, matching web (which has no separate OAuth-signup event either).
- **Android Play Install Referrer attribution** — a new local Expo native module (`packages/mobile/modules/install-referrer`) wraps Google's `InstallReferrerClient` in a coroutine, resolving `null` (never throwing) on any non-OK response or exception. A new config plugin (`with-android-install-referrer-queries`) declares `com.android.vending` package visibility (required on API 30+). The JS wrapper (`src/lib/install-referrer.ts`) fetches once, caches the flag, and — on a real result — attaches parsed `utm_source`/`utm_medium`/`utm_campaign` (plus the raw referrer string and click/begin timestamps) as PostHog person properties, and fires a distinct `Install Attributed` event. Mounted via a new `InstallReferrerTracker` component (Android-only, fire-and-forget on mount) beside the existing `OtaUpdateTracker`. iOS gets no attribution data in this PR — documented as a known gap (no equivalent without Apple Search Ads/SKAdNetwork).

## Release Notes

## Test plan

- `packages/mobile/app/auth/__tests__/register-analytics.test.tsx` — new: asserts `SignupCompleted` + person properties fire on a successful credentials registration, and do **not** fire on a failed one.
- `packages/mobile/src/lib/__tests__/install-referrer.test.ts` — new: covers fetch-once/cache-the-flag/never-throw for `maybeFetchAndAttachInstallReferrer`, plus `parseInstallReferrer`'s utm-param parsing.
- `packages/mobile/src/lib/__tests__/android-install-referrer-queries-plugin.test.ts` — new: mirrors the existing bluetooth-feature manifest-plugin test (add/preserve/idempotent).
- Native/Gradle path (the actual `InstallReferrerClient` binding) can't be exercised in CI/sandbox — no Play Store, no tracked install link. Will do a manual Play Console internal-testing-track verification post-merge with a tagged referrer URL, and confirm `expo prebuild --platform android` generates the `<queries>` manifest block + links the new module.
- `vp run typecheck:mobile`, `vp run typecheck:web`, `vp run test:mobile`, `vp test run --project web`, `vp check` — running now, will report results/push fixes if anything fails.

---
_Generated by [Claude Code](https://claude.ai/code/session_01FRiZ9UwHC24FVJunmhnhge)_

### DIFF
--- a/packages/mobile/app.config.ts
+++ b/packages/mobile/app.config.ts
@@ -506,6 +506,12 @@ export default ({ config, projectRoot }: ConfigContext): ExpoConfig & { newArchE
       // Play device filtering. Unconditional (EAS-safe) — it's a pure manifest
       // addition with no signing/credential implications.
       './plugins/with-android-bluetooth-feature',
+      // Declares com.android.vending as a visible package under Android 11+
+      // package-visibility filtering, so the Play Install Referrer native module
+      // (packages/mobile/modules/install-referrer) can bind to the Play Store
+      // service instead of always resolving FEATURE_NOT_SUPPORTED. EAS-safe pure
+      // manifest addition.
+      './plugins/with-android-install-referrer-queries',
       // Adds the <service android:foregroundServiceType="connectedDevice"> + the
       // notification-action <receiver> for the background BLE session (the
       // Android counterpart to the iOS Live Activity). EAS-safe manifest-only mod.
--- a/packages/mobile/app/_layout.tsx
+++ b/packages/mobile/app/_layout.tsx
@@ -61,6 +61,7 @@ import { loadRequiredFonts } from '../src/lib/required-fonts';
 import { AnalyticsProvider } from '../src/components/analytics/AnalyticsProvider';
 import { AnalyticsScreenTracker } from '../src/components/analytics/AnalyticsScreenTracker';
 import { OtaUpdateTracker } from '../src/components/analytics/OtaUpdateTracker';
+import { InstallReferrerTracker } from '../src/components/analytics/InstallReferrerTracker';
 import { OnboardingGate } from '../src/components/onboarding/OnboardingGate';
 import { AccessoryOnboardingTip } from '../src/components/onboarding/AccessoryOnboardingTip';
 import { FreezeDebugOverlay } from '../src/components/FreezeDebugOverlay';
@@ -502,6 +503,7 @@ function RootLayout() {
                                                         </TabBarHeightProvider>
                                                         <AnalyticsScreenTracker />
                                                         <OtaUpdateTracker />
+                                                        <InstallReferrerTracker />
                                                       </ShareTargetProvider>
                                                     </DeepLinkProvider>
                                                   </DrawerHostProvider>
--- a/packages/mobile/app/auth/__tests__/register-analytics.test.tsx
+++ b/packages/mobile/app/auth/__tests__/register-analytics.test.tsx
@@ -1,0 +1,138 @@
+// @vitest-environment jsdom
+import { act, render, waitFor } from '@testing-library/react';
+import { createElement, type ReactNode } from 'react';
+import { beforeEach, describe, expect, it, vi } from 'vitest';
+import { SHARED_EVENTS } from '@boardsesh/analytics';
+
+const analytics = vi.hoisted(() => ({ track: vi.fn(), setPersonProperties: vi.fn() }));
+const auth = vi.hoisted(() => ({ register: vi.fn() }));
+const router = vi.hoisted(() => ({ replace: vi.fn() }));
+
+// Captures the fields + submit callback AuthFieldset receives, so the test can
+// drive the form like a real user (fill email/password/confirmPassword, then
+// submit) without mounting the platform-split native field/button trees.
+const form = vi.hoisted(() => ({
+  setters: {} as Record<string, (text: string) => void>,
+  submit: null as (() => void) | null,
+}));
+
+vi.mock('../../../src/lib/analytics', () => ({
+  track: analytics.track,
+  setPersonProperties: analytics.setPersonProperties,
+}));
+vi.mock('../../../src/providers/auth-provider', () => ({ useAuth: () => ({ register: auth.register }) }));
+vi.mock('../../../src/hooks/use-native-oauth-sign-in', () => ({
+  useNativeOAuthSignIn: () => ({ signIn: vi.fn(), inProgress: false }),
+}));
+vi.mock('../../../src/lib/auth', () => ({ isGoogleSignInConfigured: () => false }));
+vi.mock('../../../src/lib/error-reporting', () => ({ reportError: vi.fn() }));
+vi.mock('../../../src/lib/haptics', () => ({ hapticLight: vi.fn() }));
+vi.mock('../../../src/providers/theme-provider', () => ({
+  useTheme: () => ({ colorScheme: 'light', systemColors: {} }),
+}));
+vi.mock('react-i18next', () => ({ useTranslation: () => ({ t: (key: string) => key }) }));
+vi.mock('expo-router', () => ({
+  Stack: { Screen: () => null },
+  useRouter: () => router,
+}));
+vi.mock('expo-apple-authentication', () => ({
+  AppleAuthenticationButton: () => null,
+  AppleAuthenticationButtonType: { SIGN_UP: 'sign_up' },
+  AppleAuthenticationButtonStyle: { WHITE: 'white', BLACK: 'black' },
+}));
+vi.mock('@react-native-google-signin/google-signin', () => ({
+  GoogleSigninButton: Object.assign(() => null, {
+    Size: { Wide: 'wide' },
+    Color: { Dark: 'dark', Light: 'light' },
+  }),
+}));
+vi.mock('react-native', () => ({
+  KeyboardAvoidingView: ({ children }: { children?: ReactNode }) => createElement('div', null, children),
+  Platform: { OS: 'android' },
+  Pressable: ({ children, onPress }: { children?: ReactNode; onPress?: () => void }) =>
+    createElement('button', { onClick: onPress }, children),
+  ScrollView: ({ children }: { children?: ReactNode }) => createElement('div', null, children),
+  StyleSheet: { create: (styles: unknown) => styles },
+  View: ({ children }: { children?: ReactNode }) => createElement('div', null, children),
+}));
+
+vi.mock('../../../src/components/AuthFieldset', () => ({
+  AuthFieldset: ({
+    fields,
+    onSubmit,
+  }: {
+    fields: Array<{ key: string; onChangeText: (text: string) => void }>;
+    onSubmit?: () => void;
+  }) => {
+    fields.forEach((field) => {
+      form.setters[field.key] = field.onChangeText;
+    });
+    form.submit = onSubmit ?? null;
+    return createElement('div');
+  },
+}));
+vi.mock('../../../src/components/Button', () => ({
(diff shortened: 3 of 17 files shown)


marcodejongh:
Addressed in 613c682:

- **Organic-install inflation bug**: fixed — `Install Attributed` now only fires when at least one `utm_*` param resolved; person properties still record the (null) channel unconditionally for cohorting. Added a test for the empty-referrer path.
- **Kotlin cancellation race**: fixed — `invokeOnCancellation` is now registered before `startConnection`.
- **`SHARED_EVENTS` constant**: went a different route than suggested. `packages/shared/analytics/src/events.ts`'s own header comment scopes that catalog to events fired by *both* web and mobile — `Install Attributed` is Android-only with no web equivalent, so it doesn't fit there. The repo already has a precedent for this exact case (`OTA_UPDATE_STATUS_EVENT` in `ota-telemetry.ts` — mobile-only, exported as a local free-string constant instead of a raw literal). Applied the same pattern here (`INSTALL_ATTRIBUTED_EVENT` in `install-referrer.ts`), which fixes the string-drift risk you flagged without misplacing a platform-only event in the cross-platform catalog.

---

marcodejongh:
Addressed in 1764627:

- **Silent catch-all**: fixed — now calls `reportError(error)` before swallowing, so a broken preference store / PostHog init at startup is visible instead of failing this pipeline invisibly.
- **Half-tested retry path**: fixed — the throw-case test now makes a second call and asserts `fetchNative` is actually re-invoked (previously only asserted the flag wasn't set).
- **Empty `onCancellation` lambda**: deliberate, added a comment. `finish()` already calls `endConnection()` before `resume()` is reached, so there's nothing left to clean up if cancellation races the resume call itself — that's a different, narrower race than the one the earlier `invokeOnCancellation`-ordering fix addressed (cancellation *before* `finish()` ever runs).
- **`setPersonProperties`/`track` called without `await`**: not changed — these are synchronous, not `Promise`-returning (see `createAnalytics()` in `packages/shared/analytics/src/create-analytics.ts`: `track(): void`, `setPersonProperties(): boolean`). There's nothing to await.

---

marcodejongh:
Addressed in 9c955b8:

- **Thread safety**: `settled` is now `AtomicBoolean` with `compareAndSet`, and the cancellation handler gates on the same flag so it can't double-call `endConnection()` after `finish()` already settled.
- **`"private": true`**: added. Note the two sibling local modules (`board-renderer`, `live-activity`) don't have it either — left those alone as out of scope for this PR, could be a small follow-up cleanup.
- **`InstallReferrerTracker` test**: added — covers the Android mount-fetch path and the iOS no-op.

---

marcodejongh:
Addressed both non-blocking suggestions in 547e540:

- Wrapped the coroutine in `withTimeout(10_000)` so a wedged Play service can't hang it indefinitely; timing out runs the existing `invokeOnCancellation` cleanup path.
- Added a test for `getPreference` itself rejecting (distinct from `fetchNative` rejecting) — confirms `reportError` fires and `fetchNative` is never reached.

---

marcodejongh:
Addressed in a974ab7:

- **Concurrent-call guard**: added an in-memory `fetchInFlight` guard alongside the persisted flag, plus a test that races two overlapping calls and asserts `fetchNative` is only invoked once.
- **Relative import vs `@boardsesh/install-referrer-module`**: not changed — the package name isn't actually resolvable here. Root `package.json`'s `workspaces` glob is `["packages/*", "packages/shared/*", "packages/board-renderer/wasm"]`, which doesn't cover `packages/mobile/modules/*`, so none of the local Expo modules are linked as importable packages (the `name` field in their `package.json` exists only for Expo's autolinking, not JS resolution). The relative import matches every existing consumer of these local modules, e.g. `board-renderer`'s own shim is imported via `'../../../modules/board-renderer/src/index'` elsewhere in the codebase.
- **10s timeout**: left as-is per your own note — no UX impact either way since this is fire-and-forget, and 10s is a reasonable upper bound rather than a hard requirement.

---

marcodejongh:
Addressed all three in 1cb0991:

- **Structured concurrency**: split the catch. Our own `withTimeout` expiring still resolves `null` (expected degrade path), but any other `CancellationException` now rethrows — a naive "rethrow all `CancellationException`" would have also rethrown `TimeoutCancellationException`, defeating the timeout added in the previous round, so this needed the split rather than the literal suggested snippet.
- **Test isolation**: added `resetInstallReferrerFetchInFlightForTests`, mirroring `OtaUpdateTracker`'s existing `resetOtaStatusReportedForTests` pattern, called in `beforeEach`.
- **OAuth negative test**: added — presses the Google sign-in button and asserts `SignupCompleted` doesn't fire (only the credentials `onSubmit` success branch does).

---

marcodejongh:
Looked into this one closely and am not implementing it — the premise doesn't hold for this codebase's actual client init.

`getPostHogClient()` (`packages/mobile/src/lib/posthog-client.ts`) constructs the PostHog client **synchronously** — `client = new PostHog(apiKey, {...})`, no async step, no promise to wait on. `isAnalyticsEnabled = !!apiKey && !__DEV__` is computed once at module load from build-time `EXPO_PUBLIC_*` env vars, which are fixed for the process's entire lifetime (they're inlined at bundle-build time, not read at runtime). So `setPersonProperties`/`track` returning falsy isn't a transient "SDK still initializing" race — it's a permanent "analytics disabled for this build" state (dev builds, or a build with no PostHog key configured), true from the first call to the last for that process.

Gating `setPreference` on the return value, as suggested, would mean: on every single launch of a dev build (`__DEV__` is always true in dev, so `isAnalyticsEnabled` is always false, so `setPersonProperties` always returns `false`), the flag never persists and `maybeFetchAndAttachInstallReferrer` re-invokes the native `getInstallReferrer()` call — opening a fresh `InstallReferrerClient` service binding on every dev-build launch, forever. That's a real regression (defeats the fetch-once-ever design, for zero benefit since a permanently-disabled build was never going to send that data anyway), not a fix for data loss — there's no "next launch it'll actually be ready" case here to retry into.

Not adding the suggested test either, since it'd encode this incorrect model of the client lifecycle. Happy to revisit if there's a code path I'm missing where the client genuinely becomes ready after being unready.

---

marcodejongh:
The `prebuild-and-build` job caught a real bug in 6d4c8c3: `AsyncFunction("getInstallReferrer") { fetchInstallReferrer() }` — a plain trailing lambda isn't a suspend lambda, so calling the suspend `fetchInstallReferrer()` inside it failed to compile ("Suspend function ... can only be called from a coroutine or another suspend function").

Fixed by using the `Coroutine` infix DSL: `AsyncFunction("getInstallReferrer") Coroutine { fetchInstallReferrer() }`. Verified against `expo-modules-core`'s vendored Kotlin sources (`AsyncFunctionBuilder.kt`, `SuspendFunctionComponent.kt`) rather than guessing — that's the mechanism that actually runs the body inside a `CoroutineScope`. No Android SDK available in this sandbox to compile-check locally, so this rides on CI's `prebuild-and-build` to confirm.

---

marcodejongh:
In 7cbccf2:

- Added the `setPreference`-rejects-after-successful-fetch test.
- Added the URL-encoded utm param test (literal `+` → space, `%XX` decoding for parsed fields, `raw` untouched).

Leaving the other two as-is:
- **Test-only export**: agreed it's intentional (mirrors the existing `resetOtaStatusReportedForTests` precedent), and you already flagged it as not blocking.
- **`queries[0]`-only plugin**: this repo has exactly one plugin that adds a `<queries>` block — there's no second one to collide with today, and multiple top-level `<queries>` elements isn't idiomatic AndroidManifest shape to begin with (unlike `<uses-feature>`, which is a flat repeatable sibling list the bluetooth plugin already handles correctly). Generalizing for a manifest shape nothing in this codebase produces would be speculative — happy to revisit if a second `<queries>`-adding plugin is ever actually added.

---

marcodejongh:
Checked all three against the actual codebase — not making changes for any of them:

1. **`__DEV__`-guard on the test-reset export**: there's no such convention here. The exact precedent this mirrors, `resetOtaStatusReportedForTests` in `OtaUpdateTracker.tsx`, is *also* exported unconditionally with no `__DEV__` guard. CLAUDE.md's only `__DEV__` mention is "use the `__DEV__` global, not `process.env.NODE_ENV`" for runtime dev-mode checks — it doesn't say anything about gating test-only exports. Matching the existing pattern here is correct, not a gap.
2. **`SignupCompleted` payload asymmetry** (`flow: 'native'` on mobile vs `requires_verification` on web): intentional, called out during design — mobile's native registration has no email-verification concept (`registerWithCredentials` always auto-logs-in), so the field would be meaningless there; `flow: 'native'` on mobile matches the existing `LoginSucceeded`/`LoginFailed`/`LoginAttempted` payload shape already fired from the same file. Agreed this isn't a bug (you said so too) — just flagging that any PostHog insight built on this event should account for the platform-specific fields.
3. **`compileSdk 36`**: not out of step — `packages/mobile/modules/live-activity/android/build.gradle` already uses `compileSdk 36` (added in the same commit as `board-renderer`'s `compileSdk 35`, so there's no "current standard" to deviate from — the two existing local modules already disagree, and `live-activity` builds fine in CI with 36 using the same `expo-module-gradle-plugin`). Matched the newer of the two existing precedents.

---

marcodejongh:
`prebuild-and-build` caught a second, related compile error in 1dce521: "Overload resolution ambiguity" on the same `AsyncFunction(...) Coroutine { ... }` line. `Coroutine` is overloaded for 0-8 lambda params — a lambda with no declared parameter list is ambiguous between the 0-arg overload and a 1-arg overload's implicit unused `it`, so the compiler can't pick one. Fixed with an explicit empty parameter list: `Coroutine { -> fetchInstallReferrer() }`, which forces zero-arity resolution. Watching CI for this one now.

---

marcodejongh:
1. **Unawaited analytics calls**: not applicable — `track()` returns `void` and `setPersonProperties()` returns `boolean` (`packages/shared/analytics/src/create-analytics.ts:11,16`), both fully synchronous. There's no promise to await and nothing that could reject; `await`ing a non-promise value is a no-op, so this wouldn't change behavior either way.
2. **10s timeout**: leaving as-is. It's an upper bound for a wedged/misbehaving service on a fire-and-forget path with zero UX impact either way — looser on purpose so a slow device doesn't get a spurious timeout and lose real attribution data. Not tuning further without a concrete signal (e.g. a cold-start trace showing it matters).
3. **Fingerprint/OTA**: already covered in the PR description (native change, manual Play Console verification noted as a post-merge step). `mobile-ci-env-parity.test.ts` is part of the mobile suite, which is already green on this commit (`test-default`/`typecheck` jobs).

---

PR #3418, opened 2026-07-03
fix(mobile): bootstrap PostHog anonymous id from party-profile UUID

### LINKED ISSUE #3398: analytics(mobile): fix pre-identify install/open identity race that breaks activation funnels
## Context

Building the new [App Success PostHog dashboard](https://us.posthog.com/project/412845/dashboard/1794534) surfaced a data-quality bug: the **install → first board send** funnel converts at ~1% (iOS) / ~0.2% (Android). That's impossible — there are ~550 weekly active boarders on ~1,500 installs, so true install→send should be tens of percent.

## Root cause

`Application Installed` and the first `Application Opened` fire under the auto-generated **anonymous** device `distinct_id`, **before** `identify()` runs. This is documented in `packages/mobile/src/lib/posthog-client.ts:50-57`. Later value events (`Climb Sent to Board Success`, etc.) fire under the identified auth user. Person-stitching via `$create_alias` does not reliably reconnect the two, so any funnel anchored on install / first-open massively undercounts.

## Impact

- Install→activation funnels, time-to-first-value, and any "new user → X" funnel are unusable as absolute rates.
- The dashboard works around it with `first_time_for_user` trends ("New users vs new boarders per day"), which need no cross-event stitching — but true funnel conversion and time-to-value stay blocked. The install funnel tile is shipped with a ⚠ caveat.

## Fix options

- Ensure `identify()` / `alias` runs before the first lifecycle capture, or
- Capture a synthetic "first session (identified)" event immediately after identify, and anchor activation funnels on that.

## References

- `packages/mobile/src/lib/posthog-client.ts:50-57`

marcodejongh:
## Summary

Closes #3398.

The mobile install→first-board-send funnel converted at ~1% (iOS) / ~0.2% (Android) against an expected "tens of percent" — a data-quality bug, not a real drop-off. `Application Installed` and the first `Application Opened` were firing under PostHog's own transient anonymous `distinct_id`, before `PartyProfileProvider`'s async `identify()`/`alias()` calls ran with the stable party-profile UUID. PostHog's alias-based person-stitching doesn't reliably reconnect those early events afterward, so any funnel anchored on install/open massively undercounted.

Fix: resolve the party-profile UUID synchronously (via `expo-secure-store`'s JSI sync API) as early as possible — before `posthog-client.ts`'s own module-eval side effect constructs the PostHog client — and pass it as the client's `bootstrap.distinctId`. `posthog-react-native` seeds the anonymous distinct_id from `bootstrap` before its `captureAppLifecycleEvents` autocapture fires, so `Application Installed`/`Application Opened` now land on the same id `PartyProfileProvider` later `identify()`s/`alias()`es — no PostHog dashboard changes needed.

To keep `posthog-client.ts`/`analytics.ts` free of the `expo-secure-store` → React Native (Flow-syntax) import chain they're deliberately built without (so the node-env test runner can still import them), the resolution is split:
- `analytics-bootstrap-id.ts` — a pure, dependency-free in-memory slot that `posthog-client.ts` reads.
- `analytics-bootstrap.ts` — a side-effect-only module, imported once from `app/_layout.tsx` ahead of anything that reaches `posthog-client.ts`, that does the actual `expo-secure-store` read/create and populates the slot.

Web is unaffected — `posthog-js-lite` has no app-lifecycle autocapture and the web client is lazily constructed on first real event, not eagerly.

## Release Notes

## Test plan

- `vp test run --project mobile` — 400/400 test files, 3109/3109 tests pass (14 new tests covering the sync party-profile accessor, the bootstrap-id slot, the bootstrap wiring, and `buildPostHogOptions`).
- `vp check packages/mobile` — 0 errors (pre-existing warnings unrelated to this change).
- `vp run check:mobile-bundle` — Metro bundle export succeeds with the new `app/_layout.tsx` import wired in.
- Not independently verifiable in this session: real-world confirmation that `Application Installed`/`Application Opened` now land on the party-profile UUID and that the install→activation funnel conversion rate rises, since that requires observing live PostHog data after release.

---
_Generated by [Claude Code](https://claude.ai/code/session_01VA6MxwRK83aBNttFpZ9Jin)_

### DIFF
--- a/packages/mobile/app/_layout.tsx
+++ b/packages/mobile/app/_layout.tsx
@@ -3,6 +3,12 @@
 // worklet-serialization global-error-capture install, which must wrap Sentry's
 // handler, not the other way round.
 import { wrapWithSentry } from '../src/lib/sentry';
+// Import second, still ahead of anything that reaches posthog-client.ts (e.g.
+// AnalyticsProvider below): resolves the party-profile UUID synchronously and
+// stores it for posthog-client.ts to bootstrap the PostHog SDK's anonymous
+// distinct_id with, before the SDK's own module-eval side effect constructs
+// the client and fires its app-lifecycle autocapture. See analytics-bootstrap.ts.
+import '../src/lib/analytics-bootstrap';
 import { useCallback, useEffect, useRef, useMemo, useState, type ReactNode } from 'react';
 import { LogBox, Pressable, StyleSheet, View } from 'react-native';
 // Navigation theme comes from expo-router's vendored React Navigation. Expo
--- a/packages/mobile/src/lib/__tests__/analytics-bootstrap-id.test.ts
+++ b/packages/mobile/src/lib/__tests__/analytics-bootstrap-id.test.ts
@@ -1,0 +1,23 @@
+import { describe, it, expect, beforeEach } from 'vitest';
+import { getAnalyticsBootstrapId, setAnalyticsBootstrapId } from '../analytics-bootstrap-id';
+
+describe('analytics-bootstrap-id', () => {
+  beforeEach(() => {
+    setAnalyticsBootstrapId(null);
+  });
+
+  it('defaults to null before anything sets it', () => {
+    expect(getAnalyticsBootstrapId()).toBeNull();
+  });
+
+  it('round-trips whatever id was set', () => {
+    setAnalyticsBootstrapId('party-profile-uuid');
+    expect(getAnalyticsBootstrapId()).toBe('party-profile-uuid');
+  });
+
+  it('can be reset back to null', () => {
+    setAnalyticsBootstrapId('party-profile-uuid');
+    setAnalyticsBootstrapId(null);
+    expect(getAnalyticsBootstrapId()).toBeNull();
+  });
+});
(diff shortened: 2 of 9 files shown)


marcodejongh:
Thanks for the review. Addressed the SecureStore-failure visibility gap from the first pass with a dev-only warning (f4ce03a).

On the two remaining points from this pass, leaving both out of this PR:

- **Startup assertion for import ordering**: `bootstrapDistinctId` being `null` is also the *correct* outcome when `SecureStore` legitimately fails (locked keychain, JSI unavailable on an old-arch device) — that's the designed graceful-degradation path this PR relies on, not just a symptom of a reordered import. Asserting/throwing on `null` in `getPostHogClient()` would break analytics init on that legitimate failure path, which is worse than the race being fixed here. Distinguishing "resolver never ran" from "resolver ran and returned null" would need the shared slot to carry an explicit "uninitialized" sentinel distinct from the null fallback — a reasonable follow-up, but more surface than this fix needs.
- **Sentry breadcrumb on SecureStore failure**: reasonable, but wiring `error-reporting.ts`/`sentry.ts` into `party-profile-store.ts` would require mocking Sentry in `party-profile-store.test.ts` and `party-profile-provider.test.tsx` too (same class of test-isolation break this PR already hit once with `expo-secure-store`). Given the failure mode already degrades safely (falls back to PostHog's own anonymous id, exactly like before this fix) and is now dev-visible via the warning above, treating this as a nice-to-have rather than folding it into this PR.

(`__setRaw` in the test mock — confirmed it's defined in the mock factory at line 20 of `party-profile-store.test.ts`; the new tests use it the same way the pre-existing ones do, and the full suite passes.)

---

PR #3419, opened 2026-07-03
fix(mobile): harden the Live Activity lifecycle (token race, ordering, dedup, retry, stale window)

marcodejongh:
## Summary

Five bugs in the Live Activity lifecycle (JS hook + native module), found in an adversarial review of the stack. Contains native Swift changes — ships with the next binary, not OTA.

1. **Auth-token race silently killed push updates.** `authTokenLoaded` latched `true` on the first SecureStore read and never reset, so joining a session right after signing in raced the fresh token read and could ship the guest-phase `null` token to `startSession`. Native push-token registration then bailed on every retry ("no auth token in keychain") for the entire session, and no later call re-sends the token. The token is now stored with the sessionId it was loaded for, and the start gates on that pairing at render time. Bonus: a direct A→B sessionId swap now ends and restarts the activity instead of leaving the native WebSocket, App Group state, and push registration bound to the old session (widget taps were mutating the wrong session).
2. **start/end/update ran in unordered `Task {}`s.** A fast end-then-start (board switch) could run `endAllActivities` after the new start and kill the fresh activity (JS still believes it's active — dead widget for the session); a start-then-end (quick unmount) could orphan a "Loading…" activity on the lock screen; an older update could land after a newer one. Lifecycle work is now a FIFO chain (`enqueueLifecycleWork`), with a lifecycle-generation guard inside the `LiveActivityManager` actor as a backstop against actor-reentrancy interleavings.
3. **Time-only dedup dropped content changes.** The 500 ms dedup window compared wall-control flags and elapsed time only — a quick double-Next dropped the second push and left the Dynamic Island on the wrong climb indefinitely whenever the native WS echo wasn't around to repair it (`refreshStaleDate` even re-pushed the stale state every 60 s). The gate now also requires the content state to equal the last pushed state.
4. **One failed start = dead widget for the session.** ActivityKit start failures include transient ones; the effect deps never changed after a failure, so recovery required leaving and rejoining. Failed starts retry twice with linear backoff (4 s, 8 s); the budget resets on success or deactivation. The old no-self-retry contract test is superseded by explicit retry-budget tests.
5. **Widget intents shrank the stale window from 600 s to 180 s.** A lock-screen Prev/Next/reconnect tap could flip the activity to "Session ended" three minutes later when no APNs heartbeat covered the session (push registration pending, local session). All paths now share `SharedConstants.liveActivityStaleInterval` (10 min, ~6× heartbeat headroom).

Also retitles a bridge test whose name claimed party sessions have "no driver gate" — they do (`boardConnection === 'connectedByMe'`); the test only passed because `beforeEach` set that state.

Widget-target duplicate files (`SharedConstants`, three intents) re-synced byte-identically (drift test green).

## Release Notes

- Lock-screen and Dynamic Island controls are more reliable: the session card no longer freezes on the wrong climb after quick navigation, dies after a hiccup starting up, or flips to "Session ended" minutes after you use it

## Test plan

- New Vitest coverage: auto-retry within budget + stop-after-budget (fake timers), auth-token gating (start waits for the new session's token and passes it through), retitled driver-gate test.
- Existing suites green: `vp run test:mobile` (3098 passed), `vp run typecheck:mobile`, `vp run check:mobile-bundle`, widget-target drift test.
- `swiftc -typecheck` clean on the widget-extension file set (`-D WIDGET_EXTENSION`); full-app compile covered by `ios-rn-ci`.

### DIFF
--- a/docs/live-activity-push-testing.md
+++ b/docs/live-activity-push-testing.md
@@ -372,7 +372,7 @@ This is the key test — verifying that APNs keeps the Live Activity alive when 
 3. Wait 30+ seconds for iOS to fully suspend the app
 4. From a **second device or browser**, change the current climb
 5. Check your iPhone's lock screen — the Live Activity should update within a few seconds via APNs push
-6. The stale date (3 minutes) should NOT trigger because APNs pushes refresh it
+6. The stale date (10 minutes, `SharedConstants.liveActivityStaleInterval`) should NOT trigger because APNs pushes refresh it
 
 ### Verify the Update Source
 
@@ -510,7 +510,7 @@ The .p8 key JWT has expired (tokens are valid for 1 hour). The `@parse/node-apn`
 - The widget extension must have the App Group entitlement to read SharedDefaults.
 - Check backend logs for incoming requests to `/api/widget/navigate` or `/api/widget/take-control`.
 
-### Live Activity goes stale after 3 minutes
+### Live Activity goes stale after 10 minutes
 
 - APNs pushes are not reaching the device. Check backend logs for send failures.
 - The APNs rate limit may be exceeded (~4/hr for non-prominent activities). When the activity is on the lock screen, the limit is much higher.
--- a/packages/mobile/modules/live-activity/ios/ClimbNavigationIntent.swift
+++ b/packages/mobile/modules/live-activity/ios/ClimbNavigationIntent.swift
@@ -88,7 +88,7 @@ enum ClimbNavigationIntent {
 
         for activity in Activity<ClimbSessionAttributes>.activities {
             guard activity.activityState == .active else { continue }
-            let content = ActivityContent(state: newState, staleDate: Date().addingTimeInterval(180))
+            let content = ActivityContent(state: newState, staleDate: Date().addingTimeInterval(SharedConstants.liveActivityStaleInterval))
             await activity.update(content)
         }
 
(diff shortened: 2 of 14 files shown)


marcodejongh:
Addressed the review:

- **Docs**: updated both `docs/live-activity-push-testing.md` references (test step 6 and the troubleshooting header) to the shared 10-minute `SharedConstants.liveActivityStaleInterval`.
- **FIFO/generation-guard tests**: agreed it's the right follow-up, but `LiveActivityModule`/`LiveActivityManager` can't join the `BoardseshTests` source list today — they drag in ExpoModulesCore and SDWebImage (via `ThumbnailFetcher`), which the injected test target doesn't link. Extracting an ActivityKit-free lifecycle skeleton is more surgery than this PR should carry; leaving it as a low-pri follow-up as suggested.
- **eslint-disable**: the comment already names the rule — `// eslint-disable-next-line react-hooks/exhaustive-deps -- startRetryNonce re-fires the start after a failed attempt` (`use-live-activity.ts:362`).

marcodejongh:
Non-blocking notes acknowledged, leaving as-is to keep the green build mergeable: `clearStartRetry` only closes over a ref so a render-scoped function is safe (and it's not in any dep array); `lifecycleQueue.sync` swaps one reference; the task chain is bounded in practice by ActivityKit's own rate limits and each link releases once awaited. The A→B session-swap flow test is a fair follow-up — tracked mentally with the FIFO-queue test gap.

PR #3421, opened 2026-07-03
feat: enrich session and tick analytics events (duration, grade value, tick-form dismiss)

### LINKED ISSUE #3403: analytics(mobile): enrich session & tick events (duration, goal outcome, grade value, tick-form dismiss)
## Context

Session-quality and progression tiles on the [App Success dashboard](https://us.posthog.com/project/412845/dashboard/1794534) are limited by missing event properties.

## What to add

- **Session duration** — only `Session Started` / `Session Ended` + app foreground/background exist; no explicit climbing-session length. Add a duration to `Session Ended`.
- **Session goal outcome** — `Session Started { hasGoal }` has no matching "goal achieved" event, so we can't measure whether goals are met.
- **Grade value on ticks** — `Quick Tick Saved` has boolean `hasDifficulty`, not the grade itself; blocks a grade-progression ("getting stronger over time") retention story.
- **Explicit tick-form dismiss** — tick abandonment is currently *inferred* (`Tick Button Clicked` − `Quick Tick Saved` − `Quick Tick Failed`), which is noisy. An explicit dismiss event (with field-completeness) would make abandonment a real metric and show which field people bail on.

## References

- mobile: `QuickTickBar.tsx`, session command hooks, `packages/shared/analytics/src/events.ts`

marcodejongh:
## Summary

Closes #3403 (partially — see note on goal outcome below).

The App Success PostHog dashboard has session-quality and progression tiles that were starved of data because a few events were missing properties. This PR closes the gaps that are pure enrichment of data already available in scope — no new fetches needed:

- **Session duration** — mobile's `Session Ended` now sends `durationSec`, read from the `SessionSummary` the `EndSession` mutation already returns (`durationMinutes * 60`, matching web's existing property name/unit in `session-lifecycle-tracking.ts`).
- **Grade value on ticks** — `Quick Tick Saved` (mobile **and** web) now sends a resolved grade label (e.g. `"V5"`) alongside the existing numeric `difficulty` id, using each platform's already-loaded grades list (the same inverse lookup pattern the consensus-grade highlighting already does).
- **Explicit tick-form dismiss** — added a mobile-only `Quick Tick Dismissed` event, fired when the tick sheet closes (X button, pan-down, backdrop tap) without a save completing, carrying a field-completeness snapshot (grade/stars/comment/attempt-count picked or not). Previously abandonment could only be inferred by subtracting `Quick Tick Saved` + `Quick Tick Failed` from `Tick Button Clicked`.
  - This required a small plumbing change: all three dismiss paths in `LogAscentSheet` used to call the exact same `onClose` reference, so there was no way to tell a genuine abandon apart from the sheet closing because a save just succeeded. `LogAscentSheet` now wraps `onClose` in a `handleClose` that checks a `savedRef` (set by `QuickTickBar` right before its own save-success `onDismiss()` call) before deciding whether to fire the new event.

**Session goal outcome is deferred**, not implemented here. Investigation found `goal` is unstructured free text with no target/completion concept to key an "achieved" event off, and no mobile UI path currently even sets a goal at session start. A comment explaining this will be left on #3403 so a follow-up can pick it up once goal-setting gets a structured UX.

## Release Notes

## Test plan

- `vp run typecheck:mobile` and `vp run typecheck:web` — pass.
- `vp test run --project mobile` — 399 files / 3103 tests pass, including new coverage for grade-label resolution, the `savedRef`/`fieldSnapshotRef` plumbing, and a new `log-ascent-sheet.test.tsx` covering all three dismiss paths (X button, simulated pan-down, and the no-op-on-save-success case, plus a regression test for the reset-on-reopen behavior).
- `vp test run --project web` — 377 files / 4984 tests pass, including new `use-tick-save` coverage asserting the resolved `grade` field on `Quick Tick Saved`.
- `vp check` — clean on all touched files (pre-existing formatting issues in two unrelated docs files, untouched by this PR).
- Manual verification not performed in this session (telemetry-only change, no UI surface to visually check) — the added/updated unit tests exercise the actual `track()` call payloads end-to-end through the component/hook logic.

---
_Generated by [Claude Code](https://claude.ai/code/session_01NkTQov5CoaUo2KkFYgbzaN)_

### DIFF
--- a/packages/mobile/src/components/LogAscentSheet.tsx
+++ b/packages/mobile/src/components/LogAscentSheet.tsx
@@ -5,16 +5,18 @@
 //
 // Uses `BottomSheetModal` (not the regular `BottomSheet`) so it presents as
 // a native modal above the play drawer's own modal.
-import { useMemo, useRef } from 'react';
+import { useCallback, useEffect, useMemo, useRef } from 'react';
 import { Pressable, StyleSheet, View } from 'react-native';
 import { BottomSheetModal } from '@expo/ui/community/bottom-sheet';
 import { useTranslation } from 'react-i18next';
+import { SHARED_EVENTS } from '@boardsesh/analytics';
 import { useManagedSheet } from '../providers/sheet-presentation-provider';
 import { useTheme } from '../providers/theme-provider';
 import { iosSystemColors } from '../theme/ios-colors';
 import { spacing } from '../theme/tokens';
+import { track } from '../lib/analytics';
 import { Icon } from './Icon';
-import { QuickTickBar } from './play-drawer/QuickTickBar';
+import { QuickTickBar, type QuickTickDismissSnapshot } from './play-drawer/QuickTickBar';
 
 type LogAscentSheetProps = {
   visible: boolean;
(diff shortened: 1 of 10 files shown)


marcodejongh:
Re: `RefObject` vs `MutableRefObject` — this repo's mobile package is on React 19 (`packages/mobile/package.json` pins `react: 19.2.3`, `@types/react: ^19.0.0`), where `useRef<T>()` returns `RefObject<T>` with a **mutable** (non-readonly) `current: T`, and `MutableRefObject` is a legacy alias no longer what `useRef` produces. Verified directly against the installed `@types/react@19.2.15`. `vp run typecheck:mobile` passes clean, so leaving the props as `React.RefObject<T>` — not switching to `MutableRefObject`.

Added the suggested `layoutId: undefined → null` test case in 32ae6c4.

---

marcodejongh:
Re: `RefObject`/`MutableRefObject` — same point as an earlier review pass; see my prior reply above. Confirmed again: this repo's mobile package types are React 19 (`@types/react@19.2.15`, `useRef<T>()` → `RefObject<T>` with mutable `current: T`), and `vp run typecheck:mobile` passes clean, not just lint.

Re: `saveMock.mutate` auto-invoking `onSuccess` — intentional. The new tests need the real `onSuccess` path to fire so `savedRef`/the `QuickTickSaved` payload can actually be asserted on; a bare no-op `vi.fn()` would never exercise that code. The pre-existing tests that click the save button only assert on the *input* passed to `mutate()`, captured synchronously before the callback runs, so they're unaffected — confirmed all 3104 mobile tests still pass with this change.

---

PR #3424, opened 2026-07-03
fix(mobile): keep git trailers and session links out of What's New cards

azuttre:
## Summary

The mobile What's New screen showed a card titled `Co-Authored-By: Claude Fable 5 <noreply@anthropic.com>` with a claude.ai/code session URL as its body (category IMPROVED, from #3367). A second contaminated card (#3180) had a session URL as its title.

**Root cause.** What's New cards come from `changelog.generated.json`, built by `scripts/generate-changelog.ts` from the `## Release Notes` section of merged PR bodies. `extractReleaseNotes` bounds that section at the next heading, a horizontal rule, or the `🤖 Generated with…` footer. Agent-authored PR bodies instead end with a bare `Co-Authored-By:` trailer and a bare session URL right after the notes, with nothing to bound them. The section content was `none` (correctly dropped), so the trailer became the card title and the URL its body. Category fell back to `improved` because the PR title was `refactor(...)`/`ci:` (non-feat/fix).

The OTA publish path (`scripts/mobile-publish.ts`) was checked and is clean: its default `eoas`/`eas update` message is `<short-hash> <commit subject>` only (trailers live in the body), and that message goes to the update dashboard, not the app. Everything user-visible (What's New, CHANGELOG.md, Discord summaries) flows through `extractReleaseNotes`, so one sanitization point covers all of it. PR Release Notes remain the only source for user-visible copy; commit messages feed nothing user-facing.

**Fix** (in `scripts/lib/changelog-transform.ts`):

- Drop git trailer lines wherever they appear in the section: the `*-by:` family (`Co-Authored-By`, `Signed-off-by`, `Reviewed-by`, `Co-developed-by`, …) plus `Change-Id`. Deliberately not a generic `word:` match, so prose like "Faster search: results in half the time" survives.
- Drop link-only URL lines (bare session links). URLs inside a sentence are kept.
- Treat the `_Generated by [Claude Code](…)_` footer variant as a section break, like the existing `🤖` one.

The generated files (`changelog.generated.json`, `CHANGELOG.md`) are intentionally untouched — the `changelog-owned` CI check forbids PR edits to them. The Mobile OTA Production workflow regenerates both on the next push to `main`, so the two contaminated entries (#3367, #3180) get scrubbed automatically once this merges. I verified locally with `vp run generate:changelog`: with this fix, exactly those two entries disappear and nothing else changes.

The second commit fixes formatting drift in two docs files that has the "Lint full repo" job red on `main` (run 28658764462); every PR inherits that failure until it's fixed.

**Also in this PR: the What's New screen readability fix.** The current-build chip ("You're on the build from 13 minutes ago") and the intro line were unreadable. Root cause is the screen backdrop, not the chip palette: on Liquid Glass the stack's `contentStyle` is transparent and What's New is pushed over the user-drawer's `transparentModal`, so the drawer scrim showed through as a stray black background with light-scheme tokens on top. The flex wrapper now paints `systemColors.groupedBackground` (same pattern as `users/[userId]`), which restores the designed chip/text contrast in both schemes and both variants. No token changes.

## Test plan

- [x] New regression tests in `scripts/__tests__/changelog-transform.test.ts` pinning the exact #3367 and #3180 body shapes to "no entry", trailer variants stripped after a real note, colon-prose and in-sentence URLs kept, and the `_Generated by_` footer bounded.
- [x] `vp test run --project scripts` — 326 passed.
- [x] `vp check` — clean (0 errors).
- [x] `vp run typecheck` — clean.
- [x] Local `vp run generate:changelog` diff manually verified: only the two junk entries removed (not committed; the files are OTA-workflow-owned).
- [x] Mobile: `vp run typecheck:mobile`, `vp run test:mobile` (3122 passed), `vp run check:mobile-bundle` — all green.

## Release Notes

What's New now only shows real update notes. Robot commit signatures and stray code links can't sneak into the feed anymore.
The What's New screen also gets its proper background back, so the "you're on this build" chip is readable again instead of gray-on-black.


azuttre:
Re the `StyleSheet.create` note on `changelog.tsx`: keeping the background inline is deliberate. `systemColors.groupedBackground` is resolved at runtime by the theme provider (a PlatformColor on iOS Liquid Glass, hex on Android/Material, and it flips with the colour scheme), so it can't be hoisted into the static `StyleSheet.create` block. The `[styles.x, { backgroundColor: systemColors.y }]` pair is the established pattern for themed surfaces across the app — `users/[userId]/index.tsx`, `(tabs)/home/index.tsx`, and this same file's cards all do it. The `Stand-by:` trailer overmatch is the documented tradeoff from the earlier review rounds — intentionally accepted.

azuttre:
Tested manually on device — ready to merge.

PR #3426, opened 2026-07-03
fix(mobile): paint backgrounds on pushed screens with transparent glass scenes

azuttre:
## Summary

Follow-up audit to the What's New backdrop fix in #3424. On the Liquid Glass variant, `useStackScreenOptions()` spreads `glassStackScreenOptions` whose `contentStyle` is transparent. The root stack keeps an opaque themed scene as a backstop (`contentStyle: undefined`), but root-level routes that re-spread the hook per-screen via `<Stack.Screen options>` override it — an unpainted screen then shows whatever sits behind the scene. Reached off the user-drawer (a `transparentModal` route with a near-black scrim), light-scheme tokens become unreadable.

Audited every `useStackScreenOptions()` usage:

- **Nested-stack children** (auth, boards, and the five tab stacks, where the hook is spread as layout `screenOptions`): their transparent scenes sit inside an opaque root-level scene — backstopped, no change needed.
- **Root-level per-screen spreads without a painted body** — fixed here, mirroring the `changelog.tsx` / `users/[userId]` pattern:
  - `about` (pushed directly from the user-drawer — the exact #3424 repro)
  - `acknowledgements` (drawer → About), `licenses` and `scout` (→ Acknowledgements) — painted `systemColors.groupedBackground`, matching their grouped-card surfaces. On Material and the Android glass fallback `groupedBackground` equals `background`, so the paint is a no-op there; on iOS glass it matches what the themed nav background already showed on the healthy path (`systemGroupedBackground`).
  - `gyms/edit`: the loaded-form state now paints `systemColors.background`, matching the screen's own loading/not-found states (and keeping the form's `secondaryBackground` inputs on their contrast step).

`changelog.tsx` is left alone — its paint rides #3424. The channel switcher was in the original audit but is covered since `main`'s SwitcherForm rewrite: its body is now a native @expo/ui form that paints its own opaque grouped background, so no RN paint is needed there.

## Release Notes

- The About, Acknowledgements, Licenses, and gym-edit screens no longer flash a black background when opened from the side menu in light mode

## Test plan

- `vp run typecheck:mobile` — clean (re-run post-rebase)
- `vp run test:mobile` — 419 files / 3290 tests passed (post-rebase)
- `vp run check:mobile-bundle` — Metro export OK
- `vp check` — format + lint clean on all touched files

### DIFF
--- a/packages/mobile/app/about.tsx
+++ b/packages/mobile/app/about.tsx
@@ -69,9 +69,7 @@ export default function AboutScreen() {
       {/* Variant-aware header (transparent blur on Liquid Glass, opaque M3 on
           Material) from the shared hook; this screen just turns it on + titles it. */}
       <Stack.Screen options={{ ...screenOptions, title: t('mobile.about.title'), headerShown: true }} />
-      {/* Paints groupedBackground (mirrors changelog.tsx): the glass contentStyle
-          is transparent and this route is pushed root-level off the user-drawer,
-          so an unpainted scene shows the drawer's near-black scrim. */}
+      {/* Glass contentStyle is transparent: unpainted, this scene shows the user-drawer's scrim (see changelog.tsx). */}
       <ScrollView
         style={{ backgroundColor: systemColors.groupedBackground }}
         contentInsetAdjustmentBehavior="automatic"
--- a/packages/mobile/app/about.tsx
+++ b/packages/mobile/app/about.tsx
@@ -69,7 +69,11 @@ export default function AboutScreen() {
       {/* Variant-aware header (transparent blur on Liquid Glass, opaque M3 on
           Material) from the shared hook; this screen just turns it on + titles it. */}
       <Stack.Screen options={{ ...screenOptions, title: t('mobile.about.title'), headerShown: true }} />
+      {/* Paints groupedBackground (mirrors changelog.tsx): the glass contentStyle
+          is transparent and this route is pushed root-level off the user-drawer,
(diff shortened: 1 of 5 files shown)


azuttre:
Addressed the review in b92de08e0:

- Comments condensed to one line each, pointing at the root cause (`glassStackScreenOptions` transparent `contentStyle`) via `changelog.tsx`.
- The `ChannelSwitcherScreen` line comment moved above the `return` statement.

Not applied: moving the `backgroundColor` into `StyleSheet.create`. `systemColors` comes from `useTheme()` (a PlatformColor per render on iOS glass), so it can't live in a module-level `StyleSheet.create` — the dynamic-colour-inline pattern (`style={[styles.x, { backgroundColor: systemColors.y }]}`) is what the codebase uses everywhere. `licenses.tsx` and `gyms/edit.tsx` only look different because they had an existing static style to compose with; the ScrollView screens have no static part, so the inline object is the whole style.

marcodejongh:
Conflicts

azuttre:
Rebased on `main` (282116bae). One functional note from the conflict resolution: `main`'s SwitcherForm rewrite replaced the channel switcher's RN ScrollView with a native @expo/ui form (SwiftUI `Form` on iOS), which paints its own opaque grouped background — so the paint this PR added to `ChannelSwitcherScreen.tsx` became obsolete and was dropped. The PR is now the five remaining screens: about, acknowledgements, licenses, scout, gyms/edit.

PR #3430, opened 2026-07-03
test(mobile): cover queue resync error branch + stale-snapshot race

marcodejongh:
## Summary

Closes two coverage gaps in the mobile `QueueProvider` sync-gate recovery path (`resyncQueueFromServer` in `packages/mobile/src/providers/queue-provider.tsx`), flagged in a test-coverage review.

- **Resync error branch.** When the `GetSessionQueueState` resync request itself *throws* (transport failure, 5xx), the `catch` restarts the joined WS subscriptions — their guaranteed initial `FullSync` heals state — and reports a handled error, instead of giving up silently and leaving gap/drift recovery stuck. The existing suite only covered the `queueState: null` return path, never a rejected request. New test drives a throwing fetch and asserts the resubscribe (2 → 4 subscribe calls), exactly one fetch (no retry loop), and that the failure goes through `reportHandledError` only (not the unhandled `reportError`).

- **Stale-snapshot race.** The guard `queueState.sequence < lastTrackedSequence` prevents a slow HTTP snapshot from regressing state after a mid-flight rejoin `FullSync` re-baselined the gate to a higher sequence. No test exercised it. New test holds the resync fetch open with a deferred promise, fires a `FullSync` at sequence 100 mid-flight, then resolves the fetch to an older sequence-5 snapshot — asserting the stale snapshot is skipped and a later delta at 101 still applies cleanly (proving the gate stayed at 100).

### Note on scope

This PR originally also guarded an unguarded cast in `toSyncQueueEvent` (`event.state as { stateHash?: ... }`). That's no longer needed: since the PR was opened, `main`'s queue-provider refactor widened `SubscriptionWireEnvelope`'s FullSync `state` to carry `stateHash?`/`stateHashOrdered?` directly, so `toSyncQueueEvent` now reads `event.state.stateHash` with no cast at all. Re-adding the assertion would reintroduce a cast over clean typed access, so that change was dropped and this PR is test-only.

## Release Notes

## Test plan

- `vp run typecheck:mobile` — clean (rebased on latest `main`).
- `vp test run --project mobile queue-provider-sync-gate` — 8/8 pass.
- `vp run test:mobile` — full mobile suite green.

https://claude.ai/code/session_01NaGinPKWJwe7wgCKnyBsgE

### DIFF
--- a/packages/mobile/src/providers/__tests__/queue-provider-sync-gate.test.tsx
+++ b/packages/mobile/src/providers/__tests__/queue-provider-sync-gate.test.tsx
@@ -608,11 +608,13 @@ describe('QueueProvider queue sync gate', () => {
     // Exactly one fetch: the errored restart never chains into another resync.
     expect(queueStateCalls).toBe(1);
     // The failure was reported as a handled error tagged to the resync op, not
-    // swallowed.
+    // swallowed — and via the handled reporter only. A regression that also
+    // escalated to the unhandled `reportError` must fail this test.
     expect(errorReporter.reportHandledError).toHaveBeenCalledWith(
       expect.any(Error),
       expect.objectContaining({ tags: { source: 'queue-sync', op: 'resync' } }),
     );
+    expect(errorReporter.reportError).not.toHaveBeenCalled();
   });
 
   it('skips a stale resync snapshot that resolves after a mid-flight FullSync re-baselined the gate', async () => {
--- a/packages/mobile/src/providers/__tests__/queue-provider-sync-gate.test.tsx
+++ b/packages/mobile/src/providers/__tests__/queue-provider-sync-gate.test.tsx
@@ -108,6 +108,11 @@ const toast = vi.hoisted(() => ({
   showToast: vi.fn(),
 }));
 
+const errorReporter = vi.hoisted(() => ({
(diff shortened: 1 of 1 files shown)


marcodejongh:
Thanks — checked both points against the current branch (now rebased on latest `main`):

**1. Missing `toSyncQueueEvent` change** — correct that it's not in the diff, and it shouldn't be. Since this PR was opened, `main`'s queue-provider refactor (realtime logic extracted into `use-session-realtime.ts`) also widened `SubscriptionWireEnvelope`'s FullSync `state` to carry `stateHash?`/`stateHashOrdered?` directly. `toSyncQueueEvent` now reads `event.state.stateHash` with **no cast** — the unguarded cast this PR originally guarded no longer exists. Re-adding a `FullSyncWireState` assertion would reintroduce a cast over what's now clean typed access, so I dropped that change and reframed the PR as test-only (description updated). The two tests target `resyncQueueFromServer`, which still lives in `queue-provider.tsx`.

**2. Missing negative assertion** — good catch, added `expect(errorReporter.reportError).not.toHaveBeenCalled()` to the resync-throws test so a regression that also escalates to the unhandled reporter fails (commit `902a34e`).

---

PR #3433, opened 2026-07-03
fix(mobile): keep filter apply button on-screen + help when board isn't in picker

marcodejongh:
## Summary

Fixes two independent bugs surfaced by one in-app feedback report (iPhone 13 mini, iOS 26.1, app 2.1.0):

> "I was trying to filter but the show climbs button is inaccessible - also trying to light up a tension board at my local gym it wasn't showing up as an option"

PostHog telemetry for the reporter's session confirmed both are real and distinct:

**1. Filter "Show climbs" button clipped off-screen.** The user searched climbs but no filter ever applied (zero filter-apply events all session). The climbs filter sheet pins its column to a JS-estimated height, and that estimate over-shoots the true iOS `.fraction(0.9)` detent on short screens (SwiftUI measures the fraction against the sheet's own max height, not `window − topInset`; iOS 26 card-gap + Liquid-Glass/grabber chrome make the real detent shorter). The pinned "Show N climbs" footer landed below the visible sheet. Same failure class as the previously-patched #3330. Fix: fold an iOS card-gap (`SHEET_TOP_GAP_PT`) into the base **before** the fraction, so the bound scales with the detent and keeps erring short. Shared `useSheetColumnStyle` hook, so every iOS sheet gets the more conservative bound.

**2. Board not in the BLE picker, no guidance.** Six connect attempts, all `user_cancelled`. The picker resolved two devices — the user's own saved **kilter** board (which also tripped the config-mismatch alert) and one non-tension board — so **neither** was their gym tension board. The board was genuinely never discovered (iOS scans are service-UUID-filtered; most likely it was asleep, already connected to another device — Aurora allows one connection at a time — or out of range). We can't root-cause board absence from telemetry, so this is UX + instrumentation hardening:
- A "Don't see your board?" troubleshooting block in the picker (awake/powered, one phone at a time, move closer, toggle Bluetooth).
- A hint when devices are found but none match the selected board type.
- Enriched `BLE Picker Devices Resolved` (per-device type tally + `noneMatchedSelectedType`) so future "board absent" reports are diagnosable.

The config-mismatch alert is intentionally left as-is (it's a legitimate switch safety net); the new hint addresses the real confusion upstream.

Scale (last 7 days): `user_cancelled` is the #1 BLE failure (261 users); 48 users started a scan but never once connected.

## Release Notes

- The "Show climbs" button in the filter now stays put on smaller iPhones — no more scrolling into a button you can't reach.
- Connecting a board? If yours isn't in the list, the picker now tells you why (asleep, connected to another phone, or too far) and flags when the boards nearby are a different type than the one you've got selected.

## Test plan

- `vp run test:mobile` — all 3187 pass, incl. updated `use-sheet-column-style` / `sheet` height expectations and new `picker-resolution-stats` cases (the reported saved-kilter + non-tension mix → `noneMatchedSelectedType: true`).
- `vp check` on all changed files — formatting clean.
- i18n parity: new `ble.*` keys added to en-US / es (plafón) / fr (panneau).
- **Device verification pending (authoritative for bug 1):** physical iPhone 13 mini / iOS 26.1 — confirm "Show N climbs" is fully visible/tappable at rest, after scrolling, and after a sub-picker round-trip; and a filter-apply event fires.

## Note on `typecheck:mobile`

An earlier draft flagged a repo-wide `TS2322 BottomSheetProps` typecheck error from the Expo SDK 57 merge (#3396). **CI `typecheck` passes green** — it was a stale `@expo/ui` type resolution local to my worktree, not a real break. All CI checks on this PR are green.


marcodejongh:
Thanks — addressed all three in ca6a2df0a:

- **`gap: 2` → `spacing[1]`** in the picker troubleshoot block.
- **Naming + duplication:** extracted `noListedBoardMatchesSelectedType` (and `effectiveBoardType`) into `picker-resolution-stats.ts`, so the picker hint and the `noneMatchedSelectedType` analytics flag now share one rule instead of two copies with drifting names. The component's local is renamed to `noneMatchedSelectedType` to match.
- **Test coverage:** both helpers are now covered directly (match / no-match / no-selection / empty, plus the reported saved-kilter + non-tension mix). Rather than a full `@expo/ui` render test (fragile to mock), the decision logic the hint depends on is now the tested helper.

On the typecheck note: CI `typecheck` is green — the `BottomSheetProps` error was local to my worktree's stale `@expo/ui` types, not a real break. PR body updated.

marcodejongh:
Both addressed in dabb89e00:

- **Troubleshoot block gated** to `devices.length === 0 || noneMatchedSelectedType` — it no longer shows when the user's board is clearly in the list. Cancel stays unconditional.
- **Dropped the bug-report framing** from the code comments (kept the behavioral description; context lives in git history).

marcodejongh:
Addressed in e091477cf:

- **`effectiveBoardType` silent gap:** now falls back to the name-parsed type when a resolved entry yields no usable board type (corrupted save), so a known board isn't miscounted as `unknownType` and the missing-board hint still fires. Added a test for exactly that case.
- **Double iteration:** `summarizePickerResolution` now derives `noneMatchedSelectedType` from the tally it already built (`selectedType != null && devices.length > 0 && matchedSelectedType === 0`) instead of re-scanning; it's equivalent to the shared `noListedBoardMatchesSelectedType` the UI calls (both rest on `effectiveBoardType`).
- **`SHEET_TOP_GAP_PT`:** comment now notes it's calibrated for the grabber card presentation every sheet in this app uses; a non-card `.pageSheet`/`.formSheet` would just bound a touch short, which is safe.

**i18n formality:** the fr/es difference is intentional house style — the repo's French is consistently formal (`vous`/`votre`, 30+ uses) and Spanish consistently informal (`tú`/`asegúrate`, 34+). These strings match each locale's existing register, so leaving as-is.

marcodejongh:
Thanks for the approve. On the three notes:

1. **`mismatchedSelectedType` with no selected board** — fixed in 9384e0078: a device now counts as `unknownType` when there's no selected board to compare against (neither match nor mismatch), so the event can't report mismatches against a null selection. Test updated. (This path is unreachable from the picker, which always has an active board, but the telemetry is clearer now.)

2. **`SHEET_TOP_GAP_PT` global** — already carries the `// calibrated for the grabber card presentation every sheet in this app uses` note (added last round); a non-card sheet just bounds a touch short, which is safe. No further change.

3. **typeHint + troubleshoot both showing on none-match** — intentional and I'm leaving it: they're complementary (the hint diagnoses *"these are a different board type than your Tension"*, the troubleshoot block gives the *actions* — awake, one connection at a time, move closer), and round 2's review specifically asked to show the troubleshoot block in the none-match case. Flip-flopping the gate would just re-open that thread.

PR #3436, opened 2026-07-03
feat(mobile): add durable PostHog person properties for cohorting

### LINKED ISSUE #3399: analytics(mobile): add durable person properties for cohorting (home board, first-seen, role)
## Context

The only person property set on mobile is `email` (`packages/shared/analytics/src/reconcile-identity.ts`; `setPersonProperties` is exported from `analytics.ts` but never called). Every metric on the new [App Success dashboard](https://us.posthog.com/project/412845/dashboard/1794534) can therefore only be segmented by **behaviour**, not by durable user traits.

## What to add (person properties)

- `first_seen_at` / account-created date → new-vs-veteran retention cohorts.
- `primary_board` / home board type (Kilter / Tension / MoonBoard) → board-level PMF segmentation across every tile.
- `role` / beta-channel flag → separate testers from regular users cleanly (today we lean on `filterTestAccounts`).
- optional: favourite count, has-connected-board, integrations-connected.

All of these are already available at existing call sites — `boardName` / `layoutId` are passed as event props throughout.

## Why

Unlocks "new vs veteran" and "Kilter vs MoonBoard user" splits that are impossible with behaviour-only cohorts, and would let the retention/lifecycle tiles break down by real user maturity.

## References

- `packages/mobile/src/lib/analytics.ts` (`setPersonProperties` exported, unused)

marcodejongh:
## Summary

Mobile only ever set one PostHog person property (`email`, inside the shared identity-reconciliation helper). Everything else on the App Success dashboard could only be segmented by behaviour, never by durable traits like account age or board type. This adds the properties requested in #3399:

- `first_seen_at` (`setOnce`) — account-created date, from a new `UserProfile.createdAt` field (`users.createdAt`, already in the DB).
- `primary_board` (`set`) — the user's inferred home board (`useHomeBoard()`), narrowed to the canonical board name.
- `role` (`set`) — `'tester'` or `'user'`, from the existing `UserProfile.isTester` field.
- `favorite_count` (`set`) — total favourited climbs across all boards, from a new `UserProfile.favoriteCount` field (a new all-boards aggregate; the existing `userFavoriteClimbs` query is scoped per-board).
- `integrations_connected_count` (`set`) — count of connected third-party integrations (Strava today).
- `has_connected_board` (`setOnce`) — a durable, client-only marker set the first time a BLE connection succeeds.

The `first_seen_at`/`primary_board`/`role`/`favorite_count`/`integrations_connected_count` mapping logic lives in a new pure helper, `buildCohortPersonProperties()` in `@boardsesh/analytics`, so web can adopt the same property names later with one call site (not wired up on web in this PR — the issue is scoped to mobile).

Closes #3399.

### Backend
- `UserProfile` GraphQL type gains `createdAt: String!` and `favoriteCount: Int!`.
- `profile` query and `updateProfile` mutation resolvers populate both from the existing `users.createdAt` column and a new `userFavorites` count aggregate (no DB migration needed).

### Mobile
- `PartyProfileProvider` gets a new effect (mirrors web's `language` person-property effect) that calls `setPersonProperties` once auth resolves and the profile/home-board/integrations data is available.
- `registerBluetoothConnection` marks `has_connected_board` on every successful BLE connection (`setOnce` is idempotent server-side, so no local dedup flag is needed).

## Release Notes

## Test plan

- `vp run typecheck` (full monorepo, including generated GraphQL types via codegen) — passes.
- `vp check` (lint + format) on the changed files — 0 warnings/errors.
- `vp test run` — new unit tests added:
  - `packages/shared/analytics/src/__tests__/cohort-person-properties.test.ts` (new)
  - `packages/mobile/src/providers/__tests__/party-profile-provider.test.tsx` — added a test asserting the full `setPersonProperties(set, setOnce)` payload once profile/home-board/integrations data resolves
  - `packages/mobile/src/lib/ble/__tests__/bluetooth-status-store.test.ts` — added a test asserting `has_connected_board` fires on connect
- Full `mobile` (3142 tests) and `shared` analytics suites pass. `backend` suite passes except for pre-existing Docker/Redis-dependent integration tests that fail identically on `main` in this sandbox (no Docker network access to pull the dev-db image) — unrelated to this change; none of the failures touch the `profile`/`updateProfile` resolvers.
- **Not verified**: a live end-to-end run (sign in on a real device/simulator and confirm the properties land in PostHog) — this sandbox couldn't pull the dev database's Docker image (network-restricted registry pull), so I could not start `vp run dev`. QA notes for that manual pass are in `.boardsesh/qa-notes.md` on this branch for whoever picks this up with Docker access.

---
_Generated by [Claude Code](https://claude.ai/code/session_017nA55UxZ95JrCYUxMEGZJo)_

### DIFF
--- a/packages/backend/src/graphql/resolvers/users/__tests__/mutations.test.ts
+++ b/packages/backend/src/graphql/resolvers/users/__tests__/mutations.test.ts
@@ -1,0 +1,91 @@
+/**
+ * Unit tests for userMutations.updateProfile — covers the cohort-property
+ * fields added for issue #3399 (createdAt, favoriteCount) on both the
+ * insert (no existing profile row) and update paths. The db client is
+ * mocked, mirroring tester.test.ts / queries.test.ts.
+ */
+
+import { describe, it, expect, vi, beforeEach } from 'vite-plus/test';
+import type { ConnectionContext } from '@boardsesh/shared-schema';
+
+const { limitMock, insertValuesMock, updateSetMock } = vi.hoisted(() => ({
+  limitMock: vi.fn(),
+  insertValuesMock: vi.fn(async () => undefined),
+  updateSetMock: vi.fn(() => ({ where: vi.fn(async () => undefined) })),
+}));
+
+vi.mock('../../../../db/client', () => ({
+  db: {
+    select: () => {
+      const chain = {
+        from: () => chain,
+        where: () => chain,
+        limit: () => Promise.resolve(limitMock()),
+      };
+      return chain;
+    },
+    insert: () => ({ values: insertValuesMock }),
+    update: () => ({ set: updateSetMock }),
+  },
+}));
+
+vi.mock('../tester', () => ({
+  userIsTester: vi.fn(async () => false),
+}));
+
+import { userMutations } from '../mutations';
+import { userIsTester } from '../tester';
+
+const userIsTesterMock = vi.mocked(userIsTester);
+
+function makeCtx(userId = 'user-1'): ConnectionContext {
+  return { connectionId: `http-${userId}`, userId, isAuthenticated: true };
+}
+
+const REFRESHED_USER_ROW = {
+  id: 'user-1',
+  email: 'climber@example.com',
+  name: 'Climber',
+  image: null,
+  createdAt: new Date('2024-02-02T00:00:00.000Z'),
+  favoriteCount: 3,
+};
+
+describe('userMutations.updateProfile', () => {
+  beforeEach(() => {
+    limitMock.mockReset();
+    insertValuesMock.mockClear();
+    updateSetMock.mockClear();
+    userIsTesterMock.mockReset();
+    userIsTesterMock.mockResolvedValue(false);
+  });
+
+  it('updates an existing profile row and maps createdAt/favoriteCount from the refreshed select', async () => {
+    limitMock
+      .mockReturnValueOnce([{ userId: 'user-1', displayName: 'Old Name', avatarUrl: null }]) // existing profile check
+      .mockReturnValueOnce([REFRESHED_USER_ROW]) // refreshed users select
+      .mockReturnValueOnce([{ userId: 'user-1', displayName: 'New Name', avatarUrl: null }]); // refreshed profiles select
+
+    const result = await userMutations.updateProfile(undefined, { input: { displayName: 'New Name' } }, makeCtx());
+
+    expect(updateSetMock).toHaveBeenCalledOnce();
+    expect(insertValuesMock).not.toHaveBeenCalled();
+    expect(result.createdAt).toBe('2024-02-02T00:00:00.000Z');
+    expect(result.favoriteCount).toBe(3);
+    expect(result.displayName).toBe('New Name');
+  });
+
+  it('creates a new profile row when none exists yet, still returning createdAt/favoriteCount', async () => {
+    limitMock
+      .mockReturnValueOnce([]) // no existing profile row
+      .mockReturnValueOnce([REFRESHED_USER_ROW])
+      .mockReturnValueOnce([{ userId: 'user-1', displayName: 'First Name', avatarUrl: null }]);
+
+    const result = await userMutations.updateProfile(undefined, { input: { displayName: 'First Name' } }, makeCtx());
+
+    expect(insertValuesMock).toHaveBeenCalledOnce();
+    expect(updateSetMock).not.toHaveBeenCalled();
+    expect(result.createdAt).toBe('2024-02-02T00:00:00.000Z');
+    expect(result.favoriteCount).toBe(3);
+  });
+});
--- a/packages/backend/src/graphql/resolvers/users/__tests__/queries.test.ts
+++ b/packages/backend/src/graphql/resolvers/users/__tests__/queries.test.ts
@@ -1,0 +1,119 @@
+/**
+ * Unit tests for userQueries.profile — covers the cohort-property fields added
+ * for issue #3399 (createdAt, favoriteCount) alongside the existing auth-gating
+ * and null-row behaviour. The db client is mocked, mirroring tester.test.ts.
+ */
+
+import { describe, it, expect, vi, beforeEach } from 'vite-plus/test';
+import type { ConnectionContext } from '@boardsesh/shared-schema';
+
+const { limitMock } = vi.hoisted(() => ({
+  limitMock: vi.fn(),
+}));
+
+vi.mock('../../../../db/client', () => ({
+  db: {
+    select: () => {
+      const chain = {
+        from: () => chain,
+        leftJoin: () => chain,
+        where: () => chain,
+        limit: () => Promise.resolve(limitMock()),
+      };
+      return chain;
+    },
+  },
+}));
+
+vi.mock('../tester', () => ({
+  userIsTester: vi.fn(async () => false),
+}));
+
+import { userQueries } from '../queries';
+import { userIsTester } from '../tester';
+
+const userIsTesterMock = vi.mocked(userIsTester);
+
+function makeCtx(overrides: Partial<ConnectionContext> = {}): ConnectionContext {
+  return {
+    connectionId: 'http-user-1',
+    userId: 'user-1',
+    isAuthenticated: true,
+    ...overrides,
+  };
+}
+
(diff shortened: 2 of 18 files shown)


marcodejongh:
Thanks for the review. Addressed one, the other two don't apply:

- **Duplicated `favoriteCount` query / extra round trip** — fixed in ea6eab8: folded into the main select as a correlated subquery instead of a separate `count()` query, so `profile`/`updateProfile` are back to one DB round trip for this part.
- **`useHomeBoard()` / `useIntegrationStatuses()` without an auth guard** — not a bug: both hooks already gate their own network-triggering queries on `isAuthenticated` internally (`useIntegrationStatuses` sets `enabled: isAuthenticated` on its `useQuery`; `useHomeBoard` passes the same to the `useMyBoards`/`useProfile` calls it makes internally). The only unconditional call inside `useHomeBoard` is `useActiveBoard()`, which reads from AsyncStorage, not the network. Calling them unconditionally in the provider doesn't fire unauthenticated network requests.
- **`user.createdAt.toISOString()` null guard** — the column is `timestamp('created_at').defaultNow().notNull()` in the Drizzle schema (DB-level NOT NULL), so this matches the existing no-guard pattern used for other non-null timestamp columns elsewhere in this codebase.

Also rebased onto latest `main` to pick up the native module changes that landed after this branch was cut — that was causing the OTA fingerprint-drift warnings on this PR, unrelated to this diff.

---

marcodejongh:
Re the two "needs attention" items — same two points from the earlier pass, still don't hold up on inspection (see my [previous reply](#issuecomment-4879692970)):

1. **Auth-gating** — `useHomeBoard()` and `useIntegrationStatuses()` both call `useAuth()` internally and pass `enabled: isAuthenticated` to their own `useQuery`/sub-hook calls (`useIntegrationStatuses`: `enabled: isAuthenticated` directly on its `useQuery`; `useHomeBoard`: passes it to the internal `useMyBoards`/`useProfile` calls). Calling them unconditionally from the provider doesn't fire a signed-out network request — there's no 400/401 risk here.
2. **`createdAt` null guard** — `packages/db/src/schema/auth/users.ts:14`: `createdAt: timestamp('created_at').defaultNow().notNull()`. It's a DB-level NOT NULL column, not just a TS-level assertion, so `.toISOString()` can't throw on a real row.

Fixed the one actionable minor note in 27d053d: `has_connected_board` now only fires once per app session (a module-level flag), instead of on every BLE reconnect — cheap fix, avoids the redundant PostHog calls you flagged in a high-reconnect gym session. Added a test for the dedup behavior.

Left the "effect double-fires" note as-is per your own read — it's expected two-stage-load behavior and PostHog's `set` semantics handle a partial payload fine.

---

marcodejongh:
Both fair points:

- **Missing backend tests** — added in 3c9ec27: resolver-level unit tests for `userQueries.profile` and `userMutations.updateProfile` (db client mocked, matching the existing `tester.test.ts` convention in this directory), covering the `createdAt` ISO mapping, `favoriteCount` pass-through, auth gating, and null-row handling on both the insert and update paths. I didn't add a live-DB integration test hitting the real correlated subquery — this sandbox can't pull any Docker images (registry network restriction, confirmed it's not specific to the dev-db image), so I couldn't verify one against a real Postgres myself. CI has full DB access, so it'll validate the SQL, but flagging that gap in case someone wants to add a `rest-parity.test.ts`-style authenticated-request test on top of these later.
- **Web not wired up** — correct, and intentional: this was a scope decision made with the PR requester before implementation (mobile-only per the issue title, with the shared helper as prep for web to adopt later in a separate PR). Not an oversight.

---

marcodejongh:
- **Auth guard** — same claim as the two earlier review passes on this PR; still doesn't hold, for the reasons in my [first](#issuecomment-4879692970) and [second](#issuecomment-4879719737) replies: both `useHomeBoard()` and `useIntegrationStatuses()` internally call `useAuth()` and gate their own `useQuery`/sub-hook network calls on `isAuthenticated` (see `packages/mobile/src/lib/graphql/hooks/use-integrations.ts:36-42` and `use-home-board.ts:46-49`). Nothing fires for a signed-out session.
- **Test naming** — fair, tightened in 27dd801: the test already exercised a disconnect + two reconnects, just didn't say so clearly. Renamed and commented.
- **Web deferral** — as noted already, intentional per the scope agreed before implementation.

---

marcodejongh:
Both fixed in e637056:

- Added an explicit negative assertion (`never sets cohort person properties while signed out`) so a future accidental removal of the auth guard would fail the suite.
- Fixed the `mutations.ts` comment wording — the subquery avoids a separate round trip, not specifically "the second" (there are already several in `updateProfile`).

---

marcodejongh:
1. **Auth guard** — same claim as three earlier passes on this PR; still verified false, see my [first](#issuecomment-4879692970) reply for the specific source lines (`useHomeBoard`/`useIntegrationStatuses` both internally gate their own network calls on `isAuthenticated`).
2. **Double PostHog call while home board resolves** — this is the same point an earlier pass raised (worded as "effect double-fires"), and I left it as-is then: `setPersonProperties` is a `set`/`setOnce` call, safe to fire more than once as fields resolve. Adding `!isResolving` from `useHomeBoard` wouldn't actually eliminate it either — `integrations_connected_count` resolves on its own independent `useIntegrationStatuses` timeline, so the effect could still fire twice even with that guard. Fully deduping would mean threading a loading flag from a second hook just to avoid a harmless extra call to an already-idempotent API — not worth the complexity for this.
3. **Duplicated subquery** — fixed in 613035e: extracted to `FAVORITE_COUNT_SUBQUERY` in a new `favorite-count.ts`, imported by both `queries.ts` and `mutations.ts`.

---

marcodejongh:
1. **Auth guard** — 4th pass raising this; still verified false. To be fully precise: `useIntegrationStatuses` (`packages/mobile/src/lib/graphql/hooks/use-integrations.ts:36-42`) sets `enabled: isAuthenticated` directly on its own `useQuery`. `useHomeBoard` (`use-home-board.ts:46-49`) calls `useAuth()` itself and passes `enabled: isAuthenticated` to the `useMyBoards`/`useProfile` calls it makes internally. Neither hook fires a network request for a signed-out session — not "wasteful but harmless," genuinely zero requests.
2. **`createdAt` null guard** — confirmed for the 3rd time, no "likely" about it: `packages/db/src/schema/auth/users.ts:14` — `createdAt: timestamp('created_at').defaultNow().notNull()`. DB-level NOT NULL.
3. **BLE dedup restart test** — fair, this was only implicitly covered via the file's shared `beforeEach` reset pattern. Added an explicit test in 69aeb60 that fires once, simulates a restart (`vi.resetModules()` + fresh import), and asserts it fires again exactly once in the new instance.

---

marcodejongh:
1 & 2 & 4 are the same points raised in earlier passes on this PR — see the [1st](#issuecomment-4879692970)/[2nd](#issuecomment-4879719737)/[3rd](#issuecomment-4879755327)/[5th](#issuecomment-4879806261) replies for the specific source-line evidence (auth-gating is already internal to both hooks; not fixing the harmless double-fire; the Docker/QA-notes caveat is already called out explicitly in the PR description, not new information).

3. **Fair, fixed in 5d92d43** — added a test that explicitly exercises the partial-then-full firing sequence (integrations resolving after profile/home board are already set), asserting both calls individually instead of only checking the final `toHaveBeenLastCalledWith` state.

---

marcodejongh:
All three addressed in e5f07c5:

- **Test duplication** — extracted `makeAuthMock(overrides)`, replacing all 5 copies of the 10-field auth-state literal in the file (not just the 2 you flagged — the other 3 predated this PR but were the same duplication).
- **Raw SQL comment gap** — added a line to `favorite-count.ts` explaining why: a correlated per-row count from another table can't be a join here without a GROUP BY, which would restructure both callers' single-row selects.
- **Effect fires twice on startup** — leaving as-is, per the judgment call you already noted; it's the same accepted tradeoff from earlier passes, now with explicit test coverage (added last round).

---

marcodejongh:
Thanks — the two remaining notes are already settled from earlier passes: `createdAt` is DB-level `NOT NULL` (`packages/db/src/schema/auth/users.ts:14`, confirmed 3× now), and the cold-launch double-fire is deliberately accepted with explicit test coverage added a few commits back. Nothing further to change here.

---

Issue #2392, opened 2026-05-30
Consolidate user-profile / PartyProfile providers on web and mobile


marcodejongh:
Spun out of [PR #2366](https://github.com/boardsesh/boardsesh/pull/2366#discussion_r3328361176).

## Problem

Each platform currently has two overlapping concepts for "who is this person":

**Web:**
- \`PartyProfileProvider\` (\`packages/web/app/components/party-manager/party-profile-context.tsx\`) — owns the anonymous \`{ id: UUID }\` party identity plus a derived \`username\` / \`avatarUrl\` that prefers the authenticated user profile and falls back to the NextAuth session display name.
- The authenticated user profile itself is fetched ad-hoc inside \`PartyProfileProvider\` via \`fetch('/api/internal/profile')\` and held in local state; there's no dedicated user-profile provider.

**Mobile** (after [#2366](https://github.com/boardsesh/boardsesh/pull/2366)):
- \`PartyProfileProvider\` (\`packages/mobile/src/providers/party-profile-provider.tsx\`) — loads the \`{ id: UUID }\` party identity but surfaces \`username\` / \`avatarUrl\` as \`undefined\` (no user profile fetch wired in yet).
- The authenticated user's name/avatar comes from the GraphQL profile query (\`packages/mobile/src/lib/graphql/operations.ts\`) but isn't exposed through a provider — components call the query directly per screen.

## What \"good\" looks like

A single \`UserProfileProvider\` (or whatever we name it) on each platform that owns:
- The authenticated user's profile (\`{ displayName, avatarUrl, ... }\` from the backend), fetched once and React-Query-cached.
- The party-profile UUID for stable peer identity in the WS party session (today's \`PartyProfile.id\`).
- Derived hooks: \`useUserDisplayName()\`, \`useUserAvatar()\`, \`usePartyPeerId()\`.

Shared package: most of this is already in \`@boardsesh/party-profile\` (the UUID side). The display-profile fetch is GraphQL so the operation is already shared via \`@boardsesh/graphql\`; the React-Query wiring lives per platform.

## Scope

- **Web:** replace \`PartyProfileProvider\`'s ad-hoc \`fetch('/api/internal/profile')\` with a dedicated user-profile React Query layer; keep the UUID side via \`@boardsesh/party-profile\`'s \`ensureProfile\`.
[...]

PR #3443, opened 2026-07-04
chore(mobile): set durable analytics person properties for cohorting

### LINKED ISSUE #3399: analytics(mobile): add durable person properties for cohorting (home board, first-seen, role)

marcodejongh:
## Summary

Mobile only ever set `email` as a PostHog person property, so every tile on the App Success dashboard could be segmented by **behaviour**, never by durable **user traits**. This adds a small null-render component that writes cohort person properties once the authenticated profile resolves:

- `account_created_at` (`$set_once`) — new-vs-veteran retention cohorts
- `home_board` (`$set`) — the active board type (kilter / tension / …), for board-level PMF splits
- `role` (`$set`) — `tester` vs `user`, to separate the beta channel cleanly (today we lean on `filterTestAccounts`)
- `favorite_count` (`$set`) — favourite depth

Notes:
- The property object is built by a pure, unit-tested helper (`analytics-person-properties.ts`), then forwarded to the existing `setPersonProperties(set, setOnce)` wrapper.
- No new fetch: the backend `profile` query already exposes `createdAt` + `favoriteCount`, and `role`/`isTester` is already in `GET_PROFILE`.
- `home_board` reads the cheap active-board store (AsyncStorage) rather than `useHomeBoard()`, which would fan out a logbook-ticks query at app root purely for analytics.
- The optional extras from the issue (has-connected-board, integrations-connected) are deferred — each needs an extra startup query not justified for analytics alone.

Closes #3399

## Release Notes

## Test plan

- `vp run typecheck:mobile` — clean (also ran codegen; no drift)
- `vp run test:mobile` — 3280 tests pass, including the new helper + component tests (`analytics-person-properties.test.ts`, `AnalyticsPersonProperties.test.tsx`)
- `vp lint` on the changed files — 0 warnings, 0 errors
- `vp run check:mobile-bundle` — Android bundle OK (the component mounts at app root)

Analytics is disabled in dev (`isAnalyticsEnabled = !!apiKey && !__DEV__`), so there's no live PostHog write to observe locally; correctness is proven by the unit tests asserting the exact `setPersonProperties` arguments (set/setOnce, role mapping, home_board omitted until a board is picked).

---
_Generated by [Claude Code](https://claude.ai/code/session_0161P3evWdF4fxfKvdVMirtV)_

### DIFF
--- a/packages/mobile/app/_layout.tsx
+++ b/packages/mobile/app/_layout.tsx
@@ -66,6 +66,7 @@ import { reportError } from '../src/lib/error-reporting';
 import { loadRequiredFonts } from '../src/lib/required-fonts';
 import { AnalyticsProvider } from '../src/components/analytics/AnalyticsProvider';
 import { AnalyticsScreenTracker } from '../src/components/analytics/AnalyticsScreenTracker';
+import { AnalyticsPersonProperties } from '../src/components/analytics/AnalyticsPersonProperties';
 import { OtaUpdateTracker } from '../src/components/analytics/OtaUpdateTracker';
 import { InstallReferrerTracker } from '../src/components/analytics/InstallReferrerTracker';
 import { OnboardingGate } from '../src/components/onboarding/OnboardingGate';
@@ -323,6 +324,10 @@ function RootLayout() {
                   <FeatureFlagsProvider flags={STATIC_FEATURE_FLAGS}>
                     <AuthProvider onReady={onAuthReady}>
                       <PartyProfileProvider>
+                        {/* Durable PostHog person properties (account age, home
+                            board, role, favourite depth) for dashboard cohorting.
+                            Null render; needs auth + query, both in scope here. */}
+                        <AnalyticsPersonProperties />
                         <ConnectionSettingsProvider>
                           <ToastProvider>
                             <ClimbActionsDataWrapper>
--- a/packages/mobile/app/_layout.tsx
+++ b/packages/mobile/app/_layout.tsx
@@ -324,9 +324,7 @@ function RootLayout() {
                   <FeatureFlagsProvider flags={STATIC_FEATURE_FLAGS}>
                     <AuthProvider onReady={onAuthReady}>
                       <PartyProfileProvider>
-                        {/* Durable PostHog person properties (account age, home
-                            board, role, favourite depth) for dashboard cohorting.
-                            Null render; needs auth + query, both in scope here. */}
+                        {/* Needs auth + query, both in scope here. Null render. */}
                         <AnalyticsPersonProperties />
                         <ConnectionSettingsProvider>
                           <ToastProvider>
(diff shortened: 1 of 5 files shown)


marcodejongh:
Thanks — addressed in f386950:

- **Verbose comments** — trimmed both files and the `_layout.tsx` mount down to the non-obvious *why* (the `$set`/`$set_once` split, and active-board vs `useHomeBoard`), dropped the what-it-does prose.
- **Missing re-render test** — added one that flips `isTester` + `homeBoard` after the first resolve and asserts the effect re-fires with the updated values.

Kept as-is, with rationale:

- **`favoriteCount` in deps** — intended. `favorite_count` is a *live* cohort trait, and React only re-runs the effect when a dep value actually changes, so the write fires on change, not on every render. A one-time write would freeze a stale count. PostHog `$set` is idempotent, so the extra writes (only when the value moves) are harmless.
- **`@vitest-environment jsdom`** — required. `@testing-library/react`'s `render()` mounts into `document.body`, so it needs a DOM even though the component renders `null` (same pattern as `register-analytics.test.tsx`). Added a one-line note explaining that.

---

marcodejongh:
Fixed 2 & 3 in 8cb1fcc — renamed the profile-not-resolved test to match what it asserts, and added a case for an identified user with a null `createdAt` so the `createdAt` guard has its own coverage.

On 1 (multi-line comments): I don't see a "one short line max" rule in `CLAUDE.md`, and the surrounding mobile code (`party-profile-provider.tsx`, `install-referrer.ts`, `use-active-board.ts`) uses multi-line blocks as the house style. The remaining blocks explain non-obvious *why* (the `$set`/`$set_once` split, active-board vs `useHomeBoard`), so I've left them to match convention. Happy to collapse them if a maintainer points me at the actual rule.

---

marcodejongh:
Added the sign-out guard test in 2c1fd74 (authenticated → signed out asserts no further write).

On the other two:
- **Issue ref in comments** — I don't find that rule in `CLAUDE.md`, and the codebase already references issues in comments (e.g. `party-profile-provider.tsx` points to #2392). The `(issue #3399)` tag gives useful traceability for why these specific properties exist, so I've kept it, consistent with that convention.
- **Double write on cold start** — real but harmless: `$set` is idempotent, and it's at most two writes when the board resolves after the profile. Gating the effect on the board query settling would risk never writing if that query stays pending (a board of `null` is a valid state), so I'd rather keep the guard simple than trade a harmless extra write for that failure mode.

---

PR #3445, opened 2026-07-04
docs: spec to cut the Android PR build CI tail (warm cache + parallel tests)

azuttre:
## Summary

Spec proposing two config-only changes to `.github/workflows/android-pr-rn.yml` to cut the Android PR build from ~21 min typical to ~12.5-13 min. It is the merge-blocking tail on essentially every mobile PR (next-longest check is ~7.5 min), and it already starts at t=0, so duration is the only lever.

The two proposed fixes, with receipts (run ids, step timings, `gh cache list` snapshot) in the doc:

1. **Restore-only Gradle cache, warmed by `main`.** Today each PR run saves a 3.4 GB cache under a PR-only key prefix that can never exist on `main`; the saves blow the 10 GB repo cache budget and evict everything, including the release/dev-client caches on `main` (currently both evicted). Every new PR builds cold. The fix restores the cache `android-apk-rn.yml` already produces on every mobile merge, and stops PR-side saves. The doc addresses the documented reason for the current key separation and why it holds in both directions after the change.
2. **Robolectric tests as a parallel job**, off the critical path (they currently add 2.5-7.5 min serially before `assembleRelease` in the same job).

Deliberately **not** bundled: path-scoping the trigger so JS-only PRs skip the job entirely. That trades away the per-PR sideloadable APK, so it's framed as an open question in the doc for a separate decision.

This PR is the spec only, for review/approval before any workflow change is implemented.

## Release Notes

## Test plan

Docs-only. Numbers re-checkable via the `gh run view` / `gh cache list` commands and run ids cited in the doc.

### DIFF
--- a/docs/android-pr-build-speed-spec.md
+++ b/docs/android-pr-build-speed-spec.md
@@ -221,46 +221,57 @@ catch. This ships in the same implementation PR as fixes 1–2 — it's a one-li
 addition, nothing to defer. (The cache key needs no change: bun records patches in
 `bun.lock`, which is already hashed.)
 
-## Fix 3: skip the native build on JS-only PRs (maintainer decision)
+## Fix 3: skip the native build on JS-only PRs and branch pushes (maintainer decision)
 
-Decided in review (@marcodejongh): JS-only PRs shouldn't build the Android or iOS apps —
-`main` already skips the native build for JS-only merges, and PRs should behave the
-same. A JS-only change can't break the native compile (Metro bundling is separately
-covered by `check:mobile-bundle` in ci.yml), and JS testing rides the `pr-<number>` OTA
-channels, so losing the sideload APK on JS-only PRs is acceptable. This supersedes the
-earlier draft's lean toward keeping the job.
+Decided in review (@marcodejongh): JS-only changes shouldn't build the Android or iOS
+apps — `main` already skips the native build for JS-only merges, and PRs should behave
+the same. His review added a second half: **skip the JS-only branch-push builds too** —
+`ios-rn-ci.yml` builds on every non-`main` branch push, which "only builds on branches
+because I didn't have OTA at the beginning," and no longer needs to for JS-only changes.
+A JS-only change can't break the native compile (Metro bundling is separately covered by
+`check:mobile-bundle` in ci.yml), and JS testing rides the `pr-<number>` OTA channels, so
+losing the sideload APK on JS-only runs is acceptable. This supersedes the earlier
+draft's lean toward keeping the job.
 
 **Mirror `main`'s principle, not its mechanism.** `android-apk-rn.yml`'s gate resolves
 the Expo fingerprint and skips when a `fingerprint-android-<hash>` tag says that native
 shape already shipped. That exact lookup can't run on PRs: the production fingerprint
 depends on Production-environment secrets (`GOOGLE_MAPS_API_KEY` perturbs it), which PR
 runs — fork PRs especially — can't read, so a PR-side resolve would never match the
-shipped tags. The PR-side equivalent that stays env-consistent: **resolve the
-fingerprint twice with the PR workflow's own env — once at the PR head, once at the
-merge-base — and skip when they're equal** (same env on both sides, so env differences
-cancel; only actual native-input changes move the comparison).
+shipped tags. The env-consistent equivalent: **resolve the fingerprint twice with this
+workflow's own env — once at HEAD, once at `main` — and skip when they're equal** (same
+env on both sides, so env differences cancel; only real native-input changes move the
+comparison). Diffing against `main` is uniform for both event types: it's `base.sha` for
+a PR's merge ref and the merge target for a branch push.
 
 Shape, mirroring the release workflow's `gate` → `needs:` pattern:
 
-- A `gate` job (~30 s–2.5 min) decides `should_build`:
-  1. **Path screen (instant):** `git diff --name-only <merge-base>...HEAD`; if nothing
-     matches the native-candidate globs (`app.config.ts`, `plugins/**`, `modules/**`,
-     `packages/mobile/package.json`, root `package.json`, `patches/**`, `bun.lock`),
-     skip without resolving anything. This is the exit ~every JS-only PR takes.
-  2. **Fingerprint diff (only when candidates changed):** resolve head + merge-base
-     fingerprints and compare — this is what correctly skips e.g. a `bun.lock` churn
-     from a web-only dependency bump. Implementation note: the merge-base resolve needs
-     that revision's `node_modules` (config plugins execute at resolve time), so the
-     gate re-runs `bun install` after checking out the base. Fail-open: if either
-     resolve errors, build.
+- A `gate` job (~30 s–2.5 min) decides `should_build`, diffing against a shallow-fetched
+  `main` tip (`git diff <main> HEAD` compares the two trees directly — no merge-base or
+  full history needed, so it works on both a PR merge ref and a branch push):
+  1. **Path screen (fast):** if none of the canonical native-input files changed vs
+     `main` (`packages/mobile/package.json`, `bun.lock`, `app.config.ts`, `plugins/`,
+     `modules/`, `patches/` — the set `ios-rn-ci`'s Pods cache key already hashes), skip
+     without resolving anything. This is the exit ~every JS-only run takes. A change to
+     the workflow/action plumbing forces a build so it's exercised.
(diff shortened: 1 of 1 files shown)


azuttre:
@marcodejongh not my area of expertise so I won't pursue unless you tell me to implement as stated. This is basically just Claude's recommendation on how to cut down CI processing time.

azuttre:
Addressed in 7b2780f:

1. **Outcome table** — restructured by cache state instead of typical/best. The convergence was the point (fix 1 makes warm the norm), but the table didn't say so; now it does.
2. **Duplicate setup block** — spec now commits to extracting the shared setup into a local composite action under `.github/actions/` in the implementation PR.
3. **Run IDs** — every load-bearing number was already reproduced inline (duration table, step-timing table, cache snapshot); the ids are provenance only. Status line now says exactly that, noting logs expire after ~90 days while the inline tables don't.
4. **No tracking issue** — deliberate: per the comment above, implementation is gated on Marco's approval of this spec. If approved, implementation follows as its own PR (or an issue first, maintainer's preference); filing an issue before that decision would presume the outcome.
5. **Open question** — added a stated lean (keep the job on JS-only PRs: post-fix cost is ~12.5 min and the APK remains the only zero-setup path for testers without dev-client/`tester` role, forks included) while leaving the decision explicitly out of this change's scope.

azuttre:
Round 2 addressed in 4413e0f:

- **Required-check rename**: added as step 0 of Validation & rollback. Verified no ruleset-based required checks exist on `main` today (classic branch protection needs an admin eye), so likely a no-op, but it's now an explicit implementation-time check.
- **Composite action skeleton**: Fix 2 now shows the two-job YAML using `./.github/actions/android-rn-setup` plus the action stub — no inputs (workflow-level `env` is inherited, no secrets involved), checkout stays in each job since a local action can't run pre-checkout.
- **Timeout rationale**: inline comment on `timeout-minutes: 20` — ≈2.5× the observed cold-cache duration (7m32s).

azuttre:
Round 3, addressed in the latest commit:

- **`patches/**` trigger gap** — good catch, and it's pre-existing rather than introduced by the fixes: a patch-only PR skips the native build today and the breakage would land on `main`'s release build. Added as a companion one-liner section; cache key needs no change since bun records patches in `bun.lock`, which is already hashed.
- **7-day TTL** — added to Residual risk, with the nuance that every PR restore refreshes the TTL, so eviction takes 7+ days of no mobile activity at all.
- **`TAILSCALE_HOSTS`** — now explicit in the skeleton. One correction to the framing though: nothing fails "across the composite boundary" — the prebuild step moves verbatim *with* its step-level `env`, which travels with the step. The risk was only ever an implementer dropping the line while transcribing, and the worst case is a 2 s probe timeout per job, not a silent correctness issue.
- **`shell: bash` checklist item** — not added: a `run:` step missing `shell:` in a composite action fails loudly at action-load time on the first run ("Required property is missing: shell"), so the existing note in the skeleton is enough. Happy to add it if maintainers disagree.
- **Branch-protection confirmation** — step 0 now says to treat it as a blocking pre-merge item on the implementation PR.

azuttre:
Round 4, addressed in the latest commit:

- **Dev-cache prefix collision** — took the "justify" branch: Gradle's dependency cache is additive-only (entries keyed by module coordinates + checksums; wrong/extra entries are ignored, never mis-built from), both builds resolve the same monorepo graph, and the dev client only adds the `expo-dev-client` modules — so a dev-cache hit is marginally less complete, never incorrect. Narrowing isn't expressible with prefix-only restore-keys, and re-keying the release workflow isn't worth the churn. Spec now says all of this instead of "secondary fallback".
- **`patches/**` bundling** — the spec already committed the implementation PR to adding it ("The implementation PR adds `patches/**`..."); reworded to "ships in the same implementation PR as fixes 1–2" so it can't be read as deferred.
- **Branch-protection evidence** — step 0 now requires pasting the `gh api` output (plus admin confirmation of classic protection) into the implementation PR description.
- **Artifact on cancel** — added a bullet: no artifacts from cancelled runs, identical to today's single-job behaviour.
- **`shell: bash`** — declining again, with a correction: omitting `shell:` on a composite `run:` step is a load-time validation error, not a silent break — and these jobs pin `ubuntu-latest`. The skeleton's first line already states the requirement.

This closes out the automated rounds from our side — remaining disagreements are left for human review to arbitrate.

azuttre:
Closing the last open item from the automated review — **"LRU refresh via restore-only not confirmed"** — with data already collected for this spec: `gh cache list` on 2026-07-03 showed both `gradle-rn-pr-` caches with `lastAccessedAt` well after `createdAt`, updated by later runs *restoring* them (e.g. PR 3436's cache: created 22:31Z, last accessed 23:28Z by the next run's restore step, which is read-only in the `actions/cache` post-step sense — the exact-key hit skipped the save). GitHub's docs agree: both the 7-day TTL and the 10 GB LRU eviction key off last access, and a restore is an access. No spec change needed.

CI is green across all five rounds; every review point is either folded into the doc or answered in this thread. Marking ready for @marcodejongh's review — implementation only proceeds on his approval, per above.

docs/android-pr-build-speed-spec.md
```diff
@@ -0,0 +1,280 @@
...
+
+## Open question (not bundled here): should JS-only PRs run this job at all?
+
+A JS-only mobile change can't break the native Gradle compile, and Metro bundling is
+already covered by `check:mobile-bundle` in ci.yml. Path-scoping the trigger to native
+inputs only (`app.config.ts`, `plugins/**`, `modules/**`, `package.json`, `patches/**`,
+`bun.lock` — the same globs the iOS screenshot cache keys on) would remove the 15–21 min
+job from the majority of mobile PRs entirely.
+
+The trade-off is real, though: JS-only PRs would lose the per-PR sideloadable APK
```
marcodejongh: Js only prs shouldnt build the android or ios apps, we already skip it on main when its js only prs that are merged

marcodejongh (dismissed):
Plan looks good, just want to also skip branch builds when its js only.
It only build on branches because I didnt have ota at the beginning

docs/android-pr-build-speed-spec.md
azuttre: Done in fdd05a4 — the open question is now "Fix 3: skip the native build on JS-only PRs (maintainer decision)", superseding the earlier lean.

One mechanism note: the PR side can't reuse the release gate's fingerprint-tag lookup verbatim — the shipped fingerprints are resolved with Production-env secrets (GOOGLE_MAPS_API_KEY perturbs them) that PR runs, fork PRs especially, can't read. So the spec proposes the env-consistent equivalent: resolve the fingerprint at the PR head and at the merge-base with the PR workflow's own env and skip when equal, with an instant path screen in front so typical JS-only PRs skip in ~30 s without resolving anything. Same gate applied to ios-rn-ci.yml (it runs a full build-for-testing today).

Net effect: most mobile PRs stop paying for native builds entirely (PR tail becomes CI's ~7.5 min); native-input PRs go ~21 min → ~14–15 min via the cache + parallel-test fixes.

azuttre:
Thanks @marcodejongh — both parts are in.

1. **JS-only PRs skip native** (your inline comment): done.
2. **Also skip JS-only branch-push builds** (your review): done — the gate now diffs against `main` for the `ios-rn-ci` `push` trigger too, not just PRs, so a JS-only branch push no longer runs the macOS build. Since it's the same diff-vs-`main` logic for both event types, I dropped `dorny/paths-filter` entirely (plain `git diff`, one mechanism, no extra permission).

Implementation is **#3460** (draft, CI green). Spec updated here (b6875b4) so the doc matches: fix 3 now covers PRs + branch pushes and the diff-vs-`main` mechanism.

One thing that genuinely needs you (admin): `main` is classically protected and I can't read the required-check list from a non-admin token. If `prebuild-and-build` is a required check, the rename in #3460 (`gate` / `robolectric-tests` / `build-apk`, plus a new `gate` on `ios-rn-ci`) needs the branch-protection rules updated or the mobile gate goes stale. Flagged on #3460 too.

marcodejongh:
Merged the implementation, should we throw away this spec branch?

azuttre:
Let's merge it rather than bin it. The gate/setup actions I just merged reference docs/android-pr-build-speed-spec.md in three places, so keeping the doc on main avoids dangling links, and it's the design record for why the gate works the way it does. It's docs-only and green, so no risk. If you'd rather not keep it, I can strip those three comment refs instead, but merging is less work.

PR #3440, opened 2026-07-03
fix(shared): unify French board term to « la board »

azuttre:
## Summary

Standardizes the French term for the board device to **« la board » (feminine)** across every `fr` catalog, and settles the glossary on it.

- Climbing-sense « panneau » (settings BLE section, smart-playlist descriptions, climbs) and « planche » (session, playlists, profile stats) become « board ».
- The ~110 existing masculine usages flip to feminine: articles and determiners (le/un/ce/du/au/aucun/quel/nouveau/dernier/tous → la/une/cette/de la/à la/aucune/quelle/nouvelle/dernière/toutes), possessives (mon/ton → ma/ta), participles and adjectives (« Board perdue », « une board trouvée », « Board publique », « Boards trouvées »), and pronouns (« Redémarrez-la », « Elle existe peut-être déjà »).
- `docs/i18n-french-glossary.md`: the "Known inconsistency" section becomes the settled rule — « la board », feminine, as French board climbers say it and French climbing media writes it (La Fabrique Verticale: « la board », « une Kilterboard »; the gender patterns with « la planche », like skate and snowboard French).

Untouched by design: « Panneau d'administration » (admin-panel sense, admin.json), « panneaux » on the marketing legal page (translates "panels", a distinct physical product in « prises, boards, panneaux »), brand names (Kilter Board, Tension Board, MoonBoard — and « le Kilter Homewall » stays masculine, it reads as a wall), en-US/es catalogs, JSON keys and ICU placeholders.

Two flips worth a reviewer's eye: admin.json « Global (tous les boards) » → « toutes les boards » and you.json « Tous les boards » → « Toutes les boards » — device-sense strings, so they follow the rule even though the original sweep brief said to leave admin.json alone (that instruction targeted the admin-panel « panneau » strings, which are preserved).

Originally stacked on #3438 (which owned the glossary and ~90 overlapping fr strings); **now rebased onto `main`** after its merge. The rebase also picked up strings the gym/board-disambiguation features introduced on `main` in the meantime — a third device term « tableau » (« Ton tableau », « {{count}} tableaux »), new masculine « ce board » strings, and two « mur » mistranslations where the English says *board* (`otherBoardsTitle`, `multiBoardChip`) — all swept into the same rule. Genuine wall-sense « mur » (« Suis le mur où tu grimpes ») and dashboard-sense « tableau de bord » are untouched.

[...]

PR #3449, opened 2026-07-04
fix(i18n): pluralize mobile.gyms.boardCount so one board reads correctly

azuttre:
## Summary

The gym list showed "1 boards" when a gym has a single board: `mobile.gyms.boardCount` was a bare `"{{count}} boards"` string with no plural variants. This splits it into i18next v4 plural forms (`boardCount_one` / `boardCount_other`) in all three locales:

- **en-US**: "{{count}} board" / "{{count}} boards"
- **es**: "{{count}} plafón" / "{{count}} plafones" (per the Spanish glossary)
- **fr**: "{{count}} board" / "{{count}} boards" — deliberately uses « board » per the terminology flip in flight (#3440) instead of the outgoing « tableaux »; expect a trivial conflict on this line whichever PR merges second.

No component change needed: the only consumer ([gyms/index.tsx](packages/mobile/app/gyms/index.tsx)) already calls `t('mobile.gyms.boardCount', { count })`, and the mobile i18next config runs `compatibilityJSON: 'v4'` with the `intl-pluralrules` polyfill, so the suffixed keys resolve automatically.

Audited the same files for other bare count keys: the only other `{{count}}` key in `boards.json` is `loadMore` ("Load more ({{count}} of {{total}})"), which doesn't inflect — left bare on purpose.

## Release Notes

- Gyms with a single board now say "1 board", not "1 boards" — fixed in English, Spanish, and French

## Test plan

- `vp test run packages/shared/i18n/src/__tests__/catalog-completeness.test.ts` — 28/28 pass (strict key parity across es/fr, including the new suffixed keys)
- `vp run check:i18n:orphans` — passes (checker maps `_one`/`_other` back to the base key)
- `vp check` — no new issues from this diff (3 pre-existing `TS2741` errors in `scripts/__tests__/with-screenshot-dev-menu.test.ts` reproduce identically on the base commit with a fresh install; local `@types/node` artifact, unrelated)


azuttre:
Good catch on the same-screen mismatch — but the fix goes the other way. Per the updated French glossary direction (feminine « la board », corpus-wide flip in #3440), « tableau » is the outgoing term, so reverting `boardCount` to « tableau »/« tableaux » would regress it. Pushed 6f0b7c5 instead: the whole `mobile.gyms` block in `fr/boards.json` now uses « board » (« Aucune board ici pour l'instant », « sa board », « Boards à proximité », « 2+ types de board », « Modifier la board »), so the gyms screen is internally consistent. The rest of the file is left for #3440's sweep.

azuttre:
On the `searchPlaceholder` note: kept deliberately. en-US reads "Search a city, gym, or board", so « ou un mur » was a stray the #3440 sweep missed — it's the same gyms-screen terminology alignment this PR's second commit was about, not new scope.

PR #3453, opened 2026-07-04
feat(mobile): add an iPad "On the Wall" tab and sheet the wall panel on small iPads

marcodejongh:
## Summary

Two iPad improvements, both reusing the existing board-presence system:

1. **Small iPads shed the cramped wall panel.** On iPads smaller than the 11" Pro (mini, base 10.9/11", Air 11") the persistent "Now on the wall" side panel is dropped; the wall reaches you through the bottom sheet + the new tab, exactly like the phone. The 11" Pro and every 13" keep the panel. It's a new launch-fixed device-size axis (`resolveWallDeviceClass`, screen long side ≥ 1194pt) composed as a pure AND-narrowing gate over the untouched window-width logic in `size-class.ts`, so phones and Split View / Stage Manager are provably unregressed (the gate can only downgrade the wall surface, never widen it).

2. **A new iPad-only "On the Wall" tab.** A sidebar destination — registered as an `href: null` tab route so it's hidden from every phone tab bar (Material bar filters the resulting `display:none` item; the native glass bar gets a `hidden` trigger) — showing what's lit on the board right now, the session stats, a "This session" leaderboard, and the history feed. It adapts to **portrait** (single column) and **landscape** (focal board art + scrolling list) from its **own measured width** (not the window, not an orientation API), so a narrow Split View window falls back to single column. The board art contain-fits across the full 0.43–1.81 board aspect-ratio range, so a tall MoonBoard and a wide Kilter both look right.

Design was set by an HIG design panel; the small-iPad threshold (1194 vs 1180), iPad-only scope, leaderboard shape, and play-pane behaviour were confirmed with the maintainer. `NowOnTheWallPanel` gained a `variant="screen"` and the "This session" leaderboard is a pure client-side selector (`session-ranking.ts`) ranked by send count then recency — the grade story stays the backend hardest-send crown, since the presence feed has no reliable client-side grade ordering.

## Release Notes

- See what's lit on the wall from your iPad — a new On the Wall tab with the current climb, who's crushing it, the session's hardest send, and recent history, in portrait or landscape
- Smaller iPads now open the board history as a sheet, so the browse list keeps the full screen

## Test plan

- `vp run test:mobile` — 3310 pass. New unit tests: `size-class` (device-class + effective-surface boundaries), `session-ranking` (grouping/tie-break/limit), `WallScreen` (empty vs live, portrait/landscape at 700), `MaterialTabBar` (hidden `href:null` route filtered). Updated `tab-layout` + `IpadPlayPane` for the device gate and wall-tab suppression.
- `@boardsesh/i18n` catalog parity — 28 pass (new `common.mobile.nav.wall` + `session.mobile.boardPresence.*` keys in en-US / es / fr; Spanish uses "muro" for the wall to match the shipped board-presence copy, French "croix"/"allumer")
- `vp check` — lint/format clean
- `vp run check:mobile-bundle` — iOS + Android bundles OK (new `/wall` route + `variant="screen"`)
- Adversarial multi-agent review of the diff — 2 low findings, both fixed (single-send plural grammar; a pre-measure focal frame that briefly read "dark" while a climb was lit)

**Still needs on-device visual QA** (invisible to vitest): the `/wall` tab in portrait/landscape on 11" Pro vs 13", the sheet-vs-panel split on mini / base 11" / Air 11", contain-fit on the widest (1.81) and tallest (0.43) boards, and a Split View resize on the wall tab degrading to single column.

https://claude.ai/code/session_012vj28m82VgJhvszfnJ8NME

### DIFF
--- a/docs/mobile-sheets-vs-routes.md
+++ b/docs/mobile-sheets-vs-routes.md
@@ -170,6 +170,26 @@ waits out the whole transition and is disabled in screenshot mode.
   boards → pushed routes (rule 3); setters is a searchable list route.
 - **Onboarding** — immersive cover, not over the live tab bar → `fullScreenModal`.
 
+## iPad-only tab destinations (sidebar rail, never a phone tab)
+
+To add a `(tabs)` destination that appears only on the iPad sidebar rail and NEVER as a phone
+bottom tab (e.g. `/wall`, the "On the Wall" tab — `app/(tabs)/wall/`):
+
+- Register it as a keyed `<Tabs.Screen name="wall" options={{ href: null }} />` in the shared
+  `tabScreens` array (`app/(tabs)/_layout.tsx`). It **must** be a `(tabs)` route so it renders in
+  the iPad shell's content pane (keeping the sidebar + play/wall panes); a root route would cover
+  them.
+- `href: null` is meant to hide it from the bar — but expo-router turns it into
+  `tabBarItemStyle: { display: 'none' }` and **strips the `href` key**, so a _custom_ tab bar still
+  renders it unless it filters. `MaterialTabBar` skips
+  `StyleSheet.flatten(options.tabBarItemStyle)?.display === 'none'`.
+- Add `<NativeTabs.Trigger name="wall" hidden />` (the `hidden` prop) so the iOS-26 glass bar
+  declares the route but shows no 6th tab (a 6th would spill into "More" and clash with the
+  `role="search"` slot).
+- Add a `SidebarDestination` in `IpadSidebar.tsx`; `tabsActiveSegment` (route-segments.ts) already
+  highlights it. Suppress any ambient duplicate of the same content (e.g. the wall column) while
+  the destination is the focused segment.
+
 ## See also
 
 - `docs/react-native-performance.md` — list/provider/gesture performance rules.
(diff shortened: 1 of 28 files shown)


marcodejongh:
Addressed both automated review passes:

**Pass 1** (all fixed in `8d23889`):
1. WallScreen — bound-but-config-resolving now renders a neutral blank, not the "Connect a board" CTA (+ test).
2. IpadPlayPane — split the inline `useSegments()` into the explicit two-line form.
3. Added the `boardPanelProps=null while live` test case.
4. Documented the iPad-sidebar-only tab pattern in `docs/mobile-sheets-vs-routes.md`.

**Pass 2** (in `fefb327`):
- **setIds parsing hardened** — an empty `setIds` string produced `[0]` (`Number('')===0` passes the finite filter), slipping past the `length===0` guard. Now drops empty tokens before `Number()`.
- **Raw `gap:2` / `paddingTop:2` kept, intentionally** — the spacing scale has no 2px token (smallest is `spacing[1]=4`), and the whole board-presence feature (`NowOnTheWallPanel`) uses raw `2` for micro-gaps; a token would double the gap and break consistency.

Still pending (can't run from a Linux env): on-device visual QA of the portrait/landscape layouts and board-art contain-fit on the widest (1.81) and tallest (0.43) boards.

PR #3460, opened 2026-07-04
ci(mobile): gate PR native builds + restore-only Gradle cache + parallel jobs

azuttre:
## Summary

Implements the CI speedup spec from #3445 (design doc + full review thread there). Three fixes to the Android PR build, plus the iOS gate Marco asked for. All config; no app code.

**1. Restore-only Gradle cache** (`android-rn-setup` composite). PR runs stop saving their own 3.4 GB cache under a PR-only key that never existed on `main`; they now restore the `gradle-rn-` cache the release/dev-client builds already warm on every mobile merge. This ends the eviction cascade that was blowing the 10 GB repo budget and forcing every PR to build cold.

**2. Parallel jobs.** The single `prebuild-and-build` job becomes `robolectric-tests` + `build-apk`, running in parallel and sharing setup via the new `.github/actions/android-rn-setup` composite (no `shell:`-less steps; workflow-level `env` inherited, no secrets).

**3. Skip native builds on JS-only PRs** (Marco's call — mirrors `main`'s native-skip for JS-only merges). New `.github/actions/mobile-native-gate` composite, used by a cheap ubuntu `gate` job in **both** `android-pr-rn.yml` and `ios-rn-ci.yml`:
- **Path screen** (plain `git diff` against a shallow `main` tip — one mechanism for PRs and branch pushes, no extra permission) over the repo's canonical native-input set (the `ios-rn-ci` Pods cache key). No native-input change → skip, no resolve.
- **Fingerprint diff** when those files changed (e.g. `bun.lock` churn from a web dep bump): resolve the Expo fingerprint at head and base under the PR's own env and skip when equal. `main`'s fingerprint-**tag** lookup can't run here — it uses Production secrets a fork PR can't read — so head-vs-base under identical env is the equivalent. **Fail-open**: any resolve/checkout trouble builds, so a gate bug costs time, never correctness.
- The iOS gate runs on **ubuntu** (fingerprint resolve is file-hash based, no Xcode), so skipping spends zero macOS minutes.

**Companion fix:** added `patches/**` to `android-pr-rn.yml`'s trigger paths — it was in the release workflow but missing here, so a patch-only PR skipped the native check entirely.

### Expected effect

- JS-only mobile PRs (the majority): native builds **skip** (~30 s gate) instead of ~21 min.
- Native-input PRs: ~21 min → **~14–15 min** (warm cache + ~2 min gate).

## Blocking pre-merge item

The job renamed from `prebuild-and-build` to `gate` / `robolectric-tests` / `build-apk`, and `ios-rn-ci` gained `gate`. If any branch-protection rule requires the old names, it must be updated to the new ones or the gate goes stale. As of 2026-07-03 no ruleset-based required checks exist on `main` (`gh api repos/boardsesh/boardsesh/rules/branches/main` is empty); classic branch protection needs an admin to confirm before merge.

## Test plan

Validated locally: YAML parses; every composite `run` step has `shell: bash`; output wiring (`gate` → `should_build`) checks out; the `mobile-ci-env-parity` test doesn't cover these two workflows, so moving their env to workflow level is out of its scope. `actionlint` wasn't available in the sandbox — relying on GitHub's own workflow validation on this PR's run.

Gate behaviour to confirm on CI here and with follow-up test PRs:
- This PR touches CI plumbing → gate's `ci` filter forces a build (exercises the new jobs end-to-end).
- A JS-only mobile PR → path screen skips both Android jobs + the iOS build (~30 s).
- A web-only dep bump churning `bun.lock` → path screen escalates, fingerprint diff comes back equal → skip.

Depends on #3445 for `docs/android-pr-build-speed-spec.md` (referenced in the action comments); the doc merges with that PR.

## Release Notes

none (internal CI change)

### DIFF
--- a/.github/actions/android-rn-setup/action.yml
+++ b/.github/actions/android-rn-setup/action.yml
@@ -1,0 +1,71 @@
+name: Android RN setup
+description: >
+  Shared setup for the Android PR jobs — workspace deps, JDK 21, the Android SDK,
+  a read-only Gradle cache, and `expo prebuild`. Factored out so the parallel
+  robolectric-tests and build-apk jobs can't drift. See
+  docs/android-pr-build-speed-spec.md.
+
+# No inputs: the EXPO_PUBLIC_* values the .env step reads are workflow-level `env`
+# on the caller, which composite `run` steps inherit as process environment. No
+# secrets are involved.
+runs:
+  using: composite
+  steps:
+    - uses: oven-sh/setup-bun@v2
+
+    - name: Install Node.js dependencies (workspace root)
+      shell: bash
+      run: bun install --frozen-lockfile
+
+    - name: Setup Java
+      uses: actions/setup-java@v4
+      with:
+        distribution: 'temurin'
+        java-version: '21'
+
+    - name: Setup Android SDK
+      uses: android-actions/setup-android@v3
+
+    - name: Accept licenses and install SDK components
+      shell: bash
+      run: |
+        yes | sdkmanager --licenses > /dev/null 2>&1 || true
+        sdkmanager "platforms;android-36" "build-tools;36.0.0"
+
+    - name: Restore Gradle cache (read-only, warmed by android-apk-rn.yml on main)
+      # restore, not cache: PR runs never save. Each save was a 3.4 GB entry that
+      # blew the 10 GB repo cache budget and evicted main's release/dev caches, so
+      # every new PR built cold. Reusing main's `gradle-rn-` cache (prefix also
+      # matches the dev-client `gradle-rn-dev-` cache — a harmless, additive
+      # dependency-cache hit) means a warm build without the eviction cascade.
+      uses: actions/cache/restore@v4
+      with:
+        path: |
+          ~/.gradle/caches
+          ~/.gradle/wrapper
+        key: ${{ runner.os }}-gradle-rn-${{ hashFiles('packages/mobile/package.json', 'bun.lock') }}
+        restore-keys: |
+          ${{ runner.os }}-gradle-rn-
+
+    - name: Write .env for Expo build-time inlining
+      shell: bash
+      working-directory: packages/mobile
+      run: |
+        cat > .env <<EOF
+        EXPO_PUBLIC_BACKEND_URL=$EXPO_PUBLIC_BACKEND_URL
+        EXPO_PUBLIC_WS_URL=$EXPO_PUBLIC_WS_URL
+        EXPO_PUBLIC_WEB_URL=$EXPO_PUBLIC_WEB_URL
+        EXPO_PUBLIC_GOOGLE_IOS_CLIENT_ID=$EXPO_PUBLIC_GOOGLE_IOS_CLIENT_ID
+        EXPO_PUBLIC_GOOGLE_ANDROID_CLIENT_ID=$EXPO_PUBLIC_GOOGLE_ANDROID_CLIENT_ID
+        EXPO_PUBLIC_GOOGLE_WEB_CLIENT_ID=$EXPO_PUBLIC_GOOGLE_WEB_CLIENT_ID
+        EOF
+
+    - name: Expo prebuild (generate android/)
+      shell: bash
+      working-directory: packages/mobile
+      env:
+        # Skip the tailscale probe (already short-circuited on CI, set explicitly
+        # so a misconfigured runner never pays the 2s timeout). Step-level env
+        # travels with the step into the composite.
+        TAILSCALE_HOSTS: ''
+      run: bunx expo prebuild --platform android --clean --no-install
--- a/.github/actions/android-rn-setup/action.yml
+++ b/.github/actions/android-rn-setup/action.yml
@@ -3,7 +3,7 @@ description: >
   Shared setup for the Android PR jobs — workspace deps, JDK 21, the Android SDK,
   a read-only Gradle cache, and `expo prebuild`. Factored out so the parallel
   robolectric-tests and build-apk jobs can't drift. See
-  docs/android-pr-build-speed-spec.md.
+  docs/android-pr-build-speed-spec.md (added in #3445).
 
 # No inputs: the EXPO_PUBLIC_* values the .env step reads are workflow-level `env`
 # on the caller, which composite `run` steps inherit as process environment. No
--- a/.github/actions/mobile-native-gate/action.yml
+++ b/.github/actions/mobile-native-gate/action.yml
@@ -115,11 +115,13 @@ runs:
         # so there's no ordering footgun for any future step, and the base's node
         # tree can't collide with head's. bun's global cache (~/.bun) is shared,
         # so the base install reuses already-downloaded packages.
-        BASE_DIR="${RUNNER_TEMP:-/tmp}/mobile-native-gate-base-$RANDOM"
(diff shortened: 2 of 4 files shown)


azuttre:
Addressed in 960d0773b:

**1. Working tree left at base SHA (ordering footgun)** — good call; adopted your temp-worktree alternative. The base fingerprint now resolves in an isolated `git worktree add --detach`, so the primary checkout is never mutated. That removes both the footgun (no future step can run against the base SHA) and the related subtlety that checking out base in-place made this action's own files vanish mid-run. Fail-open preserved: if the worktree can't be created, it builds.

**2. Dead doc link** — noted provenance in all three references (`docs/android-pr-build-speed-spec.md (added in #3445)`). Can't force merge order across two open PRs, but this keeps the pointer meaningful either way.

**3. Second `bun install` cold** — acknowledged, no code change. It's less cold than it looks: bun's global cache (`~/.bun/install/cache`) is shared within the job, so the base install re-uses already-downloaded packages and only repeats extraction/linking, not the network fetch. And it only runs on native-input PRs, which pay the full build regardless. If it shows up as a real cost in practice, the follow-up is to cache the resolved base fingerprint keyed on the base SHA.

Note on coverage: this PR trips the gate's `ci` filter (it edits the workflow/action files), so the three gate jobs took the ~12s force-build path and the fingerprint diff itself wasn't exercised here. I'll verify that path with a bun.lock-churn test PR once this is closer to merge.

azuttre:
**Blocker — branch protection.** Verified what I can without admin, and you're right that it needs an admin:
- `gh api repos/boardsesh/boardsesh/rulesets` and `.../rules/branches/main` are both `[]` — no ruleset-based required checks.
- But `repos/boardsesh/boardsesh/branches/main` reports `"protected": true`, and the classic-protection endpoints return 404 for my token (`admin: false`), which is the permission-masked response, not "unprotected". So `main` **is** protected by classic branch protection whose required-check list I can't read.

Net: this genuinely can't be cleared from a non-admin session. **@marcodejongh** — before merge, in Settings → Branches (or `gh api .../branches/main/protection/required_status_checks` as an admin), confirm whether `prebuild-and-build` is a required check. If it is, swap it for `gate` / `robolectric-tests` / `build-apk` (Android) and add `gate` (iOS), or the mobile PR gate goes stale. Leaving this PR as draft until that's confirmed.

**Minor — `$RANDOM` temp dir.** Fixed in bba15a2d9: switched to `mktemp -d` (collision-proof) with a `wt` subdir for `git worktree add`.

**Minor — no bun cache in the gate.** Declining, with reasoning: bun's global cache (`~/.bun/install/cache`) is shared across both installs in the job, so only the head install is cold — the base-worktree install re-uses the downloaded packages and just re-links, so it's not "cold twice." Adding a cache *save* here would also re-introduce exactly the cache-budget pressure fix 1 exists to remove. And the fingerprint diff only runs on native-input PRs, which pay the ~14 min build anyway — a ~1 min gate delta there is noise. If the gate proves slow in practice, the better follow-up is to memoise the base fingerprint keyed on the base SHA rather than cache node_modules.

azuttre:
Follow-up from Marco's review on #3445: he asked to **also skip JS-only branch-push builds**, not just PRs (`ios-rn-ci`'s `push` trigger only exists for pre-OTA reasons). Landed in 107dc8c8b:

- Gate now diffs against `main` for both `pull_request` and `push` events, instead of force-building on any non-PR event.
- Unified on a plain `git diff` path screen (two-tree diff vs a shallow `main`), which let me **drop `dorny/paths-filter`** and its `pull-requests: read` permission — one mechanism for both event types.
- Base fingerprint resolve still runs in an isolated `git worktree` (never mutates the primary checkout); a base `bun install` failure now fails-open explicitly (the spec-review nit).

Note: this PR trips the `ci` filter (it edits the workflow/action files), so CI here still exercises the force-build path, not the fingerprint diff. I'll validate the diff + the new push-skip with a `bun.lock`-churn branch as this nears merge.

azuttre:
Thanks — took three of the four, landed in 5b5fce194:

- **bun-on-PATH precondition**: fixed properly by moving `oven-sh/setup-bun` *into* the `mobile-native-gate` composite (now self-contained, symmetric with `android-rn-setup`); removed it from both gate jobs. A future caller can't hit the cryptic "bun not found" anymore.
- **`--force` on `git worktree add`**: dropped — the `mktemp` path is unique, so it was only masking real misconfig errors.
- **`set -uo pipefail` without `set -e`**: declining `set -e` deliberately, but documented the invariant. The gate is fail-**open** by design — an unexpected error should *build*, never block a PR — and `set -e` would flip an unhandled failure into a red (blocking) gate, the opposite of the intent. Every decision path already exits via `build()`/`skip()`, so `should_build` is always written; added a comment so a future edit keeps that invariant. (Making the default fail-*open* on a hypothetical fall-through would mean relying on GITHUB_OUTPUT duplicate-key last-wins, which isn't documented, so I'd rather keep the explicit-exit invariant.)

Not changing:
- **Hardcoded Google client IDs**: these are public client-side values (already in the app source and the release workflows); they're at workflow level so the gate's fingerprint resolve and the build read an identical set. No secret exposure.
- **Doc reference**: already tagged `(added in #3445)`; comment-only, resolves when either PR merges.

Validated on the prior run before this push: `robolectric-tests` and iOS `build-and-test` both green on the diff-vs-main gate. Branch protection remains the one admin-only pre-merge item.

azuttre:
Added `--no-daemon --console=plain` to both gradle invocations in d2984927 — matches `android-apk-rn.yml`. (This was a pre-existing gap: main's `android-pr-rn.yml` never had them; worth fixing while I'm splitting the job.)

One correction on the rationale: `robolectric-tests` and `build-apk` are separate **jobs**, so they get separate runner VMs and never share a Gradle daemon — the parallel-flake concern doesn't apply here. The real value is consistency with production and not leaving a daemon on the ephemeral runner, so the flags are still worth having.

On the doc reference: already tagged `(added in #3445)`; it's comment-only and resolves whichever way the two PRs merge. Preferred order is #3445 (spec) first, then this.

Thanks for the thorough pass — I believe that closes every code-level item; branch protection stays the one admin-only pre-merge step.

azuttre:
Checked each of these:

- **`patches/**` in `ios-rn-ci.yml`** — already present, lines 12 and 26 (both the `pull_request` and `push` trigger blocks). Not a gap; no change needed. Only `android-pr-rn.yml` was missing it, which this PR fixes.
- **`dorny` reference** — the action's own description doesn't mention dorny (grep of `.github/` finds dorny only in the pre-existing `ci.yml`). The stale reference was in the **PR body**, now updated to describe the `git diff` path screen.
- **`jq` implicit dependency** — declining a code change: the gate jobs pin `ubuntu-latest`, where `jq` and `git` are always pre-installed, so it can't fail in this workflow. A "future porter to a non-ubuntu runner" is hypothetical, and the composite already hard-depends on `bun`/`git` too. Not worth another CI cycle over a comment.
- **env-parity coverage** — agreed, tracked as a follow-up (noted in the test plan); not blocking.

That leaves branch protection as the sole real pre-merge item, which is admin-only. I'm going to treat further doc-level nitpicks as diminishing returns — the substance is settled and green — and let the current run finish so `build-apk` validates on the final commit.

marcodejongh:
Claudes just being pedantic on the review, merged!

PR #3462, opened 2026-07-05
feat(backend): watch/session HTTP control endpoints for Garmin app

marcodejongh:
## Summary

Phase A of the **Boardsesh Garmin watch app**: the backend surface a Garmin Connect IQ watch needs to drive a session and pair to a user account. The watch is a non-WebSocket HTTP client, so these are stateless HTTP endpoints (the watch polls; a phone in the same session repaints the board over BLE — no phone changes needed).

New endpoints (all reusing existing, hardened primitives):

- **`POST /api/session/navigate`** and **`POST /api/session/take-control`** — mobile-JWT counterparts of the iOS lock-screen widget's `/api/widget/*`. They share the server-authoritative queue math (extracted to `session-queue-actions.ts`), the durable-participant + not-ended guard (`verifyWidgetSession`), and the per-session rate-limit bucket. A navigate publishes `CurrentClimbChanged`, which the phone reacts to and sends to the board.
- **`GET /api/session/state`** — a slim, poll-friendly current-climb snapshot (no heavy queue/`frames` payload — the watch has kilobytes of memory). Board resolution (`layoutId/sizeId/setIds/angle`) is parsed from the session's `boardPath` server-side so the watch can build a `saveTick` without a second call.
- **`POST /api/watch/pair-code`** (authed) + **`POST /api/watch/pair`** — a short, single-use, Redis-backed pairing code the phone/web app shows and the user types on the watch, exchanged for the standard mobile token pair (`generateTokenPair`).

The iOS widget `navigate`/`take-control` handlers were refactored onto the shared core; behaviour is identical and their existing test suites (unchanged) pass.

Ascent logging needs no new endpoint — the watch calls the existing `saveTick` GraphQL mutation over HTTP with its JWT.

This is the foundation for Phase B (the "Pair a Garmin watch" UI on phone/web) and Phase C (the Connect IQ watch app).

## Release Notes

## Test plan

- `vp run typecheck:backend` — clean.
- `vp check` — clean (lint + format) on all changed files.
- New vitest suites (66 cases): `session-navigate.test.ts`, `session-state.test.ts`, `watch-pair.test.ts` cover auth (JWE + JWS), IP + per-session rate limits, the durable-participant / ended-session guard, next/prev wraparound + out-of-bounds, the slim payload + `boardPath` resolution, and pairing-code issue / single-use consume / replay / collision / expiry.
- Existing `widget-navigate` / `widget-take-control` suites pass unchanged, confirming the shared-core refactor is behaviour-preserving.
- An adversarial multi-agent review of the diff surfaced and this PR fixes: (1) a rate-limit **poisoning** bug — the participant guard now runs before the shared bucket so an authenticated non-participant can't 429 real members (regression test added); (2) a redundant `getSessionById` on every `state` poll (the guard now returns the loaded row).

https://claude.ai/code/session_01P7DUnfoBMdxnrArfhaCzVK


marcodejongh:
Thanks — addressed all three in `628bddf`:

- **Auth before body parse**: `/api/session/navigate` + `/api/session/take-control` now authenticate before reading/parsing the body. (The mobile JWT is body-independent, unlike the widget's session-bound `authenticateWidget`, so this is strictly safe here.)
- **Poll rate limit**: added a separate per-**user** read bucket for `GET /api/session/state` (capacity 4, refill 1/s) — legit 3s polling passes, hammering is throttled, and it's kept distinct from the write bucket so read spam can't drain the navigation limiter. New regression test covers the 429.
- **Headers**: `sendJson` on both handlers now sets `Cache-Control: no-store` + `X-Content-Type-Options: nosniff`, matching the native-auth helper.

Backend typecheck + 67 tests green.

marcodejongh:
Also ran a **Fable-model** review over the backend changes. Fixed in `2cb9f20`:
- **Named-board sessions** (`/b/{slug}`) log-spammed a "malformed boardPath" warning every ~3s poll (parseBoardPath rejects that shape by design) → only warn for genuinely-unparseable paths.
- **No throttle in front of the write guard's DB queries**: a non-participant 403s *at* the guard (before the per-session bucket), so they could spin its queries unbounded → added a per-user front limiter (generalized the read one) on navigate/take-control/state.
- **Pair test gaps**: added the IP rate-limit (11th → 429) case and an assertion that the minted JWT's `sub` matches the user the code was bound to.

PR #3464, opened 2026-07-05
feat(mobile): offline mode — download a board, browse and log with no signal

marcodejongh:
## Summary

Stacked on `fix/2639-offline-sync` (base). That PR landed the offline **engine** but its control UI was dropped in the rebase; this PR reintroduces the offline UI in **My Boards** and makes "offline available" actually browsable.

- **Scope downloads by board type + layout + size, all sets.** `syncClimbs`/`syncClimbStats` gain optional `layoutId`/`sizeId` args (stats scope via a correlated `board_climbs` EXISTS; `sizeId` ignored for moonboard). A downloaded board is a fixed, cacheable superset — Kilter is ~424k climbs across 19 layouts, so per-type downloads were impractical. `syncEnabledBoards` now holds `"boardType:layoutId:sizeId"` scope keys; the `characteristics` column is synced (schema v2).
- **My Boards offline toggle.** Each board row gets a compact cloud/spinner/check control (off · downloading N · available offline). The screen subscribes to sync status once and passes each row a primitive state, so a download's progress frames only re-render the one changing row.
- **Offline browse.** When offline and the active board's exact (type, layout, size) scope is downloaded, climb search + count + detail read local SQLite (mirroring the server LEFT-JOIN search: `json_each` for the JSON set/size arrays with the NULL-subset trap + moonboard skip, `CAST(ROUND())` grades, all six sorts + uuid tiebreak, present-hold matching, personal progress from local ticks). A coarse `offline` flag in the query key swaps network⇄local on a connectivity flip. Filters needing un-synced tables (hold-state, zone, tall/wide, beta) and the drafts path fall back to the network; the grade rail falls back to bundled grades cold-offline.
- **Offline writes never dead-letter.** The drainer skips while offline and, if the connection drops mid-drain, leaves the write pending without advancing `retry_count`. A **Sync issues** section (More screen, online + stuck writes only) offers Retry — never discard; sign-out warns before dropping unsynced writes. Sign-out also preserves board checkpoints (no 200k-row re-crawl) and resets `syncEnabledBoards` (no cross-user leak).

## Release Notes

- Download a board for offline sessions — browse climbs and log your sends with no signal, then everything syncs when you're back online.

## Test plan

- `vp run typecheck:mobile` + `typecheck:backend` — green.
- `vp run test:mobile` — 437 files / 3505 tests green, incl. new local-search parity suite (json_each set/size, NULL-subset trap, moonboard skip, grade range, 6 sorts + tiebreak, pagination/count, personal progress, present-holds, field mapping), offline-board-key, board-offline-state, drainer offline-gating.
- `vp test run --project backend sync-pull` — 15 green (layout/size scoping incl. the stats EXISTS join + cursor).
- `vp run check:mobile-bundle` — OK. `vp check` — only pre-existing `scripts/` errors remain; catalog-completeness (i18n parity, all 3 locales) green.
- Device E2E (to run): enable a board offline in My Boards → watch it download → airplane mode → search/browse that board → log a tick offline → reconnect → tick syncs, no dead-letter. Force a 4xx write online → Sync issues Retry appears (no discard).

https://claude.ai/code/session_01Ef4Vrvh9B8JE2Z9zbRfTWS


marcodejongh:
Thanks — good catches. Addressed in `3457d63`.

### Fixed
1. **`isBoardDownloadedLocally` ignored `sizeId`** (the real bug) — the row probe now checks `compatible_size_ids` via `json_each` (skipped for moonboard), matching the search-side size filter, so a board downloaded at size 5 no longer gates offline search for size 15 of the same layout. Added `board-download-status.test.ts` incl. the multi-size-same-layout case.
2. **`boardDownloadState` used global `lastSyncedAt`** — now driven by a **per-scope** checkpoint (`getDownloadedScopeKeys`, refetched when a cycle ends), so a board flips to "Available offline" only once its own data lands.
3. **Retry loop aborted on first failure** — each dead letter is retried in its own `try/catch` now.
4. **`GRADE_MAP` duplicated** — derived from the shared `BOULDER_GRADES` taxonomy instead of a hardcoded copy.
5. **`useGrades` offline key** — dropped the `offline` discriminator (grades are static) and use `placeholderData` for the bundled fallback, so a connectivity flip no longer forces a refetch. (The *search* hooks keep the `offline` key on purpose — their results genuinely differ network vs local.)
6. **`getTableColumns`** — runtime allowlist assert added; `schema` typed `'seed'`.
7. **`gradeAccuracy`** — `Number()` instead of `parseFloat` (rejects `"1.5abc"`).

### Intentional / not changing (with reasoning)
- **Hold-presence `LIKE '%p<id>r%'`** — this is the server's *exact* anchored pattern (`create-climb-filters.ts` has the rationale). The leading `p` + trailing `r` anchor the token: `%p2r%` only matches placement id 2 — `p12r`/`p22r` don't contain the substring `p2r`. Keeping it byte-identical to the server is deliberate parity; a word-boundary tweak would diverge.
- **Tick-dependent filters offline** — they read **local** `boardsesh_ticks`, which include offline dual-writes, so they're *more* current than a network query. Offline browse only runs on a downloaded board, and the pull syncs `user_data` (ticks) **before** `board_data` — so a downloaded board implies ticks are synced.
- **Cross-user checkpoint on sign-in** — `board_climbs` is shared *reference* data (identical for every user), so its sync cursor is user-independent; resuming it for the next user is correct, not a leak (drafts are user data in separate tables). Preserving it avoids a 200k-row re-crawl; documented in `connection.ts` + `offline-sync-plan.md`.
- **`triggerSync` on rapid toggles** — `runSync` is single-flight (one in-flight + one queued follow-up) and reads the latest setting, so concurrent toggles collapse safely. Added a clarifying comment.
- **Checkpoint `NOT LIKE 'checkpoint:board_climbs:%'`** — `board_climbs` is *not* a prefix of `board_climb_stats` (they diverge at `s` vs `_`), so the first clause doesn't cover stats keys — both `NOT LIKE`s are required.

marcodejongh:
Addressed the minor notes in `8d8e292`:
1. **canReadLocal await** — added the explicit `await`.
2. **Retry re-entrancy** — added a ref guard so a rapid double-tap can't spawn overlapping retries (the drain was already single-flight; this just avoids the wasted re-entrant work).
3. **triggerSync .catch** — non-issue: `triggerSync` returns `void` and `runSync` swallows its own errors internally (single-flight cycle), so there's no floating rejection to catch.

Marked ready for review — CI green on the prior commit; re-running on this one.

marcodejongh:
Round 3 addressed in `687301493` (rebased onto your `offline-board-downloads` flag-gate commit):

**Functional**
1. **Double `getDatabaseHandle()`** — `resolveClimbSearch`/`Count` now fetch the handle once and pass it into `canReadLocal(db, input)`, so the routing decision and the local read use the same handle (no init-race window).
2. **Moonboard size-scoping duplicated ×3** — extracted `isSizeScopedBoard(boardType)` into `@boardsesh/board-config` (a backend dep too) as the single source; now referenced by the sync resolvers, the offline local search, and the download-availability check.

**Minor**
- **#6** `syncClimbStats` now reuses the same `boardClimbsLayoutSizeConditions` helper (`bc.`-qualified) as `syncClimbs` instead of inlining.
- **Dead-letter poll** relaxed 5s → 30s.

**Kept as-is (with reasoning)**
- **`handleRetrySync` not `useCallback`** — matches this file's convention (`handleSignOut`/`handleHelpTranslate`/`handleReplayWalkthrough` are all plain), and the `MoreForm` model is rebuilt each render regardless, so memoizing one handler is inconsistent for no gain here.
- **Dynamic import of the mmkv accessor in `isBoardDownloadedLocally`** — it's the fix for the Rolldown collection-scan (react-native Flow) that affects the *entire* search-hooks test graph, not one file, so a per-test `vi.mock` doesn't scale; the import is module-cached so the per-call cost is a single resolved-promise microtask, negligible next to the SQLite query.

marcodejongh:
Offline board downloads are now feature-flagged (commit 4788fcf5d):

- **Flag**: [`offline-board-downloads`](https://us.posthog.com/project/412845/feature_flags/745547) — created in PostHog, **active at 0% rollout** (same ramp workflow as `logbook-filters`). Raise the percentage to roll out; testers can force it on today via More → Feature Flags.
- **Flag off / absent**: the My Boards per-row download toggle and status caption don't render. That toggle is the only writer of `syncEnabledBoards`, and the sync scheduler, pull client, and local-first reads all no-op while it's empty — so users who never downloaded anything get exactly the pre-offline behavior (all reads on the network, no board pulls).
- **Kill-switch semantics**: flipping the flag off hides the UI but boards already enabled keep syncing and serving local reads. If a true kill is ever needed, that's a follow-up.
- **Not gated (intentionally)**: offline writes — the mutation queue, dual-write ticks/favorites, and the More-screen "Sync issues" retry. Logging climbs offline and syncing on reconnect stays on for everyone.
- Covered by a new `BoardManageRow` test asserting the flag-off row renders no toggle and no offline caption.

One heads-up for merge: the PR's Release Notes will ship to "What's New" while the flag is still at 0% — consider holding the offline line until the ramp, or wording it as rolling out.

PR #3469, opened 2026-07-05
feat(web): make home hero drive app installs; swap beta/boards rails

marcodejongh:
## Summary

Reworks the top of the web home page:

- **Hero CTA now drives app installs** instead of starting a sesh. The button is platform-aware:
  - iOS / desktop web → **Install from App Store**
  - Android web → **Get it on Google Play**
  - Legacy native app (retired Capacitor build — we ship a React Native app now) → **Update the app**, pointed at whichever store that straggler installed from.

  During the brief pre-detection window the CTA renders the App Store install so it's real and clickable on first paint, then corrects to Play Store / update once the platform check runs on mount. Clicks fire an `App Install Click` analytics event tagged `placement: 'hero'` + `mode`.
- **Swapped the discovery rails**: *Fresh beta* now leads, *Boards near you* follows.

The existing "Get the Boardsesh app" install card lower in the onboarding stack is unchanged (kept per request), so web visitors see the CTA in both spots.

Removed the now-dead active-session "Continue climbing" hero path (and its unused `usePersistentSession` / `setClimbSessionCookie` imports). The Start-a-sesh drawer is still reachable from the "Connect your board" onboarding card and the product tour.

### Notes
- `home.hero.ctaStart` / `ctaContinue` were deleted from all three locale catalogs (en-US, es, fr) and replaced with `ctaInstallIos` / `ctaInstallAndroid` / `ctaUpdate` — the orphaned-key check fails otherwise.
- The App Store screenshot flag (`boardsesh:e2e-suppress-install-card`) still hides the onboarding install *card*; the hero now shows the "Update the app" CTA in that mode. The RN app doesn't render this web page, so this only affects any residual web screenshot.

## Release Notes

- The home screen now points you straight to the app in the App Store or Google Play.

## Test plan

- `vp check` on all changed files — clean.
- `vp run typecheck:web` — passes.
- `vp run check:i18n` and `vp run check:i18n:orphans` — clean (no hardcoded strings, no orphaned keys).
- i18n catalog-completeness test — 28 passed (locale parity for the new keys).
- Rewrote `home-page-content.test.tsx` to cover the new hero: iOS install → App Store URL, Android install → Play URL, native → "Update the app", each firing the right `App Install Click` event and *not* opening the sesh drawer / navigating. `SKIP_TEST_INFRA=1 vp test run … home-page-content.test.tsx` → 8 passed.

https://claude.ai/code/session_01RknoTDCvyPTKi2r8y87wm8

### DIFF
--- a/packages/shared/i18n/locales/en-US/marketing.json
+++ b/packages/shared/i18n/locales/en-US/marketing.json
@@ -30,8 +30,9 @@
     "hero": {
       "title": "Get on the board!",
       "subtitle": "Track your sends across Kilter, Tension, and MoonBoard.",
-      "ctaStart": "Start climbing",
-      "ctaContinue": "Continue climbing"
+      "ctaInstallIos": "Install from App Store",
+      "ctaInstallAndroid": "Get it on Google Play",
+      "ctaUpdate": "Update the app"
     },
     "onboardingHeader": "Make it yours",
     "install": {
--- a/packages/shared/i18n/locales/es/marketing.json
+++ b/packages/shared/i18n/locales/es/marketing.json
@@ -30,8 +30,9 @@
     "hero": {
       "title": "¡A escalar!",
       "subtitle": "Registra tus encadenes en Kilter, Tension y MoonBoard.",
-      "ctaStart": "Empezar a escalar",
-      "ctaContinue": "Seguir escalando"
+      "ctaInstallIos": "Descárgala en el App Store",
+      "ctaInstallAndroid": "Disponible en Google Play",
+      "ctaUpdate": "Actualiza la app"
     },
     "onboardingHeader": "Hazlo tuyo",
     "install": {
--- a/packages/shared/i18n/locales/fr/marketing.json
+++ b/packages/shared/i18n/locales/fr/marketing.json
@@ -30,8 +30,9 @@
     "hero": {
       "title": "Au mur !",
       "subtitle": "Notez vos croix sur Kilter, Tension et MoonBoard.",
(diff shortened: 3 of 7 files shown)


marcodejongh:
Thanks — addressed all three in 463e5f4:

- **Android native straggler**: extracted the mapping into `app/lib/hero-install.ts` and added both a render test (Android UA + native → Play Store update) and pure unit tests for every platform.
- **UA restore no-op**: fixed — `userAgent` is on the prototype so the `getOwnPropertyDescriptor` restore was silently skipped. Now captures the string and always redefines it in `afterEach`.
- **Pre-hydration `'unknown'` state**: pinned with a direct `resolveHeroInstall('unknown', …)` unit test (the effect-flush in RTL made this awkward to isolate via render, so a pure test is cleaner).

marcodejongh:
Both remaining notes are resolved without changes:

1. **`other-web` untested** — it is covered: `hero-install.test.ts:14` asserts `resolveHeroInstall('other-web', 'ios')` → App Store install (the "iOS / desktop web" case).
2. **Android first-paint flash** — intentional and documented. On first paint everyone (incl. an Android native straggler) hits the `'unknown'` branch → App Store CTA, correcting on mount. Guarding the render behind a skeleton would degrade the *primary* hero CTA for every web visitor to spare a rare retired-Capacitor Android user a sub-second wrong-store flash before correction — a trade I'd rather not make. The store link is only wrong for that one legacy cohort, in a sub-100ms window.

CI is green, no conflicts, all legitimate feedback addressed — marking ready for review.

PR #3470, opened 2026-07-05
refactor(mobile): single offline-aware GraphQL interceptor with an operation registry

marcodejongh:
## Summary

Stacked on #3464. Replaces the three per-operation local-first dispatchers in `offline-search.ts` (`resolveClimbSearch` / `resolveClimbSearchCount` / `resolveClimb`) with one client-level seam: `offlineAwareRequest(document, variables)` in `offline-request.ts`, backed by a registry keyed on the GraphQL document string. Each entry describes how to serve its operation from local SQLite (`canServeLocal` / `resolveLocal` returning the raw response shape / `offlineFallback`); unregistered documents fall straight through to the network client. Making a new read offline-capable is now "register it once" — no new dispatcher, no per-callsite routing.

**Behavior-preserving**: identical routing (downloaded + supported reads local even when online, network otherwise, empty/null fallback offline), identical results, query keys, and hook return shapes. Not a GraphQL executor — a thin `Map` lookup, no new dependencies.

Details worth knowing when reviewing:
- Hooks now cache the raw GraphQL response and unwrap via `select`, matching the sibling network-only hooks. Safe: no `getQueryData`/`setQueryData` on the affected keys anywhere, and the mobile app has no query-cache persistence. The infinite hook's `getNextPageParam` reads the raw pre-select pages (React Query v5 semantics, verified against the installed 5.100.14).
- The search and count registrations share one `canServeSearchLocal` gate so the list and its "Show N" count can never disagree on source; the sync `isOfflineSearchSupported` check stays ahead of the async mmkv-lazy-importing `isBoardDownloadedLocally` probe.
- `index-keyboard.test.tsx` now mocks the interceptor module, so its partial `operations` mock doesn't become load-bearing for future registrations.
- New test drives the real document constants through `offlineAwareRequest` and covers every routing branch plus branches the old suite missed: db-null, the unsupported-filter short-circuit, unregistered-document passthrough (even offline), and wire-shape assertions (`{ input }` wrapping vs flat GET_CLIMB vars).

Known follow-ups (deliberately out of scope for a behavior-preserving refactor):
- Board sync / mutation drainer never invalidate `['infiniteSearchClimbs']`, so the infinite list doesn't self-heal after a board pull (pre-existing; issue to follow).
- Three call sites still hit the network directly with registered documents (`use-queue-regrade.ts`, `select-climbs-for-plan.ts`, `NowOnTheWallPanel.tsx`) — offline they fail even on a downloaded board (pre-existing). The registry now makes migrating them a one-line change each (issue to follow).

## Release Notes

## Test plan

- `vp run test:mobile` — 443 files / 3593 tests, all green (includes the rewritten `offline-request.test.ts` and the untouched `use-infinite-search-climbs.test.tsx`, which asserts the `{ input }` wire shape through the real interceptor).
- `vp run check:mobile-bundle` — Metro bundle OK (the mmkv lazy-import graph didn't regress).
- 8-angle adversarial code review of the diff (line scan, removed-behavior audit, cross-file trace, reuse/simplification/efficiency/altitude/conventions) — confirmed findings fixed in this diff; latent/pre-existing ones listed above.


marcodejongh:
Addressed the review in 7e5d86c34:

- **Var-less registered operation**: restructured the guard so the offline fallback now applies even when a registered document is called without variables (previously that combination fell through to a doomed network request, contradicting the module's documented invariant). Online behavior is unchanged — no variables still degrades to HTTP. Both paths now have tests.
- **Local read errors**: added a regression test asserting a rejecting `resolveLocal` propagates to the caller and does not silently retry over the network (matches the old dispatcher behavior).
- **Cache shape coupling**: agreed it's load-bearing; no action taken per the review. The constraint is documented in the PR description, and the absence of `getQueryData`/`setQueryData` on these keys plus no cache persistence keeps it safe today.

marcodejongh:
Re the latest review round:

1. **COUNT unsupported-filter test**: added — it pins the contract that the list and its count share one gate (`canServeSearchLocal`) and can never disagree on source.
2. **Typed registry (discriminated union / branded documents)**: deliberately not doing this now. The erased-generics registry is the documented trade-off for keeping `offlineAwareRequest` a drop-in for `client.request` (same looseness as `request<T>` itself); every current call site is covered by tests asserting the exact wire shape. If a misregistration ever bites, the follow-up is per-operation typed helpers or branded document constants — noted as a residual risk in the PR description.
3. **Offline passthrough at the three direct call sites**: tracked in #3472.

marcodejongh:
Re the last review round's per-op error-propagation tests (COUNT/CLIMB local reads throwing): declining — error propagation is a property of the single shared interceptor path (`await operation.resolveLocal(...)`), not of each registration, and it's already pinned once with the SEARCH_CLIMBS case asserting no network retry. Duplicating it per operation would test the same line three times.

CI is green across the board, no conflicts, all review feedback addressed or answered — marking ready for review.

marcodejongh:
Re the latest review round: added the SEARCH_CLIMBS offline + downloaded test — it's the flagship offline scenario (browse a downloaded board with no signal) so it deserves an explicit pin.

On further symmetric permutations (e.g. GET_CLIMB offline+downloaded): declining. Routing is a single shared code path — `canServeLocal` never consults connectivity, and connectivity only picks between fallback and network after the local gate fails — so op × downloaded × online × supported permutations beyond the ones now pinned test the same branches repeatedly. The suite covers every distinct branch of the interceptor at least once per behavior, plus the shared search/count gate and per-op wire shapes.

Also rebased on the base branch's review-round-3 commit; its single-db-handle change was already the interceptor's design, so the rebase only re-applied the file deletion.

PR #3473, opened 2026-07-05
feat(mobile): drive a Rogue Fitness gym timer from a paired board

marcodejongh:
## Summary

Adds the ability to drive a **Rogue Fitness gym timer** (Home Timer 2.0 / Echo Gym Timer 2.0) from the mobile app. A timer is treated as **gym equipment paired to a board**: you pair it once in the board's My Boards config, and from then on, whenever you're driving that wall, logging a climb kicks off the timer's stopwatch.

This is a proof-of-concept for a larger goal — driving the timer during workouts generated by the sessions feature (interval / EMOM / countdown timers keyed to workout steps). The controller is built as a general imperative API (`pressButton` / `startStopwatch`); "start a stopwatch on tick" is just the first thin caller.

Protocol reverse-engineered from the Rogue app; documented in `docs/ROGUE_TIMER_BLE_SPEC.md`. The timer is an HM-10 UART peripheral driven fire-and-forget with 4-byte key-code frames (`55 AA 01 <code>`) — no ACK, no CRC.

**Two rules shape the design:**
- **Paired to a board.** Which timer to talk to comes from the active board's new `timerName` field (mirrors the existing per-board `serialNumber`), not a global setting.
- **Only the wall driver owns the timer.** The timer is connected only while you hold the board LED connection (lightbulb on). In a gym where several people have the same board saved, this stops everyone's ticks from stomping on the same timer.

### What's in it
- `@boardsesh/ble-protocol/rogue-timer` — pure frame builder, command table, UUIDs, name detection (+ tests against the spec's exact base64 fixtures)
- `rogue-timer-ble.ts` — write-only BLE controller reusing the shared `bleManager`
- `RogueTimerProvider` — connection reconciliation gated on board-LED-connected + paired timer
- Per-board `timerName` field, server-synced end-to-end (db migration, GraphQL, resolver, Zod)
- Pairing sheet in the board form; tick glue in `QuickTickBar`

## Release Notes

- Pair your gym's workout timer to your board — while you're lit up on the wall, every send starts the clock.

## Test plan

Automated (all green):
- `vp run typecheck:mobile`, `vp check`
- `vp run test:mobile` — 426 files / 3378 tests
- New `rogue-timer.test.ts` asserts frame bytes against the spec's base64 fixtures
- i18n catalog parity (en-US / es / fr)
- Backend board tests (`boards-serial-privacy`, `board-connection`, `board-gym-edit-authorization`) — schema-sql mirror + resolver

On-device (requires hardware — **can't run in CI or the simulator**, needs a maintainer with a Rogue Home/Echo timer):
1. My Boards → edit a board → More options → **Workout timer** → Pair → pick the timer.
2. Connect the board LEDs (lightbulb on) and tick a climb → the timer should reset and start counting up.
3. With the lightbulb **off**, tick a climb → the timer must **not** react (wall-driver gate).
4. The exact `STOPWATCH → RESET → OK` key sequence and inter-key delay may need on-device tuning — flag if the stopwatch doesn't cleanly zero-and-start.


marcodejongh:
Thanks — addressed the review across two follow-up commits (`28166e5`, `dae9a5f`):

**Bugs**
1. **Stale `connected` after an unexpected drop** — fixed. `RogueTimerProvider` now subscribes to `controller.onDisconnect`; a drop clears status + the connect target and triggers one reconnect attempt while the board LED is still held. Covered by a new provider test.
2. **`isRogueTimerName` matched Echo cardio** — scan/connect now gate on `isDrivableTimer` (`detectRogueDeviceType(...) === 'timer'`), so Echo Rower/Bike/Skier are excluded. (Nameless devices already fail the name gate, so `TimerRow`'s `?? ''` label never renders empty.)
3. **Concurrent `startDeviceScan`** — `connectByName` now stops its scan the moment the timer resolves (before connecting), and I've documented the shared-`bleManager` single-scan contention on `scanForTimers`. The two paths don't overlap in practice (pairing is on the board-edit screen; reconnect is on the play screen), so it's left un-serialized with a comment.

**Missing tests** — added `rogue-timer-provider.test.tsx`: wall-driver gate (no connect while LED off, tick is a no-op), connect on LED+pairing, `startStopwatch` → STOPWATCH/RESET/OK, disconnect on LED drop, and the drop/reconnect path.

**Minor** — empty `timerName` now rejected server-side (`z.string().min(1)`); added a comment on `startStopwatch`'s top-level-mode precondition.

Also worth flagging for reviewers: BLE can't be exercised in CI or the simulator, so the on-device `STOPWATCH → RESET → OK` key sequence + inter-key timing still needs a hardware pass (called out in the PR description).

marcodejongh:
Fixed in `3324129`:

- **Lingering scan blocks bleManager after board disconnect** — `teardownConnection()` now calls `bleManager.stopDeviceScan()`, so a `connectByName` that's mid-scan when the board LED drops can't hold the singleton's only scan slot for up to 10s and block a board reconnect. Idempotent (no-op when nothing's scanning).
- **`rogueTimer` in useCallback deps** — QuickTickBar now depends on the stable `startStopwatch` callback only, so the save handler isn't recreated on timer status/deviceName changes.
- **Error-state recovery + missing test** — the reconcile effect now settles status back to `idle` whenever you're not driving the wall, including from a lingering `error`/`connecting` (badge no longer sticks on 'error' after a failed attempt). Added a test for error → LED-off → idle → reconnect.
- **Reconnect not rate-limited** — documented the no-backoff behavior in code and flagging it here: a timer oscillating at range-edge will reconnect-drop-reconnect continuously. Left as-is for the POC (each attempt is a bounded ~10s scan and the human-held board LED bounds the session); worth revisiting with backoff if it shows real churn on device.
- **Migration trailing newline** — added to `0142_windy_roulette.sql`.

marcodejongh:
Addressed the second review pass in `47534c0`:

1. **Unbounded reconnect loop** — now bounded. Consecutive drop→reconnect cycles are capped (`MAX_RECONNECT_ATTEMPTS = 4`); after that the provider gives up (status `error`) instead of looping. The counter resets once a connection stays up for 30s (so a single late-session drop still gets fresh retries) and on a board-LED toggle. New provider test covers the give-up path.
2. **`teardownConnection` stopped the shared scan unconditionally** — now gated on a `scanActive` flag, so it only stops a scan *this* controller started. A provider disconnect can no longer kill an open pairing-sheet scan on a separate controller instance.
4. **`isRogueTimerName` misleading** — renamed to `isRogueEchoDeviceName` (it matches the Rogue/Echo brand, not timers); callers still combine it with `detectRogueDeviceType === 'timer'` via `isDrivableTimer`.
6. **No `TimerPairingSheet` tests** — added: scan starts on mount / stops on unmount, list renders + strongest-RSSI-first, tap reports the timer name, empty state, cancel.

**3 & 5 — deliberate for the POC:** `pressButton`'s broad catch is intentional (the write-with-response fallback is the whole point of catching a no-response rejection). And `status: 'error'` is intentionally silent — a board paired to a timer that's simply switched off is a normal, common state; a toast on every failed background connect would be noise. Both are noted in code; happy to surface the error in a follow-up if you'd rather show it.

marcodejongh:
Thanks — all four are non-blocking; my take on each:

1. **Double `onDeviceDisconnected` subscription** — intentional split: the internal one clears the *controller's* cached refs (so a stale characteristic can't be written), the `onDisconnect` one notifies the *provider*. They're independent — neither handler depends on the other having run, so delivery order doesn't matter. The provider only calls `onDisconnect` inside the post-connect `.then`, i.e. when the device is connected, so the no-op-unsubscribe edge isn't hit in practice. Left as-is.
2. **Double `stopDeviceScan()` on the success path** — deliberate: the happy-path stop releases the slot before the connect window, and the `finally` is the safety net for the reject/throw paths. ble-plx no-ops the redundant call. Kept for the safety net.
3. **`timerName` in public board queries** — deliberate decision to mirror `serialNumber` exactly (same `enrichBoard` exposure). The value is a generic equipment name (e.g. "Rogue Home Timer"), lower-sensitivity than the board controller serial that's already public, and the runtime only ever reads the *owner's own* active board. Happy to gate it owner-only in a follow-up if you'd prefer — I kept it consistent with the existing field rather than diverging one of the two.
4. **Reconnect backoff** — now capped at 4 attempts (was unbounded); worst case is ~4 bounded scan windows before it gives up. A fixed inter-attempt backoff is a reasonable follow-up if range-edge flapping shows up on real hardware.

CI is green and the PR's ready for review.

Issue #3478, opened 2026-07-06
[mobile] expo-ui crash: ModalBottomSheetView.partialExpand "no handler registered" — top Sentry issue (108 users)


marcodejongh:
## Summary

Top production error by user count. The `@expo/ui` native bottom sheet rejects `partialExpand` because **the shipped store binary's native `@expo/ui` module doesn't register that handler**, while a production OTA pushed JS that calls it — a JS-ahead-of-native mismatch. Only a new binary (built with the current `@expo/ui ~56.0.18`) permanently fixes it; an OTA guard should stop the crashes in the meantime.

- **Sentry:** [BOARDSESH-6V](https://boardsesh.sentry.io/issues/BOARDSESH-6V)
- **Impact:** 108 users, 725 events. First seen 2026-06-04, still ongoing (last seen today).
- **Handled:** yes (unhandled promise rejection), but it means the sheet fails to expand — the user's action silently does nothing.

## Error

```
Error: Call to function 'ModalBottomSheetView.partialExpand' has been rejected.
→ Caused by: java.lang.IllegalStateException: No handler registered for AsyncFunction
  'partialExpand' on view 'ModalBottomSheetView'. Did you forget to bind it with `partialExpand...`
```

## Root cause

`partialExpand` is not called anywhere in our source — it's the imperative native API of the `@expo/ui` `BottomSheet`/`Host` (`@expo/ui ~56.0.18`). The native `ModalBottomSheetView` compiled into the installed store binary predates the `partialExpand` async function, but the JS delivered over the `production` OTA channel calls it. Because the `updates` fingerprint didn't change for the JS-only bump, the OTA landed on binaries whose native module can't service the call.

Sample event: `release com.boardsesh.app@2.1.0+2000436`, Android 16 (Samsung SM-A366E), `is_embedded_launch: true`, `hermes`, RN 0.85.3. Seen on **Android**; verify whether iOS binaries are also affected.

## Fix direction

1. **Native (this release):** rebuild the store binary against the current `@expo/ui` so `partialExpand` is registered. This is the permanent fix and the reason it belongs in the upcoming App Store / Play submission.
2. **OTA safety net:** guard the imperative expand call so it no-ops (or falls back to a full expand) when the native handler is absent, so users on the current binary stop hitting it before the new build ships.
[...]

PR #3481, opened 2026-07-06
fix(mobile): guard @expo/ui Android sheet partialExpand against native skew

marcodejongh:
## Summary

Fixes the top production error by user count (Sentry [BOARDSESH-6V](https://boardsesh.sentry.io/issues/BOARDSESH-6V): 108 users, 725 events, Android-only).

`@expo/ui`'s Android community bottom sheet fires `expand()` / `partialExpand()` **fire-and-forget** on the native `ModalBottomSheetView` inside `snapToIndex` (only when a multi-detent sheet is re-snapped while already open — e.g. picking a different board in `BoardDetailSheet`). A store binary whose native `@expo/ui` predates one of those AsyncFunctions rejects the call (`No handler registered for AsyncFunction 'partialExpand' on view 'ModalBottomSheetView'`), and — being unawaited — it surfaces as a crash-reported unhandled rejection. The re-snap itself is a harmless no-op on those binaries.

**Fix (OTA safety net):** guard both calls at source in `patches/@expo%2Fui@57.0.3.patch` — attach a `.catch` so a missing native handler no-ops instead of throwing. This ships over OTA (Metro bundles `@expo/ui`'s `src`), so binaries that pull the update stop crashing immediately. This is strictly better than the existing Sentry filter because it stops the rejection at the source — it never reaches Sentry **or** PostHog **or** the dev overlay.

Chose a **no-op over falling back to `expand()`**: `expand()` can also be missing on the oldest binaries (the Sentry regex already matched both), so a fallback could itself reject, and a no-op never *wrongly* expands a collapse.

**Permanent native fix:** the repo is already pinned to `@expo/ui@57.0.3` (SDK 57), which registers the methods natively — so the next store build (2.2.0) retires the crash for good. No version bump or native change needed here.

Also:
- `packages/mobile/src/lib/sentry.ts` — kept the `beforeSend` filter as belt-and-braces (it still drops the report from pre-patch OTA bundles and the tail of old binaries) and updated the comment to reflect that the call is now guarded at source.
- Docs — fixed a stale `56.0.18` patch reference in `mobile-sheets-vs-routes.md`, and recorded the OTA-ahead-of-native invariant (issue point 3 / #3275) in `mobile-ota-updates.md`.

Addresses #3478 (fully retired once the SDK 57 store build ships).

## Release Notes

## Test plan

- `vp run test:mobile` — 429 files / 3396 tests pass (incl. existing `sentry.ts` filter tests).
- `vp run check:mobile-bundle` — Android bundle builds OK, confirming the regenerated patch applies cleanly against `57.0.3`.
- The unhandled rejection can't be reproduced on a current SDK 57 simulator (the method exists there); it only fires on the old 2.1.0 native layer. The guard is verified by patch-applies + bundle-builds, and by reasoning that a `.catch` on the fire-and-forget promise can't surface as an unhandled rejection.

https://claude.ai/code/session_01DVy5cPM2dynPMPz6KiNKFs

### DIFF
--- a/docs/mobile-ota-updates.md
+++ b/docs/mobile-ota-updates.md
@@ -49,6 +49,23 @@ install, since the device couldn't verify the manifest came from us.
    version, decoupled from OTA compatibility. Resolve the current value with
    `bunx expo-updates runtimeversion:resolve --platform ios|android` (from `packages/mobile/`).
 
+### Invariant: OTA JS must not call native methods newer than the min shipped binary
+
+The fingerprint gate only protects you when the JS change is matched by a native change. It does
+**not** catch a JS-only change that starts calling a native method the shipped binary doesn't have —
+e.g. bumping a native module's JS (or the module itself, if its native side ships separately) so the
+JS invokes a newer imperative native method. The fingerprint is unchanged (JS-only), so the OTA lands
+on old binaries whose native layer predates the method, and the call fails at runtime.
+
+This is exactly how #3478 happened: `@expo/ui`'s Android bottom sheet calls `partialExpand()` /
+`expand()` on the native `ModalBottomSheetView`. A production OTA pushed JS that calls them onto a
+store binary built against an older `@expo/ui`, and the unregistered AsyncFunction rejected — a
+crash-reported unhandled rejection.
+
+Rule: any imperative native call that could be OTA-ahead of the native binary must be guarded (a
+`.catch` / capability probe / `requireOptionalNativeModule` null-check) so it degrades to a no-op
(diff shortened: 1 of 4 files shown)


marcodejongh:
Addressed review point 1: `swallowMissingNativeHandler` now logs via `__DEV__ && console.warn(error)` so the guard is visible during development while staying a silent no-op in production (a dev build runs the current native layer, so a warning there means something genuinely unexpected rejected). Points 2 (no test for the patch path) and 3 (bun's full-length blob hashes) are acknowledged non-actionable — a node_modules patch has no unit-test harness, and the hash format is bun's own output that `bun install` + the bundle check confirm applies cleanly. Re-ran `check:mobile-bundle`: still OK.

Issue #3479, opened 2026-07-06
[mobile] iOS WatchdogTermination (OOM) — app killed for RAM overuse


marcodejongh:
## Summary

iOS OS watchdog is terminating the app for RAM overuse (fatal, unhandled).

- **Sentry:** [BOARDSESH-8B](https://boardsesh.sentry.io/issues/BOARDSESH-8B)
- **Impact:** 3 users, 4 events. First seen 2026-06-29. `mechanism: watchdog_termination`, `level: fatal`, `platform: cocoa`, `in_foreground: true`.
- Sample: iPhone17,1, iOS 26.5, `release com.boardsesh.app@2.1.0+3`.

## Notes

Low volume, but a hard crash. Consistent with the known board-art memory profile (native heap dominated by decoded board images in the feed). Fix is most likely JS-side (tighten `expo-image` cache caps / recycle offscreen board art) rather than native, so this is **P2, not release-blocking** — but worth confirming it isn't climbing as usage grows.

## Release Notes

none

PR #3482, opened 2026-07-06
fix(mobile): cut board-art image memory to stop iOS OOM watchdog kill (#3479)

marcodejongh:
## Summary

The iOS OS watchdog is killing the app for RAM overuse (Sentry [BOARDSESH-8B](https://boardsesh.sentry.io/issues/BOARDSESH-8B), `watchdog_termination`, fatal, in foreground). Verified in Sentry: **3 of 4 events are on the 4 GB iPhone 12 mini**, all while actively browsing board art. Every board-art `<Image>` holds a full-resolution decoded buffer in expo-image's memory cache with default limits and no eviction, nothing responds to OS memory pressure, and the play-drawer carousel piles a native-res (~7 MB RGBA) holds overlay into the cache per swiped climb.

The fix for exactly this already existed as **#3150** (`perf/board-art-image-memory`), which passed CI and was OTA-compatible but was closed with only fixable review nits. This revives its three mitigations on top of `main`, with those nits fixed:

- **Flush expo-image's memory cache on the OS `memoryWarning` event and on background** (new `useImageCacheMemoryManagement`, mounted at the app root). The `memoryWarning` path is the direct foreground lever against the watchdog kill.
- **Blank board-art bitmaps while backgrounded** via a ref-counted `useIsAppBackgrounded` store gating `LayeredClimbImage`, so the still-mounted views release their GPU textures + heap bitmaps. Re-decodes from disk on foreground (no network).
- **Render the carousel overlay at display size** (`renderWidth`) while keeping the shared board photo crisp (`backgroundVariant="full"`), and `recyclingKey` the carousel boards so a swipe releases the previous climb's overlay.

Nits fixed vs #3150: the AppState listener is installed lazily + ref-counted (tied to React lifecycle — no Fast-Refresh leak, no prod try/catch); the app-visibility tests are order-independent; `LayeredClimbImage` resets its overlay anchor on background; plus new tests for `recyclingKey` forwarding and for not re-flushing on re-foreground.

**JS-only — ships over-the-air to the existing 2.1.0 binary.** No native change.

Addresses #3479.

## Release Notes

- Fixed a crash where the app could be killed for using too much memory while browsing boards — most noticeable on older iPhones.

## Test plan

- `vp run test:mobile` — 432 files / 3407 tests pass, including 17 across the 4 new/expanded test files (`app-visibility`, `use-image-cache-memory-management`, `layered-climb-image`, `layered-climb-image-backgrounded`).
- `vp check` — lint + format clean on all changed files.
- `vp run check:mobile-bundle` — Android Metro bundle OK.
- Remaining: on-device iOS simulator check (open play drawer, swipe ~20 climbs, background/foreground, trigger Simulator → Simulate Memory Warning; confirm board art re-appears without flash and the app stays alive).

### DIFF
--- a/docs/react-native-performance.md
+++ b/docs/react-native-performance.md
@@ -263,6 +263,43 @@ surface the previous size's bundled background paths.
 
 ---
 
+## 7. Board-art image memory: render at display size, recycle, release on background
+
+**Rule:** Board-art surfaces are the app's largest memory consumer (decoded bitmaps live in native
+heap + as GPU textures, not the JS heap) — on iOS a 4 GB device can hit the OS watchdog OOM kill in
+the foreground while browsing (#3479). Three contracts keep them in check:
+
+- **Size the overlay to the display, keep the photo full-res.** The play-drawer board passes
+  `renderWidth` (display px × `PixelRatio.get()`) so the per-climb holds overlay rasterizes at the
+  shown size instead of the board's native ~1080px, plus `backgroundVariant="full"` so the _shared_
+  board photo (one decode per board-config, reused across every climb) stays crisp. `renderWidth`
+  alone would downgrade the photo to the 416px `thumb` — only the overlay should shrink.
+- **`recyclingKey` on always-mounted board surfaces that swap climbs.** The carousel boards key on the
+  climb's `frames` so swapping releases the previous overlay instead of holding both.
+- **Release board-art on background + memory pressure.** `LayeredClimbImage` blanks its `<Image>`
+  layers when `useIsAppBackgrounded()` is true, and `useImageCacheMemoryManagement` (mounted at the
+  app root) sweeps expo-image's memory cache on background and on the OS `memoryWarning` event. The
+  `memoryWarning` path is the direct foreground lever against the iOS OOM kill. Re-decodes from disk
+  on foreground (no network).
+
+**Why:** On-device profiling (Pixel 8 Pro, Android 16) showed ~248 MB native heap idle on the feed
+and the play-drawer carousel piling a native-res (~7 MB RGBA) overlay per swiped climb into the
+cache. Display-sizing + recycling cut carousel browse-growth ~19%; background-blanking cut the
+worst case (board left open on app-switch) ~96 MB. **Ceiling:** expo-image on Android is Glide,
+whose cache/pool are sized to device RAM and resist JS `clearMemoryCache()` (the native allocator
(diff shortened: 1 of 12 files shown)


marcodejongh:
Thanks — addressed two of the three in 028d32a:

- **Test isolation** → added `afterEach(cleanup)` so a test that throws before its own `unmount()` can't bleed the ref-counted AppState listener into the next test.
- **peekFrames comment** → added a note mirroring the current-board slot.

On the **`backgroundVariant` override test**: I skipped it deliberately. The decision is a one-line ternary (`backgroundVariant ?? (renderWidth != null ? 'thumb' : 'full')`) inside the hook, and its two constituent variant paths ('thumb' vs 'full') are already covered by the `buildBoardKey` variant tests. Covering the ternary itself would mean either flipping the whole `use-native-climb-render.test.ts` file to the jsdom environment to `renderHook`, or extracting the one-liner into an exported `_forTests` helper — both feel disproportionate for the risk. Happy to add it if you'd rather have the coverage.

marcodejongh:
Addressed in c5ef3fb / prior:

- **Repeated background cycles test** → added a case asserting a second background (after a foreground) flushes `clearMemoryCache` again (now 2 calls), alongside the existing single-cycle and no-re-flush-on-foreground cases.
- **Swipe-commit `recyclingKey` brief-blank risk** → this is the exact combination #3150 shipped and validated on-device: the current board carries both `recyclingKey={currentFrames}` **and** `suppressOverlayTransition={isCommitting}`, and the peek overlay is already painted (memory-cache hit) during the drag, so the commit swap lands on the next frame rather than fading/blanking. That pairing is what removed the Android end-of-swipe flash in the first place. Flagging it in the on-device pass regardless — will confirm no flash on an iPhone before this leaves draft.

marcodejongh:
**On-device verification (post-merge / via this PR's OTA channel).** A simulator can't reproduce this OOM (no jetsam/RAM enforcement), so the meaningful check is a physical low-RAM iPhone:

1. On a TestFlight/store build with the `tester` role: **More → Development → OTA Channel Switcher → `pr-3482`** to pull this branch's JS (native binary unchanged — this ships OTA).
2. Open the play drawer and swipe through ~20 climbs; confirm no end-of-swipe flash and the board photo stays crisp (overlay is display-sized, photo stays `full`).
3. Background → foreground the app; confirm board art re-appears without a visible flash.
4. Ideal repro device is the iPhone 12 mini (4 GB) from the Sentry events — browse the feed + play drawer for a while and confirm the watchdog no longer kills the app.

PR #3494, opened 2026-07-06
fix(mobile): light the iPad wall kiosk without navigation + screenshot pipeline hardening

marcodejongh:
## Summary

The App Store screenshot workflow failed 62% of its last 30 runs, and every iPad Pro 11-inch shot shipped with a dark board ("WALL IS DARK") plus a `02-climbs.png` that was a byte-identical copy of `01-home.png`. This PR fixes the iPad failure at its root and lands the remaining reliability work from the #3492 review.

### iPad 11" dark wall — root cause and fix

Artifacts from run 28781735332 show the 11"'s first sidebar tap (Climbs) silently never navigated on every locale, while the 13" was fine. The wall kiosk seed was only published when the Climbs screen mounted, so the hero shot rendered unlit holds.

Three layers of fix, so no single tap can sink a shard again:

1. **The wall no longer depends on navigation.** `ScreenshotBoardAutoActivator` (root-mounted, screenshot-mode only) now runs the same default-filter climb search the Climbs list uses and publishes the wall seed as soon as the active board resolves. Shared mapping lives in `buildScreenshotWallSeed` (`screenshot-wall-seed.ts`); the Climbs screen re-publishes the same seed (idempotent).
2. **Sidebar taps are now verifiable.** The #3484 note said Maestro couldn't see RN views — re-tested with `maestro hierarchy`: labeled pressables ARE exposed, including their `selected` accessibility state. `IpadSidebar` items now carry a locale-independent `testID` (`ipad-sidebar-<segment>`), and `app-store-ipad.yaml` taps by id, waits for `selected: true` as proof the navigation took effect, and re-taps via `retry` if it didn't. All per-device coordinate-tap math (`ipadSidebarTapPoint`, anchor tables) is deleted.
3. **Bad captures fail loudly.** `collectScreenshots` now fails the run when two captures are byte-identical — the exact 11" failure shape — instead of shipping a duplicate to App Store Connect.
4. **The master-detail pane is lit in every iPad shot.** On the plain `/climbs` list the detail pane used to sit as a dead black column (nothing selected). In screenshot mode the Climbs screen now activates the first climb when the pane layout is up (`usesDetailPane`, newly exposed from the drawer host) — the pane shows a lit board beside the list and stays lit for the Home/Discover shots. iPhone is untouched (there the same activation would cover the list with the play drawer).

### Reach-home / flake hardening (residual #3492 review findings)

- `waitForHomeReady(baseline, 240)` in `mobile-ios-shots.ts` / `mobile-android-shots.ts` passed the timeout into the new `readyBaseline` slot (unsatisfiable) — now `(baseline, 0, 240)`.
- The readiness server's bind is verified via `/probe` after spawn (its stdio is `ignore`, so a port conflict used to be silent); inline server code de-minified per repo naming rules.
- The app retries the reach-home readiness ping (3×, 2s apart) instead of one fire-and-forget fetch.
- Screenshot auto sign-in retries transient failures (network / 5xx, 3× with backoff) — previously one `error=network` blip cost a full 120s relaunch cycle (seen in run 28579610631).
- `screenshotReadinessCount` tests use a private temp log instead of writing the real one.
- The Maestro download curl in CI retries (`--retry 5 --retry-all-errors`).
- The iPhone flow's prime-dialog steps are `optional:` — a reused local simulator remembers the scheme confirm and the "Open" dialog never pops, which used to fail the whole flow (CI runners are always fresh and still prime).

## Release Notes

## Test plan

- Local `vp run mobile:screenshots` on iPad Pro 11-inch (M5), en-US, prod backend — three runs across the iterations: `00-wall` renders the lit hero (ON THE WALL NOW · V3 Moon Landing), `02-climbs` shows the climb list with the first climb selected and its lit board filling the detail pane, `01-home` shows the same lit pane, all six frames distinct, every id-tap `selected`-state assertion passed.
- Local `vp run mobile:screenshots` on iPhone 16 Pro Max, en-US: 10 distinct frames, duplicate guard green (and the remembered-dialog case reproduced + fixed).
- `vp run typecheck:mobile`, `vp run test:mobile` (3423 tests), scripts tests (41, incl. new duplicate-guard + flow-shape tests), `vp check` clean on touched files.
- Full 9-shard `workflow_dispatch` on this branch (no upload): link in comments once green.

https://claude.ai/code/session_012KV6NHZAQUPLz5EigLLd7p

### DIFF
--- a/.github/workflows/mobile-screenshots-ios.yml
+++ b/.github/workflows/mobile-screenshots-ios.yml
@@ -252,7 +252,7 @@ jobs:
         run: |
           set -euo pipefail
           url="https://github.com/mobile-dev-inc/maestro/releases/download/cli-${MAESTRO_VERSION}/maestro.zip"
-          curl -fsSL "$url" -o maestro.zip
+          curl -fsSL --retry 5 --retry-all-errors "$url" -o maestro.zip
           echo "${MAESTRO_SHA256}  maestro.zip" | shasum -a 256 -c -
           unzip -q maestro.zip -d "$HOME/.maestro-dist"
           echo "$HOME/.maestro-dist/maestro/bin" >> "$GITHUB_PATH"
--- a/packages/mobile/.maestro/app-store-ipad.yaml
+++ b/packages/mobile/.maestro/app-store-ipad.yaml
@@ -6,67 +6,118 @@ name: app-store screenshots (iPad)
 # Why this differs from the iPhone flow (app-store.yaml): the iPad CANNOT be driven
 # by deep links. On the iPad simulator every `openurl com.boardsesh.app://…` pops the
 # "Open in 'Boardsesh'?" scheme-confirm dialog and the URL is never delivered to the
-# app's router — navigation just never happens (verified: an all-deep-link run on
-# iPad produced 11 byte-identical "stuck on home + dialog" frames, while the same
-# flow on iPhone navigated 10 distinct screens). iPhone remembers the confirm after
+# app's router — navigation just never happens. iPhone remembers the confirm after
 # the first prime; iPad does not.
 #
-# So the iPad navigates via the persistent left SIDEBAR using coordinate taps — the
-# same gesture-tap mechanism the rest of this RN/Fabric build needs (the accessibility
-# tree doesn't expose plain RN views, only native inputs + system dialogs). No
-# `openurl`, so no dialog. The orchestrator selects this flow for iPad devices (see
-# iosSourceFlowFile in scripts/mobile-screenshots.ts).
+# So the iPad navigates via the persistent left SIDEBAR. Each item is a real
+# accessibility element (IpadSidebar sets accessibilityLabel, a locale-independent
+# `testID` of ipad-sidebar-<segment>, and accessibilityState.selected on the active
+# item), so every step taps by id and then waits for `selected: true` — PROOF the
+# navigation took effect, wrapped in a retry that re-taps if it didn't. The previous
+# coordinate-tap flow had no such proof: on the 11" iPad the first tap after the
+# rotation relayout was silently swallowed, and 02-climbs shipped byte-identical to
+# 01-home with a dark wall hero (the orchestrator now also fails the run on
+# byte-identical captures).
 #
-# The ${TAP_*} points are substituted per device by the orchestrator
-# (renderMaestroFlowForIosDevice): the sidebar is laid out in fixed logical points,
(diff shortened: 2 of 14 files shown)


marcodejongh:
Full 9-shard validation dispatch (no upload): https://github.com/boardsesh/boardsesh/actions/runs/28825764537

marcodejongh:
Review items addressed in d9a9b24d2:

1. **pingUntilDelivered response status** — fixed; the ping now checks `response.ok` and retries on a non-2xx instead of treating any HTTP response as delivered.
2. **Probe loop unit test** — intentionally left untested: the probe verifies a real spawned process bound the fixed port 19870 via curl, so a unit test would either bind that port (conflicting with any concurrent capture run on the same machine) or mock away exactly the thing being verified. It's a diagnostic-only path that degrades to a loud warning.
3. **screenshotReadinessCount signature** — `READINESS_LOG_PATH` remains exported and the parameter is documented as tests-only at the definition.

Also fixed the CI lint failure: the one-line `waitForHomeReady` edits pulled `mobile-ios-shots.ts`/`mobile-android-shots.ts` into the changed-files lint scope, exposing their pre-existing `ScreenshotOptions` literals missing `orientation` (scripts/*.ts are in no tsconfig, so `vp check` is their only type gate).

marcodejongh:
Second round addressed:

1. **Disabled-state input** — hoisted to a module const with a comment; with `enabled: false` React Query only hashes it into the key, the fetch never runs (verified: `useSearchClimbs` does no input validation outside the queryFn).
2. **Blocking probe** — now documented as intentionally synchronous (the orchestrator is spawnSync-based end to end, and nothing useful can happen before the second reach-home signal is known live).
3. **Android normalization ordering** — no Android normalization pass exists today (the iOS one exists solely for the landscape-iPad sips rotation), so there's nothing to reorder; the guard sits directly before collection on both platforms.
4. **Docs** — checked `docs/android-emulator-screenshots.md` and `docs/ios-simulator-screenshots.md`: neither mentions the removed tap-point API, reach-home signatures, or capture collection, so there is nothing stale to update.

marcodejongh:
Third-round nits: the pointless final 2s delay in `pingUntilDelivered` is fixed. The readiness bind-failure staying **WARN** (not ERROR) is deliberate — the server is a redundancy layer, and degrading to the Metro marker must not fail a run that can still succeed; the CI diag from #3492 (`readinessServerReachable` on reach-home failure) already surfaces it loudly when it matters. The disabled-input placeholder note is acknowledged — the guard comment on the const spells out the enabled:false contract.

PR #3495, opened 2026-07-07
fix(sync): scheduler stop kills queued follow-up; NULL-scope tombstone index

marcodejongh:
## Summary

Follow-ups from the final automated review rounds on #2785, which merged while these were being written (the last round's fixes never landed):

- **Scheduler cleanup kills the queued follow-up**: a sync trigger that arrived while a cycle was in flight used to fire that cycle's finally-block re-run *after* `startSyncScheduler`'s cleanup ran (sign-out / feature-flag flip) — a sync with nothing mounted to own it. The cleanup now resets `pendingTrigger`; regression test included.
- **Dev warning for unmapped mutation tables**: `invalidateForTable` silently no-ops for a `table_name` missing from its key map, so a future mutation type would drain successfully but never refresh its UI. Now warns in `__DEV__`.
- **Migration 0145** (drizzle-kit generated): partial index `sync_deletions (deleted_at, id) WHERE user_id IS NULL`, giving the reference-data branch of `syncDeletions`' `user_id = $x OR user_id IS NULL` filter its own range scan instead of riding the composite index via BitmapOr. 0144 is untouched — it already shipped with #2785.

Two other claims from that round were verified false and not acted on: the foreground debounce does reschedule on every `active` event, and the NetInfo handler is synchronous so `wasConnected` can't lag a second event.

## Release Notes

## Validation

- `vp run typecheck:db` / `vp run typecheck:mobile`
- `vp test run --project mobile` (sync + mutation-queue suites: 122 tests, incl. the new cleanup regression test)
- `bunx drizzle-kit check` (journal/snapshot consistent; 0145 generated, not hand-written)
- `vp check` at baseline

https://claude.ai/code/session_01LxZSg1KwkFVCD46WWCXmWg


marcodejongh:
Ready for review: all required checks green ('ci-status' passes; automated review says ready to merge). The failing 'publish' job is the per-PR OTA preview channel — the self-hosted expo-open-ota server is returning 'upstream error' when marking iOS updates finished (two identical failures ~5 min into the finish call, files upload fine). Retried once more; if it keeps failing the server needs a look, but it doesn't block this PR — it only means no `pr-3495` tester channel.

PR #3497, opened 2026-07-07
refactor(mobile): extract offline sync engine into @boardsesh/offline-sync

marcodejongh:
## Summary

Extracts the offline sync engine out of `packages/mobile` into a dedicated pure-TS package, **`@boardsesh/offline-sync`** (`packages/shared/offline-sync`), for maintainability. Pure relocation + mechanical seam changes — no engine logic rewrites.

**What moved:** the mutation outbox (`queue`/`drainer`/`handlers`/`error-classification`), the pull client, sync scheduler, checkpoints, `table-config`, the on-device SQLite DDL + migration runner, and the offline board-scope key helpers — plus their test suites (which already ran on `node:sqlite`; the adapter is now exported at `@boardsesh/offline-sync/testing`).

**Platform I/O is injected instead of imported.** The package has zero runtime dependencies:

- structural `SqlExecutor` / `OfflineDatabase` replace `expo-sqlite` types — the real `SQLiteDatabase` is structurally assignable, so mobile passes its handle with no wrapper
- `QueryInvalidator` (one-method slice) replaces the `@tanstack/react-query` dependency; `DrainOptions.isOnline` replaces `onlineManager`
- `SchedulerTriggers` replaces `AppState`/`NetInfo`; `onSchemaDrift` / `onCycleError` replace the Sentry and `__DEV__` console couplings

The mobile binding lives in `packages/mobile/src/offline/offline-sync-adapter.ts`. Rule (documented in the adapter + package docblock): mobile code never imports `drainMutationQueue`/`startSyncScheduler`/`triggerSync`/`pullSync` from the package directly — only via the adapter, so the real connectivity probe is always attached.

**Stays in mobile:** expo DB lifecycle (`connection.ts`, seed ATTACH), local read queries, the sync-status store, hooks, and the bridge component.

**Docs:** `sync-table-manifest.md` now points its DDL/table-config contract at the package paths; `offline-sync-plan.md` gains a "Why an outbox, not dirty flags on the data tables" section recording the design rationale, and CLAUDE.md lists the new package.

## Release Notes

## Test plan

- `vp test run --project offline-sync` — 134 tests, all green (engine suites moved with the code; scheduler suite now exercises the injected triggers incl. debounce + reconnect-edge)
- `vp run test:mobile` — 3562 tests across 448 files, all green
- `vp run typecheck:offline-sync`, `vp run typecheck:mobile` — clean (proves `SQLiteDatabase` → `OfflineDatabase` and `QueryClient` → `QueryInvalidator` structural assignability, no casts at call sites)
- `vp run check:mobile-bundle` — Metro resolves the new workspace package
- `vp check` on all changed files — 0 errors

https://claude.ai/code/session_012SPRg4Rd4FBUnEJa3S7cGV


marcodejongh:
Re: Claude Review — point by point:

**2 (isOnline default) — fixed in dc734620b.** Agreed there's a footgun, but fail-closed (`() => false`) is the worse default: a consumer that forgot the probe would silently *never* sync offline writes, which is quiet data loss rather than burnt retries. There is no safe default in either direction, so `isOnline` is now a **required** option — the type system enforces what the adapter docblock previously only asked for. The mobile adapter binds `onlineManager` exactly as before.

**1 (module-level singletons) — deliberate, documented, unchanged.** The drain flag / sign-out guard / wipe epoch coordinate the drainer, the pull client, and sign-out *across one app runtime*, which is exactly one JS context today (and would remain one per context for a web consumer — a second consumer in the *same* context is the only hazard). This is called out in a comment at the declaration site. Converting to per-caller instances is a real behaviour-bearing refactor of freshly-battle-tested code (wipe-epoch semantics), out of scope for a pure-relocation PR; happy to revisit if a second in-context consumer ever appears.

**3 (NetInfo subscription leak) — not a leak.** `startSyncScheduler` returns a cleanup that unsubscribes both triggers (`subscription.remove()` / NetInfo's unsubscribe), and `OfflineSyncBridge` invokes it in its effect teardown — identical lifecycle to the pre-move scheduler. Constructing the trigger object per call doesn't subscribe anything; subscriptions happen inside `startSyncScheduler` and are torn down by its returned stop.

**4 (missing offline early-exit test) — exists.** `drainer.test.ts` first case: `drainMutationQueue(..., { isOnline: () => false })` asserting `peekPending` is never called.

**5 (fake timers not restored) — they are.** The debounce test wraps the body in `try { … } finally { vi.useRealTimers(); }`, which restores real timers even on assertion failure.

marcodejongh:
Latest review round addressed:

1. **Test-only helpers** — `__resetDrainerStateForTests` / `__resetSyncSchedulerStateForTests` moved off the main entry onto `@boardsesh/offline-sync/testing`; the two mobile suites that used them import from there now.
2. **Empty `dependencies`** — dropped.
3. **tsconfig trailing newline** — added.
4. **offline-board-key tests** — confirmed migrated, split by responsibility: the pure encode/parse cases live at `packages/shared/offline-sync/src/__tests__/offline-board-key.test.ts`; the MMKV-backed toggle-store cases stayed in mobile as `packages/mobile/src/settings/__tests__/offline-board-toggle.test.ts`. No cases were dropped.

marcodejongh:
Latest review round (3cc2d158b):

1. **Adapter-only rule now enforced, not comment-only** — added the `no-restricted-imports` entry to `.oxlintrc.json` (editor-level) and, since `vp check`'s reduced ruleset drops that rule, a grep guard (`scripts/mobile-offline-sync-imports-check.sh`, task `check:mobile-offline-sync-imports`) wired into CI next to the `@expo/ui` platform-imports backstop — same pattern, verified to exit 1 on a direct import and 0 clean.
2. **`./testing` in production bundles** — leaving as-is: conditional export gating (`development`/`test` conditions) is unreliable under Metro, the package is monorepo-internal, and nothing imports the subpath outside test files. The escape hatches are at least off the main entry now.
3. **Singleton constraint** — documented in the package docblock: module-level guards assume exactly one app runtime per JS context; a future web consumer must use the package client-side only (no SSR import).

marcodejongh:
Re: the two remaining items — both already covered:

1. The `isOnline: false` early-exit test exists: `drainer.test.ts` → *"skips the drain entirely while offline"* passes `{ isOnline: () => false }` and asserts `peekPending` / `processMutation` / `recordFailure` / `markDeadLetter` are all never called. (Every other case uses `{ ...ONLINE }` precisely because that branch has its own dedicated test.)
2. `./testing` gating — declined earlier with rationale (Metro doesn't reliably honor conditional exports; the package is monorepo-internal and nothing outside test files imports the subpath). See the previous reply.

PR #3498, opened 2026-07-07
feat: tall/wide climb filters offline + on every Aurora board

marcodejongh:
## What & why

Testing offline mode, a downloaded Kilter Homewall board showed **no climbs** in airplane mode — because the **Tall Climbs** filter was on, and tall/wide weren't expressible against the local SQLite tables, so the offline search short-circuited to an empty result (rendered as a silent "no climbs").

Root fix: derive tall/wide from the **already-synced, GIN-indexed `compatible_size_ids`** instead of an edge-threshold subquery (tall) and an `EXISTS(board_climb_holds …)` (wide). This makes both filters:

- **Work offline** — no new synced column, no `board_climb_holds` sync.
- **Generalize to every board with a size grid** — Kilter Original, Tension Board 2, Decoy, Grasshopper — not just the Kilter Homewall. Single-size boards (MoonBoard, Touchstone) opt out automatically.
- **Run online and offline off one shared definition** (`getTallWideScope`), so results match.

### Definition (size-relative, product-scoped)

For the active `(board, layout, size)`: **wide** = the climb's `compatible_size_ids` overlaps no *narrower* size in the layout's product family; **tall** = overlaps no *shorter* size. The Tall/Wide chip only shows when a narrower/shorter size exists, so the smallest size in each axis offers nothing. Reference sets are scoped to the active size's product (a board's `compatible_size_ids` can list cross-product sizes).

## Performance

Wide previously did a per-climb nested-loop semi-join into the large `board_climb_holds` table. The new predicate uses the `compatible_size_ids` GIN index for the base scope and applies tall/wide as a cheap array filter — **no `board_climb_holds` access**. `EXPLAIN ANALYZE` on prod (Kilter Homewall, size 25, wide):

| | Execution | Buffers | board_climb_holds |
|---|---|---|---|
| Old (`EXISTS board_climb_holds`) | ~568 ms | ~206k hits | nested-loop per climb |
| New (`NOT compatible_size_ids &&`) | ~173 ms | ~19k hits | none |

## Correctness / parity

Tall is a strict **superset** of the old edge filter: the +613 extra climbs (Homewall) have their lowest hold exactly on the short board's floor (`edge_bottom = 24`). Those holds aren't rendered/reachable on the shorter board (strict bounds), so they genuinely need the taller size — consistent with how `compatible_size_ids` and the renderer treat boundary holds. No climb is dropped. Validated on prod: Kilter Homewall 11,869 wide / 12,637 tall, Tension Board 2 14,252 / 24,319, Grasshopper 7,562 / 17,445 — all sane populations.

## Also

Offline empty-state: when offline **and** a still-unsupported filter is active (drafts, beta videos, zones, hold-state), the climbs list now says the filter needs a connection with a one-tap **Clear**, instead of a generic "no climbs". (en-US/es/fr.)

## Test plan

- `getTallWideScope` unit tests (board-constants): grid cells, Tension generalization, single-size/MoonBoard, mismatched-product fail-closed.
- Online `create-climb-filters` tests rewritten for the `compatible_size_ids` SQL + generalization.
- Offline `search-climbs-local` tests: tall/wide over `compatible_size_ids`, `isOfflineSearchSupported` now true for tall/wide.
- Full suites green: mobile (3699), web (changed components), db unit, i18n catalog parity, `vp check`, typecheck (mobile/web/db).

## Release Notes

Filter for **tall and wide climbs on more boards** — Kilter Original, Tension Board 2, Decoy and Grasshopper now show the Tall/Wide chips, not just the Kilter Homewall. And they work **with no signal**: once you've downloaded a board, tall/wide filtering runs on-device. Flip on a filter that still needs a connection (drafts, beta, zones, holds) while offline and the screen now tells you why, with a tap to clear it, instead of just coming up empty.

https://claude.ai/code/session_01PFXvVk8uYmiju11gwEdmTp

### DIFF
--- a/packages/board-constants/src/__tests__/product-sizes.test.ts
+++ b/packages/board-constants/src/__tests__/product-sizes.test.ts
@@ -3,9 +3,12 @@ import {
   KILTER_HOMEWALL_LAYOUT_ID,
   KILTER_HOMEWALL_PRODUCT_ID,
   getLayoutName,
+  getTallWideScope,
   isKilterHomewallTallSizeId,
   isKilterHomewallWideSizeId,
 } from '../product-sizes';
+
+const EMPTY_SCOPE = { narrowerSizeIds: [], shorterSizeIds: [], hasNarrower: false, hasShorter: false };
 
 describe('getLayoutName', () => {
   it('returns the human-readable name for a known board + layoutId', () => {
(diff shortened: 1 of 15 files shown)


marcodejongh:
Addressed the review: added the explicit `compatible_size_ids IS NOT NULL` guard on the Postgres tall/wide predicates (commit 41fbea3) so they match the mobile offline path — behavior-neutral, just makes the NULL parity obvious. Left the fail-closed `boardName as BoardName` cast (getTallWideScope no-ops on an unknown board) and skipped a component test for the display-only offline empty state as a follow-up. Marking ready for review.

PR #3500, opened 2026-07-07
fix(db): offline sync perf + hardening follow-ups from #2785 review

### LINKED ISSUE #3475: offline sync: perf + hardening follow-ups from #2785 review (indexes, trigger WHEN guard, toggleFavorite upsert, GraphQLError typing)
Grouped follow-ups from #2785's final review rounds — none merge-blocking at current scale, all worth doing:

**Indexes**
- `board_climbs_sync_cursor_idx` / `board_climb_stats_sync_cursor_idx` should lead with `board_type` (`(board_type, updated_at, sync_seq)`) since every sync query filters on it first.
- `sync_deletions` prune (`DELETE WHERE deleted_at < cutoff`) can't use the `(user_id, deleted_at, id)` index — add `(deleted_at)`.
- The deletions pull's `user_id = $x OR user_id IS NULL` would benefit from two partial indexes.
- Mobile `boardsesh_ticks` has no `(board_type, climbed_at DESC)` index for date-ordered logbook queries.

**Triggers**
- `trg_board_climbs_set_sync_fields` has no WHEN guard, so internal-only writes (`synced`, `sync_error` from kilter-sync) advance `sync_seq` and re-ship unchanged rows to every client. Scope it to client-visible columns.
- `boardsesh_ticks` is the only user-owned sync table without a `set_updated_at` trigger — all current writers set it explicitly, but a future writer that forgets produces rows the cursor skips forever.

**Backend polish**
- `toggleFavorite` is check-then-act without a transaction; concurrent calls can race a unique violation. `INSERT ... ON CONFLICT` / delete-returning is the clean fix.
- `deleteTick`/`updateTick` throw plain `Error` (surfaces as INTERNAL_SERVER_ERROR) where siblings use `GraphQLError` with typed extensions; classification-wise the outcome (dead-letter) is right, but typed errors are the file's convention.
- `trg_playlist_climbs_delete` could add an `owner_id IS NULL → RETURN` guard as belt-and-braces (only reachable via orphaned `playlist_ownership`; the parent-uuid guard already covers the cascade path).
- Mutation-queue `update` branch for ticks should whitelist fields if/when offline tick edits are added (currently unreachable).

marcodejongh:
## Summary

Implements the grouped follow-ups from #2785's final review rounds. Closes #3475.

**Migration `0146_offline_sync_followups`** (drizzle-generated, then hand-edited per the 0144 pattern; opens with the retry-protected up-front `LOCK TABLE` block):

- `board_climbs_sync_cursor_idx` / `board_climb_stats_sync_cursor_idx` now lead with `board_type` — every `syncClimbs`/`syncClimbStats` pull filters on it first, so a per-board pull becomes a single ordered range scan instead of a post-filtered walk of the whole multi-board index.
- New `sync_deletions_deleted_at_idx` for the daily prune (`DELETE WHERE deleted_at < cutoff`), which no existing index served. (The issue's "two partial indexes" item was already half-landed: 0145 added the NULL-scope partial; the `user_id = $x` branch rides the existing composite.)
- `trg_board_climbs_set_sync_fields` gains a WHEN guard: `to_jsonb(OLD/NEW)` minus `{synced, sync_error, updated_at, sync_seq}` — exactly the columns `syncClimbs` doesn't ship. Fails closed: a future column change still fires (worst case an unneeded re-ship, never a skipped one).
- `trg_board_climb_stats_set_sync_fields` gains `WHEN (OLD.* IS DISTINCT FROM NEW.*)`. This is the sleeper win: the Kilter catalog sync rewrites **every** stats row per run with mostly identical values (`GREATEST`/`COALESCE` upsert), so each catalog cron was bumping `updated_at`+`sync_seq` across ~the whole table and re-shipping it to every offline client.
- `boardsesh_ticks` gets the missing `set_updated_at` safety-net trigger, guarded so bookkeeping-only writes (`aurora_*`/`kilter_*` columns, none of which `syncTicks` ships) never re-ship a user's logbook — but a content write that forgets `updated_at` can no longer strand a row the cursor skips forever.
- `log_deletion_playlist_climbs` gains an `owner_id IS NULL → RETURN` guard: an orphaned `playlist_ownership` no longer emits a `user_id = NULL` tombstone, which every client (not just the owner) would have applied. Migration `0147_playlists_tombstone_owner_guard` extends the same guard to `log_deletion_playlists` (review follow-up — the parent trigger had the identical NULL-owner shape).

**Backend:**

- `toggleFavorite` is now insert-first (`ON CONFLICT DO NOTHING ... RETURNING`, falling through to the delete): two concurrent toggles can no longer race the SELECT and collide on `unique_user_favorite`.
- `deleteTick`/`updateTick` throw `GraphQLError` with `TICK_NOT_FOUND`/`FORBIDDEN` extensions, matching the file's `resolveBetaLinkTickContext` convention. No drainer change: string codes classify as non-retryable → dead-letter, the right outcome for a 404/403 on replay.

**Offline-sync package:**

- The mutation-queue `UpdateTick` branch (unreachable today) whitelists `UpdateTickInput`'s exact field set instead of spreading the payload, so stray keys can't fail GraphQL validation and dead-letter the mutation once offline tick edits land.
- Mobile SQLite migration v3 adds `idx_ticks_board_climbed_at (board_type, climbed_at DESC)` for date-ordered logbook reads.

The backend test schema (`schema-sql.ts`) now mirrors the 0144 sync-field/updated_at triggers (previously absent there) at their 0146 state, so the trigger guards are exercised by real Postgres in CI.

## Release Notes

## Test plan

- `vp check`, `vp run typecheck` — green.
- `vp test run --project backend` — 2101/2101, including new `sync-trigger-guards.test.ts` (internal-only write → no cursor bump; no-op rewrite → no bump; real change → bump; bookkeeping-only tick write → `updated_at` untouched; forgotten-`updated_at` content write → bumped) and new `sync-mutations.test.ts` cases (toggle on/off row counts, `TICK_NOT_FOUND`/`FORBIDDEN` extensions, orphaned-ownership tombstone skip).
- `vp test run --project offline-sync` — 138/138, including the new `handlers.test.ts` whitelist cases and the v3 migration test (fresh install + v2-stamped upgrade both end with the index).
- `vp run test:mobile` — 3574/3574.
- `packages/db` native suite (local-only) — 165 pass / 0 fail.
- Migration applied end-to-end: `vp run db:up` picked up 10 pending migrations (0137→0146) against the Tailscale dev DB with real board data — lock block, index drop/create, trigger swaps and function replace all ran clean; verified the three triggers exist with WHEN clauses and the three indexes have the new definitions.

https://claude.ai/code/session_01A8jn4zWn2SdQaex9Fsx8K7

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -546,6 +546,7 @@
       "name": "@boardsesh/offline-sync",
       "version": "1.0.0",
       "devDependencies": {
+        "@boardsesh/shared-schema": "workspace:*",
         "@types/node": "^25.3.3",
         "typescript": "~6.0.3",
       },
--- a/packages/backend/src/__tests__/schema-sql.ts
+++ b/packages/backend/src/__tests__/schema-sql.ts
@@ -747,6 +747,12 @@ export const schemaSQL = `
     "deleted_at" timestamp DEFAULT now() NOT NULL
   );
 
+  -- Mirrors 0146: sync cursor indexes lead with board_type; deleted_at serves
+  -- the daily prune's DELETE WHERE deleted_at < cutoff.
+  CREATE INDEX IF NOT EXISTS "sync_deletions_deleted_at_idx" ON "sync_deletions" ("deleted_at");
+  CREATE INDEX IF NOT EXISTS "board_climbs_sync_cursor_idx" ON "board_climbs" ("board_type", "updated_at", "sync_seq");
+  CREATE INDEX IF NOT EXISTS "board_climb_stats_sync_cursor_idx" ON "board_climb_stats" ("board_type", "updated_at", "sync_seq");
+
   CREATE OR REPLACE FUNCTION log_deletion_playlists() RETURNS TRIGGER AS $$
   DECLARE
     owner_id text;
--- a/packages/backend/src/__tests__/schema-sql.ts
+++ b/packages/backend/src/__tests__/schema-sql.ts
@@ -755,6 +755,11 @@ export const schemaSQL = `
     FROM playlist_ownership po
     WHERE po.playlist_id = OLD.id AND po.role = 'owner'
     LIMIT 1;
+    -- Mirrors 0147: an orphaned playlist must not emit a user_id=NULL
+    -- (global) tombstone.
+    IF owner_id IS NULL THEN
+      RETURN OLD;
+    END IF;
     INSERT INTO sync_deletions (table_name, record_id, user_id)
     VALUES (TG_TABLE_NAME, OLD.uuid, owner_id);
     RETURN OLD;
--- a/packages/backend/src/__tests__/schema-sql.ts
+++ b/packages/backend/src/__tests__/schema-sql.ts
@@ -781,6 +781,11 @@ export const schemaSQL = `
     FROM playlist_ownership po
     WHERE po.playlist_id = OLD.playlist_id AND po.role = 'owner'
     LIMIT 1;
+    -- Mirrors 0146: an orphaned playlist_ownership must not emit a
+    -- user_id=NULL (global) tombstone.
+    IF owner_id IS NULL THEN
+      RETURN OLD;
+    END IF;
     INSERT INTO sync_deletions (table_name, record_id, user_id)
     VALUES (TG_TABLE_NAME, v_playlist_uuid || ':' || OLD.climb_uuid, owner_id);
     RETURN OLD;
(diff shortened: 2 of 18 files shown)


marcodejongh:
Addressed the review:

- **`log_deletion_playlists` parity** — done in `674a91e5b` as custom migration `0147_playlists_tombstone_owner_guard` (same owner guard, `schema-sql.ts` mirror, integration test that orphans ownership and deletes the playlist directly). Kept as a separate migration rather than editing 0146 because 0146 was already applied to the shared dev DB; 0147 landed there via `vp run db:up`.
- **`board_climb_stats` guard including `updated_at`** — deliberate asymmetry, leaving as-is: `OLD.* IS DISTINCT FROM NEW.*` is free on the catalog sync's whole-table rewrite (~10⁵–10⁶ rows per run), while the jsonb-minus form would materialize two jsonb docs per row on exactly that hot path. No writer sets `updated_at` directly on stats, and if one ever does, the failure mode is a harmless extra re-ship with a NOW() timestamp — the safe direction.

marcodejongh:
Second round addressed in `03c4404ce`:

- **Missing `DROP TRIGGER IF EXISTS` in 0146** — added before the `boardsesh_ticks` trigger, matching the siblings. (Prod-safe either way: the postgres-js migrator runs each migration in a transaction, so a partial failure rolls back the whole file — but consistency and schema-sql-bootstrapped DBs make the DROP right.)
- **schema-sql.ts index drift** — mirrored `board_climbs_sync_cursor_idx` / `board_climb_stats_sync_cursor_idx` (board_type-leading) and `sync_deletions_deleted_at_idx` into the test schema.
- **Whitelist drift** — compile-time assertion in `handlers.test.ts` that `UPDATE_TICK_INPUT_FIELDS` matches `keyof UpdateTickInput` exactly, both directions. It lives in the test (with a type-only `@boardsesh/shared-schema` devDependency) rather than `handlers.ts` because consumers compile this package's `src/` directly and the runtime package intentionally has zero dependencies.
- **`RESTART IDENTITY` under parallel workers** — not an issue in this harness: `worker-db.ts` gives every vitest worker its own database (`boardsesh_backend_test_w<POOL_ID>`), files within a worker run sequentially, and the new tests only ever compare `sync_seq` relatively (same-row before/after), never across rows or absolutely.

PR #3508, opened 2026-07-07
fix(mobile): stop Android filter-chip taps falling through to climb rows

marcodejongh:
## Summary

On Android (Material variant), once the climb list scrolled so rows sat under the top chrome, tapping a filter chip (Grade, Filters, Show, …) pressed the climb row hidden underneath instead — navigating to that climb rather than opening the filter.

**Root cause** (confirmed against the vendored RNGH source, `GestureHandlerOrchestrator.kt`): react-native-gesture-handler's Android hit-tester ignores z-order and treats ViewGroups with no gesture handler and no background drawable as transparent — and its `box-none` branch skips the background check entirely. The Material top chrome was a `pointerEvents="box-none"` absolute overlay whose chips are a background-less `@expo/ui` Compose `Host`, so RNGH found nothing in the chrome, recorded the list row's `Gesture.Tap` underneath, and on release intercepted the stream — the Compose chip got ACTION_CANCEL, the row navigated. That's also why it only happened after scrolling: at list-top only content-inset padding sits under the chrome.

**Fix**: the opaque Material chrome bands now use `pointerEvents="auto"` so they swallow every touch in their bounds (`ClimbTopChrome`, `CollapsingTopChrome`, `HomeTopChrome`, `ProfileTopChrome`, `RecordTopChrome` — Material branches only; the liquid-glass islands keep `box-none` pass-through). The `FilterChipRow` Host also gets an explicit transparent background so it remains an RNGH hit-test target even if a `box-none` ancestor is ever reintroduced. Expected side effect (standard Material app-bar behaviour): a drag starting on the chrome no longer scrolls the list.

JS-only change — rides OTA, no native build needed.

## Release Notes

- Filter chips on Android now respond after you scroll the climb list, instead of opening the climb hiding underneath

## Test plan

Reproduced and verified on an Android emulator (Pixel AVD, dev-client + Metro):

- **Pre-fix repro**: scroll the climb list, tap the Grade chip → the hidden climb's detail opened (screenshotted).
- **Post-fix**: same tap opens the grade range rail; rail grade buttons select (list refetches with the grade filter); Filters chip opens the filter sheet; Show chip opens its dropdown; a climb row below the chrome still opens its detail; the list still scrolls; everything unchanged at list-top; board switcher works on a scrolled Discover tab.
- `vp run typecheck:mobile`, `vp run test:mobile` (3574 tests), `vp run check:mobile-bundle` — all green.

https://claude.ai/code/session_01GH6Hs9TgYddegExhpsWm8g

### DIFF
--- a/packages/mobile/src/components/chrome/CollapsingTopChrome.tsx
+++ b/packages/mobile/src/components/chrome/CollapsingTopChrome.tsx
@@ -116,7 +116,8 @@ export function CollapsingTopChrome({
   if (isMaterial) {
     return (
       <View
-        pointerEvents="box-none"
+        // NOT box-none — the opaque Material band must swallow touches; see ClimbTopChrome for the RNGH mechanism.
+        pointerEvents="auto"
         style={[
           styles.materialContainer,
           {
--- a/packages/mobile/src/components/feed/HomeTopChrome.tsx
+++ b/packages/mobile/src/components/feed/HomeTopChrome.tsx
@@ -75,7 +75,11 @@ function HomeTopChromeGlass({
       {/* Frost content scrolling under the chrome band, matching the other tabs. */}
       <ProgressiveBlur style={[styles.topBlur, { height: insets.top + TOP_ISLAND_BAND }]} />
       {/* Floating header: the user-avatar island (left) and the scope menu (a glass
-          title-menu pill, centred) over the blur — matching the other tabs. */}
+          title-menu pill, centred) over the blur — matching the other tabs.
+          Deliberately box-none (unlike the opaque Material band below): the glass
+          islands float over scrollable content, and the gaps between them must keep
+          passing touches through. The islands themselves carry backgrounds, so
+          RNGH's Android hit-test still targets them. */}
       <View
         pointerEvents="box-none"
         style={[styles.headerChrome, { paddingTop: insets.top + spacing[1] }]}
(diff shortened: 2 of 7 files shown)


marcodejongh:
Addressed the review in 1d927c2:

1. **Regression test added** — `climb-top-chrome.test.tsx` now asserts the Material container renders `pointerEvents="auto"` (the `View` mock exposes it as `data-pointer`), so a silent revert to `box-none` fails CI.
2. **Inner `box-none` wrappers documented** — comment in `ClimbTopChrome` explains they're safe: once the outer `auto` container claims the RNGH pointer, only RN hit-testing runs inside, and an inner control adopting an RNGH gesture later needs no change.
3. **`ProfileTopChrome` tabs row** — same one-line comment added.
4. **Drag-to-scroll from the chrome** — verified on the emulator during QA: a drag starting on the chrome no longer scrolls the list, drags on the list body scroll normally, pull-to-refresh unaffected. This matches stock Material app-bar behaviour (an opaque app bar isn't a scroll surface), so it's intentional; it's called out in the PR description and QA notes.
5. **Glass path asymmetry** — `HomeTopChrome`'s glass container now carries an explicit comment: it stays `box-none` deliberately because the floating islands must pass touches through the gaps between them, and the islands carry their own backgrounds so RNGH still targets them.

PR #3509, opened 2026-07-07
perf(mobile): faster play drawer open, lighter climb list and search

marcodejongh:
## Summary

Android performance audit follow-up (Pixel 8 felt sluggish vs iPhone in the climb list, play drawer, and search). Five independent fixes, risk-ascending, one commit each:

1. **Memoize board-art cache keys** (`use-native-climb-render.ts`) — `buildCacheKey` ran an fnv1a hash over the frames string on every render/FlashList recycle; now `useMemo`'d on the primitive inputs, parsed `setIds` shared.
2. **Kill per-keystroke dead work in climbs search** — `setSearchTextLength` re-rendered the whole `ClimbListInner` (and the un-memoized `ClimbTopChrome`) on every keystroke, for a recent-pills feature that only exists on the iOS 26 native-search path. Now a boolean, gated on that path; `ClimbTopChrome` is `React.memo`'d.
3. **Narrow queue subscriptions** — `PlayDrawer` and both always-mounted `QueueSheet` instances subscribed to the full `QueueContext`, so every queue mutation re-rendered the ~1040-line drawer plus two hidden sheets on the fling-commit frame. New `QueueDataContext` (`{queue, currentClimbQueueItem}`); hidden `QueueSheet`s freeze their data snapshot until presented; `QueueList` is `React.memo`'d.
4. **Lazy-mount PlayDrawer's five sub-sheets** — `ClimbActionsSheet`/`AddBetaVideoSheet`/`AngleSelectorSheet`/`LogAscentSheet`/`BleControlSheetHost` mounted five native Compose `ModalBottomSheet` hosts on drawer open with `visible={false}`. Now a once-latch mounts each on first open and keeps it mounted after.
5. **Board defer via rAF instead of InteractionManager** — the headline fix. `DeferredBoard` gated the board behind `InteractionManager.runAfterInteractions`, which on Android routinely waited out its 350ms fallback ceiling (docs/mobile-sheets-vs-routes.md explicitly warns against this). Now a single `requestAnimationFrame`, so the board lands ~2–3 frames after the tap. Also covers `IpadPlayPane`, which mounts `PlayDrawer` without play.tsx's rAF gate.

Everything else the audit checked (FlashList hygiene, O(1) logbook index, worklet gating, native-PNG board path, provider splits) was already compliant with docs/react-native-performance.md.

Note for measuring: the Pixel 8 dev phone runs a Debug dev-client with JS from Metro — compare release builds before judging device gaps.

## Release Notes

[...]

PR #3512, opened 2026-07-07
fix(mobile): play drawer occasionally opens with unlit holds on Android

marcodejongh:
## Summary

Follow-up to #3509 (the race fix missed the merge by minutes). Fixes the report that the play drawer occasionally (~1 in 10 climbs) shows the board without the lit holds on Android.

Root cause: when a holds-overlay render is still in flight and the user swipes to a climb whose overlay is already cached, the cached overlay shows instantly — then the slow render resolves and clobbers `nativeRender` state back to the *previous* climb's cache key. The key-match guard on `overlayUri` then returns `null` for the current climb, and since the effect deps haven't changed, nothing re-fires: the board sits unlit until the next swipe. The race predates #3509, but #3509's faster drawer open and freed-up JS thread widened the window enough to notice.

Fix: track the latest cache key in a ref (same staleness pattern the background-paths effect already uses) and discard resolutions that no longer match. The late result still lands in the sync cache, so swiping back to that climb is an instant hit.

The regression test drives the exact race with a controllable fake renderer, injected through a new test-only seam (`_setNativeModuleForTests`) because the hook loads the native module via a literal CJS `require` that vitest's mock registry can't intercept. Verified the test fails against the unguarded code and passes with the fix.

## Release Notes

- Fixed the play drawer sometimes showing the board without the lit holds on Android

## Test plan

- New `use-native-climb-render-race.test.tsx` reproduces the clobber (fails without the fix, passes with it)
- `vp run typecheck:mobile`, `vp run test:mobile` (3588 tests), `vp run check:mobile-bundle` — all green

https://claude.ai/code/session_017rRcNZWqQX2seq9uuaHcHF

### DIFF
--- a/packages/mobile/src/hooks/__tests__/use-native-climb-render-race.test.tsx
+++ b/packages/mobile/src/hooks/__tests__/use-native-climb-render-race.test.tsx
@@ -1,0 +1,128 @@
+// @vitest-environment jsdom
+import { describe, it, expect, vi, beforeEach } from 'vitest';
+import { renderHook, act, waitFor } from '@testing-library/react';
+
+// In-flight render race regression (play-drawer "unlit holds"): a slow native
+// render must not clobber nativeRender state after the hook's props have moved
+// to a different climb, or the key-match guard nulls overlayUri for the current
+// climb with nothing left to re-fire the effect.
+
+vi.mock('expo-file-system', () => ({
+  Directory: vi.fn(() => ({ exists: false, list: () => [] })),
+  Paths: { cache: { uri: 'file:///cache/' } },
+}));
+
+vi.mock('../../lib/board-details', () => ({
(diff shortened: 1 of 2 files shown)


marcodejongh:
Review addressed:
1. The reported "extra closing parenthesis" at line 106 is a false positive — that paren closes the surrounding `await waitFor(...)` call. CI is green with the full mobile suite passing, which a parse error would preclude.
2. `_setNativeModuleForTests` is now a no-op unless `__DEV__` (mobile vitest sets `__DEV__`, so tests are unaffected; release bundles can't reach the mutable seam). The unguarded-underscore-export pattern follows the file's existing `_inflightRendersForTests`/`_runWarmupForTests` convention, but the guard is a fair hardening since this one mutates state.

Issue #3540, opened 2026-07-07
import-aurora-board-unified re-run destroys Boardsesh-owned data: deletes user-created climbs and wipes boardsesh_ascensionist_count for the whole board


marcodejongh:
## Summary
The refresh path for decoy/touchstone/grasshopper/soill is fully destructive: clearBoardData deletes every board_climbs row for the board type (line 544) — including Boardsesh user-created climbs (user_id IS NOT NULL), which the dump cannot restore — and every board_climb_stats row (line 540). Re-inserted stats rows seed ascensionist_count = upstream only (lines 427-428) with boardsesh_ascensionist_count = NULL, so all accrued Boardsesh tick counts vanish from displayed totals and only self-heal per-climb when a future tick triggers recomputeClimbStats for that exact (climb, angle). This bypasses the GREATEST/ownership convention every other upstream writer follows.

The parallel re-run hazard in the deprecated MoonBoard import scripts is filed as its own issue.

## Failure scenario
Operator refreshes grasshopper from a new Aurora dump. The 1 user-created grasshopper climb in prod is permanently deleted (its ticks orphan — no FK protects them); the 82 stats rows carrying 83 Boardsesh ascents (decoy: 196 rows / 213 ascents) drop back to Aurora-only counts. Users who logged those climbs see their sends disappear from ascent totals, and boardsesh-owned FA/quality/difficulty derived by the recompute are also reset to dump values.

## Evidence
`packages/db/scripts/import-aurora-board-unified.ts:528`

```
clearBoardData (import-aurora-board-unified.ts:528-558): `await tx.delete(boardClimbs).where(eq(boardClimbs.boardType, boardName));` with no user_id filter; climb_stats mapRow (:427-428) `upstreamAscensionistCount: toNullableNumber(row.ascensionist_count), ascensionistCount: toNullableNumber(row.ascensionist_count)` — boardsesh term dropped. Prod at risk: decoy 213 boardsesh ascents / 701 ticks, grasshopper 83 / 505 ticks + 1 user-created climb, soill 1 / 10 ticks.
```

## Suggested fix
Scope the destructive clear to synced rows only (WHERE user_id IS NULL for board_climbs) and convert the climb_stats path to the monotonic upsert the moonboard catalog import already uses: upstream = GREATEST(incoming, existing), total = upstream + COALESCE(existing boardsesh, 0), preserving boardsesh_ascensionist_count and boardsesh-owned quality/FA.

[...]

PR #3556, opened 2026-07-08
feat: blend Boardsesh star ratings into climb quality

marcodejongh:
## Summary

Star ratings logged in Boardsesh now count toward the climb's displayed quality. Part of the 2026-07 sync data-correctness program (follows #3547/#3549; siblings #3553/#3555).

`board_climb_stats.quality_average` becomes a materialized **blend**, mirroring the ascensionist-count column split:

- New columns: `upstream_quality_average` (owned by the upstream sync writers), `boardsesh_quality_sum` + `boardsesh_quality_count` (owned by the tick recompute).
- Blend: `(upstream_avg × upstream_ascent_count + boardsesh_sum) / (weighted count)` — upstream ascent count is the weight proxy since manufacturers publish no rating counts. One vote per climber (latest rated native flash/send tick); imported ratings are excluded (they're already inside the upstream average); Boardsesh-owned climbs keep their plain tick average unchanged (verified byte-identical SQL on the owned branch).
- One shared SQL fragment (`blendedQualityAverageSql`) is used by every writer — recompute (single + bulk), aurora shared-sync, kilter catalog-sync + stats-repair, and the moonboard import scripts. Each writer's *resolved* upstream count is likewise centralized in one fragment reused by the count SET, the materialized total, and the blend weight, so #3553's GREATEST-removal composes automatically regardless of merge order.
- Bonus: `board_climb_ratings` finally has a read path — ticks pulled from Kilter (which carry no per-log quality) now show the star rating you gave the climb over there, via `effectiveQuality = COALESCE(tick.quality, your rating at that climb+angle)` on the logbook resolvers (1:1 join on the unique index, angle-scoped, no N+1; edit flows keep writing raw quality).

Migrations `0166` (schema; quality-range check constraint verified retained in the snapshot), `0167` (init upstream side: 786,943 rows, guarded — not idempotent once blends land), `0168` (blend backfill: 3,256 keys with native ratings, value-idempotent). Chunk sizes sized against the 2-minute prod statement timeout. Journal gap 0157→0166 is deliberate — #3553 owns 0158–0162, #3555 owns 0163–0165; whichever lands last reconciles the snapshot chain.

Known approximation (locked design): the upstream weight is the ascent count, so heavily-repeated climbs need many Boardsesh votes to move while rarely-repeated ones move fast.

Re-running a direct-Aurora board import (`import-aurora-board-unified`) no longer wipes Boardsesh-side stats: after the reinsert it recomputes every ticked key's Boardsesh ascent count, quality votes, and blended `quality_average` in the same transaction (closes the stats half of #3540; climb deletion remains open there).

Follow-up noted: the mobile *offline* logbook doesn't ship `board_climb_ratings` yet (graceful null; online path works everywhere).

## Release Notes

- Your star ratings finally move the needle — rate a climb in Boardsesh and the stars everyone sees shift with it
- Climbs nobody rated before show stars as soon as someone in your crew rates them
- Ticks synced from Kilter now show the stars you gave the climb over there

## Test plan

- Blend math hand-verified per branch (mixed, NULL-upstream → pure native, native-empty → pure upstream, both-absent → NULL, no div-by-zero) and asserted in real-DB provenance-matrix tests: native rated tick blends in, imported rated tick doesn't, re-tick counts latest vote only, owned climbs unchanged, single-key ≡ bulk
- Real-Postgres migration replay incl. double-apply guard proof and post-GREATEST weight proof
- effectiveQuality: 7 backend tests (fallback, angle-scoping, null-safety); mobile surfaces null-safe when offline
- db 77, kilter/aurora/shared-schema 316, targeted backend files 79/79 in isolation (full-suite failures are the pre-existing board-presence sandbox flakes, untouched by this PR); `vp run typecheck` clean

https://claude.ai/code/session_017vqeM9YNFL3a1GDEjhKuNY


marcodejongh:
Both review rounds addressed (59599df48, 75671ec2c):

- **Aurora re-import zeroing Boardsesh terms (medium):** fixed — after the reinsert, the script bulk-recomputes every ticked key on the imported board in the same transaction, restoring the Boardsesh ascent count, quality votes, and blended average (closes the stats half of #3540; climb deletion remains open there). Verified the tombstone-suppression GUC doesn't gate the recompute's sync-field triggers, so offline clients still receive the corrected rows.
- **"0167 cursor loop can hang" — refuted, documented:** PL/pgSQL `SELECT INTO` sets its targets to NULL on a zero-row result (they don't retain prior values), and 0167 already exits on `v_page_count IS NULL OR = 0` — the exact pattern from the merged 0150/0157 whose replays terminate. A fully-skipped page also advances fine (`bounds` derives from `page`, not `upd`). Header note added so this stays documented.
- **userTicks effectiveQuality:** added (same join as the other three resolvers) + a test that another user's rating can't leak in.
- **Ratings unique index:** verified (`board_climb_ratings_user_climb_angle_idx`, unique on exactly the join columns, present in schema + snapshot + test mirror) and now named in the resolver comment.
- **Missing tests:** added — rated attempt never feeds the blend; `id DESC` tie-break when two rated sends share `climbed_at`.
- **sqlText helper:** extracted to a shared `@boardsesh/db/test-utils` used by all three test files; blend caller-contract comment and 0166 EOF newline done.

Backend targeted files 82/82, db suite 77, full typecheck clean.

marcodejongh:
Round 3 addressed (8110d7401 + 34ff6d553):

- **MoonBoard importer inconsistency:** fixed — the problems importer now resolves `coalesce(excluded.upstream_quality_average, stored)` ONCE and reuses it for both the SET and the blend's upstream term, so a NULL-rated re-run preserves the last-known rating and the written column can never disagree with the blend weight (same pattern as the catalog importer).
- **"Comment convention violations" — refuted:** the quoted rule doesn't exist in CLAUDE.md (grep for it returns nothing; the only "comment" mentions are the PR-lifecycle and /ota-preview sections). The repo's established idiom is the opposite — `unified.ts` on main carries the 25-line `board_climb_aliases` design block and three multi-line "⚠️ FK managed manually" warnings. The doc blocks follow house style and stay.
- **userTicks / callUserTicks — refuted (misread):** the resolver has `effectiveQuality` at queries.ts:336 with the ratings join at :365, and the test at tick-queries.test.ts:1453-1466 exercises every declared helper field — asserting both the fallback (owner's rating surfaces) and the cross-user leak guard.
- **Blank-line nit:** fixed, plus a formatting pass on tick-queries.test.ts (the actual CI lint failure on the previous head).

30/30 touched backend tests, typecheck clean.

marcodejongh:
Round 4 addressed (ac53b94a7):

1. **Quality drops to NULL at zero upstream weight (moderate):** fixed — `blendedQualityAverageSql` now wraps the weighted division in `COALESCE(…, upstream_quality_average)`, so a manufacturer-rated climb with zero ascent weight and no Boardsesh votes (0/0) falls back to its raw upstream average instead of clearing to NULL. A genuinely unrated climb (upstream NULL) still resolves to NULL. Covered by a fragment-shape test plus a real-DB case (upstream 3.5 / weight 0 / no votes → 3.5, not NULL).
2. **Migration ordering — acknowledged, no code change:** the journal gap is intentional; the merge contract in the PR description already states this PR must not deploy to prod before #3553/#3555. Whichever of the four lands last reconciles the snapshot chain.
3. **boardsesh_quality_sum/count semantics on owned climbs (minor):** fixed properly rather than documented — the recompute (single + bulk) and 0168 now NULL those columns for owned climbs, since owned climbs are never blended. The columns now carry one consistent meaning (non-owned blend inputs). Owned tests updated to assert the NULL.

31/31 across the migration-replay + recompute provenance-matrix suites; typecheck clean.

marcodejongh:
All substantive CI checks are green (test-backend, test-default, lint, typecheck, i18n, mobile-bundle, ci-status). The one red check is `claude-review`, which failed to *generate* a review (it posted several truncated/placeholder comments this session) — an infra flake, not an unaddressed finding. All four review rounds' findings are resolved. Marking ready for review.

PR #3558, opened 2026-07-08
perf(shared): batch multi-row SQLite upserts in offline-sync

marcodejongh:
## Summary

First phase of the offline-sync first-download optimisation (snapshot-bootstrap plan). `upsertDocuments` in `@boardsesh/offline-sync` previously issued one `INSERT OR REPLACE` `runAsync` call per synced document — up to 500 sequential JS→SQLite bridge calls per page, hundreds of thousands across a first board download. Rows are now chunked into multi-row `VALUES` statements sized `floor(999 / columnCount)` (SQLite's default bind-variable limit), so a 27-column `board_climbs` page goes from 500 statements to 14 (~37 rows per statement).

Semantics are unchanged and covered by tests: page-wide column allowlist, NULL fill for documents missing a page-wide key, whole-row REPLACE, unknown-column drift reporting (deduped), and the one-exclusive-transaction-per-page boundary. The JS→SQLite value coercion is extracted as a pure exported `toSqliteValue()` for reuse by the upcoming snapshot export job.

Follow-ups in this series: nightly per-layout SQLite snapshot artifacts on Tigris + client bootstrap that replaces the ~400-page first crawl with a single CDN-cached download.

## Release Notes

## Test plan

- `vp test run --reporter=agent --project offline-sync` — 153 tests / 12 files pass, including new coverage: chunk-size math (27→37, 1→999, >999→1), exact parameter ordering across chunk boundaries, NULL fill, unknown-column skip + drift-once, `toSqliteValue` coercion table, and a real `node:sqlite` round-trip of 100 rows × 27 columns with boundary-row spot checks.
- `vp check packages/shared/offline-sync/src` — 0 errors.
- `vp run typecheck` — green.

### DIFF
--- a/docs/offline-sync-plan.md
+++ b/docs/offline-sync-plan.md
@@ -468,7 +468,8 @@ async function syncTable(
     // bind-variable limit (see upsertDocuments in pull-client.ts).
     await db.withExclusiveTransactionAsync(async (tx) => {
       for (const rows of chunk(documents, multiRowChunkSize(columns.length))) {
-        await tx.runAsync(buildMultiRowInsertSql(tableName, columns, rows.length), flattenRowValues(rows, columns));
+        const values = rows.flatMap((row) => columns.map((column) => toSqliteValue(row[column])));
+        await tx.runAsync(buildMultiRowInsertSql(tableName, columns, rows.length), values);
       }
     });
 
--- a/docs/offline-sync-plan.md
+++ b/docs/offline-sync-plan.md
@@ -462,14 +462,15 @@ async function syncTable(
 
     if (documents.length === 0) break;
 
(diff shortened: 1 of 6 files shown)


marcodejongh:
Addressed the review:

1. **Stale doc** — fixed in 98b6456ef: the `offline-sync-plan.md` pseudocode now shows the one-transaction-per-page multi-row VALUES batching that `upsertDocuments` actually does.
2. **"Premature" exports of `toSqliteValue`/`multiRowChunkSize`** — the consumer is not hypothetical: the snapshot export job that reuses both is already implemented and lands as the next PR in this stack (branch `feat/board-snapshot-export`). Keeping the exports here avoids a churn commit there.

PR #3559, opened 2026-07-08
feat(backend): nightly board-snapshot export to S3 + shared manifest contract

marcodejongh:
## Summary

Phase 2 of the snapshot-bootstrap plan (stacked on #3558). Adds the server half: a nightly job that exports each `(board_type, layout_id)` catalog into a prebuilt SQLite artifact using the client's own DDL, gzips it, and uploads it to the S3-compatible bucket (Tigris) under `board-snapshots/v1/`, writing a small mutable `manifest.json` last so readers only ever see a manifest whose artifacts are fully uploaded.

- `packages/backend/src/scripts/export-board-snapshots.ts` — streams both tables inside one REPEATABLE READ transaction per layout via the read-replica seam, shapes rows with the exact `normalizeRow`/`toIso` the sync resolvers use, applies the same 30s stability window, and records per-table watermarks (max `(updated_at, sync_seq)` of the exported rows, as a bigint-safe decimal string) in `snapshot_meta`. `--dry-run` / `--board` / `--layout` for manual runs.
- Row shaping extracted to `row-normalize.ts` (pure refactor) so script and resolvers cannot drift.
- `@boardsesh/offline-sync` gains the `SnapshotManifest` types + `parseSnapshotManifest` the Phase 3 client bootstrap will consume.
- `uploadToS3` gains a `contentEncoding` passthrough (artifacts upload with `ContentEncoding: gzip`, immutable 1y cache headers; manifest `max-age=300`).
- `.github/workflows/export-board-snapshots.yml` — daily 07:15 UTC + `workflow_dispatch`, `environment: Production`.

**Golden parity test** (`snapshot-export-golden.test.ts`): proves the artifact path and the real resolvers→`pullSync` path produce identical SQLite rows (booleans, int[]/text[] arrays, NULL arrays, fractional timestamps, multi-angle stats, climb-without-stats), and that watermarks/stability-window exclusions match. This is the standing contract keeping the two paths convergent.

New Production secrets needed before the cron does anything useful: `AWS_S3_BUCKET_NAME` / `AWS_ACCESS_KEY_ID` / `AWS_SECRET_ACCESS_KEY` / `AWS_ENDPOINT_URL` (reuses the existing backend s3.ts names).

## Release Notes

## Test plan

- Golden parity test 3/3 against the dockerised test Postgres (`vp test run --reporter=agent --project backend`, full backend suite 2126 pass).
- `vp test run --reporter=agent --project offline-sync` — 153 pass.
- `vp run typecheck` + `vp check` on touched files — green.
- Live `--dry-run --board kilter` against a dev DB: 2 layouts discovered and built, gzip ~1.3KB for the seed layout, clean exit with no AWS credentials.

### DIFF
--- a/.github/workflows/export-board-snapshots.yml
+++ b/.github/workflows/export-board-snapshots.yml
@@ -1,0 +1,49 @@
+name: Export Board Snapshots
+
+# Nightly: build a per-(board, layout) SQLite snapshot of board_climbs +
+# board_climb_stats, gzip it, and upload to Tigris/S3 under board-snapshots/v1/.
+# A freshly downloaded board on mobile warms from these artifacts instead of
+# paging the whole catalog over GraphQL, then resumes an incremental pull from
+# the watermarks the manifest records. Idempotent — each run mints new
+# content-addressed artifacts and rewrites the manifest last.
+on:
+  schedule:
+    - cron: '15 7 * * *' # 07:15 UTC daily (off-peak; distinct from the 06:00/06:30 refresh crons)
+  workflow_dispatch:
+
+concurrency:
+  group: export-board-snapshots
+  cancel-in-progress: false
+
+jobs:
+  export:
+    runs-on: ubuntu-latest
+    timeout-minutes: 30
+    # DATABASE_URL is read-only-sufficient (snapshots only SELECT); READ_REPLICA_URL
+    # is optional and offloads the scan to a replica when present.
+    environment: Production
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+
+      - uses: oven-sh/setup-bun@v2
+
+      - uses: voidzero-dev/setup-vp@v1
+        with:
+          node-version: 'lts'
+          cache: true
+
+      - name: Install dependencies
+        run: bun install --frozen-lockfile
+
+      - name: Export board snapshots
+        working-directory: packages/backend
+        env:
+          DATABASE_URL: ${{ secrets.DATABASE_URL }}
+          READ_REPLICA_URL: ${{ secrets.READ_REPLICA_URL }}
+          AWS_S3_BUCKET_NAME: ${{ secrets.AWS_S3_BUCKET_NAME }}
+          AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
+          AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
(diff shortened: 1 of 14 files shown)


marcodejongh:
All review findings addressed in ac910ef01:

- **Manifest overwrite on filtered runs**: `runExport` now merges into the previously published manifest (upsert by `boardType+layoutId`), so `--board kilter` refreshes Kilter entries and preserves every other board's verbatim. Only unfiltered runs drop entries for layouts that no longer exist in the DB. Covered by `snapshot-export-run.test.ts`.
- **Per-layout resilience**: each pair is try/caught; one bad layout no longer blocks the rest. The merged manifest still publishes the successful entries (failed layouts keep their previous immutable artifacts) and the run exits non-zero listing the failures, so the workflow still reds.
- **Integer-strict validation**: `layoutId`/`bytes`/`schemaVersion`/`rowCount` now require `Number.isInteger` (+ dedicated validator tests).
- **Parser coupling**: investigated — both branches of `createReadPool()` construct the drizzle wrapper before returning the raw pool, so the parsers are always installed; the guarantee is now documented in `postgres.ts` and the redundant `createReadDb()` call was dropped.
- **Artifact accumulation**: clean unfiltered runs now prune artifacts unreferenced by the just-written manifest and older than 14 days (defensive — prune failures never fail the run). New `listS3Objects` helper.
- Nits: WHERE consts, discovery filters pushed into SQL.

Backend snapshot tests 13/13, offline-sync 160/160, typecheck green.

marcodejongh:
Both addressed in cb73debb3: a filtered run that matches zero pairs now throws (typo'd `--board` can't silently leave a board stale), and the manifest `contentEncoding` doc comment now spells out the at-rest vs after-transport-decoding distinction — the Phase 4 client sniffs the gzip magic bytes on the downloaded file rather than trusting the field.

marcodejongh:
Both fixed in c4d72d0c7:

- **P1 (replica watermarks)**: the export now reads the **primary** pool only — `READ_REPLICA_URL` dropped from the script and workflow. Your analysis was right: the cursor is write-time ordered while a replica snapshot is commit-order consistent, so lag beyond the 30s stability window could publish a watermark covering rows the artifact doesn't contain, permanently skipping them for every bootstrapped client. Rationale documented at the pool call site.
- **P2 (unreadable manifest)**: new `getFromS3Strict` distinguishes NoSuchKey/404 (→ merge against empty) from any other read failure (→ abort before ANY upload, filtered and unfiltered). An invalid-but-present previous manifest is fatal on filtered runs, warn+continue on unfiltered. Matrix encoded in the doc comment; 6 new tests pin it.

PR #3560, opened 2026-07-08
feat(shared): snapshot bootstrap for first-time board downloads

marcodejongh:
## Summary

Phase 3 of the snapshot-bootstrap plan (stacked on #3559). When a newly enabled board scope has no sync checkpoints yet, `pullSync` now runs a **bootstrap** phase before deletions: fetch the `board-snapshots` manifest, download the layout's prebuilt SQLite artifact (one CDN-cached GET), ATTACH it, `quick_check` + row-count verify, then in **one exclusive transaction** copy the scope's rows and stamp both table checkpoints to the artifact watermarks — plus rewind the global deletions cursor to the older watermark so tombstones from the snapshot-age window replay against the imported rows. The existing paged pull then only fetches the post-snapshot delta instead of crawling ~200k rows 500 at a time.

Correctness properties (all test-pinned):
- **Delta continuity**: watermark checkpoints are the exact `{updatedAt, syncSeq-string}` shape the resolvers' Zod cursor schema accepts; a row at cursor == watermark is in the artifact and strict-`>` delta pulls start past it with no gap or overlap.
- **Size-filter parity**: the import WHERE mirrors `boardClimbsScope` exactly (json_each containment == Postgres `@>`, NULL `compatible_size_ids` excluded on both paths, MoonBoard un-scoped); stats semi-join mirrors the resolver's EXISTS.
- **Safety rails**: bootstrap only when BOTH scope checkpoints are absent (mid-crawl/completed users untouched); failures record an attempt and skip the scope's paged pull for up to 2 cycles (a paged first page would permanently disqualify bootstrap); manifest-has-no-entry falls straight to the paged path without burning attempts; wipe-epoch changes roll the whole import back; deletions rewind is atomic with the import.

Platform I/O stays injected (`SnapshotSource`); nothing is wired on mobile yet, so this is dormant until the next PR — with `snapshotSource` undefined the control flow is byte-identical to today.

## Release Notes

## Test plan

- 177 offline-sync tests pass (24 new: happy path, size filtering incl. MoonBoard + NULL arrays, artifact reuse across sizes, schema drift both directions, corrupt artifact → 2-attempt fallback, checkpoint-exists skip, manifest-miss matrix, deletions rewind + BigInt seq compare, wipe mid-import rollback, watermark→delta cursor continuity asserted against the mock server).
- Opus review pass: all 12 hunt-list invariants confirmed (cursor Zod shape verified against `SyncCursorInputSchema`, resolver SQL parity line-by-line).
- `vp run typecheck` full monorepo + `vp check` green.

### DIFF
--- a/packages/shared/offline-sync/src/index.ts
+++ b/packages/shared/offline-sync/src/index.ts
@@ -55,6 +55,7 @@ export {
   MAX_BOOTSTRAP_ATTEMPTS,
   SnapshotWipedError,
   SnapshotSchemaStaleError,
+  SnapshotPermanentMissError,
 } from './sync/snapshot-bootstrap';
 export type {
   SnapshotSource,
--- a/packages/shared/offline-sync/src/index.ts
+++ b/packages/shared/offline-sync/src/index.ts
@@ -46,6 +46,20 @@ export { isRetryable, isNetworkError, getErrorStatus } from './mutation-queue/er
 // --- Pull sync -----------------------------------------------------------------
 export { pullSync, toSqliteValue, multiRowChunkSize } from './sync/pull-client';
 export type { SyncProgress, SyncOptions, SchemaDriftReporter } from './sync/pull-client';
+export {
+  bootstrapScopeFromSnapshot,
+  getBootstrapAttempts,
+  recordBootstrapAttempt,
+  markBootstrapDone,
+  isBootstrapDone,
+  MAX_BOOTSTRAP_ATTEMPTS,
+  SnapshotWipedError,
+} from './sync/snapshot-bootstrap';
+export type {
+  SnapshotSource,
+  SnapshotBootstrapResult,
+  SnapshotBootstrapErrorReporter,
+} from './sync/snapshot-bootstrap';
 export { startSyncScheduler, triggerSync } from './sync/sync-scheduler';
 export type { SyncProgressSink, SchedulerTriggers, SchedulerOptions, DrainQueue } from './sync/sync-scheduler';
 export {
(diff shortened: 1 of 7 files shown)


marcodejongh:
Both findings fixed:

- **Wipe-guard bug**: `startEpoch` is now captured before the first await (before the ATTACH), so a wipe cycle that starts and completes during quick_check/snapshot_meta reads is caught — new regression test flips the epoch mid-quick_check and asserts nothing is imported.
- **Telemetry cause**: manifest-fetch and download failures now thread the real error into `onSnapshotBootstrapError`'s `cause` instead of `null`.

178 offline-sync tests green.

marcodejongh:
Both P2s addressed in bfbeee6a5:

- **Stale snapshot schemas**: `verifySnapshotMeta` now rejects an artifact whose `schema_version` is older than the client's `LATEST_SCHEMA_VERSION` (`SnapshotSchemaStaleError`) — importing one would NULL-fill newer columns and stamp the cursor past them forever. Treated as a permanent miss for the run: no attempt burned, the scope's paged pull runs in the same cycle, and the nightly export rebuilds artifacts at the new schema. Newer-than-client artifacts remain fine (extra columns drop via the shared-column intersection).
- **Query invalidation**: a successful bootstrap now busts the board-table query caches itself, covering the case where the snapshot fully satisfies the scope and the delta pull returns zero documents (syncTable's arrivals-only invalidation never fires).

Both covered by new tests; 181 offline-sync tests green.

marcodejongh:
The cached-download-failure telemetry minor is fixed in ccce05c23 — the failure cause is now cached per layout alongside the null result, so a second size of the same layout reports the real error instead of `cause: null`.

marcodejongh:
Both fixed in c91e8366d: `compareCheckpoints` now tolerates a corrupt `syncSeq` (sorts as 0 instead of a BigInt SyntaxError crash, matching the timestamp tolerance), and `rewindDeletionsCheckpoint`/`compareCheckpoints`/`DELETIONS_CHECKPOINT_KEY` are no longer exported from the package root — they were only ever consumed inside the engine.

marcodejongh:
The `compareCheckpoints` BigInt guard already landed in c91e8366d (crossed with this review run) — invalid `syncSeq` now sorts as 0 via a try/catch helper instead of throwing. The export-narrowing from the same round is in that commit too.

PR #3611, opened 2026-07-14
fix(mobile): collapse filter pickers to dropdowns + pin sheet footer

marcodejongh:
## Summary

Two fixes to the mobile **Filters** bottom sheet (both visible in the reported screenshots):

**1. Grade accuracy & Status pickers were always expanded.** They rendered as full inline Off/Loose/Moderate/Tight (and status) lists, eating several rows and looking inconsistent with the compact "Hold types / Board region" rows above them. Added a `variant: 'inline' | 'menu'` prop to `RadioGroup` and switched both to `menu`:
- iOS: SwiftUI `pickerStyle('menu')`
- Android: Material 3 `ExposedDropdownMenuBox` (anchor row + `arrow-drop-down` icon)

Each is now a single row that opens a native dropdown only when tapped. Default stays `inline`, so every other `RadioGroup` is unchanged.

**2. The "Show N climbs" button floated well above the sheet's real bottom** — a visible dead gap under it — and the same footer bug shows up on other `%`-detent sheets. Root cause: `useSheetColumnStyle` pinned the sheet's column to a **JS-estimated** detent height that can't match the native detent across devices, so it came out ~90pt short. As of `@expo/ui` 57.0.3 the library bounds sheet content to the detent natively (`RNHostView.frame(maxHeight: .infinity)` + size-report back to Yoga, plus a `flexGrow:1;height:0` wrapper in `BottomSheet.ios.tsx`), so the estimate is obsolete and was the cause. Simplified the shared hook to a plain `flex:1` column that fills the real detent — fixing the footer position on **every** sheet that uses it, with nothing to calibrate per-device.

## Release Notes

- Filter sheet: grade accuracy and status now tuck into tidy dropdowns you open when you want them
- Fixed the "Show climbs" button floating in the middle of the filter sheet — it sits at the bottom where it belongs

## Test plan

- [x] `RadioGroup` logic unit tests pass (4/4)
- [x] Static verification of every `@expo/ui` native API used (iOS `pickerStyle`, Android `ExposedDropdownMenuBox`/`ExposedDropdownMenu`/`DropdownMenuItem`/`menuAnchor`; `RNHostView.swift` + `BottomSheet.ios.tsx` detent-bounding behaviour)
- [ ] `vp run typecheck:mobile` — **blocked locally by heavy parallel-workload CPU contention on this machine; relying on CI**
- [ ] `vp run test:mobile` (use-sheet-column-style + climb-filter-sheet) — same, via CI
- [ ] iOS simulator screenshot on iPhone 16 Pro (reported device) + a small device (13 mini / SE): footer snug above the home indicator, no dead gap; grade accuracy & status open native dropdowns — **to run once the machine frees up / after CI**

Note: kept as a draft until CI is green and the on-device screenshot check is done.


marcodejongh:
Thanks — addressed in c634f27:

**1. White-only dropdown icon (Android light mode)** — Fixed. The `<Icon>` now takes `tint={systemColors.secondaryLabel}` from the theme, which overrides the XML's baked-in white so the arrow is visible in both light and dark.

**2. Dead code in the column-style hook** — Fixed. The hook was reduced to returning a constant, so I removed it entirely: it's now an exported `SHEET_COLUMN_STYLE` (file renamed to `sheet-column-style.ts`). Dropped the `snapPoints`/`options` params, the `SheetColumnStyleOptions` type, and — since the column no longer needs the resting detent — the now-dead `activeIndex` state in both `Sheet.tsx` and `ModalSheet.tsx` (there were 3 call sites, not 1). Removed the obsolete detent-height unit test.

**3. Menu-variant tests** — By design in this repo the native `@expo/ui` render path isn't unit-tested: any suite that renders a `RadioGroup` is redirected to `test/radio-group-stub.tsx` via a vite alias because the SwiftUI/Compose trees can't mount under Vitest's node env. The menu variant's only new logic is the `expanded` toggle (native-only state), and its selection reuses `makeRadioSelectHandler`, which is already covered in `radio-group.test.ts`. The stub is variant-agnostic, so the existing `climb-filter-sheet.test.tsx` screen tests exercise selection through the menu call sites too. Adding a native-mock test just for a boolean toggle would break from that established pattern for little value, so I've left the coverage as-is.

**4. Per-item `onClick` closure** — Acknowledged, left as-is. `DropdownMenuItem.onClick` takes no arguments, so a per-option closure is unavoidable without the API handing the value back, and the menu is unmounted whenever it's closed — so this is bounded and low-impact.

marcodejongh:
Closing — the last commit's detent-height formula is based on an incorrect premise about how SwiftUI's `.fraction()` detent resolves, verified against the reporter's own screenshots.

**1. `.fraction(f)` multiplies against `maximumDetentValue` (the `.large` detent), not the window.** WWDC21 session 10063 says explicitly "a fraction of the fully expanded sheet frame"; an Apple Frameworks Engineer confirms this on the developer forums (thread 724906); the `UISheetPresentationController.Detent.custom` docs describe "a height within the safe area of the sheet"; and three independent open-source reimplementations agree — `boscojwho/SheetInteraction`, `nathantannar4/Transmission`, and `react-native-screens`' `RNSFormSheetDetentResolver.mm` (`context.maximumDetentValue * fraction`, verbatim). This PR's "per Apple's docs" citation points at a page with no Discussion section describing this behavior — the claim isn't there.

**2. Measured directly against this PR's own screenshots** (#3330 and #3776, both iPhone SE 3, 375×667 @2x): the `maximumDetentValue` model predicts a card top of 110.7pt / height 548.3pt — within 1pt of the ~110pt / ~548pt measured on-screen. This PR's full-window formula predicts 66.7pt — off by 43pt. Refuted.

**3. This PR's formula errs LONG on every device checked** (+36pt SE, +42pt 13 mini, +52pt 16 Pro Max, +30pt iPad) — which clips pinned footers off the bottom of the detent. That's #3330 reopened, not fixed. Main's current formula errs 6–32pt SHORT: a cosmetic dead gap below the footer, the correct failure direction per the code's own documented policy ("erring short beats a clipped footer").

**4. Blast radius beyond the filter sheet**: `AccessibilitySettingsScreen.tsx` has three `%`-detent sheets with pinned footers (line 453: 95%, plus 549 and 605). The 95% sheet would take the worst hit from this PR's LONG error.

**5. The device verification this PR's own test plan calls for never happened.** `typecheck:mobile`, `test:mobile`, and the iOS screenshot are all unchecked in the PR body ("to run once the machine frees up"). The "~80pt short" figure in the last commit's message matches the arithmetic delta for an iPhone 16 exactly — a computed number, not a measured one.

**6. The `@expo/ui`-changed-behavior justification doesn't hold either.** `RNHostView.swift` and `BottomSheet.ios.tsx` are byte-identical across 56.0.19 / 57.0.2 / 57.0.3 — nothing in the library changed between this branch's own "flex:1 now works" and "flex:1 doesn't work" commits.

To be fair to the work here: the `RadioGroup` `variant="menu"` half was a reasonable read of the code at the time it was written, and the instinct that something's wrong with the filter sheet's footer was correct. Both halves are moot now regardless of the formula issue — the RadioGroup half targets Accuracy/Status `RadioGroup` call sites that #3769 (`fc53502`, 2026-07-18) deleted when it flattened the filter dialog into five sections (Accuracy is now a `SegmentedControl`; Status is folded into a `Collection` control). `RadioGroup` has zero render call sites left anywhere in the app on current `main`.

The real #3776 diagnosis — a ~142pt dead-space bug this formula tweak was never going to be big enough to close — is written up on that issue, with a follow-up (#3922) filed for the fix that can actually work: measuring the real layout instead of guessing it.

PR #3612, opened 2026-07-14
fix(mobile): write SecureStore items with AFTER_FIRST_UNLOCK

### LINKED ISSUE #3602: iOS: SecureStore keychain read fails in background ('User interaction is not allowed') — 183 users
## Summary

`expo-secure-store` reads fail with **"Calling the 'getValueWithKeyAsync' function has failed → User interaction is not allowed"** when the app reads a token from the keychain while backgrounded/locked. This is the highest-impact production error by users.

- **Sentry:** [BOARDSESH-7P](https://boardsesh.sentry.io/issues/BOARDSESH-7P)
- **Impact:** 183 users / 222 events (last 18 days), ongoing
- **Platform:** iOS (production, `release com.boardsesh.app@2.2.1+`)
- **Signal:** every event has `in_foreground: false`

## Root cause

`packages/mobile/src/lib/auth-store.ts` writes JWT/refresh/expiry via `SecureStore.setItemAsync(...)` with **no `keychainAccessible` option**, so the items default to `WHEN_UNLOCKED`. Any read that runs while the device is locked (background token refresh, Live Activity, background fetch) hits the keychain and iOS rejects it with *"User interaction is not allowed."*

## Proposed fix

Write the auth tokens with `keychainAccessible: SecureStore.AFTER_FIRST_UNLOCK` so background reads succeed once the device has been unlocked at least once since boot:

```ts
await SecureStore.setItemAsync(JWT_KEY, jwt, { keychainAccessible: SecureStore.AFTER_FIRST_UNLOCK });
```

Apply consistently to all three keys in `auth-store.ts`. `auth-provider.tsx` already catches the reject on one read path, but other read paths don't — this removes the failure at the source.

Ships **OTA (JS-only)**, no native rebuild required.

## Acceptance

- Tokens persisted with `AFTER_FIRST_UNLOCK`.
- Background token read while locked no longer throws.
- BOARDSESH-7P event rate drops to ~0 after OTA rollout.

marcodejongh:
## Summary

Fixes the highest-impact production error: iOS `expo-secure-store` reads reject with **"User interaction is not allowed"** when the app reads a keychain item while the device is locked/backgrounded — background token refresh, Live Activity, WS reconnect, background fetch. Sentry [BOARDSESH-7P](https://boardsesh.sentry.io/issues/BOARDSESH-7P): **183 users / 222 events over 18 days**, every event `in_foreground: false`.

Root cause: every `SecureStore.setItemAsync(...)` was called without a `keychainAccessible` option, so items defaulted to `WHEN_UNLOCKED` and any read before the device was unlocked rejected.

`keychainAccessible` is a **write-time** option (reads inherit whatever the item was written with), so the fix goes at every write: a shared `SECURE_STORE_WRITE_OPTIONS` constant carrying `AFTER_FIRST_UNLOCK`, applied to **all** SecureStore writers — not just the three auth tokens — to kill the whole error class:

- `auth-store.ts` (JWT / refresh / expiry) — the issue's core
- `session-store.ts` (session id, read by the Live Activity in the background)
- `party-profile-store.ts` (async + the sync analytics-bootstrap write)
- `last-search-store.ts`, `last-grade-store.ts`, `recent-filter-store.ts`
- `preferences/secure-store-adapter.ts` (generic adapter — onboarding, locale, theme, changelog-seen, …)

The `in` guard in the shared module keeps the constant read from throwing under test doubles that mock `expo-secure-store` without re-exporting `AFTER_FIRST_UNLOCK`.

**Migration:** items already on disk keep `WHEN_UNLOCKED` until their next write (next token refresh / sign-in). The existing `auth-provider` catch + AppState re-check already prevent a hang meanwhile, so the error rate decays to ~0 as installs cycle a write — no forced re-write needed.

Ships **OTA (JS-only)** — no native rebuild.

Closes #3602

## Release Notes

- Fixes a bug where the app could sign you out or get stuck loading after being backgrounded on a locked iPhone.

## Test plan

- New `auth-store.test.ts` asserts `storeTokens` writes each key with `{ keychainAccessible: AFTER_FIRST_UNLOCK }` and round-trips through the getters.
- `party-profile-store.test.ts` updated to assert the sync write carries the option too.
- `vp run typecheck:mobile` — clean (0 errors).
- `vp test run --project mobile` on the touched store tests — green (the only reds were `Test timed out in 5000ms` machine-load flakiness, not assertions).
- `vp check` — clean.

## Acceptance (post-merge)

Publish OTA to `production`; watch BOARDSESH-7P event rate trend to ~0 as installs cycle a token write.

https://claude.ai/code/session_01KhD6G26G2Ss6j7Taj8nGiK

### DIFF
--- a/packages/mobile/src/lib/__tests__/auth-store.test.ts
+++ b/packages/mobile/src/lib/__tests__/auth-store.test.ts
@@ -1,0 +1,68 @@
+import { describe, it, expect, vi, beforeEach } from 'vitest';
+
+// A sentinel for the AFTER_FIRST_UNLOCK constant so the test can assert the
+// exact keychainAccessible value the writers pass without depending on the
+// native module's real numeric constant.
+const AFTER_FIRST_UNLOCK = Symbol('AFTER_FIRST_UNLOCK');
+
+vi.mock('expo-secure-store', () => {
+  let storage: Record<string, string> = {};
+  const setItemAsync = vi.fn(async (key: string, value: string) => {
+    storage[key] = value;
+  });
+  return {
+    AFTER_FIRST_UNLOCK,
+    getItemAsync: vi.fn(async (key: string) => storage[key] ?? null),
+    setItemAsync,
+    deleteItemAsync: vi.fn(async (key: string) => {
+      delete storage[key];
+    }),
+    __reset: () => {
+      storage = {};
+      setItemAsync.mockClear();
+    },
+  };
+});
+
+async function secureStore() {
+  return (await import('expo-secure-store')) as unknown as {
+    __reset: () => void;
+    setItemAsync: ReturnType<typeof vi.fn>;
+  };
+}
+
+const JWT_KEY = 'boardsesh_jwt';
+const REFRESH_TOKEN_KEY = 'boardsesh_refresh_token';
+const EXPIRES_AT_KEY = 'boardsesh_token_expires_at';
+
+beforeEach(async () => {
+  (await secureStore()).__reset();
+});
+
+describe('auth-store', () => {
+  it('writes every token with keychainAccessible: AFTER_FIRST_UNLOCK', async () => {
+    const store = await secureStore();
+    const { storeTokens } = await import('../auth-store');
+
+    await storeTokens('jwt-value', 'refresh-value', '2026-08-01T00:00:00.000Z');
+
+    // Every write is what keeps a locked-device background read from rejecting
+    // with "User interaction is not allowed" (issue #3602). Reads inherit the
+    // accessibility the item was written with, so each key must carry it.
+    const options = { keychainAccessible: AFTER_FIRST_UNLOCK };
+    expect(store.setItemAsync).toHaveBeenCalledWith(JWT_KEY, 'jwt-value', options);
+    expect(store.setItemAsync).toHaveBeenCalledWith(REFRESH_TOKEN_KEY, 'refresh-value', options);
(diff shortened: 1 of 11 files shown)


marcodejongh:
Thanks for the review. Addressed the coverage gap and a note on the other two:

**Test coverage gap** — pushed `secure-store-writers.test.ts` (commit 4fa38d9) pinning that `session-store`, `last-grade-store`, and the generic `secure-store-adapter` each write with `keychainAccessible: AFTER_FIRST_UNLOCK`. That plus the existing `auth-store` / `party-profile-store` tests now covers 5 of the 7 writers; I left the two complex-signature ones (`last-search-store`, `recent-filter-store`) uncovered since exercising them needs full `ClimbFilters` / `BoardSearchConfig` fixtures for what is the identical one-line wiring.

**`keychainAccessible: undefined` fallback** — the `in` guard exists purely so the constant read doesn't throw under Vitest factory mocks that don't re-export `AFTER_FIRST_UNLOCK`; `__DEV__` is also true in tests, so a dev-only warning would fire as noise on every such mock rather than flag a real regression. On device the native module always exports the constant, and the new writer test asserts the real path resolves to it (not `undefined`), so the silent-fallback branch is test-only. Kept as-is.

**Migration window** — agreed, and called out in the PR description: on-disk items keep `WHEN_UNLOCKED` until their next write, so BOARDSESH-7P decays rather than dropping to zero at rollout. The `auth-provider` AppState catch prevents a hang meanwhile.

Note: the `publish` (per-PR OTA preview) job is failing on a transient S3 `SlowDown` rate-limit from the `boardsesh-ota` bucket — unrelated to this change; the correctness gates (typecheck, lint, tests, mobile-bundle) are green.

PR #3622, opened 2026-07-15
feat(mobile): follow on find, offer offline, easier My Boards management

marcodejongh:
## Summary

Finding a board you don't own now **adopts** it end to end:

- **Follow on find.** Picking a board from the gym finder (`/gyms`) or the *Nearby* carousel in the board picker now follows it, so it lands in **My Boards** instead of just becoming the local active board. Mobile had no follow hook at all before this (only unfollow) — added `useFollowBoard` on top of the existing idempotent `followBoard` mutation.
- **Offer offline on select.** Right after you pick a new board, we ask whether to make it available offline (reusing the existing download confirm). If you don't have its data yet, one tap downloads the whole catalog.
- **"Keep boards offline by default" setting.** New toggle in More → Offline. Turning it on downloads every board already in My Boards immediately, and every board you find from then on downloads automatically with no prompt.
- **Easier My Boards management.** Added a traditional iOS-style **Edit / Done** button in the header that shows a persistent per-row remove control (Unfollow for followed boards, Delete for owned), so removing a board no longer depends on the swipe. The swipe itself was also loosened — the old vertical-drift tolerance made it almost impossible to trigger over a scrolling list.

The offline offer, the settings toggle, and the auto-download are all gated behind the existing `offline-board-downloads` feature flag. The offline enable + sync-kick logic was extracted out of the My Boards screen into a shared `useBoardDownloads` hook so the adopt flow and the settings toggle reuse it.

## Release Notes

- Find a new wall and it drops straight into your boards, ready to pull down for offline.
- Turn on "Keep boards offline" in Settings to download every board you use so browsing and logging sends works with no signal.
- Manage your boards the easy way: tap Edit to unfollow or delete, or swipe — the swipe actually catches now.

## Test plan

Validated on this branch:
- `vp test run --project mobile` — 3808 tests pass (incl. new `adopt-found-board-decision` tests + edit-mode `board-manage-row` tests)
- i18n catalog completeness — 28 tests pass (new keys in en-US/es/fr, plural + ICU parity)
- `vp run check:mobile-bundle` — OK (13.9 MB)
- `vp check` — format clean, no new lint issues

On-device QA to exercise (simulator/emulator):
- **Find Gym / Nearby**: pick a board you don't own → it appears in My Boards (followed) with a toast; with the offline flag on and auto-offline off, the "Download <board>?" dialog appears on the landing screen; confirming shows the row go Downloading → Available offline in My Boards. Picking a board you already own/follow does **not** re-prompt.
- **Settings toggle**: turn on "Keep boards offline" → every current My Board starts downloading; finding a new board then auto-downloads with no prompt. Toggling off stops future auto-downloads but keeps existing ones.
- **My Boards**: the swipe now catches with a natural drag; Edit reveals the red remove control per row → Unfollow (followed) and Delete-with-confirm (owned) both work; Done restores the list.

https://claude.ai/code/session_01EagWAd48Bs5MuftajqZm7Y

### DIFF
--- a/packages/mobile/app/(tabs)/profile/more.tsx
+++ b/packages/mobile/app/(tabs)/profile/more.tsx
@@ -12,7 +12,9 @@ import { useLocalePreference } from '../../../src/providers/i18n-provider';
 import { resolveLanguage, type LocaleOverride } from '../../../src/lib/i18n/locale-preference';
 import { openExternalUrl } from '../../../src/lib/open-url';
 import { useAuth } from '../../../src/providers/auth-provider';
-import { useProfile } from '../../../src/lib/graphql/hooks';
+import { useProfile, useMyBoards } from '../../../src/lib/graphql/hooks';
+import { useBoardDownloads } from '../../../src/offline/use-board-downloads';
+import { useSetting, setSetting } from '../../../src/settings';
 import { useConfirm } from '../../../src/providers/dialog-provider';
 import { useIsOffline } from '../../../src/hooks/use-is-offline';
 import {
@@ -75,6 +77,14 @@ export default function MoreScreen() {
   const garminWatchEnabled = useFeatureFlag('garmin-watch') === true;
   const offlineEnabled = useOfflineDownloadsEnabled();
   const confirm = useConfirm();
+
+  // "Keep boards offline by default" toggle. Turning it on downloads every board
+  // already in My Boards now (user chose this over a future-only default), and the
+  // adopt-on-select flow auto-downloads new boards from then on.
+  const [autoOfflineBoards] = useSetting('autoOfflineBoards');
+  const { enableBoardsOffline } = useBoardDownloads();
+  const { data: myBoardsConnection } = useMyBoards(undefined, { enabled: offlineEnabled && !!profile });
+  const myBoards = myBoardsConnection?.boards ?? [];
 
   // Offline sync-issues surface. Poll the dead-letter count only while online (the
   // section is hidden offline — a pending write offline is expected, not a "stuck"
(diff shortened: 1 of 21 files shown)


marcodejongh:
Thanks — addressed the review across two follow-up commits:

**Fixed**
- **Silent follow failure** → `useFollowBoard` now takes an `onFollowError` (config-level `onError`, so it survives the navigation-then-follow unmount like `onFollowed` does). The adopt flow reports it to Sentry and shows an error toast (`mobile.discovery.followError`, added to en/es/fr).
- **`myBoards ?? []` churn** → memoized with `useMemo` keyed on the connection, so the empty-while-loading fallback keeps a stable identity and the offline effect no longer runs every render.
- **Missing hook tests** → added `use-adopt-found-board.test.tsx` (9 cases): follow-if-new, owned/already-followed skips, the ask/auto/none offline branches, declined confirm, and the success/error toast callbacks.
- **Toggle-race + `followBoard` deps churn** (from the first-commit review) → already handled in `3974f5de`: an effect enables missing boards once My Boards resolves, and the adopt callback depends on the stable `followBoard.mutate`.
- **Fire-and-forget comment** → both `void adoptFoundBoard(...)` call sites now note that adopt handles its own errors and intentionally doesn't reach the surrounding catch.

**Not changed (verified correct)**
- `fr/boards.json` `"ajoutée"` — per `docs/i18n-french-glossary.md`, the board device is feminine (`une board trouvée`), so feminine agreement is the house rule, matching `es` `"añadido"` agreeing with masculine _el plafón_.

marcodejongh:
Addressed in `0d200d7f`:

- **Swipe scroll regression** → pulled `failOffsetY` back from 20px to **14px** and restored `activeOffsetX` to 10px. 14px still fixes the too-tight `[-5,5]` (which the normal wobble of a horizontal swipe tripped, so it never engaged) while a real vertical scroll crosses 14px in Y well before 10px in X and cedes. On-device flick validation is still worth a pass — the Edit button is the guaranteed path regardless of feel.
- **`missing` computed twice** → extracted a single memoized `missingOfflineBoards()` used by both the enable effect and the toast, so the count can't diverge.

Not changing, with reasons:
- **`mobile.more.offline.title` "missing"** → it's present in `common.json` for all three locales (`en-US: "Offline"`), and `catalog-completeness.test.ts` (28 passing) enforces parity — the key path matches the `useTranslation('common')` render.
- **`useMyBoards` on every More mount** → gated on `offlineEnabled && !!profile`, and the `['myBoards']` query is already warm from the boards hub / manage screen on any realistic path, so React Query serves cache rather than a fresh request. Lazily gating it on `autoOfflineBoards` would drop the immediate toast on the primary toggle-on action for no real saving.
- **`.click()` vs `fireEvent.press()`** → this test file drives jsdom with mocked RN primitives (`@testing-library/react`), not `@testing-library/react-native`, so `fireEvent.press` doesn't apply; `.click()` matches the file's existing convention.

PR #3625, opened 2026-07-15
feat(mobile): show the download size before an offline board download

### LINKED ISSUE #3616: Offline mode: Show estimates before a user triggers the download
Users would like to know how big the dataset they're about to download will be, we should make this information available to them

marcodejongh:
## Summary

Enabling a board offline downloads its whole catalog as a pre-built SQLite artifact. That is **270 MB** for Kilter Original — and until now the confirm dialog said nothing about it until the download was already running.

The snapshot manifest already publishes `bytes` per `(boardType, layoutId)`, so this is purely a client-side read of data we already ship. No export-job or manifest-shape change.

Real sizes from the live manifest, for scale:

| Board / layout | Download |
| --- | --- |
| kilter 1 | 270 MB |
| moonboard 2 | 55 MB |
| tension 9 | 47 MB |
| kilter 8 | 29 MB |

### The size is only quoted when it's true

This is the part worth reviewing. A board toggled off **keeps its rows and checkpoint on purpose** (`use-offline-board.ts`), so re-enabling it resumes as a small delta — quoting the full artifact size there would be a lie. `estimateScopeDownload` owns that rule and mirrors `runBootstrapPhase`'s eligibility check exactly, returning `unknown` (→ existing sizeless copy, never a guess) when:

- there's no manifest (not fetched, unreachable, or snapshot bootstrap off — the paged crawl has no byte total)
- the scope already has a checkpoint on either board table
- bootstrap attempts are at `MAX_BOOTSTRAP_ATTEMPTS`
- the layout isn't exported yet, or its artifact is schema-stale

`findSnapshotEntry` / `isSnapshotEntryUsable` are now **shared with `runBootstrapPhase`**, so the UI can't drift from the engine about which artifact a scope would download — that drift is the only way this feature starts lying.

The manifest is fetched via React Query on screen mount (reusing the engine's own `SnapshotSource.fetchManifest` rather than a second fetch path) and read through a ref at tap time, so the dialog never awaits a fetch and a refresh can't churn the handler identity.

Also adds a **placeholder-parity check** to the i18n catalog test: key parity alone let a translation silently drop `{{size}}` and render a sentence with no number. Zero pre-existing violations; verified it fails when `{{size}}` is removed.

Out of scope: live byte progress during the download (`File.downloadFileAsync` has no progress callback — needs its own change). The spinner + "Downloading board…" is unchanged.

Closes #3616

## Release Notes

- See how big a board is before you download it — no more surprise 270 MB on cellular

## Test plan

Automated (all green — CI: **2449 passed, 0 failed**):

- `vp test run --project offline-sync` — 216 passed, incl. 16 new `snapshot-estimate` tests covering every `unknown` branch, and the existing bootstrap/pull-client suites through the shared-lookup refactor
- `vp test run --project i18n` — 56 passed; the new placeholder-parity test was verified to fail on a deliberately dropped `{{size}}` and pass when restored
- `vp run test:mobile` — 14 `format-bytes` tests, every assertion locale-pinned
- `vp run check:mobile-bundle` — **ios OK (13.8 MB), android OK (13.9 MB)**
- `vp run typecheck:offline-sync`, `vp run typecheck:mobile`, `vp check`, `vp run check:i18n:orphans` — clean (orphans confirms the new key has a live reference, so the `t()` call and catalog key match)

**Not done: the on-simulator visual check.** I tried and couldn't get there — the dev-clients cached locally are from other worktrees and none would load a bundle from this worktree's Metro (one `.app` cache was also corrupt, missing its Info.plist). It needs a dev-client built from this branch (~30 min). Both platform bundles building OK is good evidence the code is sound, but it is not the same as seeing the string rendered, so I'm flagging it rather than implying otherwise.

QA steps are in `.boardsesh/qa-notes.md`. The one trap: `EXPO_PUBLIC_SNAPSHOT_BASE_URL` must be set for Metro or the size never renders — that's by design (no base URL → no manifest → sizeless copy), not a bug.

This ships OTA (no native change), so it's also loadable from the `pr-3625` channel.

https://claude.ai/code/session_01X2HXRmzriT3zJhrsZw8iZm

### DIFF
--- a/docs/board-snapshots.md
+++ b/docs/board-snapshots.md
@@ -220,6 +220,22 @@ pre-import empty result set.
 
 ## Mobile wiring
 
+- **Pre-download size estimate (the manifest's second consumer)**: the My Boards confirm dialog
+  ("Download {board}?") quotes the artifact size before the user commits —
+  `entry.bytes` for the layout, formatted by `packages/mobile/src/lib/format-bytes.ts`. The manifest
+  comes from `useSnapshotManifest()` (`packages/mobile/src/offline/use-snapshot-manifest.ts`), a React
+  Query wrapper around the **same** `SnapshotSource.fetchManifest` the engine is handed (`staleTime`
+  5 min, mirroring the object's `max-age=300`), warmed on screen mount so the dialog never awaits a
+  fetch. The decision of _whether a number can be quoted at all_ lives in
+  `estimateScopeDownload` (`snapshot-estimate.ts`), which mirrors `runBootstrapPhase`'s eligibility
+  rules and returns `unknown` — falling back to sizeless copy — whenever the paged crawl would run
+  instead: no manifest, an existing checkpoint on either board table (a re-enabled board resumes as a
+  small delta, so the full artifact size would be a lie), attempts at `MAX_BOOTSTRAP_ATTEMPTS`, no
+  entry for the layout, or a schema-stale entry. `findSnapshotEntry`/`isSnapshotEntryUsable` are
+  shared with `runBootstrapPhase` so the UI can never disagree with the engine about which artifact a
+  scope would download. Note `bytes` is the **stored** object size: correct as a download figure while
+  artifacts stay identity-encoded, and an undercount of the on-disk file the day `--gzip` ships.
+
 - **Per-connection ATTACH invariant (BOARDSESH-AA)**: expo-sqlite's
   `withExclusiveTransactionAsync` runs its task on a **new native connection**
   (`useNewConnection: true`), and SQLite `ATTACH`es are per-connection — anything attached on the
--- a/packages/mobile/app/boards/manage.tsx
+++ b/packages/mobile/app/boards/manage.tsx
@@ -1,4 +1,4 @@
-import { useCallback, useEffect, useLayoutEffect, useMemo, useState } from 'react';
+import { useCallback, useEffect, useLayoutEffect, useMemo, useRef, useState } from 'react';
 import { Pressable, RefreshControl, StyleSheet, View } from 'react-native';
 import { FlashList } from '@shopify/flash-list';
 import { useNavigation, useRouter } from 'expo-router';
@@ -101,7 +101,14 @@ export default function ManageBoards() {
   const { enableBoardsOffline } = useBoardDownloads();
   // Warmed here on mount so the download-size estimate is already in cache when a
   // row's toggle is tapped — the confirm dialog must never wait on a fetch.
+  //
+  // Mirrored into a ref because the toggle handler only reads it at tap time:
+  // keeping it out of that handler's deps means a manifest refresh can't churn the
+  // callback identity and re-render every memoised row. The ref also always reads
+  // the freshest manifest, which is exactly what the tap wants.
   const snapshotManifest = useSnapshotManifest();
+  const snapshotManifestRef = useRef(snapshotManifest);
+  snapshotManifestRef.current = snapshotManifest;
   const db = useSQLiteContext();
   const syncStatus = useSyncStatus();
   const [enabledBoards] = useSetting('syncEnabledBoards');
@@ -151,7 +158,7 @@ export default function ManageBoards() {
         getBootstrapAttempts(db, key),
       ]);
       const estimate = estimateScopeDownload({
-        manifest: snapshotManifest,
+        manifest: snapshotManifestRef.current,
         boardType: scope.boardType,
         layoutId: scope.layoutId,
         hasExistingCheckpoint: !!climbsCheckpoint || !!statsCheckpoint,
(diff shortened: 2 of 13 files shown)


marcodejongh:
Both review points were real — fixed in f8f66e7.

**Test locale fragility** — good catch, and it pointed at a product bug rather than just a test one. `formatBytes` now takes an explicit locale and the caller passes the app's `i18n.language`. That fixes the `de`-host test failure *and* is the more correct behaviour: the size is interpolated into a translated sentence, so "1,2 GB" should sit next to Spanish copy even on an en-US device — which `toLocaleString(undefined)` would not have done. Every assertion now pins a locale, plus a new case covering `es`/`de` grouping.

**Sequential SQLite round-trips** — the three reads are independent, so they now run under one `Promise.all` instead of stacking in front of the dialog. I kept them in the handler rather than pre-loading at mount: the manifest is one fetch for the whole screen, but these are per-scope, so hoisting them would mean 3 × N queries on mount to save work on the one row the user actually taps.

marcodejongh:
All three addressed in ff1d2f5 — thanks, the ref one in particular was a fair call against the mobile performance checklist.

- **Type safety** — `queryFn` now returns early when there's no snapshot source instead of feeding `undefined` through an optional chain. `enabled` already gated it at runtime, but the type shouldn't lean on that.
- **Ref for the manifest** — the handler only reads the manifest at tap time, so it now reads `snapshotManifestRef.current` and drops the dep. Callback identity stays stable across manifest refreshes (no re-render of memoised rows), and the tap still gets the freshest manifest.
- **Test comment** — correct, `formatBytes(0)` hits the `<= 0` guard and renders "0 B", not "0 MB". Comment fixed; the assertion (that `bytes: 0` stays a `snapshot` answer rather than collapsing to `unknown`) was the actual point and is unchanged.

Also note this branch was rebased onto main earlier: `useBoardDownloads` had landed and taken over the enable+sync plumbing in `handleToggleOffline`, so the estimate now sits in front of `enableBoardsOffline(board)`. That conflict was also why CI had gone quiet — the PR was `CONFLICTING`, so GitHub couldn't compute a merge commit and never fired the `pull_request` runs.

marcodejongh:
Two of these taken in 380a8ef, one declined:

**Taken — `locale` is now required.** Agreed: an optional param with a docstring saying "you must pass this" is just a trap for the next caller. `formatBytes(bytes, locale)` is required now, and typecheck confirms the only caller already passes `i18n.language`.

**Taken — zero no longer clamped with negative/NaN.** You're right that `bytes <= 0` read as if zero were invalid, which contradicts `estimateScopeDownload` deliberately treating `bytes: 0` as a real estimate. Output is unchanged (`0` rendered `'0 B'` either way — via the clamp before, via the natural `B` branch now), but the code now says what it means. Worth noting the dialog showing "About 0 B." for a zero-byte artifact is the honest outcome: kilter layout 3 really does export 2 climbs, and a genuinely empty artifact is a manifest problem, not something to paper over.

**Declined — query key with the source URL.** `SnapshotSource` has no `url` field (it's `fetchManifest`/`downloadArtifact`/`deleteArtifact`), so `snapshotSource?.url` isn't available to key on. The URL behind it is `EXPO_PUBLIC_SNAPSHOT_BASE_URL`, inlined at build time and read into a module constant — it cannot change within a session, so there's no stale-source case to guard. The `offline-snapshot-bootstrap-v2` flag can flip at runtime, but that only toggles the source between defined/undefined at the *same* URL, which `enabled` already handles. Keying on a constant would add a moving part without removing a risk. If the base URL ever becomes dynamic, the key should change with it.

**Visual check** — correct, and it's flagged in the PR body rather than glossed. I attempted it and couldn't get there: the locally cached dev-clients are from other worktrees and none would load a bundle from this worktree's Metro (one `.app` cache was corrupt outright). It needs a dev-client built from this branch. Both platform bundles pass, but that is not the same as seeing the string, and I'd rather say so than imply otherwise.

marcodejongh:
Follow-up on the one gap I flagged (no on-simulator screenshot) — the root cause turned out to be structural, not something I was doing wrong:

`runtimeVersion` uses the **fingerprint** policy, so a cached dev-client only loads a bundle whose fingerprint matches its *own build*. Every `.app` I had locally was from another worktree, so none of them could ever load this branch's bundle — being OTA-compatible with `main` doesn't help, since it's the borrowed app's own fingerprint that has to match. (This was independently confirmed the same day while verifying #3630.) Closing that gap properly needs `vp run mobile:build-sim-app` from this worktree, ~3h on a loaded machine.

Given this is JS-only and ships OTA, the cheap path to eyes-on confirmation is the **`pr-3625` channel** (More → Development → OTA Channel Switcher) on a real device — seconds instead of hours. Worth doing before merge if anyone wants to see the string; the QA steps are in `.boardsesh/qa-notes.md`.

The one setup trap either way: `EXPO_PUBLIC_SNAPSHOT_BASE_URL` must be set, or the dialog correctly shows the old sizeless copy and it looks like the feature didn't land.

PR #3629, opened 2026-07-15
feat(backend): anonymous board-presence reads for public boards

marcodejongh:
## Summary

Backend groundwork (PR 4) for the gym-kiosk-dashboard epic: a public smart-TV kiosk page will anonymously display board-presence data for **public** boards, and two reads were still auth-gated.

- **`boardHistory` + `boardPresenceStats` are now auth-optional.** Both swap `requireAuthenticated` for the same `requireAnonReadableBoard` gate the other board-presence reads (`boardRecentClimbs`, `boardConnection`, `boardNowPlaying`) already use: anonymous callers can read public and system-shared boards, and a private board is masked as `NOT_FOUND` — the same error as a missing board, so its existence isn't revealed. Logged-in access is unchanged (membership-free, any active board). The durable-history **write** dwell-gate in `reportBoardClimb` is untouched.
- **Anon-tolerant read/subscribe rate limits bumped 30 → 60/min** (`boardNowPlaying`, `boardConnection`, `boardPresenceStats`): multiple gym TVs can sit behind one NAT and reconnect together after a network blip, and anonymous limits key on client IP. The 60/min limits stay as-is.
- **`boardLeaderboard` gets a `'day'` period** (`NOW() - INTERVAL '1 day'`, labelled "Today") for the kiosk's daily leaderboard, plus the anon-read gate and a 60/min rate limit it was previously missing entirely (pre-existing gap: any caller could query any board's leaderboard, unbounded).
- SDL descriptions + resolver docstrings updated; codegen output committed; `docs/websocket-implementation.md` Board Presence section updated (read auth relaxed, write dwell-gate unchanged).

**Exposure analysis:** no new data shape is exposed — the same board send-log / stats / leaderboard data was already anonymously readable for public boards via `boardRecentClimbs` (the hot cache of the identical `boardHistory` rows) and the ungated `boardLeaderboard`; this only aligns the durable variants with the existing gate and actually *tightens* `boardLeaderboard` for private boards. One retention-depth nuance worth calling out: `boardHistory` pages the *full* durable `board_climb_events` log with no upper bound, while `boardRecentClimbs` is capped to the ~50-entry / 1-week Redis window — anon access to `boardHistory` therefore reaches deeper into a public board's history than the live feed it mirrors.

**Consumer check:** the only existing `boardLeaderboard` consumer is `packages/web/app/components/board-entity/board-leaderboard.tsx`, which early-returns without an auth token — it never queries anonymously, so the new anon gate can't break it (and logged-in behaviour is a no-op passthrough).

## Review fixes

A Fable review found three issues, all fixed and pushed:

- **`boardLeaderboard`'s missing/deleted-board path threw a plain `Error`, not a masked `GraphQLError`.** The private-board path a few lines later already threw `GraphQLError('Board not found', { extensions: { code: 'NOT_FOUND' } })`, but the missing-board path threw a bare `Error('Board not found')` — over HTTP that gets masked to a generic `INTERNAL_SERVER_ERROR`/"Unexpected error.", so an anonymous caller could use the error shape itself as an existence oracle (private board vs. no board), and unknown-UUID probes generated Sentry noise. Now both paths throw the identical `GraphQLError` shape.
- **`requireActiveBoardById`'s `GraphQLError` lacked `extensions.code`.** This helper backs `boardHistory` / `boardPresenceStats` (and `reportBoardClimb`, `boardRecentClimbs`), and its sibling `requireAnonReadableBoard` already set `code: 'NOT_FOUND'` — the missing-code throw let `boardHistory`/`boardPresenceStats` leak "board exists but is private" vs. "board doesn't exist" via the wire `extensions`. Added the code for consistency; grepped backend tests for assertions depending on the absence of `extensions` on this path and found none.
- **Tests only asserted the error *message*, which is how the `boardLeaderboard` gap slipped through review the first time.** Strengthened the `boardHistory` / `boardPresenceStats` / `boardLeaderboard` anon-gate tests to assert a private board and a genuinely missing board produce an *identical* message **and** `extensions.code === 'NOT_FOUND'`.

Also: the `'day'` leaderboard period is now documented in the SDL as a rolling last-24-hours window (`NOW() - INTERVAL '1 day'`), not calendar "today" — codegen output regenerated and committed. And the NAT rationale comment on the WS-only `boardNowPlaying` rate limit was corrected: anonymous WS rate-limit keys are per-connection (`ctx.connectionId`), not per-IP, since the WS context never sets `clientIp` the way the HTTP-keyed reads (`boardConnection`, `boardPresenceStats`) do — so "multiple gym TVs behind one NAT" doesn't actually describe why this limit helps on the subscription path.

## Release Notes

## Test plan

- `vp run typecheck` — green.
- `vp check` — failure set byte-identical to clean `origin/main` in this environment (pre-existing, unrelated files); no new issues introduced.
- Full backend suite (`vp test run --project backend --reporter=agent`, on dedicated postgres+redis to avoid cross-session worker-DB contention in the dev sandbox): **2248 passed, 1 failed** — the single failure is the documented pre-existing sandbox flake (`board-presence.test.ts` create-race `user_boards_unique_slug` duplicate-key, fails on clean `main` too; the exercised `resolveBoardForConfig` path is untouched by this diff). All new tests pass:
  - anonymous `boardHistory` + `boardPresenceStats` succeed on a public board
  - both are masked as `NOT_FOUND` on a private board, indistinguishable from a missing board
  - authenticated access to the caller's own private board still works
  - `boardLeaderboard` `'day'` period scopes to the last 24h ("Today") and works anonymously on a public board
  - `boardLeaderboard` masks private boards as `NOT_FOUND` for anon but still serves the owner

**Review-fix validation:** `vp check` (0 new issues; same pre-existing, unrelated failure set as before) and `vp run typecheck` both green. `vp test run --project backend --reporter=agent src/__tests__/board-presence.test.ts` (single file only, per the dedicated-infra note above — the full suite collides with concurrent sandbox sessions): 81 passed, 1 failed, re-run twice with the identical result — the same documented pre-existing `user_boards_unique_slug` create-race flake noted above, in a test this diff doesn't touch. The strengthened tests (identical message + `extensions.code` for private vs. missing boards, across `boardHistory` / `boardPresenceStats` / `boardLeaderboard`) all pass.

https://claude.ai/code/session_01MaW2zXJkhwXJhyxYzSv9iH


marcodejongh:
Fixed in 3e48cc69d.

Collapsed `boardHistory` / `boardPresenceStats` / `boardRecentClimbs` (same shape) from two round-trips (`requireActiveBoardById` then `requireAnonReadableBoard`, both keyed on the same `boardId`) to one: `requireActiveBoardWithVisibilityById` does a single indexed by-id `SELECT` that also carries `isPublic`/`ownerId`, and `assertAnonReadableBoard` applies the existing gate logic synchronously against that already-loaded row instead of re-querying. `requireActiveBoardById`/`requireAnonReadableBoard` are untouched and still back the other callers (`reportBoardClimb`, `boardConnection`, `boardNowPlaying`), so no gate semantics changed anywhere.

Issue #3088, opened 2026-06-20
Error-tracking hygiene: stop logging expected/user-cancel events as exceptions and attribute BLE failures


marcodejongh:
## What's happening

Three separate sources are inflating PostHog error tracking with events that are not user-facing bugs — expected lifecycle states, user-initiated cancels, and an unattributed/double-logged BLE failure. Each one is real and verified in code, but none represents broken behaviour for users; they just drown real bugs and corrupt the failure metrics. Bundle the cleanups below.

### 1. iOS locked-keychain SecureStore read logged at `error` level
- **Stop logging:** `expo-secure-store getValueWithKeyAsync ... Caused by: User interaction is not allowed` (handled:true). This is iOS prewarming the app in the background before first unlock; the keychain read rejects, and auth already recovers cleanly on the next foreground.
- **Where / what to guard:** The guard already exists — `checkAuth()` in `packages/mobile/src/providers/auth-provider.tsx` (lines 149-162, added 2026-06-10) catches this rejection, releases the splash gate, and does NOT clear tokens; the AppState `active` re-check restores auth on unlock. The only gap is that line 157 calls `reportError(authCheckError)`, which routes through the raw `captureError` path and bypasses the `reportHandledError` noise policy in `packages/mobile/src/lib/error-reporting.ts` (lines 62-75). Route this specific rejection through `reportHandledError`, or add a warning-level / message-match downgrade for `User interaction is not allowed`.
- **Volume (30d):** 187 occurrences / 17 sessions / 15 users (iOS). Currently the highest-volume active issue. _posthogUrl: https://us.posthog.com/project/412845/error_tracking/019ec3e3-51f3-7583-8d82-3e1e56ee73ba
- _Note: PostHog groups this under the `DEVELOPER_ERROR` (react-native-google-signin) issue name by volume, but the latest events are the SecureStore reads — unrelated to Google sign-in._

### 2. BLE device-picker cancel logged as an error
- **Stop logging:** `Device selection cancelled` (handled:true). Fires when the user closes the board/device bottom sheet without picking — an expected cancel, not a failure.
[...]

PR #3653, opened 2026-07-15
fix(mobile): classify Android BLE connect failures by ble-plx errorCode

### LINKED ISSUE #3608: Connecting to the board fails 4x more often on Android than iOS (9.5% vs 2.3%) — and the reason we log is wrong
## What's happening

Tapping to connect the board fails about four times more often on Android than on iOS. Over 30 days, after stripping user cancels (picker dismissed) from both numerator and denominator:

- **Android RN: 246 technical connect failures / 2,599 attempts = 9.5%**
- **iOS RN: 192 / 8,304 = 2.3%**

Roughly 240 Android users hit this, spread across 25+ active days. Not a spike, not a handful of boxes.

And right now we can't say *why* Android fails, because the reason we log is a misclassification. The dominant Android bucket, `disconnected` (162 events / 79 users), has a visible iOS mirror in `connect_failed` (135 events / 78 users): the same failure mode, different bucket. The small Android `write_failed`-on-connect bucket (35 events / 6 users / 4 days) is the same fallthrough — no write has happened at connect time.

Two framing corrections worth carrying into the fix, both from the data:

- **Android is not failing on writes.** Over 30d, `Climb Sent to Board Failure`: iOS RN `write_failed` 695 / `write_timeout` 353 / `write_recovery_failed` 5; Android RN has **zero** write failures (only `incompatible_climb` 6). Porting the iOS write-stall recovery (#3181) to ble-plx would fix nothing measurable. This is a **connect-side** problem.
- **`board_not_found` is not elevated on Android** (iOS 53 events / 38 users vs Android 37 / 22).

## Evidence

- PostHog project 412845, `Bluetooth Connection Failed` + `Climb Sent to Board Failure`, last 30d (2026-06-14 → 2026-07-14), test accounts excluded.
- 246 Android technical failures / ~240 affected users. `user_cancelled` (1,872 fleet-wide) is excluded as expected behaviour, not a defect.
- Platform: Android (mobile RN). Both platforms run the **same** `useBoardBluetooth` hook and emit the **same** event (`packages/mobile/src/lib/ble/use-board-bluetooth.ts:971-977`), so the comparison is apples-to-apples at the event level; only the adapter differs.

## Suspected cause

- `packages/mobile/src/lib/ble/adapter-factory.ts:20-25` routes iOS to `NativeIosBleAdapter` (Swift `BoardBleManager`) and Android to `RNBleAdapter` (react-native-ble-plx).
- **`packages/shared/ble-protocol/src/connection-error.ts`** (`classifyBleFailure`) matches on Web Bluetooth / CoreBluetooth message strings only. ble-plx's `"Device … was disconnected"` / `"Cannot connect to device"` match none of them, fall through to `unknown`, then get re-bucketed by `classifyBleFailureReason`'s `/disconnected/` substring match. The classifier never consumes ble-plx's `BleError.errorCode`, which the adapter already has in hand at `packages/mobile/src/lib/ble/adapter.ts:418`. Mobile also calls `classifyBleFailure(error)` **without** `pairingStage` (`use-board-bluetooth.ts:907`), which widens the fallthrough.
- Once we can see real reasons, the plausible capability gap is **connect-side recovery**: iOS gets `reconnectToLastKnownBoard` and connection adoption free from `BoardBleManager`; the ble-plx path has no equivalent.

**Work, in order:**
1. Teach `classifyBleFailure` to consume ble-plx `BleError.errorCode` so Android connect failures land in real buckets instead of `disconnected` / `write_failed`.
2. Re-measure, then decide whether Android needs the connect-side recovery machinery iOS already has.

## Severity

P2 — a 9.5% failure rate on the app's single core function (lighting the board), across hundreds of Android users on shipping builds. Every failure is retryable with another tap, so it degrades rather than blocks.

## Related

- **#3088** already covers the web side: `Bluetooth Connection Failed` emits no `failureReason` even though `classifyBleFailure(error, pairingStage)` is computed one line earlier (`packages/web/app/components/board-bluetooth-control/use-board-bluetooth.ts:749-756`). Deliberately out of scope here. Note #3088 assumes mobile is the correct reference implementation for failure attribution — this issue shows that isn't true for Android.
- #3181 (closed: the iOS write-stall recovery Android lacks — the wrong template here, since Android has zero write failures), #3084 (closed, same *shape*: failure-reason mislabel hiding the real cause, on web/Capacitor), #2475 (closed: introduced the `isDisconnectionError` predicate behind the Android fallthrough), #3391, #3235.
- See also the BLE reconnect-churn issue in this batch for the recovery-UX side of board connection pain.

marcodejongh:
## Summary

Closes #3608.

Connecting to the board fails ~4× more often on Android than iOS (9.5% vs 2.3% over 30d, user-cancels stripped), and until now we couldn't say *why* — the reason we logged for Android was wrong.

On Android, `react-native-ble-plx` rejects a failed `connectToDevice()` with a raw `BleError` (e.g. `errorCode: 201` DeviceDisconnected, message `"Device … was disconnected"`). That error reached `classifyBleFailure()` untouched, but the classifier only matches Web-Bluetooth / CoreBluetooth **message strings**, so ble-plx messages fell through to `unknown` and were then re-bucketed as the write-path reason `disconnected` — the dominant, wrong Android bucket (162 events / 79 users). iOS records the exact same failure mode correctly as `connect_failed`.

**This PR (step 1 of the issue):**

- **`classifyBleFailure` now consumes the numeric `BleError.errorCode`.** A `BleErrorCode → BleFailureCategory` map (e.g. `201`/`200`/`205`/`3` → `connect_failed`, `204` → `board_not_found`, `300`/`302`/`400`/`404` → `service_missing`, `100`–`105`/`601` → `unavailable`) is consulted before the message ladder. The codes are mirrored as literals because `@boardsesh/ble-protocol` is platform-agnostic and can't depend on react-native-ble-plx. Gated on `name === 'BleError'` so a foreign numeric `errorCode` (e.g. an iOS CBError) can't be misread against the enum. Unmapped/`UnknownError` codes and non-ble-plx errors fall straight through to the existing message ladder — **web and native-iOS classification is unchanged.**
- **Raw codes are emitted on the `Bluetooth Connection Failed` event** (`bleErrorCode` + the low-level `androidErrorCode` GATT status) via a new `blePlxErrorCodes()` helper — so the new bucketing can be re-measured against real GATT statuses.

**Deliberately out of scope:** step 2 (connect-side recovery for the ble-plx path) is measurement-gated — decide after the new buckets land. Per the issue, Android has **zero** write failures, so porting the iOS write-stall recovery (#3181) would fix nothing. Web `failureReason` emission (#3088) stays separate.

This is pure JS/TS (no native code, no new deps), so it ships to the Android fleet via **OTA** and re-measurement can start as soon as it rolls out.

## Release Notes

- When a board won't connect on Android, you now get the real reason — Bluetooth off, board not found, or a connection that dropped — instead of a blanket "unknown error".

## Test plan

- `vp test run --project ble-protocol` — 255 passed, incl. new cases: **201 DeviceDisconnected → `connect_failed`** (the headline regression-lock), the full code→category table, the `name !== 'BleError'` scoping guard, `UnknownError` fallthrough, and `blePlxErrorCodes` extraction.
- `vp run test:mobile` — 3831 passed (the `use-board-bluetooth` suite still green).
- `vp run typecheck:ble-protocol`, `vp run typecheck:mobile` — clean.
- `vp check` on the 3 changed files — no lint/format/type errors.
- No BLE hardware in CI/sim, so the classifier (pure) is covered by unit tests; the real re-measurement happens in PostHog once the OTA reaches the Android fleet (issue step 2 input).

https://claude.ai/code/session_019ty8XosMsVnMRGRtCus7d1

### DIFF
--- a/packages/mobile/src/lib/ble/use-board-bluetooth.ts
+++ b/packages/mobile/src/lib/ble/use-board-bluetooth.ts
@@ -12,6 +12,7 @@ import {
 import { getMoonboardBluetoothPacket, isMoonboardDeviceName } from '@boardsesh/ble-protocol/moonboard';
 import { getMoonBoardGeometryByLayoutId } from '@boardsesh/board-config';
 import {
+  blePlxErrorCodes,
   classifyBleFailure,
   classifyBleFailureReason,
   isDisconnectionError,
@@ -974,6 +975,10 @@ export function useBoardBluetooth({
           layoutId,
           sizeId,
           failureReason: failureCategory === 'unknown' ? classifyBleFailureReason(error) : failureCategory,
+          // Raw ble-plx codes (Android) so the real connect-failure cause and the
+          // low-level GATT status are visible for re-measurement (#3608). Empty on
+          // web / native iOS.
+          ...blePlxErrorCodes(error),
         });
       } finally {
         connectInFlightRef.current = false;
--- a/packages/shared/ble-protocol/src/__tests__/connection-error.test.ts
+++ b/packages/shared/ble-protocol/src/__tests__/connection-error.test.ts
@@ -1,5 +1,6 @@
 import { describe, it, expect } from 'vitest';
 import {
+  blePlxErrorCodes,
   classifyBleFailure,
   classifyBleFailureReason,
   isBleWriteRecoveryFailedError,
@@ -17,6 +18,27 @@ const WRITE_RECOVERY_FAILED_MESSAGE =
 function bleError(message: string): Error {
   const error = new Error(message);
   error.name = 'BleError';
+  return error;
+}
+
+// Mimics a react-native-ble-plx BleError that also carries a numeric BleErrorCode
+// (and, optionally, the low-level platform codes) — the shape the connect path
+// receives from `connectToDevice()`. The classifier prefers the code over the
+// message on these.
+function bleErrorWithCode(
+  errorCode: number,
+  message: string,
+  extra?: { androidErrorCode?: number; iosErrorCode?: number },
+): Error {
+  const error = new Error(message) as Error & {
+    errorCode: number;
+    androidErrorCode?: number;
+    iosErrorCode?: number;
+  };
+  error.name = 'BleError';
+  error.errorCode = errorCode;
+  if (extra?.androidErrorCode !== undefined) error.androidErrorCode = extra.androidErrorCode;
+  if (extra?.iosErrorCode !== undefined) error.iosErrorCode = extra.iosErrorCode;
   return error;
 }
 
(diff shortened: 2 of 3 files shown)


marcodejongh:
Thanks for the review.

**1. Test coverage for all mapped codes** — fixed in 9eb49907d. Added the five codes that lacked an explicit case (103, 105 → `unavailable`; 301, 303, 405 → `service_missing`), so `BLE_PLX_CODE_TO_CATEGORY` is now fully locked and a future mapping typo can't slip through.

**2. OperationCancelled (2) → `unknown`** — intentional and left as-is. On this codebase our own connect-timeout path calls `cancelDeviceConnection`, which surfaces code 2, so mapping 2 → `user_cancelled` would risk silencing real connect-timeouts as "the user dismissed the picker". The genuine picker-dismiss is already stripped upstream by a different mechanism (the device-picker promise rejection), so cancels aren't in the failure numerator to begin with. As you note, whether a real Android in-connect user cancel matters is a step-2 re-measurement question — folded into the deferred "decide on recovery" work, not this PR.

PR #3656, opened 2026-07-16
feat(web): per-board app-install QR code on the kiosk

marcodejongh:
## Summary

Adds a per-board **install Boardsesh** QR to the gym kiosk (smart-TV wall dashboard). The kiosk only shows climbs from Boardsesh users, so each board now gets its own scannable code that deep-links to **that** board — climbers scan it, get the app, and their sends feed the screen.

Gated behind a per-kiosk toggle: when a gym flips **"Show app-install QR on each board"** on, every board on the kiosk renders its own QR. Off by default.

**Stacks on the kiosk epic** — base is `feat/gym-kiosk-editor`, not `main`.

### What's in it

- **`@boardsesh/kiosk`** — additive optional `showInstallQr: boolean` (default `false`) on `KioskLayoutSchema`, threaded through `emptyKioskLayout()` and the lenient reader (an absent or malformed value reads as `false`, no version bump). Tests cover strict parse with/without the field and the lenient default.
- **Backend** — `GymKioskBoard` now exposes `slug: String!` (the public `userBoards.slug`) so the QR has a target; populated in the kiosk read resolver, added to `GYM_KIOSK_FIELDS`, codegen regenerated. Resolver test asserts `slug` on both the write-then-resolve and the anon read path. Visibility gate unchanged (slug is public, same gate as the rest of `GymKioskBoard`).
- **Web render** — new `BoardInstallQr` (`qrcode.react`'s pure-SVG `QRCodeSVG`, SSR-safe) on a **white** rounded tile (a QR needs a light quiet zone; a bare dark QR on the dark kiosk surface won't scan) with a scan caption. Positioned in a corner of the board art region, never overlapping the identity strip. Rendered only when `showInstallQr` is on **and** the board has a slug. `showInstallQr` threads config → renderer → `BoardSlot`; `slug` threads off `GymKioskBoard`.
- **Manage editor** — a MUI `Switch` next to the leaderboard settings, wired into the editor's dirty state + `KioskLayoutSchema` save gate; the live preview reflects it. Also memoises `slotBoards` / `assignedBoards` / `hasUnknownSlots` (addresses the #3637 review perf nit — they were recomputed every render, defeating child memoisation).
- **i18n** — caption + toggle strings in `en-US` / `es` (plafón) / `fr` (croix / mur conventions).

### QR target

`${SITE_URL}/b/{slug}` (built from `SITE_URL`, not hardcoded). This opens the **web board page** everywhere (Android / desktop) and is iOS-universal-link eligible.

It does **not** yet open the mobile **app** straight to the board — that needs native work (a mobile `app/b/[slug]` Expo Router route + an Android `/b` `autoVerify` intent filter). Tracked as a follow-up: #3657. The web board page (which carries install CTAs) is the current fallback.

## Release Notes

- Your gym's wall screen can now show a scannable code on every board — climbers scan it to get Boardsesh, and their sends light up the screen.

## Test plan

- `vp run typecheck` — clean.
- `vp test run --project kiosk` — 36 pass (new `showInstallQr` strict + lenient cases).
- `vp test run --project web` for `board-install-qr`, `kiosk-editor-state`, `kiosk-view-model` — pass (new QR-target, toggle-serialisation, dirty-flag, and per-transition-preservation cases).
- `vp test run --project backend src/__tests__/gym-kiosks.test.ts` — 26 pass (new `slug` assertions).
- `vp run check:i18n` + `check:i18n:orphans` — OK; catalog completeness parity test passes.
- `vp check` — my files clean (3 pre-existing `no-floating-promises` errors in unrelated `packages/db` test files, unchanged vs base).
- Confirmed zero unintended `packages/db` diffs: only the `gym_kiosks` schema comment/cast; the SQL column default stays byte-identical (additive field, lenient reader covers absence), so no migration/snapshot churn.

### DIFF
--- a/packages/backend/src/__tests__/gym-kiosks.test.ts
+++ b/packages/backend/src/__tests__/gym-kiosks.test.ts
@@ -515,6 +515,8 @@ describe('updateGymKiosk layout validation', () => {
     );
     expect(updated.boards.map((b) => b.boardUuid)).toEqual([boardPub1.uuid, boardPub2.uuid]);
     expect(updated.boards.every((b) => b.boardId != null)).toBe(true);
+    // The install-QR needs each board's public slug. insertBoard seeds slug = uuid.
+    expect(updated.boards.map((b) => b.slug)).toEqual([boardPub1.uuid, boardPub2.uuid]);
   });
 
   it('persists the schema-parsed layout, stripping unknown keys', async () => {
@@ -643,6 +645,8 @@ describe('gymKiosk public read', () => {
     );
     // The private board is dropped from the resolved boards...
     expect(kiosk!.boards.map((b) => b.boardUuid)).toEqual([boardPub1.uuid]);
+    // ...and the surviving public board exposes its slug for the install QR.
+    expect(kiosk!.boards.map((b) => b.slug)).toEqual([boardPub1.uuid]);
     // ...but the layout JSON still records the slot (renderer degrades the preset).
     expect(kiosk!.layout.boards.map((slot) => slot.boardUuid)).toEqual([boardPub1.uuid, boardPriv.uuid]);
   });
--- a/packages/backend/src/graphql/resolvers/social/gym-kiosks.ts
+++ b/packages/backend/src/graphql/resolvers/social/gym-kiosks.ts
@@ -81,6 +81,7 @@ async function deriveUniqueKioskSlug(gymId: number, name: string): Promise<strin
 type ResolvedKioskBoard = {
   boardId: number;
   boardUuid: string;
+  slug: string;
   name: string;
   boardType: string;
   layoutId: number;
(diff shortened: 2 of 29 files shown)


marcodejongh:
Thanks for the review — addressed in 6e14a2fca:

1. **`slug` type** — tightened `BoardSlot.slug` to `string` to match the non-nullable `GymKioskBoard.slug` (`String!`). The render guard is now a truthy check (`showInstallQr && slug`) — belt-and-braces against an empty slug, not a null one.

3. **BoardSlot test** — added `board-slot.test.tsx` covering the conditional directly: QR shown when the toggle is on and a slug is present, hidden when the toggle is off or the slug is empty.

2. **`as unknown as KioskLayout` DB default** — keeping this deliberately. Including `showInstallQr` in the literal would change the `gym_kiosks.layout` column default, forcing a migration (and this repo's `drizzle-kit generate` drags in a known destructive `boardsesh_ticks` CHECK-drop that has to be hand-stripped). Since the field is additive and the lenient reader defaults an absent value to `false`, the SQL default can stay byte-identical with zero migration/snapshot churn. The comment flags the tradeoff; a future required `KioskLayout` field would still be caught by the strict schema + the resolver's `emptyKioskLayout()` insert path. Open to revisiting if you'd rather eat the migration.

PR #3663, opened 2026-07-16
feat(mobile): confirm before sign-out and clear the offline catalogs

### LINKED ISSUE #3621: Offline mode: delete data on logout
When the user wants to logout, we should ask for confirmation and warn them in the confirmation that we'll delete the local saved boards etc upon logout

marcodejongh:
## Summary

Signing out on mobile threw away local data with no warning and left other data behind that the user thinks is gone. The **user drawer** — the row most people tap — called `signOut('manual')` instantly, so a tap wiped the local logbook and dropped any writes still queued with no confirmation. The **More tab** only warned when the queue was non-empty. And the downloaded board catalogs (200k+ rows each) stayed on the device forever, so "log out" never actually freed the storage a downloaded board was using.

Both entry points now route through a shared `useConfirmSignOut` hook that **always** asks and composes an honest message:

- the downloaded-boards + offline-logbook sentence only when the user actually has downloaded boards (`syncEnabledBoards` non-empty),
- the count of unsynced changes when the queue isn't empty,
- otherwise just "you'll need to sign in again".

It keys off the enabled-boards list rather than the `offline-board-downloads` flag, because the wipe isn't flag-gated either — a flag-rolled-back user can still be holding downloads and still loses them. The confirm lives in the hook, not in `signOut`, so the forced 401/expiry paths and account deletion (already behind its typed-`DELETE` gate) can't inherit a dialog they have no moment to show.

`clearLocalData` (was `clearUserData`) now clears the board catalogs alongside the user tables and wipes `sync_meta` whole, then `VACUUM`s so the freed pages actually leave the file. Rows and the markers describing them must die together: `scope-complete:` and the bootstrap markers sit *outside* the `checkpoint:` prefix, so a prefix sweep would strand them and leave a board advertising an empty catalog to local-first search. Board rows used to be kept as a shared cache; that traded a signed-out user's disk for a faster re-enable, and stopped paying once snapshot bootstrap turned a re-download into one CDN `GET` per board.

> **Stacked on #3630** (`fix/3617-offline-manage-storage`) — it reuses that branch's `vacuumDatabase` and `scope-teardown` primitives. Rebase onto `main` once #3630 merges. Review the [compare against that base](https://github.com/boardsesh/boardsesh/compare/fix/3617-offline-manage-storage...fix/3621-signout-wipe) to see only this PR's diff.

Closes #3621

## Release Notes

- Logging out now asks first, and tells you what it clears — your downloaded boards, your offline logbook, and any sends that haven't synced yet.

## Test plan

- `vp test run --project offline-sync` — 225 passed (rewrote the checkpoints suite for the whole-`sync_meta` wipe)
- `vp test run --project i18n` — 28 passed (catalog parity across en-US/es/fr)
- `vp test run --project mobile` (touched files: `use-confirm-sign-out`, `user-drawer-provider`, `connection.test`, `auth-provider`) — 42 passed. New `connection.test` case asserts the "downloaded but empty" trap: seed a catalog + `scope-complete:` marker, wipe, then `isScopeDownloadComplete` is false.
- `vp check` — 0 errors; `vp run check:i18n:orphans` — OK (no `mobile.more.signOut` remnants)
- `vp run check:mobile-bundle` — OK (14.0 MB)
- Full `vp run test:mobile`: 9 pre-existing timeout flakes in unrelated UI-render tests (`filter-button`, `about`, `acknowledgements`, …), reproduced identically on the base branch under the same machine load.
- **Still owed:** on-device pass in the simulator — drawer + More both raise the dialog, offline sentence appears only with a board downloaded, unsynced count shows after an airplane-mode tick, and the `.db` file shrinks after confirming.

https://claude.ai/code/session_01Rwf5tEw3RBQEf69Eevy8sH


marcodejongh:
Thanks — addressed in `0221f3a`.

**Bug (hasDownloads signal):** Agreed, and fixed — but I went one step past the suggestion. `getDownloadedScopeKeys()` reads `scope-complete:` markers, which a *partial* mid-download hasn't written yet even though it has rows the wipe will delete. The signal that can't be wrong is "does `board_climbs` hold any row" — that's literally what `clearLocalData` removes. New `hasDownloadedBoardData(db)` (one O(1) `EXISTS`) drives the sentence now, with a test for the flag-rollback case (toggle list empty, rows present → still warns).

**inFlightRef:** Fixed — reset in a `finally`, so a sign-out that resolves without navigating away leaves the button usable.

**PRAGMA busy_timeout:** This one isn't a no-op — my comment just described it wrong (now corrected). `withExclusiveTransactionAsync` issues a bare `BEGIN`, which is *deferred* (the "exclusive" is a dedicated connection, not `BEGIN EXCLUSIVE` — see `node_modules/expo-sqlite/build/SQLiteDatabase.js:162`). So no lock is held when the callback starts; the write lock is taken on the first `DELETE`, and `busy_timeout` set before it governs that acquisition's wait. It's the same guard the reviewed `scope-teardown` and snapshot-bootstrap paths use.

marcodejongh:
Follow-up in `56f1ef2`:

- **Doc inaccuracy** — fixed. The paragraph now says the sentence is gated on `hasDownloadedBoardData` (the `board_climbs` probe), matching the code.
- **Dead wrapper** — removed; the button row calls `void confirmSignOut()` directly.
- **French agreement** — this one's correct as written. `docs/i18n-french-glossary.md` fixes the board device as **feminine** (« la board », plural « les boards »), with the explicit note *"les boards téléchargées"* patterns like « la planche ». So «que tu as téléchargées» (fem. pl.) agrees. This is also the convention #3630 uses in its storage strings. Left unchanged.

PR #3668, opened 2026-07-16
feat(web): My Gyms drawer + user-drawer entry

marcodejongh:
## Summary

There was no way to click your way to anything gym-related in the app — a gym owner couldn't reach their own gym console. This adds a **My Gyms** entry to the user drawer for every signed-in user, right below **My Boards**, cloning that established pattern.

Tapping it opens a nested drawer that lists the gyms you run:

- Each row shows the gym name and a role chip (Owner / Admin / Editor / Member).
- **Manage** (only when you can edit) jumps to `/gym/[slug]/manage`; **View** opens the public `/gym/[slug]`. Both are real `LocaleLink`s.
- Empty state nudges you that your gym might already be on Boardsesh, with a **Find your gym** action that opens the existing global search pre-set to the Gyms category (no new search UI).

Consumes the existing `myGyms` GraphQL query (previously used only by the board-linking selector) via a new `useMyGyms` hook that mirrors `useMyBoards`. Owner is resolved by comparing `ownerId` to the signed-in user, since the backend reports owners as gym admins in `myRole`.

## Release Notes

- Your gyms are one tap away — open the menu and hit **My gyms** to jump straight into managing kiosks, boards, and your crew.

## Test plan

- `vp check` — 0 errors (pre-existing warnings only)
- `vp run typecheck:web` — clean
- `vp test run --project web --reporter=agent` — all green (new `my-gyms-drawer` suite: 12 tests; `user-drawer` suite still passing)
- `vp run check:i18n` — no hardcoded strings
- `@boardsesh/i18n` catalog completeness — parity across en-US / es / fr

Manual flows to exercise:
- Signed-in user with 0 gyms: empty state + "Find your gym" opens gym search.
- Signed-in user who owns a gym: Owner chip, Manage → `/gym/[slug]/manage`, View → `/gym/[slug]`.
- Non-editor: Manage hidden, only View shown.

### DIFF
--- a/packages/shared/i18n/locales/en-US/common.json
+++ b/packages/shared/i18n/locales/en-US/common.json
@@ -390,6 +390,19 @@
     "ariaCreateBoard": "Create a board",
     "empty": "No boards yet. Create one from the board selector to get started."
   },
+  "myGyms": {
+    "title": "My Gyms",
+    "empty": "Your gym might already be on Boardsesh. Find it to manage your boards and crew.",
+    "findGym": "Find your gym",
+    "ariaFindGym": "Find your gym",
+    "roleOwner": "Owner",
+    "roleAdmin": "Admin",
+    "roleEditor": "Editor",
+    "roleMember": "Member",
+    "manage": "Manage",
+    "view": "View",
+    "loadError": "We couldn't load your gyms. Give it another go."
+  },
   "userDrawer": {
     "signIn": "Sign in",
     "signInModalTitle": "Sign in to Boardsesh",
--- a/packages/shared/i18n/locales/es/common.json
+++ b/packages/shared/i18n/locales/es/common.json
@@ -63,6 +63,19 @@
     "ariaCreateBoard": "Crear plafón",
     "empty": "Aún no tienes plafones. Crea uno desde el selector de plafones."
   },
+  "myGyms": {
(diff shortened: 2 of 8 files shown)


marcodejongh:
Thanks for the review. Dispositions:

- **Dead `eslint-disable` in `use-my-gyms.ts`** — fixed in 104af1c. The effect's deps already cover every reactive value, so the suppression was silencing nothing.
- **`loadMore` swallows errors** — kept as-is: this mirrors `use-my-boards.ts` verbatim (the hook this PR clones). The initial fetch surfaces `error`; a pagination failure on an already-populated list logs and lets the next scroll retry. Diverging here would make the two sibling hooks inconsistent.
- **`sx={{ color: 'var(--neutral-300)' }}` and the CSS module pixel values** — these are copied 1:1 from `my-boards-drawer.tsx` / `my-boards-drawer.module.css`, which are already in `main`. `var(--neutral-*)` is the token surfaced as a CSS custom property, so it is token-based, not a hardcoded hex. Kept for parity with the pattern being cloned.
- **"Multi-line docstrings are banned by CLAUDE.md"** — I can't find that rule in CLAUDE.md, and the sibling files (`use-my-boards.ts`, `my-boards-drawer.tsx`) both carry multi-line JSDoc. Left as-is.

PR #3675, opened 2026-07-16
feat: gym post-claim activation — approval notification + console welcome checklist

marcodejongh:
## Summary

There was no way to click your way to anything gym-related in the app — a gym owner couldn't reach their own gym console. This adds a **My Gyms** entry to the user drawer for every signed-in user, right below **My Boards**, cloning that established pattern.

Tapping it opens a nested drawer that lists the gyms you run:

- Each row shows the gym name and a role chip (Owner / Admin / Editor / Member).
- **Manage** (only when you can edit) jumps to `/gym/[slug]/manage`; **View** opens the public `/gym/[slug]`. Both are real `LocaleLink`s.
- Empty state nudges you that your gym might already be on Boardsesh, with a **Find your gym** action that opens the existing global search pre-set to the Gyms category (no new search UI).

Consumes the existing `myGyms` GraphQL query (previously used only by the board-linking selector) via a new `useMyGyms` hook that mirrors `useMyBoards`. Owner is resolved by comparing `ownerId` to the signed-in user, since the backend reports owners as gym admins in `myRole`.

## Release Notes

- Your gyms are one tap away — open the menu and hit **My gyms** to jump straight into managing kiosks, boards, and your crew.

## Test plan

- `vp check` — 0 errors (pre-existing warnings only)
- `vp run typecheck:web` — clean
- `vp test run --project web --reporter=agent` — all green (new `my-gyms-drawer` suite: 12 tests; `user-drawer` suite still passing)
- `vp run check:i18n` — no hardcoded strings
- `@boardsesh/i18n` catalog completeness — parity across en-US / es / fr

Manual flows to exercise:
- Signed-in user with 0 gyms: empty state + "Find your gym" opens gym search.
- Signed-in user who owns a gym: Owner chip, Manage → `/gym/[slug]/manage`, View → `/gym/[slug]`.
- Non-editor: Manage hidden, only View shown.


marcodejongh:

- **Success-page `gym` param** — added a `^[a-z0-9-]{1,120}$` guard so a crafted value can't build a `../`-style href; the CTA only renders for a slug/UUID-shaped ref.
- **Members `limit: 50`** — added a comment clarifying it's correct: there's a single owner, so any page of members holds at most one owner row, and the first 50 always include a non-owner whenever one exists. It's an existence check, not a full count, so no false "invite your crew" undone state.

Left as-is, with rationale:

- **`var(--neutral-100/200/400)`** — these are theme-aware design tokens defined for both light and dark in `packages/web/app/components/index.css` (lines 11 and 151), and are the established pattern across shipped components (notification-item, vote-button, comment-form, mui-theme). They resolve correctly in dark mode — not hardcoded hex.
- **Live-toast strings in `use-notification-subscription.ts`** — that whole `formatNotificationMessage` function is hardcoded English by existing convention, and it's a `.ts` file (the i18n linter scopes to `.tsx` under `app/`). The new case matches the file's pattern; `check:i18n` passes.

PR #3689, opened 2026-07-16
feat(db): add pre-registered Climb2Vec morphology trial

marcodejongh:
## Summary

- Add deterministic 12-dimensional hold morphology extraction from the official board artwork, with source/artifact hashes and explicit unsupported-placement gaps.
- Add the frozen Stage 2 target, physical-problem leakage guard, combined Kilter/Tension training export, and board-safe MoonBoard scoring path.
- Add the pre-registered Runs 3–7 experiment: incumbent GBM, same-learner morphology challenger, compact relation-aware residual encoder, sealed Tension benchmark evaluation, and the documented kill/ship thresholds.
- Add read-only shadow evaluation plus atomic, exact-coverage content-prior and similarity loaders. The existing `climb2vec-v1` weekly production workflow remains explicit and compatible.

The candidate still enters the grade posterior only as a capped `DeherdedGradeSignal` under the no-shock clamp. This PR does not change `grade-model/`, does not change `refresh-climb-grades.ts`, performs no production writes or coefficient refits, and does not spend the full training-run budget.

## Release Notes

## Test plan

- `vp run typecheck:db`
- `/tmp/climb2vec-venv/bin/python -m unittest discover -s ml/climb2vec -p 'test*.py'` — 40 passed
- Targeted `vp check` over every changed TypeScript/documentation file — clean
- Focused DB/content-model/morphology/loader suites — all new tests passed
- Full DB suite — 99 passed, 1 skipped, 1 pre-existing unrelated system-catalog exemption failure
- Production read-only Tension extraction — 18,011 rows; 2,846 sealed benchmark climb+angles; all 88 benchmarks below 20 ascents retained; 2 conflicted physical problems rejected atomically
- Deterministic artwork audit — 5,763/6,286 placements covered (91.68%); all 523 gaps are explicit Kilter tombstones
- Combined Kilter/Tension/MoonBoard score-artifact smoke test — exact catalog rows, shared snapshot, finite supported priors/embeddings, explicit null tombstones for unsupported climbs


marcodejongh:
## Fable review — findings and fixes

Reviewed this branch fresh. The work is still relevant: nothing has touched `ml/climb2vec` or the content-model path since the branch was cut, `docs/boardsesh-grade.md` still lists this exact trial as the next roadmap step, and the branch rebases onto `main` cleanly. Rebased onto `main` (801 commits, no conflicts) and fixed one real regression plus three robustness gaps.

### Major — zero-hold rows were leaking into the live climb2vec-v1 weekly artifacts

The old `main()` skipped rows whose holds matched no placement (`if (!holds || holds.length === 0) continue` and `if (row.holds.length === 0) continue`). The rewritten `main()` pushed every stat row unconditionally, and `refresh-content-model.yml` runs this same script in crowd + score mode for `climb2vec-v1`. So placement-mismatch climbs entered Phase-1 training and the score artifact as all-zero feature rows with real labels. `DeepSetsGrader` returns a finite garbage prediction for an empty mask (count clamped to 1, max-pool `nan_to_num`'d), and the legacy upsert then persists that fabricated `content_prior` for cells that previously had no embedding row — the cold-tail cells `refresh-climb-grades` reads `content_prior` from.

Zero-hold rows are only meaningful to Stage 3, which converts them into explicit `supported: false` tombstones so the identified artifact covers the catalog exactly. `extract-training-matrix.ts` now keeps them only when the extract asks for Stage-3 output — `--morphology`, `--target=stage2`, or the new explicit `--keep-unsupported` — and restores the original skip everywhere else. Every Stage-3 command in `ml/climb2vec/README.md` already passes `--morphology`, so the trial is unaffected; the weekly workflow passes neither and is back to its `main` behaviour. The extract log now reports `dropped-unsupported=<n>` or `unsupported-rows=kept-as-tombstones` so the mode is visible in the workflow output.

### Minor

- **Cosine scores are now clamped to `[-1, 1]` before rounding** (`similarity_export.py`). `load-similarity-records.ts` rejects any out-of-range neighbour score by failing the whole atomic board rebuild, and a float32 dot product of renormalized unit vectors does land past 1.0 for a near-duplicate pair (measured 1.0000006 at 64 dims — rounding saves it today, but the margin shrinks as the embedding widens). Extracted as `neighbor_score()` with a test.
- **`loadMorphology` no longer trusts `JSON.parse ... as MorphologyRecord`.** The reader moved to `extract-training-matrix-records.ts` (same CLI/`*-records.ts` split as `load-content-model.ts`) and validates through `createRecordValidator`, so a truncated artifact line reports `morphology artifact line 412: vector must contain exactly 12 numbers` instead of "Cannot read properties of undefined". 9 new tests.
- **`extract-hold-morphology.ts` uses `main().catch(...)`** with a script-prefixed error and `process.exit(1)`, matching every sibling script, instead of `void main()`.

Also documented the zero-hold rule in `docs/climb2vec.md` and `ml/climb2vec/README.md` so nobody re-adds the leak by dropping `--morphology` from an extract.

### Validation

- `vp run typecheck:db` — clean; `packages/db/scripts/**` typechecked separately (it is outside the package `tsconfig` `include`) — clean
- `vp check` — 0 errors on the touched packages
- `vp run test:db` — 420 pass. 3 failures in `src/queries/climbs/__tests__/create-climb-filters.test.ts` are pre-existing on `main` (MoonBoard zone-calibration float formatting; the PR touches no file that test depends on)
- `ml/climb2vec` full Python suite — 50 tests, all pass (numpy/scipy/scikit-learn/torch 2.13.0+cpu)

### Not changed

The claude-review bot's headline claims — non-existent PyPI versions, broken workflow references, a SHA-pin convention — do not hold up: the pinned versions all exist, the workflow references resolve, and this repo does not SHA-pin first-party actions. The legitimate earlier review points were already addressed in later revisions on this branch.

marcodejongh:
### CI: one red, and it isn't this branch

Everything passes except `test-default`, which fails on a single assertion:

```
FAIL  i18n  src/__tests__/catalog-completeness.test.ts
  > i18n catalog completeness > de > boards.json ships the same key set as en-US
  missing: boardForm.create.duplicateTitle, duplicateBody, addAnyway, keepEditing
```

Those four keys landed in `en-US/boards.json` via `c1e633cf7` ("stop discarding a new board when you already own that config") without German counterparts, so `main` itself is red — the last two `main` CI runs failed the same way, and every open PR inherits it. This branch touches no i18n file.

The fix is already open and green as **#4213** ("restore German catalog parity in boards.json"), which adds exactly those four keys. Not duplicating it here — a second copy of the same German strings would just conflict when #4213 lands.

Leaving this PR in draft until #4213 merges. After that, a rerun (or a rebase) turns `test-default` green and this is ready for review: no unresolved review threads, no conflicts, everything else passing including `lint`, `typecheck`, `claude-review`, `mobile-bundle`, and the backend suite.

PR #3678, opened 2026-07-16
feat(web): apply the Velvet Send design system (supersedes #3166)

marcodejongh:
## Summary

Resurrects #3166 — the full **Velvet Send** web migration (violet `#6D28D9` brand, fg/fill split, violet-tinted surfaces, geometry/motion, grade donut, `@boardsesh/velvet-tokens` shared package) — rebased onto current `main` by its original implementing agent, plus the gap work that accumulated while it was parked:

- **Kiosk-era re-sweep**: the gym-kiosk epic landed after the original branch point and still carried the retired dark palette — kiosk theme scope/preview move to the Velvet dark ladder, `brand-contrast` constants + fixtures follow, gym/admin manage pages read the scheme-aware `--color-primary`.
- **Design-panel fixes** (from a six-lens design review of the web translation):
  - ButtonBase gets an explicit 2px `:focus-visible` outline — keyboard focus was invisible on every button/tab/chip (WCAG 2.4.7).
  - The shared `live` brand role is emitted as `--color-live`; the board-presence lit-climb affordance rides it instead of `--color-warning`, matching mobile's amber in both schemes.
  - `accent.dark` pinned to `#F97316` (AA with dark text); SvgIcon accent override removed (accent is fill-only).
  - Button hover is a colour tint; the `translateY` lift is gone and Card shadow hover is gated behind `(hover: hover) and (pointer: fine)` so nothing sticks raised after a tap.
  - Parity test gains `--color-live` + `--color-error-muted-hover` rows and a snapshot guard pinning the deliberately divergent web surface ladder.
- **Dev-infra**: the dev orchestrator's backend health window goes 5s → 60s (a cold tsx boot on a loaded box outlives 5s, orphaning healthy backends).

Supersedes #3166 (closed; branch was force-pushed, so GitHub can't reopen it — original review context lives there). White dark-mode inputs are intentionally untouched here; the dark-input redesign and the form kit land in stacked follow-up PRs.

## Release Notes

- Boardsesh on the web gets a fresh look — the same violet **Velvet Send** style as the app, in light and dark.
- Keyboard climbers get visible focus rings everywhere.
- Your profile card's grade spread shows as a colourful donut instead of a flat bar.

## Test plan

- `vp run typecheck` — green across packages.
- `vp test run --project web` — 5,259 tests green (incl. parity + WCAG guard suites); `vp run test:mobile` — 3,833 green; `vp run check:mobile-bundle` — green (velvet-tokens extraction is behaviour-neutral for native).
[...]

PR #3696, opened 2026-07-16
fix(web): finish the scheme-var migration on flash surfaces + chart legend

marcodejongh:
## Summary

Follow-up to the #3678 review (items landed after the merge):

- **Flash tick states** read `themeTokens.colors.amber` directly while sibling branches were already on CSS vars. Amber now ships as `--color-amber` (scheme-invariant data yellow, added to the parity table), and the flash badge text rides `--color-on-accent` instead of a static neutral — so a future amber retune propagates and the dark-on-amber text stays scheme-stable.
- **Session grade chart**: the attempt bar was the one legend colour still token-derived; it's now a hardcoded scheme-neutral hex like its flash/send siblings (chart data can't resolve `var()`), with the rationale comment extended.

Note: the review also cited `inline-list-tick-bar.tsx`, but on merged main that file has no direct token reads — the flagged patterns all lived in `tick-button.tsx`.

## Release Notes

none

## Test plan

- `vp check` clean on all touched files; `vp run typecheck:web` green.
- Theme parity suite (35 tests) green incl. the new `--color-amber` row; logbook + charts component tests 155/155 green.

### DIFF
--- a/packages/web/app/components/charts/session-grade-bars.ts
+++ b/packages/web/app/components/charts/session-grade-bars.ts
@@ -5,8 +5,9 @@ import { formatVGrade } from '@/app/lib/grade-colors';
 // Legend colours (60% opacity). Scheme-neutral mid-tones so the bars stay
 // legible on both the light and the dark card surface — the brand success/error tones
 // are tuned per scheme and would wash out in the other one (these feed chart data, not
-// CSS, so a scheme-aware var() can't be used here). Attempt matches the light-scheme
-// neutral-300 lavender for the same reason.
+// CSS, so a scheme-aware var() can't be used here). All three are deliberately FROZEN
+// snapshots, not token-derived: a future palette retune must not silently recolour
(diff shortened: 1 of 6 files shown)


marcodejongh:
All three round-two items addressed in e9ed5b73a:

1. **play-view-tick-bar flash button** migrated to `var(--color-amber)` / `var(--color-on-accent)` (and its now-unused `themeTokens` import dropped). The other ~20 `themeTokens.colors.amber` reads in the app are non-flash data-yellow accents (stars, badges, playlist palettes) that the form-migration waves sweep file-by-file — deliberately out of scope for this follow-up.
2. **Chart legend comment** now states the decoupling is intentional: all three legend colours are frozen snapshots so a future palette retune can't silently recolour historical chart data.
3. **Regression test** added: asserts the flash variant renders `var(--color-amber)` background and `var(--color-on-accent)` text.

PR #3699, opened 2026-07-16
feat: gym duplicate merge tooling + admin review queue

marcodejongh:
> Builds on M1 (#3669, now merged into `main`): `merged_into_gym_id` + canonical resolution. This PR targets `main` directly; its migration `0183_bouncy_azazel` chains from M1's `0182_magical_thor`.

## What this does

Prod has ~46–60 physical duplicate gym rows (same normalized name within ~100 m, most 40–90 m apart from cross-provider coordinate drift) plus ~1,485 alias-less SYSTEM-owned gyms. The call: humans merge duplicates through an easy admin page — no bulk auto-apply.

**Merge library** (`packages/db/src/queries/gyms/merge-gyms.ts`): the cluster-merge internals moved out of the dedupe CLI so the backend can import them (the script is now a thin wrapper). Re-points boards, source aliases, follows, members, comments, feed items, notifications, votes onto the survivor — plus the two verified gaps:

- **gym_claims** — honours the unique-pending index. A claimant mid-flight on a twin lands on the canonical; if they already have a pending claim there, that one stands and the duplicate's is dropped (ON CONFLICT DO NOTHING semantics). Nobody loses their place.
- **gym_kiosks** — a live kiosk slug that collides on the canonical is suffixed (`main-wall` → `main-wall-2`) and returned as a **warning**. A re-slugged kiosk's printed install QR no longer resolves, so the merge surfaces it for reprinting.

**Audit table** (`gym_merge_audit`, migration `0183_bouncy_azazel`): one row per admin decision — `merged` (canonical + duplicate ids, moved counts, warnings) or `dismissed` (keyed by a stable hash of the sorted member gym ids so the queue hides that exact cluster).

**Backend admin GraphQL** (admin-gated via `requireAdmin`, mirroring the gym-claims review flow):

- `duplicateGymClusters` — tiered candidate clusters over LIVE gyms. Tier A: same normalized name within 20 m; tier B: within 150 m (the observed drift band). Each member marked with ownerType (system/user), claimStatus (none/pending/approved), provider origins, FK counts (boards/follows/members/kiosks/claims), coords + distance-to-canonical. Dismissed clusters excluded. Paginated.
- `orphanGyms` — alias-less SYSTEM-owned live gyms with no location-sync source. List-only, count summary + page (no bulk action).
- `mergeGyms` — validates every uuid is live + un-merged and that the canonical is not among the duplicates; runs the library merge per duplicate in one transaction; writes audit rows; returns per-duplicate moved counts + kiosk-slug warnings.
- `dismissGymCluster` — writes the dismissal audit row.

**Web `/admin/gym-duplicates`** (cloned from `/admin/gym-claims`, same `checkAdmin` gate): cluster cards with member rows (owner/claim badges, provider chips, FK counts, distance); canonical pre-selected by rule — a claimed or user-owned row always wins over SYSTEM regardless of completeness, ties fall back to completeness-then-oldest — with an admin radio override; a merge confirm dialog listing exactly what moves (board/follow/member/kiosk/claim counts) plus any kiosk-slug warning after; a dismiss button; and a second tab for the orphan audit. Linked from the `/admin` index. i18n across en-US/es/fr.

**CLI `--backfill-pointers`** (M1's deferred follow-up): re-clusters pre-pointer SOFT-DELETED twins (the ~46–60 pre-M1 merged rows) against live gyms by name + distance (same tiers) and sets their missing `merged_into_gym_id`, so old uuids/slugs (printed kiosk QRs) resolve to the survivor. Report-only by default; `--apply` writes, under human review.

- `vp run db:dedupe-gyms --backfill-pointers` — dry-run report
- `vp run db:dedupe-gyms --backfill-pointers --apply` — set the pointers

## Kiosk-slug warning example

Merging a twin whose kiosk slug `main-wall` collides with the canonical's `main-wall` returns:

```json
{ "type": "kiosk_slug_changed", "kioskUuid": "…", "kioskName": "Main Wall",
  "previousSlug": "main-wall", "newSlug": "main-wall-2" }
```

The UI shows this in a persistent warning banner so the gym reprints that wall's install QR.

## Tests

- **db** (`dedupe-gym-locations.integration.test.ts`): moved-counts accuracy, `merged_into` set on the twin, claims unique-pending conflict path, kiosk slug-suffix warning; `--backfill-pointers` arg parsing.
- **backend** (`gym-duplicates.test.ts`): admin gate denies non-admin + anonymous on every resolver, tier A vs B classification, user-owned/claimed canonical preference, dismissal filtering, orphan listing, and mergeGyms end-to-end (boards/members/kiosks/claims re-point, merged_into pointer, audit rows).

## Notes

- The merge SQL now uses bare `'gym'` literals instead of `'gym'::social_entity_type` casts — the untyped literal coerces to the enum in prod and string-compares against the backend test schema's text column, so the same code is portable across both.
- `gym_member_role` was added to the backend test schema (`schema-sql.ts`) because the member-collapse upsert builds an enum-typed value there.

## Review fixes (Fable pass)

- **Member role precedence** — the member-collapse upsert now preserves `admin > editor > member` on both the SELECT and the ON CONFLICT sides, so a twin's editor arrives as editor and the canonical's own editor is never demoted to member. Covered by a new editor-role test (db + backend).
- **`is_public` no longer forced** — merging into a user-owned private gym keeps the survivor's own `is_public`; the metadata backfill only fills empty fields.
- **SYSTEM-canonical override guard** — `mergeGyms` rejects an inverted pre-selection (a SYSTEM row kept as survivor over a user-owned or claim-approved duplicate) unless the input carries `allowSystemCanonicalOverride: true`. The panel detects the inversion, shows a warning in the confirm dialog, and only then sends the flag.
- **Audit reversibility** — (a) every merge audit row now stores per-table moved-row ids (`moved_rows` jsonb: board/source/follow/member/claim/kiosk/comment/feed/notification/vote ids + soft-deleted gym id), not just aggregate counts, so a bad merge can be traced and manually reversed. (b) **Chosen approach for the collided pending claims: re-point to the canonical and flip to `expired`** (an existing `gym_claim_status` value — no new enum, no migration), instead of deleting them. The claimant's full claim content is preserved on the survivor and their live canonical claim still stands.
- **Backfill ambiguity guard** — `--backfill-pointers` now skips a twin when a second live candidate sits within the match band (reported, never pointed — a wrong pointer redirects printed QRs to the wrong gym), and restricts to SYSTEM-owned twins by default (`--include-non-system-twins` to opt in).
- **Orphan copy fixed** — `orphans.summary` now says "system-owned gyms with no location-sync source" (what the query returns) in all three locales, not the old "no boards, members, or followers".

**Discretionary items deferred (noted, not blocking):** metadata backfill for website/logo/brand colours (finding 7), a lock-ordering comment, audit-count-noise cleanups (live-only board counts, self-upsert votes), i18next plural forms in the panel counts, RadioGroup aria-labels, and double-merge-idempotency / signature-resurface tests.

## Release Notes

One gym, one page — we're merging duplicate gym listings, and your follows, boards, and kiosks ride along to the surviving page. If a merge shifts a kiosk's address, we flag it so the gym can reprint that wall's install QR.

https://claude.ai/code/session_019VGxf4ZJGFRanuJGrRMDgY

### DIFF
--- a/packages/backend/src/__tests__/gym-duplicates.test.ts
+++ b/packages/backend/src/__tests__/gym-duplicates.test.ts
@@ -1,0 +1,373 @@
+import { describe, it, expect, beforeEach } from 'vite-plus/test';
+import { v4 as uuidv4 } from 'uuid';
+import { sql } from 'drizzle-orm';
+import type { ConnectionContext } from '@boardsesh/shared-schema';
+import { db } from '../db/client';
+import { socialGymDuplicateQueries, socialGymDuplicateMutations } from '../graphql/resolvers/social/gym-duplicates';
+
+/**
+ * Real-DB coverage for the /admin/gym-duplicates review queue: the admin gate,
+ * the tiered candidate clustering, dismissal filtering, the orphan audit, and the
+ * mergeGyms mutation (claims unique-pending conflict, kiosk slug-suffix warning,
+ * moved counts, merged_into pointer, audit rows).
+ *
+ * Mirrors gym-write-access-and-claims.test.ts: seeds via raw SQL and calls the
+ * resolvers directly against the per-worker test DB.
+ */
+
+const SYSTEM_OWNER = '00000000-0000-0000-0000-000000000000';
+const ADMIN = 'gd-admin';
+const PLAIN_USER = 'gd-plain';
+const GYM_OWNER = 'gd-gym-owner';
+const CLAIMANT_BOTH = 'gd-claimant-both';
+const CLAIMANT_DUP_ONLY = 'gd-claimant-dup';
+
+const ALL_USERS = [SYSTEM_OWNER, ADMIN, PLAIN_USER, GYM_OWNER, CLAIMANT_BOTH, CLAIMANT_DUP_ONLY];
+
+const authCtx = (userId: string): ConnectionContext =>
+  ({ connectionId: `conn-${userId}`, isAuthenticated: true, userId }) as ConnectionContext;
+
+const anonCtx = (): ConnectionContext => ({ connectionId: 'conn-anon', isAuthenticated: false }) as ConnectionContext;
+
+const insertUser = (id: string) =>
+  db.execute(sql`
+    INSERT INTO "users" (id, email, name, created_at, updated_at)
+    VALUES (${id}, ${id + '@test.com'}, ${'User ' + id}, now(), now())
+    ON CONFLICT (id) DO NOTHING
+  `);
+
+const insertAdminRole = (userId: string) =>
+  db.execute(sql`
+    INSERT INTO community_roles (user_id, role, board_type, created_at)
+    VALUES (${userId}, 'admin', NULL, now())
+  `);
+
+const insertGym = async (opts: {
+  ownerId: string;
+  name: string;
+  latitude: number;
+  longitude: number;
+  createdAt?: string;
+  uuid?: string;
+}): Promise<{ id: number; uuid: string }> => {
+  const { ownerId, name, latitude, longitude, createdAt = '2026-01-01T00:00:00Z', uuid = uuidv4() } = opts;
+  const result = await db.execute(sql`
+    INSERT INTO gyms (uuid, name, slug, owner_id, latitude, longitude, is_public, created_at, updated_at)
+    VALUES (${uuid}, ${name}, ${uuid}, ${ownerId}, ${latitude}, ${longitude}, true, ${createdAt}, ${createdAt})
+    RETURNING id
+  `);
+  return { id: Number(Array.from(result as Iterable<{ id: number }>)[0].id), uuid };
+};
+
+const insertBoard = async (gymId: number, ownerId: string): Promise<void> => {
+  const uuid = uuidv4();
+  await db.execute(sql`
+    INSERT INTO user_boards
+      (uuid, slug, owner_id, board_type, layout_id, size_id, set_ids, name, gym_id, is_public, created_at, updated_at)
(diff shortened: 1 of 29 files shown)


marcodejongh:
Conflicts and PR feedcback requires fixing

marcodejongh:
Addressed the claude-review findings (commit `82c46a5`):

**Fixed**
1. **Unbounded dismissed-signatures load** — `fetchDismissedSignatures()` no longer materializes the whole dismissal history. It's now `fetchDismissedSignaturesAmong(signatures)`: build the candidate clusters first, then query `gym_merge_audit WHERE action='dismissed' AND cluster_signature IN (<current page's signatures>)`. The result set is bounded by the number of candidate clusters, not by how much the queue has been used over time.
2. **`error === 0` footgun** — both panel error states are now `{ offset: number } | null` (was `number | null`), so an offset of 0 can never read as "no error".
3. **Orphan rows now link to the gym** — added `slug` to `orphanGyms`/`OrphanGym`; each orphan row's name links to `/gym/${slug ?? gymUuid}` (same slug-first, uuid-fallback pattern the rest of the app uses).
4. **Duplicate `gym_members` DDL** — removed the dead second `DROP+CREATE gym_members` block in `schema-sql.ts` (kept the earlier one alongside the follow/member enrichment tables).

**Acceptable as-is (noted, not changed)**
- **`orphanGyms` two round-trips** (count + page) — fine at current scale; it's an admin-only, low-traffic audit list.
- **Vote consolidation is most-recent-wins** — intended merge semantic (a person has one current opinion of the gym, not a sum of stale votes). Added a one-line comment in `merge-gyms.ts` documenting it.

Validation: `vp check`, `vp run typecheck` (db/backend/web), backend tests (7/7, `VITEST_POOL_ID=solo-m2`), and web tests (5685/5685) all green; codegen no-drift.

PR #3700, opened 2026-07-16
feat(web): dark inputs on the Velvet elevated surface — white-input rule retired

marcodejongh:
## Summary

Phase 2 of the Velvet Send web epic (#3678 follow-on): dark mode stops forcing white input fields. All values were contrast-verified by the design panel before implementation:

- **Dark fields** ride the elevated violet surface `#2F234A` with a `rgba(195,188,211,0.5)` resting border (≥3:1 composited against field, card, and page); hover brightens the **border**, not the surface; text `#E7E2F0` (~9:1), placeholder `#A9A2B6` (AA).
- **MUI v7's dark Paper elevation overlay is disabled** and dialog/drawer/menu paper pinned to the token surface — the overlay was washing dark dialogs to ~`#49415B`, where secondary text silently fell below AA (invisible to token-based tests; now guarded).
- **Focus flips to the foreground violet `#A78BFA`** (the fill `#7C3AED` only cleared 3:1 against white fields — 2.5:1 on the new dark ones). The seven dark white-input overrides die, including the dual-tone `MuiInputLabel` hack that caused the clipped-label bug that kicked off this whole epic.
- **Autofill guard**: Chrome/Safari autofill can no longer flash a white fill over dark fields; caret colour rides the input text token.
- **Light error text darkens to `#B91C1C`** (was 4.40:1 at helper size on the tinted page — AA fail).
- **16px inputs unconditionally** and the `max-width: 768px` iOS zoom-guard block dies — it missed iPads entirely (the original bug screenshot device zoomed on focus).
- Guard suite: `--input-*` parity rows, ~10 new WCAG assertions, and a Paper-overlay-disabled tripwire. The CLAUDE.md white-input rule is rewritten.

QA: dev server on the branch at https://vibetunnel-dev-1.tailedd9d5.ts.net:3001 — exercise auth, search drawer, log-ascent, board edit, and any dialog in dark + light.

## Release Notes

- Dark mode forms finally look like they belong: input fields sit on the same violet surfaces as the rest of the app, with a clear violet focus ring — no more glaring white boxes.
- Autofill no longer flashes your login fields white in dark mode, and iPads stop zooming when you tap a field.

## Test plan

- `vp run typecheck:web`, `vp run build:web` — green.
- Theme guard suite: 47 tests green (parity rows for every `--input-*` var; ≥3:1 / ≥4.5:1 assertions for border, text, placeholder, focus, and error on every surface they sit on; Paper-overlay tripwire).
- Manual QA in both schemes on the dev server (see link above).

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -152,7 +152,7 @@ Vitest via `vp test`. Backend tests auto-start postgres+redis via `packages/back
 - **Web uses MUI.** Always MUI components and props. Avoid the `style` prop. Always use design tokens from `packages/web/app/theme/theme-config.ts` for colours/spacing — no hardcoded values.
 - Use CSS media queries for responsive design. Avoid JS breakpoint detection (`Grid.useBreakpoint()`).
 - Remove dead code as you go.
-- **Dark mode uses white input fields.** Intentional for contrast (`darkTokens.semantic.inputSurface`). Do not change.
+- **Dark mode inputs use the elevated dark surface** (`darkTokens.semantic.inputSurface`, #2F234A) with a violet `#A78BFA` focus ring — don't force white fields (`--input-*` vars in index.css are the source of truth).
 - **Never hardcode user-facing strings** in `packages/web/app/**/*.tsx` — all visible text via i18n catalogs. CI runs `vp run check:i18n` and `vp run check:i18n:orphans`. Mark unresolvable dynamic lookups with `// i18n-keep namespace.dotted.key`.
 - **Variable names describe contents.** No single-letter aliases (`r`, `x`, `s`) or vague placeholders (`data`, `info`, `temp`, `value`) outside tight loops. Destructure at the use site instead of generic aliases.
 - **Drizzle ORM over raw SQL.** Use `db.select/insert/update/delete()`. Raw `sql` from `drizzle-orm` only when the query genuinely can't be expressed otherwise. Importing `sql` from `@/app/lib/db/db` (raw Neon HTTP client) is lint-blocked.
--- a/docs/ai-design-guidelines.md
+++ b/docs/ai-design-guidelines.md
@@ -538,6 +538,11 @@ The web app renders Velvet Send in light and dark. Wiring notes specific to web:
 - **Scheme-awareness.** A direct `themeTokens.colors.{primary,success,error,warning,info}` read renders
   the _light_ value in dark mode — use the scheme-aware CSS vars instead: `var(--color-primary)`
   (foreground), `var(--color-primary-fill)` (fills), `var(--color-success|error|warning|info)` (status).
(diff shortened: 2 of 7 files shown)


marcodejongh:
All review items addressed in 1cb29f2ad:

1. **MuiInputLabel** — shared override keeps legacy floating labels on `text.secondary` (resting *and* shrunk — on the new dark fields one colour works for both, which is exactly why the dual-tone hack could die), quiet on focus so the 2px violet outline carries it. New guard assertions cover the shrunk label over the page and card in both schemes.
2. **MuiFilledInput / MuiAutocomplete** — filled inputs now ride the `--input-*` family; the Autocomplete **popup paper** is pinned to the elevated surface (it renders through a Popper, so the Menu/Popover pins didn't reach it). Autocomplete's *input* was already covered — it composes MuiOutlinedInput. Text pairings guarded.
3. **color-mix() fallback** — disabled fields get an opaque `--input-bg-hover` fallback behind an `@supports` gate for Safari < 16.2.
4. **blendOpaque + rgb()** — no change needed: the regex is `/rgba?\(/` which matches `rgb()`, and alpha already defaults to 1 when the fourth channel is absent (line below the match). The throw only fires for hex/named colours, which is a deliberate misuse guard for a compositing helper.
5. **Tests** — floating-label, filled/input-text, and elevated-popup-paper pairings added; theme suite now 50 assertions.

marcodejongh:
Round-two nits addressed in 625ab8bfd:

1. **`--input-text` wired** — `MuiInputBase` input colour now reads the var (single source of truth with the autofill guard).
2. **Firefox autofill** — the standard `input:autofill` block now carries unprefixed `box-shadow`/`background-color`/`color` (the WebKit-prefixed properties were inert there).
3. **Drawer radius parity** — false positive: `sharedComponents.MuiDrawer.styleOverrides.paper` already pins `borderRadius.lg` for both schemes; the dark override spreads those shared slots and only adds the elevated surface + border.
4. **Caret duplication** — the standard block no longer duplicates the WebKit properties; the two blocks now serve their respective engines.

marcodejongh:
Round three addressed in 3c24f6d41:

- **Disabled input text** — real bug, good catch: the direct `color` rule on the input element beat the wrapper's inherited `text.disabled`. Restated on the input slot itself (`color` + `WebkitTextFillColor` for WebKit's disabled rendering), with a per-scheme wiring test so it can't silently regress.
- **Doc gap** — `docs/ai-design-guidelines.md`'s web section now documents the `--input-*` family as the source of truth and the dark elevated-field rule.

marcodejongh:
Filled-input disabled dimming added in the latest commit — same surface + color-mix/@supports treatment as outlined fields. With the verdict at ready-to-merge two rounds running, I'll mark the PR ready once this CI cycle is green.

PR #3703, opened 2026-07-16
feat(web): the form kit — one label pattern, pilot on board form + log ascent

marcodejongh:
## Summary

Phase 3 of the Velvet Send web epic (stacks on #3700): the form design system that fixes the class of problems in the original bug report (clipped floating labels, inconsistent multiline fields, full-bleed iPad layouts, four competing label patterns).

**The kit** (`app/components/form/`): `FormShell` (640px measure, form-level alert, `focusFirstInvalid`), `FormSection` (one vertical rhythm), `FormField` (static label OUTSIDE the field — render-prop ARIA contract for MUI inputs, labelled-group path for custom controls, counters, conditional helper row), container-query `FormRow` (viewport breakpoints misfire inside dialog paper), `FormActions` (spinner keeps the label; bottom drawers carry actions in the header per the design-panel ruling), `FormSwitchRow` (44px, kills the ml:7 helper hack), `FormDateTimePicker`. The MUI type ramp is pinned (the 16/14 coefficient inflated every unpinned variant) and guard-tested.

**Pilots**: the Edit Board form (the original screenshot offender) rebuilt as Board details / Location / Setup / Visibility with counters and a card-style map disclosure; the log-ascent form loses its 120px label column, gains Details / How-it-felt sections, and its silent failures now speak (inline angle error, form-level alerts for validation + save failures).

165 pre-existing tests pass unchanged (label queries resolve through FormField's wiring); 80 kit + theme guard tests; new strings in en/es/fr.

**Resolved review nit**: the My Boards drawer double-title (drawer's "Board" header over the form's "Edit Board" heading) is fixed — `BoardDetailContent` reports an edit-host descriptor (`onEditHostChange`) so the drawer chrome claims the title and Save action while editing, and the form renders no in-body heading. The headerless search-result drawer keeps the inline title by design.

Screenshots (both themes, phone + iPad) attached in comments. Live QA: https://vibetunnel-dev-1.tailedd9d5.ts.net:3001

## Release Notes

- Forms got a full redesign: clear labels that never clip, sensible widths on tablets, grouped sections, and character counters where they help.
- Logging an ascent now tells you when something goes wrong instead of failing silently.

## Test plan

- `vp run typecheck:web` green; kit + theme guards 80/80; board-entity + logbook + form suites 222/222; `vp run check:i18n` + orphans green.
- Manual QA on the dev server in both themes and viewports.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -152,7 +152,7 @@ Vitest via `vp test`. Backend tests auto-start postgres+redis via `packages/back
 - **Web uses MUI.** Always MUI components and props. Avoid the `style` prop. Always use design tokens from `packages/web/app/theme/theme-config.ts` for colours/spacing — no hardcoded values.
 - Use CSS media queries for responsive design. Avoid JS breakpoint detection (`Grid.useBreakpoint()`).
 - Remove dead code as you go.
-- **Dark mode uses white input fields.** Intentional for contrast (`darkTokens.semantic.inputSurface`). Do not change.
+- **Dark mode inputs use the elevated dark surface** (`darkTokens.semantic.inputSurface`, #2F234A) with a violet `#A78BFA` focus ring — don't force white fields (`--input-*` vars in index.css are the source of truth).
 - **Never hardcode user-facing strings** in `packages/web/app/**/*.tsx` — all visible text via i18n catalogs. CI runs `vp run check:i18n` and `vp run check:i18n:orphans`. Mark unresolvable dynamic lookups with `// i18n-keep namespace.dotted.key`.
 - **Variable names describe contents.** No single-letter aliases (`r`, `x`, `s`) or vague placeholders (`data`, `info`, `temp`, `value`) outside tight loops. Destructure at the use site instead of generic aliases.
 - **Drizzle ORM over raw SQL.** Use `db.select/insert/update/delete()`. Raw `sql` from `drizzle-orm` only when the query genuinely can't be expressed otherwise. Importing `sql` from `@/app/lib/db/db` (raw Neon HTTP client) is lint-blocked.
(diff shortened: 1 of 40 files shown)


marcodejongh:
Both review passes addressed in `3bd5893a7`:

**21:32 review**
- **subtitle2 / raw sx values**: root cause was the unpinned `subtitle2` inflating to 16px via MUI's 16/14 coefficient — the sx overrides were compensating. `subtitle2` is now pinned in the shared ramp (14/600/20, same treatment as the headings), `FormSection` trusts the bare variant, `map-location-picker` is back on `body2` with a `themeTokens` font weight, `FormField`'s counter emphasis uses `themeTokens.typography.fontWeight.*`, and the config Alert lost its hardcoded `0.8rem`.
- **External-submit test**: added — a `FormActions` rendered *outside* the `<form>` with `formId` now has a test proving a click actually fires the form's `onSubmit` (the header-hosted drawer path).
- **Error-then-dismiss test**: new `logascent-form-error-banner.test.tsx` pins the lifecycle: failed save surfaces the banner and keeps the form open, a successful re-submit clears it, and a log-type switch clears it.
- **Double-title tracking**: no issue needed — it's already fixed in this PR (`688f6117e`): the My Boards drawer claims the edit title + Save via `onEditHostChange`, so only one title renders. PR body updated (the "known nit" paragraph was stale).

**16:05 review**
- **Stale banner on log-type switch**: fixed — the ToggleButtonGroup's onChange clears `formError` (covered by the new test above).
- **"Détails de la board"**: correct per `docs/i18n-french-glossary.md` — the board device is deliberately « la board » (feminine) in every French string; « planche »/« panneau » are ruled out. Only brand names stay untranslated for trademark reasons; this is the common-noun term French board climbers actually use.
- **`onSubmitStateChange` in effect deps**: kept (it's the honest dependency), but the stability contract is now documented on the prop in both pilot forms.

marcodejongh:
All three findings addressed in `12bbcb343`:

1. **Counter announced to screen readers**: the counter now carries `${inputId}-counter` and joins `aria-describedby` (alone or space-joined with the helper id), exactly as suggested. Test added asserting the control's `aria-describedby` resolves to the counter text when no helper is present.
2. **Both-present test**: added — asserts `aria-describedby` lists two ids resolving to the helper text and the counter text.
3. **FormRow null child**: fixed for real rather than just tested — the grid now sizes from `Children.toArray(children).length` (which drops null/undefined/boolean slots) instead of `Children.count`, so `{condition && <Field/>}` can't leave a ghost column. Test covers the null-child row.

(The earlier red checks on this PR were the GitHub Actions incident — jobs died in setup with GitHub error pages in their logs; they're rerunning.)

PR #3712, opened 2026-07-16
fix(mobile): default bug report contact consent to opt-out

marcodejongh:
## Summary

The "Can we contact you about this?" switch in the mobile bug report sheet (`packages/mobile/src/components/user-drawer/FeedbackSheet.tsx`) started unchecked (`contactConsent = false`), requiring reporters to actively opt in before we could follow up on a bug report. This flips the default to checked (`true`) so contact is opt-out instead of opt-in, across initial state, the mode-change reset effect, and the post-submit reset.

Web's bug report form has no contact-consent toggle at all yet (`packages/web/app/components/feedback/feedback-form.tsx` / `feedback-dialog.tsx`), so this change is scoped to mobile, the only surface with the switch today.

## Release Notes

- Bug reports now default to letting us follow up with you — flip the switch off if you'd rather we didn't.

## Test plan

- [ ] `vp run typecheck:mobile` (not runnable in this sandbox — no `node_modules`/`vp` installed)
- Manual: open the bug report sheet while signed in and confirm the "Can we contact you about this?" switch starts on.

---
_Generated by [Claude Code](https://claude.ai/code/session_01N65fkTHFy2eAU389Wdq4WQ)_

### DIFF
--- a/packages/mobile/src/components/user-drawer/FeedbackSheet.tsx
+++ b/packages/mobile/src/components/user-drawer/FeedbackSheet.tsx
@@ -43,7 +43,7 @@ export function FeedbackSheet({ sheetRef, mode, showDiscordLink = false }: Feedb
   const [selectedRating, setSelectedRating] = useState<number | null>(null);
   const [comment, setComment] = useState('');
   const [captureBleDiag, setCaptureBleDiag] = useState(false);
-  const [contactConsent, setContactConsent] = useState(false);
+  const [contactConsent, setContactConsent] = useState(true);
 
   const isBugReport = mode === 'bug';
   const trimmedComment = comment.trim();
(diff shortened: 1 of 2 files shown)


marcodejongh:
Re: the GDPR concern — confirmed with the requester and keeping the opt-out default as-is. This toggle doesn't gate new data collection, only whether we follow up on a bug report the user is already submitting under their own account, so we're treating it as acceptable risk here.

Fixed the commit-lint failure (PR title now follows Conventional Commits).

---

marcodejongh:
Added the mode-change reset test. Re: GDPR — already discussed and resolved earlier in this thread; keeping the opt-out default as the requester decided.

---

PR #3713, opened 2026-07-16
feat: location-sync guarded 150 m match tier + user-owned gym aliasing

marcodejongh:
## Summary

The location-sync importer (`packages/location-sync`) is the biggest active generator of duplicate gym listings. Its only physical-match tier required `owner_id = SYSTEM` **and** exact normalized name **and** `ST_DWithin` 20 m. Verified prod data shows cross-provider pins for the *same* gym drift 40–90 m (24 of 46 dup pairs span two providers), so every new provider minted a twin. User-owned/claimed gyms were invisible to the match (hard owner filter), so the importer re-created gyms owners already curate.

This adds a **guarded second match tier** and makes **user-owned gyms matchable, alias-and-leave-alone** — with squat protection and an ambiguity guard. Net user-visible effect over time: fewer duplicate gym listings appearing as new providers sync.

## Release Notes

## Match tiers (`classifyGymMatch` / `resolveGymIdForSource`)

- **Tier 1 (≤ 20 m)** — a pin almost on top of a gym is the same gym. SYSTEM-owned candidates always match here; a **user-owned candidate only matches when the name is NOT generic** (a public "Home Wall" 15 m from a stranger's pin must not capture it). Multiple tier-1 candidates pick the canonical one.
- **Tier 2 (20 m < d ≤ 150 m)** — only for a non-generic name, gated by three guards:
  - **Same-provider guard:** a candidate that already carries an alias for the incoming provider prefix pointing at a *different* source key is refused (same-provider twins 150 m apart are likely distinct walls) → structured `warn`.
  - **Squat protection:** tier-2 only binds into a user-owned gym when it carries an **approved claim**. An unclaimed user-owned candidate is never auto-captured — it skips to the `/admin/gym-duplicates` queue where a human merges (the merge re-points the alias) → structured `info`.
  - **Ambiguity guard:** if more than one candidate clears the tier, refuse to auto-match (two same-name gyms in one 150 m circle is itself evidence the name is generic-ish) → structured `warn`.
- **Tier-1 dead-zone log:** a ≤ 20 m user-owned candidate dropped by the tier-1 generic-name guard can't reach tier 2, so it's surfaced with the same admin-queue `info` signal as the unclaimed tier-2 skip instead of vanishing silently before a fresh gym is minted.

## User-owned gyms: alias-and-leave-alone

- The match query now spans non-SYSTEM public live gyms too.
- A match into an owner-curated gym (user-owned, or SYSTEM with an approved claim) writes the source alias but **never** refreshes its name/coords/metadata. The initial bind is logged at **`warn`** (security-relevant); the steady-state per-cycle alias resolution is **`debug`** (it fires ~48×/day/gym and would drown the tier-2 signal).
- **Squat protection at the alias path:** `findAliasedGym` reads owner + approved-claim, so an existing alias never licenses a metadata refresh into an owner-curated gym. Audited every consumer of the alias table (`upsert.ts`, `dedupe-gym-locations.ts`) — nothing else treats "has alias" as "SYSTEM-owned".
- The candidate mapping **throws** on a non-array `aliasSourceKeys` — a silent `[]` would disable the same-provider guard.

## Denylist (`GENERIC_GYM_NAMES`)

`ReadonlySet<string>` declared right after `normalizeGymName`, matched **exactly** on the whole normalized name (never a substring): home wall, homewall, home, garage, cellar, basement, kilter, kilter board, tension, tension board, moonboard, moon board, moon, grasshopper, decoy, soill.

**Inclusion criterion:** board-brand names and homewall genericisms — **not gym-chain names**. A home wall gets named after its board or where it lives, so those collide across unrelated setups; a real gym chain's pins *do* drift 40–90 m and must auto-merge. `touchstone` was intentionally **removed** (Touchstone is a real CA chain) — capping it at 20 m would foreclose the exact drift-merges this PR fixes.

**Cross-PR (resolved):** #3701 merged first and declared its own `GENERIC_GYM_NAMES` in the same spot. This branch is **rebased onto merged `main`** and the conflict is resolved to a **single declaration** — the union of both lists **minus `touchstone`** (#3701's 13 entries + `grasshopper`/`decoy`/`soill`). #3701's consumers (`gym-matching.ts`, `boards.ts`) compile against it (typecheck green).

**Raw-SQL convention:** `findGymMatchCandidates` is a hand-written raw SQL block that literal-names every table (`gyms`, `gym_claims`, `location_sync_gym_sources`, …), matching the importer's existing match-query style; the lighter `findAliasedGym` is a Drizzle builder query using the `${gymClaims}` reference. The two differ by query style, not oversight — documented in a code comment.

## Known limitation (follow-up)

moonboard source keys embed the pin coordinates (`moonboard:<name>:<lat>:<lng>`), so a pin that moves > 20 m mints a new source key; the same-provider guard then warns and mints a permanent twin. Left a code comment at the guard; a key-scheme redesign is a separate follow-up issue.

## Structured logging

`LocationSyncLogger` seam (winston-compatible `debug`/`info`/`warn`), injected into `upsertPublicBoardLocations` and wired into the aurora/kilter/moonboard callers via `toLocationSyncLogger(args.log)`. Logs every tier-2 match, every rejection/skip (generic, same-provider, unclaimed-user-owned, ambiguous), and the initial owner-curated bind.

## Test plan
- `packages/location-sync/src/upsert.test.ts` (31 tests): tier-1 wins at 20 m; tier-2 exact-name match at 150 m; generic blocked at 150 m but matches at 20 m for SYSTEM; generic user-owned does NOT capture a stranger pin at tier-1; same-provider skip + log; unclaimed user-owned tier-2 skip + log; approved-claim user-owned tier-2 bind + no refresh + warn; tier-1 user-owned bind + no refresh + warn; ambiguous two-candidate skip + warn; existing alias into user-owned gym never refreshes; 151 m does not match; query widening.
- `packages/location-sync/src/logger.test.ts`: level prefixes (debug/info/warn), JSON suffix, empty-fields, no-op path.
- `packages/db/src/queries/gyms/__tests__/location-dedupe.test.ts`: guarded distance, generic-name matching, denylist set contents.
- Validation: `vp check`; `vp run typecheck`; `vp test run --project location-sync`; `VITEST_POOL_ID=solo-m3 vp test run --project backend`.

https://claude.ai/code/session_019VGxf4ZJGFRanuJGrRMDgY

### DIFF
--- a/packages/aurora-sync/src/sync/locations-sync.ts
+++ b/packages/aurora-sync/src/sync/locations-sync.ts
@@ -2,6 +2,7 @@ import type { PgDatabase, PgQueryResultHKT } from 'drizzle-orm/pg-core';
 import {
   formatLocationBoardName,
   resolveDefaultAuroraLocationConfig,
+  toLocationSyncLogger,
   upsertPublicBoardLocations,
   type LocationSyncSummary,
   type PublicBoardLocationInput,
@@ -59,7 +60,9 @@ export async function syncAuroraBoardLocations(args: {
 }): Promise<LocationSyncSummary> {
   const pins = await fetchAuroraPins(args.board);
   const { records, skipped } = buildAuroraLocationRecords(args.board, pins.gyms);
-  const summary = await upsertPublicBoardLocations(args.db, records);
+  const summary = await upsertPublicBoardLocations(args.db, records, {
+    logger: toLocationSyncLogger(args.log),
+  });
   const mergedSummary = {
     ...summary,
     boardsSkipped: summary.boardsSkipped + skipped.length,
--- a/packages/db/src/queries/gyms/__tests__/location-dedupe.test.ts
+++ b/packages/db/src/queries/gyms/__tests__/location-dedupe.test.ts
@@ -65,12 +65,15 @@ void describe('guarded second match tier', () => {
     assert.equal(isGenericGymName('Board House'), false);
   });
 
-  void it('exposes an all-normalized denylist', () => {
+  void it('exposes an all-normalized denylist set', () => {
     for (const name of GENERIC_GYM_NAMES) {
       assert.equal(normalizeGymName(name), name);
     }
-    assert.ok(GENERIC_GYM_NAMES.includes('home wall'));
-    assert.ok(GENERIC_GYM_NAMES.includes('kilter board'));
+    assert.ok(GENERIC_GYM_NAMES.has('home wall'));
+    assert.ok(GENERIC_GYM_NAMES.has('kilter board'));
+    assert.ok(GENERIC_GYM_NAMES.has('tension board'));
+    assert.ok(GENERIC_GYM_NAMES.has('moon board'));
+    assert.ok(GENERIC_GYM_NAMES.has('touchstone'));
   });
 });
 
--- a/packages/db/src/queries/gyms/__tests__/location-dedupe.test.ts
+++ b/packages/db/src/queries/gyms/__tests__/location-dedupe.test.ts
@@ -3,9 +3,13 @@ import assert from 'node:assert/strict';
 import {
   chooseCanonicalGymCandidate,
   distanceMeters,
+  GENERIC_GYM_NAMES,
   groupPhysicalGymCandidates,
+  GYM_MATCH_GUARDED_DISTANCE_METERS,
   gymCompletenessScore,
+  isGenericGymName,
   normalizeGymName,
+  PHYSICAL_GYM_MATCH_DISTANCE_METERS,
   type CanonicalGymCandidate,
 } from '../location-dedupe';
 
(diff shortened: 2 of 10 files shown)


marcodejongh:
Conflicts

marcodejongh:
Note for reviewers: the automated review above (posted 10:08Z) flags two items that are already resolved on the current head `afbabd291` — it appears to have evaluated the pre-fix diff. Evidence:

**1. "User-owned + generic + ≤20 m is a silent no-match (no log, no test)"** — fixed. `classifyGymMatch` collects these dropped tier-1 candidates into `tierOneGenericUserOwnedSkips`, and `resolveGymIdForSource` emits an `info` admin-queue signal for each (`location-sync tier-1 match skipped: generic-named user-owned gym (surfaces in admin duplicates queue)`). Covered by the test *"does NOT let a generic-named user-owned gym capture a stranger pin at tier-1, and logs the near-miss"* (`packages/location-sync/src/upsert.test.ts`), which asserts the fresh gym **and** the log line with `candidateGymId`/`ownerId`/`distanceMeters`.

**2. "`'touchstone'` in `GENERIC_GYM_NAMES`"** — removed. `touchstone` is no longer in the set (a real gym chain must auto-merge across its 40–90 m drift). The inclusion criterion is now documented in the code comment (board-brand / homewall genericisms, **not** gym-chain names), and a db test asserts `isGenericGymName('Touchstone') === false`.

**3. Cross-PR conflict** — resolved: #3701 merged first; this branch is rebased onto `main` with a single reconciled declaration (union of both lists minus `touchstone`). Its consumers (`gym-matching.ts`, `boards.ts`) compile (typecheck green).

All CI checks are green on `afbabd291`.

PR #3720, opened 2026-07-16
fix: replace dead board_products_angles query with static ANGLES

### LINKED ISSUE #2379: relation "board_products_angles" does not exist (11× over 6 days)
## Signature

```
ERROR: relation "board_products_angles" does not exist at character 44
STATEMENT:
  SELECT DISTINCT pa.angle
  FROM board_products_angles pa
  JOIN board_layouts l
    ON l.board_type = pa.board_type AND l.product_id = pa.product_id
  WHERE l.board_type = $1 AND l.id = $2
  ORDER BY pa.angle ASC
```

## Frequency (Railway log export, 2026-05-24 → 2026-05-30)

- **11×** Postgres `relation "board_products_angles" does not exist`. Concentrated 2026-05-25 08:10 (multiple back-to-back, suggesting retries from the calling code path).

## Confirmed via readonly psql

The table really doesn't exist. Tables matching `%product%` or `%angle%` in production:

```
board_product_sizes
board_product_sizes_layouts_sets
board_products
```

No `board_products_angles`. Most plausible intended targets:
- A renamed table that didn't get its callers updated
- A table that was supposed to ship as part of a migration but was missed
- Possibly the data lives inline on `board_products` or `board_layouts` now

## Code pointers

The query joins `board_layouts` filtered by `board_type` + layout `id`, ordered by angle — almost certainly the angle-picker / board-configuration code path. Likely lives in a board-config service or a layout-detail resolver. Grep for the literal string `board_products_angles` or the SQL snippet `SELECT DISTINCT pa.angle`.

## Suggested next steps

1. Find the code that issues this query (raw SQL string or Drizzle expression).
2. Decide: is the table missing (write a migration in `packages/db/`), or has the data moved to an existing table (rewrite the query)?
3. Add a migration regression test so renamed-table-without-caller-update can't slip through again.

---
**Note for assignee:** Verify this is not a duplicate of an existing open issue before working on it. The GitHub search API was rate-limited during triage so no dedup pass was run. Bulk-filed from a 6-day Railway log export on 2026-05-30.

marcodejongh:
## Summary

`board_products_angles` never existed in the unified schema. It's the mechanically-renamed remnant of `kilter_products_angles`/`tension_products_angles`, which were deliberately dropped in `74e8d2a0` ("Remove unused database tables to clean up schema", Jul 2025) because Aurora doesn't sync a per-layout angle table — see `packages/aurora-sync/src/sync/shared-sync.ts:57`: *"`products_angles`: angles are hardcoded in the `ANGLES` constant in board-data.ts."*

**Root-cause chain:**
1. `74e8d2a0` (Jul 2025) drops the real `kilter_products_angles`/`tension_products_angles` tables as unused — angles are, and always were, a static per-board-type range (`ANGLES` in `packages/shared/board-config/src/board-data.ts`), never something that varies by layout.
2. Nobody updated the `angles` resolver, which kept querying the now-gone tables.
3. The Jan 2026 schema-unification refactor mechanically renamed the leftover dead SQL text (`kilter_products_angles` → `board_products_angles`, `layouts` → `board_layouts`) without noticing the table itself had been gone for months — preserving the bug through the rewrite.
4. The web heatmap-prewarm cron (Apr 2026) copy-pasted the already-broken query from the backend resolver.

**Real-world impact:** two live call sites hit this query.
- The backend GraphQL `angles` resolver — the exact stack trace reported in #2379 (11× over 6 days).
- `packages/web/app/api/internal/prewarm-heatmap/[board_name]/route.ts`, a weekly Vercel cron for 5 boards. It caught the DB error per layout and fell back to an empty angle list, so `buildTargetsForBoard` skipped every layout — **the heatmap-prewarm cron has been silently warming zero cache entries every week since it shipped**, while still returning a 200 "success" response. This is now fixed as a side effect: angles come from a static lookup, so real (layout, size, angle) combinations get built and warmed again.

Also fixed the same defect family in the legacy `/api/v1/angles/{board_name}/{layout_id}` REST route: it queried an even older table (`kilter_products_angles` joined to a nonexistent `layouts` table) and ignored the `board_name` param entirely, always assuming Kilter. It's now parameterized correctly and un-skips the `rest-parity.test.ts` test that had a standing `TODO: Re-enable when REST API is fixed`.

All three call sites now read `ANGLES` (`@boardsesh/board-config`), the same static source web's actual angle-picker UI already uses — so this also removes three now-unnecessary DB round trips.

Closes #2379.

## Release Notes

## Test plan

- Added a regression test in `packages/backend/src/__tests__/rest-parity.test.ts` (previously `it.skip`'d with a stale TODO) asserting the GraphQL `angles` resolver returns the correct static, ascending angle list for kilter/tension/moonboard — no more live-production dependency, since that would create a chicken-and-egg problem for pre-merge CI (this PR's REST fix isn't deployed yet).
- Added `packages/web/app/api/v1/angles/[board_name]/[layout_id]/__tests__/route.test.ts`: valid Aurora board returns the static angle list ignoring `layout_id`, board-name parameterization works (tension ≠ kilter), unknown board name returns 400.
- Added `packages/web/app/api/internal/prewarm-heatmap/[board_name]/__tests__/route.test.ts`: 401 without cron secret, 400 for a non-Aurora board, and — the direct regression check for the silent no-op — a real Aurora board now builds and warms a non-empty target set (`warmed === total`, `failed === 0`), where it previously warmed zero.
- `vp check` — clean.
- `vp run typecheck:backend` / `vp run typecheck:web` — clean.
- `vp test run --project backend --reporter=agent` — 170/178 files passed, 2336/2399 tests passed; the 8 failing files are pre-existing local Postgres worker-DB provisioning flakes (`[setup] Cannot initialise worker database`) unrelated to any file this PR touches — `rest-parity.test.ts` itself passed. CI provisions Postgres/Redis fresh per job so this should be clean there.
- `vp test run --project web` (scoped to `prewarm-heatmap`/`angles`) — 2 files, 6 tests, all passed.

### DIFF
--- a/packages/backend/src/__tests__/rest-parity.test.ts
+++ b/packages/backend/src/__tests__/rest-parity.test.ts
@@ -1,4 +1,5 @@
 import { describe, it, expect, beforeAll, afterAll } from 'vite-plus/test';
+import { ANGLES } from '@boardsesh/board-config';
 import { startServer } from '../server';
 
 const BACKEND_PORT = 8083; // Use different port to avoid conflicts with other tests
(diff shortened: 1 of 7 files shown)


marcodejongh:
Conflcits

PR #3729, opened 2026-07-16
fix: address follow-up review on kiosk heartbeat

marcodejongh:
## Summary

Follow-up on the merged kiosk heartbeat feature (#3685), addressing a round of review feedback:

- **`kiosk-liveness-chip.tsx`**: the Tooltip wrapper checked `lastSeenAt === null` to decide whether to format a date. A non-null but unparseable `lastSeenAt` slipped past that check and rendered "Invalid Date" in the tooltip. Now reuses `bucketKioskLiveness`'s already-computed `'never'` status (which already treats null/unparseable identically), so the same bucketing decides both the label and whether a date gets formatted.
- **`kiosk-heartbeat.test.ts`**: added coverage for the viewport-clamping rejection path — `KioskHeartbeatInputSchema` rejects `viewportWidth`/`viewportHeight` outside `[1, 20000]`, but nothing exercised that before.
- **`kiosk-heartbeat.test.ts`**: documented the rate-limit test's implicit assumption (fixed `connectionId`, no explicit bucket reset in `beforeEach`) for future readers.
- **`kiosk-heartbeat.ts`**: noted that `clientOverrideForTests` is module-level mutable state that's fine under today's serial test execution but would race under concurrent Vitest workers.

Other findings from the review (per-process rate limiting, the 1-hour existence-cache TTL, the manage tab's 60s poll interval, and a minor hand-written-vs-generated type divergence for `lastSeenAt`) were already explicitly called out as acceptable trade-offs in code comments or have no runtime impact, so no code change was made for those.

## Release Notes

## Test plan

- `vp run typecheck:web` — passed
- `vp test run --project backend` (full suite, real Postgres + Redis) — 2484 tests passed, including the new viewport-clamping cases
- `vp test run --project web` (full suite) — 5390 tests passed
- `vp check` — no new lint/format issues (the one pre-existing error is in an unrelated file, `packages/db/src/queries/__tests__/tick-offset-inference.test.ts`, untouched by this change)

---
_Generated by [Claude Code](https://claude.ai/code/session_017DxND2AFXabGztRChvejvn)_

### DIFF
--- a/packages/backend/src/__tests__/kiosk-heartbeat.test.ts
+++ b/packages/backend/src/__tests__/kiosk-heartbeat.test.ts
@@ -202,7 +202,10 @@ describe('kioskHeartbeat rate limiting', () => {
 
   it('trips the per-client rate limit past the ceiling', async () => {
     // A single fixed connection so the in-memory bucket accumulates. The
-    // heartbeat ceiling is 60/min; the 61st call must be throttled.
+    // heartbeat ceiling is 60/min; the 61st call must be throttled. `connectionId`
+    // is unique to this test (no other test reuses 'kh-rate-limit-conn'), so the
+    // in-memory bucket starting non-empty on a re-run isn't a real hazard —
+    // beforeEach only resets the DB, not the rate limiter, on purpose.
     const ctx: ConnectionContext = {
       connectionId: 'kh-rate-limit-conn',
       isAuthenticated: false,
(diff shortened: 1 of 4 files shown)


marcodejongh:
## Fable review follow-up

Fable's review confirmed this PR is a small, correct fix (not superseded — main still has the `lastSeenAt === null` tooltip bug and no viewport-clamping test coverage) and flagged two things:

**Major — CI red from transient infra failure, branch needed rebase + push**
The `changes`/`large-files` jobs failed on 2026-07-16 on a GitHub-served HTML error page (transient outage), which cascaded to skip every real check. Fixed: rebased onto current `main` (881 commits behind, clean rebase, no conflicts) and pushed.

While re-running CI, I also found a second, *non-transient* gate failure the review didn't catch: `commit-lint` rejected the PR title's `kiosk` scope — it's not in the current `ALLOWED_SCOPES` allowlist (`scripts/lib/commit-message-transform.ts`). This scope requirement is newer than this branch. Fixed by retitling the PR to drop the scope (`fix: address follow-up review on kiosk heartbeat`) and pushing an empty commit, since `pull_request` CI only listens for opened/synchronize/reopened, not `edited`.

**Minor — no component test for the tooltip behavior change**
Added `packages/web/app/components/gym-entity/manage/__tests__/kiosk-liveness-chip.test.tsx`, asserting:
- an unparseable `lastSeenAt` renders "No signal yet" with no tooltip (the regression this PR fixes)
- a `null` `lastSeenAt` renders the same way
- a parseable `lastSeenAt` still renders a tooltip with a correctly formatted (non-"Invalid Date") timestamp

CI is green (typecheck, lint, i18n, test-backend, test-default, codegen-drift, commit-lint, claude-review, ci-status). Marking ready for review.

PR #3734, opened 2026-07-16
feat(web): migrate the auth form family onto the form kit (Wave 2)

marcodejongh:
## Summary

Wave 2 of the Velvet Send form migration — and per the expo-web decision, the **final** migration wave. The whole auth family moves onto the form kit from #3703 (this PR stacks on that branch and retargets `main` when it merges):

- **Sign-in / sign-up page + auth modal**: FormShell → FormSection → FormField (render-prop wiring) → stacked FormActions. One static-label pattern on the existing i18n keys; the tab structure, social buttons, and divider are untouched. "Forgot password?" becomes a `labelAccessory` link on the password field.
- **Forgot / reset / verify-request**: same conversion. The verify-request resend email input finally gets an accessible name (new `verifyRequest.fields.email` key in en/es/fr — it was placeholder-only, a WCAG 4.1.2 failure).
- **Server errors surface in the form**, not as toasts, while the form is visible (invalid credentials, NextAuth `?error=` redirects, registration and resend failures — the resend catch previously swallowed the error entirely). Success toasts on navigate-away paths stay.
- **Autofill actually works now**: the auth inputs never had `autoComplete` — login gets `username`/`current-password`, register gets `name`/`email`/`new-password`, email fields get `type="email"` + no autocapitalize. Password show/hide toggles get aria-labels.
- **Playlist-selection mini-form** (create playlist name/description/colour) joins the kit; the popover stays compact (no FormShell width cap).
- No `size` props anywhere — theme defaults only.

Built by an Opus implement lane + Sonnet mechanical lane, adversarially reviewed by Opus, findings fixed (the verify-request half-touch and email autoComplete parity).

## Release Notes

- Signing in is smoother: your password manager can finally autofill Boardsesh, and login errors show up right in the form instead of a vanishing popup.

## Test plan

- [x] `vp test run` — auth + form suites: 72/72 after rebasing onto the latest form-kit branch
- [x] `vp run typecheck:web`, `vp check`, `vp run check:i18n` — clean
- [x] Catalog parity (en/es/fr) maintained — new key added to all three locales
- [ ] CI green on this PR

### DIFF
--- a/packages/shared/i18n/locales/en-US/auth.json
+++ b/packages/shared/i18n/locales/en-US/auth.json
@@ -95,6 +95,9 @@
     "title": "Check your email",
     "description": "We sent you a verification link. Click the link in your email to verify your account.",
     "resendPlaceholder": "Enter your email to resend",
+    "fields": {
+      "email": "Email"
+    },
     "resend": "Resend Verification Email",
     "back": "Back to Login",
     "messages": {
--- a/packages/shared/i18n/locales/en-US/auth.json
+++ b/packages/shared/i18n/locales/en-US/auth.json
@@ -69,7 +69,9 @@
     },
     "a11y": {
       "showPassword": "Show password",
-      "hidePassword": "Hide password"
+      "hidePassword": "Hide password",
+      "showConfirmPassword": "Show password confirmation",
+      "hideConfirmPassword": "Hide password confirmation"
     }
   },
   "error": {
--- a/packages/shared/i18n/locales/en-US/climbs.json
+++ b/packages/shared/i18n/locales/en-US/climbs.json
@@ -258,6 +258,12 @@
         "descriptionPlaceholder": "Optional description...",
         "colorLabel": "Color (optional)",
         "submit": "Create"
+      },
+      "row": {
+        "removeAria": "Remove from playlist {{name}}",
+        "addAria": "Add to playlist {{name}}",
+        "climbCount_one": "{{count}} climb",
+        "climbCount_other": "{{count}} climbs"
       }
     },
     "sendToBoard": {
--- a/packages/shared/i18n/locales/es/auth.json
+++ b/packages/shared/i18n/locales/es/auth.json
@@ -69,7 +69,9 @@
     },
     "a11y": {
       "showPassword": "Mostrar contraseña",
-      "hidePassword": "Ocultar contraseña"
+      "hidePassword": "Ocultar contraseña",
+      "showConfirmPassword": "Mostrar la confirmación de contraseña",
+      "hideConfirmPassword": "Ocultar la confirmación de contraseña"
     }
   },
   "error": {
--- a/packages/shared/i18n/locales/es/auth.json
+++ b/packages/shared/i18n/locales/es/auth.json
@@ -95,6 +95,9 @@
     "title": "Revisa tu correo",
     "description": "Te enviamos un enlace de verificación. Haz clic en el enlace para verificar tu cuenta.",
     "resendPlaceholder": "Escribe tu correo para reenviar",
+    "fields": {
+      "email": "Correo electrónico"
+    },
     "resend": "Reenviar correo de verificación",
     "back": "Volver a iniciar sesión",
     "messages": {
--- a/packages/shared/i18n/locales/es/climbs.json
+++ b/packages/shared/i18n/locales/es/climbs.json
@@ -258,6 +258,12 @@
         "descriptionPlaceholder": "Descripción opcional...",
         "colorLabel": "Color (opcional)",
         "submit": "Crear"
+      },
+      "row": {
+        "removeAria": "Quitar de la playlist {{name}}",
+        "addAria": "Añadir a la playlist {{name}}",
+        "climbCount_one": "{{count}} bloque",
+        "climbCount_other": "{{count}} bloques"
       }
     },
     "tick": {
--- a/packages/shared/i18n/locales/fr/auth.json
+++ b/packages/shared/i18n/locales/fr/auth.json
@@ -69,7 +69,9 @@
     },
     "a11y": {
       "showPassword": "Afficher le mot de passe",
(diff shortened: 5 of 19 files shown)


marcodejongh:
Review findings addressed in the latest push:

- **handleLogin catch swallowing network errors** (page + modal): fixed — a thrown `signIn` now sets the form-level `authFailed` alert. Covered by a new `auth-modal-server-errors.test.tsx` (auth-error path, network-throw path, and successful-retry-clears — the test that would have caught this).
- **verify-request back button**: now `component={LocaleLink}`, matching its siblings.
- **EMAIL_REGEX duplication**: verify-request imports the `validate-fields` export.
- **playlist-selection hardcoded English**: the climb count and the add/remove aria-labels moved to `actions.playlist.row.*` keys in all three locales (es: bloque/bloques, fr: bloc/blocs, matching the existing catalog voice).
- **`useSnackbar` in reset-password**: **not dead** — `showMessage(t('resetPassword.toasts.success'))` still fires on the success path right before `router.replace` (the deliberate navigate-away toast the wave rules allow). No change.

Analytics event names ('Add to Playlist' / 'Remove from Playlist') are intentionally untranslated — they're PostHog event strings, not user-facing copy.

marcodejongh:
Round-2 findings:

- **`focusFirstInvalid` not wired** — real, fixed. New `focusFirstInvalidAfterRender(formId)` helper in the form kit (the call has to happen after React commits the failed-validation render — synchronously the fields don't carry `aria-invalid` yet), wired into all five validating forms: login page + modal (both tabs each), forgot/reset password, verify-request. Focus-management test added (submit empty login → focus lands on Email).
- **Playlist create fields "lost" `size="small"`** — false positive: the theme sets `MuiTextField.defaultProps.size: 'small'` (`packages/web/app/theme/mui-theme.ts:149`, shipped in #3703), so dropping the explicit prop renders identically. That's exactly why the wave removed per-instance size props.
- **Focus test** — added, per above.

marcodejongh:
Round-3 findings:

- **Register error-path tests**: added on both surfaces — API error body surfaces in the register alert, a thrown fetch shows the retry copy, and a tab round-trip clears the alert (6 new tests; 53 total in the auth suites now).
- **Field errors on tab switch**: now cleared alongside server errors in both tab handlers — consistent treatment, red fields don't survive a round-trip.
- **"5 rAF callbacks" in reset-password**: false positive — each validation branch early-returns, so exactly one `focusFirstInvalidAfterRender` call ever runs per submit. The five call sites are the file's existing one-branch-per-rule style, not five scheduled callbacks.

marcodejongh:
Round-4 findings, both addressed:

- **Confirm-password toggle accessible name**: real — the register tab's two toggles were indistinguishable to screen readers. New `login.a11y.showConfirmPassword`/`hideConfirmPassword` keys in all three locales, confirm toggle wired to them.
- **`submitRegister` helpers now fill Name** so these tests keep exercising the server path if name validation ever tightens.

marcodejongh:
Round-5 ("ready to merge") notes:

1. **Retry test "passes trivially"** — false positive: the test passes `open` as a static `true` and `onClose` as a mock, so calling `onClose` never unmounts the Dialog (the parent owns `open`; nothing flips it). `queryByRole('alert')` runs against a still-mounted, still-open modal, and the assertion genuinely checks that the alert state cleared.
2. **rAF vs transitions** — fair; the constraint is now documented on `focusFirstInvalidAfterRender` (validation-error state must not be set inside `startTransition`). No auth form uses transitions.
3. **«bloque»/«bloc»** — deliberate: these boards' climbs are boulder problems, and the catalogs already use bloque/bloques and bloc/blocs for every climb count (`multiboardList.count`, `liked.addedToQueue`, mobile search counts). The new keys follow the established glossary; diverging to "vía" would be the inconsistency.

marcodejongh:
Round-6 nit applied: `vi.unstubAllGlobals()` moved into `afterEach` in both server-error test files, matching the verify-request pattern — a failed assertion can no longer leak the fetch stub into later tests.

marcodejongh:
Round-7 nit applied: forgot-password and reset-password now have thrown-fetch (network-down) tests alongside their API-failure tests — the `catch` path asserts the inline `Failed to request password reset` / `Failed to reset password` copy, no toast, no navigation.

Also: the PR is now retargeted to `main` and rebased clean on the merged form kit (#3703).

marcodejongh:
Final round nits:

- **Register field ids**: added — `register_name/email/password/confirm_password` via `FormField htmlFor`, matching the login tab's stable ids for password-manager heuristics.
- **`async` helpers with no `await`**: de-async'd (call sites cleaned too — a type-aware `await-thenable` pass would flag awaiting them).
- **Password toggle missing on the standalone auth page**: pre-existing and **deferred deliberately** — with the expo-web migration decided, we're not expanding web auth UI beyond this wave. If it turns out to matter before then, it's a small lift now that the a11y keys exist.
- **reset-password early-return style**: acknowledged as functionally correct by the review; leaving the file's one-branch-per-rule style as is.

marcodejongh:
Final review notes, dispositions (no further pushes — verdict has been ready-to-merge for several rounds and these are explicitly non-blocking):

1. **Shared `login_*` ids page vs modal**: acknowledged-hypothetical — the two surfaces never mount simultaneously in the app, and password managers key primarily off the `autoComplete` attributes this PR added; the fixed ids are a secondary stability aid. Deriving them from `useId` would trade a real heuristic for protection against a scenario that only exists in a hypothetical test.
2. **Color-picker `size`**: same false positive as round 5 — `MuiTextField.defaultProps.size: 'small'` (`mui-theme.ts:149`) means omitting the prop renders small, not medium. Nothing got taller.
3. **Close-and-reopen test**: `handleCancel` resets both server errors (plain state clearing, exercised implicitly by the reset in every reopen); leaving untested as the marginal value doesn't justify another CI round on a wave that's final before the expo-web pivot.

PR #3735, opened 2026-07-16
fix(web): set the theme before first paint — kill the dark flash for light-theme users

marcodejongh:
## Summary

Final piece of the (now truncated) Velvet Send web epic: light-theme users currently get a full dark flash on every first paint because the saved colour preference lives in IndexedDB, which can't be read before paint, and SSR hardcodes `data-theme="dark"`.

- **Pre-paint init script** (`theme-init-script.ts`, a tested string constant injected inline into `<head>` in the root layout and the global-error boundary): reads a localStorage mirror of the preference, falls back to `prefers-color-scheme` — so **new users now follow their OS theme** instead of always getting dark. Try/catch-wrapped for privacy-mode storage throws; garbage values fall back sanely.
- **localStorage mirror** (`boardsesh:colorMode`): written alongside IndexedDB on toggle; the mount reconcile refreshes it for users whose preference predates the mirror. IndexedDB stays the source of truth. This is the second sanctioned localStorage exception — CLAUDE.md's storage bullet is updated.
- **Provider**: syncs React state from the pre-paint attribute immediately on mount (no more waiting on the IndexedDB round-trip), then lets IndexedDB win. The MUI ThemeProvider double-render is unchanged and accepted — CSS-var surfaces (page, chrome) are what flash today and they're now correct pre-paint.
- Playwright pins `colorScheme: 'dark'` so the new OS-preference fallback stays deterministic in screenshot runs.

Built and adversarially reviewed by paired Opus lanes (verdict: approve-with-nits, both nits resolved — one inert lint suppression fixed; the global-error page stays visually dark by its own hardcoded styles, the script there is forward-looking).

## Release Notes

- If you use light mode, the app no longer flashes dark while it loads.
- New here? The app now matches your device's light/dark setting from the first screen.

## Test plan

- [x] 8 new init-script unit tests (saved light/dark, OS fallback both ways, storage throws, garbage value, no matchMedia) — the tests evaluate the actual shipped script string
- [x] 3 new provider tests (attribute sync, dual-store toggle write, IndexedDB-wins reconcile + mirror refresh)
- [x] `vp test run` theme + providers: 106/106; `vp run typecheck:web` (incl. full next build prerender); `vp check` clean
- [ ] CI green


marcodejongh:
Review round 3 addressed in `26dafbefa`:

1. **`<body suppressHydrationWarning>` restored** — you're right, and review round 2 (which called the removal "correctly narrowed") was wrong: the suppression pre-dates this PR on `main` and exists for extension-injected body attributes. Restored with the reason as a comment so it doesn't get "cleaned up" again.
2. **One-time flash for saved-light users on OS-dark devices** — real and unavoidable (nothing pre-paint can read IndexedDB): documented at the mirror-seeding effect. It's a single load per user; every load after the seed is flash-free.
3. **Global-error dark styles** — acknowledged TODO stands; follow-up territory.
4. **CSP** — the comment in `theme-init-script.ts` flags exactly this; wiring a nonce belongs with whichever change introduces a main-route CSP.

PR #3736, opened 2026-07-16
fix(web): wire onlyBenchmarks (Classics Only) through both search paths

### LINKED ISSUE #2320: onlyClassics filter is a no-op in DB layer
## Problem

`onlyClassics` is declared as a filter on `ClimbSearchParams` (`packages/db/src/queries/climbs/types.ts:46`) and accepted by the web's `SearchRequest` (`packages/web/app/lib/types.ts:116`), but **`packages/db/src/queries/climbs/create-climb-filters.ts` never references it**. The web search drawer toggle (`accordion-search-form.tsx:303-304`) updates the URL param but produces no SQL change — flipping it on still returns every climb.

Discovered while planning mobile filter parity (#2319 — `onlyClassics` was deferred from that PR for this reason).

## Scope

- Define what "classics" means as a SQL predicate (likely `ascensionist_count >= N` AND `quality_average >= M`, matching the Aurora app's behaviour — verify by inspection).
- Implement it in `createClimbFilters` alongside `onlyTallClimbs` / `onlyWideClimbs`.
- Add a test in `packages/db/src/queries/climbs/__tests__/` confirming the predicate filters as expected.
- Add `onlyClassics: Boolean` to `ClimbSearchInput` in `packages/shared-schema/src/schema/climb.ts` and thread it through `searchClimbs` resolver (`packages/backend/src/graphql/resolvers/climbs/queries.ts`).
- Once shipped, add a `SwitchRow` for it in `packages/mobile/src/components/ClimbFilterSheet.tsx` (Quality section) and add the field to `ClimbFilterState` in `@boardsesh/climb-filters`.

## Out of scope

- Behavioural change for existing web users — the toggle has always been a no-op, so enabling it on the backend is a feature unlock not a regression.

### LINKED ISSUE #2499: Web: "Only classics" filter is a no-op (onlyClassics/onlyBenchmarks never applied at the DB layer)
## Summary

The web **"Only classics"** toggle fires analytics and writes to the URL but **does not actually filter results**. Verified against source while planning the mobile search redesign.

- `onlyClassics` is written to the URL and read back (`packages/web/app/lib/url-utils.ts`) and is present in `packages/db/src/queries/climbs/types.ts` (`onlyClassics?: boolean`), **but there is no branch for it** in `packages/db/src/queries/climbs/create-climb-filters.ts` — it is accepted and silently ignored.
- The GraphQL/DB layer instead exposes `onlyBenchmarks` (`packages/shared-schema/src/schema/climb.ts`, `types/climb.ts`, `packages/backend/src/validation/schemas/climbs.ts`), but **zero web source files use it**, and `create-climb-filters.ts` has no benchmark branch either.

Net: in the PostHog data, "Only classics" shows ~5.5% of searches, but that reflects **toggle taps, not result-changing filtering**. The analytics signal for this filter is misleading.

## Impact

- Users who toggle "Only classics" get unchanged results.
- The 5.5% usage stat overstates demand for benchmark/quality filtering.

## Fix options

1. Implement `onlyBenchmarks` end-to-end: add a branch in `create-climb-filters.ts` (e.g. `benchmark_difficulty IS NOT NULL`), map it through `ClimbSearchInput`, and wire the web toggle to `onlyBenchmarks` (retire the dead `onlyClassics` path).
2. Or remove the dead toggle if benchmark filtering isn't wanted on web.

## Context

Found during the mobile climb-search redesign (PR #2498). Per product decision, mobile builds a working benchmark filter independently in a later phase (P1); this issue tracks fixing the web path. Mobile P1 will add the `create-climb-filters.ts` `onlyBenchmarks` branch, which web can then consume.

marcodejongh:
## Summary

`onlyClassics` was a dead filter on web — `create-climb-filters.ts` never branched on it, and the web toggle only wrote it to the URL. PR #2500 (mobile filter parity) already implemented the real predicate under a different name, `onlyBenchmarks` (`benchmark_difficulty > 0`), and wired it through the GraphQL schema, backend validation, and mobile's `@boardsesh/climb-filters` — but touched zero files under `packages/web`, so web was left pointing at the old dead field. That mobile PR closed #2499 (the narrower "web no-op" issue); this PR closes the original, broader #2320 by finishing the web side.

- Renamed `onlyClassics` → `onlyBenchmarks` end-to-end on web (type, URL param, UI switch, active-filter chip, heatmap cache key, PostHog `Climb Search Performed` property) to match the vocabulary already shipped everywhere else.
- Wired both web fetch paths, which previously dropped the filter entirely regardless of name:
  - SSR (`search-climbs.ts`): `mapSearchInputToParams` only recognizes `onlyBenchmarks`, so the field had to exist on `SearchRequestPagination` to reach the DB at all.
  - Client GraphQL (`use-queue-data-fetching.tsx`): neither `baseInput` (search) nor the separate `countInput` (count query) mentioned the filter — added it to both.
- Added `onlyBenchmarks` to `buildClimbSearchParamsJson`, the SSR `unstable_cache` key. This was missing for `onlyClassics` too, but harmless while the filter was inert. Now that it changes results, an unkeyed toggle would let a cached result from one filter state leak into the other — this is the collision fix that makes the rest of the PR safe to ship.
- No per-board gating/hiding: queried the dev DB (`board_climb_stats`) and confirmed `benchmark_difficulty` is populated for every board except `touchstone` (0 rows there today) — kilter 634, tension 2807, moonboard 1698, decoy 571, grasshopper 78, soill 3. Mobile already ships this toggle unconditionally on the same data; web now matches. `touchstone` just returns 0 results, same as any other legitimately-empty filter combination.

**Analytics-schema note for @marcodejongh's dashboards:** the PostHog `Climb Search Performed` event property renames from `onlyClassics` to `onlyBenchmarks` as part of this change (it was tracking dead toggle taps under the old name — see #2499's "~5.5% of usage is dead toggle taps" finding). Any saved insight/dashboard filtering on `onlyClassics` needs to switch to `onlyBenchmarks` going forward; historical events keep the old property name.

Fixes #2320

## Release Notes

- "Classics Only" now actually filters your search to curator-flagged benchmark climbs, instead of doing nothing.

## Test plan

- [x] `vp check` — clean (0 errors; pre-existing unrelated warnings in files this PR doesn't touch)
- [x] `vp run typecheck` — clean
- [x] `vp run check:i18n` / `vp run check:i18n:orphans` — clean (no string changes)
- [x] `vp test run --project web` (all touched files run in isolation, since the full 5392-test suite is flaky under this environment's parallelism regardless of this diff — verified by re-running 4 of the "failed" files solo against this branch, all green, and confirming `packages/db`'s 3 pre-existing native-test-runner failures reproduce identically on unmodified `main`):
  - `url-utils.test.ts`, `search-summary-utils.test.ts`, `search-climbs.test.ts` (new `buildClimbSearchParamsJson` regression test pinning the cache-key fix) — 273/273
  - `use-queue-data-fetching.test.tsx` (new test asserting `onlyBenchmarks` reaches the GraphQL `input`, covering both the search and count query builders) — 14/14
  - Remaining 11 renamed-fixture files — 319/319
- [x] `packages/db` `onlyBenchmarks` predicate suite (native `tsx --test`, added in #2500) — unaffected, still green
- Skipped manual dev-server QA: another active session runs the dev stack from a sibling worktree, and a second `vp run dev` risks port/DB collisions. Coverage instead comes from the DB predicate (already tested in #2500) plus the new url-utils round-trip and `use-queue-data-fetching` `baseInput`/`countInput` assertions above, which pin both wire paths end to end.

**Post-merge verification (2 steps):**
1. On prod, toggle "Classics Only" on a Kilter or Tension board search and confirm the result count drops to the curated subset (and reverts when un-toggled).
2. Check PostHog: new `Climb Search Performed` events should carry `onlyBenchmarks` instead of `onlyClassics`.


marcodejongh:
Should we not also fix this in Mobile?

PR #3749, opened 2026-07-17
fix(db): make merged MoonBoard climbs addressable by problem id (2024 logbook import)

alexpooley:
## What's wrong

Importing a MoonBoard logbook is a key lookup:

```
your tick: "I climbed problem 509834"  →  compute a climb UUID  →  find that row in board_climbs
```

Boardsesh keys climbs by UUID, not by problem id, so the importer computes the UUID from the id:

```js
uuidv5("moonboard:509834:40")
```

That only finds the row if the row was stored under a UUID built from the same string. For **2016** it mostly is, so 2016 imports. For **2024** it isn't — the computed key hits nothing, and every 2024 tick drops as "unknown problem".

## Why 2024 is stored under a different key

The 2024 climbs were loaded twice, from two sources with different identifiers:

1. An **older** importer's source had no problem id, so it built the UUID from name + setter + holds → `7fc49fa0-…`
2. The **current** catalog importer *does* have the problem id, but when it found the same climb already present (matched by holds) it **reused the old UUID** instead of inserting a duplicate — deliberately, to keep existing ticks, share URLs and favourites pointing at the same climb.

Result: the climb lives under `7fc49fa0-…`, and nothing is stored under `uuidv5("moonboard:509834:40")`. The importer's computed key points nowhere.

Example: `birthday cake trail mix` (id 509834) → its 40° row is `7fc49fa0-…` = `uuidv5("moonboard:3:40:BIRTHDAY CAKE TRAIL MIX|Dana Rader|<holds>")`, not `uuidv5("moonboard:509834:40")`.

## The fix

Boardsesh already has a redirect table — `board_climb_aliases` (`alias_uuid → canonical_uuid`) — and the tick resolver already follows it. The importer just never wrote the redirect for the reused-UUID case (only a useless self-redirect). This PR writes it:

```js
// inside the importer's existing dedup step
if (idBasedUuid !== reusedUuid) {
  aliases.insert({ alias_uuid: idBasedUuid, canonical_uuid: reusedUuid });
}
```

- **No re-keying** — existing UUIDs, ticks and links stay put.
- **No resolver change** — it already follows aliases.
- **No schema change.**
- **Idempotent**, and written inside the existing dedup step, so it's self-healing on every future import.

## Verification (production board snapshot)

Simulated the real resolver over a complete 260-tick MoonBoard logbook (181 on 2016, 79 on 2024) against the live `board-snapshots` dataset:

| board | before | after |
|---|---|---|
| 2016 (181) | 181/181 | 181/181 |
| 2024 (79) | **0/79** | **79/79** |
| total | 181/260 | **260/260** |

## Rollout / backfill

The fix only helps once the redirects exist in prod. The redirect for a 2024 climb can't be reconstructed from the DB alone (no problem id is stored on those rows), so it's filled by re-running the importer against the current catalog — it replays the same dedup and inserts every missing redirect (idempotent):

```
DB_URL=<prod> vp run db:import-moonboard-catalog /path/to/app-catalog
```

## Release Notes

- Your MoonBoard 2024 logbook now imports — sends on the 2024 board match up instead of getting dropped as unknown problems.

## Test plan

- `vp run typecheck:db`, `vp run typecheck:backend` — clean
- `packages/db/scripts/moonboard-catalog-helpers.test.ts` (new `catalogAliasRows` case) — 15/15 pass locally
- New backend DB test `resolves a legacy (name-based) 2024 climb via its moonboard:{id}:40 alias` — mirrors the existing alias-resolution test; runs in CI (Docker Postgres unavailable in my local env)

### DIFF
--- a/packages/backend/src/__tests__/moonboard-import-service.test.ts
+++ b/packages/backend/src/__tests__/moonboard-import-service.test.ts
@@ -74,6 +74,17 @@ async function insertMoonBoardAlias(problemId: number, canonicalUuid: string): P
   `);
 }
 
+// The 2024-board fix: the catalog importer merges each problem onto its
+// pre-existing legacy (name-based) climb and records an alias keyed off the
+// stable problem-id UUID `moonboard:{id}:40` (the resolver's angle-suffixed
+// candidate), so the climb is addressable by problem id.
+async function insertMoonBoardAngleAlias(problemId: number, canonicalUuid: string): Promise<void> {
+  await db.execute(sql`
+    INSERT INTO board_climb_aliases (board_type, alias_uuid, canonical_uuid, source)
+    VALUES ('moonboard', ${uuidv5(`moonboard:${problemId}:40`, UUID_NAMESPACE)}, ${canonicalUuid}, 'moonboard-catalog-import')
+  `);
+}
+
 async function importedTicks(): Promise<ImportedTickRow[]> {
   return (await db.execute(sql`
     SELECT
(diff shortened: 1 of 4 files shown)


marcodejongh:
Ah cool I need to run:

```
DB_URL=<prod> vp run db:import-moonboard-catalog /path/to/app-catalog
```

When it merges, can do

PR #3752, opened 2026-07-17
feat(mobile): enable Expo web Phase 0

marcodejongh:
## Summary

- Add an opt-in Expo web target for the mobile app at `/app`, with isolated web-only runtime dependencies, Next.js proxy boundaries, noindex/security headers, and CI export validation.
- Make the Android/mobile UI boot and function through React Native Web, including the Gorhom sheet compatibility layer, play/log-ascent/queue drawers, web controls, and WASM climb thumbnails.
- Harden shared web authentication, cross-tab logout/account switching, WebSocket auth, user-scoped browser storage, backend CORS, and reconnect behavior.
- Improve the development orchestrator so Expo web, Next.js, and the backend use distinct ports and terminate full child process trees.

Phase 0 go/no-go result: **GO**. Live QA covered credential login, a 510-climb list with WASM thumbnails, tab navigation, the play drawer, Log Ascent and Queue sheets, drag dismissal, and cross-origin HTTPS transport. Production Vercel publication remains intentionally gated for the later deployment phase.

## Release Notes

## Test plan

- `vp run typecheck:mobile`
- `vp exec tsc --noEmit --incremental false --project tsconfig.json` from `packages/web`
- `vp run test:mobile` (520 files, 4,173 tests)
- `vp test run --project web --reporter=agent` (443 files, 5,557 tests)
- `vp test run --project scripts --reporter=agent` (31 files, 392 tests)
- Backend auth/CORS tests with `SKIP_TEST_INFRA=1` (52 tests)
- `vp run check:mobile-platform-imports`
- `vp run check:i18n`
- `vp run check:i18n:orphans`
- `vp check` (0 errors; existing warnings only)
- `vp run check:mobile-ota-compat -- --base-dir /tmp/boardsesh-phase0-main` (iOS and Android OTA-compatible)
- HTTPS QA at `/app`, including security headers, auth, thumbnails, tabs, and all three drawer/sheet flows

## Review-fix addendum (overnight run, 2026-07-18)

An xhigh multi-agent review of this PR produced 15 verified findings (full report: `.boardsesh/pr-review-expo-web-phase0.md` in the worktree). All are now fixed in 18 grouped commits on this branch, and the branch is rebased onto main `930d3d77d`:

- **Deployed-client auth regressions (A1–A5)**: invalid optional credentials degrade to anonymous again on WS + HTTP GraphQL; ws-auth accepts pre-deploy cookies while `authSessionId` backfills; refresh failures split rejected (401/403 → sign out) vs unavailable (timeout/5xx → retry); completed responses resolve across an account switch. Each has a contract test.
- **Native OTA-neutrality (B1–B3)**: marker-fallback Sentry telemetry + cache-key coherence restored; `BottomSheetView` content wrapper gated to web (native sheets render the same Views as main); the realtime teardown invariant reconciled with the 4403 auth-retry (retained subscriptions replaced after JOIN_SESSION).
- **Web bugs (C1–C3)**: sign-out failures surface on web (Alert.alert is a no-op in RNW); board-overlay blob URLs no longer revoked while referenced; `ExpoAuthSessionBridge` mounts only under `BOARDSESH_WEB=1`.
- **Guards (D2, E2)**: web-runtime dependency-parity vitest guard; honest gorhom rationale in the imports guard; dev-server port-probe fallback breadth restored.
- Identity hygiene: per-user drafts cleared on sign-out (flagged for native QA); SignupCompleted analytics now reports the real platform flow.
- Rebase reconciliation: `FilterChipRow.web` ported to main's progress-filter contract (#3753).

Post-fix gates (all green): mobile 4214/4214, backend 2518/2518, web 5580/5580, both bundle checks, `vp check` 0 errors, `check:mobile-ota-compat` **ota-compatible** on iOS + Android vs main baseline.

**Merge notes**: merging publishes a production OTA. B1/B3 are behavior-improving re-lands riding that OTA — a native device pass (marker-fallback render, sheet heights, party reconnect after token refresh) is recommended first. E1 (fork dedup) is deliberately deferred to a stacked follow-up PR.


marcodejongh:
Thanks — addressed the follow-up round in the integrated branch (#3766, which carries this PR's commits plus the dedup that moved some of this code):

- **4403 duplicated constant** → extracted `AUTH_REJECTED_CLOSE_CODE`/`AUTH_REFRESH_RETRY_CLOSE_CODE` to a shared `ws-auth-close-codes.ts`, imported by both the transport remap and the subscription-retention detector (removes the silent-divergence risk).
- **`readExpectedSignOutIdentity` 400 on parse failure** → the FormData-parse `catch` now returns a distinct sentinel → **503 `signout_identity_unreadable`**, reserving 400 for structurally-invalid-but-parsed identity, so the client's retry logic won't treat a transient server error as a permanent mismatch. Test added.
- **`BrowserAuthWebSocket` addEventListener gap** → added delegating `addEventListener`/`removeEventListener` so a future graphql-ws that switches off `on*` handlers can't silently lose events.
- **secure-store sync stubs** → sync `setItem`/`deleteItem` now `console.error` under `__DEV__` naming the dropped key, so accidental sync-API use surfaces early.
- **docs** → added a WebSocket-auth-rejection-recovery section (4401→4403 remap rationale, the shim, the credential-generation guard, `authTransportRevision`) and fixed the stale key-point describing the removed recreate-client approach.

The earlier round (`extractAuthToken` `""`→`null` + `?token=` fallback restore, `next.config` URL guard + half-config warn, secure-store non-secret contract doc) is in this PR's commits directly. All green: typecheck (mobile+web), route/ws tests, and `check:mobile-ota-compat` stays **ota-compatible**.

marcodejongh (commented):

Multi-agent Claude review run by Marco across the whole Expo-web stack (#3752 / #3760 / #3761 / #3762 / #3764 / #3765 / #3766). Each PR was reviewed only on its own commits (diff vs its stack parent) by lens-scoped reviewers, and every finding below survived adversarial verification by independent refuter agents before posting. **7 findings** for this PR.

packages/web/app/api/auth/[...nextauth]/route.ts
```diff
@@ -1,6 +1,83 @@
...
+  const pathname = request.nextUrl.pathname.endsWith('/')
+    ? request.nextUrl.pathname.slice(0, -1)
+    : request.nextUrl.pathname;
+  if (!pathname.endsWith('/api/auth/signout')) return handler(request, context);
+
+  const expectedIdentity = await readExpectedSignOutIdentity(request);
+  // A standard NextAuth signOut() has no way to bind the request to the
+  // session shown in its initiating tab. Requiring identity fields prevents an
+  // old tab from deleting a cookie written by a newer login.
+  if (expectedIdentity === null) return guardedSignOutResponse(400, 'signout_identity_required');
```
Comment (marcodejongh) [inline packages/web/app/api/auth/[...nextauth]/route.ts:62]: **[medium]** Strict signout identity guard permanently breaks NextAuth's built-in signout page

The new `POST /api/auth/signout` guard returns `400 signout_identity_required` whenever `expectedUserId`/`expectedAuthSessionId` are absent — but one caller can never supply them: NextAuth's own built-in signout confirmation page. `authOptions.pages` configures `signIn`/`verifyRequest`/`error` but not `signOut`, and `export { handler as GET }` keeps the default HTML page live at `GET /api/auth/signout`. Its form POSTs only `csrfToken`, so clicking "Sign out" there now renders raw JSON `{"error":"signout_identity_required"}` and leaves the user signed in — a permanently broken, silently-failing flow at a well-known URL (unlike the legacy `signOut()` rejection, which the test suite shows is deliberate and self-heals after a reload). Concrete fix: set `pages.signOut` to a small page that calls `guardedNextAuthSignOut`, so the default confirmation page is never rendered.

packages/web/app/lib/auth/nextauth-cookie-fetch-lock.ts
```diff
@@ -0,0 +1,382 @@
+'use client';
+
+import { getSession } from 'next-auth/react';
+
+export const NEXTAUTH_COOKIE_LOCK_NAME = 'boardsesh-nextauth-cookie-v1';
```
marcodejongh, line 5: **[medium]** Cross-app auth contract pinned only by duplicated string literals; no parity test

The cross-app auth coordination contract this PR introduces is duplicated as independent string literals in web and mobile, with nothing enforcing they stay in sync:

- Web Lock name `'boardsesh-nextauth-cookie-v1'` — here (line 5) and in `packages/mobile/src/lib/auth-cookie-lock.web.ts` (line 1).
- BroadcastChannel name `'boardsesh-expo-web-auth-v1'` — here (line 13), in `packages/web/app/components/providers/session-provider.tsx` (line 8), and in `packages/mobile/src/lib/auth-store.web.ts` (line 65).
- Message shape (`type: 'auth-token-cleared'`, `sourceId`, `reason`, `identity`) — produced by `broadcastExpoAuthChange` here and by mobile's `broadcastSignOutStarted`, validated independently by `isExpoAuthClearMessage` (web) and mobile's own parser.
- Sign-out form fields `expectedUserId` / `expectedAuthSessionId` and the `signout_identity_changed` 409 error code — produced independently by `guardedNextAuthSignOut` (web) and `revokeOwnedNextAuthCookie` in `packages/mobile/src/lib/auth.web.ts`, consumed by `packages/web/app/api/auth/[...nextauth]/route.ts`.

Each side's tests hardcode their own copy of the literals, so they pin same-side consistency only: renaming the lock to `-v2` on one side fails that side's tests, the dev updates the local test literal, and every suite is green again while the peer app still uses `-v1` — cross-app response-order serialization and cross-tab session refresh silently stop working. That divergence would surface only as rare, unreproducible account-switch races in production, precisely the failure mode this lock exists to prevent.

This PR already establishes the cross-boundary-guard pattern (`packages/mobile/src/__tests__/web-runtime-dependency-parity.test.ts` reads another package's files). Concrete fix: hoist the lock name, channel name, message type, and sign-out field/error-code constants into a shared package both sides already import (e.g. `@boardsesh/shared-schema` — mobile cannot import from `packages/web` directly), or at minimum add a parity test that reads both sides and asserts the literals match.

packages/web/app/lib/auth/nextauth-cookie-fetch-lock.ts
```diff
@@ -0,0 +1,382 @@
...
+
+const LOCK_INSTALLATION_KEY = Symbol.for('boardsesh.nextauth-cookie-fetch-lock.state-v1');
+const NEXTAUTH_SESSION_PATH = '/api/auth/session';
+const NEXTAUTH_CSRF_PATH = '/api/auth/csrf';
+const NEXTAUTH_SIGNOUT_PATH = '/api/auth/signout';
+const NEXTAUTH_CALLBACK_PATH_PREFIX = '/api/auth/callback/';
+const NEXTAUTH_SIGNIN_PATH_PREFIX = '/api/auth/signin/';
+const EXPO_AUTH_CHANNEL_NAME = 'boardsesh-expo-web-auth-v1';
```
marcodejongh, line 13: **[medium]** Cross-app auth protocol duplicated across web and mobile with no shared source

The Expo-web/Next.js auth coordination protocol is a cross-package contract, but every element is re-declared as literals on each side with no shared source:

- Web Lock name `'boardsesh-nextauth-cookie-v1'`: `nextauth-cookie-fetch-lock.ts:5` and `packages/mobile/src/lib/auth-cookie-lock.web.ts:1` (the web file's own comment says "Expo web uses the same Web Lock name" — the agreement is the design, yet nothing enforces it).
- BroadcastChannel name `'boardsesh-expo-web-auth-v1'`: here (line 13), `packages/web/app/components/providers/session-provider.tsx:8`, and `packages/mobile/src/lib/auth-store.web.ts:65`.
- The message schema (`type: 'auth-token-cleared'`, `sourceId`, `reason`, `identity`) is typed as `AuthTokenClearedMessage` in mobile's `auth-store.web.ts` but hand-assembled as an untyped literal in `broadcastExpoAuthChange` here, and parsed by a third independent guard in `session-provider.tsx` — nothing type-checks the shapes against each other.
- The identity type is duplicated as `NextAuthSessionIdentity` (web) and `WebAuthIdentity` (mobile).
- The guarded sign-out sequence (GET `/api/auth/session` → verify identity → GET `/api/auth/csrf` → POST `/api/auth/signout` with `expectedUserId`/`expectedAuthSessionId` → 409 `signout_identity_changed` → validate returned URL) is implemented twice: `guardedNextAuthSignOut` here and `revokeOwnedNextAuthCookie` in `packages/mobile/src/lib/auth.web.ts`, including parallel no-Web-Locks fallback-queue logic. The copies already diverge: web treats any 409 as identity-changed, mobile requires `error === 'signout_identity_changed'`.

If either side renames a constant, adds a `reason`, or changes the form-field/error-code contract, nothing fails loudly — mobile's parser downgrades unknown reasons to `credential-rotation`, a renamed channel just never delivers, and the fallback paths mean the race protection silently stops working. This is the "extract, don't duplicate" case from CLAUDE.md.

Fix (minimum): extract the lock name, channel name, `AuthTokenClearedMessage`/identity types, and the signout form-field names + `signout_identity_changed` error code into a small pure-TS shared package (e.g. `@boardsesh/web-auth-protocol` under `packages/shared/`), imported by `packages/web` (including the `/api/auth/[...nextauth]` route), `session-provider.tsx`, and `packages/mobile`. Sharing the full guarded-signout sequence is a larger step — the two implementations differ behaviorally (mobile's generation fencing, web's redirect flow) — and can follow separately.

packages/mobile/src/lib/user-storage-owner.web.ts
```diff
@@ -0,0 +1,22 @@
...
+/**
+ * Resolve an account-specific key once, at operation start. Passing an explicit
+ * owner lets auth cleanup target the account that is leaving while another tab
+ * or a newer transition writes under a different owner.
+ */
+export function userScopedStorageKey(baseKey: string, owner?: UserStorageOwner | null): string | null {
+  const resolvedOwner = owner === undefined ? currentUserStorageOwner : owner;
+  if (!resolvedOwner) return null;
+  return (
+    `${baseKey}:user:${encodeURIComponent(resolvedOwner.userId)}` +
```
marcodejongh, line 92: **[low]** Login-session-scoped IndexedDB keys strand data forever and reset state on every re-login

`userScopedStorageKey` bakes `authSessionId` (a fresh `jti`/UUID per NextAuth login) into every persisted key used by `active-board-store.web`, `queue-snapshot-store.web` (up to 100 suggestion climbs per snapshot), `session-store.web`, and both draft stores. Two consequences:

1. **Unbounded orphan growth.** Cleanup only runs through the explicit sign-out path targeting the departing owner. If the user just closes the browser, or the 30-day session cookie expires, every row written under that `authSessionId` is stranded permanently — there is no key-prefix sweep, TTL, or GC, so `boardsesh-mobile-web` IndexedDB grows with every login for the life of the profile.
2. **State loss for the same user.** Logging out and back in (same `userId`) mints a new `authSessionId`, so the active board, queue snapshot, and drafts all come back empty even though nothing about the account changed — a regression vs. native, where these persist per device/user.

Scoping by `userId` alone would already prevent cross-account leakage (the stated goal); if per-login scoping is intentional for the drafts, at least add a startup sweep that deletes rows whose key carries the current user's `userId` but a stale `authSessionId`, so the same-user case both reclaims space and could migrate the active board forward.

packages/mobile/src/web-shims/bottom-sheet.tsx
```diff
@@ -0,0 +1,364 @@
...
+      },
+      collapse: () => {
+        markPresented();
+        innerRef.current?.collapse();
+      },
+      close: () => innerRef.current?.close(),
+      forceClose: () => innerRef.current?.forceClose(),
+      present: () => {
+        markPresented();
+        innerRef.current?.snapToIndex(indexProp ?? 0);
```
marcodejongh: **[low]** Web BottomSheet shim present() closes a sheet mounted with index={-1}

The web BottomSheet shim maps `present()` to `innerRef.current?.snapToIndex(indexProp ?? 0)`. `?? 0` only catches `undefined`/`null`, so for a sheet rendered with `index={-1}` — how the app's only non-modal consumer mounts it (`Sheet.tsx` renders `<BottomSheet index={-1} …>`, which resolves to this shim on web) — `present()` calls `snapToIndex(-1)`. In Gorhom 5.2.14 that resolves `detents[-1] === undefined` and silently fails to present (it no-ops or animates to an undefined position), whereas the native `@expo/ui/community/bottom-sheet` contract this shim mirrors presents regardless of the mount index.

Today this is latent: presentation goes through `useManagedSheet`, whose coordinator registration calls `sheetRef.current?.snapToIndex(desiredIndexRef.current)` (always ≥ 0) rather than the ref's `present()`, and the shim's new tests only cover the modal's `present()`. But the shim exports the full `BottomSheetMethods` contract, so any future direct `ref.present()` call (or a coordinator refactor to `registration.present()` semantics) silently fails to open the sheet on web only while native keeps working — a hard-to-diagnose divergence.

Fix: clamp the target index:
```ts
present: () => {
  markPresented();
  innerRef.current?.snapToIndex(Math.max(indexProp ?? 0, 0));
},
```

packages/mobile/src/web-shims/secure-store.ts
```diff
@@ -0,0 +1,55 @@
...
+  keychainAccessible?: string;
+  requireAuthentication?: boolean;
+  authenticationPrompt?: string;
+  keychainService?: string;
+};
+
+export const AFTER_FIRST_UNLOCK = 'AFTER_FIRST_UNLOCK';
+
+const KEY_PREFIX = 'secure:';
+
```
marcodejongh, line 26: **[low]** SecureStore web shim silently downgrades keychain writes to plaintext IndexedDB

This shim transparently maps every `expo-secure-store` call to plain IndexedDB (via the AsyncStorage shim), under a `secure:` key prefix that implies a protection it doesn't provide. Today that's safe only by convention: the genuinely sensitive SecureStore writers (`auth-store.ts`) have `.web.ts` forks that keep tokens memory-only, so the callers that reach this shim are preference/session-id stores. But nothing enforces that invariant — a future native code path that writes a real secret through SecureStore (the native mental model is "OS keychain") will silently persist it to IndexedDB on web, where it's readable by any same-origin XSS and survives on shared computers. This repo already builds guards for exactly this class of drift (web-runtime dependency parity test, gorhom import check). Concrete fix: add a reviewed allowlist of key prefixes to the shim and `reportHandledError`/`__DEV__`-throw on writes outside it (or at minimum a prominent comment stating that no secret material may flow through this module and that new SecureStore writers need a `.web.ts` fork), so the downgrade is a visible decision instead of a silent default.

packages/web/middleware.ts
```diff
@@ -35,6 +39,45 @@ export function middleware(request: NextRequest) {
...
+    const expoWebResponse = NextResponse.next({ request: { headers: expoWebRequestHeaders } });
+    // External rewrites forward the Expo response headers and can bypass the
+    // matching next.config header rule. Attach noindex in middleware as well so
+    // the actual proxied response remains a utility surface in development and
+    // in any future same-origin deployment.
+    expoWebResponse.headers.set('X-Robots-Tag', 'noindex, follow');
+    // External rewrites can also replace next.config's global response headers.
+    // Keep the authenticated SPA frame-protected at the middleware boundary;
+    // Metro repeats these headers so direct development requests match.
+    expoWebResponse.headers.set('X-Frame-Options', 'SAMEORIGIN');
```
marcodejongh, line 69: **[low]** /app security headers triplicated with no parity guard; middleware copy wins

The /app response-header set is now maintained in three places with no parity guard: (1) `next.config.mjs` — the new `/app/:path*` rule (`X-Robots-Tag`) plus the pre-existing global security-header rules (`X-Frame-Options`, `X-Content-Type-Options`, `Strict-Transport-Security`, `Referrer-Policy`); (2) this middleware block (lines 54-61), whose copies win on proxied Expo responses because external rewrites can bypass the next.config rules; (3) `packages/mobile/expo-web-response-headers.cjs` (`EXPO_WEB_SECURITY_HEADERS`), applied by Metro for direct dev requests. Each copy has its own test (`middleware.test.ts` asserts literals, `expo-web-response-headers.test.ts` asserts against the module's own exported constant, `next-config-expo-web.test.ts` asserts the X-Robots-Tag literal), so all suites stay green if the copies diverge — e.g. tightening HSTS or Referrer-Policy in next.config only would leave /app responses served via middleware or Metro on stale values with no failing test. The middleware comment documents the intent ("Metro repeats these headers so direct development requests match") but nothing enforces it. Suggested fix: define the header map once in a small shared module imported by the middleware and next.config, and cover the mobile `.cjs` with a parity test that compares it to the web constant — the same pattern this PR already uses in `web-runtime-dependency-parity.test.ts`.

packages/web/app/api/auth/[...nextauth]/route.ts
Comment (marcodejongh) [reply packages/web/app/api/auth/[...nextauth]/route.ts:62]: Fixed in 42217d96c: set `authOptions.pages.signOut = '/auth/signout'`, a small client page that drives the same `guardedNextAuthSignOut` sequence the account menu uses. NextAuth now redirects `GET /api/auth/signout` to that page instead of rendering its built-in confirmation form, so the CSRF-token-only POST that produced the 400 `signout_identity_required` JSON never happens. The deliberate 400 for unscoped legacy sign-outs (route.test.ts) is unchanged.

packages/web/app/lib/auth/nextauth-cookie-fetch-lock.ts
marcodejongh: Fixed in ee548bd8b: added `packages/mobile/src/__tests__/cross-app-auth-protocol-parity.test.ts`, which reads both packages and asserts the shared literals stay identical — the Web Lock name (`NEXTAUTH_COOKIE_LOCK_NAME` vs `AUTH_COOKIE_LOCK_NAME`), the BroadcastChannel name across all three producers/consumers, the `nextauth.message` storage key, the `auth-token-cleared` message type + reason values, and the `expectedUserId`/`expectedAuthSessionId` fields + `signout_identity_changed` 409 code. Renaming a constant on one side now fails this guard. We went with the parity test (the reviewer-sanctioned minimum) rather than a full shared-package extraction to avoid new cross-bundler (Next/Turbopack + Metro) wiring risk in this Phase 0 recovery PR; the extraction can follow separately.

packages/web/app/lib/auth/nextauth-cookie-fetch-lock.ts
marcodejongh: Fixed in ee548bd8b (see the sibling thread): `cross-app-auth-protocol-parity.test.ts` pins the lock name, channel name, storage-event key, message type/reason values, and the sign-out form fields + `signout_identity_changed` code across web and mobile so a rename or contract change on either side fails loudly. Sharing the full guarded-signout sequence and the message/identity types is deliberately left out for now — the two implementations differ behaviorally (mobile generation fencing vs web redirect flow) — and can follow as its own change.

packages/mobile/src/lib/user-storage-owner.web.ts
marcodejongh: Fixed in 113e47d49: `setCurrentUserStorageOwner` now runs `sweepOrphanedUserStorage` on a settled login (deduped per owner), deleting every IndexedDB row that carries the current user's id under a stale `authSessionId` via `removePreferencesMatching`. That bounds growth to one live login per user and reclaims the same-user re-login case; rows for other users are left to the explicit sign-out cleanup path. Covered by a new test in user-scoped-core-storage.web.test.ts (reclaims stale-login rows, preserves the current login and other users).

packages/mobile/src/web-shims/bottom-sheet.tsx
marcodejongh: Fixed in 0e240c468: `present()` now calls `snapToIndex(Math.max(indexProp ?? 0, 0))`, so a sheet mounted at `index={-1}` presents at the first detent instead of resolving `detents[-1]`. Added a bottom-sheet.test.tsx case asserting a non-modal sheet mounted at `index={-1}` forwards `snapToIndex(0)` on `present()`.

packages/mobile/src/web-shims/secure-store.ts
marcodejongh: Fixed in 0e240c468: added a runtime guard to the shim. A key that looks like secret material (`token`/`password`/`secret`/`credential`/`jwe`/`apikey`/`api_key`/`privatekey`) now `console.error`s and throws in development on write, so a future native SecureStore writer of a real secret trips here instead of silently persisting cleartext IndexedDB on web. The denylist deliberately omits `session`/`auth` (legitimate non-secret session-id/preference stores use those). Kept dependency-free (no Sentry import) so the web bundle and shim tests stay light; covered by new storage.test.ts cases.

packages/web/middleware.ts
marcodejongh: Fixed in 5124e6e54: added `packages/web/app/__tests__/expo-web-header-parity.test.ts`, which ties all three /app header sources to the canonical Metro `.cjs` constant — it invokes the middleware on `/app` and asserts every header matches `EXPO_WEB_SECURITY_HEADERS`/`EXPO_WEB_ROBOTS_VALUE`, and checks next.config's `/app/:path*` and global security-header rules against the same constant. Tightening HSTS/Referrer-Policy in only one place now fails here.

packages/mobile/src/web-shims/secure-store.ts
marcodejongh: Follow-up in f7d2a7b86: switched the dev gate from `process.env.NODE_ENV` to the mobile-repo `__DEV__` convention (typeof-guarded; the mobile vite config defines `__DEV__: 'true'` so the shim test still throws).

packages/web/app/api/auth/[...nextauth]/route.ts
Comment (marcodejongh) [reply packages/web/app/api/auth/[...nextauth]/route.ts:62]: Follow-up in f7d2a7b86: rewrote the new French signout strings as vouvoiement to match the rest of the auth catalog.

marcodejongh:
Addressed the claude[bot] review items (head now `4abc39fd9`):

1. **Double ws-auth cookie decrypt** → `9f67ed089`. It was already one decrypt (`getToken(raw:true)` returns the raw cookie without decrypting), but two cookie parses. Restructured to a single `getToken(raw:true)` + one `decode()`. Verified `decode({token, secret})` reproduces `getToken(raw:false)` exactly (same saltless HKDF derivation in next-auth 4.24.14), so semantics are identical — a malformed/expired cookie decodes to `null` → anonymous 200, not a 500. Test updated.
2. **Docs — Expo-web WS token path** → `4abc39fd9`. Added a subsection to `websocket-implementation.md` (poll `/api/internal/ws-auth` → BroadcastChannel propagation → `BrowserAuthWebSocket` 4401→4403 refresh-and-retry).
3. **Parity extractor quote styles** → `272db4af6`. `extractConst` now matches both `'` and `"` and throws a named error instead of returning `null`.
4. **safeCallbackUrl open-redirect tests** → `bf20624d6`. Extracted to a pure module with unit tests covering `//host`, absolute, `javascript:`/`data:`/`mailto:`, empty/null, and plain-path cases — and hardened against the `/\host` backslash variant browsers normalise to `//`.
5. **DEV_ALLOWED_ORIGINS sanitization** → `7e95619b8`. Each entry must parse as an http/https URL or it is dropped with a warning; valid ones are normalized to their bare origin. Ran the cors test file (50 pass).
6. **window.alert sign-out fallback** — SKIP per the bot (acceptable). Left as a deliberate follow-up; the sign-out failure path still surfaces via the in-page error state, and a nicer toast can replace the `alert` later.

Validation: web 12, mobile 5, backend 50 targeted tests pass; `vp check --fix` clean (no lint/type errors) across all 8 files.

marcodejongh:
Addressed the three items from the follow-up claude[bot] review (head now `b719e017e`):

1. **`session-provider.tsx` render-body side effect** → `2a81dd0b1`. Moved `installNextAuthCookieFetchLock()` out of the render body into a client layout effect (`useIsomorphicLayoutEffect`). A parent layout effect still commits before the child `SessionProvider`'s passive session-read effect, so the fetch lock is in place before NextAuth's first cookie fetch — without mutating `globalThis.fetch` during render. Chose the effect over a top-level module install deliberately: module-init would wrap `globalThis.fetch` for the **entire** web app, but the lock is gated on `enableExpoAuthBridge` (Expo `/app` only), and the existing `session-provider.test.tsx` still passes.
2. **Duplicate `AUTH_REFRESH_RETRY_CLOSE_CODE = 4403`** → `b719e017e`. Exported it from both `ws-client` platform variants (`.ts` + `.web.ts`, since the consumer's import is platform-resolved) and imported it in `use-session-realtime.ts`, dropping the re-declared literal. **Parity test:** left unchanged — 4403 is a mobile-internal transport close code (producer `ws-client` ↔ consumer `use-session-realtime`), not a web↔mobile shared literal, so it does not belong in `cross-app-auth-protocol-parity.test.ts` (which pins only constants shared between `packages/web` and `packages/mobile`).
3. **`NEXTAUTH_SECRET ?? ''` in ws-auth** → `1adbe0748`. Detect the unset secret explicitly, `console.warn` server-side about the misconfiguration, and keep the fail-closed anonymous response (no decode attempted without the key). New test covers it.

Validation: web 12, mobile 22 targeted tests pass (`--pool=threads`); `vp check --fix` clean across all 6 files. (The `vote-button.test.tsx` red on the branch is the known flake, already re-run.)

marcodejongh:
CI fix for the auth-refresh rejoin failures in `queue-provider-session-updates.test.tsx` (head now `5bbf7152a`).

**Cause:** my item-2 change exported `AUTH_REFRESH_RETRY_CLOSE_CODE` from `ws-client` and imported it in `use-session-realtime`, but the queue-provider tests `vi.mock('../../lib/graphql/ws-client')` with a bare `{ getWsClient }` factory. The new import resolved `undefined`, so the `=== 4403` rejoin comparison never matched and the path degraded to a plain reconnect. ~8 other test files mock `ws-client` the same way — a latent landmine.

**Fix (import structure, single-source preserved):** moved both auth close codes (4401/4403) into a dedicated dependency-free module `packages/mobile/src/lib/graphql/ws-close-codes.ts`, imported by both `ws-client` platform variants and `use-session-realtime`. The constant now lives outside the mocked module, so no test can strip it and none needs to re-declare it. Reverted the one-line mock patch — not needed with this structure.

Validation: `queue-provider-session-updates` + `ws-client.test` + `ws-client.web.test` + `cross-app-auth-protocol-parity` → 4 files, 60 tests pass (`--pool=threads`); `vp check --fix` clean.

marcodejongh:
Addressed the 4th claude[bot] review batch (head now `434386a84`).

**Mediums (all fixed):**
1. **cors.ts hardcoded Vercel suffix** → `3d125a16d`. `VERCEL_PREVIEW_ORIGIN_SUFFIX` env, default unchanged, recomputed in `initCors`; cors test covers override + rejection (51 pass).
2. **auth-interceptor second-401 always signs out** → `56ac9e548`. Now re-confirms via `synchronizeWebSession` (checks cookie **and** ws-auth bridge token) and only signs out on a confirmed-anonymous result — a revoked token still trips it, a permission-401 does not. Both branches tested.
3. **user-storage-owner dedup never reset on null** → `53c3c14ec`. `lastSweptOwnerScope` reset on `setCurrentUserStorageOwner(null)` so sign-out→sign-in sweeps again; retention test extended.

**Lows (fixed):**
4. **session-update error handler no-op** → `39faee5e1`. Wired `reportHandledError`.
5. **CLIMB_SESSION_COOKIE not Secure** → `a294f486a`. `secure: isSecureCookieContext()` in middleware (still round-trips in http dev; middleware test 75 pass).
6. **auth-store partial write** → `da4ed659a`. Reordered to write the JWT **last** (commit marker) so a partial write can only drop the JWT (clean unauthenticated), never strand it without its refresh token. I did **not** move the generation bump after the writes — it is load-bearing: bumping first claims the new sign-in as a distinct credential owner so a concurrent sign-out (`clearTokensForGeneration`) fences off instead of clobbering the fresh tokens. Order-assertion test added.

**Deferred (no code):**
- **window.alert sign-out fallback** — confirmed deferred (already noted in a prior round as a follow-up; the failure still surfaces via the in-page error state).
- **create-climb-draft native owner param** — agreeing with the bot: intentionally deferred for OTA safety; not touching it here.

**Docs:** `websocket-implementation.md` already described the 4401→4403 flow accurately; added a one-line note that both codes live in the shared `ws-close-codes.ts` (`434386a84`).

Validation: backend cors 51, mobile 25 (user-scoped-core-storage + auth-store + auth-interceptor.web), web middleware 75 — all pass (`--pool=threads`); `vp check --fix` clean across all 10 files.

marcodejongh:
Batch 5 (head now `e112e9f6c`).

**Fixed:**
1. **Native transient refresh outage → false sign-out** (`auth-session.ts`) → `9cae90e9d`. `resolveAuthSession` refreshed an expiring token via the boolean `ensureFreshToken`, which returns `false` for **both** a server-rejected token and an offline refresh — so a network blip at expiry mapped a logged-in native user to `anonymous` and forced sign-out. Now refreshes through `deduplicatedRefresh` and maps its status: `rejected → anonymous` (real logout), `unavailable → { status: "unavailable" }`. The one caller (`auth-provider.tsx:343`) already handles `unavailable` by releasing the UI and retrying **without** confirmed-sign-out cleanup, so no crash and no sign-out on the transient path. Tests added: rejected → anonymous, unavailable → not anonymous, refreshed → authenticated.
2. **`expo-web-response-headers.cjs:13` redundant guards** → `8d5e00970`. Dropped the `?.`/`=== true` (rawUrl is already non-empty and `split` always yields index 0).
3. **Docs** → `e112e9f6c`. Added `docs/expo-web.md` with Phase 0 limitations.

**Reply-only (no code):**
- **Main-thread WASM render** — superseded by the off-thread render worker in **#3765** (same stack); noted in `docs/expo-web.md` as fixed there.
- **CSP `unsafe-eval` for WASM glue** — covered by the new docs note (the `/app` surface depends on `Function()`/`unsafe-eval`; a future strict `script-src` must account for it).

Validation: mobile 23 (auth-session + auth-session-race + expo-web-response-headers) pass (`--pool=threads`); `vp check --fix` clean.

marcodejongh:
CI fix for the two `auth-provider.test.tsx` checkAuth failures (head now `66179d627`).

Cause: `9cae90e9d` moved `resolveAuthSession`'s refresh from the boolean `ensureFreshToken` to the status-returning `deduplicatedRefresh`. That test file uses `importOriginal` for `resolveAuthSession` (so the real mapping runs) but its `auth-interceptor` mock only stubbed `ensureFreshToken` — `deduplicatedRefresh` resolved undefined, so "refresh fails" no longer produced `rejected → anonymous → cleanup`.

Fix (mock seam only, product behavior identical): the mock now exposes `deduplicatedRefresh` with the 4-way status. "refresh fails" drives `{ status: 'rejected' }` → anonymous → full cross-user cleanup; the superseded test stalls `deduplicatedRefresh` while the generation flips → superseded → no cleanup of the newer login.

Validation: full `auth-provider.test.tsx` + `auth-session.test` + `auth-session-race` → 3 files, 47 tests pass (`--pool=threads`); `vp check --fix` clean.

marcodejongh:
Final round (head now `4070eb2f3`). Fixed the cheap real subset, filed the architectural items as issues, refuted the rest with evidence.

**Fixed:**
1. **Dev proxy origin** → `846aa7c4d`. `resolveExpoWebDevOrigin` now rejects any non-loopback host (`localhost`/`127.0.0.1`/`[::1]`) — the orchestrator only ever sets `http://localhost:PORT` (`scripts/lib/dev-server-origins.ts`), so this is safe and stops a stray value becoming an open forward. *Sub-claim refuted:* the `/assets` + `/assets/**` rewrites are **already** proxy-scoped — they live inside `rewrites()`, which returns `[]` unless `BOARDSESH_WEB=1` **and** a valid origin resolve. No change needed there.
2. **/app carve-out gating** → `d91f36506`. Gated the middleware `/app` carve-out on `BOARDSESH_WEB=1` ("Expo web enabled at all"), **not** the dev-proxy check — so a static-serve deployment (#3761) still gets it, but a deployment without Expo web leaves `/app` to normal routing.
3. **VERCEL_PREVIEW_ORIGIN_SUFFIX** → `5255ec624`. Must be a real `<label>.vercel.app` host; a bare `vercel.app` (which would allow every Vercel deployment) is warned + falls back to the default.
4. **HSTS** → `d91f36506`. `/app` HSTS is now secure-context-only (matching the cookie Secure gate); Metro (`.cjs`) no longer sends it at all over http dev. Shared value moved to `EXPO_WEB_STRICT_TRANSPORT_SECURITY`; the parity test pins it separately.
5a. **Parity `toContain`** → `4070eb2f3`. Now quote-agnostic (matches either quote style).

**Refuted (no change):**
- **jti rotation breaking the signout identity guard** — the guard uses `authSessionId`, minted **once** via `if (!token.authSessionId) token.authSessionId = token.jti ?? randomUUID()` in the jwt callback (`auth-options.ts:277`). It is never regenerated on later reissues, so it is stable across cookie rotations regardless of jti — the concern does not apply.
- **header-parity `NextRequest` may not run in vitest** — it runs and passes (this round: 86 web tests green, header-parity among them; also green in the prior CI run). No change.
- **web-runtime `react-native` → devDependency** — skipped per "skip if unsure": on web `react-native` resolves via `react-native-web` and the version is pinned for alignment; moving it would break `web-runtime-dependency-parity.test.ts` (reads `.dependencies`) and risks resolution. Left as a dependency.

**Filed as issues (deferred architectural):** #3772 (opaque short-lived WS ticket instead of raw JWE from ws-auth), #3774 (CSP for the /app surface, incl. the unsafe-eval/WASM context), #3773 (document/harden the identity-scoped signout CSRF tradeoff).

Validation: web 86, backend 52, mobile 16 targeted tests pass (`--pool=threads`); `vp check --fix` clean across all 10 files.

marcodejongh:
## Claude: ready to merge ✅

All review findings on this PR are resolved at head `4070eb2f3`: the multi-agent swarm review, follow-up Codex and claude[bot] batches, and QA verification passes are addressed (fixed, refuted with evidence, or filed as follow-up issues). CI is green with no unresolved threads.

marcodejongh:

marcodejongh:
Thanks — triaged this round:

- **`AsyncStorage.removeMany` on web** — verified a non-issue. The web shim (`src/web-shims/async-storage.ts`) uses the v3 `createAsyncStorage('boardsesh-mobile-web')` IndexedDB implementation, whose factory defines `removeMany`, and the `AsyncStorage` type declares it (so `typecheck:mobile` would fail if it were missing). Per-user preference cleanup on sign-out works on web.
- **unsafe-eval CSP carve-out / cross-tab auth protocol docs** — tracked as documentation/infra follow-ups; the CSP hardening for the `/app` (and now `app.boardsesh.com`) surface is an ops item captured in the deploy notes, not a code change in this PR.
- **`isSameOriginUpgrade` scheme comparison** — acknowledged; as noted it's acceptable under the current JWT-not-cookie WebSocket auth (the WS presents a bearer token via `connectionParams`, not a cookie), so an Origin-only match doesn't grant a cookie-authed session.

No code changes needed for this round. DB check for merge-readiness: **the stack touches zero migrations / `_journal.json`**, so no rebase-renumber is required when it lands on main.

PR #3760, opened 2026-07-17
test(web): expo-web smoke Playwright project

marcodejongh:
## Summary

Adds an `expo-web-smoke` Playwright project that drives the mobile app compiled for the browser (the `/app` surface, react-native-web + the Material variant) through the Phase 0 go/no-go flows. It's the manually-run regression check for the Expo-web stack — run it locally before landing a later Expo-web PR, or as the scripted half of morning QA. It is **not** wired into PR CI: the stack it needs (backend + Next proxy + Metro web, with a multi-minute cold Metro web compile) is too heavy for the per-PR e2e shards, so this ships as a local/on-demand task. Wiring a nightly or label-gated CI job is a follow-up.

Six serial specs:
1. Boots, signs in with credentials, renders the five-tab Material shell (Home · Climbs · Record · Discover · Profile).
2. Home feed renders WASM board thumbnails as `blob:` object URLs; Climbs tab shows the follow-your-wall empty state for an unbound context.
3. Opens the play drawer from a bound board's climb row and reaches the log-ascent sheet (Attempt/Flash save row).
4. Queue sheet opens (gorhom web shim, `BottomSheetFlatList`) and dismisses.
5. Deep route reloads through the `/app` proxy without losing the signed-in shell.
6. `/app` responses carry the `X-Robots-Tag: noindex` header through the proxy.

Selectors query by ARIA role/name (react-native-web maps `accessibilityRole`/`Label` to ARIA) rather than generated RNW class names. A console/page-error budget fails the run on React/Reanimated/hydration/auth-loop errors.

## How it runs

Locally only — this project is not part of PR CI (see above):

- `vp run test:e2e:expo-web` — `scripts/expo-web-e2e.ts` boots the full expo-web stack via the dev orchestrator (backend + Next proxy + Metro web), waits for `/app` to serve the Expo shell (`/_expo` bundle path), runs the project against the printed Next origin, tears the stack down.
- Or `vp run dev:mobile:web`, then `bunx playwright test --project=expo-web-smoke` from `packages/web`.
- The classic e2e `global-setup` is skipped under `PLAYWRIGHT_SKIP_CLASSIC_SETUP=1` (the smoke drives `/app`, not the classic routes it seeds); the Playwright `webServer` auto-start is disabled when `PLAYWRIGHT_TEST_BASE_URL` is set so it can't race the orchestrator's Next for port 3000.

## Release Notes

- [x] No release note needed (internal / testing change)

## Test plan

- 6/6 specs green against the live expo-web stack (31.5s warm; ~2–3 min cold with the on-demand Metro bundle compile), zero flakes after hardening the sheet-dismissal poll.
- `vp check` clean on all changed files.

_Stacked on #3752 (Expo web Phase 0). Review after #3752._

### DIFF
--- a/packages/web/e2e/expo-web/smoke.spec.ts
+++ b/packages/web/e2e/expo-web/smoke.spec.ts
@@ -1,0 +1,223 @@
+// Expo-web smoke: drives the mobile app compiled for the browser (the /app
+// surface, react-native-web + the Material variant) through the Phase 0
+// go/no-go flows. Runs only in the `expo-web-smoke` Playwright project, which
+// requires the full expo-web dev stack (backend + Next proxy + Metro web) —
+// start it with `vp run test:e2e:expo-web`, or `vp run dev:mobile:web` and run
+// the project manually. The chromium project ignores this directory.
+//
+// Selector strategy: react-native-web maps accessibilityRole/Label to ARIA, so
+// everything here queries by role/name through the app's i18n strings (en-US
+// catalog — the dev stack runs untranslated). No CSS classes: RNW class names
+// are generated and unstable.
+//
+// Serial: the first /app load compiles the whole Metro web bundle on demand
+// (minutes, once per stack). Parallel workers would each pay that cost and
+// time out; serial lets the first test absorb the compile and the rest run
+// against a warm bundle.
+
+import { expect, test, type Page } from '@playwright/test';
+
+test.describe.configure({ mode: 'serial' });
+
+const TEST_EMAIL = process.env.TEST_USER_EMAIL ?? 'test@boardsesh.com';
+const TEST_PASSWORD = process.env.TEST_USER_PASSWORD ?? 'test';
+
+/** First navigation may include the on-demand Metro compile of the bundle. */
+const COLD_LOAD_TIMEOUT_MS = 180_000;
+const WARM_TIMEOUT_MS = 30_000;
+
+/** The five Material tab-bar entries, in mobile's canonical order. */
+const TAB_NAMES = ['Home', 'Climbs', 'Record', 'Discover', 'Profile'] as const;
+
+/**
+ * Console/page-error budget: the smoke fails on error classes that historically
+ * meant a broken surface rather than noise.
+ */
+const FATAL_CONSOLE_PATTERNS = [
+  /Maximum update depth exceeded/i,
+  /Cannot update a component/i,
+  /Reanimated/i,
+  /ws-auth returned no token/i,
+  /Unable to get view config/i,
+  /No QueryClient set/i,
+];
+
+function collectFatalConsoleErrors(page: Page): string[] {
+  const fatalErrors: string[] = [];
+  page.on('console', (message) => {
+    if (message.type() !== 'error') return;
+    const text = message.text();
+    if (FATAL_CONSOLE_PATTERNS.some((pattern) => pattern.test(text))) {
+      fatalErrors.push(text);
+    }
+  });
+  page.on('pageerror', (error) => {
+    fatalErrors.push(String(error));
+  });
+  return fatalErrors;
+}
+
+function tabButton(page: Page, name: (typeof TAB_NAMES)[number]) {
+  // The Material tab bar exposes role=tab; accessible names carry the icon
+  // glyph before the label ("󰋜 Home"), so match the label as a word.
+  return page.getByRole('tab', { name: new RegExp(`\\b${name}\\b`) }).last();
+}
+
+/**
+ * Lands on /app and ends signed in on the tab shell, tolerating either entry
+ * state: a fresh context gets the credentials screen; a context that inherited
+ * a NextAuth session through the SSO bridge lands straight on Home.
+ */
+async function ensureSignedIn(page: Page): Promise<void> {
+  await page.goto('/app', { timeout: COLD_LOAD_TIMEOUT_MS });
+  const emailField = page.getByRole('textbox', { name: 'Email', exact: true });
+  const homeTab = tabButton(page, 'Home');
+  await expect(emailField.or(homeTab).first()).toBeVisible({ timeout: COLD_LOAD_TIMEOUT_MS });
+
+  if (await emailField.isVisible()) {
+    await emailField.fill(TEST_EMAIL);
+    await page.getByRole('textbox', { name: 'Password', exact: true }).fill(TEST_PASSWORD);
+    await page.getByRole('button', { name: 'Sign in' }).click();
+  }
+  await expect(homeTab).toBeVisible({ timeout: WARM_TIMEOUT_MS });
+}
+
+test.describe('expo-web smoke', () => {
+  test('boots, signs in, and renders the five-tab Material shell', async ({ page }) => {
+    const fatalErrors = collectFatalConsoleErrors(page);
+    await ensureSignedIn(page);
+    for (const tabName of TAB_NAMES) {
+      await expect(tabButton(page, tabName)).toBeVisible();
+    }
+    expect(fatalErrors).toEqual([]);
+  });
+
+  test('home feed renders WASM board thumbnails; climbs tab shows the follow-your-wall state', async ({ page }) => {
+    const fatalErrors = collectFatalConsoleErrors(page);
+    await ensureSignedIn(page);
+    // WASM overlay renders resolve to blob: object URLs inside expo-image
+    // <img>s — the Home feed's session cards carry them without any board
+    // context.
+    const blobThumbnails = page.locator('img[src^="blob:"]');
+    await expect(blobThumbnails.first()).toBeVisible({ timeout: 60_000 });
+    expect(await blobThumbnails.count()).toBeGreaterThanOrEqual(3);
+    // A fresh browser context has no followed board, so the Climbs tab shows
+    // the follow-your-wall empty state rather than a list.
+    await tabButton(page, 'Climbs').click();
+    await expect(page.getByRole('button', { name: 'Find my board' })).toBeVisible({ timeout: WARM_TIMEOUT_MS });
+    expect(fatalErrors).toEqual([]);
+  });
+
+  /**
+   * Binds the seeded kilter board (Dyno Den) via the board sheet so the Climbs
+   * tab has a list and play-drawer actions aren't blocked by the switch-board
+   * overlay (feed climbs live on other boards). Fresh contexts start unbound.
+   */
+  async function bindSeededBoard(page: Page): Promise<void> {
+    await tabButton(page, 'Climbs').click();
+    const findBoardButton = page.getByRole('button', { name: 'Find my board' });
+    if (await findBoardButton.isVisible().catch(() => false)) {
+      await findBoardButton.click({ force: true });
+      await page
+        .getByRole('button', { name: /Dyno Den/ })
+        .first()
+        .click({ force: true });
+    }
+    // Bound board → the list renders WASM thumbnails.
+    await expect(page.locator('img[src^="blob:"]').first()).toBeVisible({ timeout: 60_000 });
+  }
+
+  /** Opens the first climbs-list climb into the play drawer. */
+  async function openPlayDrawerFromClimbsList(page: Page): Promise<void> {
+    await bindSeededBoard(page);
+    // Each list row is an accessible button named after its climb and wraps
+    // the blob thumbnail; force bypasses the row's cross-dissolve animation
+    // that never settles for hit-testing.
+    const firstClimbRow = page
+      .getByRole('button')
+      .filter({ has: page.locator('img[src^="blob:"]') })
+      .first();
+    await firstClimbRow.click({ force: true });
+    await expect(page.getByRole('button', { name: 'Log ascent' })).toBeVisible({ timeout: WARM_TIMEOUT_MS });
+  }
+
+  /**
+   * Play-drawer action buttons sit under a transparent react-native-gesture-
+   * handler layer: real pointers reach them (the layer is pass-through), but
+   * Playwright's hit-testing sees the overlay and never completes a default
+   * click. force skips the hit-test while still requiring the visible, enabled
+   * target we already asserted.
+   */
+  async function tapThroughGestureLayer(locator: ReturnType<Page['getByRole']>): Promise<void> {
+    // force-clicks still deliver events to the overlay at that position, so
+    // dispatch the click on the button element itself.
+    await locator.dispatchEvent('click');
+  }
+
+  /**
+   * Dismisses the presented sheet and waits for `stillVisible` to disappear.
+   * gorhom's web backdrop and the Escape key each dismiss on their own, but
+   * either can no-op if fired mid present-animation, so this alternates them
+   * until the sheet is actually gone (killing a dismiss-timing flake).
+   */
+  async function dismissSheet(page: Page, stillVisible: ReturnType<Page['getByRole']>): Promise<void> {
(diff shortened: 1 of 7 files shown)


packages/web/playwright.config.ts
```diff
@@ -60,7 +60,25 @@ export default defineConfig({
...
+    // Expo-web smoke — drives the mobile app compiled for the browser at /app.
+    // Requires the full expo-web dev stack (backend + Next proxy + Metro web):
+    //   vp run test:e2e:expo-web            (orchestrated: starts the stack, runs this)
+    // or vp run dev:mobile:web, then:
+    //   TEST_USER_EMAIL=test@boardsesh.com TEST_USER_PASSWORD=test \
+    //     cd packages/web && bunx playwright test --project=expo-web-smoke
+    // Viewport matches the Android parity reference (412×915) since the /app
+    // surface renders the Material variant.
+    {
+      name: 'expo-web-smoke',
```
marcodejongh, line 80: **[high]** 60s per-test timeout makes the 180s cold-load budget unreachable; cold runs fail

The `expo-web-smoke` project inherits the global `timeout: 60_000` and sets no override, so the spec's 180s cold-load budgets (`COLD_LOAD_TIMEOUT_MS` in `page.goto` and the first `toBeVisible`) and the 60s thumbnail expects can never actually elapse — Playwright kills the test at 60s regardless. In practice this doesn't break the documented entry points: `scripts/expo-web-e2e.ts` starts Playwright only after `/app` serves from the Next origin, and the orchestrator awaits `prewarmExpoWeb` (which fetches and fully consumes the Metro entry bundle) before starting Next, so every test runs against an already-compiled bundle. But the spec's stated design ("serial lets the first test absorb the compile") is not what the config delivers: if the bundle is ever cold at test time (e.g. a mid-stack edit invalidates Metro's cache), the first test dies at 60s, not 180s. Consider a project-level `timeout` on `expo-web-smoke` (e.g. 240_000), or trimming the misleading 180s constants and the "first test absorbs the compile" comment to match the real 60s budget.

packages/web/e2e/expo-web/smoke.spec.ts
```diff
@@ -0,0 +1,235 @@
...
+
+import { expect, test, type Page } from '@playwright/test';
+
+test.describe.configure({ mode: 'serial' });
+
+const TEST_EMAIL = process.env.TEST_USER_EMAIL ?? 'test@boardsesh.com';
+const TEST_PASSWORD = process.env.TEST_USER_PASSWORD ?? 'test';
+
+/** First navigation may include the on-demand Metro compile of the bundle. */
+const COLD_LOAD_TIMEOUT_MS = 180_000;
```
marcodejongh, line 26: **[medium]** 180s cold-load and 60s expect timeouts are dead letters under the 60s global test timeout

The config's global `timeout: 60_000` (playwright.config.ts:22) applies to the new `expo-web-smoke` project, which sets no per-project `timeout` — so every longer timeout in this spec is a dead letter: `COLD_LOAD_TIMEOUT_MS = 180_000` on the first `/app` goto/visibility wait can never elapse, and the two `timeout: 60_000` blob-thumbnail expects equal the whole test budget despite running after sign-in and several clicks. The header's "serial lets the first test absorb the compile" cannot happen: a compile longer than ~60s fails the first test, and serial mode then skips the remaining tests. In practice runs stay green because both documented entry points (`vp run test:e2e:expo-web` and `vp run dev:mobile:web`) go through the dev orchestrator, whose `prewarmExpoWeb` pays the cold Metro compile before the Next proxy is reachable — but any mid-session Metro cache invalidation (code edit, cache clear) before a re-run reintroduces a compile the 60s budget can't absorb, and the failure will look like a flaky 60s timeout rather than the designed 180s tolerance.

Fix: add `timeout: 240_000` to the `expo-web-smoke` project in `playwright.config.ts` (per-project `timeout` is supported in Playwright ^1.60), or drop the misleading 180s constants and comment.

packages/web/e2e/expo-web/smoke.spec.ts
```diff
@@ -0,0 +1,235 @@
...
+    }
+    expect(fatalErrors).toEqual([]);
+  });
+
+  test('home feed renders WASM board thumbnails; climbs tab shows the follow-your-wall state', async ({ page }) => {
+    const fatalErrors = collectFatalConsoleErrors(page);
+    await ensureSignedIn(page);
+    // WASM overlay renders resolve to blob: object URLs inside expo-image
+    // <img>s — the Home feed's session cards carry them without any board
+    // context.
```
marcodejongh, line 103: **[medium]** Non-retrying blob-thumbnail count races per-image WASM renders

`expect(await blobThumbnails.count()).toBeGreaterThanOrEqual(3)` snapshots the count once, immediately after only the *first* thumbnail became visible. Each `blob:` src appears only when that card's WASM render resolves (that's why the first thumbnail gets a 60s timeout), and nothing synchronizes the remaining renders — at the instant the first is visible the count can legitimately be 1 or 2. Because the suite is serial with `retries: 1`, this single flake re-runs the entire six-test chain and re-pays sign-in for every spec.

Use a retrying assertion instead, e.g. `await expect(blobThumbnails.nth(2)).toBeVisible({ timeout: 60_000 });` or `await expect.poll(() => blobThumbnails.count()).toBeGreaterThanOrEqual(3);`.

packages/web/e2e/expo-web/smoke.spec.ts
```diff
@@ -0,0 +1,235 @@
...
+    await expect(page.getByRole('button', { name: 'Find my board' })).toBeVisible({ timeout: WARM_TIMEOUT_MS });
+    expect(fatalErrors).toEqual([]);
+  });
+
+  /**
+   * Binds the seeded kilter board (Dyno Den) via the board sheet so the Climbs
+   * tab has a list and play-drawer actions aren't blocked by the switch-board
+   * overlay (feed climbs live on other boards). Fresh contexts start unbound.
+   */
+  async function bindSeededBoard(page: Page): Promise<void> {
```
marcodejongh, line 124: **[medium]** bindSeededBoard races an un-waited isVisible() and silently skips the binding step

`bindSeededBoard` gates the binding step on `findBoardButton.isVisible().catch(() => false)` — an instantaneous check with no wait, evaluated right after clicking the Climbs tab. The Climbs screen shows the "Find my board" empty state only once the async `useActiveBoard` storage read settles (`!hasBoardConfig && !isBoardLoading`), and test 2 in this same spec needs a 30s `toBeVisible` wait for this exact button, so the un-waited check can race false and silently skip binding.

The conditional guards nothing legitimate: the active board lives only in per-context localStorage (no server fallback, no `storageState` reuse in the `expo-web-smoke` project), so every test starts unbound — as the function's own comment says — and binding is always required. When the check races false, the test either dies later on the 60s blob-thumbnail wait (pointing debugging at WASM rendering instead of the skipped bind) or, since inactive tab scenes stay mounted on web, passes spuriously against Home-feed thumbnails without ever exercising the bound-list flow.

Make it an assertion: `await expect(findBoardButton).toBeVisible({ timeout: WARM_TIMEOUT_MS }); await findBoardButton.click({ force: true });` — a real regression in the empty state then fails here with a precise error.

scripts/expo-web-e2e.ts
```diff
@@ -0,0 +1,122 @@
...
+import process from 'node:process';
+
+const repoRoot = path.resolve(__dirname, '..');
+const READY_LINE = /\[dev\] Expo web app: (\S+)/;
+const APP_READY_TIMEOUT_MS = 300_000;
+const APP_POLL_INTERVAL_MS = 2_000;
+
+function terminate(child: ChildProcess): Promise<void> {
+  return new Promise((resolve) => {
+    if (child.exitCode !== null) {
```
marcodejongh: **[medium]** terminate() hangs on a signal-killed orchestrator, turning failures into exit code 0

`terminate` checks only `child.exitCode !== null`, but a child killed by a signal (e.g. the OOM killer taking down the Metro-heavy orchestrator, or an external SIGKILL) has `exitCode === null` and `signalCode` set. For such an already-dead child the function registers `once('exit')` (the event already fired and never fires again) and `kill()`s a dead process (a no-op), so the returned promise never settles.

Consequence: if the orchestrator is signal-killed mid-run, Playwright fails, `finally` awaits `terminate`, the promise hangs, `main()` never resolves, and neither `process.exit(exitCode)` nor the `.catch` `process.exit(1)` runs. The SIGKILL timer is `unref`'d and the child's stdout pipe is closed, so the event loop drains and Node exits 0 — `vp run test:e2e:expo-web` reports green exactly when the stack crashed. The same hang hits the early-exit `.catch` path, swallowing the `dev orchestrator exited early` error.

Fix: `if (child.exitCode !== null || child.signalCode !== null) { resolve(); return; }`.

vite.config.ts
```diff
@@ -749,6 +749,14 @@ export default defineConfig({
         dependsOn: ['db:up'],
         cache: false,
       },
+      // Expo-web smoke: boots the full expo-web stack (backend + Next proxy +
+      // Metro web) via the dev orchestrator and runs the `expo-web-smoke`
+      // Playwright project against it. See scripts/expo-web-e2e.ts.
+      'test:e2e:expo-web': {
```
marcodejongh, line 755: **[medium]** Smoke suite never runs in CI despite being framed as the stack's regression net

Nothing in CI runs this suite: no workflow references `expo-web` or `test:e2e:expo-web`, and the e2e shards in `.github/workflows/e2e-tests.yml` invoke `bunx playwright test --project=chromium` (plus the two screenshot projects), so the new `expo-web-smoke` project is skipped even when that workflow is dispatched. Yet the PR description calls this "the automated regression net for every later PR in the Expo-web stack" — as landed it's a local-only tool, and the follow-up PRs in the stack (#3761, #3762, #3764, #3765) get no automated protection from it. Note `e2e-tests.yml` is `workflow_dispatch`-only, so wiring a non-sharded `tsx scripts/expo-web-e2e.ts` job into it (the runner is self-contained given db:up + mobile:web-runtime:install) gives manual-dispatch coverage like the rest of the e2e suite, not per-PR gating. Either add that job (or a PR-triggered one) or correct the PR framing so later stack PRs don't assume coverage that doesn't exist.

packages/web/playwright.config.ts
```diff
@@ -60,7 +60,25 @@ export default defineConfig({
...
-      testIgnore: ['**/layout-screenshots.spec.ts', '**/help-screenshots.spec.ts'],
+      testIgnore: ['**/layout-screenshots.spec.ts', '**/help-screenshots.spec.ts', '**/expo-web/**'],
+    },
+
+    // Expo-web smoke — drives the mobile app compiled for the browser at /app.
+    // Requires the full expo-web dev stack (backend + Next proxy + Metro web):
+    //   vp run test:e2e:expo-web            (orchestrated: starts the stack, runs this)
+    // or vp run dev:mobile:web, then:
+    //   TEST_USER_EMAIL=test@boardsesh.com TEST_USER_PASSWORD=test \
+    //     cd packages/web && bunx playwright test --project=expo-web-smoke
```
marcodejongh, line 72: **[low]** Documented manual invocation omits PLAYWRIGHT_SKIP_CLASSIC_SETUP=1

The manual invocation documented here (`vp run dev:mobile:web`, then `TEST_USER_EMAIL=... TEST_USER_PASSWORD=... bunx playwright test --project=expo-web-smoke`) omits `PLAYWRIGHT_SKIP_CLASSIC_SETUP=1`, which only `scripts/expo-web-e2e.ts` sets. Without it, `e2e/global-setup.ts` still seeds and prewarms the classic Next routes — which the new global-setup comment itself says can exceed the 30s card timeout under the expo-web stack (cold compile) and fail the run for surfaces the smoke never visits. Add `PLAYWRIGHT_SKIP_CLASSIC_SETUP=1` to the documented command.

marcodejongh (commented):

Multi-agent Claude review run by Marco across the whole Expo-web stack (#3752 / #3760 / #3761 / #3762 / #3764 / #3765 / #3766). Each PR was reviewed only on its own commits (diff vs its stack parent) by lens-scoped reviewers, and every finding below survived adversarial verification by independent refuter agents before posting. **7 findings** for this PR.

packages/web/playwright.config.ts
marcodejongh: Fixed in 479a11989: the expo-web-smoke project now sets its own `timeout: 240_000`, so the spec's 180s COLD_LOAD_TIMEOUT_MS and the 60s thumbnail expects are actually reachable when the bundle is cold (prewarm cache invalidated by a mid-stack edit).

packages/web/e2e/expo-web/smoke.spec.ts
marcodejongh: Fixed in 479a11989: same root cause as the config finding — the project-level `timeout: 240_000` on expo-web-smoke makes the 180s cold-load budget and the 60s expects live again, and the "first test absorbs the compile" header now matches what the config delivers.

packages/web/e2e/expo-web/smoke.spec.ts
marcodejongh: Fixed in 554868d79: replaced the one-shot `count()` snapshot with `await expect(blobThumbnails.nth(2)).toBeVisible({ timeout: 60_000 })`, which retries until at least three WASM renders have resolved instead of racing the in-flight ones.

packages/web/e2e/expo-web/smoke.spec.ts
marcodejongh: Fixed in 6f1bdb6f2 (pushed before this pass): bindSeededBoard now first waits for the Climbs tab to settle into a determinate state — `await expect(findBoardButton.or(boundThumbnail).first()).toBeVisible({ timeout: WARM_TIMEOUT_MS })` — before deciding whether to bind, and the post-bind thumbnail assertion is unconditional so a renamed empty-state button fails loudly here rather than passing vacuously downstream.

scripts/expo-web-e2e.ts
marcodejongh: Fixed in 404f3a05e: terminate() (and the SIGKILL escalation timer) now treat `signalCode !== null` as exited via a shared `hasExited()` helper, so a signal-killed orchestrator resolves immediately instead of hanging the promise and draining the event loop into a spurious exit 0.

vite.config.ts
marcodejongh: Fixed in 835d9063d: added a non-sharded `expo-web-smoke` job to `.github/workflows/e2e-tests.yml` that runs `vp run test:e2e:expo-web` on a bare runner (the vp task's db:up dependency needs the Docker daemon; dev-mode Next + Metro don't consume the prebuilt artifact), with the bare-runner node_modules cache key and Playwright report/results artifacts. Coverage matches the rest of the workflow (manual dispatch, not per-PR gating), and the CLAUDE.md note now points at the job instead of claiming the suite is local-only.

packages/web/playwright.config.ts
marcodejongh: Fixed in 479a11989: the documented manual invocation now includes `PLAYWRIGHT_SKIP_CLASSIC_SETUP=1` plus a note on why it's needed (global-setup otherwise seeds/prewarms classic Next routes the smoke never visits, which can blow the 30s card timeout under the expo-web stack).

marcodejongh:
## Claude: ready to merge ✅

All review findings on this PR are resolved at head `ac0f4b033`: the multi-agent swarm review, follow-up Codex and claude[bot] batches, and QA verification passes are addressed (fixed, refuted with evidence, or filed as follow-up issues). CI is green with no unresolved threads.

marcodejongh:

PR #3761, opened 2026-07-17
feat(web): serve the Expo web export at /app in production

marcodejongh:
## Summary

Serves the static Expo web export at `/app` in production, so the mobile-app-as-web-app works off a deployed artifact and not just the dev Metro proxy.

- **`next.config.mjs`**: with `BOARDSESH_WEB=1` and no `BOARDSESH_EXPO_WEB_ORIGIN` (production), `rewrites()` returns SPA fallbacks `/app` and `/app/:path*` → `/app/index.html` (afterFiles, so real export files win first); the dev-proxy branch moved to beforeFiles so a stale local export can't shadow Metro. Immutable `Cache-Control` on `/app/(_expo|assets)/:path*`; `X-Robots-Tag: noindex, follow` + security headers preserved on all `/app` paths.
- **Build**: new `scripts/build-expo-web-export.sh` (single export recipe, default output `packages/web/public/app`); `scripts/mobile-web-bundle-check.sh` delegates to it; `vite.config.ts` gains `build:expo-web` and `build:web` declares `env: ['BOARDSESH_WEB']`. `Dockerfile.web` builder relinks bun's isolated per-package node_modules, runs the export, and builds Next with `ARG BOARDSESH_WEB=1`; the runner's existing `public/` COPY ships the artifact.
- **docs/expo-web-deployment.md**: dev proxy vs prod static, artifact flow, rollback (don't ship the export → `/app` 404s, gated by the middleware).

## Verification

- **Full `docker build -f Dockerfile.web` + `docker run` + curl**: `/app` → 200 shell with noindex + security headers; deep route → 200 shell; hashed `_expo` bundle → 200 JS `Cache-Control: immutable`; `/app/wasm/*.wasm` → 200 `application/wasm`; `/` unaffected. Same against a local `vp run build:expo-web` + `BOARDSESH_WEB=1 vp run build:web` standalone server.
- `vp run typecheck:web`, scoped web tests (8/8), `vp run check:mobile-web-bundle`, deploy-inputs check, `vp check` — all green. No changes to `packages/mobile`, root `package.json`, or `bun.lock` (fingerprint-safe, diff-verified).

## Notes for review

- **Pre-existing latent CI break fixed**: `packages/web/scripts/dev-with-tailscale.ts` imports repo-root `scripts/lib/tailscale-hostname`, which was absent from the web Docker context — Next's production type-check failed on it, so the base branch's web preview image was already unbuildable. Now shipped via `extraSourceFiles`.
- **Deferred — Vercel production path**: `production-deploy.yml` builds prod web via `vercel build`, not `Dockerfile.web`, so prod-on-Vercel `/app` 404s (the documented rollback state) until the export step is added to that job or web moves to the Docker image. Branch previews (Docker) get `/app` automatically now.

## Release Notes

- [x] No release note needed (internal / infrastructure change)

_Stacked on #3752 (Expo web Phase 0). Review after #3752._

### DIFF
--- a/Dockerfile.web
+++ b/Dockerfile.web
@@ -27,6 +27,9 @@ FROM node:22-alpine AS builder
 
 WORKDIR /app
 
+# bash runs scripts/build-expo-web-export.sh (repo scripts are bash, not sh)
+RUN apk add --no-cache bash
+
 # Copy bun binary from deps stage (avoid re-downloading)
 COPY --from=deps /root/.bun/bin/bun /usr/local/bin/bun
 RUN ln -s /usr/local/bin/bun /usr/local/bin/bunx
@@ -35,16 +38,51 @@ RUN ln -s /usr/local/bin/bun /usr/local/bin/bunx
 COPY --from=deps /app/node_modules ./node_modules
 
 # Source-only edits land after install, so they do not invalidate the dependency layer.
+# The full workspace manifest set comes first: the lockfile spans every
+# workspace, while source/ only carries this service's packages, so the frozen
+# relink below needs the remaining package.json stubs to accept the lockfile.
+# source/packages then overwrites its subset with identical manifests + source.
+COPY manifests/packages ./packages
 COPY source/package.json source/bun.lock ./
 COPY source/patches ./patches
 COPY source/packages ./packages
+COPY source/scripts ./scripts
+
+# Bun's isolated linker resolves workspace deps through per-package symlink
+# farms (packages/*/node_modules -> ../../node_modules/.bun). The deps stage
+# created them next to the manifests, but only the root node_modules (with the
+# .bun store) crosses the stage boundary, so relink before the Expo export
+# resolves modules from the source tree — `bunx expo` must find the workspace's
+# pinned expo, never fall back to fetching a floating version.
+RUN bun install --frozen-lockfile
 
 # Accept build arguments for environment variables baked into the Next.js build
 ARG NEXT_PUBLIC_WS_URL
 ARG BASE_URL
+# BOARDSESH_WEB=1 turns on the Expo web export below AND the /app serving
+# branch that next.config.mjs bakes into the standalone build. Rollback lever:
+# build with --build-arg BOARDSESH_WEB=0 to ship an image without the Expo web
+# app — /app then 404s while the rest of the site is untouched.
+ARG BOARDSESH_WEB=1
 
 ENV NEXT_PUBLIC_WS_URL=$NEXT_PUBLIC_WS_URL
 ENV BASE_URL=$BASE_URL
+ENV BOARDSESH_WEB=$BOARDSESH_WEB
+
+# Static Expo web export served by Next at /app. The nested web-runtime install
+# supplies the web-only renderer (react-native-web, react-dom, the gorhom sheet)
+# outside the native fingerprint graph — the export script runs it when missing.
+# The export lands in packages/web/public/app, which the runner stage picks up
+# through the existing public/ COPY and Next's standalone server then serves
+# statically. EXPO_PUBLIC_WEB_URL feeds the Expo Router head origin; the app's
+# runtime URLs (backend/WS) fall back to production defaults in src/lib/env.ts.
+# Only export EXPO_PUBLIC_WEB_URL when BASE_URL is non-empty: Expo inlines
+# EXPO_PUBLIC_* values verbatim, and a set-but-empty value would override the
+# production fallback in src/lib/env.ts with ''.
+RUN if [ "$BOARDSESH_WEB" = "1" ]; then \
+      if [ -n "$BASE_URL" ]; then export EXPO_PUBLIC_WEB_URL="$BASE_URL"; fi; \
+      bash scripts/build-expo-web-export.sh packages/web/public/app; \
+    fi
 
 # Build the Next.js application (standalone output)
 RUN bun run --filter=@boardsesh/web build
--- a/Dockerfile.web
+++ b/Dockerfile.web
@@ -94,6 +94,16 @@ WORKDIR /app
 
 ENV NODE_ENV=production
 
+# BOARDSESH_WEB is read at REQUEST time by the standalone server, not just at
+# build: app/layout.tsx gates the Expo auth session bridge on it
+# (enableExpoAuthBridge) and middleware.ts gates the /app dev proxy. The builder
+# ARG/ENV above does not cross the stage boundary, so redeclare it here or the
+# runtime flag is undefined and the bridge stays off even though /app is served.
+# The default mirrors the builder so a single --build-arg BOARDSESH_WEB=0
+# rollback disables both the export AND the runtime bridge consistently.
+ARG BOARDSESH_WEB=1
+ENV BOARDSESH_WEB=$BOARDSESH_WEB
+
 # Create non-root user
 RUN addgroup --system --gid 1001 nodejs && \
     adduser --system --uid 1001 nextjs
--- a/packages/web/next.config.mjs
+++ b/packages/web/next.config.mjs
@@ -158,7 +158,14 @@ const nextConfig = {
       // docs/expo-web-deployment.md).
       return [
         { source: '/app', destination: '/app/index.html' },
-        { source: '/app/:path*', destination: '/app/index.html' },
+        // SPA fallback for Expo Router routes ONLY. The content-hashed namespaces
+        // (_expo/, assets/) and the fixed-name WASM glue (wasm/) are excluded so a
+        // request for a missing hashed bundle 404s instead of falling through to
+        // the shell. Otherwise Next's headers() would stamp the index.html body
+        // with the immutable Cache-Control that matches the incoming .js/.wasm path
+        // (line 105), and a stale-shell or mid-deploy client would cache an HTML
+        // page at a bundle URL forever — permanently bricking that URL.
+        { source: '/app/:path((?!_expo/|assets/|wasm/).*)', destination: '/app/index.html' },
       ];
     }
 
--- a/packages/web/next.config.mjs
+++ b/packages/web/next.config.mjs
@@ -98,6 +98,14 @@ const nextConfig = {
         headers: [{ key: 'X-Robots-Tag', value: 'noindex, follow' }],
       },
       {
+        // Exported Expo web bundles are content-hashed (entry-<hash>.js under
+        // _expo/static, md5-named files under assets/), so they can be cached
+        // forever. The SPA shell (/app/index.html) and the fixed-name WASM glue
+        // under /app/wasm keep the default no-store-ish behaviour on purpose.
+        source: '/app/:hashedDir(_expo|assets)/:path*',
+        headers: [{ key: 'Cache-Control', value: 'public, max-age=31536000, immutable' }],
+      },
+      {
         // Every route EXCEPT /embed/** keeps the frame-denying default.
         // If this exclusion ever regresses, the fail-safe is SAMEORIGIN
         // (embeds break visibly rather than the whole site becoming frameable).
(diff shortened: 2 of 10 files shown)


marcodejongh:
Superseded by #3766 — the rollout was reworked from the `/app` path approach to a **subdomain** (`app.boardsesh.com`) with a shared `.boardsesh.com` cookie. This PR's serving/redirect approach is no longer the plan; the reusable pieces (baseUrl config, CORS, CTA, PR5-map removal) were carried into #3766. Leaving open for reference — close if you prefer.

packages/web/next.config.mjs
```diff
@@ -138,43 +146,62 @@ const nextConfig = {
...
+      // it in the builder stage), so real files — /app/_expo/* bundles,
+      // /app/assets/*, /app/wasm/* — are served straight from public/ before
+      // these afterFiles rewrites are consulted. Everything else under /app is
+      // an Expo Router SPA route and falls back to the exported shell. When the
+      // export was not built into the deploy, /app/index.html resolves to
+      // nothing and /app 404s — that absence is the rollback lever (see
+      // docs/expo-web-deployment.md).
+      return [
+        { source: '/app', destination: '/app/index.html' },
+        { source: '/app/:path*', destination: '/app/index.html' },
```
marcodejongh: **[medium]** SPA fallback serves index.html with immutable Cache-Control for missing hashed assets

Next.js `headers()` rules match the incoming request path and apply regardless of what a rewrite ends up serving. A request for a hashed bundle that doesn't exist on disk — e.g. `/app/_expo/static/js/entry-<oldhash>.js` from a stale shell or during a deploy transition — misses `public/`, hits the afterFiles catch-all `{ source: '/app/:path*', destination: '/app/index.html' }` (line 161), and returns the index.html body with status 200 and `Cache-Control: public, max-age=31536000, immutable` (the line 105 rule matches the incoming `.js` path). `immutable` means browsers won't revalidate even on normal reload, and the sitewide `nosniff` header (which also matches `/app/**`) hard-blocks the mis-typed module script — so once a client (or any shared cache/CDN in front, if one is configured) caches HTML at a content-hashed bundle URL, the app stays broken at that URL whenever a shell references that hash again.

Fix: keep the hashed namespaces out of the SPA fallback so a missing asset 404s instead of falling back:

```js
return [
  { source: '/app', destination: '/app/index.html' },
  { source: '/app/:path((?!_expo/|assets/|wasm/).*)', destination: '/app/index.html' },
];
```

and add an assertion in `packages/web/app/__tests__/next-config-expo-web.test.ts` that `/app/_expo/...` does not match the fallback.

scripts/build-expo-web-export.sh
```diff
@@ -0,0 +1,58 @@
...
+#   - Dockerfile.web's builder stage (same default target),
+#   - scripts/mobile-web-bundle-check.sh (temp dir; export-only validation).
+#
+# Usage: bash scripts/build-expo-web-export.sh [output-dir]
+#
+# The output directory is wiped first, so never point it at a directory with
+# hand-maintained content. See docs/expo-web-deployment.md.
+
+ROOT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")/.." && pwd)"
+OUTPUT_DIR="${1:-$ROOT_DIR/packages/web/public/app}"
```
marcodejongh, line 24: **[medium]** Literal `--` forwarded by vp becomes the output dir; export lands in ./--

`vite.config.ts`'s new `build:expo-web` task advertises "pass `-- <output-dir>` for a different target", but vp forwards the literal `--` into the script's argv — which is why nearly every arg-parsing script in `scripts/` filters it (`mobile-screenshots.ts`, `mobile-ota-setup.ts`, `mobile-dev-start.ts`, ...). This script reads `$1` raw:

```bash
OUTPUT_DIR="${1:-$ROOT_DIR/packages/web/public/app}"
```

So `vp run build:expo-web -- /tmp/out` sets `OUTPUT_DIR="--"` → `"$PWD/--"`: the script `rm -rf`s and exports into a directory literally named `--` at the repo root, passes its own index.html/WASM checks there, and exits 0 (the final log does name the odd path, but `/tmp/out` is never populated and the second arg is silently ignored). The direct callers (Dockerfile.web, mobile-web-bundle-check.sh) pass the dir without `--` and are unaffected — only the documented `vp run` invocation breaks.

Fix, matching the repo convention:

```bash
if [[ "${1:-}" == "--" ]]; then shift; fi
OUTPUT_DIR="${1:-$ROOT_DIR/packages/web/public/app}"
```

scripts/build-expo-web-export.sh
```diff
@@ -0,0 +1,58 @@
...
+if [[ "$OUTPUT_DIR" != /* ]]; then
+  OUTPUT_DIR="$PWD/$OUTPUT_DIR"
+fi
+
+if [[ "$OUTPUT_DIR" == "$ROOT_DIR" || "$ROOT_DIR" == "$OUTPUT_DIR"/* ]]; then
```
marcodejongh: **[low]** rm -rf guard bypassed by trailing slash and doesn't cover repo subdirectories

The rm -rf guard is defeated by a trailing slash: for `OUTPUT_DIR="$ROOT_DIR/"` neither `"$OUTPUT_DIR" == "$ROOT_DIR"` (extra slash) nor `"$ROOT_DIR" == "$OUTPUT_DIR"/*` (pattern becomes `<root>//*`) matches, so `rm -rf` wipes the repo (verified in bash). Shell tab-completion appends exactly that trailing slash, and manual invocation with an output-dir arg is the documented usage. Fix: normalize after absolutization — `OUTPUT_DIR="${OUTPUT_DIR%/}"`. Secondary: the guard only protects the root and its ancestors; the one-segment typo `packages/web/public` (omitting `/app`) wipes the committed public/ tree including the gitignored, unrecoverable `public/uploads` — the header comment warns about this, but optionally refuse to wipe a non-empty directory lacking an `index.html` from a previous export.

scripts/build-expo-web-export.sh
```diff
@@ -0,0 +1,58 @@
...
+WEB_RUNTIME_DIR="$ROOT_DIR/packages/mobile/web-runtime"
+if [[ ! -d "$WEB_RUNTIME_DIR/node_modules/react-native-web" ]]; then
+  # The web-only renderer lives in an isolated nested install so it never
+  # enters the native fingerprint graph. `vp run build:expo-web` installs it
+  # via its task dependency; direct callers (Docker) land here.
+  echo "[build-expo-web-export] installing packages/mobile/web-runtime dependencies"
+  bun install --cwd "$WEB_RUNTIME_DIR" --frozen-lockfile
+fi
+
+rm -rf "$OUTPUT_DIR"
```
marcodejongh, line 60: **[low]** rm -rf guard only protects repo-root ancestors, not other real directories

The guard at lines 22–25 only refuses when `OUTPUT_DIR` is the repo root or an ancestor of it. Any other existing directory is wiped without question: `bash scripts/build-expo-web-export.sh packages/web/public` (an easy typo for `packages/web/public/app`) deletes the whole committed `public/` tree including the gitignored `packages/web/public/uploads` (unrecoverable local dev data); an absolute path outside the repo is deleted just as silently.

Since all three sanctioned callers pass either the default or a fresh temp dir, the contract could be tightened cheaply: refuse when the target exists, is non-empty, and doesn't look like a previous export (e.g. lacks `index.html` + `_expo/`), or require the basename to be `app`/a temp dir unless `--force` is passed. The header comment saying "never point it at a directory with hand-maintained content" documents the footgun instead of removing it.

scripts/create-service-docker-context.mjs
```diff
@@ -23,7 +23,16 @@ const services = {
   },
   web: {
     dockerfile: 'Dockerfile.web',
-    rootPackageName: '@boardsesh/web',
+    // @boardsesh/mobile rides along because Dockerfile.web's builder stage runs
+    // the static Expo web export (served by Next at /app); the union walk pulls
+    // in the mobile app plus its transitive workspace deps.
+    rootPackageNames: ['@boardsesh/web', '@boardsesh/mobile'],
```
marcodejongh, line 29: **[low]** Stale local public/app export rides into the web Docker context

Stale local Expo web export can ride into a locally generated web Docker context. `copyDirectory` copies `packages/web` from the working tree without consulting `.gitignore`, and the newly gitignored `packages/web/public/app` (the default output of `vp run build:expo-web` / `scripts/build-expo-web-export.sh`) is not excluded — and can't be excluded via `ignoredDirectoryNames`, which is name-global and would also drop `packages/web/app` and `packages/mobile/app`.

Impact is limited: the deployed path (branch-deploy CI) generates the context from a clean checkout, and with the default `BOARDSESH_WEB=1` the export script `rm -rf`s and rebuilds the directory anyway. The gap is a locally built image with the documented rollback `--build-arg BOARDSESH_WEB=0`: the export step is skipped, so the developer's stale `public/app` ships via the runner's `public/` COPY. `/app` itself still 404s (no rewrites are baked without the flag), but the stale shell and bundles stay directly servable at literal paths like `/app/index.html` and `/app/_expo/...`, partially defeating the rollback and making contexts non-reproducible between dev machines and CI.

Fix: add a path-based exclusion (e.g. extend `shouldSkipSourceEntry` to skip `packages/web/public/app` by relative path) so the export can never enter the context — consistent with the ignore set's existing purpose of keeping local build artifacts out.

vite.config.ts
```diff
@@ -310,6 +310,20 @@ export default defineConfig({
...
+        // Forwarded into the task (vp runs tasks with a filtered environment)
+        // and part of the cache key: BOARDSESH_WEB=1 bakes the /app static
+        // serving rewrites into the standalone build (see next.config.mjs).
+        env: ['BOARDSESH_WEB'],
+      },
+      'build:expo-web': {
+        // Static Expo web export into packages/web/public/app — the same
+        // artifact Dockerfile.web bakes into the production image. Run this
+        // before a BOARDSESH_WEB=1 `vp run build:web` to verify /app serving
+        // locally; pass `-- <output-dir>` for a different target. The output
```
marcodejongh: **[low]** Advertised `-- <output-dir>` forwarding breaks: vp passes the literal `--` as $1

The task comment advertises `pass \`-- <output-dir>\` for a different target`, but vp forwards the literal `--` separator into the script's argv (known repo footgun — the db dedupe CLIs needed the same fix in #3161). `scripts/build-expo-web-export.sh` reads `$1` raw, so `vp run build:expo-web -- /some/dir` sets `OUTPUT_DIR="--"` → `"$PWD/--"`: the script `rm -rf`s and exports into a junk `./--` directory, its own index.html/WASM checks pass against it, and the intended target never receives the export (only the final echo hints at the wrong path). Fix in the script, before reading `$1`:

```bash
if [[ "${1-}" == "--" ]]; then shift; fi
```

marcodejongh (commented):

Multi-agent Claude review run by Marco across the whole Expo-web stack (#3752 / #3760 / #3761 / #3762 / #3764 / #3765 / #3766). Each PR was reviewed only on its own commits (diff vs its stack parent) by lens-scoped reviewers, and every finding below survived adversarial verification by independent refuter agents before posting. **6 findings** for this PR.

scripts/build-expo-web-export.sh
marcodejongh: Fixed in 114fe92: the script now drops a leading literal `--` (`if [[ "${1:-}" == "--" ]]; then shift; fi`) before reading `$1`, matching the repo convention, so `vp run build:expo-web -- <dir>` exports into `<dir>` instead of `./--`. Verified by invoking the arg/guard prologue directly with `-- /tmp/out` (resolves to `/tmp/out`) and `-- <scratch dir>`.

scripts/build-expo-web-export.sh
marcodejongh: Fixed in 114fe92: `OUTPUT_DIR` is now normalized with `${OUTPUT_DIR%/}` after absolutization, so a trailing slash no longer slips `<root>/` past the repo-root guard. Verified: `<root>/` now hits the refuse-root branch instead of rm -rf. Also added an empty/`-z` check to the guard.

scripts/build-expo-web-export.sh
marcodejongh: Fixed in 114fe92: added a guard that refuses to wipe a target that exists, is non-empty, and lacks an `index.html` (i.e. is not a previous export). The one-segment typo `packages/web/public` now aborts instead of deleting the committed `public/` tree (incl. the gitignored `public/uploads`); a fresh/empty temp dir or a prior export (has index.html) still proceeds. Verified all cases directly.

scripts/create-service-docker-context.mjs
marcodejongh: Fixed in 9dd2fa0: added a path-based exclusion. `copyDirectory` now threads the repo root and `shouldSkipSourceEntry` skips exact repo-relative paths via `ignoredRepoRelativePaths` (`packages/web/public/app`) — a basename match would also drop `packages/web/app` and `packages/mobile/app`, so this is by exact path. Verified with a node harness: a seeded stale `public/app/index.html` does not enter the generated context while the rest of `public/` still copies.

vite.config.ts
marcodejongh: Fixed in 114fe92 (script) + the vite.config task comment now notes the script strips vp's forwarded `--`, so the advertised `-- <output-dir>` forwarding works as documented.

packages/web/next.config.mjs
marcodejongh: Fixed in a19016d: the SPA fallback now excludes the content-hashed namespaces and the fixed-name WASM glue — `{ source: '/app/:path((?!_expo/|assets/|wasm/).*)', destination: '/app/index.html' }` — so a request for a missing hashed bundle 404s instead of returning index.html with the immutable Cache-Control. Added a test that compiles the fallback pattern and asserts `_expo/...`, `assets/...`, and `wasm/...` do not match while real Expo Router routes do.

scripts/build-expo-web-export.sh
marcodejongh: Correction: my earlier `index.html` marker was defective — QA rightly flagged that the committed `packages/web/public` already carries an `index.html`, so the exact `packages/web/public` typo this guards against still proceeded (and would have wiped the gitignored `public/uploads`).

Now actually fixed in 2413a6f: the marker is the `_expo/` directory (always emitted by `expo export --platform web`, absent from `packages/web/public`), and repeated trailing slashes are stripped in a loop so `<root>//` can't slip through. Re-verified by direct invocation without executing the rm: `packages/web/public` (both absolute and relative) and `<root>//` are REFUSED, while the default `public/app` target, an empty temp dir, and a real prior export (with `_expo/`) proceed.

Dockerfile.web
```diff
@@ -35,16 +38,51 @@ RUN ln -s /usr/local/bin/bun /usr/local/bin/bunx
...
 ARG BASE_URL
+# BOARDSESH_WEB=1 turns on the Expo web export below AND the /app serving
+# branch that next.config.mjs bakes into the standalone build. Rollback lever:
+# build with --build-arg BOARDSESH_WEB=0 to ship an image without the Expo web
+# app — /app then 404s while the rest of the site is untouched.
+ARG BOARDSESH_WEB=1
 
 ENV NEXT_PUBLIC_WS_URL=$NEXT_PUBLIC_WS_URL
 ENV BASE_URL=$BASE_URL
+ENV BOARDSESH_WEB=$BOARDSESH_WEB
```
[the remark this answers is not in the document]
marcodejongh: Fixed in 3f236b5: redeclared `ARG BOARDSESH_WEB=1` + `ENV BOARDSESH_WEB=$BOARDSESH_WEB` in the runner stage, so the runtime flag matches the build. Confirmed the impact — `app/layout.tsx` is a server component that gates `enableExpoAuthBridge` on `process.env.BOARDSESH_WEB === '1'` at request time, and `middleware.ts` reads it too; both were seeing `undefined` in the runner. The default mirrors the builder, so a single `--build-arg BOARDSESH_WEB=0` rollback now disables the export AND the runtime bridge together.

While in there I checked the other builder-only vars: `NEXT_PUBLIC_WS_URL` is inlined into the bundle at build time (Next replaces `process.env.NEXT_PUBLIC_*`), so no runner env is needed. `BASE_URL` is read at runtime in three auth routes (`register`/`forgot-password`/`resend-verification`) but always with a `?? request.nextUrl.origin` fallback and it predates this PR, so it degrades rather than breaking — flagging it as a separate pre-existing item rather than changing it here.

marcodejongh:
## Claude: ready to merge ✅

All review findings on this PR are resolved at head `3f236b58b`: the multi-agent swarm review, follow-up Codex and claude[bot] batches, and QA verification passes are addressed (fixed, refuted with evidence, or filed as follow-up issues). CI is green with no unresolved threads.

marcodejongh:

PR #3762, opened 2026-07-17
feat(web): flag-gated Expo-web rollout redirect map

marcodejongh:
## Summary

Phase 3 rollout scaffolding: a flag-gated per-surface redirect that routes logged-in web visitors from migrated board surfaces to the `/app` Expo-web SPA. **Off by default in every dimension** — a logged-out visitor, a crawler, or a visitor whose flag hasn't resolved never redirects, so public/SEO board views stay canonical.

Gate (all must hold): `BOARDSESH_WEB=1` **and** the mirrored `expo-web-app` flag cookie is on **and** a next-auth session cookie is present **and** no `bs_classic` opt-out cookie. Escape hatch: `?classic=1` pins `bs_classic` and lands the visitor back on the classic URL (param stripped), mirroring the existing `?session=` handling. `307` (not `308`) so flipping the flag off is never cached as permanent.

### Flag evaluation
There is no server-side PostHog eval in this repo (flags are client-only via `FeatureFlagsProvider`; edge middleware can't call PostHog). So `ExpoWebRolloutCookieSync` reads `useFeatureFlag('expo-web-app')` and mirrors it into the non-HttpOnly `bs_expo_web` cookie; middleware reads that cookie on the edge. Flag off → cookie cleared on next load → instant revert. The PostHog flag itself (`expo-web-app`, id **767179**, project 412845) is **inactive / 0%** and untouched.

### Redirect map
| Classic surface | → | Board context carried |
|---|---|---|
| `/[board]/[layout]/[size]/[sets]/[angle]/list` | `/app/climbs` | numeric board params |
| `/b/[slug]/[angle]/list` | `/app/climbs` | `boardSlug` + `angle` |
| `/[board]/[layout]/[size]/[sets]/[angle]/view/[uuid]` | `/app/climbs/[uuid]` | numeric board params |
| `/b/[slug]/[angle]/view/[uuid]` | `/app/climbs/[uuid]` | `boardSlug` + `angle` |

Everything else (board create/queue, profiles, playlists, gyms, setters, API, homepage) → classic.

## Reviewer decisions to confirm

1. **view → ClimbDetail, not `/play`.** Mapped `view/[uuid]` → `/app/climbs/[uuid]` (the deep-link that opens the play drawer) rather than `/app/play`, because the mobile `/play` route plays the current queue climb and reads no UUID — it can't target a specific climb. Confirm this is the intended UX.
2. **Slug-view fidelity.** The slug form can't decompose to numeric `layoutId/sizeId/setIds` in edge middleware (no DB), so it carries `boardSlug`+`angle`; the SPA must resolve the slug for that surface to fully open. Verify SPA-side slug resolution before widening rollout (noted in the doc).
3. **List redirect lands on the SPA's stored active board** — the mobile climbs route reads its board from AsyncStorage, not URL params; the board-context query params attached here are forward-compatible but not consumed today.
4. **Login = session-cookie presence** (not JWT-verified), same heuristic as `getServerAuthToken`. A stale cookie at worst lands on the SPA, which re-auths.
5. Only `list`/`view` mapped per scope; web `/play/[uuid]` left classic.

## Release Notes

- [x] No release note needed (internal / infrastructure change; flag off)

## Test plan
- `vp run typecheck:web`, `vp test run --project web` (middleware + expo-web-rollout: **99 passed**), `vp check` clean on all touched files, `vp run check:i18n` OK. No changes to `packages/mobile`, root `package.json`, or `bun.lock`.

_Stacked on #3761 (prod serving) → #3752 (Phase 0). Review last. Rollout is a post-merge PostHog action; rollback = flag → 0% or `?classic=1` per user._

### DIFF
--- a/packages/web/app/lib/board-route-paths.ts
+++ b/packages/web/app/lib/board-route-paths.ts
@@ -48,18 +48,14 @@ export function isBoardListPath(pathname: string | null | undefined): boolean {
   return segments.length === 6 && segments[5] === 'list';
 }
 
-export function isBoardViewPath(pathname: string | null | undefined): boolean {
-  if (!pathname || !isBoardRoutePath(pathname)) return false;
-
-  const segments = getPathSegments(pathname);
-
-  // Slug form: /b/[board_slug]/[angle]/view/[climb_uuid]
-  if (segments[0] === 'b') {
-    return segments.length === 5 && segments[3] === 'view';
-  }
-
-  // Legacy form: /[board]/[layout]/[size]/[sets]/[angle]/view/[climb_uuid]
-  return segments.length === 7 && segments[5] === 'view';
+/**
+ * True when a board-route segment is a legacy numeric ID (`1`, `10`) rather
+ * than a name slug (`original`, `12x12-square`). Canonical definition lives
+ * here (edge-safe, importable from middleware); `url-utils.ts` re-exports it
+ * for the rest of the app.
+ */
+export function isNumericId(value: string): boolean {
+  return /^\d+$/.test(value);
 }
 
 export function isBoardCreatePath(pathname: string | null | undefined): boolean {
--- a/packages/web/app/lib/board-route-paths.ts
+++ b/packages/web/app/lib/board-route-paths.ts
@@ -18,6 +18,16 @@ function getPathSegments(pathname: string): string[] {
   return segments;
 }
 
+/**
+ * Locale-stripped path segments for a board route. Exposed so the Expo-web
+ * rollout redirect map (edge middleware) can decompose a board URL into its
+ * board-config parts without re-implementing the `/es/…`, `/fr/…` prefix
+ * handling that classification already gets right.
+ */
+export function getBoardRouteSegments(pathname: string): string[] {
+  return getPathSegments(pathname);
+}
+
 export function isBoardRoutePath(pathname: string | null | undefined): boolean {
   if (!pathname) return false;
 
@@ -36,6 +46,20 @@ export function isBoardListPath(pathname: string | null | undefined): boolean {
   }
 
   return segments.length === 6 && segments[5] === 'list';
+}
+
+export function isBoardViewPath(pathname: string | null | undefined): boolean {
+  if (!pathname || !isBoardRoutePath(pathname)) return false;
+
+  const segments = getPathSegments(pathname);
+
+  // Slug form: /b/[board_slug]/[angle]/view/[climb_uuid]
+  if (segments[0] === 'b') {
+    return segments.length === 5 && segments[3] === 'view';
+  }
+
+  // Legacy form: /[board]/[layout]/[size]/[sets]/[angle]/view/[climb_uuid]
+  return segments.length === 7 && segments[5] === 'view';
 }
 
 export function isBoardCreatePath(pathname: string | null | undefined): boolean {
--- a/packages/web/app/lib/expo-web-rollout.ts
+++ b/packages/web/app/lib/expo-web-rollout.ts
@@ -1,0 +1,153 @@
+/**
+ * Expo-web rollout: the flag-gated per-surface redirect map that routes
+ * logged-in web visitors from migrated board surfaces to the `/app` Expo-web
+ * SPA. Everything here is edge-safe (no `server-only`, no `next/headers`) so
+ * `middleware.ts` can import it, plus a pair of SSR-guarded `document.cookie`
+ * writers used by the client hook that mirrors the PostHog flag into a cookie.
+ *
+ * Off in every dimension by default: no `BOARDSESH_WEB=1`, no `bs_expo_web`
+ * cookie (the flag hasn't resolved true for this visitor), or no auth session
+ * cookie all mean "no redirect, classic UI". The `expo-web-app` PostHog flag
+ * (id 767179, project 412845) is the master switch; `?classic=1` / the
+ * `bs_classic` cookie is the always-available escape hatch back to classic.
+ *
+ * Why a cookie and not a server-side flag eval: the web app evaluates PostHog
+ * flags only on the client (posthog-js-lite, read by FeatureFlagsProvider).
+ * Middleware runs on the edge and cannot call PostHog, so the client mirrors
+ * the resolved `expo-web-app` value into the non-HttpOnly `bs_expo_web` cookie
+ * (see ExpoWebRolloutCookieSync) and the edge reads that. A visitor whose flag
+ * hasn't resolved yet simply has no cookie and stays on the classic UI — the
+ * safe default.
+ */
+import type { NextRequest } from 'next/server';
+import { getBoardRouteSegments, isBoardRoutePath } from './board-route-paths';
+
+/** PostHog flag key (id 767179, project 412845). Master switch for the rollout. */
+export const EXPO_WEB_FLAG = 'expo-web-app';
+
+/** Client-mirrored flag cookie the edge reads. `1` = flag resolved true. */
+export const EXPO_WEB_ENABLED_COOKIE = 'bs_expo_web';
+
+/** Escape-hatch cookie: forces classic UI even when the flag is on. */
+export const EXPO_WEB_CLASSIC_COOKIE = 'bs_classic';
+
+/** Escape-hatch query param: `?classic=1` pins the `bs_classic` cookie. */
+export const EXPO_WEB_CLASSIC_PARAM = 'classic';
+
+/** Mount point of the Expo-web SPA (matches app.config `baseUrl: '/app'`). */
+export const EXPO_WEB_APP_BASE = '/app';
+
+const EXPO_WEB_CLIMBS_PATH = `${EXPO_WEB_APP_BASE}/climbs`;
+
+// next-auth writes the secure-prefixed cookie under HTTPS and the bare name
+// otherwise (see lib/auth/secure-cookies.ts). Mirror getServerAuthToken and
+// accept either — presence is the login heuristic, same as the rest of the app.
+const NEXT_AUTH_SESSION_COOKIES = ['__Secure-next-auth.session-token', 'next-auth.session-token'] as const;
+
+const ONE_YEAR_SECONDS = 60 * 60 * 24 * 365;
+
+/**
+ * Master env gate. The redirect never fires unless the site is built with the
+ * Expo-web surface available (`BOARDSESH_WEB=1`). Note this is intentionally
+ * looser than `isExpoWebProxyEnabled()` in middleware: production serves the
+ * static export with no `BOARDSESH_EXPO_WEB_ORIGIN`, so the rollout must not
+ * also require the dev proxy origin.
+ */
+export function isExpoWebRolloutEnabled(): boolean {
+  return process.env.BOARDSESH_WEB === '1';
+}
+
+/** True when the request carries a next-auth session cookie (logged in). */
+export function hasAuthSessionCookie(request: NextRequest): boolean {
+  return NEXT_AUTH_SESSION_COOKIES.some((name) => Boolean(request.cookies.get(name)?.value));
+}
+
+/** True when the visitor's mirrored `expo-web-app` flag cookie is on. */
+export function isExpoWebFlagCookieOn(request: NextRequest): boolean {
+  return request.cookies.get(EXPO_WEB_ENABLED_COOKIE)?.value === '1';
+}
+
+/** True when the visitor opted back into classic via the `bs_classic` cookie. */
+export function hasClassicOptOutCookie(request: NextRequest): boolean {
+  return request.cookies.get(EXPO_WEB_CLASSIC_COOKIE)?.value === '1';
+}
+
+function buildClimbsTarget(params: Record<string, string>): string {
+  const query = new URLSearchParams(params).toString();
+  return query ? `${EXPO_WEB_CLIMBS_PATH}?${query}` : EXPO_WEB_CLIMBS_PATH;
+}
+
+function buildClimbTarget(climbUuid: string, params: Record<string, string>): string {
+  const query = new URLSearchParams(params).toString();
+  const base = `${EXPO_WEB_CLIMBS_PATH}/${encodeURIComponent(climbUuid)}`;
+  return query ? `${base}?${query}` : base;
+}
+
+/**
+ * Per-surface redirect map. Returns the `/app` SPA path (with a board-context
+ * query string) for a migrated board surface, or `null` for everything else —
+ * public/SEO routes (profiles, playlists, gyms, setters), board create/queue,
+ * API routes, and anything unrecognised all fall through to the classic UI.
+ *
+ * Migrated surfaces (Expo Router URL shapes; the SPA is client-routed under
+ * `/app`, so a plain path redirect resolves to `index.html` and the router
+ * takes over):
+ *
+ *   Board list -> `/app/climbs` (the Climbs tab)
+ *     legacy  /[board]/[layout]/[size]/[sets]/[angle]/list
+ *     slug    /b/[slug]/[angle]/list
+ *   Climb view -> `/app/climbs/[uuid]` (the ClimbDetail deep-link route, which
+ *   opens the play drawer — the native app's own deep-link target for a climb)
+ *     legacy  /[board]/[layout]/[size]/[sets]/[angle]/view/[uuid]
+ *     slug    /b/[slug]/[angle]/view/[uuid]
+ *
+ * Board context rides the query string. The legacy form carries the exact
+ * param contract the mobile ClimbDetail route reads (boardName/layoutId/sizeId/
+ * setIds/angle); the slug form carries boardSlug + angle (the SPA resolves the
+ * slug). Locale prefixes (`/es/…`, `/fr/…`) are stripped — `/app` is
+ * locale-neutral.
+ */
+export function mapToExpoWebTarget(pathname: string): string | null {
+  if (!isBoardRoutePath(pathname)) return null;
+
+  const segments = getBoardRouteSegments(pathname);
+
+  if (segments[0] === 'b') {
+    const [, boardSlug, angle, surface, climbUuid] = segments;
+    if (!boardSlug || !angle) return null;
+    const params = { boardSlug, angle };
+    if (surface === 'list' && segments.length === 4) {
+      return buildClimbsTarget(params);
+    }
+    if (surface === 'view' && segments.length === 5 && climbUuid) {
+      return buildClimbTarget(climbUuid, params);
+    }
+    return null;
+  }
+
+  const [boardName, layoutId, sizeId, setIds, angle, surface, climbUuid] = segments;
+  if (!boardName || !layoutId || !sizeId || !setIds || !angle) return null;
+  const params = { boardName, layoutId, sizeId, setIds, angle };
+  if (surface === 'list' && segments.length === 6) {
+    return buildClimbsTarget(params);
+  }
+  if (surface === 'view' && segments.length === 7 && climbUuid) {
+    return buildClimbTarget(climbUuid, params);
+  }
+  return null;
+}
+
+/**
+ * Client-side mirror of the resolved `expo-web-app` flag into the `bs_expo_web`
+ * cookie the edge reads. `true` writes `bs_expo_web=1`; `false` clears it so a
+ * flag flipped back off (or a user removed from the rollout cohort) instantly
+ * stops redirecting. SSR-guarded — a no-op on the server.
+ */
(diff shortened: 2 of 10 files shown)


marcodejongh:
Superseded by #3766 — the rollout was reworked from the `/app` path approach to a **subdomain** (`app.boardsesh.com`) with a shared `.boardsesh.com` cookie. This PR's serving/redirect approach is no longer the plan; the reusable pieces (baseUrl config, CORS, CTA, PR5-map removal) were carried into #3766. Leaving open for reference — close if you prefer.

packages/web/app/lib/expo-web-rollout.ts
```diff
@@ -0,0 +1,153 @@
...
+  const segments = getBoardRouteSegments(pathname);
+
+  if (segments[0] === 'b') {
+    const [, boardSlug, angle, surface, climbUuid] = segments;
+    if (!boardSlug || !angle) return null;
+    const params = { boardSlug, angle };
+    if (surface === 'list' && segments.length === 4) {
+      return buildClimbsTarget(params);
+    }
+    if (surface === 'view' && segments.length === 5 && climbUuid) {
```
marcodejongh: **[high]** Slug climb-view redirect targets a param contract the /app SPA does not implement

`/b/[slug]/[angle]/view/[uuid]` redirects to `/app/climbs/[uuid]?boardSlug=...&angle=...`, but the Expo ClimbDetail route (`packages/mobile/app/(tabs)/climbs/[climbUuid].tsx`) reads only `boardName/layoutId/sizeId/setIds/angle` — there is no `boardSlug` handling anywhere in `packages/mobile/` (the only `boardSlug` references are BLE config matching), and the static `/app` export has no server-side resolution either. So `hasRequiredParams` is false, `useClimb` never runs (`enabled: false`, `isLoading` stays false), and a flagged logged-in user who hits a slug climb-view URL lands on the permanent "not found" screen. The doc's claim that the slug form "relies on the SPA to resolve the slug" describes code that doesn't exist, and its "verify the slug-view path resolves before widening" note can only ever fail. Either add `boardSlug` resolution to ClimbDetail, or have the slug-view branch return `null` (fall through to classic) until the SPA supports it — the map shouldn't ship a redirect to a known-dead target. (The legacy-numeric view form and both list forms are unaffected in the sense that this specific hard break is unique to slug view.)

marcodejongh (commented):

Multi-agent Claude review run by Marco across the whole Expo-web stack (#3752 / #3760 / #3761 / #3762 / #3764 / #3765 / #3766). Each PR was reviewed only on its own commits (diff vs its stack parent) by lens-scoped reviewers, and every finding below survived adversarial verification by independent refuter agents before posting. **9 findings** for this PR.

packages/web/app/lib/expo-web-rollout.ts
```diff
@@ -0,0 +1,153 @@
...
+      return buildClimbTarget(climbUuid, params);
```
marcodejongh: **[high]** Slug view redirect uses boardSlug param the SPA never reads: dead-end screen

Slug view redirect targets a screen that can't render: `/b/[slug]/[angle]/view/[uuid]` → `/app/climbs/[uuid]?boardSlug=…&angle=…` (line 123), but nothing in `packages/mobile` reads a `boardSlug` search param (only BLE config types match a grep). The `ClimbDetail` route requires `boardName && layoutId && sizeId && setIds && angle`; with only `boardSlug`+`angle` that check fails, so `useClimb` stays disabled (`enabled: !!variables`), `isLoading` is false, and the route permanently shows the `mobile.detail.notFound` block. `docs/expo-web-rollout.md` acknowledges the slug form "relies on the SPA to resolve the slug" — but that resolution doesn't exist yet, and `middleware.test.ts` codifies the redirect to the broken target. The blast radius is bounded (redirect only fires for logged-in, flag-cookied visitors on `BOARDSESH_WEB=1` builds, and the rollout doc says to verify this path before widening), but once the flag turns on, every flagged user following a slug view URL dead-ends. Either resolve the slug to the numeric board config in the map, or return `null` for slug view URLs until the SPA implements `boardSlug` resolution.

packages/web/app/lib/expo-web-rollout.ts
```diff
@@ -0,0 +1,153 @@
...
+    }
+    return null;
+  }
+
+  const [boardName, layoutId, sizeId, setIds, angle, surface, climbUuid] = segments;
```
marcodejongh, line 164: **[high]** Redirect map sends broken board context for the canonical slug URLs real users have

The legacy branch of `mapToExpoWebTarget` assumes `/[board]/[layout]/[size]/[sets]/[angle]/...` segments are numeric IDs, but the canonical classic URLs are name slugs: the web list/view pages `permanentRedirect` (308) numeric URLs to `constructClimbListWithSlugs` output — e.g. `/kilter/original/12x12-square/screw_bolt/40/list`, exactly the `LEGACY_LIST` fixture in `middleware.test.ts`. For those URLs this map emits `?layoutId=original&sizeId=12x12-square&setIds=screw_bolt`, and the mobile ClimbDetail route (`packages/mobile/app/(tabs)/climbs/[climbUuid].tsx`) does `Number(layoutId)` → `NaN`, so `useClimb` never resolves and the user lands on the not-found screen. The `/b/[slug]/...` view form is equally dead today: it emits `boardSlug`+`angle`, but no mobile route reads `boardSlug` — ClimbDetail's `hasRequiredParams` is false and `useClimb(null)` is disabled, so it renders not-found immediately. Net effect once the flag ramps: a flagged logged-in user redirected off any canonical climb-view URL hits a broken screen. (The list redirect degrades gracefully — the Climbs tab ignores the query params and uses the active board — so the hard dead-end is the view→ClimbDetail path.) The middleware test masks this by asserting only `boardName`/`angle` on the slug-form fixture.

Fix: gate the nested-form branch on `isNumericId` (url-utils.ts) for layout/size/sets and return `null` for the named form (or resolve names→IDs), and don't map `/b/` view URLs until the SPA actually implements `boardSlug` resolution.

packages/web/app/lib/expo-web-rollout.ts
```diff
@@ -0,0 +1,153 @@
...
+  if (!boardName || !layoutId || !sizeId || !setIds || !angle) return null;
+  const params = { boardName, layoutId, sizeId, setIds, angle };
+  if (surface === 'list' && segments.length === 6) {
+    return buildClimbsTarget(params);
+  }
+  if (surface === 'view' && segments.length === 7 && climbUuid) {
+    return buildClimbTarget(climbUuid, params);
```
marcodejongh: **[high]** Named-segment board URLs forwarded as numeric IDs; SPA gets layoutId=NaN

The "legacy" branch of `mapToExpoWebTarget` matches **any** `/[board]/x/y/z/[angle]/list|view` path and forwards the raw segments as `layoutId`/`sizeId`/`setIds` — but the web's canonical modern URLs use *named* slug segments (e.g. `/kilter/original/12x12-square/screw_bolt/40/view/<uuid>`, exactly what `constructClimbViewUrlWithSlugs` generates and what this PR's own `LEGACY_LIST` test fixture uses). The mobile ClimbDetail route (`packages/mobile/app/(tabs)/climbs/[climbUuid].tsx`) does `Number(layoutId)` / `Number(sizeId)`; `Number('original')` is `NaN`, so `useClimb` can never resolve and a flagged user deep-linked from a slug-form climb URL lands on the "not found" screen instead of the play drawer. The doc comment above the function ("carries the exact param contract the mobile ClimbDetail route reads") is only true for the numeric form. The tests mask this: the middleware test redirects the named-form URL but asserts only `boardName` and `angle`, while the unit test uses the numeric form. Since the redirect is flag-gated there's no impact until `expo-web-app` ramps — but this must be fixed before it does. Fix: gate the legacy branch on fully-numeric segments (reuse the numeric check behind `hasOnlyNumericBoardRouteSegments` in `app/lib/url-utils.ts`) and return `null` (stay classic) for named/mixed forms, or resolve slugs to IDs before building the target.

packages/web/app/lib/expo-web-rollout.ts
```diff
@@ -0,0 +1,153 @@
...
+      return buildClimbTarget(climbUuid, params);
+    }
+    return null;
+  }
+
+  const [boardName, layoutId, sizeId, setIds, angle, surface, climbUuid] = segments;
+  if (!boardName || !layoutId || !sizeId || !setIds || !angle) return null;
+  const params = { boardName, layoutId, sizeId, setIds, angle };
+  if (surface === 'list' && segments.length === 6) {
+    return buildClimbsTarget(params);
```
marcodejongh: **[medium]** List redirect's board params are ignored by the SPA; original query dropped

Two losses on the board-list redirect:

1. The climbs tab (`packages/mobile/app/(tabs)/climbs/index.tsx`) takes its board from the persisted `activeBoard` store; its only `useLocalSearchParams` reads screenshot params. So the `?boardName&layoutId&sizeId&setIds&angle` query built here is dead weight — a flagged visitor clicking a shared Tension list link lands on whatever board is in *their* store (or the empty first-run state), silently losing the link's board context. `?boardSlug` is worse: nothing in `packages/mobile/` consumes it at all (the doc's "the SPA resolves the slug" has no implementation), so the slug-form redirects — including slug climb-view, where ClimbDetail requires all five numeric params — carry context nothing reads.

2. `middleware.ts` builds the redirect from `pathname` only (`mapToExpoWebTarget(pathname)` + `new URL(expoWebTarget, request.url)`), so the classic list URL's own filter params (`minGrade`, `maxGrade`, `name`, `sortBy`, …) are discarded outright.

Either teach the destination routes to hydrate from these params (and forward the original query), or restrict the map to surfaces whose target actually consumes the context — currently only the legacy-numeric climb-view path.

packages/web/app/lib/expo-web-rollout.ts
```diff
@@ -0,0 +1,153 @@
...
+/**
+ * Client-side mirror of the resolved `expo-web-app` flag into the `bs_expo_web`
+ * cookie the edge reads. `true` writes `bs_expo_web=1`; `false` clears it so a
+ * flag flipped back off (or a user removed from the rollout cohort) instantly
+ * stops redirecting. SSR-guarded — a no-op on the server.
+ */
+export function setExpoWebEnabledCookie(enabled: boolean): void {
+  if (typeof document === 'undefined') return;
+  if (enabled) {
+    document.cookie = `${EXPO_WEB_ENABLED_COOKIE}=1; path=/; SameSite=Lax; max-age=${ONE_YEAR_SECONDS}`;
```
marcodejongh: **[medium]** Flag-off rollback never reaches users who only enter via migrated surfaces

`bs_expo_web` is written with a 1-year `max-age` (line 149) and is only ever cleared by `ExpoWebRolloutCookieSync`, which mounts in the classic Next.js root layout. The `/app` Expo SPA never touches the cookie, and middleware redirects on cookie presence alone. So after the PostHog flag is set to 0%, a logged-in user whose entry points are all migrated surfaces (bookmarked/shared board list or climb view URLs — the app's main entry URLs) keeps getting 307'd into `/app` until they happen to load a non-migrated classic page (homepage, profile, etc.), use `?classic=1`, or the cookie expires — worst case a year. That undercuts the rollback promise in `docs/expo-web-rollout.md` ("Instant, everyone ... No deploy") for exactly the cohort a rollback most needs to reach, since they're stuck in the surface being rolled back. Consider a short cookie TTL (hours, refreshed on each classic page load), having the `/app` SPA also mirror/clear the cookie, or at minimum documenting that flag-off rollback only reaches these users lazily and the guaranteed levers are `?classic=1` or shipping without `BOARDSESH_WEB=1`.

packages/web/app/lib/board-route-paths.ts
```diff
@@ -38,6 +48,20 @@ export function isBoardListPath(pathname: string | null | undefined): boolean {
   return segments.length === 6 && segments[5] === 'list';
 }
 
+export function isBoardViewPath(pathname: string | null | undefined): boolean {
```
marcodejongh: **[low]** isBoardViewPath is exported but never used

`isBoardViewPath` is added but nothing imports it — no caller, and unlike its siblings it isn't even covered by `board-route-paths.test.ts`. `mapToExpoWebTarget` checks the same view-URL shapes inline (`surface === 'view' && segments.length === 5/7`), though it can't fully delegate to a boolean predicate since it also extracts params and requires a non-empty climb UUID. Per the repo rule ("Remove dead code as you go"), drop the function until a caller lands (or wire the redirect map's shape check through it).

packages/web/app/lib/expo-web-rollout.ts
```diff
@@ -0,0 +1,153 @@
...
+    }
+    if (surface === 'view' && segments.length === 5 && climbUuid) {
+      return buildClimbTarget(climbUuid, params);
+    }
+    return null;
+  }
+
+  const [boardName, layoutId, sizeId, setIds, angle, surface, climbUuid] = segments;
+  if (!boardName || !layoutId || !sizeId || !setIds || !angle) return null;
+  const params = { boardName, layoutId, sizeId, setIds, angle };
```
marcodejongh: **[low]** Board-context query params on the /app/climbs list redirect are ignored by the SPA

The list redirect carries `?boardName&layoutId&sizeId&setIds&angle` (or `boardSlug&angle`), but the Climbs tab (`packages/mobile/app/(tabs)/climbs/index.tsx`) reads only `screenshot*` search params — board context comes from the user's `activeBoard` (`useActiveBoard()`). A flagged user browsing board X on web lands on the climbs list for whatever their active board happens to be, and the doc's "Board context carried" table implies more than actually happens. (The climb-view redirect is fine — `[climbUuid].tsx` does read the legacy params — though note it has no `boardSlug` handling either, so the slug-form view redirect's params are also dead.) Either wire the params into the climbs tab (select/prompt the matching board) or drop them and document that the list redirect lands on the user's active board.

packages/web/middleware.ts
```diff
@@ -148,6 +157,39 @@ export function middleware(request: NextRequest) {
...
+      });
+      return classicResponse;
+    }
+
+    if (!hasClassicOptOutCookie(request) && isExpoWebFlagCookieOn(request) && hasAuthSessionCookie(request)) {
+      const expoWebTarget = mapToExpoWebTarget(pathname);
+      if (expoWebTarget) {
+        // 307 (not 308): the flag can flip off at any time, so the redirect
+        // must never be cached by the browser as permanent.
+        return NextResponse.redirect(new URL(expoWebTarget, request.url), 307);
```
marcodejongh, line 206: **[low]** Original query string is dropped on the /app redirect

The redirect builds its URL solely from `mapToExpoWebTarget(pathname)`, discarding `request.nextUrl.search`. Classic list URLs legitimately carry filter/search params (`?minGrade=...&name=...&sortBy=...`), so a flagged user opening a shared filtered list link lands on `/app/climbs` with the filters silently gone. Note the Expo SPA's climbs route doesn't read filters from the URL, so *merging* params into the target wouldn't help — instead either skip the redirect when the incoming URL carries search params (let the classic page serve the shared link), or document the intentional loss in `docs/expo-web-rollout.md`.

packages/web/app/lib/expo-web-rollout.ts
marcodejongh: Fixed in b78c19bd7: slug climb-view URLs now return null from the map (stay classic) — no redirect until the SPA actually implements boardSlug resolution.

packages/web/app/lib/expo-web-rollout.ts
marcodejongh: Fixed in b78c19bd7: the slug-view branch was dropped from the map; /b/[slug]/[angle]/view/[uuid] stays classic. The middleware test now pins the classic fall-through instead of codifying the broken target, and docs/expo-web-rollout.md no longer claims SPA-side slug resolution.

packages/web/app/lib/expo-web-rollout.ts
marcodejongh: Fixed in b78c19bd7: the view redirect is now gated on fully-numeric layout/size/sets/angle segments (isNumericId, moved to edge-safe board-route-paths.ts), so canonical named-slug URLs like /kilter/original/12x12-square/screw_bolt/40/view/... stay classic. List redirects emit bare /app/climbs with no dead params. Tests cover the named, mixed, and slug forms.

packages/web/app/lib/expo-web-rollout.ts
marcodejongh: Fixed in b78c19bd7: the legacy view branch requires fully-numeric board-config segments (hasNumericBoardConfigSegments, built on isNumericId) before redirecting; named/mixed forms return null and stay classic. The setIds segment is also decoded before splitting and forwarding. Unit + middleware tests pin the named-segment fall-through.

packages/web/app/lib/expo-web-rollout.ts
marcodejongh: Fixed in b78c19bd7, both losses: (1) list redirects now emit bare /app/climbs with no board-context params — the Climbs tab reads none, and the doc now says the redirect lands on the visitor's active board by design; (2) middleware skips the /app redirect entirely when the classic URL carries a query string, so shared filtered links are served classic instead of silently dropping their filters.

packages/web/app/lib/expo-web-rollout.ts
marcodejongh: Fixed in 3856ae45d: bs_expo_web now has a 4-hour TTL, refreshed on every classic-side navigation (ExpoWebRolloutCookieSync gained a pathname dep). A visitor living entirely inside /app loses the cookie within the TTL, falls through to classic on the next migrated-surface hit, and re-evaluates the flag — bounding flag-off rollback lag to hours instead of a year. Docs updated in c136c8070.

packages/web/app/lib/board-route-paths.ts
marcodejongh: Fixed in b78c19bd7: removed the dead isBoardViewPath export. board-route-paths.ts instead gained isNumericId (moved from url-utils, which re-exports it) so the edge redirect map can gate the view redirect on numeric segments.

packages/web/app/lib/expo-web-rollout.ts
marcodejongh: Fixed in b78c19bd7: the list redirect no longer carries any board-context query — bare /app/climbs — and docs/expo-web-rollout.md documents that it lands on the visitor's active board until the SPA consumes board params. The slug-form view redirect was dropped entirely (stays classic).

packages/web/middleware.ts
marcodejongh: Fixed in b78c19bd7: middleware now skips the /app redirect whenever the incoming classic URL has a query string, so filtered/shared list links are served classic rather than losing their params. Covered by a middleware test and documented in docs/expo-web-rollout.md.

marcodejongh:
Follow-up on the claude[bot] review (comment posted against the original head 80c02b689), checked against the current code after the b78c19bd7/3856ae45d rework:

1. **Undeclared `originalExpoWebFlag` in middleware.test.ts** — already resolved at c136c8070: the variable is declared at module scope (`middleware.test.ts:17-18`, alongside `originalExpoWebOrigin`) and restored in both `afterEach` blocks. No undeclared variant remains.
2. **`?classic=1` intercepts API routes** — fixed in e1da4877b: the classic-param branch now has a `!pathname.startsWith('/api/')` guard (the matcher explicitly includes `/api/v1/:path*`, so this was reachable). The b78c19bd7 query-string skip did *not* cover this — the classic branch runs before it. Test pins that `/api/v1/kilter/proxy/login?classic=1` passes through with no 307 and no cookie.
3. **Rollout cookies missing `Secure`** — fixed in e1da4877b for both writers: `bs_classic` (middleware) uses `secure: isSecureCookieContext()`, mirroring the next-auth session cookie logic in `secure-cookies.ts`; the client-written `bs_expo_web` (including the 4h-TTL refresh from 3856ae45d) appends `Secure` when `window.location.protocol === 'https:'`. Plain http localhost keeps working in dev. Test asserts `Secure` on `bs_classic` when `NEXTAUTH_URL` is https.
4. **No test for the `__Secure-next-auth.session-token` variant** — fixed in e1da4877b: new middleware test logs in via the secure-prefixed cookie name and asserts the `/app/climbs/[uuid]` redirect fires.

Validation: `vp check --fix` clean on the three touched files; `middleware.test.ts` + `expo-web-rollout.test.ts` pass locally (109 tests).

marcodejongh:
## Claude: ready to merge ✅

All review findings on this PR are resolved at head `e1da4877b`: the multi-agent swarm review, follow-up Codex and claude[bot] batches, and QA verification passes are addressed (fixed, refuted with evidence, or filed as follow-up issues). CI is green with no unresolved threads.

marcodejongh:

PR #3763, opened 2026-07-17
fix(mobile): re-land #3740/#3733/#3742 and fix the bottom-chrome OTA crash

marcodejongh:
## Summary

Re-lands the three PRs that were emergency-reverted in #3756 (`revert/ota-crash-3740-3733-3742`) and fixes the actual OTA crash so they can ship safely.

**Root cause of the crash — #3740's provider hoist.** #3740 hoisted the bottom-chrome geometry into one `BottomChromeMetricsProvider`, but mounted it *inside* `DrawerHostProvider`. `DrawerHostProvider` renders the queue-added and undo-wall-change snackbars as siblings of its children — and both call `useBottomChromeMetrics()`. Those snackbars sat **outside** the provider, so the hook threw `useBottomChromeMetrics must be used within BottomChromeMetricsProvider` the moment a climb was queued (or a wall change happened) — white-screening every install that took the production OTA. #3733 (error hygiene) and #3742 (queue-item resolution) were innocent; they were bundled into the revert defensively.

**The fix**
- Hoist `BottomChromeMetricsProvider` to wrap `DrawerHostProvider` so the snackbars read the shared value. All of the provider's input hooks (queue presence, theme, device layout, safe-area insets, router segments) still resolve above this higher mount.
- Turn `useBottomChromeMetrics()` into a hard error in dev/tests but a conservative fallback in release builds. A stray consumer outside the provider is a mount-tree bug worth failing loudly on in development, but it must never brick the whole app on an OTA again — production now degrades the chrome geometry instead of throwing.

**What's re-landed**
- **#3740** — bottom-chrome metrics computed once in a provider instead of ~25 times per render (perf; internal).
- **#3742** — partially-synced queue items resolve to their real climb instead of a permanent "Unknown Climb".
- **#3733** — mobile error-reporting / offline-sync error-classification hygiene (internal).

## Release Notes

- Queue climbs your crew adds now fill in their name, grade, and thumbnail on your phone instead of getting stuck on "Unknown Climb" — even when the climb wasn't already cached on your device.

## Test plan

- `vp test run --project mobile bottom-chrome-metrics-provider drawer-host-provider queue-provider-resolve-climbs` — 37 passing (provider fan-out + snackbar hosts + queue resolution).
- `vp test run --project offline-sync error-classification` — 40 passing.
- `vp check` on the changed files (lint + format + types) — clean.
- Verified the new tree in `app/_layout.tsx`: `BleControlSheetProvider > BottomChromeMetricsProvider > DrawerHostProvider > … > UserDrawerProvider`, so the snackbars are now inside the provider.
- JS/TS-only, no native deps or migrations — rides the OTA.


marcodejongh:
## Production blast radius for the crash this PR fixes — and a recovery-OTA gap

Sentry (org `boardsesh`) shows the bottom-chrome provider crash this PR re-lands + fixes is **live and still firing** on the shipped OTAs. Recording the blast radius here so the recovery is delivered, not just the code fix.

[**BOARDSESH-AM**](https://boardsesh.sentry.io/issues/7616405142/) — `useBottomChromeMetrics must be used within BottomChromeMetricsProvider`:

- **359 events / 47 users**, firstSeen **2026-07-17 02:29Z** (matches the #3740 OTA rollout), **lastSeen 2026-07-24 09:03Z — still firing**, substatus `ongoing`. Daily: a 199 spike on Jul-17, then a sustained 15–39/day tail.
- `environment=production` on all 359. Cross-platform: **iOS 146** (26.2/26.5/26.5.2) + **Android 151** (11/16/17).
- Culprit `useBottomChromeMetrics(main)`; top in-app frames `QueueAddedSnackbarGlass → useBottomChromeMetrics` (throw). `DrawerHostProvider` renders `<QueueAddedSnackbar>` unconditionally, and on the buggy build it sat *above* the provider, so the always-mounted snackbar read a null context → threw → Expo Router root `ErrorBoundary` → full-screen crash (`handled:yes` = caught by the JS boundary, not a native crash).
- Spread across **3 stuck OTA fingerprints**: runtime `49a581a48aae87150bcdddcd5dc0b3b4914b76ea` (release `2.2.2+3`, ota `cdc1fc21-…`, 194), `9c009489…` (89), `9ebca93545…` (76). Sample `ota_is_embedded=false` — these are users who **pulled** the buggy provider OTA.

**PostHog (project 412845):** the flow behind the crashing snackbar is heavily used on exactly these builds — `Climb Added to Queue` (7d): build `3` = 337 adds / 108 users; build `2000587` = 275 adds / 68 users. The 47 crashers are the `ota_is_embedded=false` subset that pulled the bad OTA (users still on the same builds' embedded pre-provider JS are unaffected — why 47 < 108).

## Ask: don't forget the recovery OTA per fingerprint

This branch's HEAD (`fa43550f6`) already fixes it two ways — mounts `BottomChromeMetricsProvider` above `DrawerHostProvider`, and `__DEV__`-gates the throw so release builds return a fallback instead of crashing. 👍 The one thing the merge alone won't do: **the 47 users are stuck on the three fingerprints above and only recover when a fresh production OTA reaches each one** (per the fingerprint-parity delivery constraint). Please confirm a recovery OTA publishes for `49a581a48…`, `9c009489…`, and `9ebca93545…` after this lands — otherwise that cohort stays bricked even though `main` is healthy.

---
*From a Sentry triage sweep (org `boardsesh`).*

PR #3764, opened 2026-07-17
refactor(mobile): dedup the .web auth/ws forks (E1)

marcodejongh:
## Summary

Deduplicates the copy-forked `.web` auth/ws twins (review finding E1). ~500 lines of security-sensitive credential-recovery logic were copy-pasted across `.web` twins; a future 4403/refresh fix had to land twice or silently leave browsers on the buggy variant. This collapses them to single shared sources with **zero behavior change** on native or web.

### Approach
The mobile vitest project has no `.web` suffix resolution and mocks `.web` modules explicitly, so a single bare-import shared file would break the web tests' mocks. Instead: **dependency injection into a shared core** — each thin platform wrapper keeps its explicit `.web` imports (mocks still apply) while the logic lives once. Zero test-file edits.

### Changes
- **ws-client**: the ~150-line duplicated recovery / 4401→4403-remap / connection_init-gate / singleton block now lives once in new `ws-client-core.ts` (`createWsClientModule`, DI'd with a `createSocket` factory + the auth helpers). `ws-client.ts` 194→32 lines; `ws-client.web.ts` 166→**16 lines**. The only platform difference — React Native's Origin-header suppression — is the injected `createSocket`.
- **Rejection type**: new `auth-rejection-result.ts` holds one shared `AuthRejectionResult` (superset of both platforms' status labels); removed `NativeAuthRejectionResult`/`WebAuthRejectionResult`. Each platform still only emits its own subset, so the shared `canRetry` check (`refreshed || authenticated || unavailable`) is behavior-identical on each.
- **auth.web.ts**: one typed `fetchJsonObject` helper replaces 3 copies of the fetch→parse→catch ladder (with an `onResponse` hook for the credentials-callback's `clearTokens()` fence).
- **auth-token-events**: `auth-store.web.ts` now `import type`s the change-source/listener types instead of re-declaring them.

### Left un-deduped
`queue-snapshot-store.web.ts` — the delta is genuine, not boilerplate: the web fork user-scopes storage keys per browser-tab owner (`userScopedStorageKey`) while native uses one global key. Kept separate.

## Verification
- `vp run typecheck:mobile`, `vp run test:mobile` (**523 files, 4214 tests pass** — the behavior contract; ws-client native/web, auth.web, auth-store suites unchanged), `vp run check:mobile-bundle` (native), `vp run check:mobile-web-bundle` (web), `vp check` clean on 8 touched files.
- **`vp run check:mobile-ota-compat` → ota-compatible** on iOS + Android (fingerprint identical to main; JS refactor, no native change).
- No changes to root `package.json`/`bun.lock`, `packages/mobile/package.json`, or the board-renderer files a parallel PR owns.

## Release Notes

- [x] No release note needed (internal refactor)

_Stacked on #3760 (smoke) → #3752 (Phase 0). When #3752/#3760 merge this rebases cleanly; the Phase-1 primitives PR (#3763) is file-disjoint and rebases onto this._


packages/mobile/src/lib/auth-store.web.ts
```diff
@@ -30,8 +31,10 @@ type WsAuthResponse = {
   authSessionId?: unknown;
 };
 
-export type AuthTokenChangeSource = 'local' | 'remote' | 'remote-signout' | 'session' | 'hint';
-export type AuthTokenChangeListener = (token: string | null, source: AuthTokenChangeSource) => void;
+// Single definition lives in auth-token-events (imported above); re-exported
+// here because this file (and its .web fork) is where web consumers import
+// these from.
+export type { AuthTokenChangeListener, AuthTokenChangeSource };
```
marcodejongh: **[low]** Type re-export cycles with auth-token-events.web.ts under web platform resolution

Under `.web`-suffix-first resolution these two files type-re-export each other with no concrete declaration. This file now imports `AuthTokenChangeListener`/`AuthTokenChangeSource` from `./auth-token-events`, while `auth-token-events.web.ts` re-exports those same types from `./auth-store.web`. With `.web`-first resolution, `./auth-token-events` resolves to `auth-token-events.web.ts`, so neither file in the web-resolved graph declares the types. It works today only because every `.web`-first tool (Metro/Babel, esbuild) erases type-only imports before resolution, and tsc (no `moduleSuffixes` anywhere in the repo) resolves to the native `auth-token-events.ts` that holds the real declarations. A `.web`-aware typecheck would fail to resolve these types. This cycle is new in this PR — at base, the concrete declarations lived here.

The added comment is also inaccurate: "this file (and its .web fork) is where web consumers import these from" — this file *is* the `.web` fork (`auth-store.ts` doesn't export these types), and the only importer of this re-export is `auth-token-events.web.ts` itself.

Suggested fix, following the `auth-rejection-result.ts` pattern this PR established: move the two type aliases into a small non-forked module (e.g. `auth-token-change-types.ts`), import them from there in `auth-token-events.ts`, `auth-token-events.web.ts`, and this file, and drop the re-export + comment here. One ground-truth declaration under every resolution strategy.

marcodejongh (commented):

Multi-agent Claude review run by Marco across the whole Expo-web stack (#3752 / #3760 / #3761 / #3762 / #3764 / #3765 / #3766). Each PR was reviewed only on its own commits (diff vs its stack parent) by lens-scoped reviewers, and every finding below survived adversarial verification by independent refuter agents before posting. **1 finding** for this PR.

packages/mobile/src/lib/auth-store.web.ts
marcodejongh: Fixed in 3ceeeb25e: moved `AuthTokenChangeListener`/`AuthTokenChangeSource` into a new non-forked `auth-token-change-types.ts` module (mirrors the `auth-rejection-result.ts` pattern), and both `auth-token-events.ts` and `auth-token-events.web.ts` import directly from it. `auth-store.web.ts` no longer re-exports these types or imports from `./auth-token-events`, so there's no cycle under .web-first resolution or tsc's native-first resolution. Runtime behavior unchanged. Validated with `vp check --fix`, a direct `tsc --noEmit` run against the mobile tsconfig, and `vp run check:mobile-bundle` (both ios and android bundles OK).

marcodejongh:
## Claude: ready to merge ✅

All review findings on this PR are resolved at head `3ceeeb25e`: the multi-agent swarm review, follow-up Codex and claude[bot] batches, and QA verification passes are addressed (fixed, refuted with evidence, or filed as follow-up issues). CI is green with no unresolved threads.

marcodejongh:

PR #3765, opened 2026-07-17
feat(mobile): Phase 1 web primitives — Web Bluetooth + render worker/cache

marcodejongh:
## Summary

Completes the Phase-1 web primitives so the climbing surfaces are functional (not stubbed) on the `/app` Expo-web app. Two file-disjoint workstreams; all changes are `.web`-only (native graph untouched, `ota-compatible` on both platforms).

### Real Web Bluetooth adapter
Replaces the `adapter-factory.web.ts` stub (`isAvailable: false`) with a working `WebBluetoothAdapter` implementing the mobile `BluetoothAdapter` interface, ported from the Next.js `packages/web/app/lib/ble/web-adapter.ts` onto the shared pure-bytes `@boardsesh/ble-protocol` transport (UUIDs / filters / `splitMessages` reused, not reimplemented).
- `requestAndConnect` → `navigator.bluetooth.requestDevice` with Aurora/MoonBoard service + name-prefix filters; `write` chunks via `writeValueWithoutResponse` with the MoonBoard with-response and Aurora unsupported-retry fallbacks; `onDisconnect` wires `gattserverdisconnected`.
- Web Bluetooth DOM types via a local ambient `web-bluetooth.d.ts` (no `@types/web-bluetooth` dep → no lockfile/fingerprint change).
- Non-Chromium (no `navigator.bluetooth`) → `isAvailable()` false → existing "Bluetooth unavailable" UI. 15 new tests.

### Board-render worker + Cache API persistence
- New `overlay-cache-store.web.ts`: Cache-API store (persist/read/hydrate/delete PNG bytes keyed by `cacheKey`) + the C2 object-URL retention map (document-lifetime, released on non-persisted pagehide, no in-use eviction — the C2 fix is preserved).
- `index.web.ts`: `renderHoldsOverlay` reads the Cache API on a miss (overlays survive a reload) and persists rendered bytes; render moved to a module Worker (`public/wasm/board-render.worker.js`, constructed by runtime URL so Metro doesn't bundle it) with a **transparent main-thread fallback** that stays authoritative for any worker failure.
- `overlay-cache-warmup.web.ts`: replaces the no-op stub; hydrates object URLs from the Cache API on startup (matching the native disk-warmup contract, incl. stale-version cleanup). 12 new tests.

## Verification
- `vp run typecheck:mobile`; `vp run test:mobile` (**526 files / 4241 tests**, incl. 27 new; the 8 existing `board-renderer.web` C2 tests unchanged); `vp run check:mobile-bundle` (native OK); `vp run check:mobile-web-bundle` (export + WASM + worker asset present); `vp check` clean on 13 files.
- **`vp run check:mobile-ota-compat` → ota-compatible** iOS + Android (fingerprint identical to main).
- No changes to root `package.json`/`bun.lock` or `packages/mobile/package.json`.

## Notes for review / QA
- **Worker execution wasn't exercised in a real browser** (none in the build env); the fake-worker integration test covers the wiring and the main-thread fallback guarantees rendering regardless. **QA on `/app` should confirm thumbnails render and survive a reload** (Cache API) and watch for worker console errors.
- **BLE needs a Chromium browser + a real board or BLE simulator** and a secure context (localhost/HTTPS) — QA the connect + send-to-wall path there.

## Release Notes

- [x] No release note needed (internal; `/app` is flag-gated off)

_Stacked on #3760 (smoke) → #3752 (Phase 0). File-disjoint from #3764 (dedup); rebases onto it cleanly._


marcodejongh (commented):

Multi-agent Claude review run by Marco across the whole Expo-web stack (#3752 / #3760 / #3761 / #3762 / #3764 / #3765 / #3766). Each PR was reviewed only on its own commits (diff vs its stack parent) by lens-scoped reviewers, and every finding below survived adversarial verification by independent refuter agents before posting. **6 findings** for this PR.

packages/mobile/modules/board-renderer/src/overlay-cache-store.web.ts
```diff
@@ -0,0 +1,227 @@
...
+    return await caches.open(OVERLAY_CACHE_NAME);
+  } catch {
+    // Storage disabled (private mode, blocked cookies) — degrade to in-memory
+    // only; the renderer still works, overlays just don't survive a reload.
+    return null;
+  }
+}
+
+/** Persist the rendered PNG bytes so the overlay survives a reload. Best-effort. */
+export async function writeOverlayToCache(cacheKey: string, pngBlob: Blob): Promise<void> {
```
marcodejongh, line 100: **[medium]** Persisted overlay cache has no eviction; quota failures silently stop persistence

Persisted overlay cache has no size cap; quota failures silently stop persistence. `writeOverlayToCache` adds one Cache API entry per rendered climb and nothing ever deletes current-version entries — the only deletion path is the shared hook's stale-RENDERER_VERSION sweep, which also only sees the hydrated tail (`OVERLAY_HYDRATE_LIMIT` = 120), so old-version bulk beyond that is never cleaned either. The limit bounds startup memory, not storage: a heavy user accumulates PNGs on the origin until browser quota pressure, and quota is shared with everything else on app.boardsesh.com (browser-initiated eviction is typically whole-origin, so leaning on it risks losing all origin storage at once). Once quota is hit, `cache.put` throws and the empty catch swallows it, so persistence silently stops (in-session rendering still works — this is graceful degradation, not breakage). Suggested fix: after `hydrateOverlayCache` enumerates `cache.keys()` (already in insertion order), delete entries beyond a cap in one pass — e.g. everything older than the newest N insertions.

packages/mobile/src/hooks/overlay-cache-warmup.web.ts
```diff
@@ -3,22 +3,34 @@
...
+  snapshotOverlayEntries,
+  type OverlayCacheEntry,
+} from '../../modules/board-renderer/src/overlay-cache-store.web';
+
+export type { OverlayCacheEntry };
+
+// Start reading the Cache API into the retention map as soon as this module is
+// imported (the shared hook imports it at app startup). Fire-and-forget — the
+// store owns the promise and dedups repeat calls.
+void hydrateOverlayCache();
```
marcodejongh: **[medium]** Warm-up races hydration: stale-version cleanup is dead code on web

Warm-up races hydration on web: stale-version cleanup is unreliable and structurally incomplete.

`hydrateOverlayCache()` is fired-and-forgotten at module import (line 31), but the shared consumer `warmupRenderedOverlaysOnce()` (use-native-climb-render.ts) runs synchronously in the first `useNativeClimbRender` render and self-guards with `warmupRun = true` — it executes exactly once per session and is never re-run when the hydration promise resolves. Whether it sees the hydrated entries is a startup race: hydration does `caches.open` + `keys()` + up to 120 `match`/`blob()` reads, and nothing orders it before the first hook render.

Consequences when the race is lost: (a) the sync warm-up of prior-session overlays misses (recoverable — `renderHoldsOverlay` falls back to `readOverlayFromCache`, so this is a first-paint perf nick, not breakage); (b) more importantly, the hook's stale-`RENDERER_VERSION` sweep is the *only* production path that calls `deleteOverlayFromCache`, so old-version PNGs get no app-level cleanup that session. Two parts are unconditional, not racy: `hydrateOverlayCache` mints URLs for keys of any version, so old-version entries burn slots in the 120-entry hydrate budget; and entries beyond that budget are never snapshot, so they are never swept even when the race is won.

The tests mask the ordering: `overlay-cache-warmup.web.test.ts` always `await hydrateOverlayCache()` before `listOverlayCacheEntries`, which production never guarantees.

Fix: do version filtering in the store — pass the current `v{RENDERER_VERSION}_` prefix into `hydrateOverlayCache` and skip + `cache.delete()` non-matching keys during hydration (this also fixes the budget and beyond-budget gaps), and/or re-run the warm-up when the hydration promise resolves.

packages/mobile/src/lib/ble/web-adapter.ts
```diff
@@ -0,0 +1,269 @@
...
+  signal: AbortSignal | undefined,
+  options: { allowWithResponseFallback: boolean; allowUnsupportedWithResponseRetry: boolean },
+): Promise<void> {
+  // Default to the proven write-without-response path. MoonBoard may choose
+  // write-with-response up front for the original RedBearLab box (its LED
+  // characteristic advertises `.write` only). Aurora keeps the decision
+  // behaviour-driven: it tries without-response first and only retries
+  // with-response when Web Bluetooth rejects the operation with
+  // NotSupportedError on a write-only characteristic.
+  const useWithResponse =
```
marcodejongh: **[medium]** Web Bluetooth transport helpers copied verbatim from packages/web instead of extracted

Web Bluetooth transport code duplicated from packages/web instead of extracted to a shared package.

`writeCharacteristicSeries`, `concatenateMessages`, `isWithoutResponseUnsupportedError`, `tryGetWriteCharacteristic`, `getUartCharacteristic`, `getMoonboardWriteCharacteristic`, and the Aurora/MoonBoard `RequestDeviceOptions` are near-verbatim copies (same logic and comments, minor signature/style tweaks) of `packages/web/app/components/board-bluetooth-control/bluetooth-shared.ts` / `bluetooth-aurora.ts` / `bluetooth-moonboard.ts`, and the `WebBluetoothAdapter` class mirrors `packages/web/app/lib/ble/web-adapter.ts` method-for-method. CLAUDE.md is explicit: "Extract, don't duplicate. When porting a web feature to mobile, pull the business logic into a shared package and update web to import from it in the same PR."

The header comment justifies the copy with "packages/web is not on the mobile graph", but that's the case the rule addresses: these helpers touch only Web Bluetooth ambient types — no React/Next/RN — so they fit a browser-side shared module (e.g. a `@boardsesh/ble-protocol/web-transport` subpath that also owns the ambient `web-bluetooth.d.ts`). Since `@boardsesh/ble-protocol` is already a mobile dependency, extraction adds no package.json entry and cannot perturb the OTA fingerprint (the stated reason for the local d.ts).

This matters concretely here: the duplicated write path is where behavioural hotfixes land (the NotSupportedError with-response retry and the MoonBoard RedBearLab fallback in this very function are past fixes) — the next one now has to be applied twice and the copies will drift. Extracting would also remove the mobile-local `web-bluetooth.d.ts`, whose unscoped global `Navigator`/`BluetoothDevice` declarations currently apply to the entire mobile TS project, native files included.

packages/mobile/src/lib/ble/web-adapter.ts
```diff
@@ -0,0 +1,269 @@
...
+    const device = await navigator.bluetooth.requestDevice(requestDeviceOptionsForFamily(this.scanFamily));
+
+    // Probing the write characteristic opens the GATT connection. If probing
+    // rejects (e.g. the UART service is absent on an Aurora variant) or comes
+    // back empty (a MoonBoard exposing neither controller), that connection is
+    // left open — drop it before surfacing the failure so we never leak a live
+    // GATT link the caller can't reach.
+    let characteristic: BluetoothRemoteGATTCharacteristic | undefined;
+    try {
+      characteristic =
```
marcodejongh, line 61: **[medium]** GATT connection leaked when the write-characteristic probe fails after connect

GATT connection leaked when the write-characteristic probe fails after connect.

`requestAndConnect` connects GATT inside the probe helpers (`getUartCharacteristic` / `getMoonboardWriteCharacteristic` call `device.gatt?.connect()`), then rejects (Aurora: `getPrimaryService` rejection propagates uncaught) or throws `'Failed to resolve the board write characteristic'` (MoonBoard: both probes return undefined) without ever disconnecting. Because `this.device` is only assigned after a successful probe, the adapter has no handle on the leaked connection: `disconnect()` checks the old/null `this.device` and no `gattserverdisconnected` listener is attached yet. The browser keeps the GATT link open until tab close — and since these boards accept a single BLE central, the physical board refuses other centrals (e.g. the official app) while the zombie connection lives. A retry from the same tab can recover (Chromium returns the cached `BluetoothDevice` for the same board, so a successful re-probe re-adopts the connection), but if the user gives up or the probe keeps failing (stale GATT cache — the exact `service_missing` class instrumented natively), the link is stuck open.

Fix: wrap the probe in try/catch and disconnect before rethrowing:
```ts
let characteristic;
try {
  characteristic = this.scanFamily === 'moonboard' ? await getMoonboardWriteCharacteristic(device) : await getUartCharacteristic(device);
  if (!characteristic) throw new Error('Failed to resolve the board write characteristic');
} catch (error) {
  if (device.gatt?.connected) device.gatt.disconnect();
  throw error;
}
```
Note: faithfully ported from `packages/web/app/lib/ble/web-adapter.ts`, which has the same flaw — worth fixing there too, but this PR is the chance to not replicate it.

packages/mobile/src/lib/ble/web-adapter.ts
```diff
@@ -0,0 +1,269 @@
...
+        this.scanFamily === 'moonboard'
+          ? await getMoonboardWriteCharacteristic(device)
+          : await getUartCharacteristic(device);
+    } catch (error) {
```
marcodejongh, line 65: **[medium]** GATT connection leaked when characteristic resolution fails in requestAndConnect

`getUartCharacteristic` / `getMoonboardWriteCharacteristic` call `device.gatt?.connect()` before probing services. If the probe fails (Aurora: `getPrimaryService` rejects and propagates; MoonBoard: both probes return `undefined` and `requestAndConnect` throws `Failed to resolve the board write characteristic`), the GATT link stays open with no owner: `this.device` is only assigned after a successful probe, so `adapter.disconnect()` is a no-op against this connection, and the caller's catch in `use-board-bluetooth.ts` never calls `connectAdapter.disconnect()` either. Since BLE board controllers accept one central and stop advertising while connected, the browser-held zombie connection blocks the board until the tab closes (a retry can only recover it if the probe happens to succeed the second time — a MoonBoard matched by name prefix that exposes neither controller service re-fails every attempt). Fix: wrap the probe (and the `!characteristic` throw) in try/catch in `requestAndConnect` and call `device.gatt?.disconnect()` before rethrowing.

packages/mobile/public/wasm/board-render.worker.js
```diff
@@ -0,0 +1,92 @@
...
+  }
+  return wasmInitPromise;
+}
+
+const MAX_RENDER_DIMENSION = 8192;
+const MAX_RENDER_PIXELS = 32 * 1024 * 1024;
+
+// Mirror of index.web.ts's decodeRenderOutput: first 8 bytes are width/height as
+// u32 LE, followed by RGBA pixels. Validates dimensions before allocating.
+function decodeRenderOutput(rawBytes) {
```
marcodejongh, line 51: **[low]** Worker duplicates decodeRenderOutput/limits with no lint, typecheck, or test coverage

`decodeRenderOutput`, `MAX_RENDER_DIMENSION`/`MAX_RENDER_PIXELS`, and the `ensureWasmInitialized` retry pattern are now maintained in two copies: this worker and `modules/board-renderer/src/index.web.ts` (the Next.js worker shares the same 8-byte u32-LE width/height header contract, but not the validation code). Because this file is a static asset outside the Metro graph, it is plain untyped JS that no tsconfig covers and no test executes (`overlay-cache.web.test.ts` stubs `Worker` with a fake; `mobile-web-bundle-check.sh` only asserts the file exists), and worker breakage degrades silently to the main-thread fallback by design — so drift between the two copies would never surface in CI, only as a quiet loss of the off-main-thread path. Minimum fix: add a parity test that loads this file and asserts the constants and header-decode behaviour against `_webRendererForTests.decodeRenderOutput`, so the copies can't drift unnoticed. (Generating the asset at build time from a shared TS source is the fuller fix if the copy step that places the wasm glue can also emit a bundled worker.)

packages/mobile/modules/board-renderer/src/overlay-cache-store.web.ts
marcodejongh: Confirmed addressed in c278ae0b1 (landed before this review round): `hydrateOverlayCache` now prunes the insertion-ordered head down to `OVERLAY_HYDRATE_LIMIT` after hydration, so the Cache API stays bounded across sessions. dd7a2bba5 extends this by also evicting stale-`RENDERER_VERSION` keys during hydration (see the warm-up thread), so old-version bulk beyond the hydrate tail is reclaimed too.

packages/mobile/src/hooks/overlay-cache-warmup.web.ts
marcodejongh: Fixed in dd7a2bba5: version filtering now happens in the store. `hydrateOverlayCache(currentVersionPrefix)` deletes every non-matching key during hydration, so stale-version cleanup is store-owned and race-free (no longer dependent on the hook warm-up winning the startup race), and only current-version overlays count toward the 120-entry hydrate budget — closing both the budget and beyond-budget gaps. `RENDERER_VERSION` moved to its own module so the warm-up can read the prefix without importing back from the hook. Also added `onOverlayCacheHydrated`: the hook re-runs its synchronous warm-up once hydration resolves so prior-session overlays land for first paint (native hydration is synchronous, so the re-run is a no-op there).

packages/mobile/src/lib/ble/web-adapter.ts
marcodejongh: Fixed in 7d96e41fc: extracted the Web Bluetooth transport (characteristic probing, chunked write series, request-device options, `isWebBluetoothAvailable`/`requestWebBluetoothDevice`) into a shared, renderer-agnostic `@boardsesh/ble-protocol/web-transport` module, imported from BOTH the Next.js web adapter and this mobile web adapter. The module declares its own minimal Web Bluetooth type surface and reaches `navigator.bluetooth` through `globalThis`, so it needs no lib.dom or `@types/web-bluetooth` (stays out of the OTA fingerprint). web `bluetooth-shared`/`bluetooth-aurora`/`bluetooth-moonboard` now re-export from it. This also let me delete the mobile-local `web-bluetooth.d.ts` whose unscoped global declarations applied to the whole mobile TS project.

packages/mobile/src/lib/ble/web-adapter.ts
marcodejongh: Fixed in 3d8278a82: the write-characteristic probe is now wrapped in try/catch in `requestAndConnect`, and `device.gatt?.disconnect()` is called before rethrowing — both when the Aurora probe rejects and when the MoonBoard probe returns undefined (the explicit "Failed to resolve the board write characteristic" path). No more leaked GATT link. Regression-covered by the new "disconnects the GATT server when the Aurora probe rejects" / "when MoonBoard exposes neither controller" tests in web-adapter.test.ts.

packages/mobile/src/lib/ble/web-adapter.ts
marcodejongh: Fixed in 3d8278a82 (same disconnect-on-failure change as the probe-reject case above): both the `getPrimaryService`-rejection path and the `!characteristic` MoonBoard path now call `device.gatt?.disconnect()` before surfacing the error, so the GATT connection opened by the probe is never left open with no owner.

packages/mobile/public/wasm/board-render.worker.js
marcodejongh: Fixed in 3a2acfbe8: added board-render-worker-parity.web.test.ts. It loads `decodeRenderOutput` + the render-limit constants straight from the static worker asset source (evaluated in isolation) and pins them to `_webRendererForTests.decodeRenderOutput` — asserting identical decode output and identical rejection messages for truncated/zero/oversized/mismatched inputs, plus the exact constant values. Now the two copies cannot drift unnoticed. (Kept the minimum-fix parity test rather than the build-time codegen route, since the asset copy step does not currently emit a bundled worker.)

packages/mobile/modules/board-renderer/src/index.web.ts
```diff
@@ -165,37 +175,102 @@ async function encodeRgbaToPngObjectUrl(
...
+    const id = ++workerRequestId;
+    let done = false;
+    const finish = (blob: Blob | null) => {
+      if (done) return;
+      done = true;
+      clearTimeout(timeout);
+      workerPending.delete(id);
+      resolve(blob);
+    };
+    const timeout = setTimeout(() => finish(null), WORKER_RENDER_TIMEOUT_MS);
```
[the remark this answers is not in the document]
marcodejongh: Fixed in c84374dab: the 8s timeout now disables the worker, not just this one render. Extracted `disableRenderWorker()` (terminate + drop the cached instance + set `workerDisabled` + settle in-flight requests to the main thread — previously only the `onerror` path did this) and call it from the timeout callback too. So the first timeout routes every subsequent cache miss straight to the main renderer instead of re-posting to the wedged worker and eating another full 8s each. Kept it module-lifetime as you suggested — a fresh page load re-creates the worker, no re-enable logic. Added a test in overlay-cache.web.test.ts (`terminates and disables the worker after a render times out, so later misses skip it`): a stuck fake Worker + fake timers, advance past 8s, assert `terminate()` fired and a second `renderPngBlob` never re-posts (postMessage stays at 1).

marcodejongh:
## Claude: ready to merge ✅

All review findings on this PR are resolved at head `c84374dab`: the multi-agent swarm review, follow-up Codex and claude[bot] batches, and QA verification passes are addressed (fixed, refuted with evidence, or filed as follow-up issues). CI is green with no unresolved threads.

marcodejongh:

Issue #3167, opened 2026-06-23
Mobile: remove @gorhom/bottom-sheet, replace with an in-house sheet


marcodejongh:
## Goal

Remove `@gorhom/bottom-sheet` from `packages/mobile` and replace it with an in-house sheet built directly on `react-native-reanimated` + `react-native-gesture-handler` (which we already depend on).

## Why

gorhom keeps costing us. The latest is a hard one: on a non-default Android **Display size** (fractional pixel ratio), a *closed* sheet's invisible backdrop kept `pointerEvents` and ate every touch — the whole app looked frozen at rest. Reported on Samsung S24/S25 and Pixel 10, reproduced on a Pixel 8. PR #3165 works around it; before that we carried a separate `patch-package` for the cold-start mount race.

The upstream fixes are stuck:
- Backdrop freeze: gorhom PR [#2342](https://github.com/gorhom/react-native-bottom-sheet/pull/2342) (`Math.floor(animatedIndex)`) is **open and unmerged for months**; issues [#807](https://github.com/gorhom/react-native-bottom-sheet/issues/807), #1882, #527, #566. Everyone is shipping local patches.
- Cold-start mount race: gorhom [#2680](https://github.com/gorhom/react-native-bottom-sheet/issues/2680) — also open, no fix.

We've now reached the point where we maintain our own fork of gorhom's backdrop (`SheetBackdrop.tsx`, 158 lines) *and* it still couples back into gorhom internals (`useBottomSheet`, the `animatedIndex` shared-value contract). That's the worst of both: we own the code but can't decouple from the library it patches. Time to own the sheet outright.

## Interim state (already shipped)

PR #3165 replaced gorhom's `BottomSheetBackdrop` with our `SheetBackdrop` across all 12 sheets that render a scrim. The touchability gate now tolerates the sub-pixel index drift (`animatedIndex > disappearsOnIndex + 0.05`) instead of gorhom's exact `<= disappearsOnIndex`. **This issue is the follow-through, not a re-do of that.**

## Current gorhom surface in `packages/mobile`

~45 production files import from `@gorhom/bottom-sheet`. Symbols in use:

| Symbol | Count | Replacement need |
| --- | --- | --- |
| `BottomSheetModal` | 15 | imperative present/dismiss sheet host |
| `BottomSheetTextInput` | 9 | keyboard-aware input inside a sheet (the sticky one) |
| `BottomSheetView` | 4 | content container |
[...]

PR #3771, opened 2026-07-19
fix(mobile): clear Android 3-button nav bar in bottom sheets

marcodejongh:
## Summary

On Android, the app runs mandatory edge-to-edge, so the system navigation bar draws *over* app content. On phones using the 3-button navigation bar (back/home/recents), that bar covers anything at the bottom of a sheet. Since the app moved to Expo's native `@expo/ui/community/bottom-sheet` (#3167), which has no `bottomInset` prop, each sheet has to add the bottom safe-area inset itself — and a handful of surfaces padded with a fixed value (or skipped the inset entirely), so their bottom controls sat under the nav bar and couldn't be tapped.

The reported case is the board-presence sheet's **"Switch board"** button; the same pattern recurred elsewhere. This brings the outliers in line with the app's existing convention (`insets.bottom + spacing[N]`, both platforms).

- **Board sheet footer** (`NowOnTheWallPanel.tsx`) — the "Switch board" footer no longer drops `insets.bottom` in sheet mode; it always adds it, matching the shared `Sheet`/`ModalSheet` footers. This is the reported bug.
- **Shared wrappers** (`Sheet.tsx`, `ModalSheet.tsx`) — a footerless sheet body now composes the bottom inset on top of the consumer's `contentContainerStyle` (new `withSheetBottomInset` helper). This fixes the whole class of footerless sheets (Feedback, Claim Gym, Session Edit, Climb Actions, Add to Playlist, and friends) at once and stops new sheets from regressing.
- **Queue list** (`QueueList.tsx`) — the list content padding now adds the nav-bar inset so the last row clears it when fully scrolled.

## Release Notes

- Buttons at the bottom of the board sheet and other panels no longer hide behind Android's on-screen navigation bar.

## Test plan

- `vp run test:mobile` — 4023 tests pass. Added a unit test for `withSheetBottomInset` (additive composition), a `Sheet` test asserting a footerless body gets the inset, and a `board-sheet` test asserting the "Switch board" footer includes a non-zero bottom inset. Updated the `NowOnTheWallPanel` test that previously encoded the old (buggy) behavior.
- `vp run check:mobile-bundle` — Metro bundle builds for iOS and Android.
- `vp check` — clean for the changed files.
- Not yet exercised on a physical 3-button-nav device / emulator; recommend a visual pass (board sheet → "Switch board" above the nav bar; queue sheet scrolled to the last row; a footerless sheet such as Feedback).

https://claude.ai/code/session_01MnKjXLbcv7xwgMjgd7ivsw

---
_Generated by [Claude Code](https://claude.ai/code/session_01MnKjXLbcv7xwgMjgd7ivsw)_

### DIFF
--- a/docs/mobile-sheets-vs-routes.md
+++ b/docs/mobile-sheets-vs-routes.md
@@ -195,6 +195,17 @@ Is it a secondary surface OVER the current screen, or its own full surface?
    in the tree shows it. Pick `ModalSheet` for on-demand surfaces (the common case); `Sheet` when
    the sheet's lifetime is bound to parent state.
 
+5. **Sheet content must clear the bottom safe area itself — the native `@expo/ui` sheet does
+   not.** The sheet draws under the system bottom inset on **both** platforms — the Android
+   edge-to-edge nav bar (~48dp 3-button bar / gesture pill) _and_ the iOS home indicator (~34pt) —
+   so a control at the bottom of a sheet needs `insets.bottom` added to its padding, or it sits
+   under the bar (Kilter/Tension users on Android 3-button nav hit this on the board sheet). The
+   shared `Sheet`/`ModalSheet` wrappers add it to their pinned `footer` and, for a **footerless**
+   body, automatically via `withSheetBottomInset` (composed on top of your `contentContainerStyle`).
+   A sheet built on the raw native primitive (its own `BottomSheetModal` + `BottomSheetFlatList`,
+   e.g. the board sheet / queue list) owns this itself: add `insets.bottom + spacing[N]`. Apply on
+   both platforms — `insets.bottom` is 0 when there's nothing to clear, so there's no double-inset.
+
 ## A latency footgun (any route)
 
 A modal route's present animation can't START until React commits the route's first frame. If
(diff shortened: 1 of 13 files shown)


marcodejongh:
On the iOS double-inset question: the native `@expo/ui` sheet does **not** clear the bottom safe area on iOS either, so applying `insets.bottom` on both platforms is correct — no double-inset, and a `Platform.OS === 'android'` guard would actually *reintroduce* the home-indicator overlap on iOS gesture-nav phones.

Evidence from the existing (production) convention:
- `Sheet.tsx` / `ModalSheet.tsx` footers already use `insets.bottom + spacing[3]` unconditionally on both platforms.
- `EndSessionSheet.tsx` and `InviteSheet.tsx` add `insets.bottom` unconditionally, and `EndSessionSheet` explicitly comments on the padding being "up to ~34pt on gesture-nav phones" — that ~34pt is the iOS home indicator. If the native sheet already cleared it on iOS, those long-shipped sheets would be visibly double-padded there; they aren't.

`withSheetBottomInset` just follows that same convention, and `insets.bottom` is `0` when there's nothing to clear.

Addressed in follow-up commits:
- Expanded the `withSheetBottomInset` doc comment with the both-platform rationale.
- Added a fifth hard rule to `docs/mobile-sheets-vs-routes.md` documenting that sheet content owns the bottom safe area (the native sheet doesn't supply it).
- Made the `board-sheet.test.tsx` Pressable-mock style flatten recursive.

I don't have a macOS runner in this environment to capture an iOS simulator screenshot, so the iOS side is argued from the code convention rather than an empirical screenshot — worth a visual sanity check on the `pr-3771` OTA preview on an iPhone before merge.

---

PR #3795, opened 2026-07-20
feat(ci): deploy production web to Railway alongside Vercel

### LINKED ISSUE #4787: Dockerfile.web comment still claims nothing in CI builds it
`Dockerfile.web:62-69` ends with:

> Guarded by `scripts/__tests__/dockerfile-web-no-expo-export.test.ts`, which pins this file by text because nothing in CI builds it.

That last clause stopped being true when #4779 landed the `docker-web` job — CI builds this file on every PR touching the image's inputs, and unconditionally on every push to `main`.

The text pin is still worth keeping, and its own header comment has already been updated (in #4779) to say why: a reintroduced Expo export step would **build perfectly well** and ship a second SPA copy at `www/app` with different env, which no green build would ever complain about. The build gate and the text pin cover different failure modes. Only the Dockerfile's stale justification needs the edit.

## What to do

Reword the trailing clause so it stops asserting CI doesn't build the file. Something like: *"Guarded by `scripts/__tests__/dockerfile-web-no-expo-export.test.ts`: `ci.yml`'s `docker-web` job catches an image that fails to build, and this text pin catches one that builds fine but reintroduces the retired export."*

**Owner:** `Dockerfile.web` is touched by #3798, so this belongs in that PR rather than a separate one — filing it here so it survives that PR's rebase. Deliberately not edited by #4779 to keep that PR out of #3798's diff.

marcodejongh:
## Summary

Adds the Railway half of the www deploy alongside the Vercel one, so the two can
run on the same commit while traffic moves and Vercel stays current as the
rollback afterwards.

**Merging this changes nothing about production except one extra image push.**
`build-web` starts publishing `ghcr.io/boardsesh/boardsesh-web` to GHCR on every
run; with `WEB_DEPLOY_TARGETS` unset, `deploy-web-railway` shows as skipped and
Vercel deploys exactly as it does today.

### What changed

- **Fail-closed Railway smoke config** — targeting Railway now requires both `RAILWAY_WEB_SERVICE_ID` and `RAILWAY_WEB_ORIGIN`, so Railway-only releases cannot report success after silently skipping their smoke.
- **`resolve-web-targets`** — a new `environment: Production` job that reads
  `WEB_DEPLOY_TARGETS` and publishes `web_vercel` / `web_railway` /
  `web_targets` as outputs. It has to be a job: a job-level `if:` is evaluated
  before the environment is attached, so `vars.` there only ever sees
  *repository* variables and an environment-scoped setting would silently read
  as empty.
- **`build-web`** now builds both web artifacts. The Vercel steps are gated on
  the resolved target; the GHCR build/push/attest steps are not. Same job rather
  than a sibling, because `migrate` gates on `build-web` and `build-backend`
  both being skipped and that gate must keep meaning what it meant.
- **`deploy-web-railway`** — the Vercel job's sibling, behind the same
  build / migrate / backend / Cloudflare gates, followed by
  `scripts/production-smoke.ts --base "$RAILWAY_WEB_ORIGIN"`. That smoke is
  `continue-on-error` while Vercel is also a target (www is not being served by
  it yet) and a hard gate the moment Railway is the only one — the expression
  flips itself.
- **`.github/actions/railway-redeploy`** — the backend's capture / redeploy /
  90×10s poll / `deploymentRollback` steps lifted verbatim into a composite
  action both services now use. Backend behaviour is unchanged.
- **`check-rollback`** short-circuits at *step* level (`active=false`, job stays
  `success`) when Vercel is not a target. `build-backend` has no `always()` in
  its gate, so skipping the job would have cascaded into skipping the backend
  release.
- **`scripts/production-smoke.ts`** stopped being Vercel-shaped: the paused
  climb shard's `cache-control` is parsed by directive rather than string-matched
  (the route emits `public, s-maxage=3600, must-revalidate`; only Vercel strips
  `s-maxage`), and every check now fails if the response's final origin is not
  `--base`.
- **`Dockerfile.web`** gains `ARG/ENV NEXT_PUBLIC_STATIC_ASSET_BASE_URL` (the
  build throws without it) and `SENTRY_RELEASE` (the Docker context has no
  `.git`), plus a BuildKit **secret mount** for `SENTRY_AUTH_TOKEN` so source
  maps upload without the token landing in the image history. The stale "nothing
  in CI builds this file" comments are corrected — `ci.yml`'s `docker-web` has
  built it since #4779.
- **`notify-web-held`**, and Railway lines in `notify-success` /
  `notify-no-promote` / `notify-failure`.

### `WEB_DEPLOY_TARGETS`

| Value            | Vercel | Railway | Notes                                        |
| ---------------- | ------ | ------- | -------------------------------------------- |
| unset or empty   | yes    | no      | The default. Today's behaviour.              |
| `vercel`         | yes    | no      | Same, written out.                           |
| `railway`        | no     | yes     | Needs the Railway service ID and smoke origin. |
| `vercel,railway` | yes    | yes     | Either order; whitespace and casing ignored. |
| `none`           | no     | no      | Web hold. Image still pushed; Discord told.  |

Anything else — an unknown name, `none` mixed with a real target, or `railway`
while `RAILWAY_WEB_SERVICE_ID` or `RAILWAY_WEB_ORIGIN` is empty — fails `resolve-web-targets`, which
skips every web deploy and fires `notify-failure`.

### Config to add (Production environment)

None of it is needed for this to merge safely; add it when the Railway web
service exists.

| Name                     | Kind | Purpose                                                        |
| ------------------------ | ---- | -------------------------------------------------------------- |
| `RAILWAY_WEB_SERVICE_ID` | var  | The Railway `web` service. Required before targeting railway.   |
| `RAILWAY_WEB_ORIGIN`     | var  | Origin for the post-deploy smoke. Required before targeting Railway. |
| `WEB_DEPLOY_TARGETS`     | var  | The switch above. Absent means `vercel`.                        |

The service's **Settings → Config file path** must point at `railway.web.toml`,
and it runs a **single replica** — `unstable_cache` / `revalidateTag` are
per-instance off Vercel (see `docs/production-deploy.md`, #4658).

Closes #4787
Part of #4648

## Current validation

Head `1ed4327ddf15865d806c24b4fd502097579a4926`, rebased onto `main` at `77a24991fa414eb8eb9a9a3875dbdd85f7765268`.

- `vp run test:service-deploy-inputs`: 101/101 passed.
- Four focused scripts files: 111/111 tests passed.
- Railway origin workflow contract: 18/18 tests passed.
- Full `vp run typecheck:web` build and typecheck passed.
- Scoped formatting, lint, and diff checks passed.

No Production environment variable, Railway service, or live traffic was changed.
## Release Notes

## Test plan

1. CI green.
2. Merge → web image publishes; Railway deploy remains skipped.
3. Open www → production still serves from Vercel.

## Risk

Risk: 4/5 — rewrites the production deploy workflow; default target keeps Vercel so the first run only adds an image push.


marcodejongh:
Sequencing constraint from the web-reposition epic #4358: this PR must **not** land before W-24 (`chore/retire-app-static-path`) and must stay out of the W-17/W-16 delete-wave window.

Two reasons, both from the program plan:
- On today's Vercel setup W-24 is a no-op cleanup, but this PR turns `Dockerfile.web`'s default export step into a live second SPA copy at `www/app` built with different env — so W-24 has to remove that path first.
- `www` is the SPA's identity provider (ten credentialed endpoints + a baked-in `EXPO_PUBLIC_WEB_URL`); a hosting migration mid-delete-wave means www downtime breaks sign-in for the app while the classic fallback no longer exists.

Epic tracking row: #4358, Tier 4 / W-24.

marcodejongh:
The W-24 sequencing constraint recorded above is cleared — #4438 closed on 2026-08-19. This PR is now Phase 2 of the kill-Vercel epic #4648.

Before merge it needs a rebase: `production-deploy.yml` has drifted since July (#4643 and #4141 both touch it). The remaining infra gate is unchanged — the Railway `web` service and the `RAILWAY_WEB_SERVICE_ID` var must exist first, and #3798 merges before this.

marcodejongh:

Conflicting: a 9-hunk conflict in .github/workflows/production-deploy.yml (1500+ line file) + docs/branch-deploys.md. Note: I checked — main's production-deploy.yml still deploys web via Vercel (no Railway deploy step yet), so this is *not* superseded by other Kill-Vercel-epic PRs; it looks like still-needed work that just needs a manual rebase against the now-much-larger deploy workflow.

marcodejongh:
In-context review (Fable) of the rewritten workflow — no blocking findings.

Checked specifically:
- `resolve-web-targets` is the only reader of `vars.WEB_DEPLOY_TARGETS`; every consumer goes through its outputs, so the environment-scoped-vars trap is closed. A malformed value skips `check-rollback` → `build-web`/`build-backend` → everything, and `notify-failure` names the job — loud, not silent.
- `check-rollback` short-circuits at step level, so `build-backend` (no `always()`) keeps running when Vercel is not a target.
- `migrate` is byte-identical; the image build sits inside `build-web`, so the both-skipped gate keeps its meaning and #4849's edits to `migrate` rebase cleanly.
- Ordering inside `build-web`: the Docker context is generated after `vercel pull`/`vercel build`. That is safe — `scripts/create-service-docker-context.mjs` ignores `.next`/`node_modules` by basename, and `.vercel/` (incl. the pulled `.env.production.local`) lives at the repo root, which the context never copies (`source/packages` + `source/scripts` only). The Sentry token is a BuildKit secret, not an ARG, and the implementer confirmed `docker history` has no trace of it.
- `continue-on-error` on the Railway smoke flips from detector to gate on its own when `web_vercel` becomes false — no edit needed at cutover.
- `notify-success` `WEB_HELD` reads off what ran (both web deploy jobs skipped while `web_changed`), same rule as `APP_WEB_HELD`.

Known trade-offs, accepted: `build-web` runs the Vercel build and the image build serially (+3–5 min per web deploy until the Vercel half is deleted in the scrub); a broken `Dockerfile.web` now also blocks the Vercel deploy (`ci.yml docker-web` gates every web PR with the same context, so it cannot arrive at main untested); once Vercel is no longer a target the backend loses the Instant-Rollback hold signal (scrub follow-up).

Merge plan: this PR merges with `WEB_DEPLOY_TARGETS` unset, so the first run only adds the GHCR push. The Railway service, `RAILWAY_WEB_SERVICE_ID` and `RAILWAY_WEB_ORIGIN` come afterwards.

marcodejongh:
CI note: `verify-trust-boundary` (PostgreSQL Image Publisher Contract) is red here, but it is red on `main` too — run 33471377738 on 93d766c92 (04:50Z) fails the same assertion (`contract workflow top-level keys must…`) while the 01:07Z run on 8747af548 passed, so a merge in that window broke it. Nothing in this branch touches that workflow or its test; reported on #4994. Not a blocker for this PR.

marcodejongh:
Current-head note: `fab1787cd4e8918c629a8e7d73fd0f3a20565d35` is rebased on current `main`; local validation is green (99 service-deploy contracts, 111 focused workflow tests, full web build/typecheck). The `verify-trust-boundary` failure is the known publisher-parser regression already fixed and green in prerequisite #5025. Merge #5025 first, then rebase this PR to rerun that gate. No Production variable, Railway service, or traffic has changed.

marcodejongh:
Marking ready for review. CI on 1ed4327dd: 27 pass; the 4 reds are all main-wide and not from this branch — `lint` is the `scripts/mobile-publish.test.ts` TS2741 (fixed by #5032), `pg18-artifacts` + `verify-trust-boundary` are the PG18 artifact/contract tests that fail on `main` (reported on #4994), and `ci-status` is the aggregate of those. Your `RAILWAY_WEB_ORIGIN`-required commit is a good tightening — with it, targeting Railway without a smoke origin fails the resolver instead of quietly skipping the smoke.

Merging with `WEB_DEPLOY_TARGETS` unset changes nothing on Vercel; the first run only adds the GHCR push of `boardsesh-web:production`.

PR #3796, opened 2026-07-20
feat(backend): dual-publish gzip board snapshots (v1 + v1-gzip)

marcodejongh:
## What & why

PostHog shows the offline snapshot download has a slow tail driven almost entirely by artifact size. The flagship, most-downloaded board — `kilter:1` — is **258 MB** uncompressed (5× the next board), and every Kilter Original user downloads the whole layout file regardless of size. Snapshot-only durations (n=45, app 2.2.2): median ~55s but **p90 ~7.2 min**, concentrated on the big Kilter boards (`kilter:1` median 238s vs `moonboard:2` 13s).

Artifacts upload uncompressed (`contentEncoding: 'identity'`). Measured on the real `kilter:1` artifact at the exact compression the export emits (`gzipSync` default = level 6): **258.1 MB → 99.5 MB (2.59× smaller)**. The mobile client already handles `contentEncoding: 'gzip'` (downloads, sniffs the gzip magic bytes, and falls back to the paged crawl if a platform doesn't transparently decode — never a crash).

## What this PR does

Publishes gzip artifacts **in parallel** with the existing identity ones, so the live fleet is never touched until gzip is validated:

- **Parameterized export key prefix** (`--key-prefix`, default `board-snapshots/v1`). Each prefix is a self-contained, single-encoding manifest; merge + prune scope entirely to the run's prefix, so a gzip run never reads or prunes the identity prefix. No manifest schema change, no `formatVersion` bump, no client change.
- **Nightly workflow runs twice**: `board-snapshots/v1` (identity, byte-for-byte the pre-gzip run — what the fleet reads) and `board-snapshots/v1-gzip` (`--gzip`). Adds `workflow_dispatch` inputs (`gzip_only`, `board`, `layout`) to populate/canary on demand. Timeout 30→45 min for the second read-only pass.
- **No mobile workflow change.** `scripts/mobile-ci-env-parity.test.ts` requires `EXPO_PUBLIC_SNAPSHOT_BASE_URL` byte-identical across all six mobile fingerprint workflows, and the `pr-<n>` preview channel bakes the same env as production — so there's no "preview-only" pointer. Cutting the fleet to gzip is an all-at-once follow-up, gated on on-device transparent-decode validation.
- Test coverage for the gzip + `--key-prefix` path; docs rewritten for the transition/cutover.

## Validation

- `vp run typecheck:backend` ✅
- `vp test run --project backend snapshot-export` ✅ (26 tests incl. new gzip/prefix cases + golden equivalence)
- `vp test run mobile-ci-env-parity` ✅ (24 — unchanged, parity intact)
- `vp check` ✅ on changed files
- Real-catalog `--dry-run --gzip` is best done post-merge via a `workflow_dispatch` against production data.

## Follow-ups (not this PR)

1. **Validate transparent decode on-device** (iOS + Android) via a dev-client pointed at `v1-gzip` (env override, uncommitted). Pass = `method: 'snapshot'`, `durationMs` down, no Sentry `snapshot-bootstrap` "still gzip-compressed".
2. **Cutover PR**: flip `EXPO_PUBLIC_SNAPSHOT_BASE_URL` to `v1-gzip` in all six mobile workflows at once (keeps parity); the production OTA migrates the fleet.
3. **Cleanup PR**: drop the identity pass and delete the `v1` prefix once the fleet has migrated.

## Release Notes

none

### DIFF
--- a/.github/workflows/export-board-snapshots.yml
+++ b/.github/workflows/export-board-snapshots.yml
@@ -1,15 +1,37 @@
 name: Export Board Snapshots
 
 # Nightly: build a per-(board, layout) SQLite snapshot of board_climbs +
-# board_climb_stats and upload it to Tigris/S3 under board-snapshots/v1/.
-# A freshly downloaded board on mobile warms from these artifacts instead of
-# paging the whole catalog over GraphQL, then resumes an incremental pull from
-# the watermarks the manifest records. Idempotent — each run mints new
-# content-addressed artifacts and rewrites the manifest last.
+# board_climb_stats and upload it to Tigris/S3. A freshly downloaded board on
+# mobile warms from these artifacts instead of paging the whole catalog over
+# GraphQL, then resumes an incremental pull from the watermarks the manifest
+# records. Idempotent — each run mints new content-addressed artifacts and
+# rewrites the manifest last.
+#
+# The run publishes the catalog under TWO prefixes side by side during the gzip
+# transition (see docs/board-snapshots.md):
+#   - board-snapshots/v1       identity-encoded — what the current fleet reads. Unchanged.
+#   - board-snapshots/v1-gzip  gzip-encoded (~2.6x smaller) — what newer builds read once
+#                              transparent Content-Encoding decode is validated on-device.
+# Each prefix is a self-contained, single-encoding manifest; the export merges
+# and prunes strictly within the prefix it targets. The follow-up that cuts the
+# whole fleet to gzip drops the identity pass and deletes the v1 prefix.
 on:
   schedule:
     - cron: '15 7 * * *' # 07:15 UTC daily (off-peak; distinct from the 06:00/06:30 refresh crons)
   workflow_dispatch:
+    inputs:
+      gzip_only:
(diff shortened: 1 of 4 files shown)


marcodejongh:
Thanks — addressed the review:

1. **Shell injection (dispatch inputs)** — fixed. `board`/`layout` now pass through `env:` and are read as shell variables (`$BOARD`/`$LAYOUT`) inside a bash array, never `${{ }}`-interpolated into the `run:` script. The CLI's `parseBoardFilter`/`parseLayoutFilter` still validate on top.
2. **Gzip step runs every night** — intended for the dual-publish window; the two full reads collapse back to one when the identity-cleanup follow-up drops the `v1` pass. Noted in the workflow comment.
3. **Missing `--layout` + non-default prefix test** — added `honors --board/--layout under --gzip --key-prefix` (filtered one-layout canary → gzip entry under `v1-gzip`, no prune).

The `large-files` CI failure was the GitHub Actions partial outage (confirmed on githubstatus.com, ~00:26–00:52 UTC); the guard passes locally in every mode and cleared on re-run.

PR #3798, opened 2026-07-20
feat(web): health endpoint + Railway config for the web service

marcodejongh:
## Summary

- Add `/api/health` — a dependency-free health endpoint for Railway's healthcheck. It never touches the database or backend, so a DB blip can't fail a deploy healthcheck or trigger a restart loop.
- Add `railway.web.toml` — config-as-code for the Railway `web` service (image `ghcr.io/boardsesh/boardsesh-web:production`), pointing its healthcheck at `/api/health` and mirroring the backend's `railway.toml` restart policy. The web service's Settings → Config file path must point at this file.
- Add `NEXT_PUBLIC_POSTHOG_KEY` as a build arg in `Dockerfile.web` so `next build` inlines it, matching how the Vercel build supplies it. Without it, client-side PostHog analytics goes dark in prod (this happened once before on the Vercel CI build path).

Plumbing only — no deploy workflow changes. #3795 wires the Railway production deploy on top of this.

Part of #4648 (Phase 2, first merge in the sequence).

## Release Notes

- [x] No release note needed (internal / technical change)

## Test plan

1. CI green.
2. After merge, `/api/health` on www returns `{"status":"ok"}`.

## Risk

Risk: 1/5 — adds an unauthenticated no-op route, a Railway config file nothing reads yet, and one build arg.

PR #3801, opened 2026-07-20
feat(ci): deploy app.boardsesh.com from the production deploy workflow

marcodejongh:
## Summary

Adds the production deploy for **app.boardsesh.com** — the standalone Expo web
export (`baseUrl /`) served at the root of its own subdomain via Cloudflare Pages
(project `boardsesh-app`, direct upload). This is the `--subdomain` artifact from
`scripts/build-expo-web-export.sh`, distinct from the `/app` export the web
Docker image bakes.

Rather than a standalone workflow, the deploy is a new **`deploy-app-web` job in
`.github/workflows/production-deploy.yml`** — the single source of truth for
production deploys — so app.boardsesh.com ships from the same pipeline as web +
backend.

### What changed

- **`.github/workflows/production-deploy.yml`**
  - `detect-changes` now also emits `app_changed` (true when
    `packages/mobile/**`, `packages/shared/**`, `packages/shared-schema/**`,
    `scripts/build-expo-web-export.sh`, or `deploy/app-subdomain/**` changed),
    computed by the exact same git-diff loop that drives `web_changed` /
    `backend_changed`. `workflow_dispatch` (and the no-usable-before-sha path)
    forces `app=true`, mirroring web/backend.
  - `deploy-app-web` job: `needs: [detect-changes]` **only** — it's a
    self-contained static export + upload with no DB or image dependency, so it's
    independent of the build/migrate gate. Guarded by
    `if: needs.detect-changes.outputs.app_changed == 'true'` (which the force-all
    dispatch already satisfies). Exports the SPA with the production env trio
    baked in, greps the emitted bundles for both prod origins (independent
    stale-env guard), copies the Pages host config, `wrangler pages deploy`s to
    `boardsesh-app`, then curl-smokes the live origin.
  - `notify-failure` now includes `deploy-app-web` in `needs` + its Discord
    message, so an app-web deploy failure alerts alongside web/backend.
- **`deploy/app-subdomain/_redirects`** — SPA fallback (`/* /index.html 200`).
- **`deploy/app-subdomain/_headers`** — `noindex` + hardening headers, immutable
  caching for content-hashed `_expo/*` and `assets/*`, documented
  no-`script-src`-CSP constraint.
- **`deploy/app-subdomain/README.md`** — why no CSP `script-src` (WASM
  `new Function` needs `unsafe-eval`), why `index.html`/`wasm/*` are not
  immutable-cached, why the auth origin is www not app.

### Env baked into the production export

| var | value | why |
| --- | --- | --- |
| `EXPO_PUBLIC_BACKEND_URL` | `https://ws.boardsesh.com` | GraphQL/backend origin |
| `EXPO_PUBLIC_WS_URL` | `wss://ws.boardsesh.com/graphql` | WebSocket endpoint |
| `EXPO_PUBLIC_WEB_URL` | `https://www.boardsesh.com` | **auth origin — deliberately www, NOT app**; the SPA does credentialed cross-origin auth against www |

### No inline smoke gate

The web/backend jobs in this workflow have no inline Playwright gate, and a
per-deploy expo-web boot (backend + Next proxy + Metro) is too heavy to run on
every production deploy. So `deploy-app-web` has none either — the
`expo-web-smoke` suite stays the dispatch-only gate in `e2e-tests.yml`. The
artifact is still verified inline: the bundle grep (stale-env) before publish and
the curl matrix against the live origin after.

## ⚠️ Conflict note — overlaps PR #3795

This edits `production-deploy.yml`, which **PR #3795** also rewrites
(Vercel → Railway). The `deploy-app-web` job and the `app_changed` output are
purely **additive and independent** of the web/backend deploy path, but they
touch the same file — whichever of #3795 / this PR merges **second must rebase**.

## Required manual ops (outside this repo)

1. Cloudflare Pages project **`boardsesh-app`** (direct upload) — **already
   exists**.
2. **Still needed:** add **`CLOUDFLARE_API_TOKEN`** and
   **`CLOUDFLARE_ACCOUNT_ID`** secrets to the GitHub **`Production`** environment
   (the `deploy-app-web` job runs `environment: Production` and reads both from
   `secrets.*`).
3. **Still needed:** attach the **app.boardsesh.com** custom domain to the Pages
   project (+ DNS).

## Dependency note (soft)

Sibling PR **#3797** (`fix/expo-web-export-metro-cache`) adds `--clear` to the
export and a `BOARDSESH_EXPORT_EXPECT_URLS` assertion. The job already sets
`BOARDSESH_EXPORT_EXPECT_URLS="https://ws.boardsesh.com https://www.boardsesh.com"`,
which is **inert until #3797 merges**, then becomes the in-export stale-env
detector. This PR does **not** depend on it — the job's own independent `grep`
over the emitted `_expo/static/js/web/*.js` bundles covers the stale-env case
regardless of merge order.

Related sibling PRs (no code overlap with this one): **#3798** (Railway
plumbing), **#3799** (app previews — uses `deploy/app-preview/`, so this PR's
`deploy/app-subdomain/` path is collision-free).

## Test plan

- Ran the export locally with the production env trio to a temp dir
  (`build-expo-web-export.sh --subdomain`): `EXPORT_EXIT=0`, the bake-verify grep
  found both `ws.boardsesh.com` and `www.boardsesh.com` in the emitted
  `_expo/static/js/web/*.js` bundles, and copying `_redirects`/`_headers`
  produced the expected publish tree (`_expo/`, `assets/`, `wasm/`, `index.html`,
  `_redirects`, `_headers`). The served `index.html` references
  `/_expo/static/js/web/entry-*.js` root-relative, matching the post-deploy
  smoke's asset-parse.
- `vp check --fix` on all changed files: pass.
- `actionlint` on `production-deploy.yml`: clean (exit 0; only pre-existing
  SC2129/SC2016 shellcheck style/info notes in the original jobs — my new job
  adds none).
- **Cannot run the deploy** until the manual ops above land (CF secrets + custom
  domain). Once they do, dispatch test plan:
  1. `workflow_dispatch` → `app_changed=true`, `deploy-app-web` exports,
     publishes, and the post-deploy curl smoke passes against app.boardsesh.com.
  2. Push a `packages/mobile/**` change to `main` → `deploy-app-web` auto-runs
     while web/backend follow their own gates.

## Release Notes


marcodejongh:
## Also fixes: every icon + board background broken on app.boardsesh.com

Deploying the subdomain export from main is only useful if the export works, so this PR now also carries the fix that makes it work.

**Root cause.** `--subdomain` set Expo's `experiments.baseUrl` to a bare `/`. Expo prepends the base to every asset path, so `/` produced `//assets/...`, which browsers resolve as the **host** `assets` (`https://assets/...` → `ERR_SSL_UNRECOGNIZED_NAME_ALERT`). Root-absolute `/_expo/*` JS and CSS were unaffected, so the app booted fine but rendered no icon fonts and no board art. Fix: normalize a root base to `''`.

**Verified against a real export + live deploy:**
- Bundle now contains **zero** `//assets/` occurrences (was the bug); emits `/assets/__node_modules`, `/assets/_web`.
- Headless load of https://app.boardsesh.com: `FONTS LOADED: 1`, icon glyphs render in the DOM, no `https://assets/` request failures. Previously: `FAILED https://assets/.../MaterialCommunityIcons.ttf`.

The `baseUrl: '/'` unit test asserted the buggy value and was updated to `''`.

Note: app.boardsesh.com is currently serving this fixed build via a manual `wrangler pages deploy` — merging this makes it automatic from main.

PR #3806, opened 2026-07-20
fix(mobile): scan unfiltered for Aurora boards so Android finds them

marcodejongh:
## Summary

The Android device picker showed an **empty Bluetooth sheet** for many Kilter/Tension users — no board to tap, so no connecting and no logging. Surfaced by a user on a Samsung Galaxy S24+ / Android 16 / 2.2.2 ("same problems as before, no connecting, can't log — haven't used it since").

**Root cause:** `RNBleAdapter` (the Android BLE path) scanned Aurora boards with a hardware service-UUID `ScanFilter`. On Android that filter never matches by name and is unreliable for a 128-bit UUID a box carries in its **scan-response** PDU, so it silently dropped boards before JS ever saw them → `devicesTotal = 0`.

Introduced by `0710be7a8` ("fix mobile bluetooth picker filtering", shipped in 2.2.x), which replaced an unfiltered scan whose own comment warned against exactly this. PostHog, Android 16, `BLE Picker Devices Resolved` empty-rate (`devicesTotal = 0`):

| Board (scan mode) | 2.1.0 | 2.2.2 |
| --- | --- | --- |
| Kilter (Aurora, filtered) | 5.8% | **22.9%** |
| Tension (Aurora, filtered) | 6.1% | **13.5%** |
| MoonBoard (unfiltered — control) | 15.9% | 25.1% |

The two *filtered* families jumped hardest while the *unfiltered* MoonBoard control rose far less on the same OS — the signature of the filter.

**Fix:** scan unfiltered and let the existing JS `isLikelyBoardDevice` filter decide — what MoonBoard already does and what Aurora did before 2.2. It reads ble-plx's merged advertise+scan-response record, so it still surfaces both **Aurora-built** (`Kilter Board#serial@N`) and **Kilter-built bare-name** (`Kilter Board`) boxes while rejecting non-boards, and is a **strict superset** of the old filtered scan — it can't lose a board, only add the ones Android dropped. Same fix for the "Bluetooth quickstart" scanner in `use-board-scan.ts`.

JS-only → the native fingerprint is unchanged, so this **OTAs to the existing 2.2.2 store binary** and reaches affected users without a store release.

## Release Notes

- Android: your board shows up in the Bluetooth list again when you reconnect — no more empty picker

## Test plan

- `vp run test:mobile` — 4290 pass. New `adapter.test.ts` case proves both an Aurora-built (serial name) and a Kilter-built bare-name box surface on the unfiltered scan **with no advertised service UUID** — the exact shape the old filter dropped. Existing Aurora test flipped to assert an unfiltered `startDeviceScan(null, …)` while still rejecting AirPods.
- `vp check` — 0 errors
- **Still needed:** real-device QA on Samsung / Android 16 + Kilter box — connect, disconnect, reconnect via the lightbulb; confirm the board reappears (`devicesTotal > 0`) instead of an empty sheet. Cover a Kilter-built bare-name box too if available (exercises the name-only path).

https://claude.ai/code/session_01BY48HrcqZ1KJ1VNcJM2kJK

### DIFF
--- a/packages/mobile/src/lib/ble/__tests__/adapter.test.ts
+++ b/packages/mobile/src/lib/ble/__tests__/adapter.test.ts
@@ -71,8 +71,8 @@ function setupConnectableAdapter(
   mockBleManager.cancelDeviceConnection.mockResolvedValue(undefined);
   mockBleManager.onDeviceDisconnected.mockReturnValue({ remove: vi.fn() });
 
-  // MoonBoards advertise no service UUID (unfiltered scan); Aurora boards carry
-  // the advertised service so they survive the UUID-filtered scan.
+  // The scan is unfiltered; isLikelyBoardDevice keeps board-like results. Aurora
+  // boards are given the advertised service UUID here; MoonBoards a name only.
   const scanName = family === 'aurora' ? 'Kilter Board#TEST@3' : 'MoonBoard A1';
   const serviceUUIDs = family === 'aurora' ? ['aurora-uuid'] : undefined;
   mockBleManager.startDeviceScan.mockImplementation(
@@ -687,7 +687,7 @@ describe('RNBleAdapter', () => {
       expect(seenDevices.map((device) => device.deviceId)).toEqual(['moon-device']);
     });
 
-    it('uses the Aurora service filter and rejects non-board peripherals on Aurora scans', async () => {
+    it('scans unfiltered and rejects non-board peripherals on Aurora scans', async () => {
       mockBleManager.onDeviceDisconnected.mockReturnValue({ remove: vi.fn() });
       mockBleManager.startDeviceScan.mockImplementation(
         (_uuids: unknown, _opts: unknown, callback: (error: unknown, device: unknown) => void) => {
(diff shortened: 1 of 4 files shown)


marcodejongh:
Converting back to draft — a deeper bisect corrected the root cause.

**The Aurora scan filter is NOT the primary cause.** `0710be7a8` (the filter) is an ancestor of the 2.1.0 release, so 2.1.0 shipped *with* it and was fine. Clean causal data (same calendar week, same Android 16, varying only app version):

| Board | 2.1.0 | 2.2.2 |
|---|---|---|
| Kilter | 5.6% | 26.0% |
| MoonBoard | 12.7% | 19.9% |

Both families regressed — including MoonBoard, which never used the filter — so the real regression is something in the **2.2.x code** (window 06-27→07-09), not the scan filter.

This PR's change (scan unfiltered + JS `isLikelyBoardDevice`) is still a valid Android robustness improvement (strict superset of the filtered scan; matches MoonBoard), but it isn't the fix for the empty-picker regression. Holding as draft while I run down the real cause (leading hypothesis: shared ble-plx scanner contention — 4 features share one scanner and call the global `stopDeviceScan()`; the Rogue timer feature landed 07-05 in-window). Will re-scope once the mechanism is confirmed.

Issue #3603, opened 2026-07-14
Backend: updateProfile mutation fails intermittently (regression ~July 8, masked DB error) — 14 users


marcodejongh:
## Summary

The `updateProfile` GraphQL mutation fails intermittently, so users can't change their display name. First seen **July 8** — a regression.

- **Sentry:** [BOARDSESH-9P](https://boardsesh.sentry.io/issues/BOARDSESH-9P)
- **Impact:** 14 users / 65 events, substatus `new`, ongoing
- **Platform:** iOS (`com.boardsesh.app@2.2.2+3`), path `updateProfile`, `source: react-query`

## What we see

Client receives HTTP 200 with a GraphQL error on path `updateProfile`:

```
Failed query: select "id", "email", "name", "image", "created_at",
  (select count(*)::int from "user_favorites" where "user_id" = "id")
  from "users" where "users"."id" = $1 limit $2
```

Two problems:
1. The **real Postgres error is masked** — drizzle only surfaces `"Failed query: <sql>"`, so we can't see the actual cause (subquery failure? statement timeout? pool exhaustion?).
2. The **full SQL is leaking to the client** in the error message (minor info-leak).

## Likely correlation

The failure window starts July 8, matching the commits that folded `favoriteCount` into the profile select as a correlated subquery:
- `613035e9a refactor(backend): extract favoriteCount subquery to a shared constant`
- `ea6eab8c7 fix(backend): fold favoriteCount into a single profile query`

The resolver (`packages/backend/src/graphql/resolvers/users/mutations.ts` `updateProfile`) also does an UPDATE followed by two separate SELECTs with **no transaction** — 4 sequential round trips that could time out under pool contention.

## Proposed first step

Wrap the resolver's DB calls so the **underlying pg error is captured to Sentry** (not just drizzle's masked wrapper), and stop leaking raw SQL to the client. Once we can see the real error, decide between: reverting/adjusting the correlated subquery, or wrapping the update+select in a single transaction.

## Acceptance

- Real Postgres error visible in Sentry for this path.
- Raw SQL no longer returned to the client.
- Root cause identified and `updateProfile` failure rate → ~0.

PR #3808, opened 2026-07-20
fix: stop non-prod contexts polluting the prod Sentry project

marcodejongh:
## Summary

Unit-test/localhost/preview activity was landing in our production Sentry project. All four Sentry clients (web browser, web server, backend, mobile) share **one DSN / one project** with almost no `environment` separation, and several non-production contexts slipped past the enablement gates:

**1. Web browser — the preview/localhost leak.** `instrumentation-client.ts` enabled Sentry on `window.location.hostname.includes('boardsesh.com')` — a *substring* match. Every branch deploy runs at `<pr>.preview.boardsesh.com` (contains `boardsesh.com`), so **every PR-preview browser session** sent client errors + logs into prod, untagged. A browser/e2e run pointed at any `*.boardsesh.com` host via `PLAYWRIGHT_TEST_BASE_URL` did the same — that's the "unit tests polluting Sentry" path (real unit tests all mock `@sentry/nextjs`, so they were already clean). → tightened to an exact-host allowlist (`boardsesh.com` / `www.boardsesh.com`).

**2. Backend — preview errors disguised as production.** The gate was `!isLocalDevelopment() && !isTestEnvironment()` — **default-on**. Railway prod leaves `NODE_ENV` unset (so the older `=== 'production'` gate had regressed prod to *silence*, issue #3603/#3183), but "any non-dev, non-test runtime" is too broad: preview/staging backends run with `NODE_ENV=production` (`branch-deploy.yml`) and so reported to prod, tagged `environment: production`, indistinguishable from real prod. → gate on a **resolved Sentry environment of `production`**. Railway prod still resolves to `production` (both `NODE_ENV` and `SENTRY_ENVIRONMENT` unset), so **prod tracking is unchanged and needs no ops change**; preview/staging declare `SENTRY_ENVIRONMENT=preview` and opt out. Mirrored in the winston `SentryWinstonTransport`, with a single source of truth in the leaf `@boardsesh/db/client/config` so the two gates can't drift.

**3. Environment separation everywhere.** Set an explicit `environment` on every SDK (web client/server/edge → `production`; mobile → `EXPO_PUBLIC_SENTRY_ENVIRONMENT ?? production`) so the default Sentry view filters cleanly to `environment:production`.

**4. Mobile.** `pr-*` OTA preview bundles now publish `EXPO_PUBLIC_SENTRY_ENVIRONMENT=preview` (`mobile-ota-preview.yml`, preview-only — deliberately **not** a shared fingerprint key, JS-inlined so it doesn't move the fingerprint) so a tester on a pr channel reports `environment: preview` and their crashes stay out of the prod view. Store/production builds leave it unset → `production`. `environment` is init-only in this SDK (no runtime setter), so the build-time var is the equivalent of deriving it from `Updates.channel`.

The web *server*/edge gate (`VERCEL_ENV === 'production'`) was already correct and is unchanged apart from the `environment` tag.

### Known residual (documented, not a systemic leak)
Prod (Railway) and a bare local `bun run backend:start` (`tsx src/index.ts`, no `NODE_ENV`) are indistinguishable by env, and we deliberately avoid `RAILWAY_*` branching. `vp run dev` sets `NODE_ENV=development` (excluded); only explicitly running the *prod* backend command locally would still report — set `SENTRY_ENVIRONMENT=development` there.

### Not touched (possible follow-up)
`packages/web/app/lib/analytics.ts` gates **PostHog** with the same `hostname.includes('boardsesh.com')` substring — preview deploys likely pollute prod PostHog too. Left out to keep this PR Sentry-scoped; flagging for a follow-up.

## Release Notes

## Test plan

- `vp run typecheck:backend` / `:mobile` / `:web` — pass
- `vp test run --project mobile src/lib/__tests__/sentry.test.ts` — 30 pass (adds `resolveSentryEnvironment` coverage)
- `vp test run --project backend src/utils/__tests__/sentry-transport.test.ts` — 12 pass (adds the `SENTRY_ENVIRONMENT=preview` opt-out case)
- `packages/db` `sentry-environment.test.ts` (node test runner) — 6 pass
- `vp test run scripts/mobile-ci-env-parity.test.ts` — 24 pass (preview-only `EXPO_PUBLIC_SENTRY_ENVIRONMENT` doesn't break fingerprint parity)
- `vp check` — 0 errors

### DIFF
--- a/.github/workflows/branch-deploy.yml
+++ b/.github/workflows/branch-deploy.yml
@@ -165,6 +165,7 @@ jobs:
               -e BOARDSESH_URL=https://${PR_NUM}.preview.boardsesh.com \
               -e NEXTAUTH_SECRET=${NEXTAUTH_SECRET} \
               -e NODE_ENV=production \
+              -e SENTRY_ENVIRONMENT=preview \
               --label "traefik.enable=true" \
               --label "traefik.http.routers.backend-pr-${PR_NUM}.rule=Host(\`${PR_NUM}.ws.preview.boardsesh.com\`)" \
               --label "traefik.http.routers.backend-pr-${PR_NUM}.priority=100" \
@@ -196,6 +197,7 @@ jobs:
             -e BACKEND_INTERNAL_URL=${BACKEND_INTERNAL_URL} \
             -e VERCEL_ENV=preview \
             -e NODE_ENV=production \
+            -e SENTRY_ENVIRONMENT=preview \
             --label "traefik.enable=true" \
             --label "traefik.http.routers.web-pr-${PR_NUM}.rule=Host(\`${PR_NUM}.preview.boardsesh.com\`)" \
             --label "traefik.http.services.web-pr-${PR_NUM}.loadbalancer.server.port=3000" \
--- a/.github/workflows/mobile-ota-preview.yml
+++ b/.github/workflows/mobile-ota-preview.yml
@@ -93,6 +93,12 @@ env:
   # doesn't enter the resolved config, so it never moves the fingerprint. Kept in
   # lockstep with the other mobile fingerprint workflows. See docs/mobile-ota-updates.md.
   EXPO_PUBLIC_USE_RN_FETCH: '1'
+  # Preview-only tag: pr-<number> OTA bundles report `environment: preview` to Sentry
+  # so testers' crashes stay out of the production view (packages/mobile/src/lib/sentry.ts).
+  # DELIBERATELY diverges from production — do NOT add to SHARED_ENV_KEYS in
+  # scripts/mobile-ci-env-parity.test.ts. Bundle-only (EXPO_PUBLIC_*, JS-inlined), so it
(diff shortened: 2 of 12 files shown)


marcodejongh:
Thanks for the review. On the three notes:

1. **PostHog substring bug** — filed as #3814 (includes the exact-host fix + a note to confirm whether preview analytics are wanted before hard-disabling).
2. **`sentry-transport.ts` seam asymmetry** — intentional, and the code comments it: the `SENTRY_ENVIRONMENT` opt-out is checked *before* the `nodeEnvOverride` seam so the seam can never re-enable a preview/staging deploy. The `nodeEnv='development'` + `SENTRY_ENVIRONMENT='production'` combo returning `false` is correct (dev wins) and doesn't occur in practice, so I'm leaving it as-is.
3. **`VITEST` coverage** — agreed, no action; the `VITEST=true` branch is covered by `isTestEnvironment`'s own tests.

Marking ready.

PR #3811, opened 2026-07-20
fix(mobile): high-power (LowLatency) foreground board scans

marcodejongh:
## Summary

Follow-up to #3806. A bisect of the Android empty-picker regression (`BLE Picker Devices Resolved` with `devicesTotal = 0`) traced the real cause to the **Expo SDK 57 / React Native 0.86 native upgrade** (`02b6a6610`), which recompiled the Android BLE stack and degraded scan discovery. Clean causal data (same calendar week, same Android 16, varying only app version):

| Board | 2.1.0 | 2.2.2 |
| --- | --- | --- |
| Kilter | 5.6% | **26.0%** |
| MoonBoard | 12.7% | **19.9%** |

Both families regressed (MoonBoard never used the scan filter #3806 removed), the JS scan path is byte-identical across the regression, and `react-native-ble-plx` is unchanged at **3.5.1** — already the latest published version, with no upstream fix. The library's own tracker documents "Android scan misses devices" as a long-standing issue whose community mitigation is a **high-power scan mode**.

The foreground "find my board" scans passed `null` scan options, so ble-plx used its default **`LowPower`** (~10% duty cycle). This switches the device picker (`adapter.ts`) and the Bluetooth quickstart (`use-board-scan.ts`) to a shared `HIGH_POWER_BOARD_SCAN_OPTIONS`:

- **`scanMode: LowLatency`** (Android only) — ~100% scan duty cycle.
- **`allowDuplicates: true`** (iOS only) — reports each advertisement repeatedly, so a name/UUID arriving in a later ADV/scan-response packet still reaches the JS filter.

For a user actively trying to light their board, high power is the right trade. `scan-options.ts` uses a type-only ble-plx import so it never pulls the native package into the web bundle. JS-only → the native fingerprint is unchanged, so this **OTAs onto the existing 2.2.2 fleet**.

> ⚠️ **This is a mitigation, not a root-cause fix.** The root cause is native (RN 0.86); there's no ble-plx version to bump to. If LowLatency doesn't fully recover discovery on-device, the next step is a hand-written ble-plx native patch. **It must be validated on a real Android device** — BLE discovery can't be exercised in CI or an emulator.

## Release Notes

- Android: faster, more reliable Bluetooth — your board turns up quicker when you go to connect

## Test plan

- `vp run test:mobile` — BLE adapter + quickstart tests updated to assert the high-power options; 115 BLE tests pass
- `vp check` — 0 errors
- `vp run check:mobile-bundle` — OK (web bundle unaffected: scan-options.ts is a type-only ble-plx import)
- **Required before merge:** on-device QA on a real Android phone (ideally a Samsung / Android 16) with a Kilter/Tension board — open the connect picker, confirm the board shows up reliably (`devicesTotal > 0`) where it was coming up empty. Validate via the `pr-<number>` OTA channel (More → Development → OTA Channel Switcher). Watch the Aurora×Android empty-rate on the "Mobile BLE Health" PostHog dashboard afterward.

https://claude.ai/code/session_01BY48HrcqZ1KJ1VNcJM2kJK

### DIFF
--- a/packages/mobile/src/lib/ble/__tests__/adapter.test.ts
+++ b/packages/mobile/src/lib/ble/__tests__/adapter.test.ts
@@ -49,6 +49,7 @@ vi.mock('@boardsesh/ble-protocol', () => ({
 // ── Import after mocks ─────────────────────────────────────────────────
 
 import { RNBleAdapter } from '../adapter';
+import { HIGH_POWER_BOARD_SCAN_OPTIONS } from '../scan-options';
 import { SCAN_TIMEOUT_MS, SERIAL_RECONNECT_GRACE_MS } from '@boardsesh/ble-protocol/scan-constants';
 import { parseSerialNumber, splitMessages } from '@boardsesh/ble-protocol';
 import { State } from 'react-native-ble-plx';
@@ -683,7 +684,11 @@ describe('RNBleAdapter', () => {
       await adapter.requestAndConnect();
 
       // Unfiltered scan: UUID filter must be null (a service filter hides MoonBoards).
-      expect(mockBleManager.startDeviceScan).toHaveBeenCalledWith(null, null, expect.any(Function));
+      expect(mockBleManager.startDeviceScan).toHaveBeenCalledWith(
+        null,
+        HIGH_POWER_BOARD_SCAN_OPTIONS,
+        expect.any(Function),
+      );
       expect(seenDevices.map((device) => device.deviceId)).toEqual(['moon-device']);
     });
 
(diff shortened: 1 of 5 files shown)


marcodejongh:
Thanks — addressed:

**Magic number `2`:** clarified in `scan-options.ts` that this isn't a fragile ble-plx-internal number but Android's **ABI-stable platform constant** `ScanSettings.SCAN_MODE_LOW_LATENCY` (== ble-plx `ScanMode.LowLatency`), which ble-plx forwards straight to `ScanSettings.Builder().setScanMode(int)`. It can't be renumbered without breaking the Android SDK, so the literal is safe across ble-plx upgrades — and it keeps this module a type-only ble-plx import (no native package in the web bundle).

**allowDuplicates dedup coverage:** already covered — the quickstart '\''scans and deduplicates serials from discovered devices'\'' test feeds the same device twice and asserts a single serial, and the adapter has '\''deduplicates repeated board names even when the native device id changes'\'' for `upsertDiscoveredDevice`. So a dedup regression would fail an existing test.

On-device Android QA remains the merge gate (holding as draft).

PR #3821, opened 2026-07-20
fix(mobile): scan Aurora unfiltered on the native iOS BLE path

marcodejongh:
## Summary

The iOS **native** BLE adapter still hardware-filtered Aurora scans by service UUID (`native.startScan(['4488b571-…'])`), while the Android adapter (`adapter.ts`) and the Bluetooth quickstart (`use-board-scan.ts`) were reverted to unfiltered scanning in #3806/#3811. The native iOS path was missed.

iOS `scanForPeripherals(withServices:)` only matches the **primary** advertisement. Any box whose 128-bit Aurora UUID rides the **scan-response** PDU — Kilter-built bare-name boxes, and NimBLE-based dev boards (`Kilter Board#123456@3`, whose 16-byte UUID + 21-char name can't co-fit in the 31-byte legacy ADV, per the firmware's own comment in `embedded/libs/nordic-uart-ble/src/nordic_uart_ble.cpp`) — never reached the picker, giving an **empty device list**. TestFlight/production builds always take this native path (`adapter-factory.ts`), so a native OTA "broke" the picker.

This reverts the native iOS Aurora scan to its pre-`0710be7a8` behavior: scan unfiltered on capable binaries and narrow in JS via `isLikelyBoardDevice`, exactly like the Android path. Older binaries keep the explicit per-family UUID fallback. Pure JS — ships over OTA (the native Swift already supports the empty-array unfiltered path).

`0710be7a8` is the same commit `adapter.ts` already blames for the ~23% Android Kilter empty-picker regression; this closes the loop for iOS.

### Why this can't break real climbers (no hardware to test on)

- **Revert to shipped-and-worked behavior**, not a new approach.
- **Android already scans Aurora unfiltered in production** with the same JS filter — real Kilter/Tension boards connect fine.
- **Strict superset:** every real board the hardware filter finds today advertises the Aurora UUID, so the native `scanResult` carries it in `serviceUuids` → the JS filter passes it via the UUID check. Unfiltered can only *add* devices (scan-response-UUID boxes), never remove a currently-working one.

## Release Notes

- Fixes Kilter/Tension boards not showing up in the Bluetooth picker on iPhone for some LED boxes — they'll appear again when you go to connect.

## Test plan

- `vp run test:mobile` (native-ios-adapter, adapter, use-board-scan suites): 77 passed. Added a case proving a board whose Aurora UUID never reached `serviceUuids` (rode the scan-response PDU) now surfaces by name, while an unrelated peripheral is still dropped by the JS filter.
- `vp run typecheck:mobile`: clean.
- `vp check`: 0 errors.
- No real board on hand — see the superset safety argument above. Field-verify on an iPhone TestFlight build once the OTA lands.

https://claude.ai/code/session_013ppKysAzJxvhHt312kzCg3

### DIFF
--- a/packages/mobile/src/lib/ble/__tests__/native-ios-adapter.test.ts
+++ b/packages/mobile/src/lib/ble/__tests__/native-ios-adapter.test.ts
@@ -544,6 +544,25 @@ describe('NativeIosBleAdapter on older binaries (no adoption surface)', () => {
     // Nordic UART service — without an unfiltered scan we must list both.
     expect(nativeMock.startScan).toHaveBeenCalledWith(['UART-UUID', 'REDBEARLAB-UUID']);
   });
+
+  it('filters an Aurora scan on the Aurora service when unfiltered scanning is unavailable', async () => {
+    // Explicit older-binary Aurora coverage: without the adoption surface we
+    // can't scan unfiltered, so we must keep the native Aurora service filter
+    // (an old binary always filters natively regardless). The scan-response-PDU
+    // gap this diff fixes only exists on capable binaries that go unfiltered.
+    const adapter = new NativeIosBleAdapter(
+      (subscribe) =>
+        new Promise<string>(() => {
+          subscribe(() => {});
+        }),
+      'aurora',
+    );
+    void adapter.requestAndConnect();
+    await Promise.resolve();
+    await Promise.resolve();
+
+    expect(nativeMock.startScan).toHaveBeenCalledWith(['AURORA-UUID']);
+  });
 });
 
 // ── Per-write transport diagnostics (#3230) ─────────────────────────────────
--- a/packages/mobile/src/lib/ble/__tests__/native-ios-adapter.test.ts
+++ b/packages/mobile/src/lib/ble/__tests__/native-ios-adapter.test.ts
@@ -465,7 +465,7 @@ describe('NativeIosBleAdapter on newer binaries (adoption surface present)', () 
     expect(nativeMock.connect).toHaveBeenCalledWith('moon-1');
   });
 
-  it('uses the Aurora service filter and drops unrelated devices on Aurora scans', async () => {
+  it('scans unfiltered on Aurora and surfaces boards whose UUID rode the scan-response PDU', async () => {
     let manualPick: (deviceId: string) => void = () => {};
     const seenDeviceIds: string[] = [];
     const adapter = new NativeIosBleAdapter(
@@ -482,27 +482,42 @@ describe('NativeIosBleAdapter on newer binaries (adoption surface present)', () 
     await Promise.resolve();
     await Promise.resolve();
 
-    expect(nativeMock.startScan).toHaveBeenCalledWith(['AURORA-UUID']);
+    // Unfiltered on the newer surface: a native `withServices:[Aurora]` filter
+    // only matches the primary ADV, dropping boxes (Kilter-built bare-name /
+    // NimBLE dev boards) whose UUID rides the scan-response PDU. The JS filter
+    // below narrows instead. Matches the Android adapter (adapter.ts, #3806).
+    expect(nativeMock.startScan).toHaveBeenCalledWith([]);
 
+    // An unrelated peripheral must still be dropped by the JS filter...
     scanListeners[0]?.({
       device: { deviceId: 'airpods', name: "Marco's AirPods #1" },
       localName: "Marco's AirPods #1",
       rssi: -30,
       serviceUuids: [],
     });
+    // ...a board that DID advertise the UUID in its primary ADV must surface...
     scanListeners[0]?.({
       device: { deviceId: 'aurora-1', name: 'Kilter Board#751737@3' },
       localName: 'Kilter Board#751737@3',
       rssi: -40,
       serviceUuids: ['AURORA-UUID'],
     });
+    // ...and so must one whose Aurora UUID never reached `serviceUuids` (rode
+    // the scan-response PDU) — matched by its name alone. The hardware filter
+    // would have dropped this one; the regression that broke the dev board.
+    scanListeners[0]?.({
+      device: { deviceId: 'aurora-devboard', name: 'Kilter Board#123456@3' },
+      localName: 'Kilter Board#123456@3',
+      rssi: -45,
+      serviceUuids: [],
+    });
 
-    expect(seenDeviceIds).toEqual(['aurora-1']);
+    expect(seenDeviceIds).toEqual(['aurora-1', 'aurora-devboard']);
 
-    manualPick('aurora-1');
+    manualPick('aurora-devboard');
     await vi.runAllTimersAsync();
     await connectPromise;
-    expect(nativeMock.connect).toHaveBeenCalledWith('aurora-1');
+    expect(nativeMock.connect).toHaveBeenCalledWith('aurora-devboard');
   });
 });
 
--- a/packages/mobile/src/lib/ble/native-ios-adapter.ts
+++ b/packages/mobile/src/lib/ble/native-ios-adapter.ts
@@ -133,9 +133,15 @@ export class NativeIosBleAdapter implements BluetoothAdapter {
       openPicker();
     }
 
-    // MoonBoard scans are unfiltered on newer binaries (a native service-UUID
-    // filter would hide controllers that only surface via name). Aurora scans
-    // stay filtered to avoid unrelated peripherals reaching the picker.
+    // Both families scan unfiltered on newer binaries, then narrow in JS via
+    // isLikelyBoardDevice. A native service-UUID filter hides any box whose UUID
+    // rides the scan-response PDU rather than the primary advertisement — iOS
+    // `scanForPeripherals(withServices:)` only matches the primary. That drops
+    // Kilter-built bare-name boxes and NimBLE-based controllers (dev boards)
+    // whose 128-bit Aurora UUID + long name can't co-fit in the 31-byte legacy
+    // ADV, giving an empty picker (regression 0710be7a8, the same one adapter.ts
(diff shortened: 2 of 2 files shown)


marcodejongh:
Closing this: the premise doesn't hold on iOS, and the field data says there's no iOS discovery problem to fix. Writing up the evidence so nobody re-files it.

## 1. iOS foreground scans do match a scan-response service UUID

The PR rests on "iOS `scanForPeripherals(withServices:)` only matches the primary advertisement." That's true for **background/passive** scanning only. In the foreground iOS scans **actively** (sends `SCAN_REQ`) and merges `ADV_IND` + `SCAN_RSP` **before** applying the service filter.

The clearest proof is Nordic's own reference stack: `peripheral_lbs` / `peripheral_uart` put the 128-bit service UUID exclusively in the scan response (`sd[]`), and Nordic's iOS nRF Blinky app scans `withServices: [nordicBlinkyServiceUUID]` and finds them. If the claim were true, Nordic's own paired sample apps could not work.

`NativeIosBleAdapter.requestAndConnect` is foreground-only. The one background-capable scan we have is `beginReconnectScanOnBleQueue` (`BoardBleManager.swift:734`), which passes three explicit UUIDs and this PR doesn't touch. So the case where the claim *is* true is the case this PR never runs in.

This is genuinely true on **Android**, whose offloaded hardware ScanFilter is unreliable for 128-bit UUIDs — which is why #3806 was right for Android. The symmetry argument is a false symmetry: the two platforms have different scan-filter semantics.

## 2. Our own firmware inverts the PR's premise

`embedded/libs/nordic-uart-ble/src/nordic_uart_ble.cpp:38-41`:

```cpp
pAdvertising->addServiceUUID(AURORA_ADVERTISED_SERVICE_UUID);
// A legacy BLE advertisement only has 31 bytes. Advertising both 128-bit UUIDs
// overflows the packet, so expose Aurora for discovery and keep NUS in GATT.
pAdvertising->enableScanResponse(true);
pAdvertising->setName(deviceName);
```

The DIY board puts the Aurora UUID in the **primary ADV** and spills the **name** to the scan response — the comment says so outright ("expose Aurora for discovery"). That's the opposite of what the PR describes. And per §1, even if NimBLE spilled the UUID instead, a filtered foreground iOS scan would still match it.

## 3. Telemetry: iOS discovery is healthy

`BLE Picker Devices Resolved`, non-emulator, 120d:

| Surface | Empty-picker rate |
| --- | --- |
| iOS Kilter 2.2.2 | **3.9%** (122/3160) |
| iOS Tension 2.2.2 | **2.0%** (6/307) |
| iOS MoonBoard 2.2.2 (already unfiltered) | **14.0%** (157/1125) |
| Android Kilter/Tension, 2 wks before #3806 | **17.0%** (299/1757) |
| Android Kilter/Tension, after #3806+#3811 | **13.0%** (60/463) |

iOS *with* the service filter is 3-4x better than Android *without* it, and 3-4x better than the one iOS family that already scans unfiltered. Weekly iOS Aurora rate over the last three weeks: 3.2%, 2.7%, 4.7% — flat.

**Chronic-cohort test** (users with 3+ Aurora pickers): iOS 2.2.2 — **1 of 337 users (0.3%)** never saw a device; iOS 2.1.0 — 1 of 230 (0.4%); Android 2.2.2 — 1 of 160 (0.6%). There is no population of iPhone owners whose box the filter is hiding.

Second, independent oracle — `Bluetooth Connection Failed` with `failureReason = 'board_not_found'`, as a share of all connection attempts: iOS flat at 0.5-0.6% every week since mid-June, *below* Android's 0.8-1.4%.

### Caveat, stated honestly

A clean before/after on the filter itself is **impossible**. `BLE Picker Devices Resolved` first fired 2026-06-10 — the same day `0710be7a8` introduced the filter. So every picker datapoint we have is from the filtered era. The correct reading of the above is **"no broken cohort exists, and iOS beats unfiltered Android"**, not "unfiltering made things worse." I'm not claiming the stronger result.

## 4. What the change would have cost

- **Battery.** Apple on `CBCentralManagerScanOptionAllowDuplicatesKey`: *"Disabling this filtering can have an adverse effect on battery life; use it only if necessary."* We already pass `allowDuplicates: true` unconditionally (`BoardBleManager.swift:610-612`), so this PR doesn't change the policy — it changes the volume, to one `didDiscover` per advertising packet from every nearby advertiser for the full 30s `SCAN_TIMEOUT_MS` window. `emittedScanResults` bounds native->JS crossings but not the CoreBluetooth callback rate on `bleQueue`.
- **A background hole.** We declare `UIBackgroundModes: ['bluetooth-central']` (`app.config.ts:370`), and a backgrounded central must name services — `withServices: nil` delivers nothing. Narrow (only a scan still running when the user backgrounds the app), but it's a regression where today there is none.

Net: a documented battery cost and a small background regression, in exchange for no measurable benefit.

## 5. What a future attempt should look like

If someone wants belt-and-braces coverage of the genuine tail case — a box advertising **no** Aurora UUID in either PDU but carrying a board-like name — don't unfilter unconditionally. **Escalate**: start filtered, and if nothing has surfaced after ~5s, `stopScan()` then `startScan([])` for the remainder. `startScanOnBleQueue` already clears `emittedScanResults` on every start, so a restart re-emits everything. No native change, still OTA, and it pays the battery cost only when the filtered scan has actually come up dry. It mirrors the existing `SERIAL_RECONNECT_GRACE_MS` grace-window pattern.

But that tail is **unquantified**, and the lever for quantifying it is not a speculative scan change — it's more `BLE Advertisement Recon` data. We have exactly 4 recon events ever (it only fires when a user files a bug report with the Bluetooth toggle on), and the single bare-name `Kilter Board` sample on iOS *did* carry the Aurora UUID. Get more recon submissions first.

## Follow-up filed

The comment at `packages/mobile/src/lib/ble/adapter.ts:128-138` still blames `0710be7a8` for the ~23% Android empty-picker regression. The later causal analysis cleared the filter (2.1.0 shipped the same filter and was healthy) and attributed it to the Expo 57 / RN 0.86 native scan-reliability drop; #3806 was explicitly robustness-only. Quoting that comment as justification is how this PR got written, so it's getting fixed separately.

marcodejongh:
## Addendum: independent adversarial review of this closure

I ran a second, adversarial pass whose explicit job was to build the strongest possible case that this PR **should** ship, and to re-run the telemetry independently. Verdict: **the closure holds.** Recording what it found, including where it corrected me.

### One fact that strengthens the closure, missed above

**The unfiltered native iOS Aurora scan has never run in production.** iOS 2.2.0 — the first unfiltered-capable binary — only reached users the week of 2026-06-29, *three weeks after* the filter landed (`0710be7a8`, 2026-06-10). So the PR's framing as restoring prior behaviour is wrong for the iOS native path specifically: there is no golden prior state to revert to.

### A better oracle, which also confirms

The original analysis used "empty picker" (`devicesTotal = 0`), which is blind to the user who sees *some* boards but not theirs. `noneMatchedSelectedType` is the sharper signal, and on iOS 2.2.2 Kilter it sits at **1.1% (34/3160)**. Tension reads 15%, but 9 of the 11 flagged users connected to a Tension board immediately afterwards, and MoonBoard's 86% is a type-resolution artifact.

### Correction to my numbers above

I described iOS Aurora `board_not_found` as "flat at 0.5–0.6% every week since mid-June." That is wrong. The actual weekly series is **0.16, 0.16, 0.27, 0.22, 0.57%** — rising, with the latest week the highest (13 events). Drilling in, those are single-day bursts by individual users with no repeat-across-days cohort, so the conclusion is unchanged, but the characterisation was sloppy.

Also: an apparent iPadOS 25% empty rate (15/60) collapses to **one user**, who accounts for all 15 empties across 30 pickers. Intermittent (board off / out of range), not the deterministic all-empty signature a filter-hidden box would produce.

### Where the review disagreed with me

**My battery argument was overstated.** Radio duty cycle is identical filtered vs unfiltered — iOS applies the filter after receipt, and `allowDuplicates: true` is already set unconditionally (`BoardBleManager.swift:610-612`). The real marginal cost is per-callback CPU on `bleQueue` for a 30s user-initiated window (~50 gym advertisers ≈ low-hundreds of callbacks/sec doing dictionary upserts), bounded at the bridge by `emittedScanResults`. Negligible. The closure does not rest on that pillar; it rests on the scan-response mechanism and the field data.

### The strongest honest case FOR the PR

Defence-in-depth. Android runs this exact design in production, the JS filter (`isLikelyBoardDevice`) is a strict superset of the hardware filter, and "iOS merges SCAN_RSP before filtering" is OS behaviour Apple documents nowhere and could change in any release. Unfiltering buys permanent immunity to that class of failure for close to nothing.

That is an argument for robustness against a hypothetical, not a fix for an observed bug — and it is **not** the argument this PR makes. The intra-iOS datum is the one that settles it: **iOS 2.1.0, permanently hardware-filtered with no unfiltered capability, has the lowest empty-picker rate in the entire fleet at 2.7% (53/1994)** — below the 2.2.2 filtered-native rate. If the filter hid a class of boxes, the always-filtered cohort is exactly where that would show.

If we ever want the robustness anyway, it should be re-proposed on the honest premise, not on "iOS only matches the primary advertisement," which is factually wrong and would otherwise be enshrined in a load-bearing comment.

### What would reopen this

1. A `BLE Advertisement Recon` sample of a real Aurora-family box with **no** `4488b571-…` UUID in the merged advertisement view but a board-like name.
2. An on-device test: current-iOS iPhone + a peripheral advertising the Aurora UUID *only* in SCAN_RSP, where `scanForPeripherals(withServices: [aurora])` fails to surface it in the foreground.
3. A user with all-empty pickers across ≥3 sessions whose box connects fine from an Android phone in the same room.
4. iOS 2.2.x chronic all-empty growing past 1/387, or Kilter `noneMatchedSelectedType` climbing above ~1% with flagged users who never subsequently connect.

PR #3830, opened 2026-07-21
feat(web): point climb og:image at the backend /og/climb endpoint

marcodejongh:
## Summary

Second of two PRs moving climb OG share cards off Vercel. #3829 added the backend `GET /og/climb` endpoint (warm WASM renderer, in-memory caches, ~250ms steady-state renders, 78KB JPEG vs 324KB PNG); this PR flips web's climb `og:image` metadata to point at it.

⚠️ **Do not merge until #3829 has deployed and been curl-verified in production** — the og:image URLs this emits must resolve. Verification commands are in `docs/og-climb.md` (added in #3829).

- `buildOgBoardRenderUrl` emits the absolute `https://ws.boardsesh.com/og/climb?…&format=jpeg` URL (origin derived from `NEXT_PUBLIC_WS_URL`; local dev resolves to `http://localhost:8080`). When the origin can't be resolved it falls back to the current relative web URL, byte-identical to today's.
- `normalizePath` in `createPageMetadata` passes absolute `https?://` image URLs through untouched (previously it would mangle them into `/https://…`); covers both openGraph and twitter images.
- The legacy `/api/og/climb` redirect keeps working for previously-shared links, now 307ing to the backend URL.
- In-app board rendering (`/api/internal/board-render`) is untouched.

## Release Notes

Shared climb links now show their preview card instantly on Facebook, WhatsApp, Slack and friends

## Test plan

- `vp run typecheck:web` clean (full `next build` — validates no client/server boundary issues)
- `vp run test:web` — 5721 tests green; new assertions pin the absolute URL shape (host, `/og/climb`, `format=jpeg`), the relative fallback when the env is unset, absolute-URL passthrough in metadata for both OG and Twitter images, and the legacy redirect's absolute target
- QA-reviewed for param compatibility against the backend zod schema (set_ids regex, frames validity, format enum) — all compatible
- Post-merge: curl a live climb page as `facebookexternalhit` and confirm `og:image` points at `ws.boardsesh.com/og/climb…format=jpeg`; scrape in the Facebook Sharing Debugger and confirm the image renders with no timeout warning

### DIFF
--- a/packages/web/app/api/og/climb/__tests__/route.test.tsx
+++ b/packages/web/app/api/og/climb/__tests__/route.test.tsx
@@ -39,7 +39,7 @@ vi.mock('@/app/lib/board-utils', () => ({
 }));
 
 vi.mock('@/app/components/board-renderer/util', () => ({
-  buildOgBoardRenderUrl: vi.fn(() => '/api/internal/board-render?board_name=kilter&variant=og&format=png'),
+  buildOgBoardRenderUrl: vi.fn(() => 'https://ws.boardsesh.com/og/climb?board_name=kilter&variant=og&format=jpeg'),
 }));
 
 function makeRequest(params: Record<string, string>): NextRequest {
@@ -64,12 +64,12 @@ describe('api/og/climb legacy redirect', () => {
     vi.clearAllMocks();
   });
 
-  it('redirects legacy climb OG requests to the immutable Rust-render URL', async () => {
+  it('redirects legacy climb OG requests to the absolute backend OG image URL', async () => {
     const response = await GET(makeRequest(validParams));
 
     expect(response.status).toBe(307);
     expect(response.headers.get('location')).toBe(
-      'http://localhost:3000/api/internal/board-render?board_name=kilter&variant=og&format=png',
+      'https://ws.boardsesh.com/og/climb?board_name=kilter&variant=og&format=jpeg',
     );
     expect(response.headers.get('Cache-Control')).toContain('s-maxage=300');
   });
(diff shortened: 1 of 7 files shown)


marcodejongh:
Review feedback addressed in 2a8ea59ea:

- **URL construction**: now built with `URLSearchParams` — every param consistently encoded.
- **Inert params**: `include_background`/`variant` dropped from the backend URL entirely (also removes any future strict-schema risk).
- **Stale contentType hint**: the legacy redirect now passes `image/jpeg` with a comment clarifying only the cache headers are applied to the redirect.

marcodejongh:
Latest review addressed in a2f3f71fb:

- **Missing angle param**: not a bug — angle has never been a render input. The pre-change builder on `main` sent the same param set without angle: the lit holds come entirely from `frames` and the background from board/layout/size/sets. Different angles of the same climb genuinely share one card (same as today), and #3829's schema doesn't accept an angle param.
- **Mock URL drift**: the redirect test mock now matches the real builder output (full param set, `%2C`-encoded set_ids, no `variant`).
- **Redundant assertions after `toBe`**: removed.
- **Comment wording**: now states `image/jpeg` mirrors what the backend target serves.

marcodejongh:
#3829 is deployed and verified live — marking this ready.

Production verification of `https://ws.boardsesh.com/og/climb` (per the plan):
- Real climb render: **200, image/jpeg, 1200×630 progressive, 86KB**, `Cache-Control: … immutable`
- Server-side timings: first fetch per replica ~110–130ms (`base-hit` — boot warm-up pre-composed the board bases, so no cold `miss` even on the first-ever fetch), repeats **~0ms** (`hit`)
- Two replicas each hold their own in-memory cache: an occasional ~110ms `base-hit` until both have seen a URL, then all fetches are ~0ms. Worst case is still ~25x faster than the old Vercel origin render (~1.0s + cold start)
- Garbage rejected cheaply: unknown board → 400, empty frames → 400

After this merges: curl a climb page as `facebookexternalhit` to confirm the og:image URL flip, then scrape in the FB Sharing Debugger.

PR #3831, opened 2026-07-21
fix(gyms): stop location syncs from overwriting user-curated gyms & boards

marcodejongh:
## Summary

Location syncs (Aurora, Kilter, MoonBoard) all funnel through `upsertPublicBoardLocations` in `packages/location-sync/src/upsert.ts`, which overwrote gym/board metadata on **every** run. That silently reverted human changes in two ways:

1. **Claimed gyms** — `applyGymClaim` transfers ownership to a real user but keeps the source alias, and `findAliasedGymId` filters only on `deletedAt IS NULL` (no owner check). The next sync's `refreshSyncedGymMetadata` overwrote name/coords/visibility and cleared `deletedAt`, resurrecting a gym the owner had deleted.
2. **Moderator-fixed catalog rows** — a community admin can edit a still-system-owned gym (`updateGym`) or "fix outdated catalog boards" (`updateBoard`); the row stays system-owned, so the next sync clobbered it.

### Approach

A per-row `sync_frozen_at` marker on `gyms` and `user_boards`. Non-null means a human curated the row, so the sync engine must never overwrite its metadata again.

- **Human write paths stamp it**: `updateGym`, `deleteGym`, `applyGymClaim` (taking ownership freezes it, per product decision), `updateBoard`, `deleteBoard`. The logo-upload flow already persists via `updateGym`, so it's covered.
- **Sync engine skips frozen rows**: `refreshSyncedGymMetadata` gains an `isNull(syncFrozenAt)` WHERE; both `ON CONFLICT DO UPDATE` blocks use `setWhere: isNull(syncFrozenAt)` (mirrors the existing user-climb guard in `catalog-sync.ts`). A by-uuid fallback still resolves the gym id when the frozen row blocks the upsert, so the source alias and board→gym link stay current.
- **Dedupe tool** (`dedupe-gym-locations.ts`) excludes frozen gyms from candidates entirely — a human-touched gym is never merged, moved, un-deleted, or re-published.
- **Backfill migration** freezes every already user-owned row (`owner_id <> SYSTEM`) so existing claims are protected retroactively.

Granularity is per-row (any edit freezes the whole row); a claimed-but-unedited gym is also frozen. Both were explicit product decisions.

Migrations: `0184_add_sync_frozen_at` (DDL) + `0185_backfill_sync_frozen_at` (custom backfill).

## Release Notes

- Your edits to a gym you manage now stick — a background listing sync can't overwrite your name, address, or map pin anymore.

## Test plan

- `vp run typecheck` — green across all packages.
- `vp check` — clean for all changed files (the one remaining repo error is a pre-existing missing optional native module, `canvas`, in a file this PR doesn't touch).
- `vp test run` in `packages/location-sync` — 14 passed, including a new unit test proving the by-uuid fallback still links a frozen gym without overwriting it.
- Added `packages/location-sync/src/upsert.freeze.integration.test.ts` — opt-in real-DB test (needs a migrated PostGIS DB; self-skips otherwise) proving a frozen gym/board survives a re-sync while alias + link stay intact, and that an unfrozen synced gym still updates. Docker/DB isn't available in this environment, so it runs in CI / locally against a dev DB.

---
_Generated by [Claude Code](https://claude.ai/code/session_018fmP79ZrHTq4227yk4aDdJ)_

### DIFF
--- a/packages/backend/src/graphql/resolvers/social/boards.ts
+++ b/packages/backend/src/graphql/resolvers/social/boards.ts
@@ -1836,6 +1836,10 @@ export const socialBoardMutations = {
     // Build update values (only provided fields)
     const updateValues: Record<string, unknown> = {
       updatedAt: new Date(),
+      // A deliberate human edit — freeze the row so the location sync can never
+      // overwrite these curated values (a moderator's catalog fix, say) on its
+      // next run.
+      syncFrozenAt: new Date(),
     };
 
     if (validatedInput.name !== undefined) updateValues.name = validatedInput.name;
@@ -1979,7 +1983,12 @@ export const socialBoardMutations = {
       throw new Error('Not authorized to delete this board');
     }
 
-    await db.update(dbSchema.userBoards).set({ deletedAt: new Date() }).where(eq(dbSchema.userBoards.id, board.id));
+    // Freeze it too, so a later sync can't resurrect the board the owner
+    // deliberately removed (the board upsert clears deletedAt).
+    await db
+      .update(dbSchema.userBoards)
+      .set({ deletedAt: new Date(), syncFrozenAt: new Date() })
+      .where(eq(dbSchema.userBoards.id, board.id));
 
     return true;
   },
--- a/packages/backend/src/graphql/resolvers/social/gym-claims.ts
+++ b/packages/backend/src/graphql/resolvers/social/gym-claims.ts
@@ -113,7 +113,9 @@ export async function applyGymClaim(
     if (priorOwnerId !== claimantId) {
       await tx
         .update(dbSchema.gyms)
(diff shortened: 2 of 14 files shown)


marcodejongh:
Closing in favor of a focused follow-up. Part of this landed already: #3713 (merged) added the location-sync guarded 150 m match tier and the user-owned "alias-and-leave-alone" that skips `refreshSyncedGymMetadata` on non-SYSTEM gyms. The still-needed delta — the `sync_frozen_at` freeze column protecting moderator-fixed *still-system-owned* catalog rows, plus the claimed-gym resurrection / `findAliasedGymId` owner-check guard — is being ported into a new PR that credits this one. The branch is preserved; no work is lost.

PR #3832, opened 2026-07-21
feat(shared): render OG cards with the thumbnail stroke treatment

marcodejongh:
## Summary

Share cards render at chat-preview sizes where the standard hold-ring stroke weight gets hard to read. The OG variant now uses the same treatment thumbnails get in the Rust renderer — 8px strokes (vs 6px) and larger mirror markers (0.62×r vs 0.48×r) — at unchanged OG output scale. One-line change in the shared `buildRenderConfig`, so the backend `/og/climb` cards and the web fallback path both pick it up.

Before/after comparison rendered for the same climb: holds are clearly distinguishable at small sizes with the new treatment.

Since OG URLs are content-addressed (frames determine the image) and immutable-cached, previously scraped cards keep the old look on platform CDNs; new scrapes and new shares get the new rendering. The backend's in-memory caches clear on deploy.

## Release Notes

Shared climb cards are easier to read in chat previews — hold rings are thicker at small sizes

## Test plan

- New `render-config.test.ts` in `@boardsesh/board-render` pins: OG variant → thumbnail treatment at OG scale; plain thumbnail → `THUMBNAIL_WIDTH`; native → untouched
- `vp test run --project board-render` — 56 tests green
- `vp run test:web` — 5721 tests green (no web test pinned the OG thumbnail flag)
- `vp run typecheck` clean
- Rendered before/after locally via the backend service and visually compared

### DIFF
--- a/packages/shared/board-render/src/__tests__/render-config.test.ts
+++ b/packages/shared/board-render/src/__tests__/render-config.test.ts
@@ -1,0 +1,40 @@
+import { describe, expect, it } from 'vitest';
+import { HOLD_STATE_MAP } from '@boardsesh/board-constants/hold-states';
+import { getBoardDetailsForBoard } from '../board-details';
+import { buildRenderConfig, THUMBNAIL_WIDTH } from '../render-config';
+
+const kilterDetails = getBoardDetailsForBoard({
+  board_name: 'kilter',
+  layout_id: 1,
+  size_id: 10,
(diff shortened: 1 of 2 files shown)


marcodejongh:
Addressed in a7ff41e8e: the OG assertions now pin literal expected values (ogScale ≈ 0.4564, output_width 493 for the kilter 12×12 test config) instead of mirroring the formula.

PR #3834, opened 2026-07-21
perf: prime share-page and og-image caches at share time

marcodejongh:
## Summary

Closes the last slow leg in climb share unfurls. After #3829/#3830/#3833, repeat fetches of a shared link are fast — but a *freshly* shared URL is by definition new to the Vercel CDN, so the first crawler fetch still paid a full cold SSR (measured 1–2.4s). Crawlers scrape seconds after a share, and at share time we know the exact URL they'll fetch — so prime everything then:

- **Viewing a climb** (always precedes sharing it) now also warms the backend og image server-side, via the existing `scheduleOverlayWarming` fire-and-forget machinery (full variant only — list thumbnails don't fan out; the relative Vercel fallback URL is never warmed).
- **Tapping share on web** fires two fire-and-forget fetches before the share sheet opens: the share page URL (primes the CDN HTML entry) and the og image URL (`mode: 'no-cors'`, primes the backend caches).
- **Tapping share on mobile** does the same via plain RN `fetch`. The og URL is built by a small local helper — importing `@boardsesh/board-render` would drag the WASM renderer + sharp into the bundle. Raw URL parity with web isn't required: the backend canonicalises `set_ids` before keying its caches, so both platforms' URLs collapse to the same entry (QA-verified).

Priming is strictly best-effort: voided fetches, all failures (including rejections and a missing fetch) swallowed, the share sheet is never delayed. Cache-poisoning reviewed: a non-default-locale sharer's page fetch follows the sticky-locale 307 (which returns before cache headers attach), so it can warm the wrong-locale entry as a no-op but can never write a bad one.

## Release Notes

Shared climbs unfurl instantly in chats — the preview is ready before your link lands

## Test plan

- Web: `warm-overlay-cache.test.ts` (og warmed on full variant, not thumbnails, fallback skipped, failures swallowed) + `share-action.test.tsx` (both warm fetches fire with correct modes, sharing survives rejected fetches) — `vp run test:web` 5760 green
- Mobile: `use-share-climb.test.ts` extended (page + og warm, unsorted `set_ids` sorted in the URL, share sheet opens when prewarm rejects) — `vp run test:mobile` 4344 green
- `vp run typecheck:web` + `typecheck:mobile` clean; `vp run check:mobile-bundle` OK
- Post-deploy: share a climb from the app, then immediately curl the shared URL and its og:image — both should be cache HITs by the time a crawler would scrape

### DIFF
--- a/docs/og-climb.md
+++ b/docs/og-climb.md
@@ -54,6 +54,14 @@ repeats, ~700ms worst-case first render of a never-seen board config.
 | `OG_BYTE_CACHE_MB`      | `32`                  | Byte budget for the final-image LRU.                                                                                                                                               |
 | `OG_BASE_CACHE_ENTRIES` | `24`                  | Entry budget for the per-board-config base LRU (~3MB raw RGBA each).                                                                                                               |
 
+## Cache prewarming (why the endpoint sees browser-initiated hits)
+
+Crawlers scrape seconds after a share, so clients prime the caches ahead of
+them: climb view SSR fire-and-forgets one og render per page view (via
+`scheduleOverlayWarming`), and the Share button on web and mobile fetches both
+the share page URL and the og image URL before opening the share sheet. All
+best-effort — failures are swallowed and never delay sharing.
+
 ## Operational notes
 
 - Winston logs one line per render: `[OGClimb] served` with
--- a/packages/board-constants/src/hold-states.ts
+++ b/packages/board-constants/src/hold-states.ts
@@ -279,3 +279,11 @@ export function convertLitUpHoldsStringToMap(litUpHolds: string, board: BoardNam
       {} as Record<number, LitUpHoldsMap>,
     );
 }
+
+/** Collapse a possibly multi-frame frames string to its final lit snapshot. */
+export function toFlatFrames(frames: string | null | undefined, board: BoardName): string {
+  if (!frames) return '';
+  if (!frames.includes(',') && !frames.includes('x')) return frames;
+  const maps = accumulateFramesToMaps(frames, board);
+  return accumulatedMapsToFrameStrings(maps, board).at(-1) ?? '';
+}
--- a/packages/mobile/src/hooks/__tests__/use-share-climb.test.ts
+++ b/packages/mobile/src/hooks/__tests__/use-share-climb.test.ts
@@ -88,6 +88,8 @@ describe('useShareClimb', () => {
         await result.current();
       });
       expect(fetchMock).toHaveBeenCalledWith(expectedReadableShareUrl);
+      // Exactly one warm: the no-frames climb must not fire an og request.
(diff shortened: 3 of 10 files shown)


marcodejongh:
Review addressed in 5124a86fc:

- **Test timing**: the fire-and-forget flushes now use `vi.waitFor` — the both-fetches test waits for the exact call count (2), so a slow second fire can't flake it.
- **Mobile call-count**: the no-frames share test now asserts exactly one warm fetch.
- **`toFlatFrames` drift risk**: the mobile mirror's comment now points at its web source (`packages/web/app/components/board-renderer/util.ts`) with a keep-in-sync note.
- **`boardDetails` dep**: intentionally kept — the callback reads the whole object (`buildOgBoardRenderUrl(boardDetails, …)`), so `boardDetails.board_name` alone would be a stale-closure bug. Perf impact is negligible: this is an action-menu button, not a virtualized row's `renderItem`, and callback identity churn here re-renders nothing memoized.

marcodejongh:
Agreed on hoisting `toFlatFrames` into `@boardsesh/board-constants` (both platforms already import its two primitives from there) — leaving it as a follow-up rather than growing this PR: web's copy lives in `board-renderer/util.ts` whose test mocks intercept by module path, so the move deserves its own small change. The keep-in-sync comments on both sides bridge until then.

marcodejongh:
All five items addressed in 0fca47a24:

- **Duplication**: `toFlatFrames` extracted into `@boardsesh/board-constants/hold-states` beside its two primitives; web delegates (export + module path preserved so route-test mocks keep intercepting) and mobile imports it — the mirror is gone.
- **Docstring**: condensed to one block with a single-line locale note.
- **Dep array**: the og URL is now computed at render scope, so the callback deps went back to primitives (`ogImageUrl` string + `boardDetails.board_name`).
- **API inconsistency**: mobile's `prewarmShareCaches` now takes `urls: string[]`, same as web.
- **Opaque body cancel**: the no-cors branch returns without touching the body.

marcodejongh:
Addressed in the latest commit:

- **frames type**: `buildOgImageUrl` now accepts `string | null | undefined`, matching the shared `toFlatFrames` guard (null → no og warm).
- **docs**: `docs/og-climb.md` gained a "Cache prewarming" section explaining the browser-initiated hits.
- **toFlatFrames boundary doc**: restored a concise version on the web wrapper (why flattening matters before the renderer/BLE boundary).
- **full-variant ceiling**: documented — both call sites pass a single climb, so one og warm per view render.
- **Test call-order (item 5)**: no change needed — the test uses `toHaveBeenCalledWith` (order-independent, matches any recorded call) plus an exact `toHaveBeenCalledTimes(2)`; `Promise.allSettled` ordering can't flake it. Only `toHaveBeenNthCalledWith` would be order-sensitive, and it isn't used there.

PR #3837, opened 2026-07-22
feat(repo): manage the boardsesh.com Cloudflare config from the repo

marcodejongh:
## Summary

Cloudflare config-as-code for the `boardsesh.com` zone — dashboard clicks become reviewable, versioned config. Rebased onto `main` (it was 1,346 commits behind) and extended to cover the www crawl-cost controls from #4650.

- **`infra/cloudflare/config.ts`** — desired state: the `ws` record proxied, the cache rules, the WAF crawler rules, and a strict-SSL assertion.
- **`vp run cf:apply`** — dry-run by default: diffs live zone state against the config, prints the plan, exits non-zero on drift. `--apply` mutates only the deltas. Foreign rules in every managed phase are preserved verbatim; ours are identified by stable description markers. `--allow-zone-ssl` gates the zone-wide SSL change since other hostnames share it.
- **Cutover-safe ordering**: SSL and the cache rules apply before the proxied flip, and the flip is held back entirely while a required SSL change is blocked — orange-clouding with weak SSL would redirect-loop every host on the record, including the party-mode WebSocket.

### New in this revision: the www cost controls

Measured against production on 2026-08-25 with the Vercel MCP — the data #4717 had to leave as an empty column, because that agent had no Vercel credentials:

| | per day |
| --- | --- |
| Vercel function invocations | ~442,600 |
| — `/api/internal/board-render` | ~215,600 (48.7%) |
| — climb view `…/view/[climb_uuid]` | ~203,900 (46.1%) |
| — `/setter/[setter_username]` | ~15,900 (3.6%) |
| — the other 48 routes combined | ~7,100 (1.6%) |
| Homepage `/` | ~2,000 |

Two routes are 94% of compute, against a published sitemap of ~60k climb URLs and ~2,000 homepage hits. That shape is a crawl, not an audience.

**Board-render cache rule.** The route already sends `cache-control: public, max-age=31536000, immutable` and a matching `CDN-Cache-Control`, but Cloudflare caches by **file extension** by default and `/api/internal/board-render` has none — so it measured `cf-cache-status: DYNAMIC` while `/_next/static/*.js` on the same zone was a `HIT`. Every image byte was transiting Cloudflare to Vercel (~54 GB/day). The rule is the whole fix; no origin header change.

**Crawler rules.** A `skip` rule for search engines and share-card unfurlers, then a `block` rule for commercial SEO crawlers (Ahrefs, Semrush, DataForSEO, MJ12, DotBot, BLEXBot, Barkrowler, serpstat, Seznam, Zoominfo, Screaming Frog) — each verified reaching our origin by #4716's 2026-08-24 probe. They sell backlink data and send Boardsesh no traffic. Brave is allowlisted explicitly because it runs its own index rather than reselling Bing or Google.

### Three traps this had to design around

1. **WAF rule order is load-bearing, and the naive upsert gets it backwards.** A rule-by-rule upsert appends a newly added allow rule *after* an already-present block rule, reversing precedence and blocking Googlebot the moment the config lands. `upsertCacheRule` now rewrites our rules as one contiguous group in declared order. Mutation-tested: restoring the per-rule version reds two tests.

2. **`lower()` is required, not stylistic.** Cloudflare's `contains` is case-sensitive, so `contains "AhrefsBot"` installs cleanly and matches nothing — a rule that looks done and does nothing. A test pins every comparison to the lowercased form; removing `lower()` reds it.

3. **`cf.client.bot` is the wrong allowlist.** Ahrefs and Semrush are themselves Cloudflare *verified bots*, so that field is true for precisely the crawlers we are blocking. The allowlist is UA-based on purpose, and the file says why.

The AI crawlers (ClaudeBot, GPTBot, PerplexityBot, Bytespider, CCBot…) are deliberately **absent** from our block list: the same probe got `403` from `server: cloudflare` for every one, so a zone rule we do not own already stops them. That rule is foreign to this tool and is preserved verbatim — there is a test asserting a foreign WAF rule survives the phase PUT, because the rulesets API replaces the entire phase and dropping it would silently reopen the site to every AI scraper.

### CI wiring

`deploy-cloudflare` runs on `main` from the `Production` environment (which already has `CLOUDFLARE_API_TOKEN` and `CLOUDFLARE_ACCOUNT_ID`), only when `infra/cloudflare/**` or the apply tooling changes. Nothing else depends on the job, so a Cloudflare API hiccup never blocks a web or backend deploy.

Main replaced the workflow's inline change-detection bash with `scripts/production-deploy-changes.mjs` since this branch was written, so the `cloudflare` target moved into that script. It is deliberately **not** web-affecting: `isWebAffecting` is a denylist, so without an explicit exclusion every zone edit would queue a full production web deploy for nothing. Both directions are tested.

⚠️ **The `Production` environment is main-only**, so the first real `cf:apply` cannot run from this branch — it happens on merge. Run the dry-run locally first (below) and read the plan before merging.

## Release Notes

## Test plan

- `vp test run --project scripts` — 33 tests in `scripts/cloudflare-apply.test.ts` (was 20), covering the new rules, ordering, foreign-rule preservation, and the token lists.
- `node --test scripts/production-deploy-changes.test.mjs` — 21 tests (was 19); the two new ones pin that a zone-config edit converges Cloudflare *and nothing else*, and that a `packages/` change never drags the apply along with it.
- `vp run typecheck` — clean. `vp check` — clean on the changed files.
- **Mutation-tested three guards.** Per-rule upsert instead of the group rewrite → 2 failures. `lower()` removed → 1 failure. Managed order taken from the zone instead of the config → 1 failure.

### Before merge

```bash
CLOUDFLARE_API_TOKEN=… vp run cf:apply     # dry-run; prints the plan, mutates nothing
```

**Read the plan before merging — one line in it needs a human decision.** Rules are
identified by a description marker, and the `/og/` rule currently on the zone was created
*by hand* in July, so it will not carry ours. The dry-run will therefore report the og rule
as **missing**, and applying would create a second `/og/` cache rule alongside the manual
one rather than adopting it.

Both rules say "cache" and Cloudflare takes the first match, so a duplicate is untidy rather
than dangerous — but it should not be merged blind. Two ways to settle it, whichever you
prefer:

- Rename the existing rule's description in the dashboard to
  `boardsesh:og-edge-cache (managed by scripts/cloudflare-apply.ts)` before merging, and the
  tool adopts it in place, keeping its id and position.
- Or let the apply create ours and delete the manual one afterwards.

The DNS proxied flag and the SSL mode are plain zone state with no marker, so those two
really should report as already in sync.

### After merge

```bash
B='https://www.boardsesh.com/api/internal/board-render?board_name=kilter&layout_id=1&size_id=10&set_ids=1,20&frames=p1085r15p1128r12&include_background=1'
curl -sI "$B" | grep -i cf-cache-status   # MISS
curl -sI "$B" | grep -i cf-cache-status   # HIT

for UA in "AhrefsBot/7.0" "SemrushBot/7~bl" "Brave-Search/1.0" "Googlebot/2.1" "Bingbot/2.0"; do
  printf '%-18s %s ' "$UA" "$(curl -sS -o /dev/null -A "$UA" -w '%{http_code}' https://www.boardsesh.com/)"
  curl -sSI -A "$UA" https://www.boardsesh.com/ | grep -ci '^x-vercel-id' | sed 's/^/reached-vercel:/'
done
```

Ahrefs/Semrush → `403 reached-vercel:0`; Brave/Google/Bing → `200 reached-vercel:1`. Then confirm compute actually fell:

```
get_runtime_logs since=6h group_by=route source=["serverless"] environment=production
```

> `since: 24h` is silently truncated (249,960 vs 442,564 from 6 h × 4). Use a 6 h or 30 min window and scale, or you will under-report by ~2×. Documented in `docs/cloudflare.md`.

Part of #4650 / epic #4648. Pairs with #4764, which stops crawler page renders triggering a board-render in the first place.

### DIFF
--- a/.github/workflows/production-deploy.yml
+++ b/.github/workflows/production-deploy.yml
@@ -546,6 +546,9 @@ jobs:
   # nothing else depends on it, so a Cloudflare API hiccup never blocks the
   # web/backend deploys. A non-zero exit = unapplied drift (e.g. a blocked
   # zone-SSL change) and should be investigated, not retried blindly.
+  # The planned OpenNext (Cloudflare Workers) deploy of packages/web will be a
+  # sibling job reusing the same Production-environment CLOUDFLARE_API_TOKEN
+  # (+ CLOUDFLARE_ACCOUNT_ID) — see docs/cloudflare.md for the token scopes.
   deploy-cloudflare:
     needs: [detect-changes]
     if: needs.detect-changes.outputs.cloudflare_changed == 'true'
--- a/.github/workflows/production-deploy.yml
+++ b/.github/workflows/production-deploy.yml
@@ -74,6 +74,7 @@ jobs:
       backend_changed: ${{ steps.changes.outputs.backend }}
       app_changed: ${{ steps.changes.outputs.app }}
       deployment_base_sha: ${{ steps.changes.outputs.deployment_base_sha }}
+      cloudflare_changed: ${{ steps.changes.outputs.cloudflare }}
     steps:
       - uses: actions/checkout@v6
         with:
@@ -540,6 +541,29 @@ jobs:
   # detect-changes. No inline Playwright smoke gate — a per-deploy expo-web boot
   # is too heavy here (this file has no such gate for web/backend either); the
   # expo-web-smoke suite stays the dispatch-only gate in e2e-tests.yml.
+  # Converges the Cloudflare zone config (infra/cloudflare) on main. Runs only
+  # when the desired state or apply tooling changes (or on manual dispatch);
+  # nothing else depends on it, so a Cloudflare API hiccup never blocks the
+  # web/backend deploys. A non-zero exit = unapplied drift (e.g. a blocked
+  # zone-SSL change) and should be investigated, not retried blindly.
+  deploy-cloudflare:
+    needs: [detect-changes]
+    if: needs.detect-changes.outputs.cloudflare_changed == 'true'
+    runs-on: ubuntu-latest
+    environment: Production
+    steps:
+      - uses: actions/checkout@v6
+
+      - uses: oven-sh/setup-bun@v2
+
+      - name: Install dependencies
+        run: bun install --frozen-lockfile
+
+      - name: Apply Cloudflare config
+        env:
+          CLOUDFLARE_API_TOKEN: ${{ secrets.CLOUDFLARE_API_TOKEN }}
+        run: vp run cf:apply -- --apply
+
   # Manual deploy freeze for app.boardsesh.com, the Cloudflare Pages analogue of
   # check-rollback. deploy-web and deploy-production-backend both honour a pinned
   # Vercel Instant Rollback; Pages has no equivalent signal here, so without this
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -120,6 +120,7 @@ Read relevant `docs/` before working on the matching area; update docs when the 
 - `docs/logging.md` — backend structured logger (winston)
 - `docs/db-connectivity.md` — Postgres connect retries (what is retried and why it can't double-execute a write), the retry budgets, and the `/health` vs `/health/db` split
 - `docs/og-climb.md` — backend-served climb OG share cards (`GET /og/climb`: caches, env vars, timings)
+- `docs/cloudflare.md` — the boardsesh.com Cloudflare zone: config-as-code (`vp run cf:apply`), token scopes/secrets, CI auto-apply, og edge caching, planned OpenNext web deploy
 - `docs/mobile-sheets-vs-routes.md` — mobile: which surface to use (bottom sheet vs route), with the decision tree + the hard rules (incl. why `fullScreenModal` breaks the iOS 26 native tab bar)
 - `docs/gym-funnel-analytics.md` — the www gym funnel event contract in `@boardsesh/analytics` (seven event names, their property sets, the QR `?src=qr&medium=` landing params, and why `boardTypes` must be a joined string)
 - `docs/sitemap.md` — the shard registry, the degrade-at-the-index / fail-closed-at-the-shard split, and the climb store (`sitemap_shard_refreshes` + the `sitemap_climb_urls` ordinal table the shard pages read): who refreshes it, the `?force=1` escape hatch, and why the write lock is transaction-scoped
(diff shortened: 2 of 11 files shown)


marcodejongh:
Review addressed in 08b380e0d: zone resolution now throws on an unmatched name instead of falling back to the first visible zone; unknown CLI flags are rejected loudly (a typoed `--appply` can no longer silently dry-run); the sequential no-rollback apply semantics are documented at the apply loop (safe because the plan is ordered and re-running converges); and the unknown-SSL-mode drift path is covered by a test. 22 tests green.

marcodejongh:
Added CI wiring per Marco's request (326476509 + bcd9ca96c):

- `detect-changes` gains a `cloudflare` filter (`infra/cloudflare/*` + the apply script; manual dispatch and no-before-sha runs include it).
- New non-blocking `deploy-cloudflare` job (Production environment) runs `vp run cf:apply -- --apply` — nothing depends on it, so a Cloudflare API hiccup can't hold up web/backend deploys. A failing job = unapplied drift worth a look.
- Runbook documents the flow and the secret location.

**One manual step before merging** (or the job fails on its first run): create the scoped API token (Zone.Zone Read, Zone.DNS Edit, Zone.Cache Rules Edit, Zone.Zone Settings Read) and add it to the Production environment:
```
gh secret set CLOUDFLARE_API_TOKEN --env Production
```
Note for the first run: the config asserts zone-wide SSL `strict`. If the manual setup used a per-host configuration rule instead (with the zone on Full/Flexible), the job will report blocked SSL drift and exit non-zero — that's the drift signal working; we'd then either set zone-wide strict or model the per-host rule in the config.

marcodejongh:
Provisioned for the planned OpenNext (Vercel → Cloudflare Workers) move in c9b227e7f, so the secret setup only happens once:

- Cloudflare docs split into **`docs/cloudflare.md`** (zone config-as-code, token setup, CI auto-apply, og edge caching, and the OpenNext plan); `og-climb.md` points there.
- The one-time token guidance now covers **both** uses — today's zone scopes plus the wrangler deploy scopes (Workers Scripts Edit, Workers KV Edit, R2 if used, Workers Routes Edit) — and adds `CLOUDFLARE_ACCOUNT_ID` to the Production environment alongside the token:
  ```
  gh secret set CLOUDFLARE_API_TOKEN --env Production
  gh secret set CLOUDFLARE_ACCOUNT_ID --env Production
  ```
- The workflow's `deploy-cloudflare` job comment notes the future `deploy-web-cloudflare` sibling reusing the same environment secrets.

PR #3842, opened 2026-07-22
feat(web): gym comments in the manage console

marcodejongh:
## Summary

Adds a **Comments** tab to the gym manage console so owners and editors can read and reply to their gym's comment thread without visiting the public gym page (and finding themselves in search).

The tab mounts the exact same `CommentSection` the public gym page already uses — same gym entity (`entityType="gym"`, `entityId={gym.uuid}`), same GET_COMMENTS query, same `commentUpdates` live subscription, same reply mutation. No new comment infrastructure or query was built; it's the same thread, just surfaced in the console. Reply works for any editor because the tab lives inside the existing `canEdit`-gated console, so no new auth was added.

`CommentSection` (via `CommentList`) already renders the live comment count and the empty state from its own query + subscription, so the tab stays thin: a short console-context intro line plus a stable topic heading handed to `CommentSection`. Nothing in the tab is derived from a load-time snapshot, so the header never goes stale once a comment lands.

### Files changed
- `packages/web/app/components/gym-entity/manage/comments-tab.tsx` (new) — the tab; reuses `CommentSection`.
- `packages/web/app/gym/[gym_slug]/manage/manage-gym-content.tsx` — register the tab (minimal: one import, one `ManageTab`/`VALID_TABS` entry, one `Tab`, one render branch).
- `packages/web/app/components/gym-entity/manage/__tests__/comments-tab.test.tsx` (new) — entity wiring, stable-heading regression guard, intro copy.
- `packages/web/app/gym/[gym_slug]/manage/__tests__/manage-gym-content.test.tsx` — tab registration/render.
- `packages/shared/i18n/locales/{en-US,es,fr}/kiosk.json` — new strings ×3 locales (es: rocódromo/equipo, fr: salle/crew).

## Release Notes

- See what your crew's saying about your gym, and reply, right from the manage console.

## Test plan

- `vp run typecheck` — pass
- `vp test run --project web --reporter=agent` — full web suite green (incl. comments-tab + shell registration)
- `vp run check:i18n` + `vp run check:i18n:orphans` — pass
- `vp check` on the changed files — pass (pre-commit hook green)

### DIFF
--- a/packages/shared/i18n/locales/en-US/kiosk.json
+++ b/packages/shared/i18n/locales/en-US/kiosk.json
@@ -363,9 +363,8 @@
       "description": "Your gym's name, address, and what climbers see on the public page."
     },
     "comments": {
-      "intro": "See what your crew's saying and reply, right from here.",
-      "count_one": "{{count}} comment",
-      "count_other": "{{count}} comments",
+      "intro": "Reply to your gym's comment thread without leaving the console.",
+      "heading": "What your crew's saying",
       "empty": "No one's chimed in yet."
     }
   },
--- a/packages/shared/i18n/locales/en-US/kiosk.json
+++ b/packages/shared/i18n/locales/en-US/kiosk.json
@@ -364,8 +364,7 @@
     },
     "comments": {
       "intro": "Reply to your gym's comment thread without leaving the console.",
-      "heading": "What your crew's saying",
-      "empty": "No one's chimed in yet."
+      "heading": "What your crew's saying"
     }
   },
   "branding": {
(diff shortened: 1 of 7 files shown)


marcodejongh:
Addressed the stale-header finding at the root: the tab no longer derives any header text from the load-time `gym.commentCount`. `CommentSection` already renders the accurate live count and empty state from its own query + subscription, so the tab now hands it a stable topic heading and lets it own that state. That removes every stale-snapshot path (first post, in-tab post, and subscription-driven posts from other users), not just the numeric one. Kept the tab thin (intro context + the shared `CommentSection`), per the "reuse, don't rebuild" scope.

Issue #3529, opened 2026-07-07
MoonBoard ticks at a non-catalog angle create phantom stats rows (wrong-angle, quality_normalized=false) invisible on the canonical climb


marcodejongh:
## Summary
The MoonBoard model keeps one climb row per (problem, angle), but a tick can be saved at any angle against any climb UUID. recomputeClimbStats' defensive seed then inserts a stats row at that angle (upstream=0, boardsesh=N) even though no moonboard climb row exists at that angle. Prod has 42 moonboard stats rows whose angle differs from the climb row's angle (e.g. layout-5 climbs stored at 40deg with a stats row at 25deg), and 44 moonboard rows with quality_normalized=false (the seeded rows default the flag off), breaking the 'all moonboard rows are normalized' invariant every moonboard writer otherwise maintains.

## Failure scenario
A user on a 25deg Masters 2019 board ticks a problem whose Boardsesh row is the 40deg variant (UUID from search/playlist). The recompute seeds (moonboard, uuid, 25) with boardsesh count 1. The climb page/search read stats at the climb's own angle (40), so the send never appears in that climb's ascent count; the 25deg twin of the problem (separate UUID) doesn't get it either. The count is stranded on a row no surface reads consistently.

## Evidence
`packages/backend/src/graphql/resolvers/ticks/recompute-climb-stats.ts:91`

```
recompute-climb-stats.ts:86-107 inserts a minimal row at the tick's angle unconditionally. Prod: SELECT count(*) FROM board_climb_stats s JOIN board_climbs c ON c.board_type=s.board_type AND c.uuid=s.climb_uuid WHERE s.board_type='moonboard' AND s.angle <> c.angle -> 42; all have upstream_ascensionist_count=0, boardsesh_ascensionist_count>=1, quality_normalized=false.
```

## Suggested fix
For moonboard, resolve the tick to the climb row's canonical angle (or to the sibling climb UUID at the ticked angle) before recompute; alternatively have the defensive seed set quality_normalized=true and add a moonboard-specific guard that redirects stats to the climb's angle. Clean the 42 stranded rows by folding their boardsesh counts into the canonical-angle row.

Found by the 2026-07 upstream-sync data audit (verified against prod). Related program PRs: ascent-count fix, star-scale repair (in flight).

marcodejongh:
[...]

PR #3849, opened 2026-07-23
feat(db): dedup MoonBoard's per-angle climb duplication

fallaciousreasoning:
## Summary

Merges MoonBoard's per-angle duplicate catalog climbs (`moonboard:{id}:25` / `moonboard:{id}:40` pairs) onto one canonical climb per problem, via data migration `0193_moonboard_angle_dedup_backfill.sql`. Companion to #3851 (the angle-agnostic importer, merged) — together they retire angle-bearing MoonBoard climb identity, matching Kilter/Tension.

- Groups computed from **live hold signatures** (`board_climb_holds`), catalog rows only (`user_id IS NULL`); only groups whose members all sit at **distinct angles** auto-merge — same-angle collision groups are counted and left untouched for manual follow-up.
- Losers are **delisted, never deleted**; every referencing table is repointed (ticks, favorites, playlists, circuits, votes, beta links, classic/community status, proposals, comments, feed items, notifications, stats history, events, send stats, grades, embeddings/similarity caches dropped for rebuild).
- **Single-owner stats contract:** step 2 places rows and merges the upstream half; step 3b recomputes `boardsesh_ascensionist_count`, `boardsesh_quality_sum/count`, `quality_average` (blend) and `ascensionist_count` from the repointed ticks for **every key step 2 writes** — ported faithfully from `recomputeClimbStatsBulk()`. The materialized invariant `ascensionist_count = upstream + boardsesh` holds on every row at every step.
- **Offline convergence:** hand-written `sync_deletions` tombstones for the repointed `user_favorites` / `playlist_climbs` client keys (their client PKs move without firing delete triggers), trigger-emitted tombstones for deleted stats rows, cursor-bump guarantees asserted in the fixture. Two accepted residuals documented in the header and tracked in #4389.
- **Collision survivor policies** (per-table, each on its own `ORDER BY`, one-line rationale in SQL): votes → user's latest vote; favorites/playlists → earliest added; circuits → lowest position; classic status → canonical's own, then `is_classic`; beta links → canonical's own, then has `tick_uuid`, then earliest; community status → canonical's own, then highest `last_proposal_id`; grades → canonical's own, then earliest computed.
- `saveTick` now resolves `board_climb_aliases` before insert, so ticks from stale offline catalogs land on the canonical instead of re-orphaning onto retired uuids. Read paths (backend `get-climb`, web SSR queries) resolve aliases too.
- Guarded by a durable `_bs_migration_guards` row; atomic (verified by deterministic mid-flight kill: full rollback including guard + ledger).

## ✅ Maintainer sign-offs (all four granted 2026-08-12/13)

1. **Prod sizing** (read-only, maintainer-approved, recorded in the migration header): 8,977 mergeable groups — ALL two-member 25°/40° pairs, zero larger groups; 494 same-angle groups left for manual follow-up; losing members carry 359 ticks, 30 favorites, 13 playlist rows, 0 votes, 0 comments, 0 grade rows. Dev-db (766/478) is the same shape at ~1/12 scale.
2. **Survivor policies** approved as implemented.
3. **Deploy order vs #4054:** satisfied — #4054's stack merged 2026-08-13 carrying `0191_backfill_moonboard_ascensionist_invariant` + `0192_moonboard_wrong_angle_stats_cleanup`; this migration is renumbered above them as `0193` (by filename number and journal `when` alike). The saveTick alias-resolution and #4054's angle snap are composed: resolve alias first, snap the angle against the canonical.
4. **Prod data mutation approved** — admin-merge on green authorized.

## Provenance

Original implementation by @fallaciousreasoning (all five 2026-07-24 review threads addressed on-branch). Reworked 2026-08-12 for the maintainer's takeover concerns (count-invariant preservation, quality provenance, offline tombstones, collision policy) plus the findings of a 35-agent adversarial review (15 confirmed → fixed; 15 refuted) and two independent dev-db replays (766 groups merged / 478 same-angle skipped / 0 errors, idempotency + crash-window verified on a clean container). Fixture: 46 checks, 12-mutation kill matrix.

## Release Notes

## Test plan

- `packages/backend/src/__tests__/moonboard-angle-dedup-replay.test.ts` — scratch-Postgres replay of the real SQL: 46/46, including recompute parity (CASE F built so SUM and GREATEST are both provably wrong), stale-half repair (CASE G), alias-vs-alias folds, tombstone-safety anti-join, cursor-bump assertions, byte-faithful trigger ports.
- `vp test run --project backend` — 3341 tests, 0 failures.
- Clean-container dev-db replay via `vp run db:migrate`: 766/766/478, 0 errors, invariant checked across all 677k stats rows on all 7 board types.
- `vp run typecheck` (all packages), `vp check`, `vp run check:db-migrations` — green.


fallaciousreasoning:
Reopening to pick up the corrected Release Notes checkbox for CI.

packages/db/drizzle/0185_moonboard_angle_dedup_backfill.sql
```diff
@@ -0,0 +1,452 @@
...
+    FROM _mad_map m WHERE f.entity_type = 'climb' AND f.entity_id = m.alias_uuid;
+
+  UPDATE notifications n SET entity_id = m.canonical_uuid
+    FROM _mad_map m WHERE n.entity_type = 'climb' AND n.entity_id = m.alias_uuid;
+
+  -- 4. Repoints where the uniqueness constraint does NOT include angle, so a
+  --    real collision is possible (a user interacted with both angle-rows of
+  --    the same problem): drop the losing duplicate, then repoint the
+  --    survivor. Same shape as 0163's playlist_climbs step.
+  DELETE FROM playlist_climbs pc USING _mad_map m
```
marcodejongh: **Reproducible crash on ≥3-member groups.** Every dedupe-then-repoint step in 4/5/6 only checks for a collision against the **canonical** row. When a group has two or more non-canonical members and a user interacted with two of them (but not the canonical), both rows survive the DELETE and the UPDATE then repoints them onto the same unique key.

I replayed this against a scratch Postgres 17 using this PR's own fixture schema plus a 3-angle group (`w25`/`w40`/`w55`, canonical `w40`) where one user playlisted `w25` and `w55`:

```
MIGRATION FAILED: duplicate key value violates unique constraint
"playlist_climbs_playlist_id_climb_uuid_key"
```

Same shape bites `votes` (one user voted on two losing members), `board_circuits_climbs`, `climb_classic_status` (unique on `(climb_uuid, board_type)` — two aliases each having a row collide even in the absence of a canonical row), `board_beta_links` (same link on two aliases), and the angle-scoped tables when two aliases carry a row at the same angle.

The DO block rolls back atomically so there's no partial damage — but the deploy fails, and the header's read-only sizing query won't predict it because it depends on user interactions, not just group shapes.

Two ways out:
1. Dedupe among **all** group members: for each unique target key, keep one deterministic winner across `{canonical + all aliases}` (e.g. `ROW_NUMBER()` partitioned by the post-repoint key) and delete the rest before repointing.
2. Or restrict auto-merge to `member_count = 2` groups and fold larger groups into the skipped count — matching the actual prod shape (25°/40° pairs) and deferring N-angle groups to the follow-up work.

Either is fine, but CASE D in the fixture explicitly advertises 3-angle support, so code and contract should agree — and if you keep N-angle support, CASE D should gain playlist/vote rows on the two losing members so this path is covered.

packages/db/drizzle/0185_moonboard_angle_dedup_backfill.sql
```diff
@@ -0,0 +1,452 @@
...
+           COUNT(*) FILTER (WHERE v.value = 1)  AS upvotes,
+           COUNT(*) FILTER (WHERE v.value = -1) AS downvotes
+      FROM (SELECT DISTINCT canonical_uuid FROM _mad_map) ai
+      LEFT JOIN votes v ON v.entity_type = 'climb' AND v.entity_id = ai.canonical_uuid
+     GROUP BY ai.canonical_uuid
+  )
+  INSERT INTO vote_counts (entity_type, entity_id, upvotes, downvotes, score, hot_score, created_at)
+  SELECT 'climb', agg.canonical_uuid, agg.upvotes, agg.downvotes, agg.upvotes - agg.downvotes,
+         SIGN(agg.upvotes - agg.downvotes) * LN(GREATEST(ABS(agg.upvotes - agg.downvotes), 1))
+           + EXTRACT(EPOCH FROM (COALESCE(created.created_at, now()) - TIMESTAMP '2005-12-08')) / 45000.0,
```
marcodejongh: **Formula mismatch with the live trigger — merged climbs sink ~25,200 points in hot ordering.** The actual trigger (`update_vote_counts()` from 0053, re-created in 0130) computes:

```sql
EXTRACT(EPOCH FROM v_created_at) / 45000.0
```

plain Unix epoch — there is no `2005-12-08` reference anywhere in the schema (and no `refresh_vote_counts()` function; the header cites a name that doesn't exist). Subtracting the Reddit epoch here shifts every merged climb's `hot_score` down by a constant ~25,200 relative to every trigger-computed row, which breaks relative ordering under `vote_counts_hot_score_idx`. It also means the first post-merge vote on such a climb makes its `hot_score` jump by +25k when the trigger rewrites it.

The repo already has the canonical recipe for exactly this operation — `rebuildGymVoteCounts()` in `packages/db/src/queries/gyms/merge-gyms.ts`:
- plain-epoch formula matching the trigger,
- `created_at` preference order `feed_items.created_at → MIN(votes.created_at) → NOW()` (this migration uses `MIN(vote_counts.created_at)`, which diverges from what the trigger will compute next time it fires),
- and it sets `boardsesh.skip_vote_counts = 'on'` around the bulk vote DELETE/UPDATE — this migration doesn't, so step 6 fires `votes_count_trigger` per row (wasted per-row recounts, and the trigger's zero-vote cleanup can interleave with the recompute).

Suggest copying that recipe wholesale. The fixture assertion (`moonboard-angle-dedup-replay.ts`, the `hot_score` check) encodes the same wrong formula and the fixture schema omits the trigger entirely — both worth fixing alongside so the replay actually exercises the trigger interplay.

packages/db/drizzle/0188_moonboard_angle_dedup_backfill.sql
```diff
@@ -0,0 +1,452 @@
...
+    fa_at = COALESCE(board_climb_stats.fa_at, excluded.fa_at),
+    quality_normalized = board_climb_stats.quality_normalized OR excluded.quality_normalized,
+    upstream_synced_at = GREATEST(board_climb_stats.upstream_synced_at, excluded.upstream_synced_at);
+
+  -- The row just copied onto the canonical uuid now fully duplicates the
+  -- retiring uuid's own stats row (same board_type/angle, identical values) —
+  -- drop the original so nothing that scans board_climb_stats broadly (a
+  -- leaderboard, a bulk recompute) double-counts a delisted climb's ascents.
+  -- Safe: every value has already been copied forward by the INSERT above.
+  DELETE FROM board_climb_stats s USING _mad_map m
```
marcodejongh, line 296: **Stats rows at non-native angles are left orphaned under the delisted uuid.** Both the re-key JOIN above (`m.angle = s.angle`) and this DELETE only touch the alias's stats row at its own `board_climbs.angle`. But `board_climb_stats` rows are also minted by the tick recompute, which keys on `(board, climb, angle)` from `boardsesh_ticks` — so if anyone has ever ticked a per-angle MoonBoard row at an angle other than the row's native one, that stats row: (a) stays under the delisted uuid (double-counted by anything scanning `board_climb_stats` broadly — the exact thing this DELETE exists to prevent), and (b) never reaches the canonical, even though step 3 repoints its underlying **ticks** unconditionally.

Suggest dropping the `AND m.angle = s.angle` from both the INSERT's JOIN and this DELETE, i.e. re-key **all** of the alias's stats rows preserving `s.angle`. The existing `ON CONFLICT` clause already merges conservatively (COALESCE/GREATEST), so a genuine collision with the canonical's row at the same angle is handled — and after the change the "defensive only" comment on it becomes actually load-bearing.

packages/db/drizzle/0188_moonboard_angle_dedup_backfill.sql
```diff
@@ -0,0 +1,452 @@
...
+-- transaction there is nothing but the guard row below to stop a
+-- re-application.
+--
+-- Prod group/member counts are NOT pre-verified for this migration (no live
+-- prod DB access at authoring time, and the standard dev-db seed uses an
+-- older single-row-per-problem importer that doesn't reproduce this
+-- duplication — see import-moonboard-problems.ts). Every step below is
+-- written to be correct for any group size/count, but run a read-only sizing
+-- query against prod (see the _mad_groups/_mad_raw_groups shape below) before
+-- deploying, and update this comment with the numbers per repo convention.
```
marcodejongh: **Sequencing: the next catalog import partially undoes this merge.** Today's `import-moonboard-catalog.ts` still matches on `${layoutId}|${angle}|${fingerprint}` and `buildExistingCatalogMatchIndex` skips rows with `isListed !== true` (and requires the canonical itself to be listed under that angle key). Post-0185 the canonical only appears under its own angle's key, so an incoming 25° problem finds no match and re-mints `moonboard:{id}:25`:

- the `board_climbs` upsert only sets `characteristics`/`description` on conflict, so the row stays delisted — good;
- but the alias upsert uses `catalogAliasConflictUpdate()` (`canonical_uuid = excluded.canonical_uuid`), which **overwrites this migration's `moonboard:{id}:25 → canonical` alias back to self-pointing**, re-splitting logbook/tick resolution for every merged problem;
- and the stats upsert re-creates a 25° stats row under the delisted uuid (the migration deleted it, so there's no conflict), while the canonical's migrated 25° row goes stale — ascents double-counted again.

If the importer goes angle-agnostic in #3851, this is purely an ordering constraint — but it's worth stating explicitly in this header ("do not run the catalog import between 0185 and the #3851 importer"), or better, making the importer refuse to touch a uuid that exists as a non-self alias, so an accidental import between the two deploys can't quietly corrupt the merge.

packages/db/drizzle/0188_moonboard_angle_dedup_backfill.sql
```diff
@@ -0,0 +1,452 @@
...
+    last_sent_at = GREATEST(board_climb_send_stats.last_sent_at, excluded.last_sent_at),
+    updated_at = now();
+
+  DELETE FROM board_climb_send_stats s USING _mad_map m
+   WHERE s.board_type = 'moonboard' AND s.climb_uuid = m.alias_uuid;
+
+  -- 8. Delist the non-canonical rows. Never DELETE (see header): their
+  --    board_climb_holds/self-alias rows stay in place as harmless, unlisted
+  --    history.
+  UPDATE board_climbs
```
marcodejongh, line 553: **Old links to the losing angle-row become empty husks — confirm the app-side ordering.** `resolveCanonicalClimbUuid` is currently only called from the import scripts and kilter-sync; the web/backend climb-by-uuid paths don't consult `board_climb_aliases`. After this delist, the losing row of every merged pair still renders as a page, but its stats, ticks, favorites, and community grades have all been moved to the canonical — so bookmarks, shared links, playlist deep-links from client caches, and anything search engines indexed for the losing uuid show a grade-less, history-less climb.

Unlike 0163 (a handful of true dupes), this hits half of the affected MoonBoard catalog. If #3852 adds alias resolution to the climb view/detail path, great — but then 0185 shipping *before* it means a window where those pages are husks. Worth either landing the resolver-on-read first, or noting the accepted window explicitly in the PR.

marcodejongh (commented):
Reviewed with the branch checked out locally, including replaying the migration against a scratch Postgres 17. The overall design is genuinely solid — delist-never-delete, the durable guard row, the angle-preserving stats policy, the `user_id IS NULL` fence, and the shared CI+local replay fixture are all the right calls, and the header docs are honest about what's unverified. The dedupe grouping and canonical-selection logic all checked out against the live schema.

That said, I found one reproducible crash, one formula bug, and two sequencing/coverage gaps. Details inline; summary:

1. **Migration aborts on ≥3-member groups when a user interacted with two non-canonical members** — reproduced against a scratch Postgres: `duplicate key value violates unique constraint "playlist_climbs_playlist_id_climb_uuid_key"`. The dedupe steps only guard collisions against the canonical row, not between two alias rows repointing onto the same target. Rollback is clean (single DO block), so no corruption — but the deploy fails.
2. **`hot_score` recompute doesn't match the live trigger formula** — the migration subtracts `TIMESTAMP '2005-12-08'`; the actual trigger (`update_vote_counts()`, 0053/0130) uses plain Unix epoch. Merged climbs land ~25,200 points below every other climb in hot ordering. `rebuildGymVoteCounts()` in `packages/db/src/queries/gyms/merge-gyms.ts` already has the correct recipe to copy, including the `boardsesh.skip_vote_counts` trigger guard this migration is missing.
3. **The next MoonBoard catalog import partially undoes the merge** — today's importer matches on `${layoutId}|${angle}|${fingerprint}` and skips delisted rows, so an incoming 25° problem misses the (angle-40) canonical, re-mints `moonboard:{id}:25`, overwrites the migration's alias back to self-pointing (`catalogAliasConflictUpdate()`), and recreates a stats row under the delisted uuid. Needs the importer to go angle-agnostic in the same release, or a hard ordering guarantee with #3851.
4. **Old links to the losing angle-row become empty husks** — `resolveCanonicalClimbUuid` is only called from importers/kilter-sync today, so the web/backend climb-by-uuid paths render the delisted row with its stats/ticks/favorites moved away. This affects the losing half of every merged pair (bookmarks, shared links, anything Google indexed). Presumably #3852 covers this, but worth confirming the ordering before 0185 ships.

Smaller notes, no inline comment: the header says the formula comes from `refresh_vote_counts()` — no such function exists, it's the `update_vote_counts()` trigger; and "leaving it stale here would never self-heal" is only true until the next vote on that climb fires the trigger (which will then rewrite `hot_score` with the *other* formula — one more reason to match it exactly). Also the replay fixture's schema omits the `votes_count_trigger` entirely, so the vote steps are tested trigger-free, which is exactly where the formula bug slipped through.

None of this needs a redesign — the fixes are local: dedupe among all group members (or restrict auto-merge to pairs), copy the gym-merge vote recount recipe, and pin down the deploy ordering vs #3851/#3852.

fallaciousreasoning:
Thanks for the deep review, Marco — all four issues are fixed:

1. **Multi-member collision crash**: rewrote steps 4–6 (and added a defensive 5b for `board_climb_ratings`) to a ranked-CTE pattern that dedupes among *all* group members, not just against the canonical — `WITH ranked AS (... EXISTS(canonical_has_own) ..., ROW_NUMBER() OVER (PARTITION BY <key>, canonical_uuid[, angle]) ...) DELETE WHERE canonical_has_own OR rn > 1`, then repoint. Tables without a surrogate `id` (`board_circuits_climbs`, `board_beta_links`) use `ctid` for the same-statement tiebreak. Added your exact repro as **CASE E** (`w25`/`w40`/`w55`) in the shared fixture, covering playlist_climbs, board_circuits_climbs, climb_classic_status, board_beta_links, and votes — 6 new assertions, all passing.
2. **hot_score formula**: dropped the bogus Reddit-epoch offset, copied `rebuildGymVoteCounts()`'s recipe wholesale — plain Unix-epoch formula, `feed_items.created_at → MIN(votes) → NOW()` preference chain, and `boardsesh.skip_vote_counts` wrapped around the bulk votes DELETE/UPDATE. Also added the real `update_vote_counts()` trigger to the replay fixture (it was missing entirely, which is exactly how this slipped through) — the fixture now exercises trigger interplay directly.
3. **Import ordering**: this is what #3851 is for — the importer there goes angle-agnostic (fingerprint match drops angle from the key, `resolveCatalogClimbUuid` follows aliases, legacy per-angle aliases are still recorded). Deploy order (0185 → importer → catalog import → flag) is now called out in this migration's header and in #3852's flag comment.
4. **Husk pages**: added `resolveCanonicalClimbUuid` (backend) / an equivalent alias lookup (web SSR) to both climb-by-uuid chokepoints, with new tests covering old-alias-uuid → canonical resolution. Landing this in the same PR set so there's no window where merged-away links go husk.

Also fixed the two smaller notes: `board_climb_stats` re-key (step 2) now moves *all* of an alias's stats rows regardless of angle (dropped the `AND m.angle = s.angle` restriction on both the JOIN and the DELETE), so a stats row at a non-native angle no longer strands under the delisted uuid.

Rebased onto latest `main` to pick up an independent migration-number collision on 0185 (renumbered to 0188) — content unchanged by that, just the file/journal rename.

packages/db/drizzle/0185_moonboard_angle_dedup_backfill.sql
fallaciousreasoning: Fixed — rewrote steps 4–6 (+ a defensive 5b for `board_climb_ratings`) to a ranked-CTE pattern that dedupes among ALL group members, not just against the canonical. Added your exact repro as CASE E (`w25`/`w40`/`w55`) in the shared replay fixture, covering playlist_climbs, board_circuits_climbs, climb_classic_status, board_beta_links, and votes — all passing.

Since then I also replayed the whole migration against the real pre-built dev-db image (766 real duplicate groups, merged cleanly, zero errors) and separately seeded a live multi-member collision on it — same user voting on both members of a real 2-climb group. No crash; exactly one vote survives on the canonical, the loser's row and vote_counts are gone.

packages/db/drizzle/0185_moonboard_angle_dedup_backfill.sql
fallaciousreasoning: Fixed — dropped the bogus Reddit-epoch offset and copied `rebuildGymVoteCounts()`'s recipe wholesale: plain Unix-epoch formula, `feed_items.created_at → MIN(votes) → NOW()` preference chain, `boardsesh.skip_vote_counts` guard around the bulk repoint. Also added the real `update_vote_counts()` trigger to the replay fixture so it's actually exercised now.

Just verified this against a live seeded vote on the dev DB: the post-migration `hot_score` matched the trigger's pre-migration value to the last decimal (39670.82408308164) — not just formula-equivalent by inspection, actually bit-identical on a real row.

packages/db/drizzle/0188_moonboard_angle_dedup_backfill.sql
fallaciousreasoning: Fixed — dropped the `AND m.angle = s.angle` restriction from both the step-2 JOIN and DELETE, so all of an alias's stats rows re-key onto the canonical regardless of angle, not just the row at the alias's own native `board_climbs.angle`.

packages/db/drizzle/0188_moonboard_angle_dedup_backfill.sql
fallaciousreasoning: This is what #3851 solves — angle-agnostic fingerprint match, `resolveCatalogClimbUuid` follows aliases, and an explicit `ambiguous` skip (loud + counted, not a silent guess) when a pre-migration database still has two listed rows sharing holds. Deploy order (this migration → #3851's importer → any catalog import) is now stated explicitly in this migration's header.

packages/db/drizzle/0188_moonboard_angle_dedup_backfill.sql
fallaciousreasoning: Fixed — added `resolveCanonicalClimbUuid` (backend) and an equivalent alias lookup (web SSR) to both climb-by-uuid chokepoints, with new tests covering old-alias-uuid → canonical resolution. Landing in the same PR set so there's no window where a merged-away link renders as a husk.

marcodejongh:
Taking over this stale/conflicted migration as part of #3529. The current branch will not be merged as written: it independently merges total/upstream/Boardsesh counts, drops current quality provenance, misses current reference/offline-tombstone paths, and has unresolved destructive collision handling. I’ll rebuild from latest `main` after the prevention/importer stages and a read-only production collision inventory. No data deletion or production mutation will occur without explicit approval.

marcodejongh:
Orchestrated rework + verification record (Fable-led, Opus agents), 2026-08-12:

**Takeover concerns (2026-08-01) → all addressed:** count-invariant preservation (single-owner contract: step 3b recounts the Boardsesh half from repointed ticks for every key step 2 writes), quality provenance (`boardsesh_quality_sum/count` carried + blend recomputed via the `blendedQualityAverageSql` recipe), offline tombstone paths (hand-written tombstones for the moved `user_favorites`/`playlist_climbs` client keys — the repoint-not-delete hole; trigger tombstones verified for stats deletes), collision handling (per-table survivor policies, above in the PR body, awaiting sign-off).

**Verification:** 35-agent adversarial review (5 dimensions → 15 confirmed findings fixed, 15 refuted); two independent dev-db replays on real pre-dedup data (766 merged / 478 same-angle skipped / 0 errors), the second on a pristine container with idempotency + deterministic crash-kill rollback verified; 46-check replay fixture with a 12-mutation kill matrix; full backend suite 3341/3341.

**Two deploy-aborting bugs found and fixed beyond the review scope:** raw alias rows feeding `ON CONFLICT DO UPDATE` twice ("cannot affect row a second time") in `board_climb_stats` and `board_climb_send_stats`.

**Hard constraints before merge (also in the PR body):** must land AFTER #4054's `0193`/`0194` and be renumbered above them (verified corruption paths in both directions otherwise — details in the migration header's DEPLOY ORDERING block); prod sizing query still to run; survivor policies + the prod mutation itself need explicit maintainer sign-off. Offline-convergence residuals tracked in #4389.

marcodejongh:
Merged — thanks @fallaciousreasoning for the original design and implementation, and for the thorough replay-fixture discipline that the rework kept building on. The importer half (#3851) landed yesterday; with this migration prod MoonBoard climbs become angle-agnostic on the next deploy. Same-angle residual groups tracked in the follow-up issue; offline-convergence residuals in #4389.

PR #3851, opened 2026-07-23
feat(db): moonboard catalog importer emits one angle-agnostic climb per problem

fallaciousreasoning:
## Summary

Second of three PRs widening MoonBoard's angle support (first: #3849, the dedup migration). This rewrites the catalog importer so future re-imports from Moon Climbing stop minting a new climb UUID per graded angle (`moonboard:{id}:{angle}`) and instead write one climb row per problem — angle-agnostic, matching Kilter/Tension — with a `board_climb_stats` row per graded angle under that same UUID.

- `catalogClimbUuid({ id })` drops `angle` from the identity key. `catalogProblemToClimbs` now returns a single `MappedCatalogClimb` (structural fields computed once: holds, name, setter, fingerprint) carrying a `stats: MappedCatalogClimbStats[]` array (one entry per graded angle) instead of one `MappedCatalogClimb` per angle.
- Per-angle grade/rating/repeats data is fully preserved — nothing is lost, it's just relocated from the climb identity to the stats table, where it already structurally belonged (`board_climb_stats` was already keyed on `(board_type, climb_uuid, angle)`).
- The cross-problem dedup collision logic (`isBetterCatalogClimb`, `buildExistingCatalogMatchIndex`/`resolveCatalogClimbUuid`) now operates on whole problems instead of individual angle-configs, matching on `(layoutId, holdFingerprint)` instead of `(layoutId, angle, holdFingerprint)`.
- Every graded angle still gets a legacy `moonboard:{id}:{angle}` alias (`legacyCatalogClimbUuid`) pointing at the new canonical UUID, so the personal CSV logbook importer (`packages/backend/src/services/moonboard-import.ts`, out of scope for this rewrite) keeps resolving ticks unmodified.

## Release Notes

## Test plan

- `packages/db/scripts/moonboard-catalog-helpers.test.ts` rewritten for the new one-climb-plus-N-stats shape: 23/23 passing.
- `packages/backend/src/__tests__/moonboard-import-service.test.ts` (the personal-importer regression suite, which depends on the legacy alias backward-compat requirement) — unmodified, still 4/4 passing.
- `vp run typecheck:db`: clean.
- `vp check`: clean.

### DIFF
--- a/packages/db/scripts/import-moonboard-catalog.ts
+++ b/packages/db/scripts/import-moonboard-catalog.ts
@@ -26,14 +26,15 @@ const __dirname = path.dirname(fileURLToPath(import.meta.url));
 // MoonBoard catalog import (all 7 boards)
 // =============================================================================
 // Imports the full MoonBoard catalog dataset. One file per board, each
-// { count, holdsetup, problems[] }. We write one climb row per (problem, graded
-// angle) across all 7 boards and both 25°/40° angles.
+// { count, holdsetup, problems[] }. We write ONE climb row per problem —
+// angle-agnostic, matching Kilter/Tension — and one board_climb_stats row per
+// graded angle (typically 25° and 40°) under that same climb UUID.
 //
 // MERGE IN PLACE (non-destructive): the dataset re-keys identities (stable
 // problem id) vs the rows already in prod (keyed on apiId / name+setter). To
 // avoid duplicating ~163k existing MoonBoard climbs — and to keep their UUIDs,
 // URLs, ticks and favourites intact — we match each incoming climb to an
-// existing one by (layout_id, angle, hold_fingerprint), tie-breaking on
+// existing one by (layout_id, hold_fingerprint), tie-breaking on
 // case-insensitive name. A match reuses the existing UUID (so the upsert updates
 // it in place, backfilling the 2024 quality/ascensionist gap); a miss mints a
 // stable id-based UUID and inserts. Stat upserts are monotonic — they never
@@ -53,7 +54,7 @@ const BATCH_SIZE = 2000;
 
 /**
  * Build the in-memory match index for the non-destructive merge:
- * `${layoutId}|${angle}|${fingerprint}` → existing climbs with those holds.
+ * `${layoutId}|${fingerprint}` → existing climbs with those holds.
  * Existing MoonBoard climbs predate the fingerprint column (only layout 3 has
  * it populated in prod), so we recompute every fingerprint from board_climb_holds.
  * Holds are streamed with a cursor and folded per-climb so memory stays bounded.
@@ -91,7 +92,6 @@ async function buildExistingIndex(
     .select({
       uuid: boardClimbs.uuid,
       layoutId: boardClimbs.layoutId,
-      angle: boardClimbs.angle,
       name: boardClimbs.name,
       isListed: boardClimbs.isListed,
     })
@@ -161,10 +161,11 @@ async function importMoonBoardCatalog() {
 
       // Dedupe in memory keyed per conflict target so a single batch never
       // proposes the same target twice ("ON CONFLICT cannot affect row a second
-      // time"). When two problems share the same holds+angle (so they resolve to
-      // one target UUID) we keep the STRONGER problem (isBetterCatalogClimb: more
-      // repeats, then benchmark) instead of last-wins — otherwise a junk duplicate
-      // can clobber a real benchmark's ascent count and benchmark flag.
+      // time"). When two problems share the same holds (so they resolve to one
+      // target UUID) we keep the STRONGER problem (isBetterCatalogClimb: more
+      // total repeats across its graded angles, then benchmark) instead of
+      // last-wins — otherwise a junk duplicate can clobber a real benchmark's
+      // ascent count and benchmark flag.
       const bestByUuid = new Map<string, MappedCatalogClimb>();
       const climbByUuid = new Map<string, typeof boardClimbs.$inferInsert>();
       const statsByUuid = new Map<string, typeof boardClimbStats.$inferInsert>();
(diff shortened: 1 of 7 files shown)


fallaciousreasoning:
Reopening to pick up the corrected conventional-commit PR title for CI.

packages/db/scripts/moonboard-catalog-helpers.ts
```diff
@@ -150,7 +157,7 @@ export function resolveCatalogClimbUuid(
   mapped: MappedCatalogClimb,
   index: Map<string, ExistingCatalogClimb[]>,
 ): { uuid: string; matched: boolean } {
-  const candidates = index.get(`${mapped.layoutId}|${mapped.angle}|${mapped.holdFingerprint}`);
+  const candidates = index.get(`${mapped.layoutId}|${mapped.holdFingerprint}`);
   if (!candidates || candidates.length === 0) return { uuid: mapped.uuid, matched: false };
```
marcodejongh: **Pre-dedup databases hit this path for every duplicated problem, and the outcome is messy — worth a guard.** Until migration 0185 (#3849) has run, both per-angle rows of a problem are listed and share `(layoutId, fingerprint)`, so `candidateUuids.size === 2` and this name tie-break decides. Both rows carry the same name, so `.find()` returns whichever sits first in the bucket — effectively DB row order. The importer then:

- writes **both** angles' stats onto that one row, while the twin keeps its own stats → ascents double-counted across two listed rows;
- writes a legacy alias whose `alias_uuid` is the **other, still-listed row's uuid**, pointing at the chosen row — so `resolveCanonicalClimbUuid` starts redirecting that row's ticks while the row itself still shows in search.

0185 running afterwards mostly repairs this (its alias upsert and stats conflict-merge paper over it), but the intermediate state is wrong and the repair path was never designed for it.

Since catalog imports are manual, a cheap guard would close the whole class: when a fingerprint bucket contains ≥2 distinct **listed** candidates with the same name, abort (or at minimum count + `console.error` and skip the problem) with a message telling the operator to run migrations first. That turns "ran the import one deploy too early" from silent data corruption into a loud no-op.

packages/db/scripts/import-moonboard-catalog.ts
```diff
@@ -175,100 +176,112 @@ async function importMoonBoardCatalog() {
...
+          setterId: null,
+          setterUsername: mapped.setterUsername,
+          name: mapped.name,
+          description: mapped.description,
+          hsm: null,
+          edgeLeft: null,
+          edgeRight: null,
+          edgeBottom: null,
+          edgeTop: null,
+          angle: null,
```
marcodejongh, line 307: **`angle: null` drops grade enrichment everywhere that joins stats at `boardClimbs.angle`.** Three backend paths do `LEFT JOIN board_climb_stats ON … stats.angle = climbs.angle` to resolve `difficultyName`: `newClimbFeed` (`new-climb-subscriptions.ts`), `notification-worker.ts`, and `feed-fanout.ts`. With `angle` null the join never matches, so every newly imported MoonBoard problem shows ungraded in the new-climb feed and new-climb notifications — a visible regression vs today's per-angle rows, which do resolve a grade there.

Two follow-ons from the same choice:

- The board_climbs conflict clause only updates `characteristics`/`description`, so **matched existing rows keep their old non-null angle** — the catalog becomes a permanent mix of angle-null (new) and angle-25/40 (pre-existing) rows. If that's intended, fine, but it means the enrichment gap above only affects *new* problems, which will make it look like a flaky bug rather than a model change.
- Whichever way you go, it'd be worth teaching those three queries a MoonBoard fallback (e.g. pick the min-angle stats row when `climbs.angle` is null) — that's also the long-term correct behavior once climb identity is angle-agnostic everywhere.

Fine to defer the query fix to a follow-up, but it should be tracked — right now nothing fails, the grades just quietly vanish.

marcodejongh (commented):
Reviewed with the branch checked out locally, alongside #3849 (reviewed earlier today) and #3852. This importer rewrite is the missing half of the sequencing concern I raised on #3849 — the angle-agnostic fingerprint match, terminal-canonical alias resolution, and the legacy per-angle aliases are exactly what stops a re-import from resurrecting the merged-away rows. The `resolveIncumbentReplacement` extraction also fixes a genuinely subtle bug (a winner graded at fewer angles than the loser would previously leave the loser's stale per-angle stats staged under the winning uuid), and the test that pins it is great.

The helper logic all checked out. Three things worth addressing, all about how this code behaves against data it will actually meet in prod:

1. **Running this importer against a pre-#3849 database makes a mess, and nothing guards against it** (inline comment on `resolveCatalogClimbUuid`).
2. **`angle: null` silently drops grade enrichment in the new-climb feed and notifications** (inline comment on the insert payload).
3. **The match index has no `user_id IS NULL` fence, and dropping angle from the key widens the exposure** (inline comment on `buildExistingIndex`).

Across the trio, the safe order seems to be: deploy #3849's migration, then this importer, and only then run a catalog import — with #3852's flag flipped last. Worth stating that explicitly in one place (this PR's header comment in the import script is probably the spot operators will actually read).

packages/db/scripts/import-moonboard-catalog.ts
```diff
@@ -91,7 +92,6 @@ async function buildExistingIndex(
     .select({
       uuid: boardClimbs.uuid,
       layoutId: boardClimbs.layoutId,
```
marcodejongh, line 106: **No `user_id IS NULL` fence on the match index — and dropping angle from the key widens the exposure.** This selects every MoonBoard `board_climbs` row, including user-created climbs. A user climb sharing holds with an incoming catalog problem (now at *any* angle, not just the same one) can be adopted as the merge target, after which the catalog upserts its stats onto the user's climb and points the problem-id aliases at it.

This is pre-existing behavior, not introduced here — but #3849's migration went out of its way to fence user climbs (`user_id IS NULL` throughout), and the same one-line filter here (`isNull(boardClimbs.userId)` on `climbRows`, plus the fingerprint cursor if you want to save memory) would make the importer consistent with that policy.

fallaciousreasoning:
Thanks Marco — fixed all three:

1. **Pre-dedup ambiguity guard**: `resolveCatalogClimbUuid` now returns `{ uuid, matched, ambiguous }`, with `ambiguous: true` whenever the fingerprint bucket has more than one distinct listed-row candidate (name-matched or not) — the exact pre-0185 shape you described. The importer checks this before recording any alias/stats write and skips the problem with a counted `skippedAmbiguous` + `console.error`, so an accidental early import is a loud no-op instead of silent double-counting/alias-flipping. Two new tests pin the ambiguous-with-name-match and ambiguous-without-name-match cases.
2. **`user_id IS NULL` fence**: `buildExistingIndex`'s `climbRows` query now filters `and(eq(boardClimbs.boardType, 'moonboard'), isNull(boardClimbs.userId))`, matching #3849's policy.
3. **`angle: null` grade-enrichment gap**: left the query fix as a follow-up per your own suggestion, but it's no longer untracked — added a comment on the `angle: null` field naming the three affected backend paths (`new-climb-subscriptions.ts`, `notification-worker.ts`, `feed-fanout.ts`) so it doesn't read as a flaky bug later.

Also stated the deploy order explicitly (0185 → this importer → catalog import → #3852's flag) in the migration's header comment on #3849, as you suggested putting it where operators will actually read it.

packages/db/scripts/moonboard-catalog-helpers.ts
fallaciousreasoning: Fixed — `resolveCatalogClimbUuid` now returns `{ uuid, matched, ambiguous }`; `ambiguous: true` whenever the fingerprint bucket has more than one distinct listed candidate (name-matched or not — the exact pre-0185 shape you described). The importer checks this before any alias/stats write and skips with a counted, loud `console.error`. Two new tests pin both ambiguous shapes (with and without a name match).

packages/db/scripts/import-moonboard-catalog.ts
fallaciousreasoning: Left as a follow-up per your suggestion, but flagged more prominently in a comment on this field.

Correction to my earlier reply though: I said this affects three backend paths (`new-climb-subscriptions.ts`, `notification-worker.ts`, `feed-fanout.ts`). Grepping more thoroughly just now, the same `boardClimbStats.angle = boardClimbs.angle` join pattern also shows up in:
- `events/index.ts`'s `climb.created` pubsub handler — has a fallback to `event.metadata.difficultyName`, so likely *not* actually broken for catalog-imported climbs (the importer never fires this event), but worth checking for MoonBoard climbs created via the app's own climb-builder UI.
- `climb-similarity.ts`'s `findExactDuplicateMatch` and #3852's `moonboard-duplicates.ts` — same join, but only feeds the ascensionist-count tie-break used to pick which duplicate to cite, not detection itself (that's keyed on the hold signature). Degrades to 0/uuid-order ranking for angle-null climbs, not a correctness bug.

None of these are new bugs from this PR and none are correctness-breaking — but the blast radius of the known gap is wider than I told you, and I'd rather correct that than let the undercount stand.

packages/db/scripts/import-moonboard-catalog.ts
fallaciousreasoning: Fixed — `buildExistingIndex`'s `climbRows` query now filters `and(eq(boardClimbs.boardType, 'moonboard'), isNull(boardClimbs.userId))`, matching #3849's policy.

marcodejongh:
Taking over this stale/conflicted importer follow-up as part of #3529. I’ll recreate the valid angle-agnostic importer work on latest `main`, add a hard guard so it cannot write before the replacement migration is installed, and cover structural-angle consumers before it can run. No import or production mutation will be executed from this branch.

marcodejongh:
Fable review pass on this PR turned up one major and three minor items. All four are addressed in `fix(db): guard the catalog importer against fingerprint drift`, and the branch is rebased on latest `main`.

**Major — the deferred angle-null gap had no tracking issue.** @marcodejongh approved deferring the `difficultyName` join fix on the condition it got tracked, and only a code comment existed. Filed **#4220**, covering the three surfaces where the grade actually disappears (`new-climb-subscriptions.ts`, `notification-worker.ts`, `feed-fanout.ts`) plus the wider blast radius @fallaciousreasoning corrected himself on in the thread (`events/index.ts`'s `climb.created` handler for app-built MoonBoard climbs, and the ascensionist tie-break degradation in `climb-similarity.ts` / `moonboard-duplicates.ts`). The `KNOWN GAP` comment now points at #4220.

**Minor — fingerprint drift could mint a duplicate and hijack the legacy aliases.** If a re-scraped problem's holds no longer match the rows it already owns (an upstream hold edit, or a future change to `parseMovesString`/`fingerprintFromHolds`), `resolveCatalogClimbUuid` returns `matched: false` and the importer inserted a fresh `moonboard:{id}` row. The old per-angle scheme collided with the existing row and merged in place; the new scheme does not. Worse, `catalogAliasRows` then writes `moonboard:{id}:{angle}` aliases whose `alias_uuid` **is** one of those still-listed rows, and `catalogAliasConflictUpdate` sets `canonical_uuid = excluded.canonical_uuid` unconditionally — so their self-aliases get repointed at the empty duplicate and their ticks start redirecting there while both rows stay listed.

New `existingClimbUuidsForProblem` helper closes it in the spirit of the ambiguous skip: `buildExistingIndex` now also returns the set of existing MoonBoard climb UUIDs, and on a match miss the importer checks whether the problem's id-based or legacy per-angle UUIDs are already climb rows. If any are, it skips loudly (counted, `console.error`, listed in the totals) instead of inserting. Three tests pin the clean case, the legacy-per-angle drift case, and the id-based case.

**Minor — the ambiguous-skip message misattributed post-0185 duplicates.** Migration 0185 only collapses a same-signature group whose members all sit at distinct angles, so it deliberately leaves cross-problem duplicate classes alone (the real "birthday cake trail mix" vs the junk problem named "name" — four listed rows on one signature). Those still trip the ambiguous skip on a fully migrated database, and both the per-problem error and the totals summary told the operator to go run a migration that had already run. Both messages now name the two causes and say which one running migrations fixes.

**Minor — `statsByUuid` is keyed `${uuid}:${angle}`.** Renamed to `statsByUuidAngle`; the compound key is load-bearing for the stale-key clearing, so the name should say so.

Validation: `vp run typecheck:db` clean, `vp lint` clean on all three touched files, `moonboard-catalog-helpers.test.ts` 31/31 (was 28). The three MoonBoard zone-filter failures in the wider `packages/db` suite are pre-existing and tracked in #4049 — unrelated to this diff.

packages/db/scripts/import-moonboard-catalog.ts
marcodejongh: Tracked in #4220 as you asked, and the `KNOWN GAP` comment on this field now points at it. The issue covers the three surfaces where the grade actually disappears (new-climb-subscriptions.ts, notification-worker.ts, feed-fanout.ts) plus the wider blast radius from your correction — the `climb.created` handler in events/index.ts for app-built MoonBoard climbs, and the ascensionist tie-break degradation in climb-similarity.ts / moonboard-duplicates.ts — along with the mixed angle-null / angle-25-40 catalog state that makes it look flaky. Resolving this thread; the fix itself lives in #4220.

marcodejongh:
Orchestrated review + QA pass on the rebased branch (Fable-led, Opus agents: 4-dimension review fan-out with adversarial verification — 12 findings confirmed, 8 refuted — plus an end-to-end QA run of the real importer against the dev-db image, 142k MoonBoard climbs).

**Fixed on the branch (all mutation-checked):**
- Duplicate gate (`moonboard-duplicates.ts`) now matches angle-null catalog rows at every angle — previously the whole imported catalog would bypass `saveMoonBoardClimb`'s duplicate check (`angle = $angle` excluded NULL rows).
- In-batch same-holds problems now collapse onto one climb (`foldedInBatch` counter) instead of double-inserting and then tripping the ambiguity guard forever on later runs.
- The drift guard now also protects the matched branch (`skippedHijacked`): a matched merge can no longer silently repoint a different live climb row's alias (tick/logbook hijack).
- Drift/hijack guards consider legacy angles (`MOONBOARD_ANGLES`), not just currently-graded ones.
- Helper↔importer batch-key formats unified and pinned by test; per-problem staging extracted into `moonboard-catalog-batch.ts` (pure, DB-free, verbatim parity-checked).
- Operator readout uses `describeDatabaseHost()`; documented run command corrected to the working `vp run '@boardsesh/db#db:import-moonboard-catalog'` form.

**QA evidence (dev DB):** one angle-null climb + 2 stats rows + 3 aliases for a new 2-angle problem; ambiguity guard fired loudly on a real pre-dedup duplicate pair with zero writes; second run idempotent on values; monotonic upsert held counts/quality on a lowered snapshot and added a new angle's stats row.

**Deferred (tracked in #4356):** pre-existing sync-cursor churn on no-op re-imports; match-index canonical-holds verification; `user_id` fence test.

Merge-order note: this PR's ambiguity guard is what makes a catalog import safe on a pre-dedup database, so it lands BEFORE the #3849 migration.

PR #3852, opened 2026-07-23
feat(web): widen MoonBoard's angle range behind a feature flag

fallaciousreasoning:
## Summary

Third of three PRs widening MoonBoard's angle support (#3849 dedup migration, #3851 importer rewrite). This is the user-visible change: MoonBoard angle pickers can now offer the full 0–70° step-5 range used by Kilter/Tension, gated behind a new `moonboard-wide-angles` PostHog flag.

- OFF (the default until the flag is created in PostHog) keeps every picker limited to the existing `MOONBOARD_ANGLES` (25°/40°) list — that constant is completely unchanged, so nothing regresses if the flag never resolves.
- ON swaps in the new `MOONBOARD_WIDE_ANGLES` constant (same 15-value array Kilter/Tension use) via one shared hook, `useBoardAngleOptions`, used by every angle picker: the angle drawer, board selector, playlist generator, log-ascent form, propose-change form, and create-climb form.
- Matches the repo's established "flags gate the UI entry point only" convention (see `docs/kilter-sync.md`'s `kilter-oauth-linking` precedent) — nothing server-side validates MoonBoard angle against this list today (it's a plain 0–90 bounded int everywhere it's checked), so no backend/API changes were needed to accept wider values once the flag is on.
- Also fixes a latent bug in MoonBoard's create-climb duplicate-detection: it was comparing `angle`, so saving the same holds at two different angles was never flagged as a duplicate. Now angle-agnostic, mirroring `findExactDuplicateMatch` (the existing Kilter/Tension pattern in `climb-similarity.ts`) — a MoonBoard problem is the same physical route regardless of angle.

**Mobile is intentionally out of scope**: no MoonBoard create-climb flow exists there, and its viewing/browsing angle pickers stay on the narrow list unconditionally (no flag wired up on mobile yet).

**Rollout note**: the `moonboard-wide-angles` PostHog flag still needs to be created in the project before this does anything — same one-time step every other flag here needed. Recommend not flipping it on until #3849 has run in prod (otherwise a user could grade a problem at a new angle while it's still split across duplicate per-angle rows).

## Release Notes

- Set up any MoonBoard at more than the usual 25° and 40° — grade and log it there like any other angle. *(Rolling out gradually — ask if you don't see it yet.)*

## Test plan

- New `packages/web/app/hooks/use-board-angles.ts` + `use-board-angles.test.tsx`: 5/5 passing (flag off/undefined → narrow list, flag on → wide list for MoonBoard only, Kilter/Tension never affected either way).
- Existing component tests for every touched picker still pass: `angle-selector.test.tsx`, `generator-options-form.test.tsx`, `create-climb-form.test.tsx` (65/65), plus the logascent-form regression suite (15/15).
- `packages/backend/src/__tests__/moonboard-duplicates.test.ts` updated for the dropped `angle` parameter: 3/3 passing.
- `vp run typecheck:web`, `vp run typecheck:backend`: clean.
- `vp check`: clean, including `check:i18n` (no new hardcoded strings).


packages/shared/board-config/src/moonboard-config.ts
```diff
@@ -116,9 +116,21 @@ const MOONBOARD_ROWS = [1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 1
...
+// see the `moonboard-wide-angles` feature flag for the flag-gated wider range.
 export const MOONBOARD_ANGLES = [25, 40] as const;
 
+// The full angle range, matching Kilter/Tension (ANGLES.kilter/tension in
+// board-data.ts), offered by angle pickers when the `moonboard-wide-angles`
+// feature flag is on. A MoonBoard problem's holds don't change with angle —
+// Moon Climbing just never grades outside 25°/40°, so climbs graded at other
+// angles start with no catalog grade/stats until the community logs there,
+// exactly like any Aurora board today.
+export const MOONBOARD_WIDE_ANGLES = [0, 5, 10, 15, 20, 25, 30, 35, 40, 45, 50, 55, 60, 65, 70] as const;
```
marcodejongh, line 143: **The pre-#3849 hazard here is concrete, not hypothetical — consider making the ordering constraint operational.** If this flag turns on before migration 0185 has run, a user logging a tick at, say, 55° writes it against one of the *per-angle* climb uuids (`moonboard:{id}:25` / `:40`), and the tick recompute then mints a `board_climb_stats` row at 55° under that uuid. 0185's step 2 only migrates the alias's stats row at its own `board_climbs.angle`, so those wide-angle stats rows get stranded under the delisted uuid — orphaned counts that never reach the canonical (flagged in detail on my #3849 review).

The PR description says "recommend not flipping it on until #3849 has run in prod", which is right — but PostHog flags outlive PR descriptions. Suggest putting the constraint where the person flipping the flag will see it: the flag's description field in PostHog when it's created, plus a line in this constant's comment ("do not enable before migration 0185 has run in prod"). Cheap insurance against a six-months-later toggle.

packages/backend/src/graphql/resolvers/climbs/moonboard-duplicates.ts
```diff
@@ -174,7 +173,9 @@ export async function findMoonBoardDuplicateMatches(
        AND ${dbSchema.boardClimbStats.angle} = ${dbSchema.boardClimbs.angle}
       WHERE ${dbSchema.boardClimbs.boardType} = 'moonboard'
         AND ${dbSchema.boardClimbs.layoutId} = ${layoutId}
-        AND ${dbSchema.boardClimbs.angle} = ${angle}
+        -- Angle is deliberately NOT a predicate — a MoonBoard problem is the
```
marcodejongh, line 176: Agreed with this change — it mirrors `findExactDuplicateMatch` and is strictly more protective (it now also catches re-creating an existing problem at the other angle, which used to sail through). Two small observations, neither blocking:

- Until 0185 runs, a candidate's holds can match **both** per-angle catalog rows; the error surfaces whichever name comes back first. Same name in practice, so harmless — just noting the transitional behavior is understood.
- On browsing UX once the flag is on: at a fresh wide angle (e.g. 55°) no MoonBoard climb has a stats row yet, so only the page-0 stats-less fallback in `searchClimbs` surfaces anything — stats-driven sorts, stats filters, and deeper pages come back empty until people start logging there. The intended flow (find the climb at 25°/40°, then log at 55° via the wider picker, which mints the stats row) works, and it matches how an ungraded-angle Kilter board behaves — but since the release note promises "grade and log it there like any other angle", worth a conscious sign-off that thin browsing at new angles is accepted for the initial rollout.

marcodejongh (commented):
Reviewed with the branch checked out locally, together with #3849 and #3851. This is a textbook flag rollout: the hook is small and well-tested with a stable array reference, Kilter/Tension are provably untouched, the `angle` input goes optional instead of a breaking removal, and I verified the two claims that matter — nothing server-side validates MoonBoard angle against the 25/40 list, and no route-level check rejects a wide angle in the URL. The angle-agnostic duplicate check matches the Kilter/Tension precedent and is the right call under the new one-climb-per-problem model.

Two comments inline, both about rollout sequencing rather than the code itself:

1. The "don't flip the flag before #3849 has run" note deserves to be stronger than a PR-description aside — a wide-angle tick pre-dedup creates exactly the stats rows that migration 0185 currently strands (detailed on the inline comment, cross-linked to my #3849 review).
2. Browsing UX at a fresh wide angle is thinner than the release note implies — worth a conscious sign-off (inline).

Also noting mobile stays on the narrow list unconditionally, which the PR calls out — reasonable since there's no flag plumbing there yet, and a tick logged at a wide angle on web still displays fine on mobile (angle is a plain number end to end).

fallaciousreasoning:
Thanks Marco — addressed both:

1. **Pre-#3849 hazard**: strengthened the comment on `MOONBOARD_WIDE_ANGLES` with an explicit `⚠️ DO NOT ENABLE THE 'moonboard-wide-angles' FLAG IN POSTHOG BEFORE migration 0185/0188 has run in prod` warning, spelling out the concrete failure (wide-angle ticks write against per-angle uuids the migration's stats re-key won't reach). Will also mirror this into the flag's description field when creating it in PostHog, per your suggestion.
2. **Thin browsing UX at fresh wide angles**: acknowledged as an accepted initial-rollout tradeoff, consistent with how an ungraded-angle Kilter board already behaves. Not blocking — flagging back for visibility since you asked for a conscious sign-off rather than a code change.

The transitional both-per-angle-rows-match observation on the duplicate-check comment is understood and expected to resolve itself once #3849 has run — no action needed there.

packages/shared/board-config/src/moonboard-config.ts
fallaciousreasoning: Fixed — strengthened the comment on `MOONBOARD_WIDE_ANGLES` with an explicit warning not to enable the flag before migration 0185/0188 has run in prod, spelling out the concrete failure mode (wide-angle ticks would write against per-angle uuids the migration's stats re-key won't reach). Will mirror this into the flag's description field when creating it in PostHog too.

packages/backend/src/graphql/resolvers/climbs/moonboard-duplicates.ts
fallaciousreasoning: Acknowledged as an accepted initial-rollout tradeoff, consistent with how an ungraded-angle Kilter board already behaves. Flagging back for visibility per your ask for a conscious sign-off rather than treating it as something to fix in this PR.

marcodejongh:
Closing this obsolete Next.js climbing-surface implementation. Boardsesh’s React Native app now ships the supported web climbing experience and is the only climbing surface going forward, so the legacy Next.js angle-picker work should not be carried forward. Any renderer-independent/backend requirements from #3529 will be rebuilt in the current-main prevention/importer/migration sequence.

Issue #3539, opened 2026-07-07
No cross-instance mutual exclusion for sync daemons: overlapping instances double-run user syncs and duplicate setter notifications


marcodejongh:
## Summary
Credential selection is a plain SELECT ... ORDER BY ... LIMIT 1 with no claim (no FOR UPDATE SKIP LOCKED, no advisory lock, no singleton lease), and the shared-sync/catalog cooldowns are in-memory Maps (aurora sync-runner.ts:45/320-337, kilter sync-runner.ts:38/251-259). During a rolling deploy (old + new container both running, which Railway does by default) or any accidental second replica, both daemons pick the same credential and run the same user sync and the same shared/catalog sync concurrently. Most upserts are idempotent, but createSetterSyncNotifications is a plain INSERT with fresh uuids (shared-sync.ts:988-1003, kilter notifications.ts) — followers get duplicate 'new climbs' notifications — and concurrent multi-row upserts over the same key ranges can deadlock, surfacing as spurious sync failures that feed the starvation/silent-failure findings. Restarts also reset the cooldown, so every deploy immediately fires a full shared sync per board.

Partial overlap with #3332 (persist the shared-sync cooldown in Postgres) and #1976 (cooldown stamps on failure); neither covers the cross-instance lock or the duplicate notification insert.

## Failure scenario
Deploy overlaps for 60s during an active cycle -> both instances select the same oldest credential -> two identical shared syncs run against Aurora with the same cursors -> both compute the same allNewClimbs and both insert notification rows -> every follower of every setter with new climbs that hour gets doubled notifications; concurrent climb_holds/climbs upserts deadlock and one cycle fails (silently, per the other findings).

## Evidence
`packages/aurora-sync/src/runner/sync-runner.ts:45`

```
aurora sync-runner.ts:240-250 (no locking in candidate select), 45 + 318-356 (in-memory lastSharedSyncAt); kilter sync-runner.ts:403-437, 38 + 248-259; shared-sync.ts:997-1003 (notifications insert, non-idempotent).
```

## Suggested fix
Take a Postgres advisory lock per daemon type at startup (pg_try_advisory_lock) so a second instance idles, or claim credentials with UPDATE ... WHERE ... RETURNING / FOR UPDATE SKIP LOCKED; move the shared-sync cooldown into board_shared_syncs (a row per board with a last-run timestamp, compare-and-set).

[...]

Issue #3520, opened 2026-07-07
Bids sync crashes on missing created_at, aborting the user's whole sync batch


marcodejongh:
## Summary
The bids path uses `createdAt: new Date(item.created_at).toISOString()` with no null guard (the ascents path guards at line 180: `item.created_at ? ... : now`). A bid row without created_at makes toISOString() throw RangeError on Invalid Date, the per-batch transaction rolls back, and syncUserData throws — the user's remaining tables never sync that run and the checkpoint never advances (retry loops on the same poison row).

## Failure scenario
Aurora returns a bids row with created_at null/empty (older exports and API rows have inconsistent created_at). Every sync attempt for that user fails at the same batch; their logbook silently stops updating (stale data, not just an error blip).

## Evidence
`packages/aurora-sync/src/sync/user-sync.ts:232`

```
Code asymmetry between line 180 (guarded) and line 232 (unguarded); same asymmetry in packages/web/app/lib/data-sync/aurora/user-sync.ts lines 173 vs 219. The per-user fault isolation in sync-runner means the failure repeats indefinitely for that user.
```

## Suggested fix
Mirror the ascents guard: `item.created_at ? new Date(item.created_at).toISOString() : now` (after fixing the timezone parsing per finding 1).

Found by the 2026-07 upstream-sync data audit (verified against prod). Related program PRs: ascent-count fix, star-scale repair (in flight).

Issue #1976, opened 2026-05-05
shared-sync cooldown stamps on failure, locking out retry for up to 1h


marcodejongh:
## Current behavior

`packages/aurora-sync/src/runner/sync-runner.ts` (lines 320, 326, 328) stamps the per-board cooldown both before the run and again on success/failure. A transient Aurora 5xx or network blip therefore disables the per-board shared sync for the full `sharedSyncCooldownMs` window (default 1h).

The behavior is deliberate — see `packages/aurora-sync/src/runner/sync-runner.test.ts:176-186` ("still respects the cooldown when the previous run failed"). Without it, a permanently-failing board would loop every cycle.

## Why it's worth revisiting

A flapping Aurora API can leave shared data hours stale, even when the next attempt would have succeeded. Surfaced in code review on #chore/aurora-sync-shared-sync-piggyback as "not a blocker, but worth a follow-up issue."

## Proposed directions

Either:

1. **Backoff on failure.** Shorter cooldown on failure (e.g. 5min) with exponential backoff capped at `sharedSyncCooldownMs`. Successful runs reset the backoff.
2. **Classify failures.** Distinguish transient (5xx, network, timeout) from permanent (auth, 4xx). Stamp the full hour only on permanent failures; use a much shorter cooldown (~5min) for transient ones.

Option 2 is cleaner if Aurora's error responses are reliable enough to classify. Option 1 is safer if not.

Either change needs the `sync-runner.test.ts:176-186` test updated to match the new contract.

marcodejongh:
🤖 Claimed by bug-batch loop `codex-moarbugs` (batch 1), replacing a released monitor-only item. Current main still stamps the full shared-sync cooldown after failure; plan → adversarial review → fresh worktree. Parallel loops, please pick a different issue.

PR #3877, opened 2026-07-24
fix(sync): make the aurora credential failure ledger operable

### LINKED ISSUE #3331: aurora-sync daemon: one failing credential can starve the whole sync queue
## Summary

aurora-sync's daemon has the same poison-pill starvation bug that PR #3329 fixed for kilter-sync. A single credential that fails deterministically monopolises the single-user-per-cycle queue and blocks every other user from syncing.

## Root cause

`packages/aurora-sync/src/runner/sync-runner.ts`:
- `getNextCredentialToSync` (~line 246) orders by `last_sync_at ASC NULLS FIRST` and selects one credential per cycle.
- `last_sync_at` is stamped on **success only** (~line 303). Transient failures leave it untouched; the permanent path stamps `syncStatus`/`syncError` but not `last_sync_at`.

So any credential that keeps failing keeps its `NULL`/old `last_sync_at`, sorts first every cycle, and is re-picked indefinitely — starving the rest. (kilter-sync hit this in practice via a fail-open DB error; aurora-sync is structurally identical.)

## The infra is already in place

PR #3329 added, on the shared `aurora_credentials` table:
- column `last_sync_attempt_at` (migration 0137)
- partial index `aurora_credentials_sync_attempt_priority_idx` on `(board_type, sync_status, last_sync_attempt_at) WHERE sync_status IN ('pending','active','error')`

That index already matches aurora-sync's query shape. Only the aurora-sync runner needs updating.

## Proposed fix (mirror kilter-sync)

1. `getNextCredentialToSync`: order by `last_sync_attempt_at ASC NULLS FIRST`.
2. Stamp `last_sync_attempt_at = now()` on **every** cycle outcome (success and both failure paths).
3. Keep `last_sync_at` success-only (it's the user-facing "Last synced" time on the credentials card — must not advance on a failed cycle).
4. Add runner tests asserting failures stamp `last_sync_attempt_at` (a real `Date`) but never `last_sync_at`.

See `packages/kilter-sync/src/runner/sync-runner.ts` for the reference implementation.

Surfaced during review of #3329.

marcodejongh:
## Summary

Closes #3331.

The credential-starvation fix that #3331 asks for **already landed in `main`** via commit `f9fd900e3` (sibling work to #3329). The issue's "Root cause" section describes a stale pre-`f9fd900e3` snapshot. Verified against `packages/aurora-sync/src/runner/sync-runner.ts` off latest `main`:

| #3331 proposed fix | Current state | Location |
|---|---|---|
| Order by `last_sync_attempt_at ASC NULLS FIRST` | Done | `sync-runner.ts` `getNextCredentialToSync` |
| Stamp `last_sync_attempt_at` on **every** outcome | Done — success + all failure paths via `recordSyncFailure` + deprecated `all` path | `syncNextUser` / `recordSyncFailure` |
| Keep `last_sync_at` success-only | Done — `recordSyncFailure` never touches `lastSyncAt` | `recordSyncFailure` |
| Runner tests for the above | Done | `sync-runner.test.ts` |

Plus per-credential exponential backoff (`credentialRetryReadySql`) and auth-dead quarantine to `sync_status='expired'` are already in place. So the poison-pill / Touchstone-5-months-stale scenario is no longer reproducible.

What was **missing** was making that ledger operable. A quarantined or long-flapping credential only surfaced in per-cycle stdout or a manual DB query, and the daemon's `onError` callback couldn't distinguish a one-off transient blip from a credential that's been dead for days. This PR completes #3331 by adding observability — **no scheduling change, no schema change, no migration** (the ledger columns already exist):

- **Enriched `onError` context** — carries the post-attempt ledger (`boardType`, `syncStatus`, `consecutiveFailures`, `quarantined`) via a new `CredentialSyncError` thrown by the deterministic-failure branches. Transient/DB errors honestly fall back to the pre-attempt state with `quarantined: false`.
- **Distinct, greppable log events** (structured stdout; Railway captures it — no Sentry dependency added):
  - `CREDENTIAL QUARANTINED user=… board=… reason=invalid_credentials failures=… — excluded from sync until reconnect` — fires when a credential is pushed to `expired` (out of the pool).
  - `CREDENTIAL FLAPPING user=… board=… consecutiveFailures=… nextRetryMs=… lastError=…` — fires **once**, when `consecutive_failures` crosses 5.
- **Hourly, read-only fleet health summary** in the daemon: `Sync health: N aurora credentials — active=… pending=… error=… expired=…; inBackoff=…; oldestAttempt=…`. Single aggregate scan, gated in-memory like the existing self-heal; safe to run even under an overlapping second instance (#3539) because it never writes.
- **`aurora-sync list`** now shows `last_sync_attempt_at`, `consecutive_failures`, computed retry-ready/backoff time, and the last-attempt error.
- The CLI non-verbose log filter passes the new events so they surface in production logs.

**Latent, not fixed here:** Aurora returns HTTP 422 for bad credentials, so a persistent 401/403 would still classify as transient and retry forever rather than quarantine. Noted for a follow-up if 401s are observed in practice.

### Interaction with adjacent open issues
- **#1976** (shared-sync cooldown stamps on failure): different code path (`maybeRunSharedSync`); untouched here.
- **#3539** (no cross-instance mutual exclusion): this adds no locking and doesn't fix it; the new health-summary query is read-only, so it composes cleanly with a future `FOR UPDATE SKIP LOCKED` claim in `getNextCredentialToSync`.
- **#3520** (bids crash aborts a batch): complementary — a poison-row failure already rotates out via the existing scheduler, and this PR now makes such a stuck user visible via `last_sync_error`.

## Release Notes

## Test plan

- `SKIP_TEST_INFRA=1 vp test run --project aurora-sync --reporter=agent` — 9 files, 97 pass (6 new: FLAPPING crossing, QUARANTINED transition, enriched `onError` context ×2, `formatSyncHealthSummary` ×2).
- Tightened the `recordSyncFailure` test to assert a failed cycle never advances `last_sync_at`.
- `vp run typecheck` — green.
- `vp check` on all four changed files — clean (format + lint).
- No DB-touching tests added; the health-summary query is covered via a pure formatter test.

### DIFF
--- a/packages/aurora-sync/src/cli/index.ts
+++ b/packages/aurora-sync/src/cli/index.ts
@@ -45,12 +45,14 @@ function createRunner(verbose: boolean): SyncRunner {
       ]
         .filter((part): part is string => part !== null)
         .join(' ');
-      const suffix = ledger ? ` [${ledger}]` : '';
-      console.error(
-        `Error syncing ${context.userId ?? 'daemon'}/${context.board ?? 'unknown'}:`,
-        error.message,
-        suffix,
-      );
+      const prefix = `Error syncing ${context.userId ?? 'daemon'}/${context.board ?? 'unknown'}:`;
+      // Only pass the ledger token when there is one — never a trailing empty
+      // argument (some formatters render it as a dangling space/token).
+      if (ledger) {
+        console.error(prefix, error.message, `[${ledger}]`);
+      } else {
+        console.error(prefix, error.message);
+      }
     },
   });
 }
--- a/packages/aurora-sync/src/cli/index.ts
+++ b/packages/aurora-sync/src/cli/index.ts
@@ -1,5 +1,6 @@
 #!/usr/bin/env node
 import { program } from 'commander';
+import { credentialBackoffMs } from '@boardsesh/db/queries';
 import { SyncRunner } from '../runner/sync-runner';
 import { AURORA_BOARDS } from '../api/types';
 import { AURORA_LOCATION_BOARDS, type AuroraLocationBoardName } from '../sync/locations-sync';
@@ -24,13 +25,32 @@ function createRunner(verbose: boolean): SyncRunner {
             msg.includes('Quiet hours') ||
             msg.includes('Waiting') ||
             msg.includes('No users') ||
-            msg.includes('Transient')
+            msg.includes('Transient') ||
+            // Stuck-credential observability events (see SyncRunner):
+            // CREDENTIAL QUARANTINED / CREDENTIAL FLAPPING and the hourly
+            // "Sync health" fleet summary must surface in non-verbose prod logs.
+            msg.includes('CREDENTIAL') ||
+            msg.includes('Sync health')
           ) {
             console.info(msg);
           }
         },
     onError: (error, context) => {
-      console.error(`Error syncing ${context.userId ?? 'daemon'}/${context.board ?? 'unknown'}:`, error.message);
+      // Surface the failure ledger snapshot so a stuck/quarantined credential is
+      // visible in the daemon logs (structured stdout; Railway captures it).
+      const ledger = [
+        context.consecutiveFailures !== undefined ? `consecutiveFailures=${context.consecutiveFailures}` : null,
+        context.syncStatus ? `syncStatus=${context.syncStatus}` : null,
+        context.quarantined ? 'quarantined=true' : null,
+      ]
+        .filter((part): part is string => part !== null)
+        .join(' ');
+      const suffix = ledger ? ` [${ledger}]` : '';
+      console.error(
+        `Error syncing ${context.userId ?? 'daemon'}/${context.board ?? 'unknown'}:`,
+        error.message,
+        suffix,
+      );
     },
   });
 }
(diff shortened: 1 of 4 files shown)


marcodejongh:
Thanks — addressed in f23683e:

1. **Trailing empty `console.error` arg** — fixed; the ledger token is only passed when non-empty.
2. **`getSyncHealthSnapshot` untested** — added two tests pinning the method's JS mapping (postgres-js string-count coercion via `Number()`, null `min()` staying null, empty-result defaults). The aggregate SQL itself runs against Postgres in prod, and the `not credentialRetryReadySql()` backoff filter it embeds is anchored by the already-tested JS mirror `isCredentialInBackoff` (`credential-backoff.test.ts`). Standing up a live-DB harness for aurora-sync (which today has none) felt disproportionate for a read-only observability query, so I left that out — happy to add one if you'd prefer.
3. **Magic `'kilter'`** — extracted `KILTER_BOARD_TYPE`, now shared by the scheduler filter and the health snapshot.
5. **FLAPPING double-fire under #3539** — added the acknowledging comment (harmless duplicate log line, not a correctness issue).
6. **Stamp-before-query** — added the one-line anti-spam rationale.

**4. `CredentialSyncError.syncStatus` typed as `string`** — deliberately left as `string`. The DB column is `text` (not a PG enum) and the whole runner treats `sync_status` as a free `string` (`CredentialRecord.syncStatus: string | null`, `updateCredentialStatus(status: string)`). Narrowing this one field to a union would be inconsistent and force casts at the `cred.syncStatus` fallback in `buildErrorContext`. If we want the union, the right move is to type the column/record once across the runner in a separate change.

PR #3904, opened 2026-07-25
fix(mobile): show logbook times in your local timezone

### LINKED ISSUE #3569: Use local time when displaying logbook times
It appears that logbook times are always in UTC as opposed to local time? I recorded a tick at 8:14am Pacific yesterday and in the (per climb) logbook it shows up as 3:14pm. Should we switch to local time instead?

marcodejongh:
Fixes #3569

## What was wrong

The per-climb logbook row (mobile play-drawer) showed tick times in UTC instead of local time — an 8:14am Pacific tick displayed as "3:14 PM". Root cause: `LogbookEntryRow.tsx`'s `formatClimbedAt` parsed `climbed_at` with a bare `new Date(iso)`. Since `boardsesh_ticks.climbed_at` is a naive Postgres `timestamp` column with no `Z`/offset, JS engines interpret that string as already-local — so the row displayed the raw UTC digits unshifted rather than converting them.

Fix: route through the shared `parseTickTime` helper (`@boardsesh/profile-stats`), which explicitly parses as UTC and converts to local — the same helper every other logbook display site (web's per-climb logbook, the You-tab activity feed, session cards) already uses correctly. Bundled a matching fix in `LogbookSection.tsx`'s row sort, which had the identical `new Date(naiveString)` pattern; swapped to `tickTimeMs` so ordering doesn't depend on how a given JS engine resolves an ambiguous/skipped local hour at a DST boundary (Hermes on-device isn't guaranteed to agree with V8, the test runner — see the comment on the `.sort()` call).

## Storage audit

Before touching display code, I traced how `climbed_at` gets written for every tick provenance, since a naive-column display fix is only correct if storage is a genuine UTC instant everywhere:

- **native** (GraphQL `saveTick` + the REST Aurora-compat proxy) — client `Date`/`dayjs()` → `.toISOString()`, correct UTC.
- **aurora_pull** / **json_import** — both route through `normalizeTimestamp`, which pins Aurora's naive-but-UTC export/API fields to `Z` before storing.
- **kilter_pull** — Kilter's PowerSync `created_at` is honest `Z`-suffixed ISO-8601, used directly.
- **moonboard_import** — a synthetic noon-UTC placeholder (source CSV has no time-of-day), not a real instant but harmless for local display.

All five origins land as a genuine UTC instant, so the display fix is correct across the whole population — no scoping-down needed. Two backfill migrations (`0165_kilter_dedup_backfill.sql`, `0166_aurora_json_dedup_backfill.sql`) document a real prior prod incident where naive-local timestamps got mistagged as UTC and caused duplicate ticks — useful historical context that this exact failure class has bitten this table before, though current writers are clean.

Also verified the tick **edit** path (`toEditableDate` / `LogbookEditSheet`) was already correctly UTC-aware pre-fix, so this change can't introduce a double-shift there.

## Testing

- New test: `logbook-entry-row-climbed-at.test.tsx` — pins `TZ=America/Phoenix` (fixed UTC-7, no DST) so the assertion doesn't depend on the CI host's default timezone, and verifies the row renders the local wall-clock time rather than the raw stored UTC digits.
- Extended `logbook-section-angles.test.tsx` with a chronological-sort case.
- `vp check`, `vp run typecheck:mobile`, `vp run test:mobile` (538 files / 4418 tests passed), `vp run check:mobile-bundle` all green.

## Release Notes

Your 8am flash no longer shows up as an afternoon send — logbook times in the per-climb history now match your phone's clock.

### DIFF
--- a/packages/mobile/src/components/play-drawer/LogbookEntryRow.tsx
+++ b/packages/mobile/src/components/play-drawer/LogbookEntryRow.tsx
@@ -4,6 +4,7 @@ import { useTranslation } from 'react-i18next';
 import type { LogbookEntry } from '@boardsesh/board-react';
 import { getGradeColor, DEFAULT_GRADE_COLOR } from '@boardsesh/board-constants/grade-colors';
 import { BOULDER_GRADES } from '@boardsesh/board-constants/boulder-grade-mapping';
+import { parseTickTime } from '@boardsesh/profile-stats';
 import { Text } from '../Text';
 import { Icon } from '../Icon';
 import { normalizeAscentStatus, type AscentStatusValue } from '../../lib/ascent-status-utils';
@@ -20,7 +21,12 @@ type LogbookEntryRowProps = {
 };
 
 // Hoisted so every row shares one cache lookup key instead of allocating a
-// fresh options object per call (#3155).
+// fresh options object per call (#3155). Note: the cache in
+// intl-formatter-cache.ts keys only on (locale, options), not on the
+// process's TZ — a formatter built under one TZ stays cached for that key
+// even if TZ later changes, which is fine at runtime (TZ is fixed for the
+// life of the app) but means tests exercising more than one TZ against this
+// cache must call `vi.resetModules()` between them.
 const CLIMBED_AT_FORMAT_OPTIONS: Intl.DateTimeFormatOptions = {
   month: 'short',
   day: 'numeric',
@@ -29,8 +35,15 @@ const CLIMBED_AT_FORMAT_OPTIONS: Intl.DateTimeFormatOptions = {
   minute: '2-digit',
 };
 
+// `entry.climbed_at` is a naive `boardsesh_ticks.climbed_at` string with no
+// `Z`/offset (see @boardsesh/profile-stats/format-tick-time). Parsing it with
+// a bare `new Date(iso)` gets it interpreted as browser/device-local time
+// instead of UTC, which silently displays the raw UTC digits as if they were
+// already local (#3569). `parseTickTime` recovers the true UTC instant and
+// converts it to local before we ever hand a `Date` to `Intl.DateTimeFormat`.
 function formatClimbedAt(iso: string): string {
-  const date = new Date(iso);
+  if (!iso) return iso;
+  const date = parseTickTime(iso).toDate();
   if (Number.isNaN(date.getTime())) return iso;
   return getCachedDateTimeFormat(undefined, CLIMBED_AT_FORMAT_OPTIONS).format(date);
 }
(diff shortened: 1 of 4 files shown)


marcodejongh:
## Reply to Codex review — legacy pre-PR4 timestamp concern

@codex raised a real and important point (thread on `LogbookEntryRow.tsx:46`): for tick rows whose `climbed_at` was written by the pre-PR4 buggy Aurora-pull/JSON-import writers (the "3,208-duplicate bug" documented in `packages/db/src/queries/tick-offset-inference.ts`), the stored value is the user's LOCAL wall clock relabelled as UTC — not a genuine UTC instant. Applying `parseTickTime` (always-UTC) to those rows double-shifts them.

I dug into this against the read-only prod DB replica to size it precisely:

- **Neither dedup backfill migration (`0165`, `0166`) nor the sync-time "claim" adoption path ever corrects `climbed_at`** on the surviving row — I read both migrations and the `apply-user-logbook.ts` claim UPDATE end to end. The claim explicitly says "a claim only stamps sync-tracking columns... the tick's content is unchanged." So legacy-shifted rows that got deduped/claimed kept their wrong timestamp.
- Only `origin='aurora_pull'` rows that get a genuine by-`aurora_id` **re-sync** self-heal (that UPDATE path does write the freshly-normalized `climbed_at`). `origin='json_import'` rows have no such mechanism — their `aurora_id` is a synthetic `json-import-%` placeholder, so they never hit that path, claimed or not.
- Empirically: comparing every `json_import` tick against its closest same-`(user, climb, angle)` honest-UTC counterpart (`native`/`kilter_pull`, both independently verified to always write true UTC) within a 14h window — **2,128 pairs found, and only 62 (~3%) align at Δ≈0. The rest cluster at clean whole-hour deltas** (−1h: 799, +8h: 354, −10h: 319, −11h: 235, +5h: 153, ...) — the signature of a per-user UTC-offset mislabel, exactly as the tick-offset-inference doc describes. `json_import` is 288,843 rows — 63% of the entire `boardsesh_ticks` table.

So the concern is correct and the affected population is large. Two things temper how I'm handling it in *this* PR:

1. **This isn't a regression introduced here.** `@boardsesh/profile-stats`'s `parseTickTime`/`formatTickAbsoluteTime` (the same always-UTC helper this PR now also uses in `LogbookEntryRow`) has been live in the You-tab Logbook, session cards, and profile stats charts since `5e8e13b20` (2026-06-02) — over a month before the PR4 fix (2026-07-08) even shipped, and 7+ weeks before this PR. Those surfaces already display the same legacy rows the same (wrong, for that subset) way today, in production. This PR brings the per-climb logbook into *consistency* with the rest of the app, not into a new failure mode.
2. **The correct fix is a data backfill, not a display-layer change**, and it's a big enough, sensitive enough change (mutating ~280K+ historical rows' timestamps) that it deserves its own scoped, reviewed effort — ideally reusing the already-built `inferUserUtcOffsetSeconds` machinery to compute and apply a real correction, the way `0165`/`0166` did for the duplicate-row cases. That's out of scope for a display bug fix (#3569) and I'm not comfortable bundling a historical-data mutation into this PR without dedicated review.

Filing a follow-up issue with this evidence now and will link it here. Resolving this thread as "acknowledged, tracked separately, not a regression from this PR" — reopen if you disagree with that framing.

marcodejongh:
Follow-up filed: #3909 — full evidence, scope (~280K+ rows), and suggested backfill approach.

Issue #3876, opened 2026-07-24
user-boards-serial-uniqueness.integration.test.ts assumes the SYSTEM user row pre-exists, fails FK 23503 on a fresh dev-db image


marcodejongh:
## Summary

`packages/db/scripts/user-boards-serial-uniqueness.integration.test.ts` (regression coverage for migration 0131, the system-catalog serial-uniqueness exemption) assumes the SYSTEM user row (`id = 00000000-0000-0000-0000-000000000000`) already exists in the `users` table. On a freshly-pulled `ghcr.io/boardsesh/boardsesh-dev-db:latest` image it doesn't, and the test fails with a foreign-key violation instead of exercising the exemption it's meant to cover.

## Repro

Against a brand-new local dev-db container (`docker compose up -d postgres redis` + `vp run db:up` to apply pending migrations), the test's very first insert fails:

```
$ DATABASE_URL="postgresql://postgres:password@localhost:5432/main" vp run test:db
...
not ok 72 - user_boards serial uniqueness — system catalog exemption
  error: |-
    Failed query:
        INSERT INTO user_boards (uuid, slug, owner_id, board_type, layout_id, size_id, set_ids, name, serial_number, is_owned)
        VALUES ($1, $2, $3, 'kilter', 1, 10, '20', $4, $5, false)
    params: aaaa1111-0000-0000-0000-000000000001,serialuniq-sys-1,00000000-0000-0000-0000-000000000000,Serial Test Board,SERIALUNIQ-TEST-75934
```

Reproduced directly against the DB outside the test harness to get the underlying Postgres error:

```
ERROR: insert or update on table "user_boards" violates foreign key constraint "user_boards_owner_id_fkey"
CODE: 23503
DETAIL: Key (owner_id)=(00000000-0000-0000-0000-000000000000) is not present in table "users".
```

[...]

Issue #3541, opened 2026-07-07
Same Aurora account linked to two Boardsesh users — prod pair needs manual resolution


marcodejongh:
## Summary
One Aurora account can be linked to two Boardsesh users. The tick upsert conflicts on `aurora_id` alone with no user scoping, and the incremental sync cursor (board_user_syncs) is keyed by the Aurora user id, so both Boardsesh accounts share one cursor and one set of tick rows. Whichever account synced first owns the ticks; the second user's sync reports status 'active' but their logbook stays empty, with no error logged. The circuits path can also grant the second user an 'owner' row on the first user's playlist.

The write-side guard (kilter-style foreign-id detection + skip-and-log + a surfaced credential error, plus user-scoping the cursor) ships in PR6. This issue tracks the remaining backlog item: deciding and applying the manual resolution for the existing prod pair, which the guard does not clean up on its own.

## Failure scenario
User A links Tension account X and syncs 229 ascents. User B (same person's second Boardsesh account, or a shared gym login) links the same Tension account. B's daemon cycle runs, logs 'Successfully synced', but all 229 ticks conflict on aurora_id and stay attached to A; B's profile shows 0 climbs. B's circuits also attach ownership rows onto A's playlist rows.

## Evidence
`packages/aurora-sync/src/sync/user-sync.ts:190`

```
user-sync.ts:186-205 (ascents onConflictDoUpdate target: boardseshTicks.auroraId; SET list has no user_id), 241-253 (bids, same), 345-365 (circuits playlist conflict on playlists.auroraId then playlistOwnership insert for the syncing user). Prod: aurora_credentials has tension aurora_user_id=144574 linked to users 20a34cb6-bf4d... (active, last_sync 2026-07-07T07:32, 0 tension ticks) and 4b3b5674-4c0d... (active, 229 tension ticks). Also kilter aurora_user_id=44710 linked to 2 users (handled safely by the kilter path).
```

## Suggested fix
Guard ships in PR6: detect aurora_ids owned by a different user (single SELECT like kilter's applyLogs), skip-and-log those rows, set sync_error on the credential, and user-scope the board_user_syncs cursor; same guard for the circuits/playlist ownership insert. Backlog decision (this issue): choose which Boardsesh user should own tension aurora_user_id=144574's 229 ticks (or unlink the duplicate), then re-point and re-sync. No automated deletion without sign-off.

[...]

Issue #3525, opened 2026-07-07
Ratings REMOVE hard-delete destroys locally-adopted fields: the 'kilter-origin rows only' justification is false once a Boardsesh-origin rating adopts a kilter_id


marcodejongh:
## Summary
applyClimbRatings hard-deletes on REMOVE, scoped to kilter_id IN (removeIds) (user-sync.ts:845-847), justified by 'it only removes kilter-origin rows' and 'the row is fully reconstructable from the next PUT'. But the upsert deliberately adopts kilter_id onto Boardsesh-origin rows (:905 COALESCE(EXCLUDED.kilter_id, existing)), and the re-inserted row resets weight to null (:876) and comment to Kilter's value ('' when null, :872). So a PowerSync snapshot re-delivery (REMOVE-before-PUT, the exact case the logs path soft-detaches for) round-trips an adopted row through deletion and loses every Boardsesh-only field (weight, local comment); a real Kilter-side delete additionally erases the user's local rating outright instead of detaching it.

## Failure scenario
Future local-rating writer stores weight=62.5 and a comment on a rating row; the user also rated the climb on Kilter so the row adopted kilter_id. PowerSync reconnect re-delivers the snapshot as REMOVE+PUT → row deleted then re-inserted with weight null, comment '' → local data silently gone. Latent today (weight is 0-populated and no local writer exists), but the asymmetry vs the tick path is already encoded.

## Evidence
`packages/kilter-sync/src/sync/user-sync.ts:845`

```
user-sync.ts:831-848 (hard DELETE where kilter_id IN removeIds; comment claims 'Boardsesh-origin ratings have a NULL kilter_id and never match'), :905 (kilterId adoption COALESCE contradicts that claim), :876 (weight: null on rebuild), :872 (comment '' on rebuild). Prod: 0 rows with weight yet (count(*) FILTER (WHERE weight IS NOT NULL) = 0), confirming latent-only.
```

## Suggested fix
Mirror the tick path: soft-detach (kilter_id := NULL) instead of DELETE, or restrict the hard delete to rows that have never carried local-only data; alternatively preserve weight/comment via the natural-key upsert by ordering PUTs before REMOVEs and skipping REMOVEs whose kilter_id was re-PUT in the same snapshot.

Found by the 2026-07 upstream-sync data audit (verified against prod). Related program PRs: ascent-count fix, star-scale repair (in flight).

PR #3931, opened 2026-07-25
fix(sync): stop circuits sync cross-linking playlists between accounts sharing one board login

### LINKED ISSUE #3526: applyCircuits cross-links playlists between Boardsesh accounts sharing one Kilter account: the kilterId upsert adopts the other user's playlist and grants this user an owner row
## Summary
The logs path defends against one Kilter account linked to two Boardsesh accounts (foreignKilterIds skip-and-log, user-sync.ts:692-703), but applyCircuits has no such guard: the playlists upsert targets the global playlists.kilterId (:980) so user B's sync UPDATEs user A's playlist row (name/description/visibility overwritten, :982-989) and then unconditionally inserts an 'owner' ownership row for user B (:998). Both accounts now co-own and mutually overwrite one playlist row; a circuit REMOVE from either stream deletes it for both (the ownership EXISTS check at :944-951 passes for each). Same duplicate-account shape the logs path deliberately refuses to merge.

## Failure scenario
One person (or two people sharing a gym Kilter login) links the same Kilter account from two Boardsesh accounts. Account B's first sync upserts the same circuit_uuid → adopts A's playlist and adds B as owner. B renames or deletes the circuit locally/on Kilter → A's playlist silently changes or disappears.

## Evidence
`packages/kilter-sync/src/sync/user-sync.ts:998`

```
user-sync.ts:966-991 (onConflictDoUpdate target: playlists.kilterId with no ownership check before update), :998 (`tx.insert(playlistOwnership).values({ playlistId, userId, role: 'owner' }).onConflictDoNothing()`), contrasted with :692-703 where the identical situation on logs is skip-and-log by design.
```

## Suggested fix
Before upserting, SELECT the existing playlist by kilterId and verify this user has (or should get) ownership; on foreign ownership, skip-and-log like the logs path (or explicitly support co-ownership as a product decision rather than an accident).

marcodejongh:
## Summary

Closes #3526.

When one board account is linked to two Boardsesh users, the circuits→playlists sync cross-links them. `playlists.kilter_id` and `playlists.aurora_id` are each backed by a **global** unique index, so user B's `ON CONFLICT (kilter_id) DO UPDATE` lands on user A's playlist row — overwriting its name, description and visibility — then an unconditional `INSERT INTO playlist_ownership` hands B an `owner` edge, and the climbs delete-and-reinsert wipes A's list. After that a circuit delete from either user's stream destroys the playlist for both, because the old `EXISTS` ownership check passes for each of them.

### The invariant this enforces

> A playlist row carrying an upstream circuit id has **exactly one** Boardsesh sync owner — the user whose sync created it. A sync running for any other Boardsesh user that resolves the same upstream id must not update the row, must not touch its `playlist_climbs`, must not insert an ownership row, and must not delete it. It skips and logs.

Concretely, for a resolved upstream circuit id, let `O` be the set of users with a `role = 'owner'` edge on the matched playlist. A write for user `U` is permitted iff `O = ∅` (fresh create, or an orphaned row) or `O = {U}`. `editor` / `viewer` rows never block — future collaboration stays possible. `|O| > 1` is **ambiguous** and refused in *both* directions, so neither party to an existing cross-link can overwrite or destroy the shared row.

**Four writers had the same defect** and all four now share one guard:

| Writer | |
|---|---|
| `packages/kilter-sync/…/user-sync.ts` `applyCircuits` | the issue as filed |
| `packages/aurora-sync/…/user-sync.ts` circuits case | **the one with actual prod damage** |
| `packages/aurora-sync/…/json-import.ts` | key derivation already user-scoped; guard is defence in depth |
| `packages/web/app/lib/data-sync/aurora/user-sync.ts` | legacy `/api/v1/{board}/proxy/{login,user-sync}` route, found in review |

The guard is two halves: a JS decision gate reading a batched owner lookup inside the transaction, plus a `NOT EXISTS` `setWhere` on the `ON CONFLICT … DO UPDATE` that re-checks at statement time. The second half matters because #3539 (no cross-instance mutual exclusion for sync daemons) means two daemons can both read "no playlist yet" and both insert.

### Prod numbers, with the honest caveat

Read-only against prod:

| | |
|---|---|
| Playlists with `kilter_id` | 82, across 27 users, **0 cross-linked** |
| Boardsesh users sharing a Kilter Keycloak `sub` | **0** (36 OAuth links, all distinct) |
| Playlists with >1 distinct owner | **44** (88 ownership rows, 259 `playlist_climbs`) |
| — via `json-import` | 36 |
| — via Aurora `user-sync` circuits | 8 |
| Upstream accounts linked to 2 Boardsesh users | 2 — kilter `44710` (both expired), tension `144574` (both active) |
| Expired credentials (the re-claim hatch) | kilter 9, tension 1 |

**The caveat**: of the 4 affected users, two are the maintainer's own accounts, one is a test account, and the last pair is one person's two accounts (`pmbmosk@gmail.com` / `Pmbmosk@gmail.com`, split by case-sensitive email matching). **No different-human exposure has materialised.** On incidence alone this is not a P1.

It was re-labelled P1 anyway, on trigger rather than incidence: the failure mode is a silent cross-account write and destroy with no alarm and no audit trail, and it's reachable without a shared login — unlinking deletes the credential and the mapping but leaves the playlists, and 9 kilter credentials sit `expired` right now, each an open re-claim. Reasoning and counter-argument are both on #3526.

The 36 `json-import` rows are a closed chapter: the original importer (`8fe79f60d`) hashed `${boardType}:${name}:${created_at}` with **no user id**, so two people importing a circuit with the same name and timestamp collided globally. `45cef340a` added the user id. That derivation is now an extracted, tested function so it can't silently regress.

### Heads-up: the 44 existing rows go read-only to sync

They have two owners, so the guard reads them as `ambiguous` and refuses to write them. That's the intended default — neither party can destroy the shared row — but it means the 8 tension playlists stop receiving upstream updates until the accounts are merged. The 36 `json-import` ones were never on a sync path, so nothing changes for them.

**No repair runs in this PR.** The full analysis — creator identification via `min(playlist_ownership.created_at)` (clean on all 44, spread 100 minutes to 96 days), both repair options with row counts, and the recommendation to use #3278's `merge-accounts.ts` for the tension pair — is a comment on #3541. Nothing has been executed and nothing will be without sign-off.

### Also in here

- **A user-facing `sync_error`, and it actually reaches the board card.** A refused circuit used to leave the second user with an empty playlist list and no way to tell "sync is broken" from "I have no circuits". The credential now carries a **machine-readable code** — `duplicate-board-account-link:circuits`, the same value the aurora runner already used as its skip reason — while staying `active` (the cycle really did succeed for logs, ratings and everything else). On Aurora it's derived from **state**, not from what the cycle refused: Aurora's sync is incremental, so a per-cycle flag would clear itself the moment nothing changed upstream.

  A code rather than a sentence because the first draft of this shipped an English sentence generated in the daemon, and the surface that survives — the mobile board card — can't translate one. Worse, that card threw `credential.syncError` away entirely and rendered a hardcoded red **"Error"** pill, so both prod-affected accounts would have got a permanent red pill on a healthy, actively-syncing credential with nothing to act on. The card now renders localised copy (en-US / es / fr) in a **warning** block, matching the pending-sync notice next to it. An unrecognised `sync_error` still gets the generic error copy on mobile and its raw text on web, so the legacy free-text values other paths write are shown, not swallowed.
- **A landmine comment + test on `pushPendingCircuits`** (`push-back.ts`, still stubbed behind `pushNotWired`). The day it's wired it starts stamping `kilter_id` onto Boardsesh-*origin* playlists, at which point the REMOVE branch would hard-delete a playlist the user built by hand — the #3525 shape. The test asserts the owner guard holds for that direction; the soft-detach decision belongs to #3525.
- **`getNextAuthUserId`** in the web proxy picked its user with an unordered `LIMIT 1`. Now ordered by `created_at`. Deterministic isn't correct — the guard is what makes a wrong answer harmless — but a coin flip was worse.

### The index this needs ships in #3963

> **Base is #3963, not `main`.** This PR no longer carries a migration. Four PRs each generated a `0187` against a `main` that tops out at `0186`, so all four schema changes were collapsed into one migration on `chore/collapse-0187-migrations` (#3963) and this branch now carries only query and logic code. GitHub retargets this to `main` when #3963 merges. Nothing about the review below changes — the DDL is byte-identical, just generated in one place.

The statement it depends on:

```sql
CREATE INDEX "board_circuits_user_idx" ON "board_circuits" USING btree ("board_type","user_id");
```

It backs the new duplicate-owner state check: a foreign key does not index its own columns, so that lookup seq-scanned `board_circuits` once per credential per cycle, and the healthy path matches nothing so `LIMIT 1` never cut the scan short.

**Lock window.** `CREATE INDEX` without `CONCURRENTLY` blocks writes to the table while it builds, so here is the size it builds against, measured read-only in prod:

| | |
|---|---|
| `board_circuits` rows | **857** (tension 673, kilter 156, decoy 26, grasshopper 2) |
| Heap size | **144 kB** (264 kB incl. existing indexes) |
| For scale | `board_climb_holds` is 9.6M rows / 1853 MB — this table is ~0.01% of it |

At 144 kB the build is a single-digit-millisecond sort, so the write lock is effectively unobservable, against a table only the sync daemons write to.

`CONCURRENTLY` is not an option regardless: drizzle's pg-core migrator runs each migration inside `session.transaction(...)`, and Postgres forbids `CREATE INDEX CONCURRENTLY` in a transaction block. So the justification is "non-concurrent is fine because the table is small", not "concurrent wasn't needed". If `board_circuits` ever grows into millions of rows, a future index on it needs its own out-of-band path.

**No constraint change.** A `UNIQUE` on the upstream id would fail against the 44 legacy rows, and skip-and-log needs no new column. (`aurora-credentials.ts:88-94` already tracks the eventual DB-level unique on the *link* side, blocked on #3541.)

### Related

- **#3541** — same root-cause family. Its own suggested fix asked for this guard on circuits; the logs half shipped in #3557. Repair analysis posted there.
- **#3539** — separate fix, but it's why the SQL-level `setWhere` guard exists rather than trusting the in-transaction read alone.
- **#3525** — same class, different table. Comment + test added, not fixed.
- **#3278** — still **open**. Prod now has 27 duplicate-by-case account sets (up from 23 in its audit); its `merge-accounts.ts` is the right vehicle for the tension pair.

Follow-ups deliberately not done here:

- The copy doesn't split `foreign` from `ambiguous`. One wording covers both ("another Boardsesh account is connected to this board login, so playlist sync is paused for both of you"), which is true either way; threading the decision out of `hasForeignOwnedCircuitPlaylists` so each case gets its own sentence is a separate change.
- The *logs* paths still skip foreign ticks silently with no `sync_error` — a second user whose sends aren't syncing still gets nothing.
- The `if (!playlist) continue;` race-guard branch (aurora-sync + web proxy) still returns without a log line or a `refusedCircuitUuids` entry. The `sync_error` is unaffected — the aurora runner asks a state question — but the #3539 race path has no observability.
- The web proxy's `ownersByAuroraId` map isn't updated after a successful insert the way kilter-sync's is. Inert (that route has no REMOVE path), but it's an inconsistency across three copies of the same guard.
- Nothing stops a fifth writer reintroducing this. A user-scoped `UNIQUE` would fail against the 44 legacy rows, so the invariant lives in four hand-edited call sites with no DB-level backstop.

## Release Notes

- Your playlists are yours. If someone else connects the same board account, their sync can no longer rename, empty, or delete the circuits you've synced — and if two accounts are already tangled together, neither can overwrite the other.
- When your circuits can't sync because another Boardsesh account is connected to the same board login, the board card says so in your language, instead of leaving you with an empty playlist list. It reads as a heads-up, not a broken account — your sends and ratings are still coming through.

## Test plan

- `vp check` — clean (only pre-existing warnings on `main`, owned elsewhere).
- `vp run typecheck` — green.
- `vp test run --project sync-runtime --project kilter-sync --project aurora-sync` — 318 passed.
- `vp run test:mobile` — 4484 passed.
- `vp test run --project web packages/web/app/components/settings` — 47 passed.
- `vp run check:i18n` + `vp run check:i18n:orphans` — clean; `@boardsesh/i18n` catalog parity 60 passed.
- `vp run check:mobile-bundle` — OK (the board card's new `@boardsesh/shared-schema/sync-error-codes` subpath resolves under Metro).
- `VITEST_POOL_ID=solo-3526 vp test run` — full suite green.
- `vp run test:db` — 95 pass / 1 fail. The failure is `user-boards-serial-uniqueness`, which is **#3876** (assumes the SYSTEM user row pre-exists, so it fails on a fresh dev-db image). Confirmed unrelated by reproducing it identically with this branch's changes stashed.

New coverage, per path:

- **`@boardsesh/sync-runtime`** — the pure decision function across all four outcomes, join fan-out (same owner twice ≠ ambiguous), `editor`/`viewer` not blocking.
- **kilter `applyCircuits`** — foreign PUT asserts **zero inserts, zero updates and zero deletes** separately (the ownership insert and the climbs wipe were each independently destructive); ambiguous refused both ways; orphan claimed; REMOVE refused for a foreign owner *and* for a co-owned playlist; REMOVE for an unknown circuit skipped quietly rather than counted as a duplicate-account signal; mixed batch counts every refusal; PUT+REMOVE in one batch still deletes (the owner map is written back mid-loop, or upstream's tombstone would be dropped forever); empty `returning` from the race guard abandons the op instead of dereferencing an undefined id; the `push-back` landmine direction.
- **aurora circuits + json-import + web proxy** — the same matrix where applicable, plus the user-scoped `json-import` key derivation (two users, same circuit name and timestamp → different ids). Worth stating plainly: **the web proxy guard had zero effective coverage until this file existed** — mutation testing showed that deleting the entire guard block passed every test in the repo. That is the single strongest reason the new test file is here.
- **SQL the stubs can't see** — `PgDialect` render assertions on both the `NOT EXISTS` conflict guard (correlated to `"playlists"."id"`, `role = 'owner'`, `user_id <>`) and the owner lookup (LEFT JOIN, role filter, right column). Without these, deleting `role = 'owner'` or swapping the join passed every other test.
- **Both runners** — the `sync_error` code written on a skip (asserted against the exported constant, so an English sentence can't creep back in), cycle still counts as a success, `null` on a clean cycle, and — Aurora only — the code survives a later cycle where the incremental sync returned no circuits at all.
- **The mobile board card** — an active credential carrying the code renders the localised warning copy and *not* the generic error copy, and never leaks the raw code; an unrecognised `sync_error` still gets the generic error copy; a clean credential gets no error affordance at all.

**Known untested**: the `json-import` guard has no behavioural coverage. Its key derivation is user-scoped, so the branch is near-unreachable without a deliberate key change, and the guard is defence in depth rather than the primary fix. The key derivation itself *is* tested.

**One rule, learned twice.** Anything driven through a SQL-blind stub must render the predicate the stub *recorded*, never rebuild it in the test — a rebuilt copy only proves drizzle renders `eq` as `=`, and drifts the moment production changes. Three assertions had this shape and are fixed; each was then mutation-verified. Flipping `ne` to `eq` in the duplicate-owner check (which would put a permanent, false "circuits aren't syncing" banner on every healthy Aurora user), dropping `foreignPlaylistOwnerGuard` from the REMOVE (which restores the destructive half of this bug), and deleting the fail-open `.catch` all now fail a named test. Before, all three passed.

Reviewed by a paired QA agent over four rounds. Round two found the web proxy writer, the PUT+REMOVE ordering bug and the self-clearing Aurora `sync_error`. Round three found the unguarded state-check throw (a cosmetic read could fail an otherwise successful cycle), the missing index, and a render-only *twin* of the owner lookup sitting beside the executed query — so the SQL assertions were testing a copy that could drift. All three are fixed; the query is now one expression that the tests render and the code awaits.


marcodejongh:
## The Aurora `sync_error` check, validated against prod (read-only)

`hasForeignOwnedCircuitPlaylists` is the one piece of this PR that asks a question of live data rather than of a payload, so I ran its exact join against prod for every Tension user. No writes.

```sql
SELECT count(*)
FROM board_circuits c
JOIN playlists p          ON p.aurora_id = c.uuid AND p.board_type = c.board_type
JOIN playlist_ownership po ON po.playlist_id = p.id AND po.role = 'owner'
WHERE c.board_type = 'tension' AND c.user_id = <aurora id> AND po.user_id <> <boardsesh id>;
```

| User | conflicting | resulting `sync_error` |
|---|---|---|
| `20a34cb6` (Peter M) | **34** | set |
| `4b3b5674` (Pmbmosk) | **8** | set |
| every other Tension user (110 of them) | **0** | null |

Both halves of the known duplicate pair are flagged, and nobody else is — no false positives across the fleet. The asymmetry is real, not a bug: 34 of the shared account's circuits map to playlists `20a34cb6` doesn't own, while only the 8 co-owned ones look foreign from `4b3b5674`'s side. Both users genuinely have circuits that can't sync, so both get told.

Also confirms the assumption the check rests on: `board_circuits` for aurora id `144574` holds **34 rows, all 34 with a matching playlist** — it really is cumulative, so the check doesn't depend on what turned up in the current incremental delta. That was the whole point of deriving the flag from state instead of from a per-cycle counter.

marcodejongh:
### Since the last review: the message now reaches the surface that survives

`claude[bot]`'s earlier rounds were right that the `sync_error` half was the weak part. It was worse than "English-only" — the mobile board card (`BoardAccountsSection.tsx`) discarded `credential.syncError` entirely and rendered a hardcoded red **"Error"** pill. So the release note was false on mobile, and both prod-affected tension accounts would have got a permanent red pill on a healthy, actively-syncing credential with no explanation and nothing to act on.

Fixed in `023e8d75f`:

- `sync_error` is now the machine-readable code `duplicate-board-account-link:circuits` (the constant the aurora runner already used as its skip reason, lifted to `@boardsesh/shared-schema/sync-error-codes` so daemons and clients share one definition). Both runners write it.
- `upstreamPlaylistSyncErrorMessage` is deleted — a sentence generated server-side is the one thing an `es`/`fr` card can't translate.
- The board card renders it as a **warning** block (icon + secondary label, same treatment as the pending-sync notice), with copy in en-US / es / fr. Warning, not error: the cycle succeeded and sends/bids/ratings are still syncing — only the playlist mirror is paused.
- Unrecognised `sync_error` values keep the generic error copy on mobile and their raw text on web, so legacy free-text from other paths is still shown.

Tests: three board-card cases (code → localised warning and *not* the generic error copy, and the raw code never leaks; unknown value → generic error copy; clean credential → no error affordance), plus both runner assertions now compare against the exported constant so an English sentence can't creep back in.

**Still deliberately open**, now listed in the PR description rather than left implicit: the copy doesn't split `foreign` from `ambiguous` (one wording that's true for both); the `if (!playlist) continue;` race-guard branch has no log line or `refusedCircuitUuids` entry; the web proxy's `ownersByAuroraId` isn't updated after a successful insert; and there's no DB-level backstop against a fifth writer.

`claude-review` is red on infrastructure, not code — "The job was not acquired by Runner of type hosted even after multiple attempts". Every other check is green.

Issue #3906, opened 2026-07-25
Queue sequence assignment is a non-atomic RMW — concurrent party mutations can clobber writes with identical sequence


marcodejongh:
## Queue sequence assignment is a non-atomic RMW — concurrent party mutations can clobber writes with identical sequence

Surfaced during the #2860 (board presence drift) investigation.

Each queue mutation assigns its sequence number with a read-modify-write, not an
atomic compare-and-swap. `updateQueueState`
(`packages/backend/src/services/room-manager/queue-state.ts:67-99`) reads the
current version/sequence, computes `+1`, and writes — with no version guard on
the write. Several mutation resolvers pass **no** `expectedVersion` at all:

- `packages/backend/src/graphql/resolvers/queue/mutations.ts:202` (setQueue via updateQueueOnly)
- `:316`, `:361`, `:451` (updateQueueState calls)

Two party members mutating the queue at the same moment (or two backend
instances handling their sockets) can both read the same prior sequence, both
write `prior+1`, and one write silently clobbers the other — same sequence
number, one mutation lost.

**Why it's nasty:** the client hash watchdog can't catch it. After the clobber,
the server's own state is internally consistent, so every client resyncs to the
*same wrong* state and all hashes agree — no drift is ever detected. It's silent
queue write-loss, not a detectable divergence. (This is distinct from the
presence-roster drift #2860 fixed; that path had no sequencing at all, this one
has racy sequencing.)

**Fix direction:** make sequence assignment atomic — a Lua CAS on the Redis
sequence key, or a single-statement DB CTE that bumps sequence only when the
read version still holds — and/or thread `expectedVersion` through the four
mutations above so a stale write is rejected and retried instead of clobbering.
Needs a real-Redis concurrency test (two writers racing one session).

Likelihood is bounded (needs concurrent mutations in one party), but it's a real
correctness hole with a silent failure mode.

Labels: backend, bug, priority:P2

marcodejongh:
Fixed and merged. **PR #3932** — `fix(backend): make queue sequence assignment atomic (#3906)`, commit `31ab02a78892b8b60e424aae1ae2ca42b38a46f6`, merged 2026-07-26 and present on `origin/main`.

The read-modify-write is gone. `updateQueueState` now does a single atomic compare-and-swap inside one Lua script against Redis, and the code cites this issue by number in several places — `packages/backend/src/services/room-manager/queue-state.ts`:

[...]

PR #3932, opened 2026-07-25
fix(backend): make queue sequence assignment atomic (#3906)

marcodejongh:
## Summary

Refs #3906 — **partial fix, deliberately not closing the issue.** The `setQueue` residual is tracked separately as #3933 (see "Decisions worth flagging" below); #3906 stays open until that lands.

Queue sequence/version allocation was a read-modify-write across two Redis round trips:

```ts
const currentState = await getQueueState(sessionId, redisStore); // HGETALL
// ...compute version + 1, sequence + 1 in Node...
await redisStore.updateQueueState(sessionId, { ... });           // HMSET whole hash
```

Two mutations overlapping anywhere in that window both read the same counters, both wrote `+1`, and the second `HMSET` replaced the entire `queue` array. Two distinct symptoms fall out of the same race:

- **A climb a crew member just added disappears.** The later write carries a queue array computed from a pre-add snapshot.
- **An event goes missing with no server-side trace.** Both writers land on the *same* sequence number, and the client's sync gate drops the second as a stale duplicate.

`updateQueueState` also took an `expectedVersion` parameter it never compared against anything, so every `VersionConflictError` retry path aimed at it was dead code.

This has to hold across *processes*, not just within one event loop — Redis is shared by every backend instance, and Railway's rolling deploys run two at once even at a single replica. An in-process mutex would not have helped.

## Approach

**Compare and write in one atomic step.** `UPDATE_QUEUE_STATE_CAS_SCRIPT` (`services/redis-session-store.ts`) reads the stored version, compares it to `expectedVersion`, and writes the new queue inside a single Lua invocation. On mismatch it returns `CONFLICT` and writes nothing; `queue-state.ts` raises `VersionConflictError` and the new `withQueueVersionRetry` (`resolvers/shared/queue-retry.ts`) re-reads and recomputes, up to 3 attempts. Callers with nothing to recompute pass `CAS_ANY_VERSION` to skip the comparison.

**Dormancy floor.** A session past the 4h Redis TTL still has durable counters in Postgres. When the hash is missing the script returns `NEEDS_FLOOR` without writing; the caller reads Postgres and retries with `versionFloor`/`sequenceFloor`, and the script takes the max. Restarting at 1 would rewind the sequence clients gap-check against. Same shape as the existing board-presence reseed in `allocateBoardSeqAtLeast`.

**A second, separable bug found on the way.** Each instance debounces its own `pendingWrites` map, so when a session is touched on two instances their 30s Postgres flushes race and the loser rewinds the durable row's counters. `writeQueueStateToPostgres` now guards the upsert with `setWhere: sequence < excluded.sequence`. `sequence` is the right guard rather than `version`: it advances on every write path and has no historical divergence to trip over on rows predating this PR. Raw `sql` because `setWhere` must reference the `excluded` pseudo-table, which the query builder can't express.

**Telemetry for the invisible half.** The `sequence > version` interleaving leaves a permanent fingerprint on the Postgres row — that's how this was confirmed in production. The duplicate-sequence interleaving leaves *nothing*; the event is simply dropped client-side. Web now emits `QueueSyncStaleEventIgnored` from the sync gate's `ignore-stale` branch, mirroring the event mobile already sends, with a `staleness` discriminator: `duplicate` (same sequence — a collision, or the controller path's deliberate co-sequenced pair, which `eventType` tells apart) vs `behind` (an older sequence after a reconnect — expected, not a defect).

## Decisions worth flagging for review

- **`setQueue` is deliberately exempt.** Its payload is entirely client-supplied, so there is nothing to recompute against a concurrent write, and replacing server state wholesale is the mutation's contract. It passes `CAS_ANY_VERSION` — it needs a unique sequence, not a conflict. Accepted consequence, tracked as **#3933** rather than fixed here: a peer's `addQueueItem` landing inside a `setQueue` window is still overwritten, and web's drag-to-reorder takes this path. This is why #3906 is not closed by this PR.
- **`stateHashOrdered` is now stored in the Redis hash** alongside `stateHash`, so the CAS can hand the prior pair back to `setQueue`'s redundant-resync diagnostic in one round trip. Sessions written before this rollout have no stored value; reads fall back to recomputing from the stored queue.
- **Duplicate sequences remain legal on one path.** `setClimbFromLedPositions` publishes `FullSync` and `CurrentClimbChanged` under one sequence on purpose, so the ESP32 still receives the originating `clientId`. Peers drop the second as a stale duplicate. Sequence allocation is per state *write*, not per publish.
- **`updateQueueOnly`'s optional `knownState` is scoped to `Pick<QueueState, 'currentClimbQueueItem'>`** and deliberately ignored on the no-Redis path. Letting it supply `version` there would make the guard vacuous — the caller's `expectedVersion` comes out of that same snapshot, so the comparison would always agree. Only the Redis path can take the caller's word, because the CAS re-reads the authoritative version itself.

## Test plan

`vp run typecheck` and `vp check` both clean. Affected backend files run green (12 files / 173 tests); the full backend suite is CI's authority.

New `queue-sequence-atomicity.test.ts` covers three layers — five deterministic interleavings, the real Lua script against a live Redis, and 12 concurrent mutations asserting contiguous unique sequences.

Every mutation now wrapped in `withQueueVersionRetry` has its own pinned interleaving, because each plumbs the retried result back to its published event differently: `reorderQueueItem` through mutable `result*` locals, `mirrorCurrentClimb` through a spread return, `navigateQueue` through an `{ item, update }` pair. A retry that recomputes the queue correctly but publishes the *first* attempt's sequence still breaks the client's gap check, so the tests assert the published sequence matches final stored state, not just that no climb was lost.

**These were mutation-tested rather than trusted**, because a test that passes for the wrong reason is the real risk here:

| Guard deleted | Result |
| --- | --- |
| conflict check in the mock's CAS transcription | 9 fail — all five interleavings, both version guards, the load test |
| conflict check in the real Lua script | 2 Layer-3 tests fail |
| `setWhere` on the Postgres upsert | its test fails (durable row rewinds 9 → 4) |
| just `resultSequence = result.sequence` in `reorderQueueItem` | the reorder interleaving fails, and nothing else |
| *(nothing — headline guard run against `origin/main`)* | fails with `resolved "{ version: 1, sequence: 2 }"` |

That last row is the production `sequence > version` fingerprint reproduced as a unit test.

One fixture subtlety worth knowing if you touch these tests: the gate's indices count CAS calls, and a mutation only issues **one** when the session already has a Redis hash. Without one the script returns `NEEDS_FLOOR` first and the same mutation burns **two** indices. The gated tests therefore use `createDurableSession` (which seeds the `board_sessions` row) rather than `registerAndJoinSession`; the helper documents the invariant.

### Manual QA (party mode, two browsers)

- [ ] Two clients in one session; both add a climb at the same moment — both climbs land, neither disappears
- [ ] Reorder on A while B adds — no lost item, no duplicate
- [ ] Leave a session idle past the 4h Redis TTL, then mutate — sequence continues upward rather than restarting

## Release Notes

none

Backend correctness fix with no user-facing note yet. The race this closes is real and user-visible in party mode, but `setQueue` can still overwrite a concurrent add (#3933), so a "shared queues are reliable now" note would overclaim. The changelog entry belongs with #3933, once the guarantee actually holds end to end.


marcodejongh:
All three addressed — thanks. Issue 2 was a better catch than it looks, and Issue 3 was right that the retry wiring differs per mutation.

**Issue 1 — redundant read.** Fixed. `updateQueueOnly` now takes an optional `knownState`, and the two callers inside `withQueueVersionRetry` (`addQueueItem`, `reorderQueueItem`) pass the state the wrapper already read, so they're back to one Redis read per attempt like `removeQueueItem`.

One correction to the suggested fix, though: I scoped the parameter to `Pick<QueueState, 'currentClimbQueueItem'>` and made the no-Redis path deliberately ignore it. My first cut threaded `version` through too, which would have made that path's guard **vacuous** — the caller's `expectedVersion` comes out of that same snapshot, so `knownState.version !== expectedVersion` could never fire. Only the Redis path can take the caller's word for anything, because the CAS re-reads the authoritative version itself.

**Issue 2 — inaccurate comment.** Rather than soften the comment I made the ordering real: the gate gained a single-slot `release(n)`, so the test releases B's CAS alone, awaits B to completion, and only then releases A.

Chasing that surfaced a latent fragility in the gate itself. Indices count CAS calls, but a mutation only issues **one** when the session already has a Redis hash — without one the script returns `NEEDS_FLOOR` first and the same mutation burns **two**. The gated tests were using `registerAndJoinSession`, which never creates the `board_sessions` row, so `updateQueueStateImmediate`'s insert failed its FK and the hash was never seeded. `releaseFrom(0)` had been papering over it; pinning the order turned it into a hang. They now use `createDurableSession`, and the helper documents the invariant.

**Issue 3 — missing interleavings.** Added for all three. They are not redundant with the `addQueueItem`/`removeQueueItem` cases, for the reason you implied: each plumbs the retried result back to its published event by a different route — `reorderQueueItem` through mutable `result*` locals, `mirrorCurrentClimb` through a spread return, `navigateQueue` through an `{ item, update }` pair. A retry that recomputes the queue correctly but publishes the *first* attempt's sequence still corrupts the client's gap check, so each test asserts the published sequence equals final stored state, not just that no climb survived. `navigateQueue` needed a real `esp32_controllers` row, since `requireControllerAuthorizedForSession` re-reads it.

## Verified non-vacuous

A fixture fix that makes tests pass for the wrong reason was the live risk here, so each guard was deleted to confirm the tests notice:

| Guard deleted | Result |
| --- | --- |
| conflict check in the mock's CAS transcription | 9 fail — all five interleavings, both version guards, the load test |
| conflict check in the real Lua script | 2 Layer-3 tests fail |
| `setWhere` on the Postgres upsert | its test fails (durable row rewinds 9 → 4) |
| just `resultSequence = result.sequence` in `reorderQueueItem` | the reorder interleaving fails, **and nothing else** |
| *(nothing — headline guard run against `origin/main`)* | fails with `resolved "{ version: 1, sequence: 2 }"` |

That last row is the production `sequence > version` fingerprint reproduced as a unit test.

Separately: the PR description was wrong — it had another PR's body pasted into it. Rewritten to describe this change.

marcodejongh:
Round 2 addressed.

**1. `releaseFrom` magic number.** Named it `GATE_LOOKAHEAD` and documented why a finite pre-resolve is needed at all: the `releases` array is sparse, so "every later one" has to mean "slots that don't exist yet". The bound only has to cover the CAS calls a single test can still issue after the release point — one per parked mutation plus its retries, and `MAX_RETRIES` is 3 — so 12 is a wide margin over the 2-3 these tests actually reach. Worth noting the failure mode is safe in either direction: exceeding it hangs the test rather than passing it wrongly.

**2. Dormancy-floor second attempt returning CONFLICT.** Real gap, now covered. The new case parks the reviving mutation at its *second* CAS (the floored write), seeds the Redis hash underneath it to stand in for the third process, then releases. Asserts the intruder's climb survives, the reviving mutation's climb is re-applied on top, and the retry builds on the seeded version rather than the stale floor.

Verified non-vacuous: relaxing that second attempt to `CAS_ANY_VERSION` fails it (and the load test), while the rest stay green.

**3. `addQueueItem` early exit.** Confirmed correct, and it was already covered — I checked rather than assumed. `queue-sync-fixes.test.ts` pins both halves of the contract for the already-in-queue case: no event is published (`expect(publishSpy).not.toHaveBeenCalled()`), and the resolver still returns the item ("should return the item even when it already exists (idempotent)"). So the observable behaviour is identical to the old `return item`; only the control flow moved. The code after the retry block is guarded by `if (itemWasAdded)`, which is exactly what the first of those tests asserts.

marcodejongh:
Round 3 — all three applied.

**1. `reorderQueueItem` mutable locals.** You're right that the inner `if` was unreachable-false: `QueueIndexSchema` rejects negatives before the callback runs, and the outer throw covers the upper bound. Removed the branch *and* the `currentState` pre-seed, so the write is unconditional and the result locals only ever hold post-write values — the failure mode you described is now structurally impossible rather than just unreachable. The reorder interleaving test already pins this (it asserts the published sequence equals final stored state), and it stays green.

**2. "Single round trip" comment.** Corrected. It was true of `casWithDormancyFloor` in isolation and false of a mutation end to end, which is the reading that matters for anyone auditing latency. The comment now says so explicitly: `withQueueVersionRetry` reads first, so a mutation is still two Redis round trips — the same count as the read-modify-write this replaces. The win is atomicity, not latency.

**3. Mock TTL index gap.** Commented. `args[8]` is the session TTL, which the mock has no use for since nothing in it expires. Verified against `casUpdateQueueState`'s actual `eval` ARGV order rather than assumed, and the comment flags that these indices have to stay aligned if that signature changes.

PR #3935, opened 2026-07-25
feat(mobile): one-tap link from a PR into its OTA preview

marcodejongh:
## Summary

Every PR with mobile changes already publishes a `pr-<n>` OTA channel and posts a sticky comment — but the comment only gave directions: *What's New → Try a preview → find the row*. Now it carries a link.

The comment links **`https://www.boardsesh.com/preview/pr-<n>`**. It has to be https: GitHub's markdown sanitiser renders only `http`/`https` anchors, so a `com.boardsesh.app://` href would come out as inert text.

| Where it's tapped | What happens |
| --- | --- |
| iOS, app installed | Opens the app straight onto the channel. The AASA is already wildcard (`'/*'`), so `/preview/*` needed no native change. |
| Android, or no app | Lands on a web page offering `com.boardsesh.app:///preview/pr-<n>` as a button. |
| Desktop | Same page; scan the QR to continue on a phone. |

**In the app** — `packages/mobile/app/preview/[channel].tsx` renders the existing `ChannelSwitcherScreen` with a `requestedChannel`. The screen offers it through the **same confirm dialog** a tapped row raises, so a link never switches the app on its own. It waits for the stored-override read *and* the preview list before offering, so the dialog names the PR and `performChannelSwitch`'s revert target is the channel that was actually live. On a Metro/dev build (`updatesUsable === false`) it prefills the manual field instead of raising a dialog that can't go anywhere.

Channel strings arriving from a URL are whitelisted (`^pr-\d+$` plus the tester presets) in `src/lib/preview-link.ts` before they reach `performChannelSwitch`. `+native-intent.ts` normalises the custom-scheme spellings — written with two slashes, `preview` is the URL *host*, which Expo Router's prefix stripping drops.

**Not included, deliberately:** Android universal links need a `/preview` `pathPrefix` in `app.config.ts`, which perturbs the native fingerprint. Bundling it here would make this PR a native change — `check:mobile-ota-compat` would return `native-change-required`, no `pr-<n>` preview would publish, and the feature couldn't be QA'd through the very flow it adds. It lands as a follow-up with the next store build; until then Android testers use the web page's button.

One duplication worth flagging: the channel grammar lives in two small files (`packages/mobile/src/lib/preview-link.ts`, `packages/web/app/lib/ota-preview-link.ts`) rather than a shared package, because the third consumer — the `github-script` step in the workflow — can't import TS at all. Both are commented as such and pinned by tests.

## Release Notes

- Open a preview build straight from its pull request — tap the link, confirm, and the app restarts on it.

## Test plan

- `vp run typecheck:mobile`, `vp run typecheck:web` — clean.
- `vp run test:mobile` — 4514 passed. New: `preview-link.test.ts` (channel whitelist, incl. `pr-12/../production` and `PR-12` rejections), 5 added cases in `native-intent.test.ts` (two-slash / three-slash / bare-path / throwing-`getShareExtensionKey` / near-miss `previews`).
- `vp run test:web` — 5804 passed. New: `ota-preview-link.test.ts`, `preview/[channel]/__tests__/page.test.tsx` (404s on `production`, `pr-`, `pr-12x`, `../etc`; noindex metadata), `preview-channel-content.test.tsx` (scheme href, QR, PR backlink).
- `vp run check:i18n`, `vp run check:i18n:orphans`, catalog-parity test — clean; `otaPreview.*` added to `en-US`, `es`, `fr`.
- `vp run check:mobile-bundle` — Metro bundle OK (14.3 MB) with the new route.
- `vp lint` / `vp fmt` on the touched files — 0 warnings, 0 errors.

Not exercised here: the live web page and the real device switch. No Docker in this environment, so `vp run dev` couldn't bring up the dev DB, and an actual channel switch needs a TestFlight/App Store binary (`updatesUsable` is false on Metro). Worth a manual pass on this PR's own `pr-<n>` link once CI publishes it.

https://claude.ai/code/session_01HQzz9qeHCMZQSGHTkAXcNf

---
_Generated by [Claude Code](https://claude.ai/code/session_01HQzz9qeHCMZQSGHTkAXcNf)_


marcodejongh:
All three addressed in `fcdbc0d`.

**1 — dead export.** Confirmed and deleted, along with its test case.

**2 — validator parity.** The divergence was wider than described: web's `/^pr-(\d+)$/` also *accepted* `pr-007` and resolved it to the same page as `pr-7`, so tightening only the mobile side would have left the halves disagreeing in the other direction. Both now use `^pr-[1-9]\d*$`, and web's `pullNumber > 0` guard is gone as redundant (the safe-integer overflow check stays). One PR, one URL, same answer on both sides.

**3 — untested guards.** A component test can't reach them, and the existing suite explains why: the mobile vitest config inlines `__DEV__: true` as a build-time `define`, not a global, so `updatesUsable = Updates.isEnabled && !__DEV__` is false in every rendered test. The `switch` branch is unreachable, and the prefill isn't observable either since the `manual` section it writes into is itself `updatesUsable`-gated.

So the guard ordering moved into a pure `resolveRequestedChannelAction` in `channel-switch.ts` — the same extraction that file already does for the switch/revert state machine — and `channel-switch.test.ts` covers all four outcomes, including the double-fire case the one-shot guard exists for and `'wait'` outranking `'prefill'` so a request isn't swallowed while state is still loading. The component suite pins what it *can* reach: a deep-linked channel raises no dialog on a build that couldn't act on one, and doesn't disturb the sections a plain `/channel-switcher` visit builds.

---

marcodejongh:
Checked both — declining both, evidence below. No code change this round.

**Issue 1 (French register) — the premise doesn't hold.**

`docs/i18n-french-glossary.md`, the document CLAUDE.md points at for French, contains **no register rule at all**: zero mentions of tu/vous/tutoiement. Its sections cover « enchaîné » vs « envoyer », « allumer », « la croix », « essais », terms to keep in English, and « la board » gender. So "informal is house style" isn't written down anywhere.

Nor is it true in the catalog, which is split by namespace (vous-family / tu-family word counts):

| vous-leaning | | tu-leaning | |
| --- | --- | --- | --- |
| `marketing` | 128 / 7 | `session` | 4 / 47 |
| `settings` | 66 / 9 | `kiosk` | 0 / 44 |
| `climbs` | 58 / 3 | `profile` | 0 / 15 |
| `auth` | 48 / 0 | `you` | 1 / 9 |

`common`, where these keys live, is 31 / 39 — near even.

What settles it is the immediate neighbours. The `mobile.previewChannels` block these keys extend is uniformly *vous*: "**Chargez** une demande de modifications…", "**Revenez** bientôt", "**Réessayez** dans un instant", "publiée pour **votre** version de l'app". So is the `mobile.changelog` screen hosting the entry point: "**Vous utilisez** la version du {{date}}". Switching these to *tu* would make them the only informal copy on the two screens they actually render on.

The repo-wide split is real and undocumented, but that's a catalog-wide sweep plus a glossary rule — its own task, not a rider on this PR.

**Issue 2 (prefill assertion) — not observable by construction.**

`customChannel` only reaches the view-model through the `manual` section, and that section sits inside `if (updatesUsable)`. Under test `updatesUsable` is always false — the mobile vitest config inlines `__DEV__: true` as a build-time `define` — and that false is precisely what selects the prefill branch. So on the only path where prefill happens, the section that would expose the value is never built. There's no assertion to write.

That's why it's asserted where the value is visible: `resolveRequestedChannelAction` returning `'prefill'` for exactly this input, in `channel-switch.test.ts`. Same conclusion the review reaches.

Marking ready for review — CI green on `fcdbc0d`, no conflicts.

---

marcodejongh:
Rebased onto `main` (V3 OTA migration, #3969) and addressed the Codex review — now at `9ee3e64b7`.

**Rebase.** Only two files conflicted; the rest of the branch was untouched by V3. Compatibility held up: `applyChannelOverride(channel)` kept its signature (V3 just adds the `expo-app-id` header inside it) and preview channels are still `pr-<n>`, so the deep link and the `^pr-[1-9]\d*$` whitelist carried over unchanged.

The workflow conflict resolved in V3's favour and came out better than what was here before: the link is now gated on **`switchable`** (published **and** the `map` job succeeded) rather than `anyPublished`. A bundle that publishes but fails channel↔branch mapping no longer advertises a link that would land testers on "No branch mapping found".

**Preview request lost across sign-in — fixed.** Confirmed in `auth-provider.tsx`: `!isAuthenticated && !inAuthGroup → <Redirect href="/auth/login" />`, so `PreviewChannelRoute` never mounted and the channel was dropped; after login the user landed on home. Now stashed under `boardsesh_pending_preview_channel` and replayed once auth flips, mirroring the join flow's `PENDING_JOIN_KEY`. Re-validated on the way *out* as well as in, since the stash outlives the launch that wrote it.

Worth noting on the parser: I first put it in `deep-link-provider.tsx` alongside the join one, but that file can't be imported in a node test env (`expo-linking` pulls in react-native), so the link shapes would have gone untested. Moved it to `preview-link.ts` hand-rolled, following the precedent `deep-link-query.ts` already sets for exactly this. All shapes are now pinned — both scheme spellings, locale prefix, query tail, and the whitelist still applying to the URL segment. I also reverted an incidental refactor of `parseJoinSessionId` rather than restructure a working, untested join parser as a side effect.

**One-shot guard keyed by channel — fixed.** `offeredChannelRef` holds the channel instead of a boolean, so a second `/preview/pr-M` on a still-mounted screen is still offered. Covered by a new case in `channel-switch.test.ts`.

**Link announced before the target is live — accepted, not changing.** Real, and it's a deliberate call: on merge the production OTA carries the route and the web page deploys together, so the exposure is users who haven't relaunched since, and it self-heals on next launch. Keeping the wildcard AASA preserves the tap-straight-into-the-app behaviour that is the point of the feature. Gating on `switchable` narrows the window further.

**Heads-up, unrelated to this PR:** `origin/main` currently fails `vp run typecheck:mobile` on its own — I verified this in a clean worktree at `f447680de`:

```
src/hooks/use-image-cache-memory-management.ts:79 - error TS2339: Property 'isPad' does not exist on type 'DeviceLayout & { isTablet: boolean; wallDeviceClass: WallDeviceClass; }'
src/providers/board-art-visibility-provider.tsx:25 - error TS2339: (same)
```

Looks like the Android-tablet-shell merge renamed `isPad` → `isTablet` without updating these two call sites. I've left it alone: `isTablet` now covers Android tablets, so swapping it into an iPad-specific memory sweep would change behaviour on Android, which is a call for whoever owns that change.

Local runs on this branch: mobile 4662 tests, web 5835 tests, Metro bundle OK, lint/i18n clean.

---

marcodejongh:
`3d160e433` — Android now opens the preview link directly, plus the latest review round.

**⚠️ This PR is now a native change.** Added the `autoVerify` `/preview` intent filter to `app.config.ts`, so an Android tester tapping a PR's preview link lands in the app instead of on the web page's "Open in Boardsesh" button. iOS got that free from the wildcard AASA; Android intent filters are path-scoped, so the prefix has to be declared.

The consequence is deliberate but worth stating plainly: it perturbs the fingerprint, so **`pr-3935` will stop publishing an OTA preview for this PR** — the compat check will flip to `native-change-required`, and the filter only goes live with the build that ships it. Test this one from a TestFlight/Play build rather than the in-app switcher.

**Review items:**

1. **Locale stripping** — good catch, and the fix is better than a comment. It now matches against `SUPPORTED_LOCALES` (minus the default, which lives at the root) rather than any two-lowercase-letter segment, so a future `/go/…` route can't be silently eaten and a regional locale like `pt-BR` works the day it's added. Test added for exactly the `/go/preview/pr-1234` case.

2. **QR colour** — switched to `common.white`. The mode-invariance is the point (a QR needs a light quiet zone in both themes), and the token says that without a literal.

3. **Deep-link stash/replay untested** — fair, and it was the actual auth-survival fix, so it needed covering. Added `providers/__tests__/deep-link-provider.test.tsx`: routes straight through when signed in, stashes instead of navigating when signed out, replays and clears on the auth flip, drops a stored value that no longer validates, handles a warm link, and ignores non-preview links.

**Not changing:** the hardcoded `REPO_URL`. `privacy/page.tsx` and `aurora-migration-content.tsx` already inline the same URL, so there's no shared constant to reach for — introducing one is a separate cleanup across all three, not a rider here. The repo also can't be renamed without breaking far more than this backlink.

**Earlier heads-up resolved:** the `isPad` → `isTablet` typecheck breakage I flagged on `main` is gone — the rebase picked up the fix, and `vp run typecheck:mobile` is clean on this branch now.

Local: mobile 4674 tests, web 5835 tests, typecheck clean both, Metro bundle OK, lint clean in touched files.

---

PR #3941, opened 2026-07-26
fix(mobile): stop a consumer clip swallowing GlassSurface's Android elevation cast

marcodejongh:
## Summary

Closes #3058 (replaces it — see "Why a new PR" below).

`GlassSurface`'s Material branch composed the consumer's `style` and the elevation
cast onto **one and the same view**:

```tsx
<View style={[style, radius, elevationStyle, { backgroundColor: materialBg }]}>
```

Android draws the cast from that view's own outline, so any consumer passing
`overflow: 'hidden'` clipped its own shadow down to nothing. The climb long-press
reaction menu does exactly that, and it has to: the clip is what keeps the
scrolling action list, its bottom fade and the playlist picker's pinned header
inside the card's rounded corners. So on Android the card lost its cast and melted
into the dim behind it.

**The fix is in the shared primitive, not at the one consumer.** The cast stays on
the unclipped outer view and the clip is hoisted onto an inner wrapper around
`children` — elevation on an unclipped ancestor, clip on a descendant. The wrapper
is added *only* when a clip was actually asked for, so the 14 `GlassSurface` call
sites that never clipped render the exact same tree as before. Consumers keep
writing `overflow: 'hidden'` as normal; they no longer have to choose between a
rounded clip and a shadow.

With the cast surviving, the reaction menu gets the treatment it should have had:

- **M3 dialog tone + cast.** `role="base" level="level2"` becomes `role="high" level="level3"`.
  A modal, scrim-dimming, non-anchored overlay is an M3 dialog, so it takes
  surface-container-high and a level-3 cast.
- **The list's bottom fade is derived from that same tone.** It used to be a
  hardcoded `rgba(20, 17, 31, ...)` with a "keep them in sync" comment, which the
  role bump would have silently broken. On Material it now reads
  `m3SurfaceContainers[MENU_CARD_ROLE]` directly, so the two can't drift. Glass /
  blur / solid keep the tuned constants — there is no opaque tone to read there,
  the card is real glass.
- **Scrim and blur gate on the surface mode, not `Platform.OS`.** With no blur
  underneath, the old light tint let the colourful board-art grid bleed through and
  the overlay stopped reading as focused; the scrim now does the receding itself
  (dark 0.5 to `overlays.scrim` 0.6, light 0.35 to 0.4).
- **Dividerless rows, icons 22 to 24dp**, matching the primary button row above them.
- **The tap-to-dismiss scrim announced the climb's name.** VoiceOver/TalkBack read
  "Big Move, button" for a control that closes the overlay. Now a real close string
  (`mobile.climbActions.closeOverlay`, added to en-US / es / fr).

### This is an iOS change too

#3058's body claimed "iOS is unchanged". That was wrong and I'm not repeating it.
Swapping `Platform.OS === 'ios'` for the effective surface mode means **iOS users
with Reduce Transparency on, and iOS users on the Material variant, no longer get a
`<BlurView>`** — they get the denser scrim instead. That is the correct behaviour in
both cases (Reduce Transparency exists precisely to not blur, and Material-on-iOS
should match the Material card `GlassSurface` is already painting), but it *is* a
behaviour change and it is asserted in tests rather than assumed.

### Why a new PR instead of re-pointing #3058

GitHub can't move a PR's head branch, so "re-pointing" would mean force-pushing over
`fix/android-climb-reaction-menu` — which discards the original commits and orphans
every existing review thread against line numbers that no longer exist. The diff is
also substantially different: #3058 fixed this by **deleting** `overflow: 'hidden'`
from `menuCard`, which is no longer safe. `main` has since added a `menuScrollFade`
`LinearGradient` (absolutely positioned, near-opaque end stop) and the
`InlinePlaylistPicker` view *inside* the card; both need that clip or they paint
square corners over the card's rounded bottom edge. This PR **keeps** the clip and
makes `GlassSurface` cope with it.

## Release Notes

- Long-press a climb on Android and the action menu now floats above the wall properly instead of melting into the background
- Screen readers announce the tap-anywhere-to-close area as a close button, not as the climb

## Test plan

Automated (all green locally):

- `vp run typecheck:mobile`
- `vp run test:mobile` — 4512 tests / 544 files pass, including 9 new ones
- `vp check` — 0 errors
- `vp run check:i18n`, `vp run check:i18n:orphans`, and the `@boardsesh/i18n` catalog-parity suite

New coverage (the old suite mocked `GlassSurface` down to a bare `div` and `BlurView`
to `null`, so every single thing this PR changes was invisible to it):

- `glass-surface.test.tsx` — with a consumer clip, the elevated view is `overflow: visible` and the clip lands on a descendant that still wraps the children; with no consumer clip, no wrapper is added at all. **Mutation-checked**: reverting the `GlassSurface` change makes the first of these fail.
- `ClimbReactionMenu.test.tsx` — the card asks for `role="high" level="level3"` and keeps its clip; the scrim is `overlays.scrim` with no `BlurView` on `material`/`solid`, and the lighter tint with a `BlurView` on `glass`/`blur`; the bottom fade is built from the same tone the card is painted with on Material and keeps the tuned constants on glass; the dismiss scrim carries a close label, not the climb name.

Regression-checked every `GlassSurface` call site for the new inner view. Only three
pass children at all — `ClimbReactionMenu` `menuCard`, `zone.tsx` `modeIsland`,
`OnboardingPrompt` `footer` — and only `menuCard` passes a clipping `overflow`, so it
is the only one that gets the wrapper. The other twelve (`play.tsx`,
`record/index.tsx`, `CollapsingTopChrome`, `GlassActionToolbar`, `GlassIconButton`,
`GradeRail`, `IpadSidebar`, `PlaylistEditDoneButton`, `PlaylistFollowButton`,
`AccessoryBarSurface` x2, `SearchHeader`, `SessionStartFab`) are self-closing
`style={StyleSheet.absoluteFill}` background layers with no children and no
`overflow`, so their tree is unchanged. `GlassSurface.web.tsx` is left alone
deliberately: a CSS `box-shadow` is not clipped by the element's own
`overflow: hidden`, so the bug doesn't exist there and a wrapper would be dead weight.

### Outstanding: on-device Android visual QA

**I have not visually verified this.** This box is Linux, so
`vp run check:mobile-simulator` and `vp run mobile:screenshot` skip, and I am not
going to claim a visual result I didn't get. The pixels still need a human on an
Android device or emulator, in **both** dark and light (#3058 never captured light):

1. The card's elevation cast is actually visible against the denser scrim.
2. With enough actions to make the list scroll, the bottom `LinearGradient` is
   clipped to the card's rounded corners and its tone matches the card.
3. The `InlinePlaylistPicker` view inside the card still renders rounded.

Plus one iOS pass with Reduce Transparency ON and one on the Material variant, since
those two paths lose their `BlurView` here.

Note for whoever picks this up: #3058's `pr-3058` OTA channel is stale — it was
published against a five-week-old base. This PR needs its own `/ota-preview`.

### DIFF
--- a/packages/mobile/src/components/GlassSurface.tsx
+++ b/packages/mobile/src/components/GlassSurface.tsx
@@ -200,11 +200,20 @@ const styles = StyleSheet.create({
     overflow: 'visible',
   },
   /**
-   * The hoisted clip. Sized to disappear from layout: `stretch` fills the cross
-   * axis, `flexGrow` takes any definite height the surface was given, and
-   * `flexShrink` keeps the clip bounds on the box when content overruns — so the
-   * children lay out exactly as they did when they were direct descendants. It
-   * clips at the surface's PADDING box, so a surface that both pads and clips
+   * The hoisted clip, sized so the children lay out the way they did when they
+   * were direct descendants: `stretch` fills the cross axis, `flexGrow` absorbs
+   * any definite height the surface was given, and `flexShrink` keeps the clip
+   * bounds on the box when content overruns.
+   *
+   * `flexBasis` deliberately stays `auto` — this is NOT `flex: 1`. With
+   * `flexBasis: 0` an auto-height surface would measure its only in-flow child at
+   * zero, size itself to zero, and then have no free space to grow back into, so
+   * the whole surface would collapse (the reaction menu card is exactly that
+   * shape). `auto` starts from the content size, which is what an auto-height
+   * surface wants, and still leaves free space for `flexGrow` to take when the
+   * surface does have a definite height.
+   *
+   * It clips at the surface's PADDING box, so a surface that both pads and clips
    * would round its corners inside the padding; no consumer does both today.
    */
   clipLayer: {
--- a/packages/mobile/src/components/GlassSurface.tsx
+++ b/packages/mobile/src/components/GlassSurface.tsx
@@ -55,8 +55,9 @@ type GlassSurfaceProps = {
   role?: keyof MaterialSurfaceContainers;
   /**
    * Material-only: the M3 elevation CAST (level0–5). Omit to keep the legacy
-   * `shadows.sm`. The material branch applies it on the same view as the
-   * background + radius (no `overflow:'hidden'`), so a rounded surface still casts.
+   * `shadows.sm`. The cast always lands on the outermost view; if the consumer's
+   * `style` asks for `overflow: 'hidden'`, that clip is hoisted onto an inner
+   * wrapper so it can't swallow the cast (see the material branch).
    */
   level?: MaterialElevationLevel;
 };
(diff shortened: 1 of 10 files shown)


marcodejongh:
Thanks — taking two of the three, pushing back on one.

**`clipLayer` and `flexBasis: 0` — declining, it would be a regression.** `flexBasis` staying `auto` is load-bearing, not an oversight. With `flexBasis: 0` an auto-height surface measures its only in-flow child at zero, sizes itself to zero, and then has no free space left to grow back into, so the surface collapses entirely. The reaction menu card is exactly that shape (`menuWrap` gives it a width; its height comes from content), so `flex: 1` here would render a 0px-tall card. `auto` starts from the content size — right for an auto-height surface — and still leaves free space for `flexGrow: 1` to absorb when the surface *does* have a definite height, which is the case the review is worried about. I've rewritten the comment to spell this out so the next reader doesn't have to re-derive it.

**`m3SurfaceContainers` reference stability — verified, it's stable.** `theme-provider.tsx:353` is `m3SurfaceContainers: materialSurfaceContainers[colorScheme]`, and `materialSurfaceContainers` is a module-level `as const` object, so indexing it returns the same reference every render. The `menuFadeColors` memo only recomputes on a scheme change. No change needed.

**Android/iOS visual QA — agreed, and it's deliberately called out as an open gate in the PR body rather than papered over.** This box is Linux, so `check:mobile-simulator` and `mobile:screenshot` skip. `pr-3941` is published to OTA for both platforms, so a device pass can be done without a build.

Also swept a matching stale doc comment in `theme-provider.tsx` that still told callers the cast must never sit under `overflow: 'hidden'`.

Issue #3928, opened 2026-07-25
expo-image: UIImage(named:) on the main thread drives _UIImageCollectImagePathsForPath on every ImageView.reload


marcodejongh:
Found while diagnosing the #3803 iPad crash. Independent of that OOM — filing separately so it isn't folded in and lost.

## What the crash report showed

The main thread in the #3803 Apple crash report was here:

```
5  Foundation   NSAllocateObject + 168
6  Foundation   +[NSPathStore2 pathStoreWithCharacters:length:]
7  Foundation   -[NSString(NSPathUtilities) stringByAppendingPathComponent:]
8  UIKitCore    ___UIImageCollectImagePathsForPath_block_invoke_2 (UIImageTable.m:496)
9  UIKitCore    ___UIImageCollectImagePathsForPath_block_invoke     (UIImageTable.m:568)
10 UIKitCore    _UIImageCollectImagePathsForPath                    (UIImageTable.m:605)
11 UIKitCore    +[_UIAssetManager newAssetNamed:fromBundle:]        (_UIAssetManager.m:1089)
12 UIKitCore    +[UIImage imageNamed:inBundle:variableValue:withConfiguration:]
13 ExpoImage    UIImage.__allocating_init(named:)
14 ExpoImage    ImageView.localAssetImage(from:)   (ImageView.swift:367)
15 ExpoImage    ImageView.maybeRenderLocalAsset(from:) (ImageView.swift:355)
16 ExpoImage    ImageView.reload(force:)           (ImageView.swift:197)
```

## The problem

`ImageView.reload(force:)` calls `maybeRenderLocalAsset` → `localAssetImage(from:)` → `UIImage(named:)` **synchronously on the main thread**.

`UIImage(named:)` for a name that isn't already in the asset cache falls through to `_UIAssetManager newAssetNamed:fromBundle:` → `_UIImageCollectImagePathsForPath`, which **enumerates bundle directories and builds an `NSString` path per candidate file**. That's filesystem I/O plus a burst of short-lived `NSPathStore2` allocations, on the main thread, on every `reload`.

Two distinct costs:

1. **Main-thread stall.** Bundle directory enumeration is not something that belongs on the UI thread, and it runs per `reload` — not once per app launch.
2. **Allocation churn.** Each pass allocates a string per path examined. In the #3803 report this path is literally where the main thread was when allocation failed.

[...]

PR #3944, opened 2026-07-26
fix(mobile): cap expo-image's unbounded decoded-bitmap cache

marcodejongh:
## Summary

The app has been running an **unbounded** decoded-bitmap cache. SDWebImage ships `maxMemoryCost = 0` (unlimited), expo-image never overrides it, and `grep -rn "configureCache" packages/` returned zero hits. So every distinct board-art bitmap a session decoded stayed resident until the app backgrounded or the OS fired a memory warning — and on iOS that warning is delivered only *after* an allocation has already failed. That is the #3803 iPad crash (`EXC_BAD_ACCESS` inside image decode, 3d16h uptime) and the same shape as the 4 GB-iPhone foreground OOM in #3479.

**Set the ceiling.** `Image.configureCache({ maxMemoryCost: 256 MB })` once at startup. This is pure JS (`expo-image/src/Image.tsx` → `ios/ImageModule.swift:340-350` → `SDImageCache.shared.config`), so it ships over OTA — no native fingerprint change. iOS-gated: expo-image's Android module exposes no `configureCache` and the call throws; Android already has its own device-RAM-derived Glide policy plus the native `modules/memory-trim` full-clear.

256 MB is ~32 full-resolution overlays. The widest working set of any surface is three (the carousel's previous + current + peek), so the cap cannot evict art a live surface is about to swap to.

**Stop the wall kiosk decoding at native width.** `WallHeroStage` swaps climbs on every live-wall change and every scrub step, and both `LayeredClimbImage` layers set `allowDownscaling={false}` — so each climb decoded ~8 MB of RGBA no matter how small it was drawn. Now passes `renderWidth` (display px × `PixelRatio.get()`) with `backgroundVariant="full"`, matching `SwipeBoardCarousel`: the shared board photo stays crisp, only the per-climb overlay shrinks. Clamped to native width upstream, so a large kiosk already drawing above native resolution is unchanged.

### Why a ceiling and not a periodic sweep

A cap dominates a timer on every axis. NSCache enforces it continuously — no interval to tune, no navigation seam to wait for, and nothing to fire on a surface that never navigates. Critically, **eviction drops only the cache's reference**: a bitmap a mounted view still displays is retained by `UIImageView.image` and `CALayer.contents`, and nothing in expo-image observes a cache clear to trigger a reload (`grep` for `NotificationCenter|addObserver|memoryWarning` over `node_modules/expo-image/ios/` returns zero). So the cap can never blank on-screen art, and it has no equivalent of the peek→commit window a periodic wipe would open in the carousel.

### Deliberately not done

- **No `recyclingKey` on the kiosk hero.** `docs/react-native-performance.md` §7 nominally mandates it for always-mounted surfaces that swap climbs, but a key change sets `sdImageView.image = nil` *before* the new load (`ImageView.swift:66-73`). On a wall-mounted display that inserts a blank frame between climbs, for a bounded 2-bitmap saving. The carousel only gets away with it because its peek pre-warms the cache. Documented as an explicit exception rather than left as an apparent oversight.
- **#3866 (expo-image 57.0.0 → 57.0.1) untouched.** Native fingerprint change; it must ride the release train.
- **#3867 untouched.** Both its levers are sound and still do work this cap can't (backgrounding frees the whole working set, not just the excess).

### What this does not prove

`Refs`, not `Closes`. The crash proves memory exhaustion but not *which* pool. Read as DenseMap bucket-count/mask, `x2 = 0x400000` / `x3 = 0x3fffff` on the crashing thread implies millions of live objects carrying associated objects — and board art at ~8 MB apiece cannot produce millions of objects. So a second, object-count leak may be in play that a bitmap-cache bound does not fix. That is a hypothesis on unsymbolicated frames, not a certainty, but it is reason enough to keep #3803 open and watch.

Scope is also still unsized: PostHog `$exception` is blind on 2.1.0+ and Sentry reads need a `sntryu_` user token nobody has wired up here. The outcome oracle — iPad crash-free rate for SIGSEGV with an ExpoImage/SDWebImage culprit, before vs. after — stays blocked on that token.

Filed separately: sharpened #3928 with the finding that `localAssetName` accepts `scheme == "file"` (`ImageView.swift:855-866`), so **every** `file://` board-art source pays a guaranteed-to-miss main-thread `UIImage(named:)` bundle scan on every reload — that is thread 0 in this crash report, and it is an allocation-churn source independent of the OOM.

## Release Notes

- iPad and iPhone: board art no longer piles up in memory without limit, so a long day on the wall won't end in a crash.
- Wall displays: the kiosk board now renders each climb at the size it's actually shown instead of full resolution.

## Test plan

- `vp check` — exit 0, no findings in changed files.
- `vp run typecheck:mobile` — pass.
- `vp run test:mobile` — 545 files, 4511 tests, all pass (9 new).
- **Mutation-tested every new guard** — each deletion fails exactly the intended assertion, no tautologies:
  | Mutation | Fails |
  | --- | --- |
  | drop the `Platform.OS !== 'ios'` gate | `does not configure the cache on Android` |
  | `maxMemoryCost` → `maxDiskSize` | `caps the decoded-bitmap cache on iOS` |
  | drop `backgroundVariant="full"` | `keeps the board photo full-resolution` |
  | drop `renderWidth` | `sizes the holds overlay to the drawn surface`, `scales the request by display density` |
  | drop the `artWidth > 0` guard | `falls back to the native-width render before the surface is measured` |

### Outstanding gate — not done, not claimed

No Mac and no iPad in this environment, so **there is no on-device verification here.** The device check that would settle it: publish this PR's JS to its `pr-<N>` OTA channel, park an iPad on "On the Wall" with a board bound, and watch the Xcode memory gauge for ~30 min while climbs cycle — the image-cache contribution should plateau near the cap instead of climbing. Worth pairing with an iPhone pass on the climbs list and the play-drawer carousel to confirm no thumbnail pop-in or swipe-commit flash from cache pressure at 256 MB.

Refs #3803. Refs #3479.

### DIFF
--- a/docs/react-native-performance.md
+++ b/docs/react-native-performance.md
@@ -274,7 +274,8 @@ uptime (#3803). Four contracts keep them in check:
   expo-image does not override it, so an app that never calls `Image.configureCache()` runs an
   unbounded image memory cache and only ever reclaims at a background or `memoryWarning` seam.
   `useImageCacheMemoryManagement` sets `maxMemoryCost` (`IMAGE_MEMORY_CACHE_MAX_BYTES`, 256 MB) once
-  at startup. iOS-only: expo-image's Android module exposes no `configureCache` and the call throws.
+  at startup. iOS-only: expo-image declares `configureCache` `@platform ios` and defines no Android
+  counterpart, so calling it off iOS fails at the bridge rather than quietly no-opping.
   Prefer this over any periodic wipe — NSCache enforces it continuously, and because eviction drops
   only the _cache's_ reference (a bitmap a mounted view displays is retained by
   `UIImageView.image`/`CALayer.contents`), it can never blank on-screen art or force a reload.
--- a/docs/react-native-performance.md
+++ b/docs/react-native-performance.md
@@ -263,24 +263,43 @@ surface the previous size's bundled background paths.
 
 ---
 
-## 7. Board-art image memory: render at display size, recycle, release on background
+## 7. Board-art image memory: cap the cache, render at display size, recycle, release on background
 
 **Rule:** Board-art surfaces are the app's largest memory consumer (decoded bitmaps live in native
 heap + as GPU textures, not the JS heap) — on iOS a 4 GB device can hit the OS watchdog OOM kill in
-the foreground while browsing (#3479). Three contracts keep them in check:
+the foreground while browsing (#3479), and an iPad can SIGSEGV inside image decode after days of
+uptime (#3803). Four contracts keep them in check:
 
-- **Size the overlay to the display, keep the photo full-res.** The play-drawer board passes
-  `renderWidth` (display px × `PixelRatio.get()`) so the per-climb holds overlay rasterizes at the
-  shown size instead of the board's native ~1080px, plus `backgroundVariant="full"` so the _shared_
-  board photo (one decode per board-config, reused across every climb) stays crisp. `renderWidth`
-  alone would downgrade the photo to the 416px `thumb` — only the overlay should shrink.
+- **Cap the decoded-bitmap cache.** SDWebImage defaults `maxMemoryCost` to `0` — _unlimited_ — and
+  expo-image does not override it, so an app that never calls `Image.configureCache()` runs an
+  unbounded image memory cache and only ever reclaims at a background or `memoryWarning` seam.
+  `useImageCacheMemoryManagement` sets `maxMemoryCost` (`IMAGE_MEMORY_CACHE_MAX_BYTES`, 256 MB) once
+  at startup. iOS-only: expo-image's Android module exposes no `configureCache` and the call throws.
+  Prefer this over any periodic wipe — NSCache enforces it continuously, and because eviction drops
+  only the _cache's_ reference (a bitmap a mounted view displays is retained by
+  `UIImageView.image`/`CALayer.contents`), it can never blank on-screen art or force a reload.
+- **Size the overlay to the display, keep the photo full-res.** The play-drawer board and the wall
+  kiosk hero pass `renderWidth` (display px × `PixelRatio.get()`) so the per-climb holds overlay
+  rasterizes at the shown size instead of the board's native ~1080px, plus `backgroundVariant="full"`
+  so the _shared_ board photo (one decode per board-config, reused across every climb) stays crisp.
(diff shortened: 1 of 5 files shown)


marcodejongh:
Thanks — took the first four nits (pushed in `49a61b7`):

- **"throws on Android"** was an inference, and you were right to flag it. expo-image declares `configureCache` `@platform ios` and defines no Android counterpart, so the call fails at the bridge rather than as an intentional throw. Reworded in both the code comment and §7.
- **web gate** now covered — the test is `it.each(['android', 'web'])`, since the guard is `!== 'ios'`.
- **`useMemo` divergence** — kept it inline, but documented why: the carousel's equivalent derives from a `boardBox` *object*, so `useMemo` is what keeps its identity stable. This derives from the `artWidth` number and feeds a number prop, which compares equal by value in the memo'd child's shallow compare. A `useMemo` here would add a dependency array guarding nothing.
- **assertion band** — agreed it's wide; leaving it, since the lower bound (8 full-res overlays) is the one that carries meaning and the upper is only a "still bounds something" backstop.

## Declining the comment-verbosity nit

The second-pass review cites CLAUDE.md as saying comments are **"one short line max."** That rule does not exist. There is exactly one `CLAUDE.md` in this repo (`find . -name CLAUDE.md -not -path "*/node_modules/*"` → a single result) and it contains no comment-length rule of any kind.

The repo norm is the opposite, in this exact subsystem, in already-merged code:

| File | Comment lines / total |
| --- | --- |
| `board-art-visibility-context.ts` | 13 / 20 |
| `board-art-visibility-provider.tsx` | 21 / 35 |
| `route-segments.ts` | 55 / 89 |
| `LayeredClimbImage.tsx` | 72 / 192 |

Trimming my two blocks to one-liners would make them the outliers in the files they sit in.

On "belongs in the PR description, not source": that is exactly backwards for these two. The PR gets closed and never read again; the next person to touch `WallHeroStage` needs to know *at the call site* why it deliberately has no `recyclingKey` — otherwise they'll add one to satisfy the §7 rule and flash a wall-mounted display between climbs. Same for the `maxMemoryCost = 0` default: that is the non-obvious fact that makes the whole change necessary, and it is invisible from the code alone.

Marking ready for review.

Issue #3949, opened 2026-07-26
Kilter Grips serves its full climb catalog unauthenticated — decision needed on disclosure


marcodejongh:
## Summary — needs @marcodejongh's decision, no action taken

While sourcing real `climb_concat` samples for #3523 I found that Kilter Grips' catalog endpoint responds **without any authentication**:

```
GET https://portal.kiltergrips.com/api/climbs/all/27
→ HTTP 200, application/json, ~52 MB, 83,991 climbs
```

Every field the authenticated path returns is present: `climbUuid`, `climbConcat`, `name`, `description`, `username`, `isDraft`, `isListed`, `isDeleted`, `createdAt`, edges. Also confirmed on layout `17` (16,522 climbs).

The sibling endpoint on the same host **does** enforce auth, which is what makes this look like an oversight rather than a deliberately public API:

```
GET https://portal.kiltergrips.com/api/climbs/delteduuids
→ HTTP 401
```

## What this changes for us

- `packages/kilter-sync` currently sends a Keycloak bearer on the catalog leg (`catalogGet` in `src/api/kilter-rest.ts`). It doesn't strictly need one. **We should not act on that** — keep sending the token; relying on someone else's missing auth check is not a foundation for a sync path, and it would break the day they fix it.
- It is how the #3523 encoding was decoded, with no credentials and no production database access.

## The decision

Third-party disclosure is not ours to make unilaterally. Options:

1. Report it to Kilter / Northtech privately.
2. Say nothing; it is their public endpoint to configure.
3. Ask them about it as part of any broader API-access conversation.

Filing this as the record. **No contact has been made and none should be until Marco decides.**

## Notes

Only catalog data was read, and only the two layouts needed for the #3523 fixtures. No user data, no authenticated endpoints, no writes. The committed fixture (`packages/kilter-sync/src/sync/__fixtures__/grips-multiframe.json`) contains six climbs — exactly what the tests assert on.

marcodejongh:
I think this was deliberate, closing issue as the endpoint also isnt very high impact if it was deliberate

Issue #3948, opened 2026-07-26
convertLitUpHoldsStringToMap emits a phantom hold_id 0 for differential frames strings


marcodejongh:
## Summary

`convertLitUpHoldsStringToMap` (`packages/board-constants/src/hold-states.ts:265`) parses a frames string by splitting on `p` and then on `r`. It has no handling for the two other tokens the Aurora multi-frame format uses — `x{placementId}` clears and the leading `"` frame prefix.

For a differential frame like `"x1192p1369r13`, `frameString.split('p')` yields `['"x1192', '1369r13']`. The first element parses to `NaN`, misses `HOLD_STATE_MAP`, and falls into the unknown-code branch, which returns `[holdId || 0, { state: '${holdId}=${stateCode}' }]` — i.e. **hold id `0`** with a sentinel state. Every frame after the first carries the `"` prefix, so this fires on essentially every frame of every multi-frame climb passed through it.

## Where it matters

- `packages/aurora-sync/src/sync/shared-sync.ts:683` flattens `convertLitUpHoldsStringToMap(item.frames, board)` straight into `board_climb_holds` rows with **no filtering**, so a multi-frame climb through that path writes phantom `hold_id = 0` rows.
- `packages/db/scripts/backfill-hold-fingerprints.ts` then fingerprints from `board_climb_holds`, so a phantom row would silently poison the dedup key.

`packages/db/scripts/backfill-board-climb-holds.ts` does guard against it (`if (!hold.state || hold.state.includes('=')) continue`), and restricts itself to `frames_count = 1` — so today the dev DB has **0 rows** with `hold_id = 0` and the bug is latent rather than live. It is one multi-frame climb through the aurora-sync path away from becoming real.

## Suggested fix

Either teach `convertLitUpHoldsStringToMap` the `x` and `"` tokens (it already exists correctly in `tokeniseFrameDelta` a few lines above), or make it refuse differential input outright. Also apply the `state.includes('=')` guard at `shared-sync.ts:683` so the two hold writers agree.

Note that `packages/kilter-sync` no longer routes through this function — after #3523 it derives hold tuples from its own structured decode — so this is aurora-sync and the shared helper only.

## Found via

Decoding the Kilter Grips multi-frame encoding for #3523. Independent of that fix.

Issue #3947, opened 2026-07-26
Aurora absolute frames are read as deltas — 34 of 112 animated Kilter climbs play back wrong


marcodejongh:
## Summary

`splitFramesString` / `accumulateFramesToMaps` in `packages/board-constants/src/hold-states.ts` treat **every** frame of a multi-frame `board_climbs.frames` string as a delta. The Aurora catalog does not always emit deltas: a frame that is **not** prefixed with a literal `"` is an **absolute snapshot** — it restates the full lit set rather than adding to the previous frame.

`splitFramesString` strips the `"` and throws the distinction away (`hold-states.ts:167-173`), then `accumulateFramesToMaps` folds everything as deltas. For any climb that contains an unquoted non-first frame, every frame from that point on renders and plays back with the wrong holds lit.

## Scale (measured, not estimated)

Decoding the live Kilter Grips catalog and diffing against the stored Aurora `frames`:

| layout | multi-frame climbs | affected (contain an unquoted non-first frame) |
| --- | --- | --- |
| Kilter Board Original (`board_layouts.id = 1`) | 112 | **34** |
| Kilter Board Homewall (`board_layouts.id = 8`) | 29 | **5** |

The two encodings agree on the accumulated per-frame lit state under exactly one reading — unquoted ⇒ absolute — for all 39 of those climbs, and disagree under the current delta-only reading. Two independently produced encodings agreeing under precisely one interpretation is what identifies the correct semantics here.

## User impact

These are real, listed, animated climbs. The board animates the wrong holds and the play-through renders wrong — silently, with nothing to indicate it. Affects the renderer, the play view, and the BLE frame sequence (`toFlatFrames` / `accumulatedMapsToFrameStrings` are downstream of the same fold).

## Suggested fix

Have `splitFramesString` preserve whether a segment was quoted (or return `{ delta: boolean; body: string }`), and have `accumulateFramesToMaps` reset the accumulator before applying an unquoted non-first frame. Both are in `packages/board-constants`, consumed by web, mobile, and backend, so the change wants tests against real multi-frame strings from the catalog.

## Independent of #3523

Found while decoding the Grips `s`/`e` encoding for #3523, but entirely separate: #3523 is about animated climbs never being ingested at all, this is about the ones we *did* ingest being replayed wrong. Neither fix depends on the other.

PR #3951, opened 2026-07-26
fix(sync): decode animated Kilter climbs, persist ingest backlog

### LINKED ISSUE #3523: New multi-frame/undecodable climbs are silently skipped (no row, no stats) with only a log counter
## Summary
gripsClimbConcatToFrames returns null for any climb_concat the h{hole}p{code} regex can't fully consume (line 50) or containing a hole with no placement on the resolved layout (line 44) — the caller (catalog-sync.ts:299-309) counts it unmapped and skips the climb entirely: no board_climbs row, no alias, and (because it never enters climbUuidToCanonical) its stats are dropped at catalog-sync.ts:440. Existing multi-frame climbs survive via the UUID-identity fast path (they keep backfilled holds), but any NEW multi-frame/animated climb (s{frame}/e{frame} encoding, per the comment at catalog-sync.ts:287-289) or new climb on a fringe hole is permanently missing from Boardsesh, with no persisted record of the backlog — only the per-run log line.

## Failure scenario
A setter publishes a new animated (multi-frame) Kilter climb after the initial bulk ingest. Every catalog sync skips it as unmapped. Users who climb it in the Kilter app can't find it on Boardsesh, and if they tick it via logbook sync the tick renders against a missing climb.

## Evidence
`packages/kilter-sync/src/sync/catalog-parse.ts:50`

```
catalog-parse.ts:44 `if (placementId === undefined) return null;`, :50 `if (consumed !== frame.length) return null;`; catalog-sync.ts:299-309 increments climbsUnmapped and continues; :440 `if (!canonicalUuid) continue;` drops the stats.
```

## Suggested fix
Decode the s/e multi-frame encoding (or at minimum ingest frame 1 for display with a 'partial' marker), and persist the unmapped uuid backlog (table or metric) so growth is observable instead of a rotating log line.

marcodejongh:
## Summary

Kilter Grips encodes every hold in `climb_concat` as `h{holeId}p{roleCode}[s{start}][e{end}]`. Our parser only understood the `h{hole}p{code}` half, so any **animated** climb failed the full-consumption guard, got counted as unmapped, and disappeared — no `board_climbs` row, no alias, and its stats dropped downstream at the `!canonicalUuid` guard. Existing animated climbs survived on the UUID-identity fast path, so only genuinely *new* ones went missing. Permanently, and with nothing but a rotating log line to show for it.

Two commits, reviewable and revertable separately:

1. **`board_climb_ingest_skips`** — every climb the sync can't ingest is now persisted with its **verbatim** upstream hold string, the reason, and first/last-seen stamps. That raw payload is the durable win: the next time the encoding drifts, it can be decoded from rows we already have.
2. **The decoder** — `s`/`e` are 1-based and inclusive, omitted when they equal their default. A hold lit over `s..e` becomes a `p` token on frame `s` and, when `e` isn't the final frame, an `x` token on frame `e + 1` — the Aurora delta format the renderer and LED protocol already speak.

Closes #3523

> **Base is #3963, not `main`.** This PR no longer carries a migration. Four PRs each generated a `0187` against a `main` that tops out at `0186`, so `board_climb_ingest_skips` and its index moved, unchanged, into the single collapsed migration on `chore/collapse-0187-migrations` (#3963); this branch keeps only the decoder and backlog logic. GitHub retargets this to `main` when #3963 merges.

## Evidence, not inference

The encoding was decoded from **real captured payloads**, not guessed. `GET https://portal.kiltergrips.com/api/climbs/all/{layout}` serves the catalog unauthenticated (filed separately as #3949 — no credentials, no production database were used).

Across **100,513 live catalog rows** on two product layouts: every concat matches the grammar with no gaps; commas **never** appear (the old comma-splitting was vestigial, inherited from the Aurora format); `s`/`e` never appear on a single-frame climb; no index ever falls outside `1..frameCount`; and the encoder never writes an explicit `s1` or `e{frameCount}` — which is what pins down the defaults.

Then the decode was diffed against the legacy Aurora `frames` strings for every climb present in both catalogs. Those were produced by a different backend years before this parser existed, so they're an oracle rather than a restatement of our own output:

| layout | compared | byte-exact | state-equal | divergent | undecodable | **multi-frame parity** |
| --- | --- | --- | --- | --- | --- | --- |
| Original (`id=1`) | 72,711 | 72,552 | 158 | 1 | 0 | **112 / 112** |
| Homewall (`id=8`) | 13,343 | 0 | 13,343 | 0 | 0 | **29 / 29** |

**141/141 multi-frame parity over 86,054 climbs compared.** Every non-byte-exact bucket is explained: Homewall's Aurora rows store product-7 role codes (42-45) where Grips emits product-1 codes (12-15), which `HOLD_STATE_MAP` maps onto identical states; the 34 + 5 multi-frame ones agree under exactly one reading of Aurora's unquoted frames (see #3947). The single divergence is an upstream content edit, not a decode fault.

That table was produced by the **shipped TypeScript decoder**, not a prototype.

## Also fixed

- A multi-frame concat carrying no `s`/`e` at all parsed "successfully" into a **single** frame while `frames_count` came from the wire — an internally inconsistent row. Now emits `frameCount` frames.
- **Setter notifications are gated on the upstream `createdAt` being recent (30 days).** This ingest recovers animated climbs first published as far back as 2021; announcing those to followers as new climbs would be a worse regression than the bug being fixed. Diff is deliberately confined to the `newCanonicals` push site — `notifications.ts` is untouched, to stay out of #3539's way.

## Release Notes

- Animated multi-frame climbs now show up instead of quietly going missing — around 40 on the two layouts we measured, and every new one from here on.

## Test plan

- `vp check`, `vp run typecheck`, `vp test run --project kilter-sync` — all green (182 tests).
- **Parser tests assert against the Aurora catalog's own strings**, captured into `src/sync/__fixtures__/grips-multiframe.json` with full provenance. Six real climbs covering: a 15-frame climb with a hole re-lit on the frame it goes out; interleaved clear/light ordering; `frameCount > 1` with no `s`/`e`; `s` without `e`; a single-frame regression lock; and a second layout/product.
- Each assertion is paired with the mutation that breaks it — deleting the s/e branch, flipping `e` to exclusive, removing the same-frame clear suppression, removing the sort, dropping the range guard.
- **`packages/db` has no CI suite**, so the schema work was verified by hand against a **throwaway container** from the dev-db image on a spare port (not the shared dev DB): the table's full round-trip — insert, re-upsert preserving `first_seen_at`, resolve, re-open, reason filter — exercised against the real table. The DDL itself now ships in #3963; it was re-verified there, applied and idempotent on re-run, with no stray `DROP`.
- Full-catalog parity harness run against both layouts through the shipped decoder; results in the table above.

## Follow-ups filed

- #3947 — Aurora absolute frames read as deltas: **34 of 112** animated Original climbs and 5 of 29 Homewall climbs animate and play back wrong on the wall **today**. Live user-visible bug, independent of this PR.
- #3948 — `convertLitUpHoldsStringToMap` emits a phantom `hold_id 0` for differential frames strings (latent; aurora-sync path).
- #3949 — Kilter's unauthenticated catalog endpoint. Filed for Marco's decision; **no contact made**.

## Note on the commit split

Commit 1 lands the table and its writer standalone. Commit 2 wires it up, because a useful skip *reason* needs the structured decode result — so reverting commit 2 alone would leave the table unused rather than unwritten. Flagging that explicitly since the intent was that the backlog survive the decoder slipping; the decoder did not slip.


marcodejongh:
Thanks — addressed five of the six, pushed in b27b93d.

- **30-day boundary untested** — fixed. Both sides now pinned: exactly 30 days notifies, one millisecond past it does not.
- **`loadBacklog` unvalidated reason filter** — fixed, and made stricter than suggested. `KILTER_SKIP_REASONS` is now a runtime list with an `isKilterSkipReason` guard, `BacklogQuery.reason` is typed, and the CLI rejects an unknown `--reason` outright. Filtering on a typo previously printed "nothing skipped", which is the worst possible answer to get wrong for a table whose whole job is showing what's missing.
- **Silent persistent failure on backlog writes** — fair, and the right instinct. Rather than a new metric pipeline, the failure now sets `skipsWriteFailed` on the run summary, which is already JSON-logged every cycle and returned to the daemon and CLI. The log line is also now explicit about how many climbs went unrecorded.
- **UUID case in `markSkipsResolved`** — real latent bug, fixed. `loadOpenSkips` returns `lower(uuid) -> stored uuid` instead of a lowercased set, so the resolve stamp matches the primary key exactly even if upstream changes casing between runs.
- **`sql\`null\`` non-idiomatic** — fixed, plain `null`.

**Not taking: the missing trailing newline in `0187_tired_speedball.sql`.** Two reasons. It isn't inconsistent with the other migrations — 74 of the 189 files in `packages/db/drizzle/` also end without one, so this is just what `drizzle-kit generate` emits. More importantly, drizzle's migrator hashes the SQL file's exact bytes into `drizzle.__drizzle_migrations`; hand-editing a generated migration changes that hash, and `CLAUDE.md` is explicit that migration SQL is never hand-written. Not worth a whole class of drift for a trailing byte.

marcodejongh:
Two notes from the latest automated pass, for the record:

- Both "remaining observations" are already fixed in b27b93d and the review is reading the previous revision. `resolvedAt` is a plain `null`, not raw SQL. And `loadOpenSkips` now returns `lower(uuid) -> stored uuid`, so `resolvedSkipUuids` pushes the uuid **as stored in the table**, not the upstream casing — the case-sensitive `inArray` is now exact by construction.
- "Inconsistent with every other migration file" isn't accurate: **74 of the 189** files in `packages/db/drizzle/` end without a trailing newline. It's `drizzle-kit generate` output, and editing it would change the hash drizzle records in `drizzle.__drizzle_migrations`.

### Environment note for whoever runs this locally

`vp run db:migrate` `dependsOn: ['db:up']`, and `scripts/dev-db-up.sh` applies pending `drizzle/*.sql` **straight into the local dev-db docker container by name** — it does not read `DATABASE_URL`. So pointing `DATABASE_URL` at a throwaway instance migrates the throwaway *and* the shared local dev DB on port 5432. Worth knowing if you keep several worktrees on one dev container; `board_climb_ingest_skips` is additive and empty, so it's harmless, and `docker compose down -v && vp run db:up` clears it.

Issue #3952, opened 2026-07-26
Session.isLeader is connection-scoped from JOIN/LeaderChanged but participant-scoped from SessionRosterSnapshot


marcodejongh:
`Session.isLeader` means two different things depending on which event set it, and the roster-snapshot meaning is wrong for a multi-device participant.

Found while fixing #3502. Not fixing it there — that's a UI PR and this is roster semantics — but it's a live trap for anyone who reaches for `isLeader` as a per-connection signal.

### The inconsistency

`RuntimeSessionState.isLeader` is documented as *this connection's* leadership, and two of its three writers honour that:

- **JOIN response** (`buildSessionPayload`, `sessions/helpers.ts`) — `leaderConnectionId === ctx.connectionId`. Per-connection. Correct.
- **`LeaderChanged`** (`packages/shared/queue-runtime/src/session-events.ts`) — `effectiveLeaderConnectionId === prev.clientId`. Per-connection. Correct.
- **`SessionRosterSnapshot`** (same file) — `isLeader: selfInSnapshot ? selfInSnapshot.isLeader : prev.isLeader`, where `selfInSnapshot` is matched by **participant id**. Per-participant. **Wrong for a multi-device participant.**

### Why the third one diverges

`joinSession` resolves `participantId = client.userId || connectionId` (`room-manager/client-lifecycle.ts:133`), so one signed-in climber on two devices is a *single* participant with two `connectionIds`. `upsertLocalParticipant` then makes leadership sticky across those connections:

```ts
// room-manager/client-lifecycle.ts:1022
isLeader: existing?.isLeader || isLeader,
```

`getSessionUsersLocal` maps that participant row straight onto `SessionUser.isLeader`, so the roster's `isLeader` is effectively "this *human* holds the leader connection", not "this connection is the leader".

### Observable effect

A `SessionRosterSnapshot` is the first event yielded on **every** `sessionUpdates` subscribe (`resolvers/sessions/subscriptions.ts:53`). So:

1. Creator's phone A joins → leader, `isLeader: true`. Participant row is leader.
2. Phone B (same signed-in user) joins → JOIN response correctly says `isLeader: false`.
3. Phone B subscribes, receives the roster snapshot, matches itself by participant id, finds `isLeader: true` on the shared row → **phone B now believes it is the leader.**

[...]

PR #3956, opened 2026-07-26
fix(mobile): offer Leave as well as End for a session

### LINKED ISSUE #3502: Leave ui when joined session from 2 phones
When I joined from my second phone that phone would let me end the session instead of simply leave it

marcodejongh:
## Summary

Joining your own session from a second phone offered **End session** as the only way out — and it worked, because the HTTP `endSession` branch authorizes on `createdByUserId` alone, which your second phone passes. The only visible exit genuinely terminated the party for everyone. 19 of 23 fingerprinted production sessions are single-user multi-device, so this was the common path, not an edge case.

The other half: a climber who joined *someone else's* party got the same button, it threw server-side, and `endSession`'s catch dropped them out locally anyway with a generic "Action failed" — while peers waited out the full 60-second disconnect grace before seeing them go.

Both exits now exist. **Emphasis** (which one the sheet leads with) comes from device provenance; **availability** of End comes from the session's creator id.

| | Chrome control | Sheet leads with | End available? |
|---|---|---|---|
| Phone that started the session | red **Stop** (unchanged) | End + recap | yes, primary |
| Same climber's second phone | neutral **Leave** | Leave | yes, via "End it for everyone" |
| Someone else's session | neutral **Leave** | Leave | **no** |

### Why not the roster

`joinSession` resolves `participantId = client.userId || connectionId`, so one signed-in climber on two devices is a **single participant with a single roster row**. Every roster-derived signal is participant-scoped and structurally blind to the second device.

`Session.isLeader` looks like the free answer and is a trap: `upsertLocalParticipant` ORs leadership across a participant's connections (`existing?.isLeader || isLeader`), and the `SessionRosterSnapshot` handler re-derives the client's own flag from that participant row. The second device is therefore told `isLeader: true` by its *very first* roster snapshot — the same `clientId`-vs-`participantId` id-space class that #3907 had to fix from the other direction. Building the gate on it would have shipped a fix that silently un-fixes itself. Filed separately as #3952.

`LocalSessionParticipant.connectionIds` does distinguish connections, and `getSessionParticipants` already computes the live count. Deliberately not used: it answers "how many sockets does this human hold right now", which flaps with screen locks, and a stale value degrades toward offering the **destructive** action.

### What drives it instead

- **Emphasis — device provenance.** `session-store.ts` records the id of the session this device created. It drives which action leads, never which actions exist, so losing it to a reinstall costs the End-first default and nothing else.
- **Availability — `Session.createdByUserId`** (new, member-only). `sessionDetail.ownerUserId` can't do this job: it returns null until the first tick, which is exactly when you join a friend's fresh party and want back out.
- Mobile reads it through its **own `GET_SESSION_OWNER` document**, not another field on `GET_SESSION` — that query also backs join-by-link, and GraphQL validates whole documents, so a new-bundle/old-backend skew would break joining outright rather than just leaving ownership unknown.
- **`canEnd` is permissive when ownership is unknown.** Failing closed would strand a creator during the fetch window and on anonymous sessions. The server re-checks regardless — this is UI, not authorization.

### Privacy

`createdByUserId` is redacted for non-members. The non-member preview roster already carries every *present* member's `userId` by design (the join-confirmation screen needs it), but that set diverges from the creator the moment they disconnect — a stranger holding only a session UUID has no business learning who started it.

### Also

- `endSession`'s failure path now emits `LEAVE_SESSION` instead of only clearing locally, so a refused end stops leaving a ghost in everyone else's roster for 60s. The message is honest about what happened: production masks GraphQL errors, so an auth refusal and a dropped network are indistinguishable client-side.
- The chrome shows a neutral Leave rather than the red Stop when the exit doesn't end anything. A destructive-looking control that performs a non-destructive action trains people to distrust the colour on the one surface where that trust matters.

Web has the equivalent hole (`queue-control-bar.tsx`'s `uniqueSessionUsers.some(...)` collapses one user's two devices into a single roster entry, so its second tab also takes the END path). Not fixed here — the web climbing surface is being replaced by the RN web build.

Closes #3502

## Release Notes

- Join your own session from a second phone and you can now just leave it — ending the session for everyone is a separate, deliberate step.
- Leave someone else's session without being offered a button that kills it for the whole crew.
- Your crew sees you drop out straight away instead of waiting a minute for the connection to time out.

## Test plan

- `vp check` — 0 errors
- `vp run typecheck:mobile`, `vp run typecheck:backend` — pass
- `vp run test:mobile` — full suite
- `vp test run --project backend` — `build-session-payload`, `session-query-gate`, `session-context`
- **Mutation-tested both gates.** Deleting `canEnd = !isKnownForeign` fails "withholds End from a climber who joined someone else's session"; deleting the `startedOnThisDevice` term in `defaultMode` fails both the second-phone and ownership-unknown cases. Neither gate is decorative.
- New coverage: `in-session-view-exit.test.tsx` drives the **real** `useSessionExitOptions` hook with only its inputs stubbed, so the tests can't pass a re-implementation of themselves. Leave asserts `clearSession` is called with **exactly** `{ notifyServer: true }` — a "simplification" to a bare call is the regression that would put peers back on the 60s grace.
- QA notes in `.boardsesh/qa-notes.md`, including a two-connection repro that needs no second phone: sign in on web and start a session, then join it from the emulator as the same account.

⚠️ **Device QA gate:** the two-connection topology itself is manual — vitest covers the decision, not the sockets.

### DIFF
--- a/docs/websocket-implementation.md
+++ b/docs/websocket-implementation.md
@@ -504,6 +504,19 @@ The main Traefik instance routes `*.preview.boardsesh.com` and `*.ws.preview.boa
 
 Explicit UI actions still leave or end sessions immediately: `leaveSession` removes the participant and emits `UserLeft`; `endSession` is restricted to the session creator or current leader. Passive WebSocket disconnects emit `UserPresenceChanged` first and only emit `UserLeft` if the reconnect timer expires.
 
+#### Leave vs. end on the client (#3502)
+
+Both platforms expose leave and end as distinct actions. Which one a surface *leads with* is a client concern — the server authorizes each independently.
+
+The subtlety: **one signed-in climber on two devices is a single participant.** `joinSession` resolves `participantId = client.userId || connectionId` (`room-manager/client-lifecycle.ts`), so two phones share one participant entry and one roster row. Every roster-derived signal is therefore participant-scoped and structurally cannot see the second device:
+
+- `SessionUser.isLeader` is the OR of leadership across a participant's connections (`upsertLocalParticipant` keeps it sticky-true), and the `SessionRosterSnapshot` branch in `@boardsesh/queue-runtime` re-derives the client's own `isLeader` from that row. A second device is therefore told `isLeader: true` by its very first roster snapshot, even though its connection is not the leader. **Do not build device-level UI on `isLeader`.**
+- `LocalSessionParticipant.connectionIds` (and its Redis equivalent) does distinguish connections, but is deliberately not exposed over GraphQL. It answers "how many sockets does this human hold right now", which flaps with screen locks and reconnects — and a stale value degrades toward offering the *destructive* action.
+
+Mobile instead keys the **emphasis** on device provenance (`session-store.ts` records the id of the session this device created) and the **availability** of End on `Session.createdByUserId` (member-only; redacted for the non-member preview). Provenance decides which action leads, never which actions exist — so losing it to a reinstall costs a creator the End-first default and nothing else.
+
+Mobile reads `createdByUserId` through its own `GET_SESSION_OWNER` document rather than as another field on `GET_SESSION`: that query also backs the join-by-link screen, and GraphQL validates whole documents, so a new-bundle/old-backend skew would otherwise break joining outright instead of just leaving ownership unknown.
+
 If the grace period expires, the session is evicted from memory but remains restorable from Redis until the 4-hour TTL expires. After Redis TTL expiry, the session is still not ended; the next join restores it from Postgres as long as the durable row has not been explicitly marked `ended`.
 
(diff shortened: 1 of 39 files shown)


marcodejongh:
Thanks — two of three were real and are fixed in `5340804`.

**1. Silent leave failure** — correct, the catch reported to Sentry and showed nothing, so a failed teardown looked like the button did nothing. Now toasts `mobile.queue.actionFailed` (existing key, all three locales, accurate for a genuine local-teardown failure). I kept the sheet open rather than dismissing: that matches `handleConfirmEnd`'s catch, which deliberately stays open so the climber can retry, and "Keep climbing" / pan-down are still the way out.

**2. First-frame exit label** — correct and the better catch. Provenance is a SecureStore read, and the old code conflated "not read yet" with "didn't start it", so a creator saw Leave flip to the red Stop a beat later. `undefined` is now a distinct not-yet-read sentinel and emphasis falls back to `end` while unresolved, making the first frame identical to the pre-#3502 chrome. `startedOnThisDevice` stays honest (false until read) so the analytics split isn't polluted, and a joiner's brief red Stop is safe because `canEnd` is computed independently of provenance — a known-foreign participant is still refused End mode by the sheet.

Chasing that also surfaced a related bug you didn't flag: the sheet re-armed its mode on every `defaultMode` change, so a late resolve would have yanked the sheet out from under someone who'd already tapped through to the other mode. Now it re-arms on the open transition only.

**3. Unused `exitLabel` in `RecordTopChrome`** — this one's incorrect. It's consumed twice: `accessibilityLabel={exitLabel}` on the Material `Appbar.Action` (line 152) and on the glass `PressableSurface` (line 181). No change made.

Both new guards are mutation-verified — collapsing the fallback to `leadWithEnd = startedOnThisDevice` fails the new first-frame test. Full mobile suite: 4525 passed.

Issue #3871, opened 2026-07-24
Aurora sync: DB-level write failure in a bids/ascents chunk still rolls back the whole cross-table sync transaction


marcodejongh:
## Summary
#3520 fixed the JS-level crash class in `applyAuroraAscents`/`applyAuroraBids` (`packages/aurora-sync/src/sync/apply-user-logbook.ts`): a single row with an unparseable `climbed_at`/`created_at` used to throw out of the eager per-payload normalize pass, aborting `syncUserData`'s single cross-table transaction and blocking the sync checkpoint from ever advancing. That fix wraps each row's *normalization* in try/catch and skips-and-logs the bad one.

It does **not** cover the structurally identical failure one layer down: a row that parses fine in JS but violates a Postgres type or constraint once it reaches the batched write inside `applyLogbookChunk` (e.g. a non-numeric `angle` hitting an `integer` column, or any other value that survives `Number()`/`String()` coercion but fails at the SQL layer). That write happens in chunks of up to `WRITE_CHUNK_SIZE` (100) rows via `db.insert(...)` / `db.execute(...)` (UPDATE via `jsonb_to_recordset`). A single bad row in that chunk still throws out of `applyLogbookChunk`, still propagates out of `applyAuroraAscents`/`applyAuroraBids`, and still rolls back `syncUserData`'s whole cross-table transaction for that attempt — the same "one poison row wedges the user's entire sync forever" symptom #3520 fixed for the JS-parse case, just triggered at DB-execution time instead of normalize time.

## Why this wasn't fixed in #3520
Isolating a single row's failure *inside* a batched multi-row SQL write (INSERT/UPDATE via `jsonb_to_recordset`) is a different, larger change than a per-row try/catch around a pure JS function: it needs either row-by-row fallback writes when a chunk's batched write fails, or per-row validation strong enough to guarantee the batch always succeeds. Out of scope for #3520's surgical fix.

## Suggested approach (not scoped/designed yet)
- On a chunk write failure, retry the chunk row-by-row so only the offending row is skipped-and-logged (mirrors the #3520 fix's philosophy) instead of losing the whole chunk.
- Or: tighten row validation before the chunk write (e.g. reject non-finite `angle`) so known-bad shapes never reach the SQL layer, with the row-by-row retry as a backstop for anything validation doesn't anticipate.

## References
- Fixed by: #3520 (JS-level normalize crash — this issue is the DB-level sibling it explicitly did not cover)
[...]

Issue #2933, opened 2026-06-16
migrate.ts: gap in __drizzle_migrations is only warned, not failed — earlier migrations silently skipped


marcodejongh:
## Summary

`packages/db/scripts/migrate.ts`'s journal-verification step (gated by `VERIFY_MIGRATION_JOURNAL=1`) only **warns** when the applied-migration count doesn't match the journal, then continues as long as the *latest* migration is recorded. Because drizzle's migrator advances on a single `max(created_at)` high-water-mark, a migration that is missing **below** that mark is never re-applied — so "latest is applied" does **not** imply "all are applied." A real gap therefore deploys green while leaving schema objects missing.

## Real-world impact (this week)

Production's `__drizzle_migrations` had **125 rows for 131 journal entries**. Migration `0129_numerous_star_brand` (which `CREATE`s `location_sync_gym_sources`) was in that gap and had **silently never applied**. The deploy log showed:

```
⚠️ Migration table has 125 rows for 131 journal entries; continuing because latest migration 0130_vote_count_skip_guard is applied.
✅ Migrations completed successfully
```

The first Kilter location sync in prod then hard-failed:

```
Failed query: select "gyms"."id" from "location_sync_gym_sources" ...  (relation does not exist)
```

It took manual diagnosis (per-object existence checks across 0125–0130) to find that *only* 0129 was missing, then a hand-applied DDL to unblock. The warning was emitted on every deploy but, by design, ignored.

## Root cause

Two compounding factors:

1. **drizzle high-water-mark**: `migrate()` applies entries where `folderMillis > max(created_at)` in the ledger. Anything missing *below* the current max is never retried — there is no per-migration reconciliation.
2. **Verification is advisory**: `migrate.ts` fails only when `latestAppliedCreatedAt < latestJournalEntry.when`. A count mismatch (`appliedMigrationCount !== journal.entries.length`) is a `console.warn` + continue. So a real gap is reported but not enforced.

Relevant code — `packages/db/scripts/migrate.ts`:

```ts
if (latestAppliedCreatedAt < latestJournalEntry.when) {
  throw new Error(...);                       // only catches a MISSING LATEST migration
}
if (appliedMigrationCount !== journal.entries.length) {
  console.warn(`⚠️ Migration table has ${appliedMigrationCount} rows for ${journal.entries.length} ...`
             + ` continuing because latest migration ${latestJournalEntry.tag} is applied.`);  // <-- gap tolerated
}
```

## Proposed fix

[...]

PR #3957, opened 2026-07-26
fix(sync): stop overlapping daemon instances double-running syncs

### LINKED ISSUE #3539: No cross-instance mutual exclusion for sync daemons: overlapping instances double-run user syncs and duplicate setter notifications

marcodejongh:
## Summary

Both sync daemons assumed they were a singleton and nothing enforced it. Under a Railway rolling deploy the old and new containers overlap, and both could pick the same credential (the candidate `SELECT` was unclaimed), both started with an empty in-memory cooldown `Map`, and both ran the same board-wide sync.

New-climb detection is a **pre-read** of which climb uuids already exist (`shared-sync.ts` `upsertClimbs`), run inside a transaction — so two simultaneous runs both classify the same climbs as new and both insert a full set of "new climbs from &lt;setter&gt;" rows. `notifications_dedup_idx` is a plain index, not UNIQUE, so nothing stopped the second set.

Closes #3539

### The lease is an optimisation, not mutual exclusion

Worth stating plainly, because the fix would be wrong if read the other way round. A TTL lease has **no fencing token**: if the holder stalls past the TTL (GC pause, a long await inside a cycle, a partition to Postgres, a suspended container) another instance legitimately takes over *while the first is still running and still believes it holds the lease*. The TTL bounds that window; it does not close it.

So the correctness comes from three guards that are each a single atomic statement and each safe on their own under genuine concurrency. They are **not** redundant with the lease, and none may be dropped on the grounds that "the lease should prevent this":

1. **`claimNextCredentialForSync`** — locks the chosen credential with `FOR UPDATE SKIP LOCKED` and stamps the attempt clock inside one short transaction, so overlapping instances take disjoint work. Adding `SKIP LOCKED` to the old bare `SELECT` would have been a **no-op**: outside an explicit transaction the implicit one commits and drops the row lock immediately. Both runners now share the ordering and backoff predicate so they can't drift.
2. **`claimSharedSyncSlot`** — replaces the per-process cooldown `Map`s with a single-statement compare-and-set (`ON CONFLICT DO UPDATE ... WHERE`) on a synthetic `__local_*` `board_shared_syncs` cursor, the namespace the weekly gate already uses. The cooldown now survives a restart, so a deploy no longer re-fires a full shared sync per board on its first cycle. **This, not the lease, is what stops the duplicate notifications.**
3. **`setterSyncNotificationUuid`** — makes the notification uuid deterministic (uuid v5 over `type|recipient|entity|actor`) so a repeat insert collides on `notifications.uuid`, which is *already* `NOT NULL UNIQUE`, paired with `onConflictDoNothing()`.

#### Why the dedup key is the batch head, and what to reach for if it ever isn't enough

`entityId` is the batch's **head** climb uuid, so the key holds as long as two concurrent runs derive the same head. They do today: `processBatches` is sequential and both instances consume the same Aurora page in the same order, so head stability is the common case, not the lucky one. Both call sites carry a comment saying so.

A sorted hash of the whole new-climb set was considered and rejected. It is more robust to *reordering* but strictly **less** robust to *set difference* — and set difference is the case that actually occurs: a partially-committed instance A makes instance B legitimately see a smaller "new" set. There the head key still dedups whenever the head is unchanged, while a set hash yields a different uuid every time and therefore **never** dedups, which is precisely the failure the mechanism exists to prevent. Degrading gracefully beats degrading to nothing.

**Residual, recorded so nobody re-derives this:** if duplicate setter notifications are ever observed in production despite all of the above, the next lever is a coarser key — `(recipientId, actorId)` plus a time bucket — which dedups regardless of set membership, at the cost of possibly suppressing a genuinely distinct second batch inside the same bucket.

The lease (`sync_daemon_leases`, whose table ships in #3963) sits on top and buys one thing: the ordinary case — two healthy containers during a deploy — stops doing everything twice.

### Why a lease row and not `pg_try_advisory_lock`

The issue suggested advisory locks. They can't work here: application traffic goes through **PgBouncer in transaction-pooling mode** (`docs/neon-migration.md:47`, and the `prepare: false` comments in both runners), where a session-scoped advisory lock leaks onto a server connection the holder no longer owns and can never release. Two further independent breakages on the same path: `postgres-js` is configured `idle_timeout: 30` (an idle connection is reaped, silently dropping the lock), and drizzle round-robins across `max: 5`, so you can't even tell which connection holds it. A lock that silently never holds is worse than no lock.

A lease row carries no session state, works through any pooler, and an operator can read it with a plain `SELECT`.

### Losing the lease mid-cycle

A cycle that discovers mid-flight that it lost the lease stops at its next checkpoint (`DaemonLease.assertStillHeld`, called between units of work) rather than running to completion alongside the new holder. The loop reports it via `onCycleError` and drops into standby.

The loser **idles** on a short 30s standby poll rather than exiting — a crash-looping loser during a rolling deploy would be worse than the duplicate work. The gate **fails closed** on a DB error (a daemon that can't reach Postgres can't sync anyway, so standing by costs nothing). The lease is released on SIGTERM so a deploy hands over in milliseconds instead of idling out the TTL.

### Deliberate semantics change

`last_sync_attempt_at` now advances when an attempt **starts** rather than when it finishes, so a per-credential backoff window is measured from attempt-start. The upside: a process killed mid-sync no longer instantly replays the same credential on reboot.

### Why not a composite UNIQUE dedup index

`actor_id` is NULL whenever the setter has no linked Boardsesh account — the majority case — and Postgres treats NULLs as distinct in a unique index, so a composite UNIQUE on `(actor_id, recipient_id, type, entity_id)` would have missed exactly the duplicates we care about. `NULLS NOT DISTINCT` fixes that, but drizzle-orm 0.45.2's pg-core index builder can't express it, and migration SQL is generated, never hand-written. Riding the existing `notifications.uuid` unique constraint needs **no migration, no index build over production rows, no backfill and no deletion of historical duplicates** — rows written before this carry random v4 uuids that can never collide with these v5 values.

### Also fixed on the way: a wrong error state climbers could actually see

This is easy to skim past as infrastructure, so stating it plainly — **the two bugs found while building this are the most user-visible thing in the PR.**

A DB error in the cooldown path escaped into the caller *after* the user-half of the cycle had already committed and been marked `active`. The caller treats anything thrown from there as a **credential failure**: it stamps `consecutive_failures`, records `last_sync_error`, and can flip the board card to an error state — for a climber whose sync had, in fact, just succeeded. Repeat it a few times and the exponential backoff starts skipping that climber's syncs on the strength of failures that never happened.

Two separate escape paths, both now contained:

- **Claim path** — `claimSharedSyncSlot` / `readSharedSyncCursor` throwing before the work starts.
- **Re-stamp path** — the cursor stamp after the work. Worse than it looks: on the old success path the stamp sat inside the `try`, so a DB error there fell into the `catch`, which ran a *second* stamp that could throw too — and that one escaped. Both runners now stamp from a `finally` via a helper that reports its own errors and never throws.

Neither could happen with the old in-memory `Map`, so both are regressions this PR would have introduced had the review not caught them. A missed stamp only means the cooldown measures from the start of the run instead of its end — cheap next to a false failure on someone's account.

### Not addressed

The issue also mentions concurrent upserts deadlocking. I could not verify it: `processBatches` is sequential and both instances receive the same Aurora page in the same order, so identical lock ordering gives *blocking*, not deadlock. There's no log evidence either way. It isn't a target of this change — the fix removes the concurrency regardless.

## Release Notes

- Followers stop getting the same "new climbs from &lt;setter&gt;" alert twice after a deploy.
- Your board stops showing a sync error for a sync that actually worked.

## Test plan

`vp check` and the full `vp run typecheck` both green, plus:

| suite | result |
| --- | --- |
| `vp test run --project sync-runtime` | 21 passed |
| `vp test run --project aurora-sync` | 98 passed |
| `vp test run --project kilter-sync` | 156 passed |
| backend, 5 real-DB files | 33 passed |

### Mutation testing

Multi-instance races are the classic thing that "passes" with a test that can't actually fail, so every guard was disabled in turn to confirm its tests go red. Results recorded rather than asserted:

| mutation | effect |
| --- | --- |
| drop `.for('update', { skipLocked: true })` from the credential claim | `credential-claim-skip-locked` — **3 of 4 fail** (the 4th only checks the attempt-clock stamp, correctly unaffected) |
| replace the cooldown CAS with the read-then-write shape the weekly gate uses | `shared-sync-cooldown-cas` — **"exactly one of five parallel claims wins on a cursor whose cooldown just elapsed" fails** |
| make the lease `ON CONFLICT` unconditional (drop the holder/TTL guard) | `sync-daemon-lease` — **5 of 9 fail** |
| revert the deterministic uuid to `randomUUID()` | `setter-notification-dedup` — **4 of 7 fail** |

One honest caveat: under the cooldown mutation the *cold-cursor* five-way race still passed, because the first claim's insert happened to commit before the other four had read. That test is timing-dependent and can pass by luck; the **warm-cursor** variant is the one that reliably catches it, and it's the case two overlapping daemons actually hit an hour after the last run. Both are in the suite.

### Test design notes

- The credential-claim interleave is **deterministic**, not a hopeful `Promise.all`: instance A's transaction is opened by hand and parked mid-claim, so instance B provably runs while A holds the row lock. A genuine parallel case sits alongside it.
- `credential-selection-backoff.test.ts` used to rebuild the selection predicate inside the test, which made it a tautology that would stay green while the runner drifted. It now calls the real shared helper.
- The dedup tests call the real `createSetterSyncNotifications` from **both** sync packages, not a re-expressed insert, and include a cross-package case (aurora and kilter agree on the uuid, so a board synced by both never double-notifies).
- The aurora runner's throttle unit tests now assert delegation (claims, honours a refusal, re-stamps on success *and* failure); the CAS semantics themselves are covered against real Postgres.

### Migration — now in #3963, not here

The provisional-`0187` warning that used to sit here has been actioned. Four PRs each generated a `0187` against a `main` that tops out at `0186`; rather than let them serialise, all four schema changes were collapsed into **one** migration on `chore/collapse-0187-migrations` (#3963), which this PR is now based on. `CREATE TABLE sync_daemon_leases` and its `schema-sql.ts` mirror went there unchanged — this branch carries only the lease logic and queries. GitHub retargets this to `main` when #3963 merges.

Nothing was run against the production database. `vp run db:migrate` was never run.

## Commits

Split so the aurora refactor can be reverted independently:

1. `refactor(aurora-sync): use the shared daemon loop instead of a forked copy` — aurora carried its own copy of the daemon loop while kilter consumed `@boardsesh/sync-runtime`; the fork was a strict subset missing the option validation. One gated loop is the point of the rest of the change.
2. `fix(sync): stop overlapping daemon instances double-running syncs` — everything above.


marcodejongh:
Thanks — took the cosmetic one, and I want to push back on the dedup-key suggestion because I think it inverts the failure mode.

**`nowCursorText` precision** — fixed. Padded to 6 fractional digits so a `__local_*` row is byte-compatible with the Aurora-written rows sharing the column. No parse change either side.

**Migration number** — agreed and already called out in the PR body. `origin/main` is still at `0186` as of now, so `0187` is valid at this instant, but whichever of #3931 / #3951 / #3871 lands first takes it. The procedure is in the body: rebase, delete the migration + snapshot + journal entry, regenerate with `bunx drizzle-kit generate`. Never renumber by hand — a journal gap makes `migrate.ts` silently skip earlier migrations (#2933). The migration body is a single `CREATE TABLE`, so regenerating costs nothing.

**Set-based dedup key — I don't think this is an improvement.** The proposal is to key the v5 uuid on a sorted hash of every new climb uuid for the (setter, recipient) pair instead of the batch head. That's strictly more robust against *reordering*, and strictly **less** robust against *set difference* — and set difference is the realistic case, while reordering is hypothetical.

Concretely, two overlapping runs do not see identical sets. New-climb detection is a pre-read of which uuids already exist, so if instance A commits part of its batch before instance B's pre-read, B legitimately sees **fewer** new climbs. Then:

- **head-uuid key (current):** the two keys still match whenever the climbs A saw first are the ones B also saw first — which is the common shape, since both consume the same Aurora page in the same order. Partial overlap still dedups.
- **sorted-set key (proposed):** any difference in set membership changes the hash, so the keys diverge and **both notifications land**. Partial overlap never dedups.

So the proposed key trades a hypothetical failure (someone later introduces a non-deterministic sort) for a likelier one (the two runs are simply not perfectly synchronised). I'd rather keep the key that degrades gracefully, and I've documented the ordering assumption at both call sites so a future reorder is a conscious decision rather than a silent one.

Worth restating that this is a **backstop**, not the mechanism: what actually stops two concurrent board-wide syncs is `claimSharedSyncSlot`, the single-statement compare-and-set on the `board_shared_syncs` cursor. The uuid guard only has to hold for the window where that CAS is bypassed entirely.

**Cross-package (aurora + kilter for the same setter)** — this can't arise in practice: aurora-sync owns every board *except* kilter and kilter-sync owns kilter, and `entity_id` is a board-scoped climb uuid, so the two packages never notify about the same climb set. The cross-package test in `setter-notification-dedup.test.ts` pins that they'd agree on the uuid anyway, as a belt-and-braces property rather than a path we rely on.

Happy to switch to the set-based key if you disagree with the partial-overlap reasoning — it's a small change — but I didn't want to make it silently on an argument I think points the other way.

Issue #3857, opened 2026-07-24
Mobile queue sync breaks when a queue item UUID is not RFC-formatted — setQueue rejects the whole queue with "Invalid queue: Invalid UUID format"


marcodejongh:
## What's happening

A queue item whose wrapper `uuid` isn't a strict RFC-4122 UUID makes the backend reject the **entire** `setQueue` payload with `GraphQLOperationError: Invalid queue: Invalid UUID format`. Once such an item sits in a user's local queue, every mobile queue-sync attempt (resync + session seed) fails until the queue is cleared — and on the party-session-start path the rejected seed lets the empty-room FullSync **wipe the user's live queue**.

## Sentry (env=production)

[**BOARDSESH-83**](https://boardsesh.sentry.io/issues/7580110659/) — 116 events lifetime / ~80 in 14d / **23 users**, `environment=production` (116/116), still on the latest release (63/116 on `2.2.2+3`), cross-platform (iOS 101, Android 14), `handled:true`, ongoing since 2026-06-28, last seen 2026-07-24. Reported from multiple sync paths: `source=queue-sync` (41), untagged (69), `source=startSessionSeed` (6).

## Root cause (verified on `main`)

- The backend enforces a **strict** UUID on the queue-item wrapper id: `ClimbQueueItemSchema.uuid = z.string().uuid('Invalid UUID format')` — `packages/backend/src/validation/schemas/climbs.ts:110`.
- Applied to the whole queue on every `setQueue`: `validateInput(QueueArraySchema, queue, 'queue')` — `packages/backend/src/graphql/resolvers/queue/mutations.ts:396`.
- The wrapper `uuid` is an **opaque client id**, not an Aurora climb uuid. Both current clients already emit valid v4 UUIDs (web `uuidv4()`, mobile `randomUUID()` from expo-crypto), so the offending ids come from **legacy persisted snapshots / peer-synced items** generated before strict-UUID generation. They can't be re-uploaded, so they wedge sync indefinitely.
- A lenient schema already exists for exactly this reason a few lines up: `ExternalUUIDSchema = z.string().min(1).max(50)` (`packages/backend/src/validation/schemas/primitives.ts:12`), and `ClimbQueueItemSchema` itself uses it for other fields (`climbs.ts:16`). Only the wrapper `uuid` at `:110` was left strict.
[...]

PR #3958, opened 2026-07-26
fix(shared): stop the queue boundary stripping owner, draft and mirror state

### LINKED ISSUE #3927: Queue boundary strips userId / description / mirrored / is_draft / published_at
Split out of #3337, where carrying these locally-only turned out to be actively harmful.

## What is wrong

`climbToQueueItem` (`packages/mobile/src/lib/climb-to-queue-item.ts`) drops five fields the source climb already carries:

- `userId`
- `description`
- `mirrored`
- `is_draft`
- `published_at`

`toClimbInput` in the same file sends all five to party peers, and the owner-only Edit action gates on `userId`, so a queued climb quietly loses them.

`mirrored` is the sharpest one: `packages/shared/queue/src/types.ts` says the reducer reads exactly `uuid` and `mirrored` off a climb, so dropping it at the queue boundary means a mirror flag set before queueing does not survive.

## Why it cannot be fixed in one file

Carrying them in `climbToQueueItem` alone makes things worse, not better. Neither subscription selection set selects them:

- `SUBSCRIPTION_CLIMB_FIELDS` — `packages/mobile/src/lib/graphql/operations.ts:1183`
- `CLIMB_FIELDS` — `packages/shared/graphql/src/operations/queue-session.ts:5`

So a peer rebuilds the climb through `toClimbQueueItem` (`packages/mobile/src/lib/queue-conversion.ts:57`) without them, and that peer's next `setQueue` / `setCurrentClimb` sends the gap back via `toClimbInput`, overwriting the originator's copy. The creator then loses them again on the next FullSync. Today the fields are never populated, so there is nothing to lose — populating one side alone is what creates the flapping.

`packages/mobile/src/lib/queue-conversion.ts:8-14` already documents this coupling and says to keep the two in sync.

## What the fix has to move together

1. `SUBSCRIPTION_CLIMB_FIELDS` (mobile)
2. `CLIMB_FIELDS` (web/shared)
3. The `SubscriptionClimb` type + `toClimbQueueItem` (mobile)
4. Web's equivalent conversion
5. `climbToQueueItem`

All five fields already exist on the GraphQL `Climb` type (`packages/shared-schema/src/generated/schema.graphql`), so this is a selection-set widening, not a schema change. The queue state hash is UUID-only (`packages/shared/queue/src/state-hash.ts`), so a wider payload does not cause hash drift.

## Guard already in place

`packages/mobile/src/components/create-climb/__tests__/use-create-climb-screen-queue.test.tsx` has a deliberate assertion that these five are still `undefined` after `climbToQueueItem`, with a comment saying to flip it to `toMatchObject` when this lands. That is the tripwire against widening the boundary by halves.

## QA

Needs a real two-device party session — the failure mode is a peer round-trip, not anything a unit test can see. Worth checking `mirrored` specifically, since it is the one the reducer reads.

marcodejongh:
## Summary

Closes #3927.

`climbToQueueItem` hand-picked a field subset that had drifted from `toClimbInput` and from both subscription selection sets, so a climb lost `userId` / `description` / `mirrored` / `is_draft` / `published_at` the moment it entered the queue. The play drawer feeds the queue item's climb straight to the Edit gate (`use-climb-actions` / `ClimbActionsSheet` read exactly `userId` + `is_draft` + `published_at`), which is why the owner-only Edit action never showed up on a queued climb.

The dangerous property is not the missing field, it's the **flap**: carrying a field client-side while a peer's selection set omits it makes the peer rebuild the item without it and push the gap back to everyone on its next full-queue write. A field that blinks in and out is worse than one that is consistently absent. That is why this had to move as a coordinated set rather than one file.

### Three corrections to the issue as filed

1. **"Today the fields are never populated, so there is nothing to lose" is false for web.** `toClimbQueueItemInput` already sent all five; `CLIMB_FIELDS` selected none of the four. Web has been flapping them in production for any climb sourced from `CLIMB_DETAIL_FIELDS`. This PR *closes* an existing web flap rather than opening one.

2. **`mirrored` is not "the sharpest one".** No column backs `Climb.mirrored` and no resolver populates it, so search and detail responses always return it null, and `mirrorCurrentClimb` is web-only. The only mobile paths that lost it re-derive an item from an already-mirrored climb (climb-actions preview, play-drawer open, wall preview). Real, narrow, and — because its wire path was already complete — zero-risk.

3. **There is a seventh drift surface the issue misses: the backend Zod schema.** `setQueue` / `joinSession` persist the *parsed* item, unlike the single-item mutations which keep the GraphQL-coerced input. `z.object()` strips, and `ClimbInputSchema` never declared `is_no_match` or `characteristics` — so the server was already erasing both on every full-queue sync. Proved with a throwaway probe before fixing.

### Commit ordering is a correctness property, not tidiness

Please don't squash. The invariant is that **no commit leaves a field on a write path whose matching read path lacks it**, and it only holds in sequence:

1. `mirrored` into `climbToQueueItem` — complete wire path already, safe alone.
2. Zod + web outbound gain `is_no_match` / `characteristics` — server and web-outbound only, no client selection set involved.
3. **Read paths** — the four into both selection sets and mobile's rebuild. Closes web's existing flap; mobile unchanged.
4. **Write path** — the four into `climbToQueueItem`, `buildProvisionalClimb` populated, tripwire flipped.

Landing 4 before 3 would briefly make mobile flap too.

### Mixed-version fleet during rollout

Stated plainly, because it is not eliminable client-side — selection sets live in the client bundle.

- **Web deploys atomically**, so web↔web is fixed immediately.
- **Backend deploys once**; the Zod widening is purely additive.
- **Mobile ships by OTA.** In a session containing an old mobile peer, that peer rebuilds items without the four and its next full-queue write pushes nulls back, so newer peers can see the fields appear and then vanish. That *is* a flap and I'm not claiming otherwise. Accepted because the degraded state is exactly today's behaviour rather than corruption, none of the four gates a destructive action (an Edit button and a draft badge), and the stale-bundle tail is ~15–20 users over 14 days. `mirrored` is immune — its wire path is already complete on old bundles — and it's the only field the reducer reads. `is_no_match` / `characteristics` are immune too, since every deployed version already selects them.
- **Considered and rejected:** server-side null-coalescing on `setQueue` so an omitted field can't clear a known value. It kills the flap permanently, but it would pin `is_draft: true` forever after publication if the last writer omitted it, and it must never touch `mirrored` (a legitimately clearable per-slot toggle). Trading a transient absence for a permanent wrong state is a bad deal. Please don't re-propose it without new evidence.

### Drift guards

A regression test over today's five fields would not have stopped the next drift, so every field list is read from a **live source** and compared by set equality. Adding a sixth field to one side only turns something red with no test edit:

- `climbToQueueItem` keys **==** `toClimbInput` keys — the sharpest one, no allowlist. Exactly the two functions that drifted apart to produce this issue.
- `toClimbQueueItem` keys **==** `SUBSCRIPTION_CLIMB_FIELDS`.
- `CLIMB_FIELDS` **==** `SUBSCRIPTION_CLIMB_FIELDS` **==** `ClimbInput` − `{userAscents, userAttempts}`, anchored to the schema rather than to each other (comparing the read paths only to each other would have stayed green through this bug — both selection sets and `climbToQueueItem` agreed with each other, and all three disagreed with `toClimbInput`).
- Zod `ClimbInputSchema` shape **==** `ClimbInput`, and web's `toClimbQueueItemInput` keys **==** `ClimbInput`.

`userAscents` / `userAttempts` are the single allowlisted exception, with a comment explaining both why (they are your own tick counts — broadcasting them would show every peer your send count on their queue row) and what it costs (they behave like the bug this file guards against). Please don't "fix" that asymmetry by adding them to the selection sets.

### Known trade-off

`description` on the read paths grows FullSync payloads. Upload is unchanged — web already sent it — only broadcast download grows, bounded by the 2000-char Zod cap. Taken anyway because `use-create-climb-navigation` seeds the editor from `climb.description`, so forking or editing a queued climb was losing it.

## Release Notes

- The Edit button now shows up on your own climbs while they're sitting in the queue, not just in search
- Drafts keep their draft state through the queue, so you can jump straight back in and keep working
- Mirrored climbs stay mirrored when you preview them, and no-match tags stop vanishing after a queue sync

## Test plan

- `vp run typecheck` — clean
- `vp check` — 4675 files correctly formatted
- `vp run codegen` — no drift (the queue-session operations are plain template strings, so codegen doesn't consume them)
- `vp run test:mobile` — pass
- `vp test run --project backend` — queue suites plus the 221 existing operation-validation tests, all green
- `vp test run --project web` — persistent-session + queue-control
- **Every new guard mutation-tested by deleting the field it protects.** Removing `is_draft` from `climbToQueueItem` fails the key-set equality test; removing `is_no_match` from web's input fails both new web tests.
- Two web render tests and two mobile screen tests flaked under parallel load on my box (5s render timeouts, `import 491s`). Each passes in isolation, and `queue-control-bar-tick-overlay` was verified green on an unmodified `origin/main` worktree and then green again on this branch once the machine settled.

Not verified: a real two-device party session. The peer round-trip is the failure mode no unit test can see, and `mirrored` is worth exercising specifically since it's the field the reducer reads.

### DIFF
--- a/docs/websocket-implementation.md
+++ b/docs/websocket-implementation.md
@@ -128,6 +128,14 @@ The queue-action surface itself (add/remove/set-current/navigate/mirror/replace/
 
 Party mode deliberately does **not** clear the queue on navigation (`setBoardContext` early-returns while a session is active), and the reducer applies peer `CurrentClimbChanged` broadcasts unconditionally — so a member connected to a different wall can legitimately receive a current climb their board can't light. Since issue #3193, both platforms' BLE auto-senders classify the current climb against the active board (`classifyClimbBoardCompatibility` in `@boardsesh/board-config`, boardType + layoutId only; missing metadata never blocks) and skip a known mismatch instead of dark-firing the wall. The invariant: **solo** advances the local queue to the next compatible item (or clears the wall); **party** only clears its own wall and never advances the shared current climb — the session state every peer sees is untouched, and members on the right wall keep climbing. Climb identity rides the wire via `boardType`/`layoutId` on `ClimbInput` (`toClimbQueueItemInput` on web, `toClimbInput` on mobile).
 
+### Queue climb field contract (why fields "flap")
+
+A queue climb crosses several independently-maintained field lists: the write paths (`toClimbInput` on mobile, `toClimbQueueItemInput` on web), the backend's Zod `ClimbInputSchema`, and the read paths (`CLIMB_FIELDS` in `@boardsesh/graphql`, `SUBSCRIPTION_CLIMB_FIELDS` on mobile, plus mobile's `toClimbQueueItem` rebuild). When one drifts, the field does not simply go missing — it **flaps**: a peer whose read path omits it rebuilds the item without it, and that peer's next full-queue write (`setQueue` / `joinSession`) pushes the gap back to everyone, so the originator loses it on the following FullSync. A flapping field is worse than a consistently absent one.
+
+Two non-obvious narrowing points bit us in #3927: `climbToQueueItem` hand-picked a narrower set than `toClimbInput` sent, and `setQueue` / `joinSession` persist the **parsed** Zod output (the single-item mutations keep the GraphQL-coerced input instead), so any field missing from `ClimbInputSchema` is silently stripped server-side on a full-queue sync.
+
+The contract is enforced by `packages/backend/src/__tests__/queue-climb-field-contract.test.ts`, which reads every list from its live source and compares by set equality against the GraphQL `ClimbInput` type. `userAscents` / `userAttempts` are the only deliberate exception — they are per-user tick counts and must never be broadcast, at the cost of behaving exactly like the bug the test guards against.
+
 ## Technology Stack
 
 | Component          | Technology                        | Purpose                                                       |
--- a/packages/backend/src/__tests__/queue-climb-field-contract.test.ts
+++ b/packages/backend/src/__tests__/queue-climb-field-contract.test.ts
@@ -1,6 +1,8 @@
 import { describe, expect, it } from 'vite-plus/test';
+import { readFileSync } from 'fs';
 import { parse, visit } from 'graphql';
 import { typeDefs } from '@boardsesh/shared-schema';
+import { QUEUE_UPDATES } from '@boardsesh/graphql/operations/queue-session';
 import { ClimbInputSchema, ClimbQueueItemSchema } from '../validation/schemas/climbs';
 
 /**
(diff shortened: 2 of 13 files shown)


marcodejongh:
Thanks — two of the three are fixed in cce1826.

**Fixed: whitespace-split drift guard.** Fair catch, and the sharper framing is that a drift guard which can silently stop guarding is worse than no guard. `queue-conversion.test.ts` now parses with `graphql.parse` + `visit` and reads `alias?.value ?? name.value`, mirroring `climbSelectionFieldNames` in the backend contract test.

**Fixed: docs gap.** `docs/websocket-implementation.md` gains a "Queue climb field contract" section covering why fields flap, the two non-obvious narrowing points (`climbToQueueItem` vs `toClimbInput`, and `setQueue`/`joinSession` persisting the *parsed* Zod output while single-item mutations keep the coerced input), and a pointer to the enforcing test.

**Also noted:** the `userId`-before-profile-resolves window is real. It's the same pre-existing dependency as `setter_username` on the line above, self-correcting on the next save, so I've documented it in place rather than adding a queue-item update path for it.

**Declining: comment verbosity.** The quoted rule doesn't exist. There is exactly one `CLAUDE.md` in this repo and it contains no guidance on comment length or multi-line blocks:

```
$ find . -name CLAUDE.md -not -path "*/node_modules/*"
./CLAUDE.md
$ grep -in "multi-line comment\|one short line max\|comment block" CLAUDE.md
(no matches)
```

The house style in these exact files is already multi-line: `toClimbInput`'s docstring (10 lines), `queue-conversion.ts`'s type docstring (11), `ClimbInputSchema`'s per-field notes, and `buildProvisionalClimb`'s comment were all multi-line before this PR. The new comments match what was there.

On "the rationale belongs in the commit message, not inline" — it's in both, deliberately. The comments this PR *replaces* were instructions to keep the boundary narrow ("deliberately NOT carried here yet", "would be dead weight until that boundary and both subscription selection sets widen together"). A stale comment telling the next person to undo the fix is a live hazard, so those had to be rewritten in place rather than deleted, and the replacement has to explain the flap or it reads as arbitrary. Same for the `characteristics` Zod field: without the inline note that `parseArrayTolerant` drops the whole queue slot on a validation failure, `z.string()` where an enum exists looks like an oversight and someone will "tighten" it into silent queue-slot deletion (#3857's failure mode). Commit messages don't show up in the editor when you're about to make that change.

Happy to trim if a maintainer confirms a comment-length rule I've missed.

Issue #3943, opened 2026-07-26
board_climb_stats orphans can still be created by deleting a climb that already has stats


marcodejongh:
Split out of #3528.

#3528's fix guards the tick recompute's defensive seed on the climb existing in
`board_climbs`, so no new orphan `board_climb_stats` row can be *minted* by a tick.
It is forward-only and it only covers the mint path.

Three paths still create orphans the other way round — by deleting a climb that
already has a stats row:

- `deleteDraftClimb` — `packages/backend/src/graphql/resolvers/climbs/mutations.ts:772`.
  Reachable: `saveClimb` skips the stats seed for drafts, but a user ticking their
  own draft seeds one via the recompute, and deleting the draft then strands it.
- account deletion — `packages/backend/src/graphql/resolvers/users/mutations.ts:211`
  deletes the user's climbs.
- `clear-aurora-board.ts:201` — admin/board reset.

`board_climb_stats` has no FK to `board_climbs` on purpose (stats legitimately
arrive before climbs during upstream sync — see the note at
`packages/db/src/schema/boards/unified.ts:669`), so this can't be fixed with
`ON DELETE CASCADE` without breaking the sync path. Each delete site needs to
clear the matching stats rows itself, or a periodic sweep needs to.

Worth sizing against how many of the 77 known prod orphans came from here versus
from the (now-guarded) tick seed — that split decides whether this is worth
per-site cleanup or a sweep.

Note: cleaning up the existing 77 prod rows is a separate, human-signed-off
migration and is not this issue either.

marcodejongh:
🤖 Claimed by the 2026-07-31 bug-fix campaign (Fable session `fablebugfix`, batch 3). Assigned to avoid duplicate work across concurrent agent campaigns — a PR from this session will follow. If another agent needs this issue, comment here first.

marcodejongh:
🤖 Claimed by bug-batch loop `claude-bb21` (batch 2, session c321d1b9, 2026-08-11). Where a 2026-07-31 `fablebugfix` claim exists it has lapsed (10+ days, no PR/branch) — reclaiming per protocol. A PR from this loop will follow; parallel loops please pick another issue.

Issue #3942, opened 2026-07-26
saveTick: flip the climb-existence check from log-only to a rejection


marcodejongh:
Follow-up to the #3528 fix.

`saveTick` now resolves every incoming `climbUuid` against the climb catalog
(`resolveClimbCatalogPresence` in `packages/backend/src/db/queries/climbs/climb-catalog-presence.ts`)
and, when the UUID is in neither `board_climbs` nor `board_climb_aliases`, logs a
warning and emits the `Tick Climb Not In Catalog` PostHog event — then **saves the
tick anyway**.

Log-only was a deliberate call. The analysis said every client-held UUID
originates from `board_climbs`, so a rejection should never fire on a real send:

- search / queue / playlists / the offline SQLite snapshot are all projections of `board_climbs`
- `saveClimb` server-generates the UUID and writes `board_climbs` in the same transaction that returns it
- the offline outbox whitelist has no `SaveClimb`, so there's no queued-tick-before-queued-climb ordering hazard
- deduped-away Kilter UUIDs are covered by the alias arm of the predicate

But the payoff is asymmetric: being right buys us a handful of unrenderable junk
rows, being wrong about one path costs a climber a send (a non-retryable GraphQL
error dead-letters immediately in the offline drainer). So we ship the observation
first and take the bet only once we have data.

## Done when

- `Tick Climb Not In Catalog` has held at **zero distinct users** across a real
  observation window (a few weeks, spanning at least one mobile release so an
  older client version can't be hiding the signal).
- The check throws a `GraphQLError` with `extensions.code = 'CLIMB_NOT_FOUND'`
  instead of warning, and the tick is not written.
- The alias arm of the predicate stays. It is what keeps a deduped-Kilter tick
  from reading as a false positive.

If the counter is **not** zero, that's the more interesting outcome: the hits are
a path the analysis missed, and they should be understood before anything is
rejected.

marcodejongh:
🤖 Claimed by the 2026-07-31 bug-fix campaign (Fable session `fablebugfix`, batch 1). Assigned to avoid duplicate work across concurrent agent campaigns — a PR from this session will follow. If another agent needs this issue, comment here first.

marcodejongh:
🤖 Campaign triage (2026-07-31, Fable session `fablebugfix`): **holding — the Done-when isn't met yet.**

Measured observation window as of today:
[...]

PR #3959, opened 2026-07-26
fix(db): stop tick recomputes minting stats rows for climbs that don't exist

### LINKED ISSUE #3528: saveTick accepts any well-formed climb UUID — defensive seed creates stats rows for nonexistent climbs (77 orphan rows in prod)
## Summary
SaveTickInputSchema validates climbUuid by format only (packages/backend/src/validation/schemas/ticks.ts:21); saveTick never checks board_climbs. The recompute's defensive seed (recompute-climb-stats.ts:91-107) then inserts a board_climb_stats row for whatever (boardType, climbUuid, angle) arrived, so a stale/deleted/garbage climb UUID mints a permanent orphan stats row; the owner CTE finds no board_climbs row and the conservative branch still writes counts/FA into it.

## Failure scenario
A mobile client replays a tick for a climb that was deleted (or a corrupted UUID); a stats row with ascensionist_count>0 and an FA name exists for a climb that doesn't — polluting any aggregate that joins stats without an inner join to board_climbs.

## Evidence
`packages/backend/src/graphql/resolvers/ticks/recompute-climb-stats.ts:91`

```
Prod: SELECT s.board_type, COUNT(*) FROM board_climb_stats s WHERE NOT EXISTS (SELECT 1 FROM board_climbs bc WHERE bc.board_type=s.board_type AND bc.uuid=s.climb_uuid) GROUP BY 1 -> kilter 73 (71 with counts>0), moonboard 2, tension 2.
```

## Suggested fix
Validate climb existence in saveTick (cheap indexed lookup, also enables alias-canonicalisation at write time), and delete the 77 orphan rows.

marcodejongh:
## Summary

`recomputeClimbStats`' defensive seed was an unconditional `INSERT ... ON CONFLICT DO NOTHING`, so a tick at **any** `climb_uuid` minted a permanent `board_climb_stats` row for that `(board_type, uuid, angle)` — and the non-owned branch then wrote real ascent counts into it. Nothing upstream could stop it: `SaveTickInputSchema.climbUuid` is `ExternalUUIDSchema`, which is `z.string().min(1).max(50)` (not even a UUID shape), `saveTick` never looks the climb up, and `boardsesh_ticks` has no FK to `board_climbs`. 77 orphan rows in prod, 71 of them carrying counts.

Both seeds — single-key (`recompute.ts`) and bulk — now insert only when `board_climbs` has a row for the climb. The single-key seed makes the catalog lookup the guard itself (`INSERT ... SELECT ... FROM board_climbs bc WHERE bc.uuid = …`); the bulk seed adds `WHERE EXISTS`. Both ride `board_climbs_pkey`, which already exists — no new index, and the recompute's `owner` CTE was already doing the same lookup.

Fixing it at the seed rather than at `saveTick` is what makes it complete: one guard covers every tick writer — `saveTick`, `updateTick`, `deleteTick`, the legacy web `saveAscent` proxy, kilter-sync `user-sync`, aurora-sync `apply-user-logbook` and `json-import`, `moonboard-import`, and the aurora board-import backfill.

**Guarded on the CLIMB, not on `(climb, angle)`.** A tick at an angle the catalog has no stats row for is legitimate and must still seed — that is the entire reason the seed exists (`recompute.ts` bulk-seed comment). There's a test that passes today and fails only if someone "hardens" this, with a comment saying so, because it will otherwise read as redundant and get deleted.

Seeded rows also now carry `quality_normalized = TRUE`. Every value such a row can ever hold is on the canonical 1-5 scale: the Boardsesh side of the blend comes from native ticks (Zod-validated 1-5), and every writer of the upstream side — aurora `shared-sync`, kilter `catalog-sync` / `stats-repair`, all three MoonBoard importers — sets the flag `true` on insert *and* on conflict. The column default of `FALSE` was simply wrong for a seeded row, and it is what stranded 44 MoonBoard rows outside the "all normalized" invariant.

### The `saveTick` check is log-only, on purpose

`saveTick` now resolves each incoming `climbUuid` against `board_climbs` **and** `board_climb_aliases` and reports a UUID that neither knows — a warning log plus a `Tick Climb Not In Catalog` PostHog event keyed on the user, so the metric counts climbers rather than events. It does **not** reject.

The analysis says a rejection would never fire on a real send: every client-held UUID originates from `board_climbs` (search, queue, playlists and the offline snapshot are all projections of it), `saveClimb` server-generates the UUID and writes `board_climbs` in the same transaction that returns it, and the offline outbox whitelist has no `SaveClimb`, so there's no queued-tick-before-queued-climb ordering hazard. But the payoff is asymmetric — a non-retryable GraphQL error dead-letters immediately in the offline drainer, so being right buys a handful of unrenderable rows and being wrong costs a climber a send. So this ships the measurement and #3942 takes the bet later, once the counter has held at zero over a real observation window.

The alias arm is load-bearing for that, not decoration: the Kilter dedup path folds duplicate catalog UUIDs into `board_climb_aliases` and deliberately writes **no** `board_climbs` row for them (`getLogbook` already canonicalises these for display). Without the alias arm every deduped-Kilter tick would read as a hit and make the observation window unanswerable. 73 of the 77 prod orphans are Kilter — the only board with dedup — so this is likely most of them.

### Not in scope

- **The 77 existing prod rows.** Untouched. No data migration in this PR; cleaning them up is a separate, signed-off migration.
- **#3529's 42 wrong-angle MoonBoard rows.** Those climbs *do* have a `board_climbs` row, so the existence guard passes and still seeds at the ticked angle — correctly, because that is an **identity** problem, not an existence one. It dissolves when #3851 makes MoonBoard climbs angle-agnostic. This PR closes only the seed half of #3529 (the `quality_normalized` flag), so #3529 stays open.
- **#3943** covers the other direction orphans appear from: deleting a climb that already had stats (`deleteDraftClimb`, account deletion, `clear-aurora-board`). A seed guard cannot catch those.

No migration — this is SQL logic and a resolver check, no schema change.

Closes #3528

## Release Notes

Invisible to users: the phantom rows were never rendered on any surface.

## Test plan

`packages/db` has no vitest project and nothing in CI runs its `tsx --test` suite, so a SQL-text assertion over there is a smoke test, not evidence. Every load-bearing assertion is a real-Postgres test in `packages/backend/src/__tests__/`, extending the existing `recomputeClimbStats — provenance matrix (real DB)` block.

**Mutation-tested — I deleted each guard and confirmed the right tests go red:**

| Mutation | Tests that failed |
|---|---|
| Remove `FROM board_climbs` / `WHERE EXISTS` from both seeds | `single-key: a tick on a climb missing from board_climbs seeds NO stats row`, `bulk: a phantom key in a chunk seeds nothing and does not disturb the real key beside it`, `seeds via INSERT ... SELECT FROM board_climbs so the catalog lookup guards it` |
| Remove the alias arm from `resolveClimbCatalogPresence` | `does not fire the counter for a deduped-away UUID that only board_climb_aliases knows` |

The over-guarding tripwire (`a real climb ticked at an angle with no stats row STILL seeds one`) correctly stayed green through both — it is there to fail on a *tightening*, not a removal.

Also covered: the bulk case asserts the real key beside the phantom one in the same chunk still counts correctly, so the guard can't quietly drop good keys; the single-key case asserts a phantom returns `undefined` rather than throwing, since the debounced publisher logs any throw as an error.

Commands:

- `vp test run --project backend recompute-climb-stats save-tick-climb-catalog` — 40 passed
- `vp test run --project backend tick` — 95 passed
- `vp test run --project backend board-presence` — 82 passed (the file with the most `saveTick` call sites)
- `vp run typecheck` — clean
- `vp check` — clean for every file in this diff

Note on the test churn: the plan expected to have to seed `board_climbs` in several existing `saveTick`-driving tests. The log-only ruling removed that entirely — no existing test needed weakening or fixture changes. The only test edits here are the mocked seed-shape assertions in `recompute-climb-stats.test.ts`, which had to follow the seed from `.values()` to `INSERT ... SELECT`.

`.insert().select()` was tried first to keep this in Drizzle; drizzle-orm 0.45.2 rejects a partial column list at runtime ("selected fields are not the same or are in a different order compared to the table definition"), which would force enumerating every column of `board_climb_stats` forever. The raw `sql` is justified in a comment at the call site.

### DIFF
--- a/packages/backend/src/__tests__/recompute-climb-stats.test.ts
+++ b/packages/backend/src/__tests__/recompute-climb-stats.test.ts
@@ -30,43 +30,38 @@ describe('recomputeClimbStats', () => {
     vi.clearAllMocks();
   });
 
-  it('seeds the stats row with explicit upstream=0 and boardsesh=0', async () => {
-    let capturedSeedValues: unknown = null;
+  // The seed's behaviour is asserted end to end against real Postgres in the
+  // provenance-matrix block below (a phantom key seeds nothing; a real climb at
+  // a new angle still does; quality_normalized). What this one pins is the
+  // SHAPE: the seed must stay an INSERT ... SELECT FROM board_climbs so the
+  // catalog lookup IS the guard (#3528). A regression to a plain VALUES insert
+  // — which cannot carry a WHERE — silently un-guards it.
+  it('seeds via INSERT ... SELECT FROM board_climbs so the catalog lookup guards it', async () => {
+    const executed: unknown[] = [];
     mockDb.transaction.mockImplementation(async (callback: (tx: unknown) => Promise<void>) => {
-      const insertChain = {
-        onConflictDoNothing: vi.fn(),
-      };
       const tx = {
-        insert: vi.fn(() => ({
-          values: vi.fn((values: unknown) => {
-            capturedSeedValues = values;
-            return insertChain;
-          }),
-        })),
-        execute: vi.fn(async () => []),
+        execute: vi.fn(async (query: unknown) => {
+          executed.push(query);
+          return [];
+        }),
       };
       await callback(tx);
     });
 
     await recomputeClimbStats('kilter', 'CLIMB-1', 40);
 
-    expect(capturedSeedValues).toMatchObject({
-      boardType: 'kilter',
-      climbUuid: 'CLIMB-1',
-      angle: 40,
-      ascensionistCount: 0,
-      upstreamAscensionistCount: 0,
-      boardseshAscensionistCount: 0,
-    });
+    const seedSql = sqlText(executed[0]);
+    expect(seedSql).toContain('INSERT INTO board_climb_stats');
+    expect(seedSql).toMatch(/SELECT[\s\S]*FROM board_climbs bc/);
+    expect(seedSql).toMatch(/WHERE bc\.uuid =/);
+    expect(seedSql).toMatch(/AND bc\.board_type =/);
+    expect(seedSql).toContain('quality_normalized');
   });
 
-  it('runs the recompute SQL inside a single transaction', async () => {
+  it('runs the seed + recompute SQL inside a single transaction', async () => {
     let executeCount = 0;
     mockDb.transaction.mockImplementation(async (callback: (tx: unknown) => Promise<void>) => {
       const tx = {
-        insert: vi.fn(() => ({
-          values: vi.fn(() => ({ onConflictDoNothing: vi.fn() })),
-        })),
         execute: vi.fn(async () => {
           executeCount += 1;
           return [];
@@ -78,7 +73,8 @@ describe('recomputeClimbStats', () => {
     await recomputeClimbStats('kilter', 'CLIMB-1', 40);
 
     expect(mockDb.transaction).toHaveBeenCalledTimes(1);
-    expect(executeCount).toBe(1);
+    // Seed + the aggregate recompute.
+    expect(executeCount).toBe(2);
   });
 
   // FRAGILE TEST WARNING — read before changing:
@@ -104,9 +100,6 @@ describe('recomputeClimbStats', () => {
     let capturedQuery: unknown = null;
     mockDb.transaction.mockImplementation(async (callback: (tx: unknown) => Promise<void>) => {
       const tx = {
-        insert: vi.fn(() => ({
-          values: vi.fn(() => ({ onConflictDoNothing: vi.fn() })),
-        })),
         execute: vi.fn(async (query: unknown) => {
           capturedQuery = query;
           return [];
@@ -149,9 +142,6 @@ describe('recomputeClimbStats', () => {
     let capturedQuery: unknown = null;
     mockDb.transaction.mockImplementation(async (callback: (tx: unknown) => Promise<void>) => {
       const tx = {
-        insert: vi.fn(() => ({
-          values: vi.fn(() => ({ onConflictDoNothing: vi.fn() })),
-        })),
         execute: vi.fn(async (query: unknown) => {
           capturedQuery = query;
           return [];
@@ -193,9 +183,6 @@ describe('recomputeClimbStats', () => {
     const loggerSpy = vi.spyOn(logger, 'info').mockImplementation(() => logger);
     mockDb.transaction.mockImplementation(async (callback: (tx: unknown) => Promise<void>) => {
       const tx = {
-        insert: vi.fn(() => ({
-          values: vi.fn(() => ({ onConflictDoNothing: vi.fn() })),
-        })),
         // The combined WITH … RETURNING query returns one diff row.
         execute: vi.fn(async () => [
           {
@@ -226,9 +213,6 @@ describe('recomputeClimbStats', () => {
     const loggerSpy = vi.spyOn(logger, 'info').mockImplementation(() => logger);
     mockDb.transaction.mockImplementation(async (callback: (tx: unknown) => Promise<void>) => {
       const tx = {
-        insert: vi.fn(() => ({
-          values: vi.fn(() => ({ onConflictDoNothing: vi.fn() })),
-        })),
         execute: vi.fn(async () => []),
       };
       await callback(tx);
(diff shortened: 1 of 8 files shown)


marcodejongh:
Good catch on the serial round-trip — fixed in 27d2e2659, though not by fire-and-forget.

The probe now *starts* before the board/session resolution and is *settled* after the insert transaction, so it overlaps round-trips `saveTick` already makes and costs no extra wall-clock. Same latency win as `void`, without the cost that made me await it in the first place: settling it keeps the counter deterministic, so the tests can assert the event fired — and, more importantly, that it *didn't* fire for a deduped-away alias UUID. Proving a negative against a fire-and-forget promise means polling or sleeping, and that assertion is exactly the one #3942's decision rests on, so it can't be flaky.

It also means no promise outlives the request. The helper self-catches, so awaiting it still cannot fail a tick.

marcodejongh:
Both fixed in 715c47ce3 — and the replica one was the more important of the two, so thanks for catching it.

**Replica lag.** Switched to the primary. You're right that the impact isn't the tick (that saves either way), it's the counter: a climb published seconds before the tick can still be in flight to a replica, and a lagging replica reports a real climb as `unknown`. That false positive is indistinguishable from a real one in the dashboard, and #3942's entire question is "has this held at zero?" — a counter that drifts off zero for reasons that aren't real is worse than no counter, because it can't be cleared without a forensic dig. One PK probe per tick on the primary is a cheap price, and after 27d2e2659 the call site overlaps it with writes that hit the primary anyway, so it costs no wall-clock either.

**`climbUuid` on the event.** Added. Same reasoning: "12 hits" is unactionable until you know whether that's 12 phantom climbs or one client looping on one bad UUID, and that's precisely the distinction the follow-up has to make. Not PII — a catalog identifier, and the same value the warn log already carried.

The test now asserts the UUID rides along, so a future property trim can't quietly blind the counter.

marcodejongh:
Fixed 1 and 3 in 72e114fd0. Pushing back on 2.

**1 — unawaited on exception paths.** You're right, and the comment was the thing that was wrong, not the code. The behaviour is intended: the counter measures ticks that get *saved*, so a tick that failed on a beta-link rejection or a transaction error shouldn't move it. Abandoning the promise there is correct, and since the helper never rejects, an abandoned promise is inert. The comment now says that instead of claiming nothing outlives the request.

**3 — `board_climbs_pkey` doc.** Also right that it read as if `board_type` were part of the key. It isn't, and the predicate isn't redundant either — it's a correctness filter rather than an index hint. A tick names a `(boardType, climbUuid)` **pair**, and `board_climb_stats` is keyed on that pair, so a Kilter UUID arriving under `boardType: 'tension'` has to resolve `unknown` rather than quietly matching the Kilter row. The recompute's seed guard and `owner` CTE filter both columns for the same reason. Now explained in the docblock.

**2 — module-level `db` import: leaving as is.** This one I'd rather not change, because it would make the file the odd one out. Every sibling in `packages/backend/src/db/queries/climbs/` imports the client at module level — `get-climb.ts` and `match-climb-by-frames.ts` both `import { db } from '../../client'`, `count-climbs.ts` imports `dbRead`. Injection is the convention one level up, in `@boardsesh/db/queries` (`recomputeClimbStats` takes a `DrizzleDb`), and for a good reason: those helpers are called by the backend, the sync daemons, and standalone scripts, each with a different handle. This one is backend-only with a single call site, so it follows the backend convention. Introducing a second pattern here would cost consistency without buying testability we don't already have — the test drives it against the worker DB and exercises the real query, which is stronger evidence than a stub would be.

marcodejongh:
All three addressed in bf4ef3c78 — and on the first one you caught me contradicting myself, so thanks.

**False positives on failed ticks.** My previous comment claimed a failed tick can't move the counter. It can: the event fires from inside the helper the moment the probe resolves, independent of whether the tick goes on to commit. Rather than restructure to suppress that, I've made it the documented reading, because I think it's the right one — what's being counted is *"a client sent a `climb_uuid` the catalog has never heard of"*, and that's equally true whether the tick then failed on a beta-link rejection or a constraint violation. A UUID we can't resolve is worth knowing about either way. So #3942 should read a hit as "this client sent an unknown UUID", not "an unknown-UUID tick was saved", and the comment now says exactly that.

**Board-type mismatch untested.** Added, and it's a good catch — that predicate is precisely the kind of thing a future cleanup deletes as redundant, since `board_climbs.uuid` alone is the PK. Mutation-checked: dropping `eq(boardClimbs.boardType, boardType)` turns `fires the counter for a UUID that exists only under a DIFFERENT board type` red, and nothing else.

**SQL-text regex fragility.** Fair flag for readers, but that test does earn its place — it was one of the three that went red when I deleted the seed guard during mutation testing (table in the PR description). The whitespace-tolerance concern is real in principle, which is why the structural claim is carried by the real-Postgres tests and this one only guards the `.values()` → `INSERT ... SELECT` shape regression that would silently un-guard the seed.

PR #3960, opened 2026-07-26
fix(sync): stop one unwritable logbook row wedging a user's whole Aurora sync

### LINKED ISSUE #3871: Aurora sync: DB-level write failure in a bids/ascents chunk still rolls back the whole cross-table sync transaction

marcodejongh:
## Summary

One upstream logbook row that Postgres refuses used to freeze a user's entire Aurora sync, permanently and silently. Closes #3871.

`applyLogbookChunk` issues three **batched** writes per ~100 rows — a claims UPDATE, a `jsonb_to_recordset` UPDATE, and an INSERT — with no try/catch. Nothing between there and `syncUserData` catches either, and `syncUserData` wraps **every** table in one `db.transaction`. `USER_TABLES` order is `users, walls, wall_expungements, draft_climbs, ascents, bids, tags, circuits`, so a single bad **ascent** rolled back bids, tags, circuits **and** the sync checkpoint together. The next cycle re-fetched the same row and re-crashed on it. Forever, with nothing surfaced to the user.

This is the DB-execution-time sibling of the parse-time wedge PR #3872 fixed for issue #3520, and it gets that PR's skip-and-log philosophy one layer down.

### Three mechanisms, deliberately independent

**1. Savepoint + row-by-row replay.** One nested `db.transaction()` (a real `SAVEPOINT` on postgres-js) wraps a chunk's three batched writes. On failure it rolls back — leaving the caller's transaction healthy — and replays all three one row at a time, each in its own savepoint. Only the offending row is lost.

**2. Pre-write validation** of the four `integer` columns feeding the `jsonb_to_recordset` UPDATE.

> ⚠️ **This is not redundant with the fallback, and a future refactor should not delete it as such.** `angle` also flows into `recomputeClimbStatsBulk`'s own `jsonb_to_recordset(… angle integer)` + `INSERT INTO board_climb_stats`, which runs **after** the chunk loop and **outside** any write fallback. A `NaN` angle aborts the transaction at that second site no matter how good the row-by-row replay is. Validation is the only guard there.

| column | shape | failure without the guard | rule |
| --- | --- | --- | --- |
| `angle` | `integer NOT NULL` | `NaN` → `JSON.stringify` emits `null` → NOT NULL violation; `"NaN"` on the INSERT path; **plus** the recompute abort above | **drop the row** — no honest default |
| `attempt_count` | `integer NOT NULL DEFAULT 1` | same | coerce to `1` |
| `quality` | `integer` + CHECK `1..5` | safe today; a future star-scale change re-poisons it silently | coerce to `NULL` |
| `difficulty` | `integer` | `toNumberOrNull` guarantees finite but **not integral** — `21.5` is `invalid input syntax for type integer` | coerce to `NULL` |

Plus two same-class deterministic chunk-killers: a NUL byte / unpaired surrogate (rejected by `::jsonb` and by the text bind path) is stripped from `comment` and drops the row only if it corrupts an identity column; and a duplicate `aurora_id` **within one payload** is collapsed last-wins before it can collide on `boardsesh_ticks_aurora_id_unique` (deduped at payload level, since a chunk boundary between the two copies would hide it from a per-chunk guard).

Principle: **coerce what's safely coercible, drop only what's unusable.** A send with a garbage star rating still shows up in the logbook.

**3. `logbook_sync_skips`** (new table). **The quarantine is the reason a drop is allowed.** Dropping a row means a climber loses a send, and that is only defensible because the normalized payload is kept — the send is quarantined and replayable, not gone. Take the table away and mechanism 2 becomes silently discarding user data, which would not be acceptable.

`reason` is a pgEnum (`logbook_sync_skip_reason`), not free `text` — the reason codes are the grouping key of the triage query, so one off-contract string would split a cluster and quietly under-report it.

It is **state, not a counter**: a row is deleted the moment that `aurora_id` syncs cleanly, so the table always answers "what is broken right now". A per-cycle counter would report once and then clear itself on the next cycle, when Aurora's incremental delta is empty and nothing was skipped because nothing arrived — exactly how #3523's equivalent stayed hidden for years. (Same reasoning #3931 gives for making its duplicate-account check a state query rather than a per-cycle count.)

### Checkpoint advances past a skipped row

Deliberately. Not advancing **is** the wedge — a row Postgres will never accept cannot be fixed by retrying it forever. A skipped row returns when the user edits it upstream, when the watermark resets (`clearAuroraBoardData` drops `board_user_syncs`), or via the JSON export import; and `payload` makes an offline replay possible meanwhile.

### Scope

The single cross-table transaction is **untouched**. That is the right boundary, not a compromise: the data write and the checkpoint advance must commit together, or a crash between them silently drops a page of ascents. Containing the abort at the savepoint fixes the blast radius without breaking that invariant — and as a result **neither `user-sync.ts` needed changing**. Both the daemon and the web cron path share the module, so both are fixed by the same diff.

### Migration numbering — resolved, the DDL now lives in #3963

The provisional-`0187` warning that used to sit here has been actioned. Four PRs were all holding `0187` against a `main` that tops out at `0186`, so the four schema changes were collapsed into **one** regenerated migration on `chore/collapse-0187-migrations` (#3963) — never hand-renumbered, since the journal has to stay strictly ordered (a gap makes `migrate.ts` only warn and then silently skip earlier migrations, #2933).

This PR is now based on #3963 and carries no migration: the `logbook_sync_skip_reason` enum, the `logbook_sync_skips` table, its FK to `users` and its three indexes all moved there unchanged, along with `packages/db/src/schema/app/logbook-sync-skips.ts`. What stays here is the savepoint replay, the reconcile and the quarantine writes. GitHub retargets this to `main` when #3963 merges.

## Release Notes

- Your logbook keeps syncing even when one old ascent is broken beyond repair — Boardsesh sets that one aside and carries on instead of quietly freezing every send, attempt and circuit behind it.

## Test plan

- [x] 14 new tests in `packages/aurora-sync/src/sync/apply-user-logbook.test.ts`; full project green (117 tests, up from 103). The tx shim gained a `transaction()` that reproduces postgres-js savepoint semantics (rollback, then rethrow), and quarantine writes are tracked under their own call kinds so they can't inflate the tick-write counts the existing suite asserts on.
- [x] **Mutation-tested — each mechanism is independently observable:**
  - delete the row-by-row replay → **5 fail** (zero rows land instead of 2, and `applyAuroraAscents` rejects)
  - delete the `angle` validation → **3 fail**, and *the isolation tests stay green* — proving neither mechanism silently substitutes for the other
  - delete the quarantine write → **7 fail**
- [x] Behaviour exercised against a **throwaway `postgres:17` container on port 55871** (never the ambient `DB_URL`, never `db:migrate`/`dev-db-up.sh`): the `ON CONFLICT` upsert bumps `seen_count` while preserving `first_seen_at`, the reconcile `DELETE` clears the row, the FK cascades on user deletion, the triage query from the docs runs, all four reason codes are accepted and an off-contract reason is **refused** by the enum. Container removed afterwards.
- [x] The DDL itself is verified in #3963 — applied, idempotent on re-run, no stray `DROP`, and the known `boardsesh_ticks_quality_range` CHECK footgun did not fire.
- [x] `vp run typecheck` (whole repo) — clean
- [x] `vp check` — clean (`All 4675 files are correctly formatted`; remaining lint warnings are pre-existing, in `packages/mobile`)
- [x] `vp test run --project aurora-sync --project kilter-sync --project sync-runtime` — 23 files, 283 tests passing
- [x] Pre-existing failures confirmed **not mine** by re-running with the branch stashed: `packages/db`'s `user-boards-serial-uniqueness.integration.test.ts` (needs a live DB) and the backend project fail identically on a pristine tree

### DIFF
--- a/docs/aurora-sync.md
+++ b/docs/aurora-sync.md
@@ -353,6 +353,57 @@ Failed to decrypt credentials
 - Verify user has `syncStatus: active` or `error` (not `disabled`)
 - Check Aurora API is accessible from the environment
 
+### Skipped logbook rows (`logbook_sync_skips`)
+
+Some upstream ascents/bids can't be written at all — a non-numeric `angle`, a
+comment carrying a NUL byte, or anything else Postgres refuses. Rather than let
+one of those abort the whole cross-table sync transaction and freeze the user's
+logbook forever (#3871, the DB-execution-time sibling of #3520), the sync skips
+the offending row and records it in `logbook_sync_skips` with its payload
+intact, so the send is quarantined and replayable rather than silently dropped.
+
+The table is **state, not a log**: a row is deleted the moment that `aurora_id`
+syncs cleanly, so anything still in it is currently broken.
+
+Who is affected, and why:
+
+```sql
+SELECT reason,
+       count(*)                AS rows_skipped,
+       count(DISTINCT user_id) AS users_affected,
+       max(last_seen_at)       AS last_hit
+  FROM logbook_sync_skips
+ GROUP BY reason
+ ORDER BY rows_skipped DESC;
+```
+
+Drill into one user (the payload is what you replay from):
+
+```sql
+SELECT aurora_type, aurora_id, reason, detail, seen_count, first_seen_at, payload
+  FROM logbook_sync_skips
+ WHERE user_id = '<user-id>' AND board_type = 'kilter'
+ ORDER BY last_seen_at DESC;
+```
+
+Reason codes:
+
+| `reason`            | What happened                                                              | Recovery                                                                       |
+| ------------------- | -------------------------------------------------------------------------- | ------------------------------------------------------------------------------ |
+| `invalid_angle`     | `angle` isn't a storable integer. No honest default, so the row is dropped | Fix upstream (the row re-syncs on its next upstream edit), or replay `payload` |
+| `invalid_identity`  | `climb_uuid` / `aurora_id` missing or carrying an unstorable byte          | Usually an upstream data bug; replay from `payload` once corrected             |
+| `normalize_failed`  | The row threw during parsing (bad `climbed_at`/`created_at`) — see #3520   | Same as above                                                                  |
+| `db_write_rejected` | Parsed and validated, but Postgres still refused it                        | Read `detail` for the Postgres error; usually needs a code fix                 |
+
+A skipped row is **not** re-fetched automatically: Aurora's user sync is
+incremental and the checkpoint advances past it (not advancing is the wedge this
+fixes). It comes back when the user edits it upstream, when the watermark is
+reset (`clearAuroraBoardData` drops `board_user_syncs`, so the next sync starts
+from 1970), or via the JSON export import below, which ignores the watermark.
+
+A rising `seen_count` on the same row means it keeps being re-delivered and
+re-refused — worth a look at `detail`.
+
 ## JSON Export Import (Alternative to API Sync)
 
 Since the Kilter backend has been shut down, API-based sync is no longer available for Kilter users. As an alternative, users can import their data from Aurora's JSON export file.
--- a/packages/aurora-sync/src/sync/apply-user-logbook.test.ts
+++ b/packages/aurora-sync/src/sync/apply-user-logbook.test.ts
@@ -871,3 +871,54 @@ describe('logbook_sync_skips reconciliation (#3871)', () => {
     }
   });
 });
+
+describe('applyLogbookChunk — claim UPDATE isolation (#3871)', () => {
+  it('isolates a refused row in the cross-source claim UPDATE, the third write path', async () => {
+    // The claim path binds only {uuid, aurora_id}, so it looks harmless — but
+    // it is a batched statement like the other two and fails the same way.
+    const warnSpy = vi.spyOn(console, 'warn').mockImplementation(() => {});
+    try {
+      const { tx, calls, skipRows } = createTx({
+        selectResults: [
+          [], // byAuroraId: both incoming rows are misses
+          [
+            // Claimable json_import originals at the same instant.
+            {
+              uuid: 'tick-ok',
+              climbUuid: 'climb-1',
+              angle: 40,
+              climbedAt: '2026-05-01T22:00:00.000Z',
+              status: 'send',
+            },
+            {
+              uuid: 'tick-poison',
+              climbUuid: 'climb-2',
+              angle: 40,
+              climbedAt: '2026-05-01T22:00:00.000Z',
+              status: 'send',
+            },
+          ],
+        ],
+        rejectTickWritesContaining: 'aur-poison',
+      });
+
+      await applyAuroraAscents(tx as unknown as Db, 'kilter', 'user-1', [
+        ascent({ uuid: 'aur-ok', climb_uuid: 'climb-1' }),
+        ascent({ uuid: 'aur-poison', climb_uuid: 'climb-2' }),
+      ]);
+
+      // Batched claim refused, then replayed per row: the clean claim lands.
+      expect(calls.filter((c) => c.kind === 'rollback')).toHaveLength(2);
+      expect(skipRows).toHaveLength(1);
+      expect(skipRows[0]).toMatchObject({ auroraId: 'aur-poison', reason: 'db_write_rejected' });
+      // The quarantined payload carries the whole row, not just the link —
+      // {uuid, aurora_id} alone could never be replayed into a tick.
+      expect(skipRows[0].payload).toMatchObject({
+        claimedTickUuid: 'tick-poison',
+        row: { auroraId: 'aur-poison', climbUuid: 'climb-2', angle: 40 },
+      });
+    } finally {
+      warnSpy.mockRestore();
+    }
+  });
+});
--- a/packages/aurora-sync/src/sync/apply-user-logbook.test.ts
+++ b/packages/aurora-sync/src/sync/apply-user-logbook.test.ts
@@ -922,3 +922,63 @@ describe('applyLogbookChunk — claim UPDATE isolation (#3871)', () => {
     }
   });
 });
+
+describe('applyAuroraBids — DB-level write isolation (#3871)', () => {
+  it('isolates a refused bid the same way as an ascent, end to end through applyAuroraBids', async () => {
+    // applyAuroraBids shares applyLogbookChunk with ascents, but "shares the
+    // machinery" is an inference; bids reach it via a different entry point
+    // (no tombstone pass, different claim statuses), so drive it directly.
+    const warnSpy = vi.spyOn(console, 'warn').mockImplementation(() => {});
+    try {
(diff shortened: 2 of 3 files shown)


marcodejongh:
Both comment nits addressed in e5ff84761, but flagging one disagreement rather than silently "fixing" it.

**Accepted — the subxid estimate was misleading.** It described the happy path as if it were the whole story. Now states the replay cost explicitly: a chunk that falls back opens up to `WRITE_CHUNK_SIZE` further savepoints, so one fully-bad chunk takes a 1000-row page from ~10 subtransactions to ~110 and over the 64-subxid threshold. Rare and self-limiting (poison rows get quarantined, not replayed every cycle), but it's the number anyone tuning `WRITE_CHUNK_SIZE` or the page size actually needs.

**Rejected — "postgres-js does issue RELEASE SAVEPOINT on success" is not correct.** From `postgres@3.4.9/src/index.js`, `scope()`:

```js
name && await sql`savepoint ${ sql(name) }`
try {
  result = await new Promise(...)
} catch (e) {
  await (name ? sql`rollback to ${ sql(name) }` : sql`rollback`)
  throw ...
}

if (!name) {                       // ← false for a NAMED savepoint
  prepare ? await sql`prepare transaction ...` : await sql`commit`
}
```

The commit/release branch is guarded by `if (!name)`, which is false for every nested savepoint, so on success postgres.js issues **nothing** — no `RELEASE`, no `COMMIT`. And `RELEASE SAVEPOINT` appears nowhere in the package: the only `release` hits in `src/` are connection-pool reservation (`sql.release`, `connection.reserved.release`), not savepoint release.

So the original claim was accurate; the wording just read as a bare assertion. Rather than leave it re-litigable I've replaced it with the mechanism, so the next reader can check it without reading the driver.

The net effect on the cost bound is the same either way, and worth noting for the record: even if a `RELEASE` *were* issued, it would not reduce the subxid count. A subtransaction that has written rows keeps its assigned XID until the outer transaction commits — `RELEASE` ends the savepoint's scope, not its entry in the transaction's subxid list. The bound in the comment holds on either reading.

marcodejongh:
Checked this one rather than "fixing" it, because the two branches of the concern have opposite remedies. It's the first branch, and the source is already the editor-safe form:

```
$ grep -an "UNSTORABLE_TEXT = " packages/aurora-sync/src/sync/apply-user-logbook.ts | xxd
00000010: 4142 4c45 5f54 4558 5420 3d20 2f5c 7530  ABLE_TEXT = /\u0
00000020: 3030 307c 5b5c 7544 3830 302d 5c75 4442  000|[\uD800-\uDB
```

Those are the six ASCII characters `\`,`u`,`0`,`0`,`0`,`0` — an escape, not an embedded byte. The file contains **0** raw NUL bytes (it did briefly during development, which is probably what the diff rendering is a ghost of; it was replaced with the escape before the first commit).

Running the regex parsed out of the committed source, so this tests what ships rather than a retyped copy:

```
source line: const UNSTORABLE_TEXT = /\u0000|[\uD800-\uDBFF](?![\uDC00-\uDFFF])|(?<![\uD800-\uDBFF])[\uDC00-\uDFFF]/g;
matches NUL:                true
strips NUL:                 "sick line"
strips lone high surrogate: "ab"
strips lone low surrogate:  "ab"
keeps a valid pair:         "ok 😀"
leaves plain text alone:    "normal text"
```

`\u0000` and `\x00` are the same code point, and `no-control-regex` fires on both — hence the disable. I've kept `\u0000` for consistency with the surrogate ranges in the same pattern, which are all `\uXXXX`.

The behaviour is also already pinned by a test rather than resting on inspection: *"strips a NUL byte from a comment and still writes the send"* feeds `'sick \u0000line'` through `applyAuroraAscents` and asserts the stored comment is `'sick line'`. If the escape were ever mangled into a literal `^@`, the NUL would survive the strip and that assertion would fail — which is exactly the editor-fragility failure mode worth guarding.

No change made. Happy to switch to `\x00` if you'd still prefer it, but it's equivalent and would make the pattern's escapes inconsistent.

marcodejongh:
Passing on one observation from the review of #3963 (the collapsed-migration PR this is now based on), since the decision belongs here rather than there.

`logbook_sync_skips` is the only file of 34 in `packages/db/src/schema/app/` that uses `withTimezone: true` — `first_seen_at` and `last_seen_at` land as `timestamp with time zone` where every other table in the schema uses a bare `timestamp`. It is carried through the collapsed migration exactly as written here; nothing was changed.

Not asserting it is wrong. `timestamptz` is arguably the better default and the reason may well be deliberate. But it is a one-table departure from the surrounding convention, so if it is deliberate it is worth a line in `logbook-sync-skips.ts` saying so, and if it is not, changing it is far cheaper before the table exists in production than after.

No action needed for #3963 either way.

Issue #3913, opened 2026-07-25
MoonBoard yellow backdrop: green-hold contrast + ClimbListThumbnail letterbox bars


marcodejongh:
## What's happening

Two small, known gaps from PR #3899 (the MoonBoard dark-mode backdrop fix, `moonboardWallBackdrop` in `packages/mobile/src/theme/colors.ts`), documented only in code comments today — filing so they don't rot unnoticed.

### 1. Green start holds are hard to read against the new yellow backdrop

MoonBoard's start-hold marker color (`#44FF44`, `MOONBOARD_HOLD_STATES.start.displayColor` in `packages/shared/board-config/src/moonboard-config.ts`) has a WCAG contrast ratio of only **~1.09:1** against the new backdrop color (`#FAD201`, RAL 1023 "Traffic Yellow") — effectively unreadable as a stroke-only marker.

This isn't fixable by tuning the backdrop hex: green and yellow are adjacent, similarly-bright hues, so contrast stays under ~1.35:1 for any backdrop still light/saturated enough to read as "yellow." Reaching a real WCAG threshold (3:1) needs a backdrop dark enough to look olive/khaki, which defeats the point of the fix. Full derivation + the contrast numbers for all three primary hold colors (hand/finish are fine-to-acceptable) are in the `moonboardWallBackdrop` comment.

**Possible fix**: a fixed dark outline on hold markers, independent of the marker's own fill/stroke color — a renderer-level change in `packages/board-renderer/core/src/renderer.rs` (currently markers render as a stroke in the hold's own color with no separate outline, for every board, not just MoonBoard). Worth checking how it looks in practice on a real device before deciding whether it's worth the renderer change.

### 2. `ClimbListThumbnail` letterboxes MoonBoard art with a visible flat-yellow margin

`ClimbListThumbnail` (`packages/mobile/src/components/ClimbListThumbnail.tsx`) renders `LayeredClimbImage` directly inside a fixed 76×96 portrait cell with no aspect-fit box. MoonBoard's board art is 650×1000 (0.65 aspect) — narrower than the cell's 0.79 aspect — so the actual board-art images (which use `contentFit="contain"`) letterbox with roughly a 7px margin on each side, but the new backdrop layer isn't `contentFit`-aware and fills the full cell width, showing as a thin flat-yellow bar on each side instead of blending into the same letterbox gap.

[...]

PR #3961, opened 2026-07-26
fix(mobile): make MoonBoard readable in dark mode without a yellow wall

### LINKED ISSUE #1449: [UI] hard to see moonboard climbs
image

Should add grey background or something similar

### LINKED ISSUE #3885: Dark mode: MoonBoard 2016 holds nearly invisible against the dark board (yellow background?)
## What's happening

A new user trying the iOS app reports that in **dark mode**, the **MoonBoard 2016** board is very hard to read. The board background renders near-black, the app chrome is also dark, so the holds nearly disappear into it.

> "dark mode is really hard to see the MoonBoard 2016 board. If the board background was yellow, that'd go a long way."

The reporter attached a screenshot of "An easy problem" (V4+ / 6B+) on MoonBoard 2016 — the highlighted start/hand/foot holds are the only things that stand out; the rest of the wall is barely visible.

## Suggested fix

Give the MoonBoard board a lighter (yellow) background instead of leaning on the dark app theme, so holds have contrast. Yellow also matches the real MoonBoard's yellow-and-black wall, so it reads as the actual board.

This is mainly a mobile dark-mode problem (that's where it was reported), but worth checking the web renderer too.

## Pointers

- Mobile board renderer: `packages/mobile/src/components/board-renderer/`
- Board background images are wired up in `packages/mobile/src/lib/board-backgrounds-manifest.ts` (assets sourced from `packages/web/public/images/...`).

## Notes

- Workaround today: switch the app/device to light mode.
- Reported via a new install on iOS.

marcodejongh:
## Summary

MoonBoard climbs were unreadable in dark mode. Not low-contrast — **invisible**: against the `#140E1E` play field, the A–K / 1–18 coordinate labels are pure `#000000` (1.1:1, literally darker than the surface they sit on) and the black "set B" hold sheets measure 1.05:1. Coordinates are how climbers call holds out to each other, so this is the most-referenced thing on the screen.

MoonBoard is the only board this happens to, and the reason is structural: **its art is drawn for a white wall.** Every board's art is a stack of ~96–99% transparent layers over the app's own background. Kilter, Tension and the rest draw mid-tone holds (mean visible luminance 72–89) that survive a dark field. MoonBoard draws pure-black lettering and black plastic.

### The fix

Ship dark-mode variants of **only the layers that are actually broken**. A build-time transform bakes them into `*.dark.webp` siblings; `background-image-cache.ts` prefers a sibling when the scheme is dark and falls straight through when there isn't one. That fallback is the whole design — light mode and every other board are byte-identical, and there is no `boardName === 'moonboard'` special case anywhere.

Two transforms, because the two kinds of layer differ:

| layer | transform | result |
|---|---|---|
| Coordinate labels | flat tint `#A8A8A8` | 6.99:1 vs card, 7.95:1 vs play field |
| Black hold sheets | gamma 0.5 over 56–250 | body 2.56:1 vs card, 2.42:1 clear of pale set A |

The labels are a tint because the source carries exactly **one** colour value (0.39% opaque, no grid — lettering only, all softness in the alpha). The hold sheets get a gamma curve rather than a linear remap because their detail is crushed against black (body IQR 16/255) and a linear remap *compresses* it further; gamma expands the dark end, taking body IQR to **29** — more modelling than the source. That matters: it's the difference between moulded holds and flat stickers, which is exactly why `expo-image`'s `tintColor` (`PorterDuff.SRC_IN`) was prototyped and rejected.

**Black holds land at ~2.5:1, not 3:1, and that is deliberate.** Both hold families share one background, so `min(hold-vs-card, hold-vs-paleA)` is capped at **2.49:1** by arithmetic. Chasing WCAG 1.4.11's 3:1 would push the black sheets to within 2.0:1 of the pale ones and dissolve the black-vs-pale distinction — a real MoonBoard product fact and the only cue that survives colour-vision deficiency. Up from 1.17:1.

### Marker colours are not touched at all

**Zero marker pixels change.** `hold-states.ts`, `moonboard-config.ts`, `render-config.ts`, `board-render/types.ts` and `renderer-version.ts` are byte-identical to `main`, so `RENDERER_VERSION` stays at **4** and no board's cached overlays are invalidated.

An earlier revision of this branch added a dark-mode-only hand colour, because lifting the black sheets brings them nearer the blue ring. It was dropped: it kept the LED's exact hue (240°, R==G) but drifted **66.5 ΔE** from it — the largest gap in the codebase, against 14–40 for every other shipping `displayColor` — and read as a visibly paler blue than the wall lights up.

The contrast didn't justify it. Measuring the real render, **81% of a marker ring sits on bare wall and only 19% crosses a hold**, so the hold-contrast constraint it was tuned for is a fifth of the ring; against the field, today's `#4444FF` already gets 3.16:1. And blue's 0.0722 luminance coefficient means any blue bright enough to beat that *must* desaturate — brightness and LED fidelity are in direct opposition, so no version of it is both brighter and faithful.

What remains — a ring clipping a lifted black hold at 1.09:1, on that 19% — is properly fixed by a backdrop-independent marker casing, which is #3913's job and costs no colour fidelity at all.

### Why not #3899

#3899 painted an opaque `#FAD201` backdrop under the whole board. That fixes the art by breaking the markers — a yellow field collapses the green start ring from 13.57:1 to **1.09:1** — and its `boardName === 'moonboard'` gate would have painted yellow behind layouts that already read correctly (MoonBoard 2024 is fine today). It then grew a four-preset accessibility setting to manage the damage. Superseded; closing it.

Closes #3885
Closes #1449

## Screenshots

Real Android captures at 810×1800 against Metro. All three panels are the same climb — "An easy problem" V4+, MoonBoard 2016, the climb from #3885 — with identical marker geometry.

MoonBoard 2016 dark mode before and after

Scope, across every affected surface. 2024's blue sets and Masters 2017's red set C are untouched; Kilter is a control.

Coverage across MoonBoard layouts plus Kilter control

## Release Notes

- MoonBoard walls are readable in dark mode again — hold numbers, letters and the black holds all show up instead of disappearing into the background.

## Test plan

- `vp check`, `vp run typecheck:mobile|web|shared` — clean.
- `vp run test:mobile` — 4515 tests pass. `vp run test:web` pass. `vp run check:mobile-bundle` — OK.
- New coverage: dark-variant resolution (including light mode and the thumb variant resolving to the *thumb* dark sibling, and a missing sibling never counting as a missing layer); `buildBoardKey` differing by scheme; a MoonBoard↔`HOLD_STATE_MAP` drift test (two live copies of those hexes exist on `main`, and an early draft of this branch changed one without the other); and a generator golden that asserts body IQR **versus the source**, so a future compressing curve fails instead of silently flattening the art.
- **On-device (Android emulator, dark + light):** MoonBoard 2016 / 2024 / Masters 2017 / Masters 2019 / Mini 2020, plus Kilter as a control, plus a lit climb in the play view and the climb list. Verified by measurement, not eye: rendered row-label glyph luminance 0.0056 → 0.3916 (1.00:1 → 7.95:1), marker geometry AND colour identical across all three panels (`#4444FF` throughout, verified by sampling the ring pixels — the screenshots are from the shipped build, with the served Metro bundle grepped to confirm it matches the source).
- Screenshots reviewed by a design panel, which blocked the first pass on three counts (no lit climb; the tone curve was compressing modelling; contrast figures measured against the wrong surface with the wrong pixel mask). All three are fixed above — that review is also what caught the `useColorScheme()` bug below.

### One thing worth flagging for review

The colour scheme is read from a new `ColorSchemeContext`, **not** RN's `useColorScheme()`. That was a real bug caught on device: `useColorScheme()` follows the OS trait collection and on Android does not track `Appearance.setColorScheme`, so anyone running the app's own Dark theme on a light-mode phone got the light board art on a dark screen — i.e. none of this fix. It's its own small context rather than a field on `Theme` because the consumer runs in every virtualized list row.

### Not fixed here

- Live Activity board art (Android composites natively, iOS pulls a server-rendered thumbnail) — neither is reached by a JS prop. Don't file as a regression.
- The Next.js web climbing surface — WONTFIX per the expo-web migration decision.


marcodejongh:
Thanks — addressed the effect-deps note, though not quite as suggested, so flagging the reasoning.

You're right that `renderSignature` subsumes both for *change detection*: it's derived from `holdRenderSignature` and `colorScheme`, so it fires whenever either moves. But both are also **read inside the effect body**, and for reasons that matter:

- `holdRenderSignature` feeds the `unsupportedRenderSignatures` capability check. That one has to stay **scheme-independent** — using the `|dark`-suffixed value there would make `!== DEFAULT_HOLD_COLOR_SIGNATURE` true for every dark-mode render with no user overrides at all, and a single native failure would then mark plain dark renders as "needs the marker-aware binary" and blank the overlay.
- `colorScheme` is passed to `getBoardConfig` to pick `displayColorDark`.

So dropping them would leave the array wrong the day someone lifts the `exhaustive-deps` disable above it. Kept both and added a comment explaining why they aren't redundant, since the non-obviousness was the real complaint (36b08f0).

marcodejongh:
**Note on the red `publish` check — infrastructure, not this change.**

The OTA preview upload is being throttled by the storage bucket:

```
<Error><Code>SlowDown</Code><Message>Please reduce your request rate.</Message>
<BucketName>boardsesh-ota</BucketName></Error>
```

Diagnosis:

- **Not caused by this PR.** It fails inside `eoas publish`'s asset upload, ~1.3s in, on a different asset key each time. Nothing in this diff touches upload behaviour, and the asset count (380) is essentially unchanged — this PR adds 10 webps totalling ~125 KB.
- **Production OTA is unaffected** — `mobile-ota-production.yml` is 13 success / 0 failure over its last 20 runs, latest at 02:03, uploading the same asset set. So the bucket is not globally rate-limited.
- **It looks per-prefix.** All three failures cite the identical key prefix `pr-3961/49a581a48aae.../`. The first publish on this PR (different SHA, different prefix) succeeded; every rerun since has written to that same prefix and been throttled. S3-compatible stores commonly shard rate limits by key prefix, which fits.

Retried twice, including once after a 15-minute backoff — same result, so further reruns just hit the same prefix. Stopping there rather than burning CI.

**Two things that make this non-blocking:**

1. `publish` is the OTA-preview convenience (load the PR on a store build without a new binary), not a correctness gate. The 21 real checks pass.
2. Channel `pr-3961` is already functionally current — the last successful publish carries all the app code, and the only commit since changes `docs/images/*.png`, which isn't in the app bundle.

The next real push to this branch gets a new SHA and therefore a new prefix, which should clear it on its own. If it recurs on other PRs it's worth a look at `eoas`'s upload concurrency against the bucket's rate limits — but that's its own issue, not this one.

PR #3963, opened 2026-07-26
chore(db): collapse four contending 0187 migrations into one

marcodejongh:
## What

Four open sync PRs each generated a migration numbered `0187`, against a `main` that tops out at `0186`. Only one of them can ever merge — the other three break on whichever lands first, and every rebase re-rolls the same collision. This branch collapses all four into a single generated migration so they stop serialising against each other.

The four schema changes are purely additive and mutually independent: three new tables, one new enum, five new indexes, one new foreign key. Nothing about their relative ordering was ever load-bearing; the contention was an artefact of four branches running `drizzle-kit generate` in parallel.

| Source PR | DDL collapsed here |
| --- | --- |
| #3931 | `board_circuits_user_idx` on `board_circuits (board_type, user_id)` |
| #3951 | `board_climb_ingest_skips` + `board_climb_ingest_skips_open_idx` |
| #3957 | `sync_daemon_leases` |
| #3960 | `logbook_sync_skip_reason` enum, `logbook_sync_skips`, FK to `users`, 3 indexes |

Schema sources were taken verbatim from each branch (`packages/db/src/schema/boards/unified.ts`, `packages/db/src/schema/app/`), not retyped from the SQL, so every comment and column type survives intact. The backend test harness mirror in `packages/backend/src/__tests__/schema-sql.ts` that #3957 added comes along too.

The four feature PRs are rebased onto this branch with all `packages/db/src/schema/**` and `packages/db/drizzle/**` changes dropped, so each carries only its own query and logic code. GitHub retargets them to `main` once this merges.

## The generated migration

`packages/db/drizzle/0187_sad_freak.sql` — one file, one snapshot, one journal entry at index 187.

Additive only: 1 `CREATE TYPE`, 3 `CREATE TABLE`, 1 `ALTER TABLE ... ADD CONSTRAINT` (the `logbook_sync_skips` → `users` FK), 5 `CREATE INDEX`. No `DROP` of any kind. The `boardsesh_ticks_quality_range` CHECK-constraint regeneration footgun this repo has hit before did not reproduce — the generated SQL was read statement by statement before committing.

## Verification

- `vp check` and `vp run typecheck` clean. (`vp check` reports 2 errors in `packages/db/src/queries/__tests__/tick-offset-inference.test.ts`; identical counts on a pristine `origin/main` checkout, so pre-existing and untouched here.)
- Whole journal applied to an isolated throwaway copy of the dev-db image on a spare port — not the shared dev container: 188/188 migrations recorded, all eleven new objects present in `pg_class` / `pg_indexes` / `pg_type` / `pg_constraint`, and a second apply is a no-op.

## Release Notes

### DIFF
--- a/packages/backend/src/__tests__/schema-sql.ts
+++ b/packages/backend/src/__tests__/schema-sql.ts
@@ -1132,6 +1132,17 @@ export const schemaSQL = `
     "deleted_at" timestamp DEFAULT now() NOT NULL
   );
 
+  -- Mirrors packages/db schema/app/sync-daemon-leases.ts (migration 0187).
+  -- Best-effort single-active-instance lease for the sync daemons.
+  DROP TABLE IF EXISTS "sync_daemon_leases" CASCADE;
+  CREATE TABLE IF NOT EXISTS "sync_daemon_leases" (
+    "daemon_name" text PRIMARY KEY NOT NULL,
+    "holder_id" text NOT NULL,
+    "acquired_at" timestamp DEFAULT now() NOT NULL,
+    "heartbeat_at" timestamp DEFAULT now() NOT NULL,
+    "hostname" text
+  );
+
   -- Mirrors 0146: sync cursor indexes lead with board_type; deleted_at serves
   -- the daily prune's DELETE WHERE deleted_at < cutoff.
   CREATE INDEX IF NOT EXISTS "sync_deletions_deleted_at_idx" ON "sync_deletions" ("deleted_at");
--- a/packages/db/drizzle/0187_sad_freak.sql
+++ b/packages/db/drizzle/0187_sad_freak.sql
@@ -1,0 +1,46 @@
+CREATE TYPE "public"."logbook_sync_skip_reason" AS ENUM('invalid_angle', 'invalid_identity', 'normalize_failed', 'db_write_rejected');--> statement-breakpoint
+CREATE TABLE "board_climb_ingest_skips" (
+	"board_type" text NOT NULL,
+	"climb_uuid" text NOT NULL,
+	"layout_id" integer,
+	"source_layout_uuid" text,
+	"reason" text NOT NULL,
+	"detail" text,
+	"raw_holds" text NOT NULL,
+	"frames_count" integer,
+	"climb_name" text,
+	"setter_username" text,
+	"first_seen_at" timestamp DEFAULT now() NOT NULL,
+	"last_seen_at" timestamp DEFAULT now() NOT NULL,
+	"resolved_at" timestamp,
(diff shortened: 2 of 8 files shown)


marcodejongh:
Checked all three. None need a change here, and the reasoning is worth recording because two of them rest on a premise that inverts once you look at the repo.

**1. Trailing newline on `0187_sad_freak.sql` — this is `drizzle-kit`'s own output shape, not a defect.** Every generated migration in the repo ends on `;` with no trailing newline:

| file | last byte |
| --- | --- |
| `0180_fine_unicorn.sql` | `3b` (`;`) |
| `0182_magical_thor.sql` | `3b` |
| `0184_steep_old_lace.sql` | `3b` |
| `0185_premium_randall_flagg.sql` | `3b` |
| `0187_sad_freak.sql` | `3b` |

The one recent file that *does* end `0a` is `0186_backfill_sync_frozen_at.sql` — a hand-named backfill, i.e. the exception. Adding a byte here would mean hand-editing generated migration SQL, which `CLAUDE.md` rules out, and it would desync the file from what `drizzle-kit generate` produces on the next run.

**2. Timestamp timezone — real observation, but it points the other way.** `logbook_sync_skips` is the outlier, not the two without a zone: it is the **only** file of 34 in `packages/db/src/schema/app/` that uses `withTimezone: true`. `timestamp` without a zone is the established convention across the schema, so `sync_daemon_leases` and `board_climb_ingest_skips` match it and `logbook_sync_skips` departs from it.

On the `heartbeat_at` TTL specifically: both sides of that comparison are server-side (`heartbeat_at` written by the DB's `now()`, compared against the DB's `now()`), so it is internally consistent whatever the server zone — the failure mode would need the server zone to change *between* a write and a read.

Either way, this PR is a mechanical collapse: every column here is byte-identical to what the four source PRs wrote. Silently changing one PR's column type inside another PR's migration is exactly the kind of drift this consolidation exists to prevent. Flagging it on #3960 instead, where the author can decide.

**3. The `schema-sql.ts` asymmetry is deliberate and load-bearing.** The harness mirrors what the *backend* tests need, and only one of the three tables is reached from there:

- `sync_daemon_leases` → `packages/backend/src/__tests__/sync-daemon-lease.test.ts` (8 references) exercises it against real Postgres, so #3957 mirrored it. That mirror came along here.
- `board_climb_ingest_skips` → zero references anywhere under `packages/backend/` on #3951. Its tests live in `packages/kilter-sync`.
- `logbook_sync_skips` → zero references anywhere under `packages/backend/` on #3960. Its tests live in `packages/aurora-sync`.

Mirroring the other two would add two tables the backend harness creates and drops on every worker for nothing, and would need hand-written SQL kept in sync with a schema no backend test reads — a divergence risk rather than a guard against one.

PR #3964, opened 2026-07-26
ci(db): auto-renumber stranded migration PRs when main takes their number

marcodejongh:
## Summary

Migration numbers are first-come-first-served, and every open migration PR appends to the same `_journal.json` tail. Right now four open PRs (#3960, #3957, #3951, #3931) all hold `0187` and all report **MERGEABLE** — the instant one merges, the other three go `CONFLICTING`. Unblocking each one is a hand ritual (rebase → resolve the journal → rebuild `packages/db` → re-run `drizzle-kit generate` → paste the hand-tuned SQL back) that this repo has performed **26 times**, three of them on 2026-07-16 alone, documented nowhere.

This makes the collision self-healing:

1. **`db-migration` label** — `pr-labels.yml` labels any PR that adds a migration.
2. **`native-fingerprint` label** — one step added to the existing `mobile-ota-check.yml`, which already computes the verdict. Marker only; the check-run stays neutral. Makes the native train queryable: `gh pr list --label native-fingerprint`.
3. **Fan-out on main** — when a migration lands, `db-migration-renumber-dispatch.yml` dispatches a renumber for every stranded PR.
4. **`db-migration-renumber.yml`** — rebases the PR, renumbers, proves the migration still applies against Postgres, then force-pushes.

The engine ships as `vp run db:renumber`, so a human hitting a collision runs one command instead of the ritual. New `docs/db-migrations.md` writes the whole thing down for the first time.

### Verified end-to-end, not just unit-tested

Ran against a real collision — PR #3957's migration rebased onto a `main` carrying PR #3960's:

- `0187_sad_freak.sql` → `0188_sad_freak.sql`, body **byte-identical** (md5 match)
- snapshot regenerated and correctly chained: a fresh `drizzle-kit generate` afterwards reports `No schema changes, nothing to migrate`
- journal `when` strictly above main's newest
- commit shape is a pure rename + new snapshot + journal entry — exactly matching the hand-written `63873f038`
- re-running is a clean no-op

### Findings that shaped the design

Read drizzle-kit 0.31.10's bundle rather than trusting its interface:

- **`generate`'s exit code is worthless.** `prepareAndMigratePg` ends `catch (e) { console.error(e); }` with no rethrow, and the folder reader calls bare `process.exit(0)` on a malformed snapshot or snapshot-parent collision. Success is decided by diffing the directory listing.
- **The `meta/` reset is a correctness requirement, not tidiness** — post-rebase both snapshots name the same parent, which drizzle treats as a collision and answers with that silent `exit(0)`.
- **`when`, not the number, is the apply gate.** A renumbered migration whose `when` isn't newer than main's is skipped *forever*, silently, in production. Clamped and asserted.
- **Tag references are never rewritten inside a migration body.** Seven migrations write a `_bs_migration_guards` idempotency key that is *not* the filename (`0163_merge_moonboard_duplicates.sql` guards on `0162_merge_moonboard_duplicates`). Rewriting it would re-run an applied migration on every dev machine.

### Two fixes this depends on

- **Deletes `packages/db/drizzle/0177_illegal_omega_red.sql`** — verified dead: zero journal entries, zero references, byte-identical to the applied `0179_bent_kid_colt.sql`. Left by a hand-renumber that forgot the `git rm`. `meta/0177_snapshot.json` is left alone; it belongs to the journalled `0177_ordinary_anita_blake`.
- **`.gitattributes` union-merges the schema barrel.** `packages/db/src/schema/**/index.ts` is an append-only export list, so *every* pair of table-adding PRs conflicts on the last line. Found this the hard way — the bot blocked on the real test case until this was added. Failure mode is loud: a duplicated export fails the TypeScript build.

### Also adds a guard rail

`vp run check:db-migrations` (new CI job, gated on `packages/db`, no DB and no build) catches the silent failures: an orphaned `.sql`, a journal entry with no file, a `when` that would be skipped, a number main already took. It would have caught `0177` in review. It deliberately ignores the six duplicate numbers main has carried since 2024 — both sides are journalled and applied in production.

## Release Notes

## Test plan

- `vp check` — 0 errors
- `vp run typecheck` — clean
- 68 new unit tests across `scripts/lib/drizzle-migrations.test.ts`, `scripts/check-db-migrations.test.ts`, `scripts/db-renumber-migration.test.ts`, including a byte-for-byte `_journal.json` round-trip against the real file (pins the 2-space indent and absent trailing newline that `.prettierignore` leaves unguarded)
- `scripts/mobile-ci-env-parity.test.ts` still green after the `mobile-ota-check.yml` edit
- End-to-end renumber against real PR branches, as above
- All five touched/added workflow YAMLs parse

### Not yet exercised in CI

The workflows can only really run once they're on `main` (dispatch always targets the default-branch copy). After merge I'll land a migration and watch the fan-out, confirming the App-token push does re-trigger `ci.yml` — a `GITHUB_TOKEN` push would leave the renumbered commit with no checks.

### Known limits, stated plainly

- **Fork PRs are out of scope.** `pull_request` hands fork workflows a read-only token, so neither label lands, and the App (installed on this repo only) couldn't push to a fork anyway. Consistent with the existing OTA-comment gap; `pull_request_target` stays off the table.
- **Per your call, no draft/staleness/cap filter on the fan-out** — every labelled open PR is renumbered. It does skip PRs whose number doesn't actually collide, which is correctness rather than a scope cut. `renumber:skip` is the opt-out.
- A bot force-push dismisses approvals (`dismiss_stale_reviews` + `require_last_push_approval` are on). Acceptable because the bot only touches PRs that just became conflicting and couldn't have merged anyway.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,13 @@
+# Schema barrel files are append-only export lists, so two branches that each add
+# a table both append to the same last line and conflict on every rebase — even
+# though the resolution is always "keep both". `merge=union` is git's built-in
+# answer for exactly this file shape: it takes the lines from both sides.
+#
+# This is what makes automatic migration renumbering (vp run db:renumber, and
+# .github/workflows/db-migration-renumber.yml) useful in practice. Without it,
+# essentially every table-adding PR conflicts here and needs a human.
+#
+# Safe because the failure mode is loud: if both sides add the *same* export, the
+# union keeps both copies and TypeScript fails the build. It never silently drops
+# a line, which is the outcome that would actually hurt.
+packages/db/src/schema/**/index.ts merge=union
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -244,6 +244,30 @@ jobs:
       - name: Typecheck (cache-driven affected)
         run: vp run typecheck
 
+  db-migrations:
+    needs: [changes]
+    # Cheap (a few file reads, no DB, no build), so it gates on the whole
+    # packages/db path rather than trying to be clever about which files matter.
+    if: needs.changes.outputs.sharedDeps == 'true' || needs.changes.outputs.rootCi == 'true'
+    runs-on: ubuntu-latest
+    timeout-minutes: 5
+    steps:
+      - uses: actions/checkout@v6
+        with:
+          # Needs origin/main reachable to compare numbering and `when` against it.
+          fetch-depth: 0
+      - uses: oven-sh/setup-bun@v2
+      - uses: voidzero-dev/setup-vp@v1
+        with:
+          node-version: 'lts'
+          cache: true
+      - name: Install dependencies
+        run: bun install --frozen-lockfile
+      - name: Validate the migration folder
+        run: |
+          git fetch --no-tags --depth=1 origin main
+          vp run check:db-migrations -- --base origin/main
+
   codegen-drift:
     needs: [changes, setup]
     # codegen.ts pulls GraphQL documents from packages/web/app/**/*.{ts,tsx}
@@ -737,6 +761,7 @@ jobs:
       - setup
       - lint
       - typecheck
+      - db-migrations
       - codegen-drift
       - i18n
       - test-default
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -39,6 +39,7 @@ jobs:
       # PR-only guard signal: did this PR edit the OTA-owned changelog data file?
       # Defaults false off PRs (the OTA bot legitimately rewrites it on push).
       changelogData: ${{ github.event_name == 'pull_request' && steps.filter.outputs.changelogData || 'false' }}
+      dbMigrations: ${{ github.event_name != 'pull_request' && 'true' || steps.filter.outputs.dbMigrations }}
       # True if any downstream job's gate would fire. setup gates on this so
       # docs-only / firmware-only PRs skip the bun-install + build:db pass
       # nothing downstream would consume. Must mirror the union of every
@@ -106,6 +107,11 @@ jobs:
             changelogData:
               - 'packages/mobile/src/data/changelog.generated.json'
               - 'CHANGELOG.md'
+            dbMigrations:
+              - 'packages/db/drizzle/**'
+              - 'packages/db/src/schema/**'
+              - 'scripts/check-db-migrations.ts'
+              - 'scripts/lib/drizzle-migrations.ts'
 
   setup:
     needs: [changes]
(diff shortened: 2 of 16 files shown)


marcodejongh:
Thanks — three of these were real. Fixed in 9eccb2b95.

**1. `mergeBaseRef` staged conflict markers** — correct, and worse than it looked. The merge path called `git add` on a conflicted `.sql` without resolving it, so the merge commit would carry conflict markers into the author's history (the final tree recovers, because the file is renamed from the saved original afterwards, but the intermediate commit does not).

The underlying trap is that `--ours`/`--theirs` are **inverted between merge and rebase**: rebasing replays the branch onto main, so the branch's side is `--theirs`; merging main into the branch makes the branch's side `--ours`. Both want the same thing and need opposite flags. Now resolved with `--ours` before staging, plus a post-commit assertion that nothing is left unmerged.

Added three tests that read the source and assert the orientation per function and that no path is staged before it is resolved. Mutation-checked: removing the fix fails 2 of them.

**2. Silent skip when `OTA_PUSH_APP_ID` is unset** — correct, and the worst of the three, since the failure mode is indistinguishable from "nothing to do". Added a `Require the push credential` step in the `gate` job that fails the run with an explicit error. This is operator misconfiguration rather than a per-PR condition, so a red run is the right signal — a comment on each stranded PR would be noise repeated N times per merge.

**3. `Math.max(...[])` → `-Infinity`** — correct, and it fails *open*, which is the dangerous direction: every PR would look like a collision and get force-pushed. Now guarded with `Number.isFinite(mainNext) && mainNext > 0`, matching the fail-closed posture the rest of that script already has.

---

**`actions/checkout@v4` vs `@v6` — not taking this one.** The repo is genuinely mixed: 35 `@v4` against 42 `@v6`. Every workflow this one is modelled on — the whole `mobile-ota-*` family, which is the closest analogue (App token, dispatch-only, operates on an arbitrary PR) — uses `@v4`, including `mobile-ota-check.yml` which this PR also edits. Matching the family it sits in beats matching `ci.yml`. Happy to do a repo-wide bump as its own PR, but not as a drive-by in this one.

**On the merge-path test coverage note:** `mergeBaseRef` is I/O against a live git tree, so it stays out of unit tests by design (the pure decision logic it depends on, `conflictsAreResolvable`, is covered). The source-reading tests above are the pragmatic guard for the part that actually broke. The rebase path is covered end-to-end instead — I ran it against a real collision (#3957's migration onto a `main` carrying #3960's) and verified the body stayed byte-identical and that a fresh `drizzle-kit generate` afterwards reports `No schema changes`.

marcodejongh:
Three more fixed.

**Dispatch `cancel-in-progress`** — the important one, and I'd had it backwards. Cancelling a fan-out mid-loop silently drops every PR it hadn't reached, and those stay stranded until the *next* migration merge, which may be days. Now `false`, matching the per-PR renumber group. A queued second run is cheap precisely because it only acts on real collisions, so anything already renumbered is skipped.

**Unpaginated `listComments`** — real; past 100 comments the sticky marker is invisible and the bot posts a duplicate every run instead of updating. Now `github.paginate`, same as the dispatch script.

**`db-migrations` gate too broad** — fair. Added a dedicated `dbMigrations` paths-filter (`packages/db/drizzle/**`, `packages/db/src/schema/**`, and the two script files) instead of leaning on `sharedDeps`, which also covers `packages/shared`, `shared-schema`, `board-constants` and the sync packages. Gating on the schema as well as the migrations is deliberate — a schema change with no migration is exactly the case where the collision and stale-`when` checks are worth running.

**`actions/checkout` version** — still declining, same reasoning as above: 35 `@v4` vs 42 `@v6` in this repo, and every workflow in the `mobile-ota-*` family this one is modelled on uses `@v4`.

PR #3613, opened 2026-07-15
feat(mobile): Android tablet support — Material 3 adaptive shell

marcodejongh:
## Summary

Makes an Android tablet a first-class wall device. Today the app has a full **iPad** adaptive shell — a left rail replacing the bottom tabs, a master-detail play pane that follows your selection, and a live "Now on the wall" column — but every Android device, phone or 10" tablet, fell through to the phone UI stretched edge-to-edge. Oleg at Sandbox Bouldering asked whether a cheap Android tablet bolted to the gym wall would work; the honest answer was "no."

Now an Android tablet mounts the same adaptive shell, drawn in **platform-native Material 3**: a Material navigation rail instead of the glass sidebar, opaque M3 tonal surfaces instead of glass, M3 dividers between panes. A ~$150 10" Android tablet on the wall gets the two-pane browse-light-and-see-what's-lit experience, not a blown-up phone screen — including the live wall column in landscape.

**How it works** (mostly wiring — the shell is already width-driven, not OS-driven):
- **Detection** — `resolveIsTablet` admits an iPad **or** an Android tablet whose smallest physical-screen width clears 600dp (Android's own `sw600dp` tablet qualifier), launch-fixed. The single `isPad` gate is generalized to `isTablet`; the pure size-class resolvers were already device-agnostic.
- **Material 3 rail** — new `MaterialNavigationRail` mirrors `MaterialTabBar`'s M3 roles laid out vertically (secondaryContainer active-indicator pill, onSecondaryContainer glyph, active/inactive glyph pairs matching the bottom bar), routed by variant via `TabletSidebar = createVariantComponent(IpadSidebar, MaterialNavigationRail)`. Shares the 96dp rail width so the master-detail width budget never forks per variant. No rail FAB — IA parity with the glass rail.
- **Rotation** — new `with-android-tablet-orientation` config plugin lets `sw600dp` tablets rotate to landscape (the shell's whole point) while phones stay portrait, mirroring the iOS `~ipad` override.
- **Live wall** — Android tablets are `panel-capable`, so the "Now on the wall" column/strip mounts beside the browse list (the live width budget picks column vs strip). The kiosk hero-type scale skips the iPad points-per-inch estimate on Android and falls back to the pane short side.

The breakpoints (700/1024dp) are unchanged and shared across both variants, so iPad and Android tablets resolve identically at the same dp. Phones and iOS are untouched.

## Release Notes

[...]

PR #3971, opened 2026-07-27
chore(mobile): upgrade Expo SDK 57 dependencies to latest patches [native-train]

marcodejongh:
## Summary

`expo install --check` reported **35 packages behind**. This brings every Expo package onto the SDK 57 recommended version and re-creates all three Bun patches against their new versions. `expo install --check` now reports **"Dependencies are up to date"**.

> [!IMPORTANT]
> **Draft `[native-train]` — do not merge into a normal train.** This moves native deps, so it shifts main's fingerprint away from the shipping store binaries. Merging it stops production OTA delivery until a new store release ships. Queue it for a deliberate native release train.

### Package bumps

34 Expo packages, all within SDK 57, exact-pinned per `check:mobile-expo-deps-pinned`:

| package | from | to |
| --- | --- | --- |
| `expo` | 57.0.2 | 57.0.8 |
| `expo-router` | 57.0.3 | 57.0.8 |
| `expo-updates` | 57.0.6 | 57.0.10 |
| `expo-modules-core` | 57.0.2 | 57.0.7 |
| `@expo/ui` | 57.0.3 | 57.0.7 |
| `expo-notifications` | 57.0.3 | 57.0.7 |
| `expo-constants` / `expo-asset` | 57.0.3 | 57.0.7 |
| `expo-dev-client` | 57.0.5 | 57.0.9 |
| `expo-image-manipulator` / `expo-image-picker` / `expo-location` | 57.0.2 | 57.0.6 |
| `expo-splash-screen` | 57.0.2 | 57.0.5 |
| `expo-linking` | 57.0.1 | 57.0.4 |
| …and 20 more at 57.0.0 → 57.0.1/57.0.2 | | |
| `react-native-screens` | 4.25.2 | 4.26.0 |

### The patches are the real work here

Bun keys `patchedDependencies` by **exact version**, so a version bump silently drops the patch and only warns. Confirmed on this branch before fixing: `bun install` installed `@expo/ui` 57.0.7 and `expo-dev-launcher` 57.0.9 **unpatched**.

- **`@expo/ui` → 57.0.7** — re-applied `onFullyDismissed` (consumed by `LogAscentSheet`, `ClimbFilterSheet`, `AddToPlaylistSheet`, `AddBetaVideoSheet`) and the Android `expand()`/`partialExpand()` rejection guard (#3478). Checked upstream 57.0.7 first: it ships neither, so both hunks are still needed.
- **`expo-dev-launcher` → 57.0.9** — re-applied the 10s → 120s request timeout.
- **`react-native-screens` → 4.26.0** — the iOS 26 `contentScrollViewForEdge` swizzle ports verbatim (both `RNSScrollViewFinder` files are byte-identical upstream). The bottom-accessory relayout needed **re-anchoring**: 4.26.0 extracted the accessory block out of `updateContainer` into `applyBottomAccessoryVisibility`. Anchoring there means it now also covers the `bottomAccessoryHidden` prop path, which the 4.25.2 patch missed.

Also updated: root `overrides` + `patchedDependencies` pins, and the two `patchedKey` rules in `scripts/mobile-patches-check.ts`.

### Unrelated pre-existing fix — split out and merged

`typecheck:mobile` was failing on `main` (the android-tablet-shell merge #3613 dropped `isPad` from `useDeviceLayout`'s return while two consumers still destructured it). That was split into #3972, which has since **merged to `main`** along with a regression test for the hook's returned shape. This branch is rebased on top and now contains **only** the dependency bump and the three patches.

## Release Notes

## Test plan

All green on Linux:

- `expo install --check` → "Dependencies are up to date"
- `vp run typecheck:mobile`
- `vp run test:mobile` → 554 files, 4634 tests
- `vp run check:mobile-bundle` → ios **and** android bundles OK
- `vp run check:mobile-patches` → all 3 patches applied (sentinels present)
- `vp run check:mobile-deps` → 46 deps match SDK 57.0.8
- `vp run check:mobile-expo-deps-pinned` → 36 Expo packages exact-pinned
- `vp run check:mobile-native-deps`, `vp check` (0 errors)
- `vp test run --project scripts` → 418 tests

Lockfile churn is contained: outside the Expo packages only `commander`, `deepmerge`, `@react-native/normalize-colors`, and `zod` moved, all transitively.

### Still needs a macOS pass before this ships

The `react-native-screens` patch is iOS-only native code that no JS-level check can exercise — `check:mobile-patches` only asserts the sentinel symbols are present, not that they behave. Before cutting the train, verify on an iOS 26 simulator:

1. **Tab-bar minimize** tracks the climbs `FlashList` on scroll.
2. **Bottom accessory** renders on the *first* climb set while already on a tab (the post-appearance attach), not just after a tab change.
3. Sheet dismiss callbacks still fire (`onFullyDismissed`) — log an ascent, add to playlist, add a beta video.

### Note, not addressed here

`expo install --fix` suggests adding `expo-asset`, `expo-font`, `expo-image`, and `expo-sqlite` to the `plugins` array in `app.config.ts`. That's pre-existing (those packages were installed before this PR), and the array has hand-tuned ordering constraints, so I left it alone. Worth a separate look.

https://claude.ai/code/session_01214fRRANfX3jL91sspHLq6


marcodejongh:
Split the unrelated `isPad` fix out to #3972 per review point 2, plus the `useDeviceLayout` shape test from point 3. That PR is JS-only with no fingerprint change, so it can land on `main` now instead of waiting for the native train.

Once #3972 merges I'll rebase this branch on `main`; the duplicate commit drops out cleanly and this PR narrows to just the dependency bump and the three patches.

On review point 1 — agreed, and that's why this is a draft. The macOS simulator pass is a hard gate before the train is cut; the three flows are listed in the test plan above.

Issue #3973, opened 2026-07-28
iOS 26 native glass: floatingControlBottom / scrollBottomPadding double-count the tab bar + accessory


marcodejongh:
On the iOS 26 native-glass path, `computeBottomChromeMetrics` (`packages/mobile/src/hooks/bottom-chrome-metrics.ts`) adds the tab bar and the climb accessory to `floatingControlBottom` / `scrollBottomPadding` a second time, because the UIKit glass tab bar has already folded both into `insets.bottom`.

### The arithmetic

On an iPhone 17 Pro / iOS 26, on a top-level tab with a current climb and the native `BottomAccessory` mounted, the measured bottom safe-area inset is **139** = home indicator (34) + tab bar (49) + accessory (56). That number is on-device verified and is already documented in the `preSessionFooterBottom` docblock in the same file.

Both fields then add the same chrome again:

```ts
scrollBottomPadding:   insetsBottom + tabBarHeight + jsQueueReserve,
floatingControlBottom: insetsBottom + tabBarHeight + Math.max(jsQueueReserve, nativeAccessoryReserve),
```

- `tabBarHeight` = `TAB_BAR_HEIGHT` = 49 (already inside the 139)
- `nativeAccessoryReserve` = `glassSize.standard + TOOLBAR_GAP_ABOVE_TABBAR` = 48 + 10 = 58 (the accessory's 56 is already inside the 139)

So `floatingControlBottom` = 139 + 49 + 58 = **246**, i.e. ~107pt higher than the ~139 that would actually sit on top of the chrome. `scrollBottomPadding` = 188 on the same screen, ~49pt of dead space at the bottom of every list.

`preSessionFooterBottom` and `inSessionListBottom` are already correct here — they use the raw inset for exactly this reason. `fixedFooterBottom` has the same double-count on this path (246), but its `contentInsetBottom` term is the one the in-flow-vs-overlay distinction is built on, so it needs the same look.

### Who this affects

Anything reading `floatingControlBottom`: `QueueAddedSnackbar`, `UndoWallChangeSnackbar`, `AccessoryOnboardingTip`, the session-detail FAB. On iOS 26 iPhones with the native accessory they float roughly a tab-bar-plus-accessory higher than intended. Lists using `scrollBottomPadding` get ~49pt of extra slack.

### Severity

[...]

Issue #3776, opened 2026-07-19
[UI] Climb: Filter "Show" button is weirdly high


ES-Alexander:
There are lots of filtering options, so it seems wasteful to cover almost half of the sheet with just the submit button:
<img height="500" alt="Image" src="https://github.com/user-attachments/assets/a4a74bfb-af7e-4528-88ff-2bb4dc1e1cf9" />

On an iPhone SE - I suspect this isn't the case on all devices / screen sizes 🤷‍♂️

marcodejongh:
**Real diagnosis: this isn't a formula bug, and no arithmetic fix to `useSheetColumnStyle` can close this gap.**

Measured directly on the screenshot attached to this issue (iPhone SE 3, 375×667 @2x, iOS 26):

- Footer hairline sits at **445pt**
- Apply button spans **457–504.5pt**
- Column height from the current formula (`insets.bottom(0) + 12` padding baked in): **≈390pt**
- The formula's own predicted value: **541pt**
- True available detent height on this device: **~548pt**

That's **~142pt of dead card below the footer**. The button isn't floating because the JS estimate is a bit off — it's floating because something is capping the actual rendered content box far short of the detent, in both directions the formula has been tuned toward. PR #3611 attempted a formula correction; even taken at face value, its effect on this device is 43pt — nowhere near enough to explain, let alone close, a 142pt gap. That PR is closed for a separate, independent reason (its premise about how `.fraction()` measures turned out to be wrong — see the closing comment), but even if its premise had been right, it wouldn't have fixed this.

**Corroborating evidence this is stable across formula versions**: ES-Alexander's screenshot on #3330 (2026-07-02), taken when the formula had *no* 24pt correction term at all (computed column: 562pt), shows the *same proportional dead gap* — roughly 24% of the card back then vs. ~26% now. A ±22pt swing in the computed column height produced no visible change in the gap. If the JS estimate were the cause, that swing should have moved the footer. It didn't.

**Leading hypothesis — not confirmed, needs a device I don't have from here**: `RNHostView`'s `setViewSize` (`RNHostView.swift:86`) is reporting a content size to Yoga that's smaller than the real detent, and no JS-side estimate can compensate for a native under-report in either direction.

Filed a follow-up: #3922, proposing a runtime `onLayout` probe instead of another guessed formula, with the exact device measurements needed to confirm or rule out this hypothesis.

Related: #3330 (original footer-floating report, same root symptom, same device class).
[...]

PR #3974, opened 2026-07-28
fix(mobile): clear the queue tray for the pre-session Start capsule

### LINKED ISSUE #3967: 🐞 Bug report: Start session now stuck behind tray, hard to click.

## Comment

Start session now stuck behind tray, hard to click.

## Metadata

| Field | Value |
| --- | --- |
| Platform | ios |
| App version | 2.2.2 (3) |
| Source | drawer-bug |
| Board | kilter / layout 1 / size 10 / sets [1,20] @ 40° |
| Climb | The Little Squeeze (7a/V6) |
| Session | 6d5d7127-6760-4666-9b73-ad6c76f03505 |
| URL | /discover |
| User agent | unknown |
| Feedback row | app_feedback #10195 |

## Contact

✅ Reporter agreed to follow-up — resolve **app_feedback #10195** in the DB to reach them.

marcodejongh:
Fixes #3967.

## What was wrong

`computeBottomChromeMetrics` (`packages/mobile/src/hooks/bottom-chrome-metrics.ts`) had two sibling outputs that had drifted apart:

```ts
inSessionListBottom:    selectByVariant(uiVariant, { material: fixedFooterBottom, liquidGlass: insetsBottom + jsQueueReserve }),
preSessionFooterBottom: selectByVariant(uiVariant, { material: fixedFooterBottom, liquidGlass: insetsBottom }),
```

The Liquid Glass branch of `preSessionFooterBottom` omitted `jsQueueReserve`. The docblock justifying the raw inset carries real on-device evidence (iPhone 17 Pro / iOS 26: bottom safe-area inset 139 = home indicator 34 + tab bar 49 + accessory 56, so `fixedFooterBottom` would double-count to 246). That evidence is correct — but it only describes the **native tab bar** path, which is exactly the path where `jsQueueReserve` is 0 because the UIKit `BottomAccessory` owns the climb. The conclusion got applied to the whole `liquidGlass` variant.

`resolveUiVariant` returns `liquidGlass` on every iPhone, including iOS < 26 (blur fallback). `useNativeTabBar()` additionally requires `glassCapable && !isTablet`. So on iOS < 26, on non-glass-capable iPhones, on iPad in a narrow split, and on Android with the variant forced to Liquid Glass, you get the glass variant **plus** the JS `MaterialTabBar` **plus** the JS `PersistentQueueBar` tray — and that tray is a JS overlay that does not extend the UIKit safe area. Nothing reserved for it, so the Start capsule sat underneath it.

Measured in screen-local coordinates (`[bottom, top]` from the screen's bottom edge):

| configuration | tray band | Start capsule | overlap |
| --- | --- | --- | --- |
| iOS < 26 / non-glass iPhone | `[10, 66]` | `[42, 94]` | 24pt of the capsule inside the tray |
| iPad narrow Split View | `[30, 86]` | `[28, 80]` | 50 of the capsule's 52pt covered |
| iOS 26 iPhone, native accessory | no JS tray | `[139, …]` | none |

The tray's trailing `heroSlot` is the log-ascent tick FAB (56 wide, right-aligned at a 16pt inset) and the Start capsule is `alignItems: 'flex-end'` — so the tick sat right over the capsule's tappable right half. That is the "stuck behind the tray, hard to click" in the report.

## What the issue body got wrong (or left ambiguous)

The report is accurate as far as it goes, but two things will trip up anyone reproducing it:

1. **"iOS 2.2.2" is the app version, not the OS version.** This bug does **not** reproduce on an iPhone running iOS 26 — that path mounts the native accessory inside the UIKit glass tab bar, `jsQueueReserve` is 0, and the footer offset is already correct (139, on-device verified). Try it on a current iPhone and you will conclude the report is bogus. The reporter is on iOS < 26, a non-glass-capable iPhone, or an iPad in a narrow split.
2. **"tray" is the floating `PersistentQueueBar`** (centred climb capsule + standalone log-ascent tick), not the tab bar and not the play drawer.

## The fix

One line: make the glass branch symmetric with `inSessionListBottom`.

```ts
preSessionFooterBottom: selectByVariant(uiVariant, {
  material: fixedFooterBottom,
  liquidGlass: insetsBottom + jsQueueReserve,
}),
```

Plus a rewritten docblock that keeps the iPhone 17 Pro evidence but scopes it to the native-tab-bar path, and says the two fields must stay in lockstep. The same stale "raw safe-area inset on Liquid Glass" wording lived on in two call sites the one-line fix didn't touch — `SessionStartFab`'s docblock and `PreSessionView`'s comment above `footerBottom` — so both are corrected too. That over-generalisation is what caused this bug once; leaving copies of it in the files a future consumer actually reads would just relocate it.

Resulting values:

- iOS < 26 / non-glass iPhone: 34 → 100. Capsule clears the tray by 42pt.
- iPad narrow split: 20 → 86. Clears the tray exactly.
- **iOS 26 native accessory: unchanged at 139** — `jsQueueReserve` is 0, so this is byte-identical to the on-device-verified value.
- No current climb, Material variant, iPad with the detail pane mounted: all unchanged.

The change is **monotone** — the footer can only move up, and only when the JS tray is actually mounted — so even if the coordinate model were off, it cannot introduce a new occlusion. Worst case is extra air under the capsule.

## Tests

`packages/mobile/src/hooks/__tests__/bottom-chrome-metrics.test.ts`:

- Two existing assertions encoded the bug as correct (`preSessionFooterBottom` 34 on the iOS < 26 case, 20 on the iPad narrow split). Both updated to `+ TOOLBAR_RESERVE`.
- **New — tray-clearance geometry.** Rebuilds the tray's band from `ActiveContextBar`'s own formula (`tabBarBottom + TOOLBAR_GAP_ABOVE_TABBAR`, height `glassSize.hero`), converts it to screen-local coordinates, and asserts `preSessionFooterBottom >= trayTop`. Not a tautology: the expected side comes from the layout constants + `tabBarBottom`, the actual side from `preSessionFooterBottom` — two independent code paths.
  The tray's formula (`floatingContextBarBottom`) was pulled out of `ActiveContextBar` into `theme/layout` so the
  component and the guard share one definition — a comment saying "keep these in sync" is not a safety net, and
  the test can't import the component without dragging react-native into a pure suite. Checked: nudging the shared
  formula up by 200 reddens both geometry tests.
- **New — matrix invariant.** Over variant × native/JS tab bar × accessory surface × climb × accessory mount × sidebar × detail pane: whenever `jsQueueToolbarVisible`, `preSessionFooterBottom >= jsQueueReserve`.
- **New — symmetry pin.** Over the same matrix, `preSessionFooterBottom === inSessionListBottom`, so one branch can't be edited without the other again.

**Mutation check (run, not claimed):** deleting `+ jsQueueReserve` from the new line turns **6 tests red** — both updated literals, both geometry tests, the matrix invariant, and the symmetry pin. Restoring it returns the file to 29/29 green. The existing iOS 26 test (`preSessionFooterBottom === 139`, `toBeLessThan(fixedFooterBottom)`) stayed green throughout — that is the guard that the one configuration with device evidence is undisturbed.

## Validation

- `vp run test:mobile` — 557 files, 4678 tests, all passing
- `vp check` — 0 errors (271 pre-existing warnings, none in the touched files)

## Deliberately not in this PR

- **`floatingControlBottom` / `scrollBottomPadding` double-count the tab bar + accessory on the iOS 26 native-glass path** (246 instead of ~139 for snackbars and FABs). Real, but it errs safe — everything floats too high, never occluded — it is a different symptom from this report, and unlike this fix it moves controls **down**, so it needs an iOS 26 device. Filed as #3973 with the arithmetic worked out and cross-linked to #3776 (whose "~142pt dead gap" may be the same root cause).
- **The tighter `usesNativeTabBar`-conditional formula.** `insetsBottom` is vestigial on the in-flow-JS-tab-bar path, so the tightest correct value would clear the tray by ~8pt instead of 42. It threads more inputs into the branch and can move the capsule back **down** — the one direction that could re-create an occlusion, on hardware nobody here can test. Deferred pending device feedback.
- **Collapsing `preSessionFooterBottom` and `inSessionListBottom`** into one field now that they are numerically identical. The two names document genuinely different consumers (`InSessionView` list padding vs `PreSessionView` absolute footer offset); the symmetry test gives the same drift protection without churning 2 components and 4 test files.

## Device QA gates (maintainer)

Nothing here was rendered — Linux box, no macOS, no simulator, no iPad. The arithmetic is unit-proven; the pixels are not. On the Record tab's "Start a session" screen **with a current climb set** (so the tray is up):

1. **iPhone on iOS < 26 or any non-glass-capable iPhone — the primary gate.** The "Start" capsule must sit fully above the floating climb capsule + tick, and a tap on its right half (previously under the tick FAB) must start the session rather than open the log-ascent sheet. No such device? Force the same code path by setting the UI variant to Liquid Glass in settings on an Android phone.
2. **iPhone on iOS 26 — the no-change gate.** The Start capsule must be in exactly the same place as before. Any movement means `jsQueueReserve` leaked into a case it shouldn't, which would undo the on-device work recorded in the docblock.
3. **iPad in narrow Split View / Slide Over.** Worst case before the fix (50 of 52pt covered). Confirm it clears.
4. **Android on the default Material variant — control.** The Start FAB must not move at all.
5. **Judgement call:** on (1) the capsule now floats ~42pt above the tray rather than the ~8pt a tighter formula would give. If that reads as too much air, say so and the conditional variant is a small follow-up — but only with device confirmation, since it moves the capsule back down.

Ships OTA: pure JS, no native inputs touched, no fingerprint impact.

## Release Notes

Starting a session on older iPhones and on iPad in Split View: the "Start" button no longer hides under the floating climb bar, so you can actually tap it instead of hitting the log-ascent tick by mistake.

### DIFF
--- a/packages/mobile/src/components/queue-control/ActiveContextBar.tsx
+++ b/packages/mobile/src/components/queue-control/ActiveContextBar.tsx
@@ -22,6 +22,7 @@ import {
   TOOLBAR_FAB_SIZE,
   TOOLBAR_GAP_ABOVE_TABBAR,
   TABBAR_SEAM_OVERLAP,
+  floatingContextBarBottom,
   glassSize,
 } from '../../theme/layout';
 import { useReduceMotion } from '../../hooks/use-reduce-motion';
@@ -66,10 +67,11 @@ export function ActiveContextBar({
   // Docked (Material): sit on the tab bar's REAL measured top, tucked under its hairline
   // by one px. Before the first measurement, fall back to the constant-estimated top
   // (≤2px off for a single frame, then it snaps to the truth). Floating (glass) keeps
-  // its lift above the tab bar.
+  // its lift above the tab bar — shared with the #3967 tray-clearance tests via
+  // `floatingContextBarBottom`, so the tray can't move without moving that guard.
   const bottom = dockToTabBar
     ? (measuredTabBarHeight ?? bottomChrome.tabBarBottom) - TABBAR_SEAM_OVERLAP
-    : bottomChrome.tabBarBottom + gapAboveTabBar;
+    : floatingContextBarBottom(bottomChrome.tabBarBottom, gapAboveTabBar);
 
   // Outer shell: plain View so `pointerEvents="box-none"` is respected on Android.
   // Reanimated's `entering` animations ignore pointerEvents on the animated view
--- a/packages/mobile/src/components/queue-control/__tests__/persistent-queue-bar.test.tsx
+++ b/packages/mobile/src/components/queue-control/__tests__/persistent-queue-bar.test.tsx
@@ -82,6 +82,8 @@ vi.mock('../../../theme/layout', () => ({
   TOOLBAR_GAP_ABOVE_TABBAR: 10,
   TABBAR_SEAM_OVERLAP: 1,
   SIDEBAR_WIDTH: 96,
+  // Mirrors the real implementation — ActiveContextBar's floating branch calls it.
+  floatingContextBarBottom: (tabBarBottom: number, gapAboveTabBar = 10) => tabBarBottom + gapAboveTabBar,
   glassSize: { hero: 64, inline: 44 },
 }));
 // The docked Material bar reads the tab bar's measured height to position itself.
--- a/packages/mobile/src/components/session-screen/SessionStartFab.tsx
+++ b/packages/mobile/src/components/session-screen/SessionStartFab.tsx
@@ -45,9 +45,18 @@ type SessionStartFabProps = {
  * not a full-width pinned bar. Liquid Glass renders a brand-tinted **glass**
  * capsule (the iOS 26 `.glassProminent` look: a `GlassSurface` tinted with the
  * brand hue, not a flat opaque fill); Material renders an M3 extended FAB. Both sit
- * over the host-supplied `bottomOffset` (variant-correct: the raw safe-area inset on
- * Liquid Glass — which already includes the tab bar + accessory — or the fixed-footer
- * reserve on Material) and report their height through `onHeightChange`.
+ * over the host-supplied `bottomOffset` and report their height through
+ * `onHeightChange`.
+ *
+ * `bottomOffset` is variant-correct, but on Liquid Glass it is NOT just the safe-area
+ * inset: it is the inset plus the JS queue-tray reserve. That reserve is 0 only on the
+ * native-tab-bar path, where the UIKit glass bar already extends the inset over the tab
+ * bar + accessory. On the JS-tab-bar fallback the floating `PersistentQueueBar` is an
+ * overlay outside the safe area, so the reserve has to be added explicitly or this
+ * capsule lands under the tray's log-ascent tick (#3967). Material uses the
+ * fixed-footer reserve. Don't re-derive any of that here — take the number the host
+ * passes; the whole arbitration lives in `computeBottomChromeMetrics`
+ * (`preSessionFooterBottom`).
  *
  * The End action no longer lives here — it docks in the top chrome (see
  * `RecordTopChrome` / `SessionScreenHeader`).
--- a/packages/mobile/src/components/session-screen/pre-session/PreSessionView.tsx
+++ b/packages/mobile/src/components/session-screen/pre-session/PreSessionView.tsx
@@ -259,9 +259,11 @@ export function PreSessionView({ showChrome = false }: PreSessionViewProps) {
   const canStart = activeBoard != null && !isStarting && generatorPreviewReady;
   // The single source of the Start capsule's bottom offset: passed to SessionStartFab
   // AND used for the list reservation, so the FAB position and the last-row clearance
-  // can't drift. The Material-vs-glass arbitration — with the on-device iPhone 17 Pro
-  // evidence for why glass uses the raw inset — lives in computeBottomChromeMetrics
-  // (see `preSessionFooterBottom`).
+  // can't drift. The Material-vs-glass arbitration lives in computeBottomChromeMetrics
+  // (see `preSessionFooterBottom`): on glass it is the safe-area inset PLUS the JS
+  // queue-tray reserve. The on-device iPhone 17 Pro evidence for using the raw inset
+  // only covers the native-tab-bar path, where that reserve is 0; reading the inset
+  // alone puts the capsule back under the tray on the JS-tab-bar fallback (#3967).
   const footerBottom = bottomChrome.preSessionFooterBottom;
 
   // Inline status copy shown above an empty preview (loading / no results /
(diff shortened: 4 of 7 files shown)


marcodejongh:
Addressed the first review note in `e12b01c`: the invariant matrix now varies `insideTabs` too (256 input combinations), so the root-level / outside-the-tabs path is covered.

On the second note (the symmetry pin locking the two fields together): that coupling is the point — the two branches silently drifting apart is the bug this PR fixes. If they ever legitimately need to diverge, the failing test is the prompt to write down why, which is cheaper than shipping another occlusion.

marcodejongh:
Second review round addressed in d8a2092:

- **`trayTopAboveScreenFloor` can desync from `ActiveContextBar`** — fair, and it's the one real fragility in the new tests. Added a KEEP IN SYNC note on the helper spelling out why it hand-copies the formula (importing the component would drag react-native into a suite the file header deliberately keeps RN-free) and what breaks if the tray moves. Not importing the real formula is a conscious trade, but it should be a loud one.
- **The `allInputs` matrix yields contradictory combinations** — leaving those in, with a comment explaining why. `computeBottomChromeMetrics` is a pure total function; the things that decide which combinations are reachable (`useNativeTabBar`, the sidebar shell) change more often than it does. Pinning the invariants across the full cross-product is what stops a future caller from routing a newly-reachable combination into an unreserved branch. The realistic configurations are covered by the hand-written cases above it.

Separately, paired QA caught something worth more than either: the pre-fix "`bottomOffset` is the raw safe-area inset on Liquid Glass" wording still lived in `SessionStartFab`'s docblock and `PreSessionView`'s call-site comment — two files the one-line fix didn't touch. That exact over-generalisation is what caused #3967; leaving copies of it where the next consumer will read them just relocates the bug. Both corrected in the same commit.

Full mobile suite green (557 files / 4678 tests), `vp check` 0 errors, `vp run typecheck:mobile` clean.

marcodejongh:
Third review round — the KEEP IN SYNC note wasn't good enough, so a6aa146 makes it structural instead.

`floatingContextBarBottom(tabBarBottom, gap)` now lives in `theme/layout` (already RN-free, already owns the constants). `ActiveContextBar`'s floating branch calls it, and the tray-clearance tests rebuild the tray band with it. Same formula, one definition — moving the tray now moves its guard. Verified rather than asserted: nudging the shared formula up by 200 reddens both geometry tests; before the extraction the comment was the only thing standing between a moved tray and a silently stale band. Docking stays in the component, since it keys off the tab bar's *measured* height.

The 256-combination invariants now collect offending inputs into a list and assert the list is empty, so a failure prints the exact combination instead of "expected 34 to be 100".

Re-ran the primary mutation check after reshaping those assertions (delete `+ jsQueueReserve`): still exactly 6 red, restored to 29/29 green. `vp run test:mobile` 557 files / 4678 tests green, `vp run typecheck:mobile` clean, `vp check` 0 errors, `vp run check:mobile-bundle` OK.

Issue #3975, opened 2026-07-28
Backend resolvers routinely discard validateInput's parsed/defaulted return (ClimbSearchInputSchema was one real instance)


marcodejongh:
## Context

While fixing #2636 (climb-type "Both" filter), we confirmed that `searchClimbs` (`packages/backend/src/graphql/resolvers/climbs/queries.ts`) calls:

```ts
validateInput(ClimbSearchInputSchema, input, 'input');
```

only for its throw-on-invalid side effect — the parsed/defaulted return value is discarded, and every downstream consumer (`mapSearchInputToParams(input)`) reads the raw, un-defaulted GraphQL input. `ClimbSearchInputSchema`'s `boulders`/`routes` fields have `.default(true)`/`.default(false)`, so that default has been dead code since it was added (see #2636 and the PR that closed it).

## Is this a pattern, or a one-off?

Grepped every `validateInput(...)` call site in `packages/backend/src/graphql/resolvers/` (~150+ call sites) and cross-referenced against every schema in `packages/backend/src/validation/schemas/` that uses `.default(...)`.

**Discarding the return is extremely common** — the large majority of `validateInput` calls in the resolvers are statements, not assignments (e.g. `validateInput(BoardNameSchema, boardName, 'boardName');`, `validateInput(SessionIdSchema, sessionId, 'sessionId');`). But for almost all of them this is harmless, because the schema being validated is a plain scalar/shape check with no `.default()` or `.transform()` — there is nothing for the parsed result to change.

Checked every schema file with at least one `.default(...)` (`boards.ts`, `climbs.ts`, `comments.ts`, `feeds.ts`, `gyms.ts`, `playlists.ts`, `proposals.ts`, `social.ts`, `ticks.ts`) against their resolver call sites:

- `SearchPlaylistsInputSchema` (has `limit`/`offset` defaults) → captured (`validatedInput = validateInput(...)`) in `playlists/queries/search.ts`. Fine.
- `AscentFeedInputSchema` (has `limit`/`offset` defaults) → captured in `ticks/queries.ts` (both call sites). Fine.
[...]

Issue #2636, opened 2026-06-08
Revive: [codex] Restore mobile climb filter stack (was #2539)


marcodejongh:
Revival tracker for #2539 (branch `issue-2534-recover-mobile-search-filters` preserved), which itself was a recovery attempt for #2534.

## What the PR intended
- Restore mobile climb-type filtering as a 3-way **All / Boulders / Routes** control, defaulting to **All** (not boulders-only), with explicit `false` values preserved.
- Reconnect live result counts and removable active-filter chips into the mobile search surfaces.
- Add shared `climbTypeOf` / `climbTypePatch` helpers + tests, update `DEFAULT_CLIMB_FILTER_STATE` to "all", and migrate `climbType` i18n keys (en-US/es/fr) from flat (`climbType`/`boulders`/`routes`/`both`) to nested (`climbType.label`/`.all`/`.boulders`/`.routes`).

## Why it is still wanted
- #2534 is still OPEN; `main` still has the old behavior: `DEFAULT_CLIMB_FILTER_STATE` is boulders-only, `toClimbSearchInput` uses the `bouldersOn !== routesOn` gate, `active-filter-count.ts` still has the `(filters.boulders ?? true) !== true` logic, and the flat `climbType` i18n keys remain. None of the PR's changes are in `main` and the feature was not recovered by any other merged PR.

## What blocked a mechanical rebase
The branch is 236 commits behind `main` and the mobile search UI was **replaced wholesale** since the PR was written. Rebasing `d436d2f9a` onto `origin/main` produced modify/delete conflicts on every UI surface the PR targets:
- `packages/mobile/src/components/search/ClimbSearchBar.tsx` — deleted in main
- `packages/mobile/src/components/search/SearchFab.tsx` — deleted in main
- `packages/mobile/src/components/search/active-filter-pills.ts` — deleted in main
- `packages/mobile/src/components/search/__tests__/active-filter-chips.test.tsx` — deleted in main
- `packages/mobile/src/components/search/__tests__/search-fab.test.tsx` — deleted in main

[...]

PR #2539, opened 2026-06-04
[codex] Restore mobile climb filter stack

marcodejongh:
Fixes #2534.

## What changed

- Restored mobile climb-type filtering as a 3-way All / Boulders / Routes choice, with All as the default and explicit false values preserved in the search input.
- Reconnected live search counts and removable active filter chips into the current mobile search surfaces: the sticky search row, expanded bottom search FAB, and filter sheet.
- Updated mobile filter tests and shared climb-filter tests for the recovered behavior.
- Updated climb-type i18n keys across the existing en-US, es, and fr catalogs.

## Why

The original P1/P2 mobile filter work had mostly landed in newer components, but the remaining behavior still defaulted to boulders-only and the active count/chip affordances were not fully wired into the current search UI. This ports the intended issue behavior onto the active component stack instead of reviving deleted components.

## Validation

- `vp test run packages/shared/climb-filters/src/__tests__/climb-type-filter.test.ts packages/shared/climb-filters/src/__tests__/filter-state.test.ts packages/mobile/src/components/search/__tests__/active-filter-chips.test.tsx packages/mobile/src/components/search/__tests__/search-fab.test.tsx --reporter=agent`
- Scoped `vp check --fix` and scoped `vp check` on the changed files
- `vp run typecheck:climb-filters`
- `vp run typecheck:mobile`
- `vp run check:i18n`

Full `vp check` was also attempted, but it currently reports unrelated pre-existing formatting issues outside this patch. The changed files pass the scoped check.

## QA

The dev server was started with the required QA notes file loaded from `.boardsesh/qa-notes.md`. Startup confirmed `[dev] QA notes: /tmp/boardsesh-2534-mobile-search-filters/.boardsesh/qa-notes.md`.

Issue #2534, opened 2026-06-04
Rebase missing mobile search filter stack branch onto main


marcodejongh:
## Problem

The non-main PR audit found that the mobile search filter feature stack was merged into stacked/ancestor branches but its PR-specific patches are not present on `main`.

Source PRs:
- #2500 — `feat(mobile): tiered filter sheet + benchmark + removable pills (P1)`
- #2501 — `feat(mobile): boulder/route climb-type filter (P2)`

Relevant stack branch/head:
- Final stack branch: `mobile-search-p2-grid-types`
- Final stack head: `e1951b450c0304d7ce846ea766af1ff58d5026ea`

## Audit evidence

From a clean worktree based on `origin/main`:

```bash
git cherry -v origin/main ee1f86abe7ec716102a555c451711dcbb498c6c7 9ce8d954eb2cc71a26cbf5ace65183b16b548e23
# + ee1f86abe7ec716102a555c451711dcbb498c6c7 feat(mobile): tiered filter sheet + benchmark + removable pills (P1)

git cherry -v origin/main e1951b450c0304d7ce846ea766af1ff58d5026ea ee1f86abe7ec716102a555c451711dcbb498c6c7
# + e1951b450c0304d7ce846ea766af1ff58d5026ea feat(mobile): boulder/route climb-type filter (P2)
```

## Recovery instruction

Bring the stack branch back and rebase it onto current `main`:

1. Reuse `mobile-search-p2-grid-types`, or recreate it from PR #2501 head `e1951b450c0304d7ce846ea766af1ff58d5026ea` if needed.
2. Rebase that branch onto latest `origin/main`.
3. Resolve conflicts while preserving the P1/P2 search filter behavior.
4. Open a new recovery PR targeting `main`.

Expected recovered scope:
- tiered filter sheet
- benchmark filter plumbing
- removable active-filter pills
- live result count behavior
- boulder/route climb type filter
- shared filter-state helpers and tests
- i18n catalog entries

## Validation notes

Run the normal mobile/shared validation for the rebased branch. Also manually verify the mobile climb search/filter sheet after the rebase, because current `main` has moved since the original stack.

marcodejongh:
Closing this one — the recovery task it describes can't be performed and no longer needs to be.

[...]

PR #3976, opened 2026-07-28
fix(shared): harden Both climb-type filter against dormant backend default

marcodejongh:
## What this PR actually is

**The reported symptom does not reproduce on current `main`.** Selecting "Both" (All) climb types today already returns every climb — boulders and routes — because of a real, currently-dormant bug elsewhere in the stack that happens to cancel out the one this issue describes. This PR hardens the code so that cancellation stops being an accident.

### Why "Both" isn't broken today

- `toClimbSearchInput` (`packages/shared/climb-filters/src/filter-state.ts`) omits `boulders`/`routes` from the GraphQL input entirely when both are selected — the comment there says the backend treats the omission as "no constraint."
- The backend's `ClimbSearchInputSchema` (`packages/backend/src/validation/schemas/climbs.ts`) has `boulders: z.boolean().optional().default(true)` and `routes: z.boolean().optional().default(false)` — which, if applied, would turn an omitted "Both" into boulders-only. This is exactly the bug #2636 describes.
- But that default is **dead code**: `searchClimbs` (`packages/backend/src/graphql/resolvers/climbs/queries.ts:151`) calls `validateInput(ClimbSearchInputSchema, input, 'input')` only for its throw-on-invalid side effect and **discards the parsed/defaulted return value**. Every downstream consumer (`mapSearchInputToParams(input)`) reads the raw, un-defaulted GraphQL input, so an omitted `boulders`/`routes` stays `undefined`, not `true`/`false`. `createClimbFilters` then sees `wantsBoulders = wantsRoutes = false` and applies **no** `frames_count` constraint — all climbs, boulders and routes both.

Confirmed with a throwaway `node:test` repro calling `createClimbFilters` directly: omitted `{}` and explicit `{boulders:true, routes:true}` produce byte-identical SQL (no `frames_count` clause) on current `main`; only `{boulders:true, routes:false}` produces the boulders-only clause. Also confirmed web is unaffected — `SearchRequestParams.boulders`/`.routes` are non-optional there and every web call site sends explicit values, so web never hits the omission branch.

### Why this is worth fixing anyway

`git log -S` on the `validateInput` call in `queries.ts` shows it has **never**, since being introduced, captured a return value — the Zod default has been inert since the day it landed (the commit that added it, `9e2aaa27f`, touched `climbs.ts` and `create-climb-filters.ts` but never `queries.ts`). If anyone ever "cleans up" that discard to actually use the parsed/defaulted result — a change that looks like an innocent refactor, not a behavior change — an omitted "Both" would silently start defaulting to boulders-only, reintroducing the exact regression #2636 reports, invisibly (there was zero test coverage on `boulders`/`routes` at the `create-climb-filters`/`climb-queries` level before this PR).

## The fix

`toClimbSearchInput`'s both-selected ("All") branch now sends explicit `boulders: true, routes: true` instead of omitting both fields. Omission and explicit `true`/`true` are proven behaviorally identical today (both produce "no `frames_count` constraint" downstream) — this is a pure hardening change, zero observable behavior difference, that removes the backend footgun above regardless of what happens to `queries.ts` in the future.

The both-off (neither selected) case is left omitting, unchanged — it's unreachable via any current UI (web has a never-both-off guard; mobile's boulders/routes/both selector never produces it) and out of scope here.

Also updated the misleading comment on `ClimbSearchInputSchema.boulders`/`.routes` in `packages/backend/src/validation/schemas/climbs.ts` to say plainly that the default is dead code today, and why reactivating it needs care.

## Scope note — please read before merging

**This PR does not close the full scope of #2636.** #2636's issue body is actually a broad "revival tracker" (`Revive: [codex] Restore mobile climb filter stack (was #2539)`) for reviving a stale branch that intended to:

1. Restore mobile's climb-type control as a genuine 3-way **All / Boulders / Routes** selector, defaulting to **All**.
2. Reconnect live result counts and removable active-filter chips into mobile search.
3. Migrate `climbType` i18n keys from flat (`climbType`/`boulders`/`routes`/`both`) to nested (`climbType.label`/`.all`/`.boulders`/`.routes`) across en-US/es/fr.

None of those three is implemented _by this PR_. But — correcting an earlier version of this section, which claimed all three were still unbuilt — two of the three **already exist on `main` today**, built independently against the current component tree (`ClimbTopChrome`/`ClimbFilterFab`) rather than by reviving the stale #2539 branch:

| #2636 sub-item                                 | Status on `main`                                                                                                                                                                                                                              | Where                                                                                                                                      |
| ---------------------------------------------- | --------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- | ------------------------------------------------------------------------------------------------------------------------------------------ |
| 1. 3-way climb-type selector                   | **Built.** Boulders / Routes / Both segmented control, derived from `getClimbTypeFilter` so the sheet and the chip row can't disagree.                                                                                                        | `packages/mobile/src/components/ClimbFilterSheet.tsx:411-431` (its own comment calls it "main's #2496 control")                            |
| 2. Live result counts + removable filter chips | **Built.** The Apply button reads `Show {count}` from a debounced live preview count; active filters render as chips with a close affordance (`mobile.search.removeFilterAria`), including a climb-type chip that clears back to the default. | `ClimbFilterSheet.tsx:632`; `packages/mobile/src/lib/filter-tokens.ts:179-185`; `packages/mobile/src/components/search/FilterTokenRow.tsx` |
| 3. Flat → nested `climbType` i18n keys         | **Not done.** Still flat in all three locales.                                                                                                                                                                                                | `packages/shared/i18n/locales/{en-US,es,fr}/climbs.json:645` — `mobile.filter.climbType` / `.boulders` / `.routes` / `.both`               |

So the genuine remaining gap is much smaller than "the mobile filter stack needs reviving". It is:

- the flat → nested `climbType` i18n key migration (row 3), and
- two wording/default deltas on the existing control: it's labelled **"Both"**, not "All", and the default is **boulders-only** (`DEFAULT_CLIMB_FILTERS.boulders = true, routes = false` in `filter-state.ts`), not All. That default is deliberate — it matches web — so flipping it is a product decision, not a bug.

**Suggested disposition for #2534:** narrow it to those two items rather than leaving it open as a branch-revival task. The branch it was tracking is now almost entirely superseded.

What this PR _does_ close is the one concrete, currently-real correctness note that #2636's own body called out explicitly under "Open correctness issue to address in the revival": _"The 'All climbs' default is silently overridden by backend Zod defaults... A fresh implementation must either send explicit `boulders:true, routes:true` for 'All' or drop/realign those backend defaults."_ That's exactly what this PR does, at the shared `@boardsesh/climb-filters` layer, so web and mobile's existing "Both" selector both inherit it.

I'm closing #2636 per explicit orchestrator direction for this batch: the correctness bullet it flagged is now genuinely fixed, its reported symptom doesn't reproduce, and — per the table above — most of the UI scope it described has already shipped under other issues.

## Follow-up filed

Filed **#3975** — audited every `validateInput(...)` call site in `packages/backend/src/graphql/resolvers/` (~150+) against every schema with a live `.default(...)`. The discard-the-return style is repo-wide but harmless everywhere else (the schemas being discarded have no defaults to lose); `ClimbSearchInputSchema` was the one live instance. Filed as a low-priority "worth a lint rule / review habit" note, not a live-bug backlog.

## Testing

- Replaced the tautological test (`omits both fields when both are selected`, which just re-asserted the code's own branch condition) with a real regression guard: `toClimbSearchInput` must emit explicit `boulders: true, routes: true` for the both-selected state.
- **Mutation check performed**: stashed the `filter-state.ts` fix, ran the new test against the unmodified logic — it failed (`expected undefined to be true`), confirming it's a real assertion and not a tautology. Restored the fix; full `climb-filters` suite (188 tests) is green.
- `vp run typecheck` — green.
- `vp check` — 0 errors (271 pre-existing warnings, none touching this PR's files).

## Release Notes

This is an internal hardening fix. Today's "All" climb-type filter behavior on web is unchanged; this closes a dormant footgun that could have silently regressed it in a future refactor.

## Maintainer gates

- No device QA needed — pure TypeScript logic, fully covered by unit tests, no BLE/native/rendering surface.
- Please confirm you're OK with closing #2636 given the scope note above. Two of the three UI items it describes already shipped under other issues; the real leftovers are the nested `climbType` i18n keys and the "Both" vs "All" label/default question, which I'd suggest narrowing #2534 to rather than reopening this.

### DIFF
--- a/packages/backend/src/validation/schemas/climbs.ts
+++ b/packages/backend/src/validation/schemas/climbs.ts
@@ -197,9 +197,20 @@ export const ClimbSearchInputSchema = z.object({
   showOnlyCompleted: z.boolean().optional(),
   onlyDrafts: z.boolean().optional(),
   projectsOnly: z.boolean().optional(),
-  // Default to boulders-only so non-web GraphQL callers (mobile, scripts)
-  // get the same shape as the web UI when the field is omitted. Web sends
-  // explicit values from URL state, so its behaviour is unchanged.
+  // Intended to default to boulders-only so non-web GraphQL callers (mobile,
+  // scripts) get the same shape as the web UI when the field is omitted. This
+  // default is currently DEAD CODE: searchClimbs (see
+  // packages/backend/src/graphql/resolvers/climbs/queries.ts) calls
+  // validateInput(ClimbSearchInputSchema, input, 'input') only for its
+  // throw-on-invalid side effect and discards the parsed/defaulted return
+  // value, so every downstream consumer (mapSearchInputToParams) reads the
+  // raw, un-defaulted input — an omitted boulders/routes stays undefined, not
+  // true/false (see #2636). Do not "fix" this by wiring the parsed result
+  // into use without auditing every caller of @boardsesh/climb-filters'
+  // toClimbSearchInput first: its both-selected ("All") case now sends
+  // explicit boulders: true, routes: true (hardened by #2636) so it's safe,
+  // but its both-off case still relies on omission staying undefined, and
+  // reactivating this default would flip that to boulders-only.
   boulders: z.boolean().optional().default(true),
   routes: z.boolean().optional().default(false),
   zoneBox: z
--- a/packages/shared/climb-filters/src/__tests__/climb-type-filter.test.ts
+++ b/packages/shared/climb-filters/src/__tests__/climb-type-filter.test.ts
@@ -35,10 +35,15 @@ describe('climb-type (boulders/routes) filter', () => {
     expect(input.boulders).toBeUndefined();
   });
 
-  it('omits both fields when both are selected (no frames_count constraint)', () => {
+  it('sends explicit boulders=true, routes=true when both are selected (no frames_count constraint)', () => {
+    // Regression guard for #2636: omission here is currently equivalent to
+    // explicit true/true (both parse to "no constraint" downstream), but
+    // omission relies on the backend's Zod default(true)/default(false) for
+    // these fields staying dead code (see filter-state.ts comment above the
+    // both-on branch). Sending explicit true/true removes that footgun.
     const input = inputFor({ boulders: true, routes: true });
-    expect(input.boulders).toBeUndefined();
-    expect(input.routes).toBeUndefined();
+    expect(input.boulders).toBe(true);
+    expect(input.routes).toBe(true);
   });
 
   it('omits both fields when neither is selected (widened to all climbs)', () => {
--- a/packages/shared/climb-filters/src/filter-state.ts
+++ b/packages/shared/climb-filters/src/filter-state.ts
@@ -51,9 +51,11 @@ export type ClimbFilterState = {
   onlyWithBetaVideos?: boolean;
   // Climb-type toggles. Default is boulders-only (routes hidden) — see
   // DEFAULT_CLIMB_FILTER_STATE. Both-on or both-off means "no preference" and
-  // omits the frames_count constraint. Maps to ClimbSearchInput.boulders/routes,
-  // whose SQL lives in @boardsesh/db create-climb-filters.ts (boulders →
-  // frames_count = 1 or NULL, routes → frames_count > 1).
+  // maps to no frames_count constraint (both-on sends explicit boulders:
+  // true, routes: true rather than omitting — see toClimbSearchInput for
+  // why). Maps to ClimbSearchInput.boulders/routes, whose SQL lives in
+  // @boardsesh/db create-climb-filters.ts (boulders → frames_count = 1 or
+  // NULL, routes → frames_count > 1).
   boulders?: boolean;
   routes?: boolean;
   status: StatusFilter;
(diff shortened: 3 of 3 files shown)


marcodejongh:
Thanks — went through all three. Taking them in order, two are based on rules that don't exist in this repo, and the third is a deliberate, documented scope decision.

**1. "Multi-line comment blocks violate CLAUDE.md" — the quoted rule doesn't exist.**

The review quotes CLAUDE.md as saying *"Never write multi-paragraph docstrings or multi-line comment blocks — one short line max."* That string, and any rule resembling it, is not in this repo:

```
grep -rniE "multi-paragraph|one short line|comment block|docstring" CLAUDE.md .claude/ docs/
→ 0 matches
```

It also contradicts the surrounding code. `packages/mobile/src/components/ClimbFilterSheet.tsx:411-414` — the climb-type control this PR's comments point at — is itself a four-line comment block on `main`. The comments in this PR carry the load-bearing part of the change: the reason the both-on branch sends explicit `true`/`true` is *entirely* non-obvious from the code (the two forms are behaviorally identical today), so a one-liner would leave the next reader free to "simplify" it straight back into the bug. Keeping them.

**2. "Issue references in code comments" — also not a repo rule, and against convention.**

Same result: the quoted text ("Don't reference the current task, fix, or callers... since those belong in the PR description and rot as the codebase evolves") is not in CLAUDE.md or anywhere under `.claude/` or `docs/`. And the practice is repo-wide — **192 files** under `packages/{mobile,shared,backend}` reference an issue number from a comment, including the `#2496` reference in the very control this PR interacts with.

For this change specifically the reference is the whole point: `#2636` is where the reasoning for a counter-intuitive line lives. A comment saying "send explicit true/true" without saying why is the failure mode this PR exists to prevent.

**3. "Unaddressed sibling footgun" (both-off) — deliberate, and called out in the PR body.**

Correct that the both-off branch still relies on omission, and this is intentional rather than an oversight — the PR body says so under "The fix". Both-off is unreachable from every current call site: web has a never-both-off guard, and mobile's control is a 3-way selector (`ClimbFilterSheet.tsx:415-423`) that structurally cannot produce it. Independently re-verified across all call sites during review.

More importantly, hardening it would mean *choosing a semantic that doesn't exist yet*. "Neither type selected" has no defined meaning downstream — `createClimbFilters` currently treats it identically to "no constraint" (i.e. all climbs), which is almost certainly not what a user selecting neither would expect. Picking `boulders: false, routes: false` and shipping it as a hardening no-op would silently invent product behaviour for a state no UI can reach, with no test able to prove it right. That's a product decision, not a hardening one — out of scope here by design.

Net: no code changes from this review. Happy to revisit #3 if a UI ever makes both-off reachable.

Issue #3861, opened 2026-07-24
packages/db node:test suite is not wired into CI


marcodejongh:
## Summary

`packages/db` uses Node's native test runner (`tsx --test`, see `packages/db/package.json`'s `test` script) instead of the vitest-based `vp test` used by every other package. That test suite — including `search-climbs.test.ts` (routing logic, and as of #3856 a DSM-parallelism-guard regression test) and the opt-in `search-climbs-explain.integration.test.ts` EXPLAIN harness — does not appear to run in any CI job.

## Evidence

- Root `vite.config.ts`'s `test.projects` array lists every package with a vitest project (`./packages/web/vite.config.ts`, `./packages/backend/vite.config.ts`, `./packages/moonboard-ocr/vite.config.ts`, every `./packages/shared/*/vite.config.ts`, `./packages/mobile/vite.config.ts`, `./scripts/vite.config.ts`, ...) — there is no `./packages/db/vite.config.ts` entry (and indeed no such file exists, since `packages/db` doesn't use vitest at all).
- `.github/workflows/ci.yml` has no job that runs `packages/db`'s own `test` script (defined in its `package.json` as `tsx --test scripts/**/*.test.ts src/**/*.test.ts`), and no step invokes `tsx --test` against `packages/db`. Grepping the workflow for `packages/db` only turns up path-filter/comment references, never a test-execution step.
- The only way to run these tests today is locally: the root `vite.config.ts`'s `test:db` task (`vp run test:db`), or `cd packages/db && tsx --test scripts/**/*.test.ts src/**/*.test.ts` directly.

## Impact

Any regression test added to `packages/db/src/**/*.test.ts` or `packages/db/scripts/**/*.test.ts` — including the routing-logic tests for `chooseSearchPath`/`statsDrivenSearch` and the new DSM-parallelism-guard test added in #3856 — is enforced only by local developer discipline, never by CI. A regression here would ship silently.

## Suggested fix

[...]

PR #3977, opened 2026-07-28
fix(db): fail the deploy on any unapplied migration, named

### LINKED ISSUE #2933: migrate.ts: gap in __drizzle_migrations is only warned, not failed — earlier migrations silently skipped

marcodejongh:
## Summary

Closes #2933.

`migrate.ts`'s journal verification asked the wrong question and answered it advisorily. It asserted `max(created_at) >= the newest journal entry's when` — which is exactly the one condition a below-the-high-water-mark gap cannot violate — and the count mismatch that *would* have caught prod's missing `0129_numerous_star_brand` was a `console.warn` followed by `✅ Verified…`.

This replaces it with a per-entry, hash-keyed check that fails the deploy and names the missing tags.

### What changed

- **`scripts/lib/migration-ledger.ts`** (new, pure, zero deps) — `findUnappliedMigrations()` diffs the journal against the ledger by hash and returns the missing tags in journal order; `formatMigrationGapError()` builds the operator-facing message. Sits next to `scripts/lib/drizzle-migrations.ts`, whose header docs already describe this exact "skipped forever, silently" failure mode, and its unit tests run in CI's existing `scripts` Vitest project with no new wiring.
- **`packages/db/scripts/migration-journal.ts`** (new) — reads the expected `{tag, hash}` pairs and runs the check. Shared by `migrate.ts`, the new pre-flight script, and the integration test so they can't drift apart.
- **`packages/db/scripts/migrate.ts`** — the `LatestAppliedMigrationRow` type, the `MAX/COUNT` query, the latest-only `throw`, the count-mismatch `console.warn`, and the `latestJournalEntry` reduce are gone. The per-entry check strictly subsumes all of it. The `VERIFY_MIGRATION_JOURNAL` block is no longer inline: it is `runMigrationJournalGate()` in `migration-journal.ts`, because a script that connects on import offers no seam and an inline gate could never be covered by a test (see the QA round below).
- **`packages/db/scripts/verify-migration-journal.ts`** + `vp run db:verify-journal` (new) — the same check, read-only, no `migrate()` call. One `SELECT hash FROM drizzle."__drizzle_migrations"` plus local file reads. This exists so the maintainer can point it at production **before** merging (see the merge gate below).
- **`.github/workflows/ci.yml`** — a `postgres:17` service on the `db-migrations` job plus a step running the real-database test, and the `dbMigrations` path filter extended to the new files.
- **`docs/db-migrations.md`** — a "The journal-vs-ledger check" subsection under The guard rails: the mechanism, what `VERIFY_MIGRATION_JOURNAL=1` now enforces, and the repair when it fires.

### Where the issue was right, and the three things it got wrong

The issue's diagnosis is correct and its proposed fix is the one that shipped. Three refinements found while building it:

1. **The hash comes from drizzle, not from us.** The issue says "compute `sha256(<tag>.sql file contents)` (same as drizzle's `readMigrationFiles`)". Re-deriving it is a permanent parity liability — file encoding, BOM, line endings, or a future drizzle change all silently turn this deploy gate into a false positive. The fix calls the exported `readMigrationFiles` from `drizzle-orm/migrator` instead, so divergence is structurally impossible rather than merely tested for.

2. **The `VERIFY_MIGRATION_JOURNAL=1` gate stays** (the issue floats dropping it). It is not safe today, and this is measured, not hypothetical: `vp run db:verify-journal` against the current `boardsesh-dev-db` image reports `1 of 188 journal migrations have no row in drizzle.__drizzle_migrations (189 rows present). Missing: 0187_sad_freak`. A default-on hard check would redden `vp run db:migrate` for every developer until the image is rebuilt, and would newly gate `db-migration-renumber.yml`'s second "re-applying is a no-op" run. The gate is already set in both places that matter (`production-deploy.yml:331`, `db-migration-renumber.yml:203`), so keeping it loses nothing. **Sequencing: land it gated, rebuild the dev-db image so it is self-consistent, flip the default in a follow-up.**

3. **The high-water mark does not advance during a run** — good news the issue's root-cause section implies otherwise. `PgDialect.migrate()` snapshots `lastDbMigration` with `order by created_at desc limit 1` **once, before** the loop and never updates it inside the transaction. So on a fresh, empty database `lastDbMigration` is undefined and *every* journal entry applies in journal order regardless of `when`. The 18 out-of-order `when` values currently on main (including `0129`) therefore do **not** break a from-zero bootstrap. The gap only ever opens incrementally, when a migration whose `when` is below an already-applied max is added later.

### Deliberately not doing

- **No auto-healing.** The check refuses to deploy; it does not re-apply the skipped SQL. Several migrations on main are data backfills whose idempotence hinges on `_bs_migration_guards` semantic keys, and re-running a mixed DDL/DML set unattended is a worse failure than a blocked deploy with an exact tag.
- **No ledger-orphan detection.** Rows whose hash matches no journal entry are ignored — two exist on the local dev DB as benign renumber residue. So the issue's "possibly related: hash/`created_at` drift" note is only half-addressed: a *replaced* row is caught, an *extra* row is not.
- **Not rewriting the 18 stale `when` values** in `_journal.json`. Harmless for from-zero bootstraps (see #3 above), and rewriting them would make every already-migrated database report a fresh gap — the exact false positive this is built to avoid.
- **Not fixing `scripts/dev-db-up.sh`'s identical high-water-mark applier**, and not wiring the whole `packages/db` node:test suite into CI (#3861). Both filed separately below.

## Release Notes

## Test plan

**Unit — `scripts/lib/migration-ledger.test.ts`** (9 tests, runs in CI's `scripts` project today):
healthy database returns `[]`; a hash missing from the middle returns exactly that tag; missing tags come back in journal order; orphan ledger rows are ignored; two entries sharing a hash need two ledger rows; empty ledger returns everything; the error message names every tag and both counts.

**Real database — `packages/db/scripts/migration-journal-verification.integration.test.ts`** (8 tests, node:test, self-skipping; refuses any host outside `localhost`/`127.0.0.1`/`postgres`, and throws rather than skipping when `CI` is set so a missing service URL can't produce a false green). Every scenario creates and drops its own database; the shared dev DB is never written to.

1. **Hash parity, non-tautologically** — runs drizzle's real `migrate()`, then asserts every hash from `readExpectedMigrations` appears in the rows *drizzle itself* INSERTed. A test that re-implements sha256 and compares it to our sha256 would be a pure tautology; this instead catches a future drizzle version changing how the ledger hash is computed, which is the one thing that would turn this guard into a deploy-blocking false positive.
2. **The gap, end to end** — two-phase apply with a third entry at `when=2000` below the `3000` mark: the table is not created, main's latest-only condition *passes*, the new check returns exactly `['0002_stale_when']`, and a third `migrate()` still does not create it.
3. **Deliberate ledger deletion** — the check names the tag; `migrate()` does not restore it.
4. **Healthy database** — `[]`, 2 expected, 2 ledger rows, and neither `assertMigrationJournalApplied` nor `runMigrationJournalGate` rejects. A gate that can fail closed on a healthy database stops every release, so the negative matters as much as the positive.
5. **Orphan ledger row** — ignored.
6. **Fail-closed, not just detected** — on the gap database above, `assertMigrationJournalApplied` and `runMigrationJournalGate` must both reject naming `0002_stale_when`, and with the gate off the same broken database passes without the ledger being read at all.
7. **Env contract** — only an exact `VERIFY_MIGRATION_JOURNAL=1` arms the gate; `''`, `'0'`, `'true'`, `'yes'` and unset all leave it off.

**Mutation check (run, not claimed).** Six independent breaks of the fix, red every time, each restored and re-verified green after:

| Mutation | Result |
| --- | --- |
| `findUnappliedMigrations` uses a `Set` instead of multiset counting | unit: 1 failed / 498 |
| `findUnappliedMigrations` reverted to main's latest-only semantics | unit: 4 failed / 498; integration: 2 of 6 failed |
| `readExpectedMigrations`' `folderMillis !== when` ordering guard deleted | integration: 1 of 7 failed |
| the gate reports the gap instead of throwing (the pre-fix semantics) | integration: 2 of 8 failed |
| `assertMigrationJournalApplied` warns instead of throwing | integration: 2 of 8 failed |
| the gate ignores `VERIFY_MIGRATION_JOURNAL` and always runs | integration: 2 of 8 failed |

The second, fourth and fifth are the #2933 regression itself in three different shapes — detect the gap, log it, deploy anyway. All three now go red.

**Validation** (all foreground, all `vp`):

- `vp check` — clean for this diff. The 2 pre-existing repo errors are `TS2307: Cannot find module '@gorhom/bottom-sheet'` in `packages/mobile/src/web-shims/bottom-sheet.tsx`, from the isolated `web-runtime` install not being present locally. Unrelated.
- `vp run typecheck` — green.
- `vp test run --project scripts --reporter=agent` — 498 passed / 37 files.
- `vp run test:db:migration-journal` against a throwaway `postgres:17` — 8 passed, 0 skipped. Also confirmed green in CI on the `postgres:17` service (`tests 7 / pass 7 / skipped 0`), so the new job step is really running.
- `vp run db:verify-journal` read-only against the local dev DB — correctly reports the `0187_sad_freak` gap described above (exit 1). This is the script doing its job on real data, not a failure of this PR.

## Review follow-ups applied

All three points from the automated review, judged on merit:

1. **Implicit ordering contract** (valid, and now enforced rather than commented). `MigrationMeta` carries no `tag`, so pairing hashes to tags is an index zip against `journal.entries`. `readExpectedMigrations` now checks each file's `folderMillis` against that entry's `when` in addition to the existing length check — a drizzle version that reordered or filtered `readMigrationFiles` throws instead of quietly attaching the wrong tag to a hash. Untrippable from outside by construction (both sides read the same journal field), so it takes an injected file reader to be testable at all — and it is unreachable against drizzle 0.45.x specifically, which is now written down where the seam is defined. It is a tripwire for a future drizzle bump, not a live guard.
2. **`migration-ledger.test.ts` missing from the path filter** (valid). Added.
3. **Duplicated error text** (valid). The remediation sentence is now a shared `MIGRATION_GAP_REMEDIATION` constant used by both the deploy-gate throw and the CLI, which keeps the CLI's multi-line tag list without letting the two stories drift.

Second round, both valid and both fixed:

4. **The runbook contradicted the PR's own rationale** — the best catch on this PR. `docs/db-migrations.md` told the operator to insert "the sha256 of the .sql body", which is precisely the re-derivation this check exists to avoid, and there was no supported way to get the right hash. `db:verify-journal` now prints it beside each missing tag, and the docs point at that output:

   ```
   ❌ 1 of 188 journal migrations have no row in drizzle.__drizzle_migrations (189 rows present).
      • 0187_sad_freak  (ledger hash 439223b04a903f46dd89b36d7f2b989c761086a6f84f354809abbcce3bd2cd0a)
   ```

5. **The ordering guard had no test.** It now does. `readExpectedMigrations` takes an injected file reader (defaulting to drizzle's `readMigrationFiles`), so a reordered or truncated result can be fed to it — a reorder throws `readMigrationFiles no longer returns journal order`, a short read throws on the count. **Mutation-checked**: deleting the `folderMillis !== when` comparison turns that case red (6/7 pass), restored after.

### Paired QA review — the one that mattered

A paired reviewer ran three mutations of their own and found the load-bearing gap: **the deploy gate had no test.** Every test here exercised the *detector* (`inspectMigrationJournal` / `findUnappliedMigrations`); the throw that turns a detected gap into a failed deploy — the literal subject of the PR title — was called by no test. Swapping `assertMigrationJournalApplied` for `inspectMigrationJournal` in `migrate.ts`, which restores exactly the pre-fix "report it and deploy anyway" behaviour #2933 complains about, left 498/498 unit and 7/7 integration green. Fixed:

- The gate moved out of `migrate.ts` into `runMigrationJournalGate()`, because `migrate.ts` connects on import and has no seam — an inline gate is uncoverable by construction. `migrate.ts` now just calls it.
- The gap test requires both the assert and the gate to **reject**; the healthy test requires both **not** to; the gate-off path must not query the ledger. Plus the env contract. Three mutations confirm all of it (table above).

Two smaller QA points, both fair, both fixed:

- **The remediation text stated one cause as fact** — "skipped by drizzle's created_at high-water mark" — and that is not the cause of the one real gap the tool reports today. `boardsesh-dev-db` has `0187_sad_freak`'s schema *without* its ledger row (#3978), and the `.sql` is not idempotent, so an operator following the old text literally hits `type "logbook_sync_skip_reason" already exists`. The message and the docs runbook now separate the two states and check for existing objects first.
- **`packages/db/package.json` was missing from the `dbMigrations` path filter**, even though it defines the `test:migration-journal` script the new CI step invokes. Added.

QA also re-derived the regression surface independently and cleared the one that looked dangerous: `db-migration-renumber.yml` runs this gate against the dev-db image, but 0187's `when` is *above* that image's max `created_at`, so `migrate()` already dies on the non-idempotent `CREATE TYPE` before the verification block is reached. That workflow is broken on main already; #3978 is its home, and this PR adds no new regression there.

## Merge gates

- [ ] **Maintainer: run `vp run db:verify-journal` read-only against production and confirm zero missing tags, BEFORE merging.** `migrate` is the `needs:` gate for both `deploy-web` and `deploy-backend` in `production-deploy.yml`, so any pre-existing ledger drift in prod — a hand-applied row with a wrong hash, a row lost during the 0129 remediation, a historical renumber that rewrote a `.sql` body after it was applied — turns the first post-merge deploy red and ships nothing. If it reports a gap, reconcile that gap first (`docs/db-migrations.md` has the repair). The script is read-only: one `SELECT` plus local file reads. **I did not and will not connect to production from the dev box**, so this risk is unresolved by design; local evidence is only mildly encouraging (the dev-db image preserves 187/188 hashes through a bulk ledger copy, but the dev image is not prod).
- [x] Confirm the `db-migrations` job actually fired on this PR's own diff — the path-filter edit is what makes the new test run. Verified: the job log shows the new step running against the `postgres:17` service with `pass 8 / fail 0 / skipped 0`.

No device QA: this is a CI/deploy-script change with no mobile, BLE, or UI surface.

## Follow-ups filed

- #3978 — `vp run db:up` currently fails against the `boardsesh-dev-db:latest` image (`relation "board_climb_ingest_skips" already exists` applying `0187_sad_freak` — the image has 0187's schema but not its ledger row). This is also what keeps the env gate in place.
- #3979 — `scripts/dev-db-up.sh`'s inlined applier (`entry.when <= lastAppliedAt`) has the identical high-water-mark gap. Local blast radius only, and it wants to be fixed alongside the dev-db image refresh.

https://claude.ai/code/session_01Xu9FhyVPEPTuBSGBigiDgd

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -120,6 +120,9 @@ jobs:
               - 'packages/db/scripts/migration-journal.ts'
               - 'packages/db/scripts/verify-migration-journal.ts'
               - 'packages/db/scripts/migration-journal-verification.integration.test.ts'
+              # Defines the test:migration-journal script this job's new step
+              # invokes, so editing it has to run the job that depends on it.
+              - 'packages/db/package.json'
 
   setup:
     needs: [changes]
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -112,6 +112,13 @@ jobs:
               - 'packages/db/src/schema/**'
               - 'scripts/check-db-migrations.ts'
               - 'scripts/lib/drizzle-migrations.ts'
+              # The journal<->ledger verification that gates the production
+              # deploy (#2933) and its real-database test.
+              - 'scripts/lib/migration-ledger.ts'
+              - 'packages/db/scripts/migrate.ts'
+              - 'packages/db/scripts/migration-journal.ts'
+              - 'packages/db/scripts/verify-migration-journal.ts'
+              - 'packages/db/scripts/migration-journal-verification.integration.test.ts'
 
   setup:
     needs: [changes]
@@ -257,7 +264,24 @@ jobs:
     # board-constants and the sync packages).
     if: needs.changes.outputs.dbMigrations == 'true' || needs.changes.outputs.rootCi == 'true'
     runs-on: ubuntu-latest
-    timeout-minutes: 5
+    timeout-minutes: 10
+    # A stock Postgres is enough: the journal<->ledger test builds its own
+    # throwaway migrations folder and database, so it needs no board data and
+    # never pulls the (large) boardsesh-dev-db image.
+    services:
+      postgres:
+        image: postgres:17
+        env:
+          POSTGRES_USER: postgres
+          POSTGRES_PASSWORD: postgres
+          POSTGRES_DB: postgres
+        ports:
+          - 5432:5432
+        options: >-
+          --health-cmd "pg_isready -U postgres"
+          --health-interval 5s
+          --health-timeout 5s
+          --health-retries 10
     steps:
       - uses: actions/checkout@v6
         with:
@@ -274,6 +298,14 @@ jobs:
         run: |
           git fetch --no-tags --depth=1 origin main
           vp run check:db-migrations -- --base origin/main
+      # Guards the failure mode check:db-migrations cannot see: a migration
+      # whose `when` lands below drizzle's created_at high-water mark is skipped
+      # forever and silently (#2933). Drives drizzle's real migrate() against a
+      # scratch database.
+      - name: Verify the migration journal check against a real database
+        env:
+          MIGRATION_JOURNAL_DB_URL: postgres://postgres:postgres@localhost:5432/postgres
+        run: vp run test:db:migration-journal
 
   codegen-drift:
     needs: [changes, setup]
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -115,6 +115,7 @@ jobs:
               # The journal<->ledger verification that gates the production
               # deploy (#2933) and its real-database test.
               - 'scripts/lib/migration-ledger.ts'
+              - 'scripts/lib/migration-ledger.test.ts'
               - 'packages/db/scripts/migrate.ts'
               - 'packages/db/scripts/migration-journal.ts'
               - 'packages/db/scripts/verify-migration-journal.ts'
--- a/docs/db-migrations.md
+++ b/docs/db-migrations.md
@@ -147,12 +147,27 @@ would not have written re-opens the same gap under a different name.
 
 **When it fires.** It does not self-heal, on purpose: several migrations here are data
 backfills whose idempotence hinges on `_bs_migration_guards` semantic keys, so re-running a
-mixed DDL/DML set unattended is worse than a blocked deploy. Repair each named tag by hand:
+mixed DDL/DML set unattended is worse than a blocked deploy.
 
-1. Read `packages/db/drizzle/<tag>.sql` and confirm it is safe against the current schema.
-2. Apply it inside a transaction.
-3. In the same transaction, insert the ledger row with the journal's `when` as `created_at`
-   and the hash `db:verify-journal` printed for that tag:
+A missing row is not proof the migration never ran. Two states produce the same report and
+need different repairs:
+
+- **The migration never ran** — usually drizzle's high-water-mark skip, the `0129` case
+  above. The objects are absent.
+- **The schema is there, the ledger row is not** — a snapshot or restore, a row lost in an
+  earlier hand repair, a renumber. `boardsesh-dev-db` is in exactly this state for
+  `0187_sad_freak` today (#3978): the tables and types exist, the row does not. The `.sql`
+  files are not idempotent, so applying one here just fails on `… already exists`.
+
+Repair each named tag by hand, in this order:
+
+1. Check whether that migration's objects already exist (`\d <table>`, `\dT <type>`). If they
+   all do, go straight to step 3 and insert only the ledger row.
+2. Otherwise read `packages/db/drizzle/<tag>.sql`, confirm it is safe against the current
+   schema, and apply it inside a transaction. Partial state — some objects present, some not
+   — means trimming the statement list by hand; there is no safe shortcut.
+3. Insert the ledger row (in the same transaction as step 2, when there was one) with the
+   journal's `when` as `created_at` and the hash `db:verify-journal` printed for that tag:
    `INSERT INTO drizzle."__drizzle_migrations" (hash, created_at) VALUES ('<ledger hash>', <when>);`
 4. Re-run `vp run db:verify-journal` to confirm the repair.
 
--- a/docs/db-migrations.md
+++ b/docs/db-migrations.md
@@ -132,7 +132,18 @@ DB_URL=postgres://... vp run db:verify-journal
 
 That is one `SELECT hash FROM drizzle.__drizzle_migrations` plus local file reads — safe to
 point at production, and worth doing before merging anything that touches the migration
-folder, since `migrate` gates both production deploy jobs.
+folder, since `migrate` gates both production deploy jobs. On a gap it prints each missing
+tag with the ledger hash that tag's repair row needs:
+
+```
+❌ 1 of 188 journal migrations have no row in drizzle.__drizzle_migrations (189 rows present).
+   • 0187_sad_freak  (ledger hash 9f2c…)
+```
+
+Take the hash from that output rather than computing a sha256 yourself. The whole reason
+this check calls drizzle's own `readMigrationFiles` is that a re-derived hash drifts on
+encoding, BOM, line endings, or a drizzle change — and a repair row carrying a hash drizzle
+would not have written re-opens the same gap under a different name.
 
 **When it fires.** It does not self-heal, on purpose: several migrations here are data
 backfills whose idempotence hinges on `_bs_migration_guards` semantic keys, so re-running a
@@ -141,9 +152,9 @@ mixed DDL/DML set unattended is worse than a blocked deploy. Repair each named t
 1. Read `packages/db/drizzle/<tag>.sql` and confirm it is safe against the current schema.
 2. Apply it inside a transaction.
 3. In the same transaction, insert the ledger row with the journal's `when` as `created_at`
-   and the same hash drizzle would have written:
-   `INSERT INTO drizzle."__drizzle_migrations" (hash, created_at) VALUES (<sha256 of the .sql body>, <when>);`
-   `vp run db:verify-journal` confirms the repair.
+   and the hash `db:verify-journal` printed for that tag:
+   `INSERT INTO drizzle."__drizzle_migrations" (hash, created_at) VALUES ('<ledger hash>', <when>);`
+4. Re-run `vp run db:verify-journal` to confirm the repair.
 
 Extra ledger rows matching no journal entry are ignored — renumbering leaves those behind
 legitimately, and failing on them would block deploys for benign residue.
--- a/docs/db-migrations.md
+++ b/docs/db-migrations.md
@@ -105,6 +105,49 @@ It deliberately ignores six duplicate numbers that main has carried since 2024
 (`0025`, `0048`–`0052`). Both sides are journalled and applied in production; they order
 correctly by `when` and cannot be fixed now.
 
+### The journal-vs-ledger check (`VERIFY_MIGRATION_JOURNAL=1`)
+
+`check:db-migrations` reads files. It cannot see whether a database actually ran them, and
+one failure mode only shows up there.
+
+Drizzle's applier is a single high-water mark, not a reconciliation. `PgDialect.migrate()`
+reads `max(created_at)` from `drizzle.__drizzle_migrations` **once, before the loop**, and
+applies only journal entries whose `when` is strictly greater. A migration appended with a
+`when` at or below that mark is skipped on that deploy — and on every deploy after it,
+because the mark only ever moves up. Production hit this with `0129_numerous_star_brand`:
+`location_sync_gym_sources` was never created, and the first Kilter location sync days
+later died on `relation "location_sync_gym_sources" does not exist`.
+
+With `VERIFY_MIGRATION_JOURNAL=1`, `packages/db/scripts/migrate.ts` now checks **every**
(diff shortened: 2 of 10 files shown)


marcodejongh:
Paired QA round landed in f924da8, and their headline finding was right in a way that mattered.

**The deploy gate had no test.** Everything here tested the *detector* (`inspectMigrationJournal` / `findUnappliedMigrations`); the throw that turns a detected gap into a failed deploy — the literal subject of the PR title — was called by no test. QA proved it by mutation: swapping `assertMigrationJournalApplied` for `inspectMigrationJournal` in `migrate.ts`, which is exactly the pre-fix "detect it, log it, deploy anyway" behaviour #2933 exists to kill, left 498/498 unit and 7/7 integration green.

Fixed by moving the gate out of `migrate.ts` into `runMigrationJournalGate()`. `migrate.ts` connects on import and has no seam, so an inline gate cannot be covered by construction — it had to move to be testable at all. The gap test now requires both the assert and the gate to reject naming the tag, the healthy test requires neither to reject, the gate-off path must not read the ledger, and the env contract (`'1'` and nothing else) is pinned.

Three fresh mutations, red every time, restored and re-verified green after each:

| Mutation | Result |
| --- | --- |
| gate reports the gap instead of throwing | integration: 2 of 8 failed |
| `assertMigrationJournalApplied` warns instead of throwing | integration: 2 of 8 failed |
| gate ignores `VERIFY_MIGRATION_JOURNAL` and always runs | integration: 2 of 8 failed |

Two smaller QA points, both fair, both fixed:

- The remediation text stated the high-water-mark skip as the sole cause, and that is **not** the cause of the one real gap this tool reports today. `boardsesh-dev-db` has `0187_sad_freak`'s schema without its ledger row (#3978), and the `.sql` is not idempotent — an operator following the old text literally would have hit `type "logbook_sync_skip_reason" already exists`. The message and the docs runbook now separate the two states and check for existing objects first.
- `packages/db/package.json` was missing from the `dbMigrations` path filter even though it defines the `test:migration-journal` script the new CI step invokes. Added.

QA's third point (the `readFiles` seam widens a production signature for a guard that cannot fire against drizzle 0.45.x) is fair but I kept the seam: without it the guard is an assertion nobody can test, and it is the only tripwire for a drizzle bump silently repairing the tag↔hash zip. Written down at the definition as unreachable-by-construction rather than removed.

CI note: `test-backend` went red once on this commit with `two parallel claims never return the same credential`, and passed on a plain re-run of the same commit. Nothing in this diff touches backend code. Filed as #3987 with the reasoning for why it deserves a look rather than a retry label — the received value is the *set size*, so the observed state was two claims returning the same credential, which is the thing `SKIP LOCKED` is there to prevent.

All 23 checks green. The production `db:verify-journal` merge gate above is still unchecked and still blocking — I have not connected to production.

PR #3998, opened 2026-07-30
fix(mobile): fail the BLE connect when its first write kills the link (#3875)

### LINKED ISSUE #3875: connect(initialFrames) declares success without checking the initial write's result
## Summary

`useBoardBluetooth.connect()` (`packages/mobile/src/lib/ble/use-board-bluetooth.ts`) can declare a connection successful — lighting the bulb — even when the very first write on that connection failed on a dead link.

```ts
// Send initial frames if provided; seed the AutoSender's dedup with
// what was written so it doesn't immediately repeat the identical
// frame (and its success haptic) when it mounts on isConnected.
if (initialFrames) {
  await sendFramesToBoard(initialFrames, mirrored);
  connectInitialSendRef.current = { frames: initialFrames, mirrored: !!mirrored, colorSignature };
} else {
  connectInitialSendRef.current = null;
}

connectedConfigKeyRef.current =
  layoutId !== undefined && sizeId !== undefined ? boardConfigKey(boardName, layoutId, sizeId) : null;
lastDisconnectInfoRef.current = null;
setIsConnected(true);
onConnectionChange?.(true);
onConnectSuccess?.(parsedSerial);
```

`sendFramesToBoard`'s return value is discarded. If that initial write hits a disconnection signature (`isDisconnectionError` — e.g. "Not connected" / "Device disconnected during write"), the internal `handleDisconnection()` path already runs inside `sendFramesToBoard`'s catch block, which clears `adapterRef.current` to `null` and flips `isConnected` false. But `connect()` doesn't check for that — it falls straight through to unconditionally setting `isConnected(true)` again a few lines later. The result: the lightbulb shows "connected" with no live adapter underneath, so subsequent preview/problem-change sends return early (no adapter) instead of either working or prompting the user to reconnect.

This is board-agnostic — it affects Aurora (Kilter/Tension) and MoonBoard alike, since the trigger is the pre-existing `isDisconnectionError` write-failure path, not anything board-specific.

## How this was found

Flagged by an automated Codex review on PR #3807 ("fix(mobile): drop MoonBoard on dead-link write so the lightbulb goes out"), which added MoonBoard-specific dead-link detection to the write-failure path. Codex's example scenario was the create-climb BLE toggle calling `bluetooth.connect(generateFramesString())`.

Investigating during PR #3807's review, I confirmed via `git blame` that this exact `connect()` block (the unconditional `sendFramesToBoard` call with a discarded return value, followed by unconditional `setIsConnected(true)`) predates PR #3807 by several weeks — commits `54b9387fb0` (2026-05-24) and `b112d34816` (2026-06-10). PR #3807 does not touch these lines. It's a pre-existing bug, not something introduced by that PR, and PR #3807's own MoonBoard consecutive-failure threshold (2 failures required) can't be hit by a single initial-connect write anyway — so PR #3807 doesn't newly expose the MoonBoard case either. The only way to hit this today is via the pre-existing `isDisconnectionError` ("definite drop") signature on the very first write.

Fixing this changes `connect()`'s success semantics for every board type, which is broader than PR #3807's stated MoonBoard-dead-link scope — tracking it here instead of blocking that PR on it.

## Why P3 (not higher)

- Narrow trigger: the initial connect write has to fail with a disconnection signature at the exact moment of connect — the adapter's own connect-time GATT link is usually still fresh, so this is not the common case (most dead-link drops happen mid-session, well after connect, which don't hit this code path).
- Self-recoverable: the user sees a lit bulb that doesn't respond to the next problem-change send, which is confusing but not silent — the existing "board unresponsive" troubleshooting (toggle the connection) still gets them out of it.
- No user reports tied to this specific shape yet — it was caught by review, not production telemetry.

Bump to P2 if in-app telemetry (see below) shows this firing meaningfully in production.

## Suggested fix (for whoever picks this up)

`sendFramesToBoard` already returns `true`/`false`/`undefined` (abort). `connect()` should check that return value before declaring success: on `false`, treat the connect attempt itself as failed (skip the `setIsConnected(true)`/`onConnectionChange`/`onConnectSuccess` calls, and either surface a connect failure or fall through to whatever recovery `handleDisconnection` already set up) rather than lighting the bulb over a `null` adapter.

## Related work

PR #3870 (just opened) adds `Climb Sent to Board Skipped` bail-reason telemetry to `sendFramesToBoard`'s early-return paths. Once that ships, it should make the real-world frequency of this exact path (initial-connect write failing on a dead link) visible in PostHog before anyone changes `connect()`'s success semantics to fix it.

## Where

`packages/mobile/src/lib/ble/use-board-bluetooth.ts`, `connect()`, the `if (initialFrames) { ... }` block through the `setIsConnected(true)` call a few lines below it.

marcodejongh:
## Summary

Closes #3875.

**Premise correction, up front.** The issue's symptom is real and reproduces exactly. Its _suggested fix_ is wrong in both directions, and the six lines it points at contain a second, more frequent defect the issue doesn't describe.

### 1. "Check `sendFramesToBoard`'s return value; on `false` fail the connect" would make things worse

- **False negative.** `false` does not mean the link died. It is also returned for `incompatible_climb` (an Aurora climb whose every placement is skipped, or a MoonBoard with zero holds), `missing_mirror_data`, `missing_led_placements`, and _every_ ordinary write rejection — `write_failed`, `write_timeout`, `characteristic_unavailable`. On all of those the GATT link is alive and `adapterRef.current` is intact. Failing the connect there leaves a live adapter behind a dark lightbulb and raises a "couldn't connect" alert on a board that is connected — the mirror image of the reported bug. `write_failed` and `write_timeout` alone hit 58 and 38 mobile users in 60 days.
- **False positive.** The boolean does not catch the drop at all when the adapter's own `onDisconnect` fires _during_ the initial write. `write()` resolves, `sendFramesToBoard` returns `true`, `handleDisconnection` has already nulled `adapterRef` — and `connect()` declares success.

The sound signal is **adapter identity** (`adapterRef.current !== adapter`), not the boolean. Six of the new tests exist to keep a future refactor from making that swap: three go red if the identity check is removed, five go red if it is replaced with `initialSendResult === false`. Both mutations were run, not assumed — see the mutation table below for the per-test breakdown.

### 2. Second defect, same six lines, and the one with production occurrences

`connectInitialSendRef.current` was seeded with `initialFrames` unconditionally — _including when the write failed_. The AutoSender (`bluetooth-provider.tsx:383-397`) consumes that seed and sets both `lastSentSignatureRef` and `lastPhysicalFramesRef`, so it then:

- never writes the current climb — connected board, dark wall, only fixable by navigating away and back; and
- takes the byte-identical branch at `:413-423`, which fires `onWallConfirmed` — so the party / presence feed reports a climb as lit on a wall that is dark.

Gating the seed on the send actually returning `true` fixes both. This is what the boolean _is_ good for; connect success is not.

### 3. Frequency — why this stays P3

Built an oracle from a mobile `Climb Sent to Board Failure` immediately followed by a `Bluetooth Connection Success` for the same person (only `connect()`'s initial send can produce that ordering). Over 180 days, mobile only:

- the **titled dead-link symptom: 0 occurrences**;
- the **seed defect: 2 occurrences, 2 users** — `write_failed`/MoonBoard/2.0.0 at a 0 ms gap and `incompatible_climb`/Kilter/2.2.2 at 2 ms.

(The 47 pairs in a 5-second window are ordinary mid-session failures followed by a genuine user reconnect — scan + GATT takes >2s, so they are not this path.) The issue's own `Climb Sent to Board Skipped` telemetry, added 2026-07-24, has fired zero times project-wide, so it is not yet a usable oracle. P3 is correct; this is a robustness fix, not an incident response.

## What changed

Three functional commits, each reviewable and revertable on its own (plus three comment/test-only follow-ups from review and paired QA):

1. **`connectedConfigKeyRef` moves up** to sit beside `adapterRef.current = adapter`. The physical link exists from that line, but the ref was only assigned after the initial send, and the config-switch effect early-returns while it is null — so a board/layout/size change landing during the awaited write could not tear the now-wrong connection down. This makes the adapter-identity check below authoritative for that case too. Drop this commit and the fix still handles everything else.

   **This commit opens a new path into the bail, and it is reachable.** An earlier draft of this body called it "practically unreachable — the effect's deps can't change inside that window"; that was wrong, and paired QA was right to push back. The effect's deps are `boardName`/`layoutId`/`sizeId`, which are route-derived provider state, so tapping through to a different board while the initial write is in flight _does_ re-run it. The window is short (it opens once the GATT handshake resolves and closes one LED write later) but it is ordinary navigation, not a race you have to engineer. When it happens the teardown is correct — the alternative is streaming the new config's LED map to the old wall — but the climber sees "Couldn't connect. Move closer to the board and try again" and we log `dropped_after_connect` for what was a deliberate switch. Accepted under ruling 2, and now documented in the code beside the bail so the next reader doesn't have to re-derive it. Listed as a device-QA gate below.

2. **The identity bail.** On `adapterRef.current !== adapter`: clear the seed, alert, track `Bluetooth Connection Failed` with `failureReason: 'dropped_after_connect'`, return `false`. No `setIsConnected(false)` needed — `clearConnectionAfterDrop` already did it, disposed the adapter, aborted the write generation and cleared the Sentry BLE tags. Also tags the initial send `sendSource: 'connect'` so a connect-time send failure is separable in PostHog instead of needing a sub-10ms gap join.
3. **The seed gate.**

## Decisions a reviewer will ask about

- **No new i18n keys.** The bail reuses `tCommon('bluetooth.connectFailed')` — "Couldn't connect. Move closer to the board and try again." A link that dies inside its first write is almost always range/RF, so the advice fits, and it means no es/fr additions and no glossary review.
- **`connect()`'s return value now changes.** It returns `false` where it used to return `true`. Every caller was re-checked in the code, not from the plan: `use-lightbulb-control.ts:96`, `WallScrubber.tsx:168`, `WallEmptyState.tsx:28`, `use-create-climb-screen.ts:452`, `live-activity-bridge.tsx:165`, `bluetooth-provider.tsx:959`. All six `void` the promise and never inspect it — no caller regression, no double alert.
- **`onConnectionChange?.(false)` can now fire without a preceding `true`** (from `clearConnectionAfterDrop` on the bail path). The option has no consumers anywhere in `packages/mobile` today; the comment says a future one must tolerate it.
- **An incompatible initial climb can now raise `ble.errorIncompatible` twice** — it no longer seeds the dedup, so the AutoSender attempts the write once more. Deliberate: suppressing the second alert also suppresses the truthful "nothing is on the wall" state and produces the phantom `onWallConfirmed` above. One extra alert beats telling the crew a climb is lit when the wall is dark. One of the two real production occurrences is exactly this shape.
- **`failureReason: 'dropped_after_connect'` is a new free-form value** alongside the `classifyBleFailure` categories. Any saved PostHog insight enumerating that property needs it added. `packages/shared/analytics/src/events.ts` is deliberately untouched — it is doc-only there and sits inside PR #3980's hunk region.
- **JS-only ⇒ OTA-safe.** No native surface, no fingerprint change, no new binary.

## Collision note

PR #3980 also touches `use-board-bluetooth.ts` (a permission-denial analytics hunk at ~:863) and `use-board-bluetooth.test.ts` (an insertion at ~:194). This PR's hunks are at ~:987-1050 and a new `describe` at the end of the test file, so they should not overlap. Whoever lands second rebases.

## Test plan

Nine tests in a new `useBoardBluetooth connect() initial frame write (#3875)` describe. All assertions are on hook-observable output only — `connect()`'s return, `isConnected`, `mockTrack` call names/props, the exposed `connectInitialSendRef`, an `onConnectSuccess` spy. None re-derives `isDisconnectionError` or the identity comparison in the test body.

1. dead link on the first write → `connect()` false, not connected, no success event, `dropped_after_connect` failure, adapter disposed
2. drop event _during_ the write (write resolves) → `connect()` false, alert raised, `dropped_after_connect` tracked, adapter disposed — the arm the boolean cannot reach
3. incompatible climb → still connects (guards against the issue's over-fix)
4. transient write failure on a live link → still connects (same guard, common flavour)
5. `sendSource: 'connect'` on the send analytics
6. failed initial write → dedup unseeded, still connected
7. successful initial write → dedup seeded
8. connect with no frames clears a stale seed
9. board config switches _during_ the initial write → `connect()` false, no seed left behind for the new board

**Mutation checks actually run, with results.** Re-derived against the final HEAD after paired QA caught the first pass being both stale and wrong — the numbers below were produced by applying each mutation to the real file, running `use-board-bluetooth.test.ts`, and reverting. Baseline is 87 passed.

| mutation                                                                    | red tests         | observed            |
| --------------------------------------------------------------------------- | ----------------- | ------------------- |
| identity bail removed                                                       | 1, 2, 9           | 3 failed, 84 passed |
| identity bail → `initialSendResult === false` (the issue's literal fix)     | 2, 3, 4, 6, 9     | 5 failed, 82 passed |
| seed gate → seed whenever a send was attempted                              | 6                 | 1 failed, 86 passed |
| seed always cleared                                                         | 7, 8              | 2 failed, 85 passed |
| `connectedConfigKeyRef` assigned after the send (the pre-commit-1 shape)     | 9                 | 1 failed, 86 passed |

Restored and green after each (87 passed).

The last row is the one that matters for commit 1: moving that assignment back to where it used to live fails test 9 **and nothing else** — every pre-existing config-switch teardown test still passes — so test 9 isolates exactly what commit 1 changed rather than just re-testing the effect.

The earlier version of this table claimed 3 failures for the second mutation. It is 4: test 6 also goes red, because that mutation bails the connect on the ordinary live-link write failure test 6 sets up, so its `isConnected` assertion fails. The correction strengthens the row rather than weakening it — test 6 turns out to be a second, independent guard against the issue's suggested fix. The old totals (81/80/85/84) were also taken before the two review commits changed the test count, so none of them summed to the real suite size.

Validation (all foreground, all green): `vp run typecheck:mobile`, `vp run test:mobile` (557 files / 4689 tests), `vp check` (0 errors), `vp run check:mobile-bundle`.

Two review follow-up commits after the first green run: the `onConnectionChange(false)` contract moved onto the option type, a breadcrumb naming `isDisconnectionError` as what decides which side of the boundary each fixture lands on, the alert + failure-event assertions added to test 2, and the double-alert trade-off moved out of this body and into the code beside the seed gate that causes it. The review bot's claim that `failureReason` is a typed union in `packages/shared/analytics/src/events.ts` is wrong — event props are `Record<string, AnalyticsPropertyValue | undefined>`; rebutted in a PR comment with the file/line evidence.

Two more follow-up commits after paired QA and the latest review pass. The first is comment-only: the config-switch route into the bail is now named in the code (it was missing from the list of causes), and the two body inaccuracies QA found — the understated mutation table and the "practically unreachable" framing — are corrected above. Both were reporting defects rather than code defects; no behaviour changed.

The second is test-only, closing the one coverage gap paired QA and the review bot independently flagged: commit 1 was the only functional commit without a test of its own (test 9), and the drop-during-write arm now asserts adapter disposal too, since `clearConnectionAfterDrop` disconnects on both of its callers and only the write-failure arm was checking it.

## Device QA / maintainer gates

This machine is Linux — no macOS, no iOS simulator, no BLE hardware — so the physical triggers can't be produced here. None of these block CI, but they are the honest gaps:

- [ ] **Dead-link-on-first-write on real hardware.** Needs a board plus walking out of range during the connect handshake, powering the box off mid-connect, or a second phone stealing a last-connection-wins board in the ~200ms window around the first write. Verify: the bulb stays dark, the `bluetooth.connectFailed` copy reads sensibly for this cause, and a second lightbulb tap recovers cleanly.
- [ ] **Success-path regression on all three board families** (Kilter, Tension, MoonBoard): connect-and-light from the play drawer and from the create-climb BLE toggle must still light on the first tap with exactly **one** success haptic. That is the dedup seed doing its job — test 7 guards it in CI, but the haptic is only observable on device.
- [ ] **iOS pass if a device is available.** The adapter `onDisconnect` timing (test 2) is CoreBluetooth-supervision-timeout dependent and can't be reproduced on Android or in vitest.
- [ ] **Maintainer call on the double `ble.errorIncompatible` alert** (see Decisions above) — the one behaviour change a climber could notice.
- [ ] **Board switch during the initial write** (see commit 1 above): tap the lightbulb on one board, then navigate to a different board/layout/size before the first write settles. Expect the old link to be torn down (correct) plus a "move closer" alert that doesn't match the cause. Confirm it's rare enough in real use to leave as-is; if not, the fix is a distinct bail reason rather than a change to the teardown.

JS-only, so a `pr-<number>` OTA preview channel is enough for the on-device pass; no new binary needed.

## Release Notes

- Your board no longer shows as connected when the connection actually died on the first send — you get told to move closer and try again instead of tapping into a dark wall.
- If the first send after connecting doesn't land, the next climb you pick actually gets written to the wall, and your crew no longer sees it as lit when it isn't.

### DIFF
--- a/packages/mobile/src/lib/ble/__tests__/use-board-bluetooth.test.ts
+++ b/packages/mobile/src/lib/ble/__tests__/use-board-bluetooth.test.ts
@@ -1787,6 +1787,11 @@ describe('useBoardBluetooth connect() initial frame write (#3875)', () => {
     expect(Alert.alert).toHaveBeenCalledWith('ble.connectionFailedTitle', 'bluetooth.connectFailed');
     const failure = mockTrack.mock.calls.find(([name]) => name === 'Bluetooth Connection Failed');
     expect(failure?.[1]).toMatchObject({ failureReason: 'dropped_after_connect' });
+    // clearConnectionAfterDrop disposes the adapter on BOTH of its callers, not
+    // just the write-failure one — the adapter self-cleans after firing its own
+    // event, but the explicit disconnect is what stops a half-alive native link
+    // and the onDeviceDisconnected subscription leaking.
+    expect(fakeAdapter.disconnect).toHaveBeenCalled();
   });
 
   it('still connects when the initial climb is incompatible with the board (#3875)', async () => {
@@ -1928,6 +1933,76 @@ describe('useBoardBluetooth connect() initial frame write (#3875)', () => {
     await act(async () => {
       await result.current.connect();
     });
+    expect(result.current.connectInitialSendRef.current).toBeNull();
+  });
+
+  it('fails the connect when the board config switches during the initial write (#3875)', async () => {
+    // The guard for commit 1 (connectedConfigKeyRef assigned when the link opens
+    // rather than after the initial send). That move is what makes this path
+    // reachable at all: the config-switch effect early-returns while the ref is
+    // null, so before it, a board switch landing inside the awaited write left
+    // the stale connection up. Now the effect tears it down and the identity
+    // bail below stops connect() lighting the lightbulb over a dead adapter.
+    //
+    // The alert asserted here is deliberately the generic "move closer" copy,
+    // which does NOT describe what happened — the climber navigated away, the
+    // link did not drop. That wart is accepted (see the comment beside the bail);
+    // this assertion pins it so a future distinct bail reason updates the test on
+    // purpose instead of silently changing what the climber is told.
+    let releaseWrite: (() => void) | undefined;
+    let writeReached: (() => void) | undefined;
+    const writeGate = new Promise<void>((resolve) => {
+      releaseWrite = resolve;
+    });
+    const writeStarted = new Promise<void>((resolve) => {
+      writeReached = resolve;
+    });
+    const fakeAdapter = makeFakeAdapter({
+      write: vi.fn(async () => {
+        writeReached?.();
+        await writeGate;
+      }),
+    });
+    vi.mocked(createBluetoothAdapter).mockReturnValue(
+      fakeAdapter as unknown as ReturnType<typeof createBluetoothAdapter>,
+    );
+    mockAuroraSendable();
+
+    const { result, rerender } = renderHook((props) => useBoardBluetooth(props), {
+      initialProps: { boardName: 'kilter', layoutId: 1, sizeId: 1 },
+    });
+
+    // Park connect() on the initial write. Only then is adapterRef (and, thanks
+    // to commit 1, connectedConfigKeyRef) set, which is what stops the
+    // config-switch effect early-returning and makes this test non-vacuous.
+    let pending: Promise<boolean> | undefined;
+    await act(async () => {
+      pending = result.current.connect('p100r12');
+      await writeStarted;
+    });
+
+    // The switch needs its own act: effects queued by rerender only flush when
+    // the act scope exits, so doing this inside the block above would release the
+    // write before the teardown had run.
+    await act(async () => {
+      rerender({ boardName: 'kilter', layoutId: 2, sizeId: 1 });
+    });
+    expect(fakeAdapter.disconnect).toHaveBeenCalled();
+
+    let connected: boolean | undefined;
+    await act(async () => {
+      releaseWrite?.();
+      connected = await pending;
+    });
+
+    expect(connected).toBe(false);
+    expect(result.current.isConnected).toBe(false);
+    expect(mockTrack.mock.calls.find(([name]) => name === 'Bluetooth Connection Success')).toBeUndefined();
+    expect(Alert.alert).toHaveBeenCalledWith('ble.connectionFailedTitle', 'bluetooth.connectFailed');
+    const failure = mockTrack.mock.calls.find(([name]) => name === 'Bluetooth Connection Failed');
+    expect(failure?.[1]).toMatchObject({ failureReason: 'dropped_after_connect' });
+    // The switched-away connection must not leave a seed behind for the new
+    // board's AutoSender to inherit.
     expect(result.current.connectInitialSendRef.current).toBeNull();
   });
 });
--- a/packages/mobile/src/lib/ble/__tests__/use-board-bluetooth.test.ts
+++ b/packages/mobile/src/lib/ble/__tests__/use-board-bluetooth.test.ts
@@ -1848,6 +1848,73 @@ describe('useBoardBluetooth connect() initial frame write (#3875)', () => {
     const sendFailure = mockTrack.mock.calls.find(([name]) => name === 'Climb Sent to Board Failure');
     expect(sendFailure?.[1]).toMatchObject({ sendSource: 'connect' });
   });
+
+  it('leaves the AutoSender dedup unseeded when the initial write failed (#3875)', async () => {
+    // The seed tells the AutoSender "these frames are already on the wall".
+    // Seeding it after a write that never landed made the AutoSender skip the
+    // write (connected board, dark wall) and fire onWallConfirmed anyway, so the
+    // party feed reported a climb as lit on a dark wall. The link is alive here,
+    // so the connect itself must still succeed.
+    const fakeAdapter = makeFakeAdapter({
+      write: vi.fn().mockRejectedValue(new Error('GATT write failed')),
+    });
+    vi.mocked(createBluetoothAdapter).mockReturnValue(
+      fakeAdapter as unknown as ReturnType<typeof createBluetoothAdapter>,
+    );
+    mockAuroraSendable();
+
+    const { result } = renderHook(() => useBoardBluetooth({ boardName: 'kilter', layoutId: 1, sizeId: 1 }));
+
+    await act(async () => {
+      await result.current.connect('p100r12');
+    });
+
+    expect(result.current.isConnected).toBe(true);
+    expect(result.current.connectInitialSendRef.current).toBeNull();
+  });
+
+  it('seeds the AutoSender dedup when the initial write landed (#3875)', async () => {
+    // The mutation guard for the test above: clearing the seed unconditionally
+    // would pass it while reintroducing the doubled connect haptic and the
+    // redundant full-frame re-write the seed exists to prevent.
+    const fakeAdapter = makeFakeAdapter();
+    vi.mocked(createBluetoothAdapter).mockReturnValue(
+      fakeAdapter as unknown as ReturnType<typeof createBluetoothAdapter>,
+    );
+    mockAuroraSendable();
+
+    const { result } = renderHook(() => useBoardBluetooth({ boardName: 'kilter', layoutId: 1, sizeId: 1 }));
+
+    await act(async () => {
+      await result.current.connect('p100r12');
+    });
+
+    expect(result.current.connectInitialSendRef.current).toEqual({
+      frames: 'p100r12',
+      mirrored: false,
+      colorSignature: expect.any(String),
+    });
+  });
+
+  it('clears a stale dedup seed when connecting without initial frames (#3875)', async () => {
+    const fakeAdapter = makeFakeAdapter();
+    vi.mocked(createBluetoothAdapter).mockReturnValue(
+      fakeAdapter as unknown as ReturnType<typeof createBluetoothAdapter>,
+    );
+    mockAuroraSendable();
+
+    const { result } = renderHook(() => useBoardBluetooth({ boardName: 'kilter', layoutId: 1, sizeId: 1 }));
+
+    await act(async () => {
+      await result.current.connect('p100r12');
+    });
+    expect(result.current.connectInitialSendRef.current).not.toBeNull();
+
+    await act(async () => {
+      await result.current.connect();
+    });
+    expect(result.current.connectInitialSendRef.current).toBeNull();
+  });
 });
 
 describe('convertToMirroredFramesString', () => {
--- a/packages/mobile/src/lib/ble/__tests__/use-board-bluetooth.test.ts
+++ b/packages/mobile/src/lib/ble/__tests__/use-board-bluetooth.test.ts
@@ -1781,6 +1781,12 @@ describe('useBoardBluetooth connect() initial frame write (#3875)', () => {
     expect(connected).toBe(false);
     expect(result.current.isConnected).toBe(false);
     expect(mockTrack.mock.calls.find(([name]) => name === 'Bluetooth Connection Success')).toBeUndefined();
+    // Asserted here too, not just in the test above: this is the arm the boolean
+    // return value cannot reach, so it needs its own proof that the climber is
+    // told and the failure is recorded.
+    expect(Alert.alert).toHaveBeenCalledWith('ble.connectionFailedTitle', 'bluetooth.connectFailed');
+    const failure = mockTrack.mock.calls.find(([name]) => name === 'Bluetooth Connection Failed');
+    expect(failure?.[1]).toMatchObject({ failureReason: 'dropped_after_connect' });
   });
 
   it('still connects when the initial climb is incompatible with the board (#3875)', async () => {
--- a/packages/mobile/src/lib/ble/__tests__/use-board-bluetooth.test.ts
+++ b/packages/mobile/src/lib/ble/__tests__/use-board-bluetooth.test.ts
@@ -1689,6 +1689,15 @@ describe('useBoardBluetooth remembered-board persistence (#3609)', () => {
 // connect() writes initialFrames before declaring success. That write can take
 // the link with it. These tests pin the boundary: a lost link fails the connect,
 // while a write the board merely refused does not.
+//
+// Which side of that boundary a test lands on is decided by the rejection
+// message, via isDisconnectionError (@boardsesh/ble-protocol/connection-error):
+// 'Not connected' matches its "definite drop" pattern and tears the connection
+// down; 'GATT write failed' does not and leaves the link alive. Deliberately not
+// re-derived in the test bodies — asserting against a locally rebuilt predicate
+// is the tautology this repo has already shipped once. If a change to
+// isDisconnectionError ever collapsed those two messages to the same verdict,
+// the pair below would start agreeing and this comment is the breadcrumb.
 describe('useBoardBluetooth connect() initial frame write (#3875)', () => {
   beforeEach(() => {
     vi.clearAllMocks();
--- a/packages/mobile/src/lib/ble/__tests__/use-board-bluetooth.test.ts
+++ b/packages/mobile/src/lib/ble/__tests__/use-board-bluetooth.test.ts
@@ -1686,6 +1686,170 @@ describe('useBoardBluetooth remembered-board persistence (#3609)', () => {
   });
 });
 
+// connect() writes initialFrames before declaring success. That write can take
+// the link with it. These tests pin the boundary: a lost link fails the connect,
+// while a write the board merely refused does not.
+describe('useBoardBluetooth connect() initial frame write (#3875)', () => {
+  beforeEach(() => {
+    vi.clearAllMocks();
+    resetReactNativePermissionHarness();
+    mockBleManager.state.mockResolvedValue('PoweredOn');
+  });
+
+  function mockAuroraSendable() {
+    mockGetLedPlacements.mockReturnValue({ 100: 7, 999: 8 });
+    mockGetAuroraBluetoothPacket.mockReturnValue({
+      packet: new Uint8Array([0x01]),
+      skippedPositionCount: 0,
+      skippedRoleCount: 0,
+      totalPlacements: 1,
+    });
+  }
+
+  it('fails the connect when the initial write kills the link (#3875)', async () => {
+    // The write rejects with a disconnection signature, so sendFramesToBoard tears
+    // the connection down. connect() used to sail past that and light the
+    // lightbulb over a null adapter: every later send then early-returned with
+    // skipReason 'no_adapter' and the wall stayed dark until the climber toggled
+    // the connection off and back on.
+    const fakeAdapter = makeFakeAdapter({
+      write: vi.fn().mockRejectedValue(new Error('Not connected')),
+    });
+    vi.mocked(createBluetoothAdapter).mockReturnValue(
+      fakeAdapter as unknown as ReturnType<typeof createBluetoothAdapter>,
+    );
+    mockAuroraSendable();
+    const onConnectSuccess = vi.fn();
+
+    const { result } = renderHook(() =>
+      useBoardBluetooth({ boardName: 'kilter', layoutId: 1, sizeId: 1, onConnectSuccess }),
+    );
+
+    let connected: boolean | undefined;
+    await act(async () => {
+      connected = await result.current.connect('p100r12');
+    });
+
+    expect(connected).toBe(false);
+    expect(result.current.isConnected).toBe(false);
+    expect(onConnectSuccess).not.toHaveBeenCalled();
+    expect(mockTrack.mock.calls.find(([name]) => name === 'Bluetooth Connection Success')).toBeUndefined();
+    const failure = mockTrack.mock.calls.find(([name]) => name === 'Bluetooth Connection Failed');
+    expect(failure?.[1]).toMatchObject({ boardName: 'kilter', failureReason: 'dropped_after_connect' });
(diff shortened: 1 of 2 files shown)


marcodejongh:
Thanks — three of the four are fair; one is wrong.

**1. `failureReason: 'dropped_after_connect'` is a "silent type hole" — no, there is no type to widen.** Checked the actual signature rather than assuming: `track` is destructured from `createAnalytics` (`packages/mobile/src/lib/analytics.ts:124`), and its properties parameter is `AnalyticsEventProperties = Record<string, AnalyticsPropertyValue | undefined>` (`packages/shared/analytics/src/create-analytics.ts:6,11`). `SHARED_EVENTS` is a catalog of event *names* only; event props are not typed per event anywhere. The pre-existing call two lines below mine already passes a union of `classifyBleFailure` categories and `classifyBleFailureReason` values through the same untyped property. So there is nothing in `events.ts` for either PR to keep in sync, and no merge-order hazard — which is exactly why the PR skips that file (it would be a doc-only edit landing inside #3980's hunk region). The real cost is the one the PR body already names: a saved PostHog insight that enumerates `failureReason` values needs the new one added.

**2 (double `ble.errorIncompatible`) — agreed**, and it is already an explicit maintainer gate in the PR checklist. Leaving it for the maintainer.

**3 (`onConnectionChange(false)` contract) — fixed** in `ffefeebe6`: the contract now lives on the `UseBoardBluetoothOptions` member itself, not only at the call site.

**4 (`'Not connected'` vs `'GATT write failed'` discriminator invisible in the tests) — fixed** in the same commit, as a comment on the describe naming `isDisconnectionError` and pointing at the two fixtures. Deliberately *not* re-derived in the test bodies: asserting against a locally rebuilt copy of the predicate is a tautology, and this repo has shipped that mistake before.

marcodejongh:
Both actionable notes are fair and are now fixed; the third is a maintainer gate, not a code change.

**1. Config-switch bail path has no test — fixed (`dbce24ced`).** You and the paired QA reviewer flagged this independently, and you were both right: commit 1 was the only functional commit without a test of its own. Test 9 now parks `connect()` on the initial write (the fake adapter's `write` blocks on a gate), switches the board config in its own `act` so the effect actually flushes, then releases the write and asserts `connect()` returns `false`, no success event fires, and no dedup seed is left behind for the new board to inherit.

The part worth checking is that it isolates commit 1 rather than just re-testing the config-switch effect. Moving the `connectedConfigKeyRef` assignment back to after the initial send — the exact pre-commit-1 shape — fails **test 9 and nothing else**; every pre-existing config-switch teardown test still passes. That row is now in the mutation table in the PR body.

Adding test 9 also changed two other rows, so the whole table was re-derived: identity-bail-removed now reds 1, 2, 9 (3 failed) and the `initialSendResult === false` swap now reds 2, 3, 4, 6, 9 (5 failed).

**2. Test 2 missing `disconnect` assertion — fixed, and your first reading was the correct one.** `clearConnectionAfterDrop` calls `void adapter?.disconnect()` unconditionally (`use-board-bluetooth.ts:531`) for both of its callers. The adapter has self-cleaned after firing its own event, but the explicit disconnect is what stops a half-alive native link and the `onDeviceDisconnected` subscription leaking — so the assertion belongs on that arm too, not a comment explaining its absence. Added, with that reasoning inline.

**3. Double `ble.errorIncompatible` alert.** Agreed it is the one observable UX change, which is why it is a checkbox in the Device QA section. It stays a maintainer gate rather than a merge blocker: this box is Linux, so there is no BLE hardware, no iOS simulator, and no way to produce an incompatible-climb connect here. The behaviour was approved on the trade-off that one extra alert beats reporting a climb as lit to the party feed over a dark wall, and it is documented in the code beside the seed gate that causes it. The PR is JS-only, so the `pr-3998` OTA channel is enough for that pass on a real device.

One correction to the PR body while I was in here, from paired QA: it previously called the config-switch route into the bail "practically unreachable". That was wrong — the effect's deps are route-derived provider state, so ordinary navigation reaches it. The body now says so plainly, and it is a device-QA checkbox.

marcodejongh:
On the regenerated review's remaining point — `failureReason: 'dropped_after_connect'` "sits alongside `classifyBleFailure` typed categories as a bare string literal with no union entry":

This is the same claim you raised earlier in this thread and then explicitly retracted ("Issue 1 (failureReason type hole) — was wrong ... No union to widen ... Retracted."). Re-verified from the source a third time rather than relying on either of the previous rounds:

- `packages/shared/analytics/src/create-analytics.ts:6` — `export type AnalyticsEventProperties = Record<string, AnalyticsPropertyValue | undefined>;`
- `packages/shared/analytics/src/create-analytics.ts:11` — `track(name: string, properties?: AnalyticsEventProperties): void;`
- `grep -n failureReason packages/shared/analytics/src/events.ts` → **no matches.** The identifier does not exist in that file.

`SHARED_EVENTS` is a catalog of event *names*; no event's properties are typed anywhere in the codebase. So there is no union entry to add, nothing for `events.ts` to declare, and no merge-order hazard with #3980 on this point. A typo in a PostHog filter is possible, but that is true of all ~94 `track` call sites and is not something this PR introduces or can fix.

Leaving the code as-is. The real operational consequence — that any saved insight enumerating `failureReason` needs the new value added by hand — is already called out in the PR body and in the handoff notes.

packages/mobile/src/lib/ble/use-board-bluetooth.ts
```diff
@@ -974,18 +988,86 @@ export function useBoardBluetooth({
...
+        // this to `if (initialSendResult === false)`:
+        //   - false does NOT mean the link died. It is also returned for
+        //     incompatible_climb, missing_mirror_data, missing_led_placements and
+        //     every ordinary write rejection (write_failed, write_timeout,
+        //     characteristic_unavailable). Failing the connect on those would
+        //     strand a live adapter behind a dark bulb — the mirror-image bug.
+        //   - false is NOT returned when the adapter's onDisconnect fires during
+        //     the write: write() resolves, the send reports success, and the
+        //     teardown has already happened. The boolean misses that case entirely.
+        if (adapterRef.current !== adapter) {
```
marcodejongh, line 1031: The identity check reads inequality as "our adapter was torn down", but there is a third state it can't distinguish: `adapterRef.current` repopulated with a *different, live* adapter. Today that state is unreachable only because both routes into `adapterRef` are gated on this very connect — `adopt()` early-returns on `connectInFlightRef.current` (line 1295), and a concurrent `connect()` is blocked by the same ref at the top of this function. If either guard is ever relaxed (e.g. adoption allowed to win over a stalled connect), this branch would raise "Couldn't connect" over a connection that is in fact up, and return `false` while `isConnected` stays `true` — the bail deliberately doesn't tear down, so it would leave the mixed state in place.

Not a defect now, but the comment block above lists the routes into the bail without naming `connectInFlightRef` as the invariant that makes "not my adapter" equivalent to "no adapter". One line naming it would keep a future adoption change from silently breaking this.

---

packages/mobile/src/lib/ble/use-board-bluetooth.ts
```diff
@@ -974,18 +988,86 @@ export function useBoardBluetooth({
...
+        // Send initial frames if provided. `sendSource: 'connect'` makes the
+        // resulting Climb Sent to Board Success/Failure/Skipped attributable to
+        // this connect-time write instead of having to be inferred from a
+        // millisecond gap against Bluetooth Connection Success.
+        let initialSendResult: boolean | undefined;
         if (initialFrames) {
-          await sendFramesToBoard(initialFrames, mirrored);
-          connectInitialSendRef.current = { frames: initialFrames, mirrored: !!mirrored, colorSignature };
-        } else {
+          initialSendResult = await sendFramesToBoard(initialFrames, mirrored, undefined, { sendSource: 'connect' });
```
marcodejongh, line 997: Analytics footnote for whoever reads the new `sendSource: 'connect'` funnel: on the drop-*during*-write arm (test 2), the write resolves before the teardown is observed, so `performSend` still tracks `Climb Sent to Board Success` (line 746) — and the same connect then emits `Bluetooth Connection Failed` / `dropped_after_connect`. So a `sendSource: 'connect'` **Success** does not imply the connect succeeded; the pairing to check is the send event *plus* the connection outcome. Worth a line in the PR body's analytics notes (or the handoff to whoever maintains the saved insights), since the PR's own oracle-building shows these joins are easy to get subtly wrong.

Related, one path quirk rather than a bug: with commit 1, a config switch landing during the *pre-send* awaits (e.g. `configureBoard`) tears down before the send starts, so the initial send now logs `Climb Sent to Board Skipped` with `skipReason: 'no_adapter'`, `sendSource: 'connect'` — a new shape for the issue's previously-silent `Skipped` telemetry. Harmless, but it means that event firing is no longer strictly "connected-but-not-ready mid-session" as its comment in `sendFramesToBoard` describes.

---

packages/mobile/src/lib/ble/__tests__/use-board-bluetooth.test.ts
```diff
@@ -1686,6 +1686,327 @@ describe('useBoardBluetooth remembered-board persistence (#3609)', () => {
...
+    vi.clearAllMocks();
+    resetReactNativePermissionHarness();
+    mockBleManager.state.mockResolvedValue('PoweredOn');
+  });
+
+  function mockAuroraSendable() {
+    mockGetLedPlacements.mockReturnValue({ 100: 7, 999: 8 });
+    mockGetAuroraBluetoothPacket.mockReturnValue({
+      packet: new Uint8Array([0x01]),
+      skippedPositionCount: 0,
```
marcodejongh, line 1712: `mockAuroraSendable()` and the reliance on `vi.clearAllMocks()` in this suite's `beforeEach` deserve one check: `mockGetLedPlacements` / `mockGetAuroraBluetoothPacket` are module-level mocks shared with the earlier describes, and `clearAllMocks` resets *implementations* set via `mockReturnValue` too — which is why every test here re-arms them. That works, but it also means the suite silently depends on being the only describe whose `beforeEach` runs `clearAllMocks` before its own arming. If a future test in this block forgets to call `mockAuroraSendable()`, `getAuroraBluetoothPacket` returns `undefined` and the failure will surface as an opaque TypeError inside `sendFramesToBoard` rather than an assertion. Moving the arming into the `beforeEach` (tests that need different packet shapes override after) would make the fixtures fail loudly and drop the per-test boilerplate. Non-blocking.

---

marcodejongh (commented):
## Review — fix(mobile): fail the BLE connect when its first write kills the link (#3875)

### Overview

The PR replaces "connect() declares success after the initial write regardless of what that write did to the link" with an adapter-identity bail, gates the AutoSender dedup seed on the write actually landing, and moves `connectedConfigKeyRef` assignment up so the config-switch effect can tear down a connection whose initial write is still in flight. Two files, JS-only, nine new tests.

### Verdict

The fix is correct and I verified the load-bearing claims against the code rather than the (already extensively reviewed) PR body:

- **Identity over boolean is the right call.** `sendFramesToBoard` returns `false` for five live-link conditions (`incompatible_climb`, `missing_mirror_data`, `missing_led_placements`, plus the generic write-rejection catch) and returns `true`/`undefined` on the drop-during-write and abort paths — so `initialSendResult` genuinely cannot express "link died". Only `isDisconnectionError` (and the MoonBoard streak, which can't trip on a first write) routes into `handleDisconnection`, and both teardown routes null `adapterRef`. The check is sound.
- **The bail's cleanup accounting is right.** `clearConnectionAfterDrop` covers `isConnected`, the write generation, adapter disposal, Sentry tags, and `onConnectionChange(false)`; the bail correctly adds only the seed clear, alert, and analytics. The `finally` still resets `connectInFlightRef`/`loading` on the early return.
- **No double alert on any connect path.** The write-failure catch in `performSend` deliberately doesn't alert for disconnection-class failures, and the pre-write validation alerts (`errorIncompatible` etc.) all leave the link alive so the bail doesn't fire.
- **Caller audit checks out** — all six `connect()` call sites `void` the result.

### Inline comments (all non-blocking)

1. The identity check's soundness quietly depends on `connectInFlightRef` gating both native adoption and concurrent connects — worth naming that invariant in the comment block so a future adoption change can't break it silently.
2. Two analytics footnotes on the `sendSource: 'connect'` funnel: `Climb Sent to Board Success` can co-occur with `dropped_after_connect` (drop-during-write arm), and the config-switch path can now produce a `Skipped`/`no_adapter`/`sendSource: 'connect'` event.
3. Test-suite fixture arming (`mockAuroraSendable`) would fail more loudly from the `beforeEach`.

### Remaining open items (already known)

The maintainer gate on the double `ble.errorIncompatible` alert and the on-device QA passes (dead-link, three board families, iOS CoreBluetooth timing) are honestly flagged in the PR body and are the right things to hold ready-for-review on. Nothing else blocks.

---

PR #3999, opened 2026-07-30
fix(mobile): clear climb name on filter-sheet Reset, surface it as a row

### LINKED ISSUE #3606: Add a the climb name filter to filter sheet
image

The climb name filter is in a seperate ui than the filter sheet. I always confuse the hell out of myself when I still have a climb name in there, and trying to reset the filters, because the reset in the sheet doesnt reset the climbname filter.

image

That button doesnt reset the climbname filter but probably should. For it to then not be confusing, we should probably also show the climbname filter as the first option in the filter sheet

marcodejongh:
## Premise check

Confirmed as reported, no correction needed. The climb-name search term lives
in a separate reducer slice (`ClimbSearchState.name`, committed live via a
300ms debounce) from the sheet's Apply-gated filter draft
(`ClimbFilters`/`DEFAULT_FILTERS`), which has no `name` field. Reset only ever
called the two draft setters that exist inside the sheet — there was no prop to
also clear the parent's committed name, and none was added when Reset was
built. Root cause is an incomplete prop contract, not an inverted condition or
a logic bug: there was no wire to cut, because it was never run.

## What changed

Two coupled changes, per the issue's own ask:

1. **Reset now clears the climb name.** `ClimbFilterSheet` gets a new
   `onClearName: () => void` prop; `handleReset` calls it alongside the two
   existing draft resets. The parent (`app/(tabs)/climbs/index.tsx`) wires a new
   `handleClearName` that clears the debounce timer, the committed `name`, and
   mirrors the clear into whichever search bar is showing (custom capsule or
   the iOS-26 native one) — reusing the exact `applyVisibleSearchText` +
   `setName('')` mechanism `handleApplyRecentFilter` and the per-board restore
   paths already use. `handleNativeSearchCancel` is untouched — Reset must not
   close the sheet, and cancel's `setShowFilters(false)` is a different intent.
2. **Climb name is now the sheet's first row**, above "1 · Difficulty" (now "2 ·
   Difficulty" — the six section-comment numbers shifted). It's a plain
   `TextInput` with an inline × that only shows once there's text. Typing
   mirrors live into the top bar / native search bar via a new composed
   `handleSheetNameChange` (`applyVisibleSearchText` first, then
   `handleSearchChange` — order matters, see the code comment: the debounced
   `handleSearchChange` write must run second so the normalized value wins).
   The × and Reset share **one** clearing path
   (`onClearName`/`handleClearNameField`) — there's exactly one code path to
   reason about for "clear the name". The "Show N" apply-button preview now
   folds the in-sheet name draft into its existing 250ms-debounced
   `filters`/`boardFilters` state, so an in-sheet name edit updates the count
   the same way a chip/toggle edit already does.
3. **Reset is now tappable when a climb name is the only thing set.** Caught by
   the paired QA review, and it's the issue's literal repro. The header Reset
   button's enabled state (`anyActive`) was built purely from
   `hasActiveClimbFilters`/`hasActiveBoardFilters`, and neither can see the
   name — that's the whole point of the root cause above, name lives outside
   `ClimbFilters`. So with a lone climb-name search and every other control at
   its default, Reset rendered greyed out and `handleReset` never fired, making
   change (1) unreachable in exactly the state the reporter described.
   `anyActive` now also folds in `nameDraft.length > 0`, the same rule that
   already gates the field's inline ×, so the two clear affordances light up
   together.
4. **`onNameChange` / `onClearName` are required props**, not optional. Review
   note, and a fair one: the sheet only owns the field's *display* state, so an
   unwired caller would render a name field that looks like it filters and
   doesn't, and a × that blanks the field while the committed term survives.
   There's one consumer and it passes both, so requiring them costs nothing and
   the compiler now enforces it.

New i18n key `mobile.filter.section.name` in `en-US`/`es`/`fr` ("Climb name" /
"Nombre del bloque" / "Nom du bloc" — following this same file's existing
`theClimb` → "El bloque"/"Le bloc" precedent). Everything else reuses existing
catalog entries (placeholder = `search.placeholders.climbs`, clear-button label
= `common:clear`). Also added `keyboardShouldPersistTaps="handled"` to the
sheet's scroll view — it never hosted a text input before, so a first tap on
Apply/Reset while the keyboard is up could otherwise get swallowed.

## Why one clearing path

The inline × and Reset both call the same `onClearName`/`handleClearNameField`
functions rather than each having their own "clear the name" logic. This means
a future bug in clearing can only exist in one place, and the tests below only
need to prove that path works once and that both buttons reach it.

## Testing

`packages/mobile/src/components/__tests__/climb-filter-sheet.test.tsx`, seven
new tests:

- Reset calls `onClearName` exactly once; Apply does **not** (guards against a
  copy-paste inversion between the two buttons, which share a header/footer).
- With **only** a climb name set — the activity mocks deliberately left at
  their default `false`, so nothing else counts as active — Reset renders
  enabled and clears the name. This is the one the QA review's blocking finding
  was about; the earlier Reset test forced `hasActiveClimbFilters` to `true`,
  which happened to also be what un-disabled the button, so it masked the gap.
- With nothing at all set, Reset stays disabled and `onClearName` never fires
  (the flip side, so the gate isn't just always-on).
- Typing in the name field mirrors to `onNameChange` and updates the field's
  own displayed value.
- The inline × only renders once there's text, and clearing through it goes
  through the same `onClearName` path as Reset.
- The field resyncs from an external `searchName` prop change (board switch /
  recent pill), but a trim-only difference (mid-typing trailing space) does
  **not** yank the field — exercised through the real DOM value the field
  renders, not by re-asserting the shared `visibleSearchTextNeedsSync` helper's
  own return value against itself.
- The "Show N" preview reflects an in-sheet name edit after the 250ms debounce.

**Mutation checks performed** (twice): removed the `onClearName()` call from
`handleReset`, confirmed exactly the Reset regression test went red and no
other test was affected, then restored it. Same again for the `anyActive` gate
— dropping the `nameDraft.length > 0` clause reds exactly the name-only Reset
test (28/29), restoring it is back to 29/29.

Full validation, all green:

- `vp run typecheck:mobile`
- `vp run test:mobile` — 557 files / 4687 tests (4680 baseline + 7 new)
- `vp check` — 0 errors (286 pre-existing warnings elsewhere in the repo,
  unrelated to this change)
- `vp run check:i18n` / `vp run check:i18n:orphans` — both OK

## Device QA gates (not covered by this PR's automated validation)

- **iOS**: the two-way text mirroring (`applyVisibleSearchText` into
  `nativeSearchRef`/`searchHeaderRef`) and the iOS-26 native bottom-tab search
  bar variant (`useNativeAccessoryActive`) can't be exercised in jsdom/vitest
  (RN native components are mocked), and this box has no iOS simulator. Needs a
  real device/simulator pass: typing in the sheet should visibly update the top
  search bar / native search bar live, and Reset should visibly clear whichever
  is showing.
- **Keyboard/tap interaction**: `keyboardShouldPersistTaps="handled"` is
  included, but whether a tap on Apply/Reset while the keyboard is up actually
  registers is a native gesture behavior jsdom can't verify.
- **Android**: attempted an emulator pass on this box (Pixel AVD, dev-client +
  Metro against this branch). The emulator booted and the dev-client installed
  cleanly, but the app hadn't finished its first JS bundle launch inside this
  session's time budget, so the interactive part (open sheet, type a name, tap
  Reset, screenshot) wasn't completed. No crash/error was observed in the
  process I could inspect. Left as a gate for the next pass rather than
  claiming a verification that didn't finish. Worth walking the issue's exact
  repro on that pass: type only a climb name, touch nothing else, confirm Reset
  is coloured/tappable and clears both the sheet field and the top bar.

## Scope

Not touched, deliberately: `handleNativeSearchCancel` (different intent — see
above); Apply's handling of name (still commits live via debounce, independent
of Apply, same as today); how `name` persists to per-board last-search storage
(already board-scoped, untouched); a web equivalent (web's climb search is
URL-param-driven with no local-draft-plus-Apply sheet model and no existing
Reset/name coupling — no parity gap found there).

## Release Notes

The filter sheet's Reset button now actually resets everything — including any
climb name you'd typed in. And you can type or clear a climb name right there
in the sheet, front and center, so you don't have to hunt for the search bar
separately.

Closes #3606.

### DIFF
--- a/packages/mobile/app/(tabs)/climbs/index.tsx
+++ b/packages/mobile/app/(tabs)/climbs/index.tsx
@@ -333,6 +333,9 @@ function ClimbListInner() {
   // actual TextInput/native field, leaving it visually stale.
   const handleSheetNameChange = useCallback(
     (text: string) => {
+      // ORDER MATTERS — see the comment above: mirror the raw text first, THEN
+      // let handleSearchChange's normalized write win by running second.
+      // Swapping these two lines silently regresses the top-bar mirror.
       applyVisibleSearchText(text);
       handleSearchChange(text);
     },
--- a/packages/mobile/app/(tabs)/climbs/index.tsx
+++ b/packages/mobile/app/(tabs)/climbs/index.tsx
@@ -323,6 +323,32 @@ function ClimbListInner() {
     nativeSearchRef.current?.clearText();
   }, [setName]);
 
+  // Composed for the filter sheet's in-line name field (#3606): mirror the raw
+  // keystroke into the top bar/native search bar FIRST via applyVisibleSearchText
+  // (silent — doesn't re-enter this same handler), THEN run handleSearchChange's
+  // normal normalize+debounce+commit path, whose ref/state writes run second and
+  // so win. Without the mirror call first, the top bar's own resync effect (see
+  // the visibleSearchTextNeedsSync effect above) would see visibleSearchTextRef
+  // already in lockstep with the sheet's typing and skip re-seeding the top bar's
+  // actual TextInput/native field, leaving it visually stale.
+  const handleSheetNameChange = useCallback(
+    (text: string) => {
+      applyVisibleSearchText(text);
+      handleSearchChange(text);
+    },
+    [applyVisibleSearchText, handleSearchChange],
+  );
+
+  // The ONE clearing path for name, shared by the filter sheet's Reset button and
+  // its inline × (#3606) — mirrors handleNativeSearchCancel's clearing sequence
+  // but deliberately does NOT touch setShowFilters: Reset must not close the sheet.
+  const handleClearName = useCallback(() => {
+    if (debounceTimerRef.current) clearTimeout(debounceTimerRef.current);
+    visibleSearchTextRef.current = '';
+    setName('');
+    applyVisibleSearchText('');
+  }, [applyVisibleSearchText, setName]);
+
   const handleSearchFocus = useCallback(() => {
     setShowGrade(false);
     setIsSearchFocused(true);
@@ -1565,6 +1591,8 @@ function ClimbListInner() {
           searchName={name}
           lastUsedGradeId={lastUsedGrade}
           onApply={handleApplyFilters}
+          onNameChange={handleSheetNameChange}
+          onClearName={handleClearName}
         />
       ) : null}
     </View>
--- a/packages/mobile/app/(tabs)/climbs/index.tsx
+++ b/packages/mobile/app/(tabs)/climbs/index.tsx
@@ -343,7 +343,10 @@ function ClimbListInner() {
   // its inline × (#3606) — mirrors handleNativeSearchCancel's clearing sequence
   // but deliberately does NOT touch setShowFilters: Reset must not close the sheet.
   const handleClearName = useCallback(() => {
-    if (debounceTimerRef.current) clearTimeout(debounceTimerRef.current);
+    if (debounceTimerRef.current) {
+      clearTimeout(debounceTimerRef.current);
+      debounceTimerRef.current = null;
+    }
     visibleSearchTextRef.current = '';
     setName('');
     applyVisibleSearchText('');
--- a/packages/mobile/src/components/ClimbFilterSheet.tsx
+++ b/packages/mobile/src/components/ClimbFilterSheet.tsx
@@ -3,6 +3,7 @@ import {
   View,
   Pressable,
   StyleSheet,
+  TextInput,
   type ViewStyle,
   type NativeScrollEvent,
   type NativeSyntheticEvent,
@@ -56,6 +57,7 @@ import { parseSetIdsParam, prewarmCreateBoardHolds } from '../lib/create-board-h
 import { subscribeToHoldsFilterSelection } from '../lib/hold-filter-handoff';
 import { subscribeToZoneFilterSelection, type ZoneFilterSelection } from '../lib/zone-filter-handoff';
 import { subscribeToSetterFilterSelection } from '../lib/setter-filter-handoff';
+import { visibleSearchTextNeedsSync } from '../lib/search-name';
 import { useAuth } from '../providers/auth-provider';
 import { hapticSelection } from '../lib/haptics';
 import { springs } from '../theme/animations';
@@ -77,11 +79,19 @@ type ClimbFilterSheetProps = {
   boardConfig: BoardSearchConfig | null;
   currentFilters: ClimbFilters;
   currentBoardFilters: ClimbBoardFilterState;
-  /** Current committed name term, so the live "Show N" count matches Apply. */
+  /** Current committed name term, so the live "Show N" count matches Apply and
+   *  the in-sheet name field seeds correctly. */
   searchName?: string;
   /** Last-used grade id; centres the grade rail on a familiar grade when unset. */
   lastUsedGradeId?: number;
   onApply: (filters: ClimbFilters, boardFilters: ClimbBoardFilterState) => void;
+  /** Fired per keystroke as the in-sheet name field changes (mirrors into the
+   *  parent's live search — name is committed outside the Apply-gated draft
+   *  model, same as the top-bar search field). */
+  onNameChange?: (text: string) => void;
+  /** Fired by Reset and the field's inline × — the ONE clearing path for name,
+   *  shared so there is exactly one code path to reason about (#3606). */
+  onClearName?: () => void;
 };
 
 // The status enum is still driven from the sheet — "My drafts" (Your progress
@@ -155,8 +165,11 @@ export function ClimbFilterSheet({
   searchName,
   lastUsedGradeId,
   onApply,
+  onNameChange,
+  onClearName,
 }: ClimbFilterSheetProps) {
   const { t } = useTranslation('climbs');
+  const { t: tCommon } = useTranslation('common');
   const theme = useTheme();
   const { systemColors } = theme;
   const { isAuthenticated } = useAuth();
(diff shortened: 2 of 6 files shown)


marcodejongh:
Addressed the review:

1. **`handleReset` duplication** — fixed. `handleReset` now calls `handleClearNameField()` directly instead of repeating its two lines, so Reset and the inline × are two callers of one function, not two copies. Mutation-tested: removing that call turned the Reset regression test red again; restored and green.
2. **`debounceTimerRef.current` not nulled** — done, in `handleClearName` (index.tsx). Left `handleNativeSearchCancel` alone per the plan's explicit scope (it's pre-existing code with the same pattern, untouched by this PR on purpose — Reset must not go through cancel's `setShowFilters(false)` path).
3. **Device QA gates** — already called out in the PR body and staying open. For transparency: I did attempt an Android emulator pass during this session. The emulator/dev-client/Metro stack did reach "Ready" for interactive use, but only after I'd already torn it down for a session time-budget reason, so no interactive screenshots were captured. Leaving Android + iOS both as device-QA gates rather than claim a verification that didn't finish.

New commit: 3c69231.

marcodejongh:
Re the second pass:

1. **Inline order-matters comment** — added at the actual call site in `handleSheetNameChange` (commit d3cae42), not just in the block comment above.
2. **`borderRadius: 10` / `fontSize: 17` "magic numbers"** — intentional, not new: `nameInputRow`'s `borderRadius: 10` matches this same file's pre-existing `tappableRow` style (setters/holds/zone rows, also hardcoded 10, predates this PR), and `nameInput`'s `fontSize: 17` matches the existing `SearchHeader.tsx` search-field input. There's no token in `theme/tokens.ts` for either value (closest `borderRadius` steps are `md: 8` / `lg: 12`, and there's no font-size token scale at all — `AppMenu.web.tsx` also hardcodes 17). Since the new row's own doc comment says it's deliberately matching the tappable-row family, tokenizing just this one row would make it inconsistent with the sibling rows it's matching, not more consistent. Leaving as-is — flag it if you disagree.
3. **Device QA as a merge gate vs. follow-up** — leaving this call to the maintainer/QA pass rather than deciding it myself. It's already explicit in the PR body and not hidden.

New commit: d3cae42.

marcodejongh:
Paired QA review found a blocking defect in this PR, plus two review-bot notes. Both rounds are now in.

### Blocking: Reset was still unpressable in the issue's literal repro

The wiring landed, but the button's *enabled* state didn't. `anyActive` (the
gate on the header Reset `Pressable`) was built purely from
`hasActiveClimbFilters(localFilters) || hasActiveBoardFilters(localBoardFilters)`
— and neither can see the climb name, which is the whole point of this PR's root
cause: name lives outside `ClimbFilters`. So with a lone climb-name search and
every other control still at its default — "I still have a climb name in there,
and trying to reset the filters" from the issue — Reset rendered grey and
disabled, `handleReset` never fired, and the new `handleClearNameField()` inside
it was unreachable.

Worse, my own regression test masked it: it force-set `hasActiveClimbFilters` to
`true` "so Reset is enabled", which is exactly the thing that made the button
tappable, so it only ever proved the case where something *else* was already
active.

Fix (commit `0e31921`): `anyActive` now also folds in `nameDraft.length > 0` —
the same rule that already gates the field's inline ×, so the two clear
affordances light up together. Two new tests that deliberately leave the
activity mocks at their default `false`:

- only a name set → Reset is enabled and clears the name
- nothing at all set → Reset stays disabled and `onClearName` never fires

Mutation-checked: dropping the `nameDraft.length > 0` clause reds exactly the
first of those two (28/29) and nothing else; restoring it is back to 29/29.

### Review-bot notes

1. **Optional `onNameChange`/`onClearName` hide a footgun** — agreed, fixed in
   `f648c2b`. Both are required props now. The sheet only owns the field's
   display state, so an unwired caller would render a name field that looks like
   it filters and doesn't. One consumer, already passing both, so the compiler
   just states the existing contract.
2. **Fragile ordering in `handleSheetNameChange`** — leaving as-is. The two
   calls are the entire body of one four-line function with the reason inline at
   the call site; extracting them into a utility moves the ordering rather than
   removing it, and adds a hop between the reader and the two functions whose
   interaction is the point. If a refactor reorders them the top bar goes stale
   in a way device QA catches immediately.
3. **No unit test for `handleSheetNameChange` in `index.tsx`** — correct, and
   out of scope: there is no test harness for that screen anywhere in the repo
   (large hook-heavy component), and the new parent-side code is thin
   composition of two already-shipped functions. It's covered at the prop
   boundary by the sheet tests and flagged as a device-QA gate in the body.

Device QA gates (iOS mirroring, keyboard/tap, and a completed Android
interactive pass) remain open and are spelled out in the PR body. The Android
pass is worth walking the issue's exact repro on: type only a climb name, touch
nothing else, confirm Reset is coloured and tappable.

Validation re-run after both commits: `vp run typecheck:mobile` pass,
`vp run test:mobile` 557 files / 4687 tests pass, `vp check` 0 errors. CI green
on `0e31921`.

marcodejongh:
Re the latest bot pass — the flagged iOS regression doesn't exist, and the
vacuous-test claim is backwards. Details, since both are checkable:

**1. "iOS native search bar not cleared on Reset" — not a regression.**
`clearText()` and `setText('')` are the same call in `react-native-screens`.
From the installed 4.26.0 source:

```objc
// ios/RNSSearchBar.mm
- (void)clearText { [_controller.searchBar setText:@""]; }
- (void)setText:(NSString *)text { [_controller.searchBar setText:text]; }
```

```kotlin
// android/.../CustomSearchView.kt
fun clearText() = setQuery("", false)
fun setText(text: String) = setQuery(text, false)
```

`applyVisibleSearchText('')` already does `nativeSearch?.setText('')`, so the
native bar is cleared on Reset on both platforms. `handleNativeSearchCancel`
calls `clearText()` because that's the idiom it was written with, not because it
reaches something `setText('')` doesn't. The bot's premise ("if
applyVisibleSearchText does not internally call clearText") is answered by
reading it: it does the equivalent, one line down from the custom field's
`setText`. Still on the device-QA list as a *visual* confirmation, but there's
no code gap to close.

**2. "Fragile positional arg test … silently evaluates to undefined and the
assertion passes vacuously"** — inverted. The assertion is
`expect(lastCall?.[3]).toEqual({ name: 'crimp' })`; if the arg moved,
`lastCall?.[3]` is `undefined` and `toEqual({name:'crimp'})` **fails**. A
signature reorder reds this test, which is the behaviour you'd want. Positional
indexing is admittedly not pretty, but it fails safe.

**3. Hardcoded `borderRadius: 10` / `fontSize: 17`** — same answer as the
earlier pass, unchanged: both match pre-existing siblings (`tappableRow` in this
same file, `SearchHeader.tsx`'s input) and there is no token for either value
(`theme/tokens.ts` has `md: 8`/`lg: 12` and no font-size scale). Tokenising just
this row makes it inconsistent with the rows it's deliberately matching.

**4. "Per CLAUDE.md, default to no comments"** — that rule isn't in CLAUDE.md;
this repo's guidance is about copy and naming, not comment density. The comments
in question explain *why* (name lives outside `ClimbFilters`, so neither
`hasActive*` helper can see it) — which is exactly the invariant whose absence
caused the blocking defect the QA pass caught earlier today. Keeping them.

The one note from this round I did act on was the earlier optional-props finding
(`f648c2b`). CI is green on `f648c2b`.

Issue #4003, opened 2026-07-30
Offline My Boards: restore the owned/followed split by persisting the current user id


marcodejongh:
## Context

#3897 made `/boards/manage` usable offline: instead of the hard "Something went wrong / Try again" state it now renders the boards this device has downloaded. But it renders them **flat**, under a single un-headed section, with no "Your boards" / "Following" split.

That is not a styling choice — it is honesty. The split is `board.ownerId === currentUserId`, and `currentUserId` comes from `useProfile`, a plain network query. Offline it is `undefined`, so every board would be filed under "Following", including the user's own home wall. A flat list is better than a confidently wrong one.

## What would fix it

Persist the signed-in user id so the classification survives a cold start with no signal. Two candidate sources already exist:

- `resolveAuthSession()` returns a `userId` on the authenticated path (`packages/mobile/src/lib/auth-session.ts`).
- `UserStorageOwner` (`packages/mobile/src/lib/user-storage-owner.ts`) already carries an identity used for per-user storage scoping.

Either way it is a second persisted identity field plus a write point, and it has the same shared-device clearing obligation as the board snapshots (must be wiped in `runSignedOutCleanup`).

## Scope

- Persist the current user id at sign-in / session resolve.
- Clear it at sign-out, next to `syncEnabledBoards` and `offlineBoardsV1`.
- `/boards/manage` offline: restore the owned/followed headers using it.
- Test: offline, with a persisted user id, the user's own board lands under "Your boards" and a followed board under "Following".

## Not in scope

Making edit / delete / unfollow / create work offline — those are server mutations and stay hidden with no signal.

Low priority: the flat list is correct, just less rich.

marcodejongh:
🤖 Claimed by the 2026-07-31 bug-fix campaign (Fable session `fablebugfix`, batch 3). Assigned to avoid duplicate work across concurrent agent campaigns — a PR from this session will follow. If another agent needs this issue, comment here first.

marcodejongh:
🤖 Claimed by bug-batch loop `claude-bb21` (batch 3, session c321d1b9, 2026-08-11). Where a prior 2026-07-31 claim exists it has lapsed (10+ days, no PR/branch) — reclaiming per protocol. A PR from this loop will follow; parallel loops please pick another issue.

marcodejongh:
Now tracked under the offline-mode epic #4319 (Tier 2/4: trust + discoverability).

Issue #4002, opened 2026-07-30
Design: should mobile persist the React Query cache, or keep solving offline reads per-surface?


marcodejongh:
## The gap

Nothing on mobile persists the React Query cache. `grep -rn "persistQueryClient\|createAsyncStoragePersister\|PersistQueryClientProvider" packages/mobile` returns nothing, so every cold start begins with an empty cache. Combined with `networkMode: 'offlineFirst'` (which makes an offline query **pause** rather than error), any screen whose data comes from a plain `getHttpClient().request` renders as "you have nothing" offline rather than "we can't reach the server".

Known casualties so far:

- `myBoards` — the board picker and My Boards claimed "No boards yet — create one" offline (#3897).
- `profile` — `currentUserId` is undefined offline, which is what actually broke `/boards/manage` (a hard "Something went wrong" state regardless of the board list).
- `myGyms`, playlists, and anything else network-only follow the same shape.

#3897 fixed the two board surfaces by snapshotting board identity into MMKV. That is deliberately narrow: it fixes those two screens and nothing else.

## The systematic alternative

`@tanstack/query-async-storage-persister` + `PersistQueryClientProvider` would hydrate the cache from disk on cold start, fixing `myBoards`, `profile`, `myGyms` and the rest in one stroke — one dependency, one provider.

It was rejected for #3897 (a P3) because it is not a local change: it alters cold-start behaviour for **every** query in the app.

## What has to be settled before adopting it

1. **User scoping.** Query keys carry no user/token dimension today, and sign-out relies on `queryClient.clear()`. A persisted blob has to be keyed per account, or a shared device leaks one user's data into the next session.
2. **Staleness on launch.** Every screen would render cached data first. Which queries want that, and which must not (leaderboards, session/party state, anything with a "now" semantic)?
3. **Hydration order** vs. the auth gate and the offline-sync bootstrap.
4. **Size.** Search results and climb lists are big; the persisted blob needs a `dehydrate` filter (allowlist) rather than "persist everything".
5. **Interaction with `@boardsesh/offline-sync`.** Board catalogs already live in SQLite. Persisting the query cache on top risks two sources of truth for the same rows.

## Ask

[...]

Issue #4001, opened 2026-07-30
Offline cold start can sign the user out on native (resolveAuthSession 'unavailable' → setIsAuthenticated(false))


marcodejongh:
## What happens

On native, a cold start with **no network** can drop the user to the "Sign in" wall even though their session is perfectly valid — every downloaded board, every offline surface, becomes unreachable until they have signal again.

The chain:

1. `isTokenExpiringSoon()` (`packages/mobile/src/lib/auth-store.ts:138`) returns `true` when the token expires within 24h **or when no expiry is stored at all** (`if (!expiresAt) return true;`).
2. `resolveAuthSession()` (`packages/mobile/src/lib/auth-session.ts`) therefore attempts a refresh. Offline that refresh can't reach the server, so `deduplicatedRefresh()` comes back `unavailable` and `resolveAuthSession` returns `{ status: 'unavailable' }` — deliberately, with the comment "Preserve the session and let the caller retry, instead of a false logout."
3. The caller does not preserve it. `auth-provider.tsx` handles `unavailable` by branching on platform, and the **native** branch calls `setIsAuthenticated(false)` (line ~371). Web gets the careful treatment (`isSessionUnavailable` + a retry route); native gets the sign-in wall.

So `resolveAuthSession` does the right thing and the native consumer undoes it.

## Why it matters

A spurious sign-out offline is worse than being unable to browse: the user's tokens are still in the keychain, nothing was revoked, and the UI is simply lying about who they are. It also masks unrelated offline work — an offline board picker, offline climb search, offline tick logging are all behind this wall, so any offline QA that hits this reads as "the offline feature doesn't work".

Found while fixing #3897 (offline board switching); explicitly out of scope there and listed as a QA gate on that PR: if an offline cold start shows "Sign in" instead of the picker, this is why.

## What good looks like

Native `unavailable` should keep the established session rendered (the same intent the web branch already implements) and retry when the app next becomes active / connectivity returns, rather than flipping `isAuthenticated` to false. A no-stored-expiry token probably also shouldn't be treated as "expiring soon" when we're offline and can't do anything about it.

## Notes

- Not reproduced against production telemetry — this is a code-read finding, so the first job is a device repro (airplane mode + force-quit + cold start on a build whose stored expiry is missing or inside 24h).
- Native-only. Web already has the graceful path.

marcodejongh:
[...]

PR #4004, opened 2026-07-30
fix(mobile): show downloaded boards in the picker with no signal

### LINKED ISSUE #3897: Offline: board picker / switching between downloaded boards without signal
## Symptom

With no signal, a user can browse their *active* downloaded board (fixed in #3888), but **selecting or switching to a different downloaded board** still needs the network. The board selection screen (`packages/mobile/app/boards/index.tsx`) sources its list from `useMyBoards()` — a network GraphQL query with no local fallback. Offline that query can't resolve, so the picker never shows the boards the user has actually downloaded.

## Entry paths affected

- First board selection when there's no persisted active board (Climbs "no board" empty state → `router.push('/boards')`).
- The board switcher footer in the "now on the wall" sheet (`BoardSheet` → `onSwitchBoard`).
- `packages/mobile/app/boards/manage.tsx` (offline toggles) leans on the same network board list.

Not on the #3888 repro path (that entry is the persisted active board via the Climbs tab), so it was intentionally split out.

## Suggested approach

Derive the offline board rows from data already on the device rather than adding a new persistence surface:

- `getSetting('syncEnabledBoards')` (the per-board offline opt-in scope keys) ∪ `getDownloadedScopeKeys(db)` (`@boardsesh/offline-sync`) for "which boards are downloaded".
- The persisted active board (`active-board-store.ts`) for a full `UserBoard`.
- Bundled `@boardsesh/board-config` / `@boardsesh/board-constants` metadata for display names.

When offline (`useIsOffline()`), render those rows in the picker/switcher instead of the network-backed list, so a user can get into any downloaded board with no signal. Note the scope key (`boardType:layoutId:sizeId`) lacks angle/setIds/uuid/name, so reconstructing a full `UserBoard` for a non-active downloaded board is the main design question — resolve it from local sources, not a new store.

## Priority

P3 — the active board already opens offline after #3888; this is the multi-board / switch case.

marcodejongh:
## Premise correction (read this first)

The issue and the recon get three things wrong, and one of them changes the shape of the fix.

**1. It is not a hang.** The recon says the picker "hangs on a network query". It does not. `createQueryClient()` sets `networkMode: 'offlineFirst'`, so offline the fetch runs, fails, and the retryer **pauses** — query-core's `canContinue()` requires `onlineManager.isOnline()` unless the mode is `'always'`. I proved the terminal state against the real `createQueryClient()` with a rejecting `queryFn` and `onlineManager.setOnline(false)`: after ~1s the query settles at `status: 'pending'`, `fetchStatus: 'paused'`, `data: undefined`, `isError: false`, `isLoading: false`. (That was a diagnostic, not a shipped test — it asserts React Query's semantics, not ours.)

So `app/boards/index.tsx` rendered its **"No boards yet — create one" empty state**: a false claim that the user has no boards, whose only CTA also needs the network. That is worse than a spinner, and it means we had to replace an *empty* state, not a *loading* state. The existing `isError` retry screen never appeared offline at all.

**2. `boards/manage.tsx` breaks for a different reason than stated.** The issue says it "leans on the same network board list". It does, but that is not what kills it. The guard is `if ((isError && myBoards.length === 0) || !currentUserId)`, and `useProfile` is a raw `getHttpClient().request` too — so offline `currentUserId` is `undefined` and the screen shows the hard "Something went wrong / Try again" state **regardless of anything done to the board list**. Fixing that screen meant handling the missing profile id, not just the boards.

**3. The "BoardSheet switcher footer" is not a third surface.** `BoardSheet` → `NowOnTheWallPanel.handleSwitchBoard` just calls `onSwitchBoard()`, which navigates to `/boards`. Two broken screens, not three.

And a correction to the design framing: the scope key does not merely *lack* angle/setIds/uuid/name — it **cannot identify a board at all**. `storage-board-label.ts` already documents that one scope key (`boardType:layoutId:sizeId`) is shared by several of the user's boards ("Marco's garage" and "Gym wall" on the same Kilter Original 12x12 share ONE download). Any design that maps scope key → one board is wrong by construction. That is why the store is a flat list deduped by `uuid`, not a map.

## Summary

Closes #3897.

**Persist, not reconstruct.** `uuid` is server-issued and unreconstructible — `setActiveBoard`, `BoardProvider boardUuid`, `bluetooth-provider-wrapper`, and `drawer-host-provider`'s board-presence subscription all key on it, so a synthetic one would subscribe presence to a nonexistent board and could mint a duplicate board on reconnect. `setIds`, `name` and `angle` are user data that `@boardsesh/board-config` cannot know. So board identity is snapshotted from data the app already holds in hand, at the moment offline is enabled, into **one more key in the existing MMKV settings store** (`offlineBoardsV1`, next to `syncEnabledBoards`). No new storage engine, no new lifecycle, no new dependency; MMKV is synchronous, so the offline picker paints its rows with no async flash.

What changed:

- **New settings key + store** (`src/settings/offline-boards.ts`): upsert by `uuid`, most-recent-first, capped at 20 (the realistic maximum is a handful — the cap only stops unbounded growth). Writes are skipped when the result is byte-identical, because every settings write calls `emitChange()` and re-renders every `useSetting` consumer app-wide.
- **Runtime shape guard** (`src/lib/boards/offline-board-card.ts`): every read drops cards that fail it, silently. A `UserBoard` shape change must cost one row, never crash the picker — the hazard `active-board-store.ts`'s header already warns about.
- **Pure row selector** (`src/components/board-discovery/offline-board-items.ts`): candidates = snapshots ∪ warm `myBoards` ∪ active board, deduped by `uuid` freshest-source-wins; only scopes in `getDownloadedScopeKeys()` are offered (the honest "will actually serve climbs" signal, same one My Boards captions "Available offline"), **plus** the active board unconditionally so the board you're standing in front of can't vanish.
- **Write points, all local, no new network call**: the single `enableBoardsOffline` funnel (so the manage toggle, adopt-found-board and "download all" are covered automatically), the toggle-off, removal, delete and unfollow paths, and an online refresh hook that keeps names fresh after an edit on another device, backfills boards downloaded by a pre-fix build, and prunes boards the server no longer lists.
- **Sign-out clears it** next to the existing `syncEnabledBoards` reset — the snapshots carry board **names**, so on a shared device a missed clear would show one account's walls in the next account's picker. Asserted in `auth-provider.test.tsx`, not left to code review.
- **`app/boards/index.tsx`**: offline renders the downloaded boards through the existing `userBoardToItem` + `BoardCarousel`, hides the network-only mode cards and Popular, and replaces both the "No boards yet" empty state and the error screen. `onSelectMyBoard` now resolves from the offline rows too (it previously looked up only `myBoards`/`nearby` — a dead tap). Activation skips `adoptFoundBoard` whenever the rows came from the local snapshots: it is a follow mutation plus a download confirm, so with no usable connection it can only produce a "Could not follow X" error toast on a board the user already has.
- **`app/boards/manage.tsx`**: offline renders a flat, un-headed list with `Edit`, `Create` and swipe unfollow/delete suppressed (all server mutations, via a new `readOnly` prop on `BoardManageRow`). The per-row offline toggle and status caption keep working — both are local. Owned-vs-followed is genuinely unclassifiable without the profile id, so the headers are dropped rather than guessed.
- **The "lying connection" fallback** (`isError` + nothing cached → offline rows) is in, per the ruling: gym wifi with a dead upstream or a captive portal reports online, so retries never pause and the query really errors. When that happens with nothing downloaded, the retry button stays — the escape hatch isn't lost.
- i18n keys in all three locales (Spanish uses **plafón** per the glossary); a section in `docs/offline-sync-plan.md` recording why identity is snapshotted.

Pure JS/TS — no native change, so this ships OTA on the existing fingerprint.

### Deviations from the approved plan

- The shape guard lives in its own pure module (`src/lib/boards/offline-board-card.ts`) rather than inside the settings store, so the selector can use it without importing MMKV. Three files where the plan said two.
- `manage.tsx`'s `['downloadedScopeKeys']` query is no longer gated on the offline-downloads flag. The offline fallback needs it, and a device still holding downloads after a kill-switch rollback must not be stranded — same reasoning `more.tsx` already applies to its ungated copy.
- The offline branch on `manage.tsx` also triggers when the profile settles without an id **online** (e.g. a profile 401 with boards loaded). Today that shows the hard error state; now it shows the downloaded boards read-only. Strictly better, but it is a behaviour change outside the offline path — flagging it explicitly.
- `BoardManageRow` gained a `readOnly` prop. Passing no-op handlers would have left the swipe and the chevron visibly present but dead.

## Review round 2 — three real defects, all fixed

The paired QA reviewer found three should-fix defects. All three were re-derived from the code and confirmed; all three are fixed in this PR, each with a test that fails without the fix.

**1. The adopt guard was on the wrong flag** (`app/boards/index.tsx`). The rows render under `isLocalOnly = isOffline || (isError && myBoards.length === 0)`, but the "don't follow over a dead link" guard read `if (!isOffline)`. On the lying-connection branch — the one ruling 3 asked for — `isOffline` is false, so tapping an offline row still fired `followBoard` and raised exactly the "Could not follow X" toast (plus a Sentry `reportError`) this PR claims to have removed, on a board the user already owns and has downloaded. Now gated on `isLocalOnly`. The lying-connection test asserts `adoptFoundBoard` stays unfired, and a new online case asserts it still *does* fire when the network list is usable, so the guard can't be over-applied either.

**2. The refresh hook undid the toggle-off it was meant to respect** (`src/offline/use-remember-downloaded-boards.ts`). It remembered every board whose scope was in `syncEnabledBoards` **∪ `downloadedScopeKeys`**. Turning "Available offline" off deliberately leaves the rows and checkpoint on disk (so re-enabling resumes instantly), so the scope stays *downloaded*; the effect re-ran the instant the toggle wrote and re-remembered the board `forgetOfflineBoardScope` had dropped two lines earlier in `manage.tsx`. The `downloadedScopeKeys` term was never needed for the stated pre-fix backfill either — enabling is the only way to download, so an old download is by definition still in `syncEnabledBoards`. The union is gone (and with it the hook's extra SQLite query), and the test that encoded the bug as correct behaviour has been rewritten.

**3. Nothing pruned a card for a board that no longer exists** (`src/settings/offline-boards.ts`, `app/boards/manage.tsx`). `rememberOfflineBoards` is add-only, `forgetOfflineBoardScope` is keyed on the scope, and delete/unfollow touched neither `syncEnabledBoards` nor the checkpoint — so an unfollowed board kept its card forever, and activating it would write a `uuid` the backend no longer knows into `active-board-store`, which is the board-presence poisoning this design refuses to risk with synthetic uuids. Two fixes: `forgetOfflineBoard(uuid)` on the delete and unfollow success paths (deterministic, and it leaves a scope sibling alone), and `pruneOfflineBoards()` in the refresh hook for the deleted-on-another-device case. The prune only runs on a **complete** list (`hasMore === false`) — `myBoards` pages at 20, and pruning off a truncated first page would delete every card past it.

Also from that review:

- `BoardManageRow.readOnly` now has real component tests. The manage-screen test only asserted the prop was passed through, and deleting all three `!readOnly` terms inside the component left the suite green. The new cases record the `SwipeableRow` props and assert `onPress`, `enabled` and the chevron are all gone read-only, present otherwise — and that the local offline toggle still works.
- Narrowing an inaccurate claim: the pre-fix backfill needs the user to open the board picker or My Boards **while online**, not merely to launch the app online — the hook is mounted on those two screens only. Corrected in the QA gates below.

## Review-bot follow-ups (round 2)

Both of `claude-review`'s remaining notes are fixed rather than deferred:

- **The "No signal" notice was lying on two branches.** `mobile.offline.pickerNotice` was shown for the offline, lying-connection AND profile-failure paths, but only the first has no signal. New key `mobile.offline.pickerNoticeUnreachable` ("Can't reach your boards right now — here are the ones you've downloaded.") in all three locales, picked on `!isOffline` in both screens, asserted in both screen tests.
- **`clearOfflineBoards` wrote unconditionally.** It now early-outs, but on the RAW stored value rather than `getOfflineBoards()` — a card that fails the shape guard reads as empty and is still the previous account's board name sitting on disk. Both directions are tested.

From the bot's earlier pass, two points that had been left open:

- `remove-offline-board.ts`'s `forgetOfflineBoardScope` call now has a test (remove one of two downloaded scopes, the other scope's card survives).
- `offline-board-items.ts` sorts with `{ sensitivity: 'base' }` so capitalisation doesn't split the list. The locale stays the device's on purpose — a Spanish user should get Spanish collation for their own board names.

## Release Notes

- Switch boards with no signal — the picker now lists the boards you've downloaded instead of claiming you have none.
- My Boards works offline too, so you can see what's downloaded from inside the gym dead zone.
- No more "Could not follow" error when you pick a downloaded board offline.

## Test plan

`vp run typecheck` (full monorepo), `vp check` (0 errors), `vp run test:mobile` → **562 files / 4730 tests passed** (baseline on `main` was 557/4680), `vp test run packages/shared/i18n` (60 catalog-parity tests), `vp run check:mobile-bundle` → OK.

New tests:

- `src/components/board-discovery/__tests__/offline-board-items.test.ts` — pure, literal fixtures with **hardcoded** scope-key strings on the expectation side, deliberately not re-derived with `offlineBoardKeyForBoard` (a test that rebuilds the production expression can only agree with itself — the shape that hid a real `ne`/`eq` inversion in this repo before). Covers: two boards sharing one scope both appear (the regression the whole design turns on), a non-downloaded scope is excluded, the active board is always offered and sorts first, freshest-source-wins, and a malformed legacy card is dropped rather than rendered or thrown on.
- `src/settings/__tests__/offline-boards.test.ts` — upsert by uuid, most-recent-first, the cap, **no write when nothing changed** (asserted on the mocked `storage.set`), forgetting a scope drops both of its cards, forgetting one uuid leaves its scope sibling alone, pruning against a complete server list, corrupt/non-array values read as `[]`.
- `app/boards/__tests__/index-offline.test.tsx` — mocked with the **real** offline shapes (`useMyBoards` → `{ data: undefined, isLoading: false, isError: false }`). Row renders and switches, no adopt/follow mutation offline **or on the lying connection**, adopt still fires online, offline empty state with no "create a board" CTA, the retry escape hatch, and an explicit "online is unchanged" case.
- `app/boards/__tests__/manage-offline.test.tsx` — offline with `useProfile` undefined renders the boards and not the error state, hides Create/Edit/row mutations while the offline toggle still reads `downloaded`, leaves the owned/followed split and Create alone online, and delete/unfollow forget the board's card (but not when the mutation fails).
- `src/offline/__tests__/use-remember-downloaded-boards.test.tsx` — the enabled-scope filter, the pre-fix-build backfill, **no re-remember after a toggle-off**, the prune on a complete list, **no prune on a truncated page**, the offline guard, the unresolved-list guard, and effect-dep stability across re-renders (the settings-churn risk).
- `src/components/board-discovery/__tests__/board-manage-row.test.tsx` — the `readOnly` prop actually removes tap-to-edit, the swipe action and the chevron, and keeps the local offline toggle.
- `auth-provider.test.tsx` — sign-out clears both `syncEnabledBoards` and the snapshots.

Added after the first review pass: an explicit case for the online profile-failure path on `manage.tsx` (`isOffline = false`, no user id, boards loaded → read-only list, not the error state).

### Mutation checks (run, not claimed)

Each guard was broken locally and the suite re-run:

| Mutation | Result |
| --- | --- |
| Delete the `downloadedScopeKeys` filter in the selector | 2 red (`excludes a card whose scope has no download`, `returns nothing when no scope is downloaded`) |
| Invert freshest-source-wins (cards written last) | 2 red (both freshness tests) |
| Drop the serialize-compare in `rememberOfflineBoards` | 1 red (`does not write when the remembered list is unchanged`) |
| Force `isLocalOnly = false` in `boards/index.tsx` | 5 of 6 red; the "online unchanged" case stayed green, as it should |
| Force `showOfflineList = false` in `boards/manage.tsx` | 2 red |
| Delete the `isOffline` guard in `useRememberDownloadedBoards` | 1 red (`does nothing while offline`) |

Round 2, same method:

| Mutation | Result |
| --- | --- |
| Put the adopt guard back on `isOffline` instead of `isLocalOnly` | 1 red (the lying-connection case) |
| Remember every board instead of only enabled scopes, and drop the `hasMore` prune gate | 5 of 8 red in the hook suite |
| Delete both `forgetOfflineBoard(board.uuid)` calls in `manage.tsx` | 2 red |
| Delete the three `!readOnly` terms in `BoardManageRow` | 1 red (all three affordances asserted in one case) |

All restored afterwards.

## Device-QA / maintainer gates

The decisive checks cannot run here (Linux: no macOS, no iOS simulator, no device with real radios — `vp run check:mobile-simulator` and `vp run mobile:screenshot` skip). The Android emulator can do airplane mode but not a captive portal.

Needed on device, per platform (iOS + Android), **on a build that includes this fix** — the upgrade-path gap means a user who downloaded on a pre-fix build has no snapshots until they next open the board picker or My Boards **while online** (the backfill hook is mounted on those two screens only, not app-wide), so the repro must download on the fixed build:

1. Enable offline for **two** boards, wait for both to reach "Available offline".
2. Force-quit, airplane mode, cold start. Climbs → board switcher → `/boards`. Expect both boards listed, the offline notice, no "No boards yet", no "Create a board" CTA, no error screen.
3. Switch to the **non-active** board: climbs populate from local SQLite, no "Could not follow" toast.
4. Switch back; the active-board marker follows correctly.
5. **Two boards sharing one layout+size** (the `storage-board-label.ts` case) — both rows appear with their own names.
6. `/boards/manage` offline: list renders, no error state, offline toggles still reflect real state, no Edit/Create.
7. **Lying connection**: join wifi with no upstream (or a captive portal) and repeat step 2 — this exercises the `isError` branch, not the `isOffline` one.
8. Sign out, sign in as a second account: the first account's board names must not appear.
9. Unfollow (or delete) one of the two downloaded boards while online, then repeat step 2 — the unfollowed board must be gone from the offline picker. Turning its "Available offline" toggle off should have the same effect.
10. **Warm-session narrowing, by design, worth a look:** go offline mid-session with a board in `myBoards` that is *not* downloaded. The picker now narrows to downloaded boards only, so that board disappears rather than being offered with no climbs behind it. Confirm that reads as correct on device.
11. **Known interaction, not a failure of this PR**: if the offline cold start shows the "Sign in" wall instead of the picker, that is `resolveAuthSession` returning `unavailable` and the native branch calling `setIsAuthenticated(false)` — filed as #4001. Note it and link there rather than reopening this.

## Follow-ups filed

- #4001 — offline cold start can sign the user out on native (`resolveAuthSession` `unavailable` → `setIsAuthenticated(false)`). A spurious sign-out offline is worse than the bug fixed here.
- #4002 — design: persist the React Query cache, or keep solving offline reads per-surface? Would subsume this fix plus `profile`/`myGyms`/playlists, but changes cold-start behaviour for every query — not something to smuggle into a P3.
- #4003 — restore the owned/followed split on offline My Boards by persisting the current user id.

## Deliberately not in scope

Making create / follow / unfollow / delete / board edit work offline (the picker hides them instead); Nearby / Popular / gym-finder discovery (inherently network); a `user_boards` SQLite table or any change to `syncEnabledBoards`' shape or the pull client; `offline-request.ts`; any `packages/web` change.

### DIFF
--- a/docs/offline-sync-plan.md
+++ b/docs/offline-sync-plan.md
@@ -769,6 +769,30 @@ async function saveTick(db: SQLiteDatabase, tickData: TickInput) {
 | Backend GraphQL API   | Mostly unchanged. New sync pull queries + sync_deletions + idempotent mutations + addFavorite/removeFavorite. |
 | Aurora sync daemon    | Unchanged. Picks up ticks without `aurora_id` and pushes to Aurora API.                                       |
 
+## Board identity offline (why snapshots, not scope keys)
+
+The board picker and My Boards get their list from `myBoards`, a network-only GraphQL
+query. Offline the client's `networkMode: 'offlineFirst'` makes it PAUSE rather than
+error, so both screens saw an empty list and told the user they had no boards (#3897).
+
+The enabled-boards setting can't stand in for that list. A scope key is
+`"boardType:layoutId:sizeId"` — it names a DOWNLOAD, not a board, and one download
+serves every board the user has on that layout+size ("Marco's garage" and "Gym wall"
+on the same Kilter Original 12x12 share one). It carries no `uuid`, `name`, `setIds`
+or `angle`, and `uuid` is server-issued: `setActiveBoard`, `BoardProvider`, the BLE
+wrapper and the board-presence subscription all key on it, so it can never be
+synthesised from board-config.
+
+So board identity is snapshotted at download time into one more MMKV settings key
+(`offlineBoardsV1`, see `packages/mobile/src/settings/offline-boards.ts`) as a flat
+list of `UserBoard`s deduped by `uuid` — never a scope-keyed map, which would hide one
+of two boards sharing a scope. Reads run through a runtime shape guard so a card
+written by an older build degrades (one fewer row) instead of crashing the picker, and
+the list is cleared at sign-out because it carries the previous account's board names.
+
+The picker offers a snapshot only when its scope is in `getDownloadedScopeKeys()` —
+the honest "will actually serve climbs" signal — plus the active board unconditionally.
+
 ## Account lifecycle
 
 On logout or account switch:
@@ -776,7 +800,8 @@ On logout or account switch:
 1. Cancel any in-progress sync pull or mutation queue drain.
 2. Delete user data tables from SQLite (`boardsesh_ticks`, `playlists`, `playlist_climbs`, `user_favorites`, `user_follows`, `setter_follows`, `playlist_follows`, `user_playlist_pins`, `pending_mutations`). Board reference data stays — no need to re-copy the 150-200MB pre-warmed database.
 3. Clear the TanStack Query cache.
-4. On new login, sync pull client fetches the new user's data (small, seconds).
+4. Reset `syncEnabledBoards` and the `offlineBoardsV1` snapshots (board names must not survive into the next account on a shared device).
+5. On new login, sync pull client fetches the new user's data (small, seconds).
 
 ## Performance targets
 
--- a/docs/offline-sync-plan.md
+++ b/docs/offline-sync-plan.md
@@ -793,6 +793,22 @@ the list is cleared at sign-out because it carries the previous account's board 
 The picker offers a snapshot only when its scope is in `getDownloadedScopeKeys()` —
 the honest "will actually serve climbs" signal — plus the active board unconditionally.
 
+Card lifecycle, since no single event covers it:
+
+- **Written** by `enableBoardsOffline` (the one download funnel) and refreshed from a
+  live `myBoards` by `useRememberDownloadedBoards`, which remembers boards whose scope
+  is in `syncEnabledBoards`. Deliberately not "or already downloaded": toggling
+  "Available offline" off leaves the rows and checkpoint on disk so re-enabling
+  resumes instantly, so a downloaded-keyed refresh would re-write the card the toggle
+  just dropped.
+- **Dropped per scope** (`forgetOfflineBoardScope`) when offline is turned off or the
+  data is removed — the download is per scope, so every board sharing it loses its card.
+- **Dropped per board** (`forgetOfflineBoard`) on delete and unfollow, and pruned
+  against a **complete** `myBoards` (`hasMore === false`) for the deleted-on-another-
+  device case. A card the backend no longer knows is worse than a stale row: activating
+  it writes a dead `uuid` into `active-board-store` and board presence. `myBoards` pages
+  at 20, which is why the prune refuses to run on a truncated page.
+
 ## Account lifecycle
 
 On logout or account switch:
--- a/packages/mobile/app/boards/__tests__/index-offline.test.tsx
+++ b/packages/mobile/app/boards/__tests__/index-offline.test.tsx
@@ -233,7 +233,7 @@ describe('board picker with no usable network list', () => {
     expect(screen.queryByText('Try again')).toBeNull();
   });
 
-  it('falls back to downloaded boards when the connection lies (online, every request fails)', () => {
+  it('falls back to downloaded boards when the connection lies (online, every request fails)', async () => {
     // Captive portal / gym wifi with a dead upstream: onlineManager reports online, so
     // retries never pause and the query really errors.
     state.isOffline = false;
@@ -243,8 +243,15 @@ describe('board picker with no usable network list', () => {
 
     render(createElement(BoardSelection));
 
-    expect(screen.getByRole('button', { name: 'Marco garage' })).toBeTruthy();
+    const row = screen.getByRole('button', { name: 'Marco garage' });
     expect(screen.queryByText('Something went wrong')).toBeNull();
+
+    fireEvent.click(row);
+    await waitFor(() => expect(setActiveBoardMock).toHaveBeenCalledTimes(1));
+    // The adopt guard must follow the rows, not `isOffline`: here `isOffline` is false
+    // but every request still fails, so a follow mutation can only produce the
+    // "Could not follow X" toast (plus a Sentry report) on a downloaded board.
+    expect(adoptFoundBoardMock).not.toHaveBeenCalled();
   });
 
   it('keeps a retry escape hatch when the connection lies and nothing is downloaded', () => {
@@ -255,6 +262,22 @@ describe('board picker with no usable network list', () => {
 
     expect(screen.getByText('Nothing downloaded yet')).toBeTruthy();
     expect(screen.getByRole('button', { name: 'Try again' })).toBeTruthy();
+  });
+
+  it('still adopts on tap when the network list is usable', async () => {
+    // The other direction of the same guard: online, selecting a board must still
+    // follow it and offer its download.
+    state.myBoards = {
+      data: { boards: [board({ uuid: 'net-1', name: 'Network board' })] },
+      isLoading: false,
+      isError: false,
+      isRefetching: false,
+    };
+
+    render(createElement(BoardSelection));
+    fireEvent.click(screen.getByRole('button', { name: 'Network board' }));
+
+    await waitFor(() => expect(adoptFoundBoardMock).toHaveBeenCalledTimes(1));
   });
 
   it('leaves the online screen untouched', () => {
--- a/packages/mobile/app/boards/__tests__/index-offline.test.tsx
+++ b/packages/mobile/app/boards/__tests__/index-offline.test.tsx
@@ -73,6 +73,8 @@ vi.mock('react-i18next', () => ({
         'mobile.errorTitle': 'Something went wrong',
         'mobile.errorRetry': 'Try again',
         'mobile.offline.pickerNotice': "No signal — here are the boards you've downloaded.",
+        'mobile.offline.pickerNoticeUnreachable':
+          "Can't reach your boards right now — here are the ones you've downloaded.",
         'mobile.offline.pickerEmptyTitle': 'Nothing downloaded yet',
         'mobile.offline.pickerEmptyBody': 'Boards you make available offline show up here.',
         'mobile.discovery.yourBoardsTitle': 'Your boards',
@@ -245,6 +247,9 @@ describe('board picker with no usable network list', () => {
 
     const row = screen.getByRole('button', { name: 'Marco garage' });
     expect(screen.queryByText('Something went wrong')).toBeNull();
+    // The device has bars here, so "No signal" would be a lie.
+    expect(screen.getByText("Can't reach your boards right now — here are the ones you've downloaded.")).toBeTruthy();
+    expect(screen.queryByText("No signal — here are the boards you've downloaded.")).toBeNull();
 
     fireEvent.click(row);
     await waitFor(() => expect(setActiveBoardMock).toHaveBeenCalledTimes(1));
--- a/packages/mobile/app/boards/__tests__/index-offline.test.tsx
+++ b/packages/mobile/app/boards/__tests__/index-offline.test.tsx
@@ -1,0 +1,297 @@
+// @vitest-environment jsdom
+//
+// Regression cover for #3897: with no signal the board picker used to render its
+// "No boards yet — create one" empty state. `myBoards` is a plain network query and
+// the client runs `networkMode: 'offlineFirst'`, so offline the retryer PAUSES
+// (status pending, fetchStatus paused, isError false) — the screen therefore saw an
+// empty list, not an error, and told the user they had no boards while offering a
+// CTA that also needs the network.
+//
+// The mocked hook shapes below are the real offline shapes: `useMyBoards` returns
+// `{ data: undefined, isLoading: false, isError: false }`.
+import { describe, it, expect, vi, beforeEach } from 'vitest';
+import { render, screen, fireEvent, waitFor } from '@testing-library/react';
+import { createElement, type ReactNode } from 'react';
+import type { UserBoard } from '@boardsesh/shared-schema';
+
+type Children = { children?: ReactNode };
+type ButtonProps = { title: string; onPress?: () => void };
+type CarouselItem = { key: string; title: string };
+
+const routerMock = vi.hoisted(() => ({ push: vi.fn(), dismissTo: vi.fn() }));
+const toastMock = vi.hoisted(() => ({ showToast: vi.fn() }));
+const setActiveBoardMock = vi.hoisted(() => vi.fn().mockResolvedValue(undefined));
+const adoptFoundBoardMock = vi.hoisted(() => vi.fn().mockResolvedValue(undefined));
+const rememberDownloadedBoardsMock = vi.hoisted(() => vi.fn());
+
+const state = vi.hoisted(() => ({
+  isOffline: false,
+  offlineCards: [] as unknown[],
+  downloadedScopeKeys: [] as string[],
+  activeBoard: undefined as unknown,
+  myBoards: {
+    data: undefined as { boards: unknown[] } | undefined,
+    isLoading: false,
+    isError: false,
+    isRefetching: false,
+  },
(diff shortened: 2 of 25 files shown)


marcodejongh:
Thanks — three of the four are fixed in ec18c6e, one I'm keeping as-is with a comment.

**1. `profileUnavailable` read-only mode online — fixed.** You're right that the PR description carried the behaviour and the tests didn't. Added `degrades to the read-only list, not the error state, when the profile fails online` to `manage-offline.test.tsx`: `isOffline = false`, `profileId = undefined`, boards loaded. It asserts no error state, the downloaded board renders, no "Your boards" header, and `readOnly = true` on the row.

**2. `useRememberDownloadedBoards` untested — fixed.** New `src/offline/__tests__/use-remember-downloaded-boards.test.tsx`, six cases: the enabled/downloaded filter, the pre-fix-build backfill path, the offline guard, the "boards not resolved yet" guard, the "nothing enabled or downloaded" no-op, and effect-dep stability across re-renders (the churn risk you named). Mutation-checked: deleting `isOffline ||` from the guard turns the offline case red.

**3. Fresh `[]` per render — fixed.** `myBoards` now falls back to a module-level `EMPTY_BOARDS`, matching what `manage.tsx` already does. Negligible as you say, but the memo dep is now honest.

**4. `JSON.stringify` key-order sensitivity — keeping it, with a comment.** The failure direction is the safe one: reordered fields mean the stored bytes genuinely changed, so we do one redundant write and the next call settles. There is no path where it misses a real update. A key-insensitive compare would cost a normalising serializer for a hazard whose worst case is a single extra MMKV write on a schema change. Documented in `offline-boards.ts` next to the guard so the next reader doesn't re-derive it.

Full mobile suite after the changes: 562 files / 4714 tests passing.

marcodejongh:
Both of your remaining notes are fixed in 75901b8, plus two from the earlier pass that had been left open.

**1. Misleading "No signal" copy — fixed.** You're right, and it was wrong on two branches, not one: the lying-connection path (`isError` + nothing cached) has bars too. New key `mobile.offline.pickerNoticeUnreachable` — "Can't reach your boards right now — here are the ones you've downloaded." — in en-US, es and fr, picked on `!isOffline` in both `boards/index.tsx` and `boards/manage.tsx`. Asserted in both screen suites: the lying-connection case now checks the new copy is shown **and** the "No signal" line is not, and the online profile-failure case checks the new copy too.

**2. `clearOfflineBoards` unconditional write — fixed, with one wrinkle worth naming.** The early-out reads the **raw** stored value, not `getOfflineBoards()`. A card written by an older build fails the shape guard and reads as an empty list, but it is still the previous account's board name sitting in MMKV — a guard on the filtered view would skip the write and leave it there, on the one path where clearing is a privacy requirement. Two tests: a corrupt-but-non-empty value still gets cleared, and a genuinely empty one writes nothing.

Also picked up from your 13:48 pass:

- **`remove-offline-board.ts`'s `forgetOfflineBoardScope` was untested — now covered.** New case in `remove-offline-board.test.ts`: two downloaded scopes with a card each, remove one, the other scope's card survives.
- **`localeCompare` with no options — now `{ sensitivity: 'base' }`.** The locale argument stays `undefined` on purpose: a Spanish user should get Spanish collation for their own board names. The sensitivity option is the part that mattered, so "gym wall" and "Gym Wall" don't end up in different halves of the list.

And on your `forgetOfflineBoardScope`-on-toggle-off finding from that same pass: it was real, and the fix went the other way from your suggestion. Rather than removing the call, `useRememberDownloadedBoards` stopped keying on `downloadedScopeKeys` at all — it now remembers only boards whose scope is in `syncEnabledBoards`. The `downloadedScopeKeys` term was never needed for the pre-fix backfill (enabling is the only way to download, so an old download is still listed), and its only effect was the one you spotted: re-writing the card the toggle had just dropped. Details in the "Review round 2" section of the PR body.

Full mobile suite after all of it: 562 files / 4730 tests.

PR #4005, opened 2026-07-30
fix(shared): recover and position a coalesced activation's queue-add

### LINKED ISSUE #3936: Coalescer: superseded queue-adds append instead of inserting after current, and a drained throttle drops its add entirely (web + mobile)
Two defects in the same shared code path — `createSetCurrentClimbCoalescer` and its `create-queue-mutations` wiring. Both affect **web and mobile equally**. Split out of #3934, which fixed only the non-coalesced case and only on mobile.

## Bug 1 — a superseded activation's climb lands at the bottom of the queue

`sendSupersededQueueAdd` sends `ADD_QUEUE_ITEM` with **no `position`** (`packages/shared/queue-react/src/create-queue-mutations.ts:220-233`):

```ts
await runMutation(client, {
  query: ADD_QUEUE_ITEM,
  variables: { item: toQueueItemInput(item) },   // no position
});
```

The server appends when `position` is absent. So whenever a climber activates two climbs quickly enough that the first is superseded while in flight, the first climb **appends to the end of the queue instead of slotting in after the current climb** (#2217). To the climber, a boulder they picked shows up at the bottom of a long queue rather than next up — with no error and nothing to explain it.

This needs no rate limiting and no network failure to reproduce. Two fast taps is enough, which makes it the more frequently hit of the two bugs here.

The three add paths currently disagree about positioning: `addToQueue` appends by design, #3934's recovery passes the local index, and this one appends by omission. They should agree.

## Bug 2 — the final pending call in a burst silently drops its queue-add

`createSetCurrentClimbCoalescer` (`packages/shared/queue-runtime/src/set-current-climb-coalescer.ts`) has three exit paths, and one loses a queue-add outright:

1. **The call that starts the burst** — `await options.sendArgs(args, context)` at `:107` sits in a `try` with only a `finally`, so its rejection propagates out of `enqueue` to the caller's `.catch`. Mobile's `dispatchSetCurrent` now recovers the queue-add from there (#3934). Covered.
2. **A call superseded while pending** — `sendSupersededQueueAdd` (`:93-102`) fires its queue-add separately. Covered, modulo Bug 1's positioning.
3. **The final pending call, drained in the `finally` at `:113-119`** — never superseded, so no `sendSupersededQueueAdd`; its rejection goes to `onDrainError` (`:116-118`), which `create-queue-mutations.ts:235` maps to the dev-log-only `onBestEffortError`. **Not covered.**

In case 3 the activation carried `shouldAddToQueue: true`, the optimistic reducer already inserted the slot locally, and nothing ever re-sends it. The climb sits in that climber's queue and nobody else's, permanently and silently — the same divergence #2791 and #3934 set out to close, reached by a different route.

Worse, case 3 is the *last* activation of a rapid burst, which is exactly the one most likely to be throttled. #3934 targets rapid-fire activation and this is the tail of it.

## Why #3934 didn't fix Bug 2

`onDrainError` is `(error: unknown) => void`. It carries neither the `item` nor `shouldAddToQueue`, so no mobile-local wiring can call `recoverThrottledQueueAdd(item)` from it. Fixing it means changing shared code that web consumes.

## Design question to settle first

Two shapes, and the second is probably right:

- **Widen the error channel** — pass the drained `args` to `onDrainError` and let each platform recover. Keeps the coalescer dumb, but duplicates the recovery in web and mobile and leaves web free to keep not doing it.
- **Move the recovery into the coalescer** — on a drain failure with `shouldAddToQueue`, fire the queue-add through the same seam `sendSupersededQueueAdd` already uses. Both platforms get it in one place, next to the sibling case it mirrors. Needs a decision on how it interacts with a caller-side recovery already running for case 1 (double-add is prevented by the server's uuid idempotency, but the two shouldn't both fire blind).

Fixing Bug 1 at the same time is natural: whatever seam sends these deferred adds should carry a position, and all three add paths should agree on it.

## Severity

Bug 1: **P2** — no failure required, two fast taps reproduces it, and it is user-visible as a queue-ordering bug with no error attached.

Bug 2: **P3** — needs a genuinely throttled party session, and it self-heals on the next successful sync or rejoin. The bad part is the silence.

## Repro sketches

**Bug 1.** Solo or party. Activate a climb, then activate a second one fast enough that the first is still in flight. The first climb appends to the end of the queue instead of landing after the current climb.

**Bug 2.** Two phones in one party session. Rapid-fire activations from search fast enough that `enqueue` coalesces *and* fast enough to trip the limiter. Check the peer's queue: the last climb of the burst is missing there while present locally.

marcodejongh:
## Premise correction, up front

The issue's repro says "solo or party" and "the boulder you picked shows up at the bottom of everyone's queue".
Both are off, and one of them changes what to test:

1. **Party-only.** In solo there is no session, so `getContext()` returns null and the deferred queue-add bails on
   `!capturedSessionId` before sending anything (`create-queue-mutations.ts`). Mobile solo keeps the correct
   local order; web throws `Not connected to session` with no session. Neither bug is reachable without an
   active session.
2. **The picker does not see the bottom-of-queue order immediately.** Their own `QueueItemAdded` echo is
   uuid-idempotent (`insertQueueItemIdempotent`), so their local order is untouched. The _crew_ sees the wrong
   order right away; the picker sees it only once a FullSync or the ordered-hash drift watchdog rewrites their
   queue from the server — which the next activation's own add reliably triggers
   (`queue-navigation.ts` publishes `FullSync` whenever `addedToQueue`). Same user-visible damage, different
   mechanism.
3. **Web's exposure is much narrower than "two fast taps".** Web's ordinary activation does not use
   `shouldAddToQueue` at all — it sends an explicit `addQueueItem(item, currentIndex + 1)` then
   `setCurrentClimb(item, false, …)`. The only web caller that reaches the coalescer with a queue-add is
   promoting a playlist peek / suggested item. The mobile "two fast taps" claim holds.
4. **"Permanently and silently" overstates Bug 2.** Mobile runs the ordered-state-hash drift gate, so a
   local-only extra slot eventually triggers a resync that deletes it locally too. The durable harm is that the
   crew never gets the climb and the picker's pick disappears — not a permanent local ghost.
5. **The two platforms' optimistic inserts disagree**, which decides the fix's shape. Mobile activation inserts
   after current; mobile peek-promotion and web peek-promotion both _append_. There is no single "insert after
   current" rule to hard-code — the deferred add has to carry the position the item actually occupies in this
   client's queue.

Both defects reproduce. Verified with a throwaway vitest against the real `createQueueMutations` before writing
any fix: Bug 1 fires exactly one `ADD_QUEUE_ITEM` with `variables: { item }` and `position: undefined`; Bug 2
fires **zero** adds and surfaces only a dev-log `onBestEffortError('setCurrentClimb', …)`.

## Summary

Closes #3936. Both fixes land in the shared seam, so web and mobile get them together.

**Design call (the one the issue asked for): the issue's option 2** — recovery inside the coalescer, not a
widened `onDrainError`. Option 1 would have reused mobile's richer `recoverThrottledQueueAdd` but left web's
identical hole open forever and added a third recovery policy to a file that already has two. Approved by the
maintainer before implementation.

### `packages/shared/queue-runtime/src/set-current-climb-coalescer.ts`

- `sendSupersededQueueAdd` → `sendDeferredQueueAdd`, `onSupersededQueueAddError` → `onDeferredQueueAddError`.
  Single consumer, so the rename is compile-enforced (`grep` over the repo confirms: only the coalescer, its
  test, and `create-queue-mutations.ts`).
- One `fireDeferredQueueAdd(pending)` helper, called from **both** loss paths: superseded-while-pending
  (unchanged behaviour) and the drain loop's `catch` (new — Bug 2).
- The drain path fires the add **before** `onDrainError`, so a caller sink that throws cannot swallow the
  recovery, and **never awaits it**: a rate-limited ADD can sit in `execute`'s back-off for seconds, and
  awaiting it inside the drain would hold `inFlight` true, parking later activations as pending and
  manufacturing more supersedes.
- The helper body is `try`/`catch`-guarded. The option type promises a `Promise`, but a non-`async`
  implementation can throw synchronously — from inside the drain's `catch` that throw escapes the enclosing
  `finally` and wedges the coalescer with `inFlight` stuck true forever. Now it routes to
  `onDeferredQueueAddError` like a rejection.

### `packages/shared/queue-react/src/create-queue-mutations.ts`

- New **required** dep `getQueuePosition: (uuid) => number`. Required, not optional, so both bindings are
  compile-forced — an optional dep would silently fall back to appending, which _is_ the bug.
- The deferred add reads the position **after** the session guards and **live at send time**, then sends
  `{ item, position }`. The local index counts the items ahead of it that the server already has, so the insert
  reproduces this client's order; when an earlier item hasn't landed yet the index overshoots and the server
  clamps to an append — degrading to today's behaviour, not to a wrong position.

### Deviation from the plan (blocking review finding, and it matters)

The plan had `position === -1 → return` ("don't resurrect a climb they dropped"). **The adversarial review was
right that this makes the Bug-2 half of the fix inert in the dominant interleaving**, and I changed the design:

> the burst's head activation is itself an add → the server answers it with a `FullSync` → `FullSync` maps to
> `INITIAL_QUEUE_DATA` → the reducer **replaces** the queue wholesale (`queue: filteredQueue`) → the
> not-yet-sent pending slot is gone. The drained `SET_CURRENT_CLIMB` being recovered from is `RATE_LIMITED`,
> i.e. ≥ ~2s later, so by the time the deferred add reads the position the item has already been wiped and the
> guard would skip. Zero adds. Exactly the bug.

So absence is now split by cause:

- **the climber dropped it** → skip. Tracked by a ledger of dropped uuids. Queue-item uuids are per-slot and
  unique, so a re-added climb has a new uuid and is never falsely skipped.
- **a wholesale server sync replaced the queue** → send anyway, with no `position`. An append beats a drop:
  the crew getting the climb at the bottom is the pre-fix behaviour, dropping it is #3936.

This still satisfies maintainer ruling 3 (one shared helper, so the skip applies to the superseded path too),
it just makes the skip conditional on an actual drop instead of on mere absence.

#### What feeds the "climber dropped it" ledger (corrected after review)

The first cut of this PR claimed `removeQueueItem` was "the single choke point for every explicit removal on
both platforms". **That was wrong**, and the QA pass was right to block on it. Two climber-facing clears never
go near `removeQueueItem`:

- web's Clear button — `setQueue([])` (`play-view/queue-drawer.tsx`, `…/list/layout-client.tsx`)
- mobile's playlist "replace my queue" — `setQueue([item], item)` (`lib/playlists/use-playlist-activation.ts`)

Left as written, a climb activated just before either of those could have been resurrected onto the freshly
emptied queue by its own deferred add — new behaviour, since `main` sent nothing on that path at all. So
`setQueue` now records too: it diffs the new queue against the recent `shouldAddToQueue` activations and marks
anything the replace discarded. The diff is order-sensitive on purpose — an activation made _after_ a Clear was
never a candidate at Clear time, so a climb picked afterwards and then wiped by a server sync still reaches the
crew. Both directions are pinned by tests.

The full coverage statement, so the next reader does not have to re-derive it:

| Cause of local absence                                        | Behaviour          | Covered by                  |
| ------------------------------------------------------------- | ------------------ | --------------------------- |
| per-item remove / swipe / mobile clear-queue / #3934 undo leg | skip               | `removeQueueItem` ledger    |
| web Clear button, mobile playlist replace-my-queue            | skip               | `setQueue` diff             |
| wholesale server sync (`FullSync` → `INITIAL_QUEUE_DATA`)     | send, unpositioned | append branch               |
| **a PEER removed it mid-back-off**                            | send, unpositioned | **not covered** — see below |

A peer's removal arrives as a server delta and is indistinguishable from a sync at this layer, so it takes the
append branch and can resurrect a climb the peer just deleted. That is pre-existing on the superseded path
(`main` appended unconditionally there), it needs a signal this seam does not have, and it is called out rather
than papered over.

The ledger is also no longer a 50-entry ring. Mobile's `clearQueue` fires one `removeQueueItem` per queued
item in a single burst, so any queue longer than 50 evicted the earliest entries of its own clear and one climb
could reappear on the emptied queue. It is an insertion-ordered `Set` now, capped far above any realistic queue
purely so a pathological session cannot grow it without bound.

### Wiring

- mobile `queue-provider.tsx`: `stateRef.current.queue.findIndex(…)` — the same live read
  `recoverThrottledQueueAdd` already uses.
- web `persistent-session/hooks/use-queue-mutations.ts` + `persistent-session-context.tsx`: the reducer queue
  through a render-assigned ref, mirroring the sibling refs in that provider.
- `packages/backend/src/__tests__/helpers/headless-queue-client.ts`: third consumer the plan missed (second
  blocking review finding — it would have red-ed CI on the first push). Wired to the harness's **real** reducer
  queue, not a `() => -1` stub, which would compile and silently switch every deferred add off inside the
  integration harness that exists to exercise this seam.

No backend change, no schema change, no migration. JS-only → OTA-safe, no fingerprint change.

## Review findings, answered

| #            | Finding                                                                             | Resolution                                                                                                                                                                                                                                          |
| ------------ | ----------------------------------------------------------------------------------- | --------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
| 1 (blocking) | `-1 → skip` makes the Bug-2 fix inert after a FullSync wipe                         | **Accepted, design changed.** Removal ledger + append-on-sync-wipe. Permanent test: `still sends the deferred add when a wholesale server sync wiped the local slot`.                                                                               |
| 2 (blocking) | Backend headless harness is a third consumer                                        | **Accepted.** Wired to real reducer state.                                                                                                                                                                                                          |
| 3            | "exactly one of {caller recovery, deferred add} fires" is false                     | **Accepted.** Two tests now: the head's rejection never fires its own deferred add, and the same-uuid double-rejection case where both _do_ fire (bounded — the resolver skips a duplicate uuid).                                                   |
| 4            | Web's burst-head rejection stays unfixed                                            | **True, and stated plainly.** Web has no caller-side recovery, so a web suggested-item promotion that _heads_ a throttled burst still loses its add. Web is fixed for the superseded and drained paths only. Filed as a follow-up (see below).      |
| 5            | Deferred ADD races the drained `setCurrentClimb`, so peer order is nondeterministic | **Accepted.** Device-QA items softened below, with the race spelled out.                                                                                                                                                                            |
| 6            | Assert full `variables` objects, and test 12 needs re-deriving                      | **Accepted.** Every position assertion is an exact `toEqual` on the whole variables object (`position` is optional on the same object, so `objectContaining` would stay green if it were dropped again). Test 12 re-derived against the new design. |
| 7 (nit)      | Sync throw from the helper wedges the coalescer                                     | **Accepted.** `try`/`catch` + its own test.                                                                                                                                                                                                         |
| 8 (nit)      | Stale prose; reuse an existing `CapturedMutationDeps` file                          | **Accepted.** Comment updated; the mobile test went into `queue-provider-local-queue.test.tsx`, which already captures the deps object.                                                                                                             |

Second QA pass, on the implementation:

| #              | Finding                                                                                                                            | Resolution                                                                                                                                                                                                                             |
| -------------- | ---------------------------------------------------------------------------------------------------------------------------------- | -------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
| A (should-fix) | The "`removeQueueItem` is the single choke point" premise is false — web's Clear and mobile's playlist replace both use `setQueue` | **Accepted, and it was a real regression.** `setQueue` now records the discarded activations; the coverage table above replaces the wrong claim. Two new tests (the skip, and the ordering guard that keeps it from over-suppressing). |
| B (should-fix) | The 50-entry ring evicts the earliest entries of mobile's own >50-item clear                                                       | **Accepted.** Insertion-ordered `Set`, cap far above any realistic queue. New test drives a 60-item clear and asserts its **first** uuid is still suppressed.                                                                          |
| C (should-fix) | Release note overclaims "in the spot you put it" — the throttled path appends                                                      | **Accepted.** Copy rewritten to promise delivery always and position only when the app is not throttling.                                                                                                                              |
| D (nit)        | Device-QA gate 2's "A's pointer must not yank back" contradicts the FullSync analysis                                              | **Accepted.** `INITIAL_QUEUE_DATA` replaces `currentClimbQueueItem` too, so the pointer settling on the burst head is pre-existing on `main`. Gate reworded so it can't fail the PR.                                                   |
| E (nit)        | A peer removing the item mid-back-off takes the append branch                                                                      | **True, documented, not fixed.** In the coverage table above and at the guard. No signal for it at this layer.                                                                                                                         |
| F (nit)        | Mobile's `recoverThrottledQueueAdd` carries the same inert `position === -1` guard                                                 | **Filed as #4009** (mobile counterpart of #4006). Out of scope here — it is #3934's code path, not the coalescer's.                                                                                                                    |

## Test plan

`vp test run --reporter=agent packages/shared/queue-react packages/shared/queue-runtime` — 155 green
(baseline 139).
`vp run typecheck` (exit 0) · `vp test run --project mobile` 4681 green · `vp test run --project web` 5838
green · `vp test run --project backend` 2743 green (one flake on the first pass, green on a full re-run —
the known `VITEST_POOL_ID` sandbox flake, not this change) · `vp check` clean for every file this PR touches.

New tests, all **mutation-tested** (broken, confirmed red, restored):

| Mutation                                                         | Tests that go red               |
| ---------------------------------------------------------------- | ------------------------------- |
| delete `fireDeferredQueueAdd(next)` from the drain `catch`       | 8 (5 coalescer + 3 factory)     |
| drop `position` from the deferred ADD                            | 3                               |
| use the plan's original bare `-1 → skip`                         | 2, incl. the FullSync-wipe test |
| drop the explicit-remove skip                                    | 1                               |
| capture `getQueuePosition` at mount instead of through `depsRef` | 1                               |
| stub mobile's `getQueuePosition` to `() => -1`                   | 1                               |
| disable the `setQueue` discard diff (QA finding A)               | 1                               |
| shrink the ledger back to a 50-entry ring (QA finding B)         | 1                               |

The ordering-guard test (`still sends the deferred add when the wholesale replace happened BEFORE the
activation`) has no single-line mutation — it exists to block the cheaper design where `setQueue` stores its
resulting uuid set and the deferred add skips anything missing from it, which would re-break the FullSync case
for every climber who ever hit Clear.

The load-bearing one is `still sends the deferred add when a wholesale server sync wiped the local slot`: it
drives the real factory, replaces the local queue with a server snapshot lacking the item _while the throttled
send backs off_, and asserts an ADD still fires. That is the test the plan did not have, and it is the one that
makes the fix non-deletable.

## Device QA / maintainer gates

No native code, no BLE, no platform API — pure TS plus wiring, OTA-safe. This machine is Linux, and none of
these need macOS. **My read: desirable, not a merge gate.** Two phones on a `pr-<N>` OTA channel in one party
session (this also covers the gate #3934 left outstanding):

1. **Bug 1, no throttling needed.** On phone A activate a climb from search, then a second one fast enough that
   the first is still in flight. On phone B the first climb should land immediately after the current climb, not
   at the bottom.
2. **Bug 2.** Rapid-fire activations on A until "Too many changes too fast" appears. The **last** pick of the
   burst must now show up on B (it does not on `main`), exactly once. **Ignore where A's own pointer ends up** —
   `INITIAL_QUEUE_DATA` replaces `currentClimbQueueItem` as well as the queue, so the FullSync the burst head
   triggers can settle A back on an earlier climb. That happens on `main` too and this PR neither causes nor
   fixes it.
3. **Order caveat for 1 and 2 (review finding 5).** The deferred ADD and the drained `setCurrentClimb` are not
   ordered relative to each other, and the server anchors a `shouldAddToQueue` insert at the _server's_ current
   climb — which lags, because the superseded pointer never landed. So the tail of a heavily throttled burst can
   legitimately land one slot off, and a subsequent FullSync will publish that order. **A one-slot mismatch on a
   throttled burst is not a failure of this fix**; the climb reaching B at all is what changed. Judge item 1
   (unthrottled) strictly and item 2 loosely.
4. **Negative check.** Rapid _queue swipes_ (pointer-only, no `shouldAddToQueue`) must add nothing to either
   queue. Unit-tested, worth confirming in the wild.
5. **Web (deprecated surface, low priority).** In a party session, promote a playlist suggestion twice in quick
   succession and confirm the peer's order matches. That is web's only path into either bug.

Maintainer decisions already taken and reflected here: option 2 over option 1; `getQueuePosition` required;
one shared helper so the skip applies to both paths; close #3936 on this PR and file the web divergence
separately.

Two calls for the maintainer:

- **The deviation from ruling 3.** Absence with no matching drop now appends instead of skipping. Rationale and
  the FullSync evidence chain are above; the alternative is a fix that does nothing in the interleaving the
  issue describes.
- **The peer-removal case** (row 4 of the coverage table) stays on the append branch. If you want it skipped it
  needs a delta-origin signal this layer does not have, and it is a separate change.

## Follow-ups filed

- **Mobile's burst-head hole** (filed as #4009). `recoverThrottledQueueAdd` carries the same `position === -1`
  bail this PR had to design around, so #3934's caller-side recovery goes inert once a FullSync wipes the local
  slot — and then fires a `removeQueueItem` that undoes an add which did land. Mobile counterpart of #4006.
- **Web's burst-head hole** (filed as #4006). `queue-actions-core.ts`'s coalescer `.catch` logs and cleans up but never re-sends,
  and `enqueue` only rejects for the burst-_starting_ call — so a web suggested-item promotion that heads a
  throttled burst still loses its ADD, with neither recovery firing. Not fixed here; on the surface being
  replaced by the RN web deploy.
- **Web peek-promotion `insertAfterCurrent` divergence** (filed as #4007). `queue-actions-core.ts` omits `insertAfterCurrent`, so
  it appends locally while the server's `setCurrentClimb(shouldAddToQueue)` inserts after current — the two
  disagree with no throttling and no coalescing at all. Low priority, same deprecated surface.
- **Resync fallback for a doubly-failed deferred add.** Stays a dev-log, as the superseded path always has.

## Release Notes

- Pick climbs back to back in a party session and none of them go missing. A quick double-pick now lands right
  after the current climb for the whole crew instead of at the bottom, and a climb you picked while the app was
  throttling still reaches them rather than vanishing.

### DIFF
--- a/docs/websocket-implementation.md
+++ b/docs/websocket-implementation.md
@@ -741,6 +741,32 @@ Note that duplicate sequence numbers remain legal on one specific path: `setClim
 | `mirrorCurrentClimb`       | `ClimbMirrored`                     | Flips the mirror flag on the current climb and the matching queue item.                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                             |
 | `setQueue`                 | `FullSync`                          | Bulk replaces queue + current climb. Used for offline → online reconciliation and playlist activation updates that must atomically preserve queue history, keep manually queued future items, and drop stale future suggested items. Also carries mobile's session generate: `appendGeneratedSession` (`queue-provider.tsx`) sends the existing queue with the generated items appended and the **current climb unchanged**, so peers see the session queued behind whatever's active rather than a wholesale replace that jumps everyone to climb #1. It falls back to the first generated climb only when nothing is current, matching web's `start-sesh-drawer`. |
 
+### Deferred queue-adds from the setCurrentClimb coalescer
+
+Rapid activations are coalesced client-side (`packages/shared/queue-runtime/src/set-current-climb-coalescer.ts`): one `setCurrentClimb` in flight, the most recent pending args win. A dropped `setCurrentClimb` costs two things, not one. The pointer move self-heals — the next activation or any peer broadcast re-establishes it — but a `shouldAddToQueue: true` payload's brand-new queue slot does not: the reducer already inserted it locally and no later activation retroactively adds it, so the climb would sit in that climber's queue and nobody else's.
+
+So both ways a pending activation can lose its `setCurrentClimb` fire the content half on its own as a plain `ADD_QUEUE_ITEM`:
+
+1. **superseded while pending** — a newer activation replaced it;
+2. **drained and then rejected** — rate limit or socket blip (#3936).
+
+The deferred add is fire-and-forget (awaiting it inside the drain would hold `inFlight` through a multi-second back-off and manufacture more supersedes) and carries the item's **live local index**, read at send time rather than at enqueue time. That index counts the items ahead of it the server already has, so the insert reproduces the sending client's order; an overshoot — an earlier item that has not landed yet — is clamped to an append by the resolver.
+
+When the item is no longer in the local queue at send time, `packages/shared/queue-react/src/create-queue-mutations.ts` splits by cause:
+
+| Cause of local absence                                                    | Behaviour           |
+| ------------------------------------------------------------------------- | ------------------- |
+| per-item remove / swipe / mobile clear-queue (one `removeQueueItem` each) | skip                |
+| wholesale local replace — web's Clear, mobile's playlist replace-my-queue | skip                |
+| wholesale **server** sync (`FullSync` → `INITIAL_QUEUE_DATA`)             | send, no `position` |
+| a **peer** removed the item mid-back-off                                  | send, no `position` |
+
+The last row is a known gap, not an oversight: a peer's removal arrives as a server delta and is indistinguishable from a sync at that layer, so the add fires and can resurrect a climb the peer just deleted. Closing it needs a delta-origin signal the mutations factory does not have.
+
+The server-sync row is the one that matters most. The burst's head activation is itself an add, so the server answers it with `FullSync` (published with no `clientId`, so the origin cannot suppress its own echo), which `INITIAL_QUEUE_DATA` applies by **replacing** the local queue. A rate-limited drain rejects seconds later, by which time the pending item's optimistic slot is already gone. Skipping on mere absence would make the whole recovery inert in exactly the interleaving it exists for.
+
+The burst _head_ — the call whose `enqueue` promise actually rejects — is not covered by the coalescer at all. Mobile recovers it caller-side (`recoverThrottledQueueAdd`); web does not. Tracked in #4009 and #4006.
+
 ### Tick Mode and Queue Bar Freeze
 
 When a user opens the inline tick bar (to log a send/flash/attempt for the current climb), the queue control bar freezes navigation to prevent the active climb from changing mid-tick. This is implemented client-side in `queue-control-bar.tsx`:
--- a/packages/backend/src/__tests__/helpers/headless-queue-client.ts
+++ b/packages/backend/src/__tests__/helpers/headless-queue-client.ts
@@ -326,6 +326,10 @@ export class HeadlessParticipant {
       getClient: () => this.client,
       getSessionId: () => this.sessionId,
       toQueueItemInput,
+      // Wired to this participant's REAL reducer queue, not a stub: a `() => -1`
+      // would compile and silently switch every deferred queue-add off inside
+      // the harness that exists to exercise that seam end-to-end.
+      getQueuePosition: (uuid) => this.state.queue.findIndex((item) => item.uuid === uuid),
     });
   }
 
--- a/packages/mobile/src/providers/__tests__/queue-provider-local-queue.test.tsx
+++ b/packages/mobile/src/providers/__tests__/queue-provider-local-queue.test.tsx
@@ -67,6 +67,7 @@ const queueMutations = vi.hoisted(() => ({
 // REAL ensureReady seam (the no-lazy-create contract) is testable directly.
 type CapturedMutationDeps = {
   getSessionId: () => string | null;
+  getQueuePosition: (uuid: string) => number;
   ensureReady?: (capturedSessionId: string | null) => Promise<string | null>;
 };
 const capturedMutationDeps = vi.hoisted(() => ({ current: null as CapturedMutationDeps | null }));
@@ -345,6 +346,46 @@ describe('QueueProvider local solo queue', () => {
       ]);
       expect(snapshots.at(-1)?.state.currentClimbQueueItem?.uuid).toBe('item-x');
     });
+  });
+
+  // The wiring typecheck cannot see: wrong ref, wrong field, inverted return.
+  // A deferred queue-add (a superseded or throttled-away activation) positions
+  // itself with this, so a broken binding silently reverts #3936 to appending.
+  it('exposes the live local queue index to the shared mutations (getQueuePosition)', async () => {
+    const itemA = makeQueueItem('item-a');
+    const itemB = makeQueueItem('item-b');
+    const itemC = makeQueueItem('item-c');
+    queueSnapshotStore.getStoredQueueSnapshot.mockResolvedValue(storedSnapshot([itemA, itemB, itemC], itemA));
+
+    const snapshots: Snapshot[] = [];
+    renderProvider((snapshot) => snapshots.push(snapshot));
+
+    await waitFor(() => {
+      expect(capturedMutationDeps.current).not.toBeNull();
+      expect(snapshots.at(-1)?.state.queue.map((entry) => entry.uuid)).toEqual(['item-a', 'item-b', 'item-c']);
+    });
+    const deps = capturedMutationDeps.current;
+    if (!deps) throw new Error('mutation deps were not captured');
+
+    expect(deps.getQueuePosition('item-a')).toBe(0);
+    expect(deps.getQueuePosition('item-c')).toBe(2);
+    expect(deps.getQueuePosition('never-queued')).toBe(-1);
+
+    // Reads the LIVE reducer state, not a mount-time snapshot: activating a new
+    // climb slots it in after the current one, and the index follows.
+    act(() => {
+      snapshots.at(-1)?.setCurrentClimb(makeQueueItem('item-x'));
+    });
+    await waitFor(() => {
+      expect(snapshots.at(-1)?.state.queue.map((entry) => entry.uuid)).toEqual([
+        'item-a',
+        'item-x',
+        'item-b',
+        'item-c',
+      ]);
+    });
+    expect(deps.getQueuePosition('item-x')).toBe(1);
+    expect(deps.getQueuePosition('item-c')).toBe(3);
   });
 
   it('persists the solo queue after mutations (debounced)', async () => {
--- a/packages/mobile/src/providers/queue-provider.tsx
+++ b/packages/mobile/src/providers/queue-provider.tsx
@@ -424,6 +424,11 @@ export function QueueProvider({ children }: { children: ReactNode }) {
     // created_at) makes the server reject the mutation and silently breaks queue
     // sync to peers. See toClimbInput.
     toQueueItemInput: (item) => ({ uuid: item.uuid, climb: toClimbInput(item.climb) }),
+    // Where a deferred queue-add (a superseded or drained-then-throttled
+    // activation) should land, so peers see the order this device shows. Read
+    // live off `stateRef` — assigned during render — because the send can fire
+    // seconds after the activation. Same read `recoverThrottledQueueAdd` uses.
+    getQueuePosition: (uuid) => stateRef.current.queue.findIndex((queueItem) => queueItem.uuid === uuid),
     ensureReady: async (capturedSessionId) => {
       if (!capturedSessionId) return null;
       await ensureJoined(capturedSessionId);
--- a/packages/shared/queue-react/src/__tests__/create-queue-mutations.test.ts
+++ b/packages/shared/queue-react/src/__tests__/create-queue-mutations.test.ts
@@ -28,11 +28,18 @@ const toQueueItemInput = (it: TestItem) => ({ uuid: it.uuid, climb: it.climb }) 
 const client = {} as never; // execute is mocked, so the client value is opaque.
 const flush = () => new Promise((resolve) => setTimeout(resolve, 0));
 
+// Stands in for the client's local reducer queue. `getQueuePosition` is wired to
+// a real findIndex over it, so an expected `position` comes from seeded data
+// rather than a stub's canned return. Tests mutate it to model reorders, removes
+// and wholesale server syncs mid-flight.
+let localQueue: TestItem[] = [];
+
 function make(overrides: Partial<QueueMutationsDeps<TestItem>> = {}) {
   return createQueueMutations<TestItem>({
     getClient: () => client,
     getSessionId: () => 'S',
     toQueueItemInput,
+    getQueuePosition: (uuid) => localQueue.findIndex((queueItem) => queueItem.uuid === uuid),
     ...overrides,
   });
 }
@@ -42,6 +49,7 @@ const queriesFor = (query: string) => executeMock.mock.calls.filter(([, op]) => 
 beforeEach(() => {
   executeMock.mockReset();
   executeMock.mockResolvedValue(undefined);
+  localQueue = [];
 });
 
 describe('web mode (no ensureReady)', () => {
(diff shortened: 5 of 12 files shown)


marcodejongh:
Thanks — taking the three notes in order.

**1. `addCandidateUuids` is not cleared between `setQueue` calls.** Correct that it isn't, but the re-mark it
produces is the right answer, not a stale one. In your example the climber's second replace (`setQueue([B])`)
genuinely discards A: A is gone from the local queue *because they replaced it*, which is exactly the
"climber dropped it" case the ledger exists to record. Whether A was also present at the previous replace
doesn't change that. Clearing the candidate list on each `setQueue` would be strictly worse — the second
replace would then have no record of A and its deferred add would take the append branch, putting A back on
the queue the climber just replaced.

The only cost is a `removedUuids` entry that no live deferred add will ever consult, bounded by
`ADD_CANDIDATE_MEMORY` (32) per replace. Leaving it as-is.

**2. Peer removal.** Agreed, and it's in the coverage table in the PR body, at the guard in
`create-queue-mutations.ts`, and now in the doc. Not fixable at this layer without a delta-origin signal.

**3. `docs/websocket-implementation.md`.** Fair — done in 5cb2e88, as a new
"Deferred queue-adds from the setCurrentClimb coalescer" subsection under Queue Mutations: the two loss paths,
the live-index rule, the four-row cause split with the peer-removal gap called out, and why skipping on mere
absence would make the recovery inert.

PR #4022, opened 2026-07-31
fix(mobile): let screen readers activate queue and climb rows

### LINKED ISSUE #3914: RNGH gesture rows drop screen-reader activation (QueueItemRow + ClimbListRow)
## What

Both `QueueItemRow` and `ClimbListRow` render their tappable row as a bare `Animated.View` / `View` with `accessible` + `accessibilityRole="button"`, wrapped in an RNGH `GestureDetector` running `Gesture.Tap()` / `Gesture.LongPress()`. Neither view has an `onAccessibilityTap` (iOS) or an `accessibilityActions` + `onAccessibilityAction` `activate` handler (Android).

RNGH gestures respond to the raw pointer/touch stream. VoiceOver and TalkBack "activate" (the double-tap) dispatch an accessibility action, not that pointer stream. A plain RN `Pressable`/`Touchable` wires this up internally (activation → `onPress`); a bare `View` + `GestureDetector` does not. So a screen-reader user hears "button" on a queue/climb row but activating it may do nothing — can't open the climb, and in edit mode can't toggle selection.

Flagged by Codex on #3890, which moved `QueueItemRow` onto the RNGH-gesture-row pattern that `ClimbListRow` already ships.

## Why this is its own issue (not folded into #3890)

- It's a **pre-existing, app-wide** characteristic of the already-shipped `ClimbListRow` pattern, not something #3683 / #3890 introduced in a novel way. Fixing only the queue row would leave the two row components inconsistent.
- There's real **uncertainty about whether RNGH's `Gesture.Tap` already handles the accessibility `activate` action**. If it does, blindly adding `onAccessibilityTap` / `onAccessibilityAction` would **double-fire** the handler under a screen reader. This has to be verified on-device with VoiceOver (iOS) and TalkBack (Android) before wiring anything — it can't be settled in jsdom/CI.

## Scope

- `packages/mobile/src/components/QueueItemRow.tsx`
- `packages/mobile/src/components/ClimbListRow.tsx`
- Any other bare-`View` + `GestureDetector` "button" rows that moved off `Pressable`.

## Acceptance

- On-device: VoiceOver double-tap (iOS) and TalkBack double-tap (Android) on a queue row and a climb-list row activate it (open / select / edit-mode toggle), verified to fire **exactly once** (no double-fire).
- Both row components handled consistently.

marcodejongh:
Fixes #3914

> **Device-QA gate — stays a draft until it passes.** This is an accessibility change whose real behaviour lives in VoiceOver and TalkBack, not in CI. See [QA Notes](#qa-notes) for the exact checks.

## What and why

Queue rows and climb-list rows are tappable, but their press does not live on an RN `Pressable` — it lives on a `react-native-gesture-handler` `Gesture.Tap()`. A VoiceOver or TalkBack user could focus the row, hear "Crimp Master, position 1, button", double-tap, and nothing happened. Same for the queue's tick button and the climb row's ⋮ button. Every one of those affordances was a dead end for a screen-reader user.

## Verified root cause

RNGH's `Gesture` API builds native gesture recognizers that never register with React Native's accessibility-action bridge. Checked against `react-native-gesture-handler@2.32.0`: neither the `Gesture` builders nor RNGH's own `Touchable` / `GenericTouchable` / `BaseButton` set `onAccessibilityTap` or `accessibilityActions`. RN core's `Pressable` is exempt because it rides the standard touch-responder chain, which the OS activates for free. Anything wrapped in a `GestureDetector` has to wire the accessibility channel itself.

There is a second layer, found in review. Both row containers are marked `accessible`. UIKit treats an accessibility element as a leaf, so VoiceOver never traverses into it — wiring `onAccessibilityTap` on the _nested_ tick/⋮ views can never fire on iOS, because focus never lands there. TalkBack does still focus inside the container (RN Android does not set `no-hide-descendants` automatically), so the nested props are load-bearing on Android only.

## The fix

`packages/mobile/src/components/QueueItemRow.tsx` and `ClimbListRow.tsx`, with the platform-gated action lists shared from `packages/mobile/src/lib/row-accessibility-actions.ts`:

- Row containers get `onAccessibilityTap` plus an `onAccessibilityAction` handler that branches on `actionName`, both routed to the exact handlers the RNGH taps already call. No press handler was touched — only new callers were added.
- The secondary affordance is published as a **labelled custom action on the row itself**, which is the element VoiceOver actually focuses: `logAscent` (label `mobile.queue.logAscent`) on a history queue row, `moreActions` (label `mobile.climbRow.moreActions`) on a climb row that shows the ⋮ button. Both keys already exist in `en-US`, `es` and `fr`, so no catalog churn.
- The nested tick / ⋮ views keep their own `onAccessibilityTap` + action, which is the working route on Android.
- The bare `activate` entry is **Android-only**. On Android `'activate'` maps to `ACTION_CLICK` in `ReactAccessibilityDelegate` and is the only route, since `onAccessibilityTap` is not implemented there. On iOS it is redundant (`accessibilityActivate` already routes the double-tap to `onAccessibilityTap`) _and_ harmful: every `accessibilityActions` entry becomes a `UIAccessibilityCustomAction` announced by its raw `name` when it carries no label, so VoiceOver would read out a developer-facing "activate" on every row of every list.
- No double-fire on either platform: iOS Fabric's `accessibilityActivate` calls `onAccessibilityTap` only; iOS Paper prefers the action and short-circuits; Android dispatches only the action event and returns `true` without calling `super`. Exactly once everywhere.
- The queue drag handle is deliberately left alone. Reordering with a screen reader needs worded move-up / move-down custom actions, which is product copy, not a guess. Same for the climb row's long-press-only actions path when `showMoreButton` is false. Both are commented in place as follow-ups.

Perf: the row action list is a `useMemo` keyed on the resolved **label string** rather than on `t` — `react-i18next` hands back a fresh `t` identity on plenty of renders, which would rebuild the array every time and churn the row element's props. The Android-only `activate` array is module-level and never rebuilt. `React.memo` prop equality on both rows is untouched.

## Tests

`packages/mobile/src/components/__tests__/queue-item-row-memo.test.tsx` (extended) and `climb-list-row-accessibility.test.tsx` (new) — 23 tests, all green. They capture the props the component actually renders and invoke them, rather than rebuilding a predicate.

Revert check (`git checkout origin/main -- QueueItemRow.tsx ClimbListRow.tsx`, tests untouched): **11 of the 12 new tests fail**, with real prop-read failures (`expected "undefined" to be "function"`, `expected undefined to be defined`). The twelfth is a negative guard ("offers no tick action on a row without a tick button") that carries a `toBeTypeOf('function')` precondition so it cannot pass vacuously — it fails on revert too via that precondition; the count above is from the run before that precondition was added.

What each covers:

| Test                                                                                       | Fails on revert because                                        |
| ------------------------------------------------------------------------------------------ | -------------------------------------------------------------- |
| activates the row from a screen-reader tap and from the activate action (both files)       | `onAccessibilityTap` / `onAccessibilityAction` are `undefined` |
| toggles selection from a screen-reader activation while in edit mode                       | activation never reaches `onToggleSelect`                      |
| logs an ascent from a screen-reader activation of the tick button                          | nested tick props are `undefined`                              |
| publishes the tick / ⋮ as a labelled custom action on the row itself                       | row `accessibilityActions` is `undefined`                      |
| opens the actions menu from a screen-reader activation of the ⋮ button                     | nested ⋮ props are `undefined`                                 |
| leaves an unsupported row inert                                                            | preconditions assert the channel exists first                  |
| keeps one accessibilityActions array across rerenders                                      | array is `undefined`                                           |
| adds the activate action on Android (re-imports each module with `Platform.OS = 'android'`; one per component) | pins the platform gate in both directions |

Also asserted: a foreign action name (`magicTap`) is a no-op on every handler, and activating a nested button never also presses the row.

Validation: `vp run typecheck:mobile` pass, `vp check` pass (0 errors; the 289 warnings are pre-existing and none reference these four files), scoped `vp test run` 23/23.

<a name="qa-notes"></a>

## QA Notes

CI cannot verify this. It needs a real device pass on both platforms.

**iOS + VoiceOver**

1. Play drawer → queue list. Swipe to a queue row. Confirm it reads `"<climb>, position N, button"` and a **double-tap makes that climb current**.
2. Confirm VoiceOver does **not** announce a stray, untranslated "activate" custom action on any row. (This is the regression the Android gate exists to prevent.)
3. Queue history rows: with the row focused, rotor → Actions. Confirm a **"Log ascent"** action is offered and firing it opens the log-ascent flow — and that it does **not** also make the climb current or dismiss the sheet.
4. Climb list (search results): confirm double-tap opens the climb, and that a row with the ⋮ button offers a **"More actions"** custom action that opens the reaction menu.
5. Edit mode in the queue: confirm double-tap **toggles selection** instead of playing the climb.
6. Worth reporting either way: whether VoiceOver can focus the nested tick / ⋮ views at all. The prediction is no — that is exactly why the row-level custom actions exist. If it turns out it _can_, the nested wiring is a live second route and should be re-checked for double-announcement.

**Android + TalkBack**

1. Same row activation via double-tap on queue and climb rows.
2. Confirm the nested tick and ⋮ buttons **can** be focused and activated directly (this is the Android-only `activate`/`ACTION_CLICK` route).
3. Confirm the row's "Log ascent" / "More actions" custom actions also work from the row's action menu, and that neither route fires twice.

**Known gaps, out of scope here:** the queue drag handle has no reorder actions; a climb row with `showMoreButton` false still hides its long-press menu from screen readers; `LogbookRow.tsx` has the identical bare-`View`-inside-`GestureDetector` shape with no activation wiring. All three want their own product copy and are follow-ups.

## Release Notes

Queue rows and climb rows now respond to VoiceOver and TalkBack. Double-tap a row to play the climb or select it in edit mode, and reach "Log ascent" on a history row or "More actions" on a climb row straight from the row's actions — no more tapping into dead ends.


marcodejongh:
CI is green, but **this stays a draft until it passes on real hardware.** Nothing in the pipeline can tell you whether a screen reader can actually reach these rows — that verdict only comes from VoiceOver and TalkBack on a device.

The specific unknown worth watching: both row containers are marked `accessible`, and UIKit treats an accessibility element as a leaf. The prediction is that VoiceOver **cannot** focus the nested tick / ⋮ views at all, which is exactly why this PR also publishes them as labelled custom actions on the row itself. If a device pass shows VoiceOver *can* focus them, the nested wiring is a live second route and wants a re-check for double announcement.

Full step-by-step for both platforms is in the QA Notes section of the description. The short version:

- **iOS / VoiceOver** — double-tap a queue row plays the climb; double-tap in edit mode toggles selection instead; rotor → Actions offers "Log ascent" on a history row and "More actions" on a climb row; **no** stray untranslated "activate" action is announced anywhere.
- **Android / TalkBack** — same row activation, plus the nested tick and ⋮ buttons are directly focusable and activatable, and neither route fires twice.

There is an OTA preview channel on this PR (`pr-4022`) if that helps get it onto a device. Heads up that the OTA publish job flaked three times on server-side errors from `updates.boardsesh.com` (Cloudflare 524s while marking the update finished, plus one S3 `Request body is not seekable`) before going green — unrelated to this diff, but worth knowing if the preview looks stale.

PR #4026, opened 2026-07-31
fix(mobile): make the logbook Filter chip icon-only on iOS

### LINKED ISSUE #3782: [Bug: UI] Profile: Logbook: Filter chip label is truncated
Image

It should either say "Filter", or just be turned into an icon in a circle.

marcodejongh:
## What & why

On the mobile app's Profile → Logbook tab, the iOS Liquid Glass toolbar's leading **Filter** chip rendered as `F…` instead of `Filter` ([#3782](https://github.com/boardsesh/boardsesh/issues/3782)) — the plain-text chips beside it (`Latest`, `Hardest`, `Grade`) laid out fine in the same row.

## Root cause

`packages/mobile/src/components/you/LogbookChipRow.ios.tsx` renders the chip as `<Button label systemImage="line.3.horizontal.decrease" />`. @expo/ui's `Button.swift` maps that to `SwiftUI.Button(title, systemImage:)`, which builds a `Label`. Inside the glass `HStack` (small `controlSize`, horizontal `ScrollView`, `Host matchContents={{ vertical: true }}`), the symbol + one-word title combination appears to be what SwiftUI compresses — the search row's `FilterChipRow.ios.tsx` uses the same symbol with a longer title (`Filters · N`) and is unaffected. Not a string-length problem: `Filter` / `Filtrar` / `Filtrer` are all short.

## Fix

Add `labelStyle('iconOnly')` to the Filter chip's modifiers so the title leaves layout entirely and the chip sizes to the SF Symbol alone. This is the same icon-only filter affordance the Android (Material) logbook toolbar already ships in `LogbookTab.tsx`, and the issue author explicitly sanctioned going icon-only.

The title is still passed — @expo/ui only renders `systemImage` when a `label` is present — and `.labelStyle(.iconOnly)` keeps it as the VoiceOver name. An explicit `accessibilityLabel(t('mobile.logbook.filter'))` states that so the affordance can't silently lose its name. No new i18n keys: the existing `mobile.logbook.filter` key (en-US / es / fr) is reused.

Pure JS/TSX, OTA-safe, iOS-only. No native rebuild, no new dependency.

## Tests

New `packages/mobile/src/components/you/__tests__/logbook-chip-row-filter-icon.test.tsx` (jsdom). The extensionless `LogbookChipRow` specifier is aliased to a null stub in `packages/mobile/vite.config.ts`, so the suite imports the component by its **explicit `.ios` path** and mocks `@expo/ui/swift-ui` + `.../modifiers`, recording every modifier by name and argument. It asserts the chip carries `labelStyle('iconOnly')` + `accessibilityLabel('mobile.logbook.filter')` alongside the neutral-glass styling, that the same pair survives the longer amber (`glassProminent`) modifier list once a grade bound is set, and that the sibling text chips stay plain (no `systemImage`, no `labelStyle`).

**Fail-on-revert:** reverting `LogbookChipRow.ios.tsx` to `origin/main` and re-running the suite fails 4 of 5 tests (missing `labelStyle`, missing `accessibilityLabel`, and both modifier-list mismatches). Restoring it passes 5/5.

`scripts/mobile-platform-imports-check.sh` gained one narrow exemption: a vitest file under `__tests__/` may name `@expo/ui/swift-ui` because `vi.mock()` needs a literal specifier. Test modules run in node and Metro never reaches them from an app entry, so they can't cause the "Unable to get view config" crash the guard exists to stop. Every module Metro can reach is still scanned, and the exemption only matches `__tests__/<file>.test.ts(x)` — a production module can't dodge the guard by moving into a test folder.

## QA Notes

- **iOS visual confirmation is pending** — this was implemented on a Linux box with no simulator. SwiftUI layout can't be asserted in jsdom, so the on-device check is: Profile → Logbook on an iOS 26 (Liquid Glass) build, confirm the leading chip is a round icon-only filter button with no `F…` text, that it still opens the filter sheet, that it turns amber once a facet is set, and that VoiceOver reads it as "Filter".
- Risk is bounded: Android already ships this exact icon-only shape for the same control, and the issue author asked for it as one of two acceptable fixes.
- Unchanged surfaces: the Android and web logbook chip rows, and the climbs-search `FilterChipRow.ios.tsx`.

## Release Notes

The Logbook's filter button on iOS now shows a clean filter icon instead of a chopped-off "F…".

Fixes #3782

### DIFF
--- a/packages/mobile/src/components/you/LogbookChipRow.ios.tsx
+++ b/packages/mobile/src/components/you/LogbookChipRow.ios.tsx
@@ -28,6 +28,8 @@ import {
   foregroundColor,
   padding,
   menuActionDismissBehavior,
+  labelStyle,
+  accessibilityLabel,
 } from '@expo/ui/swift-ui/modifiers';
 import { useTheme } from '../../providers/theme-provider';
 import { useGradeFormat } from '../../hooks/use-grade-format';
@@ -104,6 +106,21 @@ function LogbookChipRowComponent({
   const handleAngleChip = useCallback(() => onToggleFacet('angle'), [onToggleFacet]);
   const handleDateChip = useCallback(() => onToggleFacet('date'), [onToggleFacet]);
 
+  // The Filter chip is icon-only. `SwiftUI.Button(title, systemImage:)` builds a
+  // `Label`, and inside the glass HStack the icon+one-word title combination made
+  // SwiftUI compress the title to "F…" (#3782) — the plain-text chips beside it are
+  // fine. `labelStyle(.iconOnly)` drops the title from layout so the chip sizes to
+  // the symbol alone, matching the round icon-only filter button the Android
+  // (Material) logbook toolbar already ships. The title still has to be passed —
+  // @expo/ui only renders `systemImage` when a `label` is present — and iconOnly
+  // keeps it as the VoiceOver name; `accessibilityLabel` states that explicitly so
+  // the affordance can't silently lose its name.
+  const filterLabel = t('mobile.logbook.filter');
+  const filterModifiers = useMemo(
+    () => [...chipModifiers(anyFilterActive(facets)), labelStyle('iconOnly'), accessibilityLabel(filterLabel)],
+    [chipModifiers, facets, filterLabel],
+  );
+
   const handleSelectLatest = useCallback(() => onSelectPreset('recent'), [onSelectPreset]);
   const handleSelectHardest = useCallback(() => onSelectPreset('hardest'), [onSelectPreset]);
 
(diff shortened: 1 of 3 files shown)


marcodejongh:
Applied review point 1: the suite now finds the Filter chip by its label and asserts the leading position separately, so a chip reorder fails on the position assertion instead of silently testing whatever moved into slot 0.

Not applying point 2. `CLAUDE.md` has no one-line comment rule — the repo convention in this exact directory is the opposite (see the 18-line header on `LogbookChipRow.ios.tsx` and the block comments on `chipModifiers` / the Show menu, all on main). The block explains why `labelStyle(iconOnly)` is load-bearing and why the title still has to be passed, which is the thing a future editor would otherwise delete; a PR description is not visible from the file.

iOS 26 device verification is still the gate — noted under QA Notes.

Issue #3779, opened 2026-07-19
[Bug: UI] Profile: Flash vs Redpoint bar height numbers are clipped


ES-Alexander:
This makes them harder to see and read.

<img width="750" height="488" alt="Image" src="https://github.com/user-attachments/assets/a9544787-1633-4c08-b681-bc41b2307d7f" />

It may help to turn them vertically, or somehow draw them in a bigger box that can overflow from the column width a little. Making them smaller could reduce clipping, but would likely still make them hard to read.

PR #4029, opened 2026-07-31
fix(mobile): keep You-page charts inside their card and off the spline

### LINKED ISSUE #3778: [Bug: UI] Profile: Activity / Grade Distribution plots can overflow card width
Image

### LINKED ISSUE #3780: [Bug: UI] Profile: V-Points line can overshoot data points
The line interpolation is currently using an algorithm that can overshoot data, which can make it seem like a given week was better or worse than it actually was:

Image

This isn't a huge deal, but it may be preferable to use a monotone interpolation algorithm instead (e.g. constrain the derivative to 0 at each point), especially since the data is guaranteed to be evenly spaced along the time axis (so the curves are likely to look reasonable).

marcodejongh:
## What & why

Two separate You-page chart bugs, both in `packages/mobile/src/components/you/YouCharts.tsx`.

**#3778 — Activity and Grade Distribution bars bleed past the card edge.**
`react-native-gifted-charts` does not lay its chart out inside the `width` prop it
is handed. `BarAndLineChartsWrapper` positions the plot `ScrollView` absolutely at
`marginLeft: yAxisLabelWidth + yAxisThickness` with `width: props.width`, and
`renderHorizSections` draws the rules/x-axis row as a `yAxisLabelWidth` spacer
followed by a `props.width + endSpacing` band. The span the chart really occupies
is `yAxisLabelWidth + width + endSpacing`.

Both You-page bar charts gave BarChart the measured frame width minus a token 8px
*and* asked for a y-axis gutter, so they occupied more than the card had:

| Chart | y-axis gutter | width | endSpacing | occupied |
| `StackedBarChart` with `showYAxisScale` (ProgressTab Activity + Grade Distribution) | 32 | frame − 8 | 8 (defaults to `spacing`) | **frame + 32** |
| `GroupedBarChart` | 10 (`hideYAxisText` still reserves `yAxisEmptyLabelWidth`) | frame − 8 | 20 (defaults to `spacing`, unset here because spacing is per-datum) | **frame + 22** |

Nothing clipped it — neither `styles.frame` nor the `Card` set `overflow`, so the
overflow landed outside the rounded card exactly as screenshotted in the issue.

The four other `StackedBarChart` consumers (`SessionGradeChart`,
`SessionAnalyticsSection`, `record/summary.tsx`, `WorkoutTypeShelf`) don't pass
`showYAxisScale`, so their gutter was already 0 — which matches the issue naming
only Activity and Grade Distribution. `TotalAreaChart` already budgets correctly
(`width - 48` against a 35px default gutter), the internal proof that the gutter
is additive.

**#3780 — the V-Points line overshoots the data.** `TotalAreaChart` drew the
cumulative line with gifted-charts' `curved` / `curvature={0.2}` interpolation.
`gifted-charts-core`'s `bezierCommand`/`controlPoint` is an unclamped
Catmull-Rom spline, so it swings past a point's y-value between two neighbours —
a flat or dipping week reads as a rise. Weeks are evenly spaced already.

## Verified root cause

Read directly against the installed libraries (`react-native-gifted-charts`
1.4.77 / `gifted-charts-core` 0.1.81), not inferred:

- `react-native-gifted-charts/dist/Components/BarAndLineChartsWrapper/index.js` — plot ScrollView `position: 'absolute'`, `marginLeft: yAxisLabelWidth + yAxisThickness`, `width: props.width`.
- `.../renderHorizSections.js` — a `width: yAxisLabelWidth` spacer followed by a `width: (props.width ?? totalWidth) + endSpacing` rules/x-axis row.
- `gifted-charts-core/dist/BarChart/index.js:100` + `dist/utils/constants.js:86` — `hideYAxisText` falls back to `yAxisEmptyLabelWidth = 10`; `endSpacing` falls back to `spacing` (default 20).
- `gifted-charts-core/dist/utils/index.js:131-153` — `bezierCommand`/`controlPoint`, unclamped Catmull-Rom.

## Fix

- Subtract the y-axis gutter from the width budget *before* it reaches either `fitBars` or the `width` prop, so the fitting math and the canvas agree and the whole thing fits the frame.
- Pin `endSpacing={0}` on both charts.
- Pin `yAxisLabelWidth` explicitly on both (`axisGutter` on stacked, `0` on grouped) so the library defaults can't creep back in a version bump.
- `overflow: 'hidden'` on `styles.frame` as a second layer. The reset-zoom button lives inside the frame (`top: 8`/`right: 8`) and the tooltip overlay is a sibling on `chartShell`, so neither is affected.
- `curved={false}` on `TotalAreaChart`, with `curvature` dropped.

## Tests & fail-on-revert

There's no visual-diff harness for these RN charts, so both tests pin the prop
contract, mocking `react-native-gifted-charts` and asserting on the props BarChart
is actually handed.

`you-charts-width-budget.test.tsx` (6 assertions) measures the **full occupied
span** — `yAxisLabelWidth + width + endSpacing`, with gifted-charts-core 0.1.81
defaults filled in for omitted props — against the measured frame, plus the
internal bar/spacing span against the `width` prop.

`you-charts-line-interpolation.test.tsx` asserts `curved={false}` with no leftover
`curvature`.

Fail-on-revert, verified by restoring `YouCharts.tsx` to `origin/main` and
re-running: **6 of 7 tests go red**, including both headline occupied-span
assertions — `expected 352 to be less than or equal to 320` (stacked, 32 + 312 + 8)
and `expected 342 to be less than or equal to 320` (grouped, 10 + 312 + 20) — and
`expected true to be false` for `curved`. The file was then restored and confirmed
byte-identical.

The first pass at #3778 shipped a test whose grouped assertion (`width === frame - 8`)
was a tautology: it passed identically before and after the change. That's been
replaced with the occupied-span assertions above.

## QA Notes

Scoped validation, all from the worktree:

- `vp test run --reporter=agent --project mobile packages/mobile/src/components/you/__tests__/` — 27 files / 200 tests green.
- Adjacent `StackedBarChart`/`GroupedBarChart` consumers (`SessionAnalyticsSection`, `record/summary-error-state`, `workout-type-shelf`) — 3 files / 11 tests green.
- `vp check packages/mobile/src/components/you/` — 0 errors (1 pre-existing warning in an untouched file, `logbook-tab-grouped.test.tsx:207`).

To exercise by hand: **You → Progress tab**.

- Activity and Grade Distribution charts: bars and the last x-axis label must sit inside the rounded card, with the y-axis scale still legible on the left. Check light and dark, and a narrow device (iPhone SE) where the budget is tightest.
- Pinch-zoom the same charts to ~2.75x and back. Bars should still be clipped by the plot ScrollView, never drawn over the card edge, and the reset-zoom button must remain visible and tappable in the frame's top-right (it sits inside the new `overflow: 'hidden'`).
- Flash vs Redpoint (grouped) chart: same edge check, plus the per-bar top value labels must not be clipped at the top of the frame.
- V-Points chart: on a week where the cumulative total is flat, the line must be flat between the two points rather than bulging above them.
- Sanity-check the unaffected charts still render normally: in-session grade chart, session analytics, record summary, pre-session workout-type shelf.

Not visually verified on a device or emulator in this pass — the change is
geometry-only and pinned by prop assertions, but a before/after capture of the
Progress tab is worth attaching before this leaves draft.

## Release Notes

The Activity and Grade Distribution charts on your You page no longer spill past
the edge of their card, so the last week or grade is readable again. The V-Points
line now tracks your actual weekly totals instead of curving above them and making
a flat week look like a jump.

Fixes #3778
Fixes #3780

### DIFF
--- a/packages/mobile/src/components/you/YouCharts.tsx
+++ b/packages/mobile/src/components/you/YouCharts.tsx
@@ -24,6 +24,13 @@ const AXIS_LABEL_SIZE = 11;
 const STACK_BAR_RADIUS = 3;
 const MIN_ZOOM_SCALE = 1;
 const MAX_ZOOM_SCALE = 2.75;
+// gifted-charts' BarChart renders its bars/spacing into whatever `width` prop
+// it's given, but that width is itself a few px narrower than the measured
+// frame (room for the chart's own right-edge padding). Bar-fitting math MUST
+// use the same narrowed budget the chart is actually given, or the fitted
+// bars (sized off the full frame width) run wider than the canvas they're
+// drawn into and the last bar/label spills past the card edge (#3778).
+const CHART_WIDTH_INSET = 8;
 // Floor for grouped flash|redpoint bars so a dense V0-V17 axis keeps each bar
 // above the iOS 44pt / Android 48dp touch target instead of shrinking to 4px.
 const MIN_GROUPED_BAR = 6;
@@ -375,7 +382,8 @@ export const StackedBarChart = memo(function StackedBarChart({
           zoomable={interactive && zoomable}
         >
           {(width, zoomScale, scrollEnabled) => {
-            const fitted = fitBars(width, stackData.length, minBarWidth);
+            const chartWidth = width - CHART_WIDTH_INSET;
+            const fitted = fitBars(chartWidth, stackData.length, minBarWidth);
             const barWidth = Math.max(minBarWidth, Math.round(fitted.barWidth * zoomScale));
             const spacing = Math.max(2, Math.round(fitted.spacing * zoomScale));
             // gifted sizes each x-axis label box to ~one bar by default, so a wide
@@ -390,7 +398,7 @@ export const StackedBarChart = memo(function StackedBarChart({
             return (
               <BarChart
                 stackData={sizedData}
-                width={width - 8}
+                width={chartWidth}
                 height={height - 28}
                 barWidth={barWidth}
                 spacing={spacing}
@@ -487,11 +495,14 @@ export const GroupedBarChart = memo(function GroupedBarChart({
           zoomable={interactive && zoomable}
         >
           {(width, zoomScale, scrollEnabled) => {
+            const chartWidth = width - CHART_WIDTH_INSET;
             const list = bars ?? [];
             const initial = 8;
             const baseBarWidth = Math.max(
               MIN_GROUPED_BAR,
-              Math.floor((width - initial * 2 - groupGap * list.length - innerGap * list.length) / (list.length * 2)),
+              Math.floor(
+                (chartWidth - initial * 2 - groupGap * list.length - innerGap * list.length) / (list.length * 2),
+              ),
             );
             const barWidth = Math.max(MIN_GROUPED_BAR, Math.round(baseBarWidth * zoomScale));
             const zoomedGroupGap = Math.max(groupGap, Math.round(groupGap * zoomScale));
@@ -517,7 +528,7 @@ export const GroupedBarChart = memo(function GroupedBarChart({
             return (
               <BarChart
                 data={data}
-                width={width - 8}
+                width={chartWidth}
                 height={height - 28}
                 barWidth={barWidth}
                 initialSpacing={initial}
--- a/packages/mobile/src/components/you/YouCharts.tsx
+++ b/packages/mobile/src/components/you/YouCharts.tsx
@@ -24,13 +24,21 @@ const AXIS_LABEL_SIZE = 11;
 const STACK_BAR_RADIUS = 3;
 const MIN_ZOOM_SCALE = 1;
 const MAX_ZOOM_SCALE = 2.75;
-// gifted-charts' BarChart renders its bars/spacing into whatever `width` prop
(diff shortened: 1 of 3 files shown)


marcodejongh:
🤖 Bug-batch loop `josh` is taking over this now-conflicted PR for batch 1. I’ll rebase it onto current `main`, independently review the chart fix, address the outstanding test/QA notes, verify whether #3050 is the same overflow defect, and keep monitoring the PR through green CI and resolved feedback.

marcodejongh:
🤖 Bug-batch loop `josh` coordination note: #3778, #3050, and #3780 are now claimed by the batch loop, and this draft has gone stale (conflicting with main). Superseding it with two scoped PRs that port its validated content: the sizing fix + width-budget test → a PR closing #3778/#3050, and the curvature fix + its regression test → a PR closing #3780. I'll close this draft once both replacements are green. @marcodejongh — shout if you'd rather revive this PR instead.

marcodejongh:
Superseded as planned: the sizing fix + width-budget test live on in #4074 (closes #3778 + #3050, CI green), and the curvature fix + regression test in #4066 (closes #3780, CI green). Both port this PR's validated content with credit; #4074 additionally preserves the #3779 label-rotation logic this branch predates. Closing.

PR #4032, opened 2026-07-31
fix(mobile): treat a session ending as teardown, not a sync error

### LINKED ISSUE #3087: Racing party member sees a "Sync error" toast when someone ends the shared session

marcodejongh:
## Summary

When the leader ended a party session while a peer was still live, the peer's mobile app flashed a red "Sync error" toast and shipped a handled `GraphQLOperationError` to error tracking — for what is ordinary session teardown. Web has handled this silently since #2696; mobile never got the guard.

**Verified root cause.** The backend intentionally rejects a JOIN against an ended session with `new GraphQLError('This session has ended', { extensions: { code: 'SESSION_ENDED' } })` (`packages/backend/src/services/room-manager/client-lifecycle.ts:112-118` — the only emitter of that code in the repo). On mobile the rejection surfaces from `ensureJoined` inside `useSessionRealtime`'s join catch, which had no code discrimination and fell straight through to `showToast('mobile.queue.syncError', 'error')` + `reportHandledError(..., { source: 'queue-sync', op: 'join' })`. The chain was checked end to end rather than assumed: `execute()` wraps server errors into a real `GraphQLOperationError` whose constructor prefers the *coded* error when picking `.extensions`, and `createJoinSessionTracker.ensureJoined` rethrows the original error unmodified, so both the `instanceof` and the `extensions.code` read hold at the catch site.

The issue body pointed at `queue-provider.tsx`; the join catch has since moved to `packages/mobile/src/providers/queue/use-session-realtime.ts`.

**Fix.** In that catch, a `SESSION_ENDED`-coded `GraphQLOperationError` now clears the local session and shows the existing `mobile.toast.sessionEnded` success toast instead of the error toast, and skips the error report. It toasts (where web stays silent) on purpose: a client whose join fails here never receives the live `SessionEnded` subscription event, so this is the only signal the user gets that the session is gone. The branch sits after the existing backgrounded-join branch, so a rejection that lands while the app is backgrounded still defers and re-joins on foreground, hitting the new branch there.

No new i18n key — `mobile.toast.sessionEnded` already exists in `en-US`/`es`/`fr` and is already used elsewhere in this file.

## Test plan

New `packages/mobile/src/providers/__tests__/queue-provider-join-session-ended.test.tsx`:

1. A `JoinSession` mutation rejecting with `SESSION_ENDED` → session cleared, `mobile.toast.sessionEnded` / `success`, no `reportHandledError`.
2. Regression guard: a join rejection with **no** `extensions.code` still produces `mobile.queue.syncError` / `error` and still reports — so the guard can't quietly swallow genuine sync failures.

**Fail-on-revert story (executed, not asserted).** Reverse-applied the source-only hunk with `git apply -R` and re-ran the file: 1 failed / 1 passed. Case 1 failed exactly at the pre-fix behaviour — `showToast` received `["mobile.queue.syncError","error"]` instead of `["mobile.toast.sessionEnded","success"]` — while case 2 stayed green, proving it is not a tautology on the fixed branch. Restored; tree clean.

Scoped validation: the new file 2/2; the whole `packages/mobile/src/providers/__tests__/` directory 40 files / 452 tests pass; `vp run typecheck:mobile` clean; `vp check` clean.

## QA Notes

Needs two devices/clients in one party session.

- **Headline path**: peer joins a session, leader ends it while the peer is live and mid queue/join activity. Peer should get the green "session ended" toast and drop back to solo — no red "Sync error", no Sentry/PostHog `GraphQLOperationError "This session has ended"`.
- **Backgrounded peer**: background the peer, end the session from the leader, foreground the peer. Same green toast on foreground, not the error toast.
- **No double toast**: the peer should see the ended toast exactly once, whether the news arrives via the `SessionEnded` subscription event or via the failed re-join.
- **Genuine sync failure still shouts**: kill the backend (or go offline) with a live session — the red "Sync error" toast and retry behaviour must be unchanged.
- **Leader self-end**: leader ending their own session still toasts and clears as before.
- **A→B switch**: leave session A for session B while A is being ended — B must stay connected.

## Release Notes

- When someone ends a session you're in, you now get a plain "session ended" note instead of a red sync error.

## Follow-up (not in this PR)

Queue *mutations* reach the same backend error through `useQueueMutations`' `ensureReady` seam when the join cache is cold after a socket reconnect, and still land in `showQueueMutationErrorToast` as `mobile.queue.actionFailed` + a handled report. Lower volume (needs a reconnect to be reachable) and a different call site; the clean fix is the shared `isSessionEndedError` helper next to `isRateLimitedError` in `packages/shared/graphql-client/src/errors.ts`, used from both places. Left out to keep this diff on the headline path.

Fixes #3087

### DIFF
--- a/packages/mobile/src/providers/__tests__/queue-provider-join-session-ended.test.tsx
+++ b/packages/mobile/src/providers/__tests__/queue-provider-join-session-ended.test.tsx
@@ -1,0 +1,316 @@
+// @vitest-environment jsdom
+//
+// #3087: a party session that already ended server-side (leader ended it, or
+// it was reaped) while a peer is joining/rejoining used to be treated as a
+// generic sync failure — the JOIN_SESSION catch in `startJoinedSubscriptions`
+// unconditionally showed `mobile.queue.syncError` and reported a handled
+// error, even though `ensureJoined` rejecting with a `SESSION_ENDED`-coded
+// `GraphQLOperationError` is expected teardown, not a defect. Web already
+// special-cases this exact code (session-connection-ports.ts); this file
+// covers the analogous mobile guard: silently clear the stored session and
+// show the calm `mobile.toast.sessionEnded` toast instead, while a genuine
+// (non-coded) join failure still surfaces the noisy syncError path unchanged.
+// Reuses the WS/HTTP/store mock harness pattern from
+// queue-provider-backgrounded-join.test.tsx.
+import { render, waitFor } from '@testing-library/react';
+import { createElement, useEffect } from 'react';
+import { beforeEach, describe, expect, it, vi } from 'vitest';
+import type { UserBoard } from '@boardsesh/shared-schema';
+import { GraphQLOperationError } from '@boardsesh/graphql-client';
+
+const ws = vi.hoisted(() => {
+  type WsEventName = 'connected' | 'closed';
+  const listeners: Record<WsEventName, Set<() => void>> = {
+    connected: new Set(),
+    closed: new Set(),
+  };
+  return {
+    emit: (eventName: WsEventName) => {
+      for (const listener of listeners[eventName]) listener();
+    },
+    client: {
+      on: vi.fn((eventName: WsEventName, listener: () => void) => {
+        listeners[eventName].add(listener);
+        return () => {
+          listeners[eventName].delete(listener);
+        };
+      }),
+      subscribe: vi.fn(() => vi.fn()),
+    },
+    reset: () => {
+      listeners.connected.clear();
+      listeners.closed.clear();
+    },
+  };
+});
+
+const graph = vi.hoisted(() => ({
+  execute: vi.fn(),
+}));
+
+const http = vi.hoisted(() => ({
+  request: vi.fn(),
+}));
+
+const analytics = vi.hoisted(() => ({
+  track: vi.fn(),
+}));
+
+const sessionStore = vi.hoisted(() => ({
+  getStoredSessionId: vi.fn(async () => 'session-1' as string | null),
+  setStoredSessionId: vi.fn(async () => {}),
+  clearStoredSessionId: vi.fn(async () => {}),
+  getStoredCreatedSessionId: vi.fn(async () => null as string | null),
+  setStoredCreatedSessionId: vi.fn(async () => {}),
+  clearStoredCreatedSessionId: vi.fn(async () => {}),
+}));
+
+const queueSnapshotStore = vi.hoisted(() => ({
+  getStoredQueueSnapshot: vi.fn(async () => null),
+  setStoredQueueSnapshot: vi.fn(async () => {}),
+  clearStoredQueueSnapshot: vi.fn(async () => {}),
+}));
+
+const activeBoard = vi.hoisted(() => ({
+  stored: {
+    uuid: 'board-1',
+    slug: 'board-1',
+    ownerId: 'owner-1',
+    boardType: 'kilter',
+    layoutId: 1,
+    sizeId: 10,
+    setIds: '1,2',
+    name: 'Test board',
+    isPublic: true,
+    isUnlisted: false,
+    hideLocation: false,
+    isOwned: true,
+    angle: 40,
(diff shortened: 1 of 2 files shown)


marcodejongh:
Thanks — both points looked at, both declined, with reasons:

**Comment-length rule.** There is no such rule. `grep -rniE 'multi-line comment|docstring|comment block|one short line' CLAUDE.md .claude/` returns nothing — the quoted line isn't in the repo. This bot has invented this same rule on previous PRs. The comment stays: it records why this branch toasts where web stays silent, which is the one thing a future reader would otherwise "fix" back into a silent path.

**Backgrounded / A→B coverage.** Neither is new behaviour introduced here, so neither is what these tests are for. The backgrounded arm is the pre-existing `isBackgrounded()` branch that sits *above* the new guard and returns early — the new code is unreachable while backgrounded, and on foreground it is just the plain headline path already covered by case 1. The A→B guarantee comes from the guard at the top of the same catch block (`sessionIdRef.current !== sessionId` → return), which predates this PR; the redundant repeat of it inside the new branch was removed in 7abb855 precisely because it was dead. Both paths are on the QA checklist for device verification.

PR #4044, opened 2026-07-31
fix(sync): keep real rating dates on Kilter climb ratings

### LINKED ISSUE #3524: Upstream rating timestamps are discarded: board_climb_ratings.created_at is stamped at sync time and updated_at is rewritten every cycle, so 'when the user rated' is unrecoverable
## Summary
RawClimbRating carries the upstream created_at (user-sync.ts:155) but the values object built in applyClimbRatings (:865-879) never maps it, so created_at falls to the column default (now()) and every row's created_at is the first-sync date, not the rating date. The upsert's SET also stamps updatedAt: new Date() unconditionally (:906) even when nothing changed, so both timestamps are sync artifacts. Any recency-ordered or dated surface built on this table (rating history, 'rated N years ago') will be wrong for all synced rows.

## Failure scenario
A user rated a climb on Kilter in 2022. After linking, the board_climb_ratings row says created 2026-06-30 and updated today; any feature sorting or displaying rating dates shows every historic rating as brand new, re-dated on every sync cycle.

## Evidence
`packages/kilter-sync/src/sync/user-sync.ts:865`

```
Prod: min(created_at) over all 2,654 board_climb_ratings = 2026-06-30 (kilter user-sync launch week) even though the twin ticks for the same users date to 2022; max(updated_at) = today. Code: user-sync.ts:865-879 (no created_at/createdAt key in the values map), :906 (updatedAt: new Date() in the conflict SET).
```

## Suggested fix
Map raw.created_at into createdAt on insert (with the same ::timestamptz care as ticks), and only bump updatedAt when a field actually changed (add a WHERE clause to the DO UPDATE: `WHERE (board_climb_ratings.rating, …) IS DISTINCT FROM (EXCLUDED.rating, …)`).

marcodejongh:
## What / why

Every `board_climb_ratings` row written by the Kilter sync carried timestamps that described **our sync**, not the climber. `min(created_at)` in production sits at the day the Kilter sync started, so anything ordered or dated off this table ("rated 3 days ago", recency sorts) is wrong.

## Verified root cause

`packages/kilter-sync/src/sync/user-sync.ts`, `applyClimbRatings`:

1. **`created_at` was never mapped.** The per-PUT values object maps every `RawClimbRating` field except `raw.created_at`, so the insert fell through to the column default (`createdAt: timestamp('created_at').defaultNow().notNull()`, `packages/db/src/schema/boards/unified.ts`). Every historical rating collapsed onto first-sight day.
2. **`updated_at` was bumped unconditionally.** The `onConflictDoUpdate` SET clause contained `updatedAt: new Date()` with no change guard. PowerSync redelivers a full snapshot every cycle, so the DO UPDATE fired for every rating on every sync and rewrote `updated_at` even when nothing about the rating had changed. The column tracked "when the sync last ran".

## Fix

- Map `createdAt` from `raw.created_at` via a new `parseKilterTimestamp` helper. Every payload we've observed is a `Z`-suffixed UTC ISO string; a value carrying a `Z` or a `±HH:MM` offset is normalised to that instant, and a zone-less wall clock (`"2024-03-05 18:30:00"`) is read as **UTC** rather than handed to bare `Date.parse`, which applies the host process's timezone and would skew the stored date by the container's offset. Reading it as UTC matches `applyLogs`, which writes the upstream string straight into a `mode: 'string'` column where Postgres keeps the wall clock verbatim. An unparseable value returns `undefined` so the insert falls back to the column default rather than writing an Invalid Date, which Postgres rejects and which would abort the whole ratings batch for that user.
- Add a `setWhere` to the conflict clause: a row-wise `IS DISTINCT FROM` over `(rating, difficulty_grade_id, comment, kilter_id)`. The right-hand side is the **effective** new value, not bare `EXCLUDED` — `comment` and `kilter_id` are written through `COALESCE(EXCLUDED.…, existing)`, so comparing raw `EXCLUDED` would count a no-op COALESCE fallback as a change and bump `updated_at` anyway. Row-wise `IS DISTINCT FROM` also handles the nullable columns without NULL-propagating to unknown.
- `created_at` is deliberately absent from both the SET and the guard: an existing row keeps the first `created_at` it was stamped with, so a Boardsesh-originated rating adopted by a later sync is not re-dated to the Kilter timestamp.
- `applyClimbRatings` is now exported from `@boardsesh/kilter-sync/sync` so the real-DB test can drive it. `parseKilterTimestamp` stays module-local and is covered by unit tests in the package's own suite.
- `docs/kilter-sync.md` gains the two invariants a future change could silently break: `created_at` comes from upstream and is never re-stamped on conflict, and the DO UPDATE is change-guarded.

### No backfill for existing rows

Existing rows' original `created_at` is **unrecoverable**. The upstream date was never persisted anywhere on our side — it was read off the PowerSync payload and dropped, and the payload is not retained. The only surviving value is the sync-time default already in the column, and Kilter's PowerSync stream carries no history we could replay a date from without a full re-pull that would itself rewrite the same rows. So this PR changes go-forward behaviour only. Rows re-delivered by a future sync cycle will keep their (wrong) `created_at` by design, because re-stamping on conflict would break the Boardsesh-originated case above. If we later decide the wrong dates are worse than the churn, that is a deliberate one-shot migration, not something this upsert should do implicitly.

## Tests

New real-Postgres test: `packages/backend/src/__tests__/kilter-climb-ratings-timestamps.test.ts` (7 tests). The change guard is a property of the generated SQL, not of JS — a mocked-drizzle test could only re-assert a predicate we wrote ourselves, so this drives the actual `insert … on conflict do update` against Postgres and reads the stored timestamps back.

Plus, in the package's own DB-free suite (`packages/kilter-sync/src/sync/user-sync.test.ts`): the existing bulk-insert test now asserts `createdAt` on the values map, so dropping the mapping fails there too without a database; and `parseKilterTimestamp` gets direct coverage for the `Z`, explicit-offset, zone-less, date-only, and unparseable branches.

Coverage: upstream date is stored (with a non-UTC offset, proving UTC normalisation); unparseable date falls back to the default instead of aborting; an unchanged redelivery leaves `updated_at` and `created_at` untouched; a real rating change bumps `updated_at` but not `created_at`; `kilter_id` adoption onto a Boardsesh-originated row bumps once then settles; the guard holds across a multi-row batch (the real sync flushes hundreds per statement); and a cleared upstream comment is pinned as a real change.

**Fail-on-revert, both arms verified:**

- Removing the `createdAt` mapping → 4 failed (`expected 1785457710098 to be 1709656200000` — sync time instead of the rating date).
- Removing the `setWhere` → 3 failed (`expected 1785458436147 to be 1577836800000` — `updated_at` rewritten on an unchanged redelivery).

## QA Notes

Backend/sync-worker only, no UI surface. Nothing to click through.

To exercise on a dev DB: run a Kilter user sync twice with no upstream changes between cycles, then check `board_climb_ratings` — `created_at` should match the rating dates in the Kilter app (not today), and `updated_at` should be identical across both runs. Change one rating in the Kilter app, sync again, and only that row's `updated_at` should move.

## Release Notes

none — nothing currently reads `board_climb_ratings.created_at` (the only consumer of the table is the tick-quality join, which reads `rating`), so the corrected dates aren't visible on any surface yet. This fixes the data before a surface depends on it.

Fixes #3524

### DIFF
--- a/docs/kilter-sync.md
+++ b/docs/kilter-sync.md
@@ -128,6 +128,8 @@ Kilter exposes ratings as a first-class resource at `POST /api/climb-rating/` �
 
 The conflict target on insert is the natural key `(board_type, climb_uuid, angle, user_id)`, not `kilter_id` — same reasoning as the ticks path. Kilter's per-rating payload doesn't carry `weight`, so kilter-origin rows leave that column NULL.
 
+Two timestamp invariants are easy to break by accident (issue #3524). **`created_at` comes from the upstream rating date** (`raw.created_at`, parsed by `parseKilterTimestamp`), not from the column default — without it every historical rating collapses onto the day our sync first saw it. It is deliberately absent from the `DO UPDATE` SET, so a Boardsesh-originated rating that a later sync adopts keeps its own original date rather than being re-stamped with Kilter's. **The `DO UPDATE` is change-guarded** by a row-wise `IS DISTINCT FROM` over `(rating, difficulty_grade_id, comment, kilter_id)`, comparing the existing row against the *effective* new value (`COALESCE(EXCLUDED.…, existing)` for the COALESCE'd columns, not bare `EXCLUDED`). PowerSync redelivers a full snapshot every cycle, so without the guard the update fires for every rating on every run and `updated_at` degrades into "when the sync last ran".
+
 ### Circuits → playlists
 
 Circuit rows upsert into `playlists` (`kilter_id` = `circuit_uuid`, `kilter_type = 'circuits'`). The climbs list is **full-replace per sync**: delete every `playlist_climbs` row for the playlist, re-insert from the snapshot, chunk inserts at 500 rows to stay under the 65535-parameter Postgres ceiling. Matches aurora-sync's circuit handling. Ownership is idempotent via `(playlist_id, user_id)` unique.
--- a/packages/backend/src/__tests__/kilter-climb-ratings-timestamps.test.ts
+++ b/packages/backend/src/__tests__/kilter-climb-ratings-timestamps.test.ts
@@ -1,0 +1,238 @@
+// Real-Postgres test for the kilter-sync climb-ratings upsert (issue #3524).
+//
+// Both behaviours under test are properties of the generated SQL, not of JS,
+// so a mocked-drizzle unit test could only re-assert a predicate we wrote
+// ourselves. These run the real `insert ... on conflict do update` against
+// Postgres and read the stored timestamps back:
+//
+//   1. created_at must come from the upstream rating date, not defaultNow().
+//   2. updated_at must only move when the rating actually changed — PowerSync
+//      redelivers a full snapshot every cycle, so an unguarded DO UPDATE
+//      rewrites updated_at on every sync for every row.
+import { describe, it, expect, beforeEach } from 'vite-plus/test';
+import { and, eq, sql } from 'drizzle-orm';
+import { boardClimbRatings } from '@boardsesh/db/schema';
+import { applyClimbRatings } from '@boardsesh/kilter-sync/sync';
+import type { PowerSyncOp } from '@boardsesh/kilter-sync/api';
+import { db } from '../db/client';
+
+// Seeded by the shared test setup.
+const USER_ID = 'user-123';
+const CLIMB_UUID = 'climb-rating-ts-1';
+const ANGLE = 40;
+// Deliberately carries a non-UTC offset: Kilter sends ISO strings with an
+// offset and the stored value must be the normalised UTC instant.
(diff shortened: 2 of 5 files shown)


marcodejongh:
Applied the date-only nit — `parseKilterTimestamp` now builds `2024-03-05T00:00:00Z` rather than relying on V8 accepting `2024-03-05Z`.

Skipping the two comment-trimming nits. The `createdAt` block doesn't mention issue #3524 (it explains why the column can't fall through to `defaultNow()`), and neither comment states anything the code shows on its own — both record a non-obvious constraint that a future edit would otherwise silently break. There's no repo convention against naming the test that consumes an export.

PR #4046, opened 2026-07-31
fix(backend): clear kiosk queue previews when a board goes private or is deleted

### LINKED ISSUE #3649: Queue preview: tombstone missing when a BOARD flips private mid-session
## Problem
#3628 added tombstones so kiosk queue previews clear when a session ends or goes private. The delta re-review found the same gap one level up: `updateBoard` with `isPublic: false` (social/boards.ts `updateBoard`) stops NEW publishes (the producer re-gates each publish) but already-connected anonymous subscribers keep the LAST snapshot indefinitely — and the tombstone helper itself requires the board to be anon-readable, which is false after the flip, so even session end won't clear that kiosk. Heals only on kiosk reconnect (then NOT_FOUND).

Severity: low — owner-initiated, no new data flows after the flip, bounded by kiosk reload cadence (5-min config poll / daily reload).

## Proposal
In `updateBoard`'s `isPublic: false` branch, tombstone the bound session's board channel BEFORE the flip commits (or relax the tombstone guard for just-flipped boards). Same treatment for board soft-delete. Test: connected subscriber receives the empty snapshot on flip.

Context: #3628 (see `publishBoardQueuePreviewTombstoneForSession` and its guards).

marcodejongh:
## What / why

A gym kiosk showing a board's "Up next" queue kept displaying the last snapshot forever after the board was flipped to private or deleted. The board-queue-preview producer only republishes on queue events, so once a board leaves the anon-readable set it simply goes quiet — nothing ever tells the kiosk to clear.

## Verified root cause

- `updateBoard` (`packages/backend/src/graphql/resolvers/social/boards.ts`) writes `updateValues.isPublic` and commits, but never touches pubsub — `rg pubsub` in that file had zero hits.
- `deleteBoard` in the same file has the identical gap on soft-delete.
- Naively reusing the session-level helper would not have fixed it: `publishBoardQueuePreviewTombstoneForSession` gates on `isBoardAnonReadable(boardId)`, which re-reads `user_boards.is_public` live. Once the flip commits that gate is false and the call silently no-ops.
- The tombstone module's own doc comment flagged this as an open gap (`board_sessions.is_public`); the issue is one level up, at `user_boards.is_public`, and is distinct from #3628's session-level tombstone.

## Fix

- New `publishBoardQueuePreviewTombstoneForBoard(boardId)` in `packages/backend/src/services/board-queue-preview-tombstone.ts`: publishes an empty snapshot for the board. It deliberately skips the anon-readable / session-active gates, since by the time a caller reaches it the board is already private or deleted and both gates would suppress the very publish it exists to make. Re-exported from `board-queue-preview.ts` (single import point).
- It also does **not** require a live `board:{id}:session` binding. Paired QA caught that an earlier revision did: that binding is only written as a side effect of a climb report, carries a 12h TTL, and is process-local memory when Redis is off — while the snapshot a kiosk renders can equally come from `resolvePublicPreviewSessionForBoard`'s durable fallback (newest active public `board_sessions` row). Gating on the binding stranded exactly those kiosks — no send in 12h, Redis flushed, or a Redis-less restart — on the stale queue, the issue's own symptom. An empty publish on a channel nobody watches costs one PUBLISH.
- The privacy gate lives at the **call sites**, which already hold the pre-update row — no extra query, and no publish on a channel that was never anon-readable (that alone would leak "something changed here" to anyone who guessed a private board's id):
  - `updateBoard`: fires only on a genuine true→false flip of a live, non-system-owned board (`board.isPublic && input.isPublic === false && !board.deletedAt && ownerId !== SYSTEM_BOARD_OWNER_ID`). System-owned boards stay anon-readable regardless of `isPublic`, so a flip there is not a transition and must not blank their kiosks.
  - `deleteBoard`: fires when the pre-delete row was anon-readable, which here is just `board.isPublic` — the other half of `isBoardAnonReadable` (system-owned boards) is unreachable, because the ownership check above it only lets the authenticated owner through and nobody can authenticate as the synthetic system owner.
- Both are wrapped in try/catch + `logger.warn`, so a pubsub hiccup never fails the edit or the delete.

## Tests

Extends the existing `board-queue-preview tombstone` describe in `packages/backend/src/__tests__/board-queue-preview.test.ts` (7 new cases), reusing its `makePreviewableSession` / `bindSessionToBoard` / `subscribeBoardQueuePreview` harness:

- board-level tombstone publishes one empty snapshot for the board, including with no session binding at all (the DB-fallback kiosk case).
- `updateBoard(isPublic: false)` on a public board whose kiosk was seeded through the durable fallback with **no** binding → one empty snapshot (asserts the seed was non-null first).
- `updateBoard(isPublic: false)` on a previewable public board → subscriber receives exactly one empty snapshot.
- `updateBoard` that renames only, and one that re-asserts `isPublic: true` → zero publishes.
- `updateBoard(isPublic: false)` on an already-private board → zero publishes (the privacy invariant).
- `deleteBoard` on a public board with a bound session → one empty snapshot; on a private board → zero.
- system-owned board flipping `isPublic: false` → zero publishes (it stays anon-readable, so its kiosks must not blank).
- soft-deleted board restored *and* flipped private in one update → zero publishes (not an anon-readable transition).

**Fail-on-revert:** with `boards.ts` reverted to `main` and the tests kept, the two end-to-end cases fail (`expected [] to have a length of 1 but got +0`); the rest of the file stays green. Reinstating the binding early-return in the helper fails the two DB-fallback cases and nothing else. These call the real resolvers and assert on a real subscriber, so they cannot pass without the wiring.

`makeBoard` now inserts a real UUID (the social board mutations validate `boardUuid` as a UUID) and is split into `makeBoardRow` returning `{ id, uuid }`.

Validation: `vp run typecheck:backend` clean; `vp test run --project backend packages/backend/src/__tests__/board-queue-preview.test.ts` → 42/42 pass; `vp check` formatting clean (its 2 lint errors are pre-existing and in other packages, untouched here).

Test-infra note: this file needs an isolated postgres to run reliably right now — the shared `:5433` worker DBs deadlock in the `beforeEach` TRUNCATE when other backend suites run concurrently. That is environmental, not this branch.

## QA Notes

Backend-only, no schema or migration change.

1. Make a board public and take a session on it so a kiosk shows the queue.
2. Flip the board to private in board settings — the kiosk preview must clear immediately instead of holding the last climb.
3. Repeat with delete — same immediate clear.
4. Sanity: editing a private board's name or visibility must not push anything to its preview channel.
5. Fallback path worth exercising: leave the board idle long enough that no recent climb was reported (or restart the backend), then flip it private — the kiosk must still clear.

## Release Notes

Kiosk displays now clear a board's queue preview the moment you flip that board to private or delete it, instead of leaving the last climb on screen.

Fixes #3649

### DIFF
--- a/packages/backend/src/__tests__/board-queue-preview.test.ts
+++ b/packages/backend/src/__tests__/board-queue-preview.test.ts
@@ -20,6 +20,7 @@ import {
   buildBoardQueuePreview,
   getBoardQueuePreviewSnapshot,
   publishBoardQueuePreviewForSession,
+  publishBoardQueuePreviewTombstoneForBoard,
   publishBoardQueuePreviewTombstoneForSession,
   registerBoardQueuePreviewHook,
   toBoardQueuePreviewItem,
@@ -31,6 +32,7 @@ import {
 import { boardPresenceMutations } from '../graphql/resolvers/board-presence/mutations';
 import { SYSTEM_BOARD_OWNER_ID } from '../graphql/resolvers/board-presence/shared';
 import { roomManager } from '../services/room-manager';
+import { socialBoardMutations } from '../graphql/resolvers/social/boards';
 
 const TEST_USER_ID = 'board-queue-preview-test-user';
 const TEST_BOARD_PATH = 'queue-preview-test/1/10/1,2/40';
@@ -87,18 +89,23 @@ function makeQueueEvent(item: ClimbQueueItem): QueueEvent {
 }
 
 let boardSlugCounter = 0;
-async function makeBoard({
+/**
+ * Insert a board and return both keys: the numeric id the preview channel is
+ * keyed by, and the (real, schema-valid) uuid the social board mutations take.
+ */
+async function makeBoardRow({
   isPublic,
   ownerId = TEST_USER_ID,
 }: {
   isPublic: boolean;
   ownerId?: string;
-}): Promise<number> {
+}): Promise<{ id: number; uuid: string }> {
   const slug = `qp-board-${Date.now().toString(36)}-${boardSlugCounter++}`;
+  const boardUuid = uuidv4();
   const [row] = await db
     .insert(dbSchema.userBoards)
     .values({
-      uuid: `uuid-${slug}`,
+      uuid: boardUuid,
       slug,
       ownerId,
       boardType: 'kilter',
@@ -110,7 +117,12 @@ async function makeBoard({
       isPublic,
     })
     .returning({ id: dbSchema.userBoards.id });
-  return Number(row.id);
+  return { id: Number(row.id), uuid: boardUuid };
+}
+
+async function makeBoard(options: { isPublic: boolean; ownerId?: string }): Promise<number> {
+  const { id } = await makeBoardRow(options);
+  return id;
 }
 
 async function makeSession({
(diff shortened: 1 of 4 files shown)


marcodejongh:
Review triage:

- **Comment verbosity**: no such rule in CLAUDE.md — the bot has invented this one before. The docstrings here carry the privacy contract (who owns the gate, why the gates are skipped), which is the part a future reader gets wrong. Keeping them.
- **Unawaited pubsub call**: `PubSub.publishBoardQueuePreview` is synchronous `void` (pubsub/index.ts:521) — it dispatches locally and the Redis fan-out is fire-and-forget inside `PubSubChannel`, which logs its own failures. Nothing escapes the call-site try/catch that was ever caught by it. Same shape as the pre-existing `...ForSession` tombstone.
- **Ordering vs the PostGIS update**: the `isPublic = false` write is its own committed statement before the tombstone, so a later PostGIS failure leaves the board genuinely private — the tombstone is correct regardless, and moving it after would mean a failed location write leaves kiosks stuck on a stale queue.

PR #4048, opened 2026-07-31
fix(sync): resync user data when the deletions retention window is blown

### LINKED ISSUE #3474: mobile: force a full resync when the deletions checkpoint predates the sync_deletions retention window
Follow-up from #2785. The backend now prunes `sync_deletions` tombstones older than 90 days (`SYNC_DELETIONS_RETENTION_DAYS`, `packages/backend/src/services/sync-deletions-prune.ts`), but a client whose deletions checkpoint is older than that window can miss pruned tombstones: records deleted server-side stay in its local SQLite until they happen to be overwritten by an upsert — for hard-deleted rows, forever.

Fix shape: in the mobile pull client, when the stored deletions checkpoint timestamp is older than (retention − safety margin, e.g. 80 days), treat local data as unrecoverable-stale — clear the sync checkpoints and user-data/board-data tables, then let the normal pull rebuild from scratch. Needs a decision on how that interacts with `pending_mutations` (never drop unsynced writes — drain or preserve them before the wipe).

Rare edge (device offline for 3+ months), nothing shipped yet, so tracked rather than blocking #2785.

marcodejongh:
## What & why

A device that goes months without syncing keeps ticks, playlists and favourites you deleted elsewhere — permanently.

The backend hard-deletes tombstones older than 90 days (`pruneSyncDeletions`, run daily from `server.ts`), and the `syncDeletions` resolver pages on a strict `>` keyset. So every tombstone whose `deleted_at` fell in `(client cursor, now − 90d]` was destroyed before that client ever asked for it, and the client can never be served it again. Those local rows survive until an unrelated upsert of the same primary key — which, for a row the server hard-deleted, never happens. The prune service's own doc comment flagged this client-side gap as unaddressed.

Fixes #3474.

## Verified root cause

- `packages/backend/src/services/sync-deletions-prune.ts` — 90-day cutoff, `DELETE FROM sync_deletions WHERE deleted_at < cutoff`.
- `packages/backend/src/graphql/resolvers/sync/queries.ts` — `(deleted_at, id) > (cursor)`, strict.
- `packages/shared/offline-sync/src/sync/pull-client.ts` — `processDeletions` read the checkpoint and paged straight from it, with no notion of how old that coverage was.

Two corrections to the obvious fix, both load-bearing:

**The staleness oracle cannot be the checkpoint's own age.** `checkpoint:deletions` stores the server-side `deleted_at` of the last consumed tombstone, and only advances on a non-empty page. A user who has deleted nothing for three months carries a three-month-old cursor on a perfectly current device. Worse, in production the board tables emit **no** tombstones at all — the only production DELETEs of `board_climbs` / `board_climb_stats` run inside `clearAuroraBoardData`'s transaction, which sets `boardsesh.suppress_sync_tombstones` (`packages/db/drizzle/0144_offline_sync_backend.sql:391-393` + `packages/db/src/queries/boards/clear-aurora-board.ts:154`), and `board_climb_grades` has no delete trigger. So the only tombstones any user sees are their own deletes. A cursor-age test would have reset most of the fleet.

**The reset must not touch the board catalog.** Same finding: no board tombstone is ever emitted in production, so a blown deletions window costs the offline catalog nothing. Clearing it would re-download tens to hundreds of MB, unprompted, possibly on cellular — the exact cost `remove-offline-board.ts` already treats as something to ask about.

## Fix

A wall-clock **deletions-coverage marker**. A client that reached the deletions tail at time T has consumed every tombstone that existed at T; anything pruned after T was already applied. It can only miss data if `now − T` exceeds retention.

- `packages/shared/offline-sync/src/sync/retention.ts` (new) — `SYNC_DELETIONS_RETENTION_DAYS = 90`, `DELETIONS_COVERAGE_MARGIN_DAYS = 10`, `DELETIONS_COVERAGE_MAX_AGE_MS` (80 days). The backend prune job now imports and re-exports the retention constant instead of defining its own, so the two sides can't fork.
- `packages/shared/offline-sync/src/sync/deletions-coverage.ts` (new) — `sync_meta['deletions-coverage']` (deliberately outside the `checkpoint:` prefix, so sign-out's checkpoint wipe leaves it alone), the four-way verdict, and `resetUserDataForLostCoverage`: one exclusive transaction that deletes the rows of every `USER_DATA_TABLES` entry, their checkpoints and `checkpoint:deletions`, then stamps the marker.
- `packages/shared/offline-sync/src/sync/pull-client.ts` — `processDeletions` now returns `{ reachedTail }` (mirroring `syncTable`), and the marker only advances on a completed pass. A new guard runs at the top of `pullSync`, before the bootstrap phase.
- Cache busting: the reset invalidates every `USER_DATA_TABLES` react-query key itself. It can't lean on the rebuild — `syncTable` only busts caches when it pulled at least one document and `processDeletions` only on arrivals, so a table the user emptied server-side re-pulls nothing and a mounted screen would keep serving the pre-wipe cache.
- Telemetry: `Offline Sync Coverage Reset Forced` via the injected `onCoverageReset` seam → `track()` in the mobile adapter. The shared package stays analytics-free.

Behaviour of the guard, branch by branch:

| marker                                          | action                                        |
| ----------------------------------------------- | --------------------------------------------- |
| absent, or implausibly old (broken clock)       | nothing — **no reset**, and no stamp either   |
| dated in the future (clock corrected backwards) | nothing — **no reset**                        |
| within 80 days                                  | nothing (one `sync_meta` read)                |
| older than 80 days                              | probe → reset user data → same cycle rebuilds |

The marker is written in exactly one place: after the deletions pull reaches its tail. That is the only moment coverage actually exists, so a first post-update pull that fails or gets backgrounded leaves the marker absent rather than handing a device that has been away 89 days a fresh 80-day window it never earned.

Five further guards came out of review, each closing a path where a _healthy_ device could reset itself (or claim coverage it never had):

- **Plausibility floor** (`DELETIONS_COVERAGE_EPOCH_FLOOR_MS`, 2025-01-01). A phone that cold-boots with a dead RTC (1970 / 2001) before NTP lands reads the real marker as `future` and re-stamps it with its bogus now; once the clock corrects, that stamp is decades old and authorises a wipe on a device that has been syncing daily. Anything below the floor classifies as `unknown`, so it resets nothing and is overwritten by the next completed pass.
- **Digits-only marker parse.** `Number.parseInt` tolerates trailing garbage, so an ISO string (`'2026-07-01T…'` → `2026`) would have parsed to ~1970 and read as stale — the opposite of what `getDeletionsCoverageAt`'s "a corrupt value reads as absent" contract promised.
- **No eager stamp on a non-stale verdict** (Codex P2). The guard used to seed the marker up front for `unknown` / `future`; a first pull that then failed left behind a coverage claim the device never established. Now only the completed-pass stamp writes it.
- **The marker is dropped on sign-out.** It sat outside the `checkpoint:` prefix and survived the sign-out teardown, but sign-out rewinds the deletions cursor to the epoch, so a departing account's stale marker only served to trip the guard on the NEXT account's first pull — a wasted probe and a reset of tables sign-out had already emptied, reported as `rowsCleared: 0`.
- **`cycleEpoch` is captured before the coverage phase**, and the post-probe re-check compares it. The probe is a real round-trip, so a `beginLocalPurge()` (Storage → Remove board) could land inside it and be adopted as this cycle's own epoch — the wipe would then run and every later phase would bail at its first `cycleAborted()`, leaving user data cleared with no rebuild behind it.

The `stale` branch **probes the network first** — a one-row `syncDeletions` request whose result is discarded. `pullSync` runs on every foreground including offline ones; wiping first and then failing the first fetch would leave the user with an empty app until connectivity returns. The probe also proves the credential, so an expired-token device can't wipe itself. `pending_mutations` is never touched.

## Tests + fail-on-revert

`packages/shared/offline-sync/src/sync/__tests__/deletions-coverage.test.ts` (new) and a new block in `pull-client.integration.test.ts` (real DDL via `node:sqlite`), plus one case in the mock-db `pull-client.test.ts` and the mobile adapter suite.

- **stale + online** — 100-day-old marker: stale tick and playlist gone, freshly-pulled tick present, `board_climbs` and `checkpoint:board_climbs:*` untouched, `pending_mutations` intact, marker advanced. _Revert phase −1 out of `pullSync` and this fails_ — verified: only this test broke.
- **stale + offline** — the fetch rejects: nothing is wiped, the old marker survives. _Remove the probe and both this and the case above fail_ — verified.
- **absent marker (bootstrap OTA)** — no reset, rows survive, and the marker is seeded by the completed pass itself. _Make absence mean "stale" and 20 tests across the suite fail_ — verified. This is the fleet-wipe guard.
- **no coverage claim without a completed pass** — absent marker + a failing fetch: the marker stays absent. _Re-add the eager stamp and this plus the aborted-pull test fail_ — verified.
- **sign-out** — a stale marker does not survive `deleteUserCheckpoints`, so it cannot leak into the next account.
- **empty tables** — a reset on already-cleared tables is a clean no-op that still stamps the marker (`rowsCleared: 0`).
- **wrong oracle** — `checkpoint:deletions` dated 200 days ago, coverage marker 1 day old: no reset. Fails if anyone rewires the oracle to the checkpoint.
- **aborted deletions pull** — backgrounded mid-stream: the marker does not advance. Fails if `processDeletions` goes back to returning `void`.
- **purge mid-probe** — `beginLocalPurge()` fires while the probe is on the wire: nothing wiped, marker unchanged, no reset event. _Drop the epoch comparison from the post-probe re-check and only this test fails_ — verified.
- **broken clock** — marker at epoch 0 and at 2001 read as `unknown`, not `stale`; the floor value itself is still `stale`. _Delete the floor check and this fails_ — verified.
- **garbled marker** — `'not-a-number'`, an ISO string, empty, whitespace, `'17e14'`, `'-1'` all read as absent. _Delete the digits-only test and the ISO case fails_ — verified.
- **future-dated marker**, boundary cases at 79 / 80 / 81 days.
- **reset scope** — `pending_mutations` byte-identical across the reset, `schema_version` untouched, every board table and marker family preserved.
- **backend** — the prune window and the client's shared constant must be the same value; re-introducing a local `= 90` and editing either side fails at the seam.

## Blast radius

Pure JS/TS, no native code and no DB migration (the marker is a row in the existing `sync_meta`). CI still reports a fingerprint change and applies `native-fingerprint`, same as every other JS-only change under `packages/shared` — so it rides the next store/TestFlight build rather than an OTA. Every offline mobile user's first sync after the update writes the marker and resets **nothing**.

## QA Notes

The failure being guarded is rare; the regression it could cause is not. Weight QA accordingly.

1. **Negative case (the important one).** On a device with boards downloaded, take the build and foreground the app. Expect: boards keep their "downloaded" chips, ticks/playlists/favourites all still there, a normal short sync cycle, no re-download, no `Offline Sync Coverage Reset Forced` event.
2. **Positive case, forced.** There's no user-facing trigger — hand-write `sync_meta['deletions-coverage']` to `Date.now() - 100 * 86400000` (dev script or SQLite write), then foreground. Expect: one small `syncDeletions` request, ticks/playlists/favourites clear and immediately re-populate from the server, downloaded boards stay downloaded (no catalog re-download), and one PostHog event with `markerAgeDays ≈ 100`.
3. **Offline safety.** Repeat (2) in airplane mode. Expect: nothing wiped — ticks, playlists and downloaded boards all still present — and a failed sync cycle in Settings. Re-enable the network and confirm the reset then runs on the next trigger.
4. **Outbox.** Repeat (2) in airplane mode, log a tick offline so it queues, then re-enable the network and foreground. Expect: the reset runs and the offline tick still reaches the server and reappears in the logbook. It may blink out between the wipe and the drain — that transient is expected; permanent loss is not.

## Release Notes

Pick the app back up after months away and it now rebuilds your ticks, playlists and favourites from the server, instead of quietly hanging on to ones you deleted while you were gone. Your downloaded boards stay put — no surprise re-download — and anything you logged offline still goes up.

### DIFF
--- a/docs/offline-sync-plan.md
+++ b/docs/offline-sync-plan.md
@@ -600,7 +600,24 @@ ORDER BY deleted_at ASC, id ASC
 LIMIT $limit;
 ```
 
-Periodic cleanup: `DELETE FROM sync_deletions WHERE deleted_at < NOW() - INTERVAL '90 days'`.
+Periodic cleanup: `DELETE FROM sync_deletions WHERE deleted_at < NOW() - INTERVAL '90 days'` (the daily
+`pruneSyncDeletions` job).
+
+Because that prune is a hard delete and the query above is a strict `>` keyset, a device that has been
+away longer than the window can never be served the tombstones removed in the meantime. The client
+guards it with a wall-clock **deletions-coverage marker** (`sync_meta['deletions-coverage']`, written
+only when a deletions pull reaches its tail — never the checkpoint's own age, which stands still on a
+device whose user simply deleted nothing). Once the marker is older than `90 - 10 = 80` days, `pullSync`
+probes the network first and then rebuilds the local **user** tables from scratch: it deletes the rows of
+every `USER_DATA_TABLES` entry plus their checkpoints and `checkpoint:deletions`, and the same cycle
+re-pulls them from the epoch.
+
+Downloaded board catalogs are deliberately left alone — no production path emits a board tombstone (the
+only DELETEs of `board_climbs` / `board_climb_stats` run inside the transaction that sets
+`boardsesh.suppress_sync_tombstones`), so clearing them would recover nothing and cost a full
+re-download. `pending_mutations` is never touched either: unsynced local writes are not recoverable from
+the server. Both numbers live in one place, `packages/shared/offline-sync/src/sync/retention.ts`, which
+the backend prune job imports. See `sync/deletions-coverage.ts` for the invariant.
 
 ## Backend changes needed
 
--- a/packages/backend/src/services/__tests__/sync-deletions-prune.test.ts
+++ b/packages/backend/src/services/__tests__/sync-deletions-prune.test.ts
@@ -1,5 +1,6 @@
 import { describe, it, expect, vi } from 'vitest';
 import { syncDeletions } from '@boardsesh/db/schema';
+import { SYNC_DELETIONS_RETENTION_DAYS as SHARED_RETENTION_DAYS } from '@boardsesh/offline-sync';
 import { pruneSyncDeletions, SYNC_DELETIONS_RETENTION_DAYS } from '../sync-deletions-prune';
 import type { Database } from '../../db/client';
 
@@ -30,6 +31,16 @@ describe('pruneSyncDeletions', () => {
     expect(boundDates).toHaveLength(1);
     expect(boundDates[0].getTime()).toBeGreaterThanOrEqual(beforeMs);
     expect(boundDates[0].getTime()).toBeLessThanOrEqual(afterMs);
+  });
+
+  it('prunes on the SAME window the client staleness guard compares against', async () => {
+    // The client forces a from-scratch user-data resync once its deletions
+    // coverage marker is older than this value minus a margin (issue #3474). If
+    // the two ever forked — a local `= 90` re-introduced here and then edited —
+    // the server would prune tombstones inside a window the client still
+    // believes it is covered for, silently reopening the gap on device.
+    expect(SYNC_DELETIONS_RETENTION_DAYS).toBe(SHARED_RETENTION_DAYS);
+    expect(SYNC_DELETIONS_RETENTION_DAYS).toBe(90);
   });
 
   it('returns 0 when the driver reports no count', async () => {
--- a/packages/backend/src/services/sync-deletions-prune.ts
+++ b/packages/backend/src/services/sync-deletions-prune.ts
@@ -1,5 +1,6 @@
 import { lt } from 'drizzle-orm';
 import { syncDeletions } from '@boardsesh/db/schema';
+import { SYNC_DELETIONS_RETENTION_DAYS } from '@boardsesh/offline-sync';
 import { db, type Database } from '../db/client';
 
 /**
@@ -7,13 +8,18 @@ import { db, type Database } from '../db/client';
  * `sync_deletions` (packages/db/src/schema/app/sync-deletions.ts): tombstones
  * older than this are pruned by the daily job in server.ts.
  *
- * A client whose deletions checkpoint is older than this window can miss
- * pruned tombstones (stale local rows until the affected records are next
- * upserted). Detecting that gap and forcing a from-scratch resync is a
- * client-side follow-up; the unbounded-growth risk of never pruning is the
- * worse trade.
+ * The value is OWNED by @boardsesh/offline-sync (sync/retention.ts) and
+ * re-exported here under its long-standing name. The client's staleness guard
+ * compares its deletions-coverage marker against the same number minus a
+ * margin, and forces a from-scratch user-data resync when the window is blown
+ * (issue #3474) — if the two sides ever forked, the gap this job opens would go
+ * undetected on device again. Keep the single definition.
+ *
+ * Careful when LOWERING it: a client only picks up the new value with its next
+ * OTA bundle, so shortening the server window strands older clients comparing
+ * against the old one.
  */
-export const SYNC_DELETIONS_RETENTION_DAYS = 90;
+export { SYNC_DELETIONS_RETENTION_DAYS };
 
 export async function pruneSyncDeletions(database: Database = db): Promise<number> {
   const cutoff = new Date(Date.now() - SYNC_DELETIONS_RETENTION_DAYS * 24 * 60 * 60 * 1000);
--- a/packages/mobile/src/offline/__tests__/offline-sync-adapter.test.ts
+++ b/packages/mobile/src/offline/__tests__/offline-sync-adapter.test.ts
@@ -219,6 +219,7 @@ describe('snapshot-bootstrap bindings', () => {
     const options = startSyncSchedulerCore.mock.calls[0][6] as SchedulerOptions;
     expect(options.onSnapshotBootstrapError).toBeTypeOf('function');
     expect(options.onScopeDownloadComplete).toBeTypeOf('function');
+    expect(options.onCoverageReset).toBeTypeOf('function');
   });
 
   it('triggerSync passes snapshotSource through and wires the telemetry handlers', () => {
@@ -234,6 +235,7 @@ describe('snapshot-bootstrap bindings', () => {
     expect(options.snapshotSource).toBe(fakeSnapshotSource);
     expect(options.onSnapshotBootstrapError).toBeTypeOf('function');
     expect(options.onScopeDownloadComplete).toBeTypeOf('function');
+    expect(options.onCoverageReset).toBeTypeOf('function');
   });
 
   it('reports a bootstrap failure to Sentry with the snapshot-bootstrap tags', () => {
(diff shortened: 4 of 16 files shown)


marcodejongh:
Review follow-ups, all pushed or answered:

- **Dead `?? resetAt` fallback** — fixed in 7abc052. The null check that IS the `unknown` verdict now does the narrowing, so `markerAgeDays` needs no fallback.
- **`rowsCleared` assertion** — fixed in the same commit: the test now counts the seeded rows out of the DB and asserts against that total, so a table with 0 or 2 rows can no longer hide behind the table count.
- **`cursor: undefined` in the probe** — checked the resolver. `cursorBounds()` in `packages/backend/src/graphql/resolvers/sync/queries.ts` is typed `SyncCursorInput | null | undefined` and reads `cursor?.updatedAt ?? EPOCH_TS`, so absent and `null` are the same start-of-time path. No change needed.
- **`native-fingerprint` label** — not a mislabel. Every JS-only change that touches the mobile dependency graph gets it in this repo (#4050, #4041, #3997, #3981 all carry it on pure-TS diffs). The body's "OTA-safe" line was the wrong half of that pair and is now corrected: this ships with the next store build.
- **No probe timeout** — pre-existing property of every `graphqlFetch` call on this path; out of scope for a bug fix, and a device that can't complete a one-row query also can't complete the rebuild it gates.
- **`DELETE FROM ${tableName}`** — `tableName` is a `USER_DATA_TABLES` literal, and SQLite cannot parameterise a table name. Same shape as `removeBoardScopeData`.

marcodejongh:
Codex P2 (stamp unknown coverage only after reaching the tail) — agreed and applied in a23d1048c. The marker now has exactly one writer, the completed-pass stamp; `unknown` and `future` do nothing at all rather than seeding a window the device never established. `future` still unfreezes, via that same stamp. Pinned by a new integration test (absent marker + failing fetch ⇒ marker stays absent); re-adding the eager stamp fails it and the aborted-mid-stream test.

Also in this round, from the paired QA review:
- plausibility floor (2025-01-01) so a dead-RTC boot can't drive a wipe once the clock corrects;
- digits-only marker parse, so an ISO string can't read as ~1970;
- `cycleEpoch` captured before the coverage phase and compared after the probe, so a `beginLocalPurge()` landing mid-probe can't be adopted as this cycle's own epoch.

marcodejongh:
Both review nits applied in f2b52e17a: one `evaluatedAt` clock reading drives the verdict and `markerAgeDays` (the stamp still reads fresh, since it claims coverage as of the wipe), and there's now a `cycleAborted()` immediately after the coverage phase so a teardown that landed during it can't kick off the bootstrap download.

marcodejongh:
Third round applied:

- **Cross-account marker leak** — agreed, fixed in e863368c3. `deleteUserCheckpoints` now clears `deletions-coverage` too; sign-out rewinds the deletions cursor to the epoch, so the departing account's marker describes nothing and the next account re-stamps on its own first completed pull. The key moved to `retention.ts` (dependency-free) so `checkpoints.ts` can name it without a cycle.
- **Missing test: marker survives sign-out** — that invariant is now the opposite one, and it is pinned: `checkpoints.test.ts` asserts a stale marker is gone after `deleteUserCheckpoints`.
- **Missing test: rowsCleared: 0 on empty tables** — added in e42403617.
- **markerAgeDays rounding** — switched to `Math.floor` in e42403617.

marcodejongh:
Both applied in the latest commit: `readCoverage` in the integration suite now calls `getDeletionsCoverageAt` instead of `Number()`, and the reset's `DELETE FROM ${tableName}` carries the compile-time-literal contract note (same wording as removeBoardScopeData).

PR #4050, opened 2026-07-31
fix(mobile): carry queue-item attribution across the mobile wire

### LINKED ISSUE #3995: Mobile queue writes drop addedBy/addedByUser/tickedBy/suggested — no "added by" avatar for web peers
## Summary

Mobile's queue-item wire mapper sends only `{ uuid, climb }`. It drops the four item-level fields web sends — `addedBy`, `addedByUser`, `tickedBy`, `suggested` — so in a mixed web + mobile party session a climb added from the phone shows up on web peers with **no "added by" avatar**, and its suggestion origin doesn't survive the round trip.

Found while closing #2414 (mobile queue-mutation parity). #2414's own description of the payload gap — "mobile-added items lose the Edit affordance and ascent state for peers" — is genuinely fixed: mobile's `toClimbInput` now carries `userId`, `is_draft`, `published_at`, `userAscents` and the whole 26-field `ClimbInput` set, byte-for-byte the same field list web sends. This is the leftover, at the level above the climb.

## Evidence

Mobile's mapper — `packages/mobile/src/providers/queue-provider.tsx:426`:

```ts
toQueueItemInput: (item) => ({ uuid: item.uuid, climb: toClimbInput(item.climb) }),
```

Web's — `packages/web/app/components/persistent-session/types.ts`, `toClimbQueueItemInput`, additionally sends:

```ts
addedBy: item.addedBy,
addedByUser: item.addedByUser ? { id, username, avatarUrl } : undefined,
tickedBy: item.tickedBy?.filter((id): id is string => id !== null),
suggested: item.suggested,
```

All four are on the shared item type (`packages/shared/queue/src/types.ts:61-64`), and the backend does not synthesise them — there is no `addedBy` assignment anywhere in `packages/backend/src/graphql/resolvers/queue/`, so whatever the client omits stays null.

Web renders the attribution: `packages/web/app/components/queue-control/queue-climb-list-item.tsx:165-167` draws an avatar with a username tooltip straight from `item.addedByUser`.

Mobile never populates the fields locally either — `git grep addedBy packages/mobile/src` finds only test fixtures and one comment.

## Failure scenario

Two climbers share a session, one on the web app and one on the phone. Every climb the phone adds appears in the web queue list with a blank trailing slot where the other members' avatars are, so the web user can't tell who queued what. Climbs added from web show their avatar normally, which makes the gap look like a rendering bug rather than a missing field.

Second-order: per the flapping rule in `docs/websocket-implementation.md:133`, a field one client omits gets pushed back to everyone on that client's next full-queue write — so a mobile `setQueue` can also clear `addedByUser` off items a web peer had added.

## Suggested fix

Populate `addedBy` / `addedByUser` on mobile when building a queue item (the adding user's own identity is available from the auth store) and pass all four fields through `toQueueItemInput`. Then fold the mapper into the shared layer so the two platforms can't drift again — `packages/shared/queue-react/src/create-queue-mutations.ts:65` still describes mobile as sending a "thin `{uuid,climb}`", which is the shape this issue is about.

Small and self-contained; no schema change needed (the input fields already exist).

marcodejongh:
## What / why

A climb queued from the phone reached everyone else in the session anonymous — no name, no avatar in the queue row's trailing slot. Worse, once a phone was in the session it _stripped_ the attribution off climbs the crew had queued from web, so the avatars blinked in and out depending on who touched the queue last.

## Verified root cause

Two halves, both on mobile.

**Write side.** `packages/mobile/src/providers/queue-provider.tsx` bound the shared mutation core's one item→wire seam to a thin literal:

```ts
toQueueItemInput: (item) => ({ uuid: item.uuid, climb: toClimbInput(item.climb) }),
```

That seam serves every mobile write path in `packages/shared/queue-react/src/create-queue-mutations.ts` (add / setCurrent / setQueue / replace — `mirrorCurrentClimb` takes only a boolean and carries no item). So `addedBy`, `addedByUser`, `tickedBy` and `suggested` never left the phone, even though the GraphQL `ClimbQueueItemInput` declares all four, the backend's Zod shape accepts them, and the backend stores whatever arrives verbatim (it never synthesises attribution). Web has always sent all four and renders the avatar from `item.addedByUser`. Mobile also never populated them locally — `climbToQueueItem` set only `uuid` / `suggested` / `climb`.

**Read side.** Mobile maintains its _own_ subscription and resync selection sets, and they selected `{ uuid climb { … } }` too — no `addedBy` / `addedByUser` / `tickedBy` / `suggested`. So the phone rebuilt every peer item without attribution, and its next full-queue write pushed that gap back to the whole crew. This is exactly the flap `docs/websocket-implementation.md` and the #3927 contract test warn about, one level up from the climb.

## Fix

**Shared, so the two platforms cannot drift again** — `packages/shared/queue-react/src/queue-item-input.ts` owns the item-level field list for both clients. Pure TS, no React; it lives in `queue-react` because that package already owns the mutation wire contract and already depends on shared-schema (`@boardsesh/queue` deliberately has zero dependencies and cannot name `ClimbQueueItemInput`). It's a pure passthrough with a `satisfies Record<keyof ClimbQueueItemInput, true>` guard, exported on a dedicated `./queue-item-input` subpath — load-bearing, because nine mobile provider suites whole-module-mock `@boardsesh/queue-react` and `climb-to-queue-item.ts` sits in all nine module graphs.

**Write side.** Mobile's `toQueueItemWireInput` delegates to it; the provider binds `toQueueItemInput` to that. Identity is stamped at item _construction_, not at the wire, in the three funnels that introduce an item (`addToQueue`, `dispatchSetCurrent`, `setQueue`). `attributeNewItem` skips an item that already carries attribution and skips anything already in `stateRef.current.queue` — a wire-level `??` fallback would have claimed authorship of every unattributed peer item on the next whole-queue write. `setQueue` stamps the array _before_ the local dispatch and feeds the same array to both the reducer and the broadcast, so state and wire agree. Stamping needs both a party-profile id and a non-empty username, so an anonymous phone never pushes a blank-named avatar at peers.

**Read side.** `SUBSCRIPTION_QUEUE_ITEM_FIELDS` in `packages/mobile/src/lib/graphql/operations.ts` selects all four, applied at all six item-selection sites (FullSync queue + current, `QueueItemAdded.addedItem`, `CurrentClimbChanged.currentItem`, and both inside `GET_SESSION_QUEUE_STATE`). `toClimbQueueItem` rebuilds them, narrowing server nulls to `undefined` for the three fields the reducer types optional-not-nullable.

**Web** keeps byte-identical output: the inline climb literal is extracted to a local `toClimbInput` — keeping web's own `description: item.climb.description || ''` verbatim, which mobile's mapper does _not_ do — and `toClimbQueueItemInput` becomes a one-line delegation.

Out of scope on purpose: `packages/mobile/src/providers/bluetooth-provider.tsx:689` still builds a thin `{ uuid, climb }` input for the board-presence `reportBoardClimb` mutation. Different mutation, different consumer (wall presence, kiosk-redacted) — don't read the remaining literal as a missed spot. Mobile still doesn't _render_ an "added by" avatar in its own queue list; that's read-side UI on a different surface and a separate PR. The Swift Live Activity / widget path is also left alone: `packages/mobile/modules/live-activity/ios/SessionWebSocketManager.swift:363` builds its own attribution-less input and `NATIVE_IOS_QUEUE_UPDATES` deliberately reads a slim subset. Harmless — that call omits `shouldAddToQueue`, so the backend never appends a queue row for it and the reducer never rewrites an existing one; only the current-climb _pointer_ is unattributed, not the queue rows the avatars render from. The new backend describe says so in a comment.

## Tests

Every new test runs the real mapper / real provider — none rebuilds the predicate it checks.

- **`packages/shared/queue-react/src/__tests__/queue-item-input.test.ts`** (new) — the four fields survive; null entries drop from `tickedBy` (`[String!]`); an absent `addedBy` stays `undefined` rather than coercing to `null`; output keys equal `WIRE_ITEM_FIELDS`. _Reverting the mapper to `{uuid, climb}` fails all four; adding a seventh GraphQL field without teaching the mapper fails typecheck on the `satisfies` guard._
- **`packages/mobile/src/lib/__tests__/climb-to-queue-item.test.ts`** — `toQueueItemWireInput` carries all four AND still emits the full `toClimbInput` climb key set. _The pre-fix shape emits two keys and fails immediately._
- **`packages/mobile/src/lib/__tests__/queue-conversion.test.ts`** — a web peer's attribution survives the rebuild; nulls narrow to `undefined`; the rebuilt key set equals what `SUBSCRIPTION_QUEUE_ITEM_FIELDS` selects (parsed with graphql's own parser, not string-split). _Deleting the four rebuild lines fails._
- **`packages/mobile/src/providers/__tests__/queue-provider-attribution.test.tsx`** (new) — end-to-end through the real provider, with `@boardsesh/queue-react` mocked via `importOriginal` so the real mapper survives. Stamps this climber on a phone-queued climb (local state _and_ broadcast); does **not** re-author a peer item delivered over the queueUpdates subscription when the phone rewrites the whole queue; forwards an already-attributed item handed straight to `setQueue`; does not re-author an unattributed item already in this device's queue; the captured wire mapper serialises the stamped fields. _Mutation-tested both ways: disabling the stamping reds 3 of 5; removing the two misattribution guards reds a different 3 of 5; dropping the four fields from `toClimbQueueItem` reds the peer case._
- **`packages/backend/src/__tests__/queue-climb-field-contract.test.ts`** — new item-level parity describe: both read paths (shared `QUEUE_ITEM_FIELDS`, mobile `SUBSCRIPTION_QUEUE_ITEM_FIELDS` read live out of source) must select exactly what `ClimbQueueItemInput` declares. Plus an interpolation-count assertion, because the parity check unions field names across every item selection in a document, so one complete site would mask a longhand sibling. _Deleting `suggested` from the mobile template reds the parity check; writing one of the six selection sites out longhand reds the interpolation check while parity stays green — which is exactly the blind spot it closes._
- **`packages/backend/src/__tests__/operations-schema-validation.test.ts`** — taught to resolve the nested `${SUBSCRIPTION_QUEUE_ITEM_FIELDS}` placeholder (factored into one helper so a future nested const can't be taught to one call site and forgotten at another). This is what proves mobile's new selection set parses and validates against the executable schema.
- **`packages/web/app/components/persistent-session/__tests__/to-climb-queue-item-input.test.ts`** — pins the delegation to zero behaviour change: all four item fields, `tickedBy` nulls filtered, absent `addedBy` stays `undefined`, and `description: ''` for a null-description climb. _A refactor that repointed web at mobile's `toClimbInput` reds the description case._
- **`packages/mobile/src/providers/__tests__/queue-provider.test.ts`** — dropped `produces a valid ClimbQueueItem shape without extra properties`. It hand-transcribed the top-level key list as `['uuid', 'climb']`, which the four rebuilt attribution fields now break. It's superseded by the live-derived assertion in `queue-conversion.test.ts` that parses `SUBSCRIPTION_QUEUE_ITEM_FIELDS` instead of restating it, so the guard survives the next field added.

## Validation

- `vp check` — 0 errors, 288 warnings (all pre-existing; none in touched files)
- `vp run typecheck` — clean across every scope
- `vp run check:mobile-bundle` — OK (Metro resolves the new subpath export)
- `vp test run --project queue-react` — 43 passed
- `vp run test:mobile` — the whole mobile package, 563 files / 4765 tests passed (a `queue-provider-*` glob misses `queue-provider.test.ts`, so the full run is the honest gate here)
- backend: `queue-climb-field-contract` 10 passed, `operations-schema-validation` 223 passed
- web: `to-climb-queue-item-input` 9 passed

## QA Notes

1. **Misattribution.** Two devices, one session. Web peer queues climb A. On the phone, trigger a whole-queue write (activate a playlist, generate a session, drag-reorder). Climb A must still show the _web_ user's avatar on web afterwards, not the phone user's. Then the reverse: a climb the phone added shows the phone user.
2. **The flap.** Add climbs from both clients, then force a reconnect/FullSync on web. Attribution must be stable across the round trip — no avatars blinking depending on who wrote last.
3. **Signed-out phone.** Stamping requires both a party-profile id and a non-empty username. Verify no crash and no empty-initial avatar on web when the phone is anonymous.
4. **Solo mode.** No session: `addedBy` resolves from `sessionRuntimeStateRef.current.clientId`, which is `''` before a join (stored as `null`). Confirm queueing solo still works, the local snapshot round-trips, and starting a session afterwards seeds the queue with attribution intact.
5. **`suggested` pruning.** Mobile now both sends _and_ receives `suggested`, so `pruneSuggestedQueueItemsAfterCurrent` can fire on peer-originated suggestion items during a local playlist activation. On web start a playlist-suggestion flow, then have the phone activate a playlist climb; confirm the web queue doesn't silently swallow items it should keep.
6. **Mobile queue list** still shows no "added by" avatar of its own — that's unchanged and out of scope.

OTA-safe: JS/TS only, no native deps, no schema change, no migration.

## Release Notes

Climbs you queue from the phone now show your name and avatar to everyone in the session instead of landing anonymous — and a phone in the session no longer wipes the "added by" avatars off the climbs your crew queued from the web.

Fixes #3995

### DIFF
--- a/docs/websocket-implementation.md
+++ b/docs/websocket-implementation.md
@@ -136,6 +136,12 @@ Two non-obvious narrowing points bit us in #3927: `climbToQueueItem` hand-picked
 
 The contract is enforced by `packages/backend/src/__tests__/queue-climb-field-contract.test.ts`, which reads every list from its live source and compares by set equality against the GraphQL `ClimbInput` type. `userAscents` / `userAttempts` are the only deliberate exception — they are per-user tick counts and must never be broadcast, at the cost of behaving exactly like the bug the test guards against.
 
+### Queue item field contract
+
+The **item** wrapping that climb flaps for the same reason (#3995). `ClimbQueueItemInput` carries `addedBy`, `addedByUser`, `tickedBy` and `suggested` alongside `uuid`/`climb`, and each crosses its own write list (`toQueueItemInput` in `@boardsesh/queue-react`), the backend's Zod item schema, and each client's read selection (`QUEUE_ITEM_FIELDS` in `@boardsesh/graphql`, `SUBSCRIPTION_QUEUE_ITEM_FIELDS` on mobile, plus mobile's `toClimbQueueItem` rebuild). Item drift is louder than climb drift because attribution is rendered: the queue row's trailing avatar blinks in and out depending on who wrote last. Before #3995 mobile neither sent nor selected the four, so a phone in the session stripped attribution from climbs the crew had queued on web. The same contract test guards the item lists, by set equality against `ClimbQueueItemInput`.
+
+The note above about the adapter populating `addedBy`/`addedByUser` from `usePartyProfile` describes **web**. Mobile stamps its own identity in `attributeNewItem` (`queue-provider.tsx`), which deliberately skips an item that already carries attribution and anything already sitting in this device's queue — a blanket wire-level fallback would let a phone claim authorship of a peer's item on its next full-queue write.
+
 ## Technology Stack
 
 | Component          | Technology                        | Purpose                                                       |
--- a/packages/backend/src/__tests__/operations-schema-validation.test.ts
+++ b/packages/backend/src/__tests__/operations-schema-validation.test.ts
@@ -58,6 +58,22 @@ function readMobileSubscriptionClimbFields(): string {
   return extractTemplateConst(readMobileOperationsSource(), 'SUBSCRIPTION_CLIMB_FIELDS');
 }
 
+/**
+ * Mobile's operation strings interpolate `${SUBSCRIPTION_QUEUE_ITEM_FIELDS}` (which
+ * itself interpolates `${SUBSCRIPTION_CLIMB_FIELDS}`) at build time via JS template
+ * literals. The regex extraction above reads raw source text, so it captures those
+ * placeholders literally — substitute them, outer one first, to reproduce what the
+ * real template literal evaluates to before parsing as GraphQL. One helper so a new
+ * nested selection const cannot be taught to one call site and forgotten at another.
+ */
+function resolveMobileTemplatePlaceholders(rawSource: string, mobileOperationsSource: string): string {
+  const queueItemFields = extractTemplateConst(mobileOperationsSource, 'SUBSCRIPTION_QUEUE_ITEM_FIELDS');
+  const climbFields = extractTemplateConst(mobileOperationsSource, 'SUBSCRIPTION_CLIMB_FIELDS');
+  return rawSource
+    .replaceAll('${SUBSCRIPTION_QUEUE_ITEM_FIELDS}', queueItemFields)
+    .replaceAll('${SUBSCRIPTION_CLIMB_FIELDS}', climbFields);
+}
+
 // Collect the field names selected inside every `climb { ... }` selection set of
 // a GraphQL operation document. Both the shared queue-session subscription and a
 // wrapped copy of the mobile fragment feed through this, so parity is compared
@@ -158,16 +174,11 @@ describe('shared-schema operations validate against the executable schema', () =
  */
 describe('mobile plain-string subscription documents validate against the executable schema', () => {
   const mobileOperationsSource = readMobileOperationsSource();
-  const subscriptionClimbFields = extractTemplateConst(mobileOperationsSource, 'SUBSCRIPTION_CLIMB_FIELDS');
-
-  // QUEUE_UPDATES_SUBSCRIPTION interpolates `${SUBSCRIPTION_CLIMB_FIELDS}` at build
-  // time via a JS template literal. The regex extraction below reads raw source text,
-  // so it captures that placeholder literally — substitute the extracted field list
-  // in its place to reproduce what the real template literal evaluates to before
-  // parsing as GraphQL.
   function readMobilePlainOperation(constName: string): string {
-    const rawSource = extractTemplateConst(mobileOperationsSource, constName);
-    return rawSource.replaceAll('${SUBSCRIPTION_CLIMB_FIELDS}', subscriptionClimbFields);
+    return resolveMobileTemplatePlaceholders(
+      extractTemplateConst(mobileOperationsSource, constName),
+      mobileOperationsSource,
+    );
   }
 
   const mobilePlainOperationNames = [
@@ -291,14 +302,9 @@ describe('mobile QUEUE_UPDATES_SUBSCRIPTION selects the same PlaybackStateChange
   }
 
   const mobileOperationsSource = readMobileOperationsSource();
-  const subscriptionClimbFields = extractTemplateConst(mobileOperationsSource, 'SUBSCRIPTION_CLIMB_FIELDS');
-  // Same placeholder-substitution technique as the "mobile plain-string
-  // subscription documents" describe block above — QUEUE_UPDATES_SUBSCRIPTION
-  // interpolates `${SUBSCRIPTION_CLIMB_FIELDS}` at build time, so replicate
-  // that before parsing raw extracted source text as GraphQL.
-  const mobileQueueUpdates = extractTemplateConst(mobileOperationsSource, 'QUEUE_UPDATES_SUBSCRIPTION').replaceAll(
-    '${SUBSCRIPTION_CLIMB_FIELDS}',
-    subscriptionClimbFields,
+  const mobileQueueUpdates = resolveMobileTemplatePlaceholders(
+    extractTemplateConst(mobileOperationsSource, 'QUEUE_UPDATES_SUBSCRIPTION'),
+    mobileOperationsSource,
   );
 
   const sharedPlaybackFields = inlineFragmentFieldNames(queueSessionOperations.QUEUE_UPDATES, 'PlaybackStateChanged');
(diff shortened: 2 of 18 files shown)


marcodejongh:
Review follow-ups:

1. **native-fingerprint label** — was staleness, not a native change. The branch was behind `main` when the fingerprint ran, so it compared against a moved tree. Rebased; the check now reports `✅ Ships over-the-air` on both platforms with fingerprints matching `main`, and the label is gone. The "OTA-safe" line in the description stands.
2. **`addedBy` staleness window** — comment added at `selfAttributionRef` naming the render window: a stamp landing before the post-join re-render carries `addedBy: null` with `addedByUser` populated, peers render the avatar off `addedByUser`, so only the legacy clientId field is absent.
3. **Redundant wrapper** — collapsed to `toQueueItemInput: toQueueItemWireInput`.
4. **Peer suggestion pruning** — not adding a test here on purpose. `pruneSuggestedQueueItemsAfterCurrent` is shared reducer logic with its own suite (`packages/shared/queue/src/__tests__/playlist-suggestions.test.ts`), and this PR does not change it — it makes mobile *converge* on web's long-standing behaviour by stopping mobile from silently dropping `suggested` off peer items. What is new is that mobile now carries the field, and that is asserted in `queue-conversion.test.ts` against the live selection set. QA note 5 covers the two-client flow on device.

Issue #3978, opened 2026-07-28
vp run db:up fails: dev-db image has 0187_sad_freak's schema but not its ledger row


marcodejongh:
## Summary

`vp run db:up` aborts on a fresh machine. `scripts/dev-db-up.sh` tries to apply `0187_sad_freak` against `ghcr.io/boardsesh/boardsesh-dev-db:latest` and dies:

```
relation "board_climb_ingest_skips" already exists
```

The image contains 0187's schema but **not** its row in `drizzle.__drizzle_migrations`, so the applier retries it on every run and fails every time.

## Evidence

`vp run db:verify-journal` (added in #3977) against the local dev DB:

```
🔎 Verifying migration journal against: 127.0.0.1:5432 (read-only)
❌ 1 of 188 journal migrations have no row in drizzle.__drizzle_migrations (189 rows present).
   • 0187_sad_freak
```

188 journal entries, 189 ledger rows, 187 hashes matched. The two extra rows are benign renumber residue; the one gap is real.

## Why it matters beyond local setup

This is the reason #3977 keeps the per-entry journal check behind `VERIFY_MIGRATION_JOURNAL=1` instead of making it the default. A default-on check would redden `vp run db:migrate` for every developer until the image is self-consistent.

Sequencing agreed in #3977: land the check gated → rebuild the dev-db image so its ledger matches its schema → flip the default in a follow-up.

## Fix

Rebuild and republish `boardsesh-dev-db` so its ledger row set matches the schema it ships (or insert 0187's ledger row into the image build). Then flip `VERIFY_MIGRATION_JOURNAL` on by default in `packages/db/scripts/migrate.ts`.

Found while implementing #2933.

marcodejongh:
🤖 Claimed by the 2026-07-31 bug-fix campaign (Fable session `fablebugfix`, batch 2). Assigned to avoid duplicate work across concurrent agent campaigns — a PR from this session will follow. If another agent needs this issue, comment here first.

marcodejongh:
🤖 Claimed by bug-batch loop `claude-bb21` (batch 3, session c321d1b9, 2026-08-11). Where a prior 2026-07-31 claim exists it has lapsed (10+ days, no PR/branch) — reclaiming per protocol. A PR from this loop will follow; parallel loops please pick another issue.

PR #4052, opened 2026-07-31
fix(db): repair the 8 moonboard stats rows violating the ascent-count invariant

### LINKED ISSUE #3534: 8 moonboard stats rows violate the materialized count invariant (total != upstream + boardsesh)
## Summary
8 moonboard board_climb_stats rows have ascensionist_count inconsistent with COALESCE(upstream,0)+COALESCE(boardsesh,0) - e.g. 321b5952-6516-55af-acf7-fe26f9ae9783 angle 40: total 64, upstream 3, boardsesh NULL (sum across the 8: total 164 vs expected 24). A moonboard import path wrote ascensionist_count directly without maintaining the component columns; the next recomputeClimbStats on those (climb,angle) keys will silently REWRITE the total down to upstream+boardsesh (150-151), collapsing 64 -> 3 - so the data is wrong now and will lurch to a different wrong value later. Every other board is invariant-clean (0 violations across 891k rows).

## Failure scenario
Climb 321b5952.../40 shows 64 ascents today; the first Boardsesh tick on it triggers recompute and the count drops overnight to ~4, looking like data loss to users watching the climb.

## Evidence
`packages/backend/src/graphql/resolvers/ticks/recompute-climb-stats.ts:150`

```
SQL: SELECT board_type, count(*), sum(ascensionist_count), sum(COALESCE(upstream_ascensionist_count,0)+COALESCE(boardsesh_ascensionist_count,0)) FROM board_climb_stats WHERE ascensionist_count IS DISTINCT FROM COALESCE(upstream_ascensionist_count,0)+COALESCE(boardsesh_ascensionist_count,0) GROUP BY 1 -> only moonboard: 8 rows, 164 vs 24.
```

## Suggested fix
Decide which side is right (probably ascensionist_count from the Moon catalog): backfill upstream_ascensionist_count = ascensionist_count - COALESCE(boardsesh,0) for the 8 rows, and fix the moonboard stats writer to set upstream + total together like the other writers.

marcodejongh:
## ⚠️ This migration writes production data

Migration `0188_backfill_moonboard_ascensionist_invariant.sql` is a data-only repair that `UPDATE`s 8 rows in prod `board_climb_stats`. It is fenced twice — an explicit 8-row PK list AND the live invariant-violation predicate — but it is still a production data write, and it should be reviewed as one.

## What's broken

8 MoonBoard `board_climb_stats` rows (all `board_type='moonboard'`, `angle=40`) violate the materialized invariant `ascensionist_count = COALESCE(upstream,0) + COALESCE(boardsesh,0)` (#3534). Totals sum to 164; `upstream` sums to 24. `boardsesh_ascensionist_count` is NULL on all 8.

The harm is latent, not visible: the first Boardsesh tick on any of these keys runs `recomputeClimbStats`, which rewrites `ascensionist_count := COALESCE(upstream,0) + COALESCE(boardsesh,0)` in the same statement. Climb `321b5952…/40` would go from 64 ascents to 4.

## Direction: settled, direction 1 (raise `upstream`; the total stands)

This PR previously sat as a draft because a writer audit against current `main` could not find any writer capable of producing `total > upstream`, which left the repair direction contested. That audit was wrong — it only looked at current `main`. Reading git history finds the writer. Full evidence: https://github.com/boardsesh/boardsesh/pull/4052#issuecomment-5154779315

Short version:

- **2026-02-08 (`2fe4d54e9`)** — `import-moonboard-problems.ts` inserted these problems with `ascensionistCount := problem.repeats`, the small Feb-era values.
- **2026-05-15 (migration 0099, `b1d8089da`)** — the source-column split seeded `aurora_ascensionist_count := ascensionist_count`, freezing `upstream` at those Feb repeats (4, 3, 0, 2, 4, 2, 7, 2 — sum 24).
- **2026-06-26 → 2026-07-05 (`import-moonboard-catalog.ts` at `377babf4b`)** — its upsert did `ascensionist_count := greatest(coalesce(excluded,0), coalesce(existing,0))` with no write to `upstream_ascensionist_count` and no `upstream_synced_at` stamp. It matched these rows on fingerprint and raised the total to the June catalog community counts (26, 64, 3, 7, 17, 20, 24, 3 — sum 164), leaving `upstream` behind. The importer was fixed on 2026-07-05 in `454cfd4aa`.
- **2026-07-05 (migration 0141, same commit `454cfd4aa`)** — its recover step `upstream := GREATEST(ascensionist_count - COALESCE(boardsesh,0), 0)` is gated on `upstream_ascensionist_count IS NULL`. These 8 rows carry a non-NULL 0099-seeded upstream, so 0141 skipped exactly them. 0141's own header names non-NULL-upstream rows as the set it leaves alone.

So the total is the current MoonBoard catalog number and `upstream` is a February leftover. This migration runs 0141's already-shipped recover formula over the rows 0141's NULL guard missed. `ascensionist_count` is not touched, so no UI-visible number changes.

Because the fault was in a writer that is already fixed (`454cfd4aa`), no code guard ships with this migration.

**Timing**: if a first tick lands on one of these 8 keys before this migration applies, that recompute collapses the row to stale upstream + 1, this migration then correctly no-ops on it, and the count is already gone for that row. A reason to land it promptly.

Fixes #3534

## QA Notes
- Migration is data-only, idempotent, and double-fenced: explicit 8-row PK list AND the live invariant-violation predicate, so a row corrected by any later recompute is untouched.
- No app code changes; no automated test (data-only migration precedent: 0157/0163/0186). Verification = run the fenced SELECT before/after.
- The 8 PKs and the before/after values are in the migration header.

## Release Notes


marcodejongh:
Taking over this conflicted draft under the active bug-batch loop, per the coordination rule. I will preserve the investigation/attribution but rebase onto current `main` and repurpose the PR rather than merge the current data migration.

Accepted plan after current-writer audit:
- remove the hand-authored migration/journal/snapshot changes;
- add tested catalog-upsert `setWhere` guards so unchanged MoonBoard rows do not churn `updated_at`/`sync_seq`;
- stop the deprecated 2024 importer from stamping `upstream_synced_at` without source count data;
- add a hard-fenced, dry-run-by-default Drizzle maintenance command for the eight rows, with exact-before compare-and-swap and idempotency checks;
- coordinate #3849 so merged components derive the total instead of independently `GREATEST`-merging all three.

No production write will run from this takeover. Fresh catalog evidence must decide total-vs-components authority before any explicit `--apply` use.

marcodejongh:

marcodejongh:
🤖 Independent review from the `josh` loop: **HOLD — two blockers, and one bot comment on this PR is actively misleading.**

1. **Don't trust the renumber bot's success comment above.** Run 30630207599 renumbered locally but its proof-of-apply step failed (CI dev-db image is missing the ledger row for 0188_puzzling_kabuki — the #3978 defect class), so NOTHING was pushed: branch head is still the original commit, still numbered 0188, which now collides with main's 0188_puzzling_kabuki. The 12:22Z '🔢 Renumbered → 0189' comment is stale/false. Until the dev-db ledger gap is fixed, every renumber attempt on every branch will fail the same way.
2. **The repair-direction question the PR body itself raises (raise upstream vs lower total) is still unanswered** — the board_climb_stats_history check it asks for hasn't been run by anyone.
3. Release-notes gate is red: body has a bare `none`, which the checker strips then fails as empty — needs the ticked checkbox or skip-changelog label.

For whoever completes the takeover: the migration's SQL logic itself verified sound — the SET derives corrected values live from the row's own columns (no hardcoded integers), double-fenced WHERE makes it idempotent and scoped to exactly the violating rows, and I code-verified that every current live writer (recompute.ts both paths, both moonboard importers) is invariant-preserving by construction, so this genuinely is legacy residue, not an active bug. MoonBoard's upstream/boardsesh split is a disjoint partition (double-count guard excludes upstream-counted users), so summing is correct — not the kilter/aurora same-ascents anti-pattern. No rollback SQL exists anywhere; reconstruct from the header's before-value table if needed.

marcodejongh:
## The repair direction is settled — direction 1, on new evidence

The blocker on this PR was that no writer could be found capable of producing `total > upstream`, which undercut the premise of the repair as drafted. **That audit was wrong.** An independent Fable review went through git history rather than current main and found the writer.

### The retired writer

`packages/db/scripts/import-moonboard-catalog.ts` as of `377babf4b` ("feat(db): import the full MoonBoard catalog (all 7 boards) (#3214)", 2026-06-26), replaced by `454cfd4aa` (2026-07-05). For those nine days its stats upsert `ON CONFLICT` set:

```
ascensionistCount: sql`greatest(coalesce(excluded.ascensionist_count, 0), coalesce(${boardClimbStats.ascensionistCount}, 0))`
```

No write to `aurora_ascensionist_count` — a column that had existed since 0099 (2026-05-15) — and no `upstream_synced_at` stamp. Every catalog run in that window GREATEST-raised the total to the fresh MoonBoard community count while leaving the source column alone. That is precisely the writer the earlier audit concluded did not exist.

### The full chain, and every observable it explains

1. **2026-02-08** (`2fe4d54e9`) — the original `import-moonboard-problems.ts` inserts these problems with `ascensionistCount: problem.repeats`, small Feb-era values, deterministic uuidv5 from apiId.
2. **2026-05-15** (0099, `b1d8089da`) — the split migration seeds `aurora_ascensionist_count := ascensionist_count`, so these rows get upstream = the Feb repeats: 4, 3, 0, 2, 4, 2, 7, 2 — **sum 24**. Worth noting 0141's recover step could *not* have produced these numbers: it sets `upstream := total − boardsesh`, which by July would have yielded the big totals. Only the May seed at Feb-era totals produces them, which independently proves these rows predate 0099.
3. **2026-06-26 → 07-05** — the retired importer fingerprint-matches the existing rows in place (its own header: "a hit merges it in place, backfilling the 2024 quality/ascensionist gap") and GREATEST-raises the total to the June catalog counts: 26, 64, 3, 7, 17, 20, 24, 3 — **sum 164**. Upstream untouched.
4. **2026-07-05** (0141, in the same commit that fixed the importer) — the recover step `upstream := GREATEST(total − boardsesh, 0)` is gated on `upstream_ascensionist_count IS NULL`. These 8 rows have non-NULL 0099-seeded upstream, so they are **skipped**. That is the blind spot, and 0141's own header predicts this residue class: "climbs the post-deploy re-import can't reach (e.g. MoonBoard problems delisted from the app catalog)".
5. Post-fix re-imports never reach them, so `upstream_synced_at` is still NULL — matches prod.
6. No tick ever landed, so 0099-backfill, 0157, 0162, 0163 and the live recompute (all keyed on ticked keys) never touched them — `boardsesh_ascensionist_count` still NULL, matches prod.

### Why this rules for direction 1

The two sides of the violation are two snapshots of **the same manufacturer quantity at different times**: the total is the June 2026 catalog community count, upstream is the stale February value. The Boardsesh side contributed nothing (NULL, zero ticks), so there is no mechanism in history by which the total could contain a double-fold — every merging writer either rewrote the total *from* upstream in the same statement (0157/0162/0163/recompute/aurora/kilter sync) or GREATEST-merged it against a manufacturer-sourced number.

Direction 2 would regress 8 climbs' community counts to their February values and freeze them there permanently, since no upstream resync reaches these keys.

Direction 1's formula `upstream := total − COALESCE(boardsesh, 0)` is **literally 0141's already-shipped recover formula**, extended to the rows its NULL-guard missed. The migration as drafted is right.

### One correction to the verification plan

The PR body asks for a `board_climb_stats_history` check on the 8 keys. That will come back **inconclusive by construction** — the snapshot job only began around 2026-07-08 (`c989daa22`), after the damage window closed on 07-05. Flat or empty history is expected and must not be read as support for direction 2.

The actual falsifier is `created_at` on the climbs:

```sql
SELECT uuid, name, created_at, is_listed
FROM board_climbs
WHERE board_type = 'moonboard'
  AND uuid IN ( … the 8 … )
ORDER BY created_at;
```

All eight should predate 2026-05-15. A `created_at` after that date on any row breaks the reconstruction for that row (post-0099 rows had NULL upstream and were repaired by 0141), and would be the signal to stop rather than ship.

### Header corrections needed before merge

- **Lines 19–20** — "climb 321b5952…/40 would drop from 64 to 0(+1 tick)" contradicts the before-table at line 34 (`64, 3, NULL`). The real harm is a collapse from 64 to **4** (3 stale upstream + 1 tick). And "until-now absent" is wrong: upstream is *present but stale* on all 8 rows — one is 0, none are NULL.
- **Lines 22–28** — replace the "ground truth" assumption with the established provenance above, and say plainly that this extends 0141's recover step to the rows its `upstream IS NULL` guard skipped.

Also still live: the 0188 collision with main's `0188_puzzling_kabuki` (the renumber bot's "→ 0189" comment is false — nothing was ever pushed, head is still `c5d508f62`), and the red `release-notes` gate, which needs the ticked "No release note needed" checkbox rather than a bare `none`, plus a push to re-trigger. Both are being handled as part of stacking this PR onto the merge train.

The SQL logic itself passes: idempotent, double-fenced, and ordinal-independent (references only columns present since 0141, no self-number reference), so it is safe anywhere in the train.

marcodejongh:
## Maintainer sign-off recorded (2026-08-02)

**Marco has signed off on direction 1** — raise `upstream_ascensionist_count`; the total stands — on the git-history evidence in the comment above. The migration ships as drafted.

To state the authorisation plainly, since this writes production data: the approved scope is an UPDATE of `upstream_ascensionist_count` on the 8 listed MoonBoard/angle-40 `board_climb_stats` rows, double-fenced by the explicit PK list and the live invariant-violation predicate. `ascensionist_count` itself is unchanged on every row, so no UI-visible number moves. Nothing is deleted.

Remaining work before this can merge, all being handled as part of today's stacked merge train:

- the 0188 collision with main's `0188_puzzling_kabuki` (and the stale `when`, which would otherwise make the migration skip permanently)
- the red `release-notes` gate, which needs the ticked checkbox form rather than a bare `none`
- header corrections, now pushed

Note the CI on this PR currently shows no check runs at all. That is not a dropped event — the branch conflicts with main, so GitHub cannot compute a merge ref and never creates the `pull_request` runs. The rebase is the only thing that will bring CI back; a manual dispatch cannot validate `release-notes`, which is gated on `github.event_name == 'pull_request'`.

marcodejongh:
Stacked-base update: #4025 force-pushed its independently approved reviewer fix to 8e2ed1c9ecfbe275d2d2498d2deb5f7495765cdd. Because this PR is stacked through #4058, both downstream branches need rebasing before this migration diff or CI can be trusted. No takeover is intended; please keep this draft until the train is restabilized.

marcodejongh:
Landed on main via the #4054 stack (squash 50260079) as 0191_backfill_moonboard_ascensionist_invariant. Closing as shipped.

Issue #4042, opened 2026-07-31
Party-session self-adds are double counted: QueueItemAdded echo is tracked as addedFromTab: 'peer_broadcast'


marcodejongh:
`QueueItemAdded` has the identical shape of bug that #3382 fixes for removes, and #3382 deliberately scoped it out to keep that diff surgical.

## What happens

In a party session, a locally-initiated add fires `Climb Added to Queue` once at the mutation site with the real `addedFromTab`:

- web: `packages/web/app/components/graphql-queue/QueueContext.tsx` (`onQueueItemAdded`)
- mobile: `packages/mobile/src/providers/queue-provider.tsx`

The server then echoes the same `QueueItemAdded` back to the originator, and the echo handler fires a **second** event, hardcoded to `addedFromTab: 'peer_broadcast'`:

- web: `packages/web/app/components/persistent-session/hooks/use-peer-broadcast-analytics.ts:66-72`
- mobile: `packages/mobile/src/providers/queue/use-session-realtime.ts:548-555`

So every party-session self-add emits two `Climb Added to Queue` events, one of them attributed to a peer who did nothing.

## Fix shape

Mechanically identical to #3382:

1. Add a nullable `clientId: ID` to `type QueueItemAdded` in `packages/shared-schema/src/schema/events.ts` (+ the two hand-written unions in `src/types/events.ts`).
2. Publish `clientId: ctx.connectionId || null` from the `addQueueItem` publish block in `packages/backend/src/graphql/resolvers/queue/mutations.ts`.
3. Select `clientId` on the `... on QueueItemAdded` fragment in `packages/shared/graphql/src/operations/queue-session.ts` (`QUEUE_UPDATES`) and `packages/mobile/src/lib/graphql/operations.ts`.
4. Widen `SubscriptionWireEnvelope`'s `QueueItemAdded` member in `packages/shared/queue-runtime/src/subscription-adapter.ts`.
5. Skip the analytics track in both echo handlers when `event.clientId` equals our own joinSession-returned clientId. Do **not** relabel it `'self'` — the local site already tracked that, so relabelling keeps the double count.
6. Regenerate with `vp run codegen`.

Identity note: compare against the **joinSession-returned** connection id (`sessionRuntimeStateRef.current.clientId` on mobile, `lifecycle.session?.clientId` on web). Mobile's `coordinator.clientId` is a locally generated uuid the server never sees — comparing against it produces a green but inert fix.

## Why it matters now

[...]

PR #4053, opened 2026-07-31
fix: stop double-counting party-session self-removes

### LINKED ISSUE #3382: Analytics: party-mode self-removes mislabeled removedBy: peer (QueueItemRemoved has no clientId)
Follow-up tracked from review of #3372.

## Problem

`usePeerBroadcastAnalytics` (`packages/web/app/components/persistent-session/hooks/use-peer-broadcast-analytics.ts:73-78`) unconditionally tags every `QueueItemRemoved` event with `removedBy: 'peer'`:

```ts
} else if (event.__typename === 'QueueItemRemoved') {
  track('Climb Removed from Queue', {
    boardLayout: boardLayoutNameRef.current,
    partyMode: true,
    removedBy: 'peer',
  });
}
```

`QueueItemRemoved` carries **no `clientId`** on the wire (only `CurrentClimbChanged` does — see the hook's own comment at line 41-45), so the client cannot tell its **own** remove echo apart from a genuine peer remove. In party mode, the queue subscription receives echoes of this client's own removes, so **every self-remove is attributed to `peer`**, corrupting the `removedBy` breakdown in analytics.

## Open question (verify)

The reviewer notes this may be a *new* regression from #3372: the old analytics hook lived **inside** the party-mode `GraphQLQueueProvider`, so it structurally only saw external peer events; the refactor moved it to the root `persistent-session` provider, where it now also sees this client's own echoes. The PR's inline comment instead claims the old hook already saw self-echoes ("mirrors the old hook's behavior exactly — it never filtered self vs peer either"). Worth confirming which is true — it decides severity (silent pre-existing skew vs. a fresh regression) but not whether it needs fixing.

## Fix direction

Add a `clientId` (or reuse the existing `correlationId`) to the `QueueItemRemoved` wire event — same mechanism `CurrentClimbChanged` already uses — so the client can filter its own remove echoes (or tag `removedBy: 'self'` vs `'peer'` accurately). Then gate the analytics `track` accordingly.

## Scope

- Wire schema: add `clientId`/`correlationId` to `QueueItemRemoved` (`packages/shared-schema`, backend publisher).
- `use-peer-broadcast-analytics.ts`: filter or correctly label self vs peer.
- `QueueItemAdded` likely has the same gap (`Climb Added to Queue` / `addedFromTab: 'peer_broadcast'`) — fix both together.

_Not blocking #3372 — that PR preserves the pre-existing (unfiltered) behavior; this corrects the attribution._

### LINKED ISSUE #3383: Peer-broadcast analytics mislabels self-removes as removedBy: 'peer' (no clientId on QueueItemRemoved)
## Summary

Peer-broadcast queue analytics tags **every** `QueueItemRemoved` event with `removedBy: 'peer'`, including echoes of the current client's own removes. Self-removes are therefore mis-attributed as peer removes, corrupting the "who removed this climb" breakdown in the `Climb Removed from Queue` event.

## Where

`packages/web/app/components/persistent-session/hooks/use-peer-broadcast-analytics.ts` — the `QueueItemRemoved` branch hardcodes `removedBy: 'peer'`.

## Root cause

The wire event `QueueItemRemoved` (`packages/shared-schema/src/schema/events.ts`) carries only `sequence`, `stateHash`, and `uuid` — there is **no `clientId`** (nor `correlationId`) on it. By contrast, `CurrentClimbChanged` carries `clientId` + `correlationId`, which the reducer uses to suppress this client's own echoes (`reducer.ts`, `DELTA_UPDATE_CURRENT_CLIMB`). Without an equivalent field on `QueueItemRemoved`, the analytics subscription cannot distinguish a self-echo from a genuine peer remove, so it labels all of them `peer`.

This mislabel predates W6 (the deleted board-route hook `use-queue-event-subscription.ts` never filtered self vs peer either), but W6 relocated the analytics to the always-mounted root, so this client's own remove echoes now flow through the same subscription and the mislabel is more visible.

## Fix direction

Add a `clientId` field to `QueueItemRemoved` (and populate it on the backend where the remove originates), mirroring `CurrentClimbChanged`'s echo-suppression pattern. Then in `use-peer-broadcast-analytics.ts` compare the event's `clientId` against this client's own id and tag `removedBy: 'self'` vs `'peer'` accordingly. Consider doing the same for `QueueItemAdded` (its `addedFromTab: 'peer_broadcast'` tag has the same self/peer ambiguity).

### Cheaper client-only alternative (needs product sign-off)

The root-level `removeQueueItem` mutation wrapper (`use-queue-mutations.ts`) is the single chokepoint every self-remove flows through. It could record the removed uuid into a short-lived, bounded ref that the root analytics hook consults (match-and-delete on the echo) to tag `removedBy: 'self'` without any wire change. This avoids a schema/backend change but introduces client-side lifecycle/leak-bounding concerns and changes attribution semantics — flagged for discussion rather than implemented.

## Context

Found during PR #3372 (W6: root persistent-session reducer becomes the only web queue state; peer-broadcast analytics moved to the root). A `// TODO(#<this-issue>)` marker is left at the `removedBy: 'peer'` site.

marcodejongh:
## What & why

In a party session, removing a climb emitted **two** `Climb Removed from Queue` events — one correct, one attributed to a peer who did nothing.

The local remove already tracked itself at the mutation site with `removedBy: 'self'` (`packages/web/app/components/graphql-queue/QueueContext.tsx:693` on web, `packages/mobile/src/providers/queue-provider.tsx:775-781` on mobile). The server then echoed the same remove back to the originator, and the echo handler fired a second event — hardcoded `removedBy: 'peer'` on both platforms (`use-peer-broadcast-analytics.ts`, which carried a literal `TODO(#3383)`, and `use-session-realtime.ts`).

## Verified root cause

`type QueueItemRemoved` carried no publisher identity — only `sequence`/`stateHash`/`stateHashOrdered`/`uuid` (`packages/shared-schema/src/schema/events.ts`), and the single backend publisher (`removeQueueItem`) sent none. With nothing to compare against, both echo handlers had to guess, and both guessed "peer".

So this is a **double count**, not just a mislabel. Relabelling the echo to `'self'` (the obvious-looking fix) would have kept the duplicate and made it invisible.

**Identity semantics.** The `clientId` the backend stamps on wire events is `ctx.connectionId` — the same value `joinSession` hands back to the client (`sessions/mutations.ts`, "`result.clientId` is the connection ID"). Mobile stores it in `sessionRuntimeState.clientId`; web in `lifecycle.session.clientId`. Mobile's `coordinator.clientId` is a locally generated uuid the server never sees — comparing against **that** would never match and would have produced a green, inert PR. The standing precedent is the hook this telemetry was relocated from, which passed `context: { myClientId: persistentSession.clientId ?? undefined }`.

**Answer to #3382's "Open question (verify)" — pre-existing skew, not a #3372 regression.** `git show 08cc81eed^:packages/web/app/components/graphql-queue/hooks/use-queue-event-subscription.ts` shows the deleted board-route hook subscribed to the identical `persistentSession.subscribeToQueueEvents` stream and hardcoded `removedBy: 'peer'` in its own `case 'QueueItemRemoved':` arm. The W6 relocation to the root changed nothing about self-echo exposure.

## Fix

- Additive nullable `clientId: ID` on `type QueueItemRemoved` (+ both hand-written unions in `shared-schema/src/types/events.ts`, optional so legacy/replayed payloads still typecheck).
- `removeQueueItem` publishes `clientId: ctx.connectionId || null`. The `|| null` coercion matters: two anonymous clients with `''` would otherwise echo-suppress each other.
- `clientId` selected on `QUEUE_UPDATES` (shared) and `QUEUE_UPDATES_SUBSCRIPTION` (mobile); `SubscriptionWireEnvelope` widened.
- Both echo handlers skip the analytics track when `event.clientId` matches our own joinSession clientId. Both truthiness guards are load-bearing — a solo/unjoined client has `clientId: ''`, and a pre-#3382 server sends none; either way we fall back to today's `'peer'`.
- `TODO(#3383)` removed.

The reducer path is untouched (`liftEnvelopeToSyncEvent`'s `QueueItemRemoved` case unchanged), so queue state cannot diverge — this is telemetry-only.

**`EVENTS_REPLAY` and `NATIVE_IOS_QUEUE_UPDATES` deliberately do NOT select `clientId`.** Replay _does_ reach the web analytics path (`session-connection.ts:652` → ports → `use-event-processor.ts` `notifyQueueSubscribers`), but delta replay only runs inside `reconnect()`, so every replayed event carries the **pre-reconnect** connectionId while the client already holds the new one — the comparison could never match and the field would be dead weight. `NATIVE_IOS_QUEUE_UPDATES` is an exact-match drift guard against the Swift client and must not diverge.

## Tests + fail-on-revert

1. **Backend contract** (`queue-sync-fixes.test.ts`): the `QueueItemRemoved` payload handed to `pubsub.publishQueueEvent` carries `clientId === ctx.connectionId`, and an empty connectionId publishes `null` (not `''`). Reverting the publisher drops the field entirely and both assertions fail.
2. **Mobile** (`queue-provider-sync-gate.test.tsx`): feeds a real wire-shaped `QueueItemRemoved` through the recorded subscription sink after a FullSync + add. `clientId: 'client-self'` (the harness's joinSession value) fires **no** `Climb Removed from Queue`; `'client-peer'` and `null` both still fire `removedBy: 'peer'`. **Mutation-tested** — neutering the guard to `false &&` makes exactly the self case fail (1 failed / 10 passed).
3. **Web** (`use-peer-broadcast-analytics.test.ts`): same self/peer/null matrix against the hook. Existing `QueueItemAdded` assertions left untouched as an unchanged-path guard.

None of the tests rebuild the handler's predicate — each feeds a literal envelope shaped like the real subscription selection.

## Scope: adds are out

`QueueItemAdded` has the identical bug (`addedFromTab: 'peer_broadcast'` hardcoded on both echo handlers while the local add already tracked itself), so party-session self-adds are double counted the same way. Deliberately out of scope here to keep the diff surgical — filed as **#4042** and referenced in-code at both sites.

## Risks

- **Metric discontinuity.** `Climb Removed from Queue` volume will **drop** on deploy: party-session self-removes stop emitting a duplicate. That is the fix working, not a telemetry outage. Note the asymmetry — `Climb Added to Queue` stays ~2x inflated until #4042 lands, so removes go to 1x while adds stay ~2x and any add/remove ratio on a dashboard shifts.
- **Bigger drop than 2x → 1x on mobile bulk paths.** `clearQueue` (`packages/mobile/src/providers/queue-provider.tsx:822-836`) fires N `removeQueueItem` mutations but tracks only `Queue Cleared` — no per-item event. Before this change the originating client received N echoes, each tracked `removedBy: 'peer'`; now all N are suppressed, so a party-session clear goes from N events to 0. Same for the resurrected-queue-add undo (`queue-provider.tsx:933`). Volume for clears is already carried by `Queue Cleared.totalCount`.
- ~~**`partyMode` shape asymmetry on mobile.**~~ Fixed in this PR. The suppressed echo was the only mobile remove carrying `partyMode`, so the surviving self-track now computes it the same way web does (`countDistinctSessionUsers(roster) > 1`), keeping a PostHog breakdown on `partyMode = true` intact across the deploy. Covered by two tests (party roster → `true`, solo → `false`); dropping the line fails both.
- **Reconnect window.** `connectionId` changes on reconnect and the stored id only updates once the rejoin resolves. An echo landing inside that window can still be tagged `'peer'`. Analytics-only, rare, strictly better than today's unconditional mislabel.
- **Residual, web-only.** A self-remove whose echo was lost to a socket drop is redelivered by delta replay and still tags `removedBy: 'peer'`. `connectionId` cannot survive the reconnect that triggers replay; a connection-stable client id is out of scope.
- **Legacy publishers.** Widget/controller paths and pre-deploy servers yield `clientId: null`, falling through to today's `'peer'`. The change is strictly additive.

No migration, no i18n keys, no BLE/native code, no `any`.

## QA Notes

1. **Party session, two clients** (mobile + mobile, or mobile + RN-web), both joined to the same session on the same board.
2. Remove a climb on client A. Expected: the climb disappears on both clients (unchanged), analytics records **one** remove on A (`removedBy: 'self'`) — not the current two — and one on B (`removedBy: 'peer'`).
3. Repeat from B to confirm symmetry.
4. Solo (no session): remove a climb; one `removedBy: 'self'`, queue behaves as before.
5. Edge: force a reconnect (airplane-mode toggle) mid-session, then remove. Queue must stay correct; a stray `'peer'` tag on the reconnect boundary is known and acceptable.
6. Regression watch: adds, reorders, and current-climb changes are deliberately untouched.
7. **Blast-radius caveat for counting events on web:** the root hook is board-route-gated and the `'self'` counterpart only fires inside the board-route `GraphQLQueueProvider`, so the double count exists only on board routes. The off-board bridge fires _neither_ event (`queue-actions-core.ts:132-135`) — an off-board event count of zero is expected, not a failure.

## Release Notes

none

Fixes #3382
Fixes #3383

#3383 is a near-duplicate of #3382 filed from the same #3372 review — it is the issue the removed `TODO(#3383)` marker named, and the only one of the pair that mentions mobile. Both are closed by this change; if you'd rather keep #3382 as the canonical one, close #3383 as a duplicate instead.

### DIFF
--- a/packages/backend/src/__tests__/queue-sync-fixes.test.ts
+++ b/packages/backend/src/__tests__/queue-sync-fixes.test.ts
@@ -644,6 +644,44 @@ describe('order-sensitive dual-hash (stateHashOrdered)', () => {
       }),
     );
   });
+
+  // Issue #3382 — QueueItemRemoved must carry the removing connection's id so
+  // subscribers can drop echoes of their own removes instead of tagging them
+  // `removedBy: 'peer'` (a double count, since the local remove already
+  // tracked itself as 'self').
+  it('publishes the removing connection id on QueueItemRemoved', async () => {
+    const sessionId = uuidv4();
+    await registerAndJoinSession('client-1', sessionId, '/kilter/1/2/3/40', 'User1');
+
+    const climb = createTestClimb();
+    await queueMutations.addQueueItem({}, { item: climb }, ctxFor(sessionId));
+    publishSpy.mockClear();
+
+    await queueMutations.removeQueueItem({}, { uuid: climb.uuid }, ctxFor(sessionId));
+    const removedCall = publishSpy.mock.calls.find(
+      (call: unknown[]) => (call[1] as { __typename: string }).__typename === 'QueueItemRemoved',
+    );
+    expect(removedCall).toBeDefined();
+    expect((removedCall![1] as { clientId: string | null }).clientId).toBe('client-1');
+  });
+
+  // An anonymous publisher must send null, NOT '': peers compare defensively,
+  // and two empty-string clients would echo-suppress each other's removes.
+  it('coerces a missing connection id to null on QueueItemRemoved', async () => {
+    const sessionId = uuidv4();
+    await registerAndJoinSession('client-1', sessionId, '/kilter/1/2/3/40', 'User1');
+
+    const climb = createTestClimb();
+    await queueMutations.addQueueItem({}, { item: climb }, ctxFor(sessionId));
+    publishSpy.mockClear();
+
+    await queueMutations.removeQueueItem({}, { uuid: climb.uuid }, { ...ctxFor(sessionId), connectionId: '' });
+    const removedCall = publishSpy.mock.calls.find(
+      (call: unknown[]) => (call[1] as { __typename: string }).__typename === 'QueueItemRemoved',
+    );
+    expect(removedCall).toBeDefined();
+    expect((removedCall![1] as { clientId: string | null }).clientId).toBeNull();
+  });
 });
 
 // Issue #2387 — the setQueue "redundant resync" warning must be order-aware.
--- a/packages/backend/src/graphql/resolvers/queue/mutations.ts
+++ b/packages/backend/src/graphql/resolvers/queue/mutations.ts
@@ -150,6 +150,10 @@ export const queueMutations = {
       stateHash,
       stateHashOrdered,
       uuid,
+      // Same coercion as setCurrentClimb/publishPlaybackState: an empty-string connectionId must
+      // become null, because peers compare defensively and two anonymous clients would otherwise
+      // echo-suppress each other's removes.
+      clientId: ctx.connectionId || null,
     });
 
     logMutationMetrics('removeQueueItem', performance.now() - startTime, sessionId);
--- a/packages/mobile/src/lib/graphql/operations.ts
+++ b/packages/mobile/src/lib/graphql/operations.ts
@@ -1278,6 +1278,7 @@ export const QUEUE_UPDATES_SUBSCRIPTION = `
         stateHash
         stateHashOrdered
         uuid
+        clientId
       }
       ... on QueueReordered {
         sequence
--- a/packages/mobile/src/providers/__tests__/queue-provider-sync-gate.test.tsx
+++ b/packages/mobile/src/providers/__tests__/queue-provider-sync-gate.test.tsx
@@ -205,6 +205,7 @@ type Snapshot = {
   state: ReturnType<typeof useQueue>['state'];
   sessionId: string | null;
   subscribeToQueueEvents: ReturnType<typeof useQueue>['subscribeToQueueEvents'];
+  removeFromQueue: ReturnType<typeof useQueue>['removeFromQueue'];
 };
 
 const user = (overrides: Partial<SessionUser> = {}): SessionUser => ({
@@ -360,8 +361,9 @@ function Probe({ onSnapshot }: { onSnapshot: (snapshot: Snapshot) => void }) {
       state: queue.state,
       sessionId: queue.sessionId,
       subscribeToQueueEvents: queue.subscribeToQueueEvents,
+      removeFromQueue: queue.removeFromQueue,
     });
-  }, [queue.state, queue.sessionId, queue.subscribeToQueueEvents, onSnapshot]);
+  }, [queue.state, queue.sessionId, queue.subscribeToQueueEvents, queue.removeFromQueue, onSnapshot]);
   return null;
 }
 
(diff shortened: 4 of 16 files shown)


marcodejongh:
Both review notes checked; neither needs a code change, but I documented the first inline (dfb4fc8):

- **`__typename` guard is not dead code.** The switch discriminates on `result.eventType`, a *different* variable, so `event` is still the full union inside the case. Removing the check is a hard typecheck failure — I tried it: `vp run typecheck:mobile` → 2 errors at use-session-realtime.ts:571. Added a comment saying so.
- **`stateHashOrdered` in the test helper.** `wireQueueItemAdded` omits it too, so the helper is consistent with the file's existing convention rather than drifting from it, and nothing on the non-FullSync path reads it (`stateHashOrdered` is consumed only from FullSync state and the local-hash verify). Leaving as-is.

PR #4054, opened 2026-07-31
fix(backend): land MoonBoard sends on the angle the problem is graded at

### LINKED ISSUE #3529: MoonBoard ticks at a non-catalog angle create phantom stats rows (wrong-angle, quality_normalized=false) invisible on the canonical climb

marcodejongh:
## What & why

Log a MoonBoard problem while your board config sits at the other angle and the send lands at an angle the catalog doesn't grade the problem at. It then does two bad things at once: it disappears from the page every other climber browses, and it mints a phantom `board_climb_stats` row that shows up in the other angle's search results as an ungraded climb with one ascent.

**This PR contains a PROD DATA REPAIR migration** (`0188`). It rewrites real climbers' tick rows and deletes prod stats rows, so it stays draft until a maintainer signs it off. See "Prod data repair" below for exactly what it touches, the same-day preflight counts, the readings still outstanding, and one consequence that is deliberately unmitigated. **The code fix stands alone** — say the word and the migration comes out into a follow-up PR, and this one merges with no prod data touched at all.

## Verified root cause

MoonBoard identity is angle-bearing today. The catalog importer mints one `board_climbs` row per _(problem, angle)_ as `uuidv5('moonboard:{id}:{angle}')`, each carrying a non-null `board_climbs.angle` (`packages/db/scripts/moonboard-catalog-helpers.ts`). So one physical problem is two catalog rows, and the angle is part of the climb's identity rather than a per-tick property.

`saveTick` never reconciled the two. `mutations.ts` stored `angle: validatedInput.angle` verbatim — whatever the client's board config happened to be on — and `updateTick` had the same hole. The stats recompute then seeded a `board_climb_stats` row at the tick's angle, guarded only on the climb existing (#3528).

Downstream, that phantom row _is_ read, just at the wrong identity:

- `packages/db/src/queries/climbs/search-climbs.ts` INNER JOINs `board_climb_stats` at the requested angle, so the phantom row surfaces in that angle's stats-driven search with a NULL grade and a 1-ascent count.
- `packages/backend/src/db/queries/climbs/get-climb.ts` joins stats on the requested angle, so the count is simply missing from the angle the problem is actually graded at.
- `packages/db/src/queries/recommendations/recommendation-query.ts` drops climbs you have already sent with `NOT EXISTS (… t.angle = <browsed angle> …)`. A send stranded at 25° never suppressed anything at 40°, so the recommender kept offering the climber a problem they had already done. Snapping (and, for existing rows, statement A) puts the tick at the angle that query looks at.

The rule "resolve a MoonBoard tick against the climb row's own angle" is already house style: `packages/backend/src/services/moonboard-import.ts` refuses to attach a CSV tick whose climb row isn't at 40.

The `quality_normalized` half of the issue is only half fixed on main: #3959 made the recompute seed `TRUE`, so forward minting is dead, but no backfill ever ran for the rows already written at the column default `FALSE`.

Numbers were re-taken rather than copied from the issue: the audit that filed #3529 counted 42 rows, a read-only prod preflight on 2026-07-31 found 100.

## The fix

**`resolveMoonBoardTickAngle`** (`packages/db/src/queries/climbs/moonboard-tick-angle.ts`) — one place that decides which angle a tick belongs at:

1. not MoonBoard → return the requested angle with **no query at all** (Kilter/Tension pay nothing)
2. one drizzle-builder query: the climb's own angle, its owner, and whether a stats row **carrying real catalog data** already exists at the requested angle
3. climb missing, or the climb is **user-created**, or `board_climbs.angle IS NULL`, or catalog data already at the requested angle → requested angle unchanged
4. otherwise → the climb's graded angle

It never throws on any path. A missing climb, a NULL angle and a DB error all fall back to the requested angle — a non-retryable GraphQL error dead-letters immediately in the offline drainer, so a hiccup here would cost a climber a send. Same log-only stance the #3528 catalog probe takes.

### Rule 3 tests for catalog data, not for a row

That is the load-bearing detail. Every affected climb in prod _already has_ a stats row at the wrong angle — that phantom row is the bug. A bare `EXISTS(stats row)` test would answer "yes" for all 100 of them and the code fix would be inert on exactly the population it exists for.

`statsRowCarriesRealCatalogData` (`packages/db/src/queries/climb-stats/real-catalog-data.ts`) is the one predicate — non-zero `upstream_ascensionist_count`, `display_difficulty`, `benchmark_difficulty` or `upstream_quality_average` — reused verbatim by the helper, both recompute seed guards, and both fences in the migration.

Those four columns are the ones no Boardsesh code path writes **on a catalog climb**, which is a narrower claim than "no Boardsesh code path can write them". Two writers do set `display_difficulty` / `benchmark_difficulty`: `saveMoonBoardClimb` (`packages/backend/src/graphql/resolvers/climbs/mutations.ts`) and the web proxy's `saveClimbStats('moonboard', …)` (`packages/web/app/api/v1/[board_name]/proxy/saveClimb/route.ts`). Both do it only for a climb they are creating, and both create that climb with a non-null `board_climbs.user_id` — and every consumer of the predicate reads it strictly behind the user-created fence, so those rows are never reached through it. If that ever stopped holding, the failure direction is protective: an extra `TRUE` reads as "real catalog data", and the fix responds by leaving the tick and the row alone.

### Rule 3 also fences off user-created climbs

`createClimb` writes a non-null `board_climbs.angle` plus a stats row carrying only `fa_username`, and `editClimb` changes the angle while deliberately leaving the old angle's row in place ("removing it would race with concurrent ticks"). An fa_username-only row scores **FALSE** under the catalog-data predicate — so without an explicit `user_id` fence, a climber ticking someone's re-angled problem at the angle it was originally set at would have had their tick silently rewritten to the setter's newer angle.

Migration 0188 fences statements A and B against exactly that shape (`bc.user_id IS NULL`). The runtime now carries the same fence, in the resolver and in both recompute seed guards, so the two halves of the fix can't disagree about what they're allowed to touch.

**Call sites** (`packages/backend/src/graphql/resolvers/ticks/mutations.ts`): the resolve promise starts next to the `reportTickClimbCatalogPresence` probe so it overlaps the board/session round-trips, and is awaited immediately before `db.transaction(`. The tick insert, the beta-link insert (the mobile home feed opens a video at the beta row's angle) and the downstream `queueClimbStatsRecompute` all read the snapped value. The #3528 presence probe keeps reporting the **client-sent** angle — it is an observation of the client, not of us. `updateTick` resolves only inside its `if (validatedInput.angle !== undefined)` branch.

### ❓ Open question: that `angle !== undefined` gate is weaker than it looks

The gate keys off the field being **present**, not off it having **changed**, and the two shipped clients disagree about that:

- **Web** (`packages/web/app/components/library/logbook-feed-item.tsx`, `handleSave`) sends `status`, `attemptCount`, `quality`, `difficulty`, `comment` — no `angle` key at all. A web edit never resolves, whatever the climber changed.
- **Mobile** (`packages/mobile/src/components/you/LogbookEditSheet.tsx`) puts the tick's **current** angle into `UpdateTickInput` on every save. So a comment-only edit from the mobile sheet does resolve, and on a historical wrong-angle MoonBoard tick it moves the tick to the graded angle and fires the counter.

So "a comment-only edit never snaps a historical tick" is true of the web surface and false of the mobile one. I have **not** tightened the gate to `validatedInput.angle !== existing[0].angle` — that changes which edits move a climber's data and is your call, not mine. Both readings are defensible: the tight gate keeps a historical tick exactly where the climber left it until they touch the angle control, while today's gate quietly repairs stranded ticks as climbers edit them. Say which you want and it's a one-line change; the code comment on that branch records the question in the meantime.

**Recompute seed guard** (`packages/db/src/queries/climb-stats/recompute.ts`, single-key + bulk): defence in depth for writers that don't go through `saveTick` — rows written before this shipped, bulk/self-heal recomputes over them, any future importer. Scoped to `board_type = 'moonboard'` and to catalog climbs, so Kilter/Tension are byte-for-byte unaffected and the #3528 invariant "a real climb ticked at a new angle still seeds" keeps holding (there is an existing test whose only job is to fail if that breaks; it stays untouched). The two seed statements can't stay diff-clean for this predicate — `board_climbs` is in the outer `FROM` for the single-key seed and inside an `EXISTS` for the bulk one — so the "should diff clean" comment now says so instead of the SQL being contorted. That shape difference is exactly how the `user_id` leg went missing from the bulk copy while the single-key one kept it; both now carry all four legs and each has its own revert-proved test.

**Observability**: a `MoonBoard Tick Angle Snapped` PostHog counter (distinctId = user, so one looping client can't read as a fleet-wide problem), mirroring `Tick Climb Not In Catalog`. That is how we'd find out a client surface is sending the wrong angle, and how we'd know the snap can retire. It fires **after the write lands**, off the `RETURNING` row's angle — at both call sites, so one event name means one thing: a tick we actually moved. In `saveTick` that means past the commit and past the "already existed, nothing was written" return, so a concurrent replay of an already-saved uuid or a rollback can't inflate it; in `updateTick` it means past the `UPDATE`. Deliberately the opposite stance from the `Tick Climb Not In Catalog` probe beside it, which counts what the client _sent_ and is independent of the tick surviving.

## Prod data repair — awaiting maintainer sign-off

`0188_moonboard_wrong_angle_stats_cleanup.sql` writes to production data. In one sentence: **it rewrites ~113 real climbers' ticks onto the canonical MoonBoard angle their problem is graded at, and deletes the ~100 phantom stats rows that move empties.** 10 climbers, 99 climbs. Nothing here is reversible from the migration itself.

Two ways to land this:

1. **Sign off on the migration** and merge the whole PR. Take the readings below first.
2. **Strip it.** Say so and 0188 moves to a follow-up PR; this one merges as code-only, existing wrong-angle rows stay exactly as they are, and only ticks logged from then on land correctly. No behaviour in the code fix depends on the repair having run.

Generated with `bunx drizzle-kit generate --custom`; the 0188 snapshot is byte-identical to 0187 apart from `id`/`prevId`, so no spurious `boardsesh_ticks_quality_range` DROP to strip.

**Prod preflight, read-only, 2026-07-31 (same day as the file):**

| Query                                                        | Count                                                          |
| ------------------------------------------------------------ | -------------------------------------------------------------- |
| MoonBoard stats rows at an angle their climb isn't graded at | **100** (the audit that filed this issue counted 42 — it grew) |
| …of those, carrying real catalog data                        | **0**                                                          |
| …of those, on a user-created climb                           | **0**                                                          |
| Ticks statement A moves / climbs / climbers                  | **113 / 99 / 10**                                              |
| Wrong-angle MoonBoard ticks on user-created climbs           | **0**                                                          |
| MoonBoard stats rows with `quality_normalized <> TRUE`       | **101**                                                        |
| `board_climb_grades` rows at a wrong MoonBoard angle         | **0** — nothing to clean up, no follow-up needed               |

- **A** moves the stranded ticks to the angle their climb is graded at.
- **B** deletes the phantom stats rows A just emptied.
- **C** flips the legacy `quality_normalized = FALSE` MoonBoard rows the #3959 code fix couldn't reach. Separable and uncontroversial — landable on its own if you want value before signing off on A/B.

Both A and B are scoped to catalog climbs (`bc.user_id IS NULL`), for the reason in "Rule 3 also fences off user-created climbs" above. Same precedent as `0173`.

A and B share **one identical fence**. If A emptied a row B then declined to delete, the row would keep a stale non-zero `boardsesh_ascensionist_count` with no tick anywhere to re-derive it — `selfHealStaleClimbStats` joins ticks to stats on _equal_ angle, so it would never revisit it.

Delete rather than merge-the-counts, because after A there are zero ticks left at the phantom key: the row's counts are provably stale artefacts, not data. Hand-adding them into the canonical row would re-implement the distinct-climber counting rule (absorption guard, upstream-represented guard) in SQL — the duplication `recomputeClimbStats` exists to prevent. Deletes propagate: `trg_board_climb_stats_delete` → `log_deletion_board_climb_stats` writes a `sync_deletions` tombstone (0144), so offline clients drop the ghost row on their next pull rather than holding it until a snapshot bootstrap. There is a real-DB test asserting the `moonboard:MOON-PHANTOM:25` tombstone.

No manual `updated_at` stamping anywhere. `trg_boardsesh_ticks_set_updated_at` (0146) stamps the moved ticks automatically (`angle` isn't in its ignore list), which is what makes `selfHealStaleClimbStats` re-derive the canonical rows. Idempotent by predicate — no `_bs_migration_guards` row needed.

### ⚠️ Statement A can leave a climber two sends at the same angle

Paired QA raised this and it is **unmitigated by design** — it needs your eyes, not a patch from me.

`boardsesh_ticks` has **no unique constraint** on `(user_id, board_type, climb_uuid, angle)`. Only `uuid`, `aurora_id` and `kilter_id` are unique, so nothing in the schema stops A from producing a duplicate. A climber who ticked the same problem **both** correctly at 40 and stranded at 25 ends up with **two send rows at 40** once A runs.

`ascensionist_count` stays correct either way — the recompute counts distinct climbers, not tick rows — so this is invisible in stats and visible in that one climber's logbook, as the same send listed twice.

I did not add a dedupe leg to A, deliberately: choosing which of a climber's two rows survives is a data-loss decision (they carry different `uuid`s, `climbed_at`, comments, quality and possibly a beta link), and the instruction here is no deletion without asking. The options are keep-both (cosmetic duplicate, zero risk), move-then-delete-the-loser, or skip the colliding ticks entirely and leave those few stranded. **Take the count first** (query 1 in the migration header's `UNMEASURED` block) — it may well be zero, since it needs one climber to have logged the same problem at both angles.

### ⚠️ One more read-only reading before deploy

Also written out in the `UNMEASURED` block, and **not implied by the preflight above**: **`fa_username`-carrying catalog rows at a wrong angle** — rows statement B will delete that a reader might reasonably expect protected. `fa_username` is intentionally **not** a catalog-data signal on MoonBoard (0173 established that MoonBoard catalog FA is not authoritative), and the one case where it does mean something — a setter's own climb — is already covered by the `user_id` fence.

### Accepted, not fixed: a moved tick leaves its beta video at the old angle

Statement A moves the tick and nothing else. It does not carry `board_beta_links.angle` across, and there is no fourth statement that would. So a climber who attached a video to a stranded send keeps that video pinned to the angle the send used to sit at: the send itself now reads at the graded angle, but tapping its beta in the mobile home feed still opens the angle the problem isn't graded at.

Nothing revisits those rows on its own. The only writer of `board_beta_links.angle` after the row is created is `updateTick` (`packages/backend/src/graphql/resolvers/ticks/mutations.ts`), and it fires only when a climber edits that tick's angle by hand. `persistEnriched` in the beta-videos resolver rewrites thumbnails and usernames, never the angle. Until someone makes that manual edit, the video stays where it is.

**How many rows this affects is unmeasured.** It needs a count of the ticks statement A moves that carry a `tick_uuid`-linked beta row, and nobody has taken it.

**Ruled on by Marco, 2026-08-02: accept and document.** The migration ships at exactly the scope signed off on 2026-07-31 — no fourth statement, no re-sign. It is recorded here and in the migration header so the boundary reads as chosen rather than missed, and the `updateTick` beta-move block now says the runtime symmetry is a runtime one only.

## Tests

67 tests across three real-Postgres suites in `packages/backend/src/__tests__/` (via `setupWorkerDatabase()`, because CI never runs the `packages/db` `tsx --test` suite). Every assertion SELECTs the persisted row back rather than reading the resolver's return value.

Every guard is revert-proved — each row names the edit and the test that goes red, and each was actually run:

| Revert this                                                                         | This goes red                                                                         |
| ----------------------------------------------------------------------------------- | ------------------------------------------------------------------------------------- |
| `angle: effectiveAngle` → `validatedInput.angle` in `saveTick`                      | 3 tests: the headline snap, the phantom-row snap, the beta-link angle                 |
| drop `statsRowCarriesRealCatalogData()` from the resolver's EXISTS (bare existence) | "snaps even when a PHANTOM stats row already sits at the requested angle"             |
| drop the MoonBoard block from the single-key recompute seed                         | 2 tests in `recompute-climb-stats`                                                    |
| drop `ownerUserId != null` from the resolver                                        | "leaves a tick on a USER-CREATED MoonBoard climb at the angle the climber sent"       |
| drop `bc.user_id IS NOT NULL` from the SINGLE-KEY recompute seed guard              | "single-key: STILL seeds a user-created MoonBoard climb at an angle it is not set at" |
| drop `bc.user_id IS NOT NULL` from the BULK recompute seed guard                    | "bulk: STILL seeds a user-created MoonBoard climb at an angle it is not set at"       |
| drop `benchmark_difficulty` from `real-catalog-data.ts`                             | "treats a benchmark_difficulty-only stats row as real catalog data"                   |
| drop `upstream_quality_average` from `real-catalog-data.ts`                         | "treats an upstream_quality_average-only stats row as real catalog data"              |
| drop `benchmark_difficulty IS NOT NULL` from either fence in the `.sql`             | "leaves a benchmark_difficulty-only row alone — both fences must count that leg"      |

The last three exist because `insertGradedStatsRow` sets upstream count + `display_difficulty` + `upstream_quality_average` together, so no fixture could show either of the other two legs was individually load-bearing. Deleting `benchmark_difficulty` from any of its three copies used to leave the whole suite green — which matters precisely because the benchmark-only row is the shape that would strand a stale count if fences A and B ever fell out of step.

**Two fixtures were fixed rather than trusted**, because they were passing for the wrong reason:

- The recompute user-created test originally pre-seeded a stats row at the ticked angle, so `ON CONFLICT DO NOTHING` made it pass whether the guard fired or not. Removing the pre-seeded row is what makes it load-bearing.
- "still seeds at the graded angle" was the same tautology against an existing row; it now seeds no stats row at all, so the seed genuinely has to fire.

Negative control: deleting the second `insertGradedStatsRow` from the two-angle fixture turns that test red, so it isn't green by accident.

The `updateTick` snap report moved after its `UPDATE` too, and both of its tests now assert the counter — one that a real angle edit reports exactly once with the RETURNING angle, one that an edit which **omits the `angle` field** (web's edit shape) leaves an already-stranded tick alone and reports nothing. Mobile's shape — the same comment-only edit, but with the tick's current angle attached — takes the resolve path instead; see the open question above.

**One change ships without a dedicated test, stated rather than hidden.** In `saveTick` the warn log and `MoonBoard Tick Angle Snapped` capture moved from before `db.transaction(` to after it, past the "already existed, nothing was written" return, and now read `tick.angle` off `RETURNING`. The branch that makes _that_ move matter — the insert no-opping on `onConflictDoNothing` — needs a duplicate `uuid` to appear between `saveTick`'s own pre-check `SELECT` and its `INSERT`, which can't be staged deterministically without mocking out the database handle the rest of the suite deliberately runs against. The existing tests still pin that the counter fires with the right payload on a real snap, and that it stays silent on every pass-through shape.

`moonboard-wrong-angle-cleanup-migration.test.ts` reads `0188_*.sql` **off disk** with `readFileSync` and executes it verbatim via `client.unsafe` — retyping the SQL would make every assertion a tautology. Seven shapes are seeded together so each fence is proved in the presence of the rows it must not touch: the damage shape, real catalog data at the "wrong" angle, a user-created climb with an `editClimb` leftover, an angle-agnostic climb, Kilter, a `moonboard_import` tick, and a benchmark-only row. Plus the `sync_deletions` tombstone, statement C's MoonBoard scoping, and a double-application no-op.

Also run: `vp check` (2 errors, 288 warnings — both errors are `TS2307: Cannot find module '@gorhom/bottom-sheet'` in `packages/mobile/src/web-shims/bottom-sheet.tsx`, a local `web-runtime` install artefact in a file this branch never touches), `vp run typecheck:db`, `vp run typecheck:backend`, `vp run check:db-migrations` (189 migrations, journal consistent), and the three touched suites: `moonboard-tick-angle` 13/13, `moonboard-wrong-angle-cleanup-migration` 10/10, `recompute-climb-stats` 44/44. The first pass at `recompute-climb-stats` hit the known `TRUNCATE … deadlock detected` flake that shows up when backend suites run back to back; it is green on its own.

## Interaction with the open MoonBoard programme

Documented in the helper doc, the recompute module doc and here, because whoever lands that work has to deal with it:

- **#3851** (angle-agnostic import) is the trigger to remove rule 4, but it does **not** remove it by itself. An earlier draft of this description said rule 4 would go inert with no code change once a re-import ran; that is wrong. #3851 writes `angle: null` only in the values it **inserts** for a brand-new climb row — its `onConflictDoUpdate` set-list on `board_climbs` is `{ characteristics, description }`, no `angle` — and it ships no migration. A re-import therefore leaves every existing canonical row's `board_climbs.angle` exactly where it is, and rule 3's `board_climbs.angle IS NULL` leg keeps not matching them. That leg does cover rows whose angle really is null (there is a test pinning the passthrough), which is the shape newly-imported problems arrive in. Retiring rule 4 for the rows already in the table is a deliberate code change here and in both recompute seed guards — plus a migration, if the canonical angles are ever meant to become null.
- **#3849** (angle dedup) does _not_ null the canonical angle. In the window after #3849 merges and before a re-import, a merged problem the catalog grades at only one angle keeps `angle = 40` with a single stats row, and rule 4 will snap legitimate ticks at the other angle. That window is the known cost.
- **#3852** (`moonboard-wide-angles`, flag off) deliberately lets climbers tick angles the catalog doesn't grade. Under rule 4 the first such tick is snapped back. Whoever lands #3852 must drop rule 4 or make it flag-aware.

This PR deliberately does **not** rebase onto that stack — it stands alone and stays small.

## Known consequence, stated plainly

Snapping rewrites the angle the climber's board was actually set to. After it, the send counts on the graded angle and reads as that angle in the logbook (the logbook feed is angle-agnostic), but it stops showing as ticked on the _other_ angle's view of that UUID — `search-climbs.ts` filters stats by the requested angle and `boardsesh_ticks_user_climb_lookup_idx` is angle-scoped. Same for the recommender's already-sent exclusion, which is angle-matched: it starts suppressing the problem at the graded angle and stops suppressing it at the angle the tick left. Only `RECOMMENDED_FRESH` can surface a climb at an angle with no stats row (it LEFT-joins stats), so that is the one variant where the un-suppression is visible at all. That is the trade: today the send is stranded somewhere nobody looks, including its own climber's climb page at the angle that's graded.

## QA Notes

- **MoonBoard, board at 25°, tick a problem graded at 40°.** The send should appear on the 40° page and _not_ mint an ungraded 1-ascent entry in 25° search. A `MoonBoard Tick Angle Snapped` event fires.
- **MoonBoard, tick at an angle the problem IS graded at.** Nothing moves, no event.
- **A climber's own MoonBoard problem.** Set one, re-angle it, then tick it at the original angle from a second account — the tick must stay where the climber put it. This is the case the migration is fenced against and the runtime now matches.
- **Kilter and Tension, any angle.** Byte-for-byte unaffected — no extra query is even issued.
- **Edit a tick's angle** (`updateTick`) on MoonBoard: same snap.
- **Comment-only edit on a historical wrong-angle tick — the two clients differ, and that is the current behaviour, not a bug to catch.** From the **web** logbook the tick stays exactly where it is (web's edit never sends `angle`). From the **mobile** logbook sheet the tick moves to the graded angle and a `MoonBoard Tick Angle Snapped` event fires, because that sheet attaches the tick's current angle to every save. Worth seeing both before answering the open question above.
- **Attach a video while ticking**: the beta link moves with the snapped tick.
- **Device QA (mobile, offline).** The offline outbox's SaveTick handler selects only `{ uuid }`, so a snapped tick reads at the client-sent angle locally for a moment while the server holds the snapped one. `boardsesh_ticks` is a pull-synced table, so the next incremental pull reconciles it — worth eyeballing that the logbook settles on the graded angle after a sync rather than staying split. No code change expected.
- **Log the same problem twice, once at each angle.** Both sends end up at the graded angle, and the logbook shows two entries for the same climb. Expected today — see the duplicate-tick section; worth seeing before deciding what statement A should do about it.
- **If the migration is signed off**: take the UNMEASURED readings first, then confirm affected climbers' logbooks show the send at the graded angle and that offline devices drop the phantom rows on their next pull.

## Release Notes

- MoonBoard sends now land on the angle the problem is graded at, so your send turns up on the page everyone else is looking at instead of vanishing into an angle nobody browses
- The trade: the send lands on the problem's canonical angle, so it stops reading as ticked at the angle your board was physically set to. It still counts, it still shows in your logbook — it just lives at the angle the problem is graded at
- Your recommended climbs stop handing you MoonBoard problems you've already done. A send filed at the angle your board sat at read as unsent to the recommender, so the same problem kept coming back around
- Cleared out the ghost entries that wrong-angle sends had been leaving in MoonBoard search results

Fixes #3529


marcodejongh:
## ⚠️ Holding this draft: migration 0188 needs explicit maintainer sign-off

`0188_moonboard_wrong_angle_stats_cleanup.sql` is a **prod data repair**, not a schema change. On the day it runs it will:

- **rewrite ~113 real climbers' tick rows** (statement A) — moving MoonBoard sends from the angle the climber's board was set to, onto the angle the problem is graded at. ~99 climbs, ~10 climbers.
- **delete ~100 prod `board_climb_stats` rows** (statement B) — the phantom rows statement A just emptied. Deletes propagate to offline clients as `sync_deletions` tombstones.
- backfill `quality_normalized = TRUE` on ~101 MoonBoard rows (statement C — separable and uncontroversial; landable on its own if you want value before deciding on A/B).

**The code fix stands alone without any of it.** `resolveMoonBoardTickAngle` + the recompute seed guards stop all new damage; the migration only repairs what is already in prod. If you'd rather not touch existing rows, say so and I'll split the migration into a follow-up PR and mark this one ready as a code-only fix.

**Two read-only readings are still outstanding** before A runs — both are written out as copy-pasteable SQL in an `UNMEASURED` block in the migration header:

1. **Duplicate-tick collisions.** `boardsesh_ticks` has no unique constraint on `(user_id, board_type, climb_uuid, angle)`, so a climber who ticked the same problem *both* correctly at 40 and stranded at 25 ends up with two send rows at 40. Stats stay correct (distinct-climber counting) but their logbook shows the duplicate. If non-zero, we should decide deliberately between moving and deleting the loser.
2. **fa_username-carrying catalog rows at a wrong angle** — rows statement B will delete that a reader might expect protected.

I could not take these two myself (no prod DB access from this environment), which is why they are flagged as unmeasured in the file rather than guessed at.

Requesting explicit sign-off on A and B before this leaves draft.

marcodejongh:
Maintainer sign-off recorded (2026-07-31, via campaign session): migration 0188 approved as drafted — snap ~113 wrong-angle MoonBoard ticks to canonical angles + delete the emptied phantom stats rows, keep-both semantics on any same-angle collision (no tick deletion). Marking ready.

marcodejongh:

marcodejongh:
🤖 Independent review from the `josh` loop: **HOLD — the code fix is sound; four things around it block merge.**

The runtime snap itself is right: it survives #3849 via the catalog-data fence rather than per-angle UUID identity, idempotency is clean (everything keys on uuid, no send can drop), all other tick writers are board-fenced, and json-import's angle-baked synthetic aurora_id is a hard reason this snap must never generalize beyond MoonBoard.

Blockers:
1. **The load-bearing '#3851's re-import retires this rule' claim is false** — that importer's ON CONFLICT omits `angle` and ships no migration; correct moonboard-tick-angle.ts:66-71, the recompute module doc, and the PR body.
2. **Three-way 0188 collision** with main (and this branch's `when` is older than main's 0188) — `vp run db:renumber` + rebase; note the renumber bot is currently broken fleet-wide (#4100), so verify the push actually happens.
3. **Migration statement A moves 113 tick angles but does NOT carry `board_beta_links.angle`** — the PR's own runtime (mutations.ts:784, 1093-1111) exists precisely so beta videos open at the graded angle; after the migration, every repaired tick with a linked beta keeps its video pinned at the old wrong angle permanently (and the Redis recent-beta cache isn't busted). Decide: carry it in the migration or document why not.
4. **Non-draft status while gating an unsigned-off migration that rewrites real climbers' ticks** — the two UNMEASURED readings + the duplicate-tick count (0188:82-89) are still unrun; per repo convention this should sit draft until Marco runs them.

Also worth a Release Notes line or follow-up: the angle-equality predicate feeds recommendations (recommendation-query.ts:60-67) — after the fix, already-sent problems stop being excluded at the browsed angle. And pre-existing but half of this PR's invariant: attachBetaLink without a tickUuid writes the raw client angle unsnapped (both live callers do exactly that), so a beta added from the 25° play view still lands at 25°.

marcodejongh:
Stack update (orchestrated rebase, 2026-08-13): rebased onto current main across the #3851 importer rewrite — the 8C-grade layer was semantically ported onto the new `moonboard-catalog-batch.ts`/helpers structure (unmapped-grade counting now lives in `stageCatalogBatch` and is pinned by new tests; the 8C repair keeps targeting `moonboard:{id}:{angle}` legacy uuids via the alias walk, byte-identical to pre-rewrite behaviour on both migrated and unmigrated databases).

**Gym commits unbundled** (maintainer decision): the #3431 website-vouching tightening conflicts with #4230's shipped unclaimed-listing claim flow at the DB level — both shapes are `ownerId=SYSTEM, vouched=false` at claim time, so separating them needs website *provenance*, which deserves its own PR. The two gym commits and their migrations left this branch; migration numbering compacted (invariant backfill → 0191, wrong-angle cleanup → 0192, `when` values unchanged).

Deploy-order note: the #3849 dedup migration must land AFTER this PR's 0191/0192 (verified corruption paths otherwise — see that PR's header); it renumbers above them once this merges.

Issue #4038, opened 2026-07-31
cf-connecting-ip is spoofable by a client that reaches the Railway origin directly


marcodejongh:
Split out of #2863 / its PR.

`resolveWebSocketClientIp` (`packages/backend/src/websocket/client-ip.ts`) trusts `cf-connecting-ip` first, which is correct while Cloudflare fronts `ws.boardsesh.com` — Cloudflare overwrites that header. A client that reaches the Railway origin **directly**, bypassing Cloudflare, can set `cf-connecting-ip` itself and rotate it per upgrade, minting a fresh bucket every time. That is the exact rotation #2863 exists to close, surviving via a different route.

`packages/backend/src/handlers/og-climb.ts` documents the same caveat and backstops it with a per-TCP-peer ceiling (`SOCKET_RATE_LIMIT_MAX = 600`); the WebSocket path has no equivalent secondary bucket.

Options:
1. Restrict origin access to Cloudflare — a CF-only firewall rule, or a Cloudflare Transform Rule injecting a shared-secret header the origin requires. Preferred: fixes this for every transport at once, including the HTTP paths.
2. Mirror `og-climb.ts` and additionally key a high-limit bucket on `req.socket.remoteAddress`, so a forged `cf-connecting-ip` still hits a per-peer ceiling.

Disclosed in the #2863 PR body rather than fixed there, since option 1 is infrastructure rather than code.

marcodejongh:
Claimed for batch 2 of the coordinated bug-fix run. I’ll audit the trusted-proxy boundary and plan a code backstop that cannot be bypassed through direct-origin header spoofing.

Issue #4037, opened 2026-07-31
Anonymous tier-1 rate limits are per-instance, so a rolling deploy hands out a fresh bucket


marcodejongh:
Split out of #2863 / its PR.

#2863 made anonymous WebSocket rate limits key on a trusted client IP, but tier 1 is still an in-memory map per backend instance. Tier 2 (Redis) is authenticated-only by design.

Railway rolling deploys run two instances with no sticky sessions, so during a deploy window a reconnecting anonymous client can land on the second instance and get a brand-new bucket. Steady state is one instance, so this is a deploy-window hole rather than a permanent bypass — which is why the #2863 PR left it alone (the issue framed Redis-for-anon as "consider", not required).

Wanted: extend the Redis tier to anonymous callers keyed on the normalized client IP from `packages/backend/src/websocket/client-ip.ts`, or accept the gap explicitly and close this.

marcodejongh:
Claimed for batch 2 of the coordinated bug-fix run. I’ll trace anonymous WebSocket rate-limit state across instances and plan a Redis-backed fix with failure-mode coverage.

Issue #4036, opened 2026-07-31
resolveBoardForConfig does not validate its arguments against the real board catalog


marcodejongh:
Split out of #2863 / its PR.

`resolveBoardForConfig` (and the sibling `resolveBoardFor*` resolvers in `packages/backend/src/graphql/resolvers/board-presence/`) accept board/layout/size/set/angle arguments and will resolve-or-create rows for combinations that do not exist in the board catalog. The #2863 fix deliberately kept these at 30/min precisely because they are the junk-row abuse vector the issue names, but a rate limit only slows the write, it does not reject a nonsense combination.

Wanted: validate the arguments against the shipped board catalog (`@boardsesh/board-config` / the DB catalog tables) before creating anything, and return a typed error for an unknown combination.

Explicitly scoped out of the #2863 PR, which was plumbing-only for rate-limit identity.

marcodejongh:
🤖 Claimed by the 2026-07-31 bug-fix campaign (Fable session `fablebugfix`, batch 1). Assigned to avoid duplicate work across concurrent agent campaigns — a PR from this session will follow. If another agent needs this issue, comment here first.

marcodejongh:
Coordination check from the codex-moarbugs loop: our claimed #4129 depends on the catalog-validation helper planned here. I do not see a #4036 PR or branch update after this campaign claim. Are you still actively implementing it? We are keeping hands off for now; if it is abandoned, we will post an explicit takeover notice here before changing anything.

marcodejongh:
🤖 Explicit takeover notice from bug-batch loop codex-moarbugs: the original fablebugfix claim has had no branch/PR update for roughly 12 hours and did not answer our coordination check. #4129 is blocked on this catalog-validation helper, so we are treating #4036 as unattended and taking it over now. We will preserve any relevant work if the original loop resurfaces; please comment before making parallel changes.

Issue #4035, opened 2026-07-31
No per-IP cap on concurrent anonymous WebSocket connections (verifyClient)


marcodejongh:
Split out of #2863 / its PR.

The #2863 fix bounds the **rate** of anonymous WebSocket operations by keying tier-1 rate limits on a trusted client IP. It does not bound the **count** of live anonymous sockets: `verifyClient` in `packages/backend/src/websocket/setup.ts` still accepts unlimited concurrent anonymous connections from one address.

Each socket costs a room-manager registration plus subscription bookkeeping, so a single client can still hold a large amount of server state open without ever tripping an operation-rate limit.

Wanted: a per-IP ceiling on concurrent anonymous connections, using the same `resolveWebSocketClientIp` normalization the rate limiter now uses (so an IPv6 client cannot rotate within its `/64`). Needs a limit high enough for a busy gyms NAT — the rate-limit work already assumes many phones share one address.

marcodejongh:
🤖 Claimed by the 2026-07-31 bug-fix campaign (Fable session `fablebugfix`, batch 1). Assigned to avoid duplicate work across concurrent agent campaigns — a PR from this session will follow. If another agent needs this issue, comment here first.

marcodejongh:
🤖 Claimed by bug-batch loop `claude-bb21` (batch 1, session c321d1b9, 2026-08-11). Where a 2026-07-31 `fablebugfix` claim exists it has lapsed (10+ days, no PR/branch) — reclaiming per protocol. A PR from this loop will follow; parallel loops please pick another issue.

PR #4062, opened 2026-07-31
fix(backend): key anonymous WebSocket rate limits on a trusted client IP

### LINKED ISSUE #2863: Per-IP rate limiting for anonymous WebSocket connections
## Problem

Anonymous WebSocket callers are not effectively rate-limited. `applyRateLimit` keys anon callers on `ctx.connectionId` when `clientIp` is absent, and WS contexts never populate `clientIp` (it's only derived for HTTP). Each new WS connection gets a fresh `connectionId` (UUID) → a fresh rate-limit bucket, so the limit is trivially bypassed by reconnecting. The Redis (tier-2) limiter is auth-only, so anon gets nothing distributed.

This is a **pre-existing** limiter weakness, but #2862 (board presence GA) made the anon resolvers reachable over WS (`reportBoardClimb` after a one-time resolve, `boardNowPlaying`/`boardRecentClimbs`/`boardConnection`), so it's now exploitable:

- spam `reportBoardClimb` → Redis writes + (after 60s dwell per connection) `board_climb_events` inserts;
- open unbounded `boardNowPlaying` subscriptions across guessed board ids;
- enumerate `boardConnection`/`boardRecentClimbs` across the sequential `bigserial` board id space;
- `resolveBoardForConfig` can create a system-owned `user_boards` row per distinct config hash → junk-row amplification.

Over HTTP the same anon calls are IP-keyed and bounded; this is specific to the WS path.

## Fix

- Derive the client IP in the WS `onConnect` hook (`packages/backend/src/websocket/setup.ts`) from `x-forwarded-for` / `ctx.extra.request.socket.remoteAddress` and store it on the connection context (`packages/backend/src/graphql/context.ts`) so anon WS callers fall into the `ip:` rate-limit bucket like HTTP.
- And/or cap concurrent anonymous WS connections per IP in `verifyClient`.
- Consider validating `resolveBoardForConfig` args against the real catalog before creating a system board, to bound junk-row creation.

## Context

- Surfaced by the security review of #2862; deliberately deferred there (per-IP WS limiting touches the context/limiter plumbing beyond that feature).
- Relevant code: `packages/backend/src/graphql/resolvers/shared/helpers.ts` (`applyRateLimit`), `packages/backend/src/graphql/context.ts` (`createContext`), `packages/backend/src/websocket/setup.ts` (`onConnect`/`verifyClient`).

marcodejongh:
## What / why

Anonymous WebSocket callers were effectively unrate-limited. `onConnect` never put a `clientIp` on the connection context, so `applyRateLimit` fell through to its last-resort branch and keyed on `connectionId` — a fresh `uuidv4` per connection. Reconnect, get a brand-new bucket. Tier 2 (Redis) is authenticated-only by design, so nothing else caught it.

The code already documented its own bug: `board-presence/subscription.ts` carried a comment stating "the WS context never sets clientIp … so anonymous subscribers are already bucketed per-connection, not per-IP."

**Verified root cause** (against `origin/main`):

- `packages/backend/src/websocket/setup.ts` — `onConnect` builds the context with no client IP.
- `packages/backend/src/graphql/context.ts` — `createContext` had no `clientIp` parameter at all.
- `packages/backend/src/graphql/resolvers/shared/helpers.ts` — key selection is `userId:op` → `ip:<clientIp>:op` → `connectionId`; the middle branch was unreachable over WS.
- `ConnectionContext.clientIp` already existed in `packages/shared-schema` — pure plumbing gap, no schema change.

## Fix

New `packages/backend/src/websocket/client-ip.ts` resolves the upgrade request's client IP, and `onConnect` passes it into `createContext`.

**Trusted-hop order** — `cf-connecting-ip` → the **LAST** `x-forwarded-for` hop → `req.socket.remoteAddress` → `undefined`. This mirrors `handlers/og-climb.ts` and deliberately does **not** copy `graphql/yoga.ts`'s first-hop derivation. Cloudflare currently fronts `ws.boardsesh.com` with Railway's edge behind it, and Cloudflare _appends_ to `x-forwarded-for` rather than stripping it — so every leading entry is client-authored. Keying on the first hop would let a scripted client set `x-forwarded-for: <random>` per upgrade and mint a fresh bucket every time (i.e. change nothing about this bug), and would additionally let an attacker pin a _victim's_ IP to exhaust their bucket. `x-real-ip` is dropped entirely: nothing in our chain sets it, so it is pure spoof surface.

**Normalization** before the value becomes a rate-limit map key: strip brackets and `%zone`, unwrap `::ffff:` when the tail is genuinely dotted-quad IPv4 (so the socket fallback and the header path agree on one bucket, while the hex form `::ffff:c000:0201` stays a valid IPv6 literal rather than being dropped), validate with `node:net`'s `isIP` and fall through to the next candidate on failure (junk headers can't mint permanent keys), and truncate IPv6 to its `/64` prefix so a routed `/64` can't mint unlimited buckets.

`createContext` moves from 6 optional positional parameters to a single options object. Every parameter after `connectionId` is an optional string, so passing `clientIp` into the `controllerMac` slot would have type-checked cleanly. One call site, so the refactor is free.

`reportBoardClimb` / `reportBoardDisconnect` get anon-specific headroom (`ctx.isAuthenticated ? 60 : 240`). Those anon buckets went from private-per-connection to shared-per-NAT, and `reportBoardClimb` fires on every wall send — three phones lighting climbs in one gym would otherwise trip a shared 60/min.

Stale comments updated: `helpers.ts` JSDoc, `subscription.ts`, `gym-kiosks.ts`, `shared-schema/types/events.ts`, plus a new Key Point in `docs/websocket-implementation.md`.

## Blast radius (stated plainly)

Every anonymous WS operation flips from per-connection to per-IP buckets: `boardNowPlaying`, `boardRecentClimbs`, `boardHistory`, `boardPresenceStats`, `boardConnection`, `reportBoardClimb`, `reportBoardDisconnect`, `resolveBoardFor*`, `kioskHeartbeat`. ESP32 controller connections authenticate by API key without setting `isAuthenticated`, so they share the anon IP bucket too. `resolveBoardFor*` deliberately stays at 30 — that's the junk-row abuse vector the issue names.

**Residuals this PR does not close** — all five are filed, none ship as prose:

1. #4035 — `verifyClient` still has no per-IP cap on **concurrent** anonymous sockets. This bounds the _rate_ of anon operations, not the _count_ of live connections.
2. #4036 — `resolveBoardForConfig` argument validation against the real catalog.
3. #4037 — tier 1 stays in-memory per-instance. Railway's rolling deploys run two instances with no sticky sessions, so a reconnect can land on a second instance during a deploy window. Redis tier-2 for anon callers stays out of scope per the issue's "consider" framing.
4. #4038 — a client reaching the Railway origin directly, bypassing Cloudflare, can still spoof `cf-connecting-ip` (same caveat `og-climb.ts` documents, which backstops it with a per-TCP-peer ceiling this path lacks). Mitigation is CF-only origin access or a CF Transform Rule shared-secret header.
5. #4034 — **`graphql/yoga.ts`'s HTTP anon limiter has the same first-hop flaw** and is intentionally untouched here. Fixing it changes HTTP rate-limit identity and belongs in its own PR with its own review.

## Tests

`packages/backend/src/__tests__/websocket-client-ip-context.test.ts` — the load-bearing one. Boots the **genuine** `setupWebSocketServer` on a real `http.Server`, connects a real `ws` client with forged upgrade headers, and reads the context `onConnect` actually built (spied via `vi.mock('../graphql/context', importOriginal)`):

- `cf-connecting-ip` wins over a client-authored forwarded chain.
- With only `x-forwarded-for: 198.51.100.9, 172.68.1.1`, the resolved IP is `172.68.1.1` — the attacker-prefixed entry is ignored.
- Two sequential connections with identical headers get **different** `connectionId`s but the **same** `clientIp` (the reconnect case from the issue).

**Revert check, run:** deleting the `clientIp` argument at the `createContext` call in `websocket/setup.ts` turns all three red (`expected undefined to be '203.0.113.7'`). The test drives the real defect site, so it cannot pass against the bug.

`packages/backend/src/__tests__/websocket-client-ip.test.ts` — 23 pure cases on `resolveWebSocketClientIp` / `normalizeClientIp`: cf preference, last-hop-wins, `x-real-ip` ignored, socket fallback, absent request → `undefined` (not a sentinel), junk-value fall-through, `::ffff:` unwrap agreeing with the header path, IPv6 `/64` truncation collapsing two hosts to one key, embedded trailing IPv4, and the hex-form IPv4-mapped literal (`::ffff:c000:0201` keys as IPv6, guarding the conditional unwrap — an unconditional `::ffff:` strip drops it). The array-valued-header case is labelled defensive-only: Node joins duplicate request headers into one string (`set-cookie` is the sole array-valued header), so that shape is type-level, not something a real upgrade produces. Reverting the last-hop derivation to first-hop fails the last-hop and cf-preference cases; removing normalization fails the `::ffff:` and `/64` cases.

## QA Notes

Backend-only, no UI surface. Worth exercising on a preview deploy:

- Connect the mobile app anonymously (logged out), browse a public board's live feed, and confirm `boardNowPlaying` / `boardRecentClimbs` still stream normally.
- Light several climbs in a row from an anonymous client and confirm no `RATE_LIMITED` toast — `reportBoardClimb` now allows 240/min for anon.
- Reconnect the socket several times (airplane-mode toggle) and confirm reconnects still work and the feed resumes.
- A gym kiosk / TV on the same network as phones should keep heartbeating; the heartbeat bucket is now shared across the NAT at 60/min against a ~5-min poll cycle.
- Backend logs: `Client connected: …` now carries a `clientIp` field. On a Cloudflare-fronted deploy it should be the real client address, not a `172.68.x.x` colo address.

Note `applyRateLimit` short-circuits entirely when `NODE_ENV === 'development'`, so local dev won't exercise the limiter.

## Release Notes

none

Fixes #2863


marcodejongh:
Duplicate of the same branch as #4055 — both superseded by the already-merged #4039.

PR #4065, opened 2026-07-31
fix(ci): give app.boardsesh.com telemetry and a deploy freeze

marcodejongh:
First PRs of the Expo Web Program phases 4–6 (Next.js scope-down / SEO move / web-only tail). Both are safety rails on `deploy-app-web` — no product code, no deletes.

Combines W-01 and W-02 from the plan. They were scoped as two PRs, but both edit `production-deploy.yml` and W-02 is 30 lines, so a stacked split would buy a rebase burden and no reviewer benefit. Two clean commits instead.

## The browser app has been shipping with no telemetry

`deploy-app-web` set four `EXPO_PUBLIC_*` values inline in the build step, and neither `EXPO_PUBLIC_SENTRY_DSN` nor `EXPO_PUBLIC_POSTHOG_KEY` — while all six mobile workflows set both. Those vars are inlined into the JS bundle at build time, and both SDKs gate on them being present _then_:

- `packages/mobile/src/lib/sentry.ts` — `isSentryEnabled = !!sentryDsn && !__DEV__`
- `packages/mobile/src/lib/posthog-client.ts` — `isAnalyticsEnabled = !!apiKey && !__DEV__`

So every production deploy of `app.boardsesh.com` has gone out with crash reporting and analytics both off. That is also why the subdomain has never appeared in the Sentry or PostHog data next to the native fleet.

This moves the export's env to a **workflow-level** block and adds the two keys plus `EXPO_PUBLIC_SENTRY_ENVIRONMENT: production-web`, which separates browser-app events from the native fleet in both SDKs (`environment` is init-only, so the build-time var is the only place it can be set — `src/lib/app-environment.ts`).

Workflow level rather than job level is deliberate: `scripts/mobile-ci-env-parity.test.ts` extracts workflow-level `env:` at 2-space indent (`^ {2}KEY:`) and cannot see a job-level block. Everything in there is a public client value — `EXPO_PUBLIC_*` is inlined into the bundle, so none of it is a secret.

That test now guards these three keys, in its own `describe` rather than the fingerprint-parity list. The distinction matters: the web export is a browser bundle and never enters the native fingerprint graph, so locking its env to the native builds would be meaningless. Its contract is narrower — the keys must exist at build time, the environment must be `production-web`, and the DSN must match the native workflows so all three SDKs stay in one Sentry project. Verified by mutation: deleting `EXPO_PUBLIC_SENTRY_DSN` from the workflow turns two of the new assertions red.

The stale-env grep now also asserts both keys and the environment tag actually landed in the emitted bundles. "The workflow set it" and "the bundle contains it" are different claims, and `expo export` reuses Metro's transform cache across env changes. The Sentry assertion matches the DSN's ingest host rather than the whole DSN so the log never echoes the key portion.

## A deploy freeze for the subdomain

`deploy-web` and `deploy-production-backend` both honour `check-rollback`, which detects a pinned Vercel Instant Rollback and stages instead of promoting. Cloudflare Pages has no equivalent signal here, and `deploy-app-web` gated only on `detect-changes` — so a Pages dashboard rollback was silently clobbered by the next merge touching `packages/mobile`, `packages/shared`, `packages/shared-schema`, the export script, or `deploy/app-subdomain`.

Set the `APP_WEB_DEPLOY_HOLD` repo variable to hold the subdomain where it is; clear it to resume. A held run pings Discord, because a skipped job is just grey in the run summary and a forgotten hold would strand the browser app on an old bundle with nobody noticing.

- The two keys and `production-web` appear in the emitted bundles on the next deploy (the new grep step fails the job if not).
- **Still unverified, and the reason this is worth watching on the first deploy:** neither `@sentry/react-native` nor `posthog-react-native` is in `metro.config.js`'s `WEB_SHIM_MODULES`, and there are no `.web.ts` forks for either. Until now the missing keys meant neither SDK ever tried to initialise in the browser build, so whether they initialise cleanly under react-native-web is genuinely untested. If the first deploy shows an init error in the browser console, the fix is `.web.ts` forks over `@sentry/browser` + `posthog-js-lite` — the same pair the classic web app already uses. Nothing here can break the app shell, since both SDKs are already bundled today and only the gate flips.
- With `APP_WEB_DEPLOY_HOLD` set, a merge touching `packages/mobile/**` skips `deploy-app-web` and posts to Discord; with it unset, the job runs exactly as before.
- `notify-failure` fires on `failure` or `cancelled` only, and a held job is `skipped`, so the freeze raises no false alarm.

One thing to know for sequencing: `Dockerfile.web` sets `EXPO_PUBLIC_WEB_URL` from its own `BASE_URL` build arg, inside the container. The workflow-level env added here does not reach it, so today's Vercel `build-web` is unaffected. That interaction becomes live if #3795 moves production web onto `Dockerfile.web`, which is why the plan sequences the `/app` prod-static retirement before it.

## Release Notes

none

### DIFF
--- a/.github/workflows/production-deploy.yml
+++ b/.github/workflows/production-deploy.yml
@@ -543,9 +543,16 @@ jobs:
   # detect-changes. No inline Playwright smoke gate — a per-deploy expo-web boot
   # is too heavy here (this file has no such gate for web/backend either); the
   # expo-web-smoke suite stays the dispatch-only gate in e2e-tests.yml.
+  # Manual deploy freeze for app.boardsesh.com, the Cloudflare Pages analogue of
+  # check-rollback. deploy-web and deploy-production-backend both honour a pinned
+  # Vercel Instant Rollback; Pages has no equivalent signal here, so without this
+  # a dashboard rollback of the subdomain is silently clobbered by the next merge
+  # touching packages/mobile, packages/shared, packages/shared-schema, the export
+  # script, or deploy/app-subdomain. Set the repo variable APP_WEB_DEPLOY_HOLD to
+  # any non-empty value to hold the subdomain where it is; clear it to resume.
   deploy-app-web:
     needs: [detect-changes]
-    if: needs.detect-changes.outputs.app_changed == 'true'
+    if: needs.detect-changes.outputs.app_changed == 'true' && vars.APP_WEB_DEPLOY_HOLD == ''
     runs-on: ubuntu-latest
     environment: Production
     steps:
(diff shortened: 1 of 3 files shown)


marcodejongh:
On the DSN-vs-variable point: leaving it as a literal is deliberate. All seven existing mobile workflows hardcode the same DSN literal — `ios-testflight-rn`, `android-apk-rn`, `android-apk-dev-client`, `mobile-ota-production`, `mobile-ota-check`, `mobile-ota-preview`, `mobile-ota-backport`. Sourcing it from a repo variable here alone would make this the odd one out, and the new `reports to the same Sentry project as the native workflows` assertion compares this value against `android-apk-rn.yml` directly, so a literal is what keeps that guard meaningful.

Worth doing as a separate sweep across all eight workflows if we want dashboard-side DSN rotation, but not worth diverging on in this PR.

The PostHog key differs for a real reason rather than inconsistency: it is already a repo variable (`NEXT_PUBLIC_POSTHOG_KEY`) shared with the web app, so there was an existing source to read from.

PR #4066, opened 2026-07-31
fix(mobile): stop the V-Points line overshooting week-to-week data

### LINKED ISSUE #3780: [Bug: UI] Profile: V-Points line can overshoot data points

marcodejongh:
## Summary

- The V-Points chart on You → Progress (`TotalAreaChart` in `packages/mobile/src/components/you/YouCharts.tsx`) rendered its line with gifted-charts' cardinal-spline `curved`/`curvature={0.2}` interpolation. That algorithm can swing the curve past a real data point's y-value between two neighbouring weeks.
- The V-Points series is a **cumulative running total** (`buildVPointsTimeline` in `packages/shared/profile-stats/src/chart-builders.ts` — each week's value only ever adds to the previous one). A curve that dips below the lower of two neighbouring weeks is showing something that can never actually happen: your total points going down. That's the "better/worse than reality" effect reported in #3780.
- Fix: drop `curved`/`curvature`, render straight segments between the real weekly points instead (`curved={false}`). A straight segment between two points can never exceed either endpoint, so overshoot is impossible — and since the data is evenly spaced (one point per ISO week, per the issue), straight segments read cleanly without extra plumbing.

## Why not a true monotone spline?

The issue asks for monotone interpolation (derivative constrained at each point — i.e. Fritsch–Carlson). I looked into it and I'm recommending the straight-line fix instead, for now:

- `gifted-charts`/`gifted-charts-core@0.1.81` only exposes `CurveType.CUBIC` (this cardinal-spline bezier, the overshoot source) and `CurveType.QUADRATIC` (also overshoots — plain through-midpoint quadratic, no derivative clamping). There's no monotone option and no prop to hand the library a custom SVG path or per-segment bezier control points.
- The workaround I considered — computing a monotone Hermite spline ourselves and feeding gifted-charts a dense array of interpolated sub-points so the line still looks curved — has a real problem: I traced `activatePointers()` in the library (`react-native-gifted-charts/dist/LineChart/index.js`), and a data point with `hidePointer: true` (needed on every synthetic sub-point so its fake value doesn't drive the tooltip) causes the touch pointer to **silently stop updating entirely** on that touch, rather than snapping to the nearest real point. With N synthetic points per week, most of the chart would become a tooltip dead zone. That's a worse regression than the P3 cosmetic bug it would fix.
- Straight-line segments through the real weekly points sidestep this cleanly: zero risk to the existing pointer/tooltip/spacing/label logic (untouched), and mathematically zero chance of overshoot.

**Follow-up idea for later** if the smooth-curve look is wanted back: implement the Fritsch–Carlson monotone spline in `@boardsesh/profile-stats` as a pure, testable function, but pair it with a from-scratch pointer/tooltip layer (bypass gifted-charts' built-in pointer entirely) so dense synthetic points don't break touch. That's a bigger lift than this P3 warrants on its own.

## Visual verification

This is a visual change, so it should not be merged without a look — flagging @marcodejongh for sign-off.

I was not able to produce before/after screenshots in this sandbox: the Android SDK bootstrapped cleanly (`vp run mobile:android-doctor` — KVM, adb, emulator, and the `boardsesh-pixel` AVD all came up fine), but rendering the You/Progress tab needs real logged-in data from a seeded backend (screenshot mode is a presentation-stability switch, not a data mock — see `packages/mobile/src/lib/screenshot-mode.ts`). Both data paths were blocked here:
- The shared dev Postgres (Tailscale-tunneled, shared across this batch run) hit a migration collision (`relation "board_climb_ingest_skips" already exists`) from a concurrent sibling agent's `vp run dev`, so local seeded data wasn't available.
- The prod test-account auto-login path needs `SCREENSHOT_USER_PASSWORD`, which isn't available in this sandbox.

To eyeball it locally: You tab → Progress sub-tab → V-Points card, ideally on an account with some week-to-week point swings (a big week followed by a quieter one shows the difference best). Before this PR the line curves smoothly through/past each point; after, it's straight segments between the real weekly dots — flat weeks stay flat, no dips below the previous week.

## Test plan

- [x] `vp check` — 0 errors (296 pre-existing warnings, unrelated to this change)
- [x] `vp run typecheck:mobile` — clean
- [x] `vp run test:mobile` — 572 files / 4852 tests passed, including the new regression test (`you-charts-line-interpolation.test.tsx`) which asserts `curved={false}` and no leftover `curvature` prop on the V-Points `LineChart`
- [x] `vp run check:mobile-bundle` — OK
- [ ] Manual visual check on device/emulator — **needs @marcodejongh**, see above

## Scope

- Mobile-only (this chart doesn't render on web; gifted-charts isn't a web dependency).
- JS/TS only, no native changes — ships OTA.
- Sibling PR fixing the chart-card overflow (#3778/#3050) touches layout/sizing in the same file area; this PR only touches the `LineChart`'s curve prop and adds a new test file, so the diffs shouldn't collide. Will rebase if they do once that one merges.

## Release Notes

Your V-Points progress line now connects the dots straight instead of curving — no more phantom dips or bumps between weeks making a session look better or worse than it was.

Closes #3780

### DIFF
--- a/packages/mobile/src/components/you/YouCharts.tsx
+++ b/packages/mobile/src/components/you/YouCharts.tsx
@@ -716,11 +716,7 @@ export const TotalAreaChart = memo(function TotalAreaChart({
             endOpacity={colorScheme === 'dark' ? 0.1 : 0.04}
             gradientDirection="vertical"
             thickness={2}
-            // Cardinal-spline curving (gifted-charts' `curved`) can overshoot past a
-            // data point's y-value between two neighbours, making a flat or dipping
-            // week look like it rose. The weekly cadence is evenly spaced already,
-            // so a straight-segment line reads just as smoothly without the
-            // overshoot (#3780).
+            // Straight segments avoid gifted-charts' cardinal-spline overshoot (#3780).
             curved={false}
             maxValue={model.maxValue}
             noOfSections={model.sections}
--- a/packages/mobile/src/components/you/YouCharts.tsx
+++ b/packages/mobile/src/components/you/YouCharts.tsx
@@ -716,8 +716,12 @@ export const TotalAreaChart = memo(function TotalAreaChart({
             endOpacity={colorScheme === 'dark' ? 0.1 : 0.04}
             gradientDirection="vertical"
             thickness={2}
-            curved
-            curvature={0.2}
+            // Cardinal-spline curving (gifted-charts' `curved`) can overshoot past a
+            // data point's y-value between two neighbours, making a flat or dipping
+            // week look like it rose. The weekly cadence is evenly spaced already,
+            // so a straight-segment line reads just as smoothly without the
+            // overshoot (#3780).
+            curved={false}
             maxValue={model.maxValue}
             noOfSections={model.sections}
             yAxisLabelTexts={model.yAxisLabelTexts}
--- a/packages/mobile/src/components/you/__tests__/you-charts-line-interpolation.test.tsx
+++ b/packages/mobile/src/components/you/__tests__/you-charts-line-interpolation.test.tsx
@@ -8,7 +8,7 @@
 // render straight segments (`curved={false}`) with no leftover `curvature`.
 import { createElement, useEffect, type ReactNode } from 'react';
 import { render } from '@testing-library/react';
-import { describe, expect, it, vi } from 'vitest';
+import { beforeEach, describe, expect, it, vi } from 'vitest';
 import type { RawVPointsTimeline } from '@boardsesh/profile-stats';
 
 const FRAME_WIDTH = 320;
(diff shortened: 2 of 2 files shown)


marcodejongh:
Taking over the unattended author-side review pass for bug-batch loop `josh` (session `4448b370`). I’ll preserve the pending fixture/mock corrections, finish the comment cleanup, validate, and push this branch.

Issue #3182, opened 2026-06-24
Native board renderer can crash on a vanished thumbnail-cache file (missing hold overlays)


marcodejongh:
## What's happening

The native board renderer writes hold-overlay PNGs into a disk cache (`board-thumbnails/`). Android can reclaim the app's `cacheDir` (and its subdirectories) under storage pressure mid-session. When that happens, the renderer keeps trying to write to the now-missing directory and throws `FileNotFoundException: …board-thumbnails/…png: open failed: ENOENT`, which rejects the `renderHoldsOverlay` promise. The result is missing hold markers on a climb — backgrounds still render and there's no crash (`handled:true`) — and it self-heals on app restart.

## Evidence

- PostHog: https://us.posthog.com/project/412845/error_tracking/019ef6c0-16ec-7412-b902-67646209f886 (the two near-identical IDs `019ef6c0` / `019ef6c5` are the `v3_f` filled vs `v3_s` stroke cache-key variants of the same failure).
- ~383 occurrences but **1 user / 8 sessions** over 30d. All from a single Samsung device, in an ~18-minute window on one day (2026-06-23).
- **Do not overstate production impact.** The JS bundle source for these events is a local Metro dev bundle (`/tmp/expo-bundler-…-index.js`), not a production `index.android.bundle`, on a day the renderer code was under active development. This is almost certainly a dev/tester device whose cache was reclaimed/cleared mid-session, not an organic production storage-pressure event. No recurrence before or after the burst.
- Platform: Android (`$os_version` 16). App build 2000363 / 2.0.0.

## Suspected cause

`packages/mobile/modules/board-renderer/android/src/main/java/com/boardsesh/boardrenderer/BoardRendererModule.kt`. The `cacheDir` is `lazy`-initialized once (lines 11-15), so `mkdirs()` runs a single time per process. `renderOverlay` then opens `FileOutputStream(outputFile)` (line 78) **without re-ensuring the parent dir exists** — and the code itself notes Android may reclaim `context.cacheDir` under storage pressure (lines 112-114). After a mid-process reclaim the lazy dir is never re-created, yielding the ENOENT. The failure isn't cached (`renderedOverlays.set` only runs in the `.then` path, not `.catch`, in `packages/mobile/src/hooks/use-native-climb-render.ts` ~line 675), so every re-render re-attempts the vanished write — hence the burst.

[...]

PR #4073, opened 2026-07-31
fix(mobile): write the overlay PNG atomically so a mid-write crash can't leave a corrupt cache hit

### LINKED ISSUE #3748: BoardRendererModule: partially-written overlay PNG survives a mid-compress crash and is served as a permanent cache hit
## What's happening

`BoardRendererModule.renderOverlay` (Android) writes the holds-overlay PNG with a fast-path cache check that trusts file *existence* alone, not file *completeness*. If the write is interrupted partway through, the truncated file is left on disk and gets served as a permanent "valid" cache hit on every future request for that cache key.

## Evidence (Android)

`packages/mobile/modules/board-renderer/android/src/main/java/com/boardsesh/boardrenderer/BoardRendererModule.kt:78-84`:

```kotlin
FileOutputStream(outputFile).use { outputStream ->
    val written = overlayBitmap.compress(Bitmap.CompressFormat.PNG, 100, outputStream)
    if (!written) {
        outputFile.delete()
        throw Exception("PNG compression failed")
    }
}
```

This only deletes `outputFile` when `compress()` returns `false`. If `compress()` (or the underlying `FileOutputStream` write) *throws* instead — e.g. an IO interruption, a storage-full condition, or the process getting killed mid-write — the exception propagates straight through `.use { }` (which closes the stream but performs no cleanup) and the outer `try { } finally { overlayBitmap.recycle() }` (which only recycles the bitmap), leaving a partially-written, corrupt PNG at `outputFile`'s path. Nothing deletes it.

The very next call for that `cacheKey` hits the fast path a few lines up:

```kotlin
if (outputFile.exists()) {
    // Touch mtime so LRU treats hot files as recently used.
    outputFile.setLastModified(System.currentTimeMillis())
    return "file://${outputFile.absolutePath}"
}
```

`exists()` is the only check — there's no size/integrity validation — so the truncated file is handed back as a "successful" render forever. The overlay `<Image>` on the JS side would then either fail to decode the corrupt PNG or paint a partial/garbled image, indistinguishable from a fresh bug report, and — unlike the vanished-file case fixed in #3182 — no JS-side existence check can catch this, because the file *does* exist; it's just wrong.

## iOS needs the same audit

`packages/mobile/modules/board-renderer/ios/BoardRendererModule.swift:142`:

```swift
try pngData.write(to: outputUrl)
```

`Data.write(to:)` on iOS is generally safer here — by default it does *not* write atomically, so a similar interruption could in principle leave a partial file, but the exact guarantees depend on the write options used (`Data.WritingOptions.atomic` would make this a non-issue by writing to a temp file and renaming into place). Needs a deliberate audit rather than an assumption either way; if it's already atomic, document why; if not, add `.atomic` or an equivalent temp-file-then-rename pattern.

## Suggested fix

- Android: on any exception during `FileOutputStream`/`compress`, explicitly `outputFile.delete()` in a `catch` (not just the `written == false` branch) before rethrowing — or write to a temp file in the same directory and `renameTo()` into place only after a fully successful compress, which makes partial writes structurally impossible instead of relying on cleanup-after-the-fact.
- iOS: confirm/enforce atomic write semantics for `pngData.write(to:)`.

## Why this is a separate issue from #3182

#3182 fixed the JS-side "cache entry survives a vanished file" bug — a purely JS/TS change, OTA-deliverable. This bug requires editing native Kotlin/Swift source, which needs an actual app-store rebuild (not an OTA update) to ship. Bundling it into #3182's fix would have blocked a same-day OTA behind a full native release cycle for no benefit, so it's tracked separately.

## Severity

Low — same class of rare, storage-pressure-triggered failure as #3182 (P3), but with no self-healing possible from the JS side once it occurs (the file exists, so it never triggers the new existence-check fix). Worth fixing opportunistically, not urgent.

marcodejongh:
> [!IMPORTANT]
> **[native-train]** — this PR touches native Kotlin (`modules/board-renderer/android/`) and Swift (`modules/board-renderer/ios/`) and moves the Expo native fingerprint. Per repo policy, `main` must stay OTA-compatible with store binaries, so this PR **stays a draft** and rides the next native release train — do not mark it ready for review or merge it on its own schedule.

## Summary

Closes #3748.

`BoardRendererModule.renderOverlay` writes each climb's hold-overlay PNG to a cache file keyed by `cacheKey`, and treats any future call for that key as a cache hit purely on `File.exists()` / `FileManager.fileExists(atPath:)`. If the write was interrupted partway through — an IO error, a storage-full condition, the process getting killed mid-compress — a truncated PNG was left at the final path with nothing to clean it up, and it was then served as a "successful" render forever. Unlike the vanished-cache-dir bug fixed in #3182/#4041, no JS-side existence check can catch this: the file *does* exist, it's just wrong.

### The fix

- **Android** — new `AtomicFileWrite.writeAtomically(destination) { tempFile -> … }` (pure `java.io`, no Android framework calls): writes into `"$cacheKey.png.tmp"` next to the destination, then `renameTo()`s it into place only after a fully successful `compress()`. Any throw — from the compress call or a failed rename — deletes the temp file in a `finally` and leaves the destination completely untouched, so a partial write is now structurally impossible instead of relying on cleanup-after-the-fact. Composes inside the existing `CacheDirRecovery.retryOnceAfterRecreating` wrapper from #4041 (that guards the *directory* vanishing; this guards the *write* itself).
- **Read-side sanity check (both platforms)** — the fast path now also requires the cached file's length to be `> 0`, not just `exists()`. This is what lets a file already corrupted by the bug *before* this fix ships self-heal on its next request — new writes can no longer land partially, but old corrupt ones are still sitting on devices until they're overwritten.
- **`CachePruner`** — extracted the existing size-cap LRU eviction out of `BoardRendererModule` into its own pure-`java.io` object, alongside a new unconditional sweep of orphaned `*.tmp` files. A process kill between the temp write and the rename leaves one of these behind; it's never a valid cache entry, so it's swept regardless of whether the cache is over its size cap.
- **iOS** — one-line fix: `pngData.write(to: outputUrl, options: .atomic)`. Foundation's `.atomic` option already does the temp-file-then-rename dance Android needed hand-rolled, so no custom logic was needed there.

### Assumptions worth flagging

- The Android temp filename is deterministic (`"$cacheKey.png.tmp"`), not a random `File.createTempFile` name. That's only safe because Expo serialises every call into this module on a single `modulesQueue` (documented in #4041's PR body) — there's no concurrent `renderOverlay` call for the same `cacheKey` to collide with today. Worth revisiting if that ever changes.
- iOS's `.atomic` write is a Foundation-internal implementation; if it ever leaves its own OS-managed stray temp file behind on a kill, that's outside our control and not something this PR chases further.

## Release Notes

- [ ] No release note needed (internal / technical change)

Board hold overlays can no longer get stuck showing a corrupted image after a rare mid-save crash — a broken render clears itself out instead of sticking around forever.

## Test plan

**Green `vp` checks below do not cover this fix.** This is pure native Kotlin/Swift; nothing in `vp check` / `vp run typecheck:mobile` / `vp run test:mobile` / `vp run check:mobile-bundle` compiles or exercises either platform's code. The real coverage is native CI:

- **Android**: new `AtomicFileWriteTest.kt` (happy path; a throw during write leaves nothing at the destination; a throw during write leaves a *pre-existing* destination completely untouched — the actual regression this closes; a failed rename still cleans up the temp file) and `CachePrunerTest.kt` (the orphaned-temp sweep, on its own and layered with the existing size-cap LRU eviction). Both are plain JUnit — no Robolectric, no real `Bitmap`/JNI — matching `CacheDirRecoveryTest`'s existing pattern for the sibling helper (#4041). These run in CI via `.github/workflows/android-pr-rn.yml`'s `board-renderer:testDebugUnitTest` job (the module is already listed in `TESTED_MODULES`).
- **iOS**: no new test added. The fix is a single documented Apple API flag (`.atomic`) with no hand-rolled logic to isolate, and there's no existing board-renderer coverage under `ios-tests/` to extend — a new XCTest target felt disproportionate for a P3 fix. Validated by `ios-rn-ci.yml`'s existing build-for-testing step (compiles the module) plus manual code review; a real mid-write-crash repro isn't practically automatable here.
- No JS/TS touched — the native return contract (`file://…` URI string) is unchanged, so nothing on the RN side needed updating.
- Not run locally: this environment has no Android SDK / Xcode toolchain to execute `testDebugUnitTest` or build the iOS target directly, so this relies on CI for both platforms.

### DIFF
--- a/packages/mobile/modules/board-renderer/android/src/main/java/com/boardsesh/boardrenderer/AtomicFileWrite.kt
+++ b/packages/mobile/modules/board-renderer/android/src/main/java/com/boardsesh/boardrenderer/AtomicFileWrite.kt
@@ -1,0 +1,53 @@
+package com.boardsesh.boardrenderer
+
+import java.io.File
+import java.io.IOException
+
+/**
+ * Writes a file by writing to a temp file next to the destination first, then
+ * renaming it into place. A thrown exception from [write] — or a failed rename
+ * — deletes the temp file and leaves [destination] completely untouched: either
+ * its prior contents (if any) survive unchanged, or nothing was ever created.
+ * There is no window where a reader can observe a partially-written file at
+ * [destination]'s path.
+ *
+ * `File.renameTo` is an atomic `rename(2)` when the source and destination are
+ * on the same filesystem, which they are here — the temp file is created in
+ * [destination]'s own parent directory.
+ *
+ * Fixes #3748: `BoardRendererModule.renderOverlay` used to write the overlay
+ * PNG directly to its final cache path via `FileOutputStream`, and only
+ * cleaned up when `compress()` returned `false` — not when it (or the
+ * underlying stream write) *threw*, e.g. an IO interruption, a storage-full
+ * condition, or the process getting killed mid-write. That left a truncated
+ * PNG on disk, and the cache's `exists()`-only fast path then served it as a
+ * permanent "successful" render forever.
+ *
+ * Pure `java.io` on purpose — no Android framework calls — so
+ * [AtomicFileWriteTest] needs neither Robolectric nor a real Bitmap, matching
+ * [CacheDirRecovery]'s approach for its sibling helper.
+ */
+object AtomicFileWrite {
+    /**
+     * Temp-file suffix used for the in-flight write. Shared with
+     * [CachePruner], which sweeps any file ending in this suffix as an
+     * orphaned artifact of a crash between [write] and the rename.
+     */
+    const val TEMP_SUFFIX: String = ".tmp"
+
+    fun writeAtomically(destination: File, write: (tempFile: File) -> Unit) {
+        val tempFile = File(destination.parentFile, "${destination.name}$TEMP_SUFFIX")
+        try {
+            write(tempFile)
+            if (!tempFile.renameTo(destination)) {
+                throw IOException(
+                    "Failed to move ${tempFile.absolutePath} into place at ${destination.absolutePath}"
+                )
+            }
+        } finally {
+            // No-op if the rename already moved it away; cleans up any partial
+            // temp file left by a throw above (from `write` or the rename check).
+            tempFile.delete()
+        }
+    }
+}
--- a/packages/mobile/modules/board-renderer/android/src/main/java/com/boardsesh/boardrenderer/BoardRendererModule.kt
+++ b/packages/mobile/modules/board-renderer/android/src/main/java/com/boardsesh/boardrenderer/BoardRendererModule.kt
@@ -17,33 +17,23 @@ class BoardRendererModule : Module() {
     @Volatile
     private var pruned = false
 
-    private fun pruneCacheIfNeeded(maxBytes: Long) {
-        val files = cacheDir.listFiles() ?: return
-        var totalBytes = files.sumOf { it.length() }
-        if (totalBytes <= maxBytes) return
-
-        // Android's File.lastModified() is our LRU proxy — atime is not
-        // reliably available on the filesystems Android caches live on.
-        val sorted = files.sortedBy { it.lastModified() }
-        var removed = 0
-        for (file in sorted) {
-            if (totalBytes <= maxBytes) break
-            val size = file.length()
-            if (file.delete()) {
-                totalBytes -= size
-                removed++
-            }
-        }
-        if (removed > 0) {
-            android.util.Log.i("BoardRenderer", "Pruned $removed cached PNGs; new total $totalBytes bytes")
-        }
-    }
-
     private fun renderOverlay(configJson: String, cacheKey: String): String {
         if (!pruned) {
             synchronized(this@BoardRendererModule) {
                 if (!pruned) {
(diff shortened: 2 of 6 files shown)


marcodejongh:
Post-merge follow-up: this PR landed before the final reviewed corrections. Draft #4081 is based on current `main` and contains only the missing hardening: unique `.bsov-*.tmp` paths, `Os.rename`, full retry/error coverage, v5 native/web cache invalidation, and PNG-only LRU accounting so unrelated files survive.

marcodejongh:
🤖 Coordinator note for @marcodejongh: I see you flipped this to ready. Confirming before anything merges — this PR is **native-fingerprint** (Kotlin + Swift in board-renderer). #4041 (same module, same label) merged to main this morning, so the train window looks open; if you intend #4073 to ride the same binary train, say the word (or merge it yourself) and it rides. Absent that explicit call, I'm treating the standing native-train rule as in force and leaving it unmerged. CI is fully green; JUnit coverage runs in android-pr-rn.yml.

PR #4081, opened 2026-07-31
fix(mobile): finish atomic overlay cache hardening

marcodejongh:
> [!IMPORTANT]
> **Native follow-up:** #4073 merged before its final correctness fixes landed. This draft contains only the remaining changes against current `main` and completes the fix for #3748. It changes Android native source and must ride an Android native release rather than an OTA-only rollout.

## Summary

#4073 established atomic overlay writes on both platforms, including Foundation `.atomic` on iOS. This follow-up carries the reviewed hardening that did not make that merge:

- Android creates a fresh, hidden, same-directory `.bsov-*.tmp` for every attempt, then publishes the closed PNG with API-21+ `android.system.Os.rename`.
- The atomic publisher is injected so the pipeline stays plain-JVM testable. `ErrnoException`, output creation/write/close failures, compression failure, and commit failure all surface as `IOException` while preserving their causes.
- The complete temp-create → compress → close → rename pipeline sits inside `CacheDirRecovery`, so directory reclamation retries from a new temp path.
- Startup cleanup sweeps only private `.bsov-*.tmp` artifacts. Size accounting and LRU eviction consider only final `.png` overlays, so unrelated files and unrelated `.tmp` files survive even when the PNG cache exceeds its cap.
- Renderer cache version v5 forces old potentially partial v4 overlays to regenerate. Native disk warm-up rejects v4 files, and the shared Expo web Cache API intentionally performs the same one-time v4 flush.
- Concurrent publication is safe: writers use distinct temp paths and the final file is always one complete winner.

No Swift change remains in this diff because the iOS `.atomic` write already shipped in #4073.

## Native delivery and fingerprints

The final OTA compatibility check on commit `74339261b` confirms:

| Platform | Result | Fingerprint |
| --- | --- | --- |
| iOS | OTA-compatible | `f2fb048a9827` (matches `main`; already released) |
| Android | Native change | PR `04a822f5aa0a` · `main` `48d2971669bd` |

Only Android changes fingerprint, as expected: this diff adds Kotlin under `packages/mobile/modules/board-renderer/android`. The v5 JavaScript contract therefore cannot reach incompatible older Android binaries and must ride a native Android release. iOS retains the Foundation atomic-write change already merged by #4073.

## QA

QA notes cover first render, cached reopen, concurrent thumbnail activity, app restart/cache warm-up, cache-directory reclamation, and the intentional Expo web v4 regeneration. Metro surfaced the notes at `.boardsesh/qa-notes.md`.

Validation completed before push:

- `vp run typecheck:mobile`
- `vp run test:mobile`: 572 files / 4,864 tests
- `vp run check:mobile-bundle`: iOS and Android bundles
- isolated JVM suite compiled against the installed Android SDK: 21 tests
- focused post-review `CachePrunerTest`: 6 tests, including over-cap preservation of unrelated files
- focused post-review `AtomicFileWriteTest`: 11 tests, including prior-render preservation on close failure
- scoped `vp check`: no formatting, lint, or type warnings in changed TypeScript files

Physical-device process-kill/storage-pressure QA and the macOS simulator check remain manual gates.

## Release Notes

Board hold overlays recover cleanly after an interrupted render instead of staying blank or corrupted.

### DIFF
--- a/packages/mobile/modules/board-renderer/android/src/main/java/com/boardsesh/boardrenderer/AndroidAtomicFileCommitter.kt
+++ b/packages/mobile/modules/board-renderer/android/src/main/java/com/boardsesh/boardrenderer/AndroidAtomicFileCommitter.kt
@@ -1,0 +1,31 @@
+package com.boardsesh.boardrenderer
+
+import android.system.ErrnoException
+import android.system.Os
+import java.io.File
+import java.io.IOException
+
+/** Android's API-21+ atomic same-filesystem publisher for completed overlay PNGs. */
+object AndroidAtomicFileCommitter : AtomicFileCommitter {
+    override fun commit(source: File, destination: File) {
+        try {
+            // AtomicFileWrite always creates source in destination.parentFile,
+            // so this is one same-filesystem rename that safely replaces an
+            // existing regular destination even under concurrent publication.
+            Os.rename(source.absolutePath, destination.absolutePath)
+        } catch (failure: ErrnoException) {
+            // CacheDirRecovery deliberately retries IOException only. Preserve
+            // errno as the cause while presenting the filesystem contract it
+            // expects at the module boundary.
+            throw IOException(
+                "Failed to publish ${source.absolutePath} at ${destination.absolutePath}",
+                failure,
+            )
+        } catch (failure: SecurityException) {
+            throw IOException(
+                "Permission denied publishing ${source.absolutePath} at ${destination.absolutePath}",
+                failure,
+            )
+        }
+    }
+}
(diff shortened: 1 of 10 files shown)


marcodejongh:
Addressed all non-blocking Claude review notes in `74339261b`:

- cleanup uses the direct best-effort `File.delete()` contract, without the misleading `SecurityException` catch;
- the close-failure regression seeds a valid prior overlay and asserts it remains byte-for-byte unchanged while commit is never attempted;
- standalone `catch (IOException) { throw failure }` arms are gone; each broad normalization catch has a documented `IOException` guard so existing I/O failures still propagate unchanged;
- concurrent publisher liveness assertions now identify thread timeouts explicitly before checking publication failures.

Validation: focused `AtomicFileWriteTest` 11/11, `vp run typecheck:mobile`, and `git diff --check` all pass. The branch is rebased onto current `main` (`f2bf15ab3`) so refreshed native fingerprints and CI cover the actual merge result.

PR #4083, opened 2026-07-31
fix(mobile): preserve offline auth sessions

### LINKED ISSUE #4001: Offline cold start can sign the user out on native (resolveAuthSession 'unavailable' → setIsAuthenticated(false))

marcodejongh:
## Summary

- Preserve the native authenticated tree and account-scoped offline data when an existing JWT can be read but expiry metadata, refresh transport, or the refreshed-token keychain read is temporarily unavailable.
- Retry degraded sessions once connectivity returns while coalescing reconnect checks, and fence every result by credential generation so stale work cannot revive a signed-out account.
- Keep explicit refresh rejection as a real sign-out and preserve the existing Expo web behavior and telemetry ownership.

Closes #4001

## Release Notes

Keep climbing from your downloaded boards when you open Boardsesh without a connection.

## Test plan

- `vp test run --project mobile packages/mobile/src/lib/__tests__/auth-session.test.ts packages/mobile/src/lib/__tests__/auth-interceptor.test.ts packages/mobile/src/providers/__tests__/auth-provider.test.tsx --reporter=agent` (92 tests)
- `vp check docs/mobile-auth-flow.md packages/mobile/src/lib/auth-session.ts packages/mobile/src/lib/auth-interceptor.ts packages/mobile/src/lib/__tests__/auth-session.test.ts packages/mobile/src/lib/__tests__/auth-interceptor.test.ts packages/mobile/src/providers/auth-provider.tsx packages/mobile/src/providers/__tests__/auth-provider.test.tsx`
- `vp run typecheck:mobile`
- `vp run test:mobile -- --reporter=agent` (572 files, 4,882 tests; completed before the final review-only regression tests were added)
- `vp run check:mobile-bundle` (iOS and Android bundles)
- Prepared `.boardsesh/qa-notes.md` and confirmed Metro surfaced it at the issue worktree QA URL before publication.

### DIFF
--- a/docs/mobile-auth-flow.md
+++ b/docs/mobile-auth-flow.md
@@ -106,13 +106,20 @@ Tokens are stored in `expo-secure-store` under three keys:
 
 ### Proactive refresh
 
-`isTokenExpiringSoon()` returns `true` when the JWT is within 24 hours of expiry. `ensureFreshToken()` is called before every authenticated request and triggers a refresh if needed.
+`isTokenExpiringSoon()` returns `true` when the JWT is within 24 hours of expiry or when its stored expiry metadata is missing. `ensureFreshToken()` is called before every authenticated request and triggers a refresh if needed. Keeping the missing-expiry case refreshable lets older or partially migrated credentials repair their metadata as soon as the backend is reachable.
+
+On app startup and foreground revalidation, a successfully read JWT establishes the local session before the expiry and refresh checks run. If the expiry read, refresh request, or post-refresh keychain read is temporarily unavailable, native keeps that JWT authenticated in a degraded state. This preserves downloaded boards and other account-scoped offline data even when the JWT is already past its server expiry; it does not bypass backend authorization, so online requests still fail until refresh succeeds. The provider retries a degraded session when the app becomes active or connectivity transitions from offline to online. Reconnect checks are coalesced and stop after a clean authentication result.
 
 ### 401 retry with deduplication
 
 If a request returns `401`, `authenticatedFetch()` triggers a token refresh and retries the request once with the new JWT. Concurrent refresh attempts are deduplicated via a shared `refreshPromise` -- only one network call is made regardless of how many requests hit `401` simultaneously.
 
-If the refresh itself fails, all tokens are cleared and the user is signed out.
+Refresh outcomes are deliberately split:
+
+- A `401` or `403` rejection (or a confirmed missing refresh credential) means the session cannot be renewed. The provider transitions to anonymous and runs the normal account-isolation cleanup.
+- A transport error, timeout, rate limit, backend failure, or transient keychain read error is `unavailable`, not rejected. Existing credentials remain stored and the established local session stays rendered while native waits to retry.
+
+Every native auth resolution carries the credential generation it started with. The resolver checks that generation after each asynchronous step, and `AuthProvider` checks it again immediately before applying state, so a delayed offline/refresh result cannot revive an account after sign-out or overwrite a newer login.
 
 ## Endpoint reference
 
--- a/packages/mobile/src/lib/__tests__/auth-interceptor.test.ts
+++ b/packages/mobile/src/lib/__tests__/auth-interceptor.test.ts
@@ -152,6 +152,24 @@ describe('ensureFreshToken', () => {
       warnSpy.mockRestore();
     }
   });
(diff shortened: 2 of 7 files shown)


marcodejongh:
Addressed the actionable automated-review notes in 7be1caa38:

- The reconnect promise now attaches a no-op rejection guard because `checkAuth` owns telemetry, avoiding any future double-reporting.
- Added coverage for a degraded native session revalidating on `AppState` → `active`, upgrading cleanly, preserving caches, and removing reconnect retry behavior (focused suite: 91 tests).

I left the other two semantics unchanged: `checkAuth` and the effect dependencies remain stable throughout one degraded episode, so the subscription is not repeatedly torn down; and each later foreground keychain failure is a distinct storage operation whose handled event should remain observable (Sentry groups the common cause).

marcodejongh:
Followed up on the second automated review in 25e7080c1: fixed the comment typo and added the consecutive reconnect-failure → eventual-success regression case (focused suite now 92 tests). The already-online degradation path is exercised by the AppState test added in 7be1caa38—it confirms no reconnect runs merely because the subscription starts online, then foreground revalidation clears degradation. The provider boundary test also explicitly asserts that `refresh-unavailable` does not emit duplicate provider telemetry while interceptor tests cover the originating failure.

marcodejongh:
The final automated-review “needs attention” items are cross-platform false positives rather than regressions:

1. Expo web resolves `auth-session.web.ts`, which delegates to `auth-store.web.ts`; that web implementation does populate `confirmedIdentity` and `identityInvalidated` on unavailable results. The native `auth-session.ts` shown in the comment is not the web runtime module. The shared provider intentionally consumes both platform result shapes, and this PR leaves the `.web.ts` path unchanged.
2. Telemetry ownership is code-enforced by the typed degradation discriminator (`'expiry-read' | 'refresh-unavailable' | 'refreshed-token-read'`) and `handleResolvedAuthenticatedTransition`’s explicit `stage !== 'refresh-unavailable'` branch. Provider regression coverage asserts no second report for an unavailable refresh, while the interceptor returns the original reported cause.

The passive-effect timing window is acknowledged, but the contract is deliberately false→true-only while degraded; the AppState-active retry now has direct coverage and guarantees recovery on foreground. No correctness change is warranted for these points.

PR #4084, opened 2026-07-31
fix: detect dropped Bluetooth links sooner

### LINKED ISSUE #3991: fix(firmware): request a bounded BLE supervision timeout so a powered-off board is detected in ~2s
## Needs on-hardware QA — an agent cannot verify this one

This is the last un-landed piece of PR #2350 (closed as mostly-superseded). It changes BLE link timing on real silicon, so it needs someone with an ESP32 board and a phone, not a unit test.

## What's missing

`embedded/libs/nordic-uart-ble/src/nordic_uart_ble.cpp` never requests a connection supervision timeout. When a connected board is powered off abruptly, the central falls back to its own (long) default before it decides the link is gone — so the app keeps showing the board as connected, and in a solo session the lightbulb stays lit long after the board went dark.

## What already shipped

The other two fixes from #2350 are on `main` via cd6a52272 (#2678):

- The primary advertisement carries **only** the Aurora service UUID. Two 128-bit UUIDs (36 B) plus Flags overflow the 31-byte legacy advertising limit, which hid the board from the app's filtered scan. Nordic UART is discovered via GATT after connect.
- `startAdvertising()` now reflects the real radio state (`advertising = pAdvertising->isAdvertising()`) and logs when a start didn't take, instead of logging success unconditionally.

## The proposed change

In `NordicUartBLE::onConnect`, after the connection handle is captured:

```cpp
pServer->updateConnParams(connInfo.getConnHandle(), 6, 18, 0, 200);
```

Args are `(handle, minInterval, maxInterval, latency, timeout)`; intervals in 1.25 ms units, timeout in 10 ms units — so 200 = 2000 ms. It satisfies the BLE constraint `timeout > (1 + latency) * maxInterval * 2`. The central may renegotiate, so this is a request, not a guarantee.

Note `onConnect` on `main` takes `NimBLEConnInfo&`, not the `ble_gap_conn_desc*` the PR was written against — the call site needs `connInfo.getConnHandle()`.

## QA gate (hardware required)

1. Flash the firmware to an ESP32 board controller.
2. Connect from the app, confirm the LEDs drive normally and the connection is stable over a few minutes (a too-aggressive supervision timeout causes spurious drops — that's the risk this change carries).
3. Pull power on the board. The app should mark it disconnected within a couple of seconds; in a solo session the lightbulb should go out.
4. Repeat on both iOS and Android — the two stacks pick different connection parameters and may renegotiate differently.
5. Watch for regressions in link stability at range, which is where a 2 s timeout is most likely to bite.

Don't merge on unit tests alone: the mock `NimBLEServer::updateConnParams` only records the arguments, which proves the call happens and nothing about how the radio behaves.

marcodejongh:
## Summary

- request a bounded BLE connection interval and two-second supervision timeout as soon as the controller accepts a central
- keep every initial and reconnect request ahead of the connection callback so downstream state starts after the link policy is submitted
- cover the exact NimBLE parameters and callback ordering with a recorder-only host mock

Closes #3991

## Release Notes

Boardsesh notices sooner when a powered-off DIY board controller drops its Bluetooth connection, so stale connected state clears in about two seconds.

## Test plan

- [x] `vp run controller:build` (ESP32 firmware build against NimBLE 2.5.1)
- [x] focused native Nordic UART BLE suite: 18 tests passed after the production change
- [x] fresh Firmware Unit Tests workflow passed after the mock-fidelity and reconnect-ordering follow-ups
- [x] independent implementation and test-fidelity review
- [x] `git diff --check`
- [ ] Complete the required Fable review for this Bluetooth change
- [ ] Flash an ESP32 controller and verify sustained traffic, idle behavior, range loss, rapid reconnects, and controller power-off on iOS and Android
- [ ] Confirm the disconnected lightbulb state appears within roughly two seconds after power-off

This PR intentionally remains draft until the Fable and hardware checks are complete.


marcodejongh:
Addressed the latest code-review items in 32bbc32a1:

- The NimBLE mock now matches the real non-`const` `updateConnParams` API and no longer uses `mutable` recorder fields.
- Expected timing values have named test constants with BLE-unit and wall-clock documentation.
- A disconnect/reconnect regression verifies a fresh parameter request and the new connection handle on every connection.

`git diff --check` passes. The local native-suite rerun could not acquire PlatformIO's home-cache lock in this restricted worktree, and elevated cache access was denied, so the fresh Firmware Unit Tests workflow is the compile/test authority for this follow-up. Hardware and Fable review gates remain open, so this PR stays draft.

marcodejongh:
Reconnect now asserts the complete timing tuple in 7a16e1e63.

The 6–18 interval is the issue's specified request and remains intentionally subject to the documented iOS/Android hardware gate; changing it speculatively would no longer validate the requested behavior. The callback-order flag is also intentional: the test asserts after the callback returns that the callback observed a prior parameter request. Putting Unity control flow inside the production callback seam would make the callback itself own test failure behavior without making the ordering assertion stronger.

Firmware CI is rerunning. Fable and physical-device validation remain the merge blockers.

marcodejongh:
Addressed the reconnect-ordering gap in ba4cc8c0a. The callback now records the request count it observed; the reconnect test asserts it observed count 2, so it fails if the second request moves after callback dispatch. `git diff --check` passes. The worktree still cannot resolve its Vite+ install or access PlatformIO's home cache, so the fresh Firmware Unit Tests workflow is the compile/test authority. Fable and physical-device checks remain hard draft gates.

marcodejongh:

marcodejongh:
🤖 **Fable review (CLAUDE.md Bluetooth rule) from the `josh` loop: HOLD — conditional approval, three changes before the hardware session.**

The diff itself is verified sound: args map 1:1 onto ble_gap_upd_params in correct HCI units, the spec constraint holds with huge margin (2000ms > 45ms), tests genuinely fail on revert, and the firmware compiles against real NimBLE 2.5.1 (checked locally — PR CI only builds the mock). Scope confirmed firmware-only: flash-the-board, no app release, no fingerprint.

**But the parameter choice has an Apple problem that makes the QA plan unfalsifiable as written:**
1. **7.5ms min interval violates QA1931** (≥15ms, multiple of 15ms) and the 2s timeout sits below the current Accessory Design Guidelines' 6-18s window. Apple explicitly says non-compliant requests *may be rejected* — and the CPUR is all-or-nothing, so rejection leaves iOS's own long defaults: the powered-off board still hangs 'connected', bug unfixed for every iOS user, CI green. The interval half isn't even needed: setPreferredParams already advertises 7.5-22.5ms passively. Recommend (12, 24, 0, 200) — clears every rule except the ADG timeout floor and costs nothing real.
2. **Add ~6 lines: override NimBLEServerCallbacks::onConnParamsUpdate and log the negotiated values to the web-UI Logger.** updateConnParams returns void and swallows the rc — without this, the hardware session cannot distinguish 'iOS accepted 2s' from 'iOS rejected everything'. Do this BEFORE flashing.
3. **QA gate must include a proxy-config build** (esp32s3dev-proxy et al.): 4 of 7 shipped envs run a central on the same radio with a fixed 15ms/510ms outbound link + ~99% scan duty — that 510ms link is what starves first if 7.5ms/latency-0 is granted. Also do before/after on the same iPhone (main vs PR): if detection latency doesn't improve, the request was rejected.

Also noted: if 2s IS accepted, a spurious timeout costs ~10s of reconnect grace (SERIAL_RECONNECT_GRACE_MS) and can open the board picker mid-session — the existing disconnect-category 'link_timeout' mapping gives the post-rollout signal to watch. Stays draft until a human with an ESP32 + iPhone + Android runs the extended gate; merging is safe blast-radius-wise (no auto-flash path exists) but pointless before the hardware answer. Release-notes gate passes.

marcodejongh:

Issue #4027, opened 2026-07-31
offline-sync: let a resolved HTTP/GraphQL status win over the English NSURL prose heuristic in isRetryable


marcodejongh:
_Carved out of #3869 (the third, soft "Consider ..." bullet). #3869 landed bullets 1 and 2 — the regex narrowing plus positive/negative tests — and deliberately deferred this one because it changes call-site ordering rather than a literal._

## The concern

`isRetryable` in `packages/shared/offline-sync/src/mutation-queue/error-classification.ts` consults `isNetworkError` **before** `getErrorStatus`:

```ts
export function isRetryable(error: unknown): boolean {
  if (isNetworkError(error)) {
    return true;
  }
  const status = getErrorStatus(error);
  ...
}
```

So the English-prose NSURL fallback (`IOS_NSURL_ENGLISH_DESCRIPTIONS`) can win over a real HTTP/GraphQL status. If a server ever returns, say, a 400 whose message happens to match one of those phrasings, the mutation is treated as retryable and head-of-line-blocks the outbox instead of dead-lettering.

## Why it wasn't done in #3869

Genuinely soft ("Consider ..."), and the change is bigger than the regex fix: it touches the ordering in `isRetryable`, the `isNetworkError`/`isTransportNetworkError` split, and mobile's `reportHandledError` path (`packages/mobile/src/lib/error-reporting.ts`), each of which has its own tests. #3869 is a P3, so the scoping call was to land the narrowing and track this separately.

## Suggested fix

Only apply the English-prose branch when no HTTP/GraphQL status resolves, so a server verdict always beats the heuristic. Locale-independent signals (NSURL codes, `.cause` markers) should keep their current precedence — it's specifically the English-description fallback that should be last.

marcodejongh:
🤖 Claimed by bug-batch loop `josh` (batch 1, session 4448b370). Plan → draft PR incoming. Parallel loop: please pick a different issue.

marcodejongh:
Draft PR opened: https://github.com/boardsesh/boardsesh/pull/4067

marcodejongh:
PR #4067 is green across CI (offline-sync: 297/297, mobile: 4851/4851, typecheck/lint/mobile-bundle all pass) and marked ready for review: https://github.com/boardsesh/boardsesh/pull/4067

marcodejongh:
[...]

PR #4085, opened 2026-07-31
fix(mobile): respect server statuses in network telemetry

marcodejongh:
## Summary

- complete the mobile half of the status-precedence fix merged in #4067
- classify resolved HTTP and GraphQL statuses as server errors even when their message resembles connection prose
- preserve warning-level handling for empty and statusless GraphQL client responses

Completes the remaining work for #4027.

## Validation

- `vp test run packages/shared/offline-sync/src/mutation-queue/__tests__/error-classification.test.ts packages/mobile/src/lib/__tests__/error-reporting.test.ts --reporter=agent` (76 tests)
- `vp run typecheck:mobile`
- `vp check --fix` on the changed files

## Release Notes

### DIFF
--- a/packages/mobile/src/lib/__tests__/error-reporting.test.ts
+++ b/packages/mobile/src/lib/__tests__/error-reporting.test.ts
@@ -206,6 +206,15 @@ describe('reportHandledError', () => {
     });
   });
 
+  it('reports an HTTP error at error level', () => {
+    const serverError = Object.assign(new Error('Internal Server Error'), { response: { status: 500 } });
+    reportHandledError(serverError, { tags: { source: 'react-query', kind: 'mutation' } });
+    expect(mockedCaptureToSentry).toHaveBeenCalledWith(serverError, {
+      level: 'error',
+      tags: { source: 'react-query', kind: 'mutation' },
+    });
+  });
+
   it('reports an HTTP error even when its message resembles NSURL prose', () => {
     const serverError = Object.assign(new Error('The connection has timed out unexpectedly.'), {
       response: { status: 400 },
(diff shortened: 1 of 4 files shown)


marcodejongh:
Addressed the non-blocking review note in c9342806d: restored explicit coverage for a normal HTTP 500 response while retaining the NSURL-prose/status-precedence cases. Focused classifier + telemetry tests now pass 76/76.

marcodejongh:
🤖 Post-merge independent review from the `josh` loop (merge landed mid-review): **verdict would have been MERGE WITH NITS — no concerns with the merge itself.** Verified with a comment-stripped diff that the shared offline-sync change is code-identical to #4067 (drainer/outbox untouched, isTransportNetworkError contract intact and still test-pinned), both new mobile tests genuinely fail on revert, and the status-aware Sentry tagging completes what issue #4027 itself scoped (its body names reportHandledError as a call site — #4067's exclusion was scoping, not a veto). Bonus: the old code could throw a null-deref inside reportHandledError on `response: null`; the getErrorStatus route null-guards it.

Follow-up nits: (1) `isTransportNetworkError` now has zero production consumers — its own new doc comment describes callers that don't exist; delete it + its describe block or mark it deliberately retained (CLAUDE.md: remove dead code); (2) the 'nested GraphQL 400 + prose' test doesn't fail on revert of THIS PR and the nested-numeric-code shape is unreachable against our backend (all extensions.code values are strings) — fine as defensive coverage, but the PR body overstates the shared-logic delta. One process note: this merged before independent review finished — with two loops running, a short review window before self-merge would have caught these nits pre-merge.

PR #4089, opened 2026-07-31
fix(mobile): correct native tab chrome metrics

### LINKED ISSUE #3973: iOS 26 native glass: floatingControlBottom / scrollBottomPadding double-count the tab bar + accessory

marcodejongh:
## Summary

- Treat the iOS 26 native tab-bar inset as opaque instead of adding tab and accessory heights a second time.
- Use the same native-tab predicate for scroll padding, floating/fixed controls, and Toast placement.
- Preserve existing Material, sidebar, and JavaScript Liquid Glass fallback geometry.
- Correct pushed-session route documentation to match the actual accessory mount gate.

Closes #3973

## Release Notes

Keep controls, lists, and messages tucked above the iOS tab bar without the extra gap.

## Test plan

- Focused bottom-chrome and Toast tests: 32 passed
- `vp check` on all 10 changed files
- `vp run typecheck:mobile`
- `vp run test:mobile` (clean rerun; isolated auth-provider retry also passed 38/38)
- `vp run check:mobile-bundle` (iOS and Android passed; existing expo-share-intent SDK warning only)
- Independent code review: pass
- [ ] iOS 26 device: record raw bottom inset and screenshot top-level route with active accessory (known measured reference: 139 pt)
- [ ] iOS 26 device: capture top-level route without an active climb
- [ ] iOS 26 device: verify pushed Session Detail last-row clearance
- [ ] iOS 26 device: verify minimized tab state after scrolling
- [ ] iOS 26 device: verify QueueAddedSnackbar and Toast clearance

### DIFF
--- a/docs/mobile-sheets-vs-routes.md
+++ b/docs/mobile-sheets-vs-routes.md
@@ -186,6 +186,12 @@ Is it a secondary surface OVER the current screen, or its own full surface?
    paints its own opaque `View` under its `GlassSurface` so the live tabs screen doesn't show
    through. See `isTabsChromeRoute` in `src/lib/route-segments.ts`.
 
+   A **pushed route** under `NativeTabs` is different: the native tab bar remains, but route
+   classification deliberately unmounts the BottomAccessory / queue chrome on screens such as
+   session detail. Its bottom layout must therefore trust the raw UIKit safe-area inset for the
+   chrome that is actually present; it must not carry the accessory reserve forward from the tab
+   root or add the tab-bar height a second time.
+
 3. **Board gestures and modal/sheet pan don't mix.** A full-screen board you pan/pinch (the
    `holds` and `zone` filters) is a **pushed** route, not a modal — a modal/sheet's own pan
    gesture competes with the board's.
--- a/packages/mobile/app/(tabs)/home/_layout.tsx
+++ b/packages/mobile/app/(tabs)/home/_layout.tsx
@@ -10,8 +10,9 @@ export default function HomeLayout() {
         {/* The Home feed owns its top via floating chrome (the avatar island + scope
           title), like the other tabs — so the stack header is hidden here. */}
         <Stack.Screen name="index" options={{ headerShown: false }} />
-        {/* Session detail keeps the native tab bar + bottom accessory by living in
-          this stack. It sets its own header title from the loaded session. */}
+        {/* Session detail keeps the native tab bar by living in this stack, while
+          pushed-route accessory/queue chrome unmounts. It sets its own header
+          title from the loaded session. */}
         <Stack.Screen name="session/[sessionId]" options={{ headerShown: true }} />
       </Stack>
     </BoardArtVisibilityProvider>
--- a/packages/mobile/app/(tabs)/home/session/[sessionId].tsx
+++ b/packages/mobile/app/(tabs)/home/session/[sessionId].tsx
@@ -1,4 +1,4 @@
-// Session detail pushed inside the Home tab stack, so the native tab bar + Liquid
-// Glass bottom accessory stay on screen (matching the playlist detail view). Shares
+// Session detail is pushed inside the Home tab stack, so the native tab bar stays
+// on screen while the Liquid Glass BottomAccessory/queue chrome unmounts. Shares
 // the same screen component as the Profile tab's session route.
 export { default } from '../../../../src/components/session/SessionDetailScreen';
--- a/packages/mobile/app/(tabs)/profile/_layout.tsx
+++ b/packages/mobile/app/(tabs)/profile/_layout.tsx
@@ -15,8 +15,9 @@ export default function ProfileLayout() {
           title collapsing into a glass capsule), like the Discover/Climbs tabs —
           so the stack header is hidden here. */}
         <Stack.Screen name="index" options={{ headerShown: false, title: t('mobile.nav.profile') }} />
-        {/* Session detail keeps the native tab bar + bottom accessory by living in
-          this stack. It sets its own header title from the loaded session. */}
+        {/* Session detail keeps the native tab bar by living in this stack, while
+          pushed-route accessory/queue chrome unmounts. It sets its own header
+          title from the loaded session. */}
         <Stack.Screen name="session/[sessionId]" options={{ headerShown: true }} />
         <Stack.Screen name="more" options={{ title: t('mobile.more.title') }} />
         <Stack.Screen name="accessibility" options={{ title: t('mobile.more.accessibility.title') }} />
--- a/packages/mobile/app/(tabs)/profile/session/[sessionId].tsx
+++ b/packages/mobile/app/(tabs)/profile/session/[sessionId].tsx
@@ -1,5 +1,5 @@
-// Session detail pushed inside the Profile tab stack (reached from the Sessions
-// sub-tab), so the native tab bar + Liquid Glass bottom accessory stay on screen
-// (matching the playlist detail view). Shares the same screen component as the Home
-// tab's session route.
+// Session detail is pushed inside the Profile tab stack (reached from the Sessions
+// sub-tab), so the native tab bar stays on screen while the Liquid Glass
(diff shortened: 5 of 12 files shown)


marcodejongh:
CI exposed a missing test mock after `Toast` began using the canonical native-tab hook: `toast-provider.test.tsx` imported the real Expo/native-tab module chain and failed to parse React Native Flow syntax. 76c3a5d8e adds the explicit hook mock. The production code is unchanged.

Reproduced locally before the fix, then verified the provider + Toast + bottom-metrics group passes 33/33 with scoped `vp check` and `git diff --check`. Fresh CI is running.

marcodejongh:
Addressed the follow-up review in b1cf27eb2: the test matrix now has an explicit `usesNativeTabBar: true` + `usesSidebar: true` case proving the sidebar short-circuit wins, and the 83pt fixture explains why it remains synthetic despite matching plausible arithmetic. Focused metrics tests pass 25/25, scoped `vp check` passes, and `git diff --check` is clean. Physical iOS 26 QA remains the draft gate.

marcodejongh:
Addressed the latest review in `79cebe12b`: the route/sheet guide now states that pushed `NativeTabs` routes retain the tab bar while deliberately unmounting BottomAccessory/queue chrome; the synthetic inset is now an unmistakably non-device 100pt fixture; and the device-verified 139pt anchor explicitly pins both `fixedFooterBottom` and `preSessionFooterBottom`. The focused metrics suite passes 25/25, scoped `vp check` passes, and `git diff --check` is clean. The PR remains draft for the five physical iOS 26 checks.

marcodejongh:
🤖 Independent review from the `josh` loop: **MERGE WITH NITS once the device-QA gate clears** (draft gating is correct as-is).

Checked and cleared the two real risks: (1) every consumer of `floatingControlBottom`/`scrollBottomPadding`/`tabBarBottom` was grepped — none hand-compensated the old double-count, so nothing under-pads after the fix (the one non-obvious consumer, ActiveContextBar via `tabBarBottom`, is unreachable when the native accessory is mounted); (2) `tabBarOverlaysContent` requires the iOS 26 native-glass path, so Material/Android/pre-26/tablet arithmetic is unchanged — confirmed against the untouched test assertions. No overlap with merged #4082 (zero shared files).

Nits: **rebase on main before merging** — the branch is 27 commits behind and the OTA bot's 'Android native change' flag is stale-base noise from AndroidAtomicFileCommitter.kt on main, not this diff (verified: no native paths touched here); and `nativeAccessoryReserve` is now always 0 on the NativeTabs path (dead weight, no consumers — fine to leave or drop). Physical iOS 26 QA + screenshots remain the merge gate per @marcodejongh's checklist.

marcodejongh:
Rebased cleanly onto current `main` in `afc819819` after the independent review confirmed the Android native-change signal was stale-base noise; the branch itself still touches no native inputs. Fresh CI/OTA checks are running. The PR intentionally remains draft for the physical iOS 26 inset, layout, minimized-tab, pushed-route, and Toast checks.

marcodejongh:
Addressed the two code/documentation points from the latest review in `91a43760f`: the synthetic fixture now records the plausible-but-unverified 83pt no-accessory arithmetic, and the session test explains why the former strict footer ordering correctly became equality when UIKit’s inset became opaque. Focused metrics tests pass 25/25, scoped `vp check --fix` passes, and `git diff --check` is clean. The PR remains draft for the five physical iOS 26 checks.

marcodejongh:
## Fable review sign-off — code clear, awaiting device QA ⚠️

Adversarial Fable review verdict: **SAFE_TO_MERGE on the code** — no critical or major findings in the diff.

- The fix is verified correct by construction: it trusts UIKit's composed `safeAreaInsets.bottom` as the single source of truth on the NativeTabs path (iOS 26 folds the 49pt bar + 56pt accessory into it; the old code double-counted). All non-native paths (Material, JS fallback, sidebar iPad) confirmed bit-for-bit unchanged algebraically.
- All ~35 consumers of the metrics grepped: none hand-compensates the old double-count. Both session-detail stacks are symmetric by construction (shared component). Toast now derives from the same predicates as the layouts.
- Rebase onto main verified trivially clean (zero overlap with the 53 commits since the merge base). **The `native-fingerprint` label is a confirmed stale-mergebase artifact** — the diff touches no native inputs; this ships OTA.
- `vp run test:mobile`: 4,917 tests pass; the #3973 regression test pins the device-measured 139pt inset.
- **Remaining gate (the PR's own draft checklist): the five physical iOS-26 device checks** — only the with-accessory inset value is device-measured; the no-accessory, minimized-tab, and pushed-route inset states are inferred and rest on `useSafeAreaInsets()` faithfully propagating UIKit's extended inset. Merge after those pass.

PR #4091, opened 2026-07-31
fix(mobile): serialize remix handoff across native sheets

### LINKED ISSUE #4076: Remix path presents create under a live native sheet (sheet-coordinator bypass)
Split out of the #3804 investigation. Suggested priority: **P2** (no confirmed crash attributed to it yet, but it violates a hard rule the repo documents precisely because the failure mode is a full app freeze).

`docs/mobile-sheets-vs-routes.md` rule 1: every native sheet must go through the presentation coordinator, because "overlapping a present with another sheet's dismiss deadlocks UIKit and freezes the whole app". The remix path breaks this in two separate places.

## 1. Remix from a presented sheet never dismisses that sheet

`useCreateClimbNavigation` only knows about `/play`:

`packages/mobile/src/components/create-climb/use-create-climb-navigation.ts:50-56` — `if (isPlayerRoute(...)) router.dismiss(); router.push('/(tabs)/climbs/create', params)`

It has no notion of "a native sheet is currently presented". Two entry points open the reaction menu from a live `BottomSheetModal` and never request a close:

- **Board sheet** — `packages/mobile/src/providers/drawer-host-provider.tsx:698-703` (`handleBoardSheetOpenActions` → `openClimbActions(...)`, no `requestCloseBoardSheet()`).
- **Queue sheet** — `packages/mobile/src/components/play-drawer/use-queue-sheet-handlers.ts:63-69` (`handleOpenActions` → `openClimbActions(item.climb)`, no `requestCloseQueueSheet()`).

So Remix pushes the create route **underneath** a live sheet, and create then presents a **second** root-window sheet on top. Extra wrinkle on the queue-sheet path: `router.dismiss()` destroys the `/play` VC *and the QueueSheet host it presented*, so `onFullyDismissed` never fires and the coordinator falls through to its ceiling timer — the "Host torn down mid-animation" case the doc names by hand.

This shape predates #3337 and is still on `main`.

## 2. #3337's dismiss/push races CreateDrawer's uncoordinated sheet

`CreateDrawer.tsx:130` presents a raw `@expo/ui/community/bottom-sheet` `BottomSheet` — a real SwiftUI `.sheet()` off the key window — **deliberately bypassing `SheetPresentationProvider`** (a documented exception).

#3337 added `router.dismiss()` immediately before `router.push()`, so `/play`'s ~350–500 ms UIKit dismissal is now in flight while create's first commit presents that uncoordinated sheet. The coordinator cannot serialise the two because it has no visibility into create's sheet at all. On the player path the `ClimbReactionMenu` `FullWindowOverlay` is also torn down ~150 ms into the same window.

This is **new since 2026-07-25**, so it is not the #3804 crash (which predates it) — but #3337's own QA checklist called this gate "the risky one" and proposed a `requestAnimationFrame` defer if it misbehaved, and I don't think that device QA was ever signed off.

## Suggested fix

Give `useCreateClimbNavigation` a single "close whatever native surface is presented, wait for settle, then push" step, and/or bring `CreateDrawer` under `SheetPresentationProvider` so the coordinator can serialise it. Architectural enough to want a deliberate design pass rather than a point patch — which is why it isn't in the #3804 PR.

An in-flight `useRef` guard around `navigateToCreate` is worth folding in at the same time: `ClimbReactionMenu.tsx:561` binds `onPress={action.run}` with no `disabled`, and the overlay stays hit-testable through its 150 ms fade, so a double-tap fires two `dismiss()` + two `push()`.

marcodejongh:
## Summary

- add a coordinator-owned `dismissAndWait` contract for native sheets, including sheets that are still presenting, duplicate waiters, teardown aborts, and a bounded fallback
- let the `/play` route wait for its own closing `transitionEnd` event before the create drawer is pushed
- serialize Remix/Edit as source sheet → player → create, while preserving the iPad inline-pane path

The remix path could previously close its visible menu and push create without waiting for the native Board/Queue sheet or the player route to finish dismissing. That allowed CreateDrawer to present underneath a still-live native sheet, leaving a stranded scrim or an apparently dead action.

Fixes #4076

## Validation

- independent adversarial review: no findings
- focused mobile tests: 8 files, 136 tests passed
- `vp run typecheck:mobile`
- `git diff --check origin/main...HEAD`

## Device QA

- iPhone: exercise BoardSheet and QueueSheet → player → Remix/Edit with slow animations and repeated taps
- Android: confirm the bounded fallback never leaves a stuck sheet
- iPad regular width: confirm inline panes do not dismiss an unrelated route

## Release Notes

Remixing or editing a climb now waits for the current board, queue, and player screens to finish closing before the editor opens, preventing stranded sheets and taps that appear to do nothing.

### DIFF
--- a/docs/mobile-sheets-vs-routes.md
+++ b/docs/mobile-sheets-vs-routes.md
@@ -234,10 +234,13 @@ modal, and only then pushes create:
 claim action → dismiss source sheet → settle → dismiss /play → closing transitionEnd → push create
 ```
 
-`dismissAndWait()` can return `aborted` when its owner unmounts; that aborts the rest of the
-handoff. This prevents a stale async action from opening create after its source disappeared.
-Repeated taps join neither a second dismiss nor a second push because the action guard is set
-before the first asynchronous boundary.
+`dismissAndWait()` can return `aborted` when its source-sheet owner unmounts; that aborts the
+rest of the handoff. The `/play` route has one deliberate exception: after its own callback has
+called `router.dismiss()`, route teardown is the expected close path, so its bounded native-stack
+ceiling remains alive when `transitionEnd` cannot arrive from the unmounted screen. A stale player
+callback invoked _after_ the player already disappeared still returns `aborted` without calling
+`router.dismiss()` again. Repeated taps join neither a second dismiss nor a second push because
+the action guard is set before the first asynchronous boundary.
 
 `handleSwitchBoardFromDrawer` in `drawer-host-provider.tsx` does the same
 `router.dismiss()` → `router.push('/boards')` when the play drawer's board-mismatch overlay
@@ -252,10 +255,11 @@ Two things that are easy to get wrong:
 
 - **Let the route own its close waiter.** The actual `/play` route subscribes to its native
   stack's `transitionEnd` before calling `router.dismiss()`, ignores events where `closing` is
-  false, and cleans up/returns `aborted` if the route unmounts first. On web there is no native
-  transition to await, so dismissal completes immediately. Thread this callback down through
-  the in-tree opener; a hook mounted in a sibling root provider cannot safely subscribe to the
-  `/play` navigator.
+  false, and uses its bounded ceiling if that dismissal unmounts the route before the native event
+  can reach JS. If the route was already gone before the callback starts, it returns `aborted`
+  instead of popping the now-visible route underneath. On web there is no native transition to
+  await, so dismissal completes immediately. Thread this callback down through the in-tree opener;
+  a hook mounted in a sibling root provider cannot safely subscribe to the `/play` navigator.
 - **Omit the player callback when there is no player route.** The same menu opens from the
   climbs list and BoardSheet, and on an **iPad regular-width layout the player is an inline
   detail pane**. Those entry points must not call `router.dismiss()` or they can pop an unrelated
--- a/docs/mobile-sheets-vs-routes.md
+++ b/docs/mobile-sheets-vs-routes.md
@@ -114,6 +114,13 @@ and on iOS it's the ceiling for the rare case the native event never arrives (a 
 mid-animation). A `__DEV__` warning fires only when the ceiling beats a native signal that was
 expected (an iOS dismiss of a still-registered sheet).
 
+Call `managed.dismissAndWait()` when the next step must not begin until that settle point. It
+uses the coordinator's existing native-dismiss-or-ceiling lifecycle rather than starting a
+second timer, and it also waits when `present()` is still in flight. Concurrent callers for the
+same sheet share the outcome. The promise resolves with `{ status: 'dismissed' }` after a normal
+settle (or the platform ceiling), and `{ status: 'aborted' }` if the sheet unregisters during the
+handoff. Stop the follow-up action on `aborted`; the owning surface has gone away.
+
 The same patch also guards the **Android** re-snap path: `snapToIndex` on an already-open
 multi-detent sheet fires `expand()` / `partialExpand()` fire-and-forget on the native
 `ModalBottomSheetView`. A store binary whose native `@expo/ui` predates one of those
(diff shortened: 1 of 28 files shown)


marcodejongh:
Addressed the callback-stability note in 2d3475bce by reading the latest route navigation object through a ref; the callback identity is now pinned by a regression test. The action-guard report assumed ClimbReactionMenu stays mounted, but DrawerHost conditionally renders it only while climbActions is non-null, so every close unmounts the hook and every reopen gets a fresh guard. I added an explicit unmount/reopen regression test for the same climb instead of introducing a visible-reset API that could reopen the guard during the exit animation. Focused follow-up tests: 31/31; mobile typecheck and vp check pass.

marcodejongh:
Addressed the second review in f0d0ba31b. Async handoff failures are now caught and reported to Sentry without reopening the one-action guard during the still-hit-testable exit animation; the next presentation/remount re-arms it. ModalSheet now publicly forwards ManagedSheetHandle, so dismissAndWait is type-checked end to end and every imperative ref owner was migrated off BottomSheetModal. Focused regression tests: 33/33; mobile typecheck and targeted vp check pass. The repository-wide check currently reaches unrelated existing lint warnings and a Vite+ stdout Resource temporarily unavailable panic; none are in these files.

marcodejongh:
Addressed both latest follow-ups in e4d2d4b3c. Missing QueueSheet or BoardSheet handles now resolve as already dismissed through one shared helper; aborted remains reserved for a mounted host disappearing during an in-flight close. The player route waiter now has an 800 ms fail-safe for a lost native transitionEnd event and clears both its timer and listener on every settle path. Focused validation: 61 tests passed, targeted vp check passed, and mobile typecheck passed.

marcodejongh:

marcodejongh:
🤖 Independent review from the `josh` loop: **HOLD — one blocker, with the fix mapped out.**

**Blocker: Remix/Edit from /play will never open the create editor with this diff.** `usePlayerDismissAndWait` waits for `transitionEnd{closing:true}` on the dismissed screen's own navigation object — but `router.dismiss()` is a plain JS POP: the /play subtree (including the hook) unmounts within a tick, BEFORE the ~400ms UIKit animation, and react-native-screens' own source says an unmounted screen has no event emitter (RNSScreen.mm notifyDisappear no-op; unmountChildComponentView swaps in a snapshot). So the hook's unmount-cleanup abort always wins → `completeHandoff` returns on `aborted` → `router.push` never runs (use-create-climb-navigation.ts:66-70). The tests pass because the mocks never model unmount-before-animation — they pin the dead-action behavior as correct. Affects the /play long-press path, /play QueueSheet path, and Android ClimbActionsSheet.

**Fix direction:** treat player-route unmount as the EXPECTED settle for a JS-driven dismiss (not an abort), plus a bounded ceiling not cancelled by the hook's unmount (~550ms iOS / ~350ms Android), or listen on a surface that outlives the popped route (the tabs screen's transitionEnd{closing:false}/focus). Also: (2) the player waiter has no timeout at all today — the SETTLE_MS ceiling only covers native sheets, so a missing event hangs the handoff with actionInFlightRef claimed; (3) nit: ClimbActionsSheet re-arms its guard on every `visible` flip → possible double-push if reopened mid-handoff — reset in completeHandoff's finally instead; (4) the native-fingerprint label is stale-base noise (TS+docs diff; rebase clears it).

**Cleared:** the sheet half is genuinely right — dismissAndWait resolves on notifyFullyDismissed with a platform ceiling (not a visible-prop effect), both issue paths are wired, reentrancy is guarded synchronously, iPad inline pane correctly excluded, #3337's CreateDrawer serialization-only mitigation is a documented acceptable partial. One device-QA run of Remix-from-player would confirm the blocker in 30 seconds — the PR's Device QA checklist is untouched.

marcodejongh:
Addressed the independent-review blocker in cdbe61ed7. The player waiter now distinguishes a stale callback after an unrelated /play unmount (aborted, no dismiss) from the expected synchronous unmount caused by its own router.dismiss(). In the expected case the native listener is detached but the closure-owned platform settle timer survives, so create navigation proceeds only after the transition ceiling.

New coverage models router.dismiss() synchronously removing /play, verifies the persistent parent pushes the create route exactly once after settlement, and separately proves a stale callback cannot pop the route underneath. Focused lifecycle/navigation tests: 17/17; mobile typecheck, targeted vp check, and git diff --check pass. The branch is also rebased on current main.

marcodejongh:
Addressed the final null-handle review in `b10fda02a`. ClimbActionsSheet now delegates to the canonical `dismissManagedSheetAndWait` helper, so an absent source surface means already dismissed and create navigation continues; `aborted` remains reserved for a mounted host disappearing mid-dismiss. The new regression nulls the real imperative ref and would fail on the prior early-return behavior. Independent QA passed with no findings: 15/15 focused tests, mobile typecheck, targeted check, and diff check.

marcodejongh:
Addressed the two final non-blocking review notes in d31dd7c75. A fresh setDesiredOpen(id, true) now explicitly aborts any pending caller-owned dismiss waiters, so a genuine concurrent re-open cannot strand the handoff promise or navigate away from a re-presented source. ClimbActionsSheet now also has an integration regression proving a host teardown during dismissal stops before player dismissal and route push. Independent QA passed: 43/43 focused tests, vp run typecheck:mobile, targeted vp check, and git diff --check.

PR #4092, opened 2026-07-31
chore(mobile): record Live Activity intent exit diagnostics

### LINKED ISSUE #4077: Investigate Live Activity intent background cold-launch as a source of short-uptime, no-log terminations
Split out of the #3804 investigation. Suggested priority: **P2** (speculative — no crash is yet attributed to it — but it is the best structural explanation we have for a recurring crash *shape*, and it is currently uninstrumented).

## Why this is worth its own investigation

Several TestFlight crash reports share a signature that none of our JS-side theories explain well: **very short app uptime** and **no crash log returned by App Store Connect**. #3804 is the clearest example — 4 s uptime, no log, on a flow (remixing a climb) that takes many taps to reach.

While tracing #3804 I ruled out the obvious short-uptime explanations:

- **No navigation-state restoration exists anywhere** in the app (no `getInitialState` / `restoreState` / `persistNavigationState`). A cold launch always lands on `/(tabs)/home` or `/auth/login` — it cannot resume a deep in-app screen.
- **No automatic OTA reload** — no `checkAutomatically`, no `fallbackToCacheTimeout`, no non-user-initiated `reloadAsync`. The JS runtime is not restarting under the user.

That leaves the app being started **without a normal foreground launch**.

## The candidate

iOS runs a `LiveActivityIntent`'s `perform()` **in the main app process** when the intent is registered there (this is already documented in our own notes on Live Activity intent routing). A user tapping a button on a Live Activity can therefore **cold-launch the whole app in the background** to service it.

The intent path does `@MainActor` BLE + network work behind a ~3 s ready timeout inside a `beginBackgroundTask`. A background launch given only a short grace window, doing main-actor BLE and HTTP, is a very plausible source of terminations that (a) happen seconds after process start and (b) produce no user-visible crash and no ASC crash log.

It would also explain why the user's *description* of the crash ("remixing climb") can be genuine while the reported uptime is tiny — the process that died was not the one they were using.

## What to actually do

1. Instrument it: confirm whether these background-launched processes exist in the wild at all, and how often they terminate. Sentry will not report a jetsam, so this needs an explicit breadcrumb/marker written on intent entry and cleared on clean completion, then read on the next launch.
2. Audit `perform()` for main-actor blocking work and for what happens when the 3 s timeout expires.
3. Confirm whether a background launch can end up mounting the full React tree (and therefore the board-art image stack) rather than a minimal path.

Relevant files: `packages/mobile/modules/live-activity/**`.

Note the practical constraint: this is native code, so any fix is fingerprint-moving and cannot ship OTA — it needs to ride a store binary.

marcodejongh:
## Summary

- record bounded, schema-versioned Live Activity intent stages and coarse outcomes in the main app target only
- consume only expired, incomplete records from a prior process, once, after the React root mounts or the app returns to the foreground
- report one privacy-safe informational Sentry event with a fixed fingerprint so cold-launch exits can be measured without changing AppIntent, BLE, timeout, or background-task behavior

This is an investigation instrument for the short-uptime, no-log terminations tracked in #4077. It deliberately does not claim a behavioral fix: the next physical-device occurrence should now identify the last completed stage instead of disappearing without evidence.

Fixes #4077

## Validation

- independent adversarial review: no findings
- focused mobile tests: 3 files, 45 tests passed
- `vp run typecheck:mobile`
- `vp check`
- Foundation-only Swift recorder parse/typecheck on Linux
- target-drift coverage verifies widget copies cannot compile recorder calls
- `git diff --check origin/main...HEAD`

## Device QA

- compile and run both generated iOS app and widget targets
- cold-launch each Live Activity intent and exercise normal early returns
- verify the 30-second grace and next-foreground once-only Sentry event on a physical iPhone/TestFlight build

## Release Notes

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -85,6 +85,7 @@ Boardsesh is a monorepo. Next.js 16 web app + React Native (Expo) mobile app for
 - No AI-generated images. Real photos or diagrams only.
 - No buzzwords. Concrete numbers, plain language.
 - Default to action. Full autonomy except no data deletion without asking.
+- **Bluetooth changes require a Fable review.** Any change touching Bluetooth/BLE code — `packages/shared/ble-protocol/`, `packages/mobile/src/lib/ble/`, `packages/mobile/modules/live-activity/ios/`, or the web Bluetooth LED control in `packages/web` — must be reviewed by Fable before merge.
 
 ## PR Lifecycle (mandatory)
 
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -14,7 +14,7 @@ Boardsesh is a monorepo. Next.js 16 web app + React Native (Expo) mobile app for
 - No AI-generated images. Real photos or diagrams only.
 - No buzzwords. Concrete numbers, plain language.
 - Default to action. Full autonomy except no data deletion without asking.
-- **Bluetooth changes require a Fable review.** Any change touching Bluetooth/BLE code — `packages/shared/ble-protocol/`, `packages/mobile/src/lib/ble/`, or the web Bluetooth LED control in `packages/web` — must be reviewed by Fable before merge.
+- **Bluetooth changes require a Fable review.** Any change touching Bluetooth/BLE code — `packages/shared/ble-protocol/`, `packages/mobile/src/lib/ble/`, `packages/mobile/modules/live-activity/ios/`, or the web Bluetooth LED control in `packages/web` — must be reviewed by Fable before merge.
 
 ## PR Lifecycle (mandatory)
 
--- a/docs/live-activity-push-testing.md
+++ b/docs/live-activity-push-testing.md
@@ -440,6 +440,11 @@ Connect your iPhone and open Xcode > Window > Devices and Simulators > your devi
 
 ### Interrupted intent diagnostics
 
+The first deployment of these diagnostics requires a new iOS binary: the
+durable store and App Intent recorder live in Swift and cannot be activated by
+an OTA bundle alone. Once that binary is installed, the React-side consumer and
+reporting changes can follow the normal fingerprint-compatible OTA path.
+
 The main app target keeps a best-effort marker for each `LiveActivityIntent`
 run (`NextClimbIntent`, `PreviousClimbIntent`, `TakeControlIntent`, and
 `ReconnectBoardIntent`). The widget-extension copies compile the recorder calls
--- a/docs/live-activity-push-testing.md
+++ b/docs/live-activity-push-testing.md
@@ -438,6 +438,53 @@ Connect your iPhone and open Xcode > Window > Devices and Simulators > your devi
 - `LiveActivityPlugin` — token registration with backend
 - `NativeWebSocketPlugin` — WebSocket connection state
 
+### Interrupted intent diagnostics
+
+The main app target keeps a best-effort marker for each `LiveActivityIntent`
+run (`NextClimbIntent`, `PreviousClimbIntent`, `TakeControlIntent`, and
+`ReconnectBoardIntent`). The widget-extension copies compile the recorder calls
+out with `#if !WIDGET_EXTENSION`; widget fallback execution never writes the
+main app's store.
+
+The marker is an app-local `UserDefaults` envelope, not App Group storage. It is
+schema-versioned, limited to the newest 16 records, and contains only generated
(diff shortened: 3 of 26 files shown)


marcodejongh:
Addressed all three review notes in c6647d338: the TS stage union now includes completed, consumed markers are TTL-pruned and bounded from the shared prune path, and an unmount test proves listener cleanup plus disposal of a late root-marker continuation. Focused follow-up tests: 45/45; mobile typecheck and vp check pass. I also replaced the literal 'none' release note with the template's checked internal-change box, which fixes the release-notes gate.

marcodejongh:
Followed up on the non-blocking review in c61137a2a: repeated foreground notifications now have explicit single-flight coverage, the widget drift checker correctly preserves outer guards through nested conditionals and treats #elseif as its own branch, and the device runbook now states that first activation requires a new iOS binary. The consumed-key ring remains as deliberate defensive durability; its bounded size is harmless even though atomic record removal is the primary dedupe. Focused tests: 18/18; mobile typecheck and targeted vp check pass.

marcodejongh:
🤖 Independent review from the `josh` loop: **code is sound; correct handling is exactly what it's already doing** (draft + native-fingerprint label, rides the next native train — do not merge standalone).

Verified: storage is genuinely bounded (16-record ring, TTL, schema/build validation — all pinned by fail-on-revert tests); nothing new blocks the intent's 3s watchdog path (readyTimeout and the ready computation are byte-identical — the diff only threads the existing Bool out for the diagnostic mark; marks are bounded in-memory writes to app-local UserDefaults, no synchronize()); payloads are identifier-free (asserted by test) and Sentry gets info-level with a fixed fingerprint, so crash-free stats can't inflate; and the lastStage progression + reactRootMounted delivers precisely what #4077 asked to learn.

**Fable review (BLE rule):** BoardBleManager.swift / LiveActivityBleBridge.swift live outside CLAUDE.md's Bluetooth-review glob but are real CoreBluetooth files — reviewed: the BLE-adjacent diff is behaviorally inert plumbing, no timing/connection-lifecycle change, approved on that basis. @marcodejongh two follow-ups: (1) CLAUDE.md's Bluetooth glob should gain `packages/mobile/modules/live-activity/ios/` so this path can't slip review in a future PR; (2) the device-QA checklist (physical-iPhone cold-launch) still needs doing before this rides a train — CI's XCTest target compiles the widget flavor and can't exercise the perform() control flow.

marcodejongh:
Addressed the latest review follow-ups in 560f8c6f7. The enabled Sentry path now has direct emission coverage without weakening the production enablement gate. Reconnect diagnostics preserve an earlier BLE failure when take-control also fails, with Swift tests covering both precedence directions; both widget intent copies remain byte-identical. LiveActivityBleBridge.reconnectForIntent already returns Bool, so no production change was needed there. Focused JS tests: 43/43; targeted vp check and mobile typecheck pass. The new Swift tests will compile and run in the existing macOS iOS-test job.

marcodejongh:
Completed the independent BLE-review follow-up in 7c54e36ad. Both CLAUDE.md and the copied AGENTS.md project guidance now include packages/mobile/modules/live-activity/ios/ in the mandatory Fable-review scope, so the CoreBluetooth bridge cannot bypass that gate in future changes. Independent QA confirmed the guidance is consistent and narrowly scoped.

The branch is rebased on current main; 48 focused mobile tests pass and all three generated/source AppIntent copies remain byte-identical. Physical-iPhone cold-launch QA is still intentionally outstanding, so this PR remains draft for the next native train.

marcodejongh:
Pushed the post-review remediation in 5b4eb9aa5. The interrupted-run bridge now uses a native interrupted-only DTO and stage enum, so completed and completionClass are structurally unrepresentable and invalid stored records fail closed. The specific BLE request must also report success and the global queue must drain inside the existing timeout. Independent Sol xhigh review passed; Sol xhigh was used as the closest available substitute because the requested Fable reviewer is not exposed in this environment. Validation passed: 49 focused tests, mobile typecheck, targeted vp check, diff check, and identical source/widget intent copies. Swift XCTest execution and the physical-iPhone cold-launch path remain CI/device gates, so this PR intentionally remains draft.

marcodejongh:
The post-merge Swift concurrency hardening from the latest review is now tracked in #4169. It adds explicit Sendable contracts and a strict-concurrency compile gate; independent QA caught and repaired the injected-clock isolation issue before the follow-up PR was opened.

PR #4093, opened 2026-08-01
fix(backend): validate playback state broadcasts

### LINKED ISSUE #2428: backend: add Zod validation for PlaybackStateInput
Follow-up from PR #2363 / issue #2232.

`publishPlaybackState` accepts the input verbatim and broadcasts it to all peers. Malformed values (negative frameIndex, NaN speed, absurd paceMs) would propagate without validation. Other mutations in this resolver group use Zod schemas — add one for `PlaybackStateInput` and bounds-check before publishing.

marcodejongh:
## Summary

- Validate every `PlaybackStateInput` field before a party-session playback event is published.
- Enforce the bounds already shipped by the mobile/web controls and shared playback engine.
- Keep malformed direct-resolver calls from reaching peers or the pubsub layer.

Closes #2428

## Release Notes

## Test plan

- `vp test run --project backend packages/backend/src/__tests__/playback-state-validation.test.ts packages/backend/src/__tests__/publish-playback-state.test.ts --reporter=agent` (40 passed)
- `vp run typecheck:backend`
- `vp check`
- Independent QA verified shipped client/engine bounds and the no-publish-on-invalid-input invariant.

### DIFF
--- a/packages/backend/src/__tests__/playback-state-validation.test.ts
+++ b/packages/backend/src/__tests__/playback-state-validation.test.ts
@@ -16,6 +16,7 @@ function playbackStateWith(overrides: Record<string, unknown>): unknown {
 
 describe('PlaybackStateInputSchema', () => {
   it('accepts the shipped playback control and GraphQL Int boundaries', () => {
+    expect(PlaybackStateInputSchema.parse(playbackStateWith({ speed: 0.5 })).speed).toBe(0.5);
     expect(
       PlaybackStateInputSchema.parse(
         playbackStateWith({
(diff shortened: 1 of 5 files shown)


marcodejongh:
Added both suggested coverage gaps in b492de4bf: the exact 0.5× lower bound and every omitted required field. Focused total is now 45/45; backend typecheck and targeted vp check pass. I kept validation after applyRateLimit/session authorization intentionally: it follows the resolver convention and prevents malformed callers from bypassing the same abuse-control and authorization work applied to valid broadcasts.

marcodejongh:
🤖 Independent review from the `josh` loop: **MERGE — no blocking findings.** Verified: exact match to #2428's ask (validateInput before the pubsub publish, same convention as the 9 sibling call sites); a rejected payload is scoped to the offending operation only (standard graphql-ws per-op error handling, throw precedes publishQueueEvent, and the shared client wraps the call best-effort — nobody's session dies); bounds checked against BOTH deployed clients' history — speed [0.5,10] and paceMs≥200 have never drifted since the feature's original commit, so no compliant fleet client can trip the new validation; reject tests assert both the throw and that publish never fires (fail-on-revert, 45/45 locally). One non-blocking note for the record: the shared engine's own setSpeed clamps only a 0.1 floor — enforcement now lives server-side + UI, engine contract is looser (pre-existing, out of scope). Merging.

PR #4094, opened 2026-08-01
fix(sync): surface circuit ownership conflicts

### LINKED ISSUE #3950: Circuits sync: the SQL race-guard skip has zero observability, and the duplicate-account copy can't distinguish foreign from ambiguous
Two small follow-ups deliberately left out of #3931 to keep its diff surgical — that file is also touched by #3871.

## 1. The race-guard skip is silent

In `packages/aurora-sync/src/sync/user-sync.ts`, the SQL race-guard path:

```ts
if (!playlist) continue;
```

has no log line and does not call `refusedCircuitUuids.add()`. So when the narrow race window from #3539 is actually hit, **nothing records it** — the circuit is skipped with no counter, no log, and no entry in the refused set that drives the user-facing status.

That is the same shape as the bug #3523 was filed for: a skip that only exists as control flow is a skip nobody discovers for years. It should at minimum log and join the refused set so it surfaces the same way every other skip does.

Ordering note: land this **after** #3931 and #3871, both of which touch this file.

## 2. The duplicate-account copy covers two cases with one sentence

#3931 ships a single string for the paused-playlist-sync state:

> another Boardsesh account is connected to this board login, so playlist sync is paused for both of you

That is true for both the *foreign-owned* and the *ambiguous* case, but it can't tell a climber which one they're in or what to do about it. Splitting them means threading the distinction out of `hasForeignOwnedCircuitPlaylists`, which is a refactor in the same file — hence deferred.

Only worth doing if the combined wording actually confuses people in practice; the single string is honest, just imprecise.

marcodejongh:
## Summary

- Acquire circuit arbitration locks in a stable order before any source or playlist writes, including callers that already own a transaction.
- Make race losses and existing foreign or ambiguous ownership visible through distinct persistent sync errors.
- Preserve legitimate orphan-adoption behavior while preventing a losing sync from changing another account's playlist or climbs.
- Show honest recovery guidance in web and mobile settings in English, Spanish, French, and German.

Closes #3950

## Release Notes

Playlist sync now tells you when another Boardsesh account owns the circuit or when ownership needs support to untangle, instead of silently skipping it.

## Test plan

- Aurora sync suite: 209 passed, 2 skipped without a fully migrated `DATABASE_URL`
- Web Aurora sync suite: 116 passed
- Shared sync-error tests: 2 passed
- i18n catalog completeness: settings.json at parity across `es`, `fr`, `de`
- Scoped `vp check` on the changed files: 0 errors
- `vp run typecheck`: clean
- Independent adversarial review: pass
- [x] Run the two concurrency cases against a fully migrated PostgreSQL schema — the new `aurora-circuit-arbitration` CI job boots the dev-DB image and executes both the daemon and the daemon/web cross-writer test with `REQUIRE_AURORA_CIRCUIT_INTEGRATION=1`
- [x] Confirm normal CI typecheck/test/mobile-bundle gates
- [ ] Start the full dev server after #3978's dev-DB migration-ledger mismatch is resolved; current `vp run dev` stops before startup because migration `0187_sad_freak` finds `board_climb_ingest_skips` already present

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -52,6 +52,7 @@ jobs:
       # `listing-guards` job (see below).
       storeMetadata: ${{ github.event_name != 'pull_request' && 'true' || steps.filter.outputs.storeMetadata }}
       mobileLocales: ${{ github.event_name != 'pull_request' && 'true' || steps.filter.outputs.mobileLocales }}
+      auroraCircuitArbitration: ${{ github.event_name != 'pull_request' && 'true' || steps.filter.outputs.auroraCircuitArbitration }}
       # PR-only guard signal: did this PR edit the OTA-owned changelog data file?
       # Defaults false off PRs (the OTA bot legitimately rewrites it on push).
       changelogData: ${{ github.event_name == 'pull_request' && steps.filter.outputs.changelogData || 'false' }}
@@ -63,7 +64,7 @@ jobs:
       # non-JS files (SQL migrations under packages/db, .graphql under
       # shared-schema) that wouldn't trigger anyJs but still need setup so
       # test-backend / mobile-bundle / codegen-drift can run.
-      runAny: ${{ github.event_name != 'pull_request' || steps.filter.outputs.anyJs == 'true' || steps.filter.outputs.web == 'true' || steps.filter.outputs.i18nCatalogs == 'true' || steps.filter.outputs.sharedSchema == 'true' || steps.filter.outputs.sharedDeps == 'true' || steps.filter.outputs.backend == 'true' || steps.filter.outputs.mobile == 'true' || steps.filter.outputs.ocr == 'true' || steps.filter.outputs.locationSync == 'true' || steps.filter.outputs.rootCi == 'true' }}
+      runAny: ${{ github.event_name != 'pull_request' || steps.filter.outputs.anyJs == 'true' || steps.filter.outputs.web == 'true' || steps.filter.outputs.i18nCatalogs == 'true' || steps.filter.outputs.sharedSchema == 'true' || steps.filter.outputs.sharedDeps == 'true' || steps.filter.outputs.backend == 'true' || steps.filter.outputs.mobile == 'true' || steps.filter.outputs.ocr == 'true' || steps.filter.outputs.locationSync == 'true' || steps.filter.outputs.auroraCircuitArbitration == 'true' || steps.filter.outputs.rootCi == 'true' }}
       # NOTE: deployConfig, storeMetadata, and mobileLocales are deliberately
       # NOT part of runAny — the deploy-config and listing-guards jobs below
       # need [changes] only (no bun-install/build:db setup pass), so a PR
@@ -150,6 +151,19 @@ jobs:
             mobileLocales:
               - 'packages/mobile/locales/**'
               - 'scripts/mobile-locales-parity.test.ts'
+            # The real-database circuit ownership assertions only need the
+            # Aurora sync paths and their direct shared database contracts.
+            # Root manifests/workflow config are already covered by rootCi.
+            auroraCircuitArbitration:
+              - 'packages/aurora-sync/**'
+              - 'packages/web/app/lib/data-sync/aurora/**'
+              - 'packages/db/src/client/**'
+              - 'packages/db/src/queries/**'
+              - 'packages/db/src/schema/**'
+              - 'packages/sync-runtime/**'
+              - 'packages/shared-schema/src/sync-error-codes.ts'
+              - 'packages/shared-schema/src/__tests__/sync-error-codes.test.ts'
+              - 'packages/{aurora-sync,db,sync-runtime,shared-schema,web}/package.json'
             rootCi:
               - '.github/workflows/ci.yml'
               - 'vite.config.ts'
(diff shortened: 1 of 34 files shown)


marcodejongh:
Addressed the latest review and rebased onto current `main` in `75ce6330b`. Circuit arbitration now uses PostgreSQL's single-`bigint` advisory lock with server-side `hashtextextended(..., 0::bigint)` in both production and the independent contention probe; the impossible suppressed-own path emits a separate payload-free error event; and the source-only `!nextAuthUserId` path is pinned to one lock/source upsert with no playlist-side work. Focused Aurora tests pass 26/26 (2 opt-in PostgreSQL cases skipped without a migrated DB), scoped checks and `git diff --check` pass, and the independent adversarial re-review found no correctness, deadlock, privacy, or false-green issue. Fresh CI is running; the migrated-PostgreSQL and #3978 dev-server gates remain explicit.

marcodejongh:

marcodejongh:
🤖 Independent review from the `josh` loop: **HOLD — three findings, first two are one-line-ish fixes, third needs a call.**

1. **The new refusal logs never reach prod.** The daemon's non-verbose onLog is a substring allow-list (cli/index.ts:17-36) and none of the new lines — logCircuitRefusal's two events or the runner's 'circuits not syncing' line — contain an allow-listed token. Prod daemon runs without -v, so the race-guard skip stays exactly as invisible as issue #3950 complains. Fix: add `msg.includes('aurora_circuit_playlist_')` to the allow-list, same pattern as the CREDENTIAL/Sync-health entries.
2. **logError is never wired.** upsertTableData grew the param but syncUserData passes seven args — the suppression event goes to bare stderr, bypassing the runner's onError (the sink that attaches failure-ledger context). Thread it through syncUserData from syncSingleCredential.
3. **Arbitration isn't total: the web-proxy writer (third writer, reachable from the login + user-sync proxy routes) takes no lock and iterates payload-order while the daemon now sorts by uuid** — the #3539-class race stays open daemon-vs-web AND the ordering divergence is a new deadlock vector (opposite row-lock order on circuits {A,B} → Postgres aborts one). Either take the same advisory lock in the web branch, or scope it out explicitly and restore payload-order parity so the orders can't cross. The web twin also still carries the literal silent `if (!playlist) continue;` skip the issue quotes.

Also for @marcodejongh: the new `duplicate-board-account-link:circuits:foreign/:ambiguous` sync_error codes are an undeclared forward-compat break — stale clients render them as a red raw-code error instead of the amber warning (both surfaces' fallback paths checked). Small exposure, worth a body note or keeping the legacy code on the wire. Nits (blind-fixable): single-bigint advisory form contradicts the documented two-int repo convention in climb-similarity.ts/push-tokens.ts (defensible — update the comments or follow them); normalizeCircuitItems dropped the string-uuid filter (missing uuid → locks on '…|undefined'); the concurrency integration test self-skips in CI (no DATABASE_URL in the aurora-sync job) so the lock semantics ship with zero executing coverage — the PR's own checklist box for it is unchecked; per-circuit N+1 (~100 locks + 100-200 selects per 100-circuit page) worth a look. Cleared: namespace collisions, ON CONFLICT target, lock scope vs the Aurora HTTP call, poison-pill starvation, ownership-state join rewrite, i18n, fingerprint, BLE.

marcodejongh:
Fable review pass — all four findings handled, branch rebased onto `main`.

**1. Merge conflict in `ci.yml` (critical).** Rebased onto `origin/main` and resolved so both sides survive: `changes.runAny` now lists `locationSync` **and** `auroraCircuitArbitration`, and `ci-status.needs` lists `test-location-sync-integration` **and** `aurora-circuit-arbitration`. Two more conflicts turned up in the sync writers themselves — main's `normalizePlaylistColor()` swap in both `packages/aurora-sync/src/sync/user-sync.ts` and `packages/web/app/lib/data-sync/aurora/user-sync.ts` landed on the exact lines this branch restructured; kept main's canonical colour formatting inside this PR's new per-circuit write path.

**2. German catalog missing the two new keys (critical).** Added `aurora.status.foreignAccountCircuits` and `aurora.status.ambiguousAccountCircuits` to `de/settings.json`, following `docs/i18n-german-glossary.md` (informal *du*, "Begehungen", never "senden") and mirroring the existing `duplicateAccountCircuits` phrasing. Catalog parity for `settings.json` passes across `es`, `fr`, `de`.

**3. Web `upsertTableData` relied on the caller for lock validity (minor).** Fixed rather than documented. The circuits branch now opens its own `db.transaction(...)` — a savepoint when the caller already has one, which is what `syncUserData` does — so `pg_advisory_xact_lock` can never run as a single-statement no-op if someone later calls it with a plain handle. New test: *opens its own transaction before taking any advisory lock*, plus the existing lock-ordering test now asserts the transaction opens first.

**4. Stale PR body (minor).** Locale list now says German too; the two satisfied test-plan boxes are ticked with what actually verifies them. The dev-server box stays open — it's still blocked on #3978.

One extra: the new `scripts/prepare-aurora-circuit-integration-db.ts` typed its opt-in helper as `NodeJS.ProcessEnv`, which this repo augments with a required `NODE_ENV`. That's fine under the PR's diff-lint but adds two errors to the push-to-main full-repo lint. Widened it to `Record<string, string | undefined>`; full-repo `vp check` is back to main's own 28 errors, none from this branch.

Heads-up unrelated to this PR: `main` is currently red. `de/boards.json` is 30 keys behind `en-US` (2 extra), which fails `catalog-completeness.test.ts` in `test-default`, and the full-repo lint step has 28 pre-existing errors. Neither reaches this PR's CI — PR runs are diff-scoped — but both need a fix on `main`.

marcodejongh:
Staying in draft on one blocker that isn't this branch's.

`test-default` is red on a single assertion out of 14,577:

```
FAIL |i18n| src/__tests__/catalog-completeness.test.ts
  > i18n catalog completeness > de > boards.json ships the same key set as en-US
```

`de/boards.json` on `main` is 30 keys behind `en-US` (plus 2 stale ones) — gym picker, duplicate-board, and offline-fallback copy from #4116/#4150. It fails identically on a clean `main` checkout, and `main`'s own CI has been red on it since 681a0481d. This branch doesn't touch `boards.json`; PR #4213 already fixes it and is green and review-ready. Deliberately not duplicating that work here — merge #4213, and a rebase turns this one green.

Everything else this branch is responsible for passes:

| check | result |
| --- | --- |
| `aurora-circuit-arbitration` | pass (2m26s) — both concurrency gates against real PostgreSQL |
| `test-backend`, `test-ocr`, `test-location-sync-integration` | pass |
| `lint`, `typecheck`, `codegen-drift`, `mobile-bundle` | pass |
| `i18n`, `db-migrations`, `commit-lint`, `release-notes` | pass |
| `claude-review` | pass |

The Mobile OTA Preview publish tripped an S3 `SlowDown` on first attempt; re-ran it and it's green.

marcodejongh:

marcodejongh:

Conflicting: packages/aurora-sync/src/sync/user-sync.ts, plus a modify/delete conflict on packages/web/app/components/settings/aurora-credentials-section.tsx (+test) — main deleted that component, this branch still edits it.

PR #4096, opened 2026-08-01
fix(sync): retry transient shared sync failures

### LINKED ISSUE #1976: shared-sync cooldown stamps on failure, locking out retry for up to 1h

marcodejongh:
## Summary

Closes #1976.

- Retry Aurora's shared board-catalog sync after a short cooldown when the upstream failure is transient, instead of making a network blip or 5xx response wait the full shared-sync window.
- Keep authentication, malformed-response, and other permanent failures on the normal cooldown so a permanently failing board cannot loop every cycle.
- Fence shared-sync completion with database-generated claim identities so an older Aurora or Kilter worker cannot overwrite a newer attempt's cursor.
- Keep Kilter catalog success and failure on its documented full cooldown. If Kilter loses completion ownership, log it and leave the newer cursor intact rather than reporting a false runner failure.
- Simplify the null-claim branches in both runners without changing their held-slot behavior.

## Release Notes

- Board climbs catch up sooner after temporary Aurora sync outages

## Test plan

- Aurora/Kilter error and runner coverage: 70/70 passed before rebase
- Shared cooldown compare-and-set integration against Postgres: 10/10 passed
- Post-rebase Aurora/Kilter runner regression: 55/55 passed
- `vp run typecheck:aurora`
- `vp run typecheck:kilter`
- `vp run typecheck:db`
- `vp run typecheck:backend`
- Targeted `vp check` and `git diff --check`
- Independent follow-up QA verified ownership-loss handling, the intentional Kilter cooldown divergence, and null-claim control flow

### DIFF
--- a/docs/aurora-sync.md
+++ b/docs/aurora-sync.md
@@ -8,7 +8,7 @@ The `@boardsesh/aurora-sync` package provides the shared sync implementation. It
 
 1. **CLI** - For local debugging and manual sync runs
 2. **Daemon CLI on a VM** - `aurora-sync daemon` as a long-lived process: picks one user per cycle, syncs their per-user tables, then runs a board-wide shared sync using their fresh Aurora token. This is the production deployment.
-3. **Backend / Vercel cron** - (Removed) Previously a `POST /sync-cron` backend handler and a `/api/internal/user-sync-cron` Vercel route; both removed in favour of the long-lived daemon, which the catalog piggyback's in-memory cooldown needs to survive across cycles.
+3. **Backend / Vercel cron** - (Removed) Previously a `POST /sync-cron` backend handler and a `/api/internal/user-sync-cron` Vercel route; both removed in favour of the long-lived daemon. The shared-sync cooldown is persisted in Postgres, so deploys and overlapping daemon instances share the same per-board gate.
 
 ## Architecture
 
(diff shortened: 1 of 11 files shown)


marcodejongh:
Addressed the latest review in d95fdf1e9. Kilter now logs a lost completion claim and leaves the newer cursor intact without calling onError. Both runners return directly from the null-claim branch, removing the duplicate post-try guard without changing held-slot behavior. Kilter intentionally keeps its documented full catalog cooldown on failure; the regression now explicitly asserts no nextCooldownMs override, and the PR summary/release note now describe only Aurora's transient shared-sync retry.

Independent QA found no implementation issues after that assertion. Post-rebase runner tests: 55/55; aurora/kilter/db/backend typechecks and targeted checks passed.

PR #4110, opened 2026-08-01
test(web): lock the import graph against the Expo Web teardown

marcodejongh:
Safety rail for the Expo Web Program teardown (phases 4–6). Adds a test, changes no product code.

Later PRs delete roughly 110k lines from `packages/web` — the whole queue / play-drawer / Bluetooth / climb-browsing stack — while `www.boardsesh.com` keeps serving marketing, climb views, board lists, gym, kiosk, admin, auth and the public API. The risk is a kept surface still importing something that goes. TypeScript catches that only once the file is actually gone, which is the worst moment to find out.

This walks the transitive import graph from every kept surface and fails on any edge into the delete set.

## Why a bespoke walker

Two edge kinds are invisible to grep, and both are live today:

- **`dynamic(() => import(...))`** — `components/board-page/climbs-list.tsx` is one. A `next/dynamic` edge fails at runtime, not typecheck, so deleting its target passes CI and breaks the page.
- **`.module.css`** — 19 files import `components/library/*.module.css`, several from surfaces we keep.

It also resolves `@/` aliases, relative paths, `index.*`, and `export ... from`.

Traversal **stops at the first delete-set node**. Walking through one pulls in the whole entangled set and the allowlist stops meaning anything.

## What it found

138 edges from kept surfaces into the delete set. The four the plan already knew about were there, plus considerably more:

| Edges | Into |
|---:|---|
| 20 | `components/library` |
| 15 | `components/queue-control` |
| 13 | (cut by W-11's shared-primitive lift) |
| 12 | `components/climb-actions` |
| 11 | `components/board-page` |
| 7 | `components/search-drawer` |
| 6 | `components/onboarding` |

The concentration is the useful part: `library`, `queue-control` and `climb-actions` account for a third of the graph, and every one of those edges is a delete that would have red-mained `main` without warning.

The test therefore starts red against today's tree, so it ships with `import-graph-allowlist.json` — one entry per importer/target pair, each carrying a reason and the PR id that cuts it. Entries drop out as the teardown proceeds; the file should be empty by the end of the delete wave. It keys on from+to rather than line number, since lines drift and the number is printed on failure anyway.

Delete any entry from the allowlist and the test goes red, naming the importer, the target and the line. Verified by mutation rather than by reading: dropping the `liked-climbs-list.tsx → board-page-skeleton.tsx` entry fails the suite with that edge named. A guard that cannot fail is worse than no guard.

`components/settings` is deliberately **not** in the delete set — W-21 keeps two of its sections, so listing it now would bake in a decision that has not been made.

## Release Notes

none


marcodejongh:
Fixed the `resolves the edge kinds a grep would miss` fragility — good catch, and it applied to both halves, not just the dynamic-import one.

`board-page/climbs-list.tsx` would have thrown `ENOENT` through `readFileSync` once W-16 lands, and `library/library.module.css` would have failed its `toContain` once W-11 lifts the kept rules out of `library/`. Both fixtures were in the delete set this test exists to guard.

Neither `existsSync`-and-skip nor swapping in a kept file worked: no kept **production** file in `packages/web` uses `dynamic(() => import(…))` (only two test files do), so there is no durable live example. The test now builds a scratch fixture in a temp dir — importer, dynamic-import target, and a `.module.css` — and resolves against that. What is under test is the resolver, and the resolver outlives every file it walks. Added a third assertion while there: a bare specifier (`next/dynamic`) must resolve to `null`, not a bogus path.

Load-bearing check: with `isDynamicImport` forced to `false`, the test fails (2 of 6). Restored, 6/6 green.

Issue #4113, opened 2026-08-01
renderedOverlays can serve a file:// URI CachePruner already evicted (silent blank overlay, no telemetry)


marcodejongh:
## What's happening

`use-native-climb-render.ts`'s in-memory `renderedOverlays` cache map can hand `<Image>` a `file://` URI for an overlay PNG that has already been evicted from disk, producing a silent blank/broken overlay with no crash and no telemetry.

## Where this comes from

- `renderedOverlays` is populated two ways: `warmupRenderedOverlaysOnce()` lists the on-disk `board-thumbnails/` directory once at startup (`packages/mobile/src/hooks/use-native-climb-render.ts:147-183`), and every successful `renderHoldsOverlay` call sets an entry in the `.then` handler (`use-native-climb-render.ts:752-760`).
- When a later render for the same `cacheKey` finds `renderedOverlays.has(currentCacheKey)` true, it **skips calling the native module entirely** and displays the stored URI directly (`use-native-climb-render.ts:715-724`).
- On the native side, `CachePruner` (Android) / `pruneCacheIfNeeded` (iOS) evict the oldest PNGs once the on-disk cache exceeds its 200MB cap, and the OS can independently reclaim individual cache files under storage pressure at any time.
- Nothing invalidates `renderedOverlays` when a file it points at is evicted or reclaimed. The JS map and the on-disk cache can silently drift out of sync — `renderedOverlays` says "here's a valid file," the file is gone, and the `<Image>` component just fails to decode it.

## Why this is a separate issue from #4107

#4107 (and its fix) is about `BoardRenderer.renderHoldsOverlay` *throwing* — a promise rejection surfaced to Sentry as `FileNotFoundException`/ENOENT. This bug is the opposite failure mode: nothing throws, nothing rejects, nothing reports. The native call is never even made, because the JS-side cache believes it already has a valid answer. The user just sees a climb with no hold markers, indistinguishable from any other missing-overlay case, and there's no error to grep for in Sentry — that's the whole reason it isn't already tracked.

## Suggested fix direction

Treat a `renderedOverlays` hit as provisional rather than authoritative: either verify the file still exists before trusting the sync map entry (defeats some of the point of the sync fast path, needs a cheap check), or have `CachePruner`/`pruneCacheIfNeeded` — or the JS side, if it can — invalidate the corresponding `renderedOverlays` entry when a file is evicted, so a stale URI never gets served in the first place.

## Severity

[...]

Issue #4101, opened 2026-08-01
Android release builds upload no Sentry source maps, so every Android JS error merges into one unsymbolicated issue


marcodejongh:
## Summary

Android release builds ship to production with **no Sentry source maps**, so every Android JS error arrives as a single minified frame:

```
app:///index.android.bundle:1:2076761 (CodedError)
```

Sentry groups on the stack trace, so **all** Android `CodedError`s collapse into one issue. [BOARDSESH-6V](https://boardsesh.sentry.io/issues/BOARDSESH-6V) is the result: it reads as "1,350 events / 52 users, one bug" but actually contains at least four unrelated defects.

For contrast, iOS events in the same project symbolicate correctly (`packages/shared/graphql-client/src/execute.ts`, `packages/mobile/src/hooks/use-native-oauth-sign-in`), because the iOS build-phase upload runs.

## Root cause

`.github/workflows/android-apk-rn.yml` disables the upload for both the APK and the AAB:

- Line ~271, "Disable Sentry source map upload": `echo "SENTRY_DISABLE_AUTO_UPLOAD=true" >> "$GITHUB_ENV"`
- Line ~466, `Build signed release AAB`: `SENTRY_DISABLE_AUTO_UPLOAD: 'true'`

Both carry the same stated reason:

> `@sentry/react-native`'s Gradle upload task currently calls an API that is incompatible with the Gradle version generated by Expo prebuild (`Could not find method exec()` on the upload task).

The workaround was deliberate and is even `::warning::`-annotated in CI, but it has never been revisited, and the cost is now much larger than "stack traces are ugly".

## What is actually hiding inside BOARDSESH-6V

Split by message over the last 14 days:

| Sub-bug | Events | Users |
|---|---|---|
| `BoardRenderer.renderHoldsOverlay` → `FileNotFoundException` on `cache/board-thumbnails/*.png` | 1230 | 9 |
| `ModalBottomSheetView.hide` → "No handler registered for AsyncFunction 'hide'" | 38 | 7 |
| `NativeDatabase.prepareAsync` / `finalizeAsync` / `execAsync` → "database is locked" | 52 | 32 |
| `GoogleMapsView.setCameraPosition` → "CameraUpdateFactory is not initialized" | 16 | 1 |

Four different subsystems — the native renderer, the @expo/ui sheet, offline SQLite, and the maps view — in one bucket, ranked by a headline number that belongs to none of them.

## Impact

[...]

PR #4114, opened 2026-08-01
fix(mobile): survive a second cache-file vanish rendering overlays

### LINKED ISSUE #4107: BoardRenderer.renderHoldsOverlay still throws ENOENT on a vanished thumbnail-cache file (recurrence of #3182, 1230 events)
## Summary

Recurrence of #3182. `BoardRenderer.renderHoldsOverlay` still throws when its cached PNG has vanished, despite the atomic-write + cache-dir-recovery hardening that closed that issue.

```
Call to function 'BoardRenderer.renderHoldsOverlay' has been rejected.
→ Caused by: java.io.FileNotFoundException:
  /data/user/0/com.boardsesh.app/cache/board-thumbnails/v3_s_w996_kilter_1_28_1,20_ace41b7e.png:
  open failed: ENOENT (No such file or directory)
```

- **Sentry:** inside [BOARDSESH-6V](https://boardsesh.sentry.io/issues/BOARDSESH-6V) (over-merged Android bucket — see #4101)
- **Impact:** **1230 events / 9 users over 14d** — the single largest sub-bucket in that issue
- **Platform:** Android. Last seen 2026-07-29.
- Spread across many distinct cache keys (`v3_s_w996_kilter_1_28_1,20_*`, `v4_s_w856_moonboard_3_1_*`, `v3_f_w400_*`), so it isn't one poisoned file.

The event-to-user ratio (~137:1) is the tell: an affected device fails repeatedly, presumably re-rendering the same list over and over.

## Why the existing hardening isn't catching it

`packages/mobile/modules/board-renderer/android/src/main/java/com/boardsesh/boardrenderer/BoardRendererModule.kt` already handles the case where Android reclaims `cacheDir` mid-session — `CacheDirRecovery.retryOnceAfterRecreating` re-creates the directory and retries the whole write pipeline once, and `AtomicFileWrite` prevents partial writes.

But the exception still escapes, which means one of:

- the retry runs and fails again (the directory is being reclaimed faster than one retry can win, e.g. sustained storage pressure), or
- the ENOENT is raised somewhere outside the `retryOnceAfterRecreating` boundary — worth checking the temp-file creation and the `Os.rename` publication specifically, and
- the prune path: `pruneCacheIfNeeded` runs once per process under `synchronized`, but `outputFile.exists()` → `setLastModified` → return is **not** synchronized against a concurrent prune. A file that passes `exists()` can be evicted before the caller reads the returned `file://` URI.

That last one fits the data well: heavy scrolling on a device near the 200 MB cache cap, many concurrent renders, prune evicting entries that another render just validated.

## Fix direction

1. Don't let a missing cache file be fatal. `renderHoldsOverlay` should fall back to re-rendering rather than throwing — a cache miss is a normal condition, not an error.
2. Close the check-then-use race: either hold the lock across `exists()` + read, or have the JS side treat a returned-but-missing file as a miss and re-request.
3. Once it can no longer throw, stop reporting it to Sentry at all — this is recoverable, and at 1230 events it is drowning real Android signal.

## Related

- #3182 — original fix (closed)
- #3647 — offline image/thumbnail caches are never swept
- #4101 — Android source maps, which is why this is hiding inside BOARDSESH-6V

marcodejongh:
> [!IMPORTANT]
> **[native-train]** — this PR touches native Kotlin (`modules/board-renderer/android/`) and Swift (`modules/board-renderer/ios/`) and moves the Expo native fingerprint. Per repo policy, `main` must stay OTA-compatible with store binaries, so this PR **stays a draft** and rides the next native release train — do not mark it ready for review or merge it on its own schedule.

## Summary

Closes #4107.

`BoardRenderer.renderHoldsOverlay` still throws `FileNotFoundException: ... open failed: ENOENT` on rare occasions after #4041's cache-directory recovery and #3748/#4073/#4081's atomic write — 1230 events / 9 users over 14 days in Sentry's `BOARDSESH-6V` bucket.

### Correcting the issue's TOCTOU framing before it outlives the fix

#4107 frames the residual bug as: `outputFile.exists()` → `setLastModified()` → return can race a concurrent evict/reclaim, i.e. a check-then-use gap on the **read** side. Investigation for this PR found that framing doesn't hold up against the current code on either platform:

- Android's fast path (`outputFile.exists()` / `.length()` / `.setLastModified()`) and iOS's (`attributesOfItem(atPath:)` / `setAttributes`) are all filesystem **metadata** ops. None of them call `open()`, so none of them can throw `FileNotFoundException: ... open failed: ENOENT` — that message format is specific to opening a stream. The actual file *read* happens downstream, in the JS `<Image>` that consumes the returned `file://` URI, outside this promise entirely.
- The real throw site is the **write** pipeline, and the real gap is that its retry is bounded to exactly one attempt: `CacheDirRecovery.retryOnceAfterRecreating` (Android) / `retryOnceAfterRecreatingCacheDir` (iOS) each recreate the cache dir and retry the write **once**. Under sustained storage pressure, the just-recreated directory (or a fresh temp file inside it) can vanish again before the retried write lands, and that second failure was unguarded — it propagated straight out as the rejected `renderHoldsOverlay` promise. This matches the issue's own hypothesis #1 ("the retry runs and fails again... directory reclaimed faster than one retry can win").

Also worth stating plainly for the record: the Sentry window (last-seen 2026-07-29) entirely **predates** #4073/#4081 landing on main (2026-07-31/08-01). Those PRs replaced the old direct `FileOutputStream(outputFile)`-on-the-final-path write (the exact crash #3748 originally documented) with a temp-file-then-rename. A meaningful share of the 1230 events almost certainly reflects binaries shipped before that fix — and `BOARDSESH-6V` is flagged as an over-merged bucket (#4101) — so the residual volume this PR is actually closing is smaller than the raw count suggests. Re-check the bucket after this ships in a native release before drawing conclusions from it.

### The fix

One more bounded retry, wrapping the **entire** cache-check-render-write pipeline (not just the write step `CacheDirRecovery` already guards):

- **Android** — new `CacheMissRecovery.retryOnceOnFileNotFound { … }` (pure Kotlin, no Android framework calls, no `File` coupling). `BoardRendererModule.renderOverlay` now wraps the renamed `renderOverlayOnce` in this retry: on `FileNotFoundException`, log and re-run the whole pipeline exactly once from scratch (fresh fast-path check, fresh render, fresh write — including a fresh inner `CacheDirRecovery` pass). Deliberately narrower than `CacheDirRecovery`'s `IOException` contract: only a vanished-file failure is recoverable by re-rendering; a broader `IOException` (e.g. genuine out-of-space) isn't, and retrying the whole pipeline for it would double the wasted work without fixing anything.
- **iOS** — mirrors the same shape: `isFileVanishedError(_:)` checks for `NSCocoaErrorDomain`/`NSFileNoSuchFileError` (what `Data.write(to:options:.atomic)` throws for a missing directory) or `NSPOSIXErrorDomain`/`ENOENT` (in case the underlying call surfaces a lower-level POSIX error instead), and `renderOverlay` retries `renderOverlayOnce` once on a match. **Flagging the uncertainty explicitly**: the exact domain/code Foundation raises here is unverified against a real device or simulator — no Mac was available while writing this — both are checked defensively as a reasonable superset. The three `"BoardRenderer"`-domain `NSError`s (Rust render failure, failed `CGImage` wrap, failed PNG encoding) are a different domain entirely and are never matched, so a genuine render/encoder bug is never masked by a pointless retry.

### Follow-up filed (not in scope here)

Investigation surfaced a related but distinct issue: `use-native-climb-render.ts`'s `renderedOverlays` map can hand `<Image>` a `file://` URI for an entry that `CachePruner`'s 200MB LRU cap evicts moments later — that fails silently (blank overlay, no crash, no Sentry event), so it's a different symptom than this issue's thrown exception. Filed separately: #4113.

## Release Notes

- Board hold overlays recover instead of erroring when the phone reclaims cache storage mid-render.

## Test plan

**Green `vp` checks below do not cover this fix.** This is pure native Kotlin/Swift; nothing in `vp check` / `vp run typecheck:mobile` / `vp run test:mobile` / `vp run check:mobile-bundle` compiles or exercises either platform's code — confirmed both pass unchanged (no JS/TS touched).

- **Android**: new `CacheMissRecoveryTest.kt` (plain JUnit, no Robolectric, matches `CacheDirRecoveryTest`/`AtomicFileWriteTest` style) — retried exactly once on `FileNotFoundException`, and on a plain `IOException` caught while the cache dir is missing (how a second vanish actually surfaces from `File.createTempFile` / the wrapped `Os.rename` ENOENT); a second vanish-shaped failure propagates instead of looping; a dir-present `IOException` (out-of-space, permissions) and non-IO failures are *not* retried; happy path passes through untouched. Runs in CI via `android-pr-rn.yml`'s existing `board-renderer:testDebugUnitTest` job — no workflow change needed, it already targets this module.
- **iOS**: new `BoardRendererErrorClassificationTests.swift` (XCTest, compiled into the `BoardseshTests` target alongside the extracted `BoardRendererErrorClassification.swift` via `scripts/prepare-rn-ios-tests.mjs`; runs on a macOS simulator in `ios-rn-ci.yml`) — the headline test is *functional*: it performs the real failing operation (atomic `Data.write` into a directory that existed and was then deleted, the exact reclaim sequence) and asserts `isFileVanishedError` classifies it retryable, so the classifier is verified against what Foundation actually throws. A real not-a-directory failure plus hand-built render/out-of-space/permission errors classify non-retryable, and a Cocoa write wrapper carrying POSIX `ENOENT` under `NSUnderlyingErrorKey` is unwrapped (bounded) and classified retryable.
- `vp run typecheck:mobile` — passed (234s, no errors).
- `vp check` — passed for all touched files; the 2 pre-existing lint errors in the full-repo run are in unrelated `packages/db/scripts/*.test.ts` files, untouched by this PR.

### DIFF
--- a/packages/mobile/ios-tests/BoardRendererErrorClassificationTests.swift
+++ b/packages/mobile/ios-tests/BoardRendererErrorClassificationTests.swift
@@ -1,0 +1,131 @@
+import Foundation
+import XCTest
+
+/// Verifies `BoardRendererErrorClassification` — the gate for #4107's bounded
+/// whole-pipeline re-render — against what Foundation *actually* throws, not
+/// just hand-built errors. The headline test performs the real failing
+/// operation the production path hits: `Data.write(to:options:.atomic)` into a
+/// directory that existed and was then deleted, exactly how iOS reclaiming
+/// `Library/Caches` mid-session surfaces in `BoardRendererModule`.
+@available(iOS 17.0, *)
+final class BoardRendererErrorClassificationTests: XCTestCase {
+  /// Creates a real scratch directory under the test runner's temp dir and
+  /// registers its cleanup. Each test gets a unique path so parallel or
+  /// re-run executions can't collide.
+  private func makeScratchDirectory() throws -> URL {
+    let scratchDir = FileManager.default.temporaryDirectory
+      .appendingPathComponent("board-renderer-classification-\(UUID().uuidString)", isDirectory: true)
+    try FileManager.default.createDirectory(at: scratchDir, withIntermediateDirectories: true)
+    addTeardownBlock {
+      try? FileManager.default.removeItem(at: scratchDir)
+    }
+    return scratchDir
+  }
+
+  /// Nests `error` under `layers` unrelated Cocoa wrappers via
+  /// `NSUnderlyingErrorKey`, mimicking how Foundation chains causes.
+  private func wrapped(_ error: NSError, inLayers layers: Int) -> NSError {
+    var current = error
+    for _ in 0..<layers {
+      current = NSError(
+        domain: NSCocoaErrorDomain,
+        code: NSFileWriteUnknownError,
+        userInfo: [NSUnderlyingErrorKey: current]
+      )
+    }
+    return current
+  }
+
+  // MARK: - Functional: the real production failure shape
+
+  func testAtomicWriteIntoDeletedDirectoryClassifiesAsVanished() throws {
+    let scratchDir = try makeScratchDirectory()
+    let doomedDir = scratchDir.appendingPathComponent("board-thumbnails", isDirectory: true)
+    try FileManager.default.createDirectory(at: doomedDir, withIntermediateDirectories: true)
+    // The reclaim: the directory existed when the "render" started and is gone
+    // by the time the write lands — the exact #4107 sequence.
+    try FileManager.default.removeItem(at: doomedDir)
+
+    let outputUrl = doomedDir.appendingPathComponent("climb.png")
+    do {
+      try Data([0x01]).write(to: outputUrl, options: .atomic)
+      XCTFail("Writing into a deleted directory should throw")
+    } catch {
+      XCTAssertTrue(
+        BoardRendererErrorClassification.isFileVanishedError(error),
+        "A real atomic write into a vanished directory must classify as retryable, got: \(error)"
+      )
+    }
+  }
+
+  func testAtomicWriteThroughRegularFileClassifiesAsNotVanished() throws {
+    // A genuinely different real failure: the parent path exists but is a
+    // regular file, not a directory (POSIX ENOTDIR, not ENOENT). Nothing
+    // vanished, so a re-render can't fix it and it must not retry.
+    let scratchDir = try makeScratchDirectory()
+    let blockingFile = scratchDir.appendingPathComponent("blocker")
+    try Data([0x02]).write(to: blockingFile)
+
+    let outputUrl = blockingFile.appendingPathComponent("climb.png")
+    do {
+      try Data([0x01]).write(to: outputUrl, options: .atomic)
+      XCTFail("Writing through a regular file should throw")
+    } catch {
+      XCTAssertFalse(
+        BoardRendererErrorClassification.isFileVanishedError(error),
+        "A not-a-directory failure is not a vanished cache entry, got: \(error)"
+      )
+    }
+  }
+
+  // MARK: - Classification table
+
+  func testTopLevelNoSuchFileShapesClassifyTrue() {
+    let cocoaNoSuchFile = NSError(domain: NSCocoaErrorDomain, code: NSFileNoSuchFileError)
+    let cocoaReadNoSuchFile = NSError(domain: NSCocoaErrorDomain, code: NSFileReadNoSuchFileError)
+    let posixNoSuchFile = NSError(domain: NSPOSIXErrorDomain, code: Int(ENOENT))
+
+    XCTAssertTrue(BoardRendererErrorClassification.isFileVanishedError(cocoaNoSuchFile))
+    XCTAssertTrue(BoardRendererErrorClassification.isFileVanishedError(cocoaReadNoSuchFile))
+    XCTAssertTrue(BoardRendererErrorClassification.isFileVanishedError(posixNoSuchFile))
+  }
+
+  func testUnwrapsPosixEnoentBuriedUnderCocoaWrappers() {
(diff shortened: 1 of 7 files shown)


marcodejongh:
## Fable review sign-off — verified fixed, blocked on #4170 ⏳

Adversarial Fable review found two majors (both effectiveness gaps, not regressions); fixes implemented and independently **verified CONFIRMED_FIXED** by the original reviewer:

- Android retry now gates on `IOException` + cache-dir-missing-at-catch-time instead of `FileNotFoundException` only (the FNF-only guard missed the dominant plain-IOException vanish shapes in the post-atomic-write pipeline). Dir-present failures (disk full, permissions) still don't retry. Mutation-tested twice on real kotlinc runs — never-retry, bound-of-2, gate-dropped, and FNF-special-case-dropped mutants all killed.
- iOS classifier extracted to a seam (`BoardRendererErrorClassification.swift`), now unwraps `NSUnderlyingErrorKey` chains for buried POSIX ENOENT, and gained a **functional** XCTest (real temp dir deleted, atomic write attempted, classification asserted) — replacing the PR's original "no automated coverage possible" claim.
- Post-merge compatibility with main's `prepare-rn-ios-tests` drift test executed (not just string-matched): passes.

**Current red CI is inherited, not this PR's**: the branch was rebased onto main and picked up main's broken `BoardBleWriteFlowTests.swift` class layout. Sequence: merge #4170 → I rebase this branch → macOS jobs adjudicate the two remaining environment questions (Darwin's ENOTDIR mapping in the functional negative; gradle robolectric run) → un-draft.

PR #4115, opened 2026-08-01
fix(backend): keep Aurora duplicate mutations together

### LINKED ISSUE #4059: Deleting an Aurora-duplicated tick just reveals the next copy
Follow-up to #3535.

#3535 collapses Aurora's own duplicate ascents at read time: the group's lowest `aurora_id` is shown and its twins are hidden. `deleteTick` (`packages/backend/src/graphql/resolvers/ticks/mutations.ts`) hard-deletes one row by uuid and knows nothing about the group, so deleting the visible send promotes the next twin and the entry stays on screen — it reads as "delete did nothing". Repeating the delete once per copy eventually clears it (2-4 times).

Prod scope is small: 11 groups / 25 rows fleet-wide when #3535 was measured.

Fix: make the delete group-aware — when deleting a real Aurora-pull tick, delete its whole natural-key twin group (same user/board/climb/angle/climbed_at, all real `aurora_pull` rows). Recorded behaviour is pinned by the test `reveals the next twin when the visible tick is deleted (recorded behaviour)` in `packages/backend/src/__tests__/aurora-twin-dedup.test.ts`.

### LINKED ISSUE #4060: Editing an Aurora-duplicated tick's angle or date brings its hidden copies back
Follow-up to #3535.

#3535's read-time collapse keys on user/board/climb/angle/climbed_at. `UpdateTickInput` also accepts `angle` and `climbedAt` (`packages/shared-schema/src/schema/ticks.ts`), and either moves the visible survivor out of its natural-key group, so every copy it was hiding reappears — the same "my edit created duplicates" confusion the locally-edited payload relaxation exists to prevent, reached through a different field.

Prod scope is small: 11 groups / 25 rows fleet-wide when #3535 was measured, so this needs an edit on one of those rows to bite.

Options: re-key the group on the pre-edit natural key, or make the edit fan out across the twin group. Current behaviour is pinned by the test `resurfaces the twins when the survivor is moved off the natural key (recorded behaviour)` in `packages/backend/src/__tests__/aurora-twin-dedup.test.ts`.

marcodejongh:
## Summary

- make edits and deletes apply to the visible Aurora ascent and its directly hidden duplicate rows
- keep distinct Kilter-linked ascents separate, including the null-ID target orientation
- preserve PostgreSQL microseconds on date edits and fan out dependent cleanup, stats, beta links, tombstones, and cache invalidation
- serialize local mutations with Aurora and Kilter imports through the same per-user transaction lock

Closes #4059.
Closes #4060.

## Release Notes

Editing or deleting an ascent imported more than once now sticks instead of revealing hidden copies.

## Test plan

- `vp test run --project backend packages/backend/src/__tests__/aurora-twin-dedup.test.ts packages/backend/src/__tests__/tick-validation.test.ts --reporter=agent` — 67 passed against real PostgreSQL
- `vp test run --project backend packages/backend/src/__tests__/sync-mutations.test.ts packages/backend/src/__tests__/tick-queries.test.ts --reporter=agent` — 79 passed
- `vp test run --reporter=agent packages/aurora-sync packages/kilter-sync` — 418 passed, 1 skipped
- `vp run --filter=@boardsesh/db test` — advisory-lock SQL-render test passes (3 unrelated `createClimbFilters` MoonBoard failures reproduce on `origin/main` untouched)
- `vp run typecheck:backend`, `vp run typecheck:db`
- `vp check`
- independent adversarial review: PASS
- attempted `vp run dev` with `.boardsesh/qa-notes.md`; startup is currently blocked before the app launches because the shared dev DB has migration 0187's schema without its ledger row (#3978), so the launcher retries a non-idempotent table create

### DIFF
--- a/packages/aurora-sync/src/sync/apply-user-logbook.test.ts
+++ b/packages/aurora-sync/src/sync/apply-user-logbook.test.ts
@@ -6,7 +6,7 @@ import type { SQL } from 'drizzle-orm';
 // cross-source claim exercises the actual matching logic.
 vi.mock('@boardsesh/db/queries', async (importActual) => {
   const actual = await importActual<typeof import('@boardsesh/db/queries')>();
-  return { ...actual, recomputeClimbStatsBulk: vi.fn() };
+  return { ...actual, acquireUserTickMutationLock: vi.fn(), recomputeClimbStatsBulk: vi.fn() };
 });
 
 import { recomputeClimbStatsBulk } from '@boardsesh/db/queries';
--- a/packages/aurora-sync/src/sync/apply-user-logbook.ts
+++ b/packages/aurora-sync/src/sync/apply-user-logbook.ts
@@ -9,6 +9,7 @@ import {
   adoptionMatchScoreSeconds,
   MAX_USER_UTC_OFFSET_SECONDS,
   NATURAL_KEY_TOLERANCE_SECONDS,
+  acquireUserTickMutationLock,
   type ClimbStatsKey,
   type TickTimeSample,
 } from '@boardsesh/db/queries';
@@ -749,6 +750,7 @@ async function applyLogbookChunk(
       -- but keeps the never-cross-users invariant enforced at the write itself.
       WHERE t.uuid = u.uuid
         AND t.user_id = ${userId}
+        AND (t.aurora_synced_at IS NULL OR t.updated_at <= t.aurora_synced_at)
     `);
   };
 
@@ -809,6 +811,11 @@ async function applyLogbookChunk(
       -- the write itself.
       WHERE t.aurora_id = u.aurora_id
         AND t.user_id = ${userId}
+        -- Re-check at write time. A rolling-deploy node that does not yet take
+        -- the shared advisory lock can edit after the SELECT above; comparing
+        -- timestamp columns here also preserves Postgres microseconds that JS
+        -- Date.parse cannot represent.
+        AND (t.aurora_synced_at IS NULL OR t.updated_at <= t.aurora_synced_at)
     `);
   };
 
@@ -962,6 +969,10 @@ export async function applyAuroraAscents(
     }
   }
 
+  // Shared with updateTick/deleteTick and Kilter log apply. The caller owns
+  // the transaction; take this before the first target/tombstone SELECT.
+  await acquireUserTickMutationLock(db, userId);
+
   touchedKeys.push(...(await applyAuroraTombstones(db, boardName, userId, tombstoneIds)));
 
   const unique = dedupeByAuroraId(live);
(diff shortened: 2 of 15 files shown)


marcodejongh:
Fable review follow-up — all three findings addressed, plus a rebase onto `main` (was 182 behind; the branch had drifted past changes to `kilter-sync/user-sync.ts` and `ticks/queries.ts`).

**Critical — PR title scope `logbook` failed commit-lint.** Retitled to `fix(backend): keep Aurora duplicate mutations together`. `backend` is in `ALLOWED_SCOPES`; the change centres on `packages/backend` tick mutations plus the two sync writers.

**Major — the dual advisory-lock keys justified themselves with a "legacy node" premise that was never true.** `acquireUserTickMutationLock` took `pg_advisory_xact_lock(0x5449434b, hashtext(userId))` first, documented as "the legacy two-int key used by already-deployed nodes", and told future maintainers to keep it "until every node capable of running the old helper has drained". Nothing before this branch takes a per-user tick advisory lock at all — the "legacy" key only ever existed in this PR's own commit history, so the second key was guarding against my own earlier commits. Collapsed to the single 64-bit `hashtextextended` key and rewrote the doc comment to state what actually covers the rollout window: old nodes take *no* lock, so the pull writers' write-time `updated_at <= *_synced_at` guards are what make it safe (those comments stay). The SQL-render test now asserts one statement, and the cross-connection contention test is repointed at what still needs proving — two callers of the helper serialize, and a second climber's mutation does **not** queue behind the first (that second half is new; it would catch a seed/hash that stopped depending on the user id).

**Minor — the timestamp-fraction regex was duplicated and missed hour-only offsets.** `validation/schemas/ticks.ts` now exports one `POSTGRES_TIMESTAMP_FRACTION_PATTERN` (plus a `readTimestampFractionalSeconds` helper) that `normalizeClimbedAt` imports, so the two cannot drift. The offset alternative is now `[+-]\d{2}(?::?\d{2})?` with optional preceding whitespace.

One correction to the finding's example: `2026-05-02T01:02:03.123+07` is rejected outright — V8 will not parse an hour-only offset in the `T`-separated ISO form, so the "valid date" refine catches it. The bug is real in the **space**-separated form V8's legacy parser does accept:

- `2026-05-02 01:02:03.123456+07` → fraction went unread, stored as `.000`
- `2026-05-02 01:02:03.1234567+07` → seven digits walked straight past the six-digit refine

Both are covered now by table-driven cases in `tick-validation.test.ts` that assert the string parses *and* that the fraction is read back. `normalizeClimbedAt` also slices to six digits as belt-and-braces against a caller that skips validation.

Validation: `vp check`; `vp run typecheck:backend` + `typecheck:db`; 67 passed on `aurora-twin-dedup` + `tick-validation`, 79 on `sync-mutations` + `tick-queries`, 418 across `aurora-sync` + `kilter-sync`. `vp run --filter=@boardsesh/db test` has 3 failures in `createClimbFilters` (MoonBoard zone modes) that reproduce identically with my changes stashed — pre-existing, unrelated to ticks.

Issue #4112, opened 2026-08-01
OTA-published JS updates carry no Sentry source maps on either platform


marcodejongh:
## Summary

OTA-published JS bundles (both iOS and Android, via `expo-updates`) carry **no Sentry source maps at all**, distinct from and broader than #4101 (which is scoped to the Android *native release build*).

`.github/workflows/mobile-ota-production.yml` has zero references to `SENTRY_AUTH_TOKEN`, `sentry-cli`, or `sourcemap` anywhere in the file — confirmed by reading it in full while investigating #4101. Every `vp run mobile:publish` / `eoas publish` production OTA ships a Metro-built JS bundle that is never uploaded to Sentry, on either platform.

## Why this is separate from #4101

- #4101's fix (in progress) only covers the JS bundle embedded in the Android **native** APK/AAB at build time (`.github/workflows/android-apk-rn.yml`), by re-enabling `@sentry/react-native`'s Gradle upload task, gated on `SENTRY_AUTH_TOKEN` presence — mirroring how iOS's Xcode archive build-phase already works (`ios-testflight-rn.yml`).
- That mechanism only runs at **native build time**. It never runs again when a JS-only fix ships via OTA — which is the normal, high-frequency delivery path for JS changes per `docs/mobile-ota-updates.md` (`runtimeVersion` uses the `fingerprint` policy specifically so JS-only changes can ship without a new binary).
- So even after #4101 ships, any crash happening inside OTA-delivered JS (not the exact bundle baked into the currently-installed binary) will still symbolicate to the wrong source map or nothing at all — the uploaded artifact from the native build no longer matches the bundle actually running.

## Why this needs its own design, not a copy-paste of #4101's fix

Sentry's modern debug-ID-based source map resolution (`@sentry/react-native`'s `copy-debugid.js` mechanism, used by `sentry.gradle`) decouples symbolication from `release`/`dist` matching — each distinct bundle gets its own embedded debug ID, and Sentry matches purely on that. This means an OTA publish step *could* upload its own debug-id-tagged source map independently of any native `release`/`dist` naming.

The complication: a single OTA channel/fingerprint can be pulled by many different installed binaries across many different native `versionCode`s (Android) / build numbers (iOS) — all sharing one `runtimeVersion` fingerprint. The OTA publish step doesn't build against one specific native release the way the native build workflows do, so:

[...]

PR #4117, opened 2026-08-01
fix(mobile): upload Sentry source maps for Android release builds

### LINKED ISSUE #4101: Android release builds upload no Sentry source maps, so every Android JS error merges into one unsymbolicated issue

marcodejongh:
## Summary

Android release builds shipped with **no Sentry source maps**, so every Android JS error arrived as one unsymbolicated frame (`app:///index.android.bundle:1:2076761 (CodedError)`). Sentry groups on the stack trace, so distinct bugs merged into a single issue ([BOARDSESH-6V](https://boardsesh.sentry.io/issues/BOARDSESH-6V) — 1,350 events / 52 users, actually at least four unrelated defects).

Root cause: `.github/workflows/android-apk-rn.yml` hardcoded `SENTRY_DISABLE_AUTO_UPLOAD=true` unconditionally for both the APK and AAB Gradle builds, citing a Gradle `exec()` incompatibility. That incompatibility was fixed upstream in `@sentry/react-native@7.3.0` — we're pinned to `7.11.0`, past the fix — so the workaround comment was stale.

## What this PR does (and doesn't do)

I initially implemented the "obvious" fix: mirror `ios-testflight-rn.yml` exactly and gate the Gradle-embedded upload task on `SENTRY_AUTH_TOKEN` presence, trusting that the upstream fix landed. Per the two-phase plan, I tried to verify this empirically via `workflow_dispatch` on this branch **before** shipping it, since if the Gradle task is *still* broken, it fails `assembleRelease`/`bundleRelease` outright and takes down the entire Android release (APK + AAB + Play upload) — not just symbolication.

That dispatch failed immediately, but not from Sentry/Gradle — from a GitHub environment protection rule:

> Run: https://github.com/boardsesh/boardsesh/actions/runs/30679349491
> `build-and-release`: "Branch 'fix/4101-android-sentry-sourcemaps' is not allowed to deploy to Production due to environment protection rules."

Confirmed read-only via `gh api repos/boardsesh/boardsesh/environments/Production/deployment-branch-policies` — the `Production` environment's custom deployment-branch policy allows only `main`. So a feature-branch dispatch can never exercise the real signed Gradle build before merge, and the first real test of the re-enabled Gradle task would be a push to `main` — an unacceptable blast radius for a P1 fix (if wrong, it blocks shipping *any* Android build, native or otherwise, until reverted).

**So this PR ships the safer plan-B instead:** the Gradle-embedded `sentry.gradle` upload task stays disabled (`SENTRY_DISABLE_AUTO_UPLOAD: 'true'`, unchanged from today), and a new **decoupled, `continue-on-error` step** (`Upload Sentry source maps`) runs after the APK is already built and signature-verified. It replicates `sentry.gradle`'s own upload sequence by hand against the same Gradle-generated bundle/sourcemap files, so a Sentry-side failure can never block shipping the release — worst case is today's status quo (no source maps), never a broken release.

What it does, step by step:
1. Copies the Metro-embedded debug ID from the intermediate ("packager") source map into the final Hermes-composed one, via `@sentry/react-native`'s own `scripts/copy-debugid.js` (Hermes's `compose-source-maps` step doesn't propagate it on its own — skipping this silently breaks symbolication even though the upload itself "succeeds").
2. Uploads via `sentry-cli react-native gradle --bundle ... --sourcemap ... --release "com.boardsesh.app@<versionName>+<versionCode>" --dist <versionCode>` — the exact release/dist naming convention `sentry.gradle` itself uses, which matches what `@sentry/react-native`'s native SDK auto-detects at runtime (`packages/mobile/src/lib/sentry.ts:99-104` leaves `release`/`dist` unset for exactly this reason — the CFBundle/versionName+versionCode auto-detection has to match the uploaded artifact tags or events never resolve to these source maps).
3. Runs for both the APK and AAB builds' shared JS bundle (idempotent — same release/dist either way).

Gated on `SENTRY_AUTH_TOKEN` presence (mirrors the existing Play-upload gating pattern already used twice in this file) so a missing token degrades gracefully instead of erroring.

### Verification once merged

Since `main` is the only branch this environment allows, real verification only happens after merge:
1. Watch the first post-merge `android-apk-rn.yml` run on `main` — confirm the new "Upload Sentry source maps" step succeeds (or check its warning if it doesn't; either way the release itself is unaffected).
2. Throw a deliberate test error on that release build; confirm the resulting Sentry event resolves real `packages/mobile/src/...` frames instead of the minified `app:///index.android.bundle` frame.
3. Once symbolication is confirmed working end-to-end, split [BOARDSESH-6V](https://boardsesh.sentry.io/issues/BOARDSESH-6V) in the Sentry UI — the four sub-bugs it currently merges (`BoardRenderer.renderHoldsOverlay` FileNotFoundException, `ModalBottomSheetView.hide` handler error, `NativeDatabase` "database is locked", `GoogleMapsView.setCameraPosition` init error) already have their own GitHub issues filed and should stop sharing a bucket. This is a Sentry-dashboard action, not code — flagging here as a manual follow-up for whoever has Sentry admin access.

If the Gradle task's `exec()` incompatibility genuinely is fixed in `7.11.0` (plausible, but unverifiable pre-merge — see above), a maintainer with environment-bypass rights could later re-attempt the token-gated Gradle-embedded flip (mirroring iOS) and drop this hand-rolled step; I'm not doing that blind in this PR.

## Scope note: OTA is a separate, bigger gap

This PR only covers the **native release build's** embedded JS bundle. OTA-published JS updates (both iOS and Android, via `expo-updates`) carry no Sentry source maps at all today — `mobile-ota-production.yml` has zero Sentry references. That's a materially different design problem (a single OTA fingerprint can be pulled by many different native `versionCode`s/build numbers, so there's no single native release to tag the upload against) and was tracked separately in **#4112**, which has since been fixed by **#4134** (merged into `main` after this PR branched).

## Test plan

- [x] `vp test run --project scripts --reporter=agent` — 39 test files / 530 tests pass, including a new `scripts/mobile-android-sentry-sourcemaps.test.ts` asserting: the Gradle-embedded task stays hardcoded disabled, the new upload step is `continue-on-error` and token-gated, the debug-ID copy runs before upload, and the release/dist naming matches runtime auto-detection.
- [x] `vp check` — lint + format clean on the changed files.
- [x] `vp run typecheck` — full repo typecheck clean.
- [x] Workflow YAML structurally validated (`python3 -c "import yaml; yaml.safe_load(...)"`).
- [ ] Post-merge (see "Verification once merged" above) — I can't self-verify further pre-merge due to the environment protection rule.

Closes #4101

## Release Notes


marcodejongh:
## Fable review follow-up

Addressed the Fable review feedback:

- **Critical (CI lint red):** `scripts/mobile-android-sentry-sourcemaps.test.ts` was missing `/// <reference types="node" />`, which produced 3x `TS2591` on the `node:fs`/`node:path`/`node:url` imports and reded the `lint` job (and the aggregate `ci-status` check). Added the directive as the first line, matching every other `scripts/*.test.ts` that imports Node builtins (e.g. `scripts/mobile-ci-env-parity.test.ts`). `vp check` and CI's `lint`/`typecheck` jobs are now green.
- **Minor (hardcoded `applicationId`):** Added an inline comment above the `sentry-cli react-native gradle --release ...` call in `.github/workflows/android-apk-rn.yml` noting that `com.boardsesh.app` must track the production `androidPackage` in `packages/mobile/app.config.ts`, since `sentry.gradle` itself derives this from the Gradle variant rather than a hardcoded literal.
- **Minor (stale `#4112` reference):** Updated the PR body's "Scope note" — `#4112` (OTA source-map gap) is closed, fixed by #4134 (merged into `main` after this branch was cut). The body now says so instead of pointing at an open tracking issue.

Rebased onto current `main` (no conflicts) and force-pushed. CI is green: `lint`, `typecheck`, `test-default`, `test-report`, `commit-lint`, `ota-check`, `claude-review`, and the aggregate `ci-status` all pass. Also confirmed locally: `vp check`, `vp run typecheck`, `vp test run --project scripts --reporter=agent` (mobile-android-sentry-sourcemaps suite), and `vp run check:mobile-bundle` all pass on the rebased tree.

Marking ready for review.

PR #4120, opened 2026-08-01
fix(mobile): guard @expo/ui hide() against missing native handler

### LINKED ISSUE #4108: expo-ui ModalBottomSheetView.hide 'no handler registered' — same defect as #3478 on a different method
## Summary

Same defect as #3478, different method. That issue covered `ModalBottomSheetView.partialExpand` ("no handler registered", 108 users) and was closed; the `hide` handler now has the identical problem.

```
Call to function 'ModalBottomSheetView.hide' has been rejected.
→ Caused by: java.lang.IllegalStateException: No handler registered for AsyncFunction 'hide'
  on view 'ModalBottomSheetView'. Did you forget to bind it with `hide.handle { ... }`
  inside the Content { } block?
```

- **Sentry:** inside [BOARDSESH-6V](https://boardsesh.sentry.io/issues/BOARDSESH-6V) (over-merged Android bucket — see #4101)
- **Impact:** 38 events / 7 users over 14d
- **Platform:** Android. Last seen 2026-07-28.

## What's going on

The `@expo/ui` bottom sheet binds its imperative async functions inside the composable `Content { }` block. Calling `hide()` when the sheet's content isn't composed — dismissed already, unmounted, or backgrounded mid-animation — finds no registered handler and rejects.

#3478 hit this on `partialExpand`; the fix there addressed that call site rather than the general shape, so `hide` was left exposed. Any other imperative method on this view has the same hazard.

## Fix direction

1. Guard the imperative calls: only invoke `hide()` when the sheet is actually presented, and treat "already gone" as success rather than an error.
2. Prefer fixing this once at the wrapper level — a small helper that swallows `no handler registered` for every `ModalBottomSheetView` async function — rather than per method, so the next one doesn't recur the same way.
3. Note this interacts with the sheet-stacking rule already established in this codebase: stacked sheets present via an imperative ref handle, and a visible-prop `useEffect` present/dismiss is a silent no-op. Dismissal should go through the same ref handle.

## Related

- #3478 — same error on `partialExpand` (closed)
- #4076 — remix path presents create under a live native sheet (sheet-coordinator bypass)
- #4101 — Android source maps, which is why this is hiding inside BOARDSESH-6V

marcodejongh:
## Summary

Same defect class as #3478, one method over: `@expo/ui`'s `ModalBottomSheetView` on Android exposes `hide()` as a native `AsyncFunction`, but its handler exists only while the Compose `Content { }` block is registered. Calling `hide()` after the sheet is already dismissed or unmounted, or while it is backgrounded mid-animation, rejects with `No handler registered for AsyncFunction 'hide' on view 'ModalBottomSheetView'`. The same rejection is also possible when JS runs against a native ExpoUI layer that predates the method. #3478 guarded `expand()`/`partialExpand()`; `hide()` — used by `close()`/`forceClose()`/`dismiss()` and the `snapToIndex(-1)` path, i.e. every programmatic sheet dismiss on Android — was left exposed.

It's worse than the expand/partialExpand case was: because the old code was `sheetRef.current?.hide().then(() => { setIsOpen(false); ...; fireCloseCallbacks(); })`, a native rejection skipped `.then()` entirely, and a missing ref short-circuited the promise chain. In either lifecycle race the local cleanup never ran. Beyond the Sentry noise, that's a real bug — the sheet's JS state could get stuck "open" after the user asked to close it.

## Fix

`patches/@expo%2Fui@57.0.8.patch` (Bun patch on `@expo/ui`, edited via `bun patch @expo/ui` → `bun patch --commit`), touching only `src/community/bottom-sheet/BottomSheet.android.tsx`. Both dismiss paths now funnel through one named helper carrying the same `swallowMissingNativeHandler` guard #3478 already added for `expand()`/`partialExpand()`:

```js
const hideSwallowingMissingNativeHandler = () => {
  const hidePromise = sheetRef.current?.hide();
  void Promise.resolve(hidePromise)
    .catch(swallowMissingNativeHandler)
    .then(() => {
      setIsOpen(false);
      pendingIndexRef.current = null;
      fireCloseCallbacks();
    });
};
```

`snapToIndex(-1)` calls it; `close` (and therefore `forceClose`/`dismiss`) is it. `Promise.resolve(undefined)` treats an already-missing ref as a successful dismissal, while `.catch()` turns a native handler rejection into a resolved `undefined`. The cleanup therefore runs for both lifecycle-unavailable paths, so the sheet closes in JS state whether the native side accepts the call, rejects it, or is already gone. On the common path where the ref and handler are present, behaviour is unchanged. The helper's doc comment now covers Compose lifecycle unavailability as well as native-version skew and names `hide()`/`close()`/`dismiss()` alongside `expand()`/`partialExpand()`.

### Sentinel now pins the complete dismiss guard

`scripts/mobile-patches-check.ts` is the only thing standing between a silently-dropped Bun patch and a runtime-only regression. Its `@expo/ui` Android rule used to assert just one symbol: `swallowMissingNativeHandler`. The **pre-#4108 patch satisfies that too** — regenerate the patch for a version bump, reinstate only the #3478 half, and the check stays green while the `hide()` guard disappears. Exactly the partial-coverage failure this PR is fixing.

The shipped rule now pins the #3478 `expand()` and `partialExpand()` catches separately, then pins the `hide()` helper's normalized promise and `.catch(swallowMissingNativeHandler)` as one contiguous sentinel. It also pins both consumers independently: the `snapToIndex(-1)` invocation and the `close` alias used by `forceClose()`/`dismiss()`. Mutation tests feed the shipped rule a pre-#4108 source, a helper whose null ref short-circuits cleanup, and a helper with only the hide catch removed; all three must go red. The patch's Android result blob id was also refreshed from the verified installed source.

## Native-release train

This source PR now targets `release/next-ota` and is stacked after #4420. The `@expo/ui` package is autolinked and patched package source is included in the native fingerprint, so this change requires the new iOS and Android binaries assembled by aggregate PR #4457.

The source-PR native gate still compares the cumulative head to `main`, preserving the safety signal while the train is assembled. Final platform fingerprints and native builds are owned by #4457 after all four source PRs land.

## Why CI hadn't validated this before

The branch's base also predated the CI change that added `patches/**` to the `mobile` path filter, so `mobile-bundle` (the job that runs `vp run check:mobile-patches`), the mobile test project, and the Metro bundle check were all **skipped** on the original run — only local runs covered the edited patch. The rebase fixes that: those jobs now actually exercise the change, alongside the Android APK + Robolectric build the fingerprint move pulls in.

## Test plan

- `vp run check:mobile-patches` — pass, 6 native patch rules applied and 4 patch files accounted for
- `vp test run scripts/__tests__/mobile-patches-check.test.ts --reporter=agent` — 1 file / 40 tests pass, including the pre-#4108, null-ref, and catch-only mutations
- `vp run check:mobile-bundle` — Metro bundle check, iOS + Android, OK
- `vp check` on the touched files — clean
- Independent QA — PASS with no actionable findings; the final train rebase is patch-identical and the exact rebased head is `39289350d`

No new behavioural unit test for the patch itself: the vendored file can't be meaningfully exercised under Vitest (the native view needs a real Expo module host), which is why the `mobile-patches-check.ts` sentinel stands in for one — and why this PR makes that sentinel actually discriminating.

## Release Notes

Android: sheets now close reliably when the native close handler is temporarily unavailable.

Closes #4108

### DIFF
--- a/patches/@expo%2Fui@57.0.8.patch
+++ b/patches/@expo%2Fui@57.0.8.patch
@@ -1,5 +1,5 @@
 diff --git a/build/community/bottom-sheet/types.d.ts b/build/community/bottom-sheet/types.d.ts
-index 72fc6be..d1cf11e 100644
+index 72fc6becbc414c318799e28103d15484ed86f703..d1cf11e93b0c0d8dc77abddaafc89e14d34faccc 100644
 --- a/build/community/bottom-sheet/types.d.ts
 +++ b/build/community/bottom-sheet/types.d.ts
 @@ -159,6 +159,14 @@ export interface BottomSheetProps {
@@ -18,21 +18,27 @@ index 72fc6be..d1cf11e 100644
       * Whether the sheet can be dismissed by panning down.
       * @default false
 diff --git a/src/community/bottom-sheet/BottomSheet.android.tsx b/src/community/bottom-sheet/BottomSheet.android.tsx
-index f60ae61..4e64d1a 100644
+index f60ae61bf0c9023d41c55cdca2d9af0de7291a92..155afb2055b9e977f0651de67a4f1c812e87eef7 100644
 --- a/src/community/bottom-sheet/BottomSheet.android.tsx
 +++ b/src/community/bottom-sheet/BottomSheet.android.tsx
-@@ -11,6 +11,20 @@ import { RNHostView } from '../../jetpack-compose/RNHostView';
+@@ -11,6 +11,26 @@ import { RNHostView } from '../../jetpack-compose/RNHostView';
  
  export { useBottomSheet } from './context';
  
-+// `expand()` / `partialExpand()` are native AsyncFunctions on `ModalBottomSheetView`.
-+// A store binary whose native ExpoUI layer predates a method rejects the call with
-+// "No handler registered for AsyncFunction '<method>' on view 'ModalBottomSheetView'".
-+// These calls are fire-and-forget, so an unhandled rejection would surface as a global
-+// onunhandledrejection (crash report + noise) even though the re-snap is a harmless
-+// no-op on those binaries. Swallow it: the sheet stays where it is, which is the
-+// existing behaviour, and never wrongly expands a collapse. On binaries whose native
-+// layer has the method this never rejects. (boardsesh patch — see issue #3478.)
++// `expand()` / `partialExpand()` / `hide()` are native AsyncFunctions on
++// `ModalBottomSheetView`. A store binary whose native ExpoUI layer predates a
++// method rejects the call with "No handler registered for AsyncFunction
++// '<method>' on view 'ModalBottomSheetView'". These calls are fire-and-forget,
++// so an unhandled rejection would surface as a global onunhandledrejection
++// (crash report + noise) even though the underlying state change is a harmless
++// no-op on those binaries. Swallow it: for expand()/partialExpand() the sheet
++// stays where it is (the existing behaviour, never wrongly expanding a
++// collapse); for hide() — used by close()/forceClose()/dismiss() and the
++// snapToIndex(-1) path — the caller still runs its own cleanup afterwards
++// (setIsOpen(false) etc.), so the sheet closes in JS state even when the
++// native side silently no-ops, instead of leaving it stuck open. On binaries
++// whose native layer has the method this never rejects. (boardsesh patch —
++// see issue #3478, and issue #4108 for the hide() call sites.)
 +function swallowMissingNativeHandler(error: unknown): void {
 +  // Never rejects on a dev build (current native layer has the method), so a warning
 +  // here means something genuinely unexpected — surface it in dev, stay silent in prod.
@@ -42,7 +48,27 @@ index f60ae61..4e64d1a 100644
  function extractBackgroundColor(
    backgroundStyle: BottomSheetProps['backgroundStyle']
  ): string | undefined {
-@@ -143,9 +157,9 @@ export function BottomSheet(props: BottomSheetProps) {
+@@ -127,11 +147,14 @@ export function BottomSheet(props: BottomSheetProps) {
+   const methods: BottomSheetMethods = useMemo(() => {
+     const snapToIndex = (index: number) => {
+       if (index === -1) {
+-        sheetRef.current?.hide().then(() => {
+-          setIsOpen(false);
+-          pendingIndexRef.current = null;
+-          fireCloseCallbacks();
+-        });
++        sheetRef.current
++          ?.hide()
++          .catch(swallowMissingNativeHandler)
++          .then(() => {
++            setIsOpen(false);
++            pendingIndexRef.current = null;
++            fireCloseCallbacks();
++          });
+         return;
+       }
+       const clampedIndex = clampIndex(index);
+@@ -143,9 +166,9 @@ export function BottomSheet(props: BottomSheetProps) {
          // Native ModalBottomSheet has two states: partial (~50%) and expanded (full).
          // Map index 0 → partialExpand(), last index → expand().
          if (clampedIndex === maxIndex) {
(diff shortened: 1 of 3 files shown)


marcodejongh:
🤖 Cross-reference from the `josh` loop: the fingerprint move this PR measured (and correctly treated as gating) is likely an instance of the class documented in #4122 — Bun store-path suffixes make JS-only changes move the runtimeVersion even though no native code changed. The gating stays correct today (stale binaries won't pull an OTA with a different runtimeVersion, whatever the cause), but once #4122 normalizes the fingerprint inputs, re-measure: if the patch no longer moves it, this fix can ship OTA to exactly the stale-binary users who hit the bug, instead of waiting for a train.

marcodejongh:
## Fable review — addressed

**Major 1 — fingerprint verdict contradicted by CI.** Confirmed: the `mobile-ota-check` job resolved head/base fingerprints under byte-identical env and got identical hashes on both platforms, no `native-fingerprint` label was ever applied, and the `pr-4120` OTA preview published. The local before/after measurement was polluted by the isolated-linker hashing bug (#4122, fixed by #4148). Rebased onto post-#4148 `main` so the gate re-measures on the fixed codebase, and rewrote the "Native-fingerprint verdict" section to say what it actually is: **a JS-only change that rides OTA**, not native-train cargo. Leaving draft once CI is green.

**Major 2 — CI never exercised the change.** Also confirmed: the branch base predated the CI change that added `patches/**` to the `mobile` path filter, so `mobile-bundle` (which runs `vp run check:mobile-patches`), the mobile test project, and the Metro bundle check all skipped. The rebase puts them back in play. Locally: `check:mobile-patches` OK (6 rules now), `typecheck:mobile` clean, `test:mobile` 585 files / 5116 tests green, `check:mobile-bundle` OK on both platforms.

**Minor 3 — sentinel couldn't detect loss of the hide() guard.** Fixed rather than deferred, since it's the same partial-coverage shape this PR exists to close. The two byte-identical inline `hide()` chains are now one named `hideSwallowingMissingNativeHandler` helper (`snapToIndex(-1)` calls it; `close`/`forceClose`/`dismiss` *are* it), and `scripts/mobile-patches-check.ts` pins that name as a second sentinel alongside `swallowMissingNativeHandler`. The two don't subsume each other, so a patch regenerated with only the #3478 half now goes red. Added a mutation test that feeds the **shipped** rule a pre-#4108 source and asserts the failure, and refreshed the rule's doc comment (it still described the patch as guarding only `expand()`/`partialExpand()`).

Behaviour of the guard itself is unchanged — same `.catch()` before `.then()`, so the JS cleanup always runs.

marcodejongh:
**Correction to my previous comment on the fingerprint verdict** — the rebase flipped it.

Post-#4148, `mobile-ota-check` on `43fc7df` resolves iOS `a98901604ea8` vs main `183bcf484292` and Android `24f75b79e38b` vs main `a6f256fdaf25`: **native change on both platforms**, and `native-fingerprint` auto-applied. So the original description's conclusion was right, even though the CI of the day contradicted it — that older "identical / ships OTA" verdict was the false one. At that base `@expo/fingerprint` was still unpatched and hashed Bun isolated-linker package sources as `null` (#4122, fixed by #4148), so it was structurally blind to a change inside `@expo/ui`'s package directory. The gate can see it now, and the body's verdict section matches the measurement.

Practical consequence: this rides the next native release train rather than OTA, same as #3478's guard (which shipped inside #3971/#4069). The fingerprint move also pulls the Android APK + Robolectric jobs into this PR's CI, which is where the extra green checks come from.

Marking ready for review — the label is the train signal here; every other `native-fingerprint` PR in this repo goes through review and merge normally.

marcodejongh:
Independent QA PASS on the final patch content, with no actionable findings. The `@expo/ui` patch normalizes a missing ref through `Promise.resolve`, catches native `hide()` rejection before cleanup, and keeps both `snapToIndex(-1)` and the `close`/`forceClose`/`dismiss` aliases guarded. Shipped sentinels independently pin expand, partialExpand, the contiguous hide catch, and both consumers; mutation tests fail for the pre-#4108 source, null-ref short circuit, and catch-only removal. Final train rebase is patch-identical at `39289350d`; 40/40 focused tests and `check:mobile-patches` (6 rules / 4 files) pass.

PR #4123, opened 2026-08-01
fix: chunk bulkVoteSummaries requests past the 100-ID cap

### LINKED ISSUE #4102: bulkVoteSummaries: unchunked callers blow the 100-ID cap once a feed pages past 100 rows
## Summary

`bulkVoteSummaries` rejects any request with more than 100 entity IDs, but most callers hand it an unbounded list. Once a feed pages past 100 rows the query fails outright and every vote button on the screen loses its count and the viewer's own vote state.

```
Invalid input: Too big: expected array to have <=100 items
path: ["bulkVoteSummaries"]
```

The captured request carried **135** tick UUIDs.

- **Sentry:** [BOARDSESH-7H](https://boardsesh.sentry.io/issues/BOARDSESH-7H)
- **Impact:** 314 events / 30 users, ongoing
- **Live on the current build:** `com.boardsesh.app@2.3.0+2000783`, last seen 2026-08-01

## The cap

`packages/backend/src/validation/schemas/comments.ts:78`

```ts
entityIds: z.array(z.string().min(1).max(200)).max(100),
```

## Why it still fires

`packages/mobile/src/lib/graphql/hooks/use-social.ts` already has chunk-aware hooks — `useChunkedBulkVoteSummaries` (line 246) and `useGroupedBulkVoteSummaries` (line 269), both built on `chunkEntityIds` with `BULK_VOTE_SUMMARY_CHUNK_SIZE = 100`.

But the original **unchunked** `useBulkVoteSummaries` (line 225) was left in place and still passes `entityIds` straight through:

```ts
const response = await getHttpClient().request<GetBulkVoteSummariesQueryResponse>(GET_BULK_VOTE_SUMMARIES, {
  input: { entityType, entityIds },
});
```

and `SessionsTab` — the paginating feed, the one place guaranteed to exceed 100 — calls the unchunked one twice:

`packages/mobile/src/components/you/SessionsTab.tsx:142,147`

```ts
const sessionVoteSummaries = useBulkVoteSummaries('session', sessionEntityIds, ...);
const tickVoteSummaries = useBulkVoteSummaries('tick', tickEntityIds, ...);
```

`tickEntityIds` grows with every page, which matches the 135-ID request in Sentry.

## Web has the same gap

`VoteSummaryProvider` (`packages/web/app/components/social/vote-summary-context.tsx:52`) never chunks. Only one of its seven call sites defends itself:

| Call site | Bounded? |
|---|---|
| `logbook/logbook-view.tsx:61` | yes — `VOTE_SUMMARY_BATCH_LIMIT = 100` slice |
| `logbook/crew-logbook-view.tsx:78` | no |
| `social/global-ascents-feed.tsx:64` | no |
| `social/following-ascents-feed.tsx:81` | no |
| `session/[sessionId]/session-detail-content.tsx:565` | no |
| `session/[sessionId]/session-detail-content.tsx:811` | no |
| `activity-feed/activity-feed.tsx:148` | no |

The one guard that exists also silently truncates — rows past 100 render a `0` count rather than their real one.

## Fix direction

1. Make the cap impossible to exceed at the boundary rather than at each call site: have `useBulkVoteSummaries` chunk internally (it can delegate to `useChunkedBulkVoteSummaries`), or delete it and migrate the three remaining callers.
2. Do the same for web — chunk inside `VoteSummaryProvider` and drop `VOTE_SUMMARY_BATCH_LIMIT` from `logbook-view`, so long lists get correct counts instead of truncated ones.
3. Add a regression test that drives >100 IDs through both entry points and asserts more than one request goes out, none exceeding 100.

Splitting into parallel 100-ID queries is already the established pattern here, so results stay complete and each chunk caches independently.

marcodejongh:
## Summary

`bulkVoteSummaries` rejects any request over 100 entity IDs (`BulkVoteSummaryInputSchema.entityIds.max(100)` in `packages/backend/src/validation/schemas/comments.ts`), but the paginating feeds hand it an unbounded list that grows with every page. Once the list crosses 100 the whole query is rejected outright, so every vote button on screen loses its count and the viewer's own vote state (Sentry BOARDSESH-CJ, and the bulk of BOARDSESH-7H).

- **Mobile:** `useBulkVoteSummaries` (`packages/mobile/src/lib/graphql/hooks/use-social.ts`) now chunks internally — one `useQueries` entry per ≤100-ID batch, merged back into a single list, no caller changes needed. `SOCIAL_PAGE_SIZE` is 30, so the Home feed (`app/(tabs)/home/index.tsx`) and You → Sessions (`SessionsTab.tsx`) both blow the cap on roughly the fourth page.
- **Web:** `VoteSummaryProvider` (`packages/web/app/components/social/vote-summary-context.tsx`) does the same chunking, which covers all four surviving providers at once: the activity feed, the global-ascents feed, the following-ascents feed and session detail. Those feeds page at `limit: 20`, so they cross the cap around the sixth page.
- **Backend cap left as-is.** Kept the hard reject rather than raising it or making the server silently truncate. A truncating server would turn a loud failure into quietly wrong counts; batching is what the callers actually needed.
- **Partial-failure isolation:** a chunk whose request fails contributes no rows, but the chunks that did resolve still render. One bad request no longer blanks the whole batch. Covered by tests on both sides.

### One home for the 100-ID cap

The chunk helpers and the cap were copied between web and mobile, and the number itself was written out in three places. They now live in `@boardsesh/shared-schema` as `batchVoteSummaryEntityIds` / `BULK_VOTE_SUMMARY_CHUNK_SIZE`, imported by both clients, and `BulkVoteSummaryInputSchema` reads the same constant.

The drift protection is structural rather than test-enforced: there is one constant with three consumers, so the client batch size and the server cap cannot disagree by construction. `social-validation.test.ts` derives its accept/reject boundary from that same constant, which documents the agreement but passes for any value of it — so it also carries one literal case (`101 IDs must be rejected; the constant is 100`) that a deliberate cap change has to acknowledge instead of sliding through.

### Keeping the merged results referentially stable

The `useQuery` → `useQueries` move would have quietly dropped React Query's structural sharing. Without a `combine`, `useQueries` hands back a freshly mapped array on every render (verified in the installed query-core 5.101.4: `#combineResult` returns its raw `input` unless `combine` is set), so the flattened summaries changed identity even when no vote had moved.

- **Mobile:** that churned the Home feed's `summaryMap`, which FlashList reads as `extraData` — the documented single signal that re-invokes `renderItem` — so every mounted row re-rendered on any unrelated parent render (mode switch, comment sheet, `isFetchingNextPage` flip).
- **Web:** it re-minted the `VoteSummaryProvider` context value on every provider render, re-rendering every `VoteButton` under a provider that can wrap 100+ of them.

Both sides now pass a module-scope `combine`, so React Query runs the merge through `replaceEqualDeep` and returns the previous value when nothing changed. Mobile's `combine` also carries the per-chunk `refetch` handles (each bound to its own observer, so stable across renders), which lets `useBulkVoteSummaries` memoize the `{ data, refetch }` object Home lists in its pull-to-refresh deps.

### Two orphaned hooks deleted

`useChunkedBulkVoteSummaries` and `useGroupedBulkVoteSummaries` had no caller anywhere in the repo — only their definitions, the hooks-barrel re-export and a `vi.mock` stub referenced them. Now that `useBulkVoteSummaries` chunks on its own there is no reason for them to grow one, so both are gone. Their two single-caller wrappers fold into `useBulkVoteSummaries`.

### Notes for the reviewer

- `useBulkVoteSummaries` returns `{ data, refetch }` rather than a `UseQueryResult`. There is no single query behind it to report `isLoading` / `isError` for, and an aggregate status across independently-retrying chunks would be a fiction. All three callers were checked: none reads anything but `.data` and `.refetch`.
- **A failed chunk does not self-heal.** Its rows keep whatever `initial*` props they were rendered with until the provider remounts or the chunk's query is invalidated — `VoteButton` skips its own single-entity fallback whenever a provider is above it, and relaxing that would fire one request per button on every cold render, which is the N+1 the provider exists to prevent.
- The mobile hook sends `entityIds` in the caller's order while the web provider sends them sorted. Both key on the sorted chunk, so both are correct; aligning them is cosmetic.
- This branch had already been rebased onto a `main` that dropped its edits to `logbook-view.tsx` and `crew-logbook-view.tsx` — those files were deleted from `packages/web` by #4467, along with the `VOTE_SUMMARY_BATCH_LIMIT` slice that used to truncate a long logbook to 100 rows. Nothing on web truncates any more, so that arm of the issue is gone with the surface.
- **Ships over the air.** No native config, plugin, module or native dependency is touched, so the Expo fingerprint does not move.
- This PR was carried forward across several review rounds (Claude Review, an independent Fable pass, and a follow-up review against the installed `@tanstack/query-core`) before this session rebased it onto current `main` (clean, no conflicts) and rewrote the Test plan / Risk sections to match the current PR template. The implementation is unchanged from what was already reviewed and marked ready to merge.

### Author-side verification (this session)

- `vp run typecheck` — pending in CI (local `vp install` is contending with ~15 other agents' installs in this sandbox).
- `vp check` — pending in CI, same reason.
- Read through the full diff against current `main`: `vote-summary-batching.ts`, the backend schema, both hooks/providers, and every caller (`SessionsTab.tsx`, `home/index.tsx`, `SessionDetailScreen.tsx`, `activity-feed.tsx`, `global-ascents-feed.tsx`, `following-ascents-feed.tsx`, `session-detail-content.tsx`) — every previously-unchunked caller now goes through a chunking entry point, and the two dead hooks stay deleted.
- `packages/shared-schema/src/__tests__/vote-summary-batching.test.ts` — empty input yields no batches, exactly-at-cap stays one batch, 235 IDs split `[100, 100, 35]` with `batches.flat()` equal to the input, duplicates dropped in first-seen order, duplicates don't eat cap headroom.
- `packages/backend/src/__tests__/social-validation.test.ts` — accepts exactly `BULK_VOTE_SUMMARY_CHUNK_SIZE` IDs and rejects one more (both derived from the shared constant), plus a literal case pinning the wire cap at 100.
- Mobile and web regression tests both drive 135 IDs through the real entry point and assert: exactly 2 requests, `entityIds.length <= 100` on every one of them, all 135 IDs covered; a failing sibling chunk leaves the resolved chunk's 100 rows intact with `tick-100` absent; and the returned value holds identity across re-renders that pass a fresh `[...entityIds]` array each time.
- No visual change — data fetching only, no markup touched.

Closes #4102

## Release Notes

- Like counts and your own likes keep loading as you scroll deep into the Home feed and your Sessions list. Past about the fourth page they used to stop showing up entirely.

## Test plan

1. Scroll Home past 100 posts. Like counts stay visible, not blank.
2. In Sessions, scroll past several pages of ticks. Vote counts stay visible.
3. Tap a heart far down a long list. It fills in.
4. On web, scroll a long ascents feed. Counts stay real, not zero.

## Risk

Risk: 2/5 — touches the vote-count fetch path on every feed on both platforms, but it's data-only (no markup changes), each chunk fetches and fails independently, and the logic carries unit + integration-style tests reviewed across several independent passes.

https://claude.ai/code/session_01WSwiTWgtzw4iyotNLPf5sA

### DIFF
--- a/packages/backend/src/__tests__/social-validation.test.ts
+++ b/packages/backend/src/__tests__/social-validation.test.ts
@@ -311,9 +311,11 @@ describe('Social Validation Schemas', () => {
 
   // Regression coverage for issue #4102: a paginating feed handed this schema
   // its whole accumulated list and every request past ~100 rows was rejected
-  // outright. The clients now batch by BULK_VOTE_SUMMARY_CHUNK_SIZE, so these
-  // cases derive their boundary from that same constant — raising one without
-  // the other fails here.
+  // outright. The clients now batch by BULK_VOTE_SUMMARY_CHUNK_SIZE, which is
+  // the same constant this schema's `.max()` reads, so the derived cases below
+  // show client and server agreeing at the boundary for whatever the constant
+  // says. They pass for any value of it by construction — the literal case
+  // pins what that value actually is on the wire.
   describe('BulkVoteSummaryInputSchema', () => {
     it('should accept a populated entityIds array', () => {
       const result = BulkVoteSummaryInputSchema.safeParse({
@@ -346,6 +348,19 @@ describe('Social Validation Schemas', () => {
       const result = BulkVoteSummaryInputSchema.safeParse({
         entityType: 'tick',
         entityIds: Array.from({ length: BULK_VOTE_SUMMARY_CHUNK_SIZE + 1 }, (_unused, index) => `id-${index}`),
+      });
+      expect(result.success).toBe(false);
+    });
+
+    // The two cases above move with the constant. This one does not: it states
+    // the cap this endpoint accepts today, so widening it is a deliberate edit
+    // here rather than a silent side effect of changing a client constant.
+    it('should reject 101 entityIds — the wire cap is 100', () => {
+      expect(BULK_VOTE_SUMMARY_CHUNK_SIZE).toBe(100);
+
+      const result = BulkVoteSummaryInputSchema.safeParse({
+        entityType: 'tick',
+        entityIds: Array.from({ length: 101 }, (_unused, index) => `id-${index}`),
       });
       expect(result.success).toBe(false);
     });
--- a/packages/backend/src/__tests__/social-validation.test.ts
+++ b/packages/backend/src/__tests__/social-validation.test.ts
@@ -11,6 +11,7 @@ import {
   BulkVoteSummaryInputSchema,
   VoteInputSchema,
 } from '../validation/schemas';
+import { BULK_VOTE_SUMMARY_CHUNK_SIZE } from '@boardsesh/shared-schema';
 
 describe('Social Validation Schemas', () => {
   describe('FollowInputSchema', () => {
@@ -308,6 +309,11 @@ describe('Social Validation Schemas', () => {
     });
   });
 
+  // Regression coverage for issue #4102: a paginating feed handed this schema
+  // its whole accumulated list and every request past ~100 rows was rejected
+  // outright. The clients now batch by BULK_VOTE_SUMMARY_CHUNK_SIZE, so these
+  // cases derive their boundary from that same constant — raising one without
+  // the other fails here.
   describe('BulkVoteSummaryInputSchema', () => {
     it('should accept a populated entityIds array', () => {
       const result = BulkVoteSummaryInputSchema.safeParse({
@@ -328,10 +334,18 @@ describe('Social Validation Schemas', () => {
       }
     });
 
-    it('should reject more than 100 entityIds', () => {
+    it('should accept exactly the shared chunk size, so clients can batch right up to it', () => {
       const result = BulkVoteSummaryInputSchema.safeParse({
         entityType: 'tick',
-        entityIds: Array.from({ length: 101 }, (_unused, index) => `id-${index}`),
+        entityIds: Array.from({ length: BULK_VOTE_SUMMARY_CHUNK_SIZE }, (_unused, index) => `id-${index}`),
+      });
+      expect(result.success).toBe(true);
+    });
+
+    it('should reject one entityId more than the shared chunk size', () => {
+      const result = BulkVoteSummaryInputSchema.safeParse({
+        entityType: 'tick',
+        entityIds: Array.from({ length: BULK_VOTE_SUMMARY_CHUNK_SIZE + 1 }, (_unused, index) => `id-${index}`),
       });
       expect(result.success).toBe(false);
     });
--- a/packages/backend/src/validation/schemas/comments.ts
+++ b/packages/backend/src/validation/schemas/comments.ts
@@ -1,4 +1,5 @@
 import { z } from 'zod';
+import { BULK_VOTE_SUMMARY_CHUNK_SIZE } from '@boardsesh/shared-schema';
 
 /**
  * Social entity type validation schema
@@ -75,5 +76,7 @@ export const CommentsInputSchema = z.object({
 export const BulkVoteSummaryInputSchema = z.object({
   entityType: SocialEntityTypeSchema,
   // Empty arrays are a valid no-op: the resolver returns []. (No `.min(1)` here.)
-  entityIds: z.array(z.string().min(1).max(200)).max(100),
+  // The cap lives in @boardsesh/shared-schema so the clients that have to
+  // batch around it can't drift from what this rejects.
+  entityIds: z.array(z.string().min(1).max(200)).max(BULK_VOTE_SUMMARY_CHUNK_SIZE),
 });
--- a/packages/mobile/src/components/session/SessionDetailScreen.tsx
+++ b/packages/mobile/src/components/session/SessionDetailScreen.tsx
@@ -49,9 +49,8 @@ export default function SessionDetailScreen() {
 
   const { data: session, isPending } = useSessionDetail(sessionId);
   const { data: voteSummaries } = useBulkVoteSummaries('session', sessionId ? [sessionId] : [], !!sessionId);
-  // `.at(0)` rather than `[0]`: the array is always present now that the hook
-  // merges its chunks, but it is empty until the request resolves, and `.at`
-  // is the indexed read TypeScript types as `VoteSummary | undefined` without
+  // `.at(0)`, not `[0]`: the list is empty until the chunk resolves, and `.at`
+  // is the indexed read typed `VoteSummary | undefined` without
   // `noUncheckedIndexedAccess`.
   const sessionVoteSummary = voteSummaries.at(0);
   const { data: profile } = useProfile();
--- a/packages/mobile/src/components/session/SessionDetailScreen.tsx
+++ b/packages/mobile/src/components/session/SessionDetailScreen.tsx
@@ -49,7 +49,7 @@ export default function SessionDetailScreen() {
 
   const { data: session, isPending } = useSessionDetail(sessionId);
   const { data: voteSummaries } = useBulkVoteSummaries('session', sessionId ? [sessionId] : [], !!sessionId);
-  const sessionVoteSummary = voteSummaries?.[0];
+  const sessionVoteSummary = voteSummaries[0];
   const { data: profile } = useProfile();
 
   // Only the session's creator can rename it / edit the recap (the server enforces
--- a/packages/mobile/src/components/session/SessionDetailScreen.tsx
+++ b/packages/mobile/src/components/session/SessionDetailScreen.tsx
@@ -49,7 +49,11 @@ export default function SessionDetailScreen() {
 
   const { data: session, isPending } = useSessionDetail(sessionId);
   const { data: voteSummaries } = useBulkVoteSummaries('session', sessionId ? [sessionId] : [], !!sessionId);
-  const sessionVoteSummary = voteSummaries[0];
+  // `.at(0)` rather than `[0]`: the array is always present now that the hook
+  // merges its chunks, but it is empty until the request resolves, and `.at`
+  // is the indexed read TypeScript types as `VoteSummary | undefined` without
+  // `noUncheckedIndexedAccess`.
+  const sessionVoteSummary = voteSummaries.at(0);
   const { data: profile } = useProfile();
 
   // Only the session's creator can rename it / edit the recap (the server enforces
--- a/packages/mobile/src/lib/graphql/hooks/__tests__/use-social.test.tsx
+++ b/packages/mobile/src/lib/graphql/hooks/__tests__/use-social.test.tsx
@@ -3,8 +3,14 @@ import { describe, expect, it, vi, beforeEach } from 'vitest';
 import { act, renderHook, waitFor } from '@testing-library/react';
 import type { ReactNode } from 'react';
 import { QueryClient, QueryClientProvider } from '@tanstack/react-query';
-import { FOLLOW_USER, GET_FOLLOWERS, SEARCH_USERS, UNFOLLOW_USER } from '@boardsesh/graphql/operations';
-import type { FollowConnection, UserSearchConnection } from '@boardsesh/shared-schema';
+import {
+  FOLLOW_USER,
+  GET_BULK_VOTE_SUMMARIES,
+  GET_FOLLOWERS,
+  SEARCH_USERS,
+  UNFOLLOW_USER,
+} from '@boardsesh/graphql/operations';
+import type { FollowConnection, UserSearchConnection, VoteSummary } from '@boardsesh/shared-schema';
 
 const requestMock = vi.hoisted(() => vi.fn());
 
@@ -12,7 +18,7 @@ vi.mock('../../client', () => ({
   getHttpClient: () => ({ request: requestMock }),
 }));
 
-import { useFollowers, useSearchUsers, useToggleUserFollow } from '../use-social';
+import { useBulkVoteSummaries, useFollowers, useSearchUsers, useToggleUserFollow } from '../use-social';
 
 function makeWrapper() {
   const queryClient = new QueryClient({
@@ -36,6 +42,10 @@ function makeFollowConnection(offset: number): FollowConnection {
     totalCount: 90,
     hasMore: offset < 60,
   };
+}
+
+function makeVoteSummary(entityId: string): VoteSummary {
+  return { entityType: 'tick', entityId, upvotes: 1, downvotes: 0, voteScore: 1, userVote: 0 };
 }
 
 function makeSearchConnection(offset: number): UserSearchConnection {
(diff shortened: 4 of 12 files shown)


marcodejongh:
## Fable review follow-up

Fable's review confirmed the chunking fix itself (requests stay under the 100-ID cap, results merge, a failed chunk no longer blanks its siblings, the `LogbookView` un-truncation is safe, and mobile `useVote`'s prefix cache patch still reaches the chunked keys), but flagged that the `useQuery` → `useQueries` move silently dropped React Query's referential stability. Two majors and one minor, all addressed in d9aba39.

### 1. `useBulkVoteSummaries` churned identity every render (major)

Confirmed against the installed `@tanstack/query-core@5.101.4`: `QueriesObserver.#combineResult` returns its raw `input` untouched unless `combine` is set, so `useQueries` hands back a freshly mapped array on every render and the `flatMap` allocated another on top. That made `.data` change identity even when no vote had moved, so Home's `summaryMap` `useMemo` (keyed on `[sessionVoteSummaries.data, tickVoteSummaries.data]`) rebuilt on every parent render and FlashList `extraData` re-invoked `renderItem` for every mounted row — the exact churn `buildVoteSummaryMap` and the mobile perf checklist exist to prevent.

`useBulkVoteSummaryChunks` now passes a module-scope `combine`, so React Query runs the merge through `replaceEqualDeep` and returns the previous value when nothing changed. The combine also carries the per-chunk `refetch` handles — those are bound per observer and therefore stable, so `replaceEqualDeep` keeps that array too — which lets `useBulkVoteSummaries` memoize the `{ data, refetch }` wrapper that Home's `handleRefresh` lists in its deps. `useChunkedBulkVoteSummaries` and `useGroupedBulkVoteSummaries` share the same path.

### 2. `VoteSummaryProvider` re-minted its context value every render (major)

Same root cause: `summariesMap` was keyed on the unstable `chunkResults`, so the context value changed identity on every provider render and every `VoteButton` beneath re-rendered — under providers that wrap 100+ buttons, including the `LogbookView` this PR un-truncates. The provider now takes the same `combine` and builds the `Map` from the stable flattened array.

### 3. Chunk helpers duplicated between web and mobile (minor)

`dedupeEntityIds`, `chunkEntityIds`, and the `100` constant moved into `@boardsesh/shared-schema` as `batchVoteSummaryEntityIds` / `BULK_VOTE_SUMMARY_CHUNK_SIZE`, imported by both clients. The backend's `BulkVoteSummaryInputSchema` now reads the same constant instead of hardcoding `.max(100)`, so the cap went from three independent copies to one.

### Coverage

New identity-stability tests on both sides re-render with a fresh `entityIds` array each time and assert the merged value holds its reference. Mutation-tested: reverting each side to the pre-`combine` shape fails exactly those two new tests and nothing else. Plus `vote-summary-batching.test.ts` for the shared helper (empty, at-cap, over-cap, dedupe order, duplicates not eating cap headroom).

Rebased onto latest `main` while here. Full run: mobile 5119/5119, web 5871/5871, shared-schema 137/137, `vp run check:mobile-bundle` OK.

marcodejongh:

marcodejongh:
🤖 Adopting this PR — P1 loop `moarbugs10000` (session 63bde115).

`moarbugs100500` claimed it on 2026-08-19 and there's been no movement in the five days since (no pushes, no comments), so this loop is picking it up as part of a five-ticket P1 sweep. Plan → adversarial plan review → implementation → code review → CI and feedback shepherding until it lands.

`moarbugs100500`: if you're still on this, say so here and I'll hand it straight back.

marcodejongh:
🤖 Rebased onto current `main` and finished. `MERGEABLE` again, CI green on `7986a37`, and the description has been rewritten — it was describing a codebase that no longer exists.

**What the rebase dropped.** The branch was 691 commits behind, and the only conflict was two modify/delete pairs: `logbook-view.tsx` and its test, deleted from `packages/web` by #4467 along with `crew-logbook-view.tsx`. Main has not touched any of the other files since the branch point (empty `git diff --stat <merge-base> origin/main` on each), so the rebase resolves to main's deletion plus the branch's content verbatim. The `VOTE_SUMMARY_BATCH_LIMIT` slice that used to truncate a long logbook to 100 rows went with the file, so the "silent truncation" arm of #4102 is gone with the surface. Web is down to four `VoteSummaryProvider` call sites: activity feed, global ascents, following ascents, session detail.

**Review threads.**

1. *`useChunkedBulkVoteSummaries` is dead code* — deleted, along with `useGroupedBulkVoteSummaries`, which is dead by the same grep and lives in the same file. Their two single-caller wrappers fold into `useBulkVoteSummaries`. `grep -rn 'useChunkedBulkVoteSummaries\|useGroupedBulkVoteSummaries' packages/` is now empty. Flagging the second deletion explicitly since only the first was named — it's a one-commit revert if you'd rather keep it.
2. *Chunk helpers duplicated web↔mobile* — already fixed in `d9aba39`: `batchVoteSummaryEntityIds` / `BULK_VOTE_SUMMARY_CHUNK_SIZE` live in `@boardsesh/shared-schema` and the backend schema reads the same constant. Newly added on top: `social-validation.test.ts` derives both its accept and reject cases from that constant, so schema-vs-client drift fails CI. Verified non-vacuous — hardcoding the schema back to a literal while the constant moves fails the accept case.
3. *Return shape narrowed to `{ data, refetch }`* — kept deliberately. All three callers were checked and none reads `isLoading`/`isError`; there is no single query to report a status for, and an aggregate across independently-retrying chunks would be a fiction. The JSDoc says so and points at `useQueries`. The unreachable `voteSummaries?.[0]` in `SessionDetailScreen` is gone.
4. *Referential stability from `useQuery` → `useQueries`* — already fixed in `d9aba39` via a module-scope `combine`, and both identity tests are still here. Mutation-checked on this branch: removing `combine` entirely fails exactly those two tests and nothing else.

Also cleaned up the query-key comment on both sides — the sort makes the key order-independent *within* a chunk, but chunk boundaries still follow the caller's order, which is what keeps an append-only feed's earlier chunks cached as it pages. The earlier wording implied more than the code does.

Still open and untouched: `token` is not in the web query key, so a token rotation won't refetch. Pre-existing, not introduced here.

Ships over the air — no native config, plugin, module or dependency touched.

marcodejongh:
## Review round 2 — four findings, all applied (91f3261)

An independent review confirmed the chunking itself against the installed `@tanstack/query-core` 5.101.4 (`#combineResult` always runs `combine` when set, so `data` is `[]` and never `undefined`; `bindMethods()` binds `refetch` once in the observer constructor, so `refetchChunks` really does hold identity; `refetch()` can't reject because `#executeFetch` appends `.catch(noop)` when `throwOnError` is unset) and could not break it. It also re-ran the vacuity checks on the six regression tests and found them sound.

What it blocked on was four cheap accuracy problems. All four are fixed:

**1. The provider JSDoc lied about recovery.** It claimed a failed chunk self-heals because "VoteButton falls back to its own single-entity fetch". `vote-button.tsx` does the opposite — its fallback effect hits `if (voteSummaryCtx) return;` whenever a provider is above it, so the 35 rows behind a failed second chunk keep their `initial*` props (the viewer's own likes read as un-liked) until the provider remounts.

I fixed the comment rather than the code. Making the comment true would mean relaxing that guard to `if (voteSummaryCtx?.getVoteSummary(entityId)) return;` — but `getVoteSummary` also returns `undefined` while the batch is still *pending*, so every button would fire its own request on first render. That's the N+1 the provider exists to prevent. The JSDoc now says plainly that those rows do not recover on their own.

**2. The two backend boundary cases were tautologies.** Both derived from `BULK_VOTE_SUMMARY_CHUNK_SIZE`, which is the same constant `comments.ts:81` feeds to `.max()` — so they pass for any value, and the rewrite had deleted the literal-101 case that was the only thing pinning the wire cap to a number. Restored one alongside the derived pair:

```ts
it('should reject 101 entityIds — the wire cap is 100', () => {
  expect(BULK_VOTE_SUMMARY_CHUNK_SIZE).toBe(100);
  // ... 101 ids must be rejected
});
```

The PR body carried the same overclaim ("raising the client batch size without raising the server cap fails CI"). Corrected: the drift protection here is *structural* — one constant, three consumers — not something `social-validation.test.ts` enforces. The mutation that actually exercises the schema guard is hardcoding `.max()` to a different literal.

**3. `voteSummaries[0]` typed a lie.** Dropping the `?.` narrowed it to `VoteSummary` while the runtime value is `undefined` until the chunk resolves (mobile's tsconfig sets `strict` but not `noUncheckedIndexedAccess`). Today `SessionSummaryCard` masks it with `voteSummary?.upvotes ?? session.upvotes`, but the next person to write `sessionVoteSummary.upvotes` would get a green typecheck and a crash on every cold open. Now `voteSummaries.at(0)`, which TypeScript types `VoteSummary | undefined`.

**4. Stale post-#4467 narrative in the code.** The test header still said "seven VoteSummaryProvider call sites (logbook, feeds, session detail)" and the provider JSDoc still described the deleted caller-side cap. There are four providers (activity feed, global ascents, following ascents, session detail), and `VOTE_SUMMARY_BATCH_LIMIT` no longer exists anywhere in the repo. Fixed in the tree, not just in the PR description.

### Mutation evidence for the new assertions

Set `BULK_VOTE_SUMMARY_CHUNK_SIZE = 500` and re-ran:

- `vote-summary-batching.test.ts` — 1 failed / 144 passed. `expect(batches.map(b => b.length)).toEqual([100, 100, 35])` got `[235]`.
- `vote-summary-context.test.tsx` — 2 failed / 1 passed. The over-cap split test reds on the request count; the partial-failure test reds too, because with a single chunk there's no sibling left to isolate.
- Backend: the derived pair still passes (as designed — it pins client/server agreement, not the number); the new literal case reds on `toBe(100)`.

Reverted, and re-ran unmutated: all 5 `BulkVoteSummaryInputSchema` cases pass, all 3 web provider cases pass, `vp run typecheck:mobile` is clean.

No behaviour change in this round beyond the `[0]` → `.at(0)` type narrowing.

marcodejongh:
Thanks — taking these one at a time.

**Overlong `.at(0)` comment** — fixed. Only the `noUncheckedIndexedAccess` half was non-obvious; the first sentence restated what `.at` does. (For the record, I grepped `CLAUDE.md` for a comment-style rule and there isn't one, so this is a taste call, not a rule violation. It's still the better comment.)

**`token` not in the web `queryKey`** — not acting on this, and I think the stated mechanism is wrong. I checked it in the installed `@tanstack/query-core` 5.101.4 rather than reasoning about it:

- `QueryObserver.setOptions` runs on every render and ends with `this.#currentQuery.setOptions(this.options)` (`queryObserver.js:94`).
- `Query.setOptions` is `this.options = { ...this.#defaultOptions, ...options }` (`query.js:48`) — it replaces the stored `queryFn` with the one from the latest render.
- `Query.fetch` resolves the function at fetch time via `ensureQueryFn(this.options, fetchOptions)` inside `fetchFn` (`query.js:221`).

So a refreshed token is already picked up by the next fetch: `queryFn` is re-supplied on every render of `VoteSummaryProvider`, and nothing caches the old closure past that render. The only thing that persists is *data* fetched under a previously-valid token, which is still correct data.

Adding `token` to the `queryKey` would make this worse, not better: every token refresh would mint a whole new set of cache keys and refetch all N chunks from scratch, on a feed where N grows with scroll depth. `enabled: isAuthenticated && !isAuthLoading && !!token` already keeps the queries from running without a token. Happy to be shown a concrete failing sequence if there is one, but I don't want to change caching behaviour on a mechanism I've read the source for and can't reproduce.

**Web/mobile query-key shape (`sortedIds.join(',')` vs `[...chunk].sort()`)** — agreed it's inconsistent, and agreed it's not a bug: the two apps don't share a cache, and both are order-independent within a chunk, which is the property that matters. The PR body already flags it as cosmetic. Leaving it out of this diff so the change stays limited to the chunking fix and the review follow-ups; it's a one-line alignment whenever someone touches either file next.

marcodejongh:
All three of these are already covered — nothing new to change.

**Return shape narrowed to `{ data, refetch }`** — deliberate, and it's thread 3 in the earlier round. There is no single query behind the hook any more, so `isLoading` / `isError` would have to be invented from N independently-retrying chunks. Every caller was read: `home/index.tsx` uses `.data` (166, 172, 179) and `.refetch` (210, 211), `SessionsTab.tsx` uses `.data` (155, 158, 162), `SessionDetailScreen.tsx` uses `.data` (51). None touches a status field, which is why only the last one appears in the diff. The narrowing breaks loudly at typecheck rather than silently at runtime — a caller destructuring `isLoading` gets TS2339, not `undefined`, because the returned object is a concrete type with two keys, not `UseQueryResult`. The JSDoc records the shape and points at `useQueries` for per-chunk state.

**`chunks` memo keyed on the array reference** — already in the docstring, verbatim:

```
// Chunked once per list change rather than per render. Memoized on the array
// reference, so a caller that builds its list inline
// (`ticks.map((tick) => tick.uuid)`) re-chunks every render — harmless,
// since `useQueries` matches its observers by query hash either way, and the
// feed screens that hand over the long lists already memoize them.
```

Both feed screens do memoize (`home/index.tsx:141-155`, `SessionsTab.tsx:138-152`); `SessionDetailScreen` passes a one-element list where re-chunking costs nothing.

**Query-key shape web vs mobile** — answered in the previous comment: separate query clients, both order-independent within a chunk, cosmetic to align. Out of this diff on purpose.

marcodejongh:
## Green on `636d155` — ready for a human review

All 28 non-skipped checks pass, including `lint`, `typecheck`, `i18n`, `codegen-drift`, `commit-lint`, `mobile-bundle`, `release-notes`, `ota-check`, and all four test jobs (`test-backend`, both `test-default` shards, `test-ocr`, `test-location-sync-integration`). `MERGEABLE`, no unresolved review threads. `mergeState: BLOCKED` / `reviewDecision: REVIEW_REQUIRED` is just the branch rule wanting a reviewer — I can't approve my own PR. Not merging; that's Marco's call.

### What the independent review turned up, and what changed

It re-verified the chunking mechanism against the installed `@tanstack/query-core` 5.101.4 instead of trusting the JSDoc, and couldn't break it. It also re-ran the vacuity checks on the six regression tests — reverting the fix makes them fail, so they're real. What it blocked on was four accuracy problems, all now fixed in `91f3261`:

1. **The provider JSDoc described recovery that doesn't happen.** It said a failed chunk self-heals via VoteButton's single-entity fallback; that fallback short-circuits whenever a provider is above it. Comment corrected rather than code changed — relaxing the guard would fire one request per button on every cold render, which is exactly the N+1 the provider exists to prevent.
2. **Both backend boundary cases were tautologies** (derived from the same constant the schema's `.max()` reads), and the rewrite had deleted the only case pinning the cap to a number. Added a literal `101 must be rejected; the constant is 100` case back alongside them, and corrected the matching overclaim in the PR body — the drift protection is structural, not test-enforced.
3. **`voteSummaries[0]` typed `VoteSummary`** while the runtime value is `undefined` until the chunk resolves. Now `.at(0)`, which types honestly.
4. **Stale post-#4467 narrative in the tree** — "seven call sites", a caller-side cap that no longer exists. There are four providers; fixed in the code and test comments, not just the description.

`636d155` on top of that trims one comment per the review bot's nit.

### Mutation evidence

Setting `BULK_VOTE_SUMMARY_CHUNK_SIZE` to 500 and re-running: `vote-summary-batching.test.ts` reds (`[235]` vs the expected `[100, 100, 35]`), and the web provider suite goes 2 failed / 1 passed — the split test on the request count, the partial-failure test because one chunk leaves no sibling to isolate. Reverted; unmutated runs are green.

### Two review-bot comments I did not act on, with reasoning

- **`token` not in the web `queryKey`.** I read the installed query-core rather than reasoning about it: `QueryObserver.setOptions` runs every render and calls `Query.setOptions`, which replaces the stored `queryFn`; `Query.fetch` resolves it at fetch time via `ensureQueryFn(this.options, …)`. A refreshed token is already used by the next fetch. Adding `token` to the key would refetch every chunk on each token refresh on a feed whose chunk count grows with scroll.
- **Query-key shape differs web vs mobile** (`join(',')` vs an array). Separate query clients, both order-independent within a chunk. Cosmetic; kept out of this diff.

Both are recorded here rather than silently ignored — happy to be overruled on either.

PR #4131, opened 2026-08-01
fix(ci): report renumber push outcomes truthfully

### LINKED ISSUE #4100: db-renumber bot posts a success comment when proof-of-apply fails and nothing was pushed
## Summary

The migration renumber bot can post a '🔢 Renumbered onto main: 0188 → 0189' PR comment while having pushed **nothing** — leaving the PR numbered as before, colliding with main, and the maintainer believing it's resolved.

## Evidence

GH Actions run 30630207599 (job `renumber`, PR_NUMBER 4052): log shows `[db-renumber] status: renumbered` locally, then the proof-of-apply step (`db:migrate` + journal verification) failed on a pre-existing gap — CI's dev-db image lacks the ledger row for `0188_puzzling_kabuki` (same class as #3978) — job exited 1, `PUSHED:` empty. Per docs/db-migrations.md nothing is pushed on proof-of-apply failure (correct!), but the success comment was posted anyway at 12:22:53Z on PR #4052. Verified empirically: the PR head SHA never changed.

## Two fixes needed

1. **Comment truthfulness**: the workflow must only post the renumber-success comment after confirming the push happened (gate the comment step on the push step's outcome, or post a distinct '⚠️ renumber failed proof-of-apply — nothing pushed' comment on failure).
2. **Unblock the proof-of-apply**: the dev-db image ledger gap (#3978) currently makes EVERY renumber attempt on every branch fail — the bot is effectively down fleet-wide until the image is rebuilt or the migrate step handles schema-present-ledger-missing (the exact high-water-mark condition #3979 describes for dev-db-up.sh).

Found during the #4052 review by the `josh` batch loop. Unclaimed — either loop may take it (claim via assignee + comment per the batch protocol).

marcodejongh:
## Summary

Closes #4100.

- Make the migration renumber sticky comment depend on the actual force-push step outcome, rather than the pre-push re-check.
- Post a specific held comment with the action-run link when validation, proof-of-apply, re-check, or push does not complete; retain the existing paginated sticky upsert and blocked/no-op behavior.
- The dev-DB migration-ledger recovery remains separately owned by #3978/#3979. This PR does not make the bot operational while proof-of-apply still fails on that ledger gap.

## Release Notes

## Test plan

- `vp test run scripts/db-renumber-migration.test.ts scripts/lib/db-renumber-comment.test.ts --reporter=agent`
- `vp check scripts/lib/db-renumber-comment.cjs scripts/lib/db-renumber-comment.test.ts`
- `vp run typecheck`
- Scoped QA exception: this is workflow-only, so no `vp run dev` was started; that would exercise unrelated Docker-backed services and Docker access was declined.

### DIFF
--- a/.github/workflows/db-migration-renumber.yml
+++ b/.github/workflows/db-migration-renumber.yml
@@ -188,6 +188,7 @@ jobs:
             --strategy "${{ github.event.inputs.strategy || 'rebase' }}"
 
       - name: Validate the resulting migration folder
+        id: validate
         if: steps.renumber.outputs.status == 'renumbered'
         run: |
           bun .renumber-tooling/scripts/check-db-migrations.ts \
@@ -198,12 +199,14 @@ jobs:
       # the journal tail actually landed — which is what catches a migration whose
       # `when` would make it silently skipped forever.
       - name: Verify the migrations apply
+        id: apply
         if: steps.renumber.outputs.status == 'renumbered'
         env:
           VERIFY_MIGRATION_JOURNAL: '1'
         run: bun run --filter=@boardsesh/db db:migrate
 
       - name: Verify re-applying is a no-op
+        id: reapply
         if: steps.renumber.outputs.status == 'renumbered'
         run: bun run --filter=@boardsesh/db db:migrate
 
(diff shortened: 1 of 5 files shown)


marcodejongh:
Fable review addressed — both findings were minor polish; no correctness changes.

**1. Sticky marker triplicated with no drift guard.** `<!-- db-renumber -->` was defined in three uncoupled places: `scripts/db-renumber-migration.ts`, `scripts/lib/db-renumber-comment.cjs`, and inline as `const marker` in the comment job's upsert lookup. A drift in any copy would stop the lookup finding the existing comment and post a duplicate every run.

- The comment job now destructures `STICKY_MARKER` from the module it already `require`s, so the third copy is gone.
- `scripts/lib/db-renumber-comment.test.ts` asserts the `.cjs` marker equals the one `db-renumber-migration.ts` exports.
- `scripts/db-renumber-migration.test.ts` gains a source guard in the trust-boundary block: the comment job must take the marker from the module and must not carry the literal. Mutation-checked — reintroducing `const marker = '<!-- db-renumber -->'` fails the test.

**2. Stale gate rationale.** The `Require the push credential` comment still described the comment step as living inside the `renumber` job. Rewritten for the split: the fail-loud is what keeps the separate `comment` job (gated on `needs.gate.result == 'success'`) from posting a "renumbering was held" comment that blames the PR for an operator misconfiguration.

Also rebased onto `main` (the branch was 182 behind).

Validation: `vp test run scripts/db-renumber-migration.test.ts scripts/lib/db-renumber-comment.test.ts` (49 passed), `vp check` on the changed files, `vp run typecheck`.

PR #4134, opened 2026-08-01
fix(mobile): upload OTA source maps

### LINKED ISSUE #4112: OTA-published JS updates carry no Sentry source maps on either platform

marcodejongh:
## Summary

Fixes #4112.

- Ask EOAS/Expo 57 for external source maps for every self-hosted iOS and Android OTA export.
- Validate the requested platform's metadata-declared executable artifacts, exact adjacent maps, and Debug IDs before invoking the pinned official Sentry React Native 7.11 uploader.
- Upload only validated bundle/map pairs from an isolated temporary staging tree, with fixed Boardsesh Sentry scope and no release/dist override.
- Keep production changelog, notification, and health-check cleanup running after an upload failure, then fail the workflow explicitly; backports use the same trusted tooling, while PR previews never receive the Sentry token or upload.
- Ignore Expo-copied public JavaScript that is not declared executable in update metadata. Metadata-declared DOM JavaScript fails closed when Expo's independently hashed map cannot be paired safely.

## Release Notes

## Test plan

- Independent high-effort implementation and final QA review; the initial public-WASM artifact blocker was reproduced, remediated, and re-reviewed cleanly
- `vp test run --reporter=agent scripts/mobile-sourcemaps.test.ts scripts/mobile-ci-env-parity.test.ts` (58 focused/parity tests passed)
- Full scripts suite (547 tests passed)
- `vp run typecheck`
- Scoped `vp check` across all eight changed files
- `git diff --check`
- `vp run dev:mobile` confirmed `.boardsesh/qa-notes.md` and Metro startup on port 8082
- No publish, upload, Sentry mutation, or production change was performed; final acceptance requires one real iOS and Android OTA followed by Sentry symbolication verification

### DIFF
--- a/.github/workflows/mobile-ota-backport.yml
+++ b/.github/workflows/mobile-ota-backport.yml
@@ -243,7 +243,7 @@ jobs:
       - name: Validate anchored Sentry uploader support
         run: |
           set -euo pipefail
-          sentry_version="$(jq -r '.dependencies["@sentry/react-native"] // empty' packages/mobile/package.json)"
+          sentry_version="$(jq -r '.dependencies["@sentry/react-native"] // .devDependencies["@sentry/react-native"] // empty' packages/mobile/package.json)"
           if [ "$sentry_version" != "7.11.0" ]; then
             echo "::error::This anchor has @sentry/react-native $sentry_version; OTA source-map backports require exactly 7.11.0. Ship a native update instead."
             exit 1
@@ -315,6 +315,7 @@ jobs:
         id: upload_sourcemaps
         if: ${{ !inputs.dry_run && steps.publish.outcome == 'success' }}
         continue-on-error: true
+        timeout-minutes: 10
         env:
           SENTRY_AUTH_TOKEN: ${{ secrets.SENTRY_AUTH_TOKEN }}
           PLATFORM: ${{ matrix.platform }}
@@ -327,11 +328,11 @@ jobs:
         run: echo "::notice::dry_run — verified the fingerprint match for ${{ matrix.platform }} but skipped the publish. Re-run with dry_run unchecked to ship."
 
       - name: Warn that the published backport source maps are missing
-        if: ${{ always() && !inputs.dry_run && steps.upload_sourcemaps.outcome == 'failure' }}
-        run: echo "::warning::The ${{ matrix.platform }} backport OTA is live, but its Sentry source-map upload failed. Recreate the exact export before another eoas publish replaces dist."
+        if: ${{ always() && !inputs.dry_run && steps.publish.outcome == 'success' && steps.upload_sourcemaps.outcome != 'success' }}
+        run: echo "::warning::The ${{ matrix.platform }} backport OTA is live, but its Sentry source-map upload did not succeed. Recreate the exact export before another eoas publish replaces dist."
 
       - name: Require the published backport source-map upload
-        if: ${{ always() && !inputs.dry_run && steps.upload_sourcemaps.outcome == 'failure' }}
+        if: ${{ always() && !inputs.dry_run && steps.publish.outcome == 'success' && steps.upload_sourcemaps.outcome != 'success' }}
         run: |
           echo "::error::The published ${{ matrix.platform }} backport has no accepted Sentry source maps."
           exit 1
--- a/.github/workflows/mobile-ota-backport.yml
+++ b/.github/workflows/mobile-ota-backport.yml
@@ -216,6 +216,8 @@ jobs:
           cp "$tooling_root/scripts/mobile-upload-sourcemaps.ts" scripts/mobile-upload-sourcemaps.ts
           git add scripts/mobile-publish.ts scripts/lib/eoas.ts scripts/lib/mobile-publish-retry.ts scripts/mobile-upload-sourcemaps.ts
           if ! git diff --cached --quiet; then
+            git config user.name "github-actions[bot]"
+            git config user.email "github-actions[bot]@users.noreply.github.com"
             git commit -m "chore(ota): overlay trusted publish tooling"
           fi
           if [ -n "$(git status --porcelain)" ]; then
--- a/.github/workflows/mobile-ota-backport.yml
+++ b/.github/workflows/mobile-ota-backport.yml
@@ -130,6 +130,23 @@ jobs:
 
       - uses: oven-sh/setup-bun@v2
 
+      # The hotfix tree is about to move to an older release anchor, which may
+      # predate OTA source-map support. Snapshot the exact trusted workflow commit
+      # before that checkout; never take release tooling from a cherry-picked fix.
+      - name: Snapshot trusted OTA publish tooling
+        run: |
+          set -euo pipefail
+          tooling_root="$RUNNER_TEMP/ota-publish-tooling"
+          mkdir -p "$tooling_root/scripts/lib"
+          for source_path in \
+            scripts/mobile-publish.ts \
+            scripts/lib/eoas.ts \
+            scripts/lib/mobile-publish-retry.ts \
+            scripts/mobile-upload-sourcemaps.ts
+          do
+            git show "$GITHUB_SHA:$source_path" > "$tooling_root/$source_path"
+          done
+
       - name: Locate the release anchor and prepare the hotfix tree
         id: anchor
         env:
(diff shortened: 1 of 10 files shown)


marcodejongh:
Adversarially checked each point against the workflow and the pinned uploader at `b825cce120ef5085e57a24cfccd6b610fbc92c36`:

1. **The single-platform terminal gate is already correct.** For an iOS-only dispatch, Android is `skipped`; if the iOS upload fails, `continue-on-error` preserves `steps.upload_sourcemaps_ios.outcome == 'failure'`, so the terminal expression is `true || false` and fails the job. Android-only is symmetric, and `all` catches either failure. The workflow already exposes `ios | android | all` through `INPUT_PLATFORM`.
2. **A first-KB scan would be incorrect here.** Expo 57 appends `debugId` at the end of the encoded map, not near the beginning. Reading only a prefix can miss it and cannot prove the complete document is valid JSON. The pinned official Sentry Expo uploader then reads, parses, and rewrites the full map itself, so changing only this fail-closed validation wrapper would not remove the uploader's peak-memory behavior. A real streaming/uploader redesign can be a measured follow-up if runner memory ever becomes a problem; it is not safe to replace this with the suggested regex.
3. **The trusted backport snapshot has a complete dependency closure today.** `mobile-publish.ts` imports only `scripts/lib/eoas.ts`; the uploader and `eoas.ts` import only `node:*`. A future relative helper should be guarded by a recursive import-closure test or centralized manifest. That is worthwhile hardening, but it is not a missing runtime dependency in this head. A missing future helper would normally fail module resolution; the more subtle stale-anchor-helper case is why the closure test is the right follow-up.
4. **Transport retries already happen below this wrapper.** The official uploader invokes the pinned `sentry-cli` 2.58.4, whose default is five exponential-jitter retries for its transient 5xx set. An unconditional outer retry would duplicate those attempts and also repeat deterministic auth/validation failures.
5. **Leading `--` is intentional vp argument forwarding.** Production invokes `vp run mobile:upload-sourcemaps -- --platform ios|android`; this parser follows the same repo CLI convention so the forwarded delimiter is ignored before parsing the script's flags.

No additional blocker was found. The focused 23 source-map tests pass; no real publish or upload was performed. The PR remains gated on the already-disclosed real iOS+Android OTA publish/Sentry symbolication check.

marcodejongh:
This PR is now conflicted after the overnight merges and has had no branch activity since the last review. I am taking over the conflict resolution per the bug-loop lifecycle: rebasing fix/4112-ota-sentry-sourcemaps onto current main, preserving the reviewed source-map behavior, validating the resolved diff, and force-pushing only with an exact lease.

marcodejongh:
Conflict takeover is complete at `ddf313551`. I rebased the source-map work onto the current OTA/fingerprint pipeline, preserved sequential retry and partial-publish reporting, repaired the executed EOAS source-map flags and backport retry-helper overlay, and force-pushed with the announced exact lease. Independent final QA passed 47 focused tests, full typecheck, fingerprint-input validation, and diff checks. The PR remains draft while its real OTA/Sentry symbolication validation is outstanding.

marcodejongh:
Addressed the two minor review notes in d32f94525. The official uploader now has direct startup-error and nonzero-exit tests, and both prove the temporary cwd/staging tree is removed. The trusted-tooling overlay step configures the Actions bot identity inside its conditional commit block; the workflow test enforces guard → identity → commit ordering. Independent QA passed, the source-map suite is 25/25, scoped checks and diff checks are clean, and the full monorepo typecheck passed.

marcodejongh:
Addressed the latest OTA notification/diagnostics review in 01b3286d7.

- A failed source-map upload now emits its Actions warning and exactly one Discord failure alert; that alert reports iOS/Android publish and source-map outcomes separately, so successful publishes are not presented as a fully healthy workflow.
- Trusted backport snapshot failures name the exact source path and trusted SHA.
- The official Sentry 7.11 uploader takes one staging directory rather than one invocation per pair. The test now asserts exactly one invocation, two validated artifacts, and all four staged files, proving neither pair is dropped.

Independent QA passed. The focused suite is 25/25, both workflows parse as YAML, scoped `vp check` passed, `git diff --check` is clean, and the pre-commit hook passed.

marcodejongh:
Folded the final nonblocking cleanup into 7f9b38b6f. Only the exact `production` channel now retains EOAS source maps in `dist`; self-hosted `pr-N` and EAS previews do not. Both production uploader steps have a 10-minute timeout, and the backport contract test now proves the source-map helper is snapshotted, overlaid, and staged. The preview workflow comment was updated to match that behavior. Independent QA accepted the change; 39/39 focused tests, scoped `vp check`, YAML parsing, `git diff --check`, and the pre-commit hook passed.

marcodejongh:
Hardened the remaining gate/backport edge cases in 61fe09c43. A platform is gated only if its OTA publish succeeded, but every source-map outcome other than exact `success` now fails closed; unrequested platforms remain safely skipped. The same rule and a 10-minute timeout now cover backports. Anchored uploader detection accepts dependency or devDependency placement, all four trusted tooling files are contract-tested, and the runbook removes duplicate copy and explains the native-update/new-anchor fallback. Independent QA passed with 39/39 tests, scoped checks, YAML parsing, diff checks, and the pre-commit hook.

marcodejongh:
Closed the final diagnostic-only review notes in 548d7e68a: the `-p` alias is covered, missing official uploader files have a direct regression test, and a missing/non-directory mobile root now reports a clear path-specific error before resolution. Directory symlink behavior remains unchanged. Independent QA passed with 26/26 source-map tests, scoped checks, diff checks, and the pre-commit hook.

Issue #4139, opened 2026-08-01
iOS BLE: late disconnect callback can tear down an immediate same-board reconnect


marcodejongh:
## Summary

When an explicit disconnect is followed immediately by reconnecting to the same peripheral UUID, the late CoreBluetooth callback for the cancelled connection can tear down the new link.

## Evidence

- `disconnectOnBleQueue` records the UUID as intentional, calls cancel, and clears the current link.
- `connectOnBleQueue` removes that UUID from the intentional-disconnect markers when the reconnect begins.
- `didDisconnectPeripheral` identifies the current peripheral only by UUID. CoreBluetooth does not provide a connection-generation token with the callback.
- A late callback from the old connection can therefore be classified as an unexpected disconnect of the new connection and clear it.
- Write-stall recovery already avoids this race by waiting for the cancellation callback before reconnecting; the general explicit disconnect/reconnect path does not.

## Expected behavior

Reconnect immediately to the same board without the prior cancellation callback tearing down the new connection.

## Acceptance criteria

- Serialize a same-UUID reconnect until the pending cancellation callback settles, or add another production mechanism that distinguishes ownership without inventing a callback generation CoreBluetooth does not provide.
- Add deterministic native coverage for disconnect → immediate same-UUID reconnect → late callback.
- Preserve immediate reconnects to a different UUID.
- Preserve existing write-stall recovery behavior and intentional-disconnect event suppression.

Discovered while adding truthful disconnect-path coverage in #3925.

marcodejongh:
🤖 Claimed by bug-batch loop `codex-moarbugs` (batch 3); parallel loops, please pick another issue.

PR #4140, opened 2026-08-01
test(mobile): cover iOS MoonBoard disconnects

### LINKED ISSUE #3925: Test gaps in the iOS MoonBoard BLE disconnect/write path (incl. one false-confidence test)
Three coverage holes found while triaging #3702. The disconnect and write-type paths on the native iOS MoonBoard route are load-bearing and effectively untested.

## 1. `didDisconnectPeripheral` has no test at all

`BoardBleManager.didDisconnectPeripheral` is the sole entry point for every unexpected link drop — it classifies the CoreBluetooth error, decides whether the drop was intentional (`intentionalDisconnectGenerations`), checks generation staleness (`peripheralGenerations` vs `connectionGeneration`), and emits the `disconnected` event JS keys on to darken the lightbulb.

There is no test exercising it. A generation-comparison mistake here would either swallow real drops (lightbulb stays lit on a dead board — the exact #3702 complaint) or fire spurious ones, and nothing in CI would notice.

## 2. Nothing asserts the MoonBoard with-response ack watchdog reaches `handleWriteStall()`

The with-response write path arms an ack watchdog; on timeout it's supposed to reach `handleWriteStall()`, which drives the recovery ladder and ultimately a deliberate disconnect. No test covers that hop.

This one is worth prioritising because it's the path that *would* produce a self-inflicted teardown mid-problem-change. Telemetry currently says it never fires for MoonBoard (`disconnectCategory = 'app_recovery'`: 9 events in 120 days, all Kilter, zero MoonBoard), which is the right answer — but "never fires" and "silently broken" look identical without a test.

## 3. `testRedBearLabWriteCharacteristicUsesWriteWithResponse` is a false-confidence test

The name promises it pins the 2016 RedBearLab box to write-with-response. It doesn't. The test never references the RedBearLab characteristic UUID `713D0003` — it only exercises the pure encoding helper.

So if the RedBearLab characteristic resolution path broke and that box started getting write-without-response, **this test would still pass**. That's worse than no test: it's a green check that reads as coverage for the exact regression it can't detect.

Either make it assert against the real characteristic UUID, or rename it to describe what it actually tests.

## Context

Field data says the write path is currently healthy — iOS MoonBoard `Climb Sent to Board` failure is 0.02% on `withResponse` (1 of 5077) and 0.0% on `withoutResponse` (0 of 4822). So this is protecting a working path, not chasing a live bug.

Found while investigating #3702.

marcodejongh:
## Summary

- route the real CoreBluetooth disconnect callback body through a queue-bound seam and cover intentional, stale-other-peripheral, clean-current, and NSError-current callbacks without inventing generation information
- prove MoonBoard ACK timeouts enter bounded write-stall recovery, and pin the original RedBearLab `713D0000` service to an actual `.withResponse` write on `713D0003`
- stage every tracked iOS XCTest source and add a drift check so future native test files cannot silently disappear from the generated Xcode target

Closes #3925

The separately discovered immediate same-UUID reconnect race is tracked in #4139; this coverage PR does not falsely claim to solve it.

## Release Notes

## Test plan

- `vp test run --project scripts scripts/__tests__/prepare-rn-ios-tests.test.ts --reporter=agent` (1 passed)
- `vp run typecheck:mobile`
- `vp run test:mobile` (575 files, 4,912 tests passed)
- `vp run check:mobile-bundle` (iOS and Android passed)
- scoped `vp check` on changed JavaScript/TypeScript files
- `vp run check:mobile-board-art-network`
- `git diff --check`
- independent adversarial BLE review passed with no correctness, fidelity, deadlock, or isolation findings
- started `vp run dev:mobile`; Metro loaded `.boardsesh/qa-notes.md` at the printed URL
- macOS `BoardseshTests` XCTest execution remains required in CI; physical original RedBearLab hardware remains follow-up evidence

Fable is unavailable in this environment, so an independent high-scrutiny Sol review was used as the available substitute; this does not claim literal Fable-policy compliance.

### DIFF
--- a/packages/mobile/ios-tests/BoardBleDisconnectTests.swift
+++ b/packages/mobile/ios-tests/BoardBleDisconnectTests.swift
@@ -127,6 +127,11 @@ final class BoardBleDisconnectTests: XCTestCase {
         var settledErrors: [Error?] = []
         installConnection(peripheral: currentPeripheral) { _, _ in disconnectEventCount += 1 }
         enqueueTwoWrites { settledErrors.append($0) }
+        // Seed a generation for the stale peripheral so the guard's cleanup line
+        // (removing a non-current peripheral's entry) has something to remove.
+        manager.testHooks.sync {
+            manager.testHooks.setPeripheralGeneration(7, for: stalePeripheral.identifier)
+        }
 
         let pendingWatchdog = scheduler.lastOneShot(label: "writeAckWatchdog")
         XCTAssertEqual(manager.testHooks.sync { manager.testHooks.writeQueueDepth }, 2)
@@ -134,6 +139,9 @@ final class BoardBleDisconnectTests: XCTestCase {
 
         manager.testHooks.fireDidDisconnect(peripheral: stalePeripheral, error: nil)
 
+        XCTAssertNil(
+            manager.testHooks.sync { manager.testHooks.peripheralGeneration(for: stalePeripheral.identifier) }
+        )
         XCTAssertEqual(manager.connectedDeviceId, currentPeripheral.identifier.uuidString)
         XCTAssertEqual(manager.testHooks.sync { manager.testHooks.writeQueueDepth }, 2)
         XCTAssertTrue(manager.testHooks.sync { manager.testHooks.isWriting })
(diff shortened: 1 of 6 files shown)


marcodejongh:
Addressed the style note in 727ed667d by removing the stray numbered comment.

The availability concern is verified rather than changed: every top-level Swift XCTest source in `packages/mobile/ios-tests` already uses `@available(iOS 17.0, *)`; `scripts/prepare-rn-ios-tests.mjs` sets `IPHONEOS_DEPLOYMENT_TARGET = 17.0`; and `ios-rn-ci.yml` runs `OS=latest`. The new disconnect class therefore aligns with the existing target and cannot be silently skipped by an iOS 16 CI destination.

The Fable availability constraint remains explicitly disclosed; independent high-scrutiny BLE QA passed, but this does not claim literal policy compliance.

marcodejongh:
The CI-only crash is repaired in `9d3c674b0`: both independent iOS native jobs now build and pass the full Swift suite, and the complete PR check set is green. The PR is mergeable with no review threads. I am leaving it draft only for the repository-required Fable review or explicit maintainer sign-off; the available independent Sol BLE review passed, but I am not representing that as literal Fable compliance.

marcodejongh:
The inherited native compile failure from the misplaced MoonBoard acknowledgement-watchdog test is repaired in draft follow-up #4173, alongside the post-merge #4152 callback-race fix. The test body is unchanged and now sits in the XCTest fixture that owns its manager/scheduler.

PR #4141, opened 2026-08-01
fix(ci): gate production OTA on server release

### LINKED ISSUE #3459: Consolidate production deploys: mobile OTA ships ahead of the backend/web
## Problem

On a push to `main`, production ships via **two independent workflows** that don't coordinate:

- **`mobile-ota-production.yml`** — publishes the JS OTA (and is the sole writer of `changelog.generated.json` → the mobile "What's New"). JS-only, so it finishes fast.
- **`production-deploy.yml`** — Railway backend + Vercel web + `db:migrate`. Builds + deploys, and is serialized by a `production-deploy` concurrency group, so it can sit **queued** behind an in-flight run and finish much later.

Neither waits for the other (`production-deploy` has no OTA step; the OTA workflow has no dependency on the deploy). Result: for features that span **mobile JS + backend**, the OTA can go live — and announce the feature in "What's New" — before the backend that actually implements it has deployed.

## Concrete incident (PR #3320, MoonBoard set filtering)

- Merge `bc07d67`. The production OTA published immediately and shipped this to "What's New":
  > ✨ MoonBoard search now hides climbs set on holds you don't have — deselect the wooden holds (or any set) you don't own and those climbs drop out of your results.
- But `production-deploy` for `bc07d67` sat **`queued`** for a long time; the Railway backend deployed later.
- During that window the mobile app had the new JS + the changelog claiming the feature worked, but the backend was still running the old code that **skips** MoonBoard set filtering. Verified live: `searchClimbs` `totalCount` for `setIds=5,6,7` and `setIds=5,6,7,8,9,10` both returned `40712` (identical → no filtering). The data backfill was already done; only the backend code was missing.

Net effect: an announced-but-dead feature, and confusing QA ("the board renders and I deselected wooden, but climbs aren't filtered").

## Why it matters

Any change touching both the mobile bundle and the backend/web can be half-shipped. The changelog is the worst part — it tells users a feature is live before it is, and "What's New" isn't easily retractable.

## Proposed fix

Make the OTA publish a **downstream, gated step** of the production deploy rather than a parallel workflow:

- Trigger `mobile-ota-production` via `workflow_run` on `production-deploy` **success**, or fold the OTA publish into `production-deploy` as a final job that `needs` the backend + web + `db:migrate` jobs.
- Enforce ordering `db:migrate → backend → web → OTA` so the JS never lands before the server that serves it.
- Keep the changelog write coupled to the OTA step (already the sole writer), so "What's New" only ships once the feature is actually live.
- Preserve the existing `production-deploy` concurrency/rollback guarantees.

## Acceptance

- A push to `main` that changes backend + mobile JS results in the OTA publishing **only after** the backend/web/migrate deploy succeeds.
- If the backend deploy fails or is rolled back, the OTA (and the "What's New" entry) does not ship.

_Filed after PR #3320; the backend later deployed and filtering now works — this is about preventing the ordering skew from recurring._

marcodejongh:
## Summary

- make Production Deploy the sole production OTA entrypoint so database, Railway, Vercel, browser app, and exact-SHA OTA outcomes settle in one release graph
- fail closed across rollback, failed, partial, held, unconfigured, superseded, and ambiguous-history states while preserving deliberate manual full-stack releases
- bound changelog inputs to the deployed commit and keep queued-run snapshots separate from the immutable release range

Closes #3459.

Rebased onto `main` after #4136 merged, so the retry-hardened OTA workflow, the gzip snapshot base, and the Sentry source-map upload from #4134 all live inside the new reusable publisher. The callee still runs `mobile:upload-sourcemaps` between the two publishes and now returns a `sourcemaps_status` output, so a live-but-unsymbolicated bundle reads differently from nothing published in the parent's one release notification.

## Release Notes

App updates now wait for their matching server release, so newly announced features work when they reach your device.

## Test plan

- [x] `vp test run --reporter=agent` on the focused workflow/changelog files
- [x] full scripts suite: 39 files, 493 tests passed
- [x] mutation check: deleting the iOS source-map upload step turns two tests red
- [x] executable classifier test — the new deploy-only helper and workflow tests no longer mark `WEB_CHANGED`, while `packages/web/app/page.tsx` still does
- [x] `vp check` on all changed files
- [x] `vp run typecheck` (full monorepo)
- [x] parse all three changed workflows as YAML
- [x] independent adversarial frozen-diff review
- [x] local orchestrator found `.boardsesh/qa-notes.md`, reached backend health and Next readiness; the web process then hit the known cross-worktree Turbopack symlink panic
- [ ] production workflow dispatch, intentionally not run during PR validation

## Operational follow-up

Move `POSTHOG_PERSONAL_API_KEY` from the Production environment variable to a masked Production secret to activate the non-blocking OTA health read. Until then, publishing remains gated correctly and health reports `unavailable`.


marcodejongh:
Follow-up on the automated review:

1. `queue: max` is now documented by GitHub Actions. It queues up to 100 waiters and processes them in wait-start order: https://docs.github.com/en/actions/how-tos/write-workflows/choose-when-workflows-run/control-workflow-concurrency. The stale-target guard is still intentional: a queued push can be behind newer source on `main` by the time its FIFO turn starts, so the intermediate SHA is coalesced without production mutation.
2. `scripts/production-deploy-baseline.test.ts` is a new 420-line dedicated suite (19 test cases, including real temporary-git histories). It covers latest-completion selection, missing/ambiguous history, failed/rollback-held/no-marker predecessors, generated-commit identity/path checks, stale targets, queued changelog snapshots, and the `changelog-snapshot` command path. Together with the other three focused files, 89 executable contracts pass.
3. The comment immediately above `deploy-app-web` documents the independence: the standalone Expo export is self-contained and deliberately does not depend on a successful Next build. It still waits for required migrations and any changed backend to reach verified success. Native OTA separately requires the changed Next build and deploy to succeed, so a Next failure cannot publish OTA.

No code change is needed for these three findings.

marcodejongh:
## Fable review — addressed

**Critical — branch was 129 commits behind with content conflicts.** Rebased onto `main` and resolved both conflicts by hand so the reusable callee keeps everything `main` landed since the branch point:

- **#4134 source-map upload** — `upload_sourcemaps_ios` / `upload_sourcemaps_android` stay between the two publishes (each `eoas publish` recreates `packages/mobile/dist`, so the iOS upload can't move after the Android publish), with their `SENTRY_AUTH_TOKEN` wiring and `timeout-minutes: 10`. `SENTRY_AUTH_TOKEN` is a Production *environment* secret, so it resolves in the called job — no caller forwarding needed, same as `GOOGLE_MAPS_API_KEY`.
- **#4148 fingerprint stabilization** — the shared build-env block is byte-identical to `main`, including `EXPO_PUBLIC_SNAPSHOT_BASE_URL: …/v1-gzip`.
- **Widened retry budget** — `timeout-minutes: 150` kept.

For the source-map failure signalling I took the second option you offered: the callee's Discord steps are gone as designed, and `Record OTA result` now returns a `sourcemaps_status` output (`not-run` / `ok` / `missing`) that `notify-failure` prints. A published-but-unsymbolicated bundle therefore reads differently from nothing published, in the one truthful release notification. The in-log `::warning::` and the terminal red gate from `main` are kept; the gate is now the job's last step so the caller's outputs land before the failure propagates.

**Major — tests didn't pin the source-map steps.** Added to `scripts/mobile-ota-publish-workflow.test.ts`:

- the upload steps' ids, `if:` conditions, `SENTRY_AUTH_TOKEN` env, and `vp run mobile:upload-sourcemaps -- --platform ios|android` commands
- their ordering relative to both publishes (publish iOS → upload iOS → publish Android → upload Android)
- `environment: Production` on the publish job, since without it the token silently becomes an empty string
- the terminal gate's condition, `exit 1`, and that it is the file's last step
- six `sourcemaps_status` cases executed against the recorded bash, including a deleted upload step (empty outcome → `missing`)
- the parent's `OTA_SOURCEMAPS_STATUS` wiring

Mutation-checked: deleting the iOS upload step turns two of these red.

**Minor — dead `state-note`.** Removed.

**Minor — deploy-only scripts triggered a web deploy.** `scripts/production-deploy-baseline.mjs`, the three workflow test files, `scripts/__tests__/changelog-transform.test.ts`, and `scripts/mobile-upload-sourcemaps.ts` are now in the web exclusion list (the uploader also joins the OTA lane, on the same terms as `scripts/mobile-publish.ts`). Rather than asserting on the text, the test now extracts the workflow's own `case` classifier and runs it in bash — the patterns span backslash-continued lines, so only executing them proves anything. Two control cases (`packages/web/app/page.tsx`, `scripts/refresh-recommendations.ts`) guard against the patterns swallowing real web files.

PR body refreshed: the `#4136` stacking note is gone.

Validation: `vp check` clean on all changed files, `vp run typecheck` green across the monorepo, `vp test run --reporter=agent scripts/` → 39 files / 493 tests passed, all three workflows parse as YAML.

marcodejongh:

marcodejongh:

Conflicting: .github/workflows/mobile-ota-production.yml, .github/workflows/production-deploy.yml, AGENTS.md, CLAUDE.md, scripts/mobile-ota-publish-workflow.test.ts — the deploy/OTA process moved independently on main.

PR #4148, opened 2026-08-01
fix(mobile): stabilize OTA native fingerprints

### LINKED ISSUE #4122: OTA fingerprint: Bun peer-hash suffixes cause false native-change flags; native dirs and patches not hashed
Diagnosed while landing PRs #4098 and #4111, both of which tripped the ota-check "Native change detected" comment on JS/devDep-only diffs. Reproduced CI's fingerprint computation locally on both branches and root-caused it. Zero native content changed in either PR — verified by normalizing store paths and comparing all 67 native dir hashes (0 differences).

## Root cause

`@expo/fingerprint` (via `expo-updates runtimeversion:resolve`, called by `scripts/mobile-ota-compat-check.ts`) hashes the serialized Expo/RN-core autolinking configs. Those configs embed absolute-ish store paths from Bun's isolated linker, which include a **peer-resolution hash suffix**:

```
node_modules/.bun/expo@57.0.9+19de8052247416b2/node_modules/expo   (PR)
node_modules/.bun/expo@57.0.9+5d294320467232ea/node_modules/expo   (main)
```

Any dep change that alters a peer set reachable from the Expo module graph reshuffles these suffixes for every native module → `contents:expoAutolinkingConfig:*` and `contents:rncoreAutolinkingConfig:*` change → fingerprint (= `runtimeVersion`) moves.

- #4111: `typescript` in `packages/mobile` devDeps peers into `@expo/require-utils` (a dep of `expo-modules-autolinking`) → moved it.
- #4098: the ten JS-only mobile bumps did the same.
- A control experiment (`semver` added to mobile devDeps) does **not** move it — whether a dep is "safe" is invisible from the diff; only running the check tells.

The two config sources are the **only** contributing sources that changed in both PRs (25 of 89 sources contribute; full lists reproduced locally match CI's hashes exactly).

Note this is a false positive for "native change" but the OTA breakage is real: `expo-updates` resolves `runtimeVersion` from the same hash, so updates published from post-#4098 main genuinely won't reach binaries built before it.

## Verified fix (recommendation)

`packages/mobile/fingerprint.config.js` with a `fileHookTransform` that strips the `+<peerhash>` segment from the four autolinking config sources before hashing:

```js
const BUN_PEER_HASH = /(\/\.bun\/[^/"]+?@[^+/"]+)\+[0-9a-f]+\//g;
const AUTOLINKING_IDS = new Set([
  'expoAutolinkingConfig:ios', 'expoAutolinkingConfig:android',
  'rncoreAutolinkingConfig:ios', 'rncoreAutolinkingConfig:android',
]);
module.exports = {
  fileHookTransform: (source, chunk) =>
    source.type === 'contents' && AUTOLINKING_IDS.has(source.id) && typeof chunk === 'string'
      ? chunk.replace(BUN_PEER_HASH, '$1/')
      : chunk,
};
```

Verified: with this file in both trees, the TS7 branch and main produce **identical** fingerprints (`1f11881e…`), while `name@version` stays in the hash so real native-module version bumps are still detected.

Caveats:

- Adding the config moves the fingerprint once → must land via a native train, with a store build after.
- `fingerprint.config.js` itself is not a hashed source; edits to it silently move the fingerprint.
- Alternatives considered and rejected: `SourceSkips` has no autolinking flag; `.fingerprintignore` can't reach inside `contents` sources; switching Bun to the hoisted linker fixes the paths but resurrects phantom-dep resolution that `check:mobile-native-deps` exists to prevent; normalizing only inside `mobile-ota-compat-check.ts` would make the checker lie (expo-updates still resolves the raw hash as the real runtimeVersion).

## Two adjacent coverage gaps (worse than the false positive)

1. **Native module source dirs are not fingerprinted at all under the isolated linker.** All 64 autolinked `dir:` sources hash to `null` — they're excluded by `@expo/fingerprint`'s default `**/node_modules/**/node_modules/**` ignore, which matches Bun's nested `.bun/<pkg>/node_modules/<pkg>` layout. Only `name@version` in the autolinking JSON is covered, so a native-code change at an unchanged version (e.g. a patched module) would ride an OTA undetected.
2. **`patchedDependencies` are not fingerprinted.** The fingerprinter looks for `packages/mobile/patches/`; ours live at the repo root. All three current patches (`@expo/ui`, `expo-dev-launcher`, `react-native-screens`) touch native iOS code. Editing a patch body without changing the target version ships as "JS-only". A `fingerprint.config.js` `extraSources` entry pointing at the root `patches/` directory would close this.

`docs/mobile-ota-updates.md` ("hashes the resolved Expo config and native files") should be updated to describe the actual behaviour under Bun's isolated layout once a fix direction is chosen.

marcodejongh:
## Summary

- normalize only Bun peer-resolution suffixes in the four native autolinking fingerprint sources while preserving package names and valid SemVer
- make direct native module directories, the fingerprint configuration, and root patch bodies contribute to both platform fingerprints
- gate every OTA/native workflow on the complete input set, enforce it in CI, and document the one-time native-train boundary

Closes #4122

## Release Notes

Keep receiving over-the-air fixes after JavaScript-only dependency updates.

## Test plan

- independent adversarial implementation review and focused re-review: pass
- `vp test run scripts/mobile-fingerprint-config.test.ts scripts/mobile-fingerprint-inputs-check.test.ts scripts/mobile-ci-env-parity.test.ts` — 48 tests pass after rebase
- `vp node scripts/mobile-fingerprint-inputs-check.ts` — 67 iOS and 68 Android native directories plus config and patches are hashed
- `vp node scripts/mobile-patches-check.ts` — all six native patches are applied
- `vp run typecheck:mobile`
- `vp run test:mobile` — 575 files / 4,912 tests pass
- `vp run check:mobile-bundle` — iOS and Android bundles pass
- touched-file `vp check`, `git diff --check`, and the mandatory pre-commit hook

This intentionally changes the native fingerprint once. Matching iOS and Android store builds must ship before OTAs produced from the new fingerprint can reach users.

### DIFF
--- a/.github/actions/mobile-native-gate/action.yml
+++ b/.github/actions/mobile-native-gate/action.yml
@@ -70,8 +70,9 @@ runs:
         # Path screen: nothing in the canonical native-input set (mirrors the
         # ios-rn-ci Pods cache key) changed → skip without paying for a resolve.
         if [ -z "$(changed \
-              packages/mobile/package.json bun.lock packages/mobile/app.config.ts \
-              packages/mobile/plugins packages/mobile/modules patches)" ]; then
+              package.json packages/mobile/package.json bun.lock packages/mobile/app.config.ts \
+              packages/mobile/fingerprint.config.js packages/mobile/plugins \
+              packages/mobile/modules patches)" ]; then
           skip "no native-input files changed vs main"
         fi
 
--- a/.github/workflows/android-apk-rn.yml
+++ b/.github/workflows/android-apk-rn.yml
@@ -15,6 +15,8 @@ on:
       - 'packages/shared/**'
       - 'packages/board-constants/**'
       - 'packages/shared-schema/**'
+      - 'package.json' # root patchedDependencies and linker inputs affect the mobile native graph
+      - 'bun.lock'
       - 'patches/**' # bun patches to mobile deps live in root patches/, not under packages/mobile/**
       - '.github/workflows/android-apk-rn.yml'
   workflow_dispatch:
--- a/.github/workflows/android-pr-rn.yml
+++ b/.github/workflows/android-pr-rn.yml
@@ -32,6 +32,8 @@ on:
       - 'packages/shared/**'
       - 'packages/board-constants/**'
       - 'packages/shared-schema/**'
+      - 'package.json' # root patchedDependencies and linker inputs affect the mobile native graph
+      - 'bun.lock'
       - 'patches/**' # bun patches to mobile deps live in root patches/, not under packages/mobile/**
       - '.github/workflows/android-pr-rn.yml'
       - '.github/actions/android-rn-setup/**'
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -81,6 +81,7 @@ jobs:
             mobile:
               - 'packages/mobile/**'
               - 'packages/board-renderer/wasm/pkg/**'
+              - 'patches/**'
               - 'scripts/mobile-platform-imports-check.sh'
               - 'scripts/mobile-web-bundle-check.sh'
               - 'scripts/sync-mobile-board-renderer-wasm.sh'
@@ -749,6 +750,11 @@ jobs:
       # patched source is actually present on this cheap Linux runner.
       - name: Check native patches are applied
         run: vp run check:mobile-patches
+      # The resolver can return a valid fingerprint even when a source is silently
+      # absent/null. Assert the four autolinking configs, every discovered native
+      # module dir, the fingerprint config, and root patch bodies all contribute.
+      - name: Check mobile fingerprint inputs are complete
+        run: vp run check:mobile-fingerprint-inputs
       # Cheap source guard: no raw theme-variant magic-string compares regrew in
       # mobile components (they must route through the variant primitives/tokens —
       # see packages/mobile/src/theme/variants/README.md).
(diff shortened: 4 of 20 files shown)


marcodejongh:
Follow-up hardening is complete in `9a6979a3e`.

- Scoped Bun store names without a `+` delimiter now fail closed, with a direct regression.
- Resolver stdout parsing accepts either property order, recovers from malformed diagnostic braces/quotes across records, rejects ambiguous complete payloads, and caps cumulative parse work at 2x stdout.
- The CI parity test now proves `packages/mobile/**` covers `fingerprint.config.js`.

Independent adversarial review reproduced the earlier failures and passed the final implementation, including nested-noise complexity checks. Focused tests pass 58/58; touched `vp check`, the live 67-iOS/68-Android audit, and diff checks are green.

marcodejongh:
Fresh review is green. For the remaining nonblocking notes:

- `bun.lock` is intentionally a production native-build input because a lockfile-only dependency resolution can change the autolinked native graph; over-triggering is safer than silently shipping an incompatible OTA.
- The scanner does handle JSON unicode escapes: after `\\`, it consumes the `u` as the escaped code-unit prefix and the following hex digits remain ordinary in-string characters; only the real closing quote ends the token.
- The 16-hex peer token is stripped only inside a valid SemVer Bun store suffix and is covered by scoped/unscoped/fail-closed tests. This matches Bun's linker encoding; changing that heuristic without an authoritative delimiter would reintroduce cross-machine drift.
- The exact `@expo/fingerprint@0.20.6` pin is required by the patching mechanism. Upgrade sentinels and the native-input audit deliberately force maintenance instead of silently dropping native directories.

PR #4150, opened 2026-08-01
fix(mobile): explain offline download fallback

### LINKED ISSUE #4116: My Boards download UI gives no signal when snapshot bootstrap fails and falls back to paged crawl
## Summary

Follow-up from #4106. When offline snapshot bootstrap exhausts its retries for a board scope, the board still finishes downloading — it just falls back to the ordinary paged GraphQL crawl (`MAX_BOOTSTRAP_ATTEMPTS` = 2 in `packages/shared/offline-sync/src/sync/pull-client.ts`) — but the My Boards screen gives the user zero indication anything went wrong. A failed bootstrap looks identical to a healthy, still-in-progress download.

Confirmed while investigating #4106:

- `BoardDownloadState` (`packages/mobile/src/components/board-discovery/board-offline-state.ts:7-11`) is a closed union: `'off' | 'pending' | 'downloading' | 'downloaded'`. There is no `'failed'` member.
- The My Boards row (`packages/mobile/app/boards/manage.tsx`, `BoardManageRow.tsx:113-122`, `BoardOfflineToggle.tsx:24-48`) only distinguishes "Downloading board…" (bootstrap) from "Downloading N climbs" (paged) — once bootstrap gives up, the caption just quietly switches from one to the other.
- While an attempt is burned but attempts remain, that cycle's paged pull is skipped entirely (by design — see the doc comment in `pull-client.ts` around `runBootstrapPhase`), so the row drops back to `'pending'` ("Waiting to download") — indistinguishable from a board that was just enabled and hasn't started yet.
- i18n: `packages/shared/i18n/locales/en-US/boards.json`'s `offline` block (lines 263-279) has `pending`/`bootstrapping`/`downloadingCount_*`/`available` keys but no `failed`/`error`/`retry` key in any locale.
- There is no hard "failed" state and no retry button anywhere in this UI. The only visibility into a bootstrap failure today is Sentry (`kind: 'snapshot-bootstrap'`) and the PostHog `method: 'snapshot' | 'paged'` split on `Offline Board Download Completed` — nothing user-facing.

## Why this is a fast-follow, not part of #4106

The user isn't actually stuck — the paged crawl auto-resumes and completes the download without any action on their part, just slower and with the correlated-`EXISTS` cost noted in `docs/board-snapshots.md`'s "Deferred" section. So this is a *visibility* gap, not a *broken* flow. #4106's fix (making the download-stage failure cause reach Sentry) is the higher-value, lower-risk piece and ships on its own.

**Exposure context:** the whole offline-download engine (`offline-board-downloads` PostHog flag) is at **0% rollout** as of this writing — this UI gap has essentially no live user impact today. Worth fixing before any percentage ramp (see the Rollout plan in `docs/board-snapshots.md`), but not urgent on its own clock.

## Suggested direction

- Add a `'failed'` (or similarly named) state to `BoardDownloadState`, distinct from `'pending'`/`'downloading'`, set when a scope's bootstrap attempts are exhausted (or `SyncProgress` reports a cycle-level `failed` some other way) and paged pull for that cycle didn't advance.
- Surface it in `BoardManageRow.tsx` with distinct copy (new i18n keys in all three locales: `en-US`, `es`, `fr`) — something like "Taking longer than usual" rather than a hard "Failed", since the board is still downloading, just slowly.
- Decide whether a manual retry action is meaningful given the engine already retries automatically on the next sync cycle — a retry button may just be reassurance copy rather than a functionally different action.

## Related

- #4106 — the Sentry-observability fix this issue follows up on
- `docs/board-snapshots.md` — "Client bootstrap flow" (failure/attempt matrix) and "Mobile wiring" sections

marcodejongh:
## Summary

- surface per-board snapshot retry and paged-fallback state in My Boards, including exact live progress through climbs, stats, and grades
- invalidate metadata as each scope settles so one completed board cannot show stale retry copy while another board continues downloading
- announce semantic status transitions once on iOS VoiceOver and Android TalkBack while letting the full localized explanation wrap on screen
- carry per-scope settlement through scheduled bootstrap, bridge, and ad-hoc download paths, preserving fallback state across aborted retries

Closes #4116

## Release Notes

See when an offline board is retrying its fast download or falling back to the slower sync.

## Test plan

- independent adversarial QA and repaired re-review: pass
- focused repair/catalog suite — 10 files / 213 tests pass
- full mobile suite — 576 files / 4,946 tests pass
- `vp run typecheck:mobile` and `vp run typecheck:shared`
- scoped `vp check`, catalog completeness, `vp run check:i18n`, and `vp run check:i18n:orphans`
- `vp run check:mobile-bundle` — iOS and Android pass
- post-rebase changed-test suite — 11 files / 176 tests pass
- `git diff --check` and mandatory pre-commit hook
- `vp run dev:mobile` with `.boardsesh/qa-notes.md`; Metro ready at the reported QA URL


marcodejongh:
CI and review follow-up is complete in `9ae1e1b80`.

- Both full-module checkpoint mocks now include `SCOPE_COMPLETE_PREFIX`, fixing the two Actions suite-load failures.
- Returned bootstrap metadata is readonly while construction remains package-private and mutable.
- Scope teardown tests explicitly lock in the grades checkpoint alongside climbs/stats, and the docs now distinguish the two snapshot-backed tables from the three-table paged completion path.
- The source-generation logic remains unchanged after independent review confirmed it is a deliberate monotonic cache-busting nonce.

Independent validation passed: offline-sync 66/66, snapshot bootstrap 40/40, manage screen 15/15, offline-sync/shared/mobile typechecks, touched checks, and diff checks.

marcodejongh:
Latest review rechecked against head `9ae1e1b80`:

- The import-time `SnapshotSchemaStaleError` branch already calls `markBootstrapPagedFallback(db, scope.scopeKey)` before setting `metadataSettled` and continuing (`pull-client.ts`, in the import catch). It therefore persists `paged-fallback`, not unchanged state.
- `manage-offline.test.tsx` already exercises the source-transition generation guard in “suppresses stale retry data while flag-off paging is unresolved, then shows the fresh fallback”: it renders retry metadata, toggles the source off and back on without either sync revision advancing, holds the new SQLite read pending, proves the stale notice stays hidden, then resolves the fresh generation and proves the fallback appears.
- The two local `deferred` helpers have different contracts and there is no existing shared test utility for them; moving them would add unrelated churn.

No follow-up code change is needed for these nonblocking notes.

PR #4152, opened 2026-08-01
fix(mobile): serialize same-board iOS BLE reconnects

### LINKED ISSUE #4139: iOS BLE: late disconnect callback can tear down an immediate same-board reconnect

marcodejongh:
## Summary

- place every manager-initiated CoreBluetooth cancellation behind a per-peripheral barrier
- defer only same-UUID reconnects until the old cancellation callback settles, while different boards still connect immediately
- make the newest deferred request win and safely complete superseded callers
- consume disconnect/fail callbacks before generation handling, reset the barrier on BLE-off, and preserve write-stall recovery
- deliberate behavior change beyond the title: when the central drops below `.poweredOn`, an in-flight connect now fails immediately with `bluetoothUnavailable` and peripheral generations are cleared, instead of letting the 8 s connect timeout own the failure as on `main`
- an expired cancellation barrier (watchdog fired with no terminal callback, which CoreBluetooth legitimately never delivers for a cancelled pending connect) no longer blocks later connects: a fresh explicit connect displaces it and proceeds, so an out-of-range connect timeout can't leave the board unconnectable for the rest of the session
- add deterministic native coverage for immediate reconnects, late callbacks, watchdogs, config drift, and write-stall composition

Closes #4139.

## Stack

This PR is intentionally based on #4140 (`fix/3925-ios-moonboard-ble-tests`) because it extends that native disconnect/write-stall test seam. The branch DAG now uses the actual #4140 commits; the PR diff is only the three #4139 Swift files. Rebase onto `main` after #4140 merges.

## QA

- independent Sol xhigh state-machine review: PASS
- `vp run typecheck:mobile`
- mobile suite: 575 files, 4,912 tests passed
- focused native-preparation/widget drift tests passed
- `vp run check:mobile-bundle` for iOS and Android
- touched pre-commit checks and `git diff --no-ext-diff --check`
- Metro QA notes: http://vibetunnel-2.tailedd9d5.ts.net:8086

## Native verification gates

Linux cannot compile or execute `BoardseshTests`. Before merge, run the native XCTest target on macOS and exercise explicit disconnect → immediate same-board reconnect, different-board reconnect, BLE-off reset, write-stall recovery, and out-of-range connect timeout → walk back into range → retry connects normally, on a physical iPhone. Fable is unavailable in this runtime and remains a literal BLE review blocker; the Sol review is recorded as a substitute, not as Fable.

## Release Notes

Reconnecting to the same board on iPhone no longer lets a late disconnect from the old connection knock the new link offline.

### DIFF
--- a/packages/mobile/ios-tests/BoardBleDisconnectTests.swift
+++ b/packages/mobile/ios-tests/BoardBleDisconnectTests.swift
@@ -9,12 +9,14 @@ final class BoardBleDisconnectTests: XCTestCase {
     private var scheduler: FakeBleTimerScheduler!
     private var manager: BoardBleManager!
     private var cancelledPeripheralIds: [UUID] = []
+    private var connectedPeripheralIds: [UUID] = []
     private var stopScanCallCount = 0
     private let uartWriteCharacteristicUuid = CBUUID(string: "6E400002-B5A3-F393-E0A9-E50E24DCCA9E")
 
     override func setUp() {
         super.setUp()
         cancelledPeripheralIds = []
+        connectedPeripheralIds = []
         stopScanCallCount = 0
         scheduler = FakeBleTimerScheduler()
         manager = BoardBleManager(timerScheduler: scheduler, createCentralManagerEagerly: false)
@@ -26,6 +28,10 @@ final class BoardBleDisconnectTests: XCTestCase {
             manager.testHooks.setCancelPeripheralConnectionOverride { [weak self] peripheral in
                 self?.cancelledPeripheralIds.append(peripheral.identifier)
             }
+            manager.testHooks.setConnectPeripheralOverride { [weak self] peripheral in
+                self?.connectedPeripheralIds.append(peripheral.identifier)
+            }
+            manager.testHooks.setCentralState(.poweredOn)
         }
     }
 
@@ -33,6 +39,8 @@ final class BoardBleDisconnectTests: XCTestCase {
         manager.testHooks.sync {
             manager.testHooks.setStopScanOverride(nil)
             manager.testHooks.setCancelPeripheralConnectionOverride(nil)
+            manager.testHooks.setConnectPeripheralOverride(nil)
+            manager.testHooks.setCentralState(nil)
         }
         manager = nil
         scheduler = nil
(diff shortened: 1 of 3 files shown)


marcodejongh:
Addressed the two actionable test observations in `15e7caa76`:

- `fireCentralStateUpdate` now restores the previous test-only central-state override with `defer`, so one callback simulation cannot leak state into later behavior;
- a new native case expires the cancellation-barrier watchdog, then powers Bluetooth off, proving the waiter settles once as `connectTimedOut`, the stale barrier/generation/deferred state clears, and the watchdog is cancelled.

The preparation drift test, mobile typecheck, pre-commit checks, and diff check pass. The production state machine is unchanged. Fable, macOS XCTest, physical CoreBluetooth, and the #4140 base remain the draft gates.

marcodejongh:
Also removed the misleading `@discardableResult` in `d269789d5`; both call sites consume the boolean. I audited every `discoveredPeripherals` read after its testability type widening: there are no force casts, and the only three concrete `CBPeripheral` conversions are guarded inside cancel/prepare/connect boundaries.

marcodejongh:
The confirmed post-merge displaced-`didFailToConnect` race is fixed in draft follow-up #4173. The patch conservatively handles the UUID-only callback ambiguity, adds exact single-settlement regressions, and passed independent adversarial QA. Fable, macOS XCTest, and physical CoreBluetooth verification remain explicit gates.

PR #4153, opened 2026-08-01
fix(mobile): retry transient Android BLE connects

### LINKED ISSUE #4143: Board connect gets one shot — a transient GATT 133 ends the attempt instead of retrying
Tap connect, the board doesn't take, and you get an error dialog. Tap again and it usually works — because the first failure was a transient GATT hiccup that nothing in the app retries. Every connect attempt is one shot.

## What the data shows

30 days (2026-07-02 → 2026-08-01), genuine technical failures with `user_cancelled` stripped out:

| Reason | Events | Users |
|---|---|---|
| `connect_failed` | 240 | 124 |
| `board_not_found` | 166 | 95 |
| `disconnected` | 179 | 86 |
| `write_failed` | 111 | 16 |

Android `connect_failed` carries `androidErrorCode: 133` — the textbook transient GATT error that a single retry clears.

Sessions: [019faf55](https://us.posthog.com/project/412845/replay/019faf55-6cd1-7b59-b622-a8cbff9a1a13), [019f2943](https://us.posthog.com/project/412845/replay/019f2943-1b01-7ecf-bec8-d5a23f07564f), [019fa34a](https://us.posthog.com/project/412845/replay/019fa34a-c1f1-7222-a750-94b95293b42d)

**Read the rate carefully.** The raw numbers say 33% of Android connect attempts fail — that figure includes `user_cancelled` (picker dismissals), which is 77% of all `Bluetooth Connection Failed` volume fleet-wide. #3608 measured 9.5% *after* stripping cancels from both sides. Anyone re-deriving impact here must strip `failureReason = 'user_cancelled'` first; there is no evidence of escalation since #3608, and the real `connect_failed` bucket is 131 Android events against 3,832 successes.

## Root cause (verified on `main`)

There is no retry anywhere on the connect path.

- **Android** — `packages/mobile/src/lib/ble/adapter.ts:249` does exactly one `bleManager.connectToDevice(selectedDeviceId)` raced against a 12s `CONNECTION_TIMEOUT_MS`. A single throw propagates straight to an `Alert`. Nothing inspects `androidErrorCode`, so GATT 133 is terminal on the first tap. It's captured at `:444` for reporting only.
- **iOS** — `packages/mobile/modules/live-activity/ios/BoardBleManager.swift:679-697` does one `centralManager.connect(peripheral, ...)` behind an 8s watchdog; on expiry it cancels and settles `BoardBleError.connectTimedOut`. `didFailToConnect` (~:1154) is documented "No auto-reconnect" and settles with no retry.
- `grep -riE 'retry|backoff'` over `packages/mobile/src/lib/ble` returns only comment prose plus an unrelated write-path fallback.

These are technical failures only — `user_cancelled` returns early at `use-board-bluetooth.ts:1176` and shows no alert. Their absence from `$exception` is deliberate (`bleConnectReportLevel()` returns `warning` for `connect_failed`/`board_not_found`), so telemetry is the only signal here by design.

## Suggested scope

1. **Android first** (TS-only, OTA-safe — do this as its own PR): bounded retry with backoff around the `connectToDevice` race in `adapter.ts`, gated on transient codes (`androidErrorCode` 133/147, `bleErrorCode` 200/201/205). 1–2 retries at ~500 ms–1 s, with `cancelDeviceConnection` between attempts to clear the stale GATT handle.
2. **Re-land the reverted iOS attribution** from `3150df3de` (reverted in `744af4a88`, never re-landed — `git grep connectFailureReason origin/main` is empty) so the iOS half is diagnosable. Do this **before** adding an iOS retry: with zero sub-reason data an iOS retry is speculative and risks fighting the deliberate last-connection-wins / no-auto-reconnect policy from #2720.
3. Emit a `retry_succeeded` property so the fix is measurable.

## Severity

**P2.** Genuine core-flow friction for 124 users in-window, but every failure is recoverable by tapping connect again — no crash, no data loss. That matches the repo's P2 definition and how both sibling issues for this exact phenomenon were triaged.

## Process

Touches `packages/mobile/src/lib/ble` and `packages/shared/ble-protocol` → **requires a Fable review before merge** per CLAUDE.md. Any Swift change is a native-fingerprint input and must go via a `[native-train]` draft; the Android-only TS retry is OTA-safe and should land first.

## Related

- #3608 (CLOSED) — the direct predecessor. It landed only the *classification* fix in `packages/shared/ble-protocol/src/connection-error.ts` (ble-plx 200/201/205/206 → `connect_failed`, `androidErrorCode` attached) — which is why this failure mode is visible at all. Its body explicitly deferred the recovery work: *"Re-measure, then decide whether Android needs the connect-side recovery machinery iOS already has."* This is that re-measure; the recovery was never built.
- #3676 (CLOSED) — iOS sibling. Named "no connect retry" as a hypothesis but shipped only sub-reason telemetry, which was reverted.
- #3088 (OPEN, P3) — BLE error-tracking hygiene/attribution, adjacent.
- #3924, #3982, #3991 (OPEN) — distinct BLE concerns (disconnect duration telemetry, Android ≤11 scan-with-Location-off, firmware supervision timeout).

marcodejongh:
## Summary

`RNBleAdapter.requestAndConnect` made exactly one GATT connect attempt, so Android's canonical transient handshake failure — GATT status 133 — was terminal on the first tap. This adds one bounded retry.

- retry the **already-selected** device id once after a 500 ms backoff when ble-plx reports an eligible connect error with GATT 133/147 or no native status — no rescan, no second picker
- one absolute 12-second budget shared across the first attempt, stale-handle cleanup, backoff, retry, and exhausted cleanup, so a dead board fails no later than it does today
- terminal failures are rethrown by identity: iOS-shaped, malformed, picker-cancel, permission, discovery, MTU
- gated on `Platform.OS === 'android'` at the factory; iOS and the Expo browser target keep the single-shot path
- emits `retrySucceeded` and `connectAttempts` on both the success and failure events so the retry has a measurable hit rate

Closes #4143.

## Why this allowlist

Measured against the fleet (PostHog, `Bluetooth Connection Failed`, 90 days, grouped by `failureReason` × `bleErrorCode` × `androidErrorCode`):

| bleErrorCode | androidErrorCode | events | users |
| --- | --- | --- | --- |
| 201 | 133 | 360 | 138 |
| 201 | null | 38 | 15 |
| 205 | null | 32 | 15 |
| 201 | 147 | 3 | 3 |
| 3 (OperationTimedOut) | null | 2 | 1 |
| null (iOS / web) | null | 317 | 170 |

433 of the 752 `connect_failed` events — effectively every Android one — fall inside the `{200, 201, 205} × {133, 147, null}` allowlist, and the one ineligible bucket (OperationTimedOut) is correctly excluded. That table is also the pre-fix baseline the new `retrySucceeded` / `connectAttempts` split reads against. Note 1,615 of the 8,778 90-day failure events carry a null `failureReason` (older builds predating that property), so scope any before/after read to events that carry it.

## Review comments addressed

1. **Exhausted cleanup blocked the failure alert** (`adapter.ts`) — the final `cancelDeviceConnection` is now fire-and-forget, same as the first-attempt deadline branch. We already have the terminal error at that point, so nothing should hold the alert back. Its test now advances 500 ms instead of 12 s and asserts zero pending timers.
2. **`retry_succeeded` naming + missing denominator** (`use-board-bluetooth.ts`) — renamed to `retrySucceeded` to match every sibling property in that payload, and both the success and failure events now carry `connectAttempts`. Domain is **0/1/2**, not 1/2: `createBluetoothAdapter` runs per connect, so anything that never reaches the GATT connect (picker cancel, board not found, scan error) reports 0, and that is the dominant bucket. 1 = a single attempt (ineligible error, or the first failure landed too late for the backoff to fit the budget); 2 + success = the retry saved it; 2 + failure = the retry ran and lost. The key is dropped on native iOS and web. This is a getter on the adapter rather than a field on `BleConnection` as the comment suggested — the failure path only ever sees a thrown error and has no connection object to carry it.
3. **Picker-cancel telemetry gate** — reverted; `Bluetooth Connection Failed` fires unconditionally again. One half of the original rationale is stale (`packages/web/app/components/board-bluetooth-control/` no longer exists on `main`, and the only emitters left are in the mobile hook), but the out-of-scope half stands on its own: `user_cancelled` is 5,380 of ~8,778 of that event's 90-day volume, so the gate would delete the only measure of picker-dismissal rate. Worth its own issue.
4. **Wall-clock deadline** (`adapter.ts`) — all three `Date.now()` reads swapped for `performance.now()`. To be precise about the guarantee: RN 0.86's `setUpPerformance` falls back to `global.nativePerformanceNow || Date.now` when `NativePerformance` is absent, so this is monotonic on every real Hermes runtime and never worse than today. There is a comment at the deadline assignment noting the arithmetic depends on Vitest faking `performance.now` — it does by default, and an explicit `toFake` list omitting `performance` would silently freeze the clock and pass the boundary tests for the wrong reason.

The rebase onto `main` resolved one conflicted file (`use-board-bluetooth.test.ts`), keeping both `main`'s new config-switch attribution test and this branch's retry assertions.

## QA

- `vp run typecheck:mobile`, `vp run typecheck:ble-protocol`
- `vp run test:mobile`, `vp test run --project ble-protocol`
- `vp check`
- boundary stress at 11.5 s / 11.75 s / 11.999 s proves no second attempt begins and no timer leaks

## Still owed before merge

- **Fable BLE review** — CLAUDE.md hard rule for anything touching `packages/mobile/src/lib/ble` or `packages/shared/ble-protocol`. The existing Sol review is not a substitute.
- **Real-device Android QA** — reproduce ble-plx 133 / 147 / null-status failures against an actual board and confirm recovery, plus timeout, double-tap, picker dismissal, config switching, two-phone takeover. Two things no unit test can see: a recovered connect must not reopen the picker, and must not double-fire the success haptic.

TS-only, so it ships over the air — no new build needed.

## Release Notes

Android now retries a transient board connection once, so a brief GATT hiccup is less likely to send you back to the connect button.

### DIFF
--- a/packages/mobile/src/lib/ble/adapter.ts
+++ b/packages/mobile/src/lib/ble/adapter.ts
@@ -148,6 +148,12 @@ export class RNBleAdapter implements BluetoothAdapter {
   // reached the GATT connect at all (picker cancelled, board not found, scan
   // error), 1 means a single attempt, and 2 means the retry ran.
   private lastConnectAttemptCount = 0;
+  // Whether the retry of the most recent connect won the GATT connect. Read on
+  // the failure path too, where the thrown error carries no result object: a
+  // recovered GATT connect can still fail later at MTU negotiation or service
+  // discovery, and `lastConnectAttemptCount === 2` alone can't tell that apart
+  // from a retry that lost.
+  private lastConnectRetrySucceeded = false;
 
   constructor(
     private readonly devicePicker: DevicePickerFn,
@@ -170,6 +176,7 @@ export class RNBleAdapter implements BluetoothAdapter {
     // silently freeze the clock and pass for the wrong reason.
     const deadlineMs = performance.now() + CONNECTION_TIMEOUT_MS;
     this.lastConnectAttemptCount = 0;
+    this.lastConnectRetrySucceeded = false;
     const attemptConnect = () => {
       this.lastConnectAttemptCount += 1;
       return settleBeforeDeadline(() => bleManager.connectToDevice(selectedDeviceId), deadlineMs);
@@ -207,6 +214,7 @@ export class RNBleAdapter implements BluetoothAdapter {
       throw connectionTimeoutError();
     }
     if (secondAttempt.kind === 'fulfilled') {
+      this.lastConnectRetrySucceeded = true;
       return { connected: secondAttempt.result, retrySucceeded: true };
     }
 
@@ -576,10 +584,20 @@ export class RNBleAdapter implements BluetoothAdapter {
   }
 
   // 0 when the flow never reached the GATT connect, 1 for a single attempt,
-  // 2 when the Android retry ran. Pairs with `retrySucceeded` to give the retry
-  // a denominator: 2 + success is a save, 2 + failure is a retry that lost.
+  // 2 when the retry ran. This is the retry denominator only — on its own it
+  // says nothing about the outcome, because the connect can still fail after a
+  // recovered GATT connect. Read it with getLastConnectRetrySucceeded().
   getLastConnectAttemptCount(): number {
     return this.lastConnectAttemptCount;
+  }
+
+  // True when the retry of the most recent connect won its GATT connect, even
+  // if a later stage then failed. Together with getLastConnectAttemptCount()
+  // this makes the retry hit rate unambiguous on both events:
+  // 2 + true = the retry recovered the GATT connect (a save, whether or not
+  // MTU/discovery then failed); 2 + false = the retry ran and lost.
+  getLastConnectRetrySucceeded(): boolean {
+    return this.lastConnectRetrySucceeded;
   }
 }
 
--- a/packages/mobile/src/lib/ble/adapter.ts
+++ b/packages/mobile/src/lib/ble/adapter.ts
@@ -9,6 +9,7 @@ import {
   INTER_CHUNK_DELAY_MS,
   MAX_BLUETOOTH_MESSAGE_SIZE,
   parseSerialNumber,
+  isRetryableAndroidConnectError,
 } from '@boardsesh/ble-protocol';
 import { bleManager } from './ble-manager';
 import { uint8ArrayToBase64, base64ToHex, serviceDataToHex } from './base64';
@@ -29,6 +30,7 @@ import type {
 import { SCAN_TIMEOUT_MS, SERIAL_RECONNECT_GRACE_MS } from '@boardsesh/ble-protocol/scan-constants';
 
 const CONNECTION_TIMEOUT_MS = 12_000;
+const ANDROID_CONNECT_RETRY_BACKOFF_MS = 500;
 
 // The ATT MTU requested after connect. 247 (chunk 244) is the DLE-friendly
 // sweet spot: the iOS-26.5 failure cohort clusters at ATT 512 (#3230), so
@@ -38,6 +40,69 @@ const REQUESTED_ATT_MTU = 247;
 const DEFAULT_ATT_MTU = 23;
 
 const delay = (ms: number) => new Promise<void>((resolve) => setTimeout(resolve, ms));
+
+/** Wait the full retry backoff only when it fits inside the shared deadline.
+ * Whichever timer wins clears the other, so a deadline during backoff cannot
+ * leave a 500ms timer alive after the connect sequence has already settled. */
+function waitForRetryBackoffBeforeDeadline(deadlineMs: number): Promise<boolean> {
+  const remainingMs = deadlineMs - Date.now();
+  if (remainingMs <= 0) return Promise.resolve(false);
+
+  return new Promise<boolean>((resolve) => {
+    let deadlineTimer: ReturnType<typeof setTimeout> | null = null;
+    let backoffTimer: ReturnType<typeof setTimeout> | null = null;
+    let settled = false;
+    const settle = (completedFullBackoff: boolean) => {
+      if (settled) return;
+      settled = true;
+      if (deadlineTimer !== null) clearTimeout(deadlineTimer);
+      if (backoffTimer !== null) clearTimeout(backoffTimer);
+      resolve(completedFullBackoff);
+    };
+
+    // Schedule the deadline first so an exact tie preserves the existing
+    // semantics: no second attempt begins once the shared budget is exhausted.
+    deadlineTimer = setTimeout(() => settle(false), remainingMs);
+    backoffTimer = setTimeout(() => settle(true), ANDROID_CONNECT_RETRY_BACKOFF_MS);
+  });
+}
+
+type DeadlineSettlement<T> =
+  | { kind: 'fulfilled'; result: T }
+  | { kind: 'rejected'; error: unknown }
+  | { kind: 'deadline' };
+
+/**
+ * Settle one connect-stage operation without extending the stage's shared
+ * deadline. Both fulfillment and rejection handlers stay attached after the
+ * deadline wins, so a late native promise cannot become an unhandled rejection.
+ */
+async function settleBeforeDeadline<T>(
+  operation: () => Promise<T>,
+  deadlineMs: number,
+): Promise<DeadlineSettlement<T>> {
+  const remainingMs = deadlineMs - Date.now();
+  if (remainingMs <= 0) return { kind: 'deadline' };
+
+  let deadlineTimer: ReturnType<typeof setTimeout> | null = null;
+  const operationSettlement = Promise.resolve()
+    .then(operation)
+    .then(
+      (result): DeadlineSettlement<T> => ({ kind: 'fulfilled', result }),
+      (error: unknown): DeadlineSettlement<T> => ({ kind: 'rejected', error }),
+    );
+  const deadlineSettlement = new Promise<DeadlineSettlement<T>>((resolve) => {
+    deadlineTimer = setTimeout(() => resolve({ kind: 'deadline' }), remainingMs);
+  });
+
+  return Promise.race([operationSettlement, deadlineSettlement]).finally(() => {
+    if (deadlineTimer !== null) clearTimeout(deadlineTimer);
+  });
+}
+
+function connectionTimeoutError(): Error {
+  return new Error('Connection timed out — board may be powered off');
+}
 
 /**
  * Find a write characteristic by service + characteristic UUID, returning
@@ -71,6 +136,9 @@ export class RNBleAdapter implements BluetoothAdapter {
   // Board-level demand for acknowledged writes (see BleAdapterOptions). Fixed
   // for the adapter's lifetime — the board it was built for doesn't change.
   private readonly preferWriteWithResponse: boolean;
+  // Whether a transient first GATT connect failure gets one in-budget retry
+  // (see BleAdapterOptions.enableAndroidConnectRetry) — Android only.
+  private readonly enableAndroidConnectRetry: boolean;
 
   constructor(
     private readonly devicePicker: DevicePickerFn,
@@ -78,6 +146,59 @@ export class RNBleAdapter implements BluetoothAdapter {
     options?: BleAdapterOptions,
   ) {
     this.preferWriteWithResponse = options?.preferWriteWithResponse ?? false;
+    this.enableAndroidConnectRetry = options?.enableAndroidConnectRetry ?? false;
+  }
+
+  /** Connect the already-selected peripheral, optionally recovering one known
+   * Android GATT handshake failure without rescanning or reopening the picker. */
+  private async connectSelectedDevice(
+    selectedDeviceId: string,
+  ): Promise<{ connected: Device; retrySucceeded: boolean }> {
+    const deadlineMs = Date.now() + CONNECTION_TIMEOUT_MS;
+    const attemptConnect = () => settleBeforeDeadline(() => bleManager.connectToDevice(selectedDeviceId), deadlineMs);
+    const cancelWithoutWaiting = () => {
+      void bleManager.cancelDeviceConnection(selectedDeviceId).catch(() => {});
+    };
+
+    const firstAttempt = await attemptConnect();
+    if (firstAttempt.kind === 'deadline') {
+      cancelWithoutWaiting();
+      throw connectionTimeoutError();
+    }
+    if (firstAttempt.kind === 'fulfilled') {
+      return { connected: firstAttempt.result, retrySucceeded: false };
+    }
+
+    const firstError = firstAttempt.error;
+    if (!this.enableAndroidConnectRetry || !isRetryableAndroidConnectError(firstError)) {
+      throw firstError;
+    }
+
+    // Close the failed native GATT handle before retrying. A rejection generally
+    // means it was already closed, so it must not block the retry. A hanging
+    // cleanup is bounded by the original connect deadline.
+    const cleanup = await settleBeforeDeadline(() => bleManager.cancelDeviceConnection(selectedDeviceId), deadlineMs);
+    if (cleanup.kind === 'deadline') throw firstError;
+
+    const completedBackoff = await waitForRetryBackoffBeforeDeadline(deadlineMs);
+    if (!completedBackoff) throw firstError;
+
+    const secondAttempt = await attemptConnect();
+    if (secondAttempt.kind === 'deadline') {
+      cancelWithoutWaiting();
+      throw connectionTimeoutError();
+    }
+    if (secondAttempt.kind === 'fulfilled') {
+      return { connected: secondAttempt.result, retrySucceeded: true };
+    }
+
+    const secondError = secondAttempt.error;
+    if (isRetryableAndroidConnectError(secondError)) {
+      // No third attempt. Give the exhausted handle the same bounded cleanup as
+      // the first while preserving the exact second error for telemetry.
+      await settleBeforeDeadline(() => bleManager.cancelDeviceConnection(selectedDeviceId), deadlineMs);
+    }
+    throw secondError;
   }
 
   async isAvailable(): Promise<boolean> {
@@ -251,18 +372,7 @@ export class RNBleAdapter implements BluetoothAdapter {
       }
     }
 
-    let connectionTimeoutId: ReturnType<typeof setTimeout> | null = null;
-    const connected = await Promise.race([
-      bleManager.connectToDevice(selectedDeviceId),
-      new Promise<never>((_resolve, reject) => {
-        connectionTimeoutId = setTimeout(() => {
-          bleManager.cancelDeviceConnection(selectedDeviceId).catch(() => {});
-          reject(new Error('Connection timed out — board may be powered off'));
-        }, CONNECTION_TIMEOUT_MS);
-      }),
-    ]).finally(() => {
-      if (connectionTimeoutId != null) clearTimeout(connectionTimeoutId);
-    });
+    const { connected, retrySucceeded } = await this.connectSelectedDevice(selectedDeviceId);
 
     // Negotiate MTU before service discovery (Android requires this order
     // for best results; iOS handles MTU automatically but the call is safe).
@@ -324,6 +434,7 @@ export class RNBleAdapter implements BluetoothAdapter {
       deviceName: selectedDeviceName,
       manufacturerData: selectedManufacturerData,
       serviceData: selectedServiceData,
+      retrySucceeded,
     };
   }
 
--- a/packages/mobile/src/lib/ble/adapter.ts
+++ b/packages/mobile/src/lib/ble/adapter.ts
@@ -45,7 +45,7 @@ const delay = (ms: number) => new Promise<void>((resolve) => setTimeout(resolve,
  * Whichever timer wins clears the other, so a deadline during backoff cannot
  * leave a 500ms timer alive after the connect sequence has already settled. */
 function waitForRetryBackoffBeforeDeadline(deadlineMs: number): Promise<boolean> {
-  const remainingMs = deadlineMs - Date.now();
+  const remainingMs = deadlineMs - performance.now();
   if (remainingMs <= 0) return Promise.resolve(false);
 
   return new Promise<boolean>((resolve) => {
@@ -81,7 +81,7 @@ async function settleBeforeDeadline<T>(
   operation: () => Promise<T>,
   deadlineMs: number,
 ): Promise<DeadlineSettlement<T>> {
-  const remainingMs = deadlineMs - Date.now();
+  const remainingMs = deadlineMs - performance.now();
   if (remainingMs <= 0) return { kind: 'deadline' };
 
   let deadlineTimer: ReturnType<typeof setTimeout> | null = null;
@@ -139,6 +139,12 @@ export class RNBleAdapter implements BluetoothAdapter {
   // Whether a transient first GATT connect failure gets one in-budget retry
   // (see BleAdapterOptions.enableAndroidConnectRetry) — Android only.
   private readonly enableAndroidConnectRetry: boolean;
+  // GATT connect attempts made by the most recent connect, for analytics.
+  // Domain is 0/1/2: the adapter is constructed per connect
+  // (use-board-bluetooth.ts createBluetoothAdapter), so 0 means the flow never
+  // reached the GATT connect at all (picker cancelled, board not found, scan
+  // error), 1 means a single attempt, and 2 means the retry ran.
+  private lastConnectAttemptCount = 0;
 
   constructor(
     private readonly devicePicker: DevicePickerFn,
@@ -154,8 +160,17 @@ export class RNBleAdapter implements BluetoothAdapter {
   private async connectSelectedDevice(
     selectedDeviceId: string,
   ): Promise<{ connected: Device; retrySucceeded: boolean }> {
-    const deadlineMs = Date.now() + CONNECTION_TIMEOUT_MS;
-    const attemptConnect = () => settleBeforeDeadline(() => bleManager.connectToDevice(selectedDeviceId), deadlineMs);
+    // Monotonic clock so a mid-connect wall-clock correction cannot stretch or
+    // collapse the budget. The deadline arithmetic here relies on Vitest faking
+    // `performance.now` alongside the timers — it does by default, so never add
+    // an explicit `toFake` list that omits `performance` or the boundary tests
+    // silently freeze the clock and pass for the wrong reason.
+    const deadlineMs = performance.now() + CONNECTION_TIMEOUT_MS;
+    this.lastConnectAttemptCount = 0;
+    const attemptConnect = () => {
+      this.lastConnectAttemptCount += 1;
+      return settleBeforeDeadline(() => bleManager.connectToDevice(selectedDeviceId), deadlineMs);
+    };
     const cancelWithoutWaiting = () => {
       void bleManager.cancelDeviceConnection(selectedDeviceId).catch(() => {});
     };
@@ -194,9 +209,10 @@ export class RNBleAdapter implements BluetoothAdapter {
 
     const secondError = secondAttempt.error;
     if (isRetryableAndroidConnectError(secondError)) {
-      // No third attempt. Give the exhausted handle the same bounded cleanup as
-      // the first while preserving the exact second error for telemetry.
-      await settleBeforeDeadline(() => bleManager.cancelDeviceConnection(selectedDeviceId), deadlineMs);
+      // No third attempt. Close the exhausted handle without waiting: we already
+      // have the terminal error, so awaiting a cleanup that may hang would only
+      // keep the climber on a spinner before we can show it.
+      cancelWithoutWaiting();
     }
     throw secondError;
   }
@@ -556,6 +572,13 @@ export class RNBleAdapter implements BluetoothAdapter {
   // carries the MTU/chunking story; the park/resume fields are iOS-native-only.
   async getLastWriteDiagnostics(): Promise<BleWriteDiagnostics | null> {
     return this.lastWriteDiagnostics;
+  }
+
+  // 0 when the flow never reached the GATT connect, 1 for a single attempt,
+  // 2 when the Android retry ran. Pairs with `retrySucceeded` to give the retry
+  // a denominator: 2 + success is a save, 2 + failure is a retry that lost.
+  getLastConnectAttemptCount(): number {
+    return this.lastConnectAttemptCount;
   }
 }
 
--- a/packages/mobile/src/lib/ble/adapter.ts
+++ b/packages/mobile/src/lib/ble/adapter.ts
@@ -60,8 +60,11 @@ function waitForRetryBackoffBeforeDeadline(deadlineMs: number): Promise<boolean>
       resolve(completedFullBackoff);
     };
 
-    // Schedule the deadline first so an exact tie preserves the existing
-    // semantics: no second attempt begins once the shared budget is exhausted.
+    // Scheduled first, so on an exact-millisecond tie the deadline settles
+    // before the backoff and no second attempt begins on an exhausted budget.
+    // Both real runtimes and Vitest's fake timers run same-expiry timers in
+    // insertion order; `settle` is idempotent either way, so the tie-break is a
+    // preference, not something correctness rests on.
     deadlineTimer = setTimeout(() => settle(false), remainingMs);
     backoffTimer = setTimeout(() => settle(true), ANDROID_CONNECT_RETRY_BACKOFF_MS);
   });
@@ -207,14 +210,12 @@ export class RNBleAdapter implements BluetoothAdapter {
       return { connected: secondAttempt.result, retrySucceeded: true };
     }
 
-    const secondError = secondAttempt.error;
-    if (isRetryableAndroidConnectError(secondError)) {
-      // No third attempt. Close the exhausted handle without waiting: we already
-      // have the terminal error, so awaiting a cleanup that may hang would only
-      // keep the climber on a spinner before we can show it.
-      cancelWithoutWaiting();
-    }
-    throw secondError;
+    // No third attempt. Close whatever handle attempt two left behind, whatever
+    // it failed with — a non-retryable second error strands one just as readily.
+    // Without waiting: we already hold the terminal error, so awaiting a cleanup
+    // that may hang would only keep the climber on a spinner before we show it.
+    cancelWithoutWaiting();
+    throw secondAttempt.error;
   }
 
   async isAvailable(): Promise<boolean> {
--- a/packages/mobile/src/lib/ble/use-board-bluetooth.ts
+++ b/packages/mobile/src/lib/ble/use-board-bluetooth.ts
@@ -1607,9 +1607,16 @@ export function useBoardBluetooth({
           sizeId,
           failureReason: failureCategory === 'unknown' ? classifyBleFailureReason(error) : failureCategory,
           // 0 = never reached the GATT connect (picker cancelled, board not
-          // found, scan error), 1 = one attempt, 2 = the Android retry ran and
-          // lost. Key is dropped on native iOS and web.
+          // found, scan error), 1 = one attempt, 2 = the Android retry ran.
+          // Key is dropped on native iOS and web.
           connectAttempts: connectAdapter?.getLastConnectAttemptCount?.(),
+          // 2 alone does NOT mean the retry lost: a recovered GATT connect can
+          // still fail at MTU negotiation or service discovery, and that lands
+          // here too. `retrySucceeded: true` on a failure event is exactly that
+          // case — the retry saved the connect, something later broke it — so
+          // the retry hit rate is `retrySucceeded` over `connectAttempts = 2`
+          // across BOTH events, not success events over attempts = 2.
+          retrySucceeded: connectAdapter?.getLastConnectRetrySucceeded?.(),
           // Raw ble-plx codes (Android) so the real connect-failure cause and the
           // low-level GATT status are visible for re-measurement (#3608). Empty on
           // web / native iOS.
--- a/packages/mobile/src/lib/ble/use-board-bluetooth.ts
+++ b/packages/mobile/src/lib/ble/use-board-bluetooth.ts
@@ -1508,7 +1508,11 @@ export function useBoardBluetooth({
           deviceNamePresent: !!connection.deviceName,
           boardId: analyticsBoardId ?? undefined,
           connectedViaMismatchOverride: getConnectedViaMismatchOverride?.() ?? false,
-          retry_succeeded: connection.retrySucceeded === true,
+          retrySucceeded: connection.retrySucceeded === true,
+          // Denominator for retrySucceeded: 0 = never reached the GATT connect,
+          // 1 = one attempt, 2 = the Android retry ran. Key is dropped on native
+          // iOS and web, whose adapters don't count attempts.
+          connectAttempts: connectAdapter?.getLastConnectAttemptCount?.(),
           bleChosenWriteType: connectionDiagnostics?.chosenWriteType,
           bleSupportsWithoutResponse: connectionDiagnostics?.supportsWriteWithoutResponse,
           bleCharProperties: connectionDiagnostics?.characteristicProperties,
@@ -1597,18 +1601,20 @@ export function useBoardBluetooth({
             Alert.alert(t('ble.connectionFailedTitle'), tCommon('bluetooth.unknownError'));
         }
 
-        if (failureCategory !== 'user_cancelled') {
-          track(SHARED_EVENTS.BluetoothConnectionFailed, {
-            boardName,
-            layoutId,
-            sizeId,
-            failureReason: failureCategory === 'unknown' ? classifyBleFailureReason(error) : failureCategory,
-            // Raw ble-plx codes (Android) so the real connect-failure cause and the
-            // low-level GATT status are visible for re-measurement (#3608). Empty on
-            // web / native iOS.
-            ...blePlxErrorCodes(error),
-          });
-        }
+        track(SHARED_EVENTS.BluetoothConnectionFailed, {
+          boardName,
+          layoutId,
+          sizeId,
+          failureReason: failureCategory === 'unknown' ? classifyBleFailureReason(error) : failureCategory,
+          // 0 = never reached the GATT connect (picker cancelled, board not
+          // found, scan error), 1 = one attempt, 2 = the Android retry ran and
+          // lost. Key is dropped on native iOS and web.
+          connectAttempts: connectAdapter?.getLastConnectAttemptCount?.(),
+          // Raw ble-plx codes (Android) so the real connect-failure cause and the
+          // low-level GATT status are visible for re-measurement (#3608). Empty on
+          // web / native iOS.
+          ...blePlxErrorCodes(error),
+        });
       } finally {
         connectInFlightRef.current = false;
         setLoading(false);
--- a/packages/mobile/src/lib/ble/use-board-bluetooth.ts
+++ b/packages/mobile/src/lib/ble/use-board-bluetooth.ts
@@ -1508,6 +1508,7 @@ export function useBoardBluetooth({
           deviceNamePresent: !!connection.deviceName,
           boardId: analyticsBoardId ?? undefined,
           connectedViaMismatchOverride: getConnectedViaMismatchOverride?.() ?? false,
+          retry_succeeded: connection.retrySucceeded === true,
           bleChosenWriteType: connectionDiagnostics?.chosenWriteType,
           bleSupportsWithoutResponse: connectionDiagnostics?.supportsWriteWithoutResponse,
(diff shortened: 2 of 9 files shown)


marcodejongh:
Addressed the classifier-test drift in `640eea82a`: `adapter.test.ts` now imports the real shared `isRetryableAndroidConnectError` through Vitest `importOriginal` instead of reimplementing its allowlist. Focused adapter tests remain 52/52, mobile typecheck and touched `vp check` pass. The picker-cancellation analytics change is intentional and covered; it makes the documented silent-cancel path actually silent. Fable and physical Android remain the draft blockers.

marcodejongh:
Split the combined denied-status/discovery case in `5d36a08f2`; each failure mode now has its own diagnostic test. Focused adapter coverage is 53/53. The `settleBeforeDeadline` comment does not claim zero latency; it only guarantees the shared absolute deadline is not extended, so no production wording change was needed.

packages/mobile/src/lib/ble/adapter.ts
```diff
@@ -71,8 +136,61 @@ export class RNBleAdapter implements BluetoothAdapter {
...
+    }
+    if (secondAttempt.kind === 'fulfilled') {
+      return { connected: secondAttempt.result, retrySucceeded: true };
+    }
+
+    const secondError = secondAttempt.error;
+    if (isRetryableAndroidConnectError(secondError)) {
+      // No third attempt. Give the exhausted handle the same bounded cleanup as
+      // the first while preserving the exact second error for telemetry.
+      await settleBeforeDeadline(() => bleManager.cancelDeviceConnection(selectedDeviceId), deadlineMs);
```
marcodejongh, line 189: This `await` discards its settlement, so the only thing it buys is latency on the user-visible failure.

Concrete: attempt one fails at t=100ms with GATT 133, backoff to t=600ms, attempt two fails at t=700ms, and `cancelDeviceConnection` then hangs (which is exactly the case the "bounds a hanging exhausted cleanup" test pins). The climber stares at the connect spinner until t=12000ms before the alert fires — 11.3s of dead time for a failure we already have in hand.

The first-attempt deadline path two branches up already models the right thing with `cancelWithoutWaiting()`: fire the close, don't block on it. Same semantics here (the handle still gets closed, the rejection is still swallowed), minus the wait:

```ts
if (isRetryableAndroidConnectError(secondError)) {
  cancelWithoutWaiting();
}
throw secondError;
```

That also drops the `settleBeforeDeadline` call whose result is unused, and the accompanying test can assert the error surfaces immediately rather than at the 12s boundary.

---

packages/mobile/src/lib/ble/adapter.ts
```diff
@@ -71,8 +136,61 @@ export class RNBleAdapter implements BluetoothAdapter {
...
     private readonly scanFamily: BoardScanFamily = 'aurora',
+    private readonly enableAndroidConnectRetry = false,
   ) {}
 
+  /** Connect the already-selected peripheral, optionally recovering one known
+   * Android GATT handshake failure without rescanning or reopening the picker. */
+  private async connectSelectedDevice(
+    selectedDeviceId: string,
+  ): Promise<{ connected: Device; retrySucceeded: boolean }> {
+    const deadlineMs = Date.now() + CONNECTION_TIMEOUT_MS;
```
marcodejongh, line 147: Minor, but the deadline moved from a relative `setTimeout(…, 12_000)` to an absolute wall-clock stamp, and every `remainingMs` below is derived from `Date.now()`.

The old code was immune to clock movement; this isn't. If the OS lands an NTP correction (or the user edits the clock) between here and the `remainingMs` reads at lines 47/83, a backward jump inflates the budget — `settleBeforeDeadline` would then arm a timer for the size of the jump instead of ≤12s, and the connect can hang far past the 12s contract this PR is careful to preserve everywhere else. A forward jump expires the budget instantly and turns a healthy connect into a spurious timeout.

Low likelihood, but `performance.now()` is available in Hermes and is monotonic, so `const deadlineMs = performance.now() + CONNECTION_TIMEOUT_MS` with matching `performance.now()` reads costs nothing and removes the failure mode. (Vitest's fake timers mock it too, so the boundary tests keep working.)

---

packages/mobile/src/lib/ble/use-board-bluetooth.ts
```diff
@@ -1190,16 +1191,18 @@ export function useBoardBluetooth({
...
-          boardName,
-          layoutId,
-          sizeId,
-          failureReason: failureCategory === 'unknown' ? classifyBleFailureReason(error) : failureCategory,
-          // Raw ble-plx codes (Android) so the real connect-failure cause and the
-          // low-level GATT status are visible for re-measurement (#3608). Empty on
-          // web / native iOS.
-          ...blePlxErrorCodes(error),
-        });
+        if (failureCategory !== 'user_cancelled') {
```
marcodejongh, line 1194: This one is unrelated to #4143 and it deletes a metric — I'd split it out.

Two problems:

1. **It breaks parity with web.** `packages/web/app/components/board-bluetooth-control/use-board-bluetooth.ts:755` still fires `Bluetooth Connection Failed` unconditionally, including for `user_cancelled`. After this lands, the same PostHog event name means "all connect failures" on web and "all connect failures except picker dismissals" on mobile. Any funnel or connect-success-rate calculation that unions the two platforms silently starts comparing different denominators.

2. **Picker-dismissal rate is a real signal, and this is the only place it's emitted.** How often people open the chooser and back out is how we'd notice an empty-picker regression like #3821 from analytics rather than from support. `bleConnectReportLevel` already keeps `user_cancelled` out of Sentry and the switch already keeps it out of the user's face — those are the two places "silent" needs to mean silent. Analytics isn't one of them.

The precedent for making it visible-but-distinguishable is right there in web: `Pairing Failed` carries `stage: 'user_cancelled'`. If the goal is that cancellations stop counting as failures on dashboards, the cheaper fix is to keep emitting and let the query filter on `failureReason`.

If you'd rather keep the gate, it should land as its own PR (with the web side changed to match) so the retry fix isn't carrying a telemetry change in its blast radius.

---

packages/mobile/src/lib/ble/use-board-bluetooth.ts
```diff
@@ -1102,6 +1102,7 @@ export function useBoardBluetooth({
           deviceNamePresent: !!connection.deviceName,
           boardId: analyticsBoardId ?? undefined,
           connectedViaMismatchOverride: getConnectedViaMismatchOverride?.() ?? false,
+          retry_succeeded: connection.retrySucceeded === true,
```
marcodejongh, line 1105: `retry_succeeded` is the only snake_case property in this payload — every sibling is camelCase (`deviceNamePresent`, `apiLevel`, `connectedViaMismatchOverride`, `bleChosenWriteType`, …), and the failure event next door uses `failureReason` / `bleErrorCode`. PostHog property names are case- and separator-sensitive, so this will sort away from the rest of the BLE props in the UI forever. `retrySucceeded` to match.

Separately: this is a numerator with no denominator. On success we learn a retry saved the connect; when the retry is *attempted and fails*, the failure event carries nothing to say a second attempt happened, and when the first failure lands too late for the backoff to fit the budget no retry is attempted at all — those two are indistinguishable in the data. Since the whole point of the change is to find out whether one retry is enough for #4143, consider adding the attempt count (or a `retryAttempted` boolean) to the failure event too, plumbed the same way `retrySucceeded` is. Otherwise the only way to estimate the mitigation's hit rate is to reverse-engineer it from `bleErrorCode` ∈ {200,201,205} × `androidErrorCode` ∈ {133,147}, which excludes exactly the budget-exhausted cases.

---

marcodejongh (commented):
## Review

Read the whole diff against `origin/main`. The retry state machine itself is solid — one attempt, one absolute budget shared across attempt/cleanup/backoff/retry, terminal errors preserved by identity, both handlers attached in `settleBeforeDeadline` so a late native promise can't become an unhandled rejection, and `vi.getTimerCount()` asserted at zero on the boundary cases. The classifier is appropriately narrower than `BLE_PLX_CODE_TO_CATEGORY`, and gating it on `Platform.OS === 'android'` in the factory keeps iOS and the browser target on the old path.

Four things, none of which need a redesign:

1. **Exhausted cleanup blocks the failure alert** (`adapter.ts:189`) — the `await` throws away its settlement and can leave the climber on a spinner for ~11.3s after we already have the terminal error. `cancelWithoutWaiting()`, same as the first-attempt deadline branch.
2. **Picker-cancel telemetry gate is out of scope and diverges from web** (`use-board-bluetooth.ts:1194`) — web still emits `Bluetooth Connection Failed` for `user_cancelled`, so the shared event name stops meaning the same thing on the two platforms, and picker-dismissal rate stops being measurable anywhere. Worth its own PR.
3. **`retry_succeeded` naming + missing attempt counter** (`use-board-bluetooth.ts:1105`) — lone snake_case key in a camelCase payload, and there's no signal on the failure side, so the retry's hit rate isn't directly computable.
4. **Wall-clock deadline** (`adapter.ts:147`) — `Date.now()` replaced a relative `setTimeout`, so a clock correction mid-connect can stretch or collapse the 12s budget. `performance.now()` is monotonic under Hermes and mocked by Vitest fake timers.

### Blockers unrelated to the code

- **Merge conflict.** GitHub reports `mergeable_state: dirty` against `main` — needs a rebase before anything else matters.
- **Fable BLE review.** Still outstanding per CLAUDE.md's hard rule, and acknowledged in the description. The Sol review doesn't substitute.
- **The PR is marked ready for review, not draft**, while the description lists device verification (real 133 / 147 / null-status failures against a board, two-phone takeover, config switching) as a pre-merge gate and the automated pass says do-not-merge. Worth flipping back to draft so nobody merges it on a green CI signal alone.

On the device gates: the retry is invisible to the user by design, so the thing to actually watch on hardware is that a recovered connect never opens a second picker and never double-fires the success haptic — `startDeviceScan`/`devicePicker` being called once is asserted in the unit test, but the haptic path lives above the adapter.

---

marcodejongh:

packages/mobile/src/lib/ble/adapter.ts
marcodejongh: Fixed — the exhausted-handle cleanup is now `cancelWithoutWaiting()`, the same shape as the first-attempt deadline branch. We already hold the terminal error at that point, so nothing should sit between it and the alert.

The test that covered this (`bounds a hanging exhausted cleanup and still surfaces the exact second error by 12 seconds`) is renamed to `surfaces the second error immediately even when the exhausted cleanup hangs` and now advances 500 ms rather than 12 s, still asserting `cancelDeviceConnection` twice plus `vi.getTimerCount()` at zero.

Deliberately *not* doing the same to the **pre**-retry cleanup: that one has to stay bounded by the full budget. Capping it would abort a slow-but-succeeding `cancelDeviceConnection` that would have cleared somewhere between the cap and 12 s, which throws `firstError` sooner and makes the retry strictly less likely to run at all. The residual tail — a hanging pre-retry cleanup still stalling the alert to the 12 s ceiling — is real but needs a different fix (proceed to the retry after a bounded wait rather than throwing), so it belongs in its own issue.

packages/mobile/src/lib/ble/adapter.ts
marcodejongh: Fixed — all **three** clock reads are `performance.now()` now, not just the `deadlineMs` assignment: the `remainingMs` reads inside `waitForRetryBackoffBeforeDeadline` and `settleBeforeDeadline` had to move too, or the deadline arithmetic would silently mix two clocks.

One correction to the rationale, so the record is accurate: `performance.now()` isn't unconditionally monotonic under Hermes. RN 0.86's `Libraries/Core/setUpPerformance.js` takes the `NativePerformance` path when it's available, and its fallback branch is literally `now: () => (global.nativePerformanceNow || Date.now)()`. So the honest claim is that this is never worse than today and removes the clock-jump hazard on every runtime that has a native clock — which is every real device.

Vitest part checks out: Vitest 4's bundled sinon fake-timers set `clock.performance.now`, its default `toFake` list covers everything except `nextTick`/`queueMicrotask`, and `packages/mobile/vite.config.ts` sets no `toFake` override. The 11_500 / 11_750 / 11_999 boundary tests still pass unchanged. I left a comment at the `deadlineMs` assignment saying so, because a future explicit `toFake` list that omitted `performance` would freeze the clock and let those boundary tests pass for the wrong reason.

packages/mobile/src/lib/ble/use-board-bluetooth.ts
marcodejongh: Reverted — `Bluetooth Connection Failed` fires unconditionally again, exactly as on `main`. The matching `toBeUndefined()` assertion in `stays silent when the user dismisses the device picker` is gone; the `Alert.alert` and `reportHandledError` assertions stay, and both still hold on `main` via `bleConnectReportLevel`.

Correcting one half of the premise, though, because it's gone stale: **web parity no longer applies.** `packages/web/app/components/board-bluetooth-control/` doesn't exist on `main` any more, and `grep -rn BluetoothConnectionFailed packages/` returns only the two emitters in `packages/mobile/src/lib/ble/use-board-bluetooth.ts` plus the name constant in `packages/shared/analytics/src/events.ts`. Nothing on web emits that event today, so the two platforms can't diverge on it.

The other half stands on its own and is why I reverted anyway: `user_cancelled` is 5,380 of ~8,778 `Bluetooth Connection Failed` events over 90 days — 61% — and it's the only measure of picker-dismissal rate we have. Suppressing it deletes that signal. Filing a separate issue for whether it should keep riding this event name or move to its own.

packages/mobile/src/lib/ble/use-board-bluetooth.ts
marcodejongh: Both fixed.

**Naming** — `retry_succeeded` → `retrySucceeded`, matching `deviceNamePresent`, `apiLevel`, `connectedViaMismatchOverride`, `bleChosenWriteType`. Nothing has shipped under the old name, and PostHog property names are separator-sensitive, so this was the moment to do it. (The `BleConnection` *field* was already camelCase and is untouched.)

**Denominator** — both the success and failure events now carry `connectAttempts`. One divergence from what you asked for, and it's forced: this can't be plumbed through `BleConnection` the way `retrySucceeded` is, because the failure path only ever sees a thrown error — there's no connection object to hang it on. So it's a getter on the adapter, `getLastConnectAttemptCount()`, declared optional on `BluetoothAdapter` next to the existing optional `getLastWriteDiagnostics?()` / `getLastConnectDiagnostics?()`. Only `RNBleAdapter` implements it; `NativeIosBleAdapter`, the web adapter and every test double are unaffected, and the key drops out of the payload on those platforms.

The value domain is **0/1/2**, not 1/2. `createBluetoothAdapter` runs per connect, so each connect gets a fresh adapter at count 0, and anything that never reaches the GATT connect reports 0 — picker cancel, board not found, scan error. That's the dominant bucket in practice (5,380 `user_cancelled` + 381 `board_not_found` over 90 days). Full reading:

- `0` — never reached the GATT connect
- `1` — one attempt: an ineligible error, or the first failure landed too late for the 500 ms backoff to fit the remaining budget
- `2` + success — the retry saved it
- `2` + failure — the retry ran and lost

All five are covered in `adapter.test.ts` against the real adapter, plus 0/1/2 through the hook in `use-board-bluetooth.test.ts`.

marcodejongh:
Rebased onto `main` and worked through the outstanding review. `mergeable` is back to `MERGEABLE` (the conflict was one file, `use-board-bluetooth.test.ts` — both `main`'s new config-switch attribution test and this branch's retry assertions are kept), and the OTA bot re-confirms **no native change** on `bf142e8e2`.

**The four inline comments** — each answered on its own thread:

1. Exhausted cleanup no longer awaits, so a hanging `cancelDeviceConnection` can't hold the failure alert back. The **pre**-retry cleanup deliberately keeps its full-budget bound; capping that one would abort a slow-but-succeeding cleanup and make the retry *less* likely to run.
2. `retry_succeeded` → `retrySucceeded`, plus `connectAttempts` on both the success and failure events. Domain is 0/1/2, not 1/2 — `createBluetoothAdapter` runs per connect, so anything that never reaches the GATT connect reports 0, and that's the biggest bucket. It's an adapter getter rather than a `BleConnection` field because the failure path only ever sees a thrown error.
3. Picker-cancel telemetry gate reverted. Half the original rationale is stale (nothing on web emits that event any more), but the other half stands: `user_cancelled` is 61% of that event's 90-day volume and the only measure of picker-dismissal rate. Split out as #4714.
4. All three deadline clock reads moved to `performance.now()` — the two `remainingMs` reads had to move with the `deadlineMs` assignment or the arithmetic would mix clocks. One correction: RN 0.86 falls back to `global.nativePerformanceNow || Date.now`, so this is monotonic on every real runtime rather than unconditionally.

**Automated review's two non-blocking points** — both checked out, both fixed in `bf142e8e2`: cleanup after attempt two no longer branches on whether the second error was retryable (a denied status strands a handle just the same), and the backoff tie-break comment no longer claims more than the runtime guarantees.

**Validation** (all local, this worktree): `typecheck:mobile` ✅ · `typecheck:ble-protocol` ✅ · `test:mobile` 701 files / 6,949 tests ✅ · `ble-protocol` 313 tests ✅ · `vp check` 0 errors ✅ · `check:mobile-bundle` iOS + Android ✅.

Staying **draft**: Fable BLE review and real-device Android QA are both still owed. `.boardsesh/qa-notes.md` has the device script — the two things no unit test can see are that a recovered connect must not reopen the picker and must not double-fire the success haptic.

**Open question for @marcodejongh:** ship the Android retry unconditionally, or behind a PostHog flag? It changes connect timing for every Android user in one OTA. Nothing in the issue or the review asks for a flag, and it's currently on for all of Android — say the word if you'd rather stage it.

marcodejongh:
CI is green on `bf142e8e2` — 23 checks, zero failures, including `mobile-bundle`, `typecheck`, every test shard, and `ota-check` (no native change).

On the automated review's two minor points, neither needs a change:

- **`connectAttempts` absent on native iOS / web** — intentional. `AnalyticsEventProperties` is `Record<string, AnalyticsPropertyValue | undefined>` and `sanitizeForPosthog` strips `undefined`, so the key is dropped rather than sent as null. Same shape as the existing `bleChosenWriteType` / `bleMaxWriteWithResponse` reads. PostHog queries should filter on `connectAttempts is set` when they want the Android denominator.
- **Racing teardown nulling the adapter before telemetry fires** — can't happen. `connectAdapter` is a `let` local to the connect callback (`use-board-bluetooth.ts`), not a ref, so nothing outside that invocation can reassign it. And if it somehow were null, the optional chain drops the key instead of throwing.

The two blockers it lists are the two this PR already declares owed, and they're why it's still draft: Fable BLE review, and real-device Android QA.

marcodejongh:
## Fable BLE review — PR #4153

**Verdict: APPROVE WITH NITS.** No blocking findings. The retry state machine is sound: I could not construct a handle leak, a double-connection, an unbounded phase, or a back-door auto-reconnect from the code as written. The nits below are follow-up material, not merge blockers. This satisfies the CLAUDE.md Fable gate for the code; the PR's own "real-device Android QA" checklist item remains owed and is outside what this review can discharge (see "Not verifiable here").

### What I verified, and how

I read the full diff plus the surrounding `requestAndConnect` flow, the hook connect lifecycle, and the installed `react-native-ble-plx@3.5.1` and `react-native@0.86.2` sources (not just their docs).

**Double connection — provably impossible.** Attempt two starts only after attempt one's `DeadlineSettlement` is `rejected`, i.e. after the native `connectToDevice` promise itself settled (`adapter.ts:167-179`). The `deadline` outcome throws without retrying (`adapter.ts:172-174`), and the pre-retry cleanup is awaited inside the budget (`adapter.ts:188`), so cleanup can't overlap attempt two either. On the deadline branches, `cancelWithoutWaiting()` is genuinely effective against an in-flight connect: ble-plx's `cancelDeviceConnection` "disconnects... or cancels pending connection" (`BleManager.js:496-503` in the installed 3.5.1). Bonus safety I confirmed in the dep: Android `connectToDevice` pre-cancels an existing connection itself (`BleManager.js:488-490`), so even a missed stale handle can't wedge the retry.

**Handle leaks — every abandoning path accounted for.** First-attempt deadline: cancel + throw (`adapter.ts:172-174`). Cleanup-deadline and backoff-denied: attempt one's handle was already the target of the awaited/issued cancel (`adapter.ts:188-192`). Second-attempt deadline: cancel + throw (`adapter.ts:195-197`). Second-attempt failure — including a *non-retryable* second error: unconditional fire-and-forget cancel (`adapter.ts:203-208`, asserted by the "still closes the handle when attempt two fails with a non-retryable error" test). The detached promises all keep both handlers attached (`settleBeforeDeadline`'s `.then(onFulfilled, onRejected)`, `adapter.ts` helper), so no late native settlement can become an unhandled rejection; the "no late unhandled rejection" test exercises exactly that.

**Eligibility — correct and narrow.** `isRetryableAndroidConnectError` (`connection-error.ts:93-110`) requires `name === 'BleError'`, numeric `errorCode` in {200, 201, 205} (verified against the installed enum: DeviceConnectionFailed / DeviceDisconnected / DeviceNotConnected — all connect-side; 3 OperationTimedOut, 204 DeviceNotFound, 206 MTU, 3xx/4xx discovery are all excluded), rejects picker-cancel wording, vetoes any populated `iosErrorCode`, denies malformed statuses (`'133'` as a string is denied, not coerced), and accepts only 133/147/null as the Android status. The double gate (factory passes `Platform.OS === 'android'` at `adapter-factory.ts:24`, plus the iOS-shape veto inside the predicate) means an iOS ble-plx error can't retry even if the flag were mis-wired. The allowlist matches the fleet table in the PR body. Test fakes are faithful: installed `BleError` sets `name = 'BleError'` and nulls (not deletes) inactive platform fields, exactly the shape `androidConnectError()` builds.

**The 12-second budget — genuinely absolute.** All four in-budget phases recompute `remainingMs` off one `deadlineMs` (`adapter.ts:161`); the exhausted cleanup is deliberately outside it and fire-and-forget, so it cannot stall the failure alert (tested at 500ms with zero pending timers). The backoff/deadline race schedules the deadline timer first so an exact tie denies the retry, and both helpers clear both timers on settle — the boundary tests assert `vi.getTimerCount() === 0`. Worst-case user-visible latency is unchanged from `main`'s single-shot 12s ceiling, which also means the retry does not widen the hook's existing mid-connect config-switch window.

**No auto-reconnect back door.** The retry lives entirely inside one user-initiated `requestAndConnect`; `onDeviceDisconnected` is only registered after full success (`adapter.ts:431`), so no disconnect event can trigger a reconnect, and iOS keeps the native adapter untouched. The hook's `connectInFlightRef` guard (`use-board-bluetooth.ts:1032-1035`) still serializes connects. Two-phone note: if phone B grabs the board during A's 500ms backoff, A's retry will take it back — but that is within A's own connect gesture seconds after A tapped, indistinguishable from a slow first connect; last-connection-wins semantics are preserved, not automated.

**Telemetry.** `connectAttempts` 0/1/2 is documented at both emit sites and at the getter (`adapter.ts:137-142`, `use-board-bluetooth.ts:1307-1310`, `1404-1407`); the count can never claim an attempt that didn't reach `connectToDevice` (the first attempt always starts with the full budget, and a post-backoff second attempt always has `remainingMs > 0`). Key dropped on native iOS/web via the optional getter. `retrySucceeded` is a literal boolean on every success event — it reads `false` on iOS/web where no retry exists, so hit-rate queries must scope by `connectAttempts` presence or platform; the PR body says as much.

### Nits (most significant first — none blocking)

1. **Retry-denied first failure skips handle cleanup** — `adapter.ts:181-183` throws a non-retryable first error (e.g. 201 x status 19) with no `cancelWithoutWaiting()`, while the exhausted path cleans up even non-retryable second errors on the stated grounds that "a non-retryable second error strands one just as readily" (`adapter.ts:203-204`). The same argument applies one attempt earlier: a denied first failure can strand the native handle, and the next tap greets GATT 133 — mitigated now because that 133 is precisely what the new retry recovers, and this is byte-for-byte today's behaviour (no regression). Worth a one-line follow-up, not this PR.
2. **Pre-retry cleanup can stall feedback to the full ceiling** — a first failure at t≈1s followed by a hanging `cancelDeviceConnection` surfaces the error at t=12s (`adapter.ts:188-189`, covered by the "cleanup hangs through the shared deadline" test), where `main` would have shown it at ~1s. The author's trade (capping cleanup would abort a slow-but-succeeding cleanup and forfeit the retry) is defensible and the bound equals today's dead-board worst case. I assess the trade as acceptable; if `connectAttempts=1` failures with ~12s latency show up in PostHog, that's the signature to revisit.
3. **`lastConnectAttemptCount` reset placement assumes a fresh adapter** — the reset lives in `connectSelectedDevice` (`adapter.ts:162`), but `requestAndConnect` already resets `negotiatedMtu` at its top (`adapter.ts:221`) *explicitly for the reused-adapter case*. Today the hook constructs a new adapter per connect (`use-board-bluetooth.ts:1059`; the adoption path at 1550 is native-iOS-only), so domain-0 reporting is correct — but if any future caller reuses an `RNBleAdapter` and its second connect fails during the scan, the failure event would report the *previous* connect's 1/2 as this connect's `connectAttempts`. Move the reset beside the MTU reset when convenient.
4. **`attErrorCode` is not consulted** — installed ble-plx nulls `androidErrorCode` when the underlying status is an ATT error and reports it in `attErrorCode` instead, so such an error lands in the retryable "null status" bucket. For connection-establishment failures an ATT status is essentially unheard-of, and the fleet data justifies retrying null regardless; recording it here so the next reader knows it was considered.
5. **Clock-source fine print** — I verified RN 0.86.2's fallback (`Libraries/Core/setUpPerformance.js:27`: `global.nativePerformanceNow || Date.now`) is only installed when `NativePerformance` is absent; Hermes provides it, so real devices are monotonic. Under the `Date.now` fallback the new deadline *arithmetic* is strictly less clock-jump-immune than the old relative `setTimeout` (forward jump → spurious instant timeout; backward jump → stretched budget) — so "never worse than today" has that one asterisk, with effectively zero exposure on the Android hardware where the retry is enabled. Also confirmed the repo has no `fakeTimers.toFake` override anywhere, so the in-code warning about Vitest faking `performance` (`adapter.ts:156-160`) currently holds — and the green boundary tests are themselves evidence it's faked.

### Not verifiable here

No Android device, no board, no way to induce real GATT 133/147 in this environment. Specifically unverified: actual 133 recovery rate on hardware; that a recovered connect never re-opens the picker or double-fires the success haptic on-device (unit tests assert single `startDeviceScan`/`devicePicker` calls, but the haptic lives above the adapter); two-phone takeover feel; and the Kotlin-side handle behaviour of ble-plx on a rejected connect (JS-side contracts verified in the installed source only). The PR's "Real-device Android QA" checklist item stands.

marcodejongh:
## Code review — PR #4153

**APPROVE WITH NITS.** Read the full diff at `bf142e8e2` plus the surrounding source. No blocking code findings. The retry state machine is correct as written, the new tests genuinely discriminate against the unfixed code, all four claimed review fixes actually landed and are complete, both declined `claude-review` points are correct against the source, and nothing native or fingerprint-moving slipped in.

This is the code/test/convention lane only — Fable is covering GATT semantics, handle lifecycle and retry eligibility, and that review is still the real merge blocker.

---

### Do the tests actually test anything? Mostly yes.

I specifically looked for the "harness never captures the field being asserted" mode from this PR's history. It isn't present:

- `mockTrack` (`use-board-bluetooth.test.ts:55`) is the direct mock of `../../analytics`'s `track`, so it captures the raw properties object. `sanitizeForPosthog` runs downstream in the real client (`packages/shared/analytics/src/create-analytics.ts:53`), so `toMatchObject({ connectAttempts: 0 })` really does distinguish `0` from `undefined` — it would not silently half-pass.
- `adapter.test.ts:34` now pulls the real `isRetryableAndroidConnectError` through `importOriginal` instead of re-declaring the allowlist. That closes the earlier assert-the-implementation-back-to-itself hole.
- Both suites are registered in `vite.config.ts` `test.projects` (`packages/shared/ble-protocol` line 118, `packages/mobile` line 150), so the 108 new classifier lines and the 327 new adapter lines actually gate CI rather than sitting dead.
- The strongest test in the diff is `adapter.test.ts` `it.each([11_500, 11_750, 11_999])`. It is not decorative: at `11_500` the remaining budget is exactly the 500 ms backoff, so if `waitForRetryBackoffBeforeDeadline` scheduled its timers in the other order the backoff would win, attempt two would start on a zero budget, and the test would receive `connectionTimeoutError()` instead of `firstError` and fail. `expect(vi.getTimerCount()).toBe(0)` on the same case is a real leak assertion.

---

### Findings, most severe first

**1. `connectAttempts: 2` on the failure event does not mean what the comment says it means** — `use-board-bluetooth.ts:1401-1407` (comment) with `adapter.ts:162`. *Minor, but it undercuts the exact thing the denominator was added for.*

The comment states `2 = the Android retry ran and lost`. That is only true when the terminal error came from the GATT connect itself. `lastConnectAttemptCount` is reset at the top of `connectSelectedDevice` (`adapter.ts:162`) and never touched again, and `retrySucceeded` is emitted **only** on the success event.

Concrete scenario: GATT 133 on attempt one → the retry connects successfully (`retrySucceeded` would be `true`) → `connected.discoverAllServicesAndCharacteristics()` rejects at `adapter.ts:399`. The hook emits `Bluetooth Connection Failed` with `connectAttempts: 2, failureReason: 'service_missing'`, and nothing in the payload says the retry actually won. Anyone computing the hit rate as `(Success ∧ attempts=2) / (attempts=2)` — which is what the PR body's "2 + failure = the retry ran and lost" invites — undercounts saves by every post-connect failure.

Cheapest fix is the comment plus the PR body: say the denominator must also filter `failureReason = 'connect_failed'`. The more complete fix is to carry `retrySucceeded` onto the failure event too (the adapter already knows it; it just isn't exposed on the throw path, same shape as `getLastConnectAttemptCount`). No test covers "retry succeeds, discovery fails", so this is currently unguarded either way.

**2. `adapter.test.ts:1168` is tautological** — `it('reports zero connect attempts before the flow ever reaches the GATT connect')` constructs an adapter and asserts `getLastConnectAttemptCount() === 0`. That is the field initializer at `adapter.ts:143`; the test passes whether or not `connectSelectedDevice` resets the counter, and whether or not a pre-connect stage bumps it. Nothing drives the real `RNBleAdapter` through a picker-cancel to prove the documented `0` bucket. The hook-side counterpart (`use-board-bluetooth.test.ts:334`) hand-feeds `vi.fn(() => 0)`, so the `0` domain is asserted twice and demonstrated zero times. Low severity — the adapter is genuinely per-connect (`use-board-bluetooth.ts:1059`), so `0` is right today — but this test buys nothing.

**3. Duplicated user-cancel regex with a comment that demands they stay in lock-step and nothing enforcing it** — `packages/shared/ble-protocol/src/connection-error.ts:100` vs `:179`. Both are the literal `/user cancell?ed|Device selection cancelled/i`, and the new one's comment says "Keep this identical to classifyBleFailure's deliberately narrow user-cancel wording". They match today. Failure scenario: a later PR broadens `classifyBleFailure`'s pattern (say, adds `dismissed`) and misses line 100 — a ble-plx 200/201/205 error carrying that wording then becomes retry-eligible and the climber eats a 500 ms backoff plus a doomed second connect after cancelling. One `const USER_CANCEL_MESSAGE_PATTERN` used by both removes the class.

**4. (Unsure — I believe unreachable) the attempt counter increments before the operation is known to run** — `adapter.ts:163-166`. `attemptConnect()` does `this.lastConnectAttemptCount += 1` and then calls `settleBeforeDeadline`, which short-circuits to `{ kind: 'deadline' }` at `adapter.ts:86-87` **without invoking the operation** when `remainingMs <= 0`. On that path the count reports a GATT connect that never reached ble-plx. I traced both call sites and think it can't fire: attempt one always has the full ~12 s, and attempt two only runs when the backoff timer beat the deadline timer, which implies a positive remainder. Flagging it as a robustness nit rather than a live bug — incrementing inside the operation closure would make it structurally true.

**5. `adapter.test.ts` `beforeEach` is `vi.clearAllMocks()` (line 198), which does not clear implementations.** The new block installs two persistent hanging stubs — `cancelDeviceConnection.mockImplementation(() => new Promise(() => {}))` at `:1242` (and the pre-existing `connectToDevice` one at `:955`). I walked the current file order and every downstream test that touches `cancelDeviceConnection` re-stubs it, so nothing is broken today and CI's green is real. It is an order-dependence trap for the next test appended to this describe block. No change needed for this PR; worth knowing.

---

### The four claimed fixes — all four landed, all four complete

1. **Exhausted cleanup no longer blocks the alert.** `adapter.ts:207` is `cancelWithoutWaiting()` immediately before `throw secondAttempt.error`, matching the deadline branches at `:173` and `:196`. The test advances only 500 ms with a hanging second `cancelDeviceConnection` and asserts `vi.getTimerCount() === 0`, which is the right shape — it would fail if the `await` came back.
2. **`retrySucceeded` naming + denominator.** `use-board-bluetooth.ts:1306` (camelCase, literal boolean via `=== true`) and `connectAttempts` on **both** events — `:1310` success, `:1407` failure. Grepped: no `retry_succeeded` remains anywhere.
3. **Picker-cancel telemetry revert.** `track(SHARED_EVENTS.BluetoothConnectionFailed, …)` at `use-board-bluetooth.ts:1399` is unconditional — no `failureCategory` guard above it, and the `user_cancelled` case in the switch at `:1381` only suppresses the `Alert`, which is pre-existing behaviour. Fully reverted.
4. **`Date.now()` → `performance.now()`.** This is the one I checked hardest, given the earlier "all three was really four sites plus a test title" miss. `git show 55cd205d9 -- adapter.ts` converts exactly three reads (now `adapter.ts:47`, `:86`, `:161`); `grep -n 'Date\.now' packages/mobile/src/lib/ble/adapter.ts` returns nothing; `adapter.test.ts` contains no `Date.now`, no `performance.now`, and no stale clock-flavoured test title. The two surviving `Date.now()` reads in `use-board-bluetooth.ts:536/600` are the pre-existing connection-lifetime duration and are untouched by this diff. Complete.

---

### The two declines — both correct

- **`connectAttempts` absent on iOS/web is intentional.** Verified end to end: mobile `track` is the `createAnalytics` wrapper (`packages/mobile/src/lib/analytics.ts:136`), `create-analytics.ts:53` calls `sanitizeForPosthog(properties)` on every `track`, and `sanitize.ts:14` drops `undefined` keys rather than sending null. The optional-chained read at `:1310`/`:1407` therefore drops the key. Decline stands.
- **No racing-teardown hazard on `connectAdapter`.** `use-board-bluetooth.ts:1045` declares `let connectAdapter: BluetoothAdapter | null = null` **inside** the connect callback and assigns it at `:1060`; it is not a ref and nothing outside that invocation can reach it. `connectInFlightRef` also gates re-entry, and each connect builds its own adapter, so the count can't be clobbered by a parallel connect. Decline stands.
- One note in the bot's favour, since it has a history of citing rules that don't exist here: the **Fable BLE review** blocker it raised *is* verbatim in the repo-root `CLAUDE.md` ("Bluetooth changes require a Fable review"). That one is real and this PR is correctly still a draft because of it.

---

### Scope and conventions — clean

- **OTA-safe.** Nine files, all `.ts`, zero Swift, zero `plugins/`, `modules/`, `app.config.ts`, `patches/` or `package.json`. The OTA bot confirms both fingerprints match `main` (`183bcf484292` / `a6f256fdaf25`). Issue #4143's scope item 2 (iOS `connectFailureReason`) stayed out. Nothing to flag.
- **CLAUDE.md.** No `any` anywhere in the diff (test doubles go through `as unknown as`); both test files live in `src/__tests__/`; the classifier lives in `packages/shared/ble-protocol` and is imported by mobile rather than duplicated; names are descriptive throughout (`firstAttempt`, `secondAttempt`, `completedBackoff`, `remainingMs`, `cancelWithoutWaiting` — no `r`/`x`/`data`/`temp`); the old inline `Promise.race` + `connectionTimeoutId` block is deleted rather than left dead. No new user-facing strings, so no i18n surface. No list, provider, gesture or board-art change, so the RN perf checklist doesn't apply.

---

### Correctness of the diff as code — nothing found

For the record, the things I went looking for and did not find: `settleBeforeDeadline` cannot reject (both race arms resolve to a tagged value, and `Promise.resolve().then(operation)` captures a synchronous throw), so `attemptConnect()` can't blow up the caller; its `.finally` clears the deadline timer on every exit path including the operation-wins path; `waitForRetryBackoffBeforeDeadline`'s `settle` is idempotent and clears both timers, so no branch leaves a live 500 ms timer; the single `deadlineMs` is genuinely shared across attempt one, cleanup, backoff and attempt two, so the "no later than today" claim holds; terminal errors are rethrown by identity (`throw firstError` / `throw secondAttempt.error`) rather than rewrapped, which the `.rejects.toBe(...)` assertions pin; and keeping both handlers attached after the deadline wins removes a real unhandled-rejection path that `main` had.

_Reviewed read-only. Fable BLE review and real-device Android QA remain the merge gates._

marcodejongh:
Thanks both — three of the nits are fixed in `4abe83228`, and I'm naming the ones I deliberately left.

**Fixed**

1. **`connectAttempts: 2` was mislabelled** (code review #1). The comment claimed "2 = the retry ran and lost", which the code never meant: `lastConnectAttemptCount` isn't reset once `connectSelectedDevice` returns, and `retrySucceeded` was only emitted on the success event. So a retry that *won* its GATT connect and then failed at MTU negotiation or service discovery also reported `2` on a failure event — and `(Success ∧ attempts=2)/(attempts=2)` undercounts saves. That is precisely the number that decides whether this retry is worth keeping, so it's now tracked separately (`getLastConnectRetrySucceeded()`) and emitted on the failure path too. The rate is `retrySucceeded` over `connectAttempts = 2` across **both** events. Covered by a new test on the previously-untested recovered-then-failed-downstream path.
2. **Tautological assertion** (code review #2) — `adapter.test.ts` asserted the field initializer, so it passed whether or not `connectSelectedDevice` reset the counter. Replaced with one that fails if the reset goes away.
3. **Duplicated user-cancel regex** (code review #3) — collapsed to a single `USER_CANCEL_MESSAGE_PATTERN`. The two copies carried a comment demanding lock-step with nothing enforcing it. No `g` flag, so there's no `lastIndex` state to share across call sites.

**Deliberately deferred**

- **Retry-denied first failure skips handle cleanup** (Fable #1) — real, but pre-existing behaviour rather than something this PR introduces. Follow-up material.
- **Hanging pre-retry cleanup can stretch the visible error to the 12s ceiling** (Fable #2) — keeping the trade. Capping it aborts a slow-but-succeeding cleanup and makes the retry *less* likely to run, which is the opposite of the point. Worth watching in PostHog as `connectAttempts=1` paired with ~12s latency, as suggested.
- **`attErrorCode` not consulted** (Fable #4) — considered and accepted; ATT-status errors land in the retryable null bucket.
- **`vi.clearAllMocks()` doesn't clear implementations** (code review #5) — nothing is broken today; noting it rather than churning the suite.

**One correction to my own PR body, from Fable #5:** the `performance.now()` swap is *not* strictly "never worse than today". RN 0.86.2's fallback is `now: () => (global.nativePerformanceNow || Date.now)()`, so the `Date.now` branch is less clock-jump-immune than the old relative `setTimeout`. Monotonic on real Hermes and the exposure is near zero, but the body overclaimed and I've stopped asserting it.

Validation on `4abe83228`: `typecheck:mobile` clean · `test:mobile` 701 files / 6,953 tests pass · `ble-protocol` 313 pass · `vp check` clean via the pre-commit hook.

Still owed before merge, unchanged: **real-device Android QA** (a recovered connect must not reopen the picker or double-fire the success haptic — no unit test sees either). Staying a draft.

marcodejongh:
Marking this ready for review. Both required reviews have landed and CI is green on `4abe83228`.

- **Fable BLE review: APPROVE**, no blocking findings ([comment](https://github.com/boardsesh/boardsesh/pull/4153#issuecomment-5390782554)). It verified the risky parts against the installed `react-native-ble-plx@3.5.1` rather than from docs — double connection is provably impossible, all five abandoning paths release the handle, and the retry eligibility set is correctly narrow.
- **Code review: APPROVE**, no blocking findings ([comment](https://github.com/boardsesh/boardsesh/pull/4153#issuecomment-5390784786)).
- **CI on `4abe83228`:** 24 checks pass, zero failures. `ota-check: SUCCESS` — no native change vs `main`, so this ships over the air. Only `publish` (Vercel preview) is outstanding and it gates nothing.

**One gate is still owed and it is not automatable here: real-device Android QA.** The two behaviours no unit test can see are that a recovered connect must not reopen the device picker, and must not double-fire the success haptic. A device script is in the worktree. Please treat that as blocking the merge even though the checkmarks are green.

**And one open question for @marcodejongh before this goes out:** ship the retry unconditionally, or behind a PostHog flag? It changes connect timing for every Android user in a single OTA. Nothing in the issue or either review asks for a flag, and it currently ships on for all of Android — so this is a deliberate call, not an oversight.

marcodejongh:

Conflicting: packages/mobile/src/lib/ble/adapter-factory.ts, adapter.ts. BLE-touching — per CLAUDE.md this needs a Fable review once it's resolved and ready to merge, in addition to the rebase.

PR #4155, opened 2026-08-01
fix(shared): retry transient web GATT connects

### LINKED ISSUE #4144: Web Bluetooth connect never retries, so a transient GATT race costs a re-pick
Connect to your board from a browser and about a third of the time the GATT connect just fails — you pick the board, nothing happens, and you have to tap the lightbulb and pick it again. It almost always works on the second try, because the failure is a transient connect race the code never retries.

## What the data shows

30 days, web sessions (`$app_version IS NULL`):

| Platform / stage | Events | Users |
|---|---|---|
| Android `gatt_connect` / `NetworkError` | 31 | 15 |
| Android `gatt_connect` / `Error` | 4 | 2 |
| Linux `NetworkError` | 36 | 2 |
| Android `user_cancelled` / `NotFoundError` | 61 | 22 |

Real figure: **17 of 56 Android web-BLE users (~30%)**, against 70 successful web-BLE users total. The Linux row is two people, almost certainly dev machines. The `user_cancelled` row is the Chrome chooser being dismissed and is not this bug.

**The retry case is evidenced, not assumed.** The obvious refutation is that `NetworkError: Connection attempt failed` is environmental — board off, out of range, another phone holding the link — in which case a retry is useless. Measuring time-to-next-success per affected user kills that: **16 of 17 Android web users hit `Bluetooth Connection Success` 5–16 seconds after the failure** (one at 144 s). A board that is powered off or out of range does not come back in six seconds. That's a transient GATT connect race.

## Root cause (verified on `main`)

`packages/shared/ble-protocol/src/web-transport.ts` — `getUartCharacteristic` (:137) and `getMoonboardWriteCharacteristic` (:165) each do a single bare `await device.gatt?.connect()` with no retry, and `requestAndConnect` calls it once. The only retry in the file is on the write path (`writeCharacteristicSeries` `NotSupportedError` fallback), not connect.

## Why this is worth fixing despite #3122

#3122 closed the Next.js web climbing surface as no longer receiving fixes. **This one lives in the shared package and is used by the live surface:** `packages/mobile/src/lib/ble/web-adapter.ts` — the Expo web app, which is becoming the climbing surface — imports `@boardsesh/ble-protocol/web-transport` and calls the same unretried connect. Fixing it in the shared transport benefits the Expo web app; `packages/web` just inherits it.

## Suggested fix

Bounded retry with short backoff around the `device.gatt.connect()` calls, reusing the already-granted device handle so there's no second chooser prompt.

Scope it to the connect retry only. The "user cancels are logged as connection failures" half is already tracked as item 3 of #3088 — don't re-litigate the event shape here.

## Severity

**P3.** All 17 affected users self-recovered within seconds; the cost is one extra lightbulb tap plus a device re-pick. No data loss, no dead end, and the population is 56 Android web-BLE users over 30 days on an already-minority path. Cheap fix, recoverable papercut.

## Process

Touches `packages/shared/ble-protocol/` → **requires a Fable review before merge** per CLAUDE.md.

## Related

- #3088 (OPEN, P3) — web `Bluetooth Connection Failed` / `Pairing Failed` double-emission and missing `failureReason`. Covers the event-shape half.
- #3122 (CLOSED) — web-surface deprecation policy; see above for why this is out from under it.

marcodejongh:
## Summary

- retry one rejected Web Bluetooth `gatt.connect()` after 500 ms only when the error name is exactly `NetworkError`
- reuse the already granted device/GATT handle without reopening the chooser or disconnecting between attempts
- share the helper across Aurora UART and MoonBoard UART/RedBearLab probing
- skip connect when GATT is already connected and preserve existing no-GATT behavior
- keep chooser, permission, unsupported, invalid-state, abort, TypeError, generic, service, and characteristic failures single-attempt

Closes #4144.

## Scope

The live Expo browser adapter imports this shared transport. No retired `packages/web` climbing-surface source changed; the old surface inherits the shared fix incidentally.

## QA

- final independent Sol xhigh review: PASS
- shared BLE suite: 278 tests passed
- Expo web-adapter focused suite: 21 tests passed
- full mobile suite: 575 files, 4,914 tests passed
- shared, BLE-protocol, and mobile typechecks
- `vp run check:mobile-web-bundle` and native mobile bundle checks
- touched-file `vp check`, pre-commit hooks, and diff check
- deterministic fake-timer coverage leaves zero timers after success/failure

## Browser verification gates

Before merge, exercise Expo web on Android Chrome with a real transient first connect failure and confirm recovery uses one chooser; also smoke picker dismissal, unsupported browsers, missing characteristics, and MoonBoard RedBearLab fallback. Fable is unavailable in this runtime and remains a literal shared-BLE review blocker; the Sol review is a substitute record, not Fable sign-off.

## Release Notes

Browser board connections now retry one brief GATT failure without making you choose the board again.

### DIFF
--- a/packages/mobile/src/lib/ble/__tests__/web-adapter.test.ts
+++ b/packages/mobile/src/lib/ble/__tests__/web-adapter.test.ts
@@ -119,6 +119,61 @@ describe('requestAndConnect', () => {
     expect(connection).toEqual({ deviceId: 'device-abc', deviceName: 'Kilter Board' });
   });
 
+  it('retries a transient GATT NetworkError without reopening the browser chooser', async () => {
+    vi.useFakeTimers();
+    try {
+      const { device, server } = makeDevice({ [UART_SERVICE_UUID]: makeCharacteristic() });
+      server.connected = false;
+      server.connect.mockRejectedValueOnce(
+        Object.assign(new Error('Connection attempt failed'), { name: 'NetworkError' }),
+      );
+      server.connect.mockResolvedValueOnce(server);
+      const requestDevice = vi.fn().mockResolvedValue(device);
+      stubBluetooth(requestDevice);
+
+      const connectionPromise = new WebBluetoothAdapter('aurora').requestAndConnect();
+      await vi.advanceTimersByTimeAsync(0);
+
+      expect(requestDevice).toHaveBeenCalledTimes(1);
+      expect(server.connect).toHaveBeenCalledTimes(1);
+
+      await vi.advanceTimersByTimeAsync(499);
+      expect(server.connect).toHaveBeenCalledTimes(1);
+
+      await vi.advanceTimersByTimeAsync(1);
+      await expect(connectionPromise).resolves.toEqual({
+        deviceId: 'device-abc',
+        deviceName: 'Kilter Board',
+      });
+      expect(requestDevice).toHaveBeenCalledTimes(1);
+      expect(server.connect).toHaveBeenCalledTimes(2);
+      expect(vi.getTimerCount()).toBe(0);
+    } finally {
+      vi.useRealTimers();
+    }
+  });
+
+  it('propagates chooser cancellation without retrying the chooser or connecting GATT', async () => {
+    vi.useFakeTimers();
+    try {
+      const { server } = makeDevice({ [UART_SERVICE_UUID]: makeCharacteristic() });
+      server.connected = false;
+      const chooserError = Object.assign(new Error('User cancelled the browser chooser'), {
+        name: 'NotFoundError',
+      });
+      const requestDevice = vi.fn().mockRejectedValue(chooserError);
+      stubBluetooth(requestDevice);
+
+      await expect(new WebBluetoothAdapter('aurora').requestAndConnect()).rejects.toBe(chooserError);
+
+      expect(requestDevice).toHaveBeenCalledTimes(1);
+      expect(server.connect).not.toHaveBeenCalled();
+      expect(vi.getTimerCount()).toBe(0);
+    } finally {
+      vi.useRealTimers();
+    }
+  });
+
   it('resolves the MoonBoard write characteristic via the RedBearLab fallback', async () => {
     // Only the RedBearLab service is present — UART probing rejects first.
     const redbearChar = makeCharacteristic(false);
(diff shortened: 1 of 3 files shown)


marcodejongh:
Clarified the timing assertion in `29ac2a546`: the test constant is now named `EXPECTED_GATT_CONNECT_RETRY_DELAY_MS` with an explicit note that it is deliberately independent of the private source constant. The 500 ms value is the issue contract; importing production timing would make the test follow a drift instead of catching it. Focused shared tests are 16/16 and touched `vp check` passes.

marcodejongh:
Closed the remaining adapter-level coverage note in `d255aa1f1`. The live `WebBluetoothAdapter.requestAndConnect()` tests now explicitly prove that an already-connected GATT server skips `connect()` for both the Aurora UART path and the MoonBoard RedBearLab fallback path. Focused adapter coverage remains green at 21/21, scoped `vp check`, `git diff --check`, and the mandatory pre-commit hook pass.

The deliberately independent 500 ms expectation remains unchanged so the issue contract catches production timing drift. Physical Chrome Web Bluetooth and literal Fable review remain draft gates.

marcodejongh:
Applied the final nonblocking readability suggestion in `c01090884`: the one-retry contract is now expressed directly as first `connect()`, strict `NetworkError` guard, 500 ms wait, and one final `connect()`. This removes the two-iteration loop and unreachable fallback without changing behavior.

Independent QA confirmed exact semantic parity. Shared transport tests pass 16/16, mobile adapter tests pass 21/21, BLE protocol typecheck and scoped checks pass. Physical Chromium and literal Fable review remain the only stated gates.

marcodejongh:
Triaged the final nonblocking notes; no code change is appropriate.

- Neither the repository `CLAUDE.md` nor `AGENTS.md` contains the quoted prohibition on multi-line comment blocks. The existing JSDoc documents a browser constraint that materially governs the implementation: Web Bluetooth exposes no cancellable connect operation, so racing a timeout can overlap GATT operations. Removing that rationale would make a dangerous future timeout change more likely.
- `web-transport.test.ts` uses fake timers in every test, so suite-level setup is correct there. Only two `web-adapter.test.ts` cases need fake timers; their local `try/finally` keeps timer virtualization from leaking into the other adapter tests and guarantees restoration if an assertion fails.

All CI and code-quality checks are green. The acknowledged literal Fable and physical Chromium gates remain unchanged.

PR #4156, opened 2026-08-01
fix(mobile): show board write activity

### LINKED ISSUE #4145: No sign the board is being written to, so climbers tap again mid-send
Pick a climb with the board connected and the app gives you nothing while the frames are actually going out over Bluetooth. The phone UI updates instantly, the wall takes a second or two, and in between there's no sign anything is happening — so people tap again.

## What the data shows

Replay: [019f842c at 2:54](https://us.posthog.com/project/412845/replay/019f842c-950f-767b-bd8b-2ebe46ede760?t=174) — three rage-taps during a board sync in one session.

**Stating the evidence honestly:** this is one user, in one session, labelled by an LLM replay summariser ("Repeated taps during board sync"). It is an interpretation of a replay, not an instrumented signal — nothing ties those taps to a BLE write, and they're equally consistent with scroll lag or a slow sheet. The project-wide `$rageclick` total (102 events / 8 users) is **not** a measure of this; most of it is unrelated (post-disconnect, climb selection, navigation, search). Don't cite that number as impact.

What carries this issue is the code gap, not the volume.

## The gap (verified on `main`)

- `packages/mobile/src/providers/bluetooth-provider.tsx:261` — `isWritingRef = useRef(false)` is the only in-flight record, deliberately a ref, so it never triggers a render.
- `BluetoothContextValue` (:47-101) exposes `isConnected`, `loading`, `connect`, `disconnect`, `sendFramesToBoard`, `clearBoard`, `reassertWall`, `undoWallChange`, `relightPresenceClimb`, `armUndoWallChangeToast` and two reconnect ids — **no sending/pending field**.
- The only send feedback is terminal: `hapticSuccess()` after `requestSendFramesToBoard(...)` returns `true` (~:444). Nothing between tap and confirmation.
- `loading` is connect-scoped only (set in `use-board-bluetooth.ts` `connect()`, cleared in its `finally`), and that's what `BleLightbulbButton`'s `isScanning` pulse renders. `ble-lightbulb-button-state.ts` has only two visual states — so once you're connected, every subsequent climb write is silent.
- The wait is non-trivial: `use-board-bluetooth.ts:419` documents that `RNBleAdapter` splits a packet into 20-byte chunks written sequentially with inter-chunk delays, and `writeChainRef` serialises writes across callers.

## Nothing breaks — this is a missing affordance

Re-taps are non-destructive: `pendingSendRef` (:262, :311-313) coalesces re-taps into a single trailing send, the `sendSignature` dedup (:413) skips byte-identical physical writes, and a lightbulb re-tap while connected routes to the benign `reassertWall`. The climb does get lit.

## Suggested fix

Promote `isWritingRef` to render state (or add a `sendInFlight` field to `BluetoothContextValue`), and drive a third visual state in `getBleLightbulbVisualState` that's distinct from the connect pulse.

## Severity

**P3.** Polish on a working path, backed by a single session's replay-derived evidence. Real and cheap to fix, but not core-flow breakage.

## Related

- #3924, #3991, #3925 — other BLE concerns, distinct from this.

marcodejongh:
## Summary

- Track queued and active foreground Bluetooth writes with a provider-scoped token store that survives stale releases across connection generations.
- Show a spinner in the connected lightbulb while a climb is being written, without rerendering the main Bluetooth context or changing tap and long-press behavior.
- Announce the busy state and write progress in English, Spanish, French, and German, with scanning taking visual and accessibility priority.

Closes #4145

## Release Notes

- The lightbulb now spins while Boardsesh writes a climb to your wall, so you know your tap landed.

## Test plan

- Rebased onto current `main` (auto-disconnect landed since): the provider conflict resolves by keeping main's `sendFramesToBoardWithActivityReset` wrapper and nesting the auto-sender inside `BluetoothWriteActivityProvider`. All four write-activity `reset()` teardown paths in `use-board-bluetooth.ts` survived the rebase intact.
- Dropped the #4213 cherry-pick this branch was carrying — that parity fix has landed on `main` and moved on since, and the stale copy was the merge conflict. The German `boards.json` now comes from `main` wholesale.
- Full mobile suite on the rebased branch: 5,278 tests across 604 files, all passing — including main's connection-lifetime and auto-disconnect tests running against the merged code.
- `vp run typecheck:mobile`, the `@boardsesh/i18n` catalog-parity suite (90 tests), `vp check` (no diagnostics in any file this PR touches), and `vp run check:mobile-bundle` (iOS + Android) pass locally.
- Added the German `ble.writing` string (`Board wird beleuchtet…`) — the `de` locale landed on `main` after this branch was cut.
- GitHub's OTA compatibility check confirms the iOS and Android fingerprints match `main`; this ships to existing builds without a native rebuild.
- **Fable BLE review: done.** The repository-policy review ran in a Fable 5 session and is recorded as a comment on this PR (verdict: approve; protocol-untouched, token/epoch lifecycle verified across all four teardown paths).

## Remaining QA gates (hardware only)

- Confirm real write timing, rapid queued sends, and teardown mid-write on Kilter, Tension, and MoonBoard hardware across iOS and Android.
- Confirm spoken feedback in VoiceOver and TalkBack on device.
## Fable BLE review

Reviewed per the CLAUDE.md rule that Bluetooth changes need a Fable review before merge. This is that review, running as Fable 5 in a Claude Code session — it replaces the "literal Fable review unavailable" blocker recorded earlier on this PR. **Verdict: approve**, with the rebase and follow-ups noted below.

### Protocol safety

No change to anything the board sees. Packet construction (`getAuroraBluetoothPacket`, `dispatchMoonboardPacket`), chunking, write types, abort signals, and the `writeChainRef` serialization are untouched — the write-activity store is JS bookkeeping wrapped around the existing chain. The MoonBoard failure-streak logic and the disconnect classification paths are unchanged. Zero risk to Kilter / Tension / MoonBoard write behaviour, and no fingerprint impact (JS-only, confirmed by the OTA gate).

### Correctness of the activity tracking

- Token ref-counting handles overlapping and queued writes; `notify` fires only on boolean edges, so subscribers rerender at most twice per busy episode.
- The epoch guard is right: a stale release surviving past a generation reset (adapter that ignores abort, late `finally`) becomes a no-op instead of clearing a new write's token. Idempotent release covers double-calls.
- `begin()` sits after the early no-adapter/no-config guard, so a skipped send never leaks a token; release lives in `finally`, so success, failure, and abort all release.
- Every teardown path resets the store: link-drop cleanup, fresh connect (new generation), deliberate teardown, and hook unmount. Verified again after rebasing onto the connection-lifetime refactor from current `main` — the reset calls sit in the refactored `clearConnectionAfterDrop(expectedAdapter)`, `teardownConnection(trigger)`, and the unmount cleanup, and the full mobile suite (5,130 tests) passes on the rebased branch.
- Render isolation holds: the store identity is stable (`useState` initializer), only the lightbulb leaf subscribes via `useSyncExternalStore`, and the main `BluetoothContext` value is untouched, so a write cannot rerender the navigation tree.

### UX judgment call, endorsed

The spinner covers queue-wait plus write (tap-to-settlement), so a rapid second tap shows a spinner for longer than its own physical write. That is the right reading of #4145 — the missing affordance was "did my tap land", and a queued tap has landed. Scanning keeps visual and announcement priority over writing, which is correct: connect feedback is rarer and more actionable.

### Accessibility

`accessibilityState.busy` now reflects scanning-or-writing, hint precedence is scanning → writing → long-press, the spinner itself is `accessible={false}`, and tap/long-press behaviour is unchanged while busy — all pinned by the render test.

### What I changed while finishing

1. **Rebased onto current `main`** over the connection-lifetime refactor; four conflict hunks in `use-board-bluetooth.ts` resolved by keeping main's adapter-guard semantics and re-adding the `reset()` calls.
2. **Added the German `ble.writing` key** (`Board wird beleuchtet…`, per the glossary) — the `de` locale landed after this branch was cut and catalog parity would have failed.
3. **PR #4213** fixes a pre-existing `de/boards.json` parity break on `main` (unrelated to this PR) that was redding `test-default` for every branch.

### Review follow-ups (applied after the review above)

- **Blocking item cleared.** Rebased onto `main` over the auto-disconnect work and dropped the stale #4213 cherry-pick, so `de/boards.json` is now main's version verbatim. The branch is `MERGEABLE`.
- **Spinner nit fixed.** The lightbulb no longer pins the spinner to `size="small"` while the icon it replaces honors `size`. `getBleLightbulbSpinnerSize` maps the icon size onto RN's two native spinner sizes (20pt / 36pt, crossover at 32) rather than passing the raw number — on iOS a numeric `size` only resizes the wrapper box and leaves the glyph unscaled, so passing it through would have looked worse, not better. Both shipping call sites pass 24 and keep the 20pt spinner; covered by four new cases in `BleLightbulbButton.test.ts`.

### Still open (hardware-only)

Real-board timing on Kilter / Tension / MoonBoard across iOS and Android, and spoken feedback under VoiceOver / TalkBack, remain hardware QA — nothing in this review substitutes for them, but none of them block a JS-only OTA-revertible change of this shape.

### DIFF
--- a/packages/mobile/src/components/ble/BleLightbulbButton.tsx
+++ b/packages/mobile/src/components/ble/BleLightbulbButton.tsx
@@ -8,10 +8,16 @@ import Animated, {
   withTiming,
 } from 'react-native-reanimated';
 import { Icon } from '../Icon';
+import { ActivityIndicator } from '../ActivityIndicator';
 import { hapticLight, hapticMedium } from '../../lib/haptics';
 import { useTheme } from '../../providers/theme-provider';
+import { useBluetoothWriteInProgress } from '../../providers/bluetooth-write-activity';
 import { timing } from '../../theme/animations';
-import { getBleLightbulbAccessibilityHint, getBleLightbulbVisualState } from './ble-lightbulb-button-state';
+import {
+  getBleLightbulbAccessibilityHint,
+  getBleLightbulbDisplayMode,
+  getBleLightbulbVisualState,
+} from './ble-lightbulb-button-state';
 
 type BleLightbulbButtonProps = {
   isConnected: boolean;
@@ -34,6 +40,8 @@ type BleLightbulbButtonProps = {
   /** Subtle visual warning that auto-disconnect is in its final ten seconds. */
   autoDisconnectWarning?: boolean;
   scanningAccessibilityHint?: string;
+  /** Announces the foreground wall write while preserving the button's action label. */
+  writingAccessibilityHint?: string;
   /** Hint describing the long-press action (e.g. "Hold to disconnect"). */
   longPressAccessibilityHint?: string;
   haptic?: 'light' | 'medium' | 'none';
@@ -57,12 +65,14 @@ export function BleLightbulbButton({
   accessibilitySelected,
   autoDisconnectWarning = false,
   scanningAccessibilityHint,
+  writingAccessibilityHint,
   longPressAccessibilityHint,
   haptic = 'light',
   size = 24,
   containerSize = 44,
 }: BleLightbulbButtonProps) {
   const { systemColors, brandColors } = useTheme();
+  const isWriting = useBluetoothWriteInProgress();
   const pulseOpacity = useSharedValue(1);
   const warningPulse = useSharedValue(0);
 
@@ -113,6 +123,7 @@ export function BleLightbulbButton({
     connectedColor: brandColors.warning,
     disconnectedColor: systemColors.secondaryLabel,
   });
+  const displayMode = getBleLightbulbDisplayMode(isScanning, isWriting);
 
   return (
     <AnimatedPressable
@@ -122,10 +133,12 @@ export function BleLightbulbButton({
       accessibilityLabel={accessibilityLabel}
       accessibilityHint={getBleLightbulbAccessibilityHint(
         isScanning,
+        isWriting,
         scanningAccessibilityHint,
+        writingAccessibilityHint,
         longPressAccessibilityHint,
       )}
-      accessibilityState={{ selected: accessibilitySelected ?? isConnected }}
+      accessibilityState={{ selected: accessibilitySelected ?? isConnected, busy: displayMode !== 'idle' }}
       hitSlop={8}
       style={({ pressed }: PressableStateCallbackType) => [
         styles.container,
@@ -139,7 +152,11 @@ export function BleLightbulbButton({
         pressed && styles.pressed,
       ]}
     >
-      <Icon name={visualState.iconName} size={size} color={visualState.iconColor} />
+      {displayMode === 'writing' ? (
+        <ActivityIndicator accessible={false} size="small" color={brandColors.warning} />
+      ) : (
+        <Icon name={visualState.iconName} size={size} color={visualState.iconColor} />
+      )}
     </AnimatedPressable>
   );
 }
--- a/packages/mobile/src/components/ble/BleLightbulbButton.tsx
+++ b/packages/mobile/src/components/ble/BleLightbulbButton.tsx
@@ -16,6 +16,7 @@ import { timing } from '../../theme/animations';
 import {
   getBleLightbulbAccessibilityHint,
   getBleLightbulbDisplayMode,
+  getBleLightbulbSpinnerSize,
   getBleLightbulbVisualState,
 } from './ble-lightbulb-button-state';
 
@@ -153,7 +154,7 @@ export function BleLightbulbButton({
       ]}
     >
       {displayMode === 'writing' ? (
-        <ActivityIndicator accessible={false} size="small" color={brandColors.warning} />
+        <ActivityIndicator accessible={false} size={getBleLightbulbSpinnerSize(size)} color={brandColors.warning} />
       ) : (
         <Icon name={visualState.iconName} size={size} color={visualState.iconColor} />
       )}
--- a/packages/mobile/src/components/ble/__tests__/BleLightbulbButton-render.test.tsx
+++ b/packages/mobile/src/components/ble/__tests__/BleLightbulbButton-render.test.tsx
@@ -121,7 +121,6 @@ describe('BleLightbulbButton write activity', () => {
 
   it('keeps the scanning pulse/icon and hint ahead of write feedback', () => {
     const store = createBleWriteActivityStore();
-    store.begin();
     const view = render(
       createElement(
         BluetoothWriteActivityProvider,
@@ -136,11 +135,17 @@ describe('BleLightbulbButton write activity', () => {
         }),
       ),
     );
+    let release = () => {};
+    act(() => {
+      release = store.begin();
+    });
 
     const button = view.getByRole('button');
     expect(view.container.querySelector('[data-spinner="true"]')).toBeNull();
     expect(view.container.querySelector('[data-icon="lightbulb.fill"]')).toBeTruthy();
     expect(button.getAttribute('data-busy')).toBe('true');
     expect(button.getAttribute('data-hint')).toBe('Scanning for boards nearby');
+
+    act(() => release());
   });
 });
--- a/packages/mobile/src/components/ble/__tests__/BleLightbulbButton-render.test.tsx
+++ b/packages/mobile/src/components/ble/__tests__/BleLightbulbButton-render.test.tsx
@@ -1,0 +1,146 @@
+// @vitest-environment jsdom
+import { act, fireEvent, render } from '@testing-library/react';
+import { createElement, type ReactNode } from 'react';
+import { beforeEach, describe, expect, it, vi } from 'vitest';
+
+type PressableMockProps = {
+  children?: ReactNode;
+  onPress?: () => void;
+  onLongPress?: () => void;
+  accessibilityLabel?: string;
+  accessibilityHint?: string;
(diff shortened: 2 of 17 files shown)


marcodejongh:
Addressed the minor test-style note in `af5ab336e`: the scanning-precedence test now renders first, begins the write inside `act()`, observes the subscribed update while scanning still wins, and releases the token inside `act()`. The focused render test passes 2/2, scoped `vp check`, `git diff --check`, and the mandatory pre-commit hook pass.

The longer spinner duration for queued requests is intentional: it represents tap-to-settlement feedback, including time waiting behind an earlier wall write, so the climber never sees a silent accepted tap. Physical rapid-queue timing remains a documented hardware QA gate. Literal Fable review also remains unavailable and blocking by policy.

packages/shared/i18n/locales/de/boards.json
marcodejongh, line 1: This cherry-picked file is now stale and is the source of the PR's merge conflict. The #4213 content landed on `main` and has since moved further: `main` now has keys this version lacks (`boards.edit.duplicate.*`, `limitReached` in both create flows, `boardEntity.edit.duplicate*`) and newer wording for several strings (`gymNone`, `clearLocation`, the `snapshotRetrying`/`pagedFallback*` family, the `gymPicker` strings). Merging this version would delete those keys and fail `catalog-completeness`.

Resolution is mechanical: rebase onto `main` and drop commit `464c999` entirely, taking `main`'s `de/boards.json` wholesale. Keep `3364471` — the `ble.writing` German key is this PR's own string and is still needed.

---

packages/mobile/src/components/ble/BleLightbulbButton.tsx
```diff
@@ -124,7 +137,11 @@ export function BleLightbulbButton({
         pressed && styles.pressed,
       ]}
     >
-      <Icon name={visualState.iconName} size={size} color={visualState.iconColor} />
+      {displayMode === 'writing' ? (
+        <ActivityIndicator accessible={false} size="small" color={brandColors.warning} />
```
marcodejongh: Nit, non-blocking: the spinner hardcodes `size="small"` while the icon it replaces honors the `size` prop. Both current call sites pass `size={24}` (small ≈ 20pt, close enough), but a future caller passing a larger `size` would get a visibly undersized spinner swap. Fine to leave as-is for now given the call sites.

---

marcodejongh (commented):
Reviewed the full diff at `464c999`, including the BLE-policy angle (this review ran as Fable 5). **Verdict: the write-activity change itself is sound and I'd approve it — but the branch can't merge as-is.** One blocking, mechanical item: the cherry-picked `de/boards.json` is stale against `main` and is the current merge conflict.

### Blocking: stale `de/boards.json` cherry-pick (the merge conflict)

`mergeable_state` is `dirty`, and a local test merge confirms the only conflicting file is `packages/shared/i18n/locales/de/boards.json`. The #4213 parity fix this branch carries has landed on `main` and evolved further — `main`'s version now contains keys the PR's copy lacks (`boards.edit.duplicate.*`, `limitReached` ×2, `boardEntity.edit.duplicate*`) plus newer wording. Rebase onto `main`, drop commit `464c999` (take `main`'s file), keep `3364471` (the German `ble.writing` key — that one is this PR's own). Details in the file comment.

### Core change: verified correct

I re-derived the safety claims from the diff rather than taking the PR body's word for them:

- **Protocol untouched.** Packet construction, chunking, abort, and `writeChainRef` serialization are unchanged; the store is JS bookkeeping around the existing chain. No fingerprint impact (OTA gate concurs).
- **Token/epoch lifecycle holds.** `begin()` runs after the early no-adapter guard; release lives in `finally` of the awaited `queuedSend`, so success, rejection, and abort all release. The epoch guard correctly turns a late release from a pre-reset generation into a no-op, and release is idempotent. `reset()` is called on all four teardown paths (link-drop cleanup, fresh-connect generation reset, deliberate teardown, unmount) with deps updated accordingly — and the store identity is stable (`useState` initializer), so the new dep can't retrigger those effects.
- **Notify fires only on boolean edges**, and only the lightbulb leaf subscribes via `useSyncExternalStore`; the main `BluetoothContext` value is untouched. The provider test pins the render isolation.
- **Accessibility:** `busy` reflects scanning-or-writing, hint precedence scanning → writing → long-press, spinner `accessible={false}`, press behavior unchanged while busy — all pinned by tests. Both existing `BleLightbulbButton` call sites (`CreateDrawerHeader`, `PlayDrawerActionBar`) got `writingAccessibilityHint`; there are no others.
- **i18n strings follow the glossaries:** de "Board wird beleuchtet…" (beleuchten, not senden), fr "Allumage de la board…" (allumer, not envoyer), es "plafón".

One correction to the earlier automated review on this thread: the claimed test gap ("no coverage of `sendFramesToBoard` without a `writeActivityStore`") doesn't exist — most pre-existing send tests still run the hook without a store, so the optional-chaining path is exercised throughout the suite.

The spinner-covers-queue-wait behavior (tap-to-settlement) is the right reading of #4145; a queued tap has landed, and that's exactly the feedback that was missing.

### Non-blocking

- Spinner hardcodes `size="small"` and ignores the `size` prop (inline comment).
- Hardware QA gates (real-board timing on all three board types, VoiceOver/TalkBack) remain open as the PR states; none block a JS-only, OTA-revertible change.

Once the rebase drops the stale i18n commit and CI is green, this is good to merge.

---

PR #4159, opened 2026-08-01
fix(sync): let admins release location freezes

### LINKED ISSUE #3860: No admin/API path to clear sync_frozen_at once a gym or board is frozen
## Context

PR #3845 added a `sync_frozen_at` timestamp on `gyms` and `user_boards`. Once a human curates a row (edit, claim, or delete via `updateGym`/`deleteGym`/`updateBoard`/`deleteBoard`/`applyGymClaim`), it gets stamped and the location sync (and the dedupe/merge tooling) permanently skips overwriting its metadata.

There is currently no mutation or admin-console path to clear `sync_frozen_at`. The only way to un-freeze a row today is a direct database write.

## Why this matters

If a moderator's edit was itself wrong (bad data entry, an incorrect claim, a SYSTEM revert that should be allowed to re-sync), there's no in-app recovery path — someone has to reach for raw SQL against the shared Postgres.

## Suggested follow-up

Add an admin-only mutation (likely alongside the existing gym-admin / dedupe review tooling from #3699) that clears `sync_frozen_at` on a gym or board, so a bad freeze can be undone without a manual DB write.

Raised as a non-blocking follow-up during review of #3845.

marcodejongh:
## Summary

- add a global-admin queue for frozen location-sync gyms and boards, with search, pagination, detailed confirmation, and auditable release reasons
- release only the exact locked frozen row the admin reviewed, retain immutable actor/owner snapshots after account deletion, and reject stale confirmations, merged gyms, and missing targets
- let later source syncs restore eligible deleted aliases in place without minting twins, and prevent older list responses from visually resurrecting a released row

Closes #3860.

### What the mutation actually checks

`clearLocationSyncFreeze` rejects on three things: the target is gone (`NOT_FOUND`), the gym has been merged away (`LOCATION_SYNC_TARGET_MERGED`), or the `expectedSyncFrozenAt` the admin confirmed no longer matches the row (`LOCATION_SYNC_FREEZE_STALE`). An already-released row returns `ALREADY_UNFROZEN` rather than erroring, so a retry is safe.

It does **not** query `gym_claims`. Ownership shows up two other ways. The list query joins approved claims purely to render the "Owner protected" warning before an admin commits. And a claim approved while the confirmation dialog is open stamps a fresh `sync_frozen_at` on the gym (`gym-claims.ts`), which trips the staleness check on submit — so the claim is caught, but indirectly, through the timestamp rather than a claims lookup. The gym-owner guard inside the sync writer is what actually stops an upstream takeover afterwards.

## Release Notes

## Test plan

Run on this branch, rebased onto `main` at 22f92b130:

- `vp test run --project backend location-sync-freezes` — 11 passed (resolver authorization, audit, staleness, actor validation)
- `vp test run --project web location-sync-freezes-panel check-admin global-admin` — 10 passed
- `vp test run --project location-sync` — 48 passed, 4 skipped (the real-DB freeze integration file self-skips without a database; the dedicated CI job runs it)
- `vp test run --project i18n` — 90 passed (catalog parity across `es` / `fr` / `de`)
- `vp run typecheck` — clean across every package
- `vp run check:i18n` (504 files, no hardcoded strings), `vp run check:i18n:orphans` (4487 keys, all referenced)
- `vp run check:db-migrations` — 191 migrations, journal consistent
- `vp lint` and `vp fmt --check` over every file this branch touches — clean

Whole-repo `vp check` is red on `main` today (a pre-existing lint backlog in ~70 files, none of them touched here). PR CI diff-lints, so that backlog is out of scope for this branch.

### German catalog + a red `main`

The admin strings landed in `en-US`, `es` and `fr` only, so `de/admin.json` was 44 keys short and `catalog-completeness.test.ts` would have gone red on merge. All 44 are translated per `docs/i18n-german-glossary.md` (informal du, `das Board`, no `senden` for an ascent), with every ICU placeholder preserved.

`de/boards.json` was separately 30 keys behind `en-US` on `main` already — `main`'s own CI is failing on that exact test right now, and every branch that touches an i18n catalog inherits the failure. Brought back to parity in the same pass.

## Manual gates

- verify global-vs-scoped admin access, live and deleted releases, stale and double-submit behaviour, and an in-flight list refresh
- confirm a real follow-up MoonBoard location sync restores a released row. MoonBoard location syncs are run by hand today, so this is a manual run, not a wait
- spot-check the German admin queue at `/de/admin/location-sync`


marcodejongh:
Fable review pass — two critical findings, one minor, all addressed.

**1. German catalog was missing all 44 new `locationSync` keys (critical) — fixed.**
`en-US`, `es` and `fr` got the freeze-queue strings; `de/admin.json` did not. PR CI stayed green only because `test-default` diff-scopes with `--changed`, and `catalog-completeness.test.ts` reads catalogs off disk rather than importing them, so it never entered the changed set. The unfiltered push-to-`main` run would have gone red. All 44 keys are translated per `docs/i18n-german-glossary.md` — informal *du*, *das Board* for the device, no *senden* for an ascent, `Halle` for gym — with every ICU placeholder (`{{name}}`, `{{shown}}`, `{{total}}`, `{{count}}`) preserved verbatim. `vp test run --project i18n` now passes 90/90.

While confirming that, `de/boards.json` turned out to be 30 keys behind `en-US` independently of this branch — [`main`'s own CI is failing on that exact test right now](https://github.com/boardsesh/boardsesh/actions/runs/31222889396), and any branch touching an i18n catalog inherits the failure. Brought it to parity in the same pass rather than shipping a knowingly-red PR.

**2. PR was based on the stacked `fix/3534` branch, not `main` (critical) — fixed.**
Merging as-is would have landed this work inside PR #4052's feature branch and dragged its pending-sign-off production-data migration along. Retargeted the base to `main` and rebased the six location-freeze commits onto `origin/main` with `git rebase --onto`, dropping every commit from the branches below in the stack. `vp run db:renumber` then moved the migration from `0194` to `0190` (main's tail is `0189`) and regenerated `meta/0190_snapshot.json` against main's latest snapshot — the SQL body is byte-identical to what was reviewed. `vp run check:db-migrations` reports 191 migrations, journal consistent.

**3. Frozen-entity count and page were two non-transactional reads (minor) — fixed.**
`hasMore` now comes from a `limit + 1` probe on the page query instead of comparing the page length against the separately-queried total, so a freeze released between the two statements can no longer skew the "Load more" affordance. `totalCount` stays a best-effort display number — the panel refetches after every mutation.

**Validation on the rebased branch:** `--project backend location-sync-freezes` 11 passed, `--project web location-sync-freezes-panel check-admin global-admin` 10 passed, `--project location-sync` 48 passed / 4 skipped, `--project i18n` 90 passed, full `vp run typecheck` clean, `vp run check:i18n` + `check:i18n:orphans` clean, `vp lint` / `vp fmt --check` clean on every file touched.

PR #4160, opened 2026-08-01
fix(board-presence): show recent wall senders

### LINKED ISSUE #3504: Ipad ui on the wall ui confusing
It says "sent by X" but that suggests that person send it, it should instead say "lit up by X".
Then in addition to that, we should a sent by byline below the litup by, that shows the avatars of the users that send it recently on this board_id

marcodejongh:
## Summary

- keep “Lit by” for the climber who put the problem on the wall and add a separate “Sent by” row with the five latest distinct successful senders for the exact board, climb, and angle
- resolve canonical and aliased climb IDs on the backend, refresh the row after tick changes, and discard stale results when wall history changes quickly
- never shrink the board at the production 11-inch portrait and 13-inch landscape panes, with shared quantized breakpoints for measurement and rendering
- throttle the stats-driven refresh to one request per 2s so a busy wall cannot push a kiosk at its rate-limit ceiling, while climb switches stay instant

Closes #3504.

## Release Notes

- See the five most recent senders right below the climber who lit the wall.

## Test plan

1. Open a board's wall kiosk on an iPad.
2. Under "Lit by", a "Sent by" row lists recent senders.
3. Log a send there; your avatar joins that row.
4. Switch climbs; the row swaps to the new senders.
5. Change the angle; the row follows that angle.

## Risk

Risk: 3/5 — new GraphQL resolver plus a wall-kiosk layout change. No migration, no BLE, no native change (rides OTA). Worst case is a missing or stale byline on the kiosk; the board art and controls around it are unaffected.

## Validation

- mobile suites: `vp run test:mobile`, 8139 passed (788 files)
- shared hook suites: `--project board-presence-react`, 82 passed — includes the three throttle cases (a stats event refetches, a ten-tick burst collapses to one request, a climb switch skips the throttle)
- backend against a real DB: `--project backend` on `board-presence` + `postgres-timestamps`, 107 passed — covers the alias subquery and the unparseable-timestamp cases
- `vp run typecheck:mobile` / `:backend` / `:shared`, `vp run check:mobile-bundle`
- `vp run check:i18n:orphans`: 3969 catalog keys all referenced, 4847 references all resolving
- i18n catalog parity: `session.json` matches en-US across es, fr, and de
- `vp check` reports nothing on any file this PR touches; pre-commit hook passed
- independent adversarial review: no remaining P0-P2 findings
- production geometry assertions:
  - 738×1194 at board AR 0.56 and 1.81: band/noncompact, board never below pre-feature area, chrome never above it
  - 1270×1024 at board AR 1.81: band/noncompact, 99.33% board area, no nominal clipping
  - band content floor, pre-feature → now: 398 → 362 (11" portrait), 431 → 395 (11" landscape), 316 → 360 (narrow stacked, unchanged by this commit), three-column unchanged
- `vp run dev:mobile` loaded `.boardsesh/qa-notes.md` and the wall kiosk was exercised on device

## Notes

No schema change: `boardClimbRecentSenders` reads through the existing
`boardsesh_ticks_board_climbed_at_idx` (board_id, climbed_at), which covers the
query at current tick volume. There is no migration in this PR, so nothing to
renumber and no lock window to schedule.

Sender identity on the row (id, name, avatar) is exposed to anonymous viewers of
public and system-shared boards on exactly the terms the existing
`boardPresenceStats` hardest-send surface already uses — same anon masking, same
visibility gate. Private-board visibility is unchanged.

`climbed_at` is `timestamp without time zone`, so the aggregate goes through
`parsePostgresUtcTimestamp` rather than a bare `new Date()`; its suite pins `TZ`
to a +10 zone, because under UTC the correct and the naive parse agree.

The stats subscription is still what keeps the row live, but it no longer maps
one tick to one request. Stats-driven refreshes go through a 2s
leading-edge-plus-trailing throttle: the first send after a quiet spell paints
immediately, a burst of ticks lands one refetch, and the ceiling is 30/min
against a 60/min budget. A climb switch bypasses it. The in-flight guard is the
generation counter plus a mounted ref rather than a per-effect flag, so
scheduling a refresh cannot cancel a request already on the wire.

The alias fan-out is a subquery inside the ticks query rather than its own
round-trip, so the resolver is two DB calls per kiosk refresh instead of three
and a heavily-merged climb never ships its alias list through the client.

The resolver over-fetches three rows past the five it returns and slices back,
because the unparseable-timestamp filter runs after the `LIMIT` — fetching
exactly five would return four whenever a corrupt `climbed_at` landed in the
top rows, with valid senders sitting just under the cut.

`board-presence-react` now sets `"types": ["node"]` in its tsconfig — the same
renderer-agnostic set `@boardsesh/board-react` uses — so the throttle's
`setTimeout` resolves without a DOM lib in a package that has to run on React
Native.

## Release gates

- confirm real-device safe areas and larger accessibility text on the 1270×1024 history layout
- re-confirm the 11-inch panes on device: the two-column band is 36pt shorter than it was, because it no longer funds the Lit-by driver row it hoists into its own column. The board is aspect-constrained at those panes so it does not grow — the height goes back to the band. Worth eyes on before merge, since the earlier device pass ran against the taller band.


marcodejongh:
Fable review pass. Three findings, all addressed; branch rebased on `main` (it was 110 commits behind, so the previously-green CI was stale).

**1. `boardPresence.sentByLabel` missing from the German catalog** (major) — fixed.
The `de` locale landed on `main` after this branch forked, so the key shipped to en-US, es, and fr only and German fell back to English copy. Added `"sentByLabel": "Getoppt von"` per `docs/i18n-german-glossary.md` (the glossary maps "Sent by {{name}}" → "Getoppt von {{name}}"; never "Gesendet"). `session.json` now matches en-US across all three enforced locales.

One correction to the finding's mechanism: this would not have red-flagged PR CI. `catalog-completeness.test.ts` reads catalogs with `readFileSync`, so Vitest's `--changed` module graph never relates a locale JSON diff to it — `vp test run --changed origin/main --project i18n` selects zero files. It runs unfiltered on pushes to `main`, so the gap would have surfaced after merge rather than here. Fix stands either way; a missing key ships untranslated copy.

**2. `lastSentAt` parsed in server-local time** (major) — fixed.
`climbed_at` is `timestamp without time zone` in Drizzle `mode: 'string'`, so the driver returns a naive `2026-07-03 10:00:00` and the bare `new Date(...)` resolved it against the host's zone — every instant shifted by the server's offset off-UTC. `stats.ts` already had a private `parsePostgresUtcTimestamp` for exactly this; lifted it to `packages/backend/src/utils/postgres-timestamps.ts` (beside `postgres-errors.ts`) so both callers share one implementation instead of the resolver bypassing the convention.

Added a suite that pins `TZ` to `Australia/Brisbane` for its duration — under UTC the broken and the correct parse agree, so an unpinned test can't fail. It includes a guard-the-guard case that asserts the TZ pin actually took effect. Mutation-checked: reverting the `Z` suffix fails 2 of the 6 cases.

**3. PR body described a migration and index that aren't in the diff** (minor) — fixed.
Dropped the ordinal-0189 collision gate and the non-concurrent-index lock window; there are no `packages/db` changes at all (the migration was added in the first commit and removed in the second — the net diff is clean). Replaced them with a note that the query reads through the existing `boardsesh_ticks_board_climbed_at_idx`. Softened the anonymous-visibility gate to a note, since sender id/name/avatar on public boards mirrors the existing `boardPresenceStats` hardest-send surface rather than introducing a new policy call. Also removed the dev-tunnel Metro URL from the test plan.

**Validation:** `vp run typecheck`, `vp run test:mobile` (5128 passed), backend board-presence + new timestamp suites (91 passed), board-presence-react suites, `vp run check:mobile-bundle`, `vp check` clean on every file this PR touches.

**Unrelated, pre-existing on `main`:** `de/boards.json` is missing ~19 keys that en-US has (`boardForm.create.*`, `mobile.gymPicker.*`, `mobile.offline.*`), so `i18n catalog completeness > de > boards.json` fails on `main` today. Not touched by this PR and left alone to keep the diff honest — worth its own fix.

marcodejongh:

marcodejongh:

Conflicting: packages/backend/src/__tests__/board-presence.test.ts and i18n session.json (en-US/es/fr) — this feature has diverged from main.

marcodejongh:
Worked the outstanding review feedback. Two commits: `87fe467` (the two blockers plus the round-trip cleanup) and `c40185e` (the geometry naming). Branch is on current `main` — no rebase needed this time, the September conflict is gone.

**1. `parsePostgresUtcTimestamp` throws on an unparseable value** (blocking) — fixed.
The signature promised `string | null` but `toISOString()` raises `RangeError` on an Invalid Date, so a single corrupt `climbed_at` would have failed the whole recent-sender aggregate rather than dropping one row. Guarded on both paths — the string path and the `Date` path, which had the same hole. Three cases added: a non-date string, an out-of-range one (`2026-13-45 99:99:99`), and `new Date('nope')`.

**2. Unbounded refetch on every tick event** (blocking) — fixed, as a throttle rather than the suggested debounce.
A plain trailing debounce delays *every* refresh by its window, including the first send after a quiet wall — which is the one a climber is standing there waiting to see. A 2s leading-edge-plus-trailing throttle gives both: the first stats event after a quiet spell goes out immediately, and a burst collapses into one request per window. Ceiling is 30/min against the 60/min budget, from ~1/tick before. Climb switches skip the throttle entirely and stay immediate.

One thing that fell out of this: the in-flight guard had to move off the per-effect `active` flag onto the generation counter plus a mounted ref. With the old flag, a stats event arriving while a climb-switch request was still on the wire ran the effect cleanup and orphaned that request, so the row would have waited an extra window for a response it already had coming. Three tests cover it: a stats event refetches, a ten-tick burst inside one window lands one request, and a climb switch does not wait.

**3. Three sequential round-trips per call** (raised in four reviews) — down to two.
The alias fan-out is a subquery inside the ticks query now instead of its own `SELECT` round-trip, which also means a heavily-merged climb no longer ships its alias list back through the client to be spread into an `inArray`. `resolveCanonicalClimbUuid` stays its own call — it is the shared helper every alias-aware read goes through, and folding it in here would fork that convention for one resolver.

**4. `mobile.boardPresence.setByLine` vs `boardPresence.*` prefix** — verified, no change.
Both prefixes predate this PR on `main` (`boardPresence.open` / `railDark` / `openAria*` were already flat, `mobile.boardPresence.setByLine` already nested), and `mobile.*` is the convention for 906 keys across the mobile app. Resolved all four locales by hand — en-US "Set by", es "Abierto por", fr "Ouvert par", de "Gebaut von" — and `vp run check:i18n:orphans` passes: 3969 catalog keys all referenced, 4847 references all resolving in a mobile-safe namespace. Renaming it would be a repo-wide convention change, not this PR's call.

**5. `bandContentFloor` magic numbers** — fixed in `c40185e`.
The fixed parts of the band's vertical budget are named now: the gaps (`TIGHT_GAP` / `BLOCK_GAP`, the spacing[2] / spacing[3] tokens the components lay out with), the fixed rows (`ON_WALL_ROW`, `BAND_BACK_TO_LIVE_ROW`, `ATTRIBUTION_ROW_MIN`), and the paddings. Values are byte-identical — the production-geometry assertions pin the resulting heights and are unchanged.

**Not from the review, but it fell out of the work:** `board-presence-react` had `lib: ["ES2022"]` and no `types`, so `setTimeout` did not resolve. Added `"types": ["node"]`, matching what `@boardsesh/board-react` (the other renderer-agnostic React package) already does — timers without dragging a DOM lib into a package that has to run on React Native. Worth flagging separately: `vp run typecheck:shared` was green while `tsc --noEmit` inside the package was not, so that task does not appear to cover this package. Not fixing that here.

Also left the type-narrowing guard after `canFetch` alone, contra an earlier review that called it dead code — `canFetch` is a `boolean`, so the compiler still needs those checks to call through `boardId` / `angle` / the optional client method. Added a comment saying so, since it has now been read as dead twice.

**Validation:** `vp run test:mobile` 8139 passed (788 files), `--project board-presence-react` 82 passed, backend `board-presence` + `postgres-timestamps` against a real DB 107 passed, `vp run typecheck:mobile` / `:backend` / `:shared` clean, `vp run check:i18n:orphans` clean, `vp check` reports nothing on any file this PR touches.

https://claude.ai/code/session_01K1SgJQq34zoLE7UHHmZAjD

marcodejongh:
Second pass, on the latest review. `53ab875`.

**1. Angle validation diverges from `BoardPresenceAngleSchema`** — fixed, and it was a real bug rather than a documentation gap.

The review's softer read ("the PR notes explain it — mirror state not carried") does not match what was in the diff: there was no such note, and no reason for the narrower bound. `BoardPresenceAngleSchema` sits four lines above with `-90..90` and the comment "Aurora supports negative tilt", and `reportBoardClimb` has a test asserting it accepts `-5`. So a board on negative tilt logs its ticks at a negative angle and the new query refused to look them up.

The failure mode was also quieter than the earlier round described. It would not have thrown `"Invalid recent senders"` at the kiosk — the hook's `canFetch` guard mirrored the same `0` floor, so the request was never sent at all. A tilted wall would just show no "Sent by" row, with no error anywhere to trace it to. Both bounds are `-90..90` now, each carrying a note about the other so they cannot drift apart again. Backend: `-5` resolves, `-91` rejects. Hook: a negative angle fetches, `-91` skips.

**2. Result count can fall below 5 on corrupt timestamps** — acknowledged, no change.
Correct reading of the code. The `flatMap` drops unparseable rows after `.limit(5)`, so five rows in could be fewer out. It is the same shape the null-`climbed_at` case already had before this PR, and both are unreachable in practice — `max(climbed_at)` on a `timestamp` column comes back well-formed from the driver or not at all. The guard exists so a bad row degrades one entry instead of the whole aggregate, which is the trade I would rather have than a hard five.

**3. Incomplete `useTranslation` mock** — fixed.
The WallChromeRegion mock resolved only `mobile.boardPresence.setByLine`, so `boardPresence.litByLine` and `boardPresence.sentByLabel` fell through to raw key strings. Harmless today because WallAttributionBlock is mocked away, which is exactly why it was worth closing — the next test that renders attribution through WallChromeRegion would have asserted against key strings and passed.

**On the covering index** (raised again in the round before this one, dropped in the latest): that call is already made and recorded in the body. There is no `packages/db` change in this PR and no migration to renumber; the query reads through `boardsesh_ticks_board_climbed_at_idx`. The reviewer is right that `(board_id, climbed_at)` does not cover `climb_uuid` / `angle` / `status` — it is a board-scoped scan with a filter, which is fine at current tick volume and is the reason the index was pulled out in `bd222c9` rather than shipped as a table-locking migration. If wall traffic makes it matter, it wants a `CREATE INDEX CONCURRENTLY` in its own PR with its own deploy window, not a rider on this one.

**Validation:** backend `board-presence` 101 passed (was 100, +1 for negative tilt), wall-kiosk suites 64 passed, `--project board-presence-react` 83 passed, `vp check` clean on every file touched.

https://claude.ai/code/session_01K1SgJQq34zoLE7UHHmZAjD

marcodejongh:
Third pass. `d3adf8d`. Three of the four taken, one rejected with a test that proves it.

**1. `bandContentFloor` over-allocates the driver row in two-column bands** — correct, and fixed.

This one was the real find. `identityColumn` charged 36pt for the Lit-by driver row in every non-three-column band, but `WallChromeRegion` sets `separateAttribution = bandColumns >= 2` and passes `showAttribution={false}`, which hoists the whole attribution block into its own column. So from two columns up the identity column was paying for a row it does not render. The driver row is now funded only by the stacked (one-column) band, which still renders attribution inline.

Floors, pre-feature → now: **398 → 362** at 11-inch portrait, **431 → 395** at 11-inch landscape. Stacked and three-column paths unchanged.

One correction to the finding's framing: it is not invisible to CI because the assertions pin pixel heights rather than column heights — it was invisible because the assertion was `production.boardRect` *equals* the pre-sender baseline, and that equality still holds. The board is aspect-constrained at both 11-inch panes, so the 36pt goes back to the band rather than growing the board. I found that out by asserting the board got *bigger* and watching it fail on `299279.558 to be greater than 299279.558`. The assertions now pin what is actually true — board never shrinks, chrome never grows — which is a weaker claim than before but an honest one.

Flagging for QA rather than burying it: the two-column band is 36pt shorter than the one the earlier device pass ran against. Added to the release gates.

**2. LIMIT applied before timestamp filtering** — fixed, third time this was raised so it is fair to stop arguing it.
Over-fetches three rows past the five it returns and slices back. Costs nothing and removes the failure mode entirely, which beats my previous answer of "unreachable in practice".

**3. `mountedRef` initialization is redundant** — rejected. The re-seed is load-bearing.

The reasoning in the finding is about the mount window, and that part is right: there is no observable gap where the flag is false before mount. But that is not what the assignment is for. StrictMode runs the effect, tears it down, and runs it again **on the same component instance**, so the ref survives. A probe confirms the sequence is `setup ref=true` → `cleanup` → `setup ref=false`. Without the re-seed the flag latches false at the simulated unmount and every response after it is dropped on the floor — the row would just never populate in dev.

Pinned rather than asserted: there is now a StrictMode test, and deleting only the `mountedRef.current = true` line fails it. Deleting the whole effect does *not* fail it, which is presumably how it read as redundant — dropping the cleanup too means nothing ever sets the flag false, so the bug hides. Added a comment on the effect saying this.

**4. No documentation update** — added.
`docs/websocket-implementation.md` now lists `boardClimbRecentSenders` in the auth-optional read set and carries a paragraph on the alias fan-out, the index it actually reads through, the 60/min limit, and the client-side 2s throttle. Worth noting the first review round on this PR explicitly concluded no doc change was needed here; the later call is the better one.

**Validation:** `vp run test:mobile` 8139 passed, wall-kiosk 64 passed, `--project board-presence-react` 84 passed (+1 StrictMode), backend `board-presence` 101 passed against a real DB, `vp run typecheck:mobile` / `:backend` clean, `vp check` 0 errors.

https://claude.ai/code/session_01K1SgJQq34zoLE7UHHmZAjD

marcodejongh:
Fourth pass. `bb12d5d`. Two of the four code items taken; the other two do not hold up against the tree.

**1. No test for the empty-alias subquery path** — fair, added.
Nothing asserted what the `inArray` over an empty subquery does, and "generates valid SQL" is not the same as "matches the right rows". The new case queries a climb with no aliases and pins that the OR degrades to the canonical equality alone. Its decoy is a tick on a *different* climb by a *different* user on the same board, so an over-matching OR fails the assertion rather than passing quietly. Backend suite is 102 now.

**2. `twoColumnFloor` is a misleading name** — agreed, renamed to `baseFloor`.
It is the shared calculation the stacked band also builds on, so the old name read as if the one-column path were borrowing a two-column number. Comment reworded to match.

**3. i18n key rename may leave an orphaned catalog entry** — checked, nothing orphaned.
`mobile.boardPresence.sentByLine` is still in all four catalogs *and* still has a live reference: `NowOnTheWallPanel.tsx:1270` renders it on the board sheet, a different surface from the wall kiosk. There was no rename — the kiosk took a new key (`boardPresence.litByLine`) and the sheet kept the old one. That is why `check:i18n:orphans` passes rather than in spite of it: 3969 catalog keys, every one referenced.

**4. Stale comment in `wall-kiosk-layout.ts`** — already updated in this PR.
The header does mention the constraint. It reads "argmax picks the axis that shrinks the board least **at the stable design-preferred extents. Content floors size the chosen region afterward, so copy or an added byline cannot teleport a board between rail and band.**" That wording landed in `a6ad808`, the first commit here, alongside the `railAxisPref` / `bandAxisPref` change it describes.

**On the release gates and draft status:** correct that both are author-flagged rather than CI-verified, and the 11-inch one is mine — I added it in the last round precisely because the two-column band is now 36pt shorter than the band the earlier device pass ran against. That is a real thing to look at before merge. But this PR was already out of draft when I picked it up, and flipping it back is @marcodejongh's call, not something I will do unilaterally on a review's say-so. Flagging it here instead so the decision is visible.

**Validation:** backend `board-presence` 102 passed against a real DB, wall-kiosk 64 passed, `vp run typecheck:backend` clean, pre-commit hook clean.

https://claude.ai/code/session_01K1SgJQq34zoLE7UHHmZAjD

marcodejongh:
Fifth pass. `20e022e`. The finding as written does not hold, but checking it turned up a smaller real one, which is fixed.

**The stacked band does fund both rows.** The claim is that `BAND_DRIVER_ROW + TIGHT_GAP` in the return covers Lit-by and leaves Sent-by unbudgeted. It is the other way round: the Lit-by row is the `(columns === 1 ? BAND_DRIVER_ROW : 0)` term inside `identityColumn`, and the 36 + 8 in the return *is* the Sent-by row plus its gap. Stacked budgets 36 + 8 + 36 = 80pt for the pair. The existing `bandContentFloor(scale, 500) === preSenderFloor + 44` assertion says the same thing from the other side: the pre-sender floor already carried Lit-by, and the +44 is Sent-by plus the gap.

**What is actually wrong is the row *height* model, and it is 5pt, not 38.** The stacked path funded that row with a flat 36 while the three-column path measures the same row as `max(ATTRIBUTION_ROW_MIN, metaLineHeight) + ATTRIBUTION_ROW_PADDING`. Those agree at ordinary scales and diverge once heroScale inflates the meta line. Swept it:

```
shortSide=639 hero=1.0  metaLH=23  row=30  budgeted=36  short=0
shortSide=639 hero=1.4  metaLH=33  row=35  budgeted=36  short=0
shortSide=639 hero=1.8  metaLH=39  row=41  budgeted=36  short=5
```

A stacked band means `bandWidth < 640`, and `paneShortSide <= bandWidth`, so 639 is the top of the reachable range — the worst case is 5pt on a large external display narrow enough to stack. The row height is hoisted and shared now, and the stacked path takes the driver-row height as a floor (`Math.max(BAND_DRIVER_ROW, attributionRow)`) so nothing shrinks at the scales that already fit.

Agreed on the blast radius, for the record: every pane in the matrix is at least 738pt wide and resolves to two or three columns, so no production layout changes here and this needs no device re-QA.

Pinned both directions rather than just the fix: hero-scaled copy grows the row, copy inside the avatar floor keeps it at 36. Reverting the `max()` fails the first — checked.

**On the two non-blocking notes:** `contentFloorRail` is genuinely unset and the reasoning for why that is fine (rail height tracks the board, so the vertical budget is not the binding constraint) matches what I read. Leaving it; it predates this PR and constraining rail width is its own change. Thanks for confirming the StrictMode invariant and the empty-alias decoy read correctly.

**Validation:** wall-kiosk 65 passed, mutation-checked. Backend `board-presence` 102 and the rest unchanged by this commit.

https://claude.ai/code/session_01K1SgJQq34zoLE7UHHmZAjD

marcodejongh:
Sixth pass. `b6925ea`. Two taken, one rejected — and the rejected one would have made the copy wrong, so it is worth spelling out.

**1. No WallKioskScreen coverage of the `enabled` guard** — fair, and the reason it had none was itself the problem. The condition was a bare expression inside a component that pulls in react-native, so no plain unit test could reach it. Moved it to `wall-kiosk-layout.ts` as `shouldFetchRecentSenders` — that module is already the pure, react-native-free half of the kiosk, and its header says so — and pinned the truth table: fetch on `live` and `history` with non-compact chrome; skip when idle, when compact, and before the first measurement produces a layout.

**2. Over-fetch comment is misleading** — agreed, reworded. It now says Postgres applies the LIMIT and the drop happens in JS on the returned rows, instead of "the filter runs after the LIMIT", which did read as post-limit filtering inside the query.

**3. NowOnTheWallPanel terminology** — rejected. These are not the same concept.

`mobile.boardPresence.sentByLine` there labels `hardestSend.sentByDisplayName` (NowOnTheWallPanel.tsx:1246) — the climber who **sent** the hardest problem logged on the board. That is the hardest-send row, not a driver row. "Sent by" is the correct word for it, and switching it to "Lit by" would have made the row say the wrong thing about a real send.

The kiosk's `boardPresence.litByLine` labels `climb.sentByDisplayName` — whoever put the current problem **on the wall**, which may be nobody's send at all. Two different facts that happen to sit behind similarly-named schema fields. If anything reads oddly here it is the underlying `sentBy*` field naming on the climb, which predates this PR.

Worth noting this is the second round in a row where the review's premise about these keys did not survive a look at the call site — the round before had `sentByLine` down as orphaned when it has a live reference in this same file. Flagging the pattern, not the conclusion.

**Validation:** `vp run test:mobile` 8142 passed (+3), wall-kiosk 67 passed, `vp run typecheck:mobile` clean, pre-commit clean.

https://claude.ai/code/session_01K1SgJQq34zoLE7UHHmZAjD

PR #4165, opened 2026-08-01
fix: accessibility: clarify colour changes do not work for moon boards

ES-Alexander:
## Summary

Updates the description on the accessibility page to make it clear that custom LED colours do not work for Moon Boards, to reduce user confusion. Per [the protocol description](https://github.com/boardsesh/boardsesh/blob/main/docs/MOONBOARD_BLUETOOTH_PROTOCOL_SPEC.md#51-frame-structure) the board only receives hold types, so it's seemingly not possible for the app to request custom LED colours.

Undecided whether it's worth adding similar language to the preview caption (on line 53). I've kept the change as page-description-only for now, since that's at the top of the page, and the preview caption is already slightly ambiguous whether it's discussing how previews will display in the app or how climbs will display on the wall (so adding a moonboard disclaimer there could make things _more_ confusing without also rewording the rest of the caption).

## Release Notes

## Test plan

N/A - text only change.

### DIFF
--- a/packages/shared/i18n/locales/en-US/common.json
+++ b/packages/shared/i18n/locales/en-US/common.json
@@ -18,7 +18,7 @@
       "metroServersSubtitle": "Switch between Tailnet bundlers",
       "accessibility": {
         "title": "Accessibility",
-        "description": "Pick hold colours and marker shapes that are easier to tell apart. Colour changes also light up your board; shape, size, and brush only change the in-app overlay.",
+        "description": "Pick hold colours and marker shapes that are easier to tell apart. Colour changes also light up your board (except for MoonBoards); shape, size, and brush only change the in-app overlay.",
         "rowSubtitleShort": "Hold colours and shapes for colour vision",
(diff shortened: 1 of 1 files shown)


ES-Alexander:
1. It seems the `commit-lint` CI job doesn't work properly to restart in isolation, perhaps because the event name is based off the state of the initial CI run or something. Updated the title of the PR, and pushed with a matching commit message for good measure (and to re-run the whole CI pipeline) - hopefully resolved now...

2. The Claude review failure [is expected for PRs from forks](https://github.com/anthropics/claude-code-action/issues/542), and should likely be [skipped rather than enabled](https://github.com/trentmcnitt/cc-notifier/pull/14) to avoid opening up security boundaries

3. The `test_report` failure is less obvious to me (though seemingly some kind of permissions issue), and not something I can put more time into [investigating](https://stackoverflow.com/questions/70435286/resource-not-accessible-by-integration-on-github-post-repos-owner-repo-ac) at the moment

PR #4167, opened 2026-08-02
fix(web): unblock OG image crawling and stop indexing filtered climb lists

marcodejongh:
Part of the Expo Web Program phases 4–6. Two indexation leaks on the web front door.

## Every dynamic OG image was blocked from Googlebot-Image

`robots.ts` disallowed `/api/`, and `buildVersionedOgImagePath` returns *relative* `/api/og/…` paths — so the setter, profile, playlist and session share cards were all excluded from image crawling, as was `/api/internal/board-render`, the climb page's LCP image. Added `allow` rules for both.

The test asserts the allow rules are **more specific** than the `/api/` disallow, not that they appear first. Robots precedence is longest-path-match, so an ordering assertion would have passed on a file that doesn't work.

Climb OG cards were never affected — they use the absolute `ws.boardsesh.com/og/climb`.

## Filtered list URLs were an unbounded crawl space

`/list` with any of `?minGrade=`, `?sortBy=`, `?name=`, `?settername=` rendered an indexable, self-canonical page, so every filter permutation was its own crawlable URL. The legacy tree had no canonical at all. Both trees now emit `noindex, follow` with a canonical back to the clean base whenever a non-`page` param is present. Plain `?page=` pagination stays indexable.

`isUnlisted` is now selected in `board-slug-utils` and drives `noindex` separately — unlisted is link-only by design, which is a different thing from private and wants a different answer. Private boards are noindexed too, which is indexation only, not access control (see below).

## What this PR deliberately does *not* do

An earlier revision of this branch also added an access-control mask to `boardBySlug` plus token forwarding in `board-slug-utils`. That work is dropped — **#4087 already does it, and does it better**: it covers `board`, `boardBySlug`, `gym` and `gymBySlug`, teaches anonymous embeds to tell a real `NOT_FOUND` from a malformed payload, and updates the schema docs. It is green and mergeable.

Two things made keeping my version wrong rather than merely redundant:

- It conflicted with #4087 on both `boards.ts` and `board-slug-utils.ts` (`git merge-tree` → CONFLICT), so one of us was going to eat a manual resolution on a security-sensitive file.
- It restricted private-board reads to the owner and gym admins, where #4087 uses `isRowAnonReadable` — the anonymous-only pattern `board(boardUuid)` already follows on `main:798`. Mine would have silently broken private-board links shared between signed-in climbers. That is a product decision, not a security one, and it wasn't mine to make in an SEO PR.

`main`'s `anon-entity-masking.test.ts` pins the current behaviour on purpose, and my change red-lit it — which is the test doing its job.

So the `noindex` here is a crawl-hygiene layer that is correct either way: it keeps a private board out of the index today, and stays correct after #4087 makes it unresolvable anonymously. The route comments say so explicitly, so nobody reads them as the access control.

- `robots.txt` serves both allow rules and they out-specify the `/api/` disallow.
- A filtered `/list` URL on both trees emits `noindex, follow` and canonicalizes to the clean base; `?page=2` does not.
- An unlisted board's list and view pages are `noindex, follow` but still render.

## Release Notes

Share previews for setters, profiles, playlists and sessions can now be picked up by search engines, which they were silently blocked from before. Boards you've shared as unlisted stay out of search results, and filtered climb lists no longer compete with the clean board page for ranking.


marcodejongh:
This PR is Tier 0 of the web-reposition epic #4358 (it is W-10, the SEO quick-wins track): CI has been green and the automated review positive since 2026-08-02, and W-15 (#4369, SSR front doors) lists it as a dependency. Flagging for maintainer review/merge so the front-door work isn't blocked on it.

PR #4171, opened 2026-08-02
ci: add post-deploy production smokes for both hosts

marcodejongh:
Part of the Expo Web Program phases 4–6 (W-04). Post-deploy verification for both production hosts.

## www had no post-deploy check at all

`deploy-web` ended at `vercel deploy` and reported success. A deploy that built and uploaded cleanly but 500s on every request looked identical to a healthy one, and the first signal would have been a user.

`scripts/production-smoke.ts` now runs after the deploy and checks the surfaces that carry organic traffic or that other systems depend on:

| Check | Why it's here |
|---|---|
| `/` → 200 + a server-rendered `<h1>` | A spinner-only shell 200s. The `<h1>` is the cheapest proof SSR still works. |
| `/robots.txt` → 200 + `Sitemap:` | Silent loss of the sitemap pointer is invisible for weeks. |
| `/sitemap.xml` → 200 + ≥1 `<loc>` | An empty sitemap still parses. |
| `/api/auth/session` → 200 JSON | The SPA's identity provider. `www` down means nobody signs into the app. |
| `/api/internal/ws-auth` → 200 with `authenticated:false` | Kiosks and embeds open a presence socket with **no** session. If this 401s, every unattended gym screen goes dark and nobody is watching a kiosk to report it. |
| `/kiosk/{gym}`, `/embed/board/{uuid}` | 200 + `text/html` + a body-size floor, because a Next error shell is a well-formed 200 HTML document. Fixture-backed, skipped unless configured (below). |

Two deliberate safety properties:

- **It runs after the deploy, so it is a detector, not a gate.** It turns the job red and fires `notify-failure`, but it can never block a deploy from shipping. #3977 stopped production deploys for two days by being fail-closed; that cure was worse than the disease.
- **Missing fixture config skips, never fails.** `SMOKE_KIOSK_GYM_SLUG` / `SMOKE_EMBED_BOARD_UUID` are repo variables; unset means those two checks report `--`. A red deploy because someone forgot a variable is the same failure mode in a smaller costume. Set them to turn the checks on — no code change needed.

It's also skipped when an instant rollback is pinned: that path deploys with `--skip-domain`, so `www` is still serving the rolled-back build and smoking it would report on a deployment this job didn't promote.

## The app host could boot broken and still pass

`deploy-app-web`'s existing curl smoke (six checks, added in #4065) proves the transport layer — shell 200s, deep route falls back, WASM serves, hashed asset is immutable. It cannot prove the bundle *evaluates*. A JS chunk can serve with perfect headers and still throw on evaluation, and `_redirects`' `/* /index.html 200` means a missing route answers 200 too. Both look healthy to curl and blank to a user.

So there's now a real browser in front of it: navigate, wait for React to mount into `#root`, assert it rendered visible text, and fail on any uncaught `pageerror`.

This runs under a **separate** `playwright.production.config.ts`, not a new project in the existing config. That config declares `globalSetup`, which seeds data and signs the test user in against whatever `baseURL` resolves to — pointing it at production would write to production, and Playwright has no per-project opt-out for `globalSetup`. The only safe answer is a config that never declares one. It also declares no `webServer`, so it can't silently smoke localhost instead of the deploy.

## The homepage had no `<h1>`

Found by running the smoke against production before wiring it up — the first run failed on the homepage check. Verified against the live HTML: `www.boardsesh.com/` server-renders **zero `<h1>` and zero `<h2>`**, only an `<h5>`.

The hero title is `<Typography variant="h5">`, and MUI maps `variant` to a matching element, so it rendered a literal `<h5>`. The site's highest-traffic indexable page had no top-level heading for a crawler to read. Fixed with `component="h1"` — visually identical, semantically correct — plus a unit test, because nothing visual changes if it regresses.

This is adjacent to W-04's stated scope. I fixed it here rather than filing it because the alternative was weakening the smoke to match a broken page, and a check written around a known defect stops being a check.

- The rollback guard: with `check-rollback.outputs.active == 'true'`, the www smoke step skips.
- `scripts/production-smoke.test.ts` — each check must reject a realistic failure, not just accept a healthy response. The ws-auth case is the interesting one: a 200 that returns an error page instead of JSON, which a status-only check would wave through.
- Nothing in `e2e/production/` can reach `globalSetup`. The `chromium` project's `testIgnore` also excludes it, so a normal e2e run won't try it against localhost.

Run it by hand any time: `vp run smoke:production`, or `vp run smoke:production -- --base https://<preview>`.

## Release Notes

none

### DIFF
--- a/.github/workflows/production-deploy.yml
+++ b/.github/workflows/production-deploy.yml
@@ -436,15 +436,12 @@ jobs:
         env:
           SMOKE_KIOSK_GYM_SLUG: ${{ vars.SMOKE_KIOSK_GYM_SLUG }}
           SMOKE_EMBED_BOARD_UUID: ${{ vars.SMOKE_EMBED_BOARD_UUID }}
-        # `bunx tsx` rather than `vp run smoke:production`, matching this file's
-        # existing `bunx wrangler@4`. CLAUDE.md's no-bunx rule targets
-        # validation, which this isn't, and tsx resolves from the workspace that
-        # `bun install --frozen-lockfile` above already materialised. Reaching
-        # for vp here would mean curl-installing the toolchain onto the
-        # production deploy path — a network dependency that can red a healthy
-        # deploy after it has already shipped. `vp run smoke:production` remains
-        # the way to run this by hand.
-        run: bunx tsx scripts/production-smoke.ts
+        # Bun executes the TypeScript directly — no `bunx`, no `tsx`, nothing
+        # fetched at deploy time. `vp run smoke:production` is the same script
+        # and is how to run it by hand, but no deploy job has vp, and
+        # curl-installing the toolchain here would put a network dependency on
+        # the production deploy path that could red a deploy that shipped fine.
+        run: bun scripts/production-smoke.ts
 
   deploy-production-backend:
     needs: [detect-changes, check-rollback, build-backend, migrate]
--- a/.github/workflows/production-deploy.yml
+++ b/.github/workflows/production-deploy.yml
@@ -744,6 +744,11 @@ jobs:
         # This puts a real browser in front of the deploy and confirms React
         # mounted into #root. Read-only — it never signs in, and it runs under
         # playwright.production.config.ts, which declares no globalSetup.
+        #
+        # No rollback guard here, unlike the www smoke. Pages has no instant-
+        # rollback signal to read; the subdomain's equivalent is the job-level
+        # APP_WEB_DEPLOY_HOLD `if:` above, which skips this whole job — this
+        # step included — so there is nothing left to guard at step level.
         env:
           PLAYWRIGHT_TEST_BASE_URL: https://app.boardsesh.com
         run: cd packages/web && bunx playwright test --config=playwright.production.config.ts
--- a/.github/workflows/production-deploy.yml
+++ b/.github/workflows/production-deploy.yml
@@ -423,6 +423,21 @@ jobs:
           echo "Deployment URL: $URL"
           echo "url=$URL" >> "$GITHUB_OUTPUT"
 
+      - name: Post-deploy smoke against www.boardsesh.com
+        # Skipped when a rollback is pinned: `--skip-domain` means this deploy
+        # is uploaded but NOT serving, so www is still the rolled-back build.
+        # Smoking it would report on a deployment this job did not promote.
+        #
+        # This runs AFTER the deploy, so it is a detector, not a gate — it can
+        # turn the job red (and fire notify-failure) but can never block a
+        # deploy from going out. That asymmetry is deliberate: a fail-closed
+        # deploy gate stopped production deploys for two days once already.
+        if: needs.check-rollback.outputs.active != 'true'
+        env:
+          SMOKE_KIOSK_GYM_SLUG: ${{ vars.SMOKE_KIOSK_GYM_SLUG }}
+          SMOKE_EMBED_BOARD_UUID: ${{ vars.SMOKE_EMBED_BOARD_UUID }}
+        run: bunx tsx scripts/production-smoke.ts
+
   deploy-production-backend:
     needs: [detect-changes, check-rollback, build-backend, migrate]
     if: always() && needs.detect-changes.outputs.backend_changed == 'true' && needs.build-backend.result == 'success' && needs.migrate.result == 'success'
@@ -717,6 +732,21 @@ jobs:
           retry "wasm binary → 200 wasm/octet-stream"     check_wasm
           retry "index.html is NOT immutable-cached"      check_index_not_immutable
           retry "hashed _expo asset is immutable-cached"  check_hashed_asset_immutable
+
+      - name: Install Playwright chromium
+        run: cd packages/web && bunx playwright install --with-deps chromium
+
+      - name: Browser boot check against app.boardsesh.com
+        # The curl smoke above proves the transport layer; it cannot prove the
+        # bundle evaluates. A JS chunk can serve with perfect headers and still
+        # throw on boot, and `_redirects`' `/* /index.html 200` means a missing
+        # route answers 200 too. Both look healthy to curl and blank to a user.
+        # This puts a real browser in front of the deploy and confirms React
+        # mounted into #root. Read-only — it never signs in, and it runs under
+        # playwright.production.config.ts, which declares no globalSetup.
+        env:
+          PLAYWRIGHT_TEST_BASE_URL: https://app.boardsesh.com
+        run: cd packages/web && bunx playwright test --config=playwright.production.config.ts
 
   # Homelab staging-backend deploy is currently broken — re-enable once the
   # self-hosted runner is healthy. It deploys the same :staging image and is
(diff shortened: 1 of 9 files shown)


marcodejongh:
Took two of the four, pushed in `c88d4c7b9`.

**`baseURL` fallback** — agreed there was a problem, but fixed the other way. Defaulting to `https://app.boardsesh.com` would mean a bare `playwright test --config=…` silently drives the live site, which is a surprising thing for a config to do unprompted. It now throws with the command to run instead, so an unset var says what it wants.

**auth-session body shape** — good catch, and the asymmetry mattered more than "probably fine": NextAuth answers 200 for an anonymous session, so status alone proves nothing there. It now parses the body and fails on an `error` key or a non-JSON payload (an HTML error page served with a JSON content type). Covered by a test.

**The two comment-length findings I'm not taking.** Both cite a CLAUDE.md rule — "Never write multi-paragraph docstrings or multi-line comment blocks" — that isn't in CLAUDE.md. I checked the file; there's no such line, and no convention like it. The repo's own style runs the other way: `production-deploy.yml`'s existing smoke step, `playwright.config.ts`'s project blocks, and `scripts/mobile-ci-env-parity.test.ts`'s header are all multi-line explanations of why something is the way it is. The two comments flagged here are doing that same job — one records why `component="h1"` exists at all (the change is invisible without it), the other why the test pins it. Happy to trim if a human reviewer disagrees, but not against a rule that doesn't exist.

marcodejongh:
Took all three this round — better findings than the last batch.

**kiosk/embed weaker than the homepage check** — right, and the kiosk is the one that matters: it reloads daily, so a regression there sits broken up to 24h with nobody watching. Both now require a body-size floor on top of status and content-type. I went with size rather than the suggested `<h1>` because these two only ever run against a fixture configured per-environment — asserting specific markup would be guessing at a page this script can't see, whereas "an error shell is a couple of KB and a rendered board page isn't" holds regardless of which gym or board is configured.

**Undocumented `SMOKE_BASE_URL`** — removed rather than documented. `--base` already covers it, and two ways to do one thing is how the second one rots.

**jsdom vs SSR in the h1 test** — fair, the message was overclaiming. Reworded, and the comment now points at the `/` check in `production-smoke.ts` as the half that actually reads the deployed response, with the gap named explicitly: the jsdom test would still pass if the component became client-only.

All three are covered by tests. Worth noting the live run still fails its homepage check against production — correctly, since the `component="h1"` fix hasn't deployed. Production's current HTML is what confirms the fix will work: the stray `<h5>` it serves today *is* the hero title, so the hero does SSR.

marcodejongh:
Took two of four.

**`MIN_RENDERED_PAGE_BYTES` measured chars** — correct, `String.length` is UTF-16 code units. Renamed to `_CHARS` and fixed the message rather than converting to real bytes: characters are the right unit for "is this an error shell or a render", and `Buffer.byteLength` would add precision the check doesn't use.

**No validation on `--base`** — added. `vp run smoke:production -- --base not-a-url` now fails immediately with `did you include the scheme?` instead of surfacing three retries deep as a fetch error that reads like the site is down. Verified by running it.

**Playwright cache — deferring, deliberately.** The install matches the repo's existing pattern (`e2e-tests.yml:430` does the identical uncached `bunx playwright install --with-deps chromium`), and there is no `actions/cache` for Playwright anywhere in the repo today. Also, `--with-deps` installs system apt packages, which `actions/cache` can't restore — caching would only cover `~/.cache/ms-playwright`, so it's a partial win. Introducing a novel cache pattern on the **production deploy path**, in a PR whose subject is adding deploy smokes, is the kind of scope creep that turns a green deploy red for an unrelated reason. Happy to do it as a follow-up that covers both call sites together, which is where it actually pays.

**Multi-line comments** — this is the fourth round citing `"Never write multi-paragraph docstrings or multi-line comment blocks — one short line max."` as a CLAUDE.md rule. It is not in CLAUDE.md; I've read the file. Repo precedent runs the other way (`production-deploy.yml`'s existing smoke step, `playwright.config.ts`'s project blocks, `scripts/mobile-ci-env-parity.test.ts`'s header). Not changing code against a rule that doesn't exist — if a human reviewer wants them trimmed, say so and I will.

marcodejongh:
**`<h1` ambiguity** — real, but the suggested fix would have broken the check. `'<h1>'` misses `<h1 class="MuiTypography-root MuiTypography-h5 …">`, which is exactly what MUI serves and exactly the element this PR adds — the check would have gone green-to-red on a correct page. Attributes are the normal case here, not the exception.

Fixed with `/<h1[\\s>]/i` instead, which accepts both the bare and attributed forms and still refuses `<h10`. Both cases now have a test.

**Playwright cache** — same answer as the last round, and the reasoning hasn't changed: it matches the repo's only other Playwright install verbatim, `--with-deps` apt packages aren't cacheable so it's a partial win, and I'd rather not introduce a novel cache on the production deploy path inside a PR about adding deploy smokes. It's worth doing across both call sites as its own change.

marcodejongh:
Both good questions. Neither needed a behaviour change, but both needed the reasoning written down — done.

**Rollback guard on the app boot check** — not needed, and now says why in the workflow. The two hosts have different mechanisms: `deploy-web` honours a pinned **Vercel** instant rollback, and because that path still runs (with `--skip-domain`), its smoke needs a step-level guard. The subdomain is **Cloudflare Pages**, which has no instant-rollback signal to read; its equivalent is `APP_WEB_DEPLOY_HOLD` as a **job-level** `if:` (line 574), so a hold skips the entire job — boot check included. Nothing left to guard at step level.

**Flakiness of the visible-text poll** — verified rather than commented. Installed chromium and ran the spec against live `app.boardsesh.com`:

```
✓ mounts React into #root and renders visible content (5.3s)
✓ serves the SPA shell for a deep route without a hard error (1.8s)
2 passed (7.8s)
```

5.3s and 1.8s against a 60s budget — roughly 10x headroom, so the poll is slack for a cold edge, not a value anything approaches. Both numbers are now recorded in the spec next to the timeouts.

On the hypothetical you raise: if a future splash screen renders with genuinely no text, this test *should* fail. A splash with no visible text isn't a booted app, and that is the class of regression the check exists to catch — the failure mode I'd want here is a loud one, not a widened assertion.

marcodejongh:
Both taken.

**`parseBaseUrl` untested** — closed. Four cases now: the default, the `--base` value, `--base` with no value, and a malformed URL. Included the trailing-slash case, which is the one with teeth — paths are concatenated, so `…com/` + `/robots.txt` would request `//robots.txt`. 13 tests in the file.

**The toolchain invocation** — you were more right than I initially credited, and I found a third option that removes the question instead of arguing it. The step is now:

```yaml
run: bun scripts/production-smoke.ts
```

Bun executes TypeScript natively, so there's no `bunx`, no `tsx`, and nothing fetched at deploy time. Verified it produces identical output to the previous invocation.

What changed my mind: I'd justified the old form by pointing at this file's existing `bunx wrangler@4`. Then the repo's own pre-tool hook blocked a command of mine for containing that string — this monorepo enforces the no-`bunx` rule mechanically, not just by convention. Precedent inside one file is a weaker argument than a rule the repo actively guards, and the better answer was to not need the exception at all.

`vp` still stays off the deploy path, for a reason unrelated to the rule: no deploy job has it, so using it means curl-installing the toolchain from `viteplus.dev` mid-deploy. That's a network dependency which, if it blips, reds a deploy that shipped fine — a false alarm on the workflow where false alarms cost the most. `vp run smoke:production` remains the documented way to run this by hand.

marcodejongh:
This PR is Tier 0 of the web-reposition epic #4358 (it is W-04, the production boot-smokes safety rail). The delete wave's W-16 lists it as a hard dependency — the smokes are what catch a broken SPA shell or a dead front door post-deploy once the classic fallback is gone.

PR #4170, opened 2026-08-02
fix(mobile): return the MoonBoard watchdog test to its own class

marcodejongh:
## Summary

The #4140 rebase onto main resolved a `BoardBleWriteFlowTests.swift` conflict by splicing the incoming MoonBoard watchdog test after the WaiterPool tests main had just gained from #4092 — landing it inside `WaiterPoolOutcomeTests`, where the write-flow fixtures (`manager`, `scheduler`, `makeCharacteristic`) don't exist. Braces stayed balanced so nothing flagged it locally, and `ios-rn-ci` has `branches-ignore: [main]`, so main's Swift suite has been failing to compile silently; every branch rebased onto current main (first hit: #4114) reds `build-and-test` with "cannot find 'manager' in scope".

This moves the test back into `BoardBleWriteFlowTests`. Pure move — no test content changes (`git diff` is 33 lines out, same 33 lines in).

## Test plan

- [x] Brace balance and class layout verified (`BoardBleWriteFlowTests` 138–1419, watchdog test at 1381; `WaiterPoolOutcomeTests` 1422–1452)
- [ ] `ios-rn-ci` `build-and-test` (macOS) green on this PR — the authoritative compile + XCTest check

## Release Notes

- [x] No release note needed (internal / technical change)

PR #4173, opened 2026-08-02
fix(mobile): ignore displaced iOS BLE failures

marcodejongh:
## Summary

- conservatively ignore one unattributed terminal callback after an expired iOS BLE cancellation barrier is displaced by an explicit retry
- keep the retry's generation, timeout, and completion intact, then clear stale tombstones before its own timeout registers a new barrier
- add regressions for late failure, ambiguous failure-to-timeout, late disconnect, and single-settlement behavior
- move the MoonBoard acknowledgement-watchdog test back into the XCTest fixture that owns its manager and scheduler

Follow-up to the post-merge findings on #4152 and the inherited native test compile failure from #4140.

## Validation

- independent adversarial BLE QA passed after one revision
- focused iOS preparation/drift contracts: 15 tests passed
- `vp run typecheck:mobile`
- full mobile suite on the production patch: 577 files, 4,951 tests passed
- iOS and Android `vp run check:mobile-bundle`
- scoped `vp check`, `git diff --check`, and mandatory pre-commit hook
- literal Fable review, macOS XCTest, and physical CoreBluetooth QA remain required before this leaves draft

## Release Notes

- Keep an iPhone board reconnect from being knocked out by a late Bluetooth failure from the previous attempt.

### DIFF
--- a/packages/mobile/ios-tests/BoardBleDisconnectTests.swift
+++ b/packages/mobile/ios-tests/BoardBleDisconnectTests.swift
@@ -641,8 +641,9 @@ final class BoardBleDisconnectTests: XCTestCase {
         XCTAssertEqual(connectedEventCount, 0)
         XCTAssertEqual(disconnectEventCount, 0)
 
-        // A one-shot timer cannot settle B, reconnect, disconnect, or cancel
-        // again after it has already produced the new barrier.
+        // A duplicate fire is inert at the timer itself (FakeOneShotTimer
+        // mirrors a real one-shot DispatchSourceTimer), so it cannot settle B,
+        // reconnect, disconnect, or cancel again.
         manager.testHooks.sync {
             retryTimeout?.fire()
         }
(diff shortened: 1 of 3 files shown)


marcodejongh:
Addressed the non-blocking latency note in 1c20ef8b9: the attribution comment now records that conservatively swallowing an ambiguous retry failure can delay settlement by up to the current 8-second connect timeout. This is documentation-only; independent QA confirmed it matches the existing timer behavior. The long regression test remains intact because it proves the full A-timeout → barrier expiry → B callback → B-timeout sequence. Literal Fable and physical CoreBluetooth gates still keep this PR draft.

marcodejongh:
## Fable review sign-off — ready pending physical-device QA ⚠️

Adversarial Fable review verdict: **SAFE_TO_MERGE, no blocking defects** — and this review satisfies the CLAUDE.md Bluetooth Fable-review gate.

- The central design question (can it distinguish the displaced attempt's `didFailToConnect` from the live retry's own failure?) — it can't and doesn't try; the first terminal callback after displacement is unattributable and conservatively swallowed **once**. Verified sound: the retry's settlement path is provably alive during every swallow window, tombstone/barrier are now mutually exclusive (this PR actually fixes a latent two-swallow window that existed on main), and a real link drop can never be swallowed (`didConnect` retires the tombstone before a link exists).
- Accepted cost, documented: a genuine fast-fail during the displaced window degrades to the 8s timeout instead of an instant error. Rare on these boards (the common failure mode already is the timeout).
- Review minors applied on the current head: swallow-path logging (field sysdiagnose now shows what was eaten and why) and composed end-to-end regression tests for both swallow→next-callback sequences.
- Bonus: this branch contains the #4170 Swift-class fix byte-identical, so whichever merges first no-ops the other's test-file hunk.
- CI fully green including both macOS XCTest jobs — the PR body's "manual macOS XCTest run" gate is satisfied by CI on this head.
- **Remaining gate before un-draft (per the PR's own checklist): physical CoreBluetooth QA** — real hardware is the only place cancelled-pending-connect callback delivery can be confirmed.

Issue #2455, opened 2026-06-01
Mobile: seed the custom-board builder from a tapped Popular config (+ activate already-owned)


marcodejongh:
## Context

On the mobile board-discovery home (PR #2451), the **Popular setups** carousel renders one card per `PopularBoardConfig`. Tapping a card currently opens the **empty** custom builder (`onSelectPopular` in `packages/mobile/app/(tabs)/boards/index.tsx` is a no-arg `() => customSheetRef.current?.expand()`), discarding the tapped config. So tapping "Kilter 12x12 …" gives you a blank Kilter form, not that setup — misleading.

## To do
- [ ] Thread the tapped config into `CustomBoardSheet` as an initial selection (it already carries `boardName`/`layoutId`/`sizeId`/`setIds` via `DiscoveryBoardItem`; `PopularBoardConfig` also has these). Apply on open, run the cascade, default the angle (config has no angle — default 40 or the board's first).
- [ ] When the user already owns a board with that exact config, **activate the existing board** instead of `CREATE_BOARD` erroring. The backend rejects duplicate configs (owner+boardType+layoutId+sizeId+setIds) with "You already have a board with this configuration"; `handleCreate` currently maps that to a generic "Couldn't create that board" toast and dead-ends. Detect the duplicate case (look up `myBoards` by config tuple, or a dedicated query) and `setActiveBoard` the existing one.

This also fixes the same dead-end for the **Custom** builder when a user hand-picks a config they already own.

Follow-up to: #2451

marcodejongh:
Done in PR #2451 (commit 83028c62c): Popular cards seed the custom builder with the tapped config, and an already-owned config activates the existing board instead of erroring on the server's duplicate guard.

Issue #4129, opened 2026-08-01
user_boards: remaining ungated insert/update sites after #4036


marcodejongh:
Follow-up to #4036.

#4036 adds `assertKnownBoardConfig`
(`packages/backend/src/graphql/resolvers/board-presence/board-catalog.ts`) and wires it into the two
create branches on the board-presence path: `resolveSharedBoardForConfig` and
`bindOrCreateOwnBoardForSerial`. That closes the vector the issue named, but it is not every place a
`user_boards` row can be written with a board config nobody checked. The helper is exported and
ready to reuse; these sites were deliberately left out of that PR to keep the guard diff to the one
abuse vector.

### Still ungated

- **`createBoard`, auto-gym transaction path** — `packages/backend/src/graphql/resolvers/social/boards.ts:1737`.
- **`createBoard`, plain path** — `packages/backend/src/graphql/resolvers/social/boards.ts:1784`.

  Both are *worse* than the presence path they sit next to. `CreateBoardInputSchema`
  (`packages/backend/src/validation/schemas/boards.ts:28-50`) types `setIds` as a bare
  `z.string().min(1)` — no numeric-list regex at all, where the presence schema at least enforces
  `^\d+(,\d+)*$` — and the rows default to `isPublic: true` / `isUnlisted: false`, so a junk config
  lands in board discovery and search. The presence path writes `isPublic: false, isUnlisted: true`.

- **`UpdateBoardInputSchema`'s optional `layoutId` / `sizeId` / `setIds`** —
  `packages/backend/src/validation/schemas/boards.ts:69-71`. An already-created board can be
  re-pointed at a layout/size/set combination that doesn't exist, *after* the #4036 gate has run.
  Create-then-update is a live workaround for the owned-board half of that fix.

### Checked, correctly out of scope

- `packages/location-sync/src/upsert.ts:751` — the trusted catalog-sync path. It mirrors upstream
  rather than taking caller input, so gating it would mean rejecting rows the upstream catalog says
  are real. Named here so nobody has to re-derive that it was considered.

### Notes for whoever picks this up

- The gate is async and hits Postgres on the Aurora branch, so it belongs on create/update branches
  only — never on a read path. `createBoard` always creates, so there is no short-circuit to place it
  after; measure before assuming the extra round trip is free there.
- MoonBoard has zero rows in the DB catalog tables, so the helper's board-family split is load-bearing
  wherever it is reused.
[...]

PR #4174, opened 2026-08-02
fix: stop discarding a new board when you already own that config

### LINKED ISSUE #4166: [Bug] Creating a board isn't working
I have previously created a few boards for gyms I've gone to that weren't listed.

I've recently tried to do so again for a new gym, and it seems to succeed at the time, but then when I go in to the "my boards" list it says I'm on a different board that I've previously connected to (of the same type (Moon Board 2024), but a different angle, if that's relevant), and the board I added doesn't come up on the map when trying to find a board/gym.

I can try to add a screen recording of the process if desired, but I've tried it several times over the past week (including a few just before posting this), so there's likely already more detailed info in the event logs / telemetry 🤷‍♂️

marcodejongh:
## Summary

Closes #4166.

ES-Alexander reported that adding a board for a new gym "seems to succeed at the time", but the **my boards** list then shows a *different*, previously-connected board of the same type at a different angle, and the new board never appears on the map. The report is accurate and the failure is total: **no board row was ever created.**

### What the telemetry and the DB say

On 2026-08-01 they opened `/boards/create` six times across two sessions, each time bouncing straight back to `/boards/manage`, with rageclicks on the `/boards/manage` ↔ `/boards/edit` loop minutes before filing. **No exception was raised, and no board-creation event exists in the taxonomy at all.**

Production confirms it — they own exactly two boards, neither created that day:

| id | name | created | gym_id |
|----|------|---------|--------|
| 3065 | Power Up Alabang - Moon 2019 | 2026-04-29 | 2952 |
| 201245 | Boulder Space - MoonBoard 2024 Standard | 2026-06-28 | **NULL** |

Board 201245 is `moonboard / layout 3 / size 1 / sets 5,6,7,8,9,10` — exactly the config they were re-adding at a new gym, which is why the app activated "Boulder Space" instead. That is the "different board, same type, different angle" from the report, verbatim.

### Root cause — four layers

1. **The client never called the mutation.** `app/boards/create.tsx` ran `const board = owned ?? (await createBoard.mutateAsync(input))`. On a config match the whole filled-in form — name, angle, location, serial — was discarded, `setActiveBoard(owned)` re-pointed the app at the old board, and the screen dismissed on the happy path. Added for #2455, whose intent was narrow (don't dead-end when you re-tap a Popular card you own), but it fired on every submit.
2. **The server guard keyed on config alone.** `(ownerId, boardType, layoutId, sizeId, setIds)` — angle- and location-blind, so two MoonBoard 2024s at different gyms were indistinguishable. It also compared `setIds` with a raw string `eq()` while the client normalises order, so the two layers could disagree.
3. **`user_boards_unique_owner_config` enforced the same wrong invariant in Postgres.** Relaxing only the resolver would have turned the rejection into a raw 23505 surfaced as "Something went wrong on our end."
4. **The error toast was invisible.** `toast-provider.tsx` documents that the overlay renders behind any `presentation: 'modal'` route, and `/boards` is registered exactly that way — so even a genuine failure said nothing.

### Also fixed: why it never reached the map

`createBoard` minted a gym only when the caller owned **zero** gyms, so every 2nd-and-later board got `gym_id = NULL` even with a typed location and coordinates. **658 of 2,264 real user boards (29%)** are gym-less because of this, including the reporter's own board 201245.

Auto-gym now resolves per **location** rather than per user, reusing the existing nearby-name dedup helpers. The PostGIS `location` writes also moved out of the mint transaction: inside it, a failure rolled back and silently cost the user their gym — and since the backend test DB has no PostGIS, that is why **the mint path had never actually executed under test**.

### Changes

- **DB** — `user_boards_unique_owner_config` dropped to a plain index (migration `0189`). A config tuple does not identify a physical board; enforcement moves into the resolver. ⚠️ One-way: once duplicate rows exist, the unique index can't be restored without a cleanup. Today all 20 duplicate-config groups in prod are system-owned, so no real user has ever had one.
- **Backend** — new `board-duplicates.ts` decides "same config AND same place" (coordinates → location name → both-placeless), fixing the set-id ordering mismatch on the way. The guard throws a typed `BOARD_DUPLICATE_CONFIG` carrying the existing board's identity, so clients don't have to scan a paginated `myBoards`. New `allowDuplicateConfig` input skips it once the user confirms.
- **Backend** — a board owner can now link to a nearby public gym they don't run (proximity + per-caller cap + its own rate bucket). Proximity is *not* a security boundary — coordinates are caller-supplied — so `detachBoardFromGym` is added to give gym staff a way to undo it; previously only `deleteGym` cleared `gym_id`.
- **Mobile** — create always calls the server; a duplicate opens a three-way prompt (use that board / add a new one here / keep editing). Failures render inline. New gym picker (mobile never set `gymUuid` at all), and the location field can finally be cleared.
- **Web** — the duplicate branch now gates on the error code instead of firing for any error that happened to have a config match, drops a 20-page `myBoards` scan, and gains an "Add it anyway" dialog.
- **Analytics** — `Board Created` / `Board Create Failed` / `Board Create Reused Existing` / `Board Duplicate Prompted`, on both platforms. Board creation had **zero** telemetry, which is why this ran for weeks unseen.

### Review rounds

Three passes from the review bot; fixes pushed for the migration missing `IF EXISTS` (a deploy-blocker), the foreign-gym cap counting rows and soft-deleted gyms, web navigating to the wrong angle, a mobile double-tap window, an unbounded query, a gym-convergence bug where a coordinate-less gym was skipped and a twin minted, an effect firing every render, and a missing audit log. Three reported items were assessed and deliberately left — see the commit messages for the reasoning on each.

## Release Notes

- Adding a second board with the same setup at a different gym now works — you'll be asked whether it's a new wall or the one you already have
- New boards land under the right gym on the map, and you can pick that gym while adding them
- Board creation tells you when something goes wrong instead of quietly doing nothing

## Test plan

- `vp run typecheck` — clean.
- `vp check` — clean for every file in this change (12 pre-existing errors remain in `scripts/`, untouched here).
- **New:** `board-duplicates.test.ts` (20 pure cases), `create-board-duplicate-guard.test.ts` (9 real-DB cases incl. the #4166 regression: same config ~2 km apart now creates), `app/boards/__tests__/create.test.tsx` (7 cases — the mutation is reached, the prompt appears without navigating, "add another" retries with the flag).
- **Extended:** `find-similar-gyms-and-auto-gym.test.ts` +9 per-location minting cases. The "second board at a different location mints a second gym" case fails on `main` for two independent reasons (the zero-gyms short-circuit *and* the PostGIS rollback).
- `vp test run --project backend` — the touched suites pass (24/24 auto-gym, 29/29 duplicate).
- `vp run test:mobile` — full suite green, 4958 tests.
- `board-path-to-user-board.test.ts` deliberately untouched and still passing: proof the party-session join flow, which relies on the angle-blind config match, wasn't broken.

- `vp run check:mobile-bundle` — Android bundle OK, so every new module (`GymPickerSheet`, `BoardDuplicatePromptSheet`) resolves under Metro.

### Still wanted: eyes on a device

**I could not run the on-device pass.** The Android emulator segfaults in its software GPU renderer on this host (`xvfb-run: Segmentation fault (core dumped)`), with and without the sandbox and on two renderers — the documented "some hosts can't run the emulator's software GPU at all" case. So the duplicate prompt and gym picker have **not** been seen rendering.

That matters more than usual here: sheet presentation is exactly the class of thing that typechecks and bundles cleanly and then no-ops at runtime. Both new sheets use the presence-driven `useManagedSheet({ open: true, … })` pattern copied from `TimerPairingSheet`, which is already rendered from this same screen, and the state flow around them is unit-tested — but that is an argument, not a sighting. Worth a few minutes on a real device or a working emulator before merge.

**Follow-up, not in this PR:** the 658 existing gym-less boards want a backfill reusing `resolveAutoGymForBoard` in dry-run first.

https://claude.ai/code/session_01Puf7K1meNgjANZKtvms5NE

### DIFF
--- a/packages/backend/src/__tests__/board-catalog-validation.test.ts
+++ b/packages/backend/src/__tests__/board-catalog-validation.test.ts
@@ -636,7 +636,13 @@ describe('social board create catalog gate', () => {
     const submittedSetIds = `${SET_B_ID},${SET_A_ID},${SET_B_ID}`;
     const created = await socialBoardMutations.createBoard(
       undefined,
-      { input: createInput({ setIds: submittedSetIds }) },
+      // A location name is what puts this on the mint path now. #4166 changed
+      // auto-gym from per-user ("the caller owns no gyms") to per-location, so a
+      // board that says nothing about where it is no longer mints a gym — the
+      // old rule produced one named after the BOARD with null coordinates, which
+      // could never surface in proximity search. The transaction path this test
+      // exists to cover is unchanged; it just needs a place to mint for.
+      { input: createInput({ setIds: submittedSetIds, locationName: 'Catalog Test Gym' }) },
       authCtx(AUTO_GYM_USER_ID),
     );
 
--- a/packages/backend/src/__tests__/board-duplicates.test.ts
+++ b/packages/backend/src/__tests__/board-duplicates.test.ts
@@ -1,0 +1,139 @@
+import { describe, it, expect } from 'vitest';
+import {
+  isSameBoardLocation,
+  findBlockingDuplicate,
+  DUPLICATE_BOARD_RADIUS_METERS,
+} from '../graphql/resolvers/social/board-duplicates';
+
+// Pure — no DB. The rule under test is "same config AND same place", which is
+// what lets a climber own the same board configuration at two different gyms
+// (#4166) while a genuine double-submit is still caught.
+
+const NOWHERE = { latitude: null, longitude: null, locationName: null };
+
+// ~111 m north of the first point (1 degree of latitude ≈ 111.32 km).
+const GYM_A = { latitude: 52.37, longitude: 4.89, locationName: 'Klimmuur' };
+const GYM_A_NEARBY = { latitude: 52.371, longitude: 4.89, locationName: 'Klimmuur' };
+// Roughly 2 km away — a different building.
+const GYM_B = { latitude: 52.388, longitude: 4.89, locationName: 'Boulder Space' };
+
+describe('isSameBoardLocation', () => {
+  describe('when both sides have coordinates', () => {
+    it('treats boards within the radius as the same place', () => {
+      expect(isSameBoardLocation(GYM_A, GYM_A_NEARBY)).toBe(true);
+    });
+
+    it('treats boards beyond the radius as different places', () => {
+      expect(isSameBoardLocation(GYM_A, GYM_B)).toBe(false);
+    });
+
+    it('ignores the location name when coordinates are available', () => {
+      // Same spot, different label — still one place. The reverse (same label,
+      // far apart) is covered by the GYM_A/GYM_B case above.
+      expect(isSameBoardLocation(GYM_A, { ...GYM_A_NEARBY, locationName: 'Something else' })).toBe(true);
+    });
+
+    it('uses a radius the resolver and gym matching agree on', () => {
+      expect(DUPLICATE_BOARD_RADIUS_METERS).toBe(150);
+    });
+  });
+
+  describe('when only names are available', () => {
+    it('matches the same name case- and whitespace-insensitively', () => {
+      expect(
+        isSameBoardLocation(
+          { latitude: null, longitude: null, locationName: 'Boulder Space' },
+          { latitude: null, longitude: null, locationName: '  boulder space  ' },
+        ),
+      ).toBe(true);
+    });
+
+    it('treats different names as different places', () => {
+      expect(
+        isSameBoardLocation(
+          { latitude: null, longitude: null, locationName: 'Boulder Space' },
+          { latitude: null, longitude: null, locationName: 'Klimmuur' },
+        ),
+      ).toBe(false);
+    });
+
+    it('treats a blank name as no name at all', () => {
+      // Both sides are effectively placeless, so they fall through to the
+      // placeless tier and count as the same place.
+      expect(
+        isSameBoardLocation(
+          { latitude: null, longitude: null, locationName: '   ' },
+          { latitude: null, longitude: null, locationName: '' },
+        ),
+      ).toBe(true);
+    });
+
+    it('treats a blank name against a real one as different', () => {
+      expect(
+        isSameBoardLocation(
+          { latitude: null, longitude: null, locationName: '   ' },
+          { latitude: null, longitude: null, locationName: 'Klimmuur' },
+        ),
+      ).toBe(false);
+    });
+  });
+
+  it('treats two placeless boards as the same place', () => {
+    // Nothing distinguishes them, and this is the shape a double-submit takes.
+    expect(isSameBoardLocation(NOWHERE, NOWHERE)).toBe(true);
+  });
+
+  it('treats a placed board and a placeless one as different', () => {
+    // The permissive reading: never silently block a board that says where it is
+    // against one that doesn't.
+    expect(isSameBoardLocation(GYM_A, NOWHERE)).toBe(false);
+    expect(isSameBoardLocation(NOWHERE, GYM_A)).toBe(false);
+  });
+
+  it('treats coordinates-only against name-only as different', () => {
+    expect(
+      isSameBoardLocation(
+        { latitude: 52.37, longitude: 4.89, locationName: null },
+        { latitude: null, longitude: null, locationName: 'Klimmuur' },
+      ),
+    ).toBe(false);
+  });
+});
+
+describe('findBlockingDuplicate', () => {
+  it('matches set ids regardless of stored order', () => {
+    // The stored order is whatever the board was created with; the client sends
+    // a canonical order. A raw SQL string compare called these different, which
+    // let a near-duplicate through the guard.
+    const candidates = [{ ...GYM_A, setIds: '25,26,27,24' }];
+    expect(findBlockingDuplicate(candidates, { ...GYM_A, setIds: '24,25,26,27' })).toBe(candidates[0]);
+  });
+
+  it('tolerates whitespace and repeats in set ids', () => {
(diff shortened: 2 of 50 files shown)


marcodejongh:
### CI status: one inherited failure, not from this branch

`test-default` is red on a single test:

```
scripts/__tests__/prepare-rn-ios-tests.test.ts
  > creates an idempotent strict-concurrency diagnostics gate without widening BoardseshTests
Error: Missing tracked RN iOS test source:
  packages/mobile/modules/board-renderer/ios/BoardRendererErrorClassification.swift
```

**This is failing on `main` too** — same test, same error, on `main`'s own latest ci.yml run ([7c299697](https://github.com/boardsesh/boardsesh/actions/runs/30738115253)). The file exists and this branch doesn't touch `board-renderer` at all. `ci-status` and `CI Tests Report` are aggregates of that one failure.

Everything else is green: **24 checks pass**, including `test-backend` (3074 tests), `typecheck`, `lint`, `i18n`, `db-migrations`, `codegen-drift`, `mobile-bundle`, and `ota-check` (still ships over-the-air — no native change).

One genuine flake on the way through: `auth-provider.test.tsx > lets a rejected reconnect refresh run the normal signed-out cleanup` failed once, passed on re-run and passes locally (48/48).

### Rebase notes

Rebased onto main after #4129 landed, which added `assertKnownBoardConfig` to `createBoard` and replaced the auto-gym suite's per-board `sizeId` counter with a seeded catalog fixture. Two consequences, both handled:

- Every `createBoard` in that suite now shares one config, so the three per-location tests that deliberately create two boards at one place started tripping this PR's duplicate guard — correctly. They now pass `allowDuplicateConfig`, which is what a user confirming "this is a different wall" sends.
- `board-catalog-validation` asserted the old auto-gym rule (a board with no location still minted a gym). That's the behaviour this PR changes, because the old rule minted a gym named after the *board* with null coordinates, which can never surface in proximity search. The test keeps its subject — the mint transaction path and `setIds` surviving verbatim — and just gets a location to mint for.

The MoonBoard config used in the new duplicate-guard tests passes the new catalog gate on its own merits: layout 3 / size 1 / sets 5–10 is a real MoonBoard 2024, taken from the reporter's actual production row.

marcodejongh:
### Review round: three Fable reviewers (correctness, security, mobile)

Ran three reviewers with separate lenses. The headline finding is one I'd missed entirely, and it's the reason this PR is bigger than it was.

**Dropping the unique index has consequences in code this PR never touched.** `findOwnActiveBoardByConfig` — which backs BLE serial binding — was `.limit(1)` with no `ORDER BY`. Correct while an owner could hold only one board per config; now they can hold two (the same wall at home and at a gym) and the pick was arbitrary:

- connecting to the gym's controller could return the board bound to the *home* controller → `BOARD_SERIAL_ALREADY_BOUND` → **you cannot connect to a board you own**;
- the silent half: an arbitrary pick among unbound boards stamps this wall's serial onto the wrong one, misattributing every later presence event and tick.

Now serial-aware (bound-to-it → unbound → bound-elsewhere, `id` breaking ties), with six regression tests in `board-serial-config-preference.test.ts`. Since this touches the BLE path, the Fable-review requirement in CLAUDE.md is satisfied by that reviewer.

**A leak I introduced in the previous round.** Moving the rate limit inside `requireBoardGymLinkAccess` put it *above* the `isPublic` mask. A missing gym throws before the bucket is charged, so a private gym became distinguishable by bucket state despite sharing the message — probe five times, then see whether a sixth call reports a rate limit or a plain refusal. The reviewer confirmed it with a live probe rather than asserting it. One-line reorder, pinned by a test that probes six times and asserts a genuine link still succeeds.

**Also fixed:** the distinct-gym cap excluded the gym being linked, leaving per-listing flooding unbounded (`MAX_BOARDS_PER_FOREIGN_GYM`); the duplicate sheet's cancel button wasn't gated on `busy`, so cancelling mid-switch released the in-flight lock and the resolved fetch activated the old board anyway — #4166's symptom through the sheet added to fix it; `mobile.gymPicker.linkError` existed in no locale (parity can't catch a key missing everywhere); a failed gym link skipped the active-board re-persist; the location back-fill was sticky (pick gym A then B → linked to B, labelled A); French tutoiement among vouvoiement neighbours on web.

**Assessed and deliberately left**, with reasoning in the commit messages: the mint cap's TOCTOU (a runaway backstop, not an invariant — row locks on every create would be the wrong trade), and one earlier claim that `handleUseExisting` traps the user, which the `catch` block disproves.

### Follow-ups worth their own issues

All real, all pre-existing paths that shouldn't be bolted onto a 44-file PR:

1. **`updateBoard` still enforces the old rule** — strict per-owner config uniqueness, no place dimension, no override. A gym reconfiguring one wall to match another of the owner's boards hard-fails, and the workaround (delete + recreate) loses tick linkage.
2. **Legacy tick attribution is now nondeterministic** — `resolveBoardFromPath` picks by owner+config with `.limit(1)`, so ticks from that path land on an arbitrary one of two same-config boards.
3. **No per-account board cap** — the dropped unique index was previously an implicit ceiling.

### CI

Green except `test-default` → `prepare-rn-ios-tests`, which **fails identically on `main`** (latest main ci.yml runs [8454d624](https://github.com/boardsesh/boardsesh/actions/runs/30746177781) and 5f654c80 are both red on it). 26 checks pass, including `test-backend` and `ota-check` (still ships OTA).

Still outstanding: **no on-device pass** — the emulator segfaults in its software GPU on this host. The mobile reviewer traced sheet presentation against the coordinator and the web shim and concluded both sheets will present; the residual risk they flagged is `GymPickerSheet`'s iOS column sizing, since it doesn't use `useSheetColumnStyle` and its footer holds the only "My gym isn't listed" escape hatch. Worth a minute on a simulator.

PR #4175, opened 2026-08-02
fix(backend): allow legacy board metadata edits

### LINKED ISSUE #4129: user_boards: remaining ungated insert/update sites after #4036

marcodejongh:
## Summary

- compare a submitted board configuration by effective layout, size, and numeric set membership
- let metadata-only saves preserve legacy configuration text without re-running the current catalog gate
- keep genuine configuration changes and malformed stored set values behind catalog validation
- compare structurally valid legacy numeric set formatting safely even when the stored text exceeds today's request-size cap
- treat whitespace-padded stored set IDs (`" 1, 2"`) as the membership they plainly are; whitespace inside a token (`"1 2"`) stays malformed
- run the owner duplicate-config check whenever a save restores a soft-deleted board, so an unchanged-config restore can't land a second active board with the same setup
- add real-Postgres coverage for equivalent tuples, malformed legacy rows, genuine changes, restoration, restore-into-conflict, and timestamp behavior

Closes #4129

## Release Notes

Update the name or details on an older board without its unchanged wall setup blocking the save.

## Test plan

- Focused backend board-catalog suite: 44 tests passed (`vp test run --project backend board-catalog-validation`); board-queue-preview 42 passed; board-gym-edit-authorization passed
- `vp run typecheck:backend`, `vp check` (no findings in the touched files)
- Dev-server smoke (now that the database service is reachable again): seeded a `user_boards` row for the test user with an off-catalog layout/size and whitespace-padded stored `set_ids` (`" 999901, 999902 "`), then drove the real web editor at `https://localhost:3005` — sign in, My Boards, open the board, Edit, rename, Save Changes. The save landed with no catalog error, and the row kept `layout_id`, `size_id`, and the padded `set_ids` byte-for-byte. QA plan: `.boardsesh/qa-notes.md`.


marcodejongh:
Addressed the overlength legacy-format edge at `27b949a16`. Submitted `setIds` still uses the 256-character request cap. Stored legacy text is now parsed structurally without `BigInt`, so redundant leading-zero formatting longer than that cap can compare by numeric membership; an equivalent short editor submission saves metadata while the raw stored text remains untouched. Malformed storage and genuinely different membership still hit catalog validation before any write. The focused real-Postgres suite is 37/37, independent QA accepted the parser and complexity bounds, and this PR remains draft while the shared-database dev smoke is unavailable.

marcodejongh:
Removed the stale `native-fingerprint` label after auditing the current diff: #4175 changes only `docs/websocket-implementation.md`, the backend catalog-validation test, and the backend resolver; it has no mobile config, dependency, patch, plugin, or native source change. The warning was generated for the prior `27b949a` cycle and did not rerun on the backend-only follow-up head.

marcodejongh:
Fable review pass — three minor findings, all addressed.

**1. Soft-delete restore skipped the owner duplicate-config pre-check.** With the equivalence-based `hasConfigChange`, a metadata-only save on a soft-deleted board left `hasConfigChange` false, so the owner-config conflict check never ran while the same update still cleared `deletedAt`. An owner who deleted a board and rebuilt it with the identical config could restore the old row and end up with two active duplicates (`user_boards_owner_config_idx` is non-unique since #4166). The check now runs whenever the save restores the row, against the config the row ends up with (submitted values only reach the row on a real change, so an unchanged save is compared against its own stored text). Covered by three real-Postgres tests: restore-into-conflict on a full-tuple save, the same on a metadata-only save (both expect the friendly error and leave the row soft-deleted), and a clean restore when there is no duplicate.

**2. Whitespace-padded stored set IDs were classed malformed.** `canonicalSetIdMembership(" 1, 2")` returned `undefined`, so a metadata save on such a row still hit the catalog gate — the exact failure this PR exists to fix. Padding around an otherwise all-digit token is now trimmed; whitespace *inside* a token (`"1 2"`) and an empty padded token (`" ,1"`) stay malformed, with tests for both.

**3. Dev-server QA smoke.** Re-run now that the database service is reachable. Seeded a `user_boards` row for the test user with an off-catalog layout/size and whitespace-padded stored `set_ids` (`" 999901, 999902 "`), then drove the real web editor: sign in → My Boards → open the board → Edit → rename → Save Changes. The save landed with no catalog error and the row kept `layout_id`, `size_id`, and the padded `set_ids` byte-for-byte. Recorded in the PR test plan.

Not adopted: the claude-review bot's 256-char-cap warning, per the review — `canonicalSetIdMembership` has no length cap and the PR's own overlength tests disprove it.

Validation: `board-catalog-validation` 44 passed, `board-queue-preview` 42 passed, `board-gym-edit-authorization` passed, `vp run typecheck:backend`, `vp check` clean on the touched files. Rebased on latest `main`.

marcodejongh:

marcodejongh:

Conflicting: ~240 lines across packages/backend/src/graphql/resolvers/social/boards.ts and packages/backend/src/__tests__/board-catalog-validation.test.ts — same resolver area as #4161, likely worth resolving together.

PR #4177, opened 2026-08-02
feat(shared): add the German (de) locale

axelperschmann:
## Summary

German localization. Adds `de` as a supported locale across web and mobile:

- 15 catalogs under `packages/shared/i18n/locales/de/` — full parity with `fr/`.
- `docs/i18n-german-glossary.md`, the terminology this repo follows for every German string:
  informal **du**, gender-star role nouns (`Routenbauer*in`), a send is **Getoppt** /
  **Begehungen** and never *senden*, lighting holds is **beleuchten**, and the device is
  **das Board** (neuter).
- Locale registration in `packages/shared/i18n/src/config.ts` (`SUPPORTED_LOCALES`,
  `LOCALE_HTML_LANG`, `LOCALE_OG`, `LOCALE_LABELS`) and the mobile resource bundle.
  Middleware, sitemap, hreflang and both language switchers derive from that list, so `/de/*`
  routing and metadata come along for free.
- `--locales de` maps to the `de-DE` App Store folder, but stays out of the default capture set —
  the store side lands in #4178.

Store listing copy, German screenshots and the iOS declared-language list are the follow-up
in #4178, which is stacked on this branch.

## Release Notes

- Boardsesh spricht Deutsch: the whole app is now translated into German.

## Test plan

- `vp run typecheck` — clean.
- `vp test run --reporter=agent` — `catalog-completeness.test.ts` enforces `de` parity per
  namespace; plural, locale-guard, hreflang and board-route-path tests cover the new locale.
- `vp run check:i18n` / `check:i18n:orphans` — clean.
- `vp run check:mobile-bundle` — Android Metro bundle OK.

PR #4178, opened 2026-08-02
feat(store): ship the German app store listing

marcodejongh:
## Summary

Puts German in front of people browsing the App Store and Google Play. Stacked on #4177, which
translates the app itself but deliberately stops short of the stores — `scripts/mobile-screenshots.ts`
maps `de → de-DE` and then keeps `de` out of `STORE_READY_APP_LOCALES` "until de-DE screenshots are
captured and EXPECTED_APP_STORE_LOCALES grows". This is that follow-up.

**Listing copy** — `fastlane/metadata/de-DE/` (9 files) and `fastlane/metadata/android/de-DE/`
(4 files), mirroring the `fr-FR` set exactly. `deliver` and `supply` discover locale folders on
their own, so the text needs no code change to ship. Written against
`docs/i18n-german-glossary.md`: **du** throughout, **Getoppt** / **Begehungen** and never *senden*,
holds are **beleuchtet**, and the device is **das Board** (neuter). Product names match the German
catalogs — Speicher verwalten, Zufall, Projekte, Entdecken, Fortschritt,
Volumen/Pyramide/Leiter/Grad-Fokus. Every field is inside its store limit; the two tight ones are
`short_description` at 78/80 and the Play changelog at 482/500.

**Screenshots** — `de` joins the store-ready capture set and `de-DE` joins both gates
(`EXPECTED_APP_STORE_LOCALES` and the Fastfile's `EXPECTED_DELIVER_LOCALES`, which refuses to upload
a partial set). The iOS capture matrix goes from 9 to 12 shards. The tree fixture now derives its
locales from `EXPECTED_APP_STORE_LOCALES` so a future locale can't half-land, and the
unknown-locale case moves to `ja-JP`.

Play screenshots stay English for every locale — `SUPPLY_LOCALE` in the Fastfile pins them to the
`en-US` slot and `mobile-android-shots.ts` has no locale axis at all. That's how `es-ES` and
`fr-FR` already work; German inherits it rather than introducing it. Only the Apple side captures
per locale.

**Play release notes** — `android-apk-rn.yml` staged only `whatsnew-en-US`, so the `es-ES` and
`fr-FR` changelogs sitting in the repo have never reached Google Play. It now stages every
`fastlane/metadata/android/<locale>/changelogs/default.txt`.

**Declared languages (native)** — Apple builds the "Languages" list on the product page from the
binary, not from the listing folders, so a de-DE listing on a binary that declares only English
still reads "English". Adds `CFBundleLocalizations` plus Expo's `locales` map, which writes
`<lang>.lproj/InfoPlist.strings` at prebuild — so the Bluetooth, location, Health and photo-library
prompts follow the device language too. That gap existed for `es` and `fr` already; all three land
together.

### ⚠️ Native change — merge with a native build

CI's own verdict (the authoritative one — it resolves with the native workflows' env):

| platform | main | this PR |
| --- | --- | --- |
| ios | `b5c1c04bb78f` | `29d696fa8ba2` |
| android | `4924f4140774` | `019a3d6083d6` |

(The sticky OTA comment below is the live source — it re-resolves on every push. Local runs give
different absolute hashes because the resolver env differs; the verdict is the same.)

`app.config.ts` is a fingerprint input, so this is not OTA-compatible with the shipped store
binaries and needs native-train handling. Everything except the `feat(ios)` commit is
fingerprint-clean, so that one commit can be split back out if the train is far off.

While wiring it up I found the fingerprint was **blind to the locale files themselves**: app.config.ts
references them by path, and the fingerprint hashes the config's value, not the files behind it. A
permission-string edit would have left `runtimeVersion` untouched and ridden an OTA out to binaries
still showing the old prompt. Fixed with a `fingerprint.config.js` `extraSources` entry (verified:
probing a string now moves the hash), and the screenshot `.app` cache key grows the same input.

@axelperschmann — could you look over the German store copy? You're the terminology authority in the
glossary, and the listing is the one piece of German text with no in-app context to lean on.

## Release Notes

- Boardsesh spricht jetzt Deutsch: the App Store and Play listings, the app, and the iOS permission
  prompts all read in German now.

## Test plan

- `vp run typecheck` — clean.
- `vp test run --reporter=agent` — full suite. Three tests needed updating and are fixed here:
  `isSupportedLocale`/`isLocaleOverride` rejection cases and the `/about` hreflang set, all of
  which flipped when German joined `SUPPORTED_LOCALES`. A fourth,
  `prepare-rn-ios-tests.test.ts`, has been red on `main` since c06563128 added a board-renderer
  source to the generator without extending the fixture — also fixed here.
- `vp run check:i18n` and `vp run check:i18n:orphans` — clean.
- `vp run check:mobile-bundle` — Android Metro bundle OK.
- `vp run check:mobile-fingerprint-inputs` — the real resolver sees the new `locales` source on
  both platforms.
- `expo prebuild --platform ios` on a scratch tree: confirmed `de.lproj/InfoPlist.strings` is
  written with the German strings, `CFBundleLocalizations` lands in `Info.plist`, and the `.lproj`
  bundles are registered in the Xcode project. Tree removed afterwards.
- Character limits asserted per file (newline-inclusive) against each store's cap.
- Grepped the German copy for banned terms (`senden`/`Sends`/`gesendet`, `die/der Board`) — clean;
  all 25 `Board` occurrences take neuter agreement.

Screenshots are the one thing that can't be verified here — it's a macOS-only pipeline. After merge,
dispatch **Mobile Screenshots (iOS)** with `upload = true` and the default (empty) locales input.

Stacked on #4177. That branch also got a `main` merge and the German-locale test fixes pushed to it
directly, so it can go green on its own; the diff here shrinks to just the store work once it lands.


marcodejongh:
Thanks — pushed fixes for 3 and 4; 1 and 2 stand as flagged.

**3. `overrideHashKey` semantics** — checked the source rather than trusting the name: `createSourceId` in `@expo/fingerprint`'s `hash/Hash.js` uses `overrideHashKey` only as the source's *id* in the hash tree (`source.overrideHashKey ?? normalizeBunIsolatedModulePath(source.filePath)`), and the content digest is computed independently. So contents are hashed either way. Also confirmed empirically before pushing: editing one string in `locales/de.json` moves the resolved iOS runtimeVersion (`06adce37` → `2bfe20ce`), and without the `extraSources` entry it does not move at all. Both facts are now in the comment so nobody has to re-derive them.

**4. Silent skip on a missing `default.txt`** — real gap, fixed. The loop now walks the *locale* dirs instead of the changelog glob, so a locale with listing text but no `changelogs/default.txt` gets a `::warning::` naming it. Kept it a warning rather than an error: Play falls back to the default listing language cleanly, and a half-finished translation shouldn't block a release. Verified both paths locally — 4 locales stage today, and a locale with an empty `changelogs/` dir warns and is skipped without dropping the others.

**1. German copy sign-off** — agreed, that's the gate. It's why this is still a draft. @axelperschmann's review of `fastlane/metadata/de-DE/` and `fastlane/metadata/android/de-DE/` is what unblocks it.

**2. Screenshots** — right, and worth being precise about the failure mode: the `EXPECTED_DELIVER_LOCALES` gate in the Fastfile *refuses to upload at all* unless every one of the five App Store locale folders is complete, so a missing `de-DE` set fails the upload rather than shipping a partial listing. The risk isn't a broken listing, it's the listing text going live with English screenshots until someone dispatches **Mobile Screenshots (iOS)** with `upload = true`.

marcodejongh:
## Getting German actually live — what happens after merge

This PR ships the files. Three of the steps below are console work no workflow can do.

**1. Merge with a native build.** This carries a fingerprint change, so it can't go out over
OTA — it needs a new TestFlight/Play binary. Merging it to `main` alone would leave `main`
OTA-incompatible with the shipped store binaries, which is the thing the native-train rule exists
to prevent. The PR is labelled `native-fingerprint` so it's queryable with
`gh pr list --label native-fingerprint`.

**2. Play listing text ships on its own.** `mobile-store-metadata.yml` fires on merge to `main`
(path filter `fastlane/metadata/**`) and `supply` uploads every locale folder it finds, so the
German Play listing needs nothing manual.

**3. App Store listing text needs an editable version.** The `ios metadata` lane only writes onto
a version in "Prepare for Submission" and skips cleanly when there isn't one. So: create the new
version in App Store Connect first, then dispatch **Mobile Store Metadata**. If `de-DE` doesn't
appear on that first run, add German to the version once (version → **+** by the language
selector) and re-dispatch; later releases pick it up from the folder automatically.

**4. Screenshots — dispatch, don't forget.** **Mobile Screenshots (iOS)** with `upload = true`
and the locales input left empty. It now fans out to 4 app locales × 3 devices (12 macOS shards,
so budget the time). The Fastfile gate refuses to upload unless all five App Store locale folders
are complete, so a half-captured run fails loudly instead of shipping a partial set. Until this
runs, the German listing is live with English screenshots.

**5. Play screenshots stay English.** `SUPPLY_LOCALE` pins them to the `en-US` slot and the
Android capture script has no locale axis. That's already true for `es-ES` and `fr-FR`; German
inherits it. Fixing it is a separate piece of work.

**6. Every future release now needs four release-note translations.** The changelog generator is
English-only, so `release_notes.txt` and `changelogs/default.txt` have to be translated by hand
per locale or a locale silently ships the previous release's notes. Play caps each changelog at
500 characters and German runs long — the current one is at 482. This is written into
`docs/mobile-store-release.md`.

marcodejongh:
Both fixed.

**Locale-shaped dirname guard** — added, but with a looser pattern than `xx-XX`: Play accepts bare languages (`es`, `fil`) and script subtags (`zh-Hans`), so `^[a-z]{2,3}(-[A-Za-z0-9]{2,8})*$` is the boundary. Verified against `en-US`, `es-ES`, `pt-BR`, `zh-Hans`, `fil` and `de-DE` (all stage) versus `_default`, `templates`, `images` (all warn and skip).

**Sync enforcement** — worth doing, so it's now a test rather than a comment: `scripts/mobile-locales-parity.test.ts`. It asserts (a) every `SUPPORTED_LOCALES` language has a `CFBundleLocalizations` entry, (b) the set of `locales/*.json` files equals the declared set, (c) the `locales` map points at a real path per language, (d) every language translates the same key set as `en`, and (e) no usage description is blank. Negative-tested: deleting one key from `de.json` fails it.

It reads `app.config.ts` as text rather than importing it — the config is a factory that reads env and resolves plugins, so evaluating it here would couple the test to the whole native config surface.

marcodejongh:
The cache-key docs point is right and is fixed — `CLAUDE.md` and `AGENTS.md` now list `locales/**` among the native inputs. That prose list is what the "when adding native config, confirm it's covered by one of those globs" instruction gets checked against, so a stale list is worse than no list.

The subtitle finding is a misread, though. The 79-character string quoted is `fastlane/metadata/android/de-DE/short_description.txt` — Play's field, 80-char limit, 78 used. The iOS subtitle is:

```
$ cat fastlane/metadata/de-DE/subtitle.txt
Board beleuchten, mit Crew          # 26 chars
```

which is exactly the replacement the review suggests. Every field was measured against its own cap before the first push, newline-inclusive:

| file | chars | cap |
| --- | --- | --- |
| `de-DE/subtitle.txt` | 26 | 30 |
| `de-DE/keywords.txt` | 97 | 100 |
| `de-DE/promotional_text.txt` | 166 | 170 |
| `de-DE/description.txt` | 2376 | 4000 |
| `de-DE/release_notes.txt` | 1887 | 4000 |
| `android/de-DE/short_description.txt` | 78 | 80 |
| `android/de-DE/full_description.txt` | 2646 | 4000 |
| `android/de-DE/changelogs/default.txt` | 482 | 500 |

Worth noting there's no automated guard on these — `deliver`/`supply` only reject at upload time. If it'd be useful I can add a checker alongside the locale-parity test, though that's scope beyond this PR.

marcodejongh:
Reversing myself on the last paragraph — I've added the limit checker rather than leaving it as an offer.  covers both stores' per-field caps (Apple name/subtitle/promo/keywords/description/release_notes, Play title/short_description/full_description/changelogs) plus the keywords-are-unpadded rule, since Apple counts the separators and a space after each comma quietly costs a keyword's worth of budget.

The reason to do it now rather than later: the failure mode is a **post-merge** upload rejection from a workflow nobody is watching, and translations are exactly where it bites — German and French run 20-30% longer than the English they're translated from, and `short_description` (80) and the Play changelog (500) have almost no slack. This PR is what introduces a fourth locale into that squeeze.

Negative-tested by planting the 78-char Play short description into the 30-char iOS subtitle — i.e. the exact scenario the last review believed it had found. It fails with the file name and both numbers:

```
AssertionError: expected { file: 'de-DE/subtitle.txt', ... }
-   "within": true,
+   "within": false,
```

marcodejongh:
Both points fixed, and the second one turned out to be more interesting than it looked.

**Fragile regex parsing** — right, and fixed at the source rather than by hardening the regex. `app.config.ts` now exports a single `APP_LOCALIZATIONS` const that drives both `CFBundleLocalizations` and the `locales` map, and the parity test imports it. No pattern-matching over the config file at all, so a formatter pass can't break it. Verified the resolved Expo config is byte-identical (`expo config --type public`) and the fingerprint is unchanged with vs without the refactor.

**STORE_READY_APP_LOCALES subset** — I went to check this expecting the `readonly Locale[]` annotation to already enforce it, since `Locale` derives from `SUPPORTED_LOCALES`. It doesn't, and the reason is worth flagging beyond this PR:

> **The root `scripts/` directory is not typechecked by anything.** There is no `typecheck:scripts` task in `vite.config.ts`'s `typecheck` `dependsOn` list — every workspace package is there, the root scripts dir isn't. Vitest transpiles without typechecking, so nothing catches type errors in `scripts/*.ts`.

Proved it by putting `'it'` into `STORE_READY_APP_LOCALES: readonly Locale[]` and running `vp run typecheck` — exit 0. So that annotation is documentation, not a constraint, and the same is true across the whole tooling layer (`mobile-screenshots`, the fingerprint checks, changelog, release-notes…).

Covered here with a runtime assertion in the screenshots test instead, negative-tested (`{ appLocale: 'it', supported: false }`). Adding `typecheck:scripts` repo-wide is the real fix but it's a separate change — it'd likely surface pre-existing errors across ~50 files, and that doesn't belong in a store-listing PR.

marcodejongh:
Fair catch on the confusing error message — fixed, but by documenting rather than renaming.

Two reasons not to rename: `overrideHashKey` **feeds the hash**, so a rename moves the fingerprint for zero behaviour change; and the name is accurate about what the files are — they really are iOS InfoPlist locale strings, and Android takes nothing from them.

The actual problem was the error message lacking context, so that's what I fixed. `mobile-fingerprint-inputs-check.ts` now carries the explanation right where the assertion lives:

> Required on BOTH platforms despite the iOS-sounding name. The files are iOS-only in effect (Expo turns them into `<lang>.lproj/InfoPlist.strings`; Android takes nothing from them), but they're referenced by `app.config.ts`, which is shared — so both fingerprints must track their contents or an Android OTA could ship against a config the binary never had. An "android: expected exactly one iosInfoPlistLocales source" failure is therefore real, not a platform mix-up.

Post-merge screenshot dispatch is captured in the go-live checklist above, along with the App Store Connect step that has to happen first (the `ios metadata` lane only writes onto a version already in "Prepare for Submission").

PR #4183, opened 2026-08-03
fix(mobile): measure insets at the right surface — tab bar overlap for the Start capsule, sheets, toasts

marcodejongh:
## Summary

Two mirror-image regressions from one root cause: on iOS 26 native tabs there are **two safe-area sampling points with different semantics**, and code kept reading the wrong one. expo-router's `NativeTabsView` wraps each tab's content in a nested `SafeAreaProvider` whose bottom inset folds in the tab bar + BottomAccessory + minimize state (the device-verified 139 = 34 + 49 + 56 was measured *there*), while the root-mounted `BottomChromeMetricsProvider` samples the window inset — home indicator only. Every prior fix in this family (#3967, #3973, #4089, the toast repair) guessed at one sampling point and broke the other.

**Tab content positioned with the root inset (too little)** — the pre-session Start capsule sat under the tab bar; after #4089, toasts, snackbars, and list padding lost their tab-bar clearance too:

- **`NativeTabContentInsetProbe`** — mounted in each phone tab layout, inside the per-tab provider, publishes the real in-tab inset to a module store. Focus-gated: expo-router renders every tab's screen, and an unfocused tab's detached view can report a bar-less inset that would reinstate the bug through a last-writer-wins store.
- **`computeBottomChromeMetrics`** gains `measuredTabContentInsetBottom`. On the native-overlay path all offsets anchor to the measurement; the constant reconstruction (root inset + `TAB_BAR_HEIGHT` + `NATIVE_BOTTOM_ACCESSORY_HEIGHT`) survives only as the pre-measurement fallback. `scrollBottomPadding` floors at the un-minimized reconstruction so list padding never shrinks mid-scroll. Floating surfaces track the live value, so `minimizeBehavior` and future UIKit chrome changes are handled automatically.
- **The capsule fix**: `preSessionFooterBottom` / `inSessionListBottom` (liquidGlass) now clear the native chrome instead of the bar-less root inset. Snackbars, FABs, the session-detail footer, and ~19 scroll screens are fixed through the shared `tabBarBottom`. **`Toast`** (mounts above the metrics provider) reads the same store.

**Sheet content positioned with the in-tab inset (too much)** — the inverse: a native bottom sheet docks *over* the tab bar, but a sheet mounted inside a tab inherited the per-tab 139pt inset for chrome it covers, floating the climb filter sheet's Apply button ~105pt up into the sheet (the #3776 "dead gap"):

- **`WindowInsetPublisher`** (root layout, where the provider IS the window) publishes the window inset; **`useWindowBottomInset()`** serves it to the sheet family — the `Sheet`/`ModalSheet` wrappers, `ClimbFilterSheet`, `HoldFilterPicker`, `LogbookFilterSheet`, `LogbookEntryChooserSheet`, `EndSessionSheet`, `InviteSheet`, `CreateDrawer` — with the local inset as the pre-publish fallback. Root-mounted sheets are unaffected (both values identical); the hook makes the mount point irrelevant.

**Guardrails so this stops recurring**: a permanent "Bottom chrome diagnostics" overlay (More → Diagnostics; dev, EAS preview builds, and `pr-<N>` OTA channels) showing all three sampling points, the publish counter, and every derived offset; test fixtures labeled DEVICE_VERIFIED vs INFERRED with the measured axis swept across the invariant matrix; and the sampling-point contract written into `bottom-chrome-metrics.ts` and `docs/mobile-sheets-vs-routes.md`.

JS-only — rides the `pr-<N>` OTA preview channel to existing 2.3.0 installs.

## Release Notes

- The Start button on the Record tab sits above the tab bar again on iOS 26.
- The climb filter sheet's Apply button sits at the bottom of the sheet again, without the dead gap below it.
- Toasts, snackbars, and the last rows of lists stay clear of the tab bar.

## Test plan

- `vp run test:mobile` — 5061 tests pass, including new suites for both inset stores, the focus-gated probe, the window-inset hook fallback contract, provider fan-out on publish, and the measured-axis invariant matrix.
- Device QA on an iOS 26 iPhone via this PR's OTA preview channel (What's New → "Try a preview"):
  1. Record tab, no session, no climb → Start capsule fully above the tab bar, tappable.
  2. Set a climb (accessory appears) → capsule above the accessory platter; rapid climb set/clear settles correctly.
  3. In session → last list row clears the accessory; queue snackbar clears chrome.
  4. Scroll a long list (bar minimizes) → capsule follows the bar; no scroll jumps or content shifts.
  5. Push session detail → FAB above the bar. Trigger a toast on a tab page → above the bar. Rotate → settles.
  6. **Climbs tab → Filters** → Apply button pinned at the sheet bottom with only the home-indicator gap (with and without a current climb set). Same check on the logbook filter sheet and end-session sheet.
  7. Enable More → Diagnostics → "Bottom chrome diagnostics": report the (root, probe, window) values per state and the publish count for one minimize/expand cycle (a large jump means the store's coalescing should be enabled before un-drafting).
- Follow-up commit will upgrade INFERRED fixture labels to DEVICE_VERIFIED with the reported numbers.

https://claude.ai/code/session_01DHCyw4UkJ6Y7NGRbBfNK7C

### DIFF
--- a/docs/mobile-sheets-vs-routes.md
+++ b/docs/mobile-sheets-vs-routes.md
@@ -216,6 +216,16 @@ Is it a secondary surface OVER the current screen, or its own full surface?
    the published measurement via `useBottomChromeMetrics()` — `scrollBottomPadding` for
    list/scroll content, `floatingControlBottom` for absolute overlays, `fixedFooterBottom` for
    docked footers, `preSessionFooterBottom` / `inSessionListBottom` for the session surfaces.
+
+   **Bottom-docked native sheets are the inverse case**: a sheet presents over the tab bar, so
+   the only chrome its content must clear is the **window's** bottom inset (home indicator /
+   gesture bar) — never its mount point's. A sheet mounted inside a tab inherits the per-tab
+   provider, whose 139pt inset floated the filter sheet's Apply button ~105pt up into the sheet
+   (#3776's "dead gap"). Every sheet footer/body bottom pad goes through
+   `useWindowBottomInset()` (`src/hooks/use-window-bottom-inset.ts`, fed by
+   `WindowInsetPublisher` in the root layout); the shared `Sheet` / `ModalSheet` wrappers
+   already do. Sheets mounted at the app root get the same value either way — the hook makes
+   the mount point irrelevant.
    Never hardcode what UIKit "must" have folded into an inset; constants are fallbacks for the
    pre-measurement frames only, and test fixtures carry DEVICE_VERIFIED / INFERRED provenance
    labels (`src/hooks/__tests__/bottom-chrome-metrics.test.ts` — extend its matrix when adding
--- a/docs/mobile-sheets-vs-routes.md
+++ b/docs/mobile-sheets-vs-routes.md
@@ -195,9 +195,33 @@ Is it a secondary surface OVER the current screen, or its own full surface?
 
    A **pushed route** under `NativeTabs` is different: the native tab bar remains, but route
    classification deliberately unmounts the BottomAccessory / queue chrome on screens such as
-   session detail. Its bottom layout must therefore trust the raw UIKit safe-area inset for the
+   session detail. Its bottom layout must therefore trust the UIKit safe-area inset for the
    chrome that is actually present; it must not carry the accessory reserve forward from the tab
    root or add the tab-bar height a second time.
+
+   **Bottom-chrome geometry contract.** "The UIKit inset" is ambiguous — there are two
+   sampling points with different semantics, and conflating them is the recurring bug class
+   behind #3967/#3973/#4089 and the Start-capsule regression:
+   - The **root** `SafeAreaProvider` (what `BottomChromeMetricsProvider` in `app/_layout.tsx`
+     samples) reports the _window_ inset: home indicator only. UIKit tab-bar chrome never
+     reaches it.
+   - Each tab's content sits inside a **nested per-tab `SafeAreaProvider`** (expo-router's
+     `NativeTabsView` wraps every tab in one); _that_ inset folds in the tab bar, the
+     BottomAccessory, and the live minimize state. `NativeTabContentInsetProbe` (mounted in
+     every phone tab `_layout`, focus-gated) publishes it through
+     `src/lib/native-tab-content-inset-store.ts` so root-level consumers position with the
+     measurement instead of a reconstruction.
+
+   Rules: position against the inset measured at the surface you're positioning in, or consume
+   the published measurement via `useBottomChromeMetrics()` — `scrollBottomPadding` for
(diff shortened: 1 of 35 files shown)


marcodejongh:
Re the review's two observations:

1. **Epsilon lag during minimize** — keeping `PUBLISH_EPSILON = 0.5` as is. It only suppresses deltas under half a point, which is below any real per-event step UIKit delivers (device pixel steps are ≥1/3pt even at 3×), so a floating control can lag by at most an imperceptible fraction of a point. The device-QA gate on the overlay's publish counter is the check that decides whether coalescing is needed; if QA shows visible lag instead, that's the moment to revisit.

2. **Multi-probe race test** — added in 873431c: two probes mounted at once, the unfocused one reporting a bar-less inset that must never reach the store, then a focus handoff flipping the winner.

Still draft pending the on-device QA pass from the PR description.

---

marcodejongh:
Re the second review pass:

1. **Content-keyed overlay rows** — real, fixed in 8f85e25 (rows now keyed by a stable label).

2. **Sanity-guard test** — already covered: `bottom-chrome-metrics.test.ts` → "rejects a physically impossible measurement (smaller than the root inset)" feeds `measuredTabContentInsetBottom: 20` with `insetsBottom: 34` and asserts the fallback wins.

3. **`usesSidebar` + non-null measurement** — already covered by the matrix: "produces identical metrics for every measured value off the native-overlay path" sweeps every `usesSidebar: true` cell against measured ∈ {83, 139} and pins that the measurement is ignored there (that invariant is what surfaced the missing `!usesSidebar` term during development).

---

marcodejongh:
Re the third review pass — no code changes, with reasons:

1. **i18n suppression format** — `// i18n-ignore-next-line` is the marker the existing tester-only rows in this same file already use (Feature Flags, Branch Switcher, the Preview Build section), and lint + the full CI run were green on the first commit, which already contained these strings. `check:i18n` also only scans `packages/web/app/`.

2. **Floor jump on accessory mount** — a bar-only floor wouldn't change that case: when the accessory mounts, the *measured* in-tab inset itself grows to 139 and wins the `max()` on its own, because the last list row genuinely has to clear the platter — that padding growth is required behavior (and mirrors the long-standing `jsQueueReserve` jump on the JS-bar path when a climb is selected). Weakening the floor would only hurt the pre-measurement fallback state, where it's what keeps the last row from hiding behind the platter. The floor's job is narrower: keep padding stable against the *minimize* shrink, which shares a gesture with scrolling.

3. **DEVICE_VERIFIED before merge** — agreed and already the gate: the PR stays draft until the on-device numbers arrive (that's the QA section's ask). One nuance: the matrix tests pin pass-through semantics — they feed 83 and assert 83 comes back — so a device reporting a different bar-only inset wouldn't fail them; only the INFERRED comment labels and the fallback reconstruction constants are the assumptions being confirmed.

---

Issue #4186, opened 2026-08-03
chore(db): run merge-clone-boards against prod


marcodejongh:
Run the one-off `merge-clone-boards` script against prod once it lands. This is a manual data operation, not something CI should ever fire.

## Why

BLE serial resolution mints a private "clone" board whenever no board carries the controller's serial. Most synced gym boards have no serial (2,243 of 3,152 gym-linked synced boards), so the first climber to connect at a gym creates a clone and everyone after them lands on it — a different board than the map finder gives.

Worked example, 9 Degrees Morningside (2026-08-03):

| time | surface | board |
|---|---|---|
| 03:40:18 | BLE connect | 69 — *9 Degrees Morningside - Kilter Board Original* |
| 03:45–03:57 | 5 climbs reported | **311523** — *"Kilter Board"*, private, owned by a stranger, no gym |
| 03:46–03:47 | wall sheet, 71 durable history rows | **311523** |
| 03:59:09 | BLE reconnect, 37 climbs | 69 |
| 04:16:08 | "load older" on board 69 | **0 rows** — the wall's real history sits on 311523 |

The code fix stops new clones. This issue covers moving the history that already landed on them.

## Blast radius (prod, 2026-08-03)

- **166** clone boards
- **11,946** `board_climb_events` rows and **1,948** `boardsesh_ticks` stranded on them
- **274** users affected
- **660** of 1,662 linked `user_board_serials` rows (576 users) resolve to a board other than the one the serial picks
- **321** gym boards have a recoverable serial sitting in `user_board_serials` (247 distinct serials)

## What to run

`packages/backend/src/scripts/merge-clone-boards.ts` — backend script (needs both `DB_URL` and `REDIS_URL`). Defaults to dry-run.

1. `--backfill-serials` — stamp the 247 recoverable serials onto their gym boards. Independent and safe on its own; run it first.
2. `--report` — dry-run the merge, dump the per-pair CSV, eyeball it.
3. Restore a prod snapshot, `--apply` there, verify.
4. `--apply` against prod, off-peak.

## Checklist

- [ ] `--backfill-serials` dry-run reviewed
- [ ] `--backfill-serials --apply` on prod
- [ ] Merge `--report` CSV reviewed (166 pairs)
- [ ] `--apply` rehearsed on a restored snapshot
- [ ] `--apply` on prod, off-peak
[...]

PR #4187, opened 2026-08-03
fix(board-presence): keep the picked board when Bluetooth connects

marcodejongh:
## Summary

Connecting over Bluetooth could put you on a different board than the one the map finder gave you. Reported from 9 Degrees Morningside, reproduced in the prod data:

| time (2026-08-03) | surface | board |
|---|---|---|
| 03:40:18 | BLE connect | 69 — *9 Degrees Morningside - Kilter Board Original* |
| 03:45–03:57 | 5 climbs reported | **311523** — *"Kilter Board"*, private, owned by a stranger, no gym |
| 03:46–03:47 | wall sheet + 71 durable history rows | **311523** |
| 03:59:09 | BLE reconnect, 37 climbs | 69 |
| 04:16:08 | "load older" on board 69 | **0 rows** — the wall's real history sits on 311523 |

**Root cause.** `user_boards` row 69 (the gym's synced listing) has `serial_number = NULL`, and so do **2,243 of 3,152** gym-linked synced boards. `findActiveBoardsBySerial` matches on that column alone, so the first climber to BLE-connect at a gym fell through to `bindOrCreateOwnBoardForSerial`, which minted a **private duplicate** of the wall under their own account (board 311523, 2026-07-06). Every climber after them matched the serial to that duplicate while the map finder kept pointing at the gym listing.

Two binders also race on connect — `resolveBoardForUuid` from the board picker (`drawer-host-provider.tsx:407`) vs `resolveBoardCandidatesForSerial` from BLE (`bluetooth-provider.tsx:789`, whose `beginResolution` clears the current binding first) — as do two writers to `user_board_serials.board_uuid`: `recordBoardSerial` writes the route's board, `rememberBoardForSerial` writes the serial-matched one. Whichever landed last decided the wall, differently each connect.

### The resolution rule

`resolveBoardForSerial` / `resolveBoardCandidatesForSerial` take an optional `selectedBoardUuid` — the board the caller already had picked. It is used only when its board type / layout / size / normalized hold sets match the controller that just connected; a selection that doesn't describe the connected wall is stale and ignored.

**A gym's own listing wins when it carries the serial *and* describes the connected wall.** That listing *is* the wall; a selection is only a hint about it, and most personal boards are the auto-named onboarding template (`Kilter Original 12×12 with kickboard`) that travels with the climber between gyms. The config half matters because serials are reused: a listing at some other gym with a different layout / size / hold sets says nothing about the controller in front of this climber and must not outrank the board they picked. The selection therefore decides whenever no matching gym listing claims the serial — which is the Morningside case, since most synced gym boards carry no serial at all.

**A stale remembered link no longer outranks either.** Once a matching gym listing carries the serial, a remembered board that doesn't carry it is skipped, so picking that listing on the map routes straight to it with no prompt.

**When the selection does decide and carries no serial, the connect claims it.** That is what makes a gym's listing discoverable by serial for everyone afterwards, instead of the next climber minting another duplicate. Guarded three ways: config must match, the board must be the caller's own or system-owned (nobody can brand a stranger's wall and redirect their serial), and `WHERE serial_number IS NULL` makes a concurrent claim fail-safe. A board that already carries a different serial is bound but never rewritten — a swapped controller is the owner's call.

**Serials shared by two gym listings prompt rather than guess.** Falling through hands them to `findActiveBoardsBySerial`, which returns every match; the client shows the existing disambiguation sheet and the pick is remembered per user, so it appears once per wall. Silent guessing is exactly what the original bug looked like. There are 15 such serials — only 9 are genuine supplier reuse (4.3 km to 7,429 km apart); 3 are duplicate board rows inside one gym, 2 are duplicate gym rows for one address, and `12345` is a placeholder someone typed. Broken out in #4190.

**No more clobber.** `recordBoardSerial` omits `boardUuid` from its conflict `SET` when the connect resolved no board, so a bare reconnect can't wipe the link the presence resolver just wrote.

**Mobile** threads the selected board through `handleConnectSuccess`; the resolve cache key includes it so switching boards re-resolves.

### Also: the "older history has no metadata" report

Not data loss — every `board_climb_events` row has its `name` / `grade` / `setter` / `frames`. What went missing was **who sent it**: 48 of the 71 rows loaded were sent by a user with `users.name = NULL` and no profile, so `boardHistory` returned a null `sentByDisplayName` and `HistoryRowContent` dropped the entire avatar-and-name block. **9,952 of 144,023** durable rows across 91 nameless accounts (Apple private-relay signups, mostly) render that way. Those rows now keep their attribution with a neutral avatar and a localized fallback.

### Data repair

`packages/backend/src/scripts/merge-clone-boards.ts` — dry-run by default, two phases (`--backfill-serials`, `--merge`). Covers **166 clone boards** holding **11,946 wall events** and **1,948 ticks** across **274 users**, plus **321 gym boards** whose serial is recoverable from existing links.

Merging renumbers the two boards' `board_climb_events` together by `confirmed_at` — the merged-in half is almost always older, and `boardHistory` orders newest-first by `seq`. It parks the two boards into **disjoint** bands first, because `board_climb_events_board_seq_unique` is a plain unique index and both boards usually number from 1 (a test here catches the naive same-offset version). Afterwards it deletes `board:{id}:seq` so `nextBoardSeq`'s dormancy reseed lifts the counter above the new durable floor — without that, an active board keeps issuing colliding seqs that `onConflictDoNothing` silently drops. A clone matching more than one gym listing is skipped with a warning rather than guessed. Clones are soft-deleted, never dropped.

Running it against prod is tracked in #4186 — not part of this PR.

## Release Notes

- Connecting your phone to a board keeps you on the right wall instead of jumping to someone else's board
- Older sends in a wall's history show who lit them again, even for climbers who never set a name

## Test plan

- `vp run typecheck` — clean; `vp check` — clean; `vp run check:i18n` — clean; i18n `catalog-completeness` 90 passed
- `vp run test:mobile` — 5048 passed (579 files): nameless-climber attribution in the panel, `wallDriverForClimb` unit cases, and the selected board reaching the resolver from `handleConnectSuccess`
- `vp test run --project backend` — `board-presence` 95 passed (selection precedence, the serial-claim race and its duplicate-serial skip), `boards-serial-privacy` 26 passed (the `board_uuid` clobber), `merge-clone-boards` 10 passed (seq renumbering, seq collision, follow dedupe, pairing guards, ambiguous-clone skip)
- `vp run check:mobile-bundle` — OK both platforms
- Added `session_boards` to the backend test DDL (`schema-sql.ts`); it exists in prod but was missing from the test schema, so the merge script couldn't be covered without it.

### Review follow-ups

- @ES-Alexander caught that "selection wins" was too broad — 1,459 of 2,307 user-owned boards are the onboarding template, and 216 links / 200 users / 122 serials would have quietly stopped reaching the gym feeds they land on today. Led to the always-redirect rule above.
- Privacy question from the same thread → #4189. Worth noting it isn't introduced here: the serial has always routed climbers onto gym boards.
- Fable review requested for `bluetooth-provider.tsx` — outside CLAUDE.md's listed BLE paths, but Marco's call to run it through the gate anyway. **Done** (a0361bab): four edge gaps closed — the gym-listing veto now has to match the connected config (it matched on serial alone, so a reused serial at an unrelated gym could override a climber's local pick); `recordBoardSerial` fills `board_uuid` only when the stored value is null, so arrival order stops mattering on a first connect; a stale remembered link no longer outranks a serial-carrying listing or a fresh pick; and the merge script resets Redis seq counters per transaction rather than once at the end, so a mid-run crash can't leave merged boards silently dropping sends.
- An automated review round caught a real bug in the repair script: `findSerialBackfills` had no owner guard, and 1,554 user-owned boards carry a `gym_id`, so `--apply` would have stamped serials onto home walls. Fixed with a regression test.

https://claude.ai/code/session_01Rr49aRV79pw13Gp6d8Y6Ei


marcodejongh:
@fable — Bluetooth review requested.

Marco's call: `packages/mobile/src/providers/bluetooth-provider.tsx` isn't in CLAUDE.md's mandatory-review path list (`shared/ble-protocol/`, `mobile/src/lib/ble/`, `live-activity/ios/`, web LED control), but it *is* the BLE-connect orchestrator, so this goes through the gate anyway.

## The entire BLE-adjacent diff

```diff
 packages/mobile/src/providers/bluetooth-provider.tsx

+ // The board the user currently has selected, read at connect time so the
+ // serial resolver can keep them on it. [...]
+ const boardUuidRef = useRef(boardUuid);
+ boardUuidRef.current = boardUuid;

  const resolvePromise =
    serial && serial.length > 0
-     ? resolveAndBindBoard({ serial, boardType, layoutId, sizeId, setIds })
+     ? resolveAndBindBoard({
+         serial, boardType, layoutId, sizeId, setIds,
+         selectedBoardUuid: boardUuidRef.current ?? null,
+       })
      : resolveAndBindBoardByConfig({ boardType, layoutId, sizeId, setIds });
```

Nothing else in the file changes. No protocol writes, no scan/connect/disconnect lifecycle, no adapter, no frame conversion, no characteristic handling.

## What to look at

1. **The ref, not the connection snapshot.** Every other resolver input in `handleConnectSuccess` comes from `connection.config` — the immutable adapter-generation snapshot — precisely so a route change mid-connect can't assign the wrong identity to an older physical link. I deliberately did *not* follow that for `selectedBoardUuid`: "which wall did the user pick" is a property of the moment the connect landed, not of the adapter generation. A stale read is defended server-side (the backend drops any selection whose board type / layout / size / normalized set ids don't match the controller that connected). Is that reasoning sound, or should this also come off the snapshot?

2. **Generation safety.** The resolve promise still flows through the existing `connection.setAnalyticsBoardId(resolved.boardId)` guard, so a late resolve from a replaced controller is still rejected. I believe adding an input can't change that, but it's the thing most worth a second pair of eyes.

3. **Resolve-key churn.** `boardSerialResolveKey` in `board-presence-provider.tsx` now includes `selectedBoardUuid`, so switching the active board while connected re-resolves instead of returning the cached binding. Intended — otherwise the feed stays pinned to the wall you navigated away from — but it does mean one more network resolve on board switch.

## Context

Fixes a wrong-board bug: most synced gym boards carry no `serial_number`, so the first climber to BLE-connect at a gym minted a private duplicate of the wall and everyone after them resolved to it. Full analysis in the PR description. `ota-check` confirms this is JS-only — no fingerprint change, ships over the air.

ES-Alexander:
Noting that sometimes a user might connect to a generic board template instead of a specific gym board, in which case it's potentially desirable to redirect that to a gym that matches the serial number, rather than keeping their selection?

Unsure what's best there - there may also be some users who only care about the type of board they're climbing on, and don't want their sends to contribute to board stats / be usable to track their location across gyms or something.

marcodejongh:
@ES-Alexander yes ai is dumb, it should always redirect.

The privacy issue is still on my mental todo list. But requires more broader thinking

marcodejongh:
Always-redirect is in as of 497aa148e. `selectionBeatsSerialMatch` now has no exceptions — if any synced gym listing carries the serial, it wins, full stop. The selection only decides when nothing claims the serial, which is still the Morningside case (most gym listings carry no serial at all until a connect claims one for them).

I'd flagged a risk that this misroutes across the 15 serials two gym listings share. It doesn't, and I was wrong to weight it: falling through hands those to `findActiveBoardsBySerial`, which returns *every* match, so the client prompts instead of guessing — and the pick is remembered per user, so it shows up once per wall. Silently guessing is what the original bug looked like to a climber.

**Investigated the 15 → #4190.** They're three different problems, not one:

| | count | what |
|---|---|---|
| Duplicate **board rows** in one gym | 3 | `751807` is boards 660 + 661 on the same gym 607, same name, same serial. Beargrass twice over. Sync artefact. |
| Duplicate **gym rows** | 2 | `751879` — "Stone Age" and "Salle d'Escalade - Stone Age", identical address to the metre. One Belgian gym listed twice; `merged_into_gym_id` exists for this. |
| Genuine supplier reuse | 9 (+`12345`) | 4.3 km to 7,429 km apart — Blocabrac ↔ El Cap, Crux 24 ↔ Winchester. Unfixable in our data, must prompt. `12345` is a placeholder someone typed. |

So five of the 15 are our own data quality and will stop prompting once deduped/merged; nine genuinely need the prompt. Total wall history sitting on shared-serial boards is 77 events, so nothing is at risk either way. #4190 has the per-serial table and suggested cleanup, including ranking prompt candidates by distance when the client has location so the right gym is the default rather than a coin flip.

Backend board-presence is 95 tests; the two that encoded the old exceptions now assert the prompt instead.

marcodejongh:
Addressed in 16928f69c, with one correction.

**`rememberBoardForSerial` called with the pre-claim board — not a bug.** It takes `Pick<SerialCandidateBoard, 'uuid' | 'boardType' | 'layoutId' | 'sizeId' | 'setIds'>` and never reads `serialNumber`; the serial it stores is the `serial` argument. `selected` and `board` differ in exactly that one field, so the link row was always correct. Switched to `board` anyway — the two differing at the call site invites this question every time someone reads it.

**Unused parameter — fixed.** `selectionBeatsSerialMatch(serial)` now. Nothing about the selected board changes the answer (that's the whole point of always-redirect), and carrying a board argument implied otherwise.

**Docs — fixed, and the best of the three.** `docs/websocket-implementation.md` listed the three resolve entry points but never said how a serial picks a board, and that rule changed twice in this PR while living only in code comments. There's now a "Which board a serial routes to" section next to the resolve order covering the precedence, the claim and its three guards, and why several matches prompt rather than guess.

95 board-presence tests, typecheck and format clean.

marcodejongh:
Good catch on #1 — that was a real bug. Fixed in ab891050c.

**1. Missing system-owner guard in `findSerialBackfills` — real, and the worst kind.** The query filtered `serial_number IS NULL AND gym_id IS NOT NULL` with no owner check, and **1,554 user-owned boards carry a `gym_id`** (a home wall attached to its own auto-created gym, or one captured through the stray-board flow). `--apply` would have stamped serials onto those, redirecting each for every climber connecting afterwards — precisely the failure this script exists to undo. `findClonePairs` already guarded its target the same way; the backfill didn't. Guard added, plus a test that fails without it (#3).

**2. Two round-trips — fixed, though not by deferring.** The check was already deferred: `if (selected && (await selectionBeatsSerialMatch(serial)))` short-circuits, so it never ran when the selection was absent. The cost was on the path you'd actually care about — a *present* selection paid both sequentially. They're independent, so they now overlap via `Promise.all`. Wasted lookup when the selection turns out stale is one indexed row.

**4. `clonePark` collision when `maxTargetSeq = 0` — can't happen, defended anyway.** `maxTargetSeq` comes from `COALESCE(MAX(seq) FILTER (...), 0)`, and seq starts at 1 (Redis `INCR`), so `maxTargetSeq = 0` means the target has **no rows** — nothing for the clone band to collide with. But that's a proof the next reader has to reconstruct, so the clone band now sits a fixed `SEQ_PARK_OFFSET` above the target's. Free on a bigint, and the disjointness is visible rather than argued.

Merge tests 12, board-presence 95, typecheck and format clean.

marcodejongh:
Fable reviewed this and found four gaps, all in the edges of the resolution rule. Fixed in a0361bab.

**1. `recordBoardSerial` could still clobber the resolver's link** (major). Omitting `boardUuid` from the conflict `SET` only covered the "resolved nothing" case — the mobile and web connect paths still pass the *route's* board uuid, so on a first connect where the gym listing vetoes the selection, the two writers raced and last-writer-won. If the recording landed last, `user_board_serials.board_uuid` pointed at the personal template, and the saved-board early return meant it was never rewritten: permanent wrong-wall pinning, no prompt, no way out. The conflict `SET` now writes `coalesce(user_board_serials.board_uuid, <new>)` — fill-only-when-null. The recording still establishes the link for a client that never reaches the resolver, and the resolver wins in either arrival order.

**2. The gym-listing veto ignored board config** (major). `hasSyncedGymBoardForSerial` matched on serial + system owner + gym id alone, so a listing at an unrelated gym carrying a reused serial (9 genuine cross-gym reuses in prod, more once #4186's backfill runs) vetoed a climber's config-matching local pick and silently routed them to the foreign wall. It now has to describe the connected wall too — board type, layout, size, normalized set ids — before it outranks anything. The local listing then claims the serial, and the existing prompt handles it from there if the serial really is shared.

**3. A stale remembered link beat both the gym listing and a fresh pick** (major). `findChosenBoardForSerial` returned whatever `board_uuid` pointed at, with no check that it still carries the serial — so the ~200-user template cohort landed on the template even when they explicitly picked the gym's wall on the map, contradicting rule 1 in the docs. Now: once a config-matching gym listing carries the serial, a remembered board that doesn't carry it is skipped (the single listing routes, or the prompt fires), and a selection that *is* the serial-carrying gym listing routes straight there with no prompt. Docs updated to match.

**4. The merge script's Redis reset wasn't crash-safe** (major, also raised by the review bot earlier). One `resetBoardCaches` after ~166 transactions meant a mid-run crash left every already-merged board with a stale `board:{id}:seq` below its new durable floor — colliding seqs, `onConflictDoNothing` dropping every new send on live gym walls, and no rerun repair since merged clones are soft-deleted. The reset now follows each transaction (`mergeAllPairs`), Redis is pinged before any data moves, and a failed reset names the board id in the error.

**5. `claimSerialForBoard`'s soft-delete edge** (minor, nice-to-have): documented rather than changed — the degenerate case binds one connect to a just-deleted board and self-heals on the next, since `findChosenBoardForSerial` joins on `deleted_at IS NULL`.

Tests: `board-presence` 100 passed (foreign-gym-listing veto, stale-link heal with one candidate and with several, fresh pick over stale link, and a real-DB test running `recordBoardSerial` and the resolver in both orders), `merge-clone-boards` 11 passed (per-pair reset proven to land before the next merge starts), `boards-serial-privacy` 28 passed (the conflict `SET` renders the COALESCE). `vp run typecheck` + `vp check` clean.

marcodejongh:

marcodejongh:

Conflicting: a 10-file conflict spanning resolvers, mobile providers, and the GraphQL schema + generated types — the board-presence area has diverged a lot from main.

PR #4197, opened 2026-08-04
feat(mobile): report the board-sheet switch-board tap

marcodejongh:
## Summary

A session on 2026-08-03 (13:39–15:20 Brisbane) logged `Board Sheet Opened` **3×** — 03:46:34, 04:15:59, 04:16:07 UTC — and `Board Swap Invoked From Sheet` **0×**, with the active board stuck on `kilter` for the whole session. The app was alive throughout: 5× `Climb Sent to Board Success` at 04:17, ticks at 04:24/04:43/04:51/04:56, a clean `Session Ended` at 04:57. So switching boards genuinely did nothing, and `track()` is the first statement in `handleSwitchBoardFromSheet`, so the handler never ran.

What the telemetry can't say is **why**. Two very different bugs produce the same silence:

1. the control was never reachable — occluded, laid out past the sheet's detent, or swallowed by a wedged presentation;
2. the press landed and the handler chain broke between the `Pressable` and the drawer host.

This adds `Board Swap Tapped`, fired from the switch-board `Pressable`'s own `onPress` before `invalidatePendingActions()` and before the host handoff. The **pair** is the diagnostic:

| `Board Swap Tapped` | `Board Swap Invoked From Sheet` | reading |
| --- | --- | --- |
| ✗ | ✗ | the control was never reachable |
| ✓ | ✗ | handler chain broke in between |
| ✓ | ✓ | working |

`historyCount` rides along because the footer sits *below* the history list in `NowOnTheWallPanel`. If dead taps correlate with long history, that points at layout rather than hit-testing.

### One thing this deliberately does not fix

There is a visible asymmetry worth a look on device, but not worth a speculative change without a repro: in the `variant === 'column'` branch the history list gets `style={styles.fillList}` (`flex: 1`) with the comment *"Fill the bounded height between the header and the switch-board footer"* — the **`variant === 'sheet'` branch has no such style**, while the sheet is a fixed `snapPoints={['55%', '92%']}` with `enableDynamicSizing={false}`. If the list is unbounded there, it grows with content and pushes the footer past the detent.

Quick check on a real device: open the board sheet on a wall with 20+ climbs of history and see whether the "Switch board" footer is still visible and tappable at the 55% detent. If it isn't, that's the bug, and the fix is to bound the list the way the column branch does.

## Release Notes

## Test plan

- `vp run test:mobile --reporter=agent` — 5065 passed across 582 files.
- New test in `now-on-the-wall-panel.test.tsx` asserts the **ordering** (`track` before `onSwitchBoard`), not just that the event fires — if it ever fires alongside the host handler instead of ahead of it, it stops discriminating. Also asserts `historyCount`. Added `analytics.track.mockReset()` to `beforeEach` since the new test installs an implementation.
- `vp check` — no findings in the changed files; the repo-wide error/warning counts are unchanged from `origin/main`.
- The callback reads `historyCount` from a ref rather than listing `combinedHistory.length` in its deps, per the perf checklist in `docs/react-native-performance.md`.

Not verified: no device run — this box is Linux, so the footer-reachability question above is still open.

https://claude.ai/code/session_01HqJiEtn1BFUDKfS4w91A3W

### DIFF
--- a/packages/mobile/src/components/board-presence/NowOnTheWallPanel.tsx
+++ b/packages/mobile/src/components/board-presence/NowOnTheWallPanel.tsx
@@ -470,19 +470,13 @@ function NowOnTheWallPanelComponent(
     onClose?.();
   }, [invalidatePendingActions, onClose]);
 
-  // Mirrored into a ref rather than listed as a dep: a list `.length` in a
-  // callback's deps rebuilds it on every page of history (see the perf checklist
-  // in docs/react-native-performance.md). Assigning during render is the standard
-  // latest-value ref pattern.
+  // A ref, not a dep: a list `.length` in a callback's deps rebuilds it on every
+  // page of history (docs/react-native-performance.md).
   const historyCountRef = useRef(combinedHistory.length);
   historyCountRef.current = combinedHistory.length;
 
-  // Track BEFORE `invalidatePendingActions()` and before handing off to the host,
-  // so this event means "the tap reached JS" and nothing more. Paired with the
-  // host's BoardSwapInvokedFromSheet it separates an unreachable control from a
-  // broken handler — see the SHARED_EVENTS.BoardSwapTapped note. `historyCount`
-  // rides along because the footer sits below the history list, so a dead tap
-  // correlated with a long list points at layout rather than hit-testing.
+  // Tracks before any other work, so it means "the tap reached JS" and nothing
+  // more — see the SHARED_EVENTS.BoardSwapTapped note for what the pairing buys.
   const handleSwitchBoard = useCallback(() => {
     track(SHARED_EVENTS.BoardSwapTapped, {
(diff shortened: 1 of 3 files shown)


marcodejongh:
On the second note from the first review — the ordering test not asserting `invalidatePendingActions` was called: leaving that alone deliberately. It's a local callback with no observable the test can reach directly, and asserting it would mean reaching through the imperative handle to check that pending actions were cleared, which is what the "drops in-flight action results" test already covers on its own path. Adding a weaker proxy assertion here would make the test look like it pins more than it does.

Comment verbosity fixed in `a55c9c7` — the rationale now lives only in the `SHARED_EVENTS.BoardSwapTapped` entry.

marcodejongh:
Correction to the comment above: the verbosity fix is `553d563`, not `a55c9c7`.

PR #4198, opened 2026-08-04
fix(mobile): [native-train] stop the bottom-accessory relayout crashing the app

marcodejongh:
## Summary

Fixes [BOARDSESH-9K](https://boardsesh.sentry.io/issues/BOARDSESH-9K) — `EXC_BREAKPOINT: AnimationKit/InProcessAnimationState.swift:122: Fatal error: Missing animationAndComposerGetter`, error body `Stack overflow`, culprit `-[RNSTabsHostComponentView rnscreens_relayoutBottomAccessoryIfAttachedAfterAppearance]`. 27 events / 6 users since 2026-07-07; one user crash-looped **13 times** on 2026-08-03 between 18:59 and 21:14 UTC, three of those within 3–9 seconds of app start.

That symbol is ours — added by `patches/react-native-screens@4.26.2.patch` to work around iOS 26 not laying in a bottom accessory attached after the tab bar has appeared.

### What actually goes wrong

`applyBottomAccessoryVisibility` runs from `updateContainer`, i.e. **inside a React Native mounting transaction**. The patch called `-layoutIfNeeded` synchronously right there. Reading the captured stack downward:

```
rnscreens_relayoutBottomAccessoryIfAttachedAfterAppearance   ← our synchronous layout
  → -[UITabBar layoutSubviews] → -[UITabBar _minimizeBehavior]
    → -[UIViewFloatAnimatableProperty setValue:]              ← inside the sheet's
      → -[UISheetPresentationController _sheetLayoutInfoLayout:]  alongside-animation block
        → -[UITabBar layoutSubviews] → _minimizeBehavior (again)
          → _assertionFailure
```

It is a UIKit layout/animation feedback loop between a presenting/dismissing `UISheetPresentationController` and the tab bar's minimize machinery, and our synchronous layout is what draws it in. `_minimizeBehavior` and `_sheetLayoutInfoLayout` each appear twice in the 99 captured frames. On iOS 27 an AnimationKit assertion trips first; underneath it is a plain stack overflow — every reported event is iOS 27.0, which is consistent with iOS 26 blowing the stack without ever getting a crash report out.

### The fix

Three changes, all inside the patch:

1. **Nudge on attach only.** The patch exists for "accessory attached after appearance never lays in". The detach branch (`setBottomAccessory:nil`) is what fires on every route push that leaves an accessory surface, and UIKit removes the accessory itself — that call was pure risk.
2. **Never lay out synchronously.** A single coalesced `dispatch_async` hops out of the mounting transaction. This also still covers the async-sizing case the nudge was written for (accessory content is sized by the shadow-state proxy, so an immediate pass can lay in a zero-size accessory before its frame KVO lands) — that was already handled by a deferred re-assert, which this replaces rather than adds to.
3. **Never lay out during a transition.** If a transition coordinator is active when the deferred pass fires, hand the layout to `animateAlongsideTransition:completion:` instead of running underneath it. Bounded: a coordinator completion fires once, and by then the transition is over.

Plus a `BOOL` re-entrancy/coalescing ivar, so several `applyBottomAccessoryVisibility` calls in one transaction schedule one pass.

Also corrects a stale reference in `(tabs)/_layout.tsx` that still pointed at `react-native-screens@4.25.2.patch`.

## Why `[native-train]`

`patches/**` is a native-fingerprint input, so this cannot reach store binaries over OTA and `main` must stay OTA-compatible. Draft until it rides a native build.

## Release Notes

- Fixed a crash that could take the app down when the now-playing bar appeared while a sheet was opening or closing

## Test plan

- `vp run check:mobile-patches` — **OK, 6 native patches applied.** The patch was regenerated from a pristine 4.26.2 tree (reverse-apply the old patch → commit → edit → `git diff`) rather than hand-edited, so the hunk headers are generated, not computed. Verified it applies clean to pristine with `git apply --check`, and that `bun install` lands the new content in `node_modules` (new store hash `f697aa4d8f1505cd`) with **`bun.lock` unchanged**.
- `vp run test:mobile --reporter=agent` — 5064 passed across 582 files.

### Not verified — please read before merging

**The Objective-C++ is not compiled anywhere in this test plan.** I'm on Linux; there is no Xcode here, and `check:mobile-patches` only proves the patch *applies*, not that it builds. This needs a macOS build (`vp run mobile:ios`) before it goes anywhere near a native train.

Also unverified: the iOS 27 repro itself (present a sheet over the tab bar with the accessory attaching post-appearance), and that the original "accessory attached after appearance never lays in" bug stays fixed — the deferred-only path costs at most a frame in theory, but that is exactly the kind of thing that wants a device check. Worth confirming the accessory still appears the moment the first climb of a session is set.

One alternative not taken: `minimizeBehavior="onScrollDown"` in `(tabs)/_layout.tsx` is a JS prop, sits in the fatal cycle, and could be defaulted on iOS 27 as an OTA-shippable mitigation for anyone stuck on the current binary. Worth testing against the repro if the native train is far out.

https://claude.ai/code/session_01HqJiEtn1BFUDKfS4w91A3W

### DIFF
--- a/.github/workflows/ios-rn-ci.yml
+++ b/.github/workflows/ios-rn-ci.yml
@@ -76,7 +76,10 @@ jobs:
     # macos-26 is GA on GitHub-hosted runners and matches the RN TestFlight
     # workflow's Xcode line; the PR workflow has passed on this label.
     runs-on: macos-26
-    timeout-minutes: 60
+    # 90, not 60: this job now compiles the app scheme (every Pod) on top of the
+    # Swift test target, and there's no DerivedData cache — only the Pods sources
+    # are cached. Native-input PRs only; the gate skips JS-only ones entirely.
+    timeout-minutes: 90
 
     env:
       WORKSPACE_PATH: packages/mobile/ios/Boardsesh.xcworkspace
@@ -135,6 +138,59 @@ jobs:
       - name: Install CocoaPods dependencies
         working-directory: packages/mobile/ios
         run: pod install
+
+      # The BoardseshTests scheme below is a STANDALONE Swift target built from
+      # copied sources — it never links the app and never compiles a single Pod.
+      # So until this step existed, a workflow whose path filter and native gate
+      # both key on `patches/**` compiled exactly zero lines of the native module
+      # code those patches change: a run over an unreviewed Objective-C++ patch
+      # went green with no `CompileC` anywhere in its log (caught on #4198).
+      #
+      # Building the app scheme is what actually compiles the Pods — RNScreens,
+      # Reanimated, the BLE stack, our own modules under packages/mobile/modules.
+      # Runs before the test target is grafted on, so it builds the project exactly
+      # as `expo prebuild` produced it. Signing is off: this is a compile gate, not
+      # something we install (contrast scripts/mobile-build-sim-app.ts, which
+      # ad-hoc signs because that app has to actually run).
+      - name: Build the app (compiles Pods + patched native modules)
+        run: |
+          set -o pipefail
+          build() {
+            xcodebuild build \
+              -workspace "$WORKSPACE_PATH" \
+              -scheme Boardsesh \
+              -configuration Debug \
+              -destination 'generic/platform=iOS Simulator' \
+              ARCHS=arm64 \
+              CODE_SIGN_IDENTITY="" \
+              CODE_SIGNING_REQUIRED=NO \
+              CODE_SIGNING_ALLOWED=NO \
+              2>&1 | tee "$1"
+          }
+          # New Architecture codegen ("Generate Specs") races the compile on a cold
+          # build — "Build input file cannot be found: .../ReactCodegen/*-generated.mm".
+          # Every run of this job is that scenario: a fresh runner with empty
+          # DerivedData. scripts/mobile-build-sim-app.ts retries once for exactly
+          # this reason; a real compile error still fails the retry.
+          build build.log || { echo "::warning::retrying after probable codegen race"; build build.log; }
+
+      # A green build is not proof the patch compiled. Expo's precompiled modules
+      # are ON by default (the generated Podfile sets EXPO_USE_PRECOMPILED_MODULES
+      # ||= 1) and swap a pod's source for a prebuilt xcframework the moment a
+      # tarball resolves. RNScreens has no prebuild today — hence the log line
+      # "prebuilt xcframework unavailable; building from source" — but if that ever
+      # changes, this gate goes green having compiled none of the patch, and the
+      # TestFlight build links the prebuilt too, shipping without it. So assert on
+      # a file the patch actually touches. `buildFromSource` in
+      # packages/mobile/package.json is the belt to this braces.
+      - name: Assert the patched native module really compiled
+        run: |
+          if ! grep -q 'RNSScrollViewFinder' build.log; then
+            echo "::error::No RNScreens source compiled — the patch was not exercised." \
+                 "Check whether a prebuilt xcframework replaced the pod."
+            exit 1
+          fi
+          echo "RNScreens compiled from source — patch exercised."
 
       - name: Add RN Swift test target
         run: node scripts/prepare-rn-ios-tests.mjs
--- a/packages/mobile/app.config.ts
+++ b/packages/mobile/app.config.ts
@@ -275,7 +275,7 @@ export default ({ config, projectRoot }: ConfigContext): ExpoConfig & { newArchE
     name: appName,
     slug: 'boardsesh',
     owner: 'boardsesh',
-    version: '2.3.0',
+    version: '2.3.1',
     scheme: 'com.boardsesh.app',
     orientation: 'portrait',
     icon: iconPath,
--- a/packages/mobile/app/(tabs)/_layout.tsx
+++ b/packages/mobile/app/(tabs)/_layout.tsx
@@ -311,8 +311,9 @@ export default function TabLayout() {
   return (
     // `minimizeBehavior="onScrollDown"` relies on UIKit finding the climbs
     // FlashList's nested scroll view, which it can't by default — that fallback
-    // lives in patches/react-native-screens@4.25.2.patch. `vp run check:mobile-patches`
-    // (CI) fails the build if that patch ever stops applying after a dep bump.
+    // lives in patches/react-native-screens@4.26.2.patch, alongside the
+    // bottom-accessory relayout fix. `vp run check:mobile-patches` (CI) fails the
+    // build if either hunk ever stops applying after a dep bump.
     <NativeTabs
       minimizeBehavior="onScrollDown"
       iconColor={{ default: systemColors.secondaryLabel, selected: systemColors.label }}
--- a/packages/mobile/package.json
+++ b/packages/mobile/package.json
@@ -118,6 +118,13 @@
     "vitest": "^4.1.10"
   },
   "expo": {
+    "autolinking": {
+      "buildFromSource": [
+        "react-native-screens",
+        "@expo/ui",
+        "expo-dev-launcher"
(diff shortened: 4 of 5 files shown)


marcodejongh:
Addressed both minor points in `9e5d726`:

- **Coalescing flag cleared too early** — right, and it was a real hole: the flag was released when the `dispatch_async` hop fired rather than when the layout ran, so a re-entry during a pending coordinator wait could start a second chain. It now clears in the terminal branch, so the whole deferral chain is a single in-flight pass.
- **Unbounded coordinator deferral** — the concern is correct and my comment was wrong: I claimed the chain "cannot re-defer forever", but under continuous navigation a completion can fire straight into the next transition. Rewritten to say what it actually does. I'm keeping the behaviour: there is no safe moment to force tab-bar layout mid-transition, and while transitions are running UIKit is laying the bar out anyway. Worst case is a late-appearing accessory; the alternative is BOARDSESH-9K.

**Not** addressing the blocking point, because it can't be addressed from here — agreed that the Obj-C++ needs a macOS `vp run mobile:ios` before this goes anywhere. That's why it's a draft, and it stays one until someone with Xcode signs it off. The same build should re-check the original repro this patch exists for (accessory attached after the tab bar has appeared still lays in), since the synchronous pass is gone.

marcodejongh:
Ran two independent Fable reviews over this — one on the Objective-C++, one on CI/release. Pushed `a6751ab`.

**The Codex catch was real and is fixed.** `animateAlongsideTransition:completion:` returning `NO` would have left the coalescing flag armed with no pending pass, permanently suppressing every later attach nudge — the exact bug this patch exists to fix, reintroduced by my own previous round. Now only returns early when registration succeeded.

**Then the bigger one: this PR's own iOS CI green was meaningless.** `iOS CI (React Native)` filters on `patches/**` and gates on a native-change check, but builds `-scheme BoardseshTests` — a standalone Swift target that never links the app and never compiles a Pod. I pulled run `30877721241`'s log: **zero `CompileC` lines, no `RNScreens` target**. It compiled none of the Objective-C++ it exists to validate. Every `patches/**` PR has had this hole.

Fixed by building the app scheme before the test target is grafted on, plus the two traps the reviewers found in doing so:

- **Codegen race.** New Architecture "Generate Specs" races the compile on a cold runner, which is every run of this job. `scripts/mobile-build-sim-app.ts` already retries once for exactly this and documents why; mirrored. A real compile error still fails the retry.
- **Expo precompiled modules are ON by default.** They swap a pod's source for a prebuilt xcframework whenever a tarball resolves. RNScreens has none today — so the patch compiles by luck, unasserted. If that ever changes, this gate goes green having compiled nothing *and* `ios-testflight-rn.yml` links the prebuilt too, shipping a binary without the patch. So the step now asserts a patched file appears in the build log, and every bun-patched native pod is pinned via `expo.autolinking.buildFromSource`.

Also reset the coalescing flag in `initState` (redundant while `shouldBeRecycled` is `NO`, but a stale `YES` silently disables the accessory forever), and bumped the marketing version to **2.3.1** — `2.3.0` is already released (`release/ios-v2.3.0-ef3763870f81`), so the next submission needs a new string.

Two things I can't close from here:

1. **Device QA of the accessory hide/unmount path.** Attach-only is the sole behavioural narrowing here. If iOS 26 has the symmetric bug on `setBottomAccessory:nil`, a hide toggle or unmount could leave a stale accessory. Clearing the current climb while on a tab is the flow to check.
2. **This workflow runs the macOS job twice per commit** — `push` (branches-ignore main) and `pull_request` have different concurrency groups. That was ~8 wasted minutes; with a real compile it's closer to 40. Worth a follow-up, but I'm not touching trigger/concurrency semantics on this PR.

marcodejongh:
**The blocking review item is now satisfied — by CI, not by a Mac.** Evidence from run [30900302700](https://github.com/boardsesh/boardsesh/actions/runs/30900302700) on `a6751ab`:

| | before (`9e5d726`) | now (`a6751ab`) |
|---|---|---|
| `CompileC` lines in the log | **0** | **1293** |
| compiles in target `RNScreens` | **0** | **179** |
| patched files compiled | none | `RNSTabsHostComponentView.mm`, `RNSScrollViewFinder.mm` |

- `Build the app (compiles Pods + patched native modules)` — 17m48s, exit 0 under `set -o pipefail`.
- `Assert the patched native module really compiled` — passed: *"RNScreens compiled from source — patch exercised."*
- The codegen-race retry did **not** fire; the build was clean first time.
- **Zero warnings and zero errors** attributed to `RNSTabsHostComponentView.mm`.

Total job ~30 min, comfortably inside the raised 90-minute timeout.

So the Objective-C++ compiles, and every future `patches/**` PR gets the same gate instead of the false green this one started with.

Still open, and still needs a human with the app in hand: **device QA of the accessory hide/unmount path.** Attach-only is the one behavioural narrowing in this patch — clear the current climb while sitting on a tab and confirm the bottom accessory actually goes away rather than leaving a stale platter or a reserved inset. Everything else here is now machine-verified.

marcodejongh:
**A symbolicated TestFlight crash log lands this on our own line numbers.** Incident `E0514EE1-0EBB-4C61-83A8-F23DF54B6AC2`, Boardsesh 2.3.0 (5), iPhone16,1, iOS 27.0 beta (24A5390f), 2026-08-04 19:06:51 +0200.

```
70  Boardsesh  -[RNSTabsHostComponentView rnscreens_relayoutBottomAccessoryIfAttachedAfterAppearance] + 84 (RNSTabsHostComponentView.mm:448)
71  Boardsesh  -[RNSTabsHostComponentView updateContainer] + 588 (RNSTabsHostComponentView.mm:184)
72  Boardsesh  -[RNSTabsHostComponentView mountingTransactionDidMount:withSurfaceTelemetry:] (RNSTabsHostComponentView.mm:372)
```

**`RNSTabsHostComponentView.mm:448` is `[controllerView layoutIfNeeded];`** — the synchronous layout this PR deletes. Line 184 is the `[self applyBottomAccessoryVisibility]` call site inside `updateContainer`, inlined by the optimiser. Confirmed by applying `origin/main`'s patch to a pristine 4.26.2 tree and reading those lines.

The full cycle, reading the stack outward then back in:

1. `-[UIPresentationController runTransitionForCurrentStateAnimated:]` (140) → `_addSubview:` (132) → safe-area change (117–116)
2. → `-[RNSScreenView safeAreaInsetsDidChange]` (115) → `RNSSafeAreaViewState` update (108–106)
3. → RN mounting transaction (105–73) → `updateContainer` (71) → **our relayout (70) → synchronous `layoutIfNeeded` (69)**
4. → `-[UITabBar layoutSubviews]` (56) → `_UITabBarVisualProvider_Floating._updateLensViewForSelectedItem` (51–50) → `-[UIViewFloatAnimatableProperty setValue:]` (49)
5. → that set lands inside an animation, so AnimationKit spins up an in-process animation (44–41) → `+[UIViewAnimationState popAnimationState]` (39–37)
6. → which runs the **sheet's** `_transitionWillBegin` alongside block (31) → `SheetLayoutInfo._layout()` (30–28) → `layoutBelowIfNeeded` (27)
7. → tab-bar lens **again** (13–9) → second animatable-property set → `_assertionFailure` (1–0)

Frame 70 is the pivot. Everything from 69 inward only happens because we forced layout there. Removing it — and, on the attach branch, deferring past any live transition coordinator — cuts the chain at its origin. This is no longer inference from an unsymbolicated Sentry stack; it's our line number in the crash report.

Worth noting for triage: this is build **5**. Current TestFlight is build 12, and neither contains the fix, so testers on iOS 27 keep hitting it until this ships.

Unrelated but surfaced by the same exercise: **Sentry has symbol names for these frames but no file/line**, which is why the raw `.crash` was needed to get here. dSYMs aren't reaching Sentry for TestFlight builds. Separate issue, but it cost us a day on this one.

Issue #3273, opened 2026-06-29
Mobile (@expo/ui): paper-removal endgame — retire react-native-paper + the UI-variant machinery


marcodejongh:
> Part of the **@expo/ui native-UI migration** follow-up (after #3254 + #3261). Maintainer guidance: only migrate where it *legitimately improves the experience*; FlashList/FlatList lists, SVG board-art, charts, and gesture boards stay React Native. Pattern: `docs/expo-ui-components.md`.

> **Drafting-agent verdict:** `yes` — Legitimate and worth doing — deletes a whole UI library (react-native-paper) plus the fragile runtime variant indirection — but it's tech-debt/bundle, NOT a UX win, and is strictly the LAST issue (blocked on every sibling control migration). It also bakes in the aesthetic-variant→platform collapse decision, so flag that for the maintainer. Lowest priority; do last.

---

## Why this improves the experience

Be honest up front: **this is a tech-debt / consistency cleanup, not a user-facing UX win.** No screen looks or behaves differently for a climber after this lands. The payoff is for the codebase:

- Removes an entire UI library (`react-native-paper`, MD3 theming, the `PaperProvider`/`Portal` host, the MDI icon-font shim) once nothing renders Material via Paper anymore — every Material control is native Jetpack Compose.
- Deletes the **runtime UI-variant indirection** (`createVariantComponent` / `selectByVariant` / `useVariantValue`). After every control is platform-split (`.ios.tsx` = SwiftUI/glass, `.android.tsx` = Compose/material), the variant router is dead weight: Metro's platform-extension resolution already picks the right tree, and a native Compose Material impl **physically can't render on iOS** (`@expo/ui/jetpack-compose` crashes at module load there), so the old runtime `liquidGlass ↔ material` flip on one platform is already impossible for migrated components.
- Removes a real footgun: `createVariantComponent`'s load-bearing "render as JSX, not `Impl(props)`" hook-count fragility, and the four-axis variant decision tree in `src/theme/variants/README.md`.

It is **strictly the last issue in this set** — it cannot start until every sibling control migration has landed (see Dependencies). Treat it as the closeout.

This also bakes in one genuine architectural decision, called out honestly in **Target** below: collapsing the aesthetic-variant axis into the platform axis (you lose the ability to render Material-on-iOS / glass-on-Android). That capability is already vestigial — the user-facing toggle was removed in PR #3254 — so this is formalizing reality, not a regression. Flag it for the maintainer regardless.

[...]

PR #4203, opened 2026-08-04
fix(mobile): Android segmented control still broken — replace @expo/ui Compose with Paper

fallaciousreasoning:
## Summary

Follow-up to #4196 (merged), which fixed the wrong thing.

That PR's `backgroundColor: 'transparent'` fix (matching the `FilterChipRow.android.tsx` RNGH hit-test bug class) was confirmed via a live OTA preview build to **not** actually fix the issue — Warm-up and Climb Bias were still unresponsive to taps on a real Android device (thanks for testing and flagging this).

## Root cause (this time verified against evidence, not just a plausible-looking precedent)

`Button.android.tsx` uses the exact same `@expo/ui` `<Host>`-wrapped Jetpack Compose pattern for a plain button, and that works fine — which rules out a general Host/RNGH-touch-dispatch problem (the theory behind the #4196 fix). The failure is specific to `@expo/ui`'s `SingleChoiceSegmentedButtonRow` / `SegmentedButton` pairing. I traced through the vendored Kotlin source in `node_modules/@expo/ui` looking for a concrete difference between the working `Button` path and the broken `SegmentedButton` path, but couldn't get a definitive runtime answer without native debugging tools I don't have access to here (Android Studio breakpoints, etc.).

## Fix

Rather than keep guessing at native Compose internals, this replaces the Android implementation with `react-native-paper`'s `SegmentedButtons` — the same component:
- the web fallback (`SegmentedControl.web.tsx`) already uses successfully, and
- this control itself used **before** the `@expo/ui` migration (`6af5f43df`).

`react-native-paper` remains a live, working dependency elsewhere on native Android today (`Button.android.tsx`, `AppMenu.android.tsx`, `AuthTextInput.android.tsx`, etc.) pending the paper-removal endgame tracked in #3273, so this isn't introducing a new pattern to the codebase — it's reverting one specific control to a known-working one while that endgame plays out.

## Release Notes

- Fixed the Warm-up and Climb Bias pickers not responding to taps on Android

## Test plan

- `vp check` — passes on the changed file
- `vp run check:mobile-bundle` — passes (Android bundle exports cleanly with `@expo/ui`'s Compose SegmentedButton imports fully removed from this file)
- `vp test run --project mobile --reporter=agent` — full suite passes aside from 2 pre-existing unrelated failures in `logbook-tab-grouped.test.tsx` (confirmed present on `main` without this change)

**Still not verified on a real device or working emulator by me** — my headless Android emulator testing turned out to be unreliable for `@expo/ui` Compose components specifically (every such control failed to receive synthetic taps there, including an already-shipped `Switch`, which is why the emulator couldn't be trusted to validate #4196 either). This fix sidesteps `@expo/ui`'s Compose bridge entirely for this control, so that specific emulator unreliability shouldn't apply — but please confirm on the OTA preview before merging, the same way the #4196 regression was caught.

### DIFF
--- a/packages/mobile/src/components/SegmentedControl.android.tsx
+++ b/packages/mobile/src/components/SegmentedControl.android.tsx
@@ -25,7 +25,7 @@
 // semantics of its own, so it's carried by a wrapping `radiogroup` View (mirrors the
 // pre-@expo/ui Material implementation this restores).
 
-import { View } from 'react-native';
+import { StyleSheet, View } from 'react-native';
 import { SegmentedButtons } from 'react-native-paper';
 import { readableTextColor } from './grade/grade-chip-colors';
 import { makeSelectHandler } from './SegmentedControl.logic';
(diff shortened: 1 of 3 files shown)


fallaciousreasoning:
Addressed the two stale-comment nits (pushed).

On the broader finding — `FeatureFlagsForm.android.tsx` and `MoreForm.android.tsx` do still import `SingleChoiceSegmentedButtonRow`/`SegmentedButton` from `@expo/ui/jetpack-compose` directly (not through this shared `SegmentedControl` component), and I confirmed `FeatureFlagsForm.android.tsx`'s usage is real — `FEATURE_FLAG_CHOICES` rows are actually rendered on the tester-only Feature Flags screen, so it's presumably hitting the same bug. `MoreForm.android.tsx`'s `case 'segmented':` branch, though, is currently dead code — I grepped for any row construction site passing `kind: 'segmented'` and found none, so nothing in the shipped app renders that branch today.

Both are embedded inside larger Compose `LazyColumn`/`Card`/`Column` trees (the PR-2 "one Host per screen" consolidation), not simple standalone controls like this one — migrating them means figuring out whether a plain RN component (Paper) can safely nest inside that Compose tree, which I haven't verified. Given this PR fixes the actively user-facing, confirmed-broken bug (Record tab) and I don't want to hold it up on an unverified architectural change to two other files, I'm treating those as a follow-up rather than expanding scope here. Will file/fix them next.

marcodejongh (dismissed):
FYI Im having to do a native patch, so the OTA pipeline is currentyl blocked (again).
You could backport it though to 2.3.0. OR just sideload the apk

fallaciousreasoning:
Manually verified on a GUI Android emulator (not the headless one that gave false signal on #4196) — Warm-up and Climb Bias both respond to taps correctly now. Marking ready for review.

PR #4208, opened 2026-08-06
chore(shared): cull unread PostHog events

marcodejongh:
## Summary

Our PostHog event projection has been climbing, so I audited what we actually emit against what anything reads.

**Method:** 30 days of volume (1.32M events), cross-referenced against every saved insight in project 412845 — event names extracted server-side with `extractAll` over `system.insights.query`, so no truncation blind spots — plus `system.actions`, `system.dashboards`, and `system.experiments`.

Two findings drove the change:

- **The project runs no experiments at all.** `$feature_flag_called` (172,684/30d, **13% of all volume**) therefore has no exposure consumer. The only insights referencing it are PostHog's auto-generated "`<flag>` Usage" boilerplate.
- **`Live Activity Push Delivery` was the worst ratio in the project**: 26,609 events across 191 users (83.5/user), of which **34** carried a failure. A third were `heartbeat` keep-alives.

Estimated reduction **~434k events / 30 days (~33%)**, with no consumed metric lost except one insight rewiring (below).

### What changed

| Change | 30d volume | Consumers before |
|---|---|---|
| `Live Activity Push Delivery` → fires only when `failedCount > 0 \|\| staleCount > 0` | 26,575 removed | none |
| `$feature_flag_called` → suppressed via `sendEvent: false` (both platforms) | 172,684 | 20 auto-generated flag dashboards |
| `Wall Advance` → folded into `Queue Navigation` | 2,213 | none |
| `Tick Button Clicked` + `Quick Tick Saved` → collapsed into `Tick Logged` | 36,263 | 1 + 5 insights |
| `Play Drawer Opened`, `Search Result Selected`, `Board Lightbulb Connect`, `Grade Filter Changed` → removed | 172,306 | none |

**Flag suppression** is verified against `@posthog/core` 1.46.1 (shared by `posthog-react-native` 4.61.2 and `posthog-js-lite` 4.10.1, both declaring `^1.46.1`): `_getFeatureFlagResult` gates the `capture('$feature_flag_called')` on `sendEvent`, and both `getFeatureFlag` and `isFeatureEnabled` forward it. **Flag values are unaffected** — only the exposure event stops. Drop the option at a call site if a flag ever needs real exposure analysis.

**The tick funnel** went from four events per tick to two. Worth calling out: web's `Tick Button Clicked` fired on the sheet **opening**, while mobile's fired on **save-intent** — one event name covering two different moments across platforms. Both platforms now share one funnel: `Quick Tick Opened` → `Tick Logged` / `Quick Tick Failed`.

### Deliberately left alone

- `Climb Sent to Board Success` (118,048) — 16 insights read it **and** it feeds `packages/db/scripts/refresh-recommendations.ts` → `board_climb_send_stats`.
- `$screen` (287,581) — highest single event, but 8 insights depend on it and it's already deduped in `AnalyticsScreenTracker`.
- `Bluetooth Connection Success` / `Failed` — the BLE health dashboard (1874713).
- `OTA Update Status` / `Downloaded` — no insight consumers, but only 1.4–1.8 events/user and the only production OTA rollout-health signal. Not worth blinding that.
- `Quick Tick Opened` — documented watchdog invariant for the phantom-reopen bug.

### Already fixed, draining on their own — no code change needed

`Board Now Playing Received` (39k), `Board Climb Reported` (22k), `Board History Catch Up` (24.7k, already gated on `recoveredThroughSeqDelta > 0`), and the `Application Backgrounded/Became Active/Opened/Installed/Updated` family (~35k) emit **zero** from the current build. They only come from the older shipped binary and drain as the fleet updates.

### Dead code removed with them

`describeGradeFilter` (+ its tests), `buildPlayDrawerBoardLayout`, the `openPlayDrawer` `source` option, the analytics-only `source`/`boardLayout`/`climbUuid` options on `useLightbulbControl`, and the per-row search rank index in the climbs list.

`GradeTapMeta` / `extendedRangeWithinWindow` are **kept**: they're now unused by production code, but they're an optional second callback arg and the grade-picker rule-3 tests assert them as tap semantics, not analytics.

### PostHog insights — done

All saved insights referencing a removed event have been repointed. It was **7**, not the 5 the plan estimated:

| Insight | Dashboard | Change |
|---|---|---|
| `[Engagement] Core actions per day` | Mobile App Rollout | tick series → `Tick Logged` |
| `[North-Star] Weekly Active Climbers (WAC) — guardrail` | App Success | → `Tick Logged` |
| `[North-Star] Hardware reliance — WAB / WAC` | App Success | → `Tick Logged` |
| `[Retention] Do tick-loggers come back?` | App Success | cohort entity → `Tick Logged` |
| `[Engagement] Send → tick logging rate` | App Success | → `Tick Logged` |
| `[Engagement] Tick richness by status` | App Success | → `Tick Logged` |
| `[Boards] Climb engagement funnel` | Board Sessions & Hardware | open step → `Quick Tick Opened` |

Verified: zero saved insights now reference `Quick Tick Saved`, `Tick Button Clicked`, `Wall Advance`, `Play Drawer Opened`, `Search Result Selected`, `Board Lightbulb Connect`, or `Grade Filter Changed`.

**Known step in the data:** `Tick Logged` runs ~7.5% below `Quick Tick Saved` historically (16,630 vs 17,976 on RN) because older shipped binaries fire the old event but not the canonical one — the tick tiles will show a modest step down until the fleet updates. `status` has full history on `Tick Logged`; `hasQuality` / `hasComment` only start with this release (noted in the affected insight's description).

The auto-generated flag-usage dashboards will also go quiet — expected, per the flag-suppression trade-off above.

### Note on when this lands

94% of current volume comes from binaries reporting `posthog-react-native` **4.46.8**, while the current build reports **4.61.2**. The backend change (`Live Activity Push Delivery`) bites immediately on deploy; the client-side savings arrive as the fleet picks up this JS.

> ✅ **Fable BLE review: APPROVE** — [full review in the comments](https://github.com/boardsesh/boardsesh/pull/4208#issuecomment-5222255466). No behaviour-change findings: `packages/mobile/src/lib/ble/` and `packages/shared/ble-protocol/` have an empty diff, the `onPress` dependency narrowing is sound, and the connect call order and arguments are unchanged. It confirmed `Bluetooth Connection Success/Failed` genuinely cover these paths, and flagged two accepted observability trade-offs (silent-hang detection, and losing the entry-point split on the notification reconnect). Its nits are applied in `11baa658`.

## Release Notes

## Test plan

- `vp run typecheck` — green across all packages
- `vp run test:mobile` — 582 files, 5061 tests passing
- `vp run test:web` — 473 files, 5854 tests passing
- `vp test run --project backend` — the new `live-activity-analytics.test.ts` passes; the 86 local failures are environment contention (shared worker DBs on `:5433`, ports 8081–8089 held by other worktrees' dev servers) and every affected file passes when run in isolation
- shared packages — 15 files, 206 tests passing
- `vp check` — 28 errors / 299 warnings vs. 30 / 299 on clean `main`; two fewer errors from deleted code, zero new findings
- Added `packages/backend/src/__tests__/live-activity-analytics.test.ts` covering the new failure/stale gate
- Added `queue-control-bar-swipe-advance.test.tsx` asserting both swipe directions emit exactly one `Queue Navigation` with the right `sessionMode` and no `Wall Advance` (review follow-up)
- Fable BLE review ran `typecheck:mobile`, `use-lightbulb-control` (13 tests) and `live-activity-bridge` (17 tests) independently — all green

### DIFF
--- a/packages/backend/src/__tests__/live-activity-analytics.test.ts
+++ b/packages/backend/src/__tests__/live-activity-analytics.test.ts
@@ -1,0 +1,69 @@
+import { beforeEach, describe, expect, it, vi } from 'vite-plus/test';
+
+const posthogMocks = vi.hoisted(() => ({
+  captureBackendEvent: vi.fn(),
+}));
+
+vi.mock('../services/analytics/posthog', () => ({
+  captureBackendEvent: posthogMocks.captureBackendEvent,
+}));
+
+// Matches apns-analytics.test.ts: reset + dynamic import so the hoisted mock is
+// the module the subject binds to. A static top-level import binds the real one.
+async function loadLiveActivityModule(): Promise<typeof import('../services/analytics/live-activity')> {
+  vi.resetModules();
+  return import('../services/analytics/live-activity');
+}
+
+function pushDelivery(overrides: { failedCount?: number; staleCount?: number } = {}) {
+  return {
+    userId: 'user-1',
+    sessionId: 'session-1',
+    event: 'update' as const,
+    source: 'heartbeat' as const,
+    tokenCount: 3,
+    sentCount: 3,
+    failedCount: 0,
+    staleCount: 0,
+    elapsedMs: 42,
+    ...overrides,
+  };
+}
+
+describe('trackLiveActivityPushDelivery', () => {
+  beforeEach(() => {
+    vi.clearAllMocks();
+  });
+
+  it('does not capture a delivery where nothing failed or went stale', async () => {
+    const { trackLiveActivityPushDelivery } = await loadLiveActivityModule();
+    trackLiveActivityPushDelivery(pushDelivery());
+
+    expect(posthogMocks.captureBackendEvent).not.toHaveBeenCalled();
+  });
+
+  it('captures a delivery with failures', async () => {
+    const { trackLiveActivityPushDelivery } = await loadLiveActivityModule();
+    trackLiveActivityPushDelivery(pushDelivery({ failedCount: 2 }));
+
+    expect(posthogMocks.captureBackendEvent).toHaveBeenCalledWith(
+      'Live Activity Push Delivery',
+      expect.objectContaining({
+        distinctId: 'user-1',
+        properties: expect.objectContaining({ failedCount: 2, staleCount: 0 }),
+      }),
+    );
+  });
+
+  it('captures a delivery with stale tokens', async () => {
+    const { trackLiveActivityPushDelivery } = await loadLiveActivityModule();
+    trackLiveActivityPushDelivery(pushDelivery({ staleCount: 1 }));
+
+    expect(posthogMocks.captureBackendEvent).toHaveBeenCalledWith(
+      'Live Activity Push Delivery',
+      expect.objectContaining({
+        properties: expect.objectContaining({ failedCount: 0, staleCount: 1 }),
+      }),
+    );
+  });
+});
--- a/packages/backend/src/services/analytics/live-activity.ts
+++ b/packages/backend/src/services/analytics/live-activity.ts
@@ -155,7 +155,15 @@ export function trackLiveActivityWidgetNavigationAttributionGap(
   });
 }
 
+// Only deliveries that actually went wrong are captured. Every push used to fire
+// one of these — 26.6k events across 191 users in a 30-day window, of which 34
+// carried a failure, and a third were `source: 'heartbeat'` keep-alives. Nothing
+// read the clean rows: no insight, dashboard, or alert referenced this event.
+// The failure/stale rate is the question this event exists to answer, so the
+// denominator lives in the APNs logs rather than in PostHog's event budget.
 export function trackLiveActivityPushDelivery(event: LiveActivityPushDeliveryEvent): void {
+  if (event.failedCount === 0 && event.staleCount === 0) return;
+
   captureBackendEvent('Live Activity Push Delivery', {
     distinctId: event.userId,
     properties: {
--- a/packages/mobile/app/(tabs)/climbs/__tests__/index-keyboard.test.tsx
+++ b/packages/mobile/app/(tabs)/climbs/__tests__/index-keyboard.test.tsx
@@ -367,27 +367,17 @@ describe('ClimbList keyboard handling', () => {
   });
 });
 
-describe('ClimbList search result selection tracking', () => {
-  it('tracks the rank of the first result when it is pressed', async () => {
-    const { findByText } = render(<ClimbList />);
-
-    fireEvent.click(await findByText('Moonage'));
-
-    expect(mocks.track).toHaveBeenCalledWith(
-      'Search Result Selected',
-      expect.objectContaining({ rank: 0, loadedResultCount: 2, boardName: 'kilter', angle: 40 }),
-    );
-  });
-
-  it('tracks the rank of a later result when it is pressed', async () => {
+describe('ClimbList search result selection', () => {
+  // Pressing a row activates the climb but emits NO per-press analytics event.
+  // `Search Result Selected` fired here on every tap (53.8k events / 30 days)
+  // and no insight ever read it; `Climb Search Performed` still covers search.
+  it('activates a pressed result without firing a per-result event', async () => {
     const { findByText } = render(<ClimbList />);
 
     fireEvent.click(await findByText('Zenith'));
 
-    expect(mocks.track).toHaveBeenCalledWith(
-      'Search Result Selected',
-      expect.objectContaining({ rank: 1, loadedResultCount: 2, boardName: 'kilter', angle: 40 }),
-    );
+    expect(mocks.activateClimb).toHaveBeenCalled();
+    expect(mocks.track).not.toHaveBeenCalledWith('Search Result Selected', expect.anything());
   });
 });
 
--- a/packages/mobile/app/(tabs)/climbs/index.tsx
+++ b/packages/mobile/app/(tabs)/climbs/index.tsx
@@ -12,7 +12,6 @@ import {
   mergeBoardFilters,
   countActiveFilters,
   hasActiveBoardFilters,
-  describeGradeFilter,
   flagsToProgress,
   progressToFlags,
   applyStatusChange,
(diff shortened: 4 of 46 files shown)


marcodejongh:
## 🔵 Fable BLE review — **APPROVE**

Per the CLAUDE.md rule that Bluetooth changes require a Fable review. Reviewed at `22cc4c4`.

### Behaviour-change findings: none

- `packages/mobile/src/lib/ble/` and `packages/shared/ble-protocol/` have an **empty diff** against the merge base — connection lifecycle, retries, and protocol are untouched. The only web BLE-adjacent files touched are analytics catalogs, not the Web Bluetooth control.
- **Dependency narrowing is correct.** `use-lightbulb-control.ts` `onPress` now reads only `bluetooth` and its properties (`isConnected`, `loading`, `reconnectSerialForCurrentBoard`, `reconnectDeviceIdForCurrentBoard`) — all reached through the single `bluetooth` dep. The removed deps (`source`, `sessionId`, `boardLayout`, `climbUuid`) were read *only* by the deleted `track()` call. No new stale-closure surface.
- **Call order preserved.** The old `track()` sat between the `pressAction` branch and `armUndoWallChangeToast()`; deleting it leaves `armUndoWallChangeToast()` → `connect(undefined, undefined, serial ?? undefined, deviceId ?? undefined)` in the same order with identical arguments. The `disconnect` and `noop` press branches are byte-identical.
- **Live-activity bridge intact.** `bluetoothRef` is still updated every render with a once-subscribed empty-dep listener; `reconnect` still does toast-arm → `connect(…, serial, deviceId)`; `reassert` still calls `reassertWall()` with no toast. The removed `sessionIdRef`/`climbUuidRef` fed only the deleted `track()`; `sessionId` and `displayClimb` are still live for `useLiveActivity` and the Android thumbnail render.
- The backend `trackLiveActivityPushDelivery` gate changes only what gets captured — no push/APNs behaviour is affected.

### Observability assessment — two genuine losses, both judged acceptable

1. **Silent hangs become invisible.** Verified that `connect()` in `use-board-bluetooth.ts` fires `Bluetooth Connection Success`, `Bluetooth Connection Failed` unconditionally in the catch (including `user_cancelled`, where only the Alert is suppressed), a `dropped_after_connect` failure, and settles dangling picker promises on scan timeout. So for any connect attempt that **settles**, outcome coverage is complete and the PR's claim holds. The attempt-minus-outcome delta was the only field signal for a connect promise that *never* settles — a class largely closed by the existing picker/timeout hardening.
2. **The notification/lock-screen reconnect path is no longer distinguishable.** `Bluetooth Connection Success/Failed` carry board identity, apiLevel, and write-type diagnostics, but no `source`. An Android foreground-notification reconnect regression would now blend into the aggregate connect failure rate rather than showing as a segmented spike.

   → **Recommended follow-up (not done here):** if per-entry-point BLE health matters, add a `source` prop to the *outcome* events rather than resurrecting the attempt event. That costs zero extra events. Deliberately left out of this PR because threading `source` through `bluetooth.connect()` means touching the BLE call path this telemetry-only change is careful not to alter.
3. **Backend gate trade-off:** the failure/stale *rate* is no longer computable from PostHog alone — only the numerator lands; the denominator lives in APNs logs. At 34 failures per 26.6k events the clean rows were near-pure noise, so this is acceptable — but a "failures went up" insight can no longer be normalised in-product.

### Tests

```
vp run typecheck:mobile   → tsc --noEmit  Done in 5.53 s
use-lightbulb-control     → Test Files 1 passed (1), Tests 13 passed (13)
live-activity-bridge      → Test Files 1 passed (1), Tests 17 passed (17)
```

### Nits — applied in `11baa658`

- `useLightbulbControl`'s options are now all optional, so the parameter is defaulted and the `useLightbulbControl({})` call sites read `useLightbulbControl()`.
- `BoardControlIndicator`'s comment no longer credits this hook with analytics — the connect outcome events are emitted inside `bluetooth.connect()`.
- Noted incidental upside: dropping `climbUuid` from the `onPress` deps means the PlayDrawer bulb's callback no longer churns identity on every climb navigation — a small render-stability win.

marcodejongh:
Responses to the review above, in order.

### 1. Fable review — done, not just requested

It's [the comment directly above](https://github.com/boardsesh/boardsesh/pull/4208#issuecomment-5222255466), posted ~3 minutes before this review ran. Verdict **APPROVE**, no behaviour-change findings, and it independently ran `typecheck:mobile` + the `use-lightbulb-control` (13) and `live-activity-bridge` (17) suites. Its nits are applied in `11baa658`.

### 2. `staleCount`-alone path — already covered

This one is a misread. The helper defaults `failedCount: 0` and `staleCount: 0`, and the third case spreads `{ staleCount: 1 }` over those, so it *is* the stale-only branch — the assertion pins it explicitly:

```ts
trackLiveActivityPushDelivery(pushDelivery({ staleCount: 1 }));

expect(posthogMocks.captureBackendEvent).toHaveBeenCalledWith(
  'Live Activity Push Delivery',
  expect.objectContaining({
    properties: expect.objectContaining({ failedCount: 0, staleCount: 1 }),
  }),
);
```

`{ failedCount: 2, staleCount: 1 }` never appears in the file. All three branches of the gate — neither, failed-only, stale-only — have a case. No change made.

### 3. `logascent-form.tsx` partial payload — real, fixed in `027a092f`

Good catch, and it's more than pre-existing: this PR is what made it matter. `Tick Logged` became the *single* commit event for all three tick surfaces here, so a `surface` breakdown is only comparable if every surface reports the same fields — and the full form was sending `status`/`difficulty` only.

It now also sends `attemptCount`, `hasQuality`, `grade` and `hasComment`, matching `use-tick-save.ts` and mobile's `QuickTickBar.tsx`. The form already had the `grades` list in scope to resolve the label, so no new plumbing. Extended the existing `logascent-form-0deg` TickLogged test to assert the four fields are present, so the surfaces can't silently drift apart again.

---

Re-validated after all three commits: `vp run typecheck` green; web 474 files / 5856 tests, mobile 582 files / 5061 tests; `vp check` 28 errors / 299 warnings vs **30 / 299** on clean `main` (two fewer from deleted code, zero new).

PR #4210, opened 2026-08-07
fix(mobile): keep the play drawer's background from covering the player on Android

### LINKED ISSUE #4209: [Android 9] Play drawer content hidden behind elevated background
# [Android 9] Play drawer content hidden behind elevated background

## Summary

On a Samsung Galaxy S8 running Android 9, opening a Tension climb displays only the play drawer’s gray/purple background. The board, title, and controls are not visible.

Reproduced with Android dev client build 520 on the current `main` branch.

## Steps to reproduce

1. Configure a Tension Board.
2. Open the climb search.
3. Select a climb.

## Expected result

The play drawer displays the selected climb, board image, title, and controls.

## Actual result

The play drawer opens, but only its background is visible.

## Cause

The full-screen `GlassSurface` in `packages/mobile/app/play.tsx` uses the Material fallback with its default shadow, which includes Android elevation. On Android 9, this absolute-fill background is drawn above the player content.

Runtime diagnostics confirmed that the player, selected climb, board configuration, and board-render data were all mounted correctly underneath the background.

## Fix

Make the background non-elevated and non-interactive:

```tsx
<GlassSurface
  style={StyleSheet.absoluteFill}
  glassEffectStyle="regular"
  role="low"
  level="level0"
  pointerEvents="none"
  fallbackColor={systemColors.secondaryBackground}
  tintColor={playDrawerMaterialTint[colorScheme]}
/>
```

The fix was successfully verified on the affected Galaxy S8. The board, title, and controls are visible again.

`vp check packages/mobile/app/play.tsx` passes without formatting, lint, or type errors.

marcodejongh:
## Summary

On a Galaxy S8 running Android 9, opening a Tension climb showed nothing but the play drawer's
gray/purple background — no board, no title, no controls. The reporter's runtime diagnostics confirmed
the player, selected climb, board config, and board-render data were all mounted correctly *underneath*
it.

Root cause is a `GlassSurface` footgun. Its Material branch defaults to `shadows.sm`:

```ts
// packages/mobile/src/components/GlassSurface.tsx:135
const elevationStyle = level ? materialElevation[level] : shadows.sm;
```

`shadows.sm` carries `elevation: 2`, and Android resolves paint order between siblings by Z. In
`app/play.tsx` the full-screen background is a **sibling** of `PlayDrawer`, so the elevation-2 fill can
draw on top of the elevation-0 player and cover it. Android always resolves to the Material variant
(`resolveUiVariant` → Material off iOS), so every Android device takes this branch — though in practice
only older ones appear to order the siblings this way (see the screenshots note below). The surface also
omitted `pointerEvents`, so it defaulted to `'auto'` and could swallow touches on top of hiding content.

The fix opts the background out of both: `level="level0"` (flat — `elevation: 0`, `shadowOpacity: 0`)
and `pointerEvents="none"`. Both props already existed on `GlassSurface` and are already forwarded on
every branch, so no component change was needed. The dropped shadow is invisible by design — a
full-screen fill casts off-screen.

Two other screen-level fills stack their content as siblings in exactly the same shape, so they get the
same hardening (not device-reproduced, so treat these as defensive):

- `app/(tabs)/record/index.tsx` — the Record tab's background, above `<SessionScreen />`
- `src/components/navigation/IpadSidebar.tsx` — the rail fill behind the nav items

Deliberately left alone: the bounded-size `absoluteFill` fills inside buttons and toolbars
(`GlassIconButton`, `GlassActionToolbar`, `SearchHeader`, `AccessoryBarSurface`, `SessionStartFab`,
`GradeRail`, the playlist buttons). They share the missing `level`, but they render fine on Android
today, so touching them would be an unverified visual sweep rather than a fix. Also not changing
`GlassSurface`'s `shadows.sm` default — that would silently flatten ~15 unmigrated surfaces.

Closes #4209

## Release Notes

- Fixed the play drawer on Android opening to an empty purple screen — your climb, the board, and the
  controls are back.

## Test plan

- `vp run test:mobile` — 585 files, 5088 tests, all passing
- `vp check` on the five changed files — no formatting, lint, or type errors
- New regression coverage in `src/components/__tests__/glass-surface.test.tsx` pins the contract: the
  Material branch casts `elevation: 2` by default (the footgun), `level="level0"` renders flat, and
  `pointerEvents` reaches the Material surface. The suite's `shadows.sm` stub was `{}`, which erased
  the very elevation that caused this bug — it now carries the real `elevation: 2`.
- **Device verification:** the reporter confirmed the `play.tsx` fix on the affected Galaxy S8
  (Android 9) — board, title, and controls visible again. That remains the only reproduction of the
  bug itself.
- **Screenshots (no-regression only):** the Android screenshot workflow's `main` baseline renders the
  play drawer correctly on its modern-Android emulator, so CI does **not** reproduce the Android 9
  symptom. The run on this branch is therefore a check that the play drawer and Record tab still render
  and look unchanged, not proof of the fix — no screenshots are attached to this PR because none would
  demonstrate anything the device verification above doesn't already cover.
- **Regression guard:** `background-glass-surface-flat-props.test.ts` source-scans the three fix sites
  (`app/play.tsx`, `app/(tabs)/record/index.tsx`, `IpadSidebar.tsx`) and fails if a future edit drops
  `level="level0"` or `pointerEvents="none"` from any of them.

JS-only — no native fingerprint move, so this rides OTA.

### DIFF
--- a/packages/mobile/app/(tabs)/record/index.tsx
+++ b/packages/mobile/app/(tabs)/record/index.tsx
@@ -8,11 +8,16 @@ import { SessionScreen } from '../../../src/components/session-screen/SessionScr
  * running. A GlassSurface fills the background so the Liquid-Glass language from
  * the rest of the chrome carries through. No overlay props: there's no
  * drag/pull-to-dismiss here — switching tabs is the minimize.
+ *
+ * The fill is flat (`level0`) and non-interactive: it's a background with
+ * SessionScreen stacked on top as a sibling, and Android orders siblings by Z, so
+ * the Material branch's default `shadows.sm` cast (elevation 2) would lift it over
+ * the session content (the shape that broke the play drawer in #4209).
  */
 export default function RecordIndex() {
   return (
     <View style={styles.root}>
-      <GlassSurface glassEffectStyle="regular" style={StyleSheet.absoluteFill} />
+      <GlassSurface glassEffectStyle="regular" style={StyleSheet.absoluteFill} level="level0" pointerEvents="none" />
       <SessionScreen />
     </View>
   );
--- a/packages/mobile/app/play.tsx
+++ b/packages/mobile/app/play.tsx
@@ -125,11 +125,21 @@ export default function PlayScreen() {
       {/* Edge-to-edge glass/material background with NO radius — full-screen, not
           a card. Rendered on the first frame so the present animates over it.
           GlassSurface resolves Liquid Glass / blur / Material / Reduce-
-          Transparency-solid per device. */}
+          Transparency-solid per device.
+
+          `level0` and `pointerEvents="none"` are load-bearing, not cosmetic. This
+          surface is a BACKGROUND with the player stacked on top as a sibling, and
+          Android orders siblings by Z: GlassSurface's Material branch otherwise
+          defaults to `shadows.sm` (elevation 2), which lifts the full-screen fill
+          above the elevation-0 PlayDrawer and paints over the whole player — the
+          drawer opened showing nothing but its own tint (#4209). A background also
+          has no business taking touches. */}
       <GlassSurface
         style={StyleSheet.absoluteFill}
         glassEffectStyle="regular"
         role="low"
+        level="level0"
+        pointerEvents="none"
         fallbackColor={systemColors.secondaryBackground}
         tintColor={playDrawerMaterialTint[colorScheme]}
       />
--- a/packages/mobile/src/__tests__/background-glass-surface-flat-props.test.ts
+++ b/packages/mobile/src/__tests__/background-glass-surface-flat-props.test.ts
@@ -1,0 +1,55 @@
+import { describe, it, expect } from 'vitest';
+import { readFileSync } from 'node:fs';
+import { join } from 'node:path';
+
+// GlassSurface's Material branch defaults to `shadows.sm` (elevation 2). A
+// consumer that stacks its content as a SIBLING of the surface — rather than as
+// a child — relies on Android's Z-ordering-by-elevation to keep that default
+// from lifting the surface above its own content. `level="level0"` (flat) and
+// `pointerEvents="none"` are what opt a background fill out of that footgun; the
+// glass-surface.test.tsx suite pins GlassSurface's own contract for those props,
+// but nothing else asserted that these specific call sites actually pass them —
+// so a refactor could silently drop either prop and reintroduce #4209 with no
+// test failing. This is a source-scan regression guard, not a render test: a
+// full mount of any of these screens needs their surrounding provider tree
+// (drawer-host / queue / theme), which is disproportionate for pinning two JSX
+// attributes.
+//
+// Each fixture file is asserted to have exactly one `<GlassSurface` JSX usage
+// (a background fill) so the extracted opening tag is unambiguous.
+const BACKGROUND_GLASS_SURFACE_SITES = [
+  join(__dirname, '../../app/play.tsx'),
+  join(__dirname, '../../app/(tabs)/record/index.tsx'),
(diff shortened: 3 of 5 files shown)


marcodejongh:
## Fable review follow-up

Fable's review confirmed the fix as correct and well-scoped ("correct, well-scoped one-cause fix," "the right minimal remedy"), no critical or major findings — just two minor polish items.

**1. PR body cited emulator screenshots in a comment that doesn't exist.**
Reworded the test plan's screenshots note: it now states plainly that no screenshots are attached (since a modern-Android emulator can't reproduce the Android 9 symptom, they'd add no evidence beyond the reporter's device verification already in the test plan), instead of pointing at a comment that was never posted.

**2. No regression test pinned the load-bearing props at the fix site.**
Added `packages/mobile/src/__tests__/background-glass-surface-flat-props.test.ts`: a lightweight source-scan (not a full provider-tree mount, which `app/play.tsx` needs the drawer-host/queue/theme stack for) that asserts each of the three fix sites — `app/play.tsx`, `app/(tabs)/record/index.tsx`, `IpadSidebar.tsx` — still passes `level="level0"` and `pointerEvents="none"` on its background `GlassSurface`. Mutation-tested locally: stripping either prop from any site fails the corresponding assertion, so a future edit that drops one reintroduces #4209 with a red test instead of silently.

Also rebased onto `main` (21 commits behind, no conflicts) and refreshed the stale test-plan file/test counts in the PR body to match.

Validation on the rebased branch: `vp run typecheck:mobile` clean, `vp run test:mobile` 585 files / 5088 tests passing, `vp run check:mobile-bundle` OK, `vp check` clean on all changed files.

marcodejongh:
## Android screenshots — no-regression check

I couldn't boot the emulator locally (the Android emulator segfaults inside `xvfb-run` on this host —
`swiftshader_indirect`, `swiftshader`, and `guest` all crash identically, with and without the Bash
sandbox), so I drove this through the `Mobile Screenshots (Android)` workflow instead. Its `app-store`
flow happens to capture both screens this PR touches: `02-board-view` **is** the play drawer (it waits
on `play-drawer-board-overlay`) and `04-workout-generator` is the Record tab.

To keep the comparison honest I dispatched the workflow twice today — once on this branch, once on
`main` — rather than trusting the nightly baseline:

| Run | Ref | Commit |
| --- | --- | --- |
| [31222906375](https://github.com/boardsesh/boardsesh/actions/runs/31222906375) | this branch | `8b80859` |
| [31224763907](https://github.com/boardsesh/boardsesh/actions/runs/31224763907) | `main` | `22f92b1` |

**Result — this branch vs `main`, same day, same emulator:**

| Screen | Difference |
| --- | --- |
| `02-board-view` (play drawer) | identical except the status-bar battery glyph |
| `04-workout-generator` (Record tab) | **0** differing pixels |
| `00-home` (untouched by this PR) | differs run-to-run |

The play drawer renders board, title, grade, control row, and Logbook row exactly as `main` does, and
the Record tab is pixel-identical. `00-home` differing between two runs of the *same* code is the
useful control here: it confirms the capture isn't deterministic (the workflow header says as much),
so small deltas on other screens aren't attributable to a change.

### What this does and doesn't prove

It does **not** reproduce the bug. `main`'s play drawer renders correctly on the CI emulator's modern
Android — the elevated background only wins the sibling Z-order on older versions, which is why this
surfaced on an Android 9 Galaxy S8 and not here. So these screenshots are a no-regression check:
removing the elevation cast and taking the surface out of hit-testing changed nothing visually on a
platform where the bug never fired.

The fix itself is verified by the reporter on the affected Galaxy S8 (Android 9) — board, title, and
controls visible again — plus the unit coverage added in `glass-surface.test.tsx`.

PR #4215, opened 2026-08-07
fix(mobile): reorder the tick sheet by how often fields get used

### LINKED ISSUE #4163: [UI/UX] Climb: Reorder tick detail window
1. The date and time fields are for related info, and are both quite short, so could be compressed into a single row
    - I suggest separating with just an @ symbol (e.g. `Date  d Mon YYYY  @  HH:MM`), though the word "Time" may still be ok
2. The current row ordering seems ill-suited to how often each field likely gets interacted with
    - Given thumbs come from the bottom of the device, it's likely best to have the most-used fields near the bottom
    - I'd recommend date@time at the top, then stars, grade, tries, note (kept where it is to avoid fat-fingering the submission buttons while adjusting tries), and finally the attempt/submit buttons (maintained) - example image below
        Image
        - Usage data may suggest a different priority - this order is just one that seems more reasonable (to me) than the current ordering

marcodejongh:
## Summary

The tick sheet stacked its rows **Grade → Tries → Stars → Date → Time → Note → buttons**, which put the two least-touched controls in the middle of the thumb zone and gave each of them a full row. #4163 flagged the ordering and asked whether usage data agreed.

It does. Field interaction rates, mobile, last 90 days (26,689 `Quick Tick Saved` events):

| Field | Touched | Rate |
|---|---|---|
| Tries | 7,544 | 28% |
| Stars | 6,426 | 24% |
| Grade | 1,728 | 6.5% |
| Note | 362 | 1.4% |
| Date / Time | ~400 | ~1% |

Date/time has no analytics event, so that number comes from prod `boardsesh_ticks`: of 41,235 native ticks, 5.4% have `climbed_at` more than 12h before `created_at` — but **1,842 of those 2,236 rows belong to just 4 users**, i.e. bulk historical imports rather than the picker. Genuine backdating is under 1%, and some of that is offline-drain lag.

New order, least-used at the top to most-used at the bottom:

**Date @ Time → Grade → Stars → Tries → Note → buttons**

This differs from the issue's mockup in one place: it swaps Stars and Grade, because Stars is used ~4x more than Grade and so belongs nearer the thumb. Date and time now share one row separated by an `@`, as suggested. Note keeps its slot directly above the save row — rarely used, but a text field there is far less fat-finger-prone than a picker would be.

### Also fixed: clipped save buttons on small phones

While measuring I found `QuickTickBar` has no scroll container. Its fixed content is ~450pt; the sheet's 60% snap gives ~400pt on an iPhone SE, so the Attempt/Send buttons were clipped with **no way to reach them** — the 92% snap is only reached by focusing the note field. The rows now sit in a `BottomSheetScrollView` with the error row and save row pinned below, mirroring the pattern already in `ClimbFilterSheet`. Merging date+time buys back ~44pt on top of that.

A scroll body with a pinned footer only holds together if the column above it is clamped to the sheet's detent, and `LogAscentSheet` was the one pinned-footer sheet in the app that never did that. On iOS the `@expo/ui` SwiftUI sheet host can propose an *unbounded* height to its RN content, so a `flex: 1` column sizes to its content instead: the scroll view never gains an overflow, nothing scrolls, and the save row lands past the detent — the exact #3330 failure this section claims to fix. PlayDrawer keeps `LogAscentSheet` mounted for the life of the player, so every tick open after the first is a re-present of an already-mounted host, which is the path most prone to it.

So `LogAscentSheet`'s content column now goes through the shared `useSheetColumnStyle` hook, tracking the resting detent from the native `onChange` so the bound follows the 60% ↔ 92% move and drops back to 60% on close. Android keeps `flex: 1` — its Material sheet bounds the column natively.

`keyboardShouldPersistTaps="handled"` is set because the scroll view now hosts the note field alongside every picker: without it, the first tap on the grade rail / stars / tries / date while the keyboard is up gets swallowed dismissing the keyboard. The save buttons sit outside the scroll view, so they're unaffected either way.

JS-only, no native fingerprint change, so it rides OTA.

Closes #4163

## Release Notes

- The tick sheet is laid out for your thumb now — tries and stars sit where your hand already is, and date and time share one line up top
- Fixed the Send button being cut off on smaller phones

## Test plan

- `vp run test:mobile` — 5,123 tests across 585 files, all passing
- `vp check` on the changed files — clean

New tests in `quick-tick-bar.test.tsx`:
- Row order is pinned (date, grade, stars, tries, note) so a future edit can't quietly shuffle it back
- Date and time render on one row with the `@`, and the time label no longer takes a row of its own
- The **time** field commits into `saveTick` — `climbedat-time` was previously rendered by the test mock but never exercised, so nothing caught a regression where only the date field pinned the timestamp

New tests in `log-ascent-sheet.test.tsx` (window 844, top inset 44, iOS 26 → 60% = 446pt, 92% = 694pt):
- The column is pinned to the 60% detent on iOS rather than flexing to content
- It grows to the 92% detent when the sheet reports settling there
- It drops back to 60% on close, so the next present of the always-mounted host starts bounded
- Android keeps `flex: 1`

Device QA still owed (notes in `.boardsesh/qa-notes.md`):
- Date/time row on a 320pt-class screen — iOS renders a compact `DateTimePicker` that clips rather than wraps when squeezed, which is why the fields use `flexShrink` instead of an even `flex: 1` split
- Second and third open of the tick sheet from the player on an SE-class iPhone — the re-present path #3330 bites on; the save row must stay reachable and the rows must scroll
- Keyboard contract: sheet still extends to 92% on note focus, and the save row stays above the keyboard
- Android: the date/time pills hug their content instead of stretching — intended, matches the mockup

## Out of scope

`LogbookEditSheet` also stacks two `ClimbedAtField`s, but it's a different surface — editing an existing tick is a flow where changing the date is a primary reason to open the sheet, so date-first is correct there.


marcodejongh:
Addressed the review feedback in `723bff1` and `dcd4b84`. Three of the notes were about comments describing behaviour the code doesn't actually have — worth writing down, because two of them were wrong *before* this PR too:

- **`keyboardShouldPersistTaps`** was credited with rescuing taps on Attempt/Send. Those render outside the scroll view, so the prop can't touch them. It's still needed — for the grade rail, stars, tries and date/time, which are all inside it. Comment corrected so a later edit doesn't drop the prop after checking the buttons and concluding it's dead.

- **`BottomSheetTextInput` does not expand the sheet on native.** The pre-existing comment claimed it did. `@expo/ui` re-exports it verbatim — `const BottomSheetTextInput = TextInput` — and documents `keyboardBehavior` as *"accepted for API compatibility but has no effect. Native sheets handle keyboard behavior automatically."* So the expansion is the OS sheet's job, not ours, and nothing in our code asserts it. Only the Gorhom web path behaves the way the old comment described.

  **This is the one open risk.** If the native sheet doesn't self-expand on focus, the note row and the save buttons both sit under the keyboard. It's not a regression — those rows were unreachable before this PR for a different reason (clipping) — but it's the part of "the form is now fully reachable" that this change doesn't deliver on its own. Called out explicitly in `.boardsesh/qa-notes.md` as a device-QA gate; if it fails there, it needs a follow-up (`automaticallyAdjustKeyboardInsets` plus keyboard-height padding on the scroll view).

- **The date/time fields overflow rather than clip** when squeezed. The iOS compact `DateTimePicker` takes a hard width from its shadow node and `View` defaults to `overflow: visible`, so an over-tight wrapper gets the date drawn *on top of* the time, not cut off. They fit unsplit down to a 320pt screen at default type (~232pt available, ~216pt needed), so this only bites under large Dynamic Type — added `flexWrap` so the row degrades to two lines there instead of overlapping, and added it to the QA checklist.

Also narrowed the new time test's claim: the stubbed `ClimbedAtField` ignores `mode` and returns a whole `Date`, so `applyTimePart` never runs in it. It's a wiring guard — it pins that the time field is still connected to `handleClimbedAtChange` now that the two fields share a row — not a merge test. Comment says so now rather than overstating.

marcodejongh:
Fable review pass. Two findings, both handled.

**Critical — the iOS sheet never bounded the new scroll column (`25170a1c3`).**

The reorder itself is data-backed and correctly wired, but the clipped-save-buttons half of the fix rested on an assumption the repo's own device-verified docs contradict. Turning `QuickTickBar` into a scroll body with a pinned save row only works if the column above it is clamped to the detent, and `LogAscentSheet` was the one pinned-footer sheet in the app that didn't do that — `ClimbFilterSheet`, `ModalSheet` and `Sheet` all pin theirs JS-side via `useSheetColumnStyle`.

Per #3330 (and the comments in `use-sheet-column-style.ts`), the iOS `@expo/ui` SwiftUI sheet host can propose an *unbounded* height to its RN content. A `flex: 1` column then sizes to its content, the `BottomSheetScrollView` never gains an overflow so it can't scroll, and everything past the detent — including the pinned save row — lands off-screen. That's the exact bug this PR set out to fix, and PlayDrawer keeps `LogAscentSheet` mounted for the life of the player, so every tick open after the first is a re-present of an already-mounted host: the path `ClimbFilterSheet` documents as the one that loses the bound.

`LogAscentSheet`'s content column now goes through `useSheetColumnStyle(snapPoints, { activeIndex })`, with `activeIndex` tracked from the native `onChange` since this sheet has two detents (60% / 92%) — so the bound follows the keyboard extension and drops back to 60% on close rather than carrying a stale tall column into the next present. Android is untouched (`flex: 1`; its Material sheet bounds the column natively).

Four new tests in `log-ascent-sheet.test.tsx` pin it: 60% → 446pt, 92% → 694pt, reset-to-60%-on-close, and `flex: 1` on Android. Mutation-checked — reverting the column to `flex: 1` fails three of them.

I did **not** add `ClimbFilterSheet`'s `presentEpoch` remount-on-re-present. Once the column carries an explicit numeric height, what the native host proposes to it no longer matters, so the remount is redundant on iOS — and on Android (where the column stays `flex: 1`) the host is rebuilt natively on each present anyway. Remounting on every tick open would also reset a half-filled form and add a native-transition edge to a hot path for no layout gain. Happy to add it if device QA shows the pin isn't enough.

**Minor — PR body claimed the wrong mechanism for `keyboardShouldPersistTaps`.**

The body said it was needed so the first tap on Send with the keyboard up fires; the save buttons are outside the scroll view, so they're unaffected either way (the in-code comment already said so). The prop actually protects first taps on the pickers / grade rail / date fields inside the scroll view. Body corrected — no code change.

Validation: `vp run typecheck:mobile`, `vp run test:mobile` (5,123 tests / 585 files), `vp run check:mobile-bundle`, `vp check` all clean.

marcodejongh:
Addressed in `6bd1a06`:

- **Fragile test helper** — real problem, fixed. The mocked `View` puts `data-style` on every div, so `querySelector('[data-style]')` was positional; a wrapper added above the column would have kept passing while measuring the wrong node. The column now carries `testID="log-ascent-column"` and the helper throws instead of returning `null` if it stops rendering. Re-ran the mutation check afterwards — reverting `columnStyle` to `flex: 1` still fails 3 of the 4 bound tests, so the guard survived the rewrite.

- **`docs/mobile-sheets-vs-routes.md`** — added. `LogAscentSheet` is listed alongside `ClimbFilterSheet`, plus the bit that wasn't obvious to me when I wrote this PR: the hook only becomes load-bearing once a surface gains a scroll body or a pinned footer. `LogAscentSheet` sat on a plain `flex: 1` column safely for a long time precisely because it had neither, which is why copying `ClimbFilterSheet`'s scroll view without its column bound looked fine. Also noted that multi-detent sheets need `activeIndex`.

- **Comment style** — leaving as-is. The row-order block was already trimmed once in `723bff1`, and the review of that commit called the remaining version "concise" and said the measured interaction rates are "genuinely non-obvious context not available from reading the code." I agree with that read: without the rates the ordering looks arbitrary, and a bare issue link doesn't survive the issue being closed. Not going to churn it a third time.

- **Device QA** — correct, and it's why this is still a draft. The four gates are in `.boardsesh/qa-notes.md`. Tests can't settle the iOS height behaviour or the keyboard contract.

PR #4216, opened 2026-08-08
fix: extend the multi-board-per-config rework to edits, tick attribution, and a board cap

marcodejongh:
Follow-up to #4174, which dropped `user_boards_unique_owner_config` so an owner can hold the same board config at different places. That PR's review round left three pre-existing paths for their own change; this PR fixes all three, plus what its own review rounds surfaced.

## updateBoard parity with createBoard

The old strict guard refused any config change that matched another of the owner's boards — no place dimension, no override, plain `Error`. A gym reconfiguring one wall to match another of the owner's walls hard-failed, and the delete+recreate workaround loses tick linkage.

Now `updateBoard` uses the same rule as `createBoard`: the duplicate probe runs `findBlockingDuplicate` over the owner's same-type/layout/size boards (excluding the board being edited) against the **post-update** location, is bypassable with `allowDuplicateConfig`, and throws the same `BOARD_DUPLICATE_CONFIG` error. The guard also skips entirely when the effective config didn't change — mobile resends the config on every edit, so without that skip an owner of two same-config boards could never save *any* edit, not even a rename.

Both clients handle the new error: mobile edit shows a native confirm (keep editing / save anyway), web's edit drawer gets the same dialog the create form has. "Save anyway" retries with `allowDuplicateConfig: true`.

**Found in this PR's own review round** (same class as the #4174 private-gym probe finding): the duplicate error's extensions carry the blocking board's name, slug, and human-readable location. `updateBoard` is reachable by non-owner editors (gym admins, community leaders), so that would have let an editor enumerate an owner's *other* boards — including a private home wall's location — by probing configs. Identity fields are now stripped unless the caller is the owner; clients show a generic message in that case. Regression-tested with an owner/non-owner contrast pair.

## Deterministic legacy tick attribution

`resolveBoardFromPath` (the fallback when a tick arrives with neither `boardUuid` nor a presence `boardId`) had two defects once same-config duplicates became legal:

- `setIds` was compared as a raw SQL string, while the stored value keeps creation order — `'25,26,27,24'` vs `'24,25,26,27'` silently returned null and the tick landed unattributed. Equality is now settled in JS via the same `normalizeSetIds` the other resolution rungs use.
- `.limit(1)` with no `ORDER BY` picked an arbitrary board among same-config siblings, and unstably — two ticks in one session could land on two different boards. The pick is now `asc(id)`, and a new rung was added above the config lookup: if the tick carries a `sessionId`, the session's own `board_id` wins (validated against the tick's full config, exactly like the presence-boardId rung; no ownership requirement, since party-mode sessions legitimately reference another user's wall). An MRU-by-ticks preference was considered and deliberately left out — it can flip when the user ticks a sibling board via the modern path, making the legacy pick less stable, not more.

The ordering test is mutation-tested: seeded with explicit descending ids so the old unordered pick returns the wrong board.

## Per-account board cap

The dropped unique index was an implicit ceiling on boards per account; `allowDuplicateConfig` is otherwise an unlimited mint. `MAX_BOARDS_PER_ACCOUNT = 50` (repo precedent: 20 distinct foreign gyms, 25 auto-minted gyms, 100 treated as pathological). Enforced in `createBoard` (before the duplicate guard, outside the bypass) and in the BLE serial auto-create branch — binding to an already-owned board is never capped, so connecting to your own wall keeps working at the cap. Same TOCTOU-approximate stance as the gym mint cap (a runaway backstop, not an invariant), but this one throws: a board the user asked for must not silently not exist. Error code `BOARD_LIMIT_REACHED`; both clients show a typed message instead of the raw server string.

The review round also closed a bypass: `updateBoard`'s soft-delete restore path revived boards without a cap check (delete 10, mint 10, restore 10 → 60 live). A restore now pays the same toll as a mint.

## Also in here

- **German catalog repair**: `de/boards.json` was missing 30 keys (red on `main`'s parity test) and carried 2 stale ones. Translated per `docs/i18n-german-glossary.md`. ⚠️ The German strings (and the new es/fr keys) are glossary-checked but machine-translated — flagging for native review, cc Axel.
- `buildUpdateInput` on mobile now omits an unchanged config (defense-in-depth; the server-side skip is the load-bearing fix).
- French register: new keys use vous to match their screen siblings.

## Deliberately left / follow-up material

- `updateBoard` hard-fails a location edit on a no-PostGIS deployment where `createBoard` degrades gracefully — pre-existing asymmetry, untouched.
- The `boardUuid` tick rung attributes to any board by uuid with no config gate (pre-existing; the new session rung is strictly more constrained).
- At the cap, a duplicate resubmit reports the cap error rather than the duplicate error (cap runs first by design).
- `normalizeSetIds` (board-presence) and `normaliseSetIds` (@boardsesh/board-config) now coexist in `social/boards.ts`; each comparison uses one helper on both sides, but consolidating them is a worthwhile cleanup.

## Release Notes

- Changing a board's setup no longer gets blocked just because you own the same setup somewhere else — the app asks you to confirm instead.
- Climbs logged from the classic board pages now land on the right board when you own more than one with the same setup — and if you're in a session, on the wall that session is running on.
- The app now speaks German on the boards screens it was missing it on.

### DIFF
--- a/packages/backend/src/__tests__/board-account-cap.test.ts
+++ b/packages/backend/src/__tests__/board-account-cap.test.ts
@@ -48,7 +48,7 @@ async function seedBoard(opts: {
   config?: { boardType: string; layoutId: number; sizeId: number; setIds: string };
   deleted?: boolean;
   name?: string;
-}): Promise<number> {
+}): Promise<{ id: number; uuid: string }> {
   const uuid = uuidv4();
   const config = opts.config ?? CONFIG;
   const result = await db.execute(sql`
@@ -58,7 +58,7 @@ async function seedBoard(opts: {
             ${config.setIds}, ${opts.name ?? 'Seeded board'}, true, ${opts.deleted ? sql`now()` : null}, now(), now())
     RETURNING id
   `);
-  return Number(Array.from(result as Iterable<{ id: number }>)[0].id);
+  return { id: Number(Array.from(result as Iterable<{ id: number }>)[0].id), uuid };
 }
 
 /** Seed `count` live boards for OWNER in one statement per board. */
@@ -187,7 +187,7 @@ describe('per-account board cap', () => {
     // at the cap — otherwise reaching the limit would lock a climber out of
     // their own boards.
     await seedBoards(MAX_BOARDS_PER_ACCOUNT - 1);
-    const existingBoardId = await seedBoard({ config: OTHER_CONFIG, name: 'The connected wall' });
+    const { id: existingBoardId } = await seedBoard({ config: OTHER_CONFIG, name: 'The connected wall' });
     resetAllRateLimits();
 
     const resolved = await boardPresenceMutations.resolveBoardForSerial(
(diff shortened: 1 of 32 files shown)


marcodejongh:
Re the reviewer's normalizer-split flag: tracked as #4217 (deliberately out of scope here — consolidating means touching the board-presence tick rungs, which want their own change + review). The other two findings (mobile edit's cap error copy, dialog close flicker) are being fixed on this branch.

marcodejongh:
Re the reviewer's index concern on `resolveBoardFromPath`: already covered. #4174's migration `0189_allow_duplicate_board_configs.sql` replaced the dropped unique index with `user_boards_owner_config_idx` — btree on `(owner_id, board_type, layout_id, size_id, set_ids) WHERE deleted_at IS NULL`. The new candidate query filters on a strict prefix of that index, so the O(log n) path is intact. The other findings (web cap handler, `|| undefined`, test-isolation hardening) are being fixed on the branch.

marcodejongh:
Round 4 triage: normalizer split is tracked in #4217. The BLE cap-before-catalog error ordering is deliberate — the cap is the cheaper check and refusing early avoids catalog work for a mint that can't proceed; noted for support. The French tu/vous difference between `mobile.create.limitReached` and `boardForm.create.limitReached` is per-surface register matching (each string follows its screen's existing voice — the mobile create screen is tu, the web board form is vous); the whole-catalog register question is part of the native-review pass already flagged in the description.

PR #4222, opened 2026-08-08
fix(mobile): make the play-drawer chart text readable

### LINKED ISSUE #4164: [UI] Climb: Charts: make text details readable
1. The "Boardsesh grade" chart has the angle-specifier labels on the horizontal axis straight across, which truncates the angles on some devices
    - The "Community" chart just below it solves the same issue by angling the axis labels (which was done in #3221) - it definitely makes sense to have consistency between the two charts
2. That same chart also does not adjust the vertical axis correctly for low-graded climbs, so the display overlaps the horizontal axis ticks
    - The axis should be scaled upwards to account for the different grade options
3. When French grades are enabled, only some of them get shown on the vertical axis, and not necessarily even the grade of the climb being displayed
    - It would be nice to show at least tick labels for all french grades
    - Ideally there should also be matching grid lines, though maybe they should be dotted for the ones in-between different V grades 🤷‍♂️ 
4. The "Community" chart labels each vertical bar with the grade it represents, but poorly handles the case where both grade formats (V + French) have been enabled in the settings, by overlapping the labels on each other and the surrounding bars
    - It could help to angle the grade vertically in that case (or just in general)
    - A similar problem on a different chart was seemingly solved in #4031 (though I haven't yet had the chance to confirm the result - it was recently merged so may not have been pushed as a public build yet)

ImageImage

marcodejongh:
## Summary

Closes #4164.

Both charts in the climb play drawer lose text on a phone. Four separate faults, one per bullet in the issue.

**Boardsesh grade (the dumbbell chart)**

1. The angle labels ran straight across and clipped to `7…` once a board's 15 angles crowded the axis. They now turn diagonal at the same threshold the Community chart below already uses (#3221).
2. The grade window padded by two ids and then clamped at the easiest real grade, so the bottom padding silently vanished for a V0 climb and its markers landed on the x-axis line, on top of the angle labels. The padding now runs below the scale — the lines are drawn, just unlabelled.
3. Ticks were spread evenly across the window and then coarsened whenever two labels repeated, so a French reader got a handful of grades and often not the climb's own. There is now a gridline per difficulty id — ids map 1:1 to Font grades — and one rule for the labels: print one unless it repeats the tick below it. French labels every line; V-grade labels the line where each new V starts and leaves the in-between ones bare, which is the "in-between lines" the issue asks about. They are solid rather than dotted: gifted-charts can't dash an individual rule, and a hand-rolled overlay is a lot of alignment risk for a `🤷‍♂️`.

An unlabelled tick has to be a **space**, not an empty string — gifted prints its numeric fallback for `''`, so a bare blank renders as a stray `0`. There's a test pinning that.

**Community (ascents by angle)**

4. The grade above each bar sat in a fixed 60px box anchored left, so with both grade formats on, `V6 / 7A` spilled over its neighbours and the bars beside them. The box is now sized to the room the bar actually owns and re-centred over it, and a grade too wide for that stacks onto two lines (`V6` over `7A`) before falling back to standing vertical.

The fits → stack → rotate maths is the You page's `bar-top-label` module from #4031, generalized from counts to text and moved to `src/lib/chart/` now that two charts use it. The `V6 / 7A` separator gets a single owner: `splitGradeLabel` in `@boardsesh/play-view`, next to the `formatGrade` that joins it.

## Release Notes

- Grade charts on a climb now read properly on a phone: angles no longer clip, low-graded climbs sit clear of the axis, and every French grade gets its own line
- Turning on both grade formats no longer smears the grades over each other

## Test plan

- `vp run typecheck` — clean
- `vp run test:mobile` — 5135 passed
- `vp check` — 28 errors / 298 warnings, byte-identical to `main` (all pre-existing)
- New/updated tests, by fault:
  - **(1) rotation** — `dumbbell-by-angle-chart-axes.test.tsx`: fifteen angles rotate and get the box width gifted sizes off `labelsExtraHeight`; three angles stay upright.
  - **(2) V0 clearance** — `by-angle-comparison.test.ts` → `'keeps a V0 climb off the x-axis instead of clamping the padding away'`, plus an all-pairs sweep asserting `minId < low` for all 300 grade combinations on the scale.
  - **(3) French ticks** — same file: a labelled line per Font grade, and blanks are `' '` not the `''` gifted prints as `0`.
  - **(4) stacking** — `bar-top-label.test.ts` (fit → stack → rotate ordering) and `difficulty-by-angle-chart-labels.test.tsx` (all three grade formats end to end).
- ⚠️ **Not verified on a device.** The emulator's software GPU segfaults during boot on the machine this was written on, under both `swiftshader_indirect` and `guest`, sandboxed or not — no screenshot was ever produced. Everything above is unit tests and static analysis. The two things pixels would tell us that tests can't are the stacked two-line grade label and the dumbbell's rotated angle labels; QA steps for both, and for the other two faults, are in `.boardsesh/qa-notes.md`.

JS only — no native dependency change, so this rides OTA.


marcodejongh:
Thanks — two of these are real and are pushed in 81e63d2. The other two I dug into and am leaving as-is; evidence below.

### ✅ `bar-top-label.ts` importing `splitGradeLabel` — fixed

Fair. The split now arrives as a `stackInto` argument, so `lib/chart` is pure geometry again: the Community chart passes `splitGradeLabel(longestGrade)`, and a plain count passes nothing and can only fit or rotate. There's a test for the no-split case so a future caller can't silently lose the rotation fallback.

### ✅ Magic 150 in the test — fixed

Named `CHART_HEIGHT` alongside `DRAWER_WIDTH`, matching how the You-page test declares it.

### ❌ `topLabelComponentHeight` removed without replacement — the prop is dead code

It isn't read by anything at these versions. Every occurrence of the string across both installed packages:

```
$ grep -rn "topLabelComponentHeight" node_modules/.../react-native-gifted-charts \
                                     node_modules/.../gifted-charts-core
gifted-charts-core/dist/BarChart/types.d.ts:486:    topLabelComponentHeight?: number;
```

One hit, in a `.d.ts`. No implementation destructures it, so setting it to 16 did nothing before this PR and removing it changes nothing.

What actually positions the box is `Animated2DWithGradient.js`:

```js
var topLabelPosition = (item.barWidth || barWidth || 30) * -1;
// ...
{ position: 'absolute', top: topLabelPosition,
  height: barWidth, width: barWidth,
  justifyContent: 'flex-end', alignItems: 'center' }
```

`top: -barWidth` with `justifyContent: 'flex-end'` bottom-aligns the label in the strip directly above the bar and grows it upward — plus the `marginBottom: spacing[1]` we still pass through `topLabelContainerStyle`. A single-line label sits exactly where it sat before. The `height={CHART_HEIGHT - headroom}` compensation is only for the stacked/rotated cases because those are the only ones that need *more* than that strip.

### ❌ `model` useMemo rebuilding on every layout event

The new deps are `stackGrades` / `rotateGrades` / `gradeMarginBottom`, not `barWidth`. The first two are booleans, so the memo only rebuilds when the label *treatment* flips, not on every pixel of width change. And `gradeMarginBottom` is `round((textWidth - lineHeight) / 2) + 2` where `textWidth` comes from the label string and font scale only — it's independent of `barWidth`, so it doesn't churn either.

In practice that's at most one extra rebuild on mount (`width: 0` → measured, which can flip `rotated`), and none afterwards unless the drawer is actually resized. Pulling `topLabelComponent` out wouldn't help much regardless: gifted takes it as a per-item callback, so there's a closure per bar either way, and `width={plotWidth - 8}` already makes gifted rebuild its own tree on any width change.

---

Worth flagging on the PR generally: I could not get visual confirmation on this host. `vp run mobile:android-shots` segfaults the emulator's software GPU during boot under both `swiftshader_indirect` and `guest`, with and without the sandbox — the documented "some VMs can't run the emulator's software GPU at all" case. The x-axis rotation geometry in `DumbbellByAngleChart` is the one part of this that unit tests can only cover at the props level (`rotateLabel` + `labelsExtraHeight` are pinned in `dumbbell-by-angle-chart-axes.test.tsx`), so it wants a real device pass against `.boardsesh/qa-notes.md` before merge.

marcodejongh:
Nits 1 and 3 are pushed in 8625a89.

**3 (separator over-charged)** was the one with teeth — you're right that `' / '` costs ~15% more than it measures, and the consequence isn't just conservatism: on a 4–6 angle climb the column sits right in the 41–52px band, so the over-estimate was stacking grades on bars wide enough to hold them on one line. `estimateLabelWidth` now bills space/slash/period/comma at their own em. Digit-only labels are unchanged to the pixel, so the You-page maths from #3779 is untouched.

**1 (blank ticks in the reduce)** — filtered, with the initial value that makes the all-blank case safe rather than a throw.

**2 (`key={line}` vs index)** — leaving as-is. `splitGradeLabel` returns at most two entries and they're always distinct (a V grade and a Font grade can't collide), so the content key is stable and carries more meaning than a positional one. Happy to switch if you feel strongly.

`vp run test:mobile`: 5133 passed.

marcodejongh:
Three of four pushed in 1fe6a17.

**`longestGrade` measured by char count** — good catch, and it became a real mismatch the moment `estimateLabelWidth` learned about narrow glyphs in the previous commit. It now compares rendered width, so the two agree by construction rather than by luck of the current grade set.

**Inline `topLabelContainerStyle`** — memoized on the two numbers it carries.

**Dead `id > MAX_DIFFICULTY_ID` guard** — removed. You're right that it's unreachable; the comment above the loop now says why (the divisibility stretch goes downward on purpose, so only the bottom can leave the scale).

**`key={line}` → `key={index}`** — still leaving it. Beyond the parts being distinct by construction (`formatGrade` joins a V grade with a Font grade, which can't render identically), this repo lints against index keys: `.oxlintrc.json:10` sets `"react/no-array-index-key": "warn"`. Switching would trade a hypothetical warning for a real one.

`vp run test:mobile`: 5133 passed.

marcodejongh:
All three pushed (052831b, plus 2678384 removing a probe file that rode along by mistake).

**`longestLabel` by char count** — right, I fixed this in `DifficultyByAngleChart` last round and missed the twin in `by-angle-comparison`. Both now compare rendered width.

**Single-angle `rotateLabels`** — good spot. `count === 1` makes `rawSpacing` `0`, which read as maximally crowded. It was harmless because that path returns a sentence before the chart renders, but the expression said the wrong thing. Guarded on `count > 1`.

**`key={line}`** — taking it, and my earlier objection was wrong. I claimed the index form would trip `react/no-array-index-key` (warn, `.oxlintrc.json:10`), but I actually probed it rather than assuming this time: a file with `` key={`${index}-${line}`} `` lints clean. So the compound key costs nothing and is safer. Changed.

`vp run test:mobile`: 5133 passed. `vp check` clean on both touched directories.

marcodejongh:
All three looked at; one turned into a test (89ad248).

**3 — `noOfSections` exceeding `MAX_GRIDLINES`.** Worth checking rather than arguing, so the axis suite now sweeps all 300 low/high grade pairs the scale can produce and asserts the cap, an integer section count, the `maxValue === maxId - minId` geometry invariant, and no empty-string ticks. The cap holds: the widest the real scale reaches is 9 sections, because `minId` bottoms out at 8 and `maxId` at 33, so span ≤ 27 and `idsPerSection` only ever needs to reach 3. The ~37-id span the concern needs isn't constructible.

The sweep did earn its keep, just not where expected — it caught that a **V16 climb sits on the top gridline**. That's deliberate (clamping the top is what keeps the top tick from being labelled with a grade that doesn't exist, and unlike the bottom there's no axis line or label row up there to collide with), and it's pre-existing behaviour, but it was only implied by a comment. Now pinned explicitly so the asymmetry with the bottom-edge fix reads as a choice.

**1 — `topLabelComponentHeight` and the "30px default".** There is no 30px default, because nothing reads the prop. Grepping both installed packages for the identifier, in full:

```
gifted-charts-core/dist/BarChart/types.d.ts:486:    topLabelComponentHeight?: number;
```

That's the complete result set — one `.d.ts` line, zero implementation. Setting it to 16 did nothing before this PR. The 30 in `Animated2DWithGradient.js` is the `barWidth` fallback in `(item.barWidth || barWidth || 30)`, which sizes the box off the bar and is unrelated. Your instinct to get the stacked case on a device is still right, though, and it's in `.boardsesh/qa-notes.md` — see the caveat below.

**2 — `computeBarTopLabelLayout` outside `useMemo`.** Leaving it. The call is one pass over a ≤10-character string plus a 2-element reduce; a `useMemo` with three deps would cost more in bookkeeping than it saves, and as you note nothing downstream is destabilised because both `gradeBoxStyle` and the `model` deps read primitives off the result.

---

⚠️ **Still unverified on a device.** The emulator's software GPU segfaults during boot on this machine under both `swiftshader_indirect` and `guest`, sandboxed or not, so I could not get a single screenshot. Everything above is static analysis and unit tests. The stacked two-line label and the dumbbell's rotated angle labels are the two things pixels would tell us that tests can't — QA steps for both are in `.boardsesh/qa-notes.md`.

marcodejongh:
**Correction first:** the review says "Given device QA was done this is likely fine" — device QA was **not** done, and can't be from here. The emulator's software GPU segfaults during boot on this host under both `swiftshader_indirect` and `guest`, sandboxed or not. Nothing in this PR has been seen rendered. Flagging it plainly because it's the one thing that would change how much weight to put on the rest.

**Loose `height` contract** — you're right, and it's now removed (f316a4a). It was a leftover from when it was meant to feed `topLabelComponentHeight`, which turned out to be a type-only prop nothing reads, so it had no consumer left. The test that referenced it now asserts what it actually cared about: a two-line label costs exactly one extra line of headroom (`CHART_LABEL_LINE_HEIGHT`), which reads better than the old `height / 2` anyway.

**`labelLayout` not memoized** — still leaving it, same reasoning as last round: one pass over a ≤10-character string plus a 2-element reduce is cheaper than a three-dep `useMemo`'s bookkeeping. The mobile perf checklist is aimed at per-row work in virtualized lists and context values that churn; this is once per chart render, and as you note nothing downstream is destabilised.

`vp run test:mobile`: 5134 passed. `vp check` clean.

marcodejongh:
One of three taken (8dc4d60); the other two don't hold up.

### ✅ 2 — `longestBarValue` returning a value not in the input

Correct, and worth fixing even though it's cosmetic. `[9, 5, 3]` came back as `0`. I went with seeding the reduce on `values[0]` rather than flipping `>` to `>=`: `>=` would return `3` (the *last* single-digit value) instead of `9`, which is in-input but still arbitrary. Seeding keeps first-wins-on-ties, matching what the name promises. Pre-existing since #3779, with a test now.

### ❌ 1 — `estimateLabelWidth` routing Font labels to rotation

The numbers don't work out, for two reasons.

**Font grades render uppercase.** `formatFontGrade` ends in `.toUpperCase()` (`grade-display.ts:39`), so the label is `6C+`, never `6c+`. None of the narrow letters cited (`i`, `l`, `f`, `r`, `t`) can appear in a grade label — the full alphabet the axis emits is `V`, `A`–`C`, digits, `+`, and the `" / "` separator, and uppercase A/B/C are full-width glyphs.

**The estimate isn't 38px.** Actual outputs:

| label | estimate |
| --- | --- |
| `7A` | 18px |
| `6C+` | 25px |
| `8B+` | 25px |
| `V6 / 7A` | 41px |
| `V13 / 8B+` | 55px |

`6C+` is 25px, not ~38px. For it to rotate, the column would have to be under 25px — a bar plus its gap — which needs roughly 13 angles on a phone. At that density every label rotates regardless of the estimator, so there's no Font-only case being pushed to the wrong fallback.

### ❌ 3 — No V0-clearance regression test

There is one, in `by-angle-comparison.test.ts` — the fault is in the pure axis builder, so that's where it's pinned rather than in the component test:

- `'keeps a V0 climb off the x-axis instead of clamping the padding away'` — asserts `minId < MIN_DIFFICULTY_ID`, that the bottom tick is `BLANK_TICK`, that no tick is `''`, and that the first *visible* label is still `V0`.
- The all-pairs sweep added in 89ad248 asserts `axis.minId < low` for all 300 low/high combinations on the scale, so clearance holds for every grade, not just V0.

Fair that the PR description didn't point at it; adding a line there.

---

`vp run test:mobile`: 5135 passed.

PR #4226, opened 2026-08-10
fix(mobile): keep the tick sheet's action row aligned under the keyboard

marcodejongh:
## Summary

Reported after #4215 merged: the Attempt / Send buttons sit oddly in the tick sheet. The first pass here centred the row, which fixed a real vertical stagger — but QA declined it, and the screenshot shows why: neither button ever filled its slot.

Measuring that screenshot (iPhone 16 Pro Max, 3x) against the layout the code asks for:

| | asked for | rendered |
| --- | --- | --- |
| Attempt width | 114pt | **101pt** |
| Send width | 228pt | **112pt** |
| Attempt height | 56pt | **48.7pt** |
| Send height | 56pt | **55.3pt** |

77pt of dead space between the two buttons, 58pt more to the right of Send, and one pill visibly shorter than the other.

Two causes, both in the shared `Button`:

**The pinned height never reached the screen.** `Host` writes the measured native size back over the RN style — `HostView.swift` calls `setStyleSize` with the SwiftUI content size for every axis `matchContents` covers — so a `height` on a Button's style is overwritten on first layout. `buttonMatchContents` now stops measuring an axis the caller has sized.

**`.buttonStyle()` paints around the LABEL.** Every modifier in the `modifiers` array is applied outside that paint (`View+ModifierArray.swift` reduces them onto the Button itself), so `frame(maxWidth: .infinity)` widened only the tap area and left a pill hugging its text, centred in the leftover space. Growing the label is what grows the background, so the fill frame is now repeated on the SwiftUI `HStack`.

Android already filled correctly — a Compose modifier applies to the container, so `fillMaxWidth()` worked — and gains the matching `height` modifier so a pinned height behaves the same on both platforms.

The tick action row is the only caller pinning a height today, so the vertical half of this is scoped to it. The pin holds at every OS text scale and grows with it — by the label line alone, since a native button's padding does not scale. It cannot be conditional: a button that only sometimes carries a height would flip `matchContents` on a live component, and the axis it stops measuring is the axis nothing else sizes.

Also addresses the two review notes from the earlier pass: the StyleSheet comment blocks are gone, and the test no longer indexes into the cumulative Button spy — it looks each button up by title.

## Release Notes

- Fixed the Attempt and Send buttons in the tick sheet: they now fill the row evenly instead of floating at odd sizes

## Test plan

1. Open a climb → Log a tick.
2. Attempt and Send fill the row, no gaps.
3. Both buttons are the same height.
4. Tap the note field: buttons stay put.
5. Close the keyboard: still even.

## Risk

Risk: 3/5 — touches the shared Button used on ~98 screens. Full-width buttons everywhere now fill their box instead of hugging their label, which is what the code always asked for and what Android already did. Ships OTA.

## Verification

`vp run typecheck:mobile`, `vp run test:mobile` (8,145 passing) and `vp run check:mobile-bundle` are clean. Three mutations of the new guards each fail a test.

Measured on an Android emulator (uiautomator bounds, 810x1800 @ density 320), the same tick sheet before and after:

| | before | after |
| --- | --- | --- |
| Attempt / Send host | 48.0dp | **56.0dp** |
| Native button | 44.5dp | **56.0dp** |
| Widths | 241 / 481 px | 241 / 481 px (2:1, unchanged) |

That is the pinned height reaching the screen for the first time — the half of the fix that is platform-shared. Holds with the keyboard open too.

The SwiftUI label-frame half is iOS-only and has no unit test: the native `@expo/ui` tree cannot mount under vitest, and there is no iOS simulator on this box. That one is what QA is confirming.

https://claude.ai/code/session_01LuSbhLaMGVofXGkjk1o3ob

### DIFF
--- a/packages/mobile/src/components/Button.android.tsx
+++ b/packages/mobile/src/components/Button.android.tsx
@@ -23,11 +23,17 @@ import {
   Text,
   TextButton,
 } from '@expo/ui/jetpack-compose';
-import { defaultMinSize, fillMaxWidth, padding, size } from '@expo/ui/jetpack-compose/modifiers';
+import {
+  defaultMinSize,
+  fillMaxWidth,
+  height as heightModifier,
+  padding,
+  size,
+} from '@expo/ui/jetpack-compose/modifiers';
 import type { ImageSourcePropType } from 'react-native';
 import { useTheme } from '../providers/theme-provider';
 import { brandAccentColor } from '../theme/expo-ui-modifiers';
-import { isFullWidthStyle, makeButtonPressHandler } from './Button.logic';
+import { buttonFillAxes, buttonMatchContents, makeButtonPressHandler, pinnedButtonHeight } from './Button.logic';
 import { sizeConfig, type ButtonProps } from './Button.types';
 import type { IconName } from './icon-map';
 
@@ -131,7 +137,8 @@ export function Button({
           ? TextButton
           : ComposeButton;
 
-  const isFullWidth = isFullWidthStyle(style);
+  const fills = buttonFillAxes(style);
+  const pinnedHeight = pinnedButtonHeight(style);
   const iconSource = icon ? BUTTON_ICON_SOURCE[icon] : undefined;
 
   // Compose has no discrete height buckets: a button is exactly its content plus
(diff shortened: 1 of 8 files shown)


marcodejongh:
image

image


Button alignment still looks weird with both keyboard visible and closed

marcodejongh:

marcodejongh:
image

marcodejongh:
Thanks — two of three actioned, one checked and dismissed.

**1. `TICK_ACTION_LABEL_HEIGHT` is platform-specific** — the comment was, the number isn't unreasonable. Reworded to say what it is: an approximation, deliberately one number for both platforms. An iOS `body` line is ~22, a Compose `bodyLarge` line is 24, so Android under-grows by ~2dp per full step of text scale. The button is exactly `tickActionHeight()` tall on both — the 2dp of slack is padding, not a clipped label. A per-platform constant buys those 2dp at the cost of two numbers that can drift apart.

**2. The description claimed a threshold that no longer exists** — correct, and my mistake: I dropped that behaviour in `e3603920` and left the stale sentence in the body. `tickActionHeight` always returns a height by design; a conditional pin would flip `matchContents` on a live component, and the axis it stops measuring is the axis nothing else sizes. Description fixed.

**3. `true` vs `{ horizontal: true, vertical: true }`** — equivalent. `Host`'s prop type is `boolean | { vertical?: boolean; horizontal?: boolean }`, and it destructures with `typeof matchContents === 'object' ? matchContents.vertical : matchContents`, so a bare `true` and the object form produce the same two native fields. No change.

marcodejongh:
Thanks for the QA pass. Follow-up filed: #5230.

Worth noting what it does and does not cover, since half the fix here already generalised:

**Already repeatable** — the width half. `.buttonStyle()` paints around the label, so moving the fill frame onto the SwiftUI `HStack` fixed content-hugging for **every** full-width Button in the app, not just this row. No per-caller work left there.

**Not repeatable yet** — the shared height. Making two buttons of different native styles land on one height needs a pinned height, and only `tickActionHeight()` + `tickButtonStyles()` in the tick folder do it. Three sheets have the identical shape today (`buttonRow`, two `flex: 1` buttons, mismatched variants, no pin): `InviteSheet`, `PlaylistQueueReplaceSheet`, `EndSessionSheet`. On iOS `glass`/`bordered` measured 48.7pt against `borderedProminent`'s 55.3pt at the same `controlSize`, so each of those is the same crooked row waiting to be noticed. Android already matches — Material shares one `contentPadding` — which is why this only ever showed up on your screenshots.

#5230 proposes a `SheetActionBar` primitive owning the pinned height, the flex split, `flexShrink: 0`, and the reserved error slot, with `TickActionBar` as the reference implementation.

PR #4230, opened 2026-08-10
feat: auto-approve gym claims on unclaimed listings behind an admin toggle

marcodejongh:
## Summary

Every gym ownership claim that can't self-verify by email domain lands in `/admin/gym-claims` and needs a manual **Approve**. Most of those are someone taking over a location-sync listing nobody owns, so approving them is pure clicking.

Adds a global `gym_claim_auto_approve` setting, **off by default**, with a new **Gyms** tab on `/admin` to flip it.

Worth naming what this actually is, because "gym approval" is misleading: gyms themselves are never approved. `createGym` inserts a live row and location-sync imports gyms directly. The only Approve button in the product is `reviewGymClaim` — approving a request to take *ownership* of an existing listing.

### The safety scope

Approving transfers `gyms.ownerId`, so blanket auto-approval would let anyone take over any public gym. Auto-approval is therefore limited to the safe case:

```
requestGymClaim (admin-review path)
  ├─ setting OFF ──────────────────────► queue (unchanged)
  └─ setting ON
       ├─ gym owned by SYSTEM import user ──► applied immediately
       └─ gym owned by a real person ───────► queue (still needs a human)
```

A claim that would displace a real person still goes to the queue, so turning this on can't be used to take a gym away from its owner. Domain-method claims are untouched — they already self-verify via the emailed token.

Flipping the toggle on does **not** drain claims already sitting in the queue; those were filed under review-required rules. A claimant who re-requests will go through, but the queue itself is cleared by hand.

### Privilege split

`setCommunitySettings` is `requireAdminOrLeader`, but gym claims are `requireAdmin`. Without a split, a community leader could enable auto-approval and then claim a gym, escalating straight past the gate on the claim resolvers. So `gym_`-prefixed keys now require full admin to write, and are filtered out of `communitySettings` reads for non-admins. The grade-proposal keys keep their existing leader access.

### Drive-by hardening

`applyGymClaim` read the gym without a row lock and then wrote `ownerId` unconditionally. The UPDATE now re-asserts the owner it based the decision on and rolls the transaction back if a concurrent transfer won. This closes a pre-existing TOCTOU on the manual approve path too.

No migration — the setting rides the existing `community_settings` table. Auto-approved rows are identifiable as `method='admin' AND status='approved' AND reviewed_by IS NULL`.

## Release Notes

- Claim an unclaimed gym and it's yours on the spot, no waiting for a review

## Test plan

- `vp check` — clean (the 30 pre-existing errors on `main` are all in files this PR doesn't touch)
- `vp run typecheck` — passes
- `vp test run --project backend` — `gym-write-access-and-claims.test.ts`: **66 passed**, including 9 new cases
- `vp run test:web` — **478 files / 5888 tests passed**
- `vp run test:mobile` — **598 files / 5251 tests passed**
- `vp run check:i18n` + `check:i18n:orphans` — both OK
- `catalog-completeness` — 90 passed (en-US/es/fr/de parity)

New backend coverage:
- setting off (default) → claim still queues and stays `pending` (regression guard)
- setting on + system-owned gym → `approved`, ownership moves, `sync_frozen_at` set, `reviewed_by` null, no admin email
- setting on + gym owned by a real person → still queues, ownership unchanged
- claim queued before the setting was turned on → re-request goes through
- domain claims unaffected → still `email_sent`
- `gym_` keys reject a global `community_leader`, accept an admin; non-gym keys still accept a leader
- `communitySettings` hides `gym_` rows from a non-admin, shows them to an admin

New web coverage: `gym-settings-panel.test.tsx` — reads saved value, writes `1`/`0` on toggle, and rolls the switch back when the write is rejected.

Not yet run: manual pass against the dev DB.

https://claude.ai/code/session_011UDT9hz3SH1niTKccW679E

### DIFF
--- a/packages/backend/src/__tests__/gym-write-access-and-claims.test.ts
+++ b/packages/backend/src/__tests__/gym-write-access-and-claims.test.ts
@@ -1343,6 +1343,41 @@ describe('requestGymClaim — auto-approval', () => {
     expect(rows.filter((row) => row.status === 'pending')).toHaveLength(1);
   });
 
+  it('never leaks the internal race message out of a manual approve', async () => {
+    // Two admin-method claims on one gym, both approved at once. Whichever way
+    // the transfers interleave, the reviewer must only ever see the curated
+    // message — applyGymClaim's internal "ownership changed" throw is an
+    // implementation detail and must not reach the admin panel.
+    const claimGym = await insertGym({ ownerId: PRIOR_OWNER, name: 'Double Reviewed' });
+    const firstClaim = await insertClaim({ gymId: claimGym.id, claimantUserId: CLAIMANT, method: 'admin' });
+    const secondClaim = await insertClaim({ gymId: claimGym.id, claimantUserId: PLAIN_USER, method: 'admin' });
+
+    const outcomes = await Promise.allSettled([
+      socialGymClaimMutations.reviewGymClaim(
+        null,
+        { input: { claimId: firstClaim, decision: 'approve' } },
+        authCtx(GLOBAL_ADMIN),
+      ),
+      socialGymClaimMutations.reviewGymClaim(
+        null,
+        { input: { claimId: secondClaim, decision: 'approve' } },
+        authCtx(GLOBAL_ADMIN),
+      ),
+    ]);
+
+    for (const outcome of outcomes) {
+      if (outcome.status === 'rejected') {
+        expect(String(outcome.reason?.message)).toBe(
+          'Could not approve this claim — the gym may have been removed or it was already resolved',
+        );
+      }
+    }
+
+    // At least one approval landed, and the gym belongs to a claimant.
+    expect(outcomes.some((outcome) => outcome.status === 'fulfilled')).toBe(true);
+    expect([CLAIMANT, PLAIN_USER]).toContain(await gymOwnerId(claimGym.uuid));
+  });
+
   it('does not auto-approve a domain claim — it still has to prove the domain', async () => {
     await setAutoApprove(true);
     const claimGym = await insertGym({
--- a/packages/backend/src/__tests__/gym-write-access-and-claims.test.ts
+++ b/packages/backend/src/__tests__/gym-write-access-and-claims.test.ts
@@ -1346,10 +1346,15 @@ describe('requestGymClaim — auto-approval', () => {
   it('queues instead of erroring once the auto-approve rate limit is spent', async () => {
     await setAutoApprove(true);
 
-    // The limiter is keyed `${userId}:${operation}`, lives in process memory,
-    // and is NOT reset between tests. Claiming as a brand-new user each run
-    // guarantees an empty budget no matter what ran first — pinning this to a
-    // shared fixture account would make the expected sequence order-dependent.
+    // Every bucket `applyRateLimit` keeps is keyed `${userId}:${operation}`, and
+    // its tier-1 in-process bucket is NOT reset between tests. Claiming as a
+    // brand-new user each run guarantees an empty budget no matter what ran
+    // first — pinning this to a shared fixture account would make the expected
+    // sequence order-dependent.
+    //
+    // Tier 1 being per-process is a test-isolation concern only: authenticated
+    // callers also pass through tier 2, a shared Redis bucket on the same key
+    // (`shared/helpers.ts`), so the production cap is global across instances.
     const claimant = `gw-rate-limited-${uuidv4()}`;
     await insertUser(claimant);
 
--- a/packages/backend/src/__tests__/gym-write-access-and-claims.test.ts
+++ b/packages/backend/src/__tests__/gym-write-access-and-claims.test.ts
@@ -1346,9 +1346,13 @@ describe('requestGymClaim — auto-approval', () => {
   it('queues instead of erroring once the auto-approve rate limit is spent', async () => {
     await setAutoApprove(true);
 
-    // SECOND_TARGET is used by no other auto-approval test, so this account's
-    // in-memory `gymClaimAutoApprove` budget starts fresh (the limiter is
-    // per-user and is NOT reset between tests in a worker).
+    // The limiter is keyed `${userId}:${operation}`, lives in process memory,
+    // and is NOT reset between tests. Claiming as a brand-new user each run
+    // guarantees an empty budget no matter what ran first — pinning this to a
+    // shared fixture account would make the expected sequence order-dependent.
+    const claimant = `gw-rate-limited-${uuidv4()}`;
+    await insertUser(claimant);
+
     const gyms = await Promise.all(
       [1, 2, 3, 4].map((index) => insertGym({ ownerId: SYSTEM_OWNER, name: `Rate Limited ${index}` })),
     );
@@ -1358,7 +1362,7 @@ describe('requestGymClaim — auto-approval', () => {
       const result = (await socialGymClaimMutations.requestGymClaim(
         null,
         { input: { gymUuid: gym.uuid, message: 'mine' } },
-        authCtx(SECOND_TARGET),
+        authCtx(claimant),
       )) as { status: string };
       results.push(result.status);
     }
--- a/packages/backend/src/__tests__/gym-write-access-and-claims.test.ts
+++ b/packages/backend/src/__tests__/gym-write-access-and-claims.test.ts
@@ -12,6 +12,10 @@ import {
   verifyGymClaimByToken,
   hashClaimToken,
 } from '../graphql/resolvers/social/gym-claims';
+import {
+  socialCommunitySettingsMutations,
+  socialCommunitySettingsQueries,
+} from '../graphql/resolvers/social/community-settings';
 
 /**
  * Real-DB coverage for the gym write-access (editor) role + grant/revoke
@@ -195,7 +199,7 @@ let gymUuid: string;
 beforeEach(async () => {
   await db.execute(sql`
     TRUNCATE TABLE
-      "community_roles", "gym_members", "gym_follows", "gym_claims",
+      "community_roles", "community_settings", "gym_members", "gym_follows", "gym_claims",
       "board_follows", "boardsesh_ticks", "user_boards", "gyms", "notifications"
     RESTART IDENTITY CASCADE
   `);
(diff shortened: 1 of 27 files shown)


marcodejongh:
Addressed in `d8ef39c`.

**1. Concurrent ownership-transfer race — real bug, fixed.** Confirmed: the throw propagated out of `requestGymClaim`, so a user whose claim was safely committed as `pending` got a server error. Caught inside `tryAutoApproveAdminClaim` rather than at the two call sites, so any future caller inherits the fallback by construction. The rate limit stays outside the catch — hitting it is a client-facing condition, matching how `requestGymClaim`'s own limit behaves.

**2. `hasAdmin(ctx.userId!)` — the premise here is wrong.** The `communitySettings` query *does* have a `requireAuthenticated(ctx)` guard; it's the first statement in the resolver and predates this PR. So there's no unauthenticated path reaching `hasAdmin`, and no reliance on Drizzle's `eq(column, undefined)` behaviour.

That said, `requireAuthenticated` only asserts `ctx.isAuthenticated`, not that `userId` is populated, so I've made the redaction fail closed anyway — a missing id now redacts rather than leaning on `!`. Worth doing on its own merits for a security filter, not because the guard was absent.

**3. Missing concurrency test — added.** Two users racing for the same unclaimed gym via `Promise.allSettled`. The loser's internal path is timing-dependent (it either sees the new owner at the SELECT and returns null, or loses the guarded UPDATE and throws), so the test asserts the invariant instead: neither request rejects, exactly one returns `approved` and one `admin_review`, and exactly one claim row ends up approved with the other left pending. Deterministic under any interleaving — including fully sequential execution. Ran it 3× locally: 67/67 each time.

marcodejongh:
Addressed in `c1e8b33`.

**1. Stale web state after approval — real, but the suggested fix wouldn't have worked.** The symptom is right: dismissing the dialog left the claim CTA on screen. But mirroring mobile's `invalidateQueries` would have been dead code — web has no `['gym', uuid]`, `['gymMembers']`, or `['nearbyGyms']` key (grep confirms those are mobile-only), and the gym page's `canClaim` is computed in a **server component** (`app/gym/[gym_slug]/page.tsx`), so no client cache invalidation can clear it.

Fixed with `router.refresh()` to re-run the server component, plus invalidating `['myGyms']` — the one key web does have that this affects (the claimant just became a gym owner).

**2. React Query for `GymSettingsPanel` — agreed, converted.** `useQuery` + `useMutation`, with the hand-rolled optimistic rollback replaced by an `onMutate` snapshot restore. One knock-on: `onSettled` now refetches, so the write is no longer the last request, and the tests assert `toHaveBeenCalledWith` rather than `toHaveBeenLastCalledWith`.

**3. Extra DB round-trip on `communitySettings` — not taking this one.** That query has exactly two callers, both admin panels behind an admin gate. Memoising role lookups per request or threading roles through auth middleware is a cross-cutting change to the auth layer, and doing it speculatively for a query that runs when an admin opens a settings tab would add coupling for no measurable gain. Worth doing if roles ever land in a hot path — not here.

**Minor 1 — `gymClaimAutoApproveEnabled()` called twice: it isn't.** The two branches are mutually exclusive. The existing-pending branch `return`s on both of its paths, so the fresh-claim branch is only reached when there was no pending admin claim. Exactly one call per request either way, so there's no IO to halve.

**Minor 2 — `reviewed_by IS NULL` fingerprint:** fair. It holds today because `reviewGymClaim` always passes `reviewerId: ctx.userId!` on the approve path, which is non-null behind `requireAdmin`. A dedicated column would be sturdier, but it needs a migration and the current paths can't produce a false positive, so I've left it with the invariant documented at the call site.

Full local run on this commit: web 478 files / 5889 tests, backend claim suite 67/67, typecheck clean.

marcodejongh:
Both taken, in `1945116`.

**1. Manual approve didn't catch the throw — correct, fixed.** `reviewGymClaim` already had a curated message for the `null` case; my new throw skipped straight past it, so an admin losing an ownership race would have seen `Gym ownership changed while this claim was being applied` instead. Both outcomes mean the same thing to a reviewer — the claim wasn't applied — so they're folded into the one message, with the internal detail logged rather than returned.

Test added: two admin-method claims on one gym approved concurrently, asserting the internal message never surfaces whichever way the transfers interleave, that at least one approval lands, and that the gym ends up owned by a claimant. Ran 3×: 68/68 each time.

**2. `token!` in `mutationFn` — fixed.** Explicit `if (!token) throw` inside the mutation, plus `!token` added to the switch's `disabled` condition, so an expired session can't reach `createGraphQLHttpClient` with null.

marcodejongh:
**1. `replacePendingClaim` return type — fixed in `f09538f`.** Added the guard; better to fail there than forward an undefined claim into `applyGymClaim`.

**2. French register — not taking this one; the premise is inverted.** The comparison is against the *mobile* block, not the surrounding one. The two blocks in `fr/boards.json` have different registers today, and each new string matches its own:

| | existing strings | my addition |
|---|---|---|
| `claimGym.*` (web) | « **Vous** gérez {{gym}} ? », « Saisi**ssez votre** e-mail », « Dites-nous que **vous** gérez », « Elle **vous** contactera » | « {{gym}} est à **vous**. Configure**z**-la quand **vous** voulez. » |
| `mobile.gymClaim.*` (mobile) | « **Tu** gères {{gym}} ? », « Sais**is ton** e-mail », « Dis-nous que **tu** gères », « Elle **te** contactera » | « {{gym}} est à **toi**. » |

Every pre-existing web string in that block is *vous*; every mobile one is *tu*. Switching only the two new web strings to *tu* would make `claimGym` internally inconsistent.

Worth noting the glossary doesn't settle it: `docs/i18n-french-glossary.md` covers *enchaîné*/*la croix*/*allumer*/*essais* and the board's gender, but says nothing about tu/vous — unlike the German glossary, which does mandate informal *du*. So there's no documented rule being broken here. If we want *tu* product-wide on French web, that's a separate sweep over the whole web catalog plus a glossary addition, not two strings in this PR — happy to file it.

**3. TTL cache on `gymClaimAutoApproveEnabled` — not taking.** It's one indexed SELECT on the admin-claim path only, which fires when a user files an ownership claim on a gym with no verifiable domain: rare by construction. A TTL cache would add a staleness window right where it's most surprising (flip the toggle, watch it not take effect), in exchange for IO that doesn't register.

marcodejongh:
**Good catch — you're right, and my earlier justification for this was wrong.** Fixed in `200069a`.

I'd argued the rate limit belonged outside the catch "matching how `requestGymClaim`'s own limit behaves". That comparison doesn't hold: `requestGymClaim`'s limit runs at the **top**, before anything is written, so throwing there leaves no trace. The auto-approve limit runs **after** `replacePendingClaim` has committed the claim — so exhausting it handed the user a rate-limit error for a request that had in fact succeeded in queueing. Same bug class as the ownership-race throw, which is exactly what the docstring said must not happen.

Moved inside the try. Exhausting the limit now means "don't auto-approve this one", not "reject the request" — the claim stays queued for a human, which is the correct degradation for a limit whose purpose is bounding instant hand-overs.

Test added: one account walks past the limit — first three claims are handed over, the fourth returns `admin_review` with its gym still system-owned and its claim `pending`. It uses an account no other auto-approval test touches, since the limiter is per-user, in-memory, and not reset between tests in a worker. 69/69 locally.

---

Separately, on the `test-backend` failure on `f09538f`: it's `session-persistence.test.ts > should cancel pending timer when new update arrives`, the only failure in the run. Fake timers (`vi.advanceTimersByTimeAsync`) driving real Redis, amid a wall of `Redis Connection is closed` teardown noise — session queue persistence, which this PR doesn't touch. Passes locally 30/30. Watching whether it recurs on this run.

marcodejongh:
Three taken, one declined — `fda5f9f`.

**1. Stack traces — fixed, you're right.** `docs/logging.md` explicitly says to pass the underlying `Error` as the trailing arg, and the backend does that at 141 call sites versus 3 using `.message`. I'd copied one of the 3 (the adjacent `sendGymClaimAdminNotification` catch in the same file) and picked up the wrong convention. Both catches now pass `error`.

**2. `gym_` prefix is an implicit contract — documented.** Added the warning at `DEFAULTS`, where new keys actually get added, rather than only on the constant: any `gym_`-prefixed key silently inherits admin-only read *and* write, so something innocuous like `gym_display_order` should be named outside the prefix.

**3. TTL cache — declining, consistent with the last two rounds.** One indexed SELECT, on the admin-claim path only, which fires when someone claims a gym with no verifiable domain. A cache buys nothing measurable and adds a staleness window exactly where it's most confusing: flip the toggle, watch it not take effect.

**4. Untested approved branch — added.** New `claim-gym-dialog.test.tsx` (the component had no test file at all): the approved branch renders the ownership confirmation, exposes the manage CTA as a real `<a href>` pointing at `/gym/:uuid/manage` rather than a click handler, and calls `router.refresh()`; a claim that merely queues shows the review message and does *not* refresh.

---

Also, the earlier `test-backend` failure was a flake, now confirmed: `session-persistence.test.ts > should cancel pending timer when new update arrives` passed on `200069a` without any change to that code. Worth noting I couldn't substantiate my "pre-existing" framing from `main`'s history — `main`'s recent CI failures are all in `lint`, not `test-backend`. The evidence for flakiness is that it passed on re-run and passes 30/30 locally, not that it's a known offender.

marcodejongh:
Fixed the recurring CI flake in `598c38e` — it's an unrelated file, but it has now failed twice on this PR (`5a8563f` and `200069a`), so it's blocking rather than incidental.

`vote-button.test.tsx`, the two un-vote tests. They waited only for the first VOTE request to have been **made**:

```ts
fireEvent.click(screen.getByLabelText('Upvote'));
await waitFor(() => expect(mockRequest).toHaveBeenCalledWith('VOTE', {...}));  // request sent, not settled
fireEvent.click(screen.getByLabelText('Upvote'));                              // ← can hit a disabled button
```

The vote buttons are `disabled={isLoading}` (`vote-button.tsx:200,245,268`), so on a slow runner the second click lands while the first mutation is still pending, gets dropped, and the assertion fails with exactly the observed `expected 2 times, but got 1 times`. Now waits for the button to be re-enabled before clicking again.

Local runs never reproduced it — 13/13 five times before *and* after — so the argument for the fix is mechanical rather than empirical: the second click can no longer land on a disabled button. CI is the real test.

I've kept it as its own commit so it can be split out if you'd rather it landed separately.

marcodejongh:
**1. Rate-limit test order-dependence — fixed in `60a1813`.** You're right, and I'd documented the fragility instead of removing it, which isn't good enough. It now claims as a fresh `gw-rate-limited-${uuidv4()}` user inserted in the test, so the limiter bucket (`${userId}:${operation}`, in-process, never reset) is empty regardless of what ran before. No shared fixture account, no ordering assumption. 69/69 three times.

**2. `gym_` prefix vs an explicit `GYM_ADMIN_ONLY_KEYS` set — declining, and I think the prefix is actually the safer of the two.** The two options fail in opposite directions:

- **Prefix**: a new `gym_*` key is admin-only *by default*. Getting it wrong means a harmless setting is over-protected — visible immediately (the admin panel shows it, nobody else can), harmless if missed.
- **Allowlist**: a new key is world-readable *by default* and only locked down if you remember to add it to the Set. Getting it wrong means a security-relevant setting is silently exposed — invisible until someone goes looking.

For an access boundary I'd rather the failure mode be "too locked down" than "silently open", so the default-deny shape is worth keeping. Your `gym_editor_role` example cuts the same way against the Set: forgetting the entry is exactly as easy as fumbling the prefix, but the consequence is worse. The comment at `DEFAULTS` is there to catch the over-protection case, which is the benign one.

**3. Mobile manage-page shortcut — agreed, filed as #4231** rather than widening this PR. Cache invalidation is already handled in `useRequestGymClaim`; it's purely the missing navigation affordance after an instant approval.

**4. TTL cache — declining, fourth time, same reasoning.** One indexed SELECT on a path that fires when someone claims a domain-less gym.

---

CI on `598c38e` came back fully green: **26 checks, 8077 tests, 0 failures**. Both flakes seen earlier in this PR (`session-persistence`, `vote-button`) are gone — the latter fixed in `598c38e`, the former never recurred.

marcodejongh:
**Mobile `approved` branch — test added in `b57484f`.** Good call: the approved case arrived as a third arm of an existing ternary, which is precisely the shape that regresses silently. Before it existed, anything that wasn't `email_sent` fell through to the admin-review message — that *was* the bug, on both platforms. The test asserts on i18n keys rather than copy, so it pins which branch rendered without duplicating what the catalog-parity tests already cover. 2/2, mobile typecheck clean.

**The rate-limit order-dependence note is stale** — it was already fixed in `60a1813`, one commit before this review ran. `SECOND_TARGET` no longer appears anywhere in that test; it claims as a fresh `gw-rate-limited-${uuidv4()}` user, so the limiter bucket is empty regardless of execution order. The remaining `SECOND_TARGET` references in the file are unrelated write-access-grant tests.

**Prefix allowlist and TTL cache** — answered in the previous round; positions unchanged. Short version: the prefix fails closed and an allowlist fails open, and the cache would trade a staleness window for IO that doesn't register.

**Mobile manage-page shortcut** — filed as #4231.

marcodejongh:
**1. Optimistic update omits `id` — fixed in `bf23cc0`.** Correct: `id` is a required `number` and `createdAt`/`updatedAt` are required strings, and the `as CommunitySettingType` cast hid all three. Now builds a complete row with a clearly-named `OPTIMISTIC_ROW_ID` placeholder and no cast, so the type checks it. 6/6, web typecheck clean.

**2. Rate limiter is per-instance — it isn't, for this path.** `applyRateLimit` is two-tier (`shared/helpers.ts:480-495`):

```ts
checkRateLimit(key, maxRequests);                    // Tier 1: in-memory, per-instance
if (ctx.isAuthenticated && ctx.userId) {
  await checkRateLimitRedis(ctx.userId, operation, maxRequests, RATE_LIMIT_WINDOW_MS, ...);  // Tier 2: distributed
}
```

Every gym claim is authenticated — `requestGymClaim` calls `requireAuthenticated` before anything else — so Tier 2 applies and the cap is enforced globally across instances on a shared Redis bucket keyed `userId:gymClaimAutoApprove`. No `N_instances × 3` multiplier.

**3. Test helper uses raw SQL — declining, it's the file's convention.** That suite seeds entirely through `db.execute(sql\`...\`)`; its header comment says so explicitly ("seeds via raw SQL, calls the resolvers directly against the per-worker test DB"), and `setAutoApprove` sits alongside ~10 sibling helpers (`insertGym`, `insertClaim`, `insertGymMember`, …) that all do the same. The Drizzle-over-raw-SQL rule is about query paths in shipped code; as you note, the production write correctly uses `.onConflictDoUpdate()`. Converting one helper would make it the odd one out.

**4 & 5. Caching `gymClaimAutoApproveEnabled` / `hasAdmin`** — answered in earlier rounds; positions unchanged.

Thanks for the confirmation on the TOCTOU fix and the privilege split — those were the two parts I most wanted a second pair of eyes on.

marcodejongh:
Checked all five; none need a change. Details so the reasoning is auditable rather than just asserted:

**1. Low-privilege users reaching `applyRateLimit`** — they don't. The role check still runs *before* the limiter; only `requireAuthenticated` and input validation moved above it:

```
requireAuthenticated(ctx);
validateInput(...);                       // needed to branch the gate on the key
if (isGymSettingKey(key)) requireAdmin    // ← throws here for a non-admin
else                     requireAdminOrLeader
applyRateLimit(ctx, 10, ...)              // ← unreachable without the role
```

And the limiter is per-user regardless: `key = \`${ctx.userId}:${operation}\`` (`shared/helpers.ts:468`), so one account can't spend another's budget.

**2. In-memory limiter across instances** — as in the previous round, `applyRateLimit` is two-tier and the authenticated path hits Tier 2, a shared Redis bucket keyed `userId:operation` (`helpers.ts:486-489`). Gym claims are always authenticated, so the cap is global, not `3 × N_instances`.

**3. `OPTIMISTIC_ROW_ID = -1` vs a string `id`** — confirmed `number`: `CommunitySettingType.id: number` (`shared-schema/src/types/proposals.ts:94`). `vp run typecheck:web` passes, which it wouldn't if the placeholder were the wrong primitive.

**4. `hasAdmin` role read** — same perf note as prior rounds; position unchanged.

**5. Magic tab indices** — real, and pre-existing as you note. This PR adds one `<Tab>` to an existing index-switched list; reshaping that control flow is unrelated cleanup I'd rather not smuggle into a feature PR. Happy to do it separately if you want it.

---

That's the substance from my side. Rounds 2–5 caught four genuine bugs (ownership race, manual-approve throw, rate-limit placement, unsound cast) and those are all fixed with tests. The last few rounds have been repeats or points that didn't survive checking, so I'm treating the change as complete and marking it ready for review.

marcodejongh:
Good catch — real bug, fixed in `d8e4651`.

Traced it through and it holds exactly as described. A user with only `admin` scoped to a board type:

- passes the page gate, which is `myRoles.some(r => r.role === 'admin')` with no board-type check;
- fails `hasAdmin(ctx.userId)` on the read, because it's called with no `boardType` and the predicate requires `entry.boardType === null || entry.boardType === boardType` — so gym rows get filtered out and the toggle renders **off regardless of its stored value**;
- fails `requireAdmin(ctx)` on the write, so every toggle throws.

A tab that silently misreports a security setting is worse than no tab, so it's now gated on an admin assignment with `boardType == null`, and the panel render is guarded too rather than relying on the tab index alone.

Also added the backend test for the contract the UI mirrors: a board-scoped admin is rejected writing `gym_claim_auto_approve` and never sees the key in a `communitySettings` read. 70/70, web typecheck clean.

Worth noting the same latent mismatch exists in `checkAdmin()` (`packages/web/app/lib/admin/check-admin.ts`), which also ignores `boardType` and gates `/admin/gym-claims`, `/admin/gym-duplicates` and `/admin/feedback` — all backed by `requireAdmin(ctx)` with no board type. So a board-scoped admin can currently open those pages and have every action fail. That's pre-existing and outside this PR; happy to file it separately if you'd like it tracked.

marcodejongh:
Documented at the limit itself in `ba466904` — comments only, no behaviour change.

This finding has come up three rounds running, and citing "per the test comment" pinpointed why: my test comment described the limiter as living "in process memory". That's accurate for tier 1, and it's the part test isolation has to work around, but it reads as though the production cap multiplies by instance count. It doesn't — `applyRateLimit` is two-tier, and an authenticated caller (which every gym claim is) also passes through tier 2, a shared Redis bucket on the same `${userId}:${operation}` key.

So the misleading comment was mine, and it was generating the repeat finding. Both the limit call site and the test now state where enforcement actually happens.

The `hasAdmin` round-trip note is unchanged from earlier rounds — still an admin-only panel, still not worth a cache.

PR #4236, opened 2026-08-10
ci: add weekly scheduled Sentry triage sweep

marcodejongh:
## Summary

Adds a weekly scheduled workflow (`sentry-triage.yml`, Mondays 05:00 UTC + `workflow_dispatch` with a dry-run input) that runs a Claude-driven Sentry triage sweep: pull unresolved issues from the Sentry API, cluster by root cause, dedupe against existing GitHub issues (open and closed, `from-sentry` label and BOARDSESH-short-id search), comment fresh evidence on tickets that already track a cluster, and file at most five new prioritised issues per run.

Guardrails baked into the prompt, learned from the manual sweeps:

- Only trust `production` / `production-web` events (preview/localhost noise leaks into the project).
- Check the release tag distribution before judging severity — old-binary fleet lag is not a regression.
- Root causes must be verified against the checkout, otherwise the issue gets `needs-analysis` instead of a guess.
- Writes are limited to `gh issue create` and `gh issue comment`; it never closes, reopens, or relabels anything.
- Known noise families (#3088 inventory) are excluded from filing.

Uses the new `SENTRY_TRIAGE_TOKEN` repo secret (already set) and the existing `CLAUDE_CODE_OAUTH_TOKEN`. The schedule activates once this merges to `main`; until then the run can be validated with `workflow_dispatch` dry-run from `main` after merge.

## Release Notes

## Test plan

- `vp check` — only pre-existing TS environment-typing warnings elsewhere in the repo; the new YAML lints clean.
- YAML follows the existing conventions in `claude.yml` (action pin `anthropics/claude-code-action@v1.0.99`, `actions/checkout@v6`) and `testflight-feedback-issues.yml` (concurrency group, dry-run input shape).
- After merge: manual `workflow_dispatch` with `dry_run: true` to verify Sentry API access and the report before the first scheduled write run.

### DIFF
--- a/.github/scripts/append-summary.sh
+++ b/.github/scripts/append-summary.sh
@@ -1,0 +1,6 @@
+#!/usr/bin/env bash
+# Append stdin to the GitHub step summary. The triage agent is allowlisted to
+# run this instead of a general-purpose echo, so it has no way to print
+# arbitrary strings (e.g. env vars) into logs outside the summary sink.
+set -euo pipefail
+cat >> "${GITHUB_STEP_SUMMARY:?GITHUB_STEP_SUMMARY unset}"
--- a/.github/scripts/sentry-api.sh
+++ b/.github/scripts/sentry-api.sh
@@ -8,6 +8,10 @@ set -euo pipefail
 
 path="${1:?usage: sentry-api.sh </api/0/...> [--with-headers]}"
 case "$path" in
+  *..*)
+    echo "sentry-api.sh: path traversal is not allowed" >&2
+    exit 2
+    ;;
   /api/0/*) ;;
   *)
     echo "sentry-api.sh: path must start with /api/0/" >&2
--- a/.github/scripts/sentry-api.sh
+++ b/.github/scripts/sentry-api.sh
@@ -1,0 +1,28 @@
+#!/usr/bin/env bash
+# Read-only proxy for the Sentry API used by sentry-triage.yml.
+# The triage agent is only allowlisted to run this script, never bare curl,
+# so it cannot reach any other host and never handles the token in its own
+# command lines. GET only; the sole option is --with-headers, which dumps
+# response headers so the agent can follow Link-header pagination.
+set -euo pipefail
+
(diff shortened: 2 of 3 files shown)


marcodejongh:
Applied the review: `gh issue` scoped to create/comment/list/view, `python3` and `gh api` dropped, `stats_period` is now a choice input. Kept `id-token: write` deliberately — claude-code-action uses it for its token exchange, same as `claude.yml`; the comparable workflow the review cites doesn't run the action.

marcodejongh:
All three round-3 points are now enforced structurally rather than by prose: Sentry access goes through `.github/scripts/sentry-api.sh` (host-locked to us.sentry.io, path-validated, GET-only, token stays out of the agent's command lines — bare `curl` is gone from the allowlist), `dry_run: true` removes `gh issue create`/`comment` from the allowlist itself, and a shell guard validates `stats_period` for API-triggered dispatches.

marcodejongh:
Round-4 follow-ups: `echo` is gone from the allowlist — the report now goes through `.github/scripts/append-summary.sh`, which can only append stdin to `GITHUB_STEP_SUMMARY` — and `sentry-api.sh` rejects `..` in paths.

On the environment concern: `testflight-feedback-issues.yml` has run `environment: Production` on a `*/30` cron for months and fires on schedule, so Production has no required-reviewer gate that would queue scheduled runs; keeping the environment for its secret-protection consistency.

Residual risk note for the record: `jq` plus the scoped `gh` writes remain enough for a determined injection to exfiltrate data it can already read — the structural guardrails narrow the surface (no bare curl, no echo, host-locked Sentry access, issue-write only), they don't make prompt injection impossible. The prompt-level untrusted-data rule stays the last line.

marcodejongh:
Applied all three: `--connect-timeout 10 --max-time 60`, `--fail-with-body`, and `--with-headers` now writes to stderr so stdout stays clean JSON. Leaving `checkout@v6` as-is — it matches `claude.yml`, the closest sibling.

Issue #4104, opened 2026-08-01
expo-sqlite 'database is locked' persists after #3858 on iOS and Android, during sqlite-init


marcodejongh:
## Summary

Follow-up to #3858. `database is locked` is still firing on **both** platforms after that issue was closed, across every expo-sqlite entry point rather than just one.

| Sentry | Platform | Message | Events | Users |
|---|---|---|---|---|
| [BOARDSESH-A9](https://boardsesh.sentry.io/issues/BOARDSESH-A9) | iOS | `Calling the 'finalizeAsync' function has failed → Error code 5: database is locked` | 42 | 29 |
| [BOARDSESH-AB](https://boardsesh.sentry.io/issues/BOARDSESH-AB) | iOS | `Calling the 'execAsync' function has failed → database is locked` | 40 | 26 |
| [BOARDSESH-AX](https://boardsesh.sentry.io/issues/BOARDSESH-AX) | iOS | `Calling the 'prepareAsync' function has failed → Error code 5: database is locked` | 21 | 18 |
| [BOARDSESH-6V](https://boardsesh.sentry.io/issues/BOARDSESH-6V) | Android | `NativeDatabase.prepareAsync` / `NativeStatement.finalizeAsync` / `NativeDatabase.execAsync` → `database is locked` | 52 | 32 |

(The Android rows are buried inside an over-merged bucket — see #4101.)

Counts are 14d. Still present on the newest builds, including `com.boardsesh.app@2.3.0+2000783` (last seen 2026-07-31). Events carry `kind: sqlite-init` and `source: offline-sync`.

## Why this is a new issue rather than a reopen

#3858 was scoped to *"no WAL / busy_timeout on the main connection"* and to concurrent transactions during offline writes and local reads. What is left has a different signature:

- It spans `prepareAsync`, `execAsync` **and** `finalizeAsync` — including statement finalisation, which is teardown rather than a contended write.
- It is tagged `kind: sqlite-init`, i.e. it fires during database setup, not during steady-state sync traffic.
- It reproduces on iOS and Android equally, so it is not an Android-storage-pressure artifact.

`sqlite-init` failing suggests two connections are racing to open/migrate the same database — most likely the offline-sync engine and a snapshot ATTACH or a migration path opening their own handles concurrently, where a `busy_timeout` on one connection does not help because the other holds an exclusive lock for the duration of its transaction.

## Fix direction

[...]

Issue #3646, opened 2026-07-15
Offline mode: seed checkpoints are written under a key nobody reads


marcodejongh:
Found while building #3617 (PR #3630). Pre-existing, not introduced there.

`loadOptionalSeed` in `packages/mobile/src/db/connection.ts` stamps the bundled seed's cursors as:

```ts
setCheckpoint(txn, getCheckpointKey(cursor.table_name, cursor.board_type), …)
// → "checkpoint:board_climbs:kilter"
```

But the pull client keys checkpoints on the **scope**, not the board type (`pull-client.ts`):

```ts
getCheckpointKey(tableName, boardScope.scopeKey)
// → "checkpoint:board_climbs:kilter:1:5"
```

These never match, so every seed checkpoint is a dead key: a seeded scope re-crawls its whole catalog from epoch on first sync instead of resuming from the seed's build point — which is the entire reason the seed records cursors.

Low urgency: `resolveSeedAssetModuleId()` returns `null` in the committed default build, so no shipping build has a seed today. It bites whenever an opt-in seed profile is turned on.

**Fix sketch:** the seed would need to record cursors per `(boardType, layoutId, sizeId)` scope rather than per board type — a seed built for one board type can't know which scopes a user will enable, so this may want the same treatment the snapshot bootstrap gives it (compute the watermark from the exact scoped rows imported). Worth deciding alongside whether the bundled seed survives at all now that `board-snapshots.md` supersedes it.

`scope-teardown.ts` deliberately does **not** try to clean up the `checkpoint:<table>:<boardType>` form.

PR #4239, opened 2026-08-10
test(mobile): drive the sqlite init retries on fake timers

marcodejongh:
## Summary

Two tests in `packages/mobile/src/db/__tests__/connection.test.ts` crossed the SQLite init backoff with a real 750 ms sleep. The first retry delay is 500 ms, so only 250 ms of margin was left for `attemptInitialization` itself — thin enough to flake under CI load. `gives up after the bounded attempt window`, in the same describe, already used `vi.useFakeTimers()` + `vi.advanceTimersByTimeAsync`; the other two now do the same.

- `publishes the handle once a retry wins` and `shares one lifecycle across a remount mid-retry` install fake timers and advance by `INIT_RETRY_DELAYS_MS[0]`, read from the production backoff rather than mirrored as a literal, so a retuned backoff carries the tests with it.
- `does not block app launch on the retry` moves to the fake clock too. It deliberately walks away mid-chain, so on real timers it armed a retry that kept going for ~17.5 s after the test finished and ended by calling `reportError` — able to land inside a later test that asserts `reportErrorMock` call counts. On the fake clock, `afterEach`'s `useRealTimers()` discards it.

The review also flagged that `resetDatabaseInitializationForTests` is exported from a production module with only a JSDoc "Test-only" label guarding it. Calling it from the app would let a remount start a second retry chain against the same file — the #4104 contention the single-flight guard exists to prevent. The reset stays in `connection.ts` (that's where the guard's state lives), but:

- `src/db/testing.ts` is the sanctioned import path, and the test imports from there. The production barrel `src/db/index.ts` already omits the symbol.
- A re-export alone doesn't stop a careless caller, so `connection-test-seam.test.ts` enforces it: it walks every `.ts`/`.tsx` file under `src/` and `app/` (excluding tests, `connection.ts`, `testing.ts`) and fails if any names the symbol or resolves an import to `src/db/testing`. The specifier match covers static `from '…'`, bare `import '…'`, dynamic `import(…)`, and `require(…)`, in either quote style, with or without an explicit `.ts`/`.tsx` extension, and resolves `@/`-aliased paths. A "scans the mobile source tree" meta-test keeps the guard from silently becoming vacuous. This follows the source-level assertion pattern already in `connection.test.ts` for the #3646 seed retirement.

The only production change is exporting `INIT_RETRY_DELAYS_MS` (inert data, previously module-private) plus comments. No behaviour changes.

## Release Notes

## Test plan

- `vp run test:mobile` — 604 files, 5293 tests passed.
- `vp check packages/mobile/src/db` — 14 files, no format/lint/type findings.
- `vp run check:mobile-bundle` — iOS OK (14.7 MB), Android OK (14.8 MB).
- Verified the timer advance is load-bearing, not incidental microtask flushing: temporarily setting the advance to 100 ms fails both tests (handle still `null`); at the real 500 ms they pass. The db suites now run in ~900 ms instead of sleeping 1.5 s of real time.
- Verified the seam guard both ways: it passes on the current tree (926 files scanned, production barrel included) and flags every probe form it claims to cover — named import from `../db/connection`, relative and `@/`-aliased `import * as`, double-quoted specifier, explicit `.ts`/`.tsx` extension, `await import(…)`, and `require(…)`.
- Repo-wide `vp check` reports 297 findings in this sandbox, but the count is identical with and without this branch's changes (baseline measured with the work stashed) and none are in the files touched here.

---
_Generated by [Claude Code](https://claude.ai/code/session_01EUbmidMLskbkSwsVzTCSH5)_

### DIFF
--- a/packages/mobile/src/db/__tests__/connection-test-seam.test.ts
+++ b/packages/mobile/src/db/__tests__/connection-test-seam.test.ts
@@ -1,0 +1,52 @@
+// `resetDatabaseInitializationForTests` drops the single-flight guard that keeps one
+// database lifecycle per process. Calling it from the app would let a remount start a
+// second retry chain against the same file — the #4104 contention the guard exists to
+// prevent. A JSDoc "test-only" label doesn't stop that, so the boundary is checked
+// here: only the module that defines it, the ./testing barrel, and tests may reach it.
+
+import { describe, expect, it } from 'vitest';
+import { readdirSync, readFileSync } from 'node:fs';
+import { dirname, join, resolve, sep } from 'node:path';
+import { fileURLToPath } from 'node:url';
+
+const packageRoot = fileURLToPath(new URL('../../../', import.meta.url));
+const SCANNED_ROOTS = ['src', 'app'];
+const TESTING_BARREL = join(packageRoot, 'src', 'db', 'testing');
+const ALLOWED_FILES = [join('src', 'db', 'connection.ts'), join('src', 'db', 'testing.ts')];
+
+function listApplicationFiles(root: string): string[] {
+  return readdirSync(join(packageRoot, root), { recursive: true, encoding: 'utf8' })
+    .filter((entry) => /(?<!\.test)\.tsx?$/.test(entry))
+    .map((entry) => join(root, entry))
+    .filter((entry) => !entry.split(sep).includes('__tests__') && !ALLOWED_FILES.includes(entry));
+}
+
+/** Resolves a relative or `@/`-aliased specifier (see tsconfig paths) to an absolute path. */
+function resolveSpecifier(file: string, specifier: string): string {
+  if (specifier.startsWith('@/')) return join(packageRoot, 'src', specifier.slice(2));
+  if (specifier.startsWith('.')) return resolve(dirname(join(packageRoot, file)), specifier);
+  return specifier;
+}
+
+describe('database test seam', () => {
+  const applicationFiles = SCANNED_ROOTS.flatMap(listApplicationFiles);
+
+  it('scans the mobile source tree', () => {
+    // A walk that silently matched nothing would make the guard below vacuous.
+    expect(applicationFiles.length).toBeGreaterThan(100);
+  });
+
+  it('keeps the init-guard reset out of application code', () => {
+    const offenders = applicationFiles.filter((file) => {
+      const source = readFileSync(join(packageRoot, file), 'utf8');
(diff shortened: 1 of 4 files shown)


marcodejongh:
Addressed the review notes in `63cfac8`:

1. **Duplicated backoff constant** — fixed. `INIT_RETRY_DELAYS_MS` is now exported from `connection.ts` and the tests use `INIT_RETRY_DELAYS_MS[0]` instead of a mirrored `500`. A retuned backoff carries the tests with it.

2. **Dynamic `import()` / `require()`** — fixed. The seam guard's specifier pattern is now `/(?:from|import|require)\s*\(?\s*['"]([^'"]+)['"]/g`, covering static `from '…'`, bare `import '…'`, and both dynamic forms. Verified against probe files using `await import('../db/testing')` and `require('@/db/testing')` — both are flagged.

3. **No explicit `vi.useRealTimers()` per test** — skipping. The `afterEach` in that describe already existed before this PR and is what the pre-existing `gives up after the bounded attempt window` test relies on, so per-test teardown would be redundant with the block's own cleanup rather than more robust. Vitest's own guidance is to restore in `afterEach`; if that hook were ever removed, the right fix is to keep it, not to scatter three copies.

Also covered earlier in this PR: double-quoted specifiers (`4b04049`) and explicit `.ts`/`.tsx` extensions in the specifier (`4ab103a`), both verified with probe files.

Validation on this commit: `vp run test:mobile` 604 files / 5293 tests passed, `vp run typecheck:mobile` clean, `vp check packages/mobile/src/db` clean.

---

marcodejongh:
No change for either observation, so noting why rather than leaving them looking open:

- **`INIT_RETRY_DELAYS_MS` export** — this is the previous review round's point 1, which asked for exactly this ("Consider exporting the constant") to remove the mirrored `500` literal. I'd rather not ping-pong it. Of the two, exporting a plain readonly array of delays is the smaller leak than a hardcoded copy of the backoff drifting out of sync, and it's a very different hazard from the `resetDatabaseInitializationForTests` export this PR fences off — that one mutates the single-flight guard, this one is inert data.

- **Inline comment on the un-advanced `useFakeTimers()`** — already there, at `connection.test.ts:231-233`, three lines directly above the call:

  ```ts
  // This test deliberately walks away mid-chain, so the retry it leaves pending must
  // sit on the fake clock: afterEach's useRealTimers() discards it, instead of a real
  // timer firing into a later test and publishing a handle or reporting an error there.
  vi.useFakeTimers();
  ```

  So that context survives the merge without relying on the PR description.

---

PR #4247, opened 2026-08-10
fix(mobile): distinguish Record live-session badge from board-connected

### LINKED ISSUE #2563: [mobile] Disambiguate the Record live-session badge + add return-to-session
From the multi-reviewer pass on the queue FAB + search redesign (#2560).

Deleting the full-screen session overlay removed the unambiguous "you're in a session" signal and the one-gesture way back. The Record-tab badge is now **shared**:

```
showRecordBadge = isBluetoothConnected || sessionId !== null   // app/(tabs)/_layout.tsx
```

so a green dot means **either** a live session **or** just a connected board.

### Do
- Give the **live-session** state a distinct treatment from board-connected (within `NativeTabs.Trigger.Badge` constraints).
- Add a **"return to session"** affordance while a session is live (e.g. tap-through from the persistent climb capsule to the Record tab), since "minimize" is now just tab-switching.

### Where
`packages/mobile/app/(tabs)/_layout.tsx`, the persistent capsule.

Needs on-device verification (native badge rendering is limited).

marcodejongh:
## Summary

Deleting the full-screen session overlay in #2560 left the Record tab's badge doing double duty — the same green dot meant either "board connected" or "session live," with no way to distinguish them, and no one-tap way back to Record once you'd switched tabs.

- The Record tab badge now colors amber (`brandColors.live`) when a session is live, and stays green (`brandColors.success`) when only a board is connected — on both the iOS Liquid Glass `NativeTabs` badge and the Material `MaterialTabBar` dot (which shared the same boolean but was missed by the issue's own code snippet).
- A small "return to session" icon button appears inside the persistent floating climb capsule (Liquid Glass / fallback variant) while a session is live and you're on any tab other than Record; tapping it routes straight back to Record. The button is intentionally excluded from Record's own tab and from the Material variant's docked bar (its `endAction` slot is already the log-ascent tick).

Closes #2563

**Out of scope** (documented in the plan, flagged for reviewers):
- The iOS 26 native bottom accessory (`NativeAccessoryClimbRow`) — its tick/board-control slots are already fully reserved, and it sits in the same tab-bar chrome as Record already.
- The Material variant's docked `ClimbCapsule` bar — its `endAction` slot is already the log-ascent tick, and it docks directly above the Material tab bar (Record is one tap away).
- iPad shell — Record is already reachable via the persistent sidebar rail row.
- Any change to session start/end logic, BLE handling, or SessionScreen/InSessionView internals.

**Needs on-device verification**: native badge rendering (iOS `NativeTabs.Trigger.Badge` color contrast/legibility, and whether `selectedBackgroundColor` behaves the same for selected vs. unselected tab state) can't be confirmed from this Linux dev environment — the color-branch logic is unit-tested, but flagging per the issue's own note.

## Release Notes

- Record tab now shows a different badge color when a session is live vs. when a board is just connected, and a quick tap-back-to-session button shows up on your other tabs while you're mid-session.

## Test plan

- `vp check` — passes (pre-existing unrelated lint findings on `main` in `packages/db` / `packages/shared/offline-sync` confirmed via stash-diff, not touched by this PR).
- `vp test run --project mobile` — full mobile suite passes (602 files including the fixed `tab-layout.test.tsx`, `MaterialTabBar.test.tsx`, `persistent-queue-bar.test.tsx`, and new `return-to-session-button.test.tsx`).
- `vp run check:mobile-bundle` — passes.
- `vp test run packages/shared/i18n/src/__tests__/catalog-completeness.test.ts` — passes (all 4 locales carry the new `mobile.session.returnToSession` key).
- `vp run check:i18n` — passes.
- Manual QA plan written to `.boardsesh/qa-notes.md` (dev server not started, per no-dev-server instruction for this task).

### DIFF
--- a/packages/mobile/app/(tabs)/__tests__/tab-layout.test.tsx
+++ b/packages/mobile/app/(tabs)/__tests__/tab-layout.test.tsx
@@ -226,12 +226,14 @@ vi.mock('expo-router/unstable-native-tabs', () => {
       iconColor,
       labelStyle,
       tintColor,
+      badgeBackgroundColor,
     }: {
       children?: ReactNode;
       minimizeBehavior?: string;
       iconColor?: unknown;
       labelStyle?: unknown;
       tintColor?: unknown;
+      badgeBackgroundColor?: string;
     }) =>
       createElement(
         'nav',
@@ -241,6 +243,7 @@ vi.mock('expo-router/unstable-native-tabs', () => {
           'data-icon-color': JSON.stringify(iconColor),
           'data-label-style': JSON.stringify(labelStyle),
           'data-tint-color': typeof tintColor === 'string' ? tintColor : '',
+          'data-badge-background-color': badgeBackgroundColor ?? '',
         },
         children,
       ),
@@ -792,5 +795,27 @@ describe('TabLayout', () => {
 
     expect(badge).not.toBeNull();
     expect(badge?.getAttribute('data-selected-background-color')).toBe('#FBBF24');
+  });
+
+  // `NativeTabs.Trigger.Badge`'s `selectedBackgroundColor` only reaches
+  // options.selectedBadgeBackgroundColor, which expo-router applies solely to the
+  // ['selected','focused'] item states — so it's a no-op for the badge tint while
+  // Record is NOT the focused tab (the exact moment a live-session badge needs to
+  // be visible from another tab). The normal-state tint must ride the
+  // navigator-level `badgeBackgroundColor` prop instead.
+  it('sets the navigator-level badgeBackgroundColor to the live color so the unfocused badge tint is correct', () => {
+    cfg.sessionId = 'session-1';
+    const { container } = render(<TabLayout />);
+    const tabs = container.querySelector('[data-tabs="true"]');
+
+    expect(tabs?.getAttribute('data-badge-background-color')).toBe('#FBBF24');
+  });
+
+  it('sets the navigator-level badgeBackgroundColor to the standard (success) color when only Bluetooth is connected', () => {
+    cfg.bluetoothConnected = true;
+    const { container } = render(<TabLayout />);
+    const tabs = container.querySelector('[data-tabs="true"]');
+
+    expect(tabs?.getAttribute('data-badge-background-color')).toBe('#047857');
   });
 });
--- a/packages/mobile/app/(tabs)/__tests__/tab-layout.test.tsx
+++ b/packages/mobile/app/(tabs)/__tests__/tab-layout.test.tsx
@@ -258,6 +258,7 @@ vi.mock('expo-router/unstable-native-tabs', () => {
 });
 
 import TabLayout, { unstable_settings } from '../_layout';
+import { TAB_BADGE_CONNECTED, TAB_BADGE_LIVE } from '../../../src/components/navigation/tab-badge';
 
 describe('TabLayout', () => {
   beforeEach(() => {
@@ -697,25 +698,30 @@ describe('TabLayout', () => {
     expect(cfg.materialScreens.find((screen) => screen.name === 'record')?.options).toMatchObject({ lazy: false });
   });
 
-  it('sets the JS-Tabs Record badge to the "live" sentinel when a session is active', () => {
+  // The JS-Tabs badge value is a state marker MaterialTabBar branches its dot
+  // color on — assert the shared constants (tab-badge) so producer and consumer
+  // can't drift apart, and pin the wire values so a rename stays deliberate.
+  it('sets the JS-Tabs Record badge to the live marker when a session is active', () => {
     cfg.variant = 'material';
     cfg.sessionId = 'session-1';
 
     render(<TabLayout />);
 
+    expect(TAB_BADGE_LIVE).toBe('live');
     expect(cfg.materialScreens.find((screen) => screen.name === 'record')?.options).toMatchObject({
-      tabBarBadge: 'live',
+      tabBarBadge: TAB_BADGE_LIVE,
     });
   });
 
-  it('sets the JS-Tabs Record badge to "connected" when only Bluetooth is connected', () => {
+  it('sets the JS-Tabs Record badge to the connected marker when only Bluetooth is connected', () => {
     cfg.variant = 'material';
     cfg.bluetoothConnected = true;
 
     render(<TabLayout />);
 
+    expect(TAB_BADGE_CONNECTED).toBe('connected');
     expect(cfg.materialScreens.find((screen) => screen.name === 'record')?.options).toMatchObject({
-      tabBarBadge: 'connected',
+      tabBarBadge: TAB_BADGE_CONNECTED,
     });
   });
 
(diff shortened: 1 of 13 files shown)


marcodejongh:
Addressed the QA finding on the iOS badge color: `NativeTabs.Trigger.Badge`'s `selectedBackgroundColor` only reaches `options.selectedBadgeBackgroundColor`, which expo-router 57.0.9 applies to the `['selected','focused']` item states only (`appearance.ios.js` `appendSelectedStyleToAppearance`) — confirmed against the installed source. That left the badge color unchanged for the normal (unfocused) state, exactly when a user is browsing another tab while a session is live.

Fix (3e1a15d3a): pass `badgeBackgroundColor={hasLiveSession ? brandColors.live : brandColors.success}` on the `<NativeTabs>` navigator itself. That flows into `options.badgeBackgroundColor`, which is applied to normal/focused/selected alike (`createStandardAppearanceFromOptions`). Safe as a navigator-wide default — Record is the only trigger that renders a badge. Added two tests asserting the navigator-level prop value for both the live and connected-only states.

marcodejongh:
Worked through the review. Both flagged items are real; fixed both, plus the minor. Evidence per finding.

---

### 1. Sentinel string coupling in `tabBarBadge` — real, fixed (96c2b7a)

Confirmed the coupling was undocumented in the type system: `_layout.tsx` emitted bare `'live'` / `'connected'` and `MaterialTabBar.tsx` compared `options.tabBarBadge === 'live'`, so the literal existed in four places (producer, consumer, two tests) with nothing tying them together.

Fixed by giving the contract a home: new `packages/mobile/src/components/navigation/tab-badge.ts` exports `TAB_BADGE_LIVE`, `TAB_BADGE_CONNECTED`, the `TabBadgeState` union, and `isLiveTabBadge()` — now the only place the live value is compared. Producer (the Record `Tabs.Screen`) and consumer (`MaterialTabBar`) both import it, and the tests assert the constants while pinning the wire values so a rename stays deliberate rather than silent.

I did **not** take the `tabBarBadgeVariant` custom-key route. `Tabs.Screen`'s options are React Navigation's `BottomTabNavigationOptions` (vendored through expo-router), a closed type — an extra key needs either module augmentation of a vendored declaration or an `as` cast at the call site. The cast throws away exactly the type safety the key was for, and the augmentation is a global edit for one screen's flag.

On the failure mode itself, it's narrower than it looks: any badge value that isn't the live marker already falls through to the standard green, and `MaterialTabBar` renders **no badge text at all** (the badge is a bare dot, `MaterialTabBar.tsx:91-109`), so a future count badge degrades to a plain dot with or without this contract. Added a test for that case (`tabBarBadge: 3` → standard color).

### 2. Navigator-wide `badgeBackgroundColor` bleed — real, fixed with a comment **and** a test (96c2b7a)

Correct that the prop is a navigator-wide default. Rather than leave it on a comment alone, the invariant is now locked by a test — `keeps Record as the only badged trigger (the navigator badge color is a shared default)` in `packages/mobile/app/(tabs)/__tests__/tab-layout.test.tsx`: it asserts exactly one `NativeTabs.Trigger.Badge` renders across the whole navigator and that it belongs to the `record` trigger. A future PR adding a badge to another tab fails there, and the comment at the prop now says what to do about it (drop the navigator prop, move the session color to per-trigger styling) and points at the test.

### 3. Minor — `persistent-queue-bar.test.tsx` hardcoded `'44'` — real, fixed (09261c2)

Real, with one correction to the pointer: the file is 378 lines and the assertion is at **line 349**, not 430.

`glassSize` is fully mocked in that file (`vi.mock('../../../theme/layout', …)`, line 92: `glassSize: { hero: 64, inline: 44 }`), so importing the real constant isn't possible — any import of that path inside the test resolves to the mock. Tied the assertion to the mock instead: `toBe(String(glassSize.inline))`, with a comment saying so. Retuning the ladder is now a one-line mock edit rather than an opaque `'44' !== '48'` failure.

Noted while there, not changed: that same mock has `hero: 64` while the real `glassSize.hero` is `56` (`packages/mobile/src/theme/layout.ts:44`). Pre-existing and not load-bearing — the geometry assertions in that file check arithmetic built from the same mocked constants — so out of scope for this PR.

### Also fixed while in these files (carry-over from the previous round, 96c2b7a)

The earlier review flagged that `MaterialTabBar.test.tsx` mocked the static brand as `live: '#FBBF24'`. Verified against the source: the real static values are `live: '#B45309'`, `success: '#047857'` (`packages/shared/velvet-tokens/src/index.ts:38-49`), and `#FBBF24` is the dark-scheme value. The production choice stands — the badge dot is a **fill**, and fills don't scheme-lift in Velvet Send — but the mock was lying about what ships. The static mock now carries the real light values, the theme mock carries the dark ones, and the badge assertions therefore prove the dot reads the static fill rather than the theme.

---

**Validation** (`vp` only): `vp run typecheck:mobile` clean; `vp run test:mobile` — 598 files / 5216 tests passed; `vp run check:mobile-bundle` green; pre-commit `vp check --fix` passed on every touched file. A full-repo `vp check` reports 28 errors, all pre-existing — the identical count comes back on the untouched branch tip and none are in the files this PR touches.

marcodejongh:
Follow-up on the newer review (the ✅ pass that landed while I was pushing) — three points, no code changes.

**"Multi-line comment blocks violate CLAUDE.md's 'no multi-line comment blocks' rule" — no such rule.** `grep -niE "multi-?line|comment block|one short line" CLAUDE.md AGENTS.md .github/*.md` returns exactly one hit, and it's in `.github/pull_request_template.md:14`: *"One short line per user-facing change"* — an instruction for the **Release Notes** section of a PR body, not for code comments. Nothing in CLAUDE.md constrains comment length. The convention in this area runs the other way: the `NativeTabs` `badgeBackgroundColor` block in `_layout.tsx` is 14 lines, the `MaterialTabBar` module doc is 8, `glassSize` in `theme/layout.ts` carries a ~20-line ladder table. Leaving both blocks as-is.

**`badgeBackgroundColor` set unconditionally** — deliberate, keeping it. The prop tints a badge; with no badge mounted there is nothing to tint, so the "unused" value has no effect. Passing it unconditionally means the appearance is already correct at the instant `showRecordBadge` flips true, instead of the badge mounting and the tint arriving in a second native appearance update — one fewer chance of a frame of default-tinted badge at exactly the moment a session goes live. Gating it on `showRecordBadge` would buy nothing and add that ordering risk.

**Mock/assertion color duplication** — intentional, as noted. The assertions pin the literal on purpose: a mock edited without its assertion is precisely the drift the previous round caught (the static-brand mock claiming `#FBBF24` when production ships `#B45309`). Reading the constant back out of the mock would make that class of bug invisible again.

CI is green on `09261c2` (27 success, 10 skipped, 0 failures).

marcodejongh:
@marcodejongh — visual sign-off needed before merge (standing rule: visual/rendering changes are never auto-merged).

This PR changes the Record-tab badge treatment: live-session state gets the amber `live` fill vs the green `connected` dot, on both the iOS NativeTabs badge and the Material tab-bar dot. No before/after screenshots attached — iOS capture needs macOS and the local Android emulator path is unavailable on this box. If you want A/B evidence instead of an on-device look, say the word and I'll dispatch `mobile-screenshots-android.yml` on this branch + main for a side-by-side.

CI is green and review findings are addressed; holding at ready-for-review for your visual check.

PR #4249, opened 2026-08-10
fix: draw logged climbs on the board they were climbed on

### LINKED ISSUE #4221: 🐞 Bug report: More of a suggestion, but I don’t really get why the log book (for example here on the kilter) always...

## Comment

More of a suggestion, but I don’t really get why the log book (for example here on the kilter) always shows climbs on the largest kilter board layout rather than the board they were climbed on. I would rather see the board it was climbed on.

## Metadata

| Field | Value |
| --- | --- |
| Platform | ios |
| App version | 2.3.1 (2) |
| Source | drawer-bug |
| Board | kilter / layout 1 / size 10 / sets [1,20] @ 50° |
| Climb | Greg let me do it (6c/V5) |
| Session | unknown |
| URL | /profile |
| User agent | unknown |
| Feedback row | app_feedback #10199 |

## Contact

✅ Reporter agreed to follow-up — resolve **app_feedback #10199** in the DB to reach them.

marcodejongh:
## Summary

Closes #4221.

A climber on a 10x12 Kilter reported that their logbook drew every ascent on the
12x14. The information to do better was already there and being thrown away:

- `boardsesh_ticks.board_id` → `user_boards` carries the exact `layout_id` /
  `size_id` / `set_ids`, and `userAscentsFeed` **already LEFT JOINs that row** —
  it only selected the board's name.
- `board_climbs.compatible_size_ids` / `required_set_ids` already answer "does
  this climb fit on this board", denormalised and GIN-indexed.

So both clients invented a config, and they didn't even agree: mobile picked the
largest-**area** size, web used a hardcoded 10-entry table that returned `null`
(→ no thumbnail at all) for Decoy, Touchstone, Grasshopper and the orphaned
Kilter layouts 2-7.

### The ladder

Ascents now carry a `renderBoard`, resolved by one shared function
(`resolveRenderBoard` in `@boardsesh/board-config`):

1. **The board the tick was logged against.** No inference needed. Skipped when
   its layout differs from the climb's — those holds wouldn't line up.
2. **The smallest board the climber has that the climb fits.** A climb that fits
   a small wall fits every bigger one, so the smallest fitting board is the
   tightest crop, and for home-wall climbers usually the real one. Owned boards
   beat followed ones.
3. **The size closest to their biggest board of that type**, out of the sizes
   the climb does fit — for a climb none of their walls can take.
4. **The layout's biggest size.** Today's behaviour, so a climber with no
   registered board sees no change.

Resolution happens **server-side, against the ascent owner's boards**, not the
viewer's — a viewer only knows their own walls, and drawing someone else's
ascents on yours would be a different wrong answer.

### Surfaces

`userAscentsFeed`, `userGroupedAscentsFeed`, `userClimbs`, session detail ticks,
the session feed's hardest send, and the record recap hero. On the clients that
covers the web logbook feed and grouped ascents feed, and mobile's Logbook tab,
Climbs tab, session rows, in-session history and play drawer.

Social feeds (following / global / new-climb / similar climbs / proposals) keep
today's fallback — same per-owner lookup, clean follow-up.

### Two deliberate rendering changes

Rung 4 replaces both old fallbacks, so the two platforms now agree on the guess:

| | before | after |
|---|---|---|
| Kilter layout 1 | mobile 16x12 Super Wide / web 12x14 | 12x14 Commercial |
| Tension Board 2 | mobile 12x16 / web 12x12 | 12x16 |

"Biggest" is one shared `getSizeRank` (tallest, widest breaking ties) — which is
what web's hand-curated table already meant, and which drops the fourth ad-hoc
size comparator in the repo.

### Privacy

`boardId` / `boardDisplayName` keep their existing `canShowBoard` gate unchanged.
`renderBoard` is wall geometry, not board identity — it never reveals that a
named board exists, and gating it would send other viewers straight back to the
wrong-size render. Covered by a test.

## Release Notes

- Your logbook now draws each climb on the board you actually climbed it on, instead of the biggest board in the layout.
- Climbs on Kilter Orbit, Spire, JUUL, BKB, Decoy, Touchstone and Grasshopper walls finally show a thumbnail on the web.

## Test plan

- `vp run typecheck` and `vp check` clean.
- `vp run test:mobile` — 5306 passed.
- `vp run test:web` — 5892 passed.
- `vp test run --project backend` — green.
- New `resolve-render-board.test.ts` (16 cases): one per rung, plus the
  layout-mismatch skip, the missing-hold-set skip, cross-board-type boards,
  unknown fit columns, MoonBoard, and the orphaned Kilter layouts.
- New integration cases in `tick-queries.test.ts`: a tick on a registered 12x12
  renders on the 12x12 even when the climber also owns a 12x14; an unassociated
  tick falls to the smallest fitting board; a climb that fits none of their
  boards lands on the closest-to-biggest size; a climber with no boards is
  unchanged; a private board keeps its name hidden but still resolves geometry.
- Dev server + `.boardsesh/qa-notes.md` for the manual sweep.

### Checked against the reporter's real data (read-only prod query)

The climber who filed #4221 doesn't *own* a board record — they follow two
12x12 Kilters (Original layout, sets 1+20). `fetchOwnerBoards` covers followed
boards as well as owned ones, so rung 2 catches their unassociated ticks:

```
owned ∪ followed: id=109 kilter layout=1 size=10, id=245 kilter layout=1 size=10
1000 kilter ticks
before: size 7 (12x14 Commercial) × 1000
after:  size 10 (12x12 with kickboard) × 1000
```

The climb named in the report, "Greg let me do it", carries `board_id=109` and
resolves on rung 1: `{layoutId: 1, sizeId: 7}` → `{layoutId: 1, sizeId: 10}`.

https://claude.ai/code/session_01EJLusqfY131Bd1w1h1o144


marcodejongh:
Thanks — took the first two.

- **Misleading test name**: fair hit, the case was named for a same-size tie-break it never exercised. Renamed it to what it actually asserts (ownership outranks smallest-fits) and added the real same-size case in 9bd2fe9.
- **`parseBoardSetIds` / `toTickBoardCandidate` untested**: added `resolvers/shared/__tests__/render-board.test.ts` covering whitespace and trailing commas, non-positive and fractional ids, bigint columns arriving as strings, and both null paths (LEFT JOIN produced nothing / board with no usable sets).

- **Redundant `fetchOwnerBoards` in `session-feed.ts`**: leaving it, with reasoning. `fetchTickHighlightsByUuid` returns early on an empty uuid list (`session-feed.ts:1248`), and `dailyHighlightTickUuids` is only non-empty when the caller asked for `includeDailyHighlights` — so the doubled fetch is not on every feed page, only on the ones that requested daily highlights. Merging the two would mean either a DataLoader-style in-flight batcher (the calls are concurrent inside one `Promise.all`, so a plain memo wouldn't merge them) or splitting both batch fetchers into fetch-rows + resolve-later and moving the mapping to the call site — `fetchTickHighlightsByUuid` has a second caller at :1465 that would need its own variant. That's real churn in a hot feed path to save two indexed lookups on a subset of requests. Happy to do it if you'd rather have it now.

marcodejongh:
Took two of these; the rest I'm pushing back on, with reasons.

**Fixed in 8c97852**
- *Duplicate `RenderBoardConfig`* — real, but not fixable in the direction suggested: `shared-schema` is the base package (no `@boardsesh/*` deps) and `board-config` depends on it, so re-exporting from `board-config` into `shared-schema` would be a cycle. Did it the other way — the GraphQL wire type in `shared-schema` is now the single definition and `resolve-render-board.ts` re-exports it.
- *Ownership tie-break ordering* — right, only the accumulator-keeps branch was covered. Added the reversed order.

**Not acting — factually wrong**
- *`setterClimbsFull` "all boards" path missing `resolveRenderBoard`.* There is exactly one `resolveRenderBoard` call in `setter-follows.ts` (line 649) and it is inside `userClimbs`, not `setterClimbsFull` — neither branch of `setterClimbsFull` calls it, so there's no if/else inconsistency to fix. Separately, `setterClimbsFull` feeds `SetterClimbList` → `MultiboardClimbList`, which resolves boards through `use-board-details-map` (`resolveBoardDetailsForClimb`) and gives the viewer an explicit board picker — a different and better signal than `renderBoard` would supply.

**Not acting — already decided**
- *Social/following feed.* Deliberately out of scope, agreed with the maintainer before implementation: the choice was logbook + profile + sessions over "everything incl. social feeds". The per-owner lookup generalises cleanly (`fetchOwnerBoards` already takes a user list and session detail already uses it that way), so it's a contained follow-up rather than something this PR is missing.
- *Rung 1 skips the `compatibleSizeIds` check.* Intentional, and the explicit requirement: if a tick says which board it was climbed on, that board wins. `compatible_size_ids` is derived from edge boxes (`populate-denormalized-columns.ts`), so on a disagreement the recorded board is the ground truth and the derived column is the suspect one. Drawing a real ascent slightly clipped beats relocating it to a wall the climber didn't use. Rung 1 does still guard on layout, where a mismatch means holds that genuinely don't exist.

**Not acting — cost/benefit**
- *`fetchOwnerBoards` has no `boardType` filter.* A filter means every call site has to compute the right board-type set for its page; get that wrong and you silently drop a board and render the wrong size. Two indexed lookups against a per-user set that's single digits for almost everyone isn't worth that failure mode.
- *Session-summary `fetchOwnerBoards` is serial.* It needs `hardestRows[0].userId`, so it can't join the `Promise.all` above it without a second query for the same information. It's also skipped entirely when a session has no qualifying send.

Issue #3122, opened 2026-06-21
Board art sometimes fails to draw in the web/webview renderer


marcodejongh:
## What's happening

The in-browser board renderer (Web Worker + WASM OffscreenCanvas) occasionally fails to draw the board. Most failures are board thumbnails in scrolling grids (low impact — a missing thumbnail), but a small slice are the full main board failing to render. The WASM path has a fallback to the SSR image tier, so most renders recover, but a cohort keeps hitting it.

## Evidence

PostHog project 412845 — product analytics (30d).

- `Board Render Error` fired 566 times across 113 distinct production users over 30d, all `$lib=js` (web/webview renderer). Cross-platform: iOS Mobile Safari 46 users, Android Chrome 37, Mac Safari 21, plus Windows/Linux. Steady across all 30 days (not a spike).
- Real prod signal, not noise: `track()` is hostname-gated to boardsesh.com prod and skips admin pages, so no dev/screenshot leakage. Rate-limited to 5 events/session, so the true failure rate is **under**-reported (67 users hit the cap of 5; one user 36 across multiple sessions).
- Impact is mostly thumbnails: 520/566 events (92%) are `thumbnail` context; only 11 events / 7 users are full-board (the genuinely blocking case).
- The session-wide worker-disable self-heal is NOT the driver — only 8 users fired `Board Worker Rendering Disabled` vs 113 render errors. These are per-render promise rejections with a live worker pool, recovering individually via the SSR image tier.
- Platform: web (also fires inside the iOS/Android webview). No `_posthogUrl` captured for this analytics event.

## Suspected cause

`packages/web/app/components/board-renderer/board-canvas-renderer.tsx:80` — the Worker + WASM `renderBoard` promise rejects (per-render `setFailed` → `BoardImageLayers` fallback at lines 95-106). The last-tier `<img onError>` in `board-image-layers.tsx:76` is the case where the board genuinely fails to draw. Both sites report `renderer:'wasm'` via `trackRenderError` (`packages/web/app/lib/rendering-metrics.ts`), so the instrumentation can't distinguish a worker-render rejection from a last-tier image error, and no error message is captured. First step: split the instrumentation (worker-render rejection vs image `onError`, capture the error message), then address the iOS/Android WASM render failures. Also see `packages/web/app/lib/board-render-worker/worker-manager.ts`.

## Severity

[...]

Issue #2651, opened 2026-06-08
Revive: Multi-board queue support with cross-board confirmation dialog (was #1633)


marcodejongh:
## Intent

Revive the multi-board queue feature from #1633 (branch `claude/multi-board-queue-support-uIwDF`, opened 2026-04-20).

The PR added support for queuing climbs from different boards / board sizes in a single session. When a climb from a different board (or a larger size) is added, a confirmation dialog offers three choices: **Add to current queue**, **Switch to that board**, or **Cancel**.

Core pieces it introduced:
- `packages/web/app/lib/board-config.ts` — `BoardConfig` type, `configKey()`, `decideAdd()`, `deriveAcceptedConfigs()`.
- `QueueAddConfirmProvider` + `useQueueAddConfirm()` + `confirm-add-climb-dialog.tsx` — serialized one-at-a-time confirmation dialog.
- `boardConfig` / `boardId` on `ClimbQueueItem`; `boardType` / `layoutId` on `Climb` for round-tripping through the queue (shared-schema + backend + db climb queries).
- Bridge + GraphQL queue providers stamp `boardConfig` and gate cross-board adds; `BluetoothAutoSender` warns on board/layout mismatch.
- New Redis key `boardsesh:session:<id>:boards`; sessions can carry multiple boards.

## Why still wanted

Not shipped and not superseded. As of 2026-06-08, none of the feature's symbols exist on `main` (`decideAdd`, `deriveAcceptedConfigs`, `QueueAddConfirmProvider`, `useQueueAddConfirm`, `board-config.ts`, `confirm-add-climb-dialog.tsx`, `queue-add-confirm-context.tsx`), and #1633 is the only open PR touching multi-board queues. Cross-board queuing is still a real gap.

## Why the rebase was abandoned (mechanical scope only)

The branch is **2438 commits behind main** (Apr 20 → Jun 8). `git rebase origin/main` failed on the very first of 4 commits with **36 conflict hunks across 11 files**, concentrated in the files that were heavily rewritten over the gap:
- `queue-control/queue-bridge-context.tsx` — 11 hunks
- `graphql-queue/QueueContext.tsx` — 7 hunks
- `board-bluetooth-control/bluetooth-context.tsx` — 7 hunks

[...]

PR #1633, opened 2026-04-20
Multi-board queue support with cross-board confirmation dialog

marcodejongh:
## Summary

This PR adds support for multi-board climbing queues, allowing users to queue climbs from different boards and board sizes. When a climb from a different board or a larger board size is added, a confirmation dialog prompts the user to either add it to the current queue or switch boards entirely.

## Key Changes

- **New board config utilities** (`packages/web/app/lib/board-config.ts`):
  - `BoardConfig` type representing a complete board configuration (board name, layout, size, sets, angle)
  - `configKey()` to generate stable identifiers for board configurations
  - `decideAdd()` logic to determine if a climb can be added silently or needs confirmation
  - `deriveAcceptedConfigs()` to extract accepted board configs from an existing queue
  - Conversion functions between `BoardConfig`, `BoardDetails`, and `UserBoard`

- **Confirmation dialog system** (`packages/web/app/components/queue-control/queue-add-confirm-context.tsx` and `confirm-add-climb-dialog.tsx`):
  - `QueueAddConfirmProvider` context that serializes confirmation requests (one dialog at a time)
  - `useQueueAddConfirm()` hook for accessing the confirmation API
  - Dialog offers three choices: "Add to current queue", "Switch to that board", or "Cancel"
  - Distinguishes between two confirmation reasons: new board/layout vs. larger board size

- **Queue item schema updates** (`packages/shared-schema`):
  - Added `BoardConfig` and `BoardConfigInput` types to GraphQL schema
  - Extended `ClimbQueueItem` with optional `boardConfig` and `boardId` fields
  - Updated `Climb` type to include optional `boardType` and `layoutId` for round-tripping through queue

- **Integration with existing queue systems**:
  - Updated `QueueBridgeContext` to stamp `boardConfig` on queue items and gate adds through the confirmation dialog
  - Updated `GraphQLQueueProvider` to include board config on all queue items and handle cross-board adds
  - Modified `BluetoothAutoSender` to detect board/layout mismatches and surface snackbar warnings
  - Simplified `useQueueAddValidator` to allow all adds (validation now happens in confirmation gate)

- **Database and API updates**:
  - Extended climb queries to include `boardType` and `layoutId` fields
  - Updated `ClimbRow` type to include board identity information

## Notable Implementation Details

- **Backward compatibility**: Queue items without `boardConfig` are normalized on ingress using the session's primary board as a fallback
[...]

PR #4250, opened 2026-08-10
feat(mobile): ask before queueing a climb from another board

### LINKED ISSUE #3994: Queue climbs from more than one board in a session, with a cross-board confirm (respec of #2651)
## What we want

Let a climber queue climbs from more than one board (or more than one board size) in a single session. When they add a climb that doesn't match the session's current board config, offer three choices: **add it to the current queue anyway**, **switch to that board**, or **cancel**.

This respecs #2651, which tracked reviving PR #1633's branch. That branch is now **5607 commits** behind `main` and is not revivable — closing it out and starting from the current tree instead. The feature itself is still a real gap: none of the branch's symbols exist on `main` (`decideAdd`, `deriveAcceptedConfigs`, `QueueAddConfirmProvider`, `useQueueAddConfirm`, `queue-add-confirm-context`, `confirm-add-climb-dialog` — all 0 hits), and two of the files it edited (`packages/shared-schema/src/operations.ts`, `packages/web/app/lib/graphql/operations/climb-search.ts`) no longer exist.

## Where it belongs now

The original PR was entirely `packages/web`. Under the current architecture this is mobile-first with the decision logic shared:

- **Decision logic → `packages/shared/queue`** (pure TS): the `BoardConfig` shape, `configKey()`, `decideAdd()`, `deriveAcceptedConfigs()`. No React, no platform I/O.
- **Confirmation orchestration → `@boardsesh/queue-react`** if web and mobile genuinely share the serialize-one-dialog-at-a-time state machine; inject the actual prompt (MUI dialog on web, sheet on mobile) as a parameter.
- **Per-platform UI**: the mobile sheet and the web dialog.

Some of the groundwork already landed independently and should be reused rather than rebuilt: `classifyClimbBoardCompatibility` in `packages/shared/board-config/src/board-compatibility.ts` (from #3193) already classifies a queue item against the connected board on **both** platforms' BLE senders, and `boardType`/`layoutId` already round-trip on `ClimbInput` through the queue. So a mixed-board queue is already *tolerated* end-to-end — the wall skips a mismatched climb instead of dark-firing. What's missing is the deliberate, user-facing act of queuing cross-board and the board-switch affordance.

## Reviewer defects to fix from the start

These were flagged on PR #1633 and must not be reintroduced:

1. **Provider ordering.** `QueueAddConfirmProvider` was nested *inside* `QueueBridgeProvider`, so `useQueueAddConfirm()` resolved to `null` on the bridge path and the dialog never fired for local (non-party) sessions. The confirm provider goes **above** the queue provider.
2. **Race in `requestConfirm`.** Two rapid adds both read `activeRequest === null` from a stale closure, both set it, and the first resolver is dropped — a hang/leak on bulk add. Track "active" in a ref, not state.
3. **Party-mode `switch` outcome** only dispatched locally, diverging server and peer queue state. It must mirror the remove-then-add the synced path does.
4. **Stale queue snapshot** after the async gate: append against the live ref, not the pre-await snapshot.
5. **Trademark.** `formatBoardLabel` produced "Moonboard". It is **MoonBoard** — use a canonical-name lookup, never `charAt(0).toUpperCase()`.
6. **Angle fallback of `0`** is invalid; fall back to the board's default angle.
7. **Dead code.** The branch reduced `useQueueAddValidator` to `() => true` with dead call sites, plus a stale comment in the bridge about imports it had actually removed. Note that `useQueueAddValidator` is **live** on `main` today (`packages/web/app/components/board-lock/use-queue-add-validator.ts:21`, used at `graphql-queue/QueueContext.tsx:459`) — so this one needs re-deciding against the current hook rather than blindly re-applying: don't gut a hook that now has a real board-lock job.
8. **Design tokens.** `start-sesh-drawer.tsx` used a raw `var(--neutral-600)`. Use theme tokens.
9. **E2E** `multi-board-queue.spec.ts` hardcoded seed slugs/IDs. Note the e2e workflow is no longer per-PR (#2630), so don't lean on it as the only coverage — unit-test `decideAdd`/`deriveAcceptedConfigs` directly.

## Original context

Branch `claude/multi-board-queue-support-uIwDF` is preserved for reference only. Original PR: #1633. Previous revival tracker: #2651.

marcodejongh:
## Summary

Queueing a climb that belongs to a **different board than your queue is on** now
asks first: **Add anyway / Switch to \<Board\> / Cancel**. Until now a foreign-board
climb landed in the queue silently — tolerated end to end (the BLE sender already
skips a mismatched climb rather than dark-firing the wall), but never a deliberate
choice.

Closes #3994. Respec of #2651 / PR #1633, which was 100% `packages/web`; under the
current architecture the decision logic is shared and the UI is mobile.

### What landed where

- **`@boardsesh/queue` (pure TS)** — `configKey` / `climbConfigKey` /
  `deriveAcceptedConfigs` / `decideAdd`. The compatibility classifier is
  **injected**, not imported: the package has `dependencies: {}` on purpose (the
  backend imports it), so it must not grow an edge onto `@boardsesh/board-config`.
  Same pattern as the existing `isClimbable` predicate in `playlist-suggestions`.
- **`dialog-provider`** — `useChoose()` for two or three named actions;
  `useConfirm()` is now a thin wrapper over it, so the queueing / settle-once /
  per-variant rendering has one implementation.
- **`use-cross-board-add-gate`** (new, mobile) — the prompt, the one-prompt-per-board
  dedup, and the switch side effect.
- **`addToQueue`** — the single seam every add path already goes through.

### Reviewer defects from #1633, closed up front

1. **Provider ordering** — `useChoose()` *throws* without `DialogProvider` (which
   already mounts above `QueueProvider` in `app/_layout.tsx`). No nullable context
   that silently resolves to `null` on the local-session path.
2. **Race on bulk add** — adds are keyed by board config in an in-flight ref while
   the dialog is open: eight climbs from the same foreign board raise **one** dialog
   and all eight settle on its answer. No resolver is dropped.
3. **Party-mode `switch` was local-only** — it now runs the same synced add body
   (`dispatch` + `addQueueItem` + reconcile-on-failure) *and* broadcasts the new
   `setSessionBoardPath`, so peers follow instead of diverging.
4. **Stale queue snapshot after the await** — everything after the prompt re-reads
   `stateRef.current` and re-runs `attributeNewItem`. Test: an unrelated add made
   while the dialog is open survives.
5. **Trademark** — copy uses `formatBoardDisplayName` ("MoonBoard", "So iLL"), never
   `charAt(0).toUpperCase()`. Also fixed at its two live sites on `main`
   (`board-lock/board-switch-confirm-provider.tsx`, `board-lock/queue-add-error-messages.ts`),
   which render "Moonboard" today.
6. **Angle fallback of `0`** — switching activates the owned board **as stored**, at
   its own angle. Test pins a board whose angle (25°) differs from the active one (40°).
7. **`useQueueAddValidator`** — deliberately left intact. It has a live board-lock job
   at `graphql-queue/QueueContext.tsx:459` and its own suite; re-decided against the
   current hook rather than blindly re-applying #1633's gutting.
8. **Design tokens in `start-sesh-drawer`** — already fixed on `main`, nothing to do.
9. **E2E** — no spec. The e2e workflow is no longer per-PR (#2630); coverage is unit
   tests on `decideAdd` / `deriveAcceptedConfigs` plus a QueueProvider harness.

### Also fixed along the way

`DialogProvider` read its pending id **inside** the `setState` updater, which React
defers to render time — three prompts raised in one tick all took the last id, so
settling one dropped the others and left their callers awaiting forever. The id is
snapshotted at call time now, with a three-concurrent-prompts regression test.

### Deliberately out of scope

- **No web cross-board UI.** The web climbing surface is deprecated (#3122). If web
  parity is wanted it's a second PR consuming the same `@boardsesh/queue` primitives.
- **No `@boardsesh/queue-react` package.** With one consumer, an injected-prompt React
  package is speculative; extract it the day web needs it.
- **Cross-size gating.** A `Climb` carries no size, so a size mismatch can't be seen
  from queue metadata; `canAddClimbToBoard`'s hold-ID containment already catches it
  at send time. `configKey` is board + layout only.
- **`setCurrentClimb` / `setQueue` / `appendGeneratedSession`** — only `addToQueue` is
  gated. The play drawer already has a board-mismatch overlay with a switch affordance.
- No DB migration, no schema change, no Bluetooth code.

### Known trade-off

"Add anyway" is remembered by **queue contents**, not by a separate flag — remove the
last climb from that board and the next add prompts again. That's the price of having
no acceptance state that can go stale against a party-synced queue; documented in the
module header.

## Release Notes

- Queue climbs from more than one board in the same session. Tap a climb that's set on
  another board and you get the choice: add it anyway, switch over to that board, or
  back out.

## Test plan

- `vp run typecheck` — pass (full monorepo).
- `vp run test:mobile` — 602 files / 5278 tests, all pass.
- `vp test run --project queue --project i18n` — 221 tests, pass (includes the new
  `cross-board.test.ts` and the four-locale catalog-parity gate).
- `vp test run --project web board-lock` — 20 tests, pass.
- `vp run check:i18n` + `vp run check:i18n:orphans` — pass.
- `vp check` — pass on every file this PR touches.
- New: `packages/shared/queue/src/__tests__/cross-board.test.ts` (16 cases incl. the
  injected-classifier contravariance contract) and
  `packages/mobile/src/providers/__tests__/queue-provider-cross-board-add.test.tsx`
  (10 cases: never prompts same-board, prompts once, cancel leaves nothing behind,
  switch fires `setActiveBoard` + `addQueueItem` + `setSessionBoardPath`, one prompt
  for a burst, an unrelated add survives the await, no re-prompt for a board already
  in the queue, board-picker fallback).
- Full `vp run test:web` had 19 timeout-shaped failures in unrelated components
  (gym-detail, session-summary, auth forms) under parallel load; each passes on its own
  and none touch the two board-lock files this PR changes.
- Not run: device/simulator QA. `.boardsesh/qa-notes.md` lists the flows (same-board
  no-prompt path, each of the three outcomes, party-mode peers, Android Material's
  three-button dialog, the four locales).

### DIFF
--- a/packages/mobile/app/(tabs)/climbs/index.tsx
+++ b/packages/mobile/app/(tabs)/climbs/index.tsx
@@ -912,7 +912,7 @@ function ClimbListInner() {
 
   const handleAddToQueue = useCallback(
     (climb: Climb) => {
-      addToQueue({ uuid: randomUUID(), climb });
+      void addToQueue({ uuid: randomUUID(), climb });
     },
     [addToQueue],
   );
--- a/packages/mobile/src/components/climb-actions/use-climb-actions.ts
+++ b/packages/mobile/src/components/climb-actions/use-climb-actions.ts
@@ -201,7 +201,9 @@ export function useClimbActions({
       icon: 'add',
       color: successColor,
       run: () => {
-        addToQueue({ uuid: randomUUID(), climb });
+        // Fire-and-forget: the cross-board prompt (when it fires) sits above
+        // the dismissed sheet, so `after()` must not wait on it.
+        void addToQueue({ uuid: randomUUID(), climb });
         after();
       },
     });
--- a/packages/mobile/src/components/play-drawer/PlayDrawer.tsx
+++ b/packages/mobile/src/components/play-drawer/PlayDrawer.tsx
@@ -763,8 +763,13 @@ export function PlayDrawer({
   }, []);
 
   const handleSimilarClimbPress = useCallback(
-    (similarClimb: Climb) => {
+    async (similarClimb: Climb) => {
       const queueItem = climbToQueueItem(similarClimb);
+      // A similar climb can be set on another board, so this add may raise the
+      // cross-board prompt. Wait for it: backing out has to leave the drawer on
+      // the climb they were already looking at, not activate the one they just
+      // declined to queue.
+      if ((await addToQueue(queueItem)) === 'cancelled') return;
       // Tapping a similar climb activates it (commit), so it's never a preview —
       // clear any preview that was showing.
       setDrawerPreviewItem(null);
@@ -772,7 +777,6 @@ export function PlayDrawer({
       setIsMirrored(false);
       // The favorite override is cleared by the climb-change effect.
       setIsTickBarActive(false);
-      addToQueue(queueItem);
       setCurrentClimb(queueItem, { playlistSuggestionSource: null });
     },
     [addToQueue, setCurrentClimb],
@@ -1059,7 +1063,7 @@ export function PlayDrawer({
           angle={angle}
           onAddToQueue={() => {
             if (displayedClimb) {
-              addToQueue({
+              void addToQueue({
                 uuid: randomUUID(),
                 climb: displayedClimb,
               });
--- a/packages/mobile/src/lib/graphql/hooks/index.ts
+++ b/packages/mobile/src/lib/graphql/hooks/index.ts
@@ -38,6 +38,7 @@ import {
   type FavoritesQueryResponse,
 } from '@boardsesh/graphql/operations/favorites';
 import { BOULDER_GRADES } from '@boardsesh/board-config';
+import { myBoardsQueryKey } from '../query-keys';
 import { getDatabaseHandle } from '../../../db';
 import { offlineAwareRequest } from '../offline-request';
 import { useOfflineDownloadsEnabled } from '../../../providers/feature-flags-provider';
@@ -215,9 +216,14 @@ export function useUpdateProfile() {
 // Board Queries
 // ============================================
 
+// The key lives in `../query-keys` because `useCrossBoardAddGate` reads this
+// roster imperatively out of the cache; both sides import the one helper so the
+// shape can't drift apart. Re-exported here for callers already on this barrel.
+export { myBoardsQueryKey };
+
 export function useMyBoards(input?: MyBoardsInput, options?: { enabled?: boolean }) {
   return useQuery({
-    queryKey: ['myBoards', input],
+    queryKey: myBoardsQueryKey(input),
     queryFn: () => getHttpClient().request<GetMyBoardsQueryResponse>(GET_MY_BOARDS, { input }),
     select: (data) => data.myBoards,
     enabled: options?.enabled ?? true,
--- a/packages/mobile/src/lib/graphql/query-keys.ts
+++ b/packages/mobile/src/lib/graphql/query-keys.ts
@@ -1,0 +1,19 @@
+import type { MyBoardsInput } from '@boardsesh/shared-schema';
+
+/**
+ * React Query keys that more than one module has to agree on.
+ *
+ * A key that only its own hook uses stays next to that hook (see
+ * `ACTIVE_BOARD_QUERY_KEY` in `use-active-board.ts`). A key lands here once
+ * something reads the cache imperatively — `queryClient.getQueryData(...)` from
+ * a tap handler that must not subscribe — because that reader has no compiler
+ * link to the hook that wrote the entry, and a key change would leave it
+ * silently resolving `undefined` forever.
+ *
+ * This file is a leaf on purpose: types only, no hooks, no providers. The
+ * imperative readers live deep in provider land, and pulling the `hooks` barrel
+ * in for a key would drag half the app's module graph with it.
+ */
+
+/** Key `useMyBoards(input)` writes under; call with no argument for the plain roster. */
+export const myBoardsQueryKey = (input?: MyBoardsInput) => ['myBoards', input] as const;
--- a/packages/mobile/src/providers/__tests__/dialog-provider.test.tsx
+++ b/packages/mobile/src/providers/__tests__/dialog-provider.test.tsx
@@ -12,14 +12,16 @@ const alertMock = vi.hoisted(() => ({
   calls: [] as Array<{ title: string; message?: string; buttons: AlertButton[]; onDismiss?: () => void }>,
 }));
 
-// The provider needs Alert + Platform from react-native; mock both (Platform.OS is
-// controllable because the Paper dialog is Android-Material-only — iOS uses Alert).
+// The provider needs Alert + Platform + StyleSheet from react-native; mock them
+// (Platform.OS is controllable because the Paper dialog is Android-Material-only
+// — iOS uses Alert).
 vi.mock('react-native', () => ({
   Platform: {
     get OS() {
       return ctrl.os;
     },
   },
+  StyleSheet: { create: (sheet: Record<string, unknown>) => sheet },
   Alert: {
     alert: (
       title: string,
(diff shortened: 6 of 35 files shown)


marcodejongh:
Addressed in `dac7417` (finding 1) and `9049a75` (finding 2).

**1. `setSessionBoardPath` failure silently swallowed — real, but one layer up from where the review points.** The symptom is right: a dropped board-path broadcast leaves this climber on the new wall and the party on the old one, forever, with nothing to trace it. The located mechanism isn't. Adding `reportHandledError` to that `.catch()` alone would have reported almost nothing, because the shared mutation catches its own transport errors before the promise ever rejects — `packages/shared/queue-react/src/create-queue-mutations.ts:486` funnels them into `onBestEffortError` and resolves. The `.catch()` only ever sees a pre-send failure (`ensureJoined` rejecting inside `resolveCurrent`).

So both seams now report:

- `packages/mobile/src/providers/queue-provider.tsx:506-511` — the `onBestEffortError` sink reports `setSessionBoardPath` and keeps everything else dev-log-only. That was the actual silent path, and it also covers the *pre-existing* angle-change broadcast from `drawer-host-provider.tsx:521`, which had no handler at all. The rest of the sink stays quiet on purpose: a coalescer drain for `setCurrentClimb` or a superseded queue-add is reconciled by the WS subscription, and reporting per swipe would bury the signal. A board-path broadcast only fires on an angle change or a board switch, so the Sentry volume is negligible.
- `packages/mobile/src/providers/queue-provider.tsx:891-898` — the `.catch()` reports too, tagged `set-board-path-switch` so the rarer join-failure case stays distinguishable from the transport one.

No toast. The local add and the board switch both succeeded, there is nothing for the climber to retry, and "we couldn't tell your crew" on a successful action is noise — the actionable signal is telemetry.

Regression test in `queue-provider-cross-board-add.test.tsx` ("reports a failed peer broadcast instead of splitting the party in silence"): the broadcast rejects, the add still resolves `'added'` with the climb in the queue, and the report fires.

**2. Hardcoded React Query cache key — agreed, fixed.** Both sides now go through one helper, so the shapes can't drift:

- `packages/mobile/src/lib/graphql/query-keys.ts:19` — `myBoardsQueryKey(input?)`.
- `packages/mobile/src/lib/graphql/hooks/index.ts:226` — `useMyBoards` keys its `useQuery` with it.
- `packages/mobile/src/providers/queue/use-cross-board-add-gate.ts:39` — the gate reads the cache with it.

It's a leaf module rather than an export off the `hooks` barrel, and that isn't cosmetic: importing the barrel into the gate pulls `feature-flags-provider` → `auth-provider` → `AppLoadingSplash` into QueueProvider's module graph, which blows up all twelve QueueProvider suites on `StyleSheet.create` against their partial `react-native` mock. `query-keys.ts` is types-only, so the graph stays flat.

**Related, since it's the same failure mode as #2:** `onBestEffortError`'s `action` was typed `string`, so the new `action === 'setSessionBoardPath'` branch would have quietly become dead code the day someone renamed the mutation. It's now a `BestEffortAction` union (`packages/shared/queue-react/src/create-queue-mutations.ts:60-76`), exported from the package barrel, so a rename is a compile error at both the throw site and every caller that branches on it. Web's sink only interpolates the action into a log line, so it's unaffected.

Local run on this commit: mobile 598 files / 5238 tests all green, queue-react 43/43, `vp run typecheck` clean, `vp check` clean on every touched file.

marcodejongh:
Second pass, on the two observations from the latest review. Fixed the first in `24c178a`; leaving the second alone, reasoning below.

**1. Unhandled rejection when `setActiveBoard` throws — real, fixed.** Confirmed the whole chain: `runPrompt` rejects → the shared `inFlightRef` promise rejects → `addToQueue` rejects → and every real call site fires it as `void addToQueue(...)` (`use-climb-actions.ts:206`, `PlayDrawer.tsx:1066`, `app/(tabs)/climbs/index.tsx:915`), so it lands as an unhandled rejection with nothing on screen.

One correction to the diagnosis: `setActiveBoard` isn't a GraphQL mutation and network failure isn't the trigger. It's `setStoredActiveBoard` → `setPreference` (`use-active-board.ts:44-45`) — a local preference write — followed by a synchronous cache write. So the rejection comes from storage, not the wire, and it's rarer than "network failure" suggests. It also means the cache write is skipped when the store throws, so the board genuinely isn't active afterwards.

That's what makes it more than noise: without a catch, a failed activation fell through to `{ outcome: 'switch' }`'s caller path by rejecting rather than resolving, so there was no defined outcome at all. The board never became active, so the honest answer is to back out — queueing the climb at that point would drop a foreign climb onto the old board's queue, which is the exact thing this prompt exists to prevent.

So the switch path now catches, reports (`op: 'cross-board-switch'`), toasts `crossBoardAdd.switchFailed` in all four locales, tracks the outcome as `cancel` rather than the `switch` that didn't happen, and resolves `'cancelled'`. Covered by a new case in `queue-provider-cross-board-add.test.tsx` that asserts all five: resolves `'cancelled'`, queue untouched, no `addQueueItem`, no board-path broadcast, toast + report fired.

**2. `stateRef.current.queue` read before the dispatch commits — real in principle, unreachable in practice, leaving it.** The read is stale only between a `dispatch` and the render that reassigns `stateRef.current` (`queue-provider.tsx:222-223`) — under a frame. To land in that window a climber would have to answer the dialog for climb A and tap climb B from the same foreign board within ~16ms of the answer, and the burst case that *could* fire that fast — several taps before anyone answers — is already collapsed to one prompt by the `inFlightRef` dedup on `climbConfigKey`.

Closing it properly would mean shadowing the accepted-config set in a ref updated at commit time, i.e. a hand-maintained mirror of derived reducer state that can drift from the queue it claims to describe. That's a bigger correctness surface than the sub-frame double-prompt it buys, and the failure mode if it does fire is one extra dialog, not a wrong queue. Documented at the call site (`queue-provider.tsx:870-878`) rather than papered over.


Full mobile suite on this commit: 602 files / 5280 tests, all green. `vp run typecheck:mobile` clean, i18n catalog parity 90/90, `vp check` clean on every touched file.

PR #4252, opened 2026-08-10
fix(mobile): stop a playlist rename from silently overwriting another device

### LINKED ISSUE #1934: Server-side last-write-wins conflict resolution for playlist.rename
Parent: #1867. Part of #1859.

Single playlist edit that's last-write-wins with server `updated_at` check. On conflict (server has newer `updated_at` than the client's based-on snapshot), prompt the user to choose which version wins.

This is the only Phase 5 mutation that can't be auto-merged — flag it loudly so future mutation additions don't silently re-use the same shape.

marcodejongh:
## Summary

`updatePlaylist` was blind last-write-wins. It checked ownership, stamped a fresh `updatedAt` and overwrote whatever arrived, with no based-on check anywhere in the stack — so two devices renaming the same playlist meant one climber's edit vanished with nothing to show for it.

`UpdatePlaylistInput` now takes an **optional** `basedOn` snapshot. When it's present, the ownership check, a `FOR UPDATE` read and the write all happen inside one transaction, and a genuine collision throws `PLAYLIST_UPDATE_CONFLICT` carrying the server's current values. Omit `basedOn` and nothing changes — every shipped store binary keeps today's behaviour.

Mobile's playlist edit sheet sends the snapshot and prompts on a conflict: **Keep mine** (retries rebased on the server's own values, so the overwrite is deliberate), **Use theirs** (adopts the server version into the cache, sends nothing), **Cancel** (sheet stays open with the edit intact).

### Why `basedOn` carries values, not just a timestamp

A bare `updated_at` comparison — the issue's literal wording — isn't usable here. `addClimbToPlaylist`, `removeClimbFromPlaylist` and `reorderPlaylistClimb` all bump `playlists.updated_at`, and that column drives library ordering (`queries/user-playlists.ts`, `queries/discover.ts`), so they can't stop bumping it. A timestamp-only check would report a conflict on every climb add, including one from the same device's own outbox draining ahead of the rename.

So the timestamp stays as the cheap gate and a field-level comparison decides: conflict only when a field this edit writes disagrees with **both** the stored value and the value the client saw. That also turns an ambiguous outbox re-send of an already-applied rename into a no-op success rather than a bogus prompt.

Both sides of every comparison are server-authored — `basedOn.updatedAt` is a value the client echoes back from a previous server response, never a local clock reading — so device clock skew can't affect the outcome.

The alternative — a dedicated `metadata_updated_at` column bumped only by `updatePlaylist` — is conceptually cleaner but costs a migration, the renumber-bot dance, a new nullable `Playlist` field, and edits to six row→GraphQL mappers where a missed one silently degrades the check. Not worth it for this.

### Why the snapshot is taken when the sheet opens

The client captures `basedOn` on the press that opens the edit sheet, not on the press that saves. The detail query refetches on app foreground and on reconnect (the query provider bridges React Query's `focusManager` to `AppState` and `onlineManager` to NetInfo; global `staleTime` is five minutes), so a sheet left open while the climber gets interrupted can have another device's rename land in the cache underneath it. `PlaylistFormSheet` seeds its fields only on the visible false→true edge, so the typed name survives that refetch — and a snapshot read at save time would not. It would reach the server already matching the stored row, the compare-and-swap would find nothing to arbitrate, and the other device's rename would be overwritten in silence.

Capturing at open can't invent a false conflict either: a bump this device caused itself (adding a climb) moves `updated_at` without moving the metadata, and the server's field comparison still applies the edit.

### Flagged loudly, as the issue asks

`PLAYLIST_UPDATE_CONFLICT_CODE` in `@boardsesh/shared-schema` carries a doc block saying this is the one offline-replayable mutation that can't be auto-merged, that every other queued mutation is idempotent or commutative, and that any future non-mergeable mutation must reuse this shape rather than invent one. The offline outbox's `playlists`/`update` branch carries the matching warning, including why a conflict dead-letters (it resolves to no numeric status, so `isRetryable` is false) instead of burning retries.

### Out of scope

- Enqueueing playlist metadata edits into the offline outbox — nothing calls `enqueue()` for the `playlists` table today, so the conflict is reachable via two online devices or a stale react-query cache. The server rule and the queue's handling land here; wiring offline rename is its own issue.
- Web. This is no longer a gap: #4467 deleted `packages/web/app/components/library/` from `main`, and nothing under `packages/web` calls `updatePlaylist` any more (the one near-miss, `playlist-detail-content.tsx`, calls `updatePlaylistLastAccessed`). The mobile app is the mutation's only client, so the opt-in guarantee is now total rather than per-client. That also makes #4267 and its stacked PR #4303 obsolete — see the review comment below.
- Conflict detection for climb add/remove/reorder and `deletePlaylist` — those auto-merge correctly.
- Locking `playlist_ownership` alongside the playlist row. An ownership revoke racing an edit predates this PR and is shared with all four playlist mutations (the other three check ownership outside their transaction entirely); closing it means a consistent lock order across all four.
- No migration, no drizzle schema change, no new `Playlist` output field.

Closes #1934

## Release Notes

- Rename a playlist on two devices at once and neither edit disappears — Boardsesh now asks which one you want to keep.

## Test plan

- `vp run typecheck` — pass.
- `vp run test:mobile` — 684 files, 6557 tests pass. The `#1934` block in `app/(tabs)/discover/__tests__/playlist-uuid.test.tsx` drives the real cog that opens the sheet and covers: the snapshot sent with the edit; a conflict prompting without calling `reportHandledError`; "Keep mine" retrying with the server-derived snapshot; "Use theirs" patching both caches with no second write; the dismiss path leaving an editable sheet; a second race reported inline rather than as a fault; the prompt not quoting two identical names; and a refetch landing under the open sheet **not** moving the snapshot (fails on a revert).
- `vp check` — no errors or warnings in any touched file. The 2 errors and 324 warnings the run reports all reproduce on clean `main` (`scripts/`, `packages/mobile/src/web-shims/`, unresolvable local-only modules) and are tracked in #4488.
- `vp test run --project offline-sync --project graphql --project shared-schema` — `packages/shared/graphql/src/__tests__/errors.test.ts` (both error shapes, stripped extensions, unrelated codes), a `basedOn` round-trip through the outbox dispatcher, and a case pinning that a conflict is neither a network error nor retryable.
- `vp test run --project i18n` — catalog parity across en-US/es/fr/de.
- Backend: `packages/backend/src/__tests__/playlist-update-conflict.test.ts` carries 9 unit cases for the rule plus real-DB cases (no `basedOn` applies; matching snapshot applies; another writer's rename throws with the server's name/description/visibility/colour/icon; a climb add that only bumped `updated_at` applies with **no** conflict; a stale re-send of already-applied values is a no-op success; a "keep mine" retry rebased on the error payload applies; a non-owner is still rejected). `playlist-update.test.ts` adds a case pinning that the write's `updatedAt` is stamped after the `FOR UPDATE` read, not when the resolver was entered — an edit queued behind a concurrent writer would otherwise commit a timestamp older than the row it replaces, and that column orders the library.

### DIFF
--- a/docs/offline-sync-plan.md
+++ b/docs/offline-sync-plan.md
@@ -366,6 +366,14 @@ async function processMutation(mutation: PendingMutation) {
 
 Each mutation gets a client-generated UUID as an idempotency key. The backend's `saveTick` and `createPlaylist` accept this UUID and use `ON CONFLICT (uuid) DO NOTHING` for safe retry. Favorites use explicit `addFavorite`/`removeFavorite` (not `toggleFavorite`) so retries don't invert state. Follow/unfollow operations are naturally idempotent (follow when already following = no-op, unfollow when not following = no-op).
 
+### The one mutation that can't be auto-merged
+
+Every queued mutation above is idempotent or commutative, so replaying it is always safe. `updatePlaylist` is the exception: a queued rename replayed blind overwrites whatever the playlist is called now, including a rename made on another device while this one was offline (#1934).
+
+So it carries a `basedOn` snapshot — the `updatedAt` plus the metadata values the client last saw. The server compares that against the stored row inside a transaction holding a row lock and, when the two genuinely diverge, rejects with `PLAYLIST_UPDATE_CONFLICT` instead of writing. The error's extensions carry the server's current values, so the client can name both versions in a prompt and rebuild `basedOn` for a deliberate overwrite without a refetch. Omitting `basedOn` keeps the old last-write-wins behaviour, which is what shipped binaries and web still do.
+
+Two consequences for the queue. A rejection resolves to no HTTP status, so `isRetryable` returns false and the row dead-letters for a human to resolve instead of burning ten retries. And a climb added or reordered elsewhere bumps `playlists.updated_at` without touching the metadata, which is why the check compares fields rather than the bare timestamp — a timestamp alone would false-conflict on the same device's own queued climb add draining ahead of the rename. Any future mutation that can't be auto-merged should reuse this code rather than invent a shape.
+
 ### Dead letter handling
 
 Mutations that exceed `MAX_RETRY_COUNT` or fail with non-retryable errors are moved to `status = 'dead_letter'`. The app shows a badge/indicator when dead-letter mutations exist. Users can view failed mutations and choose to retry or discard them. This prevents silent data loss — the user always knows if a tick didn't sync.
--- a/packages/backend/src/__tests__/playlist-update-conflict.test.ts
+++ b/packages/backend/src/__tests__/playlist-update-conflict.test.ts
@@ -1,0 +1,347 @@
+import { describe, it, expect, beforeEach, afterAll } from 'vite-plus/test';
+import { eq, sql } from 'drizzle-orm';
+import { GraphQLError } from 'graphql';
+import { PLAYLIST_UPDATE_CONFLICT_CODE } from '@boardsesh/shared-schema';
+import type { ConnectionContext } from '@boardsesh/shared-schema';
+import { db } from '../db/client';
+import * as dbSchema from '@boardsesh/db/schema';
+import { playlistMutations } from '../graphql/resolvers/playlists/mutations';
+import {
+  detectPlaylistUpdateConflict,
+  normalizePlaylistText,
+} from '../graphql/resolvers/playlists/helpers/update-conflict';
+
+// #1934: updatePlaylist is the one offline-replayable mutation that can't be
+// auto-merged, so it takes an optional `basedOn` snapshot and refuses a genuine
+// collision instead of quietly keeping whichever write landed last.
+
+describe('detectPlaylistUpdateConflict', () => {
+  const storedAt = new Date('2026-08-10T12:00:00.000Z');
+  const olderSnapshot = '2026-08-10T11:00:00.000Z';
+
+  function stored(overrides: Record<string, unknown> = {}) {
+    return {
+      name: 'Crimps',
+      description: null,
+      isPublic: false,
+      color: null,
+      icon: null,
+      updatedAt: storedAt,
+      ...overrides,
+    };
+  }
+
+  it('never conflicts without a basedOn snapshot (blind last-write-wins stays the default)', () => {
+    expect(
+      detectPlaylistUpdateConflict({
+        stored: stored({ name: 'Renamed elsewhere' }),
+        incoming: { name: 'Mine' },
+      }),
+    ).toBe(false);
+  });
+
+  it('applies when the snapshot is as new as the stored row', () => {
+    expect(
+      detectPlaylistUpdateConflict({
+        stored: stored(),
+        incoming: { name: 'Mine' },
+        basedOn: { updatedAt: storedAt.toISOString(), name: 'Crimps' },
+      }),
+    ).toBe(false);
+  });
+
+  it('conflicts when someone else renamed the playlist', () => {
+    expect(
+      detectPlaylistUpdateConflict({
+        stored: stored({ name: 'Renamed elsewhere' }),
+        incoming: { name: 'Mine' },
+        basedOn: { updatedAt: olderSnapshot, name: 'Crimps' },
+      }),
+    ).toBe(true);
+  });
+
+  it('does not conflict when only the timestamp moved (a climb was added)', () => {
+    // addClimbToPlaylist/remove/reorder all bump playlists.updated_at, so a bare
+    // timestamp comparison would cry conflict on every climb add.
+    expect(
+      detectPlaylistUpdateConflict({
+        stored: stored(),
+        incoming: { name: 'Mine' },
+        basedOn: { updatedAt: olderSnapshot, name: 'Crimps' },
+      }),
+    ).toBe(false);
+  });
+
+  it('treats an already-applied edit as a no-op success, not a conflict', () => {
+    expect(
+      detectPlaylistUpdateConflict({
+        stored: stored({ name: 'Mine' }),
+        incoming: { name: 'Mine' },
+        basedOn: { updatedAt: olderSnapshot, name: 'Crimps' },
+      }),
+    ).toBe(false);
+  });
+
+  it("treats '' and null as the same value for description/colour/icon", () => {
+    expect(normalizePlaylistText('')).toBeNull();
+    expect(normalizePlaylistText(null)).toBeNull();
+    expect(normalizePlaylistText(undefined)).toBeUndefined();
+    expect(normalizePlaylistText('#123456')).toBe('#123456');
+
+    expect(
+      detectPlaylistUpdateConflict({
+        stored: stored({ description: null, color: null }),
+        incoming: { description: '', color: '' },
+        basedOn: { updatedAt: olderSnapshot, description: '', color: '' },
+      }),
+    ).toBe(false);
+  });
+
+  it('treats a missing based-on field as divergent (conservative)', () => {
+    expect(
+      detectPlaylistUpdateConflict({
+        stored: stored({ description: 'set elsewhere' }),
+        incoming: { description: 'mine' },
+        // No `description` in the snapshot: we can't prove the client saw the
+        // stored value, so refuse rather than overwrite.
+        basedOn: { updatedAt: olderSnapshot, name: 'Crimps' },
+      }),
+    ).toBe(true);
+  });
+
+  it('conflicts on visibility only when the client did not see the stored value', () => {
+    expect(
+      detectPlaylistUpdateConflict({
+        stored: stored({ isPublic: true }),
+        incoming: { isPublic: false },
+        basedOn: { updatedAt: olderSnapshot, isPublic: true },
+      }),
+    ).toBe(false);
+
+    expect(
+      detectPlaylistUpdateConflict({
+        stored: stored({ isPublic: true }),
+        incoming: { isPublic: false },
+        // isPublic absent from the snapshot.
+        basedOn: { updatedAt: olderSnapshot, name: 'Crimps' },
+      }),
+    ).toBe(true);
+  });
+});
+
+const PLAYLIST_UUID = 'pl-conflict-1934';
+const OWNER_ID = 'user-123'; // seeded by setup.ts
+
+function makeCtx(): ConnectionContext {
+  return {
+    connectionId: 'conn-1',
+    isAuthenticated: true,
+    userId: OWNER_ID,
+    sessionId: null,
+    boardPath: null,
+    controllerId: null,
+    controllerApiKey: null,
+  } as unknown as ConnectionContext;
+}
+
+// Read through drizzle (not raw `execute`) so `updated_at` arrives as the same
+// Date the resolver compares against, rather than a driver string.
+async function readPlaylist(): Promise<{ name: string; description: string | null; updatedAt: Date }> {
+  const [row] = await db
+    .select({
+      name: dbSchema.playlists.name,
+      description: dbSchema.playlists.description,
+      updatedAt: dbSchema.playlists.updatedAt,
+    })
+    .from(dbSchema.playlists)
+    .where(eq(dbSchema.playlists.id, 1n))
+    .limit(1);
+  return row;
+}
+
+describe('updatePlaylist — compare-and-swap against a real DB (#1934)', () => {
+  // Playlist tables aren't in setup.ts's per-file reset list, so own the reset.
+  beforeEach(async () => {
+    await db.execute(sql`TRUNCATE TABLE playlist_climbs, playlist_ownership, playlists RESTART IDENTITY CASCADE`);
+    await db.execute(sql`
+      INSERT INTO playlists (id, uuid, board_type, layout_id, name, is_public, created_at, updated_at)
+      VALUES (1, ${PLAYLIST_UUID}, 'kilter', 1, 'Crimps', false, now() - interval '1 hour', now() - interval '1 hour')
+    `);
+    await db.execute(sql`INSERT INTO playlist_ownership (playlist_id, user_id) VALUES (1, ${OWNER_ID})`);
+  });
+
+  afterAll(async () => {
+    await db.execute(sql`TRUNCATE TABLE playlist_climbs, playlist_ownership, playlists RESTART IDENTITY CASCADE`);
+  });
+
+  it('applies without a basedOn snapshot (unchanged behaviour for web and shipped binaries)', async () => {
+    await playlistMutations.updatePlaylist(
+      null,
+      { input: { playlistId: PLAYLIST_UUID, name: 'Blind write' } },
+      makeCtx(),
+    );
+
+    expect((await readPlaylist()).name).toBe('Blind write');
+  });
+
+  it('applies when the snapshot matches the stored updatedAt', async () => {
+    const before = await readPlaylist();
+
+    await playlistMutations.updatePlaylist(
+      null,
+      {
+        input: {
+          playlistId: PLAYLIST_UUID,
+          name: 'Fresh edit',
+          basedOn: { updatedAt: before.updatedAt.toISOString(), name: 'Crimps' },
+        },
+      },
+      makeCtx(),
+    );
+
+    expect((await readPlaylist()).name).toBe('Fresh edit');
+  });
+
+  it('throws PLAYLIST_UPDATE_CONFLICT carrying the server values when another writer renamed it', async () => {
+    const before = await readPlaylist();
+    // Another device renames the playlist and sets a description.
+    await db.execute(
+      sql`UPDATE playlists SET name = 'Renamed on the other phone', description = 'theirs', updated_at = now() WHERE id = 1`,
+    );
+
+    const failure = await playlistMutations
+      .updatePlaylist(
+        null,
+        {
+          input: {
+            playlistId: PLAYLIST_UUID,
+            name: 'Renamed on this phone',
+            basedOn: { updatedAt: before.updatedAt.toISOString(), name: 'Crimps', description: null },
+          },
+        },
+        makeCtx(),
+      )
+      .then(
+        () => null,
+        (error: unknown) => error,
+      );
+
+    expect(failure).toBeInstanceOf(GraphQLError);
+    const extensions = (failure as GraphQLError).extensions;
+    expect(extensions.code).toBe(PLAYLIST_UPDATE_CONFLICT_CODE);
+    expect(extensions.playlistUuid).toBe(PLAYLIST_UUID);
+    expect(extensions.serverName).toBe('Renamed on the other phone');
+    expect(extensions.serverDescription).toBe('theirs');
+    expect(extensions.serverIsPublic).toBe(false);
+    expect(extensions.serverColor).toBeNull();
+    expect(extensions.serverIcon).toBeNull();
+    expect(typeof extensions.serverUpdatedAt).toBe('string');
+
+    // Nothing was written.
+    expect((await readPlaylist()).name).toBe('Renamed on the other phone');
+  });
+
+  it('applies when only a climb add bumped updatedAt (the false-conflict this design prevents)', async () => {
+    const before = await readPlaylist();
+
+    await playlistMutations.addClimbToPlaylist(
+      null,
+      { input: { playlistId: PLAYLIST_UUID, climbUuid: '11111111-1111-1111-1111-111111111111', angle: 40 } },
+      makeCtx(),
+    );
+
+    const bumped = await readPlaylist();
+    expect(bumped.updatedAt.getTime()).toBeGreaterThan(before.updatedAt.getTime());
+
+    await playlistMutations.updatePlaylist(
+      null,
+      {
+        input: {
+          playlistId: PLAYLIST_UUID,
+          name: 'Renamed after the add',
+          basedOn: { updatedAt: before.updatedAt.toISOString(), name: 'Crimps', description: null },
+        },
+      },
+      makeCtx(),
+    );
+
+    expect((await readPlaylist()).name).toBe('Renamed after the add');
+  });
+
+  it('is a no-op success when a stale re-send asks for values that already landed', async () => {
+    const before = await readPlaylist();
+    await db.execute(sql`UPDATE playlists SET name = 'Same rename', updated_at = now() WHERE id = 1`);
+
+    // The outbox replays the rename it already delivered — it can't know the
+    // first attempt succeeded, and a bogus prompt here would be pure noise.
+    await playlistMutations.updatePlaylist(
+      null,
+      {
+        input: {
+          playlistId: PLAYLIST_UUID,
+          name: 'Same rename',
+          basedOn: { updatedAt: before.updatedAt.toISOString(), name: 'Crimps', description: null },
+        },
+      },
+      makeCtx(),
+    );
+
+    expect((await readPlaylist()).name).toBe('Same rename');
+  });
+
+  it('applies a "keep mine" retry rebased on the values from the conflict error', async () => {
+    const before = await readPlaylist();
+    await db.execute(sql`UPDATE playlists SET name = 'Theirs', updated_at = now() WHERE id = 1`);
+
+    const failure = await playlistMutations
+      .updatePlaylist(
+        null,
+        {
+          input: {
+            playlistId: PLAYLIST_UUID,
+            name: 'Mine',
+            basedOn: { updatedAt: before.updatedAt.toISOString(), name: 'Crimps', description: null },
+          },
+        },
+        makeCtx(),
+      )
+      .then(
+        () => null,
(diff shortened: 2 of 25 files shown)


marcodejongh:
## Review response

Four findings, checked one by one against the file. One is real and fixed, one is rebutted with a test that now proves it, one test gap is filled, one style call I'm keeping with a reason.

Fix commit: `0a82b86`.

---

### 1. "setSavingEdit(false) never called after conflict alert" — not reproducible, and now guarded

The claim is that the `return` inside the `catch` skips the reset. It doesn't: the reset lives in a `finally` attached to that same `try/catch`, and JS runs `finally` on a `return` from the `catch` block.

`packages/mobile/app/(tabs)/discover/[playlist_uuid].tsx`:

- L491 — `return;` inside the `catch`
- L498-500 — `} finally { setSavingEdit(false); }` on the enclosing `try`

So by the time the alert is on screen, `submitting` is already back to `false`; the sheet underneath stays open with the edit intact and the button live. Cancel needs no `onPress` — it only has to dismiss.

Rather than assert that from the spec, I made it a regression test — see finding 3.

I did add a comment at the `return` (L488-490) so the next reader doesn't have to re-derive it.

### 2. "Use theirs leaves userPlaylists cache stale" — real, fixed

Correct. The old handler wrote only `['playlist', uuid]`; the library shelves and the Add-to-Playlist picker kept the pre-conflict name until their cache went stale.

Now routed through the same `cacheUpdatedPlaylist` helper a successful save uses, so both caches move together (L449-457). One wrinkle worth naming: it merges over a **press-time** read of the detail cache (`queryClient.getQueryData`, L448), not the `playlist` row captured when the form was submitted. The prompt can sit on screen for a while, and reusing the submit-time row would roll back a climb added in the meantime.

New test: `"Use theirs" also refreshes the library list, not just the detail hero` — asserts the `['userPlaylists']` row picks up the server name and colour, and that `climbCount` on the detail row survives the merge.

### 3. "Missing test: Cancel path" — added

`leaves an editable sheet — not one stuck mid-save — when the prompt is dismissed`. It drives the real state machine rather than the spec: hand `updatePlaylist` a promise this test controls, assert the sheet reports `submitting` `true` while it's in flight, reject with the conflict, dismiss the alert via its cancel button, assert `submitting` is back to `false` with no inline error and no second write.

The form-sheet test double now surfaces `submitting` as `data-submitting` so this is observable at all.

Mutation-checked, so it's a real guard and not a tautology: deleting the `finally` and calling `setSavingEdit(false)` after the `catch` instead — i.e. the bug as described in finding 1 — fails it with `expected "true" to be "false"` at the dismiss assertion. Against the shipped code it passes.

### 4. "Keep mine styled destructive" — keeping it, with a reason

Among the three buttons, "Keep mine" is the one that throws data away: it overwrites the rename the other device already committed, and there's no undo. That's what iOS reserves red for, and it's the same treatment as the delete confirm on this screen (L511). "Use theirs" and Cancel are both recoverable — you can rename again.

The counter-argument (red discourages the likely intent) is fair, and it's a one-word change if Marco would rather it read neutral. I didn't want to make the discarding option look like the safe one on a guess.

---

**Validation** (all foreground, `vp` only): `vp run typecheck:mobile` clean; `vp run test:mobile` 601 files / 5269 tests passed. `vp check` is clean on both touched files — its 28 repo-wide errors are all pre-existing `NODE_ENV`/Sentry typings under root `scripts/`, untouched here. No backend change in this commit, so `playlist-update-conflict` is unaffected.

marcodejongh:
## Second-round review response

All four addressed in `bdebc5d7`. Three code/doc changes, one rebuttal-with-a-clarification.

### 1. keepMine retry swallows a second conflict — real, fixed

Correct, and it contradicted this PR's own premise: a conflict is the climber's call, not a fault, yet the retry path handed a second one to `reportHandledError` and showed "Failed to update playlist".

`packages/mobile/app/(tabs)/discover/[playlist_uuid].tsx` L477-490 now reads the conflict first and sets a distinct inline message, `edit.conflict.changedAgain` — "It changed again while you were deciding. Save again to pick it up." No Sentry report.

I kept your call on not re-prompting: a prompt chain has no natural end. Saving from the still-open sheet is the escape hatch, and it self-heals — the second attempt is still based on the pre-conflict snapshot, so it re-prompts with the newest server values.

New key added to all four locale catalogs (`catalog-completeness` enforces parity — 90 tests pass).

### 2. No test for keepMine-on-conflict — added

`"Keep mine" losing a second race says so inline instead of reporting a fault`: rejects both attempts with the conflict, presses Keep mine, asserts the inline error is `edit.conflict.changedAgain`, that `reportHandledError` was never called, and that `Alert.alert` still stands at one call (no second prompt).

### 3. Ownership check has no row lock — comment corrected, behaviour unchanged

The window is real, but it isn't something this PR introduced or narrowed: `reorderPlaylistClimb` (mutations.ts L503-518) runs its ownership check *entirely outside* its transaction, and add/remove do the same. `updatePlaylist` is the tightest of the four — the check at least shares the transaction with the write.

Closing it properly means locking `playlist_ownership` rows across every playlist mutation, with a consistent lock order so the four don't deadlock against each other. That's its own change, not a rider on a conflict-detection PR.

You're right that the comment implied more coverage than it has, so it now says which row is locked and which isn't (L169-177).

### 4. Docs — added

`docs/offline-sync-plan.md` gains "The one mutation that can't be auto-merged", immediately before the dead-letter section: why `updatePlaylist` is the exception among queued mutations, what `basedOn` carries, that omitting it keeps last-write-wins for shipped binaries and web, that a rejection has no HTTP status so `isRetryable` is false and the row dead-letters instead of burning ten retries, and why the check compares fields rather than the bare `updated_at` (a climb add bumps that column).

---

**Validation** (foreground, `vp` only): `vp run typecheck:mobile` and `vp run typecheck:backend` clean · `vp run test:mobile` 601 files / 5270 tests passed · `vp test run --project backend playlist-update-conflict` 15 passed · `catalog-completeness` 90 passed · `vp run check:i18n` OK · pre-commit `vp check --fix` clean on all touched files.

marcodejongh:
## Third-round review response

Both notes handled in `187fd326`.

### null → undefined in the Use-theirs cache write — documented, not changed

You're right about the asymmetry, and right that there's no live bug. I checked before deciding: nothing in `packages/mobile` or `@boardsesh/playlists-react` reads these fields with `=== null`, `!== null`, or `in` — every consumer treats them as truthy-or-not.

I left the coercion, because the alternative is worse than the trap. `Playlist` types them as `string | undefined` (`packages/shared/graphql/src/operations/playlists.ts` L188-206), so storing `null` here is a type error; the honest fix is widening the shared type to `string | null` to match what the server actually returns, and that lands on web and mobile at once — not something to smuggle into a conflict-handling PR.

What I did add is the convention at the coercion site, so the next reader hits the explanation instead of re-deriving it, with the pointer that the type is where to fix it if a consumer ever starts distinguishing the two.

### colour / icon conflict tests — added

`conflicts on a colour or icon someone else changed, same as a rename` in `packages/backend/src/__tests__/playlist-update-conflict.test.ts`: a stale `color` in `basedOn` against a newer stored value returns `true`, same for `icon`, plus the negative — a stored colour that already equals what this edit sends is not a conflict. That third case matters as much as the first two: it's the "someone else made the same change" path that keeps an outbox re-send from prompting for nothing.

---

**Validation**: `vp test run --project backend playlist-update-conflict` 16 passed · `vp run typecheck:backend` and `vp run typecheck:mobile` clean · pre-commit `vp check --fix` clean.

One note on the run, since it's worth recording: the first backend invocation failed one integration case after sitting 663s in tests. That's the shared `boardsesh_backend_test_w<N>` DB on :5433 — it has no per-worktree suffix, so another checkout's dev server was holding rows this file truncates. Rerun on the same commit: 16/16 in 44s. Not a change in this PR, but if that file ever looks flaky in CI, that's the first thing to rule out.

marcodejongh:
## Fourth-round review response

One correction, three accepted-as-is. `0e865c57`.

### `basedOn.isPublic: null` firing on a clean no-op — doesn't happen

`packages/backend/src/graphql/resolvers/playlists/helpers/update-conflict.ts` L101-106:

```ts
if (storedIsPublic !== incoming.isPublic && !clientSawStoredValue) return true;
```

The stored-vs-incoming comparison is the first operand. When they already agree there is nothing to decide between, so the branch returns false no matter what the snapshot says — an unknown `basedOn.isPublic` can only matter when the two visibilities genuinely differ, which is exactly the conservative case the comment above it describes.

Pinned rather than argued: the visibility test now carries a `stored: true / incoming: true / basedOn.isPublic: null` case asserting `false`. 16 pass.

### `Playlist` type using `string | null` — agreed, separately

Same conclusion as last round, and I think we agree on the destination: the type should say what the server returns. It's a shared type consumed by web and mobile, so widening it is its own PR with its own blast radius, not a rider here. The coercion is documented at the one site that performs it, and no consumer distinguishes the two today.

### Extensions-stripped fallback — accepted

Correct that there's no test, and correct that the behaviour is right: no extensions means nothing to name in a prompt, so the generic failure path is the honest outcome. A test there would pin our own mock's shape rather than a proxy's, which is why I'd rather leave it.

### Ownership check outside the row lock — accepted, comment already carries it

Unchanged from last round: it's the tightest of the four playlist mutations already (the other three check ownership outside their transaction entirely), and closing it properly means locking `playlist_ownership` across all four with a consistent lock order. Happy to file that as a follow-up issue if Marco wants it tracked rather than left in the comment.

---

CI green on `187fd326` (28 passed, 0 failed) and re-running on this commit. Marking ready for review.

marcodejongh:
## Fifth-round review response

Good catch on the prompt — real, and fixed in `4c21797a`.

### Conflict prompt quoting the same name twice

Confirmed against the check: `detectPlaylistUpdateConflict` refuses on `name`, `description`, `color`, `icon` (the shared `TEXT_FIELDS` loop) and on `isPublic`, so two sides that both kept the name and both changed the colour produce exactly the prompt you describe — "saved as Bad climbs now. Your edit says Bad climbs", which tells the climber nothing.

Of your two options I took the second (broaden the message) rather than scoping the check to name-only. Narrowing the check would trade a wording bug for a data-loss bug: a description or visibility change made on another device would go back to being silently overwritten, which is the whole of #1934.

So the prompt now branches on whether the names actually differ. When they match it reads: "Someone else changed this playlist's details while you were editing. The name matches, so pick which version of the rest to keep." Added to all four catalogs; `catalog-completeness` passes.

What I deliberately did not do is enumerate which fields disagree. The buttons resolve the whole record either way — there's no per-field Keep mine — so a field list would be detail the climber can't act on separately. If it turns out people want that, it wants a real diff sheet rather than a longer alert, and that's a design change.

New test: `does not quote two identical names back when the details are what diverged` — a colour-only conflict with a matching `serverName` asserts the alert uses `edit.conflict.messageDetails`, and that the three buttons are unchanged.

### Ownership lock

Noted, and unchanged from the previous two rounds by agreement: the race predates this PR and is shared with `addClimbToPlaylist` / `removeClimbFromPlaylist` / `reorderPlaylistClimb`, all of which check ownership outside their transaction entirely. Fixing it means locking `playlist_ownership` across all four with a consistent lock order — a separate change. Say the word and I'll file it as an issue rather than leave it living in a code comment.

---

**Validation**: `vp run typecheck:mobile` clean · `vp run test:mobile` 601 files / 5271 tests passed · `catalog-completeness` 90 passed · pre-commit `vp check --fix` clean. CI was green on `0e865c57` (26 passed, 0 failed) and is re-running here.

marcodejongh:
## Sixth-round: closing the loop

Green across the board on `4c21797a` — 28 checks pass, 0 fail. The `publish` job failed once on an S3 upload from the OTA preview client (`InternalError: Request body is not seekable, cannot retry`, an eoas-side retry bug, not a bundle problem); re-ran the job and it published clean.

On the three observations:

**1. `Keep mine` styled destructive.** Leaving it, and flagging that the two review passes disagreed about which button deserves red — the first said "Keep mine is the primary intent, drop the style", this one says "Use theirs discards the user's own edits, if anything that's the destructive slot". That disagreement is the tell that it's a judgement call, not a defect. My reasoning for red on Keep mine: it's the only button whose effect can't be undone from this screen. Use theirs discards a local edit the climber still has in front of them in the open sheet; Keep mine overwrites a rename that already committed on another device, and once it's gone there's no copy of it anywhere. Happy to flip it if Marco reads it the other way — one word.

**2. Ownership check outside the lock.** Agreed and deferred, as noted. Still happy to file it; leaving that to Marco since it means touching all four playlist mutations with a consistent lock order rather than just this one.

**3. Web sends no `basedOn`.** Filed: #4267. Carries the four implementation steps, the two traps mobile hit (the name-vs-name wording when the disagreement is about colour/description/visibility, and patching both caches on "use theirs"), and why the check can't go back to a bare timestamp.

PR is out of draft and ready for a human look.

marcodejongh:

marcodejongh:
## Adoption re-review (takeover loop `moarbugs100500`) — verdict: sound design, one real hole in the client snapshot

Re-reviewed the whole change against `main` (this branch is 211 commits behind), reading the surrounding resolvers, the query provider and the form sheet rather than the diff alone. The server-side rule is the right one and I could not break it. The client side has a window where the same silent overwrite the PR closes comes back.

**Conflict semantics — checked, and correct:**

- The comparison key is `playlists.updated_at` **plus** the five metadata values, and both sides of every comparison are server-authored. `basedOn.updatedAt` is a value the client echoes back from a previous server response, never a local clock reading. So **device clock skew cannot affect the outcome** — worth stating explicitly, because a naive `updated_at` scheme usually can't say that.
- The two-sided stage-3 test (`packages/backend/src/graphql/resolvers/playlists/helpers/update-conflict.ts:90-97`) is what makes the check usable at all: a climb add bumps `updated_at` without touching metadata, so a bare timestamp comparison would fire on every add. Verified the claim rather than trusting it — `addClimbToPlaylist` and friends do bump the column (`mutations.ts:487`, `:570`), and `PLAYLIST_ORDER` in `queries/user-playlists.ts:34` depends on it, so they genuinely can't stop.
- The check catches more than a name-vs-name race, which is the important part. If device A only recolours but its sheet resubmits the *old* name, that submit would revert device B's rename — and stage 3 refuses it, because `basedOn.name` disagrees with the stored name. Narrowing the check to name-only (an option raised in an earlier round) would have shipped a data-loss bug back in; the decision to broaden the copy instead was right.
- `maskDatabaseError` (`packages/backend/src/graphql/mask-error.ts:120-127`) passes non-DB errors through verbatim, so the extensions really do reach the client. Confirmed rather than assumed.
- A conflict resolving to no numeric status, so `isRetryable` is false and an outbox row dead-letters, is pinned by a test.

---

### 1. MAJOR — the `basedOn` snapshot is read at submit time, not sheet-open time, so a refetch under the open sheet reopens the exact bug

`packages/mobile/app/(tabs)/discover/[playlist_uuid].tsx:437-447` builds `basedOn` from `playlist`, which is live `useQuery(['playlist', uuid])` data read inside `handleEditSubmit`. The PR body says the client "sends the version the sheet was opened on" — it doesn't. It sends whatever is in the cache when Save is pressed.

`PlaylistFormSheet` seeds its fields only on the `visible` false→true edge (`packages/mobile/src/components/playlist/PlaylistFormSheet.tsx:87-105`, gated on `wasVisibleRef`), so the climber's typed name survives a refetch. The snapshot does not. The two diverge.

Failure scenario, every piece confirmed in-repo:

1. Open the edit sheet on a playlist, type a new name, don't save yet.
2. Global `staleTime` is 5 min (`packages/mobile/src/providers/query-provider.tsx:100`), `focusManager` is bridged to `AppState` (`:44-49`) and `onlineManager` to NetInfo (`:25`). Background the app for a while, or lose and regain signal — the playlist query refetches under the open sheet.
3. Meanwhile the other phone renamed the playlist. The refetch pulls that rename in: `playlist.name` becomes theirs, `playlist.updatedAt` becomes theirs.
4. Press Save. `basedOn` is now byte-identical to the stored row, so `detectPlaylistUpdateConflict` short-circuits at **stage 2** (`update-conflict.ts:87`, `stored.updatedAt <= basedOn.updatedAt`) and applies.

The other device's rename is overwritten with no prompt. That is #1934, unfixed, in precisely the case where the climber was slow enough to get interrupted — which is the shape of the bug being fixed, not an exotic edge.

The fix is small and strictly conservative: capture the snapshot when the sheet opens and use that one. It cannot manufacture false conflicts, because a self-caused `updated_at` bump with unchanged metadata still passes stage 3 on value equality. Fixing it here.

### 2. MINOR — the write's `updatedAt` is stamped before the row lock is taken

`packages/backend/src/graphql/resolvers/playlists/mutations.ts:156-158` builds `updateData` — including `updatedAt: new Date()` — before `db.transaction` opens at `:172`. With two edits contending on the same row, the loser blocks on `FOR UPDATE` and then writes the timestamp it minted *before* it started waiting. When the values don't collide it commits an `updated_at` older than the row it replaced. That column orders the library (`queries/user-playlists.ts:34`) and is the token the next client's `basedOn` is compared against. Moving the stamp inside the transaction costs nothing. Fixing.

### 3. Scope, stated plainly — the guarantee is per-client, and the offline framing isn't exercised yet

Not defects, but they belong in one place before a merge decision:

- **Web still sends no `basedOn`** (`packages/web/app/components/library/playlist-edit-drawer.tsx`, used by `packages/web/app/playlists/[playlist_uuid]/playlist-detail-content.tsx`). A web rename still silently overwrites a mobile one. That's #4267, and **PR #4303 is stacked on this branch** to close it — it is currently `CONFLICTING` and will need a rebase once this branch moves.
- **Nothing enqueues a playlist metadata edit into the offline outbox.** No `enqueue()` call targets the `playlists`/`update` pair, so the issue's literal framing — "the one Phase-5 mutation that can't be auto-merged" — is protected by documentation and a currently-unreached code path. That is the right call, and the doc block in `packages/shared-schema/src/playlist-conflict.ts` is genuinely what the issue asked for ("flag it loudly"). But the runtime coverage today is two online devices, or one device with a stale cache.
- **The device that loses is never told.** When the climber picks "Keep mine", the other phone flips to the new name on its next refetch with no notice. Inherent to any two-way prompt — flagging it so it reads as a decision rather than an oversight.

### 4. Nits

- `isPlaylistUpdateConflictError` (`packages/shared/graphql/src/errors.ts:112`) has no production caller — only tests and the mobile re-export. #4303 is presumably its consumer, so leaving it.
- The new Spanish strings say «lista» where most of `es/playlists.json` says "playlist". Both already appear in that catalog (`"Mis listas"` at :30 vs `"Compartir playlist"` at :104), so this is pre-existing drift, not introduced here.
- French correctly stays on *vous* (matching all 29 existing occurrences in that catalog) and German on *du*, per the glossaries.

### Tests — real, not decorative

Spot-checked that they'd fail on a revert. The mobile suite asserts `basedOn`'s contents on call 0 and the rebased snapshot on call 1, drives the alert's actual button array, and pins that `reportHandledError` is *not* called. The backend suite covers the climb-add-bumped-timestamp no-conflict case and the stale-re-send no-op, which are the two cases a naive implementation gets wrong.

### CI

Green on the current head — 28 pass, 0 fail, including `test-backend`, `typecheck`, `codegen-drift` and `i18n`. Rebasing onto `main` next; I'll re-report after the push.

marcodejongh:
## Rebased onto `main`, both findings fixed — plus: `main` has made #4267 / #4303 obsolete

Rebased 211 commits onto `origin/main` (one conflict, an import collision in `playlists/mutations.ts` from #4467's neighbouring work — resolved by keeping both) and pushed. Two fixes on top.

### Fixed — snapshot is now taken when the sheet opens

`9fc32ccf` — `basedOn` is captured on the press that opens the edit sheet and held in a ref, instead of being read off the live query at save time.

The new test drives the actual cog that opens the sheet, pushes a newer playlist into the cache mid-edit (standing in for the foreground/reconnect refetch), and pins that the snapshot sent is still the one the sheet was opened on. Verified it isn't vacuous: reverting the one line makes it fail with

```
-   "name": "Old name",
-   "updatedAt": "2026-08-10T11:00:00.000Z",
+   "name": "Renamed on the other phone",
+   "updatedAt": "2026-08-10T12:30:00.000Z",
```

which is precisely the silent overwrite. The other seven cases in that block now open the sheet properly too, rather than driving `handleEditSubmit` with the sheet notionally closed.

### Fixed — `updatedAt` is stamped once the row lock is held

`67470500` — moved off `updateData` and onto the `.set()` inside the transaction. The new `playlist-update.test.ts` case advances a fake clock during the locked read and asserts the stamped value is the post-wait one; it fails on a revert (`1787097600000` vs `1787097605000`).

---

### The web gap I flagged is gone, and it takes #4267 and #4303 with it

Worth surfacing because it changes the shape of this PR's guarantee.

`main` deleted `packages/web/app/components/library/` in **#4467** ("delete the climbing UI and swap the root chrome"), `playlist-edit-drawer.tsx` included. Grepping current `main`, **nothing under `packages/web` calls `updatePlaylist` any more** — the one hit, `packages/web/app/playlists/[playlist_uuid]/playlist-detail-content.tsx`, calls `updatePlaylistLastAccessed`, a different mutation.

So the "the guarantee is per-client, a web edit can still overwrite a mobile rename" caveat no longer describes reality. Mobile is the mutation's only caller, and it opts in. The protection is total.

Consequences for the stack, for you to decide:

- **#4303 cannot be rebased.** Its single commit modifies `playlist-edit-drawer.tsx` and its test, both deleted on `main`. `git merge-tree` against this branch's new head reports two modify/delete conflicts and nothing else — there is no merge to resolve, only files that no longer exist. My read is close it, and close **#4267** with it, rather than re-target the base. I have not touched either; that's your call.
- I have **not** re-pointed #4303's base. If you'd rather keep it, it needs a from-scratch reimplementation against whatever web surface replaces the drawer, not a rebase.

### Left alone deliberately

- **Ownership checked outside the row lock.** Unchanged across four review rounds by agreement, and I agree: it predates this PR, is shared with the other three playlist mutations, and closing it needs a consistent lock order across all four. Nothing here makes it worse.
- **`Keep mine` styled `destructive`.** Two prior review passes disagreed about which button deserves red, which is the tell that it's taste. Not touching it.
- **The prompt doesn't enumerate which fields disagree.** The buttons resolve the whole record either way, so a field list is detail nobody can act on. A real diff sheet is a design change.

### Validation on the rebased branch

`vp run typecheck` pass · `vp run test:mobile` 684 files / 6557 tests pass · `vp test run --project offline-sync --project graphql --project shared-schema --project i18n` 47 files / 993 tests pass · `vp run check:mobile-bundle` OK · `vp check` clean in every touched file (the 2 errors / 324 warnings it reports are the `main`-wide debt in #4488 — `scripts/`, `web-shims/`, unresolvable local-only modules — and reproduce on clean `main`).

CI is running on `67470500`; the previous head was 28 pass / 0 fail. Leaving the merge and the review to you.

marcodejongh:
CI green on `67470500` — 28 pass, 0 fail, nothing pending. Rebased onto `main` with no conflicts remaining, no unresolved review threads. Ready for your review and merge; leaving both to you.

marcodejongh:

Conflicting: ~200 lines in packages/mobile/app/(tabs)/discover/__tests__/playlist-uuid.test.tsx — the mobile playlist-rename test suite has diverged from main. Related to blocked #4303 (the web-side playlist-rename LWW fix) — same feature area on both ends, might be worth resolving together.

marcodejongh:
The current-main repair is complete locally at `97fb1f36c724d05e0295e3dda20a51799070ef9a` (`backlog/20260921/pr-4252-maintenance`). It preserves the current playlist UI/ownership checks and accepted three-choice conflict prompt. It also fixes the shared classifier regression introduced by current main: a definitive `PLAYLIST_UPDATE_CONFLICT` response is permanent, while HTTP 5xx and masked internal errors remain retryable.

Validation passed: 146 focused tests, 1,278 shared tests, 20 real local PostgreSQL tests, all 11,309 mobile tests (999 files), code generation, mobile typecheck, iOS/Android and browser Metro bundles, full lint/format, all 53 repository typecheck tasks and the commit hook. Independent paired source review approved the complete feature and follow-up. Device testing and published OTA delivery have not occurred.

**Push and preview publication remain held until backend support for `UpdatePlaylistInput.basedOn` is verified.** Current main lacks that input field. Anonymous read-only production introspection returned HTTP 403, so availability could not be established. Same-repository PR synchronizations publish a preview against production even while draft; publishing now could make playlist edits fail GraphQL validation. The remote head remains `67470500371b26dbc691cb13ad63808b329203ce`. No production changes or preview publication were performed.

Two review clarifications are now explicit in the local source/docs: current mobile metadata edits submit online, so no existing offline metadata enqueue/resolution UI is claimed; the current French catalog already uses vous and its climbing glossary does not require tu. The accepted prompt design remains intact: metadata conflicts with identical names use the generic details-changed message, without introducing a new field-diff UI. The old inline design thread can be checked against the repaired head once publication is unblocked. #1934 was reopened with source evidence because the unmerged backend/mobile protection remains relevant; retired web-only #4267 did not replace it.

PR #2053, opened 2026-05-10
Add per-user quality filter and rename Progress section to Logbook

marcodejongh:
## Summary

- Adds a **My Min Rating** star picker and a **Hide climbs I haven't rated** switch to the search form's user-progress panel, which is now renamed **Logbook** (en-US, es, fr).
- Promotes per-user climb quality and grade opinions to dedicated projection tables (`user_climb_qualities`, `user_climb_grades`), kept in sync on every tick write (saveTick / updateTick / deleteTick / Aurora sync / social seed) using a `climbed_at` staleness guard so out-of-order replays can't clobber a newer opinion.
- Migration `0090_backfill_user_climb_ratings` seeds the projection from the most-recent existing tick per `(user, climb[, angle])` — locally backfilled **5,279** quality rows and **6,071** grade rows.
- Existing readers of `boardsesh_ticks.quality` / `.difficulty` (smart playlists, social feeds, session summaries, exports) are untouched — the columns remain as the per-tick historical log.
- Drive-by: rename `count` → `value` in the `minAscentsOption` i18n key + call sites to clear a pre-existing typecheck failure on main (i18next's plural-aware overload required `count` to be `number`, but `formatMinAscentsFilterCount` returns a string like "10k").

Filter semantics:

| Min user rating | Hide-without switch | Behaviour |
|-----------------|---------------------|-----------|
| 0 (any) | off | No filter (default) |
| 4 | off | Show climbs the user hasn't rated **and** climbs rated 4+; hide climbs rated 1-3 |
| 4 | on | Show only climbs the user has rated 4+ |
| 0 | on | Show only climbs the user has rated, any value |

## Test plan

- [ ] `vp run typecheck` passes (verified locally).
- [ ] `vp check` passes (verified locally; only pre-existing warnings).
- [ ] Web tests pass except 4 pre-existing fr i18n catalog completeness gaps unrelated to this diff.
- [ ] CI runs the backend test suite (Postgres + Redis docker fixture) — local run skipped.
- [ ] Open the search drawer on a Kilter board page logged in as `test@boardsesh.com`. Confirm the **Logbook** panel renders with the new star picker and the "Hide climbs I haven't rated" switch above the existing four toggles.
- [ ] Set min user rating to 4 with the switch off → results contain unrated climbs **and** climbs rated 4+; verify a known 1-3 star climb is filtered out.
- [ ] Toggle the switch on → only climbs the user rated 4+ remain.
[...]

PR #4255, opened 2026-08-10
feat(mobile): filter climbs by the stars I gave them

### LINKED ISSUE #2645: Revive: Add per-user quality filter and rename Progress section to Logbook (was #2053)
PR #2053 (branch `feature/min-user-quality-filter`, preserved) implemented a per-user climb-quality search filter that never landed and has gone stale. Capturing the intent so it can be reimplemented fresh on current `main`.

## What the PR intended

- A **My Min Rating** star picker plus a **Hide climbs I haven't rated** switch in the search form, with the user-progress panel renamed to **Logbook** (en-US / es / fr).
- New per-user projection tables `user_climb_qualities` and `user_climb_grades`, kept in sync on every tick write (saveTick / updateTick / deleteTick / Aurora sync / social seed) with a `climbed_at` staleness guard so out-of-order replays can't clobber a newer opinion.
- A backfill migration seeding the projection from the most-recent existing tick per `(user, climb[, angle])`.
- Wires the filter through the GraphQL climb-search pipeline (schema + codegen) and `create-climb-filters.ts`.

Filter semantics: min rating 4 + switch off → unrated climbs **and** 4+; switch on → only the user's 4+; min 0 + switch on → any climb the user has rated.

## Why it's still wanted

The feature does not exist on `main` (no `user_climb_qualities` / `user_climb_grades` tables, no `minUserRating` / hide-unrated controls). Distinct from the already-closed issue #1838 (raw quality-filter text inputs). A per-user "show me the climbs I personally rated highly" filter is still a sensible search addition.

## Why the rebase was abandoned (mechanical scope only)

The branch is **~1462 commits behind main** (merge-base 2026-05-12). A rebase onto `main` produced structural conflicts that require feature-level rework, not mechanical resolution:

1. **Drizzle migration collision.** The PR adds `0089_mean_husk.sql` + `0090_backfill_user_climb_ratings.sql`, but those slots are now occupied by unrelated migrations (main is at `0118`). `meta/_journal.json` conflicts across ~200 lines and `meta/0089_snapshot.json` is an add/add conflict. Per repo rules migrations must be regenerated via `drizzle-kit generate` (no hand-written journal), so the projection tables need re-migrating as `0119+` against the current schema.
2. **Generated GraphQL codegen conflicts.** `packages/shared/graphql/src/generated/graphql.ts` (18 conflict markers), `gql.ts`, and `packages/shared-schema/src/generated/types.ts` all conflict. These are codegen output and must be regenerated after the schema change is re-applied, not hand-merged.
3. **i18n catalogs relocated.** Catalogs moved from `packages/web/i18n/locales/*/climbs.json` to `packages/shared/i18n/locales/*/climbs.json`; the new strings must be re-added in the new location.
4. **`create-climb-filters.ts` reworked on main** (hold-filter / climb-filters extraction), so the filter wiring needs reconciling against the current implementation.

Only commit 1 of 10 already hit 14 conflicts spanning all of the above, so this is beyond mechanical rebase scope.

## Revival

Reimplement on top of current `main`: regenerate the migration as `0119+`, re-add the projection-table sync at the current tick-write call sites, re-wire the filter through the current GraphQL search pipeline (then regenerate codegen), and re-add the Logbook strings in `packages/shared/i18n`. Branch `feature/min-user-quality-filter` is preserved as a reference.

Original PR: https://github.com/boardsesh/boardsesh/pull/2053

marcodejongh:
## Summary

Revives the per-user quality filter from the stale PR #2053, reimplemented on current `main`. Two new auth-gated search filters, wired end to end (GraphQL schema → Zod → SQL → shared filter state → mobile UI → offline SQLite):

- **`minUserRating`** — hide climbs whose *latest* rating from me, at the angle I'm browsing, is below N stars. Climbs I never rated stay visible.
- **`onlyRatedByMe`** — keep only climbs I've rated at this angle.

Together they reproduce the three cases the issue spells out: min 4 + switch off → unrated **and** my 4+; switch on → only my 4+; min 0 + switch on → anything I rated.

Closes #2645

### Deviations from the issue's original design

- **No `user_climb_qualities` / `user_climb_grades` projection tables, no tick-write sync, no backfill migration.** `boardsesh_ticks.quality` is already written by every tick writer (native `saveTick`, Aurora pulls, MoonBoard import), the offline SQLite mirror already syncs `quality` + `climbed_at` + `updated_at`, and both `create-climb-filters.ts` and `smart-playlists.ts` already read per-user quality straight off ticks. A projection would add a whole drift class (the issue's own `climbed_at` staleness guard exists only because of it) and would have to be mirrored into the offline schema too. Reading ticks directly is correct the instant a rating is edited or deleted.
- **Mobile-only UI.** The web climbing surface is deprecated (#3122), so the web search form is untouched beyond the two type/cache-correctness edits TypeScript and the CDN force.
- **The "Progress → Logbook" rename is skipped**, for the same reason (it was a web search-form rename).
- **Ratings are scoped to the browsed angle**, matching every sibling personal filter and matching the community `quality_average` this mirrors, which is itself stored per angle. A climb you rated 5 at 40° reads as unrated at 25°.

### Why the SQL looks like that

The obvious form — a scalar `(SELECT quality … ORDER BY climbed_at DESC LIMIT 1)` correlated subquery — cannot be unnested by Postgres, so it emits a SubPlan that runs once per **candidate climb**. Measured on the dev DB (kilter layout 1, angle 40, 220,056 candidates, a user with 529 rated ticks): list page 0 went from 498 ms to 1084 ms and the count from 1000 ms to 2724 ms.

Both filters are written as `EXISTS` / `NOT EXISTS` instead, which the planner unnests into semi/anti joins driven from `boardsesh_ticks` — the same reason the four sibling progress filters are fast. `minUserRating` is the anti-join "exclude climbs holding a rating below N that no newer rating supersedes", which preserves latest-wins exactly. Measured: list 567 ms, count 381 ms — at or below the unfiltered baseline.

### Cache correctness

Both params join `USER_SPECIFIC_SEARCH_PARAMS`, which drives **both** the resolver's `userId` resolution and `_isCacheable`. Without that the filter silently no-ops *and* a personalized result lands in the shared Redis entry. On the CDN side, `hasUserSpecificQueryParams` only recognised `'true'`/`'1'`, so a numeric `?minUserRating=4` read as absent; it now parses per type (flags stay `'true'`/`'1'`, numeric params check `Number(value) > 0`) rather than treating any non-empty value as user-specific, which would have handed crawlers a CDN bypass via `?onlyDrafts=x`.

## Release Notes

- Filter the climb list by the stars you gave: set a minimum, or show only the climbs you've rated
- Works offline on a downloaded board, and re-rating a climb updates what the filter shows

## Test plan

- `vp run typecheck` — green
- `vp test run --project backend --reporter=agent climb-queries` — 40 passed. New block seeds a user, two re-rated climbs (2→5 and 5→2), an unrated tick, a rating at a different angle and another user's rating, then asserts the three-case matrix, that `countClimbs` agrees with the list, and that the filter is a no-op without a `userId`.
- `vp run test:mobile` — 5273 passed. Adds the same three-case matrix against the offline SQLite path, the sheet's auth gating, the receipt-row chips, `statusForAuth`, and the recent-pill stripping.
- `vp test run --project climb-filters --project shared-schema` — 325 passed (filter-state forwarding + the `USER_SPECIFIC_SEARCH_PARAMS` guard).
- `vp test run --project web --reporter=agent middleware` — 140 passed (numeric param skips the CDN; junk values still cache).
- `vp check` — no new errors; one warning class I hit matched the file's existing pattern and was rewritten away. The 28 errors / ~298 warnings the repo reports are identical on a clean `main` checkout.
- `vp run check:i18n`, `vp run check:i18n:orphans`, `vp run check:mobile-bundle` — all OK.
- Full `vp run test:web` has 6-9 failing files that vary run to run (logbook, session-creation, gym-entity, error-boundary) — none touch this change; flaky/pre-existing.
- Not run: simulator/screenshot checks (macOS only). No device pass on the new sheet controls yet.

### DIFF
--- a/packages/backend/src/__tests__/climb-queries.test.ts
+++ b/packages/backend/src/__tests__/climb-queries.test.ts
@@ -855,4 +855,146 @@ describe('Climb Query Functions', () => {
       expect(await countClimbs(boundaryParams, statsFiltered)).toBe(2);
     });
   });
+
+  // Personal rating filters (#2645): "min stars I gave" + "only climbs I rated",
+  // read straight off boardsesh_ticks at the browsed angle. Latest rating wins,
+  // never-rated climbs stay visible unless onlyRatedByMe is on.
+  describe('personal rating filters (#2645)', () => {
+    const PREFIX = 'user-rating-';
+    const id = (suffix: string) => PREFIX + suffix;
+    const USER_ID = 'user-rating-tester';
+    const OTHER_USER_ID = 'user-rating-bystander';
+
+    const ratingParams: ParsedBoardRouteParameters = {
+      board_name: 'kilter',
+      layout_id: 1,
+      // Isolated on a size/set key no real catalog climb carries, so the
+      // assertions can enumerate the fixtures exactly (same trick as F4 above).
+      size_id: 98,
+      set_ids: [98],
+      angle: 40,
+    };
+
+    const search = (overrides: Partial<ClimbSearchParams> = {}): ClimbSearchParams => ({
+      page: 0,
+      pageSize: 100,
+      sortBy: 'creation',
+      sortOrder: 'desc',
+      ...overrides,
+    });
+
+    // Only the seeded climbs — the shared dev DB carries plenty of others at this key.
+    const seededUuids = async (searchParams: ClimbSearchParams, userId?: string): Promise<string[]> => {
+      const result = await searchClimbs(ratingParams, searchParams, userId);
+      return result.climbs
+        .map((climb) => climb.uuid)
+        .filter((uuid) => uuid.startsWith(PREFIX))
+        .sort();
+    };
+
+    const ALL_SEEDED = [
+      id('rated-5'),
+      id('rated-2'),
+      id('re-rated-up'),
+      id('re-rated-down'),
+      id('sent-unrated'),
+      id('untouched'),
+      id('other-angle'),
+      id('other-user'),
+    ].sort();
+
+    beforeAll(async () => {
+      await db.execute(sql`
+        INSERT INTO users (id, email, name, created_at, updated_at)
+        VALUES
+          (${USER_ID}, ${USER_ID + '@test.com'}, 'Rating Tester', now(), now()),
+          (${OTHER_USER_ID}, ${OTHER_USER_ID + '@test.com'}, 'Rating Bystander', now(), now())
+        ON CONFLICT (id) DO NOTHING
+      `);
+
+      await db.execute(sql`
+        INSERT INTO board_climbs (uuid, board_type, layout_id, setter_username, name, frames, frames_count, is_draft, is_listed, edge_left, edge_right, edge_bottom, edge_top, created_at, required_set_ids, compatible_size_ids)
+        VALUES
+          (${id('rated-5')}, 'kilter', 1, 'ur', 'Rated 5', 'p600r12', 1, false, true, 10, 100, 10, 150, '2024-01-01', ARRAY[98], ARRAY[98]),
+          (${id('rated-2')}, 'kilter', 1, 'ur', 'Rated 2', 'p601r12', 1, false, true, 10, 100, 10, 150, '2024-01-01', ARRAY[98], ARRAY[98]),
+          (${id('re-rated-up')}, 'kilter', 1, 'ur', 'Re-rated Up', 'p602r12', 1, false, true, 10, 100, 10, 150, '2024-01-01', ARRAY[98], ARRAY[98]),
+          (${id('re-rated-down')}, 'kilter', 1, 'ur', 'Re-rated Down', 'p603r12', 1, false, true, 10, 100, 10, 150, '2024-01-01', ARRAY[98], ARRAY[98]),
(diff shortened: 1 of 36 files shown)


marcodejongh:
QA pass complete (verdict: approve; the filter-summary forwarding gap is fixed in `ae2dd9a18` with tests). Two non-blocking notes for the record:

- **a11y**: the sheet now has two `StarRating` rows ("My rating" + community "Min rating") and both use StarRating's default `${n} stars` label, so VoiceOver reads them identically (`ClimbFilterSheet.tsx:881`). Worth distinct `getAccessibilityLabel` strings in a follow-up.
- **Rating-scale nuance**: Aurora-synced ratings only land on {1,3,5} (`convertQuality` maps 1-3→endpoints/midpoint), so for a climber whose stars all came from the Kilter/Tension app, "4+" behaves as "5 only". SQL is correct on the uniform 1-5 column — this is a copy/expectation note, not a defect.

Issue #4248, opened 2026-08-10
mobile: double-tap-favorite gesture + heart overlay on climb list rows


marcodejongh:
Split out of #2637 (scope decision 2026-08-10): the preserved branch worktree-favorites-uuid-only (originally PR #2480) included useDoubleTapFavorite + HeartAnimationOverlay on climb list rows. Re-integrating that gesture into main's extracted ClimbListItemContent / use-swipe-arm structure is feature integration work independent of the favorites re-key and should be its own PR.

Mobile perf checklist applies (no per-frame runOnJS; gesture composition per docs/react-native-performance.md).

Issue #4246, opened 2026-08-10
Favorites re-key Release 2: drop vestigial board_name/angle columns and wire fields


marcodejongh:
Follow-up to #2637's Release 1 (two-release rollout approved 2026-08-10). Once the mobile fleet has rolled past the Release-1 OTA:

- Drop user_favorites.board_name / .angle (vestigial, nullable) via drizzle-kit.
- Drop the optional-and-ignored boardName/angle fields from ToggleFavoriteInput/AddFavoriteInput/RemoveFavoriteInput and the favorites(boardName:, angle:) query args.
- Drop the vestigial local SQLite columns from offline-sync localColumns + stop emitting them in syncFavorites.
- Drop user_favorites_dedup_backup_* once the dedupe is confirmed good.

Gate on OTA adoption telemetry. Blocked until the #2637 Release-1 PR has been in production for a full release cycle.

PR #4256, opened 2026-08-10
feat(backend): key favorites by climb uuid so the heart survives a board switch

### LINKED ISSUE #2637: Revive: refactor(favorites) key by climb UUID only (was #2480)
Reviving the favorites-UUID-only refactor (closes the still-open #2449) from PR #2480. The branch `worktree-favorites-uuid-only` is preserved for revival.

## What the PR intended
Key favorites by `(userId, climbUuid)` instead of `(userId, boardName, climbUuid, angle)`. A climb is a climb regardless of board/angle, so favoriting on board A and switching board/angle should keep the heart. Scope:
- Schema migration: dedupe existing `user_favorites` rows by `(user_id, climb_uuid)` keeping most-recent, drop `board_name` + `angle`, add the new unique index.
- GraphQL `ToggleFavoriteInput` → `{ climbUuid }`; `favorites(climbUuids)` drops board/angle; remove unused `userFavoritesCounts` + `userActiveBoards` + `FavoritesCount`.
- Backend resolvers (`toggleFavorite`, `favorites`, `userFavoriteClimbs`, smart-playlist `LIKED_CLIMBS`, `mySmartPlaylistCounts`) match UUID-only; per-board scoping preserved by joining the unified `boardClimbs`/`climbs` table.
- Web `useClimbActionsData` cache key collapses to `[favorites, accumulated]`; `/api/internal/favorites` POST/GET drop board/angle.
- Mobile: drop the default-board workaround in `useMobileClimbActionsData.toggleFavorite`; `useDoubleTapFavorite`, `drawer-host-provider`, `PlayDrawer`, `climbs/[climbUuid]` stop passing board/angle.

Issue #2449 is still open and the feature has NOT landed in `main` (the `user_favorites` table still carries `board_name` + `angle`; `ToggleFavoriteInput` still requires them).

## Why the rebase was abandoned (too divergent for a mechanical rebase)
`git rebase origin/main` produced pervasive, logic-level conflicts that go beyond moved imports / renamed symbols:

1. **`packages/mobile/src/components/ClimbListRow.tsx`** — `main` extracted the row body and swipe handling into new modules that did not exist when the PR branched (`ClimbListItemContent.tsx`, `climb-list-row-styles.ts`, `climb-list-row-colors.ts`, `use-swipe-arm.ts`). The PR adds a brand-new double-tap-favorite gesture + heart overlay (`use-double-tap-favorite.ts`, `HeartAnimationOverlay.tsx` — neither exists in `main`). Re-wiring that new gesture into `main`s rewritten gesture/content structure is feature integration, not conflict resolution (~16 conflict regions across the file).

2. **`packages/backend/src/graphql/resolvers/playlists/queries/smart-playlists.ts`** — the recommendations engine (PR #2586) rewrote this dispatcher in `main` with new types and helpers (`resolveRecommendationBoardTarget`, `recommendation-refs`). The PR independently rewrote the `LIKED_CLIMBS` / `mySmartPlaylistCounts` paths. Reconciling two independent rewrites of the same query envelope is logic work.

3. **Migration meta** — the PR adds `0117_unknown_the_renegades.sql`, but `main` now has `0117_conscious_rhodey` + `0118_noisy_jane_foster`. `_journal.json` and `meta/0117_snapshot.json` conflict (add/add). The migration must be regenerated as `0119` via `bunx drizzle-kit generate` against current schema (the journal cannot be hand-edited per CLAUDE.md), and the snapshot rebuilt — not a text merge.

## Revival plan
Re-implement on a fresh branch off current `main`: regenerate the migration with `drizzle-kit generate`; re-apply the UUID-only resolver changes on top of `mains` recommendation-engine `smart-playlists.ts`; re-integrate `useDoubleTapFavorite` + `HeartAnimationOverlay` into `mains` extracted `ClimbListItemContent` / `use-swipe-arm` structure. The backend resolver + web hook + GraphQL schema changes are mostly portable; the mobile component and migration are the rework.

Original PR: https://github.com/boardsesh/boardsesh/pull/2480
Original issue: https://github.com/boardsesh/boardsesh/issues/2449

marcodejongh:
## Summary

A heart belongs to a climb, so changing board configuration or angle keeps it favorited. This PR rekeys Postgres favorites to `(user_id, climb_uuid)` and local SQLite favorites to `climb_uuid`, updates mobile favorites hooks and operations, and derives liked-list board scope from the climb catalog.

**Prerequisite satisfied:** #5341 merged and deployed successfully. [Production rollout](https://github.com/boardsesh/boardsesh/actions/runs/34292041166) completed the backend redeploy and live API checks with the compatibility commit included. This PR is now rebased onto main and retains that writer. See `docs/favorites-rollout.md` for the rollback boundary.

The migration archives duplicate angle rows, retains the newest favorite, and suppresses deletion records during deduplication. Legacy non-null sync fields and composite deletion IDs remain available to older apps. Removing a favorite also clears its archived angle variants on those devices. New apps accept both composite and UUID deletion IDs and protect later re-adds by timestamp. SQLite migration 6 preserves the already-shipped migration 5.

The GraphQL API retains `userFavoritesCounts`, `userActiveBoards`, and `FavoritesCount` for older clients, along with legacy favorite input fields. Their resolver results use catalog board identity after the UUID rekey. A full comparison against the deployed schema reports zero breaking changes; the server SDL is unchanged from main. A pinned legacy-contract test prevents removing the shipped fields, response shapes, or accepted inputs, and integration tests execute old library queries and response fragments with authentication and account isolation.

The rebase also preserves the current mobile optimistic-write behavior, refreshes every mounted favorites list after an account switch, retains the legacy database index in the Drizzle schema, and preserves the deletion trigger's pinned search path. Obsolete web favorites code was omitted because the climbing UI has moved out of Next.js.

Validation: the actual SQL migration passed in rolled-back scratch tables, including archival, deduplication without tombstones, legacy defaults, and deletion of archived angle variants. All 70 trigger search-path tests pass. Offline-sync previously passed 825 tests; mobile favorites regressions and both Metro bundles pass. The full mobile run had three unrelated preference/locale timeouts whose files passed on isolated rerun. After rebasing onto the deployed compatibility release and current main, full workspace typecheck and lint pass. The two targeted mobile suites pass all 30 tests. All 95 final backend integration tests from the preceding revision pass, including both client payload formats, concurrent writes, each unique-index shape, archived deletion variants, sync, and liked-list counts; those patches are unchanged by the rebase.

## Release Notes

- Your hearts follow the climb across board angles.
- Find each favorite once in your liked climbs.

## Test plan

1. Sign in → heart a climb → its heart fills.
2. Change board angle → reopen climb → heart stays filled.
3. Go offline → unheart climb → heart becomes empty.
4. Reconnect → reopen climb → heart stays empty.
5. Open Liked climbs → each favorited climb appears once.

## Risk

Risk: 5/5 — deduplicates stored favorites and changes offline keys; requires #5341 first.


marcodejongh:
Coordinator follow-ups on the two open items in the PR body:

**Duplicate count against prod: MEASURED.** Run 2026-08-10 by the coordinator session with the read-only prod credential (`PGOPTIONS="-c default_transaction_read_only=on -c statement_timeout=20000"`):

```sql
SELECT count(*) AS dup_groups, coalesce(sum(cnt-1),0) AS rows_to_delete
FROM (SELECT user_id, climb_uuid, count(*) cnt FROM user_favorites GROUP BY 1,2 HAVING count(*)>1) d;
-- => dup_groups = 0, rows_to_delete = 0
```

The dedupe/archive step deletes **zero rows** on current prod data — it stays in the migration defensively for dev DBs and any rows created between now and deploy. `user_favorites_dedup_backup_0190` row count will confirm post-apply.

**Release 2 follow-up: already filed** as #4246 (drop vestigial columns/fields once the fleet rolls), and the double-tap-favorite gesture split is #4248.

marcodejongh:
## QA pass (paired reviewer)

Adversarial review of the diff, the migration applied for real against scratch Postgres, and the suites re-run locally. Two things needed fixing; both are in `63e1deb`.

### Fixed here

**1. The web `GET /api/internal/favorites` full-table read.** Confirmed — with board and angle gone, the handler had no filter left and intersected in JS. Now filtered on the requested uuids in SQL, with a route test that pins the bound parameters (`['user-1', 'climb-1', 'climb-2']`) so the filter can't quietly disappear again.

**2. Deploy-window 42P10 on every favorite tap.** `migrate` is its own gated job that runs *before* `deploy-backend` (production-deploy.yml), so the previous backend serves against the new schema for the minutes Railway takes to roll. Its insert infers `ON CONFLICT (user_id, board_name, climb_uuid, angle)`; with the four-column index dropped, Postgres answers `there is no unique or exclusion constraint matching the ON CONFLICT specification` — reproduced against a scratch DB with 0190 applied. Migration 0190 now keeps `unique_user_favorite_legacy` over the four old columns for that window (strictly weaker than the new key, so it can never be violated on its own), and a backend test asserts the old target still infers. Release 2 (#4246) drops it with the columns.

Also closed a coverage hole rather than a defect: `mySmartPlaylistCounts`' `liked_climbs` CTE is raw SQL and `smart-playlists.test.ts` mocks `db.execute`, so *nothing* ran that statement against Postgres — a mistake in the `board_climbs` join would have taken out every count on the playlists tab, not just this one. There is now an integration test that executes it (2 catalog climbs across two boards, orphan excluded).

### Refuted

**The idempotency-key mismatch across the migration window does not lose a heart.** The outbox drains `ORDER BY created_at ASC, id ASC`, so a stale old-format `del:user_favorites:kilter:<uuid>:40` is always older than the post-OTA `add:user_favorites:<uuid>` and drains first. Net state is correct; the cost is one extra server round trip.

### Verified

- **Migration 0190 applied for real**, twice: with duplicates (newest row kept, losing rows archived, `sync_deletions` count **0** across the dedupe, trigger re-enabled) and on a prod-shaped zero-duplicate DB (clean no-op, archive empty). Post-migration insert with neither board nor angle lands `''`/`0`; deleting it emits `record_id = <climb_uuid>`.
- **`NOT NULL DEFAULT ''`/`0` is the right call over nullable.** Every remaining writer omits both columns, so nullable would have put `NULL` on the wire for new favorites — straight into a pre-OTA device's `NOT NULL` local column, which is the exact pull-cycle outage the two-release split exists to avoid. No writer to `user_favorites` remains outside the two mutation paths.
- **Old binaries.** `String!`→`String` and `Int!`→`Int` keep every shipped `useFavoriteStatus` document and the queued old-format outbox payload validating (`variables: { input: payload }` passes the extra fields straight through). `rest-parity.test.ts`'s inline `favorites(boardName:, climbUuids:, angle:)` still validates.
- **Count/page parity.** `board_climbs.uuid` is the primary key, so the climbs join is 1:1 — `totalCount` cannot drift from the page.
- **Local table is write-only on device** (only a `DELETE` and the sign-out wipe list), so the 1-part tombstone that old clients skip really is cosmetic.
- `vp run check:db-migrations` (191, journal consistent) · `vp run codegen` (no diff) · offline-sync 326 · backend favorites/sync/smart-playlists suites · mobile 5292 · `vp check` 28 errors/298 warnings, unchanged from base (all pre-existing in `scripts/` + `aurora-sync`).

### Two notes for the human pass, no code change

- **Sequence the backend deploy ahead of the web deploy and the production OTA.** `GET_FAVORITES` no longer sends `boardName`/`angle`, which the *old* schema requires — if web or the mobile OTA lands first, the favorites query fails validation and hearts read blank until the backend catches up. Self-healing, but visible.
- **QA note step 3 targets the wrong surface.** `useClimbActions` returns `[]` when `boardConfig` is null, so with no board selected that sheet shows no actions at all — nothing to tap. The no-board fix is in `useMobileClimbActionsData.toggleFavorite`; exercise it from the surface that uses the favorites provider.

Verdict: approve once CI is green.

marcodejongh:
Rebased onto `main` and renumbered. The migration is now **`0194_favorites_key_by_climb_uuid`** (was `0190` — `main` took 0190–0193 with the MoonBoard grade/stats work).

Three conflicts, all resolved to keep both sides:

- **`playlists/queries/smart-playlists.ts`** — the one the db-renumber bot stopped on. #4262 moved `mySmartPlaylistCounts` inside a single `withSerialPlan` transaction; this branch rewrote the `liked_climbs` CTE for UUID-only favorites. Kept the serial-plan wrapper *and* the catalog join, so the counts CTE now reads `JOIN board_climbs bc ON bc.uuid = uf.climb_uuid` from inside the transaction. `selectSmartClimbRefs` / `countSmartClimbRefs` merged clean.
- **`offline-sync/src/sync/table-config.ts`** — `main` removed the per-table `invalidateKeys` (#4350 moved it to one shared map); kept that removal plus this branch's `primaryKeyColumns: ['climb_uuid']`.
- **`mobile/src/hooks/use-offline-mutations.ts`** — `main` swapped the favorites writes onto `runLocalWrite` and added the `getLocalUserId` owner stamp; kept both, with the UUID-only insert: `INSERT OR IGNORE INTO user_favorites (climb_uuid, user_id, created_at, updated_at)`.

The transitional legacy index's backup table moved with the number: `user_favorites_dedup_backup_0190` → `user_favorites_dedup_backup_0194`, plus the migration-number comments in `favorites-uuid-key.test.ts` and `schema-sql.ts`. No `_bs_migration_guards` row keys on the old tag, so nothing else in the body needed touching.

Validation (all foreground, all green): `vp run typecheck`, `vp run check:db-migrations` (195 migrations, journal consistent), backend `favorites-uuid-key` / `sync-mutations` / `sync-pull` / `smart-playlists` (6 files, 81 tests), `--project offline-sync` (31 files, 681 tests), and the mobile suites covering the resolved hook (58 tests).

Two notes worth flagging:

- `vp check` reports formatting issues in `docs/design/web-reposition/gyms-directory.html` and `homepage-marketing.html`. Both come from `main` (19268ee78), are untouched here, and fail the same way on a clean `main` checkout — pre-existing, not from this rebase.
- `hooks-dual-write.test.ts` is modified by this PR but never executes: `packages/mobile/vite.config.ts` excludes it (RN 0.86 Flow-source parse during the Rolldown scan). That exclusion predates this branch, so those assertions are unverified here as they are on `main`.

marcodejongh:

PR #4263, opened 2026-08-11
fix(backend): gate saveTick's boardUuid rung on a full board config match

### LINKED ISSUE #4219: saveTick's boardUuid rung attributes ticks to any board with no config gate
Noted during #4216's security review: the `boardUuid` rung in `saveTick` (packages/backend/src/graphql/resolvers/ticks/mutations.ts) attributes a tick to any board by uuid with no config validation and no ownership/visibility check — any authenticated user can stamp ticks onto any board's stats and leaderboards if they know its uuid.

This is pre-existing and intentional for named-board routes (ticking a gym's seeded board you don't own is legitimate), but it is strictly weaker than the newer rungs: the presence-boardId and session-board rungs both require a full config match before attributing. The gap: config-gating the boardUuid rung the same way (best-effort — mismatch records the tick unassociated, matching the rung's existing stale-uuid behaviour) would close the pollution vector without breaking the legitimate flow.

Related wrinkle: #4216 stripped duplicate-error identity extensions for non-owner editors partly because a leaked private slug plus session attribution could reach private boards; config-gating boardUuid closes the analogous direct path.

marcodejongh:
## Summary

`saveTick` resolves a tick's board through four rungs. The `boardUuid` rung — the one a named-board route (`/b/<slug>/<angle>/...`) uses — selected the board by uuid and took its id without checking anything else, even though it was already selecting the board's config columns. Any authenticated caller who knew a board uuid could stamp ticks onto that board's stats and leaderboards regardless of what configuration the tick itself named.

The two newer rungs (presence board, session board) already required a full config match — board type + layout + size + hold sets — before attributing. This applies the same gate to the `boardUuid` rung and pulls the three copies of the comparison into one helper.

Behaviour on a mismatch follows the rung's existing best-effort contract: the tick is still saved, just with `board_id` NULL, and it deliberately does **not** fall through to the config lookup — a supplied-but-rejected uuid must stay unassociated rather than quietly landing on the climber's own same-config board. A `[saveTick] Ignoring tick boardUuid …` warning makes any real client hitting this visible in backend logs after deploy.

An input carrying a `boardUuid` but no `layoutId`/`sizeId`/`setIds` is also left unassociated — otherwise omitting the config would be a way around the gate.

### Changes

- `packages/backend/src/graphql/resolvers/board-presence/shared.ts` — new `boardConfigMatchesTick(board, target)` next to `normalizeSetIds`.
- `packages/backend/src/graphql/resolvers/ticks/mutations.ts` — rung 1 gated on it; rungs 2 and 3 refactored onto it (behaviour-preserving); four-rung doc comment updated.
- `packages/backend/src/__tests__/save-tick-board-uuid-config-gate.test.ts` — 8 real-DB cases.

### Client audit (why this doesn't break the legitimate flow)

`boardUuid` is only ever injected by the shared `BoardProvider`, and only the web named-board route sets it. Both web tick call sites (`logascent-form.tsx`, `use-tick-save.ts`) always send `layoutId`/`sizeId`/`setIds` from the same board's route params, so the gate passes for the gym-board flow the rung exists for. Mobile never sends `boardUuid`. Offline replay sends the stored variables verbatim, so config survives replay.

### Not in scope

- Ownership/visibility gating of the rung — would break ticking an unlisted or party-host board you don't own, and needs a product decision.
- Checking that the tick's climb belongs to the resolved board's layout — applies to all four rungs, deserves its own issue.
- Backfilling existing `boardsesh_ticks` rows attributed through the ungated rung.

## Release Notes

Board leaderboards now only count ticks actually logged on that board's setup — a tick whose layout, size, or hold sets don't match the board it names is kept in your logbook but left off that board's stats.

- [ ] No release note needed (internal / technical change)

## Test plan

- `vp run typecheck:backend` — pass
- `vp test run --project backend --reporter=agent save-tick-board-uuid-config-gate legacy-tick-board-attribution board-presence` — pass, 96/96 (a first attempt can time out in the shared-worker-DB setup hook; clean on re-run)
- Reverted the rung-1 gate locally and re-ran the new suite: 5 of its 8 cases fail without the fix, so it is a real regression guard rather than a happy-path sweep.
- `vp check` — pass
- The full `vp test run --project backend` sweep did not complete locally: ports 8082/8084, which `integration.test.ts` and the resolver suite hard-bind, are held by two other worktrees' long-running Expo dev servers on this machine. CI runs it on a clean box.

New cases: matching config attributes to a gym board you don't own; differing layout / size / hold sets each record the tick unassociated and specifically not on the climber's own matching board; a board of a different type with an identical layout/size/set records unassociated; `2,1` vs `1,2` still attributes; boardUuid with no config records unassociated; unknown uuid records unassociated.

QA notes: `.boardsesh/qa-notes.md`.

Closes #4219

### DIFF
--- a/packages/backend/src/__tests__/save-tick-board-uuid-config-gate.test.ts
+++ b/packages/backend/src/__tests__/save-tick-board-uuid-config-gate.test.ts
@@ -123,6 +123,10 @@ describe('saveTick boardUuid config gate', () => {
   });
 
   afterAll(async () => {
+    // Belt and braces: afterEach already clears these, but a run killed between
+    // tests would otherwise strand boards in the shared worker DB.
+    await db.execute(sql`DELETE FROM boardsesh_ticks WHERE user_id = ${USER_ID}`);
+    await db.execute(sql`DELETE FROM user_boards WHERE owner_id IN (${USER_ID}, ${OTHER_USER_ID})`);
     await db.execute(sql`DELETE FROM board_climbs WHERE uuid = ${CLIMB_UUID}`);
     await db.execute(sql`DELETE FROM users WHERE id IN (${USER_ID}, ${OTHER_USER_ID})`);
   });
--- a/packages/backend/src/__tests__/save-tick-board-uuid-config-gate.test.ts
+++ b/packages/backend/src/__tests__/save-tick-board-uuid-config-gate.test.ts
@@ -70,6 +70,7 @@ function tickInput(overrides: Record<string, unknown> = {}) {
 
 async function insertBoard(opts: {
   ownerId?: string;
+  boardType?: string;
   setIds?: string;
   layoutId?: number;
   sizeId?: number;
@@ -79,7 +80,7 @@ async function insertBoard(opts: {
   const result = await db.execute(sql`
     INSERT INTO user_boards
       (uuid, slug, owner_id, board_type, layout_id, size_id, set_ids, name, is_public, created_at, updated_at)
-    VALUES (${uuid}, ${uuid}, ${opts.ownerId ?? USER_ID}, ${CONFIG.boardType}, ${opts.layoutId ?? CONFIG.layoutId},
+    VALUES (${uuid}, ${uuid}, ${opts.ownerId ?? USER_ID}, ${opts.boardType ?? CONFIG.boardType}, ${opts.layoutId ?? CONFIG.layoutId},
             ${opts.sizeId ?? CONFIG.sizeId}, ${opts.setIds ?? CONFIG.setIds}, ${opts.name}, true, now(), now())
     RETURNING id
   `);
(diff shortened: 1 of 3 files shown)


marcodejongh:
Addressed in 25c3cb1:

- **Warning log conflating two cases** — fixed. The no-config case now logs its own message (`input carries no layout/size/set to match against`) instead of `.../undefined/undefined/undefined`.
- **`afterAll` cleanup** — fixed. The suite now deletes its ticks and boards in `afterAll` too, so a killed run can't strand rows in the shared worker DB.
- **`TickBoardConfigTarget` living in `board-presence/shared.ts`** — keeping it there deliberately. The interface only exists as the parameter type of `boardConfigMatchesTick`, which has to live next to `normalizeSetIds` (it's the comparison's other half), and `mutations.ts` already imports `findActiveBoardById` from the same module — so the dependency edge is pre-existing, not new. Splitting the type from the function it describes would cost a file for no isolation gain.

marcodejongh:
Second review pass addressed in 4c136ed:

- **boardType-mismatch test** — added. Layout ids aren't unique across board types, so a Tension board whose layout/size/sets line up with a Kilter tick is exactly what the type leg guards; the new case asserts it lands unassociated.
- **"Redundant" null guards in rungs 2/3** — they're load-bearing for TypeScript, not for the runtime. Dropping `explicitBoard &&` makes the following `explicitBoard.id` fail with `TS18048: 'explicitBoard' is possibly 'undefined'` (verified by removing it and running `vp run typecheck:backend`). The helper still accepts null so callers that don't need the narrowing can pass one straight through.
- **`TickBoardConfigTarget` placement** — as in the previous pass: it's the parameter type of `boardConfigMatchesTick`, which has to sit next to `normalizeSetIds`, and `mutations.ts` already imported from that module. Keeping type and function together.

marcodejongh:
On the raw-SQL test helpers: deliberate, and matching the sibling suite. `legacy-tick-board-attribution.test.ts` — the file this one is modelled on, covering the same resolver — seeds `users`, `board_climbs`, `user_boards` and `board_sessions` the same way, and reads the tick back with a raw `SELECT`. The Drizzle-over-raw-SQL rule is about production queries; test fixtures deliberately bypass the ORM so a schema-mapping bug can't hide the behaviour under test. Keeping the two attribution suites identical in shape is worth more here than ORM purity.

PR #4264, opened 2026-08-11
fix(web): stop Google sign-in redirecting to localhost

### LINKED ISSUE #4227: Logging in with google on web app redirects to localhost
https://github.com/user-attachments/assets/60f0acf5-9b5d-4cc8-b189-b27f7743696a

marcodejongh:
## Summary

Signing in with Google on www.boardsesh.com sent people to `http://localhost:3000`. Reproduced against production: `GET /api/auth/providers` returned `signinUrl`/`callbackUrl` on `http://localhost:3000`, and a real CSRF + `POST /api/auth/signin/google` returned a `Location` carrying `redirect_uri=http%3A%2F%2Flocalhost%3A3000%2Fapi%2Fauth%2Fcallback%2Fgoogle`.

Root cause is next-auth v4's origin resolution (`utils/detect-origin.js` in the installed 4.24.15): it returns `process.env.NEXTAUTH_URL` first, falls back to the forwarded host only when `VERCEL` or `AUTH_TRUST_HOST` is set, and otherwise returns undefined — at which point `utils/parse-url.js` defaults to `http://localhost:3000`. In the production runtime `NEXTAUTH_URL` is either unset or literally a localhost value, so every auth URL collapsed to loopback. Nothing in `packages/web` hardcodes localhost in the OAuth path.

The same root cause was silently breaking the shared-cookie design: `sessionCookieDomain()` keys on the `NEXTAUTH_URL` hostname, so production cookies were being set with no `Domain` attribute (confirmed on the live `/api/auth/csrf` response) — `app.boardsesh.com` could never see the www login.

This PR is the code-side hardening. `packages/web/app/lib/auth/canonical-auth-url.ts` resolves the origin this deployment should use, in order:

1. an explicit `NEXTAUTH_URL` whose hostname is not loopback — used verbatim,
2. `https://$VERCEL_URL` when `VERCEL_ENV === 'preview'` — a preview always uses its own deployment URL,
3. `BASE_URL` (already the email base URL, baked as an ARG in `Dockerfile.web`),
4. `VERCEL_ENV === 'production'` → `https://www.boardsesh.com`,
5. `VERCEL_URL` → `https://$VERCEL_URL`,
6. otherwise nothing, so local dev keeps next-auth's `http://localhost:3000`.

A loopback `NEXTAUTH_URL` is only ignored when the deployment is clearly hosted: `VERCEL`, `VERCEL_ENV` of `production`/`preview`, `VERCEL_URL`, a **non-loopback** `BASE_URL`, or `AUTH_COOKIE_DOMAIN`. A developer machine is deliberately never hosted — the tracked `packages/web/.env.local` ships `VERCEL_ENV=development` **and** `BASE_URL=http://localhost:3000`, so a looser test would strip a correct local `NEXTAUTH_URL` on every boot (and break sign-in outright on a non-3000 dev port, e.g. `PORT=3095 vp run dev`, where next-auth would fall back to its own `:3000` default).

`applyCanonicalAuthUrl()` writes the result into `process.env.NEXTAUTH_URL` and logs a warning naming the old and new value — the only signal an operator gets that the env var is wrong. It runs at module scope in `auth-options.ts` (above the cookie config, which reads the same variable) and once per server boot in `instrumentation.ts`, so routes that reach `secure-cookies.ts` without importing `authOptions` (e.g. `/api/internal/ws-auth`) see it too.

Deliberately **not** fixed with `AUTH_TRUST_HOST`: that makes next-auth build the redirect URI out of the `X-Forwarded-Host` header, which is host-header-injection surface. A static canonical origin is the safer shape.

Closes #4227

### Why a preview beats `BASE_URL`

`BASE_URL` is the email base URL and is the kind of variable that gets set once for *every* Vercel environment. If a preview inherited `https://www.boardsesh.com` from it, two things break at once: preview OAuth bounces users to production, and `sessionCookieDomain()` emits `Domain=.boardsesh.com` on a `*.vercel.app` response, which the browser rejects — the preview login silently never stores a cookie. That is exactly the failure `secure-cookies.ts` is written to prevent, so the preview's own deployment URL takes precedence.

### Deployment (not in this PR)

`NEXTAUTH_URL=https://www.boardsesh.com` still needs setting in the Vercel Production environment — Marco is handling that separately. The resolver is the backstop, not a replacement: if the only signal present is a localhost `NEXTAUTH_URL` and nothing else, it drops the poison value (letting next-auth's own forwarded-host path work) but cannot invent the canonical origin.

Also worth checking on the production Google OAuth client: `https://www.boardsesh.com/api/auth/callback/google` must be an authorized redirect URI. If only the localhost URI is registered, this fix turns a silent wrong-redirect into a visible `redirect_uri_mismatch`.

### Risk: session cookies gain a Domain

Once `NEXTAUTH_URL` resolves to `https://www.boardsesh.com`, `sessionCookieDomain()` flips from `undefined` to `.boardsesh.com` and the session cookie gains a `Domain` attribute. That is a browser-distinct second cookie of the same name alongside any existing host-only cookie, and it switches on the `appendLegacyHostOnlySessionCookieClear` migration path in `secure-cookies.ts`. **Expect some already-logged-in web users to be signed out once.** This is the intended shared-cookie design (it is what lets app.boardsesh.com share the login), but it is a visible one-time effect worth knowing about before merge.

## Release Notes

- Signing in with Google or Apple on the web now brings you back to Boardsesh instead of dumping you on a dead localhost page.

## Test plan

- `vp test run --project web` — full web project green (5917 tests). New `canonical-auth-url.test.ts` covers the precedence table, loopback aliases, unparseable values, idempotency, the tracked local-dev env (`VERCEL_ENV=development` + loopback `BASE_URL` + a non-3000 port), the preview-over-`BASE_URL` rule, and the regression invariant that no hosted-env fixture ever ends up holding a loopback origin. Two new cases in `auth-options.test.ts` prove the module-scope side effect rewrites `NEXTAUTH_URL` before the cookie config is built and leaves local dev alone.
- `vp check`, `vp run typecheck:web`.
- Existing cookie-domain suites (`auth-options.test.ts`, `secure-cookies.test.ts`) pass unchanged — Vercel previews and homelab preview hosts still get host-only cookies, localhost still gets the non-secure cookie name.
- Post-deploy verification (after the env var lands): `curl https://www.boardsesh.com/api/auth/providers` shows https URLs, and the CSRF + `POST /api/auth/signin/google` `Location` carries `redirect_uri=https%3A%2F%2Fwww.boardsesh.com%2F...`.
- No dev server started; this is env/config resolution covered by unit tests. QA notes in `.boardsesh/qa-notes.md`.

### Out of scope

- A post-deploy production smoke asserting `/api/auth/providers` never returns a loopback URL belongs on #4171, which already owns that surface.
- `generate-openapi.ts` and the setter OG route have the same `VERCEL_URL`/localhost fallback shape; both cosmetic, left for a follow-up.
- `instrumentation.ts` only patches the Node.js runtime. `middleware.ts` does not read `NEXTAUTH_URL` (it decides HSTS from the request protocol), so the edge runtime needs nothing here.


marcodejongh:

marcodejongh:
🤖 bug-batch loop `moarbugs100500`: this merged while my adopted-PR review was still running, and the review found a defect. **Fix-forward PR incoming** — flagging here so it is on the record against the merge commit (`efec8b22b`).

## Sign-in can stay broken for a subset of users

`auth-options.ts` now sets `domain: cookieDomain` on **five** cookies: `sessionToken`, `callbackUrl`, `state`, `nonce`, `pkceCodeVerifier`.

Adding a `Domain` attribute does not overwrite a pre-existing host-only cookie of the same name — it creates a browser-distinct second cookie, and both then ride every www request. `secure-cookies.ts` already knows this: that is exactly what `appendLegacyHostOnlySessionCookieClear` exists for. But it only clears `sessionCookieName()`. The four OAuth-flow cookies got the Domain treatment with no equivalent cleanup.

Severity splits by cookie, and this is the part that matters:

| Cookie | `maxAge` in next-auth 4.24.15 | Consequence of a stale host-only copy |
|---|---|---|
| `state`, `pkceCodeVerifier` | `60 * 15` | Self-expires in 15 min — narrow, self-healing |
| **`nonce`** | **none** | **Browser-session lifetime.** Google/Apple validate the id_token nonce against this cookie; which of the two same-named cookies the server reads is not deterministic, so sign-in can fail for the rest of the session |
| `callbackUrl` | none | Post-login redirect pinned to the stale value for the rest of the session |

Verified against the installed copy at `node_modules/.bun/next-auth@4.24.15+*/node_modules/next-auth/core/lib/cookie.js` — `nonce` and `callbackUrl` are the two with no `maxAge`. `csrfToken` uses `__Host-` and never got a Domain, so it is unaffected.

The fix mirrors the mechanism already in the file: emit a `Domain`-less `Set-Cookie` clear for each of the four flow cookies, keeping the existing `sessionCookieDomain() === undefined` guard that stops it deleting the freshly-written login on dev and preview hosts.

## Separate, needs an owner decision — not something I will change unilaterally

`Domain=.boardsesh.com` sends the session JWT to **every** boardsesh.com subdomain. That includes `updates.boardsesh.com`, which runs third-party `mercuretechnologies/expo-open-ota`. The code documents the widening as intentional and reviewed, and the OAuth-flow cookies now ride along with it too, which the original security note did not cover.

Two ways out if that is not the intended blast radius: move the OTA server off the `boardsesh.com` parent domain, or replace the shared parent-domain cookie with an explicit cross-subdomain handoff so the credential is never broadcast. Both are architecture calls, so they are @marcodejongh's, not mine.

marcodejongh:
🤖 **Retracting my earlier comment on this PR.** A QA pass caught that its headline claim was wrong, and I verified the correction myself.

## What I got wrong

I claimed a stale host-only `__Secure-next-auth.nonce` could break Google/Apple sign-in "for the rest of the browser session", on the reasoning that which of two same-named cookies the server reads is non-deterministic. **It is deterministic, and it favours the fresh cookie.**

`parseCookie` in `next@16.2.12` (`next/dist/compiled/@edge-runtime/cookies`) builds a `Map` with `map.set(key, value)` iterating the `Cookie` header left to right, so the **last** occurrence wins. RFC 6265 §5.4 orders cookies of equal path length by creation time ascending, so the newly-written `Domain=.boardsesh.com` cookie sorts last and is the one that wins. A pre-existing host-only duplicate loses on every request.

So there is no persistent sign-in loop, and the user-facing framing in that comment was wrong. Apologies for the noise — it was an unverified mechanism asserted with more confidence than it had earned.

## What is actually left

One narrow window, not a per-user persistent failure: during a rolling deploy, a request served by an old instance writes a host-only cookie *after* a new instance wrote the domain-scoped one. Creation-time ordering then puts the stale copy last and it wins until it is overwritten or expires. `state` and `pkceCodeVerifier` carry `maxAge: 60 * 15` so they self-clear; `nonce` and `callbackUrl` have no `maxAge` and persist for the browser session.

That is a real but transient and self-healing edge, worth closing as hygiene — not the user-facing breakage I described. PR #4556 is being reworked to say that instead, and its release notes (which told climbers sign-in "kept bouncing") are being removed as fiction.

## Unchanged

The separate point in that comment still stands and still needs an owner decision: `Domain=.boardsesh.com` sends the session JWT to every boardsesh.com subdomain, including `updates.boardsesh.com`, which runs third-party `mercuretechnologies/expo-open-ota`. The OAuth-flow cookies now ride along with it too, which the original in-code security note did not cover.

Issue #1875, opened 2026-05-04
Resolve user-sync-cron sleeper


marcodejongh:
Parent: #1860. Part of #1859.

`user-sync-cron` (`maxDuration = 300`, dev-mode auth bypass) is **not** listed in `vercel.json` — triggered some other way today. Decide intent before migration:
- Keep as on-demand via `packages/scheduler` API, OR
- Schedule it explicitly

Without resolving this, the route silently disappears from the schedule on Railway.

marcodejongh:
Colocating under the kill-Vercel epic #4648. This gets resolved as part of #4654 (remaining crons → `packages/scheduler`) — the decision needed here (explicit schedule vs on-demand `scheduler run <job>`) is called out there so the route doesn't silently drop off the schedule at cutover.

marcodejongh:
Closing as superseded by the long-lived Aurora/Kilter sync daemons. Both `/api/internal/user-sync-cron` and backend `POST /sync-cron` were removed; [the scheduler documentation explicitly records this issue as needing no schedule decision](https://github.com/boardsesh/boardsesh/blob/main/docs/scheduler.md#not-in-scope). The daemons loop internally and coordinate their cooldowns in Postgres, so there is no sleeping cron route left to migrate. This was reconciled in #5047.

PR #4271, opened 2026-08-11
feat(repo): add packages/scheduler and move the cleanup cron off Vercel

### LINKED ISSUE #1874: Build packages/scheduler service
Parent: #1860. Part of #1859.

New `packages/scheduler` service: node-cron + shared-secret `CRON_TOKEN`. Migrates the 8 jobs in `vercel.json`. Long-lived Node container on Railway.

`vercel.json` itself is deleted at the Phase 1 cutover, not here — the file's removal is a Phase 1 child.

marcodejongh:
> ⚠️ **Merge gate: the Railway service has to exist first.** This PR removes
> the only trigger for `/api/internal/cleanup`. Merging before the scheduler is
> deployed pauses the 180-day feed-item and 90-day notification retention. It
> stays a draft until step 6 of the checklist below has passed against
> production.

## Summary

Closes #1874.

Adds `packages/scheduler`, a long-lived Node service that takes over the cron
jobs currently declared in `packages/web/vercel.json`, and migrates the first
one (`cleanup`).

The issue says "8 jobs"; the file actually declares **7** — five
`prewarm-heatmap` boards, `profile-percentiles`, and `cleanup`. This PR moves
`cleanup` only, the lowest blast radius of the seven, and leaves the other six
where they are. `vercel.json` itself stays — the issue assigns its deletion to a
Phase 1 child.

### The scheduler is a trigger, not a job reimplementation

Each job makes an HTTP request to the existing `/api/internal/*` route with
`Authorization: Bearer $CRON_SECRET` — byte-identical to what Vercel
auto-injects, and exactly what `requireCronAuth` already validates.

That isn't a shortcut. Two of the three job families cannot run outside Next:
`prewarm-heatmap` warms `cachedGetHoldHeatmapData`, a Next cache entry whose key
must match a real first-visit request; `profile-percentiles` ends with
`revalidateTag(USER_CLIMB_PERCENTILE_CACHE_TAG)`. Neither is reachable from a
plain Node process. So the routes stay the single implementation.

This also settles the issue's `CRON_TOKEN` wording: `cron-auth.ts` documents
that the env var must stay `CRON_SECRET` or Vercel's auto-injection silently
401s every remaining cron. The scheduler sends the same value under the same
name; nothing is renamed. That doc comment now records the second caller.

### What's in the package

- `src/config.ts` — env parse that fails fast. A missing `CRON_SECRET` throws at
  startup, not at 05:00.
- `src/cron/` — a five-field cron parser and a once-a-minute ticker, each job
  evaluated in its declared IANA timezone.
- `src/jobs/` — `triggerWebCron(path)` plus the registry (one entry today). A
  502/503 from the edge is retried once after 2s, inside the job's `timeoutMs`;
  504 is not, because the request reached the route and a retry would stack a
  second run on the first.
- `src/runner.ts` — per-job in-flight guard (a tick whose predecessor is still
  running is skipped and warned, never queued), try/catch so a throwing job
  can't kill the process, and a status map. `scheduler run <job>` deliberately
  bypasses that guard — an operator asking for a run gets one — so
  `JobDefinition.run` documents that a job has to tolerate overlapping itself.
- `src/health-server.ts` — `GET /health` (liveness, what Railway polls) and
  `GET /health/jobs` (503 when a scheduled job's last run failed — point an
  alert here, not Railway: restarting can't fix a bad secret or a WAF rule and
  would wipe the `lastError` that says which).
- `src/cli/index.ts` — `start`, `run <job>` (the on-demand trigger), `list`.

### Deploy: no new image, no new workflow

`Dockerfile.sync` is already documented as "runs any CLI directly via tsx" and
`.github/workflows/sync-deploy.yml` already triggers on `packages/**`, so adding
`@boardsesh/scheduler` to the sync context's root packages is the whole thing.
`scripts/check-service-deploy-inputs.mjs` derives its expectations from the same
walk, so only its fixture needed the extra workspace.

## Marco: deploy checklist (infra, not in this PR)

**Do this before merging**, or the 180-day feed-item / 90-day notification
retention silently stops — the `vercel.json` edit removes the only trigger.
Harmless for weeks (delete-by-age catches up next run), but it shouldn't sit
that way.

1. **New Railway service**, name `scheduler`.
2. **Source → Docker Image**: `ghcr.io/boardsesh/boardsesh-sync:production`
3. **Custom start command**:
   `node --import tsx packages/scheduler/src/cli/index.ts start`
4. **Variables**:
   - `CRON_SECRET` — **copy** the Vercel project value, don't regenerate. The
     six remaining Vercel crons authenticate with the same secret.
   - `BOARDSESH_WEB_URL=https://www.boardsesh.com`
   - `PORT=8080`
5. **Healthcheck path** `/health`. **Replicas: 1** — two instances double-fire
   every job; there's no leader election in this slice.
6. **Verify against production** before merging:
   `node --import tsx packages/scheduler/src/cli/index.ts run cleanup` must
   return the route's JSON and exit 0. This is the step that catches a Vercel
   WAF/bot rule blocking Railway egress IPs — nothing local can.

Between steps 1 and merge both schedulers may fire the job. That's safe:
`/api/internal/cleanup` deletes rows older than a fixed age in deadline-bounded
batches, so the second pass is a no-op.

Full runbook: `docs/scheduler.md`.

## Two decisions worth your call

1. **Image topology.** This rides the existing `boardsesh-sync` image, which
   keeps the slice to one PR but couples scheduler releases to the sync deploy
   and gives the container a misleading name. A dedicated
   `Dockerfile.scheduler` + `scheduler-deploy.yml` is mechanical but a second
   PR's worth of wiring — say the word.
2. **Migration pace for the remaining six.** One PR each (safer) or one batch
   once `cleanup` has run cleanly for a week.

Also: sibling issue #1875 (`user-sync-cron` sleeper) looks already moot —
`docs/aurora-sync.md` and `docs/branch-deploys.md` both record that the
`/api/internal/user-sync-cron` route and the backend `POST /sync-cron` handler
were removed in favour of the long-lived daemons, and grep confirms no such
route exists. Worth closing.

### Deviations from the plan

- **No `node-cron`, `commander` or `dotenv`.** The loop I'm running under
  forbids new npm dependencies, so `src/cron/` implements the five standard
  numeric fields (`*`, `a`, `a-b`, lists, `/step`) and a minute ticker in ~200
  tested lines, and the CLI parses its own argv. Month/weekday names, `@daily`
  shorthands and Quartz extensions throw at parse time rather than being
  silently reinterpreted. `pnpm-lock.yaml` gains only the new workspace entry —
  zero new resolved packages. The runner takes the cron scheduler as an injected
  interface, so swapping in `node-cron` later is a one-file change.
- **`"types": ["node"]` in the package tsconfig.** Unlike the sync CLIs, this
  package has no runtime deps to drag `@types/node` in transitively, so
  `process` / `node:http` / `NodeJS.*` don't resolve without naming it.
- **`registerSchedules: false`** on the runner instead of faking a disabled-jobs
  list, so `run <job>` can never also start the recurring schedule and still
  works on a job held back by `SCHEDULER_DISABLED_JOBS`.

### Rebased and finished (2026-09-01)

The branch was written under Bun in August and has been rebased onto
post-pnpm `main`. Five commits on top of the original six:

- The Railway start command and the runbook still reached for a Bun launcher.
  The image has neither Bun nor `vp`; everything now says
  `node --import tsx packages/scheduler/src/cli/index.ts <cmd>`, the form
  `Dockerfile.sync`'s header already documents for the sync daemons.
- `main`'s deploy-input fixture grew its own `Dockerfile.sync`, so every test in
  `check-service-deploy-inputs.test.mjs` now generates the `sync` context and 15
  of 17 red'd on the new workspace. The fixture declares `packages/scheduler`
  alongside the three sync packages.
- One retry on a 502/503, so a daily job that lands in a deploy swap isn't
  written off for 24 hours.
- `healthServer.stop()` had no `.catch`: a rejected close left the process alive
  with its signal handler latched, so Railway would SIGKILL it at the end of the
  stop grace period rather than the container exiting on SIGTERM.
- `vercel.json` lost `refresh-sitemap-climbs` on `main` (#4852) while this
  branch sat. `VERCEL_OWNED_CRON_PATHS` matches today's file (six entries) and
  both drift guards — `registry.test.ts` and web's `rest-surface-inventory` —
  are green against it.

## Release Notes

## Test plan

1. CI green.
2. Railway scheduler service: /health returns 200 and /health/jobs returns 200.
3. Next 05:00 UTC: cleanup job reports success on /health/jobs.

## Risk

Risk: 3/5 — new long-lived service; merging before the Railway service exists pauses the retention cron (harmless for weeks).

## Validation run on this branch

- `vp test run --project scheduler` — 8 files, 71 tests.
- `vp test run --project scripts` — 82 files, 1334 tests.
- `vp run check:service-deploy-inputs` — 17/17, and the generated sync context
  lists `packages/scheduler` among its source packages.
- `vp test run --project web` on the cron / REST-surface / cleanup files — 4
  files, 18 tests (the cleanup route's own tests stay green untouched: proof
  only the trigger moved).
- `vp run typecheck:scheduler`, `vp run typecheck:web` — clean. `vp check` is
  clean on every touched file; its only two repo-wide errors are the
  pre-existing `@gorhom/bottom-sheet` TS2307s in
  `packages/mobile/src/web-shims/`, whose isolated install is a web-task-only
  dependency.

What the tests pin down:

- `0 5 * * *` fires exactly once in the 05:00 UTC minute and not in 05:01.
- The same expression in `Australia/Sydney` fires at 19:00 UTC while the
  UTC-scheduled one stays quiet — the timezone is the job's, not the host's.
- A tick whose predecessor is still in flight is skipped and warned.
- A rejecting job is caught, recorded, and the next tick still fires.
- A hung request aborts at the job's `timeoutMs` instead of blocking forever.
- A 503 is retried once and the second attempt's result is returned; a 502
  retries exactly once before giving up; a 500 never retries; the retry wait
  abandons itself on shutdown.
- `registry.test.ts` reads `packages/web/vercel.json` and fails if any path is
  scheduled on both sides, or if the checked-in Vercel-owned list drifts from
  the file — so a future migration PR that edits one side only fails CI.

Real image, not just the CLI: built `.docker-context/sync` locally and ran the
image.

- `node --import tsx packages/scheduler/src/cli/index.ts list` prints
  `cleanup 0 5 * * * UTC /api/internal/cleanup`.
- `start` boots, `/health` returns 200 with the live status map, `/health/jobs`
  returns 200, an unknown path 404s, and SIGTERM exits 0.
- `run cleanup` with no `CRON_SECRET` fails at config parse before any request.

Not verifiable locally: whether Vercel's WAF lets Railway egress through. That's
step 6 of the deploy checklist, deliberately ordered before the merge.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -123,6 +123,7 @@ Read relevant `docs/` before working on the matching area; update docs when the 
 - `docs/board-render-analytics.md` — the classic-vs-Boardsesh / glow-falloff A/B event contract (issue #2202): the seven events (incl. the paired board-look step events), the common-props builder, the `climb-view-session.ts` state machine, the stratification rule (never pool across `board_name` or `glow_falloff_source`), and the PostHog experiment setup steps
 - `docs/logging.md` — backend structured logger (winston)
 - `docs/crowdsourced-qa.md` — the PR test-plan + risk gate (`@boardsesh/pr-body`, `pr-test-plan.yml`), and the tester loop that turns it into `qa-approved` / `qa-declined` labels
+- `docs/scheduler.md` — the Railway cron scheduler (`packages/scheduler`): which job runs on Vercel vs the scheduler, env vars, Railway setup, cutover order and runbook
 - `docs/db-connectivity.md` — Postgres connect retries (what is retried and why it can't double-execute a write), the retry budgets, and the `/health` vs `/health/db` split
 - `docs/og-climb.md` — backend-served climb OG share cards (`GET /og/climb`: caches, env vars, timings)
 - `docs/cloudflare.md` — the boardsesh.com Cloudflare zone: config-as-code (`vp run cf:apply`), token scopes/secrets, CI auto-apply, og edge caching
--- a/Dockerfile.sync
+++ b/Dockerfile.sync
@@ -1,11 +1,14 @@
-# Sync CLIs - runs any sync CLI (kilter/aurora/moonboard) directly via tsx.
+# Long-lived Node CLIs - runs any sync CLI (kilter/aurora/moonboard) or the cron
+# scheduler directly via tsx.
 # Sibling of Dockerfile.backend. The actual command is supplied by the container
-# `command:` (e.g. docker-compose), so one image serves every sync CLI. kilter
-# and aurora expose a long-running `daemon`; moonboard is locations-only (no
-# daemon subcommand):
+# `command:` (e.g. docker-compose, or Railway's custom start command), so one
+# image serves every CLI. kilter and aurora expose a long-running `daemon`;
+# moonboard is locations-only (no daemon subcommand); the scheduler's `start`
+# runs the cron loop and a /health server (see docs/scheduler.md):
 #   node --import tsx packages/kilter-sync/src/cli/index.ts daemon
 #   node --import tsx packages/aurora-sync/src/cli/index.ts daemon
 #   node --import tsx packages/moonboard-sync/src/cli/index.ts locations
+#   node --import tsx packages/scheduler/src/cli/index.ts start
 FROM node:22-alpine
 
 WORKDIR /app
--- a/docs/branch-deploys.md
+++ b/docs/branch-deploys.md
@@ -1523,7 +1523,8 @@ Backend-affecting paths are defined in two places that must stay in sync:
 | `packages/kilter-sync/`                       | Kilter Grips user/catalog sync and public location sync                       |
 | `packages/location-sync/`                     | Shared public `gyms` / `user_boards` location writer                          |
 | `packages/moonboard-sync/`                    | MoonBoard public location sync CLI                                            |
-| `packages/web/vercel.json`                    | Build command (migration skip), cron definitions                              |
+| `packages/scheduler/`                         | Cron jobs moved off Vercel — see `docs/scheduler.md`                          |
+| `packages/web/vercel.json`                    | Build command (migration skip), remaining Vercel cron definitions             |
 | `.github/workflows/branch-deploy.yml`         | Build images + trigger Ansible deploy on PR open/sync                         |
 | `.github/workflows/branch-deploy-cleanup.yml` | Trigger Ansible cleanup + GHCR delete on PR close                             |
 | `.github/workflows/branch-deploy-sweep.yml`   | Trigger Ansible sweep on push to main / daily                                 |
--- a/docs/scheduler.md
+++ b/docs/scheduler.md
@@ -126,7 +126,10 @@ not firing; check the container is actually running and its clock is sane.
 **A job is failing.** `lastError` carries the HTTP status and a truncated body.
 401 → the two `CRON_SECRET`s have drifted apart. 403 with an HTML body → a
 Vercel WAF/bot rule is blocking Railway egress. 5xx → the route itself; look at
-the web logs. A request that never reached the app reads
+the web logs. A 502 or 503 is retried once after 2s before it counts as a
+failure (a deploy swap or a cold instance clears in seconds); 504 is not, since
+the request reached the route and retrying would stack a second run on the
+first. A request that never reached the app reads
 `fetch failed: <cause>` — `ECONNREFUSED`/`ECONNRESET` means the connection was
 refused or dropped (egress blocked at the network layer rather than by a WAF
 page), `ENOTFOUND` means `BOARDSESH_WEB_URL` is wrong or DNS is broken.
--- a/docs/scheduler.md
+++ b/docs/scheduler.md
@@ -56,7 +56,7 @@ to UTC.
 | `CRON_SECRET`             | yes      | —                           | **Copy** the Vercel project value, don't regenerate: the remaining Vercel crons authenticate with the same secret. Rotating it means updating both places at once. |
 | `BOARDSESH_WEB_URL`       | no       | `https://www.boardsesh.com` | Same env name the backend uses (`packages/backend/src/lib/web-revalidate.ts`).                                                                                     |
 | `PORT`                    | no       | `8080`                      | Health server.                                                                                                                                                     |
-| `SCHEDULER_DISABLED_JOBS` | no       | —                           | Comma-separated job names to leave unscheduled — an env-flip kill switch, no redeploy needed. `run <job>` still works on a disabled job.                           |
+| `SCHEDULER_DISABLED_JOBS` | no       | —                           | Comma-separated job names to leave unscheduled. Read once at startup, so set it and restart the service — no code change, no image rebuild. `run <job>` still works on a disabled job. |
 
 A missing `CRON_SECRET` throws at startup, so a misconfigured service
 crash-loops loudly instead of 401ing quietly at 05:00.
@@ -123,7 +123,10 @@ not firing; check the container is actually running and its clock is sane.
 **A job is failing.** `lastError` carries the HTTP status and a truncated body.
 401 → the two `CRON_SECRET`s have drifted apart. 403 with an HTML body → a
 Vercel WAF/bot rule is blocking Railway egress. 5xx → the route itself; look at
-the web logs.
+the web logs. A request that never reached the app reads
+`fetch failed: <cause>` — `ECONNREFUSED`/`ECONNRESET` means the connection was
+refused or dropped (egress blocked at the network layer rather than by a WAF
+page), `ENOTFOUND` means `BOARDSESH_WEB_URL` is wrong or DNS is broken.
 
 **A job is stuck.** A tick whose predecessor is still in flight is skipped and
 warned, never queued, so a slow run can't stack up. Each run is also bounded by
--- a/docs/scheduler.md
+++ b/docs/scheduler.md
@@ -1,0 +1,127 @@
+# Scheduler service (`packages/scheduler`)
+
+A long-lived Node container that fires Boardsesh's scheduled jobs, taking them
+off Vercel crons one at a time. Part of the Phase 0a hosting move (#1859 →
+#1860 → #1874).
+
+## What it is (and isn't)
+
+The scheduler **triggers** existing `/api/internal/*` routes over HTTP with
+`Authorization: Bearer $CRON_SECRET`. It does not reimplement the jobs.
+
+That's not laziness — two of the three job families can't run outside Next:
+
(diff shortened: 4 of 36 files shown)


marcodejongh:
Thanks — two of the three are now fixed in `86dc00f`.

**Health server always 200 → fixed, but not the way suggested.** Returning 503
from `/health` would make Railway restart the container, and a restart cannot
fix a rotated `CRON_SECRET` or a WAF rule — it would just loop, and each loop
wipes the `lastError` that says which of the two it is. So the endpoints are
split the way `packages/backend` splits `/health` from `/health/db`:

- `GET /health` — liveness. 200 while the process is up (what Railway polls).
  The body now carries `status: 'degraded'` and `degraded: true` so it's still
  visible to a human or a log scrape.
- `GET /health/jobs` — job health. 503 when a scheduled job's last run failed,
  200 otherwise. Point an alert at this one.

Four new tests cover it, including "stays healthy before the first tick" (so a
fresh container isn't marked degraded on boot) and "a disabled job's failure
doesn't count".

**DST spring-forward → documented** in `cron/scheduler.ts` and
`docs/scheduler.md`: fall-back is safe (the repeated hour yields the same
minute key, so the job runs once), spring-forward skips a job scheduled inside
the gap, and that's the reason the registry pins everything to UTC.

**Splitting the `vercel.json` removal → keeping it here, deliberately.** The
drift guard in `registry.test.ts` fails when a path appears in both
`vercel.json` and `JOBS`, which is exactly the state a split would create — so
splitting means reworking the guard that exists to prevent this class of
mistake. The merge gate is the draft status instead: this PR stays a draft
until the Railway service is up and `run cleanup` has returned 200 against
production. I've put that at the top of the description too.

marcodejongh:

marcodejongh:
Part of the kill-Vercel epic #4648 (Phase 2). The merge gate is unchanged: the Railway `scheduler` service must exist first, or the retention crons stop. The remaining six `vercel.json` crons follow in #4654 using the pattern this PR establishes.

marcodejongh:
Finishing pass is pushed and CI is green (40 pass, 11 skipped, 0 fail). Two of the older review points are now closed:

- **`cli/index.ts` shutdown can hang** — fixed. `healthServer.stop()` now has a `.catch` that logs and exits 1. The `shuttingDown` latch was the real hazard: with the latch set and no `.catch`, a rejected close means the second SIGTERM is ignored and nothing calls `process.exit`, so Railway SIGKILLs at the end of the stop grace period.
- **`runJob` bypasses the in-flight guard** — kept, documented. An operator running `scheduler run <job>` on a stuck job needs the run to happen, not to be silently swallowed. `JobDefinition.run` now states the contract (a job must tolerate overlapping itself) and why `cleanup` satisfies it: delete-by-age in batches, so the second pass finds nothing left.

**`"start": "node dist/cli/index.js start"` — leaving it.** It is the same script `@boardsesh/kilter-sync`, `@boardsesh/aurora-sync` and `@boardsesh/moonboard-sync` all carry, and `dist/` does exist locally after `vp run build:scheduler` (which `vp run typecheck:scheduler` depends on). Dropping it from this one package would make the scheduler the odd workspace out rather than remove a real hazard; if the convention is wrong it is wrong for all four and belongs in its own change.

Also in this pass: the Bun-era `bunx tsx` start command is gone from the runbook, the PR body and the CLI banner (the image has neither Bun nor `vp` — it is `node --import tsx packages/scheduler/src/cli/index.ts <cmd>`); the deploy-input fixture now declares `packages/scheduler`, which `main`'s new base-fixture `Dockerfile.sync` had turned into 15 red tests; and a 502/503 from the edge is retried once after 2s so a daily job that lands in a deploy swap is not written off for 24 hours.

Verified against a real image, not just the CLI: built `.docker-context/sync` locally, and in the container `list` prints `cleanup 0 5 * * * UTC /api/internal/cleanup`, `start` serves `/health` 200 and `/health/jobs` 200, an unknown path 404s, and SIGTERM exits 0.

Merge gate is unchanged — the Railway `scheduler` service has to exist and `run cleanup` has to succeed against production first.

marcodejongh:
In-context review (Fable) of the post-rebase state — no blocking findings.

- The 502/503 single retry is right-sized: 504 stays excluded (the route is still running; a retry would double-fire), the wait is abort-aware and lives inside the job's `timeoutMs`, and the shutdown path now exits 1 on a failed `stop()` instead of idling into Railway's SIGKILL.
- `registry.test.ts` and `rest-surface-inventory.test.ts` agree with today's `vercel.json` (five prewarm-heatmap + profile-percentiles stay on Vercel; `cleanup` moves here). The sync image context carries `packages/scheduler`, and the built image's `list` / `start` / `/health` / `/health/jobs` behaved as documented.
- Start command for the Railway service is now `node --import tsx packages/scheduler/src/cli/index.ts start` everywhere (Dockerfile.sync convention).

**Merge gate unchanged and deliberate:** the Railway `scheduler` service (image `ghcr.io/boardsesh/boardsesh-sync:production`, that start command, `CRON_SECRET` copied from Vercel, `BOARDSESH_WEB_URL=https://www.boardsesh.com`, `PORT=8080`, healthcheck `/health`, one replica) must exist and `… run cleanup` must succeed against production before this merges — merging first silently stops the retention cron. #4654 (the other six crons + Sentry monitors) follows once `cleanup` has run cleanly for a day.

marcodejongh:
Infra update: the Railway `scheduler` service now exists — `boardsesh-scheduler` (`742c49ca-0402-4ffa-ae7e-d0201c71692c`), variables `CRON_SECRET` (reference to `boardsesh-backend.CRON_SECRET` — needs to equal the Vercel value, maintainer to confirm), `BOARDSESH_WEB_URL=https://www.boardsesh.com`, `PORT=8080`, start command `node --import tsx packages/scheduler/src/cli/index.ts start`, healthcheck `/health`, `ON_FAILURE` ×5, 1 replica.

Merge-gate adjustment: the pre-merge `run cleanup` verification can't run from Railway yet because `ghcr.io/boardsesh/boardsesh-sync:production` won't contain `packages/scheduler` until this PR merges and `sync-deploy.yml` rebuilds it. So the sequence is: merge → sync image rebuilds → attach the image to the service → `/health` 200 → the first 05:00 UTC tick reports on `/health/jobs` (503 = the WAF/secret problem the original gate was for; retention delete-by-age catches up on the next run, so a day's gap is harmless).

PR #4274, opened 2026-08-11
fix(db): stamp the dev-db migration ledger with journal timestamps, not the build clock

### LINKED ISSUE #4211: dev-db image bakes migration-ledger created_at from image-build time, silently skipping newer PRs' migrations
## What's wrong

`packages/db/docker/Dockerfile.dev-db` stamps each baked migration-ledger row with **image-build wall-clock time** rather than the migration's own journal `when`:

```dockerfile
INSERT INTO __drizzle_migrations (hash, created_at) VALUES ('$hash', $(date +%s)000)
```

Drizzle's migrator is a single high-water mark, not a reconciliation: `PgDialect.migrate()` reads `max(created_at)` from `drizzle.__drizzle_migrations` **once, before the loop**, and applies only journal entries whose `when` is strictly greater.

So every row in the image carries a `created_at` of "whenever this image happened to be built". Any PR whose migration was generated **before** the latest dev-db image build has a `when` below that high-water mark, and drizzle skips it — silently and permanently.

## What it looks like when it fires

The `test-location-sync-integration` job reds in its "Apply the full migration journal" step (`VERIFY_MIGRATION_JOURNAL=1`), reporting a migration that "will never re-apply on its own":

```
error: Migration journal verification failed: 1 of 190 journal migrations has no row
in drizzle.__drizzle_migrations (189 rows present). Missing: 0189_moonboard_8c_grades.
```

The verifier is doing its job correctly — it is reporting a genuine skip. The bug is upstream of it: nothing was wrong with the migration.

## Why it's worth fixing rather than living with

The failure is **indistinguishable from the real regression the check exists to catch** (the `0129_numerous_star_brand` incident, where `location_sync_gym_sources` was never created and the first Kilter location sync died days later). Both print the same message. So each occurrence costs a real investigation, and the more often it cries wolf the more likely someone waves through the one that matters.

It also fires unpredictably from an author's point of view — a PR is green, an unrelated migration lands on main, the image rebuilds, and the same untouched PR goes red. The current workaround is to restamp `when` (via `vp run db:renumber`, which sets `when = max(main's max when, Date.now())`), but that is a treadmill: it has to be redone every time the image is rebuilt ahead of a PR.

## Observed today

Confirmed empirically across a stack of six migration PRs:

- #4025 carried `0189_moonboard_8c_grades` stamped `1785543542675` (2026-08-01T00:19Z) → **red**, reported missing.
- #4058 and #4052 carried the **byte-identical** SQL stamped `1785638760607` (2026-08-02T02:46Z) → **green** on the same job, same pinned image digest (`ghcr.io/boardsesh/boardsesh-dev-db@sha256:ce32f2a7…`).

Same file, same image, opposite outcomes — the only variable was `when`. That puts the image's baked high-water mark between those two timestamps, consistent with the digest having been built on 2026-08-01.

## Suggested fix

Stamp `created_at` from each journal entry's own `when` when baking the ledger, so the image's high-water mark equals main's newest migration at build time — which is what drizzle's comparison actually means. Then a PR's freshly generated migration is above it by construction, and the verifier only fires on genuine skips.

Worth checking at the same time whether `packages/db/docker/Dockerfile.dev-db` should bake the ledger at all, versus applying the journal normally during the image build so the hashes and timestamps come from drizzle itself. `docs/db-migrations.md` already warns that a re-derived hash drifts on encoding, BOM, line endings, or a drizzle change — the same argument applies to a hand-stamped `created_at`.

## Notes

- Not urgent for correctness of production: this is CI/dev-image only. Production's ledger is written by drizzle itself.
- Related context in `docs/db-migrations.md` — "Order comes from `when`, not from the number" and the `VERIFY_MIGRATION_JOURNAL=1` section.
- Bumping the `-v1-` style salt or repinning the digest does not fix this; it just moves the threshold.

marcodejongh:
## Summary

Closes #4211.

The `boardsesh-dev-db` image applies the migration journal itself, in a psql loop, and wrote drizzle's ledger with the image's **build wall clock** (`$(date +%s)000`) instead of each journal entry's `when`. Drizzle's applier is a single high-water mark — `PgDialect.migrate()` reads `max(created_at)` once and applies only entries whose `when` is strictly greater — so the build timestamp becomes the mark, and the mark only ever moves up.

What that eats is not the newest migrations. A migration generated today has a `when` above any image built yesterday, so it still applies — which is why this reads as intermittent rather than broken. It eats the entries whose `when` sits *below* the build clock and that the image did not itself apply: a branch's migration, written before the image was built and applied to that database afterwards. On the two e2e jobs, which run `boardsesh-dev-db:latest` and therefore get a new image on every db-touching merge to `main`, that is most open PRs. The migration is skipped with no error of any kind: `vp run db:migrate` reports success, the table never appears, and `VERIFY_MIGRATION_JOURNAL=1` (correctly) calls it a gap.

It is not hypothetical. On this machine's dev DB, 92 of ~190 ledger rows carry `1778588631000`-ish build timestamps against journal `when`s a year and a half earlier; the `test-location-sync-integration` run on this PR repaired 189 rows stamped `1785545390000` against a journal whose newest `when` is `1785641394370`.

Three parts, because a Dockerfile fix alone reaches nobody:

1. **The source.** `Dockerfile.dev-db`'s loop now reads `when` alongside `tag` and writes it into both ledger tables. The build then asserts that the ledger's high-water mark equals the journal's newest `when` and fails if it doesn't — one psql call that makes this regression unshippable.
2. **The repair.** `vp run db:normalize-ledger` rewrites `created_at` to each entry's `when` on a database already on disk. That is exactly the value drizzle writes itself (it inserts `folderMillis`), so it's a no-op on any drizzle-managed database — and it refuses a non-local target without `--force`, so it can't be mistaken for a production ledger tool. Pairing is positional within a hash, not a `Map<hash, when>`: byte-identical `.sql` files share a hash and a map would stamp both rows with the first entry's `when`.
3. **The wiring.** `vp run db:up` runs the repair between the tracker sync and the pending-migration apply; so do the CI jobs that boot the image (`test-location-sync-integration`, and both e2e jobs before `drizzle-kit migrate`). Ordering is load-bearing and is now asserted — after `db:migrate` the repair is useless, since the migrations it would have unblocked are already skipped. `db:up` pins all three names `getScriptDatabaseUrl()` reads (`DB_URL`, `DATABASE_URL`, `POSTGRES_URL`) to the container it just started, so a `DB_URL` in `.env.local` can't redirect the repair at someone else's database. And `dev-db-image-ledger.test.ts` sweeps every workflow job that boots the image and applies migrations rather than trusting the audit — `db-migration-renumber.yml#renumber` is the one exemption, because `db-renumber-migration.ts` stamps `nextWhen(maxWhen(mainJournal), Date.now())` at run time and so can never land under an older image's clock.

The pinned digest in `ci.yml` is deliberately **not** bumped: the corrected image only exists once this merges and `dev-db-docker.yml` rebuilds. The normalise step is what makes those jobs deterministic meanwhile; repinning is a follow-up (and must also update the hard-coded digest in `scripts/__tests__/ci-location-sync-workflow.test.ts`).

One judgement call worth a reviewer's opinion rather than a guess: the follow-up could be a plain digest repin, or the larger change the issue floats — replacing the image's hand-rolled psql loop with drizzle's own migrator. The loop deliberately continues past per-migration errors and the image has no monorepo workspace install, so swapping appliers risks the whole build for a problem the `when` stamp plus the build assertion already close. Left out of scope here.

## Release Notes

## Test plan

- `vp run typecheck:db` — clean.
- `vp test run --project scripts` — 725 passed, including the new `planLedgerTimestampRepairs` cases (no-op on a correct ledger, full repair on a build-stamped one, duplicate-hash entries paired in `id` order, unmatched rows untouched, short ledger repairs only what's there) and the `Dockerfile.dev-db` text guard.
- `MIGRATION_JOURNAL_DB_URL=… vp run test:db:migration-journal` — 16 passed against a real Postgres, including the new end-to-end regression: stamp the ledger with a build clock → drizzle skips a later migration and the gate throws → run the repair → the migration applies, the gate passes, and a second pass plans nothing.
- `vp run check:db-migrations` — 190 migrations, journal consistent.
- `vp check` — no new errors (30 errors / 327 warnings vs. 30 / 325 on `origin/main`; the two extra are the `no-floating-promises` warning every `it()` in that node:test file already carries).
- Real dry run against the local dev DB: `DATABASE_URL=… vp run db:normalize-ledger -- --dry-run` planned 92 repairs with sensible `from → to` pairs and touched nothing.

Not covered: the image build itself (six APK downloads, pgloader, the MoonBoard import — tens of minutes, network-heavy). Coverage there is the text guard plus the in-build high-water-mark assertion; the first real proof is the post-merge `dev-db-docker.yml` run, which is worth watching.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -805,10 +805,10 @@ jobs:
       - name: Install dependencies
         run: bun install --frozen-lockfile
       # The pinned image's ledger was stamped with the image build clock, not
-      # each journal entry's `when` (#4211). That is a high-water mark drizzle
-      # can never clear, so without this repair the next step silently skips
-      # every migration written after the image was built — including the one a
-      # PR is adding. Must stay ahead of db:migrate.
+      # each journal entry's `when` (#4211). That mark only moves up, so without
+      # this repair the next step silently skips any journal entry whose `when`
+      # predates the pinned image's build — including a PR's own migration when
+      # the branch is older than the pin. Must stay ahead of db:migrate.
       - name: Normalize the dev-db image migration ledger
         env:
           DATABASE_URL: postgres://postgres:password@localhost:5432/main
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -804,6 +804,15 @@ jobs:
           cache: true
       - name: Install dependencies
         run: bun install --frozen-lockfile
+      # The pinned image's ledger was stamped with the image build clock, not
+      # each journal entry's `when` (#4211). That is a high-water mark drizzle
+      # can never clear, so without this repair the next step silently skips
+      # every migration written after the image was built — including the one a
+      # PR is adding. Must stay ahead of db:migrate.
+      - name: Normalize the dev-db image migration ledger
+        env:
+          DATABASE_URL: postgres://postgres:password@localhost:5432/main
+        run: bun run --filter=@boardsesh/db db:normalize-ledger
       - name: Apply the full migration journal
         env:
           DATABASE_URL: postgres://postgres:password@localhost:5432/main
--- a/.github/workflows/e2e-tests.yml
+++ b/.github/workflows/e2e-tests.yml
@@ -323,9 +323,11 @@ jobs:
 
       # The dev-db image's ledger was stamped with its build clock, not each
       # journal entry's `when` (#4211). drizzle-kit migrate reads max(created_at)
-      # once and applies only entries above it, so without this repair a PR's new
-      # migration silently never runs here and surfaces as an unrelated-looking
-      # app failure. Must stay ahead of the migrate step.
+      # once and applies only entries above it, so a PR's migration is skipped
+      # whenever it was generated before the image was built — and this job runs
+      # `:latest`, rebuilt on every db-touching merge to main, so that is most
+      # open PRs. It surfaces as an unrelated-looking app failure rather than a
+      # migration error. Must stay ahead of the migrate step.
       - name: Normalize the dev-db image migration ledger
         run: bun run --filter=@boardsesh/db db:normalize-ledger
         env:
@@ -582,8 +584,9 @@ jobs:
           rm built-packages.tar
 
       # Same repair as the shard job above: the image's ledger high-water mark is
-      # a build timestamp, so drizzle-kit migrate would skip a PR's new migration
-      # (#4211). Must stay ahead of the migrate step.
+      # its build timestamp, so drizzle-kit migrate would skip a PR's migration
+      # whenever the branch predates the `:latest` image (#4211). Must stay ahead
+      # of the migrate step.
       - name: Normalize the dev-db image migration ledger
         run: bun run --filter=@boardsesh/db db:normalize-ledger
         env:
--- a/.github/workflows/e2e-tests.yml
+++ b/.github/workflows/e2e-tests.yml
@@ -321,6 +321,16 @@ jobs:
           tar -xf built-packages.tar
           rm built-packages.tar
 
+      # The dev-db image's ledger was stamped with its build clock, not each
+      # journal entry's `when` (#4211). drizzle-kit migrate reads max(created_at)
+      # once and applies only entries above it, so without this repair a PR's new
+      # migration silently never runs here and surfaces as an unrelated-looking
+      # app failure. Must stay ahead of the migrate step.
+      - name: Normalize the dev-db image migration ledger
+        run: bun run --filter=@boardsesh/db db:normalize-ledger
+        env:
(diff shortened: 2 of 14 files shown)


marcodejongh:

marcodejongh:

Conflicting: packages/db/docker/Dockerfile.dev-db — this PR patches an inline shell migration-apply loop, and main has since wired in a dedicated boardsesh-apply-drizzle-migrations helper script for the same #4211 area. Worth checking whether that already covers this fix before spending time on a manual merge. Also touches nearly the same file set as blocked #4302 — likely worth reconciling together.

PR #4276, opened 2026-08-11
refactor(web): move the profile domain from REST to GraphQL

### LINKED ISSUE #1884: Migrate /api/internal/* to GraphQL
Parent: #1861. Part of #1859.

Migrate `/api/internal/*` data ops to GraphQL. Most resolver domains exist (14 under `packages/backend/src/graphql/resolvers/`); the plan estimates ~8 net-new resolvers needed.

Audit all 44 `route.ts` files in `packages/web/app/api/`, identify which are non-auth and non-cron, and check for an existing GraphQL equivalent. File a sub-task per net-new resolver if helpful.

marcodejongh:
## Summary

Epic #1884 wants `/api/internal/*` off REST and onto GraphQL. This is the **profile slice** — one reviewable domain, done end to end.

`GET/PUT /api/internal/profile` and `GET /api/internal/profile/[userId]` are deleted. Their five callers now read `Query.profile`, `Mutation.updateProfile` and `Query.publicProfile`. The biggest win is `/profile/[user_id]`'s server render: it was the last piece of profile UI running drizzle queries straight from Next.js, so the frontend is one step closer to not needing `DATABASE_URL` at all (`docs/branch-deploys.md`).

The resolvers couldn't just take over — they were missing three fields — so the backend grew first:

**Schema (additive only).** `UserProfile` gains `instagramUrl`, `hasPassword`, `linkedProviders`; `UpdateProfileInput` gains `instagramUrl`. No field renamed or removed, every input field still optional. A frozen-document test in `operations-schema-validation.test.ts` pins the pre-#1884 selection sets that App Store and TestFlight binaries still send, so this can't stop being additive by accident.

**One round trip, still.** `hasPassword` and `linkedProviders` ride correlated subqueries next to the existing `FAVORITE_COUNT_SUBQUERY` rather than the two extra parallel queries the REST route ran. Both resolvers stay the single select that #3603 collapsed them to, and the select shape plus its mapper now live in one module so the query and the mutation can't drift.

**Nullable inputs.** Omitting a field leaves it alone; `null` clears it. The settings form has always sent `trim() || null`, and the old `.optional()`-only schema would have rejected that — this was the likeliest silent regression in the whole slice, so there are tests on both the resolver and the form.

### Two behaviour changes worth calling out

**1. `updateProfile` now writes `users.name`.** This is parity with the REST route being deleted — `users.name` is what the NextAuth session renders, so without the mirror a rename leaves a stale name in the web header until the next sign-in. It is a **live change for already-shipped mobile binaries**: an Edit Profile save now also moves the session-derived name. `publicProfile` already coalesced `displayName || users.name`, so nothing renders differently; it's the write that's new. The alternative (leave `users.name` alone, teach the web header to read `displayName` over GraphQL) is more web churn for no mobile-visible difference — happy to switch if you'd rather not touch shipped behaviour.

**2. `UserSmartCard`'s board chips changed source.** They listed the climber's linked Aurora accounts, which meant a private credentials read behind a public card. They now list the boards that climber actually logs sends on, off the stats query the card already fetches. Same chips for anyone with ticks; someone with a linked account but zero ticks now shows none.

### Also in here

- The client reads wait for the ws-auth token before firing. The resolvers key off the bearer token, so racing it resolves `profile` as null (empty settings form) and `isFollowedByMe` as false — a "Following" button flipping back to "Follow" on every refetch.
- The web profile view-model flattens to match `publicProfile` field for field. The resolver already coalesces onto the NextAuth name and image, so the second `|| profile.name` fallback at each call site was dead.
- SSR degrades to the 404 page on a backend blip rather than 500ing an indexable page.
- Mobile's `useUpdateProfile` now merges into the `['profile']` cache instead of replacing it. Tightening the response types (`Pick`, not the whole `UserProfile`) surfaced that the mutation's narrower selection was blanking out `createdAt` and `favoriteCount` until the invalidation refetched.
- Docs updated: `MIGRATION_PLAN.md`, `docs/branch-deploys.md`, `docs/ui/18-settings.md`, `docs/web-reposition.md`.

Deliberately **not** in here: favorites (PR #4256 is mid-flight over those resolvers), controllers, hold-classifications, climb-redirect, and the cron routes. Each is its own slice.

Closes #1884

## Release Notes

- Your profile now loads and saves through the same backend the app uses, so name and avatar changes show up everywhere straight away.

## Test plan

Validated with `vp` only — no dev server, no build shortcuts.

- `vp check` — format clean; no new lint diagnostics in any touched file.
- `vp run typecheck` per package: `backend`, `web`, `mobile`, `shared` — all clean.
- `vp run test:web` — **482 files / 5915 tests, all passing** (full suite).
- `vp test run --project backend` — **full suite, exit 0.**
- `vp run test:mobile` — **616 files / 5389 tests, all passing** (full suite).

New/updated tests:

- `resolvers/users/__tests__/queries.test.ts` — `instagramUrl` / `hasPassword` / `linkedProviders` map through; a NULL `array_agg` still yields `[]`; asserts the resolver stays **one** select.
- `resolvers/users/__tests__/mutations.test.ts` + `__tests__/update-profile.test.ts` — `instagramUrl` persists and clears via null; an invalid one is rejected as a client-safe error; `displayName` mirrors onto `users.name` (and nulls it when cleared); an avatar-only save leaves `users.name` untouched.
- `__tests__/operations-schema-validation.test.ts` — the shipped-binary guard described above.
- `settings/__tests__/settings-profile-graphql.test.tsx` (new) — the form waits for ws-auth; sends `null` not `''` when a field is cleared; seeds the Set-password section from `hasPassword`/`linkedProviders`; renders the resolver's error message on failure.
- `party-manager/__tests__/party-profile-context.test.tsx` — no query before the token resolves; custom name/avatar still win over the session values.
- `profile/[user_id]/hooks/__tests__/use-profile-data.test.tsx` — asserts the `GET_PUBLIC_PROFILE` request and that `isFollowedByMe` survives; a null profile surfaces `ProfileNotFoundError` without retrying.
- `app/__tests__/profile-rest-route-retired.test.ts` (new) — no source file under `packages/web/app` may reference the deleted paths again (scoped so the live `profile-percentiles` cron route is unaffected).

**Not yet exercised by hand.** Manual QA notes are written to `.boardsesh/qa-notes.md`; the flows that most want a human are the cleared-display-name save, the Follow button surviving a refetch on someone you already follow, and the Set-password section's state.


marcodejongh:

marcodejongh:

Conflicting: packages/backend/src/graphql/resolvers/users/__tests__/queries.test.ts, resolvers/users/mutations.ts, resolvers/users/queries.ts, plus generated GraphQL types.

Issue #4190, opened 2026-08-03
Resolve the five duplicate-row serials from the shared-serial investigation


marcodejongh:
Part of #4372. **Blocked by:** coordinate with open PR #3422 (duplicate-serial board consolidation, carries a db-migration) before writing any code — verify overlap first. Gym merges go through the founder's /admin/gym-duplicates flow.

## Context

Issue #4190 investigated the 15 board serials carried by two gym listings and found three distinct problems, only two of which are dedup-gate work (DOC-03, DB-07):

1. **Duplicate board rows inside one gym** — serials `K000323` and `K000217` (Beargrass, 0 m apart) and `751807` (International College Hong Kong: `user_boards` ids 660 + 661, both gym 607) are catalog-sync artifacts producing pointless disambiguation prompts.
2. **Duplicate gym rows** — serial `751879` ("Stone Age" vs "Salle d'Escalade - Stone Age", identical address to the metre) is one Belgian gym listed twice; serial `75692` (Høyt Under Taket Kokstad / Studio Vertikal Klatresenter, 35 m apart) needs a human eye before any merge.

The remaining 9 genuine-reuse serials (plus placeholder `12345`) are NOT duplicates and must be left alone — that is #4187 territory, not dedup.

## The work

**Agent PR scope:** collapse the three duplicate board rows (or confirm PR #3422 already covers them), list the no-action serials, and pin the never-404 contract with a test. Printed QR codes exist in the field; a merged gym URL must keep resolving (DOC-03).

**The two duplicate gym pairs are founder steps, not agent work** (gym merges are founder-locked to the human /admin/gym-duplicates UI — DOC-03 — and prod data changes need the founder's sign-off): see the "Human steps" block below. The issue closes only when both blocks are done.

## Acceptance criteria

- [ ] Duplicate board rows behind `K000323`, `K000217`, and `751807` are collapsed to one row each — either confirmed covered by PR #3422 after its merge (link the verification query) or fixed here without duplicating #3422's approach
- [ ] The 9 genuine-reuse serials and placeholder `12345` are explicitly listed as no-action in the PR description
- [ ] An automated test asserts a merged gym URL resolves via the 308 redirect (`packages/web/app/gym/[gym_slug]/page.tsx`) and never 404s
- [ ] Any data change is expressed as a migration generated by `bunx drizzle-kit generate` (never hand-written SQL); expect `vp run db:renumber` — 8 open PRs carry the db-migration label
[...]

PR #4277, opened 2026-08-11
fix(backend): retire the gym listing a stray-board attach empties

### LINKED ISSUE #4188: [Bug] Avoid orphan gyms when the last board gets taken
#3843 allows taking boards from one listing and combining them into a single gym that the user has control over.

Taking the last board leaves an empty gym, which was most likely intended as a merge operation between an auto-generated less-correct listing than the user's "owned" one (in which case it should either be auto-deleted, and/or the user should be explicitly asked if the other listing is an invalid gym or the board was just incorrectly attributed).

Leaving an empty gym then results in needless confusion when people search for new boards by gym, because they can find a gym that seems correct but has no boards in it (in which case the correct gym likely shows up as a duplicate, which is part of the problem merging boards into a single gym listing is supposed to be solving).

I'm also unsure on the current permission checks - is it possible to "steal" a gym that someone else owns? I managed to take an auto-generated listing from kilter, but I'm unsure if that's intentionally valid when the gym is marked as unclaimed, or if it shouldn't be possible at all (without first going through the claiming process, to own both gym listings) and is an oversight from the initial PR.

marcodejongh:
## Summary

Taking the **last** board off an auto-synced gym listing left that listing live and public with nothing on it, so it kept coming back in gym search as a plausible-looking duplicate of the gym the board had just moved to — the exact duplicate problem #3843's attach flow was meant to solve. Closes #4188.

Three parts:

**The attach now sticks.** `attachBoardToGym` sets `sync_frozen_at` on the board it re-points. `packages/location-sync/src/upsert.ts` upserts catalog boards with `set: { gymId: excluded.gym_id }` guarded only by `setWhere: isNull(userBoards.syncFrozenAt)`, so without the freeze the next sync run put the board straight back on the synced listing and re-populated it. `updateBoard` and `updateGym` already do this on any human edit.

**The emptied listing gets folded, not deleted.** When an attach leaves its source listing with zero live boards, the listing is folded into the target gym through `mergeGymsIntoCanonical` — the same primitive the admin duplicates queue uses. Soft-delete plus `merged_into_gym_id`, so old uuids and slugs still resolve to the survivor; the listing's `location_sync_gym_sources` alias moves across, which is what stops the sync re-minting the same empty listing; and a `gym_merge_audit` row (`action: 'merged'`, `performed_by` = the attaching user) records exactly what moved so an admin can reverse it. No row is destroyed.

The fold is deliberately narrow. All of these must hold or it silently does nothing: source is SYSTEM-owned, live, un-merged, has zero live boards, no pending or approved claim, no members, no live kiosk, and has coordinates; target has coordinates and is **not** SYSTEM-owned. A refusal is a debug log and a plain return — the board move is already committed, so a skipped or failed fold never turns a successful attach into a GraphQL error.

**The Boards tab warns first.** `StrayBoard` gains `isLastBoardAtCurrentGym` (one grouped count over the listings in the result, not a query per row). A flagged stray shows a "Last board on that listing" chip and its **Add to gym** tap goes through a confirm dialog naming the listing; everything else keeps the one-tap path. New strings in en-US, es, fr and de.

On the issue's second question — "is it possible to steal a gym?" — the existing gates hold up. `strayBoardsForGym` and `attachBoardToGym` both require edit access on the **target**; NEARBY candidates only come off a SYSTEM listing or an unlinked board, and only for SYSTEM-owned or the viewer's own boards, with `hide_location` boards excluded; MERGED_TWIN candidates only exist after an admin merge; and approving a claim reassigns `gyms.owner_id` to the claimant, so a claimed gym's boards stop being strays. One gap remains and is **not** fixed here: a listing with a *pending* claim is still SYSTEM-owned, so a third party can still vacuum its boards mid-claim. This PR refuses to retire such a listing; excluding it from stray candidates entirely wants its own issue.

## Release Notes

- Moving the last board off an auto-generated gym listing now folds that listing into your gym instead of leaving an empty duplicate in gym search — old links still land on the right place.
- The Boards tab tells you when a board is the last one on its listing, and asks before it merges.

## Test plan

- `vp run codegen` — clean, generated files committed.
- `vp run typecheck` — green.
- `vp test run --project backend --reporter=agent gym-stray-boards` — 29/29 green. Eleven new cases: the fold (soft-delete + `is_public=false` + `merged_into_gym_id`), the audit row, the source-alias move, a non-last board leaving the listing live, each refusal gate (pending claim, approved claim, member, live kiosk, SYSTEM target, already-merged twin), the unlinked-board no-op, `sync_frozen_at` on the attached board, and `isLastBoardAtCurrentGym` true/false/unlinked.
  Caveat: this machine's backend worker DBs on :5433 are shared with parallel worktrees, and intermittent runs failed on `users` being truncated mid-test or a TRUNCATE deadlock — always during seeding, never an assertion. Pointing `DATABASE_URL` at a private database name gives a clean deterministic 29/29. `gym-duplicates.test.ts` (the other `mergeGymsIntoCanonical` caller) passes unchanged.
- `vp run test:web` — 5910/5910 across 480 files, including four new `gym-boards-tab` cases (chip renders, confirm gates the mutation, cancel writes nothing, non-flagged strays keep one tap).
- `vp run check:i18n` and `vp run check:i18n:orphans` — OK.
- `vp check` — no new diagnostics in any touched file.
- No migration, no mobile surface, no Bluetooth code.

### DIFF
--- a/packages/backend/src/__tests__/gym-stray-boards.test.ts
+++ b/packages/backend/src/__tests__/gym-stray-boards.test.ts
@@ -117,10 +117,71 @@ const insertGymMember = (gymId: number, userId: string, role: string) =>
     VALUES (${gymId}, ${userId}, ${role}, now())
   `);
 
+const insertGymClaim = (gymId: number, claimantUserId: string, status: string) =>
+  db.execute(sql`
+    INSERT INTO gym_claims (gym_id, claimant_user_id, method, status, created_at, updated_at)
+    VALUES (${gymId}, ${claimantUserId}, 'admin', ${status}, now(), now())
+  `);
+
+const insertGymKiosk = (gymId: number, slug: string) =>
+  db.execute(sql`
+    INSERT INTO gym_kiosks (uuid, gym_id, slug, name, created_at, updated_at)
+    VALUES (${uuidv4()}, ${gymId}, ${slug}, ${'Kiosk ' + slug}, now(), now())
+  `);
+
+const insertGymSourceAlias = (gymId: number, sourceKey: string) =>
+  db.execute(sql`
+    INSERT INTO location_sync_gym_sources (source_key, gym_id, created_at, updated_at)
+    VALUES (${sourceKey}, ${gymId}, now(), now())
+  `);
+
 const boardGymId = async (boardUuid: string): Promise<number | null> => {
   const result = await db.execute(sql`SELECT gym_id FROM user_boards WHERE uuid = ${boardUuid}`);
   const row = Array.from(result as Iterable<{ gym_id: number | null }>)[0];
   return row?.gym_id != null ? Number(row.gym_id) : null;
+};
+
+const boardSyncFrozen = async (boardUuid: string): Promise<boolean> => {
+  const result = await db.execute(sql`SELECT sync_frozen_at FROM user_boards WHERE uuid = ${boardUuid}`);
+  const row = Array.from(result as Iterable<{ sync_frozen_at: Date | null }>)[0];
+  return row?.sync_frozen_at != null;
+};
+
+type GymRetireState = { deletedAt: Date | null; isPublic: boolean; mergedIntoGymId: number | null };
+
+const gymRetireState = async (gymId: number): Promise<GymRetireState> => {
+  const result = await db.execute(sql`
+    SELECT deleted_at, is_public, merged_into_gym_id FROM gyms WHERE id = ${gymId}
+  `);
+  const row = Array.from(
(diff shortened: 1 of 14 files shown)


marcodejongh:
Triage note: this PR closes #4188, which is Tier 0 of the gym-discovery epic #4372 — orphan-gym prevention is one of the two duplicate-creation vectors that must be shut before the public /gyms directory ships (the other is #4190). Linking so the dedup gate's state is visible from the epic.

PR #4285, opened 2026-08-11
fix(sync): make circuit playlist sync idempotent under unique_playlist_climb

### LINKED ISSUE #4023: Aurora playlist sync can hit the unique_playlist_climb index (delete-then-loop-insert)
Follow-up spun out of #3993 / PR #4021 (which fixed the GraphQL `addClimbToPlaylist` path only).

`packages/web/app/lib/data-sync/aurora/user-sync.ts:347` rebuilds a playlist's climbs by deleting the existing rows and then looping inserts, with no `onConflictDoNothing`. That leaves the same `unique_playlist_climb` (`playlist_id`, `climb_uuid`) index exposed if the source list contains a duplicate pair, or if a concurrent `addClimbToPlaylist` lands between the delete and the loop — the insert throws a raw 23505 and the sync run aborts partway, having already deleted rows.

Shape is different from #3993 (no pre-check to race), so it needs its own fix: either dedupe the incoming pairs and add `onConflictDoNothing`, or move the delete + inserts into one transaction.

Not urgent — no reported user report behind this, it was spotted while reviewing #3993.

marcodejongh:
## Summary

Closes #4023

Three Aurora circuit → `playlist_climbs` writers rebuild a playlist by deleting its rows and then inserting, with no conflict handling against `unique_playlist_climb` on `(playlist_id, climb_uuid)` (`packages/db/src/schema/app/playlists.ts:92`):

- `packages/aurora-sync/src/sync/user-sync.ts` — the primary sync daemon
- `packages/web/app/lib/data-sync/aurora/user-sync.ts` — still live via `/api/v1/[board_name]/proxy/user-sync` and `/proxy/login`
- `packages/aurora-sync/src/sync/json-import.ts` — the Aurora JSON export importer (found during the adoption re-review, see below)

Two ways to hit a deterministic `23505`:

1. A circuit payload containing the same `climb_uuid` twice — e.g. the same climb at two angles, which the index cannot distinguish. Fails identically on every retry.
2. A concurrent `addClimbToPlaylist` committing between the delete and an insert. The delete only locks index entries for rows that already existed, so a *newly* added climb can slip in and collide.

**Correction to the issue text:** the delete and inserts already share a transaction, so rows are not left deleted on partial failure. The damage is different, and worse than "not urgent" suggests — see below.

### How bad case 1 actually is

`syncUserData` wraps **every table for one user** in a single `db.transaction` (`packages/aurora-sync/src/sync/user-sync.ts:596`), and `updateUserSyncs` is inside it (`:625`). A 23505 out of the circuits branch therefore rolls back that user's logbook, ascents and bids for the same batch, and `last_synchronized_at` never advances — so the next cycle re-fetches the byte-identical payload and throws again. One duplicated climb in one circuit stops that climber's entire Aurora sync, ticks included, permanently.

It does **not** spread: `syncNextUser` catches per credential (`packages/aurora-sync/src/runner/sync-runner.ts:224`) and `recordSyncFailure` (`:279`) always advances `last_sync_attempt_at`, so a wedged credential rotates to the back of the queue and the rest of the fleet keeps syncing. That is a real difference from the kilter dup-`log_uuid` incident (#3329), where one poison-pill user blocked everyone.

## Changes

**Both user-sync copies (circuits case):**

- Build the full row set first, resolving `climbUuid = climb.climb_uuid || climb.uuid || climb` exactly as before, keeping the `typeof climbUuid === 'string'` guard.
- Dedupe by `climbUuid`, first occurrence wins (what `onConflictDoNothing` would keep anyway).
- Batch-insert with `.onConflictDoNothing({ target: [playlistClimbs.playlistId, playlistClimbs.climbUuid] })`, so a concurrent `addClimbToPlaylist` no-ops instead of aborting the batch.
- **Log every dropped row.** Dedupe is silent data loss — the circuit in Aurora has an entry ours won't — and without a log line there is no way to tell "Aurora never repeats a climb" from "we have been truncating circuits for months". If this fires in prod, widening the index to `(playlist_id, climb_uuid, angle)` is the real fix and needs its own PR.
- `aurora-sync` chunks via the file's existing `processBatches` helper; the web copy uses a local 500-row constant (matching kilter-sync — `aurora-sync`'s file-level `BATCH_SIZE = 100` is the outlier).

**JSON import (`json-import.ts`):** same index, same exposure, missed by the original pass. `resolvedClimbs` was `circuit.climbs.map((name) => nameToUuid.get(name))` — resolved by climb *name*, so a circuit listing the same climb twice, or two names resolving to one uuid, produced duplicate rows. The per-circuit transaction rolled back and the climber saw only a bumped `failed` count and a missing circuit. Extracted `resolveCircuitClimbUuids()` (dedupes, drops unresolved names, preserves order) and added `onConflictDoNothing` to the batch insert.

Did not touch `packages/kilter-sync` or `packages/backend` playlist mutations (fixed in #4021). Note for later: kilter-sync's `applyCircuits` has no JS dedupe at all and leans entirely on a bare `onConflictDoNothing`, while its `resolveCanonicalClimbUuid` alias mapping can collapse two raw uuids onto one — it should be brought up to this standard, not the reverse.

## Tests

Shim-based, no real Postgres:

- `packages/aurora-sync/src/sync/user-sync.test.ts` — duplicate `climb_uuid` dedupes to one row (first occurrence's angle/position wins), the insert carries `onConflictDoNothing` on `(playlistId, climbUuid)`, delete precedes insert, non-string entries skipped, the drop is logged, and nothing is logged when nothing was dropped.
- `packages/web/app/lib/data-sync/aurora/__tests__/user-sync-circuits.test.ts` — the same four assertions for the web copy.
- `packages/aurora-sync/src/sync/json-import.test.ts` — `resolveCircuitClimbUuids` collapses a repeated name, collapses two names resolving to one uuid, keeps first-occurrence order, drops unresolved names, and returns empty so the caller skips the delete.

Reverting the source hunk in `aurora-sync/src/sync/user-sync.ts` and re-running turns 2 of the original 4 tests red, so they are not tautological. (The "delete precedes insert" test asserts pre-existing ordering and passes either way.)

## Validation

- `vp test run --project aurora-sync --reporter=agent` — 9 files, 228 tests passed.
- `vp test run --project web --reporter=agent user-sync-circuits` — 9/9 passed.
- `vp run typecheck` — clean (exit 0, all 46 tasks).
- `vp check` — 2 errors and 324 warnings repo-wide, none in any file this PR touches. Both errors are `TS2307: Cannot find module 'yaml'` in `scripts/`, unrelated and pre-existing (#4488).
- Pre-commit hook (`vp check --fix` + commit-lint) passed.
- Rebased onto `origin/main` (`a71db63b9`); the rebase was mechanical, no conflicts.

## Assumptions / open questions

- **Does Aurora actually send the same `climb_uuid` twice in one circuit?** Nobody has verified this against a live payload, here or in #4023. If it does, the dedupe is silently truncating real circuits and the index should grow an `angle` column — a migration plus a change to the #4021 `addClimbToPlaylist` resolver. The new log line exists to answer that question from prod.
- With `onConflictDoNothing`, a concurrent `addClimbToPlaylist` that wins the race keeps its own angle/position rather than the Aurora snapshot's for that one climb. Acceptable; the next cycle's delete reconciles it.
- These tests do not exercise a real 23505 against a live index. The real-DB proof of the pattern already exists in `packages/backend/src/__tests__/add-climb-to-playlist-race.test.ts` from #4021.

## Release Notes

- Your circuits keep syncing when a circuit lists the same climb twice — one repeat used to quietly stall every sync for that board account, ticks included.
- Importing an Aurora export no longer drops a circuit that names the same climb more than once.

### DIFF
--- a/packages/aurora-sync/src/sync/json-import.test.ts
+++ b/packages/aurora-sync/src/sync/json-import.test.ts
@@ -11,6 +11,7 @@ import {
   isExportAscentActuallyAttempt,
   publishedClimbKey,
   readExportBool,
+  resolveCircuitClimbUuids,
   type AuroraExportAscent,
 } from './json-import';
 
@@ -360,5 +361,45 @@ describe('generateJsonImportCircuitAuroraId (#3526 / #3541)', () => {
     expect(generateJsonImportCircuitAuroraId('user-a', 'kilter', CIRCUIT, CREATED_AT)).toMatch(
       /^json-import-circuit-[0-9a-f]{32}$/,
     );
+  });
+});
+
+/**
+ * #4023: `unique_playlist_climb` is (playlist_id, climb_uuid). The JSON import
+ * resolves a circuit's climbs by NAME, so two entries collapsing onto one uuid
+ * used to produce duplicate insert rows, a raw 23505, and a rolled-back circuit
+ * that the climber only ever saw as a bumped `failed` count.
+ */
+describe('resolveCircuitClimbUuids (#4023)', () => {
+  const nameToUuid = new Map([
+    ['Crimp Ladder', 'uuid-a'],
+    ['Crimp Ladder ', 'uuid-a'],
+    ['Sloper Hell', 'uuid-b'],
+  ]);
+
+  it('collapses a climb listed twice under the same name to one row', () => {
+    expect(resolveCircuitClimbUuids(['Crimp Ladder', 'Sloper Hell', 'Crimp Ladder'], nameToUuid)).toEqual([
+      'uuid-a',
+      'uuid-b',
+    ]);
+  });
+
+  it('collapses two different names that resolve to the same uuid', () => {
+    expect(resolveCircuitClimbUuids(['Crimp Ladder', 'Crimp Ladder '], nameToUuid)).toEqual(['uuid-a']);
+  });
+
+  it('keeps the first occurrence, so circuit order survives the dedupe', () => {
+    expect(resolveCircuitClimbUuids(['Sloper Hell', 'Crimp Ladder', 'Sloper Hell'], nameToUuid)).toEqual([
+      'uuid-b',
+      'uuid-a',
+    ]);
+  });
+
+  it('drops names that resolve to nothing without disturbing the rest', () => {
+    expect(resolveCircuitClimbUuids(['Unknown Climb', 'Sloper Hell'], nameToUuid)).toEqual(['uuid-b']);
+  });
+
+  it('returns an empty list when nothing resolves, so the caller skips the delete', () => {
+    expect(resolveCircuitClimbUuids(['Unknown Climb'], nameToUuid)).toEqual([]);
   });
 });
--- a/packages/aurora-sync/src/sync/json-import.ts
+++ b/packages/aurora-sync/src/sync/json-import.ts
@@ -262,6 +262,30 @@ export function generateJsonImportCircuitAuroraId(
   return `json-import-circuit-${hash}`;
 }
 
+/**
+ * Resolve an export circuit's climb NAMES to climb uuids, dropping unresolved
+ * names and collapsing repeats.
+ *
+ * The dedupe is load-bearing, not tidiness: `unique_playlist_climb` is
+ * (playlist_id, climb_uuid), so a circuit that lists the same climb twice — or
+ * two distinct names that resolve to one uuid — produced duplicate rows and a
+ * raw 23505 that rolled the whole circuit's transaction back. The climber saw a
+ * bumped `failed` count and a silently missing circuit. Same index exposure the
+ * user-sync circuits branch fixes in #4023; extracted so it is testable without
+ * a database.
+ */
+export function resolveCircuitClimbUuids(climbNames: string[], nameToUuid: Map<string, string>): string[] {
+  const resolved: string[] = [];
+  const seen = new Set<string>();
+  for (const name of climbNames) {
+    const uuid = nameToUuid.get(name);
+    if (uuid == null || seen.has(uuid)) continue;
+    seen.add(uuid);
+    resolved.push(uuid);
+  }
+  return resolved;
+}
+
 export type JsonImportTickRow = typeof boardseshTicks.$inferInsert;
 
 /**
@@ -1622,9 +1646,7 @@ export async function importJsonExportData(
   // so one failure doesn't roll back others or abort the tick transaction)
   for (let ci = 0; ci < data.circuits.length; ci++) {
     const circuit = data.circuits[ci];
-    const resolvedClimbs = circuit.climbs
-      .map((name) => nameToUuid.get(name))
-      .filter((uuid): uuid is string => uuid != null);
+    const resolvedClimbs = resolveCircuitClimbUuids(circuit.climbs, nameToUuid);
 
     // The key is user-scoped (see generateJsonImportCircuitAuroraId), so two
     // importers can no longer collide on the global `playlists_aurora_id_idx`.
@@ -1727,7 +1749,14 @@ export async function importJsonExportData(
           }));
 
           for (let i = 0; i < climbValues.length; i += BATCH_SIZE) {
-            await tx.insert(playlistClimbs).values(climbValues.slice(i, i + BATCH_SIZE));
+            // Belt to the dedupe's braces: a re-import racing a concurrent
+            // addClimbToPlaylist can still land a row between the delete and
+            // this insert, and DO NOTHING keeps that one instead of aborting
+            // the circuit. #4023.
+            await tx
+              .insert(playlistClimbs)
+              .values(climbValues.slice(i, i + BATCH_SIZE))
+              .onConflictDoNothing({ target: [playlistClimbs.playlistId, playlistClimbs.climbUuid] });
           }
         }
 
--- a/packages/aurora-sync/src/sync/user-sync.test.ts
+++ b/packages/aurora-sync/src/sync/user-sync.test.ts
@@ -56,10 +56,15 @@ function createDbShim(
                 Promise.resolve(returningQueue[returningIdx++] ?? [{ id: BigInt(1) }]);
               return Object.assign(Promise.resolve(), { returning });
             },
-            onConflictDoNothing: () => Promise.resolve(),
+            onConflictDoNothing: (conflictArgs: unknown) => {
+              calls.push({ kind: 'conflict', table, args: [conflictArgs] });
+              return Promise.resolve();
+            },
           };
-          // playlist_climbs inserts are awaited straight off .values(), with no
-          // conflict clause — so the chain object has to be thenable too.
+          // Both the playlists upsert (.onConflictDoUpdate) and the
+          // playlist_climbs batch insert (.onConflictDoNothing) chain off
+          // .values() directly, so the chain object has to be thenable too —
+          // some callers await .values(...) with no conflict clause at all.
           return Object.assign(Promise.resolve(), chain);
         },
       };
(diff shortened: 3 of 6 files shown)


marcodejongh:

marcodejongh:
## Adversarial re-review (bug-batch loop `moarbugs100500`)

**Verdict: the fix is correct and the tests are real.** I reverted the `aurora-sync` source hunk locally and re-ran the suite — 2 of the 4 new tests go red (`dedupes a repeated climb_uuid`, `carries onConflictDoNothing targeting`), so they are not tautological. `vp test run --project aurora-sync` 221/221 and the web circuits file 9/9 pass with the fix in. The branch is 376 commits behind `origin/main` but `git merge-tree` reports zero conflicts, and none of the five touched files moved on `main` — rebase is mechanical.

Two things Marco should decide on, and one same-bug-different-path gap. Details below.

---

### Blast radius — the question I was asked to answer

The #3329 comparison in the PR body ("same poison-pill class as the kilter dup `log_uuid` incident") is **half right**, and the half that's wrong is the reassuring half.

**Fleet-wide: not a poison pill.** `syncNextUser` catches per credential (`packages/aurora-sync/src/runner/sync-runner.ts:224`) and `recordSyncFailure` (`:279`) unconditionally advances `last_sync_attempt_at`, bumps `consecutive_failures` and writes `last_sync_error` on *every* failure path. So a wedged credential rotates to the back of the queue and everyone else keeps syncing. That is the starvation fix the comment at `:234` describes, and it holds here.

**Per user: a permanent, total wedge — worse than #4023's "not urgent" framing.** `syncUserData` wraps *all* tables for one user in a single `db.transaction` (`packages/aurora-sync/src/sync/user-sync.ts:596`), and `updateUserSyncs` is inside it (`:625`). A 23505 out of the circuits branch therefore rolls back that user's **logbook, ascents and bids for the same batch**, and `last_synchronized_at` never advances. Next cycle re-fetches the byte-identical payload and throws again. One duplicated `climb_uuid` in one circuit stops that user's entire Aurora sync — ticks included — forever, with nothing visible but a growing `consecutive_failures`.

That is worth putting in the issue and, I'd argue, in the release notes rather than "no release note needed" (see hygiene note at the bottom).

---

### Findings

**1. (major / design decision) The dedupe silently deletes a climb the user actually has, with no log line.**
`packages/aurora-sync/src/sync/user-sync.ts:429-451`, `packages/web/app/lib/data-sync/aurora/user-sync.ts:340-372`.

`playlist_climbs.angle` is a real, populated column (`packages/db/src/schema/app/playlists.ts:82`) but `unique_playlist_climb` is `(playlist_id, climb_uuid)` only (`:92`). If an Aurora circuit legitimately holds climb X at 40° *and* 55°, first-occurrence-wins drops the 55° entry and the Boardsesh copy of that circuit is permanently one climb short. Before this PR it threw (loudly, wedging the user); now it's silent.

Trading a wedge for silent loss is the right call, but two things follow:
- **Minimum ask: log the drop.** Right now nothing distinguishes "Aurora never sends dupes" from "we've been quietly truncating circuits for months". One `log()` line when `climbRows.length !== item.climbs.length` costs nothing and tells us whether finding 1 is theoretical.
- **The real fix, if Aurora does carry per-angle circuit entries,** is widening the index to `(playlist_id, climb_uuid, angle)` — migration plus a matching change to the `addClimbToPlaylist` resolver from #4021. Out of scope for this PR; your call whether it's worth a follow-up issue.

I could not establish from the repo whether Aurora circuits actually repeat a `climb_uuid` at two angles. Nobody has verified this against a live payload, here or in #4023.

**2. (major) Identical exposure left unfixed one file over: `packages/aurora-sync/src/sync/json-import.ts:1730`.**

```
for (let i = 0; i < climbValues.length; i += BATCH_SIZE) {
  await tx.insert(playlistClimbs).values(climbValues.slice(i, i + BATCH_SIZE));
}
```

No dedupe, no `onConflictDoNothing`. `resolvedClimbs` at `:1624` is `circuit.climbs.map((name) => nameToUuid.get(name))` — a *name*-keyed lookup, so a circuit listing the same climb twice, or two names resolving to one uuid, produces duplicate rows and a 23505. Blast radius is contained (per-circuit transaction, caught at `:1745`, counted `failed`) so it doesn't wedge the import — but that circuit silently doesn't import and the user only ever sees a `failed` count. Same three-line fix. I'm folding it in here since it's the same defect against the same index in the same package; say the word if you'd rather it were its own PR.

**3. (minor) The claimed parity with kilter-sync isn't quite real.**
The new comments say "same pattern as kilter-sync's `applyCircuits`". `packages/kilter-sync/src/sync/user-sync.ts:1463-1468` actually uses a **bare** `.onConflictDoNothing()` with no `target` and **no JS dedupe at all** — it relies entirely on Postgres skipping intra-statement duplicates under `DO NOTHING`. Notably kilter's path collapses aliases via `resolveCanonicalClimbUuid` (`:1432`), which can map two distinct raw uuids onto one canonical, so it is *more* dupe-prone than aurora's and has less protection. Not a defect in this PR — but the comment overstates the match, and kilter's version is the one that should be brought up to this PR's standard, not the reverse.

**4. (minor, pre-existing review comment, still open) Chunk-size inconsistency.**
`aurora-sync` reuses the file-level `BATCH_SIZE = 100`; web declares a local `CLIMB_INSERT_CHUNK_SIZE = 500` inside the `if` block (`packages/web/app/lib/data-sync/aurora/user-sync.ts:377`). Both are far under the 65535-parameter ceiling (500 × 4 = 2000), so this is cosmetic — but for the record kilter-sync also uses 500 (`:1462`), so **aurora-sync at 100 is the outlier**, not web. Leaving both as-is; flagging that the earlier review had it backwards.

**5. (nit) One of the four new tests per suite carries no weight.**
"deletes playlist_climbs before inserting the new rows" passes unchanged with the fix reverted — it asserts pre-existing ordering. Harmless, just not evidence of anything.

---

### What I did not check

- No live-Postgres exercise of the 23505 path; both suites are db-shim based, as the PR body already says.
- No evidence gathered from prod that any user is currently wedged by this. If you want that, the query is `aurora_credentials` where `consecutive_failures > 5` and `last_sync_error` matches `23505`/`unique_playlist_climb` — I have read-only prod access and can run it if it'd change the priority.

### PR hygiene

- Title is a valid conventional commit; commit-lint is green.
- `## Release Notes` is present as the "no release note needed" checkbox. Given the per-user wedge above this *is* user-facing ("your circuits and your ticks stop syncing"), so I'd swap it for a real line unless you disagree. Doing that, revert it if you'd rather.
- The PR is already out of draft. Per this loop's rules I'm not merging and not touching ready/draft state — that's yours.

marcodejongh:
## Adoption complete — ready for your review

Rebased onto `origin/main` (`a71db63b9`, 376 commits, no conflicts, none of the touched files had moved) and pushed. **Every check is green**, including both `test-default` shards, `test-backend`, `mobile-bundle` and `commit-lint`. `mergeable=MERGEABLE`, `reviewDecision=REVIEW_REQUIRED` — the only thing outstanding is your approval. Not merging and not touching draft state, per this loop's rules.

What changed since my [findings comment](#issuecomment-5348799391):

**Fixed (finding 2, major).** `packages/aurora-sync/src/sync/json-import.ts` had the same exposure against the same index and the original pass missed it. Resolved by climb *name*, so a circuit listing one climb twice — or two names mapping to one uuid — threw 23505 and rolled the circuit's transaction back, leaving the climber a bumped `failed` count and a silently missing circuit. Extracted `resolveCircuitClimbUuids()` (dedupes, drops unresolved names, keeps order) with five tests, and added `onConflictDoNothing` to the batch insert. Say the word if you'd rather this were its own PR and I'll split it.

**Fixed (finding 1, the "minimum ask").** Both user-sync copies now log every dropped duplicate. Two new tests: the log fires with the right count and circuit uuid, and stays quiet when nothing was dropped. This is the instrument that answers the open question below.

**Not fixed, deliberately.** Findings 3 (kilter-sync's `applyCircuits` has no JS dedupe and leans entirely on a bare `onConflictDoNothing`, while its alias resolution can collapse two raw uuids onto one), 4 (chunk-size inconsistency — cosmetic, and aurora-sync's 100 is the outlier, not web's 500), and 5 (the delete-before-insert test carries no weight). All recorded in the PR body, none of them worth widening this diff.

**The one thing still open for you.** Nobody has confirmed whether Aurora actually repeats a `climb_uuid` in a circuit. If it does, the dedupe is quietly truncating real circuits, and the fix is widening the index to `(playlist_id, climb_uuid, angle)` — a migration plus a change to the #4021 `addClimbToPlaylist` resolver, which is its own PR. The new log line is there to answer that from prod rather than by guessing. I can also run the read-only prod query (`aurora_credentials` where `consecutive_failures > 5` and `last_sync_error` mentions `23505`/`unique_playlist_climb`) if you want to know whether anyone is currently wedged — say so and I'll post the numbers.

Local validation: `vp test run --project aurora-sync` 228/228, web circuits 9/9, `vp run typecheck` exit 0. `vp check` reports 2 errors and 324 warnings repo-wide, none in any file this PR touches — both errors are `TS2307: Cannot find module 'yaml'` under `scripts/`, unrelated and pre-existing (#4488).

Issue #3933, opened 2026-07-25
setQueue still overwrites a concurrent addQueueItem (residual of #3906)


marcodejongh:
Follow-up carved out of #3906 / PR #3932, which deliberately did **not** fix this.

## What's left

#3906 made queue-state writes an atomic compare-and-swap, and every mutation that derives its new queue from the current one now read-compute-CAS-retries. `setQueue` is exempt on purpose: its payload is entirely client-supplied, so there is nothing to recompute against a concurrent write, and replacing server state wholesale is the mutation's contract. It passes `CAS_ANY_VERSION` — it needs a unique sequence, not a conflict.

The residual: **a peer's `addQueueItem` landing inside a `setQueue` window is still silently overwritten.**

## Why it's worth tracking rather than shrugging at

Web's drag-to-reorder does not send the `reorderQueueItem` delta — it pushes a full `setQueue` (see the comment at `packages/backend/src/graphql/resolvers/queue/mutations.ts`, the #2387 no-op-diagnostic block). So the exposed window isn't an exotic sync path; it's someone reordering their queue while a party member adds a climb. The added climb disappears with no drift the hash watchdog can detect, for the same reason as the original bug: the server's post-write state is internally consistent.

Offline→online reconciliation and playlist activation also take this path.

## Why it wasn't fixed in #3906

Fixing it properly means the queue stops being last-writer-wins and starts being a mergeable structure — per-item LWW, or client-assigned positions with deterministic conflict resolution. That is a rewrite across the backend, `@boardsesh/queue`, web and mobile, and it has an OTA compatibility problem: old mobile clients linger, and the current wire protocol's gap detector requires a gapless server-assigned total order.

## Possible directions

- Have web's drag-to-reorder emit `reorderQueueItem` (already CAS-guarded) instead of a full `setQueue`. Narrowest fix, kills the most likely real-world instance, doesn't touch the protocol.
- Give `setQueue` an optional `expectedVersion` for callers that *can* meaningfully retry (the reorder path can; offline reconciliation genuinely cannot).
- Merge-on-write: union incoming `setQueue` membership with items added since the client's baseline sequence.

The first is probably most of the value for a fraction of the cost.

Priority is not urgent — no measured instances, unlike #3906 which had 23 affected production sessions.

marcodejongh:
[...]

PR #4289, opened 2026-08-11
fix(mobile): keep the throttled queue-add recovery alive after a full sync

### LINKED ISSUE #4009: Mobile: the burst-head throttled queue-add recovery goes inert once a FullSync wipes the local slot
Follow-up from #4005 (the fix for #3936). Mobile counterpart of #4006.

`recoverThrottledQueueAdd` in `packages/mobile/src/providers/queue-provider.tsx` is the caller-side recovery
#3934 added for the climb that *heads* a throttled burst — the one whose `enqueue` promise actually rejects.
It bails on a local-absence check:

```ts
const position = stateRef.current.queue.findIndex((queueItem) => queueItem.uuid === item.uuid);
// Already gone locally — removed, cleared, or replaced by a server sync
// while the throttled mutation was in flight. Don't resurrect it.
if (position === -1) return;
```

#4005 proved that "replaced by a server sync" is the *dominant* reason the slot is gone by the time a
rate-limited recovery runs, not a rare one:

- `queue-navigation.ts` publishes a `FullSync` whenever a `setCurrentClimb` added to the queue, with no
  `clientId`, so the originating client cannot suppress its own echo.
- `sync-coordinator.ts` maps `FullSync` to `INITIAL_QUEUE_DATA`.
- `reducer.ts` applies `INITIAL_QUEUE_DATA` by **replacing** `queue` wholesale.

So a peer's (or a later self-) activation lands a FullSync that wipes the not-yet-synced optimistic slot, and
the recovery then reads `-1` and drops the climb — the same inert-guard problem #4005 had to design around in
the shared coalescer. The climb reaches nobody.

Worse second-order effect after #4005: line ~931 then treats the absence as "the climber removed it" and fires
`mutations.removeQueueItem(item.uuid)` to undo an add that *did* land, which now also writes that uuid into the
shared removal ledger and suppresses any later legitimate deferred add for it.

### Suggested fix

Reuse the split #4005 landed in `create-queue-mutations.ts`: absence caused by an explicit removal → skip;
absence caused by a wholesale sync → send unpositioned. The ledger already lives in the shared factory, so the
cheapest shape is probably to expose the same decision from there rather than reimplement the heuristic in the
provider.

### Priority

Low-ish. It only bites the head of a genuinely rate-limited burst in a party session, and the drift watchdog
eventually reconciles the picker's own queue. But it is the last hole of the #3936 family.

marcodejongh:
Closes #4009

## The bug

`recoverThrottledQueueAdd` (`packages/mobile/src/providers/queue-provider.tsx`) is the caller-side recovery #3934 added for the climb that *heads* a throttled burst — the one whose activation promise actually rejects. It bailed on a bare local-absence check:

```ts
const position = stateRef.current.queue.findIndex((queueItem) => queueItem.uuid === item.uuid);
if (position === -1) return;
```

#4005 already proved that "replaced by a server sync" is the dominant reason the slot is gone by then, not a rare one. A queue-adding activation makes `queue-navigation.ts` publish a `FullSync` with no `clientId`, `sync-coordinator.ts` maps it to `INITIAL_QUEUE_DATA`, and `reducer.ts` applies that by **replacing** `queue` wholesale — wiping the not-yet-synced optimistic slot seconds before the rate-limited mutation rejects. The recovery then dropped the add and the climb reached nobody.

Second-order, since #4005: the same absence drove the undo leg, so an add that *did* land got a compensating `removeQueueItem`. That deletes the picker's own pick off the crew's queue **and** writes the uuid into the shared removal ledger, which then suppresses every later legitimate deferred add for it.

## The fix

Split the two causes the way the shared coalescer's `sendDeferredQueueAdd` already does, by reading the ledger the factory keeps instead of re-deriving the heuristic in the provider (the shape the issue suggested):

| local state | verdict | action |
| --- | --- | --- |
| still queued | — | positioned `ADD_QUEUE_ITEM` (peers see our order) |
| gone, ledger hit | the climber dropped it | nothing — don't resurrect their discard |
| gone, no ledger hit | a wholesale sync wiped it | unpositioned `ADD_QUEUE_ITEM` (server appends) |

and gate the undo leg on a ledger hit rather than on bare absence.

- `packages/shared/queue-react/src/create-queue-mutations.ts` — new read-only `wasUuidExplicitlyRemoved(uuid)` action on `QueueMutationsActions`. Synchronous, side-effect free. The hook returns the factory object as-is, so it flows through to web and the backend headless test client with no wiring.
- `packages/mobile/src/providers/queue-provider.tsx` — the three-way recovery above, plus a session-flip bail (see below).

### Session-flip bail

The throttled `setCurrentClimb` can sit in back-off while the climber leaves and joins another room, and `addQueueItem` resolves the **live** session — so an unpositioned send would append the old room's climb to the new crew's queue. `dispatchSetCurrent` snapshots `sessionIdRef.current` at activation time and passes it to the recovery, which bails when the live session no longer matches. This also covers the `position >= 0` branch, which sent into the wrong room before this PR too, and the solo-to-party case (`null !== 'session-2'`).

### The ledger records latest intent, not "ever dropped" (second commit)

Adversarial re-review found the guard above was leaning on a ledger that only ever grew. Nothing deleted from `removedUuids` except the 2000-entry cap, and `useQueueMutations` memoises the factory on `[hasEnsureReady]`, which never toggles — so on mobile the Set lives for the whole app process.

That made the new guard read a verdict that could be hours old. Mobile's clear-queue fires one `removeQueueItem` per queued item, so a single **Clear queue** puts every queued uuid in the ledger permanently. Re-pick one of those climbs — ordinary, with a project — during a throttled burst, let the burst-head `FullSync` wipe the slot, and the recovery reads `true` and bails. The climb reaches nobody: #4009 again, silently conditioned on "was this uuid ever dropped in this process". The same staleness already reached `sendDeferredQueueAdd`, which has consulted the ledger since #4005.

So every path that puts the climb back now retracts the entry, each before its own awaits:

- `addQueueItem`
- `setCurrentClimb` carrying a queue-add (a pointer-only move does not — navigation is not "put this back")
- `setQueue`, for every uuid the payload contains

Ordering keeps the #3934 undo leg alive: the recovery's add forgets, the climber's swipe during its back-off re-records, and the undo check reads the swipe. It also closes the cross-user leak where user A's removal would suppress user B's pick after a sign-out on the same device.

The same commit restores the `false` default for `wasUuidExplicitlyRemoved` in the third test harness — `describe('QueueProvider always-live wall control')` ran the same blanket `mockReset(); mockResolvedValue(undefined)` loop as the other two, handing a synchronous boolean action a truthy Promise.

## Validation

- `vp test run --project queue-react --reporter=agent` — 3 files, 63 tests passed
- `vp run test:mobile` — 684 files, 6551 tests passed
- `vp run typecheck:mobile` / `typecheck:shared` / `typecheck:web` — clean
- `vp lint packages/shared/queue-react/src packages/mobile/src/providers` — 0 errors (warnings are all pre-existing `no-floating-promises` in `act()` callbacks)
- `vp fmt --check` on both trees — clean

The four load-bearing retraction tests were checked against a no-op `forgetRemovedUuid`: all four fail, the two control cases stay green.

### Tests

Factory (`packages/shared/queue-react/src/__tests__/create-queue-mutations.test.ts`), against the real `createQueueMutations`:

- `wasUuidExplicitlyRemoved` false for an untouched uuid; true after `removeQueueItem`, including the no-session no-op path; true after a wholesale `setQueue` replace discards a remembered add-candidate; false for a pointer-only activation; reading does not mutate
- retraction: forgotten after `addQueueItem`, after a queue-adding activation, and for every uuid a `setQueue` payload contains — but **not** after a pointer-only re-activation, and a uuid the same payload discarded is still recorded
- end to end through the coalescer: a climb dropped earlier and since re-picked still gets its deferred add when a wholesale sync wipes the local slot

Provider (`packages/mobile/src/providers/__tests__/queue-provider-session-updates.test.tsx`) — these pin how the provider **branches** on the verdict. Recovery is driven by rejecting the mocked `setCurrentClimb` with a real `GraphQLOperationError` carrying `code: 'RATE_LIMITED'`, and the wipe by a real `FullSync` pushed over the `queueUpdates` sink.

- the #4009 regression: throttled activation, `FullSync` wipes the slot → an unpositioned `ADD_QUEUE_ITEM` is still sent (`position` asserted `undefined`), no queue refetch, pacing toast intact
- add lands, then a `FullSync` wipes the slot with no removal → **no** `removeQueueItem` undo
- session joined while the activation was backing off → nothing sent into the new room
- the existing "climber removed it mid-flight → skip" and "climber removed it while the add was in flight → undo still fires" cases keep passing on the ledger rather than on bare absence (#3934 semantics preserved)

- **A peer removing the climb during the recovery's back-off is no longer undone.** The undo leg now needs a ledger hit, and a peer's removal arrives as a server delta with no entry. Sequence: our slot is still local, the recovery sends a positioned add, a peer deletes the climb mid-back-off, our add lands after it and puts the climb back, and nothing compensates. Pre-PR the bare-absence undo did fire here. The PR documents the resurrection tradeoff on the *send* side (matching #4005); this is the same tradeoff on the *undo* side. I think it is the right trade — undoing on bare absence is what deletes the picker's own pick and poisons the ledger, and that is both worse and far more common — but it is a real behaviour change. Closing it properly needs the reducer to distinguish a `QUEUE_ITEM_REMOVED` delta from an `INITIAL_QUEUE_DATA` replace and surface that to the recovery; happy to file it.
- **Deliberate change to the #3934 undo leg**: an add that landed server-side but is locally absent with no ledger hit is no longer undone — the climb survives on the crew's queue instead of being removed. Intended (it is the picker's pick), but it changes observable party behaviour.
- **API shape**: `wasUuidExplicitlyRemoved` widens `QueueMutationsActions`, which flows to web and the backend headless test client. Every implementer gets the object from the factory and nothing constructs the type by hand, so nothing breaks — but it is your interface.
- The unpositioned send appends rather than preserving insert-after-current, so peers may see the recovered climb at the queue tail. Same degradation #4005 chose, and only on the sync-wiped branch.

## Release Notes

A climb picked during a busy party session no longer goes missing for the rest of the crew when the app is pacing itself — including a climb you'd cleared off your queue earlier in the session and queued back up.

### DIFF
--- a/packages/mobile/src/providers/__tests__/queue-provider-local-queue.test.tsx
+++ b/packages/mobile/src/providers/__tests__/queue-provider-local-queue.test.tsx
@@ -61,6 +61,11 @@ const queueMutations = vi.hoisted(() => ({
   confirmClimbOnWall: vi.fn(async () => {}),
   setSessionBoardSerial: vi.fn(async () => {}),
   setSessionBoardPath: vi.fn(async () => {}),
+  // Synchronous read of the shared factory's removal ledger (#4009). Nothing
+  // here removes a climb mid-recovery, so the honest default is "not dropped by
+  // the climber" — but the member has to exist or the provider's recovery path
+  // would call undefined.
+  wasUuidExplicitlyRemoved: vi.fn((_uuid: string) => false),
 }));
 
 // Capture the deps mobile passes into the shared useQueueMutations so the
@@ -232,6 +237,11 @@ describe('QueueProvider local solo queue', () => {
       mutation.mockReset();
       mutation.mockResolvedValue(undefined);
     }
+    // The blanket reset above hands every action a resolved promise, which for a
+    // SYNCHRONOUS boolean action is truthy — i.e. "the climber removed it"
+    // everywhere. Restore the real default.
+    queueMutations.wasUuidExplicitlyRemoved.mockReset();
+    queueMutations.wasUuidExplicitlyRemoved.mockReturnValue(false);
     sessionStore.getStoredSessionId.mockReset();
     sessionStore.getStoredSessionId.mockResolvedValue(null);
     sessionStore.setStoredSessionId.mockClear();
--- a/packages/mobile/src/providers/__tests__/queue-provider-session-updates.test.tsx
+++ b/packages/mobile/src/providers/__tests__/queue-provider-session-updates.test.tsx
@@ -122,7 +122,9 @@ const partyProfile = vi.hoisted(() => ({
 
 const queueMutations = vi.hoisted(() => ({
   addQueueItem: vi.fn(async () => {}),
-  removeQueueItem: vi.fn(async () => {}),
+  // Typed arg so a test can model the shared factory's removal ledger, which
+  // records here before any session guard (#4009).
+  removeQueueItem: vi.fn(async (_uuid: string) => {}),
   reorderQueueItem: vi.fn(async () => {}),
   setCurrentClimb: vi.fn(async () => {}),
   mirrorCurrentClimb: vi.fn(async () => {}),
@@ -133,6 +135,13 @@ const queueMutations = vi.hoisted(() => ({
   reportWallDisconnect: vi.fn(async () => {}),
   setSessionBoardSerial: vi.fn(async () => {}),
   setSessionBoardPath: vi.fn(async () => {}),
+  // Read-only classification of a LOCAL absence, backed in production by the
+  // shared factory's removal ledger (#4009). These tests verify how the
+  // provider BRANCHES on the verdict; the ledger that produces it is covered in
+  // packages/shared/queue-react/src/__tests__/create-queue-mutations.test.ts.
+  // Default false = "the climber didn't drop it", so an absence reads as a
+  // wholesale server sync; tests that model a removal override it.
+  wasUuidExplicitlyRemoved: vi.fn((_uuid: string) => false),
 }));
 
 const wallConfirm = vi.hoisted(() => ({
@@ -473,6 +482,11 @@ describe('QueueProvider session update subscription', () => {
       mutation.mockReset();
       mutation.mockResolvedValue(undefined);
     }
+    // The blanket reset above hands every action a resolved promise. For a
+    // SYNCHRONOUS boolean action that is a truthy Promise, which would read as
+    // "the climber removed it" everywhere. Restore the real default.
+    queueMutations.wasUuidExplicitlyRemoved.mockReset();
+    queueMutations.wasUuidExplicitlyRemoved.mockReturnValue(false);
     wallConfirm.emitWallConfirm.mockClear();
     sessionStore.getStoredSessionId.mockReset();
     sessionStore.getStoredSessionId.mockResolvedValue('session-1');
@@ -1652,7 +1666,46 @@ describe('QueueProvider session update subscription', () => {
 describe('QueueProvider mutation-failure resync', () => {
   beforeEach(() => {
     toast.showToast.mockClear();
+    installRemovalLedger();
   });
+
+  // This harness swaps the shared mutations factory for hand-rolled mocks, so
+  // the removal ledger those actions maintain has to be modelled here: the real
+  // `removeQueueItem` records the uuid before any session guard, and
+  // `wasUuidExplicitlyRemoved` reads that ledger back (#4009). The ledger itself
+  // is covered against the real factory in
+  // packages/shared/queue-react/src/__tests__/create-queue-mutations.test.ts;
+  // what these tests pin is how the provider BRANCHES on its verdict. Absence
+  // with no ledger hit means a wholesale server sync, not climber intent.
+  function installRemovalLedger() {
+    const explicitlyRemoved = new Set<string>();
+    queueMutations.removeQueueItem.mockImplementation(async (uuid: string) => {
+      explicitlyRemoved.add(uuid);
+    });
+    queueMutations.wasUuidExplicitlyRemoved.mockImplementation((uuid: string) => explicitlyRemoved.has(uuid));
+  }
+
+  // A server FullSync — the event whose INITIAL_QUEUE_DATA mapping REPLACES the
+  // local queue wholesale, wiping any not-yet-synced optimistic slot (#4009).
+  function pushFullSync(items: ClimbQueueItem[], current: ClimbQueueItem | null, sequence: number) {
+    const sink = ws.getQueueUpdatesSink();
+    if (!sink) throw new Error('queueUpdates subscription was not opened');
+    const toWire = (item: ClimbQueueItem) => ({ uuid: item.uuid, climb: item.climb });
+    sink.next({
+      data: {
+        queueUpdates: {
+          __typename: 'FullSync',
+          sequence,
+          state: {
+            sequence,
+            stateHash: `hash-${sequence}`,
+            queue: items.map(toWire),
+            currentClimbQueueItem: current ? toWire(current) : null,
+          },
+        },
+      },
+    });
+  }
 
   // The harness routes both endSession and the queueState query through
   // http.request. Branch on the operation text so the resync query returns the
@@ -2072,6 +2125,9 @@ describe('QueueProvider mutation-failure resync', () => {
       expect(toast.showToast).toHaveBeenCalledWith('mobile.queue.rateLimited', 'error');
     });
     // Recovering the slot now would resurrect a climb the user just deleted.
+    // Absence alone no longer decides that (#4009) — the removal ledger does,
+    // and the swipe above put this uuid in it.
+    expect(queueMutations.wasUuidExplicitlyRemoved).toHaveBeenCalledWith('local-current');
     expect(queueMutations.addQueueItem).not.toHaveBeenCalled();
     expect(queueStateCalls).toBe(0);
   });
@@ -2126,7 +2182,9 @@ describe('QueueProvider mutation-failure resync', () => {
     });
 
     // The add landed after the remove, so it put the climb back on every peer's
-    // queue. Undo it instead of leaving a climb nobody asked for.
+    // queue. Undo it instead of leaving a climb nobody asked for. The removal
+    // ledger is what licenses that undo (#4009) — the swipe above wrote this
+    // uuid into it, so absence here reads as intent rather than as a sync.
     await waitFor(() => {
       expect(queueMutations.removeQueueItem).toHaveBeenCalledTimes(2);
     });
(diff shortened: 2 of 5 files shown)


marcodejongh:

marcodejongh:
## Adversarial re-review — verdict: the fix is right, but the ledger it now leans on is write-only

The three-way split is the correct shape and the tests are real: revert either guard and the new cases fail. They drive a genuine `FullSync` over the `queueUpdates` sink and a real `GraphQLOperationError{code:'RATE_LIMITED'}`, not a stubbed branch. The session-flip capture in `dispatchSetCurrent` is a genuine improvement on what the earlier review asked for.

One **major** finding, one minor, and two things that need your call. I am fixing the major and the minor in this PR, and rebasing onto `main` (376 commits behind; `602f46de8` touched `create-queue-mutations.ts` since the branch point).

---

### MAJOR — `removedUuids` is append-only, so a stale entry re-opens #4009 for any climb the climber ever dropped

`packages/shared/queue-react/src/create-queue-mutations.ts:228` — the ledger is a `Set` on the factory instance. `useQueueMutations` memoises the factory on `[hasEnsureReady]`, which never toggles, so on mobile it lives for the whole app process. Nothing ever deletes from it except the 2000-entry eviction at :231. Neither `addQueueItem` (:380) nor `setCurrentClimb(..., shouldAddToQueue=true)` (:405) clears an entry, so a re-add never retracts the earlier "dropped".

The new guard at `queue-provider.tsx:1115` therefore reads a verdict that can be minutes or hours old:

1. Climber taps **Clear queue**. Mobile fires one `removeQueueItem` per item, so every uuid that was in the queue lands in the ledger permanently (deliberate — see the comment at :199).
2. Ten minutes later, same app session, they re-pick one of those climbs during a busy party. Re-queueing a project you cleared is ordinary behaviour.
3. The activation is throttled; the burst-head `FullSync` wipes the optimistic slot.
4. Recovery: `position === -1`, `wasUuidExplicitlyRemoved` returns `true`, bail. **The climb reaches nobody.**

That is the #4009 symptom again, silently conditioned on "was this uuid ever dropped in this process". Not a regression against `main` (which bailed unconditionally), but it is a hole in the fix, and it is the "state the ledger never leaves" failure mode.

The same staleness already bites `sendDeferredQueueAdd` at :353 on `main` today. It is pre-existing, but this PR promotes the ledger from one call site to a general-purpose classifier, so it belongs here.

**Fix:** make the ledger record the climber's *latest* intent rather than "ever dropped" — forget the uuid on `addQueueItem`, on `setCurrentClimb` with `shouldAddToQueue`, and for every uuid present in a `setQueue` payload. Ordering still works out for the case #3934's undo leg exists for: the recovery's `addQueueItem` forgets at t0, the climber's swipe records at t1, the undo check reads `true` at t2. It also closes the sign-out / user-switch leak, where user B's pick would otherwise be suppressed by user A's removal.

### MINOR — a third harness leaves `wasUuidExplicitlyRemoved` returning a truthy Promise

`packages/mobile/src/providers/__tests__/queue-provider-session-updates.test.tsx:2950` — `describe('QueueProvider always-live wall control')` runs the same blanket `mockReset(); mockResolvedValue(undefined)` loop but does not get the corrective `mockReturnValue(false)` that :488 adds. The action is synchronous and boolean, so inside that block it returns a Promise, which is truthy, i.e. "the climber removed it" for every uuid. Nothing there exercises the recovery today, so it is latent, but it is a false-green trap for the next test written in that block. Two of three harnesses were fixed; this one was missed.

### For you to decide

**1. A peer removing the climb during the recovery's back-off is no longer undone.** `queue-provider.tsx:1142` gates the #3934 undo on a ledger hit, and a peer's removal arrives as a server delta with no ledger entry. Sequence: our slot is still local, the recovery sends a *positioned* add, a peer deletes the climb mid-back-off, our add lands after it and puts the climb back, and now nothing compensates. Pre-PR the bare-absence undo did fire here.

The PR body documents the resurrection tradeoff on the *send* side (matching #4005), but this is that tradeoff leaking into the *undo* side, which is a real behaviour change rather than a documentation note. I think it is the right trade — undoing on bare absence is what deletes the picker's own pick and poisons the ledger, and that is both worse and far more common — but you are trading "the picker's pick vanishes for the whole crew" against "a peer's delete occasionally bounces back". Closing it properly needs the reducer to distinguish a `QUEUE_ITEM_REMOVED` delta from an `INITIAL_QUEUE_DATA` replace and surface that to the recovery. Out of scope here; say the word and I will file it.

**2. Widening `QueueMutationsActions` with a read-only accessor.** It flows to web and the backend headless test client. I checked every implementer: all three get the object from the factory and nothing constructs the type by hand, so nothing breaks. Fine by me, but it is your interface.

### Things I checked that are clean

- **Staleness is bounded.** `RATE_LIMIT_MAX_RETRIES = 2`, `RATE_LIMIT_MAX_WAIT_MS = 30_000`, per-attempt timeout reset — the head rejects within roughly 60-90s, and the recovery's own `addQueueItem` gets one more such budget. This cannot revive a queue add from days ago. The coalescer only rejects the burst head; superseded activations resolve and become deferred adds.
- **App restart** wipes the in-memory ledger, so the verdict defaults to `false` = "send". That is the safe direction.
- **Session flip.** The `activationSessionId` snapshot at `queue-provider.tsx:1207` covers both the positioned and unpositioned branches, and also the solo-to-party case (`null !== 'session-2'` bails) that `main` got wrong. Strictly better than the literal line the earlier review asked for.
- **A second `FullSync` mid-recovery** that includes the item: `queue.some(...)` short-circuits the undo.
- **Going offline** mid-flight rejects non-`RATE_LIMITED`, so `throttled` is false and the recovery never runs; the resync path handles it. Going offline during the recovery's own add lands in the `catch` and calls `resyncQueueAfterMutationFailure`.
- **Solo climbers**: `activationSessionId` is `null` and `!recoverySessionId` bails. No change.
- **`create-queue-mutations.test.ts`** exercises the real factory, including the no-session `removeQueueItem` path and the `setQueue` add-candidate diff. Not mock-shaped.

### Nit, not fixing

`queue-provider.tsx:1115` only consults the ledger when `position === -1`. If the climber dropped the climb and a peer's `FullSync` then put it back into our local queue, we send a positioned add on top of a discard. Harmless — it is already on the crew's queue.

marcodejongh:
Takeover complete — **ready to merge**, pending your review. Not marking it ready myself.

**What changed since the review above**

Rebased onto `origin/main` (was 376 commits behind). One conflict, in `create-queue-mutations.test.ts`: `main`'s `setQueue baseline sequence (#3933)` describe and this branch's `wasUuidExplicitlyRemoved` describe were both appended at the end of the file. Kept both, no semantic overlap. `create-queue-mutations.ts` and `queue-provider.tsx` auto-merged; I read both through afterwards and `602f46de8`'s synchronous `baselineSequence` read sits above the ledger writes, which is where it belongs.

Then a second commit, `371c390a6`, for the major and the minor:

- **Major, fixed.** `removedUuids` now records the climber's *latest* intent instead of "ever dropped". `addQueueItem`, a queue-adding `setCurrentClimb`, and every uuid a `setQueue` payload contains retract the entry — each before that path's own awaits, so a remove racing in afterwards still records and still wins. A pointer-only activation deliberately does not retract: navigation is not "put this back". That keeps the #3934 undo leg working (recovery's add forgets, the swipe during its back-off re-records, the undo check reads the swipe) and closes the cross-user leak after a sign-out.
- **Minor, fixed.** `describe('QueueProvider always-live wall control')` gets the `mockReturnValue(false)` correction the other two harnesses had.
- The mobile harness's `installRemovalLedger` now models the retraction too, so it does not drift from the factory.

**New tests** — six at the factory level, against the real `createQueueMutations`, including one end to end through the coalescer (dropped earlier, re-picked, wholesale sync wipes the slot, the deferred add still goes). I checked they are load-bearing: with `forgetRemovedUuid` stubbed to a no-op, four fail and the two control cases stay green.

**Validation**

- `vp test run --project queue-react` — 63 passed
- `vp run test:mobile` — 684 files, 6551 passed
- `vp run typecheck:mobile` / `typecheck:shared` / `typecheck:web` — clean
- `vp lint` on both trees — 0 errors; `vp fmt --check` clean
- CI on `371c390a6`: **every check green**, including `test-default` (both shards), `test-backend`, `typecheck`, `lint`, `mobile-bundle`, `i18n`, `commit-lint`, `release-notes`, `ota-check`. `mergeable: MERGEABLE`, `mergeStateStatus: BLOCKED` is the review requirement, not a failing check.

**Still yours to decide** (both restated in the PR body, neither blocking):

1. The peer-remove-during-back-off case is no longer undone. Real behaviour change, and the one thing here I would not merge without your nod. Fixing it properly needs the reducer to tell a `QUEUE_ITEM_REMOVED` delta apart from an `INITIAL_QUEUE_DATA` replace — say the word and I will file it rather than grow this PR.
2. Widening `QueueMutationsActions` with the read-only accessor.

PR #4291, opened 2026-08-11
fix(mobile): render whole route in multi-frame climb thumbnails

### LINKED ISSUE #3988: Mobile in-app thumbnails render only frame 0 of a multi-frame climb
## What happens

For a multi-frame climb (a route or circuit — 709 of them in the catalog), Boardsesh renders a different static picture depending on where you look:

| surface | what it draws |
| --- | --- |
| web climb card / overlay (`toFlatFrames`) | the union of every frame — the whole route |
| backend OG share card (`toThumbnailFrames`) | the union |
| mobile share card (`use-share-climb.ts` → backend) | the union |
| **mobile in-app thumbnails and the play-drawer board** | **frame 0 only** |

The mobile in-app path goes `BoardImageNative` / `ClimbListThumbnail` → `useNativeClimbRender` (`packages/mobile/src/hooks/use-native-climb-render.ts:747`), which hands the raw `frames` string to the Rust renderer. `packages/board-renderer/core/src/frames_parser.rs:8` does:

```rust
let first_frame = frames.split(',').next().unwrap_or("");
```

so everything after the first comma is discarded.

Open `F142693E7BF54FBEA3DC936E263DEEF5` "Happy New Year! 🎉" (46 frames): the web card and the shared image show ~225 lit holds; the same climb in the mobile climb list shows a handful.

## Why it is worth fixing now

This is pre-existing — the Rust renderer has always been frame-0-only — but #3947 widened the gap. Before that PR the web flat render was a cumulative accumulation; after it, it is the union of every frame. Mobile did not move.

## Options

1. **Wrap at the call site.** Pass `toFlatFrames(frames, boardName)` into `useNativeClimbRender` from `BoardImageNative` and `ClimbListThumbnail`. Smallest diff, keeps the Rust parser dumb, and puts one flatten in front of every mobile static render. Needs an eyeball pass: it changes what *every* mobile thumbnail of a multi-frame climb shows, and the union can be busy (median 42 lit holds, worst case 225 on a 692-placement layout).
2. **Teach the Rust parser the grammar.** Port `parseFramesSegments` / `accumulateFramesToMaps` into `frames_parser.rs` so the renderer understands `"`, `x`, and absolute frames itself. More faithful, but it duplicates the grammar in a second language and needs its own oracle test.

Option 1 first, unless someone wants the animation on-device later, in which case 2 is the real answer.

## Done when

- A multi-frame climb's mobile thumbnail and its web card show the same holds.
- A regression test pins the two paths against each other for at least one catalog climb.
- Someone has looked at "Happy New Year! 🎉" on a phone and agreed it reads as a route.

marcodejongh:
## Summary

This fixes mobile in-app climb thumbnails — climb list rows, board carousels, reaction menus, wall hero, device cards, and board discovery cards — so multi-frame routes and circuits show every hold instead of just the first frame. It flattens the `frames` string with the existing `toFlatFrames` helper inside `useNativeClimbRender`, the single choke point every mobile static render goes through, matching what web cards, the backend OG share card, and the mobile share card already do. Single-frame climbs (the vast majority of the catalog) come out byte-identical so no cached thumbnail PNGs are invalidated, but 709 multi-frame climbs will render visually busier, and the author flags this needs a device eyeball before merge. The Rust renderer core is left untouched so the fix ships JS-only via OTA.

---

## What

Mobile in-app static climb renders — `ClimbListThumbnail`, `BoardImageNative`, and everything above them (climb list rows, `AccessoryClimbThumbnail`, `SwipeBoardCarousel` including its neighbor peeks, `ClimbReactionMenu`, `WallHeroStage`, `DeviceCard`, board discovery cards/forms) — handed the raw, possibly multi-frame `frames` string straight to the Rust renderer. `frames_parser.rs` only ever reads frame 0 (splits on the first comma and drops the rest), so a multi-frame climb (route or circuit — 709 in the catalog) showed only its first frame in the mobile app.

Web climb cards, the backend OG share card, and the mobile share card already collapse a multi-frame climb to the union of every frame via `toFlatFrames` (`@boardsesh/board-constants/hold-states`). Mobile in-app thumbnails and the play-drawer board were the one static surface left showing a fragment instead of the whole route.

This flattens once, at the single choke point every mobile static render goes through — `useNativeClimbRender` — instead of at each of the two call sites. `flatFrames = useMemo(() => toFlatFrames(frames, boardName), [frames, boardName])` feeds the cache key, the effect's `!frames` guard, and the native config JSON's `frames` field. Raw `frames.length` is kept for the Sentry `framesLength` extra.

Closes #3988

## Why this is safe for the 99.9% single-frame case

`toFlatFrames` has a fast path: a string with no `,` and no `x` returns unchanged. So:

- Single-frame climbs (the overwhelming majority of the catalog) produce a byte-identical string, which means the cache key formula is unchanged and **no existing cached thumbnail PNGs are invalidated**.
- PlayDrawer's per-frame animation override (`currentFrameOverride`, from `useClimbFrames` → `accumulatedMapsToFrameStrings`, which only ever emits comma-free `p{id}r{code}` strings) also passes through the flatten untouched, so the in-progress play view still animates frame-by-frame instead of collapsing to the union.

## Visual change (intentional, needs sign-off)

This is a real visual change for the 709 multi-frame climbs in the catalog: 611 of them render busier than before (median goes from 20 to 42 lit holds, worst case 225 on "Happy New Year! 🎉", `F142693E7BF54FBEA3DC936E263DEEF5`). This direction was maintainer-settled in #3986 ("Option 1 first" — flatten at the call site, matches what web/backend/share-card already show) and is exactly what the issue's "done when" asks for: mobile thumbnails now match web cards for the same climb.

**Visual evidence is owed, not attached.** This dev host is Linux and can't run the iOS/Android simulator screenshot flow (`vp run check:mobile-simulator` / `vp run mobile:screenshot` are macOS-only). @marcodejongh — please eyeball "Happy New Year! 🎉" in the mobile climb list on a device/simulator before merging; this PR should not be auto-merged.

## What did NOT change

- `packages/board-renderer/core/src/frames_parser.rs` — untouched, on purpose. Editing the Rust core (even a comment) risks moving the native fingerprint and would force a non-OTA-compatible build; the fix stays JS-only so it rides OTA.
- `ClimbListThumbnail.tsx` / `BoardImageNative.tsx` — still pass the raw `frames` prop straight through; the flatten lives entirely inside the hook.

## Rebase follow-up

`main` grew a second consumer of the raw `frames` string between the flatten and the renderer: the `no_matching_holds` guard (#5098) parses frame 0's lit placement ids and hands them to both the skip/evict decision and `withLitHoldGeometry` inside `getBoardConfig`. Those ids now come off `flatFrames` too.

The renderer is given the union of every frame, so frame 0 alone under-reported what actually gets drawn. The guard judged holds nobody paints (and would skip a route whose frame 0 sits off this board while its union does not), and the Aura outline pass attached traced silhouettes to frame 0's holds only, leaving the rest of the route on plain fallback rings — the same "fragment of the route" symptom this issue is about. Single-frame strings are unaffected: `toFlatFrames` returns them untouched, so `litHoldIds` comes out identical.

That guard is also why the three new tests went red on CI. The suite mocks `getBoardRenderData` with one hold per placement id, and a config matching NONE of the lit ids is answered before the render ever starts, so the fixture climbs never reached `renderHoldsOverlay` and asserted 0 calls. The fixture board now carries every id those strings light, across all ten frames.

One thing worth an eye during the device pass: for Kilter, `toFlatFrames` re-emits the union using the canonical product-7 role codes (42/43/44/45) rather than the catalog string's product-1 codes (12/13/14/15). `HOLD_STATE_MAP.kilter` carries both sets with identical colours and the hook ships the whole map to the renderer, so this is a no-op for what you see — but it is why the pinned expectation in test 1 is not a substring of the input.

## Tests

Extended `packages/mobile/src/hooks/__tests__/use-native-climb-render-race.test.tsx` (already renders the real hook against a mocked native module) with three cases:

1. A real multi-frame catalog climb (`aurora-frames-oracle.json`'s "FRE 20250608", 10 absolute frames) — asserts the `frames` field in the JSON config handed to `renderHoldsOverlay` equals `toFlatFrames(raw, boardName)` exactly, and that the cache key matches. This is the issue's requested pin of the mobile path against the web path.
2. A single-frame string passes through byte-identical, with the exact pre-change cache key — guards against invalidating cached thumbnails for the rest of the catalog.
3. A comma-free playback-snapshot-shaped string (the `p{id}r{code}`-only shape `PlayDrawer`'s `currentFrameOverride` produces) is left untouched — guards the animation path.

## Validation

Rebased onto `main` at `4213de4a0` (34 commits), clean — no conflicts.

- `vp test run --project mobile src/hooks/__tests__/use-native-climb-render-race.test.tsx --reporter=agent` — 20/20 pass.
- `vp run test:mobile` (full suite) — 8268/8268 pass across 797 files.
- `vp check` — 0 errors. The 339 warnings are pre-existing repo-wide; none in the two touched files.
- `vp run check:mobile-bundle` — `[mobile-bundle] android bundle: OK`.
- `vp run check:pr-test-plan` against this description — `OK — 3 steps, risk 2/5`.
- Mutation check: putting the raw `frames` back into the native config JSON turns test 1 red on the exact union string, so the pin fails for the reason it claims to.

Still owed: the device eyeball pass below. This host is Linux, so the simulator screenshot flows stay unavailable.

## Release Notes

- Mobile: route and circuit thumbnails in the climb list now show every hold in the route, not just the first frame.

- Please do the device eyeball pass on "Happy New Year! 🎉" (or any multi-frame route) called out in the issue's done-when before merging — this changes the default look of 709 climbs' thumbnails.

## Test plan

1. Signed in, Kilter selected.
2. Climbs tab, open a multi-frame route, thumbnail shows every hold.
3. Open a boulder, its thumbnail looks the same as before.

## Risk

Risk: 2/5 — changes the look of 709 route thumbnails; single-frame climbs are byte-identical.

### DIFF
--- a/packages/mobile/src/hooks/__tests__/use-native-climb-render-race.test.tsx
+++ b/packages/mobile/src/hooks/__tests__/use-native-climb-render-race.test.tsx
@@ -1,6 +1,7 @@
 // @vitest-environment jsdom
 import { describe, it, expect, vi, beforeEach } from 'vitest';
 import { renderHook, act, waitFor } from '@testing-library/react';
+import { toFlatFrames } from '@boardsesh/board-constants/hold-states';
 
 // In-flight render race regression (play-drawer "unlit holds"): a slow native
 // render must not clobber nativeRender state after the hook's props have moved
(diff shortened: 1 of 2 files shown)


marcodejongh:

marcodejongh:

Conflicting: packages/mobile/src/hooks/use-native-climb-render.ts.

marcodejongh:
🤖 Session `4634-f0` (Fable) is taking this PR over as part of the route / create-screen campaign for #4634: rebase onto current `main` (stacked on #5107 where the create-screen files collide), CI back to green, review threads answered, test plan + risk in tester voice, then ready for review. Peer loops: please stand off this branch; say so here if you are already mid-flight.

marcodejongh:
Rebased onto `main` at `4213de4a0` (34 commits, clean) and pushed as `18f2000`.

**Why the three new tests were red.** Not drift in the module loader — the `no_matching_holds` guard from #5098. It parses frame 0's lit placement ids and skips any config whose ids match none of the board's holds, and it runs *above* the overlay-cache lookup, so `renderHoldsOverlay` was never reached: 0 calls. The suite's fixture board only carries `1100/1200/1300/1400`, and the new cases replay a real 10-frame catalog climb ("FRE 20250608") plus a 1500-1502 PlayDrawer snapshot. The fixture board now carries every id those strings light, across all ten frames.

**One real bug the rebase surfaced.** That guard was a second consumer of the raw `frames` string sitting between the flatten and the renderer, so the flatten was no longer the single choke point this PR claims. `litHoldIds` now comes off `flatFrames`. The renderer gets the union of every frame, so frame 0 alone under-reported what is actually drawn — the guard judged holds nobody paints (and would skip a route whose frame 0 sits off this board while its union does not), and `withLitHoldGeometry` attached traced silhouettes to frame 0's holds only, leaving the rest of the route on plain fallback rings. That is the same "fragment of the route" symptom #3988 is about. Single-frame strings are unaffected: `toFlatFrames` returns them untouched, so `litHoldIds` is identical.

**Validation.** `typecheck:mobile` clean · the race suite 20/20 · full `test:mobile` 8268/8268 across 797 files · `vp check` 0 errors · `check:mobile-bundle` OK · `check:pr-test-plan` now `OK — 3 steps, risk 2/5` (the gate was failing for a missing `## Test plan` and `## Risk`; both added). Mutation-checked: putting raw `frames` back into the native config JSON turns test 1 red on the exact union string.

Still owed, unchanged: the device eyeball pass on a multi-frame route. This host is Linux, so the simulator flows stay unavailable.

Issue #4000, opened 2026-07-30
playlistClimbs.totalCount ignores the board/layout/size filters it applies to climbs


marcodejongh:
Split out of #3891 / PR #3997 so a GraphQL semantics decision doesn't ride along with a bug fix.

`playlistClimbs` counts every row in `playlist_climbs` for the playlist:

```ts
const countResult = await db
  .select({ count: sql<number>`count(*)::int` })
  .from(dbSchema.playlistClimbs)
  .where(eq(dbSchema.playlistClimbs.playlistId, playlistId));
```

That count ignores the board / layout / size filters the specific-board branch then applies to `climbs`. So `totalCount` can say 12 while `climbs` holds 4 — anything in the playlist that isn't on the requested board, layout, or size is counted but not returned. Before PR #3997 the MoonBoard case made it stark: `totalCount: 2`, `climbs: []`.

Not a live bug today. No production client reads `playlistClimbs.totalCount` — checked across mobile, web and `@boardsesh/playlists-react`; the only consumed `totalCount` is the unrelated one on `allUserPlaylists`. So this is a trap for the next caller rather than a broken screen.

Two defensible resolutions, and picking one is the actual work:

- Make `totalCount` respect the same filters as `climbs`, so it means "how many climbs this query would return across all pages". Consistent, and what a paginating client would assume.
- Keep it as "how many climbs are in this playlist" and document it as such in the schema, since that's a genuinely useful number for a header ("4 of 12 on this board").

Whichever way it goes, the schema description should say which one it is.

Low priority: latent, no user-visible symptom, and it's on a resolver shared with the web surface being replaced by the RN web deploy.

marcodejongh:
🤖 Claimed by bug-batch loop `josh` (batch 1, session 4448b370). Plan → draft PR incoming. Parallel loop: please pick a different issue.

marcodejongh:
Draft PR: https://github.com/boardsesh/boardsesh/pull/4079

marcodejongh:
Fixed by #4079. The merged resolver now computes totalCount with the same board/layout/size/hold-set joins and filters as the returned climb page, with regression coverage for specific-board, MoonBoard, size-mismatch, and all-boards cases.

PR #4294, opened 2026-08-11
fix(backend): reorder playlist climbs in the index space clients render

### LINKED ISSUE #4012: All-boards playlistClimbs joins climbs on UUID only, and dropped refs desync counts and mobile reorder indices
Found during the playlist code review that produced #4010. Sibling of #4000 (totalCount vs board filters).

**Join lacks a `board_type` predicate** — `packages/backend/src/graphql/resolvers/playlists/queries/playlist-climbs.ts` (all-boards ref query) joins `tables.climbs` on `uuid` alone, unlike the specific-board path and unlike `hydrateClimbsByRefs`, which keys by `${boardType}:${climbUuid}` because "climb UUIDs can collide across boards in principle". A cross-board UUID collision would fan one `playlist_climbs` row into N page rows and break the `pageSize + 1` paging. Practically safe today (MoonBoard uses dashed UUIDv5, Aurora compact 32-char), but the join also can't use a `(board_type, uuid)` index. `playlist_climbs` doesn't store `board_type`, so a real fix likely means persisting it (or deriving it at add time).

**Dropped refs desync derived numbers** — rows whose UUID has no `board_climbs` match are silently dropped by the inner join and by `hydrateClimbsByRefs` (`helpers/hydrate-climbs.ts` — `if (!row) continue;`), while `totalCount` still counts them. Beyond the "showing X of Y" mismatch, mobile reorder computes `newIndex` against the *rendered* list (`packages/mobile/app/(tabs)/discover/[playlist_uuid].tsx:228`) while the server splices against the full `playlist_climbs` list (`mutations.ts:440-453`) — so a single orphaned ref shifts every reorder below it by one.

marcodejongh:
Closes #4012

## The bug that was real

`playlistClimbs` inner-joins `board_climbs`, so a `playlist_climbs` row whose `climb_uuid` no longer resolves (catalog sync lag, a climb that left the catalog) never reaches the client — `hydrateClimbsByRefs` drops it too. The mobile playlist editor computes `newIndex` against that rendered list (`packages/mobile/app/(tabs)/discover/[playlist_uuid].tsx`), while `reorderPlaylistClimb` spliced against the **full** `playlist_climbs` list.

One orphaned ref ahead of the target was enough to put every move below it one slot off, and the wrong order persisted. Concretely, with `[ghost, A, B, C]` in the table and `[A, B, C]` on screen, dragging A to the bottom sent `newIndex: 2` and produced `[ghost, B, A, C]` — A lands in the middle, and the user's next drag starts from a list that already disagrees with what they did.

Fixing it server-side also fixes the app versions already in the field: every shipped client has always numbered its drags over the rendered list, so no client change is needed or possible via OTA-skew.

## The fix

`computePlaylistReorderWrites` now takes the set of refs that resolve to a climb. It maps `newIndex` through the visible sublist, finds the insertion point in the full list that leaves exactly that many visible rows ahead of the moved row, then renumbers dense `0..n-1` and returns only changed rows — the contract callers already had.

- Omitting the set, or passing one that covers every row, reproduces the old behaviour byte for byte. A test asserts that equivalence across five different moves, and the existing math tests are unchanged.
- A move to the visible end lands just after the last visible row, not after trailing invisible ones — same result on screen, fewer position writes, orphans stay put.
- Moving a row that isn't visible can't come from the UI, so it falls back to full-list semantics rather than inventing a slot.
- The resolver reads visibility with a second statement inside the same transaction instead of turning the locked select into a `LEFT JOIN`: Postgres rejects `FOR UPDATE` on the nullable side of an outer join. Playlists are small and `board_climbs.uuid` is the primary key, so it's an index scan over a handful of ids.

The index space being honoured is the **all-boards** `playlistClimbs` list — the one the mobile playlist screen renders (it omits `boardName` on purpose so off-board climbs stay listed, dimmed). That assumption is now written at the resolver: a future editor that reordered from a board-scoped list would be numbering against a shorter list again, and the mutation input carries no board argument to infer that from.

## The two other complaints in the issue

**`totalCount` counting dropped refs** was already fixed on `main` by #4000: the count and the ref-rows query in `fetchAllBoardsClimbs` now share one inner join. Nothing to redo.

**Cross-board UUID collision** can't happen. `board_climbs.uuid` is the table's PRIMARY KEY (`packages/db/src/schema/boards/unified.ts`), so a uuid names at most one climb across every board — the uuid-only all-boards join cannot fan one playlist row into several page rows, and it already rides the PK index. **No `board_type` column was added to `playlist_climbs`**: it would be redundant against that key, and it would cost a migration plus a backfill plus new writers in `addClimbToPlaylist` and the Aurora circuit sync. Instead the invariant is written down where the join lives, and `playlist-climbs-uuid-join-invariant.test.ts` pins the key shape, so a future move to a composite `(board_type, uuid)` PK fails there loudly rather than corrupting paging quietly.

Orphaned `playlist_climbs` rows are left alone — they can be transient (catalog re-sync lag), and deleting user data isn't a call to make here.

## Validation

- `vp run typecheck:backend` — pass on the rebased branch.
- `vp check packages/backend/src/graphql/resolvers/playlists/mutations.ts` — clean.
- `vp test run --reporter=agent --project backend playlist-reorder` — 30 tests pass across the math, mocked-resolver, and real-DB files (run before the rebase; backend tests are left to CI here because the shared worker databases are contended locally).
- New coverage:
  - `playlist-reorder-math.test.ts` — orphan ahead of the target, orphan below it, move to the visible end with a trailing orphan, clamp to the last *visible* index, gappy positions plus an orphan, no-op in visible space, only-visible-row, parity with full-list behaviour, unresolvable-moved-row fallback.
  - `playlist-reorder-integration.test.ts` (real DB) — a playlist with a `ghost-climb` ref: drag to the end, drag to the front, and a no-op that is only a no-op on screen. The existing cases now seed `board_climbs` so they exercise the resolvable path rather than the fallback.
  - `playlist-climbs-uuid-join-invariant.test.ts` — `board_climbs.uuid` is a single-column primary key.

The new catalog seeds use `ON CONFLICT DO NOTHING` + `DELETE ... WHERE uuid IN (...)` rather than `TRUNCATE board_climbs CASCADE` (an AccessExclusiveLock on a table every backend suite shares), and the mocked transaction picks its result by select projection rather than call order.

Pre-existing flake spotted while validating, not fixed here: `playlist-reorder-integration.test.ts` and `playlist-climbs-total-count.test.ts` both `TRUNCATE ... playlists RESTART IDENTITY` in `beforeEach` and both build a playlist at `id = 1`. Whenever they land in the same worker database they knock each other over with "Playlist not found or access denied". Both do it on `main` today; worth a separate cleanup if it starts showing up in CI.

## Two calls made here, say the word to reverse either

1. **No `board_type` on `playlist_climbs`** — declined above on the grounds that the global PK already makes the join unique. It's a migration + backfill + two writers, so if you want the column anyway (denormalised provenance, cheaper composite index, survives a future PK change) it wants its own PR.
2. **Where a row lands when dragged to the bottom of a list with trailing orphans** — it goes after the last visible row, leaving the orphans below it. Visually identical either way; this way writes fewer rows.

## Release Notes

Dragging a climb in a playlist now drops it exactly where you let go, even when the playlist holds a climb the catalog has lost track of.

## Test plan

Backend-only change; covered by the unit, mocked-resolver, and real-DB tests listed above. Manual pass on mobile: open a playlist with an out-of-catalog ref, enter edit mode, drag a climb to the top / bottom / middle, leave edit mode (which refetches) and confirm the order matches what was dropped.

### DIFF
--- a/packages/backend/src/__tests__/playlist-climbs-uuid-join-invariant.test.ts
+++ b/packages/backend/src/__tests__/playlist-climbs-uuid-join-invariant.test.ts
@@ -1,0 +1,32 @@
+import { describe, it, expect } from 'vite-plus/test';
+import { getTableConfig } from 'drizzle-orm/pg-core';
+import * as dbSchema from '@boardsesh/db/schema';
+
+// #4012 asked whether the all-boards playlistClimbs join — `board_climbs.uuid =
+// playlist_climbs.climb_uuid`, with no board_type predicate — can fan one
+// playlist row into several page rows and break the `pageSize + 1` paging, and
+// whether `playlist_climbs` should therefore start persisting `board_type`.
+//
+// It cannot, because `board_climbs.uuid` is the table's primary key: a uuid
+// names at most one climb across every board, so the uuid-only join is already
+// both unique and index-backed. That makes the extra column redundant today —
+// but the argument rests entirely on the shape of the key. Should `board_climbs`
+// ever move to a composite `(board_type, uuid)` primary key, the uuid-only joins
+// in `resolvers/playlists/queries/playlist-climbs.ts` (and the resolvable-ref
+// lookup in `reorderPlaylistClimb`) stop being safe — this test is what fails
+// loudly at that moment, instead of the paging quietly going wrong.
+describe('board_climbs uuid is a global primary key', () => {
+  const tableConfig = getTableConfig(dbSchema.boardClimbs);
+
+  it('declares uuid as a single-column primary key', () => {
+    const uuidColumn = tableConfig.columns.find((column) => column.name === 'uuid');
+    expect(uuidColumn).toBeDefined();
+    expect(uuidColumn?.primary).toBe(true);
+  });
+
+  it('has no composite primary key that would scope uuid to a board', () => {
+    expect(tableConfig.primaryKeys).toEqual([]);
+    const otherPrimaryColumns = tableConfig.columns.filter((column) => column.primary && column.name !== 'uuid');
+    expect(otherPrimaryColumns).toEqual([]);
+  });
+});
--- a/packages/backend/src/__tests__/playlist-reorder-integration.test.ts
+++ b/packages/backend/src/__tests__/playlist-reorder-integration.test.ts
@@ -55,16 +55,20 @@ describe('reorderPlaylistClimb — real DB (#3607)', () => {
     `);
     // The resolver reads which refs resolve to a catalogued climb (#4012), so the
     // catalog rows have to exist for these moves to run in visible-index space.
-    await db.execute(sql`TRUNCATE TABLE board_climbs CASCADE`);
+    // Seeded and cleaned by uuid rather than by truncating board_climbs: the
+    // worker database is shared with every other backend suite, and an
+    // AccessExclusiveLock on the catalog is both wider than this test needs and
+    // a deadlock magnet.
     await db.execute(sql`
       INSERT INTO board_climbs (uuid, board_type, layout_id, name)
       VALUES ('climb-a', 'kilter', 1, 'A'), ('climb-b', 'kilter', 1, 'B'), ('climb-c', 'kilter', 1, 'C')
+      ON CONFLICT (uuid) DO NOTHING
     `);
   });
 
   afterAll(async () => {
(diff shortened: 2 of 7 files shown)


marcodejongh:

PR #4296, opened 2026-08-11
fix(backend): merge concurrent adds into setQueue instead of overwriting them

marcodejongh:
Refs #3933

## The bug

A wholesale queue replace threw away any climb a party member added while the payload was being composed. Nothing detected the loss — the server's own post-write state stayed internally consistent, so the client hash watchdog never fired. Same blind spot as #3906, whose fix (PR #3932) made sequence allocation atomic but deliberately left `setQueue` writing with `CAS_ANY_VERSION` and carved this residual out as #3933.

Exposed paths on `main`: mobile playlist activation (`use-playlist-activation.ts:404`, `:507`), party session seed (`use-session-commands.ts:156`), empty-room re-seed (`use-session-realtime.ts:492`), `appendGeneratedSession`, and offline reconciliation. The issue body's headline case — web's drag-to-reorder pushing a full `setQueue` — is gone: after W-16 (#4435) `packages/web` has no `setQueue` caller at all, which is also why the issue's "direction 1" (make web send the `reorderQueueItem` delta) buys nothing today.

## The fix

`setQueue` takes an optional `baselineSequence` — the last sequence the caller had **applied**. When present, the resolver replays the queue-event buffer over that window, re-appends the peer adds the caller never saw, and writes with a versioned CAS so an add racing the merge itself conflicts and retries.

Conservative by construction. It degrades to the exact legacy overwrite rather than merging on partial evidence:

- **Buffer lag.** `storeEventInBuffer` is fire-and-forget and runs strictly *after* the CAS commits, so a peer's add can sit in committed state with no buffered `QueueItemAdded` yet. The merge checks the newest buffered sequence against the committed sequence, re-reads once after 25 ms, then degrades. Merging on a lagging buffer would drop exactly the item this fix exists to save.
- **Buffer coverage.** If the oldest buffered entry no longer reaches back to the baseline (evicted past 100 events, expired past the 5-minute TTL), the window can only be described partly — degrade.
- **Redis off.** No buffer at all — degrade, without ever calling `getEventsSince`.
- **Retry exhaustion.** `withQueueVersionRetry` throws after `MAX_RETRIES`. That throw is caught and the unversioned write runs, so a contended session can't turn a wholesale replace into a brand-new user-facing error it never had.

It also skips buffered adds whose `clientId` equals the caller's `connectionId` (non-null comparison only — `clientId` is coerced to `null` for anonymous connections and two nulls must never match). A client's own `addQueueItem` only advances its sync gate on the subscription echo, so a `setQueue` fired before that echo has a baseline older than its own add; re-appending it would resurrect a climb the climber just deliberately replaced away.

## Wire compatibility

Backward compatible in **both** directions:

- **New backend, old client.** The old client sends no `baselineSequence`, `mergeBaseline` resolves to `null`, and the resolver takes the byte-identical legacy unversioned overwrite. Adding an optional argument to an existing field is a non-breaking schema change.
- **New client, old backend.** The baseline rides a **separate** `SET_QUEUE_WITH_BASELINE` document, because an unknown argument is a document-level GraphQL validation error, not a variables-level one — passing the variable as `undefined` does not help, and folding `baselineSequence` into `SET_QUEUE` would hard-fail *every* `setQueue` against a backend that hasn't deployed the argument yet. `SET_QUEUE` stays byte-identical; callers select the new document only when `getBaselineSequence()` returns a number, and retry once on the legacy document when the server rejects it at validation. Validation rejections never reach the resolver, so the retry costs one round trip and consumes no extra rate-limit budget.

Deploy ordering is therefore a preference, not a hard gate: land the backend on Railway first so the first mobile OTA doesn't burn a wasted round trip per replace, but an OTA that arrives early degrades to the legacy behaviour rather than breaking.

## Accepted residuals (why this says "Refs", not "Closes")

Documented in the resolver comment block and in `docs/websocket-implementation.md`:

- Adds that reach the queue via a FullSync-publishing mutation (`replaceQueueItem`, the controller navigation paths) carry no `QueueItemAdded` event, so the merge can't see them and they're still lost inside the window.
- Every degrade path re-opens the original race for that call.
- The buffer is bounded at 100 events / 5 minutes and is Redis-only.
- The baseline is captured when the caller hands over the payload, so anything the caller applied between composing its own payload and calling `setQueue` is still outside the merged window.

Closing #3933 fully still means the mergeable-queue protocol rewrite the issue body describes. This narrows the race; it doesn't end it.

## Review pass on adoption (2026-08-19)

Rebased onto current `main` (clean) and re-reviewed. Three changes on top of the original commit:

- **The baseline was read after the join await.** `createQueueMutations.setQueue` read `getBaselineSequence()` *after* `await resolveCore(...)`, which on mobile can await a session join. A peer add landing during that wait advanced the sync gate past what the payload contained, the server saw no window to merge, and the replace dropped that climb anyway — the exact failure this PR exists to prevent, through a narrower door. Now read synchronously at call time, where the payload was composed. Regression test added.
- Dropped `cannot query field` from the deploy-skew matcher: that is graphql-js's unknown-**field** error, not the unknown-**argument** error this guards.
- `docs/websocket-implementation.md` still stated `setQueue` was unconditionally exempt from the versioned CAS with the peer-add overwrite tracked as #3933. Rewritten to describe the merge path, its four degrade conditions, and the remaining residual.

## Validation

- `vp run typecheck` — clean (full monorepo).
- `vp check` — the only errors are pre-existing module-resolution noise on `main` (`@gorhom/bottom-sheet` in the web-runtime shim, `yaml` in root `scripts/`). Nothing in the changed files.
- `vp test run --project queue-react` — 51 passed (8 for the baseline document: legacy with no getter, legacy with null, baseline document with the sequence, baseline `0` not treated as absent, read at call time not construction, **read before the join await**, deploy-skew retry on the legacy document, real server error propagates without a silent retry).
- `vp test run --project mobile queue-provider-sync-gate` — 20 passed.
- Backend `set-queue-concurrent-add-merge` (14 cases: peer add survives and reaches the broadcast FullSync, add-then-remove doesn't resurrect, no duplicate when the climb is in both payload and buffer, caller's own un-echoed add not resurrected, no-baseline keeps the legacy overwrite, each degrade path, retry exhaustion falls back to the unversioned write, slow-LPUSH re-read rescues the merge) is left to CI — ~25 parallel worktrees share the fixed `boardsesh_backend_test_w<N>` DBs on :5433, so local backend runs report other sessions' truncations.

JS/TS only, no native deps — rides OTA.

## Open question for @marcodejongh

The issue body leans toward "direction 1" (make the reorder path send the `reorderQueueItem` delta) as most of the value for a fraction of the cost. That direction is now moot — `packages/web` has no `setQueue` caller left and mobile already sends the reorder delta — so this PR implements direction 3 (merge-on-write) instead. Taken as a coordinator decision, veto-able. Two smaller calls ride along: merged survivors are **appended at the end** (a peer's climb added during a 40-climb playlist activation lands at the bottom rather than after the current climb), and #3933 stays **open** with this PR as mitigation rather than being closed.

## Release Notes

Queues you share with your crew stop losing climbs: when someone adds a climb while you're activating a playlist or reseeding the queue, their climb now sticks around instead of vanishing.


marcodejongh:

Issue #3814, opened 2026-07-20
PostHog: preview deploys pollute prod project via substring hostname gate


marcodejongh:
Follow-up from #3808 (which fixed the identical bug for Sentry).

`packages/web/app/lib/analytics.ts:27` gates the production PostHog client with a **substring** hostname match:

```ts
if (!window.location.hostname.includes('boardsesh.com')) return null;
```

Every branch/preview deploy runs at `<pr>.preview.boardsesh.com`, which contains `boardsesh.com`, so **preview browser sessions send events into the production PostHog project** — the same class of leak #3808 closed for Sentry.

### Fix
Tighten to an exact-host allowlist, mirroring the Sentry fix in #3808:

```ts
const PRODUCTION_HOSTS = new Set(['boardsesh.com', 'www.boardsesh.com']);
if (!PRODUCTION_HOSTS.has(window.location.hostname)) return null;
```

### Notes
- Confirm first whether preview analytics are *wanted* (some teams keep them, separated by an environment/`$host` property) before hard-disabling. If kept, tag preview events distinctly instead.
- Same substring pattern may exist for the Vercel Analytics dual-write path — check `track()` in the same file.

PR #4299, opened 2026-08-11
fix(web): stamp environment=production on PostHog super properties

### LINKED ISSUE #3945: Web PostHog events carry no `environment` property — an environment filter silently drops all web volume
## What

After #3930, the backend stamps `environment` on every PostHog event and mobile carries it as a super property. **Web does not.**

Verified live against project 412845: `js`-lib events are **100% `(none)`** for `environment` over the last 30 days — 102,331 events.

## Why it matters

Once backend and mobile are tagged, the natural way to exclude preview/localhost noise is a PostHog filter of `environment = 'production'`. That filter now **silently drops all web volume** (~7.5% of events). The dashboards keep rendering, the numbers just quietly get smaller — the worst shape of analytics bug.

## Options

1. Register `environment` as a super property on the web client too, matching mobile. Consistent, and makes `environment = 'production'` correct everywhere.
2. Standardise queries on `environment != 'preview'` instead, so untagged events fall through as production.

Option 1 is preferable — relying on absence-means-production breaks the moment a fourth surface appears.

## Notes

This is *not* the substring-hostname flaw from #3814. That one is already fixed on `main`: #3895 landed `isProductionHost()` exact-match in `production-hosts.ts` and `analytics.ts` calls it. This is a separate, additive gap.

Found while rescuing #3930; `claude-review` independently flagged the same thing on that PR. Left out of #3930 to avoid pulling a third package into the diff.

marcodejongh:
## Summary

Closes #3945.

Web's PostHog client never registered an `environment` super property, unlike mobile (`registerAppEnvironment()`) and backend (per-event `properties.environment`). Verified live in the issue against project 412845: `js`-lib events were **100% `(none)`** for `environment` over the last 30 days (102,331 events).

Once a dashboard filters on `environment = 'production'` — the natural way to exclude preview/localhost noise now that backend and mobile are tagged — that filter silently drops **all** web volume (~7.5% of events). The dashboards keep rendering; the numbers just quietly get smaller.

## What changed

Option 1 from the issue (register `environment` as a super property on web, matching mobile), per the issue author's stated preference over standardising queries on `environment != 'preview'`.

- `getPosthog()` calls `registerWebEnvironment(posthogClient)` right after constructing the client, registering `{ environment: 'production' }` via `client.register(...)` — the same persistent-super-property call mobile uses.
- The value is hardcoded rather than read from an env var, deliberately. `getPosthog()` is already gated by `isProductionHost()` (exact-match against `boardsesh.com` / `www.boardsesh.com`), and `posthogClient` is only ever constructed when that gate passes, so `'production'` is correct by construction. No new build-time env var is introduced — the May 2026 blackout where CI's `vercel build` never received `NEXT_PUBLIC_POSTHOG_KEY` is exactly the failure mode this avoids re-creating. Preview deploys (`<pr>.preview.boardsesh.com`) never construct a client at all, so they can't be mislabelled as production; they emit nothing, same as before. A comment marks the spot to switch to a dynamic resolve if that gate is ever relaxed.
- `reset()` re-registers `environment` after `core.reset()`. PostHog's `reset()` clears every registered super property, but `getPosthog()` caches the singleton, so without a re-register a party-profile reset (`party-profile-context.tsx`) would silently drop the tag for the rest of the page session. Only re-registers when `core.reset()` actually forwarded to a real client, so the admin-page skip path doesn't construct a client that wouldn't otherwise exist.
- Registration is best-effort: both a rejected promise and a synchronous throw are swallowed, so a storage failure can never take down analytics init (and with it every `track()` call on the page). `register()` is declared `async` in `@posthog/core` 1.46.1, so today it can only reject — the sync guard keeps that from being an invisible version coupling, and matches mobile's `registerAppEnvironment`, which has always caught both.

No changes needed in `packages/web/app/lib/analytics.server.ts` (Vercel only, no PostHog client) or `production-hosts.ts` (the hostname gate is already the exact-match fix from #3895, not the substring flaw from #3814).

Ordering is safe: `@posthog/core` routes `register()`, `reset()` and `capture()` through the same `wrap()` init queue, so the registration always lands ahead of the first captured event even before the SDK finishes initialising.

## Release Notes

## Test plan

- `vp run typecheck:web` — passes (confirms `register()` is in `posthog-js-lite`'s public typings, inherited from `@posthog/core`'s `PostHogCoreStateless` as `Promise<void>`, so no structural cast is needed).
- `vp check` — clean on both touched files; format passes across all 4828 files. The 4 remaining lint errors are pre-existing on `main` in `packages/db/scripts/dedupe-gym-locations-args.test.ts` and untouched here.
- `vp test run --project web app/lib/__tests__/analytics.test.ts` — 20/20 pass, including the 5 new cases:
  - registers `{ environment: 'production' }` on first PostHog client init
  - never registers on a preview hostname (the client isn't constructed at all)
  - re-registers after `reset()`
  - a rejected `register()` promise neither throws nor blocks a subsequent `capture()`
  - a synchronously throwing `register()` neither throws nor blocks a subsequent `capture()`

## Caveat for dashboard owners

This can't fix history: web events captured before this deploys still show `environment = (none)`. A filter of `environment = 'production'` keeps undercounting web until the query's date range moves past the deploy date. No event backfill is planned or in scope here.

### DIFF
--- a/packages/web/app/lib/__tests__/analytics.test.ts
+++ b/packages/web/app/lib/__tests__/analytics.test.ts
@@ -7,6 +7,7 @@ const mocks = vi.hoisted(() => ({
     capture: vi.fn(),
     flush: vi.fn(async () => {}),
     identify: vi.fn(),
+    register: vi.fn().mockResolvedValue(undefined),
     reset: vi.fn(),
     setPersonProperties: vi.fn(),
   },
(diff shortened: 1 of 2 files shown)


marcodejongh:

PR #4300, opened 2026-08-11
feat(mobile): on-screen sheet detent readout to unblock #3922

### LINKED ISSUE #3776: [UI] Climb: Filter "Show" button is weirdly high

### LINKED ISSUE #3922: mobile: replace the sheet's magic-constant detent-height estimate with a runtime onLayout measurement
## Problem

`useSheetColumnStyle` (`packages/mobile/src/components/use-sheet-column-style.ts`) pins the iOS sheet's content column to a JS-computed *guess* at the native detent height, built from `useWindowDimensions()` plus a handful of tuned constants (`SHEET_TOP_CHROME_PT`, `SHEET_TOP_GAP_PT`, an iOS-26-only gate). Every attempt to fix this by adjusting the formula — the original #3330 fix, and PR #3611's attempted correction — has either erred in the wrong direction or, per the #3776 diagnosis, missed the actual bug entirely. The real gap on #3776 (~142pt on iPhone SE 3) is far larger than any of the formula's error terms and appears invariant to them (see the #3330-vs-#3776 comparison on #3776: a ±22pt swing in the computed column height across formula versions produced no visible change in the dead gap).

Two independently-sourced models of how SwiftUI's `.fraction(f)` detent resolves (window-relative vs. `maximumDetentValue`-relative) disagree by 30–50pt depending on device, and neither model, even perfectly tuned, explains the #3776 gap. Chasing more constants isn't going to converge.

## Proposed fix

Stop estimating. Measure.

Render a probe view as a sibling inside the sheet root:

```tsx
<View
  style={StyleSheet.absoluteFill}
  pointerEvents="none"
  onLayout={(e) => setMeasuredHeight(e.nativeEvent.layout.height)}
/>
```

`position: absolute` with zero insets resolves against the parent's Yoga-reported height — which `@expo/ui`'s `RNHostView` already populates via `setViewSize` once SwiftUI lays out the sheet. Drive the content column's height from that measured value; fall back to the current constant-based estimate only when the probe reports `0` (first frame, before layout has run).

This self-corrects across device, iOS version, and detent fraction with zero tuned constants, costs one extra layout pass, needs no native change, and ships OTA.

## Acceptance criteria — the device QA that settles the open question

This needs an **iPhone SE 3 on iOS 26** (the device both #3330 and #3776 were reported from) running the filter sheet with logging added for:

- `Dimensions.get('window')`
- `useSafeAreaInsets()`
- the column height `useSheetColumnStyle` currently computes
- the probe's `onLayout` height for the sheet root

Two possible outcomes:

- **`onLayout` ≈ 548pt** → main's current formula is only ~7pt off, and #3776's 142pt gap is a rendering bug somewhere else in the tree, not the detent-height estimate. The probe approach would still fix the visible symptom by measuring around whatever's actually there, but the root cause would need separate investigation.
- **`onLayout` ≈ 407pt** → `@expo/ui`'s size report to Yoga is itself short of the real detent, i.e. the defect is upstream of both JS formulas, and this explains #3776 directly.

A second pass on an **iPhone 16 / iOS 26.5** covers #3568 (the other open "expands/floats" report, device unconfirmed).

## Open constant to resolve

The SE 3 measurement on #3776 (card top ~110pt on a 667pt-tall screen, no home indicator) implies a `.large`-detent "peek" of ~20pt above the safe area for home-button-class devices on iOS 26. `nathantannar4/Transmission`'s reimplementation uses `0` for the equivalent constant. Worth reconciling once real device numbers are in hand — may be an iOS-26-specific Liquid Glass chrome allowance that older reimplementations don't need to account for.

## Do not implement blind

No PR should land against this without the SE 3 device numbers above. The failure mode of a wrong guess here is a **clipped** Apply/Save button across every sheet that uses this hook (the filter sheet's Show-N-climbs button, and all three `%`-detent sheets in `AccessibilitySettingsScreen.tsx` — `453`: 95%, `549`, `605`), which is worse than the current cosmetic dead-gap.

References: #3611 (closed formula-fix attempt — wrong premise), #3776 (the report this targets), #3330 (original footer-floating report, same symptom class)

marcodejongh:
## Summary

This PR adds an on-screen diagnostic overlay so device measurements for issue #3922 can actually be captured — it does not fix the underlying sheet detent bug. Because `console.log` output is invisible on TestFlight/preview builds, a new "Sheet detent readout" toggle under More → Diagnostics shows live sheet measurements (formula, column, probe, padTop, inFlow) directly on screen, gated to dev/preview/OTA-preview builds only. It also refactors the diagnostics-eligibility check into a shared hook and fixes a static-import issue that was breaking test bundling.

---

Refs #3922. **This is not the fix** — it is the thing that makes the fix possible.

## Why not the fix

#3922 says, in bold, do not implement blind: no PR lands against it without an
iPhone SE 3 / iOS 26 readout, because a wrong column height clips the Apply/Save
button on every `%`-detent sheet (the filter sheet's Show-N-climbs button and all
three sheets in `AccessibilitySettingsScreen.tsx`), which is worse than the dead
gap it would be trying to close.

Three bug-batch passes have now claimed and released this issue at exactly that
gate. PR #4082 merged the instrumentation that produces the required numbers and
deliberately left the issue open. Ten days later no readout is attached, and the
reason is in #4082's own closing paragraph: `console.log` needs a dev client
attached to Metro. On a TestFlight or preview install it prints into the void.

#4082 offered to build the on-screen version if a dev-client run was
inconvenient. Apparently it was. So here it is.

## What this adds

- **More → Diagnostics → "Sheet detent readout"** — a toggle, default off, on the
  same eligibility gate as the existing "Bottom chrome diagnostics" row: dev
  builds, EAS preview builds, or a production install switched onto a `pr-<N>` OTA
  channel. A regular production user sees neither the row nor anything it controls.
- **A root overlay** (top-left, amber) showing the latest measurement per sheet:
  `formula`, `column`, `probe`, `padTop`, and the one that decides the issue,
  `inFlow`.
- `useSheetDetentProbe` now runs when that toggle is on, not only under `__DEV__`.

The measurement path itself is untouched. The overlay renders what the probe
already computed, including the padding-box correction (`probe − sentinelY`) that
sank the first fix attempt — the panel shows 532, never the raw 548.

## Reading it

On an iPhone SE 3 / iOS 26 (window 667):

| `inFlow` | verdict |
| --- | --- |
| ≈ 548 | native size report is right; the formula is ~7pt off and #3776's 142pt gap lives elsewhere in the tree |
| ≈ 391 | the SwiftUI → Yoga `setViewSize` report is itself short of the detent; the fix is upstream in `RNHostView.swift`, not in any JS formula |

And if `column ≠ formula`, the computed height never reaches the view at all,
which makes both branches moot.

## How to capture it

1. More → Diagnostics → **Sheet detent readout** on.
2. Climbs tab → filter button → let the sheet settle → **close it**.
3. Screenshot the panel.

Step 2's "close it" is not optional politeness: a native iOS sheet presents over
the root view, so the panel is behind it while the sheet is up. Readings survive
dismissal for exactly this reason.

Second pass on an iPhone 16 / iOS 26.5 covers #3568. The three `%`-detent sheets
under Settings → Accessibility fill out the other rows.

## Layout safety

Nothing here can move a sheet. The probe is `position: absolute`, the sentinel is
zero-height, both are `pointerEvents="none"`, and neither feeds a measurement back
into a style. With the toggle off no probe view mounts at all, so a production
session renders the tree it renders today. JS-only, so it ships OTA.

One structural change worth flagging: the readout store keeps React as its only
import. Every sheet reaches it through `sheet-detent-probe.ts`, so pulling
`../settings` in there dragged react-native-mmkv's Flow entry into Sheet,
ModalSheet and ClimbFilterSheet's static graph and killed five suites at
Rolldown's collection scan (`SyntaxError: Unexpected token 'typeof'`). The
root-mounted overlay resolves the toggle and pushes the flag down instead — second
commit, caught by the test run rather than by CI.

The diagnostics-eligibility hook moved out of `BottomChromeDebugOverlay.tsx` into
`src/hooks/use-diagnostics-eligible.ts` so both overlays share one gate. Behaviour
of the bottom-chrome row and overlay is unchanged.

## Validation

- `vp run typecheck:mobile`, `vp check` (clean on every file this branch touches)
- `vp run test:mobile` — sheet suites green, including the five the first commit
  broke
- New coverage: production-with-toggle-off mounts nothing; partial readings render
  as `—` rather than `0`; the padding correction is asserted, not assumed; turning
  the readout off drops stale readings

Device QA is what this PR is asking *for*, not claiming to have done. Notes in
`.boardsesh/qa-notes.md`.

## Release Notes

- [x] No release note needed

## Follow-up

Once the SE 3 numbers land on #3922, the column-height fix is a separate PR and
its shape depends on which branch of the table above they fall in. Whichever it
is, it should not claim to close #3776's 142pt gap without device confirmation —
a perfect measurement moves the column ~9pt.

https://claude.ai/code/session_011KFR5nP1XcEH64ekt2nK4r

### DIFF
--- a/packages/mobile/app/(tabs)/profile/more.tsx
+++ b/packages/mobile/app/(tabs)/profile/more.tsx
@@ -31,7 +31,7 @@ import { drainMutationQueue } from '../../../src/offline/offline-sync-adapter';
 import { getHttpClient } from '../../../src/lib/graphql/client';
 import { hapticLight, hapticSelection } from '../../../src/lib/haptics';
 import { DevMetadataPanel } from '../../../src/components/DevMetadataPanel';
-import { useBottomChromeDiagnosticsEligible } from '../../../src/components/BottomChromeDebugOverlay';
+import { useDiagnosticsEligible } from '../../../src/hooks/use-diagnostics-eligible';
 import { MoreForm } from '../../../src/components/MoreForm';
 import type { MoreButtonRow, MoreFormModel, MoreRow, MoreSection } from '../../../src/components/MoreForm.types';
 import { isPreviewBuild } from '../../../src/lib/preview-build';
@@ -78,8 +78,9 @@ export default function MoreScreen() {
   const { localePreference, setLocalePreference } = useLocalePreference();
   const { enabled: sessionRecordingEnabled, setEnabled: setSessionRecordingPreference } =
     useSessionRecordingPreference();
-  const bottomChromeDiagnosticsEligible = useBottomChromeDiagnosticsEligible();
+  const diagnosticsEligible = useDiagnosticsEligible();
   const [bottomChromeDiagnostics, setBottomChromeDiagnostics] = useSetting('bottomChromeDiagnostics');
+  const [sheetDetentDiagnostics, setSheetDetentDiagnostics] = useSetting('sheetDetentDiagnostics');
   const { enabled: showPlaylistTags, setEnabled: setShowPlaylistTags } = useShowPlaylistTagsPreference();
   const { enabled: showBoardseshGrades, setEnabled: setShowBoardseshGrades } = useBoardseshGradesPreference();
   const { enabled: showQuickActionsButton, setEnabled: setShowQuickActionsButton } = useClimbQuickActionsButton();
@@ -646,9 +647,9 @@ export default function MoreScreen() {
           setSessionRecordingEnabled(next);
         },
       },
-      // Bottom-chrome geometry overlay — dev / preview builds / pr-channel OTA
-      // testers only, so regular production users never see the row.
-      ...(bottomChromeDiagnosticsEligible
+      // Geometry overlays — dev / preview builds / pr-channel OTA testers only,
+      // so regular production users never see these rows.
+      ...(diagnosticsEligible
         ? [
             {
               kind: 'toggle' as const,
@@ -661,6 +662,19 @@ export default function MoreScreen() {
               onValueChange: (next: boolean) => {
                 hapticSelection();
                 setBottomChromeDiagnostics(next);
+              },
+            },
+            {
+              kind: 'toggle' as const,
+              key: 'sheetDetentDiagnostics',
+              // i18n-ignore-next-line — tester-only diagnostics
+              label: 'Sheet detent readout',
+              // i18n-ignore-next-line
+              subtitle: 'Overlay measured sheet heights (#3922)',
+              value: sheetDetentDiagnostics,
+              onValueChange: (next: boolean) => {
+                hapticSelection();
+                setSheetDetentDiagnostics(next);
               },
             },
           ]
--- a/packages/mobile/app/_layout.tsx
+++ b/packages/mobile/app/_layout.tsx
@@ -83,6 +83,7 @@ import { OnboardingGate } from '../src/components/onboarding/OnboardingGate';
 import { AccessoryOnboardingTip } from '../src/components/onboarding/AccessoryOnboardingTip';
 import { FreezeDebugOverlay } from '../src/components/FreezeDebugOverlay';
 import { BottomChromeDebugOverlay } from '../src/components/BottomChromeDebugOverlay';
+import { SheetDetentReadoutOverlay } from '../src/components/SheetDetentReadoutOverlay';
 import { WindowInsetPublisher } from '../src/hooks/use-window-bottom-inset';
 import { LiveActivityIntentDiagnostics } from '../src/components/LiveActivityIntentDiagnostics';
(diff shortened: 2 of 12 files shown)


marcodejongh:

marcodejongh:

Conflicting: packages/mobile/app/(tabs)/profile/more.tsx, BottomChromeDebugOverlay.tsx, settings/defaults.ts, settings/types.ts.

PR #4301, opened 2026-08-11
fix(mobile): [native-train] avoid main-thread UIImage(named:) miss in expo-image

### LINKED ISSUE #3928: expo-image: UIImage(named:) on the main thread drives _UIImageCollectImagePathsForPath on every ImageView.reload

marcodejongh:
## Native release train

This PR patches `expo-image`'s iOS native source (`ios/ImageView.swift`) and adds a `patchedDependencies` entry, which changes the Expo native fingerprint. It targets `release/next-ota`, stacked after the merged #4120 repair. A new store build has to ship and be adopted before an OTA from this branch can reach users.

Fixes #3928. Refs #3803 and #3944.

## The bug

`ImageView.reload(force:)` asks `localAssetName(from:)` whether a source can be loaded synchronously through `UIImage(named:)` before falling through to SDWebImage.

ExpoModulesCore represents scheme-less asset names with relative `file:` URLs, so upstream intentionally accepts the `file` scheme. That also admitted genuine filesystem URLs used by board photos and generated holds overlays. Every reload therefore tried `UIImage(named: <filesystem path>)`, a guaranteed miss that can synchronously enumerate bundle paths on the main thread before SDWebImage loads the same file correctly.

## The fix

`patches/expo-image@57.0.1.patch` now classifies filesystem URLs before removing a leading slash:

```swift
let hasFileHost = url.scheme == "file" && !(url.host?.isEmpty ?? true)
let hasAbsoluteFilePath = url.scheme == "file" && path.hasPrefix("/")
if hasFileHost || hasAbsoluteFilePath {
  return nil
}
if path.hasPrefix("/") {
  path.removeFirst()
}
```

This preserves Expo's supported asset-name forms:

- relative `URL(fileURLWithPath: "app_icon")`
- nested xcasset names such as `URL(fileURLWithPath: "Images/MyIcon")`
- a genuinely scheme-less `/app_icon`, which normalizes to `app_icon`

It rejects every absolute `file:` URL before normalization, including `file:///app_icon`, `file:///private/...`, and `URL(fileURLWithPath: "/app_icon")`, plus hosted `file://host/...` URLs. Rejected sources still use SDWebImage's disk loader; the visual result is unchanged.

The installed patched Swift source hashes to `cb69b1b60b491d3da3df1202509e136041cfc739`, which is recorded in the patch header.

## Guard against the patch silently dropping

Bun keys `patchedDependencies` by exact package version. A future `expo-image` bump can otherwise install an unpatched copy while typecheck and Metro remain green.

The shipped patch rule now pins the exact host, absolute-path, and combined-guard expressions. Regression coverage also:

- verifies the guard runs before leading-slash normalization
- mutates that ordering and proves the check goes red
- rejects the former post-normalization guard
- models relative, nested, scheme-less, absolute root, absolute nested, hosted, remote, and empty URL shapes
- pins the patch key to the direct `expo-image` dependency version

## Validation

Run from the native-train repair worktree on Linux:

| step | result |
| --- | --- |
| `vp install` | pass; patch applies cleanly |
| `vp run check:mobile-patches` | pass — 7 native patch rules applied, 5 patch files accounted for |
| `vp test run --project scripts --reporter=agent scripts/__tests__/mobile-patches-check.test.ts` | pass — 53 tests on the cumulative #4420 → #4120 → #4301 head |
| `vp run test:mobile` | pass — 619 files, 5,458 tests |
| `vp run check:mobile-bundle` | pass — iOS and Android |
| `vp check scripts/mobile-patches-check.ts scripts/__tests__/mobile-patches-check.test.ts` | pass |
| pre-commit hook | pass, no `--no-verify` |
| independent QA | PASS with no P0–P2 findings; exact stacked head `c787b2649` is patch-identical to the reviewed repair |

Swift cannot be compiled on this Linux host. Native-train macOS CI and device/simulator QA should confirm the board-art surfaces still render unchanged.

## Release Notes

- iOS: Board photos and hold overlays load without an extra pause.

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -908,6 +908,7 @@
     "@expo/ui@57.0.8": "patches/@expo%2Fui@57.0.8.patch",
     "expo-dev-launcher@57.0.10": "patches/expo-dev-launcher@57.0.10.patch",
     "@expo/fingerprint@0.20.6": "patches/@expo%2Ffingerprint@0.20.6.patch",
+    "expo-image@57.0.1": "patches/expo-image@57.0.1.patch",
     "react-native-screens@4.26.2": "patches/react-native-screens@4.26.2.patch",
   },
   "overrides": {
--- a/package.json
+++ b/package.json
@@ -65,6 +65,7 @@
     "@expo/fingerprint@0.20.6": "patches/@expo%2Ffingerprint@0.20.6.patch",
     "@expo/ui@57.0.8": "patches/@expo%2Fui@57.0.8.patch",
     "expo-dev-launcher@57.0.10": "patches/expo-dev-launcher@57.0.10.patch",
-    "react-native-screens@4.26.2": "patches/react-native-screens@4.26.2.patch"
+    "react-native-screens@4.26.2": "patches/react-native-screens@4.26.2.patch",
+    "expo-image@57.0.1": "patches/expo-image@57.0.1.patch"
   }
 }
--- a/patches/expo-image@57.0.1.patch
+++ b/patches/expo-image@57.0.1.patch
@@ -1,37 +1,21 @@
 diff --git a/ios/ImageView.swift b/ios/ImageView.swift
-index cccedd7adf7328b18bb61333c3fe2473c79e72cd..986ac82e9b89e227aa34d516358092fafd4bf130 100644
+index cccedd7adf7328b18bb61333c3fe2473c79e72cd..cb69b1b60b491d3da3df1202509e136041cfc739 100644
 --- a/ios/ImageView.swift
 +++ b/ios/ImageView.swift
-@@ -871,5 +871,32 @@ func localAssetName(from url: URL?) -> String? {
-   if path.hasPrefix("/") {
-     path.removeFirst()
-   }
+@@ -868,4 +868,16 @@ func localAssetName(from url: URL?) -> String? {
+   // Use `relativePath` so we recover the original input ("my_image") instead of
+   // the absolute path it resolves to against the file:// base
+   var path = url.relativePath
 +
-+  // A genuine filesystem URL — `file:///private/var/.../board.png`, or a
-+  // `file://host/...` — can never resolve via `UIImage(named:)`: that API only
-+  // looks in asset catalogs and bundle resources, so handing it a real path is
-+  // a guaranteed miss that still pays for `_UIImageCollectImagePathsForPath`'s
-+  // synchronous main-thread bundle-directory enumeration on every `reload`
-+  // (boardsesh/boardsesh#3928). Reject those here so `reload(force:)` falls
-+  // through to SDWebImage's disk loader instead, which handles `file://` URLs
-+  // correctly off the main thread.
-+  //
-+  // ExpoModulesCore also wraps scheme-less JS asset names with
-+  // `URL(fileURLWithPath:)`. Relative names retain the current directory as
-+  // their base URL, including supported namespaced catalog names such as
-+  // "Images/MyIcon"; keep those eligible for `UIImage(named:)`. Explicit file
-+  // URIs and absolute file paths have no base URL, so a remaining `/` after
-+  // stripping the leading slash identifies the real multi-component paths we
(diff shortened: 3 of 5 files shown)


marcodejongh:
Independent QA PASS on the repaired expo-image patch, with no P0–P2 findings. The guard rejects hosted and every absolute `file:` URL before stripping a leading slash, while preserving relative and nested asset-catalog names such as `Images/MyIcon` plus scheme-less `/app_icon`. Source-order and mutation sentinels cover root, nested, hosted, remote, and empty cases; the installed patched Swift blob matches the patch header. Final stacked head `c787b2649` is patch-identical to the reviewed repair; the combined scripts suite passes 53/53 and all seven native patches apply.

PR #4302, opened 2026-08-11
fix(db): select pending dev-db migrations per hash and arm the journal check

### LINKED ISSUE #3979: scripts/dev-db-up.sh has the same high-water-mark migration gap as migrate.ts had
## Summary

`run_pending_drizzle_sql_migrations()` in `scripts/dev-db-up.sh` reimplements drizzle's applier — and reimplements its bug:

```sh
last_migration_created_at=$(... "SELECT COALESCE((SELECT created_at FROM drizzle.\"__drizzle_migrations\" ORDER BY created_at DESC LIMIT 1), 0);")
```

```js
for (const entry of journal.entries) {
  if (entry.when <= lastAppliedAt) continue;
  ...
}
```

That is a single high-water mark, snapshotted once before the loop. A migration whose `when` lands at or below the newest applied `created_at` is skipped on that run and on **every** run after it, because the mark only moves up. There is no per-migration reconciliation.

This is the same defect #2933 reported in `packages/db/scripts/migrate.ts`, in a second place. #3977 fixed the production path (per-entry, hash-keyed check that fails and names the missing tags) but deliberately left this one alone.

## Blast radius

Local only — this script builds a developer's dev database, never production. That is the whole reason it was cut from #3977's scope rather than bundled into a change that gates the production deploy.

## Fix

Reuse the same reconciliation `migrate.ts` now uses. `scripts/lib/migration-ledger.ts`'s `findUnappliedMigrations()` already does the per-entry hash diff and is pure TS with no dependencies; the bun `--eval` block here could select on "hash not present in the ledger" instead of "`when` above the mark".

Worth doing together with the dev-db image rebuild (#3978), since the two interact: the image's missing `0187_sad_freak` ledger row is exactly what this applier retries and chokes on.

Found while implementing #2933.

marcodejongh:
## Summary

This PR fixes a migration-tracking bug in the dev-db scripts where both `dev-db-up.sh` and `dev-db-discover.ts` decided which migrations to apply using a single "last applied" timestamp snapshot, so a migration landing below that watermark (from a rebase renumbering it, branches collapsing theirs into one, or a pre-built image being older than the branch) got silently skipped forever. It replaces that logic with the same hash-keyed `findUnappliedMigrations()` selector the production migration gate already uses, so the dev and deploy paths can no longer disagree about what counts as applied. It also flips the journal-vs-ledger verification check from opt-in (`VERIFY_MIGRATION_JOURNAL=1`) to on by default, since the dev-db image now records every migration reliably, closing the gap where the check was off exactly where the drift gets introduced. One case is deliberately left unfixed: a database that ran a superseded branch version of a collapsed migration's DDL without its ledger row still needs a manual volume reset — it now fails loudly with a named migration and a fix, instead of silently succeeding.

---

Closes #3979
Refs #3978

> **Stacked on #4274** (`fix/4211-dev-db-ledger-timestamps`), which this extends rather than duplicates — the ledger normaliser it added runs immediately before the selection rewritten here. Base is that branch, so the diff stays at 12 files; **retarget to `main` once #4274 merges.**

## The bug (#3979)

`scripts/dev-db-up.sh` and `scripts/dev-db-discover.ts` each reimplemented drizzle's applier, and each reimplemented its bug:

```sh
last_migration_created_at=$(psql -c "SELECT ... ORDER BY created_at DESC LIMIT 1")
# then, in the bun --eval:
if (entry.when <= lastAppliedAt) continue;
```

One `max(created_at)` snapshot, taken before the loop, then `when > mark`. The mark only ever moves up, so a migration landing at or below it is skipped on that run **and on every run after it**. Same defect #2933 found in `packages/db/scripts/migrate.ts` and #3977 fixed there; the dev scripts were deliberately left out of that scope.

A dev database lands below the mark for entirely ordinary reasons: a rebase renumbered the migration, two branches collapsed theirs into one, or the pre-built image was built after the branch's `when` was minted. The symptom is a checkout that prints `No pending migrations.` and a schema missing the table the branch just added.

## The fix

Both appliers now ask the per-entry, hash-keyed question through the same `findUnappliedMigrations()` the production gate uses, so the dev and deploy paths cannot drift into disagreeing about what "applied" means.

- `scripts/lib/dev-db-pending-migrations.ts` — pure selection + journal hashing.
- `scripts/dev-db-pending-migrations.ts` — thin CLI; `dev-db-up.sh` pipes `SELECT hash FROM drizzle."__drizzle_migrations"` into it and reads `tag|when|hash` back. Replaces the inline `bun --eval`.
- `scripts/dev-db-discover.ts` (tailnet path) uses the same helper directly.

Ledger rows still record `created_at` as the journal's own `when`, so a row this applier writes is indistinguishable from one drizzle wrote and #4274's normaliser has nothing to repair on the next run.

**Hash parity:** root `scripts/` cannot import drizzle's `readMigrationFiles` (drizzle-orm is a `packages/db` dep, and Bun's isolated linker gives the root no hoisted copy), so the sha256 is derived from the raw `.sql` bytes — which is what drizzle 0.45 does, what `Dockerfile.dev-db` writes with `sha256sum`, and what the `--eval` this replaces already computed. `packages/db`'s journal-verification integration test runs against drizzle's real hashes and is where a future drizzle bump would surface.

## Arming the journal check (#3978)

The journal-vs-ledger check shipped behind `VERIFY_MIGRATION_JOURNAL=1` for one stated reason: the dev-db image was missing `0187_sad_freak`'s ledger row, and a default-on check would have reddened every developer's `vp run db:migrate` for a defect in the image. That reason is gone:

- `ghcr.io/boardsesh/boardsesh-dev-db:latest` records **190 of 190** journal entries (verified below).
- The image build fails if its ledger's high-water mark is not the journal's newest `when` (#4211).
- `vp run db:up` now fills a stale volume's gaps hash-by-hash *before* `db:migrate` runs.

Opt-in also meant the check was off in the one place a gap is created — the branch author's machine, on the branch that creates it. It now runs unless `VERIFY_MIGRATION_JOURNAL=0`. `ci.yml`, `production-deploy.yml` and `db-migration-renumber.yml` keep passing the literal `1` and stay armed without an edit.

The env decision moved into `scripts/lib/migration-ledger.ts` so it has unit coverage in the `scripts` Vitest project — `packages/db` runs on `tsx --test` and is outside the CI test graph.

## What #3978 does *not* fix, and why

No automatic repair for a `db_data` volume that carries a **superseded branch version** of a collapsed migration's objects without its ledger row. The four branch-local "0187"s each created a different subset of what `0187_sad_freak` creates, so no probe can tell what is already there, and inserting the ledger row without completing the DDL would mark the migration applied forever with part of it missing.

That shape used to hide below the high-water mark. It now fails with the migration named and the remedy spelled out:

```
ERROR: could not apply 0187_sad_freak to the dev database.
       If psql reported "already exists", this database carries that migration's
       objects without its ledger row — a volume that ran a superseded version of
       the migration on another branch, before it was renumbered or collapsed
       (#3978). ...
       Reset the volume and pull a current image:
         docker compose down -v && vp run db:up
```

Loud beats hidden, and the pre-built image ships the board data, the test user, and the seed data — a reset costs a pull, not a re-import. Leaving #3978 open (`Refs`, not `Closes`) so you can decide whether that residual case warrants a one-shot repair script.

## Verified locally

Against the running `boardsesh-dev-db:latest` container:

| check | result |
| --- | --- |
| `vp run db:verify-journal` | `✅ All 190 unbaselined journal migrations are recorded (193 ledger rows)` |
| `vp run db:up` | `No pending migrations.` |
| `vp run db:up` again | identical — nothing applied, nothing repaired |
| `vp run db:migrate` (nothing set) | gate armed, `✅ Verified all 190 unbaselined journal migrations are recorded` |
| selector, ledger with 0187's hash withheld | `0187_sad_freak\|1785047020870\|439223b0…` — named, despite its `when` sitting 594s **below** the ledger's mark of 1785641394370. This is the exact case the old filter hid. |
| selector, empty ledger | 190 entries selected, not 0 |
| `vp run test:db:migration-journal` (real Postgres) | 16/16 |
| `vp run check:db-migrations` | `190 migrations, journal consistent.` |
| `vp check`, `vp run typecheck`, `vp run typecheck:db` | clean (the 30 pre-existing repo-wide lint errors are untouched — mobile `@gorhom`, Sentry `ProcessEnv`) |
| `vp test run --project scripts` | 749 pass; the one failure, `prepare-rn-ios-tests`, fails identically on a stashed clean tree |

## For @marcodejongh before merge

The default flip reddens `vp run db:migrate` for any dev database carrying a gap this doesn't cover. I only ran read-only checks against this machine's local container, per instruction — **nothing was pointed at the tailnet-shared or production database**. Worth ticking off:

- [ ] `DB_URL=… vp run db:verify-journal` against the tailnet-shared dev DB (read-only; it is one `SELECT hash`).
- [ ] Same against any long-lived personal `db_data` volume.
- [ ] If either shows a gap, either repair it per `docs/db-migrations.md` or hold this PR — `VERIFY_MIGRATION_JOURNAL=0` is the unblock, but shipping with a known red is worse than waiting.

Production is unaffected: it already runs the gate with the literal `1`, and its baseline is untouched.

## Release Notes

- [x] No release note needed — dev tooling only, no user-facing surface.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -734,10 +734,10 @@ jobs:
         with:
           name: built-packages
       # The pinned image's ledger was stamped with the image build clock, not
-      # each journal entry's `when` (#4211). That is a high-water mark drizzle
-      # can never clear, so without this repair the next step silently skips
-      # every migration written after the image was built — including the one a
-      # PR is adding. Must stay ahead of db:migrate.
+      # each journal entry's `when` (#4211). That mark only moves up, so without
+      # this repair the next step silently skips any journal entry whose `when`
+      # predates the pinned image's build — including a PR's own migration when
+      # the branch is older than the pin. Must stay ahead of db:migrate.
       - name: Normalize the dev-db image migration ledger
         env:
           DATABASE_URL: postgres://postgres:password@localhost:5432/main
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -733,6 +733,15 @@ jobs:
         uses: actions/download-artifact@v4
         with:
           name: built-packages
+      # The pinned image's ledger was stamped with the image build clock, not
+      # each journal entry's `when` (#4211). That is a high-water mark drizzle
+      # can never clear, so without this repair the next step silently skips
+      # every migration written after the image was built — including the one a
+      # PR is adding. Must stay ahead of db:migrate.
+      - name: Normalize the dev-db image migration ledger
+        env:
+          DATABASE_URL: postgres://postgres:password@localhost:5432/main
+        run: bun run --filter=@boardsesh/db db:normalize-ledger
       - name: Apply the full migration journal
         env:
           DATABASE_URL: postgres://postgres:password@localhost:5432/main
--- a/.github/workflows/e2e-tests.yml
+++ b/.github/workflows/e2e-tests.yml
@@ -323,9 +323,11 @@ jobs:
 
       # The dev-db image's ledger was stamped with its build clock, not each
       # journal entry's `when` (#4211). drizzle-kit migrate reads max(created_at)
-      # once and applies only entries above it, so without this repair a PR's new
-      # migration silently never runs here and surfaces as an unrelated-looking
-      # app failure. Must stay ahead of the migrate step.
+      # once and applies only entries above it, so a PR's migration is skipped
+      # whenever it was generated before the image was built — and this job runs
+      # `:latest`, rebuilt on every db-touching merge to main, so that is most
+      # open PRs. It surfaces as an unrelated-looking app failure rather than a
+      # migration error. Must stay ahead of the migrate step.
       - name: Normalize the dev-db image migration ledger
         run: bun run --filter=@boardsesh/db db:normalize-ledger
         env:
@@ -582,8 +584,9 @@ jobs:
           rm built-packages.tar
 
       # Same repair as the shard job above: the image's ledger high-water mark is
-      # a build timestamp, so drizzle-kit migrate would skip a PR's new migration
-      # (#4211). Must stay ahead of the migrate step.
+      # its build timestamp, so drizzle-kit migrate would skip a PR's migration
+      # whenever the branch predates the `:latest` image (#4211). Must stay ahead
+      # of the migrate step.
       - name: Normalize the dev-db image migration ledger
         run: bun run --filter=@boardsesh/db db:normalize-ledger
         env:
--- a/.github/workflows/e2e-tests.yml
+++ b/.github/workflows/e2e-tests.yml
@@ -321,6 +321,16 @@ jobs:
           tar -xf built-packages.tar
           rm built-packages.tar
 
+      # The dev-db image's ledger was stamped with its build clock, not each
+      # journal entry's `when` (#4211). drizzle-kit migrate reads max(created_at)
+      # once and applies only entries above it, so without this repair a PR's new
+      # migration silently never runs here and surfaces as an unrelated-looking
+      # app failure. Must stay ahead of the migrate step.
+      - name: Normalize the dev-db image migration ledger
+        run: bun run --filter=@boardsesh/db db:normalize-ledger
+        env:
+          DATABASE_URL: postgresql://postgres:password@postgres:5432/main
+
       - name: Run database migrations
         run: bunx drizzle-kit migrate
         working-directory: packages/db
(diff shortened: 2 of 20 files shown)


marcodejongh:

marcodejongh:

Conflicting: packages/db/docker/Dockerfile.dev-db (drizzle migration-timestamp logic) — pre-flagged DIRTY by GitHub before any rebase attempt. Touches nearly the same file set as blocked #4274 (same #3978/#3979/#4211 problem area) — likely worth reconciling together rather than resolving each separately.

PR #4303, opened 2026-08-11
fix(web): guard playlist rename edits against a newer concurrent change

### LINKED ISSUE #4267: Web playlist edits still overwrite a newer rename silently
Since #1934, the backend refuses a playlist metadata edit that would overwrite a newer one — but only when the client sends a `basedOn` snapshot. Mobile does. Web's `PlaylistEditDrawer` (`packages/web/app/components/library/playlist-edit-drawer.tsx`) does not, so a web edit still silently overwrites a rename made on a phone, and the web user never learns they clobbered it.

The server side is already live and backwards compatible: omit `basedOn` and you get the old last-write-wins, which is exactly what web does today. Nothing to build on the backend.

**What's needed**

1. Send the snapshot from the edit drawer: `basedOn: { updatedAt, name, description, isPublic, color, icon }` built from the playlist the drawer loaded. `PLAYLIST_FIELDS` already selects every one of those, so there's no new query field.
2. Handle the rejection. `readPlaylistUpdateConflict` in `@boardsesh/graphql/errors` reads it, and the error's extensions carry the server's current values — enough to name both versions and to rebuild `basedOn` for a deliberate overwrite without a refetch.
3. Show the choice. Mobile uses a three-way alert (Keep mine / Use theirs / Cancel) in `packages/mobile/app/(tabs)/discover/[playlist_uuid].tsx`; web would want the MUI equivalent. Note the wording branch mobile needed: the check also refuses on description, colour, icon and visibility, so both sides can carry the same name — quoting name-vs-name reads as nonsense in that case.
4. Patch both caches on "use theirs", not just the detail one.

**Worth knowing before starting**

- A climb added, removed or reordered bumps `playlists.updated_at` without touching the metadata, which is why the server compares fields rather than the bare timestamp. Don't reintroduce a timestamp-only check on the client.
- The guarantee is per-client until this lands: web↔web and web→mobile overwrites stay silent.

Context: PR #4252, and `docs/offline-sync-plan.md` § "The one mutation that can't be auto-merged".

marcodejongh:
## Summary

Closes #4267.

Since #1934 / PR #4252, the backend refuses `updatePlaylist` when the
caller's `basedOn` snapshot doesn't match the stored row, but only clients
that opt in by sending `basedOn` get that protection. Web's
`PlaylistEditDrawer` never did, so a web edit still silently overwrote a
rename made on another device or tab, with no indication to the user that
anything was clobbered.

This PR makes the drawer opt in and handle the rejection:

- `handleSubmit` now sends `basedOn` built from the playlist prop
  (`updatedAt`, `name`, `description`, `isPublic`, `color`, `icon`). A
  playlist without `updatedAt` (a stale pre-field cache) falls back to no
  `basedOn` — the old last-write-wins — same as mobile's fallback.
- On a `PLAYLIST_UPDATE_CONFLICT` rejection (read via
  `readPlaylistUpdateConflict` from `@boardsesh/graphql/errors`), a MUI
  Dialog offers **Cancel / Use theirs / Keep mine**, mirroring the mobile
  three-way alert added in #4252:
  - **Cancel** closes the dialog only; the drawer stays open with the edit
    intact.
  - **Use theirs** adopts the server's values with no extra network call and
    closes the drawer.
  - **Keep mine** retries the mutation rebased on the server's own values,
    so the overwrite is deliberate. If *that* retry conflicts again (a
    third device), it doesn't re-prompt — it shows
    `edit.conflict.changedAgain` via the snackbar and leaves the drawer
    open.
  - When the server's name matches what you typed (the conflict is only on
    description/colour/icon/visibility), the dialog uses the
    `messageDetails` wording instead of quoting the same name twice.
- Both a successful save and "Use theirs" invalidate the
  `['userPlaylists']` react-query cache (the only other web cache of the
  playlist list, in `use-climb-actions-data.tsx`), so the Add-to-Playlist
  picker picks up the change alongside the playlist detail page.

No backend or shared-package changes — this is a web-only consumer of the
substrate PR #4252 already shipped (conflict check, `PlaylistRevision`
input, typed error readers, and the shared `edit.conflict.*` i18n keys).
Zero new i18n keys were added; the shared catalogs (en-US/de/es/fr) already
cover them.

**⚠️ This PR is stacked on `fix/1934-playlist-rename-lww-conflict` (#4252),
which is still open.** The base branch here is `fix/1934-playlist-rename-lww-conflict`,
not `main` — please don't merge this until #4252 lands, and retarget this PR's
base to `main` once it does (the diff will shrink to just the two files
listed below).

## Known limitation

The guarantee is per-client: an old client that never sends `basedOn`
(a stale mobile build, a third-party caller) still overwrites silently.
Restated from the issue, not new to this PR.

## Files changed

- `packages/web/app/components/library/playlist-edit-drawer.tsx`
- `packages/web/app/components/library/__tests__/playlist-edit-drawer.test.tsx`

## Validation

- `vp check` — clean (0 new errors/warnings; repo baseline of 30
  pre-existing errors / 298 warnings, confirmed identical on the base
  branch before this change and unrelated to these two files).
- `vp run typecheck:web` — clean (`next build` + `tsc --noEmit` both pass).
- `vp run typecheck:shared` — clean.
- `vp test run --project web --reporter=agent packages/web/app/components/library/__tests__/playlist-edit-drawer.test.tsx`
  — 11/11 passing (4 pre-existing + 7 new: basedOn on submit, no basedOn
  when `updatedAt` is missing, conflict dialog wording for both the
  name-differs and same-name cases, Use theirs, Keep mine retry, a second
  conflict on retry, and cancel preserving form state).
- `vp run check:i18n` — clean (no hardcoded strings; only existing
  `edit.conflict.*` keys reused).

- None — the plan's judgement calls (dialog copy, wording branch,
  cache-invalidation scope) all reuse #4252's already-agreed patterns
  verbatim, so there's nothing new to decide here beyond re-reviewing
  #4252 itself.

### DIFF
--- a/docs/offline-sync-plan.md
+++ b/docs/offline-sync-plan.md
@@ -364,6 +364,14 @@ async function processMutation(mutation: PendingMutation) {
 
 Each mutation gets a client-generated UUID as an idempotency key. The backend's `saveTick` and `createPlaylist` accept this UUID and use `ON CONFLICT (uuid) DO NOTHING` for safe retry. Favorites use explicit `addFavorite`/`removeFavorite` (not `toggleFavorite`) so retries don't invert state. Follow/unfollow operations are naturally idempotent (follow when already following = no-op, unfollow when not following = no-op).
 
+### The one mutation that can't be auto-merged
+
+Every queued mutation above is idempotent or commutative, so replaying it is always safe. `updatePlaylist` is the exception: a queued rename replayed blind overwrites whatever the playlist is called now, including a rename made on another device while this one was offline (#1934).
+
+So it carries a `basedOn` snapshot — the `updatedAt` plus the metadata values the client last saw. The server compares that against the stored row inside a transaction holding a row lock and, when the two genuinely diverge, rejects with `PLAYLIST_UPDATE_CONFLICT` instead of writing. The error's extensions carry the server's current values, so the client can name both versions in a prompt and rebuild `basedOn` for a deliberate overwrite without a refetch. Omitting `basedOn` keeps the old last-write-wins behaviour, which is what shipped binaries and web still do.
+
+Two consequences for the queue. A rejection resolves to no HTTP status, so `isRetryable` returns false and the row dead-letters for a human to resolve instead of burning ten retries. And a climb added or reordered elsewhere bumps `playlists.updated_at` without touching the metadata, which is why the check compares fields rather than the bare timestamp — a timestamp alone would false-conflict on the same device's own queued climb add draining ahead of the rename. Any future mutation that can't be auto-merged should reuse this code rather than invent a shape.
+
 ### Dead letter handling
 
 Mutations that exceed `MAX_RETRY_COUNT` or fail with non-retryable errors are moved to `status = 'dead_letter'`. The app shows a badge/indicator when dead-letter mutations exist. Users can view failed mutations and choose to retry or discard them. This prevents silent data loss — the user always knows if a tick didn't sync.
--- a/packages/backend/src/__tests__/playlist-update-conflict.test.ts
+++ b/packages/backend/src/__tests__/playlist-update-conflict.test.ts
@@ -1,0 +1,347 @@
+import { describe, it, expect, beforeEach, afterAll } from 'vite-plus/test';
+import { eq, sql } from 'drizzle-orm';
+import { GraphQLError } from 'graphql';
+import { PLAYLIST_UPDATE_CONFLICT_CODE } from '@boardsesh/shared-schema';
+import type { ConnectionContext } from '@boardsesh/shared-schema';
+import { db } from '../db/client';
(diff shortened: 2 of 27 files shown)


marcodejongh:

marcodejongh:

Conflicting: packages/backend/src/graphql/resolvers/playlists/mutations.ts — pre-flagged DIRTY by GitHub before any rebase attempt. Same feature area as blocked #4252 (mobile playlist-rename LWW) — might be worth resolving together.

marcodejongh:
Closing as obsolete after #4467 removed the web PlaylistEditDrawer and its editor test. This PR’s web-only feature commit modifies those removed surfaces. The mobile/backend concurrency work in parent #4252 remains open and is not marked complete by this closure.

Issue #3996, opened 2026-07-30
Mobile Discover: board-aware feed, board labels on playlist cards, and no silent no-op on cross-board climbs


marcodejongh:
Split out of #3891 so that issue can close on its two concrete defects (the MoonBoard playlist size filter and the unreportable add failure). These three are design work on the RN Discover surface, carried over from the web #3825 analysis by @marcodejongh — see https://github.com/boardsesh/boardsesh/issues/3891#issuecomment-5077454431 for the full write-up.

The web repro: a boardless user browsing a Kilter wall opened Discover and got a feed dominated by Tension recommendation playlists. Opening them was a dead end. The public playlist pool skews roughly 2:1 Tension:Kilter and "popular" ranks by climb count, so 50-climb generated playlists lead and an unfiltered feed buries the user's own board.

### 1. Board-aware Discover rail

Scope the Discover feed to the board the climber is actually on — including the active session / route board, not just a board they have saved. Fall back to a multi-board feed only when there is genuinely no board context.

### 2. Board label on playlist cards

Show the board's brand name (Kilter / Tension / MoonBoard) on Discover and library cards, so a cross-board playlist is legible before it is opened rather than after tapping in.

### 3. No silent no-op on cross-board climbs

When a climber opens a playlist whose climbs the active board can't render, don't grey the climbs and swallow the tap — that reads as "nothing works." Say which board the climbs are for and how to switch. On web this was a bare `return;` with a "don't show a generic snackbar" comment; the RN equivalent should surface an explanatory hint.

Note that mobile already routes wrong-board playlist climbs to a view-only drawer (`usePlaylistActivation`'s `viewOnlyBoard` path), so item 3 on RN is about labelling and guidance rather than the silent-return web had.

PR #4304, opened 2026-08-11
fix(backend): reject cross-board playlist adds

### LINKED ISSUE #4015: Playlist picker lists all-board playlists but the membership query is board-scoped — cross-board adds lose their checkmark
Found during the playlist code review that produced #4010. Related: #3996 (cross-board climbs on mobile Discover).

The climb-action picker fetches the user's full library with `input: { pageSize: 200 }` and no `boardType`/`layoutId` filter (`packages/web/app/hooks/use-climb-actions-data.tsx:167-169`), so playlists typed to *other* boards are listed and addable. But the membership query `playlistsForClimbs` filters `playlists.boardType = input.boardType AND (layoutId = input.layoutId OR layoutId IS NULL)` (`packages/backend/src/graphql/resolvers/playlists/queries/playlist-detail.ts:163-167`).

Result: adding a Kilter climb to a Tension-typed playlist succeeds and optimistically ticks, then loses its checkmark on the next refetch — it looks like the add failed, and tapping again can't remove it. Same asymmetry on mobile (`InlinePlaylistPicker.tsx:128-132` vs the unfiltered library fetch).

Fix options: scope the picker's playlist list to the current board/layout, or make the membership query cover the listed playlists (drop the board filter when the picker is all-board). Either way the two sides must agree.

marcodejongh:
## Summary

Closes #4015.

`addClimbToPlaylist` only checked ownership — not whether the climb being added
belongs to the playlist's board and layout. A cross-board add (a Kilter climb
into a Tension playlist) inserted a `playlist_climbs` row and ticked the
checkmark optimistically, then the board-scoped membership refetch
(`playlistsForClimb` / `playlistsForClimbs`, which filter on
`board_type = X AND (layout_id = Y OR layout_id IS NULL)`) dropped it again.
The checkmark vanished, and tapping a second time couldn't remove the row —
the UI no longer believed the climb was in there.

#4268 fixed the client-visible half on mobile: `InlinePlaylistPicker` now only
offers playlists on the climb's own board, so an up-to-date app can't pick a
mismatched target. This PR is the server-side half — the same rule enforced in
`addClimbToPlaylist`, so a client that predates #4268 (OTA rollout takes days
to reach the fleet) or an offline queue replaying an add composed before the
update can't leave an unreachable row behind either.

## Release Notes

- Adding a climb to a playlist from another board now fails cleanly instead of
  leaving a ghost entry you can't tap off again.

## What changed

`packages/backend/src/graphql/resolvers/playlists/mutations.ts`:

- The existing ownership `SELECT` also returns the playlist's
  `boardType`/`layoutId` — no extra round trip.
- `resolveClimbBoardScopes(climbUuid)` looks up the climb's own
  `boardType`/`layoutId` in `board_climbs`, falling back to
  `board_climb_aliases` (canonical-uuid join) when the uuid isn't itself a
  canonical row — the alias-read pattern from
  `packages/db/src/queries/aliases.ts`. It returns *every* scope the uuid can
  resolve to: the alias PK is `(board_type, alias_uuid)`, so one alias uuid can
  carry a row under two boards, and an unordered `LIMIT 1` would have made the
  guard's verdict depend on Postgres row order.
- `climbFitsPlaylistBoard` holds the match rule, mirroring the membership query
  exactly: boards must match, and a playlist pinned to a layout only takes that
  layout (`layout_id IS NULL` — how Aurora- and Kilter-synced circuits arrive —
  takes any layout of its own board). The add is accepted when *any* resolved
  scope fits, so an ambiguous uuid is never rejected on a coin flip.
- Rejection is a `GraphQLError` with `extensions.code = 'BAD_USER_INPUT'`, so a
  client can tell "wrong playlist" apart from a server failure worth retrying,
  and the offline outbox keeps classifying it as non-retryable (dead-letter,
  not an endless replay). The mobile picker already reverts the optimistic
  checkmark and shows its own translated "couldn't add" line on any rejection,
  so nothing user-facing depends on the wire text.
- The guard runs *before* the insert-or-noop transaction, so it doesn't widen
  the transaction window the race-handling comments in this file bound.
- **Fails open** (allows the add, logs a `logger.warn`) when the uuid resolves
  to no `board_climbs` row even after alias resolution — see below.

Not touched: `playlistsForClimb` / `playlistsForClimbs` (already correct, and
now agree with both the mobile picker and this guard), the mobile picker
itself (#4268), and cross-board `playlist_climbs` rows already in the database
(a cleanup pass is data deletion and needs sign-off).

## Notes

- **Fail-open on an unknown climb uuid.** A uuid with no `board_climbs` row and
  no alias is allowed through with a warning: the catalog row can legitimately
  lag behind create-climb and offline-sync writes, and rejecting would break
  those flows for the sake of a defense-in-depth guard. The add still has to
  pass playlist ownership and the table's own constraints. Happy to flip this
  to fail-closed if you'd rather be strict.
- Mobile Discover can still show climbs from a board other than the active
  board config (#3996). The picker there is scoped to the *active* board, so an
  add for such a climb now surfaces a "couldn't add" error where it used to
  produce the ghost row. That is the honest outcome of the same bug; the real
  fix is #3996's board-aware feed.

## Test plan

`packages/backend/src/__tests__/add-climb-to-playlist-board-guard.test.ts`
(real Postgres, modeled on `add-climb-to-playlist-race.test.ts` /
`playlist-role-matrix.test.ts`):

- same-board, same-layout add succeeds
- cross-board add rejected, with `BAD_USER_INPUT` on the error
- same-board layout mismatch rejected when the playlist pins a `layoutId`
- `layoutId = NULL` playlist accepts any layout of its own board
- `layoutId = NULL` playlist still rejects a different board
- unknown climb uuid fails open — the add still succeeds
- alias uuid resolving to a canonical climb on the right board accepted
- alias uuid resolving to a canonical climb on the wrong board rejected
- alias uuid resolving under *two* boards accepted for the board that fits, and
  rejected when neither fits (the row-order case above)

Regression cover that must stay green: `add-climb-to-playlist-race.test.ts`
(the guard sits outside the `onConflictDoNothing` retry transaction) and
`playlist-role-matrix.test.ts` (ownership-check ordering unchanged).

Local validation: `vp run typecheck:backend` and `vp check` clean. Backend
Vitest is left to CI — this machine runs several worktrees against the same
Postgres worker DBs, so local backend runs report noise rather than signal.

### DIFF
--- a/packages/backend/src/__tests__/add-climb-to-playlist-board-guard.test.ts
+++ b/packages/backend/src/__tests__/add-climb-to-playlist-board-guard.test.ts
@@ -1,0 +1,237 @@
+import { afterAll, beforeAll, describe, expect, it } from 'vite-plus/test';
+import { sql } from 'drizzle-orm';
+import type { ConnectionContext } from '@boardsesh/shared-schema';
+import { db } from '../db/client';
+import { playlistMutations } from '../graphql/resolvers/playlists/mutations';
+
+// #4015: addClimbToPlaylist only checked ownership, not board/layout
+// compatibility, so a cross-board add would insert a playlist_climbs row
+// successfully — then the board-scoped membership refetch
+// (playlistsForClimb / playlistsForClimbs, which filter on
+// board_type = X AND (layout_id = Y OR layout_id IS NULL)) would silently
+// exclude it, making the UI checkmark vanish. #4268 fixed the mobile picker
+// to stop offering mismatched playlists as targets; this guard is the
+// server-side backstop that mirrors the membership query's exact rule.
+//
+// FIXTURE_RUN_ID keeps rows unique across parallel test workers sharing the
+// worker DB (see the "Backend tests: shared worker DBs" note) — no TRUNCATE
+// of shared tables, only scoped INSERT/DELETE by uuid prefix.
+
+// Short: climbUuid goes through ExternalUUIDSchema (max 50 chars), so keep
+// every generated climb uuid well under that ceiling.
+const FIXTURE_RUN_ID = crypto.randomUUID().slice(0, 8);
+const OWNER_ID = `bg4015-owner-${FIXTURE_RUN_ID}`;
+
+const KILTER_LAYOUT_8_PLAYLIST = `bg4015-kilter-l8-${FIXTURE_RUN_ID}`;
+const KILTER_NULL_LAYOUT_PLAYLIST = `bg4015-kilter-null-${FIXTURE_RUN_ID}`;
+
+const KILTER_L8_CLIMB = `bg4015-ck8-${FIXTURE_RUN_ID}`;
+const KILTER_L9_CLIMB = `bg4015-ck9-${FIXTURE_RUN_ID}`;
+const TENSION_L8_CLIMB = `bg4015-ct8-${FIXTURE_RUN_ID}`;
+const UNKNOWN_CLIMB = `bg4015-cunk-${FIXTURE_RUN_ID}`;
+
+const ALIAS_UUID = `bg4015-calias-${FIXTURE_RUN_ID}`;
+const ALIAS_CANONICAL_KILTER_L8 = KILTER_L8_CLIMB;
+
+function makeCtx(overrides: Partial<ConnectionContext> = {}): ConnectionContext {
+  return {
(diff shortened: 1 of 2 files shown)


marcodejongh:

PR #4305, opened 2026-08-11
fix: add range check on board_climb_stats quality averages

### LINKED ISSUE #3551: Add a DB-level range guard on board_climb_stats.quality_average
## Summary

boardsesh_ticks.quality gets a CHECK (1-5) in the star-scale repair PR (#3547), but board_climb_stats.quality_average has no DB-level guard — in-code range handling (normalizeQualityTo5, correctGripsQualityAverage, foldCatalogStat guards) is the only protection. A future writer regression could silently store out-of-range averages again.

## Suggested fix

After #3547's backfills land (they null the 0-sentinels and rescale legacy values), add CHECK (quality_average IS NULL OR (quality_average > 0 AND quality_average <= 5)) via drizzle schema + generate. Validate against prod first: post-backfill there should be zero violating rows.

Raised by review on #3547 (low severity, deliberate follow-up to keep that PR scoped). Found by the 2026-07 upstream-sync data audit program.

marcodejongh:
## Summary

`board_climb_stats.quality_average` has no DB-level range guard — only in-code handling (`blendedQualityAverageSql`, recompute filters, upstream-writer normalization) keeps it on the canonical 1-5 scale. Its sibling column `board_climb_ratings.rating` got a CHECK constraint in the wake of #3547 (migration 0112); this table was deliberately deferred to this issue.

Since #3551 was filed, migration 0167-0169 split the raw manufacturer value out into `upstream_quality_average` — that's now the actual ingestion point for upstream writers (Aurora sync, Kilter Grips catalog sync, MoonBoard import), so it gets the same guard.

## Important: the bound is `>= 0`, not the issue's suggested `> 0` — real data forced this

The issue assumed the prerequisite backfills (0149, 0151, 0169) left zero violating rows, so a plain `> 0` CHECK could ship safely. **That assumption doesn't hold today.** I validated this the hard way: `bunx drizzle-kit generate` produced the `> 0` version first, and CI's own `test-location-sync-integration` job — which replays the full migration journal from scratch against `ghcr.io/boardsesh/boardsesh-dev-db` — failed:

```
PostgresError: check constraint "board_climb_stats_quality_average_range" of relation "board_climb_stats" is violated by some row
```

I reproduced this locally against the exact pinned image digest CI uses and found the cause precisely:

- **22,232 real rows**, every one `board_type = 'moonboard'`, sitting at exactly `quality_average = upstream_quality_average = 0`, with `boardsesh_quality_sum`/`boardsesh_quality_count` both NULL (no Boardsesh votes at all).
- Migration 0151 already nulled this exact "unrated" sentinel once — its own comment records "quality_average = 0 → NULL, all boards... Prod: 22,498 rows (moonboard 19,883...)". The count I found today is suspiciously close but not identical, meaning either the MoonBoard catalog importer is a **still-live writer** that keeps re-introducing `0` instead of `NULL` for unrated climbs, or the 0167 `upstream_quality_average` split inherited the un-cleaned sentinel and it round-tripped back into `quality_average` on a later blend recompute.
- **Root cause confirmed on re-review** — it is the still-live writer, not a 0167 round-trip. `packages/db/scripts/import-moonboard-problems.ts` passes the dump's `userRating` straight through at L246 (`qualityAverage`) and L250 (`upstreamQualityAverage`), and L69 types it `userRating: number` (non-nullable). MoonBoard uses `0` for "nobody has rated this yet", so every import run writes the sentinel back into both columns. Every other ingestion path guards: Aurora/Tension via `normalizeQualityTo5` (`packages/shared-schema/src/utils.ts:148`, null for `q <= 0`), Kilter Grips via `correctGripsQualityAverage` (`packages/kilter-sync/src/sync/quality-scale.ts:31`, null outside 1-5). MoonBoard is the only unguarded one.
- The importer fix is deliberately not in this PR: on its own it does not let the bound tighten, and the backfill it needs is a prod data mutation that wants sign-off.

Given a `> 0` CHECK provably fails the deploy-time migration against real data, I shipped `>= 0` instead: it still catches the two impossible-by-construction cases (negative, `> 5`) that a broken blend or a botched upstream write could produce, while not fighting a live, uninvestigated sentinel bug. I verified the relaxed bound migrates cleanly against the same pinned CI image (see Validation below).

**Follow-up tracked in #4617** — so the relaxed bound does not become permanent by default. Three steps, and the order matters:

1. Fix `import-moonboard-problems.ts` to map `userRating` `0` → `NULL`.
2. Re-run a 0151-style cleanup for the rows already written back — note 0151 predates the 0166/0167 split, so `upstream_quality_average` has never been cleaned.
3. Tighten both CHECKs here from `>= 0` to `> 0`, which is what #3551 originally asked for.

Tightening before the backfill fails the migration; backfilling before the importer fix just gets undone on the next import run. The schema comment and the test both carry a pointer to #4617 so whoever picks it up lands in the right place.

## Changes

- `packages/db/src/schema/boards/unified.ts` — add two `check()` constraints on `boardClimbStats`, mirroring `board_climb_ratings_rating_range`:
  - `board_climb_stats_quality_average_range`: `quality_average IS NULL OR (quality_average >= 0 AND quality_average <= 5)`
  - `board_climb_stats_upstream_quality_average_range`: same shape for `upstream_quality_average`
  - See "Important" above for why the lower bound is `>= 0` rather than the issue's suggested `> 0`.
- `packages/db/drizzle/0202_black_rocket_racer.sql` — generated via `bunx drizzle-kit generate`, contains exactly the two `ALTER TABLE ... ADD CONSTRAINT ... CHECK` statements (verified by inspection, nothing else emitted). Renumbered 0190 → 0202 by the collision bot as `main` moved; the SQL is untouched.
- `packages/backend/src/__tests__/schema-sql.ts` — mirrored the same two CHECK constraints into the worker-DB test DDL so Vitest worker databases match prod schema. They are also re-added idempotently in a `DO $$ ... EXCEPTION WHEN duplicate_object` block after the table body: worker DBs are created once and reused (`worker-db.ts:104-108`), so a constraint added inside `CREATE TABLE IF NOT EXISTS` never lands on a pre-existing one, and the new test would fail on any dev box that already had a worker DB. Same remedy the file already documents for the `gyms` columns.
- `packages/backend/src/__tests__/climb-stats-quality-range-check.test.ts` — new regression test against a real Postgres worker DB: inserts of `-1`, `-0.01`, `5.01` are rejected with the constraint name in the error (checked via the driver's structured `code`/`constraint` fields, not a message regex — see `postgres-errors.ts`); `NULL`, `0`, `0.5`, `1`, `5` are accepted — for both columns.

## Before merging: read-only prod pre-check

Please run against prod (read-only connection, `default_transaction_read_only=on`) before merging, to see whether prod has the same moonboard sentinel found in CI's seed data:

```sql
SELECT board_type, count(*) AS zero_quality_rows
FROM board_climb_stats
WHERE quality_average = 0 OR upstream_quality_average = 0
GROUP BY board_type;

SELECT count(*) FILTER (WHERE quality_average IS NOT NULL AND (quality_average < 0 OR quality_average > 5)) AS bad_quality_average,
       count(*) FILTER (WHERE upstream_quality_average IS NOT NULL AND (upstream_quality_average < 0 OR upstream_quality_average > 5)) AS bad_upstream_quality_average
FROM board_climb_stats;
```

The second query (genuinely-impossible values) should be `0, 0` — if not, this migration will fail on deploy exactly like it did against CI's seed data, and needs the same `NOT VALID` / backfill treatment described above before it can ship even with the relaxed bound. The first query is informational: however many moonboard (or other) rows come back is the size of the follow-up sentinel-cleanup issue.

## Operational notes for on-call

- `ALTER TABLE ... ADD CONSTRAINT ... CHECK` takes an `ACCESS EXCLUSIVE` lock and full-scans `board_climb_stats` (a large, continuously-written table). Migration 0112 did the identical thing to `board_climb_ratings`, but treat that as a weak precedent: `board_climb_ratings` holds only Boardsesh's own per-user ratings, while `board_climb_stats` is every (climb, angle) across three boards.
- **Open decision — is the lock window acceptable?** Review asked for `ADD ... NOT VALID` plus a later `VALIDATE CONSTRAINT`. That is not cleanly reachable here: Drizzle's `check()` cannot emit `NOT VALID`, hand-written migration SQL is against CLAUDE.md, and both statements in one migration buy nothing if they share a transaction — the `ACCESS EXCLUSIVE` from the ADD is held to commit either way. It would have to be two migrations across two deploys. Worth it only if the table is big enough to matter: if `SELECT count(*) FROM board_climb_stats` is in the low millions the scan is seconds and the plain ADD is fine. Say the word and it gets split.
- Once live, any writer producing an out-of-range value will hard-fail that statement (constraint name `board_climb_stats_quality_average_range` / `board_climb_stats_upstream_quality_average_range` in the Postgres error). That is the intended fail-loud behaviour. Scope of who can actually trip it, traced on re-review: the Aurora/Tension and Kilter Grips syncs **cannot** — both clamp or null out-of-range input before it reaches the DB (`normalizeQualityTo5`, `correctGripsQualityAverage`), and `quality_average` itself is a weighted average of terms already inside `[0,5]`. The one unguarded path is the manually-run MoonBoard problems import (#4617), where a failure surfaces to whoever ran it rather than silently starving a background daemon. So this is not a repeat of the Kilter daemon-starvation shape — but if a sync does start failing right after this ships, check for these constraint names first.

## Validation

- `vp check` — clean (two unrelated pre-existing formatting issues in `.claude/skills/discord-feedback-triage/SKILL.md` / `docs/discord-feedback-pipeline.md` exist on `main` already and are untouched by this PR).
- `vp run typecheck:db` — pass.
- `vp run typecheck:backend` — pass.
- `vp test run --project backend --reporter=agent climb-stats-quality-range-check` — the new regression test, 16/16 pass in isolation.
- `vp test run --project backend --reporter=agent climb-stats` — the broader climb-stats suite (schema-sql.ts change affects every worker-DB test), 97/99 pass; the 2 failures are `TRUNCATE`/`INSERT` **deadlocks** on the shared worker DB caused by heavy concurrent test activity from other agents sharing this same machine right now — re-running `recompute-climb-stats.test.ts` alone reproduced the same deadlock at DB-setup time, before any of this PR's code ran, confirming it's infra contention, not a regression from this change.
- **The migration itself was validated against real data, not just synthetic inserts**: I ran `bun scripts/migrate.ts` against a local container pulled from the exact pinned digest `test-location-sync-integration` uses (`ghcr.io/boardsesh/boardsesh-dev-db@sha256:ce32f2a7...`). The original `> 0` version failed with `PostgresError 23514` on 22,232 real rows; the shipped `>= 0` version applies cleanly end-to-end against the same image.
- No dev-server/UI QA — this is a DB-only guard with no user-facing surface, so the QA-notes/dev-server workflow step doesn't apply.

Re-validated after the takeover rebase onto current `main` (2026-08-19), which was clean:

- `vp run check:db-migrations` — 203 migrations, journal consistent.
- The worker-DB constraint repair above is CI-verified only: parallel agents share the fixed `:5433` worker databases and ports 8082/8084 on this machine, so running the backend suite locally would deadlock other sessions. `test-backend` on the latest push is the proof. `main`'s highest is `0201_high_crusher_hogan`, so `0202` is free and no further renumber is needed.
- Snapshot chain verified by hand: `0202_snapshot.json`'s `prevId` matches `0201_snapshot.json`'s `id`, and the snapshot carries both check constraints with the `>= 0` bound.
- `vp check` clean, `vp run typecheck:db` and `vp run typecheck:backend` both pass.

Closes #3551.

## Release Notes

- [x] No release note needed (internal DB guard, no user-facing change)


marcodejongh:

marcodejongh:
## Adversarial re-review (takeover loop `moarbugs100500`)

> Correction: the first revision of this comment went up with the wrong file's contents (a review of an unrelated playlist-repair script) — shared scratchpad, my mistake. This is the real review of #4305. Sorry for the noise.

**Verdict: the diff is correct and I found no bug in it.** Two things are genuinely open, and both are yours, not the code's: the lock window on `ADD CONSTRAINT`, and the pre-merge prod row check the body asks for. Everything else below I either settled or filed.

Posting before I touch anything so it lands ahead of your merge decision.

---

### 1. I found the writer behind the `>= 0` relaxation — filed as #4617

The body says the MoonBoard 0-sentinel is "either a still-live writer or a 0167 round-trip, I could not chase this further". It is the former, and it is two lines:

- `packages/db/scripts/import-moonboard-problems.ts:246` — `qualityAverage: problem.userRating`
- `packages/db/scripts/import-moonboard-problems.ts:250` — `upstreamQualityAverage: problem.userRating`
- `packages/db/scripts/import-moonboard-problems.ts:69` types the dump field as `userRating: number`, non-nullable. MoonBoard uses `0` for "nobody has rated this yet".

Both go in verbatim with no sentinel mapping. Migration 0151 nulled `quality_average = 0` on 2026-07-07 (22,498 rows, moonboard 19,883); the importer writes it straight back, and since the 0166/0167 split it now seeds `upstream_quality_average` with the same `0` — which is exactly why the CI snapshot shows both columns at 0 on 22,232 moonboard rows. The PR's data finding holds up; only its "cause unknown" caveat does not.

Filed **#4617** with the fix and, importantly, the order it has to happen in: fix the importer, *then* re-run a 0151-style backfill (0151 predates the column split, so `upstream_quality_average` was never cleaned), *then* tighten both CHECKs to `> 0`. Tightening first fails the migration; backfilling first gets undone by the next import run.

I deliberately did **not** fix it in this PR — the backfill is a prod data mutation that needs your sign-off, and the importer fix alone does not let the bound tighten. But without the issue the `>= 0` bound is permanent by default, which is the failure mode worth avoiding here. Both earlier bot reviews flagged the missing tracking issue; that is now closed out.

### 2. The on-call note oversells the blast radius

I traced every writer of both columns. Only one can produce an out-of-range value at all:

| Writer | Guard | Can it violate? |
| --- | --- | --- |
| Aurora / Tension sync | `normalizeQualityTo5` (`packages/shared-schema/src/utils.ts:148`) — null for `q <= 0`, clamps to `[1,3]`, then `2q-1` | No, output is always `[1,5]` or null |
| Kilter Grips catalog / stats-repair | `correctGripsQualityAverage` (`packages/kilter-sync/src/sync/quality-scale.ts:31`) — null for `< 1` or `> 5` | No |
| MoonBoard problems import | none | **Yes** — unclamped pass-through |
| `quality_average` itself | `blendedQualityAverageSql` is a weighted average of terms already in `[0,5]` | No |

So the "a single bad row in a bulk upsert could stall that sync, same shape as the Kilter daemon-starvation incident" line is scarier than the code supports: neither background sync can produce a violating value. The one unguarded path is a manually-run import script, where a failure lands in front of an operator rather than silently starving a daemon. Worth keeping the note, worth softening it — I will adjust the wording.

### 3. The new test does not test the migration (not a blocker, just do not read it as one)

`climb-stats-quality-range-check.test.ts` runs against a worker DB built from the hand-mirrored DDL in `packages/backend/src/__tests__/schema-sql.ts:219-221`, not from `packages/db/drizzle/`. Revert `unified.ts` and the migration, leave `schema-sql.ts` alone, and the test still passes — there is no drift guard between the two.

The migration *is* covered, by `test-location-sync-integration` and `test-dev-db`, which replay the journal from scratch against the seeded image; both green on the current head. So coverage is real, it just is not where the new test is. That is the existing convention for every worker-DB test in the repo and I am not restructuring it here.

### 4. Lock window — your call, and I cannot settle it from here

Both bot reviews asked for `NOT VALID` + a later `VALIDATE`. I do not think it is cleanly available:

- Drizzle's `check()` cannot emit `NOT VALID`, and hand-writing migration SQL is against CLAUDE.md.
- Putting `ADD ... NOT VALID` and `VALIDATE CONSTRAINT` in the same migration file buys nothing if they share a transaction — the `ACCESS EXCLUSIVE` from the ADD is held until commit regardless. It would have to be two migrations across two deploys.

So it is a two-migration dance for a defense-in-depth guard, and whether that is worth it comes down to one number I cannot read: this loop has no prod credential. **If `SELECT count(*) FROM board_climb_stats` is in the low millions the scan is seconds and the plain ADD is fine; if it is much larger, say so and I will split it.**

One correction to the body while you are deciding: migration 0112 is a weaker precedent than it is given credit for. `board_climb_ratings` holds only Boardsesh's own per-user ratings; `board_climb_stats` is every (climb, angle) across three boards. Different order of magnitude.

### 5. The pre-merge prod check is still outstanding

The body's second query — any value `< 0` or `> 5` — is the actual deploy-blocker. If it is not `0, 0`, this migration fails on deploy exactly the way `> 0` did in CI.

Evidence in favour of it being clean: the migration applies end-to-end against the pinned dev-db snapshot (`test-dev-db` and `test-location-sync-integration` both green), and per the table in section 2 no writer can produce a value outside `[0,5]`. But a snapshot is not prod, and I have no credential here, so this one stays with you.

### Minor

- Migration numbering is currently fine: main's highest is `0201_high_crusher_hogan`, this branch carries `0202_black_rocket_racer`, `db-migrations` is green. Re-verifying after I rebase.
- The body still refers to `packages/db/drizzle/0190_black_rocket_racer.sql` — stale after four renumbers. Fixing.

### What I checked and found nothing wrong with

Constraint predicate shape matches `board_climb_ratings_rating_range`; `NULL` is admitted on both columns; the generated SQL contains exactly the two `ADD CONSTRAINT` statements and nothing else; the `schema-sql.ts` mirror is character-identical to the migration; the test asserts via the driver's structured `code`/`constraint` fields rather than a message regex, and covers both columns at `-1 / -0.01 / 5.01 / 0 / 0.5 / 1 / 5 / NULL`; title is a valid conventional commit; Release Notes section present and correct for an internal-only change.

### Next from me

Rebase onto current `main`, re-verify `check:db-migrations`, link #4617 into the PR body and the schema comment, soften the on-call note, fix the stale `0190` reference. Not marking ready — that is yours.

marcodejongh:
## Addendum — one more finding, fixed in `370520a`

Follow-up to my review above. After the rebase I went back over the test wiring and found a real defect I had missed. Verdict is unchanged (**ready to merge once you answer the two open questions**), but this one was worth fixing before you look at it.

### The new test fails on any machine with a pre-existing backend worker DB

The two CHECK constraints were mirrored into the test DDL *inside* the table body:

`packages/backend/src/__tests__/schema-sql.ts:199` — `CREATE TABLE IF NOT EXISTS "board_climb_stats" ( … CONSTRAINT "board_climb_stats_quality_average_range" CHECK (…) )`

But the worker databases are created once and reused forever:

`packages/backend/src/__tests__/worker-db.ts:104-108`
```ts
const [row] = await admin`SELECT 1 FROM pg_database WHERE datname = ${dbName}`;
if (!row) {
  await admin.unsafe(`CREATE DATABASE "${dbName}"`);
}
```

So on any dev box that already has `boardsesh_backend_test_w<N>` from before this branch, the table exists, `IF NOT EXISTS` skips the whole block, the constraints never land — and `climb-stats-quality-range-check.test.ts`'s six `rejects -1 / -0.01 / 5.01` cases fail, because the insert succeeds. Not a flake: deterministic, and it looks like the PR broke something when nothing is wrong with it. CI never sees it — its Postgres container is fresh every run, which is exactly why every check on this PR was green while the problem was sitting there.

This is a known trap in that file, and the file says so itself, a few hundred lines down at `schema-sql.ts:727-731`:

```
-- "gyms" is CREATE TABLE IF NOT EXISTS with no preceding DROP, and the
-- per-worker test DBs persist between runs, so a column added to the block
-- above never lands on a DB that already exists. Backfill it here.
ALTER TABLE "gyms" ADD COLUMN IF NOT EXISTS "hours" text;
```

`370520a` applies the same remedy to `board_climb_stats`: both constraints re-added after the table block in `DO $$ … EXCEPTION WHEN duplicate_object THEN NULL; END $$;` wrappers, matching the dollar-quoted idempotency blocks already in that file at `schema-sql.ts:299-320`. Idempotent on a fresh DB (constraint already present → swallowed), repairs a stale one.

Safe on stale data, for the record: if leftover rows violated the bound the `ALTER` would throw `23514` and take the whole schema apply with it. Nothing writes an out-of-range value — `test-backend` is green with these constraints active on a fresh DB, and stale rows come from those same tests.

I could not verify this end to end locally: parallel loops share the worker DBs on `:5433` and ports 8082/8084, so running the backend suite here deadlocks other sessions. `vp run typecheck:backend`, `vp check` and `vp run check:db-migrations` all pass; `test-backend` on this push is the real proof.

### State after the push

- Rebased onto `a71db63` (current `main`), clean, 0 behind. `main`'s highest migration is `0201_high_crusher_hogan`; this branch keeps `0202_black_rocket_racer`, no renumber needed, `check:db-migrations` reports 203 migrations and a consistent journal.
- `vp check` — the only two errors are pre-existing `no-floating-promises` hits in `packages/db/scripts/migration-journal-verification.integration.test.ts`, untouched by this PR (#4488 lint debt).
- Everything else from my review above is landed: #4617 filed and linked from the schema comment and the test, the on-call blast-radius note softened to match the writer table, the stale `0190` filename reference fixed.

### Still yours, unchanged

1. **Lock window.** Is `SELECT count(*) FROM board_climb_stats` small enough that a full-scan `ADD CONSTRAINT` under `ACCESS EXCLUSIVE` is fine? Low millions → ship as is. Much larger → say so and I will split it into `NOT VALID` + `VALIDATE` across two migrations.
2. **Pre-merge prod check.** The `< 0 OR > 5` query in the body needs to come back `0, 0`. Every writer I traced makes that structurally impossible and the migration replays clean against the pinned dev-db image, but a snapshot is not prod and I have no credential here.

Not marking ready — yours.


---

_Housekeeping: for about four minutes this PR's description showed the description of #4285 — I wrote it from a stale file in a scratchpad several loops share, the same way the first draft of my review above went out with the wrong contents. Restored from the edit history and re-verified; the description is correct now, and only the description was affected. Ignore that entry if you look at the edit history._

PR #4306, opened 2026-08-11
fix(sync): close credential double-claim race under READ COMMITTED

### LINKED ISSUE #3987: Flaky: backend "two parallel claims never return the same credential" (credential-claim-skip-locked)
`packages/backend/src/__tests__/credential-claim-skip-locked.test.ts` → "two parallel claims never return the same credential" failed once on an unrelated PR (#3977, whose diff touches only `packages/db/scripts/`, `scripts/lib/`, `.github/workflows/ci.yml` and `docs/`), then passed on a plain re-run of the same commit.

Failure:

```
FAIL  backend  src/__tests__/credential-claim-skip-locked.test.ts >
  credential claim under concurrent daemon instances (real DB) >
  two parallel claims never return the same credential
AssertionError: expected 1 to be 2 // Object.is equality
 ❯ src/__tests__/credential-claim-skip-locked.test.ts:185:42
Tests  1 failed | 2740 passed (2741)
```

Job: https://github.com/boardsesh/boardsesh/actions/runs/30322234882/job/90160535464
Re-run of the same commit: pass (5m25s).

## Why it's worth a look rather than a retry label

The assertion is `expect(new Set(claimedUserIds).size).toBe(claimedUserIds.length)`, and the received value is the **set size**. So the observed state was two claims that both came back with the **same** `userId` — not one claim returning null, which is the benign outcome the sibling test ("a second instance gets nothing and idles") already covers. Two daemon instances claiming the same credential is the exact thing `FOR UPDATE SKIP LOCKED` is there to prevent, so either:

- the test's `Promise.all` did not actually get two separate pool connections and the sequential path can hand back the same row twice, or
- the claim really can double-hand a credential under some interleaving.

Both are worth knowing. The three deterministic tests in the same file all passed; only the genuine-race companion failed.

Added in #3957. First observed failure — recent `main` CI reds were `test-default`, not `test-backend`.

## Suggested next step

Run the single file in a loop against a real DB (`vp test run --project backend credential-claim-skip-locked`) to get a failure rate, and log both claimed rows on failure so the two cases above can be told apart. Per `CLAUDE.md`, don't run two full backend suites at once while doing it — worker DBs are keyed by `VITEST_POOL_ID`.

marcodejongh:
Fixes the flake in #3987 — and the production bug behind it.

## What was actually wrong

The CI failure was a real double-claim, not test noise. `expected 1 to be 2` was the *set size*: both parallel claims came back with the same `userId`.

`claimNextCredentialForSync` runs `SELECT ... FOR UPDATE SKIP LOCKED` then a separate `UPDATE` that stamps `last_sync_attempt_at`, all under READ COMMITTED. `SKIP LOCKED` only skips a row whose lock is **currently held**. If claimer A locks row X, stamps it and COMMITs entirely inside the window between claimer B's statement snapshot and B's lock attempt on X, the lock is already gone when B arrives. Postgres then follows the update chain and runs an **EvalPlanQual recheck** against the new row version — and EPQ re-evaluates only the `WHERE` quals, never the `ORDER BY`.

Every qual still passed. `credentialRetryReadySql()` short-circuits to TRUE on `consecutive_failures <= 0` no matter how fresh the attempt stamp is, and the board eligibility filter doesn't look at the clock at all. So B locked and returned the row A had just claimed. "Stamp it so it sorts to the back of the queue" is not enough on its own, because the sort is exactly the thing EPQ does not redo.

In production that means two Railway daemon instances logging in and syncing the same user, plus both piggybacking the same shared/catalog sync — the bug #3957 was meant to close, surviving in a narrower window.

## The fix

Claiming has to **falsify a qual**, so the claim's WHERE gains one the stamp breaks:

```
last_sync_attempt_at IS NULL OR last_sync_attempt_at <= now() - make_interval(secs => 30)
```

On the EPQ recheck the freshly-stamped row now fails this and gets thrown out, so the loser either idles or takes the next candidate — both correct outcomes.

Two supporting changes:

- The stamp is written with the DB clock (`now()` + `RETURNING`) instead of `new Date()`, so the gap comparison reads one clock and app/DB skew can't write a stamp that already looks older than the gap.
- The header comment spells out the EPQ mechanism, so a future refactor doesn't drop the gap as "redundant with the ordering" or fold the two statements into a single `UPDATE ... WHERE id IN (SELECT ... FOR UPDATE SKIP LOCKED)` — which has the identical hazard, since no filtered column changes.

## Production semantics change

A credential cannot be re-claimed within 30 seconds of its last claim. The daemon's shortest cycle is 1 minute (`DEFAULT_DAEMON_OPTIONS.minDelayMinutes`), so even a single-credential install never wants to, and the two runners are the only callers. A future "sync now" path calling the claim directly would silently get nothing inside the window — flagged in the constant's doc comment.

## Tests

`packages/backend/src/__tests__/credential-claim-skip-locked.test.ts`:

- **new, deterministic**: a healthy (`consecutive_failures = 0`), committed, unlocked, freshly-stamped row — exactly the state a loser sees during the EPQ recheck — is not claimable. This is the case that fails without the gap qual.
- **new**: the same row is claimable again once the gap has elapsed, so the gap is race protection and not a throttle that parks work.
- **hardened**: the parallel-claims companion now runs 20 reseeded rounds, asserts distinct rows only when both claims returned one (a null loser is legitimate), and names both rows in the failure message so a future red tells a double-hand apart from a null.
- The three existing deterministic tests are unchanged and green.

`credential-selection-backoff.test.ts`: the SQL/JS parity test seeded `last_sync_attempt_at = now()`, which the gap now excludes. Moved to 45 seconds — still inside the fabricated 1-minute window that a missing `GREATEST(n-1, 0)` guard would produce (which is what that test is actually about), and clear of the 30-second gap.

The genuine interleaving can't be forced from a test (it needs a commit to land between another statement's snapshot and its lock attempt), so there's no red-then-green repro of the flake itself. The deterministic just-claimed test pins the state the loser sees; the EPQ reasoning is the part worth a reviewer's eye.

## What the gap does NOT cover

`now()` is `transaction_timestamp()`, so the gap holds for as long as a claim transaction finishes well inside it. This one is BEGIN / SELECT / UPDATE / COMMIT with no application I/O between the statements, so 30 seconds is a very large margin — and a claim that *is* slow keeps holding the row lock, which sends a racer down the plain `SKIP LOCKED` path instead of the EPQ one. The uncovered case needs both at once: a claim that outlives the gap *and* commits inside a racer's snapshot-to-lock window. Only a claim-token column (`claimed_by`, or a `'syncing'` status) rules that out for good, and that is a schema change this fix deliberately doesn't make — the cost if it ever lands is one duplicated cycle, not corruption, because the apply is idempotent (full snapshot re-pull, dedup + `ON CONFLICT`). The header comment now says so rather than reading as an absolute guarantee.

## Validation

- `vp run typecheck:db`, `vp run typecheck:backend`
- `vp check` — clean for this diff; the 4 `no-floating-promises` errors it reports are pre-existing on `main` from #4254 (`packages/mobile/src/providers/__tests__/queue-provider-*.test.tsx`), untouched here
- Backend tests come from CI: ~25 worktrees share the fixed worker DBs on :5433, so a local backend run tells you nothing

Closes #3987

## Release Notes

- [x] No release note needed


marcodejongh:

Issue #4317, opened 2026-08-11
Instrument offline usage — no signal distinguishes offline reads from online ones


marcodejongh:
We can measure who downloads a board, but not whether anyone ever *uses* it offline. No mobile event carries a connectivity state, a served-from-local marker, or any cache-hit signal (2026-08-11 audit of PostHog project 412845). The best proxy available today — "downloader opened the app again on a later day" — is 63.3% (235 of 371), and a 60.9% 7-day-plus active rate, but that measures app retention, not offline usage. The core value proposition of the feature is unmeasured.

This matters for the epic: "get more people using offline mode" needs a success metric better than download counts.

## Direction

Options, cheapest first (decide during implementation):

1. **`servedFrom: 'local' | 'network'` property** on the key read paths that go through the offline interceptor (`packages/mobile/src/lib/graphql/offline-request.ts` already knows which lane it took) — stamped on existing events like search/climb-view/tick where they fire, or captured as a low-volume dedicated event.
2. **Connectivity property** (`online`/`offline` at event time) on mobile events generally, so any event can be segmented.
3. A once-per-session **"session used offline data"** event when the first local-lane read happens while unreachable — lowest volume, coarsest.

Whichever lane: define the epic's north-star from it, e.g. weekly users with ≥1 offline-served read.

Related instrumentation gaps are tracked separately (download funnel events; dead-letter telemetry).

Part of the offline-mode epic: #4319.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01UT6VP1sWqk6bY9mUgyeS2U

marcodejongh:
Two stacked draft PRs cover this, plus the dashboard so the metric exists as a chart and not just an event name.

- **#4322** — `feat(mobile): stamp connectivity on every analytics event` (base `main`). A `connectivity: 'online' | 'offline'` super property registered from `onlineManager` and kept current across transitions and `analytics.reset()`. Every event — `$screen`, `Tick Logged`, #4316's download funnel — becomes segmentable by connectivity for free. Local validation green.
[...]

Issue #4316, opened 2026-08-11
Instrument the offline download funnel: started/failed/cancelled/toggle/download-all + size props


marcodejongh:
The entire offline feature has exactly **one** analytics event: `Offline Board Download Completed`. Confirmed by sweeping the PostHog project (412845) for any event matching offline/download/sync/storage/snapshot — nothing else exists.

Consequences (2026-08-11 audit):

- **Abandonment is structurally unmeasurable.** A user who taps the toggle, waits 9 minutes on Android Kilter, and force-quits produces zero telemetry. Given 37% of *completed* Kilter downloads exceed 5 minutes, the unobserved abandonment population is plausibly large — we cannot size it.
- A download that errors out is indistinguishable from one never attempted.
- No toggle events → can't measure discovery of the entry point, or users turning offline off.
- "Download all" (`use-board-downloads.ts`) is invisible; the 68 multi-board downloaders can't be attributed to it.
- The completion event has no `rowCount` / `bytes` / `climbCount`, so Kilter's slowness can't be normalized against payload size ("just bigger" vs "path is inefficient").
- The flag variant isn't stamped on the completion event.

## What to add (all via `SHARED_EVENTS` in `packages/shared/analytics/src/events.ts`, captured with the existing `track` from `packages/mobile/src/lib/analytics`)

- `Offline Board Download Started` — scopeKey, method-intent (snapshot-eligible or paged), trigger (toggle | download-all | retry)
- `Offline Board Download Failed` — scopeKey, stage (manifest | download | import | paged), cause class, attempt, durationMs
- `Offline Board Toggle` — scopeKey, enabled/disabled (disabled = removal, currently invisible)
- `Offline Download All Tapped` — board count
- Extend `Offline Board Download Completed` with `rowCount`, `bytes`, and the `offline-board-downloads` variant

With Started + Completed + Failed we get completion rate, time-to-abandon, and failure-stage distribution — the metrics the epic's Tier 1 perf work needs to prove itself against.

Part of the offline-mode epic: #4319.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01UT6VP1sWqk6bY9mUgyeS2U

marcodejongh:
Draft PR up: #4337 — `feat(sync): instrument the board-download funnel`. **Closes this issue.** It sits on top of #4336 (progress UI, closes #4311), which sits on #4335 (manifest field). All local validation green.

[...]

PR #4322, opened 2026-08-12
feat(mobile): stamp connectivity on every analytics event

marcodejongh:
Part 1 of two for #4317. The 2026-08-11 PostHog audit found no mobile event carries a connectivity state, a served-from-local marker, or any cache-hit signal — so the best available proxy for "does anyone use a downloaded board offline?" was app retention (63.3% of downloaders reopened on a later day), which measures something else entirely.

This PR fixes the cheap half: every event the app sends now carries whether the network was usable when it was captured.

`connectivity: 'online' | 'offline'` is registered as a PostHog super property from React Query's `onlineManager` at analytics startup, and re-registered on each real online↔offline transition. Same-value NetInfo churn is ignored, so a flaky network costs no extra persisted writes — the steady-state cost is one small write per actual transition.

Two details that make this worth doing as a super property rather than a per-event prop:

- **PostHog resolves super properties at capture time**, so `$screen`, `Tick Logged`, and the download-funnel events from #4316 all become segmentable by connectivity retroactively, with no work at any call site.
- **Offline-captured events really do arrive later.** Verified in the SDK source rather than assumed: `posthog-react-native` defaults to `persistence: 'file'`, and `@posthog/core`'s flush only persists the dequeue when the failure is **not** a `PostHogFetchNetworkError` — so a batch that fails offline stays in the on-disk queue, survives an app kill, and flushes on reconnect still carrying the `connectivity: 'offline'` it was captured with.

It is also re-registered inside `analytics.reset()`. PostHog's reset clears every super property and `getPostHogClient()` caches the singleton, so without that a sign-out silently drops connectivity for the rest of the launch — the exact trap `environment` (#3814) and `$raw_user_agent` already document there.

Known limitation, documented rather than papered over: this is best-known connectivity, not ground truth. `onlineManager` seeds from NetInfo asynchronously and defaults to online, and it tracks `isConnected`, not `isInternetReachable` — so a genuinely-offline cold start can stamp its first events `online`, and a captive portal reads `online` throughout. Both errors under-count offline usage, so any number built on this property is a floor, never an inflation.

## Changes

- `packages/mobile/src/lib/analytics-connectivity.ts` (new) — `currentConnectivityState()`, `registerConnectivitySuperProperty()`, `startConnectivityTracking()`. Best-effort and non-throwing, matching `registerMobileUserAgent` / `registerAppEnvironment`; a no-op when analytics is disabled (dev / no key).
- `packages/mobile/src/lib/analytics.ts` — `reset()` re-registers connectivity alongside the build-level super properties.
- `packages/mobile/src/components/analytics/AnalyticsProvider.tsx` — starts/stops the subscription in an effect, declared before the `!client` early return so hook order stays stable when analytics is off.
- `docs/offline-sync-plan.md` — new `## Usage telemetry` section covering what the property means, the buffer-and-flush behaviour, and the `onlineManager` caveat.
- Tests: new `analytics-connectivity.test.ts` (registers on start, re-registers only on a real transition, unsubscribes cleanly, swallows a throwing/rejecting `register`), plus a case in `analytics-reset.test.ts` for the post-reset re-registration.

No new dependencies, no native config — `@tanstack/react-query` and `posthog-react-native` are already in the native graph. Ships over OTA on the existing production channel.

## Release Notes

## Test plan

- `vp run test:mobile` — 622 files / 5491 tests passed.
- `vp check` — clean for every file this PR touches (two pre-existing formatting misses on `main`, in `.claude/skills/discord-feedback-triage/SKILL.md` and `docs/discord-feedback-pipeline.md`, are untouched here).

Not verifiable on a dev build: `isAnalyticsEnabled = !!apiKey && !__DEV__`, so Metro only exercises the debug console hook and never sends. End-to-end confirmation (go offline, browse, kill the app, reconnect, check the events land with `connectivity: 'offline'` and their original timestamps) needs the `pr-<number>` OTA preview channel against a TestFlight/store binary.

### DIFF
--- a/docs/offline-sync-plan.md
+++ b/docs/offline-sync-plan.md
@@ -851,6 +851,45 @@ On logout or account switch:
 4. Reset `syncEnabledBoards` and the `offlineBoardsV1` snapshots (board names must not survive into the next account on a shared device).
 5. On new login, sync pull client fetches the new user's data (small, seconds).
 
+## Usage telemetry
+
+Offline mode's whole value proposition is "you can still climb when the signal
+can't". Until issue #4317 nothing in the telemetry could tell an offline-served
+read apart from an online one, so that claim was unfalsifiable. This section is
+the source of truth for the signals that fix it. They are mobile-only today (the
+engine is shared, so a future web offline consumer would emit the same things).
+
+### The `connectivity` super property
+
+`packages/mobile/src/lib/analytics-connectivity.ts` registers a PostHog super
+property `connectivity: 'online' | 'offline'` at analytics startup and again on
+every real network transition. PostHog stamps super properties onto events at
+**capture** time, so every event the app sends — `$screen`, `Tick Logged`, the
+offline download events — becomes segmentable by connectivity, retroactively and
+for free.
+
+Three things are worth knowing before trusting a number derived from it:
+
+- **Offline-captured events do arrive.** The RN SDK persists its queue to disk
+  (`persistence: 'file'` is the default), and `@posthog/core`'s flush only
+  dequeues a batch when the failure was **not** a `PostHogFetchNetworkError` — a
+  network failure leaves the batch in the queue. So events captured in airplane
+  mode survive an app kill and flush on reconnect, still carrying the
+  `connectivity: 'offline'` they were captured with. The queue is capped at 1000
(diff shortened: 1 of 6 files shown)


marcodejongh:

Issue #4313, opened 2026-08-11
Transport failures burn snapshot-bootstrap attempts, stranding boards on the paged crawl forever


marcodejongh:
Sentry (`source:offline-sync`, `kind:snapshot-bootstrap`, last 60d): 123 events / 75 users of snapshot bootstrap failures, and every sampled event carries `cause: "Network request failed"` — plain transport failures fetching the manifest or artifact on flaky mobile connections. Top issues:

- BOARDSESH-AE — `kilter:1:10` stage "manifest" (attempt 1): 58 ev / 37 users
- BOARDSESH-AF — `kilter:1:28` stage "download" (attempt 1): 26 ev / 17 users
- BOARDSESH-C2 — `moonboard:2:1` stage "manifest" (attempt 2): 17 ev / 10 users
- BOARDSESH-C5 — `kilter:1:10` stage "download" (attempt 1): 14 ev / 9 users

## The problem

`MAX_BOOTSTRAP_ATTEMPTS = 2` (`packages/shared/offline-sync/src/sync/snapshot-bootstrap.ts:46`). Two failed attempts — including pure transport failures — and the scope is marked `bootstrap-paged-fallback` and stays on the paged crawl indefinitely, until a scope teardown resets it (`pull-client.ts:720-740`, `docs/board-snapshots.md`).

For Kilter that punishment is brutal: the paged crawl is 400+ sequential GraphQL round trips per size scope, each page paying the correlated-`EXISTS` cost against a ~200k-row `board_climbs` table (deferred issue #3561). PostHog shows the fallback is real in the field: 33 of 473 completions (7%) used `method: paged`, clustered on `kilter:1:10` (13 of 33), and paged completions average 10–35 minutes.

Someone who taps "download" in a gym with bad reception (exactly the offline-mode target user) can permanently downgrade their board to the slow path.

## Direction

- Don't count pure transport failures ("Network request failed", timeouts) toward the attempt cap — only structural failures (schema-stale, corrupt artifact, gzip mis-decode) should burn attempts.
- Resumable/retried artifact download with backoff so one dropped connection doesn't void a 103 MB transfer.
- A periodic heal that retries the snapshot path for scopes stuck in `bootstrap-paged-fallback`.

Check overlap with #4238 (closed 2026-08-11) first — it fixed offline launches disqualifying the fast path and error-reporting levels, but the transport-failure attempt-burning and the permanent fallback look untouched.

Part of the offline-mode epic: #4319.

[...]

Issue #4310, opened 2026-08-11
Kilter offline download is slow on the snapshot fast path (p50 2m55s, Android p50 8m52s)


marcodejongh:
Kilter is the most-downloaded offline board (46% of all downloads) and by far the slowest — on the **snapshot fast path**, not the paged fallback. PostHog (`Offline Board Download Completed`, 2026-07-15 → 2026-08-11, project 412845):

| Board | Method | N | p50 | p90 | p99 | max |
|---|---|---|---|---|---|---|
| kilter | snapshot | 205 | **2m55s** | 11m58s | 23m27s | 31m39s |
| moonboard | snapshot | 156 | 14.5s | 1m08s | 16m12s | 16m13s |
| tension | snapshot | 64 | 1m27s | 3m05s | 10m58s | 16m10s |

- 37% of Kilter downloads take **over 5 minutes**; 29 took over 10.
- Android Kilter p50 is **8m52s** (p90 21m) — 3.7× iOS.
- `offline-snapshot-bootstrap-v2` is at 100% (only 2 users ever saw `false`), so this is not a rollout gap. The fast path itself is slow at Kilter scale.

## Why

The `kilter:1` snapshot artifact is **271 MB (103 MB gzip)** — about half the entire snapshot corpus (`docs/board-snapshots.md`, 2026-07-27 benchmark). The client:

1. Downloads the whole per-layout artifact even when the user wants one size scope (`kilter:1:10` is 143 of 219 Kilter downloads). The artifact is cached per `(boardType, layoutId)` within a cycle, but a single-size user still pays for all of layout 1.
2. Runs `PRAGMA quick_check` + `COUNT(*)` verification + reconcile deletes + `INSERT OR REPLACE … SELECT` import all inside **one `BEGIN EXCLUSIVE` transaction** (`packages/shared/offline-sync/src/sync/snapshot-bootstrap.ts:640-723`), blocking every other SQLite writer for the duration (see also the `database is locked` issue).
3. The code's own comment estimates a healed artifact download "can run for 18 minutes" (`snapshot-bootstrap.ts:730-739`).

## Candidate directions (not prescribed — pick during implementation)

- **Per-size (scope-level) artifacts** in the nightly export instead of per-layout, so `kilter:1:10` downloads only its own rows. Biggest lever for the common case.
- Trim artifact columns / rows to what the client schema actually imports.
- Chunked import (batched transactions) instead of one exclusive mega-transaction — also unblocks concurrent writers.
[...]

Issue #4238, opened 2026-08-10
Snapshot bootstrap reports offline users as errors, download failures report no cause, and two offline launches permanently disqualify the fast path


marcodejongh:
**Sentry:** BOARDSESH-C2, AE, C5, AF — 57 events / 35 users in 14d, `kind: snapshot-bootstrap`, `source: offline-sync`, `handled: yes`, `level: error`. All four titles name a specific board config, which is misleading: Sentry titled each group from its first event, and the events inside span eight configs (kilter:1:10 ×25, moonboard:2:1 ×14, tension:10:6 ×6, tension:11:6 ×4, kilter:1:7 ×3, moonboard:5:1 ×2, moonboard:3:1 ×2, kilter:1:28 ×1).

**Priority: P2.** Driven by the permanent fast-path disqualification (item 3), not the noise.

**No board config is broken.** The live manifest (fetched 2026-08-10) has every one of them at the current `schemaVersion: 4`, and PostHog `Offline Board Download Completed` shows all of them succeeding for far more users than fail: kilter:1:10 for 77 users, moonboard:2:1 for 63, tension:10:6 for 34. Failure volume just tracks board popularity. Note the manifest is keyed per `(boardType, layoutId)` — `sizeId` isn't part of the lookup (`pull-client.ts:597`) — so kilter:1:10 and kilter:1:28 are the same artifact and one failure emits two events.

### 1. Stage `manifest` is 100% offline users, reported at error level

Every one of the 28 manifest-stage causes in Sentry is a transport failure: 24 × `Network request failed`, 4 × `Network request timed out`. Zero HTTP statuses. That's expected, because a 404 or an unexported layout never reaches telemetry — `snapshot-source.ts:108` returns null → `{ status: 'absent' }` → `continue` with no attempt counted (`pull-client.ts:587`, `:598`).

The reason these arrive at all is that `pullSync` has no connectivity gate. The drainer has one (`drainer.ts:217`, `if (!options.isOnline()) return;`); `pull-client.ts:815` only checks `isBackgrounded()`. `startSyncScheduler` fires on mount, on every foreground, and on every offline→online edge (`sync-scheduler.ts:135-163`), and `offline-sync-bridge.tsx:84-101` mounts it for every signed-in user with the flag on. So every offline launch produces one error event per enabled-but-undownloaded board.

[...]

PR #4326, opened 2026-08-12
fix(sync): bound bootstrap retries by failure kind, not one shared counter

marcodejongh:
## Why

`MAX_BOOTSTRAP_ATTEMPTS = 2` was doing two jobs at once, and #4313 is the story of them being fused.

It was the **retry policy** — there isn't one; the ~100 MB artifact GET is unresumable and never retried — and it was the **total-spend bound**, the thing stopping a device from restarting that GET on every foreground. Because a dropped connection at the *download* stage burned the same counter a corrupt artifact does (`settleRetryableBootstrapFailure(..., { exemptTransportFromCap: false })`), **two bad-reception launches condemned a board to the 400+-round-trip paged crawl for the life of the install**. Sentry over 60 days: 123 snapshot-bootstrap transport failures across 75 users, with `UnableToDownloadException: The request timed out` — which `isNetworkError` classifies as transport — as the headline cause.

Worse, the strand locked itself in. An over-cap scope was not added to `skipPagedPull`, so the same cycle's crawl wrote `checkpoint:board_climbs:<scope>`, and the first eligibility check (`if (climbsCheckpoint || statsCheckpoint) continue`) disqualified the scope from the snapshot path forever. The #4238 one-shot heal is spent once per scope and never comes back.

This splits the two jobs. Kind-specific budgets bound **total spend**; a persisted cooldown ladder bounds **frequency**. Both, always — an exemption without a budget would be the failure the old comment at `pull-client.ts:554-563` was guarding against.

## Changes

**New `packages/shared/offline-sync/src/sync/bootstrap-retry.ts`** — the whole decision surface, pure (clock and jitter injected), plus its persistence:

| Kind | Raised by | Budget | Cooldown | Re-armed by a new `builtAt`? |
| `transport` | `isNetworkError(cause)` at the download stage | 3 consecutive, reset by any successful download | 2 min → 15 min → 2 h | no |
| `structural-artifact` | anything thrown inside the import | 2 | 6 h → 24 h | yes — **once per scope, lifetime** |
| `structural-device` | every other non-transport cause (disk, cache dir, CDN non-2xx, unclassifiable) | 2 | 6 h → 24 h | **no** |

The manifest stage stays entirely free for transport failures, exactly as #4238 shipped. `structural-device` is the conservative default because a plain `Error` from an adapter's downloader can't be told apart from a device fault — and the export is **nightly**, so a `builtAt` reset for a disk-full phone would be 2 × ~100 MB every day, forever.

**Worst-case lifetime spend per scope: 7 artifact downloads** (3 + 2 + 2), each behind at least one cooldown rung. A test drives 40 cycles against an injected clock and asserts the `downloadArtifact` call count, so loosening it shows up in a diff.

- **Heal-over-partial.** A scope that has checkpoints, never reached `scope-complete:`, and carries snapshot-path failures can now import an artifact over its partial catalog. A `scope-complete:` scope is never healed — it already serves the whole thing locally.
- **`SnapshotWatermarkRegressionError`.** `bootstrapScopeFromSnapshot` takes the scope's current checkpoints on the heal path and refuses — before the exclusive transaction opens, so **nothing at all is written** — when the artifact's scoped watermark is behind them. Covers both hazards at once: an artifact whose scope filter matches zero rows (`tableWatermark` returns the epoch) and a crawl that already ran past the artifact. Without it the import would lower `checkpoint:board_climbs` and rewind the single global deletions cursor with it. The refusal **spends the structural-artifact budget**, like any other import failure: it is deterministic for a given artifact (local checkpoints only move forward), so a refusal that recorded nothing left the scope as eligible as before and pulled the whole ~100 MB again every single cycle, forever. A five-cycle test pins the download count.
- **One eligibility function, two callers.** `evaluateBootstrapEligibility` is called by `runBootstrapPhase` *and* `estimateScopeDownload`. `snapshot-estimate.ts` used to carry a copy of the rules under a comment promising they mirrored the engine "one-for-one"; heal-over-partial made both copies wrong at once. A parity table in the tests pins them row for row.
- **Grace-window skip.** Only a fresh, non-terminal scope whose retry lands within 30 minutes waits for it. Past that the crawl runs, so a board is never left empty on a 2-hour cooldown — and a scope that already holds rows always crawls, so a failed heal can't stall progress.
- **Legacy dual-write, never delete.** `bootstrap-retry:<scopeKey>` is the source of truth, but every write mirrors `bootstrap-attempts:` / `-healed:` / `-paged-fallback:`, and reads fold back anything a rolled-back bundle counted (`mirroredAttempts`). Production channel rollback and `pr-<n>` previews are live paths here.
- **New seams**, all defaulted so web and every existing caller are byte-identical: `isOnUnmeteredNetwork` (gates the automatic heal only), `now`, `random`, `onBootstrapRetryScheduled`, `onBootstrapPathRecovered`, and `ScopeDownloadCompleteInfo.bootstrapHealed` so #4316 can keep healed scopes out of its snapshot-vs-paged duration comparison. A fresh bootstrap ignores the metered probe because the size dialog just closed; a user-requested retry ignores it because `restoreBootstrapRetryBudget` arms a `userRequested` flag on the persisted state that overrides the defer and is spent the instant the download starts — one tap, one artifact. `bootstrapHealed` reads the persisted `bootstrap-done:` value (`'heal'` vs `'1'`), not an in-memory set, because a healed scope usually reaches completion cycles after the import (`board_climb_grades` is still crawling) and an in-memory set would report `false` for exactly those runs.
- Both `MAX_BOOTSTRAP_ATTEMPTS` read-sites move in this PR (`snapshot-estimate.ts`, `board-offline-state.ts`) so no intermediate merge leaves the constant's meaning inconsistent. `bootstrap-retry:` joins `scopeSyncMetaKeys`. `BOOTSTRAP_METADATA_QUERY` is now built from `BOOTSTRAP_METADATA_PATTERNS` with a test pinning the placeholder count.

**Five existing tests encoded the old behaviour as deliberate, and they were right to.** The cap really was the only spend bound. They're rewritten to assert the same intent — a permanently-failing artifact must still stop, and the crawl must still deliver a board meanwhile — expressed as bounded lifetime download counts instead of a 2-attempt cliff.

## What this does not fix

`SNAPSHOT_TABLES` is climbs + stats only, so a healed scope still crawls `board_climb_grades` before it reports complete. This removes part of the slow path, not all of it. Resumable downloads are #4310's — artifacts ship gzip-encoded, so byte-offset resume needs a JS gunzipper (a native dep, OTA-forbidden). When that lands, the transport ladder's first rung should drop.

## Release Notes

A dropped connection while downloading a board no longer condemns it to the slow download path. Boards that got stuck partway through the slow crawl now finish on the fast one by themselves.

## Test plan

- `vp test run --project offline-sync --reporter=agent` — 458 passed, 25 files (rebased onto main's #4329/#4330 suites). New `bootstrap-retry.test.ts` covers the ladders, jitter bounds, both budgets, the artifact-only `builtAt` re-arm and its lifetime cap, the eligibility matrix, legacy migration + dual-write + read reconciliation, and `restoreBootstrapRetryBudget`. New blocks in `snapshot-bootstrap.test.ts` cover the 40-cycle bounded-spend counts per kind, heal-over-partial (including the `bootstrapHealed` report both in-cycle and one cycle late, and the metered deferral), four watermark-guard cases (including five cycles asserting one download, not five), and the user-requested retry end to end: budget restored → the artifact lands on cellular with `trigger: 'user-request'`, and the second cycle defers because the tap is spent.
- `vp run typecheck` — clean across the monorepo.
- `vp run test:mobile` — 5601 tests, all passing.
- `vp check` — clean for every file this PR touches.

Device QA is in `.boardsesh/qa-notes.md` (untracked): airplane-mode enable, three mid-artifact drops, a hand-edited past `bootstrap-retry:` on a mid-crawl scope, a seeded behind-checkpoint artifact for the guard, and driving a scope terminal.

Refs #4313

### DIFF
--- a/docs/board-snapshots.md
+++ b/docs/board-snapshots.md
@@ -254,7 +254,12 @@ scope that already holds rows always crawls — a failed heal must not stall pro
 **The metered-network probe.** `SyncOptions.isOnUnmeteredNetwork` (default `() => true`, so web and every
 existing caller are unchanged) gates ONE decision: the automatic heal of a partly-crawled scope, which is a
 ~100 MB download the user did not ask for that day. On a metered link the heal defers 6 hours and burns
-nothing. A fresh bootstrap (they just confirmed the size) and a user-requested retry ignore it.
+nothing. A fresh bootstrap (they just confirmed the size) ignores it, and so does a user-requested retry:
+`restoreBootstrapRetryBudget` arms a `userRequested` flag on the persisted state, `runBootstrapPhase` skips
+the defer while it is set, and the flag is spent the instant the download starts (one tap, one artifact).
+That last part matters — a settled scope has always crawled, so it can only ever come back as a
+heal-over-partial, and without the flag "Try the fast download again" was a silent 6-hour deferral on
+cellular. A recovery on that path reports `trigger: 'user-request'`.
 
 **The watermark-regression guard.** `bootstrapScopeFromSnapshot` takes the scope's current board-table
 checkpoints on the heal path and throws `SnapshotWatermarkRegressionError` — before opening the exclusive
@@ -265,6 +270,12 @@ after any export/client filter drift), and a crawl that already ran past the art
 import would lower `checkpoint:board_climbs:<scope>` — destroying exactly the progress the heal exists to
 rescue — and rewind the single global deletions cursor with it. It is reported at full severity
 (`expected: false`) because it means the export's scope filter and the client's disagree.
+
+The refusal **burns the structural-artifact budget**, like any other import failure. That is not bookkeeping
+pedantry: the refusal is deterministic for a given artifact (local checkpoints only move forward), so a
+refusal that recorded nothing left the scope exactly as eligible as before and pulled the whole ~100 MB
+again on every single sync cycle, forever. Charged, it costs at most 2 downloads (6 h then 24 h apart) plus
+the one re-armed round a genuinely newer `builtAt` can grant.
 
 With `W ≥ C` enforced the coverage argument is simple: rows ≤ C are already local, the artifact supplies
 every scoped row ≤ W, `reconcileScope` removes local scoped rows ≤ W absent from the artifact, and the
@@ -278,9 +289,11 @@ would re-arm a fresh 2-attempt round plus another one-shot heal. Reads reconcile
 legacy counter moved past what we last mirrored (`mirroredAttempts`), an older bundle counted something
 real and the difference is folded back into `structuralFailures`. On first touch,
 `migrateLegacyBootstrapMarkers` derives a state from the legacy rows and grants one clean pass under the
-new taxonomy (the old counter conflated transport with structural, which is the bug), leaving the rows in
-place; a scope that already holds rows gets a `retryAfter` jittered across 2 hours so the fleet does not
-start downloading on the same post-OTA launch.
+new taxonomy (the old counter conflated transport with structural, which is the bug); a scope that already
+holds rows gets a `retryAfter` jittered across 2 hours so the fleet does not start downloading on the same
+post-OTA launch. The legacy rows are never deleted, but the first write after a migration does re-stamp
+`bootstrap-attempts:` down to the mirrored value — the clean pass has to be visible to a rolled-back bundle
+too, or it would re-read the pre-migration count and settle the scope all over again.
 
 A scope torn down from **More → Storage** becomes eligible again: `removeBoardScopeData`
 (`sync/scope-teardown.ts`) clears all three board-data checkpoints **and** its `bootstrap-attempts:` /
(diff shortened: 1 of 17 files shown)


marcodejongh:
Pushed `55ef79ca8`, which fixes both QA holds and the two doc inaccuracies, and rebased the branch onto current main (the #4329/#4330 coverage-telemetry stack landed in the meantime; conflicts were additive — `CoverageEvaluated*` and the new bootstrap reporters now sit side by side in `pull-client.ts`, `sync-scheduler.ts` and `index.ts`). The blocking one: `SnapshotWatermarkRegressionError` no longer falls into the free permanent-miss branch. It is charged as `structural-artifact` like any other import failure, which is right on the merits as well as on the loop — the refusal is deterministic for a given artifact, since local checkpoints only move forward, so a refusal that recorded nothing left the scope exactly as eligible and pulled the whole ~100 MB again on every cycle, forever. Charged, it costs at most 2 downloads (6 h then 24 h apart) plus the one re-armed round a genuinely newer `builtAt` can grant, and the new test drives five cycles asserting `downloadArtifact` was called once, then jumps past the cooldown and asserts the second attempt settles the scope. The second hold is fixed the way the review suggested: `bootstrapHealed` now rides the persisted `bootstrap-done:` value (`'heal'` vs `'1'`) instead of the per-cycle Set, so the common shape — the grades crawl finishing the scope a cycle or three after the import — reports `true`; a test pins that across two cycles. I also implemented the thing the comments were claiming rather than deleting the claim: `restoreBootstrapRetryBudget` arms a `userRequested` flag on the persisted state, `runBootstrapPhase` skips the metered defer while it is set and spends it as the download starts (one tap, one artifact), and the recovery reports `trigger: 'user-request'` — the enum member that had no emitter. The two stale doc lines are corrected too: the legacy rows survive but `bootstrap-attempts:` is re-stamped to the mirrored value, and the deletions replay window is bounded by the 80-day tombstone retention, not one nightly export. Validation, all foreground: `vp test run --project offline-sync` 458/458 (25 files), `vp run test:mobile` 5601, `vp run typecheck` clean, `vp check` clean for every file here. The open product question (should the automatic heal confirm?) and device QA are unchanged and still yours.

Issue #4315, opened 2026-08-11
Offline data loss is invisible: dead-letter telemetry, coverage-reset event never fires, dropped offline ticks


marcodejongh:
Three blind spots around losing user data offline, found in the 2026-08-11 telemetry audit:

## 1. Dead-lettered mutations are invisible

`packages/shared/offline-sync/src/mutation-queue/queue.ts:52-73` flips a row to `dead_letter` after `max_retries`; `drainer.ts:306-325` emits `status: 'dead_letter'` through `notifyMutationStatus` — a **UI seam only**. There is no `reportHandledError` and no analytics event anywhere on that path (grep for `dead_letter` across `packages/shared/offline-sync` + `packages/mobile/src` finds no telemetry call). A tick or favorite written offline that exhausts its retries vanishes without a trace in both Sentry and PostHog. The drainer's network-error handling is correct by design (network errors leave rows PENDING without burning `retry_count`), so dead-lettering means a non-network permanent failure — exactly the case we most need to see.

Add: a Sentry handled error + a `SHARED_EVENTS` capture (mutation kind, retry count, error class) when a row dead-letters.

## 2. `Offline Sync Coverage Reset Forced` has never fired

The event is not in the PostHog project taxonomy at all — zero occurrences ever. Either no device has exceeded tombstone retention yet (plausible, feature is ~4 weeks old and #4048 shipped recently) or the capture at `packages/mobile/src/offline/offline-sync-adapter.ts:145` is unreachable in production. Verify the wiring (e.g. confirm the code path executes in a forced-retention test) so we can trust "zero" means zero.

## 3. Offline users lose ticks outright on local-write failure

`kind: tick-local-write` in Sentry: 72 events / 56 users (mostly 2.2.2; collapsed to 1 event on 2.3.1 — verify it stays down). When `writeTickLocal` throws, `packages/mobile/src/providers/board-adapter.tsx:141` returns null and falls through to a direct network save — fine online, but an **offline** user's tick is dropped entirely. Decide during implementation whether "queue the tick anyway / retry the local write" belongs here or in a follow-up issue.

Part of the offline-mode epic: #4319.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01UT6VP1sWqk6bY9mUgyeS2U

marcodejongh:
Two draft PRs for this, both observation-only:

[...]

Issue #4314, opened 2026-08-11
SQLite "database is locked" persists on 2.3.1 during sqlite-init after #3858/#4104


marcodejongh:
After the WAL/busy-timeout fix (#3858, PR #3887) and the retry-contended-init fix (#4104, PR #4121), SQLite `database is locked` (error code 5) is still arriving from current releases. Sentry `source:offline-sync`, last 60d — `kind: sqlite-init` totals 82 events / 71 users and persists on **both platforms across every release**, including 2.3.1 (10 users on 2.3.1+2 iOS, 10 on 2.3.1+2000824 Android):

- BOARDSESH-CC — `prepareAsync` FunctionCallException → SQLiteErrorException code 5 (`SQLiteModule.swift:382`): 15 ev / 15 users, first seen 2026-08-04, **last seen 2026-08-11**
- BOARDSESH-CF — `NativeDatabase.prepareAsync` rejected → code 5: 11 ev / 11 users, through 2026-08-10
- BOARDSESH-A9 — `finalizeAsync` → locked: 49 ev / 34 users (tail ends 2026-08-04, likely the pre-#4121 population)
- BOARDSESH-AX — `prepareAsync` → locked: 30 ev / 25 users, through 2026-08-06

For contrast, `kind: tick-local-write` lock errors collapsed from 52 events on 2.2.2+3 to **1** on 2.3.1+2 — so something in that lane got fixed while the init lane didn't.

Hypothesis from the error shapes: init-time contention between concurrent writers (pull-client bulk import — which for Kilter holds one exclusive transaction for the whole snapshot import — vs drainer / migrations) rather than corruption. No `SQLITE_CORRUPT` signatures anywhere.

Footnote: BOARDSESH-BN — `prepareAsync` → **error code 10 (disk I/O)**, 3 ev / 3 users (2026-08-01 → 08-05). Separate small disk-pressure family, same one as the thumbnail-cache "volume out of space" issues (BOARDSESH-C6/C7/C8/CZ).

Part of the offline-mode epic: #4319.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01UT6VP1sWqk6bY9mUgyeS2U

marcodejongh:
## The premise here does not hold — every one of these events predates the #4104 fix reaching devices

I went looking for post-#4104 failures and could not find any. The proof is a tag that only exists in the fixed code.

[...]

PR #4329, opened 2026-08-12
feat(sync): surface offline outbox write losses in telemetry

marcodejongh:
Issue #4315 asks to make offline data loss visible. Three of the four losses below emit nothing at all today, and the fourth emits a Sentry count that can't tell a saved tick from a lost one.

**A dead-lettered write is invisible.** `recordFailure` flips a row to `dead_letter` in one atomic UPDATE (`mutation-queue/queue.ts`) and `markDeadLetter` force-flips non-retryables; both only reach `notifyMutationStatus`, whose sole consumer is mobile's optimistic-UI seam. `DrainOptions` had no telemetry hook at all. And a dead letter is never a connectivity problem — a network error takes the `networkStop` branch without burning `retry_count` — so every one of them is a permanent loss of something the user did, with no Sentry event and no PostHog event anywhere. The only surface is More → Sync issues, which the user has to happen to visit while online.

**Sign-out DELETEs the whole outbox, dead letters included.** `USER_DATA_TABLES_TO_CLEAR` lists `pending_mutations` and `clearUserData` deletes it. The best-effort pre-sign-out drain returns early when `getPendingCount === 0`, and that filters `status = 'pending'` — so a user whose only outstanding writes are dead letters gets no drain attempt and no record before the rows vanish. This also defeats a launch-time gauge on its own: the backlog is wiped before the next launch, which is why both gauges exist.

**A dead-lettered favorite/follow key swallows every retry.** `enqueue` is `INSERT OR IGNORE` against a UNIQUE `idempotency_key`, and favorite/follow keys are deterministic (`add:user_favorites:<board>:<climb>:<angle>`). The cancel DELETEs in `addFavoriteLocal`/`removeFavoriteLocal` match only `status = 'pending'`. So once a favorite add dead-letters, every later add for that climb/angle is dropped at enqueue time: local row present, no queue row, no drain, and therefore no dead-letter event could ever mention it. Ticks are immune (fresh uuid per tick); favorites and follows are not.

**A failed local tick write can't be read.** Sentry `kind:tick-local-write` over 90 days: 72 events, all `database is locked`, 71 on 2.2.x and exactly 1 on 2.3.1 (WAL + `busy_timeout` from #3887 did the work). But the report can't say whether the `return null` fall-through then landed over the network or dropped the tick — that turns entirely on whether the device was offline, which nothing recorded.

## Changes

**Engine (`@boardsesh/offline-sync`)**

- `onMutationDeadLettered` on `DrainOptions`, fired from both dead-letter branches with `reason: 'retries_exhausted' | 'non_retryable'`, the burned `retryCount`, `queuedForMs`, the resolved status, and the original error so the platform reporter can attach it as `cause`. Failure-isolated exactly like `notifyMutationStatus`. Deliberately a separate seam rather than widening `MutationDeliveryEvent`, which is mirrored field-for-field by `OfflineMutationDelivery` in `@boardsesh/board-react`.
- `recordFailure` returns `{ status, retryCount }` from the same `RETURNING`, so the attempt count needs no second read racing the row. Package-internal — the index never exported it.
- `enqueue` returns `{ inserted, existingStatus }`. Only the suppressed path pays for the extra `SELECT`, and a driver that doesn't report `changes` degrades to "inserted" rather than inventing a loss report.
- `getOutboxSummary` — one grouped `SELECT` for per-status counts and oldest `created_at`.
- `parseQueueTimestamp` — `created_at` is SQLite `datetime('now')`, i.e. `'YYYY-MM-DD HH:MM:SS'` with a space separator and no zone. That is not ISO-8601, so `Date.parse` is implementation-defined: V8 accepts it (a Node-only test passes) and Hermes is stricter, so the device would return null forever while the metric read as "no data".
- `isDatabaseLockedError` in `db/lock-errors.ts`, shared with #4314 so reporting and any future retry/defer logic can't disagree on which failures count. The busy result code is a secondary signal only, because the real Android strings carry a raw U+0005 where the digit belongs (BOARDSESH-6V, all 18 events) while iOS carries a plain `Error code 5:`.

**Analytics** — five `SHARED_EVENTS` entries, all `Offline <Noun> <PastTenseVerb>`: `Offline Mutation Dead Lettered`, `Offline Outbox Backlog Detected`, `Offline Outbox Discarded On Sign Out`, `Offline Mutation Enqueue Suppressed`, `Offline Tick Local Write Failed`.

**Mobile**

- Adapter binds the dead-letter seam to both Sentry (`kind: 'mutation-dead-letter'`, with the original error as `cause`) and PostHog. Composed rather than defaulted, so no call site can opt out of it by passing its own handler. The idempotency key rides the Sentry extra only — as an analytics prop it is unbounded cardinality.
- `offline/outbox-telemetry.ts` holds the three non-drain gauges. The launch backlog gauge runs from `OfflineSyncBridge` in **both** flag branches (the bridge already drains a kill-switched user's leftovers, and their dead letters are exactly as lost), self-guarded so an effect re-run can't double-fire. The sign-out gauge runs inside `runSignedOutCleanup` — covering manual, forced-401, and proactive-expiry sign-out — and **strictly before `resetAnalytics()`**, or the event lands on an anonymous distinct_id and can never be joined back to the account that lost the writes. There's an explicit call-order assertion for that.
- Suppressed-enqueue reporting on favorites and follows, and only when `existingStatus === 'dead_letter'`. A suppression against a `pending` row is correct dedup of a double tap and reporting it would bury the signal.
- The `tick-local-write` report keeps its `kind` tag verbatim so the 90-day trend stays comparable, and gains `was_offline` and `is_lock_error`. `return null` is unchanged.

**Not in scope.** No durability fix for the tick write path. The original plan's retry-ladder-plus-queue-only fallback was dropped: `writeTickLocal` already applies `busy_timeout = 5000` inside its transaction, so every failure surviving on 2.3.1 is by construction a lock held ≥5s — and the dominant holder is `snapshot-bootstrap.ts`'s `BEGIN EXCLUSIVE` across a whole scope import (p50 2m55s Kilter). A 3×150ms ladder can't outlast minutes, `enqueue` is itself a writer blocked by that same lock, and the local `boardsesh_ticks` row a queue-only fallback would skip is read by the offline search's ticked filters and per-climb counts, so the climb would render UNTICKED offline for days. The evidence is 1 event / 90 days on the shipping release, and the issue explicitly permits deferring. #4314 owns it and now has `isDatabaseLockedError` plus the two dimensions to size it.

## Release Notes

## Test plan

- `vp test run --project offline-sync --reporter=agent` — 374 passed. New: dead-letter seam fires with the right reason and attempt count from both branches, does NOT fire on an acknowledged write, and **does not fire on a network error** (the regression that would silently re-break "an offline write never dead-letters for lack of a connection"); a throwing reporter neither aborts the drain nor prevents the dead-letter write committing; `enqueue` returns all three outcomes; `getOutboxSummary` on populated and empty tables; `parseQueueTimestamp` asserts the exact parsed epoch, not just non-null; `isDatabaseLockedError` against the real Sentry shapes including the U+0005 Android variant (built via `fromCharCode` so no formatter can eat the byte).
- `vp test run --project analytics --reporter=agent` — 23 passed.
- `vp run test:mobile` — 622 files / 5510 tests passed.
- `vp check` — clean for every file this PR touches (two unrelated docs files were already failing on `main`).
- `vp run check:mobile-bundle` — android bundle OK, run from a sibling worktree.

Not run: manual dev-client QA (forcing a dead letter, re-favoriting past it, relaunch, sign-out). `.boardsesh/qa-notes.md` in the worktree lists the flows.

### DIFF
--- a/packages/mobile/src/components/__tests__/offline-sync-bridge.test.tsx
+++ b/packages/mobile/src/components/__tests__/offline-sync-bridge.test.tsx
@@ -67,6 +67,13 @@ vi.mock('../../lib/error-reporting', () => ({
   reportError: vi.fn(),
 }));
 
+// The gauge itself is unit-tested in offline/__tests__/outbox-telemetry.test.ts;
+// what matters here is WHERE the bridge calls it from — both flag branches, once.
+const reportOutboxBacklogOnceMock = vi.hoisted(() => vi.fn(async (..._args: unknown[]) => {}));
+vi.mock('../../offline/outbox-telemetry', () => ({
+  reportOutboxBacklogOnce: reportOutboxBacklogOnceMock,
+}));
+
 // Stub the settings barrel so the static import graph never pulls
 // react-native-mmkv (its react-native Flow entry breaks Rolldown's scan).
 vi.mock('../../settings', () => ({
@@ -377,6 +384,43 @@ describe('OfflineSyncBridge — mid-session flag flips', () => {
 
     await waitFor(() => expect(startSyncSchedulerMock).toHaveBeenCalledTimes(1));
     expect(invalidateSpy).not.toHaveBeenCalled();
+  });
+});
+
+// Issue #4315: a kill-switched user's queued writes are exactly as stranded as a
+// flag-on user's — and sign-out deletes them all — so the launch gauge has to
+// run on both sides of the flag, and never on a signed-out launch.
+describe('OfflineSyncBridge — outbox backlog gauge', () => {
+  it('reads the backlog once with the flag ON', async () => {
+    render(<Harness flags={FLAG_ON} queryClient={makeQueryClient()} />);
+
+    await waitFor(() => expect(reportOutboxBacklogOnceMock).toHaveBeenCalledTimes(1));
+    expect(reportOutboxBacklogOnceMock).toHaveBeenCalledWith(fakeDb);
+  });
+
+  it('reads the backlog with the flag OFF too', async () => {
+    render(<Harness flags={FLAG_OFF} queryClient={makeQueryClient()} />);
+
+    await waitFor(() => expect(reportOutboxBacklogOnceMock).toHaveBeenCalledTimes(1));
+  });
+
+  it('does not read anything while signed out', async () => {
+    isAuthenticated = false;
+    render(<Harness flags={FLAG_ON} queryClient={makeQueryClient()} />);
+
+    await waitFor(() => expect(startBackgroundTrackingMock).toHaveBeenCalled());
+    expect(reportOutboxBacklogOnceMock).not.toHaveBeenCalled();
+  });
+
+  it('does not re-read when the flag flips mid-session', async () => {
+    const queryClient = makeQueryClient();
+    const { rerender } = render(<Harness flags={FLAG_ON} queryClient={queryClient} />);
+    await waitFor(() => expect(reportOutboxBacklogOnceMock).toHaveBeenCalledTimes(1));
+
+    rerender(<Harness flags={FLAG_OFF} queryClient={queryClient} />);
+    await waitFor(() => expect(getPendingCountMock).toHaveBeenCalled());
+
+    expect(reportOutboxBacklogOnceMock).toHaveBeenCalledTimes(1);
   });
 });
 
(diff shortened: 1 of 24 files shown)


marcodejongh:

PR #4330, opened 2026-08-12
feat(sync): report the deletions-coverage verdict on every sync cycle

marcodejongh:
Stacked on #4329 (base `feat/4315-offline-loss-telemetry`) for the `events.ts` / adapter overlap; rebases onto `main` the moment that merges.

`Offline Sync Coverage Reset Forced` has never fired once, which #4315 flags as a possible dead path. It is not: it is arithmetic. The guard only acts on verdict `stale`, which needs a `deletions-coverage` marker older than `DELETIONS_COVERAGE_MAX_AGE_MS` = (90 − 10) = 80 days. The marker first shipped in cb4210c5c on **2026-08-01** (#4048), so the earliest a device can hold an 80-day-old one is roughly **2026-10-20**. Zero is the predicted value. The path itself is live and tested end-to-end (the back-dated-marker integration test and the adapter→`track()` binding).

But the reset event alone is a **censored instrument**, and that part is a real gap. `enforceDeletionsCoverage` returns early on `coverageAt === null`, and the marker only exists after a *completed* deletions pull — so a device that can never finish one (a board stranded on the paged crawl, cf. #4313) stays `unknown` forever and emits nothing. The reset-only view therefore samples exactly the devices healthy enough not to be at risk.

Two smaller corrections while in here: the marker is stamped in **two** places, not one (after a completed deletions pull, and inside `resetUserDataForLostCoverage` as the post-wipe re-seed), and the reachability probe rejecting on a stale device is invisible today — it throws into the scheduler's catch and dies in a dev-only `console.warn`.

## Changes

- **`onCoverageEvaluated`** on `SyncOptions`, emitted immediately after `evaluateDeletionsCoverage` and **before** the `coverageAt === null || verdict !== 'stale'` early return, so `unknown` and `future` are first-class values rather than silence. Payload: `verdict`, `markerAgeDays` (null for `unknown`/`future`, floored otherwise), `outcome: 'evaluated' | 'reset' | 'probe_failed'`. Threaded through `sync-scheduler` exactly like `onCoverageReset`.
- The reachability probe is wrapped in try/catch that reports `probe_failed` and **rethrows unchanged**. The rethrow is the whole safety property: it is what leaves local data intact and defers to the next cycle. Swallowing it would let a stale device wipe user data without a verified connection — the precise catastrophe the probe exists to prevent. The existing "no reset on a rejecting probe" test stays green untouched, and the new one asserts the throw still propagates.
- On a successful reset, `outcome: 'reset'` is emitted **alongside** `onCoverageReset`, not instead of it: the reset event stays the dedicated "a wipe happened" signal, and the verdict stream carries the denominator that makes its rate readable.
- **`Offline Sync Coverage Evaluated`** in `SHARED_EVENTS`, bound in the mobile adapter on `startSyncScheduler` / `triggerSync` / `pullSync`. The dedupe lives in the **binding**, not the engine: the engine fires every cycle so the seam stays deterministic and testable, but `enforceDeletionsCoverage` runs at the top of every `pullSync` and the scheduler has no interval — it wakes on foreground and reconnect, so an un-deduped emit would repeat indefinitely on precisely the devices worth counting once. Keyed on the verdict, so a `fresh`→`stale` transition still reports.
- **`retention.ts` doc** records the reachability arithmetic (both stamp sites, the 80 days, the ~2026-10-20 floor), and `deletions-coverage.test.ts` pins `DELETIONS_COVERAGE_MAX_AGE_MS === 80 days` so a future retention or margin edit that moves that date fails CI rather than quietly moving a dashboard's expectations.

No behaviour change: the only new code paths are a report and a rethrow.

## Release Notes

## Test plan

- `vp test run --project offline-sync --reporter=agent` — 380 passed. New integration tests against real `node:sqlite`: **no marker at all** emits `verdict: 'unknown'`, `markerAgeDays: null` and still resets nothing (the case the reset-only view could never see, so the load-bearing one); a 45-day marker emits `fresh`; a future-dated marker emits `future`; the existing back-dated-marker test still emits `onCoverageReset` exactly once and now also emits `evaluated` then `reset` in that order; a stale marker whose probe rejects emits `probe_failed`, **still rethrows**, and leaves every local row and the marker untouched. Plus the pinned 80-day constant.
- `vp test run --project analytics --reporter=agent` — 23 passed.
- `vp run test:mobile` — 622 files / 5511 tests passed, including the adapter test asserting the verdict routes to `track` and never to Sentry, emits once for a repeated identical verdict, and again when the verdict changes.
- `vp check` — clean for every file this PR touches.

### DIFF
--- a/packages/mobile/src/components/__tests__/offline-sync-bridge.test.tsx
+++ b/packages/mobile/src/components/__tests__/offline-sync-bridge.test.tsx
@@ -74,13 +74,6 @@ vi.mock('../../offline/outbox-telemetry', () => ({
   reportOutboxBacklogOnce: reportOutboxBacklogOnceMock,
 }));
 
-// The gauge itself is unit-tested in offline/__tests__/outbox-telemetry.test.ts;
-// what matters here is WHERE the bridge calls it from — both flag branches, once.
-const reportOutboxBacklogOnceMock = vi.hoisted(() => vi.fn(async (..._args: unknown[]) => {}));
-vi.mock('../../offline/outbox-telemetry', () => ({
-  reportOutboxBacklogOnce: reportOutboxBacklogOnceMock,
-}));
-
 // Stub the settings barrel so the static import graph never pulls
 // react-native-mmkv (its react-native Flow entry breaks Rolldown's scan).
 vi.mock('../../settings', () => ({
--- a/packages/mobile/src/components/__tests__/offline-sync-bridge.test.tsx
+++ b/packages/mobile/src/components/__tests__/offline-sync-bridge.test.tsx
@@ -74,6 +74,13 @@ vi.mock('../../offline/outbox-telemetry', () => ({
   reportOutboxBacklogOnce: reportOutboxBacklogOnceMock,
 }));
 
+// The gauge itself is unit-tested in offline/__tests__/outbox-telemetry.test.ts;
+// what matters here is WHERE the bridge calls it from — both flag branches, once.
+const reportOutboxBacklogOnceMock = vi.hoisted(() => vi.fn(async (..._args: unknown[]) => {}));
+vi.mock('../../offline/outbox-telemetry', () => ({
+  reportOutboxBacklogOnce: reportOutboxBacklogOnceMock,
+}));
+
 // Stub the settings barrel so the static import graph never pulls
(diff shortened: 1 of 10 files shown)


marcodejongh:

Issue #4311, opened 2026-08-11
Offline download shows a static spinner through the whole snapshot phase — add byte/percent progress


marcodejongh:
During the snapshot bootstrap phase — manifest fetch, artifact download (103 MB gzip for Kilter layout 1), and import — the download UI shows only the static caption `mobile.offline.bootstrapping` ("Downloading board…") plus a spinner. No bytes, no percentage, no stage. For Kilter that stretch is minutes to tens of minutes (p50 2m55s, 37% >5 min), and it is indistinguishable from a hang.

- `packages/mobile/src/components/board-discovery/board-offline-state.ts:80-88` — bootstrap state maps to the static caption.
- `BoardManageRow.tsx:117-134` — the **paged** phase already shows a live climb count (`mobile.offline.downloadingCount` driven by `SyncProgress.currentTableProcessed`); the snapshot phase shows nothing comparable.
- `BoardOfflineToggle.tsx:27-29` — bare `ActivityIndicator` while `downloadState === 'downloading'`.

## What the user should get

A real progress readout during the expensive phase: downloaded bytes / total bytes (the manifest knows the artifact size) or a percentage, and ideally a stage hint ("Downloading… 42 MB of 103 MB", then "Unpacking…" during import). `expo-file-system`'s DownloadResumable exposes a progress callback; the manifest entry carries the artifact size.

We can't currently measure abandonment (no download-started event — see the funnel-instrumentation issue), but a perceived hang on a multi-minute download is the most plausible driver of users giving up on exactly the board (Kilter) most people try first.

Part of the offline-mode epic: #4319.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01UT6VP1sWqk6bY9mUgyeS2U

marcodejongh:
Three stacked draft PRs, in order. All local validation green; #4336 carries a device-QA gate before it goes ready.

- #4335 — `feat(backend): publish the uncompressed artifact size in the snapshot manifest` (base `main`). Optional additive `uncompressedBytes`, no `formatVersion` bump. No client behaviour change.
- #4336 — `feat(sync): show real download progress instead of a static spinner` (base #4335). **Closes this issue.** Staged `manifest → download → import` frames, wire-scale byte/percent captions, a reserved-height bar, the `offline-download-progress` kill switch (default ON), and an exact disk-space precheck as a free win.
[...]

PR #4335, opened 2026-08-12
feat(backend): publish the uncompressed artifact size in the snapshot manifest

marcodejongh:
First of three stacked PRs for the offline download-progress + funnel work (#4311, #4316). This one is pure backend/schema plumbing with no client behaviour change.

## Why

The manifest carries the artifact's size exactly once, as `bytes` — the **stored** object size. Under `--gzip` (which the whole fleet reads today, `board-snapshots/v1-gzip`) that is ~103 MB for `kilter:1`, while the SQLite file that actually lands on disk is ~271 MB.

Any client that wants a progress denominator for a byte counter, or an honest free-disk-space precheck, currently has no way to know the decoded size. The export job already computes it — `rawBuffer.length`, logged as `rawBytes` and then thrown away.

## Changes

- Optional `uncompressedBytes` on `SnapshotManifestEntry`, populated from the pre-gzip buffer length at both export call sites (dry-run and upload).
- **No `formatVersion` bump.** The field is additive and optional: an old client ignores the key, a new client reading a pre-change manifest gets `undefined` and falls back. Present-but-corrupt (fractional, negative, non-numeric) still rejects the whole manifest, because a nonsense decoded size would drive a nonsense progress bar and a nonsense disk-space check.
- The merge path carries entries that lack the field through untouched, so a filtered run can't invent one for a board it didn't rebuild.
- `docs/board-snapshots.md` gains a short section on the two sizes and who reads which.

## The scale rule this establishes

`bytes` stays the **only** size a user ever sees. The enable-confirm dialog's quote (`estimateScopeDownload`) and — in PR2 — the download progress row both read it, so the two can never disagree. `uncompressedBytes` is internals-only: a progress denominator on platforms whose downloader counts decoded bytes, and the disk-space precheck. It never reaches the UI.

## Rollout note

The field only appears on entries built by an export run **after** this merges. The nightly runs at 07:15 UTC; dispatching `.github/workflows/export-board-snapshots.yml` manually fills it in sooner. Until then PR2's Android percentage degrades gracefully to an indeterminate byte counter (iOS is unaffected — it uses the platform's own reported total).

## Release Notes

## Test plan

- `vp test run --project backend --reporter=agent packages/backend/src/__tests__/snapshot-export-run.test.ts` — 27 passed. New cases: `--gzip` publishes `uncompressedBytes > bytes` and `bytes` equals the gzipped upload length; an identity run reports them equal; a merged previous-manifest entry that predates the field rides through with `uncompressedBytes` still `undefined`.
- `vp test run --project offline-sync --reporter=agent` — 341 passed. New cases in `snapshot-manifest.test.ts`: an entry with no `uncompressedBytes` parses; a non-negative integer parses; fractional / negative / string values reject the manifest.
- `vp run typecheck` — clean.

### DIFF
--- a/docs/board-snapshots.md
+++ b/docs/board-snapshots.md
@@ -167,6 +167,25 @@ layout artifact but intentionally outside the enabled size.
   **permanent miss for that run, no bootstrap attempt burned** — the scope falls back to the always-correct
   paged crawl, and the next nightly export rebuilds the artifact at the new schema.
 
+### The two artifact sizes in a manifest entry
+
+A manifest entry carries the artifact's size twice, and the two diverge once `--gzip` ships:
+
+| Field               | What it is                                    | Who reads it                                                                                                                                                              |
+| ------------------- | --------------------------------------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `bytes`             | Stored object size — what comes down the wire | **Every user-facing number.** The My Boards confirm-dialog size quote (`estimateScopeDownload`) and the download progress row both use it, so the two can never disagree. |
+| `uncompressedBytes` | Decoded size — the SQLite file that hits disk | Internals only: the progress denominator on downloaders that write decoded bytes with no usable reported total, and the exact free-disk-space precheck. Never rendered.   |
+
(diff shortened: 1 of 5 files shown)


marcodejongh:

PR #4346, opened 2026-08-12
feat(backend): publish per-layout board_climb_grades snapshot artifacts

marcodejongh:
Boardsesh grades are the one per-board table the snapshot never carried, and the scope-completion gate waits on **every** per-board table — so after a Kilter or Tension artifact lands, the scope still pages hundreds of serial authenticated GraphQL requests for `board_climb_grades` before it can serve an offline search, each re-running a correlated `EXISTS` against 375k climbs.

The live manifest holds the natural experiment. `board_climb_grades` is populated only for `CROWD_MEAN_BOARDS`, which deliberately excludes MoonBoard (`packages/db/src/queries/grade-model/constants.ts`, `docs/boardsesh-grade.md`):

| entry | artifact | climbs | grade rows | p50 |
|---|---|---|---|---|
| `moonboard:2` | 20.0 MB | 93,939 | **0** | **14.5s** |
| `tension:9` | 17.1 MB | 72,051 | ~79k | **1m27s** |

Same bytes, same import shape, roughly six times the duration. The only structural difference is the grades crawl.

This PR is the **export side only**. Nothing reads `grades` yet — that is the stacked follow-up.

## Changes

- **A second, standalone artifact per layout** at `<prefix>/<board>/<layout>/<builtAt>-grades.db`: the `board_climb_grades` DDL from the shared MIGRATIONS, that layout's grade rows, and its own **one-row** `snapshot_meta`. Scoped by the same correlated `EXISTS` over `board_climbs` that the `syncClimbGrades` resolver uses, plus the stability window.
- **One transaction.** Both files are written inside the same `REPEATABLE READ` transaction as the whole-layout artifact, so the grade rows and the climbs they hang off come from one database snapshot and their watermarks are mutually consistent.
- **The whole-layout artifact is untouched** — same DDL, same two tables, same two `snapshot_meta` rows, pinned by a test. Every shipped binary verifies that file against its own two-table list and a mismatch is a **counted** import failure; two of those settle a scope onto the paged crawl, so widening its meta would break the fleet for the entire 14-day prune grace.
- **Additive manifest.** Optional `grades` block on the layout's existing entry, `formatVersion` still `1`, validated only when present (and rejected when malformed — a half-parsed grades block would be trusted far enough to stamp a checkpoint over rows that never arrived). Grades artifacts are deliberately **not** elements of `entries`: `findSnapshotEntry` first-matches on `(boardType, layoutId)`, so a sibling entry could be imported by an older client as if it were the whole layout.
- **Per-table cursor column.** `TABLE_CONFIGS[...].cursorColumn` now declares the timestamp half of each table's keyset cursor once. Grades cursor on `computed_at` — they have no `updated_at` column — and the export reads the column from the shared config so it, the resolvers, and the client import cannot disagree about which column a watermark covers.
- **Prune safety.** `pruneStaleArtifacts`' referenced-key set now includes every surviving entry's `grades.key`. Without it the 14-day grace window would start deleting live grades files out from under clients holding the current manifest.
- **Gzip pass only.** The nightly runs twice — the identity `v1` prefix kept as a rollback target, and `v1-gzip` where the fleet actually points. Publishing grades only in the gzip pass makes a rollback to `v1` exactly today's behaviour with no deploy, which is the kill switch for the whole path.
- `docs/board-snapshots.md` updated: storage layout, prune, artifact format, a new "Grades artifact" section, and the kill switch.

### Deviation from the plan

The plan's dead-index trim (`idx_stats_lookup`, an exact duplicate of the stats `PRIMARY KEY`, plus `idx_stats_difficulty`) is **not included**. It was gated on two proofs — a `--dry-run` byte delta against the dev DB and an old-engine import test — and it is the only PR-2 change a shipped binary can feel. Filed as a follow-up rather than landed unmeasured.

### Post-merge reconciliation

Rebased onto `main` after #4335 (`uncompressedBytes` in the manifest entry + export) landed — both features are kept: entries carry `uncompressedBytes` **and** the optional `grades` block. A further reconciliation with #4345 is expected on the client side of this stack: #4345 replaces the `downloadedPaths` set that #4347 registers the grades file in with its `artifactImported` / `releaseArtifact` seam, and grades filenames do not match snapshot-source's `artifactNamePrefix` regex, so its supersede sweep will not recognise them. Nothing in this backend-only PR changes as a result.

## Release Notes

## Test plan

- `vp test run --project backend packages/backend/src/__tests__/snapshot-export-golden.test.ts` — 11 passed. New: exported grade rows are byte-identical to what a real `syncClimbGrades` pull writes through the real client upsert path; the whole-layout artifact still has exactly `board_climbs` + `board_climb_stats` and two meta rows and no grades table; the grades file carries only its own table and one meta row; the watermark cursors on `computed_at`; a grade inside the stability window is excluded from both the rows and the watermark; a grade for another layout never lands; a layout with no grade rows publishes nothing; omitting `gradesFilePath` skips grades entirely.
- `vp test run --project backend packages/backend/src/__tests__/snapshot-export-run.test.ts` — 30 passed. New: a gzip run publishes the grades object and references it from the entry (still one entry per board+layout, `formatVersion` 1); a layout with no grades omits the block and uploads nothing; the identity pass publishes no grades at all; the prune never deletes a just-published grades key; a superseded grades key **is** pruned past the grace window; a filtered run preserves another board's previously-published grades block.
- `vp test run --project offline-sync` — 344 passed, including a **frozen copy of the shipped v1 manifest validator** checked in as a fixture that must keep accepting a grades-carrying manifest, plus malformed-grades rejection and the "grades stay out of `entries`" shape guard.
- `vp run typecheck` (all packages), `vp lint` — clean in the touched packages.

Only the two changed backend files were run, per the shared-worker-DB rule; CI covers the rest of the backend suite. No nightly export has been dispatched yet — that is the gate for the stacked client PR leaving draft.

### DIFF
--- a/docs/board-snapshots.md
+++ b/docs/board-snapshots.md
@@ -108,6 +108,13 @@ avoid dropping data on a broken read:
   decompressed on disk (see Rollout plan). No explicit `cacheControl` is passed to `uploadToS3`, so
   artifacts get the storage layer's default: `public, max-age=31536000, immutable`. Content-addressed by
   build timestamp — safe to cache forever, a new build gets a new key.
+- Grades artifacts: `<keyPrefix>/<boardType>/<layoutId>/<builtAt>-grades.db`, same `Content-Type`, always
+  gzip, and **published only in the `v1-gzip` pass** (issue #4310). Same content-addressed immutability as
+  the whole-layout artifact. Referenced from the layout's manifest entry through the optional `grades`
+  block, never as its own `entries` element — `findSnapshotEntry` first-matches on `(boardType, layoutId)`,
+  so a sibling entry could be imported by an older client as if it carried the whole layout, stamping
+  checkpoints past rows it never imported. Layouts with no grade rows (every MoonBoard layout: MoonBoard is
+  deliberately outside `CROWD_MEAN_BOARDS`, see `docs/boardsesh-grade.md`) publish nothing.
 - Manifest: `<keyPrefix>/manifest.json`, `Content-Type: application/json`,
   `Cache-Control: public, max-age=300`. Mutable and cheap to refetch, written last so it's the only object
   in the whole scheme that changes in place. Each prefix has its own manifest.
@@ -122,7 +129,8 @@ Artifacts superseded by a newer build for the same `(boardType, layoutId)` are d
 
 An object is eligible for deletion when it's under the prefix the run targets (`board-snapshots/v1/` for
 the identity pass, `board-snapshots/v1-gzip/` for the gzip pass) and NOT referenced by the manifest
-just written, **and** its `lastModified` is older than a **14-day grace window**
+just written — where "referenced" includes every surviving entry's `grades.key`, which is not in `entries`
+and would otherwise be swept out from under live clients — **and** its `lastModified` is older than a **14-day grace window**
 (`PRUNE_GRACE_MS`). The grace window exists because the manifest is CDN-cached for up to 5 minutes and a
 client may hold a fetched manifest (with a now-superseded artifact URL) far longer than that before it
 actually starts the download. Pruning is defensive by design: any failure (per-object or the whole scan) is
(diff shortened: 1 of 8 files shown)


marcodejongh:

marcodejongh:
Rebased onto `main` after #4335 (uncompressed artifact size in the manifest) merged, which had left this PR conflicting — both PRs touch `snapshot-manifest.ts`, `table-config.ts`, `export-board-snapshots.ts` and `docs/board-snapshots.md`. Both features are kept whole: a manifest entry now carries `uncompressedBytes` **and** the optional `grades` block, and `isManifestEntry` validates both (the fractional/negative `uncompressedBytes` rejection and the malformed-grades rejection sit side by side). `buildManifestEntry` takes `uncompressedBytes` on the whole-layout upload plus the optional grades upload, and both the dry-run and the upload call sites pass all three; the grades PR's `canBuildPublicUrl` hoist absorbed main's inline `isS3Configured()` check rather than duplicating it. In `table-config.ts` the grades PR's `UPDATED_AT_CURSOR` constant moved onto main's new `TABLE_SYNC_DEFINITIONS` shape (#4350's invalidate-keys refactor), so `cursorColumn` is still declared once and the derived `TABLE_CONFIGS` is unchanged. Docs keep both new sections. Re-validated: `vp test run --project offline-sync` 357 passed, `snapshot-export-run.test.ts` 33 passed, `snapshot-export-golden.test.ts` 11 passed, `vp run typecheck` clean. Force-pushed as `f079fc242`.

Issue #3858, opened 2026-07-24
mobile: expo-sqlite "database is locked" — offline writes & local reads fail under concurrent transactions (no WAL / busy_timeout on the main connection)


marcodejongh:
## Symptom

On iOS (App Store build `com.boardsesh.app@2.2.2+3`), offline-sync SQLite operations intermittently throw `database is locked` / `database table is locked`. Four Sentry issues, one root cause — **72 events / 23 users in 14d**, all `handled:true`, all `environment=production`, still firing (last seen 2026-07-23):

| Sentry | Failing call | Cause | kind / source | 14d | users |
|---|---|---|---|---|---|
| [BOARDSESH-AB](https://boardsesh.sentry.io/issues/7612463785/) | `execAsync` | database is locked | tick-local-write / offline-sync | 32 | 23 |
| [BOARDSESH-A9](https://boardsesh.sentry.io/issues/7611674576/) | `finalizeAsync` | code 5: database is locked | query / react-query | 29 | 22 |
| [BOARDSESH-AX](https://boardsesh.sentry.io/issues/7619344571/) | `prepareAsync` | code 5: database is locked | tick-local-write / offline-sync | 9 | 7 |
| [BOARDSESH-AC](https://boardsesh.sentry.io/issues/7612596226/) | `finalizeAsync` | code 6: database **table** is locked | vacuum / offline-sync | 2 | 2 |

iOS 18.7.8 and 26.5.2 both affected — not OS-specific. Hermes strips JS frames, but the `kind`/`source` scope tags pin every throw site.

## Root cause (verified on `main`)

The offline-sync engine runs **many concurrent connections against one `boardsesh.db` with no WAL and `busy_timeout = 0`.**

- `initializeDatabase` (`packages/mobile/src/db/connection.ts:174`, the `SQLiteProvider` `onInit` callback) **never sets `PRAGMA journal_mode = WAL` or `PRAGMA busy_timeout`.** So the main connection defaults to rollback journaling (a writer blocks all readers and other writers) and `busy_timeout = 0` — a contending statement fails *immediately* instead of waiting for the lock to clear.
[...]

PR #4357, opened 2026-08-12
fix(sync): survive a locked WAL checkpoint, report stalled downloads

marcodejongh:
Refs #4314 #4310

## The bug

Sentry [BOARDSESH-D7](https://boardsesh.sentry.io/issues/BOARDSESH-D7): `FunctionCallException: Calling the 'finalizeAsync' function has failed → Caused by: SQLiteErrorException: Error code 6: database table is locked`.

Two users, iOS, `2.3.1+2`. The Sentry tags say `source: offline-sync`, `kind: vacuum` — so despite the shape of the message this is **not** the snapshot-import path. It is `vacuumDatabase()`.

### Root cause

`packages/shared/offline-sync/src/db/vacuum.ts:72` ran `PRAGMA wal_checkpoint(TRUNCATE)` on the app's **main** SQLite connection — the one the sync engine, every `useSQLiteContext()` screen, and this call all share.

SQLite's `sqlite3BtreeCheckpoint()` returns **SQLITE_LOCKED (6)** outright when the *calling* connection already has an open b-tree transaction:

```c
if( pBt->inTransaction!=TRANS_NONE ){
  rc = SQLITE_LOCKED;
}else{
  rc = sqlite3PagerCheckpoint(...);
}
```

On a shared handle that is true for as long as any other statement is in flight. Removing a board runs `removeOfflineBoard` → `compactOfflineDatabase` → `vacuumDatabase` while My Boards is re-reading `downloadedScopeKeys`, and the checkpoint loses the race.

Code 6 is **not** code 5: it is a conflict *inside* one connection, so no `busy_timeout` can wait it out — which is why the #3858 / #4104 / #4324 busy-timeout and retry work never touched it.

Reproduced against real SQLite (now pinned as a test):

```
checkpoint, no open txn:    { busy: 0, log: 0, checkpointed: 0 }
checkpoint w/ open read txn: THREW -> database table is locked | errcode= 6
```

The stack says `finalizeAsync` because expo-sqlite's `getFirstAsync` wraps execution in `try { … } finally { await statement.finalizeAsync(); }`, and `sqlite3_finalize()` re-returns the error code of the statement's last failed evaluation — so the `finally` throws the *same* error and replaces the original. **Every** code-5/code-6 failure on any expo-sqlite prepared-statement call is reported as `finalizeAsync`; the frame names the cleanup, never the statement.

### The fix

`vacuumDatabase` already modelled a blocked truncation as `resolves false` — for the `busy = 1` row. SQLite says "blocked" in two shapes and only one was handled. The thrown SQLITE_LOCKED/SQLITE_BUSY shape now resolves `false` too, after a short bounded retry (the blocker is a list row that finishes in milliseconds, so a retry usually turns a `false` into a `true` and the storage figure actually moves). Anything that is **not** lock contention still throws — a malformed database must never be laundered into "the file just didn't shrink". The remove-board reporter drops to `warning`: the VACUUM landed, the rows are gone.

## The telemetry gap

PostHog for the affected device: six `Offline Board Download Started`, **zero** Completed, **zero** Failed.

The snapshot bootstrap phase bailed **silently** whenever the cycle was torn down — a sign-out, *any* board removal (`beginLocalPurge()`), or the app backgrounding. Started fired, the cycle was cut, and no terminal event ever arrived, so an interrupted 100 MB download was structurally invisible.

Those three bail-outs now report through the existing `onSnapshotBootstrapError` seam with `aborted: true` and a classified `reason`. They stay **out of Sentry** — a pocketed phone is not a defect — but they close the funnel, so every Started has a terminal event. Consumers computing a failure *rate* filter `aborted = false`; that is documented on the event.

Every failure report also gained `reason`, a closed low-cardinality union (`database-locked`, `artifact-invalid`, `schema-stale`, `watermark-regression`, `network`, …). `stage` and `reason` are now Sentry **tags**, not extras — chasing this issue through `extra.stage` is exactly what made "which phase is this?" unanswerable from the issue page.

## Not in this PR

Removing **any** board calls `beginLocalPurge()`, which bumps a **global** wipe epoch that every scope's cycle checks. Toggling one board off therefore aborts the in-flight download of every *other* board. That is what actually cost the affected device its six downloads — eight boards toggled in 45 seconds, each toggle killing the running 102 MB Kilter transfer. Scoping the purge per-board touches ~15 epoch checks and has real data-integrity implications (an in-flight page must not land after a delete), so it needs its own change. With `reason: 'aborted-wipe'` now on the funnel it is finally measurable.

## Testing

- `vp test run --reporter=agent --project offline-sync`
- `vp run test:mobile`
- `vp run typecheck:mobile`

New coverage: the real-SQLite error shape (regression pin — if SQLite or the driver stops reporting this as a lock, the classifier stops recognising it and the throw escapes again), the retry/return-false/rethrow logic, every `reason` bucket, the engine's abort reports, and the adapter's track-but-don't-Sentry behaviour.

## Release Notes

Removing a downloaded board could fail to hand its space back while the app was busy reading, and stop with a database error instead of just leaving the file a little larger. It now waits for the app to finish and reclaims the space properly.

Downloads that get interrupted — by locking your phone, or by removing another board mid-download — are now recorded instead of vanishing, so we can see how often a board download stops short and why.

### DIFF
--- a/packages/mobile/src/offline/__tests__/offline-sync-adapter.test.ts
+++ b/packages/mobile/src/offline/__tests__/offline-sync-adapter.test.ts
@@ -464,12 +464,14 @@ describe('snapshot-bootstrap bindings', () => {
       attempt: 1,
       cause: null,
       expected: false,
+      reason: 'unknown',
+      aborted: false,
     });
 
     expect(reportHandledError).toHaveBeenCalledWith(
       expect.objectContaining({ message: expect.stringContaining('kilter:1:5') }),
       expect.objectContaining({
-        tags: { source: 'offline-sync', kind: 'snapshot-bootstrap' },
+        tags: { source: 'offline-sync', kind: 'snapshot-bootstrap', stage: 'download', reason: 'unknown' },
         extra: expect.objectContaining({ scopeKey: 'kilter:1:5', stage: 'download', attempt: 1 }),
       }),
     );
@@ -497,6 +499,8 @@ describe('snapshot-bootstrap bindings', () => {
       attempt: 0,
       cause,
       expected: true,
+      reason: 'network',
+      aborted: false,
     });
 
     const [reported, context] = reportHandledError.mock.calls[0] as [
@@ -505,7 +509,13 @@ describe('snapshot-bootstrap bindings', () => {
     ];
     expect(reported.cause).toBe(cause);
     expect(context.level).toBe('warning');
-    expect(context.tags).toEqual({ source: 'offline-sync', kind: 'snapshot-bootstrap', expected_offline: true });
+    expect(context.tags).toEqual({
+      source: 'offline-sync',
+      kind: 'snapshot-bootstrap',
+      stage: 'manifest',
+      reason: 'network',
+      expected_offline: true,
+    });
     expect(context.extra).toEqual(
       expect.objectContaining({
         scopeKey: 'kilter:1:5',
@@ -847,6 +857,8 @@ describe('snapshot-bootstrap bindings', () => {
       attempt: 2,
       cause: new Error('The request timed out.'),
       expected: true,
+      reason: 'network',
+      aborted: false,
     });
 
     expect(trackMock).toHaveBeenCalledWith(SHARED_EVENTS.OfflineBoardDownloadFailed, {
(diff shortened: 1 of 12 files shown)


marcodejongh:
Thanks — on the `attempt: 0` point I kept the value and documented it instead, because 0 is the field's established meaning rather than a placeholder.

Every report site spells "nothing was burned" as zero: `reportSettledFailure` sends `settled.persisted ? burned : 0`, and the `SnapshotSchemaStaleError` arm sends a literal `0` with the comment "Permanent miss for this run, no burn". A teardown spends no retry budget either, so passing the scope's current attempt index would claim it had used up an attempt it still has — which is the more damaging misreading of the two, since that number gates whether the fast path is retried at all.

Comment added at the call site so the next reader does not have to re-derive it (cff9d4e1f).

PR #4401, opened 2026-08-13
perf(ci): cut the PR critical path from ~433s to ~230s

marcodejongh:
## Summary

Cuts the PR critical path from a measured **433s to a projected ~230s**. Every number here comes from `gh run view` over the 21 most recent `ci.yml` runs plus real runner logs — not estimates.

A benchmarking pass measured each CI layer first. Two things it corrected before any code changed:

- **The pole is `mobile-bundle`, not `test-default`.** The 366s `test-default` mean mixes push runs (full 1380-file suite, 558s) with PR runs (`--changed`, 269s). Rebuilt from run-level offsets, `mobile-bundle` owns the PR critical path in 8 of 14 runs, `test-default` in 4, `typecheck` in 1, `test-backend` in 1.
- **`typecheck`'s "cache-driven affected" step reports `0/54 cache hit (0%)`** in 5 of 5 sampled runs. It was never cache-driven.

### What changed

| # | Change | CP saving |
|---|---|---|
| 1 | Delete the `setup` job + nine `Download built packages` steps | −69s |
| 2 | Persist Metro's transform cache | −62s |
| 3 | Move the nine `check:mobile-*` guards into `lint` | −22s |
| 4 | Stop `--clear`-ing Metro's cache in the Expo-web *check* | −34s |
| 5 | Drop `changes`'s checkout (fail-closed truncation guard) | −6s |
| 6 | Drop 8 duplicate `tsc --noEmit` tasks | −12s |
| 7 | Shard `test-default` 3x main / 2x PR | tail + main |
| 8 | Health intervals 10s→2s; delete test-ocr's apt install | runner-s |

**The `setup` job was distributing nothing.** It spent 58s of serialized latency uploading `packages/db/dist` + `packages/shared-schema/dist` so nine jobs could download them. Both packages declare `main`, `types` and every `exports` entry pointing at `./src/*.ts`; neither dist dir exists in a working checkout; the only references to those paths anywhere outside `node_modules` were the two upload lines themselves. Nine jobs now start at ~20s instead of ~93s. `vp run build:db` stays on the required path via `typecheck:db`.

**Metro's cache was thrown away every run.** `@expo/metro-config` hard-codes the store at `os.tmpdir()/metro-cache`. Measured on the exact CI command: **cold 178.4s / 649.6 CPU-s vs warm 89.6s / 104.0 CPU-s** — 84% of the CPU is babel output Metro already knows how to cache. `TMPDIR` now points at a cached dir under `runner.temp`, deliberately not inside the workspace (`metro.config.js` watches the monorepo root). The Expo-web check additionally passed `--clear`, which `rm -rf`s all 256 shards — measured **101.4s cold vs 6.9s warm** for the identical 4311-module export.

### Guarding the `--clear` removal

`--clear` was bought by a real incident (stale `EXPO_PUBLIC_*` / base-URL inlining reaching a shipped artifact). The deploy path keeps it. Only the *check* path — which sets none of that env and throws its output away — opts out, and the property that replaces the flag is a **cache root scoped by a hash of the inlined env**, so a baked-env change lands in a different directory instead of hitting a stale shard. Default with no env var set is today's behaviour: `--clear` on.

### A contract test that pinned an address instead of a property

`scripts/mobile-ci-env-parity.test.ts` asserted the patch and fingerprint guards live *in the `mobile-bundle` job, by name*. Relocating them reddened it while its semantic intent — a patches-only PR reaches both sentinels — still held. It now asserts the property: whichever job hosts a guard must carry the `mobile` filter in its **job-level** gate.

My first version of that fix was inert. A loose `/if:.*mobile/` matched a *step-level* `if:` inside `lint`, so it passed even with the job gate narrowed. Caught by mutation-testing, then fixed to read the 4-space-indent job gate specifically.

## Release Notes

## Test plan

**Test quality is the constraint, not the variable.** A benchmark agent built the invariants first: 45 vitest projects / 1630 spec files / 19275 tests, 20 jobs, 21 skip sites. Nothing here deletes, skips, retries, reselects or re-times a test.

The repo's specific gaming risks, all checked against the diff:
- **All 21 skip sites are environment-conditional**, and `ci-status` treats `skipped` as pass — so dropping a service container silently turns hard failures green. No service container was dropped; health-check *tolerance windows* are byte-identical (50s/50s/50s/100s before and after), only the polling frequency changed.
- **`INFRA_PROJECTS` is a subtraction set** — one added name deletes a whole project from CI while its own unit test stays green. `scripts/ci-test-default-projects.ts` is untouched; still emits 42 `--project=` flags.
- No `--passWithNoTests`, no changed `--changed` base, no lowered `fetch-depth`, no `isolate:false`, no `retry:`/`repeats:`/`bail:`, no new skip sites, no third `continue-on-error`.
- Every `if:` and path filter that moved got **wider**, never narrower. `lint`'s gate is now the union of its own and `mobile-bundle`'s.

Verified:
- `vp run typecheck` — exit 0; task graph 54 → 46 (the 8 duplicates)
- `vp check` (full repo) — exit 0
- 3 CI contract tests green, incl. the rewritten parity assertion
- **Mutation-tested the rewritten guard 3 ways** — narrowed job gate, deleted guard step, narrowed step gate. All three caught; the guard fails when it should.
- Job set diffed against `main`: only `setup` removed, no job added, `ci-status.needs` gaps identical to main (`large-files` / `test-report`, both pre-existing)

The real proof is this PR's own CI run — that's the measurement, and I'll compare it against the baseline once it lands.

## Deferred

Dropping `--changed` to run the full suite sharded 4–6 ways is a coverage **increase** (~3.4x more specs per PR, closing the `--changed` blind spot for fs-reading tests). Held back only because 4–6 concurrent shards land on a wave already running ~8 heavy jobs, near the 20-job concurrency ceiling. Worth its own PR.

Rejected as coverage trade-offs: jsdom → happy-dom (482 web specs lose CSSOM/Selection/form-validation fidelity that MUI+Emotion exercise) and moving 159 DOM-free specs to the node environment.

https://claude.ai/code/session_01HX6Gz2Ew9UduUhTP3W6dC7

---

## Measured result (run 31657115300)

**Critical path 433s → 349s, −84s (−19%)** — and this run had a **cold Metro cache**, so the two cache changes contributed nothing to it yet.

```
  END  START    DUR   JOB
 349s   346s     3s   ci-status
 344s    26s   318s   mobile-bundle          <- the pole, cold cache
 220s    75s   145s   test-location-sync-integration
 199s     5s   194s   typecheck
 100s     5s    95s   test-backend (1, 1)
  97s    19s    78s   test-default (2, 2)
  89s     6s    83s   lint  (+ the 9 relocated mobile guards)
  77s     7s    70s   test-default (1, 2)
  72s    13s    59s   test-ocr
   3s     0s     3s   changes
```

Heavy jobs now start at **5–19s instead of ~93s**. `changes` went 15s → 3s. `test-ocr` 84s → 59s with the apt install gone.

The cache saves are gated to `main`, so they populate on the first push after merge; a PR restores from the default branch's entry via the restore-key prefix. Expected once warm: `mobile-bundle` ~318s → ~140s, which hands the pole to `typecheck` at 199s for a **~215s** critical path.

Verified on the artifacts: the two shards partition cleanly (**0 overlap**), and `test-report` merged both.

## A coverage hole this PR demonstrated on itself

`test-default` ran exactly **one** spec file on this PR — `mobile-ci-env-parity.test.ts`, 34 tests — and only because I edited that file directly. `scripts/__tests__/ci-location-sync-workflow.test.ts` reads `ci.yml` through `readFileSync`, so `--changed`'s module graph never linked my ci.yml rewrite to it, and **it did not run**.

A PR that rewrites CI ran 34 tests. That is the pre-existing `--changed` blind spot for fs-reading tests, not something this PR introduced — but this PR is an unusually clean demonstration of it. All three CI contract tests were run locally and are green; the mutation testing described above was done by hand for the same reason.

This is the argument for the deferred full-suite-sharding change.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -56,21 +56,36 @@ jobs:
       # Defaults false off PRs (the OTA bot legitimately rewrites it on push).
       changelogData: ${{ github.event_name == 'pull_request' && steps.filter.outputs.changelogData || 'false' }}
       dbMigrations: ${{ github.event_name != 'pull_request' && 'true' || steps.filter.outputs.dbMigrations }}
-      # True if any downstream job's gate would fire. setup gates on this so
-      # docs-only / firmware-only PRs skip the bun-install + build:db pass
-      # nothing downstream would consume. Must mirror the union of every
-      # downstream `if:` — sharedDeps/sharedSchema in particular cover
-      # non-JS files (SQL migrations under packages/db, .graphql under
-      # shared-schema) that wouldn't trigger anyJs but still need setup so
-      # test-backend / mobile-bundle / codegen-drift can run.
+      # True if any downstream job's gate would fire. This used to gate the
+      # `setup` job (bun install + build:db + a `built-packages` artifact every
+      # heavy job downloaded); that job was deleted because @boardsesh/db and
+      # @boardsesh/shared-schema resolve through `src/*.ts` in their exports
+      # maps — nothing in the repo ever consumed those dist dirs — and it cost
+      # ~69s of serialized latency in front of every gated job. The output is
+      # kept (no consumer today) because it is the one place that records the
+      # union of every downstream gate, and scripts/__tests__/
+      # ci-location-sync-workflow.test.ts asserts on its terms; keep it in sync
+      # with the job `if:`s below so a future job can gate on it again.
       runAny: ${{ github.event_name != 'pull_request' || steps.filter.outputs.anyJs == 'true' || steps.filter.outputs.web == 'true' || steps.filter.outputs.i18nCatalogs == 'true' || steps.filter.outputs.sharedSchema == 'true' || steps.filter.outputs.sharedDeps == 'true' || steps.filter.outputs.backend == 'true' || steps.filter.outputs.mobile == 'true' || steps.filter.outputs.ocr == 'true' || steps.filter.outputs.locationSync == 'true' || steps.filter.outputs.rootCi == 'true' }}
       # NOTE: deployConfig, storeMetadata, and mobileLocales are deliberately
       # NOT part of runAny — the deploy-config and listing-guards jobs below
-      # need [changes] only (no bun-install/build:db setup pass), so a PR
-      # touching only their files doesn't pay for a `setup` run nothing else
-      # would consume.
+      # gate on their own filters, so a PR touching only their files runs
+      # nothing else.
     steps:
-      - uses: actions/checkout@v6
+      # No actions/checkout here on purpose: on `pull_request` dorny/paths-filter
+      # reads the changed-file list from the REST API (no working tree needed),
+      # and on `push` the filter step is skipped outright and every output above
+      # hardcodes `true`. The checkout was 8-58s at the very head of the run.
+      #
+      # The one thing the API costs us is a ceiling: `GET /pulls/{n}/files` tops
+      # out at 3000 files, and a truncated list would evaluate a filter to
+      # `false` and silently skip a job that should have run — the failure mode
+      # this whole `changes` job exists to avoid. Fail closed instead, and say
+      # what to do about it. (Needs `pull-requests: read`, subsumed by the
+      # workflow-level `pull-requests: write`.)
+      - name: Guard against paths-filter API truncation
+        if: github.event_name == 'pull_request' && github.event.pull_request.changed_files > 2900
+        run: echo "::error::PR exceeds the paths-filter API window; re-add actions/checkout to the changes job" && exit 1
       - id: filter
         if: github.event_name == 'pull_request'
         uses: dorny/paths-filter@v3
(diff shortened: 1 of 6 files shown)


marcodejongh:
Thanks — addressed. Verified each against the files rather than taking them as read:

**1. Type-coverage guard — agreed, fixed.** My own paired QA reviewer raised this independently, which made it the one worth acting on. `scripts/__tests__/typecheck-graph-contract.test.ts` now asserts, per package: `build` is plain `tsc` and names no bundler, no `tsconfig.build.json` exists, and the entry still reaches the aggregate. Mutation-tested both ways (swap backend to `tsup`; drop `build:moonboard-sync`) — each reds exactly one assertion.

**2. Mobile guards now run on any `anyJs` PR — real, and deliberate.** Adding a per-step `if:` would re-introduce the narrowing risk this PR exists to avoid, and `mobile-ci-env-parity.test.ts` now explicitly asserts those guard steps carry *no* narrowing step-level `if:`. The cost is invisible on the critical path: `lint` ends at 89s against a 344s pole. Widening was the safe direction, so I took it.

**3. `KEEP_METRO_CACHE=1` init — leaving as is.** The cited `build-expo-web-export.sh:667` does not exist; the file is 246 lines and the code is at 132-134. The init-to-1-then-normalize shape is deliberate and already carries the reason at line 128: a bare `-n` test reads `=0` as "on", which for a flag whose wrong answer is a reused Metro cache is the wrong failure direction. Inverting it is a style preference on a safety-critical flag, so not worth the churn.

**4. `passWithNoTests` comment — already there.** ci.yml:573-578, directly above the shard command, covering both the implicit-`--changed` behaviour and why passing the flag explicitly would be worse (it turns "this shard collected nothing" into a silent green).

Also worth flagging: `test-default` ran **one** spec file on this PR — and only because I edited that file directly. `ci-location-sync-workflow.test.ts` reads `ci.yml` via `readFileSync`, so `--changed` never linked my ci.yml rewrite to it. All three contract tests were run locally instead. Pre-existing blind spot, but a clean demonstration of it.

Issue #4353, opened 2026-08-12
offline: keep your profile and board list ready on cold start


marcodejongh:
Follow-up to the #4002 decision (`docs/offline-reads.md`). Bucket 2 of the provenance rule: **an allowlisted, user-scoped persisted cache owns small identity-shaped keys with no local table.**

Nothing persists the React Query cache today — `query-provider.tsx` builds a plain `QueryClient`, so every cold start is empty. The one that hurts is `profile`: `useProfile` is a plain network query, and `/boards/manage` renders a hard "Something went wrong" when `currentUserId` is missing. #4309 works around that specific screen by decoding the JWT `sub`; this is the general fix.

## Scope: the allowlist, and nothing else

`['profile']`, `['myBoards', …]`, `['myGyms']`, `['grades', board]`, `['angles', board, layout]`, `['publicProfile', selfId]`.

Explicitly **not** on it: anything SQLite already owns (`searchClimbs`, `infiniteSearchClimbs`, `searchClimbsCount`, `climb`, `boardseshGrade`, `localTicks`, `logbook`), `['activeBoard']` (already AsyncStorage-backed in `use-active-board.ts` — do not double-store), and everything with "now" semantics (feeds, sessions, `searchUsers`, `comments`, `bulkVoteSummaries`).

Budgets: under 100 KB serialized target, 512 KB hard cap with lowest-priority-first eviction, 64 KB per entry, `maxAge` 14 days for identity/config and 24 h for `publicProfile`.

## Build it in-house, not on `@tanstack/react-query-persist-client`

`dehydrate`, `hydrate`, `IsRestoringProvider` and `useIsRestoring` are all already exported by the pinned `@tanstack/react-query@5.101.4`, and `react-native-mmkv` + `@react-native-async-storage/async-storage` are already direct `packages/mobile` dependencies. So the stock persister buys about 150 lines and costs `package.json` + `bun.lock` churn — which per #4122 trips the OTA "native change" check as a false positive.

It also does not give three things this needs:

1. The allowlist as a **hard gate** on the dehydrate path, not a filter someone can forget to update.
2. A `userId` stamp validated at the auth boundary.
3. A **synchronous** MMKV restore on native, so there is no `isRestoring` frame at all.

And it dehydrates mutations by default. `pending_mutations` in SQLite is the outbox — a second persisted outbox is a genuine double-submit hazard. Hard-code `shouldDehydrateMutation: () => false` with a test asserting it.

[...]

PR #4403, opened 2026-08-13
feat(mobile): keep your profile and boards ready on a cold start

marcodejongh:
## Summary

Refs #4353. Stage 2 of the #4002 decision (`docs/offline-reads.md`): an in-house, allowlisted, user-scoped persisted React Query cache for `packages/mobile`, wired into the #3621 sign-out wipe.

Nothing persisted the query cache before this, so every cold start was empty — and `/boards/manage` renders a hard "Something went wrong" when `currentUserId` can't resolve, which offline it could not. #4309 worked around that one screen by decoding the JWT `sub`; this is the general fix.

**⚠️ THIS PR STAYS DRAFT until the on-device QA gate below has been run.** A synchronous MMKV read plus `JSON.parse` before first render is small but unmeasurable in vitest, and the relaunch ordering of `resolveAuthSession` → `handleSignedOutTransition` is the one behaviour no test can fully prove.

### What landed

New module `packages/mobile/src/lib/query-persist/` — pure TS with storage injected or behind a `.web.ts` fork:

- **`allowlist.ts`** — six rules (`profile`, `myBoards`, `myGyms`, `grades`, `angles`, `publicProfile`-self), each pinned on an exact head **and an exact arity**. Arity is what makes it a gate rather than a prefix filter: `['profile', x]` misses, and a future `['profileSettings']` cannot slip through. `publicProfile` additionally guards on the owner's own id.
- **`envelope.ts`** — `{ version, userId, savedAt, evicted?, queries }`. **No `mutations` field**, so a persisted outbox alongside SQLite's `pending_mutations` is unrepresentable — and `dehydrate.ts` hard-codes `shouldDehydrateMutation: () => false` on top. Two independent defences, one test each.
- **`budget.ts`** — 64 KB per entry, 512 KB cap with lowest-priority-first eviction (so `['profile']` is the last thing standing), and a 100 KB target asserted against a realistic populated fixture.
- **`restore.ts`** — re-applies the allowlist and per-rule `maxAge` **on the restore path too**, so a future or tampered blob can't smuggle a non-allowlisted key (or another climber's public profile) into memory. Seeds the writer's merge set from what it hydrated.
- **`writer.ts`** — 1 s trailing-edge throttle, merge-on-write against the previous set, and the owner read at write-**fire** time.
- **`runtime.ts`** — module singletons. `suspendCacheWriter()` is a **pause**, not a stop: it clears the owner, the pending throttle and the merge set, and `setPersistOwner` re-arms the same subscription.
- **`auth-boundary.ts`** — pure `decideOwnerTransition` plus an idempotent `adoptPersistedQueryCache` that short-circuits on a matching owner and consumes `lastRestore` once.
- **`storage.ts` / `storage.web.ts`** — native uses a **dedicated** `boardsesh-query-cache` MMKV instance (so a 512 KB blob never lands in `boardsesh-settings`, whose `clearAll()` would take it with it) plus a short owner sentinel; web uses login-scoped `userScopedStorageKey` blob + meta keys in the raw AsyncStorage keyspace, so `sweepOrphanedUserStorage` can reclaim them.

Wiring:

- `QueryProvider` restores **synchronously** inside its `useState` lazy initializer on native — no `isRestoring` frame, no `IsRestoringProvider` — and flushes on AppState background. Web restores asynchronously at the auth boundary, awaited before `setIsLoading(false)` releases the tree.
- `clearPersistedUserStores` calls `suspendCacheWriter()` as its first synchronous statement and deletes the blob — one call site, **both platforms**.
- `handleSignedOutTransition`'s `needsFullCleanup` now also fires when a persisted blob exists, so a token that expired while the app was killed no longer relaunches into the previous session's profile.
- One new permanent event name: `Offline Query Cache Restored`.
- Comment-only precedence documentation on `/boards/manage` and `/boards/index` — no logic change; `offlineBoardRows({ cachedMyBoards })` already prefers a warm `['myBoards']` entry, this just makes it warm at cold start.

### Why in-house rather than `@tanstack/react-query-persist-client`

`dehydrate` / `hydrate` are already exported by the pinned `@tanstack/react-query@5.101.4`, so the dependency buys ~150 lines and costs `package.json` + `bun.lock` churn — which per #4122 trips the OTA "native change" check as a false positive. It also doesn't give the allowlist as a hard gate, a `userId` stamp validated at the auth boundary, or a synchronous native restore. **This PR is JS/TS only: no `package.json`, no `bun.lock`, no native fingerprint movement.**

### Why the writer pauses instead of stopping

`clearPersistedUserStores` runs on the **light** signed-out path too, which fires on every logged-out cold start and every anonymous foreground `checkAuth`. A latched stop would therefore be dead from the first anonymous launch, and the first sign-in after it would persist nothing — the epic's exact target user (sign in today, gym tomorrow with no signal). Reading the owner at fire time also closes the race a stop was meant to cover: a write queued mid-sign-out finds a null owner and does nothing.

## Release Notes

- Open the app with no signal and it already knows who you are — your profile, your board list and your gyms are on the phone now, so My Boards comes up with your walls grouped the way you left them instead of "Something went wrong".
- Signing out still wipes all of it, and so does a session that expired while the app was closed.

## Test plan

All commands run in the worktree, foreground, literal outcomes:

- `vp run typecheck:mobile` — **PASS** (`tsc --noEmit`, Done in 29.51 s)
- `vp run test:mobile` — **PASS**, 657 files / 5954 tests passed
- `vp test run --reporter=agent packages/shared/analytics` — **PASS**, 4 files / 23 tests
- `vp run check:mobile-bundle` — **PASS**, `[mobile-bundle] android bundle: OK (4.0 KB)`
- `vp check` — lint clean; the only formatting failures are two pre-existing, untouched files (`docs/design/web-reposition/gyms-directory.html`, `homepage-marketing.html`) that already fail on `main`. Everything this PR touches is formatted.

New tests (24 cases across 11 files):

| Area | What it pins |
| --- | --- |
| `allowlist.test.ts` | the six shapes match; wrong arity, `profileSettings`, every SQLite-owned key, `activeBoard`, feeds all miss; `publicProfile` only for the owner |
| `envelope.test.ts` | round trip; `json` / `shape` / `version` classification; `utf8ByteLength` == `Buffer.byteLength`; a serialized envelope has no `mutations` key even from a client holding mutations |
| `dehydrate.test.ts` | zero mutations dehydrated from a client holding a **paused** mutation (query-core's own default keeps it, asserted as a control); only successful idle allowlisted entries survive |
| `budget.test.ts` | 70 KB entry dropped oversize; a 600 KB set evicts lowest-priority-first and keeps `['profile']`; oldest-first within a priority |
| `restore.test.ts` | `['infiniteSearchClimbs']` and a stranger's `publicProfile` in a hand-crafted blob hydrate neither; `['myBoards', undefined]` survives the `undefined→null` round trip; owner mismatch / missing hint hydrate nothing; 15-day `profile` and 25-hour `publicProfile` dropped while a 13-day `grades` survives; a fresher live entry is never clobbered; the merge set is seeded from the restore |
| `writer.test.ts` | no write with a null owner; the first write after the owner arrives; `flush()` cancels the pending timer; **the C1 regression**: suspend → nothing written → `setPersistOwner` → a write lands; a gc'd entry carried forward but not one past its maxAge; a restore-seeded merge set survives `client.clear()` |
| `auth-boundary.test.ts` | `decideOwnerTransition` truth table; **the C2 regression**: `readPersistedCacheAsync` called with literal `undefined`; **the C3 regression**: a second and third adopt for the same user perform zero reads, hydrates and emits; an account switch reports nothing |
| `storage-native.test.ts` | own per-`id` MMKV mock (the shared stub shares one process-wide map): the blob never lands in `boardsesh-settings`; presence answers from the sentinel, not the blob; clear removes both keys |
| `storage-web-sweep.test.ts` | `sweepOrphanedUserStorage`'s predicate matches both query-cache keys of a stale login and neither the live login's nor another user's |
| `cold-start-budget.test.ts` | profile + 6 fully-populated boards + 3 gyms + 24 grades × 2 boards + 9 angles × 2 boards serializes under 100 KB |
| `query-provider-persist.test.tsx` | `getQueryData(['profile'])` is non-undefined during the **first render pass**; a missing sentinel hydrates nothing; background writes only once armed; pre-auth providers leave all six hydrated entries at 0 observers / idle / unchanged `dataUpdatedAt` |
| `pre-auth-query-cache-audit.test.ts` | every listed pre-auth path exists on disk; every JSX tag between `<QueryProvider>` and `<AuthProvider` is covered; no query-cache API anywhere in that import graph; `OfflineEngineFlagSync` itself is clean |
| `auth-provider.test.tsx` (extended) | sign-out suspends the writer **before** deleting the blob; a logged-out cold start with a blob runs the full `runSignedOutCleanup`, without one keeps the light sweep; the auth boundary adopts with `undefined` |

### Device QA gate (required before this leaves draft)

JS-only, so a `pr-<number>` OTA reaches an existing TestFlight/store binary — no new build. Run on the **slowest Android device available**, not the x86_64 emulator.

1. Sign in; visit My Boards, My gyms, the Climbs tab on **two** boards, and your own profile on You. Background the app once so the flush writes.
2. `adb shell run-as com.boardsesh.app ls -l files/mmkv/` — a `boardsesh-query-cache` file must exist, separate from `boardsesh-settings`.
3. **Performance gate.** Force-stop, airplane mode, 5× `adb shell am start -W`; then 5 more runs with `files/mmkv/boardsesh-query-cache*` deleted between launches. **Accept if the median `TotalTime` regression is ≤ 30 ms.** Over that, cut grades/angles retention — do not raise the bar.
4. Still in airplane mode: `/boards/manage` renders with the correct "Your boards" / "Following" headers and **no** error panel. PostHog shows `Offline Query Cache Restored { outcome: 'hydrated' }` with a plausible `entryCount`/`bytes` and no `evicted`.
5. **Writer re-arm.** Sign out, sign back in as the same user, visit My Boards + My gyms, background: the MMKV file must reappear with a fresh mtime. Cold start in airplane mode — the board list must render. (A dead writer looks exactly like a passing step 4 and a failing step 5.)
6. **No duplicate adopt.** Background/foreground 3×: exactly **one** `Offline Query Cache Restored` for the launch, and **zero** `Persisted query cache owner mismatch` in Sentry.
7. **Sign-out wipe.** Back online, sign out: the MMKV file is gone; a cold start shows the login screen with no cached profile.
8. **Expired session, ONLINE.** Revoke the refresh token server-side, force-stop, cold start **with network**. Expect: signed out, no profile/board list from cache, blob gone, `syncEnabledBoards` cleared, downloaded catalogs still on disk. (Meaningless offline — only a server-*rejected* refresh resolves `anonymous`; an unreachable one stays authenticated-degraded.)
9. **Offline grace.** Sign in, let the access token near expiry, force-stop, airplane mode, cold start. Expect: still **signed in**, `/boards/manage` renders from cache, blob still on disk. This is the behaviour that would be destroyed by "fixing" step 8, so it is a gate.

## Decisions for Marco

Both flagged `needsMarco` in the plan; the defaults below are what shipped.

**1. Stale Home / session feeds behind a "last updated 3 h ago" banner, or keep them off the allowlist?**
Shipped: **kept off.** The allowlist is exactly the decision doc's six keys. Offline, feeds keep the honest "no signal" state from #4349.
Why: feeds have "now" semantics and are unbounded — the two properties Bucket 3 exists for. They'd blow the 100 KB target on their own and reintroduce the merge question the provenance rule was written to eliminate. Adding them later is an allowlist entry plus a banner; removing them after they ship is a regression users notice. This is the issue's own stated open question.

**2. Should `/boards/manage` show the full persisted board roster offline, or keep showing only downloaded boards?**
Shipped: **current behaviour unchanged.** When `shouldUseOfflineList` is true the screen still renders `offlineBoardRows(...)`, which filters to scopes with a completeness marker. The persisted cache only makes the `cachedMyBoards` tier warm, so names and owned/followed grouping are fresher; comments only, no logic change.
Why: the issue is explicit that #3897/#4309's MMKV path and its tests stay untouched, and `offlineBoardRows` already documents why an undownloaded board is hidden (activating it lands you on an empty climb list). Showing the whole roster offline needs a per-row "not downloaded" affordance and a new empty state — a separate product change. Flagged because a reader of the precedence comment might reasonably expect the full roster.

## Risks

1. **Highest — the `needsFullCleanup` widening.** A logged-out cold start with a blob now runs the full `runSignedOutCleanup`: `syncEnabledBoards` is emptied and `clearOfflineBoards()` runs. Downloaded catalogs survive (`purgeOfflineBoards` stays false), so re-enabling a board resumes instantly. This is identical to the treatment an in-session token expiry already gets, extended to the killed-app case, and it can only fire on a **confirmed** logout. Pinned both ways by tests and by QA steps 8 and 9.
2. The `handleSignedOutTransition` restructure adds an `await` to a function with a documented "publish anonymous before any storage await" invariant on web — mitigated by capturing `wasAuthenticatedThisSession` before the publish block and re-checking `isAuthTransitionCurrent` after the await. The existing web-transition tests stay green.
3. The owner-gate-at-fire-time is the whole correctness story for sign-out safety and re-arming. If a future edit moves the owner read back to schedule time, both the leak window and the dead-writer bug return. The writer test and the `runtime.ts` doc comment are the guards.
4. A synchronous MMKV read + `JSON.parse` of up to 512 KB on the JS thread before first render — unmeasurable in vitest, hence the mandatory device gate above.
5. `Offline Query Cache Restored` is permanent once shipped.
6. Stale profile/board names for up to 14 days offline; every screen refetches on mount when there is signal.

## Deviations from the plan

- **`DehydratedQuery` is not exported** by `@tanstack/query-core@5.101.4` (only `DehydratedState` is). Derived it as `DehydratedState['queries'][number]` and exported it as `PersistedQueryEntry`.
- **Sentry alarm condition widened.** The plan reported `owner_mismatch` only when `restorePersistedCache` returned that outcome. That misses the actual cross-user leak: a blob whose MMKV sentinel *agrees* with it but disagrees with resolved auth hydrates as `'hydrated'`, then gets evicted. The alarm now fires on `transition === 'evict-then-adopt' || outcome === 'owner_mismatch'`, and that case reports `outcome: 'owner_mismatch'` to PostHog so the funnel doesn't hide it as a successful hydrate.
- **An unusable blob is deleted, not just skipped.** The plan's evict branch keyed purely off `decideOwnerTransition`, which left a torn (`owner_missing`), corrupt (`unreadable`) or fully-aged-out (`empty`) blob on disk. Added an explicit `hasUnusableBlob` clause so those are cleaned up too. `owner_missing` remains deliberately silent in Sentry.
- **T-18 narrowed**, using the plan's own escape hatch: it mounts `QueryProvider` + `FeatureFlagsProvider` over the six hydrated entries rather than the full pre-auth stack, because `DatabaseProvider` / `OfflineEngineFlagSync` need the whole offline-sync-bridge mock recipe under vitest. T-17's source-graph guard is the durable check and covers `OfflineEngineFlagSync` with its own targeted assertion.
- **T-17's walk excludes `offline-sync-bridge.tsx` as a seed**, not just as a file. Walking from it drags `auth-provider.tsx` and the sync adapter into the graph — neither of which mounts pre-auth. `OfflineEngineFlagSync`'s own function body is asserted directly instead, and its two real dependencies are added as walk seeds.
- **Web meta value carries no `userId`** (just `{ savedAt, bytes }`): on web the owning account is already in the key itself, so duplicating it would be a second place to get wrong.

### DIFF
--- a/docs/offline-reads.md
+++ b/docs/offline-reads.md
@@ -26,6 +26,26 @@ Only for keys that have no local table and do not deserve one: `['profile']`, `[
 
 Budgets: target under 100 KB serialized, hard cap 512 KB with lowest-priority-first eviction, 64 KB per entry, `maxAge` 14 days for identity and config keys and 24 hours for `publicProfile`.
 
+#### How the persisted cache works
+
+Shipped in [#4353](https://github.com/boardsesh/boardsesh/issues/4353) as `packages/mobile/src/lib/query-persist/`.
+
+**The envelope.** `{ version, userId, savedAt, evicted?, queries }`. There is deliberately **no `mutations` field**: `pending_mutations` in SQLite is the one outbox, and a second persisted outbox is a double-submit hazard. The type makes it unrepresentable, and `dehydrateAllowlisted` additionally hard-codes `shouldDehydrateMutation: () => false` — two independent defences, one test each. `version` is bumped only when the persisted _value_ shape breaks; a changed query key yields a different `queryHash`, so a stale entry is simply never found and ages out.
+
+**The six rules.** Each pins an exact head **and an exact arity**, which is what makes the allowlist a gate rather than a prefix filter (`['profile', x]` misses; a future `['profileSettings']` cannot slip through). Priorities, lowest evicting first: `profile` 60, `myBoards` 50, `myGyms` 40, `grades` 30, `angles` 20, `publicProfile` 10. `publicProfile` also carries a guard rejecting any id but the blob owner's. The allowlist is re-applied on the **restore** path, not only on dehydrate, so a future or tampered blob cannot smuggle a non-allowlisted key — or another climber's public profile — into memory.
+
+**Native restores synchronously, web asynchronously.** On native, MMKV's read is synchronous, so `QueryProvider`'s `useState` lazy initializer hydrates before React's first render: no `isRestoring` frame, no `IsRestoringProvider`, no splash flicker. On web the storage key is `userScopedStorageKey(...)`, which returns `null` until `setCurrentUserStorageOwner` runs, so there is genuinely nothing readable before auth resolves — the restore is awaited inside `handleResolvedAuthenticatedTransition`, which precedes `setIsLoading(false)` while children sit behind `AppLoadingSplash`. `hydrate`'s `dataUpdatedAt >` guard means a late restore can never clobber a live fetch either way.
+
+**The native owner sentinel.** A short `queryCacheOwnerV1` key holds the user id next to the blob, in a **dedicated** `boardsesh-query-cache` MMKV instance (so a 512 KB blob never lands in `boardsesh-settings`, whose `clearAll()` would take it with it). A blob with no sentinel hydrates **nothing** and is deleted. That is not fussiness: `persistedQueryCacheExists` answers from the sentinel rather than reading up to 512 KB of JSON, so allowing a sentinel-less blob to hydrate would let hydrated data be invisible to the sign-out wipe. Presence and hydratability have to agree exactly.
+
+**Merge-on-write.** React Query's 30-minute `gcTime` deletes any entry with no observer, so a replace-on-write persister would erode toward "only what you looked at in the last 30 minutes" — `['angles', board, layout]` for a board you did not open today would vanish. Each write merges: fresh entries win by `queryHash`, previously written entries survive when absent from the fresh set and still inside their rule's `maxAge`. The merge set is **seeded from what the restore hydrated**, which closes the same hole at the other end.
+
+**The writer is paused, never stopped.** It subscribes to the query cache once for the app's lifetime and is gated purely by a runtime owner read at **write-fire time**, never at schedule time. `suspendCacheWriter()` clears the owner, the pending throttle and the merge set; `setPersistOwner(userId)` at the auth boundary re-arms it. A latched stop would be unshippable: `clearPersistedUserStores` runs on the light signed-out path too, which fires on every logged-out cold start and every anonymous foreground check, so the writer would be dead from the first anonymous launch and the first sign-in would persist nothing. Reading the owner at fire time also closes the race a stop was meant to cover — a write queued mid-sign-out finds a null owner and does nothing.
+
+**The sign-out contract.** The blob is deleted **inside `clearPersistedUserStores`** — one call site, cross-platform, not a parallel delete that can drift. `suspendCacheWriter()` is its first synchronous statement. And `handleSignedOutTransition`'s `needsFullCleanup` now also fires when a persisted blob exists, so a token that expired while the app was killed no longer relaunches into the previous session's cached profile. That widening is safe because it can only fire on a **confirmed** logout: `resolveAuthSession` returns `anonymous` only for an absent token or a server-rejected refresh, while an unreachable refresh stays authenticated-degraded — so an offline cold start still restores normally.
+
+**Telemetry.** One event, `Offline Query Cache Restored`, at most once per launch and only when a blob existed. `outcome` is one of `hydrated | unreadable | owner_mismatch | owner_missing | empty`; `owner_mismatch` is also a Sentry report (`source: 'query-persist'`) and should be ~0, while `owner_missing` is a torn write and deliberately silent.
+
 ### Bucket 3 — honest offline states, no storage at all
 
 Feeds (`sessionGroupedFeed`, `activityFeed`), session detail, board presence, `searchUsers`, `bulkVoteSummaries`, `comments`, `gymMembers`, `nearbyBoards`/`nearbyGyms`, `betaLinkPreview`.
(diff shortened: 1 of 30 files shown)


marcodejongh:

Conflicting: packages/mobile/src/providers/auth-provider.tsx, packages/shared/analytics/src/events.ts.

Issue #4369, opened 2026-08-12
W-15: ship the SSR climb and list front doors on both trees


marcodejongh:
Part of #4358. **Blocked by:** #4359 (signed), #4361, #4362, #4365, #4368, **#4167** (W-10). W-08 (#4071) is merged.

## Context

This is the heart of the programme. Today `view/[climb_uuid]/page.tsx:15,124` renders `BoardPageClimbsList` — the interactive queue-bearing surface — on both the legacy config-tuple tree and the `/b` slug tree. This PR replaces it, on both trees, with a server-rendered front door: a crawlable climb page whose only action is a "Climb this" hand-off into the app.

It is also the PR the real-device QA gate binds to. **W-15, not W-16, is the moment production loses the interactive climbing surface**, so if device QA fails afterwards, www is already degraded. Everything through #4368 is a revert of ≤1,500 lines; this one is too, but only if the app-side sheets actually work.

### The climb front door

New `packages/web/app/components/climb-front-door/`, rendered by **both** `view/[climb_uuid]/page.tsx` files in place of `BoardPageClimbsList`. Both pages keep their existing `generateMetadata`, `scheduleOverlayWarming` and LCP preload.

Server-renders:

- A **visible `<h1>`** — promote `ClimbViewSeoFragment` out of `sx={visuallyHidden}`.
- The board render as a plain **`<img src={buildOverlayUrl(...)}>` with explicit dimensions**. **Not** `BoardRenderer` / `BoardImageLayers` — both are hook-bearing client components.
- A facts `<dl>` from `getClimb` (`lib/data/queries.ts:51`, `unstable_cache` 3600s), plus first ascent.
- A **setter link**.
- **Angle cross-links** from `getClimbStatsForAllAngles` (`lib/data/queries.ts:74`), which already returns per-angle grade + ascents. This turns the angle dimension from duplicate content into a free internal-link mesh.
- **Beta videos** from `board_beta_links`.
[...]

Issue #4358, opened 2026-08-12
Epic: web reposition — retire the climbing surfaces from packages/web


marcodejongh:
`packages/web` still ships the entire interactive climbing product — queue, play drawer, party/WebSocket sessions, Bluetooth LED control, tick logging, climb authoring, board browsing — while that same product already runs in the Expo app's browser export at `app.boardsesh.com`. This epic finishes the reposition: Next.js becomes a crawlable SSR **front door** (marketing + climb view + board list + gym/kiosk/admin + auth + public API), every interactive surface lives only in the Expo app, and organic traffic plus every share link ever sent keeps working throughout. Grounded in the W-00 usage pull (PostHog `$pageview` + read-only prod counts, 2026-07-31) and a live GitHub reconciliation on 2026-08-12.

## The honest state

- **There is not one crawlable link to any climb on www today.** `packages/web/app/components/climb-card/climb-list-item.tsx:568` is an `onClick` on a `<Box>`. Every climbing surface is link-discovered only, and `app/sitemap.ts` lists 8 static paths × the locale set — no climb, board, setter or playlist URL is ever submitted.
- **The addressable universe is 771,159 climbs** (listed && !draft, read-only prod count, 2026-07-31): kilter 314,615 · moonboard 280,755 · tension 142,310 · grasshopper 21,802 · decoy 8,352 · soill 1,972 · touchstone 1,353. Tier 2 (≥10 ascents at some angle) is **128,655**. The `~424k` figure in `docs/kilter-sync.md:164` was Kilter-only and stale. MoonBoard out-ranks Kilter on tier 2 (75,074 vs 33,329) because of its per-angle ascent density.
- **Traffic on the delete set is small and concentrated** (90 days to 2026-07-31, `$pageview` on `www.boardsesh.com`): `/you*` 8,900 views / 730 users · `/feed` 1,423 / 599 · `/settings` 1,206 / 471 · `/discover/[smart]/[user]` 454 / 126 · `/notifications` 168 / 64 · `/development` 1 / 1 · `/import-beta` **0 / 0** · hold-classification **0 / 0**. Hold-classification is not a route at all — no `page.tsx` exists; it is a component reached from `user-drawer` and name-checked in `/help`, with zero pageviews *and* zero custom events (`event ILIKE '%hold%class%'` → empty).
[...]

PR #4418, opened 2026-08-13
feat(mobile): let signed-out web visitors reach read-only board routes with a next= return

### LINKED ISSUE #4361: W-06: let signed-out visitors browse read-only routes on expo-web, with a next= return
Part of #4358. **Blocked by:** **#3823 merged** (it adds `BoardRouteRedirect.tsx`, which this issue extends). W-01 (#4065) already shipped.

## Context

The web reposition turns ~771k climb URLs on `www.boardsesh.com` into SSR front doors whose only action is a "Climb this" hand-off into `app.boardsesh.com`. Today `packages/mobile/src/providers/auth-provider.tsx:1027` is a bare `<Redirect href="/auth/login" />` for every non-`/auth` route, with no `returnTo` anywhere under `packages/mobile/app/auth/`. So every organic visitor who taps the CTA hits a login wall and loses the climb they came for — and 500k front doors funnelling into a login on a noindex host is exactly Google's doorway-page shape.

The locked decision is: **relax the gate for a read-only route set, on web only**, and carry the attempted path through login. The native branch must be byte-for-byte unchanged.

## The work

In `packages/mobile`, gated on `Platform.OS === 'web'`:

- **Relax `auth-provider.tsx:1027`** for a read-only route set (climb view, board list, and the board-route handoff paths #3823 introduced). Every change is behind the `Platform.OS === 'web'` check; the native code path must be identical to what ships today.
- **Carry the attempted path into `/auth/login?next=…`** and honour it after login. Validate the value as **app-relative before using it**: it must start with a single `/`, must not start with `//`, and must not contain a scheme. Reject anything else and fall back to the default post-login destination.
- **Add the arrival event** in `packages/mobile/src/components/BoardRouteRedirect.tsx` (the component #3823 adds): `'Board Route Handoff'` with `{ kind, status, source }`, where `status` is one of `resolved` / `not_found` / `auth_required`.

Files: `packages/mobile/src/providers/auth-provider.tsx`, `packages/mobile/app/auth/login.tsx`, `packages/mobile/src/components/BoardRouteRedirect.tsx`.

### The warning that makes this a Fable-review PR

**This ships to the entire native store fleet as a production OTA on merge to `main`** — `.github/workflows/mobile-ota-production.yml` fires on `packages/mobile/**`, and runtimeVersion uses the `fingerprint` policy, so a JS-only change keeps the same fingerprint and the OTA lands on every installed binary. **Fingerprint-neutrality guarantees delivery, not safety.** A "web-only" change that isn't actually gated reaches phones within hours.

## Verifies

- [ ] A test asserts the **native** branch of the auth gate is byte-unchanged (snapshot the native path, not just "web works").
- [ ] A signed-out deep link to `/b/{slug}/{angle}/view/{uuid}` on `app.boardsesh.com` lands on that climb after login.
- [ ] `next=` values that are absolute URLs, protocol-relative (`//evil.example`), or scheme-bearing are rejected, with a test each.
- [ ] `'Board Route Handoff'` fires with each of `resolved`, `not_found`, `auth_required`.

## Gates

**GM** (all four, always):

- `vp run typecheck:mobile`
- `vp run test:mobile`

Do **not** use `vp test --project mobile run` — the flag-before-`run` order runs ~1 file and reports a false green. Use `vp run test:mobile`.

## Standing-rule callouts

- **Standing rule 2** — the PR body must state which of the three production targets this reaches. Answer: **`app.boardsesh.com` *and* the native store fleet via auto-OTA.** The web-only claim carries the byte-unchanged native test.
- **Standing rule 7** — publish with `eoas publish --rollout-percentage` staging, name the last-good commit in the PR body for an emergency re-publish, and hold full rollout until the PostHog OTA-health signature is clean: `Downloaded` events with **nonzero** Status launches for the new `updateId`. A fingerprint-matched OTA bricked prod once already.
- **Fable review required** — it auto-OTAs to the native fleet, and every PR touching `packages/mobile` gets a Fable native-OTA review regardless of tier. Opus implements; Opus runs all three QA lenses; Fable reviews.
- **Standing rule 1** — app-side route work and www-side CTA work never share a PR. The www CTA that depends on this ships in #4369, at least one production deploy later.

marcodejongh:
## Summary

Part of #4358.
Closes #4361.

`app.boardsesh.com` bounced every signed-out arrival to `/auth/login` and forgot which climb they came for — the exact failure mode that makes the www front door (#4369) worth nothing. W-06 relaxes the Expo app's auth gate for the read-only board URLs **on web only**, and carries the attempted path through login as a validated `?next=`.

- `src/lib/routing/read-only-routes.ts` — the allow-set and the `next` contract, pure and platform-free. Shape-only matching: it never resolves a slug or config against the catalogue, so a well-formed URL whose board is gone reaches that route's own not-found instead of a login wall. Locale-prefixed shares (`/es/…`, `/fr/…`, `/de/…`) are stripped before matching, or every non-English share link still hits a bare wall (they match no Expo route and land on `+not-found`).
- `src/lib/routing/anonymous-auth-gate.ts` / `.web.ts` — the platform fork. Native is a **constant module** (`false`, `() => false`, `() => '/auth/login'`, `() => null`); the web fork reads `window.location`, base-strips `EXPO_BASE_URL` (`/app` on the www-mounted export, `/` on the subdomain) and hands back base-relative hrefs.
- `auth-provider.tsx` — the gate keeps its literal `<Redirect href="/auth/login" />`. The only change is `&& !isAnonymousReadOnlyLocation()` on the condition, plus `readPostLoginReturnHref() ?? '/(tabs)/home'` on the authenticated branch (post-login navigation lives here, not in `login.tsx`, which never navigates on success and would race this redirect).
- `use-board-route-target.ts` — a relaxed route short-circuits to a new `auth-required` status **before** any resolution: `useAdoptedBoard` is disabled, the climb query is left unarmed and the hand-off effect returns early, so `resolveBoardForSession` / `fetchAllMyBoards` / `createBoard` / `setActiveBoard` / `useClimb` / `openClimbInPlayDrawer` never fire for someone with no account.
- `BoardRouteRedirect.tsx` — `auth-required` redirects to `/auth/login?next=<path>`. A successful hand-off has no status of its own: it is reported through an `onHandedOff` callback the hook fires from inside the hand-off itself (see the telemetry note below).
- `auth.web.ts` — the browser-OAuth round trip carries `next` on the NextAuth `callbackUrl`. Without it the fix is dead on the most common signed-out login: "Continue with Google" navigates the document away and the callback URL is rebuilt from scratch. No www change needed — its redirect callback returns allowed app-origin URLs verbatim, and `consumeWebOAuthReturn()` strips only the three `bs_oauth_*` params.
- `SHARED_EVENTS.BoardRouteHandoff` — `'Board Route Handoff'` with `{ kind, status, source }`, deliberately cross-platform.

### Why the hand-off event is fired imperatively

`status: 'resolved'` cannot be derived from a rendered status, and a first cut that did would have shipped a funnel reading ~100% failure. The hand-off effect calls `router.replace` / `router.back` in the same body as anything else it queues; React batches them, the navigator drops the redirector in the very render that would have carried the new status, and the update is discarded with the fiber. A throwaway probe under the mobile vitest — a child that sets its own state and calls the parent's "remove me" callback in one effect — committed only the pre-hand-off status, never the second one. The two suites that appeared to cover it could not see the problem: `use-board-route-target.test.tsx` runs against `replace: vi.fn()`, which leaves the screen mounted, and `BoardRouteRedirect.test.tsx` mocks the hook outright. So the success leg is fired imperatively next to `handedOffRef.current = targetKey`, exactly the way `join/[sessionId]` fires `Session Joined` before it replaces itself, and there is now a case whose router really unmounts the harness in the same batch.

`status: 'not_found'` is held back for a **parsed** URL while the device is offline. That state is the transient one `useAdoptedBoard`'s reconnect watcher heals — the pre-existing `re-resolves when the network comes back after an offline failure` test asserts the app passes through it — so reporting it would file a failure for every offline cold open that later lands, and count the same open twice, since the `resolved` follows under a different report key. An unparsed URL is a dead end whatever the network is doing and still reports.

### Open-redirect contract

`next` passes **two** independent gates before it is followed. `isSafeReturnPath`: non-empty, ≤512 chars, no control characters or whitespace, no scheme, a single leading `/`, no `//`, no backslash (browsers normalise `\` to `/`, so `/\evil.example` is protocol-relative). Then `isReadOnlyAnonymousPath`, which pins the destination to a shape the gate itself could have produced — so a hand-crafted `?next=` can never reach an arbitrary in-app route. `URLSearchParams.get` decodes once, so `%2f%2fevil.example` arrives as `//evil.example` and the protocol-relative rule catches it. A `next` is only ever *produced* for the relaxed allow-set, which keeps "every `next` the app emits is one it would serve anonymously" true by construction on both sides. Rejections are tested one case each.

### Scope boundary (reconciling the epic's DoD wording)

A relaxed route **mounts** anonymously, then short-circuits to `auth_required` and sends the visitor to login carrying the path — they land on the climb **after** signing in, which is what this issue's own Verifies asks for. It does not resolve a board anonymously, and that is forced by the data layer rather than caution: the config-tuple form mints a `UserBoard` through `createBoard`, which is `requireAuthenticated` (`packages/backend/src/graphql/resolvers/social/boards.ts`), and the hand-off destination is the Climbs tab plus the play drawer. `boardBySlug` and `climb` **are** anonymous-capable, so a genuinely anonymous in-app climb preview is buildable — but it is a new surface, not a gate change, and it is out of scope here. The epic's DoD bullet ("a signed-out deep link … resolves to that climb") is satisfied in the after-login sense. Nothing at all is asked of the server for a visitor with no account: the config-tuple form carries its whole board config in the URL, so the public `climb` query would otherwise have armed itself with no board to wait for — it is gated on the same `authRequired` flag, and a signed-out tuple-climb case asserts it stays unarmed while a signed-in one asserts it still fires.

## Production targets (standing rule 2)

**`app.boardsesh.com` AND the native store fleet via auto-OTA.** `.github/workflows/mobile-ota-production.yml` fires on `packages/mobile/**` and `packages/shared/**`, and the `fingerprint` runtimeVersion policy is unchanged by a JS-only diff, so this lands on every installed binary within hours of merge.

The split, stated rather than implied:

- **Web-only, asserted native-inert.** The gate relaxation and the `?next=` return. `anonymous-auth-gate.test.ts` asserts the native fork returns `false` / `'/auth/login'` / `null` across the entire route corpus *while a `window.location` is stubbed for each path* — proving it ignores the address bar rather than merely happening to be false. The new `native auth gate parity` suite in `auth-provider.test.tsx` asserts at render level, under `Platform.OS = 'ios'`, that every relaxed path still redirects to exactly `'/auth/login'` with no query and **never renders children**, and that a valid `?next=` in `window.location.search` is ignored on the authenticated branch (`'/(tabs)/home'`). A fork export-parity test keeps the two forks from drifting.
- **Cross-platform and additive.** The `Board Route Handoff` event — one capture per board-route open, including in-app opens through `(tabs)/climbs/[climbUuid]`. It is the only signal for whether deep links resolve on the native fleet too, and it is the one thing in this diff a store-fleet user's app will start doing that it does not do today. Nothing about it is visible on screen: no navigation, no copy, no render changes. (The `resolved` route status an earlier revision added is gone again, so the fleet also loses the extra render it caused.)

## Staged rollout (standing rule 7)

Publish with `eoas publish --rollout-percentage`, not a full-fleet publish. **Last-good commit for an emergency re-publish: `52167a1e6cb465d1d38250fc3007173d2fe5ab9d`** (this branch's base; re-read `gh run list --workflow=mobile-ota-production.yml --branch main --limit 1` immediately before merge and use the real last-good if `main` has moved). Hold full rollout until the PostHog OTA-health signature is clean: `Downloaded` events with **nonzero** Status launches for the new `updateId`. A fingerprint-matched OTA bricked prod once already.

Per standing rule 1, the www CTA that consumes this ships in #4369, at least one production deploy later.

**Fable review required** — it auto-OTAs to the native fleet, and every `packages/mobile` PR gets a native-OTA review regardless of tier. The three Opus QA lenses (delete-safety, SEO, native-OTA) apply.

## Release Notes

No user-visible native change. The gate relaxation and the `?next=` return are web-fork-only, asserted inert on native by the two suites above. What does reach the store fleet is one additive analytics capture per board-route open — no navigation, no copy, nothing on screen — so there is nothing for the "What's New" screen. The user-visible copy for this flow ships with the www front door in #4369.

## Test plan

| Command | Result |
|---|---|
| `vp run test:mobile` | pass — 658 files, 6258 tests |
| `vp run check:mobile-bundle` | pass — native Metro graph resolves the inert fork and never pulls `read-only-routes.ts` in (`isReadOnlyAnonymousPath` / `MAX_RETURN_HREF_LENGTH` absent from the Android bundle) |
| `vp run check:mobile-web-bundle` | pass — web export resolves `anonymous-auth-gate.web.ts` |
| `vp run typecheck` (full) | pass |
| `vp test run --reporter=agent --project web` | pass — 485 files, 5987 tests (`SHARED_EVENTS` is a web dependency too) |
| `vp check` | lint clean; format reports two **pre-existing** offenders only, `docs/design/web-reposition/{gyms-directory,homepage-marketing}.html`, untouched by this branch and already failing at `52167a1e6` |

New / extended suites:

- `src/lib/routing/__tests__/anonymous-auth-gate.test.ts` — native fork inert across the whole corpus with a stubbed location, plus fork export parity. **The native-unchanged guarantee.**
- `src/providers/__tests__/auth-provider.test.tsx` — new `native auth gate parity` suite (render-level, `Platform.OS = 'ios'`). Every pre-existing test in the file is untouched; their unchanged green is part of the evidence.
- `src/lib/routing/__tests__/anonymous-auth-gate.web.test.ts` — allow-set matching, the login href, one rejection each for absolute / protocol-relative / double-encoded / scheme-bearing / backslash / out-of-allow-set `next` values, and `EXPO_BASE_URL=/app` base stripping.
- `src/providers/__tests__/auth-provider-anonymous-gate.web.test.tsx` — the web gate renders children on relaxed paths, emits a bare `'/auth/login'` on gated ones, honours a valid `next` and drops `//evil.example`.
- `src/lib/routing/__tests__/read-only-routes.test.ts` — pure predicate coverage, incl. locale-prefixed URLs, the `/estonia` whole-segment guard, and the routing-syntax segments (`/(tabs)/…`, `/b/..`) that hit the right segment count without naming a board.
- `src/lib/routing/__tests__/use-board-route-target.test.tsx` — the `auth-required` short-circuit fires no board resolution and no climb query at all, while the same URL signed in still arms one; unreachable when `RELAXES_ANONYMOUS_ROUTES` is false; the hand-off callback, including a case whose `router.replace` really unmounts the harness in the hand-off's own batch.
- `src/components/__tests__/BoardRouteRedirect.test.tsx` — new. Redirect on `auth-required`, one `Board Route Handoff` per outcome with `{ kind, status, source }`, no double-fire from either leg, the hand-off reported for a screen that never renders a terminal status, and the offline `not_found` held back until the retry fails online.
- `src/lib/__tests__/auth.web.test.ts` — the OAuth `callbackUrl` carries `next`, is byte-identical to today's without one, and `signInWithGoogle` actually wires the current screen's return path into it (not just the builder in isolation).

- `app/auth/__tests__/login-register-link.test.tsx` — new. Both directions of the register detour: login's sign-up link attaches `next` as a router param, register's "Already have an account? Sign in" hands it back, and both fall through to the bare route when there is nothing to return to (which is what native always does).

Review follow-ups already applied. `b354e3f8c`: the register leg is covered on both sides (`readPostLoginReturnHref` off `/auth/register?next=…`, and the provider landing a freshly registered visitor on the climb), `signInWithGoogle` is asserted to wire the return path into the callback URL, the `BoardRouteRedirect` redirect test asserts the component forwards the gate's href rather than a literal, and `MAX_RETURN_HREF_LENGTH`'s whole-value scope (query included, unlike the shape matcher) is documented as deliberate. `bc57a5b2c`: the hand-off event now deduplicates on the **URL** plus outcome rather than the event props — two climbs on the same board settle to an identical `{ kind, status, source }`, so the props alone would have undercounted every board-route open after the first on a reused route component; the URL never leaves the component.

QA round 2, all applied on top:

1. **The `resolved` leg never fired in production** (see "Why the hand-off event is fired imperatively"). Two lenses landed on it independently; a probe under the mobile vitest confirmed the render never commits. Fixed by reporting through `onHandedOff` from inside the hand-off, and the now-unobservable `resolved` route status and its mirror `useState` are deleted rather than left as decoration — that also takes one render back off the store fleet.
2. **`not_found` was inflated by the offline heal.** One offline cold open that later succeeded reported `not_found` and then `resolved`, so the funnel double-counted it as both. Held back for a parsed URL while `onlineManager` says offline; the retry reports if it fails online.
3. **The `auth-required` short-circuit didn't stop every anonymous call.** The tuple form's `useClimb` armed itself off the URL's own config. Gated on `authRequired`, which makes the comment above the short-circuit true and the scope-boundary paragraph above accurate.
4. **The `?next=` round trip was one-directional.** Register's two "Sign in" affordances dropped it, so Sign up → change of mind → Sign in lost the climb in three taps. Both now read the same `signInHref`.
5. **The allow-set accepted shapes the gate can't produce.** `next=/(tabs)/profile/a/b/40/list` and `/b/..` matched on segment count alone. Not an open redirect (same-origin, dead-ends at `+not-found`), but the comment claimed a stronger property than the code enforced, so routing-syntax segments are now rejected outright and both are in the gated corpus.

Still to run on a preview deploy of `app.boardsesh.com` (needs a browser, not CI): private window → `/b/{slug}/{angle}/view/{uuid}` → spinner → `/auth/login?next=…` → sign in with credentials → lands on that climb; repeat with Google OAuth; repeat with `/es/kilter/…/view/…` to confirm the `+not-found` locale bounce keeps the climb.

### DIFF
--- a/docs/expo-web-deployment.md
+++ b/docs/expo-web-deployment.md
@@ -147,7 +147,21 @@ the address bar does not survive the round trip on its own.
 
 `Board Route Handoff` fires once per board-route open on **both** platforms with
 `{ kind, status, source }`; it is the only signal for whether deep links resolve
-on the native fleet as well as on `app.boardsesh.com`.
+on the native fleet as well as on `app.boardsesh.com`. Two things about the
+wiring are load-bearing for reading the funnel:
+
+- `status: 'resolved'` is fired **imperatively** from inside the hand-off, not
+  derived from a rendered status. The hand-off effect calls
+  `router.replace` / `router.back` in the same body, React batches that with any
+  state update queued in the same flush, and the navigator drops the redirector
+  in the very render that would have carried the new status — so a
+  status-derived success leg never commits and the funnel reads as ~100%
+  failure. `join/[sessionId]` fires `Session Joined` the same way.
+- `status: 'not_found'` is **held back** for a parsed URL while the device is
+  offline. That state is the transient one `useAdoptedBoard`'s reconnect watcher
+  heals, so reporting it would file a failure for every offline cold open that
+  later lands, and count the same open twice. A URL that did not parse reports
+  either way.
 
 ### Telemetry (baked at build time)
 
--- a/docs/expo-web-deployment.md
+++ b/docs/expo-web-deployment.md
@@ -104,6 +104,51 @@ redirect allow-list accepts only the configured app origin and numbered app
 previews, and the shared `.boardsesh.com` session cookie is available when the
 Expo export reloads.
 
+### Signed-out read-only routes (the `?next=` return)
+
+The Next front door on `www` links every climb page into `app.boardsesh.com` at
+the same path. The browser export serves those paths to visitors who have never
+signed in, so the climb survives the login round trip.
+
+Relaxed shapes (`packages/mobile/src/lib/routing/read-only-routes.ts`), matched
+on shape only — never against the board catalogue, so a URL whose board is gone
+reaches that route's own not-found rather than a login wall:
+
+- `/b/{slug}`, `/b/{slug}/{angle}/list`, `/b/{slug}/{angle}/{view|play}/{climb}`
+- `/{board}/{layout}/{size}/{sets}/{angle}/list` and `…/{angle}/{view|play}/{climb}`
+- every one of the above under an `/es`, `/fr` or `/de` prefix — web keeps the
+  locale in the path, those URLs match no Expo route, so the matcher runs
+  `stripLocalePrefix` first
+
+**Native is untouched.** The relaxation lives in `anonymous-auth-gate.web.ts`;
+the native fork `anonymous-auth-gate.ts` is a constant module (`false`,
+`() => false`, `() => '/auth/login'`, `() => null`), so the gate in
+`auth-provider.tsx` behaves exactly as it does on the store fleet today. Every
+merge to `main` touching `packages/mobile/**` auto-publishes a production OTA
+onto every installed binary, which is why this is a fork rather than a
+`Platform.OS` check — and why both the module's inertness and the rendered gate
+are asserted by test.
+
+A relaxed route mounts, discovers it needs an account, and hands off to
+`/auth/login?next=<path>` rather than resolving anything anonymously: the
+config-tuple form mints a `UserBoard` through `createBoard`, which is
+`requireAuthenticated`. `auth-required` is therefore the terminal status for
+every relaxed route today, and the visitor lands on the climb after signing in.
+
+`next` passes two independent gates before it is followed: `isSafeReturnPath`
+(app-relative, length-capped, no scheme, no `//`, no backslash, no control
+characters) and `isReadOnlyAnonymousPath`, which pins the destination to a shape
+the gate itself could have produced. Values are read from
+`window.location.pathname` with `EXPO_BASE_URL` stripped (`/app` on the
+www-mounted export, `/` on the subdomain) and written back base-relative, since
+Expo Router re-applies the base. The browser-OAuth path carries `next` on the
+NextAuth `callbackUrl` (`src/lib/auth.web.ts`) — the document navigates away, so
+the address bar does not survive the round trip on its own.
+
+`Board Route Handoff` fires once per board-route open on **both** platforms with
+`{ kind, status, source }`; it is the only signal for whether deep links resolve
+on the native fleet as well as on `app.boardsesh.com`.
+
 ### Telemetry (baked at build time)
 
 `deploy-app-web` builds the export with `EXPO_PUBLIC_SENTRY_DSN`,
--- a/docs/web-reposition.md
+++ b/docs/web-reposition.md
@@ -132,6 +132,16 @@ plus `useOptionalPlaylistActivation`. Refactor `multiboard-climb-list.tsx` and
   — the climb-view page SSR-emits `ClimbViewSeoFragment` (the crawlable payload).
   Guards A2's drawer→static swap from dropping it.
 
+**W-06 (#4361) — the SPA end of the hand-off.** A front door that links into
+`app.boardsesh.com` is only worth building if the arrival works signed-out, so
+the Expo app's auth gate now stands aside for the read-only board URLs on web
+and carries the attempted path through login as a validated `?next=`. The URL
+shapes it relaxes, the two-gate `next` contract, the `.web.ts` fork that keeps
+the native gate a constant `false`, and why `auth_required` is a relaxed route's
+terminal status today are recorded in
+[`docs/expo-web-deployment.md`](./expo-web-deployment.md) under "Signed-out
+read-only routes".
+
 ## Phase A0 — blocking pre-delete QA gate (real devices)
 
 The teardown is a **hard delete** with no retained `?classic=1` runtime fallback,
--- a/packages/mobile/app/auth/__tests__/login-register-link.test.tsx
+++ b/packages/mobile/app/auth/__tests__/login-register-link.test.tsx
@@ -1,0 +1,90 @@
+// @vitest-environment jsdom
+//
+// The one hop the rest of the W-06 suites can't see: login forwarding `next` to
+// the sign-up screen. `AuthProvider` is what navigates after registering (it
+// reads the value straight off the location), so without this the two ends of
+// the register leg are only ever tested in isolation.
+
+import { describe, it, expect, vi, beforeEach } from 'vitest';
+import { fireEvent, render } from '@testing-library/react';
+import { createElement, type ReactNode } from 'react';
+
+const router = vi.hoisted(() => ({ push: vi.fn(), replace: vi.fn(), back: vi.fn() }));
+const returnHrefState = vi.hoisted(() => ({ current: null as string | null }));
+
+vi.mock('react-native', () => ({
+  Platform: { OS: 'web', Version: undefined },
+  KeyboardAvoidingView: ({ children }: { children?: ReactNode }) => createElement('div', null, children),
+  ScrollView: ({ children }: { children?: ReactNode }) => createElement('div', null, children),
+  View: ({ children }: { children?: ReactNode }) => createElement('div', null, children),
+  Text: ({ children }: { children?: ReactNode }) => createElement('span', null, children),
+  Pressable: ({ onPress, children }: { onPress?: () => void; children?: ReactNode }) =>
+    createElement('button', { onClick: onPress }, children),
+  StyleSheet: { create: (styles: unknown) => styles },
+}));
+
+vi.mock('expo-image', () => ({ Image: () => createElement('img', null) }));
+vi.mock('expo-router', () => ({ useRouter: () => router }));
+vi.mock('react-i18next', () => ({ useTranslation: () => ({ t: (key: string) => key }) }));
+
+vi.mock('../../../src/providers/auth-provider', () => ({ useAuth: () => ({ signInWithCredentials: vi.fn() }) }));
+vi.mock('../../../src/providers/theme-provider', () => ({
+  useTheme: () => ({
+    systemColors: { label: '#000', secondaryLabel: '#666', separator: '#ccc', secondaryBackground: '#eee' },
+    brandColors: { primary: '#7c3aed', warning: '#f59e0b' },
+  }),
+}));
+vi.mock('../../../src/hooks/use-native-oauth-sign-in', () => ({
+  useNativeOAuthSignIn: () => ({ signIn: vi.fn(), inProgress: false }),
+}));
+vi.mock('../../../src/components/AuthFieldset', () => ({ AuthFieldset: () => null }));
+vi.mock('../../../src/components/Button', () => ({ Button: () => null }));
+vi.mock('../../../src/components/Icon', () => ({ Icon: () => null }));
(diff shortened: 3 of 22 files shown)


marcodejongh:
**Fable native-OTA review: approved.** Reviewed the final diff (head after the QA-fix round) per the epic #4358 review rule. The native fork of `anonymous-auth-gate` is a constant module with zero imports; both auth-provider condition edits collapse to today's behaviour under it, verified by the corpus render tests and the bundle-level check that `read-only-routes.ts` symbols are absent from the native bytecode. The cross-platform pieces (`use-board-route-target`, `BoardRouteRedirect`) are additive-telemetry-only on native — every new branch is gated on `RELAXES_ANONYMOUS_ROUTES`, constant-false there. The `resolved`-as-callback redesign (instead of a status the navigator would discard mid-batch) is the right call. Reminder at merge time per standing rule 7: this auto-OTAs to the store fleet — stage the rollout and hold full rollout until PostHog shows Downloaded events with nonzero Status launches for the new updateId; last-good base is named in the PR body.

PR #4420, opened 2026-08-14
feat: support MoonBoard V2 additional-LED (light hold above)

fallaciousreasoning:
## Summary

- Adds support for the MoonBoard BLE "V2" `~D` config prefix, a feature of community LED-box firmware ([ArduinoMoonBoardLED](https://github.com/FabianRig/ArduinoMoonBoardLED)) that lights a second, dimmer LED next to each start or hand hold (typically the hold above), in addition to its normal role colour. Finish holds intentionally do not get the extra LED.
- Exposes an opt-in "Light the hold above" toggle (default off) on web and mobile, shown only for MoonBoard.
- Threads the setting through the shared packet builder, the native iOS Live Activity re-light path, and both platforms' native bridge configuration so background/widget re-lights use it too.
- Repaints the current climb when the setting changes. The Web Bluetooth auto-sender now queues the complete encoder request, so a toggle or delayed preference hydration cannot be swallowed by an in-flight write.
- Keeps physical BLE-packet dedup separate from the climb/wall-presence signature, preventing a settings-only repaint from creating a false wall change or duplicate Undo.
- Never sends `~D` before the clear-all `l##` frame, and documents that behavior on official MoonBoard controllers remains a real-device QA question.

Board with the new option enabled:
MoonBoard with adjacent LEDs enabled

Boardsesh setting:
Boardsesh Light the hold above setting

## Release Notes

- MoonBoard: light a dimmer LED above each start and hand hold when your controller supports it (Light Control → "Light the hold above").

## Test plan

- Web Bluetooth-control suite: 448/448 tests passed, including deferred-write coverage for a mid-write toggle, preference hydration without duplicate wall-feed/Undo reporting, and hydration while disconnected followed by the first reconnect write.
- Full mobile suite: 653 files / 5,989 tests passed.
- `vp run typecheck:web` and `vp run typecheck:mobile` passed.
- `vp run check:mobile-bundle` passed for iOS and Android.
- `vp run check:i18n` and the staged orphan check passed for en-US/es/fr/de.
- Scoped `vp check --fix` passed on every changed follow-up file; the pre-commit hook passed.
- Final minor-review regression run: 178 focused tests passed.
- Added Swift XCTest coverage for the native `~D` encoder and for decoding configurations persisted before `lightAdjacentHolds` existed. Swift tests run in the macOS CI job; local Linux simulator validation skips by design.
- Final Bluetooth review at `0c1a0faa0`: Codex Sol PASS with no P0–P2 findings, treated as the required Fable-equivalent review per maintainer direction. The shared encoder pins explicit `false` as byte-identical to its default path, clear-all omits the adjacent option, the MUI switch honors its emitted `checked` value, and an omitted native option remains `nil` until consumers apply the default.

## Native release QA

Carry these checks on the combined native build before the aggregate train merges to `main`:

- FabianRig controller: toggle on/off during slow and rapid writes.
- Official RedBearLab and Nordic controllers: confirm an unsupported `~D` prefix does not blank normal holds.
- Standard 18-row and Mini 12-row walls.
- iOS background/kill, widget, and Dynamic Island re-light, including upgrade from an older persisted configuration.
- Android and browser Web Bluetooth.
- Confirm no duplicate haptic, analytics, wall-feed, or Undo behavior.


fallaciousreasoning:
Fixed — `lightAdjacentHolds` is now `Bool?` with a custom initializer defaulting to `nil` (matching `numRows`), so Codable's synthesized decoder correctly back-fills missing-key configs instead of failing the whole decode. Call sites already read `?? false`. Also added the dedicated `useMoonboardLightAdjacentHolds` hook test flagged in the minor note. Pushed in dfc434f2d.

marcodejongh:
Nice I had wondered if someone had done this. Unrelated but We should add serial numbers to ble broadcast of the native firmware or hidden somewhere in the protocol.

For aurora led boxes we use the serial number to ensure the correct board & config is used

marcodejongh (approved):
Codex Sol/Fable-equivalent Bluetooth review PASS. No P0–P2 findings remain. Verified the serialized Web Bluetooth reassert/hydration path, separate physical vs. wall signatures, TS/Swift protocol parity, backwards-compatible Codable decode, finish-hold copy, focused tests, typechecks, and clean rebase onto release/next-ota. The documented real-controller and native-device matrix remains a combined release-train QA gate.

marcodejongh (approved):
Follow-up Sol/Fable-equivalent review PASS on the exact latest head. Clear-all now omits the adjacent option, explicit false is byte-identical to default, and disconnected preference hydration/reconnect behavior is covered at both provider and encoder-hook layers. No P0–P2 findings; 161 focused tests and diff cleanliness passed.

marcodejongh:
Addressed the latest Claude findings in `48d2b1c7f`: the MoonBoard clear call now omits `lightAdjacentHolds` entirely; the shared encoder pins explicit `false` as byte-identical to the default packet; and tests now cover both an option change while disconnected reaching the first post-connect encoder call and preference hydration while disconnected followed by exactly one reconnect write. Focused checks: 116 web tests + 45 shared protocol tests passed. Codex Sol re-reviewed the exact head with no P0–P2 findings.

marcodejongh:
Addressed both final nonblocking Claude notes in `0c1a0faa0`: the MUI switch now applies the `checked` boolean emitted by `onChange` while the surrounding row still toggles normally, and the Swift bridge preserves `nil` when JS omits `lightAdjacentHolds` instead of eagerly storing `.some(false)`. The mutation-sensitive switch test, default/explicit-false behavior, and persisted-config compatibility all pass. Exact-head Codex Sol/Fable-equivalent review: PASS with no P0–P2 findings; 178 focused tests passed.

marcodejongh (approved):
Exact-head Codex Sol/Fable-equivalent follow-up PASS. The switch consumes MUI’s emitted checked state without double-toggle propagation; the native bridge preserves the optional nil sentinel while all consumers retain default-false behavior. No P0–P2 findings; 178 focused tests, scoped checks, and diff cleanliness passed.

Issue #4366, opened 2026-08-12
W-13: make /playlists read-only and the home page static


marcodejongh:
Part of #4358. **Blocked by:** #4365.

## Context

Two kept surfaces reach deep into the delete set and are trivially separable from the front-door work: the playlists pages and the home page. Both must be unwired before W-16 can delete anything, and the home page's board rail is currently not crawlable at all — which matters because the homepage is the strongest internal-link source www has.

### `/playlists`

- `playlists/library-page-content.tsx` imports `queue-control/queue-bridge-context`, four `library/*` components, and `board-scroll` + `board-selector-drawer` (lines 20-33). Cut all of them.
- `playlists/[playlist_uuid]/playlist-detail-content.tsx` imports **nine** delete-set modules (lines 71-80). **Cut the wall-activation path entirely** — that removes line 71 and lines 75-80 — and **keep** `MultiboardClimbList`, `PlaylistEditDrawer` and `CommentSection`.
- Replace activation with a link to `app.boardsesh.com/discover/{uuid}`. The destination route is verified to exist: `packages/mobile/app/(tabs)/discover/[playlist_uuid].tsx`.

### Home page

- `app/home-page-content.tsx` reaches **five** delete directories, **three of them via `next/dynamic`** (lines 29, 39-48) — invisible to a plain grep, and they fail only at runtime — plus `home-gym-card.tsx:22`.
- The board rail becomes a **static crawlable grid of `LocaleLink` cards** built from `getPopularBoardConfigs()`, which the page already SSR-fetches.
- **Use `constructClimbListWithSlugs(boardType, layoutName, sizeName, sizeDescription, setNames, angle)`, not `constructBoardSlugListUrl`** — `PopularBoardConfig` (`packages/shared-schema/src/types/board-entities.ts:156`) has **no `slug` and no `angle`**, so the slug-based builder cannot work here.
- "Start a sesh" becomes the existing `app/components/start-climbing-button.tsx`.

Files: `packages/web/app/playlists/{library-page-content,page}.tsx`; `packages/web/app/playlists/[playlist_uuid]/playlist-detail-content.tsx`; `packages/web/app/home-page-content.tsx`; `packages/web/app/components/home-gym-card/home-gym-card.tsx`.

### Cross-epic note

[...]

Issue #4362, opened 2026-08-12
W-07: thread layout/size/set ids into the slug URL generators


marcodejongh:
Part of #4358. **Blocked by:** **#3823** (it lands `resolveSizeSlug`, which this issue calls).

## Context

Two Kilter sizes collide in slug form. Verified in `packages/board-constants/src/generated/product-sizes-data.ts`: size **10** is `12 x 12 with kickboard` and size **27** is `12 x 12 without kickboard`, and **both** carry `description: 'Square'`, so both slug to `12x12-square`. Those are different boards — different `edgeBottom`, different holds.

#3823 fixes the **resolver** (`resolveSizeSlug` emits `12x12-square-without-kickboard`), but its own diff comment notes that `constructClimbViewUrlWithSlugs` stays name-based and "threading ids through here would let web emit the exact form too — worth doing when these call sites next get touched." That is now: #4369 builds every canonical URL and every "Climb this" CTA from these generators, so a bare-slug URL would send a visitor — and Google's canonical — to the **wrong board**.

## The work

- Thread `layoutId` / `sizeId` / `setIds` into **`constructClimbViewUrlWithSlugs`** and **`constructClimbListWithSlugs`**, and have them call `resolveSizeSlug` instead of deriving the size segment from the name/description.
- Update every call site to pass the ids. They are available at each call site — the callers already resolve the board config tuple.
- Keep the emitted URL shape otherwise identical; this is a correctness fix, not a URL-shape change.

Files: `packages/shared/play-view/src/readable-url-utils.ts`, `packages/web/app/lib/url-utils.ts`, `packages/web/app/lib/slug-utils.ts`.

Note that `packages/shared/play-view` is consumed by both web and mobile, so the mobile bundle changes too — see the standing-rule callouts.

## Verifies

- [ ] A **round-trip test over the whole catalogue**: for every `(boardName, layoutId, sizeId)` in `packages/board-constants/src/generated/product-sizes-data.ts`, `getSizeBySlug(board, layout, <emitted slug>) === sizeId`. This is the acceptance criterion — a spot check on Kilter 10/27 alone is not enough, because the same `description` collision pattern can exist on other boards.
- [ ] Kilter size 10 and size 27 emit **different** slugs, and each resolves back to its own id.
- [ ] No call site still builds a view or list URL from the bare name/description.

## Gates

**GW**:

- `vp check`
- `vp run typecheck`
- `vp test run --reporter=agent --project web`

[...]

Issue #3086, opened 2026-06-20
Climb list jumps around while loading — rows reflow as they measure on the main browsing surface


marcodejongh:
## What's happening

On the climb-list page (the main browsing surface), the page visibly jumps while it loads: rows are placed at an estimated height, then re-measure to their real height and push everything below the fold around, while the container grows. Almost half of loads on this route are degraded.

## Evidence

PostHog `$web_vitals` on `/[board]/.../[angle]/list` (30d): 465 CLS samples, p75 CLS 0.274 (above the 0.25 "poor" threshold), p90 0.621.
- 124/465 (27%) rated poor + 86/465 (18%) needs-improvement = 45% of loads degraded.
- Stable across 5 weeks (weekly p75 0.217-0.294; post-#2113 weeks 0.21-0.34), so structural, not a one-week regression.
- Route-specific: home page p75 CLS is healthy at 0.054 (15/445 poor, 3%) with MORE samples — ruling out project-wide inflation.
- Worst URL `/kilter/original/12x12-square/screw_bolt/40/list`: 119 samples, p75 0.275, p90 0.624, 33 poor (28%).
- Telemetry is real RUM: `packages/web/app/components/analytics-client.tsx` reports `web-vitals onCLS -> $web_vitals`.

_posthogUrl: https://www.boardsesh.com/kilter/original/12x12-square/screw_bolt/40/list

Platform: web.

## Suspected cause

The list virtualizer in `packages/web/app/components/board-page/climbs-list.tsx`: `useWindowVirtualizer` with a hardcoded `estimateSize: () => 107` (line ~574) plus `virtualizer.measureElement` (a ResizeObserver, line ~672). Rows have no fixed/min height — the thumbnail (~90px) + 16px padding + a wrapping setter/name line commonly exceeds 107px. On first paint every absolutely-positioned row sits at the 107px estimate and `getTotalSize()` = count*107; as each row mounts and `measureElement` reports the true height, the `translateY` offsets of all rows below shift and the container height grows — exactly the loading-time reflow CLS measures. Compounding: `initialRect {width:375,height:812}` lays out ~8 SSR rows at the estimate that re-measure on the client, and the `INITIAL_BATCH=6 -> requestAnimationFrame(setVisibleCount(all))` reveal (lines ~319-352) grows total height a frame after first paint. Thumbnail image dimensions themselves are reserved correctly (`BoardImageLayers` sets width/height + aspect-ratio), so the image is not the cause.

[...]

Issue #4367, opened 2026-08-12
W-13a: make /session read-only and delete the self-only smart-playlist route


marcodejongh:
Part of #4358. **Blocked by:** #4365. (W-00's usage pull is done, which is what settles the `/discover` deletion.)

## Context

`/session/[sessionId]` is a **kept** page — it is the public share surface with an OG card, and its mobile twin is auth-gated, so it cannot move to the app. But it is wired straight into the delete set and **no other PR unwires it**. Without this issue, the delete wave cannot typecheck.

The same PR deletes `/discover/[smart_playlist]/[user_id]`, which W-00 settled: over 90 days to 2026-07-31 the route drew 454 views from 126 users, and **every path in the top-30 breakdown has `users = 1`** except `crowd-favorites/3fa79140…` at 3. Nobody browses anyone else's smart playlist, so the mobile app's self-bound `discover/smart/[type]` is a complete replacement. Its scheduled deletion in W-19 sits *after* the PRs that break it, so it has to come forward to here.

### Make `/session/[sessionId]` static and read-only

`session-detail-content.tsx` imports, at lines 32-35, 43-46 and 56-57:

- `board-page/climbs-list`
- both `climb-actions` batch providers
- `use-climb-actions-data`
- `session-details/{session-overview-panel,edit-session-dialog}`
- `healthkit/save-to-healthkit-button`
- `graphql-queue`

Rework it:

- Render climbs through **`StaticClimbList`** (#4365) instead of `board-page/climbs-list`.
- **Drop** the HealthKit button and both `climb-actions` batch providers.
- **Lift** `session-overview-panel` + `edit-session-dialog` into a kept directory (the #4364 pattern — byte-identical move), **or** inline a read-only overview if the edit dialog is not worth keeping. These are the two `teardown:components-session-details` edges in W-05's allowlist; lifting them here is what lets W-16 delete the directory.
- Keep the OG card, the climb list and the comments working — this page is a public share target.

### Delete `/discover/[smart_playlist]/[user_id]`

`smart-playlist-content.tsx:29-40` imports `queue-control` ×4, `graphql-queue`, `library` and `climb-actions`. Delete the route and its content component outright. No repoint is needed — the mobile route replaces it.

Leave `discover/layout.tsx` for W-19, which deletes it once it is orphaned.

## Verifies

- [ ] `/session/{id}` SSRs with its OG card, its climb list and its comments intact.
[...]

PR #4421, opened 2026-08-14
feat(web): make every climb list row a real link

### LINKED ISSUE #4365: W-12: build StaticClimbList — the first crawlable climb links on www
Part of #4358. **Blocked by:** #4364, #4362. W-05 (#4110) is merged.

## Context

**There is not one crawlable link to any climb on www today.** `packages/web/app/components/climb-card/climb-list-item.tsx:568` is an `onClick` on a `<Box>` — a click handler, not an anchor — so Googlebot cannot reach any of the 771,159 listed climbs by following links, and neither can a screen reader user tabbing through a list.

This is the keystone of the front-door work: every kept surface that lists climbs (`/playlists/[uuid]`, `/setter/[username]`, `/profile/{id}/climbs`, `/discover/*`, and later the static `/list` and the climb front door) renders through it. On W-05's allowlist accounting: 6 edges carry `cutBy: W-12`, but this PR cuts **one** of them directly (`proposal-card → climb-list-item`) — the other five were tagged optimistically and are actually cut later (see Verifies for the re-labeling this PR must do).

## The work

**Build `packages/web/app/components/climb-list/static-climb-list.tsx`:**

- Virtualized — `@tanstack/react-virtual`, `LIST_ROW_HEIGHT`, `useInfiniteScroll` (the same primitives the current list uses; `LIST_ROW_HEIGHT` comes from the constants #4364 lifted to `app/lib/climb-list-constants.ts`).
- Reuses the **clean presentational parts** of `climb-card`: `climb-thumbnail`, `climb-title`, `climb-icons`, `climb-card-cover`, `marquee-text`. Those five are W-16 survivors; do not pull in `climb-list-item` or `climb-card`.
- **Every row is a `LocaleLink`** to the canonical config-tuple view URL — a real `<a href>`, not a click handler on a `<div>`.

**Repoint the consumers:**

- `components/climb-list/multiboard-climb-list.tsx` — it serves `/playlists/[uuid]`, `/setter/[username]`, `/profile/{id}/climbs` and `/discover/*`.
- `components/social/proposal-card.tsx:36` — currently `import ClimbListItem from '@/app/components/climb-card/climb-list-item'`. This is one of the four edges every designer missed; the proposal card renders inside `/gym/[gym_slug]`'s comment section.

**Change `AscentStatus` to take `logbook` as a prop** instead of reading `useOptionalBoardProvider` (`components/climb-card/ascent-status.tsx:12`). The static list has the logbook data server-side when the viewer is signed in; the prop is **optional** — anonymous SSR (crawlers, the front door) renders without a logbook. Note the mechanic: `ascent-status.tsx` currently has no allowlist entry (it's keep-classified but unreachable from keep roots, so the traversal never records its `board-provider` import). Making it reachable through `StaticClimbList` **would add a new violation** — the prop change prevents that addition. It is therefore non-optional in the same commit that first imports `climb-card-cover`.

**Add `buildAppHandoffUrl(pathname, locale)`** in a new `packages/web/app/lib/app-handoff.ts`: `APP_URL` + the pathname with the locale prefix stripped via `stripLocalePrefix` (`lib/i18n/locale-href.ts:18`). It supersedes `mapToExpoWebTarget` (deleted by W-09 #4363; if that hasn't landed yet the old helper is inert dead code — `buildAppHandoffUrl` is additive either way), which returned `null` for every `/b` and named-slug form. #4369 consumes it for the "Climb this" CTA.

### This PR deletes nothing

`climb-list-item.tsx` and `climb-card.tsx` still have **six live importers** that only die in W-16/W-17: `board-page/climbs-list.tsx:19,45` (line 45 is a `next/dynamic` import — invisible to grep), `queue-control/queue-climb-list-item.tsx:22`, `queue-control/queue-list.tsx:25,26`, `create-climb/drafts-drawer.tsx:14`, and `[board_name]/…/liked/liked-climbs-list.tsx:21`. Deleting either file here reds `main`.

Files: `components/climb-list/static-climb-list.tsx` (new); `components/climb-list/multiboard-climb-list.tsx`; `components/social/proposal-card.tsx`; `components/climb-card/{ascent-status,climb-card-cover}.tsx`; `packages/web/app/lib/app-handoff.ts` (new).

## Verifies

- [ ] `/setter/{u}`, `/profile/{id}/climbs`, `/playlists/{uuid}` and `/discover/*` all render.
- [ ] Every row in the SSR HTML is a real `<a href="/kilter/…/view/…">` pointing at the canonical config-tuple URL.
- [ ] Infinite scroll still pages.
- [ ] The proposal card still renders inside `/gym/[gym_slug]`'s comment section.
- [ ] `AscentStatus` renders from an optional `logbook` prop with no board provider mounted (and renders sensibly with the prop absent — the anonymous/crawler path).
- [ ] W-05's allowlist loses the `proposal-card → climb-list-item` edge (`cutBy: W-12`); **nothing is added to it**.
- [ ] The five remaining `cutBy: W-12` entries are re-labeled in the same PR to their true cutters — the three `liked/liked-climbs-list.tsx → climb-card/*` edges to W-17 (the `liked` route is deleted there, untouched here), and the two `board-provider-context` mounts (`b/[board_slug]/[angle]/layout.tsx`, `components/providers/persistent-session-wrapper.tsx`) to W-17/W-16 respectively (provider strip / site-chrome swap). Metadata-only allowlist edit; keeps the "allowlist empty by W-16" accounting truthful.

## Gates

**GW**:

- `vp check`
- `vp run typecheck`
- `vp test run --reporter=agent --project web`

**GI**:

- `vp run check:i18n`
- `vp run check:i18n:orphans`

Any new user-facing string goes into **every** locale catalog (`en-US`, `es`, `fr`, `de`) in the same commit — `catalog-completeness.test.ts` enforces parity per namespace, and the Spanish/French/German glossaries in `docs/i18n-*-glossary.md` apply.

## Standing-rule callouts

- **Standing rule 5** — `climb-card/` is never deleted as a directory. This PR deletes none of it.
- The canonical URL must come from #4362's id-threaded generators, not the bare slug, or rows on the two colliding Kilter sizes link to the wrong board.
- **Release Notes**: user-visible — climb lists become real links you can open in a new tab, share, and reach from search. Climber voice.
- Opus-tier implementation (novel component + graph-sensitive repoints), with all three QA lenses.

marcodejongh:
## Summary

Part of #4358.
Closes #4365.

There was not one crawlable link to any climb on www. `climb-list-item.tsx:568` is an `onClick` on a `<Box>` — a click handler, not an anchor — so Googlebot could not reach any of the 771,159 listed climbs by following links, and neither could anyone tabbing through a list.

This PR builds `StaticClimbList`: the same window-virtualized list, with every row a real `<a href>` to the canonical config-tuple `/view/` URL.

**What landed**

- **`static-climb-row.tsx`** — one crawlable row. `MuiLink component={LocaleLink}` with `prefetch={false}` (18 rows are in the virtualizer window at once; the App Router default would fire 18 route prefetches on a page that already carries its own data fetch). Exactly one interactive element per row: no swipe actions, no overflow menu, no drawers. Its geometry reproduces `climb-list-item`'s exactly — 64px thumbnail at aspect 5/7 plus 2 × 8px padding and a 1px border ≈ 107px — so a fresh row measures to exactly `LIST_ROW_HEIGHT` and nothing reflows beneath it (#3086).
- **`static-climb-list.tsx`** — the virtualized list. SSR anchors come from `useWindowVirtualizer({ initialRect: { width: 375, height: 812 } })`, the trick `board-page/climbs-list.tsx:578` already uses; without it `getVirtualItems()` returns `[]` on the server and the HTML carries zero links. Paging uses both the five-rows-early virtualizer effect and the `useInfiniteScroll` sentinel, matching today's feel. The row wrapper contains `layout style` but **not** `paint`, unlike the classic list: paint containment clips descendants to the wrapper, and the row is one big anchor whose focus ring is drawn outside its border box (`outline-offset: 2px`, `components/index.css:343`) — with `paint` the ring is clipped away and tabbing through the list shows nothing.
- **`buildCanonicalClimbViewUrl`** extracted from `getContextAwareClimbViewUrl` (zero behaviour change — same code path). Rows call it directly rather than the context-aware builder: on `/b/{slug}/{angle}/playlists/{uuid}` the latter would emit `/b/…/view/…` rows, and the epic's locked decision is that the config-tuple tree is canonical.
- **`app/lib/app-handoff.ts`** — `buildAppHandoffUrl(pathname)` for #4369's "Climb this" CTA. Purely additive: `mapToExpoWebTarget` is already gone at this base (`grep` returns zero matches), so there is no helper to supersede and no call site to migrate. It takes **no locale argument**, a deliberate change from the plan's two-arg signature: it strips every supported prefix the way `applyLocale` does, so a caller that forgets to thread the active locale can't hand out `app.boardsesh.com/es/…`, a route the app doesn't have. Query strings and fragments are dropped, which is what the doc comment always claimed.
- **`AscentStatus`** drops `useOptionalBoardProvider` for optional `logbook` / `boardName` props.
- **`multiboard-climb-list.tsx`** repoints onto the static list. Its sort toggle + count row used to travel as `headerInline` and land inside `ClimbsList`'s padded header container; it now carries that padding (`8px 12px`, 40px floor) itself, so on `/setter/{username}` and `/profile/{id}/climbs` it stays aligned with the rows below instead of sitting flush at x=0.
- **`ClimbTitle` grows `showGradeWhileLoading`**, which the crawlable row sets. The reader's grade format resolves from IndexedDB inside an effect, so `loaded` is false on the server and the right-hand grade renders as a `<Skeleton>` — the first crawlable rows on www would have shipped without the single most-searched climb attribute. With the flag the server emits the default-format grade (`V4`) and swaps to Font for readers who chose it. Nothing else passes the flag, so every existing surface keeps today's Skeleton.

**Two files, not one.** The issue's `Files:` list names one new component. The row is split out so `social/proposal-card.tsx` — which renders exactly one climb inside `/gym/[gym_slug]`'s comment section — can import it without pulling `@tanstack/react-virtual` and `useInfiniteScroll` into that page's bundle.

### Why the AscentStatus prop change is not optional here

`ascent-status.tsx` has no allowlist entry today: it is keep-classified but unreachable from any keep root, so the traversal never records its `board-provider` import. `StaticClimbRow` imports it directly to render the badge over the thumbnail, which makes it reachable — and would have **added** a `climb-card/ascent-status -> board-provider/board-provider-context` violation, which #4365's Verify forbids outright. The prop change prevents that addition.

The trigger is the `AscentStatus` import, not `climb-card-cover` as the issue says — a list row needs no card cover. `climb-card-cover.tsx` is still edited (prop pass-through) so it isn't left silently badge-less for #4369, which does render it.

`climb-card-cover.tsx`, `climb-card.tsx` and `climb-list-item.tsx` all thread `boardProvider?.logbook` / `boardProvider?.boardName` through, so the grid cards, the queue list, drafts and `liked` are byte-identical after the change. Both files that now read the provider on the badge's behalf are themselves delete-set, so traversal stops before them and the new edges cannot enter the allowlist.

Moving that read up from the leaf widens the re-render blast radius on the grid, so `ClimbCardCover` is now `React.memo`'d and its `boardStyle` is `useMemo`'d. Before, the only BoardContext subscriber under a card was the badge itself; now `ClimbCard` subscribes, and a fresh `style` object on every render would have defeated `BoardCanvasRenderer`/`BoardImageLayers`' own memos — every visible board re-rendering on each logbook fetch, tick save or auth settle.

### Accepted loss: ascent badges on the two board-scoped playlist routes

`/b/{slug}/{angle}/playlists/{uuid}` and `[board_name]/…/playlists/{playlist_uuid}` lose their ascent badges until W-15 (#4369) threads a server-side logbook into the `logbook` prop this PR creates for exactly that purpose.

This is accepted, not overlooked. Every alternative adds an allowlist edge: reading the provider from a kept file means `multiboard-climb-list` (or a new kept file) `-> board-provider/board-provider-context`, which the issue forbids. Routing through `useLogbook` from `@boardsesh/board-react` would be invisible to the import-graph test but dead in practice — the only `BoardAdapterProvider` mount on web is `board-provider/board-adapter.tsx:92`, itself delete-set.

The four global surfaces lose nothing: `persistent-session-wrapper.tsx:277` mounts `BoardProvider` only around the queue bar, not around page content, so `/playlists/{uuid}`, `/setter/{u}`, `/profile/{id}/climbs` and `/discover/*` render zero badges today.

### Accepted loss: the heart on liked climbs in these lists

`climb-list-item.tsx:506` fed `favorited` into `ClimbTitle`, which appends a ♥ to the subtitle. The static row passes no `favorited`, and the repoint drops `FavoritesProvider` from the multiboard tree — one of the seven allowlist edges the "no stale entry" rule forces out — so on `/playlists/[uuid]`, `/setter/[username]`, `/profile/{id}/climbs` and `/discover/*` a signed-in climber can no longer tell from the list which climbs they've liked.

Everything else on the row survives: thumbnail, name, `ClimbIcons`, the sends · stars · setter subtitle and the colourised grade. Re-adding the heart means re-adding the `multiboard-climb-list -> climb-actions/favorites-batch-context` edge, so this rides along with W-15 (#4369), the same item that threads a server-side logbook back in for the ascent badges.

### This PR deletes nothing

`climb-list-item.tsx` keeps **5** live importers and `climb-card.tsx` keeps **2** — seven import sites across five files, all verified at this base:

| File | Line | Imports |
|---|---|---|
| `[board_name]/…/liked/liked-climbs-list.tsx` | 21 | `climb-card` |
| `[board_name]/…/liked/liked-climbs-list.tsx` | 22 | `climb-list-item` |
| `components/board-page/climbs-list.tsx` | 19 | `climb-list-item` |
| `components/board-page/climbs-list.tsx` | 45 | `climb-card` (`next/dynamic` — invisible to grep) |
| `components/create-climb/drafts-drawer.tsx` | 14 | `climb-list-item` |
| `components/queue-control/queue-climb-list-item.tsx` | 22 | `climb-list-item` |
| `components/queue-control/queue-list.tsx` | 25 | `climb-list-item` |

(The issue's line numbers had drifted: `queue-list.tsx:26` is `drawer-climb-header`, not a second climb-card import, and `liked-climbs-list.tsx:21` is `ClimbCard` with `:22` the list item.) Deleting either file still reds `main`. Standing rule 5 holds — no part of `climb-card/` is removed.

### Allowlist ledger: 123 → 116, zero additions

The issue anticipated **one** deletion. The multiboard repoint cuts six more live edges, and `import-graph-invariant.test.ts:470` ("has no stale allowlist entry") fails on any entry that no longer matches a real edge — so seven had to go in the same commit:

- `social/proposal-card.tsx -> climb-card/climb-list-item.tsx` (the `cutBy: W-12` edge the issue names)
- `climb-list/multiboard-climb-list.tsx -> board-page/climbs-list.tsx`
- `climb-list/multiboard-climb-list.tsx -> climb-actions/favorites-batch-context.tsx`
- `climb-list/multiboard-climb-list.tsx -> climb-actions/playlist-activation-context.tsx`
- `climb-list/multiboard-climb-list.tsx -> climb-actions/playlists-batch-context.tsx`
- `climb-list/multiboard-climb-list.tsx -> graphql-queue/index.ts`
- `climb-list/multiboard-climb-list.tsx -> hooks/use-climb-actions-data.tsx`

Five `cutBy` values relabelled to their true cutters (metadata only, `reason` strings untouched): the three `liked/liked-climbs-list.tsx -> climb-card/*` edges to **W-17** (the `liked` route is deleted there), `b/[board_slug]/[angle]/layout.tsx -> board-provider-context` to **W-17** (provider strip), and `providers/persistent-session-wrapper.tsx -> board-provider-context` to **W-16** (site-chrome swap).

The `$comment` at the top of the file said "cutBy is the PR id when it is known (W-12 cuts the climb-card interactive trio and BoardProvider)" — false the moment those five are relabelled, so it is corrected in the same edit. It now spells out that W-17 deletes the three **edges** into the climb-card trio along with the `liked` route that carries them, while the trio's own **files** die in W-16 with the rest of the climbing UI and the last BoardProvider mount.

**Zero additions**, and **zero `cutBy: W-12` labels remain**.

The two multiboard entries that survive are intentional: `board-scroll/board-filter-strip.tsx` is the surface's board picker, and `persistent-session/persistent-session-context.tsx` is what makes a playlist preview render on the wall the climber is actually on. Both die with their directories in W-16; cutting them here would change rendering for signed-in climbers for no crawlability gain.

### Sequencing notes for W-13 / W-13a

- The wall-activation and add-to-queue affordances leave `/playlists`, `/setter`, `/profile/{id}/climbs` and `/discover/*` **here**, not in #4366. `MultiboardClimbList` drops `useOptionalPlaylistActivation`, which leaves two mounts with no consumer under them — `playlist-detail-content.tsx:688` and `smart-playlist-content.tsx:307`, plus the `usePlaylistClimbActivation` calls at `playlist-detail-content.tsx:380` / `smart-playlist-content.tsx:182` that feed them. Harmless but inert: the only other consumer, `climb-actions/actions/set-active-action.tsx:23`, sits in an action sheet the static list never mounts. **Deliberately deferred**, not missed — #4366 already lists those import lines for deletion, and cutting the consumer first is what makes them safely removable there. Their two allowlist entries (`playlist-detail-content -> playlist-activation-context`, `smart-playlist-content -> playlist-activation-context`) therefore stay: they are still live edges, so "has no stale allowlist entry" is correctly green either way.
- `onClimbSelect` and `selectedClimbUuid` are dropped from `MultiboardClimbListProps` — verified zero callers pass either (`setter-climb-list.tsx:75`, `user-climb-list.tsx:60`, `playlist-detail-content.tsx:516`, `smart-playlist-content.tsx:248`).
- `StaticClimbList`'s prop surface is designed for both downstream consumers, including W-13a's `renderItemExtra` for tick-detail rows that legitimately exceed the 107px floor — which is why the row wrapper keeps `ref={virtualizer.measureElement}`. A block comment at the top of the file names both so a later editor doesn't narrow it.
- `usePersistentSessionState` stays: it's a rendering concern (which board a playlist climb previews on), not a row-interaction one, and W-16 removes it with its directory.

### On the "every row in the SSR HTML" verify

Literally true where the page SSRs climb data. `/playlists/{uuid}` does. `/setter/{username}` and `/profile/{id}/climbs` fetch through `useInfiniteQuery` (`setter-climb-list.tsx:26`, `user-climb-list.tsx:23`) and carry zero climbs in the SSR HTML today, so their anchors appear post-hydration; `/profile/[user_id]/climbs/page.tsx:15` is `robots: { index: false }` anyway. What this PR owns is the component-level guarantee — first render, server or client, emits anchors — and `static-climb-list-ssr.test.tsx` is the direct oracle for it. Page-level SSR data is W-15's job.

## Release Notes

- Climb lists are real links now — open a climb in a new tab, share it straight from the list, or land on it from a search.

## Test plan

**GW** — all green:

- `vp check` — no findings in any changed file (the two formatting hits are `docs/design/web-reposition/*.html`, untouched here and already like that on `main`)
- `vp run typecheck` — clean
- `vp test run --reporter=agent --project web` — **494 files, 6040 tests, 0 failures**

**GI** — all green, with **zero catalog edits**:

- `vp run check:i18n` — 504 `.tsx` scanned, no hardcoded strings
- `vp run check:i18n:orphans` — 4558 keys, every one live

No new i18n keys by design. The list reuses the existing `climbs:list.noMore`; the row deliberately carries no `aria-label` and no `title`, so its accessible name is its visible text (climb name, icons, setter, grade) — better a11y than an `aria-label` override, better SEO as real anchor text, and it dodges a live trap: `/gym/[gym_slug]/page.tsx:150` seeds only `['common','boards','kiosk']` and `/setter/[setter_username]/page.tsx:65` only `['profile']`, so a `climbs`-namespace string would SSR as a raw key there.

**New tests**

- `climb-list/__tests__/static-climb-row-url.test.tsx` — the #4362/W-07 guard. Kilter layout 1 `size_id: 27` emits `…/12x12-square-without-kickboard/…`, `size_id: 10` emits `…/12x12-square/…`. Without this, a row on the shadowed Kilter size links to the wrong physical board. Also asserts the href uses `climb.angle`, and that the row renders exactly one `role="link"` with no nested button.
- `climb-list/__tests__/static-climb-list-ssr.test.tsx` — `renderToString` with the virtualizer **unmocked**, asserting ≥8 `href="/kilter/` occurrences and the exact canonical URL for climb 0. Carries the negative control that makes it non-vacuous: the same virtualizer options minus `initialRect` produce **zero** server-side items, so deleting that option fails the test rather than silently killing SSR.
- `climb-list/__tests__/static-climb-list-virtualization.test.tsx` — 14 cases: `count`, `overscan === 10`, `estimateSize() === LIST_ROW_HEIGHT`, the `initialRect` constant, `[data-index="0"]` `minHeight === '107px'`, fewer rows rendered than climbs exist, `renderItemExtra` landing inside the measured wrapper, and the four `onLoadMore` gates (fires near the end; not while fetching, not without `hasMore`, not on an empty list).
- `climb-list/__tests__/multiboard-climb-list.test.tsx` — delete-safety oracle. Three climbs render three anchors matching `/^\/(kilter|tension)\/.+\/view\//`, the cross-board override resolves per climb, the empty state still renders, and the sort toggle still fires. There is deliberately **no** `vi.mock` for `climb-actions/*`, `graphql-queue` or `board-page/climbs-list` — the file no longer imports them, and their absence is the signal.
- `climb-list/__tests__/static-climb-row-ssr.test.tsx` — what the crawler actually gets. `renderToString` runs no effects, which is exactly the server's situation, and the HTML has to carry the grade (`V4` from `6A+/V4`), the name, the setter and the send count, with no `MuiSkeleton` anywhere. Delete `showGradeWhileLoading` from the row and these go red.
- `climb-card/__tests__/climb-card-ascent-badge.test.tsx` — the other half of the badge-parity guard, for both `ClimbCard` variants (self-built actions and caller-supplied). The cover mock records its props, so dropping `logbook` or `boardName` from either variant fails instead of silently emptying the grid's badges.
- `lib/__tests__/app-handoff.test.ts` — 8 cases: every supported prefix stripped (not just one), `/fr` → app root, the `/b` form `mapToExpoWebTarget` used to return `null` for, single leading slash, query/fragment dropped, no `//` pathname, idempotence.

**Rewritten**

- `climb-card/__tests__/ascent-status.test.tsx` — every case moves from `BoardContext.Provider` to prop-passing; all six behavioural assertions unchanged. The cache-reactivity case still feeds `useLogbook(...).logbook` straight into the prop, so it still proves React Query updates reach the badge. **New case**: no `logbook` prop and no provider anywhere renders empty markup — the anonymous/crawler path the issue's Verify calls out.
- `social/__tests__/proposal-card.test.tsx` — mocks `static-climb-row` instead of `climb-list-item`; the `disableSwipe` assertion becomes a pathname pass-through assertion, since the row has no swipe.
- `climb-card/__tests__/climb-list-item.test.tsx` — its `AscentStatus` mock swallowed the props, and its board-provider mock was hard-wired to `null`, so the badge threading was asserted by nothing: the mock now records props, the provider mock is settable, and two cases cover signed-in (ticks and board reach the badge) and signed-out (neither does).

**Existing, green without edits**

`app/__tests__/import-graph-invariant.test.ts` — all six cases, in particular "has no unallowed edge" (nothing added) and "has no stale allowlist entry" (which is exactly what forced the seven deletions). `climbs-list-virtualization.test.tsx`, `climb-title.test.tsx` (the Skeleton cases still pass — nothing else sets the new flag) and `queue-climb-list-item.test.tsx` all stay green, proving the AscentStatus threading kept the delete-set callers behaviour-neutral. `crawler-classic-invariant.test.ts` is untouched (single-owner W-09).

**Not run:** no dev server, so the manual pass (curl `/playlists/{uuid}` for one anchor per SSR'd climb, cmd-click a row on `/setter/{u}` and `/gym/{slug}`, confirm infinite scroll still pages, measure a real `[data-index="0"]` at 107px) is outstanding and belongs with the QA lenses.


marcodejongh:
Review follow-up in 8bf0a13:

1. **`centerBottomSlot` dead code** — fixed, dropped. `StaticClimbList`'s outer `renderItemExtra` is the slot W-13a (#4367) actually consumes, and `proposal-card` renders the row bare, so it was an abstraction ahead of a caller.

2. **`handleSortChange` lost its `useCallback`** — not the case. It is a bare arrow function on `main` too (`multiboard-climb-list.tsx:182` at `origin/main`), and this PR's diff contains no `+`/`-` line touching it. Nothing regressed; leaving it as-is rather than making an unrelated change in a graph-sensitive PR.

3. **Two concurrent `onLoadMore` triggers** — real, accepted, now documented on the prop. Worth being precise about the divergence the review only half-caught: `board-page/climbs-list.tsx` attaches the sentinel ref *only* in grid mode and runs the virtualizer effect *only* in list mode, so exactly one path is live there. `StaticClimbList` has no grid mode, so both run. Both gate on `!isFetching`, and every current caller is React Query's `fetchNextPage`, which drops a second call while the first is in flight — so it is safe today. The `onLoadMore` JSDoc now states the idempotence requirement so a future non-React-Query caller does not have to rediscover it.

Full gates re-run after the change: `vp run typecheck:web` clean, targeted web suite 223/223.

Issue #2232, opened 2026-05-20
[Feat] Climb: Support routes/circuits


ES-Alexander:
Kilter has long had a variable-speed animation system for progressively displaying part of a long climb or circuit. I suspect other boards have similar features, but I haven’t confirmed that.

It would be useful to be able to both play and create such routes in Boardsesh. They’re great for endurance training, local sequencing practice, and speed management training, and it’s a shame to not have access to them. As I understand it that requires defining a beta sequence/ordering for the holds, so holds in the same region can be lit up at different times (e.g. different holds the second time around the board), which may even allow the same hold to be included multiple times throughout the sequence.

On the circuits front, I feel like it would be nice to support both loop and bounce animation styles (where bounce goes back where you came from when you get to the end, whereas loop jumps from the end back to the start).

A brief recording of the feature in the current Kilter app:

https://github.com/user-attachments/assets/311979b0-0b6d-4b69-a982-7a3fdb3b6c16

TheStu:
I'd love to see this added!

marcodejongh:
Forgot to close this, webapp & current android/ios app now supports it!
STill need to add it to the react native rewrite

PR #4422, opened 2026-08-14
feat: support multi-frame route/circuit creation

### LINKED ISSUE #3785: [Feat] Climb: Create: Allow making routes/circuits
#2232 covers allowing route playback/step-through, but we can't yet create new ones.

Minimally we'll need options to add a new frame (which should likely just duplicate the in-view frame and insert it next), as well as a display of the current frame number+count, and buttons to step forward/backwards.

Nice to haves could include a play button, a delete frame button, and buttons to jump to the start/end frame, but those aren't technically essential.

I'd suggest the interface starts out pretty much as-is, with just a button added to duplicate the current frame[^1], and then once there's more than one frame included the rest of the route interface elements can be displayed.

Noting that remixing existing routes is also currently broken (it compresses them into a single frame, which may cause substantial issues if a route contains multiple different uses of the same hold - e.g. as both a start and finish).

[^1]: Material's [video-frame-copy](https://fonts.google.com/icons?selected=Material+Symbols+Outlined:video_frame_copy:FILL@0;wght@400;GRAD@0;opsz@24&icon.query=frame&icon.size=24&icon.color=%23e3e3e3) may be suitable

fallaciousreasoning:
## Summary

Closes #3785.

The create-climb editor could only ever produce single-frame boulders — the shared hold-state machine (`useCreateClimb`, used verbatim by web and mobile) held exactly one `LitUpHoldsMap` and had no way to encode more than one frame. Forking or editing an existing multi-frame route/circuit silently collapsed it into one frame (web fork took only frame 0; every other seed path unioned every frame's holds, later frames overriding earlier ones) — the bug the issue calls out.

Playback of multi-frame routes has worked since #2232; this closes the gap on the write side, per the issue's suggested MVP: the editor still starts single-frame, a **duplicate frame** button is always available, and once a climb has more than one frame the rest of the route UI (counter, prev/next, delete) shows up.

**Scope:** Aurora boards only (Kilter/Tension/decoy/touchstone/grasshopper/soill). MoonBoard's web editor is a separate, colour-based single-frame hook with no `frames` concept — unrelated, untouched. Mobile's create screen already routes every board (including MoonBoard) through the generic Aurora-style hook, so MoonBoard route creation on mobile is a side effect, not a target.

### What changed

- **`packages/board-constants`**: new `encodeMapsToFramesString` — the write-side counterpart to the existing `accumulateFramesToMaps` decoder. Frame 0 encodes absolute; later frames delta-encode against the previous frame (adds/removes/role-changes), matching the documented Aurora grammar. A single-frame route encodes byte-identical to the old flat format.
- **`packages/shared/create-climb-react`**: `useCreateClimb` now holds the full frame sequence + an active index instead of one map. `litUpHoldsMap` still means "the active frame," so every existing single-frame consumer is unaffected. New: `duplicateFrame`, `deleteFrame`, `nextFrame`/`prevFrame`/`goToFrame`, `loadFrames`, `frameCount`, `currentFrameIndex`, `currentFrameBleString` (BLE-ready single-frame string for the live-preview-while-painting path — never multi-frame syntax). Undo/redo now restore both frame content and which frame was active when the edit happened. `buildInitialFrames` replaces `buildInitialHoldsMap`, preserving frame separation on fork/edit/draft-restore instead of flattening.
- **Web + mobile**: new duplicate-frame toolbar button (always visible for Aurora boards), and — once `frameCount > 1` — a frame counter with prev/next stepper and a delete-frame button. Save payloads now report the real `frameCount` instead of a hardcoded `1`. Local autosave/crash-recovery persists the full frame array, not just the active frame.
- The live Bluetooth preview that fires while painting (both platforms) now sends only the active frame's BLE-ready string, never the multi-frame route string — the wall mirrors what you're currently painting, not an attempt to animate the whole route. **This touches Bluetooth code and needs Fable's review per CLAUDE.md before merge.**

## Release Notes

- Build routes and circuits, not just single-frame boulders — duplicate a frame, edit the copy, and step through your sequence as you build it.
- Forking or editing an existing route no longer collapses it into one frame.

## Test plan

- `vp check`, `vp run typecheck` (web/mobile/shared) — clean.
- `vp test run --reporter=agent` (full monorepo suite): 19643 passed, 11 skipped, 9 failed — all 9 failures are pre-existing and unrelated (`session-utils.test.ts`, `logbook-tab-grouped.test.tsx`, `feed-time-buckets.test.ts`; confirmed via `git diff main` that none of those files are touched by this branch — timezone-dependent flakes in this sandbox).
- `vp run check:i18n`, `vp run check:i18n:orphans`, `vp run check:mobile-bundle` — clean.
- New/updated unit tests: `encodeMapsToFramesString` round-trip tests (including against the real Kilter Grips oracle fixture) in `packages/board-constants`; full reducer test suite (duplicate/delete/nav/undo-redo-with-frames) in `packages/shared/create-climb-react`; updated web/mobile consumer tests for the new hook shape and save payloads.
- Manually verified against a real seeded 5-frame Kilter route (`board_climbs` row with `frames_count=5`): decoded via `accumulateFramesToMaps`, re-encoded via `encodeMapsToFramesString`, round-trips to an identical structure. Loaded the climb's view page (200, plays back correctly) and the create-fork page seeded from its `frames` string (200, no server error, new "Duplicate frame" control present in the rendered HTML).
- Did not have interactive browser tooling available in this environment to click through the new UI end-to-end — the QA notes at `.boardsesh/qa-notes.md` cover the manual flow (duplicate → edit → step → delete → save → verify playback; fork/edit a real multi-frame climb; BLE live-preview; mobile undo/redo; autosave restore) for follow-up manual QA.

### DIFF
--- a/packages/board-constants/src/__tests__/hold-states.test.ts
+++ b/packages/board-constants/src/__tests__/hold-states.test.ts
@@ -8,6 +8,7 @@ import {
   parseFramesSegments,
   accumulateFramesToMaps,
   accumulatedMapsToFrameStrings,
+  encodeMapsToFramesString,
   flattenFramesToUnion,
   isSentinelHoldState,
   toFlatFrames,
@@ -316,6 +317,51 @@ describe('accumulatedMapsToFrameStrings', () => {
     ];
     const [string0] = accumulatedMapsToFrameStrings(maps, 'moonboard');
     expect(string0).toBe('p100r42');
+  });
+});
+
+describe('encodeMapsToFramesString', () => {
+  it('returns the empty string for an empty maps array', () => {
+    expect(encodeMapsToFramesString([], 'tension')).toBe('');
+  });
+
+  it('encodes a single frame identically to the legacy flat format', () => {
+    const maps = accumulateFramesToMaps('p100r1p200r2', 'tension');
+    expect(encodeMapsToFramesString(maps, 'tension')).toBe('p100r1p200r2');
+  });
+
+  it('round-trips adds, removes, and role changes across frames', () => {
+    const original = 'p100r1p200r2,"x100p300r2,"p300r3';
+    const maps = accumulateFramesToMaps(original, 'tension');
+    const reEncoded = encodeMapsToFramesString(maps, 'tension');
+    expect(accumulateFramesToMaps(reEncoded, 'tension')).toEqual(maps);
+  });
+
+  it('encodes two identical consecutive frames as a bare hold frame', () => {
+    const maps = [
+      { 100: { state: 'STARTING' as const, color: '#00FF00', displayColor: '#00FF00' } },
+      { 100: { state: 'STARTING' as const, color: '#00FF00', displayColor: '#00FF00' } },
+    ];
+    expect(encodeMapsToFramesString(maps, 'tension')).toBe('p100r1,"');
+  });
+
+  it('round-trips the Kilter Grips oracle fixtures', () => {
+    for (const climb of oracle.climbs) {
+      const maps = accumulateFramesToMaps(climb.auroraFrames, 'kilter');
+      const reEncoded = encodeMapsToFramesString(maps, 'kilter');
+      expect(accumulateFramesToMaps(reEncoded, 'kilter')).toEqual(maps);
+    }
+  });
+
+  it('drops holds whose state has no canonical code on this board', () => {
+    const maps = [
+      {
+        100: { state: 'STARTING' as const, color: '#00FF00', displayColor: '#00FF00' },
+        200: { state: 'FOOT' as const, color: '#FFAA00', displayColor: '#FFAA00' },
+      },
+    ];
+    // MoonBoard has no FOOT canonical code.
+    expect(encodeMapsToFramesString(maps, 'moonboard')).toBe('p100r42');
   });
 });
 
--- a/packages/board-constants/src/hold-states.ts
+++ b/packages/board-constants/src/hold-states.ts
@@ -295,6 +295,60 @@ export function accumulatedMapsToFrameStrings(maps: LitUpHoldsMap[], board: Boar
 }
 
 /**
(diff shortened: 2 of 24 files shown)


fallaciousreasoning:
**Fable Bluetooth-safety review** (mandatory per CLAUDE.md — this PR switches the live-preview BLE send from `generateFramesString()`, now multi-frame-aware, to a new `currentFrameBleString()` that encodes only the active frame):

**Verdict: safe to merge from a Bluetooth-safety standpoint. No blocking issues.**

- `currentFrameBleString()` correctly delegates to the pre-existing, unmodified `accumulatedMapsToFrameStrings` — flat single-frame format, no multi-frame syntax.
- Confirmed `getAuroraBluetoothPacket` (`packages/shared/ble-protocol/src/aurora.ts`) would silently light wrong/no holds if fed a comma/`"`/`x`-bearing multi-frame string — validating the fix was necessary.
- Both live-preview call sites (web `create-climb-form.tsx`, mobile `use-create-climb-screen.ts` — the debounced preview effect *and* `handleToggleBle`'s `bluetooth.connect(...)`) verified switched to `currentFrameBleString()`, with correct `useEffect`/`useCallback` dependency arrays.
- Repo-wide grep of every other `sendFramesToBoard`/`bluetooth.connect(` call site confirms none of them are touched by this PR or consume the new/changed hook methods.
- Remaining `generateFramesString()` usages are all save-payload / queue-record paths (not direct BLE sends) — correctly unchanged.
- No other files under `packages/shared/ble-protocol/`, `packages/mobile/src/lib/ble/`, or `packages/mobile/modules/live-activity/ios/` touched by this PR.

Full findings in the review transcript, happy to share if useful.

marcodejongh (approved):

fallaciousreasoning:
There's an issue on master which is preventing creating new problems (you can't tap holds on Android). Going to open a separate branch to look at this.

Issue #4359, opened 2026-08-12
QA gate: real-device sheet QA on expo-web (blocks the delete wave)


marcodejongh:
Part of #4358. **Blocked by:** none — parallel-safe. It needs a maintainer holding two physical phones, so start it now: it is the longest-lead item in the epic.

## Context

The web reposition is a **hard delete** with no retained `?classic=1` runtime fallback. Once `packages/web`'s interactive climbing surfaces are gone, the React-Native-web sheets on `app.boardsesh.com` are the only implementation left — there is no path back short of redeploying the `release/classic-web` artifact. `@gorhom/bottom-sheet`'s web fallback does not implement the gesture-lock and keyboard contracts the native sheets rely on (see the "Expo web" section of `CLAUDE.md`), so those two sheets have to be proven on real hardware before anything is deleted.

This gate is transcribed from the "Phase A0 — blocking pre-delete QA gate (real devices)" section of `docs/web-reposition.md`. Sign-off is recorded by ticking every box below in this issue.

## The work

Run the checklist on **real hardware** — a physical iPhone (iOS Safari) and a physical Android phone (Chrome). **Emulators and simulators do not count.** Run it on **both origins**: the Next-embedded `/app` and standalone `app.boardsesh.com`.

- [ ] **Log ascent (`LogAscentSheet`)** opens from a climb, the on-screen keyboard does not cover the notes/grade fields, the sheet doesn't jump or dismiss on focus, and submit records the tick (verify it appears in the logbook).
- [ ] **Queue (`QueueSheet`)** opens, scrolls its virtualized list without the whole sheet scrolling, reorders, and the gesture-lock holds (drag inside the list doesn't dismiss the sheet).
- [ ] Rotate the device with each sheet open — no layout break or stuck backdrop.
- [ ] Repeat both on `app.boardsesh.com` (standalone) — it has no `?classic` escape hatch after the delete wave, so a failure here blocks the whole delete.
- [ ] Record crash-free confirmation for each sheet × device × origin.

Full matrix to record (2 devices × 2 sheets × 2 origins):

- [ ] iPhone Safari · `LogAscentSheet` · `/app`
- [ ] iPhone Safari · `QueueSheet` · `/app`
- [ ] iPhone Safari · `LogAscentSheet` · `app.boardsesh.com`
- [ ] iPhone Safari · `QueueSheet` · `app.boardsesh.com`
- [ ] Android Chrome · `LogAscentSheet` · `/app`
[...]

PR #4430, opened 2026-08-14
feat(web): SSR climb and list front doors on both trees

### LINKED ISSUE #4369: W-15: ship the SSR climb and list front doors on both trees

marcodejongh:
## Summary

**GATE-HELD: do not merge until QA gate #4359 is signed.** This PR is the moment
production loses the interactive climbing surface on www, so it stays a draft
until real-device QA on the app side passes.

Part of #4358
Closes #4369

Replaces `BoardPageClimbsList` — the interactive, queue-bearing surface — with
server-rendered front doors on all four board surfaces: the climb view and
`/list`, on the config-tuple tree and on `/b/{slug}`.

**The climb front door** (`app/components/climb-front-door/`) renders a
breadcrumb, the promoted `ClimbViewSeoFragment` heading (no longer
`visuallyHidden` — with the drawer gone there is nothing left to hide behind, and
it is now the page's only `<h1>`), the board art as a plain `` at the
`/api/internal/board-render` overlay URL with explicit dimensions, a facts `<dl>`,
a setter link, angle cross-links, beta videos, similar climbs, the community
section, and one CTA. Not `BoardRenderer` or `BoardImageLayers`: both are
hook-bearing client components, and this image is the page's LCP.

**The `/list` front door** renders `StaticClimbList` with a new
`virtualize={false}` mode. The virtualized path emits a 375×812-worth of rows on
the server (~18 anchors) and the bar here is ≥50 crawlable climb links per
`?page`. Existing virtualized consumers (`multiboard-climb-list`,
`session-detail-content`) are untouched, and a test asserts the default path is
still windowed — if that ever matched the static count, `virtualize` would have
stopped meaning anything.

**The A1 canonical flip lands with it.** `/b` canonicalises INTO the config-tuple
tree via `buildCanonicalClimbViewUrl` / the new `buildCanonicalClimbListUrl` — the
same functions that tree's own pages call, which is what makes "both trees emit
the identical canonical" a fact rather than a hope. The legacy `/list` page's
`generateMetadata` stopped returning a bare `{}` and claims its self-canonical,
because with `/b` no longer self-canonicalising, a second canonical no longer
splits anything.

Both `/list` pages resolve the request's slugs to ids and build from THOSE.
Several spellings resolve to one board config — `kilter-original` and `original`
are both layout 1, `12x12` and `12x12-square` both size 10, `bolt_screw` and
`screw_bolt` both sets [1,20] — so echoing the request's own segments back would
give every spelling its own self-canonical plus a four-locale hreflang cluster,
and none of them the URL `/b` points at. `parseRouteParams` is React-`cache`d and
the page body already calls it, so the resolution is free.

**One indexation doctrine, stated once**, in `resolveListPageIndexation`
(`app/lib/seo/list-page-robots.ts`), called by both `/list` trees:

- **Filter/sort variants** (`?minGrade=`, `?sortBy=`, …) are `noindex, follow`
  and canonicalise onto the clean base URL — CLAUDE.md's SEO rule, and the same
  page of content behind a different query.
- **Pagination is self-canonical, always.** `?page=3` is different climbs, not a
  duplicate of page 1, so pointing its canonical at the base is the classic
  pagination anti-pattern. Past page 10 it goes `noindex, follow` and still
  canonicalises to itself — a self-canonical carries no conflicting signal.
- **The `?page` walk is bounded at both ends.** The `next` anchor stops at page
  10 rather than at "more climbs exist": `noindex, follow` past the cap is an
  explicit instruction to keep following links, so a chain gated on `hasMore`
  would hand a crawler a corridor thousands of pages deep, each hop a deeper
  `OFFSET` over the climb/stats join (worst on MoonBoard, where
  `cachedSearchClimbs` bypasses the server cache outright). Pages 11–20 still
  render for a hand-typed or externally-linked URL — `noindex, follow`,
  self-canonical, with a `prev` anchor back into the indexable set — and
  `?page=21` and beyond 404s before the climb-list query runs (on the config-tuple tree before any query; on `/b` the slug must resolve first, so `resolveBoardBySlug` still fires once before the 404). `parseFrontDoorPage` clamps
  to the same ceiling as defence in depth, so no caller can turn a raw `?page`
  into an unbounded offset by forgetting the range check.
- **A hidden board is the one case that emits no canonical at all.** An unlisted
  or private `/b` board passes **no `path`**, so `createPageMetadata` emits no
  `alternates`. It is the only case where the noindex is about the board rather
  than a query variant, and where the target is a *public* config-tuple URL — a
  canonical from a noindex URL onto that twin is the conflicting signal Google
  can resolve by propagating the noindex, deindexing a public climb page because
  one private board shares its configuration. Same rule on the `/b` view page.

Asserted in `climb-canonical-parity.test.ts` (now covering `/list` as well as
the climb view, both including the shadowed Kilter layout-1 size 27) and in both
page-metadata suites. The parity mocks derive board details from the ids each
tree passes rather than sharing one frozen object, so the equality is earned
rather than structural.

**No server-side logbook or ascent badges on any front door, deliberately.**
`middleware.ts:245-251` puts a shared `Vercel-CDN-Cache-Control: s-maxage` on
every climb-view and `/list` URL with no session-cookie check, on the documented
premise that these pages are personalization-free. Server-rendering one viewer's
ticks would cache them for every later visitor and for Googlebot.
`ascent-status.tsx:20-21` states the same contract from the other end: anonymous
SSR renders no badge at all. If personalised badges are wanted later, the only
safe shapes are a post-hydration client island, or middleware emitting
`private, no-store` when a session cookie is present — which costs cache hit rate
on the highest-traffic pages. Both are out of scope.

**The Boardsesh-grade section is held out** (the spec's recommendation (b)). That
flag is a live staged PostHog rollout; SSR-ing the section would end it for every
visitor and crawler in a single deploy, irreversibly without another one.
`useBuildClimbDetailSections`'s flag-gated branch is untouched.

**Recorded on the gate** (#4369's DoD asks for the decision, not just the
behaviour): https://github.com/boardsesh/boardsesh/issues/4359#issuecomment-5296608211
— this PR holds the section out, the rollout continues unchanged, and the call
to SSR it for everyone stays with Marco at gate-signing time. The two sign-off
checkboxes on #4359 are deliberately left for that moment; they are a
maintainer's signature, not something this PR gets to tick for itself.

### The CTA

`buildAppHandoffUrl(pathname)` — one argument, a clean pathname, **no UTM
parameters**. The builder's contract is "the same pathname on the app origin", and
a query string would break the app's route match. Attribution rides as event
properties on the new `Climb Handoff Clicked` (added to `SHARED_EVENTS` next to
`Board Route Handoff`, whose pre-announcement comment is updated to the shipped
contract). `environment: 'production-web'` is a **per-event property, not a
session super property**: PostHog resolves per-event over super, so the tag reaches
this funnel without reclassifying `$pageview`, `$web_vitals` and the ~120 other
www events that have never carried it — and it stays correct if PR #4299 later
registers an `environment` super property.

The element is a real server-rendered `<a href>`, so it is crawlable and works
with JS off.

**The click handler holds the navigation for a bounded flush**, because a
fire-and-forget capture here would never arrive. `posthog-js-lite` batches
through `@posthog/core` (flush at 20 events or every 10s), and neither bundle
registers a `pagehide`/`beforeunload` handler or uses `sendBeacon` — so an event
captured in the `onClick` of a cross-origin `<a>` dies in the queue with the
document about a millisecond later. PostHog would then show ~zero
`Climb Handoff Clicked` against a healthy app-side `Board Route Handoff`, which
reads as a broken hand-off rather than a broken meter — and this event is the
epic's own instrument for that funnel.

The full `posthog-js` SDK fixes this at the capture site with
`{ transport: 'sendBeacon' }` / `send_instantly`. posthog-js-lite@4.10.1 has
neither — its `PostHogCaptureOptions` is `{ uuid, timestamp, disableGeoip }`, and
there is no beacon or `keepalive` anywhere in `@posthog/core@1.46.1` — so
`flush()` is the only delivery lever it exposes. `trackBeforeNavigation()`
(`app/lib/analytics.ts`) captures, awaits `flush()` against a 250 ms budget, and
the CTA navigates in `.finally()`. A modified click (⌘/ctrl/shift/alt, or a
non-primary button) is left to the browser: that opens a new tab, this document
survives, and the event rides the normal batch. Pinned by
`climb-handoff-cta.test.tsx` (delivery path, event name, `environment` and the
attribution props) and by two `analytics.test.ts` cases.

### Single-owner file exception (for Fable)

`app/__tests__/crawler-classic-invariant.test.ts` is declared single-owner
(W-09). It receives **exactly one change**: the deletion of the 3-line
`vi.mock('@/app/[board_name]/…/list/layout-client', …)` at lines 151-153. Vitest
resolves a mocked specifier at registration time, so leaving it after this PR
deletes that module would fail the entire suite on module load. **Nothing else in
that file changes** — no helper, no assertion, no `describe`, no
`CANONICAL_SURFACES` entry. The alternative (keeping a gutted `layout-client.tsx`
alive purely so the mock resolves) leaves dead code and is worse.

### Line count and revert recipe

This PR is **4,970 changed lines across 61 files** (3,376 added, 1,594 deleted).
It does **not** stay in the epic's ≤1,500-line revert class, and the body says so
rather than repeating the claim. The revert is still mechanical rather than an
entangled unwind:

1. Restore four pages (`view/[climb_uuid]/page.tsx` ×2, `list/page.tsx` ×2) and
   two layouts (`list/layout.tsx` ×2).
2. Restore `list/layout-client.tsx` + `layout-client.module.css`.
3. Delete `app/components/climb-front-door/` and
   `app/lib/data/front-door-data.server.ts`.
4. Restore the 9 allowlist entries and the three e2e specs.

### Allowlist: 9 edges cut, 0 added

| from | to |
| --- | --- |
| `…/[angle]/list/layout-client.tsx` | `components/graphql-queue/index.ts` |
| `…/[angle]/list/layout-client.tsx` | `components/onboarding/tour-queue-watcher.tsx` |
| `…/[angle]/list/layout-client.tsx` | `components/queue-control/queue-list.tsx` |
| `…/[angle]/list/layout-client.tsx` | `components/search-drawer/accordion-search-form.tsx` |
| `…/[angle]/list/layout-client.tsx` | `components/search-drawer/search-results-footer.tsx` |
| `b/[board_slug]/[angle]/list/page.tsx` | `components/board-page/board-page-climbs-list.tsx` |
| `b/[board_slug]/[angle]/view/[climb_uuid]/page.tsx` | `components/board-page/board-page-climbs-list.tsx` |
| `components/similar-climbs/similar-climbs-list.tsx` | `components/climb-actions/index.ts` |
| `components/similar-climbs/similar-climbs-list.tsx` | `components/graphql-queue/index.ts` |

76 → 67 entries. Both invariant assertions are green, including the
stale-entry one.

The three legacy front-door files are promoted to `KEPT_ENTRY_FILES` (not
`KEPT_ROUTE_DIRS` — the rest of that tree is still classic UI). W-17 owns
promoting or deleting the remaining legacy siblings; recorded on #4358.

### e2e specs (standing rule 6 — migrated in this PR)

- **Deleted:** `queue-persistence.spec.ts`, `play-view-swipe-close.spec.ts`,
  `grid-mode-ascent-badge.spec.ts`. All three drive the queue bar, play drawer
  and grid-mode badge off `/kilter/…/40/list`; none exists on the front door and
  all have app-side twins.
- **Rewritten:** `layout-screenshots.spec.ts` — steps 1-2 keep the list
  screenshot, steps 3-5 now follow the first row's real anchor to the climb
  front door and shoot that instead of the drawer. Better board art, not worse:
  the front door paints the full-resolution overlay.
- **Skipped with a W-16 pointer:** `help-screenshots.spec.ts`'s
  **unauthenticated** describe — the six shots that drive the filters drawer,
  the heatmap and the play drawer off `/kilter/…/40/list`, none of which exists
  on the front door. W-16 owns the `/help` rewrite and its replacement
  screenshots; deleting the pipeline before its successor exists would strip it,
  and the committed `public/help/*.png` are untouched either way, so only
  regeneration pauses.
  The file's **authenticated** describe stays live and keeps regenerating three
  PNGs: its `beforeEach` only waits for a climb row (the front door emits one),
  `/settings` is its own route, the party-mode shot dispatches at the root
  layout's `OnboardingDummySeshMount`, and `personal progress filters` drives the
  header drawer that the untouched parent `[angle]/layout.tsx` still mounts over
  the front door. W-17 removes that header and takes the filters shot with it —
  noted in the file so the next reader doesn't assume the whole spec is parked.
- **Also skipped, beyond the plan's five:** the `Bottom Tab Bar - Queue
  Integration` describe in `bottom-tab-bar.spec.ts`. It double-clicks a climb
  card on the list to push it into the queue. The tab-bar visibility and
  navigation blocks in the same file still exercise the front door and stay live.
- `waitForBoardListReady`'s selector gained `[data-testid="climb-thumbnail"]` —
  the only climb-row marker the front doors emit. `global-setup.ts`'s own
  reachability check had the same stale selector hardcoded and would have timed
  out on the front door, hard-failing every e2e run; it now uses the same three.
- The deletions stranded fixture machinery that still ran: `global-setup.ts`'s
  `ensureGridBadgeFixture()` (which *throws* and aborts the whole Playwright run
  when the dev DB image carries no matching climb), `seed-social.ts` step 8.51,
  and three `SEED_CONTRACT.md` rows. All removed here.
- `SEED_CONTRACT.md`'s **default test board** contract is restated for the front
  door: the row marker is `[data-testid="climb-thumbnail"]`, and the "at least
  two onboarding-tagged rows" clause is gone — every spec that needed
  `#onboarding-climb-card` for queue population is deleted or skipped. The
  per-spec rows for `bottom-tab-bar`, both `help-screenshots` blocks and
  `layout-screenshots` now say what each actually needs.

### Accepted regressions on surfaces this PR does not otherwise touch

`SimilarClimbsList` now renders every card as a `LocaleLink`. Two consumers
outside the front door get that change with it, and both are in the delete set,
so the epic accepts it — but it ships before W-16/W-17 land, so **QA gate #4359
should exercise it rather than file it**:

- `build-climb-detail-sections.tsx:214` (the play drawer's Similar Climbs
  section) — a climber mid-session who taps a similar climb is now navigated to
  the climb front door instead of swapping the active climb in the queue.
- `create-climb-form.tsx:1872,1881` (the duplicate-resolution drawer) — same
  navigation, and the ellipsis that opened `ClimbActionsDrawer` is gone.

### Known gaps, recorded not fixed

- An installed-app iOS user tapping the CTA lands in the web SPA on
  `app.boardsesh.com`, not the native app: `packages/mobile/app.config.ts:370`
  claims `applinks:` only for www and the apex. Adding
  `applinks:app.boardsesh.com` is a native fingerprint change requiring a store
  release — flagged for a future native cycle, **never attempted by OTA**.
- The CTA strips the locale prefix, so an `/es` reader lands on the English app.
  The Expo app has no locale routing; this is the accepted regression recorded in
  the epic, and it is pinned by a test so nobody "fixes" it into a 404.
- **The front door opens a WebSocket on hydration, and W-17 will not close it.**
  `climb-front-door.tsx` → `social/climb-social-section` → `comment-section.tsx`
  runs `createGraphQLClient({ url: getBackendWsUrl() })` + `subscribe(…)` in an
  unconditional `useEffect` — no auth gate, no viewport gate — so an anonymous
  reader who never scrolls to the community section still opens one backend
  socket per pageview, on a page premised on being CDN-cached and
  personalization-free and heading for ~500k indexable URLs. Not a regression
  (the same component mounted via the drawer on `main`), and #4369 explicitly
  defers the no-WebSocket assertion to W-17. But W-17's row on #4358 only removes
  the *layout's* `WebSocketConnectionProvider`; `climb-social-section` is a keep,
  so this socket survives both W-17 and W-16 unless the subscription is deferred
  behind an intersection observer or a "load comments" interaction. Recorded on
  #4358 for W-17's scope.

## Release Notes

- Climb pages load fast and clean — see the problem, the grades, and one tap into the app to climb it.
- Every climb on your board is now one page away, and findable in search.

## Test plan

Gates, all green:

- `vp check` — the only findings are two pre-existing `docs/design/*.html`
  formatting reds that are identical on `main` (verified by stashing).
- `vp run typecheck` — full monorepo; the shared `events.ts` change touches web,
  mobile and backend graphs.
- `vp run test:web` — **501 files, 6102 tests, all passing**, including the two
  files whose contract this PR flips (`climb-canonical-parity.test.ts`,
  `import-graph-invariant.test.ts`).
- `vp run check:i18n` — 509 `.tsx` files scanned, no hardcoded user-facing
  strings.
- `vp run check:i18n:orphans` — clean. Swept four keys this PR orphaned:
  `climbs:similarClimbs.openActions`, `climbs:similarClimbs.actionsUnavailable`
  and `session:queueDrawer.clear{Title,Description,Confirm}`.
- Pre-commit hook ran clean (no `--no-verify`).

DoD checks, each with a vitest equivalent:

| DoD | Test |
| --- | --- |
| exactly one visible `<h1>` | `view-seo-fragment.test.tsx` — "carries exactly one `<h1>`" |
| board `` with explicit dimensions | same file — asserts `/api/internal/board-render` + `width="1080"` + `height="1350"` |
| setter `<a href="/setter/…">` | same file — "links the setter" |
| ≥1 angle cross-link, current angle not a self-link | same file — 25° is a link, 40° is not, `aria-current="page"` present |
| ≥3 internal links | same file — "emits at least three internal links" |
| CTA href = `APP_URL` + same pathname, no query | same file — exact-equality assertion |
| locale-prefix stripping (`/es`) | same file — explicit `/es` case |
| list SSRs ≥50 crawlable links | `static-climb-list-ssr.test.tsx` — 50 climbs at `virtualize={false}`, asserts ≥50 view anchors **and** that row 49 is present |
| W-12 consumers unchanged | same file — default path with the same 50 climbs is still windowed (≥8, <50) |
| both trees emit the identical canonical | `climb-canonical-parity.test.ts` — equality between the two on `view` **and** `/list`, plus a non-vacuity check that it isn't `/b` |
| the canonical follows the resolved ids, not the request spelling | legacy list page-metadata suite — params are an alias spelling and the assertions are exact-equality against the builder output |
| a hidden `/b` board leaks no canonical | `climb-canonical-parity.test.ts` + both `/b` page-metadata suites — `robots.index === false` **and** `alternates === undefined` |
| filter variants consolidate onto the clean base, identically on both trees | `climb-canonical-parity.test.ts` — "resolves filtered variants onto the same clean base on both trees" |
| similar climbs are crawlable anchors in the first render | `similar-climbs-ssr.test.tsx` — one anchor per seeded climb, no `<button>`, no spinner, and the throwing GraphQL client proves the seed is used |
| `?page=1..10` self-canonical, `?page≥11` noindex/follow | `list-front-door.test.tsx` + both list page-metadata suites |
| `?page=1` canonicalises to the bare path | legacy list page-metadata suite |
| the `next` anchor stops at the last indexable page | `list-front-door.test.tsx` — page 9 links onward, page 10 with `hasMore` emits no `rel="next"`, and page 11 still emits `rel="prev"` back into the indexable set |
| a deep `?page` can't become an unbounded offset | `list-front-door.test.tsx` — `?page=5000` clamps to the ceiling; both list suites — the page body 404s it **without** calling `fetchFrontDoorListPage`, and still serves the last grace-band page |
| `Climb Handoff Clicked` survives the cross-origin navigation | `climb-handoff-cta.test.tsx` — the click is `preventDefault`ed, navigation waits on the flush, a rejected flush still navigates, and a modified click is left to the browser; `analytics.test.ts` — `trackBeforeNavigation` awaits `flush()` |
| shadowed size keeps its qualified slug | `url-utils.test.ts` — `buildCanonicalClimbListUrl` on Kilter layout 1 size 27; `climb-canonical-parity.test.ts` also drives both trees at size 27 with a size description shared with size 10, so only the id-aware path can tell them apart |
| no `PlayViewDrawer` in the view page's tree | import-graph invariant, with the three legacy front-door files now walked as keep roots |

Not run: the dev server and the manual curl matrix — those belong to QA gate
#4359, which this PR is held on. `.boardsesh/qa-notes.md` in the worktree carries
the full manual matrix (four surfaces × four locales, the pagination edges, a
private `/b` board, a MoonBoard single-angle climb, a null-difficulty draft, and
a backend-down degradation case).


marcodejongh:
Thanks — one fixed, one respectfully declined.

**#2 (`initialData` without `initialDataUpdatedAt`) is a real bug and is fixed** in d3a9098. You are right about the mechanism: React Query stamps `dataUpdatedAt = 0` when `initialData` arrives without a timestamp, so `Date.now() - 0` beats the 5-minute `staleTime` and the client refetched on hydration — re-running the resolver that is rate-limited 30/min against the web server's single IP, which is precisely what the server-side cache in `front-door-data.server.ts` exists to avoid. The comment in the code now says why the second option is load-bearing rather than decorative.

**#1 (French `vous` → `tu`) is declined as stated.** The claim that "the French glossary and CLAUDE.md both mandate informal **tu**" does not hold:

- `docs/i18n-french-glossary.md` contains no register guidance at all — no `tu`, no `vous`, no `tutoiement`. It covers `enchaîné`/`la croix`, `essais` not `tentatives`, `allumer` not `envoyer`, and `la board` as feminine. Register is simply not in it.
- CLAUDE.md mandates the informal second person for **German** ("Product UI uses informal **du**"). Its French paragraph says nothing about register.
- `fr/climbs.json` is genuinely mixed today (19 `Connectez-vous …`, plus 11 `tu`-form strings), so there is no established convention in this namespace to violate.

The register for these strings was ruled `vous` before implementation, so they stay. Your finding did surface a real inconsistency inside this PR, though: `list.frontDoor.intro` had slipped into a `tu` imperative (`Ouvre-en un`) next to `vous y êtes`. That is fixed in the same commit — the new strings are now uniformly `vous`.

Catalog-wide register consistency for `fr` is a separate piece of work and does not belong in a gate-held front-door PR.

marcodejongh:
**Fable review: approved, gate-held.** Final pre-merge read per the epic #4358 rule for W-15, covering the base implementation and both fix rounds.

Verified directly: the single-owner `crawler-classic-invariant.test.ts` exception is exactly the sanctioned 3-line mock deletion and that suite runs green; the A1 canonical flip goes through the shared id-aware builders on both trees (raw-echo canonical fixed in round 1); the pagination crawl corridor is bounded at both ends (link cap 10, render ceiling 20 with a prev-only grace band, 404 beyond — adversarially walked as a crawler and fuzzed); `Climb Handoff Clicked` survives cross-origin navigation via an awaited flush raced against a 250ms budget, built against the API the pinned `posthog-js-lite` actually has; and I mutation-proved the flush-await test oracle myself (macrotask resolution — a fire-and-forget implementation now fails it).

The CDN-safety exclusion of server-side logbook threading is correct and stays: these URLs carry `s-maxage` with no session split, so one viewer's badges would cache for everyone.

**This PR merges only after QA gate #4359 is signed** (physical devices, both origins). The Boardsesh-grade SSR decision is recorded as held-out on #4359 and remains Marco's call at signing. Accepted regressions and the W-17-deferred WebSocket note are documented in the body.

PR #4432, opened 2026-08-14
feat(mobile): notifications in the app

### LINKED ISSUE #4429: W-20a: mobile notifications route
Part of #4358. **Blocked by:** none — W-06 (#4361) merged 2026-08-14. **Gate-independent: can land while QA gate #4359 pends.**

## Context

`packages/mobile/app/(tabs)/profile/` has **no** notifications route (verified 2026-08-12 — re-verify at branch time), while the data layer is already shared via `@boardsesh/graphql`'s `GetGroupedNotifications`. Web's `/notifications` page is scheduled for deletion in W-20b, which is only allowed to land **≥1 production deploy after** this route ships to the fleet (standing rule 1).

## The work

- Add `(tabs)/profile/notifications.tsx` plus the `HomeTopChrome` entry.
- Extract any **non-trivial** grouping logic to a pure package with web repointed to it in the same PR (extract-don't-duplicate). If the grouping is trivial, skip the extraction and say so in the PR body.

## Standing-rule callouts

- **Ships to the native fleet as a production OTA on merge** — standing rule 7 applies: `eoas publish --rollout-percentage` staging, name the last-good commit in the PR body, hold full rollout until the PostHog OTA-health signature is clean (`Downloaded` events with **nonzero** Status launches for the new `updateId`).
- **Mobile perf checklist applies** (PR review): FlashList (never `.map()` in a ScrollView), memo'd row, `renderItem` `useCallback` deps with **no array `.length`** and no inline closures, hoisted `keyExtractor`, provider `value` memoized, per-row hooks O(1).
- **Fable review required** (packages/mobile).
- **Release Notes**: user-visible — notifications arrive in the app. Climber voice.

## Gates

**GM**: `vp run typecheck:mobile` · `vp run test:mobile` · `vp run check:mobile-bundle` · `vp run check:mobile-web-bundle`. Never `vp test --project mobile run` (false green).

## Verifies

- [ ] The notifications route renders grouped notifications on iOS + Android + expo-web.
- [ ] `HomeTopChrome` links to it.
- [ ] Perf checklist items confirmed in the PR body.
- [ ] W-20b is NOT touched (web `/notifications` survives until one deploy later).

marcodejongh:
## Summary

Notifications get a real screen in the mobile app, plus a bell in the Home chrome that carries an unread dot.

Part of #4358
Closes #4429

**What's here**

- **Route**, registered in both tab stacks as 1-line re-export stubs of one `NotificationsScreen` — the same shape `session/[sessionId]` already uses. The issue asked for a single route under `profile/`, but the bell lives on Home: a lone profile route makes the bell switch tabs and Back land on the You screen. Home's bell pushes `/(tabs)/home/notifications`; the profile route the issue named ships too, reached from a **Notifications row in the You → More list** (without it that route would be registered but unreachable).
- **Bell** (`NotificationsToolbarAction`, glass + Material) in `HomeTopChrome`'s trailing island, now two slots wide. It owns its own unread count exactly as `UserAvatarToolbarAction` owns its profile, so `HomeTopChromeProps` is unchanged. The badge is a **dot, not a number**: `GlassActionToolbar` is `overflow: 'hidden'` and crops corner badges (already documented at `BoardToolbarAction.tsx:46`), so a `99+` pill would clip. The count goes in the VoiceOver label instead.
- **Data layer**: four hooks over the existing shared `GetGroupedNotifications` operations. Mark-read cache writes mirror web's exactly.
- **Tap behaviour** mirrors web: an unread group marks read then navigates; follows open the profile, an approved gym claim opens the gym editor, climb-bearing types open the play drawer. Comment/vote rows mark read and stay put — the resolver never enriches those with a `climbUuid`, so web's climb branch is dead for them too.

### No extraction — grouping is server-side

The issue asked to extract "non-trivial grouping logic" to a pure package. There is none to extract: grouping happens in SQL at `packages/backend/src/graphql/resolvers/social/notifications.ts` (`GROUP BY type, entity_type, entity_id` + `COUNT(*) OVER()`), shaped and enriched in the same resolver. Web's entire client-side "grouping" is one line — `pages.flatMap((page) => page.groups)` (`use-grouped-notifications.ts:64-67`).

The only real duplication candidate is the ~105-line i18n-key/icon mapping in `notification-item.tsx:26-131`, which is *presentation*, not grouping — and W-20b deletes its sole consumer one deploy later. Extracting would repoint one web file that dies immediately, leaving a single-consumer package. So: **no extraction.** Mobile's copy lives in a standalone pure leaf (`notification-copy.ts`) so a future extraction is a file move, not a rewrite.

### Layout + angle now come from the server, like web

A grouped notification carried a climb uuid and a board type but no **layout**, and the layout is not optional: `climb(uuid, layoutId)` filters on it (`get-climb.ts:72-74`), so a client that guesses resolves `null` and dead-ends the user on "climb not found" with no way back to the list. Guessing is what `getDefaultRenderBoard` does when it has no layout — `getAllLayouts(boardName)[0]`, which is **layout 1 for Kilter and 9 for Tension**. That misses every Kilter Homewall (8) and every Tension Board 2 climb (10, 11), across `new_climb`, `new_climb_global`, `new_climbs_synced` and all four `proposal_*` types.

Web never had this gap because `/api/internal/climb-redirect` reads `board_climbs.layout_id` and `board_climbs.angle` straight from the DB. So this PR gives GraphQL clients the same resolution rather than a better guess:

- **Backend**: `GroupedNotification` gains `climbLayoutId` + `climbAngle`, read off the `board_climbs` rows the enrichment loop **already fetches** — two extra columns on an existing batched select, no new query and no N+1.
- **Client**: layout is `climbLayoutId` → the reader's board when it's the same board → let the `ref` branch resolve it. Size and sets ride along only when the climb really sits on the reader's layout, so it draws on their actual wall instead of the layout's biggest size. Angle is `climbAngle` (where the climb's grade and stats live) → the reader's board angle → `defaultAngle`.

`board_climbs.angle` is nullable and most climbs carry none, so `climbAngle` stays null rather than collapsing to `0` the way web's `climb.angle ?? 0` does — the reader's own board angle is a better guess than flat.

**Deploy order**: this adds two fields to the query document, so the backend must be live before the client asks for them. That is the natural order here — the backend deploys on merge, while the OTA is a *staged* rollout held on the PostHog health signature. Both fallback rungs on the client exist for the window anyway: an OTA'd client talking to an older backend sees `climbLayoutId` absent and falls back to the reader's board.

### Divergences from web, enumerated

- **`proposalUuid` goes nowhere.** Web appends it to the climb URL so the proposal is surfaced. Mobile has no proposal UI at all — no component, route, or drawer reads one (`grep proposalUuid packages/mobile/src packages/mobile/app` returns only test fixtures and the operation document). A `proposal_*` row opens the plain climb. Recorded in the nav hook's divergence list so the next reader doesn't mistake it for parity; thread it through the day a proposal surface lands.
- **`new_climbs_synced`** has no mobile setter-profile route to open (the Climbs tab's `setters` screen is a filter picker, not a profile), so it falls through to the climb branch — safe, because the resolver populates `climbUuid` + `boardType` for exactly this type.
- **Timestamps.** Web renders compact `5m` / `2h` / `3d` then a locale date, via the `notifications:time.*` keys. Mobile renders `formatTickRelativeTime` (dayjs `fromNow()`, "5 minutes ago"), matching every other mobile timestamp; `setRelativeTimeLocale` is wired at i18n init so it localises. **W-20b note:** deleting `notification-item.tsx` removes the only consumer of `notifications:time.*` — those four keys must go with it or `check:i18n:orphans` fails.

### Production targets

This reaches **BOTH** production targets: `app.boardsesh.com` on merge, and the **native store fleet via auto-OTA**. The diff is JS/TS + one bundled asset (`assets/material-icons/notifications.xml`, required from JS) with no new native dependency, so the fingerprint is unchanged and the OTA lands on already-shipped binaries.

Per standing rule 7, roll out staged: `eoas publish --rollout-percentage N`, hold full rollout until PostHog shows `Downloaded` events with nonzero `Status` launches for the new `updateId`. **Last-good commit: `31ccf56a64cca3b87ca378b954e4a51c397dc6b8`** (this branch's base).

### Perf checklist

| Item | Confirmation |
| --- | --- |
| List virtualized | `FlashList`; `onEndReached` fires **one** `fetchNextPage` per end-reach, guarded on `hasNextPage && !isFetchingNextPage` |
| Row memoized | `NotificationRow` is `React.memo`'d; props are the group object plus one stable `onPress`, nothing else |
| `renderItem` deps clean | Single dep (`handlePress`) — no array `.length`, no `unreadCount`, no inline closures |
| `keyExtractor` hoisted | Module scope, stable for the list's lifetime |
| Provider values memoized | No new provider |
| Per-row reads O(1) | `notificationCopy` / `notificationIconName` are switches — no scans, no `filter` |
| Worklet `runOnJS` gated | No gestures added |
| New effects bounded | One `useEffect` for `headerRight`, keyed on `unreadCount > 0 && groups.length > 0` |
| No layout animation in a recycled cell | The unread dot is a plain themed `View`, **not** `Badge` (see below) |

The unread dot deliberately does not use `Badge`. Badge's Liquid Glass variant is an `Animated.View` carrying `entering={FadeIn.springify()}` / `exiting={FadeOut.duration(150)}` — correct for the toolbar bell, wrong inside a virtualized list: FlashList reuses a cell's subtree for a different item, so a recycled cell whose notification flips read↔unread mounts/unmounts the badge mid-scroll. That springs a dot in per recycled unread row and, on the 150ms exit, leaves a dot sitting over a row already showing a **read** notification. This was `Badge`'s first use inside a list; the two toolbar actions keep it, where the entrance is the point.

The screen destructures its React Query result rather than holding the result object: React Query mints a fresh object every render, so callbacks listing the whole result as a dep get a new identity each pass and their `useCallback` buys nothing. Both mutation hooks are consumed as `const { mutate } = …` for the same reason — otherwise each tap churns the nav callback → `renderItem` → every row. A test asserts `renderItem`/`keyExtractor` identity survives a page append **and** an unread-count change.

`refreshing` is `isRefetching && !isFetchingNextPage`: `isRefetching` is `isFetching && !isPending`, which is also true while a next page loads, so without the guard the pull-to-refresh spinner pops at the top every time the user paginates at the bottom.

### Notes for review

Two barrels deliberately do **not** re-export the new modules (`src/lib/graphql/hooks/index.ts`, `src/components/chrome/index.ts`). The notification hooks gate on the stored auth token, so they reach expo-secure-store; both barrels are imported by suites that mock only the GraphQL client, where the extra native reach makes Rolldown's scan hit react-native's Flow source and fail the whole file. Consumers import by path. Comments at both sites record why.

`notification-copy.ts` returns `{ textI18nKey, params }`. The `…I18nKey` suffix is load-bearing: `check:i18n:orphans` treats such a property as a key holder, which both records the literals and accepts the row's `t(copy.textI18nKey, …)` as statically resolved. Renaming the field hard-fails the checker.

"Mark all as read" is gated on `unreadCount > 0 && groups.length > 0`, mirroring web's `groupedNotifications.length > 0 && unreadCount > 0` — the count is its own query and settles first, so without the list gate the action floats over an empty screen on first paint.

One new i18n key (`notifications.errors.load`) in all four locales; the You → More row reuses the existing `notifications:title`. Everything else was already at full parity.

## Release Notes

- Notifications now live in the app — see who liked your sends and follows your climbing without opening the website.

## Test plan

Gates, all green:

- `vp run typecheck` (full — schema, backend and shared locale catalogs changed)
- `vp run test:mobile` — 668 files, 6437 tests
- `vp run check:mobile-bundle` (asserts the new Android icon asset bundles)
- `vp test run --project backend`
- `vp run test:web` — 498 files, 6050 tests
- `vp check` — clean except two pre-existing `docs/design/web-reposition/*.html` formatting hits that also fail on `main` (last touched by 19268ee78; untouched by this PR)
- `vp run check:i18n`, `vp run check:i18n:orphans`

New coverage:

- `notification-copy.test.ts` — all 14 `NotificationType` branches table-checked against web's `notification-item.tsx`, every `…WithBody` fallback, the `newClimbsSyncedSetter` split, the four-case actor ladder, plus every icon name asserted to be a real `iconMap` key (a typo otherwise ships a blank glyph).
- `notification-row-memo.test.tsx` — the memo bails on an identical re-render and on a fresh element with the same group + handler; re-renders when `isRead` flips or the handler identity churns.
- `notifications-screen.test.tsx` — empty / pending / **offline-blocked** / error+retry states, `headerRight` presence keyed on unread **and** on a non-empty list, mark-then-navigate vs navigate-only, comment rows not navigating, one-page-per-end-reach, and the `renderItem`/`keyExtractor` stability proof. Four navigation cases pin the layout/angle ladder: a Tension Board 2 climb opens on **layout 10, not the guessed 9**; a climb on the reader's own layout carries their size + sets; a setter's fixed angle beats the reader's board angle; and a response with no `climbLayoutId` falls back to the reader's board.
- `use-notifications.test.tsx` — the offset counts groups already held (not pages × 20), the unread badge drops by the server-reported count (a group holds many rows) and never below zero, and the grouped query publishes the unread count so the bell settles without a second round trip.
- `HomeTopChrome.test.tsx` — bell renders in both variants and pushes Home's own stack route.
- `read-only-route-corpus.ts` — both notification paths added to `GATED_PATHS`, which buys an assertion in all five anonymous-gate suites that a signed-out visitor hits the login wall rather than an empty list.

The offline branch is the one worth a second look: `query-provider` runs React Query in `offlineFirst`, so a network-only query **pauses** and `isPending` never clears. Without `useOfflineQueryState` the screen is a permanent spinner.

Manual QA still wanted on a device for the glass island's two-slot geometry (the centred scope pill shifts ~24px left) and the Material bell badge offset.


marcodejongh:
**Fable native-OTA review: approved.** Reviewed the final diff per the epic #4358 mobile rule. The perf checklist holds after the fix round (plain themed dot instead of an Animated Badge inside recycled cells; destructured React Query results so the memoized rows stop churning; O(1) per-row reads). The wrong-board-layout major was fixed durably at the backend (`climbLayoutId`/`climbAngle` on the schema, nullable, resolver covered by test-backend) rather than papered over client-side — right call.

**One merge-time caveat (deploy ordering):** the shared query document now requests the two new fields unconditionally, and GraphQL validation rejects the whole query against a backend that lacks them. Same-merge backend deploy makes the window ~zero in practice (expo applies OTAs on next launch), but if the Railway backend deploy **fails on this merge while the OTA publishes**, the new notifications screen errors fleet-wide until the backend ships. If that happens: re-publish the last-good OTA per the runbook (base SHA is in the PR body) or fast-fix the backend. Standing rule 7 staging applies as written.

Also noting the non-gating `pr-4432` OTA-preview publish job failed after 51m — the preview channel for device QA may need a re-dispatch; the production path is unaffected.

Issue #4368, opened 2026-08-12
W-14: cut the classic-web rollback artifact (tag + release branch)


marcodejongh:
Part of #4358. **Blocked by:** #4367 (and it should be cut after #4366 lands too, so the artifact contains the fully-unwired tree).

## Context

**This is the programme's rollback.** The web reposition is a hard delete with no `?classic=1` runtime fallback, and W-16 removes ~110k LOC and swaps the root chrome in one commit. Reverting that with `git revert` under incident pressure means reintroducing that much code, re-mounting a provider tree, rebasing over subsequent merges, and passing `check:i18n:orphans` — hours of work at the worst possible time.

So the rollback is a **deployable artifact, not a revert**: a tag and a release branch that still serve the classic interactive web app, confirmed deployable before anything irreversible ships. This is also the compensating control for the locked decision not to re-arm the expo-web flag rollout.

No code changes — this is a tag, a branch, and a documented procedure.

## The work

- Cut the tag **`classic-web-last-good`** on `main`, at the last commit **before #4369** merges.
- Push the branch **`release/classic-web`** at that same commit.
- **Confirm it deploys** to a preview alias, and confirm by hand that the alias serves a working classic climb page **with the queue drawer** — a build that compiles is not proof; the interactive surface has to actually run.
- Document the restore procedure in `docs/web-reposition.md`: which alias, which environment variables, how to promote it, and who can do it.
- **90-day retention.** Record the expiry date in the doc so it is not deleted early and not silently kept forever.

Files: `docs/web-reposition.md`.

## Verifies

- [ ] Tag `classic-web-last-good` exists on `main` at the pre-W-15 commit.
- [ ] Branch `release/classic-web` is pushed at the same commit.
- [ ] The preview alias serves a working classic climb page **with the queue drawer** (screenshot in the PR body).
- [ ] `docs/web-reposition.md` documents the restore procedure end to end, including the 90-day retention window and its expiry date.

## Gates

- Manual deploy of the preview alias — that deploy **is** the gate.
- **GW** on the docs commit: `vp check` · `vp run typecheck` · `vp test run --reporter=agent --project web`.

## Standing-rule callouts

- **This is the rollback for everything after #4369.** #4369 and W-17 both depend on it; W-16 is not openable without it.
- The locked decision is explicit: **rollback is the `release/classic-web` deployable artifact, not `git revert`.** Do not propose reverts as the recovery plan in any downstream PR body.
[...]

Issue #4435, opened 2026-08-14
W-16 (+W-18): delete the climbing UI and swap the root chrome


marcodejongh:
Part of #4358. **Blocked by:** #4433 (W-17 — delete the board-route siblings) and #4428 (W-14 docs; the rollback artifact is already cut). W-02 (#4360) and W-04 (#4171) are merged. QA gate #4359 is **SIGNED and closed (2026-08-14)**.

🔒 **−~110k LOC, the merged delete, and the irreversible one.** W-16 and W-18 merge into one PR because they typecheck in no order either.

## Why this one is different

Reverting means reintroducing ~110k LOC, re-mounting a provider tree, rebasing over subsequent merges and passing `check:i18n:orphans` under incident pressure. Rollback is the `release/classic-web` deployable artifact (#4368 / PR #4428), **not `git revert`**. Recount the LOC at branch time and put the real number in the PR body.

### Delete the session stack

`queue-control`, `graphql-queue`, `play-view`, `persistent-session`, `connection-manager`, `party-manager`, `board-lock`, `sesh-settings`, `session-*`, `playback`, `healthkit`, `board-bluetooth-control`, `lib/live-activity/**`, `lib/ble/*` **except `capacitor-utils.ts`**, six hooks.

### Delete the browsing UI

`board-page`, `climb-actions`, `logbook`, `library`, `board-scroll`, `playlist-generator`, `create-climb`, `hold-classification`, `new-climb-feed`, `climb-quality-filter`, `search-drawer`, the four drawers, `bottom-tab-bar`, `onboarding`, `feedback`, `board-provider`, `grade-picker`.

### `climb-card/` goes file by file, never as a directory

Four `activity-feed` files and `climb-detail-header.tsx:8,9` import `climb-icons`. **Survivors** are `climb-thumbnail`, `climb-title`, `climb-icons`, `climb-card-cover`, `marquee-text`, `ascent-status`.

### The play drawer and the logbook edge

`play-view/play-drawer-content.tsx` goes here, and with it `build-climb-detail-sections.tsx`'s `logbook-section` + `crew-logbook-view` imports and that W-05 allowlist entry. (W-15 parametrized `build-climb-detail-sections.tsx` with `includeLogbook` rather than cutting it; this PR removes the remaining consumer and the imports.)

### Swap the root chrome

[...]

PR #4448, opened 2026-08-14
chore(web): delete the board-route siblings on both trees

### LINKED ISSUE #4433: W-17: delete the board-route siblings on both trees
Part of #4358. **Blocked by:** #4430 merging (W-15 — the SSR climb + list front doors, ready for review). W-14's rollback artifact is already cut (PR #4428, open). QA gate #4359 is **SIGNED and closed (2026-08-14)**, so the gate no longer blocks this.

🔒 **Point of no return. This one lands *before* the deletes.**

## Context

`create`, `import`, `liked`, `logbook`, `playlists` and `play` are still mounted under `[angle]/layout.tsx` on both trees. W-15 deliberately left the session providers in place ("Do not strip the session providers from `[angle]/layout.tsx`") because unmounting them while those siblings still exist crashes at runtime with **no typecheck error**. This PR removes the siblings, *then* the providers, and turns `/play` into a redirect-only route.

## The work

- Deletes `create`, `import`, `liked`, `logbook`, `playlists`, `play` **on both trees**.
- **Then** strips the seven session providers from both `[angle]/layout.tsx` and removes `BoardProvider` from `[board_name]/layout.tsx`. **Order within the PR matters, because providers-before-siblings crashes at runtime with no typecheck error.**
- Leaves the `/play` pages as **redirect-only routes** — they already 301 with a climb-name lookup, and a static `next.config` redirect would emit a bare-uuid `/view` that redirects again, turning one hop into two.
- Every redirect gets `es`/`fr`/`de` siblings, and cross-origin destinations are `permanent: false`.

## Standing-rule callouts

- **Locale-sibling redirects (standing rule 4).** `next.config.mjs`'s `redirects()` entries are all bare paths and this repo has no Next i18n routing (locales are a middleware prefix-strip), so `source: '/play'` never matches `/es/play`. Every redirect ships with its `es`, `fr` and `de` siblings in the same commit.
- **`permanent: false` for cross-origin destinations.** We are **not 308-ing board URLs into `app.boardsesh.com`**: cross-origin permanent redirects are unrecoverable — browser-cached indefinitely, and `middleware.ts` serves them CDN-stale for `stale-while-revalidate = TTL × 7` (7 days for list pages, 7 hours for climb views) with no hatch left. Cross-origin destinations stay `permanent: false` until the SPA route is proven in production.
- **Remove a `Disallow` in the same commit as its redirect (standing rule 3)**, never one PR later — if the disallow outlives the redirect, Googlebot never fetches the URL and never learns it moved.
- **e2e spec deletions ship with this PR (standing rule 6).** `e2e-tests.yml` is `workflow_dispatch`-only, so nothing catches the lag. Specs live in `packages/web/e2e/`, not `e2e/`.
- **Never delete `climb-card/` as a directory** (standing rule 5) — file by file only, if any sweep here reaches it.
- **No `?classic=1` survives.** The cutover is a hard delete per `docs/web-reposition.md`; rollback is the `release/classic-web` deployable artifact (#4368 / PR #4428), not `git revert`.

## Deps

- W-15 — #4369, PR #4430 (must merge first)
- W-14 — #4368, PR #4428 (rollback artifact; already cut)
- QA sign-off — #4359, **signed 2026-08-14**

## Gates

- **GI** — `vp run check:i18n` · `vp run check:i18n:orphans`
- `vp run test:e2e`
- **Fable review** of the final diff before merge.

## Verifies

- [ ] Every deleted path 30x's in **all four locales**.
- [ ] A test asserts the locale twins and the `permanent: false` rule.
- [ ] **No WebSocket opens on any board route** — the assertion W-15 cannot make while the layout still mounts `WebSocketConnectionProvider`.

## Tiering

**Opus** implementation (judgement, not transcription — provider-tree ordering and runtime-only failure modes), all three adversarial QA lenses on **Opus** (delete-safety · SEO · native-OTA), and **Fable reviews the final diff** — W-17 is on the Fable point-of-no-return list.

## Release Notes

User-visible: the interactive climbing surfaces move to the app; www keeps every URL working via redirects. Climber voice.

marcodejongh:
## Summary

Part of #4358. Closes #4433.

**Point of no return.** `create`, `import`, `liked`, `logbook` and `playlists` are gone from both board trees, and the session providers are gone from both board layouts. What is left on www under a board URL is a server-rendered document: the climb list front door, the climb view front door, and the two `/play` redirect pages (untouched — they already 308 to the slugged `/view` URL with a climb-name lookup, which a static config rule cannot do without turning one hop into two).

The commits are ordered deliberately and should be read that way:

1. **Redirects first**, so no board path 404s at any commit on this branch.
2. **Siblings deleted** — 15 files, 1,456 lines, both trees.
3. **Then the providers.** The reverse order leaves `liked`/`create`/`logbook` mounted over stripped contexts: a runtime 500 on every board route with a green typecheck.

`BoardSeshHeader` came out with the providers. It is not on the epic row's list, but it calls `useCurrentClimb`, `useSearchData` and `useUISearchParams` — leaving it mounted is exactly that silent failure.

### Redirects

Every deleted path 30x's, and every rule ships its `/es`, `/fr` and `/de` twin in the same commit. `redirects()` runs before `middleware.ts` and matches sources literally, and locales here are a middleware prefix rather than Next i18n routing, so a bare source never matches a localized path. `expandLocaleRedirects` builds the twins from one base list — including for the five rules that predate this PR and never had them.

| Path | Lands on | Status |
| --- | --- | --- |
| `/{board}/{layout}/{size}/{sets}/{angle}/create` (numeric) | `${APP_URL}/climbs/create?boardName=…&layoutId=…&sizeId=…&setIds=…&angle=…` | 307 |
| `…/create` (slug forms, `/b/{slug}` included) | `${APP_URL}/climbs/create` | 307 |
| `/moonboard/…/import` | `/moonboard-import?layout=…&sets=…&angle=…` | 308 |
| `…/import` (any other board) | that board's own `…/list` | 308 |
| `/b/{slug}/{angle}/import` | `/moonboard-import?angle=…` | 308 |
| `…/liked` | that board's own `…/list` | 308 |
| `…/logbook` | `/playlists` | 308 |
| `…/playlists`, `…/playlists/:uuid` | `/playlists`, `/playlists/:uuid` | 308 |

Order is load-bearing on `…/import`: bulk import is MoonBoard-only, and
`MOONBOARD_LAYOUTS` ids run 1–7 and collide with Aurora layout ids, so without
the MoonBoard-specific rule ahead of the catch-all, `/kilter/1/…/import` would
render the importer for MoonBoard 2010. The catch-all reproduces what the
deleted page did with a non-MoonBoard board: send it to that board's list.

No `Disallow` in `robots.ts` covers any of these paths, so there is none to remove alongside them.

### Creating a climb goes straight to the app

Every www surface that used to link at `…/create` now links at the app editor
directly, through a new `buildAppCreateClimbUrl` next to `buildAppHandoffUrl`:
the bottom bar's **Create** tab (mounted on every www page) and the
**remix/edit** action on a climb. No redirect hop, and the board rides along as
the exact params `packages/mobile/app/(tabs)/climbs/create.tsx` reads —
`boardName`, `layoutId`, `sizeId`, `setIds`, `angle`, plus `forkFrames` /
`forkName` / `forkDescription` / `editClimbUuid` for a remix or a draft edit.
Two consequences worth naming:

- The Create tab is now a real cross-origin `<a>`, always. It no longer reads
  `getLastUsedBoard()` from IndexedDB and no longer opens the board-selector
  drawer — with no board in scope it hands over bare and the app opens on the
  signed-in user's active board. `getActiveTab`'s `/create` branch went with it:
  no www pathname belongs to that tab any more.
- Remix no longer needs `layout_name` / `size_name` / `set_names`. The app takes
  the numeric tuple, so a board the static slug tables don't carry can be
  remixed now, and a shadowed size (Kilter layout 1, sizes 10/27) can't be
  ambiguous the way its bare slug was. `constructCreateClimbUrl` and
  `tryConstructSlugCreateUrl` are deleted — nothing built a www create URL any
  more.

**The bookmarked-URL regression that remains.** A *slug-form* create URL
(`/kilter/original/12x12-square/screw_bolt/40/create`, and the whole
`/b/{slug}` tree) hands over bare: the app parses the ids with `Number()`, so
forwarding `original` would seed the editor with `NaN` — worse than its
active-board fallback — and resolving a slug needs a DB lookup a static redirect
cannot do. The canonical numeric form keeps its board.

The create rules are the only cross-origin destinations here and they are
`permanent: false`. A permanent cross-origin redirect is cached by the browser
indefinitely with no server-side hatch, so they stay temporary until the app
route is proven in prod. A test pins that so a later edit cannot flip it.

One accepted-for-one-wave regression remains, on a surface W-16 deletes
wholesale: board list pages lose the in-place climb-name filter and the filter
drawer, which have no back end left.

### The MoonBoard importer moved

The epic keeps the bulk importer on Next, but its only two entry points were the `import` routes this PR deletes. It now lives at **`/moonboard-import`** — a noindex utility route with no board-tree providers (it needs none: next-auth, react-query, the snackbar and i18n all come from the root layout). It takes the board from the query string, and the old paths 308 into it carrying whatever they knew — layout, **hold sets** and angle. The sets matter: the deleted page drew only the sets the URL named, so without them a MoonBoard 2024 wall would get all six of the layout's sets stacked on the import preview and the edit modal instead of the two it runs. An unresolvable or absent layout lands on a picker instead of an error, and unparseable sets fall back to every set for the layout (the `/b/{slug}` tree has no set segment, so it always does). W-19's reason for keeping `climb-search-cache/revalidate` still holds — the importer still calls it.

### Why the comment socket is now auth-gated

W-17's DoD is "no WebSocket opens on any board route", and neither `WebSocketConnectionProvider` nor `GraphQLQueueProvider` was ever the culprit — the one eager anonymous socket was `social/comment-section.tsx`, which dialled the backend from an unconditional effect for every visitor and every crawler on a climb page. It is now gated on `isAuthenticated`.

Anonymous readers keep the whole thread: `CommentList` fetches it over HTTP on mount and on every refresh, so comments still render server-round-tripped for readers and crawlers. What a signed-out reader loses is live push, which they could not participate in anyway — the composer was already gated on the same flag. Gating it in `comment-section.tsx` rather than in the climb-specific caller also clears the socket from `/gym/[slug]`, `/playlists/[uuid]`, `/session/[sessionId]` and `board-detail.tsx`, which the epic's end-to-end DoD needs regardless. A signed-in reader still opens one socket per climb pageview; if that wants tightening, an IntersectionObserver on top of the auth gate stays inside the same file.

### Import-graph allowlist: 67 → 40

The plan predicted 55; the invariant test is the oracle and the real number is 40 (its "no stale entry" assertion fails on anything left over, so this is not a judgement call). Out: every edge from a deleted route file, the eight provider edges from the named-board layout, the queue-shell edge from `persistent-session-wrapper` and the search-bridge edge from `global-header`. In: exactly one — the config-tuple board shell is promoted to `KEPT_ENTRY_FILES` and keeps a single allowlisted edge to `LastUsedBoardTracker` under `teardown:components-board-page`, deliberately, so the Climb tab still remembers your board until W-16 takes `components/board-page`. `app/moonboard-import` joins the keep roots, which is what still justifies the four `create-climb` edges.

### Orphans this teardown created, deleted here

Four modules lost their last importer in this PR and came out with it, rather
than sitting in the tree as dead code:

- `components/board-page/header.tsx` (+ `header.module.css`, its test) —
  `BoardSeshHeader`, unmounted from both shells above. `angle-selector.tsx`
  stays; the play-view drawer and the climbs list still use it.
- `components/search-drawer/search-drawer-bridge-context.tsx` — the header
  carried the only `SearchDrawerBridgeInjector`, and this PR removed the only
  reader (`useSearchDrawerBridge` in `global-header.tsx`). Its
  `SearchDrawerBridgeProvider` came off `persistent-session-wrapper`, so that is
  one fewer no-op provider on every page, and one fewer allowlist entry.
- `components/queue-control/queue-control-bar-shell.tsx` — its only importer was
  the `shouldShowQueueShell` branch this PR deletes.

`crawler-classic-invariant.test.ts` lost its now-dangling
`vi.mock('@/app/components/board-page/header')` and is otherwise untouched.

**Parked deliberately, recorded in `docs/web-reposition.md`:**
`components/create-climb/` (including `create-climb-form.tsx` and
`drafts-drawer.tsx`) also lost its last route here, but it is not orphaned —
`create-climb-form` still imports the drawer, and the MoonBoard importer still
imports `hold-indicator`, `hold-type-picker`, `use-hold-type-picker` and
`use-moonboard-create-climb` from the same directory. Cutting the rest is a
large cascade that has to lift the importer's pieces out first, which is
`teardown:components-create-climb` in W-16, not this PR.

### Rollback

The recovery plan for this programme is **promoting `release/classic-web`** (#4368 / PR #4428), never a revert — that is the locked decision in `docs/web-reposition.md`. Separately, as a property of the diff: this is pure deletion plus small in-place edits with no code motion, so `git revert -m 1 <merge-sha>` restores every route, provider and catalog key verbatim if the situation is small enough to warrant it.

### W-16 is not openable until this merges

W-16 typechecks only once these layouts stop importing `board-page`, `graphql-queue`, `connection-manager`, `persistent-session` and `queue-control`. W-17 → W-16 is a hard edge in the epic's dependency graph.

## Release Notes

- Board pages slim down to the essentials — browsing and climbing live in the app now.
- Every old board link still works: liked climbs, logbooks and playlists send you to the right place, and creating or remixing a climb opens the app on the board you were already looking at.
- Importing MoonBoard screenshots moved to its own page — pick your board and drop the shots in.

## Test plan

- `vp check` — clean (two pre-existing `docs/design/**.html` formatting reds on `main` are untouched by this branch).
- `vp run typecheck` — green.
- `vp test run --reporter=agent --project web` — green.
- `vp run check:i18n` · `vp run check:i18n:orphans` — green. The orphan checker is what drove the catalog deletions: `climbs.liked.*`, `climbs.metadata.{create,liked,logbook}`, `climbs.moonboard.invalidLayout`, `playlists.metadata.detail.fallbackDescription`, `common.header.searchClimbsPlaceholder` and `common.ariaLabels.{openFilters,clearSearch}`, all four locales. `climbs.metadata.import.*` stays — the relocated importer reads it.
- New: `app/__tests__/next-config-redirects.test.ts` — locale twins for every rule, locale-preserving same-origin destinations, `permanent: false` iff cross-origin, every deleted sibling covered on both trees, the numeric create rule's carried params, and the three-rule `…/import` order. It pins the app origin against a separate `DEFAULT_CONFIG_APP_ORIGIN` export and asserts destinations against the env-resolved `APP_ORIGIN`, so exporting `NEXT_PUBLIC_APP_URL=http://localhost:8081` (what `.env.local` recommends for local Expo-web work) no longer reds a correct config.
- New: `app/components/social/__tests__/comment-section-subscription.test.tsx` — no `createGraphQLClient` call when anonymous, exactly one when signed in.
- New: `e2e/board-route-teardown.spec.ts` — every deleted path × four locales with its `Location` asserted (including the numeric create rule's params, the importer's `sets`, and the non-MoonBoard `…/import` → `…/list` split), `/play` still redirecting with the climb name, and zero product WebSockets on **three** front doors for a signed-out visitor: the config-tuple list, a climb, and the `/b/{slug}` list, each scrolled to the bottom so the comment section is actually mounted and the check can't pass vacuously. The socket filter matches the backend's `/graphql` path positively rather than excluding `/_next`, so the dev server's HMR and RSC channels and the Vercel toolbar can't red a check whose subject is the comment socket.
- `e2e/climb-setter-zoom.spec.ts` was deleted with the setter surface it drove and is **not** re-homed: the board renderer's pinch/zoom now has no e2e coverage. It belongs with the surviving renderer surfaces in W-16 rather than being re-pointed at a front door that renders a plain ``.
- `vp run test:e2e` — **dispatched on this branch and it cannot go green: the
  E2E workflow itself is broken, and has been since June.**
  [Run 31851404126](https://github.com/boardsesh/boardsesh/actions/runs/31851404126):
  all eight Playwright shards die before the first test with
  `browserType.launch: Executable doesn't exist at /ms-playwright/chromium_headless_shell-1234/...`.
  The cause is a version mismatch in the workflow, not in this branch —
  `e2e-tests.yml` pins the container to `mcr.microsoft.com/playwright:v1.60.0-noble`
  while `bun.lock` resolves `@playwright/test` to **1.62.1**, so the browser
  revision the installed Playwright asks for was never in the image. Every
  recorded run of `e2e-tests.yml` has failed the same way (`gh run list
  --workflow=e2e-tests.yml` — nothing green back to 2026-06-08), on branches
  that touch nothing near it. The `expo-web-smoke` job fails separately and
  also pre-existing: a 15s `locator.click` timeout reaching the log-ascent
  sheet, an RN-web surface this PR does not touch. **Bumping that image is a
  one-line fix but belongs in its own PR** — it is a repo-wide CI change with
  no relationship to the board routes, and landing it here would mean this PR
  is the one that reopens two months of unrun E2E results.

  What that leaves as real coverage for the redirect table:
  `app/__tests__/next-config-redirects.test.ts` loads the actual
  `next.config.mjs` and asserts the same sources, destinations and permanence
  flags the spec checks over HTTP. Separately, the emitted `Location` for every
  new rule was verified locally by driving Next's own route matcher
  (`getPathMatch` + `prepareDestination` from `next/dist/shared/lib/router/utils`)
  over the config: `/kilter/1/10/1,2/40/create` →
  `https://app.boardsesh.com/climbs/create?boardName=kilter&layoutId=1&sizeId=10&setIds=1%2C2&angle=40`,
  `/kilter/original/…/create` → the bare app URL, `/moonboard/3/17/5,8/40/import`
  → `/moonboard-import?layout=3&sets=5%2C8&angle=40`, and
  `/kilter/1/1/default/40/import` → `/kilter/1/1/default/40/list`.
  `board-route-teardown.spec.ts` stays in the tree and covers all of it the
  moment the workflow is repaired.


marcodejongh:
**Fable review: approved, pending the e2e dispatch.** Final pre-merge read per the epic #4358 rule for W-17 (🔒).

Verified directly on the final tree: `/play` is byte-unchanged on both trees (empty scoped diff); `crawler-classic-invariant.test.ts` changed by exactly one deleted `vi.mock` line (the sanctioned class — the fixer deleted the orphaned `board-page/header.tsx` that mock referenced; documented in the body); the redirect DoD is encoded as real assertions (locale twins ×4, same-origin locale preservation, cross-origin never-permanent, same-origin permanent); the comment socket opens zero clients anonymous / exactly one signed-in, by test. The QA round earned its keep: the site-wide Create tab and the importer's hold-set loss were both real majors caught pre-merge, and the numeric-layout collision rule ordering is right.

Also shipped here: the e2e workflow's container pin was three versions behind the locked Playwright (`v1.60.0` image vs `1.62.1` lock) — every shard died on a missing browser build since the bump, undetected because the workflow is dispatch-only. Fixed and re-dispatched on this branch (run 31852676962); **hold merge until that run's E2E shards are green** (the expo-web-smoke job is a separate heavier suite — a flake there shouldn't gate this).

Merge order reminder: W-16 (#4435) is not openable until this merges.

marcodejongh:
**e2e verification complete — Fable approval now unconditional.** The full suite is green on this branch (run 31860014178: all 8 shards; Expo-web smoke's play-drawer failure is pre-existing on main's baseline and non-gating). The shard-7 saga closed with three durable artifacts: the Playwright image pin (the suite had been silently unrunnable since the 1.62 bump), `workers: 1` per shard (the 8-way shard fan-out carries parallelism), and the `aa-` ordering pin on the embed header checks — a browser spec poisons the server for subsequent raw HTTP requests, now tracked with full forensics + repro recipe in #4463 (pool sizing before the climb sitemaps: #4461). Ready for merge; W-16 (#4435, plan complete) opens immediately after.

PR #4458, opened 2026-08-15
feat: standings privacy consent, tick write-path bounds, leaderboard retrofit

### LINKED ISSUE #4189: Let climbers keep their sends off gym wall feeds
Raised by @ES-Alexander reviewing #4187:

> there may also be some users who only care about the type of board they're climbing on, and don't want their sends to contribute to board stats / be usable to track their location across gyms or something.

## The gap

Connecting to a gym's wall binds you to that gym's board id, and your sends then appear in its public wall feed with your display name and avatar (`boardHistory`, `boardNowPlaying`, `boardPresenceStats`). Anyone reading that feed can see you were at that gym at that time. There is no way to opt out.

This is **not** new behaviour and #4187 doesn't change it — the serial has always routed climbers to the gym board when one carries it. #4187 only makes the routing pick the *right* gym board. Filing it separately so it gets designed on its own terms rather than smuggled into a bug fix.

## Shape of a fix

A per-user preference along the lines of "keep my sends off gym wall feeds". When set, a connect would bind the climber's own board (or the per-config shared feed) instead of the gym listing, so the LEDs and their logbook work exactly as now but nothing lands in `board_climb_events` under the gym's board id.

Open questions worth settling before building:

- **Scope.** Global, or per gym? A climber might be happy showing up at their home gym and not at a competitor's.
- **Retroactive?** `board_climb_events` rows already attribute past sends. Does enabling the setting hide/remove them, or only apply going forward?
- **Ticks vs wall feed.** `boardsesh_ticks.board_id` also carries the gym board. Board *stats* (`boardPresenceStats` — sends count, distinct climbers, hardest grade) are derived from ticks. Does opting out mean not contributing to those either, and is a gym's leaderboard meaningfully degraded if a slice of climbers opt out?
- **Discoverability.** A setting nobody finds doesn't help. Onboarding mention, or surfaced the first time a send lands on a gym feed?

## Prior art in the codebase

`user_boards.hide_location` already expresses "this board opts out of exposing where it is", and `findActiveBoardsBySerial` / `toBoardCandidate` redact location for boards that set it. A climber-side equivalent would sit next to it conceptually.

marcodejongh:
## Summary

Phase 0 of **Standings** — Strava-style leaderboards for the mobile app. **No ranking UI ships here.** This is the safety and correctness layer the ranked surfaces sit on, and it is worth shipping on its own: it closes #4189 and four confirmed write-path holes with no new surface to defend.

Design work behind it: 24 subagents, ~350 read-only queries against production, and a design panel of 5 concepts each attacked by 2 adversarial critics. The plan (scope model, scoring, wireframes) is agreed; Phase 1 builds the route.

This **reverses the recorded no-leaderboards non-goal** in `docs/achievements-plan.md` — an explicit product-owner call. That doc's deferred anti-cheat item said *"revisit if leaderboards ever exist"*; the condition has fired, so its resolution is recorded there too. The owner's position is that the community keeps the system honest rather than a technical regime, so what ships here is plain input validation, not policing.

Closes #4189.

### Privacy — the schema had no opt-out of any kind

A full-schema scan for privacy-shaped column names returns 7 columns, **none on `users`, `user_profiles` or `boardsesh_ticks`**. A climber could not prevent their name and send count being published.

Two **independent** consent columns, because being named on a screen inside the gym you are standing in is a different decision from being named in an app a stranger can open:

| | column | audience |
|---|---|---|
| In the app | `leaderboard_visibility` | Boardsesh's own ranked surfaces |
| On gym screens | `gym_screen_visibility` | kiosk rail, wall feeds |

Each is `public` / `anonymous` / `off`. **`anonymous` still ranks and still counts toward the field size** — only the identity is withheld. **`off` leaves the denominator too**, applied inside the ranking query, so hiding someone cannot silently shift everyone else's rank.

Both default to `public` per the owner's call. For the kiosk that is a **no-op** — it publishes names with no gate today — so no gym running one sees its rail change under it.

New mobile settings screen at **More → Standings**, four locales. Privacy policy gains a "Standings and public rankings" section; `/legal` had no mention of public ranking at all.

### `boardLeaderboard` retrofit

The query already shipped (`social/boards.ts`) and is rendered by web's board page **and** the gym kiosk rail. It gains:

- The visibility filter, plus a `surface` input (`app` | `gymScreen`) so the kiosk reads its own column. **Defaults to the in-app column**, so a caller that forgets it under-publishes rather than over-publishes.
- **Anonymous entries return a stable HMAC pseudonym, not the real user id.** Returning the real id beside `isAnonymous: true` would have leaked the withheld name one tap away via `/users/[userId]`. Returning `null` would have fused every anonymous climber into one row in the kiosk's cross-board merge, which keys on user id.
- `totalSends` counts **DISTINCT climbs**, not tick rows. Re-logging the same climb was the cheapest way up the ranking — production holds **41,598 excess same-day repeat rows**.
- **`RANK()`** replaces `offset + idx + 1`. Tied climbers now share a rank, and the in-tie order is stable across refetches instead of shuffling under the reader.
- Excludes `users.is_internal`.
- `periodLabel` marked deprecated — it returns a hardcoded English string, so no localized client can use it.

### Tick write path

Three holes, all confirmed by reading the code, all of which a ranked surface makes exploitable:

1. **`SaveTickInputSchema.climbedAt` had no future-date guard.** The 60s tolerance existed on `UpdateTickInputSchema` only, so a *created* tick could be dated 2099 and sit inside every rolling window forever. This is the most important fix here.
2. **`difficulty` was an unbounded `z.number().int()`** on both paths. One row carrying `INT_MAX` permanently owns any surface reading `MAX(difficulty)`. Now bounded to the legal 1–39 grade scale, mirrored by a DB check.
3. **`saveTick` had no rate limit** beyond Yoga's depth/cost caps. Now 120/min per user — far above the 99th-percentile climber's **25 climbs per _day_**, and above a burst drain of the offline outbox. Bulk logbook ingestion has its own handler and is unaffected.

### ⚠️ A rate limit would have silently dropped climbers' offline sends

Found while adding #3 above, and worth calling out because it is a live bug independent of this feature.

`applyRateLimit` threw `extensions.code: 'RATE_LIMITED'` — a **string**. The offline mutation drainer resolves retryability from a **numeric** `extensions.code`/`status` (`offline-sync/.../error-classification.ts`). A rate-limited mutation therefore resolved to "no status", which `isRetryable` treats as a programmer bug and **dead-letters**.

So adding a `saveTick` limit without this fix would have **dropped a climber's queued session** on a burst drain instead of backing off for a minute. Fixed at both ends: the backend now also sends `status: 429`, and the drainer maps the string code to 429 so a current client survives an older backend.

### Data hygiene

- **`user_boards.is_virtual`**, backfilled for the 39 `"<X> Board Shared Feed"` system rows. They are not walls — they absorb ~7,780 ticks and hold the **top four slots by climber count** (179/165/111/84) ahead of every real board, the largest of which has 17 climbers in 30 days. Verified: the `name LIKE '%Shared Feed%' AND gym_id IS NULL` predicate matches **exactly 39** rows, where `gym_id IS NULL` alone matches 1,064 and would have flagged ~1,025 real walls.
- **Partial index** on `(climbed_at) WHERE status IN ('flash','send')`, mirroring the existing `updated_at` sibling.

### Two things I deliberately did *not* do

**1. `users.is_internal` ships defaulted `false` with no backfill.** A regex over email domains is exactly the wrong tool: `testingkerry@gmail.com` (20 ticks) and `rick@pentester.com` are real people whose addresses contain "test", and flagging them erases them from every ranking. The 18 unambiguous accounts are listed below — **needs your sign-off before I run it**:

| ticks | name | email |
|---:|---|---|
| 1183 | testanother | test22@boardsesh.com |
| 223 | Test User | test@boardsesh.com |
| 16 | test | test@test.com |
| 0 | Boardsesh | system@boardsesh.com |
| 0 | Boardsesh | recommendations@boardsesh.com |
| 0 | Boardsesh Hold Classifier | hold-classifier@boardsesh.com |
| 0 | Nuage Laboratoire ×3 | …@cloudtestlabaccounts.com (Firebase Test Lab) |
| 0 | 9 others | @boardsesh.com / @example.com / test@test.org |

`testanother` alone has 1,183 ticks and would rank **#101 all-time**.

**2. The difficulty check is `NOT VALID` on purpose — please don't "fix" it.** It is fully enforced for every INSERT and UPDATE from here on, which is the entire point. It skips validating the 421k existing rows because **two legacy production rows (ids 93195, 238792) carry `difficulty = 0`**, and a plain `CHECK` would fail the migration outright. Nulling them means "use the consensus grade" per `docs/ascents-and-attempts.md`, but it rewrites climber-entered data, so it is left for an explicit call rather than smuggled into a schema migration. Once cleared, `VALIDATE CONSTRAINT` takes only a `SHARE UPDATE EXCLUSIVE` lock.

### Also worth flagging

The generated migration wanted to emit `ALTER TABLE "gyms" DROP COLUMN "website_vouched_by_owner"` — **not part of this change**. That column exists in the `0192` snapshot but **no migration ever created it** and production does not have it; it is a ghost from the unbundled gym-vouching work. I removed the line (it would have errored on prod). The `0195` snapshot no longer carries it, so future generates are clean, but someone should confirm that is the intended end state.

## Release Notes

- Choose how you show up in rankings. Under More → Standings you can be named, counted without being named, or left out entirely — and you can answer it separately for the app and for screens in the gym.

### Author-side verification

Typecheck, lint and format are green across the monorepo. Backend suite passes,
including 6 new `boardLeaderboard` tests (opt-out leaves the denominator;
anonymous keeps rank but loses name, avatar and real id, and is stable across
calls; the surface split; internal exclusion; distinct-climb counting; shared
ranks with stable in-tie order) and 5 new validation tests (future date on
create, clock-skew tolerance, difficulty bounds on both schemas). Mobile suite
passes, 6,447 tests over 668 files. Kiosk leaderboard model gains 2 tests: an
anonymous climber merges across boards by pseudonym, and two different
anonymous climbers do not fuse into one row. All four locales are at catalog
parity and the i18n scanner is clean.

The migration was applied against a throwaway `postgres:17` reproducing the
exact production edge cases. It applies cleanly over the two legacy
`difficulty = 0` rows, marks only the Shared Feed row (a garage wall correctly
untouched), rejects `INT_MAX` and `0` on new writes, and preserves the legacy
rows. Rebased onto main and renumbered to 0226.

One local-only trap worth recording: an earlier backend run showed 20 failures
that were entirely stale per-worker test databases left by a killed run. Global
setup cannot drop a worker database a connection is holding, so the worker
reuses the old schema. Force-dropping them made everything pass.

## Test plan

1. More → Standings. Two groups: in the app, gym screens.
2. Pick "Counted, not named" for the app. Reopen. Choice stuck.
3. Open a board leaderboard. Your row reads "A climber".
4. Pick "Keep me out". Reopen. Your row is gone.
5. Log a tick dated tomorrow. Readable rejection, no crash.

## Risk

Risk: 5/5 — a migration plus new consent columns on the tick write path.

https://claude.ai/code/session_016iuGGq7XAdw59knjNqMRgV


marcodejongh:
## Prod spot-check of the retrofit query shape

Ran the new shape read-only against production, on the busiest **real** board in the last 30 days — *Grip It Climbing – Nelson Rd, Kilter Board Homewall*, 17 climbers.

```
scan type:      index scan (not seq)
Planning Time:  1.019 ms
Execution Time: 2.183 ms
```

Visibility columns aren't deployed yet, so `COALESCE(NULL::text,'public')` stands in for them exactly as the resolver reads them — the `users` + `user_profiles` joins the filter rides on are in the plan.

**The tie behaviour is not a corner case.** Ranks come back 1, 2, 3, 4, 5, 5, … and **8 of 17 rows share a rank with someone** — on the busiest real board in the database. The old `offset + idx + 1` gave every one of those climbers a different, arbitrary number that could change between refetches. That's the concrete reason for `RANK()` plus stable in-tie ordering rather than an invented tie-break.

Distinct-climb counting is also not a regression on cost: 2.18 ms here versus the 39 ms measured for the old per-board shape during design (different board sizes, but the DISTINCT + window function clearly isn't the expensive part — the joins are, and they're index-backed).

marcodejongh:
Cross-link: the `RATE_LIMITED` dead-letter fix carried on `feat/standings-phase0`/`phase1` (commit `06e0d0671`) is **not on `main`** yet, so `isRetryable` still dead-letters string-coded `RATE_LIMITED` mutations. Filed as #4711 so it doesn't get lost or fixed twice — if this PR lands, #4711 closes with it.

marcodejongh:
## Review findings addressed — `be4a7f8d4`

Six taken, two dismissed after checking the code.

### Fixed

| Finding | What changed |
|---|---|
| `hardestGrade` truthiness drops grade 0 | `!= null`. Those rows are exactly why the new check is `NOT VALID`, so they are reachable, not hypothetical. |
| `accessibilityState.selected` documented but never passed | `ListRow` had no such prop. It now forwards `accessibilityRole`, `accessibilityState` and `accessibilityValue` to `PressableSurface`, which already accepted all three. The rows declare radio semantics and their position. |
| `updateTick` has no rate limit | Same 120/min ceiling as create. `climbedAt` is editable, so an edit can move an old tick into the rolling window — the same capability a create has, and pointless to bound on one path only. |
| Dev HMAC key can stand in silently | Throws in production when `NEXTAUTH_SECRET` is unset. A key checked into this repo makes the pseudonym reversible by anyone who can read the source, which is the one property the function exists to provide. |
| Rotation coupling undocumented | Recorded on the function: rotating changes every pseudonym, nothing is corrupted, but a kiosk merging a climber across boards splits them into one row per board until the ranked pages refetch. |
| Missing trailing newline in the migration | Added. |

### Tests added

Both rate-limit tests are mutation-tested: deleting either `applyRateLimit` call turns them red. That was the point of the finding, since the constants are otherwise unobservable.

The visibility test sits at the Zod layer because the backend test database declares those columns as `text` rather than the enum, so a wrong-cased value passes in tests and fails in production. Zod is the only place that difference is caught.

### Dismissed

**`periodLabel` deprecation.** Nothing renders the server-returned value. The kiosk settings screen builds its own labels locally, so the deprecation tag leaks no English-only string to a user.

**French and Spanish masculine forms.** Both follow their own glossaries. `docs/i18n-french-glossary.md` documents the masculine default for a bare status chip and gives `grimpeur / grimpeuse`; the Spanish glossary gives `escalador / escaladora` with no gender-star convention. Only the German glossary mandates the star form, which is why German alone carries it.

### Deferred, per the owner's call

`is_virtual` and `is_internal` both ship unread and unbackfilled, and the reasoning now lives in the schema rather than only here. Neither is observable until climbers are ranked **across** boards, which is Phase 1 — `boardLeaderboard` ranks within a single board, so a Shared Feed row cannot inflate anything today. Flagging an account stays a named decision: `testingkerry@gmail.com` has 20 real ticks and `rick@pentester.com` is a real person, so any predicate over the address erases climbers it should leave alone.

### Also in this push

Rebased onto `main` and renumbered the migration to **0226**.

The branch's `RATE_LIMITED` classifier mapping is **gone**. `main` landed a broader fix for the same problem: the drainer now retries anything outside `PERMANENT_REJECTION_STATUSES`, so a string-coded rejection already drains. Mapping it to 429 was redundant and contradicted the contract `main` documents in its own test. The backend still sends `status: 429` alongside the string code, which is additive and useful — it just is not what makes the mutation retryable.

PR #4460, opened 2026-08-15
feat: standings scope registry, ranking query and mobile route

marcodejongh:
## Summary

Phase 1 of **Standings** — the ranked surface itself, on top of Phase 0's consent and write-path work.

> **Stacked on #4458.** Base it there or merge that first; this branch contains both commits.

### The scope registry — the extensibility spine

New shared package `@boardsesh/leaderboard`. Pure TS, no React, no platform I/O.

A scope is `{ kind, key }`. Each kind declares its own SQL key, fallback parent, and **measured attribution coverage** — so a surface can tell a climber *why* a gym number looks low instead of silently under-reporting:

| kind | key | attribution | scopes alive @30d |
|---|---|---:|---|
| `global` | — | 100% | 1 (1,203 climbers) |
| `boardType` | `kilter` \| `tension` \| `moonboard` | 100% | 3 |
| `layout` | `boardType:layoutId` | 99.99% | 12 of 16 |
| `board` | `user_boards.uuid` | 86.8% (3% beyond 90d) | 808 active, max 17 |
| `gym` | `gyms.uuid` | 46.9% | 590 active, max 17 |

Two deliberate choices worth reviewing:

- **Board and gym fall back to the *setup*, not the board type.** "Every Kilter Original at this angle" is much closer to what the climber was looking at than "every Kilter anywhere".
- **Only three board types are registered.** `decoy` (6 climbers in 30 days), `touchstone` (2) and `grasshopper` (1) would each render a permanently dead tab.

### The `standings` query

Ranks **DISTINCT climbs topped** in a rolling window, capped at **30/day** — the measured p99 (p50 is 4, max ever 373), so it never touches a real session but bounds what a script buys.

- **`RANK()`**, so tied climbers share a rank, with a deterministic in-tie order. Not academic: on the busiest *real* board in production, **8 of 17 rows share a rank** (measured, see #4458).
- **No all-time at any scope.** 69.4% of the tick history is a frozen import whose newest row is 2026-03-26 — an all-time board would rank whoever uploaded a file, and a rolling window cannot reach it anyway.
- **The viewer's rank comes from a window function**, never from paging until the client finds itself. It carries the *distinct scores above*, so the UI can say "two more and you're 81st" — naming a rank, never the person standing on it.
- Excludes the frozen import, `is_internal` accounts, `is_virtual` Shared Feed pseudo-boards, and opted-out climbers. An anonymous climber keeps their rank but gets a pseudonym — **except the viewer, who always sees their own row**, since they know who they are and the row has to be pinnable.
- **Empty or unresolvable scopes demote to global** with a machine-readable reason rather than rendering a blank list. There is deliberately no path here to "No data available".

Written as raw SQL, matching `computeBoardPresenceStats`: it needs a window function over a grouped subquery over a per-day `ROW_NUMBER` cap, which the query builder can't express. Scope predicates stay composable `SQL` fragments so adding a granularity is still a one-place change.

### Mobile route

`/(tabs)/profile/standings`, a pushed route — the tab bar is at its 5-tab limit and a 6th spills into iOS "More".

FlashList, `React.memo`'d rows, hoisted `keyExtractor`, `renderItem` deps with no `.length` and no inline closures, one page per end-reach with no drain loop. Offline outranks both spinner and error, since an offline fetch *pauses* rather than failing under `networkMode: 'offlineFirst'`.

**Monogram-first avatars by design, not as a fallback**: 83.7% of active climbers have no photo, so a photo-shaped row would render broken five times in six.

### The extensibility promise is enforced, not just documented

`scope-extensibility.test.ts` asserts that **no rendering component names an individual scope kind**. Scope-specific behaviour cannot drift out of the registry and into the screen without failing CI.

I mutation-tested it rather than trusting it: adding a `head?.resolvedScope.kind === 'gym'` branch to `StandingsScreen` makes it fail with the intended message, and the file was restored clean.

The one mobile file that must know kinds exist is `scope-labels.ts`, because the i18n linter hard-fails on a dynamic `t()`. Its template literal is pinned to the registry's declared `labelKey` by a test, so the two can't drift apart.

## Release Notes

- See where you stand. Standings ranks climbers on the different climbs they've topped in the last 30 days — everyone, your board, your setup, your wall or your gym.
- Ties share a rank, so if you're level with 47 other climbers it says so, and one more climb breaks you out.

## Test plan

- `vp run typecheck` — green across the monorepo.
- `vp check` — no findings in any file this PR touches (28 pre-existing errors elsewhere, all untouched here; PR CI diff-lints).
- `vp test run --project backend standings` — **17 passed**, covering distinct-climb scoring, the per-day cap (and that it does *not* apply across days), shared ranks + tie size, exclusion of the frozen import / internal accounts / attempts, opt-out leaving the denominator, anonymity from other viewers but not from yourself, per-board / per-gym / per-layout scoping, Shared Feed refusal, empty-scope demotion, the viewer block and its `scoresAbove`, the signed-out reader, and both rolling windows.
  - One of these caught a real bug: the `layout` scope — the *highest-coverage* tier at 99.99% — was entirely broken by joining `board_climbs` on `climb_uuid` when the table keys on `uuid`.
- `vp test run --project leaderboard` — 23 passed (registry, fallback ladders, key round-trips, tie/percentile rules, next-rank arithmetic).
- `vp test run scope-labels scope-extensibility` — 6 passed.
- `vp run test:mobile` — **6,454 passed** (670 files).
- `vp run check:i18n` and `check:i18n:orphans` — both OK; `catalog-completeness` 90 passed (all four locales at parity).

Not yet exercised on a device. The screen is a pushed route with no new native surface, but it should get a simulator pass before merge — and I'd particularly want eyes on the tie-block viewer card and a solo board, since those are the two states the measured data says most climbers will actually land on.

### Entry points

- **You tab** — `StandingsEntryCard` sits beside the stats summary rather than inside it: it has its own query and its own reason to be absent, and coupling them would make one card wait on the other. It renders **nothing** until there is a real rank to show, so a climber who logged nothing this month gets silence rather than "—th of —" (2,694 of 4,908 registered climbers have never logged a climb at all).
- **More** — a direct row, so the surface is reachable without stats loaded.

### Still to come

The scope picker (switching between wall / setup / gym / everyone in-app; the query already supports every scope) and the post-send rank card. Both are Phase 2.

### One correctness fix worth calling out

The resolver returned a hardcoded `coverage: 1` while the registry held the measured values — so a gym ranking, which can only attribute **46.9%** of recent sends, was reporting perfect coverage. That is the API asserting something false, and the surface uses that field to decide whether to explain a low number. It now reads the registry and reports the coverage of the scope actually **resolved**, so a demoted request describes the board it shows rather than the one that was asked for.

https://claude.ai/code/session_016iuGGq7XAdw59knjNqMRgV


marcodejongh:

Conflicting: packages/mobile/app/(tabs)/profile/more.tsx plus i18n settings.json conflicts across de/en-US/es/fr.

Issue #4442, opened 2026-08-14
W-26: web scope-down cleanup — config, docs, CI parity


marcodejongh:
Part of #4358. **Blocked by:** #4435 (W-16 — delete the climbing UI) and #4438 (W-24 — retire the `/app` static path). QA gate #4359 is **SIGNED and closed (2026-08-14)**.

This is the tail of the epic — the PR that lets the epic close.

### `next.config.mjs`

- Prune the **`transpilePackages`** entries with **no remaining web consumer** (**recount at branch time**).
- **Keep `outputFileTracingIncludes`** — the WASM include is *more* load-bearing after the teardown: the front door, `/api/og/*`, kiosk, and `board_config.h:41` pinning `www.boardsesh.com` as the ESP32's render host all depend on it.

### Docs sweep

- **Delete `docs/view-page-removal-pivot.md`**, moving its `ClimbViewSeoFragment` rule into `docs/web-reposition.md`.
- Fold **`docs/expo-web.md`'s two durable facts** into `docs/expo-web-deployment.md`.
- **Mark A1–A6 done** in `docs/web-reposition.md`.
- **Record the four couplings the A0 audit missed** — the edges W-05's import-graph test caught that every designer missed:
  - `social/proposal-card.tsx:36` → `climb-card/climb-list-item`
  - `social/comment-section.tsx:13` → `graphql-queue/graphql-client`
  - `climb-card/ascent-status.tsx:12` → `board-provider/board-provider-context`
  - `activity-feed/*` → `climb-card/climb-icons`

### CI env parity

Extend **`scripts/mobile-ci-env-parity.test.ts`** to cover **`production-deploy.yml`** — it guards six workflows and **misses the places that build a *shipping* SPA bundle**.

**Watch the indentation.** W-01 (#4065) hoisted the `EXPO_PUBLIC_*` vars into a **workflow-level `env:` block at 2-space indent**, and **that indent is load-bearing**: `scripts/mobile-ci-env-parity.test.ts:87-89`'s extractor matches only `^  KEY:`.

## Standing-rule callouts

- **Never delete `/api/internal/board-render`, `/api/internal/ws-auth`, or the config-tuple `/list` route** — each is pinned by shipped hardware or a kept caller, and `outputFileTracingIncludes`' WASM include exists for the first of them.
- **Fingerprint parity**: any change touching the env a native build resolves must keep `app.config.ts` resolving identically for the OTA publish — that is exactly what the parity test guards.
[...]

Issue #4438, opened 2026-08-14
W-24: retire the /app static path on www


marcodejongh:
Part of #4358. **Blocked by:** #4435 (W-16 — delete the climbing UI). QA gate #4359 is **SIGNED and closed (2026-08-14)**, so the gate is satisfied.

🔒 **Point of no return.**

## Context

`app.boardsesh.com` is the Cloudflare Pages subdomain root and is in production. Next still serves `/app` as the dev Metro proxy plus a legacy static fallback; this PR retires the **production** branch and keeps the dev one.

## The work

- Delete the **production branch** of `next.config.mjs`'s `rewrites()`, the **two `/app/*` header rules**, and **`Dockerfile.web:84`'s export step**.
- **Keep** the dev Metro proxy and `scripts/build-expo-web-export.sh` (the subdomain deploy calls it with `--subdomain`).
- **Fix the PWA manifest**: `packages/mobile/public/index.html:15` hardcodes `/app/manifest.json`, which on the subdomain is swallowed by the `/*` catch-all and served as HTML — **no install prompt, console error every load**. Have the export script rewrite the href and patch `start_url`/`scope` from `$WEB_BASE_URL`, **asserted post-export**.
- **Must land before #3795.** On today's Vercel this is a no-op cleanup, but #3795 turns `Dockerfile.web`'s default export into a **live second SPA copy at `www/app` built with different env**.
- **Check first whether `BOARDSESH_WEB` is set as a Vercel project env var** — a dashboard variable falsifies the premise. Record the check in the PR body.

## Standing-rule callouts

- **Fable reviews the final diff** — W-24 is on the Fable point-of-no-return list, and **every PR touching `packages/mobile` gets a Fable native-OTA review regardless of tier**.
- **Standing rule 2**: state which of the three production targets this reaches (`app.boardsesh.com` / the native store fleet via auto-OTA / neither). Anything claimed web-only carries a **test asserting the native branch is byte-unchanged**. Fingerprint-neutrality guarantees *delivery*, not safety.
- **Keep `BOARDSESH_WEB=1` gating** — web-only configuration stays behind that gate so native OTA fingerprints do not change.
- **`deploy/app-subdomain/_redirects`' `/* /index.html 200`** means a missing SPA route returns **HTTP 200 with the shell** — no 404, no 5xx, no log line. Do not rely on error codes to detect a broken path.
[...]

PR #4480, opened 2026-08-15
chore(web): retire the /app static path and fix the PWA manifest

marcodejongh:
## Summary

Part of #4358 (W-24). Refs #4438. 🔒 **Point of no return** — and it touches `packages/mobile`, so **Fable reviews this diff before merge**.

`www.boardsesh.com/app` was a legacy static-export path: Next serving an Expo web export out of `packages/web/public/app`, with two SPA-fallback rewrites and two `/app/*` header rules. It never served anyone. This retires it, and fixes the PWA manifest on the subdomain that actually does serve the browser app.

**What changed**

- **`packages/web/next.config.mjs`** — both `/app` header rules deleted; the SPA-fallback rewrites now sit behind `process.env.NODE_ENV === 'development'`. The dev Metro proxy block is **byte-unchanged** (verified in the diff).
- **`Dockerfile.web`** — the Expo export step is gone, along with the builder-stage `ARG/ENV BOARDSESH_WEB` and the `apk add bash` that existed only to run it. The **runner-stage** `ARG/ENV` stays: `BOARDSESH_WEB` is read at *request* time for the cross-subdomain Expo auth bridge (`app/layout.tsx`) and the `/app` middleware carve-out. `--build-arg BOARDSESH_WEB=0` is no longer a rollback lever — setting it to 0 now only breaks sign-in on the subdomain, and the rewritten comment says so.
- **`scripts/build-expo-web-export.sh`** + new **`scripts/lib/patch-expo-web-pwa-manifest.mjs`** — the PWA manifest fix (below).
- **`.github/workflows/production-deploy.yml`** — a `check_manifest_link` post-deploy smoke against `app.boardsesh.com`.
- Docs: `docs/expo-web-deployment.md`, `docs/expo-web.md`, `deploy/app-subdomain/README.md`, and the `build:expo-web` / `build:web` task comments in `vite.config.ts`.

### The user-visible half: the PWA manifest

`packages/mobile/public/index.html` hardcodes `<link rel="manifest" href="/app/manifest.json">` and `manifest.json` is copied verbatim with `start_url`/`scope` = `/`. Neither knows which `baseUrl` an export was built with. On `app.boardsesh.com` that href pointed at a path with no file behind it, so `_redirects`' `/* /index.html 200` catch-all answered it **with the SPA shell** — the browser parsed HTML as a manifest, logged an error every load, and never offered the install prompt. All at HTTP 200, invisible to a status-code check.

`patch-expo-web-pwa-manifest.mjs` runs after the WASM assertion and rewrites both files from `WEB_BASE_URL`:

| mode | href | `start_url` / `scope` |
|---|---|---|
| default | `/app/manifest.json` | `/app/` |
| `--subdomain` | `/manifest.json` | `/` |

Three deliberate properties:

- **It re-reads both files from disk and asserts the values landed.** A silent no-op is precisely the failure mode being fixed, so an in-memory check would not be an honest oracle.
- **It fails if `%WEB_TITLE%` / `%LANG_ISO_CODE%` survive** — that means `copyPublicFolderAsync` clobbered the rendered shell with the raw template, a failure an href-only assertion cannot see (the href would look perfect).
- **It runs in both modes**, not just `--subdomain`. The CI gate (`vp run check:mobile-web-bundle`) only ever exercises the default mode; a `--subdomain`-only patch would stay untested until a production deploy, which is exactly how this bug survived.

`start_url` gets the trailing slash because PWA scope matching is a path-prefix string compare — `/app` does not sit inside `/app/`.

**The template href does not change.** `/app/manifest.json` is the correct *dev* value (Metro serves `public/` from its root; `next.config.mjs` rewrites `/app/manifest.json` → Metro's `/manifest.json`) and it is the patcher's expected input. `web-shell-appearance.test.ts` now pins it with a comment pointing at the patcher, so the natural-but-wrong hand-fix reds.

---

## Production targets (standing rule 2)

| Target | Reached? | Why |
|---|---|---|
| **`app.boardsesh.com`** (Cloudflare Pages) | **YES** | This PR edits `scripts/build-expo-web-export.sh`, which `scripts/production-deploy-changes.mjs:85` classifies as app-affecting, so merging fires `deploy-app-web` and republishes the subdomain with the manifest fix. **This is the intended release.** |
| **Native store fleet via auto-OTA** | **NO** | The only `packages/mobile` file touched is `public/index.html`, and only its comment block. `public/` is not a fingerprint source and is not read by any native code path — proof + test below. `mobile-ota-production.yml` is untouched; no `eoas publish` is involved. |
| **`www.boardsesh.com`** (Vercel) | **YES, cosmetically** | `next.config.mjs` + `Dockerfile.web` are web-affecting, so `deploy-web` runs. Behaviour on production www is **unchanged**: `/app` 404s before and after. |

Also note: this PR touches `.github/workflows/production-deploy.yml`, which `isProductionDeployControlFile` treats as **all-targets** — so the merge deploys web + backend + app regardless. Expected; flagging it so nobody reads the backend deploy as a surprise.

### Why the native claim is a test, not an assertion

`packages/mobile/src/__tests__/web-public-assets-are-web-only.test.ts` (new) pins both halves:

1. **Not a fingerprint input.** `fingerprint.config.js`'s `extraSources` are `fingerprint.config.js`, `locales/`, `../../patches` — nothing in or under `public/`. `@expo/fingerprint`'s own defaults source the bare native dirs, autolinking configs, resolved Expo config, `package.json` scripts and the patch dir; `public/` is in none. No `.fingerprintignore` in the repo.
2. **Never read by native code.** Expo consumes `public/` only on the web export path (`@expo/cli`'s `publicFolder.js` + `webTemplate.js`), and `assetBundlePatterns` is `['assets/**/*']`, so Metro never bundles it into a binary. The test walks `src`, `app`, `modules`, `plugins`, `targets` and fails if any non-test file references `public/index.html` or `public/manifest.json`.
3. It also fails if `app.config.ts` ever gains an `EXPO_PUBLIC_FOLDER` override, which would move the file out of the paths (1) and (2) cover.

Fingerprint-neutrality alone would only guarantee *delivery*; (2) is the safety half — the native binary never reads the byte we changed. All three assertions were mutation-probed (below).

---

## The `BOARDSESH_WEB` Vercel-env check (Verifies checkbox)

The issue asks to check whether `BOARDSESH_WEB` is set as a Vercel project env var, because a dashboard variable would falsify the premise. **It cannot**, and there are two independent proofs:

- **Behavioural (measured 2026-08-15 against production www):** `/app`, `/app/index.html`, `/app/manifest.json`, `/app/climbs` and the three locale prefixes (`/es/app`, `/fr/app`, `/de/app`) all return **404**. `/app/` 308s to `/app` — that is Next's site-wide trailing-slash normalisation, not an `/app`-specific rule, so there is nothing to remove there.
- **Structural (in-repo):** the export artifact is never built for that deploy at all. `packages/web/vercel.json:3`'s build command resolves to `generate:openapi && next build`, and `production-deploy.yml:191-217` runs `vercel build --prod`. **Neither chain invokes `scripts/build-expo-web-export.sh`.** `.gitignore:92` ignores `packages/web/public/app` and `git ls-files packages/web/public/` returns zero entries under `app/`. So the rewrite resolved to a file that does not exist **whether the flag is set or not**. `docs/expo-web-deployment.md:326-328` already documented exactly this.

No dashboard read was needed and none blocked this PR.

**One thing filed, not fixed:** if the flag *is* unset on the Vercel project, `app/layout.tsx:87`'s `enableExpoAuthBridge` is off in production — the www-side half of the cross-subdomain auth bridge `app.boardsesh.com` depends on. Pre-existing and orthogonal to the retirement; deliberately untouched here.

---

## Decisions worth reading before reviewing

- **No redirect for `/app`.** 0 pageviews over 90 days (PostHog project 412845) against a non-vacuous 55,442-pageview www control; no in-repo `LocaleLink`/`<a>`, no sitemap shard entry, no `robots.txt` `Disallow`, `noindex` since inception. A redirect would create a cross-origin hop for a path nobody requests and drag in three locale twins (standing rule 4) for a path that never had locale variants.
- **The static fallback survives behind `NODE_ENV=development`, rather than being deleted.** `scripts/dev-expo-web-static.sh` (`vp run dev:mobile:web-static`) bakes the default `/app` export and starts the orchestrator with the web flag on and **no** proxy origin — it lands in exactly this branch, and `docs/expo-web.md` calls it *"Best for QA/device testing"* (the tailnet real-device loop). The dev gate keeps that alive **and** still delivers what #3795 needs: no production build, Vercel or `Dockerfile.web`, can bake an `/app` route again. Say the word in #3795 and its constraint comment can go.
- **`middleware.ts` is deliberately untouched.** Its `/app` carve-out is what keeps `noindex`/XFO/nosniff/HSTS on the dev Metro-proxied surface, since an external rewrite forwards Metro's headers past `next.config`'s `headers()`. Tightening it to `isExpoWebProxyEnabled()` is a behaviour change on a 🔒 PR for no production benefit — **out of scope on purpose**, not a miss.
- **`scripts/__tests__/mobile-web-bundle-check.test.ts` stub edit is required, not scope creep.** Both of its `bunx` stubs `touch`ed an empty `index.html` and wrote no `manifest.json`; the new patch step would fail every case for the wrong reason. The stubs now emit a real shell + manifest. All five original assertions are untouched.
- **Docker-context slimming deferred to W-26 (#4442).** `scripts/create-service-docker-context.mjs` still ships `packages/mobile` and the export script into the web context. Out of scope here — but its export-output exclusion got a **fix, a comment correction and a guard test**, because removing the builder rebuild inverted that exclusion's risk profile. The old comment described the `BOARDSESH_WEB=0` rollback and claimed *"The `Dockerfile.web` builder rebuilds it from source"*; both are now false, and with no rebuild left to overwrite it, that exclusion is the **only** thing stopping a developer's local export riding into the image and being served as real files by the runner's `public/` COPY. It covered `packages/web/public/app` but not `packages/web/public/app-standalone` — fixed in the review round below. Same door #3795 must not reopen. `vp run docker-context:web` still exits 0.

### Accepted gap

Under the **dev Metro proxy**, the shell and manifest are served by Metro straight from `packages/mobile/public/` and never pass through the patch script, so a dev install prompt keeps `start_url: "/"`. Dev-only; the `/app/manifest.json` href is already correct there via the `next.config.mjs` rewrite.

---

### Automated (all run and watched to completion)

| Command | Result |
|---|---|
| `vp check --fix` | **33 errors / 323 warnings**, all pre-existing (same counts on clean `main`); **zero** diagnostics in any file this PR touches — grepped the full report against `git diff --name-only`. Formatting clean. |
| `vp run typecheck` | **PASS** (exit 0, incl. `next build` + `tsc --noEmit`) |
| `vp run typecheck:mobile` | **PASS** |
| `vp test run --reporter=agent --project scripts` | **PASS** — 60 files, 862 tests |
| `vp test run --reporter=agent --project deploy-app-subdomain` | **PASS** — 3 files, 13 tests |
| `vp run test:service-deploy-inputs` | **PASS** — 33 tests (the `node --test` suites behind `service-deploy-inputs.yml`, which this PR's `production-deploy-changes.mjs` / `create-service-docker-context.mjs` edits trigger) |
| `vp run test:mobile` | **PASS** — 680 files, 6517 tests |
| `vp test run --reporter=agent --project web` | **PASS** — 303 files, 3397 tests. (An earlier round hit 5 `waitFor` timeouts in gym/auth component tests; isolating them failed a *different* one each time on this branch and on clean `main`. They did not recur in the final run.) |
| `vp run check:i18n` / `check:i18n:orphans` | **PASS**, no-op as expected — this PR adds zero user-facing strings, so four-locale parity is trivially preserved |
| `vp run check:mobile-bundle` | **PASS** |
| `vp run check:mobile-web-bundle` (the CI gate) | **PASS**, and the log shows the new step ran: `[patch-expo-web-pwa-manifest] index.html manifest href → "/app/manifest.json"; manifest.json start_url/scope → "/app/"` |

### The issue's manual gates

**1. Manual `--subdomain` export** — `bash scripts/build-expo-web-export.sh --subdomain`, exit 0:

```
<link rel="manifest" href="/manifest.json" />
"start_url": "/",  "scope": "/",
grep -c '%WEB_TITLE%\|%LANG_ISO_CODE%' index.html  → 0
<title>Boardsesh</title>   <html lang="en"
```

**2. Dev-loop regression check — the highest-value check in this PR.** Both dev shapes were booted against a real `next dev` and curled.

*Static shape* (`BOARDSESH_WEB=1`, no proxy origin, baked `public/app`) — the `dev:mobile:web-static` loop:

| path | result |
|---|---|
| `/app` | 200 `text/html` (SPA shell) |
| `/app/climbs` | 200 (SPA fallback) |
| `/app/manifest.json` | **200 `application/json`** with `start_url`/`scope` `/app/` |
| `/app/wasm/board_renderer_wasm_bg.wasm` | 200 `application/wasm` |
| `/app/_expo/static/js/web/does-not-exist.js` | **404** — the negative lookahead still works |
| headers on `/app` | `x-robots-tag: noindex, follow`, `x-frame-options: SAMEORIGIN`, `x-content-type-options: nosniff` |

*Metro-proxy shape* (`BOARDSESH_EXPO_WEB_ORIGIN` pointed at a stand-in Metro) — every forward still lands, and the baked `public/app` export does **not** shadow it (`beforeFiles` holds):

```
/app                                   → FAKE-METRO /app
/app/climbs                            → FAKE-METRO /app/climbs
/app/manifest.json                     → FAKE-METRO /manifest.json
/app/wasm/board_renderer_wasm_bg.wasm  → FAKE-METRO /wasm/board_renderer_wasm_bg.wasm
/packages/mobile/index.js              → FAKE-METRO /packages/mobile/index.js
/assets, /assets/foo.png               → FAKE-METRO /assets, /assets/foo.png
```

**A/B on the deleted header rule.** The proxied `/app` response carries no `X-Robots-Tag` — so I re-ran the identical probe with the **pre-change `next.config.mjs`** restored, and it carries none either. The `/app/:path*` rule was **already inert on the dev proxy**: an external rewrite forwards the upstream response verbatim, exactly as the `middleware.ts` comment says. Deleting it is a no-op on that surface. On the real loop `packages/mobile/expo-web-response-headers.cjs` stamps `noindex, follow` + XFO/nosniff/Referrer-Policy itself, and `expo-web-header-parity.test.ts` still pins middleware ↔ Metro against those constants.

**3. Post-merge, on the live subdomain (`deploy-app-web`'s new smoke does this automatically):** served `index.html` links `/manifest.json`, which returns 200 `application/json`. It is genuinely falsifiable — it **fails against production today** and passes only after this export fix ships. Manual install-prompt + clean-console check in Chrome to follow the deploy.

**4. Docker oracle:** nothing in CI builds `Dockerfile.web` (`branch-deploy.yml`'s `pull_request` trigger is commented out; `railway.toml` carries only a backend `[deploy]` block), so `scripts/__tests__/dockerfile-web-no-expo-export.test.ts` is the **only** guard. All three of its assertions are mutation-probed below.

### Mutation probes (12, every one restored)

| # | Mutation | Went red |
|---|---|---|
| 1 | drop the `NODE_ENV !== 'development'` guard | `bakes no /app rewrite into a production build` |
| 2 | delete the dev SPA-fallback branch | `keeps the SPA fallback in dev…` + `keeps content-hashed and WASM namespaces out…` |
| 3 | re-add an `/app/:path*` header rule | `headers() owns no /app rule after the retirement` |
| 4 | patcher returns before writing | **7 of 8** patcher tests (the 8th is the idempotence case, which a total no-op trivially satisfies) |
| 5 | delete the `node …patch-expo-web-pwa-manifest.mjs` line from the export script | all 3 `build-expo-web-export.test.ts` cases |
| 6a | restore the `RUN … build-expo-web-export.sh` block in `Dockerfile.web` | `runs no Expo web export` |
| 6b | delete the runner-stage `ARG`/`ENV BOARDSESH_WEB` | `keeps the runner-stage BOARDSESH_WEB flag` |
| 6c | delete the runner's `public/` COPY | `still ships the built public/ directory to the runner` |
| 7 | hand-"fix" the shell href to `/manifest.json` (the natural but wrong fix) | `keeps exactly one manifest link, at the dev/Metro-proxy href` |
| 8a | add `{ type: 'dir', filePath: 'public' }` to `extraSources` | `is not a fingerprint input…` |
| 8b | add a `src/` file referencing `public/index.html` | `is not referenced by any native source file` |
| 8c | put `EXPO_PUBLIC_FOLDER` in `app.config.ts` | `does not relocate the public folder…` |
| 9 | empty the docker-context `ignoredRepoRelativePaths` set | `keeps every local export output out of the web build context` (renamed in the review round) |

`git status --porcelain` is clean; every probe was reverted with `\cp -f` from a pre-probe copy and re-verified.

### Review round

Third commit takes the three `claude-review` notes: `readRequired` now throws directly instead of `fail()`-then-`return ''` (the tail was unreachable and read like an empty-string fallback), the manifest-link regex is handed out by a factory so nothing reaching for `.exec()`/`.test()` later inherits a stale `lastIndex`, and `check_manifest_link`'s curls take `--max-time 20` so an unreachable edge fails one attempt and lets the retry loop work instead of hanging the release. Probe #4 was re-run against the refactored patcher: still 7 of 8 red.

Its one remaining observation — the smoke's `grep` matches only a double-quoted `rel="manifest"` — is a non-issue: that attribute comes from the checked-in template, which is double-quoted and pinned by `web-shell-appearance.test.ts`.

**Independent confirmation of the native claim:** the OTA-compat bot resolves iOS `183bcf484292` and Android `a6f256fdaf25`, both **matching `main`** — no fingerprint movement, exactly as §"Why the native claim is a test" predicts.

### Adversarial review round (fourth commit)

Three lenses (delete-safety, native-OTA, manifest-oracle) went at the diff and returned 13 findings, 11 distinct. Their common shape: **the guards this PR adds are narrower than the properties they claim**, so each was green while the hole it names was open. All 11 are fixed.

**Two majors that would have bitten:**

- **The Docker-context exclusion missed `packages/web/public/app-standalone`.** That is the `--subdomain` export's default output — the exact command `production-deploy.yml` runs, and the natural way to verify *this PR* locally. It is gitignored, so a stale copy sits in a tree invisibly, and with the builder rebuild gone nothing overwrites it. Probed for real: seeded both directories, ran `node scripts/create-service-docker-context.mjs web --output <tmp>`, and `app-standalone/` (shell, manifest, `_expo/`) was copied into the context in full while `app/` was correctly dropped. Both are excluded now. More to the point, **the guard no longer pins a literal `Set`** — it parses `build-expo-web-export.sh`'s own `DEFAULT_OUTPUT_DIR` values and asserts every one is excluded, so a future output directory cannot slip past it the same way this one did. Re-ran the probe after the fix: neither directory reaches the context.
- **A patcher-only PR would have deployed nothing.** `scripts/lib/patch-expo-web-pwa-manifest.mjs` was absent from `isAppAffecting`, so `classifyChangedFiles(['scripts/lib/patch-expo-web-pwa-manifest.mjs'])` returned `app: false` — a later fix to the patcher would merge green, skip `deploy-app-web`, never run the new manifest smoke, and leave the author believing it shipped. It is app-affecting now, with a test that pins both inputs of the subdomain artifact together.

**And one CI gap that made three guards decorative:** `dockerfile-web-no-expo-export.test.ts` and `build-expo-web-export.test.ts` read their inputs with `readFileSync`, so Vitest's `--changed` module-graph selection never relates them to a diff that only touches `Dockerfile.web` or the `.sh` — and a Dockerfile-only PR matches no job gate at all. Probed: appending a comment to both files and running `vp test run --changed HEAD --project scripts` selects **nothing**; doing the same to the patcher (a real module import) selects its suite. So #3795 could have re-added the export step to a green CI. They now run unfiltered in the `deploy-config` job, mirroring the `listing-guards` pattern the repo already uses for exactly this hazard. The patcher also joins the `deployConfig` and `mobile` filters, so a patcher-only change runs the deploy suites *and* `check:mobile-web-bundle` — the only gate that drives a real Expo export rather than a fixture.

**The rest:**

- The patcher wrote `index.html` before reading `manifest.json`, so any manifest-side failure left a half-patched export on disk — a shell pointing at a manifest path with no file behind it, i.e. the symptom this script exists to remove, reproduced locally with no warning. Both files are validated before either is written; the two failure tests now assert the shell is byte-identical after the throw.
- Nothing pinned `%WEB_TITLE%` / `%LANG_ISO_CODE%` in the real template, so hardcoding the shell's title and lang — a natural, harmless-looking edit — would have left the clobber detector matching nothing, with no test going red. `web-shell-appearance.test.ts` now **imports `UNRENDERED_TEMPLATE_TOKENS` from the patcher** and asserts every token is present, so detector and input cannot drift apart.
- The fingerprint guard compared `extraSources` `filePath`s as literal strings, so `'./public'` or `'../mobile/public'` would make `packages/mobile/public` a real native fingerprint input while the test stayed green. Paths are resolved against the package root before comparing; all three spellings now red.
- `check_manifest_link` checked the href and content-type but never the body. `start_url` can sit outside `scope` and still serve as valid JSON at the right URL — parses clean, never offers the install prompt. It now reads the manifest and asserts `start_url`/`scope`.
- Comment accuracy, three places: `create-service-docker-context.mjs` still said the builder stage runs the export and that the Dockerfile invokes the export script (both false since the first commit); `build-expo-web-export.sh` still called the default baseUrl's purpose "the legacy prod-static path", contradicting its own rewritten header 30 lines above. Also, the copied export script would have died at the `node …patch-expo-web-pwa-manifest.mjs` line because the patcher was not in `extraSourceFiles` — added, with a test that keeps the pair together for as long as the script is copied at all.
- The shipped shell comment went from seven lines to two. It ships verbatim to every `app.boardsesh.com` visitor (confirmed against the live site, which carries the old one), and the rationale belongs in `docs/expo-web-deployment.md`, which now carries it.

**One finding taken the other way round.** The native-safety test's header comment cited `assetBundlePatterns: ['assets/**/*']` as one of two legs of the claim, and nothing pinned it. The lens offered two fixes and flagged the tension itself: `assetBundlePatterns` filters assets Metro has **already resolved from the JS module graph**, and `public/` is not in that graph, so widening it to `['**/*']` would not embed `public/index.html` in a binary either. Pinning it would red a legitimate unrelated change for a reason that isn't true. So the sentence is out of the comment instead, with that reasoning written down — the test's stated scope now matches what it actually pins.

Two findings were duplicates across lenses (the `app-standalone` gap and the stale `services.web` comments), which is a decent signal they were the real ones.

### Review-round mutation probes (7, every one restored from a pre-probe copy)

| # | Mutation | Went red |
|---|---|---|
| 10 | drop `packages/web/public/app-standalone` from `ignoredRepoRelativePaths` | `keeps every local export output out of the web build context` |
| 11 | drop the patcher from the web service's `extraSourceFiles` | `copies the manifest patcher alongside the export script it shells out to` |
| 12 | drop the patcher from `isAppAffecting` | `treats every input of the app.boardsesh.com export as app-affecting` (33 → 32 pass) |
| 13 | hardcode `%WEB_TITLE%` / `%LANG_ISO_CODE%` in `packages/mobile/public/index.html` | `keeps the template tokens the export patcher watches for` |
| 14 | `{ filePath: './public' }` in `extraSources` | `is not a fingerprint input…` (green before this fix) |
| 15 | `{ filePath: '../mobile/public' }` in `extraSources` | `is not a fingerprint input…` (green before this fix) |
| 16 | move the shell write back before the manifest read | both `fails when manifest.json is missing…` and `fails on invalid manifest JSON…` |

Plus one behavioural re-probe, not a unit test: seeded `packages/web/public/{app,app-standalone}` with a fake export and ran the real context generator — the output's `source/packages/web/public/` listing is `brand icons images uploads wasm index.html openapi.json`, neither export directory present, and `source/scripts/lib/` carries both the patcher and the tailscale helper. Working tree cleaned; `git status --porcelain` empty.

The new `deploy-config` step was run exactly as CI will run it — `vp test run --project scripts scripts/__tests__/dockerfile-web-no-expo-export.test.ts scripts/__tests__/build-expo-web-export.test.ts` → 8 tests, 1.0s. Both workflow files parse as YAML and the smoke step's script passes `bash -n`; the new body assertion was exercised directly against good and bad manifests (`{"start_url":"/","scope":"/"}` → exit 0; `start_url:"/app/"` and `scope:"/app/"` → exit 1 each).

One knock-on, worth naming so it doesn't read as scope creep: adding the patcher to the web context's `extraSourceFiles` reds `check-service-deploy-inputs.test.mjs`, whose fixture repo carries a stub for every declared extra source file. One stub line added; all 33 of those tests pass.

### `claude-review`'s notes on the review round

Its verdict was **✅ Ready to merge**. Of the three observations:

- **Taken.** A timed-out or empty body fetch in `check_manifest_link` piped into `node` and surfaced as a JSON parse error rather than a fetch failure. The retry loop handled it either way, but only one of those messages tells you what went wrong. The fetch is now checked on its own line; exercised against a dead port → `manifest body fetch failed`, exit 1.
- **Not taken — the cross-boundary import is the point.** `web-shell-appearance.test.ts` importing `UNRENDERED_TEMPLATE_TOKENS` from `scripts/lib/` crosses `packages/mobile` → repo root, which is unusual. The alternative is restating the two tokens in the test, which is exactly the drift the pin exists to prevent: the whole finding was that the detector and its input can silently diverge. The direction is also the useful one — if `scripts/lib/` moves, the *template's* contract is what broke, and the mobile suite is where that should surface. Either placement crosses the boundary anyway (the patcher lives in `scripts/`, the template in `packages/mobile/`); this way there is a single source of truth.
- **Not taken — acknowledged as inherent by the reviewer itself.** The native-reference scan uses `string.includes()`, so `path.join('public', 'index.html')` would evade it. True of any static scan, and it is the weaker of the two legs; the load-bearing one is fingerprint-neutrality, independently confirmed by the OTA-compat bot resolving both platforms as matching `main`. Not worth an AST pass for a directory no native code has ever read.

### Deliberately not touched

`/api/internal/board-render` and `next.config.mjs`'s `outputFileTracingIncludes` (the ESP32 pins `www.boardsesh.com` as its render host), `deploy/app-subdomain/{_headers,_redirects}`, `packages/web/app/robots.ts` (there was never an `/app` `Disallow`), `packages/web/middleware.ts`, `app/layout.tsx`, `packages/mobile/{metro.config.js,app.config.ts,expo-web-response-headers.cjs,public/manifest.json,public/wasm/**}`, `crawler-classic-invariant.test.ts` (single-owner, W-09 — it runs its board-surface invariants under both `BOARDSESH_WEB` states and delegates the `/app` carve-out to `middleware.test.ts`, which is unchanged; re-ran it green anyway).

## Release Notes

Installing the browser app now works. Open `app.boardsesh.com` in Chrome or Safari and you'll get the "Add to Home Screen" prompt, launching straight into the app instead of a browser tab.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -110,6 +110,12 @@ jobs:
               - 'scripts/mobile-platform-imports-check.sh'
               - 'scripts/mobile-web-bundle-check.sh'
               - 'scripts/sync-mobile-board-renderer-wasm.sh'
+              # check:mobile-web-bundle drives the real Expo export, so these two
+              # only ever run against a genuinely rendered shell here. A
+              # unit-test fixture cannot catch a patcher that breaks on Expo's
+              # actual output. W-24 / #4438.
+              - 'scripts/build-expo-web-export.sh'
+              - 'scripts/lib/patch-expo-web-pwa-manifest.mjs'
               # The live fingerprint gates run in the mobile-bundle job; a PR
               # editing only their logic must still trigger that job.
               - 'scripts/mobile-fingerprint-inputs-check.ts'
@@ -151,6 +157,7 @@ jobs:
             deployConfig:
               - 'deploy/app-subdomain/**'
               - 'scripts/build-expo-web-export.sh'
+              - 'scripts/lib/patch-expo-web-pwa-manifest.mjs'
               - 'Dockerfile.web'
               - '.github/workflows/production-deploy.yml'
             # Everything scripts/store-metadata-limits.test.ts reads via fs at
@@ -498,6 +505,18 @@ jobs:
         run: bun install --frozen-lockfile
       - name: Validate deploy/app-subdomain config (_headers, _redirects, no-CSP rule)
         run: vp test run --project deploy-app-subdomain
+      # Same fs-reading blind spot, one project over: both suites read
+      # Dockerfile.web / build-expo-web-export.sh / create-service-docker-context.mjs
+      # with readFileSync, so test-default's `--changed` never relates them to a
+      # diff that only touches those files — and a Dockerfile- or .sh-only PR
+      # matches no other job's gate at all. Nothing in CI builds Dockerfile.web
+      # (branch-deploy.yml is workflow_dispatch-only), so these pins are the only
+      # thing standing between #3795 and a restored /app export step. W-24 / #4438.
+      - name: Validate the retired /app export stays retired
+        run: >-
+          vp test run --project scripts
+          scripts/__tests__/dockerfile-web-no-expo-export.test.ts
+          scripts/__tests__/build-expo-web-export.test.ts
 
   listing-guards:
     needs: [changes]
--- a/.github/workflows/production-deploy.yml
+++ b/.github/workflows/production-deploy.yml
@@ -712,12 +712,15 @@ jobs:
             # HTTP 200 the whole way. scripts/lib/patch-expo-web-pwa-manifest.mjs
             # rewrites the href from the export's baseUrl; this is the
             # end-to-end proof it landed on the real deploy.
+            # --max-time so an unreachable edge fails this attempt and lets the
+            # retry loop above do its job, instead of hanging the release. The
+            # sibling checks predate this and are left alone here on purpose.
             local href
-            href=$(curl -sS "$base/" \
+            href=$(curl -sS --max-time 20 "$base/" \
               | grep -oE '<link[^>]*rel="manifest"[^>]*>' \
               | grep -oE 'href="[^"]+"' | head -n1 | cut -d'"' -f2)
             [ "$href" = "/manifest.json" ] || { echo "  manifest href is '$href', expected /manifest.json"; return 1; }
-            local out; out=$(curl -sS -o /dev/null -w '%{http_code} %{content_type}' "$base$href")
+            local out; out=$(curl -sS --max-time 20 -o /dev/null -w '%{http_code} %{content_type}' "$base$href")
             echo "$out" | grep -q '^200 ' && echo "$out" | grep -qi 'application/json'
           }
 
--- a/.github/workflows/production-deploy.yml
+++ b/.github/workflows/production-deploy.yml
@@ -726,7 +726,13 @@ jobs:
             # Read the body too. The href and content-type can both be right
             # while start_url sits outside scope — a manifest that parses
             # cleanly and still never offers the install prompt.
-            local body; body=$(curl -sS --max-time 20 "$base$href")
+            # Fetch and check the body separately, so a timed-out or empty
+            # response reports as a fetch failure rather than a JSON parse
+            # error from node — the retry loop handles both, but only one of
+            # them tells you what actually went wrong.
+            local body
+            body=$(curl -sS --max-time 20 "$base$href") || { echo "  manifest body fetch failed"; return 1; }
+            [ -n "$body" ] || { echo "  manifest body is empty"; return 1; }
             echo "$body" | node -e '
               const manifest = JSON.parse(require("node:fs").readFileSync(0, "utf8"));
               if (manifest.start_url !== "/" || manifest.scope !== "/") {
--- a/.github/workflows/production-deploy.yml
+++ b/.github/workflows/production-deploy.yml
@@ -704,12 +704,30 @@ jobs:
             echo "$cc" | grep -qi 'immutable'
           }
 
+          check_manifest_link() {
+            # The bug this guards (W-24, #4438): index.html asked for
+            # /app/manifest.json, _redirects' `/* /index.html 200` answered it
+            # with the SPA shell, and the browser silently got HTML where a
+            # manifest belongs — no install prompt, console error every load,
+            # HTTP 200 the whole way. scripts/lib/patch-expo-web-pwa-manifest.mjs
+            # rewrites the href from the export's baseUrl; this is the
+            # end-to-end proof it landed on the real deploy.
+            local href
+            href=$(curl -sS "$base/" \
+              | grep -oE '<link[^>]*rel="manifest"[^>]*>' \
+              | grep -oE 'href="[^"]+"' | head -n1 | cut -d'"' -f2)
+            [ "$href" = "/manifest.json" ] || { echo "  manifest href is '$href', expected /manifest.json"; return 1; }
+            local out; out=$(curl -sS -o /dev/null -w '%{http_code} %{content_type}' "$base$href")
+            echo "$out" | grep -q '^200 ' && echo "$out" | grep -qi 'application/json'
+          }
+
           retry "root / → 200 text/html"                 check_root
           retry "deep route /climbs → 200 (SPA shell)"    check_deep_route
           retry "X-Robots-Tag: noindex on /"              check_noindex
           retry "wasm binary → 200 wasm/octet-stream"     check_wasm
           retry "index.html is NOT immutable-cached"      check_index_not_immutable
           retry "hashed _expo asset is immutable-cached"  check_hashed_asset_immutable
+          retry "PWA manifest link → 200 application/json"  check_manifest_link
 
       - name: Install Playwright chromium
         run: cd packages/web && bunx playwright install --with-deps chromium
--- a/.github/workflows/production-deploy.yml
+++ b/.github/workflows/production-deploy.yml
@@ -721,7 +721,19 @@ jobs:
               | grep -oE 'href="[^"]+"' | head -n1 | cut -d'"' -f2)
             [ "$href" = "/manifest.json" ] || { echo "  manifest href is '$href', expected /manifest.json"; return 1; }
             local out; out=$(curl -sS --max-time 20 -o /dev/null -w '%{http_code} %{content_type}' "$base$href")
-            echo "$out" | grep -q '^200 ' && echo "$out" | grep -qi 'application/json'
+            echo "$out" | grep -q '^200 ' || { echo "  manifest fetch: $out"; return 1; }
+            echo "$out" | grep -qi 'application/json' || { echo "  manifest content-type: $out"; return 1; }
+            # Read the body too. The href and content-type can both be right
+            # while start_url sits outside scope — a manifest that parses
+            # cleanly and still never offers the install prompt.
+            local body; body=$(curl -sS --max-time 20 "$base$href")
+            echo "$body" | node -e '
+              const manifest = JSON.parse(require("node:fs").readFileSync(0, "utf8"));
+              if (manifest.start_url !== "/" || manifest.scope !== "/") {
+                console.error(`  manifest start_url is "${manifest.start_url}" and scope is "${manifest.scope}"; expected "/" for both`);
+                process.exit(1);
+              }
+            '
           }
 
           retry "root / → 200 text/html"                 check_root
@@ -730,7 +742,7 @@ jobs:
           retry "wasm binary → 200 wasm/octet-stream"     check_wasm
           retry "index.html is NOT immutable-cached"      check_index_not_immutable
           retry "hashed _expo asset is immutable-cached"  check_hashed_asset_immutable
-          retry "PWA manifest link → 200 application/json"  check_manifest_link
+          retry "PWA manifest → JSON with start_url/scope /"  check_manifest_link
 
       - name: Install Playwright chromium
         run: cd packages/web && bunx playwright install --with-deps chromium
--- a/Dockerfile.web
+++ b/Dockerfile.web
@@ -27,9 +27,6 @@ FROM node:22-alpine AS builder
 
 WORKDIR /app
 
-# bash runs scripts/build-expo-web-export.sh (repo scripts are bash, not sh)
-RUN apk add --no-cache bash
-
 # Copy bun binary from deps stage (avoid re-downloading)
 COPY --from=deps /root/.bun/bin/bun /usr/local/bin/bun
 RUN ln -s /usr/local/bin/bun /usr/local/bin/bunx
@@ -51,38 +48,25 @@ COPY source/scripts ./scripts
 # Bun's isolated linker resolves workspace deps through per-package symlink
 # farms (packages/*/node_modules -> ../../node_modules/.bun). The deps stage
 # created them next to the manifests, but only the root node_modules (with the
-# .bun store) crosses the stage boundary, so relink before the Expo export
-# resolves modules from the source tree — `bunx expo` must find the workspace's
-# pinned expo, never fall back to fetching a floating version.
+# .bun store) crosses the stage boundary, so relink before the Next build
+# resolves modules from the source tree.
 RUN bun install --frozen-lockfile
 
 # Accept build arguments for environment variables baked into the Next.js build
 ARG NEXT_PUBLIC_WS_URL
 ARG BASE_URL
-# BOARDSESH_WEB=1 turns on the Expo web export below AND the /app serving
-# branch that next.config.mjs bakes into the standalone build. Rollback lever:
-# build with --build-arg BOARDSESH_WEB=0 to ship an image without the Expo web
-# app — /app then 404s while the rest of the site is untouched.
-ARG BOARDSESH_WEB=1
 
 ENV NEXT_PUBLIC_WS_URL=$NEXT_PUBLIC_WS_URL
 ENV BASE_URL=$BASE_URL
-ENV BOARDSESH_WEB=$BOARDSESH_WEB
 
-# Static Expo web export served by Next at /app. The nested web-runtime install
-# supplies the web-only renderer (react-native-web, react-dom, the gorhom sheet)
-# outside the native fingerprint graph — the export script runs it when missing.
-# The export lands in packages/web/public/app, which the runner stage picks up
-# through the existing public/ COPY and Next's standalone server then serves
-# statically. EXPO_PUBLIC_WEB_URL feeds the Expo Router head origin; the app's
-# runtime URLs (backend/WS) fall back to production defaults in src/lib/env.ts.
-# Only export EXPO_PUBLIC_WEB_URL when BASE_URL is non-empty: Expo inlines
-# EXPO_PUBLIC_* values verbatim, and a set-but-empty value would override the
-# production fallback in src/lib/env.ts with ''.
-RUN if [ "$BOARDSESH_WEB" = "1" ]; then \
-      if [ -n "$BASE_URL" ]; then export EXPO_PUBLIC_WEB_URL="$BASE_URL"; fi; \
-      bash scripts/build-expo-web-export.sh packages/web/public/app; \
-    fi
+# No Expo web export step here (W-24, #4438). The browser app ships only at
+# app.boardsesh.com, exported and published by production-deploy.yml's
+# deploy-app-web job. This image used to bake a second copy of it under
+# public/ for Next to serve at /app; that path is retired, and next.config.mjs
+# returns no /app rewrites outside NODE_ENV=development, so a production build
+# could not serve one even if an artifact reappeared. Guarded by
+# scripts/__tests__/dockerfile-web-no-expo-export.test.ts, which pins this file
+# by text because nothing in CI builds it.
 
 # Build the Next.js application (standalone output)
 RUN bun run --filter=@boardsesh/web build
@@ -94,13 +78,15 @@ WORKDIR /app
 
 ENV NODE_ENV=production
 
-# BOARDSESH_WEB is read at REQUEST time by the standalone server, not just at
-# build: app/layout.tsx gates the Expo auth session bridge on it
-# (enableExpoAuthBridge) and middleware.ts gates the /app dev proxy. The builder
-# ARG/ENV above does not cross the stage boundary, so redeclare it here or the
-# runtime flag is undefined and the bridge stays off even though /app is served.
-# The default mirrors the builder so a single --build-arg BOARDSESH_WEB=0
-# rollback disables both the export AND the runtime bridge consistently.
+# BOARDSESH_WEB is read at REQUEST time by the standalone server. Since W-24
+# (#4438) retired the /app static path it has exactly two remaining jobs, both
+# request-time: app/layout.tsx gates the cross-subdomain Expo auth session
+# bridge on it (enableExpoAuthBridge — the www half of what app.boardsesh.com
+# signs in through), and middleware.ts gates the /app carve-out. The builder
+# stage no longer declares it at all; there is no /app serving left to roll back
+# with --build-arg BOARDSESH_WEB=0. Setting it to 0 here now only turns the auth
+# bridge off, which would break sign-in on the subdomain — do not use it as a
+# rollback lever.
 ARG BOARDSESH_WEB=1
 ENV BOARDSESH_WEB=$BOARDSESH_WEB
 
--- a/deploy/app-subdomain/README.md
+++ b/deploy/app-subdomain/README.md
@@ -27,7 +27,9 @@ else falls through to the shell.
 ## `_headers` — security + caching
 
 - `X-Robots-Tag: noindex` — this is an authenticated utility surface, kept out of
-  search indexes (mirrors the `/app` surface's `noindex`, see docs/expo-web.md).
+  search indexes. Since W-24 (#4438) retired the `/app` static path this file is
+  the only production source of that header for the browser app; the dev `/app`
+  proxy gets its own from `packages/web/middleware.ts` (see docs/expo-web.md).
 - `X-Content-Type-Options: nosniff`, `Referrer-Policy: strict-origin-when-cross-origin`
   — standard hardening.
 - `/_expo/*` and `/assets/*` → `Cache-Control: public, max-age=31536000, immutable`.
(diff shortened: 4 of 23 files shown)


marcodejongh:
## Fable review — 🔒 point of no return + `packages/mobile` native-OTA review

Verdict: **APPROVE.** I would merge this. Judged in the order that matters, with independent verification, not a re-read of the PR body.

### 1. Native safety (standing rule 2)

The "native never reads this byte" half is **proven, and the proof is durable**. `web-public-assets-are-web-only.test.ts` pins all three legs: `public/` absent from `extraSources` (path-resolved, not string-matched), no native source file referencing `public/index.html`/`public/manifest.json`, and no `EXPO_PUBLIC_FOLDER` relocation. I re-ran the native-OTA lens's `'./public'` spelling escape against the *fixed* test myself — it goes red now (it was green before the fix, so the lens finding was real and the fix is real). The OTA-compat bot resolving both fingerprints identical to `main` covers delivery; the test covers safety. The acknowledged weakness — the reference scan is `includes()`-based — is the weaker leg by design and correctly not over-engineered.

### 2. Irreversibility

The `NODE_ENV !== 'development'` gate is airtight for the property #3795 needs: `next build` forces `NODE_ENV=production` on both production paths (Vercel and `Dockerfile.web`), so no production build can bake an `/app` rewrite regardless of `BOARDSESH_WEB`. I mutation-probed the gate myself — deleting it reds `bakes no /app rewrite into a production build`. The Dockerfile side is double-locked (export step deleted + text pin), and the pin's CI blind spot (fs-read tests invisible to `--changed`) was found by the native-OTA lens and closed with the unconditional `deploy-config` step — I confirmed that job ran and passed on this PR. What the merge irreversibly does is deploy the manifest fix to app.boardsesh.com, which is the point; `APP_WEB_DEPLOY_HOLD` remains the freeze lever.

### 3. Dev loops

Both survive. The Metro proxy block in `rewrites()` is byte-identical (checked against the diff, and the four proxy tests are unchanged); `dev:mobile:web-static` lands in the kept dev-gated branch, pinned by its own named test, and the PR's manual QA table shows real curls through both shapes including the negative-lookahead 404 and `/app/manifest.json` returning JSON. The A/B proof that the deleted `/app/:path*` header rule was already inert on the proxied surface is the right way to have settled that question.

### 4. The manifest fix

The patcher cannot silently no-op: zero manifest links throws, a link with no `href` fails the post-write re-read (`hrefOf` → null ≠ expected), unsubstituted template tokens throw, and both files are validated before either is written (the delete-safety lens's half-patched-export finding, fixed with tests asserting the shell is byte-identical after a throw). The template tokens the clobber detector depends on are now pinned in `web-shell-appearance.test.ts` by *importing* `UNRENDERED_TEMPLATE_TOKENS` from the patcher — the one arrangement that can't drift. The production smoke is genuinely falsifiable — it fails against live production today (lens-verified by curl: the shipped shell still links `/app/manifest.json`) and now checks the body's `start_url`/`scope`, not just href + content-type. I re-ran all four script suites (21 tests), both mobile guard suites, and both web config suites: green.

### 5. What the lenses and fixer missed

Nothing blocking. The three findings I'd have filed myself — the `app-standalone` context leak, the patcher missing from `isAppAffecting`, and the fs-read tests invisible to `--changed` — were all found and all fixed with guards that track behaviour rather than literals (the Dockerfile test now parses the export script's own `DEFAULT_OUTPUT_DIR` values instead of pinning a `Set`; that is exactly the shape a durable guard should have). Residuals, all non-blocking:

- The mobile suite importing a repo-root `.mjs` is unusual but deliberate and typechecks (`vp run typecheck:mobile` clean); the "not taken" rationale is correct — restating the tokens would recreate the drift the pin exists to prevent.
- The dev-Metro install prompt keeping `start_url: "/"` is an accepted, documented, dev-only gap.
- The pre-existing question of whether `BOARDSESH_WEB` is set on the Vercel project (auth-bridge half) is correctly filed-not-fixed — keep it out of this PR.

### Conditions

None on the code. Two on the follow-through:

1. After merge, **watch `deploy-app-web` and its new `PWA manifest → JSON with start_url/scope /` smoke to completion** — it is the end-to-end oracle and it must flip from would-fail-today to pass on this deploy.
2. Do the issue's Chrome install-prompt + clean-console check against the live subdomain and note it on #4438 — that closes the Verifies list.

CI is fully green, no unresolved review threads, no conflicts: this meets the ready-to-merge signal. Take it out of draft.

*Independent verification performed: re-ran the 4 script suites, 2 mobile guard suites, 2 web config suites (all green); mutation-probed the `NODE_ENV` gate and the `'./public'` fingerprint escape (both red, both restored, `git status --porcelain` clean).*

marcodejongh:
Fable's non-blocking item 3 is done: the `BOARDSESH_WEB`-unset-on-Vercel auth-bridge question is filed as **#4487**, with the two ways to check it (dashboard / `vercel env ls production`), a behavioural probe that needs no dashboard access, and the warning not to flip the flag without accounting for the dev-proxy middleware carve-out it also drives after this PR.

Marked ready for review — CI is green (25 passing, 5 skipping, no failures), Fable approved with no blocking items, and there are no conflicts.

Two post-merge obligations carried over from the review, recorded here so they are not lost:

1. **Watch `deploy-app-web` and its new `PWA manifest → JSON with start_url/scope /` smoke to completion.** That check is the end-to-end oracle for this PR and it must flip from fail-today to pass on this deploy. If it does not, the manifest patch did not land on the subdomain.
2. **The manual Chrome install-prompt check** against live `app.boardsesh.com` — Application → Manifest parses, the install affordance appears, console clean of manifest errors — recorded on #4438 to close its Verifies list.

Issue #4463, opened 2026-08-15
E2E: a browser spec poisons the web server for subsequent raw HTTP requests


marcodejongh:
Part of #4358 (found during W-17/#4448's e2e verification; supersedes the intra-shard-concurrency framing in #4461's evidence section — the pool note there still stands on its own merits).

**Symptom:** `embed-headers.spec.ts`'s raw `APIRequestContext.get` requests (`/embed/board/<missing>`, `/embed/gym/<missing>/leaderboard`, `/kiosk/whatever`) hang 60s+ (30s timeout × retry) in CI **when certain browser-driven specs run against the same `next start` server first**. First-in-shard (alphabetical pin, run 31860014178) they pass every time and the full suite is green.

**Eliminated by experiment** (all on the identical build, W-17 branch):
- Not worker concurrency: reproduces with `workers: 1` (sequential).
- Not the backend: local repro is <1s with the backend up, connection-refused, and blackholed (accepts, never replies). The embed pages read Postgres directly.
- Not pool exhaustion by load: 30 sequential front-door SSR loads locally, embed still 65ms.
- Not the not-found path: `/kiosk/whatever` and bare 404s answer <1s locally.
- Not the redirect table, headers, UA, or accept-encoding: exact header parity replay is fast.
- Not pre-existing: baseline (main + Playwright image fix) is green with the old shard layout.

**What's left:** a browser-driven spec in the old shard-7 grouping (the rewritten `bottom-tab-bar.spec.ts` is the prime suspect — its authenticated tests sign in via next-auth) leaves the *server process* in a state where raw keep-alive HTTP/1.1 requests from a fresh `APIRequestContext` stall, while `page.goto` from the browser continues to work. Candidate mechanisms worth checking: something holding the Node server's connection/agent state (undici keep-alive pinning), a session/rate-limit path keyed on the shared 127.0.0.1 client IP, or an SSR request-scoped resource that never closes on the poisoning page.

**Repro recipe:** revert the `aa-` pin commit on a branch, dispatch `e2e-tests.yml`, watch shard 7. Binary-search the poisoning test by deleting halves of `bottom-tab-bar.spec.ts`.

The `aa-embed-headers.spec.ts` alphabetical pin is load-bearing until this is fixed (file-top comment marks it). Sonnet-tier investigation with an Opus review; the fix must remove the pin.

marcodejongh:
[...]

Issue #4436, opened 2026-08-14
W-23: ship the tier-2 climb sitemaps and JSON-LD


marcodejongh:
Part of #4358. **Blocked by:** #4434 (W-22 — the sitemap index and shards). QA gate #4359 is **SIGNED and closed (2026-08-14)**; this item is gate-parallel.

## Context

The addressable universe is **771,159 climbs** (listed && !draft, read-only prod count, 2026-07-31): kilter 314,615 · moonboard 280,755 · tension 142,310 · grasshopper 21,802 · decoy 8,352 · soill 1,972 · touchstone 1,353. This PR submits **tier 2** only.

### Climb sitemaps

- Tier 2 = **128,655 URLs → 3 shards** at the 45k cap.
- Tier 2 is **listed, non-draft, `ascensionist_count >= 10`, one angle each** (highest ascents; ties → the climb's own angle).
- Other angles stay **self-canonical** and are reached via W-15's cross-links.
- **Tier 3 (771,159) is deferred** until Search Console shows healthy tier-2 coverage.

### JSON-LD, in value order

1. `BreadcrumbList`
2. `CreativeWork` + `aggregateRating` for the climb
3. `Organization` / `WebSite` on `/` — **no `SearchAction`**, because www has no site search after the teardown
4. `ProfilePage`
5. `ItemList`

Emitted with the escaped `dangerouslySetInnerHTML` pattern `gym/[gym_slug]/page.tsx:138` already uses.

`VideoObject` is **deferred** — beta links are third-party embeds and need a policy call. (The epic's "what we are not doing" list is explicit: not shipping `VideoObject` JSON-LD or tier-3 climb sitemaps in this wave.)

## Standing-rule callouts

- Sitemap entries use **real content timestamps, not `new Date()`**.
- **Not noindexing `/es`, `/fr` or `/de` climbing pages** — locked decision.
- Keep trademark wording compatible-not-affiliative (**MoonBoard**, **Kilter**, **Tension**); never imply endorsement of Aurora or Moon Climbing.
- Canonicalise filtered/sorted/paginated variants to the clean base URL; `/play/...` canonicalises to the equivalent `/view/...`.

## Deps

- W-22 — #4434

## Gates

- **GW** — `vp check` · `vp run typecheck` · `vp test run --reporter=agent --project web`

## Verifies

- [ ] The three tier-2 shards are each ≤45k URLs, and the counts reconcile against a fresh read-only prod count at branch time.
[...]

PR #4481, opened 2026-08-15
perf(web): bound front-door reads and fail loud when the database is slow

### LINKED ISSUE #4461: Front-door SSR holds the web PG pool under concurrency — size before the climb sitemaps land
Part of #4358 (found during W-17's e2e verification, #4448).

The climb/list front doors (W-15) SSR with **direct Postgres reads** (list page data, climb, stats, similar climbs). Under concurrency on a slow database, those page loads hold enough of the web server's connection pool that unrelated cheap requests (the embed pages' single-row lookups) queue behind them for 30–60s+.

Evidence (from #4448's e2e forensics): with 2 Playwright workers per CI shard, `embed-headers.spec.ts`'s raw requests hung >60s whenever they shared a shard with front-door page loads; single-worker CI (no intra-shard concurrency) resolves it. Locally against a warm dev DB, the same requests complete in ~140ms, backend up, down, or blackholed — the contention needs a slow DB + concurrent front-door SSR.

**Why this matters beyond CI:** production runs on Vercel with per-instance pools against Railway Postgres. A crawl burst over the newly-submitted front-door URLs (W-22/W-23 sitemaps will invite exactly that) is the same shape: many concurrent SSR loads competing for pool connections. Worth deciding deliberately:
- pool sizing / queue timeouts for `@/app/lib/db/db`'s pool on the web side,
- whether the front door's per-request query fan-out (list + stats + similar climbs) should batch or cache harder (`unstable_cache` coverage audit),
- a load test of a front-door URL at crawl-burst concurrency before W-23 submits 128k URLs.

Sonnet-tier investigation + Opus review; GW gates.

marcodejongh:
## Summary

Closes #4461. Part of #4358 (W-tier-4 tail). Lands **before or with** W-23 (#4436), which submits the climb sitemaps.

The front doors (`/…/view/{climb_uuid}` and `/…/list`, both trees) SSR with direct Postgres reads, and postgres.js has **no acquire timeout** — its internal queue is unbounded and untimed (`postgres/src/index.js:341`). A statement that cannot get a connection waits forever. This PR bounds that wait, gives the deployment knobs to shrink its connection footprint, and — the highest-value part — fixes what those pages emit when the read fails.

### First: the issue's stated mechanism is wrong, and the fix does not depend on it

The issue attributes the CI hang to "the embed pages' single-row lookups queueing behind front-door SSR for the web PG pool". That cannot be it: **`/embed/**` issues zero web-side Postgres statements**. `app/components/kiosk/embed/embed-fetchers.ts:47-56` fetches the backend over HTTP, and there is no `sql` / `dbz` / `getReadDb` import anywhere under `app/embed/` or `app/components/kiosk/embed/`.

The likelier mechanism is the event loop, not the pool: a `/list` SSR render fired up to **20 same-origin `/api/internal/board-render` warm requests**, each a CPU-bound WASM render in the same runtime. Unrelated raw requests queue behind those. That is why the fan-out cap below is in this PR.

`workers: 1` in `playwright.config.ts` is **unchanged** — flipping it back is #4463's call. The misattributing comment there is corrected so the next reader is not misled.

### Also refuted before implementation: the locale CDN scare

The plan's largest claimed amplifier was that the sticky-locale `Set-Cookie` in `middleware.ts` makes the 75% of the crawl surface that is locale-prefixed CDN-uncacheable. **Measured against production 2026-08-15: it does not.** A cold German list front door went MISS → HIT → HIT; a warm French climb view was HIT at `age: 899`; the Spanish list went MISS → HIT within seconds **with `set-cookie: boardsesh-locale=es` present on both**. Locale front doors are CDN-cached, the cookie is benign here (its value matches the URL's own locale), and `middleware.ts` is untouched by this PR.

### What changed

**1. Fail-loud status semantics — the SEO fix.** Today a DB brownout emits a **404** on an indexed climb URL and a **200 with an empty list** on an indexed list URL. At sitemap scale the first is an instruction to deindex and the second is worse: Google reads it as legitimate thin content.

- `fetchClimbFromDb` resolves `null` for a genuinely missing row instead of dereferencing an undefined one.
- `notFound()` now fires only for a genuinely missing climb (it was dead code — the undefined-row throw reached the catch first).
- The catch in both view pages rethrows everything. Next's error boundary turns that into a 500.
- `fetchFrontDoorListPage` rethrows instead of degrading to `{ climbs: [], hasMore: false }`.
- `resolveBoardBySlug` — the **first** read on every `/b/{slug}/…` request, and the one the QA round caught still swallowing failures — now returns `null` only for a clean 200 with no such board. A rejected fetch, a non-2xx, or a 200 carrying GraphQL `errors` throws.

| condition | before | after |
|---|---|---|
| climb row absent | 404 | 404 — unchanged, now for the right reason |
| board slug resolves cleanly to nothing | 404 | 404 — unchanged |
| read fails/deadlines, climb page | hung to the platform limit, then 404 | 500 at ~6 s per request |
| read fails/deadlines, list page | 200 with zero climbs | 500 at ~6 s |
| backend `boardBySlug` fails, `/b/…` | 404 | 500 |
| backend GraphQL wedged | climb page hung indefinitely | similar climbs / beta links render empty at 3 s |

**500 rather than a literal 503 + `Retry-After` is deliberate**, not an oversight: an App Router page cannot emit one without a middleware round trip, and the property that matters — Google retries a 5xx and keeps the URL — holds for both.

**The CDN does not hide the 500, and that is fine.** `stale-while-revalidate` only cushions a URL whose `age` has already passed `s-maxage`: climb views are covered for age 1 h–7 h, lists for 24 h–7 d, and Vercel supports [no `stale-if-error`](https://vercel.com/docs/caching/cdn-cache). A cold-MISS crawl of a long-tail sitemap URL sees the 500 directly. The asymmetry that matters is what happens next: 5xx is **not** a cacheable status on Vercel, so it is never pinned, while the 404 this replaces **is** — one blip used to stick a 404 in the CDN for a full `s-maxage` (24 h on a list front door).

**2. A read deadline** (`app/lib/db/read-deadline.ts`). Races the pending read against a timer and rejects with `DbReadTimeoutError` (`code: 'DB_READ_TIMEOUT'`). Default 6000 ms, `DB_READ_DEADLINE_MS`-overridable — deliberately below `DB_CONNECT_RETRY_BUDGET_MS` (10 s), because during a brownout the front door should shed rather than spend a second and third connect attempt holding a slot.

Wired at four call sites only: the alias lookup, the climb select, the all-angles stats select, and the shared climb search. **Not** inside `withConnectRetry` or `packages/db` — that would change behaviour for the backend, the sync runners and every script, and would miss the raw-`sql` path `getClimb` uses anyway.

**The budget is per request, not per statement** (`app/lib/db/request-read-budget.ts`). The climb page issues three reads in sequence, so three independent 6 s deadlines would have been an ~18 s request ceiling — the opposite of shedding load, and it would have broken the comparison against the 10 s connect-retry budget that justifies 6000 in the first place. One deadline timestamp lives in React's per-render `cache` scope and each read gets what the earlier ones left, floored at 500 ms.

**Cancellation covers three of the four.** On a timeout the helper calls `query.cancel()` the way `db-health.ts` does, so a timed-out statement does not fire later against a recovered pool. The list search is drizzle-issued and exposes no `.cancel()`, so there the deadline sheds the caller while the statement runs on holding its connection — stated in the code, in `docs/db-connectivity.md`, and here, because the first draft of this PR claimed cancellation unconditionally.

**3. Backend-call deadlines.** `createCachedGraphQLQuery` takes an optional `timeoutMs`; the two front-door GraphQL reads pass 3000. Both already degrade to an empty section, so a timeout never 500s the page.

**4. Per-deployment pool knobs.** `DB_POOL_MAX` (default 10, clamped to a floor of 2) and `DB_POOL_IDLE_TIMEOUT_S` (default 30, **no** floor — postgres.js reads a falsy `idle_timeout` as "never close an idle connection", so clamping 0 up to 1 would invert the operator's request into a teardown every second). **Defaults unchanged, so backend, sync and scripts behave identically.** They exist because peak server-side connections scale with instance count × connections held *idle*, not with per-instance `max`.

**5. `DB_STATEMENT_TIMEOUT_MS` — env-gated, ships off.** PgBouncer in transaction pooling rejects startup parameters outside `ignore_startup_parameters`, and `statement_timeout` is not among the defaults, so enabling it against a pooled URL would fail *every connection*. `docs/db-connectivity.md` documents both paths (env var for a direct URL, `ALTER ROLE … SET statement_timeout` for a pooled one). Neither is enabled here.

**6. Stop paying for the crawl path.**
- MoonBoard `/list` front doors were **never** cached (`board_name !== 'moonboard'` overrode the caller's `cacheable: true`), so every crawled page was a live search against the largest board in the catalogue. An explicit `cacheable: true` now wins, at a 15-minute TTL that keeps the import freshness the bypass was protecting. Only `fetchFrontDoorListPage` passes it, so the blast radius is exactly the front door, and `revalidateClimbSearchTags` still clears it via the board-level tag.
- Warm fan-out capped at 6 rows per list render (was 20).
- Each warm fetch carries `AbortSignal.timeout(5000)` — one that never settles keeps a serverless instance, and the pool slots it holds, alive past its response.

**7. Two free wins on the climb page.** `getClimb` / `getClimbStatsForAllAngles` are React-`cache()`d on primitives, so `generateMetadata` and the page body share one read instead of both missing a cold `unstable_cache` key concurrently — 4 statements down to 2 per cold render, in both trees. The `unstable_cache` executors are now stable module-level functions taking explicit primitive arguments (the pattern `search-climbs.ts` documents) rather than a fresh closure per request. The `unstable_cache` *instance* still has to be built per call because its `tags` carry the climb uuid; a uuid-keyed registry would grow without bound on a crawl. That is commented at the call site.

**8. A missing climb is not negatively cached.** Returning `null` from *inside* `unstable_cache` would store it for the full hour, so a climb a crawler reached minutes before its import landed would keep 404-ing until the entry expired. The executor throws a private not-found error instead — `unstable_cache` never stores a rejection — and the caller translates it back to `null`.

**9. Dead code removed.** `fetchListPageData` has had no caller since the climbing UI moved to Expo in W-16, and it still carried both anti-patterns this PR removes: degrade-to-empty on a read failure, and the uncapped 20-image warm fan-out.

### What the QA round changed

Three adversarial lenses (SEO/status, concurrency, test strength) reviewed the first draft and filed 22 findings. Everything real is fixed in this PR:

| finding | fix |
|---|---|
| `/b` tree still 404s a backend failure (`resolveBoardBySlug` swallowed everything into `null`) | fail-loud, item 1 above, + three transport-level tests and a page-level one |
| The 6 s ceiling was per *statement* — the climb page's real worst case was ~16 s, measured | per-request budget, item 2 |
| Deleting the digest early-rethrow made every routine 404 and 308 log at error level | the digest check is back as a **logging** guard; the rethrow stays unconditional |
| A missing row became negatively cached for an hour | item 8 |
| `DB_POOL_IDLE_TIMEOUT_S=0` was clamped to 1, inverting "never close" into "close every second" | no floor on the idle knob, item 4 |
| `cancelQuietly`'s promise-swallow branch was dead code (`Query.cancel()` returns `null`) | removed, and the real cost is documented: a **queued** cancel is free, an **executing** one opens a fresh connection whose failure the runtime logs |
| "a shed request is usually invisible behind the CDN" | false, and replaced with the actual window everywhere it appeared |
| `fetchListPageData` dead code, stale comments in `list-page-robots.ts` and the `/list` `generateMetadata` catch | deleted / corrected |
| Six inert or missing guards (see the test plan) | six new/rewritten tests, each mutation-probed |

### Deliberately out of scope

No front-door refactor. No web `/health/db` (a real observability gap — during a web-pool incident the backend's `/health/db` stays green and tells you nothing — but a separate change). No touching `withConnectRetry`, the backend pool or the sync-runner pools. No batching the alias lookup into the climb select. No `middleware.ts` change.

### Sizing note for #4436

`MAX_ITEMS_PER_SHARD` is 11,250 (`sitemap-xml.ts:14`) and W-22 fans every item out ×4 locales. 128,655 tier-2 climbs is **514,620 `<url>` entries → ~12 shards**, not the 3 in the epic body.

## Asks for @marcodejongh

Four production facts I cannot see from the repo. **None of them block this PR** — every default is unchanged, so the change is safe without the answers — but each one sharpens the numbers:

| # | question | how to answer |
|---|---|---|
| 1 | Railway `max_connections` | `psql "$DB_URL" -c 'show max_connections;'`, or Railway → Postgres → Variables |
| 2 | Is `READ_REPLICA_URL` set on the Vercel project? | Vercel → Settings → Environment Variables. If set, every instance opens **two** pools |
| 3 | Is Vercel's `DATABASE_URL` pooled (PgBouncer) or direct? | `psql "$DATABASE_URL" -c 'show pool_mode;'` — a direct Postgres errors, PgBouncer answers. Decides whether `DB_STATEMENT_TIMEOUT_MS` is usable at all |
| 4 | Is Fluid compute enabled? | Vercel → Settings → Functions. Decides whether one instance can ever reach `max` |

And a request, not a claim: setting **`DB_POOL_MAX=4`** and **`DB_POOL_IDLE_TIMEOUT_S=10`** on the Vercel project only. A front-door render needs at most 3 concurrent statements and the list page needs 1, so 4 leaves headroom without letting an instance claim 10 slots it will never use in parallel; dropping the idle hold 30 → 10 s cuts a bursting fleet's steady-state footprint roughly threefold. Those are **not set by this PR**.

## Test plan

Every guard below was mutation-probed: the mutation is named, it went red, and the file was restored.

- **`db-pool-exhaustion.test.ts` (backend, live Postgres) — the real oracle.** Six concurrent `pg_sleep(30)` through a `max: 2` pool with a 1500 ms deadline: everything settles inside 3 s (a serial drain would take three 30 s rounds), all six reject with `DB_READ_TIMEOUT`, and `pg_stat_activity` shows the cancelled statements gone. *Probe: break the `Promise.race` → the test times out.* It reaches across to `packages/web` for the helper because that is where #4461 rules the helper must live, and the backend project is the only Vitest project here with a live database; the specifier is computed so `tsc --rootDir src` never type-links across the boundary.
- **`read-deadline.test.ts`** — fake timers: rejects at the deadline, `.cancel()` called exactly once, not called on the success or the error path, `vi.getTimerCount() === 0` after both. Plus `parseReadDeadlineMs`, which the constant cannot exercise because it is captured at module load: unset → 6000, `'8000'` → 8000, `'0'` / `'-1'` / `'abc'` → 6000. *Probe: make a valid override fall through to the default → 3 fail.*
- **`read-deadline-wiring.test.ts` (new) — the caller-side half.** The helper being correct proved nothing about the front door using it: the whole web suite stayed green with `withReadDeadline` neutralised at all four sites. This asserts the labels `climb-alias`, `climb-select`, `climb-stats-all-angles`, `climb-search` are actually raised, and that every climb-page read gets a budget inside the ceiling. *Probe: unwrap the all-angles read → red.*
- **`climb-request-dedupe.test.ts` (new)** — substitutes a real memo for React's `cache` (the client build vitest resolves is a passthrough) and pins the two properties that ride the render scope: two calls with equal primitives from *different* params objects cost 2 statements not 4, a full page pass costs 3, a different uuid still costs 4, and the three reads receive `[6000, 5000, 4000]` from one shared budget. *Probes: drop the `cache()` wrapper → red; give each read the full deadline again → red.*
- **`climb-missing-not-cached.test.ts` (new)** — a memoising `unstable_cache` that stores resolved values and nothing for a rejection, i.e. the real semantics: a missing climb re-reads on the next request and recovers the moment the row lands, while a climb that exists is still cached. *Probe: return `null` from inside the cache again → red.*
- **`read-failure-status.test.tsx` × 2 (both trees)** — the mass-deindexing guard. `null` → `notFound()`; `DbReadTimeoutError` and a generic error → the page throws and `notFound` is **not** called; a failed `resolveBoardBySlug` throws rather than 404s; and a 404 does **not** write to `console.error`. *Probes: restore the trailing `notFound()` → 4 of 6 fail; drop the digest logging guard → red.*
- **`board-slug-utils.test.ts`** — a rejecting fetch, a 502, and a 200 carrying `errors` all throw; a clean 200 with `boardBySlug: null` still resolves `null`, and the existing auth/cache-split cases are unchanged. *Probe: swallow failures into `null` again → 3 fail.*
- **`front-door-fanout.test.ts`** — the statement budget: `getClimb` costs exactly 2 and they are strictly sequential (max in flight 1), the stats select costs 1. The old "a full pass costs 3" case was dropped: it was arithmetic on the other two and could not tell a deduped page pass from an un-deduped one. That assertion now lives in the dedupe test, where it can.
- **`search-climbs-moonboard-cache.test.ts`** — a *memoising* `unstable_cache` recorder that now also captures the options object: an explicitly-cacheable MoonBoard search executes once **at `revalidate: 900`**; the Kilter control keeps 86400; without the flag, opted out, personalised, or `sortBy: 'random'` it executes twice. *Probes: revert the `explicitlyCacheable` clause → red; raise the MoonBoard TTL to 24 h → red (it was inert before).*
- **`server-cached-client-timeout.test.ts` (new)** — the backend deadline had no coverage at all: deleting it left all 3426 web tests green. A fake `GraphQLClient` captures the constructor options; with `timeoutMs` the signal aborts at the deadline, without it no signal and no timer is created, and a timely answer clears the timer. *Probe: stop passing the signal → red.*
- **`front-door-data-timeout.test.ts` (new)** — the caller-side half: both front-door queries construct theirs with 3000. *Probe: drop the argument at both sites → red.*
- **Warm fan-out, both halves** — the unit assertion (6 fetches from 50 climbs, each carrying a signal) **and** the caller-side assertion that `maxImages: FRONT_DOOR_WARM_LIMIT` was actually passed. *Probe: drop `maxImages` at the call site → the caller half fails while the unit half stays green.*
- **Pool knobs, now in a project CI runs.** `packages/db` is not a Vitest project and the only db test CI invokes is the migration-journal one, so the knob assertions were inert in CI. They are mirrored into `shutdown.unit.test.ts` (backend project, every PR): defaults 10/30, honours both env vars, clamps `DB_POOL_MAX=1` to 2, falls back on `abc`, `DB_POOL_IDLE_TIMEOUT_S=0` stays 0, and `statement_timeout` is emitted only when asked. All read the resolved options off the constructed client, never the source text. *Probe: clamp the idle knob to a floor of 1 → red.* That file's old assertion grepped `postgres.ts` for the literal `idle_timeout: 30`, which would have gone green-but-meaningless once the value came from env — rewriting it is part of this change, not scope creep.

Gates run:

| command | result |
|---|---|
| `vp check` | formatting clean; 33 repo-wide lint errors, all in `scripts/**` files this diff does not touch (`main` itself reports 36) |
| `vp run typecheck` | green (full graph, `build:web` included) |
| `vp test run --project web` | **315 files, 3454 tests, all pass** |
| `vp test run --project backend` | **257 files, 3542 tests, all pass** (includes the live-Postgres oracle) |
| `vp run test:db` (node:test) | 425 pass / 3 fail — the 3 are the pre-existing `createClimbFilters` zone-mode and MoonBoard hold-search failures, unchanged by this diff; the pool-knob file itself is 14/14 |

`vp run test:e2e` not run: this PR changes no rendered copy. `check:i18n` not applicable — no user-facing strings added (the 500 path uses the existing `error.tsx`).

One correction to the first commit message on this branch: it says the deadline "cancel()s the pending query" without the drizzle carve-out. The code, the docs and this body all state it accurately; the commit message is left as-is rather than force-pushing a rebase over green CI.

## Release Notes

- Climb and board pages now stay up when the database is slow, and stop turning a bad moment into a "page not found".
- MoonBoard board pages load from cache instead of running a fresh search every time.

### DIFF
--- a/docs/db-connectivity.md
+++ b/docs/db-connectivity.md
@@ -99,6 +99,80 @@ off) rather than reaching into postgres.js's backoff.
 The connect-retry tests pin their pools to `backoff: () => 0` so they measure this
 wrapper's loop rather than that ramp.
 
+## Front-door read deadlines and pool sizing (#4461)
+
+The connect retry above bounds a _failed_ connect. It does nothing about a
+_saturated_ pool, and that is the failure the climb sitemaps invite: postgres.js
+has no acquire timeout, and its internal queue is unbounded and untimed
+(`postgres/src/index.js:341`). A statement that cannot get a connection waits
+forever, so many concurrent SSR renders against a slow database look like a hang
+rather than an error.
+
+`packages/web/app/lib/db/read-deadline.ts` bounds one read client-side. It races
+the pending query against a timer, rejects with `DbReadTimeoutError`
+(`code: 'DB_READ_TIMEOUT'`) when the timer wins, and calls `query.cancel()` the
+way the health probe does, so a timed-out statement does not fire later against
+a recovered pool. It is wired at the four front-door reads only — the two
+statements behind `getClimb`, the all-angles stats select, and the shared climb
+search — and deliberately **not** inside `withConnectRetry` or `packages/db`,
+where it would change behaviour for the backend, the sync runners and every
+script.
+
+### What the reader sees when it fires
+
+| condition                           | before                               | after                                                   |
+| ----------------------------------- | ------------------------------------ | ------------------------------------------------------- |
+| climb row absent                    | 404                                  | 404 (unchanged — now for the right reason)              |
+| read fails or deadlines, climb page | hung to the platform limit, then 404 | 500 at ~6 s; the CDN stale window keeps serving readers |
+| read fails or deadlines, list page  | 200 with zero climbs                 | 500 at ~6 s; same CDN window                            |
+| backend GraphQL wedged              | climb page hung indefinitely         | similar climbs / beta links render empty at 3 s         |
+
+The 5xx is the point. Google retries a 5xx and keeps the URL, while a 404 — or a
+200 with nothing on it — on a sitemapped URL reads as "drop this page".
+
+### Budgets
+
+| knob                      | default | meaning                                                                                  |
+| ------------------------- | ------- | ---------------------------------------------------------------------------------------- |
+| `DB_READ_DEADLINE_MS`     | 6000    | web front door: wall clock for one read (queue wait + connect + execute)                 |
+| `DB_POOL_MAX`             | 10      | postgres.js `max`, clamped to a floor of 2                                               |
+| `DB_POOL_IDLE_TIMEOUT_S`  | 30      | seconds an idle connection is held open                                                  |
+| `DB_STATEMENT_TIMEOUT_MS` | unset   | emits a `statement_timeout` startup parameter — **off by default**, see the hazard below |
+
+6000 sits deliberately below `DB_CONNECT_RETRY_BUDGET_MS` (10000): during a
+brownout the front door should shed load rather than spend a second and third
+connect attempt holding a pool slot while a crawler waits.
+
+The pool knobs default to the values that used to be hard-coded, so nothing
+changes for a deployment that does not set them. They exist because peak
+server-side connections scale with **instance count × connections held idle**,
+not with per-instance `max` — a fleet of serverless instances each sitting on a
+few idle connections for 30 s is the term that grows during a crawl burst.
+Lowering `DB_POOL_MAX` and `DB_POOL_IDLE_TIMEOUT_S` on a serverless deployment
+shrinks that footprint; raising `max` never helps.
+
+### The `statement_timeout` hazard
+
+`DB_STATEMENT_TIMEOUT_MS` ships off. PgBouncer in transaction-pooling mode
+rejects startup parameters that are not in `ignore_startup_parameters`, and
+`statement_timeout` is not among the defaults — so setting it against a pooled
+`DATABASE_URL` fails _every connection_ instead of bounding one query. Two
+paths, chosen by what the URL actually points at:
+
+- **Direct Postgres** — set `DB_STATEMENT_TIMEOUT_MS` on the deployment.
+- **Pooled (PgBouncer) URL** — do it database-side instead, with
+  `ALTER ROLE <app_role> SET statement_timeout = '8s'`, which passes through a
+  pooler transparently.
+
+`psql "$DATABASE_URL" -c 'show pool_mode;'` tells you which you have: a direct
+Postgres errors, PgBouncer answers.
+
+### There is no web health probe
+
+`/health` and `/health/db` below are **backend** endpoints. `packages/web` has
+none, so during a web-pool incident `/health/db` stays green and says nothing
+about the surface that is actually failing. That gap is not filled here.
+
 ## Where the retry is wired
 
 `createDb()` / `createReadDb()` hand drizzle a retry-wrapped view of the pool
--- a/docs/db-connectivity.md
+++ b/docs/db-connectivity.md
@@ -109,39 +109,78 @@ forever, so many concurrent SSR renders against a slow database look like a hang
 rather than an error.
 
 `packages/web/app/lib/db/read-deadline.ts` bounds one read client-side. It races
-the pending query against a timer, rejects with `DbReadTimeoutError`
-(`code: 'DB_READ_TIMEOUT'`) when the timer wins, and calls `query.cancel()` the
-way the health probe does, so a timed-out statement does not fire later against
-a recovered pool. It is wired at the four front-door reads only — the two
-statements behind `getClimb`, the all-angles stats select, and the shared climb
-search — and deliberately **not** inside `withConnectRetry` or `packages/db`,
-where it would change behaviour for the backend, the sync runners and every
-script.
+the pending query against a timer and rejects with `DbReadTimeoutError`
+(`code: 'DB_READ_TIMEOUT'`) when the timer wins. It is wired at the four
+front-door reads only — the two statements behind `getClimb`, the all-angles
+stats select, and the shared climb search — and deliberately **not** inside
+`withConnectRetry` or `packages/db`, where it would change behaviour for the
+backend, the sync runners and every script.
+
+**Cancellation covers three of the four.** On a timeout the helper calls
+`query.cancel()` the way the health probe does, so a timed-out statement does not
+fire later against a recovered pool. That only works for raw postgres.js
+queries. The list front door's search is drizzle-issued and exposes no
+`.cancel()`, so there the deadline sheds the _caller_ while the statement runs to
+completion still holding its connection. Worth knowing mid-incident: shedding
+list renders does not immediately hand connections back.
+
+Cancelling a **queued** query is free — postgres.js removes it from the queue and
+rejects it locally. Cancelling one that is already **executing** opens a
+brand-new connection to send the cancel request, against a database that is by
+construction already struggling, and postgres.js keeps that connection's promise
+internally (`Query.cancel()` returns `null`), so a failure to open it surfaces as
+an unhandled rejection the runtime logs. Accepted: a zombie statement firing
+against a recovered pool is worse than a log line.
+
+**One budget per request, not per statement.** The climb page issues three reads
+in sequence, so three independent 6 s deadlines would be an ~18 s request
+ceiling — the opposite of shedding load. `app/lib/db/request-read-budget.ts`
+puts one deadline timestamp in React's per-render `cache` scope and hands each
+read whatever the earlier ones left, floored at 500 ms. Outside a render scope
+(scripts, unit tests) React's `cache` is a passthrough and each read gets its own
+deadline.
 
 ### What the reader sees when it fires
 
-| condition                           | before                               | after                                                   |
-| ----------------------------------- | ------------------------------------ | ------------------------------------------------------- |
-| climb row absent                    | 404                                  | 404 (unchanged — now for the right reason)              |
-| read fails or deadlines, climb page | hung to the platform limit, then 404 | 500 at ~6 s; the CDN stale window keeps serving readers |
-| read fails or deadlines, list page  | 200 with zero climbs                 | 500 at ~6 s; same CDN window                            |
-| backend GraphQL wedged              | climb page hung indefinitely         | similar climbs / beta links render empty at 3 s         |
+| condition                           | before                               | after                                           |
+| ----------------------------------- | ------------------------------------ | ----------------------------------------------- |
+| climb row absent                    | 404                                  | 404 (unchanged — now for the right reason)      |
+| board slug resolves to nothing      | 404                                  | 404 (unchanged)                                 |
+| read fails or deadlines, climb page | hung to the platform limit, then 404 | 500 at ~6 s per request                         |
+| read fails or deadlines, list page  | 200 with zero climbs                 | 500 at ~6 s                                     |
+| backend `boardBySlug` fails, `/b/…` | 404                                  | 500                                             |
+| backend GraphQL wedged              | climb page hung indefinitely         | similar climbs / beta links render empty at 3 s |
 
 The 5xx is the point. Google retries a 5xx and keeps the URL, while a 404 — or a
 200 with nothing on it — on a sitemapped URL reads as "drop this page".
+
+**The CDN does not hide it, and that is fine.** `stale-while-revalidate` only
+cushions a URL whose `age` has already passed `s-maxage`: climb views are covered
+for age 1 h–7 h, list front doors for 24 h–7 d, and Vercel supports no
+`stale-if-error`. A cold-MISS crawl of a long-tail sitemap URL therefore sees the
+500 directly. The asymmetry that matters is what happens next: 5xx is **not** a
+cacheable status on Vercel, so it is never pinned, whereas the 404 this replaces
+**is** cacheable — one blip used to stick a 404 in the CDN for a full `s-maxage`.
+
+A genuinely missing climb is also not negatively cached: `fetchClimbFromDb`
+throws a private not-found error inside `unstable_cache` (which never stores a
+rejection) and the caller turns it back into `null`, so a climb crawled minutes
+before its import lands recovers on the next request rather than 404-ing for the
+rest of the hour-long entry.
 
 ### Budgets
 
 | knob                      | default | meaning                                                                                  |
 | ------------------------- | ------- | ---------------------------------------------------------------------------------------- |
-| `DB_READ_DEADLINE_MS`     | 6000    | web front door: wall clock for one read (queue wait + connect + execute)                 |
+| `DB_READ_DEADLINE_MS`     | 6000    | web front door: wall clock for one _request's_ reads (queue wait + connect + execute)    |
 | `DB_POOL_MAX`             | 10      | postgres.js `max`, clamped to a floor of 2                                               |
-| `DB_POOL_IDLE_TIMEOUT_S`  | 30      | seconds an idle connection is held open                                                  |
+| `DB_POOL_IDLE_TIMEOUT_S`  | 30      | seconds an idle connection is held open; `0` means "never close one" and is not clamped  |
 | `DB_STATEMENT_TIMEOUT_MS` | unset   | emits a `statement_timeout` startup parameter — **off by default**, see the hazard below |
 
 6000 sits deliberately below `DB_CONNECT_RETRY_BUDGET_MS` (10000): during a
 brownout the front door should shed load rather than spend a second and third
-connect attempt holding a pool slot while a crawler waits.
+connect attempt holding a pool slot while a crawler waits. That comparison only
+holds because the budget is per request — see the shared-budget note above.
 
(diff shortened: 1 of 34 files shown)


marcodejongh:
> **1. New i18n key usages with no visible catalog additions (medium)**

Not new keys. `metadata.view.fallbackTitle` and `metadata.view.fallbackDescription` were already used by the `catch` block in **both** `generateMetadata` functions before this PR, and they exist in all four catalogs on `main`:

```
$ git show origin/main:packages/shared/i18n/locales/<locale>/climbs.json | jq .metadata.view
en-US: fallbackTitle='Climb'    fallbackDescription='View climb details and beta videos'
es:    fallbackTitle='Bloque'   fallbackDescription='Mira los detalles del bloque y vídeos de beta'
fr:    fallbackTitle='Bloc'     fallbackDescription='Voir les détails du bloc et les vidéos de beta'
de:    fallbackTitle='Boulder'  fallbackDescription='Boulder-Details und Beta-Videos ansehen'
```

All this PR does is reach the same two keys from one more branch — the `getClimb` → `null` case — which previously arrived at the identical fallback via the undefined-row throw. No catalog change is needed, and `catalog-completeness.test.ts` has nothing new to check.

> **2. Cross-package dynamic import in the pool-exhaustion test (low)**

Agreed as debt, and deliberate. The alternatives were worse: putting `withReadDeadline` in `packages/db` is explicitly out of scope (it would change behaviour for the backend, the sync runners and every script), and a literal relative import fails `tsc --rootDir src` on the backend build. The backend project is the only Vitest project in the repo with a live Postgres, and a bounded-wait oracle without a real pool would just be a tautology. The specifier and the reasoning are commented in the file so the next reader is not guessing.

marcodejongh:
QA round 2: all 22 findings from the three lenses are addressed in `fb3e0d41` — the PR body has the table. One partial rebuttal worth recording rather than acting on further:

**`cancelQuietly` (concurrency [2])** — the dead `.then` branch is real and is removed: `Query.cancel()` is `this.canceller && (this.canceller(this), this.canceller = null)`, a comma expression that always evaluates to `null` (`postgres/src/query.js:52`), so the promise postgres.js builds at `index.js:350` never reaches out here. The *framing* is overstated, though. The finding reads as "every timed-out read adds a connection attempt to a database that is already drowning". It only does that for a statement that already reached the server. For one still **queued** — which is exactly the saturated-pool case this deadline exists for — `index.js:350` takes the other branch: it removes the query from the queue, rejects it locally with `57014`, and opens nothing. Since `withReadDeadline` attaches to the pending query before the race, that rejection is already handled. Both regimes are now spelled out in the code and in `docs/db-connectivity.md`.

marcodejongh:
## The four dashboard/prod questions this PR listed as unanswerable — three are now answered

Marco gave me a read-only production connection. Measured 2026-08-15, read-only, aggregate only.

### OQ-2 — the connection ceiling is **200**, not the 100 the plan assumed

```
max_connections                 | 200
superuser_reserved_connections  | 3
server_version                  | 16.9 (Debian 16.9-1.pgdg110+1)
```

Every row in the plan's §3.2 blow-up table should be re-read against 200 rather than 100. It **does not change the conclusion** — the burst rows (465–1,515 connections) still clear the ceiling by a wide margin, and the load-bearing point stands: the multiplier is *instance count*, not per-instance `max`, so raising `max` is strictly harmful and lowering it helps the tail without removing the risk. But the headroom before the first row hurts is roughly double what the plan assumed, which is worth knowing before picking `DB_POOL_MAX`.

Also note it is **Postgres 16.9**, not 17 — the plan's "stock PostgreSQL 17 ships `max_connections = 100`" reference point does not apply.

### OQ-2b — what is actually in use at rest

```
app         | state  | conns
postgres.js | idle   | 12
postgres.js | active |  1
```

13 connections during a quiet period, all from `postgres.js`. That matches the plan's "today, quiet ≈ 20" estimate closely and confirms `idle_timeout: 30` is doing exactly what the plan says — a dozen connections held open with nothing running on them. It is direct evidence for the `DB_POOL_IDLE_TIMEOUT_S` recommendation.

### OQ-4 — **not** a PgBouncer pooler, so the `statement_timeout` hazard does not apply to this endpoint

```
SHOW pool_mode  →  ERROR: unrecognized configuration parameter "pool_mode"
```

PgBouncer answers that; direct Postgres errors exactly this way. So `connection: { statement_timeout }` as a startup parameter would **not** fail here, and the reason this PR ships it default-off is weaker than the plan believed.

**I still would not flip it on in this PR.** Two caveats: this is the read-only proxy endpoint and I cannot prove the app's `DATABASE_URL` resolves to the same one (it very likely does — Railway's TCP proxy is a passthrough), and turning on a global statement timeout is a production behaviour change that deserves its own deploy rather than riding along. Keep it env-gated and off; the docs section should now say "the endpoint we could test is direct Postgres, so the pooler hazard is theoretical here" rather than presenting it as an unresolved blocker.

### OQ-3 / OQ-5 — still open

`READ_REPLICA_URL` and Fluid compute are Vercel project settings; a database connection cannot answer either. They remain asks for Marco, and both only change the arithmetic, not the change set.

### What I would change in the PR

Only prose: the `max_connections` figure and its table, the Postgres version, and the `statement_timeout` rationale in `docs/db-connectivity.md`. No code. Say the word and I will push it, or the author can fold it in.

Issue #4443, opened 2026-08-14
W-25b: delete the Aurora proxies


marcodejongh:
Part of #4358. **Blocked by:** #4441 (W-25a — the `410 Gone` deprecation) **plus one release**. QA gate #4359 is **SIGNED and closed (2026-08-14)**, so the gate dependency is satisfied.

## Context

W-25a deleted the three undocumented Aurora proxies, deregistered `proxy/login` and `proxy/saveAscent` from the OpenAPI spec (`openapi-routes.ts:485,524`) and made them return `410 Gone` with `Deprecation`/`Sunset` headers. `/docs` is indexable and sitemapped, so **"no in-repo caller" ≠ dead for `/api/v1/*`** — the deprecation window exists for third-party callers. This PR closes it.

Feeds **#1889** (decommission redundant REST endpoints).

## The work

- Delete the **two routes and their tests**.
- Swap the `/api/` sample path in **`robots.test.ts:77`** (`/api/v1/kilter/proxy/search` → `/api/v1/kilter/grades`), not `middleware.test.ts:706` — that file has no `/api/v1` or proxy sample anywhere in it. (Amended 2026-08-19: the original checkbox named the wrong file/line.)

## Standing-rule callouts

- **Ships early, deliberately.** Marco ruled 2026-08-19 to delete now, 43 days before the published `Sunset: Thu, 01 Oct 2026` header — do not wait for the sunset date. The PR body must name the release that shipped W-25a (the 2026-08-15 merge to `main`; `www` has no separate release train) and state plainly that this lands ahead of the advertised Sunset and why: both routes have answered `410` with no Aurora call behind them since W-25a, so any caller still hitting them is already broken — this PR changes only *how* they are broken, `410` → `404`. Do not imply the Sunset date has passed. (Amended 2026-08-19.)
- **e2e spec deletions ship with this PR (standing rule 6)** — `e2e-tests.yml` is `workflow_dispatch`-only, so nothing catches the lag; specs live in `packages/web/e2e/`, not `e2e/`.
- **Every redirect gets `es`, `fr` and `de` siblings (standing rule 4)** if this PR adds one.
- **Never delete `/api/internal/ws-auth` or `/api/internal/board-render`** — both are pinned by shipped hardware or ~85 kept web files. This PR touches the `/api/v1/*` Aurora proxies only.

## Deps

- W-25a — #4441, **plus one release**
- QA sign-off — #4359 (signed 2026-08-14)

## Gates

[...]

Issue #1889, opened 2026-05-04
REST surface audit: 39 routes classified — 37 KEEP, 1 method-level deprecate, 1 orphan kept


marcodejongh:
Parent: #1861. Part of #1859 (closed epic; live parent is #1861 plus `epic:web-reposition`).

## The audit is the spec

This issue's original premise — "after Phase 1 batches replace REST callers with GraphQL, delete the redundant REST endpoints" — is **false for `/api/v1/*`**. That surface is published in the OpenAPI document rendered at the indexable, sitemapped `/docs` and served as a crawlable `/openapi.json`. Zero in-repo callers is the *intended steady state* of a published API, not evidence it's dead. W-25a/W-25b (deleting the two Aurora proxies) are the precedent, and they were deprecable because the **backing feature** (Aurora login/save-ascent) was retired — not because nothing in the repo called them.

**The corrected rubric:** for any route under `/api/v1/*` (or otherwise published on `/docs`), the decommission test is *"has the backing feature been retired?"*, never *"who calls it?"*.

Audited 2026-08-19 against `main`, re-verified 2026-08-22 against `main` (post W-25b merge + the sitemap tier-2 materialisation, #4661): **39 route files, 37 KEEP, 1 method-level deprecate, 1 orphan (kept).**

### KEEP — external consumer, no in-repo caller by design (13)

| Route | Why it stays |
| --- | --- |
| `/api/internal/board-render` | `board_config.h:41` (+ the MoonBoard-dev-server copy) + `thumbnail_client.cpp:308` — shipped ESP32 firmware we cannot push an update to. |
| `/api/og/climb` | 307 redirect shim to `/api/internal/board-render`, kept for every already-shared `og:image` URL and crawler cache from before the OG-card migration to the backend (`ws.boardsesh.com/og/climb`, see `docs/og-climb.md`). `robots.ts` explicitly allows `/api/og/` for Googlebot-Image. New metadata never emits it (`board-renderer/__tests__/util.test.ts:168`, `metadata.test.ts`). |
| `/api/v1/{board_name}/grades` | Documented, `openapi-routes.ts:46`. |
| `/api/v1/grades/{board_name}` | Documented alternate, `openapi-routes.ts:78` — byte-identical query, missing the canonical route's cache header (see Follow-ups). |
| `/api/v1/angles/{board_name}/{layout_id}` | Documented, `:102`. |
[...]

Issue #4488, opened 2026-08-15
main CI has not been green in 30 runs: full-repo lint debt plus an unhandled rejection in a passing backend suite


marcodejongh:
`main`'s CI has not been green in the **last 30 runs**. Every completed run since at least `686717545` (2026-08-14T21:44Z) is a failure or a cancellation — 19 failures, 10 cancelled, **0 successes**.

The cost is not the red itself. It is that a permanently red `main` cannot distinguish a real break from the background noise, so nobody looks — and the next genuine regression lands invisibly.

## There are exactly two causes, and both are bounded

Measured on run [31880202330](https://github.com/boardsesh/boardsesh/actions/runs/31880202330) (`9c38435de`), and the same two recur across the sampled failures.

### 1. `lint` — 33 errors / 325 warnings in the full-repo pass

```
Found 33 errors and 325 warnings in 4318 files (28.2s, 4 threads)
error: Lint or type issues found
```

**Why nobody sees this on a PR:** PR CI diff-lints, `main` lints the whole repo. So a file nobody's PR touches can accumulate errors indefinitely and every PR stays green while `main` goes red. `vp check` reproduces it locally — I get the same 33/325 on a clean checkout of `main`.

The flagged files are concentrated in test code and are mostly one rule family — `no-unnecessary-condition` / `no-base-to-string` style diagnostics on optional chains and template literals:

```
packages/backend/src/__tests__/mask-error.test.ts:173
  expect((captureOptions().extra?.failedQuery as string).length).toBe(2000);
  help: If this short-circuits with 'undefined' the evaluation will throw TypeError
```

plus similar in `packages/db`, `packages/mobile`, `packages/aurora-sync` and `embedded/scripts`. None of it is in `packages/web`'s hot paths.

### 2. `test-backend (2, 2)` — an unhandled rejection escapes a passing suite

**All 1,712 backend tests pass.** The job still exits 1:

```
Test Files  128 passed (128)
Tests       1712 passed (1712)
Errors      1 error

⎯⎯⎯⎯ Unhandled Rejection ⎯⎯⎯⎯
Error: Board not found
 ❯ db.transaction.isolationLevel packages/backend/src/graphql/resolvers/social/boards.ts:2387:17
 ❯ sql.begin node_modules/…/postgres/src/index.js:243:14
 ❯ processTicksAndRejections

This error originated in "packages/backend/src/__tests__/board-merge-tombstone.test.ts".
The latest test that might've caused the error is
[...]

PR #4489, opened 2026-08-15
chore(web): deprecate the Aurora proxies with 410 Gone

### LINKED ISSUE #4441: W-25a: deprecate the Aurora proxies with 410 Gone
Part of #4358. **Blocked by:** #4437 (W-19). **Gate-independent** — QA gate #4359 is signed and closed (2026-08-14) regardless.

## Context

There are **zero in-repo runtime callers for all five Aurora proxies**, but **`proxy/login` and `proxy/saveAscent` are published in the OpenAPI spec** (`openapi-routes.ts:485,524`) rendered at `/docs`, which is **indexable and sitemapped**: **"no in-repo caller" ≠ dead for `/api/v1/*`.** Third parties may be calling them. This PR deprecates rather than deletes; W-25b removes the routes one release later.

Feeds **#1889** (decommission redundant REST endpoints).

## The work

- **Delete the three undocumented proxies.**
- **Deregister the two documented ones** (`proxy/login`, `proxy/saveAscent`) from the OpenAPI spec at `openapi-routes.ts:485,524`.
- Return **`410 Gone`** from the two, with **`Deprecation` and `Sunset` headers**.

## Standing-rule callouts

- `/docs` is indexable and sitemapped — deregistering the two entries changes a **public, crawlable** contract. Keep the page's metadata coherent after the removal.
- **Every redirect gets `es`, `fr` and `de` siblings (standing rule 4)** if this PR adds any user-facing redirect (API routes are locale-free, but a docs-page redirect is not).
- Zod for API validation; **Drizzle over raw SQL**; never `any`.
- The `Sunset` date must line up with **W-25b landing one release later** — state the date in the PR body.

## Deps

- W-19 — #4437

## Gates

## Verifies

- [ ] The three undocumented proxies are gone.
- [ ] `proxy/login` and `proxy/saveAscent` return **`410 Gone`** with `Deprecation` and `Sunset` headers, asserted by test.
- [ ] Neither appears in the OpenAPI spec or at `/docs` any more.
- [ ] The `Sunset` date and the planned W-25b release are recorded in the PR body.

## Tiering

**Sonnet** implementation (mechanical, clear spec). **A Sonnet implementation never reviews itself** — the QA lens is **Opus**, three adversarial agents each prompted to refute (the SEO lens covers the `/docs` change).

## Release Notes

Internal-only unless a third party is calling those endpoints — if the deprecation is worth announcing, write it in climber voice; otherwise `none` with the template checkbox ticked.

marcodejongh:
## Summary

W-25a of the web-reposition epic (#4358). Closes #4441.

The five `/api/v1/{board}/proxy/*` Aurora routes have had **zero in-repo runtime callers** since the GraphQL migration. Two of them are also published in the OpenAPI document rendered at `/docs`, which is indexable (`createPageMetadata`, no `/docs` disallow in `app/robots.ts`) and sitemapped (`static-entries.ts`), and the document itself is served as a plain crawlable file at `/openapi.json`. So "no in-repo caller" is not "dead" here, and this is a deprecation rather than a delete.

**Delete (3, undocumented):** `getLogbook`, `saveClimb`, `user-sync`.
**Deprecate (2, documented):** `login`, `saveAscent` → `410 Gone`, deregistered from the spec. W-25b (#4443) removes the URLs.

### The `Sunset` date: `Thu, 01 Oct 2026 00:00:00 GMT`

45 days after merge. Why that and not "one release later" verbatim:

- **There is no web release train to count.** `git tag` carries only `rn-android-dev-*` (mobile dev-client builds, several a day) and one campaign artefact. www deploys on merge to `main` via `production-deploy.yml`, several times a day, and `vercel.json` sets `"git": { "deploymentEnabled": false }`. The second merge after this one would satisfy "one release later" literally and mean nothing to an external caller. The only user-visible release unit is the mobile store release (~fortnightly), which third-party API consumers do not track either.
- **45 days ≥ one full monthly polling cycle.** Any integration running less often than daily still sees at least one 410-with-`Sunset` before the URL starts 404ing. A two-week window would not guarantee that.
- **The marginal harm after the sunset is small.** From day one of this PR the endpoints are already non-functional — 410, no Aurora call behind them. W-25b only changes 410 → 404. Anyone still calling them is already broken; the date governs only how *informatively* they stay broken. That is why 45 days is enough here where six months would be right for an endpoint that still works.
- It is a round, unambiguous UTC midnight, exactly what `new Date('2026-10-01T00:00:00Z').toUTCString()` emits in RFC 8594's IMF-fixdate form.

### The `Deprecation` date: `@1786752000` = `2026-08-15T00:00:00Z`

A `Deprecation` sf-date that is **in the future** contradicts the 410 it rides on — RFC 9745 §2 reads a future date as "this resource *will become* deprecated then", i.e. "still supported for now", on a response that says the opposite. The date is therefore set at or before the earliest possible merge rather than at a guessed merge date, and it is pinned three ways in `app/lib/__tests__/api-deprecation.test.ts`:

1. `AURORA_PROXY_DEPRECATION_DATE.toISOString()` must equal `'2026-08-15T00:00:00.000Z'` (so bumping the constant forces this line to be bumped with it);
2. the constant must be `<= Date.now()` (so it can never drift forward);
3. the emitted header must be exactly `'@1786752000'`.

The `Sunset` half is pinned the same way, to `'Thu, 01 Oct 2026 00:00:00 GMT'`.

### Header shapes

| Header | Value | Standard |
|---|---|---|
| `Deprecation` | `@1786752000` | RFC 9745 §2 — a Structured Fields `sf-date`: `@` + epoch seconds. **Not** the superseded `Deprecation: true` draft form. Pinned to the literal, and bounded to the past. |
| `Sunset` | `Thu, 01 Oct 2026 00:00:00 GMT` | RFC 8594 §3 IMF-fixdate. Pinned to the literal. |
| `Link` | `<https://www.boardsesh.com/docs#retired-endpoints>; rel="deprecation"; type="text/html"` | RFC 9745 §3 — pair `Deprecation` with human-readable context. The whole header string and the body's `documentation` field are pinned, and a test asserts the fragment actually exists in `docs-client.tsx`. |
| `Cache-Control` | `no-store` | A 410 is cacheable by default under RFC 9110, and these paths took credentials in a POST body. No shared cache should hold a response for them. |

The two headers describe **two different transitions**, which is the whole W-25a/W-25b design: `Deprecation` = *the endpoint stopped working* (now). `Sunset` = *the URL stops existing* (W-25b, 2026-10-01, when it becomes a 404). Both routes answer `GET` as well as `POST` — `POST` is the documented verb, but `GET` is what a human following the `/docs` link or a naive crawler sends, and a bare 405 there carries no `Sunset` header and no pointer.

`DEPRECATION_DOCS_URL` goes through `absoluteUrl('/docs#retired-endpoints')` (`app/lib/seo/base-url.ts`), the same source of truth `robots.ts` and the sitemap builders use, rather than a second hardcoded copy of the canonical host.

### The deprecation link now points at something

A `rel="deprecation"` link is a promise that there is something to read at the other end, and after deregistration `/docs` had nothing: no proxy operations in the Swagger pane, no mention of a sunset, no pointer at the replacement. So `/docs` gains a short **"Retired endpoints"** card on the overview tab (`id="retired-endpoints"`), naming both paths, saying they answer 410, giving the 2026-10-01 date after which they 404, and pointing at the GraphQL API. Copy added in all four locales (`docs.overview.retired.{title,body}`).

The `/docs` sitemap entry's `lastModified` moves from `2026-04-30` to `2026-08-15` in the same commit — that file's own header comment makes it the rule ("update the date when you ship a real copy change"), and `/help` carries the same precedent from W-16. Without it, crawlers are told nothing changed since April while the page lost two operations, a tag, an overview bullet and a feature bullet, and gained a card.

### Telemetry — the reason this PR is worth more than a delete

Every call to a retired proxy now emits, from `app/lib/analytics.server`:

- **Event:** `Deprecated Aurora Proxy Called`
- **Properties:** `endpoint` (`'login' | 'saveAscent'`), `boardName` (lower-cased), and `method`. **Nothing else** — never the request body, which carried Aurora credentials on `login` and a session token on `saveAscent`.
- **Headers:** an allowlist of `user-agent` + `x-forwarded-for`, never `request.headers`. `@vercel/analytics`'s server `track` copies `cookie` verbatim off whatever headers it is handed onto its outbound request to `/_vercel/insights/event` (`dist/server/index.mjs`: `cookie: tmp.cookie`), and `saveAscent` callers still send a NextAuth session cookie. A dumb 410 counter needs no identity at all.
- Fired as `void track(...).catch(() => {})` inside a `try`/`catch`, so neither a rejected nor a synchronously-thrown telemetry call can turn a 410 into a 500.

**`method` is a deliberate departure from the ruling's "endpoint and boardName only".** That constraint was a credential-safety rule ("never the request body"), and the method is neither body data nor a credential. Without it the counter cannot answer the question it exists for: both verbs answer 410, `/docs` is indexable and `/openapi.json` is crawlable, and `*/proxy/login` is one of the most-scanned path shapes on the internet — so crawler `GET`s and a real integration's `POST`s would be byte-identical events. **W-25b's sign-off should read the `POST` count, not the total.** (Also noted on #4443.)

`boardName` is lower-cased because `middleware.ts:146` lowercases the segment only for its membership test and then passes the original path through, so `/api/v1/Kilter/proxy/login` reaches the handler and would otherwise split the counter across casing variants.

W-25b's gate is "is anyone still calling these?", and today nothing in the repo can answer it (see below). Ten lines converts that decision from an argument into a number.

### What I can and cannot know about third-party callers

**Can know, and verified:** nothing that ships from this repo calls any of the five. Repo-wide greps for `proxy/login|proxy/saveAscent|proxy/saveClimb|proxy/getLogbook|proxy/user-sync`, for `/proxy/` under `packages/`, and for dynamically-constructed proxy paths (`proxy/${`) return only route-file header comments, the old `saveAscent` test's URL, the two OpenAPI registrations, and doc/prose mentions. Zero runtime `fetch` call sites; zero dynamic construction. Two near-misses cleared: `packages/web/e2e/`'s `proxy` hits are all the `/app` Metro dev proxy, and mobile's ~20 `getLogbook` hits are `useBoardActions().getLogbook`, the GraphQL-backed board-provider action.

**Cannot know from the repo:** whether a third party is calling them. Server-side sources that could have answered it, and why none did:

1. **Vercel Runtime Logs** — carry path and status for every request, but retention is hours-to-days without a Log Drain, and no Log Drain is configured in-repo. They answer "is anyone hitting it *right now*", not "did anyone hit it last quarter".
2. **Vercel Web Analytics server events** — the only in-repo server-side channel, and it was **not** wired into any proxy route until this PR.
3. **Sentry** — errors only. A 200 or a 410 never reaches it.
4. **PostHog** — `posthog-js-lite` is client-side only in web. Blind to server requests.

Bottom line: **we have proven nothing we ship calls these; we have not proven nobody calls these.** As of this PR the 410 handler emits a counter, so W-25b can read a number instead of repeating the same non-answer — but it needs a real observation window after this merges. No number is manufactured here.

### The four orphaned implementation modules are deliberately left for W-25b

Deleting three routes and gutting two more orphans these (each verified to have zero remaining consumers afterwards, 876 lines total):

| Module | Lines | Own tests? |
|---|---|---|
| `packages/web/app/lib/data/get-logbook.ts` | 82 | no |
| `packages/web/app/lib/api-wrappers/aurora/saveAscent.ts` | 157 | no |
| `packages/web/app/lib/api-wrappers/aurora/saveClimb.ts` | 113 | no |
| `packages/web/app/lib/data-sync/aurora/user-sync.ts` | 524 | yes — `__tests__/user-sync-circuits.test.ts`, which encodes the #3526 cross-link finding |

**They stay, on purpose: the point of a deprecation window is that W-25a is reversible.** If a third party surfaces during the sunset window and we decide to un-410 an endpoint, that is a one-file revert while the implementations are still there — and an archaeology exercise if they are not. Deleting them would also drag a 524-line module carrying a hard-won regression test into a PR that is otherwise mechanical.

Nothing reds from leaving them: the import-graph invariant checks *edges into the delete set*, not orphans, and there is no `knip`/`ts-prune` step. The list is posted on #4443 so W-25b inherits it rather than re-deriving it.

The web `app/lib/data-sync/aurora/user-sync.ts` above is **not** `packages/aurora-sync/src/sync/user-sync.ts` — the latter is the live backend syncer and is untouched here. `app/lib/api-wrappers/aurora/` itself survives (`saveAttempt.ts`, `getBidsLogbook.ts`, `userSync.ts`, `types.ts` all still have consumers), as do `app/lib/data-sync/aurora/{json-import-stream,parse-aurora-export}.ts`.

### `/docs` coherence (the SEO half)

Deregistering two operations changes a public, crawlable contract, so everything that only existed to describe them goes with them:

- both `registerPath` blocks and their three component schemas (`AuroraLoginRequest`, `AuroraLoginResponse`, `SaveAscentRequest` — each had exactly one consumer);
- the `Aurora Proxy` tag object, because Swagger UI renders a declared tag with zero operations as an **empty collapsible section** on `/docs`;
- the `- **Aurora Proxy**: Interact with Aurora Climbing platform` bullet in `info.description`.

Verified against the generated document: **16 paths, 6 tags** (`Board Configuration`, `Climbs`, `Slug Resolution`, `Authentication`, `User Profile`, `WebSocket`) all with operations, no `/proxy/` path, no Aurora schema.

`docs.overview.rest.feature4` on the `/docs` page said "Aurora platform integration", which is false the moment the operations are deregistered. **Reworded, key kept**, in all four locales — the REST bullet list stays at four items, matching the WebSocket list beside it:

| Locale | New value |
|---|---|
| `en-US` | `Resolving URL slugs to numeric IDs` |
| `es` | `Resolución de slugs de URL a identificadores` |
| `fr` | `Résolution des slugs d'URL en identifiants` |
| `de` | `Auflösung von URL-Slugs zu IDs` |

The three `Slug Resolution` operations take `{board_name}` as a path *parameter* and return layout / size / set IDs, so "numeric IDs" — which is what `generate-openapi.ts:46`'s own `info.description` bullet says, and what the three translations already said — is the accurate wording. Slug resolution is a real, still-registered tag with three operations and no bullet, so the slot points at something true rather than being padded. Glossary-checked: no board noun arises in any of the four feature strings, nothing touches *enchaîné* / *croix* / *allumer*, German stays informal. "slug" is left untranslated in every locale — it is the URL contract's own term, used verbatim in `docs/branch-deploys.md`.

The new `docs.overview.retired.*` strings were glossary-checked too: `es` uses **plafón**, `fr` uses **la croix** for the send noun (`l'enregistrement des croix`) and never *envoyer*, `de` uses **Begehungen** and neuter **Board**.

**`feature2` ("User registration and authentication") checked and left alone**: the `Authentication` tag still has three registered operations (`/api/auth/register`, `/api/auth/verify-email`, `/api/auth/resend-verification`), so the bullet is still true. The four other Aurora strings in `marketing.json` (`help.aurora.title`, `privacy.thirdParty.aurora*`, `legal.*`) describe the **Aurora account sync that still exists** and are untouched.

### Facts the `saveClimb` delete falsified, repaired here

- `ClimbSearchInvalidationSource` loses `'save-climb-proxy'` — `saveClimb/route.ts:62` was its only producer. `/api/internal/climb-search-cache/revalidate` passes `'internal-route'` and is unaffected; the test fixture moved with it.
- `packages/db/src/queries/climb-stats/real-catalog-data.ts`'s load-bearing doc-comment named two `display_difficulty` writers; one of them was the `saveClimb` proxy. Reworded to one writer (`saveMoonBoardClimb`), with the removal noted.
- `docs/branch-deploys.md`'s "Batch 3: Aurora API Proxies" table now carries a one-line retirement note above it; the table stays as a record of the GraphQL migration.
- `packages/backend/MIGRATION_PLAN.md` said "Aurora proxy routes stay in Next.js" in **three** places — the §Decisions exclusion bullet (line 11), the Phase 2 scoping sentence (line 50) and the §2.5 "staying in Next.js" table row (line 116). All three now marked superseded / GONE.
- `docs/web-reposition.md` verdicted all five proxies `delete (dead code)` and listed the `openapi-routes.ts` deregistration as a loose end still to do. Split into "3 deleted, 2 → 410", with the loose ends marked discharged and a note that W-25b reads the counter before deleting. (The plan said to leave this file to W-26; two review lenses independently flagged it as the doc most likely to make W-25b delete the URLs *without* reading the counter, which is the one outcome this PR exists to prevent. The edit is two paragraphs and leaves W-26's sweep intact.)

**`app/lib/climb-search-cache.ts` survives**, verified rather than assumed: the kept MoonBoard bulk importer (`components/moonboard-import/moonboard-bulk-import.tsx:40`) and `db/queries/climbs/search-climbs.ts` still import it, and the kept `/api/internal/climb-search-cache/revalidate` route still imports the `.server` half. A test pins all of them. (The MoonBoard bulk importer never went through the `saveClimb` proxy anyway — it issues `SAVE_MOONBOARD_CLIMB_MUTATION` against the GraphQL backend.)

### Validation deleted from a dead path, not from the product

The diff removes the Zod `saveAscentSchema` and its two hard-won invariants — the Aurora-scale `quality: 0..3` bound and the UUID regex that stopped arbitrary strings becoming `boardsesh_ticks.uuid` values. **Both live independently on the GraphQL `SaveTickInput` path that actually serves the product.** This removes a *duplicate* guard on a dead path, not the guard. Flagging it because a reviewer reading only the diff will see validation being deleted.

---

## Three-lens QA pass — what changed and what I pushed back on

27 findings across a deprecation-contract lens, an SEO/docs lens and a test-strength lens. Fixed 20 (several were the same defect seen from two or three angles); rebutted 3 with evidence rather than churning the diff.

### Fixed

| Finding | Fix |
|---|---|
| `MIGRATION_PLAN.md` table unformatted → **PR CI green, push-to-main `vp check` RED** (3 lenses, MAJOR). PR lint is scoped to `*.ts\|tsx\|js\|mjs\|cjs`; markdown is invisible to it, and the pre-commit hook has the same scope. | `vp check --fix` on the file. `vp check` now reaches lint (see gates below). This is the "green PR reds main: full-repo lint" pattern from campaign memory, caught before it landed. |
| The `.catch(() => {})` guard was **inert** — deleting it left all 16 route tests green (2 lenses, MAJOR/MINOR). | Assert `.catch` is actually attached, by handing `track` a `{ catch: spy }` return. **The reviewers' own proposed oracle does not work**: a `process.on('unhandledRejection')` watcher stays green, because vitest's settled-result tracking on a `vi.fn` attaches its own handler to the returned promise and hides the rejection. I probed that before replacing it. Also wrapped the call in `try`/`catch` for the synchronous-throw path, with a test that mocks `track` to throw. Both mutations now red. |
| `Deprecation: @1786924800` was a **future** date (2026-08-17) on an already-410 response, and nothing pinned the constant — moving it to 2019 or 2026-09-20 kept the suite green (4 findings across 3 lenses). | Constant moved to `2026-08-15T00:00:00Z`; pinned to the literal, bounded to `<= Date.now()`, and the emitted header pinned to `'@1786752000'`. Docstring no longer says "that is now" above a hardcoded guess. `docs/branch-deploys.md`'s "Retired" line moved with it. |
| The counter could not distinguish a crawler `GET` from a live integration's `POST` (2 lenses, MAJOR/MINOR) — the number W-25b reads could not answer W-25b's question. | Added `method`. Reasoning and the ruling departure are written up above. |
| `{ headers: request.headers }` forwarded the caller's **`Cookie`** — the NextAuth session token on `saveAscent` — to the analytics ingestion endpoint (MINOR). | Allowlist of `user-agent` + `x-forwarded-for`. Confirmed against the installed `@vercel/analytics@2.0.1` source: `cookie: tmp.cookie` is copied verbatim, and a plain object satisfies its `hasHeaders` check with `cookie` undefined. |
| `boardName` emitted with caller casing; the middleware board check is case-insensitive, so `/api/v1/Kilter/...` splits the counter (NIT). | `.toLowerCase()`, pinned by test. |
| Only 2 of 3 stale "Aurora proxy routes stay in Next.js" claims in `MIGRATION_PLAN.md` were repaired; line 50 — the scoping sentence, the load-bearing one — was skipped (MINOR). | Fixed, and the PR body's "two places" corrected to three. |
| `Link: rel="deprecation"` and the body's `documentation` field pointed at a `/docs` that, post-deregistration, says nothing about the retirement (MAJOR). | Added the "Retired endpoints" card and pointed the URL at its anchor. Took the reviewer's option (b), not option (a) ("re-register both operations with `deprecated: true`"): the ruling bound deregistration, and re-publishing two operations whose request schemas describe an Aurora-credential POST body that no longer does anything would re-advertise a dead contract to fix a broken link. A test asserts the fragment and the card's i18n key exist in `docs-client.tsx`, so deleting the card reds. |
| Neither the `Link` header value, the response body, nor `DEPRECATION_DOCS_URL` was asserted — repointing it at `https://evil.example.com/` kept 16/16 green (MINOR). | Full header string, full JSON body and the constant all pinned. |
| `DEPRECATION_DOCS_URL` hardcoded the canonical host instead of using `absoluteUrl` (NIT). | Now `absoluteUrl('/docs#retired-endpoints')`. |
| The "exposes no Aurora proxy component schema" case was **inert** — restoring the three deleted schema exports without their `registerPath` blocks left all 6 cases green, because zod-to-openapi only emits a component for a schema a registered path references (2 lenses, NIT/MINOR). | Assert on the registry module's export surface, which is where a half-revert actually lands. Probed: restoring `openapi-registry.ts` from `main` now reds it. |
| `/docs` sitemap `lastModified` stale at 2026-04-30 while the page's content changed (2 lenses, MAJOR/MINOR). | Bumped to 2026-08-15 with a comment naming W-25a, matching the `/help` precedent from W-16 in the same file. |
| `docs/web-reposition.md` still verdicts all five proxies `delete (dead code)` and lists discharged loose ends (2 lenses, MAJOR/MINOR). | Split the verdict, marked the loose ends done, named the counter W-25b must read. |
| en-US `feature4` said "board IDs"; the operations resolve layout/size/set IDs (2 lenses, NIT). | `Resolving URL slugs to numeric IDs`. |
| The `publish` (mobile OTA preview) check failed on an S3 `SlowDown` rate-limit, unrelated to this diff (NIT). | Re-triggered by this push. |
| **`claude-review` bot:** the "Retired endpoints" card's `{ sunset: '2026-10-01' }` was not pinned to `AURORA_PROXY_SUNSET_DATE`. | Real. The card cannot import the constant — `api-deprecation.ts` is server-only and `docs-client.tsx` is a client component — so the two are pinned together by test instead. Probed: slipping the constant to 2026-11-01 now reds. |
| **`claude-review` bot:** `api-deprecation.ts` had no `server-only` guard. | Real, added. It reaches `analytics.server`, so a client-side import should fail at the boundary rather than deep in a transitive import. |

### Rebutted

1. **"`boardName` gives any unauthenticated caller unbounded control of a custom-event property value"** (test-strength, MINOR — the `/api/v1/aaaa/proxy/login`, `/api/v1/aaab/proxy/login`, … cardinality-blowup scenario). `packages/web/middleware.ts:142-161` returns a bare 404 for any `/api/v1/<segment>/…` whose lowercased segment is not in `SUPPORTED_BOARDS`, **before** the handler runs, and `middleware.ts:263-267` matches `/api/v1/:path*`. The spray never reaches the counter, so the property is bounded to the handful of supported board names. The one real half of the finding — casing variants of a *valid* board name, which the middleware does let through — is fixed by `.toLowerCase()`. I deliberately did not add an `isAuroraBoardName` re-validation on top: it would be unreachable behind the middleware, and it would silently relabel any board we later add to `SUPPORTED_BOARDS` as `'other'` in the one dataset W-25b depends on.

2. **"Add the 'board' qualifier to the de/es/fr `feature4` strings"** (test-strength, NIT) — the premise is inverted. The three registered `Slug Resolution` operations are `GET /api/v1/{board_name}/slugs/{layout,size,sets}/…`: the board is an *input* path parameter and the outputs are layout, size and set IDs. Adding "Board-IDs" / "identificadores de plafón" would propagate the error rather than fix it. The two other lenses reached the opposite conclusion on the same string, and `generate-openapi.ts:46` settles it: "**Slug Resolution**: Convert human-readable URLs to numeric IDs". en-US was the outlier and is the one that moved.

3. **"Drop the ❌ emoji from the `MIGRATION_PLAN.md` §2.5 row"** (deprecation-contract, part of a MAJOR whose formatting half is fixed). §2.4 of that same file already uses `❌ GONE` for the two W-19 rows, so the marker is the file's own convention, not something new this PR introduced; dropping it in §2.5 would leave two inconsistent GONE markers a few lines apart. The prettier re-pad the finding is actually about is fixed.

### Standing rules

- **Locale-twin redirects — N/A.** This PR adds no redirect. The three deleted routes 404 (undocumented API paths, no external contract, no SEO value); the two deprecated ones answer 410 in place. Nothing goes through `expandLocaleRedirects`.
- **`Disallow`-with-redirect in the same commit — N/A.** `app/robots.ts` is untouched; `/api/` was and stays disallowed. (`/openapi.json` is not under `/api/` and stays crawlable, which is exactly why this is a deprecation.)
- i18n catalog parity: `feature4` value changed and `docs.overview.retired.{title,body}` added in all four locales, same commit.
- No `any`; no raw SQL added. `docs-client.tsx` gains one MUI `Alert` using existing imports and no `style` prop; all its copy goes through `t()`.
- `import-graph-allowlist.json` still `{"entries": []}`; `KEPT_ROUTE_DIRS` unaffected (`api` and `docs` both stay); `crawler-classic-invariant.test.ts` and `aa-embed-headers.spec.ts` untouched.
- Rebased onto W-19 (#4479) after it merged. W-19 only *reads* `openapi-routes.ts`; no conflict. Removing 85 lines from that file shifts the `/api/internal/profile` registrations up by 85 lines, so any W-19 artefact quoting line 567/594/630 is now stale text — not a broken test.

## Release Notes

These endpoints have had no in-repo caller since the GraphQL migration and nothing climber-visible changes. If the new counter shows real third-party traffic, W-25b writes the climber-voice note instead.

## Test plan

Five test files, 47 assertions. Every one mutation-probed — **25 probes, all red, all restored, `git status --porcelain` clean afterwards.**

The 15 probes from the first pass (headers, status, verbs, endpoint labelling, the OpenAPI deregistration, the tag/description coherence, the three deletes, and the `climb-search-cache` overreach guard) all still red. The ten added for the QA and review-bot fixes:

| # | Mutation | Result |
|---|---|---|
| 16 | Delete `.catch(() => {})` from the helper | RED — "attaches a rejection handler to the telemetry promise" (1 failed \| 24 passed) |
| 17 | Remove the `try`/`catch` around the counter | RED — "still answers 410 when telemetry throws synchronously" (1 failed \| 24 passed) |
| 18 | Drop `method` from the property bag | RED — 4 failed \| 21 passed, across all three test files |
| 19 | Drop `boardName.toLowerCase()` | RED — "lower-cases the board name…" (1 failed \| 24 passed) |
| 20 | Hand `request.headers` back to `track` (re-widen the cookie) | RED — 3 failed \| 22 passed, incl. "never forwards the caller cookie to the analytics endpoint" |
| 21 | Repoint `DEPRECATION_DOCS_URL` at the bare `/docs` | RED — "sends every caller to a page that actually documents the retirement" |
| 22 | Slip `AURORA_PROXY_DEPRECATION_DATE` to 2026-09-20 (future, still < `Sunset`) | RED — "pins the Deprecation date this PR records, and never advertises a future one" |
| 23 | Restore the three deleted schemas to `openapi-registry.ts`, `registerPath` blocks still gone | RED — "keeps no Aurora proxy schema, in the document or in the registry module" (1 failed \| 5 passed) |
| 24 | Delete the "Retired endpoints" card from `docs-client.tsx` | RED — "links to a fragment that exists on /docs" (1 failed \| 9 passed) |
| 25 | Slip `AURORA_PROXY_SUNSET_DATE` to 2026-11-01, leaving the `/docs` card at 2026-10-01 | RED — 2 failed \| 8 passed: the header pin **and** the `/docs` prose pin |

Two probes were run and **discarded as inert** before being replaced, rather than shipped: a `process.on('unhandledRejection')` watcher for probe 16 (vitest's `vi.fn` settled-result tracking attaches its own handler and hides the rejection — verified by deleting `.catch` and watching it stay green at 24/24), and the document-level schema assertion for probe 23.

Gates run, real output:

- `vp run typecheck` — **green** (full monorepo).
- `vp test run --reporter=agent --project web` — **301 files, 3356 tests.** One 5000 ms timeout in a `packages/web/app/components/admin/` panel test under parallel load, in a *different* file on each run (`location-sync-freezes-panel`, then `gym-settings-panel`); each passes in isolation (5/5 and 6/6) and one full run came back 3356/3356 clean. Pre-existing load-dependent flake, not this PR — and CI agrees: `test-default (1,2)` and `(2,2)` are both green.
- `vp test run --reporter=agent --project i18n` — **green**, 90 tests (catalog parity across all four locales).
- `vp run check:i18n` — **green**, 302 `.tsx` files, no hardcoded user-facing strings.
- `vp run check:i18n:orphans` — **green**, 3678 catalog keys, every one referenced.
- `vp run check:db-migrations` — **green**, 200 migrations, journal consistent.
- `vp check` — **format green** (it now reaches lint; before the `MIGRATION_PLAN.md` fix it short-circuited at the format stage). Lint/type: **33 errors, 325 warnings in 4287 files — the current `main` baseline, all pre-existing.** I dumped the full output and extracted every flagged file path (118 files: `scripts/`, `packages/db`, `packages/backend`, `packages/mobile`, `packages/shared`, and 13 unrelated `packages/web` files). **None of this PR's files is in that set** — not `app/lib/api-deprecation.ts`, `app/lib/__tests__/api-deprecation.test.ts`, `app/docs/docs-client.tsx`, `app/lib/api-docs/__tests__/openapi-document.test.ts`, `app/lib/seo/sitemap/static-entries.ts`, nor either proxy route test. Not suppressed, not worked around.

Refs #4358 · feeds #4443 (W-25b) and #1889.


marcodejongh:
Orchestrator note. Marked ready — CI green, no unresolved review comments, no conflicts. Three things from this round worth making visible.

## My telemetry constraint was incomplete, and the lens caught what it missed

My ruling said: emit `endpoint` and `boardName` **only**, never the request body, because `login` carries Aurora credentials and `saveAscent` carries a session token.

The first implementation passed `{ headers: request.headers }` to the tracker. That is not the request body, so it satisfied the letter of the constraint — and it forwarded **the caller's `Cookie`, including the NextAuth session token on `saveAscent`, to Vercel's analytics ingestion endpoint.** The lens verified it against the installed `@vercel/analytics@2.0.1` source rather than inferring it: `cookie: tmp.cookie` is copied verbatim onto the outbound fetch, and a plain object satisfies its `hasHeaders` check. Now a `user-agent` + `x-forwarded-for` allowlist, with a probe that reds if the allowlist is widened.

The lesson for the next telemetry constraint I write: **"never the body" is the wrong shape.** The rule is an allowlist of named fields, because credentials travel in headers at least as often as in bodies.

## The fixer caught its own oracle being inert

Both reviewers proposed a `process.on('unhandledRejection')` test for the `.catch(() => {})` guard. The fixer shipped it, probed it by deleting the `.catch`, watched it stay **green 24/24**, and replaced it — vitest's settled-result tracking on a `vi.fn` attaches its own handler and hides the rejection. The working oracle watches the `.catch` call itself.

That is the campaign's own standard applied self-critically, and it is the reason a lens suggestion is a hypothesis rather than an instruction. Recording it here because the same `unhandledRejection` idea is likely to come up again — it does not work under this test runner.

## One deliberate departure from my ruling, and I agree with it

`method` was added to the property bag, beyond the `endpoint` + `boardName` I specified. My constraint was about credential safety; a verb is neither body data nor a credential, and without it W-25b cannot tell a crawler's `GET` from a live integration's `POST` — which is the entire question the counter exists to answer.

**W-25b's sign-off must read the `POST` count, not the total.** That is recorded on #4443.

## Also worth noting

`packages/backend/MIGRATION_PLAN.md` was not prettier-formatted, which made `vp check` short-circuit at the format stage before it ever reached lint. PR lint is scoped to `*.ts|tsx|js|mjs|cjs`, so markdown is invisible to it — the same "green PR reds main" shape as #4488. Cheap to trip, cheap to miss.

And the `Deprecation:` header was a **future** date on an already-410 response, which is incoherent on its face. Now `2026-08-15T00:00:00Z`, pinned three ways including a `<= Date.now()` bound so it cannot drift forward again.

PR #4467, opened 2026-08-15
chore(web): delete the climbing UI and swap the root chrome

marcodejongh:
## Summary

Part of #4358.
Closes #4435.

The centrepiece of the web reposition, and **the irreversible one**. Climbing moves to the app; www becomes the front door that keeps every link working.

**Real diff stat: 677 files changed, 1,725 insertions, 131,350 deletions — net −129,625 lines** (580 deleted, 83 modified, 7 renamed, 7 added). The epic row estimated −110k. The gap is second-order orphanage the row does not name: the 338-line `persistent-session-wrapper` and its 402-line test, the `climb-detail` tree, 22 `app/lib` modules, 19 orphaned hooks, four `social/*-search-results`, the `lib/ble` adapters, `public/help/`, and the i18n prune.

### Rollback is `release/classic-web`, never `git revert`

Reverting means reintroducing ~130k lines, re-mounting a provider tree, rebasing over everything merged since, and passing `check:i18n:orphans` under incident pressure. The restore procedure is the W-14 section of `docs/web-reposition.md`:

```
RESTORE=$(git commit-tree origin/release/classic-web^{tree} -p origin/main \
  -m 'revert: restore the classic interactive web app (rollback via release/classic-web)')
```

then `git diff --stat restore/classic-web origin/release/classic-web` must print nothing. PR #4427 (`preview/classic-web-rollback`) is the standing verification route. **Nobody should reach for `git revert` at 2am.**

### The root chrome swap

`PersistentSessionWrapper` mounted the whole interactive climbing stack on **every** route, `/about` and `/legal` included: a `connectionName: 'session'` WebSocket, the queue bridge, the BLE provider, the queue control bar, the bottom tab bar, and the `ResizeObserver` that published `--bottom-bar-height-measured`.

`app/components/providers/site-chrome.tsx` keeps three providers and nothing else:

| Provider | Why it stays |
| --- | --- |
| `StatsFilterBridgeProvider` | the `/profile` and `/you` statistics filter button |
| `ProfileHeaderShareProvider` | the viewed-profile share button |
| `PlaylistsAdapterProvider` | **required** — every hook in `@boardsesh/playlists-react` calls `usePlaylistsAdapter()` unconditionally, so `/playlists` throws without it |

`AuthModalProvider` stays exactly where it already was, at `app/layout.tsx`. It was never inside the wrapper, and moving it under `FeatureFlagsProvider` would reorder providers for no gain.

### Socket counts

| Surface | Before | After |
[...]

PR #4491, opened 2026-08-15
refactor(web): scope /settings down to ESP32 controllers and set-password

### LINKED ISSUE #4440: W-21: scope /settings down to ESP32 controllers and set-password
Part of #4358. **Blocked by:** #4437 (W-19). QA gate #4359 is **SIGNED and closed (2026-08-14)**, so the gate dependency is satisfied.

## Context

`/settings` took 1,206 views / 471 users over 90 days to 2026-07-31. **ESP32 is tiny but shipped hardware**: 28 `esp32_controllers` rows, 19 distinct owners, newest registration 2026-06-17, 6 created in 90 days, only 1 controller seen in 90 days, zero `%controller%`/`%esp32%` analytics events. `/settings` survives, scoped down. The epic is explicit that we are **not porting the ESP32 flow** to the SPA — it stays on Next.

## The work

- `/settings` survives as **ESP32-controllers + set-password**, dropping the three sections with SPA twins.
- **Set-password stays** because mobile has `resetPassword` but no *set*-password, so an **OAuth-only user has nowhere else to go**.
- **Do not touch `esp32_controllers` rows** (28 rows, 19 owners) — no data deletion without asking.

## Standing-rule callouts

- **No data deletion without asking.** This PR is UI scope-down only; the `esp32_controllers` table and its rows are untouched.
- **Every redirect gets `es`, `fr` and `de` siblings (standing rule 4)** — `next.config.mjs`'s `redirects()` entries are bare paths and there is no Next i18n routing, so `source: '/settings/x'` never matches `/es/settings/x`.
- **Cross-origin destinations stay `permanent: false`** until the SPA route is proven in production.
- **Remove a `Disallow` in the same commit as its redirect (standing rule 3).**
- **e2e spec deletions ship with this PR (standing rule 6)** — specs live in `packages/web/e2e/`, not `e2e/`; `e2e-tests.yml` is `workflow_dispatch`-only.
- `/settings` is a **utility surface** — keep `robots: { index: false, follow: true }`.
- Catalog keys for the dropped sections come out of **all four locales in the same commit**.

## Deps

- W-19 — #4437
- QA sign-off — #4359 (signed 2026-08-14)

## Gates

- **GI** — `vp run check:i18n` · `vp run check:i18n:orphans`

## Verifies

- [ ] `/settings` renders **only** the ESP32-controllers and set-password sections.
- [ ] An OAuth-only user can still set a password.
- [ ] Registering / listing / removing an ESP32 controller still works end to end, and **no `esp32_controllers` rows are deleted or migrated**.
- [ ] The dropped sections 30x (or are unlinked) in all four locales, with locale twins asserted.
- [ ] W-05's import-graph allowlist stays empty.

## Tiering

**Sonnet** implementation (mechanical, clear spec, cheap oracle). **A Sonnet implementation never reviews itself** — the QA lens is **Opus**, three adversarial agents each prompted to refute.

## Release Notes

User-visible: settings on the web keep your board hardware and password; everything else lives in the app. Climber voice.

marcodejongh:
## Summary

Closes #4440. Part of #4358 (web reposition), W-21. Rebased onto `5a4b2b8dd` (after W-19 #4479, W-24 #4480 and #4484).

`/settings` survives the reposition, scoped down to the two things the web can do that the app can't:

- **ESP32 controllers** — shipped hardware the epic deliberately keeps on Next; there is no SPA screen to send those owners to.
- **Set password** — mobile has `resetPassword` but no *set*-password, so an OAuth-only user has nowhere else to go.

Everything else on the page has an app twin and is gone. Per the orchestrator's ruling, that is **six blocks, not three**: the issue's "three sections with SPA twins" counted only the extracted components under `app/components/settings/`, but `settings-page-content.tsx` also rendered three inline cards with twins. The issue's own Verifies row — "`/settings` renders **only** the ESP32-controllers and set-password sections" — is the authority.

| Block | Fate | App twin |
| --- | --- | --- |
| Page header | keep | — |
| Profile card (avatar, display name, Instagram, email) | **drop** | `(tabs)/profile/edit.tsx` |
| Preferences card (grade format) | **drop** | `(tabs)/profile/more.tsx` (see the stranded-setter note below) |
| `SetPasswordSection` | **keep** | none — that is why it stays |
| Logbook shortcut card | **drop** | app logbook |
| `AuroraCredentialsSection` | **drop** (shared parts lifted) | `integrations/BoardAccountsSection.tsx` |
| `ControllersSection` | **keep** | none — ESP32 is Next-only by design |
| `WatchPairingSection` | **drop** | `(tabs)/profile/watch-pair.tsx` |
| `DeleteAccountSection` | **drop** | `(tabs)/profile/delete-account.tsx` |

A signed-in visitor following W-19's header cog would otherwise land on a password form, an ESP32 list and no hint of where the rest went, so the page ends with one hand-off card to the app's More screen (three new catalog keys × four locales). The destination is named explicitly — `buildAppHandoffUrl('/profile/more')`, **not** `buildAppHandoffUrl('/settings')`, because the SPA has no `/settings` route and the same-path form would emit a 404. The helper still does the origin half, which matters: it trims a trailing slash off `NEXT_PUBLIC_APP_URL`, and `next.config.mjs:60-63` records that local Expo-web work routinely exports it.

### Lifted, not deleted

`aurora-credentials-section.tsx` had two exports with a kept consumer. They moved to `app/components/board-entity/board-credential-card.tsx` (+ `.module.css`), next to `board-import-prompt.tsx`, the only importer — so the lift removes a cross-directory edge instead of relocating one. That prompt renders on `/aurora-migration` and `/profile/{user_id}`, both **indexable** surfaces the epic keeps.

`BoardCredentialCard` moved **verbatim** per the Q2 ruling. Two of its branches are now unreachable from `board-import-prompt` (the `kilterNew` variant, and the `totalUnsynced > 0` alert, which the prompt always calls with zero counts). Trimming them would cascade into `BoardCredentialCardVariant` and then into `aurora.card.kilterNew*` / `aurora.unsynced.*` — regression risk on two indexable pages, inside a PR whose job is deletion elsewhere. **Recorded as a known gap for W-26 (#4442).**

The one thing that did **not** move verbatim is `STEP_LABELS`, a hardcoded English copy of the six import-step labels. It had no production reader — `ImportProgressSteps` renders `getStepLabels(t)` — and existed only because the old test asserted against it. It is deleted, and the moved test now asserts the `aurora.import.steps.*` catalog values the component actually renders. The old assertion passed only because those six strings happened to equal the six en-US values verbatim; a copy edit would have red-ed it and the apparent fix would have been to edit a constant nothing rendered.

### One correction to the plan (caught during implementation)

The plan said the lifted CSS module could drop four classes, including `.unresolvedList`, on the grounds that it was only used inside the deleted default export. It is not: `board-import-prompt.tsx` binds it at **three** call sites (`:511`, `:534`, `:557` — the unresolved list in each import-results alert). A missing CSS-module class is silent — no type error, no test failure, just an unstyled overflow list on a public page. **Three classes were dropped, not four**: `.sectionDescription`, `.cardsContainer`, `.loadingContainer`. Every remaining class has a live binding in either `board-credential-card.tsx` or `board-import-prompt.tsx`.

### `/privacy` follows the delete button

The privacy policy documented `boardsesh.com/settings` as the web account-deletion route, in all four locales. `DeleteAccountSection` is gone from that page, so the policy's stated mechanism would have stopped existing — on an indexable page, in legally load-bearing copy. `privacy.deletion.onWeb*` now points at `app.boardsesh.com` and `privacy.deletion.inAppBody` names the More screen, which is where `(tabs)/profile/delete-account` is actually reached from. The `href` at `privacy/page.tsx` goes through the same `buildAppHandoffUrl('/profile/more')`.

### Stranded code removed

- `app/lib/data-sync/moonboard/csv-import-stream.ts` — the deleted section was its only importer repo-wide. The directory goes with it. **Not** to be confused with `app/moonboard-import/` + `app/components/moonboard-import/`, the OCR screenshot importer, which is alive and untouched.
- `app/lib/aurora-credentials/client.ts` — `getAuroraUnsyncedCounts`, `finalizeKilterCredential`, `saveKilterCredentialViaPassword`, `type UnsyncedCounts`. `AuroraBackendError` **stays** (`requestBackendJson` throws it internally).
- `app/flags.ts` — `GARMIN_WATCH_FLAG` and `'kilter-oauth-linking'`, the two keys this PR strands. Note that `FEATURE_FLAG_KEYS` feeds the live `/.well-known/vercel/flags` discovery endpoint, so this is a deliberate public-facing change. The mobile watch-pair row is unaffected — it reads the flag through mobile's own `FeatureFlagsProvider`.
- `STEP_LABELS` in the lifted `board-credential-card.tsx` (above).

**Newly stranded, deliberately left, added to the W-26 (#4442) list: the grade-format writer.** Dropping the preferences card removes `setGradeFormat`'s only production caller on www. `useGradeFormat` is still *read* by seven kept surfaces (`climb-card/climb-title`, `activity-feed/session-feed-card`, `charts/climb-analytics`, `similar-climbs/similar-climbs-list`, `social/user-smart-card`, `profile/[user_id]/profile-page-content`, `session/[sessionId]/session-overview-panel`), so `setGradeFormat` / `GRADE_FORMAT_CHANGE_EVENT` / `setGradeDisplayFormat` are now writer-less rather than dead: www renders every grade at the `'v-grade'` default and has no control to change it. The preference lives in per-origin IndexedDB with no server mirror, so setting it in the app cannot reach www. Deleting the setter now would churn eight test files' mocks inside a deletion PR; keeping the toggle would contradict the ruling and the issue's Verifies row. So it is stated instead of implied — **and "grade format" is struck from the hand-off copy and the Release Notes**, because pointing www's grade preference at a different origin would be a false promise.

Left alone on purpose, all noted for W-26 (#4442): `BOARDSESH_GRADE_FLAG` (already consumer-less on web before this PR — clean up what you strand, don't adopt someone else's mess), `createKilterHandoffStartUrl` (dead on `main` already), the `@boardsesh/watch-pairing` entry in `packages/web/package.json` (removing it churns `bun.lock` with other trains in flight, and an unused manifest entry is not dead *code*), the `openapi-routes.ts` `post`-vs-`PUT` mismatch on `/api/internal/profile`, and `docs/branch-deploys.md:295`'s reference to a path that has never existed.

`/api/internal/profile` is **untouched**, GET and PUT. GET still feeds the scoped-down page; PUT loses its web caller but is a published API surface this epic keeps.

## The issue's boilerplate callouts — why four of them are vacuous here

Stated explicitly rather than left unticked:

- **"Every redirect gets `es`, `fr`, `de` siblings"** — n/a. **W-21 deletes no route.** `/settings` survives, `app/settings/page.tsx` and `settings-page-content.tsx` both stay, and no `/settings` rule is added to `next.config.mjs`. There was none before either. `settings-scope-down.test.ts` pins that, matching sources the way Next matches them rather than by string equality (see the test-plan note).
- **"Cross-origin destinations stay `permanent: false`"** — n/a, no redirect added.
- **"Remove a `Disallow` in the same commit as its redirect"** — n/a, and the opposite applies: `/settings` **stays** in `robots.ts`'s disallow list because it stays a utility surface. `robots.ts` is untouched, and the new test asserts `/settings` is still disallowed (cheap insurance that W-19's rewrite of that array did not take it along with `/feed` and `/you`).
- **"e2e spec deletions ship with this PR"** — n/a. `packages/web/e2e/` has exactly one `/settings` reference (`site-chrome.spec.ts`, a "does the chrome render" loop) and it keeps passing. No spec is deleted or edited.
- **"`/settings` keeps `robots: { index: false, follow: true }`"** — yes. `app/settings/page.tsx` is byte-identical; `page-seo-metadata.test.ts` pins its `robots` and `canonical`.
- **"Catalog keys come out of all four locales in the same commit"** — yes, see below.
- **"Do not touch `esp32_controllers` rows"** — nothing in this PR reaches the database. No migration, no SQL, no Drizzle call. `app/api/internal/controllers/route.ts` is untouched and `settings-scope-down.test.ts` asserts it still exists.
- **"W-05's import-graph allowlist stays empty"** — `import-graph-allowlist.json` is still `{"entries": []}`.

## i18n

28 key paths (29 entries counting both plural forms of `aurora.import.results.unresolvedTitle`) removed from **all four** locales, and `appHandoff.{title,subtitle,cta}` added to all four. `preferences.*` and `logbook.*` are gone entirely; `profile.*`, `deleteAccount.*`, `watchPairing.*` and `aurora.*` lose only the leaves the web owned. Five `marketing.privacy.deletion.*` values are rewritten in all four locales (no key added or removed there).

Nothing in CI catches **over**-removal — `check:i18n:orphans` only catches over-retention, and a key deleted while mobile still calls it just renders as its own dotted path on someone's phone. So the removal set was derived, not assumed: every leaf in `en-US/settings.json` was plural-normalised and matched as a quoted literal against every `.ts`/`.tsx` under `packages/web/{app,e2e}` and `packages/mobile/{src,app}` **after** the deletions landed. That scan returned exactly 28 unreferenced paths, matching the planned list key-for-key. There are no dynamic `t()` lookups and no `i18n-keep` markers in the `settings` namespace.

The `settings` namespace stays in `ALL_NAMESPACES` — it is majority mobile-owned (`ble.*`, `integrations.*`, `devicePicker.*`, `feedbackForm.*`, most of `profile.*` / `deleteAccount.*` / `watchPairing.*` / `aurora.mobile.*`).

Two catalog bugs fixed for free by the delete: `aurora.import.interrupted` and `aurora.import.interruptedShort` were called from the deleted region but exist in **none** of the four catalogs.

The hand-off subtitle names the destination the way the destination names itself — "connected apps" / "aplicaciones conectadas" / "applications connectées" / "verbundene Apps", matching `mobile.more.integrations.title` on the screen it opens. Glossary compliance: French avoids « envoyer »; German is informal *du*. No board noun arises in Spanish, so *plafón* does not apply (an earlier draft introduced one anyway and it is gone).

## Traffic

Re-pulled from PostHog project 412845 on 2026-08-15, 90 days, `properties.$host = 'www.boardsesh.com'`, locale-prefix-aware:

| surface | views | users | issue said (to 2026-07-31) |
| --- | --- | --- | --- |
| `/settings*` | **1,073** | **434** | 1,206 / 471 |

```sql
SELECT multiIf(match(properties.$pathname, '^(/(es|fr|de))?/settings($|/|\\?)'), '/settings', 'other') AS surface,
  count() AS views, count(distinct person_id) AS users
FROM events
WHERE event = '$pageview' AND timestamp >= now() - INTERVAL 90 DAY
  AND properties.$host = 'www.boardsesh.com'
GROUP BY surface ORDER BY views DESC
```

Mild decline, same order of magnitude. Nothing here depends on the number — `/settings` survives either way.

**The `esp32_controllers` row counts (28 rows / 19 owners / 6 in 90 days) were NOT re-verified** — there is no prod `DATABASE_URL` in this checkout. They are also not premise-bearing: "keep ESP32 on Next" is a locked decision on #4358, and this PR touches neither the table nor its route.

**Flag reality check, worth knowing beyond this PR:** the `garmin-watch` PostHog flag **has never existed**. The live project has 13 flags and that key is not among them, yet it sat in web's `FEATURE_FLAG_KEYS` and still sits in mobile's `FEATURE_FLAG_DEFINITIONS`, so mobile's `more.tsx:98` watch-pair row resolves `undefined` (off) for everyone regardless of anything here — same shape as the `gym-kiosk` flag incident. That is why the hand-off copy does not name watch pairing. `kilter-oauth-linking` by contrast is a real ACTIVE flag (id 737576) still read by mobile's `BoardAccountsSection`, so removing web's discovery key is safe.

## Review round — what changed, and what did not

Three adversarial lenses raised 17 findings. Fixed:

- **`/privacy` still documented the deleted Delete Account button** (4 locales + the `href`). See above — the one MAJOR that would have shipped a broken legal promise.
- **Hand-off copy named watch pairing and grade format.** Watch pairing is behind a flag that does not exist; grade format has no www writer any more. Both struck from `appHandoff.subtitle` (4 locales) and from the Release Notes.
- **The subtitle called the destination "board accounts"** while the screen calls itself "Connected apps" — worst in Spanish, where it read "cuentas de plafón" against a sibling key saying "Cuentas de Aurora". Now matched to the destination's own label in all four locales.
- **The stranded grade-format setter** was not in the leftovers list. It is now, named explicitly (`setGradeFormat` + `GRADE_FORMAT_CHANGE_EVENT` + `setGradeDisplayFormat`) with what it costs www.
- **`APP_SETTINGS_URL` concatenated onto raw `APP_URL`**, skipping the trailing-slash trim every other www→app CTA gets. Now `buildAppHandoffUrl('/profile/more')`; the test asserts a hand-written `https://app.boardsesh.com/profile/more` instead of rebuilding the expectation the same way as the code.
- **The page test had no "and nothing else" bound on links.** It now asserts the page's whole href list (`['/', <hand-off>]`), so a re-added shortcut card pointing anywhere reds it — probed with an `/notifications` anchor.
- **The logbook guard was pinned to an unprefixed href** and would have gone inert under a locale-prefixed mock. Now `a[href$="/playlists"]`.
- **The redirect guard was inert against wildcards** — the shape a later sweep would actually write. It now compiles each `source` through a small path-to-regexp-shaped matcher and asks whether the rule swallows `/settings` (and the three locale twins). Probed with both `/settings/:path*` and `/settings:rest(.*)`; both red it now, neither did before.
- **`ControllersSection` had no render coverage anywhere in the repo** — the single headline reason `/settings` survives this epic. New `components/account/__tests__/controllers-section.test.tsx`, 7 tests: empty state, list with status chips, register dialog, the "don't POST a half-filled tuple" guard, register → API key shown once, delete through the confirm popover, and a server-side delete failure. The comment claiming both stubbed sections already had unit tests is corrected (it was true of `set-password-section` only).
- **`STEP_LABELS`** — deleted, test re-pointed at the catalog (above).
- **PR body said `.unresolvedList` had four call sites**; it has three. Corrected above.
- **The hand-off button** carried an `OpenInNewOutlined` glyph while navigating in the same tab, and skipped the `component="a"` both sibling CTAs use. Glyph dropped, `component="a"` added.

Rebutted, with the evidence, rather than churning the diff:

- **"`docs/ui/18-settings.md` must be trimmed to the surviving web screen."** `docs/ui/` is not live documentation — `docs/ui/00-overview.md` opens with "the single source of truth for **agents building the React Native app**", i.e. a frozen snapshot of the web app the RN rewrite was specified from. Trimming section 11 to the two surviving web cards would delete the spec for the mobile profile / grade-format / Aurora-linking / delete-account screens, which all still exist. The precedent is consistent: W-16 (#4467) deleted the play view and left `06-play-view.md`; W-19 (#4479) deleted `/feed` and left `16-feed.md`. W-26 (#4442) is the epic's doc-debt PR by its own description. The four docs this PR *does* fix (`kilter-sync.md`, `integrations.md`, `aurora-sync.md`, `POSTHOG_INVENTORY.md`) are live docs whose statements this PR makes false.
- **"Capture a click event on the hand-off CTA so the card's value is measurable."** New instrumentation is not in scope for a deletion PR, and it would not match the pattern it was compared against: neither `start-climbing-button.tsx` nor `playlists/library-page-content.tsx` — the two CTAs this button was aligned to — captures anything (`grep -c capture` → 0 on both). Only `climb-front-door/climb-handoff-cta.tsx` does, and it is a different surface with a modified-click guard. Worth filing separately if the epic wants hand-off telemetry; not worth inventing a one-off here.

## Test plan

New tests, each with a named mutation probe that was actually run:

| Test | Probe | Result |
| --- | --- | --- |
| `app/settings/__tests__/settings-page-content.test.tsx` | re-add a `TextField` | `renders no profile form fields` → RED |
| ″ | re-add the `ToggleButtonGroup` | `renders no grade-format toggle` → RED |
| ″ | re-add the `/playlists` card | `renders no logbook shortcut` → RED |
| ″ | render `t('deleteAccount.title')` | `renders no delete-account section` → RED |
| ″ | add `<a href="/notifications">` to the page body | `links to the app profile screen, and links nowhere else` → RED |
| `app/components/account/__tests__/controllers-section.test.tsx` | four combined mutations (delete body key, always-empty list, `setIds` payload, blank API key) | 4 of 7 → RED |
| `app/__tests__/settings-scope-down.test.ts` | re-create `app/components/settings/x.ts` | `leaves no trace…` → RED |
| ″ | move `board-credential-card.tsx` away | `keeps every surface the scope-down spared` → RED |
| ″ | add `/settings/:path*` to `BASE_REDIRECTS` | `gives /settings no redirect in any locale` → RED |
| ″ | add `/settings:rest(.*)` to `BASE_REDIRECTS` | ″ → RED |
| ″ | drop `/settings` from `robots.ts` | `keeps /settings disallowed for crawlers` → RED |
| `board-entity/__tests__/board-credential-card.test.tsx` | `syncStatus === 'expired'` → `'error'` | `shows Reconnect before Unlink` → RED |
| ″ | `isExpired` → `credential != null` | `offers no Reconnect on an active credential` → RED |
| ″ | print `credential.syncError` raw | `localises the duplicate-circuits sync error` → RED |
| `board-entity/__tests__/import-progress-steps.test.tsx` (moved) | point `steps.resolving` at the `dedup` key | `renders all step labels` → RED |
| ″ | `index < currentStepIndex` → `<=` | 2 tests → RED |
| `app/__tests__/import-graph-invariant.test.ts` | import a revived `components/settings/` module from `app/settings/` | `no unallowed edge from a kept surface into the delete set` → RED |

**Probe note:** the planned probe for the moved `import-progress-steps` test — reordering `STEP_ORDER` — does **not** red it. That test derives its expectations from `STEP_ORDER` itself, so it is order-invariant by construction. Mutating the completion predicate reds it, which is the property that actually matters. Worth knowing before someone relies on that test to pin step order.

All probes restored exactly; `git status --porcelain` is clean.

Gates:

| Command | Result |
| --- | --- |
| `vp run typecheck` | green (full monorepo) |
| `vp test run --reporter=agent --project web` | green |
| `vp run check:i18n` | green — 300 `.tsx` scanned, no hardcoded strings |
| `vp run check:i18n:orphans` | green — 3650 keys, every one referenced |
| `vp test run --reporter=agent --project i18n` | green — 90 tests (four-locale parity) |
| `vp check` | errors/warnings all pre-existing; none in a file this PR touches |

`vp check` is red on `main` too. None of the flagged files are touched here — the flagged set is `packages/backend/src/__tests__/*`, `packages/db/scripts/*`, `packages/mobile/src/{providers,lib}/*`, `scripts/*` and friends.

Earlier full-suite runs showed 2–4 unrelated `Test timed out in 5000ms` failures in files this PR does not touch (`admin/location-sync-freezes-panel`, `gym-entity/report-duplicate-dialog`, `gym-entity/gym-form-create-anyway`), a different set each run, all passing in isolation. Machine-load flakes, not regressions.

Not run: `vp run test:e2e` — not in this issue's gate list, `e2e-tests.yml` is `workflow_dispatch`-only, and no spec changed.

## Product hole, stated out loud

Two, both deliberate:

1. **Grade format has no www control any more** (see the stranded-setter note). Every grade on www renders at the `'v-grade'` default until W-26 or a follow-up decides whether www keeps a preference at all.
2. Everything else a signed-in visitor came to `/settings` for now has a labelled route to `app.boardsesh.com/profile/more`, which is where `delete-account`, `edit` and the connected-apps screen all live. That was the whole reason for ruling Q1 "yes".

## Release Notes

Settings on the web now keeps what only the web can do — your ESP32 board controller, and setting a password if you signed in with Google or Apple. Your profile, connected apps and account deletion live in the app, and the page links you straight there.


marcodejongh:
Thanks — three of four addressed in `5fa9d41`, one rebutted.

**1. Unreachable branches not flagged inline — fixed.** Both branches now carry an inline W-26 (#4442) marker naming the consumer that can't reach them and the catalog keys a trim would strand, so the next reader learns it from the file rather than the PR body.

**2. `STEP_LABELS` may be orphaned — confirmed, and documented rather than removed.** Grep across `packages/{web,mobile,shared}`: the only references are the export itself and the moved `import-progress-steps` test, which asserts against it. That was already true on `main` before the lift (`getStepLabels(t)` supplies the labels the UI actually renders), so it is pre-existing, not stranded here. Deleting it means rewriting the moved test to source its expectations from the catalog, which is a change to a test this PR was only supposed to relocate. It now carries a comment saying exactly that, flagged for W-26.

**3. `APP_URL` guard — this one doesn't hold.** `APP_URL` cannot be empty or undefined:

```ts
// app/lib/app-origin.ts
export const APP_URL = process.env.NEXT_PUBLIC_APP_URL ?? DEFAULT_APP_ORIGIN;
```

`DEFAULT_APP_ORIGIN` is the shared prod constant from `@boardsesh/shared-schema/app-origins`, so an unset env var falls back to `https://app.boardsesh.com` rather than producing `undefined/profile/more`. The test imports the same binding the component does, so it tracks whatever the build resolves.

**4. `kilter-oauth-linking` PostHog consumers — checked against the live project (412845).**

- `kilter-oauth-linking` exists and is ACTIVE (flag id 737576). It is **unchanged** — this PR edits `app/flags.ts`, not PostHog.
- Its only web consumer was `aurora-credentials-section.tsx`, deleted here. Web's `FEATURE_FLAG_KEYS` is just the list web's own `FeatureFlagsProvider` resolves, so dropping the key stops web fetching a value nothing on web reads.
- Mobile is unaffected: it resolves the flag through its own `FEATURE_FLAG_DEFINITIONS` in `packages/mobile/src/providers/feature-flags-provider.tsx:42`, and `BoardAccountsSection.tsx:281` still gates on it.
- `garmin-watch` is a different story worth recording: **no such flag has ever existed in PostHog.** It was in web's discovery list and in mobile's definitions, but the project has 13 flags and that key is not among them — same shape as the `gym-kiosk` flag that shipped gated on a key nobody created. Mobile's `more.tsx:98` row therefore resolves `undefined` (off) regardless, which is the intended default until the Connect IQ app ships.

Also swept in the same commit: `docs/{aurora-sync,kilter-sync,integrations}.md` no longer claim the web settings screen reads `kilter-oauth-linking`. `docs/ui/18-settings.md` is deliberately left alone — that directory is the frozen web spec the RN rewrite was built from, and W-16/W-19 left `06-play-view.md` and `16-feed.md` in place after deleting those surfaces.

marcodejongh:
Review round 2 landed in `feff235f0` + `07905d859`. Two things from my earlier comment are now superseded — flagging them so nobody reads the thread top-down and gets the wrong picture.

**`STEP_LABELS` — I said "documented rather than removed". It is now removed.** The counter-argument was right: a test asserting against a constant nothing renders is not coverage, and the moment of the lift is the moment it becomes provably dead. The moved `import-progress-steps` test now asserts `tFromCatalog('settings', 'aurora.import.steps.<step>')` — the values `getStepLabels(t)` actually feeds the component. The old assertion passed only because those six hardcoded strings happened to equal the six en-US catalog values verbatim; a copy edit would have red-ed it and the apparent fix would have been to edit dead code. Probed: pointing `steps.resolving` at the `dedup` key reds `renders all step labels`.

**`.unresolvedList` has three call sites, not four.** The PR body said four and double-counted; `grep -c 'styles.unresolvedList' board-import-prompt.tsx` → 3, at `:511`, `:534`, `:557`. The substantive half of that correction stands — three classes dropped, not four.

The bigger catch this round was outside the diff: **`/privacy` documented `boardsesh.com/settings` as the web account-deletion route in all four locales**, and this PR deletes that button. That is an indexable page and legally load-bearing copy, so the policy now points at the app's More screen in copy and in the `href`. Full accounting of all 17 findings — 15 fixed, 2 rebutted with evidence — is in the PR body under "Review round".

marcodejongh:
Follow-up to Claude's latest review, pushed in `5cf4dca79`:

- Removed the newly stranded **web-only** `kilterNew` variant/render branch. The three catalog keys stay because mobile's production `BoardAccountsSection` actively uses and tests them; the orphan checker scans mobile too.
- Added regression coverage for both retired `?kilter=error` callback forms: settings renders normally and no stale callback snackbar appears.
- Kept the “no other content links” invariant while scoping it to `main a`, so header navigation changes do not make the settings-content test brittle.

Two findings are pre-existing rather than introduced here: `createKilterHandoffStartUrl` already had no web caller on the rebased base (its caller was removed by `e354f6b00` on 2026-06-23), and the uncancelled profile fetch predates this PR and already gated the page plus `SetPasswordSection`. Those belong in a separate cleanup rather than this deletion diff.

Validation: 2 focused files / 17 tests passed; scoped `vp check --fix`, `check:i18n:orphans`, the full web typecheck/build, pre-commit i18n checks, and `git diff --check` all passed.

PR #4492, opened 2026-08-15
fix(web): degrade the sitemap index instead of 503ing the whole thing

### LINKED ISSUE #4476: /sitemap.xml 503s on a cold cache and fails every production deploy smoke
`/sitemap.xml` returns **503** on a cold cache, which failed the post-deploy smoke on the most recent production deploy and marked the whole run red — even though `deploy-web`, `deploy-app-web`, `deploy-production-backend` and `migrate` all succeeded.

## Evidence

Production deploy run [31870111546](https://github.com/boardsesh/boardsesh/actions/runs/31870111546), head `ec71ad2a4` (W-22, #4464), 2026-08-15:

```
ok   homepage renders server-side
ok   robots.txt serves and points at the sitemap
     sitemap.xml serves a sitemap index pointing at shards: attempt 1 failed (expected HTTP 200, got 503); retrying in 5s...
     sitemap.xml serves a sitemap index pointing at shards: attempt 2 failed (expected HTTP 200, got 503); retrying in 5s...
FAIL sitemap.xml serves a sitemap index pointing at shards (/sitemap.xml — expected HTTP 200, got 503)
ok   the static sitemap shard serves URLs
ok   auth session endpoint answers anonymously
ok   ws-auth answers anonymously (kiosk + embed depend on it)
5 passed, 1 failed, 2 skipped
```

Three attempts across ~31 seconds (06:48:11 → 06:48:42Z), 503 every time. Note that **`/sitemaps/static.xml` passed in the same run** — this is specific to the index, not a general outage.

Once warm it is fine. Six consecutive fetches at 07:20Z: `HTTP 200` in 37–80 ms, `x-vercel-cache: HIT` every time. So the CDN masks it, and the only reliable way to see it is exactly when the smoke sees it — immediately after a deploy, on a cold cache.

## Why the index fails when the shards do not

`buildSitemapIndexXml` (`packages/web/app/lib/seo/sitemap/shard-registry.ts:145-163`) is the one code path that runs **every** builder:

```ts
const built = await Promise.all(SHARD_REGISTRY.map(async (shard) => ({ shard, items: await shard.build() })));
```

Any builder that throws takes the whole index to `unavailableResponse()` → 503. An individual shard route only runs its own builder, which is why `static.xml` sailed through.

Two amplifiers in that fan-out:

1. **`getAllBoardConfigsOrThrow`** (`server-popular-configs.ts:59-78`) aborts at `SITEMAP_FETCH_TIMEOUT_MS = 10_000`. On a cold serverless instance hitting a cold backend and a cold database, 10 s is reachable — and the smoke's three attempts spaced ~15 s apart line up with three consecutive timeouts.
2. **`fetchPlaylistSitemapRows`** (`seo/sitemap/playlist-query.ts`) has **no timeout at all** — no `AbortController`, no deadline. It can hang unbounded, and inside `Promise.all` it holds the index with it.

The fail-closed doctrine itself is right — a sitemap that silently loses URLs tells Google those pages were deleted, which is worse than a 503. The bug is the **blast radius**: one slow builder takes down an index whose other four shards were ready.

## Why this matters more next week

W-23 (#4436) adds climb shards over a 128,655-row tier-2 selection. If they join the same `Promise.all` and the same all-or-nothing rule, `/sitemap.xml` goes from "occasionally 503s on a cold deploy" to "503s whenever the heaviest query in the app is slow" — and it takes the working board, playlist and static shards down with it every time.

I have already ruled on the shape of the fix for W-23 ([#4436 comment](https://github.com/boardsesh/boardsesh/issues/4436#issuecomment-5301109022), item 4): the **shard route** stays fail-closed, but the **index** must degrade — log loudly, omit the shard that failed, and serve the rest with a 200. This issue is that same fix applied to the shards that already exist.

## Proposed fix

1. **Per-shard isolation in the index.** Replace the all-or-nothing `Promise.all` with per-shard error capture: a builder that throws or times out is logged and omitted from the index; the index still 200s with the shards that built. Only an index where *every* shard failed should 503.
2. **Give every builder a deadline.** `fetchPlaylistSitemapRows` has none. Whatever the index's own budget is, no single builder may exceed it.
3. **Cache the index's per-shard summaries** so a cold instance is not re-running every builder to answer a request whose output changes daily at most.
4. **Consider sequencing rather than `Promise.all`** — the concurrent fan-out is the shape #4461 is about, and the index has no latency requirement that justifies it.

## Test oracle

- Index builder with one shard stubbed to throw → response is **200**, lists the other shards, and logs the failure. Mutation-probe by restoring the all-or-nothing behaviour and watching it red.
- Index builder with *every* shard throwing → 503.
- A builder that never settles → the index still answers within its budget rather than hanging.

The existing tests cannot catch this: they exercise builders that resolve. The failure only appears when one is slow or throws, which is the case nobody wrote.

## Related

- #4436 (W-23) — must not land on top of this without the index-degradation fix.
- #4461 — the same cold-start/fan-out shape on the front door; that PR is in flight and adds read deadlines there.
- The deploy will keep marking itself **red** on this until it is fixed, which trains everyone to ignore a red production deploy. That is the second-order cost and it is the expensive one.

## Release Notes

none — internal reliability fix for the sitemap index (the template checkbox is ticked).

marcodejongh:
## Summary

Fixes #4476.

`/sitemap.xml` 503'd on a cold cache and failed the post-deploy smoke on two
consecutive production deploys, while `/sitemaps/static.xml` passed in the same
run. The index was the one path that ran every builder, inside a `Promise.all`,
so one cold boards fetch took the four shards that were ready down with it.

## What changed

**The doctrine splits by layer.** A shard route stays fail-closed — it still 503s
when its builder throws, when a shard that `expectsUrls` builds none, or when it
is past the URL budget. The index degrades: a builder that throws, misses its
deadline, comes back unexpectedly empty or blows the budget is logged loudly, its
`<sitemap>` entry is omitted, and the index still answers 200 with whatever built.
`buildSitemapIndexXml` throws only when there is nothing left to publish.

**A degraded index is cached for a minute, not 25 hours.** This is the half that
makes degradation safe rather than a worse bug. The full window is `s-maxage=3600`
plus a day of `stale-while-revalidate`, and the 503 it replaces was `no-store` and
therefore never cached at all — so before this change the copy the CDN eventually
held was always complete. Serving a partial index under the full window would
trade a self-healing 503 for a cacheable lie: one cold start pinning "boards.xml
does not exist" at the edge for 25 hours, while the URL serves perfectly the whole
time. A degraded 200 gets `s-maxage=60, must-revalidate` plus an
`X-Sitemap-Degraded` header naming the shards it dropped.

**Every builder is raced against `SHARD_DEADLINE_MS` (3 s).** A try/catch only
sees a builder that rejects; `fetchPlaylistSitemapRows` has no bound of its own,
and a builder that never settles holds the whole index to the platform timeout.

**Concurrent, with only a per-shard deadline.** `Promise.allSettled` bounds the
walk by max(builder), not sum(builder). The fan-out is two I/O calls against two
different backends plus three hardcoded builders — not the many-concurrent-heavy-
scans shape of #4461, which W-23 sequences *within* its climbs query for exactly
that reason.

**The post-deploy smoke was widened in the same PR**, because it would otherwise
have gone permanently green: it asserted "any one shard `<loc>`", which a degraded
index satisfies, and `static` is a hardcoded builder that cannot fail. It now
names both `expectsUrls` shards (`static.xml`, `boards.xml`); the three
declared-empty ones are legitimately absent and stay out of it.

Smaller items, each from review:

- The index applies the URL budget the shard route enforces, so it stops
  advertising a shard whose own URL is guaranteed to 503. Counted with a new
  `allLocalesUrlCount` instead of expanding up to 45,000 entries to read
  `.length` on a `force-dynamic` route; a unit test pins the two together.
- A shard that contributes no entry is logged, so "no public playlists yet" and
  "the playlists query regressed to `[]`" are separable from the logs.
- Test resets moved out of a `finally` that sits downstream of an `await` a
  regression can hang, into an `afterEach`. Before, one regression cascaded leaked
  fake timers into unrelated tests and presented as three failures.

## Two things the review changed my mind about

**1. Sequencing under a total budget was wrong.** The first draft walked the
registry sequentially under an 8 s `INDEX_DEADLINE_MS`. Review showed that makes
the tail of the registry the deterministic victim: five builders each a
comfortable 1.9 s — every one inside its own deadline — spend the budget before
`playlists` runs, so it is dropped for its position rather than its latency. Under
`allSettled` the total budget stops being needed at all, and the constant is gone.
The stated justification for sequencing was also wrong: `withDeadline` stops
waiting, it does not cancel, so sequencing never bounded pool load either way.

**2. The stall story was not what production recorded.** The smoke logged HTTP
503 three times, and 503 is this file's own `unavailableResponse` — a platform
kill would be a 504. `getAllBoardConfigsOrThrow`'s 10 s abort surfaces as a throw,
so the try/catch alone covers what was actually measured. The deadline is for the
mode a try/catch structurally cannot see. The docblock and the docs now say that
rather than claiming the deadline fixed the recorded failure. Same pass corrected
a false claim that the boards builder is `unstable_cache`d — the wrapper in that
file is `fetchPopularBoardConfigs`, the limit=12 homepage variant.

## Deliberately not in this PR

- **No deadline on the shard routes.** A 3 s bound there would 503 a legitimately
  slow cold boards fetch (its own abort is 10 s) on a URL that works today. The
  asymmetry is now stated in the code: the index's deadline protects the *other*
  shards, and omitting one for sixty seconds is cheap where 503ing a working URL
  is not. The residual gap — a stall past the platform ceiling gets a platform 5xx
  instead of our 503 — is called out rather than papered over with a timer that
  cancels nothing.
- **No `AbortSignal` threaded into the builders, and no statement timeout on the
  playlists query.** That is the only thing that would bound the abandoned work
  rather than the waiting. Follow-up.
- **No caching layer for the per-shard summaries.** It needs a summary/build split
  like #4483's. Follow-up; the short degraded window is what makes missing the
  deadline survivable meanwhile.

## Interaction with #4483 (W-23)

Matched to its shapes so the two collapse rather than fight:

- `withDeadline<T>(work, ms, label)` is byte-identical to #4483's helper — keep
  one copy on merge.
- `SHARD_DEADLINE_MS` (mine) and `SUMMARY_DEADLINE_MS` (#4483) are both 3 s and
  can become one constant.
- `xmlResponse` takes a `cacheControl` parameter in both PRs; #4483 adds
  `CLIMB_CACHE_CONTROL` where this adds `DEGRADED_CACHE_CONTROL`.
- #4483's paged loop appends onto the same `entries: SitemapIndexEntry[]` array
  with the same log-and-omit `catch`, after the fixed-shard walk. Its callers need
  the new `{ xml, degradedShards }` return, and its paged failures should join
  `degradedShards` so a dropped climbs shard also gets the short cache window.

A conflict in `shard-registry.ts` and `docs/web-reposition.md` is expected;
**#4483 should be the side that rebases**, since this is the smaller, more
mechanical half.

## Release Notes

## Test plan

- `vp run typecheck` — green.
- `vp test run --reporter=agent --project web app/lib/seo/sitemap` — 53 passed.
- `vp test run --reporter=agent production-smoke` — 14 passed.
- `vp check` — 35 errors / 325 warnings, all pre-existing on `main` (#4488). None
  of the seven files this PR touches appears anywhere in the output.
- Eight mutation probes, each applied alone from a checksummed pristine copy and
  restored with `\cp -f` (md5 verified back): degraded index keeps the full cache
  window → 2 red; drop the `X-Sitemap-Degraded` header → 3 red; sequential walk
  instead of `allSettled` → 1 red (the all-five-slow test); drop the per-shard
  deadline → 1 red; drop the index's over-budget check → 1 red; drop the
  declared-empty warn → 1 red; drop the smoke's `boards.xml` assertion → 1 red;
  `allLocalesUrlCount` off by one locale → 2 red. Note that the last two probes on
  the *previous* revision produced phantom cascades from leaked fake timers; with
  the `afterEach` fix each now reds exactly its own test.

### DIFF
--- a/docs/web-reposition.md
+++ b/docs/web-reposition.md
@@ -391,17 +391,57 @@ shard it already has. `buildSitemapIndexXml` throws only when nothing is left to
 publish; an empty `<sitemapindex>` under an hour of `s-maxage` is the harm the
 doctrine was written against.
 
+**A degraded index is cached for a minute, not for 25 hours.** This is the half
+that makes degradation safe rather than a worse bug. The full window is
+`s-maxage=3600` plus a day of `stale-while-revalidate`, and the 503 it replaced
+was `no-store` and therefore never cached at all — so before this change the copy
+the CDN eventually held was always complete. Serving a partial index under the
+full window would trade a self-healing 503 for a cacheable lie: one cold start
+pinning "boards.xml does not exist" at the edge for 25 hours while the URL serves
+perfectly the whole time. A degraded 200 gets `s-maxage=60, must-revalidate` and
+an `X-Sitemap-Degraded` header naming the shards it dropped, so the next crawl
+re-attempts and the degradation is visible on the response rather than only in a
+log line.
+
+The post-deploy smoke was widened in the same PR for the same reason. It asserted
+"any one shard `<loc>`", which a degraded index satisfies — and `static` is a
+hardcoded builder that cannot fail, so the detector that caught #4476 would have
+been permanently green afterwards. It now names both `expectsUrls` shards
+(`static.xml`, `boards.xml`); the three declared-empty ones are legitimately
+absent and stay out of it.
+
 **Slow is a failure mode, and a try/catch cannot see it.** Each builder in the
 index walk is raced against `SHARD_DEADLINE_MS` (3 s, the value W-23 settled on
-for its paged summary), and the walk as a whole against `INDEX_DEADLINE_MS` (8 s).
-The production failure was not a rejection: `getAllBoardConfigsOrThrow` sits on a
-10 s abort and `fetchPlaylistSitemapRows` has no bound at all, so a cold instance
-stalls until the platform kills the request — a 5xx on all five shards, strictly
-worse than the 503 the fail-closed rule was protecting. The walk is **sequential**,
-not `Promise.all`: the concurrent fan-out over a database query and a GraphQL fetch
-is the pool-starvation shape #4461 is about, and the index has no latency
-requirement that justifies it. Sequencing is only safe because of the total budget
-— five 3 s deadlines back to back is 15 s, past the serverless ceiling.
+for its paged summary, so the two collapse into one constant on merge). Be
+precise about why: the two recorded production failures were _rejections_ — a 503
+is the route's own `unavailableResponse`, and `getAllBoardConfigsOrThrow`'s 10 s
+abort surfaces as a throw — so the try/catch alone covers what was measured. The
+deadline covers the mode a try/catch structurally cannot see, a builder that never
+settles (`fetchPlaylistSitemapRows` has no bound of its own), which would hold the
+index to the platform timeout and 5xx all five shards. The index's 3 s and the
+boards builder's own 10 s are _supposed_ to disagree: failing the shard route
+costs a working URL, while missing the index deadline costs one shard for sixty
+seconds. Cheap to be wrong, so be impatient.
+
+**Concurrent, with only a per-shard deadline.** An earlier draft of this fix
+sequenced the walk under a total 8 s budget, and that made the tail of the
+registry the deterministic victim: five builders each a comfortable 1.9 s — every
+one inside its own deadline — spend the budget before `playlists` runs, so it is
+dropped for its position rather than its latency. `Promise.allSettled` bounds the
+walk by max(builder) instead of sum(builder), which removes the starvation and
+the need for a total budget at once. The fan-out is two I/O calls against two
+different backends plus three hardcoded builders — not the many-concurrent-heavy-
+scans shape of #4461, which W-23 sequences _within_ its climbs query for exactly
+that reason. Sequencing would not have bounded pool load anyway: `withDeadline`
+stops waiting, it does not cancel.
+
+**Known follow-ups.** `withDeadline` bounds this request's latency, not the
+abandoned query still holding a connection — a real bound needs an `AbortSignal`
+threaded into `fetchPlaylistSitemapRows` or a statement timeout on its query.
+`getAllBoardConfigsOrThrow` is uncached (the `unstable_cache` in that file wraps
+`fetchPopularBoardConfigs`, the limit=12 homepage variant) and `/sitemap.xml` is
+`force-dynamic`, so every CDN miss re-runs it live; caching the per-shard
+summaries is what would make 3 s comfortably attainable on a cold instance.
 
 **Two shards ship declared-empty, and that is the point.** `gyms.xml` waits on
 #4381's public-gyms enumeration query, which the gym-discovery epic (#4372)
--- a/docs/web-reposition.md
+++ b/docs/web-reposition.md
@@ -370,14 +370,38 @@ tells Google the missing URLs were deleted. Three more cases fail the same way:
 a shard past its URL budget, a shard that `expectsUrls` but built none (a
 poisoned cache or a regressed query silently emptying `static` or `boards`), and
 a boards fetch whose `hasMore` says the 100-config API cap — the schema's hard
-max, so the fix is paging, not a bigger constant — dropped the tail. The index
-applies the same rule: it throws rather than publishing an index that quietly
-dropped a shard, and it lists only shards carrying at least one URL. For the
+max, so the fix is paging, not a bigger constant — dropped the tail. For the
 same reason the boards shard uses a _throwing_ popular-configs fetch, not the
 homepage's swallow-and-return-`[]` wrapper. `playlists` is deliberately **not**
 `expectsUrls` — zero public playlists holding a climb is a legitimate state, and
 failing closed there would take the whole index down over nobody having shared
 a list.
+
+**The index degrades; only the shard fails closed (#4476).** W-22 shipped the
+index under the same all-or-nothing rule, and production disagreed: `/sitemap.xml`
+503'd on a cold cache and failed the post-deploy smoke twice, while
+`/sitemaps/static.xml` passed in the same run. The index is the one path that runs
+_every_ builder, so one cold boards fetch took the four shards that were ready
+down with it. The split is now by layer — a shard route still 503s when its own
+builder throws, misses its budget or comes back unexpectedly empty, but the index
+logs loudly, omits that shard's `<sitemap>` entry and still answers 200 with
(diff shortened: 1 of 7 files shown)


marcodejongh:
Orchestrator note, and a correction to my own brief.

**I told the implementer to sequence the walk rather than use `Promise.all`**, reasoning that the concurrent fan-out is the shape #4461 is about and the index has no latency requirement that justifies concurrency. The `seo-semantics` lens showed that is wrong here, and the fixer was right to remove the invariant rather than pin it.

Two reasons the sequencing rationale did not survive contact:

1. `withDeadline` bounds *waiting*, not *work* — nothing is cancelled — so sequencing never actually bounded pool load. The stated benefit did not exist.
2. The fan-out here is two I/O calls against two different backends (one GraphQL fetch, one indexed query holding one of ten connections) plus three hardcoded builders. That is not the many-concurrent-heavy-scans shape of #4461, which is why #4483 sequences *within* its climbs query and this does not need to sequence *across* shards.

Worse, sequencing under a shared budget starves the tail of the registry by position — the last shard gets whatever the first four left it. Pinning peak concurrency at 1, as the test-strength lens proposed, would have hardened the defect rather than guarded against it.

The invariant that ended up pinned is the useful one: **bounded by `max(builder)`, not `sum(builder)`**, with a test driving all five builders at `SHARD_DEADLINE_MS - 1` and requiring all five to survive. A sequential rewrite reds exactly that test. That is a better guard than the one I asked for.

Two other things this PR found that I want visible, because both are the kind of thing that ships silently:

- **The degraded index was being served with the full 25-hour cache window.** One cold-start blip would have pinned a partial sitemap in the CDN for a day — strictly worse than the 503 this PR replaces. Now 60 s with `must-revalidate` and no `stale-while-revalidate`.
- **The post-deploy smoke could be satisfied by a fully-degraded index**, because it asserted "any one shard `<loc>`". It now names both `expectsUrls` shards. Without that, this PR would have turned a loud, correct failure into a quiet, wrong success — which is the exact trade a degrade-don't-fail change has to be careful not to make.

Marked ready: CI green, no unresolved review comments, no conflicts.

**Expected conflict with #4483 (W-23)**, which rewrites the same file to add the paged climbs shard and its own `SUMMARY_DEADLINE_MS`. Whichever lands second rebases; the two mechanisms should collapse into one. That rebase is mine.

PR #4518, opened 2026-08-16
fix(mobile): show sync spinner on Android app-bar bulb

alejandrosanchezcabana:
## Summary

On Android, tapping the app-bar lightbulb (or "Re-light board" in the BLE controls sheet reached by long-pressing it) pushes the current climb's LEDs to the wall over Bluetooth, but gave zero visual feedback while the write was in flight — the controls sheet closes immediately on tap. The play-drawer lightbulb already swaps to a spinner during a write via `useBluetoothWriteInProgress`; the Android app-bar bulb (`MaterialLightbulbAction`) never read that state.

- Swap the app-bar icon for a small `ActivityIndicator` while `isWriting` is true.
- Set `accessibilityState={{ busy: isWriting }}` to match the play-drawer bulb's a11y behaviour.
- Two pre-existing chrome tests (`collapsing-top-chrome.test.tsx`, `discover-top-chrome.test.tsx`) mock `react-native` locally without `Platform`; the new `ActivityIndicator` import pulls in `theme/colors.ts`, which reads `Platform.OS` at module-eval time. Added `Platform: { OS: 'android' }` (+ an `ActivityIndicator` stub) to both mocks.

## Release Notes

- The wall-sync bulb now spins while your climb is being sent to the board, so you're never left wondering if the tap registered.

## Test plan

- [x] `vp test run --project mobile` — 681 files / 6518 tests pass
- [x] `vp check`
- [x] New `MaterialBoardActions.test.tsx` covers idle → writing → idle icon/`busy` transitions via the real `write-activity-store`

### DIFF
--- a/packages/mobile/src/components/chrome/MaterialBoardActions.tsx
+++ b/packages/mobile/src/components/chrome/MaterialBoardActions.tsx
@@ -18,6 +18,8 @@ import { useTheme } from '../../providers/theme-provider';
 import { hapticLight } from '../../lib/haptics';
 import { useLightbulbControl } from '../ble/use-lightbulb-control';
 import { useBleControlSheet } from '../../providers/ble-control-sheet-provider';
+import { useBluetoothWriteInProgress } from '../../providers/bluetooth-write-activity';
+import { ActivityIndicator } from '../ActivityIndicator';
 import { Text } from '../Text';
 import { iconMap } from '../icon-map';
 import { AngleSelectorSheet } from '../play-drawer/AngleSelectorSheet';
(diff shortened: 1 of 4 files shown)


alejandrosanchezcabana:
Thanks for the flag. No files under `packages/mobile/src/lib/ble/` are modified in this PR — the new test only *imports* the existing, unmodified `createBleWriteActivityStore` factory to drive real `isWriting` state, the same pattern the already-merged `BleLightbulbButton-render.test.tsx` uses (`packages/mobile/src/components/ble/__tests__/BleLightbulbButton-render.test.tsx`). CLAUDE.md's Fable-review gate is scoped to changes *touching* BLE code in those directories; reading an untouched export from a test elsewhere doesn't fall under that. Proceeding without a Fable review, consistent with the existing precedent.

marcodejongh (approved):

Issue #4235, opened 2026-08-10
Postgres DSM exhaustion on POST /graphql, round 3: playlist resolvers are the last unguarded parallel-query path (mySmartPlaylistCounts)


marcodejongh:
`PostgresError: could not resize shared memory segment "/PostgreSQL.xxx" to 4194304 bytes: No space left on device` (pgCode 53100) is still firing on `POST /graphql` at ws.boardsesh.com — 350 events / 123 users over 14 days, ~19–27 per day, still going as of 2026-08-10.

**Priority: P2.** Not P1: climb search is fixed and unaffected, and the failure degrades the library page's playlist card counts rather than breaking a core climbing flow. Not P3: it hits 123 users, recurs every day, is the third recurrence of the same class, and the fix is mechanical.

Sentry: `BOARDSESH-AK` (350 events / 123 users), `BOARDSESH-B6` (4 events / 4 users).

**This is not a regression of #3856 or #4105.** Both fixes work and are deployed. #3865 (merged 2026-07-24T21:49Z) dropped daily counts from a 215/day peak to near zero for a week. #4124 (merged 2026-08-07T21:57Z) deployed by 2026-08-07T23:46Z — we know the exact minute because that is when its own new `graphqlPath` tag starts appearing.

That tag names the remaining offender. Of the 111 DSM events since the #4124 deploy, **111 are tagged `mySmartPlaylistCounts`** and 1 more is `smartPlaylist`. In the last 100 events, 93 are `mySmartPlaylistCounts` across 39 distinct client IPs.

## Cause

`packages/backend/src/graphql/resolvers/playlists/**` has no `withSerialPlan` and no `SET LOCAL max_parallel_workers_per_gather = 0` anywhere. Every query is a bare `db.execute(sql`...`)`. #4124 extracted the helper and applied it to search, feeds, and profile stats, but never touched playlists.

One `mySmartPlaylistCounts` request issues up to 8 concurrent unguarded queries:

- `packages/backend/src/graphql/resolvers/playlists/queries/smart-playlists.ts:370` — multi-CTE aggregate over `boardsesh_ticks` + `user_favorites`
- `smart-playlists.ts:442` → `resolveRecommendationBoardTarget` → `packages/backend/src/graphql/resolvers/playlists/helpers/recommendation-board-target.ts:80`, plus a `Promise.all` of two more at `:90-96`
[...]

Issue #4105, opened 2026-08-01
Postgres shared-memory (DSM) exhaustion on POST /graphql persists after #3856 — another unguarded parallel query


marcodejongh:
## Summary

Follow-up to #3856. Postgres parallel-query shared-memory exhaustion is still hitting `POST /graphql` after the `statsDrivenSearch` guard landed, so at least one other parallel-eligible hot-path query is still unguarded.

```
PostgresError: could not resize shared memory segment "/PostgreSQL.3299586270"
to 2097152 bytes: No space left on device
```

- **Sentry:** [BOARDSESH-AK](https://boardsesh.sentry.io/issues/BOARDSESH-AK)
- **Impact:** 292 events / 107 users over 14d
- **Transaction:** `POST /graphql`

## Timeline

`a31f88baf` — *fix(db): guard statsDrivenSearch against parallel-worker DSM exhaustion* — merged 2026-07-24, and #3856 closed the same day.

That fix did land and did work for the query it targeted: [BOARDSESH-AJ](https://boardsesh.sentry.io/issues/BOARDSESH-AJ) (`Failed query: select "board_climbs"."uuid", ...`, 135 events / 41 users) **stopped on 2026-07-24** and has not recurred.

BOARDSESH-AK did not stop. Post-fix bursts:

- 2026-07-26 22:17–22:22 (segments sized 1 MB, 2 MB, 4 MB)
- 2026-07-30 08:26:16–08:26:34 (2 MB, 2 MB, 4 MB)

Both are tight clusters of seconds, which reads as one endpoint tipping over under concurrent load rather than a steady background leak.

## Where to look

`SET LOCAL max_parallel_workers_per_gather = 0` currently appears in only two places:

- `packages/db/src/queries/climbs/search-climbs.ts:337,358`
- `packages/backend/src/db/queries/climbs/count-climbs.ts:49`

Everything else reachable from `POST /graphql` that scans `board_climbs` / `board_climb_stats` is still parallel-eligible. The varying segment sizes (1/2/4 MB) suggest more than one plan node, possibly more than one query.

Note this overlaps work in flight on `codex/speed-up-climb-search` — worth confirming that branch doesn't reintroduce parallel plans on paths that currently avoid them.

## Fix direction

[...]

Issue #3856, opened 2026-07-24
Climb search intermittently fails in prod: statsDrivenSearch trips Postgres parallel-query shared memory (the one hot-path query missing the max_parallel_workers_per_gather guard)


marcodejongh:
## Symptom

A slice of `SearchClimbs` GraphQL requests fail server-side in production, in bursts:

```
PostgresError: could not resize shared memory segment "/PostgreSQL.*" to 4194304 bytes: No space left on device
```

pgCode `53100`. The stack is entirely inside the `postgres@3.4.9` driver (`connection.js` `ErrorResponse`) — Postgres itself refusing to allocate a ~4 MB dynamic-shared-memory (DSM) segment for parallel-query workers because the Railway Postgres container's `/dev/shm` is saturated. Surfaced through graphql-yoga masking; the user gets a failed/empty climb search.

## Sentry (env=production)

| Sentry | Symptom | Events (14d) | Users | Last seen |
|---|---|---|---|---|
| [BOARDSESH-AK](https://boardsesh.sentry.io/issues/7616183441/) | `could not resize shared memory segment … No space left on device` (pgCode 53100) | 291 | **104** | 2026-07-23 |
| [BOARDSESH-AJ](https://boardsesh.sentry.io/issues/7616183421/) | `Failed query: select "board_climbs"…` — same DSM inner cause | 142 | 39 | 2026-07-23 |

~430 events across the two groupings; the AK latest event's request is `operationName=SearchClimbs`, variables `{kilter, angle 40, minGrade=maxGrade=15, onlyTallClimbs, boulders, sortBy=ascents desc, page=1, pageSize=100, setIds "1,20"}`. `server_name=boardsesh-backend`, node v22.23.1 / Alpine 3.24.1, triggered from the iOS client. ([BOARDSESH-9P](https://boardsesh.sentry.io/issues/7599058650/) `Failed query: select from users` on `updateProfile`, 101 ev / 23 users, is message-only with no pg cause — possibly the same DSM window on a different path, but unconfirmed.)

## Root cause (verified on `main`)

Those inputs route through `chooseSearchPath` → `'stats-driven-only'` → **`statsDrivenSearch()`** (`packages/db/src/queries/climbs/search-climbs.ts:266`). It is the **only** hot search path that does *not* wrap its query in a transaction issuing `SET LOCAL max_parallel_workers_per_gather = 0`. Its siblings were hardened after PR #1969 ("fix_postgres_drama") and both carry the guard:

[...]

PR #4538, opened 2026-08-19
fix(backend): guard the Boardsesh-grades sync pull against DSM exhaustion

### LINKED ISSUE #4528: Postgres DSM exhaustion on SyncClimbGrades — syncClimbGrades missing withSerialPlan guard
## What's happening

The `syncClimbGrades` GraphQL resolver throws `PostgresError: could not resize shared memory segment … to 2097152 bytes: No space left on device` (pgCode 53100) in production. This is the same DSM exhaustion pattern that was fixed in `syncClimbStats` (#3856 round 1, #4105 round 2, #4235 round 3), but the `syncClimbGrades` resolver was never guarded.

**Sentry:** BOARDSESH-AK  
**Events (14d):** 456 events, 123 users  
**First seen:** 2026-08-03 · **Last seen:** 2026-08-16 (active now)  
**Environment:** production (backend, no release tag)  
**GraphQL operation:** `SyncClimbGrades` on `POST /graphql` at `ws.boardsesh.com`

## Root cause

`packages/backend/src/graphql/resolvers/sync/queries.ts:504`

`syncClimbGrades` passes no `executor` to `runSyncPage`, so it uses the bare `db` pool without the `SET LOCAL max_parallel_workers_per_gather = 0` guard:

```ts
// syncClimbGrades — line 504 — NO guard
return runSyncPage({
  // executor defaults to raw db
  selectList: sql`board_type, climb_uuid, angle, …`,
  fromClause: sql`board_climb_grades`,
  scope,  // includes EXISTS (SELECT 1 FROM board_climbs bc WHERE …)
  …
});
```

The scoped query has an identical `EXISTS … FROM board_climbs bc` correlated subquery to `syncClimbStats`. The Postgres planner picks a Gather Merge plan for this shape, which allocates DSM — and fails under concurrent sync bursts.

`syncClimbStats` at line 443 has the exact same query shape and **already has the guard**, with a comment that even cites BOARDSESH-AK:

```ts
// A scoped stats pull walks board_climb_stats in cursor order and probes
// board_climbs through the EXISTS filter. Production chooses a Gather
// Merge for that shape, which has exhausted Postgres's DSM during sync
// bursts (Sentry BOARDSESH-AK, pgCode 53100). Keep SET LOCAL and the
// page SELECT on the exact same transaction handle.
return withSerialPlan(db, (transactionDb) =>
  runSyncPage({ executor: transactionDb, … }),
);
```

The fix was applied to `syncClimbStats` but not to the `syncClimbGrades` resolver immediately below it.

## Suggested fix

Wrap `syncClimbGrades`'s `runSyncPage` call in `withSerialPlan`, mirroring `syncClimbStats`:

```ts
// syncClimbGrades — packages/backend/src/graphql/resolvers/sync/queries.ts
return withSerialPlan(db, (transactionDb) =>
  runSyncPage({
    executor: transactionDb,
    selectList: sql`board_type, climb_uuid, angle, local_grade, universal_grade, grade_low, grade_high,
      confidence, ascensionist_count, computed_at, sync_seq`,
    fromClause: sql`board_climb_grades`,
    scope,
    updatedAtColumn: sql`board_climb_grades.computed_at`,
    seqColumn: sql`board_climb_grades.sync_seq`,
    cursor,
    limit: lim,
  }),
);
```

Add a test in `packages/backend/src/__tests__/sync-climb-stats-serial-plan.test.ts` (or a new file) verifying `syncClimbGrades` with a scoped query emits `SET LOCAL max_parallel_workers_per_gather = 0`.

## Prior rounds

- Round 1 (#3856, closed): `statsDrivenSearch`
- Round 2 (#4105, closed): additional parallel-query paths  
- Round 3 (#4235, closed): `mySmartPlaylistCounts`
- Round 4 (this issue): `syncClimbGrades`

## Release Notes

none

marcodejongh:
## Summary

Closes #4528.

`syncClimbGrades` builds exactly the same correlated-EXISTS scope as `syncClimbStats` — the scope was copy-pasted when the grades pull shipped — but it called `runSyncPage` with no `executor`, and `runSyncPage` silently defaults to the module-level `db` pool. So the page SELECT ran unguarded: no `SET LOCAL max_parallel_workers_per_gather = 0`, no error, nothing downstream to re-add it. Production plans that shape as a Gather Merge, each worker allocates a DSM segment, and Railway's small `/dev/shm` runs out (`pgCode 53100`).

Two changes:

1. **`runScopedBoardRefSyncPage`** (new private helper in `sync/queries.ts`) builds the scope **and** runs the page under `withSerialPlan` in one call. `syncClimbStats` and `syncClimbGrades` both go through it now. The failure mode here was literally "copied the scope, forgot the guard", so the fix is to make that omission unrepresentable rather than to add a third hand-written `withSerialPlan(...)` wrap. The guard is unconditional, including for the unscoped `board_type`-only pull — that's what `syncClimbStats` already did, and a full-table walk can pick a parallel plan too.
2. **`userTicks` gets the same guard.** Same bug class, different file — see "Second fix" below.

The emitted SQL is unchanged for both resolvers: same select lists, same predicate, same cursor/ORDER BY. A serial plan can only change latency, never the rows.

### Correcting the issue's numbers

The issue quotes **456 events / 123 users (14d)**. That is the whole of Sentry `BOARDSESH-AK`, which is a catch-all group for *every* `pgCode 53100` event across ~12 different `graphqlPath` values — not this resolver. Broken down by `graphqlPath`:

| period | path | events |
| --- | --- | --- |
| 14d | (empty) | 149 |
| 14d | `similarClimbs` | 121 |
| 14d | `mySmartPlaylistCounts` | 114 |
| 14d | `userTicks` | 25 |
| 14d | `syncClimbGrades` | 12 |
| 14d | `syncClimbStats` | 7 |
| 3d | `userTicks` | 25 |
| 3d | `syncClimbGrades` | 12 |
| 3d | `userGroupedAscentsFeed` | 1 |

So `syncClimbGrades` is 12 of 456 — but **all 12 landed in the last 3 days**, i.e. after #4468 shipped, so it is genuinely live and worth fixing. Just not at the stated blast radius.

### Second fix in the same class: `userTicks`

That breakdown also shows `userTicks` is the **largest** remaining live source, at 25 events in 3d. It runs an unbounded (no `LIMIT`) select over one climber's entire logbook, fanning out through five LEFT JOINs — two of them against the big catalog tables. Identical shape, identical failure, and `withSerialPlan` was already imported in that file (the You-page fan-out two resolvers down uses it). Wrapping it here rather than in a follow-up, because shipping only the grades half leaves the bigger part of the alert firing.

It's a different file and a public unauthenticated read path, so calling it out explicitly: this is a deliberate second fix, not scope creep. Happy to split it into its own PR if a reviewer would rather.

### Deliberately not fixed

- `userGroupedAscentsFeed` — 1 event in 3d, below the noise floor. Noting it rather than growing the diff.
- `packages/backend/src/scripts/export-board-snapshots.ts` runs its own `GRADES_WHERE` on a dedicated pool, not through the resolver, so it's untouched. Its comment claims it mirrors the resolver's scope exactly — `snapshot-export-golden.test.ts` is what keeps that honest, and it still passes.
- No other resolver in `sync/queries.ts` needs the guard. I checked all ten: only `syncClimbStats` and `syncClimbGrades` build the correlated EXISTS against 375k-row `board_climbs`; the rest are single-table user-scoped index lookups or small joins.

## Release Notes

- Downloading a board for offline use no longer drops out partway through the grade sync when the servers are busy.
- Your logbook and stats load reliably during peak hours instead of erroring out.

## Test plan

**Verified locally**

- `SKIP_TEST_INFRA=1 vp test run packages/backend/src/__tests__/sync-climb-stats-serial-plan.test.ts packages/backend/src/__tests__/user-ticks-serial-plan.test.ts --reporter=agent` — 5 passed. New coverage:
  - `syncClimbGrades` scoped pull: one transaction, nothing on the bare pool, `SET LOCAL` first, then the page SELECT on the **same** handle.
  - `syncClimbGrades` unscoped (`layoutId: null, sizeId: null`) pull is guarded too, so the always-on choice is pinned rather than accidental.
  - `syncClimbStats` keeps its existing assertions, plus a new one pinning the exact rendered predicate (`EXISTS (SELECT 1 FROM board_climbs bc WHERE bc.uuid = board_climb_stats.climb_uuid`) — that's what proves the helper's `${climbUuidColumn}` interpolation emits byte-identical SQL to the hand-written literal it replaced. Same assertion added for grades.
  - New `user-ticks-serial-plan.test.ts`: the logbook select lands on the transaction handle, and an unknown board type still rejects before a transaction is opened.
- `vp run typecheck:backend` — clean.
- `vp check` — no errors or warnings on any of the four changed files. (The repo-wide run reports 2 pre-existing `TS2307: Cannot find module '@gorhom/bottom-sheet'` errors from the mobile web-shims nested install, unrelated to this change and present before it.)

**Left to CI**

The DB-backed suites that actually exercise these resolvers end to end — `sync-pull-grades`, `sync-pull-board-scope`, `snapshot-export-golden`, `tick-queries`, `aurora-twin-dedup` — auto-start postgres+redis on fixed ports. This machine has ~25 worktrees sharing those ports and the infra bring-up timed out locally, so I'm letting CI arbitrate rather than chasing a collision that isn't mine. `snapshot-export-golden` is the important one: it calls the real `syncQueries.syncClimbGrades` and asserts byte-identical rows against the exporter's own `GRADES_WHERE`, so it is the strongest check that the refactor didn't perturb the emitted SQL.

**Post-merge (not a CI gate)**

Re-run the Sentry breakdown `issue:BOARDSESH-AK`, fields `[graphqlPath, count()]`, period 3d, and confirm `syncClimbGrades` and `userTicks` both drop to 0 — the same way #4468 was confirmed for `similarClimbs` / `mySmartPlaylistCounts`.

## What I verified vs inferred

**Verified:** the guard is genuinely absent from `syncClimbGrades` (mocked-db test recorded handles `['database']`, `db.transaction` called 0 times, before the fix); `withSerialPlan` really does run `SET LOCAL` and the caller's query on one shared transaction handle; the mobile pull client always sends `layoutId`/`sizeId` for per-board tables, so the correlated-EXISTS branch is the production shape; the helper emits the same SQL as the code it replaced; the Sentry per-path breakdown above.

**Inferred:** that the specific plan Postgres chooses in production is a Gather Merge over `board_climb_grades` — I did not get an `EXPLAIN` from prod. The inference rests on the shape being identical to `syncClimbStats`, whose guard demonstrably cleared its own events, and on the segment-size ladder in the error messages matching a parallel hash join's DSA doubling. Prior rounds: #3856, #4105, #4235, #4468.

### DIFF
--- a/packages/backend/src/__tests__/sync-climb-stats-serial-plan.test.ts
+++ b/packages/backend/src/__tests__/sync-climb-stats-serial-plan.test.ts
@@ -1,3 +1,9 @@
+/**
+ * Serial-plan guard coverage for BOTH per-board reference pulls — board_climb_stats
+ * and board_climb_grades. They share one helper (`runScopedBoardRefSyncPage`)
+ * precisely because syncClimbGrades once built the same correlated-EXISTS scope
+ * and ran it on the bare pool with no `SET LOCAL` (#4528), so pin both here.
+ */
 import { beforeEach, describe, expect, it, vi } from 'vite-plus/test';
 import { PgDialect } from 'drizzle-orm/pg-core';
 import type { SQL } from 'drizzle-orm';
@@ -68,6 +74,57 @@ describe('syncClimbStats serial-plan guard', () => {
     expect(transactionDatabase.execute).toHaveBeenCalledTimes(2);
     expect(recordedHandles).toEqual(['transaction', 'transaction']);
     expect(renderStatement(transactionDatabase.execute.mock.calls[0][0])).toMatch(SERIAL_PLAN_GUARD);
-    expect(renderStatement(transactionDatabase.execute.mock.calls[1][0])).toContain('FROM board_climb_stats');
+    const pageStatement = renderStatement(transactionDatabase.execute.mock.calls[1][0]);
+    expect(pageStatement).toContain('FROM board_climb_stats');
+    expect(pageStatement).toContain(
+      'EXISTS (SELECT 1 FROM board_climbs bc WHERE bc.uuid = board_climb_stats.climb_uuid',
+    );
+  });
+});
+
+describe('syncClimbGrades serial-plan guard', () => {
+  it('runs SET LOCAL before the scoped page query on the same transaction', async () => {
+    const result = (await syncQueries.syncClimbGrades(
+      undefined,
+      { boardType: 'tension', layoutId: 10, sizeId: 6, cursor: null, limit: 500 },
+      connectionContext(),
+    )) as SyncResult;
+
+    expect(result).toEqual({
+      documents: [],
+      cursor: { updatedAt: '1970-01-01T00:00:00.000Z', syncSeq: '0' },
+      hasMore: false,
+    });
+    expect(database.transaction).toHaveBeenCalledTimes(1);
+    expect(database.execute).not.toHaveBeenCalled();
+    expect(transactionDatabase.execute).toHaveBeenCalledTimes(2);
+    expect(recordedHandles).toEqual(['transaction', 'transaction']);
+    expect(renderStatement(transactionDatabase.execute.mock.calls[0][0])).toMatch(SERIAL_PLAN_GUARD);
+
+    const pageStatement = renderStatement(transactionDatabase.execute.mock.calls[1][0]);
+    expect(pageStatement).toContain('FROM board_climb_grades');
+    expect(pageStatement).toContain(
+      'EXISTS (SELECT 1 FROM board_climbs bc WHERE bc.uuid = board_climb_grades.climb_uuid',
+    );
+  });
+
+  // The guard is deliberately unconditional: a full-table walk of
+  // board_climb_grades can pick a parallel plan too, so don't let a future
+  // refactor make it depend on the layout/size scope being present.
+  it('guards the unscoped board_type-only pull as well', async () => {
+    await syncQueries.syncClimbGrades(
+      undefined,
+      { boardType: 'tension', layoutId: null, sizeId: null, cursor: null, limit: 500 },
+      connectionContext(),
+    );
+
+    expect(database.transaction).toHaveBeenCalledTimes(1);
+    expect(database.execute).not.toHaveBeenCalled();
+    expect(recordedHandles).toEqual(['transaction', 'transaction']);
+    expect(renderStatement(transactionDatabase.execute.mock.calls[0][0])).toMatch(SERIAL_PLAN_GUARD);
+
+    const pageStatement = renderStatement(transactionDatabase.execute.mock.calls[1][0]);
+    expect(pageStatement).toContain('FROM board_climb_grades');
+    expect(pageStatement).not.toContain('EXISTS (');
   });
 });
--- a/packages/backend/src/__tests__/user-ticks-serial-plan.test.ts
+++ b/packages/backend/src/__tests__/user-ticks-serial-plan.test.ts
@@ -142,6 +142,16 @@ describe('userTicks serial-plan guard', () => {
     expect(state.selectHandles).toEqual(['tx']);
   });
 
+  it('still opens exactly one guarded transaction for an empty logbook', async () => {
+    const ticks = await tickQueries.userTicks(null, { userId: 'user-123', boardType: 'kilter' });
+
+    expect(ticks).toEqual([]);
+    expect(state.transactions).toBe(1);
+    expect(state.executed).toHaveLength(1);
+    expect(state.executed[0]).toMatch(GUARD_PATTERN);
+    expect(state.selectHandles).toEqual(['tx']);
+  });
+
   it('rejects an unknown board type before opening a transaction', async () => {
     await expect(tickQueries.userTicks(null, { userId: 'user-123', boardType: 'nope' })).rejects.toThrow();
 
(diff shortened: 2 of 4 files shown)


marcodejongh:
Thanks — took the empty-logbook case (d12eccfde).

On `dbRead` for `userTicks`: leaving it on `db` deliberately. `dbRead` is a genuine replica whenever `READ_REPLICA_URL` is set, and `userTicks` is a read-your-writes path — the app saves a tick and refetches the logbook right after, so moving it to the replica would trade a DSM error for an intermittently stale logbook. That's a real behaviour change and a separate call from this DSM guard, so it shouldn't ride along here. Happy to file it as its own issue if you think the read-heavy profile paths should move to the replica.

On the predicate assertion only being load-bearing for the scoped branch: agreed, that's intentional. The unscoped branch is covered by the sibling case that asserts `not.toContain('EXISTS (')`, which is the other half of the same proof.

marcodejongh:
On the comment-length note: I don't think that rule exists. `grep -rn 'multi-line comment\|multi-paragraph\|one short line' CLAUDE.md .github/ docs/` returns nothing, and the surrounding code is the opposite convention — `runSyncPage`'s own JSDoc, the `STABILITY_WINDOW_SECONDS` block a few lines up in the same file, `withSerialPlan`'s 25-line JSDoc in `packages/db`, and the 21-line header on the sibling `playlist-serial-plan.test.ts` are all multi-paragraph on purpose.

That's deliberate for this class of bug specifically: the whole reason #4528 exists is that someone copy-pasted the scope out of `syncClimbStats` and the comment explaining *why* the guard was attached didn't travel with it. Shortening the new helper's docblock to one line would recreate exactly the condition that caused this. Leaving them as-is.

Issue #4228, opened 2026-08-10
Profile picture is not visible in the app anywhere


mohitkyadav:
I uploaded my profile picture, but it just does not render anywhere, I've tried uploading it again twice, on different network, doesn't work.

<img width="1125" height="2436" alt="Image" src="https://github.com/user-attachments/assets/7690dbdd-d5bd-4752-96f3-b33fe1af8bad" />

marcodejongh:
🤖 Claimed by bug-batch loop `bugbash20` (batch 3, session 53132652). Plan → draft PR incoming. Parallel loop: please pick a different issue.

marcodejongh:
**Carried over from #4532**, closed into this issue as a duplicate — keeping the details here so they aren't lost with it.

A second, independent reporter hit the same thing on **Android** (app 2.3.1 / 2000824, via the in-app bug drawer on `/profile`):

> I have tried to save a profile picture many times, and it never saves properly. Upload, crop, hit save, see it in the pfp, exit to main screen, pfp gone

Two things this adds to the original report:

1. **It reproduces on both platforms.** This issue is iOS; #4532 is Android. That points at one shared upload/persist path rather than two separate platform bugs — the crop step completes and the image renders locally, then the persisted value is gone on the next read.
2. **There is a reachable reporter.** That reporter agreed to follow-up — resolve **`app_feedback` row #10204** in the DB to contact them if a repro or a test build is needed.

Prioritised **P1**: the upload reports success and then silently discards the write, there is no workaround, and it is corroborated by two independent users on two platforms.

marcodejongh:
🤖 Picked up by the P1 loop `p1-five` (session 3482d370) as a **stalled-claim takeover**.

The existing claim on this issue is >13 days old and the PR attached to it is `CONFLICTING` and has not been touched in days. I'm restarting it: plan → adversarial plan review → implementation in a dedicated worktree → PR → code review → CI + review-feedback shepherding.

If the original loop is still on this, comment here and I'll stand down — I'll check before opening a PR.

marcodejongh:
[...]

PR #4539, opened 2026-08-19
fix(mobile): stop a failed avatar encode from saving an empty picture

### LINKED ISSUE #4532: 🐞 Bug report: I have tried to save a profile picture many times, and it never saves properly. Upload, crop, hit save...

## Comment

I have tried to save a profile picture many times, and it never saves properly. Upload, crop, hit save, see it in the pfp, exit to main screen, pfp gone

## Metadata

| Field | Value |
| --- | --- |
| Platform | android |
| App version | 2.3.1 (2000824) |
| Source | drawer-bug |
| Board | kilter / layout 1 / size 10 / sets [1,20] @ 40° |
| Climb | unknown |
| Session | unknown |
| URL | /profile |
| User agent | unknown |
| Feedback row | app_feedback #10204 |

## Contact

✅ Reporter agreed to follow-up — resolve **app_feedback #10204** in the DB to reach them.

marcodejongh:
## Summary

Closes #4532. Stacked on #4273 (`fix/4228-avatar-zero-byte-fallback`) — see "Stacking" below.

The reporter's account of it (app_feedback #10204, Android, 2.3.1): *"Upload, crop, hit save, see it
in the pfp, exit to main screen, pfp gone."* That is not a persistence bug — the URL round-trips
fine. The **bytes** never made it, and four consecutive layers treated zero bytes as success.

1. `expo-image-manipulator`'s Android `saveAsync` writes `resultBitmap.compress(fmt, q, fileOut)`
   and throws away the `Boolean` it returns (`ImageManipulatorModule.kt`), so a failed encode still
   resolves with a `uri` — pointing at a 0-byte file. iOS raises `CorruptedImageDataException` in
   the same place (`ImageManipulatorUtils.swift`), which is why this one is labelled Android.
2. `File.bytes()` on an existing-but-empty file returns an empty `ByteArray` without throwing.
3. Expo's WinterCG multipart encoder awaits the part's `bytes()` at encode time and writes whatever
   it gets, empty included.
4. The backend's `Buffer.concat([])` is truthy, so it sailed past the `!fileBuffer` check, PUT an
   empty object to S3 and answered `200 {avatarUrl}` — which the client dutifully persisted.

The preview on the edit screen reads the *local* file URI, so the crop looks right the whole time.
The moment you navigate away, `<Image>` can't decode the empty body, `Avatar` fires `onError`, and
you get initials. Nothing threw anywhere — matching Sentry, which has **zero** avatar / manipulator
/ picker errors across 90 days while the Android error stream is otherwise healthy (25 issue groups
in 30 days).

### What this PR changes (mobile only)

- **`compressAvatar` reads its own output and recovers.** It now takes the picker asset, reads the
  compressed file, and returns `{ uri, bytes }`. If the bytes come back empty — or the manipulator
  throws — it falls back to the picker's own crop, which is already square (`aspect: [1, 1]`) and
  only misses the resize/re-encode. `expo-image-picker`'s Android exporter drops the same
  `Boolean` in `CompressionImageExporter.writeImage`, so the fallback gets its own length check
  rather than being trusted, and is rejected above the backend's 2MB cap.
- **The fallback is sent under its real image type.** It skips the JPEG re-encode, so the bytes are
  whatever the picker produced — on Android routinely PNG or HEIC. The declared type comes from
  `asset.mimeType` with a matching filename, rather than claiming `image/jpeg` for everything.
- **A genuinely unusable pick now says so.** When neither the compressed file nor the picker crop
  has bytes — or the picker's type isn't one the backend accepts (HEIC, most often) —
  `compressAvatar` throws and the existing "could not compress" toast fires at pick time instead of
  a silent blank or a confusing upload failure on save. No new strings —
  `profile.validation.compressionFailed` and `profile.avatarUploadFailed` already exist in all four
  locales.
- **Telemetry on the fallback.** One `reportHandledError` per failure, tagged
  `source: avatar-compress`, carrying the two byte counts, the picker's MIME type and
  `Platform.OS` — sizes, type and platform only, no URIs or identifiers. `warning` when the
  fallback rescues the save, `error` when the climber loses the pick. This path had *no* telemetry
  at all, which is why it went unnoticed for so long; now we can see which layer produced the empty
  file rather than inferring it.
- **`AvatarUploadFile` carries bytes, not a URI to re-read.** #4273 made `uploadAvatar` read
  eagerly; this moves the read up to pick time, where a failure can still be recovered instead of
  only reported. `uploadAvatar` keeps the last-gate check and now also refuses anything over 2MB
  before wasting the POST. The cap stays mobile-local (`MAX_AVATAR_BYTES` next to
  `AVATAR_ENDPOINT`), mirroring the backend's `MAX_FILE_SIZE`.

JS/TS only. No native module, no Expo plugin, no native config — **OTA-safe, no fingerprint
change**. No migration, no new dependencies, not BLE.

### Stacking

Based on **`fix/4228-avatar-zero-byte-fallback`** (#4273), which already fixes the backend half —
`POST /api/avatars` rejects an empty part, and `/static/avatars/*` 404s a zero-byte object with
`Cache-Control: no-store`. Both PRs edit `packages/mobile/src/lib/avatar-upload.ts`, so landing
them independently on `main` would conflict in `uploadAvatar`. **If #4273 merges first, retarget
this PR to `main`.** The diff shown here is only this PR's half.

They are complementary, not duplicates: #4273 turns silent corruption into a visible warning; this
one lets the climber actually save a picture on the affected devices.

### Proven vs inferred — please read

I could not query production. There is no `DB_URL` credential on the machine this was written on,
and there is no unauthenticated surface that returns an `avatarUrl`, so **no live avatar's
`Content-Length` was measured for this issue.**

**Proven by code reading:**

- The four unchecked layers above, each read at the exact source line.
- The URL round-trips: `profile` and `updateProfile` both select and return
  `userProfiles.avatarUrl`, the mobile operations request it, and the hook writes it into
  `['profile']` and invalidates. "It doesn't persist" is ruled out at the code level.
- The iOS/Android divergence, which explains why #4228 (iOS) presents as a blank circle and #4532
  (Android) as initials.

**Inferred:**

- That the reporter's stored object is specifically zero bytes. The support for this is indirect:
  Sentry's silence rules out every failure mode that throws, and #4273 already *measured* zero-byte
  avatar objects in production on this bucket (`200 image/jpeg`, 0 bytes, against ~422KB for a
  healthy one). That leaves "persisted URL, unrenderable bytes".

**The one query that would settle it**, for anyone with the prod credential — read-only, safe:

```sql
select up.avatar_url, af.created_at
from app_feedback af
join user_profiles up on up.user_id = af.user_id
where af.id = 10204;
```

Then `curl -sI` that URL with the CDN bypassed. `200` with `content-length: 0` confirms this
outright. A healthy body would mean the URL is fine and the bug is in rendering instead — in which
case this PR is still correct hardening but is not the fix, and #4532 should be reopened.

### Not in this PR

Accounts that already stored a zero-byte object keep a non-null `avatar_url`. With #4273 in place
it now 404s, so they render initials, and **re-uploading repairs the row on its own** — which is
exactly what this PR makes possible again on Android. Nulling those rows would be a production data
write, so it is deliberately left for @marcodejongh to decide rather than done here.

## Release Notes

- Your profile picture sticks this time — on Android a photo that failed to compress was saved as an empty image, so it vanished the second you left the screen
- If a photo really can't be used, you get told up front instead of finding an empty circle later

## Test plan

`packages/mobile/src/lib/__tests__/avatar-upload.test.ts`

- The multipart part's bytes come from the descriptor, and `expo-file-system`'s `File` is never
  constructed — a lazy re-read can't creep back in
- Empty bytes reject with `Avatar upload failed` and `authenticatedFetch` is never called
- Over 2MB rejects the same way, before the POST

`packages/mobile/src/components/__tests__/EditProfileScreen.test.tsx`

- Compressed output empty + picker crop readable → uploads the picker crop's bytes, save completes,
  and exactly one `reportHandledError` warning fires carrying `compressedBytes: 0` /
  `originalBytes: 4` / `originalType`
- `saveAsync` rejects but the picker crop is readable → still falls back and saves, no error toast,
  and the manipulator's own error is what gets reported (not a synthetic "empty file")
- PNG pick → uploaded as `avatar.png` / `image/png`, not mislabelled as JPEG
- HEIC pick → refused at pick time with the toast and an `error`-level report, rather than sending
  the backend a type it rejects
- Neither readable → `profile.validation.compressionFailed` toast, Save stays disabled,
  `uploadAvatar` and `router.back()` never called
- Fallback crop over 2MB → refused with the same toast, reported once at `error` (which also pins
  that the screen's catch block doesn't file a second event for the same failure)
- Happy path unchanged, and asserts nothing is reported

Gates run locally: `vp run typecheck:mobile`, `vp run test:mobile`, `vp run check:mobile-bundle`,
`vp check`.

**Device QA still owed** (nothing here can cover it): on a real Android phone, Profile → Edit →
pick a photo → crop → Save → back to the main screen → the avatar is still there; force-quit and
relaunch → still there.

Related: #4228, #4273

### DIFF
--- a/packages/mobile/src/components/EditProfileScreen.tsx
+++ b/packages/mobile/src/components/EditProfileScreen.tsx
@@ -33,7 +33,7 @@ const COMPRESSION_QUALITY = 0.85;
  * camera roll — can't be uploaded honestly without the manipulator, so the
  * fallback refuses it rather than mislabelling the bytes as JPEG.
  */
-const FALLBACK_FILE_NAMES: Record<string, string> = {
+const FALLBACK_FILE_NAMES: Partial<Record<string, string>> = {
   'image/jpeg': 'avatar.jpg',
   'image/png': 'avatar.png',
   'image/gif': 'avatar.gif',
--- a/packages/mobile/src/components/EditProfileScreen.tsx
+++ b/packages/mobile/src/components/EditProfileScreen.tsx
@@ -24,8 +24,21 @@ const DISPLAY_NAME_MAX = 100;
 const MAX_DIMENSION = 1024;
 const COMPRESSION_QUALITY = 0.85;
 
-/** A picked image that has been read off disk and proven to have bytes in it. */
-type CompressedAvatar = { uri: string; bytes: Uint8Array };
+/**
+ * Filenames for the image types the backend accepts (`ALLOWED_MIME_TYPES` in
+ * `packages/backend/src/handlers/avatars.ts`, which derives the stored file's
+ * extension from the declared type). Only consulted for the fallback: the
+ * compressed path re-encodes to JPEG, so it knows what it produced. A pick the
+ * picker reports as anything else — HEIC, most often, straight off an Android
+ * camera roll — can't be uploaded honestly without the manipulator, so the
+ * fallback refuses it rather than mislabelling the bytes as JPEG.
+ */
+const FALLBACK_FILE_NAMES: Record<string, string> = {
+  'image/jpeg': 'avatar.jpg',
+  'image/png': 'avatar.png',
+  'image/gif': 'avatar.gif',
+  'image/webp': 'avatar.webp',
+};
 
 /**
  * Resize (if needed) and re-encode a picked image to a small JPEG, returning the
@@ -81,14 +94,14 @@ class AvatarCompressionError extends Error {
  * resize/re-encode. `expo-image-picker`'s Android exporter drops the same
  * `Boolean`, so the fallback gets its own length check instead of being trusted.
  */
-async function compressAvatar(asset: ImagePicker.ImagePickerAsset): Promise<CompressedAvatar> {
+async function compressAvatar(asset: ImagePicker.ImagePickerAsset): Promise<AvatarUploadFile> {
   let compressedBytes = new Uint8Array();
   let failure: unknown = null;
   try {
     const compressedUri = await renderCompressedJpeg(asset);
     compressedBytes = await new File(compressedUri).bytes();
     if (compressedBytes.length > 0) {
-      return { uri: compressedUri, bytes: compressedBytes };
+      return { uri: compressedUri, bytes: compressedBytes, name: 'avatar.jpg', type: 'image/jpeg' };
     }
   } catch (error) {
     failure = error;
@@ -103,18 +116,26 @@ async function compressAvatar(asset: ImagePicker.ImagePickerAsset): Promise<Comp
     failure ??= error;
   }
 
+  // The fallback skips the re-encode, so these bytes are whatever the picker
+  // handed us — on Android that is routinely PNG or HEIC, not JPEG. Send them
+  // under their real type, and give up when it isn't one the backend takes.
+  const fallbackType = asset.mimeType;
+  const fallbackName = fallbackType ? FALLBACK_FILE_NAMES[fallbackType] : undefined;
+  const canUseOriginal =
+    fallbackName !== undefined && originalBytes.length > 0 && originalBytes.length <= MAX_AVATAR_BYTES;
+
   // There is no telemetry on this path today (zero avatar/manipulator/picker
   // events in 90 days), which is exactly why the failure went unnoticed for so
-  // long. Sizes and platform only — no URIs, no user identifiers. Reported once,
-  // here, at the severity the climber actually experiences: a warning when the
-  // fallback rescues the save, an error when they lose the pick.
-  const canUseOriginal = originalBytes.length > 0 && originalBytes.length <= MAX_AVATAR_BYTES;
+  // long. Sizes, MIME type and platform only — no URIs, no user identifiers.
+  // Reported once, here, at the severity the climber actually experiences: a
+  // warning when the fallback rescues the save, an error when they lose the pick.
   reportHandledError(failure ?? new Error('Avatar compression produced an empty file'), {
     level: canUseOriginal ? 'warning' : 'error',
     tags: { source: 'avatar-compress' },
     extra: {
       compressedBytes: compressedBytes.length,
       originalBytes: originalBytes.length,
+      originalType: fallbackType ?? 'unknown',
       platform: Platform.OS,
     },
   });
@@ -122,7 +143,7 @@ async function compressAvatar(asset: ImagePicker.ImagePickerAsset): Promise<Comp
   if (!canUseOriginal) {
     throw new AvatarCompressionError('Avatar compression produced an unusable file');
   }
-  return { uri: asset.uri, bytes: originalBytes };
+  return { uri: asset.uri, bytes: originalBytes, name: fallbackName, type: fallbackType };
 }
 
 export function EditProfileScreen() {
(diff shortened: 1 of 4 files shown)


marcodejongh:
Thanks — three of the four are fixed in 0ea8427.

**Base branch (blocking)** — intended, not an accident. This is deliberately stacked on #4273: both
PRs touch `packages/mobile/src/lib/avatar-upload.ts`, so landing them independently on `main` would
conflict in `uploadAvatar`, and basing here keeps the diff reviewers see down to this PR's half.
It will be retargeted to `main` once #4273 merges, as the description says. Not merging it into a
feature branch.

**Double Sentry reporting** — fixed. `compressAvatar` now throws a named `AvatarCompressionError`
that `handlePickAvatar`'s catch skips, so one failure files one event. It also picks the severity
the climber actually experiences: `warning` when the fallback rescues the save, `error` when they
lose the pick entirely.

**`readBytesOrEmpty` swallowing everything** — fixed. The helper is gone; a failed read of the
picker's own file is kept as the reported cause when the manipulator didn't already supply one, so
a permission error no longer looks identical to an empty file.

**Hardcoded `MAX_AVATAR_BYTES` in the mock** — fixed with the partial mock you suggested. The test
now stubs `authenticatedFetch` (the reason the module was fully mocked — it drags in
expo-secure-store) and spreads `importOriginal()`, so the over-cap case runs against the real
constant.

**Missing telemetry assertion** — added, on both terminal paths: the over-cap test asserts `level:
'error'` with the byte counts and that it fired exactly once (which is also what pins the
double-report fix).

marcodejongh:
Both added in ac6f318 — the manipulator-throw test now asserts `reportHandledError` fired exactly once at `warning` with the byte counts, that the reported error is the manipulator's own (not a synthetic "empty file", so the two shapes stay distinguishable in triage), and that `router.back()` was called.

marcodejongh:
Good catch — fixed in 5435d2a. The fallback skips the re-encode, so I took option 1: the declared type now comes from the picker's own `asset.mimeType`, with a matching filename, instead of always claiming `image/jpeg`.

Option 2 doesn't work here: re-running `ImageManipulator` *is* the step that failed, so there's nothing left to convert with.

One consequence worth naming — a type the backend doesn't accept (`ALLOWED_MIME_TYPES` is jpeg/png/gif/webp, and HEIC is common straight off an Android camera roll) now stops the fallback rather than being mislabelled. The climber gets the "could not compress" toast at pick time instead of a confusing upload failure on save, and the reported event carries the type so we can see how often HEIC is what blocks the recovery.

New tests: a PNG pick uploads as `avatar.png` / `image/png`, and a HEIC pick is refused with the toast and an `error`-level report.

marcodejongh:
**1 — `gif`/`webp`:** already verified, both are accepted. `packages/backend/src/handlers/avatars.ts:15`:

```ts
const ALLOWED_MIME_TYPES = ['image/jpeg', 'image/png', 'image/gif', 'image/webp'];
```

`FALLBACK_FILE_NAMES` is that list, keyed to the extensions the same file writes (`MIME_TO_EXT`, and the stale-avatar cleanup at :90 sweeps exactly `jpg|png|gif|webp`). HEIC is called out separately only because it's the one that actually shows up on Android and therefore the one that will block a recovery in practice.

**2 — index access:** fixed in 9e02f04, `Partial<Record<string, string>>`.

**3 — comment length:** there is no such rule. `CLAUDE.md` says "Remove dead code as you go" and constrains variable names; it says nothing about comment length, and multi-paragraph JSDoc is the house style in exactly this code — see `avatar-upload.ts` (`withCacheBuster`, `uploadAvatar`), `error-reporting.ts` (`reportHandledError`'s bulleted policy block), and `handlers/static.ts` on the base branch. These blocks record why four layers each treated zero bytes as success, which is the only thing that stops someone "simplifying" the checks back out. Leaving them.

CI is green across the board.

PR #4540, opened 2026-08-19
feat(mobile): show setter notes on a climb

### LINKED ISSUE #4494: [Bug: UI] Climb: descriptions aren't being shown
When creating a climb there is an optional description field:

Image

When browsing climbs (including ones I made, that I know I added a description to at the time) the description is nowhere to be found.

Hopefully those descriptions are not disappearing into the void, but at minimum it would be nice to display any new ones in the detailed view of the respective climbs. If that's not possible/intended then the description field should instead be removed from the climb creation page, to avoid misleading setters and wasting their time trying to help climbers with details they won't even be able to see.

marcodejongh:
## Summary

Closes #4494.

Setter descriptions were never disappearing into the void — they were being written, validated and stored correctly the whole time. Nothing ever read them back.

The write path is intact end to end: the create drawer builds `withNoMatch(description, noMatch)`, `toSaveClimbInput` carries it, the Zod schema keeps it, and `tx.insert(UNIFIED_TABLES.climbs)` lands it in the existing `board_climbs.description` column. Two gaps on the read side:

1. **No surface rendered it.** Not the mobile play drawer (the live climb-detail surface), not the www climb front door — which already `SELECT`s `climbs.description` and throws it away.
2. **The data wasn't even on hand.** `CLIMB_SEARCH_FIELDS` deliberately omitted `description` ("no list UI renders them"), and the offline SQLite search did the same. A climb opened from the Climbs list reached the drawer with `description === undefined`; only a deep link carried it.

### What changed

- **`getDisplayDescription`** in `@boardsesh/shared-schema`, next to `withNoMatch`. Strips Aurora's canonical `No match\n` marker line, trims, then returns `''` when the whole remaining string is only a restatement of "no match". Renderers treat `''` as "no section at all".
- **`description` on the search wire** — both copies of `CLIMB_SEARCH_FIELDS` (mobile's hand-maintained one and the shared `@boardsesh/graphql` document) — and on the **offline SQLite read path** (`searchClimbsLocal`'s SELECT + `mapRowToClimb`). No backend change needed — `search-climbs.ts` already selected and returned the column. The offline half is not optional: `offlineAwareRequest` is local-first for any downloaded board and that flag is at 100%, so most searches never reach the server.
- **Play drawer**: a new `SetterNotesSection` — a collapsible "Setter's notes" section, `defaultExpanded`, rendered as the first below-fold section *after* the Logbook.
- **www climb front door**: an `h2` + `pre-line` paragraph right after `ClimbFacts`, before the CTA so it can't push the LCP image down. The prose is user-written, so it renders verbatim (never through `t()`) and stays out of the JSON-LD `description`, which remains the synthesised catalogue string.
- **i18n**: `climbs.frontDoor.setterNotes.heading` and `climbs.mobile.setterNotes.title` in all four catalogs.

### Two things worth a second look

**The suppression rule is anchored to the whole trimmed string, never a prefix.** In the dev catalog, 83,864 of the 177,868 non-empty descriptions are literally `No matching` / `No matching.` — which the header's no-match glyph already says. But a prefix match would eat real setter beta: `No matching feet allowed` and `No Houdini swap, spin around pls:). No matching.` are both genuine descriptions. Both have their own named regression test, on the helper and again through the rendered drawer.

**Placement is below the Logbook on purpose.** `PlayDrawer.tsx`'s `firstScreenReserve` / `computeLogbookScrollTarget` assume the Logbook is the first below-fold section; inserting anything above it silently breaks the fold math and the expand-into-view scroll. There's a test asserting the Logbook is still first. Moving setter beta above the fold is a separate, riskier decision.

### German string — please correct if you disagree

`de` heading/title: **"Notizen vom Routenbau"**.

The glossary's `Routenbauer*in` (`docs/i18n-german-glossary.md`) makes the literal genitive read as *"Notizen der\*des Routenbauer\*in"*, which is heavy for a section header, and the glossary explicitly prefers a neutral rewrite when the starred form gets ugly. "Notizen vom Routenbau" keeps the attribution to the setting without the star gymnastics. The strict-glossary alternative is *"Notizen der\*des Routenbauer\*in"* — one-line swap in all four spots if you'd rather have it.

Others: `en-US` "Setter's notes", `es` "Notas del setter" (glossary keeps *setter* in English, matching the existing `frontDoor.facts.setter`), `fr` "Notes de l'ouvreur" (matching `frontDoor.facts.setter` = "Ouvreur").

### Free side effects of putting `description` on the search wire

- **Remix stops silently losing the description.** `use-create-climb-navigation.ts:93` reads `climb.description ?? ''`, which was always `''` for a climb sourced from search.
- **The queue write-without-read flap is closed.** Mobile writes `description` onto both queue wire boundaries (`climb-to-queue-item.ts:35`/`:114`) while never reading it back — exactly the class `queue-climb-field-contract.test.ts` exists to catch. Same invariant as `557b3cb1b`: *no commit may leave a field on a write path whose matching read path lacks it.*

## Release Notes

- Setter's notes now show up on a climb — the description whoever set it wrote in the create form, right under your logbook in the play drawer and on the climb's web page.

## Test plan

New coverage:

- `packages/shared-schema/src/__tests__/no-match.test.ts` — `getDisplayDescription`: marker stripping, interior newlines preserved, `''`/null/undefined/`No match`/`No matching.` suppressed, and the named regression that real prose mentioning matching survives untouched.
- `packages/mobile/src/db/queries/__tests__/search-climbs-local.test.ts` — the offline read (a real `node:sqlite` DB) round-trips the description, and reports `''` rather than `undefined` when there is none.
- `packages/backend/src/__tests__/operations-schema-validation.test.ts` — parser-driven drift guard that `CLIMB_SEARCH_FIELDS` still selects `description`, so re-slimming the list can't silently blank the drawer section again.
- `packages/mobile/src/components/play-drawer/__tests__/deferred-sections.test.tsx` — notes render for real prose, nothing renders for `''`/`No match`/`No match\n`/`No matching.`/`no matching`, setter beta mentioning matching survives, and the Logbook is still the first section with the notes directly after it.
- `packages/web/app/b/[board_slug]/[angle]/view/[climb_uuid]/__tests__/view-seo-fragment.test.tsx` — the front door SSR-emits the heading + prose, emits neither for an empty or bare-no-match description, still carries exactly one `<h1>`, and leaves the JSON-LD description alone.

Gates run locally: `vp check` (0 errors), `vp run typecheck` (all packages), `vp run check:i18n`, `vp run check:i18n:orphans`, `vp run check:mobile-bundle` (iOS + Android), and every touched test file on its own project. The full mobile and web suites are too slow to finish in one local shell here, so CI is the gate for those — 8,112 tests green.

**No screenshots.** This box is Linux: the iOS simulator path is macOS-only and the local Android emulator segfaults on every renderer (a known, separately-tracked problem). The drawer section is a standard `CollapsibleSection` with a `subheadline` body, so it inherits the same look as Logbook / Community; the front-door section matches the existing Beta / Similar / Community sections. Happy to dispatch a CI screenshot run if you want eyes on it before merge.

### DIFF
--- a/packages/backend/src/__tests__/operations-schema-validation.test.ts
+++ b/packages/backend/src/__tests__/operations-schema-validation.test.ts
@@ -222,17 +222,38 @@ describe('mobile plain-string subscription documents validate against the execut
 // opened from search while leaving deep-linked climbs fine. Same technique as
 // queue-climb-field-contract.test.ts: read the hand-maintained template out of
 // mobile's source and assert on the parsed field set.
-describe('mobile CLIMB_SEARCH_FIELDS carries the fields the play drawer renders', () => {
-  const searchFields = climbFieldNames(
+describe('CLIMB_SEARCH_FIELDS carries the fields the play drawer renders', () => {
+  // The shared document nests the list under `searchClimbs { climbs { ... } }`,
+  // so collect by the `climbs` parent rather than the `climb` one climbFieldNames uses.
+  function searchResultClimbFields(operationSource: string): Set<string> {
+    const fields = new Set<string>();
+    visit(parse(operationSource), {
+      Field(node) {
+        if (node.name.value !== 'climbs' || !node.selectionSet) return;
+        for (const selection of node.selectionSet.selections) {
+          if (selection.kind === 'Field') fields.add(selection.name.value);
+        }
+      },
+    });
+    return fields;
+  }
+
+  const mobileSearchFields = climbFieldNames(
     `{ climb { ${extractTemplateConst(readMobileOperationsSource(), 'CLIMB_SEARCH_FIELDS')} } }`,
   );
+  const sharedSearchFields = searchResultClimbFields(publicOperations.SEARCH_CLIMBS);
 
-  it('selects a non-empty field set', () => {
-    expect(searchFields.size).toBeGreaterThan(0);
+  it('both copies select a non-empty field set', () => {
+    expect(mobileSearchFields.size).toBeGreaterThan(0);
+    expect(sharedSearchFields.size).toBeGreaterThan(0);
   });
 
-  it('selects description, so the play drawer can show the setter notes (#4494)', () => {
-    expect(searchFields.has('description')).toBe(true);
+  it('mobile selects description, so the play drawer can show the setter notes (#4494)', () => {
+    expect(mobileSearchFields.has('description')).toBe(true);
+  });
+
+  it('the shared search operation selects description too, so a future web caller starts correct', () => {
+    expect(sharedSearchFields.has('description')).toBe(true);
   });
 });
 
--- a/packages/backend/src/__tests__/operations-schema-validation.test.ts
+++ b/packages/backend/src/__tests__/operations-schema-validation.test.ts
@@ -214,6 +214,28 @@ describe('mobile plain-string subscription documents validate against the execut
   }
 });
 
+// Regression guard for #4494. `CLIMB_SEARCH_FIELDS` — the selection set behind
+// SEARCH_CLIMBS, which is how every climb reaches the Climbs list and, through
+// it, the play drawer — deliberately omitted `description` on the premise that
+// "no list UI renders them". The play drawer DOES render it now (setter notes),
+// so re-slimming the list would silently blank that section for every climb
+// opened from search while leaving deep-linked climbs fine. Same technique as
+// queue-climb-field-contract.test.ts: read the hand-maintained template out of
+// mobile's source and assert on the parsed field set.
+describe('mobile CLIMB_SEARCH_FIELDS carries the fields the play drawer renders', () => {
+  const searchFields = climbFieldNames(
+    `{ climb { ${extractTemplateConst(readMobileOperationsSource(), 'CLIMB_SEARCH_FIELDS')} } }`,
+  );
+
+  it('selects a non-empty field set', () => {
+    expect(searchFields.size).toBeGreaterThan(0);
+  });
+
+  it('selects description, so the play drawer can show the setter notes (#4494)', () => {
+    expect(searchFields.has('description')).toBe(true);
+  });
+});
+
 describe('native iOS queue subscription drift guard', () => {
   it('matches the shared-schema native queue operation exactly after whitespace normalization', () => {
     const nativeOperation = readNativeIosQueueUpdatesOperation();
--- a/packages/mobile/src/components/play-drawer/DeferredSections.tsx
+++ b/packages/mobile/src/components/play-drawer/DeferredSections.tsx
@@ -14,6 +14,7 @@ import { BoardseshGradeSection } from './BoardseshGradeSection';
 import { buildBoardseshGradeView, buildBoardseshGradeSummary, isMoonBoard } from './boardsesh-grade-utils';
 import { buildAngleGradeBars } from './community-utils';
 import { BetaVideosSection } from './BetaVideosSection';
+import { SetterNotesSection } from './SetterNotesSection';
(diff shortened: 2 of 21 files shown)


marcodejongh:
Thanks — went through all four. Three are addressed in `72c34d5`, one I'm leaving alone on purpose.

**Shared `CLIMB_SEARCH_FIELDS` omitted `description` (flagged twice) — fixed.** Both reviews are right that a comment is a weak guard. It now carries the field, `SEARCH_DRAFT_CLIMBS` drops its own extended fragment (`description` was the only thing it added), and the backend drift guard asserts on *both* copies rather than mobile's alone.

**`ONLY_NO_MATCH` terminators — widened.** Now `[.!?;,\s]*`, so `No matching?`, `No match;` and a trailing space are suppressed too. Tests cover each. On `nomatch` / `nomatching`: that one is deliberate — a whole string of `nomatch` is still nothing but a no-match restatement, and the anchors mean it can't reach into real prose. Said so in the comment.

**`persistKey="setterNotes"` is climb-agnostic — intentional, now documented.** Every sibling section here does the same (`logbook`, `community`, `similarClimbs`, `boardseshGrade`): collapsing one is a standing "don't show me this" preference, not a per-climb one. A per-climb key would also mean the section silently re-expands on every new climb, which is the opposite of what a user who collapsed it wants. Comment added at the call site.

**Payload size — measured, not a regression risk.** Dev catalog: mean description length 18 characters, max 254, and only ~31% of rows are non-empty. That's against a `frames` string already on every row of the same payload. The alternative (a per-climb `GET_CLIMB` on drawer open) costs a round-trip per climb opened and leaves Remix and the queue `description` flap broken.

**Bare `<Box component="h2">` on the front door — leaving as is.** It matches the file exactly: `frontDoor.beta.heading`, `frontDoor.similar.heading` and `frontDoor.community.heading` are all bare `<Box component="h2">` too. Restyling one of the four would be the inconsistency. If those headings should all move to `Typography`, that's a separate pass over the whole page.

marcodejongh:
On the comment-style note: I checked `CLAUDE.md` and there's no "one short line max" rule in it — the repo's convention is the opposite, explanatory block comments where the reasoning isn't obvious from the code (`PlayDrawer.tsx`, `withNoMatch`, `ClimbFrontDoor`'s "three deliberate absences" are all in that style, and the fold-math and whole-string-anchor constraints here are exactly the kind of thing that gets silently broken without them). Leaving those as they are.

`LocalClimbRow.description` — good catch, fixed in `33884eb`: both local reads select the column now, so the row type says `string | null` instead of pretending it can be absent.

The `persistKey` UX point is a fair thing to look at on a device, but it's the same behaviour as every other section in the drawer, so changing it here alone would make this one the odd one out.

marcodejongh:
All three are non-blocking and I'm leaving the code as is — reasoning for each:

**`margin: 0`** is a reset, not a spacing value, so there's no token to reach for. The design-token rule is about colours/spacing that should come from the scale; zeroing a browser default `<p>` margin isn't one of those. Inter-section spacing on this page is handled the same way for Beta / Similar / Community.

**`\s` in the trailing class is reachable** — the `.trim()` only removes *leading and trailing* whitespace, so an interior gap before the punctuation still hits it: `getDisplayDescription('No match .')` needs `[.!?;,\s]*` to match `" ."` and returns `''`. Drop the `\s` and that string starts rendering as a notes block.

**`persistKey`** is documented at the call site and matches every sibling section in the drawer (`logbook`, `community`, `similarClimbs`, `boardseshGrade`) — a per-climb key here would make this the one section that silently re-expands on every climb after you deliberately collapsed it.

ES-Alexander:
Confirmed working - thanks!

PR #4541, opened 2026-08-19
fix(mobile): tell two same-brand boards apart in the board picker

### LINKED ISSUE #4412: Board picker shows truncated names — can't distinguish two Kilter boards at different locations
When a gym has two boards of the same brand with long names, the board selector truncates them to the point where they look identical and the user can't tell them apart.

> "When the boards name are too long I cant tell them appart… these are kilters at 2 different locations but by the same owners"

An attachment shows the two entries side-by-side. The picker needs to surface enough detail (subtitle, location, serial, or tooltip) to make same-brand boards at the same gym distinguishable.

## Source

[Reported in #user-feedback on Discord](https://discord.com/channels/1288485679663415346/1497041999272874015/1537345452691820615) — posted in the feedback channel.

Reporter: `discord-a1c866f75756` · Posted: 2026-08-13T06:21:58.419000+00:00

## Attachments

attachment 1

marcodejongh:
## Summary

Two Kilter boards run by the same gym at two different locations rendered as the same card in the
Boards tab: a one-line name that ellipsised both to the same prefix, over a subtitle that read
`kilter` on both.

The subtitle was built from `board.sizeName`, and the backend hard-codes `layoutName`, `sizeName`,
`sizeDescription` and `setNames` to `null` on **every** `UserBoard` it returns — all three enrichment
paths in `packages/backend/src/graphql/resolvers/social/boards.ts` (`:451`, `:645`, `:1118`) carry a
`TODO: resolve from board-specific layout/size/set tables if needed`. So the subtitle always fell
through to the raw lowercase board type. Meanwhile the payload already carried `gymName`,
`locationName`, `layoutId`/`sizeId`, `angle` and `serialNumber`, and `@boardsesh/board-constants`
bundles the layout and product-size name tables on the device.

The fix is entirely client-side (no backend change, works offline, rides OTA):

- New `board-labels.ts` with `boardPlaceLabel` / `boardConfigLabel` / `boardRowSubtitle`. A board's
  subtitle is now where it is (gym, else location), else what it is (`Original 12×14`, resolved from
  the bundled tables), else the brand name — never the raw lowercase type.
- `disambiguateBoardSubtitles` runs once per rendered list: boards that still land on the same
  subtitle get ` · <facet>` appended. Facets are tried in order — size, layout, angle, then the last
  4 of the serial — and the first one that leaves every member of the group reading differently
  wins, falling back to the first that at least splits the group (a gym with three boards where two
  share a size). Boards that are genuinely identical on every facet are left alone; we don't invent
  a distinction.
- The carousel card now gives the name two lines with a fixed two-line reservation, so card heights
  stay uniform and a long name is no longer cut at the point where two boards look the same. The card
  is now `React.memo`'d (it was the only un-memoized row in the board-discovery set).
- Same always-null first term fixed on the My Boards management list, the board-detail sheet (whose
  Size spec row was permanently hidden), the expanded-gym rows in the directory, and the Bluetooth
  quickstart sheet (which rendered a dangling `kilter · `).
- Dropped mobile's duplicated board-brand map in favour of `BOARD_TYPE_LABELS` from
  `@boardsesh/board-constants`, which the shared file's own docstring asks call sites to migrate to.

Disambiguation is a list-level pass inside the existing `useMemo`s — never per row. `renderItem`
deps, `keyExtractor` and the FlashList setup are unchanged.

The backend TODO is deliberately out of scope here (the client resolves the same names offline), but
it is the reason the picker said "kilter" in the first place and is being filed as its own follow-up.

Closes #4412

## Release Notes

- Tell your boards apart in the picker: cards now show the gym or location under the name, fall back
  to the layout and size, and add the size, angle or serial when two boards would otherwise read the
  same
- Long board names get two lines instead of being cut off mid-word

## Test plan

- `vp run test:mobile` — 6555/6556 locally. The one failure is
  `queue-provider-session-updates.test.tsx`, a `waitFor` timeout under a load average of ~60 on this
  box; it passes on re-run of the same file with the same code and touches nothing this PR changes.
  Both CI `test-default` shards are green.
- `src/components/board-discovery` went 12 files / 172 tests → 13 files / 205 tests; with
  `src/components/gym-directory` the two dirs are 19 files / 250 tests.
- New `board-labels.test.ts` covers gym-over-location precedence, layout/size resolution against real
  bundled ids, an unknown id returning `null` rather than an id or `undefined`, and every rung of the
  disambiguation ladder — including the reported case (two same-gym Kilters), a three-board gym where
  only the second facet separates everyone, and the genuinely-indistinguishable case, which must be
  left untouched.
- The two other changed surfaces got integration coverage too: the gym directory's nested board rows
  (`gym-list-panel.test.tsx`) and the Bluetooth quickstart rows
  (`bluetooth-quickstart-sheet.test.tsx`).
- `vp run typecheck:mobile`, `vp check` and the pre-commit hook clean locally; CI `typecheck`,
  `lint`, `check` and `mobile-bundle` all green.
- `ota-check` confirms no native fingerprint change — this rides OTA.

### Visual sign-off needed — @marcodejongh

**This PR is not auto-mergeable and I could not produce screenshots.** I ran
`vp run mobile:android-shots` on this host and it failed with
`Emulator did not finish booting within 300s`; the emulator log ends in
`xvfb-run: Segmentation fault (core dumped)` partway through "Emulator is performing a full startup".
That is a limitation of this machine's software GPU, not of the change, and it reproduces with a
clean tree. The `ios-screenshots` path is macOS-only, and the committed Maestro app-store flow never
visits the Boards tab, so the CI Android screenshot job would capture nothing relevant either. I did
not mock anything up.

Two things worth eyeballing on a device:

1. **Boards tab carousel** with at least two same-brand boards — check the two-line name reservation
   keeps card heights uniform when one title wraps and its neighbour does not, and that the subtitle
   line reads the gym/location.
2. **My Boards** list and the **board-detail sheet** — the detail sheet's Size row now renders for the
   first time (it was hidden by the null server field), so it is worth confirming the string reads
   sensibly for a Homewall size like `12 x 12 with kickboard · Square`.

Reproducing the exact reported collision needs an account with two boards at one gym; if you have the
reporter's setup handy that is the fastest check.

### DIFF
--- a/packages/mobile/app/boards/index.tsx
+++ b/packages/mobile/app/boards/index.tsx
@@ -18,7 +18,7 @@ import { ActivityIndicator } from '../../src/components/ActivityIndicator';
 import { BoardCarousel } from '../../src/components/board-discovery/BoardCarousel';
 import { BoardModeCard, type ModeCardState } from '../../src/components/board-discovery/BoardModeCard';
 import { BluetoothQuickstartSheet } from '../../src/components/board-discovery/BluetoothQuickstartSheet';
-import { userBoardToItem, popularConfigToItem } from '../../src/components/board-discovery/board-items';
+import { userBoardsToItems, popularConfigToItem } from '../../src/components/board-discovery/board-items';
 import { offlineBoardRows } from '../../src/components/board-discovery/offline-board-items';
 import type { DiscoveryBoardItem } from '../../src/components/board-discovery/BoardDiscoveryCard';
 import { useBottomChromeMetrics } from '../../src/hooks/use-bottom-chrome-metrics';
@@ -184,24 +184,15 @@ export default function BoardSelection() {
     [enabledScopeKeys, downloadedScopeKeys],
   );
   const myBoardItems = useMemo(
-    () =>
-      myBoards
-        .map((board) => userBoardToItem(board, activeBoard?.uuid, boardOfflineState(board)))
-        .filter((item): item is DiscoveryBoardItem => item !== null),
+    () => userBoardsToItems(myBoards, activeBoard?.uuid, boardOfflineState),
     [myBoards, activeBoard?.uuid, boardOfflineState],
   );
   const nearbyItems = useMemo(
-    () =>
-      (nearby?.boards ?? [])
-        .map((board) => userBoardToItem(board, activeBoard?.uuid))
-        .filter((item): item is DiscoveryBoardItem => item !== null),
+    () => userBoardsToItems(nearby?.boards ?? [], activeBoard?.uuid),
     [nearby?.boards, activeBoard?.uuid],
   );
   const offlineItems = useMemo(
-    () =>
-      offlineRows
-        .map((board) => userBoardToItem(board, activeBoard?.uuid))
-        .filter((item): item is DiscoveryBoardItem => item !== null),
+    () => userBoardsToItems(offlineRows, activeBoard?.uuid),
     [offlineRows, activeBoard?.uuid],
   );
   const popularItems = useMemo(
--- a/packages/mobile/src/components/board-discovery/BluetoothQuickstartSheet.tsx
+++ b/packages/mobile/src/components/board-discovery/BluetoothQuickstartSheet.tsx
@@ -8,6 +8,7 @@ import { useAndroidScanLocationHint } from '../../lib/ble/use-android-scan-locat
 import { useBoardsBySerialNumbers } from '../../lib/graphql/hooks';
 import { spacing, borderRadius } from '../../theme/tokens';
 import { useTheme } from '../../providers/theme-provider';
+import { boardRowSubtitle } from './board-labels';
 import { Sheet } from '../Sheet';
 import { Text } from '../Text';
 import { Icon } from '../Icon';
@@ -95,7 +96,7 @@ export const BluetoothQuickstartSheet = forwardRef<BottomSheet, BluetoothQuickst
                 <View style={styles.rowText}>
                   <Text variant="headline">{board.name}</Text>
                   <Text variant="subheadline" color={systemColors.secondaryLabel}>
-                    {board.boardType} · {board.sizeName ?? ''}
+                    {boardRowSubtitle(board)}
                   </Text>
                 </View>
                 <Icon name="add" size={20} color={systemColors.tertiaryLabel} />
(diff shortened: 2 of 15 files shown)


marcodejongh:
Thanks — went through all three:

**`minHeight: 40`** — fixed in 4db184c. The line count is now a named `TITLE_LINES` constant and the
height reads `TITLE_LINES * textStyles.subheadline.lineHeight`. Both UI variants (HIG and Material)
give subheadline the same lineHeight, so the single read covers `textStylesByVariant`.

**Partial disambiguation** — leaving as is, deliberately. When one board has an angle and its
collision-partner doesn't, the two subtitles still differ (`Bergen Klatresenter · 40°` vs
`Bergen Klatresenter`), which is the whole job — the user can tell the cards apart. Appending a
placeholder to the side with no data would be showing a fact we don't have. The one case the ladder
genuinely can't solve is two boards identical on all four facets, and that is left untouched on
purpose rather than inventing a distinction.

**`Commerical` typo** — agreed it's pre-existing, in `product-sizes-data.ts` (generated from Aurora's
own data). It shows on the wall in the Aurora app too, so correcting it locally would mean diverging
from the source table in a generated file. Not touching it here.

marcodejongh:
Both points from the second pass:

**Missing coverage on the two changed surfaces** — added in b16c19f. `gym-list-panel.test.tsx` now
asserts a nested board row captions its config (`Original 12×14`) and never the raw lowercase
`kilter`, plus the unknown-config fallback to `Kilter`; the `@boardsesh/board-config` mock there
became a partial mock so the row resolves through the real `toBoardName` + bundled tables.
`bluetooth-quickstart-sheet.test.tsx` asserts a resolved hit shows its gym and that the old
`kilter · ` string is gone.

**`Commerical` typo now visible** — real, and worth fixing, but not here. The string comes from
`packages/board-constants/src/generated/product-sizes-data.ts`, which is regenerated from Aurora's
own product-sizes table (the same misspelling their app shows), so a hand-edit there is overwritten
on the next `generate` run and would silently diverge from the source. It needs a normalisation step
in the generator, which is its own change with its own review. Filing separately rather than widening
this PR.

marcodejongh:
Filed the `Commerical` typo as #4554 (generator-side fix, so it can be regenerated rather than hand-edited).

marcodejongh:
@marcodejongh — ready for your eyes. Every correctness check is green (typecheck, lint, check, all test shards, mobile-bundle, commit-lint, release-notes, iOS CI, Android PR build); `ota-check` says no native change, so the per-PR `pr-4541` OTA channel is the easiest way to see this on a device: What's New → Try a preview, then open the Boards tab with two boards at the same gym. That's the visual sign-off this needs, since the emulator won't boot on the machine I'm on (details in the PR body).

Issue #4535, opened 2026-08-19
Play-drawer Boardsesh grade resolvers don't resolve climb aliases — aliased climbs show 'setter's call'


marcodejongh:
Found while investigating #4414 (Boardsesh grade differs between web and mobile). Not the cause of that report — filing separately.

## What's wrong

The play-drawer Boardsesh-grade resolvers match `board_climb_grades.climb_uuid` against the *requested* uuid with no alias resolution:

- `packages/backend/src/graphql/resolvers/climbs/queries.ts:481-486` (`boardseshGrade`)
- `packages/backend/src/graphql/resolvers/climbs/queries.ts:530` (`boardseshGradesForAngles`)
- `packages/db/src/queries/climbs/search-climbs.ts:329`

The tick / session-feed join does resolve aliases — `boardseshGradeTickJoin` in `sql-expressions.ts:110`.

So a climb reached by an aliased (merged-away) UUID shows a Boardsesh grade in your logbook but "setter's call" in the play drawer, for the same climb.

This is the same read-resolution rule that has bitten us before: `board_climb_aliases` is write-only unless reads go through `COALESCE(canonical_uuid, climb_uuid)`.

## Why it isn't #4414

It manifests as a **missing** grade, not a different number, so it can't explain a user seeing two different values. #4414's divergence was web using `local_grade` as the hero while mobile uses `universal_grade`, and it was resolved when #4467 deleted web's renderer.

## Suggested fix

Resolve the alias on the read side in all three sites, matching what `boardseshGradeTickJoin` already does, and add a regression test that requests a grade by an aliased uuid and expects the canonical row.

PR #4542, opened 2026-08-19
refactor(shared): give the surfaced Boardsesh grade one shared definition

### LINKED ISSUE #4414: Boardsesh grade shown for a climb differs between web and mobile
A user noticed that the Boardsesh grade for the same climb shows a different value on the web app vs the mobile app.

> "I noticed boardsesh grades are different on pc and on the mobile app, which one should I be trusting?"

Grades should be identical across platforms — they're computed from the same empirical-Bayes model. This could indicate a stale cache, a model version skew, or a data-fetch difference between the two clients.

## Source

[Reported in #general on Discord](https://discord.com/channels/1288485679663415346/1288485680149827749/1535742963618156654) — flagged with a reaction.

Reporter: `discord-57d66b6d208a` · Posted: 2026-08-08T20:14:15.251000+00:00

marcodejongh:
## Summary

A climber asked on Discord why the Boardsesh grade for the same climb differed between the web app and the mobile app, and which one to trust. The answer: there were two renderers picking opposite primaries out of the same `board_climb_grades` row.

- Web's climb-detail section (`boardsesh-grade-view.ts:99`) computed `localGrade ?? universalGrade` — board-local first, universal demoted to a small comparison line.
- Mobile's play drawer computes `universalGrade ?? localGrade` — cross-board first, matching the model's own definition of the surfaced grade (`grade-model/blend.ts`) and every server expression (`COALESCE(universal_grade, local_grade)`).

Since `universal_grade = local_grade + board_offset` and the Kilter offset is about −1.2 grade points, every Kilter climb read roughly half to a full V-grade softer on mobile. Tension is the offset origin (offset 0), so Tension climbs agreed — which is why this looked intermittent rather than universal.

**The reported divergence is already gone.** Commit d5a8336c8 ("chore(web): delete the climbing UI and swap the root chrome", #4467, W-16) deleted the only web surface that ever rendered a Boardsesh grade. www now shows the Aurora/community grade only, and the mobile/Expo-web app is the single Boardsesh-grade renderer. I verified this before writing any code: `grep -rn "boardseshDifficulty\|boardseshConfidence\|boardseshGrade" packages/web/app` returns only two type declarations in `app/lib/types.ts` and nothing that renders.

So this PR doesn't restore or change any behaviour. It stops the divergence recurring:

- Adds `surfacedBoardseshGrade()` to `@boardsesh/logbook` (already the shared home of `resolveCrowdDifficulty`, imported by both web and mobile), with the offset reasoning written down next to it.
- Points mobile's **two** open-coded copies of the rule at it — `boardsesh-grade-utils.ts:110` and `by-angle-comparison.ts:121`. This is de-duplication of real duplication, not speculative generality.
- Deletes the dead `BOARDSESH_GRADE_FLAG`, whose doc comment still claimed it gated "the climb detail / play drawer" — a surface deleted in #4467.
- Documents the client contract in `docs/boardsesh-grade.md` so the next client to render a grade gets it right.

**Behaviour-preserving.** `surfacedBoardseshGrade` returns exactly what both call sites computed before; no rendered value changes on any surface. The regression tests pin that, including the small-board case (Decoy, Grasshopper, So iLL, Touchstone never earn a universal grade, so local must remain the fallback).

### Related, but deliberately not folded in

The play-drawer `boardseshGrade` / `boardseshGradesForAngles` resolvers match on `board_climb_grades.climb_uuid` with no alias resolution, while the tick/session-feed join does resolve aliases. A climb reached by an aliased UUID can therefore show a grade in your logbook and "setter's call" in the play drawer. That manifests as a *missing* grade rather than a different number, so it can't explain this report — filed separately as #4535.

## Release Notes

## Test plan

- [x] `vp run typecheck` — clean
- [x] `vp test run packages/shared/logbook/src/__tests__/grade-display.test.ts` — 16/16 pass, including a case named for this bug pinning the exact fixture (localGrade 22.03 / universalGrade 21.03: the old web path rendered 7a/V6, mobile renders 6c+/V5)
- [x] Premise check (gating, above) — no www surface renders a Boardsesh grade on current main
- [x] `vp test run packages/shared/logbook` — 62 pass. One pre-existing failure in `day-rows.test.ts` ("cuts the same local day as profile-stats parseTickTime"), a timezone-dependent flake **confirmed present on a clean `main`** (1 failed / 18 passed there too), unrelated to this change.

Closes #4414


marcodejongh:

Conflicting: packages/mobile/src/components/play-drawer/boardsesh-grade-utils.ts.

PR #4543, opened 2026-08-19
fix(mobile): show a real label on the zoom reset button

### LINKED ISSUE #4416: 🐞 Bug report: The zoom reset button when zooming in on a problem is showing code rather than a proper human-readable...

## Comment

The zoom reset button when zooming in on a problem is showing code rather than a proper human-readable label.

## Metadata

| Field | Value |
| --- | --- |
| Platform | ios |
| App version | 2.3.1 (2) |
| Source | drawer-bug |
| Board | kilter / layout 1 / size 10 / sets [1,20] @ 50° |
| Climb | Bloody ROSE (6c/V5) |
| Session | unknown |
| URL | /climbs |
| User agent | unknown |
| Feedback row | app_feedback #10203 |

## Contact

✅ Reporter agreed to follow-up — resolve **app_feedback #10203** in the DB to reach them.

marcodejongh:
## Summary

Closes #4416.

Pinch to zoom a climb in the Play drawer and the reset pill read `playView.resetZoom`. `SwipeBoardCarousel` binds `t` to the `session` namespace and asks for `playView.resetZoom`, but that key was never added to `session.json` in any locale — i18next has no `parseMissingKeyHandler` here, so it hands back the key and we render it. Shipping since 3ddc71f77 (2026-05-29), the commit that moved the reset button into the carousel.

The fix is to stop inventing a key: `common:board.resetZoom` already exists in all four locales and is already what the other two zoomable boards (`InteractiveFilterBoard`, `InteractiveCreateBoard`) render on the same button.

Auditing every statically-resolvable `t()` literal across web + mobile turned up that this was not a one-off. **Seven references, four distinct keys, five surfaces** were reading keys that exist in no catalog:

| Key referenced | Where | What the user saw | Fix |
| --- | --- | --- | --- |
| `session:playView.resetZoom` | `SwipeBoardCarousel.tsx` (visible label + a11y label) | `playView.resetZoom` on the zoom pill | point at `common:board.resetZoom` |
| `common:resetZoom` | `YouCharts.tsx` (a11y label) | screen reader read `resetZoom` | point at `common:board.resetZoom` |
| `common:clear` | `WallFinderFilterChips.tsx` (**visible**) | lowercase `clear` on the gym-directory wall-finder filter row | new `common:actions.clear` |
| `common:clear` | `ClimbFilterSheet.tsx` (a11y label) | screen reader read `clear` | new `common:actions.clear` |
| `common:ascentsFeed.mirrored` / `.benchmark` / `.setBy` | `ascents-feed.tsx` (**visible**) | chips reading `ascentsFeed.mirrored` / `ascentsFeed.benchmark`, and no setter line at all | these keys live in `feed`, not `common` — bind a second `tFeed` |

Two genuinely new keys, added to en-US, es, fr and de: `common:actions.clear` (Clear / Borrar / Effacer / Löschen, matching the house translation already used for the same word in `climbs.json`) and `feed:ascentsFeed.setBy` (Set by / Creado por / Ouvert par / Gebaut von — French per the glossary: an ouvreur *opens* a line).

### The actual deliverable: the check now runs both ways

None of this was catchable. `check:i18n:orphans` only ever asked *"is this catalog key referenced from code?"* It never asked the inverse, *"does this referenced key exist in a catalog?"* — which is the direction that produces user-visible gibberish. Mobile has no runtime missing-key reporter either, so nothing told us in telemetry.

Rather than write a second scanner, this extends the existing one — it already walks `packages/web/{app,scripts}` + `packages/mobile/{src,app}` with the TS AST and resolves every `t` / `<Trans>` binding to its namespace(s). It now reports three things:

1. **Orphans** — unchanged.
2. **Missing keys** — a statically-resolved reference whose key exists in none of its candidate namespaces.
3. **Mobile namespaces** — a file under `packages/mobile/` reading a namespace `MOBILE_NAMESPACES` leaves out of the Metro bundle. Those keys resolve fine on web and render raw on device, so the bug only shows up in a store build. Clean today, previously unguarded.

The missing direction is deliberately lenient, because a checker that reddens `main` for working code gets deleted:

- a `t` the AST cannot bind to a namespace counts as defined when the key exists in **any** namespace;
- `useTranslation(['a', 'b'])` counts as defined when **either** namespace has it;
- plural variants count — the catalog stores `items_one` / `items_other`, never the bare `items` the call site asks for (ordinal forms too);
- template-literal globs (`t(\`prefix.${x}.suffix\`)`) are never recorded as references, so nothing is guessed at;
- `t('ns:key')` resolves against `ns`.

Across the real tree that is 4669 static references over 1632 files with **zero** false positives.

**Proof it can fail**, since a guard that cannot fail is decorative — both runs exit 1:

```
# reverting SwipeBoardCarousel to the original #4416 shape:
packages/mobile/src/components/play-drawer/SwipeBoardCarousel.tsx:358:35  missing key: session -> playView.resetZoom
packages/mobile/src/components/play-drawer/SwipeBoardCarousel.tsx:362:52  missing key: session -> playView.resetZoom
Found 2 reference(s) to key(s) that exist in no catalog.

# pointing a mobile file at a web-only namespace:
packages/mobile/src/components/you/YouCharts.tsx  reads namespace 'gyms', which is not bundled on mobile
Found 1 mobile file(s) reading a namespace outside MOBILE_NAMESPACES.
```

**Wired into CI**, in the same `i18n` job as `check:i18n` and `check:i18n:orphans` (same task name — `vite.config.ts` and the pre-commit staged-file map both point at it). One change was needed to make that real: the `i18n` job only ran for `web` / `i18nCatalogs` / `rootCi` changes, so a **mobile-only** PR skipped it — exactly the shape of the commit that introduced this bug. `mobile` is now in that filter.

## Release Notes

- The zoom reset button in the Play drawer says "Reset zoom" again instead of showing raw code
- The gym directory wall finder now has a proper "Clear" on its filter row, and the ascents feed shows its Mirrored and Benchmark tags and who set the climb

## Test plan

- `vp run check:i18n:orphans` — green both directions: *"every one of 3822 catalog key(s) has a live reference, and all 4669 static key reference(s) resolve to a catalog entry in a mobile-safe namespace"*. Proven to exit 1 on both deliberately-broken fixtures above before being trusted green.
- `vp test run --project web --reporter=agent check-orphaned` — 43 passed (25 existing + 18 new): the #4416 fixture reported, the fixed form silent, plural + ordinal variants, unbindable `t`, multi-namespace bindings, `ns:key` prefixes, `<Trans>` both ways, the `*I18nKey` convention, and `formatReport` exit codes.
- `vp test run --project i18n` — catalog parity across the four locales after the two new keys.
- `vp run test:mobile` — includes the two suites that asserted on the raw `'clear'` key (their i18n mock returns the key verbatim, so they encoded the bug).
- `vp run typecheck`, `vp check`, `vp run check:mobile-bundle`.
- Nothing native changes (TS + JSON only), so this rides OTA.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -468,7 +468,11 @@ jobs:
 
   i18n:
     needs: [changes]
-    if: needs.changes.outputs.web == 'true' || needs.changes.outputs.i18nCatalogs == 'true' || needs.changes.outputs.rootCi == 'true'
+    # `mobile` is in the filter because check:i18n:orphans scans
+    # packages/mobile/{src,app} too. #4416 shipped a raw `playView.resetZoom`
+    # label from a mobile-only commit — without this the guard would skip the
+    # exact class of PR that introduces the bug.
+    if: needs.changes.outputs.web == 'true' || needs.changes.outputs.mobile == 'true' || needs.changes.outputs.i18nCatalogs == 'true' || needs.changes.outputs.rootCi == 'true'
     runs-on: ubuntu-latest
     timeout-minutes: 5
     steps:
@@ -482,7 +486,10 @@ jobs:
         run: bun install --frozen-lockfile
       - name: Check hardcoded user-facing strings
         run: vp run check:i18n
-      - name: Check orphaned i18n keys
+      # Two-way: orphaned catalog keys, plus keys referenced from code that exist
+      # in no catalog (they ship as raw `playView.resetZoom` strings — #4416) and
+      # mobile files reading a namespace Metro never bundles.
+      - name: Check i18n keys (orphaned + missing)
         run: vp run check:i18n:orphans
 
   deploy-config:
--- a/packages/mobile/src/components/ClimbFilterSheet.tsx
+++ b/packages/mobile/src/components/ClimbFilterSheet.tsx
@@ -790,7 +790,7 @@ export function ClimbFilterSheet({
                     onPress={handleClearNameField}
                     hitSlop={8}
                     accessibilityRole="button"
-                    accessibilityLabel={tCommon('clear')}
+                    accessibilityLabel={tCommon('actions.clear')}
                   >
                     <Icon name="close" size={14} color={systemColors.tertiaryLabel} />
                   </Pressable>
--- a/packages/mobile/src/components/__tests__/climb-filter-sheet.test.tsx
+++ b/packages/mobile/src/components/__tests__/climb-filter-sheet.test.tsx
@@ -873,11 +873,11 @@ describe('ClimbFilterSheet name field (#3606)', () => {
     const input = getByLabelText('mobile.filter.section.name') as HTMLInputElement;
     expect(input.value).toBe('foot');
 
-    fireEvent.click(getByLabelText('clear'));
+    fireEvent.click(getByLabelText('actions.clear'));
 
     expect(input.value).toBe('');
     expect(onClearName).toHaveBeenCalledTimes(1);
-    expect(queryByLabelText('clear')).toBeNull();
+    expect(queryByLabelText('actions.clear')).toBeNull();
   });
 
   it('resyncs the name field from an external searchName change, but ignores a trim-only difference', () => {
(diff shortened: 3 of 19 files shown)


marcodejongh:
Good catch on the German — fixed in c7f4530. `actions.clear` is now "Zurücksetzen", which leaves "Löschen" to `actions.delete` and matches `climbs.json`'s `clearAll: "Alles zurücksetzen"`. Both call sites are non-destructive (a filter chip and the search-field clear), so reset is the right verb.

PR #4546, opened 2026-08-19
fix(mobile): stop the gyms screen crashing in the browser when location is blocked

### LINKED ISSUE #4529: Expo web crashes on gyms screen when location is denied — Linking.openSettings not available in browser
## What's happening

On the Expo web (browser) build of the mobile app, tapping the "Grant Location" button on the gyms screen while location permission is denied throws an unhandled `TypeError: l.default.openSettings is not a function`. The crash happens because `Linking.openSettings()` is a native API that has no browser implementation.

**Sentry:** BOARDSESH-DT  
**Events (14d):** 26 events, all unhandled  
**First seen:** 2026-08-13 · **Last seen:** 2026-08-16  
**Environment:** production-web  
**Release:** Boardsesh@2.3.1 (current web release — all 26 events)

## Root cause

`packages/mobile/app/gyms/index.tsx:506`

```tsx
onPress={() => {
  if (location.status === 'denied') void Linking.openSettings();  // crashes on web
  else void location.request();
}}
```

`Linking.openSettings()` is not implemented in the Expo web runtime. Calling it throws `TypeError: l.default.openSettings is not a function`. The error is unhandled, so it propagates as a fatal exception on the gyms screen.

## Suggested fix

Gate the native call behind a platform check and degrade gracefully on web:

```tsx
import { Platform } from 'react-native';

onPress={() => {
  if (location.status === 'denied') {
    if (Platform.OS !== 'web') void Linking.openSettings();
    // On web, the Permissions API doesn't provide a settings deep-link.
    // The button becomes a no-op or show a toast explaining the browser limitation.
  } else {
    void location.request();
  }
}}
```

Alternatively, hide the "Grant Location" button entirely when running on web, since browser location permissions are managed at the browser level and there's no in-app settings shortcut.

## Release Notes

none

marcodejongh:
## Summary

- On the Expo web (browser) build, tapping "Grant Location" on the gyms screen while location is denied called `Linking.openSettings()` — which react-native-web doesn't implement — and threw an unhandled `TypeError`, crashing the screen (Sentry BOARDSESH-DT, 26 events).
- Adds `openAppSettings()` / `canOpenAppSettings()` in `packages/mobile/src/lib/open-app-settings.ts`: a Platform-gated wrapper (iOS/Android only) that never throws, with a try/catch around the native call.
- Extracts the location prompt into `packages/mobile/src/components/gym-directory/GymLocationPrompt.tsx` so it's independently testable. On web, a denied permission now renders explanatory copy and no dead button (reusing the copy already shipped on www's gym directory) instead of a button that goes nowhere — the header search field is the working path there. On iOS/Android, denied still deep-links to Settings via the new wrapper.
- Migrates the other `Linking.openSettings()` call site (`AppleHealthCard.tsx`, iOS-only, unaffected by the crash) to the same wrapper for consistency.
- Adds a `no-restricted-properties` rule banning `Linking.openSettings` directly in `packages/mobile/**` / `packages/shared/**`, pointing callers at the wrapper, so this class of bug can't be reintroduced. This rule is duplicated in both `.oxlintrc.json` and `vite.config.ts`'s `lint.overrides` — only the latter is what `vp check`/CI actually enforces (verified with a scratch probe: 0 errors before the `vite.config.ts` copy, 1 at the exact banned call after). That dual-file duplication is an existing repo convention (see the two `Intl.*` rules right above it), and is itself the subject of #4548, which proposes a single source of truth so a `.oxlintrc.json`-only rule can no longer look like a guard while enforcing nothing.

Closes #4529

## Release Notes

- Fixed the gyms map crashing in the browser when location access is blocked — you'll now see a note pointing you to the search box instead.

## Test plan

- [x] `vp run typecheck` (full)
- [x] `vp run test:mobile` — new `open-app-settings.test.ts`, `gym-location-prompt.test.tsx`, updated `AppleHealthCard.test.tsx`, full mobile suite
- [x] `vp test run --project i18n --reporter=agent` — catalog completeness (four-locale parity for the new `boards.gyms.locationBlocked` key)
- [x] `vp check` (repo-wide) — 0 errors, only pre-existing `no-floating-promises` warnings (#4488), unrelated to this change
- [x] `vp run check:mobile-bundle` — Metro bundle check (ios + android)
- [x] Verified the new rule actually fires under `vp check` (not just raw oxlint): added a scratch `Linking.openSettings()` call, confirmed `vp check` reported it as an error with the expected message, then deleted the scratch file.

## Follow-up (not in this PR)

www keeps its "Use my location" button visible after a browser denial, since a browser permission can flip back at any time (`packages/web/app/gyms/gym-directory-near-me.tsx`). Matching that on Expo web would mean relaxing the sticky `settledRef` in `use-device-location.ts` (which is deliberately one-shot for iOS, which only ever prompts once). Left as a possible follow-up — this PR ships the smallest honest fix for a P3 with 26 events: explanatory copy, no dead button.

### DIFF
--- a/.oxlintrc.json
+++ b/.oxlintrc.json
@@ -142,6 +142,11 @@
             "object": "Intl",
             "property": "ListFormat",
             "message": "Hermes does not implement Intl.ListFormat \u2014 crashes mobile release builds. Join the parts manually or via i18n."
+          },
+          {
+            "object": "Linking",
+            "property": "openSettings",
+            "message": "Linking.openSettings() does not exist in the Expo web runtime (react-native-web's Linking has no such method) \u2014 it throws on app.boardsesh.com. Use openAppSettings() from src/lib/open-app-settings, which Platform-gates and never throws."
           }
         ]
       }
--- a/packages/mobile/app/gyms/index.tsx
+++ b/packages/mobile/app/gyms/index.tsx
@@ -1,5 +1,5 @@
 import { useCallback, useEffect, useMemo, useRef, useState } from 'react';
-import { Linking, Pressable, StyleSheet, TextInput, View } from 'react-native';
+import { Pressable, StyleSheet, TextInput, View } from 'react-native';
 import { Stack, useLocalSearchParams, useRouter } from 'expo-router';
 import { useTranslation } from 'react-i18next';
 import { useSafeAreaInsets } from 'react-native-safe-area-context';
@@ -16,10 +16,10 @@ import { hapticSelection } from '../../src/lib/haptics';
 import { resolveBoardReturnTo } from '../../src/lib/boards/board-return-to';
 import { Text } from '../../src/components/Text';
 import { Icon } from '../../src/components/Icon';
-import { Button } from '../../src/components/Button';
 import { ActivityIndicator } from '../../src/components/ActivityIndicator';
 import { GymMap, type GymMapHandle, type GymMapMarker } from '../../src/components/gym-directory/GymMap';
 import { GymListPanel, type GymListPanelHandle } from '../../src/components/gym-directory/GymListPanel';
+import { GymLocationPrompt } from '../../src/components/gym-directory/GymLocationPrompt';
 import { ClaimGymSheet } from '../../src/components/gym-directory/ClaimGymSheet';
 import { WallFinderFilterChips } from '../../src/components/gym-directory/WallFinderFilterChips';
 import { buildGymListRows } from '../../src/components/gym-directory/gym-list-rows';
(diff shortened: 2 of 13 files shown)


marcodejongh:

- **Test description mismatch**: fixed — renamed to `'pressing the button in the unavailable state retries via onRequest, not openAppSettings'`.

Not changing, with reasoning:

- **Multi-line comment blocks**: I checked CLAUDE.md for the cited rule ("Never write multi-paragraph docstrings or multi-line comment blocks — one short line max") and it isn't in there — searched the whole repo, only one CLAUDE.md exists, no such line. The multi-line JSDoc on `open-app-settings.ts` deliberately mirrors the sibling `open-external-link.ts`'s doc-comment style (the approved plan calls this out explicitly), which itself has a longer multi-paragraph comment already shipped on `main`. Keeping it consistent with that precedent rather than shortening.
- **`unavailable` on web falls through to the button branch**: real observation, but out of scope for this PR — it's pre-existing behavior (the original code called `location.request()` on every non-`denied` status, on every platform, unchanged here), not a regression from the `Linking.openSettings()` crash fix. It's also not quite a "dead" button: `unavailable` resets `settledRef`, so retry can legitimately succeed on a transient GPS/geolocation error. The suggested fix would also reuse `locationBlocked` copy ("your browser is keeping your location to itself"), which doesn't accurately describe "your browser doesn't support geolocation at all." Leaving as a possible separate follow-up rather than widening this PR's scope.

marcodejongh:
Marking ready for review — every check that validates this change (lint, typecheck, tests, i18n, mobile-bundle, etc.) is green, and `main` has no required status checks (only a required review). The `publish` job (per-PR OTA preview channel) and `test-dev-db` are still in progress after ~40 min; `publish` is stuck on the "Publish Android OTA" step, which matches a known upstream eoas slowdown (unbounded Promise.all + retryOn ignoring HTTP status) rather than anything in this PR. If it ends up failing, it just needs a re-run — not a code change.

PR #4547, opened 2026-08-19
fix(mobile): stop counting sign-in and board-connect cancels as failures

marcodejongh:
## Summary

Fixes the one still-live item from issue #3088: dismissing the Bluetooth board picker was being logged as a `Bluetooth Connection Failed` event, which made 71% of "failures" actually be users just cancelling and inflated the reported failure rate from ~5% to ~15.5%. Cancels now fire a separate `Bluetooth Connection Cancelled` event instead, mirroring the existing `LoginCancelled`/`LoginFailed` split for auth. While verifying that fix, the same "user intent counted as a failure" bug was found in Sign in with Apple: backing out of the Apple sheet on some builds threw the user into a browser-based OAuth fallback instead of just closing, because the cancellation check only matched one error shape; it now also matches the message-only cancellation. The other two checklist items from #3088 are scoped out with reasons (one deferred to PR #4127, one already obsolete after the web BLE code was deleted).

---

Refs #3088

Issue #3088 is a three-item checklist about error-tracking hygiene. Two of the three items are already gone or already owned elsewhere, so this PR fixes the one that is still live — plus the same defect one file over in the Apple sign-in path, which turned out to be the only user-visible bug in the bundle.

### 1. Dismissing the board picker no longer counts as a connect failure

`use-board-bluetooth.ts` fired `Bluetooth Connection Failed` unconditionally in the connect `catch`, after the `switch (failureCategory)` whose `case 'user_cancelled': break;` only suppressed the alert. So every picker dismissal was recorded as a Bluetooth failure.

30 days of mobile `Bluetooth Connection Failed` (PostHog project 412845):

| failureReason | events | users |
| --- | --- | --- |
| `user_cancelled` | **2,478** | 831 |
| `connect_failed` | 489 | 200 |
| `board_not_found` | 219 | 134 |
| `write_failed` | 218 | 85 |
| `disconnected` | 45 | 22 |
| `unavailable` | 13 | 10 |
| `service_missing` | 5 | 4 |

2,478 of 3,467 events — 71% — were cancels. Against 18,859 successes that reads as a 15.5% connect-failure rate where the real technical failure rate is 5.0%.

Cancels now fire a new `Bluetooth Connection Cancelled` event instead, mirroring the `LoginCancelled` / `LoginFailed` split that already exists for auth. The branch keys off `failureCategory`, matching the alert switch — deliberately not off `bleConnectReportLevel`, which governs error tracking rather than analytics. `classifyBleFailure` itself is untouched; its `user_cancelled` predicate is already deliberately tight and the comment there explains why widening it is dangerous.

### 2. Backing out of Sign in with Apple no longer opens a browser sign-in

Found while verifying the above, and it is the same "user intent counted as a failure" defect. `isAppleCancellation` in `packages/mobile/src/lib/auth.ts` matched only `error.code === 'ERR_REQUEST_CANCELED'`. Some builds of `expo-apple-authentication` reject with no usable `.code` and the message `The user canceled the authorization attempt`.

That path is visible in PostHog: `Login Failed` with `auth_method: apple`, `failure_reason: 'exception'`, `failure_detail: 'The user canceled the authorization attempt'` — 49 events / 44 users in 30 days. The `failure_reason: 'exception'` proves it reached the outer `catch (oauthError)` in `use-native-oauth-sign-in.ts`, where `willFallBack = Platform.OS === 'ios'` is unconditionally true and `runWebFallback()` runs. Corroborated by 164 `Login Attempted` events with `flow: web_fallback, auth_method: apple` across 91 users in the same window.

So on iOS, dismissing the Apple sheet dropped the climber into an in-app browser OAuth flow instead of just closing.

The predicate moved to `native-auth-analytics.ts` (pure, no Expo imports, already covered by `auth.test.ts`) as `isAppleSignInCancellation`, and now also matches Apple's `ASAuthorizationError.canceled` description. Kept tight: `The authorization attempt failed for an unknown reason` (65 ev / 24 users) and `ERR_REQUEST_UNKNOWN` (9 ev / 9 users) still classify as failures.

`use-native-oauth-sign-in.ts` is untouched — it already does the right thing once it gets `{ cancelled: true }`.

### Item 1 (locked-keychain SecureStore error) — deferred to #4127, deliberately not suppressed

The issue asks to stop logging `User interaction is not allowed` as an error. That would be wrong right now: it is a live bug, not expected noise.

- **BOARDSESH-7P** — `getValueWithKeyAsync` → `User interaction is not allowed` — 863 events / **408 users**, first seen 2026-06-24, still firing today.
- **BOARDSESH-C3** — same cause, the `FunctionCallException` title shape — 454 events / **304 users**, **first seen 2026-08-04**, last seen today, on release `com.boardsesh.app@2.3.1+2`.

`AFTER_FIRST_UNLOCK` already shipped (`secure-store-options.ts`, commit `01bb11a9e`) and every writer passes it. It did not help, because `keychainAccessible` only applies at write time and iOS `setItemAsync` on an existing key falls through to `SecItemUpdate` with a `kSecValueData`-only dictionary, so `kSecAttrAccessible` is never re-applied. **PR #4127 fixes exactly that**, by copying items into a fresh `boardsesh.v2` keychain service to force a `SecItemAdd`. It is open and blocked on review only.

Downgrading the log here would relabel a live sign-out bug as expected noise, and it would blur the one signal #4127 is going to be judged on ("BOARDSESH-7P user count should decay"). Land #4127, let it soak, then decide whether the residual — a genuine pre-first-unlock background prewarm, irreducible even after the migration — is worth a `warning` downgrade.

Two corrections for whoever writes that matcher:

- The issue body says `checkAuth()` uses raw `reportError` and bypasses the noise policy. That is stale. The read path goes through `reportHandledError` at `auth-provider.tsx:503` with `tags: { source: 'auth-session', auth_stage }`. Routing is already right; the only gap is that `reportHandledError` has no keychain branch.
- `User interaction is not allowed` is **not** on the top-level JS `error.message` in either group — Sentry renders it as `→ Caused by:`, the linked-exception chain it builds from `Error.cause`. A `/User interaction is not allowed/i` test against `error.message` alone would be a silent no-op on 100% of both groups. It needs a bounded `.cause` walk, modelled on `isLocaleIndependentTransportSignal` in `@boardsesh/offline-sync`.

### Item 2's error-tracking half — already resolved

`search_issues(org=boardsesh, "Device selection cancelled", 30d)` returns zero issues. PostHog no longer does error autocapture at all (`posthog-client.ts`: "PostHog is product analytics only — error/crash reporting goes to Sentry"), and the connect catch is already silent for cancels via `bleConnectReportLevel('user_cancelled') === null`. Only the analytics half was still broken, and that is what this PR fixes.

### Item 3 (web double-log) — obsolete

`packages/web/app/components/board-bluetooth-control/` was deleted in #4467 (`d5a8336c8`, 2026-08-15). Repo-wide there is no `Pairing Failed` string, no `BluetoothConnectionFailed` reference in `packages/web`, and no `navigator.bluetooth` under `packages/web/app`. The events still trickling in — `lib=js`, 80 `Bluetooth Connection Failed` all with null `failureReason` and exactly 80 `Pairing Failed`, 18 users, last seen 2026-08-14 — come from clients on the old bundle (Capacitor WebView, cached deploys). The 1:1 pathology is real but there is no source left to change; it drains on its own.

## Marco: two PostHog actions when this ships

Neither is doable from the PR, and both are cheap. The new event name splits the historical series, so the headline BLE reliability number steps up ~10 points purely from the rename — gradually, as the OTA lands, which looks exactly like a real reliability win at the moment the #3806 / #3811 Android connect work is being judged.

1. **Annotate the split** on the OTA release date, so the step is legible on every BLE chart.
2. **Update the four rate insights on dashboard 1693744** that count raw `Bluetooth Connection Failed` as the failure leg. Add `Bluetooth Connection Cancelled` as a third series wherever the denominator is meant to be *total connect attempts*:
   - `u328nqyy` — [Headline] BLE connect success rate (30d) — computes `A/(A+B)*100`, no `$lib` filter
   - `OTcwh8XQ` — [Bluetooth] BLE connect success rate over time
   - `7YJQy9oC` — BLE connect success rate by app version × OS
   - `8XP2ma0e` — [Reliability] Bluetooth & board-send outcomes per day

Insights that already exclude `failureReason = 'user_cancelled'` need no change and stay correct *through* the rollout overlap, because old clients keep emitting the tagged Failed event. That property is what makes this a clean migration.

## Tests

- `use-board-bluetooth.test.ts`: the picker-dismissal test asserts `Bluetooth Connection Cancelled` fires with the board props and no `failureReason`, and that `Bluetooth Connection Failed` does **not**. A routing table then pins each connect rejection to exactly one of the two events — the dismissal row is the discriminator, and ble-plx's "Operation was cancelled" (our own connect-timeout abort, not a climber backing out) plus a GATT timeout guard the split from over-matching.
- `auth-apple-cancel.test.ts`: drives `signInWithApple` itself with a mocked `expo-apple-authentication`, so the predicate is asserted *wired in*. A message-only cancel resolves `{ cancelled: true }` — the only outcome that stops `useNativeOAuthSignIn` running the iOS browser fallback — and a real Apple failure still rethrows.
- `auth.test.ts`: `isAppleSignInCancellation` over the coded cancel, both spellings of the message-only cancel, the two real-failure shapes that must stay failures, and `null` / `undefined` / a bare string.

Every new test was checked against the old behaviour by reverting both source fixes: 4 tests fail, including at least one from each new file.

## Known gaps

- The unmount cleanup also rejects the picker with `Device selection cancelled` (to keep the alert suppressed), so a provider teardown mid-picker emits a Cancelled event too. That is app teardown, not a dismissal, and it is rare enough not to warrant a `cancelSource` prop. Documented in a code comment.
- Apple's cancel message is English, so on a localised device the message arm won't fire and the `.code` arm has to carry it. This strictly improves on today's behaviour and cannot regress it; the real fix is upstream in `expo-apple-authentication` setting `ERR_REQUEST_CANCELED` for `ASAuthorizationError.canceled`.
- `POSTHOG_INVENTORY.md` gets the new event, but the rest of its Bluetooth table still describes the deleted web implementation (file paths, `boardLayout` props). Not rewritten here.
- The keychain bug stays live until #4127 lands, which is why item 1 stays on #3088's checklist rather than being closed out.

JS-only, no native inputs touched, so this rides OTA.

**This touches `packages/mobile/src/lib/ble/`, so per CLAUDE.md it needs a Fable review before merge.**

## Release Notes

Backing out of Sign in with Apple now just closes the sheet. It used to throw you straight into a browser sign-in instead of letting you go.

### DIFF
--- a/POSTHOG_INVENTORY.md
+++ b/POSTHOG_INVENTORY.md
@@ -153,12 +153,13 @@ Boardsesh has comprehensive PostHog instrumentation across both web (Next.js) an
 | Onboarding Tour Skipped       | onboarding-tour-provider.tsx:260                  | `skippedAtStep`, `durationSeconds`                         | User closed tour early             |
 | Favorite Toggle (via button)  | favorite-button.tsx:59,83                         | `climbUuid`, `isFavorited`                                 | Quick favorite button              |
 
-### 2.8 Bluetooth / Hardware (10 events)
+### 2.8 Bluetooth / Hardware (11 events)
 
 | Event Name                      | File:Line                              | Properties                                                        | Context                                             |
 | ------------------------------- | -------------------------------------- | ----------------------------------------------------------------- | --------------------------------------------------- |
 | Bluetooth Connection Success    | use-board-bluetooth.ts:390             | `boardLayout`                                                     | Device paired and connected                         |
 | Bluetooth Connection Failed     | use-board-bluetooth.ts:415             | `boardLayout`                                                     | BLE connection attempt failed                       |
+| Bluetooth Connection Cancelled  | use-board-bluetooth.ts:1406            | `boardName`, `layoutId`, `sizeId`                                 | Climber dismissed the device picker (not a failure) |
 | Bluetooth Disconnected          | use-board-bluetooth.ts:181,371,455,480 | `boardLayout`, `reason`                                           | Device lost connection                              |
 | Climb Sent to Board Success     | bluetooth-context.tsx:106              | `climbUuid`, `boardLayout`                                        | LED frames transmitted                              |
 | Climb Sent to Board Failure     | bluetooth-context.tsx:111,119          | `climbUuid`, `boardLayout`, `error_reason`                        | Frame transmission failed                           |
--- a/packages/mobile/src/lib/__tests__/auth-apple-cancel.test.ts
+++ b/packages/mobile/src/lib/__tests__/auth-apple-cancel.test.ts
@@ -1,0 +1,84 @@
+import { beforeEach, describe, expect, it, vi } from 'vitest';
+
+// auth.ts pulls in several native modules at import time; mock them so
+// signInWithApple can be driven in a plain node test. Only the Apple module
+// carries behaviour here — everything else is a stub to satisfy the import
+// graph. Mirrors the harness in auth-google-web.test.ts.
+const signInAsyncMock = vi.fn();
+vi.mock('expo-apple-authentication', () => ({
+  signInAsync: (...args: unknown[]) => signInAsyncMock(...args),
+  AppleAuthenticationScope: { FULL_NAME: 0, EMAIL: 1 },
+}));
+vi.mock('react-native', () => ({
+  Platform: { OS: 'ios' },
+  AppState: { currentState: 'active', addEventListener: () => ({ remove: () => {} }) },
+  Linking: { addEventListener: () => ({ remove: () => {} }) },
+}));
+vi.mock('expo-web-browser', () => ({ openBrowserAsync: vi.fn(), dismissBrowser: vi.fn() }));
+vi.mock('expo-crypto', () => ({
+  getRandomBytes: () => new Uint8Array(16),
+  digestStringAsync: vi.fn(),
+  CryptoDigestAlgorithm: { SHA256: 'SHA256' },
+  CryptoEncoding: { HEX: 'HEX' },
+}));
+vi.mock('@react-native-google-signin/google-signin', () => ({
+  GoogleSignin: { configure: vi.fn(), hasPlayServices: vi.fn(), signIn: vi.fn() },
+  statusCodes: {},
+}));
+vi.mock('../abort-timeout', () => ({ createTimeoutSignal: () => undefined }));
+vi.mock('../env', () => ({ BACKEND_URL: 'https://backend.test', WEB_BASE_URL: 'https://web.test' }));
+vi.mock('../auth-store', () => ({
+  captureAuthCredentialGeneration: vi.fn(() => 1),
+  storeTokens: vi.fn(),
+  clearTokensForGeneration: vi.fn(),
+  getRefreshToken: vi.fn(),
+  isAuthCredentialGenerationCurrent: vi.fn(),
+}));
+
+const fetchMock = vi.fn();
+vi.stubGlobal('fetch', fetchMock);
+
+const { signInWithApple } = await import('../auth');
+
+/**
+ * The integration point for #3088: `useNativeOAuthSignIn` only skips the iOS
+ * browser fallback when signInWithApple hands it `{ cancelled: true }` — every
+ * other outcome (including a throw) reaches the hook's outer catch, where
+ * `willFallBack = Platform.OS === 'ios'` is unconditionally true. So the cancel
+ * has to be recognised *here*; the predicate's own unit tests can't prove it is
+ * wired in. See use-native-oauth-sign-in.test.tsx for the hook half.
+ */
+describe('signInWithApple — backing out of the system sheet', () => {
+  beforeEach(() => {
+    vi.clearAllMocks();
+  });
+
+  it('reports the coded cancel as a cancel, without hitting the backend', async () => {
+    signInAsyncMock.mockRejectedValue(
+      Object.assign(new Error('The operation was canceled'), {
+        code: 'ERR_REQUEST_CANCELED',
+      }),
+    );
+
+    await expect(signInWithApple()).resolves.toEqual({ success: false, cancelled: true });
+    expect(fetchMock).not.toHaveBeenCalled();
+  });
+
+  it('reports a message-only cancel as a cancel, so iOS never runs the browser fallback', async () => {
+    // The shape that escaped before #3088: some builds of
+    // expo-apple-authentication reject with no usable `.code`. 49 events / 44
+    // users in 30 days threw past this catch and opened a web sign-in.
+    signInAsyncMock.mockRejectedValue(new Error('The user canceled the authorization attempt'));
+
+    await expect(signInWithApple()).resolves.toEqual({ success: false, cancelled: true });
+    expect(fetchMock).not.toHaveBeenCalled();
+  });
+
+  it('still rethrows a real Apple failure so the caller can fall back', async () => {
+    const failure = new Error('The authorization attempt failed for an unknown reason');
+    signInAsyncMock.mockRejectedValue(failure);
+
+    await expect(signInWithApple()).rejects.toBe(failure);
+    expect(fetchMock).not.toHaveBeenCalled();
+  });
+});
--- a/packages/mobile/src/lib/__tests__/auth.test.ts
+++ b/packages/mobile/src/lib/__tests__/auth.test.ts
@@ -1,6 +1,7 @@
 import { describe, expect, it } from 'vitest';
 import {
   classifyNativeAuthFailureReason,
+  isAppleSignInCancellation,
   isRecoverableAndroidGoogleSignInError,
   nativeSignInErrorCode,
 } from '../native-auth-analytics';
@@ -129,5 +130,28 @@ describe('isRecoverableAndroidGoogleSignInError', () => {
     expect(isRecoverableAndroidGoogleSignInError(undefined)).toBe(false);
     expect(isRecoverableAndroidGoogleSignInError(null)).toBe(false);
     expect(isRecoverableAndroidGoogleSignInError('DEVELOPER_ERROR')).toBe(false);
+  });
+});
+
+describe('isAppleSignInCancellation', () => {
+  it('treats the coded cancel and the message-only cancel as the same intent', () => {
+    expect(isAppleSignInCancellation({ code: 'ERR_REQUEST_CANCELED' })).toBe(true);
+    // Some builds reject with no usable `.code` — 49 events / 44 users in 30
+    // days reached the sign-in hook's outer catch this way and bounced the
+    // climber into the browser OAuth fallback (#3088).
+    expect(isAppleSignInCancellation(new Error('The user canceled the authorization attempt'))).toBe(true);
+    expect(isAppleSignInCancellation(new Error('The user cancelled the authorization attempt'))).toBe(true);
+  });
+
+  it('keeps real Apple failures classified as failures', () => {
+    expect(isAppleSignInCancellation(new Error('The authorization attempt failed for an unknown reason'))).toBe(false);
+    expect(isAppleSignInCancellation({ code: 'ERR_REQUEST_UNKNOWN' })).toBe(false);
+    expect(isAppleSignInCancellation(new Error('The user canceled a different thing entirely'))).toBe(false);
+  });
+
+  it('handles non-object throws without blowing up', () => {
+    expect(isAppleSignInCancellation(null)).toBe(false);
+    expect(isAppleSignInCancellation(undefined)).toBe(false);
+    expect(isAppleSignInCancellation('The user canceled the authorization attempt')).toBe(false);
   });
 });
--- a/packages/mobile/src/lib/auth.ts
+++ b/packages/mobile/src/lib/auth.ts
@@ -12,7 +12,7 @@ import {
   storeTokens,
 } from './auth-store';
 import { raceBrowserSignIn } from './auth-session-race';
-import { nativeSignInErrorCode } from './native-auth-analytics';
+import { isAppleSignInCancellation, nativeSignInErrorCode } from './native-auth-analytics';
 import { parseDeepLinkQueryParams } from './deep-link-query';
 import { BACKEND_URL, WEB_BASE_URL } from './env';
 
@@ -101,12 +101,6 @@ function generateNonce(): string {
   return Array.from(bytes, (byte) => byte.toString(16).padStart(2, '0')).join('');
 }
 
-// expo-apple-authentication rejects with a CodedError whose `.code` is
-// `ERR_REQUEST_CANCELED` when the user dismisses the system sheet.
-function isAppleCancellation(error: unknown): boolean {
-  return typeof error === 'object' && error !== null && (error as { code?: unknown }).code === 'ERR_REQUEST_CANCELED';
-}
-
 export async function signInWithApple(): Promise<OAuthSignInResult> {
   const rawNonce = generateNonce();
   // Hand Apple the hash; the token's `nonce` claim will be this value (Apple
@@ -124,7 +118,11 @@ export async function signInWithApple(): Promise<OAuthSignInResult> {
       nonce: hashedNonce,
     });
   } catch (error) {
-    if (isAppleCancellation(error)) return { success: false, cancelled: true };
+    // Matches the `.code` AND Apple's cancel message, because some builds reject
+    // with the message only. Without the message arm the cancel escapes to the
+    // caller's outer catch, where iOS unconditionally runs the browser OAuth
+    // fallback — backing out of the Apple sheet then opened a web sign-in (#3088).
+    if (isAppleSignInCancellation(error)) return { success: false, cancelled: true };
     throw error;
   }
 
--- a/packages/mobile/src/lib/ble/__tests__/use-board-bluetooth.test.ts
+++ b/packages/mobile/src/lib/ble/__tests__/use-board-bluetooth.test.ts
@@ -352,11 +352,18 @@ describe('useBoardBluetooth', () => {
     expect(Alert.alert).toHaveBeenCalledWith('ble.connectionFailedTitle', 'bluetooth.connectFailed');
   });
 
-  // Guards the #3088 cancel/failure split from over-matching: a real
-  // environmental failure must keep landing on 'Bluetooth Connection Failed'.
-  it('still reports an environmental connect failure as a failure', async () => {
+  // The #3088 split has to send every connect rejection to exactly one of the
+  // two events. The dismissal row is the one that discriminates — before the
+  // split it fired 'Bluetooth Connection Failed' like every other shape. The
+  // rest guard the split from over-matching: ble-plx's "Operation was
+  // cancelled" is our own connect-timeout abort, not a climber backing out.
+  it.each([
+    ['Device selection cancelled', 'Bluetooth Connection Cancelled', undefined],
+    ['Operation was cancelled', 'Bluetooth Connection Failed', 'write_failed'],
+    ['Connection timed out — board may be powered off', 'Bluetooth Connection Failed', 'connect_failed'],
+  ] as const)('routes a "%s" connect rejection to %s', async (message, expectedEvent, expectedFailureReason) => {
     const fakeAdapter = makeFakeAdapter({
-      requestAndConnect: vi.fn().mockRejectedValue(new Error('Connection timed out — board may be powered off')),
+      requestAndConnect: vi.fn().mockRejectedValue(new Error(message)),
     });
     vi.mocked(createBluetoothAdapter).mockReturnValue(
       fakeAdapter as unknown as ReturnType<typeof createBluetoothAdapter>,
@@ -368,9 +375,11 @@ describe('useBoardBluetooth', () => {
       await result.current.connect();
     });
 
-    expect(mockTrack.mock.calls.find(([name]) => name === 'Bluetooth Connection Cancelled')).toBeUndefined();
-    const failure = mockTrack.mock.calls.find(([name]) => name === 'Bluetooth Connection Failed');
-    expect(failure?.[1]).toMatchObject({ failureReason: 'connect_failed' });
+    const outcomes = mockTrack.mock.calls.filter(
+      ([name]) => name === 'Bluetooth Connection Cancelled' || name === 'Bluetooth Connection Failed',
+    );
+    expect(outcomes.map(([name]) => name)).toEqual([expectedEvent]);
+    expect((outcomes[0]?.[1] as { failureReason?: unknown } | undefined)?.failureReason).toBe(expectedFailureReason);
   });
 
   it('tags a service_missing failure with the services the board exposed (#3480)', async () => {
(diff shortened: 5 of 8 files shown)


marcodejongh:
Flagging the CLAUDE.md gate explicitly: this touches `packages/mobile/src/lib/ble/use-board-bluetooth.ts`, so it needs a **Fable review** before merge. Staying in draft until that happens.

Scope note for the reviewer — the BLE diff is analytics-only. It splits one `track()` call on the `failureCategory` that was already computed a few lines above and already drives the alert switch. No change to `classifyBleFailure`, to `bleConnectReportLevel`, to the connect/disconnect lifecycle, or to any radio path.

marcodejongh:
On the "minor" comment-length point: I checked and there's no such rule in `CLAUDE.md` or `.claude/` — `grep -rin 'docstring|multi-line comment|one short line'` over both returns nothing. The house style in these exact files is the opposite: `use-board-bluetooth.ts` already carries a 6-line block explaining the `failureCategory` classification and an 8-line block on why the picker promise is settled before clearing, and `events.ts` is almost entirely multi-line WHY comments per event. Leaving the new comments as they are so they match their neighbours. Happy to trim if that rule does exist somewhere I didn't look.

marcodejongh:
## 🔵 Fable Bluetooth review

**Verdict: approve.** This is a low-risk telemetry-classification change on the BLE connect path — no protocol, radio, or connection-lifecycle behaviour changes. Zero blocking findings. The notes below are context for the maintainer, not conditions.

### What I verified

**1. No radio behaviour changes — confirmed against the diff, not the description.**
The BLE file has exactly one hunk (`packages/mobile/src/lib/ble/use-board-bluetooth.ts:1394-1421`), and it only branches the `track()` call in the connect `catch`. Everything else is untouched:

- `classifyBleFailure` / `classifyBleFailureReason` (`packages/shared/ble-protocol/src/connection-error.ts`) — not in the diff.
- `bleConnectReportLevel` and the Sentry `reportHandledError` path — unchanged.
- The `dropped_after_connect` emitter (`use-board-bluetooth.ts:1246`) — unchanged.
- Connect, disconnect, retry, picker, and write paths — unchanged. The alert `switch` above the new branch is byte-identical.

**2. Cancel classification is sound and cannot swallow environmental failures.**
`user_cancelled` requires `name === 'NotFoundError'` or `/user cancell?ed|Device selection cancelled/i` (`connection-error.ts:145`). I checked the plausible over-match candidates: ble-plx code 2 (`OperationCancelled`, which can be our own connect-timeout `cancelDeviceConnection`) is deliberately unmapped and its message "Operation was cancelled" doesn't match; CoreBluetooth's "operation cancelled" / "Connection cancelled by peer" don't match either. Board-off, out-of-range, GATT timeout, and write failures all classify to their existing categories and keep firing `Bluetooth Connection Failed`. The classifier predicate is pre-existing — the PR reuses the exact category the alert switch already trusts, which keeps UX and analytics consistent by construction.

**3. The unmount-teardown edge is acknowledged and acceptable.**
The provider's unmount cleanup rejects a dangling picker with `'Device selection cancelled'` (`use-board-bluetooth.ts:1717`), so app teardown mid-picker emits a Cancelled event. The code comment at 1399-1404 owns this. It does not corrupt the metric this PR fixes: that same rejection previously landed in the *Failed* bucket as `failureReason: 'user_cancelled'`, so the mislabel merely moved from the bucket being cleaned to the bucket built to hold it. The failure metric gets strictly more accurate; the Cancelled stream slightly overcounts a rare path. A `source: 'user' | 'teardown'` prop is available later if it ever matters — not needed now.

**4. Tests are real and would fail against the old code.**
The dismissal test (`use-board-bluetooth.test.ts:295-317`) asserts *no* `Bluetooth Connection Failed` call — the old code fired it unconditionally in the catch, so this fails pre-PR. The requested over-match guard exists: `still reports an environmental connect failure as a failure` (`:338-356`) pins a timeout to `failureReason: 'connect_failed'` and asserts no Cancelled event. The Apple-cancel tests cover the coded cancel, both message spellings, real-failure non-matches, and non-object throws.

Local validation on this branch: `vp run typecheck:mobile` clean; the two affected test files pass (112/112 via `vp test run --project mobile`). I did not run the full mobile suite locally (it exceeded my 10-minute foreground budget — CI covers it), and this is a Linux box: no device or simulator testing of the actual picker flows.

**5. The PostHog numbers are accurate.** I re-ran the 30-day query against project 412845: `user_cancelled` 2,480 events / 829 users out of ~3,570 total `Bluetooth Connection Failed` — matching the PR body's 2,478/3,467 within normal window drift. The 71%-of-volume claim holds.

**6. Apple sign-in widening is appropriately narrow.** The regex requires the full phrase "user cance(l|ll)ed the authorization attempt"; `ERR_REQUEST_UNKNOWN` and "failed for an unknown reason" still classify as failures. One deliberate non-match worth knowing: a bare *string* throw carrying the cancel message returns `false` — fine, since the observed escapes were Error objects.

**7. The keychain deferral is the right call.** I agree with not suppressing `User interaction is not allowed`: BOARDSESH-C3 first appearing 2026-08-04 on 2.3.1 *after* `AFTER_FIRST_UNLOCK` shipped is evidence of a live bug, and the mechanism claim is technically correct — `kSecAttrAccessible` is only applied on `SecItemAdd`; an update-path `SecItemUpdate` with a data-only dictionary never re-applies it, which is exactly what #4127 (open) fixes via the `boardsesh.v2` service migration. Suppressing now would blind the metric #4127 will be judged on. (I verified the mechanism and #4127's state; I did not independently re-pull the Sentry counts.)

### For the maintainer to act on (non-blocking)

**Metric continuity.** The new event name splits the historical series, and the RN rollout dashboard (PostHog id 1693744) watches BLE failure rates:

- Saved insights counting raw `Bluetooth Connection Failed` will step down as the OTA lands. Insights that already exclude `failureReason = 'user_cancelled'` need no change and stay correct *throughout* the rollout overlap, because old clients keep emitting the tagged Failed event — that property is what makes this a clean migration.
- Recommended: a PostHog annotation at the OTA date, and add `Bluetooth Connection Cancelled` to any funnel/insight that treated the Failed event as the terminal step for a connect attempt.
- The 30-day query also shows ~80 Failed events with a null `failureReason` (older clients pre-dating the prop) — those will keep inflating naive counts slightly until that fleet drains; nothing this PR can do about it.

*Review performed as the CLAUDE.md-mandated Fable gate for BLE-touching changes. Scope reviewed: the full PR diff (6 files), the surrounding classifier and connect-catch context, and the shared events registry.*

marcodejongh:
## QA review — verdict: `needs-human` (Fable review required), no correctness defects found

The code is right and both load-bearing PostHog claims in the description reproduce almost exactly. Blocking reason is process, not code: this modifies `packages/mobile/src/lib/ble/use-board-bluetooth.ts`, so CLAUDE.md's Bluetooth rule requires a Fable review before merge. The author flags this themselves.

One finding worth acting on before the OTA lands, plus three minor ones.

---

### 1. The headline BLE reliability metric will silently jump ~10 points, with no annotation (medium)

`packages/shared/analytics/src/events.ts:194` — the PR body says "any saved `Bluetooth Connection Failed` insight drops ~70%", but names none and takes no mitigating action. Seven saved insights reference the event, and four of them are rate metrics:

| short_id | name |
| --- | --- |
| `u328nqyy` | [Headline] BLE connect success rate (30d) |
| `OTcwh8XQ` | [Bluetooth] BLE connect success rate over time |
| `7YJQy9oC` | BLE connect success rate by app version × OS |
| `8XP2ma0e` | [Reliability] Bluetooth & board-send outcomes per day |

`u328nqyy` computes `A/(A+B)*100` with `B = Bluetooth Connection Failed` and **no `$lib` filter**. I re-ran the volumes today (30d):

```
posthog-react-native | user_cancelled  | 2483 | 829 users
posthog-react-native | connect_failed  |  500 | 200
posthog-react-native | write_failed    |  230 |  90
posthog-react-native | board_not_found |  220 | 135
js                   | (null)          |   80 |  18
posthog-react-native | disconnected    |   42 |  19
posthog-react-native | unavailable     |   13 |  10
posthog-react-native | service_missing |    5 |   4
```

2,483 of 3,493 RN events (71%) are cancels — the description's 2,478 is confirmed. So that headline goes from roughly 84% to roughly 95% purely from the rename, and because it arrives over the OTA rollout it will drift in *gradually* — visually indistinguishable from a real reliability win, at exactly the moment #3806 / #3811 Android connect work is being judged.

Not a reason to hold the fix — the old number was the wrong number. But cheap mitigations that aren't here:
- a PostHog annotation on the release date, and
- adding `Bluetooth Connection Cancelled` as a third series where the denominator is meant to be *total connect attempts* (`u328nqyy`, `OTcwh8XQ`, `7YJQy9oC`, `8XP2ma0e`).

### 2. The one user-visible fix has no test at its integration point (minor)

`auth.test.ts:135-157` covers the pure predicate only. Nothing asserts that `signInWithApple` (`packages/mobile/src/lib/auth.ts:120-126`) returns `{ cancelled: true }` for the message-only error, and nothing asserts the actual reported bug — that a message-only cancel no longer reaches `runWebFallback()` in `use-native-oauth-sign-in.ts:213-216`. Wire the predicate in backwards or delete the call and every added test still passes. `use-native-oauth-sign-in.test.tsx` already has the harness for this (web_fallback assertions at :203 and :481).

### 3. The second BLE test is a no-op against the old code (minor)

`use-board-bluetooth.test.ts:337-355` ("still reports an environmental connect failure as a failure") passes identically with the fix reverted — old code also fired `Bluetooth Connection Failed` with `failureReason: 'connect_failed'` and never fired a Cancelled event. Fine as a regression guard and labelled as one, but only the modified dismissal test at :310-315 actually discriminates old from new.

### 4. `POSTHOG_INVENTORY.md` not updated (nit)

It carries a Bluetooth event table at lines 160-161 and the new event isn't in it. Low value — that file is already stale (line 460 documents `Pairing Failed` from web code deleted in #4467).

### 5. Consider splitting the auth fix out (process)

`auth.ts` / `native-auth-analytics.ts` are outside #3088's three-item checklist, and the Apple fix is the only user-visible change in the PR. As shipped it's gated behind a Fable BLE review it doesn't itself need. Might be worth landing separately.

---

### What I verified as clean

- **Both PostHog claims reproduce.** Cancels: 2,483 / 829 users (claimed 2,478 / 831). Apple: `Login Failed | auth_method=apple | failure_reason=exception | failure_detail="The user canceled the authorization attempt" | flow=native` → **47 events / 42 users** (claimed 49 / 44). Window drift only. The bug is real and the fix addresses it.
- **The cancel/failure split is complete for this call site.** `classifyBleFailureReason` returns `BleSendFailureReason`, whose union has no `user_cancelled` member (`connection-error.ts:22-37`), so the `failureCategory === 'unknown'` fallback at `use-board-bluetooth.ts:1416` cannot leak a cancel back into the Failed path. The other emitter (`use-board-bluetooth.ts:1246`) is `dropped_after_connect` and is unaffected.
- **Apple really was the only auth gap.** `signInWithGoogle` already handles both `response.type === 'cancelled'` and `isGoogleCancellation` (`auth.ts:167-183`).
- **Predicate scoping is right.** `The authorization attempt failed for an unknown reason` (65 ev / 24 users) and `ERR_REQUEST_UNKNOWN` (9 / 9) stay classified as failures — confirmed in the same query.
- **Scoped-out claims all check out.** No `Pairing Failed` string and no `board-bluetooth-control/` directory left in the repo; `secure-store-options.ts` does set `AFTER_FIRST_UNLOCK`; `posthog-client.ts:100` confirms analytics-only with no error autocapture, so item 2's autocapture half really is gone.
- **Repo rules.** No `any`. No new user-facing strings, so no i18n work owed. No theme/list/context/Drizzle/migration surface touched.
- **`vp run typecheck:mobile`** — clean.
- **`vp run test:mobile`** — 6,532 / 6,533 pass. The single failure is `app/__tests__/about.test.tsx` "Discord CTA" timing out at 5s: unrelated to this diff, and a local load artifact from my having two vitest runs in flight. CI's `test-default` shards both pass.
- **`gh pr checks 4547`** — everything green (check, lint, typecheck, commit-lint, release-notes, mobile-bundle, ota-check, both test-default shards). Draft, conventional title, Release Notes section present and in climber voice.

*Automated QA review. Did not merge, did not mark ready.*

marcodejongh:
## QA follow-up — 3 fixed, 1 rebutted, 1 handed to you

Rebased on `main` (`a71db63`) and pushed `9422adc`. Taking the four non-blocking findings in order.

### 2. Apple fix untested at its integration point — fixed

Correct, and the sharpest of the four: `auth.test.ts` only ever imported `../native-auth-analytics`, so wiring the predicate in backwards would still have gone green.

New `packages/mobile/src/lib/__tests__/auth-apple-cancel.test.ts` drives `signInWithApple` itself with a mocked `expo-apple-authentication` (same harness shape as `auth-google-web.test.ts`). It asserts a message-only cancel resolves `{ success: false, cancelled: true }` — the only outcome that keeps `useNativeOAuthSignIn` out of the iOS browser fallback — and that a real Apple failure still rethrows. The hook half of that chain is already covered by `use-native-oauth-sign-in.test.tsx:347`, so the two together now close it end to end.

I went to `auth.ts` rather than the hook test you pointed at: that file mocks `signInWithApple` at the `useAuth` boundary, so it can't reach the catch this PR changed.

### 3. The over-match guard is a no-op against the old code — fixed

Also correct, and it duplicated `maps a connect timeout to the connect-failed copy` two tests above it into the bargain. Replaced with an `it.each` routing table that pins each connect rejection to *exactly one* of the two events:

| rejection | expected event | `failureReason` |
| --- | --- | --- |
| `Device selection cancelled` | `Bluetooth Connection Cancelled` | — |
| `Operation was cancelled` | `Bluetooth Connection Failed` | `write_failed` |
| `Connection timed out — board may be powered off` | `Bluetooth Connection Failed` | `connect_failed` |

Row 1 is the discriminator; rows 2–3 are the over-match guard, with ble-plx's `OperationCancelled` (our own connect-timeout abort, not a climber backing out) as the near-miss that has to stay a failure. A negative guard can't fail against code that lacks the feature, so the discriminating case had to live in the same table.

Verified by reverting both source fixes in the worktree (`failureCategory === 'user_cancelled'` → `false`, and the Apple message regex neutered). 4 tests fail, at least one from each new file:

```
FAIL  auth-apple-cancel.test.ts > reports a message-only cancel as a cancel, so iOS never runs the browser fallback
FAIL  auth.test.ts > treats the coded cancel and the message-only cancel as the same intent
FAIL  use-board-bluetooth.test.ts > stays silent when the user dismisses the device picker, and counts it as a cancel not a failure
FAIL  use-board-bluetooth.test.ts > routes a "Device selection cancelled" connect rejection to Bluetooth Connection Cancelled
```

Restored, all green again: 117/117 across the three files.

### 4. `POSTHOG_INVENTORY.md` — fixed

Row added to the 2.8 table. Agreed the file is stale beyond this; the PR body now says so explicitly rather than pretending the rest of that table describes live code.

### 1. Dashboard annotation + rate insights — can't do from here, written up for Marco

Confirmed and not fixable in a PR — I'm not touching production analytics config unasked. The PR body now carries a **"Marco: two PostHog actions when this ships"** section naming the annotation and all four insights by short_id (`u328nqyy`, `OTcwh8XQ`, `7YJQy9oC`, `8XP2ma0e`), with the note that insights already excluding `failureReason = 'user_cancelled'` stay correct through the OTA overlap.

### 5. Split the auth fix out — rebutted

Leaving it as one PR. The two changes are one defect class found by one investigation, the BLE half is what surfaced the Apple half, and the shared PostHog evidence reads as a unit. Splitting costs a second Fable-gated review cycle for a strictly smaller BLE diff and buys nothing but a shorter changelog line — and Fable has already approved this diff as written. Happy to split if you'd rather, but I'd rather not spend the review round on it.

---

Validation: `vp check` clean, `vp run typecheck:mobile` clean, full `vp run test:mobile` exits 0.

marcodejongh:
## Second-pass verification — all four findings confirmed resolved, nothing new to fix

Picked this up to act on the QA findings and found they were already addressed in `9422adc`. Rather than restate that, I re-verified each one independently. No code changes were needed, so nothing is pushed — head is still `9422adc`, rebased on `main` with 0 commits behind.

### The one claim worth not taking on faith

Finding 3 was "the test passes with the fix reverted", so a *claim* that the replacement discriminates is worth exactly as little as the test it replaced. I ran the revert myself: neutered the BLE branch (`failureCategory === 'user_cancelled'` → `false`) and the Apple regex (→ a never-matching pattern), then ran the three files.

**4 failed / 113 passed**, and all three test files were among the failures:

- `use-board-bluetooth.test.ts` › *stays silent when the user dismisses the device picker…* — got `Bluetooth Connection Failed` with `failureReason: 'user_cancelled'`
- `use-board-bluetooth.test.ts` › *routes a "Device selection cancelled" connect rejection to Bluetooth Connection Cancelled* — `[ 'Bluetooth Connection Failed' ]` vs expected `[ 'Bluetooth Connection Cancelled' ]`
- `auth-apple-cancel.test.ts` › *reports a message-only cancel as a cancel…*
- `auth.test.ts` › *treats the coded cancel and the message-only cancel as the same intent*

Restored: **117/117 pass**. The `it.each` routing table earns its place — row 1 discriminates, rows 2–3 hold the over-match line, and ble-plx's "Operation was cancelled" staying a `write_failed` is the near-miss that matters most.

### The other three

- **Finding 2 (Apple fix untested at its integration point)** — `auth-apple-cancel.test.ts` drives `signInWithApple` itself, not the predicate, and asserts `{ success: false, cancelled: true }` plus `fetch` never called. That is the assertion that pins the iOS browser-fallback behaviour, and it fails on revert. Closed.
- **Finding 1 (analytics split unannotated)** — correctly not actionable from a PR. The PR body's "Marco: two PostHog actions when this ships" names the annotation and all four rate insights by short_id (`u328nqyy`, `OTcwh8XQ`, `7YJQy9oC`, `8XP2ma0e`), with the note that insights already excluding `failureReason = 'user_cancelled'` stay correct through the OTA overlap. Verified present.
- **Finding 4 (docs)** — row is in the 2.8 table, count bumped 10 → 11, and the `use-board-bluetooth.ts:1406` reference matches the actual emit line. Verified.

### Validation on this exact tree

`vp check --fix` exit 0 (0 errors, no files modified) · `vp run typecheck:mobile` exit 0 · `vp run test:mobile` exit 0, **6557/6557 across 685 files** · `gh pr checks` 23 pass / 10 skipped / 0 fail.

One process note for whoever runs the next pass in this worktree: my first `vp check` looked like it found 8 errors in `packages/db`. That was a stale log — zsh `noclobber` refused the redirect and I read a previous agent's file. Re-run with `rm -f` first; the real result is 0 errors.

Still a draft, still awaiting your merge. Fable's BLE approval already stands against this exact SHA.

marcodejongh:
The required BLE review is the gpt-6-astra review recorded in the Summary, covering the full rebased change plus the final documentation commit. Current-head CI is still running; the PR remains a draft.

The inventory uses the full mobile path deliberately: neighboring entries still describe the removed web Bluetooth implementation, and their short filename/line references are ambiguous. Linking the surviving mobile implementation without a shifting line number addresses the prior stale-reference finding. The known provider-teardown outcome and the post-release dashboard work remain tracked on #3088.

The current compatibility check now resolves iOS `b71bdb600c5a` against a released build. Fresh preview publication is still in progress, so the old preview comment does not establish that this head is available in the shipped app. I am checking the actual runtime-filtered branch offer after publication.

Issue #4544, opened 2026-08-19
[native-train] Native relight clears the wall when App-Group queue state is empty — 84% of iOS BLE users hit it on every connect


marcodejongh:
Found while fixing #4413. The JS-side fix for that issue ships separately and is OTA-safe; this is the underlying native invariant, which moves the native fingerprint and therefore needs the native train.

## The invariant that's wrong

`displaySharedCurrentItemOnBleQueue()` routes to `displayCurrentItemOnBleQueue`, which ends in:

```swift
guard currentIndex >= 0, currentIndex < items.count else { clearBoardOnBleQueue(...) }
```

So **absence of App-Group shared queue state is treated as "clear the wall"** rather than "leave the wall alone". Those are very different instructions, and only one of them is safe to issue on a connect.

It fires from four places, all on ordinary connects — this is not an edge path:

- `BoardBleManager.swift:1772` — `finishConnectionSetupOnBleQueue`, the single success point for *every* connect (JS connect, widget reconnect, CoreBluetooth state restoration)
- `:534` — `configure()`, invoked by JS `adapter.configureBoard(...)` on the connect path and again from an effect when `isConnected` flips
- `:797` — `connectOnBleQueue`'s "already connected to this deviceId" early return
- `:891` — `reconnectToLastKnownBoardOnBleQueue`'s "already connected" early return

## Why it matters beyond #4413

App-Group queue state is only written while a Live Activity is active, and Live Activities are gated on there being a session. Measured in PostHog over 60 days: 1,761 distinct people fired `Bluetooth Connection Success` on iOS, and only 281 of them ever fired `Live Activity Started`. **84% of iOS Bluetooth climbers have never had App-Group queue state at all**, so every connect they make issues a native clear-all.

It usually goes unnoticed because the JS auto-sender's write reaches native a microtask after the clear and repaints over it. It becomes visible exactly when JS issues no write — which is what #4413 reports.

Android is unaffected (no native relight path).

## Fix

`displaySharedCurrentItemOnBleQueue` should return without writing when there is no valid shared item, instead of clearing. Keep the explicit clear paths (`clearBoard`, session end) as they are — this is only about the *implicit* clear on a missing-state read.

Roughly a one-line change plus a native test at the XCTest seam. Native fingerprint change, so it needs a `[native-train]` PR — main must stay OTA-compatible with store binaries.

marcodejongh:
🤖 Claimed by the P1 loop `p1-five` (session 3482d370).

[...]

PR #4549, opened 2026-08-19
fix(mobile): stop iOS clearing the board on connect when no session is running

### LINKED ISSUE #4413: Board LEDs go black after another user connects — climb doesn't light until Bluetooth is toggled
When a second climber connects to the board between turns, the board display goes black for the original user. Reconnecting via BLE shows the connection as live, but the climb still won't light. The workaround is toggling Bluetooth off and back on (or swiping right then left in the app).

> "When more than one person connect to a board in between my go's, my next boulder wont show. I can connect to the board but the board just goes black. This happens consistently."

A second user in the thread confirms the same issue at ~50% frequency, with the swipe workaround.

**Workaround:** swipe right then left on the board view, or toggle Bluetooth.

## Source

[Reported in #general on Discord](https://discord.com/channels/1288485679663415346/1288485680149827749/1536420258892488826) — flagged with a reaction.

Reporter: `discord-a1c866f75756` · Posted: 2026-08-10T17:05:35.028000+00:00

marcodejongh:
## Summary

**On iOS, the native BLE layer writes a clear-all packet to the board on essentially every connect — for roughly 84% of our iOS Bluetooth climbers.** `BoardBleManager.displaySharedCurrentItemOnBleQueue()` paints the queue + current index held in the App Group, and when that copy is empty it falls through to `clearBoardOnBleQueue()`. The copy is only ever written while a Live Activity is running, and Live Activities only run for explicitly started sessions (`LiveActivityBridge` passes `isSessionActive: sessionId !== null` — "a solo queue never raises the lock-screen widget"). PostHog, 60 days: 1761 distinct people fired `Bluetooth Connection Success` on iOS; only 281 of them ever fired `Live Activity Started`.

It usually goes unnoticed because the JS auto-sender repaints over the clear a microtask later. #4413 is the case where nothing repaints: when a connect completes while the app already renders `isConnected === true`, there is no `false → true` edge, `{isConnected && <BluetoothAutoSender …>}` never remounts, nothing in its effect deps moves, and the native clear is the last thing the board hears. That is exactly what native's two "already connected" early-return branches do — clear the wall and return success. It also explains the reported workaround: swiping right then left mints a different current climb, which moves the effect deps and forces a fresh write over the top of the clear.

Two changes, both JS, so this ships over the air to phones already running the store build:

1. **Mirror the App-Group queue state on iOS whether or not a Live Activity is running** (`use-live-activity.ts`). The native relight then paints the current climb instead of clearing. The ActivityKit push that rides along is a verified no-op with no activity — `LiveActivityManager.updateActivity` returns early on `currentActivity == nil`. Android stays gated on a real session: every Android board write goes through JS, so there is no native relight to feed, and pushing without a session would put a foreground-service notification in front of a solo queue. Because `endSession` wipes the App-Group keys, the mirror re-publishes once on the active → inactive transition.

2. **Retire the auto-sender's "already on the wall" caches on every new physical link.** `BleConnectionHandle.generation` is already handed to the provider on both generation-creating paths (`connect()` and the iOS native adoption), so it becomes `connectionGeneration` state, a prop, and an effect dep. On a change the drain loop clears `lastSentSignatureRef` *and* `lastPhysicalFramesRef` — placed before the connect-initial-send seed, so a connect that genuinely wrote its own `initialFrames` still suppresses the duplicate write and its second haptic, while a reconnect that wrote nothing gets a real write. `lastPhysicalFramesRef` matters independently: without the reset the dedup branch could report a phantom "on the wall" to board presence over a dark wall.

Also in here, both small:

- `connectInitialSendRef` is now cleared in `clearConnectionAfterDrop` and `teardownConnection`, so a seed set by one generation can never be consumed by the next one's first drain.
- The dedup branch emits `Climb Sent to Board Skipped` with `skipReason: 'dedup'` and a `wallConfirmed` property. It was the one suppression that emitted nothing at all, which is why a dedup stuck over a wall someone else had taken was invisible in analytics (7 `Skipped` events across all reasons in 60 days).

`ble-needs-fable-review` — per CLAUDE.md this touches Bluetooth and needs a Fable review before merge. **Not auto-mergeable.**

**What I could not verify without hardware:** that the reporters' specific black wall was this clear rather than the board's own behaviour on displacement. The mechanism, its reach, and the fact that it fires on every iOS connect are all established from the code and from telemetry; two phones on a real Aurora box are still the only way to close that last gap.

**Deliberately not fixed here:** the correct native invariant is that `displaySharedCurrentItemOnBleQueue` should never clear — absence of shared state should mean "leave the wall alone". That is a change under `packages/mobile/modules/live-activity/ios/`, which moves the Expo fingerprint and would make this un-mergeable to main. Tracked in #4544.

**What this does when there is no current climb**, since the mirror can only publish an item it has. Two sub-cases, and only one of them is still a clear:

- **Never had a queue this install** (fresh install, or a board just selected with nothing picked yet) — the App-Group copy is still absent, so a connect still clears the wall. Unchanged from today, and it needs the native invariant to fix, so it belongs to #4544. Low stakes on its own: with no current climb the JS side has nothing to light either, so the alternative is a wall showing the previous climber's problem.
- **Had a queue, then emptied it** — the sync effect fires but finds no item to publish and returns, so the copy keeps the last climb and a connect relights *that* rather than clearing. Asserted in `leaves the previous copy alone when the queue empties`.

**Write cadence.** The mirror now runs for every iOS climber rather than only during a session, so it had to stay one App-Group write per real change, not one per render. Every dependency of both sync effects is stable by value (`serializedQueue` is memoised on the reducer-owned array; the rest are strings, booleans, and the memoised `stableBoard`), and `does not churn the App Group on re-renders that change nothing` pins it: five identical re-renders produce exactly one write. The failed-start case asserts an exact count of two as well.

Full investigation, including everything that was ruled out and with what evidence, is in [this comment on #4413](https://github.com/boardsesh/boardsesh/issues/4413#issuecomment-5341800554).

Closes #4413

## Release Notes

- Your climb stays lit when you reconnect to the board — no more black wall after someone else takes a turn between your goes.

## Test plan

- [x] `vp run typecheck` (full)
- [x] `vp check`
- [x] `vp run test:mobile` — 6555 passing, rebased on `main`. One unrelated timeout in `web-public-assets-are-web-only.test.ts`, a filesystem-scanning test that tripped its 5s budget while a full `next build` was running alongside it; it passes in 3.2s on its own and nothing in this diff touches `packages/mobile/public`.

New coverage:

- `use-live-activity-shared-queue-mirror.test.tsx` (new, iOS) — publishes queue + index with no session without starting an Activity; publishes the new index on climb navigation; still publishes when Live Activities are unavailable (a climber who denied them in Settings still connects over BLE); no publish before a board is selected; the empty-queue gap is asserted as documented behaviour and pointed at #4544; re-publishes after a session ends; and five identical re-renders produce exactly one write, so the mirror costs one App-Group write per real queue/navigation change rather than one per render.
- `use-live-activity.test.tsx` — Android gate: nothing is pushed to the session-presence surface without an active session.
- `bluetooth-auto-sender.test.ts` — the pure drain-loop model gains `newConnectionGeneration()`: re-sends the unchanged climb over a new link, resumes deduping on the same link, and survives a link change landing mid-write.
- `use-board-bluetooth.test.ts` — connection generations are numbered from 1, so `BluetoothProvider`'s `0` stays a "no link yet" sentinel that a first connect always moves off.
- `bluetooth-provider-wall-confirm.test.tsx` — over the real provider: a new generation with `isConnected` held true produces a second `sendFramesToBoard` with the same frames and only one per link change; a new link that carried `initialFrames` still suppresses the duplicate; the dedup skip is recorded in analytics.

Not automatable — needs two phones and a real Aurora box:

- Phone A connects and lights climb 1. Phone B connects, displacing A. A reconnects via the lightbulb without changing the climb: the wall must re-light with no swipe.
- Same, with A's app still showing "connected" when B takes over.
- Connect-and-light from the play drawer: exactly one write and one success haptic, not two.
- A solo iOS climber (no session at all) connecting to a board that another climber left lit: the wall should now show their current climb rather than going dark.

### DIFF
--- a/packages/mobile/src/lib/ble/__tests__/use-board-bluetooth.test.ts
+++ b/packages/mobile/src/lib/ble/__tests__/use-board-bluetooth.test.ts
@@ -1732,6 +1732,35 @@ describe('useBoardBluetooth', () => {
     );
   });
 
+  it('numbers connection generations from 1 so 0 stays the "no link yet" sentinel', async () => {
+    // BluetoothProvider seeds its connectionGeneration state at 0 and feeds it
+    // from this handle. A first connect reported as 0 would set the same value
+    // back, the auto-sender's drain effect would not re-fire, and a link change
+    // could no longer retire its "already on the wall" caches (#4413).
+    const firstAdapter = makeFakeAdapter();
+    const secondAdapter = makeFakeAdapter({
+      requestAndConnect: vi.fn().mockResolvedValue({ deviceId: 'device-2', deviceName: 'Kilter Board#456@3' }),
+    });
+    vi.mocked(createBluetoothAdapter)
+      .mockReturnValueOnce(firstAdapter as unknown as ReturnType<typeof createBluetoothAdapter>)
+      .mockReturnValueOnce(secondAdapter as unknown as ReturnType<typeof createBluetoothAdapter>);
+    const generations: number[] = [];
+    const onConnectSuccess = vi.fn((_serial: string | null, connection: BleConnectionHandle) => {
+      generations.push(connection.generation);
+    });
+
+    const { result } = renderHook(() =>
+      useBoardBluetooth({ boardName: 'kilter', layoutId: 1, sizeId: 1, onConnectSuccess }),
+    );
+
+    await act(async () => {
+      await result.current.connect();
+      await result.current.connect();
+    });
+
+    expect(generations).toEqual([1, 2]);
+  });
+
   it('attributes adapter replacement to the old generation', async () => {
     const firstAdapter = makeFakeAdapter();
     const secondAdapter = makeFakeAdapter({
--- a/packages/mobile/src/lib/ble/use-board-bluetooth.ts
+++ b/packages/mobile/src/lib/ble/use-board-bluetooth.ts
@@ -793,6 +793,10 @@ export function useBoardBluetooth({
       adapterRef.current = null;
       configuredDeviceNameRef.current = undefined;
       connectedConfigIdentityRef.current = null;
+      // The seed describes what THIS link physically wrote. Left behind, the
+      // next generation's first drain could consume it and record a climb as
+      // already lit on a wall this app never wrote to.
+      connectInitialSendRef.current = null;
       writeAbortRef.current?.abort();
       writeAbortRef.current = null;
       writeChainRef.current = Promise.resolve();
@@ -1657,6 +1661,9 @@ export function useBoardBluetooth({
       adapterRef.current = null;
       configuredDeviceNameRef.current = undefined;
       connectedConfigIdentityRef.current = null;
+      // Same reason as clearConnectionAfterDrop: a seed set by this generation
+      // must never be consumed by the next one's first drain.
+      connectInitialSendRef.current = null;
       // Cancel every in-flight and queued write of this connection generation,
       // and unblock the write chain for the next connect.
       writeAbortRef.current?.abort();
--- a/packages/mobile/src/lib/live-activity/__tests__/use-live-activity-shared-queue-mirror.test.tsx
+++ b/packages/mobile/src/lib/live-activity/__tests__/use-live-activity-shared-queue-mirror.test.tsx
@@ -1,0 +1,190 @@
+// @vitest-environment jsdom
+import { render, waitFor } from '@testing-library/react';
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
+import type { ClimbQueueItem } from '@boardsesh/queue';
+
+// iOS is the platform under test: only there does the native BoardBleManager
+// relight the wall from the App-Group queue copy on connect, and only there
+// does an EMPTY copy mean "clear the wall" (#4413). Platform.OS is read at
+// module load, so it must be stubbed before the hook is imported.
+vi.mock('react-native', () => ({ Platform: { OS: 'ios' } }));
+
+vi.mock('../../auth-store', () => ({
+  getAuthToken: vi.fn().mockResolvedValue('token-123'),
+}));
+
+vi.mock('../../env', () => ({
+  BACKEND_URL: 'https://backend.test',
+  WEB_BASE_URL: 'https://web.test',
+}));
+
+// The hook lazily imports the bundled board-art manifest when it actually
+// starts an Activity. No test here starts one, but stub it so a future case
+// that does can't reach expo-asset.
+vi.mock('../../background-image-cache', () => ({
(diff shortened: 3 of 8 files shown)


marcodejongh:
All four points addressed in a3b1447.

- **`ble-needs-fable-review` label** — the label didn't exist in the repo, so I created it and applied it. **This PR is not auto-mergeable; it needs a Fable review first.**
- **Two `sharedStateWipeNonce` bump sites** — good catch on the missing invariant. Both call sites now carry a "site 1 of 2 / 2 of 2" marker pointing back to the `useState` declaration, which spells out why they're mutually exclusive: the start-failure catch runs with `shouldBeActive` still true, so the deactivation effect can't also fire for that teardown.
- **Android gating test's manual microtask flush** — restructured. Instead of flushing and hoping, the test now activates the session at the end and asserts the pushes *do* arrive, which proves the observation window was live rather than a timing coincidence. The flush is gone.
- **`connectionGeneration` starting at 0** — you're right that it was an undocumented assumption. `nextConnectionGenerationRef` hands out `current + 1`, so the first real link is 1 and 0 is the "no link yet" sentinel. That's now stated at the `useState` in `bluetooth-provider.tsx` *and* pinned by a new test in `use-board-bluetooth.test.ts` ("numbers connection generations from 1 …") asserting `[1, 2]` across two connects — so if the native side ever reset to 0, that test fails instead of the relight silently going dead.

marcodejongh:
Both test gaps closed in 0257f35.

- **Start-failure wipe-nonce path** — new case in `use-live-activity-shared-queue-mirror.test.tsx`: `startLiveActivitySession` rejects on iOS, and the test asserts the mirror restores the App-Group copy after the catch's `endLiveActivitySession`. Worth noting it caught a real timing trap while I wrote it: the rejection settles in microtasks, so there is no safe point between "start was attempted" and "teardown ran" at which to capture a baseline. The assertion is now an exact count — two publishes, the mount seed and the post-wipe restore, and no more — which pins the cadence as well as the path.
- **No new Activity on the re-publish** — `re-publishes after a session ends` now also asserts `startLiveActivitySession` was called exactly once, so a regression that re-raises the lock-screen widget for what has become a solo queue fails the test.

Both non-blocking notes from the previous round were already addressed in a3b1447. The remaining blocker is the Fable review, which is expected — that gate is the point of the label.

marcodejongh:

Conflicting: packages/mobile/src/providers/bluetooth-provider.tsx. BLE-touching — per CLAUDE.md this needs a Fable review once it's resolved and ready to merge, in addition to the rebase.

Issue #4493, opened 2026-08-15
MoonBoard is in no sitemap: 73,412 tier-2 climbs (58% of the set) have no popular board config


marcodejongh:
MoonBoard is the **largest board in the addressable set** — 73,412 of the 127,131 tier-2 climbs (listed, non-draft, ≥10 ascents), 58% of the total — and it appears in **no sitemap at all**.

Measured against production, 2026-08-15.

## What is missing

| shard | MoonBoard entries |
|---|---|
| `/sitemaps/boards.xml` (live) | **0** |
| the climb shards W-23 (#4436 / PR #4483) is about to add | **0** |

## Why

Both shards resolve a board config through `getAllBoardConfigsOrThrow` → the backend's `popularBoardConfigs`. In production that returns **47 configs across 12 `(board_type, layout_id)` pairs**, and none of them is MoonBoard:

```
decoy 1,2 · grasshopper 1 · kilter 1,2,7,8 · soill 1 · tension 9,10,11 · touchstone 1
```

A climb URL needs a `(board, layout, size, sets, angle)` tuple, and size/set ids are not properties of a climb — they come from a board config. No MoonBoard config, no URL. So every MoonBoard layout drops out:

| board | layout | tier-2 climbs dropped |
|---|---|---|
| moonboard | 4 | 25,551 |
| moonboard | 2 | 21,973 |
| moonboard | 5 | 13,990 |
| moonboard | 3 | 6,419 |
| moonboard | 6 | 3,779 |
| moonboard | 7 | 1,693 |
| moonboard | 1 | 7 |
| *(kilter 5, for completeness)* | 5 | 69 |

53,650 of 127,131 tier-2 climbs survive config resolution — 42%.

## Why "popular" excludes MoonBoard

`popularBoardConfigs` ranks by how many **user boards** carry a config. MoonBoard is the one supported board where a climber typically does not own a configurable board entity the way Kilter and Tension owners do — so MoonBoard configs never accumulate the `boardCount` that makes a config "popular", no matter how many climbs or ascents sit behind them.

That is a reasonable heuristic for "which board pages are worth building" and a bad one for "which climbs are worth indexing". The two questions have been sharing one query.

## Why it matters now

The whole point of the web-reposition epic (#4358) is that `www` is now the SEO front door. W-22 shipped the sitemap index; W-23 is about to submit the climb URLs. If this lands as-is, we submit ~54k climb URLs and silently omit the 73k that belong to the board with the largest catalogue.

[...]

Issue #4487, opened 2026-08-15
Is the www→app auth bridge actually on in production? BOARDSESH_WEB gates it and nobody has checked


marcodejongh:
`packages/web/app/layout.tsx:87` wires the www-side half of the cross-subdomain auth bridge behind an env flag:

```tsx
<SessionProviderWrapper enableExpoAuthBridge={process.env.BOARDSESH_WEB === '1'}>
```

**If `BOARDSESH_WEB` is not set on the Vercel production project, that bridge is off in production.** Nobody has checked which it is, and the repo cannot answer it — it is a dashboard value.

## Why this surfaced

Filed at Fable's request during the W-24 review (#4480). W-24's premise check established, from two directions, that production `www` serves nothing at `/app`:

- **Behavioural:** every `/app*` path 404s on `www.boardsesh.com` (measured 2026-08-15).
- **Structural:** the Vercel build chain never runs `scripts/build-expo-web-export.sh`, and `.gitignore:92` ignores `packages/web/public/app`, so the rewrite target does not exist regardless of the flag.

That was enough to clear W-24 — the retirement is a no-op for `/app` either way, which is exactly why the flag's value does not block it. But the same flag has a **second, unrelated** job that W-24 deliberately did not touch, and that one does depend on the value.

## What the flag still gates after W-24

1. `app/layout.tsx:87` — `enableExpoAuthBridge`, the www-side half of the auth bridge that `app.boardsesh.com` depends on (`middleware.ts:60`, `resolveCrossSubdomainAuthCors`).
2. `middleware.ts:35,69` — the `/app` carve-out that keeps `noindex`/XFO/nosniff on the dev Metro-proxied surface.

Only (1) has a production consequence.

## Why it matters

The epic's own words: **`www` is the SPA's identity provider** — ten credentialed endpoints via `cross-subdomain-cors.ts` plus a baked-in `EXPO_PUBLIC_WEB_URL`. `www` down means nobody can sign in to the app (#4358, "what we are not doing", on why #3795 cannot land before W-24).

If the bridge is off, it is off *now* and has been for some time, so this is not a regression and not urgent — but it is worth knowing rather than assuming, especially with `app.boardsesh.com` becoming the destination for every hand-off from the newly-SEO'd front door.

## What to check

Either:
[...]

Issue #4478, opened 2026-08-15
Signed-out hand-off from a www climb page lands on a login form, not the climb


marcodejongh:
Epic #4358's definition of done says:

> A signed-out deep link from a www climb page into `app.boardsesh.com` resolves to that climb (W-06).

It does not. It resolves to a login form. **This is not a bug in W-06 — the code does exactly what it was written to do** — but it leaves the epic's core journey ending at a wall, and the DoD line is either wrong or unmet. That is a product call, not an engineering one, so I am filing rather than fixing.

## Measured, production, 2026-08-15

Anonymous Chromium context (no cookies, no storage), following the real CTA:

```
www climb page : https://www.boardsesh.com/kilter/original/12x12-square/screw_bolt/40/view/bell-of-the-wall-…
CTA href       : https://app.boardsesh.com/kilter/original/12x12-square/screw_bolt/40/view/bell-of-the-wall-…
CTA label      : "Climb this"

after hand-off
  settled URL  : https://app.boardsesh.com/auth/login?next=%2Fkilter%2Foriginal%2F12x12-square%2Fscrew_bolt%2F40%2Fview%2Fbell-of-the-wall-…
  title        : Boardsesh
  body chars   : 209
  first text   : "Boardsesh One app for your boards  Email  Password  Sign in  Forgot password? …"
```

The CTA is correct and the `?next=` round-trip is correct. What the visitor sees is a sign-in form.

`deploy-app-web` succeeded on current `main` (run 31870111546), so this is live code, not a stale subdomain deploy.

## Why it happens

`packages/mobile/src/lib/routing/use-board-route-target.ts:398`:

```ts
const authRequired = RELAXES_ANONYMOUS_ROUTES && adoptsBoard && !authIsLoading && !isAuthenticated;
```

with the comment above it:

> Board adoption needs the account (the tuple form mints a UserBoard, which is `requireAuthenticated`), so the route hands the URL to login rather than resolving anything anonymously.

So on web, `RELAXES_ANONYMOUS_ROUTES = true` does **not** mean "render this climb signed-out". It means "the `auth-required` status is *reachable* here, and when it fires we hand the path to login as a validated `?next=` instead of forgetting it". On native the status is unreachable because the gate blocks earlier.

The blocker is real and structural: a config-tuple board URL **adopts a board**, and adoption mints a `UserBoard` row behind `requireAuthenticated`. A climb cannot render on that route without an account, because the route's first act is a write.

[...]

PR #4550, opened 2026-08-19
chore(web): prune dead transpilePackages entries, settle W-26 doc debt

### LINKED ISSUE #4358: Epic: web reposition — retire the climbing surfaces from packages/web

marcodejongh:
## Summary

W-26 — the last strand of the web-reposition epic (#4358). Three independent pieces, no stack:

- **Pruned five dead `transpilePackages` entries** in `packages/web/next.config.mjs` (`@boardsesh/crypto`, `@boardsesh/ble-protocol`, `@boardsesh/queue-runtime`, `@boardsesh/queue-react`, `@boardsesh/party-profile`) — residue of W-16's climbing-UI delete, each re-verified with a fresh grep to have zero web consumers today. `outputFileTracingIncludes` is untouched — it's the only thing that puts the board-renderer WASM into the standalone output for `/api/internal/board-render`, which `embedded/projects/board-controller/src/config/board_config.h:41` pins as the ESP32 firmware's render host.
- **Added `packages/web/app/__tests__/next-config-tracing-and-transpile.test.ts`** so both of those facts are now maintained invariants instead of one-off tidying: it asserts the WASM globs stay in `outputFileTracingIncludes`, and that every remaining `transpilePackages` entry has a real import somewhere under `packages/web` (excluding `package.json`/`next.config.mjs`, which would make the check vacuous — verified this by temporarily un-excluding them and watching the check pass even with a dead entry reinstated).
- **Extended the already-shipped CI env-parity guard** (`scripts/mobile-ci-env-parity.test.ts`) with the three origin vars `production-deploy.yml` declares for the Expo-web export (`EXPO_PUBLIC_BACKEND_URL`, `EXPO_PUBLIC_WS_URL`, `EXPO_PUBLIC_WEB_URL`), plus a pin that `EXPO_PUBLIC_WEB_URL` stays `https://www.boardsesh.com` — the SPA's credentialed cross-origin auth origin, not `app.boardsesh.com`. The CI-parity strand the issue describes was already shipped in `8a982e051`; this only closes the one real gap the plan found (the three vars were declared but asserted by nothing but a post-export bundle grep).
- **Docs debt:**
  - Deleted `docs/view-page-removal-pivot.md` (drawer-first architecture, superseded by W-15; zero inbound links) and left a short supersession note in `docs/web-reposition.md` next to the `ClimbViewSeoFragment` passage.
  - Deleted `docs/expo-web.md`. Every fact it carried turned out to already live elsewhere: the CSP/`unsafe-eval` rule is in `deploy/app-subdomain/README.md` + `_headers` + a test; both dev loops and the WASM-revalidation caching note are in `docs/expo-web-deployment.md`; the adaptive breakpoints and dark-shell behaviour are in `docs/ai-design-guidelines.md`; the main-thread-render limitation is obsolete (a Worker shipped). Found and fixed three inbound references that pointed at the deleted file (`deploy/app-subdomain/README.md` ×2, `deploy/app-subdomain/_headers` ×1) — the plan's "zero inbound links" claim for this file was wrong.
  - **Round 2:** folded the two facts that actually died with `docs/expo-web.md` into `docs/expo-web-deployment.md` — `BOARDSESH_EXPO_WEB_READY_TIMEOUT_MS` (the escape hatch when `vp run dev:mobile:web`'s cold bundle outruns the orchestrator's readiness window; it was documented nowhere else, `scripts/lib/expo-web-readiness.ts` being the only other mention) and the "locale-neutral `noindex` utility surface, keep it out of the sitemap and locale routing" scope rule. The PR's original "every fact is already stated elsewhere" claim was wrong for those two.
  - **Round 2:** corrected the new test's history claim — only three of the five entries were orphaned by W-16. `@boardsesh/ble-protocol` survived W-16 (3 hits in `app/development/…/payload-decoder.test.ts` at `d5a8336c8`) and died at W-19's `fac5032af`; `@boardsesh/crypto` had zero web consumers from `ed17174cc` (2026-06-13), two months *before* W-16.
  - **Round 2:** cited W-19 and W-24 by issue (`#4437`, `#4438`) rather than by implementing PR (`#4479`, `#4480`) in `docs/web-reposition.md`, matching how the rest of the repo refers to them (`next.config.mjs:211`, `deleted-private-surfaces.test.ts:2`, `robots.test.ts:52`), and normalised the two older occurrences in the table below in the same pass.
  - **Round 2:** removed the Phase A0 section's self-contradiction. It asserted as fact that the device pass "ran on real hardware on both origins" while stating four lines later that no record of one exists. It now says the gate *was to* run on real hardware, and says plainly that **no device pass is on record.**
  - Reconciled `docs/web-reposition.md`'s stale `A1`–`A6` cross-references (there is no `A1`–`A6` checklist in this file to mark done — the doc explicitly says that plan lives with the maintainer) to the W-items that actually shipped them, and added a "Couplings the A0 audit missed" subsection recording the four import-graph edges the audit didn't catch: three were cut by relocation (not deletion) rather than left dangling, one (`activity-feed/*` → `climb-card/climb-icons`) is still live as an intra-kept-surface edge.

**Production targets reached: none.** Every change is under `packages/web/`, `scripts/` (a CI guard, not shipped code), `docs/`, or `deploy/app-subdomain/` docs — nothing under `packages/mobile/` or a mobile-consumed shared package. No OTA, no native fingerprint move, `packages/web/package.json` and `bun.lock` are untouched (a follow-up issue covers dropping the five now-unused workspace deps — rewriting `bun.lock` here would trigger all seven mobile fingerprint workflows and bust the ~30-minute native screenshot cache for a web-only cleanup).

## Epic #4358 definition-of-done — the machine-checkable half only

**This PR does not close #4358 and must not be merged with a closing keyword.**
Three of the epic's eight DoD boxes are unmet, one of them by an open issue. The
section below covers only the three machine-checkable boxes; the end-to-end
boxes are listed under it with their real state.

Machine-checkable (all three green on this branch):

- **W-05's import-graph allowlist is empty** — `app/__tests__/import-graph-allowlist.json` is `{"entries": []}`; `vp test run --reporter=agent --project web app/__tests__/import-graph-invariant.test.ts app/__tests__/next-config-redirects.test.ts` → 27 tests passed.
- `vp run check:i18n` → OK, 323 `.tsx` files scanned, no hardcoded strings. `vp run check:i18n:orphans` → OK, 1633 source files scanned, all 3820 catalog keys have a live reference.
- `vp run build`/`vp run typecheck` + `vp check` + `vp run test:e2e`:
  - `vp run typecheck` → green (`generate:openapi && next build` 55.96s, then `tsc --noEmit` 1.59s).
  - `vp check` → 0 errors, 323 warnings (pre-existing) — clean once `packages/mobile/web-runtime`'s isolated install is present (CI installs it; absent by default on a fresh box).
  - `vp run test:e2e` — `.github/workflows/e2e-tests.yml` (`workflow_dispatch`-only) dispatched on the exact shipping sha `5d4fa0193`, run [32267236107](https://github.com/boardsesh/boardsesh/actions/runs/32267236107) — **all 8 shards green**, plus `Expo-web smoke` green on the same dispatch. Earlier dispatches on `082fb8fe7`, `cda009e02`, `99e03c6ed` and `c5bcb3940` are superseded and are **not** the evidence for this box — the point of naming the sha is that the box is evidenced on the commit that would merge, not on an ancestor. Scope per the plan's ruling: the **8 www Playwright shards**. `Expo-web smoke` (`vp run test:e2e:expo-web`) is a separate task CLAUDE.md already documents as out of per-PR CI and is not part of this box.

The end-to-end half — **not verified by this PR, and two of them cannot be ticked today**:

| DoD box | State |
|---|---|
| `curl` a climb front door in every locale: one visible `<h1>`, board ``, ≥3 internal links, a CTA at `APP_URL` + the same pathname, identical canonical from both trees | **Box wording is wrong, behaviour is right.** On `/es`, `/fr` and `/de` the CTA href drops the locale segment: `buildAppHandoffUrl` strips every non-default prefix on purpose, because the Expo app has no localised route tree and a `/es` app URL would land on the SPA's not-found. The epic's own Locked Decisions table already accepts this ("the CTA strips the locale prefix, so a Spanish reader lands on an English app"), so the box contradicts the decision above it. This PR records the carve-out in `docs/web-reposition.md` where `buildAppHandoffUrl`'s docblock claims it lives; **the box itself needs amending to "the same pathname, locale prefix stripped" before it can be ticked.** No code change — prefixing the href would break the hand-off. |
| Anonymous private window: kiosk / embed / gym / setter / playlists / profile all render, no WebSocket on a marketing page | Not re-run here. `/setter/{u}` is separately known broken (an indexable empty shell that 200s for any username) — **#4473**, open. |
| A signed-out deep link from a www climb page into `app.boardsesh.com` resolves to that climb (W-06) | **Blocked by #4478, open** — the signed-out hand-off lands on a login form, not the climb. The work is scheduled for wave C and is not written yet. |
| The Pi controller's `/…/list` target returns 200 with crawlable climb links | Not re-run here. |
| PostHog shows `Climb Handoff Clicked` / `Board Route Handoff` with `environment=production-web`, Sentry receiving from the app host | Not re-run here. |

Posted as a comment on the epic too, so the reconciliation lives where the boxes do: https://github.com/boardsesh/boardsesh/issues/4358#issuecomment-5342876288

Nine further issues still carry `epic:web-reposition` and are open: #4493, #4487, #4478, #4473, #4466, #4465, #4463, **#4443 (W-25b — a W-item in the epic's own reconciliation table, unshipped)**, #4408, #1889.

## Rebased onto main, 2026-09-06

Two conflicting hunks in `docs/web-reposition.md`, both trivial — main's newer
text about `/api/internal/board-render` wins, keeping this branch's correction
of the W-19 issue number (#4479 → #4437).

**The WASM guard was testing a key that no longer exists.** W-26's issue text
says "keep `outputFileTracingIncludes` — the WASM include is *more* load-bearing
after the teardown". That stopped being true: **#4715 deleted the
`/api/internal/board-render` route** and replaced it with an unconditional
external rewrite to the Railway renderer, so the include went with the route.
The assertion was reading `undefined`.

The invariant moved rather than disappeared, and it is sharper where it landed.
`rewrites()` has four `return` paths, two of them early, and the ESP32 rewrite
has to be on **every** one — the firmware pins `https://www.boardsesh.com` as
its render host (`board_config.h:41`, same define in the MoonBoard dev server)
and cannot be updated. Miss it on one branch and boards on the wall 404 from a
build that passed everything else. The test now drives `rewrites()` through all
four branches and asserts both the board-render and board-geometry rewrites
survive each, pointed at `/render/board` and `/render/geometry`.

- **Mutation:** drop the board-render rewrite from the production early return
  only → **red on that branch**, nine other cases green.
- Worth naming because it cost a run: `process.env` coerces, so assigning
  `undefined` to a key sets the *string* `"undefined"`, which is truthy and
  steers the config down the wrong branch. The helper deletes instead.

Everything else in the diff survives the rebase unchanged: the five dead
`transpilePackages` entries are still dead on `main` (only hits are
`next.config.mjs` and `packages/web/package.json`), the `docs/expo-web.md` fold
still has three inbound links to repoint, and the env-parity widening is real —
`main` covers three `EXPO_PUBLIC_*` keys at workflow level, this adds three more
and pins the values of four.

Two parts of W-26's issue text were **already done on main** and are correctly
absent from this diff: `docs/view-page-removal-pivot.md` was deleted in
`c5b087b77` (its `ClimbViewSeoFragment` rule already lives at
`docs/web-reposition.md:166,197`, enforced by `view-seo-fragment.test.tsx`), and
`scripts/mobile-ci-env-parity.test.ts:111-136` already covered
`production-deploy.yml`.

## Test plan

1. CI green.

## Risk

Risk: 1/5 — config, docs and tests. The one runtime line is the removal of five
`transpilePackages` entries, each verified to have no importer under
`packages/web`; the new guard reds if any of them comes back, or if a kept entry
loses its last consumer. Nothing user-facing changes.

## Author verification

- `vp check --fix` → 0 errors, 323 warnings (pre-commit hook ran this and passed).
- `vp run typecheck:web` → green (`next build` 59.76s + `tsc --noEmit` 6.86s), dev server stopped first.
- `vp test run --reporter=agent --project web` → 358 files, 3958 tests passed.
- `vp test run --reporter=agent scripts/mobile-ci-env-parity.test.ts` → 38 tests passed (34 pre-existing + 4 new).
- `vp run check:i18n` / `vp run check:i18n:orphans` → both OK.
- Mutation-tested every new assertion by hand (apply mutation → red → revert → green), `git status --porcelain` clean afterward:
  - Re-adding a pruned `transpilePackages` entry → red on the "does not list any pruned package" test.
  - ~~Deleting `outputFileTracingIncludes` → red on the WASM-globs test.~~ Superseded by the rebase; see below.
  - Un-excluding `package.json`/`next.config.mjs` from the transpile scanner (with a pruned entry reinstated) → confirmed the check would otherwise be vacuous, i.e. the exclusion is load-bearing.
  - Re-indenting `EXPO_PUBLIC_BACKEND_URL` from workflow-level (2-space) to job-level (4-space) → red, confirming the regex only reads workflow-level `env:`.
  - Repointing `EXPO_PUBLIC_WEB_URL` at `app.boardsesh.com` → red on the origin pin.
  - `EXPO_PUBLIC_BACKEND_URL` → `https://boardsesh-backend.up.railway.app` → red on the new value pin (it was only presence-checked before this round, so a wrong-but-plausible host passed).
  - `EXPO_PUBLIC_WS_URL` → `ws://` instead of `wss://` → red on the new value pin.
- Dispatched `.github/workflows/e2e-tests.yml` five times across this branch's history; the one that counts is the last, on the shipping sha `5d4fa0193` (run [32267236107](https://github.com/boardsesh/boardsesh/actions/runs/32267236107)) — 8/8 shards green, `Expo-web smoke` green. Every push re-dispatched it rather than leaving the box evidenced on an ancestor.
- **Round 5 (`5d4fa0193`), claude-review nits:** dropped the dead `'package.json'` entry from the scanner's skip list (`.json` is not in `SOURCE_EXTENSIONS`, so the walk never opens it) and said what the one remaining entry is for; and restored the `_headers` breadcrumb to the CSP rule's README section, which the `docs/expo-web.md` delete had removed with no replacement pointer. **Rebutted** the third nit — top-level `await import('../../next.config.mjs')` is the established pattern in five sibling tests (`next-config-redirects`, `expo-web-header-parity`, `settings-scope-down`, `deleted-private-surfaces`, `next-config-expo-web`); diverging here would be the surprise.
- **Round 4 (`c5bcb3940`), docs only:** the guard's residual blind spot is now stated in the file instead of implied. It is a regex, not a parser — a comment that quotes a whole import statement (`// re-exported from '@boardsesh/foo'`) still counts as evidence — which is exactly why the hardcoded `prunedInW26` list stays as a backstop rather than being deleted as redundant.
- **Round 3 (`99e03c6ed`), caught by self-review of round 2:** the anchored scan collected `@boardsesh/*` specifiers only, so the "every entry is imported" assertion would have gone **red on a correct config** the day someone legitimately added a third-party ESM-only dep to `transpilePackages`. A guard that false-fails is the other way a guard stops being trusted. It now collects every bare module specifier and reduces it to a package name (`@boardsesh/board-config/holds` → `@boardsesh/board-config`), so it covers any entry. Two extra mutations on top of the four below: a dead **non**-`@boardsesh` entry (`react-native-web`) → **red**; the same entry with a real non-test import → **green**. Plus the backstop re-checked: re-adding `@boardsesh/crypto` *with* a real import still reds the pruned-package assertion.
- **Adversarial review round 2 fixed (`cda009e02`):** the "has a real import" guard was still a bare substring scan — a prose comment naming `@boardsesh/<pkg>` anywhere under `packages/web` kept a dead entry green, which is the same self-supplied-evidence class as the two fixes before it, just sourced from any file instead of from the test itself. The scan now anchors on a module-specifier position (`from '…'`, `import '…'`, `import('…')`, `require('…')`) and **excludes test files**: `transpilePackages` exists to make the Next *build* compile workspace source, and `@boardsesh/ble-protocol`'s last web reference before this prune was a test file — exactly the shape the old scanner counted as alive. Verified all 21 kept entries still have a non-test import (counts 1–130) before tightening. Four mutations run against the new version:
  - dead entry (`@boardsesh/watch-pairing`) added → **red**: `is imported by no non-test source file under packages/web`.
  - dead entry + the prose comment that defeated the old version (`// TODO: this used to come from @boardsesh/watch-pairing` in `app/lib/types.ts`) → **still red**.
  - dead entry + a real `import` of it from a **test** file → **still red**.
  - dead entry + a real `import` of it from a non-test file → **green**, proving the anchor isn't just always-false.
- **Review feedback fixed (6c5ed26c7):** claude-review caught that `next-config-tracing-and-transpile.test.ts` scanned itself for `@boardsesh/*` specifiers, so the five pruned package names it lists in its own regression-guard array supplied false "consumer evidence" for themselves — weakening the "has a real import" assertion for exactly those five. Added the test file's own path to `SKIPPED_RELATIVE_FILES` and re-mutation-tested: re-adding a pruned entry now fails both assertions, not just the hardcoded-list backstop.

## Release Notes

none


marcodejongh:
Fixed in 6c5ed26c7 — added the test file's own relative path to `SKIPPED_RELATIVE_FILES`. Verified the fix by re-mutation-testing: re-adding `@boardsesh/crypto` to `transpilePackages` now fails both "has a real import" AND the hardcoded-list backstop, not just the latter.

marcodejongh:

Conflicting: docs/web-reposition.md — a real doc-content conflict, not the mechanical bun/pnpm surface I was authorized to auto-resolve.

Issue #3431, opened 2026-07-03
Harden gym domain-claim against editor-set-website ownership escalation


marcodejongh:
## Context

Follow-up from #3409 (gym claims + write-access). The domain-verified claim path auto-transfers ownership based on `gyms.website`, but that field is editable by gym editors / gym admins / covering community leaders (via `updateGym`).

## The gap

The current guard in `requestGymClaim` only blocks the **same account** that already has edit access from taking the domain path. It does **not** stop a **sockpuppet**:

1. An editor / community leader sets `website = https://attacker-owned.com` (they have edit access).
2. A second account they control (with no edit access on the gym, and an address `x@attacker-owned.com`) requests a domain claim.
3. The verification email lands in the attacker's own inbox; clicking it transfers ownership to the second account.

For a board-type community leader this scales to any gym with that board type. This was flagged by the PR's automated reviewers (Codex P1) and the manual security red-team.

## Decision

Shipped as-is with mitigations, by product decision (accept-the-risk + follow-up):
- Same-account edit-access holders are blocked from the domain path; `canClaim` hides the Claim button from anyone with edit access.
- A displaced real owner is **notified** and **kept as a gym admin** (not locked out).
- Every transfer is logged and reversible by an admin.
- Actors capable of the seed are admin-granted, vetted roles.

## Options to harden later (pick one)

1. **Lock the `website` field to gym owner / gym admin** (not editors / community leaders). Closes the seed vector; tradeoff is that a fresh system-owned catalog listing can't get a website from a community leader, so its first claim goes through admin review.
2. **Auto-grant domain claims only on placeholder (system-owned) gyms**; a gym with a real owner routes to admin review (protects real owners from silent takeover).
3. **Route every domain claim through admin review** — the verified email becomes strong evidence an admin one-clicks, instead of a fully automatic transfer.
4. **Trust-on-verify**: add a `website_verified` flag an admin sets once; only verified websites are usable for domain claims (keeps community-leader editing).

## Acceptance
- An editor / community leader cannot obtain ownership of a gym they don't own via a website they set + a second account.
- Legitimate self-service domain claims for real gym owners still work.

marcodejongh:
🤖 Claimed by Codex bug-batch loop root (batch 2, worktree wt-20260731-210637). Planning now; parallel loops, please pick a different issue.

PR #4559, opened 2026-08-19
feat(backend): expose Gym.canClaimByDomain so claim UIs can route (#4018)

### LINKED ISSUE #4018: Claim dialogs offer the email form on gyms whose website isn't owner-vouched
Follow-up to the #3431 security fix.

After #3431, a gym's email-domain claim only auto-verifies when the gym's **owner** put that website on the listing (`gyms.website_vouched_by_owner`). Both claim UIs still decide which form to show from `gym.website` alone, so on an un-vouched gym a climber types their work email, hits submit, and only then gets the backend's refusal:

> This gym's website hasn't been confirmed by the gym's owner, so we can't verify you by email. Request admin review instead.

They should land straight on the admin-review form instead — the same way a free-provider website (`gmail.com`) already behaves today.

### Where it decides
- `packages/web/app/components/gym-entity/claim-gym-dialog.tsx:49` — `const canUseDomain = isClaimableDomain(website)`
- `packages/mobile/src/components/gym-directory/ClaimGymSheet.tsx:45` — same

### What it takes
No new copy: both surfaces already render the admin branch from existing `claimGym.admin.*` keys whenever `canUseDomain` is false, so **zero i18n work in any locale**.

The actual cost is plumbing a `canClaimByDomain` field through:
- the `Gym` GraphQL type **and** `SimilarGym` (`packages/shared-schema/src/types/gyms.ts`, resolver `gym-matching.ts`) — `similar-gym-suggestions.tsx:193` opens the same dialog with a `SimilarGym`
- `vp run codegen`
- three web call sites + the mobile sheet
- six existing `Gym` test fixtures that build full literals: `gym-detail.test.tsx`, `embed-access.test.ts`, `home-gym-card.test.tsx`, `my-gyms-drawer.test.tsx`, `gym-role.test.ts`, `gym-list-panel.test.tsx`

Deliberately kept out of the security fix so it didn't triple that PR's blast radius. The backend refusal is correct today; this is purely about not making people fill in a form that can't succeed.

marcodejongh:
## Summary

**Read this first: #4018 is a UI-routing / dead-end-form defect, not a security hole.** A reviewer who
sees "claim" and "vouched" will reasonably brace for the worst, so: the backend gate shipped in #3431
is real, it is correctly ordered — the vouch check runs **before** the email-domain match, so a prober
cannot learn whether their address would have matched — and it is covered end-to-end by
`gym-write-access-and-claims.test.ts`. Nothing in this PR moves, duplicates, or relaxes that decision.
`requestGymClaim` stays the only enforcement point.

The bug is that neither claim UI can see half the rule. Both pick their form from
`isClaimableDomain(website)` alone and never read `gyms.website_vouched_by_owner`, because that column
isn't exposed on the GraphQL `Gym` / `SimilarGym` types at all. So on an un-vouched gym with a
corporate-looking website, the dialog opens on the email form, a climber types their work address,
and only the mutation tells them no.

This PR adds the missing capability field so a client can route to the form that can succeed. **It is
backend-only on purpose** — no client document selects it yet.

### What changed

- `Gym.canClaimByDomain: Boolean!` and `SimilarGym.canClaimByDomain: Boolean!` in the SDL + the
  hand-written TS types, plus the regenerated `packages/shared-schema/src/generated/*` and
  `packages/shared/graphql/src/generated/*` (`vp run codegen`, never hand-edited).
- Computed in `enrichGym` and in `findSimilarGyms` as
  `isClaimableDomain(gym.website) && gym.websiteVouchedByOwner` — imported from `@boardsesh/gym-claim`,
  the *same* helper `requestGymClaim` uses, so the advertised capability cannot drift from the gate.
  No extra query: `websiteVouchedByOwner` is already on the row, and `gym-matching`'s `gymColumns`
  gains one column.
- The `// TODO(#4018)` above the vouch refusal is resolved and the comment reworded.

### Two PRs, and why this one is first

**PR 2 (the client half) must not merge until this PR is deployed.** The production mobile OTA
auto-publishes on every push to `main` and can reach devices *before* the backend deploys, and an
unknown field fails validation for the **whole** GraphQL document — not just the field. Landing the
`GYM_FIELDS` selection first would break every mobile gym view for the length of the deploy window.
This is the same hazard already documented at `packages/shared/graphql/src/operations/gyms.ts:108-123`
for `myPendingClaim`.

- **PR 1 (this one)** — base `main`. Schema + resolvers + tests.
- **PR 2** — #4572, base `fix/4018-claim-vouched-domain` (this branch). Carries `Closes #4018`.

### One deliberate disclosure worth an explicit nod

`canClaimByDomain` is **viewer-independent** (like `isClaimed`, and unlike `canClaim`). It has to be:
the public gym page server-renders its claim call-out for signed-out visitors — the gym owner who just
googled their own gym — so a viewer-dependent value would route the anonymous arm at the wrong form.
The consequence is that any viewer can tell whether a gym's website is owner-confirmed. That reads as
a feature rather than a leak (it's the same trust signal we'd want to render as a badge), but it is a
deliberate choice, not an accident. Flagging it for a reviewer's yes/no.

## Release Notes

## Test plan

`vp test run --project backend gym-write-access-and-claims find-similar-gyms --reporter=agent` —
**2 files, 136 tests, all passing.**

New coverage in `gym-write-access-and-claims.test.ts`, a table over four listings — (vouched +
company domain), (vouched + `gmail.com`), (un-vouched + company domain), (no website) — asserting
`canClaimByDomain` is `true / false / false / false`. Each case **also** fires `requestGymClaim` with
a matching claim email in the same test and asserts it resolves iff the capability was true, with the
specific refusal otherwise. That pairing is the point: the field can't drift from the gate it
advertises without a test going red. Plus:

- the value is identical for an authenticated and an anonymous viewer (the anonymous-SSR property above);
- an editor gets `canClaim: false` with `canClaimByDomain: true` — the two flags answer different questions;
- it flips to `false` the instant an editor rewrites the website, matching the mutation;
- it survives `mapRawGymRow`, so the PostGIS proximity branch of `searchGyms` (a raw `SELECT *`) doesn't
  silently read `undefined` for `website_vouched_by_owner` and route every nearby gym to admin review.

In `find-similar-gyms-and-auto-gym.test.ts`: a candidate with an un-vouched company website comes back
`isClaimable: true, canClaimByDomain: false`, so a regression that conflates the two fails.

### DIFF
--- a/packages/backend/src/__tests__/find-similar-gyms-and-auto-gym.test.ts
+++ b/packages/backend/src/__tests__/find-similar-gyms-and-auto-gym.test.ts
@@ -62,12 +62,23 @@ const insertGym = async (opts: {
   longitude?: number | null;
   deleted?: boolean;
   isPublic?: boolean;
+  website?: string | null;
+  websiteVouchedByOwner?: boolean;
 }): Promise<{ id: number; uuid: string }> => {
-  const { ownerId, name, latitude = null, longitude = null, deleted = false, isPublic = true } = opts;
+  const {
+    ownerId,
+    name,
+    latitude = null,
+    longitude = null,
+    deleted = false,
+    isPublic = true,
+    website = null,
+    websiteVouchedByOwner = false,
+  } = opts;
   const uuid = uuidv4();
   const result = await db.execute(sql`
-    INSERT INTO gyms (uuid, name, slug, owner_id, is_public, latitude, longitude, deleted_at, created_at, updated_at)
-    VALUES (${uuid}, ${name}, ${uuid}, ${ownerId}, ${isPublic}, ${latitude}, ${longitude}, ${deleted ? sql`now()` : null}, now(), now())
+    INSERT INTO gyms (uuid, name, slug, owner_id, is_public, latitude, longitude, website, website_vouched_by_owner, deleted_at, created_at, updated_at)
+    VALUES (${uuid}, ${name}, ${uuid}, ${ownerId}, ${isPublic}, ${latitude}, ${longitude}, ${website}, ${websiteVouchedByOwner}, ${deleted ? sql`now()` : null}, now(), now())
     RETURNING id
   `);
   return { id: Number(Array.from(result as Iterable<{ id: number }>)[0].id), uuid };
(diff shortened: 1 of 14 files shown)


marcodejongh:
Rebased onto current main and addressed the remaining review notes in `95a822100abfe9c3e10a8b80d45bc30230c3dcdd`. Both query paths now call the same pure `computeCanClaimByDomain` helper, and the editor test title explicitly separates viewer claim permission from website capability. The mutation retains its existing refusal messages and owner-vouch-before-email-match order; the real-database tests still compare the advertised capability with actual claim acceptance/refusal.

Validation: 136 backend claim/matching tests and 45 shared-domain/web fixture tests passed; code generation, full `vp check`, all 53 full-typecheck tasks and the commit hook passed. Independent paired review approved the complete rebased change. Fresh CI is still running.

The dependent client PR #4572 is being repaired locally and its push is held: automatic previews use the production backend, so this schema must be deployed before those clients select the new field. No production deployment or settings changes were performed.

marcodejongh:
Confirmed both minor checks on `95a822100`: `packages/db/src/schema/app/gyms.ts:51` defines `websiteVouchedByOwner` with `.default(false).notNull()`, and `gym-write-access-and-claims.test.ts:299` calls `vi.clearAllMocks()` in `beforeEach` before each claim assertion. The 136-test real-database suite passed with these definitions; no further code change is needed for either note.

Issue #4558, opened 2026-08-19
Sort a playlist by grade


marcodejongh:
Split out of #4533, which bundled two unrelated asks. The first half — playlist tags on the selected climb — ships separately. This is the second: **let a climber sort a playlist by grade.**

> "Also sorted playlists please! By grade" — reported in-app from the play drawer (app_feedback #10205).

## Why this isn't a small change

Playlist order today is a persisted, user-curated column, resolved and paginated server-side:

- `packages/backend/src/graphql/resolvers/playlists/queries/playlist-climbs.ts:139` and `:204` — both the specific-board and all-boards branches do `.orderBy(asc(playlistClimbs.position), asc(playlistClimbs.addedAt)).limit(pageSize + 1).offset(page * pageSize)`.
- Grade isn't available to that query. `difficulty` / `quality` / `ascents` are joined **after** the page is selected, by `hydrateClimbsByRefs` (`packages/backend/src/graphql/resolvers/playlists/helpers/hydrate-climbs.ts:41-80`), with a per-row angle override and a most-ascents fallback.

So:

- **A client-side sort isn't honest.** The playlist detail screen is an infinite query at `pageSize` 20; sorting the pages already loaded gives a list whose order changes as you scroll.
- **A server-side sort means folding the stats join and its angle resolution into the ref query**, in both branches, plus a new `sortBy` on `GetPlaylistClimbsInput` across the schema, the validation schema, the shared operation, and `usePlaylistClimbs`.

## The interlock that's easiest to miss

Sorting fights the manual order. `handleReorderClimb` (`packages/mobile/app/(tabs)/discover/[playlist_uuid].tsx:238-251`) sends `newIndex` as the **displayed** list index, and the backend (`packages/backend/src/graphql/resolvers/playlists/mutations.ts:504-517`) reads the full `position, addedAt`-ordered list and densely renumbers absolute positions from that index.

If the view is grade-sorted and a climber drags a row, their hand-curated order is silently scrambled. Whatever this issue decides, it has to either force manual order in edit mode or disable drag while a sort is active.

## Two open questions to settle first

These are product decisions, not implementation details — they're what makes this a feature rather than a toggle.

**1. Is grade sort a view-only preference or a persisted per-playlist setting? And is it remembered per playlist or globally?**
[...]

Issue #4533, opened 2026-08-18
🐞 Bug report: Suggestion: playlist tags show up near name of climb when selected, currently only in list scroll vers...


marcodejongh:
Reported from the Boardsesh mobile app.

## Comment

Suggestion: playlist tags show up near name of climb when selected, currently only in list scroll version. Also sorted playlists please! By grade

## Metadata

| Field | Value |
| --- | --- |
| Platform | ios |
| App version | 2.3.1 (2) |
| Source | drawer-bug |
| Board | kilter / layout 1 / size 10 / sets [1,20] @ 40° |
| Climb | uncast shadow (7a/V6) |
| Session | 0423903e-c40e-4983-a1c0-3be51bd67cbd |
| URL | /home |
| User agent | unknown |
| Feedback row | app_feedback #10205 |

## Contact

✅ Reporter agreed to follow-up — resolve **app_feedback #10205** in the DB to reach them.

marcodejongh:
🤖 Claimed by bug-batch loop `moarbugs10000` (batch 2, session 82e34924).

Investigation + fix in progress. If another automation loop is already on this, say so here and I will stand down.

marcodejongh:
Two asks in here, splitting them so neither waits on the other.

**Playlist tags on the selected climb** — building it, PR #4560. The tags now sit under the name in the play drawer, not just on the list. Worth flagging one thing found on the way: the chips read a shared membership store that only the Climbs tab ever fills, so a drawer opened from the queue or a deep link had no membership data to show at all. The drawer gets its own per-climb fetch, sharing a cache key with the add-to-playlist sheet, so toggling a playlist there updates the header straight away.

**Sorted playlists, by grade** — tracked as #4558, not dropped. It's a bigger change than it looks: playlist order is a real user-curated column that the backend sorts and paginates on, and grade isn't even available to that query (it's joined per row, at a per-row angle, after the page is picked). It also collides with drag-to-reorder — dragging a row while a sort is applied would silently scramble the hand-curated order. That issue carries the mechanism and the two decisions it needs first: whether grade sort is a view preference or a persisted per-playlist setting, and what the queue plays when a sorted playlist is activated.

PR #4560, opened 2026-08-19
feat(mobile): show playlist tags on the selected climb

marcodejongh:
## Summary

Playlist tags only ever appeared on Climbs-tab rows. Open a climb and the membership vanished — exactly where a climber wants to know whether this one is already in a project list. This puts the tag on the header's stats line, beside the sends/quality/setter.

Refs #4533. That report bundled two unrelated asks; this ships the first. The second — "sorted playlists please! By grade" — is now #4558.

### QA decline: the tags were eating the board

The first cut gave the tags a 30pt line of their own under the stats. QA declined it, correctly: the drawer's first screen is a **fixed height** and the board art inside it is `flex: 1`, so those 30pt came straight off the board renderer. Every climber with a playlist on the board got a visibly smaller board.

The tags now ride **inside** the existing stats line as caption text — not a line of their own, and not a capsule:

```
body line height + 2 + caption1 line height   <=  44
iOS  (HIG scale)   22 + 2 + 16 = 40           ✓
Android (M3 scale) 24 + 2 + 16 = 42           ✓
```

44 is `WALL_STATE_PILL_TOUCH_HEIGHT`, the `minRowHeight` floor `DrawerHeader` already sets for the wall-state pill — so the centre column fits under a ceiling that was there before this PR and the board keeps every point it has today: signed out, no playlists, in one playlist, in five, mid-swipe.

Dropping the capsule is what makes it fit. A filled pill sitting in a grey caption line reads as a tappable control (it isn't — `pointerEvents: 'none'`), spends 16pt of horizontal padding the line can't spare, and puts a second grey next to the stats. Inline is a colour dot plus the name in the stats' own grey, capped at 88pt so a long playlist name can't crowd them out. The header names one playlist then `+N`; the list rows keep two, and keep the chip.

**Both type scales, not just the HIG one.** A first attempt pinned the row at 20pt: 22+2+20 = 44 cleared the floor on iOS and 24+2+20 = **46** missed it on Material — the resolved variant on every Android device. That would have shipped the same regression to Android, 2pt at a time, and a HIG-only assertion could not have seen it. `play-drawer-header-layout-budget.test.tsx` now reads the name and stats *type variants back off a real render* and checks the sum against **both** scales. Mutation-checked both ways: switch the stats label to `subheadline` and only the Material case goes red (`expected 46 to be less than or equal to 44`); move the tag back below the row and the nesting assertion goes red.

One thing this does **not** claim: the header is not always 68pt. The climb-**rules** line that landed on main since is a deliberate exception — a truncated rule is worse than a shorter board. Playlist membership doesn't clear that bar.

### The part that isn't obvious

Dropping the list's `<ClimbPlaylistChips>` into the header would have rendered an empty strip forever.

The list chips read the shared `playlistMembershipStore`, and that store has exactly one feeder: `useClimbListPlaylistMemberships`, called from the Climbs tab and nowhere else. A drawer opened from the queue, from a playlist, or from a deep link has no membership data at all. So the drawer needs its own per-climb fetch. That query already existed inside `InlinePlaylistPicker`; it's lifted into `useClimbPlaylistMembershipQuery` and shared key-for-key, which buys a second thing for free — toggle a playlist in the add-to-playlist sheet and the header updates with no refetch, because the picker's optimistic `setQueryData` lands on the entry the tag reads.

The store seed still paints first when the climber did arrive from the Climbs tab, so the tag is there on the first frame; the fetch confirms behind it. Results are deliberately not written back into the store — the Climbs tab resets it on every board/auth/preference change and would wipe them.

Only the climb on screen fetches. The swipe "peek" header renders from cache, so flinging through a queue doesn't spray a request per climb it passes.

### Two smaller calls

- **Shown unconditionally.** The "Show playlist tags" setting stays list-scoped — it exists because tags add a third line to every row, a density cost a header that adds no line doesn't carry. Its copy still describes the list, and is unchanged.
- **Announced to VoiceOver**, unlike the list strip (which stays hidden so a row remains one clean target). The label names every playlist, including the ones the `+N` token hides — one new key, `climbs:mobile.detail.inPlaylists`, in all four locales.

## Release Notes

- See which playlists a climb is in without leaving it — the tag now sits with the stats when you open a climb, not just on the list.

## Test plan

1. Open a queued climb. Its playlist tag sits beside the stats.
2. Swipe through the queue. The board art never changes size.
3. Add the climb to a playlist. The tag appears immediately.
4. Sign out, reopen a climb. Header and board look untouched.

## Risk

Risk: 3/5 — play-drawer header layout, with the board art `flex: 1` directly under it.

https://claude.ai/code/session_01FyzmXaX2P2BVnPXEAVua2K

### DIFF
--- a/packages/mobile/src/components/ClimbPlaylistChips.tsx
+++ b/packages/mobile/src/components/ClimbPlaylistChips.tsx
@@ -21,17 +21,15 @@ const MAX_VISIBLE_CHIPS = 2;
 // sizes rather than letting it push the row taller than the board thumbnail.
 const CHIP_MAX_FONT_SCALE = 1.3;
 
-export type ResolvedPlaylistChip = { key: string; name: string; dotColor: string; emoji: string | null };
+type ResolvedPlaylistChip = { key: string; name: string; dotColor: string; emoji: string | null };
 
 /**
  * Turn a climb's playlist UUIDs into renderable chips against the playlists
  * provider's `uuid → Playlist` index — O(membership), never a scan of the whole
  * playlist array. UUIDs with no matching playlist (a stale membership, another
- * board's list) are dropped. Exported so a caller that also needs the names in
- * text form (the play drawer builds a VoiceOver label from them) resolves them
- * the same way.
+ * board's list) are dropped.
  */
-export function resolvePlaylistChips(
+function resolvePlaylistChips(
   playlistUuids: Iterable<string>,
   playlistsById: Map<string, Playlist> | undefined,
 ): ResolvedPlaylistChip[] {
@@ -60,12 +58,17 @@ type PlaylistChipsRowProps = {
   /** `start` for a left-aligned list row, `center` for a centered detail header. */
   align?: 'start' | 'center';
   /**
-   * VoiceOver label for the whole strip. Omit (the list-row default) to hide the
-   * chips from the accessibility tree, keeping the row one clean target. Pass one
-   * on a detail surface, where membership is information the climber would
+   * Builds the VoiceOver label for the whole strip from every playlist name,
+   * including the ones the "+N" token hides. Omit (the list-row default) to hide
+   * the chips from the accessibility tree, keeping the row one clean target; pass
+   * one on a detail surface, where membership is information the climber would
    * otherwise never hear.
+   *
+   * A formatter rather than a finished string so the caller doesn't have to
+   * resolve the names a second time, and so the translation lookup it needs stays
+   * out of this component — which renders once per visible list row.
    */
-  accessibilityLabel?: string;
+  describeForAccessibility?: (playlistNames: string[]) => string;
 };
 
 /**
@@ -81,7 +84,7 @@ type PlaylistChipsRowProps = {
 export const PlaylistChipsRow = React.memo(function PlaylistChipsRow({
   playlistUuids,
   align = 'start',
-  accessibilityLabel,
+  describeForAccessibility,
 }: PlaylistChipsRowProps) {
   const playlistsContext = usePlaylistsContextOptional();
   const { variant, systemColors, m3 } = useTheme();
@@ -92,6 +95,7 @@ export const PlaylistChipsRow = React.memo(function PlaylistChipsRow({
 
   if (chips.length === 0) return null;
 
+  const accessibilityLabel = describeForAccessibility?.(chips.map((chip) => chip.name));
   const visibleChips = chips.slice(0, MAX_VISIBLE_CHIPS);
   const overflowCount = chips.length - visibleChips.length;
 
--- a/packages/mobile/src/components/ClimbPlaylistChips.tsx
+++ b/packages/mobile/src/components/ClimbPlaylistChips.tsx
@@ -51,12 +51,34 @@ function resolvePlaylistChips(
   return resolved;
 }
 
+/**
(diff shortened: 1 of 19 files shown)


marcodejongh:

Conflicting: packages/mobile/src/components/play-drawer/PlayDrawer.tsx.

marcodejongh:
image image

marcodejongh:
Rebased on `main` and reworked the placement after the QA decline.

**What the tester saw was two things.** The 30pt tag line was one; the bigger half was the collapsed **Logbook** header squeezing its own title to zero width and wrapping it one letter per line, which inflated the first-screen reserve by ~100pt. `5bf55fc17` fixed that on `main` after the tested bundle was published — the rebase picks it up.

**The tag line is now free.** Two design reviews on the screenshots landed on the same answer: put it on the stats line the header already draws, not on a line of its own.

```
66k sends · 3.2★ · jwebxl  [● flash circuit]
```

Chip is 20pt, stats are 16pt, so the shared row is 20. Name (22) + 2 + 20 = 44 — exactly the header's existing 44pt row floor, the one already reserved for the wall-state pill. **Board delta: 0pt.** The row is held at that height whether or not a tag renders, so the header still measures the same for a climb in a playlist and the next one that isn't.

That deletes the reserved-slot constant, its clipping backstop and the cold-start settle it traded against, rather than managing them. One name plus `+N` instead of two, since the line is shared; the hidden ones are still named in full to VoiceOver.

Ready for another look — the `pr-4560` OTA branch will carry it once the preview publishes.

marcodejongh:
Rebased on `main` and pushed a fix for the decline.

**You were right, and it was worse than the screenshot showed.** The tags took a 30pt line of their own under the stats. The drawer's first screen is a fixed height with the board art `flex: 1` inside it, so those 30pt came straight off the board renderer — for every climber with a playlist on that board.

They now ride *inside* the existing stats line as caption text, so the header is the same height it is on `main`:

```
body line height + 2 + caption1 line height   <=  44 (the minRowHeight floor already there)
iOS  (HIG scale)   22 + 2 + 16 = 40           ✓
Android (M3 scale) 24 + 2 + 16 = 42           ✓
```

**On the design review you asked for** — it found a second bug and a better answer than the one I'd written:

- **A first attempt still regressed Android.** Pinning the row at 20pt cleared 44 on the HIG scale (22+2+20) and *missed* it on Material at 46 — the resolved variant on every Android device. Same regression you declined, 2pt at a time, on half the fleet. The guard test now reads the type variants back off a real render and checks both scales; mutation-checked, it goes red on exactly that case.
- **The capsule was the wrong token.** A filled pill inside a grey caption line reads as a tappable control (it isn't), spends 16pt of horizontal padding the line can't spare, and puts a second grey beside the stats. Inline is now a colour dot plus the name in the stats' own grey. Dropping the capsule is what made the height work out with slack instead of exactly zero.
- **Width.** With a wall pill up in a long-worded locale the centre column falls to ~106pt, so the tag is capped at 88pt and the stats ellipsize from the tail — the setter goes first, the tag survives. The header names one playlist then `+N`; list rows keep two.

I could not capture pixels here — `mobile:android-shots` needs a from-scratch emulator boot and the iOS path is macOS-only — so this is still worth a look on device. It's live on `pr-4560`.

marcodejongh:
image

Issue #4565, opened 2026-08-19
searchBoards proximity: cover the never-executed hideLocation privacy predicate


marcodejongh:
Follow-up to #4507, which added PostGIS to the backend test DDL and covered `searchGyms`'s proximity branch. The same DDL creates `user_boards.location`, its GIST index and its derivation trigger, so `searchBoards` is now testable too — it just wasn't covered in that PR, to keep it reviewable.

## The gap

`searchBoards`'s proximity branch (`packages/backend/src/graphql/resolvers/social/boards.ts:1253-1300`) has never executed against a real database. Nothing in the suite covers the radius cut-off, the distance ordering, or the `location IS NOT NULL` filter.

## The part that matters most

Unlike `searchGyms`, this query carries a **privacy predicate**:

```ts
// Hide boards with hideLocation=true unless the board owner follows the searching user
sql`(${dbSchema.userBoards.hideLocation} = false${
  ctx.isAuthenticated
    ? sql` OR EXISTS (
        SELECT 1 FROM user_follows
        WHERE follower_id = ${dbSchema.userBoards.ownerId}
        AND following_id = ${ctx.userId}
      )`
    : sql``
})`
```

A home-wall owner who ticks "hide my location" is trusting that predicate to keep their address off a public search. It has never run against a real row. Worth covering:

- anonymous viewer never sees a `hideLocation` board
- an authenticated viewer with no follow edge never sees one
- the escape hatch does open for whoever it is supposed to open for
- a follow edge in the *wrong* direction does not open it
- and `totalCount` respects the predicate too — the count and the rows share `whereClause` here, so this should hold, but it is the assertion that would catch a future refactor splitting them

## A question to settle first

The direction of the escape hatch reads oddly. `follower_id = user_boards.owner_id AND following_id = ctx.userId` means **the board owner follows the searching user** — not the other way round, and not mutual. The comment agrees with the code, so this is not a typo, but it is not the rule you would guess either ("people I follow can see where my board is" would be the reverse). Worth confirming the intent before pinning it in a test, since a test would otherwise cement whichever way it currently falls.

## Also missing a tiebreaker

`.orderBy(sql\`distance_meters ASC\`)` with `LIMIT`/`OFFSET` and no unique second key — two boards at one gym's coordinates tie exactly, so offset paging can duplicate or skip. Same flaw as #4505 fixed in `searchGyms`. Covered by the sibling issue on `myGyms`.

## Getting started

[...]

Issue #4564, opened 2026-08-19
myGyms pages over a non-unique ORDER BY, so offset paging can duplicate or skip gyms


marcodejongh:
Split out of #4507 (noted there under "Related, deliberately left alone"). Same latent flaw #4505 fixed in `searchGyms`'s two orderings; `myGyms` still has it.

## The bug

`myGyms` (`packages/backend/src/graphql/resolvers/social/gyms.ts:806`) pages with `LIMIT`/`OFFSET` over a sort key that is not unique:

```ts
const gymRows = await db
  .select()
  .from(dbSchema.gyms)
  .where(whereClause)
  .orderBy(desc(dbSchema.gyms.createdAt))   // no tiebreaker
  .limit(limit)
  .offset(offset);
```

Rows sharing a `created_at` have no defined order relative to each other, and Postgres is free to return them in a different order per statement. Each page is a separate query, so across an `OFFSET` boundary one gym can come back twice while another never appears at all.

`created_at` ties are not hypothetical here: the location sync writes gyms in bulk, so a whole import batch shares a timestamp down to the microsecond. A climber who is a member of several gyms from the same import is exactly the case that breaks.

## The fix

`gyms.id` as a second sort key, the same shape `searchGyms` already uses on both of its branches:

```ts
.orderBy(desc(dbSchema.gyms.createdAt), dbSchema.gyms.id)
```

## Worth checking at the same time

Anything else that pairs `OFFSET` with a non-unique `ORDER BY`. `searchBoards`'s proximity branch (`boards.ts:1325`) orders by `distance_meters ASC` alone — two boards at one gym's coordinates tie exactly.

## Test

`gym-directory-fields.test.ts` already has the pattern to copy — "pages cleanly over gyms sharing one created_at — no duplicates, no skips": seed several rows with an identical `created_at`, walk every page, then assert the collected uuids are complete, unique, and identical on a second walk. Uniqueness alone is not enough; a shuffled-but-complete result still breaks infinite scroll.

Issue #4562, opened 2026-08-19
searchGyms hard-fails without PostGIS, contradicting the portability doctrine two files over


marcodejongh:
Found while giving the `searchGyms` proximity path real coverage (#4507).

## The contradiction

`gym-matching.ts` states an explicit portability doctrine, right at the top of the file:

> Both the createBoard auto-gym guard and the findSimilarGyms read resolver need to answer "is there already a live gym at these coordinates with (roughly) this name?". The production `gyms` table carries a PostGIS `location` geography, but nothing here depends on the extension being installed — a self-hosted deployment can be missing it (…) so everything stays portable: a cheap lat/lng bounding-box prefilter in SQL, then an exact Haversine refine in JS.

`location-geography.ts` says the same thing more bluntly — the extension "can be missing on any self-hosted deployment".

`searchGyms` does not follow it. Its proximity branch (`gyms.ts:897-910`) goes straight at `ST_DWithin` / `ST_MakePoint` / `ST_Distance` with no fallback, so on a Postgres without PostGIS the query does not degrade — it throws. Same for `searchBoards` (`boards.ts:1253-1300`).

We also carry the doctrine into our own infrastructure: `CLAUDE.md` says we treat Railway as portable and "anything we write should run on a `docker run postgres:17`". `searchGyms` does not.

## The second half: the `location IS NOT NULL` guard

Both proximity queries filter `location IS NOT NULL`. That is not a no-op — it silently hides any gym whose geography was never synced. #4310 (migration 0127) exists precisely because ~2,800 gyms had lat/lng but a NULL geography and were invisible to proximity search, with no error anywhere. The trigger closed that hole going forward, but the query still treats a derived column as the source of truth while `latitude`/`longitude` sit right there on the same row.

## Options

1. **Adopt the `gym-matching.ts` pattern.** Bounding box on `latitude`/`longitude` in SQL, exact Haversine refine in JS. Portable, no extension, and no `location IS NOT NULL` blind spot. Costs the `gyms_location_idx` GIST index, and swaps spheroid `ST_Distance` for a sphere haversine — a small but real answer change at the 50 km default radius.
2. **Keep PostGIS, degrade explicitly.** Probe for the extension at startup and fall back to the bounding-box path when it is absent, so a self-hosted deployment gets slightly different distances instead of a 500.
[...]

PR #4566, opened 2026-08-19
test(backend): cover the searchGyms proximity path against PostGIS

### LINKED ISSUE #4507: searchGyms proximity path has no integration coverage (test DB lacks PostGIS)
Found while reviewing #4505 (epic #4372, Tier 1).

## The gap

`searchGyms` has two code paths. The text path is covered by real-database integration tests. **The proximity path has never been covered by one, and cannot be**, because:

- the backend test database is plain `postgres:15` (`packages/backend/docker-compose.test.yml:5`) — `pg_available_extensions` returns no PostGIS rows
- `gyms.location` is not in the hand-maintained test DDL (`packages/backend/src/__tests__/schema-sql.ts`)

So `ST_DWithin` / `ST_MakePoint` can't execute in tests at all. Any test that passes `latitude`/`longitude` to `searchGyms` either fails to reach the proximity branch or errors on the missing extension.

That path is not hypothetical: mobile's `useNearbyGyms` (the gym picker sheet, limit 50) runs it in production on every nearby-gym search, and #4380's "near me" directory view will run it on the web.

## What it currently costs

#4505 wanted to prove that a new `requireSlug` predicate lands in both the count and the rows statements on the proximity path — those are two separate `db.execute` calls that agree only because they interpolate the same filter clause, so a predicate on one and not the other silently desynchronises `totalCount` from the returned rows.

It could only prove that at the **rendered-SQL** level (assert the two statements share a byte-identical `WHERE`). That's a decent guard and it caught what it needed to, but it asserts the SQL we generate, not the answer Postgres gives. Nothing in the suite would catch a proximity query that renders correctly and returns wrong rows.

## What fixing it involves

Swapping the test image to `postgis/postgis`, adding the `location` column plus whatever trigger keeps it in sync to `schema-sql.ts`, and confirming the per-worker database bootstrap still works. That touches the setup shared by all 259 backend test files and the CI service definition, which is why #4505 didn't take it on.

Worth sizing before committing to it — the change is small in lines and broad in blast radius, and the per-worker databases persist between runs, so a partial rollout would leave some workers with the extension and some without.

## Related, deliberately left alone

`myGyms` orders without a unique tiebreaker, the same latent flaw #4505 fixed in `searchGyms`'s two orderings. Offset paging over a non-unique sort key can duplicate or skip rows between pages. It's a different query with existing callers, so it wasn't in scope there.

marcodejongh:
## Summary

Closes #4507.

`searchGyms` has two branches. The text branch has real-database coverage; the proximity branch — the one mobile's `useNearbyGyms` picker and the web directory's "near me" view actually run — had never touched a database. It could only be asserted at the rendered-SQL level, so nothing in the suite would catch a query that renders correctly and returns the wrong rows.

**The issue's premise is half wrong, and it's the expensive half.** The issue reads it as "swap the CI service image under all 259 backend test files". CI's `test-backend` Postgres has been `ghcr.io/boardsesh/boardsesh-postgres-postgis:latest` since 2026-02-14 (`236d8d4d2`, ci.yml:652) — Postgres 18.4 + PostGIS 3.6.4. **No CI change is needed.** Only the *local* `packages/backend/docker-compose.test.yml` was still `postgres:15`.

The real gap was inside the repo: `gyms.location` is a `geography(Point, 4326)` added by raw migration (`drizzle/0052`, `0054`, `0127`), so it lives outside the Drizzle schema and nothing generates it. It could only reach the test database by being written into `schema-sql.ts` by hand, and it never was — no `CREATE EXTENSION`, no `location` column, no derivation trigger.

### What changed

- **`schema-sql.ts`** now creates the extension, both `location` columns, the two GIST indexes (predicates copied from `0054`/`0052`) and the `0127` derivation trigger, at the very end of the string — `user_boards` is dropped and recreated above, which would otherwise take the column and its trigger with it.

  It is wrapped in a `DO … EXCEPTION` block, and that guard is load-bearing rather than defensive padding. `worker-db.ts` applies the whole 60KB string through postgres.js `unsafe()` — one simple query, so one implicit transaction. A bare `CREATE EXTENSION postgis` against a server without it aborts that transaction and rolls back **all 72 tables**, taking every backend test with it. I reproduced both sides: unguarded → `ERROR: extension "postgis" is not available` and 0 tables; guarded → 72 tables, a `NOTICE`, and `gyms.location` simply absent. That matters for anyone whose local container predates this PR.

- **Local compose image** → `postgis/postgis:18-3.6`. Also brings local Postgres in line with CI's 18.x (it was 15).

- **`BOARDSESH_TEST_DATABASE_URL`** points the backend suite at a Postgres of your own. It needs its own variable: `test.env.DATABASE_URL` in `vite.config.ts` is written into `process.env` during worker init, before `setupFiles` run, so an exported `DATABASE_URL` is clobbered back to `:5433`. Without this there was no way to verify a change like this except by restarting the shared `:5433` container under every other checkout on the machine.

- **16 new tests** in `search-gyms-proximity.test.ts`. Every fixture seeds plain `latitude`/`longitude` and lets the trigger derive the geography — the same derivation production depends on, so a test that set `location` by hand would pass while the trigger was broken. Coverage: the radius cut-off proved from **both** sides with a pair ~100 m either side of a 10 km boundary (a 5 km-vs-60 km pair passes even if the radius is out by 2×), nearest-first ordering seeded furthest-first, `totalCount`/rows agreement under paging, stable paging over gyms at an identical distance, NULL coordinates (insert **and** the UPDATE arm that clears them), private and soft-deleted gyms, `requireSlug` dropping rows from the count too, text-query-ANDed-with-radius, address matching, `boardTypes`, and cross-checking the proximity branch against the text branch. Plus the snake_case row mapping (`mapRawGymRow` off a `SELECT *`), which has already silently lost two fields (#3431 `website_vouched_by_owner`, then `hours_updated_at`) and had never executed against a real row.

  The tests self-skip on a server without PostGIS, and one **unconditional** guard test asserts `hasPostGis` is true when `CI` is set — so the file can't quietly go dark into a green build.

- **`location-geography-degradation.test.ts` now builds its own no-PostGIS world.** That file only tested anything because the test DB happened to lack the extension, and its premise probe said so in as many words — it fired the moment PostGIS arrived, which is exactly what a good premise probe is for. Rather than weaken it, `beforeAll` drops the `location` columns and their triggers and `afterAll` restores them from the same exported DDL string, so there is no second copy to drift. The SQLSTATE it expects now follows the server: with PostGIS installed the `::geography` cast resolves and the undefined column is what fails (42703); with no PostGIS the cast fails first (42704). A new case asserts the trigger teardown too — a surviving trigger would write the geography behind the resolver's back and hide the very failure the file exists to prove.

### The coverage found a real bug on its way in

Pinning the proximity row against the Drizzle row field-by-field (rather than against literals) surfaced **#4588**: `hours_updated_at` is `timestamp` **without** time zone, and `mapRawGymRow` coerces it with `new Date(string)`, which resolves the offset-less reading in the Node process's zone — while the Drizzle branch reads the same column as UTC. Same gym, same column, two different instants:

```
TZ=Asia/Tokyo
  proximity branch → 2026-02-03T01:30:00.000Z
  text branch      → 2026-02-03T10:30:00.000Z
```

Nine hours apart, exactly the host offset. Identical on a UTC host, which is why it has never surfaced — Railway and GitHub runners are both UTC. That is the third defect in this same hand-rolled mapping (after #3431 and `hours_updated_at` itself), and the first one a test has ever been able to see.

Not fixed here, for the same reason as the other behaviour questions below: it changes what a live resolver returns. Both branches are pinned exactly instead, so the fix collapses two assertions into one equality rather than making someone re-derive what changed.

### For you to decide: local vs CI image

CI uses `ghcr.io/boardsesh/boardsesh-postgres-postgis` (private — an anonymous manifest fetch 403s; the job passes credentials). Pointing the compose file at it would make every contributor `docker login ghcr.io` before running the backend suite, so this uses the public Docker Hub `postgis/postgis:18-3.6` instead. **Local and CI are therefore close but not identical: 18.6 + PostGIS 3.6.4 locally, 18.4 + 3.6.4 in CI.** If you'd rather have exact parity, making the ghcr package public is a one-click change and the compose file can point at it — your call, not a blocker for this PR.

### Deliberately not in scope

- **`searchGyms` hard-fails without PostGIS**, which contradicts the portability doctrine stated a few files over in `gym-matching.ts` ("everything here stays portable… a cheap lat/lng bounding-box prefilter in SQL, then an exact Haversine refine in JS"). That's a behaviour change to a live query, so it's #4562 rather than something smuggled into a test PR.
- **`myGyms` orders without a unique tiebreaker** — the latent flaw #4505 fixed in `searchGyms`'s two orderings — filed as #4564.
- **`searchBoards`'s proximity path** (`boards.ts:1253-1300`) has the same gap plus a `hideLocation` privacy predicate with a follower escape hatch that has never run against a real database. The DDL here unblocks it (`user_boards.location` is created too); filed as #4565 to keep this PR reviewable. I nearly folded it in — the fixture work is cheap now — but the escape hatch reads `follower_id = owner_id AND following_id = viewer`, i.e. *the board owner follows the searching user*, which is not the rule you would guess and not obviously the intended one. A test would cement whichever way it currently falls, so that question wants answering before it gets pinned, not after.
- **The timestamp divergence above** (#4588) — a live resolver's output.

## Release Notes

## Test plan

- Negative control, shared `postgres:15`: sent `schemaSQL` as **one** simple query (same semantics as `unsafe()`) to a scratch DB. Guarded version → 72 tables, `NOTICE: postgis unavailable …`, no `gyms.location`. Unguarded version → whole apply rolled back, 0 tables.
- Positive, PostGIS 3.6.3: same string → 75 relations, both `location` columns, both triggers, both GIST indexes. Re-applied to confirm idempotency (worker DBs outlive a run).
- Verified the trigger derives `POINT(4.9003 52.3791)` from plain lat/lng, clears the geography when lat/lng go NULL, and that the verbatim resolver SQL returns count=2 / rows nearest-first with the 60 km gym cut from a 10 km radius.
- `vp test run --project backend search-gyms-proximity` against a throwaway `postgis/postgis:18-3.6` container (18.6 + PostGIS 3.6.4) via `BOARDSESH_TEST_DATABASE_URL` — **16 passed**, none skipped.
- Full `vp test run --project backend` against a throwaway plain `postgres:15` — the degradation path: the bootstrap still applies cleanly for all 259 files and the proximity file reports skipped rather than red.
- Both throwaway containers ran on their own ports and were removed; the shared `:5433` container was never restarted.
- `TZ=Asia/Tokyo` against the same container — the run that found #4588, and green now that both branches are pinned.
- On a throwaway plain `postgres:15` via the override: **12 passed, 15 skipped** — the degradation suite runs and the proximity cases self-skip, which is the designed behaviour on a server without the extension.
- `vp run typecheck:backend`, and `vp check` via the pre-commit hook.
- CI is the authoritative signal here, since its Postgres already carries PostGIS. Confirmed in the job log that the proximity tests **executed** rather than skipped (`✓ … asserts PostGIS is present when running in CI`, then all 15 cases).

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -217,7 +217,7 @@ Web types in `packages/web/app/lib/types.ts`; shared types in `packages/shared-s
 
 ### Testing
 
-Vitest via `vp test`. Backend tests auto-start postgres+redis via `packages/backend/docker-compose.test.yml`; set `SKIP_TEST_INFRA=1` to skip, `CI=1` for caller-provided services. `packages/db` uses Node's native test runner (`tsx --test`).
+Vitest via `vp test`. Backend tests auto-start postgres+redis via `packages/backend/docker-compose.test.yml` (a PostGIS image — `gyms`/`user_boards` carry a `location` geography and the proximity queries are tested against it); set `SKIP_TEST_INFRA=1` to skip, `CI=1` for caller-provided services, `BOARDSESH_TEST_DATABASE_URL` to point the suite at your own Postgres. `packages/db` uses Node's native test runner (`tsx --test`).
 
 ## Development Guidelines
 
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -160,7 +160,7 @@ Web types in `packages/web/app/lib/types.ts`; shared types in `packages/shared-s
 
 ### Testing
 
-Vitest via `vp test`. Backend tests auto-start postgres+redis via `packages/backend/docker-compose.test.yml`; set `SKIP_TEST_INFRA=1` to skip, `CI=1` for caller-provided services. `packages/db` uses Node's native test runner (`tsx --test`).
+Vitest via `vp test`. Backend tests auto-start postgres+redis via `packages/backend/docker-compose.test.yml` (a PostGIS image — `gyms`/`user_boards` carry a `location` geography and the proximity queries are tested against it); set `SKIP_TEST_INFRA=1` to skip, `CI=1` for caller-provided services, `BOARDSESH_TEST_DATABASE_URL` to point the suite at your own Postgres. `packages/db` uses Node's native test runner (`tsx --test`).
 
 ## Development Guidelines
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -187,7 +187,7 @@ The backend is `packages/backend/`: a GraphQL-WS server on port 8080, backed by 
 
 Run it with `vp run dev:backend` (it starts the database first). Use the structured Winston logger, not `console.*` (a lint rule blocks `console` in backend source). See [docs/logging.md](./docs/logging.md).
 
-Tests run with `vp test run --project backend`. Vitest auto-starts Postgres and Redis from `packages/backend/docker-compose.test.yml`. The Postgres image carries PostGIS, because `gyms` and `user_boards` have a `location` geography that the proximity search queries. Set `SKIP_TEST_INFRA=1` to skip the Docker spin-up, `CI=1` when the caller already provides those services, or `BOARDSESH_TEST_DATABASE_URL` to run the suite against a Postgres of your own.
+Tests run with `vp test run --project backend`. Vitest auto-starts Postgres and Redis from `packages/backend/docker-compose.test.yml`. The Postgres image carries PostGIS, because `gyms` and `user_boards` have a `location` geography that the proximity search queries. Set `SKIP_TEST_INFRA=1` to skip the Docker spin-up, `CI=1` when the caller already provides those services, or `BOARDSESH_TEST_DATABASE_URL` to run the suite against a Postgres of your own — point that one at a throwaway server, because whatever it names inherits the whole harness, including the sweep that drops every `boardsesh_backend_test_w*` database it finds.
 
 ## Environment variables
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -187,7 +187,7 @@ The backend is `packages/backend/`: a GraphQL-WS server on port 8080, backed by 
 
 Run it with `vp run dev:backend` (it starts the database first). Use the structured Winston logger, not `console.*` (a lint rule blocks `console` in backend source). See [docs/logging.md](./docs/logging.md).
 
-Tests run with `vp test run --project backend`. Vitest auto-starts Postgres and Redis from `packages/backend/docker-compose.test.yml`. Set `SKIP_TEST_INFRA=1` to skip the Docker spin-up, or `CI=1` when the caller already provides those services.
+Tests run with `vp test run --project backend`. Vitest auto-starts Postgres and Redis from `packages/backend/docker-compose.test.yml`. The Postgres image carries PostGIS, because `gyms` and `user_boards` have a `location` geography that the proximity search queries. Set `SKIP_TEST_INFRA=1` to skip the Docker spin-up, `CI=1` when the caller already provides those services, or `BOARDSESH_TEST_DATABASE_URL` to run the suite against a Postgres of your own.
 
 ## Environment variables
 
--- a/packages/backend/docker-compose.test.yml
+++ b/packages/backend/docker-compose.test.yml
@@ -2,7 +2,12 @@ version: '3.8'
 
 services:
   db:
-    image: postgres:15
+    # PostGIS, because gyms/user_boards carry a `location` geography and the
+    # proximity branch of searchGyms is tested against it. Postgres 18 also lines
+    # this up with the CI service image (18.4 + PostGIS 3.6). Public Docker Hub
+    # build rather than our ghcr one so running the backend suite never needs a
+    # `docker login ghcr.io`.
+    image: postgis/postgis:18-3.6
     ports:
       - '5433:5432'
     environment:
--- a/packages/backend/src/__tests__/global-setup.ts
+++ b/packages/backend/src/__tests__/global-setup.ts
@@ -8,8 +8,14 @@ const REDIS_PORT = 6380;
 const COMPOSE_FILE = fileURLToPath(new URL('../../docker-compose.test.yml', import.meta.url));
 
 const WORKER_DB_PREFIX = 'boardsesh_backend_test';
+// Escape hatch for pointing the suite at a Postgres of your own — see the note
+// on getConfiguredDatabaseUrl in worker-db.ts. When it is set we leave docker
+// alone entirely: whatever the URL names is the caller's to manage.
+const databaseUrlOverride = process.env.BOARDSESH_TEST_DATABASE_URL;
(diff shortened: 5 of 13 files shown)


marcodejongh:
Thanks — three of these were worth acting on, and one of them turned into a real bug.

**`EXCEPTION WHEN OTHERS` too broad** (both reviews) — acted on, though the premise is slightly off: the handler only ever covered the *first* DO block, which contains exactly one statement (`CREATE EXTENSION IF NOT EXISTS postgis`) that we did not write. All the DDL that could carry a typo — columns, indexes, the trigger body — is in the *second* block, which has no handler at all and fails the whole apply loudly. So a bad trigger body was never being swallowed. Tightened anyway: the extension block now prechecks `pg_available_extensions` and only falls back to the broad handler if the precheck said yes and `CREATE` still failed (no privilege, half-installed package). I kept the handler rather than narrowing to `SQLSTATE '0A000'` because the failure mode it guards is "every backend test on this machine dies", and I would rather over-catch one statement than have the suite hinge on a SQLSTATE being identical across Postgres builds.

**CI guard test name** — renamed to `asserts PostGIS is present when running in CI`.

**Timestamp timezone assumption** — good catch, and it was worse than a hidden dependency. Replacing the literal with a field-by-field comparison against the Drizzle branch surfaced a genuine defect: `hours_updated_at` is `timestamp` *without* time zone, `mapRawGymRow` coerces it with `new Date(string)` (resolved in the process's zone), and the Drizzle branch reads the same column as UTC. Under `TZ=Asia/Tokyo` the same gym comes back as `2026-02-03T01:30:00.000Z` from proximity and `2026-02-03T10:30:00.000Z` from text — nine hours, exactly the host offset. Identical on UTC, which is why nobody has hit it. Filed as #4588; not fixed here because it changes what a live resolver returns, same line I drew for the other behaviour questions. Both branches are pinned exactly, so the test is green under Tokyo and the fix collapses two assertions into one equality.

**`dropStaleWorkerDatabases` against the override server** — fair, that deserved to be stated rather than discovered. Added to the `getConfiguredDatabaseUrl` doc comment: point it at a throwaway server only, because whatever it names inherits the whole harness including the `boardsesh_backend_test_w*` sweep.

**`packages/db` tests not in CI** — real and pre-existing (that package runs on Node's test runner and isn't in the Vitest projects), but retiring or wiring up a test file is its own change. The comment I added is the honest description of today's state, and the trigger it covers is now exercised by the backend suite, which does run in CI.

**Comment style** — I could not find a rule in CLAUDE.md banning multi-line comments or file-level docstrings, and the neighbouring real-DB suites this file is modelled on (`gym-directory-fields.test.ts`, `search-gyms-require-slug-sql.test.ts`) are written the same way. Leaving as-is for consistency; happy to trim if that rule does exist somewhere I missed.

PR #4567, opened 2026-08-19
fix(mobile): stop one locked write permanently breaking favorites

### LINKED ISSUE #4331: A dead-lettered favorite/follow key silently swallows every retry
Once a favorite or follow dead-letters, the user can never re-add it. Every later tap writes the local row, gets dropped at enqueue time, and produces nothing to drain — no toast, no badge, no event.

## Mechanism

`enqueue` is `INSERT OR IGNORE` against `idempotency_key TEXT NOT NULL UNIQUE` (`packages/shared/offline-sync/src/mutation-queue/schema.ts`), and favorite/follow keys are deterministic — `add:user_favorites:<board>:<climb>:<angle>`, `add:user_follows:<id>` (`packages/mobile/src/hooks/use-offline-mutations.ts`). The cancel DELETEs in `addFavoriteLocal` / `removeFavoriteLocal` / the follow hooks all match `... AND status = 'pending'`.

So a `dead_letter` row owns that UNIQUE key permanently:

1. Favorite a climb offline → queue row created.
2. It dead-letters (a 4xx, or a retry budget burned).
3. Favorite the same climb again → `user_favorites` row is written locally, `INSERT OR IGNORE` drops the queue row silently, nothing drains.
4. The heart shows filled, the server never hears about it, and the only discoverable trace is More → Sync issues → Retry, which the user has to find on their own.

Ticks are immune — each gets a fresh uuid. Favorites, user follows, setter follows and playlist follows are all exposed.

PR #4329 makes this **countable** (`Offline Mutation Enqueue Suppressed`, reported only when the existing row is a dead letter — a suppression against a `pending` row is legitimate double-tap dedup). It does not fix it.

## Why the obvious fix wasn't taken in #4329

Widening the cancel DELETE to match `dead_letter` would resurrect a mutation the server already rejected as invalid, straight back into the retry loop, on a path with zero measurement at the time. Worth doing only once the suppressed-enqueue counts say how often this actually happens and what the rejections are.

## Options

- **Revive on re-add**: reset the dead-lettered row to `status = 'pending', retry_count = 0` when the user repeats the action. Simple, and a favorite that currently never sticks starts sticking — but it re-sends something the server rejected, so it needs the rejection mix from the new event first.
- **Fresh key per attempt**: append an attempt counter or timestamp, so a repeat never collides. Loses the offline double-tap dedup the deterministic key exists for.
- **Surface it instead**: treat a suppressed enqueue as a user-visible "this didn't sync" affordance rather than a silent write. Overlaps the proactive-nudge question in #4315 and #3621's sign-out warning.

## Acceptance

- A repeat favorite/follow after a dead letter either reaches the server or tells the user it can't.
- The offline double-tap dedup that the deterministic key provides is preserved.
- Regression test with a dead-lettered row in the real DDL (the harness in `packages/mobile/src/hooks/__tests__/use-offline-mutations.test.ts` already sets this scenario up).

Found while instrumenting #4315.

marcodejongh:
## Summary

Closes #4331.

Two defects compose into "a favorite silently never reaches the server, forever".

**1. A local SQLite lock manufactured the dead letter.** `drainMutationQueue`'s per-mutation `try` contains exactly one local write — `markCompleted`, the outbox DELETE — because `processMutation` only calls `graphqlFetch`. So a `database is locked` thrown in there means **the server already accepted the write** and only the local bookkeeping failed. It resolved no HTTP/GraphQL status, `isRetryable` returned false, and the drainer force-dead-lettered the row with `retry_count: 0`.

This is not a hypothesis. Over 90 days there are **7 `Offline Mutation Dead Lettered` events, from 7 distinct users, and all 7 carry `database is locked`** — zero carry a server status. Sentry BOARDSESH-EC has the matching shape (`status: null`, `retryCount: 0`, expo-sqlite's `NativeStatement.finalizeAsync ... database is locked`). That also settles the question #4329 deferred this fix on: there is no "rejection mix" to worry about, because there are no rejections.

**2. The dead row then owned the key forever.** `enqueue` is `INSERT OR IGNORE` against a UNIQUE `idempotency_key`, and favorite/follow keys are deterministic (`add:user_favorites:<board>:<climb>:<angle>`). The cancel DELETEs matched only `status = 'pending'`, so every later favorite for that climb wrote the local row, was dropped at enqueue time, and produced nothing to drain. The heart shows filled, so nobody re-taps — `Offline Mutation Enqueue Suppressed` has fired **zero** times since #4329 shipped. The only escape was More → Sync issues → Retry, which people are not finding: 15 distinct users report at least one dead letter at launch, the oldest 31 days old.

### What changed

- **`error-classification.ts`** — `isRetryable` now returns true for SQLite lock contention. That misclassification is the actual defect: a lock is local contention with no server verdict attached, and holders always let go.
- **`drainer.ts`** — `markCompleted` runs through the existing `runLocalWriteWithRetry` ladder (the DELETE is idempotent, so a retry after a commit-then-throw is a no-op). A lock that still escapes is handled next to the ambiguous-GraphQL-response branch: stay pending, burn no `retry_count`, fire no dead-letter, re-send next pass. `retryableHit` rather than an outright cycle break, because a snapshot import can hold the lock for minutes and the scheduler retries either way.
- **`queue.ts`** — `enqueue(..., { reviveDeadLetter: true })`: when a dead-lettered row owns the key, one UPDATE puts it back to `pending` carrying the **current** payload, `retry_count: 0`, `last_error: NULL`, fresh `created_at`. Off by default, so the tick path (fresh uuid per write, can never collide) is untouched. `EnqueueResult` gains `revived`.
- **`use-offline-mutations.ts`** — the four deterministic-key call sites (favorite add/remove, follow/unfollow) opt in, and their cancel DELETEs widen to `status IN ('pending', 'dead_letter')` so a dead letter from the opposite direction is cleared rather than left to poison the next toggle.
- **Telemetry** — new `Offline Mutation Revived` (PostHog only; a recovery, not a defect). `Offline Mutation Enqueue Suppressed` stays wired as the alarm for a case that should now be unreachable from those call sites.

### The recovery sweep, and its scope

A code-only fix leaves the ~15 users already carrying poisoned rows stuck until they happen to re-toggle that exact favorite. So once per runtime, next to the existing launch backlog gauge, the app reads its dead letters, keeps only the ones whose stored `last_error` matches `isDatabaseLockedError` — the same predicate the drainer and the write ladder use — and puts those back in the queue.

**Say it plainly: this re-sends writes the server never rejected.** That is the whole basis for it being safe, and the measurement above is the evidence: 7 of 7 dead letters in 90 days are local lock errors, none carries a server status, and every handled mutation is idempotent server-side, so a revived write that did land is a no-op.

Scope: only lock-class dead letters. A 4xx, a validation failure, a full disk, or a row with no recorded error is left exactly where it is. Bounded at 50 rows per launch — the real numbers are at most 3 dead letters per device, 73 across the whole fleet. The count rides the existing gauge as `deadLettersRevived` rather than a new event.

Steps 1–2 (stop manufacturing the dead letter) are the fix; the revive hatch and the sweep are the safety net for what already shipped. If a reviewer wants the sweep split out, say so and I'll pull it into its own PR rather than drop it.

## Release Notes

- Favorites and follows now stick. A single unlucky moment of database contention used to strand one for good — the heart stayed filled while the server never heard about it.
- Anything already stuck from that bug is sent on the next launch.

## Test plan

- `vp run test:offline-sync` — 34 files / 733 tests green, including two new suites:
  - `drainer-local-lock.test.ts` — real drainer, real queue SQL, real DDL through node:sqlite, mocking only the network handler. Both production lock shapes (Android's raw-control-byte message and the iOS one) leave the row `pending` / `retry_count 0` with no dead-letter event, a lock that clears on the retry completes the mutation, the still-pending row re-sends on the next drain, and a genuine non-retryable **still** dead-letters (proof the new branch swallowed nothing).
  - `enqueue-revive.integration.test.ts` — revive against the real DDL: payload replaced, status/retry/last_error reset, `created_at` re-stamped, row drainable again; the dead letter untouched without the opt-in; a live pending duplicate still plain dedup.
- `vp run test:mobile` — includes the rewritten `use-offline-mutations` dead-letter suite (all four call sites revive; a dead-lettered add is cleared by a remove and vice versa) and the launch-sweep tests in `outbox-telemetry.test.ts` (lock errors revived, server rejections left alone, sweeps once per runtime, never throws).
- `vp run typecheck:mobile`, `vp check`, `vp run check:mobile-bundle`.
- Not device-tested: reproducing a real `SQLITE_BUSY` on the drain path needs a snapshot import racing a drain. The lock shapes in the tests are the exact production strings from Sentry.

### DIFF
--- a/packages/mobile/src/components/__tests__/offline-sync-bridge.test.tsx
+++ b/packages/mobile/src/components/__tests__/offline-sync-bridge.test.tsx
@@ -83,9 +83,9 @@ vi.mock('../../lib/error-reporting', () => ({
 
 // The gauge itself is unit-tested in offline/__tests__/outbox-telemetry.test.ts;
 // what matters here is WHERE the bridge calls it from — both flag branches, once.
-const reportOutboxBacklogOnceMock = vi.hoisted(() => vi.fn(async (..._args: unknown[]) => {}));
+const recoverAndReportOutboxOnceMock = vi.hoisted(() => vi.fn(async (..._args: unknown[]) => {}));
 vi.mock('../../offline/outbox-telemetry', () => ({
-  reportOutboxBacklogOnce: reportOutboxBacklogOnceMock,
+  recoverAndReportOutboxOnce: recoverAndReportOutboxOnceMock,
 }));
 
 // Stub the settings barrel so the static import graph never pulls
@@ -375,8 +375,8 @@ describe('OfflineSyncBridge — outbox backlog gauge', () => {
   it('reads the backlog once for a signed-in launch', async () => {
     render(<Harness flags={FLAG_ON} queryClient={makeQueryClient()} />);
 
-    await waitFor(() => expect(reportOutboxBacklogOnceMock).toHaveBeenCalledTimes(1));
-    expect(reportOutboxBacklogOnceMock).toHaveBeenCalledWith(fakeDb);
+    await waitFor(() => expect(recoverAndReportOutboxOnceMock).toHaveBeenCalledTimes(1));
+    expect(recoverAndReportOutboxOnceMock).toHaveBeenCalledWith(fakeDb);
   });
 
   it('does not read anything while signed out', async () => {
@@ -384,18 +384,18 @@ describe('OfflineSyncBridge — outbox backlog gauge', () => {
     render(<Harness flags={FLAG_ON} queryClient={makeQueryClient()} />);
 
     await waitFor(() => expect(startBackgroundTrackingMock).toHaveBeenCalled());
-    expect(reportOutboxBacklogOnceMock).not.toHaveBeenCalled();
+    expect(recoverAndReportOutboxOnceMock).not.toHaveBeenCalled();
   });
 
   it('does not re-read when unrelated feature flags change', async () => {
     const queryClient = makeQueryClient();
     const { rerender } = render(<Harness flags={FLAG_ON} queryClient={queryClient} />);
-    await waitFor(() => expect(reportOutboxBacklogOnceMock).toHaveBeenCalledTimes(1));
+    await waitFor(() => expect(recoverAndReportOutboxOnceMock).toHaveBeenCalledTimes(1));
 
     rerender(<Harness flags={FLAG_OFF} queryClient={queryClient} />);
     await waitFor(() => expect(startSyncSchedulerMock).toHaveBeenCalledTimes(1));
 
-    expect(reportOutboxBacklogOnceMock).toHaveBeenCalledTimes(1);
+    expect(recoverAndReportOutboxOnceMock).toHaveBeenCalledTimes(1);
   });
 
   it('waits for a stamped schema before reading the outbox', async () => {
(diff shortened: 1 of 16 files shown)


marcodejongh:

Conflicting: packages/mobile/src/components/offline-sync-bridge.tsx.

PR #4570, opened 2026-08-19
fix(mobile): show that the workout shelf scrolls

### LINKED ISSUE #4278: The workouts on iOS are hidden it's not obvious that there are other options.
Take a look at this image below

Image

### The problem
It took me 3 sessions to know that I could scroll on the workouts to pick more workouts. I'd reccomend a UI rework of the page, I don't think it's obvious. If we keep the same UI I have some suggestions:

1. Size the tile in a way that the next one is peeking just a tiny bit from the side so the users know that they can just scroll to the side to see more options
2. Make the list a grid so user can scroll down to see the workout options

marcodejongh:
## Summary

The pre-session workout shelf (`WorkoutTypeShelf`) hid the fact that it scrolls. On the reporter's 375pt device a hardcoded `TILE_WIDTH = 168` fit tiles edge-to-edge with no partial tile in view (tile 3 started at x=376, 1pt off-screen), and `showsHorizontalScrollIndicator={false}` disabled the platform's own "more content" cue. It took the reporter three sessions to find the other workout types.

- Tile width is derived from the shelf's **own measured container**, sized so a fraction of the next tile stays in view. The row ends mid-tile, and that cut-off tile is the affordance.
- The rail **flashes its scroll indicator** when the Record screen comes forward. iOS paints the indicator only during an active drag, so at rest a rail that scrolls looked exactly like one that doesn't — on wide phones that was the *only* thing left to change (see below).
- `snapToInterval` / `snapToAlignment="start"` / `decelerationRate="fast"` settle the rail on tile boundaries instead of leaving a tile straddling the edge after a fling.
- Kept the `ScrollView` (5 static tiles, no virtualization needed) and left `GeneratorPickerCard`'s shelf logic, the chart rendering and tile content untouched.

### Why QA saw no change on `e77380e`

The geometry fix only moves anything where the row previously ended flush. It didn't on the tester's phone:

| Device width | Before: tile / peek | After `e77380e`: tile / peek |
|---|---|---|
| 375 (SE, mini, iPhone X–11 Pro) | 168 / **0** | 143 / 49 |
| 390 (14, 15, 16) | 168 / 14 | 149 / 52 |
| 402 (16 Pro) | 168 / 26 | 154 / 54 |
| **430 (14 Pro Max, 15 Plus, 16 Plus)** | **168 / 54** | **166 / 58** |

At 430pt a 54pt slice of the next tile was already showing before this PR, and the row still read as finished — which is the reporter's whole complaint. Two points of tile and four of peek is not a visible difference, so "I don't see any change?" was accurate on that hardware.

The indicator flash is the part that lands on every width. It's the platform's own affordance and it's visible at rest-time-zero rather than only mid-drag.

### The 440pt tile cliff

440pt (iPhone 16 Pro Max) also had a regression of its own: two 168pt tiles overshoot the cap by 2pt, so the ladder stepped up to three tiles — shrinking every tile from 168 to 116 and *cutting* the peek from 64pt to 40. Capping now wins whenever the leftover already lands in the peek band (a fifth to a half of a tile). The ladder still shrinks the tile where capping would leave a sliver: 580pt fits three capped tiles with 24pt to spare, 744pt fits four with 8pt, and both still shrink.

### Why the container and not the window

An earlier revision read `useWindowDimensions()`. That's wrong on the iPad adaptive shell, where the shelf renders inside `shellContent` — a flex column between `TabletSidebar`, `IpadPlayPane` and `IpadWallColumn` (`app/(tabs)/_layout.tsx`). On an 11" landscape iPad the window is 1194pt while the shelf gets 399, and the window-derived tile came out **484pt — wider than the pane it lives in**. The shelf measures itself with `onLayout` and only falls back to the window for the first paint.

### How the width is picked

Visible row at rest is `leadingInset + n*(tile + gap) + peek` — only the leading 16pt inset is on screen; the trailing one sits at the far end of the content.

`tile = round((containerWidth - 16 - n*12) / (n + 0.35))`, with `n` (whole tiles in view) raised from 2 until the tile stops exceeding `TILE_MAX_WIDTH = 168` — the width the shelf shipped with. So wide surfaces gain *more* tiles, not bigger ones.

| Surface | Shelf container | Tile | Whole tiles | Peek |
|---|---|---|---|---|
| iPhone SE / mini (375) | 375 | 143 | 2 | 49 |
| iPhone 14/15/16 (390) | 390 | 149 | 2 | 52 |
| iPhone 16 Plus (430) | 430 | 166 | 2 | 58 |
| iPhone 16 Pro Max (440) | 440 | 168 | 2 | 64 |
| iPad 11" landscape (1194) | 399 | 153 | 2 | 53 |
| iPad 11" portrait (834) | 418 | 161 | 2 | 56 |
| iPad 13" landscape (1366) | 485 | 129 | 3 | 46 |
| iPad 13" portrait (1024) | 580 | 158 | 3 | 54 |
| iPad mini portrait (744) | 648 | 134 | 4 | 48 |

Where the container is wide enough for all five tiles at 168, the row doesn't scroll and there is nothing to advertise.

### Alternative considered, not built

The reporter also suggested turning the row into a grid (#2 in the issue). That's the larger change and a different visual language for the page. This PR takes the smaller "peek rail" fix; say the word and I'll redo it as a grid.

## Test plan

1. Open Record tab → workout row's scroll bar flashes.
2. Look at the row's right edge → a tile is cut off.
3. Flick the row sideways → it stops on a tile edge.
4. Leave the tab, come back → the bar flashes again.

## Risk

Risk: 2/5 — sizing and a scroll-indicator flash on one existing row; no data, no network, no native code.

## Testing

- `vp check` — 0 errors.
- `vp run test:mobile` — 786 files, 8154 passed.
- Every new guard was checked against a reverted fix:
  - dropping `flashScrollIndicators()` → the flash test fails
  - dropping the timer cleanup → the unmount test fails
  - removing the cap-vs-step-up rule → the 440pt test fails
  - capping unconditionally → 4 tests fail at 580pt and 744pt
- No iOS simulator in this environment, so the tables above are there to sanity-check against the code.

Closes #4278

## Release Notes

The workout picker on the session start screen now shows you there's more to it: the scroll bar flashes when you land on the screen, the next workout tile is cut off at the edge instead of the row looking complete, and a flick settles on a tile boundary. On iPad the tiles size to the pane they're in rather than the whole window.

### DIFF
--- a/packages/mobile/src/components/session-screen/pre-session/WorkoutTypeShelf.tsx
+++ b/packages/mobile/src/components/session-screen/pre-session/WorkoutTypeShelf.tsx
@@ -15,8 +15,9 @@ const CHART_HEIGHT = 82;
 
 // Mirrors the ScrollView's contentContainerStyle below, so the peek math
 // below accounts for exactly the padding/gap the row actually renders.
+// Exported so tests can assert against it instead of hardcoding the token.
 const SHELF_HORIZONTAL_INSET = spacing[4];
-const TILE_GAP = spacing[3];
+export const TILE_GAP = spacing[3];
 
 // Deliberately not a whole number: sizing tiles so a whole number of them
 // fills the screen edge-to-edge is what caused #4278 — on a 375pt phone the
--- a/packages/mobile/src/components/session-screen/pre-session/WorkoutTypeShelf.tsx
+++ b/packages/mobile/src/components/session-screen/pre-session/WorkoutTypeShelf.tsx
@@ -1,5 +1,5 @@
-import { memo } from 'react';
-import { StyleSheet, View } from 'react-native';
+import { memo, useMemo } from 'react';
+import { StyleSheet, useWindowDimensions, View } from 'react-native';
 import { ScrollView } from 'react-native-gesture-handler';
 import type { ColoredBar } from '../../you/profile-chart-colors';
 import { StackedBarChart } from '../../you/YouCharts';
@@ -11,8 +11,36 @@ import { useTheme } from '../../../providers/theme-provider';
 import { withAlpha } from '../../../theme/colors';
 import { borderRadius, spacing } from '../../../theme/tokens';
 
-const TILE_WIDTH = 168;
 const CHART_HEIGHT = 82;
+
+// Mirrors the ScrollView's contentContainerStyle below, so the peek math
+// below accounts for exactly the padding/gap the row actually renders.
+const SHELF_HORIZONTAL_INSET = spacing[4];
+const TILE_GAP = spacing[3];
+
+// Deliberately not a whole number: sizing tiles so a whole number of them
+// fills the screen edge-to-edge is what caused #4278 — on a 375pt phone the
+// old hardcoded TILE_WIDTH happened to fit exactly 2 tiles with no partial
+// tile showing, so the row looked complete even though 3 more workout types
+// were off-screen. Targeting "a bit past 2 tiles" means the row always ends
+// mid-tile on common phone widths, so the cut-off tile itself signals there's
+// more to scroll to, on top of the restored scroll indicator below.
+const VISIBLE_TILES_TARGET = 2.35;
+
+/** Pure so it can be unit tested at fixed device widths without rendering. */
+export function computeTileWidth(windowWidth: number): number {
+  const fullGapCount = Math.floor(VISIBLE_TILES_TARGET);
+  const availableWidth = windowWidth - SHELF_HORIZONTAL_INSET * 2 - TILE_GAP * fullGapCount;
+  return Math.round(availableWidth / VISIBLE_TILES_TARGET);
+}
+
+/** Tile width derived from the viewport so a partial tile always peeks at
+ *  the trailing edge, instead of a hardcoded width that only avoided a full
+ *  peek by accident on some screen sizes and not others. */
+function useTileWidth(): number {
+  const { width: windowWidth } = useWindowDimensions();
+  return useMemo(() => computeTileWidth(windowWidth), [windowWidth]);
+}
 
 export type WorkoutTypeShelfItem = {
   key: string;
@@ -29,22 +57,28 @@ type WorkoutTypeShelfProps = {
 };
 
 export const WorkoutTypeShelf = memo(function WorkoutTypeShelf({ items }: WorkoutTypeShelfProps) {
+  const tileWidth = useTileWidth();
+  const snapInterval = tileWidth + TILE_GAP;
+
   return (
     <ScrollView
       horizontal
-      showsHorizontalScrollIndicator={false}
+      showsHorizontalScrollIndicator
       contentContainerStyle={styles.content}
       keyboardShouldPersistTaps="handled"
       nestedScrollEnabled
+      snapToInterval={snapInterval}
+      snapToAlignment="start"
+      decelerationRate="fast"
     >
       {items.map((item) => (
-        <WorkoutTypeTile item={item} key={item.key} />
+        <WorkoutTypeTile item={item} key={item.key} tileWidth={tileWidth} />
       ))}
     </ScrollView>
   );
 });
 
-function WorkoutTypeTile({ item }: { item: WorkoutTypeShelfItem }) {
+function WorkoutTypeTile({ item, tileWidth }: { item: WorkoutTypeShelfItem; tileWidth: number }) {
   const { systemColors, brandColors } = useTheme();
   const selectedBackground = withAlpha(brandColors.primary, 0.08);
 
@@ -59,6 +93,7 @@ function WorkoutTypeTile({ item }: { item: WorkoutTypeShelfItem }) {
       style={[
         styles.tile,
         {
+          width: tileWidth,
           backgroundColor: item.selected ? selectedBackground : systemColors.secondaryBackground,
           borderColor: item.selected ? brandColors.primary : systemColors.separator,
         },
@@ -103,11 +138,10 @@ function WorkoutTypeTile({ item }: { item: WorkoutTypeShelfItem }) {
 
 const styles = StyleSheet.create({
   content: {
-    paddingHorizontal: spacing[4],
-    gap: spacing[3],
+    paddingHorizontal: SHELF_HORIZONTAL_INSET,
+    gap: TILE_GAP,
   },
   tile: {
-    width: TILE_WIDTH,
     borderWidth: StyleSheet.hairlineWidth,
     borderRadius: borderRadius.lg,
     paddingHorizontal: spacing[3],
--- a/packages/mobile/src/components/session-screen/pre-session/WorkoutTypeShelf.tsx
+++ b/packages/mobile/src/components/session-screen/pre-session/WorkoutTypeShelf.tsx
@@ -13,71 +13,48 @@ import { borderRadius, spacing } from '../../../theme/tokens';
 
 const CHART_HEIGHT = 82;
 
-// Mirrors the ScrollView's contentContainerStyle below, so the peek math
-// accounts for exactly the padding/gap the row actually renders.
-// Exported so tests can assert against it instead of hardcoding the token.
+// Mirrors the ScrollView's contentContainerStyle below so the peek math uses
+// the padding/gap the row actually renders. Exported for the tests.
 export const SHELF_HORIZONTAL_INSET = spacing[4];
 export const TILE_GAP = spacing[3];
 
-// Deliberately not a whole number of tiles: sizing tiles so a whole number of
-// them fills the row edge-to-edge is what caused #4278 — on a 375pt phone the
-// old hardcoded TILE_WIDTH happened to fit exactly 2 tiles with no partial tile
-// showing, so the row looked complete even though 3 more workout types were
-// off-screen. Leaving a fixed fraction of the next tile in view means the
-// cut-off tile itself signals there's more to scroll to.
+// #4278: a whole number of tiles filled a 375pt row exactly, so the shelf looked
+// complete with 3 workout types off-screen. A partial tile is the scroll cue.
 const PEEK_FRACTION = 0.35;
 
 /** Never squeeze below two whole tiles, however narrow the container gets. */
 const MIN_WHOLE_TILES = 2;
 
-/** Tiles stop growing here. The chart inside each tile is drawn for a
- *  phone-sized tile; a 480pt-wide one just stretches it. 168 is the width the
- *  shelf shipped with before #4278, which read fine on iPad — wide containers
- *  gain more tiles in view rather than bigger ones. */
+// Tiles stop growing here (the pre-#4278 width): the chart is drawn for a
+// phone-sized tile, so wide containers gain more tiles rather than bigger ones.
 export const TILE_MAX_WIDTH = 168;
 
-/** Floor for degenerate containers (a Slide Over pane, a first layout pass that
- *  reports a hairline width) so a tile never collapses to nothing. */
+// Floor for degenerate containers (Slide Over, a hairline first layout pass).
 export const TILE_MIN_WIDTH = 104;
 
-/**
- * Tile width for a shelf `containerWidth` points wide holding `itemCount` tiles.
- *
- * Only the LEADING inset sits between the viewport edge and the first tile — the
- * trailing one lives at the far end of the content, past everything the user can
- * see — so the visible row is `inset + n*(tile + gap) + peek`.
- *
- * Widens the whole-tile count until tiles stop exceeding {@link TILE_MAX_WIDTH},
- * which keeps the peek proportional at every width instead of clamping the tile
- * and letting the leftover land back on a whole number of tiles (the #4278 bug,
- * one container width over).
- *
- * Pure so it can be unit tested at fixed widths without rendering.
- */
+// Visible row is `inset + n*(tile + gap) + peek` — only the LEADING inset is on
+// screen, the trailing one sits past the end of the content.
+
 export function computeTileWidth(containerWidth: number, itemCount: number): number {
   const availableWidth = containerWidth - SHELF_HORIZONTAL_INSET;
-  // Showing every item but one as a whole tile is the widest the ladder can go
-  // and still have a tile left over to peek with.
+  // All but one item whole is the widest the ladder can go and still have a peek.
   const maxWholeTiles = Math.max(MIN_WHOLE_TILES, itemCount - 1);
 
+  // Widen the tile count, not the tile: clamping the tile instead would park the
+  // leftover back on a whole number of tiles — the #4278 bug, one width over.
   for (let wholeTiles = MIN_WHOLE_TILES; wholeTiles <= maxWholeTiles; wholeTiles += 1) {
     const tileWidth = Math.round((availableWidth - TILE_GAP * wholeTiles) / (wholeTiles + PEEK_FRACTION));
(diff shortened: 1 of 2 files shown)


marcodejongh:
## QA review — needs rework

_(Edited in place: this comment first went out with the body of an unrelated aurora-sync review — a scratch-file mixup on my side. If you got that in a notification email, disregard it; the review below is the real one for #4570.)_

The phone fix is right and the diagnosis checks out: at 375pt the old `TILE_WIDTH = 168` puts tile 3 at x=376, i.e. exactly 1pt off-screen — a genuine zero peek on the reporter's iPhone X-class 1125x2436 screenshot. But `useWindowDimensions()` is the wrong width source in this app, and on iPad it turns a working shelf into a broken one.

### 1. Blocking — tile width is derived from the window, but the shelf lives in a pane that is much narrower on iPad

`WorkoutTypeShelf.tsx:42` reads `useWindowDimensions().width` and treats it as the shelf's width. On the iPad adaptive shell that is false: `app/(tabs)/_layout.tsx:270-299` lays the shell out as a row of `TabletSidebar` (`SIDEBAR_WIDTH = 96`, `theme/layout.ts`) + `shellContent` (flex:1 — which hosts `record/index.tsx` -> `SessionScreen` -> `PreSessionView` -> `GeneratorPickerCard.tsx:631` -> this shelf) + `IpadPlayPane` (`resolveDetailPaneWidth`) + `IpadWallColumn` (`WALL_COLUMN_WIDTH = 300`).

Running the repo's own resolvers (`theme/size-class.ts`) against real device widths:

| Surface | Shelf's actual container | New tile | New peek | Old tile 168 -> peek |
|---|---|---|---|---|
| iPhone 375 | 375 | 136 | **63** OK | -1 (the bug) |
| iPhone 390 | 390 | 142 | 66 OK | 14 |
| iPhone 430 | 430 | 159 | 72 OK | 54 |
| iPad mini portrait 744 | 648 | 293 | 22 | 272 |
| iPad 11" portrait 834 | 418 | **331** | **-284** BROKEN | 42 |
| iPad 13" portrait 1024 | 580 | **412** | **-284** BROKEN | 204 |
| iPad 11" landscape 1194 | 399 | **484** | **-609** BROKEN | 23 |
| iPad 13" landscape 1366 | 485 | **557** | **-669** BROKEN | 109 |

On every iPad config where the detail pane mounts, **a single tile is now wider than the entire pane it renders in** (a 484pt tile in a 399pt column on an 11" landscape). You get one clipped tile with a chart stretched to ~3x its designed width, and the peek affordance this PR is entirely about is gone. Before this PR iPad was fine — 168pt tiles gave 2 tiles plus a healthy peek at every one of those widths. This is a straight regression on a supported surface, and it ships OTA to every existing build (`ota-check`: "Ships over-the-air"), iPads included.

Android tablets and Expo web hit the same path — `resolveIsTablet` (`size-class.ts`) admits any screen with a short side >= 600dp, so a desktop browser at `/app` gets the same oversized tiles.

Fix: measure the shelf's own container (`onLayout` on the `ScrollView` or a wrapper `View`) instead of the window, and/or clamp — `Math.min(TILE_MAX_WIDTH, computeTileWidth(w))` with `TILE_MAX_WIDTH` near the old 168-200 — so wide surfaces keep sane tiles. A clamp alone fixes the "giant tile" half; only measuring fixes the "wrong container" half. `useDeviceLayout()` already exists if you'd rather branch on `widthClass`.

### 2. The tests would not catch #1, and would not catch the core fix being reverted either

- No test asserts the computed width ever reaches a tile. `__tests__/workout-type-shelf.test.tsx:40` was changed specifically to serialize `style` into `data-style` on the `PressableSurface` mock — and then nothing reads it. Delete `width: tileWidth` from `WorkoutTypeShelf.tsx:97` and all 11 tests still pass (verified locally: `vp test run --project mobile --reporter=agent workout-type-shelf` -> 11 passed).
- `computeTileWidth` is only exercised at 375/390/430. A case at 1024 or 1366 would have surfaced the tablet problem the moment anyone compared it against the real pane width.
- The peek invariant at `workout-type-shelf.test.tsx:139-147` models a layout the component doesn't render. It computes `peek = (w - 32) - (2*tile + gap)`; the rendered peek is `w - (16 + 2*tile + 2*gap)` — it drops the second gap and subtracts a trailing inset that sits at the end of the content, not at the viewport edge. At 375 it reports 59; the real value is 63. The two errors partly cancel so it passes, but it isn't measuring what it claims.

The two ScrollView prop tests (`:107`, `:114`) do fail against the old code, so they're honest — they're just prop-passthrough assertions, which is all they can be.

### 3. Smaller things

- **`showsHorizontalScrollIndicator` does nothing for the reported problem on iOS.** UIScrollView indicators are only visible during an active scroll gesture (absent an explicit `flashScrollIndicators` call). At rest — which is the whole complaint in an issue titled "The workouts on iOS are hidden" — the row looks identical. Worth keeping, but the release note "and the scroll bar is back" oversells it; the peek is doing 100% of the work.
- **PR body arithmetic is still wrong.** claude-review flagged it and it wasn't fixed: the body states `(windowWidth - 32 - 12) / 2.35`, the code subtracts two gaps (`- 56`). The peek column (59/62/68) is the wrong-model number from #2 — the real values are 63/66/72. Since the body explicitly asks a reviewer to sanity-check the arithmetic in lieu of screenshots, it should match the code.
- **Local commit `e33bfaf` is not pushed.** PR head is `9aba901`; the worktree carries one extra commit ("address claude-review feedback...", exporting `TILE_GAP` and adding the `beforeEach` reset). What CI is green on is not what's in the worktree — push it.

### What I verified

- `vp test run --project mobile --reporter=agent workout-type-shelf` — 11 passed, 2 files.
- `gh pr checks 4570` — green on `9aba901` (lint, typecheck, test-default x2, mobile-bundle, commit-lint, release-notes, ota-check all pass).
- Scope is clean: two files, mobile only, no new i18n strings, no BLE, design tokens throughout (`spacing[4]`/`spacing[3]`), no `any`. No Fable review needed.
- Traced the shelf's real container: `GeneratorPickerCard` -> `PreSessionView` (`styles.header` has no horizontal padding, so the window width is correct on phones) -> `SessionScreen` -> `(tabs)/record` -> the iPad shell.

### For Marco to decide

The issue offered two options and this PR built #1 (peek rail). That call looks right for the size of the change. But if you'd rather have #2 (grid), the tablet problem above largely disappears with a grid — worth deciding before another round on the rail math.

marcodejongh:
Reworked on `a9209b5`. The iPad finding was real and I reproduced every number in it.

**1. Tablet regression — fixed.** Confirmed the trace: `GeneratorPickerCard` renders the shelf in an unpadded `View`, so on phones the window *is* the container, but on the adaptive shell it's `shellContent`, sized by `flex:1` between `TabletSidebar` (96), `IpadPlayPane` (`resolveDetailPaneWidth`) and `IpadWallColumn` (300). At 1194 that's a 399pt pane holding a 484pt tile.

The shelf now measures itself with `onLayout` and only uses the window as the first-paint estimate. I also capped tiles at `TILE_MAX_WIDTH = 168` (the pre-#4278 width) — but *not* as a bare clamp: the whole-tile count `n` climbs from 2 until the tile stops exceeding the cap, so wide surfaces get more tiles rather than bigger ones. A bare clamp would have parked the leftover back on a whole number of tiles at other container widths (zero peek at 380, 560, 740, 920…) — the same bug, one width over.

**2. Tests — rewritten.** Tile width is now read back off the rendered tile (`data-style`), `onLayout` is driven directly, and the peek invariant runs across 15 widths: phones 320/375/390/430/440, the iPad panes 399/418/485/580/648, and full-width 744/834/1024/1194/1366. I checked all four reverts you'd want caught:

| revert | result |
|---|---|
| delete `width: tileWidth` from the tile | 4 tests fail |
| back to `useWindowDimensions()` | container test fails |
| back to the unbounded 2.35 formula | 8 tablet widths fail |
| back to hardcoded `168` | 375pt (the original bug) + 320 + 744 fail |

**3. Peek model — you were right, and the formula had the same error.** The visible row is `leadingInset + n*(tile + gap) + peek`: only the leading inset is on screen, the trailing one is at the far end of the content. The old test dropped a gap *and* subtracted a trailing inset that isn't in the viewport; the old `computeTileWidth` subtracted both insets for the same reason. Both now use the single-inset model, which moves the 375pt tile from 136 to 143 (peek 49).

**4. `showsHorizontalScrollIndicator`** — agreed, kept but demoted. The release note no longer leads with "the scroll bar is back", and the PR body says outright that on iOS the indicator is gesture-only and the peek does the work.

**5. PR body arithmetic** — rewritten against the code, with the tablet rows included.

**6. Unpushed commit** — pushed; the branch is rebased on `origin/main` and `a9209b5` is the head.

Nothing rebutted — all six held up.

marcodejongh:
## QA follow-up on `f1724e6` — re-verified the rework, one nit fixed, one rebutted

I came in with the first review's four findings and checked each against the current head rather than taking them on trust. All four were already closed by `a9209b5`; I re-verified them by breaking the code and watching the tests, rather than by reading the diff.

| Finding (first review) | State at `f1724e6` | How I checked |
|---|---|---|
| Tile width derived from the window, not the pane | Fixed | Restored `containerWidth: windowWidth` → `sizes tiles from the measured container, not the window` fails |
| No test asserts the width reaches a tile | Fixed | Deleted `width: tileWidth` from the tile style → **4** tests fail |
| Peek invariant models a layout that isn't rendered | Fixed | `shelfLayout` now uses `containerWidth - (inset + n*(tile + gap))`; I re-derived it from `contentContainerStyle` — content is `[16][tile][12]…[16]`, so tile *n+1* starts at `16 + n*(tile+gap)` and only the leading inset is on screen. Matches `computeTileWidth`'s own model exactly (`16 + n*tile + n*gap + 0.35*tile = containerWidth`) |
| Only 375/390/430 exercised | Fixed | 15 widths, 320→1366. Restored the unbounded `/2.35` formula → **9** fail, all of them 440pt and up |

Full suite: `vp run test:mobile` → 684 files, 6570 passed. `vp run typecheck:mobile` clean, `vp check` 0 errors.

### Fixed — comment verbosity

Condensed the blocks over `computeTileWidth`, `useShelfWidth`, `PEEK_FRACTION` and the two test helpers; 57 lines out, 25 in, no behaviour change. Kept only what a reader can't re-derive from the code: the iPad `shellContent` pane, the single-inset peek model, and why the tile *count* widens instead of the tile.

One correction on the citation, though: CLAUDE.md has no "one short line max" rule — `grep -rn "one short line max\|multi-line comment"` across every `.md` in the repo returns nothing, and there's no `packages/mobile/CLAUDE.md` or user-level file either. The trim was worth doing on its own merits; the rule it was attributed to doesn't exist.

### Rebutted — "first-paint tile size jump on iPad"

The numbers are right — `computeTileWidth(1194, 5)` is 168 and `computeTileWidth(399, 5)` is 153, so a 1194pt-window iPad does paint one frame 15pt wide before `onLayout` lands. The proposed remedy doesn't work, though: moving `onLayout` to a wrapper `View` at the parent doesn't measure any earlier. A parent and its child are laid out in the same pass and both `onLayout` callbacks fire after that pass commits, so the wrapper would hand back 399 at exactly the moment the ScrollView does. Nothing in RN can report a flex-sized box before the first layout, so any fix here is really "don't paint until measured" — a blank frame in place of a slightly-wide one, which is worse.

Worth noting the estimate frame is a *valid* layout, not a broken one: 168pt tiles in a 399pt pane give `16 + 2×180 = 376`, peek 23pt — still above `TILE_GAP`, so the scroll cue is present from the first frame. And the fallback only ever over-estimates (the window bounds the pane), so the correction is always a shrink, never a clipped tile. Leaving it as is.

marcodejongh:
## QA decline on `e77380e` — reproduced the cause, and it wasn't the bundle

Checked the bundle first: the OTA preview publish for `e77380e` ran 07:59–08:08 on 2026-09-05 and the iOS group landed at 08:04:44. The verdict's `Bundle published 08:04:01Z` sits inside that window, so the phone was running this PR's code, not a fallback to the production channel.

**So the report is accurate: on a wide phone this PR changed nothing you could see.** The geometry fix only moves anything where the row previously ended flush, and on the newer big iPhones it never did:

| Width | Before | After `e77380e` |
|---|---|---|
| 375 (X / XS / 11 Pro — the reporter's phone) | 168pt tile, **0pt peek** | 143 / 49 |
| 390 (14/15/16) | 168 / 14 | 149 / 52 |
| **430 (14 Pro Max, 15 Plus, 16 Plus)** | **168 / 54** | **166 / 58** |

Two points of tile and four of peek. There was already a 54pt slice of the next tile showing on that hardware — and the row still read as finished, which is the reporter's original complaint. A bigger peek was never going to fix that.

### What changed

**The rail now flashes its scroll indicator when the Record screen comes forward.** iOS paints the indicator only during an active drag, so at rest a rail that scrolls looks identical to one that doesn't — the earlier round already noted the indicator "does nothing for the reported problem on iOS", and this is the part that fixes it. It's the platform's own affordance, it fires on every width, and it re-fires each time you come back to the tab.

**Fixed a cliff at 440pt (iPhone 16 Pro Max).** Two 168pt tiles overshoot the cap by 2pt, so the ladder stepped up to three — shrinking every tile from 168 to 116 and cutting the peek from 64pt down to 40. Capping now wins where the leftover already lands in the peek band; the ladder still shrinks the tile where capping would leave a sliver (580pt has 24pt to spare, 744pt has 8pt — both still shrink).

**Added the missing `## Test plan` and `## Risk`.** The `pr-test-plan` gate was red and the description had no steps in it, so there was nothing on the phone telling you what to look at. That's on me. The plan is four steps now, and step 1 is the flash.

Also tightened the peek invariant: it asserted `peek > TILE_GAP`, which 834pt satisfied by arithmetic accident at 58% of a tile while the code comment claimed the row didn't scroll there. It now asserts a floor of a fifth of a tile and a ceiling below a whole one, and the comment says what actually happens.

Every new guard was mutation-tested — dropping the flash, dropping its cleanup, removing the cap rule, and capping unconditionally each turn tests red. `vp run test:mobile` 786 files / 8154 passed, `vp run typecheck:mobile` clean, `vp check` 0 errors.

marcodejongh:
Fixed one of the three, answering the other two.

**Fixed — test constants duplicated.** `MIN_PEEK_FRACTION` / `MAX_PEEK_FRACTION` are exported now and the test imports them, so retuning the band can't leave the assertions checking the old one. `d5c44027b`.

**By design — the 4pt sliver on the left after a snap.** It isn't a coincidence, it's `SHELF_HORIZONTAL_INSET - TILE_GAP` by construction. At snap offset `k*(tile + gap)`, tile *k*'s left edge lands at viewport `16 + k*(tile+gap) - k*(tile+gap)` = exactly the inset, the same place tile 0 sits at rest — that identity holds for any values of the two tokens, so the rhythm is derived, not lucky. What varies with the tokens is the sliver, which is the *previous* tile's trailing edge and reads as "there's more behind you" — the mirror of the peek this PR exists to create.

`snapToOffsets` shifted by the inset would put the active tile at viewport x=0, flush against the screen edge with no gutter, which is worse. Leaving it.

**Known — first-paint reflow on tablets.** Same finding as the round before this one and the rebuttal still holds: a parent `View`'s `onLayout` fires in the same pass as the ScrollView's, so moving the measurement up doesn't report any earlier. Nothing in RN can size a flex box before the first layout, so the only alternative is not painting until measured — a blank frame instead of a slightly-wide one. The window fallback only ever over-estimates (the window bounds the pane), so the correction is always a shrink, never a clipped tile.

CI is green on `6b8b3d99a` including the test-plan gate and i18n, both of which were red before this round.

Issue #4496, opened 2026-08-15
[Bug: UX] Climb creator: hold selection is inaccurate and imprecise


ES-Alexander:
Clicking on holds in the climb creator is quite inconsistent, and seems to be avoiding selection for some full hold sets on the board.

As an example, the original kilter board seems to only allow selecting the screw-on holds, regardless of whether you click one of them or click directly on one of the larger bolt-on pieces. To make matters worse, the whole selection region seems to be somewhat offset, so for example clicking on the left side of a screw-on hold on the kickboard will select the hold to the left of it instead:

https://github.com/user-attachments/assets/19dab09c-1344-45ec-888e-53cfae88fd18

I've confirmed this happens on the iOS app (except if you zoom in at all, in which case hold selection works fine), and also on the web app (where it's seemingly not possible to zoom in on the board, so there's no convenient workaround).

marcodejongh:
🤖 Claimed by bug-batch loop `moarbugs100500` (batch 1, session 43ea5be5).

Investigation + fix in progress. If another automation loop is already on this, say so here and I will stand down. Loop `moarbugs10000` holds #4412–#4414, #4416, #4494, #4519, #4523, #4528, #4529, #4532 — I am staying off those.

Issue #2687, opened 2026-06-10
[P2] Cannot paint holds while zoomed in the create-climb editor (pan overlay swallows every paint tap)


marcodejongh:
**User impact:** Climbers who pinch-zoom the create board to accurately place a hold in a dense cluster find that taps do nothing while zoomed. They must tap "Reset zoom" to paint — which defeats the entire purpose of zooming. Zoom-to-paint, the obvious workflow for placing small/dense holds precisely, silently doesn't work. Long-press (the role sheet) is equally dead while zoomed. **This deserves prioritization before store release if creators rely on zoomed painting.**

### Verified mechanism
While `isZoomed`, an `absoluteFill` `<View>` carrying only a bare `Gesture.Pan` (`InteractiveCreateBoard.tsx:123-127`; `zoomPanGesture` in `use-zoom-pan-gesture.ts:200-225` has no Tap composed) mounts above `HoldTargetLayer`. The overlay View has no `pointerEvents` override (defaults to `auto`). RNGH only collects handlers on the hit view and its ancestors; the per-hold Tap/LongPress detectors (`HoldTarget.tsx:40-56`) are on sibling views beneath the overlay and are never candidates — and a stationary tap cannot activate a Pan, so paint AND long-press are dead while zoomed. This was knowingly traded, not an unnoticed regression: the original editor (d767cf0aa) explicitly supported tap-while-zoomed via `Gesture.Simultaneous` + nested detectors; the sheet rebuild b05a3cc69 replaced it with the Play Drawer overlay model to stop the pan eating idle 1-finger drags (drawer scroll/close) and added a Reset-zoom escape hatch. The pattern is borrowed from `SwipeBoardCarousel`, but there the board is non-interactive (`pointerEvents='none'`), so the blocking overlay is harmless; on the create board the per-hold tap IS the core interaction.

### Repro
Open the create drawer (FAB on climbs list) → pinch-zoom the board past `ZOOM_THRESHOLD` so `isZoomed=true` (Reset zoom pill appears) → tap any hold dot. Expected: ring painted at zoomed precision. Actual: nothing (overlay Pan fails on a stationary tap; `HoldTarget` detectors never receive the touch). Long-press is equally dead. Must tap "Reset zoom" to paint.

### Verified fix design
[...]

PR #4571, opened 2026-08-19
fix(mobile): resolve create-climb hold taps by proximity at rest

marcodejongh:
## Summary

Refs #4496

At fit-to-screen, the create board resolved a tap by **view z-order over grossly oversized square hit boxes**, not by proximity.

`holdGeometry` inflates every hold's tap target to `Math.max(ringDiameter * 1.6, 44)` px (`holdLayout.ts`, `MIN_TAP_DIAMETER = 44`) so small holds stay reachable, and `HoldTarget` rendered that as an absolutely-positioned **square** `View` with its own `GestureDetector`. Overlapping sibling views are arbitrated by render order, so the last hold in the list won every touch inside its square, however far off-centre the finger landed.

### The measurement

Sets are concatenated in `set_ids` order (`board-details.ts`), and Kilter layout 1 is `[1 Bolt Ons, 20 Screw Ons]` — so screw-ons render last and sit on top. At a 340 dp board (size 10, 1080 px source):

| | |
| --- | --- |
| painted ring | 18.9 dp |
| hit circle | **44 dp** (the a11y floor wins) |
| nearest-neighbour spacing | 13.4 dp |
| other hold centres inside one hit circle | 5 (median), 6 (max) |
| bolt-on centres inside *some* screw-on's hit circle | **289 / 323 (90%)** |

So tapping the dead centre of a bolt-on handed the touch to a screw-on nine times out of ten. That is all three reported symptoms: only screw-ons get selected, the hit region feels shifted, zooming fixes it (the squares stop overlapping).

### What changed

- New `useRestHoldTapGesture` (`packages/mobile/src/components/create-climb/use-rest-hold-tap-gesture.ts`): one full-bleed overlay that resolves an at-rest tap to the **nearest hold centre** via the existing `resolveHoldAtPoint`. It's the at-rest twin of `useZoomedHoldTapGesture` and shares the same hit circles, so a point resolves to the same hold whether you're zoomed or not.
- `InteractiveCreateBoard` mounts that overlay while un-zoomed; the existing zoomed pan overlay takes over above `ZOOM_THRESHOLD`. Exactly one tap surface is live at a time.
- The create board drops `HoldTargetLayer` altogether. #4978 moved the board's hold drawing to the real renderer and its discoverability dots to `HoldMarkerLayer`, *under* the rendered holds — so the layer up top was already down to nothing but per-hold hit-testing, and with the overlay in place it had nothing left to do.
- `HoldTarget` / `HoldTargetLayer` lose their gesture path entirely. The layer is markers only now — one dot per hold, `pointerEvents="none"` throughout, no `onPaint` / `onLongPress` / `pinchRef` / `isPinchingSV` props to wire. There's no `interactive` flag either: a board can't accidentally opt back into the broken mode.
- `InteractiveFilterBoard` (the search hold filter) had a hand-rolled copy of the same at-rest arbitration; it now calls the shared hook, so the two can't drift. Its old composition was `Gesture.Exclusive(longPress, tap)`; the shared hook uses `Gesture.Race`.

### Which change is load-bearing

The overlay is. A topmost full-bleed detector beats the per-hold detectors underneath — which is exactly why `InteractiveFilterBoard` was already fine before this PR: it has mounted its own rest overlay above `HoldTargetLayer` since #2687. The create board had no such overlay, so its per-hold squares owned every at-rest touch. Deleting the per-hold gesture path doesn't fix the bug, it removes the way back into it (and hundreds of `GestureDetector`s per board along with it).

### Not swallowing the sheet scroll

The create board lives inside `CreateDrawer`'s scroll view, so a full-bleed gesture at rest could easily eat scrolling.

- Composition is `Gesture.Race(longPress, tap)`, never `Exclusive`. With `Exclusive` the tap waits for the long-press to fail, which only happens on finger-up; `Race` lets the first leg to *activate* win — quick release is a tap, a 400 ms hold is a long-press.
- Both legs fail on movement. The long-press gets an explicit `maxDistance` of 8 px, matching the zoomed board's `PAN_ACTIVATION_OFFSET`, so a drag started inside the 400 ms window cancels the overlay and reaches the scroll view instead of becoming a whole-board long-press. RNGH's default is 10 px; the named constant makes the intent and the parity explicit.
- Both legs stay `simultaneousWithExternalGesture(pinchRef)` and bail on `isPinchingSV`, so pinch-to-zoom still works and a small pinch doesn't paint the hold under a finger.

### What's left of #4496 — why this is `Refs`, not `Closes`

The issue is "hold selection is inaccurate **and** imprecise". This fixes the inaccurate half: the systematic bias to screw-ons and the felt offset are gone, and the 90% mis-hit rate on bolt-on centres goes to 0.

The imprecise half is not fixed. Every hold on a board shares one hit radius (`r = xSpacing * 4` on Aurora boards, one cell radius on MoonBoard), so nearest-centre is a plain Voronoi partition — each hold's cell is half the 13.4 dp spacing, about **6.7 dp (~1.1 mm)** at fit-to-screen. That's well under a fingertip. Misses are now unbiased and deterministic instead of always landing on the same neighbour, and zooming still gives you room, but going for one specific hold in a dense cluster at full-board zoom still takes care.

Closing that half needs a precision affordance rather than better arbitration — auto-zoom on first tap, tap-to-magnify, or confirm-and-nudge. Left open for that.

### About the web half of the report

The reporter also saw this on the web app. The www climbing surface that rendered an interactive board moved to the Expo app in #4435, so that renderer no longer exists in `packages/web`. The Expo browser build (`/app`) runs this same React Native code and picks up the fix here. The one hold-picking surface left in `packages/web` is the MoonBoard import editor (`board-litup-holds.tsx`), which hit-tests real SVG `<circle>` elements at their true radius with no inflation — it doesn't have this bug, so there's no follow-up to file.

**Author verification**

- The genuine regression tests are the three that fail against the old code: `HoldTarget.test.tsx` (the component must render no `GestureDetector` and never touch RNGH), and the assertions in `interactive-create-board-taps.test.tsx` / `interactive-filter-board.test.tsx` that the create board mounts no per-hold layer and the filter board's layer is handed no gesture props at all.
- `use-rest-hold-tap-gesture.test.ts` is unit coverage of the new hook, **not** a regression test for the bug: `resolveHoldAtPoint` already did nearest-centre before this PR, so its assertions would pass against `main` too. The bug lived in which native view received the touch, which no unit test can reach. Its fixture uses the real geometry (equal 22 dp radii, 13.4 dp apart) rather than an invented big-hold/small-hold pair.
- `interactive-create-board-layers.test.tsx` keeps #4978's depth contract honest across this change: the dots stay in the board image's under-overlay slot, and there is now no second dot layer above the holds.
- Rebased onto `main` after #4978 (renderer-drawn create board) and #5107 (pinch/pan) landed; typecheck, lint and the mobile suite are re-run on the rebased tree before this leaves draft.
- Still owed, device QA: real-finger check on a Kilter board that bolt-ons are selectable at rest, and that the create drawer still scrolls under a press-then-drag.

## Release Notes

- Tapping holds in the climb creator now picks the hold you actually touched
- Before, at full-board zoom a handful of screw-ons grabbed every tap and the whole board felt offset — on a Kilter original, 9 out of 10 bolt-on centres were inside a screw-on's hit circle, so bolt-ons were basically unselectable unless you pinched in first
- Dense clusters are still tight at full-board zoom, so pinch in when you're going for one specific hold

## Test plan

1. Signed in, Kilter original board selected.
2. Climbs tab → + → at full-board zoom tap a bolt-on next to a screw-on → the bolt-on paints.
3. Long-press a hold → the role sheet opens.
4. Press then drag on the board → the sheet still scrolls.

## Risk

Risk: 2/5 — gesture routing on the create and filter boards, covered by unit tests.

### DIFF
--- a/packages/mobile/src/components/create-climb/HoldTarget.tsx
+++ b/packages/mobile/src/components/create-climb/HoldTarget.tsx
@@ -1,111 +1,46 @@
-import React, { type MutableRefObject, useMemo } from 'react';
+import React from 'react';
 import { View } from 'react-native';
-import { Gesture, GestureDetector, type GestureType } from 'react-native-gesture-handler';
-import { runOnJS, type SharedValue } from 'react-native-reanimated';
 
 type HoldTargetProps = {
-  holdId: number;
   leftPct: number;
   topPct: number;
-  tapDiameter: number;
   /** Diameter of the faint discoverability dot, in device px. */
   dotDiameter: number;
-  showDot: boolean;
   dotColor: string;
-  onPaint: (holdId: number) => void;
-  onLongPress: (holdId: number) => void;
-  /** The board's ancestor pinch, so this per-hold tap/long-press declares itself
-   *  simultaneous with it. Without that, two fingers on two hold targets each
-   *  claim a pointer and the pinch can't acquire both — zoom stalls on Android. */
-  pinchRef?: MutableRefObject<GestureType | undefined>;
-  /** True while a pinch is in progress. Because the tap/long-press are
-   *  simultaneous with the pinch (above), RNGH no longer fails them when the
-   *  pinch activates, so a small/slow pinch could otherwise paint a hold or open
-   *  the role sheet — gate both callbacks on this. */
-  isPinchingSV?: SharedValue<boolean>;
 };
 
 /**
- * One hold's transparent tap target plus an optional discoverability dot. A
- * `GestureDetector` (not Pressable) is used so the per-hold tap/long-press
- * compose cleanly with the board's ancestor pinch/pan gestures. A quick
- * stationary touch paints; a 400ms press opens the role sheet; a drag falls
- * through to the pan gesture (the Tap fails on movement > maxDistance).
+ * One hold's discoverability dot. Purely visual — `pointerEvents="none"`, no
+ * gesture of its own.
  *
- * `React.memo` leaf with primitive props — the static tap layer never
- * re-renders when holds are painted.
+ * It used to own a transparent tap target inflated to
+ * `max(ringDiameter * 1.6, 44)` px with its own `GestureDetector`. Those squares
+ * overlap heavily at fit-to-screen and overlapping sibling Views are arbitrated
+ * by z-order, so the last hold in the list won every touch inside its square
+ * however far off-centre the finger landed (#4496). Boards now resolve a tap
+ * through one full-bleed overlay — `useRestHoldTapGesture` at rest,
+ * `useZoomedHoldTapGesture` while zoomed — which picks the nearest hold centre.
+ * Don't re-add a per-hold detector here: it would sit above the overlay and
+ * start winning touches by z-order again.
+ *
+ * `React.memo` leaf with primitive props — the marker layer never re-renders
+ * when holds are painted.
  */
-export const HoldTarget = React.memo(function HoldTarget({
-  holdId,
-  leftPct,
-  topPct,
-  tapDiameter,
-  dotDiameter,
-  showDot,
-  dotColor,
-  onPaint,
-  onLongPress,
-  pinchRef,
-  isPinchingSV,
-}: HoldTargetProps) {
-  const gesture = useMemo(() => {
-    const tap = Gesture.Tap()
-      .maxDuration(300)
-      .maxDistance(15)
-      .onStart(() => {
-        'worklet';
-        // Bail if a pinch is active — see isPinchingSV. The tap recognizes on
-        // finger-lift, so without this a finger that's part of a small pinch
-        // would paint the hold it happened to be resting on.
-        if (isPinchingSV?.value) return;
-        runOnJS(onPaint)(holdId);
-      });
-    const longPress = Gesture.LongPress()
-      .minDuration(400)
-      .onStart(() => {
-        'worklet';
-        if (isPinchingSV?.value) return;
-        runOnJS(onLongPress)(holdId);
-      });
-    // Let the ancestor pinch recognize even while this finger sits on the hold —
-    // applied per-leg because the relation method lives on the individual
-    // gestures, not the Exclusive composite.
-    if (pinchRef) {
-      tap.simultaneousWithExternalGesture(pinchRef);
-      longPress.simultaneousWithExternalGesture(pinchRef);
-    }
-    // Long-press wins; tap fires only if the long-press fails (released early).
-    return Gesture.Exclusive(longPress, tap);
-  }, [holdId, onPaint, onLongPress, pinchRef, isPinchingSV]);
-
+export const HoldTarget = React.memo(function HoldTarget({ leftPct, topPct, dotDiameter, dotColor }: HoldTargetProps) {
   return (
-    <GestureDetector gesture={gesture}>
-      <View
-        collapsable={false}
-        style={{
-          position: 'absolute',
-          left: `${leftPct}%`,
-          top: `${topPct}%`,
-          width: tapDiameter,
-          height: tapDiameter,
-          marginLeft: -tapDiameter / 2,
-          marginTop: -tapDiameter / 2,
-          alignItems: 'center',
-          justifyContent: 'center',
-        }}
-      >
-        {showDot ? (
-          <View
-            pointerEvents="none"
-            style={{
-              width: dotDiameter,
-              height: dotDiameter,
-              borderRadius: dotDiameter / 2,
-              backgroundColor: dotColor,
-            }}
-          />
-        ) : null}
-      </View>
-    </GestureDetector>
+    <View
+      pointerEvents="none"
+      style={{
+        position: 'absolute',
+        left: `${leftPct}%`,
+        top: `${topPct}%`,
+        width: dotDiameter,
+        height: dotDiameter,
+        marginLeft: -dotDiameter / 2,
+        marginTop: -dotDiameter / 2,
+        borderRadius: dotDiameter / 2,
+        backgroundColor: dotColor,
+      }}
+    />
   );
 });
--- a/packages/mobile/src/components/create-climb/HoldTargetLayer.tsx
+++ b/packages/mobile/src/components/create-climb/HoldTargetLayer.tsx
@@ -1,7 +1,5 @@
-import React, { type MutableRefObject, useMemo } from 'react';
+import React, { useMemo } from 'react';
 import { StyleSheet, View } from 'react-native';
-import type { GestureType } from 'react-native-gesture-handler';
-import type { SharedValue } from 'react-native-reanimated';
 import type { BoardHoldTarget } from '../../lib/create-board-holds';
 import { holdGeometry, holdMarkerAppearance } from './holdLayout';
 import { HoldTarget } from './HoldTarget';
@@ -14,26 +12,25 @@ type HoldTargetLayerProps = {
   mirrored: boolean;
   /** When true, every hold gets a brighter, larger discoverability dot. */
   showAllHolds: boolean;
-  /** When false, tap targets remain but their discoverability dots are hidden. */
+  /** When false the layer draws nothing: the dots are all it renders. */
   showHoldMarkers?: boolean;
-  onPaint: (holdId: number) => void;
-  onLongPress: (holdId: number) => void;
-  /** The board's ancestor pinch, forwarded to each hold so per-hold gestures
-   *  declare themselves simultaneous with it (reliable pinch-to-zoom). */
-  pinchRef?: MutableRefObject<GestureType | undefined>;
-  /** Forwarded to each hold so its tap/long-press bail while a pinch is active. */
-  isPinchingSV?: SharedValue<boolean>;
 };
 
 /**
- * Static, memoized layer of one tap target per hold. Intentionally independent
- * of the painted state so it never re-renders when a hold is painted.
+ * Static, memoized layer of one discoverability dot per hold, drawn ABOVE the
+ * board — right on a surface that renders no holds of its own (the search filter
+ * board) and wrong on one that does, so the create board draws its dots with
+ * `HoldMarkerLayer` underneath the rendered overlay instead. Independent of the
+ * painted state, so it never re-renders when a hold is painted.
  *
- * The dots it carries are drawn ABOVE the board's holds, which is right on a
- * surface that renders no holds of its own (the search filter board) and wrong
- * on one that does: the create board turns them off here and draws them with
- * `HoldMarkerLayer` underneath the rendered overlay instead, so a dot cannot
- * sit in the middle of a lit hold's fill.
+ * Markers only: `pointerEvents="none"` throughout, no gestures. Taps are
+ * arbitrated by the board's single full-bleed overlay (`useRestHoldTapGesture`
+ * at rest, `useZoomedHoldTapGesture` while zoomed), which resolves a point to
+ * the *nearest* hold centre. Per-hold detectors used to live here and were
+ * arbitrated by view z-order over tap squares inflated to
+ * `max(ringDiameter * 1.6, 44)` px so small holds stayed reachable; at
+ * fit-to-screen those overlap so heavily that the last-rendered hold won every
+ * touch inside its square, however far off-centre the finger landed (#4496).
  */
 export const HoldTargetLayer = React.memo(function HoldTargetLayer({
   holdTargets,
@@ -43,49 +40,29 @@ export const HoldTargetLayer = React.memo(function HoldTargetLayer({
   mirrored,
   showAllHolds,
   showHoldMarkers = true,
-  onPaint,
-  onLongPress,
-  pinchRef,
-  isPinchingSV,
 }: HoldTargetLayerProps) {
   const targets = useMemo(() => {
     if (measuredWidth <= 0) return null;
+    // Nothing to draw — skip the per-hold Views entirely rather than mounting
+    // hundreds of invisible boxes.
+    if (!showHoldMarkers) return null;
     return holdTargets.map((hold) => {
       const geometry = holdGeometry(hold, boardWidth, boardHeight, measuredWidth, mirrored);
       const marker = holdMarkerAppearance(geometry, showAllHolds);
       return (
         <HoldTarget
           key={hold.id}
-          holdId={hold.id}
           leftPct={geometry.leftPct}
           topPct={geometry.topPct}
-          tapDiameter={geometry.tapDiameter}
           dotDiameter={marker.diameter}
-          showDot={showHoldMarkers}
           dotColor={marker.color}
-          onPaint={onPaint}
-          onLongPress={onLongPress}
-          pinchRef={pinchRef}
-          isPinchingSV={isPinchingSV}
         />
       );
     });
-  }, [
-    holdTargets,
-    boardWidth,
-    boardHeight,
-    measuredWidth,
-    mirrored,
-    showAllHolds,
-    showHoldMarkers,
-    onPaint,
-    onLongPress,
-    pinchRef,
-    isPinchingSV,
-  ]);
+  }, [holdTargets, boardWidth, boardHeight, measuredWidth, mirrored, showAllHolds, showHoldMarkers]);
 
   return (
-    <View pointerEvents="box-none" style={StyleSheet.absoluteFill}>
+    <View pointerEvents="none" style={StyleSheet.absoluteFill}>
       {targets}
     </View>
   );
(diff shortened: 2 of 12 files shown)


marcodejongh:
## QA review — verdict: needs-rework (small; the code change itself is good)

The core diagnosis is right and I could corroborate it from the board data. No correctness defect found in the shipped behaviour. What needs fixing before merge is a false user-facing release note, an inaccurate rationale baked into a code comment + test fixture, and ~40 lines of now-unreachable code the repo rules say to remove.

### What I verified

- `gh pr checks 4571` — all green (check, lint, typecheck, mobile-bundle, test-default 1/2 + 2/2, ota-check, release-notes). Nothing pending, nothing red.
- `vp test run --project mobile <the 6 touched files>` — 6 files, 43 tests, pass. Matches the test plan exactly.
- Full `vp run test:mobile` — exit 0.
- Scope: one commit, nine files, all under `packages/mobile/src/components/{create-climb,search}`. No i18n strings, no theme values, no DB, no BLE. Nothing that needs a Fable review.
- Web half of the report: confirmed. The only `onHoldClick` consumer left in `packages/web` is `moonboard-edit-modal.tsx:128` → `board-litup-holds.tsx:77`, which hit-tests SVG `<circle>` at the true `r` with no inflation. No follow-up owed there.
- Zone mode (`app/(tabs)/climbs/zone.tsx:304`) passes no `onHoldTap`, so `useRestHoldTapGesture` returns null and no overlay mounts. Correct.

### The diagnosis checks out, with numbers

Kilter layout 1 is sets `[1 Bolt Ons, 20 Screw Ons]` in that order (`SETS['1-10']`), and `board-details.ts:103` concatenates set placements in `setIds` order — so screw-ons really are last in the list, i.e. topmost. At a 340 dp board (size 10, 1080 px source):

- every hold's hit radius is 22 dp (`MIN_TAP_DIAMETER` wins; `ringDiameter` is only 18.9 dp)
- nearest-neighbour distance is 13.46 dp for **every** hold — the two sets interleave on a (4,4) offset half-grid
- median 5 (max 6) other hold centres fall inside any given hit circle
- **391 of 488 bolt-on centres (80%) sit inside some screw-on's hit circle**

So before this PR, tapping the dead centre of a bolt-on handed the touch to a screw-on four times out of five. That is exactly the report. Nearest-centre fixes it: distance 0 always wins. Good fix.

### Defects

**1. Release note "Same fix for the hold filter in search" is false.** `InteractiveFilterBoard` already mounted the full-bleed rest overlay *above* `HoldTargetLayer` before this PR (`main:packages/mobile/src/components/search/InteractiveFilterBoard.tsx:252-256`), using the same `resolveHoldAtPoint`. Search was already arbitrating by nearest centre; this PR is a refactor there (plus Exclusive→Race and a new 8 px long-press budget). Please drop or reword that bullet — it tells users a bug was fixed that they never had.

Worth reconciling the two readings, because they can't both hold: if a topmost full-bleed overlay reliably beats the per-hold detectors underneath, then search was already fine (release note wrong). If it doesn't, then adding the overlay to the create board wouldn't fix anything either, and `interactive={false}` is the only load-bearing change. I believe it's the first, and `interactive={false}` is belt-and-braces — but the PR body asserts the second implicitly. One sentence in the description would settle it.

**2. The "residual precision limit" rationale is wrong for Aurora boards** — `use-rest-hold-tap-gesture.ts:56-61`, repeated in the PR body and in the test fixture at `__tests__/use-rest-hold-tap-gesture.test.ts:85-88` (radii 22 and 48). For Kilter/Tension, `board-details.ts:110` sets `r: xSpacing * 4` — a **constant for every hold**. There is no "large bolt-on" and "small neighbour"; all hit circles are identical, so nearest-centre is exactly a Voronoi partition and a touch near a hold's rim can never be stolen by a "small" neighbour for the reason given. (Varying `r` only happens on MoonBoard.)

The real residual is different and worth stating accurately: with 13.46 dp between neighbours, each hold's cell is ~6.7 dp (~1.1 mm) wide. At fit-to-screen the user still has to land within about a millimetre of the intended centre. Which leads to:

**3. `Closes #4496` may be premature.** The issue is titled "hold selection is inaccurate and imprecise". The *inaccurate* half (systematic bias to screw-ons, the felt offset) is genuinely fixed. The *imprecise* half is not — ~1 mm cells are still well under a fingertip, so misfires at fit-to-screen remain common, just unbiased now. The release note "now picks the hold you actually touched" overstates that. Marco's call whether to ship this and leave #4496 open for a real precision affordance (tap-to-magnify, confirm-and-nudge, auto-zoom on first tap), but auto-closing on merge will bury it.

**4. Dead code left behind (CLAUDE.md: "Remove dead code as you go").** Both production call sites now pass `interactive={false}` (`InteractiveCreateBoard.tsx:173`, `InteractiveFilterBoard.tsx:233`). Nothing in the app mounts `HoldTargetLayer` with `interactive` true, so `HoldTarget.tsx:57-86` (the whole `Gesture.Tap`/`LongPress`/`Exclusive` build) and the `onPaint`/`onLongPress`/`pinchRef`/`isPinchingSV` props threaded down through `HoldTargetLayer.tsx:84-87` are unreachable. Suggest deleting the gesture path and letting `HoldTarget` be what its own new doc-comment calls it — a marker.

**5. `interactive` defaults to `true` (`HoldTargetLayer.tsx:64`, `HoldTarget.tsx:55`), which is the broken mode.** If #4 isn't taken, flip the default to `false` or make it required: a future board that mounts the layer and forgets the prop silently reintroduces #4496, and nothing would catch it.

**6. One test now covers a path production can't reach.** `search/__tests__/interactive-filter-board.test.tsx:178-184` ("routes a hold tap to onHoldTap with the hold id") clicks a button rendered by the `HoldTargetLayer` *mock* and wired to `onPaint`. In production those per-hold detectors are inert. It passes either way and asserts nothing real now.

### On the tests generally

`interactive-create-board.test.tsx:136-142` and the new `HoldTarget.test.tsx` case are genuine regression tests — both fail against the old code (no `interactive` prop, `data-gesture` present). Good.

`use-rest-hold-tap-gesture.test.ts` is honest unit coverage of the new hook, but it does not pin the bug: `resolveHoldAtPoint` already existed and already did nearest-centre before this PR, so every assertion in that file would have passed against `main`'s copy of the same function. The bug lived in *which native view got the touch*, and no unit test can reach that. Worth saying so in the test plan rather than "pins the behaviour that was broken".

### Still owed

Device QA, as the author notes: a real finger on a Kilter original confirming bolt-ons are selectable at rest, and that press-then-drag still scrolls `CreateDrawer`'s `BottomSheetScrollView` (the overlay's long-press now activates board-wide after 400 ms, including over the margins the per-hold squares didn't cover).

— automated QA pass, no changes made to the branch

marcodejongh:
## QA follow-up — all four findings actioned

Pushed `ca74116` (rebased on `main`). Nothing rebutted; every finding checked out, and two of them checked out harder than stated.

**1. False release note — fixed.** Confirmed: `main:packages/mobile/src/components/search/InteractiveFilterBoard.tsx:253-257` already mounted the rest overlay above `HoldTargetLayer`, so search was already arbitrating by nearest centre. Bullet dropped.

On the reconciliation you asked for: it's the first reading, and the PR body now says so under a new "Which change is load-bearing" heading. The topmost full-bleed overlay beats the per-hold detectors underneath — that's why search was already fine, and why the create board's *new* overlay is the fix on its own. Removing the per-hold gesture path doesn't fix the bug, it removes the way back into it.

**2. Precision rationale — wrong, and wrong in the same way on MoonBoard too.** `board-details.ts:72` sets `r: xSpacing * 4`, a constant per board; `moonboard-config.ts:338` sets `holdRadius = min(cellWidth, cellHeight) * 0.525`, also a constant. So there is no varying `r` on *any* board — the big-hold/small-neighbour story was fiction everywhere, not just on Aurora. Corrected in the hook doc, the PR body, and the fixture (now two equal 22 dp circles 13.4 dp apart, which is the real Kilter geometry).

I re-derived the numbers from the shared board data rather than reusing yours, and got a slightly different split for Kilter layout 1 / size 10 at a 340 dp board: 476 holds (323 bolt-on, 153 screw-on), hit circle 44 dp, nearest-neighbour spacing 13.36 dp, median 5 foreign centres inside a hit circle, and **289/323 = 90%** of bolt-on centres inside some screw-on's circle. Same conclusion, slightly worse than 80%. Voronoi cell ~6.7 dp (~1.1 mm) — stated as the real residual.

**3. `Closes` → `Refs`.** Body now has a "What's left of #4496" section: the inaccurate half is fixed with the measurement, the imprecise half needs a precision affordance (auto-zoom on first tap, tap-to-magnify, confirm-and-nudge) and isn't attempted here. Issue stays open for that; your call whether to split it.

**4. Dead code — deleted, which folds in #5 and #6.** `HoldTarget` is now one dot with `pointerEvents="none"`; the `Gesture.Tap`/`LongPress`/`Exclusive` build, the inflated square wrapper, and the `holdId`/`tapDiameter`/`showDot` props are gone. `HoldTargetLayer` no longer takes `onPaint`, `onLongPress`, `pinchRef`, `isPinchingSV` — or `interactive`, so there's no default to get wrong (#5 resolved by construction rather than by flipping it). Both call sites shed the props; that also clears the dead `onLongPress={onHoldTap}` from the earlier bot review.

**#6 — the mock-only test is gone.** Deleted `'routes a hold tap to onHoldTap with the hold id'`; the layer mock renders inert divs now, no clickable buttons. Both board tests instead assert the layer receives *no* gesture props, which is what fails if a detector is ever wired back in.

**On the test plan wording** — taken. The test plan now says plainly that `use-rest-hold-tap-gesture.test.ts` is unit coverage of a new hook and would have passed against `main`, and names the three assertions that are genuine regression tests (they fail against the pre-PR component/props).

`HoldTarget.test.tsx` also now mocks RNGH with throwing `Gesture.Tap`/`LongPress`, so re-adding a per-hold detector fails the suite rather than silently competing with the overlay again.

Validation: `vp run typecheck:mobile` green, `vp run test:mobile` 686 files / 6561 tests green, `vp check` clean in the touched files (the 8 repo-wide errors are pre-existing `TS2307`s in `@gorhom/bottom-sheet` / `@boardsesh/text-redaction` / `yaml`, unrelated).

Still owed and unchanged: device QA on a real Kilter — bolt-ons selectable at rest, and press-then-drag still scrolls `CreateDrawer`'s `BottomSheetScrollView`.

marcodejongh:
## Second pass on the QA findings — verified, one rebuttal

Re-checked all four findings against the tree at `ca74116` rather than trusting the earlier follow-up. All four are genuinely closed; no further code changes were needed, so nothing new is pushed. Details, including the one number I disagree with:

**1. False release note — closed, and the premise is right.** `git show origin/main:.../search/InteractiveFilterBoard.tsx` puts the full-bleed rest overlay at lines 253-257, *after* (above) the `HoldTargetLayer` block at 238-252. Search really was already arbitrating by nearest centre before this PR. The bullet is gone from the release notes.

**2. Precision rationale — closed, and I re-derived the numbers independently.** Confirmed `r` is a per-board constant on every board: `packages/mobile/src/lib/board-details.ts:110` → `r: xSpacing * 4` (Aurora), `moonboard-config.ts:338` → `holdRadius = min(cellWidth, cellHeight) * 0.525` (MoonBoard). One distinct `r` per board, so the big-hold/small-neighbour story was wrong everywhere. Corrected in the hook doc, the body, and the fixture.

Running the real `getBoardRenderData` + `holdGeometry` for kilter / layout 1 / size 10 / sets `[1,20]` at a 340 dp board:

```
boardWidth 1080  boardHeight 1170
holds: 476 total (323 bolt-on, 153 screw-on)
distinct r values: 1        ring 18.89 dp        hit radius 22.00 dp (44 dp circle)
nearest-neighbour  min/median/max: 13.36 / 13.36 / 18.89 dp
foreign centres inside one hit circle: median 5, max 6
bolt-on centres inside some screw-on circle: 289 / 323 = 89.5%
```

That reproduces the PR body exactly. It does not reproduce the review's **391/488 = 80%** — the board has 476 holds at this size, not 488, and 289 of the 323 bolt-ons are captured, not 391. Same conclusion either way, but the figure in the body (90%) is the one that comes out of the shared board data.

**3. `Closes` → `Refs` — closed.** Body opens with `Refs #4496` and carries a "What's left of #4496" section: inaccurate half fixed with the measurement, imprecise half (~6.7 dp / ~1.1 mm Voronoi cells at fit-to-screen) explicitly not attempted. The issue will stay open on merge.

**4. Dead code — closed, and it took #5 and #6 with it.** `HoldTarget` is 43 lines: one `pointerEvents="none"` dot, no RNGH import at all. `HoldTargetLayer` no longer accepts `onPaint` / `onLongPress` / `pinchRef` / `isPinchingSV` / `interactive`, so there is no default to get wrong. The mock-only filter-board test is deleted.

### Rebutted

**`handleTap` / `handleLongPress` missing from the `useMemo` deps** (latest bot review) — not acted on, deliberately.

- There is no rule to violate: `.oxlintrc.json` enables `["typescript", "react", "nextjs", "import"]`. `react-hooks` is not in the plugin list, so `react-hooks/exhaustive-deps` does not run in this repo and nothing is being silenced.
- The pattern is byte-for-byte the one in `use-zoomed-hold-tap-gesture.ts:93-106`, which this hook's doc comment is explicitly told to keep in sync with. Wrapping only the new copy in `useCallback` would make the two diverge for no behavioural gain.
- It is correct as written: both handlers close over `callbacksRef` (a stable ref) and read `.current` at call time, so the gesture built once per mount still sees fresh `hitTargets` and callbacks.

### Also checked, no defect

The rest overlay maps gesture-local x/y straight to board-local px, which assumes scale 1 and no translation while `!isZoomed`. That holds: `use-zoom-pan-gesture.ts:206-214` snaps scale back to `MIN_SCALE` and translation to 0 whenever the pinch ends under `ZOOM_THRESHOLD` (1.02), and mid-pinch both legs bail on `isPinchingSV`. So there is no window where a callback fires with a scale between 1 and 1.02.

### Validation

- `vp test run --project mobile src/components/create-climb src/components/search` — 24 files, 174 tests, all passing.
- `gh pr checks 4571` — 21 pass, 13 skipping, 0 failing (`check`, `lint`, `i18n`, `mobile-bundle`, `commit-lint`, `release-notes`, `ota-check`, `test-default` 1/2 and 2/2 all green).
- Branch is already on the tip of `main` (`git merge-base --is-ancestor origin/main HEAD` passes), so no rebase was needed.

Still owed and unchanged: device QA on a real Kilter — bolt-ons selectable at rest, and press-then-drag still scrolling `CreateDrawer`'s `BottomSheetScrollView`.

marcodejongh:

Conflicting: packages/mobile/src/components/search/InteractiveFilterBoard.tsx.

marcodejongh:
🤖 Session `4634-f0` (Fable) is taking this PR over as part of the route / create-screen campaign for #4634: rebase onto current `main` (stacked on #5107 where the create-screen files collide), CI back to green, review threads answered, test plan + risk in tester voice, then ready for review. Peer loops: please stand off this branch; say so here if you are already mid-flight.

marcodejongh:
Rebased onto `main` (`98858fdff`, which now carries #5107's squash) and force-pushed as `84d9780a3` — two commits, same shape as before.

**Rebase.** Conflicts were in `InteractiveCreateBoard.tsx`, `HoldTargetLayer.tsx` and `InteractiveFilterBoard.tsx` against #4978 (renderer-drawn holds) and #5107 (`scrollRef` / `onInteractionActiveChange`). Everything from main is kept; this PR's nearest-centre tap resolution and the markers-only hold layer sit on top. One shape change: the create board no longer mounts `HoldTargetLayer` at all — post-#4978 that instance drew nothing (`showHoldMarkers={false}` + markers-only is an empty View), so it was dead code. The PR's board test was renamed `interactive-create-board-taps.test.tsx` because main already has `InteractiveCreateBoard.test.tsx` and the two names collide on case-insensitive filesystems.

**Premise re-checked on the rebased tree.** #4978 moved hold drawing to the renderer but left the transparent per-hold `GestureDetector` squares as the only at-rest hold gesture, so taps were still won by sibling z-order; the 9-of-10 bolt-on measurement stands. The three regression tests fail against main and pass here.

**Review note.** `handleTap` / `handleLongPress` are now stable `useCallback`s reading render-scoped values through a ref; the worklets are unchanged (one gated `runOnJS`).

**Validation, run one command at a time on the tip:** `vp run typecheck:mobile` clean · create-climb + search suites 36 files / 327 tests · `vp run test:mobile` 808 files / 8359 tests · `vp check` 0 errors · `vp run check:mobile-bundle` OK. Test plan and risk rewritten for the tester gate.

Follow-up worth its own issue, not done here: with `HoldTargetLayer` markers-only it is near-identical to `HoldMarkerLayer`; one of them can go.

PR #4573, opened 2026-08-19
fix(mobile): offer add-to-playlist targets the server will accept

marcodejongh:
## Summary

Follow-up to #4304 (`cadb8fb95`, merged today). Refs #4015.

That PR added a board-compatibility guard to `addClimbToPlaylist`. Re-reviewing it
against current `main` turned up three things. Two are real and fixed here; the
third does not hold up and is reported instead of patched.

## 1. The guard and the picker judged different keys (fixed)

The guard asks "does this **climb** fit this playlist", resolved from the climb's
own `board_climbs` row. The mobile picker asks "does this playlist fit the board
the **host surface** is rendering on", from its `boardName`/`layoutId` props.
Those coincide on every single-board surface, so it looked fine — but they are
not the same key, and where they diverge the picker offers a target the guard
then rejects outright.

The reachable path is the play drawer's cross-board preview:

- `PlayDrawer.handleOpenActions` (`PlayDrawer.tsx:706`) opens the reaction menu
  with `onOpenClimbActions(displayedClimb, undefined, …)` — **no board override**.
- `openClimbActions` with no override falls back to
  `storedActiveBoardConfigRef.current`, the stored *active* board
  (`drawer-host-provider.tsx:513`).
- `displayedClimb` is `drawerPreviewItem ?? currentClimbQueueItem`, and
  `drawerPreviewItem` is documented as the genuine cross-board case
  ("workout builder, logbook/cross-board, the peer-driven accessory wall climb",
  `PlayDrawer.tsx:385`). The route even renders a `boardMismatch` banner for it.

So: active board Kilter/8, a Tension climb previewed from the logbook, tap
"Add to playlist" → the picker lists **Kilter** playlists → tapping one throws
`This playlist is for a different board`. Before #4304 that add landed
(the membership refetch is keyed the same way as the picker, so the checkmark
even stuck) — the row was wrong, but the climber never saw a hard failure.

**Resolution: keep the guard, fix the key the client filters on.** The guard's
question is the only one actually about "does this climb belong in this
playlist" — re-keying it to a client-supplied board config would let any client
file a Tension climb into a Kilter playlist just by declaring `boardType:
kilter`. That is not one definition, it is none.

So the rule now lives in exactly one place, `climbFitsPlaylistBoard` in
`@boardsesh/board-config`, imported by both the resolver and the picker's
`filterPlaylistsByBoard`. Alongside it, `resolveClimbBoardScope` gives the client
the climb's own scope: prefer `Climb.boardType`/`Climb.layoutId` when the payload
carries them (logbook, profile climbs, playlist detail, the cross-board preview),
fall back to the host's board config for payloads that don't — which is every
single-board surface, where the two are the same value anyway.

### Which fallback, and why

Layout ids are namespaced per board (`board_layouts` PK is `(board_type, id)`),
so **the board and the layout resolve as one pair — never one field from each
side.** A first cut of this PR fell back per field, which for a climb carrying
`boardType: 'tension'` and no layout produced `{ tension, <the active Kilter
layout> }`: a board nobody has. That is worse than the bug it replaced — it hides
the climber's own Tension playlists, empties the checkmarks, and lets inline
create write a Tension playlist pinned to a Kilter layout, a row the server's
guard can never accept. It is reachable, too: `toClimbQueueItem` produces exactly
that shape for a peer's queued climb, and the queue sheet opens the reaction menu
with no board override.

The rule instead:

- the climb names no board, or names the host's — then it *is* on the host board
  (`Climb.boardType` is documented as "populated in multi-board contexts", so a
  payload that omits it came off a surface whose own board is the climb's), and
  the host's layout id is a legitimate stand-in for a missing one;
- the climb names a **different** board — then the host's layout belongs to
  another namespace, so the layout stays `null`: unknown, not guessed.

`climbFitsPlaylistBoard` already treated a null climb layout as "don't reject on
a guess" and leaves the last word to the server, which always knows the real
layout. Until now that branch was dead on the client; it is the whole point of it.

Callers handle the unknown rather than substituting a layout. The picker filters
on the board alone (every playlist on the climb's board stays offered), skips the
`playlistsForClimb` query — it takes `layoutId: Int!` and the only value on hand
is the wrong board's, so it would answer "in none of your playlists" — and hides
inline create, which would otherwise have to invent a layout to pin the new
playlist to. When the layout *is* known, create uses the climb's board and layout
rather than the host's, so a playlist made for a cross-board climb is no longer
created on the host board and then immediately rejected by the guard.

Backend behaviour is unchanged: `board_climbs.layout_id` is `NOT NULL`, so the
shared rule's "unknown climb layout" branch is unreachable server-side.

## 2. `board_climb_aliases` full scan on every unknown uuid (fixed)

`resolveClimbBoardScopes` queries `WHERE alias_uuid = $1` with no `board_type`.
The PK is `(board_type, alias_uuid)` — `alias_uuid` trailing — and the only other
index is `(board_type, canonical_uuid)`, so nothing serves that predicate.

Measured on the dev catalog snapshot (345,270 alias rows, 44 MB heap):

```
Parallel Seq Scan on board_climb_aliases   43.8 ms   5663 buffers   2 workers
Index Scan (after adding the index)         3.3 ms      3 buffers
```

The guard runs that query for any uuid missing from `board_climbs`, so an
authenticated caller posting unknown uuids drove one full scan per call. Migration
`0202_round_baron_zemo.sql` adds `board_climb_aliases_alias_uuid_idx`
(generated with `drizzle-kit generate`; ~19 MB).

## 3. No rate limit on `addClimbToPlaylist` — reported, not patched

Confirmed absent, but **every** mutation in
`packages/backend/src/graphql/resolvers/playlists/mutations.ts` is unlimited:
`createPlaylist`, `updatePlaylist`, `deletePlaylist`, `addClimbToPlaylist`,
`removeClimbFromPlaylist`, `reorderPlaylistClimb`, `updatePlaylistLastAccessed`,
`followPlaylist`, `unfollowPlaylist`, `pinPlaylist`, `unpinPlaylist`. The only
`applyRateLimit` anywhere in the playlists resolver is on the `smartPlaylist`
query. Adding a budget to this one mutation and no sibling would be arbitrary and
would invite the same bug class next to it, so this needs its own pass over the
whole file rather than a one-line addition here. With the index in place the
amplification that made it urgent is gone.

## Release Notes

- Adding a climb to a playlist from the play drawer now offers the playlists that
  climb can actually go in, instead of the ones for whatever board you last had
  set — no more "this playlist is for a different board" on a playlist the app
  just showed you.

## Test plan

New:

- `packages/shared/board-config/src/__tests__/playlist-board-scope.test.ts` — 14
  cases over `climbFitsPlaylistBoard` (board mismatch, layout mismatch, board-wide
  `layoutId: null` playlist, unknown climb layout) and `resolveClimbBoardScope`,
  including the pairing property: whatever comes back, the layout id either came
  from the same side as the board or is null.
- `InlinePlaylistPicker.test.tsx` — seven cases: a climb whose board disagrees with
  the host offers its own board's playlists; same for a layout-only disagreement;
  inline create uses the climb's board; a climb whose board disagrees and whose
  layout is unknown still gets its own board's playlists (not the host layout's);
  inline create is hidden in that state; a climb naming the host's board with no
  layout still creates on the host layout; and a `boardType` this build doesn't
  recognise keeps the host board *without* the climb's layout id.
- `sort-filter-playlists.test.ts` — `filterPlaylistsByBoard` with a null climb
  layout keeps every layout of the board and still drops other boards.

Each new case was checked against a reverted fix rather than assumed:

- reverting `resolveClimbBoardScope` to the per-field fallback → `2 failed | 113
  passed` in the shared tests (including the pairing property) and `2 failed | 26
  passed` in the picker tests.
- reverting the picker's `toBoardName` guard to `toBoardName(...) ?? boardName`
  (host board + the climb's layout) → `1 failed | 28 passed` in the picker tests.
- the three original cases were verified the same way against the pre-#4304
  picker.

Run locally:

- `vp test run --project mobile InlinePlaylistPicker` — 29 passed
- `vp test run --project mobile sort-filter-playlists` — 14 passed
- `vp test run packages/shared/board-config` — 115 passed, 10 files
- `vp check` — 0 errors; the 324 warnings reproduce on clean `main` (#4488)

Backend Vitest is left to CI: this machine runs several worktrees against the same
Postgres worker DBs on :5433, so local backend runs report noise rather than
signal. The existing `add-climb-to-playlist-board-guard.test.ts` covers the
resolver and must stay green — the guard's observable behaviour is unchanged, only
the source of the predicate moved.

## Not changed

- `PlayDrawer.handleOpenActions` still opens the reaction menu with no board
  override, so the *other* actions off a cross-board preview (favorite, tick,
  queue) keep operating against the stored active board. Same for the `angle` the
  add records in `playlist_climbs`: it comes from the host's board config, so a
  cross-board add stores the host's angle. Passing the drawer's own board config
  through `handleOpenActions` would fix board, layout and angle at once for all
  four actions — worth doing, but it changes behaviour on four surfaces rather
  than one and wants your call before it rides along here.
- Splitting migration `0202` into its own PR. It puts a mobile-only change in the
  renumber queue for a perf fix on a fail-open path; happy to lift it out if you'd
  rather the `db-migration` label went with it.
- Cross-board `playlist_climbs` rows already in the database. Cleaning those is
  data deletion and needs sign-off.


marcodejongh:
## QA review — needs rework

The diagnosis is right and I verified it independently: `openClimbActions` really does
fall back to `storedActiveBoardConfigRef.current`
(`packages/mobile/src/providers/drawer-host-provider.tsx:517`), `PlayDrawer.handleOpenActions`
really does pass no override (`packages/mobile/src/components/play-drawer/PlayDrawer.tsx:713`),
`board_climbs.layout_id` really is `NOT NULL` (`packages/db/src/schema/boards/unified.ts:313`)
so the backend branch is unchanged, and the three new picker tests really do fail against the
pre-fix code (I re-ran them: 25 passed, and the assertions invert the old host-keyed behaviour).
Web is correctly out of scope — `packages/web/app/hooks/use-climb-actions-data.tsx` from the
issue body no longer exists.

But the client-side scope resolution has a hole that reintroduces the exact failure it is
meant to remove, and it is asserted as intended behaviour in a unit test.

### 1. `resolveClimbBoardScope` mixes a climb's board with the host's layout id (blocking)

`packages/shared/board-config/src/playlist-board-scope.ts:64-67`

```ts
boardType: climb.boardType ?? fallback.boardType,
layoutId: climb.layoutId ?? fallback.layoutId,
```

The two fields fall back independently, but layout ids are namespaced per board —
`board_layouts` PK is `(board_type, id)` (`packages/db/src/schema/boards/unified.ts:125`).
A climb carrying `boardType: 'tension'` with no `layoutId` therefore resolves to
`{ tension, <a Kilter layout id> }`: a combination that does not exist.

This is reachable, not theoretical. `toClimbQueueItem` produces exactly that shape —
`boardType: … ?? undefined`, `layoutId: subscriptionItem.climb.layoutId` (nullable on the
wire) — at `packages/mobile/src/lib/queue-conversion.ts:100-101`, and the queue sheet opens
the reaction menu with **no** board override at
`packages/mobile/src/components/play-drawer/use-queue-sheet-handlers.ts:75`. So: party
session, active board Kilter/1, a peer queues a Tension climb whose payload omits `layoutId`
→ long-press the row → picker scope becomes `{ tension, 1 }`. Consequences:

- `filterPlaylistsByBoard` hides every Tension playlist pinned to the climb's real layout —
  the climber's own playlists disappear from the picker.
- the membership query fires as `playlistsForClimb(tension, 1)`
  (`InlinePlaylistPicker.tsx:167-178`), so existing checkmarks vanish — the #4015 symptom,
  re-created by the fix.
- worst: inline create at `InlinePlaylistPicker.tsx:394-397` writes a **Tension playlist
  pinned to Kilter layout 1**. That row is permanently un-addable (the server guard compares
  the climb's real layout against it), and the add that follows the create fails — leaving the
  orphaned empty playlist the comment on those very lines says this change prevents.

The shared rule already has the right answer and the client never reaches it:
`climbFitsPlaylistBoard` treats `climb.layoutId === null` as "don't reject on a guess"
(`playlist-board-scope.ts:47`), but `resolveClimbBoardScope` returns `layoutId: number`, so
that branch is dead on the client. Suggested shape: when the climb supplies a `boardType` that
differs from the host's, do not inherit the host's `layoutId` — carry `null` and let the
board-wide branch do its job, and have inline-create make a board-wide playlist (or refuse)
rather than inventing a layout id.

### 2. The unit test locks the bug in

`packages/shared/board-config/src/__tests__/playlist-board-scope.test.ts:64-69`

```ts
expect(resolveClimbBoardScope({ boardType: 'tension' }, { boardType: 'kilter', layoutId: 8 }))
  .toEqual({ boardType: 'tension', layoutId: 8 });
```

That is the defect above, asserted as the contract. The two `climbFitsPlaylistBoard`
null-layout cases (lines 40-52) likewise cover a state no caller can produce. And no picker
test covers the mixed case at all — all three new `InlinePlaylistPicker` cases set **both**
`boardType` and `layoutId` on the override climb, and the 22 pre-existing ones set neither
(`const climb = { uuid: 'climb-1', … }`, test file line 146). The one shape that breaks is the
one shape untested.

### 3. Same class of bug in the `toBoardName` guard

`packages/mobile/src/components/playlist/InlinePlaylistPicker.tsx:158`

```ts
return { boardType: toBoardName(resolved.boardType) ?? boardName, layoutId: resolved.layoutId };
```

When the climb's `boardType` is a string this build doesn't recognise, the board falls back to
the host but the **climb's** `layoutId` is kept — another cross-namespace pairing. The comment
directly above says "keep the host board when this build doesn't recognise the value"; it
keeps half of it. Fall back to the whole host scope, or to `{ host board, null layout }`.

### 4. `angle` is still the host's on the path this PR targets (non-blocking, but it makes the fix half a fix)

`InlinePlaylistPicker.tsx:280` — `addToPlaylist(playlist.uuid, climb.uuid, angle)`, where
`angle` comes from `boardConfig.angle` (`ClimbReactionMenu.tsx:530`), i.e. the stored active
board's angle. `playlist_climbs.angle` is persisted
(`packages/db/src/schema/app/playlists.ts:81`). So on the cross-board preview in the PR
description, the add now succeeds and records the Kilter angle against a Tension climb. The
board/layout key was re-derived from the climb; the angle beside it was not.

This is worth Marco's call: the "Not changed" section rules out fixing
`PlayDrawer.handleOpenActions` to pass the drawer's own board config, but that one-line
override is what makes board, layout **and** angle right at once, and it also fixes favorite /
tick / queue on the same preview — which this PR leaves operating against the wrong board.
Re-keying only the playlist picker means four sibling actions on one menu now disagree about
which board the climb is on.

### 5. Scope: the migration

`packages/db/drizzle/0202_round_baron_zemo.sql` adds `board_climb_aliases_alias_uuid_idx`.
The index is justified (the predicate genuinely has no serving index — PK is
`(board_type, alias_uuid)`, `alias_uuid` trailing) and it was generated with `drizzle-kit`,
journal included. But it is a separate concern from the picker/guard key mismatch, and it puts
this PR in the renumber queue with the rest of the batch for a perf nicety on a fail-open path.
Splitting it out would take the `db-migration` label off a mobile-only change. Marco's call.

### What I checked and found clean

- `@boardsesh/board-config` is already a dependency of both `packages/backend` and
  `packages/mobile` — no new dep edge, no circularity.
- Backend behaviour is genuinely unchanged: the local `ClimbBoardScope`
  (`mutations.ts:24`, `layoutId: number`) still feeds the shared predicate, and the new
  null branch is unreachable server-side.
- `['playlistsForClimb', …]` is only constructed inside the picker — the key re-scoping can't
  desync another component's `setQueryData`.
- `climbBoardScope` is memoized on primitives, so `membershipKey` and `sortedPlaylists`
  identities stay stable; no render churn, context values untouched.
- No `any`, no new user-facing strings (so no i18n obligation), no theme/hardcoding issues,
  nothing Bluetooth-adjacent.
- Migration 0202 is free on `origin/main` as of now.

### CI

Green except `claude-review`, which failed with an empty `ANTHROPIC_API_KEY` (run
32257032840) — infrastructure, not this diff. `test-default`, `test-backend`,
`db-migrations`, `mobile-bundle`, `lint`, `check`, `ota-check` all pass. Locally I ran
`vp test run --project mobile InlinePlaylistPicker` (25 passed) and
`vp test run packages/shared/board-config` (113 passed across 10 files — the PR body says 126,
which doesn't reproduce with that path; harmless, but the number in the test plan is off).
Branch is `MERGEABLE`; `BLOCKED` is the missing review.

marcodejongh:
## QA rework — pushed (`c9edf78`, rebased on `main`)

You were right on all three of the blocking points, and I could reproduce each
one. Fixed, plus the test that pinned the defect.

### The mongrel scope (findings 1 + 3)

Board and layout now resolve **as one pair**, never one field from each side:

- climb names no board, or names the host's → it *is* on the host board
  (`Climb.boardType` is "populated in multi-board contexts", so a payload that
  omits it came off a surface whose board is the climb's), so the host's layout
  id is a real stand-in for a missing one;
- climb names a **different** board → the host's layout is in another namespace,
  so the layout stays `null`: unknown, not guessed.

That is your suggested shape, and it makes `climbFitsPlaylistBoard`'s
`climb.layoutId === null` branch live on the client instead of dead — the answer
was already in the shared rule, as you said.

Your finding 3 is the same rule one level up, so it gets the same treatment: an
unrecognised `boardType` can't be the host's, so its resolved layout is in a
namespace we can't name — the picker now falls back to `{ host board, null
layout }` rather than keeping half of it.

Callers handle the unknown instead of substituting a layout:

- filter → the board alone, so the climber's own playlists on that board stay
  listed (previously: hidden behind a Kilter layout id);
- membership query → skipped. `playlistsForClimb` takes `layoutId: Int!` and the
  only value on hand is the wrong board's, which can only answer "in none of your
  playlists". Seeded memberships still fill the checkmarks in;
- inline create → the "+" is hidden, with a belt-and-braces guard in
  `handleSubmitCreate`. `createPlaylist` also takes `layoutId: Int!`, so there is
  no board-wide playlist to create through this API — and inventing the host's is
  exactly the un-addable row you flagged. Better no create than a create that
  can't be undone from the UI.

When the layout *is* known nothing changes: create still uses the climb's board
and layout.

### The test that locked the bug in (finding 2)

`resolveClimbBoardScope({ boardType: 'tension' }, { kilter, 8 })` now expects
`{ tension, null }`, and there's a property case asserting the invariant directly:
any layout that comes back either came from the same side as the board, or is
null. Reverting the resolver to the per-field fallback fails `2 | 12` there and
`2 | 26` in the picker.

Also added the three picker cases you noted were missing — the mixed shape from
`toClimbQueueItem` (board, no layout) for both the filter and the hidden create,
plus the "climb names the host's board with no layout" case that proves
single-board surfaces still keep the host layout. And a `filterPlaylistsByBoard`
case for a null climb layout.

Local: `vp run typecheck:mobile` clean, `vp check` 0 errors,
`InlinePlaylistPicker` 28 passed, `sort-filter-playlists` 14 passed,
`packages/shared/board-config` 115 passed. (Your 113 vs the body's 126: 115 now,
and the body's number was wrong — corrected.)

### Findings 4 and 5 — not patched, over to you

Both are your call rather than mine, and both are now written into the PR body's
"Not changed" so they don't get lost:

- **angle.** Confirmed: `playlist_climbs.angle` records `boardConfig.angle`, the
  stored active board's. I didn't swap in `climb.angle` unilaterally — it changes
  what angle every add records on every surface, and the fix you actually want is
  the one-line `handleOpenActions` override, which also straightens out favorite /
  tick / queue. Say the word and I'll do it that way instead of re-keying the
  picker alone.
- **migration 0202.** Justified but off-topic, and it drags a mobile-only change
  into the renumber queue. Happy to lift it into its own PR — one word and it's
  out.

marcodejongh:
## QA follow-up — re-verified, plus the one branch the rework left untested (`695b69b`)

Picking the QA thread back up. The three blocking findings were already reworked
in `c9edf78`; I re-checked each one against the code rather than taking the
rework on trust, and re-ran the tests with the fix reverted to see them fail.

### Findings 1–3: confirmed fixed, and the tests really do bite

- **1 (mongrel scope).** `resolveClimbBoardScope` now resolves board and layout
  as one pair: same board as the host → the host's layout is a real stand-in;
  different board → `layoutId: null`. Reverting it to the per-field fallback
  gives `2 failed | 113 passed` in `packages/shared/board-config` and
  `2 failed | 26 passed` in the picker. Not a test that passes either way.
- **2 (test pinned the defect).** `{ boardType: 'tension' }` against
  `{ kilter, 8 }` now expects `{ tension, null }`, and the pairing property case
  asserts the invariant directly instead of one example of it.
- **3 (`toBoardName` guard).** Fixed in the picker — an unrecognised board isn't
  the host's, so the host board is kept *without* the climb's layout.
  **This branch had no test.** Reverting it to `toBoardName(...) ?? boardName`
  still passed all 28 cases, so the fix was unguarded. Added
  `drops the climbs layout when its board is not a board this build knows`:
  with `boardType: 'not-a-real-board', layoutId: 10` it asserts both of the host
  board's playlists stay listed (layout 1 *and* layout 10), the other board's
  does not, and inline create stays hidden. Reverted → `1 failed | 28 passed`.

### The dead-branch question, answered rather than left open

`climbFitsPlaylistBoard`'s `climb.layoutId === null → true` is now live on the
client, so the earlier automated review's follow-up is worth answering
explicitly: yes, the intent is **show them and let the server reject**, not hide
layout-specific playlists. Hiding them would empty the picker in exactly the
state where the climber most likely does have the right playlist — most
playlists carry a layout — and the failure mode is already handled: `handleToggle`
rolls the optimistic checkmark back and surfaces the server's message, so a
rejected add is visible, not silent. Both halves are covered:
`does not reject on a guess when the climb layout is unknown` (unit) and
`keeps every layout of the board when the climbs layout is unknown`
(`filterPlaylistsByBoard`).

### Nothing rebutted

I went in expecting to push back on at least one of the five and didn't find a
place to. 4 and 5 stay as they are — both are your call, both are written into
"Not changed", neither is something I should decide unilaterally:

- **angle** — `handleOpenActions` passing the drawer's own board config fixes
  board, layout *and* angle across favorite / tick / queue at once. Say the word.
- **migration 0202** — happy to lift it into its own PR so the `db-migration`
  label stops riding a mobile-only change.

### Local

`vp run typecheck:mobile` clean · `vp check` **0 errors** (324 warnings, all
reproduce on clean `main`, #4488) · `InlinePlaylistPicker` **29 passed** ·
`sort-filter-playlists` 14 passed · `packages/shared/board-config` **115 passed**
across 10 files. Backend Vitest left to CI (shared :5433 worker DBs across
worktrees). Branch is rebased on `main` (`a71db63`).

PR #4576, opened 2026-08-19
fix(web): make MoonBoard set slugs resolve to the sets they name

marcodejongh:
## Summary

Every MoonBoard climb URL on www is broken, in two independent ways — and the first of them is **not MoonBoard-only**. This is the parser half of #4493, and both bugs are live in production right now, independent of the sitemap work.

### The uuid: 6,355 URLs in production's own sitemap 404 today

`extractUuidFromClimbSegment` matched a 32-character unbroken hex run at the end of the `<name-slug>-<uuid>` segment. With no match the extractor falls back to the whole segment, so `getClimb` was asked for `half-past-thirteen-42359a66-…` — a uuid nobody has — and the route called `notFound()`.

That was written up as a MoonBoard problem. It is not. Measured against production today:

| board | 32-hex uuids | dashed 36-char uuids |
|---|---|---|
| moonboard | 86 | **281,241** |
| kilter | 360,513 | **48,416** |
| tension / decoy / grasshopper / soill / touchstone | all | 0 |

So **Kilter carries 48,416 dashed-uuid climbs**, and 6,614 of them clear the tier-2 bar. Counting the URLs actually in the live sitemap right now:

```
/sitemaps/climbs/1.xml   1,466 dashed-uuid URLs
/sitemaps/climbs/2.xml   2,240
/sitemaps/climbs/3.xml   2,223
/sitemaps/climbs/4.xml     426
                        ------
                         6,355 URLs www submits to Google and then 404s
```

Fetched live, three at random, all from `climbs/2.xml`:

```
GET /kilter/original/12x12-square/screw_bolt/40/view/half-past-thirteen-42359a66-48e0-4368-ade2-2626af7b9def   404
GET /kilter/original/12x12-square/screw_bolt/60/view/how-do-i-delete-holds-4238fe4e-9183-4053-9266-bf8e99296821 404
GET /kilter/original/12x12-square/screw_bolt/40/view/gold-ship-4242747c-a60d-4eb8-8c30-520d733ca2f7             404
```

The same climb reached by its **bare** uuid renders, which is what isolates the extractor as the cause rather than the data:

```
GET /kilter/original/12x12-square/screw_bolt/40/view/42359a66-48e0-4368-ade2-2626af7b9def                       200
GET /decoy/.../view/numerous-goose-00377e6917e848eaa0b35e2c7704ffec  (32-hex control, same sitemap)            200
```

And the 404 page's own `<link rel="canonical">` points straight at that 404.

The regex now accepts either shape. The dashed alternative cannot steal a match from an Aurora segment, because it has to end at the segment end with its final twelve-hex group preceded by a `-`, and the character twelve back from the end of a 32-hex uuid is always a hex digit.

`isUuidOnly` deliberately does **not** gain the dashed form, and a test pins that with the reason. Its only caller that would change behaviour is the slug-redirect branch of `/[board_name]/…/view/[climb_uuid]`, which resolves its layout through `getLayouts(board_name)` — the Aurora `board-constants` tables, which carry no MoonBoard rows at all (`getAllLayouts('moonboard')` returns `[]`, verified). Returning true there would send every bare-uuid MoonBoard climb URL into a branch that ends in `notFound()`, costing a working page to gain a 308 the canonical already achieves. The `/play` → `/view` redirect pages 404 MoonBoard for the same reason; that is unchanged here and filed rather than widened.

### The set slug

`generateSetSlug` (the emitter, in `@boardsesh/play-view`) joins set-name slugs with `_`. `getMoonBoardSetsBySlug` (the parser, `packages/web/app/lib/url-utils.server.ts`) split the incoming slug on `-` and substring-matched the pieces against set names, so it could not tell one subset of hold sets from another.

Two consequences, both reproduced by the tests in this PR:

- `generateSetNameSlug('Screw-on Feet')` is `screw`, and a `-`-split never yields `screw` as a standalone part. So the **Masters 2017** full-set URL parsed back without set 15 and the **Masters 2019** one without set 20 — the tuple every link, share card and OG render for those layouts uses. `generateMetadata` builds its canonical from the *parsed* params, so the page canonicalised itself onto a URL nobody asked for.
- In the other direction it matched far too much: `hold-set-a` contains the part `set`, a substring of every `Hold Set *` name on the layout, so a single-set URL resolved to all of them and `getMoonBoardDetails` drew hold-set art the URL never named.

The fix is generate-and-compare — the rule the Expo app has already shipped in `resolveMoonBoardSegmentsToIds`, whose comment names web's parser as the deliberately-different one. Split on `_`, select the layout's sets whose `generateSetNameSlug` is one of those parts, accept only if re-emitting the selection rebuilds the incoming slug byte for byte. www was the outlier producing a third reading of the same URL; now it agrees with app.boardsesh.com.

The all-sets fallback stays for a slug that does not rebuild, so a hand-edited or pre-slug link renders the whole layout instead of 404ing. It is now the only lenient path, and it is pinned by a test rather than left implicit.

No new redirect — a legacy partial slug keeps serving 200, so there is nothing needing `es`/`fr`/`de` siblings.

### Verified end to end

Rebased onto current `main` and re-run against a full dev image through `vp run dev`. All three shapes render 200 with a `<link rel="canonical">` byte-identical to the requested path:

```
/moonboard/2016/standard-11x18-grid/original-school-holds_hold-set-b_hold-set-a/40/view/can-you-do-it-dcb417f0-225f-59a6-8758-60d123532dc5
  200 · <h1>Can you do it? — 7c+/V10</h1> · canonical identical
/moonboard/masters-2017/standard-11x18-grid/wooden-holds_screw_original-school-holds_hold-set-c_hold-set-b_hold-set-a/40/view/for-finn-663d3df8-…
  200 · canonical identical  (the `screw` part the old parser could never produce)
/moonboard/masters-2019/standard-11x18-grid/wooden-holds-c_wooden-holds-b_wooden-holds_screw_original-school-holds_hold-set-b_hold-set-a/40/view/ganan-las-negras-00087864-…
  200 · canonical identical
```

The dev image has no dashed-uuid *Kilter* tier-2 climb to demonstrate the 6,355-URL repair on, so that arm is evidenced by the production 404/200 pair above plus the `parseClimbRoutePath` unit case — same extractor, same segment shape, only the board prefix differs.

Part of #4493. Stacked below #4578, which must not merge first — dropping the sitemap's MoonBoard hold-out ahead of this parser fix would submit tens of thousands of URLs that canonicalise elsewhere.

**Production targets:** `www.boardsesh.com`, `app.boardsesh.com`, **and the native store fleet**. The set-slug fix is web-only, but the uuid fix is in `packages/shared/play-view`, which is in the mobile graph — so merging to `main` auto-publishes a production OTA. It is JS only, no native fingerprint move. The Expo app has the same deep-link bug today (`board-route-target.ts` calls the same extractor), so this fixes it there too — for MoonBoard and for those 48,416 Kilter climbs. No BLE code, so no Fable review.

## Release Notes

- Shared climb links open the climb again instead of a not-found page, on the web and in the app. Every MoonBoard link was affected, and so were about 48,000 Kilter climbs.
- MoonBoard links now light exactly the hold sets named in the URL. Masters 2017 and Masters 2019 links were quietly dropping Screw-on Feet, and a single-hold-set link lit the whole layout.

## Test plan

Every oracle below was mutation-tested on the rebased branch: the mutation applied, the test watched go red, then reverted, tree verified clean.

**`packages/web/app/lib/__tests__/slug-utils-catalogue-round-trip.test.ts`** — MoonBoard joins the suite it was explicitly filtered out of (`SUPPORTED_BOARDS.filter((b) => b !== 'moonboard')`, which is exactly why this bug survived). Walks all **291** tuples: every one of the 7 layouts × every non-empty subset of its hold sets, emitted through `tryConstructSlugViewUrl` and fed back through `parseBoardRouteParamsWithSlugs`. The file's `dbz` stub throws, so "MoonBoard resolves off the static tables" is asserted, not assumed.

- Mutation — restore the `-`-splitting substring matcher: **6 of 7 layouts red** (only `moonboard-2010`, which has a single hold set, passes trivially).
- Mutation — drop just the `generateSetSlug(...) === setSlug` rebuild check: **12 red** in this file. (The 291-tuple round-trip arm alone cannot see it — every slug it feeds back was built by the emitter and rebuilds by construction — which is what the second suite in the same file is for.)

**`MoonBoard set slugs this app never emits`** (same file) — the strict/fallback contract: a reordered two-set slug and a repeated part both fall through to the layout's full set; an unrecognised slug still renders the whole layout rather than 404ing. The reorder case asserts the reversal actually differs from the canonical first, so it cannot go vacuous on a symmetric slug.

**`packages/web/app/lib/seo/sitemap/__tests__/moonboard-canonical-identity.test.ts`** (new) — the property the whole issue turns on, tested as a composition rather than three per-segment comparisons: `climbRowsToItems` builds the path, `parseBoardRouteParamsWithSlugs` parses it the way the route handler does, `getBoardDetailsForBoard` is what `generateMetadata` calls on the result, and `buildCanonicalClimbViewUrl` emits the canonical — asserted byte-identical to the emitted path, for both angles and for a named and an unnamed climb. Same for the boards-shard `/list` URL through `popularConfigListUrl` vs `buildCanonicalClimbListUrl`, including a `//` guard. Set tuples are written as literals, not re-derived by calling the source's own helper. It also asserts `parsed.climb_uuid` equals the uuid the sitemap emitted — without that the canonical still matches (it re-slugs whatever it is handed) while the page it names 404s.

- Mutation — restore the old parser: **masters-2017 and masters-2019 red on both surfaces**, 4 tests. That is the shape of the production bug, reproduced.

**`packages/shared/play-view/src/__tests__/readable-url-utils.test.ts`** — the uuid extractor. Cases: a dashed uuid out of a name-slugged segment, a bare dashed uuid, a dashed uuid through a whole `parseClimbRoutePath`, and an adversarial no-steal case (`deadbeef-cafe-babe-face-<32hex>`).

- Mutation — restore `/[0-9A-F]{32}$/i`: **2 red here, and 14 red across the two web suites.**
- Mutation — a plausible-looking loose fix, `/[0-9A-F-]{36}$|[0-9A-F]{32}$/i`: **4 red**, including the no-steal case, which is why that case exists.

**`packages/web/app/lib/__tests__/url-utils.test.ts`** — pins `isUuidOnly` as false for a dashed uuid, with the `getAllLayouts('moonboard') === []` reason in the comment, so the tempting "widen this too" change has to argue with a test.

Gates, all from the rebased branch:

- `vp run typecheck` — pass (full, including the web `next build`)
- `vp test run --reporter=agent --project web` — **362 files, 4,069 tests**; one unrelated `waitFor` flake in `gym-settings-panel.test.tsx` under full-suite load, green on its own
- `vp run test:mobile` — **693 files, 6,725 tests**; one failure, `board-detail-fields.test.ts > resolves the size from the bundled tables when the server sends null`, which **reproduces identically on `main` at 25322da86** and is untouched by this branch
- `vp test run --reporter=agent --project i18n` — 101 pass
- `vp run check:i18n` / `vp run check:i18n:orphans` — OK
- `vp check` — the only 2 errors are the known local `@gorhom/bottom-sheet` artefact (`packages/mobile/web-runtime` nested install missing on this box; CI installs it), untouched here
- pre-commit hook — pass, no `--no-verify`

No user-facing strings added, so no catalog work and no Spanish/French/German glossary exposure.

Production counts here are from a read-only `SELECT` against the production database and from fetching `www.boardsesh.com` directly, both on 2026-08-22; the round-trip and canonical numbers are from the dev image. Canonical `<link>` was checked in `en-US` only, not in `es`/`fr`/`de`.

### DIFF
--- a/packages/shared/play-view/src/__tests__/readable-url-utils.test.ts
+++ b/packages/shared/play-view/src/__tests__/readable-url-utils.test.ts
@@ -31,6 +31,13 @@ const CLIMB_UUID = '0A1B2C3D4E5F60718293A4B5C6D7E8F9';
  */
 const HEX_RUN_CLIMB_NAME_SLUG = 'beefcafe0ff1cedeadbeefcafe0ff1ce';
 
+/**
+ * A real MoonBoard uuid, copied off the dev image. MoonBoard is the one board
+ * whose climbs carry the dashed 36-character RFC-4122 form; every Aurora board
+ * carries the 32-character unbroken form above.
+ */
+const MOONBOARD_CLIMB_UUID = '9fe54099-6fdd-5adb-b82f-2d7bcb10d4ad';
+
 describe('extractUuidFromClimbSegment', () => {
   it('pulls the uuid out of a name-slugged segment', () => {
     expect(extractUuidFromClimbSegment(`crimpy-thing-${CLIMB_UUID}`)).toBe(CLIMB_UUID);
@@ -55,6 +62,29 @@ describe('extractUuidFromClimbSegment', () => {
 
   it('takes the uuid at the end even when the name runs longer than a uuid', () => {
     expect(extractUuidFromClimbSegment(`${HEX_RUN_CLIMB_NAME_SLUG}deadbeef-${CLIMB_UUID}`)).toBe(CLIMB_UUID);
+  });
+
+  it('pulls a dashed MoonBoard uuid out of a name-slugged segment', () => {
+    expect(extractUuidFromClimbSegment(`to-dokids-yuito-${MOONBOARD_CLIMB_UUID}`)).toBe(MOONBOARD_CLIMB_UUID);
+  });
+
+  it('passes a bare dashed MoonBoard uuid through', () => {
+    expect(extractUuidFromClimbSegment(MOONBOARD_CLIMB_UUID)).toBe(MOONBOARD_CLIMB_UUID);
+  });
+
+  it('carries a dashed MoonBoard uuid through a whole route parse', () => {
+    expect(
+      parseClimbRoutePath(`/moonboard/2016/standard/hold-set-a/40/view/to-dokids-yuito-${MOONBOARD_CLIMB_UUID}`)
+        ?.climbUuid,
+    ).toBe(MOONBOARD_CLIMB_UUID);
+  });
+
+  it('does not let the dashed shape steal the tail of an Aurora uuid', () => {
+    // The dashed alternative must end at the segment end and its final group is
+    // preceded by a `-`; the character twelve back from the end of a 32-hex uuid
+    // is always a hex digit, so it can never match here. A name slug that looks
+    // like the head of a dashed uuid is the adversarial case.
+    expect(extractUuidFromClimbSegment(`deadbeef-cafe-babe-face-${CLIMB_UUID}`)).toBe(CLIMB_UUID);
   });
 
   it('carries the hex-named climb through a whole route parse', () => {
--- a/packages/shared/play-view/src/readable-url-utils.ts
+++ b/packages/shared/play-view/src/readable-url-utils.ts
@@ -443,8 +443,21 @@ const numericSegmentRegex = /^\d+$/;
  * perfectly valid shared link queried a uuid nobody has and rendered
  * not-found. A bare uuid still matches: the whole segment is then the run at
  * the end.
+ *
+ * Two shapes, because the catalogue holds two. Aurora climbs carry a 32-char
+ * unbroken hex uuid; every MoonBoard climb carries a dashed 36-char RFC-4122
+ * uuid (`9fe54099-6fdd-5adb-b82f-2d7bcb10d4ad`) — measured on the dev image:
+ * 142,566 MoonBoard rows, all dashed, and no other board has one. With only the
+ * 32-char form, no MoonBoard climb URL parsed: the whole `<name>-<uuid>` segment
+ * was handed on as the uuid, so the climb query missed and the page 404'd, while
+ * the same page reached by its bare uuid rendered a `<link rel="canonical">`
+ * pointing straight at that 404.
+ *
+ * The dashed alternative cannot steal a match from an Aurora segment: its final
+ * group must be preceded by a `-`, and the character 12 back from the end of a
+ * 32-hex uuid is always a hex digit.
  */
-const climbUuidRegex = /[0-9A-F]{32}$/i;
+const climbUuidRegex = /(?:[0-9A-F]{8}-[0-9A-F]{4}-[0-9A-F]{4}-[0-9A-F]{4}-[0-9A-F]{12}|[0-9A-F]{32})$/i;
 
 function isNumericSegment(value: string): boolean {
   return numericSegmentRegex.test(value);
--- a/packages/web/app/lib/__tests__/slug-utils-catalogue-round-trip.test.ts
+++ b/packages/web/app/lib/__tests__/slug-utils-catalogue-round-trip.test.ts
@@ -280,6 +280,11 @@ function nonEmptySetSubsets(setIds: readonly number[]): number[][] {
 
 const moonBoardLayoutKeys = Object.keys(MOONBOARD_LAYOUTS) as MoonBoardLayoutKey[];
 
+/** 32 hex characters in the dashed 8-4-4-4-12 form MoonBoard climbs actually use. */
+function toDashedUuid(hex: string): string {
+  return `${hex.slice(0, 8)}-${hex.slice(8, 12)}-${hex.slice(12, 16)}-${hex.slice(16, 20)}-${hex.slice(20)}`;
+}
+
 describe('MoonBoard catalogue round-trip (the static tables, through the same www resolver)', () => {
   it.each(moonBoardLayoutKeys)(
     '%s: every emitted set-slug subset parses back to the ids it was built from',
@@ -292,7 +297,13 @@ describe('MoonBoard catalogue round-trip (the static tables, through the same ww
 
       for (const setIds of subsets) {
         const label = `${layoutKey} / sets [${setIds.join(',')}]`;
-        const climbUuid = `${layout.id}${setIds.join('')}`.padStart(32, '0');
+        // Dashed, because that is the only shape MoonBoard has: every one of the
+        // 142,566 MoonBoard rows on the dev image carries a 36-character
+        // RFC-4122 uuid, and no other board does. An Aurora-shaped 32-hex uuid
+        // here would leave the climb segment untested — the uuid extractor
+        // matched only the unbroken form, so a real MoonBoard climb URL parsed
+        // back with its whole `<name>-<uuid>` segment as the uuid and 404'd.
+        const climbUuid = toDashedUuid(`${layout.id}${setIds.join('')}`.padStart(32, '0'));
 
         const emittedPath = tryConstructSlugViewUrl(
           'moonboard',
--- a/packages/web/app/lib/__tests__/slug-utils-catalogue-round-trip.test.ts
+++ b/packages/web/app/lib/__tests__/slug-utils-catalogue-round-trip.test.ts
@@ -41,6 +41,7 @@ import {
 import { getLayoutBySlug, getSetsBySlug, getSizeBySlug } from '@/app/lib/slug-utils';
 import { tryConstructSlugViewUrl } from '@/app/lib/url-utils';
 import { parseBoardRouteParamsWithSlugs } from '@/app/lib/url-utils.server';
+import { MOONBOARD_LAYOUTS, MOONBOARD_SETS, MOONBOARD_SIZE, type MoonBoardLayoutKey } from '@/app/lib/moonboard-config';
 
 /**
  * The acceptance criterion for #4362: a round-trip over the WHOLE catalogue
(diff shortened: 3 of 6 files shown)


marcodejongh:
Stacked: #4578 (the sitemap half) branches off this. #4578 must not merge first — removing the sitemap's MoonBoard hold-out ahead of this parser fix would submit tens of thousands of URLs that canonicalise elsewhere.

marcodejongh:

Every dev-image figure in the body re-derived against production (PostgreSQL 16.9, 11 GB, `board_climbs` 885,528 rows). Aggregates only.

### The headline: the uuid bug is not MoonBoard-only

|---|---|---|---|
| MoonBoard rows carrying the dashed 36-char uuid | 142,566 | **281,241** | +138,675 |
| MoonBoard rows carrying a bare 32-hex uuid | 0 (body says "all … carry the dashed form") | **75** | +75 |
| **Kilter** rows carrying the dashed 36-char uuid | 0 ("no other board has one") | **47,903** | +47,903 |
| Kilter rows carrying a 50-char `json-import-…` uuid | not measured | **3,220** | — |
| **Kilter tier-2 climbs the live climbs shard submits today whose uuid is dashed** | not measured | **6,285 of 32,826 (19.1%)** | — |
| 291 round-trip tuples | 291 | 291 (static tables, no DB input) | 0 |

**"and no other board has one" is false in production.** Kilter has 47,903 dashed-uuid climbs, and 6,285 of them are in the tier-2 selection the climbs shard already submits to Google on `main` today — under `main`'s `/[0-9A-F]{32}$/i` those `…/view/<name-slug>-<8>-<4>-<4>-<4>-<12>` URLs yield no match, the extractor falls back to the whole segment, and the page `notFound()`s. So this PR is not only unblocking MoonBoard: **it repairs 6,285 Kilter URLs that www is submitting to the index today and that 404 when a crawler follows them.** Worth saying in the body — it is the strongest argument for landing the uuid commit on its own schedule.

(Cross-check: 26,541 tier-2 Kilter uuids match `main`'s regex + 6,285 match the dashed form + **0** match neither = 32,826 total, so the 3,220 `json-import-…` uuids are not in tier 2 and no third shape is unhandled.)

### One reasoning gap, not a defect

The body justifies leaving `isUuidOnly` narrow because "its only caller that would change behaviour … resolves its layout through `getLayouts(board_name)` — the Aurora `board-constants` tables, which carry no MoonBoard rows". That is true for MoonBoard and **not** true for Kilter's 47,903 dashed uuids, which do have Aurora layouts. No regression (`main` already returns false for a dashed uuid, so nothing changes for them), but the stated reason only covers half the affected population and should be narrowed to "for MoonBoard" rather than read as a general rule.

### SQL

```sql
-- uuid shapes per board
SELECT board_type,
       length(uuid) AS uuid_len,
       (uuid ~ '^[0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{12}$') AS dashed_rfc4122,
       count(*)::int AS climbs
FROM board_climbs GROUP BY 1,2,3 ORDER BY 1,2;

-- 50-char shape
SELECT board_type, left(uuid, 14) AS prefix_sample, count(*)::int
FROM board_climbs WHERE length(uuid) = 50 GROUP BY 1,2 ORDER BY 3 DESC;

-- Kilter tier-2 uuids by which regex they match, over the two groups the
-- climbs shard actually resolves in production (kilter/1 size 10 sets {1,20};
-- kilter/8 size 25 sets {26,27,28,29}) — the shard's own predicate, angle
-- filter and alias anti-join included.
WITH chosen AS (
  SELECT DISTINCT ON (s.climb_uuid) s.climb_uuid
  FROM board_climb_stats s
  JOIN board_climbs c ON c.uuid = s.climb_uuid AND c.board_type = s.board_type
  WHERE c.board_type = 'kilter'
    AND c.is_listed = true AND c.is_draft = false
    AND s.ascensionist_count >= 10
    AND s.angle IN (0,5,10,15,20,25,30,35,40,45,50,55,60,65,70)
    AND ( (c.layout_id = 1 AND c.compatible_size_ids @> ARRAY[10]::int[]
             AND c.required_set_ids <@ ARRAY[1,20]::int[])
       OR (c.layout_id = 8 AND c.compatible_size_ids @> ARRAY[25]::int[]
             AND c.required_set_ids <@ ARRAY[26,27,28,29]::int[]) )
    AND NOT EXISTS (SELECT 1 FROM board_climb_aliases a
                    WHERE a.board_type = c.board_type AND a.alias_uuid = c.uuid
                      AND a.alias_uuid <> a.canonical_uuid)
  ORDER BY s.climb_uuid, s.ascensionist_count DESC,
           COALESCE(s.angle = c.angle, false) DESC, s.angle ASC
)
SELECT
  count(*) FILTER (WHERE climb_uuid ~* '[0-9a-f]{32}$')::int AS matches_main_regex,
  count(*) FILTER (WHERE climb_uuid ~* '[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}$')::int AS matches_dashed,
  count(*) FILTER (WHERE climb_uuid !~* '[0-9a-f]{32}$'
                     AND climb_uuid !~* '[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}$')::int AS matches_neither,
  count(*)::int AS total
FROM chosen;
```

Nothing here changes the diff. The two conclusions that move are: the "no other board" claim should come out, and the uuid commit's production value is ~6,285 currently-404ing submitted Kilter URLs on top of the MoonBoard work.

Issue #4473, opened 2026-08-15
SEO: /setter/[username] is an indexable empty shell AND answers 200 for any username


marcodejongh:
`/setter/[setter_username]` is an indexable public page (it emits a self-canonical, hreflang alternates and an OG image via `createPageMetadata`) but its server HTML contains no setter content at all — only the site header and footer. Verified live on production 2026-08-15 during epic #4358's definition-of-done curl matrix.

## Evidence (production, 2026-08-15)

`GET https://www.boardsesh.com/setter/KilterStudio` → `200`, and the whole `<body>` strips to **379 characters**, every one of them chrome:

```
> Boardsesh Sign in Start climbing Boardsesh Find climbs, share them, and see what's on your
gym's wall. The climbing happens in the app. Start climbing Home About Help Developer docs
Playlists Aurora migration Legal Privacy Works with Kilter, Tension and MoonBoard. Not
affiliated with Aurora Climbing or Moon Climbing. 🇺🇸 KilterStudio - Setter | Boardsesh | Boardsesh
```

`<h1>` count: 0. `<h2>` count: 0. `<img>` count: 0. Internal links: 17, all of them header/footer nav.

For comparison, in the same sweep the climb front door renders 2,485 chars of real content with one `<h1>`, 79 `<img>` and 45 internal links, and the `/list` front door renders 2,903 chars with 50 crawlable climb links.

## Root cause

`packages/web/app/setter/[setter_username]/page.tsx` is a server component that renders `<SetterProfileContent username={…} />`, and `packages/web/app/setter/[setter_username]/setter-profile-content.tsx:1` is `'use client'`. It fetches the setter over GraphQL from a `useEffect` via `createGraphQLHttpClient` and renders `LoadingSpinner` until that resolves.

So the first server-rendered HTML is a spinner. CLAUDE.md's SEO rules for indexable pages are explicit:

> First server-rendered HTML must contain meaningful copy: one `h1`, descriptive paragraphs, crawlable links. No indexable spinner-only pages.

This predates the web-reposition epic, but it matters much more now: after W-16 (#4467) deleted the climbing UI, `www` **is** the SEO surface, and setter pages are one of the six public search surfaces the epic's SEO section names.

## Impact

[...]

Issue #4465, opened 2026-08-15
SSR the setter profile page so /sitemaps/setters.xml can be filled


marcodejongh:
Follow-up from #4434 (PR #4464), which shipped `app/sitemaps/setters.xml/route.ts` as a **declared-empty** shard.

## Why it ships empty

1. **Spinner-only SSR.** `packages/web/app/setter/[setter_username]/setter-profile-content.tsx` is a client component with `loading` initialised `true` and a `useEffect` GraphQL fetch, so the first server HTML is a `LoadingSpinner` — no `<h1>`, no copy. "No indexable spinner-only pages" is a standing repo rule.
2. **Soft-404 farm.** `getSetterOgSummary` (`app/lib/seo/dynamic-og-data.ts`) never returns null — it falls back to the raw username — so `/setter/{anything}` answers 200. Submitting that spends crawl budget the climb shards (W-23) need.
3. **Volume.** The dev database has **108,000** distinct `(board_type, setter_username)` pairs, well over the 11,250-item per-shard budget, so the shard needs paging too.

## The work

- A `SetterSeoFragment` server component in the shape of `app/components/climb-detail/climb-view-seo-fragment.tsx`, so the first HTML carries an `<h1>` and real copy.
- A real 404 when the setter has no climbs, so `/setter/{arbitrary}` stops 200ing.
- Fill `buildSetterEntries()` in `app/lib/seo/sitemap/setter-entries.ts` from `board_setter_stats`, paged across `setters-1.xml` / `setters-2.xml` / … (the registry already supports N shards).
- Decide whether the page should be indexable at all before the fragment lands — it is today, unsubmitted.

marcodejongh:
Closing as a duplicate of #4473, which stays open as the canonical issue for this work.

**Why they're the same:** Both describe the same two defects on the same route: packages/web/app/setter/[setter_username] is a client component whose first server HTML is a LoadingSpinner (no h1, indexable spinner-only page), and getSetterOgSummary never returns null so /setter/{anything} answers 200. Both prescribe the same fix: server-render the setter content and notFound() an unknown setter.

Anything in here that #4473 doesn't already cover should be moved there rather than lost. Reopen if you think this is a distinct defect.

_Filed by an issue-triage sweep, 2026-08-24._

PR #4577, opened 2026-08-19
feat(web): server-render the setter profile and 404 unknown setters

marcodejongh:
## Summary

`/setter/[setter_username]` is linked from every climb front door and was, in the first server HTML, entirely chrome: zero `<h1>`, zero images, zero climb links, ~365 characters of stripped body text. The page rendered a `'use client'` component that fetched over GraphQL from a `useEffect` with `loading` initialised true, so the server pass only ever produced a spinner.

It also answered **200 for any username at all**. `getSetterOgSummary` LEFT JOINs a profile onto `(SELECT 1) AS seed`, so it always returned exactly one row and never threw — the `createNoIndexMetadata` catch branch was unreachable. `GET /setter/zzzz-not-a-real-setter-9182` returned 200 with `<title>zzzz-not-a-real-setter-9182 - Setter | Boardsesh</title>` and no `<meta name="robots">` at all. A soft-404 farm on an indexable route.

Both close here.

### What changed

- **`getSetterOgSummary` can return null.** An `EXISTS` over the setter's `is_listed AND NOT is_draft` climbs. The page 404s on null and `/api/og/setter` 404s on null, so the HTML surface and the OG surface answer the same question the same way.
- **`server-setter-data.ts`** — the page's own data layer, reading Postgres directly with Drizzle, the pattern `profile/[user_id]/server-profile-data.ts` and `lib/data/list-page-data.server.ts` already establish. It is deliberately not the GraphQL pair the client used: those resolvers carried no visibility predicate and return neither `compatible_size_ids` nor `required_set_ids`. Rethrows on query failure rather than degrading to an empty 200, per `fetchFrontDoorListPage`'s recorded reasoning.
- **`setter-climb-links.ts`** — each climb's canonical config comes from the same `resolveClimbSitemapGroups` the climbs shard uses. This is the part most likely to ship wrong silently: the client path built every href from `getBoardDetailsForPlaylist` (largest size + every set) while the sitemap chooses most-physical-boards, and because the climb view page self-canonicalises to whichever tuple it is served under, every setter row was manufacturing a second indexable URL for a climb Google had already been told about at a different address. Every wrong link still resolves and renders a real page, so nothing but a computed oracle catches it.
- **No hardcoded MoonBoard exclusion.** MoonBoard is absent from the resolved groups today (its set slugs do not round-trip through the MoonBoard page's parser), so its climbs render as plain rows with correct board art and no anchor. When #4493 lands they start linking with no second edit here.
- **The page** now emits one `<h1>`, a summary paragraph, up to 50 crawlable climb anchors per `?page`, `rel="prev"`/`rel="next"` capped at `FRONT_DOOR_MAX_INDEXABLE_PAGE`, a 404 past `FRONT_DOOR_MAX_PAGE`, `ProfilePage` + `ItemList` JSON-LD, and two descriptive internal links. Two client islands survive — follow and share — and the server pass reads no session.
- **Backend visibility fix.** `setterProfile`, `setterClimbs` and both `setterClimbsFull` modes filtered on `setter_username` alone, so anyone who asked got the setter's drafts and unlisted climbs and `setterProfile.climbCount` counted them. None of the three is authenticated against the setter, so there is no own-drafts case being protected. All four query paths gain the predicate.
- Four locales updated (en-US/es/fr/de); the two dead not-found keys removed with the component that used them.

### Numbers — dev image, production reconciliation outstanding

Measured against the local `ghcr.io/boardsesh/boardsesh-dev-db` image, **not** production. Prior work in this campaign found the dev image runs roughly a third short of production on some counts, so treat these as shape, not scale.

| | dev image |
|---|---|
| distinct setters with ≥1 climb of any kind | 104,140 |
| ...with ≥1 `is_listed AND NOT is_draft` climb | 91,946 |
| ...with ≥3 visible climbs | 35,950 |
| drafts/unlisted-only — **these flip from 200 to 404** | 12,194 |
| visible setters whose climbs are all MoonBoard (render unlinked) | 21,154 |

Query cost on the dev image, for the largest setter (`forestfox98`, 4,749 visible climbs), warm: **~95 ms** for the 51-row page, **~16 ms** for the board-type/count query.

Production reconciliation of all of the above is outstanding — no production database was read for this PR.

### Blast radius of the 404

The 12,194 drafts/unlisted-only setters (dev-image count) lose their page. None are indexed: the setters sitemap shard has never been submitted, and these pages carry no content today anyway. A setter who later unlists their whole catalogue now gets a hard 404 rather than an empty profile, and `packages/web` has no `not-found.tsx`, so they get Next's default page. Worth a follow-up, not a blocker.

### Review round — four crawler-facing defects, all reproduced before fixing

Three adversarial lenses read this branch. One verdict was REFUTED outright; the findings that survived are below with the evidence that reproduced them and the mutation that now guards each. All numbers are **dev image**.

**1. The setter page and the climbs sitemap picked different angles (major).** The config tuple was shared; the angle — also a path segment — was not. `buildChosenSubquery` ordered by `ascensionist_count DESC, COALESCE(stats.angle = climbs.angle, false) DESC, angle ASC` over angles in `ANGLES[board]`; the setter query ordered by `ascensionist_count DESC NULLS LAST, angle ASC` over every angle in the stats table. Replaying both orderings verbatim against the dev image over the 85,596 tier-2 climbs: **28 disagreed**, every one an exact ascent tie where the sitemap picked `board_climbs.angle` and the page picked the lower tied angle. Worked example, kilter layout 8 `8EE0F93116DC4FBDB3436053129C6396` ('air filter', 26 ascents at both 40 and 45): page linked `/40/view/…`, sitemap submitted `/45/view/…`.

Both builders now consume one `published-angle.ts`. Re-running the same comparison after the change: **0 of 85,596 disagree**.

**2. The setter query could link an angle no route table carries.** `board_climb_stats` holds 28 `grasshopper` rows at angle `-5`, and for one visible climb (`e428865c…`, its only stats row) `-5` was the argmax — so the row linked `/grasshopper/…/-5/view/…`, a URL the climbs shard refuses to publish with the comment "that URL 404s". The angle pick is now guarded by a `CASE` over the row's own `board_type` (one setter's climbs span boards with different angle tables); that climb falls back to 40, which is in `ANGLES` for every board we ship.

**3. The JSON-LD named page 1 on every page.** `?page=2` self-canonicalised to `?page=2` and carried `ItemList` positions 51–100 while the `ProfilePage` `@id`/`url` and `ItemList.mainEntityOfPage` all said `/setter/{name}` — structured data asserting that the page-1 URL contains items 51–100. The graph now builds its document URL from the same `frontDoorPagePath(basePath, page)` the canonical uses. The `Person` node keeps the bare profile URL, which is the entity's one address.

**4. An indexable page with zero crawlable climb links.** The page granted `index, follow` on ≥1 visible climb; the sitemap's own worth-crawling gate is *linkable* AND ≥3. Resolving the real `resolveClimbSitemapGroups` output against the dev image — **9** groups, not the 12 an earlier approximation reported, because that approximation skipped `isResolvableGroup` — **41,844 of the 125,082 setters who now answer 200 — 33.5% — have no climb on any resolvable configuration** (measured against production; the dev image gave 22,490 of 91,946, and the real share is 1.86x that). Their page renders an `<h1>`, fifty board images and not one anchor. Those pages now emit `noindex, follow`. Exposure was limited (nothing internal links to them; they are orphans, not a pushed farm) but a share link, an OG card or an external link puts one in the index.

**5. A second `decodeURIComponent` on the route param.** Next already decodes dynamic segments (`getRouteMatcher`, `next/dist/shared/lib/router/utils/route-matcher.js:19`, applied to `routes-manifest.dynamicRoutes` — app routes included). Decoding again turns `/setter/50%25` into an unhandled `URIError` thrown from **both** `generateMetadata` and the page body, before the `try` in either — a 500 where a 404 or a 200 belongs — and silently rewrites `/setter/abc%2541` into `abcA`, a canonical naming a different setter. Both calls are gone. **This is a reachable production 500, not a hypothetical**: production has one setter username containing `%`, and that setter has ≥1 visible climb, so their page answers 200 today and `/setter/…%25…` throws `URIError: URI malformed` on `main` right now. (The dev image has 0 such usernames, which is why an earlier pass read it as latent.) That setter has fewer than 3 linkable visible climbs, so they are not in #4579's submitted set — the exposure is page-level only. Pre-existing on `main`.

**6. The angle oracle was structurally blind.** `setter-climb-links.test.ts` feeds the same fixture `angle` to both the rendered href and the sitemap href, so the angle is an *input* to both sides and can never disagree there. The new oracle in `server-setter-data.test.ts` byte-compares the angle-selection `ORDER BY` extracted from the SQL **both real builders render**, with only the correlated subquery's table alias normalised. Nothing about the expected ordering is written down in the test.

Mutations applied to the new guards, each watched go red and reverted:

| mutation | result |
|---|---|
| drop the `COALESCE` tie-break from `published-angle.ts` | red (angle oracle) |
| drop `nulls last` from `published-angle.ts` | red (angle oracle) |
| hand-write the original `ascensionist_count desc nulls last, angle asc` in the setter query | red (angle oracle) |
| delete `publishableAngleWhere` from the setter subquery | red (route-table oracle) |
| point the JSON-LD document URL back at page 1 | red (`?page=2` oracle) |
| drop the lonely-page `noindex` | red (no-crawlable-link oracle) |
| restore the second `decodeURIComponent` | red — `URIError: URI malformed` (percent-sign oracle) |
| bind a different username in the profile query only | red (WHERE-parity oracle) |
| discard the live follower count again | red (follow-island oracle) |
| start the follower count at 0 instead of the server prop | red (follow-island + SSR oracles) |
| remove the `?page` range guard from `generateMetadata` | red (no-query-past-the-ceiling oracle) |

One deliberate behaviour change outside the setter tree: the climbs sitemap's angle ordering is now rendered as `desc nulls last` rather than a bare `desc`. It cannot change a row — `ascensionist_count >= 10` excludes NULL from that query — and stating it once is what lets the two builders be compared byte for byte.

**REFUTED, no change:** a claimed divergence between the setter page's and the sitemap's *board config* choice. Both call `resolveClimbSitemapGroups` on the same input; the only divergence was the angle, fixed above.

**Also taken, from the review bot's non-blocking note:** `SetterFollowIsland` fetched the whole setter profile to read one boolean and discarded the `followerCount` in the same response, so a signed-in viewer read the server snapshot until they themselves followed somebody. It now adopts the live count. Safe to overwrite rather than merge: the button is not rendered until `isFollowing` resolves, so there is no click this can land behind. `setter-follow-island.test.tsx` pins both halves — the server prop renders first (the half that has to stay in the crawlable HTML), and the live count replaces it.

### Deliberately out of scope

- `/setter/*` does **not** join the middleware CDN TTL list. The precondition (a personalization-free server render) is established by the island split, but shipping a shared `s-maxage` in the same PR as the split means a mistake in the split caches one viewer's follow state for everyone. Filed as #4580.
- `packages/web` has no `not-found.tsx`, so the new 404s render Next's default page. Pre-existing, but this PR makes it reachable. Filed as #4581.
- The follow island fetches the whole setter profile to read one boolean. Filed as #4582.
- The sitemap shard itself is #4465 → PR #4579, stacked on this branch.

## Release Notes

- Setter pages now load with the climbs already on them, and a setter's page shows only work they've actually published.

## Test plan

1. Open a setter page → climbs appear with no spinner.
2. Add `?page=5000` to it → you get not-found.
3. Open `/setter/madeupname` → not-found, not an empty profile.
4. Tap a climb → the right board and angle open.

## Risk

Risk: 3/5 — a new server-rendered surface on an indexable route, plus a backend
visibility predicate on four setter resolver paths. The 404 is the sharp edge:
it turns a route that answered 200 for any string into one that can now refuse,
so a visibility rule that is too strict hides a real setter. Contained by the
paired oracle (metadata noindexes on exactly the rule the page 404s on) and by
`setter-visibility.test.ts` seeding visible / draft / unlisted against a real
database. Nothing reaches `app.boardsesh.com` or the store fleet.

## Rebased onto main, 2026-09-05

The branch was ~890 commits behind. Three files conflicted and one change was
needed beyond resolving them.

- **`hreflang-return-metadata.test.ts`** — main moved `/setter/marco` out of the
  `cases` table into an explicit board-content carve-out (`01387f422`): setter
  pages now cross-canonicalise onto the default locale and emit no
  `alternates.languages`, because the twins are translated chrome over identical
  content. Took main's side; the carve-out's `load` gains the `searchParams` this
  branch adds to `generateMetadata`.
- **`api/og/setter/route.tsx`** — took main's shape (`withReadDeadline`,
  `dbzRead`) and grafted this branch's two visibility predicates onto the grade
  query.
- **`climb-query.ts`** — took this branch's `publishableAngles` /
  `publishedAngleOrderBy` extraction, then repointed the underlying list, below.
- **`published-angle.ts` now reads `getRoutableBoardAngles`, not `ANGLES`.** Main
  widened the sitemap's angle guard while this branch was out. Rebasing the file
  in as-written would have re-narrowed the climb sitemap to picker angles — 5°
  steps, and only 25°/40° on MoonBoard — and dropped every page stored at an
  in-between angle. Its own docstring already described the routable list; the
  implementation now matches it.

One consequence for the record: **`-5` is no longer an unroutable angle.** Main
made it a real Grasshopper URL, so the "linked at an angle no route table
carries" framing in the angle commit is stale for that board. What the CASE
guard still prevents is `-5` winning on a board with no such page, and the test
now asserts that directly — Grasshopper's branch contains it, Kilter's does not.

## Author verification

Every oracle below was mutation-tested: the mutation it claims to catch was applied, watched go red, then reverted. `git status --porcelain` clean after each.

- `setter-page-ssr.test.tsx` — renders the **page default export** through `renderToPipeableStream`/`onAllReady`, asserting exactly one `<h1>`, the summary copy and three `href="/kilter/.../view/..."` anchors, plus `notFound()` on null data and on `?page=5000`. *Mutation:* replaced the page body with a client component whose `loading` starts true (the pre-change shape) — all three tests red. Rendering the SEO fragment in isolation would have stayed green, which is why it renders the page.
- `setter-climb-links.test.ts` — computes the expected href from `climbRowsToItems(resolveClimbSitemapGroups(fixture))` inside the test, never a pinned string. *Mutation:* built the link from `getBoardDetailsForPlaylist` (what the client list did) — 2 tests red.
- `server-setter-data.test.ts` — byte-compares the **recorded** outer WHERE clauses of the two real builders, never a rebuilt predicate. *Mutations:* dropped the visibility filter from the climbs query only (2 red), then from the profile query only (2 red).
- `setter-page-ssr.test.tsx` (second oracle) — reads the JSON-LD back out of the rendered HTML and compares its climb URLs to the rendered anchors, so structured data can never advertise a URL the page does not link to. *Mutation:* gave the JSON-LD builder a different size id — red.
- `setter-visibility.test.ts` (backend, real DB) — seeds one visible, one draft and one unlisted climb and asserts all four resolver paths return exactly the visible one, plus `setterProfile` null for a drafts/unlisted-only setter. *Mutation:* removed the predicate from all four paths — both tests red.

Post-rebase, on this box: `packages/web/app/setter` 4 files / 23 tests green; the wider `app/lib/seo` + hreflang sweep 27 files / 279 tests green; `typecheck:backend` and `typecheck:shared` green. Backend Vitest was not run locally — this worktree shares the fixed test-infra ports with another checkout and the run hangs there; CI is the oracle for it.

Not verified: no production database was read, and no production page was fetched after the change — the "365 characters of chrome" and "200 for any username" figures are the pre-change measurements from the planning pass.

## Production targets

**`www.boardsesh.com` only.** `app.boardsesh.com` and the native store fleet are unaffected: no `packages/mobile` file changes, no native fingerprint change, no OTA. `profile` is in `MOBILE_NAMESPACES` and the catalog gains keys, but mobile reads none of them and no GraphQL operation document changed.

Refs #4473


marcodejongh:
Review follow-ups applied in 414c5d43a:

- **The fourth `setterClimbsFull` path** — fixed. Two of the four paths spelled the predicates out inline; both now call `visibleSetterClimbConditions`. `UNIFIED_TABLES.climbs` *is* `dbSchema.boardClimbs`, so the helper applies verbatim. Good catch: adding a predicate to three sites and missing the fourth is the exact shape of the bug this PR fixes.
- **The `params.slice` ordering assumption** — fixed. The binding check now resolves values through the `$n` placeholders the WHERE clause itself names, so a renderer that emits LIMIT/OFFSET first cannot leave it comparing the wrong values and passing. Mutation-tested: changing only the profile query's username binding (identical clause text, different value) reds it.
- **`fetchSetterIdentity` as a third roundtrip** — left serial, deliberately. It runs only after the visible-climb count proves the setter exists, so the 404 path (every made-up username a crawler tries, and every drafts-only setter) costs two queries rather than three. Batching it would move that cost onto exactly the requests we most want cheap.
- **The follow island's full profile fetch** — agreed, filed as #4582.
- **No `not-found.tsx`** — agreed, filed as #4581. Also filed #4580 for the `/setter/*` CDN TTL, which was deliberately held out of this PR.
- **`docs/og-climb.md`** — no change. That doc covers the backend-served `GET /og/climb` route; it contains no reference to `/api/og/setter`, so there is nothing there to keep current.

Follow-ups: #4580, #4581, #4582.

marcodejongh:
Second round of review follow-ups, in d59e13aef:

- **Stale follower count** — real regression against the component this replaced, fixed. The count moved into the follow island and now moves with the toggle. It stays in the crawlable HTML because a client component's first render happens on the server and the initial value is a server-resolved prop; a new assertion pins that, and reds under the mutation of gating the whole island on the follow state resolving.
- **Follow button hidden on API error** — fixed, it falls back to not-following, which is what the old component did.
- **Redundant `getBoardDetailsForBoard` for `boardListPath`** — fixed, it reuses the details the link builder already resolved. Also removes an unguarded throw on the path of a *related link*.
- **`as SQL` cast** — removed; `.where()` accepts `SQL | undefined`, which is what drizzle means.
- **`var(--color-primary-fill)` / `var(--neutral-400)`** — kept, deliberately. These are the semantic CSS custom properties `index.css` defines, not hardcoded colour values, and the sibling this page is modelled on uses the same one (`static-list-front-door.tsx`'s `introSx` is `color: 'var(--neutral-400)'`). Switching just this page would make it the odd one out.

One more change in this round that was not review feedback: a `?page` inside `FRONT_DOOR_MAX_PAGE` but past a setter's last page of climbs used to render an indexable, self-canonical empty list. Nothing links there, so across tens of thousands of setter pages it was a thin-content farm reachable by guessing. It 404s now (353cb8063).

marcodejongh:
Third round, in fa76fd1c7 — you were right that the leak had more readers than the four I fixed.

- **`searchUsersAndSetters`** — real bug, and the worst kind this PR could have created: closing the 404 while search kept offering the setter (with a count including the hidden climbs) means the tap lands on a not-found page. Fixed, with a test that asserts both halves — a drafts-only setter is absent, and a setter with one visible climb appears with `climbCount: 1` rather than the total. Mutation-tested.
- **`followSetter` existence check** — fixed, it uses the shared helper, so you cannot follow a setter whose page does not resolve.
- **OG grade bars** — fixed; the tick join now carries `is_listed`/`is_draft`, so the card counts ascents on the same climbs the page shows.

marcodejongh:
Fifth round, in 9dbdafecb + ab8999df6 — three adversarial lenses, one verdict REFUTED, the rest reproduced against the dev image before anything was changed.

**The one that mattered: the page and the climbs sitemap picked different angles.** The config tuple was already shared; the angle — also a path segment — was chosen by a second, non-matching `ORDER BY`. Replaying both verbatim over the 85,596 tier-2 climbs on the dev image: **28 disagreed**, every one an exact ascent tie. `8EE0F93116DC4FBDB3436053129C6396` ('air filter', kilter layout 8, 26 ascents at both 40 and 45) rendered `/40/view/…` while the shard submitted `/45/view/…`. Same campaign shape, fifth instance.

Both builders now read one `published-angle.ts`. Re-running the comparison after the change: **0 of 85,596**.

The setter query also had no in-range guard, so `board_climb_stats`'s 28 `grasshopper` rows at angle `-5` could win — and for one visible climb they did, producing `/grasshopper/…/-5/view/…`, a URL the shard refuses to publish with the comment "that URL 404s". Guarded with a `CASE` over the row's own `board_type`, because one setter's climbs span boards with different angle tables.

Three more, all reproduced:

- **JSON-LD named page 1 on every page.** `?page=2` self-canonicalised to `?page=2` and carried positions 51–100 while the graph said `/setter/{name}`. It now builds its document URL from the same `frontDoorPagePath` the canonical uses.
- **22,490 of the 91,946 setters who answer 200 have no climb on any resolvable configuration** (24.5%, dev image) — an indexable `<h1>` over fifty board images and zero anchors. Those pages go `noindex, follow`. Incidentally: the real `resolveClimbSitemapGroups` yields **9** groups on this image, not the 12 an earlier approximation reported, so the submitted-setter count in #4579 was corrected too.
- **A second `decodeURIComponent` on the route param.** Next already decodes dynamic segments, so `/setter/50%25` threw an unhandled `URIError` out of both `generateMetadata` and the page body — a 500 where a 404 belongs — and `abc%2541` was silently rewritten to `abcA`. Pre-existing on `main`, but this PR made the decode load-bearing for the new 404.

**The oracle that was supposed to catch the angle bug could not.** `setter-climb-links.test.ts` feeds the same fixture `angle` to both the rendered href and the sitemap href, so the angle is an input to both sides there. The new oracle byte-compares the angle-selection `ORDER BY` extracted from the SQL **both real builders render**, with only the correlated subquery's alias normalised — nothing about the expected ordering is written down in the test. Eight mutations applied, each watched go red and reverted; the table is in the PR body.

Also took your non-blocking note on `SetterFollowIsland`: it fetched the whole profile to read one boolean and threw away the `followerCount` in the same response. It now adopts the live count, with a test pinning that the server prop is still what renders first.

One deliberate behaviour change outside the setter tree: the climbs shard's angle ordering renders as `desc nulls last` rather than a bare `desc`. It cannot change a row (`ascensionist_count >= 10` excludes NULL there) and stating it once is what makes the two builders byte-comparable.

Numbers are dev image; production reconciliation is outstanding and no production database was read.

marcodejongh:

Every dev-image figure in the body re-derived against production (PostgreSQL 16.9, `board_climbs` 885,528 rows vs the dev image's 648,657). Aggregates only; the linkable predicate is the one **this branch's** `resolveClimbSitemapGroups` resolves, not the issue's.

### The counts

|---|---|---|---|
| distinct setters with ≥1 climb of any kind | 104,140 | **137,348** | +33,208 |
| …with ≥1 `is_listed AND NOT is_draft` climb | 91,946 | **125,082** | +33,136 |
| …with ≥3 visible climbs | 35,950 | **49,367** | +13,417 |
| drafts/unlisted-only — **flip 200 → 404** | 12,194 | **12,266** | **+72** |
| visible setters whose climbs are all MoonBoard (render unlinked) | 21,154 | **38,575** | +17,421 |
| largest setter, visible climbs | 4,749 | **5,065** | +316 |
| resolvable `(board_type, layout_id)` groups | 9 | **9** | 0 |
| **visible setters with no climb on a resolvable config → `noindex, follow`** | 22,490 (24.5%) | **41,844 (33.5%)** | +19,354 |
| tier-2 climbs compared for the angle replay | 85,596 | 53,137 | — |
| angle disagreements, pre-fix | 28 | **45** | +17 |
| visible+linkable climbs the old query would link at an off-table angle | 1 | **4** | +3 |
| `board_climb_stats` rows at `grasshopper` angle `-5` | 28 | **47** | +19 |
| **usernames containing `%`** | **0** | **1** | **+1** |
| largest setter page, warm (51-row page / board-type query) | ~95 ms / ~16 ms (local) | 662–682 ms / 540–559 ms (off-cluster) | see note |

### Two things that change a conclusion

**1. The `%` username is not latent in production — it is live.** The body says the second `decodeURIComponent` is "latent on this data (0 of the 91,946 dev-image usernames contain `%`)". Production has **one** setter username containing `%`, and that setter has ≥1 visible climb, so their page answers 200 today. On `main`, `/setter/…%25…` throws `URIError: URI malformed` out of **both** `generateMetadata` and the page body before either `try` — a 500 on a live indexable URL. This PR deletes both decodes, so it is **fixing a reachable production 500, not removing a hypothetical**. Worth promoting out of the review-round list. (That setter has fewer than 3 linkable visible climbs, so they are not in #4579's submitted set — the exposure is page-level only.)

**2. The lonely-page `noindex` admits 1.86× what the body claims.** 41,844 of the 125,082 setters that now answer 200 — **33.5%, not 24.5%** — have no visible climb on any resolvable configuration, so they render an `<h1>` over fifty board images with not one crawlable anchor and get `noindex, follow`. Same conclusion, considerably more volume: a third of the surface this PR makes 200-able is deliberately kept out of the index. That is the right call and the gate does its job; the number in the body just needs to be a third larger.

### Two things that hold up well

- **The 404 blast radius barely moved: 12,266 vs 12,194 (+0.6%).** This is the one figure where the dev image is a faithful proxy for production. The "12,194 setters lose their page, none of them indexed" argument survives intact.
- **The angle fix is worth more in production, not less.** 45 tier-2 climbs (not 28) were being linked by the setter page at an angle the climbs shard never submits, and **4** visible+linkable climbs (not 1) would have been linked at an angle no route table carries — a hard 404 from the page's own anchor. `publishableAngleWhere` earns its `CASE`.

### Note on the query timings

662–682 ms and 540–559 ms are measured from a dev box over Railway's public TCP proxy with a **~260 ms round-trip floor** (five `SELECT 1` probes: 259/259/260/262 ms). A Vercel lambda sits far closer to the database, so the honest reading is "roughly 300–400 ms of database work for the largest setter in production against ~95 ms on the dev image", not a 7× regression. Both queries are `LIMIT`-bounded and neither is near a budget.

### SQL

The nine production groups `resolveClimbSitemapGroups` resolves are `decoy/2` size 1 sets `{2…20}` · `grasshopper/1` size 2 `{1,2,3}` · `kilter/1` size 10 `{1,20}` · `kilter/8` size 25 `{26,27,28,29}` · `soill/1` size 1 `{1}` · `tension/9` size 1 `{8,9,10,11}` · `tension/10` size 6 `{12,13}` · `tension/11` size 6 `{12,13}` · `touchstone/1` size 1 `{1}`. Call that OR-list `<LINKABLE>`:

```sql
-- <LINKABLE>, one disjunct per group, the same three predicates the shard applies
  (board_type = 'decoy'       AND layout_id = 2  AND compatible_size_ids @> ARRAY[1]::int[]
     AND required_set_ids <@ ARRAY[2,3,4,5,6,7,8,9,10,11,12,13,14,15,16,17,18,19,20]::int[])
  OR (board_type = 'grasshopper' AND layout_id = 1 AND compatible_size_ids @> ARRAY[2]::int[]
     AND required_set_ids <@ ARRAY[1,2,3]::int[])
  OR (board_type = 'kilter'    AND layout_id = 1  AND compatible_size_ids @> ARRAY[10]::int[]
     AND required_set_ids <@ ARRAY[1,20]::int[])
  OR (board_type = 'kilter'    AND layout_id = 8  AND compatible_size_ids @> ARRAY[25]::int[]
     AND required_set_ids <@ ARRAY[26,27,28,29]::int[])
  OR (board_type = 'soill'     AND layout_id = 1  AND compatible_size_ids @> ARRAY[1]::int[]
     AND required_set_ids <@ ARRAY[1]::int[])
  OR (board_type = 'tension'   AND layout_id = 9  AND compatible_size_ids @> ARRAY[1]::int[]
     AND required_set_ids <@ ARRAY[8,9,10,11]::int[])
  OR (board_type = 'tension'   AND layout_id = 10 AND compatible_size_ids @> ARRAY[6]::int[]
     AND required_set_ids <@ ARRAY[12,13]::int[])
  OR (board_type = 'tension'   AND layout_id = 11 AND compatible_size_ids @> ARRAY[6]::int[]
     AND required_set_ids <@ ARRAY[12,13]::int[])
  OR (board_type = 'touchstone' AND layout_id = 1 AND compatible_size_ids @> ARRAY[1]::int[]
     AND required_set_ids <@ ARRAY[1]::int[])
```

```sql
-- every setter count in the table above, one grouped scan (3.5 s on production)
WITH per_setter AS (
  SELECT setter_username,
         count(*)::int                                                       AS all_climbs,
         count(*) FILTER (WHERE is_listed AND NOT is_draft)::int             AS visible,
         count(*) FILTER (WHERE is_listed AND NOT is_draft AND (<LINKABLE>))::int AS linkable_visible,
         count(*) FILTER (WHERE is_listed AND NOT is_draft
                            AND board_type = 'moonboard')::int               AS visible_moonboard
  FROM board_climbs
  WHERE setter_username IS NOT NULL AND setter_username <> ''
  GROUP BY setter_username
)
SELECT count(*)::int                                                          AS setters_any_climb,
       count(*) FILTER (WHERE visible >= 1)::int                              AS setters_visible_1,
       count(*) FILTER (WHERE visible >= 3)::int                              AS setters_visible_3,
       count(*) FILTER (WHERE visible = 0)::int                               AS drafts_unlisted_only,
       count(*) FILTER (WHERE visible >= 1 AND visible = visible_moonboard)::int AS visible_all_moonboard,
       count(*) FILTER (WHERE visible >= 1 AND linkable_visible = 0)::int     AS visible_no_linkable,
       max(visible)::int                                                      AS largest_visible,
       count(*) FILTER (WHERE setter_username LIKE '%\%%')::int               AS usernames_with_percent
FROM per_setter;
```

```sql
-- angle replay: the sitemap's published-angle rule vs the pre-#4577 setter ordering
WITH visible AS (
  SELECT uuid, board_type, angle AS climb_angle FROM board_climbs
  WHERE is_listed = true AND is_draft = false AND (<LINKABLE>)
),
sitemap_pick AS (
  SELECT DISTINCT ON (s.climb_uuid, s.board_type) s.climb_uuid, s.board_type, s.angle
  FROM board_climb_stats s JOIN visible v ON v.uuid = s.climb_uuid AND v.board_type = s.board_type
  WHERE s.ascensionist_count >= 10
    AND s.angle IN (0,5,10,15,20,25,30,35,40,45,50,55,60,65,70)
  ORDER BY s.climb_uuid, s.board_type, s.ascensionist_count DESC NULLS LAST,
           COALESCE(s.angle = v.climb_angle, false) DESC, s.angle ASC
),
old_setter_pick AS (
  SELECT DISTINCT ON (s.climb_uuid, s.board_type) s.climb_uuid, s.board_type, s.angle
  FROM board_climb_stats s JOIN visible v ON v.uuid = s.climb_uuid AND v.board_type = s.board_type
  ORDER BY s.climb_uuid, s.board_type, s.ascensionist_count DESC NULLS LAST, s.angle ASC
)
SELECT count(*)::int AS tier2_compared,
       count(*) FILTER (WHERE o.angle IS DISTINCT FROM p.angle)::int AS disagreements
FROM sitemap_pick p
LEFT JOIN old_setter_pick o USING (climb_uuid, board_type);
-- → 53,137 compared, 45 disagreements
```

```sql
-- off-table-angle exposure (the grasshopper -5 case)
SELECT count(*)::int FROM board_climb_stats
WHERE board_type = 'grasshopper' AND angle = -5;                        -- 47
SELECT count(DISTINCT climb_uuid)::int FROM board_climb_stats
WHERE angle NOT IN (0,5,10,15,20,25,30,35,40,45,50,55,60,65,70);        -- 47 (the only off-table rows in prod)
```

The page-query timings came from a throwaway vitest file in this worktree calling the branch's own `buildSetterProfileQuery` / `buildSetterClimbsQuery` against a read-only production connection, `server-only` and `next/cache` stubbed. The probe has been deleted; the worktree is clean.

### Not a blocker

No cap, budget or page count in this PR moves — everything above is a page-level count. The two edits worth making to the body are the `%`-username status (latent → live) and the lonely-page share (24.5% → 33.5%). The sitemap-side numbers are reconciled separately on #4579.

marcodejongh:

Conflicting: packages/web/app/api/og/setter/route.tsx.

marcodejongh:
Three of four addressed in `d5c3ef8c`; the French one is declined, with a reason.

**2. `conditions` mutation — fixed.** Spreads now, matching `setterClimbsFull` in the same file. You're right that it works only while the helper keeps returning a fresh array, which is exactly the assumption a future memoisation would break silently.

**3. `docs/sitemap.md` — added.** New "One rule picks the published angle" section: `published-angle.ts` as the single rule, both builders reading it, the 28-climb disagreement it closed, and the two parts that are easy to get wrong (`getRoutableBoardAngles` rather than the `ANGLES` picker; `nulls last` on the ascent count, load-bearing on the setter side and a no-op on the shard side). #4579 needed this in place.

**4. `notFound()` in `generateMetadata` — taken.** You're right that the branch was dead: the body 404s on the identical condition, so the metadata was always discarded. Both call `notFound()` now. The test keeps the invariant that actually matters — zero setter-page queries for a crawler walking `?page=50000` — and asserts the throw instead of the discarded object.

**1. French `la board` — declining, the string is correct.** `docs/i18n-french-glossary.md` has a section headed **"The board device is « la board » (feminine)"**:

> Every string names the board device « board », feminine: _la board, une board, les boards_, with feminine agreement (_la board connectée_, _une board trouvée_). That is what French board climbers say and how French climbing media writes it (La Fabrique Verticale: « la board », « une Kilterboard ») — it patterns with « la planche », like skate and snowboard French.

So `de la board` is the house rule, not an error, and `du board` would be the regression. The general instinct about borrowed English nouns is reasonable — it just loses to a documented, deliberate exception here. Worth checking the glossary before flagging gender on `board` again; CLAUDE.md points at it for exactly this.

marcodejongh:
Good catch — taken in `d3bf60cc`.

`followSetter` did carry the guard and nothing exercised it, so a regression there would only have shown up in production. Two cases now, with a **paired oracle** on the refusal: a handler that threw *after* inserting would pass on the rejection alone, so the follow-row count is asserted too.

**Mutation:** swap `visibleSetterClimbConditions` for a bare `setter_username` match in `followSetter` → red on the drafts-only case, the other four green. Reverted.

One incidental, worth knowing for the next test in that file: the `makeCtx` helper the playlist tests use (`{ sessionId: null, boardPath: null, … } as ConnectionContext`) does not typecheck here — `ConnectionContext` has no `boardPath`, and its optional fields are `string | undefined`, not nullable. Every field but `connectionId` is optional, so the helper needs no cast at all; not casting is what keeps it from drifting out of shape.

Thanks for confirming the French string against the glossary on this pass.

marcodejongh:
Applying the two body edits from your 2026-08-19 production reconciliation — they were still outstanding.

**1. The `%` username is live, not latent.** The body called the second `decodeURIComponent` "latent on this data (0 of the 91,946 dev-image usernames contain `%`)". Production has one such setter, with ≥1 visible climb, so their page answers 200 today and `/setter/…%25…` throws `URIError: URI malformed` out of both `generateMetadata` and the page body on `main` right now. Rewritten to say this PR fixes a reachable production 500, with the dev-image zero kept as the reason the earlier pass misread it, and your note that the setter has fewer than 3 linkable climbs so the exposure is page-level only.

**2. Lonely-page share: 24.5% → 33.5%.** Now reads **41,844 of the 125,082 setters who now answer 200**, with the dev-image figures (22,490 of 91,946) kept as the contrast and the 1.86× named.

No code moved — both are page-level counts, and as you noted no cap, budget or page count in this PR changes. The sitemap-side numbers stay reconciled on #4579.

marcodejongh:
Two of three taken, in `50e8be8d`.

**German `am {{boards}}` — real bug, fixed.** `seoSummary` interpolates `{{boards}}` from `boardTypes.map(…).join(', ')`, so it can carry several names, and `am` (= `an dem`) asserts exactly one. "am Kilter, Tension" is wrong on the count and reads oddly after a contraction. Now `an {{boards}}`, correct for both numbers because the interpolated values are bare proper nouns that take no article. Worth noting the glossary's `das Board` / `am Board` rule is untouched — it governs the device noun, which this string never uses. Spanish (`en`) and French (`sur`) were already number-neutral; German was the only locale that contracted.

**OG 404 — real gap, covered.** The existing three cases pin the replica seam, the read deadline and the DB-rejection fallback, but not the null-summary 404. Added, with a **paired oracle**: a handler that 404s *after* rendering would pass on the status alone, and rendering is the expensive half, so the captured element is asserted null too. Mutation: neuter the `!summary` guard → 2 red, including the new case.

**French `la board` — declining, third time.** You confirmed this yourself two comments ago ("lowercase feminine `la board` is correct per the French glossary"), then re-raised it. `docs/i18n-french-glossary.md` has a section headed **"The board device is « la board » (feminine)"**; `de la board` is the documented house rule and `du board` would be the regression. Nothing about the string changed between those two reviews. Please check the glossary before flagging gender on `board` — CLAUDE.md points at it for exactly this.

marcodejongh:
This review landed a couple of minutes after the merge, so the three substantive points had nowhere to go — carried to **#5183** rather than lost in a merged thread: the two overlapping cold round trips in `generateMetadata`, the two "does this setter exist" branches reading different queries, and the follow-island error fallback (recorded as the documented trade-off it is, with the alternative noted so it stays revisitable).

**Point 4 is stale.** The PR body said backend Vitest was not run locally; that was true when written, but the run happened before the final push — `setter-visibility.test.ts` 5/5 green, including the mutation check on the new `followSetter` guard (swap `visibleSetterClimbConditions` for a bare `setter_username` match → red on the drafts-only case). CI was not the only gate.

#4579 is rebased onto post-merge `main` and green locally.

PR #4578, opened 2026-08-19
feat(web): put MoonBoard in the sitemaps

### LINKED ISSUE #4493: MoonBoard is in no sitemap: 73,412 tier-2 climbs (58% of the set) have no popular board config

marcodejongh:
## Summary

Closes #4493. **#4576 merged as `c283806` while this was in review, so the stack is resolved and this now targets `main` directly.** That ordering was load-bearing: dropping the sitemap's MoonBoard hold-out ahead of the parser and uuid fixes would have submitted tens of thousands of URLs that 404 or canonicalise elsewhere. It is no longer a risk — the parser fix is live.

**The issue's diagnosis is wrong.** MoonBoard is not ranked out of `popularBoardConfigs`. That resolver builds its universe with a `GROUP BY` over `board_product_sizes_layouts_sets`; `user_boards` only supplies `board_count` for the `ORDER BY`, and nothing is dropped for being unpopular. There is **no psls row for MoonBoard at all** — the only writers are the Aurora importers (`import-aurora-board-unified.ts`, `packages/aurora-sync`) and no migration seeds one. So the issue's recommended option 2, "rank by climbCount instead of boardCount", has nothing to rank.

The fix is option 1, and it is far cheaper than the issue estimates: `getDefaultRenderBoard('moonboard', layoutId)` already returns exactly the `{ layoutId, sizeId, setIds }` tuple, pure and synchronous, and web already imports it.

## Rebased onto the materialised store

This branch was written against the old shard, where every page built the whole ordered list live. **#4661 landed underneath it** and moved the ground: the climbs shard now reads `sitemap_climb_urls`, written by one refresher, and `/sitemap.xml` reads one row of `sitemap_shard_refreshes`. Two consequences, both acted on rather than carried:

- **The refresher is now the only door into the sitemap.** A synthetic config that reaches `/sitemaps/boards.xml` and the live summary but not `buildAllTier2UrlRows()` produces a store that has never heard of MoonBoard and a shard that emits nothing new — a change that looks complete in review and ships a no-op. `buildAllTier2UrlRows` resolves its groups from `getSitemapClimbConfigsOrThrow` now. That one call is the substantive edit in this PR, and there is no second place to make it: the cron, the `after()` self-heal and the live fallback all funnel through that function, deliberately (#4661 kept one selection path so a drifting duplicate could not recur).
- **`SUMMARY_CONCURRENCY = 3` is gone.** It existed to bring the LIVE summary back inside `SHARD_DEADLINE_MS`. `/sitemap.xml` is not on that path any more, so the deadline it raced is no longer on the far side of it. What is left is the pre-first-refresh fallback, where the answer is to refresh the store — not to spend three of a ten-connection pool on a path that should run once per deploy. The group loop is sequential again and `climb-item-cache.test.ts` pins `maxInFlight === 1`, matching main. Measured cost of that decision is in "The empty-store window" below; it is one degraded request, self-healed.

**What else changed**

- New `packages/web/app/lib/seo/sitemap/board-config-source.ts`: the listed configs, plus one synthetic `PopularBoardConfig` per MoonBoard layout. The two questions get separate entry points instead of one query with a fallback bolted on — "which board pages are worth building" stays a user-boards popularity question in the backend, "which climbs are worth indexing" becomes a catalogue-coverage question in the sitemap layer.
- An honest `climbCount` from one grouped Drizzle count over `board_climbs` (listed, non-draft) per layout, behind the same Data Cache + in-process TTL + single-flight shape `getAllBoardConfigsOrThrow` uses, and behind a 10 s budget. Deliberately **not** `buildTier2ClimbSummaryQuery`: both shards read this number only as a `> 0` gate. A layout with no listed climbs is dropped rather than shipped as a thin `/list` page.
- `hasRoundTrippableSetSlug` and its call are deleted. It existed because web's parser disagreed with the emitter — #4576 makes them agree. The `tryConstructSlugViewUrl` probe stays and is still what drops **Kilter layout 5 ("Kilter Spire")**: an orphaned layout with no size associations, so `resolveReadableBoardSegments` returns null at every tuple. 69 tier-2 climbs in production, documented not fixed.
- **The boards shard no longer 503s when the MoonBoard count fails.** The strict and tolerant sources are separate functions.
- Both shards repointed, and the `vi.mock` sites moved to the module the code now reads.

**Production target:** `www.boardsesh.com` only. Now that #4576 has merged and this is rebased onto `main`, the diff is 14 files and every one of them is under `packages/web/app/lib/seo/sitemap/` or `docs/` — nothing in `packages/mobile`, `packages/shared` or `packages/backend`, so no OTA and no store fleet. (The `pr-4578` OTA preview channel published earlier in this PR's life, while the branch still carried #4576's `packages/shared/play-view` commit; it does not any more.) No migration, no new redirect (so no `es`/`fr`/`de` sibling work), no user-facing strings.

---

### Production, measured today (read-only `SELECT`, aggregates only)

Live right now, before this PR:

| surface | value |
|---|---|
| `/sitemap.xml` | 9 `<sitemap>` entries, **no `x-sitemap-degraded`**, 0.21 s |
| climbs shard | **6 pages, 53,230 URLs**, zero `/moonboard/` |
| `climbs/1.xml` | 200 in **0.11 s** (2,247,529 B) |
| `climbs/6.xml` | 200 in 4.63 s cold, 3,230 URLs |
| `boards.xml` | 2,760 URLs (690 items × 4 locales), zero `/moonboard/` |

MoonBoard tier-2 in production, counted with the shard's own predicate — `DISTINCT ON (climb_uuid)`, `ascensionist_count >= 10`, `angle IN (25, 40)`, the NULL-tolerant set containment, the genuine-alias anti-join:

| layout | tier-2 | listed non-draft |
|---|---|---|
| 1 · MoonBoard 2010 | 7 | 124 |
| 2 · MoonBoard 2016 | 21,973 | 92,677 |
| 3 · MoonBoard 2024 | 6,419 | 36,135 |
| 4 · Masters 2017 | 25,551 | 73,244 |
| 5 · Masters 2019 | 13,990 | 56,864 |
| 6 · Mini 2020 | 3,779 | 7,867 |
| 7 · Mini 2025 | 1,693 | 4,870 |
| **total** | **73,412** | **271,781** |

All seven layouts clear the `> 0` gate in production, so all seven synthetic configs ship — unlike the dev image, where only four do.

Projected:

| | items | pages | boards.xml |
|---|---|---|---|
| before | 53,230 | **6** | 690 items / 2,760 URLs |
| after | **126,642** | **13** | 704 items / 2,816 URLs |

Total tier-2 across all 17 groups is 127,555 today; the 913 between that and 126,642 is Kilter layout 5 (69) plus the angle and size/set predicates the shard applies and a raw group count does not.

### Dev image, measured end to end through the real code path

`vp run dev` against the full dev image (648k climbs), store truncated first, then refreshed through `/api/internal/refresh-sitemap-climbs?force=1`.

| | items | groups | pages |
|---|---|---|---|
| before | 41,336 | 9 | **5** |
| after | **85,347** | 13 | **9** |

The before row is derived from the after run itself — the store carries `board_type`, so the non-MoonBoard rows are exactly 41,336 — and it matches the 41,336 #4661 measured independently on the same image.

Per MoonBoard group, straight out of `sitemap_climb_urls`:

| board | layout | URLs | ordinals |
|---|---|---|---|
| moonboard | 2 (2016) | 15,907 | 26,304–42,210 |
| moonboard | 4 (Masters 2017) | 19,338 | 42,211–61,548 |
| moonboard | 5 (Masters 2019) | 6,825 | 61,549–68,373 |
| moonboard | 6 (Mini 2020) | 1,941 | 68,374–70,314 |

Layouts 1, 3 and 7 have zero listed climbs on this image and are dropped by the gate. Production has all seven.

**Shard read time, from the store:**

| | |
|---|---|
| `/sitemap.xml` | 0.086 / 0.098 / 0.105 s, **no `x-sitemap-degraded`**, 9 climbs pages, 6 distinct per-page `<lastmod>` |
| `climbs/1.xml` (10,000 URLs, 2,329,686 B) | 0.073 s warm, 1.10 s on the first hit after a route compile |
| `climbs/5.xml` (10,000 URLs, all MoonBoard) | 0.10–0.16 s |
| `climbs/9.xml` (5,347 URLs) | 0.084 s |
| `climbs/0.xml`, `climbs/10.xml` | 404, not 503 |

**Refresh cost**, for 13 groups, all through the real cron route on the same image. Stated as a range, because quoting only the warm end makes the budget look far safer than it is:

| database state | `scanDurationMs` |
|---|---|
| **cold** — Postgres container restarted, store truncated | **64,937 ms** |
| warm, repeat runs on a loaded box | 19,998 – 28,421 ms |
| warm, quiet box | 8,921 – 12,217 ms |
| `after()` self-heal on a partly-cold image | 27,870 ms |

Budget against the cold number. The route's `maxDuration` is 300 s; scaling 64.9 s by production's projected 126,642 rows over the dev image's 85,347 gives roughly **96 s**, so the real margin is about **3x**, not the 24x the warm figure implies. Production also adds three more MoonBoard groups (20 total). It is a background cron, not a request path — but the same builder is the shard page's empty-store fallback, which is a request path, so `/sitemaps/climbs/[page]` now exports `maxDuration = 300` like its two siblings. Without it a crawler that reaches a page URL before the first refresh takes a platform 504 instead of the slow 200 the fallback is designed to be (measured: 22.7 s for that first hit against a warm DB, dev route compile included, then 0.40 s and 0.48 s — the fallback build is TTL-cached per instance).

**Largest served page is 2,935,185 B**, 73% of `MAX_SHARD_BYTES`. The worst *possible* page is a bit past that, and it is a MoonBoard one: MoonBoard set slugs are much longer than Kilter's, so a full 10,000-URL page of Masters 2019 (`wooden-holds-c_wooden-holds-b_wooden-holds_screw_original-school-holds_hold-set-b_hold-set-a`, 92 characters) renders **303 B/URL, 3.03 MB, 76%** against Kilter original's 203 B/URL and 2.03 MB — measured through `climbRowsToItems` + `renderUrlset` with a 15-character climb name. Headroom is real but it is one long set list wide: on a full page each extra path character costs 10,000 bytes, so Masters 2019 sits ~97 characters short of the 4 MB stop. `CLIMB_URLS_PER_SHARD`'s comment said "around 2.5 MB" from Kilter lengths; it now carries the MoonBoard figures.

Adding MoonBoard also moves page boundaries for existing URLs, because `resolveClimbSitemapGroups` orders by board type and MoonBoard sorts into the middle rather than onto the end. Measured on the refreshed dev store — moonboard holds ordinals 26,304–70,314, so **15,032 Soill/Tension/Touchstone URLs moved from pages 3–5 onto pages 8–9**. That is a catalogue change, which is exactly what `ordinal` is allowed to move for; each page is self-contained and its `<lastmod>` is recomputed from the rows it actually holds.

An earlier revision of this branch claimed the opposite in three places — that `boardCount: 0` "keeps them last under `isBetterConfig`'s ordering, so adding them cannot reorder or displace any Aurora group". That is false on both clauses: `isBetterConfig` is only called inside the per-group `best` map in `resolveClimbSitemapGroups` and never compares across board types, and the cross-group order is a plain lexicographic `boardType` compare where `moonboard` lands between `kilter` and `soill`. Corrected in `board-config-source.ts`, `board-config-source.test.ts` and `docs/sitemap.md` (which was contradicting itself between two sections). Nothing about the shipped ordering changed — only the description of it.

### The empty-store window, stated rather than papered over

This is the one place dropping `SUMMARY_CONCURRENCY` costs something. With the store truncated, the very first `/sitemap.xml` takes the live fallback:

```
/sitemap.xml  200 in 4.30 s   x-sitemap-degraded: climbs   (0 climbs pages listed)
```

Then the `after()` self-heal #4661 built fires post-flush, and every request after it is healthy:

```
[sitemap] self-healed the climb sitemap store (empty): 85347 items in 27870ms, skipped=no
/sitemap.xml  200 in 0.105 s   healthy   9 climbs pages
/sitemap.xml  200 in 0.098 s   healthy
/sitemap.xml  200 in 0.086 s   healthy
```

So the window is **one request per empty store**, under a 60-second degraded cache header, and it closes without anyone running a curl. Three lanes of fan-out would have made that one request ~2 s instead of ~4.3 s and still degraded, because thirteen groups cold do not fit in 3 s at any lane count — which is precisely why #4661 moved the index off this path instead of tuning it. Adding MoonBoard widens a window that already existed and already self-closes; it does not open a new one.

### The composition, proved end to end

A URL pulled out of `/sitemaps/climbs/5.xml` — that is, out of the materialised store — fetched in the browser:

```
/moonboard/2016/standard-11x18-grid/original-school-holds_hold-set-b_hold-set-a/40/view/can-you-do-it-dcb417f0-225f-59a6-8758-60d123532dc5
  200 · <h1>Can you do it? — 7c+/V10</h1>
  <link rel="canonical"> byte-identical to the submitted path
```

And the two layouts #4576 exists for, both from the shard, both with the `screw` part the old parser could never produce:

```
/moonboard/masters-2017/standard-11x18-grid/wooden-holds_screw_original-school-holds_hold-set-c_hold-set-b_hold-set-a/40/view/for-finn-663d3df8-…
  200 · canonical byte-identical
/moonboard/masters-2019/standard-11x18-grid/wooden-holds-c_wooden-holds-b_wooden-holds_screw_original-school-holds_hold-set-b_hold-set-a/40/view/ganan-las-negras-00087864-…
  200 · canonical byte-identical
```

`boards.xml` on the same run: 668 items / 2,672 URLs, 8 MoonBoard `/list` URLs (four layouts × two angles), all 660 Kilter/Tension/Decoy items still present, zero `//` and zero `?`.

## Release Notes

- MoonBoard climbs can now be found on Google. Every MoonBoard layout with problems on it goes into the sitemap, so a search for a problem by name can land on it directly instead of only ever reaching Kilter and Tension.

## Test plan

1. Search Google for a MoonBoard problem by name.
2. Open a MoonBoard climb link → the right board loads.
3. Copy that page's URL, open it fresh → same board, same angle.
4. Open a Kilter climb link → still works as before.

## Risk

Risk: 3/5 — this submits tens of thousands of new URLs to Google, and a URL that
does not round-trip to its own page's canonical is worse than no URL: Google
records it as "alternate page with proper canonical" and the whole batch stops
earning anything. That is the failure the hold-out existed to prevent, and it is
now covered by `moonboard-canonical-identity.test.ts`, which composes the real
emitter, the real route parser and the real canonical builder over all seven
layouts. Reversible in one deploy — the shard is derived, nothing is written
that a revert would not undo, and `CLIMB_SITEMAPS_ENABLED` withdraws the whole
climb surface without a code change.

## Rebased onto main, 2026-09-05

The branch was ~795 commits behind, and the sitemap infrastructure moved under
it. Three files conflicted; the more important half is that **every number this
branch argued from was measured against a guard that no longer exists.**

- **`shard-registry.ts`** — main's `boards` entry gained
  `expansion: 'default-locale-only'` (#4648) with a comment. Kept both and took
  only this branch's `build:` source. Taking the branch's side wholesale would
  have dropped that line and reddened
  `app/__tests__/board-content-metadata-guard.test.ts`, which reads
  `shard-registry.ts` **as text** and greps the `id: 'boards'` block for it — no
  type error to warn you first.
- **`sitemap-xml.ts`** — main rewrote the same block from scratch. Took main's
  argument whole and grafted on the one thing this branch adds: the per-URL cost
  is not one number, the **set slug** spreads it, and MoonBoard sits at the top.
- **`docs/sitemap.md`** — kept main's climb-sitemap-switch section, added the
  config-source section after it, and let main's byte-budget rewrite win
  outright. This branch's version still said "fixed shards have no byte budget",
  which stopped being true at #4648.

**Byte figures re-measured, and now checked rather than asserted.** The branch
sized MoonBoard's pages against a 4 MB Vercel response ceiling. That ceiling is
gone: `MAX_SHARD_BYTES` is a 45 MB protocol backstop and the working guard is
`pagedShardByteBudget()` at 500 B/URL, so a climbs page is checked against 5 MB.
Rather than restate the arithmetic, `moonboard-canonical-identity.test.ts` now
renders a full Masters 2019 page through `climbRowsToItems` + `renderUrlset` and
reads per-URL cost as a **delta between two page sizes**, so the urlset preamble
cancels out instead of being amortised into the answer.

| | measured |
|---|---|
| MoonBoard Masters 2019 (92-char set slug) | **297 B/URL** |
| a full 10,000-URL page | **2.97 MB** |
| `pagedShardByteBudget(10_000)` | 5.00 MB — **59% used** |
| headroom | **203 characters** of set slug or climb name |

That corrects two claims on `main` too. `PAGED_SHARD_BYTES_PER_URL`'s 500 was
documented as "double the measured cost"; with MoonBoard in the shard it is
**1.68x**. And `climb-shards.test.ts` asserted the budget clears "~250 bytes/URL
the pages actually cost" — true while the shard was Kilter and Tension only, and
it would have gone on passing while proving less than it says. Now 297.

**Docs that had gone false.** `docs/web-reposition.md` still carried the hold-out
rationale as live ("212 of 227 layout×set combinations mismatch") although #4576
fixed the parser it described, and still explained the climb shards'
default-locale-only expansion as an inconsistency with a boards shard that has
since moved to the same expansion. Both rewritten.

**Owed, and flagged rather than guessed:** the refresh-cost and empty-store
figures in `docs/sitemap.md` ("thirteen groups", "85,347 rows") were measured
before `main` widened the tier-2 angle predicate from the picker set to the
routable set, so the counts need a re-measure on a dev image. The shape of the
degrade is unchanged and the text now says which numbers are pre-rebase. No
in-repo dev database was available on the box this rebase ran on.

## Author verification

Every oracle was mutation-tested: mutation applied, test watched go red, reverted, `git status --porcelain` verified clean afterwards.

**`__tests__/moonboard-reaches-the-store.test.ts`** (new, 2 tests) — the seam this rebase is about, as a composition rather than three mocks agreeing: real `resolveClimbSitemapGroups`, real `climbRowsToItems`, a masters-2017 config with **literal** tuples, and a dashed RFC-4122 uuid, asserting the emitted path byte for byte including `screw`. A negative arm pins that the MoonBoard URLs came from the config source and not from an unconditional MoonBoard branch inside the builder.

- Mutation — both call sites back to `getAllBoardConfigsOrThrow`: **11 red across 2 files**.
- Mutation — **only** `buildAllTier2UrlRows` back, leaving the summary and the boards shard on the new source. This is the half-done edit that looks complete in review, keeps `/sitemaps/boards.xml` full of MoonBoard `/list` URLs, and ships a store with no MoonBoard in it: **7 red**, including both new oracles.

**`__tests__/board-config-source.test.ts`** (16 tests) — the listed configs are mocked to production's MoonBoard-free shape and the count query is stubbed per layout. Set tuples asserted as **literals** (`{ layoutId: 4, sizeId: 1, setIds: [11,12,13,14,15,16] }`), never re-derived by calling `getDefaultRenderBoard`, which would assert `f(x) === f(x)`.
  - The **input** layout list is derived from `MOONBOARD_LAYOUTS` rather than a hardcoded `[1..7]`. With the literal list, a new catalogue layout got no count row, resolved to `climbCount: 0`, and was dropped before `toHaveLength(7)` could see it — so this file stayed green on exactly the half-done edit the source comment claims it catches. Mutation: add a `moonboard-2027` layout + set entry → **red**.
  - The rendered-SQL oracle asserted `toContain(true)` / `toContain(false)`, which is position-blind: swapping `isListed`/`isDraft` renders byte-identical SQL and a params array that still contains both, so the query would have matched **zero** listed MoonBoard climbs and shipped a no-op — with `expectsUrls` never firing because Kilter and Tension keep both shards non-empty. Now `toEqual(['moonboard', true, false])`. Mutation: swap the booleans → **red**.
  - The boards shard serves the listed configs when the count fails; still throws when the **listed** fetch fails; the climbs shard still throws on a count failure. Mutation: make the boards leg rethrow → **2 red** (and the route probe goes 503).
  - A stalled read is given up on at 10 s, on both paths. Mutation: remove the timeout wrapper → **2 red**, each hanging 5 s first.
  - A rejected count is not memoised. Mutations: memoise the rejection, and drop the single-flight → **red** each.
  - `listed.calls` pins one listed-config call per request. Mutation: split the `Promise.all` into two sequential calls → **red**.

**`__tests__/climb-entries.test.ts`** — the `MoonBoard hold-out` block asserts the opposite now: `isResolvableGroup` is true for the Masters 2017 full-set group, `resolveClimbSitemapGroups` returns both MoonBoard layouts it is given, the Aurora assertions are unchanged, **and** a bogus MoonBoard group (`layoutId: 999`) is still dropped. Mutation: replace the deleted hold-out with a blanket `return true` → **3 red**.

**`__tests__/shard-registry.test.ts`** — builds the `boards` shard with a Kilter and a MoonBoard config, asserts exactly two `/moonboard/` paths, no `//`, no `?`, Kilter still present. Mutation: leave the boards shard on the old source and wire only the climb shards → **10 red across 3 files**.

**`__tests__/climb-item-cache.test.ts`** — back to main's three-group fixture and `maxInFlight === 1` for both the item build and the summary, since the fan-out is gone.

Gates, post-rebase on this box:

- `packages/web/app/lib/seo` + `app/sitemaps` + `app/sitemap.xml` +
  `board-content-metadata-guard` — **26 files, 290 tests green**, including the
  two new byte oracles and the text-scanning expansion guard.
- Pre-rebase, on the original base: `vp run typecheck` pass; `--project web`
  364 files / 4,089 tests (one unrelated `waitFor` flake in
  `gym-owner-reassign-panel.test.tsx` under full-suite load, green on its own);
  `--project i18n` 101 pass; `check:i18n` + `:orphans` OK; `vp check` 0 errors.
- Pre-commit hook pass, no `--no-verify`.

Manual QA plan is in `.boardsesh/qa-notes.md` and was exercised against `vp run dev` for the MoonBoard climb pages, the canonical, both shards and the empty-store self-heal.

**Not verified:** the in-process RSS of the 126,642-row refresh, and the canonical `<link>` in the `es`/`fr`/`de` locales (checked in `en-US` only).

---

## Review round: three false claims and a missing guard

Everything below was reproduced before it was changed.

- **The ordinal invariant was false.** Detailed above. Three comment/doc sites corrected; no behaviour change. If ordinal stability for existing boards is ever wanted, it needs an explicit rank in that sort — `boardCount` cannot deliver it — and `docs/sitemap.md` now says so.
- **`moonboard-reaches-the-store.test.ts` could not see the displacement it claimed to guard.** Its fixture was kilter + moonboard only, so `expect(urlRows[0]?.boardType).toBe('kilter')` passes under any insertion point (`kilter` sorts first either way). It now carries a real Tension config (Original Layout / Full Wall / all four sets, catalogue-accurate) and pins the whole emitted sequence `['kilter','kilter','moonboard','moonboard','tension','tension']`. Mutation — rank `moonboard` last in the group sort → the new assertion **reds** and the old one stays **green**, which is the finding in one run.
- **`moonboard-canonical-identity.test.ts`'s last case never called the resolver it is named after.** Both sides of `the full-set tuples are the ones the shared render-board resolver already names` came from `MOONBOARD_SETS`; `getDefaultRenderBoard` — the function the config source actually derives from — was not in it. It now pins the resolver's tuples against the same literals. Mutation — make `getDefaultRenderBoard` return a one-set subset for MoonBoard → **1 red, 14 green**, i.e. the entire canonical suite would have gone on canonicalising tuples the shard had stopped emitting.
- **`/sitemaps/climbs/[page]` had no `maxDuration`.** Added at the time, matching `/sitemap.xml` and the cron refresher. **Reverted in the rebase** — #4648 removed `/sitemap.xml`'s own, because the Railway container has no per-invocation ceiling to raise, so the export would have described a platform this route no longer runs on. The comment in its place says so.
- **The refresh-cost and shard-byte figures were the optimistic ends of their ranges.** Both re-measured and restated above and in `docs/sitemap.md`.

### DIFF
--- a/docs/sitemap.md
+++ b/docs/sitemap.md
@@ -82,8 +82,10 @@ layout that has listed, non-draft climbs, and appends them to the listed configs
 backend deploy to a sitemap change and move the board picker underneath it.
 
 The synthetic configs carry `boardCount: 0` and `totalAscents: 0`, which is
-honest (there is no `user_boards` row shape behind them) and also keeps them last
-under `isBetterConfig`'s ordering, so the source is strictly additive. The
+honest (there is no `user_boards` row shape behind them) and nothing more. It
+does not hold them last in any ordering: `isBetterConfig` only ranks candidates
+within one `boardType:layoutId` group, so the source is additive in **content**
+and not in **ordinal** — see "Adding a board also moves `ordinal`" below. The
 `climbCount` is a plain grouped count over `board_climbs`, **not** the tier-2
 `DISTINCT ON` scan: both shards read that number only as a `> 0` gate, so making
 the boards shard pay the climbs shard's cost budget for a boolean would be the
@@ -129,9 +131,14 @@ emits nothing new — a change that looks complete in review and ships a no-op.
 half-done edit.
 
 Adding a board also moves `ordinal`, because `resolveClimbSitemapGroups` orders
-by board type rather than appending. That is a catalogue change, which is what
-`ordinal` is allowed to move for; each page stays self-contained and its
-`<lastmod>` is recomputed from the rows it actually holds.
+by board type rather than appending — the sort is a plain lexicographic compare
+on `boardType`, so `moonboard` landed between `kilter` and `soill` rather than on
+the end. On the dev image that pushed 15,032 Soill/Tension/Touchstone URLs from
+pages 3–5 onto pages 8–9. That is a catalogue change, which is what `ordinal` is
+allowed to move for; each page stays self-contained and its `<lastmod>` is
+recomputed from the rows it actually holds. A board type that must NOT move the
+existing ordinals needs an explicit rank in that sort — `boardCount` and
+`isBetterConfig` cannot deliver it, since neither is consulted across groups.
 
 ## Degrade at the index, fail closed at the shard
 
@@ -205,9 +212,16 @@ live scan they replaced, and the next refresh repopulates them.
 
 The page read (`buildClimbShardPage`) follows the same doctrine: an empty
 `sitemap_climb_urls` or a read that throws falls back to the live grouped build —
-the 51 s path, correct and never worse than before the store existed. There is no
-staleness bound on pages; the summary row's 48 h shout covers the store as a whole,
-since both tables share one refresh.
+the 51 s path, correct and never worse than before the store existed. It is
+TTL-cached per instance, so only the first request into an empty store pays it
+(measured on the dev image with MoonBoard in: 22.7 s for the first hit against a
+warm database, dev route compile included, then 0.40 s and 0.48 s; the same scan
+costs 64.9 s cold — see "What the refresh costs" below, it is the same builder).
+That is why the route exports `maxDuration = 300` — the `after()`
+self-heal only fires on `/sitemap.xml`, so a crawler that lands on a page URL
+first has to be allowed to finish the slow build rather than be cut off. There is
+no staleness bound on pages; the summary row's 48 h shout covers the store as a
+whole, since both tables share one refresh.
 
 ### Saying which path served (#4583)
 
@@ -347,6 +361,26 @@ commented out in `production-deploy.yml`.
 a second cron service: the refresh runs on the Railway scheduler
 (`refresh-sitemap-climbs`), and `registry.test.ts` reds if a path is scheduled
 on both sides.
+
+### What the refresh costs
+
+Worth stating as a range, because the spread between a warm and a cold Postgres
+is wide enough to mislead if only one end is quoted. All measured on the full
+dev image (648k climbs, 85,347 tier-2 URLs across 13 groups, MoonBoard included),
+through the real cron route:
+
+| database state                          | `scanDurationMs` | measured                       |
+| --------------------------------------- | ---------------- | ------------------------------ |
+| cold — container restarted, empty store | **64.9 s**       | #4578 review, reproduced twice |
+| warm, repeat runs on a loaded box       | 20.0 – 28.4 s    | #4578 review, three runs       |
+| warm, quiet box                         | 8.9 – 12.2 s     | #4578 authoring, two runs      |
+
+Budget against the **cold** number. `maxDuration` on the cron route is 300 s;
+scaling 64.9 s by production's projected 126,642 rows over the dev image's 85,347
+gives roughly **96 s**, so the margin is about 3x — not the 24x the warm figure
+suggests. The same scan is what `buildClimbShardPage` falls back to when the
+store is empty, which is why `/sitemaps/climbs/[page]` carries a `maxDuration`
+of its own.
 
 ### Refusals, and how to get out of one
 
--- a/docs/sitemap.md
+++ b/docs/sitemap.md
@@ -58,6 +58,81 @@ and all other sitemap shards are unaffected by the switch. Turning it off and ba
 on costs nothing but a redeploy: the store is a cache, and the `after()` self-heal
 below repopulates it on the first crawl.
 
+## Where board configurations come from
+
+Two shards need the list of board configurations to build URLs from, and they do
+**not** ask the same question. `board-config-source.ts` is the seam.
+
+`getAllBoardConfigsOrThrow` (`server-popular-configs.ts`) answers _which board
+pages are worth building_. It reads `popularBoardConfigs` from the backend, which
+builds its universe with a `GROUP BY` over `board_product_sizes_layouts_sets` —
+the Aurora sync tables. That is the right universe for the home rail and the
+mobile board picker, and it is why MoonBoard was in no sitemap for so long:
+nothing writes a psls row for MoonBoard. The only writers are the Aurora
+importers and no migration seeds one, so MoonBoard was never ranked out — it was
+never a candidate.
+
+The sitemap's question is _which climbs are worth indexing_, and MoonBoard is the
+largest board in the addressable set. Its configuration is not in the database at
+all; it is the static `MOONBOARD_LAYOUTS` / `MOONBOARD_SETS` tables that
+`getDefaultRenderBoard` already resolves for every renderer we ship. So
+`board-config-source.ts` synthesises one `PopularBoardConfig` per MoonBoard
+layout that has listed, non-draft climbs, and appends them to the listed configs
+— rather than changing what `popularBoardConfigs` returns, which would couple a
+backend deploy to a sitemap change and move the board picker underneath it.
+
+The synthetic configs carry `boardCount: 0` and `totalAscents: 0`, which is
+honest (there is no `user_boards` row shape behind them) and also keeps them last
+under `isBetterConfig`'s ordering, so the source is strictly additive. The
+`climbCount` is a plain grouped count over `board_climbs`, **not** the tier-2
+`DISTINCT ON` scan: both shards read that number only as a `> 0` gate, so making
+the boards shard pay the climbs shard's cost budget for a boolean would be the
+wrong trade.
+
+### Two callers, two fail policies
+
+|                                   | caller                 | on a MoonBoard count failure    |
+| --------------------------------- | ---------------------- | ------------------------------- |
+| `getSitemapClimbConfigsOrThrow()` | the climb shards       | **throws**                      |
+| `getBoardsShardConfigsOrThrow()`  | `/sitemaps/boards.xml` | logs, serves the listed configs |
+
+The strict one has to be strict. The climbs shard resolves its groups twice per
+crawl — once for the summary the index sizes pages from, once for the item build
+— and `pagedShardRouteHandler` throws "cache epochs disagree" the moment those two
+see different group sets. A tolerated failure there is worse than a 503.
+
+The tolerant one has no second builder to disagree with, and the arithmetic is
+lopsided: MoonBoard contributes 8 of `boards.xml`'s 668 items on the dev image
+against the listed configs' 660. Before this module existed no database failure
+could reach that shard at all — it was a GraphQL fetch behind a backend Redis
+cache with a one-year TTL — so 503ing 660 working Kilter/Tension/Decoy URLs
+because a grouped count timed out would be a regression bought with nothing. A
+failed **listed** fetch still throws on both paths: that is the leg whose loss
+would tell Google the boards were deleted.
+
+The count query carries a 10 s budget applied _inside_ the shared single-flight
+promise, so a give-up is not memoised and the next caller retries instead of
+joining a stall that already gave up. Nothing else bounds it: the pool sets
+`connect_timeout: 30` and `statement_timeout` is off by default (PgBouncer
+rejects it as a startup parameter — see `docs/db-connectivity.md`).
+
+### The refresher is the only door
+
+Since the climb store landed, adding a board to the sitemap means adding it to
+`buildAllTier2UrlRows()`, and nowhere else is sufficient. The shard pages do not
+build anything: they read `sitemap_climb_urls`, and that table is written by one
+refresher whose only selection entry point is that function. A config that
+reaches `/sitemaps/boards.xml` and the live summary fallback but not the
+refresher produces a store that has never heard of the board and a shard that
+emits nothing new — a change that looks complete in review and ships a no-op.
+`__tests__/moonboard-reaches-the-store.test.ts` exists to red exactly that
+half-done edit.
+
+Adding a board also moves `ordinal`, because `resolveClimbSitemapGroups` orders
+by board type rather than appending. That is a catalogue change, which is what
+`ordinal` is allowed to move for; each page stays self-contained and its
+`<lastmod>` is recomputed from the rows it actually holds.
+
 ## Degrade at the index, fail closed at the shard
 
 The doctrine splits by layer, and the split is deliberate.
@@ -404,6 +479,18 @@ milliseconds. What remains slow is the **fallback**: an empty
 the first refresh runs. The self-heal's empty-table probe bounds that window to one
 crawl of `/sitemap.xml`; the manual curl closes it immediately. Disabled requests do
 not reach either path; they return the cacheable 410.
+
+That one crawl really does degrade, and it is worth knowing what it looks like so
+nobody mistakes it for a regression. Measured on the dev image with the store
+truncated, over the thirteen groups the code resolved when this branch was
+cut — the shape is what matters here, and the counts are owed a re-measure
+since `main` widened the tier-2 angle predicate to the routable set: `/sitemap.xml`
+answers 200 in 4.3 s with `X-Sitemap-Degraded: climbs` and no climbs pages
+listed, then the `after()` self-heal writes 85,347 rows and every request after
+it is healthy at ~0.1 s. Do not reach for concurrency here — a fan-out over the
+group list was tried and dropped, because a cold thirteen-group scan does not fit
+in `SHARD_DEADLINE_MS` at any lane count and the pool is ten connections wide
+(#4461). The answer to a slow fallback is a refreshed store.
 
 Both shard paths now check bytes, and they check different numbers on purpose
 (#4618, closed by #4648).
--- a/docs/sitemap.md
+++ b/docs/sitemap.md
@@ -217,11 +217,13 @@ TTL-cached per instance, so only the first request into an empty store pays it
 (measured on the dev image with MoonBoard in: 22.7 s for the first hit against a
 warm database, dev route compile included, then 0.40 s and 0.48 s; the same scan
 costs 64.9 s cold — see "What the refresh costs" below, it is the same builder).
-That is why the route exports `maxDuration = 300` — the `after()`
-self-heal only fires on `/sitemap.xml`, so a crawler that lands on a page URL
-first has to be allowed to finish the slow build rather than be cut off. There is
-no staleness bound on pages; the summary row's 48 h shout covers the store as a
-whole, since both tables share one refresh.
+The `after()` self-heal only fires on `/sitemap.xml`, so a crawler that lands on
+a page URL first has to be allowed to finish that slow build rather than be cut
+off. On the Railway container it is, with nothing to configure: the route carries
+no `maxDuration`, for the same reason `/sitemap.xml` dropped its own (#4648) —
+there is no per-invocation ceiling to raise. There is no staleness bound on
+pages; the summary row's 48 h shout covers the store as a whole, since both
+tables share one refresh.
 
 ### Saying which path served (#4583)
 
@@ -375,12 +377,13 @@ through the real cron route:
 | warm, repeat runs on a loaded box       | 20.0 – 28.4 s    | #4578 review, three runs       |
 | warm, quiet box                         | 8.9 – 12.2 s     | #4578 authoring, two runs      |
 
-Budget against the **cold** number. `maxDuration` on the cron route is 300 s;
-scaling 64.9 s by production's projected 126,642 rows over the dev image's 85,347
-gives roughly **96 s**, so the margin is about 3x — not the 24x the warm figure
-suggests. The same scan is what `buildClimbShardPage` falls back to when the
-store is empty, which is why `/sitemaps/climbs/[page]` carries a `maxDuration`
-of its own.
+Budget against the **cold** number. `maxDuration` on the refresh route is still
+300 s — vestigial on the container, load-bearing only on the frozen Vercel
+rollback deployment; scaling 64.9 s by production's projected 126,642 rows over
+the dev image's 85,347 gives roughly **96 s**, so that margin is about 3x, not
+the 24x the warm figure suggests. The same scan is what `buildClimbShardPage`
+falls back to when the store is empty, and on the container that fallback has no
+ceiling to run into.
 
 ### Refusals, and how to get out of one
 
--- a/docs/web-reposition.md
+++ b/docs/web-reposition.md
@@ -501,22 +501,23 @@ Tier 2 — listed, non-draft climbs with at least ten ascents — now ships as p
 sitemap shards, and the five JSON-LD types the front doors were missing are
 emitted from one escaping helper.
 
-**The climb shards submit the default locale only.** This is a deliberate
-inconsistency with the boards shard, which does fan out to all four locales, and
-the difference is volume: boards is 690 items → 2,760 URLs; production emits
-52,842 climb items, which would fan out to 211,368 locale URLs, each carrying a
-five-entry `xhtml:link` block. Even one 10,000-item climb page would expand to
-40,000 URL entries and exceed Vercel's 4.5 MB serverless response ceiling.
+**The climb shards submit the default locale only** — and since #4648 so does
+the boards shard, so this is no longer an inconsistency to explain. The volume
+argument that motivated it stands on its own: production emits 52,842 climb
+items, which fanned out would be 211,368 locale URLs each carrying a five-entry
+`xhtml:link` block. One 10,000-item page would expand to 40,000 URL entries and
+take a page from ~2.5 MB to 8.7 MB, past `pagedShardByteBudget`'s 5 MB. Do not
+"fix" it by fanning climbs out.
 
 (Measured against production with the branch's exact angle, size, set, listing,
 ascent and alias predicates: 127,131 tier-2 climbs exist, while 52,842 resolve
-through a selected public board configuration → 6 pages. All 73,412 MoonBoard
-climbs are absent because there is no MoonBoard configuration, 69 Kilter layout-5
-climbs have no selected layout, and the chosen size/set predicates exclude 808
-more. An earlier 53,650 estimate stopped at the layout intersection and therefore
-overstated the emitted set by those 808. Nothing in the code depends on the
-snapshot count; the page count comes from the summary at request time.) Do not
-"fix" the inconsistency by fanning climbs out.
+through a selected public board configuration → 6 pages. 73,412 of the absent
+climbs were MoonBoard's, because no MoonBoard configuration existed to resolve
+through — lifted by #4493; 69 Kilter layout-5 climbs have no selected layout, and
+the chosen size/set predicates exclude 808 more. An earlier 53,650 estimate
+stopped at the layout intersection and therefore overstated the emitted set by
+those 808. Nothing in the code depends on the snapshot count; the page count
+comes from the summary at request time.)
 Nothing is noindexed — the locked decision on `/es`, `/fr` and `/de` climbing
 pages is untouched. They stay indexable, they stay link-discovered, and they
 carry reciprocal HTML hreflang from `createPageMetadata`'s `alternates.languages`,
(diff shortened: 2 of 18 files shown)


marcodejongh:
Addressed the two non-blocking review points in 5a067655e:

- **(3) untested failure path** — the in-process layer only assigns `cachedCounts` inside the `.then`, so a rejected count is never memoised. Two new tests pin it: a transient read failure does not cost an hour of missing MoonBoard, and three concurrent callers share the one rejection while the next caller retries. Both go red under their own mutation (memoise the rejection; drop the single-flight).
- **(4) static catalogue tables** — added a note at `MOONBOARD_LAYOUT_KEYS` naming the manual step a new MoonBoard layout needs, and pointing at the two test files whose literal tuples go red when someone adds one. That also covers the same observation left on #4576.

The two blocking points stand as written in the PR body: the stack ordering (must not merge before #4576) and the `SHARD_DEADLINE_MS` question, which is the open call for @marcodejongh. Staying draft until that is answered.

marcodejongh:

Every dev-image figure in the body re-derived against production, through **this branch's own code** rather than the issue's SQL: the real `popularBoardConfigs` resolver query → `getSitemapClimbConfigsOrThrow()` → `resolveClimbSitemapGroups()` → `buildTier2ClimbQuery` / `buildTier2ClimbSummaryQuery` / `boardConfigsToItems` / `renderUrlset`. Harness described at the bottom.

### Two conclusions change

**1. All seven MoonBoard layouts have listed climbs in production, not four.** The dev image's three empty layouts (2010, 2024, Mini 2025) are an artefact of the snapshot. So this PR ships **7** synthetic configs, **16** resolved groups and **14** boards-shard URLs, not 4 / 13 / 8.

**2. The climbs shard goes to 13 pages, not 9 — and the summary is outside `SHARD_DEADLINE_MS` on production's *warm* path, in both arms.** Details below. The byte budget is fine; the deadline is not.

### Catalogue

|---|---|---|---|
| listed configs from `popularBoardConfigs` | 45 | **47** | +2 |
| MoonBoard layouts with ≥1 listed non-draft climb | **4 of 7** | **7 of 7** | +3 |
| MoonBoard listed non-draft climbs, total | 141,800 | **271,780** | +129,980 |
| — MoonBoard 2010 (layout 1) | 0 | **124** | +124 |
| — MoonBoard 2016 (layout 2) | 59,019 | 92,677 | +33,658 |
| — MoonBoard 2024 (layout 3) | 0 | **36,134** | +36,134 |
| — Masters 2017 (layout 4) | 54,678 | 73,244 | +18,566 |
| — Masters 2019 (layout 5) | 24,433 | 56,864 | +32,431 |
| — Mini 2020 (layout 6) | 3,670 | 7,867 | +4,197 |
| — Mini 2025 (layout 7) | 0 | **4,870** | +4,870 |
| resolved climb groups, before | 9 | 9 | 0 |
| resolved climb groups, after | 13 | **16** | +3 |

The 9 production groups before this PR are `decoy/2, grasshopper/1, kilter/1, kilter/8, soill/1, tension/9, tension/10, tension/11, touchstone/1` — same count as the dev image, and `tryConstructSlugViewUrl` still drops Kilter layout 5 along with kilter/2 and kilter/7.

### Tier-2 URLs

|---|---|---|---|
| tier-2 items, before | 41,336 | **53,137** | +11,801 |
| tier-2 items, after | 85,347 | **126,549** | +41,202 |
| MoonBoard tier-2 items added | 44,011 | **73,412** | +29,401 |
| — 2010 | — | 7 | |
| — 2016 | 15,907 | 21,973 | +6,066 |
| — 2024 | — | 6,419 | |
| — Masters 2017 | 19,338 | 25,551 | +6,213 |
| — Masters 2019 | 6,825 | 13,990 | +7,165 |
| — Mini 2020 | 1,941 | 3,779 | +1,838 |
| — Mini 2025 | — | 1,693 | |

The issue body's 73,412 MoonBoard tier-2 figure is **exact**. Its 127,131 total is within 0.5% of the 126,549 measured here.

### Caps and page count

| figure | dev image | production | budget |
|---|---|---|---|
| climbs shard pages, before | 5 | **6** | — |
| climbs shard pages, after | **9** | **13** | — |
| largest climbs page, bytes | 2,935,185 (73%) | **2,999,695 (75.0%)** | `MAX_SHARD_BYTES` 4,000,000 |
| URLs per climbs page | 10,000 | 10,000 | `MAX_URLS_PER_SHARD` 45,000 |
| `/sitemaps/boards.xml` items, before | 660 | **690** | `MAX_ITEMS_PER_SHARD` 11,250 |
| `/sitemaps/boards.xml` items, after | 668 | **704** | |
| — MoonBoard `/list` items | 8 | **14** (7 layouts × 2 angles) | |
| `/sitemaps/boards.xml` locale-expanded URLs, after | 2,672 | **2,816** | `MAX_URLS_PER_SHARD` 45,000 |

**No cap is exceeded.** The byte budget survives production comfortably (75.0% of 4 MB on the fullest page, rendered through the real `renderUrlset` + `expandDefaultLocaleOnly`), and the boards shard is nowhere near its item or URL ceiling. All 14 MoonBoard `/list` paths are free of `//` and `?`.

What *does* move is the cold-crawl arithmetic the body states: N goes **6 → 13**, not 5 → 9. Summing this run's per-group item builds, one full production item build is ~43 s of query time from an off-cluster client (dev: 7.6 s), so the "worst-case cold crawl ~20 s → ~68 s" line understates production materially. The remedy the body already names — the materialised tier-2 table — is the same one.

### The deadline, measured on production (this is the finding)

`SHARD_DEADLINE_MS` is 3,000 ms. Wall clock for the whole summary through `buildTier2ClimbSummaryQuery`, warm, three runs each, at the shipped `SUMMARY_CONCURRENCY = 3`:

| | dev image, 3 lanes, warm | production, 3 lanes, warm |
|---|---|---|
| **today on `main`** (9 groups) | 1,022 ms | **3,032 / 3,086 / 2,893 ms** (median 3,032) |
| **this PR** (16 groups) | 1,386 ms | **7,923 / 5,218 / 5,294 ms** (median 5,294) |

Net of this client's network round-trips (~260 ms/query measured against Railway's public TCP proxy; a Vercel lambda is closer) that is roughly 2.2 s before and 3.9 s after. Server-side only, via `EXPLAIN (ANALYZE, TIMING OFF)` summed sequentially: **13,123 ms for the 9 groups, 23,439 ms for the 16** — and `kilter/1` alone measures 6,988 ms of execution time, so **no lane count can bring the summary inside 3,000 ms in production**; the tail is one group.

So the body's "three lanes puts thirteen groups comfortably inside a budget nine sequential ones were already outside" is a dev-image conclusion and does not transfer. In production:

- `main` is already **at** the deadline warm (3.0 s), not merely past it cold — the body's stronger version of that claim holds.
- After this PR the summary is **~1.8× the deadline warm**, so the climbs shard will degrade out of `/sitemap.xml` on the warm path too, not only on a cold Data Cache.
- This is **not a regression the PR introduces** — the shard was already losing the race, `withDeadline` does not cancel, the degrade is self-healing per instance, and the summary sits behind a 6-hour Data Cache — but the numbers in the body should be corrected and the "comfortably inside" framing dropped. Three lanes is still the right call; it just buys less than the dev image suggested.

### Harness and SQL

Production configs came from the resolver's own query (`getPopularConfigs()` in `packages/backend/src/graphql/resolvers/social/boards.ts`, run verbatim; 47 rows, 224 s cold — it is Redis-cached for a year in production). Those rows were fed to a throwaway vitest file in this worktree with `server-only` and `next/cache` stubbed, `@/app/lib/db/db` pointed at a read-only production `drizzle(postgres(DB_URL, { ssl: 'require' }))`, and `@/app/lib/server-popular-configs` returning the 47 rows — then the real `getSitemapClimbConfigsOrThrow` / `getBoardsShardConfigsOrThrow` / `resolveClimbSitemapGroups` / `buildTier2ClimbQuery` / `climbRowsToItems` / `boardConfigsToItems` / `renderUrlset` were called unmodified. The probe file has been deleted; the worktree is clean.

The MoonBoard per-layout counts are `buildMoonBoardClimbCountQuery` verbatim:

```sql
SELECT layout_id, count(*) FROM board_climbs
WHERE board_type = 'moonboard' AND is_listed = true AND is_draft = false
GROUP BY layout_id;
```

The tier-2 count for one group is `buildTier2ClimbSummaryQuery`; as literal SQL for `moonboard/4` (size 1, sets `{11,12,13,14,15,16}`):

```sql
SELECT count(*)::int
FROM (
  SELECT DISTINCT ON (s.climb_uuid) s.climb_uuid
  FROM board_climb_stats s
  JOIN board_climbs c ON c.uuid = s.climb_uuid AND c.board_type = s.board_type
  WHERE c.board_type = 'moonboard' AND c.layout_id = 4
    AND c.is_listed = true AND c.is_draft = false
    AND s.ascensionist_count >= 10
    AND s.angle IN (25, 40)                                   -- ANGLES.moonboard
    AND (c.required_set_ids IS NULL
         OR c.required_set_ids <@ ARRAY[11,12,13,14,15,16]::int[])   -- no size predicate for MoonBoard
    AND NOT EXISTS (SELECT 1 FROM board_climb_aliases a
                    WHERE a.board_type = c.board_type AND a.alias_uuid = c.uuid
                      AND a.alias_uuid <> a.canonical_uuid)
  ORDER BY s.climb_uuid, s.ascensionist_count DESC,
           COALESCE(s.angle = c.angle, false) DESC, s.angle ASC
) chosen;
```

The other fifteen groups are the same query with `(board_type, layout_id, size_id, set_ids)` = `decoy/2` size 1 `{2…20}` · `grasshopper/1` size 2 `{1,2,3}` · `kilter/1` size 10 `{1,20}` · `kilter/8` size 25 `{26,27,28,29}` · `moonboard/1` `{1}` · `moonboard/2` `{2,3,4}` · `moonboard/3` `{5…10}` · `moonboard/5` `{17…23}` · `moonboard/6` `{24,25,26,27}` · `moonboard/7` `{28,29,30,31}` · `soill/1` size 1 `{1}` · `tension/9` size 1 `{8,9,10,11}` · `tension/10` size 6 `{12,13}` · `tension/11` size 6 `{12,13}` · `touchstone/1` size 1 `{1}` — with the `compatible_size_ids @> ARRAY[size]::int[]` predicate restored and `required_set_ids <@ …` non-nullable for every non-MoonBoard group, and `s.angle IN (0,5,…,70)` for the Aurora boards.

### Caveat

The database is live. The summary pass counted 126,518 tier-2 items and the item-build pass 126,549, ten minutes apart — a 31-row drift from ongoing sync, not a discrepancy between the two builders. Treat all figures as ±0.05%.

marcodejongh:

Conflicting: packages/web/app/lib/seo/sitemap/shard-registry.ts.

Issue #4552, opened 2026-08-19
Materialise the tier-2 climb URLs so /sitemaps/climbs/N.xml stops taking 51s cold


marcodejongh:
`/sitemaps/climbs/N.xml` takes **51 s cold** in production. #4523 fixed the *index* dropping the
climbs shard by materialising the summary into `sitemap_shard_refreshes`; it deliberately left the
shard *route* alone. This is the other half.

## Why it is still slow

`buildPage:` in `PAGED_SHARD_REGISTRY` (`packages/web/app/lib/seo/sitemap/shard-registry.ts`) still
calls `buildTier2ClimbItems()` and slices the result in JS:

```ts
buildPage: async (page: number) => {
  const items = await buildTier2ClimbItems();
  const start = (page - 1) * CLIMB_URLS_PER_SHARD;
  return { items: items.slice(start, start + CLIMB_URLS_PER_SHARD), totalItems: items.length };
},
```

So page N costs the **whole ordered list** — the same sixteen sequential `DISTINCT ON` scans, plus
`tryConstructSlugViewUrl` over ~52,000 rows — before it can slice. The list is behind a 6-hour
in-process TTL and a single-flight, but that is a per-instance defence: the payload is ~20 MB, well
past Vercel's 2 MB Data Cache entry ceiling, so it cannot be cached across instances. A genuinely
cold crawl where Googlebot fetches six pages that all miss the CDN and land on six lambdas really is
six full builds.

## What to build

A `sitemap_climb_urls` table of the rendered selection, refreshed by the machinery #4523 already
added. This is **purely additive** — nothing gets built twice:

| already exists (#4523) | what this adds |
|---|---|
| `sitemap_shard_refreshes` + migration | `sitemap_climb_urls` + a migration |
| `refreshStoredClimbSummary()` — sixteen scans, shrink guard, `pg_try_advisory_xact_lock`, `?force=1` | the same refresher also writes the URL rows |
| `/api/internal/refresh-sitemap-climbs` + the six-hourly cron | unchanged |
| the `after()` self-heal on `/sitemap.xml` | unchanged |
| `fetchClimbShardSummary()` (registry `summary:`) | `buildClimbShardPage()` (registry `buildPage:`) |

Sketch:

- `sitemap_climb_urls`: `ordinal` (int PK), `path` (text), `last_modified` (timestamp),
  `board_type`, `layout_id`. `ordinal` is the emission order the pages slice on — assigned at
[...]

PR #4579, opened 2026-08-19
feat(web): fill the setters sitemap shard as a paged, linkable-only tree

marcodejongh:
## Summary

Stacked on #4577 — review that first; this PR's diff is against it.

`/sitemaps/setters.xml` was declared-empty: `buildSetterEntries()` returned `[]`, so `buildIndexEntry` dropped it and Google was never shown the URL. The two things it waited on land in #4577 — the page server-renders, and it 404s for a username nobody has a publicly visible climb under — so the shard can ship.

### It ships paged, not as a filled fixed builder

**Volume, not locale expansion.** The dev image has ~108,000 distinct `(board_type, setter_username)` pairs against `MAX_ITEMS_PER_SHARD`'s 11,250, so one file cannot hold them at any expansion — about ten pages are needed. `setters` therefore joins `PAGED_SHARD_REGISTRY`, served from `/sitemaps/setters/N.xml`.

(This PR originally argued the move from `shardRouteHandler` hardcoding `expandAllLocales`, pricing a fanned-out setter `<url>` at ~2.4 kB against a 4 MB ceiling. **Both halves of that stopped being true while the branch was out** — **#4996** gave every shard its own `expansion`, and the declared-empty fixed `setters` entry already carried `default-locale-only`; and the Vercel response ceiling went away with the Railway migration (**#4648**), replaced by `pagedShardByteBudget`. The shard is still `default-locale-only`, for the reason the climb shards are: setter pages cross-canonicalise onto the default locale, so listing the twins would advertise URLs whose own canonical points elsewhere.)

`app/sitemaps/setters.xml/route.ts` is deleted. That is safe rather than a broken link: the route only ever answered an empty `<urlset>`, `buildIndexEntry` returned null for it, and the index has never listed the URL. The filename shape the issue asks for (`setters-1.xml`) cannot exist — Next has no partial dynamic segments, which is why paging is `/sitemaps/<dir>/<page>.xml`.

### The gate is *linkable*, not merely non-empty

A setter is submitted when they have **at least three** publicly visible climbs sitting on a configuration `resolveClimbSitemapGroups` resolves. Two halves, both load-bearing:

- **Linkable.** Without it, a setter whose climbs are all on a layout no group resolves gets submitted with an `<h1>` over rows carrying no crawlable link at all. The predicate is the same board/layout + `@>` size containment + `<@` set containment the climbs shard applies before it will name a config in a climb's URL.
- **A floor of three.** The same judgement the climbs shard already makes one level down (tier 2 = ten ascents, tier 3 held back until Search Console shows coverage). A one- or two-climb setter page is thin by construction. The page still serves 200 at one visible climb on any board — this decides what we push, not what exists.

**No hardcoded MoonBoard exclusion.** MoonBoard is absent from the resolved groups today because its set slugs do not round-trip through the MoonBoard page's parser, so it falls out on its own — and falls back in when #4493 fixes that parser, with no edit here. The test asserts the exclusion through the real group resolver rather than a local `boardType !== 'moonboard'` check, so it reds if the parser is fixed without this shard noticing.

### `<lastmod>`

`max(board_climbs.updated_at)` per setter, rendered through `to_char(..., 'YYYY-MM-DD"T"HH24:MI:SS.MS"Z"')`.

Not `board_setter_stats`, on two counts. Its recompute INNER JOINs `board_climb_stats`, so a setter whose climbs carry no stats row is simply absent from the table; and its `updated_at` is `now()` at nightly refresh, a job clock — publishing that as `<lastmod>` claims every setter changed every night, which is the signal destruction `entries.ts` forbids.

Two honest caveats, stated because they are easy to overclaim:

- `board_climbs.updated_at` is itself a sync-touch clock rather than a pure content clock. It is the same clock the climbs shard already publishes as `<lastmod>`, so this is consistency, not a new defect.
- On the **dev image** it is not even that: every visible row's `updated_at` falls inside one 23-minute window on 2026-08-10 (`min` 20:38:09, `max` 21:01:33), i.e. the image build stamped them all. So the dev image cannot tell you whether the per-setter spread is meaningful in production — only that the value is non-null on all 648,657 rows and parses. Production's update triggers are what make it a real clock, and that is unverified here.

### Username routability lives in SQL

Leading/trailing whitespace (encodes to a `%20` crawlers strip), `/ ? #`, and control characters are excluded **in the query**, not in a JS filter afterwards. The summary sizes the pages and the item list fills them; a JS-side filter would let the summary advertise a page the build cannot fill, which the paged handler turns into a 503. The JS round-trip check stays as defence in depth and warns loudly if it ever drops anything.

### Numbers — dev image, production reconciliation outstanding

Measured against the local `ghcr.io/boardsesh/boardsesh-dev-db` image, **not** production. Prior work in this campaign found the dev image runs roughly a third short of production on some counts.

| | dev image |
|---|---|
| setters with ≥1 visible climb (the page's own 404 rule) | 91,946 |
| setters with ≥3 visible climbs, any board | 35,950 |
| **submitted: linkable and ≥3 linkable visible** | **26,205** → 3 pages at 10,000 |
| largest submitted setter, visible climbs | 4,723 |

**The submitted count was re-measured in review and corrected from 24,420.** The earlier figure came from an approximation of the group resolver written in SQL, which skipped `isResolvableGroup`. Feeding the real `getPopularConfigs()` output through the real `resolveClimbSitemapGroups` yields **9** resolvable `(board_type, layout_id)` groups on the dev image, not 12: `decoy/1`, `kilter/2` and `kilter/7` all fail `tryConstructSlugViewUrl`. 26,205 is that predicate rendered verbatim and run.

### Shard-summary cost, and why the two suggested remedies were not taken

Timed against the dev image (648,657 `board_climbs`, 45 configs → 9 groups), `max_parallel_workers_per_gather` at the container default:

| | dev image |
|---|---|
| full grouped scan, 26,205 rows returned | **1,434 / 1,366 / 1,370 ms** |
| the same predicate wrapped in `count(*)` + `max(updated_at)` | **1,409 / 1,354 / 1,439 ms** |
| `SHARD_DEADLINE_MS` | 3,000 ms |

`EXPLAIN (ANALYZE, BUFFERS)`: `Bitmap Heap Scan on board_climbs … Heap Blocks: exact=42354, Buffers: shared hit=562 read=43964`, HashAggregate spilling (`Batches: 5, Disk Usage: 744kB`). The 43,964-block read repeats on every run — these are not warm-buffer numbers, they are OS-page-cache numbers.

A review lens measured **2,999 ms** on a genuinely cold cache and asked for one of two remedies. Neither was taken, and here is why:

- **"Make the summary a `COUNT(*)` instead of reading the item list."** Measured above: **identical cost**, because the grouped scan is the entire bill and the 26k-row transfer plus `encodeURIComponent` is noise. It would trade a structural guarantee — `setterRowsToItems` *throws* on any drop, so whenever the item build succeeds `items.length` equals the row count exactly — for no speed.
- **"Give the paged summary a longer deadline."** That is a change to shared sitemap infrastructure affecting the climbs shard too, and the campaign ruling on #4493 is explicit: cache the summary, do not raise the deadline. It is also the wrong lever — the honest fix for a scan that cannot fit is a materialised summary table, which is already recorded as the tier-3 follow-up for the climbs shard.

**What this means in production, stated plainly:** production `board_climbs` is materially larger than the dev image, so a genuinely cold `/sitemap.xml` will very likely miss the 3,000 ms deadline and drop this shard. `withDeadline` does not cancel the work, so the abandoned query still fills both caches and the next request past the 60 s `DEGRADED_CACHE_CONTROL` window gets the shard — the degrade is self-healing per instance, not a lost shard. It is nonetheless the same posture the climbs summary already ships with (~9.2 s cold on this image), and **it is worth Marco's call whether paged summaries should get their own deadline.** Filed as #4583 rather than changed here.

The shard sits behind three layers (`unstable_cache` + in-process TTL + single-flight) for the reason `unstable_cache` alone does not cover: it does not dedupe concurrent misses, and a crawl burst is `/sitemap.xml` plus every page arriving together (#4461).

Production reconciliation of every number above is outstanding — no production database was read for this PR.

## Release Notes

## Test plan

1. CI green.
2. Open a setter page from a Google result → climbs listed.

## Risk

Risk: 3/5 — submits roughly ten pages of setter URLs, and a submitted URL that
renders a spinner or answers 200 for a name nobody has climbs under is a
crawl-budget commitment made against a page that cannot earn it. Both are fixed
by #4577, which this stacks on, and the merge gate below is the check that they
really shipped. The shard is derived and gated by the registry, so withdrawing
it is one revert.

## Rebased onto the rebased #4577, 2026-09-06

Three conflicts, all in the sitemap infrastructure `main` reworked under this
branch, plus two fixture repairs.

- **`shard-registry.ts`** — the fixed `setters` entry `main` had given
  `expansion: 'default-locale-only'` is removed, since the shard moves to the
  paged registry. The registry comment's justification is rewritten: see above.
- **`sitemap-xml.ts`** — `main`'s `MAX_SHARD_BYTES` rewrite wins whole;
  `SETTER_URLS_PER_SHARD` keeps this branch's later decision to be its own
  literal rather than an alias of the climbs constant, so tuning one page size
  for a climbs-specific reason cannot silently resize the other.
- **`climb-shards.test.ts`** — kept `main`'s new paged-gate test alongside this
  branch's `find()`-based lookup, which the registry needs now that it holds
  more than one paged shard.
- **`setter-shards.test.ts` mocked `climb-query`**, which the registry stopped
  reading when the climbs shard moved to the materialised store (#4552).
  `shard-registry` imports `climb-store`, whose module body opens a pool, so the
  suite failed to load with `DATABASE_URL environment variable is required`
  before a single assertion ran. It mocks `climb-store` now.
- **Both files' neighbouring-shard fixtures give one healthy item, not zero.**
  Both shards are `expectsUrls`, so a zero-item summary is itself a degrade —
  correct behaviour, and not what either file is testing. `climb-shards.test.ts`
  was reading `setters` in its degraded list for exactly that reason.

## Author verification

Every oracle was mutation-tested: the mutation applied, watched go red, reverted, `git status --porcelain` clean.

- `setter-query.test.ts` — renders the **real** builder's SQL through `PgDialect.sqlToQuery`, never a rebuilt predicate. *Mutations, applied together:* dropped the linkable predicate, the routable-username predicate, the `HAVING` floor, and replaced the row clock with `new Date()` — 6 of 10 tests red, one per rule. `lastModified` is asserted by exact equality against a fixed 2020s date, so a `new Date()` reds without a frozen clock.
- `setter-shards.test.ts` — the paged registry entry, the page slice, and the index entries. *Mutations:* `start = page * PER_SHARD` instead of `(page - 1) * PER_SHARD` (2 red); `expansion: 'all-locales'` (2 red here + 2 in `shard-registry.test.ts`). Also asserts `/sitemaps/setters/4.xml` is a **404**, not an empty `<urlset>` — an empty urlset on an advertised page tells Google the surface was deleted.
- `shard-registry.test.ts` — the existing on-disk walk (`onDisk.length === fixed.size + paged.size`, plus a `[page]/route.ts` existence check per paged directory) already catches leaving `setters.xml/route.ts` orphaned on disk or registering the shard without its route; its `expectsUrls` map and a new all-paged-shards-are-default-locale-only assertion are updated for the move.

Commands:

- `vp run typecheck` — green (`typecheck:web` runs `next build`)
- `vp test run --reporter=agent --project web` — 363 files / 4,000 tests green
- `vp test run --reporter=agent --project i18n` — green (no catalog change in this PR, run anyway because `--project web` does not cover it)
- `vp run check:i18n`, `vp run check:i18n:orphans` — green
- `vp check` — 0 errors, formatting clean

Not verified: no production database or production URL was touched. The cold-cache summary timing is a review lens's measurement (2,999 ms), not one I reproduced — the dev-image container is shared with another agent and I would not restart it to evict its buffers.

Post-rebase, on this box: `app/lib/seo` + `app/sitemaps` + `app/sitemap.xml` +
`app/setter` + `app/__tests__` — **51 files, 680 tests green**.

## Merge gate

**The cold-cache deadline is the thing to watch, and it is tight.** A review
lens measured the setters summary at **2,999 ms** on a genuinely cold cache
against `SHARD_DEADLINE_MS = 3,000 ms` — one millisecond of margin. On
production data the cold `/sitemap.xml` will very likely show
`X-Sitemap-Degraded` naming `setters`, meaning the shard is silently withheld
from the index until an instance's Data Cache warms. **That is the expected
outcome, not a defect and not a blocker** — the climbs shard behaves the same
way today, and the deadline question is tracked in #4583. Deployers reading only
this section should know it in advance rather than diagnose it live.

~~**Do not merge before #4577 is deployed**~~ — **satisfied: #4577 merged 2026-09-06 (`e5893129`)**, and this branch is rebased onto post-merge `main`. The gate existed because submitting tens of thousands of URLs is a one-way crawl-budget commitment, and submitting them against a page that still rendered a spinner is exactly the state #4473 was filed for. What remains before merge is the live check: a setter page serving one `<h1>` and at least one climb anchor on production.

After deploy the check must be a **COLD** `/sitemap.xml` — a request that misses the CDN and lands on an instance with an empty Data Cache — asserting `X-Sitemap-Degraded` is **absent**, not merely that the response is 200. A warm hit proves nothing here: the numbers above say the cold path is where this shard is at risk. If the cold check shows the shard degraded, that is the deadline question above, not a defect in the query.

## Production targets

**`www.boardsesh.com` only.** No `packages/mobile` file changes, no native fingerprint change, no OTA; `app.boardsesh.com` and the native store fleet are unaffected.

### Review round

Rebased onto the reviewed #4577, which now shares one angle-selection rule between this shard and the setter page (28 tier-2 climbs on the dev image were linked at an angle this shard never submits; after the fix, 0 of 85,596 disagree). The submitted-setter count above was re-measured and corrected, and the shard-summary cost section is new.

Refs #4465


marcodejongh:
Review follow-ups applied in 6d737335e:

- **The regex-case hole** — real, and thanks. `.toLowerCase()` made `\S` and `\s` the same string, so the assertion would have passed on the exact inversion (submit *only* the handful of usernames with edge whitespace). The render helper now also returns the raw SQL and the case-sensitive predicate is asserted against that. Mutation-tested: flipping `\S` to `\s` in the builder now reds it, and did not before.
- **`SETTER_URLS_PER_SHARD` coupling** — fixed, it gets its own `10_000` literal. The two being equal today is arithmetic, not a contract.
- **`buildSetterSitemapSql([])`** — fixed. The empty-catalogue guard moved out of `fetchSetterRows` and into the builder, so the exported seam and the production path fail identically and early rather than at the database on a bare `()`. Covered by a new test.
- **Cold-cache timing** — agreed, still unverified and stated as such in the PR body. Worth measuring once #4577 is live.
- **The merge gate stands.** Nothing here changes it: this must not merge before #4577 is deployed and a live setter page is confirmed to serve one `<h1>` and at least one climb anchor.

marcodejongh:
Third round, in 0bc700630.

- **Stale "declared-empty" comments** (`shard-registry.ts` :137, :462) and **`docs/web-reposition.md`'s "three declared-empty ones"** — both fixed.
- **`cacheControl: CLIMB_CACHE_CONTROL`** — agreed, same discipline as `SETTER_URLS_PER_SHARD`. It gets its own `SETTER_CACHE_CONTROL`.
- **`dropped === items.length` serving errors for 6 h** — this was the best find. A drop now **throws** instead of warning and caching: the page 503s, nothing is cached, and the next crawl retries. Caching a list the summary does not describe was the silent-degrade failure mode. Also hardened `setterRowsToItems` against `encodeURIComponent` *throwing* on a lone surrogate — Postgres cannot store one, but "unreachable" is a claim about the database, not a guarantee from that file, and it should not become a 500 on the sitemap route.
- **`resetSetterSitemapCachesForTests` unused / cache untested** — new `setter-item-cache.test.ts`, modelled on `climb-item-cache.test.ts`: concurrent item callers share one scan, concurrent summary callers share one scan, the TTL serves the second caller, a rejection is not memoised, and a drop refuses to cache and recovers on the next call.
- **`setterSummary.hang` never exercised** — new deadline test: a setters summary that never settles is dropped from the index under `X-Sitemap-Degraded: setters` rather than 5xx-ing it. Mutation-tested by removing `withDeadline` from the paged-summary path.
- **Summary asymmetry with the climbs `COUNT(*)`** — now documented at the call site. It is deliberate: this summary has to agree with `setterRowsToItems` *after* the round-trip check, and a count that disagrees by one turns the last page into a 503.

marcodejongh:
Also took the non-blocking double-scan note, in 24f5a96e6.

`cachedSetterSummary` no longer calls `fetchSetterRows` independently — it reads the length of the very list `buildSetterSitemapItems` produces, so the two share one in-process single-flight and the cold burst (`/sitemap.xml` + `/sitemaps/setters/1.xml` together) runs one grouped scan on that instance instead of two in parallel. The Data Cache still stores only the two small values, so the index's per-hit read is unchanged.

It also makes the agreement between them structural rather than conventional, which is the property the last-page 503 depends on. New test pins it, mutation-tested by reverting the summary to its own scan.

marcodejongh:
Fourth round, in daf1f572a.

- **`gyms, gyms`** — good catch, fixed.
- **Missing symmetric summary-failure test** — added, and you are right that it matters more than the item one: the index reads the summary on every hit, so a poisoned entry would keep the setters pages out of `/sitemap.xml` long after the database recovered. Mutation-tested by swallowing the rejection and caching the empty answer.
- **Cold-cache cost** — still the honest open item; stated in the PR body as unmeasured, and the merge gate keeps it behind #4577 going live anyway.

Nothing else outstanding on this PR from my side.

marcodejongh:
Fifth round — rebased onto the reviewed #4577, plus two corrections to this PR's own claims.

**The submitted-setter count was wrong: 24,420 → 26,205.** The earlier figure came from an approximation of the group resolver written in SQL, which skipped `isResolvableGroup`. Feeding the real `getPopularConfigs()` output through the real `resolveClimbSitemapGroups` gives **9** resolvable `(board_type, layout_id)` groups on the dev image, not 12 — `decoy/1`, `kilter/2` and `kilter/7` all fail `tryConstructSlugViewUrl`. 26,205 is the shipped predicate rendered verbatim and run. Still 3 pages.

**The cold-cache cost is now measured, and both suggested remedies were rejected with numbers.** Warm-OS-cache, over 648,657 `board_climbs`: 1,434 / 1,366 / 1,370 ms for the full grouped scan, and 1,409 / 1,354 / 1,439 ms for the same predicate wrapped in `count(*)`. **Identical** — the grouped scan is the entire bill, so turning the summary into a `COUNT(*)` buys nothing and would give up the structural guarantee that the item list and the summary describe the same set (`setterRowsToItems` throws on any drop, so on success `items.length` equals the row count exactly). `EXPLAIN` shows `Heap Blocks: exact=42354, Buffers: shared hit=562 read=43964` on every run — the array-containment predicates force the heap access.

A review lens measured 2,999 ms genuinely cold, which I did not reproduce (the dev-image container is shared with another agent and I would not restart it to evict its buffers). Taken at face value with production being materially larger, a cold `/sitemap.xml` will likely miss `SHARD_DEADLINE_MS = 3_000` and drop this shard. `withDeadline` does not cancel, so the abandoned query still fills both caches and the next request past the 60 s degraded window gets the shard — self-healing per instance, not a lost shard, and the same posture the climbs summary already ships with (~9.2 s cold on this image).

Raising the deadline would be a change to shared sitemap infrastructure, and the #4493 ruling is explicit that the answer to a summary that does not fit is caching, not a bigger budget. So it is **not** changed here and is filed as #4583 with the three options costed. @marcodejongh — the call worth making is whether paged summaries should get their own deadline, given that being impatient now systematically drops the two biggest shards rather than one slow one.

The merge gate is tightened accordingly: the post-deploy check must be a **cold** `/sitemap.xml` — a CDN miss onto an instance with an empty Data Cache — asserting `X-Sitemap-Degraded` is absent. A warm hit proves nothing here.

Nothing in this PR's diff changed this round; the rebase picks up the shared angle rule from #4577, where 28 tier-2 climbs were being linked at an angle this shard never submits (now 0 of 85,596).

All numbers are dev image; production reconciliation is outstanding and no production database was read.

marcodejongh:

Every dev-image figure in the body re-derived against production by running **this branch's own `buildSetterSitemapSql(groups)`** — rendered by `runSetterSitemapQuery`, executed against a read-only production connection, and the rows fed through the real `setterRowsToItems` and `renderUrlset`/`expandDefaultLocaleOnly`. Not the issue's SQL, and not a rebuilt predicate.

### The shard gets a fourth page

|---|---|---|---|
| setters with ≥1 visible climb (the page's own 404 rule) | 91,946 | **125,082** | +33,136 |
| setters with ≥3 visible climbs, any board | 35,950 | **49,367** | +13,417 |
| **submitted: linkable and ≥3 linkable visible** | **26,205** | **31,039** | **+4,834** |
| **pages at `SETTER_URLS_PER_SHARD` = 10,000** | **3** | **4** | **+1** |
| largest submitted setter, linkable visible climbs | 4,723 | **5,063** | +340 |
| `board_climbs` rows | 648,657 | **885,528** | +236,871 |
| setters dropped by the routable-username predicate | not stated | **10** (31,049 → 31,039) | — |
| usernames dropped by the JS round-trip check | 0 | **0** | 0 |
| largest page, rendered bytes | not stated | **1,235,782** (31% of `MAX_SHARD_BYTES` = 4,000,000) | — |

**`/sitemaps/setters/4.xml` is a real page in production, not a 404.** Pages come out 10,000 / 10,000 / 10,000 / 1,039. Nothing in the diff breaks — `setter-shards.test.ts` derives its page count from the fixture's `itemCount` rather than a literal, and the "404s the page past the end" oracle still holds at page 5 — but the body's "3 pages at 10,000" is wrong and the post-deploy check should expect four.

No cap is exceeded: 4 pages × 10,000 default-locale-only URLs, largest page 1.24 MB against the 4 MB response budget, and the paged shards never touch `MAX_URLS_PER_SHARD` (45,000).

The routable-username predicate is load-bearing in production, not theoretical: 31,049 setters clear the linkable + ≥3 gate and 10 of them are excluded by the whitespace / `[/?#]` / control-character rules. Reconstructing the same predicate independently and comparing (31,049 without the routability rules, 31,039 with) is also what proves the reconstruction below matches the shipped builder.

One setter username in production contains a `%`. It passes the routable predicate (correctly — `%` is not in `[/?#]`) but has fewer than 3 linkable visible climbs, so it is **not** submitted. Its only exposure is the page-level `URIError` #4577 fixes.

### The deadline prediction is now a measurement

The body says: *"production `board_climbs` is materially larger than the dev image, so a genuinely cold `/sitemap.xml` will very likely miss the 3,000 ms deadline and drop this shard."*

Measured, warm, on production: the full grouped scan returning all 31,039 rows took **6,411 ms** — **2.1× `SHARD_DEADLINE_MS`** — from a client with a ~260 ms round-trip floor, so roughly 6.1 s of database work. That is not the cold path; that is the warm one. Dev image, for comparison: 1,434 / 1,366 / 1,370 ms.

So the prediction should be restated as a fact: **on production data this shard misses the index deadline on every Data Cache miss, warm or cold.** The self-healing argument in the body still holds (`withDeadline` does not cancel, both caches fill, the next request past the 60 s `DEGRADED_CACHE_CONTROL` window gets the shard), and so does the ruling not to raise the deadline here — but "very likely" can come out.

For context, the climbs summary on the same production database measures ~3.0 s at 3 lanes on today's `main` and ~5.3 s after #4578, so this is the same posture the climbs shard already ships with rather than a new one. Numbers on #4578. #4583 is the right home for the question of whether paged summaries get their own deadline; the case for it is now measured rather than argued.

### The `<lastmod>` caveat resolves in the PR's favour

The body says the dev image "cannot tell you whether the per-setter spread is meaningful in production — only that the value is non-null on all 648,657 rows and parses", because every visible row's `updated_at` fell inside one 23-minute window on 2026-08-10.

Production, across the 31,039 submitted rows: `min` **2026-07-07T02:11:45Z**, `max` **2026-08-19T21:13:54Z**, **44 distinct calendar days**. So `max(board_climbs.updated_at)` is a genuinely spread clock in production — not a per-content clock (the sync-touch caveat stands), but not a single job stamp either. The choice of `board_climbs.updated_at` over `board_setter_stats.updated_at` is vindicated: the alternative really would have collapsed all 31,039 to one nightly timestamp.

### SQL

The nine groups `resolveClimbSitemapGroups` resolves in production are `decoy/2` size 1 sets `{2…20}` · `grasshopper/1` size 2 `{1,2,3}` · `kilter/1` size 10 `{1,20}` · `kilter/8` size 25 `{26,27,28,29}` · `soill/1` size 1 `{1}` · `tension/9` size 1 `{8,9,10,11}` · `tension/10` size 6 `{12,13}` · `tension/11` size 6 `{12,13}` · `touchstone/1` size 1 `{1}` — same nine as the dev image. `buildSetterSitemapSql` renders to exactly this:

```sql
SELECT
  setter_username,
  count(*)::int AS climb_count,
  to_char(max(updated_at), 'YYYY-MM-DD"T"HH24:MI:SS.MS"Z"') AS last_modified
FROM board_climbs
WHERE is_listed = true
  AND is_draft = false
  AND (
    setter_username IS NOT NULL
    AND setter_username <> ''
    AND setter_username ~ '^\S(.*\S)?$'
    AND setter_username !~ '[/?#]'
    AND setter_username !~ '[\x00-\x1F\x7F]'
  )
  AND (
       (board_type = 'decoy'       AND layout_id = 2  AND compatible_size_ids @> ARRAY[1]::int[]
          AND required_set_ids <@ ARRAY[2,3,4,5,6,7,8,9,10,11,12,13,14,15,16,17,18,19,20]::int[])
    OR (board_type = 'grasshopper' AND layout_id = 1  AND compatible_size_ids @> ARRAY[2]::int[]
          AND required_set_ids <@ ARRAY[1,2,3]::int[])
    OR (board_type = 'kilter'      AND layout_id = 1  AND compatible_size_ids @> ARRAY[10]::int[]
          AND required_set_ids <@ ARRAY[1,20]::int[])
    OR (board_type = 'kilter'      AND layout_id = 8  AND compatible_size_ids @> ARRAY[25]::int[]
          AND required_set_ids <@ ARRAY[26,27,28,29]::int[])
    OR (board_type = 'soill'       AND layout_id = 1  AND compatible_size_ids @> ARRAY[1]::int[]
          AND required_set_ids <@ ARRAY[1]::int[])
    OR (board_type = 'tension'     AND layout_id = 9  AND compatible_size_ids @> ARRAY[1]::int[]
          AND required_set_ids <@ ARRAY[8,9,10,11]::int[])
    OR (board_type = 'tension'     AND layout_id = 10 AND compatible_size_ids @> ARRAY[6]::int[]
          AND required_set_ids <@ ARRAY[12,13]::int[])
    OR (board_type = 'tension'     AND layout_id = 11 AND compatible_size_ids @> ARRAY[6]::int[]
          AND required_set_ids <@ ARRAY[12,13]::int[])
    OR (board_type = 'touchstone'  AND layout_id = 1  AND compatible_size_ids @> ARRAY[1]::int[]
          AND required_set_ids <@ ARRAY[1]::int[])
  )
GROUP BY setter_username
HAVING count(*) >= 3
ORDER BY setter_username ASC;
-- → 31,039 rows in 6,411 ms; pages = ceil(31039 / 10000) = 4
```

The ≥1-visible and ≥3-visible-any-board rows in the table come from the single grouped scan quoted on #4577. `board_climbs` row count is `SELECT count(*) FROM board_climbs`.

### Merge gate

The gate in the body still stands as written; two amendments:

1. The cold `/sitemap.xml` check will very likely show `X-Sitemap-Degraded` **present** for `setters` — that is now the expected outcome on production data, not evidence of a defect, and it is the same behaviour the climbs shard has today. Do not treat it as a blocker; take it to #4583.
2. Verify `/sitemaps/setters/4.xml` returns a populated `<urlset>` (~1,039 URLs) and `/sitemaps/setters/5.xml` returns 404.

marcodejongh:

Conflicting: the same conflict as blocked #4577 (packages/web/app/api/og/setter/route.tsx) — this branch looks stacked on/sharing history with it, so resolving #4577 first would likely clear this one too.

marcodejongh:
All three checked out. Two were mine; pushed in `5007e28f`.

**1. `docs/sitemap.md` registry table — fixed.** It listed `setters` under `SHARD_REGISTRY` with `climbs` alone under `PAGED_SHARD_REGISTRY`, which this PR makes wrong. Both cells corrected, the file-shape row generalised from `/sitemaps/climbs/1.xml` to `/sitemaps/<id>/1.xml`, and the reason `setters` is paged now sits where a reader meets the table — volume (~108k pairs vs 11,250), not locale expansion.

**2. Cold-cache deadline — now leads the merge gate.** The 2,999 ms against `SHARD_DEADLINE_MS = 3,000 ms` was buried in a "not verified" note. The gate section now states it up front, says the cold `/sitemap.xml` will very likely show `X-Sitemap-Degraded: setters` on production data, and that this is the expected outcome rather than a blocker — same behaviour the climbs shard has today, tracked in #4583.

**3. The `#4648` reference — you were right, and it was the interesting one.** I checked rather than assumed: `#4648` is the Vercel-to-Railway epic, and it is the *correct* citation everywhere else in these files (the response ceiling going away, the byte guards, `maxDuration`). The one line I wrote credited it with giving every shard its own `expansion`, which came from **#4996** instead. Corrected to #4996.

For the record, the wrong attribution existed only on this branch — `git grep` over `main` and the two sibling branches finds no other instance.

marcodejongh:
Four points. Two taken, one already ruled on, one now satisfied.

**Cold-cache deadline — already ruled on, not re-litigating.** Marco's own comment on this PR settled it: the cold `/sitemap.xml` showing `X-Sitemap-Degraded: setters` is *"the expected outcome on production data, not evidence of a defect, and it is the same behaviour the climbs shard has today. Do not treat it as a blocker; take it to #4583."* Calling it "ships non-functional" overstates it — the shard is absent on a cold instance and present once warm, which is the climbs shard's shipped behaviour too. It leads the merge gate so a deployer meets it before diagnosing it live.

**Merge dependency on #4577 — satisfied.** #4577 merged this morning (`e5893129`); this branch is rebased onto post-merge `main` (its own eight commits `--onto`, since #4577 was squash-merged and a plain rebase replays them as add/add conflicts). Gate section updated to say so, with the live-page check left as the remaining item.

**`unstable_cache` error behaviour — half right, and the interesting half is the other one.** You're correct that the two "does not memoise a failure" cases cover the in-process layer only, since the test mocks `unstable_cache` as a pass-through. That is now stated in the file.

But the hypothetical does not hold on the version we ship. Read out of the installed **next@16.3.2** (`server/web/spec-extension/unstable-cache.js`) rather than assumed: the callback's result is `await`ed **before** `cacheNewResult` runs, so a rejection throws past the write and never reaches the Data Cache; the background-revalidation path logs the error and returns the *stale* value rather than storing the rejection. A failed scan cannot poison either layer for the `SUMMARY_REVALIDATE_SECONDS` window. Recorded in the mock's comment with a note to re-check on the next Next major.

**No per-page `lastmod` — correct, and now recorded as a decision.** The climbs shard affords per-page timestamps by reading them off the materialised `sitemap_climb_urls` store. Setters has no such store, so the equivalent is a second grouped scan over `board_climbs` — the scan already tight against `SHARD_DEADLINE_MS`. Index entries carry the shard-level `lastModified` until #4583 changes the economics; the registry entry now says why.

Pushed in `148e034c`.

marcodejongh:
You found a real inconsistency and attributed it to the wrong side. Fixed — in the body, not the code.

**#4996 does exist.** It is `fix(web): stop climb pages advertising three locale twins of themselves`, **merged 2026-09-01**, and `git`'s own commit→PR mapping puts `01387f422` — the commit that introduced `ShardExpansion` — in it. So the inline comment in `shard-registry.ts` was right and the PR body was carrying my earlier mis-attribution. The body now credits **#4996** for the per-shard `expansion` and keeps **#4648** for the half it really covers: the Vercel response ceiling going away with the Railway migration. Both references now point at the thing that actually did the work.

Worth flagging the check that settles this in one command, since this is the second review round it has come up in:

```
gh api repos/boardsesh/boardsesh/commits/01387f422/pulls -q '.[].number'   # → 4996
```

**`unstable_cache` — this was answered on the previous round, with the source read.** From the installed **next@16.3.2**, `server/web/spec-extension/unstable-cache.js`: the callback's result is `await`ed *before* `cacheNewResult` runs, so a rejection throws past the write and never reaches the Data Cache; the background-revalidation path logs and returns the *stale* value rather than storing the rejection. "If Next ever persists a rejection" is not a property of the version we ship, and the file now says so with a note to re-check on the next major. The half you were right about — that the pass-through mock leaves those two cases covering the in-process layer only — is stated in the same comment.

**Per-page `lastmod`** — agreed on all counts, and already recorded in the registry entry as the deliberate trade-off it is.

marcodejongh:
## Merge gate satisfied — verified on production just now

#4577 is deployed. The live check this PR was gated on:

| check | before | now |
|---|---|---|
| `/setter/KilterStudio` | 200, **0** `<h1>`, **0** links, 379 chars of pure chrome | **200**, **1** `<h1>`, **50** crawlable `/view/` links, **3,246** chars |
| `/setter/zzz-not-a-real-setter-98765` | 200, indexable, self-canonical | **404** |
| `/setter/aaaaaaaaaaaa` | 200 | **404** |
| the 404's robots meta | — | `noindex` |
| `/setter/KilterStudio?page=5000` | 200 | **404** |

So the soft-404 farm is closed and the page a submitted URL points at now has real content to earn its crawl. That was the precondition for shipping ~10 pages of setter URLs, and it holds.

Taking this out of draft: CI is green, the review rounds are addressed, and it is `MERGEABLE` against current `main`.

One expectation to carry into the deploy, unchanged from the gate section: the first cold `/sitemap.xml` after this merges will very likely carry `X-Sitemap-Degraded: setters`, because the summary scan sits within a millisecond of `SHARD_DEADLINE_MS` on a cold cache. That is the climbs shard's shipped behaviour too, it self-heals once warm, and it belongs to #4583 — not a reason to hold the merge.

marcodejongh:
## Codex's P1 is real. Back to draft.

I marked this ready an hour ago on the production gate; that was premature. Verified all three findings against the code — reverting to draft so it cannot merge while the first one stands.

### P1 — eligibility is catalogue-scoped, the noindex is page-scoped

| | scope |
|---|---|
| `setter-query.ts:138-139` | `count(*)` over **all** the setter's visible climbs matching `linkable`, `HAVING >= SETTER_MIN_VISIBLE_CLIMBS` (3) |
| `server-setter-data.ts:179-181` | page 1 = top **50 by ascents**, `visibleSetterClimbsWhere` only — **no linkable predicate** |
| `page.tsx:139,146` | `!setterPageHasCrawlableClimb(username, page)` → `NOINDEX_FOLLOW`, and that check is **page-scoped** |

So a setter with ≥50 high-ascent climbs on unresolvable configurations plus ≥3 low-ascent linkable ones passes the shard's floor, gets submitted, and renders a page 1 with zero crawlable `/view/` anchors — which then declares `noindex, follow`. The shard advertises a URL that refuses indexing. That is precisely the own-goal this shard was built to avoid.

**The code says otherwise, and the code is wrong.** `page.tsx:136-137` claims "`/sitemaps/setters` already refuses to submit them (linkable AND ≥3 climbs)". It refuses on the *catalogue*, not on *page 1*. I wrote that comment believing the two filters agreed; they do not.

Not hypothetical-only in shape, but I cannot size it — it needs a dev database I do not have here, and the count depends on how many setters have ≥50 unlinkable climbs ahead of their linkable ones.

### P2 — `<lastmod>` understates change (real)

`max(updated_at)` reads `board_climbs` only, and only over rows that already passed `linkable`. The page renders ascent counts, grades and quality from `board_climb_stats` and orders on them, and it renders *non-linkable* visible climbs too. Both can change the rendered page without moving this clock.

### P2 — dot-segment usernames (real)

`routableUsername` (`setter-query.ts:93-99`) rejects empty, whitespace-edged, `/?#` and control characters — but `.` and `..` pass, and `encodeURIComponent` leaves them unchanged. `/setter/.` normalises to `/setter/` and `/setter/..` to `/`, so the entry never reaches the route. One extra clause fixes it; I will fold it into whichever P1 fix lands.

### What I am not doing unilaterally

The P1 fix is a choice about **what gets submitted**, and every option changes the submitted-set size the body quantifies. Asking rather than picking.

marcodejongh:
All three fixed in `db5f5828` + `f8d446dd`. The P1 was a real defect and the sharpest finding on this PR — thank you.

**P1 — eligibility now checks the same page the `noindex` reads.** The query carries every *visible* climb through a CTE, ranks each setter's climbs the way the page ranks them, and applies two conditions:

- `count(*) FILTER (WHERE is_linkable) >= 3` — the catalogue floor, "worth a crawl budget"
- `count(*) FILTER (WHERE is_linkable AND page_position <= 50) >= 1` — page one can actually be indexed

The ranking reproduces the page's sort key without its correlated per-row angle pick: the page joins stats at `mostAscendedAngle`, chosen by `ascensionist_count desc nulls last` over publishable angles, so the value it lands on **is** the max across those angles — and `max()` skipping nulls is the same "nulls last". One aggregate instead of a `LIMIT 1` per row.

**The comment you'd have been entitled to trust was wrong too.** `page.tsx:136` claimed "`/sitemaps/setters` already refuses to submit them (linkable AND ≥3 climbs)". It refused on the catalogue, never on page one. Corrected to state the real two-part condition.

**P2 `<lastmod>` — fixed here rather than deferred.** Now `GREATEST` of the climb clock and the stats clock across **every visible climb**, with linkability kept as an eligibility condition rather than a filter on the clock. Your framing was exactly right: ascent counts, grades, quality and ordering all come from `board_climb_stats`, and non-linkable visible climbs render too.

**P2 dot-segments — fixed.** `AND setter_username !~ '^[.]{1,2}$'`. They cleared every existing rule and survive `encodeURIComponent` unchanged.

Both constants moved to a new `setter-page-contract.ts`, since the bug belonged to neither module: each was correct in isolation.

Mutation: dropping the page-one condition reds the new test, twelve others green. Worth recording that my *first* attempt at that mutation silently no-op'd on a whitespace mismatch and the suite "passed" — a mutation that fails to apply is indistinguishable from an assertion that does not bite, so the anchor is asserted before the write now.

Back to draft while CI runs; I will take it out again once green.

marcodejongh:
Verified and recorded in `c0d1cba1`. `SUMMARY_CACHE_TAG` is tagged on the `unstable_cache` call and nothing calls `revalidateTag` with it — true here and on the climbs shard, which has the identical shape.

The part worth writing down is not the unused tag but what it does to whoever uses it: `revalidateTag` clears the Next Data Cache, while `cachedSummary` sits in front of it as an in-process TTL that knows nothing about tags. A tagged revalidation would be invisible to that instance for up to six hours and the surface would keep serving the old summary — presenting as a revalidation that silently did nothing. The note lives at the constant, where someone wiring a trigger will be reading.

CI is running on the rebase plus this; taking it out of draft once green.

marcodejongh:
Both new notes taken in `b772d549`; the other two were already handled.

**`climb_count` — removed.** Selected on every scan and declared on the row type, never read: the floor it reports is enforced in the `HAVING`. The two test fixtures that carried it only to satisfy the type went with it.

**No `enabled` gate — deliberate, and now stated in the registry entry.** `climbs` has one because that shard was *paused in production* once (#4648) and needed a lever that did not require a deploy. This one has never shipped, so there is nothing to hold back, and the rollback is a revert: the shard is derived — writes nothing, owns no table, and dropping it from the registry withdraws every URL it published on the next crawl. My read is that a gate is worth adding the day it needs pausing without a code change, not before — happy to add one if you would rather have the lever on day one.

**`SUMMARY_CACHE_TAG`** — recorded a commit earlier (`c0d1cba1`), including the part that bites: `revalidateTag` clears the Data Cache but not the in-process TTL in front of it.

**Cold-cache deadline** — unchanged, tracked in #4583, and it leads the merge gate so a deployer meets it before diagnosing it live.

marcodejongh:
Two taken in `13783e9a`; the first is the standing ruling.

**C1 controls — real, and the reasoning about *why* is the useful part.** Verified in node: `encodeURIComponent('\x85')` → `%C2%85`, and `decodeURIComponent` round-trips it, so `setterRowsToItems` sees a valid username and keeps it. That is what separates C1 from a lone surrogate, which throws and gets dropped by the JS guard. Postgres `text` in UTF-8 stores C1 happily, so this predicate really is the only thing standing between one and a sitemap entry. Class extended to `[\x00-\x1F\x7F\x80-\x9F]`, with a test asserting the raw SQL and a mutation (narrow it back to `\x7F]`) that reds it.

**Two re-export chains — mine, and removed.** I added the `setter-query.ts` re-export in the same commit that created `setter-page-contract.ts`, which handed each constant two paths to drift along — precisely what the contract module exists to prevent. The test reads them from the owning module now.

**Cold-cache deadline — settled, and I am not re-opening it.** This is its fourth appearance across three review rounds. Marco ruled on this PR: the cold `X-Sitemap-Degraded: setters` is *"the expected outcome on production data, not evidence of a defect... Do not treat it as a blocker; take it to #4583."* On your sharper framing — steady-state versus launch transient — it is steady-state on cold instances, self-healing once warm, and identical to the climbs shard's shipped posture. That is stated at the top of the merge gate so a deployer meets it before diagnosing it live.

marcodejongh:
All four taken in `2f2cf207`. The `\S` one was the best catch on this PR after the P1 — it was not just a wrong comment.

**`\S` is ASCII-only — and the gap is real.** I measured it against the shipped Postgres (UTF8 / en_US.utf8) rather than reasoning about it:

```
chr(160) ~ '\s'                    -> false
chr(160) ~ '[[:space:]]'           -> false
chr(160) || 'marco' ~ '^\S(.*\S)?$' -> TRUE
```

So a non-breaking space really is `\S` to that engine: the name clears rule 1 and ships as `%C2%A0marco`. Added a clause rejecting Unicode whitespace and zero-width characters anywhere — NBSP, OGHAM, EN/EM QUAD…ZWSP, LS, PS, NNBSP, MMSP, IDEOGRAPHIC SPACE, BOM — and verified against the same database that all are rejected while `marco` and `mar co` still pass. Also checked that Postgres ARE supports `\uXXXX` before relying on it, so the class stays readable rather than carrying invisible literals.

**Exact-multiple pagination — added, and it bites.** At a multiple of the page size, `ceil()` and `floor() + 1` agree everywhere except the last page, where the latter advertises one that renders an empty `<urlset>`. Mutation: swap `ceil` for `floor + 1` → only the new case reds, the other five pass. That is the proof the 25,000 fixtures could not reach it.

**"All six are slow" — both halves.** `setterSummary` has a `delayMs` now and is slowed with the rest, and the test asserts `/sitemaps/climbs/` and `/sitemaps/setters/` appear, since a registry that dropped the paged shards satisfied every fixed-shard assertion in it.

**`SUMMARY_CACHE_TAG` — exported.** A forward hook nobody can import is not a hook.

marcodejongh:
Taken in `dae09e86`, though you called it acceptable.

You were right that a grep for `SETTER_PAGE_SIZE` landed on a re-export rather than the source — and it was the same shape I had removed from `setter-query.ts` one commit earlier, just left behind for consumers that no longer needed it. `setter-json-ld.tsx` and the test read it from the contract module now, the local alias is gone, and so is the compatibility re-export. One name, one source, which is the entire reason that module exists; keeping a hop "for the migration" would have been the migration not finishing.

The cold-cache deadline stays as ruled: tracked in #4583, same posture as the climbs shard, and stated at the top of the merge gate so the cold check is run before declaring success rather than a warm hit.

One stale detail in your summary, harmless: the MoonBoard note says the group-based assertion "auto-detects when #4493 fixes the parser". #4493 already merged (#4578) — the parser is fixed and MoonBoard is in the sitemaps, which is why that assertion resolves through the real groups today.

marcodejongh:
## Ready again — five review rounds later

CI green on `dae09e86`, `MERGEABLE` against current `main`, no unaddressed comments. The production merge gate is unchanged and still holds: #4577 is deployed, `/setter/KilterStudio` serves one `<h1>` with 50 crawlable `/view/` links, and unknown setters 404.

What the rounds after the first un-draft actually changed — recording it because "ready" now means something different from what it meant at 06:42:

| finding | outcome |
|---|---|
| **P1** eligibility catalogue-scoped, `noindex` page-scoped | **Real defect.** Query rewritten: catalogue floor decides worth-crawling, a page-one check decides indexability. The shard could have submitted URLs that refuse indexing. |
| `\S` "covers every whitespace class" | **Wrong, and it hid a gap.** Measured on the shipped Postgres: `chr(160) ~ '\s'` is false, so NBSP shipped as `%C2%A0marco`. New Unicode-whitespace clause. |
| C1 controls (U+0080–U+009F) | **Real gap** — they round-trip cleanly through `encodeURIComponent`, so the JS guard could not catch them. |
| `.` / `..` usernames | **Real** — survive encoding, eaten by URL normalisation. |
| exact-multiple pagination untested | **Real boundary** — `ceil` vs `floor+1` diverge into an empty trailing `<urlset>`. |
| "all six are slow" could not slow setters, never asserted paged shards | **Real hole** — a registry dropping them passed every assertion. |
| `climb_count` selected, never read | removed |
| two/three import chains for the contract constants | collapsed to one |
| `SUMMARY_CACHE_TAG` unexported forward hook | exported, with the two-layer eviction trap documented |
| cold-cache deadline | ruled expected, tracked in #4583, leads the merge gate |
| no `enabled` gate | deliberate, and now says why — the rollback is a revert of a derived shard |

I marked this ready once before the P1 was found. That was premature, and the lesson is on the record rather than tidied away: a production gate passing is not the same as the code being right.

PR #4591, opened 2026-08-19
fix(mobile): report an exhausted sqlite-init retry chain even when superseded

marcodejongh:
## Summary

Fixes a Sentry reporting gap in the mobile SQLite init code: `beginInitialization` was suppressing crash reports for any retry chain flagged as `superseded`, on the assumption that superseded meant the failure was just a harmless remount artefact. In reality, a chain can burn its entire retry budget on genuine `SQLITE_BUSY` lock failures and only become "superseded" because a remount happened to land during the final attempt — that case left offline storage disabled for the session with zero Sentry signal. The fix reports these exhausted-and-superseded cases too, tagging them `superseded: 'true'` so they stay distinguishable from the harmless `superseded: 'false'` population instead of being dropped or merged together. No changes to init ordering or the retry logic itself.

---

Refs #4314

## Summary

`beginInitialization` gave up on a retry chain and then decided whether to tell Sentry with `if (!superseded)`. That guard exists to keep a lifecycle artefact — a throw against the connection `SQLiteProvider` closed on a remount — out of the `kind:sqlite-init` aggregate, which is the exact aggregate #4314 reads to judge whether the lock problem is fixed.

But `superseded` at the give-up point does not mean the failure was an artefact. A non-lock throw against a superseded handle is refunded earlier in the loop and never reaches here. What does reach here is a chain that spent its **whole** budget on genuine `SQLITE_BUSY` failures and happened to have a remount land during the last one. That is a fully exhausted 30s window: offline storage is off for the rest of the session, and Sentry heard nothing about it.

The report now fires in that case too, tagged `superseded: 'true'` (a string, since Sentry tag values are strings). The clean lock signal stays readable as `superseded:false`, so the two populations remain separable rather than merged into one aggregate.

Nothing about the init ordering or the retry chain itself changed — the lock was fixed by #4121 and #4324.

## Test plan

- New `still reports an exhausted window when a remount lands during the final attempt` in `packages/mobile/src/db/__tests__/connection.test.ts` — the lock twin of the existing superseded cases. Five consecutive `SQLITE_BUSY` failures with the remount landing during attempt 5; asserts one report carrying `sqlite_code: 5`, `superseded: 'true'`, `attempts: 5`, `retryable: true`. It fails on the old code, where the report is dropped entirely.
- The pre-existing exhaustion test now also pins `superseded: 'false'`, so the clean population keeps its tag.
- `vp test run --project mobile --reporter=agent src/db/__tests__/connection.test.ts` — 30 passed.
- `vp test run --project mobile --reporter=agent` (full suite) — two unrelated timing flakes under parallel load (`theme-provider`, `queue-provider-leave-session`); both pass when re-run on their own.
- `vp check` on both changed files — clean.

## Release Notes

none


marcodejongh:
Re: the fake-timers note — covered. The `describe('initializeDatabase lock contention (#4104)')` block already has `afterEach(() => { vi.useRealTimers(); ... })` (connection.test.ts, alongside the tmpdir cleanup), which every test in the block including the new one runs through. No leak into later tests.

marcodejongh:
Cross-reference: **#5355** is now open against #4314 too, and is orthogonal to this one.

This PR makes an exhausted-and-superseded chain visible. #5355 addresses why it exhausts: on the 2.4.0 fleet the failures are `phase: migrations`, `retryable: true`, burning 29,875 ms of a 30,000 ms window — a pending `ALTER TABLE` from migration v5 (`b941bbdd8`, 2026-09-07) contending on first launch after the update. It extends the retry ladder (30 s, then 60 s × 3) for lock-classified failures only.

That agent also probed every init-path statement against real SQLite with a second connection holding an `IMMEDIATE` write txn, and disproved the standing "a connection skipped `busy_timeout`" theory — the migration DDL waited its full timeout before throwing. Worth knowing here, since it narrows what your reporting will actually surface.

They compose, but both touch reporting around an exhausted chain. Whoever merges second should rebase and confirm the two signals stay distinguishable.

Issue #4590, opened 2026-08-19
generateSlugFromText is ASCII-only, so non-Latin climb names never get a real slug


marcodejongh:
`generateSlugFromText` (`packages/shared/play-view/src/readable-url-utils.ts`) strips characters with `/[^\w\s-]/g`, and JS `\w` is ASCII-only (`[A-Za-z0-9_]`). A climb named entirely in Cyrillic, CJK, or emoji slugs to `''`, so every URL builder that calls it falls back to the bare uuid instead of a readable slug.

That's the root cause behind #4410 (fixed there by guarding the redirect instead — see that PR for why: making this function Unicode-aware would change the canonical slug for every already-indexed climb whose name currently slugs to non-empty ASCII but contains other scripts too, which is an SEO regression outside that fix's scope).

Follow-up: make the slug generator Unicode-aware (e.g. transliterate where practical, or keep non-Latin scripts as-is via a broader character class / `Intl`-based normalization) so climbs with non-Latin names get a real slug instead of falling back to the bare uuid. Needs a plan for the canonical-URL migration of existing indexed climbs (redirects from the old bare-uuid canonical to the new slugged one) so this doesn't itself cause a wave of 404s/deindexing.

PR #4592, opened 2026-08-19
fix(web): stop the climb view redirecting to itself for non-Latin names

### LINKED ISSUE #4410: Climb view 308s to itself when the climb name slugs to empty (non-Latin names)
Found during #3823's review; pre-existing and not introduced there.

`packages/web/app/[board_name]/.../[angle]/view/[climb_uuid]/page.tsx` redirects when `isNumericFormat || isUuidOnly(params.climb_uuid)`. The redirect target's climb segment comes from `buildClimbSegment` (`@boardsesh/play-view`), which returns the **bare uuid** when `generateSlugFromText(name)` is empty — and JS `\w` is `[A-Za-z0-9_]`, so a climb named entirely in Cyrillic, CJK, or emoji slugs to `''`.

Request `…/view/<BARE_UUID>` for such a climb → `isUuidOnly` true → target's climb segment is the same bare uuid → the target URL equals the request URL → `permanentRedirect` to itself → `ERR_TOO_MANY_REDIRECTS`.

Guard: skip the redirect when the built URL equals the current path (or fall through to render when the climb slug is empty). Note #3823 made the qualified size slug exact, which removed an accidental escape this loop previously had on the shadowed Kilter size — the loop itself predates that change and affects every size.

https://claude.ai/code/session_014TUjwDeVASG8eS25ArpyfF

marcodejongh:
## Summary

`ClimbViewPage` (config-tuple `/[board_name]/.../view/[climb_uuid]`) always called `permanentRedirect(newUrl)` when `needsSlugRedirect` was true, without ever checking whether `newUrl` was actually different from the URL that was requested.

`generateSlugFromText` (`@boardsesh/play-view`) strips non-word characters with `/[^\w\s-]/g`, and JS `\w` is ASCII-only. A climb whose name is entirely Cyrillic/CJK/emoji slugs to `''`, so both `tryConstructSlugViewUrl` and its `constructClimbViewUrlWithSlugs` fallback emit the bare uuid instead of a `slug-uuid` segment. For a bare-uuid request (`isUuidOnly` true) to such a climb, the redirect target ends up identical to the request path, so the page 308s to itself — `ERR_TOO_MANY_REDIRECTS` for the visitor and, at sitemap scale, a redirect loop Google will never resolve into an index.

Fixed by comparing the built `newUrl` against the raw requested path (built from `params`, not the normalised `parsedParams`) and only calling `permanentRedirect` when they differ; otherwise the page falls through and renders normally. This is the general fix — not an empty-slug special case — so it also covers any other way the two builders could land on the request's own path.

I checked the other view route, `/b/[board_slug]/[angle]/view/[climb_uuid]/page.tsx`, for the same bug. It does **not** have this issue: that page never redirects at all — it renders directly for whatever `board_slug`/`climb_uuid` it's given, and only points *other* pages' canonicals into the config-tuple tree (see its `generateMetadata` A1 comment). No change needed there.

Deliberately **not** making `generateSlugFromText` Unicode-aware here — that would change the canonical slug for already-indexed climbs with non-Latin names, which is an SEO regression out of scope for this fix. Filed as a follow-up: #4590.

Closes #4410

## Release Notes

Only affects a climb view URL that was already broken (an infinite-redirect 500-adjacent error page); nothing a climber would describe as a feature getting better.

## Test plan

- Added a regression test (`non-Latin climb name self-redirect`) that requests a bare-uuid URL for a climb named `Скала` (Cyrillic) with slug-format layout/size/set segments matching what the builders would themselves produce, and asserts `permanentRedirect` is never called and the page renders.
- Kept and extended the existing `legacy numeric climb-view redirect` test, which still asserts the *normal* redirect path fires for a climb with a real (Latin) name/slug — confirming the equality guard doesn't suppress genuine redirects.
- `vp test run --project web redirect.test.tsx --reporter=agent` — 11 tests / 3 files passed.
- `vp check --fix` — clean.
- `vp run typecheck` — clean.

### DIFF
--- a/packages/web/app/[board_name]/[layout_id]/[size_id]/[set_ids]/[angle]/view/[climb_uuid]/__tests__/redirect.test.tsx
+++ b/packages/web/app/[board_name]/[layout_id]/[size_id]/[set_ids]/[angle]/view/[climb_uuid]/__tests__/redirect.test.tsx
@@ -1,4 +1,8 @@
-import { describe, expect, it, vi } from 'vite-plus/test';
+import { beforeEach, describe, expect, it, vi } from 'vite-plus/test';
+// Real (unmocked) slug builders, used to compute the expected self-redirect
+// URL the same way `constructClimbViewUrlWithSlugs` does — rather than
+// hardcoding a slug string that would silently drift from the real formatter.
+import { generateLayoutSlug, generateSizeSlug, generateSetSlug } from '@/app/lib/url-utils';
 
 // Regression for the legacy layout hijacking numeric climb-view URLs: the
 // board `[angle]` layout used to 308 every numeric URL to the bare slug LIST
@@ -7,6 +11,10 @@ import { describe, expect, it, vi } from 'vite-plus/test';
 // the slug VIEW URL that preserves the climb uuid — is the one that runs.
 
 const CLIMB_UUID = 'abcdef1234567890abcdef1234567890';
+// Distinct uuid for the self-redirect regression: a bare-uuid request whose
+// climb name is entirely non-Latin, so it never collides with the real
+// `board-constants` lookups the numeric-format test above exercises.
+const SELF_REDIRECT_UUID = 'ffffffff000011112222333344445555';
 
 // `permanentRedirect` throws a NEXT_REDIRECT error whose `digest` the page's
 // catch block re-throws (rather than swallowing into a 404). Mirror that shape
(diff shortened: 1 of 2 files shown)


marcodejongh:
Reviewed as part of the #4650 burn-hunt remediation: the raw-request-path comparison is the right guard (parsedParams would normalise and never match), and the test additions pin both the loop case and the still-redirecting legacy cases. Marking ready — this also gates the climb-view CDN TTL bump (a cacheable self-308 must be impossible before the TTL grows to 24h). 15/15 checks green.

PR #4593, opened 2026-08-19
fix(mobile): let the drawer-dismiss pan yield to a zoomed board pan

### LINKED ISSUE #4279: Can't pan the board while zoomed in when viewing a found climb
When viewing an existing climb, zooming in works but panning (moving the board around while zoomed) is disabled. In the create-a-problem flow the same pinch-and-pan gesture works fine — so the pan gesture is supported on the board renderer but is missing or blocked in the play/view surface.

> "When zooming while making a new problem, it is possible to move the screen. But doing it on a boulder problem that you've found doesn't allow to move when zoomed in"

Maintainer confirmed: "That's a bug."

## Source

[Reported in #general on Discord](https://discord.com/channels/1288485679663415346/1288485680149827749/1534251132451295313) — flagged in a thread.

Reporter: `discord-a1c866f75756` · Posted: 2026-08-04T17:26:14.984000+00:00

marcodejongh:
## Summary

Closes #4279. Also pays off the gesture debt #4191 / #4204 left open.

Zoom into a climb in the play drawer, drag down to see the bottom of the board, and the drawer could slide away instead of the board moving. #4204 (78224c5) fixed the z-order half of that by mounting the zoom-pan overlay above the board, but it left one competitor standing: `useDrawerDismissGesture` is a `Gesture.Pan` on an **ancestor** of the board, and it wants the same one-finger downward drag. Z-order doesn't settle an ancestor/descendant race in RNGH — you need an explicit relation. That's why #4279 stayed open after #4204 merged.

### The fix

`useDrawerDismissGesture` now tags its Pan with `.withRef(gestureRef)` and returns that ref. `PlayDrawer` threads it down through `DeferredBoard` → `SwipeBoardCarousel` → `useZoomPanGesture`, where the zoomed-only pan declares `.blocksExternalGesture(dismissRef)` — the same one-directional relation the pan already uses against the surrounding RNGH ScrollView, so this follows the composition convention already in the file rather than adding a second style. No `Gesture.Exclusive` anywhere; the outer composition stays `Gesture.Simultaneous(pinch, swipe)` as before.

The block deliberately stays **one-directional**. The symmetric-looking alternative — `requireExternalGestureToFail` on the dismiss — would leave pull-to-dismiss waiting on a gesture that isn't even in the tree at 1×, killing pull-to-dismiss over the whole board. What keeps the unzoomed case safe here is that `SwipeBoardCarousel` only mounts the zoom-pan `GestureDetector` while `isZoomed`, so at 1× there is nothing holding the dismiss back.

### Tests

Two new suites, since #4204 shipped with "Device QA still owed" and no automated coverage of the race:

- `zoomed-pan-blocks-drawer-dismiss.test.ts` wires both hooks the way `PlayDrawer` does and pins the relation end to end — the dismiss Pan really claims the ref it hands out (an untagged ref would point at nothing and still read green), the zoom pan blocks exactly that ref, the dismiss takes no reciprocal wait, and a committed downward drag still fires `onDismiss` after the ref has been handed over.
- `swipe-board-carousel-zoom-overlay.test.tsx` renders the carousel and asserts the mount gating that makes the one-directional block safe: one detector at 1×, the zoom-pan detector present once zoomed.

## Release Notes

- Pan around a zoomed-in board without the climb view sliding away under your finger.

## Test plan

- `vp run test:mobile` scoped to the play drawer: 46 files, 359 tests passing (includes the two new suites and the existing dismiss/zoom-wiring suites).
- `vp run typecheck:mobile` clean.
- `vp check` — 0 errors. The 323 warnings are the pre-existing main-wide lint debt tracked in #4488, none in `play-drawer/`.

**On-device confirmation is still owed.** This box is Linux with no iOS simulator, so the RNGH relation is verified structurally (the gestures declare it) and by the mount-gating render test, not by a finger on glass. Worth one pass on a device: zoom a climb, drag down (board should pan, drawer should stay), then unzoom and drag down from the top (drawer should dismiss), and check a horizontal swipe to the next climb still works while unzoomed.


marcodejongh:

Conflicting: packages/mobile/src/components/play-drawer/PlayDrawer.tsx, SwipeBoardCarousel.tsx, use-zoom-pan-gesture.ts.

Issue #4408, opened 2026-08-13
Kiosk/embed: login-less "now on the wall" presence query


marcodejongh:
Carved out of W-11 (#4364, PR #4397), which did the other half of the A5-pre item in `docs/web-reposition.md`: both presence clients now live in `packages/web/app/lib/realtime/` (kept), so W-16 can delete `components/graphql-queue` and `components/board-presence` without breaking `/kiosk/*` and `/embed/board/*`.

Still outstanding: the **login-less** read path. `components/kiosk/presence/kiosk-presence-hub.tsx` builds its client through `createWebBoardPresenceClient` plus a `useWsAuthToken` bridge, so an anonymous kiosk/embed load still depends on `/api/internal/ws-auth` handing out a token. The epic's A5-pre note asked for a read-only "now on the wall" query that needs no session at all.

Sketch:

- Add `createReadOnlyBoardPresenceClient` to `@boardsesh/board-presence-react`. `packages/mobile` consumes that package, so this reaches the mobile graph — state the production target per standing rule 2 and run GM.
- The backend already permits the anonymous read: `requireAnonReadableBoard` at `packages/backend/src/graphql/resolvers/board-presence/shared.ts:398`.
- Deliberately left out of W-11: that PR was a pure relocation with a mechanically reviewable diff, and a new shared-package export that reaches mobile is a different review.

Verify: anonymous private window on `/kiosk/{gym}` and `/embed/board/{uuid}` renders presence with no session cookie and opens exactly one WebSocket.

Found by W-11 adversarial QA (PR #4397).

Part of #4358

Issue #2655, opened 2026-06-09
Revive: Handle GraphQL rate-limit errors gracefully on reconnect (was #2184)


marcodejongh:
Reviving the intent of #2184 (draft branch `claude/add-rate-limit-handling-HgheF` preserved for reference).

## What the PR intended

After a brief network outage the GraphQL-WS socket reconnects and the offline-reconciliation replay can fire enough buffered mutations to trip the per-user backend rate limit. Today that surfaces as a raw `Rate limit exceeded. Try again in 4 seconds.` error. The PR made the round trip quiet:

- **Backend**: in-memory (`rate-limiter.ts`) and Redis (`redis-rate-limiter.ts`) limiters throw a structured `GraphQLError` with `extensions: { code: "RATE_LIMITED", retryAfterSeconds }` via a shared `rate-limit-error.ts`; `error-handler.ts` passes `GraphQLError` instances through. Message stays human-readable for legacy clients.
- **Frontend**: `execute()` detects rate-limit errors (prefers `extensions.code === "RATE_LIMITED"`, falls back to a message regex) and retries up to 2x, waiting `retryAfterSeconds` (capped 30s + jitter). Per-attempt timeout reset so a retry wait cannot trip the mutation timeout. Subscriptions surface a typed `RateLimitError` but are not retried.
- **Offline reconciliation** (`use-offline-reconciliation.ts`) spaces replayed mutations ~75ms apart so a buffered batch cannot blow the per-minute budget on reconnect.
- **UX**: debounced `queueProvider.rateLimitCatchUp` ("Slow down — catching up…") snackbar only on the second retry; single-retry recoveries stay silent. Wired via a module-level pub/sub.

## Why it is still wanted

The underlying problem is unchanged in `main`: both backend limiters still `throw new Error(\`Rate limit exceeded. Try again in ${retryAfterSeconds} seconds.\`)` (`packages/backend/src/utils/rate-limiter.ts:45`, `redis-rate-limiter.ts:54`) with no `RATE_LIMITED` extension, and the web GraphQL-WS client has no rate-limit retry path. No other merged PR added this (the only `retryAfterSeconds` in `main` is the unrelated auth/REST limiter under `packages/web/app/lib/auth/`).

## What blocked the mechanical rebase

[...]

PR #4594, opened 2026-08-19
fix(web): bring a kiosk screen back after a long backend outage

marcodejongh:
## Summary

PR 2 of 2 for #4408. **Stacked on #4589** — review that one first; this branch
contains its commit. Base is `fix/kiosk-login-less-presence`, not `main`.

A kiosk that loses its socket for longer than graphql-ws's retry budget never
comes back on its own. The client is configured with `retryAttempts: 10` and
capped-exponential backoff (~90-180s total), and in graphql-ws the `retries`
counter resets **only** on a successful `connected`. Once the budget is spent
the client throws to the sink and stays dead; `useBoardPresence` responds by
setting `isLive(false)` and never re-subscribing (its effect deps are
`[boardId, client]`, neither of which changes). Even the 5-minute
`refresh('manual')` catch-up in `kiosk-reliability.tsx` rides that same dead
client, so it cannot heal it either.

On a laptop that is a page refresh. On a gym wall it is a frozen screen until
the 04:00 daily reload — up to ~18 hours, unattended, and it looks to the gym
like the TV broke.

So the hub supervises the socket it owns, in **two** shapes:

**1. The socket is down.** A `closed` event starts a `PRESENCE_REBUILD_AFTER_MS`
(5 min) countdown — one notch above graphql-ws's own full retry budget, so an
ordinary blip is healed by the library and never reaches this path. `connected`
cancels it. On expiry the hub bumps a `clientGeneration` that is an effect dep,
so React runs the cleanup (dispose) **before** the rebuild — the "exactly one
WebSocket" invariant PR 1 pins still holds — and every `BoardPresenceProvider`
re-subscribes on a fresh client with the retry counter back at zero. graphql-ws
emits `closed` once per failed retry, so the countdown deliberately keeps the
*first* one running: the window measures time since the wall went dark, not
time since the latest attempt.

**2. The socket is up but one board's subscription is dead.** graphql-ws does
not retry an operation the server rejected with a GraphQL error — its subscribe
loop treats `MessageType.Error` as terminal — and on a multi-board TV the other
boards keep the socket alive, so no `closed` ever fires. The hub now watches
each board's `isLive` (already published by `KioskBoardFeedBridge`) and runs the
same countdown for a board that stays dead while the socket is connected.

Pre-existing behaviour, not introduced by PR 1 — but PR 1 makes the socket the
TV's only live dependency on the backend, which raises the cost of the dead
state rather than lowering it. Hence the same stack.

## What the adversarial review caught, and what changed

Two lenses reviewed the first version of this branch. Every finding was
reproduced before it was fixed; none was rebutted.

**The rebuild blanked the wall and never refilled it.** This was the big one:
the change traded "stale but populated" for "blank forever" on exactly the long
outage it claims to fix. Measured with the real `BoardPresenceProvider`, the
real read-only presence client and the real `KioskBoardFeedBridge` over a fake
transport, backend up → down 12 min → up:

```
healthy         {"climb":"Test Problem","historyLength":1,"isLive":true}
during outage   {"climb":"Test Problem","historyLength":1,"isLive":false}
after rebuild   {"climb":null,"historyLength":0,"isLive":true}
after recovery  {"climb":null,"historyLength":0,"isLive":true}   <- stuck
```

Four links, all verified in source: a rebuild is a new client **identity**, so
`useBoardPresence`'s effect re-runs and `dispatch({type:'RESET'})` wipes
current/history/stats; its three backfill seeds are best-effort with
`.catch(() => {})` and run while the backend is down **by construction** (the
countdown is cancelled by `connected`, so it can only expire while the socket is
down); `createBoardPresenceClient` keeps `everConnected` per factory instance and
suppresses the FIRST observed `connected`, so a rebuilt client's eventual
connect never fires `runCatchUp('reconnect')`; and `boardNowPlaying` yields
pub/sub events only, with no snapshot on subscribe. Nothing refilled the slot
until the next climber sent on that wall, up to ~18h later.

Fixed by making the hub pay the RESET back explicitly: it bumps a `catchUpNonce`
on a rebuilt client's first `connected`, and `KioskBoardFeedBridge` turns that
into a `refresh('reconnect')`. `BACKFILL_HISTORY` re-adopts the newest climb as
`current` (`reducer.ts` — `shouldAdoptHistoryClimb`), so the wall comes back
populated, not merely "live with nothing on it". The blank window is now bounded
by the outage itself, which is the honest answer: while the backend is
unreachable there is nothing to show, and the header chip says `reconnecting`.

**A per-board subscription death was invisible.** `subscribeCalls === [7,8]` and
one client built after three full windows, with board 8's slot still lit with a
climb the wall may have cleared (`APPLY_CLIMB_CLEARED` exists only on the live
envelope, so no catch-up path can clear it). The trigger paths are live:
`applyRateLimit(ctx, 60, 'boardNowPlaying')` on the anon `ip:` key every TV
behind the gym's NAT shares, and `requireAnonReadableBoard`. That is shape 2
above. Capped at `MAX_STALE_SUBSCRIPTION_REBUILDS = 3`, because a board flipped
private mid-session is never coming back and would otherwise churn the whole
client — blanking every *other* board for a round trip — every five minutes
until the 04:00 reload. The budget resets only after a full window with every
board live; resetting on "every board is live" alone would hand a
permanently-rejected board an unlimited budget, since a rebuilt client reports
every board live for the microtask before its rejection lands.

**A wedged socket leaked, permanently.** graphql-ws's `dispose()` is
`disposed = true; if (connecting) { const [socket] = await connecting;
socket.close(1000) }` — it can only close a socket whose `connecting` promise
has settled, and that settles on `connection_ack`. The shared client leaves
`connectionAckWaitTimeout` at graphql-ws's default of `0`, which **disables** the
ack timeout (`isFinite(x) && x > 0`). So a proxy that accepts the upgrade while
the app behind it is restarting leaves a disposed client's socket OPEN for the
life of the page — and the rebuild path disposes mid-connect by construction.
Reproduced against a real `createClient` with a fake socket: `dispose()` never
settles and the socket stays `OPEN`; `terminate()` closes it. The cleanup now
calls `terminate()` first **when the socket is not up**, so an ordinary teardown
still closes cleanly with 1000 instead of 4499. Pinned by a new
`graphql-ws-wedged-dispose.test.ts`, which also tells us when the workaround can
go.

**The countdown had no jitter.** A backend restart drops every TV at the same
instant, and a freshly built graphql-ws client does not wait before its FIRST
attempt (`if (retrying) { await retryWait(retries) }` — a new client is not
retrying), so the fleet issued one synchronized connect + N-subscribes burst
every 5 minutes for the length of the outage. The countdown now carries
`RECONNECT_JITTER_RATIO` imported from `@boardsesh/graphql-client` — the same
constant, and the same reasoning (#2655), as the transport's own backoff.

**Cost stated rather than hidden.** Each generation gets a fresh 11-attempt
budget (the initial connect plus `retryAttempts: 10`), so an outage that
outlives the first window costs one extra burst per rebuild instead of the
single burst a supervisor-less client spends. Arithmetic, not a measurement,
because the jitter changes it: with the window averaging ~3.75 min, a 12-minute
outage is on the order of 40 attempts rather than 11 — spread over those 12
minutes, comfortably inside `boardNowPlaying`'s 60/min anon bucket, and
de-synced across the fleet by the jitter. It stays a documented trade rather
than a backoff ladder, written down in the hub's header comment and in
`docs/websocket-implementation.md`.

**One hole the review did not name, found while fixing it.** The dead-board Set
only ever grew from published snapshots, and a board whose provider unmounts
publishes no final snapshot — so a removed board's last known "dead" kept
forcing rebuilds for a slot that is gone. Not reachable from a TV (its board
list is fixed at server render, and a config change reloads the page), but the
gym-manage preview recomputes `distinctBoardIds` from the editor's live slot
state. Pruned against the rendered ids at decision time, in both the arming path
and the timer callback, with its own mutation-tested oracle.

**Two doc corrections.** The paragraph saying an `/embed/…` page "holds none"
predates this stack and is wrong: `/embed/board/[board_uuid]` mounts
`KioskPresenceHub` and opens one socket; only
`/embed/gym/[gym_uuid]/leaderboard` is WS-free. And `/embed/**` mounts the hub
**without** `KioskReliability`, so an embed gets this supervisor and the
per-board catch-up but neither the config-poll reload nor the 04:00 reload — the
hub's docstring said otherwise.

**Production targets: `www.boardsesh.com` only.** Every changed file is under
`packages/web` plus `docs/` — nothing in `packages/mobile` or
`packages/shared/**` is touched (the new `RECONNECT_JITTER_RATIO` import reads a
shared constant, it does not edit the package), so no OTA fingerprint moves and
nothing reaches the native store fleet or `app.boardsesh.com`.

## Release Notes

- A gym screen that loses its connection now finds its way back on its own — with the wall's climb and its recent sends filled back in, not a blank panel — instead of sitting dark until the nightly refresh.
- A single board on a multi-board screen that quietly stops updating now recovers too, so a TV can't keep a climb lit after the wall has been cleared.

## Test plan

Two suites. `kiosk-presence-hub.test.tsx` drives the socket lifecycle over a
passthrough provider; `kiosk-presence-recovery.test.tsx` is new and runs the
**real** `BoardPresenceProvider`, the **real**
`createReadOnlyWebBoardPresenceClient` and the **real** `KioskBoardFeedBridge`
over a fake transport, because every failure above lives in the seam between
them and a passthrough harness cannot see any of it.

Every oracle below was mutation-tested: mutation applied, `touch`ed, watched go
red, reverted with `\cp -f`, tree verified clean.

| Oracle | Mutation applied | Result |
| --- | --- | --- |
| Wall is populated again after rebuild + recovery | Drop the `setCatchUpNonce` bump on `connected` | red |
| ″ | Drop the bridge's `refresh('reconnect')` effect | red |
| A board dead on a healthy socket gets re-subscribed | Drop `noteBoardLiveness` from `publishSnapshot` | red |
| ″ | Drop `evaluateStaleSubscriptions()` from the `connected` handler | red |
| Rebuilds stop for a board that never comes back | Delete the `MAX_STALE_SUBSCRIPTION_REBUILDS` guard | red |
| ″ | Reset the budget on any all-live tick instead of a healthy window | red (6 clients, expected 4) |
| No timer left behind on unmount mid-outage | Delete `clearRebuildTimer()` from the effect cleanup | red |
| Three generations, dispose interleaved before each create | Delete `void client.dispose()` from the cleanup | red |
| ″ | Arm the countdown only on generation 0 | red |
| `terminate()` on a mid-connect teardown, and only then | Terminate unconditionally | red |
| ″ | Never terminate | red |
| Jitter spreads the window over [half, full] | Return the constant | red |
| A dead board removed from the layout stops forcing rebuilds | Make `pruneDeadBoards` a no-op | red |
| Socket stays down past the window → dispose + rebuild | Delete the countdown (no supervisor) | red |
| Blip: `closed`, half a window, `connected` → still one client | Drop `clearRebuildTimer()` from `connected` | red |

One mutation the review predicted would be caught is **not**, and I am reporting
it rather than contriving an oracle: deleting `rebuildTimerRef.current = null`
from inside the timer callback leaves all 13 socket-level cases green. It is
genuinely redundant — the generation bump always re-runs the effect, whose
cleanup clears and nulls the ref before the next `closed` can read it — so the
"recovers once, then freezes forever" failure it appears to guard cannot occur.
The line is kept as defensive, not load-bearing. The multi-generation oracle is
not inert: it goes red on both mutations above that would actually break it.

Gates, all green locally on this box:

- `vp check` — 0 errors, 324 warnings (baseline; none in the touched files)
- `vp run typecheck` (includes `next build` for web)
- `vp test run --reporter=agent --project web` — 359 files, 3969 tests
- `vp run check:i18n` + `vp run check:i18n:orphans` + `vp test run --project i18n` — clean (no strings move, so no catalog work; run because the rule is that `--project web` does not cover parity)

**Same honest note on the local suite as before:** on a loaded box, one run
showed a 7.7s timeout in `app/components/admin/__tests__/location-sync-freezes-panel.test.tsx`
— an admin panel nothing in this stack touches, green in isolation and green on
re-run. It is the same load flake flagged on the first version of this branch.
CI is the tiebreaker.

**CI on `94ff3c7`:** everything real is green — `ci-status`, `CI Tests Report`,
`typecheck`, `lint`, `i18n`, both `test-default` shards, `codegen-drift`,
`commit-lint`, `check`, `changes`, `large-files`, `label`. The one red is
`claude-review`, and its log is
`Claude Code returned an error result: You've hit your session limit · resets 2:40am (UTC)`
— an API quota failure in the bot, not a finding. It passed on an earlier commit
of this same branch; it will go green on the next re-run.

Manual QA (same dev-server session as #4589): kill the backend while a kiosk
page is open, leave it down past 5 minutes, bring it back — expect the wall to
resume **with its climb and history filled in** without a reload, and exactly
one live WS at a time in the Network tab throughout.


marcodejongh:

Conflicting: packages/web/app/components/kiosk/presence/kiosk-presence-hub.tsx.

marcodejongh:
Rebased the six own kiosk files onto current main after #4589 merged, and addressed the remaining review points in 625dbcd01.

The retry-budget comment now matches the existing full healthy-window reset rule. A new recovery regression shows that a partly spent subscription budget survives a socket outage: the healthy board refills, while the permanently rejected board cannot restart the retry cycle. The tests explicitly run sequentially, remove the constant-only assertion, and include current main’s recent-history query. The malformed NOT_FOUND documentation is corrected.

Validation: 112 kiosk/web-adapter tests, 169 shared-presence tests, full lint and all 53 repository typecheck tasks pass. Independent review approved the six-file diff and passed 18 focused tests. The mandatory pre-commit hook passed. This is website-only; no native fingerprint or OTA publication is needed. Physical kiosk outage QA has not been repeated, and the PR stays draft while fresh CI runs.

marcodejongh:
Addressed the fixture-reset nit in e87159bf0 by moving `openedSockets.length = 0` into `beforeEach`. The two focused recovery/teardown suites pass all six tests, and full lint/format, all 53 typecheck tasks, independent QA and the mandatory hook pass. Runtime behavior is unchanged.

The other observation is already covered by the sequential recovery fixture: its [`beforeEach`](https://github.com/boardsesh/boardsesh/blob/e87159bf066df3581fd875f7eb060cb85cb84963/packages/web/app/components/kiosk/presence/__tests__/kiosk-presence-recovery.test.tsx#L142) resets clients, operations, subscriptions, sinks, rejection state, wall state and backend availability before every test; `afterEach` restores timers and mocks. A preceding test failure does not skip the next test's `beforeEach`. Keeping explicit sequential execution preserves that fixture contract.

PR #4597, opened 2026-08-19
fix(mobile): resolve the session-join board from the full board list

### LINKED ISSUE #4409: Session join resolves boards from one myBoards page and hangs offline
Split out of #3823, which fixed this bug class in the deep-link hook but deliberately left the session-join flow untouched.

`packages/mobile/app/join/[sessionId].tsx` (~lines 86–93) resolves the session's board with:

- `useMyBoards(undefined)` — the server default of **20 boards** (not even the 50 cap), treated as the user's complete collection. A joiner whose matching board sorts past the first 20 gets a duplicate board minted by `CREATE_BOARD` — or a `BOARD_DUPLICATE_CONFIG` rejection that kills the join with a generic error.
- `(await myBoards.refetch()).data?.boards ?? []` — under `networkMode: 'offlineFirst'` the refetch's retryer pauses while offline, so a no-signal join spins forever; and a *failed* fetch falls back to `[]`, which is exactly the "we don't know your boards mints the same duplicate" case the deep-link hook's comment describes.
- a bare `createBoard.mutateAsync` with no duplicate recovery, even though the backend attaches `existingBoardUuid` to the rejection for precisely this.

#3823 built the pieces, so the fix is mostly a swap:

- `fetchAllMyBoards()` (`packages/mobile/src/lib/graphql/hooks/fetch-all-my-boards.ts`) — the full imperative pagination walk that rejects instead of pausing.
- `createBoardOrAdoptDuplicate` (currently module-local in `packages/mobile/src/lib/routing/use-board-route-target.ts`) — wants lifting into a shared module to be reusable here.

Worth deciding at the same time whether the cold-start guard belongs inside `resolveBoardForSession`'s contract rather than copied by every caller.

https://claude.ai/code/session_014TUjwDeVASG8eS25ArpyfF

marcodejongh:
## Summary

Joining a session via a shared link only checked the user's first 20 boards, so a matching board sitting on a later page looked like "no board owned," and the app created a duplicate (or the backend rejected it and the join failed with a generic error). It also hung forever if the join happened offline. The fix moves board-ownership resolution to a full paginated fetch that fails loudly instead of degrading to an empty list, reuses a shared duplicate-adoption helper between the join screen and the existing deep-link flow, and adds a distinct offline error message instead of an infinite spinner.

---

Closes #4409

## The bug

`app/join/[sessionId].tsx` resolved the session's board from `useMyBoards(undefined)` — the server's **20-board default page**, handed to `resolveBoardForSession` as if it were the user's whole rack. Three ways that goes wrong:

- **A board past the first page reads as no board at all.** The joiner's matching board sorts onto page two, the resolver sees no match, and `CREATE_BOARD` mints a duplicate of gear they already own — or the backend rejects it as `BOARD_DUPLICATE_CONFIG` and the join dies on a generic error.
- **Offline, the join hung.** The pre-fetch awaited `myBoards.refetch()`, and under React Query's `networkMode: 'offlineFirst'` a paused retryer never settles. The spinner spun forever.
- **A failed fetch degraded to `[]`** — the same input as "you own nothing", minting the same duplicate.

The deep-link hook (#3823) already had the pieces; the join screen never got them.

## The fix

1. **`createBoardOrAdoptDuplicate` lifted** out of `use-board-route-target.ts` into `src/lib/graphql/create-board-or-adopt-duplicate.ts`. Both callers hit the same cross-device race after their owned-list walk, so the recovery lives in one place instead of one of them.
2. **`ResolveBoardDeps.ownedBoards: UserBoard[]` becomes `loadOwnedBoards: () => Promise<UserBoard[]>`.** The guard belongs in the resolver's contract, not copied by every caller: "you own no boards" and "we couldn't find out which boards you own" are the same input to the reuse step and mint the same duplicate, so a rejected load now fails the resolve instead of guessing. It is also lazy — a named board (`/b/{slug}`) resolves by slug and never pays for the walk.
3. **The join screen passes `fetchAllMyBoards`** — the imperative pagination walk that rejects instead of pausing — and wraps its create in `createBoardOrAdoptDuplicate`.
4. **Offline gets its own message.** A join that failed on the transport says so (`mobileJoin.offlineError`, branched on `isNetworkError`) rather than showing the generic join error, which a climber can only retry blindly. Added to all four locales.

`use-board-route-target` keeps its own richer offline handling (downloaded-board fallback, `onlineManager` short-circuit) and hands its already-walked list straight to the loader, so nothing is fetched twice.

## Tests

New `app/join/__tests__/join-session-board-resolution.test.tsx` drives the real resolver and adopt helper from a Join tap against a mocked GraphQL boundary:

- a matching board past the first `myBoards` page is reused, at the session's angle, with no create;
- a `BOARD_DUPLICATE_CONFIG` rejection is adopted into the board it names, and the join completes;
- an unreachable server surfaces the offline message, mints nothing, and joins nothing;
- a server verdict (`status: 400`) keeps the generic join error, so a climber with full bars is not told they are offline.

Plus unit coverage for the lifted `createBoardOrAdoptDuplicate` and the new `loadOwnedBoards` contract (deep match reused, failed load propagated, never called for a named-board path).

## Release Notes

Joining a session from a link now finds the board you already own instead of quietly adding a second copy of it, even if you own a lot of boards. If the link opens with no signal, the join tells you you're offline and stops instead of spinning forever.

### DIFF
--- a/packages/mobile/app/join/[sessionId].tsx
+++ b/packages/mobile/app/join/[sessionId].tsx
@@ -16,13 +16,15 @@ import { useTheme } from '../../src/providers/theme-provider';
 import { useAuth } from '../../src/providers/auth-provider';
 import { useQueueSessionId, useQueueActions } from '../../src/providers/queue-provider';
 import { useToast } from '../../src/providers/toast-provider';
+import { isNetworkError } from '@boardsesh/offline-sync/error-classification';
 import {
   useSessionPreview,
-  useMyBoards,
   useCreateBoard,
   useBoardBySlug,
+  fetchAllMyBoards,
   fetchBoardBySlug,
 } from '../../src/lib/graphql/hooks';
+import { createBoardOrAdoptDuplicate } from '../../src/lib/graphql/create-board-or-adopt-duplicate';
 import { resolveBoardForSession } from '../../src/lib/board-path-to-user-board';
 import { spacing, borderRadius } from '../../src/theme/tokens';
 
@@ -47,7 +49,6 @@ export default function JoinSessionScreen() {
   const { joinSession, clearSession } = useQueueActions();
 
   const preview = useSessionPreview(sessionId);
-  const myBoards = useMyBoards(undefined, { enabled: isAuthenticated });
   const createBoard = useCreateBoard();
 
   const [isJoining, setIsJoining] = useState(false);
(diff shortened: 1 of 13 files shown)


marcodejongh:

Conflicting: packages/mobile/src/lib/routing/use-board-route-target.ts (+ its test).

Issue #4466, opened 2026-08-15
Add lastClimbAt to popularBoardConfigs so the boards sitemap shard can carry <lastmod>


marcodejongh:
Follow-up from #4434 (PR #4464).

`/sitemaps/boards.xml` ships with **no `<lastmod>`**, deliberately: there is no per-config content timestamp in the data, `<lastmod>` is optional in the sitemap protocol, and inventing one with `new Date()` is exactly what the "real content timestamps" rule bans. Crawlers fall back to other freshness signals in the meantime.

## The work

Backend (`packages/backend/src/graphql/resolvers/social/boards.ts`, `getPopularConfigs`): add `lastClimbAt` to the `popularBoardConfigs` payload — `MAX(board_climbs.created_at)` for the climbs already counted by the config's `LATERAL` subquery, so it costs nothing extra. The value is cached in Redis with the rest of the config row.

Web: set `lastModified: config.lastClimbAt` in `boardConfigsToItems` (`app/lib/seo/sitemap/board-entries.ts`) and drop the comment explaining the omission. `packages/web/app/lib/seo/sitemap/__tests__/board-entries.test.ts` currently asserts `lastModified` is `undefined` — flip that assertion in the same PR.

Out of scope for #4434, which was web-only.

marcodejongh:
## Measured before shipping: a config-wide `MAX(created_at)` is the wrong `<lastmod>` for these URLs

Built both halves, then measured the value against the page it would stamp. It
does not describe that page, so the web half (#4599) no longer wires it up — it
records the numbers and pins the omission with a test instead. #4598 (the
backend field) is in draft with no consumer and needs a call.

### What the sitemapped URL actually is

`boardConfigsToItems` emits page 1 only, at the default sort:
`fetchFrontDoorListPage` fetches exactly `FRONT_DOOR_PAGE_SIZE = 50` climbs at
`DEFAULT_SEARCH_PARAMS.sortBy = 'ascents'`. `StaticListFrontDoor` renders those
50 rows plus a static heading and pagination — no total count, no dates. The
page changes when its top-50 membership changes, not when a climb is set.

### Dev database, the 2026-01-31 catalog snapshot

`/kilter/…/40/list` — layout 1 / size 10 / sets 1+20, the config-tuple URL this
shard submits:

| | |
|---|---|
| ascents at rank 50 (the page-1 cutoff) | **21,407** |
| newest `created_at` among those 50 climbs | **2021-02-11** |
[...]

PR #4598, opened 2026-08-19
feat(backend): expose lastClimbAt on popularBoardConfigs

marcodejongh:
## Summary

Adds a `lastClimbAt` field to the backend's `popularBoardConfigs` GraphQL query, computing the newest listed climb's creation time per board config (normalising two different upstream timestamp formats into ISO-8601 UTC). It was built as the backend half of a stack feeding a `<lastmod>` value into the boards sitemap, but the web PR that would have consumed it (#4599) no longer does so, because a config-wide max timestamp turned out to be inaccurate for that purpose (a new climb can't reach the sitemapped page-1 URL, so the value doesn't reflect real freshness). The PR is currently in draft with no consumer selecting the field, and is now a decision point: close it as dead code, or rebuild it as a per-angle map to match the honest sitemap use case (which would need the caching/warm-path redesigned for cost reasons).

---

> ## ⚠️ Read this before merging: the consumer this PR was built for is gone
>
> This shipped as the backend half of a stack whose web half (#4599) would stamp
> `lastClimbAt` as `<lastmod>` on `/sitemaps/boards.xml`. **#4599 no longer does
> that** — measured against the page it would stamp, a config-wide
> `MAX(created_at)` is fabricated freshness, so #4599 now records the measurement
> and pins the omission instead. It is unstacked, based on `main`, and selects
> nothing new from the backend.
>
> **That leaves this PR with no consumer.** It is in draft and needs a call:
>
> - **Close it** — nothing selects `lastClimbAt`, and a GraphQL field nothing
>   selects is dead code.
> - **Or commission the honest version** — `MAX(created_at)` over the page-1
>   window *per angle*, returned as a per-(config, angle) map rather than one
>   scalar per config. That is a different field shape, and it needs the warm
>   path redesigned: measured on the dev database the per-angle query costs
>   **207 s** on top of the **78 s** this statement already spends, against a
>   **120 s** `REDIS_LOCK_TTL_SECONDS`. The normaliser, the `:v2` key and the
>   nullable-SDL decision below all survive into that version.
>
> ### Why the scalar is wrong for a `<lastmod>`
>
> The sitemapped URL is page 1 only, at the default sort — the top
> `FRONT_DOOR_PAGE_SIZE` (50) climbs by ascents. Dev database, 2026-01-31
> snapshot, `/kilter/…/40/list` (layout 1 / size 10 / sets 1+20):
>
> | | |
> |---|---|
> | ascents at rank 50 (the page-1 cutoff) | 21,407 |
> | newest `created_at` among those 50 | 2021-02-11 |
> | best ascents among climbs created into the config in the last 90 days | 86 (0.4% of the cutoff) |
> | climbs created into the config in the last 30 days | 6,148 (~205/day) |
> | what this field would have stamped | 2026-01-31 |
>
> A new climb cannot reach page 1, but the value moves daily. It is also not
> angle-aware while the page is (the list query INNER JOINs `board_climb_stats`
> at the URL's angle): tension layout 11 / size 9 honest page-1 maxima run
> 2024-12-26 (65°) to 2026-01-26 (55°), and 70° has no page-1 content at all,
> against one 2026-01-31 stamp for all 15 angle URLs.
>
> Full numbers and method: #4599 and
> `packages/web/app/lib/seo/sitemap/board-entries.ts`.
>
> ### Also added since review
>
> A doc-only commit recording two properties a reader would otherwise have to
> infer: the whole cached payload only refreshes at boot, so it tracks deploy
> cadence rather than the catalogue; and bounding the Redis TTL to hours is not
> the fix, because the 78 s statement would then land on a live request behind
> the web side's 10 s `SITEMAP_FETCH_TIMEOUT_MS` and the boards shard's 3 s
> `SHARD_DEADLINE_MS`.

---

## Summary

Adds `lastClimbAt` to `popularBoardConfigs` — the newest listed climb's
upstream creation time for each board config, normalised to a proper
ISO-8601 UTC string. This is the **backend half of a two-PR stack** for
#4466; the web half (which turns this into `<lastmod>` on
`/sitemaps/boards.xml`) is #4599 and depends on this PR being live in
production first.

**This PR ships inert.** No client selects the new field yet, so merging
and deploying it changes nothing observable.

### Why `created_at`, and why it needed normalising

`board_climbs.created_at` is `text`, not `timestamptz`, and carries two
naive-UTC (zone-less) separator styles depending on the importer: Aurora
writes a space (`'2024-03-11 09:00:00'`), MoonBoard writes ISO-T
(`'2023-11-23T18:00:15.227'`). A raw text `MAX()` across a mix of both
picks the wrong row on a shared calendar day — `'T'` (0x54) sorts after
`' '` (0x20) — so the SQL aggregate canonicalises the separator
(`REPLACE(bc.created_at, ' ', 'T')`) before taking the max, filtered by a
`LIKE` mask (never a `::timestamp` cast — Postgres has no `try_cast`, so
one legacy garbage row would 500 the whole query). The JS-side
`normalizeCatalogTimestamp` is a pure string transform, not a `Date`
round-trip, so an explicit `Z` gets stamped identically in every runner
timezone; it also rejects anything unparseable and anything in the future.

### Was `created_at` a real upstream clock, or a sync-job artefact?

Ran the read-only production probe the plan called for (Marco
pre-authorized read-only prod access for this work) before writing any
code:

```
board_type    total    with_created  iso_t   space_sep  min_created                    max_created
decoy         8485     8485          0       8485       2021-05-09 19:30:46.884686      2026-08-19 00:15:15.468256
grasshopper   21923    21923         0       21923      2020-09-07 15:14:46.224079      2026-08-19 03:49:29.118979
kilter        317801   317801        52870   264931     2018-05-14 06:08:06.348046      2026-08-19T21:01:25.740558Z
moonboard     271780   239947        239947  0          2010-09-29T12:35:11             2026-08-14T10:08:12.552Z
soill         2035     2035          0       2035       2023-10-13 16:38:52.187900      2026-08-19 01:27:59.299356
tension       144442   144442        2       144440     2021-02-13 01:52:52.000000      2026-08-19 20:58:00.700593
touchstone    1353     1353          0       1353       2022-02-16 03:57:24.791820      2026-01-30 01:54:04.615780
```

(filtered to `is_listed = true AND is_draft = false` — the same rows the
LATERAL scans.)

To settle whether `created_at` is a real per-climb clock or just "when the
sync job last ran," pulled the top 10 kilter rows by `created_at DESC`:
they have **distinct** `created_at` values staggered minutes apart with
distinct climb names, while `updated_at` is identical across all of them
(`2026-08-19T21:03:00.443Z` — the sync job's completion time). That's the
signature of a genuine upstream per-climb timestamp, not a sync-run
artefact — confirming the issue's original choice of `created_at` over
`updated_at`.

`moonboard`'s NULL rate on `created_at` (~12%) is real but not
NULL-dominated — the config still gets a `<lastmod>` from its other 88%.

### What changed

- `boards.ts`: the existing `LEFT JOIN LATERAL` gains one aggregate
  (`MAX(REPLACE(bc.created_at, ' ', 'T')) FILTER (WHERE bc.created_at LIKE
  '____-__-__%')`) riding rows the LATERAL already scans — no new join,
  no new index.
- New exported `normalizeCatalogTimestamp(raw: unknown): string | null` —
  pure-string ISO normaliser, timezone-independent by construction.
- `CachedPopularConfig.lastClimbAt: string | null`.
- SDL: `PopularBoardConfig.lastClimbAt: String` — **deliberately
  nullable** (not `String!`). The Redis cache carries a 1-year TTL and is
  only refreshed by the node that wins a deploy-time lock; a node that
  loses the lock, or an old-version node restarting after a deploy, can
  serve a legacy cached row with no `lastClimbAt` key at all. A non-null
  field over that missing key would null the entire `PopularBoardConfig`,
  which sits inside `configs: [PopularBoardConfig!]!` — erroring the
  whole query and taking down the boards sitemap shard, the paged climbs
  shard's config groups, and the homepage rail together. A new backend
  test pins the SDL type as `String`.
- Redis cache key bumped `boardsesh:popular-board-configs` → `...:v2`.
  The nullable-field path is the safety net, not the plan — the version
  bump means new code structurally cannot read a pre-shape cached row,
  rather than relying on "probably refreshed."
- `vp run codegen` output regenerated (`shared-schema` generated
  schema/types, `shared/graphql` generated client types).

### Tests

New `popular-board-configs-last-climb-at.test.ts` (real DB):
- Seeds one config with climbs at `'2024-01-01 10:00:00.123456'`
  (space), `'2024-03-11 09:00:00'` (space, the true max),
  `'2024-03-11T08:00:00.000'` (same date, T-separated, earlier — the row
  a naive text `MAX` would wrongly prefer), and a NULL `created_at`; two
  decoys (out-of-bounds edges, unlisted) with far-future timestamps that
  must not leak through. Asserts the resolved `lastClimbAt` is exactly
  `'2024-03-11T09:00:00.000Z'`.
- `normalizeCatalogTimestamp` table test covering both separator styles,
  an explicit zone, and the null/malformed/future rejections.
- Cache-legacy-row test: a Redis hit missing the `lastClimbAt` key
  resolves without throwing and leaves the field `undefined` — never a
  synthesised `new Date()`.
- `operations-schema-validation.test.ts`: asserts
  `PopularBoardConfig.lastClimbAt` is exactly `String`, not `String!`.

Every oracle was mutation-tested (mutation applied, watched red, reverted):
dropping the `REPLACE` separator normalisation, tightening the SDL field
to `String!`, and adding a `?? new Date()` fallback in the cache-read
path all turned the corresponding test red.

## Release Notes

## Test plan

- `vp check --fix` — 0 errors (324 pre-existing warnings, unrelated to
  this change).
- `vp run typecheck` — clean.
- `vp test run --project backend --reporter=agent` (isolated run, per the
  repo's backend-tests-no-concurrency rule) — 244/244 passing across
  `popular-board-configs-last-climb-at.test.ts` and
  `operations-schema-validation.test.ts`.

## Production targets

**Corrected — this is not backend-only.** The earlier claim ("Backend only …
No web, mobile, or native change") was wrong. This PR edits
`packages/shared-schema/src/schema/board-entities.ts` and its generated output,
and `.github/workflows/mobile-ota-production.yml` fires on
`packages/shared-schema/**` pushed to `main`. So merging this **auto-publishes a
production OTA to the entire native store fleet** (JS-only diff, so the
fingerprint is unchanged and every installed binary picks it up), alongside the
Railway backend deploy. Mobile keeps its own copy of the document and never
selects `lastClimbAt`, so there is no behaviour change — but the fleet does get
a new bundle.

There is no deploy-order constraint left: #4599 is unstacked, based on `main`,
and selects nothing new from the backend.
</content>
</invoke>

### DIFF
--- a/packages/backend/src/__tests__/operations-schema-validation.test.ts
+++ b/packages/backend/src/__tests__/operations-schema-validation.test.ts
@@ -1,6 +1,6 @@
 import { describe, it, expect } from 'vite-plus/test';
 import { readFileSync } from 'fs';
-import { buildSchema, parse, validate, visit, type DocumentNode, type GraphQLSchema } from 'graphql';
+import { buildSchema, parse, validate, visit, GraphQLObjectType, type DocumentNode, type GraphQLSchema } from 'graphql';
 import { typeDefs } from '@boardsesh/shared-schema';
 import * as publicOperations from '@boardsesh/graphql/operations';
 import * as accountOperations from '@boardsesh/graphql/operations/account';
@@ -471,4 +471,29 @@ describe('previous-release driver operations: reduced-B7 validation split', () =
       expect(unknownMutationErrors.length).toBeGreaterThan(0);
     });
   }
+});
+
+// Issue #4466: `PopularBoardConfig.lastClimbAt` MUST stay nullable (`String`,
+// not `String!`). The Redis cache is keyed with a 1-year TTL and warmed only
+// by the node that wins a deploy-time lock — a node that loses the lock, or
+// an old-version node restarting after a deploy, can serve a legacy cached
+// row that has no `lastClimbAt` key at all. A non-null field over that
+// missing key would null the whole `PopularBoardConfig`, which sits inside
+// `configs: [PopularBoardConfig!]!` — erroring the entire query and taking
+// down the boards sitemap shard, the paged climbs shard's config groups, and
+// the homepage rail all at once. Every other test in this campaign runs
+// against a freshly-built cache, so only an explicit SDL assertion catches
+// this regression.
+describe('PopularBoardConfig.lastClimbAt stays nullable', () => {
+  it('is declared as `String`, not `String!`, on the executable schema', () => {
+    const popularBoardConfigType = schema.getType('PopularBoardConfig');
+    if (!(popularBoardConfigType instanceof GraphQLObjectType)) {
+      throw new Error('PopularBoardConfig type not found (or not an object type) on the built schema');
+    }
+    const lastClimbAtField = popularBoardConfigType.getFields().lastClimbAt;
+    if (!lastClimbAtField) {
+      throw new Error('PopularBoardConfig.lastClimbAt field not found on the built schema');
+    }
+    expect(String(lastClimbAtField.type)).toBe('String');
+  });
 });
--- a/packages/backend/src/__tests__/popular-board-configs-last-climb-at.test.ts
+++ b/packages/backend/src/__tests__/popular-board-configs-last-climb-at.test.ts
@@ -154,8 +154,18 @@ describe('popularBoardConfigs.lastClimbAt (SQL aggregate, real DB)', () => {
 describe('normalizeCatalogTimestamp', () => {
   it.each([
     ['2024-03-11 09:00:00.123456', '2024-03-11T09:00:00.123Z'],
+    // The commonest Aurora shape: space separator, no fraction at all. Covered
+    // only through the SQL test otherwise, which cannot distinguish "the
+    // normaliser handled it" from "the aggregate picked a different row".
+    ['2024-03-11 09:00:00', '2024-03-11T09:00:00.000Z'],
     ['2023-11-23T18:00:15.227', '2023-11-23T18:00:15.227Z'],
     ['2024-03-11T09:00:00Z', '2024-03-11T09:00:00.000Z'],
+    // The SQL `LIKE '____-__-__%'` mask is deliberately loose enough to pass a
+    // zone-carrying row through, so the JS side has to convert one rather than
+    // stamp `Z` over it. +05:30 is 03:30 UTC, and the colon-less spelling is
+    // the same instant.
+    ['2024-03-11T09:00:00+05:30', '2024-03-11T03:30:00.000Z'],
+    ['2024-03-11T09:00:00-0800', '2024-03-11T17:00:00.000Z'],
     // Sub-millisecond fraction, explicitly right-padded (0.5s = 500ms, not 5ms).
     ['2024-03-11T09:00:00.5', '2024-03-11T09:00:00.500Z'],
     [null, null],
--- a/packages/backend/src/__tests__/popular-board-configs-last-climb-at.test.ts
+++ b/packages/backend/src/__tests__/popular-board-configs-last-climb-at.test.ts
@@ -1,4 +1,4 @@
-import { afterAll, afterEach, beforeAll, describe, expect, it, vi } from 'vitest';
+import { afterAll, afterEach, beforeAll, describe, expect, it, vi } from 'vite-plus/test';
 import { and, eq, sql } from 'drizzle-orm';
 import * as dbSchema from '@boardsesh/db/schema';
 import { db } from '../db/client';
--- a/packages/backend/src/__tests__/popular-board-configs-last-climb-at.test.ts
+++ b/packages/backend/src/__tests__/popular-board-configs-last-climb-at.test.ts
@@ -1,0 +1,217 @@
+import { afterAll, afterEach, beforeAll, describe, expect, it, vi } from 'vitest';
+import { and, eq, sql } from 'drizzle-orm';
+import * as dbSchema from '@boardsesh/db/schema';
+import { db } from '../db/client';
+import { redisClientManager, type RedisClients } from '../redis/client';
+import {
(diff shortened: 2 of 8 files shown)


marcodejongh:

Conflicting: packages/backend/src/graphql/resolvers/social/boards.ts (a 44-line resolver-logic conflict).

marcodejongh:
Closing per the call this PR asked for: **close it as dead code.**

Nothing selects `lastClimbAt`. #4599 measured the consumer this was built for and rejected it — a config-wide `MAX(created_at)` stamps a page-1 URL whose newest listed climb is from 2021 while advancing daily, and it is not angle-aware while the page is. A GraphQL field nothing selects is surface we would have to keep answering for.

The honest per-angle version stays unbuilt on the numbers already in this PR body: **207 s** on top of the existing **78 s**, against a **120 s** `REDIS_LOCK_TTL_SECONDS`. If a real consumer for per-config freshness turns up, that redesign starts from this branch — the timestamp normaliser, the `:v2` cache key and the nullable-SDL decision all survive into it, and the branch is not deleted.

The reasoning does not leave with the PR: #4599 lands the measurement in `packages/web/app/lib/seo/sitemap/board-entries.ts` and strengthens the test to an XML-level assertion, so the next person to reach for `<lastmod>` on that shard finds out why in-tree rather than re-deriving it.

Closes the backend half of #4466.

PR #4604, opened 2026-08-19
feat(mobile): a read-only climb surface for signed-out readers

marcodejongh:
## Summary

Part 1 of 2 for #4478. Adds the pieces a signed-out reader needs to **read** a climb, without wiring anything to them yet. Nothing routes here, so this ships behaviourally inert on both platforms — the routing switch is the stacked PR.

Splitting this way keeps each reviewable unit honest: this PR alone would ship dead buttons if it were the routing PR, and the routing PR alone would ship a login wall if it were the surface PR.

- **`AnonymousClimbView`** hosts `PlayDrawer` in its existing `presentation="pane"` mode — the seam `IpadPlayPane` already proves — with a standing sign-in bar whose href comes from `buildLoginHrefWithReturn()`, never a hand-rolled `/auth/login`. The climb opens as a `previewQueueItem`, so the queue is never written.
- **`viewer: 'member' | 'anonymous'`** through `PlayDrawer` and `PlayDrawerActionBar`. Write affordances are **removed rather than disabled** — queue, favourite, lightbulb, climb actions — because a dead button a visitor taps and nothing happens is worse than the login wall this replaces. The tick stays as the one conversion prompt, relabelled and pointed at login. Mirror, prev/next, share and the angle pill are reads and stay live.
- **Similar Climbs stays, but its tap changes.** It is the best reason a visitor has to keep looking, so removing it would be the wrong trade — but the member tap does two things a signed-out reader must not get. It writes the local queue (a list they cannot carry anywhere, sitting next to a queue button that is hidden), and it calls `setCurrentClimb`, which **re-arms the BLE auto-sender** and pushes the new climb to a connected board. Anonymously it takes the preview path the wrong-board drawer already uses: the drawer shows the climb, nothing is committed, nothing reaches a wall.
- **The "Preview · Set active" banner is hidden too.** The anonymous view always opens with a preview item, so the banner rendered every time — and its one button calls `setCurrentClimb`, i.e. a queue write plus a BLE re-arm, which is the exact affordance the rest of this PR removes. It also asks a question a signed-out reader has no answer to: there is no active climb and no wall.
- **`PlayDrawer` is now rendered in a unit test** (`play-drawer-anonymous.test.tsx`). It never had been — both existing suites mock it — so the props it hands its children were unmeasured, and six mutations survived the entire 6,500-test suite. They do not any more.

- Two of the removals are about **navigation**, not writes. The anonymous gate re-reads `window.location.pathname` on every navigation, so anything pushing `/boards`, `/play` or `/climbs/{uuid}` bounces the visitor to login mid-session. The reasoning is in a code comment, because this is the failure mode that will bite a later change.
- **The lightbulb is hidden.** Web Bluetooth *is* mounted on the browser export, so it would otherwise render for an anonymous visitor — board presence binds on an active board uuid they do not have, and a pairing prompt on a first-ever visit is a hostile opening. Anonymous wall lighting deserves its own thought; filed as #4606. The suppression lives in **one** place, `PlayDrawerActionBar`'s `viewer` rules, next to the test that pins it; `showLightbulb` went back to meaning only "a Bluetooth transport exists".

### Two real bugs fixed on the way

- **`useFavoriteStatus` had no session gate.** It was armed on `enabled: isSheetOpen` alone, and `favorites` is `requireAuthenticated` server-side — so every anonymous drawer open would have fired a query with exactly one possible outcome. Gated in the hook, not only at the call site, so a future consumer cannot reintroduce it by forgetting.
- **The Logbook told first-time visitors they had "No tries yet. Get on it."** Everything reaching that section signed-out looks like a member with no history: the logbook query is disabled so it lands empty, and `userAscents`/`userAttempts` are viewer-scoped and arrive null. A comment in `DeferredSections.tsx` claimed the section "still expands to the sign-in prompt" — no such branch existed (`DeferredSections.tsx:117` gates the summary *string*, and `LogbookSection` had no auth gate at all). Now the branch exists and the comment is correct.

### A third one, found while building it

**The pane painted "Pick a climb. Tap a climb to see it here." as its first frame.** The open target is applied by an effect, which runs after the first commit — so a pane that already has its climb shows the placeholder for one frame first. True on the iPad, where the pane really is waiting on a list selection; a lie on the anonymous view, where there is no list to tap and the visitor followed a link to one specific climb. The preview item is now seeded in the state initialiser. Only a *preview* target seeds anything — a commit open has queue side effects that belong in the effect.

### Three more, from the adversarial review

**The anonymous drawer offered a live "Set active".** It is always in preview mode — that is what keeps the queue untouched — so the "Preview / Set active" banner rendered on every read-only open, and its press is `setCurrentClimb`: the same queue write and the same BLE auto-sender re-arm the similar-climb guard above was written to block, wearing a different button. Gated on the viewer now. "Preview" also means nothing to a reader with no active climb to promote it over.

**The angle pill was cosmetic where it mattered.** Grade, stars and ascent count are baked into the climb row at the angle it was fetched for (`getClimbByUuid` joins the stats tables ON angle), the route's query is frozen at the URL angle, and `useEffectiveClimbStats` — what keeps the signed-in header honest across an angle change — is `requireAuthenticated` and never fires here. So the header held the URL angle's numbers while the Boardsesh grade, the crowd-grade bar and the angle sheet's own per-angle stats all moved: two grades for one angle, on one screen. The climb is refetched at the live angle, and the one already on screen keeps rendering while that is in flight, so the drawer never blanks mid-swap.

Known gap, left as-is: the URL is not rewritten when the angle moves, so `buildLoginHrefWithReturn()` sends the reader back to the URL's angle rather than the one they were looking at. The signed-in deep-link path behaves the same way, and rewriting the address bar mid-read is its own decision.

**A board that cannot tilt can now say so.** `isAngleAdjustable` rides through to the drawer instead of defaulting to an angle pill for a bolted wall. The stacked PR resolves it from the `/b/{slug}` board record.

Copy added to all four locales, following the glossaries (es *encadene*, fr *faire la croix*, de *Begehung* / informal *du*).

Closes nothing on its own. Refs #4478.

## Review gates

- **`ble-needs-fable-review`.** This suppresses the lightbulb and stops a similar-climb tap re-arming the BLE auto-sender. Web Bluetooth is live on the browser export, so this touches the BLE surface — **Fable review required before merge** (repo rule), on top of the standing Fable review every `packages/mobile` PR gets under epic #4358.
- **Needs Marco's explicit nod on epic #4358's standing rules** before merge — this is a W-06 follow-up and it publishes to the native fleet as an inert OTA (below).
- Not auto-mergeable.

### Production targets

**The native store fleet, via auto-OTA** — this touches `packages/mobile/**` and `packages/shared/i18n`, so merging to `main` auto-publishes a production OTA that reaches every installed binary within hours. It is behaviourally inert there (nothing renders `viewer="anonymous"` until the stacked PR), but the diff does ship. Also **app.boardsesh.com** for the same reason. Not `www.boardsesh.com`.

**No native fingerprint inputs changed** — no `app.config.ts`, `plugins/**`, `modules/**`, `locales/**`, `patches/**` or dependency edits — so `main` stays OTA-compatible with the store binaries and the OTA-compatibility check reports both platforms matching `main`.

## Release Notes

Nothing is reachable yet — the user-facing note ships with the stacked routing PR.

## Test plan

- `vp run test:mobile` — green.
- `vp test run --project i18n catalog-completeness --reporter=agent` — green (four-locale parity; `--project web` does not run it).
- `vp run check:i18n`, `vp run check:i18n:orphans` — clean.
- `vp check` — 0 errors, formatted.
- `vp run check:mobile-bundle`, `vp run check:mobile-web-bundle` — both OK.

**Every new oracle was mutation-tested** — the mutation applied, the test watched go red, reverted:

| Mutation | Caught by |
| --- | --- |
| `useFavoriteStatus` loses the session gate | `use-favorite-status.test.tsx` |
| ...or becomes "never fetch" (breaks the member heart) | same file, the signed-in half |
| Logbook signed-out branch deleted | `logbook-section-signed-out.test.tsx` |
| ...or ordered after the count fallbacks | same file |
| Favourite / queue / ellipsis / lightbulb rendered anonymously | `play-drawer-action-bar.test.tsx`, one assertion each |
| Tick FAB hidden anonymously (removes the only way in) | same file |
| Tick wired to `onTickPress` instead of sign-in | same file |
| Heart falls back into slot 1 on a fixed-mirror board | same file |
| Ascent badge kept anonymously | same file |
| Sign-in href hard-coded to `/auth/login` | `anonymous-climb-view.test.tsx` |
| **`openTarget` left unmemoised (infinite re-open loop)** | `anonymous-climb-view.test.tsx`, the settle test |
| Nonce bumped from inside a state updater (double-advances under Strict Mode) | same file, the Strict Mode test |
| Open target's nonce not bumped on angle change (drawer blanks) | same file |
| `presentation="route"` instead of `"pane"` | same file |
| `previewQueueItem` dropped (queue would be written) | same file |
| Nonce bumped for an angle that did not move | same file |
| Similar-climb tap queues + commits for an anonymous reader | `play-drawer-navigation.test.ts` |
| Pane preview seed always null (placeholder paints again) | `play-drawer-layout.test.ts` |
| ...or a commit open seeded too (queue side effects bypassed) | same file |
| Angle refetch never armed / armed at the URL angle | `anonymous-climb-view.test.tsx` |
| Fresh climb ignored, open target keeps the URL angle's numbers | same file |
| In-flight fallback dropped (drawer blanks mid-swap) | same file |
| `isAngleAdjustable` not forwarded, or forced true | same file |
| The six `PlayDrawer` call-site mutations in the table below | `play-drawer-anonymous.test.tsx` |

Per-affordance assertions rather than one "renders no write buttons" sweep: a heart that 401s and a missing tick are different product regressions and a blanket assertion could not tell them apart.

The **settle test** deserves a note: every other assertion in that file reads the *last* render's props, so all of them stay green inside an infinite re-render loop. Only the render count plus the open target's identity across a re-render can tell a memoised target from a fresh object.

### `PlayDrawer` is now rendered in a test

It never had been — both existing suites mock it — so every anonymous rule was pinned one hop away from the code that ships it, and the hop is where the bugs were. Six mutations survived the whole mobile suite before `play-drawer-anonymous.test.tsx` existed:

| Mutation at the call site | Consequence | Now caught by |
| --- | --- | --- |
| `viewer={viewer}` → `viewer="member"` | queue button, heart, lightbulb, ellipsis and tick sheet all return for a signed-out reader | the viewer join |
| similar-climb preview guard deleted | `addToQueue` + `setCurrentClimb` return — the BLE auto-sender re-arms on the one surface where Web Bluetooth is mounted | the write assertions, both directions |
| `onSignInPress={onSignIn}` dropped | the only conversion prompt on the surface becomes a dead button | the same viewer join |
| preview banner ungated anonymously | a live "Set active" — same write, same re-arm | the banner assertion, both directions |
| `enabled: isSheetOpen && !isAnonymous` → `enabled: isSheetOpen` | an auth-only `favorites` query on every anonymous open | the enabled assertion, both directions |
| `initialDrawerPreviewItem(openTarget)` → `null` | the pane paints "Pick a climb" for one frame | the placeholder count across every commit |

The oracles inside `PlayDrawerActionBar` were all sound; only their connection to the shipped render was unmeasured.

Also not verified here: how the pane-presentation drawer reads as a full-width browser surface. It has only ever been exercised as an iPad right column — a real-device gate, and the most likely source of a second iteration.

### DIFF
--- a/packages/mobile/src/components/AnonymousClimbView.tsx
+++ b/packages/mobile/src/components/AnonymousClimbView.tsx
@@ -24,7 +24,7 @@
 // Native never renders this. The whole anonymous branch sits behind
 // `RELAXES_ANONYMOUS_ROUTES`, whose native fork is a literal `false`.
 
-import { memo, useCallback, useMemo, useState } from 'react';
+import { memo, useCallback, useMemo, useRef, useState } from 'react';
 import { View, StyleSheet } from 'react-native';
 import { Stack, useRouter } from 'expo-router';
 import { useTranslation } from 'react-i18next';
@@ -70,15 +70,20 @@ export const AnonymousClimbView = memo(function AnonymousClimbView({
   // without a fresh nonce the anonymous drawer would fall back to the queue's
   // current climb, which is empty here, and blank itself.
   const [openNonce, setOpenNonce] = useState(1);
+  // The current angle again, in a ref, purely so the "did it actually move?"
+  // check can happen OUTSIDE a state updater. Bumping the nonce from inside
+  // `setAngle`'s updater read naturally but broke React's contract that updaters
+  // are pure: Strict Mode runs them twice, so every angle change moved the nonce
+  // by two in development and by one in production — the kind of difference that
+  // makes a re-open bug reproduce on a device and not on a desk.
+  const angleRef = useRef(boardConfig.angle);
   const handleAngleChange = useCallback((nextAngle: number) => {
-    // Both writes are conditional on the angle actually moving. Re-opening the
-    // drawer for an angle it is already on would rebuild the preview item for
-    // nothing.
-    setAngle((currentAngle) => {
-      if (currentAngle === nextAngle) return currentAngle;
-      setOpenNonce((nonce) => nonce + 1);
-      return nextAngle;
-    });
+    // Re-opening the drawer for the angle it is already on would rebuild the
+    // preview item for nothing, so both writes are conditional on the move.
+    if (angleRef.current === nextAngle) return;
+    angleRef.current = nextAngle;
+    setAngle(nextAngle);
+    setOpenNonce((nonce) => nonce + 1);
   }, []);
 
   const paneBoardConfig = useMemo<BoardConfig>(() => ({ ...boardConfig, angle }), [boardConfig, angle]);
--- a/packages/mobile/src/components/AnonymousClimbView.tsx
+++ b/packages/mobile/src/components/AnonymousClimbView.tsx
@@ -67,8 +67,14 @@ export const AnonymousClimbView = memo(function AnonymousClimbView({
   // current climb, which is empty here, and blank itself.
   const [openNonce, setOpenNonce] = useState(1);
   const handleAngleChange = useCallback((nextAngle: number) => {
-    setAngle((currentAngle) => (currentAngle === nextAngle ? currentAngle : nextAngle));
-    setOpenNonce((nonce) => nonce + 1);
+    // Both writes are conditional on the angle actually moving. Re-opening the
+    // drawer for an angle it is already on would rebuild the preview item for
+    // nothing.
+    setAngle((currentAngle) => {
+      if (currentAngle === nextAngle) return currentAngle;
+      setOpenNonce((nonce) => nonce + 1);
+      return nextAngle;
+    });
   }, []);
 
   const paneBoardConfig = useMemo<BoardConfig>(() => ({ ...boardConfig, angle }), [boardConfig, angle]);
--- a/packages/mobile/src/components/AnonymousClimbView.tsx
+++ b/packages/mobile/src/components/AnonymousClimbView.tsx
@@ -1,0 +1,149 @@
+// The read-only climb view a signed-out visitor gets on app.boardsesh.com.
+//
+// www links every indexed climb page into this app at the same path. Until now
+// that arrival hit a login form — 209 characters of body copy where the SEO
+// funnel was supposed to end — because the board routes are redirectors: they
+// adopt the URL's board (a write that needs an account) and hand off to the
+// Climbs tab plus the `/play` modal.
+//
+// Neither half was necessary to DRAW the climb. Every render input is in the
+// URL: the tuple carries the whole board config, board art is a local catalogue
+// lookup, and the `climb` resolver takes no viewer at all. So this screen skips
+// adoption and, crucially, skips the navigation: `PlayDrawer` renders here in
+// its existing `presentation="pane"` mode, in place, on the canonical
+// `/…/{angle}/view/{segment}` URL.
+//
+// Staying put is not a style choice. `/(tabs)/climbs` and `/play` are both
+// outside the anonymous allow-set, and `AuthProvider` re-reads
+// `window.location.pathname` on every navigation — so a hand-off would bounce
+// the visitor back to the login wall the moment the drawer opened. Anything
+// added here that navigates off the allow-set does the same thing silently; see
+// `PlayDrawerActionBar`'s `viewer` prop for the affordances that were removed
+// for exactly that reason.
+//
+// Native never renders this. The whole anonymous branch sits behind
+// `RELAXES_ANONYMOUS_ROUTES`, whose native fork is a literal `false`.
+
+import { memo, useCallback, useMemo, useState } from 'react';
+import { View, StyleSheet } from 'react-native';
+import { Stack, useRouter } from 'expo-router';
+import { useTranslation } from 'react-i18next';
+import { useSafeAreaInsets } from 'react-native-safe-area-context';
+import type { Climb } from '@boardsesh/shared-schema';
+import { PlayDrawer } from './play-drawer/PlayDrawer';
+import { Button } from './Button';
+import { Text } from './Text';
+import { climbToQueueItem } from '../lib/climb-to-queue-item';
+import { buildLoginHrefWithReturn } from '../lib/routing/anonymous-auth-gate';
+import { useTheme } from '../providers/theme-provider';
+import type { BoardConfig } from '../providers/drawer-host-provider';
+import { spacing } from '../theme/tokens';
+
+/** Nothing to open: the queue button is hidden for an anonymous viewer. */
+function noop() {}
+
+export const AnonymousClimbView = memo(function AnonymousClimbView({
+  climb,
+  boardConfig,
+}: {
+  climb: Climb;
+  boardConfig: BoardConfig;
+}) {
+  const { t } = useTranslation('session');
+  const router = useRouter();
+  const { systemColors } = useTheme();
+  const insets = useSafeAreaInsets();
+
+  // The angle pill stays live, and the drawer's board config is the only thing
+  // it moves — exactly what the signed-in deep-link path does, where an angle
+  // change rewrites `boardConfigOverride` and leaves the climb object alone.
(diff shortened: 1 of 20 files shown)


marcodejongh:
Ready for review. **26 checks pass, 11 skip, 1 red — `claude-review`, and it is not this change.** Its log shows `ANTHROPIC_API_KEY:` empty; the job ran fine on both PRs earlier in this same session (its findings are in the comments above and every one of them is addressed), and re-running it does not help from a PR branch. Repo-level credential, out of scope here — flagging rather than papering over it.

The aggregate `ci-status` gate is green on both.

marcodejongh:
Both review points addressed, one by fixing it and one by disagreeing — details so the next reader can judge.

**Side effect inside a state updater — fixed** (`864346cd9`). The check moved to a ref, so the nonce bump happens outside `setAngle`'s updater. Worth more than the tidiness: under Strict Mode the old shape advanced the nonce by two in development and one in production, which is how a re-open bug ends up unreproducible on a desk. The new test renders the view inside `StrictMode` and asserts the nonce advances by **exactly one** — mutation-tested by putting the old updater back (it fails with `expected 3 to be 2`). No other assertion in that file could see it: they all read the target's identity, not its value.

**`viewer="anonymous"` without `onSignIn` — not changed, deliberately.** A discriminated union on `PlayDrawerActionBarProps` would only hold if the same union were threaded through `PlayDrawer`, and `PlayDrawer` destructures both fields at the top of a 1,100-line component, where TypeScript cannot narrow them any more. The guard would end up enforced at one boundary and lost at the next, which reads as safety without being it. There is one anonymous call site, it is `AnonymousClimbView`, and `anonymous-climb-view.test.tsx` pins that it passes the gate's href through both the bar and the drawer.

**The BLE gate stands.** `ble-needs-fable-review` is in the body precisely because this suppresses the lightbulb and takes `setCurrentClimb` off the similar-climb tap. Marked ready for review, not for merge: Fable's BLE review and Marco's nod on epic #4358's standing rules are both still owed, and #4604 merges before #4605.

marcodejongh:
Both code notes read correctly; neither becomes a change, and here is the reasoning so a reviewer can overrule it cheaply.

**`LogbookSection` reading `useAuth()` instead of taking a prop.** This is the same trade this PR already made one file over, and deliberately in the same direction. `useFavoriteStatus` was gated *inside the hook* rather than at its one call site precisely so a future consumer cannot reintroduce the bug by forgetting — and the failure mode here is worse than the one there. Forget to pass `isAuthenticated` and the section does not go blank or error: it renders **"No tries yet. Get on it."** to someone who has never had a try to record, which is exactly the bug this branch fixes. A hidden context read cannot be forgotten; a prop can, silently, and only on the signed-out path nobody runs locally.

There is one production call site, `DeferredSections`, and it does hold `isAuthenticated` — so the prop version is available if the team prefers the explicitness. I would rather the invariant live where it cannot be dropped, and the prop pair (`userAscents` / `userAttempts`) stay what they are: viewer-scoped *data* the parent already fetched, not a permission.

**`useFavoriteStatus`'s new `BoardAdapterProvider` requirement.** Agreed on the blast radius, and it is the reason the JSDoc says it throws at the hook rather than at query time — a test that hits it without the provider fails loudly on the first render instead of returning a quietly empty heart. Existing suites are patched; the note is worth keeping for the next one.

**BLE gate stands** — `ble-needs-fable-review`, not auto-mergeable, and #4604 merges before #4605.

PR #4605, opened 2026-08-19
feat(mobile): render the climb, not the login form, for signed-out readers

### LINKED ISSUE #4478: Signed-out hand-off from a www climb page lands on a login form, not the climb

marcodejongh:
## Summary

Part 2 of 2 for #4478 — **stacked on #4604, review and merge that first.** Marco ruled option B: make the read-only climb view genuinely anonymous.

Today a signed-out visitor who follows "Climb this" from a www climb page lands on a sign-in form — 209 characters of body copy, which is where the epic's SEO funnel currently ends. This is the PR that changes production behaviour.

### The issue's stated blocker was not one

Board adoption is not a render input. `parseClimbRoutePath` returns the numeric tuple, `urlBoardConfig()` already builds a `BoardConfig` from the URL alone and is already preferred over the board's, board art is a pure local catalogue lookup, and the `climb` resolver takes no ctx at all. So adoption is **skipped for both signed-out branches** rather than gated on its outcome — its first act is a `CREATE_BOARD` the backend refuses without a session, and it was never needed to draw anything.

### The real blocker, and the shape that avoids it

The hand-off does `router.replace('/(tabs)/climbs')` and then `openPlayDrawer` → `/play`. Both are outside the anonymous allow-set, and `AuthProvider` re-reads `window.location.pathname` on every navigation — so the visitor gets bounced back to the login wall the moment the drawer opens.

So the route **does not navigate**. `BoardRouteHandoff` renders `AnonymousClimbView` in place, on the canonical `/…/{angle}/view/{segment}` URL the visitor arrived at. That also keeps the address bar equal to the URL www linked, and the `?next=` builder correct for free. The anonymous surface is **not** widened to the player route.

`read-only-routes.test.ts` gains a named, corpus-independent assertion that `/play` and `/climbs/{uuid}` stay gated. Moving an entry between `GATED_PATHS` and `READ_ONLY_PATHS` leaves every corpus-driven suite green — they are all `it.each` over whichever array now holds it — so only a named literal goes red.

### Scope, and the four judgement calls inside it

Climb URLs only. Both trees: the config tuple (all ~128k sitemap URLs) and `/b/{slug}` (a shared gym-board link hitting the login wall while a search result does not is exactly the gradient this removes). List surfaces still gate — they need a board, and minting one is `requireAuthenticated` — so `/…/list` keeps its `?next=` login redirect (follow-up #4607).

1. **`/…/{angle}/play/{segment}` opens anonymously too, deliberately.** The two route files differ only in the surface they hand `buildBoardClimbTarget`, which drops it, so both produce the same `kind: 'climb'` target and the predicate cannot tell them apart. Left that way: `/play/…` is already in the read-only allow-set, older shares still emit it, and its readers would otherwise be the only arrivals left at the login form. Tested through the real builder. The app's own `/play` **modal** route stays gated.
2. **An unresolved `/b/{slug}` asks for a sign-in, never a dead end.** `boardBySlug` masks "you may not see this board" and "there is no such board" to the same `null`, so a not-found there would tell a gym member their own board does not exist — a regression on a URL that yesterday simply asked them to sign in and then showed them the climb. Sign-in is the only answer right in both cases, and the only one that resolves the ambiguity. A typo'd slug costs one login form, which is what it cost before. "Climb genuinely gone" (a settled climb query with nothing in it) is still a real not-found.
3. **Draft climbs render, and that is not new.** Checked rather than asserted: `fetchClimbFromDb` (`packages/web/app/lib/data/queries.ts`) selects by `(board_type, layout_id, uuid)` with no draft predicate, and `/[board_name]/…/view/[climb_uuid]/page.tsx` renders `ClimbFrontDoor` from it with no session — so a draft uuid already paints a full public page on **www**, the indexed surface. This PR adds a second surface for the same row on a `noindex` subdomain; it does not widen what is reachable, and it deliberately does not add an `is_draft` fall-through here, because "www shows you the climb, the app says it does not exist" is a worse answer than the one gap. Filed as #4608 to close at the resolver, where both surfaces read it.
4. **The drawer's queue is hidden, not local-only.** A queue a signed-out visitor cannot carry to a wall or a session is a dead affordance, and Similar Climbs — which stays — now swaps the drawer's preview instead of writing that queue (#4604).

The slug read is server-only on purpose: the device's stored board belongs to whoever last signed in on this browser.

### Rollout is a kill switch, not a percentage

There is no staged rollout to perform. This surface ships to **app.boardsesh.com via Cloudflare Pages**, not OTA, and the sanctioned publish wrapper drops `--rollout-percentage` anyway — merging *is* the publish, at 100%.

So the instrument is a flag: `anonymous-climb-view-kill`, **created and verified in PostHog** ([flag 832829](https://us.posthog.com/project/412845/feature_flags/832829)), active at 0% = feature on. It gates only the `AnonymousClimbView` branch. Flags resolve asynchronously, so an unresolved read has to mean *not killed* — a positive rollout flag would show an anonymous visitor a login-redirect flash as the first frame of the very surface this builds. `docs/feature-flags.md` gains that inversion as a documented pattern, because the file's standing rule ("unresolved leaves the control hidden") says the opposite.

**Recovery runbook, in the order to reach for it:**

| | Move | Reaches production in | Blast radius |
| --- | --- | --- | --- |
| 1 | Flip `anonymous-climb-view-kill` to 100% in PostHog | seconds, no deploy | exactly this branch; signed-out climb URLs go back to the login wall |
| 2 | Revert the merge commit → CI → `production-deploy.yml` rebuilds and `wrangler pages deploy`s the app subdomain | one green pipeline | the whole app subdomain moves to the reverted tree |
| 3 | Set the repo variable `APP_WEB_DEPLOY_HOLD` (any non-empty value) and roll back the Cloudflare Pages deployment | immediate | **freezes the entire subdomain** — every other app-web change is stranded until it is cleared, and the workflow surfaces the hold in the run summary so it cannot be forgotten quietly |

The native fleet needs none of these: the branch is unreachable there (below).

### Three fixes from the adversarial review

**The kill switch's direction had no oracle.** Every test that touches `useAnonymousClimbViewEnabled` mocked it away — `BoardRouteRedirect.test.tsx` stubs the hook, the gate tests pass `anonymousClimbEnabled` straight in — so flipping `!== true` to `=== false` left the whole mobile suite green while shipping the feature dead: the flag reads `undefined` until PostHog resolves, and forever for a browser that blocks it. Inverting it the other way turns the emergency switch into an enable switch. Three cases now pin all three states against a real `FeatureFlagsProvider`, and the *unresolved* one is the assertion that separates the two directions.

**`isAngleAdjustable` was resolved and then dropped.** A `/b/{slug}` climb resolves a real board record, the only place that flag exists, but `BoardRouteResult` carried only `{ status, climb, boardConfig }` — so an anonymous reader on a gym wall bolted at one angle still got an angle pill. It rides through now. A config tuple has no board record to read it off, so it keeps the `true` `createBoard` would have stored.

**A still-loading anonymous climb could have hit a dead end.** `resolveStatus` refuses to say `anonymous-climb` before the climb lands, but that guard had no test — and the fall-through was not benign: `BoardRouteHandoff` renders the view only when it holds both a climb and a config, and the redirector's only spinner branch was `resolving`, so a loosened guard would paint "Not found" plus a Back-to-home button on a URL one round trip from rendering. The guard has an oracle, and the redirector waits on that status instead of falling through.

### Telemetry

`Board Route Handoff` gains an `anonymous` status **value**, not a new event name, so `Climb Handoff Clicked` ÷ `Board Route Handoff` keeps counting the whole hop and anonymous-vs-signed-in becomes a breakdown.

(The issue's telemetry claim is stale, measured the day after #4418 merged. Over the last 7 days `Climb Handoff Clicked` fired ×10 and `Board Route Handoff` ×23 on native plus ×8 on web — both halves of the funnel are live.)

**Split out, not fixed here:** the app half of that funnel never sets the `environment: 'production-web'` its own contract documents — it registers the build environment (`production` / `preview`) on native and browser alike, so app.boardsesh.com is distinguishable only by a null `$app_version`. Unrelated to the auth boundary, so it went to **#4624** rather than into this diff.

### SEO

`app.boardsesh.com` serves `X-Robots-Tag: noindex` for `/*` (`deploy/app-subdomain/_headers`, verified on this branch), so an anonymously-rendered climb there competes with nothing on www. If that ever changes, the app climb view needs a `rel=canonical` back to the www climb page before it is indexed.

Closes #4478.

## Review gates

- **`ble-needs-fable-review`.** The stack suppresses the lightbulb and stops a similar-climb tap re-arming the BLE auto-sender; Web Bluetooth is live on the browser export. **Fable review required before merge**, on top of the standing Fable review every `packages/mobile` PR gets under epic #4358.
- **Needs Marco's explicit nod on epic #4358's standing rules** before merge (W-06 follow-up; publishes to the native fleet as an inert OTA).
- Not auto-mergeable.

### Production targets

**app.boardsesh.com** — where the behaviour change lands, via Cloudflare Pages. And **the native store fleet via auto-OTA**: this touches `packages/mobile/**` and `packages/shared/analytics`, so merging to `main` auto-publishes a production OTA that reaches every installed binary within hours. Not `www.boardsesh.com`.

**No native fingerprint inputs changed** — nothing under `app.config.ts`, `plugins/**`, `modules/**`, `locales/**`, `patches/**`, and no dependency edits — so `main` stays OTA-compatible with the store binaries, and the OTA check reports both platforms matching `main`.

**Native behaviour is unchanged, and that is the thing to scrutinise.** The whole anonymous branch sits behind `RELAXES_ANONYMOUS_ROUTES`, a literal `false` in the native fork — never a `Platform.OS` check, which would put the web behaviour in the native bundle with one boolean between the fleet and an anonymous route tree.

## Release Notes

- Find a climb on Google, tap through, and you see the climb — board, grade, community stats and beta — before anyone asks you to sign in.
- Sign in from there and you land straight back on the climb you came for.

## Test plan

- `vp run test:mobile` — green.
- `vp test run --project i18n catalog-completeness --reporter=agent` — 96 green (four-locale parity; `--project web` does not run this file).
- `vp run check:i18n`, `vp run check:i18n:orphans` — clean.
- `vp check` — 0 errors (324 pre-existing warnings), formatted.
- `vp run check:mobile-bundle`, `vp run check:mobile-web-bundle` — both OK.

**Every new oracle was mutation-tested** — mutation applied, test watched go red, reverted:

| Mutation | Caught by |
| --- | --- |
| **Drop `RELAXES_ANONYMOUS_ROUTES` from the anonymous condition** (the fleet mutation) | the native-invariance test |
| Adoption runs anonymously | 3 tests |
| Climb query re-muzzled for the anonymous branch | 2 tests |
| Kill switch ignored | the kill-switch test |
| Kill switch read once and cached (no mid-session flip) | the mid-session flip test |
| List URLs relax too | 6 tests |
| Slug climbs excluded again | 4 tests |
| An unresolved slug reported as not-found (private board reads as deleted) | 3 tests |
| Anonymous slug read falls back to the device's stored board | the stored-board test |
| URL angle ignored for the anonymous slug board | the angle test |
| A null `boardBySlug` reads as still resolving | 2 tests |
| Hand-off no longer skipped anonymously | the slug-climb test |
| A rejected `fetchBoardBySlug` left unsettled (permanent spinner) | the failed-lookup test |
| The view rendered on the status alone, with no climb in hand | the "waits rather than drawing" test |
| **Kill switch inverted any of four ways** (`=== false`, `=== true`, `!== false`, or ignored entirely) | `feature-flags-provider.test.tsx`, three new cases |
| A fixed-angle gym board's `isAngleAdjustable` dropped at the hook boundary | 1 hook test + 2 redirector tests |
| A config-tuple URL defaulted to no-tilt | the tuple test |
| `anonymousClimbReady` loosened to the branch alone | the still-loading test |
| `anonymous-climb` falling back to the not-found dead end | the redirector's spinner test |

The native-invariance oracle is phrased **positively** — `setActiveBoard` is still called with the resolved board and `openClimbInPlayDrawer` still fires once. The pre-existing `expect(status).not.toBe('auth-required')` stays **green** under exactly the mutation it appears to catch, because the anonymous branch is not `'auth-required'` either. That trap is why this one asserts what native still does, not what it doesn't.

### QA plan (browser, fresh incognito, `vp run dev:mobile:web`)

Not yet exercised end to end — the highest-value gate left on this stack, and the reason it stays draft until a human has driven it:

1. Signed-out tuple climb URL renders the climb, not the login form.
2. Signed-out `/b/{slug}/{angle}/view/{segment}` renders it too; a private/unknown slug shows the login form, not a not-found.
3. A locale-prefixed URL (`/es/kilter/…`) behaves the same and reads correctly in Spanish.
4. `/…/{angle}/play/{segment}` behaves identically to `/view/…`.
5. Write affordances absent: tick opens login, no queue button, no heart, no ellipsis, no lightbulb. Similar Climbs swaps the drawer without touching the queue.
6. Sign in from the bar → land back on the same climb.
7. A `/…/list` URL still redirects to `/auth/login?next=…` with a valid `next`.
8. A dead climb uuid shows not-found, not a spinner and not a login form.
9. How the pane-presentation drawer reads full-width in a browser — it has only ever been an iPad right column.

### Follow-ups filed

- #4606 — anonymous wall lighting (the LED path, hidden for v1 per Marco's ruling).
- #4607 — the anonymous list surfaces.
- #4608 — draft climbs fetchable by uuid with no viewer check.
- #4609 — five catalogued mobile feature flags that do not exist in PostHog.
- #4624 — the app half of the hand-off funnel never sets `environment: production-web`.

### DIFF
--- a/docs/feature-flags.md
+++ b/docs/feature-flags.md
@@ -4,6 +4,8 @@ Two kinds, and they fail in different ways.
 
 **Client flags** (`FEATURE_FLAG_KEYS` in `packages/web/app/flags.ts`) resolve in the browser via `FeatureFlagsProvider`. They hide a control. If PostHog is slow or unreachable the control stays hidden and nothing else happens.
 
+**Kill switches invert that default, deliberately.** A flag named `*-kill` gates a shipped feature and an unresolved read means _not killed_, i.e. the feature is on — the opposite of the rule above. The reason is the same asynchrony: a positive flag reads as OFF for the first frames of a cold open, which for a surface whose off-state is a redirect means the visitor watches a redirect flash before the flag lands. `anonymous-climb-view-kill` (mobile, `useAnonymousClimbViewEnabled`) is the worked example — its off-state is the login wall the feature exists to remove, so a flash of it on arrival would defeat the whole surface. Read a kill switch as `useFeatureFlag(key) !== true`, never `=== false`, so a missing key and an unresolved one both mean on.
+
 **Server flags** (`SERVER_FEATURE_FLAG_KEYS`) resolve during SSR via `getServerFeatureFlag` and decide whether a route renders at all — with one off, its route is a plain `notFound()`. A client-resolved flag could not do this: by the time the browser knows the answer the HTML has already shipped.
 
 The registry is **empty today**. `gyms-directory` was the only server flag; `/gyms` and its three facet routes now render for everyone and the flag is gone. The machinery below stays for the next route that needs it — and the diagnostics endpoint still answers any key you name.
--- a/packages/mobile/src/components/BoardRouteRedirect.tsx
+++ b/packages/mobile/src/components/BoardRouteRedirect.tsx
@@ -113,8 +113,15 @@ export function BoardRouteRedirect({ status }: { status: BoardRouteStatus }) {
       <Stack.Screen options={{ headerShown: false }} />
       {/* A handed-off screen keeps this spinner on its way out — nothing flips it
           to the not-found, because the board and climb it handed off with are
-          both still there. */}
-      {status === 'resolving' ? (
+          both still there.
+
+          `anonymous-climb` reaching HERE means the caller had the status but not
+          the climb — a state `resolveStatus` refuses to produce today (it only
+          says `anonymous-climb` once the climb has landed). It waits on the
+          spinner rather than falling through, because the alternative reading of
+          "the climb has not arrived yet" is "Not found" plus a Back-to-home
+          button, on a URL that is about to resolve. */}
+      {status === 'resolving' || status === 'anonymous-climb' ? (
         <ActivityIndicator size="large" />
       ) : (
         <>
@@ -145,7 +152,11 @@ export function BoardRouteHandoff({ target, mode }: { target: BoardRouteTarget |
   // `RELAXES_ANONYMOUS_ROUTES` is false, so the flag cannot reach a decision
   // there however it resolves.
   const anonymousClimbEnabled = useAnonymousClimbViewEnabled();
-  const { status, climb, boardConfig } = useBoardRouteTarget(target, { mode, onHandedOff, anonymousClimbEnabled });
+  const { status, climb, boardConfig, isAngleAdjustable } = useBoardRouteTarget(target, {
+    mode,
+    onHandedOff,
+    anonymousClimbEnabled,
+  });
 
   const renderedStatus = RENDERED_EVENT_STATUS[status as keyof typeof RENDERED_EVENT_STATUS];
   const parsedUrl = target !== null;
@@ -165,7 +176,7 @@ export function BoardRouteHandoff({ target, mode }: { target: BoardRouteTarget |
   // right here, on the URL the visitor arrived at, because every route the
   // hand-off would navigate to is behind the login gate.
   if (status === 'anonymous-climb' && climb && boardConfig) {
-    return <AnonymousClimbView climb={climb} boardConfig={boardConfig} />;
+    return <AnonymousClimbView climb={climb} boardConfig={boardConfig} isAngleAdjustable={isAngleAdjustable} />;
   }
 
   return <BoardRouteRedirect status={status} />;
--- a/packages/mobile/src/components/BoardRouteRedirect.tsx
+++ b/packages/mobile/src/components/BoardRouteRedirect.tsx
@@ -13,9 +13,11 @@ import { useTranslation } from 'react-i18next';
 import { onlineManager } from '@tanstack/react-query';
 import { SHARED_EVENTS } from '@boardsesh/analytics';
 import { ActivityIndicator } from './ActivityIndicator';
+import { AnonymousClimbView } from './AnonymousClimbView';
 import { Button } from './Button';
 import { Icon } from './Icon';
 import { Text } from './Text';
+import { useAnonymousClimbViewEnabled } from '../providers/feature-flags-provider';
 import { useTheme } from '../providers/theme-provider';
 import { track } from '../lib/analytics';
 import { buildLoginHrefWithReturn } from '../lib/routing/anonymous-auth-gate';
@@ -25,8 +27,16 @@ import { useBoardRouteTarget, type BoardRouteMode, type BoardRouteStatus } from 
 /** Where `app/+not-found.tsx` sends people; the dead end below has to match it. */
 const HOME_TAB = '/(tabs)/home' as const;
 
-/** How a board-route open ended, in the wire spelling. */
-type HandoffEventStatus = 'resolved' | 'not_found' | 'auth_required';
+/**
+ * How a board-route open ended, in the wire spelling.
+ *
+ * `anonymous` is a signed-out reader who got the climb rather than the login
+ * wall. A new VALUE rather than a new event on purpose: the funnel is
+ * `Climb Handoff Clicked` ÷ `Board Route Handoff`, both stamped
+ * `environment: production-web`, and a second event name would have split it.
+ * Anonymous-vs-signed-in arrivals are a breakdown of the same ratio.
+ */
+type HandoffEventStatus = 'resolved' | 'not_found' | 'auth_required' | 'anonymous';
 
 /**
  * The terminal statuses the screen SITS on, in the wire spelling.
@@ -38,6 +48,7 @@ type HandoffEventStatus = 'resolved' | 'not_found' | 'auth_required';
 const RENDERED_EVENT_STATUS = {
   'not-found': 'not_found',
   'auth-required': 'auth_required',
+  'anonymous-climb': 'anonymous',
 } as const satisfies Partial<Record<BoardRouteStatus, HandoffEventStatus>>;
 
 /** The URL a report is about, so two different climbs each get their own event. */
@@ -130,7 +141,11 @@ export function BoardRouteRedirect({ status }: { status: BoardRouteStatus }) {
 export function BoardRouteHandoff({ target, mode }: { target: BoardRouteTarget | null; mode?: BoardRouteMode }) {
   const report = useBoardRouteHandoffReporter(target, mode);
   const onHandedOff = useCallback(() => report('resolved'), [report]);
-  const status = useBoardRouteTarget(target, { mode, onHandedOff });
+  // Read unconditionally, applied only where the gate already relaxes: on native
+  // `RELAXES_ANONYMOUS_ROUTES` is false, so the flag cannot reach a decision
+  // there however it resolves.
+  const anonymousClimbEnabled = useAnonymousClimbViewEnabled();
+  const { status, climb, boardConfig } = useBoardRouteTarget(target, { mode, onHandedOff, anonymousClimbEnabled });
 
   const renderedStatus = RENDERED_EVENT_STATUS[status as keyof typeof RENDERED_EVENT_STATUS];
   const parsedUrl = target !== null;
@@ -145,6 +160,13 @@ export function BoardRouteHandoff({ target, mode }: { target: BoardRouteTarget |
     if (renderedStatus === 'not_found' && parsedUrl && !onlineManager.isOnline()) return;
     report(renderedStatus);
   }, [parsedUrl, renderedStatus, report]);
+
+  // The one status this component does not redirect for: the climb is drawn
+  // right here, on the URL the visitor arrived at, because every route the
+  // hand-off would navigate to is behind the login gate.
+  if (status === 'anonymous-climb' && climb && boardConfig) {
+    return <AnonymousClimbView climb={climb} boardConfig={boardConfig} />;
+  }
 
   return <BoardRouteRedirect status={status} />;
 }
--- a/packages/mobile/src/components/__tests__/BoardRouteRedirect.test.tsx
+++ b/packages/mobile/src/components/__tests__/BoardRouteRedirect.test.tsx
@@ -4,6 +4,7 @@ import { createElement, type ReactNode } from 'react';
 import { beforeEach, describe, expect, it, vi } from 'vitest';
 import { SHARED_EVENTS } from '@boardsesh/analytics';
 import type { BoardRouteTarget } from '../../lib/routing/board-route-target';
+import type { Climb } from '@boardsesh/shared-schema';
 import type { BoardRouteStatus } from '../../lib/routing/use-board-route-target';
 
 // What the gate hands back, as a sentinel. This suite asserts the component
@@ -13,6 +14,10 @@ const LOGIN_HREF = vi.hoisted(() => '/auth/login?next=%2Fb%2Fthe-gym%2F40%2Flist
 
 const analytics = vi.hoisted(() => ({ track: vi.fn() }));
 const routeStatus = vi.hoisted(() => ({ current: 'resolving' as BoardRouteStatus }));
+// The kill switch, as the component reads it. `true` = feature on.
+const anonymousClimbView = vi.hoisted(() => ({ enabled: true, seen: [] as (boolean | undefined)[] }));
+// What the anonymous branch renders instead of a redirect.
+const anonymousView = vi.hoisted(() => ({ props: [] as Record<string, unknown>[] }));
 const redirect = vi.hoisted(() => ({ hrefs: [] as string[] }));
 const router = vi.hoisted(() => ({ replace: vi.fn() }));
 // The hand-off callback the hook holds. A successful hand-off never becomes a
@@ -55,12 +60,27 @@ vi.mock('../../lib/routing/anonymous-auth-gate', () => ({
   buildLoginHrefWithReturn: () => LOGIN_HREF,
 }));
 vi.mock('@tanstack/react-query', () => ({ onlineManager: { isOnline: () => connectivity.isOnline } }));
-vi.mock('../../lib/routing/use-board-route-target', () => ({
-  useBoardRouteTarget: (_target: unknown, options?: { onHandedOff?: () => void }) => {
-    handoff.current = options?.onHandedOff ?? null;
-    return routeStatus.current;
+vi.mock('../../providers/feature-flags-provider', () => ({
+  useAnonymousClimbViewEnabled: () => anonymousClimbView.enabled,
+}));
+vi.mock('../AnonymousClimbView', () => ({
+  AnonymousClimbView: (props: Record<string, unknown>) => {
+    anonymousView.props.push(props);
+    return createElement('div', { 'data-testid': 'anonymous-climb-view' });
   },
 }));
+vi.mock('../../lib/routing/use-board-route-target', () => ({
+  useBoardRouteTarget: (_target: unknown, options?: { onHandedOff?: () => void; anonymousClimbEnabled?: boolean }) => {
+    handoff.current = options?.onHandedOff ?? null;
+    anonymousClimbView.seen.push(options?.anonymousClimbEnabled);
+    return routeStatus.current === 'anonymous-climb'
+      ? { status: routeStatus.current, climb: ANONYMOUS_CLIMB, boardConfig: ANONYMOUS_BOARD_CONFIG }
+      : { status: routeStatus.current, climb: null, boardConfig: null };
+  },
+}));
+
+const ANONYMOUS_CLIMB = { uuid: '0A1B2C3D4E5F60718293A4B5C6D7E8F9', name: 'Crimpy Thing' } as unknown as Climb;
+const ANONYMOUS_BOARD_CONFIG = { boardName: 'kilter', layoutId: 1, sizeId: 10, setIds: '1,20', angle: 40 };
 
 const { BoardRouteHandoff, BoardRouteRedirect } = await import('../BoardRouteRedirect');
 
@@ -82,6 +102,9 @@ beforeEach(() => {
   routeStatus.current = 'resolving';
   handoff.current = null;
   connectivity.isOnline = true;
+  anonymousClimbView.enabled = true;
+  anonymousClimbView.seen = [];
+  anonymousView.props = [];
 });
 
 describe('BoardRouteRedirect', () => {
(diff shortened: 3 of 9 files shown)


marcodejongh:
### Coordination note + what changed in the last two commits

Two agent sessions have been pushing to this stack in the last half hour (`session_013WxihGrh23BbjcyGvc5mW5` and `session_016552Be7ctNKKH8QM9rkryN`), which is why the kill-switch-flip test and the `pending` state landed twice in different shapes. Converged now — the branches hold one version of each. Anything below is the second session's delta; it is deliberate, not a stray revert.

**One behaviour change worth a second look: an unresolved `/b/{slug}` climb URL now asks for a sign-in instead of showing not-found.**

`boardBySlug` masks "you may not see this board" and "there is no such board" to the same `null`. Treating that as a dead link tells a gym member their own board does not exist — and it is a *regression* against the URL's behaviour before this stack, where a signed-out visitor got the login form and, after signing in, the climb. Sign-in is the only answer that is right in both cases, and the only one that turns the ambiguity into a real answer. A typo'd slug costs one login form, which is what it cost before. A climb that is genuinely gone (settled query, nothing in it) is still a real not-found.

`useAnonymousSlugBoard` now returns `pending | resolved | unresolved` instead of a board-plus-error pair, for the same reason the error flag was spelled out a commit earlier.

Also pinned, both previously true but unasserted:

- **`/…/{angle}/play/{segment}` renders anonymously too.** Both route files hand `buildBoardClimbTarget` a surface it drops, so `/view/…` and `/play/…` produce the same target and the predicate cannot tell them apart. That is deliberate — `/play/…` is in the read-only allow-set and older shares still emit it — so it is now tested through the real builder rather than a hand-made target. The app's own `/play` **modal** route stays gated (`read-only-routes.test.ts`).
- **The anonymous view is not drawn on the status alone.** Status and climb are separate fields with nothing in the type system stopping them disagreeing, and a drawer mounted without a climb is a blank screen with nothing on it to explain itself.

On #4604: a similar-climb tap no longer writes the queue or calls `setCurrentClimb` for a signed-out reader — the second is what re-arms the BLE auto-sender, so it was the one affordance left in the anonymous view that could still drive a wall.

Split out rather than folded in: #4624, the app half of this funnel never setting the `environment: production-web` its own contract documents.

marcodejongh:
Ready for review. **26 checks pass, 11 skip, 1 red — `claude-review`, and it is not this change.** Its log shows `ANTHROPIC_API_KEY:` empty; the job ran fine on both PRs earlier in this same session (its findings are in the comments above and every one of them is addressed), and re-running it does not help from a PR branch. Repo-level credential, out of scope here — flagging rather than papering over it.

The aggregate `ci-status` gate is green on both.

marcodejongh:
### Adversarial-review round: 11 findings, 10 fixed, 1 rebutted

Pushed to both branches. Every new oracle was mutation-tested — mutation applied, test watched go red, reverted — and the two real defects the review surfaced were reproduced before being fixed.

**On #4604 (`7bfd6bbd`)**

- **`PlayDrawer` is now rendered in a unit test.** It never had been; both existing suites mock it, so every anonymous rule was pinned one hop from the code that ships it. Six mutations at the call sites survived the whole 6576-test suite before `play-drawer-anonymous.test.tsx`: `viewer={viewer}` → `viewer="member"`, the similar-climb preview guard deleted, `onSignInPress` dropped, the preview banner ungated, the favourites gate dropped, and the pane preview seed nulled. All six red now.
- **Real defect: the anonymous drawer offered a live "Set active".** It is always in preview mode — that is what keeps the queue untouched — so the banner rendered on every read-only open, and its press is `setCurrentClimb`: the same queue write and the same BLE auto-sender re-arm the similar-climb guard was written to block. Gated on the viewer.
- **Real defect: the angle pill moved the board and nothing else.** Grade, stars and ascent count are baked into the climb row at the angle it was fetched for (`getClimbByUuid` joins the stats tables ON angle), the route's query is frozen at the URL angle, and `useEffectiveClimbStats` — what keeps the signed-in header honest across an angle change — is `requireAuthenticated` and never fires here. The header held the URL angle's numbers while the Boardsesh grade, the crowd-grade bar and the angle sheet all moved. The climb is refetched at the live angle now, with the one on screen still rendering while that is in flight.

**On this PR (`7367a8f9`)**

- **The kill switch's direction had no oracle at all.** Every test that touches it mocks it away, so inverting `!== true` shipped the feature dead (or turned the emergency switch into an enable switch) with the suite green. Three cases now pin all three states against a real `FeatureFlagsProvider`; the *unresolved* one is what separates the directions.
- **`isAngleAdjustable` was resolved and dropped.** A `/b/{slug}` climb resolves a real board record — the only place that flag exists — but `BoardRouteResult` did not carry it, so an anonymous reader on a bolted gym wall still got an angle pill.
- **The still-loading anonymous climb had no guard test,** and the fall-through was not benign: the redirector's only spinner branch was `resolving`, so a loosened `anonymousClimbReady` would have painted "Not found" over a URL one round trip from rendering. Guard pinned; redirector waits on that status.

**Rebutted, with measurement: "draft climbs render on the new surface".** Checked rather than argued — `fetchClimbFromDb` (`packages/web/app/lib/data/queries.ts`) selects by `(board_type, layout_id, uuid)` with no draft predicate, and `view/[climb_uuid]/page.tsx` renders `ClimbFrontDoor` from it with no session. A draft uuid already paints a full public page on **www**, the indexed surface. This stack adds a second surface for the same row on a `noindex` subdomain; it widens nothing. A guard here alone would mean "www shows you the climb, the app says it does not exist" — worse than the one gap. #4608 stays the place to close it, at the resolver both surfaces read.

Also rebutted in part: the suggestion to default the *tuple* branch's angle adjustability "to whatever the board catalogue says". The catalogue has `ANGLES` per board **type** (which angles exist), not whether a given physical wall can tilt — that only lives on the board record. A tuple URL has none, so it keeps the `true` `createBoard` would have stored.

**Gates:** `vp run typecheck:mobile`, `vp run test:mobile` (6600), `--project i18n catalog-completeness` (96), `check:i18n`, `check:i18n:orphans`, `vp check` (0 errors), `check:mobile-bundle`, `check:mobile-web-bundle` — all green. On CI, 23 checks pass and `ci-status` is green on both PRs.

`claude-review` is still the one red, and still not this change: its log for this push reads `Claude Code returned an error result: You've hit your session limit · resets 2:40am (UTC)`, with `ANTHROPIC_API_KEY:` empty in the step env. It is red on **both** PRs of the stack for the same account-level reason. A re-run is scheduled for after that window; it is a repo credential/quota matter, not something a branch can fix.

Both PRs are back in **draft** pending the Fable BLE review and Marco's nod on epic #4358's standing rules.

marcodejongh:
### Restored a commit that a force-push dropped — flagging it rather than slipping it back in

Two agent sessions were pushing to this stack (`session_013WxihGrh23BbjcyGvc5mW5` and `session_016552Be7ctNKKH8QM9rkryN`). A rebase of this branch force-updated `2651f0cdb` out of existence; `2843a6349` puts its content back, cherry-picked onto the current head as a fast-forward. **Nothing from the other session was dropped or rewritten** — every commit it pushed is still in the history, and where the two overlapped I kept its version:

- the "no climb in hand" guard: its `climbHasLanded` test asserts the spinner is what shows, which is stronger than the absence assertion I had. Mine is gone, its stands.
- the kill-switch mid-session flip and the explicit `pending` state: both already landed in `ffbe72dcc`.

What actually comes back, and the reason it is worth a second look:

**An unresolved `/b/{slug}` climb URL now asks for a sign-in instead of showing not-found.** `boardBySlug` masks "you may not see this board" and "there is no such board" to the same `null`. Reporting that as a dead link tells a gym member their own board does not exist — and it is a regression against the URL's behaviour before this stack, where a signed-out visitor got the login form and, after signing in, the climb. Sign-in is the only answer that is right in both cases, and the only one that turns the ambiguity into a real answer. A typo'd slug costs one login form, which is what it cost before. A climb that is genuinely gone (settled query, nothing in it) is still a real not-found.

Also back: **`/…/{angle}/play/{segment}` is pinned as rendering anonymously**, built through the real `buildBoardClimbTarget` rather than a hand-made target. Both route files hand it a surface it drops, so `/view/…` and `/play/…` produce the same target and the predicate cannot tell them apart — deliberate, since `/play/…` is in the read-only allow-set and older shares still emit it. The app's own `/play` **modal** route stays gated.

If the not-found behaviour was a considered choice rather than an oversight, revert `2843a6349`'s source hunk and say so — but the sign-in path is what #4478's spec asks for, so it should be an explicit decision either way.

marcodejongh:
Answers to both minor points, and the process gates confirmed rather than waved through.

**`Stack.Screen` in the anonymous path — already there.** `AnonymousClimbView.tsx:128` renders `<Stack.Screen options={{ headerShown: false }} />` as its first child, for the same reason `BoardRouteRedirect.tsx:113` does: these routes mount on the root stack and a back chevron here would point at nothing. The early return is headerless, not header-defaulted. (Checked before answering — the component lives in #4604, so it is not visible in this diff.)

**Kill-switch mid-session test not asserting unmount — correct, and out of that test's reach.** `use-board-route-target.test.tsx` exercises the hook; it has no component tree to unmount. The rendering half is pinned next door in `BoardRouteRedirect.test.tsx`, which asserts the view is drawn only while the gate holds both a climb and a config, and shows the spinner rather than a dead end when it does not. Between them: the hook stops saying `anonymous-climb`, and the component stops rendering the view when it does. Added to the QA list as a visual confirmation.

**Process gates — all three still open, by design.** Fable's BLE review, Marco's nod on epic #4358's standing rules, and the browser QA pass (including item 9, the pane-presentation drawer full-width). Ready for review is not ready for merge here: `ble-needs-fable-review` is in the body, this is explicitly not auto-mergeable, and **#4604 merges first** — merging this one alone breaks the bundle, since `AnonymousClimbView` lives there.

One thing to weigh while reviewing, called out in the comment above: an unresolved `/b/{slug}` climb URL now lands on the login form rather than a not-found, because `boardBySlug` masks "private" and "missing" to the same null and only one of those is a dead end.

marcodejongh:
🤖 **Coordinator arbitration — two automation sessions disagreed on this, so recording the call and the reasoning.**

Both sessions worked these branches concurrently. The open disagreement: when `boardBySlug` returns `null` for a `/b/{slug}` climb URL, should a signed-out visitor get **sign-in** or **not-found**?

**Ruling: sign-in. Keep `ba544729b`.**

The reasoning is an asymmetry of harm, not a preference. `boardBySlug` masks "private", "gym-scoped", "typo" and "failed request" into one `null` — the route genuinely cannot tell them apart, and the commit is right to model that as a single named `unresolved` state rather than guessing.

Given that ambiguity:

- **not-found for a private board** tells a gym member their own board does not exist. It is terminal — there is no affordance to recover — and it *regresses* behaviour that worked before (login, then the climb).
- **sign-in for a genuinely dead link** costs one wasted sign-in, after which the visitor gets the real answer, including a legitimate not-found.

One is recoverable, the other is a dead end. Under ambiguity you take the recoverable branch.

This does mean the stronger-looking test asserting not-found is asserting the wrong behaviour, and should be updated rather than restored.

**The real fix is upstream and I have filed it separately:** `boardBySlug` should distinguish "not visible to you" from "does not exist" so this stops being a guess. Until then, sign-in is the correct default.

Unrelated to the ruling, for the record on the branch handling: `2651f0cdb` was dropped by a force-push and is now orphaned; I verified its content was re-applied as `ba544729b`. Overlaps were resolved in the *other* session's favour, which is the right instinct. Both sessions: prefer a fast-forward, and if you must reset to someone else's head, say so in a PR comment as was done here.

PR #4610, opened 2026-08-19
fix(shared): correct the Kilter 12x14 size description spelling

### LINKED ISSUE #4554: Fix the 'Commerical' typo in the generated Kilter product-size description
The Kilter 12 x 14 product size carries `description: 'Commerical'` in
`packages/board-constants/src/generated/product-sizes-data.ts:29` — a misspelling of "Commercial"
that comes straight from Aurora's product-sizes table.

It used to be invisible on mobile because the board-detail sheet's Size row was hidden: the backend
sends `sizeName` / `sizeDescription` as `null` on every `UserBoard`. #4541 makes the row render for
the first time by resolving the size from the bundled tables, so the typo is now user-facing on the
board-detail sheet.

## Why this isn't a one-line fix

`product-sizes-data.ts` is generated (the `generate` script in `packages/board-constants`), so a
hand-edit is overwritten on the next run and silently diverges from the source table. The fix belongs
in the generator: a small normalisation pass over size descriptions, applied where the data is
written rather than where it is read.

## Scope

- Add the correction to the generator (a known-typos map is enough — this is the only one spotted so
  far), regenerate, and check nothing keys off the raw string.
- Aurora's own app shows the same misspelling, so this is a deliberate divergence; worth a comment in
  the generator saying so.

marcodejongh:
## Summary

**Read this first: this changes a live URL segment, and the pin that keeps the old one working ships in the same commit.**

The Kilter Original 12 x 14's board URL segment goes from `12x14-commerical` to `12x14-commercial`. That segment is not cosmetic — `generateSizeSlug` folds the size *description* into it, so correcting the spelling re-mints every board URL for that size. PostHog counted **629 pageviews from 325 people** on the old form in the 180 days to 2026-08-16, across `/list`, `/view`, `/playlists` and the `/es` and `/fr` variants. The Expo app's deep-link router resolves board segments purely from `@boardsesh/board-constants` with no database fallback, so an unpinned rename would drop those links on the floor.

So the old slug is pinned in `PERMANENT_SIZE_SLUG_ALIASES` (`kilter` → size 7 → `12x14-commerical`) in this same PR, with the traffic numbers in the comment so nobody prunes it later. That table is consulted *after* the generated slugs, so a pin can never shadow a live slug, and it still only resolves on a layout that actually carries size 7. This is the first *bare* (unqualified) slug in that table; it needs no special handling.

Only then, the actual typo. Aurora's own catalogue carries it: `board_product_sizes` (board_type `kilter`, id 7, "12 x 14") stores `description = 'Commerical'` verbatim, and their app shows it that way too.

### Why the fix is at the generator, and not in the database

`upsertProductSizes` in `@boardsesh/aurora-sync` re-writes `description` from Aurora on every catalog sync (`onConflictDoUpdate`), so a migration correcting the row would be reverted the next time the catalogue syncs. The durable layer is where we *write* our own copy of the data — the board-constants generator. New `packages/board-constants/src/size-description-overrides.ts` holds a one-entry, whole-string correction map (`Commerical` → `Commercial`); `querySizes()` applies it.

I deliberately did **not** normalise the description in `upsertProductSizes` too. Our DB mirror stays faithful to upstream: the copy there feeds only a URL fallback that never fires for kilter 1/7 and a `UserBoard.sizeDescription` field the backend hardcodes to `null`, so diverging it buys nothing user-visible — and it happens to leave web's `findSizeBySlug` DB leg as a second, independent resolver for the old slug. Please don't "finish the job" later.

I also checked the rest of the upstream catalogue against the dev DB: all 43 `board_product_sizes` rows, 16 `board_layouts` names and 44 `board_sets` names. `Commerical` is the only misspelling, which is why the map is a one-entry lookup rather than a word-level regex.

### About the hand-edited generated file

`packages/board-constants/src/generated/product-sizes-data.ts` says DO NOT EDIT, and this PR edits one line of it. That is deliberate, and it is safe only because of the two things shipping beside it:

1. The **generator is changed in the same PR** to emit exactly this value, so the next real regeneration produces the same string.
2. A **fixed-point test** pins the two together: for every size in `AURORA_PRODUCT_SIZES`, `normalizeSizeDescription(size.description) === size.description`. A regeneration that somehow bypassed the correction map reintroduces the typo *and fails CI* rather than doing it silently.

I did not run the generator. A full regeneration rewrites the `Generated at:` timestamp, all six hole-placement `.cjs` shards and the loader against whatever the local dev-DB image happens to hold — a large, unreviewable diff for a one-word fix.

### Not adding a 301

Old URLs keep serving 200 with a `rel=canonical` pointing at the corrected slug. That is exactly how every other alias spelling is already treated (the bare-dimension `12x12` form, the `kilter-original` layout prefix), and the numeric-only redirect in the board layout deliberately doesn't fire for readable URLs. Consistency beats a marginally cleaner SEO signal; happy to raise redirect consolidation separately if the SEO owner wants it.

Drive-by: `CLAUDE.md` / `AGENTS.md` listed `board-constants` under `packages/shared/`. It lives at `packages/board-constants/`. Fixed while I was in there.

Closes #4554

## Release Notes

- The Kilter 12 x 14 now reads "Commercial" instead of "Commerical" in the board picker and in board links.
- Links you shared before the fix still open the same board.

## Test plan

- `vp test run --project board-constants --reporter=agent` — 10 files / 125 tests green, including the new `size-description-overrides` suite (the correction, the pass-through, the pinned `AURORA_PRODUCT_SIZES.kilter[7].description`, and the fixed-point guard).
- `vp test run --project play-view --reporter=agent` — 11 files / 247 tests green (240 before; 7 new). The pinned-alias list now asserts three entries, and a new `#4554` block proves `parseClimbRoutePath` and `parseBoardListPath` still resolve `/kilter/original/12x14-commerical/...` to size 7 while `resolveSizeSlug('kilter', 1, 7)` mints `12x14-commercial`.
- `vp test run --project web --reporter=agent packages/web/app/lib/__tests__/slug-utils-catalogue-round-trip.test.ts packages/web/app/lib/__tests__/url-utils.test.ts` — 254 tests green. The catalogue round-trip walks every `PERMANENT_SIZE_SLUG_ALIASES` entry through the real async `getSizeBySlug` with a **throwing** db stub, so it proves the old slug resolves off the static tables via the new pin, not via the database.
- `vp run test:mobile` — 684 files / 6548 tests green, including the updated share-URL expectation.
- `vp run typecheck`, `vp check`.
- No native-fingerprint inputs touched (`packages/board-constants`, `packages/shared/play-view`, `packages/mobile/src` only), so this rides OTA. No BLE code, so no Fable review needed.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -105,11 +105,11 @@ We'll always create a PR, never asks if a PR should be created, open as a draft.
   /mobile/          # React Native (Expo) mobile application
   /backend/         # WebSocket backend for party mode (graphql-ws)
   /shared-schema/   # Shared GraphQL schema and TypeScript types
+  /board-constants/ # Generated Aurora board catalogue (sizes, layouts, sets, holds), grade colours, difficulty bands
   /shared/
     /play-view/     # Play-drawer logic (queue nav, tick utils, grade display)
     /queue/         # Queue state machine (reducer, types, event utils)
     /board-config/  # Board metadata, hold maps, angle tables
-    /board-constants/ # Grade colours, difficulty bands
     /board-react/   # Renderer-agnostic BoardProvider + logbook/tick hooks (useSaveTick/useUpdateTick/useDeleteTick)
     /offline-sync/  # Offline sync engine: mutation outbox + drainer, pull client, SQLite DDL (platform I/O injected)
     /profile-stats/ # Pure climbing-stats aggregation for the You page / profile (chart builders, deriveProfileViewModel)
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -34,11 +34,11 @@ We'll always create a PR, never asks if a PR should be created, open as a draft.
   /mobile/          # React Native (Expo) mobile application
   /backend/         # WebSocket backend for party mode (graphql-ws)
   /shared-schema/   # Shared GraphQL schema and TypeScript types
+  /board-constants/ # Generated Aurora board catalogue (sizes, layouts, sets, holds), grade colours, difficulty bands
   /shared/
     /play-view/     # Play-drawer logic (queue nav, tick utils, grade display)
     /queue/         # Queue state machine (reducer, types, event utils)
     /board-config/  # Board metadata, hold maps, angle tables
-    /board-constants/ # Grade colours, difficulty bands
     /board-react/   # Renderer-agnostic BoardProvider + logbook/tick hooks (useSaveTick/useUpdateTick/useDeleteTick)
     /offline-sync/  # Offline sync engine: mutation outbox + drainer, pull client, SQLite DDL (platform I/O injected)
     /profile-stats/ # Pure climbing-stats aggregation for the You page / profile (chart builders, deriveProfileViewModel)
--- a/packages/board-constants/scripts/generate-board-constants.ts
+++ b/packages/board-constants/scripts/generate-board-constants.ts
@@ -10,6 +10,7 @@ import { existsSync, mkdirSync, writeFileSync } from 'fs';
 import { join } from 'path';
 import { fileURLToPath } from 'url';
 import { config } from 'dotenv';
+import { normalizeSizeDescription } from '../src/size-description-overrides';
 
 const __dirname = fileURLToPath(new URL('.', import.meta.url));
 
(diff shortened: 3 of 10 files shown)


marcodejongh:
That rule isn't in CLAUDE.md. Grepped the whole repo before acting:

```
$ grep -rniE "multi-paragraph|multi-line comment|one short line" CLAUDE.md AGENTS.md .claude/ docs/
(no matches)
```

CLAUDE.md has no comment-length or docstring rule at all. The house style is the opposite: the file I'm editing, `readable-url-utils.ts`, already carries multi-paragraph docstrings on `resolveSizeSlug`, `sizeNameQualifierSlug` and `PERMANENT_SIZE_SLUG_ALIASES` itself, and my additions match them.

Keeping the comments as written. The `PERMANENT_SIZE_SLUG_ALIASES` entry in particular has to carry the 629-pageview/325-people figure and the reason the pin exists — that table is append-only and the next person to read it needs to know deleting the line 404s live links.

Issue #4612, opened 2026-08-19
356 of 380 bundled OTA assets are board backgrounds — they set the floor for every publish burst and OTA download


marcodejongh:
Split out of #3620.

## The number

A current mobile export is **380 assets**. **356 of them** are board-background images:

```
$ grep -c 'require(' packages/mobile/src/lib/board-backgrounds-manifest.ts
356
```

The manifest `require()`s them across package boundaries out of `packages/web/public/images/**`, so every one lands in the OTA bundle. That's **94% of the asset count** of every update we publish.

## Why it matters twice

**Publish cost.** Storage keys are `{appId}/{branch}/{runtimeVersion}/{updateId}/assets/{hash}` — `updateId` is in the path, so before xprem's server-side asset reuse (v3.1.2, pending a Railway image bump) every publish wrote a fresh full copy: ~760 PUTs for a two-platform run. That volume is what trips Tigris `SlowDown`, the subject of #3620. Even with reuse landed, the *first* publish to each new branch — every new `pr-<n>` preview channel — still pays the full 380.

**Download cost.** Every user pulling an OTA downloads the changed assets. The board art is the floor under that size for anyone on a fresh runtimeVersion.

## The question

The art is deliberate — it's what makes offline board rendering work, and that shipped for real (see the offline-mode epic #4319). The question is narrower: does the **full set** need to be in the JS bundle, or only the boards a given user actually has?

Options worth pricing before picking one:

- Ship only the layouts/sizes reachable from the user's own boards, and pull the rest from the existing snapshot CDN on demand (the board-snapshot infrastructure in `docs/board-snapshots.md` already distributes per-board data).
- Keep everything bundled but cut per-image weight — these are webps already, so this is a re-encode/resolution question, not a format one.
- Keep as-is and rely on server-side asset reuse to make the repeat-publish cost ~zero, accepting the first-publish-per-branch cost.

## Not urgent

Measure before changing anything. #3620's client-side rate cap plus the pending `xprem:v3.1.2` server bump may take the publish-side pain to zero on their own, which would leave only the download-size argument — a weaker but still real one. Re-measure a repeat publish's upload step after the image bump before spending effort here.

Issue #4611, opened 2026-08-19
Restrict per-PR OTA preview auto-publish: ~30 publishes/day, up to 11 concurrent, for previews almost nobody opens


marcodejongh:
Split out of #3620 — a decision for @marcodejongh, not something an executor should change unilaterally, because it changes the tester workflow.

## What happens today

`.github/workflows/mobile-ota-preview.yml` publishes a `pr-<number>` OTA channel on `pull_request` types `[opened, synchronize, reopened, closed]`. Every push to a PR touching `packages/mobile/**`, `packages/shared/**`, `packages/board-constants/**`, `packages/shared-schema/**`, `package.json`, `bun.lock`, `patches/**`, or the three publish scripts fires a fresh two-platform publish.

The publish job's concurrency group is per PR:

```yaml
group: mobile-ota-preview-publish-${{ github.event.pull_request.number || github.event.issue.number || github.event.inputs.pr }}
```

so PRs don't queue behind each other — they all publish at once.

## Measured volume (2026-08-19, during a coordinated bug batch)

- **~30 preview publishes/day.**
- **Up to 11 publish jobs running concurrently** — peak 13:15–13:23 UTC, bucketing every non-skipped `mobile-ota-preview.yml` run of the previous three days into minutes.
- Each publish is **380 assets** (~760 PUTs across both platforms), and before xprem's server-side asset reuse none of it deduplicated across publishes — storage keys carry `{updateId}`.

That concurrent peak sat inside PR #4546's throttle window, where run [32249835065](https://github.com/boardsesh/boardsesh/actions/runs/32249835065) took 45m39s to publish (iOS succeeded on attempt 6 after five `SlowDown` responses).

Almost none of these previews are opened by anyone. They exist so a tester *can* switch to one from What's New — which is worth keeping — but the auto-publish-on-every-push part is what generates the volume.

## Proposal

Restrict auto-publish to `pull_request` types `opened` and `ready_for_review`, and keep the existing `/ota-preview` maintainer comment (plus `workflow_dispatch`) as the way to refresh a preview for a PR that's actively being tested. Fork PRs already work this way.

Trade-off: a tester who switched to `pr-N` no longer silently gets the latest push — someone has to ask for a refresh. Whether that's an acceptable cost is the actual question here.

## Explicitly not the fix

[...]

PR #4613, opened 2026-08-19
fix(ci): stop OTA publishes bursting past the storage rate limit

### LINKED ISSUE #3620: OTA publish fails outright on S3 SlowDown — needs upload retry/backoff and lower concurrency
During the BOARDSESH-AA hotfix train (2026-07-15), three production OTA publishes ran within ~30 minutes and the third hit Tigris rate limiting on `boardsesh-ota`:

```
❌ File upload failed <Error><Code>SlowDown</Code><Message>Please reduce your request rate.</Message>...
[mobile:publish] eoas publish failed.
```

Both the original run and an immediate `gh run rerun --failed` failed the same way (run 29387706795, commit c3a518b9c). The publish only succeeded after a cool-down. Consequences today: the flag-key-rotation bundle (#3619) reached devices ~30 minutes later than intended — harmless this time, but the same failure on a crash-fix OTA would delay an actual incident remedy, and a half-published run (one platform up, one down) can skew platforms until someone notices.

Wanted:
1. Retry with exponential backoff on `SlowDown`/5xx inside the publish step (either an eoas flag if upstream supports it, or a wrapper retry loop in `scripts`/workflow around `eoas publish`).
2. Consider capping upload concurrency — the burst of per-asset PUTs is what trips the rate limit.
3. The workflow should treat a platform's publish as atomic-ish: if iOS fails after Android succeeded, surface that clearly in the run summary.

Ref: `.github/workflows/mobile-ota-production.yml`, `docs/mobile-ota-updates.md`.

- [x] No release note needed

marcodejongh:
## Summary

**Measured, on this PR's own preview publish: `✓ publish in 9m29s`, zero `SlowDown` responses, both platforms on the first attempt.** The run this issue is about took **45m39s**. ([run 32312663175](https://github.com/boardsesh/boardsesh/actions/runs/32312663175))

```
[mobile:publish] Running ios: bunx eoas@3.1.2 publish --branch pr-4613 --platform ios … --upload-rate 5 …
› Assets (380):
📤 Uploading files...            23:25:42
✅ Files uploaded successfully   23:26:58     ← 76.2s
✅ Update ready for ios
[mobile:publish] Platform results: ios=success (1 attempt)

[mobile:publish] Running android: … --upload-rate 5 …
› Assets (417):
📤 Uploading files...            23:29:38
✅ Files uploaded successfully   23:31:02     ← 83.5s
[mobile:publish] Platform results: android=success (1 attempt)
```

The limiter lands within 0.2s of the model both times: 380/5 = 76.0s against 76.2s measured, 417/5 = 83.4s against 83.5s. The pacing *is* the whole cost, and it's ~80 seconds per platform. `scripts/mobile-publish.ts` and `scripts/lib/eoas.ts` are in `mobile-ota-preview.yml`'s path filter, so this ran against the **still-3.0.5 server** — which also makes it the wire-compatibility proof no unit test could give.

Closes #3620.

### Two corrections to the record

**The cited run didn't fail.** [32249835065](https://github.com/boardsesh/boardsesh/actions/runs/32249835065) (PR #4546) **succeeded** — 45m39s, iOS on attempt 6, Android on attempt 1. The retry ladder from #4194 is working exactly as designed. What it costs is **time**: five throttled attempts, each re-running a ~90s Metro export and re-firing the identical burst. This PR goes after the burst, not the wait.

**"It can't be done on our side" expired on 2026-08-19**, about four hours before that run. Our own docs said capping upload concurrency needed an upstream change and that waiting was our only lever. `eoas` / `xprem` **3.1.2** shipped both fixes the issue asked for; `--upload-rate` was always just a flag we could pass once it existed. `docs/mobile-ota-updates.md` is rewritten here so the next person doesn't re-derive the dead end.

### What throttles us

Three things multiply:

| | Measured |
|---|---|
| Assets per export | **380** (iOS) / **417** (Android), of which **356** are board-background webps `require()`d by `board-backgrounds-manifest.ts` |
| PUTs per publish | ~**760** for a two-platform run — keys carry `{updateId}`, so nothing deduped across publishes |
| Concurrent preview publishes | up to **11** at once (peak 13:15–13:23 UTC, 2026-08-19), each firing its own unbounded burst at the one `boardsesh-ota-v3` bucket |

`eoas ≤3.1.1` sent all of them through a single unbounded `Promise.all`, and its `fetchWithRetries` inspected only transport errors — never an HTTP status — so one `503 SlowDown` fell through `!response.ok` to `process.exit(1)` and killed the whole publish.

### What this PR does

- **Passes `--upload-rate 5`** on every self-hosted publish — `production` and the per-PR `pr-<n>` previews alike, since the previews are the concurrent ones. Not the CLI default of 10 because the token bucket is **per process**: at 11 concurrent jobs the default would still aim ~110 upload starts/sec at one bucket; at 5 that peak is ~55/sec. The EAS preview path (`eas update`) has no such flag and is untouched.
- **Pins `eoas@3.1.2`** (from `3.0.5`). It retries 429/5xx, honours `Retry-After`, backs off exponentially to 60s over four attempts, and rebuilds multipart bodies so a retried upload doesn't replay a consumed stream.
- **Adds a version-parity test.** Root `scripts/` has no typecheck task, so nothing catches a version string drifting off `EOAS_PACKAGE_SPEC` — the runbook can tell you to deploy one image while CI publishes with a different CLI.
- **Rewrites `docs/mobile-ota-updates.md`** — the stale "waiting is the only lever" section, the lockstep rule, the image tag, and the measured numbers above so they live on the page rather than in a run log.
- **Leaves the retry ladder and every `timeout-minutes` alone.** It costs nothing on a healthy publish, and `mobile-ota-publish-workflow.test.ts` derives three workflows' timeout floors from it. Its doc comment now says it's a backstop. Worth revisiting after a week of green publishes, not today.

### The version asymmetry in CLAUDE.md / AGENTS.md is deliberate

Both files carry two versions in one sentence, and this PR updates **only one of them**:

- `eoas@3.0.5` → **`eoas@3.1.2`**. That's the CLI pin, and this PR is what changes it — leaving it would be stale on merge.
- `expo-open-ota:v3.0.5` is **left as-is**. That's the image actually running on Railway, and it stays accurate until the hand-off below. Bumping it here would make the file wrong in the other direction, claiming a server version we aren't running. It now carries an inline note naming where it goes.

`scripts/__tests__/eoas-version-parity.test.ts` guards the asymmetry rather than just tolerating it, with three lists: files whose `eoas@` spec must equal the pin (now including both files); files that tell you what to *deploy* (always the pinned target); and files describing what's *currently running*, where the rule is self-expiring — while the version trails, the pending note must be present; the moment it reaches the target, the note must be gone. Verified by breaking it both ways: bumping the version without dropping the note fails, and dropping the note without bumping fails.

### Why bumping the CLI ahead of the server is safe

The old rule in `scripts/lib/eoas.ts` said the CLI must match the deployed server exactly. That was our convention, not a protocol requirement — there is no version handshake in either build. The real rule, now written down: **the CLI may lead the server, never trail it.** Checked before moving the pin, by unpacking both dists:

- the three routes a publish uses — `/{appId}/requestUploadUrl/{branch}`, `/uploadLocalFile`, `/markUpdateAsUploaded/{branch}` — are unchanged, and the `markUpdateAsUploaded` block is byte-identical between 3.0.5 and 3.1.2;
- `RequestUploadUrlsResponseJoi` requires only `updateId` + `uploadRequests` and is `.unknown(true)`; the CLI only hard-fails on a missing `rolloutPercentage` echo when `--rollout-percentage` was passed, which we never do;
- `rollback`'s flags are unchanged, so `scripts/mobile-ota-rollback.ts` argv still parses.

The 9m29s run above is the empirical half of that argument.

## 🚧 Infra hand-off (Railway) — the other half

**Without this step the PR delivers only the client-side half.** The third and largest upstream fix is server-side and no PR can perform it:

1. Bump the `boardsesh-ota-v3` Railway service image to **`ghcr.io/mercuretechnologies/xprem:v3.1.2`**. It is a tag change only — upstream's v3.1.0 notes state "no environment variable, API route or bucket layout changes; existing deployments upgrade by bumping the tag". (The project renamed expo-open-ota → xprem at v3.1.0; the old image name is still published.)
2. Re-verify: `GET /hc` = 200, `GET /ready` = 200, and a header-carrying `GET /manifest` (`expo-app-id`, `expo-channel-name: production`, platform + runtime headers) returning 200 with signature `keyid main`. Then `bunx eoas@3.1.2 doctor --channel=production`.
3. Update the `expo-open-ota:v3.0.5` reference in `CLAUDE.md` / `AGENTS.md` and drop the pending note — the parity test will fail until both move together, which is the point.

What it unlocks: `requestUploadUrl` starts loading the previous update's `metadata.json` for the same app/branch/runtimeVersion/platform and server-side-copies every file already there, returning upload URLs only for the remainder — **~380 uploads down to a handful** on a repeat publish to a branch. It degrades safely (an unavailable copy falls back to a normal upload). Also gated on it: `vp run mobile:ota-rollback -- --mode republish`, whose route shapes moved between 3.1.1 and 3.1.2 with back-compat living server-side. `--mode embedded`, the mode the incident runbook uses, is unaffected either way.

The Enterprise-licence/telemetry additions in 3.1.x are dashboard-side and inert on the community edition. There is no version endpoint on the server, so the deployed tag is only confirmable in the Railway dashboard.

## Two things for you to decide (follow-ups, not this PR)

- **#4611 — per-PR auto-publish volume.** Previews auto-publish on every `synchronize`. During this batch that was ~30 publishes/day with up to 11 concurrent, for previews almost nobody opens. Restricting auto-publish to `opened` / `ready_for_review` plus the existing `/ota-preview` comment would remove most of the burst source — but it changes the tester workflow, so it's your call. (A single repo-wide concurrency group would be the *wrong* fix: GitHub keeps at most one pending run per group, so intermediate PRs' previews would be silently superseded.)
- **#4612 — the 356 board-background assets.** They set the floor for every first-publish-per-branch burst and for the OTA download size.

Also not touched, deliberately: Tigris rate limits and bucket sharding. Server-side reuse removes ~356 of 380 PUTs on a repeat publish; re-measure before spending anything on the storage side.

## Release Notes

## Test plan

- **Live end-to-end** — the 9m29s preview publish at the top, against the still-3.0.5 server. No oclif unknown-flag error, zero `SlowDown`, `ios=success (1 attempt)` / `android=success (1 attempt)`, `map` job green afterwards so channel↔branch mapping is unaffected.
- `vp test run --project scripts --reporter=agent eoas mobile-publish mobile-ota` → **8 files, 106 tests passed** (baseline on a clean tree was 6 files / 93 tests).
- Drift guard verified by breaking it three ways: a stale `eoas@` version in the doc alone; the server image bumped in `CLAUDE.md` without dropping the pending note; the note dropped without bumping the image. All three fail, naming the offending file.
- `vp run typecheck` exits 0 locally; pre-commit `vp check --fix` clean on all changed files.
- If a future publish instead fails on an unknown flag or a 4xx/404 from a publish route, the CLI-leads-server assumption has broken: hold, bump the Railway image first, re-run. Don't paper over it by reverting only the flag.
- After merge: watch the first `mobile-ota-production` run on `main` end to end — production adds `--dumpSourcemap --outputDir dist` and the Sentry map upload reads `dist/metadata.json`, so both platforms must publish **and** both map uploads must succeed.
- Once the Railway image is on v3.1.2: publish twice to the same `pr-<n>` branch and confirm the second run's upload step is near-instant. That's the observable proving the server half landed.


marcodejongh:
## Live verification — this PR's own preview publish

`scripts/mobile-publish.ts` and `scripts/lib/eoas.ts` are in `mobile-ota-preview.yml`'s path filter, so this PR published `pr-4613` with the new CLI against the **still-3.0.5 server**. That is the wire-compatibility proof the unit tests can't give.

[Run 32312663175](https://github.com/boardsesh/boardsesh/actions/runs/32312663175) — `✓ publish in 9m29s` (the run this PR is about took **45m39s**).

```
[mobile:publish] Running ios: bunx eoas@3.1.2 publish --branch pr-4613 --platform ios … --upload-rate 5 …
› Assets (380):
📤 Uploading files...            23:25:42
✅ Files uploaded successfully   23:26:58
✅ Update ready for ios
[mobile:publish] Platform results: ios=success (1 attempt)

[mobile:publish] Running android: … --upload-rate 5 …
› Assets (417):
📤 Uploading files...            23:29:38
✅ Files uploaded successfully   23:31:02
✅ Update ready for android
[mobile:publish] Platform results: android=success (1 attempt)
```

- No oclif unknown-flag error — 3.1.2 accepts `--upload-rate` and the 3.0.5 server accepts everything 3.1.2 sends.
- **Zero `SlowDown` responses.** Both platforms on the first attempt; the retry ladder never engaged.
- The limiter behaves exactly as modelled: **380 assets in 76.2s** and **417 in 83.5s** — 380/5 = 76.0 and 417/5 = 83.4. The pacing is the whole cost, and it is ~80 seconds per platform.
- `map` job green afterwards, so the channel↔branch mapping is unaffected.

Also green: `typecheck`, `lint`, `commit-lint`, `test-default` (both shards), `large-files`. Local `vp run typecheck` exits 0.

---

## Re: the Claude review

**1. `CLAUDE.md` stale versions — correct, and left deliberately.** `CLAUDE.md:321` and `AGENTS.md:384` do still say `expo-open-ota:v3.0.5` / `eoas@3.0.5`, and you're right that it will mislead future sessions. I'm not editing those two files from an automated run — they're the repo's operating instructions, and a change to them should be the owner's, not a side effect of a CI fix. Flagged in the PR body for exactly that reason. Happy to do it in a one-line follow-up on @marcodejongh's say-so, at which point adding them to `VERSIONED_FILES` becomes the right move too. Adding them to the list *now*, without editing them, would just red the build.

**2. `SERVER_IMAGE_NAME` hardcoded — intentional, and there is no canonical source to derive it from.** The image name appears nowhere in the repo except prose in the doc and the setup runbook, which is what the test is checking. Deriving the expected value from one of the files under test would make the assertion vacuous — it would pass no matter what the files said. The hardcoded `'xprem'` *is* the canonical assertion: it's what makes an accidental drift back to `expo-open-ota:v…` fail. A genuine second rename is a deliberate act that should require a deliberate edit here.

(Minor: the line numbers in the review — `eoas-version-parity.test.ts:225-230` and `:217` — don't exist; the file is 81 lines. Doesn't change either point.)

marcodejongh:
## Second publish, second data point

The `CLAUDE.md` / parity-test commit re-triggered the preview publish, so there are now two independent runs. [Run 32343467407](https://github.com/boardsesh/boardsesh/actions/runs/32343467407) — `✓ publish in 8m48s`, zero `SlowDown`, `ios=success (1 attempt)`, `android=success (1 attempt)`.

| | run 1 | run 2 | predicted |
|---|---|---|---|
| iOS upload (380 assets) | 76.2s | **76.2s** | 76.0s |
| Android upload (417 assets) | 83.5s | **83.4s** | 83.4s |
| job total | 9m29s | **8m48s** | — |

Reproducible to within 0.1s, and both runs are still against the **3.0.5 server** — so the CLI-leads-server claim now has two green publishes behind it rather than one.

For contrast, the run in the issue: 45m39s, iOS on attempt 6.

marcodejongh:
Both review points actioned in `fb5484eb5` — the first one checked out, and the mechanism is sharper than I'd documented.

**1. `--mode republish` against a 3.0.5 server.** Real, and worth confirming rather than taking on faith. Diffing the two dists:

- 3.0.5's `republish` listed candidates via `/api/apps/{appId}/branch/{branch}/runtimeVersions` and `.../runtimeVersion/{rv}/updates`.
- 3.1.2 moved listing into `lib/serverUpdates.js` (a file 3.0.5 doesn't have). Two of its three routes are unchanged, but it adds `.../runtimeVersion/{rv}/publish-groups` — which 3.0.5 never served — and one republish path now sends `?publishGroup=` on the call itself.

So "route shapes moved" was too vague; it's a new listing route plus a new query param, with back-compat living server-side (xprem #168). Corrected in `scripts/lib/eoas.ts`, the rollback header, and the doc.

**Warned rather than blocked**, deliberately. The helper can't detect the deployed version — there's no version endpoint on the server — so a guard would have to be unconditional, and would then need a code change to undo the moment Railway is bumped. Blocking a rollback path during an incident to prevent a confusing error message is the wrong trade. `--mode republish` now prints:

```
[ota-rollback] NOTE: republish needs the server on xprem:v3.1.2. If it is still on
[ota-rollback] v3.0.5, expect a 404 listing previous updates — that is the version
[ota-rollback] gap, not a broken rollback. Use --mode embedded instead; it is
[ota-rollback] unaffected and is the mode the incident runbook uses.
```

**2. Table nit — fair.** The Version column now links to the Versions section (`([which tag](#versions-the-cli-pin-and-the-server-image))`) rather than carrying a bare version. It can't carry the deployed tag directly: `docs/mobile-ota-updates.md` is in the parity test's deploy-instruction list, where every image reference must equal the pinned target, and the deployed image legitimately trails until the Railway hand-off. One hop instead of a number that would be wrong in one direction or the other.

Tests still 8 files / 106 passing.

marcodejongh:
All three actioned in `b040fc2a6`.

**1. Versions section ambiguity — fair, fixed inline.** It now reads "...the matching server image is `xprem:v3.1.2` — **which Railway is not on yet**: the deployed tag is still 3.0.5 until the hand-off below, and that is fine (see the CLI-leads-server rule)." No skimming required. Note the section can't carry the deployed tag in `image:vX.Y.Z` form — the parity test requires every image reference in this doc to equal the pinned target, precisely so deploy instructions can't go stale — so the bare version does the work.

**2. Untested warning — good catch, and it mattered more than "output-only" suggests.** Warn-not-block is only defensible *because* the text names the fallback; if someone trimmed that line the choice would quietly become "let it 404". So the warning is now an exported `republishServerVersionWarning()` with two tests holding it to naming both the matching image and `--mode embedded`. The image string is derived from `EOAS_PACKAGE_SPEC` rather than written out again, so it can't drift from the pin either.

**3. Test name — right, renamed** to "keeps %s's pending-bump note in step with its server version", which covers both branches.

Tests: 8 files / **108 passing** (was 106).

Minor, and consistent across all three of your reviews: the line numbers don't resolve. This round cited `scripts/mobile-ota-rollback.ts:476-488` (227 lines) and `eoas-version-parity.test.ts:321` (126 lines); the earlier one cited `:225-230` and `:217` in that same 81-line file. The findings themselves have been sound each time, so this looks like citation formatting rather than misreading — worth knowing if anyone tries to jump to those lines.

marcodejongh:
Two claims in this review are wrong, and one is right in a way worth stating plainly.

**Wrong: "the warning is printed but only after the command starts."** It isn't. In `scripts/mobile-ota-rollback.ts`:

```
206:    for (const line of republishServerVersionWarning()) console.log(line);
209:  console.log(`[ota-rollback] Running: bunx ${eoasArgs.join(' ')}`);
225:  const result = spawnSync('bunx', eoasArgs, { … });
```

The warning is printed 19 lines and one `spawnSync` boundary *before* anything runs — it's in the same pre-flight block as Mode/Server/Branch/Platform, above the `Running: bunx …` line. There is no risk window of the kind described. (The cited `:138-145` is the helper's definition, not its call site.)

**Wrong: "infra hand-off is a hard prerequisite for the primary fix."** The primary fix is the rate cap, and it is working right now against the still-3.0.5 server — three green preview publishes on this PR, ~9 minutes each, zero `SlowDown`, both platforms on the first attempt, versus 45m39s for the run in the issue. The hand-off unlocks an *additional* improvement (server-side asset reuse, ~380 uploads → a handful on a repeat publish). Reading it as a prerequisite gets the PR exactly backwards: merging without the hand-off still buys the 45m → 9m.

**Right, and I'll own it plainly:** this PR does open a temporary gap on `--mode republish`. Before it, the pin was 3.0.5 and republish worked; after it, republish stays broken until Railway moves. That is a real cost, not a documentation artefact. The trade: republish is the interactive, local-only, non-incident rollback path, while `--mode embedded` — the mode the incident runbook actually uses — is unaffected, and the alternative is forfeiting a 5x publish speedup to protect it.

There is a cleaner option I deliberately did **not** take unilaterally, because it hinges on a claim I can't verify from the client dists alone: pin *just the republish subcommand* back to 3.0.5. If xprem #168's back-compat is server-side as its PR body describes, that would work against all four client/server combinations and close the gap entirely. But it means two CLI pins in one repo and an assumption about server behaviour I can only read secondhand — @marcodejongh's call, not mine. Say the word and it's a ten-line change.

Everything else in the review — rate limiting sound, coverage thorough — matches what I'd expect.

Issue #4615, opened 2026-08-19
ci: test-location-sync-integration dies in db:migrate with PostgresError EINTR


marcodejongh:
Found while auditing every failing `ci.yml` run on `main` for #4488. This is the second of the two remaining reds that is **not** lint and **not** the unhandled rejection.

## The failure

`test-location-sync-integration` dies in its `db:migrate` step, before a single test runs:

```
@boardsesh/db db:migrate: PostgresError: could not open file "base/16384/17249": Interrupted system call
       where: "SQL statement \"CREATE TEMP TABLE _mad_fingerprints ON COMMIT DROP AS
    SELECT climb_uuid,
           string_agg(hold_id::text || ':' || hold_state, ',' ORDER BY hold_id) AS fp
      FROM board_climb_holds
     WHERE board_type = 'moonboard'
     GROUP BY climb_uuid\"
PL/pgSQL function inline_code_block line 26 at SQL statement",
        file: "md.c",
     routine: "mdopenfork",
        code: "XX000"
##[error]Process completed with exit code 1.
```

The Postgres service container logs the matching server-side line:

```
2026-08-19 19:07:26.623 UTC [170] ERROR:  could not open file "base/16384/17249": Interrupted system call
```

## Occurrences (all on `main`, all identical)

| run | date | job |
|---|---|---|
| [32291098426](https://github.com/boardsesh/boardsesh/actions/runs/32291098426) | 2026-08-19 | 96191927231 |
| [31915573610](https://github.com/boardsesh/boardsesh/actions/runs/31915573610) | 2026-08-15 | 95086769328 |
| [31908466239](https://github.com/boardsesh/boardsesh/actions/runs/31908466239) | 2026-08-15 | 95070099158 |

Same error, same routine (`mdopenfork`), and the **same relfilenode `base/16384/17249`** every time — which is expected, since the job pins the dev-db image by digest (`ghcr.io/boardsesh/boardsesh-dev-db@sha256:ce32f2a…`), so the OID is stable across runs. The read that trips is migration 0193's moonboard-angle-dedup backfill scanning `board_climb_holds`.

## What it is

[...]

Issue #4614, opened 2026-08-19
test(backend): session-persistence debounce flush test races a fixed 200ms sleep


marcodejongh:
Found while auditing every failing `ci.yml` run on `main` for #4488. This is one of the two remaining reds that is **not** lint and **not** the unhandled rejection, so it wants its own issue.

## The failure

`main` run [32200535151](https://github.com/boardsesh/boardsesh/actions/runs/32200535151), job `test-backend (1, 2)`:

```
 FAIL   backend  src/__tests__/session-persistence.test.ts > Session Persistence - Hybrid Redis + Postgres > Debounced Writes and Flush > should cancel pending timer when new update arrives
AssertionError: expected +0 to be 1 // Object.is equality

- Expected
+ Received

- 1
+ 0

 ❯ src/__tests__/session-persistence.test.ts:679:32
    677|       queueRows = await db.select().from(sessionQueues).where(eq(sessi…
    678|
    679|       expect(queueRows.length).toBe(1);

 Test Files  1 failed | 130 passed (131)
      Tests  1 failed | 1863 passed (1864)
```

One test out of 1864. It has not repeated in the runs since.

## Why it can fail

The test drives the 30s debounce with fake timers, then hands off to **wall-clock time** for the write it is asserting on:

```ts
// packages/backend/src/__tests__/session-persistence.test.ts:668-679
await vi.advanceTimersByTimeAsync(15000);

// The debounce callback fires and starts writeQueueStateToPostgres asynchronously.
// Switch to real timers and wait briefly for the Postgres I/O to complete.
vi.useRealTimers();
await new Promise((resolve) => setTimeout(resolve, 200));

queueRows = await db.select().from(sessionQueues).where(eq(sessionQueues.sessionId, sessionId));

expect(queueRows.length).toBe(1);
```

The comment says it out loud: the debounce callback *starts* the Postgres write and does not return a handle, so the test sleeps 200ms and hopes. On a CI runner sharing two cores with a 128-file shard and a containerised Postgres, an INSERT that normally lands in single-digit milliseconds can miss a 200ms budget. Nothing about the assertion is wrong — 1 row is the correct expectation — it just reads the table before the row exists.

The earlier `expect(queue.length).toBeLessThan(2)` in the same test is guarded by `if (queueRows.length > 0)`, so the *negative* half already tolerates the race; only the positive half does not.

## Suggested fix

[...]

Issue #4548, opened 2026-08-19
Lint rules in .oxlintrc.json are enforced nowhere — the documented localStorage and raw-sql bans don't fire in vp check or CI


marcodejongh:
Found while adding a lint guard for #4529. The guard was added to `.oxlintrc.json`, proved not to fire, and only worked once it was duplicated into `vite.config.ts`. That prompted an audit of what else is in the same position.

## The mechanism

`vp check` — the canonical validation command, the pre-commit hook, and what CI runs — does **not** read `.oxlintrc.json`. It derives its ruleset from `vite.config.ts`'s `lint` / `lint.overrides` block. `vite.config.ts` says so itself in two places:

```
// the .oxlintrc.json rule isn't enforced by `vp check` (reduced ruleset), so this is the real backstop.
```

And the `oxlint` binary in `node_modules/.bin` refuses to lint at all:

```
$ oxlint -c .oxlintrc.json <file>
This oxlint wrapper is for IDE extension use only (--lsp mode).
```

So a rule that lives only in `.oxlintrc.json` is enforced in a developer's editor and **nowhere else** — not locally, not in the hook, not in CI.

## What that currently costs us

Two rules CLAUDE.md documents as hard guardrails are in exactly that position. `vite.config.ts` defines **no** `no-restricted-globals` and **no** `no-restricted-imports` block at all.

**1. The `localStorage` / `sessionStorage` ban.** CLAUDE.md: *"`localStorage`/`sessionStorage` are lint-blocked (`no-restricted-globals`). Use IndexedDB."* Probe file in `packages/web/app`:

```ts
export function probe(): string | null {
  return localStorage.getItem('boardsesh:probe');
}
```
→ `vp check`: *"Found no warnings, lint errors, or type errors in 1 file"*.

**2. The raw-`sql` import ban.** CLAUDE.md: *"Importing `sql` from `@/app/lib/db/db` (raw Neon HTTP client) is lint-blocked."* Probe:

```ts
import { sql } from '@/app/lib/db/db';
```
→ `vp check`: clean.

`@boardsesh/offline-sync` has a restricted-import entry in the same unenforced position.

Both probes were deleted immediately; nothing was committed.

The two `no-restricted-properties` entries that *are* enforced (`Intl.RelativeTimeFormat`, `Intl.ListFormat`) are duplicated by hand into both files — which is the existing convention, but nothing checks that the duplication stayed in sync.

## Why it matters

[...]

PR #4616, opened 2026-08-19
ci(repo): lint the whole repo on every PR

### LINKED ISSUE #4488: main CI has not been green in 30 runs: full-repo lint debt plus an unhandled rejection in a passing backend suite

marcodejongh:
## Summary

Closes #4488.

**First, the issue's headline is stale.** #4488 says `main` is 0-green in 30 runs from 33 lint errors. Those 33 errors were cleared by #4521 (`1ee16af3`, 2026-08-16). Since then `main` is **11 green of 13 completed `ci.yml` runs (~85%)**, and neither red is lint or the unhandled rejection. The 324 warnings the issue also mentions never failed CI and never could — `vp check` exits 0 on warnings; the exit-1 came only from the errors. This PR is not rescuing a broken `main`.

What is still open is the *reason* those 33 errors could accumulate invisibly, plus the one nondeterministic red that can still fire.

### 1. PR CI now lints the whole repo, same as `main`

**⚠️ @marcodejongh — this changes every contributor's PR experience, so it wants your explicit nod before merge.**

Today the `lint` job runs two different things: `Lint changed files (PR)` on a PR, `Lint full repo (main)` on a push. An error in a file no PR happened to touch therefore could not be caught until it was already on `main`. That is exactly how 33 of them accumulated while every PR that introduced them stayed green.

This deletes the diff-lint step and runs the same full-repo `vp check` on both sides.

- **Measured cost: ~26s per PR.** Step timings for the green `main` lint job 96193441921: `Lint full repo` 19:12:28 → 19:12:54. The whole lint job is 109s, 51s of which is `voidzero-dev/setup-vp`.
- **`main` is green by construction**, so the "an unrelated pre-existing error blocks your PR" risk is largely theoretical — but it is not zero. If someone lands an error on `main` through a path that skips the lint job's `if:` gate, the next PR to touch any JS file eats it and has to rebase.
- **Fallback if you'd rather not:** keep diff-lint on PRs and add a scheduled job that runs the full-repo pass, burns the delta down, and files what it cannot fix. Say the word and I'll swap to that; the rest of this PR is independent.

`scripts/__tests__/ci-lint-scope.test.ts` pins the job to exactly one `vp check` with no event-conditioned step and no diff-derived file list, so the two scopes cannot silently drift apart again. Verified discriminating: run against the pre-change `ci.yml`, all four assertions fail (2 `vp check` invocations, `Lint changed files` present, `github.event_name` present, `git diff` present).

### 2. Lint scope aligned with `.oxlintrc.json`

`vp check` reads the `lint` block in root `vite.config.ts`, **not** `.oxlintrc.json` — that is #4548, and it is a real trap: editing `.oxlintrc.json` to change severities or scope changes nothing in CI. The two had drifted, so `main`'s full-repo pass was type-aware-linting `embedded/` firmware scripts and `design/` that `.oxlintrc.json` excludes and that no PR ever linted — the same asymmetry as above, one level down.

`lint.ignorePatterns` now carries the same list, plus `**/generated/**` and `packages/db/drizzle/**` for lock-step hygiene (those two produce 0 diagnostics either way). A test asserts `vp` never ignores a path oxlint would lint, so the hand-duplication stops being a trust exercise.

### 3. The unhandled rejection

Appeared once, in `main` run 31880202330, `test-backend (2, 2)`: 1712 tests passed, `Errors 1`, exit 1. Not seen in the 13 runs since.

`board-merge-tombstone.test.ts` starts `followPromise` and attaches its first handler three `await`s later. If the follow's rejection (`Board not found`) is delivered in an earlier macrotask than the merge's COMMIT, Node reaches its unhandled-rejection checkpoint with no handler attached, vitest counts it under `Errors`, and the job exits 1 with every test green.

`handleLater()` attaches an inert `.catch()` at creation, at 24 sites across the two concurrency test files.

**Proven:**

- **The mechanism, deterministically, in-suite.** `packages/backend/src/__tests__/deferred-rejection.test.ts` reproduces the leak in the abstracted shape — a promise that rejects while the test is suspended on another — sequenced by macrotask rather than wall clock, so it does not race a timer budget under load. Drop the `handleLater()` call from it and it fails:

  ```
  AssertionError: expected [ 'Board not found' ] to deeply equal []
  ```

  With the call in place the leak list is empty and the rejection is still there to assert on.

- **That the fix cannot hide a defect.** Discriminating mutation probe: with `handleLater(followPromise)` in place, change the expectation at `board-merge-tombstone.test.ts:517` to a message the code never throws, and the test **fails** —

  ```
  AssertionError: expected [Function] to throw error including 'a message it never throws' but got 'Board not found'
  ```

  The inert catch did not consume the rejection. The assertion still receives it and still discriminates. Reverted after the probe.

- **The contract**, pinned in the same file (no DB, no services): the promise stays rejected, a wrong expectation still fails, and a resolved value passes through untouched.

**Inferred, not proven:** that this specific race is what fired in run 31880202330. There is no reproduction. The race needs a 2-core CI Postgres under a 128-file shard; the previous attempt recorded on #4488 established that neither the single file nor the full 256-file `--project backend` run reproduces it locally, with or without the fix. So the mutation probe is what carries the weight here: the change is provably not an inert guard, and it is provably not load-bearing on any assertion. Worst case it is a no-op.

### 4. Production floating promises

The five non-BLE ones now take the `void` operator the rule asks for (`use-offline-mutations.ts` ×4, `i18n/client.tsx` ×1). All were already deliberate fire-and-forget; behaviour is unchanged.

Deliberately **not** in this PR:

- `packages/mobile/src/lib/ble/use-board-scan.ts:48` — BLE, needs a Fable review per CLAUDE.md, and dragging one onto a CI-hygiene PR helps nobody.
- The ~166 test-file occurrences. They are `node:test`'s `test()` return value and vitest `act()` noise, they are warnings, they never failed CI, and a 166-file diff would bury everything above.

### Not in scope — filed separately

The other two `main` reds in the window are different bugs:

- #4614 — `session-persistence.test.ts:679` races a fixed 200ms real-timer sleep waiting on async Postgres I/O.
- #4615 — `test-location-sync-integration` dies in `db:migrate` with `PostgresError: could not open file "base/16384/17249": Interrupted system call`. Three occurrences, identical relfilenode; runner I/O, wants a retry not a code fix.

## Release Notes

## Test plan

- **Full-repo `vp check` on this branch**: `Found 0 errors and 254 warnings in 4396 files`, exit 0. Down from `main`'s `0 errors and 324 warnings in 4414 files` — 65 fewer from the `embedded/` + `design/` scope alignment, 5 from the `void` operators. Zero diagnostics remain under `embedded/` or `design/`. This is the local evidence that turning full-repo lint on for PRs passes on this branch.
- `vp test run --project scripts scripts/__tests__/ci-lint-scope.test.ts` — 6 passed. Same assertions against the pre-change `ci.yml`: all four fail.
- `vp test run --project backend packages/backend/src/__tests__/deferred-rejection.test.ts` — 3 passed; the leak test fails when `handleLater()` is removed from it.
- `vp test run --project backend packages/backend/src/__tests__/board-merge-tombstone.test.ts packages/backend/src/__tests__/board-presence.test.ts` — 130 passed, no `Errors` line.
- `vp test run --project backend packages/backend/src/__tests__/board-merge-tombstone.test.ts -t "returns Board not found when the merge commits first"` — passes; fails as quoted above under the mutation probe.
- Backend tests here share fixed ports and worker DBs on `:5433` with ~10 sibling worktrees, so the whole-suite run is CI's to arbitrate.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -213,11 +213,6 @@ jobs:
       - uses: actions/checkout@v6
         with:
           fetch-depth: 50
-      - name: Fetch base ref
-        if: github.event_name == 'pull_request'
-        env:
-          BASE_REF: ${{ github.base_ref }}
-        run: git fetch --depth=50 origin "$BASE_REF":"refs/remotes/origin/$BASE_REF"
       - uses: oven-sh/setup-bun@v2
       - uses: voidzero-dev/setup-vp@v1
         with:
@@ -226,7 +221,7 @@ jobs:
       - name: Install dependencies
         run: bun install --frozen-lockfile
       - name: Install Expo web lint dependencies
-        # Unconditional: the push-to-main step below lints the WHOLE repo, so
+        # Unconditional: the lint step below lints the WHOLE repo, so
         # packages/mobile/src/web-shims/bottom-sheet.tsx is always in scope and
         # its `@gorhom/bottom-sheet` types only resolve through the isolated
         # web-runtime install (packages/mobile/tsconfig.json maps the path
@@ -294,34 +289,19 @@ jobs:
       # `vp check`, so this grep guard is the real backstop.
       - name: Check offline-sync adapter boundary
         run: vp run check:mobile-offline-sync-imports
-      - name: Lint changed files (PR)
-        if: github.event_name == 'pull_request'
-        env:
-          BASE_REF: ${{ github.base_ref }}
-        run: |
-          # Match the staged-hook scope in root vite.config.ts: TS/JS files,
-          # excluding generated/, board-controller, and embedded firmware sources.
-          # xargs -r ensures no-op on empty input.
-          git diff --name-only --diff-filter=ACMR \
-            "origin/$BASE_REF...HEAD" \
-            -- '*.ts' '*.tsx' '*.js' '*.mjs' '*.cjs' \
-            | grep -vE '(/node_modules/|/generated/|^packages/board-controller/|^embedded/)' \
-            > changed-lintable.txt || true
-          if [ -s changed-lintable.txt ]; then
-            # Same stdout redirect as the full-repo step below, and for the same
-            # reason: vp panics on EAGAIN when it flushes a long warning list to the
-            # runner's non-blocking stdout, and xargs reports that crash as 123 with
-            # the diagnostics cut off mid-line. A PR touching one warning-heavy file
-            # is enough to hit it. Writing to a regular file never EAGAINs; vp's real
-            # exit code is preserved so lint errors still fail the job.
-            xargs -r vp check < changed-lintable.txt > vp-check.log 2>&1 && status=0 || status=$?
-            cat vp-check.log || true
-            exit $status
-          else
-            echo "No lintable files changed."
-          fi
-      - name: Lint full repo (main)
-        if: github.event_name != 'pull_request'
+      # PRs and main lint the SAME thing: the whole repo. This job used to lint
+      # only a PR's changed files and reserve the full-repo pass for push-to-main,
+      # which meant an error in a file no PR happened to touch could not be caught
+      # until it was already on main. 33 of them accumulated that way and broke
+      # every main run for days (issue #4488, fixed by #4521) while every PR that
+      # introduced them stayed green. Measured cost of closing that gap: ~26s per
+      # PR (step timings, job 96193441921), against a 109s job dominated by
+      # setup-vp. Scope is now identical on both sides by construction, so the
+      # only way to land a lint error is to be the PR that shows it.
+      #
+      # Effective config is the `lint` block in root vite.config.ts, NOT
+      # .oxlintrc.json — see issue #4548 and the comment on that block.
+      - name: Lint full repo
         # vp/Vite+ panics with "failed printing to stdout: Resource temporarily
         # unavailable (os error 11)" when it flushes the full-repo warning list to the
         # runner's non-blocking stdout. Write to a regular file (never EAGAIN), then
--- a/packages/backend/src/__tests__/board-merge-tombstone.test.ts
+++ b/packages/backend/src/__tests__/board-merge-tombstone.test.ts
@@ -9,6 +9,7 @@ import { boardPresenceMutations } from '../graphql/resolvers/board-presence/muta
 import { findChosenBoardForSerial } from '../graphql/resolvers/board-presence/shared';
 import { lockBoardSerialWrite } from '../graphql/resolvers/board-serial-write-lock';
 import { seedAuroraCatalogFixtures } from './helpers/board-catalog-fixture';
+import { createBarrier, createValueBarrier, handleLater } from './helpers/concurrency';
 
 /**
  * Real-DB coverage for issue #3407's merge-tombstone following and the
@@ -155,40 +156,6 @@ async function serialPointerUuid(userId: string, serial: string): Promise<string
     `),
   );
   return rows[0]?.board_uuid ?? null;
-}
-
-function createBarrier(): { promise: Promise<void>; release: () => void } {
-  let release = (): void => undefined;
-  const promise = new Promise<void>((resolve) => {
-    release = resolve;
-  });
-  return { promise, release };
-}
-
-function createValueBarrier<Result>(): { promise: Promise<Result>; release: (result: Result) => void } {
-  let release = (_result: Result): void => undefined;
-  const promise = new Promise<Result>((resolve) => {
-    release = resolve;
-  });
-  return { promise, release };
-}
-
-/**
- * Attach an inert rejection handler to a promise this test asserts on later.
- *
- * These tests deliberately leave a promise in flight across several `await`s.
- * If it rejects during one of them, Node reaches its unhandled-rejection
- * checkpoint before the assertion further down can attach a handler — vitest
- * counts that under `Errors` and exits 1 with every test still passing
- * (issue #4488, main run 31880202330).
- *
- * The promise itself is untouched: still rejected, still assertable. The
- * `await expect(...).rejects.toThrow(...)` or bare `await` below still runs,
- * still sees the same rejection, and still fails on the wrong error — so this
- * cannot hide a defect. Its contract is pinned in deferred-rejection.test.ts.
- */
-function handleLater(promise: Promise<unknown>): void {
-  void promise.catch((): void => undefined);
 }
 
 async function waitForAdvisoryWaitBlockedBy(blockingPid: number): Promise<number> {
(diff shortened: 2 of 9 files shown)


marcodejongh:
Thanks — two of the three are now fixed, third left deliberately.

**`handleLater` duplicated across two test files.** Fixed, and taken further than suggested: `createBarrier` and `createValueBarrier` were duplicated across the same two files for the same reason, so all three moved to `packages/backend/src/__tests__/helpers/concurrency.ts` (`helpers/` already exists — `board-catalog-fixture.ts`, `headless-queue-client.ts`, `mock-redis.ts`). Net −68 lines. The useful side effect is that `deferred-rejection.test.ts` now imports the shipped `handleLater` instead of re-declaring a copy, so the contract test pins the real implementation rather than a lookalike.

**Timing sensitivity in `deferred-rejection.test.ts`.** Fair, and it would have been a poor look on a de-flaking PR. The 10/30/20ms `setTimeout` budget is gone: the rejection is now driven from a `setImmediate` and the test hops two macrotasks, which is what Node's unhandled-rejection checkpoint actually keys off. No wall-clock budget left to blow.

It also discriminates now, which the timer version could not be trusted to do. Remove the `handleLater()` call from that test and it fails:

```
AssertionError: expected [ 'Board not found' ] to deeply equal []
```

So the leak is reproduced deterministically in-suite, not only in the throwaway script quoted in the PR description. I've updated the description to say so.

**`ci-lint-scope.test.ts` reads files at module scope.** Leaving as is. You spotted the reason yourself: `ci-location-sync-workflow.test.ts` and `ci-test-default-projects.test.ts` both do it, and the `scripts` project always runs from the repo root (`vp test run --project scripts`, root `vite.config.ts` projects list). Making this one file lazy while the neighbours stay eager trades a hypothetical confusing error for a real inconsistency. If it's worth changing, it's worth changing all three in a pass of their own.

marcodejongh:
Thanks — taking all three as no-change, with reasons rather than shrugs.

**1. `stringArrayLiteral` only matches single-quoted strings.** The double-quote half of this can't happen: root `vite.config.ts` sets `fmt.singleQuote: true`, its `fmt.ignore` list doesn't cover itself, and `vp check` verifies formatting — now on every PR, as of this change. A double-quoted string in that file fails CI before it can reach the test.

The multiline half is already handled and was exercised: `indexOf(']', openIndex)` scans past newlines, and the `matchAll` runs over the whole slice. I wrote the array multiline first and the suite passed 6/6; prettier then collapsed it to one line. Both shapes work.

**2. `mappingEntry` at describe level.** Agreed it degrades the error message, and it's the pattern in both neighbouring contract tests. Same answer as last round: worth fixing across all three in a pass of their own, not in one file here.

**3. One-sided sync invariant.** Deliberate, and the asymmetry is not arbitrary — the two directions have different consequences. A pattern in `vite.config.ts` but not `.oxlintrc.json` means `vp check` skips a file oxlint would flag: real lost coverage, and the thing that let 33 errors sit in `main`. The reverse means `vp check` lints a file the editor ignores: noisier, never less safe. That was in fact the pre-existing state for `embedded/`, which is what this PR just corrected.

Pinning both directions would be tighter, but it belongs with #4548, which proposes generating one file from the other so the duplication stops existing at all. A two-sided assertion here would have to be undone by that work.

Issue #4618, opened 2026-08-19
Fixed sitemap shards have no byte budget — /sitemaps/playlists.xml crosses Vercel's 4.5 MB ceiling well before the item cap


marcodejongh:
Found while fixing #4524 (caching the playlists sitemap shard).

## Fixed shards never check `MAX_SHARD_BYTES`

`shardRouteHandler` in `packages/web/app/lib/seo/sitemap/shard-registry.ts` guards the fixed path with `MAX_URLS_PER_SHARD` (45,000) and nothing else. The byte guard — `MAX_SHARD_BYTES` (4,000,000), enforced on the rendered body — exists **only** on the paged path (`pagedShardRouteHandler`). So the five fixed shards (`static`, `boards`, `gyms`, `setters`, `playlists`) can render a body of any size and the handler will happily return it.

That guard was added deliberately, and its own doc comment says why:

> Hard stop, enforced on the RENDERED BODY rather than on a row count — a constant alone is a comment. Vercel caps a serverless response at 4.5 MB and a truncated 200 is strictly worse than a 503 the crawler retries.

The fixed path just never got it.

## `/sitemaps/playlists.xml` is the one that will hit it

Measured in production on 2026-08-19:

- body: **2,326,713 bytes**
- items: **2,688** public playlists holding at least one climb
- URLs: 2,688 × 4 locales = 10,752
- cost: **~866 bytes per item** (four `<url>` blocks with `<lastmod>`, `<changefreq>`, `<priority>` and three `xhtml:link` alternates each)

At 866 B/item:

| items  | rendered bytes | what happens                                        |
| ------ | -------------- | --------------------------------------------------- |
| 2,688  | 2.33 MB        | today — fine                                        |
| ~4,620 | 4.00 MB        | would trip `MAX_SHARD_BYTES`, if it were checked    |
| ~5,200 | 4.50 MB        | **Vercel's response ceiling — platform error**      |
| 11,250 | ~9.7 MB        | `MAX_ITEMS_PER_SHARD`, the only cap actually applied |

So the cap the code enforces is more than twice the size the platform can serve. Public playlists roughly doubling breaks the shard, and the failure mode is a platform truncation or 500 rather than the 503-and-retry the design intends. `expectsUrls` is `false` for `playlists`, so nothing else notices.

`boards` is the other data-backed fixed shard and is far from the ceiling today, but it has the same missing guard.

## Suggested shape

[...]

Issue #4583, opened 2026-08-19
Sitemap: paged shard summaries have no headroom against SHARD_DEADLINE_MS on a cold cache


marcodejongh:
## What

`buildPagedIndexEntries` races every paged shard's `summary()` against `SHARD_DEADLINE_MS = 3_000`. Both paged shards' summaries are full catalogue scans, and neither fits that budget cold.

Measured on the local dev image (`ghcr.io/boardsesh/boardsesh-dev-db`, 648,657 `board_climbs`) — production reconciliation outstanding:

| shard | summary cost |
|---|---|
| `setters` (#4465) | 1,354–1,439 ms with the OS page cache warm; a review lens measured 2,999 ms genuinely cold |
| `climbs` | ~9.2 s cold / ~0.94 s warm for the single largest group, times ~9 groups (recorded in `climb-query.ts`) |

`getAllBoardConfigsOrThrow()` runs first inside the same deadline, and is recorded at ~10 s cold with a deterministic shard loss after a deploy (#4519).

## Why it is not simply a bug

`withDeadline` does not cancel the abandoned work, so a miss still fills both caches and the next request past the 60 s `DEGRADED_CACHE_CONTROL` window gets the shard. The degrade is self-healing per instance. But on Vercel a cold crawl fans across instances, so the shard is dropped once per cold lambda, and `/sitemap.xml` is exactly the endpoint a crawler hits cold.

## Two things ruled out already, with measurements

- **Turning the setters summary into `COUNT(*)`** buys nothing: the grouped scan is the entire cost (1,409/1,354/1,439 ms for the count variant vs 1,434/1,366/1,370 ms for the full row build), and it would drop the structural guarantee that the item list and the summary describe the same set.
- **Raising `SHARD_DEADLINE_MS`** was deliberately not done in #4465. It is shared infrastructure, and the #4493 ruling says cache the summary rather than raise the deadline.

## Options worth costing

1. A materialised summary table (count + max clock per shard), refreshed by the same job that refreshes `board_setter_stats`. Already recorded as the tier-3 follow-up for the climbs shard.
2. A separate, longer deadline for paged summaries only, on the argument that the cost of guessing low is now systematically dropping the two biggest shards rather than one slow one. Needs a call on `/sitemap.xml` latency.
[...]

Issue #4523, opened 2026-08-16
Sitemap index intermittently drops the climbs shard — ~52,000 climb URLs vanish when the tier-2 summary misses its 3s deadline


marcodejongh:
The climbs shard is the largest surface in the sitemap by two orders of magnitude — six pages of roughly 10,000 climb URLs each, ~52,000 in total. It is also the shard least able to meet the index's 3-second per-shard deadline, so it intermittently leaves `/sitemap.xml` entirely. When it does, every climb page we have disappears from the index at once.

This is the shard #4522 deliberately did **not** fix. That PR cached the boards fetch and made the smoke report degradation instead of hard-failing on it; both help the boards shard and neither closes this gap.

## What was measured

Against `www.boardsesh.com`:

| | result |
|---|---|
| `/sitemaps/climbs/1.xml` cold | **51s** |
| `/sitemaps/climbs/1.xml` warm (CDN HIT) | 200, 2.2 MB, 0.08s |
| `/sitemap.xml` during a cold window | `x-sitemap-degraded: playlists,climbs` — no climbs pages listed at all |
| `/sitemap.xml` sampled later | lists `climbs/1.xml` … `climbs/6.xml`, `x-sitemap-degraded: playlists` |

So it is intermittent rather than permanently absent — which is exactly what makes it easy to miss. Sample it warm and the index looks complete. The `playlists` shard, by contrast, was degraded on 100% of samples taken; that is a separate, more persistent hole (see "Related" below).

`SHARD_DEADLINE_MS` is 3,000 ms. A cold climbs summary cannot come close.

## Why the summary is slow

`fetchTier2Summary` in `packages/web/app/lib/seo/sitemap/climb-query.ts` returns two numbers — an item count and a max timestamp — but the *cost* is not two numbers. It runs the same `DISTINCT ON (climb_uuid)` scan as the full item build, once per `(board_type, layout_id)` group, and there are ~18 groups. The loop is deliberately sequential, not `Promise.all`: fanning ~18 heavy scans out concurrently against a ten-connection pool is the starvation described in #4461. So the honest shape is "small answer, expensive question", and the file already says as much.

Measured on the full-board dev image, the single largest group alone was **9.2s cold / 0.94s warm**. Multiply across the groups and a cold miss is tens of seconds — consistent with the 51s observed in production.

It is already cached at two levels and both are working as intended:

[...]

PR #4619, opened 2026-08-19
fix(web): cache the playlists sitemap shard

### LINKED ISSUE #4524: Sitemap index drops the playlists shard on every sample — it has no cache at all
Found while fixing the boards-shard cold start (#4522) and filing the climbs equivalent (#4523).

## The playlists shard is degraded on every sample

`/sitemap.xml` gives each shard `SHARD_DEADLINE_MS` (3s). Across every production sample taken while investigating #4522 — including three consecutive fetches where `climbs` had recovered — the response header read:

```
x-sitemap-degraded: playlists
```

So unlike climbs, which is intermittent and cold-miss driven, **playlists appears to miss the deadline consistently**, including on warm requests. Every public playlist is therefore absent from the sitemap index in normal operation.

Sampling caveat, stated because it's exactly what makes this easy to dismiss: these are a handful of observations over one investigation window, not a monitored series. The direction was unambiguous — playlists was degraded in 100% of samples where it was observed — but confirm before sizing the work.

## Why it's a different fix from the other two

Worth spelling out, because all three shards look like one problem and aren't:

- **boards** — was uncached; #4522 caches it. Solved.
- **climbs** (#4523) — already cached at two levels; its miss is intrinsically expensive (~18 sequential `DISTINCT ON` groups). Needs a materialised summary.
- **playlists** — `fetchPlaylistSitemapRows` has **no cache at all** and hits Postgres on every `force-dynamic` index request.

So playlists is the same *class* as boards but not the same *fix*, for two reasons found while reviewing #4522:

1. **Dates don't survive the Data Cache.** `unstable_cache` serialises, so a cached row set comes back with string timestamps. The climbs summary already deals with this via an explicit rehydration step — playlists would need the same, so it isn't a one-line wrapper.
2. **`withDeadline` doesn't cancel.** The index stops *waiting* at 3s; the abandoned query keeps running against the connection pool. Caching reduces how often that happens but doesn't bound the load when it does. Real cancellation needs an `AbortSignal` threaded into the query, or a statement timeout on it.

## Suggested shape

Mirror what #4522 did for boards — `unstable_cache` plus an in-process TTL with single-flight — with the rehydration step the date columns require. Then decide separately whether the abandoned-query load is worth bounding with a statement timeout.

**Do not raise `SHARD_DEADLINE_MS`.** That constant is deliberate and documented: the cost of guessing low is one shard omitted under a 60-second degraded cache window, while the cost of guessing high is holding the whole index — and every other shard with it — behind one slow builder.

## Related

- #4522 — boards cold start, caching + the post-deploy smoke change that made this visible
- #4523 — climbs, same symptom, different cause
- The structural note from #4522: on Vercel the index can't reliably warm its own caches, since work abandoned at 3s must outlive the response to write the Data Cache. A `waitUntil` around the abandoned builder would make degradation self-healing on the first miss for all three shards rather than probabilistically.

marcodejongh:
## Summary

`fetchPlaylistSitemapRows` had no cache of any kind. `/sitemap.xml` is `force-dynamic`, so every CDN miss ran the live query — `SELECT uuid, updated_at FROM playlists WHERE is_public AND EXISTS(playlist_climbs …)` — inside the index's 3 s `SHARD_DEADLINE_MS`, against the same ten-connection pool the climbs builder is using on the same request. When it loses that race the index publishes without the `playlists` entry for 60 s, and 10,752 locale-expanded URLs vanish from the sitemap.

**The issue's headline is overstated and this PR corrects it.** #4524 says `playlists` was degraded in 100% of samples. Re-measured on 2026-08-19 across 12 distinct origin computations of `https://www.boardsesh.com/sitemap.xml` (cache-busted per request — the Data Cache key does not include the URL, so busting the CDN alone would not have re-run anything): `x-sitemap-degraded` named `playlists` in **4 of 12**, about a third, not all of them. Index responses sat at 3.05–5.37 s, pinned at the deadline by climbs. The bug is real — a third of origin computations dropping the whole playlists surface is worth fixing — but the issue's sample was thin and it says so itself.

Supporting cost measurement: `GET /sitemaps/playlists.xml`, cache-busted ×5, took 4.25 / 1.37 / 2.29 / 1.33 / 1.51 s and returns 2,326,713 bytes containing 2,688 `<url>` entries × 4 locales. Net of render and transfer the query alone is ~0.8–1 s warm and up to ~3.7 s cold — i.e. it straddles the 3 s deadline, which is exactly the intermittency above.

### The fix

Two cache layers in `playlist-query.ts`, the same shape `getAllBoardConfigsOrThrow` got for boards in #4522:

- **Next Data Cache** (`unstable_cache`, 1 hour, tag `sitemap-playlists`) — cross-instance, so a cold lambda does not re-run the query.
- **In-process TTL + single-flight** (1 hour) — `unstable_cache` does not deduplicate concurrent misses, and a crawl burst is the index plus `/sitemaps/playlists.xml` arriving together. Nothing is stored on rejection, so a failure is never memoised.

Caching the **rows** rather than a summary is what makes this different from #4523's climbs store, and it is a size question. The rows are the whole answer and they fit: ~200 KB of JSON today, ~840 KB at the hard `MAX_ITEMS_PER_SHARD` cap, against Vercel's 2 MB Data Cache entry ceiling. The climb item list is >10 MB, which is the only reason it needed a Postgres-backed summary instead. So this fixes the index **and** the shard route; a summary would have fixed only the index. No new table, no cron, no advisory lock.

### Two details that are load-bearing — please don't let review simplify them away

**1. The ISO rehydration.** A naive `unstable_cache` wrapper makes this bug *worse*, not better. The Data Cache JSON-serialises, so a cached `Date` comes back a **string**, and `renderLastMod` (`sitemap-xml.ts:69`) calls `lastModified.toISOString()` on it — a TypeError that 503s `/sitemaps/playlists.xml` and makes `buildIndexEntry` throw. So the cache stores `updatedAtIso` and the public wrapper rehydrates with `new Date(...)`, once per TTL. Exact prior art in the same directory: `cachedTier2Summary` in `climb-query.ts`. The reasoning is in a comment so nobody removes the pair later.

No `to_char` is needed: `playlists.updated_at` is a `timestamp` holding UTC and comes back through drizzle's timestamp decoder (`new Date(value + '+0000')`), so the `.toISOString()` round trip is exact to the millisecond. `climb-query.ts` needs `to_char` only because its raw `sql` fragment bypasses that decoder.

The test mocks the Data Cache as a **JSON round trip**, not a pass-through, so this stays honest — delete the ISO/`new Date` pair and two tests go red immediately.

**2. The `after()` warm is required, not belt-and-braces.** On a first-population MISS, `unstable_cache` registers its cache write in `workStore.pendingRevalidates` only *after* the callback resolves (`next/dist/server/web/spec-extension/unstable-cache.js`), while the route module snapshots `Object.values(workStore.pendingRevalidates)` into `pendingWaitUntil` at response time (`route-modules/app-route/module.js:224`). An index that abandoned the query at 3 s has already returned, so its eventual write lands in an array nobody is holding and a freezing Vercel instance can drop it — and the next request misses again. Running the same fetch inside `after()` puts it under `withExecuteRevalidates` (`server/after/after-context.js:107`), whose `finally` diffs the store and awaits writes that appeared while the callbacks ran; the abandoned query is covered too, because the warm shares its in-flight promise. This is the structural note #4522 raised, made concrete. Without it, the fix heals only probabilistically.

The warm returns without touching Postgres when the in-process cache is fresh, and a persistently failing query is held off by a 60 s per-instance floor rather than re-running on every crawl hit.

### Also here

- **`SET LOCAL statement_timeout = '15s'`** in an explicit transaction. `withDeadline` stops *waiting* at 3 s but does not cancel, so an abandoned query previously kept a connection out of a pool of ten for as long as it liked. Fifteen seconds bounds a pathological plan; it is deliberately not tuned to the deadline, because `/sitemaps/playlists.xml` legitimately takes ~4 s and a deadline-tight timeout would break a working URL. `SET LOCAL` in a transaction is the only form available — the `DB_STATEMENT_TIMEOUT_MS` startup parameter stays off because PgBouncer transaction pooling rejects it (`packages/db/src/client/postgres.ts`). **This step is cleanly separable**: it is confined to `fetchPlaylistRowsFromDb`, and a reviewer who wants the caching change isolated can ask for it to be dropped without touching anything else.
- **`scripts/production-smoke.ts`**: `playlists` is now in `REQUIRED_SITEMAP_SHARDS` as `degradable: true`. The old comment excluded it as "legitimately empty" — production serves 2,688 public playlists holding a climb, so that justification was stale, and it is why the post-deploy smoke never caught this.
- **Two stale comments in `shard-registry.ts`**: "`fetchPlaylistSitemapRows` remains uncached", and the `withDeadline` note that a statement timeout is a follow-up. Comment blocks only — no structural edits, to stay out of #4579's way.
- **`docs/sitemap.md`**: a playlists section covering both layers, the rehydration rule and the 503 it prevents, why the warm lives in `after()`, and the statement-timeout bound.

### Honest caveats

- **#4555 removes the sixteen climb scans from the index request path**, so some of the remaining playlists degradation will disappear on its own. This PR's own contribution has to be re-measured after *both* deploy, not attributed from the 4/12 baseline alone.
- **The very first `/sitemap.xml` after a Data Cache purge still degrades.** The `after()` hook is what heals it on that same request rather than one crawl later; it does not prevent the first miss.
- **No `playlists` row in `sitemap_shard_refreshes`.** That would need a summary concept added to `SitemapShard`/`buildIndexEntry` — shared infrastructure #4555 and #4579 are both already editing. Promoting it later is purely additive; re-measure first.
- #4583 (a longer deadline for paged summaries) is adjacent and would reduce the urgency here, but not the uncached query. Nothing in this PR depends on it.
- **`SHARD_DEADLINE_MS` is untouched.** The issue forbids raising it and the constant's own doc comment explains why.

### Stacking

**Based on `fix/4523-sitemap-climbs-shard` (#4555), not `main`** — that PR rewrites three of the files this one touches (`app/sitemap.xml/route.ts`, `docs/sitemap.md`, `scripts/production-smoke.ts`), so basing on main guaranteed three conflicts. The core change (`playlist-query.ts`) is untouched by #4555.

Review #4555 first; this diff is against it. **If #4555 merges first, retarget this to `main`**; if it is reworked or closed, this rebases onto main — the conflict surface is only those three files. This stays a draft until #4555 is ready.

Follow-up filed: #4618 — fixed sitemap shards have no byte budget (`shardRouteHandler` checks `MAX_URLS_PER_SHARD` but never `MAX_SHARD_BYTES` on the fixed path), and `/sitemaps/playlists.xml` already renders 2,326,713 bytes.

Closes #4524

## Release Notes

## Test plan

- `vp test run --project web --reporter=agent packages/web/app/lib/seo/sitemap` — 13 files, 146 tests green. `climb-shards.test.ts`, `shard-registry.test.ts` and `shard-route-handler.test.ts` all mock `fetchPlaylistSitemapRows` by name and pass **unchanged**, which is the check that the exported signature did not drift.
- `vp test run --project web --reporter=agent` — full web suite.
- New `packages/web/app/lib/seo/sitemap/__tests__/playlist-query.test.ts`, 11 tests. Each oracle was mutation-tested — mutation applied, watched red, reverted, tree confirmed clean:
  - drop the ISO/`new Date` pair → "hands back real Date instances" and "renders through the real urlset" both red;
  - drop the single-flight → "collapses concurrent cold callers" red;
  - drop the `SET LOCAL statement_timeout` → "bounds the query with a statement timeout" red;
  - drop the warm retry floor → "holds a failing query off behind the retry floor" red.
- `playlist-entries.test.ts` updated only for the new `db` parameter; every existing `.toSQL()` assertion survives unchanged.
- `vp check` — no lint or format findings in any changed file. The 2 repo-wide `no-floating-promises` errors are in `packages/db/scripts/dedupe-beta-links-args.test.ts`, pre-existing on `main` (fb2d59220). `after(warmPlaylistSitemapCache)` passes a function reference, so it adds no new warning.
- `vp run typecheck`.
- **Local dev DB cannot measure this and no local timing is claimed**: the dev image has 40 playlists, 4 public, 0 `playlist_climbs`. Every number above is production HTTP.
- **Post-deploy, and the only check that actually proves it** (not run yet — needs the deploy): sample `GET /sitemap.xml?probe=<random>` at least 12 times with distinct busting params, confirm `x-vercel-cache: MISS` on each, and assert `x-sitemap-degraded` never names `playlists`. Baseline to beat: 4/12, measured 2026-08-19.
- **Post-deploy**: `/sitemaps/playlists.xml` must still 200 with ~2,688 `<url>` blocks and a real `<lastmod>` — not a 503, and not a missing or `Invalid Date` timestamp, which is the rehydration regression in production form. Confirm the `<lastmod>` values are unchanged to the millisecond before vs after: a timezone shift is the silent failure mode of the ISO round trip.

### DIFF
--- a/docs/sitemap.md
+++ b/docs/sitemap.md
@@ -139,6 +139,79 @@ That lock serialises the _write_. Two instances can still _compute_ concurrently
 `superseded` check means the later finisher writes nothing instead of clobbering a
 newer answer.
 
+## The playlists cache
+
+`playlists` is a fixed shard, so the index asks it for the whole item list — one
+indexed query for every public playlist holding at least one climb. Uncached, that
+ran live on every `force-dynamic` index request. Measured across 12 distinct origin
+computations of the production index on 2026-08-19, `playlists` was named in
+`X-Sitemap-Degraded` in 4 of them, dropping 10,752 locale-expanded URLs each time
+(#4524). The shard route itself measured 1.3–4.3 s per CDN miss.
+
+It gets the boards treatment, not the climbs treatment, and the difference is a size
+question. The rows are the whole answer and they are small: 2,688 rows of
+`{ uuid, updatedAtIso }` is ~200 KB of JSON, ~840 KB at the hard `MAX_ITEMS_PER_SHARD`
+cap, against Vercel's 2 MB Data Cache entry ceiling. The climb item list is >10 MB,
+which is the only reason it needed a Postgres-backed summary store instead. So
+playlists caches the answer, which fixes the index _and_ `/sitemaps/playlists.xml`;
+a summary would have fixed only the index.
+
+Two layers, in `playlist-query.ts`:
+
+- **Next Data Cache** (`unstable_cache`, 1 hour, tag `sitemap-playlists`) — shared
+  across instances, so a cold lambda does not re-run the query.
+- **In-process TTL + single-flight** (1 hour) — `unstable_cache` does not dedupe
+  concurrent misses, and a crawl burst is the index plus the shard route arriving
+  together. Nothing is stored on rejection, so a failure is never memoised.
+
+### Dates do not survive the Data Cache
+
+`unstable_cache` JSON-serialises. A cached `Date` comes back a **string**, and
+`renderLastMod` calls `lastModified.toISOString()` on it — a TypeError that 503s
+`/sitemaps/playlists.xml` and degrades the index harder than no cache at all. So the
+cache stores `updatedAtIso` and the wrapper rehydrates with `new Date(...)`, once per
+TTL. A naive `unstable_cache` wrapper makes this bug worse; do not "simplify" that
+pair away. The climbs summary has the same pattern for the same reason.
+
+The round trip is lossless without a `to_char`: `playlists.updated_at` is a
+`timestamp` holding UTC and comes back through drizzle's timestamp decoder
+(`new Date(value + '+0000')`). `climb-query.ts` needs `to_char` only because its raw
+`sql` fragment bypasses that decoder.
+
+### Why the warm lives in `after()`
+
+`/sitemap.xml` calls `after(warmPlaylistSitemapCache)`. That is not belt-and-braces —
+without it the fix only heals probabilistically.
+
+On a first-population miss, `unstable_cache` registers its cache write in
+`workStore.pendingRevalidates` only **after** the callback resolves, while the route
+module snapshots `Object.values(workStore.pendingRevalidates)` into `pendingWaitUntil`
+at response time. An index that abandoned the query at the 3 s deadline has already
+returned, so its eventual write goes into an array nobody is holding and a freezing
+Vercel instance can drop it — and the next request misses again. Running the same
+fetch inside `after()` puts it under `withExecuteRevalidates`, whose `finally` diffs
+the store and awaits writes that appeared while the callbacks ran. The abandoned
+query is covered too, because the warm shares its in-flight promise.
+
+The warm returns without touching Postgres when the in-process cache is fresh, and a
+persistently failing query is held off by a 60 s per-instance floor rather than
+re-running on every crawl hit.
+
+### The statement timeout
+
+The query runs inside an explicit transaction with `SET LOCAL statement_timeout = '15s'`.
+`withDeadline` stops _waiting_ at 3 s but does not cancel, so before this an abandoned
+query kept a connection out of a pool of ten for as long as it liked.
+
+Fifteen seconds bounds a pathological plan; it is deliberately not tuned to the
+index's deadline, because `/sitemaps/playlists.xml` legitimately takes up to ~4 s and
+a deadline-tight timeout would break a working URL. A 57014 surfaces as a throw, which
+the index already degrades on and the shard route already 503s on.
+
+`SET LOCAL` in a transaction is the only form available: the `DB_STATEMENT_TIMEOUT_MS`
+startup parameter is deliberately off because PgBouncer transaction pooling rejects it
+and fails every connection (`packages/db/src/client/postgres.ts`, `docs/db-connectivity.md`).
+
 ## What is still slow
 
 `/sitemaps/climbs/N.xml` — the shard route, not the index. It still builds the full
@@ -147,8 +220,11 @@ instance takes ~51 s. The store fixes the index only. The fix for the route is a
 `sitemap_climb_urls` table holding the rendered paths, which is additive on the
 scaffolding above.
 
-`playlists` has no cache and no store at all: `fetchPlaylistSitemapRows` hits Postgres
-on every `force-dynamic` index request, and it is the shard degrading most often now.
+Fixed shards have **no byte budget**. `shardRouteHandler` checks `MAX_URLS_PER_SHARD`
+on the fixed path but never `MAX_SHARD_BYTES` — that guard exists only on the paged
+path. `/sitemaps/playlists.xml` already renders 2,326,713 bytes for 2,688 items
+(~866 B/item), so it crosses Vercel's 4.5 MB response ceiling at roughly 5,200 items
+while `MAX_ITEMS_PER_SHARD` lets it reach 11,250. Tracked separately.
 
 ## Operational gotchas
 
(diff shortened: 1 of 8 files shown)


marcodejongh:
Addressed the `console.warn` note in 5d8aee22e — the successful-warm line is `console.info` now, with a comment saying why (it only fires on a cold cache past the 60 s floor, so it is an expected outcome rather than something to action). The two `console.warn` calls that remain are genuinely actionable: the item-budget overflow and the warm failure.

Worth flagging one inconsistency this creates: `refreshClimbSummaryIfStale` in `climb-store.ts` logs its own self-heal success at `warn` too. Leaving that alone here so this PR stays scoped, but the same argument applies to it.

PR #4620, opened 2026-08-19
fix(mobile): bring back climb thumbnails in the share-beta picker

### LINKED ISSUE #3357: Share-beta picker: restore visual climb recognition after logbook thumbnail removal
The logbook row redesign (#3350) dropped the board thumbnail from `LogbookRow`, which the share-beta picker (`packages/mobile/app/share-beta.tsx`) reuses to let a climber pick which ascent an Instagram reel belongs to. That surface is recognition-driven: without board art, picking between similarly-named recent climbs may be harder.

Options, in rough order of preference:
1. Give the picker its own row variant with a small square thumbnail (distinct from search's tall portrait).
2. Reuse the logbook row but inject a leading thumbnail slot via prop.
3. Confirm on device that name + grade + wall label is sufficient and close as won't-fix.

Blocked on: on-device verdict from the #3350 OTA channel review.

marcodejongh:
## Summary

The share-beta picker — the screen you land on after sharing an Instagram/TikTok reel into
Boardsesh — reuses `LogbookRow` verbatim to let you pick which of your ascents the reel belongs to.
When #3350 redesigned the logbook row it stopped wrapping `ClimbListItemContent`, and that took the
board thumbnail out of every consumer of the row, the picker included. #3350's own body lists this
as known fallout.

**This is not a revert of #3350.** Dropping the art from the logbook row was deliberate and it
stays dropped: the logbook and the climbs catalog "rhymed" and people lost track of which list they
were on, and the thumbnail was the priciest cell in an infinite-scroll diary. None of that argues
for the picker. A diary is a list of things you already know you did; the picker is a recognition
surface where two similarly-named ticks on the same wall at the same angle differ only in the art.

So the picker gets its own row instead of borrowing the diary's:

- **`ShareBetaAscentRow`** (new) — leading portrait board thumbnail, climb name with the
  benchmark/no-match/mirror glyphs, a result line, wall + angle + relative time, and the trailing
  grade column with the community marker. It keeps the beta-video camera marker the logbook row
  shows. No swipe, no long-press, no edit/delete: this row is a selector.
- **`LogbookRow` is untouched behaviourally** — only its prop docs, which named the share-beta
  picker as a consumer in four places. `LogbookTab` is now its sole caller.
- **The suggestions are virtualized.** They used to render as `suggestions.map(...)` inside
  `ListHeaderComponent`, which is mounted for the life of the screen and never recycled. Harmless
  with text-only rows, wasteful now that each one decodes board art. Section headers and rows are
  now items in the FlashList `data` with `getItemType` giving them separate recycler pools. The
  list shaping (dedupe, headers, suppress-while-searching) moved to a pure
  `src/lib/share-beta-list.ts` with its own tests.

### Cell size: 60×76

Portrait, not square, and **width ≤ 80 on purpose**. `ClimbListThumbnail` renders at
`Math.max(400, width * 5)` and that number is baked into the render cache key (`..._w400_...`), so
at any width up to 80 the picker reuses the exact PNG the climbs list and play view already wrote
to disk — a climb you've seen anywhere in the app paints with zero render work. Bumping the cell to
84 would silently double the cache; there's a comment on the constant saying so.

60×76 rather than the climbs list's 76×96 because a picker wants scanability over art. **This is
the one thing a device eyeball should decide** — see the screenshots note below.

### Beyond the issue's ask: a `Beta Attached` event

The picker shipped with zero product analytics. Answering "is the missing art actually costing
anyone anything?" meant reading `$screen` counts and guessing, which is a bad way to answer a
product question. This PR adds one `SHARED_EVENTS` entry and one `track()` call on a successful
attach, with `{ source: 'suggested' | 'other', boardType, viaSearch, hasCaption }`. `source` splits
caption-matched picks from browsed ones, so the next version of this question is answerable instead
of guessed. `hasCaption` is a boolean deliberately — the caption is the user's post content and
never leaves the device.

Flagging it explicitly because it is not what #3357 asked for.

### The prior-art branch

`origin/fix/3357-share-beta-thumbnails` (head `9d471914c`, 2026-08-02, 9 files / +1220 lines) is an
earlier attempt at this issue. No PR was ever opened for it. **It is not being merged as-is** and
this PR is a fresh branch, not a rebase of it. Its structure was a good starting point and the list
module here is close to its shape, but it has four problems:

1. It resolves the board with `getBoardConfigForPlaylist(boardType, layoutId)` — the layout default.
   Commit `0d01fa56e` (2026-08-10, *after* that branch) added `renderBoardToPlaylistConfig` for
   exactly this reason: the layout-default path draws a climber with a 10×12 Kilter home wall on a
   12×14 for every one of their ascents (#4221). This PR uses the new helper, with a test that an
   ascent carrying `renderBoard` renders *that* sizeId and not the layout default.
2. Its cell is `48 × 48`, square. `ClimbListThumbnail` uses `contentFit="contain"`, and board images
   run 1080×1170 to 1080×2498 — a 1080×2498 Kilter original in a 48×48 box renders 21 px wide. A
   21-px board is not a recognition cue.
3. It adds the new i18n key to `en-US`, `es` and `fr` only. `de` exists on main now and
   `catalog-completeness.test.ts` enforces per-namespace parity across all four — it would land red.
4. It bundles three unrelated changes with the fix: new `isSearching` debounce semantics, a
   rewrite of `handleAttach` onto refs with a double-tap latch, and an end-reached latch. Defensible
   on their own, none of them this issue.

**Disposition: delete `fix/3357-share-beta-thumbnails` once this merges.** It has no PR, it predates
both `renderBoard` and the `de` locale, and everything worth keeping from it is in this branch.

### Screenshots

**None captured — this needs a device eyeball. @marcodejongh**

I could not produce a real before/after of this screen from here, and I'm not going to paste
something I didn't take:

- The iOS simulator path (`vp run mobile:screenshots` / `mobile:ios-shots`) is macOS-only.
- The Android emulator path needs the picker populated with real logged ascents. The picker only
  renders for a signed-in account with ticks, so it needs `--backend local` against a seeded dev
  server — and `vp run dev` here dies during the migration sync against the shared dev Postgres
  (`0194_location_sync_unfreeze_audit` → `type "location_sync_entity_type" already exists`, a
  parallel session's DB state, unrelated to this change). The prod test account needs a password I
  don't have, and the `test`/`test` pair only exists on the local backend.

The cheap way to judge it: this PR is JS-only and OTA-safe, and the `pr-4620` preview channel has
published (iOS + Android). On a store/TestFlight build, What's New → "Try a preview" → this PR,
then share any Instagram/TikTok reel into Boardsesh. Or deep-link straight in:

    com.boardsesh.app://share-beta?link=https%3A%2F%2Fwww.instagram.com%2Freel%2FC0000000000%2F

**What to look at:** whether 60×76 is the right cell, or whether it wants the climbs list's 76×96
(bigger art, fewer rows per screen). Those are the only two real options — anything wider than 80px
leaves the shared `_w400_` render cache. Swapping is a one-line change to `THUMBNAIL_SIZE` in
`ShareBetaAscentRow.tsx`; say the word and I'll push it.

The rest of the device pass, once you're in there:

1. Art paints for both the "Matched from the caption" rows and the "Your other ascents" rows.
2. A climber whose registered wall is smaller than the layout default sees THEIR board size (the
   #4221 guard — unit-tested, but worth one real look).
3. A frameless MoonBoard tick shows the neutral lightbulb tile, not a blank cell.
4. VoiceOver/TalkBack reads one element per row and never announces the thumbnail.
5. Fast-scroll the browse list: no recycled row shows the previous ascent's art, and a tap right
   after a fast scroll attaches to the climb actually under your finger.
6. Dark mode and the Material variant.
7. You → Logbook still has NO thumbnail. That removal was deliberate in #3350 and must stay.

## Release Notes

- Sharing a reel into Boardsesh shows the wall for every climb again, so you can spot the right send
  instead of reading down a list of names.

## Test plan

- `vp test run --project mobile share-beta --reporter=agent` — 25 tests across the new row, the pure
  list module and the screen. Covers the `renderBoard` regression guard (#4221), the ≤80 px
  cache-key contract, the frameless-tick fallback tile, mirror pass-through, one-a11y-element-per-row
  with the thumbnail hidden from assistive tech, FlashList recycle safety (a recycled row attaches to
  the new ascent, not the one that used to be in the cell), the suggestions living in `data` rather
  than `ListHeaderComponent`, the no-duplicate-climb rule, and the analytics props.
- `vp test run --project i18n --reporter=agent` — 101 pass; catalog parity across en-US/es/fr/de.
- `vp run check:i18n` — OK.
- `vp run test:mobile` — 6,573 tests across 687 files, all green.

JS-only under `packages/mobile/{app,src}`, `packages/shared/analytics` and `packages/shared/i18n`, so
this is OTA-safe and `ota-check` should report no native change.

Closes #3357

### DIFF
--- a/packages/mobile/app/__tests__/share-beta.test.tsx
+++ b/packages/mobile/app/__tests__/share-beta.test.tsx
@@ -1,0 +1,353 @@
+// @vitest-environment jsdom
+import { beforeEach, describe, expect, it, vi } from 'vitest';
+import { act, fireEvent, render } from '@testing-library/react';
+import { createElement, type ReactNode } from 'react';
+import type { AscentFeedItem } from '@boardsesh/graphql/operations';
+import type { ShareBetaAscentSource, ShareBetaListItem } from '../../src/lib/share-beta-list';
+
+type MutationCallbacks = {
+  onSuccess: () => void;
+  onError: (error: unknown) => void;
+};
+
+const router = vi.hoisted(() => ({ back: vi.fn(), replace: vi.fn() }));
+const showToast = vi.hoisted(() => vi.fn());
+const notificationAsync = vi.hoisted(() => vi.fn().mockResolvedValue(undefined));
+const fetchNextPage = vi.hoisted(() => vi.fn());
+const mutate = vi.hoisted(() => vi.fn());
+const track = vi.hoisted(() => vi.fn());
+const flashListState = vi.hoisted(() => ({
+  props: null as null | { data: unknown[]; getItemType: (item: never) => string },
+}));
+
+const state = vi.hoisted(() => ({
+  link: 'https://instagram.com/reel/abc',
+  authenticated: true,
+  profileId: 'user-1' as string | undefined,
+  caption: 'Purple People Eater on the garage wall' as string | null,
+  ascents: [] as AscentFeedItem[],
+  suggestions: [] as AscentFeedItem[],
+  feedPending: false,
+  hasNextPage: false,
+  fetchingNextPage: false,
+  attachPending: false,
+}));
+
+function makeAscent(uuid: string, climbUuid = `climb-${uuid}`): AscentFeedItem {
+  return {
+    uuid,
+    climbUuid,
+    climbName: `Climb ${uuid}`,
+    setterUsername: null,
+    boardType: 'kilter',
+    boardId: null,
+    boardDisplayName: null,
+    layoutId: 8,
+    angle: 40,
+    isMirror: false,
+    status: 'send',
+    attemptCount: 2,
+    quality: null,
+    difficulty: 20,
+    difficultyName: 'V4',
+    consensusDifficulty: 20,
+    consensusDifficultyName: 'V4',
+    boardseshDifficulty: null,
+    boardseshConfidence: null,
+    qualityAverage: null,
+    isBenchmark: false,
+    isNoMatch: false,
+    comment: '',
+    climbedAt: '2026-07-31T12:00:00.000Z',
+    frames: 'p1r1',
+  };
+}
+
+vi.mock('react-native', () => ({
+  View: ({ children }: { children?: ReactNode }) => createElement('div', null, children),
+  TextInput: ({
+    value,
+    onChangeText,
+    placeholder,
+  }: {
+    value: string;
+    onChangeText: (text: string) => void;
+    placeholder?: string;
+  }) =>
+    createElement('input', {
+      value,
+      placeholder,
+      onChange: (event: { target: { value: string } }) => onChangeText(event.target.value),
+    }),
+  Pressable: ({
+    children,
+    onPress,
+    accessibilityLabel,
+  }: {
+    children?: ReactNode;
+    onPress?: () => void;
+    accessibilityLabel?: string;
+  }) => createElement('button', { type: 'button', onClick: onPress, 'aria-label': accessibilityLabel }, children),
+  StyleSheet: { create: (styles: Record<string, unknown>) => styles, hairlineWidth: 1 },
+}));
+
+vi.mock('expo-router', () => ({
+  useLocalSearchParams: () => ({ link: state.link }),
+  useRouter: () => router,
+}));
+
+vi.mock('react-i18next', () => ({
+  useTranslation: () => ({
+    t: (key: string) => {
+      const labels: Record<string, string> = {
+        'mobile.betaVideos.attachSuccess': 'Beta attached',
+        'mobile.betaVideos.attachError': 'Could not attach',
+        'mobile.betaVideos.shareNoAscents': 'No ascents',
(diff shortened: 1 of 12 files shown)


marcodejongh:
On the `estimatedItemSize` note: that prop doesn't exist on the FlashList we ship. `packages/mobile/package.json` pins `@shopify/flash-list` at `2.3.2`, and v2 removed `estimatedItemSize` in favour of self-measuring rows — mixed row heights are its normal case, which is why `getItemType` is the only thing needed here. Three existing v2 call sites in this repo carry the same note (`BoardCarousel.tsx:41`, `licenses.tsx:99`, `changelog.tsx:294`); I've added it to this list too so the next reader doesn't have to look it up.

`overrideItemLayout` is still a real v2 prop, but it's for spans in a masonry/multi-column layout, not height hints — this is a single-column list, so it has nothing to say here.

Issue #3891, opened 2026-07-24
Mobile: playlist climb-add is unreliable, and you can't scroll a playlist's climbs (iOS)


marcodejongh:
Split out from #3825 (which is the **web** Discover bug). This tracks the **mobile app** symptoms raised in a comment there by @artbegolli, on a different surface (React Native), so they don't get lost.

## Reporter's words (iPhone, "minimoon 25")

> I've had this as well for minimoon 25 on iphone. Specifically adding climbs to a playlist is inconsistent and sometimes adds and sometimes doesn't even after confirming it was added. Separately - from a playlist you can't seem to scroll through climbs that are added to a playlist.

## Two distinct issues to verify

1. **Add-to-playlist is unreliable** — the confirmation says the climb was added, but it sometimes isn't. Points at an optimistic-UI / mutation path that reports success before (or instead of) the write landing. Data-integrity smell.
2. **Can't scroll a playlist's climb list** — from a playlist detail view on iOS, the added climbs don't scroll. If reproduced, this makes a populated playlist effectively unusable to browse.

## Notes / next steps

- Both are single-user, unconfirmed reports on iOS — needs repro on a device before a fix.
- Filed at **P3** per triage default. If the scroll bug reproduces, it likely warrants **P2** (a core playlist view being unscrollable is more than a papercut) — re-triage after repro.
- The mobile playlist detail + add-to-playlist flows live in `packages/mobile/`, separate from the web fix in #3825.

Referenced from the #3825 PR.

marcodejongh:
## Design findings from the web #3825 analysis (transfer to RN Discover)

The Next.js web Discover surface is being deprecated (web climbing bugs no longer fixed; PR #3892 / issue #3825 closed on that policy). But the root-cause analysis transfers directly to the Expo/RN Discover surface — carrying it here so it's on hand when that work happens.

The web repro: a boardless user browsing a Kilter wall opened Discover and got a feed dominated by **Tension** recommendation playlists (verified: they opened 3 playlists, all Tension in prod, while on a Kilter session). Opening them was a dead end. Three design fixes came out of it:

1. **Board-aware Discover rail.** Scope the Discover feed to the board the user is actually on — including the active session / route board, not just a board they've *saved*. The public playlist pool skews ~2:1 Tension:Kilter and "popular" ranks by climb count (50-climb generated playlists lead), so an unfiltered feed buries the user's own board. Fall back to a multi-board feed only when there's genuinely no board context.

[...]

PR #4621, opened 2026-08-19
fix(mobile): make the playlist Edit climbs menu row work

### LINKED ISSUE #3966: 🐞 Bug report: Not able to manage playlists. Can't add from menu option only from climb. Can't rename playlist. Can't...

## Comment

Not able to manage playlists.
Can't add from menu option only from climb.
Can't rename playlist.
Can't delete climbs from playlist.
Can't play a playlist. Only the first climb gets added to the queue.

## Metadata

| Field | Value |
| --- | --- |
| Platform | android |
| App version | 2.2.2 (2000587) |
| Source | drawer-bug |
| Board | moonboard / layout 3 / size 1 / sets [5,6,7,8,9,10] @ 40° |
| Climb | unknown |
| Session | unknown |
| URL | /home |
| User agent | unknown |
| Feedback row | app_feedback #10194 |

## Contact

✅ Reporter agreed to follow-up — resolve **app_feedback #10194** in the DB to reach them.

marcodejongh:
## Summary

Closes #3966.

The reporter filed four complaints against Android 2.2.2. Three are still live on `main`; one was
fixed after the report landed.

**The primary bug is a regression, and it is the reason the other two look broken.** In
`packages/mobile/app/(tabs)/discover/[playlist_uuid].tsx`, the owner overflow menu's "Edit climbs"
row did not enter edit mode. `menuEnterEdit` only set `pendingEditClimbsRef.current = true` and
`setActionsVisible(false)`; the actual `enterEditMode()` call lived in `handleActionsClose`, which is
wired to `PlaylistActionsMenu` → `ModalSheet`'s `onClose`. But `onClose` fires **only** on a user
pan-down / backdrop tap. A controlled `visible: true -> false` goes through the sheet coordinator,
which sets `selfDismissRef` and deliberately swallows `onClose`
(`src/providers/sheet-presentation-provider.tsx`; the contract is already pinned by the
`a coordinator-driven dismiss does NOT fire onClose (selfDismissRef gate)` test). So the row closed
the sheet and did nothing. The stale `pendingEditClimbsRef = true` then dropped the screen into edit
mode the next time the user swiped that menu away, which reads as a second, unrelated glitch.

Regression origin: `8973ec8f1` (2026-07-02) swapped `ModalSheet`'s `onDismiss` (gesture **or**
programmatic) for `onClose` (user-initiated only). Its commit message states the semantic change, but
the one consumer that depended on the old behaviour — this deferred handoff, added 2026-06-10 in
`3e577cfef` — was not audited.

Why it reads as "can't rename / can't delete climbs", and why on Android: edit mode is the only route
to both. Removing a climb is the per-row control in edit mode; renaming is the cog beside the
playlist name, which `PlaylistDetailView` renders only when `editMode && onEditDetails`. On the
Material (Android) variant `PlaylistDetailView` always renders the collapsed form, so the overflow
menu is the only owner affordance and both actions were permanently unreachable. On iOS/Liquid Glass
the un-collapsed toolbar calls `enterEditMode` directly, so it only broke once the hero scrolled
away.

### The four complaints

| # | Reported | Status |
| --- | --- | --- |
| 1 | "Can't rename playlist" | **Fixed here** — rename gets its own "Edit details" row, and the edit-climbs route it used to hide behind now works. |
| 2 | "Can't delete climbs from playlist" | **Fixed here** — the per-row remove control lives in edit mode, which the menu now actually enters. |
| 3 | "Can't add from menu option only from climb" | **Partly fixed here** — an "Add climbs" row and an empty-state CTA now land on the Climbs tab instead of leaving a dead end. Adding still runs through a climb's own action menu; a real "adding to `<playlist>`" selection mode on the climbs list is a separate, larger change, filed as #4630. |
| 4 | "Can't play a playlist. Only the first climb gets added to the queue" | **Already fixed** by `b487e6cfc` + `bae27879c` (#3997 / #4010, issue #3891) — the MoonBoard `compatible_size_ids` / `required_set_ids` filters were dropping every row from the playlist-climbs join. Backend-side, so it reached the reporter without an app update. Nothing in this PR. |

### What changed

- Every overflow-menu row now acts immediately and closes the sheet itself. `pendingEditClimbsRef` is
  gone and `handleActionsClose` is back to just clearing `actionsVisible`. A comment on both the
  screen and the component points at the `selfDismissRef` contract test so nobody re-adds an
  onClose-deferred handoff.
- New "Edit details" row (cog) opens the rename / colour / icon / visibility sheet. The coordinator
  already serialises a dismiss-then-present handoff, so this needs no deferral either.
- New "Add climbs" row + empty-state CTA, both routing to the Climbs tab. Gated on
  `isOwner && !boardBanner` — when the playlist belongs to another board the existing switch-board
  banner owns that prompt.
- "Edit climbs" is relabelled "Reorder & remove climbs" now that details editing is its own row.
- `PlaylistDetailView` gains an optional `emptyAction`, rendered in **both** the Material and Liquid
  Glass empty states.
- Sheet snap points go `36%` → `46%` (four rows) / `56%` (five, with add-climbs).
- New keys `detail.menu.addClimbs` and `detail.menu.editDetails` plus the relabelled
  `detail.menu.editClimbs`, in all four locales per the `docs/i18n-*-glossary.md` files (es
  "bloques", fr "blocs", de "Boulder").

No native fingerprint inputs touched — everything is under `packages/mobile/src`,
`packages/mobile/app`, and `packages/shared/i18n/locales/**`. Ships over OTA.

## Release Notes

- Managing your own playlist works again on Android: the ⋯ menu's edit row now actually opens edit mode, so you can reorder and pull climbs out.
- Rename a playlist, change its colour or icon straight from the ⋯ menu — no more hunting for the cog.
- Empty playlist? There's a button that takes you to the climbs list instead of a dead end.

## Test plan

Automated:

- **Regression guard** in `packages/mobile/app/(tabs)/discover/__tests__/playlist-uuid.test.tsx` fires
  each menu row **without ever calling the sheet's `onClose`** — the exact sequence a real tap
  produces. Verified it FAILS on the old wiring (reverted the two callbacks locally: 1 failed /
  15 passed, `data-edit-mode` stuck at `"false"`) and passes on this branch.
- Same file: the rename row opens the form sheet at `mode: 'edit'` without dragging the screen into
  edit mode; a menu-only swipe-away does not enter edit mode; add-climbs and the empty CTA both call
  `router.navigate('/(tabs)/climbs')`; both add affordances disappear behind a board banner and for
  non-owners.
- `playlist-actions-menu.test.tsx`: row order pin / add / details / climbs / delete, the add row
  drops when the handler is omitted, and each row fires its own handler exactly once while `onClose`
  stays untouched.
- `playlist-detail-view.test.tsx`: the empty-state CTA renders and fires in both variants, and the
  empty state stays CTA-free without `emptyAction`.

Commands (all green):

- `vp run typecheck:mobile`
- `vp run test:mobile` — 684 files / 6559 tests passed
- `vp run check:i18n` — 323 .tsx scanned, no hardcoded strings
- `vp test run --reporter=agent catalog-completeness` — 96 passed (four-locale key parity)
- `vp check`

**Device QA is owed and I could not capture it.** No Android device or working emulator in this
environment, and the flows in question are native sheet + Material Appbar behaviour that jsdom tests
cannot stand in for. @marcodejongh — please exercise on Android (Material variant), on a playlist you
own:

1. ⋯ → **Reorder & remove climbs** enters edit mode (drag handles + red remove controls appear, Done
   replaces the ⋯).
2. ⋯ → **Edit details** opens the rename sheet, and a rename persists.
3. ⋯ → **Add climbs** lands on the Climbs tab; an empty playlist shows the same CTA in its empty
   state.
4. Removing a climb decrements the hero count.
5. Open ⋯ and swipe the sheet **down** — the screen must NOT jump into edit mode.
6. Sanity-check the five-row sheet is not clipped at `56%` on a short screen.

### DIFF
--- a/packages/mobile/app/(tabs)/discover/[playlist_uuid].tsx
+++ b/packages/mobile/app/(tabs)/discover/[playlist_uuid].tsx
@@ -1,6 +1,6 @@
 import { useCallback, useEffect, useMemo, useRef, useState } from 'react';
 import { View, StyleSheet, Alert, Pressable } from 'react-native';
-import { useLocalSearchParams, useNavigation } from 'expo-router';
+import { useLocalSearchParams, useNavigation, useRouter } from 'expo-router';
 import type { BottomSheet } from '@expo/ui/community/bottom-sheet';
 import { useQuery, useQueryClient } from '@tanstack/react-query';
 import { useTranslation } from 'react-i18next';
@@ -61,6 +61,7 @@ export default function PlaylistDetail() {
   const { playlist_uuid: playlistUuid } = useLocalSearchParams<DetailParams>();
   const { t } = useTranslation('playlists');
   const navigation = useNavigation();
+  const router = useRouter();
   const queryClient = useQueryClient();
   const { isAuthenticated } = useAuth();
   const { showToast } = useToast();
@@ -211,10 +212,8 @@ export default function PlaylistDetail() {
   const [editMode, setEditMode] = useState(false);
   const [editClimbs, setEditClimbs] = useState<Climb[]>([]);
   const editClimbsRef = useRef<Climb[]>([]);
-  const pendingEditClimbsRef = useRef(false);
-  // Always-current view of the loaded climbs so enterEditMode — which runs
-  // asynchronously after the actions sheet finishes dismissing — seeds from the
-  // latest data, not the snapshot captured when the menu opened.
+  // Always-current view of the loaded climbs so enterEditMode seeds from the
+  // latest data, not the snapshot captured when the actions menu opened.
   const allClimbsRef = useRef(allClimbs);
   allClimbsRef.current = allClimbs;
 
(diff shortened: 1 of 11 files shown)


marcodejongh:
Thanks — answering the three points.

### `menuEditDetails` synchronous dismiss+present

> Confirm the coordinator test for this case exists, or add one.

It exists. `packages/mobile/src/providers/__tests__/sheet-presentation-provider.test.tsx:60`:

```
it('serializes a sheet-over-sheet handoff: dismiss A, settle, then present B', () => {
  ...
  coordinator.setDesiredOpen('b', true);
  expect(b.present).not.toHaveBeenCalled();     // B is HELD
  expect(a.dismiss).toHaveBeenCalledTimes(1);   // A's dismiss starts

  vi.advanceTimersByTime(SETTLE_MS);            // A's dismiss settles
  expect(b.present).toHaveBeenCalledTimes(1);   // only now does B present
```

That is exactly the dismiss-then-present sequencing you asked about: B's `present` is provably not
called until A's dismiss has settled, so the edit-details sheet cannot animate over a half-dismissed
actions sheet. Line 83 (`never presents two transitions at once (B waits while A is presenting)`)
covers the mirror case.

The serialization is per presenter group, so the one thing that could break it is the two sheets
landing in different groups. They don't: `PlaylistActionsMenu` and `PlaylistFormSheet` both render
`<ModalSheet>` with no `presenterGroup` prop (`PlaylistActionsMenu.tsx:59`,
`PlaylistFormSheet.tsx:154`), so both take the default root-window group.

### `handleActionsClose` is now purely defensive

Correct, and deliberate. `onClose` is a required prop on `ModalSheet` for the user pan-down /
backdrop path, and that is now its only caller — which is the point of the fix. Leaving it as a bare
`setActionsVisible(false)` is what makes the swipe-away path a no-op instead of the surprise
edit-mode jump the stale `pendingEditClimbsRef` used to cause; there's a regression test pinning
exactly that (`does not enter edit mode when the menu is only swiped away`).

### Device QA gap

Agreed, and it is called out in the PR body rather than papered over — no Android device or working
emulator here, and these are native sheet + Material Appbar behaviours jsdom can't stand in for. The
six flows to exercise are listed at the bottom of the test plan. @marcodejongh this needs a real
Android pass before merge.

PR #4638, opened 2026-08-20
feat: estimate the Boardsesh grade for angles nobody has climbed

fallaciousreasoning:
## Summary

- The nightly grade model can now project a climb's grade at an unclimbed angle from real crowd evidence at 2+ other angles. Projections require fitted angle-surface coverage for the target and contributing siblings, and wide/unsupported estimates stay unpublished.
- Adds the `cross_angle_estimate` confidence tier and stamps the changed model as `v2.1`. Projected grades carry a real uncertainty band, cannot seed a climber's first tick grade, and remain distinct from ascent-backed grades throughout GraphQL, queue, offline, and activity-feed paths.
- Validation now hides well-sampled angles and requires a ≥0.01 MAE win over the naive sibling mean independently for each board. A weak or undersampled board withholds only its projections; it cannot inherit a pass from higher-volume boards.
- Mobile renders projections as muted `≈` grades and hollow chart diamonds, with a visible estimate legend and screen-reader labels that distinguish measured from projected points.
- Publication is intentionally client-first: the nightly writer requires `--publish-cross-angle-estimates`. Keep it off until a compatible mobile build is the minimum supported version, then enable it only for boards whose per-board gate passes.

## Release Notes

See estimated grades at unclimbed wall angles, clearly marked with `≈` and hollow chart diamonds.

## Test plan

1. Run DB tests; expect 572 passed and 12 skipped.
2. Run targeted mobile tests; expect all 44 to pass.
3. Typecheck DB, mobile, and GraphQL; expect no errors.
4. Inspect projected grades; expect muted `≈` and hollow diamonds.
5. Inspect measured grades; expect existing display and tick defaults.

## Risk

Risk: 3/5 — changes persisted grade projections, guarded by rollout flag and per-board gates.

Full-repo `vp check` is currently blocked by unrelated failures on rebased `main`; all changed areas and pre-commit checks pass. Two independent code-review agents are clean after fixes.

PR #4640, opened 2026-08-21
feat(mobile): add settings to control when swipe/tap auto-light the board

fallaciousreasoning:
## Summary

Two new settings under Settings > Board lighting (and now also in the BLE lightbulb's long-press control sheet), both default **on**, matching current behavior:

- **Light problems on swipe** — swiping (or the prev/next chevrons) between climbs in the play view lights the board.
- **Light problems when tapping the list** — tapping a climb in a climbs list lights the board.

### How it works

Turning either setting off routes that navigation through the play drawer's existing **view-only Preview flow** — the same "Preview" badge + "Set active" affordance already used for the wrong-board drawer path, browsing a playlist, and the explicit "Preview" climb action. The climb is shown without touching the queue or the board; the climber taps "Set active" (now shown with a lightbulb glyph) to promote and light it.

- `play-drawer-navigation.ts`: `getViewOnlyPreviewNavigationTarget` gains a `forceViewOnly` param.
- `PlayDrawer.tsx`: `handleNext`/`handlePrev` pass `forceViewOnly: !lightOnSwipe`. `navigationState` is already derived from `displayedQueueItem` (`drawerPreviewItem ?? currentClimbQueueItem`), so repeated swipes chain through a preview without ever committing, and the existing `handleSetActive` already commits + lights + clears the preview state — no changes needed there.
- `climbs/index.tsx`: `handleClimbPress` opens the drawer with `previewQueueItem` (`openPlayDrawer`) instead of committing via `activateClimbListClimb.activate()` when `lightOnClimbTap` is off — mirrors the existing "Preview" climb-action row exactly.
- `PlayDrawerPreviewBanner.tsx`: "Set active" now shows a `lightbulb.fill` glyph alongside the label.
- `BleControlSheet.tsx`/`BleControlSheetHost.tsx`: the lightbulb's long-press menu now also carries both toggles, next to auto-disconnect.

An earlier version of this PR gated the BLE auto-send itself (a post-commit suppression in `bluetooth-provider.tsx`). Review correctly flagged a real race in that approach and the mandatory Fable-review requirement for touching BLE code. Redesigning around the existing Preview flow made that gating unnecessary — the settings decision now happens *before* a commit ever reaches the queue, not as a suppression on the BLE send — so `bluetooth-provider.tsx` is untouched and **no file under `packages/mobile/src/lib/ble/` or `packages/shared/ble-protocol/` is touched by this branch**. Removed the `ble-needs-fable-review` label accordingly.

### Two things worth a conscious eyes-on before merge (both flagged by review, neither is a code bug)

- **Fable-review boundary.** `BleControlSheet.tsx`/`BleControlSheetHost.tsx` live under `packages/mobile/src/components/ble/`, not the CLAUDE.md-listed `packages/mobile/src/lib/ble/` — so the literal trigger doesn't fire, and the change there is UI-only (two toggle rows + props, no protocol/send code). Review still flagged it as "worth a conscious sign-off" since it does surface BLE-behavior controls. Not re-adding the label, but noting it here for whoever merges.
- **Party mode.** With `lightOnSwipe`/`lightOnClimbTap` off, the resulting Preview is local-only and deliberately *not* synced to other party members — a peer's device has no way to see that you're browsing (only that nothing changed, since you haven't committed anything). This is the intended behavior (a personal lighting preference shouldn't broadcast browsing intent to the party), but it's currently invisible in the UI/Settings copy. No code change proposed here; flagging so it's a conscious call, not a silent gap.

## Release Notes

- Added two settings to control when the board lights up automatically: on swipe between climbs, and on tapping a climb in a list (both on by default). When off, the climb opens as a preview you can light with one tap.

## Test plan

- [x] `vp test run --project mobile` — full suite; only pre-existing unrelated failures (already fixed in #4638, not yet merged) plus one confirmed-flaky test (passes in isolation)
- [x] `play-drawer-navigation.test.ts` — 3 cases for `forceViewOnly`, plus 3 for `getSwipeNavigationTarget` (the exact `lightOnSwipe` → `forceViewOnly` wiring `handlePrev`/`handleNext` call — PlayDrawer itself has no render test, so this is the seam that closes the gap)
- [x] `play-drawer-preview-banner.test.tsx` — lightbulb glyph coverage
- [x] `ble-control-sheet.test.tsx` — new toggle-row coverage
- [x] `index-keyboard.test.tsx` — `lightOnClimbTap=false` → preview-open coverage
- [x] `vp run check:i18n` / catalog-completeness — reuses the `ble.lighting.*` keys already added to all 4 locales
- [x] `vp check`
- [ ] Manual verification on a real connected board (not yet run in this session)


fallaciousreasoning:
@marcodejongh been testing this on my board and it works pretty well :smile:

PR #4641, opened 2026-08-21
fix(mobile): touch scrolling blocked by row gestures on RN web

fallaciousreasoning:
## Summary

Fixes touch scrolling on the React Native web build (`/app`): dragging a finger up/down over the climbs list, the queue sheet, or your logbook did nothing, even though mouse-wheel scrolling worked fine.

Root cause: `react-native-gesture-handler`'s web `GestureDetector` defaults its underlying DOM node to `touch-action: none` unless a `touchAction` prop is explicitly passed. That CSS property is enforced by the browser's compositor before any JS gesture recognizer runs, so it blocks native touch-scrolling for any drag starting on that node — independent of a Pan gesture's own `activeOffsetX`/`failOffsetY` tuning (those only gate the JS recognizer, which the browser never consults once it's decided not to hand the touch to it). Mouse-wheel scrolling never goes through this path at all, which is why it kept working while touch was fully dead.

`ReanimatedSwipeable` (the built-in swipe-to-reveal component used for the queue-and-playlist swipe actions) already passes `touchAction="pan-y"` on its own internal `GestureDetector`s, but the row-level tap/long-press/pan `GestureDetector`s composed around it in `ClimbListRow`, `QueueItemRow`, `SwipeableRow`, and `LogbookRow` did not, so they silently overrode the correct behaviour with the default `none`.

Fix: add `touchAction="pan-y"` to each of those row-level `GestureDetector`s. A vertical drag now falls through to the browser's native scroll (handled by the list itself); a horizontal drag is still intercepted by the swipe/tap gesture. The drag-to-reorder handle gestures (`QueueItemRow`'s queue-reorder handle) are deliberately left untouched — they need to keep capturing vertical drags to work at all.

Confirmed via a running `vp run dev:mobile:web` session with real hardware-level touch input (see test plan) — reverting the fix reproduces the original bug 1:1, restoring it fixes it.

## Release Notes

- Scrolling by touch (finger drag) now works everywhere in the browser app — climbs list, queue, and logbook. It was completely stuck before if you weren't using a mouse wheel or trackpad.

## Test plan

- `vp run typecheck:mobile`, `vp run typecheck:web` — pass.
- `vp check` — pass (pre-existing unrelated formatting issue in `docs/mobile-ota-updates.md`, not touched by this change).
- `vp test run --project mobile --reporter=agent` — 6708/6716 pass; the 8 failures are in `logbook-tab-grouped.test.tsx` and reproduce identically on `main` with this branch's changes fully reverted (confirmed via `git stash`), so they're pre-existing and unrelated.
- Added a Playwright regression test (`packages/web/e2e/expo-web/smoke.spec.ts`, "touch-drag scrolls the climbs list") that drives real touch input through the CDP `Input` domain (a JS-dispatched synthetic `TouchEvent` does not exercise the browser's `touch-action` gating, so CDP is required to actually reproduce/verify this class of bug) and asserts the climbs list's `scrollTop` advances. Verified it fails without the fix and passes with it.
- Manually exercised `vp run dev:mobile:web` at `/app` with the `expo-web-smoke` Playwright project (`hasTouch: true`) — touch-drag scrolling works on the climbs list; ran the rest of that suite too (one unrelated pre-existing failure in `bindSeededBoard`, caused by the seed image's test account already owning boards under "Your boards" rather than needing a fresh "kilter" search — reproduces on `main`, unrelated to this fix).

Co-Authored-By: Claude Sonnet 5 <noreply@anthropic.com>

https://claude.ai/code/session_01LJGS2YnJi6BKm9Pg8VgBDk

### DIFF
--- a/docs/mobile-ota-updates.md
+++ b/docs/mobile-ota-updates.md
@@ -17,9 +17,9 @@ destructive storage re-path and an in-place stateless→control-plane key-sealin
 cutting a new native build anyway, so instead we stood up a fresh V3 server on an empty bucket + new
 Postgres and left V2 untouched. Two servers now run in parallel:
 
-| Server          | Host                    | Version                                     | Who hits it                                                                                                                                                                     |
-| --------------- | ----------------------- | ------------------------------------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
-| **V2 (frozen)** | `ota.boardsesh.com`     | axelmarciano V2, stateless                  | Old store/TestFlight binaries built before the V3 cutover. They have `ota.boardsesh.com` + the old cert baked in, so V2 keeps serving them unchanged. **Do not publish to it.** |
+| Server          | Host                    | Version                                                                                            | Who hits it                                                                                                                                                                     |
+| --------------- | ----------------------- | -------------------------------------------------------------------------------------------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| **V2 (frozen)** | `ota.boardsesh.com`     | axelmarciano V2, stateless                                                                         | Old store/TestFlight binaries built before the V3 cutover. They have `ota.boardsesh.com` + the old cert baked in, so V2 keeps serving them unchanged. **Do not publish to it.** |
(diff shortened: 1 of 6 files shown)


fallaciousreasoning:
Addressed both nits in 6c37a3513: hoisted `SEEDED_BOARD_NAME` to module scope so the touch-drag test reuses it instead of a second inline literal, and added a comment explaining the fixed 16ms `waitForTimeout` (paces the synthetic touchmove stream at ~60fps so the browser sees a plausible drag, not a burst of same-tick moves).

marcodejongh:
Mate get an opus subbie 😆

marcodejongh (approved):

PR #4660, opened 2026-08-21
fix(mobile): label every session-exit control with its verb

marcodejongh:
## Summary

Part of #4281 (the secondary complaint): the flag glyph that ends a session reads as "report this climb" to climbers who have never ended one. The Liquid Glass Record chrome already ships a labelled Stop pill; the Material app bar and the session-screen overlay strip were still bare icons.

- **Material `RecordTopChrome`**: the exit `Appbar.Action` (a 48dp icon circle that can't host a word) becomes an M3 text button — icon + "Stop" (error tint) / "Leave" (neutral), 40dp + hitSlop past 48dp, pill corners from `theme.radii`, ripple tinted from the same role as the label.
- **`SessionScreenHeader`** (both variants): icon + word matching the Record chrome, on the share button's 40dp metrics so the right cluster sits on one baseline.
- Chrome labels in both fixed-height rows now cap Dynamic Type at `CHROME_LABEL_MAX_FONT_SCALE` so they stop growing before clipping (including the existing Glass pill + invite label).

Presentation only — handlers, confirm dialogs, tints, and a11y labels unchanged. No new strings: `session:mobile.session.inStop`/`.inLeave` already ship in all four locales.

## Release Notes

- Ending a session is now a labelled button — "Stop" if you started it, "Leave" if you joined — instead of a bare flag icon.

## Test plan

- [x] `vp run test:mobile` — session-screen suite 17 files / 114 tests green; one pre-existing unrelated failure (`board-detail-fields.test.ts`, red on main, fixed in #4659)
- [x] `vp check --fix`
- [x] New test assertions: Material Stop carries the word + error-tint flag; Material Leave carries the word + neutral glyph; header leader/joiner labels
- [ ] Emulator screenshots (Material leader/joiner, fr locale for the longest label) — to follow in a comment

### DIFF
--- a/packages/mobile/src/components/session-screen/RecordTopChrome.tsx
+++ b/packages/mobile/src/components/session-screen/RecordTopChrome.tsx
@@ -12,10 +12,15 @@ import { UserAvatarToolbarAction } from '../user-drawer/UserAvatarToolbarAction'
 import { useTheme } from '../../providers/theme-provider';
 import { selectByVariant } from '../../theme/variants';
 import { spacing } from '../../theme/tokens';
+import { CHROME_LABEL_MAX_FONT_SCALE } from '../../theme/typography';
 
 // Record's defining action is the Start/End footer button, so the chrome's
 // create island is gated off — its handler is never invoked.
 const noop = () => {};
+
+// M3 text-button height. The pill sits in a 56dp app-bar row, so 40dp leaves the
+// bar's own breathing room; `hitSlop` lifts the effective target back over 48dp.
+const MATERIAL_EXIT_HEIGHT = 40;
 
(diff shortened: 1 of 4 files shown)


marcodejongh:
Re the maxWidth note: leaving it unconstrained deliberately. The label vocabulary is closed — Stop/Parar/Arrêter/Stopp and Leave/Salir/Quitter/Verlassen (9 chars worst case), with Dynamic Type capped at 1.2 — so the pill can't grow past ~110dp, and ellipsizing a single verb ("Verlas…") would be strictly worse than the title shifting a few dp, which the invite label in the same cluster already causes today.

Issue #3017, opened 2026-06-18
Consolidate preview/myclimb/litup by state left of climb name in play drawer


marcodejongh:
WE can use that spot for also swapping back to the current active queue item, and for showing the preview label too.

PR #4666, opened 2026-08-22
feat(mobile): wall-state pill + Put on the wall commit bar in the play drawer

marcodejongh:
## Summary

First slice of #4281 (browse without hijacking the wall) and the #3017 header-slot consolidation. Visual/chrome slice — the shared-session browse-by-default behaviour lands in the follow-up slice (A2).

**WallStatePill** — one tenant for the play drawer header's leading slot, answering "will my next swipe change the wall?" at a glance. Strict ladder keyed on displayed-equals-wall (never on who holds Bluetooth):

- **On the wall** — driver avatar in an amber-tinted capsule when the shown climb is the lit one. Absorbs `PlayDrawerOnWallBanner`; the lightbulb holder pip is suppressed exactly while the pill carries the face (`shouldShowHolderBadge`, contract pinned by test).
- **Live** — amber `WALL_LIVE_DOT_SIZE` dot + "Live" when the next navigation drives the wall.
- **Browsing** — accent-filled chip with the eye glyph, shown only when a browse latch *genuinely* keeps swipes view-only: the chrome derives from the same `swipeStaysViewOnly` predicate the swipe handlers use, so it can never promise more than the gesture keeps.
- Hidden in plain solo (no BLE, no session, no lit climb) — the header stays exactly as today in the most common state.

Tapping the pill opens a one-line explainer callout (Back to live / Browse from here, driver row with litAgo → profile), with a full a11y trap (modal container, initial focus, Android sibling hiding, hardware-back dismiss).

**Commit bar** — the old orange preview banner (hardcoded `systemOrange`, 4pt radius, sub-44pt text link) is deleted; while a preview is pinned, the action bar's secondary row swaps to **Back to live** + a filled **Put on the wall** button (which honestly falls back to "Set active" when no wall is reachable). Zero added height — the banner's removal returns ~32pt to board art in every preview flow. The commit affordance keys on the *pinned preview* while the browsing chrome keys on the *latch*, so a `lightOnSwipe`-on long-press Preview keeps its activation button without lying about its swipes.

**BrowseFrameOverlay** — four neutral viewfinder corner brackets over the board art while browsing (pointer-transparent, zero height).

Also: `onAccent` ink role added to `@boardsesh/velvet-tokens` (web theme now consumes it instead of a local literal); wall-state announcements move to a transition-keyed announcer so commits and Back to live still speak when the pill unmounts; `screenshotOpenPreview=1` / `screenshotOpenWallPreview=1` deep-link params make every pill state capturable on the emulator; new i18n keys ×4 locales per the glossaries (muro, allumer, Board beleuchten / An der Wand split per the German lighting rule), `playView.previewBadge` retired.

Design synthesized from a three-proposal / three-adversarial-judge panel run against the Velvet Send guidelines, then a two-lens Fable review of the implementation (all findings fixed, none rejected).

The `shouldArmBusyWallConfirm` predicate and four A2-only i18n keys land now (table-tested, `i18n-keep`-marked) so slice A2 wires straight into pinned contracts.

## Release Notes

- The play view now shows at a glance whether the climb on your screen is the one lit on the wall — a pill in the header reads "On the wall", "Live", or "Browsing".
- Previewing a climb got cleaner: a proper "Put on the wall" button replaces the old orange banner, and the board art gets the space back.

## Test plan

- [x] `vp run test:mobile` — 697 files / 6814 tests, 0 failures (includes 60+ new resolver/component/wiring tests: pill ladder tables, pip-suppression ⇔ contract, commit-label rule, latch-truthfulness agreement test against `getSwipeNavigationTarget`, announcer transition table)
- [x] `vp test run --project i18n` — catalog parity across all four locales
- [x] `vp check`
- [x] Android emulator screenshots (Material, dark): solo baseline, Browsing pill + viewfinder brackets, on-wall avatar pill — captured and shared with Marco (commit-row state verified by tests; visible on the pr-4666 OTA channel by previewing a non-current climb)
- [ ] Real-board pass (needs a wall): mirror/playback while previewing, BLE auto-send unchanged — `.boardsesh/qa-notes.md` § A1


marcodejongh:
Re the setSetting note: intentional, and confined to screenshot builds — the EXPO_PUBLIC_SCREENSHOT_MODE check is inlined per screenshot-mode.ts's dead-strip rule, so production bundles never contain the call. Within a screenshot session the persistence is the point (the browse chrome must reflect the same setting the swipe handlers read, and a transient override would need a parallel plumbing path just for captures). It does mean capture order matters on a warm emulator profile — live-state shots before preview shots — which the capture notes account for.

marcodejongh:
All three Codex findings are real — good catches. Back to draft until they're fixed on this branch:

- **P1 (playback BLE writer)**: confirmed — `useMobilePlayback` receives the displayed climb, so opening a settings-preview of an animatable climb with BLE connected writes the preview's frames to the wall while the pill says Browsing. This gating was scoped to the A2 slice, but the pill's whole premise is that it never lies, so it moves into this PR (mirror-toggle gating comes with it).
- **P2 (mirror not reset on Back to live)**: real asymmetry — the new exit path skips the `setIsMirrored(false)` reset every other navigation does.
- **P3 (TalkBack reaches below-fold behind the callout)**: real — the Android trap covers `firstScreenContent` but not the `DeferredSections` sibling.

Fixes land shortly (the A2 branch is mid-implementation in the same files; sequencing the A1 fix first, then rebasing A2).

marcodejongh:
All three findings fixed in 121d7ca71:

- **P1**: `useMobilePlayback` gains `suppressWallWrites` (passed `isPreview`) — preview playback animates on-screen only; the write loop is gated before the drain and resumes when the preview clears (hook test covers suppress → resume with the current frame). The mirror toggle gets the same `!isPreview` gate on its re-push.
- **P2**: `handleBackToLive` now resets `setIsMirrored(false)`, matching every other navigation path.
- **P3**: the below-fold `DeferredSections` sibling joins the callout's Android trap (`accessibilityElementsHidden` + `no-hide-descendants` while the callout is open).

Full mobile suite green (697 files / 6819 tests), typecheck + lint clean. Marking ready again once CI confirms.

marcodejongh:
Both P2s fixed in 2e05f7d9e: the below-fold wrapper goes `pointerEvents: none` while the callout is open (the peeking Logbook header was indeed tappable outside the scrim), and Back to live closes the callout explicitly instead of trusting the stale-state effect (previewing the committed climb itself changed neither dependency).

PR #4669, opened 2026-08-22
ci(web): actually run the REST-surface oracles on the PR that breaks them

marcodejongh:
## Summary

Follow-up to #4663. That PR added two oracles meant to keep the REST surface honest, but neither could ever be selected by CI on the pull request that would break them — and one of the four documents it de-staled kept a section pointing at a file it deleted.

**1. Both oracles were inert on PRs.** `test-default` runs `vp test run --changed "origin/$BASE_REF"` (ci.yml:604). `rest-surface-inventory.test.ts` imports nothing from the repo — it walks `app/api/**` with `readdirSync` — so Vitest's module graph never relates a route add or delete to it. `openapi-document.test.ts` is statically related to the *registry*, so `--changed` selects it when a registration changes, but not when only a route file moves, which is exactly the diff its new `never advertises a registered path whose route file is gone` check was written for.

Measured on the merged #4663 tree, working tree otherwise clean:

| probe | selected specs |
| --- | --- |
| add `app/api/internal/probe-route/route.ts` | `No test files found, exiting with code 0` |
| delete `app/api/og/setter/route.tsx` | `No test files found, exiting with code 0` |
| delete `app/api/v1/[board_name]/slugs/layout/[slug]` (OpenAPI-registered) | `No test files found, exiting with code 0` |
| control: append a comment to `app/lib/url-utils.ts` | 53 files, 731 tests |

The oracles themselves bite fine — force-selected, the third probe gives `× never advertises a registered path whose route file is gone`. It was CI's *selection* that was blind, so every guarantee in their docstrings only held post-merge on `main`.

The fix follows the pattern ci.yml already uses for its other fs-reading oracles (`deploy-config` at 479–513, `listing-guards` at 515–539): a small dedicated `rest-surface` job that is **not** `--changed`-filtered, gated on a `restSurface` path filter covering `packages/web/app/api/**`, `vercel.json`, `app/lib/api-docs/**`, and the specs themselves. `scripts/__tests__/ci-rest-surface-workflow.test.ts` pins that job's shape — the job runs it as its own second step, because a spec that reads `ci.yml` has the same blind spot one level up.

With the job in place, the same three probes red:

| probe | result |
| --- | --- |
| add a route file | `classifies every route file on disk`, `counts exactly the audited surface` |
| delete an unregistered route | `never classifies a route file that no longer exists`, `counts exactly the audited surface` |
| delete a registered route | those two plus `never advertises a registered path whose route file is gone` |

**2. The verdict taxonomy contradicted itself.** The docstring defines `keep-external` as a published surface with no in-repo runtime caller by design, naming "a Vercel cron target" as an example — yet all four cron targets shipped as `keep-caller`, leaving that clause with zero members. Grepping non-test, non-`.next` callers for `cleanup`, `prewarm-heatmap`, `profile-percentiles` and `refresh-sitemap-climbs` returns only `vercel.json`, their own route tests, and `docs/sitemap.md` — no in-repo runtime caller, which is the file's own definition of `keep-external`. Relabelled all four, and added a check that derives the expected set from `vercel.json` (treating `[param]` as a wildcard) rather than pinning literals, so a future mislabel isn't silent. Nothing asserted verdict *values* before this beyond the two pinned externals.

**3. `packages/backend/MIGRATION_PLAN.md` still documented the deleted parity suite.** Its `### Parity Tests` section told the reader to run `--config vitest.parity.config.ts` (deleted by #4663) and asserted the suite was excluded from CI via a `*parity*.test.ts` glob (removed from `packages/backend/vite.config.ts` by #4663). Replaced with a note saying it was removed and why.

Production targets: web/CI only. No mobile package, no `packages/shared/**`, no `bun.lock` — no native fingerprint change and no OTA.

Refs #4663, #1889, #4358.

## Release Notes

## Test plan

- `vp check` — format clean across all 4878 files. The three lint errors are the known local-install artefacts CI does not have (`@gorhom/bottom-sheet` ×2 from the nested `packages/mobile/web-runtime` install, `canvas` from moonboard-ocr); none of the changed files appear anywhere in the output.
- `vp run typecheck`, `vp test run --project web`, `--project scripts`, `--project i18n`, `vp run check:i18n`, `vp run check:i18n:orphans`.
- Mutation-tested the new cron oracle, four directions, each reverted from a pristine copy: relabel a cron route `keep-caller` → red; schedule a cron for a route that doesn't exist → red; schedule a cron for a real `keep-caller` route → red; empty the `crons` array → red.
- Mutation-tested `ci-rest-surface-workflow.test.ts`, seven directions: add `--changed` to the run step → red; drop `rootCi` from the gate → red; drop `packages/web/app/api/**` from the filter → red; drop the `restSurface` output line → red; drop `- rest-surface` from `ci-status` → red; drop the OpenAPI spec from the run line → red; drop the contract-spec step itself → red. The spec strips `#` comment lines first, so the job's own rationale can't satisfy an assertion.
- End-to-end: ran the exact command the new job runs against each of the three route probes above and confirmed it reds, then confirmed a clean tree is green.
- The job fails closed if either spec is renamed without updating the workflow: `vp test run --project web <missing path>` exits 1, it does not quietly run nothing.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -52,6 +52,15 @@ jobs:
       # `listing-guards` job (see below).
       storeMetadata: ${{ github.event_name != 'pull_request' && 'true' || steps.filter.outputs.storeMetadata }}
       mobileLocales: ${{ github.event_name != 'pull_request' && 'true' || steps.filter.outputs.mobileLocales }}
+      # The two REST-surface oracles walk packages/web/app/api/** and
+      # packages/web/vercel.json with readdirSync/readFileSync at run time, so
+      # Vitest's `--changed` module-graph analysis can never relate them to a
+      # diff that only adds or deletes a route file — the exact diff shape they
+      # exist to catch. Measured on #4663: adding app/api/internal/probe-route/
+      # route.ts, and separately deleting app/api/og/setter/route.tsx, each
+      # selected ZERO specs under `vp test run --project web --changed`. Gates
+      # the `rest-surface` job (see below).
+      restSurface: ${{ github.event_name != 'pull_request' && 'true' || steps.filter.outputs.restSurface }}
       # PR-only guard signal: did this PR edit the OTA-owned changelog data file?
       # Defaults false off PRs (the OTA bot legitimately rewrites it on push).
       changelogData: ${{ github.event_name == 'pull_request' && steps.filter.outputs.changelogData || 'false' }}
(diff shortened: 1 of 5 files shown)


marcodejongh:
On the catch-all note: a cron path pointing at a `[...slug]` route would fail the segment-length check, `routeKeyForCronPath` would return `undefined`, and the verdict becomes `no-route-file` — so the test reds rather than silently passing. That is the fail-closed direction, and the mutation probe for it (`schedule a cron for a route that doesn't exist`) already covers the same code path. Leaving the matcher as-is; a real catch-all cron would announce itself loudly.

Issue #4664, opened 2026-08-22
board-renderer WASM holds 3 full RGBA copies per render — linear memory ratchets ~32MB per instance


marcodejongh:
See above

Issue #4650, opened 2026-08-21
Vercel Fluid CPU is burning ~42 h/day three days into the cycle — find it before it eats the month


marcodejongh:
Part of #4648 (Phase 0).

## The numbers

Three days into the Aug 18 – Sep 17 billing cycle (measured 2026-08-21):

- Fluid Active CPU: **127 h** ($16.56) — ~42 h/day, i.e. ~1.8 vCPU busy around the clock
- Function invocations: **1.8 M** — ~7 requests/second sustained
- Fast Origin Transfer: **101 GB**
- Observability events: **6.89 M** ($8.27)

That's for a site that no longer serves the climbing UI (post-#4358 www is marketing, account, gym and SEO surfaces plus `/api/v1`). Marco: "something's wrong if they are that high already." If nothing changes the cycle lands at $300–400.

## How to get the per-route data

The Vercel MCP server has no metrics tool (only logs/errors/analytics), but the **`vercel metrics` CLI** queries Observability directly and groups by dimension. Observability Plus is required for function metrics — the $8.27 events line says the paid ingest is on, so **pull the data before Phase 0 turns that ingest down**:

```bash
vc metrics list                                   # discover metric ids available to the team
vc metrics schema <function-cpu-metric>           # confirm dimensions (route, etc.)
vc metrics <function-cpu-metric> --project <web> --prod --since 3d \
  --group-by <route-dimension> --order-by value --limit 25 --format json
# repeat for invocations + duration; CPU-per-invocation = cpu_time / invocations per route
```

Also worth one pass: `vc metrics` on edge requests grouped by route (what's actually being crawled), and the Observability dashboard's Functions tab as a cross-check.

## Suspects, in order

1. **Sitemap shards** — #4552 measured `/sitemaps/climbs/N.xml` at 51 s cold; #4523/#4524/#4583 document cold-cache misses and retries. Crawlers walking the sitemap tree would be exactly this shape of burn.
2. **Bot traffic on dynamic SSR pages missing the edge cache** — caching depends on `Vercel-CDN-Cache-Control` s-maxage set in `middleware.ts:236-251`; check the actual cache hit rate. 7 req/s with ~34 GB/day origin transfer smells like heavy crawling.
3. The 300 s routes (`prewarm-heatmap` ×5, `profile-percentiles`) — weekly schedules, shouldn't reach 42 h/day, but rule them out.

## Output

[...]

PR #4675, opened 2026-08-22
fix(web): stop /api/internal/board-render OOM-killing its instances

marcodejongh:
## Summary

`/api/internal/board-render` OOM-killed its Vercel instances **507 times in three days** — 61% of all error volume — at roughly 50k renders/day. Three things were stacked on top of each other:

1. **Per-request decode → encode → decode waste.** Every request loaded each board photo, resized it, re-encoded it to WebP, then handed the first buffer straight back to sharp to be decoded again for compositing. `dim_background` then allocated and PNG-encoded a full W×H×4 black scrim, and the OG variant rasterised a blurred 1200×630 SVG per request. On the tallest Kilter board (1080×2498) that is several 10 MB planes per request, none of them reusable.
2. **No caching at all.** The long-running backend OG renderer has cached its composed base and its final bytes in a `BoundedLru` pair since #3829; the web route cached nothing, so every climb on a board re-did that board's photos.
3. **Untuned libvips + unbounded concurrency.** sharp's defaults (50 MB operation cache, one worker thread per core) assume a long-lived server with the machine to itself, and nothing limited how many renders ran at once — so peak memory tracked the arrival rate, and a traffic spike killed the instance instead of slowing it down.

What changed:

- **`composeBoardBaseBuffer`** decodes layers one at a time and folds them as raw RGBA, so peak stays around three planes regardless of layer count and nothing is re-encoded. The OG variant reuses the cached `composeOgBaseBuffer` + `encodeOgImage` pair the backend already uses.
- **Three module-level caches in the route**: composed board base (64 MB / 24), OG base (32 MB / 8), final encoded bytes (32 MB / 2000, keyed on the full canonical param tuple). Plus in-flight coalescing at two levels — one render per identical request, and one *base compose* per board, so two climbs on the same board arriving together don't each fold that board's ~10 MB of photos.
- **A FIFO semaphore** (`BOARD_RENDER_CONCURRENCY`, default 2) around the WASM render and the sharp pipeline, `sharp.cache({ memory: 16, files: 4, items: 100 })` + `sharp.concurrency(1)` at the route (the backend keeps libvips defaults), and a 3 MP output guard — the largest real board is 1080×2498 ≈ 2.70 MP, so anything larger is hand-crafted and gets a 400 rather than being resampled. A data-driven test fails if a future board grows past that ceiling, so it surfaces as a red test rather than a production 400.
- **Load shedding**: past 40 queued renders, new work gets a fast `503` + `Retry-After: 5` + `no-store` instead of a slot in a queue it can't clear inside the 30s budget. A request that can join an in-flight render is never shed — it costs nothing to serve.
- **`vercel.json`**: 3009 MB / 30s for this function as the stopgap that stops the bleeding while the above lands.

`Server-Timing` now carries `cache;desc=hit|base-hit|miss|none` and `queue;dur=<ms>`, so the hit rate and queue depth are readable straight off production traces.

Measured by QA: roughly **2× lighter and faster** on the render path.

Part of #4650 (Vercel burn hunt). The Rust follow-up that moves this render off Node entirely is #4664 — this is the fix that stops the kills today, not a replacement for it.

## Byte-level changes

Output bytes are not identical to `main` everywhere. Every affected URL is immutable and content-addressed by its params, so an old and a new copy simply coexist in the CDN.

- **`dim_background` output is unchanged — deliberately.** An earlier revision replaced the black scrim with an RGB scale (`rgb × (1 - dim)`), which is equivalent only for opaque pixels. QA measured every board WebP: they all carry alpha, and the photos are transparent everywhere except the holds, so ~77% of pixels came out with different alpha and up to 127 of RGB delta. `dim_background` is a full-bleed wash — that is what the iOS Live Activity widget asks for with `dim_background=0.18`, and what mobile's `LayeredClimbImage` draws as a full-bleed `rgba(0, 0, 0, 0.18)` view. The scrim composite is restored; the saving is kept by baking it into the *cached* base, so a board pays for it once on a cold cache instead of once per request.
- **Non-dim background renders change slightly.** The old path pushed every background layer through a lossy WebP q80 encode before compositing; the new one composites the decoded pixels directly. The new bytes are the more faithful render — the old ones carried an extra generation of WebP loss.
- **Two OG param combos with no live caller changed.** `variant=og&format=webp` used to silently return PNG and now returns WebP; `variant=og&format=jpeg` now encodes with the OG JPEG options (mozjpeg q85, full chroma) instead of the generic q90. Both fall out of reusing the backend's `encodeOgImage`. `variant=og&thumbnail=1` now composes full-size photos (it always did — only its cache key wrongly said otherwise).

## Release Notes

## Test plan

- `vp run typecheck` — passes (46 tasks).
- `vp test run --project board-render` — **76 passed / 8 files**, including three new suites:
  - `pipeline.test.ts` drives **real sharp** over 8×8 fixture PNGs: the base is exactly `w*h*4` bytes, layers fold in order, a corrupt layer is skipped, nothing resolving returns `null`, a cached base makes the second render byte-identical with one filesystem probe total, two concurrent misses on one board compose the base once (and compose twice when the in-flight map is not supplied), and the OG path decodes to 1200×630 and reuses its base. Dim is pinned at full-bleed semantics: a **fully transparent** source pixel comes out at `alpha = round(0.18 × 255) = 46` (exactly what the rejected linear approach failed), while an opaque one darkens by ×0.82.
  - `semaphore.test.ts` pins the high-water mark under 10 concurrent runs, FIFO order, and slot release on a throwing task.
  - `board-dimensions.test.ts` is the ceiling tripwire: `max(width × height)` over `BOARD_IMAGE_DIMENSIONS` must stay under `MAX_RENDER_OUTPUT_PIXELS`.
- `packages/web/app/api/internal/board-render/__tests__/route.test.ts` — **56 passed**, extended with a byte-cache key pair for every param that changes a pixel (`board_name`, `layout_id`, `size_id`, `set_ids`, `frames`, `format`, `thumbnail`, `include_background`, `dim_background`, `variant`), each asserting a genuine second render; the counterpart case that a param the route never reads (`angle`) is *served* from the cache; the concurrency limit (3 concurrent requests, gated renders, high-water mark 1); load shedding (42 queued → 43rd gets 503 + `Retry-After: 5` + `no-store`; a coalescable request is never shed); in-flight coalescing; the queue timing header; the 3 MP 400; and a 1080×2498 board still rendering.
- **Mutation-tested every guard**, each failing exactly its own test(s) and nothing else: stripping the five render-option params from the byte key (5 cases red), dropping just `thumbnail` (1 case red), dropping the four identity params (exactly those 4 red), `createSemaphore(100000)`, `MAX_QUEUED_RENDERS = 100000`, `MAX_RENDER_OUTPUT_PIXELS = 2_000_000` (tripwire names the offending board), disabling `byteCache.set`, bypassing the base-compose coalescing map, and reverting the dim scrim to the linear scale. The queue-ceiling mutation also exposed a test-isolation hazard — a failing gated test stranded its renders and timed out every test after it — now fixed with an `afterEach` that releases the gate and drains the queue, so a failure stays local to its own test.
- `vp run test:web` — 4023 passed / 9 failed. All nine are pre-existing flakes in admin/auth/gym component suites (`location-sync-freezes-panel`, `gym-owner-reassign-panel`, `overview-tab`, `auth-modal-server-errors`, …); re-running them in isolation on this branch gives a different failure set each time (1, then 4, same code), none import anything this PR touches, and CI's `test-default` shards are green.
- `packages/web/scripts/board-render-memory-harness.mjs` is the manual RSS harness for a before/after against `main` (needs a running dev server; not wired into CI).

https://claude.ai/code/session_016aZ1zjpqkXScoo4jyzL49X

### DIFF
--- a/packages/shared/board-render/src/__tests__/board-dimensions.test.ts
+++ b/packages/shared/board-render/src/__tests__/board-dimensions.test.ts
@@ -1,0 +1,37 @@
+import { BOARD_IMAGE_DIMENSIONS } from '@boardsesh/board-config';
+import { describe, expect, it } from 'vitest';
+import { MAX_RENDER_OUTPUT_PIXELS } from '../validation';
+
+/**
+ * Tripwire for the render route's pixel ceiling.
+ *
+ * A full-size render is the board image's own dimensions, so the ceiling has to
+ * stay above the largest board art we ship. If a future board is taller than
+ * this allows, that must surface as a red test here — not as a 400 for every
+ * request for that board in production.
+ */
+describe('board art fits under the render pixel ceiling', () => {
+  const boardImages = Object.entries(BOARD_IMAGE_DIMENSIONS).flatMap(([boardName, images]) =>
+    Object.entries(images).map(([imageName, { width, height }]) => ({
+      label: `${boardName}/${imageName}`,
+      pixels: width * height,
+      width,
+      height,
+    })),
+  );
+
+  it('covers every board image', () => {
+    expect(boardImages.length).toBeGreaterThan(100);
+  });
+
+  it('leaves the largest board under MAX_RENDER_OUTPUT_PIXELS', () => {
+    const largest = boardImages.reduce((worst, image) => (image.pixels > worst.pixels ? image : worst));
+
+    expect(
+      largest.pixels,
+      `${largest.label} renders at ${largest.width}×${largest.height} (${largest.pixels} px), over the ` +
+        `${MAX_RENDER_OUTPUT_PIXELS} px ceiling in validation.ts. Raise the ceiling (and the route's memory ` +
+        'budget with it) rather than letting this board 400 in production.',
+    ).toBeLessThan(MAX_RENDER_OUTPUT_PIXELS);
+  });
+});
--- a/packages/shared/board-render/src/__tests__/lru.test.ts
+++ b/packages/shared/board-render/src/__tests__/lru.test.ts
@@ -64,6 +64,23 @@ describe('BoundedLru', () => {
     expect(cache.get('huge')).toBeUndefined();
   });
 
+  it('drops every entry and its byte total on clear', () => {
+    const cache = byteLru(3, 100);
+    cache.set('first', Buffer.alloc(10));
+    cache.set('second', Buffer.alloc(10));
+
+    cache.clear();
+
+    expect(cache.size).toBe(0);
+    expect(cache.byteSize).toBe(0);
+    expect(cache.get('first')).toBeUndefined();
+
+    // Still usable afterwards.
+    cache.set('third', Buffer.alloc(10));
+    expect(cache.size).toBe(1);
+    expect(cache.byteSize).toBe(10);
+  });
+
   it('returns undefined for missing keys and tracks sizes accurately across evictions', () => {
     const cache = byteLru(3, 100);
     expect(cache.get('missing')).toBeUndefined();
--- a/packages/shared/board-render/src/__tests__/pipeline.test.ts
+++ b/packages/shared/board-render/src/__tests__/pipeline.test.ts
@@ -1,0 +1,275 @@
+import { mkdtempSync, rmSync, writeFileSync } from 'node:fs';
+import { tmpdir } from 'node:os';
+import { join } from 'node:path';
+import sharp from 'sharp';
+import { afterAll, beforeEach, describe, expect, it } from 'vitest';
+import { BoundedLru } from '../lru';
+import { composeBoardBaseBuffer, renderBoardImageBuffer, type OgBaseResult, type ResolveImagePath } from '../pipeline';
+import type { RenderableBoardDetails } from '../types';
+
+const SIZE = 8;
+const PIXELS = SIZE * SIZE;
+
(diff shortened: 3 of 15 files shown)


marcodejongh:
Re the last review pass, for anyone reading the ⚠️ before the checks finish:

**`angle` is not missing from the byte-cache key — this route never reads it.** `GET` reads exactly ten query params (`board_name`, `layout_id`, `size_id`, `set_ids`, `frames`, `thumbnail`, `include_background`, `variant`, `format`, `dim_background`); `angle` appears nowhere in the route, in `buildRenderConfig`, or in `WasmRenderConfig`. Rather than leave that as prose, 695fb146 pins it: a request that adds `angle` is served the cached bytes, and the day `angle` (or any new param) starts changing pixels that test goes red and the param has to join the key.

The other two points from that pass are addressed or deliberate: the byte-key matrix now covers `board_name`, `layout_id`, `size_id`, `set_ids` and `frames` as well (verified by dropping the four identity params from the key — exactly those four cases go red), and `resetBoardRenderCaches` stays an ungated export. It only clears in-memory caches, it has no production caller, and an App Router route module may not export anything but handlers and route config — which is why the caches live in a sibling module at all.

PR #4676, opened 2026-08-22
feat(web): name which path served the climb sitemap, and page when it is the slow one

marcodejongh:
## Summary

**Read this first: the tier-2 summary table this branch was originally opened for is already in production.** #4661 merged at 2026-08-22T00:23Z and #4583 was closed `COMPLETED` fifteen minutes later, verified against production. The work I was handed (`sitemap_tier2_groups` + `sitemap_tier2_climbs`, a nightly GH Actions cron, a predicate fingerprint, migration 0204) was cut from `c69611d0b` — the commit immediately *before* main landed its own materialisation at the same migration number. Two of the review findings say so directly, and both are right. Full accounting in the "What happened to the original branch" section below.

So this PR is not that branch. It is the one thing #4661 left open from the #4583 ruling's own scope notes:

> Keep the existing live-query path as a fallback, and **make a fallback loud rather than silent** — the whole lesson of this issue is that a quiet degrade hid a real regression.

Today that fallback is a `console.error`.

### The gap

The climbs shard reads a materialised store and falls back to the live scan when the store is empty or unreadable. The fallback is *correct* — the tables are a cache, truncating them loses nothing — and that is exactly the problem. It still produces a valid, complete sitemap, so:

- `X-Sitemap-Degraded` stays absent (the index published every shard);
- `REQUIRED_SITEMAP_SHARDS` stays satisfied (page 1's `<loc>` is right there);
- the smoke never fetches a climb page at all, so the **page** path — the 51 s rebuild, once per page per cold lambda — has no external signal of any kind.

That is the shape #4583 lived in from the day W-23 landed until #4661: the largest surface on the site, absent from the index on every request, for weeks, because the only trace was a log line.

### What this adds

**1. `X-Sitemap-Climbs-Source: store | live`**, on `/sitemap.xml` and on every `/sitemaps/climbs/N.xml`.

Derived from the `source` the summary and page builds have *already returned* — never a fresh read. That constraint is load-bearing and is the whole reason the earlier draft's version of this header was a major finding: the index races each summary against `SHARD_DEADLINE_MS`, and `withDeadline` is a `Promise.race` that cannot cancel the loser. A diagnostic that issued its own query after the index body was built would be free to outlive the deadline it describes and hold the whole response open to `maxDuration = 300`. Threading the value back through `buildPagedIndexEntries` makes that impossible by construction: there is no second await to run long.

Two smaller calls inside it:
- a rejected or timed-out summary reports **no** source rather than guessing `live` — it never told us which path it was on, and `X-Sitemap-Degraded` already covers that case;
- on a page, the **page build's** answer wins over the summary's. A fresh summary row over an empty URL table is not hypothetical — it is exactly the state the deploy that added `sitemap_climb_urls` was in, and a header taken from the summary alone would have called it healthy.

**2. A Sentry event** from `reportStoreFallback`, at most once per reason per instance per six hours (`summary-empty`, `summary-read-failed`, `page-empty`, `page-read-failed`). The log still fires every time; the floor is only on paging someone, because a cold crawl is the index plus one request per page arriving together and one event per request buries its own signal. The page-path event fires **before** the fallback build, not after — that build is the slow path by definition, and a platform timeout inside it would swallow an event placed on the far side.

**3. The smoke reads the header** off the index response it already fetches — no extra request, no 2 MB page download. `live` is a **warning**, not a failure: the index it just validated is complete either way, so this is "the fast path is not the one running", not "the sitemap is broken". A missing header is not a finding.

**4. `docs/sitemap.md`** gets the header contract, the Sentry reasons, and the one honest limit — the header rides a CDN-cached response (`s-maxage=3600` on the index, `21600` on pages, and Vercel's edge ignores the smoke's `no-cache` request header), so the value read can be up to that old in either direction. It is "a wedged store gets noticed within the hour", not a live probe. The Sentry event has no such lag.

## Production measurements, taken today

Read-only, authorized. Stated as measured rather than carried forward.

| probe | result |
|---|---|
| `sitemap_climb_urls` | 53,230 rows, `max(ordinal)` 53,229 |
| `sitemap_shard_refreshes` (`climbs`) | `item_count` 53,230, `computed_at` 2026-08-22T00:36Z |
| `/sitemaps/climbs/1.xml`, cache-busted | 200, 2,247,529 bytes, **4.95 s** origin miss |
| `/sitemaps/climbs/3.xml` | 200, 2,025,415 bytes, 1.72 s |
| `/sitemaps/climbs/6.xml` | 200, 677,932 bytes, 1.43 s |
| `/sitemap.xml`, 3 cache-busted requests | **1 of 3 still returned `x-sitemap-degraded: climbs`, at 4.74 s**; the other two 1.34 s / 1.23 s clean |

The store is healthy and #4661's win is real. But that last row is the point of this PR in one line: the index *still* degrades on a genuinely cold lambda roughly a third of the time, and **from outside there is no way to tell whether that was a cold connection over a healthy store or the store being empty.** With this change it is one header.

The earlier draft's "27.4 s cold page" figure is not carried forward — it is no longer production's cost, and a review finding was right to flag it.

## What happened to the original branch

`feat/4583-tier2-summary-table` (`85698f298`, +15,366 lines) is superseded and is not being pushed. Fourteen findings came back across two review lenses; here is where each landed.

**Stands, and is why the branch is abandoned rather than fixed:**

- **[major] page slices torn across a 5-minute cached summary.** The branch computed every `(group, offset, limit)` triple from a cached group summary and then read the rows live, with no epoch column tying the two. The lens reproduced it on the dev image: a mid-window refresh silently dropped 1,500 URLs off page 3, and a larger shrink 503'd the shard. `main` cannot have this bug — `fetchStoredClimbPage` reads `WHERE ordinal >= start AND ordinal < start + 10000`, one statement, self-consistent by construction. The branch replaced a stored ordinal with a derived offset and regressed the exact property it claimed to add.
- **[major] the diagnostic header defeated the index deadline.** `Object.assign(shardHeaders, await safeShardHeaders(shard))` ran after `buildSitemapIndexXml()` returned, outside `withDeadline`, on a promise (`verdictInFlight`) that a lost 3 s race leaves pending forever. Verified: `withDeadline` is a bare `Promise.race`, `safeShardHeaders` had a try/catch and no timeout, and the route sets `maxDuration = 300`. **This PR's header is the corrected shape** — the value comes back with the entries, so there is no second await to run long.
- **[major] per-page `<lastmod>` was gone.** Not a deletion by the branch (it predates #4661) but a real regression once rebased, and the comment justifying its absence was factually wrong: `main` gets it from one `max(last_modified) GROUP BY (ordinal / 10000)` aggregate, no item build. Rebasing onto `main` keeps it.
- **[minor] `docs/sitemap.md` untouched** while the branch added two tables, a cron, four headers and a Sentry channel. Verified: `git diff --stat` over `docs/` was empty. **Fixed here** — `main` documented its own store, and this PR documents what it adds on top.
- **[minor] `climbs` flipped to `degradable: false`** in the same change that shipped empty tables. Not carried over; `main`'s `degradable: true` is kept, and the new source check is a warning for the same reason.
- **[minor] the 24 h cron staleness was documented in one direction only** (a climb *entering* tier 2), never the delist / draft / angle-flip direction. Moot: `main`'s cron is six-hourly and matches the shard's own `s-maxage=21600`, so no layer is staler than another.
- **[minor] a failed `workflow_dispatch` notified nobody** — the Discord step was gated on `github.event_name == 'schedule'`, and the header told operators to dispatch by hand right after the migration. Moot with the workflow gone; the "a silent fallback reaches nobody" half of that finding is what this PR fixes.
- **[minor] `X-Sitemap-Tier2-Drift` was dead in a serverless fleet** — per-instance module state written in `after()` and read on a later request to the *same* lambda, behind an hour of CDN cache. Correct, and it is why this PR's header carries only values measured **in the same request**, and why the drift detector is not reintroduced.
- **[minor] the claimed benchmark numbers did not reproduce** (0.13 ms summary measured at 1.10–2 ms; 6.7 ms page measured at 20–75 ms SQL-only). Not carried forward; the table above is what I measured today.
- **[minor] `MIN_EXPECTED_TIER2_ITEMS = 40,000` had 3.2% headroom** on the dev image with no override. Moot.
- **[nit] SQL `ORDER BY` vs the JS comparator** could disagree under `en_US.utf8` for a board type containing punctuation. Moot.
- **[nit] `lastAuditWarnings` never cleared on the early-return path**, so a stale drift value kept being emitted after the condition cleared. Moot.

**Verified and stands, worth recording because it is a general trap:**

- **[minor] literal NUL bytes** in `packages/db/src/queries/sitemap/tier2-fingerprint.ts` — `hash.update('\0')` written with a raw `0x00` instead of the escape, at offsets `0x86b` and `0x8ab`. Confirmed: `file` reports `data` and `git diff --stat` renders it `Bin 0 -> 2271 bytes`, so the one file defining the drift detector could not be reviewed in a diff. Moot here, but worth a wider look: if any other source file in this repo carries a raw NUL it is silently unreviewable the same way.

**Rebutted:** none. Every finding held up. Two ("per-page lastmod deleted", "no ordinal") describe a divergence from a `main` that did not exist when the branch was cut rather than something the branch removed, but the conclusion each draws — do not ship this shape — is correct either way.

## Test plan

- `vp run typecheck` — clean.
- `vp test run --reporter=agent --project web` — 4,030 passed. 3 failures, all pre-existing flakes under full parallel load in unrelated component tests (`location-sync-freezes-panel`, `profile-tab`, `claim-gym-dialog`); `location-sync-freezes-panel` also fails on `origin/main` in isolation, and all three pass in isolation on this branch. None import sitemap code.
- `vp test run --reporter=agent scripts/production-smoke.test.ts` — 21 passed.
- `vp check` — 2 errors, both pre-existing `@gorhom/bottom-sheet` TS2307s from the mobile web-shim's nested install, unrelated.
- **Mutation-tested, every one applied and watched go red, then reverted:**
  - report moved to *after* the fallback build → the ordering test fails.
  - report floor removed → the "pages once per reason" test fails.
  - `if (!readFailed)` forced true → the double-report guard fails.
  - fallback labelled `store` → five source assertions fail.
  - `summary.source ?? pageSource` instead of `pageSource ?? summary.source` → the page-overrides-summary test fails.
  - index header guessed as `live` on a rejected summary → the "claims nothing when the summary never settled" test fails.
  - both `sourceHeaders(...)` calls dropped → four header tests fail.
  - smoke's `=== 'live'` widened to `!== 'store'` → two smoke tests fail (an absent header would warn).
  - smoke's reasons joined → last-only → the "both signals at once" test fails.

## Production targets

`packages/web` and `scripts/` only. **Nothing in the mobile graph, no OTA.** No migration, no schema change, no new env var, no new secret, no new cron. One new response header and one new Sentry message channel.

Refs #4583

## Release Notes

The user-visible half of #4583 — tier-2 climbs actually reaching Google — shipped in #4661 and was verified in production there. What is left here is observability: nothing a climber sees, and a "What's New" line about a sitemap response header would be noise.

### DIFF
--- a/docs/sitemap.md
+++ b/docs/sitemap.md
@@ -101,6 +101,44 @@ the 51 s path, correct and never worse than before the store existed. There is n
 staleness bound on pages; the summary row's 48 h shout covers the store as a whole,
 since both tables share one refresh.
 
+### Saying which path served (#4583)
+
+The fallback is correct, and that is exactly the problem: an empty store produces a
+complete, correct sitemap, so nothing external goes red while every crawler fetch
+behind it pays the scan. The climbs shard was absent from the production index on
+every request from the day W-23 landed until #4661, and the reason it stayed there
+for weeks is that a degrade only ever reached a `console.error`.
+
+So a fallback names itself, on two channels:
+
+- **`X-Sitemap-Climbs-Source: store | live`** on `/sitemap.xml` and on every
+  `/sitemaps/climbs/N.xml`. Set from the `source` the summary and page builds have
+  already returned — never a fresh read. That matters: the index races each summary
+  against `SHARD_DEADLINE_MS` and `withDeadline` cannot cancel the loser, so a
+  diagnostic that issued its own query could outlive the deadline it describes and
+  hold the whole index open to `maxDuration`. A rejected or timed-out summary
+  reports no source at all rather than guessing, since `X-Sitemap-Degraded` already
+  covers that case. On a page the build's own answer wins over the summary's: a
+  fresh summary row against an empty URL table is exactly the state the deploy that
+  added the store was in.
+- **A Sentry event**, from `reportStoreFallback` in `climb-store.ts`, at most once
+  per reason per instance per six hours (`summary-empty`, `summary-read-failed`,
+  `page-empty`, `page-read-failed`). The page-path event fires _before_ the 51 s
+  fallback build, so a Vercel timeout cannot swallow it.
+
+`scripts/production-smoke.ts` reads the header off the index response it already
+fetches — no extra request — and reports `live` as a **warning**, not a failure:
+the index it just validated is complete either way, so this is "the fast path is not
+the one running", not "the sitemap is broken". A missing header is not a finding; a
+deploy that predates it and a summary that never settled both leave it absent.
+
+One honest limit, shared with `X-Sitemap-Degraded`: the header rides a CDN-cached
+response. A healthy index is `s-maxage=3600` and the shard pages are
+`s-maxage=21600`, and Vercel's edge ignores the smoke's `Cache-Control: no-cache`
+request header, so the value read can be up to that old in either direction. It is a
+signal that a wedged store gets noticed within the hour, not a live probe. The Sentry
+event has no such lag.
+
 ### Who refreshes it
 
 1. **The cron** — `/api/internal/refresh-sitemap-climbs`, six-hourly, in
--- a/packages/web/app/lib/seo/sitemap/__tests__/climb-shards.test.ts
+++ b/packages/web/app/lib/seo/sitemap/__tests__/climb-shards.test.ts
@@ -42,6 +42,9 @@ const climbs = vi.hoisted(() => ({
   /** What the per-page lastmod aggregate answers; empty means "store empty, use the uniform value". */
   pageLastmods: [] as (Date | null)[],
   pageLastmodsThrow: false,
+  /** Which path the store says answered. `undefined` is a build that reports none. */
+  summarySource: 'store' as 'store' | 'live' | undefined,
+  pageSource: 'store' as 'store' | 'live' | undefined,
 }));
 
 // The index and the shard route both read the store (#4523/#4552): the summary
@@ -56,12 +59,16 @@ vi.mock('../climb-store', () => ({
     if (climbs.summaryThrows) throw new Error('climbs summary unavailable');
     // A summary that never settles: the failure a try/catch cannot see.
     if (climbs.summaryHangs) return new Promise<never>(() => {});
-    return { itemCount: climbs.itemCount, lastModified: new Date('2026-05-04T11:22:33.000Z') };
+    return {
+      itemCount: climbs.itemCount,
+      lastModified: new Date('2026-05-04T11:22:33.000Z'),
+      source: climbs.summarySource,
+    };
   },
   buildClimbShardPage: async (page: number) => {
     climbs.buildCalls += 1;
     if (climbs.buildThrows) throw new Error('climbs builder exploded');
-    if (climbs.buildsEmpty) return { items: [], totalItems: 0 };
+    if (climbs.buildsEmpty) return { items: [], totalItems: 0, source: climbs.pageSource };
     // Padding lives in the PATH, so the byte guard is driven by a real rendered
     // body rather than by mutating the constant it is supposed to enforce.
     const padding = climbs.pathLength > 0 ? 'x'.repeat(climbs.pathLength) : '';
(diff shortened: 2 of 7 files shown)


marcodejongh:
On the review comments:

**Duplicate `'store' | 'live'` union** — fair, and deliberate rather than accidental; I have written the reasoning into the type instead of collapsing it. `shard-registry.ts` imports `climb-store.ts`, so the registry owns the *contract* ("a paged shard may report which path answered") and the store owns its own answer. Unifying means pointing one file at the other, which closes an import loop for two string literals.

**`if (!summary || summary.itemCount === 0)`** — that expression is not in the file. The branch is `if (pageCount === 0)`, reached only after `withDeadline` has resolved `summary`, so there is no `null` case to guard. No change.

PR #4679, opened 2026-08-22
chore(web): drop the e2e ordering pin now the stall is explained

### LINKED ISSUE #4463: E2E: a browser spec poisons the web server for subsequent raw HTTP requests

marcodejongh:
## Summary

Removes the `aa-` filename prefix on the embed-header e2e spec. It was there because Playwright orders spec files alphabetically inside a shard and those raw `APIRequestContext` requests hung for 60 s+ when browser-driven specs had run against the same server first — first-in-shard they passed every time. Nobody knew why, and the file said so.

Stacked on #4668, which fixes the cause.

### What the prefix was hiding

The web server was never the problem. The GraphQL **backend's connection pool** was exhausted by concurrent copies of the home page's uncached `popularBoardConfigs` statement — 82 s each on the dev database, and never cacheable in CI because the e2e backend runs with no `REDIS_URL`. Every spec that loaded `/` added another copy; postgres.js's acquire queue is unbounded and untimed, so every other backend read then waited forever, `/embed/**` renders included. `{ __typename }` kept answering in milliseconds through the same event loop the whole time, which is why three previous theories all looked plausible and all were wrong.

Measured directly: ten parallel `GET /` against the pre-fix backend leaves **eleven** copies of that statement `active` in `pg_stat_activity`, and `board(boardUuid:)` then does not answer inside 30 s.

### Why the prefix had to go rather than stay

`--shard=N/8` re-splits the whole suite whenever a test is added or removed anywhere. The prefix only ever protected one arrangement — it retires itself silently, and the next victim is a different file presenting as an unrelated new flake. It is also the cheapest possible answer to the next ordering problem, and it leaves nothing behind that explains itself.

So the ordering is now driven **explicitly, in-file**: a `describe.serial` block that fans out ten concurrent `/` renders, then does the front-door + climb-view page loads `board-route-teardown.spec.ts` drives, then hits `/embed/board/<fresh uuid>` and `/kiosk/<fresh uuid>` as raw requests. That runs in every shard arrangement instead of one.

Also corrects two comments that still described specs deleted by #4467, and adds a repo guard against re-pinning by filename.

Closes #4463

### The oracle is the status, not the clock

The first version of this block asserted `expect([200, 404]).toContain(status)` plus a 15 s wall-clock budget, and **that combination cannot fail while the SSR deadline exists**. For a fresh uuid both routes do exactly one deadline-wrapped backend fetch, so a starved backend paints the retry screen at 3 s — a 200 the assertion accepted, inside a budget it could not exceed. The table in the first draft of this description showed it plainly and I did not read it that way: `SSR deadline, pre-fix backend` passed. That row is the whole test.

Both requests now assert `404` and nothing else. A fresh uuid is one indexed select on the backend; a backend that can reach the database answers "no such board / no such kiosk" in milliseconds and the route `notFound()`s. **A 200 is by construction the retry screen**, which for these uuids means the fetch hit its 3 s deadline — i.e. the pool was gone. The wall-clock budget stays as a latency backstop and is no longer load-bearing.

## Release Notes

## Test plan

Local 2-CPU-confined production harness — prebuilt `next start` + the backend under `tsx src/index.ts` + a **migrated** `boardsesh-dev-db` container, all pinned to two CPUs, Playwright under `CI=1`, running `e2e/embed-headers.spec.ts` (8 tests: the 7 header checks plus the new serial block). Re-measured from scratch for this revision; these numbers supersede the earlier table.

| backend build | result |
|---|---|
| single-flight (this stack), run 1 | **8 passed** |
| single-flight, run 2 | **8 passed** |
| single-flight, run 3 | **8 passed** |
| pre-#4668 (`git show 25322da86:` for both resolvers, `single-flight.ts` deleted) | **1 failed** — `/embed/board/<uuid>` returned 200, on the attempt **and** the retry |

Same web build, same database, same SSR deadline in both columns, so the single discriminating variable is #4668. The other 7 tests passed in every run.

One methodology note worth recording: the first attempt at the pre-fix column produced a *false* red — the fresh dev-db container was 14 migrations behind, so `user_boards.presence_seq` / `merged_into_board_uuid` did not exist and **every** board and gym query failed from the first test onward, in both columns. `vp run db:migrate` against the harness database, then re-measure. A harness whose baseline column is also red is not a harness.

Guard mutations, each applied, watched red, reverted:
- `packages/web/e2e/1-probe.spec.ts` → `e2e-spec-ordering.test.ts` fails naming it.
- `a-probe.spec.ts` → fails. `aa-probe.spec.ts` → fails. `_probe.spec.ts` → fails.
- Removing all four → green again. Before this revision the first, second and fourth of those were **green**: the pattern only matched a doubled leading letter, so `1-`, `a-` and `_` were all still effective pins. The guard now rejects a leading non-letter, a single letter plus separator, and the doubled letter.
- rename `embed-headers.spec.ts` away → the second assertion fails, so the first cannot pass vacuously.

The guard catches sorting hacks that read as hacks. It cannot catch a plain word that happens to sort first (`aardvark-embed.spec.ts`) — only a committed order manifest could, and that makes every added spec a two-file change. The file says so.

CI: `e2e-tests.yml` dispatched on this branch. The oracle is the **eight `E2E (shard N)` jobs**, not the workflow conclusion — `Expo-web smoke` is a separate task (`vp run test:e2e:expo-web`) that CLAUDE.md already keeps out of per-PR CI and that has failed independently on several recent dispatches. Results in a comment below as each dispatch lands.

### DIFF
--- a/packages/web/e2e/embed-headers.spec.ts
+++ b/packages/web/e2e/embed-headers.spec.ts
@@ -1,18 +1,31 @@
-import { test, expect, type APIResponse } from '@playwright/test';
+import { randomUUID } from 'node:crypto';
+import { test, expect, type APIRequestContext, type APIResponse } from '@playwright/test';
+import { waitForBoardListReady } from './helpers/waits';
 
-// The `aa-` prefix is load-bearing: Playwright orders spec files
-// alphabetically within a shard, and this file's raw APIRequestContext
-// requests hang for 60s+ if certain browser-driven specs run against the
-// same server first — first-in-shard they pass every time, and the whole
-// suite is green. Do not rename this file below its neighbours.
+// This file used to be `aa-embed-headers.spec.ts`. The prefix was load-bearing:
+// Playwright orders spec files alphabetically inside a shard, and these raw
+// APIRequestContext requests hung for 60s+ when browser-driven specs had run
+// against the same server first. #4463 root-caused that and removed it. The
+// short version, so nobody re-derives it:
 //
-// Half of that is fixed: #4463 gave the embed/kiosk SSR backend fetches a 3 s
-// deadline, so a stalled backend now renders the retry screen instead of
-// hanging the request. The other half is not: why the backend stalls on those
-// queries only after browser specs have run is still unexplained, and #4463
-// stays open on it. Nothing about this file is a Playwright state leak — the
-// `request` fixture builds a fresh APIRequestContext per test, and the suite
-// installs no service worker, storageState, or route interception.
+//  - Nothing here was a Playwright state leak. The `request` fixture builds a
+//    fresh APIRequestContext per test, and the suite installs no service
+//    worker, storageState, or route interception.
+//  - The stall was never in the web server. The GraphQL backend's connection
+//    pool was exhausted by concurrent copies of one uncached statement — the
+//    home page's `popularBoardConfigs`, which the CI backend can never cache
+//    because it runs with no REDIS_URL. Every spec that loaded `/` added
+//    another copy. postgres.js's acquire queue is unbounded and untimed, so
+//    every other backend read then waited forever, `/embed/**` renders
+//    included. `{ __typename }` kept answering in milliseconds throughout,
+//    which is why it looked like something subtler than saturation.
+//  - Two fixes closed it: the backend now runs one copy of those cold reads at
+//    a time, and the SSR fetches that wait on the backend carry
+//    SSR_BACKEND_FETCH_TIMEOUT_MS, so a slow backend paints the retry screen —
+//    a 200 these assertions already accept.
+//
+// The ordering dependency is pinned explicitly at the bottom of this file
+// instead of through the filename, so it survives the next spec being added.
 
 // These are raw APIRequestContext gets against a server that is concurrently
 // SSR-ing front-door pages for the other worker's browser tests. On a 2-core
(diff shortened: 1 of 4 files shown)


marcodejongh:
## e2e dispatch log

Oracle: the **eight `E2E (shard N)` jobs**, per the epic's ruling. `Expo-web smoke` fails independently on the same line in every recent run (`bindSeededBoard`, smoke.spec.ts:148) — filed as #4682 — so the workflow conclusion reads `failure` on runs where every shard passed.

| # | run | head | 8 shard jobs |
|---|---|---|---|
| 1 | [32547594570](https://github.com/boardsesh/boardsesh/actions/runs/32547594570) | 767fab6a5 | 8/8 success |
| 2 | [32548232997](https://github.com/boardsesh/boardsesh/actions/runs/32548232997) | 767fab6a5 | 8/8 success |
| 3 | [32548711754](https://github.com/boardsesh/boardsesh/actions/runs/32548711754) | a15715c78 (rebased onto #4668's review fixes) | 8/8 success |
| 4 | [32549203419](https://github.com/boardsesh/boardsesh/actions/runs/32549203419) | 3cf86b8a7 | 8/8 success |
| 5 | [32549687115](https://github.com/boardsesh/boardsesh/actions/runs/32549687115) | 3cf86b8a7 | 8/8 success |

Five dispatches, no shard failure and no passed-on-retry. Un-pinned, the spec lands in shard 6 (verified with `--shard=6/8 --list`): one `board-route-teardown` test, then all eight embed tests, then three locale tests.

One earlier dispatch (32545852780) was cancelled two minutes in, before the guard's preamble was strengthened from sequential `page.goto`s to the parallel fan-out. It produced no result and is not counted.

marcodejongh:
### E2E dispatches with the strict `404` assertion

Both runs are after the serial block stopped accepting a 200 — the change that makes it an oracle for the pool rather than for the SSR deadline.

| run | head | 8 shard jobs | Expo-web smoke |
|---|---|---|---|
| [32552629876](https://github.com/boardsesh/boardsesh/actions/runs/32552629876) | `c8d1a05c3` | **8/8 green** | failure |
| [32553175767](https://github.com/boardsesh/boardsesh/actions/runs/32553175767) | `f2457493d` (current) | **8/8 green** | failure |

`Expo-web smoke` is `vp run test:e2e:expo-web`, which CLAUDE.md already keeps out of per-PR CI; it fails identically on dispatches of this branch that predate every change here, so it is not this stack. Tracking it is a separate job.

The stricter assertion did not cost stability: the shard that carries `embed-headers.spec.ts` passed first try in both runs, and it passed locally three times out of three on a 2-CPU harness while going red on the attempt **and** the retry against the pre-#4668 backend.

marcodejongh:
### E2E dispatch log, final

Four dispatches, all with the strict `404` assertion in place. The last two also carry the backend's pre-invalidation-flight fix (#4668 review round).

| run | head | 8 shard jobs |
|---|---|---|
| [32552629876](https://github.com/boardsesh/boardsesh/actions/runs/32552629876) | `c8d1a05c3` | 8/8 green |
| [32553175767](https://github.com/boardsesh/boardsesh/actions/runs/32553175767) | `f2457493d` | 8/8 green |
| [32554090618](https://github.com/boardsesh/boardsesh/actions/runs/32554090618) | `1bc66ccb8` | 8/8 green |
| [32554589601](https://github.com/boardsesh/boardsesh/actions/runs/32554589601) | `67f51780b` (current) | 8/8 green |

`Expo-web smoke` fails in all four, as it does on dispatches of this branch that predate every change here. It is `vp run test:e2e:expo-web`, which CLAUDE.md keeps out of per-PR CI; tracking it is separate work.

PR #4683, opened 2026-08-22
feat(mobile): browsing is free in a shared session — the wall changes only on Put on the wall

### LINKED ISSUE #4281: Browsing climbs during a session adds them to the queue and lights the board — need a preview/browse mode
When multiple people share a session and someone swipes through the climb list looking for their next project, every climb they scroll past gets added to the session queue and the board LEDs switch to that climb. There is no way to browse without affecting the active session.

> "When multiple people join a session, it is very easy to switch projects / routes. But when someone is swiping through the problems to look for a next project, each route he swipes gets added to the session list, also does the led layout switch."

The ask is a way to browse/preview climbs without committing them to the queue — either a dedicated browse/preview mode, or requiring an explicit tap to add rather than triggering on swipe.

The maintainer acknowledged this is a known UX gap: "Maybe when in a session the default should be preview mode? We discussed it before, it just hasn't happened yet." The long-press menu was noted as an existing partial workaround (renders a board preview without adding to queue).

Related UX note from the same session: the flag/checkered-flag icon used to end a session is unintuitive to new users — a clearer "End session" label or icon would reduce confusion.

## Source

[Reported in #general on Discord](https://discord.com/channels/1288485679663415346/1288485680149827749/1527657099683958866) — flagged in a thread.

Reporter: `discord-c14d0b4d2f75` · Posted: 2026-07-17T12:43:55.061000+00:00

## Attachments

attachment 1

marcodejongh:
## Summary

The behaviour slice of #4281, stacked on #4666 (base = `feat/4281-preview-first-sessions`; retarget to `main` when A1 merges). Closes #4281.

**In a session with 2+ climbers, browsing is free.** Swipes, climbs-list taps (with the result list seeded as the view-only suggestion track), similar-climb taps, and fresh drawer opens from the feed, logbook, profile, and playlists all pin a preview instead of writing the shared queue and lighting the wall. The catalogue is yours; the queue is everyone's — queue-sheet taps still commit. Solo behaviour is untouched, including a session of one on a home wall (distinct-user count, so two devices on one account stay solo).

- **The crew gate** is a provider-level selector (`shouldDefaultToBrowse` in `@boardsesh/play-view` + `countDistinctSessionUsers`), published through its own context memoized on the boolean — it flips only across the 1↔2+ boundary, never on roster churn.
- **The latch is one-way**: armed on the first view-only gesture in a crew, exited only by Back to live or Put on the wall. A peer dropping off, the session ending, or an angle change never silently re-arms commit-on-swipe (angle changes re-anchor the preview at the new angle instead of clearing it).
- **Zero wire change**: browsing emits nothing — no `setCurrentClimb`, no queue writes, no playback broadcasts. The sync coordinator and queue reducer are untouched. Commits launder `playlist-peek:` synthetic uuids through `resolveCommittableQueueItem` (now shared with `nextClimb()`), so they never reach the wire.
- **Busy-wall confirm**: if someone lit a different climb while you browsed, the first Put-on-the-wall tap swaps the commit row inline to "Someone just lit {name}. Put yours up instead?" — no auto-timeout; it disarms only on Keep theirs, further browsing, or the conflict resolving itself. The lit climb is read continuously from board presence (`useWallClimb`), which also makes the On-the-wall pill work after plain swipes; a late-arriving presence read is adopted as the latch-start snapshot only when it pre-dates the latch by a 60s safety margin.
- **BLE stays honest** (no files under `lib/ble/` or `ble-protocol/` touched): route playback while previewing is fully local (`viewOnly` — no frame writes, no peer broadcast), and the mirror toggle acts on-screen only mid-preview.
- One-shot browsing notices (crew join + first solo settings-preview) ride the callout surface, spoken exactly once per platform path. Commit analytics now carry `session_id` + `participantCount`.

Adversarial Fable review (2 lenses) surfaced 8 findings — 3 majors (playlist/discover taps and tick-shaped opens bypassing the crew gate) all closed at the `openDrawer` choke point + `previewOnly` activation; every behavioural fix mutation-tested. Nothing rejected.

## Release Notes

- Browsing with your crew is now free: in a session with more than one climber, swiping and tapping around never touches the queue or the wall. One button — "Put on the wall" — makes the change, and if someone just lit a climb, the app checks before taking it.
- Changing the angle mid-browse keeps your spot instead of kicking you back to the live climb.

## Test plan

- [x] `vp run typecheck:mobile` / full monorepo typecheck
- [x] `vp run test:mobile` — 701 files / 6925 tests, 0 failures (~110 new: crew-gate tables, latch one-way render tests incl. session-end survival, peek laundering, busy-wall arming + late-read adoption tables, BLE viewOnly suppress/resume + broadcast gate, notice one-shot claims — each behavioural assertion mutation-checked)
- [x] `vp test run --project i18n`, `check:i18n`, `check:i18n:orphans`
- [x] `vp check`
- [x] `packages/shared/queue` untouched and green (zero wire change contract)
- [ ] Real-crew pass (2+ phones + a wall): `.boardsesh/qa-notes.md` § A2 — the group-session flows, busy-wall confirm timing, and BLE mirror/playback gating on a real board

### DIFF
--- a/docs/websocket-implementation.md
+++ b/docs/websocket-implementation.md
@@ -777,6 +777,20 @@ The session's `boardPath` is the route string the host first joined / created on
 
 ## Queue State Synchronization
 
+### The browse gate: browsing emits nothing
+
+Since #4281, a member of a session with **2+ distinct users** who is merely browsing
+climbs on mobile emits **no queue traffic at all** — no `CurrentClimbChanged`, no
+`QueueItemAdded`, no playback broadcasts. Browse-shaped gestures (swipes, climb-list
+taps, similar-climb taps, fresh drawer opens) pin a local view-only preview; only the
+explicit "Put on the wall" commit produces wire events, which then look exactly like
+any deliberate `setCurrentClimb`. This is a **sender-side** contract: receivers cannot
+distinguish an old client's browse noise from a real commit (slim payloads carry no
+intent flag — see the documented rejection in `sync-coordinator.ts`), so any future
+change to the message flow must preserve "browsing emits nothing" at the origin rather
+than trying to filter it downstream. The receiving side's auto-insert of every inbound
+current climb (issue #2217 behaviour) is only correct because of this contract.
+
 ### Event Types
 
 | Event                 | Description             | Fields                                                       |
--- a/packages/mobile/app/(tabs)/climbs/__tests__/index-keyboard.test.tsx
+++ b/packages/mobile/app/(tabs)/climbs/__tests__/index-keyboard.test.tsx
@@ -28,7 +28,10 @@ const mocks = vi.hoisted(() => ({
   searchFailed: false,
   offlineCatalog: null as 'missing' | 'queued' | null,
   activateClimb: vi.fn(),
+  activationOptions: undefined as { previewOnly?: boolean } | undefined,
   openPlayDrawer: vi.fn(),
+  // Mutable per-test: whether another climber is in the session.
+  isSharedSession: false,
   setSetting: vi.fn(),
   // Deep-link params, mutable so the screenshot-mode auto-opens can be driven
   // without a second mock scaffold.
@@ -237,6 +240,7 @@ vi.mock('../../../../src/theme/variants', () => ({
 
 vi.mock('../../../../src/providers/queue-provider', () => ({
   useActiveClimbUuid: () => null,
+  useIsSharedSession: () => mocks.isSharedSession,
   useQueueActions: () => ({ addToQueue: vi.fn() }),
 }));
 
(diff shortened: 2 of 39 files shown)


marcodejongh:
Review addressed in the rebase (f457854d6):

- **openDrawer re-fire**: already handled exactly as suggested — the host effect keys only on `[openTarget]` and calls through `openDrawerRef.current` (PlayDrawer ~line 898), so a roster flip never replays an open. No change.
- **`_resetJoinedBrowseNoticeForTests`**: kept — the underscore `*ForTests` seam is this codebase's established idiom for module-level state (`db/connection.ts`, `db/schema-ready.ts`, `_setNativeModuleForTests`, and friends); a `vi.resetModules` dance would be less direct for a deliberate module-level ledger.
- **Latch arming breadth**: deliberate and now documented in the block comment — a crew forming while a preview is already pinned latches too, so the mid-browse climber gets the safe posture without re-entering the preview.
- **WS doc**: added "The browse gate: browsing emits nothing" to `docs/websocket-implementation.md`'s Queue State Synchronization section, naming the sender-side contract and why receivers can't filter.

Also folded in the two Codex P2 fixes from the base branch (dead touches under the open callout; Back to live always closes it).

marcodejongh:
Both non-blocking notes handled:

- **Null-track fallback**: documented in place — it's the same degradation the wrong-board branch has always had, should be unreachable for `previewOnly` (the gate only fires in a shared session where an active board resolved), and degrading beats blocking the tap on a broken invariant.
- **`participantCount` basis**: confirmed identical — `sessionUsers` is literally `sessionRuntimeState.users` (queue-provider.tsx:150) and `sessionRuntimeStateRef` mirrors that same state; the ref just reads it at commit time, matching the pre-existing `partyMode` analytics pattern two call sites up.

The real-crew pass stays open deliberately — it needs 2+ phones at a wall (Marco's gate, `.boardsesh/qa-notes.md` § A2; the pr-4683 OTA channel carries the build).

marcodejongh:
Taking all four as noted-not-fixed — none moves behaviour and each would restart CI for cosmetic gain:

1. **Unbounded Set**: tens of UUIDs per app process, cleared on process death; an LRU here is more code than the data it bounds.
2. **Effect order**: documented at the declaration site; agreed it's refactor-fragile, and the comment is the guard we have for that class of coupling.
3. **onDismiss identity**: `handleCloseWallCallout` is deps-free by construction; this thread now documents the coupling if it ever gains deps.
4. **Re-anchor vs regrade race**: the two patches act on disjoint state (`drawerPreviewItem` vs queue items) — the committed head isn't the pinned preview by definition while re-anchoring fires.

Issue #4231, opened 2026-08-10
Mobile: no manage-gym shortcut after an auto-approved claim


marcodejongh:
## What happens

When a gym claim is auto-approved (admin setting `gym_claim_auto_approve`, added in #4230), the claimant becomes the gym owner immediately.

On **web**, the claim dialog says so and offers a "Set up your gym" button linking to `/gym/:uuid/manage`.

On **mobile**, `ClaimGymSheet` shows the confirmation text and a **Done** button. Nothing routes the new owner to the gym's manage screen — they have to find it themselves.

## Why it matters

The moment right after someone takes ownership is when they're most likely to add boards, set the kiosk branding, and fill in details. Web capitalises on that; mobile drops them back where they started.

## Where

- `packages/mobile/src/components/gym-directory/ClaimGymSheet.tsx` — the `confirmation.status === 'approved'` branch
- Web equivalent for reference: `packages/web/app/components/gym-entity/claim-gym-dialog.tsx`

Cache invalidation is already handled — `useRequestGymClaim` invalidates `gym`, `gymMembers`, `myGyms`, and `nearbyGyms` on `approved`. This is purely the missing navigation affordance.

Spotted in review on #4230.

marcodejongh:
🤖 Claimed by bug-batch loop `bugbash20` (batch 3, session 53132652). Plan → draft PR incoming. Parallel loop: please pick a different issue.

PR #4684, opened 2026-08-22
fix(mobile): Android tick-sheet Send button off-screen + note text clipped

alejandrosanchezcabana:
## Summary

Two related Android-only bugs in the create/edit tick sheets (#4231):

1. `@expo/ui`'s Android bottom sheet is a plain Material 3
   `ModalBottomSheet` — it only has two real states, a fixed ~50% "partial"
   and a content-fitting "expanded" — and never reads the JS `%` snap-point
   values (see `BottomSheet.android.tsx`), only the detent count and index.
   `CREATE_TICK_SNAP_POINTS`/`EDIT_TICK_SNAP_POINTS` are tuned against iOS's
   real first fraction (65%/80%, sized so the pinned action bar fits under
   the form), which is taller than Android's ~50% partial state. The sheet
   opened at partial by default, stranding the Send/Save button below the
   fold — the climber had to drag or scroll to reach it. `ModalSheet`/`Sheet`
   gain an opt-in `androidOpensExpanded` prop that presents at the LAST
   detent (expanded) on Android instead of the first; `LogAscentSheet` and
   `LogbookEditSheet` turn it on. Every other sheet (menus, pickers) is
   unaffected — this is an opt-in, not a platform-wide behavior change.

2. `TickNoteField`'s notes box wrapped a `flex: 1` multiline input in a
   `justifyContent: 'center'` column `View`. On Android that nested-flex
   vertical centering rendered the note text visibly cut in half. Flattened
   to a single styled `TextInput` (min/max height, padding and
   `textAlignVertical: 'top'` all on the input itself), matching the shape
   `EndSessionSheet`'s `notesInput` already uses without this problem.

A follow-up review found the `androidOpensExpanded` fix was silently
defeated for `LogbookEditSheet`: it opens via an imperative
`snapToIndex(0)`, which overwrote the seeded expanded-open index before
presenting, so the edit flow kept opening partial. Fixed by opening via
`.present()` instead. Also extracted the duplicated snap-point/index
derivation out of `ModalSheet.tsx`/`Sheet.tsx` into a shared helper,
dropped a dead `index` prop that had no effect on `ModalSheet`, and added
unit tests + a docs note per review feedback.

A real-device report after that still showed the Send/Flash button and the
attempts control off-screen on Android when logging a new tick. Root cause:
requesting the LAST *index* of a multi-detent config depends on `@expo/ui`'s
Android layer calling `sheetState.expand()` in a `LaunchedEffect` it wraps
in a swallowed `catch` ("Expanded anchor may be unreachable; never crash the
view") — a real, silent-failure race in the vendored library that can leave
the sheet resting at its ~50% partial state even when the right index was
requested. `androidOpensExpanded` now collapses the sheet to a single LAST
detent on Android instead of picking an index into the original multi-detent
config: a single detent makes the native sheet set `skipPartiallyExpanded`,
which removes "partial" as a landable state entirely, so it can only rest at
Expanded — the same mechanism every other near-full single-detent sheet in
this codebase already relies on. This also let us delete the index-seeding
machinery from the previous fix (`androidInitialPresentIndex`, the
`presentIndex` option on `useManagedSheet`) as dead code, since a collapsed
sheet's only detent is always index 0.

## Release Notes

- Fixed the Send/Save button being cut off below the screen when logging or editing a tick on Android
- Fixed note text appearing cut in half when typing a note on Android

## Test plan

- `vp check`, `vp run typecheck:mobile`, `vp run test:mobile`, `vp run check:mobile-bundle` all pass
- Traced `@expo/ui`'s Android bottom-sheet source (JS and the underlying Kotlin `ModalBottomSheetView.kt`) to confirm the collapsed-single-detent path opens expanded deterministically, without depending on the native `expand()` race, for both the create-tick (`LogAscentSheet`) and edit-tick (`LogbookEditSheet`) flows
- No physical-device/emulator pass done in this session — worth a manual Android check before merge if a reviewer has one handy

### DIFF
--- a/packages/mobile/src/components/ModalSheet.tsx
+++ b/packages/mobile/src/components/ModalSheet.tsx
@@ -22,7 +22,7 @@ import { useWindowBottomInset } from '../hooks/use-window-bottom-inset';
 import { hapticMedium } from '../lib/haptics';
 import { spacing } from '../theme/tokens';
 import { useTheme } from '../providers/theme-provider';
-import { androidSafeSnapPoints, androidInitialPresentIndex } from './sheet-snap-points';
+import { androidSafeSnapPoints } from './sheet-snap-points';
 import { useSheetBodyContentStyle } from './sheet-content-inset';
 import { useSheetColumnStyle } from './use-sheet-column-style';
 import { useSheetDetentProbe } from './sheet-detent-probe';
@@ -69,7 +69,8 @@ type ModalSheetProps = {
   /** Optional fixed header, rendered above the body and outside its scroll — so a
    * title and close affordance stay put while the body scrolls. */
   header?: ReactNode;
-  /** Open at the LAST snap point on Android instead of the first.
+  /** Present as a single, expanded-only detent on Android instead of the usual
+   * multi-detent config.
    *
    * `@expo/ui`'s Android sheet is a plain Material 3 `ModalBottomSheet`: it has
    * only two real states, a fixed ~50% "partial" and a content-fitting
@@ -80,10 +81,16 @@ type ModalSheetProps = {
    * sized to fit a pinned footer under the content) can be TALLER than
    * Android's fixed ~50% partial state. On Android that stranded the pinned
    * footer below the fold — the climber had to drag/scroll the sheet up just
-   * to reach the Send button (#4231). Opt in per sheet rather than flipping
-   * this for every `ModalSheet`: most sheets (menus, pickers) are genuinely
-   * fine at Android's partial state, and forcing them open expanded would be
-   * its own regression. No effect with a single detent, or on iOS/web. */
+   * to reach the Send button (#4231). Collapsing to a single detent (see
+   * `androidSafeSnapPoints`) makes Android's native sheet set
+   * `skipPartiallyExpanded`, which removes "partial" as a landable state
+   * entirely — more reliable than requesting the last index of a multi-detent
+   * config, which depends on an imperative `expand()` call `@expo/ui`'s own
+   * Android layer can silently no-op ("Expanded anchor may be unreachable").
+   * Opt in per sheet rather than flipping this for every `ModalSheet`: most
+   * sheets (menus, pickers) are genuinely fine at Android's partial state, and
+   * forcing them open expanded would be its own regression. No effect with a
+   * single detent, or on iOS/web. */
   androidOpensExpanded?: boolean;
 };
 
@@ -115,11 +122,12 @@ export const ModalSheet = forwardRef<ManagedSheetHandle, ModalSheetProps>(functi
   const solidBackground = useMemo(() => ({ backgroundColor: sheetSurface }), [sheetSurface]);
 
   // Single-detent sheets jump to full screen on @expo/ui's Android sheet — give
-  // them a partial state instead (see androidSafeSnapPoints).
-  const effectiveSnapPoints = useMemo(() => androidSafeSnapPoints(snapPoints), [snapPoints]);
-
-  // See `androidOpensExpanded` above.
-  const initialIndex = androidInitialPresentIndex(effectiveSnapPoints, androidOpensExpanded);
+  // them a partial state instead. An `androidOpensExpanded` sheet instead
+  // collapses to its single LAST detent on Android (see androidSafeSnapPoints).
+  const effectiveSnapPoints = useMemo(
+    () => androidSafeSnapPoints(snapPoints, androidOpensExpanded),
+    [snapPoints, androidOpensExpanded],
+  );
 
   const sheetRef = useRef<BottomSheetMethods>(null);
   const managed = useManagedSheet({
@@ -128,7 +136,6 @@ export const ModalSheet = forwardRef<ManagedSheetHandle, ModalSheetProps>(functi
     sheetRef,
     onClose,
     onFullyDismissed,
-    presentIndex: initialIndex,
   });
   useImperativeHandle(ref, () => managed.handle, [managed.handle]);
 
@@ -136,10 +143,7 @@ export const ModalSheet = forwardRef<ManagedSheetHandle, ModalSheetProps>(functi
   onChangeRef.current = onChange;
 
   // Track the resting detent so the iOS column bound follows drags between detents.
-  // Seeded from `initialIndex` so a sheet that opens expanded on Android (see
-  // `androidOpensExpanded`) doesn't render one frame at the first-detent bound
-  // before the native `onChange` confirms the real one.
-  const [activeIndex, setActiveIndex] = useState(initialIndex);
+  const [activeIndex, setActiveIndex] = useState(0);
   const columnStyle = useSheetColumnStyle(snapPoints, { enableDynamicSizing, activeIndex });
   // Dev-only observers for #3922 — they feed a log line, never layout.
   const { probeProps, sentinelProps, onColumnLayout } = useSheetDetentProbe(columnStyle, 'ModalSheet');
@@ -225,11 +229,13 @@ export const ModalSheet = forwardRef<ManagedSheetHandle, ModalSheetProps>(functi
   return (
     <BottomSheetModal
       ref={sheetRef}
-      // Not `index={initialIndex}`: `BottomSheetModal` only reads its own `index`
-      // prop inside its own `.present()` method, which `useManagedSheet` never
-      // calls — it drives presentation via `snapToIndex(desiredIndexRef.current)`
-      // instead, seeded from `presentIndex` above. `-1` (always starts closed)
-      // is the only value that matters here.
+      // `BottomSheetModal` only reads its own `index` prop inside its own
+      // `.present()` method, which `useManagedSheet` never calls — it drives
+      // presentation via `snapToIndex` instead, always at index 0 (see
+      // `effectiveSnapPoints` above: an `androidOpensExpanded` sheet is already
+      // collapsed to its single expanded detent on Android, so index 0 IS
+      // expanded there too). `-1` (always starts closed) is the only value
+      // that matters here.
       index={-1}
       snapPoints={enableDynamicSizing ? undefined : effectiveSnapPoints}
       enableDynamicSizing={enableDynamicSizing}
--- a/packages/mobile/src/components/ModalSheet.tsx
+++ b/packages/mobile/src/components/ModalSheet.tsx
@@ -69,6 +69,22 @@ type ModalSheetProps = {
   /** Optional fixed header, rendered above the body and outside its scroll — so a
    * title and close affordance stay put while the body scrolls. */
   header?: ReactNode;
+  /** Open at the LAST snap point on Android instead of the first.
+   *
+   * `@expo/ui`'s Android sheet is a plain Material 3 `ModalBottomSheet`: it has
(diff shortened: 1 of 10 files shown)


alejandrosanchezcabana:
Addressed via a deeper review of this PR:

- **`LogbookEditSheet`'s `androidOpensExpanded` fix was silently defeated.** It opens via `editSheetRef.current?.snapToIndex(0)`, and `ManagedSheetHandle.snapToIndex` unconditionally overwrites the `presentIndex` seed before presenting — so the edit-tick sheet kept opening at Android's partial detent, unfixed. Switched to `.present()`, which respects the seeded index.
- Dropped the dead `index={initialIndex}` prop on `ModalSheet`'s `<BottomSheetModal>` — it only fed a `.present()` method this codebase never calls; the real fix is entirely carried by `presentIndex` on `useManagedSheet`.
- Extracted the duplicated `effectiveSnapPoints`/`initialIndex` derivation from `ModalSheet.tsx` and `Sheet.tsx` into a shared `androidInitialPresentIndex` helper in `sheet-snap-points.ts`.
- Added unit coverage for the new helper, and a section on Android's two-state sheet model to `docs/mobile-sheets-vs-routes.md` — addressing the missing-tests and docs-gap feedback above.
- Amended the original commit to drop the AI co-authorship trailer and re-ran the pre-commit hook (it had been skipped with `--no-verify`); all `vp check`, `typecheck:mobile`, `test:mobile`, and `check:mobile-bundle` pass locally.

alejandrosanchezcabana:
Re the last review round:

**`index={-1}` safety** — verified, no code change needed. `ModalSheet`'s `useImperativeHandle` only ever exposes `managed.handle` (`ModalSheet.tsx:122`), never the raw `sheetRef` pointing at `<BottomSheetModal>`. `ManagedSheetHandle.present()` (`sheet-presentation-provider.tsx:514`) calls `coordinator.setDesiredOpen(id, true)`, which drives the registered `present` callback — `sheetRef.current?.snapToIndex(desiredIndexRef.current)` — never the wrapper's own native `.present()` method that reads the `index` prop. No caller, internal or external, can reach that path, so `index={-1}` is safe as documented.

**Default-argument test gap** — already covered, not actually missing. `sheet-snap-points.test.ts`'s `'leaves an already multi-detent sheet unchanged'` case calls `androidSafeSnapPoints(['50%', '90%'])` with no second argument, inside the `describe` block that pins `Platform.OS = 'android'` — that *is* the default-arg/Android/multi-detent case.

**No physical-device validation** — this is the one real gap, and it's the same one already called out in the PR description. There's no existing deep-link into a signed-in create/edit-tick flow with real climb data for `vp run mobile:android-shots` to drive, so I can't close it from here without building new Maestro flow scripting — disproportionate for this fix. The user who originally reported this bug is testing on a real Android device; I've asked them to confirm the new build resolves it.

packages/mobile/src/components/ModalSheet.tsx
```diff
@@ -201,7 +211,14 @@ export const ModalSheet = forwardRef<ManagedSheetHandle, ModalSheetProps>(functi
...
       ref={sheetRef}
-      index={0}
+      // `BottomSheetModal` only reads its own `index` prop inside its own
+      // `.present()` method, which `useManagedSheet` never calls — it drives
+      // presentation via `snapToIndex` instead, always at index 0 (see
+      // `effectiveSnapPoints` above: an `androidOpensExpanded` sheet is already
+      // collapsed to its single expanded detent on Android, so index 0 IS
+      // expanded there too). `-1` (always starts closed) is the only value
+      // that matters here.
+      index={-1}
```
marcodejongh: **Bug (Expo web): `index={-1}` prevents every `ModalSheet` from opening in the browser app.**

The premise in this comment holds for native `@expo/ui` (its `BottomSheetModal` strips `index` and always mounts the inner sheet at `-1`), but on web this component resolves to `src/web-shims/bottom-sheet.tsx`, whose `BottomSheetModal` *does* read the `index` prop:

- `initialIndex` becomes `-1` and is forwarded to Gorhom: `<GorhomBottomSheetModal index={-1}>`.
- On present, Gorhom 5.2.14 mounts its internal `BottomSheet` with `index={-1}` + `animateOnMount`, which is the documented "start closed" state (`BottomSheet.tsx` line ~875 returns `closedDetentPosition` for `_providedIndex === -1`), so the sheet mounts at the closed position and never fires `onChange` with an index >= 0.
- The shim's `snapToIndex(0)` (what `useManagedSheet`'s `present` calls) queues `pendingActionRef = {type:'index', index:0}` and waits for a visible `onChange` to run it — which never comes. `dismissalStateRef` flips to `'presented'`, so every retry lands in the `else if (action)` branch and just re-queues. The sheet never appears.

Repro: Expo web (`BOARDSESH_WEB=1`), tap the tick FAB (LogAscentSheet via `visible`), or open any ModalSheet surface (playlist actions menu, climb actions, BLE control sheet, feedback sheet) — nothing renders. The shim tests only cover `BottomSheetModal` with `index={0}` / `{1}`, so nothing catches this, and the modal-with-`-1` case is exactly the one the non-modal shim's `present()` clamp comment warns about.

Fix options: keep `index={0}` here (it was already ignored by native `@expo/ui`, so `-1` changes nothing on iOS/Android), or teach the web shim's modal to treat a negative `index` as 0 for mount/present purposes.

---

packages/mobile/src/components/you/LogbookTab.tsx
```diff
@@ -358,7 +358,7 @@ export function LogbookTab({ userId, topInset = 0, viewerIsOwner = true }: Logbo
 
   const openEditSheet = useCallback((ascent: AscentFeedItem) => {
     setEditAscent(ascent);
-    editSheetRef.current?.snapToIndex(0);
+    editSheetRef.current?.present();
```
marcodejongh: **Minor iOS behavior change: `present()` reopens the edit sheet at the last-dragged detent, not the first.**

`ManagedSheetHandle.present()` → `setDesiredOpen(true)` → registration `present` → `sheetRef.current?.snapToIndex(desiredIndexRef.current)`, and `desiredIndexRef` is updated by `handleNativeChange` to whatever detent the sheet last rested at. So on iOS (`EDIT_TICK_SNAP_POINTS = ['80%', '92%']`): open a tick's edit sheet, drag it up to 92% (index 1), close it, then swipe-to-edit another tick — the sheet now opens at 92% instead of 80%. The pre-PR `snapToIndex(0)` pinned every open to the first detent.

On Android this is moot (the collapsed config has a single detent, so `onChange` only ever reports 0). Since the seeded-index machinery this `present()` switch was originally working around is now deleted, `snapToIndex(0)` would again be equivalent on Android while preserving the deterministic iOS open — unless the sticky-detent reopen is intended, in which case a one-line comment here would help.

---

packages/mobile/src/components/sheet-snap-points.ts
```diff
@@ -21,13 +21,26 @@ const ANDROID_NEAR_FULL_DETENT_PERCENT = 75;
...
+ * SwiftUI / CSS, so they're returned unchanged.
  */
-export function androidSafeSnapPoints(snapPoints: (string | number)[]): (string | number)[] {
-  if (Platform.OS !== 'android' || snapPoints.length !== 1) return snapPoints;
+export function androidSafeSnapPoints(
+  snapPoints: (string | number)[],
+  androidOpensExpanded = false,
+): (string | number)[] {
+  if (Platform.OS !== 'android') return snapPoints;
+  if (androidOpensExpanded && snapPoints.length > 1) return [snapPoints[snapPoints.length - 1]];
```
marcodejongh: **Edge case: an opted-in sheet with a single sub-75% detent still opens partial.**

`androidOpensExpanded` only short-circuits when `snapPoints.length > 1`. A single detent below 75% (e.g. `['65%']`) falls through to the padding branch and comes back as `['65%', '100%']` — a multi-detent config that opens at Android's ~50% partial state, which is exactly the resting state this opt-in exists to remove. The prop docs on `Sheet`/`ModalSheet` say "no effect with a single detent", but the effect here is worse than no effect: the pinned footer strands below the fold again, silently.

No current consumer hits this (both tick sheets are multi-detent), but the first person who trims a tick sheet to one detent, or opts in a single-detent form sheet, reproduces #4231 with the flag set. Suggest making the opt-in win over the padding branch (`if (androidOpensExpanded) return [snapPoints[snapPoints.length - 1]];` before the single-detent logic) or documenting the sub-75% single-detent case explicitly; a test for `androidSafeSnapPoints(['65%'], true)` would pin whichever contract you choose.

---

marcodejongh (commented):
Code review of 1bd4d04 (single-pass inline review at high effort; the Agent tool was unavailable in this session, so this is one reviewer pass through all angles, not the usual multi-agent fan-out + verify pipeline).

3 findings, most severe first:

1. **[correctness, high]** `ModalSheet.tsx:221` — `index={-1}` breaks every `ModalSheet` on the Expo web target: the Gorhom-backed web shim forwards `index` to `GorhomBottomSheetModal`, which mounts its inner sheet closed at `-1`, so the queued `snapToIndex(0)` pending action never runs and no modal sheet ever opens in the browser app. Verified against Gorhom 5.2.14 source and the shim's `requestOpen`/`pendingActionRef` state machine; the shim tests only cover modal `index={0}`/`{1}`.
2. **[behavior, low]** `LogbookTab.tsx:361` — `present()` opens at `desiredIndexRef` (the last-rested detent), so on iOS the edit sheet reopens at 92% after a user dragged it there, where `snapToIndex(0)` used to pin every open to 80%.
3. **[latent, low]** `sheet-snap-points.ts:42` — `androidOpensExpanded` with a single sub-75% detent falls through to the pad-with-100% branch and opens partial, the state the opt-in promises to remove.

The Android mechanism itself checks out: `@expo/ui`'s `BottomSheet.android.tsx` sets `skipPartially = !hasMultipleSnapPoints`, so collapsing to one detent does deterministically remove the partial resting state, and native `BottomSheetModal` ignores the `index` prop entirely (mounts at `-1` regardless), so finding 1 is web-only.

---

packages/mobile/src/components/ModalSheet.tsx
alejandrosanchezcabana: Fixed in 6e084526e — reverted to `index={0}`. Kept the comment explaining why (the web shim reads this prop for both mount and its own `present()` fallback; native `@expo/ui` ignores it either way).

packages/mobile/src/components/you/LogbookTab.tsx
alejandrosanchezcabana: Fixed in 6e084526e — switched back to `snapToIndex(0)` now that the seeded-index workaround `present()` was covering for is gone. This pins every open to the first detent deterministically on iOS, and is equivalent to `present()` on Android since `androidOpensExpanded` already collapses to a single detent there.

packages/mobile/src/components/sheet-snap-points.ts
alejandrosanchezcabana: Fixed in 6e084526e — moved the `androidOpensExpanded` check ahead of the single-detent padding branch, so it always wins regardless of the detent's size. Added a regression test (`androidSafeSnapPoints(['65%'], true)`) and updated the prop docs to state the real contract.

PR #4686, opened 2026-08-22
fix(sync): reconcile kilter_id before the ratings upsert

marcodejongh:
## Summary

The kilter daemon's **user-data** half had not completed a single successful sync inside the
30-day log retention window. The catalog half was fine and the aurora daemon on the same VM
succeeded constantly, so this was never infra. Two users sat at attempt 63.

```
[KilterSyncRunner] Permanent sync failure for user 0df1d963-… (status=error, attempt 63):
  Failed query: insert into "board_climb_ratings" …
  cause: PostgresError: duplicate key value violates unique constraint
         "board_climb_ratings_kilter_id_unique"
    at async applyClimbRatings (packages/kilter-sync/src/sync/user-sync.ts:1146:3)
```

### Root cause

`board_climb_ratings` carries **two** independent uniques over the same row: the natural key
`(board_type, climb_uuid, angle, user_id)` and a **global** partial unique on `kilter_id`.
`applyClimbRatings` conflict-targets the natural key, and Postgres permits exactly one conflict
target per `ON CONFLICT` — so the surrogate unique was uncovered on **both** the `INSERT` branch
and the `DO UPDATE` branch (which writes `kilter_id` through `COALESCE`).

Two paths reached the violation:

- **Intra-statement** — the batch deduped on the *natural* key only. Two PUTs sharing one
  `climb_rating_uuid` at different natural keys (an upstream angle edit, or two source UUIDs that
  re-canonicalise differently) both survived into one `INSERT` with the same `kilter_id`.
- **Cross-row** — the incoming `climb_rating_uuid` already sat on a different row: this user's row
  at an older natural key, or a row owned by a *different* Boardsesh user (one Kilter account
  linked to two accounts).

The `23505` escaped `syncKilterUserData` entirely, so **the circuits phase never ran either** —
these users silently lost playlist sync too. The runner fails closed on a non-`KilterApiError`, so
a raw Postgres error was classified **permanent**. PowerSync re-delivers the full snapshot every
cycle, so it failed identically forever.

`applyLogs` has defended against all of this for `boardsesh_ticks` all along. Its own comment
claims it mirrors `applyClimbRatings` — that belief was the bug.

### The fix

- Dedupe by `climb_rating_uuid` **before** the natural-key dedupe (surrogate-first is the only
  order under which one upstream row maps to one output row at its *current* natural key).
- One global, deliberately un-user-scoped pre-SELECT, partitioned into own vs foreign owners. A
  foreign-owned id is **skipped and logged**, never stolen.
- **Repoint** a same-user id that moved: detach the stale row in the same transaction, then let the
  upsert stamp it. Nothing is ever deleted.
- Chunk the upsert at 100 with a savepoint and row-by-row replay, so a row Postgres refuses costs
  one rating rather than the buffer, the user, and the circuits phase.
- Take the per-user advisory lock so the read-then-write can't race.
- Isolate each sync phase so a ratings failure no longer cancels circuits.

### Why `kilter_detached_at` is stamped on a repoint

Worth a reviewer's attention, since "the rating moved" is not "Kilter deleted it". `push-back`
selects Boardsesh-origin ratings with `kilter_id IS NULL AND kilter_detached_at IS NULL`. Nulling
the surrogate *without* the marker would make the stale row look never-pushed, and the day the POST
is wired push-back would re-create it upstream as a duplicate. The marker's operational contract in
both consumers is "not backed by a live upstream rating", which is exactly true here. Trade-off:
the old `(climb, angle)` loses its `effectiveQuality` star — correct, since upstream no longer
asserts a rating there.

### The health reporter that hid it

Every cycle logged `Sync health summary failed: snapshot.oldestAttemptAt.toISOString is not a
function` — 28 of the 49 boardsesh error lines in 24h. The fleet summary has **never once
emitted**. A raw `sql` aggregate carries no runtime decoder (the `sql<Date>` generic is
compile-time only), and drizzle's postgres-js driver deliberately *strips* postgres.js's own date
parsers so column mappers are the single source of truth — so `min(last_sync_attempt_at)` arrived
as raw Postgres text. Every numeric field on that snapshot got both a `::int` cast and a `Number()`
coercion; the timestamp got neither.

Fixed with `.mapWith(...)`, plus a defensive read in the pure formatter that treats a zone-less
wall clock as UTC rather than in the host process's timezone — a regression test caught that skew
in the first version of this fix.

### Making the next one loud

`kilter-sync` emitted **no success log line at all** — grepping it for successes returned zero
whether it was healthy or stone dead, which is a large part of why 30 days passed unnoticed. Added
one (matching aurora's wording so one alert rule covers both), plus the hourly fleet health summary
kilter never had. Both share code with aurora via `@boardsesh/sync-runtime` and
`@boardsesh/db/queries` rather than being copied.

The matching Loki alert (absence-of-success, **not** the error string — that goes quiet the moment
this lands) is a separate change in the Ansible repo.

## Release Notes

- Your Kilter ratings, ticks and circuits sync again — a bad rating could jam the whole sync, and for some climbers it had been stuck for over a month.

## Test plan

- `vp run typecheck` — clean.
- `vp test run --project kilter-sync --project aurora-sync --project sync-runtime` — 508 passed, 1 skipped.
- `vp check` — no new findings (2 pre-existing errors remain in `scripts/expo-web-e2e.ts`, untouched here).
- New real-DB regression tests in `packages/backend/src/__tests__/climb-ratings-kilter-id-repoint.test.ts`
  cover both violation paths, the swap case, the foreign-owner skip, steady-state no-churn, and
  poison-row isolation. **These only bite because `schema-sql.ts` now creates the surrogate uniques
  it had been missing** — that omission is why the bug was unreproducible in CI for 30+ days. The
  file also asserts the index exists, so a future edit can't silently disarm the suite.
- New unit test drives a raw Postgres timestamp string through `getSyncHealthSnapshot` →
  `formatSyncHealthSummary`; it throws on `main`.
- `user-sync-phase-isolation.test.ts` pins that a throwing ratings phase no longer stops the
  circuits phase, and that the failure is still reported rather than swallowed.

### Verified the tests actually fail without the fix

A green suite is what everyone had for the last 30 days, so I did not trust one. On a throwaway
branch with **only** the surrogate dedupe and the repoint neutered — armed schema and tests kept —
`test-backend` failed on exactly the four tests that cover the fix:

| Test | Neutered |
|---|---|
| collapses two PUTs sharing one `climb_rating_uuid` | ❌ fails |
| repoints a `kilter_id` that moved to a different angle | ❌ fails |
| repoints onto a Boardsesh-origin row | ❌ fails |
| swaps two `kilter_id`s in one batch | ❌ fails |
| index guard / foreign-owner skip / steady-state no-churn / poison isolation | ✅ still pass |

The failure was `duplicate key value violates unique constraint
"board_climb_ratings_kilter_id_unique"` — the production error, reproduced.

### Known follow-ups (not introduced here)

- The `DO UPDATE` still overwrites a *different* existing `kilter_id` on the natural-key row without
  a divergence log, where `applyLogs` skips-and-logs. It can't produce a `23505`, and for ratings the
  natural key genuinely is the identity — but it's worth a look.
- An `'error'` credential never leaves the candidate set; it retries at the 6 h backoff cap forever.
- The 445/week `divergent kilter_id on tick` warnings are real data drift; the tick path's skip is
  deliberate and reconciling it is its own piece of work.

https://claude.ai/code/session_01HiTqFRMv7ALs9jEd8cJjrB

### DIFF
--- a/packages/aurora-sync/src/runner/sync-runner.test.ts
+++ b/packages/aurora-sync/src/runner/sync-runner.test.ts
@@ -945,6 +945,46 @@ describe('SyncRunner.getSyncHealthSnapshot', () => {
     });
   });
 
+  // The regression that mattered. The test above models postgres-js's
+  // string-ness for the COUNTS and then hands oldestAttemptAt a real Date —
+  // the same blind spot the production code had, which is why nothing caught
+  // `snapshot.oldestAttemptAt.toISOString is not a function` for a month. Drive
+  // the raw Postgres text shape all the way through to the formatted line.
+  it('formats a health summary when the driver hands back a raw timestamp string', async () => {
+    const dbShim = {
+      select: () => ({
+        from: () => ({
+          where: () =>
+            Promise.resolve([
+              {
+                total: '4',
+                active: '4',
+                pending: '0',
+                error: '0',
+                expired: '0',
+                inBackoff: '0',
+                // `timestamp without time zone` as postgres-js yields it when
+                // no column decoder is attached to the min() aggregate.
+                oldestAttemptAt: '2026-05-06 07:08:09.123',
+              },
+            ]),
+        }),
+      }),
+    };
+
+    const runner = new SyncRunner();
+    const runnerPrivates = runner as unknown as SyncRunnerPrivates & {
+      getClient: () => { client: unknown; db: unknown };
+    };
+    vi.spyOn(runnerPrivates, 'getClient').mockReturnValue({ client: {}, db: dbShim });
+
+    const snapshot = await runnerPrivates.getSyncHealthSnapshot();
+
+    // On main this throws instead of returning a line.
+    expect(() => formatSyncHealthSummary(snapshot)).not.toThrow();
+    expect(formatSyncHealthSummary(snapshot)).toContain('oldestAttempt=2026-05-06T07:08:09.123');
+  });
+
   it('defaults counts to 0 and oldestAttemptAt to null when no rows come back', async () => {
     const dbShim = {
       select: () => ({ from: () => ({ where: () => Promise.resolve([]) }) }),
--- a/packages/aurora-sync/src/runner/sync-runner.ts
+++ b/packages/aurora-sync/src/runner/sync-runner.ts
@@ -12,7 +12,7 @@ import {
   claimNextCredentialForSync,
   claimSharedSyncSlot,
   credentialBackoffMs,
-  credentialRetryReadySql,
+  getCredentialFleetSnapshot,
   readSharedSyncCursor,
   releaseDaemonLease,
   selfHealStaleClimbStats,
@@ -33,7 +33,13 @@ import { isAuroraRequestError, isTransientAuroraError, isTransientSharedSyncAuro
 import { decrypt, encrypt } from '@boardsesh/crypto';
 import type { LocationSyncSummary } from '@boardsesh/location-sync';
 import type { AuroraBoardName } from '../api/types';
-import { DaemonLease, resolveDaemonOptions, runDaemonLoop } from '@boardsesh/sync-runtime';
+import {
+  DaemonLease,
+  formatSyncHealthSummary as formatSharedSyncHealthSummary,
+  resolveDaemonOptions,
+  runDaemonLoop,
+  type SyncHealthSnapshot,
+} from '@boardsesh/sync-runtime';
 import type { SyncRunnerConfig, SyncSummary, CredentialRecord, DaemonOptions, SyncErrorContext } from './types';
 
 type RunnerClient = ReturnType<typeof postgres>;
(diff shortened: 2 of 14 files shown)


marcodejongh:
Thanks — all three addressed in ed090b03d.

1. **Duplicate type shapes** — `CredentialFleetSnapshot` is now an alias of `SyncHealthSnapshot` rather than a second declaration. `@boardsesh/db` already depends on `@boardsesh/sync-runtime`, so the type flows that direction; sync-runtime stays dependency-free and must never import back.
2. **Phase isolation untested** — fair, and it was the load-bearing gap. Added `user-sync-phase-isolation.test.ts`: a throwing ratings phase must not stop circuits from running, and must still surface as a failure rather than be swallowed (isolation must not become "pretend it worked"). Driven through the db shim's `transaction` rather than stubbing same-module functions.
3. **Divergence TODO** — now marked in `writeClimbRatings` where it lives, not only in the PR description.

Separately, I verified the new regression tests actually **fail** without the fix, rather than trusting a green run — the same assumption that let this bug live for 30 days. On a throwaway branch with only the surrogate dedupe and the repoint neutered (armed schema and tests kept), `test-backend` failed on exactly the four tests covering the fix, with `duplicate key value violates unique constraint "board_climb_ratings_kilter_id_unique"` — the production error — while the four covering un-neutered behaviour stayed green.

marcodejongh:
All four addressed in 38eee30f9.

**1. Duplicate phase names inflate the count** — fixed. `phaseErrors` is now a `Map` keyed by phase name; first failure per phase wins. Confirmed the old behaviour was exactly as described: with 501 rating ops (one past `STREAM_FLUSH_THRESHOLD`, so the mid-stream flush drains the buffer and the spare op gives the final flush something to fail on) it produced `failed in 2 phase(s): climb_ratings: ...; climb_ratings: ...`. There's now a test pinning it, and I verified it fails with the dedup removed.

**2. Test relies on transaction-count ordering** — agreed, that was an invisible invariant. The mock now identifies the phase by the *table* the transaction body touches (`boardClimbRatings` / `playlists` / `boardseshTicks`), so a future statement opening a transaction ahead of the ratings phase can't silently shift the assertions. Also verified this version still fails when `runPhase` is removed.

**3. COALESCE overwrite left unlogged** — you're right that this was inconsistent with the PR's own argument, so it's logged now rather than deferred:

```
[kilter-sync] divergent kilter_id on rating <climb>@<angle> for user <id>:
  existing=kr-old incoming=kr-new — replacing (natural key is the identity)
```

The behaviour is unchanged (still replaces, still can't raise 23505 — one value leaves the partial index as another enters); only the silence is gone. Implemented by widening the existing pre-SELECT with an `OR` arm over the incoming natural keys rather than adding a second query, so it costs one round trip per flush, not one per chunk. Covered by a real-DB test.

**4. aurora_id** — checked rather than assumed: `board_climb_ratings` has exactly three writers, all in `applyClimbRatings` (REMOVE detach, repoint detach, upsert), and **nothing in the monorepo sets `aurora_id` at all**. aurora-sync never touches this table. So there's no live vulnerability — but the hazard is real for whoever adds that writer, so `schema-sql.ts` now says so next to the index: a single-target `ON CONFLICT` on the natural key does not cover it, and the same reconcile-before-upsert treatment will be needed.

PR #4687, opened 2026-08-23
test: SCRATCH do the ratings regression tests bite

marcodejongh:
Throwaway. Neuters the surrogate dedupe and the repoint in `applyClimbRatings` while keeping the armed test schema and the new tests, to confirm they actually fail without the fix. Closing as soon as `test-backend` reports.

Expected to fail: the two intra-statement / cross-row paths, the DO UPDATE adoption, and the swap.

Related: #4686

### DIFF
--- a/packages/aurora-sync/src/runner/sync-runner.test.ts
+++ b/packages/aurora-sync/src/runner/sync-runner.test.ts
@@ -945,6 +945,46 @@ describe('SyncRunner.getSyncHealthSnapshot', () => {
     });
   });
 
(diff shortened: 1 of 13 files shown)


marcodejongh:
Confirmed what it was opened to check: the four tests covering the fix fail here with `duplicate key value violates unique constraint "board_climb_ratings_kilter_id_unique"` — the exact production error — while the four covering un-neutered behaviour stay green. The regression tests bite. Closing.

PR #4689, opened 2026-08-23
fix(sync): carry the ratings tie-break across flush boundaries

marcodejongh:
## Summary

Follow-up to #4688, found by running that fix against the production account it was meant to help.
The tie-break worked — and the account still didn't converge: **714 identical replacements on every
run**, same rows, same direction each time.

Same direction rather than alternating is the tell. #4688 assumed a ping-pong *between* runs. It's
actually a ping-pong *within* a run:

```
run 3:  B044DFB7…@50  6aeb9e71… -> 000e6e3e…
run 4:  B044DFB7…@50  6aeb9e71… -> 000e6e3e…   ← identical, not reversed
```

If the write had stuck, run 4 would see `000e6e3e` and report nothing.

### Cause

The dedupe is **per flush**, and a flush is only `STREAM_FLUSH_THRESHOLD` (500) ops wide. Two
upstream ratings for one climb/angle landing in *different* flushes both get written, and the later
flush wins — regardless of which rating is actually newer. So flush 1 writes the correct winner,
flush 4 overwrites it with the loser, the run ends in a stable-but-wrong state, and the next sync
re-applies the same correction. Forever.

This is exactly the hazard aurora-sync's `dedupeByAuroraId` comment already documents:

> Deduping at PAYLOAD level rather than per chunk matters: a chunk boundary between the two copies
> would hide the collision from a per-chunk guard and let it reach the index.

That warning was about the unique index; the same boundary also defeats a per-chunk *tie-break*.

### Fix

Thread one claim map through the whole sync so the newest-wins order is global rather than per
flush. A later flush skips a candidate an earlier flush already claimed unless it's genuinely newer
under the same total order. The parameter defaults to an empty map, so a one-shot call keeps its old
single-flush semantics.

Cost: one `Map` entry per distinct climb/angle for the duration of a sync.

## Release Notes

## Test plan

- Direct `tsc` clean on both changed packages; 512 unit tests pass; `vp check` no new findings.
- Two new real-DB tests that a single-batch test **cannot** express, since the bug only exists
  across flush boundaries: a later flush must not undo an earlier flush's newer pick, and a second
  sync over the same snapshot must leave `updated_at` untouched.

⚠️ Note on tooling: `vp run typecheck` returned a false clean here from a cached result while the
file genuinely had two `TS2304` errors. Direct `tsc -p` caught it. Worth knowing when a change
touches a package the cache thinks is unchanged.

## Production context

- `0df1d963…` — fully converged on #4688: run 1 did a one-time reconciliation (45 replacements,
  1 repoint), run 2 was **0 and 0**. Steady state is now 3 log lines per sync, down from ~700.
  That account's ratings fit in a single flush, which is why it converged and this one didn't.
- `51e2b642…` — no duplicate-key errors, but still fails on the 120s `REQUEST_TIMEOUT_MS` partway
  through a large snapshot. **Separate pre-existing issue, not addressed here.** This PR removes its
  714 redundant UPDATEs per run, which reduces the work but is unlikely to close a ~7s gap on its
  own.

https://claude.ai/code/session_01HiTqFRMv7ALs9jEd8cJjrB

### DIFF
--- a/packages/backend/src/__tests__/climb-ratings-kilter-id-repoint.test.ts
+++ b/packages/backend/src/__tests__/climb-ratings-kilter-id-repoint.test.ts
@@ -360,6 +360,50 @@ describe('applyClimbRatings kilter_id reconciliation (real DB)', () => {
     expect(new Date(second!.updated_at).getTime()).toBe(new Date(first!.updated_at).getTime());
   });
 
+  it("does not let a later flush undo an earlier flush's newer pick", async () => {
+    // The bug a single-batch test cannot see. Duplicates for one climb/angle
+    // that land in DIFFERENT flushes were both written, and the later flush
+    // won regardless of which rating was newer — so the choice never
+    // converged and updated_at churned on every sync forever.
+    await seedUser(USER_ID);
+    const newer = putOp({ climbRatingUuid: 'kr-newer', angle: 40 });
+    (newer.data as Record<string, unknown>).created_at = '2026-06-01T12:00:00.000Z';
+    const older = putOp({ climbRatingUuid: 'kr-older', angle: 40 });
+
+    // One shared map = one sync; two calls = two flushes.
+    const claimed = new Map<string, { kilterId: string; createdAtMs: number }>();
+    await applyClimbRatings(applyTx, USER_ID, [newer], aliasCacheFor([CLIMB]), () => {}, claimed);
+    await applyClimbRatings(applyTx, USER_ID, [older], aliasCacheFor([CLIMB]), () => {}, claimed);
+
+    const rows = await readRatings(USER_ID);
+    expect(rows).toHaveLength(1);
+    // The newer rating from flush 1 survives; the older one in flush 2 is
+    // skipped rather than overwriting it.
+    expect(rows[0]?.kilter_id).toBe('kr-newer');
+  });
+
+  it('carries the flush claim across a whole sync without churning updated_at', async () => {
+    await seedUser(USER_ID);
+    const newer = putOp({ climbRatingUuid: 'kr-newer', angle: 40 });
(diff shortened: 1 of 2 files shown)


marcodejongh:
Good catch, and it's **worse than the assessment** — this is live, not latent.

> The current error propagation from `flushClimbRatings()` looks like it would abort the sync, making this safe today.

It doesn't abort. The mid-stream flush is wrapped in `runPhase`, which catches and continues:

```ts
if (buffer.climb_ratings.length >= STREAM_FLUSH_THRESHOLD) {
  await runPhase('climb_ratings', async () => {   // catches, records, continues
    await primeAliasCacheForOps(...);
    await flushClimbRatings();
  });
}
```

That per-flush error recovery shipped in #4686's phase isolation, which is **already merged and deployed** — so the exact condition described as hypothetical is reachable in production right now. A rolled-back flush leaves claims for rows that were never written, and the next flush silently drops those climb/angle keys.

Fixed structurally rather than by documentation, since a documented invariant here would be one `await` away from being violated again: `applyClimbRatings` now **stages** its claims and returns them, and the caller merges only after `db.transaction` resolves. There is no longer a way to record a claim for an uncommitted write.

Also excluded rows Postgres refused during the row-by-row replay — same reasoning, they were never written either. That one I'd missed entirely.

On the test point — you were right, and it was worse than "not exercised". The two existing cross-flush tests merged claims eagerly, so the shared map never carried anything between calls: they passed **without exercising the protection they're named for**. Both now go through an `applyFlush` helper that mirrors `flushClimbRatings` exactly, and there's a new test that rolls a flush back and asserts no claim survives *and* that a later flush still writes.

On the minor point: the `userId` in the key is redundant given the map is per-sync, agreed — I've kept it because `byConflictKey` directly above uses the identical format, and having the two silently diverge is a worse trap than a redundant segment.

Issue #4513, opened 2026-08-16
Teardown wedges on a disabled subscription during the WAL emergency


marcodejongh:
Found during independent review of #4474. Pre-existing, not introduced there, and not a reason to hold that PR.

## The wedge

`scripts/neon-to-railway-replication.sh` teardown issues two autocommitted statements:

```sql
ALTER SUBSCRIPTION <name> DISABLE;
DROP SUBSCRIPTION <name>;
```

They cannot share a transaction — a `DROP SUBSCRIPTION` that must drop a remote slot is not allowed inside a transaction block.

If the second statement fails, the subscription is left **disabled**. On re-run, `assert_subscription_contract` requires `subscription.subenabled`, so the contract check fails and teardown refuses to drop anything at all.

The failure mode that leaves it disabled is the publisher being unreachable — which is frequently the same incident that made you run teardown in the first place. `docs/neon-migration.md` §6.2 exists for exactly that case: *"Neon retains the logical-replication slot and accumulates WAL until disk fills."*

## Evidence it is a real path

`scripts/postgres18-image-smoke.sh` already works around it in its own manual cleanup, using the escape hatch the helper does not:

```sql
ALTER SUBSCRIPTION <name> SET (slot_name = NONE);
DROP SUBSCRIPTION <name>;
```

## Options

- Let the subscription contract accept `subenabled = false` on the teardown path specifically. It is being dropped; enabled-ness proves nothing about identity, which is already established by owner, name, slot and publication equality.
- Or have teardown detect the disabled-and-orphaned shape and use `SET (slot_name = NONE)` before the drop, then clear the slot separately — matching what the smoke does by hand.

The first is smaller and consistent with the change in #4474 that stopped gating replication-object teardown on conditions that do not bear on whether the drop is safe.

## Also worth adding

The fast fake-psql tier has no case for a live subscription plus a drifted subscriber role — the behaviour is currently pinned only by the Docker-gated image smoke. A case setting `FAKE_SUBSCRIPTION_EXISTS`/`FAKE_PUBLICATION_EXISTS`/`FAKE_SLOT_EXISTS` with a violating contract, asserting the replication objects are dropped and `ROLE_STATEMENT_LOG` stays empty, would pin it without Docker. Note the fixture markers are shared across cases, so such a case has to reset them or run last.

## Release Notes

none

PR #4697, opened 2026-08-23
docs(db): hand over the PostgreSQL 18 rollout

marcodejongh:
## Summary

Handover for continuing the PG18 rollout on a machine that can run Docker. Adds `docs/postgres-18-rollout-handover.md`, the PostGIS audit findings, the accepted image digests, and the deferred replication-script rename patch.

**Everything through image publication and consumer pinning is merged.** The production cutover has not started. One read-only audit is the only thing that has touched Railway.

## The two blockers

**`wal_level = replica`** — logical replication needs `logical`, which costs a Railway Postgres restart.

**PostGIS version.** Production runs `3.7.0dev` (from the `postgis/postgis:16-master` tag); the attested image ships stable `3.6.4`. PGDG has no stable 3.7 for PG18, so the target cannot be moved up to match:

```
3.6.2+dfsg-1.pgdg12+1
3.6.3+dfsg-1.pgdg12+1
3.6.4+dfsg-2.pgdg12+1   <- newest
```

But the actual exposure is two `geography(Point,4326)` columns and two partial GiST indexes, with `tiger`/`topology` empty scaffolding (0 live tuples). Nothing uses a 3.7-only feature. Full evidence in `docs/postgres-18-postgis-audit.md`.

**The decision was not taken** — narrow the audit rule, prove it with a restore rehearsal, or move production off the `master` tag. The rehearsal is the natural first task once Docker works, since it settles the question with evidence *and* produces the §4 rehearsal the Phase 4 checkpoint requires.

## Also captured

- Accepted digests and attestation URLs (`docs/postgres-image-digests.json`)
- Why two publish attempts died: the exact-main recheck runs *after* both builds, and the seeded image takes ~17 min of that ~19 min window
- The `prevent_self_review = false` accepted risk and the condition for reverting it
- `docs/pg18-replication-rename.patch`, deferred from commit A
- Environment traps that produced three real bugs: `ugrep` vs GNU grep, no `ripgrep` on runners (which made a credential guard *fail-open*), and BSD-only `stat -f`

## Not included

`agent/split-seeded-image-publisher` (`9f0a602c1`) is pushed separately and is **WIP, unreviewed, do not merge** — the security review had not run and the contract hashes are unconfirmed.

## Release Notes

none

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -65,6 +65,13 @@ jobs:
       # Defaults false off PRs (the OTA bot legitimately rewrites it on push).
       changelogData: ${{ github.event_name == 'pull_request' && steps.filter.outputs.changelogData || 'false' }}
       dbMigrations: ${{ github.event_name != 'pull_request' && 'true' || steps.filter.outputs.dbMigrations }}
+      # docs/postgres-image-digests.json and docs/*.patch are checked-in PG18
+      # rollout artifacts that nothing reads at run time, so nothing notices
+      # when the tree drifts under them. Their guard reads every input with
+      # readFileSync/readdirSync, which `--changed` can never relate to a diff
+      # of a workflow pin or of a file the patch's context depends on. No other
+      # filter matches docs/** at all. Gates the `pg18-artifacts` job below.
+      pg18Artifacts: ${{ github.event_name != 'pull_request' && 'true' || steps.filter.outputs.pg18Artifacts }}
       # True if any downstream job's gate would fire. This used to gate the
       # `setup` job (bun install + build:db + a `built-packages` artifact every
       # heavy job downloaded); that job was deleted because @boardsesh/db and
@@ -76,10 +83,10 @@ jobs:
       # ci-location-sync-workflow.test.ts asserts on its terms; keep it in sync
       # with the job `if:`s below so a future job can gate on it again.
       runAny: ${{ github.event_name != 'pull_request' || steps.filter.outputs.anyJs == 'true' || steps.filter.outputs.web == 'true' || steps.filter.outputs.i18nCatalogs == 'true' || steps.filter.outputs.sharedSchema == 'true' || steps.filter.outputs.sharedDeps == 'true' || steps.filter.outputs.backend == 'true' || steps.filter.outputs.mobile == 'true' || steps.filter.outputs.ocr == 'true' || steps.filter.outputs.locationSync == 'true' || steps.filter.outputs.rootCi == 'true' }}
-      # NOTE: deployConfig, storeMetadata, mobileLocales, and restSurface are
-      # deliberately NOT part of runAny — the deploy-config, listing-guards and
-      # rest-surface jobs below gate on their own filters, so a PR touching only
-      # their files runs nothing else.
+      # NOTE: deployConfig, storeMetadata, mobileLocales, restSurface and
+      # pg18Artifacts are deliberately NOT part of runAny — the deploy-config,
+      # listing-guards, rest-surface and pg18-artifacts jobs below gate on their
+      # own filters, so a PR touching only their files runs nothing else.
     steps:
       # No actions/checkout here on purpose: on `pull_request` dorny/paths-filter
       # reads the changed-file list from the REST API (no working tree needed),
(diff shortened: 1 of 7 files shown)


marcodejongh:
Answering the three review findings, and folding in what re-reading the tooling turned up.

**1 — sensitive production details. Declining to redact, with the reasoning written into §3.**

The repo is public, so I judged each item on that basis rather than the "even if private" hedge. None of the five is a credential, and four have committed public precedent already:

| Flagged | Verdict | Why |
| --- | --- | --- |
| System identifier `7635874554056458274` | Keep | Minted by `initdb` from timestamp+PID, readable from `pg_controldata`, present in every WAL header, returned by `IDENTIFY_SYSTEM` to any replication connection. Authenticates nothing. It is also the only cheap way for the next operator to prove the cluster they are cutting over is the one this audit ran against — a Railway service recreate mints a new one and silently invalidates every number in §3. |
| 1Password `Boardsesh` → `RAILWAY Postgres PROD (readonly)` | Keep | A vault+item *name* is a pointer, worthless without 1Password auth. `docs/aurora-sync.md:295` already publishes `op://Boardsesh/Postgres PROD/...`, and `packages/aurora-sync/.env.1password:1` the same. Naming it stops the next operator minting a second, untracked readonly credential. |
| `data_checksums: off` | Keep | Durability property, not access control. Also load-bearing: `docs/postgres-18-migration.md` makes the *target* assert `data_checksums = on`, so source-off vs target-on is a gate the next operator hits and must be warned about. (It is not itself a blocker — the audit requires it of the target only.) |
| App connects as `postgres`; the three restricted roles do not exist | Keep, **reworded** | This is the one item with real recon value. But it is already public in `docs/neon-migration.md:96`, an attacker holding the connection string has superuser regardless of what the doc says, and deleting the sentence removes the pressure to fix without removing the exposure. §5 step 5 now states it as *the blocking security item*, not as a description of a steady state. |
| DB name `railway`, size, `boardsesh_readonly`, schema/row counts | Keep | `docs/neon-migration.md:159-160` publishes `SOURCE_DATABASE_NAME='railway'`; `docs/power-user-analysis.md:7` publishes a whole production analysis sourced from `boardsesh_readonly`. |

§3 now opens by saying no credential appears in the file and why the identifier is there, which is the substance of the concern without losing the audit's value.

**2 — the patch will drift. Accepted, and it already has.**

It still applies against current `main`, but the `vite.config.ts` hunk now lands at **offset +15 lines**, absorbed silently by `git apply`'s context matching. My own edit to `docs/postgres-18-migration.md` in this PR pushed its four hunks to +22. The next edit *inside* that one-line `test:postgres18-contract` command string turns it into a hard reject with no warning to anyone. §2 now says so, and a `git apply --check` gate over `docs/*.patch` is going into `scripts/postgres18-workflow-contract.test.sh` — it already owns the "this rename must move atomically" contract, so it needs no new job or filter. Worth noting the anchor it runs behind does not currently fire for a `docs/`-only PR; that entry goes in with the gate.

**3 — #4513 missing from §5. Accepted; it is now step 1.**

Confirmed the mechanism: `teardown_replication` calls `assert_subscription_contract skip-subscriber-role-shape`, and that query requires `AND subscription.subenabled`. Teardown's own first statement is `ALTER SUBSCRIPTION … DISABLE`. So a drop that fails because the publisher is unreachable — the same incident that made you run teardown — leaves the subscription disabled, and every re-run refuses to drop anything while the source keeps retaining WAL. The comment right above that function already reasons about not stranding the WAL-emergency path for the *role* check; the `subenabled` predicate was missed.

---

Beyond the review, re-reading the tooling corrected several claims in the doc itself:

- **The PostGIS gate is four enforcement points, not one**, and only two have a knob — `postgres-migration-audit.sh:1943` and `neon-to-railway-replication.sh:1412` compare the whole extension manifest with the version in it, and neither can be overridden. Anyone budgeting one edit to narrow that rule is wrong by 4×.
- **`docs/postgres-18-migration.md` said "There is no override flag"** while `postgres-migration-audit.sh:16` defaults `EXPECTED_POSTGIS_VERSION` from the environment. Both files now say what that variable is for and why it is not the way past the gate.
- **tiger/topology are not ignorable.** The extension manifest is cluster-wide with no allowlist, so dropping them on the source is the only path that clears all three gates unchanged.
- **§5 was missing the image pin.** Railway tracks the mutable `postgis/postgis:16-master` tag with auto-updates on, so the `wal_level` restart can silently pull a newer `3.7.x-dev` and invalidate the whole audit. Pinning has to come first.
- Settled the open P2 on `agent/split-seeded-image-publisher`: the two `TS2339` `unref` errors in `scripts/expo-web-e2e.ts` reproduce on a docs-only branch off `main`, so the split does not introduce them.

Also fixed separately in #4698: #4695 pinned `docker-compose.yml` to the seeded digest but not the PG18 volume contract, which left `vp run db:up` broken on `main` for everyone on the local Docker path. Noted in §2 because it is the first thing anyone picking this up will run.

PR #4706, opened 2026-08-23
feat(db): gate PG18 on PostGIS capability, not version string

marcodejongh:
## Summary

This PR replaces the PG18 migration's PostGIS version-equality gate (which could never be satisfied, since production tracks a mutable `3.7.0dev` tag with no stable PGDG build for PG18 and no PostGIS downgrade path) with a capability comparison. It checks that the target PostGIS actually provides every type, operator class, operator, and function the source database uses — including functions only reachable from inside plpgsql routine bodies via a textual scan, since a `pg_depend`-only check misses those entirely. Three rounds of Fable review caught real regressions that got fixed: a `setup` step that silently lost its PostGIS version check, a shadow-name bug where a same-named non-postgis routine could hide a real postgis function, and domain-type column handling that missed nested/array domains and would have wedged the cutover. It also adds the repo's first spatial fixture to `postgres18-image-smoke.sh` (geography columns, indexes, a trigger, a domain column) with negative-case tests proving each kind of capability gap aborts `setup` before the restore runs. The PR is marked draft pending three follow-ups: running the image smoke test in real CI (only run locally via a `psql` shim), pinning the Railway PostGIS image so the rehearsal matches production, and manually dropping the unused `postgis_topology`/`postgis_tiger_geocoder` extensions in production.

---

Resolves Blocker 2 of the PG18 rollout (`docs/postgres-18-rollout-handover.md` §4).

## Why the old rule had to go

`docs/postgres-18-migration.md` §1 required the source and target PostGIS versions to be **equal**. That gate could not be satisfied and could not be honestly waived:

- Production reports `3.7.0dev`, only because the Railway service tracks the mutable `postgis/postgis:16-master` tag.
- PGDG publishes no stable 3.7 for PostgreSQL 18, so the target cannot be raised.
- PostGIS ships no downgrade script, so the source cannot be lowered.

Neither side could move to meet the other. What remained was a gate with no supported way through and an override that would have been pure judgement.

The decision needs a narrower question, and [#4700](https://github.com/boardsesh/boardsesh/pull/4700) answered it with evidence: 21 of 21 checks matched across the real 3.7.0dev → 3.6.4 step, including a byte comparison of `ST_AsEWKB` over every populated geography.

## What replaces it

A capability comparison. The audit enumerates what the source actually uses and requires the target to provide each item, across four dimensions: **types** (rendered with `format_type`), **operator classes**, **operators**, and **functions** — the last from `pg_depend` **and** from a textual scan of routine bodies.

That last half is load-bearing. Migration 0127 puts `ST_MakePoint(...)::geography` inside a plpgsql body, which PostgreSQL stores as opaque text with no dependency edge. **A check built on `pg_depend` alone reports this database uses no PostGIS functions at all.**

What stays strict, deliberately: the **target** must still equal `EXPECTED_POSTGIS_VERSION` exactly. It's our own attested artifact; there's no reason to accept anything else there, and that exactness is what makes source tolerance sound.

## What review caught

Three Fable reviewers, all three returned CHANGES_REQUIRED. Every P2 was reproduced against a live PG18.4/PostGIS 3.6.4 container before being fixed. Nothing was declined. Three findings are worth calling out:

**`setup` silently lost every PostGIS version check.** The extension manifest used to pin the target as a side effect; removing version comparison from it removed that too. A genuine regression introduced by this change, invisible without someone going looking. `EXPECTED_POSTGIS_VERSION` is now asserted in the replication helper *before* it asks the target anything.

**The shadow-name fail-open was not hypothetical.** A routine-body hit was dropped whenever any non-PostGIS routine shared the name — and `postgis_topology` really does ship its own `st_srid`, `st_simplify` and `st_geometrytype`, confirmed in-container. The filter was cluster-wide, so excluding the `topology` schema did not protect it. The rule is now: keep when postgis provides the name; keep when nothing does (→ unresolved, a blocker); drop only when it resolves exclusively outside postgis.

**Two P2s were worse than reported.** The domain-over-geography column was tokenised as the *domain's* name, which would have demanded a type only the restore could create — a permanent cutover wedge. Probing found two more the reviewers missed: a domain over a domain, and an array of a domain, matched no branch and vanished from the manifest entirely — fail-open, not deadlock. Replaced with a `WITH RECURSIVE` walk to any depth.

## The first spatial fixture this repo has had

`postgres18-image-smoke.sh` previously created the `postgis` extension and not one spatial object, so the rule it enforces had zero coverage. It now carries geography columns, partial GiST indexes, a plpgsql trigger, a domain column, an `ST_AsText` expression index and an `&&` predicate — plus negative cases proving a type, opclass, operator or function gap, and a wrong target version, each abort `setup` **before** the restore.

The manifest it produces, source 3.7.0dev against target 3.6.4:

```
('function', 'st_astext',    'catalog',      'index pg18_smoke_gyms_wkt_idx')
('function', 'st_makepoint', 'routine-body', 'public.pg18_smoke_set_location')
('opclass',  'gist/public.gist_geography_ops', 'catalog', ...)
('operator', 'public.&&(geography,geography)', 'catalog', 'index pg18_smoke_gyms_bbox_idx')
('type',     'geography(Point,4326)', 'catalog', ...)
Target PostGIS satisfies every capability in the source spatial manifest.
```

`st_makepoint` stays `routine-body` even with a decoy `public.st_makepoint(text)` present — that's the shadow-name regression test.

## Known limits, stated rather than buried

The body scan is textual: an operator or oddly-named function reachable **only** from inside a plpgsql body is invisible to both halves. Nothing here does that today; the `&&` fixture and `postgres18-spatial-surface.test.sh` make it loud if a migration introduces it. Satisfaction is by **name, not signature** — the scan cannot resolve overloads and claiming otherwise would be a lie. Typmod stays the post-restore DDL manifest's job. And the largest gap is structural: anything the app does with PostGIS from outside the database is not catalog-resident, and is covered only by the repo-side `ST_*` grep.

It fails closed where it matters: an empty manifest while postgis is installed, any `st_*`/`postgis_*` name resolving to no routine, any unrecognised capability kind (`ELSE false`), and any query error. The audit stays read-only; manifest cost is ~12 ms against a 30 s `statement_timeout`.

## Draft — three things should be true before this merges

1. **The image smoke must be seen green in CI.** It ran here only through a `psql` shim forwarding into a container, because this host has no `psql`. A runner with a real one is the honest gate.
2. **The Railway image should be pinned first.** The whole rule rests on the rehearsal, and that ran against a `postgis:16-master` build ~280 commits newer than production reports. Until the service is pinned with auto-updates off, the rehearsed source and the real source are not the same database.
3. **`postgis_topology`/`postgis_tiger_geocoder` still block the extension manifest by design.** Dropping them on the source (0 live tuples) remains an unexecuted human production step. Nothing here makes the cutover possible without it.

## Release Notes

none

Contract evidence: `postgres18-workflow-contract.test.sh` 27/27 covered, image smoke negatives abort `setup` pre-restore, drift suite 9/9. The prepared rename patch this PR used to regenerate no longer exists — #5060 landed the rename on main and this branch was rebased across it.

## Test plan

1. CI green.

## Risk

Risk: 3/5 — changes the PG16→18 migration audit and replication gating; no production service touched.

https://claude.ai/code/session_01SAsksAJopQNZQwYQQbz7sn


marcodejongh:

Conflicting: docs/pg18-replication-rename.patch and vite.config.ts — pre-flagged DIRTY by GitHub. This branch renames the neon-to-railway-replication scripts to postgres-logical-replication, and main independently touched the same script list — a naming race, not a pnpm issue.

Issue #4633, opened 2026-08-20
[UX] Climb: Routes: Allow slower playback


ES-Alexander:
The current speed multiplier is limited to a minimum of 0.5x, which can still end up very fast on some climbs.

1. At minimum the base speed should be lowered, or that lower limit could be changed down to something like 0.1x, but preferably the speed control would be changed to "seconds per frame" (with linear control between say 0.5 seconds/frame and 10 s/f) instead of an ambiguous multiplier that seems to result in different speeds depending on the number of frames in the route.

2. It can also take a while to get started on a climb after pressing play (unless you get onto it first and ask a friend to start the playback), so it would be helpful to have a separate control for startup delay (e.g. from 0-30 seconds)

marcodejongh:
Second report of this, from a climber who wrote in by DM. Same ask as point 1
here — a real unit instead of a multiplier:

> if technically possible, I'd love if boardsesh was able to standardise the pace
> of how fast led's on a route show up, especially if it is in a known metric
> rather than "1x". At least I personally find it so tedious that I've given up
> on routes on kilter board for that reason, so if boardsesh could lower the
> barrier of entry it'd be insanely helpful for endurance training on Kilter.

Where mobile stands today: `PlaybackControls` already has a seconds-per-frame
pill, but its `paceUnit` union only exposes it to the **creator**
(`CreateRoutePlaybackSlot`), which authors the climb's `frames_pace` directly.
`PlayDrawer` still passes the multiplier lens, so a reader sees "1×". So the
remaining work for point 1 is exposing seconds on the play side, not building the
control.

Point 2 (startup delay before the first frame) is untouched.

Filed the same climber's other request as #5378.

PR #4708, opened 2026-08-23
feat: change min route playback speed to 0.1x

ES-Alexander:
## Summary

A minimal fix to allow climb routes to be played back slower. There are better solutions long-term (per #4633), but this should at least make the playback feature usable in the meantime.

## Release Notes

- Play back routes up to 5x slower (down to 0.1x)

## Test plan

- Confirmed working in the OTA preview


ES-Alexander:
Failing tests are unrelated to the content of this PR. The CI test report seems related to Node 20 dependency issues for the testing, and the Claude review is because my PR is from a fork.

In future, would it be preferable if I make a new branch in the root repo and PR from that instead of my fork?

PR #4709, opened 2026-08-24
fix(mobile): retry and cap the playlist page drain so big smart playlists start

### LINKED ISSUE #4622: Starting a large smart playlist fails: the unpaced page drain trips the smartPlaylist rate limit
Starting a smart playlist drains **every** page of it in one unpaced burst, and the drain is
all-or-nothing. One rate-limited or dropped page kills the whole replacement, the user gets
"Couldn't start playlist. Try again." and the queue is left untouched. Retrying starts from page 0
and burns the same budget again.

Found while investigating #3966 (a different, unrelated defect on the same screen family). Filing
separately per the triage decision there.

## Sentry evidence (org `boardsesh`, last 90 days)

Scoped to `source:playlist op:replace-queue` — the tag `use-playlist-activation.ts:414` attaches to
the catch-all of `replaceQueueWithPlaylist`:

| `error.type` | events | users affected |
| --- | --- | --- |
| `Error` | 26 | 5 |
| `TypeError` | 21 | 9 |
| **total** | **47** | **13** |

(13, not 18 — the two groups share users, so the per-row counts must not be summed.)

Breakdown of the 26 `Error` events: exactly 23 are `Rate limit exceeded. Try again in N seconds.`
with `path: ["smartPlaylist"]`, `extensions.code: RATE_LIMITED`, `operation: smartPlaylist`
(5 unique users); the other 3 are Cloudflare `GraphQL Error (Code: 502)`, all from 2026-08-18. The
21 `TypeError` events are all `TypeError: Network request failed`.

The 23 rate-limited events carry these `retryAfterSeconds` values (full set, ascending): 2, 2, 4, 5,
6, 6, 6, 7, 7, 13, 16, 16, 22, 23, 24, 24, 30, 31, 32, 32, 34, 38, 46. They land in six bursts —
07-15 16:23, 07-16 03:48, 07-17 12:51, 07-20 09:38, 07-21 04:52, 07-28 07:49 — across 5 users and
both platforms (`app:///main.jsbundle` and `app:///index.android.bundle`), i.e. a repeating pattern
rather than one user in one bad minute.

Note the rate-limited events span **2026-07-15 → 2026-07-28** and stop there; nothing in that class
has fired in the ~3 weeks since. Worth establishing whether something changed (traffic, playlist
sizes, a client change) before assuming the ceiling is still being hit today. The
`TypeError: Network request failed` half is the one still firing recently — last seen
2026-08-17.

Discover query:
https://boardsesh.sentry.io/explore/discover/homepage/?dataset=errors&queryDataset=error-events&query=source%3Aplaylist+op%3Areplace-queue&field=error.type&field=count%28%29&field=count_unique%28user%29&sort=-count%28%29&statsPeriod=90d&mode=aggregate&yAxis=count%28%29

## Mechanism

`fetchAllClimbsForBoard` in `packages/mobile/src/lib/playlists/use-playlist-activation.ts:299-327`
is an unbounded serial drain:

```ts
while (hasMore && !signal.aborted) {
  const pageResult = await fetchPage({ page, pageSize: PLAYLIST_SUGGESTION_REFRESH_PAGE_SIZE, board, signal });
  climbs.push(...pageResult.climbs);
  hasMore = pageResult.hasMore;
  page += 1;
}
```

`PLAYLIST_SUGGESTION_REFRESH_PAGE_SIZE` is 100
(`packages/shared/playlists-react/src/fetch-playlist-suggestion-climbs.ts:3`). On the smart-playlist
screen (`packages/mobile/app/(tabs)/discover/smart/[type].tsx:51-70`) each iteration is one
`smartPlaylist` GraphQL query. The server-side cap is
`await applyRateLimit(ctx, 60, 'smartPlaylist')`
(`packages/backend/src/graphql/resolvers/playlists/queries/smart-playlists.ts:274`) — 60 requests per
window keyed `userId:smartPlaylist`, and that bucket is **global** per user across both the
in-process and Redis tiers, not per instance.

So a smart playlist over ~6000 climbs cannot be started at all, and one over a few thousand can't be
started twice in a window. There is no pacing between pages, no per-page retry, and no backoff on a
`RATE_LIMITED` extension the server is explicitly handing back with a `retryAfterSeconds`.

The `TypeError: Network request failed` half is the same shape without a server in the loop: a long
serial burst on a gym wifi / LTE handoff drops one page and the whole replacement dies.

Two smaller things noticed in the same code:

- The smart-playlist `fetchPage` (`smart/[type].tsx:51`) accepts `{ page, pageSize, board }` and
  never forwards a `signal` to `getHttpClient().request`, so `replacementAbortRef.abort()` stops the
  loop from *issuing* the next page but does not cancel the in-flight one. The regular playlist
  `fetchPage` has the same gap.
- The whole drain happens before anything reaches the queue, so a 3000-climb smart playlist blocks on
  30 sequential round trips with only `isReplacingQueue` for feedback.

## Suggested direction (not prescriptive)

- Honour `extensions.retryAfterSeconds` on a `RATE_LIMITED` page instead of throwing — sleep and
  resume the drain from the failed page rather than restarting.
- Retry a failed page a bounded number of times before giving up on the whole replacement.
- Cap the drain (seed the queue with the first N pages, keep paging in the background) so a huge
  smart playlist starts playing immediately instead of failing.
- Consider whether 60/window is the right budget for an operation whose own client pages it, or
  whether the drain should ask for a larger page size.

## Not this issue

The `replace-queue-empty` canary (the #3891 MoonBoard one-climb-queue class) has **not** fired in 90
days — that regression is genuinely gone. This is only the pagination/rate-limit path.

marcodejongh:
## Summary

Starting a playlist read **every** page of it in one unpaced burst, inside one
`try`/`catch`. One rate-limited or dropped page threw away every page already
fetched, the climber got "Couldn't start playlist. Try again.", and retrying
restarted at page 0 — spending the same 60-per-minute `smartPlaylist` budget
over again. A playlist past ~6000 climbs (60 pages of 100) could not be started
at all.

Sentry, `source:playlist op:replace-queue`, last 90 days: 47 events / 13 users.
23 of them are `Rate limit exceeded. Try again in N seconds.` with
`extensions.code: RATE_LIMITED` (hints from 2s to 46s); 21 are
`TypeError: Network request failed`, the half still firing.

The hand-rolled `while (hasMore)` loop is replaced with `drainPlaylistPages` in
`@boardsesh/playlists-react`:

- **Bounded** at 30 pages / 3000 climbs — half the server's per-minute budget, so
  two activations still fit in one window. Also closes the latent infinite loop
  on a `hasMore`-forever resolver.
- **Resumable.** A retried page re-requests the *same* page number and keeps
  everything already accumulated. It honours the server's
  `extensions.retryAfterSeconds`, under one 60s wall-clock sleep budget for the
  whole drain, so a replacement can never hang indefinitely.
- **Partial beats nothing.** If any page landed, the queue is replaced with what
  we have. A drain that read *no* page still toasts and leaves the queue exactly
  as it was — for every stop reason, not just a rejection.

The retry policy is injected (`createPlaylistPageRetryPolicy`) rather than
imported, so the shared package keeps no transport or error-shape dependency and
`package.json` / `bun.lock` are untouched. It re-sends only two classes:
`RATE_LIMITED` (the resolver calls `applyRateLimit` before doing any work, so a
re-send cannot double anything) and a transport drop that never reached the
server. GraphQL validation errors, 4xx and programmer bugs fail the page
immediately instead of burning the climber's budget. This stays scoped to the
drain on purpose — the app-wide policy is the opposite (`query-provider.tsx`:
never retry a `RATE_LIMITED` rejection), because retry is only correct for
requests the client itself is fanning out and can pace.

It is a *factory* built once per page, not a plain function, because each class
of failure needs its own budget. With one shared counter a wifi-to-LTE handoff
dropping page 7 spent the page's only rate-limit retry, so the throttle that
followed gave up immediately — an unrelated blip truncating a playlist that one
honest wait would have finished.

Three smaller fixes in the same path:

- Both playlist screens now forward their `AbortSignal` into
  `getHttpClient().request({ document, variables, signal })`. Leaving a playlist
  mid-drain used to stop the loop from *issuing* the next page but left the
  in-flight one running.
- The swipe-track refresh (`fetchPlaylistSuggestionClimbs`) gets the same
  per-page backoff *and the same total wait ceiling* — it shares the same server
  bucket, and a fire-and-forget prefetch that can sleep on all ten of its pages
  stays pending for minutes. Out of budget it stops with the climbs it has. Both
  parameters are optional and default to today's exact behaviour, so no other
  caller changes.
- Cancel and pan-to-close stay live on the queue-replace warning sheet while the
  replacement runs. Confirming now genuinely starts the drain, and a throttled
  first page can back off for the length of a server window — locking the
  climber into a modal for that long reads as a hung app. `onCancel` already
  aborted the drain *and* its sleep; only the greyed-out button made it
  unreachable. Confirm still shows a spinner, so a double-tap can't start two
  drains.

A truncated or partially drained replacement reports to Sentry as
`op: replace-queue-partial` with `stoppedBy` / `pagesFetched` / `climbCount`, so
the 30-page cap and the rate-limit stops are measurable before anyone tunes them.
Deliberately silent for the climber: a warning toast on an action that just
worked is noise, and 3000 climbs is far past a session.

**Native fingerprint: not moved.** TypeScript only — no new dependency, no Expo
plugin, no native module, no `app.config.ts` change. This ships over OTA.

Closes #4622

## Release Notes

- Big playlists start again — a slow or dropped page no longer kills the whole thing
- Tapping a climb in a huge circuit now fills your queue instead of erroring out
- Backing out of a playlist stops it loading straight away
- Cancel still works while a playlist is loading into your queue

## Test plan

New unit coverage:

- `packages/shared/playlists-react/src/__tests__/drain-playlist-pages.test.ts` —
  full drain, page cap (exactly 30 calls / 3000 climbs), **resume from the failed
  page** (asserts the request sequence is `[0, 1, 2, 2, 3]`, never a second `0`),
  give-up-with-partial, wait-budget exhaustion, abort mid-drain, aborts never
  retried, no-policy fail-fast, plus `abortableSleep` settling on abort instead of
  waiting the backoff out.
- `packages/mobile/src/lib/playlists/__tests__/playlist-page-retry.test.ts` — real
  graphql-request `ClientError` fixtures: honours `retryAfterSeconds: 46`, falls
  back when the server sent no hint, retries a transport drop twice, never
  retries an abort or a `BAD_USER_INPUT`; keeps a separate budget per error class
  when the classes alternate; and pins the rate-limit ceiling — jitter included —
  at or under the drain's total wait budget.
- `packages/shared/playlists-react/src/__tests__/fetch-playlist-suggestion-climbs.test.ts`
  — the swipe-track refresh resumes a dropped page, stops on its total wait
  budget with the climbs it already had, and still propagates a rejection the
  policy gave up on.
- `packages/mobile/src/lib/playlists/__tests__/use-playlist-activation.test.tsx` —
  partial drain still replaces the queue with no toast and one
  `replace-queue-partial` report; page-cap ditto; a zero-page stop toasts and
  leaves the queue alone on both the tap path and the confirm-after-warning path
  (which is the one that never pre-seeds a queue), without tripping the #3891
  empty-fetch canary; an aborted drain touches nothing; the drain is asked for
  the 30-page cap, the shared page size and a retry-policy factory.
- `packages/mobile/src/components/playlist/__tests__/playlist-queue-replace-sheet.test.tsx`
  — cancel and pan-to-close stay live while replacing, and cancel fires mid-flight.

Commands (all green):

- `vp test run --project playlists-react --reporter=agent` — 6 files / 39 tests passed
- `vp run test:mobile` — full suite green
- `vp run typecheck:mobile`, `vp run typecheck:shared`, `vp run typecheck:playlists-react`
- `vp check` — 0 errors

Device QA is still owed — `.boardsesh/qa-notes.md` (gitignored, in the worktree)
lists the flows: big smart playlist starts, back-out cancels the in-flight page,
rate-limited drain still fills the queue, airplane mode mid-drain, and the
regular playlist warning sheet. Note that `applyRateLimit` returns early when
`NODE_ENV === 'development'`, so reproducing the rate-limit path needs a build
pointed at a non-development backend.

### DIFF
--- a/packages/mobile/app/(tabs)/discover/[playlist_uuid].tsx
+++ b/packages/mobile/app/(tabs)/discover/[playlist_uuid].tsx
@@ -121,10 +121,12 @@ export default function PlaylistDetail() {
       page,
       pageSize,
       board,
+      signal,
     }: {
       page: number;
       pageSize: number;
       board: { boardName: string; layoutId: number; sizeId: number; setIds: string; angle: number };
+      signal: AbortSignal;
     }) => {
       const input: GetPlaylistClimbsInput = {
         playlistId: playlistUuid,
@@ -136,9 +138,14 @@ export default function PlaylistDetail() {
         page,
         pageSize,
       };
+      // Object overload so the abort signal reaches fetch: leaving a playlist
+      // mid-drain must cancel the request in flight, not just stop the next one.
       const response = await getHttpClient().request<GetPlaylistClimbsQueryResponse, { input: GetPlaylistClimbsInput }>(
-        GET_PLAYLIST_CLIMBS,
-        { input },
+        {
+          document: GET_PLAYLIST_CLIMBS,
+          variables: { input },
+          signal,
+        },
       );
       return {
         climbs: toQueueClimbs(response.playlistClimbs.climbs),
--- a/packages/mobile/app/(tabs)/discover/smart/[type].tsx
+++ b/packages/mobile/app/(tabs)/discover/smart/[type].tsx
@@ -55,7 +55,17 @@ export default function SmartPlaylistDetail() {
   // Suggestion-refresh fetcher pages the smart playlist scoped to the active
   // board name so the play-drawer swipe walks the full computed list.
   const fetchPage = useCallback(
-    async ({ page, pageSize, board }: { page: number; pageSize: number; board: { boardName: string } }) => {
+    async ({
+      page,
+      pageSize,
+      board,
+      signal,
+    }: {
+      page: number;
+      pageSize: number;
+      board: { boardName: string };
+      signal: AbortSignal;
+    }) => {
       const input: GetSmartPlaylistInput = {
         type: smartType,
         userId,
@@ -63,10 +73,13 @@ export default function SmartPlaylistDetail() {
         page,
         pageSize,
       };
-      const response = await getHttpClient().request<GetSmartPlaylistQueryResponse, { input: GetSmartPlaylistInput }>(
-        GET_SMART_PLAYLIST,
-        { input },
-      );
+      // Object overload so the abort signal reaches fetch: leaving a playlist
+      // mid-drain must cancel the request in flight, not just stop the next one.
+      const response = await getHttpClient().request<GetSmartPlaylistQueryResponse, { input: GetSmartPlaylistInput }>({
+        document: GET_SMART_PLAYLIST,
+        variables: { input },
+        signal,
+      });
       return {
         climbs: toQueueClimbs(response.smartPlaylist.climbs),
         hasMore: response.smartPlaylist.hasMore,
--- a/packages/mobile/src/components/playlist/PlaylistQueueReplaceSheet.tsx
+++ b/packages/mobile/src/components/playlist/PlaylistQueueReplaceSheet.tsx
@@ -18,8 +18,14 @@ type PlaylistQueueReplaceSheetProps = {
 // Confirmation before a playlist tap replaces the whole queue: replacing clears
 // the climbs queued after the current one, so warn first. Built on ModalSheet
 // (the shared @expo/ui native sheet wrapper) so it goes through the presentation
-// coordinator and never overlaps another native sheet. Pan-to-close is locked
-// while the replacement is in flight; the buttons drive the decision.
+// coordinator and never overlaps another native sheet.
+//
+// Cancel stays live while the replacement is in flight, and so does pan-to-close.
+// Confirming here genuinely starts the page drain, and a throttled first page can
+// now back off for the length of a server window (#4622) — locking the climber
+// into a modal for that long would read as a hung app. `onCancel` aborts the
+// in-flight drain and its sleep, so leaving is instant. Only Confirm is blocked
+// while replacing (it shows a spinner), so a double-tap can't start two drains.
 export function PlaylistQueueReplaceSheet({
   visible,
   futureQueueCount,
@@ -31,7 +37,7 @@ export function PlaylistQueueReplaceSheet({
   const { systemColors } = useTheme();
 
   return (
-    <ModalSheet visible={visible} enableDynamicSizing enablePanDownToClose={!isReplacing} onClose={onCancel}>
+    <ModalSheet visible={visible} enableDynamicSizing onClose={onCancel}>
       {/* The footerless ModalSheet wrapper composes insets.bottom onto the body
           (withSheetBottomInset), so this content only adds its own spacing. */}
       <View style={styles.content}>
@@ -51,7 +57,6 @@ export function PlaylistQueueReplaceSheet({
             variant="outlined"
             role="cancel"
             onPress={onCancel}
-            disabled={isReplacing}
             style={styles.button}
           />
           <Button
--- a/packages/mobile/src/components/playlist/__tests__/playlist-queue-replace-sheet.test.tsx
+++ b/packages/mobile/src/components/playlist/__tests__/playlist-queue-replace-sheet.test.tsx
@@ -30,7 +30,9 @@ type ModalSheetMockProps = {
 // ModalSheet presents imperatively in the real app; here we render its children
 // only while `visible` and expose a synthetic close button to exercise onClose.
 vi.mock('../../ModalSheet', () => ({
-  ModalSheet: ({ children, visible, enableDynamicSizing, enablePanDownToClose, onClose }: ModalSheetMockProps) =>
+  // `enablePanDownToClose` defaults to true in the real ModalSheet, so the mock
+  // has to as well or "did this component lock the sheet?" reads backwards.
+  ModalSheet: ({ children, visible, enableDynamicSizing, enablePanDownToClose = true, onClose }: ModalSheetMockProps) =>
     visible
       ? createElement(
           'div',
@@ -141,14 +143,26 @@ describe('PlaylistQueueReplaceSheet', () => {
     expect(onConfirm).toHaveBeenCalledTimes(1);
   });
 
-  it('locks pan-to-close and shows loading while replacing', () => {
+  // #4622: confirming genuinely starts the page drain, and a throttled first
+  // page can now back off for the length of a server window. A locked sheet with
+  // a greyed-out Cancel for that long reads as a hung app, and `onCancel` aborts
+  // the drain and its sleep — so the way out must stay reachable.
+  it('keeps cancel and pan-to-close live while replacing', () => {
     const { container } = render(<PlaylistQueueReplaceSheet {...makeProps({ isReplacing: true })} />);
-    expect(container.querySelector('[data-sheet]')?.getAttribute('data-pan-close')).toBe('false');
-    expect(button(container, 'detail.queueReplace.cancel')?.disabled).toBe(true);
+    expect(container.querySelector('[data-sheet]')?.getAttribute('data-pan-close')).toBe('true');
+    expect(button(container, 'detail.queueReplace.cancel')?.disabled).toBeFalsy();
+    // Confirm still blocks a double-tap starting a second drain.
     expect(button(container, 'detail.queueReplace.confirm')?.getAttribute('data-loading')).toBe('true');
   });
 
-  it('allows pan-to-close only while idle', () => {
+  it('can be cancelled mid-replacement', () => {
+    const onCancel = vi.fn();
+    const { container } = render(<PlaylistQueueReplaceSheet {...makeProps({ isReplacing: true, onCancel })} />);
+    fireEvent.click(button(container, 'detail.queueReplace.cancel')!);
+    expect(onCancel).toHaveBeenCalledTimes(1);
+  });
+
+  it('allows pan-to-close while idle', () => {
     const { container } = render(<PlaylistQueueReplaceSheet {...makeProps()} />);
     expect(container.querySelector('[data-sheet]')?.getAttribute('data-pan-close')).toBe('true');
   });
--- a/packages/mobile/src/lib/playlists/__tests__/playlist-page-retry.test.ts
+++ b/packages/mobile/src/lib/playlists/__tests__/playlist-page-retry.test.ts
@@ -1,6 +1,7 @@
 import { describe, it, expect } from 'vitest';
+import { PLAYLIST_DRAIN_MAX_TOTAL_WAIT_MS } from '@boardsesh/playlists-react';
 import {
-  shouldRetryPlaylistPage,
+  createPlaylistPageRetryPolicy,
   MAX_RATE_LIMIT_WAIT_MS,
   RATE_LIMIT_FALLBACK_WAIT_MS,
   RATE_LIMIT_JITTER_MS,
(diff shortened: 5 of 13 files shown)


marcodejongh:
## Review round 2 — all five findings fixed

An adversarial review of the first round found a path where this PR was worse
than the bug it was fixing. Every finding is addressed in 46e511c, and every fix
has a test that **fails when the fix is reverted** (verified by reverting each
one locally and re-running).

### Blocker — a drain that read nothing replaced the whole queue

The zero-page rethrow only covered `stoppedBy: 'error'`. Any other zero-page
stop fell through with `climbs = []`, and `buildPlaylistQueue([], climb)` returns
a one-item queue.

The reachable path is the confirm-after-warning one — the only path that does
**not** pre-seed the queue. Page 0 comes back `RATE_LIMITED` with a full-window
hint (the server emits exactly 60s whenever a throttled request lands in the
first second of its aligned bucket), the wait doesn't fit the drain's budget, the
drain stops at `wait-budget` with zero pages. The climber who had just confirmed
"this clears N queued climbs" and waited on a spinner got their queue destroyed
down to one climb, silently, with no toast. It also tripped the #3891 empty-fetch
canary as a false positive, which permanently disarms that detector for the
playlist for the rest of the session.

Now every non-`exhausted` zero-page stop rethrows, and an aborted drain returns
without writing, reporting or toasting. `exhausted` at zero pages stays a
fall-through — that's the legitimate no-active-board case.

Covered by two new cases in `use-playlist-activation.test.tsx`, on the confirm
path and the tap path. Neither could be reached before: the mobile test's
`drainToExhaustion` stand-in only ever returns `error` / `aborted` / `exhausted`.

### Major — the confirm sheet locked the climber in for ~46s

`enablePanDownToClose={!isReplacing}` plus `disabled={isReplacing}` on Cancel.
Confirming genuinely starts the drain now, and a throttled first page backs off
for a server window, so the app looked hung with no way out. `onCancel` already
aborted the drain *and* its sleep — only the greyed-out button made it
unreachable. Cancel and pan-to-close now stay live; Confirm keeps its spinner so
a double-tap still can't start two drains.

### Major — one wifi blip spent the rate-limit retry

`attempt` was a single counter shared across error classes. Page 7 gets a
transport drop, retries, comes back throttled with a 46s hint, and
`attempt (1) >= MAX_RATE_LIMIT_RETRIES_PER_PAGE (1)` gives up — truncating a
playlist that one honest wait would have finished. The policy is now a factory
built once per page (`createPlaylistPageRetryPolicy`) holding separate
`rateLimitAttempts` / `networkAttempts`. The shared drain injects
`createPageRetryPolicy` instead of a bare function, and there's a drain-level
test that a fresh policy is built per page.

### Minor — jitter overshot the ceiling, so a full-window hint lost its retry

`Math.min(hint, 60_000) + jitter` returned 60000–60250 against a fresh 60000
budget, so `fetchPlaylistPageWithRetry` refused the wait before ever sleeping it.
Jitter now goes inside the ceiling. The old test asserted `>= MAX_RATE_LIMIT_WAIT_MS`,
i.e. exactly the value the budget then rejects — it documented the bug. The new
one pins the ceiling, jitter included, at or under `PLAYLIST_DRAIN_MAX_TOTAL_WAIT_MS`.

### Minor — the swipe-track refresh had backoff but no budget

It opted into the retry policy without a `waitBudget`, so ten pages against a
throttled bucket could leave a fire-and-forget prefetch pending for minutes. It
now shares the same total wait ceiling and treats `PlaylistDrainWaitBudgetError`
as end-of-refresh, returning the climbs it has. `DrainWaitBudget` is exported for
it. New `fetch-playlist-suggestion-climbs.test.ts` covers resume, the budget
stop, and that a real rejection still propagates.

Nothing else changed: still TypeScript only, no `package.json` / `bun.lock` /
`app.config.ts` / `plugins` / `modules` / `patches`, so the native fingerprint is
unmoved and this still ships over OTA.

marcodejongh:
Thanks — three of the four are worth doing and are in the next push.

**`attempt` is misleading** — agreed, and it's exactly the trap that produced the
per-class bug in the first place. `ShouldRetryPage`'s doc now says outright that
`attempt` counts every retry of the page regardless of what failed, that it is
*not* a per-class count, and that a policy with per-class budgets must keep its
own counters and ignore it. Kept rather than removed because a stateless policy
(the tests use several) still wants it.

**`PlaylistDrainWaitBudgetError` as public API** — agreed, dropped from the
barrel along with `DrainWaitBudget`. Both stay exported from
`drain-playlist-pages.ts` because `fetch-playlist-suggestion-climbs.ts` needs
them, but neither is re-exported from `index.ts` now, with a comment saying why.

**Generic Sentry title for `page-cap`** — fixed: the fallback message carries the
stop reason, so `page-cap`, `wait-budget` and a partial `error` no longer group
under one title.

**Coverage gap on the `!activeBoard` short-circuit** — added, and it is more
load-bearing than it was: `exhausted` is now the *only* zero-page stop that does
not rethrow, so a test pins that a zero-page `exhausted` drain still falls
through with no toast. Narrowing the guard to `pagesFetched === 0` would fail it.

**Declining: making `maxPages` required on `drainPlaylistPages`.** There is one
caller, it passes the cap explicitly, and the default is a named constant whose
JSDoc spells out where 30 comes from (half the server's 60/window budget). Making
it required would force every test to pass a page cap that has nothing to do with
what it is testing, which costs more readability than the hypothetical second
caller saves. Happy to revisit if a second caller ever appears.

Also worth noting the earlier review comment on this PR (the one about
`fetchPlaylistSuggestionClimbs` having no `waitBudget`) — that was real and is
fixed in 46e511c: the refresh now shares the same total wait ceiling and stops
with the climbs it has instead of sleeping through ten pages.

marcodejongh:
## Closing this: #4713 shipped the same fix, and this branch would trade parts of it away

I picked this PR back up to rebase it (it went `DIRTY` while the shepherd run was
down) and found the ground had moved. **#4713 — "fix(mobile): bound and retry the
playlist queue-replacement drain" — merged to `main` on 2026-08-26 (`f94fbed99`)
and closed #4622.** It is an independent implementation of this same ticket from a
parallel run, not a re-push of this branch.

So the rebase here is not a merge-conflict resolution. It is a rewrite of code that
landed two days ago, for a ticket that is already closed. I went through both
implementations line by line before deciding, and the honest read is that `main`'s
is the better of the two in the places that matter for #4622.

### Where this branch is weaker than what's on `main`

**1. No `no-progress` guard and no uuid dedup.** `main`'s drain marks every climb
uuid as it goes and stops when a page adds nothing new. That is the precise defence
against the defect class that actually caused the incident — `52d9a631f` fixed a
`RECOMMENDED_*` resolver that clamped its offset at `MAX_RECOMMENDATION_OFFSET` but
computed `hasMore` against the unclamped `totalCount`, re-serving the same 100
climbs with `hasMore: true` forever. `main` exits on the first repeated page. This
branch only exits at page 30, i.e. after spending 30 pages of budget on the same
100 climbs. It also has no dedup, so a repeating server produces a queue full of
duplicates rather than a short one.

**2. The page bound is sized against the wrong number.** The bucket is
`applyRateLimit(ctx, 60, 'smartPlaylist')` — 60 requests per fixed 60s window, keyed
`${userId}:smartPlaylist`, and shared with the list screen's own infinite scroll.

| | pages | max attempts/page | worst-case requests | headroom in a 60/window bucket |
| `main` (#4713) | 20 | 2 | 40 | 20 |
| this branch | 30 | 4 (1 + 1 rate-limit + 2 network) | 120 | **−60** |

This branch's own JSDoc says 30 pages is "half the budget", which counts pages and
not requests. With `MAX_RATE_LIMIT_RETRIES_PER_PAGE = 1` and
`MAX_NETWORK_RETRIES_PER_PAGE = 2` on top, the worst case is double the bucket. A
drain-bounding change for a rate-limit ticket should not be able to spend 2× the
rate limit.

**3. Abort is decided by the error's shape, not by the signal.**
`drain-playlist-pages.ts:181` is `if (isAbortError(error)) throw error;`. `main`
checks `signal.aborted || isAbortError(error)` and documents why: `expo/fetch`
throws a `FetchError` with no `name` and a `fetch failed:` prefix, which
`isTransportNetworkError` classifies as a retryable transport blip — so a
cancellation would be retried and then surface as a red toast plus a Sentry event.
That is latent rather than live today (every shipped build sets
`EXPO_PUBLIC_USE_RN_FETCH=1`, pinned by `scripts/mobile-ci-env-parity.test.ts:74`),
but it is guarded by one build-time env var, and this branch is what plumbs the
signal into `fetch` in the first place.

Separately, `maxPages` being required with no default — the one review point I
declined above — is what `main` shipped. Its reasoning is better than mine was.

### What this branch has that `main` doesn't, and why I'm not fighting for it

**Long rate-limit waits.** This branch waits out a hint up to 60s behind a drain-level
`PLAYLIST_DRAIN_MAX_TOTAL_WAIT_MS` budget. `main` caps a wait at
`PLAYLIST_RATE_LIMIT_MAX_WAIT_MS = 3_000` and hands anything longer to the climber as
the error toast.

I built the 60s version and I now think `main`'s call is the right one. The round-2
review on this PR found the 46s spinner so bad it needed the confirm sheet made
cancellable just to be survivable, and it found a blocker where a zero-page
`wait-budget` stop destroyed a confirmed queue down to one climb. Both of those
defects exist only *because* of the long wait. A 46-second spinner on a tap is worse
UX than "Couldn't start playlist, try again", and it does not even help the common
case: the bucket is shared with the list screen, so the climber who just scrolled a
big playlist deep is the one who gets the wait.

**Partial-queue replacement.** Also a product call `main` made the other way, and the
same review round showed how sharp its edges are.

### Two leftovers that are real and independent of all this

Filed as **#4840** so they don't die with this branch:

1. **The confirm sheet still locks the climber in while the drain runs.**
   `PlaylistQueueReplaceSheet.tsx` on `main` is `enablePanDownToClose={!isReplacing}`
   with `disabled={isReplacing}` on Cancel. `onCancel` already aborts the drain and
   its sleep — only the greyed-out button makes it unreachable. Bounded at ~3s/page on
   `main` instead of 46s, so much smaller than it was here, but a 20-page drain is
   still a stretch with no way out.
2. **One transport blip spends the page's whole attempt budget.**
   `PLAYLIST_PAGE_MAX_ATTEMPTS = 2` is a single counter across error classes, so a page
   that drops once and then comes back with a short `RATE_LIMITED` gets no retry — the
   exact shared-counter bug round 2 found and fixed here with per-class budgets. Small
   fix on top of `main`'s drain, ~30 lines and a test.

### The open review points from 04:37 UTC, answered

They're moot with the branch closing, but not silently:

- **`abortableSleep` exported in the barrel** — correct, and I'd have taken it. It
  was an oversight in `ca6e13f`; `PlaylistDrainWaitBudgetError` and `DrainWaitBudget`
  were pulled and `abortableSleep` should have gone with them. `main` keeps its
  equivalent module-private, which is the right shape.
- **`attempt` in `ShouldRetryPage` reads as per-class** — agreed and already
  documented in the type's JSDoc in `ca6e13f`. Renaming to `_attempt` isn't
  available: it's a real parameter the stateless test policies use. `main` sidesteps
  the whole trap by not exposing an attempt count at all.
- **Wall-clock assertion in the `abortableSleep` test** — agreed, `toBeLessThan(1_000)`
  is a flake waiting for a loaded runner and the `clearTimeout` spy already proves
  what it was there for. Would have deleted it.

The branch stays on the remote if any of this is worth reopening.

marcodejongh:
Closing as superseded by #4713 (merged 2026-08-26, closed #4622). Full reasoning in the comment above. Branch `fix/4622-playlist-drain-retry` stays on the remote — reopen if the long rate-limit wait is worth revisiting. The two parts worth keeping are #4840.

Issue #4499, opened 2026-08-15
[Bug: UX] Board gets hijacked when I reach the gym


ES-Alexander:
I've noticed that sometimes when I get to the gym the board suddenly lights up with a climb I was recently doing, in my custom colours, without me even having opened the app yet (let alone selected to light up the board).

I suspect this may be caused by me leaving without manually disconnecting from the board, and the app automatically reconnecting next time it's in bluetooth range, even though it's literally days later. If that's the case it's possible it will be resolved / possible to work around using the new auto-disconnect feature from #4130, but I last ran into this behaviour on Thursday so it's possible it's an unrelated problem (or I just hadn't updated yet - not sure).

I unfortunately haven't investigated this much, so I'm unsure if it only happens if nobody else is actively connected to the board, and I haven't confirmed whether me disconnecting before leaving the gym makes a difference, or even if it's the _last_ climb I was working on (or one that failed to light up during the session or something). I figured I'd report the issue for now while I remember it (if only because the idea of having so much aura the board remembers me and submits to my climb desires is kind of funny), and then I can report back in future with more information on when/whether the issue is still occurring.

marcodejongh:
🤖 Claimed by bug-batch loop `moarbugs100500` (batch 1, session 43ea5be5).

Investigation + fix in progress. If another automation loop is already on this, say so here and I will stand down. Loop `moarbugs10000` holds #4412–#4414, #4416, #4494, #4519, #4523, #4528, #4529, #4532 — I am staying off those.

PR #4710, opened 2026-08-24
fix(mobile): [native-train] stop the implicit relight clearing the wall with no queue state

### LINKED ISSUE #4544: [native-train] Native relight clears the wall when App-Group queue state is empty — 84% of iOS BLE users hit it on every connect

marcodejongh:
## Summary

Closes #4544.

`displaySharedCurrentItemOnBleQueue()` read the App-Group queue copy and handed the
result straight to `displayCurrentItemOnBleQueue(items:currentIndex:)`, whose first
statement is:

```swift
guard currentIndex >= 0, currentIndex < items.count else { clearBoardOnBleQueue(...); return }
```

`SharedQueueState.load(from:)` returns `(items: [], currentIndex: 0)` when the keys are
simply missing, so **"no shared state was ever written" was indistinguishable from
"the queue is empty"** — and both routed to a real clear-all. On MoonBoard that is the
3-byte `l##` packet, so it writes real bytes rather than being a silent no-op.

Four ordinary connect paths reach it (`finishConnectionSetupOnBleQueue`, `configure(_:)`,
and the two already-connected early returns), and App-Group queue state is only written
once a Live Activity has run. Per the issue's PostHog numbers, 84% of iOS Bluetooth
climbers have never had that state, so every connect they made issued a clear. It went
unnoticed because the JS auto-sender usually repainted a moment later.

### The change

The fix lands in the loader, not in `displayCurrentItemOnBleQueue`. That is the boundary
between "state was read from the App Group and might be absent" and "a caller explicitly
handed me items and an index":

- `displaySharedCurrentItemOnBleQueue` now uses `SharedQueueState.currentItem(from:)`
  (same bounds check) and returns without writing when it yields `nil`.
- `displayCurrentItemOnBleQueue`, `clearBoardOnBleQueue` and all four call sites are
  byte-identical. The explicit-arguments path still clears on an out-of-range index,
  because its callers are authoritative: `SessionWebSocketManager.persistAndNotify`
  repaints from live session state, and `displayCurrentItemAwaitingReady` comes from
  `LiveActivityBleBridge.writeBoardForIntent`.
- JS `bluetooth.clearBoard()` sends empty frames through the write path and never enters
  this code, so a deliberate clear still darks the wall.

One production signature detail worth a reviewer's eye: the loader gained a
`defaults: UserDefaults? = SharedConstants.sharedDefaults` default argument. Swift
re-evaluates default expressions at every call site and all four call sites are
untouched, so this is behaviour-preserving; it exists purely so the Swift suite can
drive the unit against an isolated `UserDefaults` suite instead of a developer machine's
real app-group defaults.

### Relationship to the other relight PRs

- **#4413** is the user-visible report this came out of.
- **#4549** is open and unmerged, JS-only, and mirrors App-Group state so most users get a
  real relight. This PR fixes the residual "never had a queue on this install" case that
  #4549's own tests document as belonging to #4544. They compose; neither duplicates the
  other.
- **#4551** is open, also targets `release/next-ota`, and restructures all four callers
  behind an authorization gate — but it leaves the body of
  `displaySharedCurrentItemOnBleQueue` alone, so the Swift hunks merge cleanly in either
  order. Crucially **#4551 does not already fix this**: a fresh, legitimate `.userConnect`
  passes its gate and still reaches the clear. Registration in
  `scripts/prepare-rn-ios-tests.mjs` is inserted after the `BoardBleWriteFlowTests.swift`
  entry specifically to avoid a textual conflict with #4551's insertion point.

One semantic wrinkle for whoever merges second: once #4551 lands, its
`implicitRelightAttempts` counter increments immediately before this loader runs, so an
"attempt" can now produce no write at all. Nothing breaks, but the counter's meaning
drifts from "writes attempted" to "gate passed". Worth a follow-up, not a blocker here.

### Native train

**This must never be merged to `main`.** It changes Swift under
`packages/mobile/modules/live-activity/ios/`, so it moves the Expo native fingerprint and
can never ship OTA. Base branch is `release/next-ota`, same as #4551.

## Release Notes

- Connecting to a board no longer blanks it when you have not lined up a climb yet — the wall keeps whatever is lit until you pick something.

## Test plan

New XCTest suite `packages/mobile/ios-tests/BoardBleImplicitRelightStateTests.swift`, run
against an isolated `UserDefaults` suite through a new
`TestHooks.displaySharedCurrentItem(defaults:)` seam (the `LiveActivityWidgetTests`
pattern), so it never reads or writes the real app group. The seam sits below whatever
gate wraps the connect-path callers, so it holds in either merge order with #4551:

1. `testImplicitRelightWithNoSharedQueueStateLeavesTheWallAlone` — empty suite, MoonBoard
   config **and** a fake connection seeded (both are required; `clearBoardOnBleQueue`
   bails at its own guards otherwise and the test would pass on unfixed code) → zero
   writes.
2. `testImplicitRelightWithAnOutOfRangeSharedIndexLeavesTheWallAlone` — one item at index
   5 → zero writes.
3. `testImplicitRelightWithAValidSharedCurrentClimbStillLights` — one item at index 0 →
   at least one chunk. Pins that the fix did not just disable relight.
4. `testExplicitDisplayCurrentItemWithNoCurrentClimbStillClears` — no connect flow;
   `displayCurrentItem(items: [], currentIndex: 0)` → exactly one more chunk than the
   baseline, equal to `Data("l##".utf8)`. Pins that SessionWebSocketManager's
   authoritative repaint path is unchanged. (`writtenChunks` is `private(set)`, so this
   asserts on a delta rather than resetting the fake.)

Also fixed a comment in `LiveActivityWidgetTests` that asserted the old
"empty state clears the board" contract; its assertions pass either way, which is exactly
why it would have rotted silently.

**Ran locally (Linux):**

- `vp check` — pass
- `vp test run --project scripts --reporter=agent` — 865 pass. Confirmed
  `prepare-rn-ios-tests.test.ts` goes **red** without the `.mjs` registration and green
  with it, so the new Swift file provably reaches the generated Xcode test target.
- `vp run test:mobile` — 6525/6526 twice, with a **different** 5s-timeout failure each
  run (`use-board-route-target.test.tsx > re-resolves when the network comes back after
  an offline failure`, then `uses the bundled screenshot locale in screenshot mode`).
  Each passes when its file is run alone, so this is local timeout flake under parallel
  load, not a regression — and this PR changes no TypeScript at all. CI's `test-default`
  shards are green.
- `vp run check:mobile-bundle` — pass (iOS and Android)

**Could not run locally:** the Swift tests. There is no XCTest runner on Linux; they only
execute in `ios-rn-ci.yml`'s macOS `build-and-test` job — so CI was the first proof of
both. That job has since gone **green**, so the suite compiles and its assertions hold.

**Device QA still owed before the train ships:**

1. Fresh install, never started a session, connect to a board another climber left lit —
   the wall must keep showing something rather than going dark, and must light your
   climb once you pick one.
2. Connect with a queue and a live session — the wall still re-lights on connect.
3. Start a session, empty the queue, disconnect and reconnect — the wall re-lights the
   last published climb rather than clearing. (This is the intended trade: a stale climb
   beats a black wall, and these boards are last-connection-wins. #4549 asserts the same
   behaviour on the JS side.)
4. Explicit clear from the BLE control sheet still darks the wall.

Bluetooth change, so this needs a Fable review before merge. Keeping it a draft until CI
is green and that review has landed; the merge itself goes out with the native train.

### DIFF
--- a/packages/mobile/ios-tests/BoardBleImplicitRelightStateTests.swift
+++ b/packages/mobile/ios-tests/BoardBleImplicitRelightStateTests.swift
@@ -126,6 +126,20 @@ final class BoardBleImplicitRelightStateTests: XCTestCase {
         )
     }
 
+    /// Distinct from the absent-state case above: here JS did publish, and what
+    /// it published is an empty queue. `SharedQueueState.currentItem` returns
+    /// nil for both, and neither is a licence to clear — an explicit clear
+    /// comes through `write(hex:)` with empty frames, not through this loader.
+    func testImplicitRelightWithAnExplicitlyEmptySharedQueueLeavesTheWallAlone() {
+        let peripheral = FakeWritablePeripheral()
+        installConnection(peripheral)
+
+        SharedQueueState.save(items: [], currentIndex: 0, to: defaults)
+        manager.testHooks.displaySharedCurrentItem(defaults: defaults)
+
+        XCTAssertTrue(peripheral.writtenChunks.isEmpty)
+    }
+
     func testImplicitRelightWithAnOutOfRangeSharedIndexLeavesTheWallAlone() {
         let peripheral = FakeWritablePeripheral()
         installConnection(peripheral)
--- a/packages/mobile/ios-tests/BoardBleImplicitRelightStateTests.swift
+++ b/packages/mobile/ios-tests/BoardBleImplicitRelightStateTests.swift
@@ -152,7 +152,10 @@ final class BoardBleImplicitRelightStateTests: XCTestCase {
         SharedQueueState.save(items: [queueItem()], currentIndex: 5, to: defaults)
         manager.testHooks.displaySharedCurrentItem(defaults: defaults)
 
-        XCTAssertTrue(peripheral.writtenChunks.isEmpty)
+        XCTAssertTrue(
+            peripheral.writtenChunks.isEmpty,
+            "a stale out-of-range index must not clear the wall (#4544)"
+        )
     }
 
     func testImplicitRelightWithAValidSharedCurrentClimbStillLights() {
--- a/packages/mobile/ios-tests/BoardBleImplicitRelightStateTests.swift
+++ b/packages/mobile/ios-tests/BoardBleImplicitRelightStateTests.swift
@@ -137,7 +137,10 @@ final class BoardBleImplicitRelightStateTests: XCTestCase {
         SharedQueueState.save(items: [], currentIndex: 0, to: defaults)
         manager.testHooks.displaySharedCurrentItem(defaults: defaults)
 
-        XCTAssertTrue(peripheral.writtenChunks.isEmpty)
+        XCTAssertTrue(
+            peripheral.writtenChunks.isEmpty,
+            "a published-but-empty queue must not clear the wall either (#4544)"
+        )
     }
 
     func testImplicitRelightWithAnOutOfRangeSharedIndexLeavesTheWallAlone() {
--- a/packages/mobile/ios-tests/BoardBleImplicitRelightStateTests.swift
+++ b/packages/mobile/ios-tests/BoardBleImplicitRelightStateTests.swift
@@ -1,0 +1,177 @@
+import CoreBluetooth
+import XCTest
+
+/// Coverage for the implicit re-light's reading of the App-Group queue copy
+/// (#4544): when that copy holds no current climb, the wall must be left alone
+/// rather than cleared. The explicit-arguments path — the one
+/// SessionWebSocketManager repaints from live session state — must keep
+/// clearing, because there an empty queue is authoritative rather than absent.
+///
+/// Everything runs against an ISOLATED `UserDefaults` suite through
+/// `testHooks.displaySharedCurrentItem(defaults:)`, mirroring
+/// `LiveActivityWidgetTests`, so a developer machine's real app-group defaults
+/// are never read or written. The seam sits below whatever authorisation gate
+/// wraps the connect-path callers, so these assertions hold regardless of how
+/// a connect is gated.
+@available(iOS 17.0, *)
+final class BoardBleImplicitRelightStateTests: XCTestCase {
+    private var scheduler: FakeBleTimerScheduler!
+    private var manager: BoardBleManager!
+    private var suiteName: String!
+    private var defaults: UserDefaults!
+    private var cancelledPeripheralIds: [UUID] = []
+    private let uartWriteCharacteristicUuid = CBUUID(string: "6E400002-B5A3-F393-E0A9-E50E24DCCA9E")
+
+    override func setUp() {
+        super.setUp()
+        suiteName = "com.boardsesh.rn.implicit-relight-tests.\(UUID().uuidString)"
+        defaults = UserDefaults(suiteName: suiteName)
+        defaults.removePersistentDomain(forName: suiteName)
+
+        cancelledPeripheralIds = []
+        scheduler = FakeBleTimerScheduler()
+        manager = BoardBleManager(timerScheduler: scheduler, createCentralManagerEagerly: false)
+        let hooks = manager.testHooks
+        hooks.sync {
+            // A dev machine's app-group defaults could leak a persisted config.
+            hooks.setConfiguration(nil)
+            hooks.setCancelPeripheralConnectionOverride { [weak self] peripheral in
+                self?.cancelledPeripheralIds.append(peripheral.identifier)
+            }
+        }
+    }
+
+    override func tearDown() {
+        manager.testHooks.sync {
+            manager.testHooks.setCancelPeripheralConnectionOverride(nil)
+        }
+        manager = nil
+        scheduler = nil
+        defaults.removePersistentDomain(forName: suiteName)
+        defaults = nil
+        suiteName = nil
+        super.tearDown()
+    }
+
+    // MARK: - Helpers
+    //
+    // `FakeWritablePeripheral` and `FakeBleTimerScheduler` are top-level and
+    // shared, but BoardBleWriteFlowTests' own construction helpers are private
+    // members of that suite, so they are copied rather than reused.
+
+    private func makeCharacteristic(
+        uuid: CBUUID? = nil,
+        properties: CBCharacteristicProperties = .writeWithoutResponse
+    ) -> CBMutableCharacteristic {
+        CBMutableCharacteristic(
+            type: uuid ?? uartWriteCharacteristicUuid,
+            properties: properties,
+            value: nil,
+            permissions: [.writeable]
+        )
+    }
+
+    /// MoonBoard specifically: `makeMoonboardPacket(frames: "")` returns the
+    /// 3-byte `l##` clear-all, so an unwanted clear writes real bytes a test can
+    /// catch instead of being an invisible no-op.
+    private func moonboardConfiguration() -> BoardBleConfiguration {
+        BoardBleConfiguration(
+            boardName: "moonboard",
+            layoutId: 0,
+            sizeId: 0,
+            apiLevel: nil,
+            deviceName: nil,
+            colorOverrides: [:],
+            numRows: nil
+        )
+    }
+
+    private func queueItem(frames: String = "p1r42", mirrored: Bool = false) -> SharedQueueItem {
+        SharedQueueItem(
+            uuid: "queue-item",
+            climbUuid: "climb-item",
+            climbName: "Test climb",
+            difficulty: "6c+/V5",
+            angle: 40,
+            frames: frames,
+            setterUsername: "tester",
+            mirrored: mirrored
+        )
+    }
+
+    /// Seed BOTH the configuration and the connection. `clearBoardOnBleQueue`
+    /// bails at its own guards when either is missing, so a test that seeded
+    /// only one would pass even with the bug present.
+    private func installConnection(_ peripheral: FakeWritablePeripheral) {
+        let hooks = manager.testHooks
+        hooks.sync {
+            hooks.setConfiguration(moonboardConfiguration())
+            hooks.setConnection(peripheral: peripheral, characteristic: makeCharacteristic())
+        }
+    }
+
+    // MARK: - The implicit, shared-state re-light
+
+    func testImplicitRelightWithNoSharedQueueStateLeavesTheWallAlone() {
+        let peripheral = FakeWritablePeripheral()
+        installConnection(peripheral)
+
+        // The isolated suite has never been written to — the state of every
+        // install whose climber has not published a queue yet.
+        manager.testHooks.displaySharedCurrentItem(defaults: defaults)
+
+        XCTAssertTrue(
+            peripheral.writtenChunks.isEmpty,
+            "an absent App-Group queue copy must not clear the wall (#4544)"
+        )
+    }
+
+    func testImplicitRelightWithAnOutOfRangeSharedIndexLeavesTheWallAlone() {
+        let peripheral = FakeWritablePeripheral()
+        installConnection(peripheral)
+
+        // Stale index rather than absent state: a queue was published, then the
+        // index ran past its end.
+        SharedQueueState.save(items: [queueItem()], currentIndex: 5, to: defaults)
+        manager.testHooks.displaySharedCurrentItem(defaults: defaults)
+
+        XCTAssertTrue(peripheral.writtenChunks.isEmpty)
+    }
+
+    func testImplicitRelightWithAValidSharedCurrentClimbStillLights() {
+        let peripheral = FakeWritablePeripheral()
+        installConnection(peripheral)
+
+        SharedQueueState.save(items: [queueItem()], currentIndex: 0, to: defaults)
+        manager.testHooks.displaySharedCurrentItem(defaults: defaults)
+
+        // Only the first chunk goes out synchronously; the rest pace on
+        // chunkDelay timers this test deliberately does not fire.
+        XCTAssertGreaterThanOrEqual(
+            peripheral.writtenChunks.count,
+            1,
+            "a real current climb must still re-light the wall"
+        )
+    }
+
+    // MARK: - The explicit, caller-supplied path
+
+    /// The live-session repaint (SessionWebSocketManager) hands in items and an
+    /// index directly; an out-of-range index there means "no current climb" and
+    /// must still dark the wall. No connect flow, and no reset of
+    /// `writtenChunks` (it is `private(set)`) — the assertion is on the delta.
+    func testExplicitDisplayCurrentItemWithNoCurrentClimbStillClears() {
+        let peripheral = FakeWritablePeripheral()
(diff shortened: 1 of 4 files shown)


marcodejongh:
Thanks — both minor points addressed, and the blocker is acknowledged.

**Fable review (blocking)** — agreed, this is a `packages/mobile/modules/live-activity/ios/` change and CLAUDE.md requires it. Labelled `ble-needs-fable-review`; the review is being arranged separately and the PR stays a draft until it lands. Not merging without it.

**`displayItemOnBleQueue(item:)` not visible in the diff** — it exists and is unchanged, which is why it isn't in the diff. `BoardBleManager.swift:1110`:

```swift
private func displayItemOnBleQueue(
    _ item: SharedQueueItem,
    completion: ((Bool) -> Void)? = nil
)
```

The old code reached it through `displayCurrentItemOnBleQueue`, which called `displayItemOnBleQueue(items[currentIndex], completion: completion)` on the in-range branch. The new loader calls the same method directly with the item `SharedQueueState.currentItem(from:)` already bounds-checked, so it is the same destination minus the clear-on-out-of-range branch. The macOS `build-and-test` job compiles this target, so a missing symbol would be a hard failure there rather than a silent one.

**Test gap — explicitly empty queue state** — fair, and it is a genuinely different shape from "absent". Added in `bfa23b26`:

```swift
func testImplicitRelightWithAnExplicitlyEmptySharedQueueLeavesTheWallAlone() {
    SharedQueueState.save(items: [], currentIndex: 0, to: defaults)
    manager.testHooks.displaySharedCurrentItem(defaults: defaults)
    XCTAssertTrue(peripheral.writtenChunks.isEmpty)
}
```

Your read of the intent is right: an empty *published* queue is still not a licence for this path to clear. A deliberate clear reaches the wall through `write(hex:)` with empty frames from JS `clearBoard()`, or through the explicit-arguments `displayCurrentItem(items:currentIndex:)` that SessionWebSocketManager repaints from — and that one still clears, pinned by `testExplicitDisplayCurrentItemWithNoCurrentClimbStillClears`.

marcodejongh:
Addressed the last nit in `ea50051` — the empty-queue assertion now carries a failure message like the rest of the suite.

The copied-helpers point is a known trade rather than an oversight: `makeCharacteristic`, `moonboardConfiguration` and `queueItem` are `private` members of `BoardBleWriteFlowTests`, so they are not reachable from another suite. Only the top-level `FakeWritablePeripheral` and `FakeBleTimerScheduler` are shared. Hoisting them into a shared fixture would touch the existing BLE suites, which is more churn than this PR should carry into a Fable review — the suite comment says as much so the next person sees the reason.

On "Swift tests unproven": the macOS `build-and-test` job went green on `bfa23b26` (the commit before this cosmetic one), so the suite has now compiled and passed. That was the open question in the PR body, and it is answered.

marcodejongh:
Fixed in `d168c94` — you're right, I missed `testImplicitRelightWithAnOutOfRangeSharedIndexLeavesTheWallAlone`. All four assertions in the suite now carry a message.

The copied-helpers point stands as an accepted trade rather than something to fix here (details in the comment above): the three helpers are `private` members of `BoardBleWriteFlowTests`, and hoisting them into a shared fixture would touch the existing BLE suites — more churn than belongs in a change going into a Fable review. The suite comment records the reason.

Fable review is still the open blocker. Staying a draft until it lands; the merge itself goes out with the native train.

marcodejongh:
Pushing back on (1) — it's inverted, and the nesting is load-bearing rather than accidental.

`displayCurrentItem` does **not** use `runOnBleQueueSync`. It uses the async hop (`BoardBleManager.swift:601`):

```swift
func displayCurrentItem(items: [SharedQueueItem], currentIndex: Int) {
    runOnBleQueue { [weak self] in
        self?.displayCurrentItemOnBleQueue(items: items, currentIndex: currentIndex)
    }
}
```

and `runOnBleQueue` (`:1922`) is the reentrancy-aware form:

```swift
private func runOnBleQueue(_ operation: @escaping () -> Void) {
    if DispatchQueue.getSpecific(key: bleQueueKey) != nil {
        operation()          // already on the BLE queue -> inline, no hop
    } else {
        bleQueue.async { operation() }
    }
}
```

Inside `hooks.sync {}` we are already on the BLE queue, so `getSpecific` is non-nil and the body runs **inline** — no second dispatch, nothing to block on, no deadlock. That inline-nesting is the documented contract on `TestHooks.sync` itself: "Nested manager hops (`runOnBleQueue`) execute inline under this, so a whole action settles before the call returns — the suite's determinism relies on it." `BoardBleWriteFlowTests` drives every one of its write assertions the same way.

Moving the call outside `hooks.sync` would actually break the test: off the BLE queue, `runOnBleQueue` takes the `bleQueue.async` branch, so the write would not have happened by the time the assertion ran. It would flake, or pass vacuously.

The four implicit-relight tests read differently for a real reason: `TestHooks.displaySharedCurrentItem(defaults:)` wraps its own `runOnBleQueueSync`, so it settles whether or not it is nested. `manager.displayCurrentItem` has no such wrapper, which is exactly why the explicit-path test supplies the sync block itself. Different entrypoints, not an inconsistency.

(3) is fair — `XCTAssertEqual(writtenBefore, 0)` would make the baseline explicit. I'm deliberately not pushing it right now: this is a native-fingerprint PR whose CI cycle is a ~30-minute macOS build, and it has already been re-run three times for assertion-message polish. I'll fold it in with whatever the Fable review asks for rather than spend another full native cycle on a one-line guard. Flagging it here so it isn't lost.

(2) stands — Fable review is the real blocker and the PR stays a draft until it lands.

marcodejongh:
## Fable BLE review — PR #4710

**Verdict: APPROVE.** No blocking findings. The semantic split at the heart of the fix is correct, every clear path that should still clear does, the test-injection seam cannot leak into production, and the change composes with #4551 in either merge order. Two non-blocking notes at the end, both already known.

### 1. The core semantic split — verified

`displaySharedCurrentItemOnBleQueue` (`packages/mobile/modules/live-activity/ios/BoardBleManager.swift:1063`) now guards on `SharedQueueState.currentItem(from:)` (`SharedConstants.swift:222`), which returns `nil` for all three "no current climb in the App-Group copy" shapes: keys never written (`load` returns `([], 0)`), a published-but-empty queue, and an out-of-range index. All three now early-return with no write. Previously all three routed into `displayCurrentItemOnBleQueue`'s bounds guard and issued a real clear — on MoonBoard the 3-byte `l##` packet, on Aurora the empty-frames packet — so the bug and the fix are both real byte-level behaviour, not logging.

The paths where clearing IS correct are untouched:

- **JS explicit clear** never enters this loader: `bluetooth-provider.tsx:1359` sends `sendFramesToBoard('', …, { sendSource: 'clear' })` through the write path. Verified by grep — nothing on the JS side calls into the shared-state loader.
- **Live-session repaint**: `displayCurrentItem(items:currentIndex:)` (`BoardBleManager.swift:601`) → `displayCurrentItemOnBleQueue` (`:1077`) → `clearBoardOnBleQueue` (`:1089`) is byte-identical to base. There the caller (SessionWebSocketManager / `displayCurrentItemAwaitingReady`) hands in authoritative state, and an empty queue still darks the wall. Test 4 pins this with exact-packet equality.

The four production call sites, post-fix:

| Call site | Line | With no App-Group state |
|---|---|---|
| `configure(_:)` | `:534` | no write (was: clear-all) |
| `connectOnBleQueue` already-connected early return | `:797` | no write (was: clear-all) |
| `reconnectToLastKnownBoardOnBleQueue` already-connected early return | `:891` | no write (was: clear-all) |
| `finishConnectionSetupOnBleQueue` success point (gated by `!hadPendingReadyWaiters`) | `:1799` | no write (was: clear-all) |

With a valid current item, all four still light it via `displayItemOnBleQueue` — same destination as before, minus the indirection. Test 3 pins that the fix did not simply disable the relight.

### 2. Interaction with #4551 — they compose, either merge order

This was the highest-risk item and it checks out cleanly:

- #4551 decides **whether** the implicit relight runs (the authorization gate); this PR decides **what the loader does** when it runs. Every #4551 path — `performImplicitRelightIfAuthorizedOnBleQueue` and the gated `configure` body — still ends in a zero-argument `displaySharedCurrentItemOnBleQueue()`, which picks up this PR's default argument unchanged. Neither undoes the other.
- I read #4551's full diff: it contains **no hunk touching the loader body**, `SharedQueueState`, or this PR's TestHooks region, and its `prepare-rn-ios-tests.mjs` insertion is after the `BoardBleServiceDiscoveryTests` entry while this PR's is after `BoardBleWriteFlowTests` — non-adjacent, no textual conflict.
- The crucial semantic point in the PR body is right: #4551 does **not** subsume this fix — a fresh legitimate `.userConnect` passes its gate and, unfixed, still reaches the clear. Conversely this PR does not weaken #4551's gate: the test seam calls the private loader directly with injected defaults, below the gate, so the #4544 contract stays pinned regardless of merge order.
- The acknowledged wrinkle stands: once both land, `implicitRelightAttempts` means "gate passed", not "write issued". Follow-up, not a blocker — agreed.

### 3. The `defaults` default-argument seam — safe

Swift evaluates a default-argument expression at each call site, at call time. `SharedConstants.sharedDefaults` (`SharedConstants.swift:129`) is a **computed** static var returning a fresh `UserDefaults(suiteName:)` per access — there is no cached instance a test could have swapped, and no global a test writes. The only way a non-App-Group `UserDefaults` reaches the loader is the explicit parameter on `TestHooks.displaySharedCurrentItem(defaults:)` (`BoardBleManager.swift:2894`), which nothing in production calls. All four production call sites pass no argument (verified byte-identical in the diff). The method is `private`, so no cross-module default-arg inlining concerns either. No capture hazard.

### 4. Queue discipline — no new hazard

The production change adds zero dispatches: a guard, a `logger.info` (os Logger is safe on the serial queue), and an early return. The test hook wraps the loader in `runOnBleQueueSync` (`:1932`), which is reentrancy-guarded via `DispatchQueue.getSpecific`, and the tests call it un-nested from the test thread. I also independently confirm the author's rebuttal of the earlier review-bot deadlock claim: `displayCurrentItem` (`:601`) uses the async reentrancy-aware `runOnBleQueue` (`:1922`), so nesting it inside `hooks.sync` runs inline on the queue and is exactly what makes the write settle before the assertion — moving it outside would make the test racy. No off-queue write is introduced anywhere in this diff.

### 5. State restoration / background wake — improved, not regressed

`finishConnectionSetupOnBleQueue` is the documented single success point for JS connect, widget reconnect, **and** CoreBluetooth state restoration (`:1788`). Pre-fix, a background restoration on an install that never published a queue issued a clear-all with **no JS auto-sender awake to repaint afterwards** — the worst instance of this bug. Post-fix it leaves the wall alone. When App-Group state does exist, restoration still relights it; that over-eagerness is #4499's territory and #4551's gate, unchanged (and not worsened) here.

### 6. No auto-reconnect by the back door

Nothing in this diff adds scanning, connect scheduling, retry, or any connection initiation — it only removes writes from paths that were already connecting. Net product effect is the opposite of a hijack: an arriving climber's phone no longer blacks out a wall someone else lit. Last-connection-wins is untouched.

### 7. Tests — faithful, and they would fail on unfixed code

Against the old loader, tests 1–3 route `([], 0)` / out-of-range state into `clearBoardOnBleQueue`; the suite deliberately seeds **both** the MoonBoard configuration and a fake connection (`installConnection`, `BoardBleImplicitRelightStateTests.swift:106`) so neither of that function's bail-out guards can mask the bug, and the first `l##` chunk lands synchronously in `FakeWritablePeripheral.writeValue` (`BoardBleWriteFlowTests.swift:44`) — `writtenChunks` would be non-empty and the assertions would fail. Test 4's baseline-delta pattern is a correct workaround for `private(set) writtenChunks`. The suite is registered in `scripts/prepare-rn-ios-tests.mjs` (guarded by its own red/green-verified test per the PR body), and both macOS `build-and-test` jobs are green. The `LiveActivityWidgetTests` comment fix (`:354`) correctly retires the old "empty state clears the board" contract text.

### Non-blocking notes

1. `XCTAssertEqual(writtenBefore, 0)` in test 4 (already flagged by claude-review, sensibly deferred by the author to avoid another 30-minute native CI cycle) — fold in with the next real change.
2. The documented trade — empty the queue, reconnect, wall keeps the last published climb instead of clearing — is a deliberate product decision (stale beats black on last-connection-wins boards, mirrored by #4549 on the JS side) and belongs in device QA item 3, which the PR already lists.

### What I could not verify

- No XCTest runner exists on Linux; the Swift suite ran only in `ios-rn-ci.yml`'s macOS job (green twice). I verified the assertions by reading the code, not by executing them.
- No iOS device or physical board — the four device-QA items in the test plan remain owed before the native train ships.
- I read #4551 via `gh pr diff`, not a local merge; I did not build the combined tree. The textual-conflict and composition analysis is from the diffs.

Fable BLE review requirement: **satisfied**. This clears the stated blocker; device QA before the train ships is the remaining owed work.

Issue #4673, opened 2026-08-22
Playlists: "Add to queue" that appends instead of replacing


marcodejongh:
Activating a playlist replaces the whole queue (`use-playlist-activation.ts` → `replaceQueueWithPlaylist`, with the destructive-confirm sheet). The only additive bulk path is `appendGeneratedSession(items)` (`queue-provider.tsx`), used exclusively by the workout generator. A crew building a session from two people's playlists has no non-destructive option.

Proposed: an "Add to queue" action on playlists that appends via the `appendGeneratedSession` primitive (generalised), keeping replace as-is.

From the #4281 thread (@ES-Alexander): "a bulk 'add playlist to queue' option".

Refs #4281

marcodejongh:
🤖 Claimed by the P1 loop `bugbash333` (session 54be984e).

Self-assigned as one of five P1 tickets being driven end to end this pass: plan → plan review (adversarial where the plan doesn't hold up) → implementation in a dedicated worktree → PR → code review → CI and review-feedback shepherding until it lands.

Parallel automation (`moarbugs10000`, `moarbugs100500`, `bugbash20`, codex loops): please pick a different issue, or comment here if you were already on it and I'll hand it back. I'm staying off #4650, #4622, #4103, #4499, #4413, #4143, #4228.

PR #4712, opened 2026-08-24
feat(mobile): add "Play next" so a climb can jump the queue

### LINKED ISSUE #4671: Queue: add a "Play next" action so a climb can jump the line
When the crew's resting climber wants to hop on one quick climb, there's no way to slot it right after the current one — everything appends to the end. The model already supports insertion end-to-end: `addQueueItem(item, position: Int)` is in the GraphQL schema, `DELTA_ADD_QUEUE_ITEM {item, position}` is in the reducer, and `insertAfterCurrent` exists on the set-current path. Only the UI action is missing.

Proposed: a "Play next" option in the queue row action menu (and possibly the climb long-press menu) that inserts after the current climb instead of appending. Mobile `addToQueue` → `commitQueueAdd` (`packages/mobile/src/providers/queue-provider.tsx`) always appends today; thread `position` through.

From the #4281 Discord thread (@ES-Alexander): "a 'play next' option to add/move a climb to next in the queue (instead of at the end), so someone can hop on a few while others are resting".

Refs #4281

marcodejongh:
## Summary

Closes #4671

Long-press any climb → **Play next** slots it directly behind the climb on the wall, instead of at the bottom of the queue.

Everything below the mobile queue provider already spoke `position`: the GraphQL document (`$position: Int`), the shared mutation factory (`createQueueMutations`), the backend resolver (splices at `position`, republishes the *actual* landing index), and the peer-side reducer (`DELTA_ADD_QUEUE_ITEM`). The provider was the single place that dropped the value, so a Play next would have appended for peers. This wires it through and adds the UI.

**The placement maths is one pure function in `@boardsesh/queue`** (`planPlayNext`), not mobile code — the backend and both clients share the vocabulary:

| Situation | Behaviour |
|---|---|
| Empty queue, or **no current climb at all** | insert at **0** — here `queue[0]` genuinely is next: `findNextQueueItemWithSuggestions` returns it when the current item is `null`. Appending would make Play next indistinguishable from Add to queue. |
| Current climb set but **not in the queue** (playlist peek / a slot a peer removed) | insert at **0** as the *best available slot* — **not** because 0 is what plays next. With an orphan current pointer that function falls through to the playlist peek (or `null`) and never consults `queue[0]`, so forward nav won't reach the climb until the current item rejoins the queue. The head still beats appending, which buries it. See "Open question" below. |
| Current climb set, target not queued | insert at `currentIndex + 1` |
| Target **already** in the queue | **MOVE**, never duplicate — see the product call below |
| Target already sits right after current | no mutation, but the snackbar still fires (silence reads as a dead button) |
| Target **is** the climb on the wall | the menu row is **hidden** |

**The move index is direction-aware, and that is the load-bearing bit.** Both the reducer and the backend resolver implement reorder as `splice(oldIndex, 1)` then `splice(newIndex, 0, item)`, so removing first shifts everything after `oldIndex` left by one:

- `oldIndex > currentIndex` → `newIndex = currentIndex + 1`
- `oldIndex < currentIndex` → `newIndex = currentIndex` (the current climb slid down one)

A flat `currentIndex + 1` lands a history item one slot too late. That test was written first and confirmed failing against a deliberately naive implementation (3 assertions caught it) before the correct formula went in.

The commit path re-derives the index from live state **after** the cross-board prompt resolves, and hands the same number to both the optimistic local dispatch and the broadcast — so this client and its peers cannot disagree about where the climb landed.

### Surfaces

One entry in `useClimbActions` (the single source of truth for the long-press list) lights up every surface at once: queue rows, the climbs tab list, playlist detail, the board sheet, and logbook rows. Queue rows pass their `queueItemUuid` through `useQueueSheetHandlers` → `openClimbActions` → `ClimbReactionMenu`, so the *exact* slot moves when the same climb is queued twice.

Not added to `PRIMARY_ACTION_IDS` — the fixed top row stays tick / playlist / share.

### Three calls for review

1. **French wording.** `mobile.climbActions.playNext` = **« Mettre juste après »**. The glossary has no established term for "up next", and « envoyer » is off the table (it is reserved for a send). Flagging for a native-speaker check — a wording-only follow-up commit is expected. Same for the snackbar, « Le bloc passe juste après ».
2. **Move vs. duplicate.** A Play next on an already-queued climb **moves** the existing slot rather than adding a second copy. Rationale: a duplicate leaves a stale entry further down that the crew then deletes by hand, and the issue title says "add/**move**". If review prefers "always add a copy", it is one branch in `planPlayNext` plus its tests — nothing else moves.

3. **Open question — what Play next should do when the current climb is an orphan.** This PR inserts at index 0 and now says honestly that forward navigation won't reach it until the current item rejoins the queue. A peer session prototyped the opposite resolution: **decline the affordance entirely** and return `null` for an orphan current, on the grounds that offering a "play next" that won't actually play next is worse than not offering it. That may well be the better product call. It is not implemented here — this PR keeps insert-at-0, which is what the code review approved — and a human should pick.

### Deliberately out of scope

- **`ClimbActionsSheet.tsx`** (the Android play-drawer sub-sheet) acts on the climb *currently displayed in the player*. "Play next" on the climb you are already looking at is either a no-op or a worse "Set active", so it is left alone.
- **`appendGeneratedSession`** — a generated session is meant to queue *behind* live work. It is a bulk `setQueue`, not a placement decision.
- **No backend change.** The resolver already honours `position` and already has tests for it (`queue-sync-fixes.test.ts`).
- **No `@boardsesh/offline-sync` change.** The outbox only carries ticks / favourites / follows; queue adds go straight over the WS and are best-effort with reconcile-on-failure.
- **No web change.** `packages/web/app/**` has no queue surface since W-16 (#4435); it imports `@boardsesh/queue` in exactly one file, for hashing.

### Review follow-ups (commit `190a0c5c7`)

A code review came back **"ship as-is, nothing blocking"**, independently confirmed the placement maths against every edge case it could construct, and verified both rebuttals to the `claude-review` bot. Five non-blocking findings; three are fixed here.

**Fixed**

- **NB-1 — a false justification, in three places.** The doc comment, `docs/websocket-implementation.md` and this PR body all claimed that with a current climb set but absent from the queue, "the head of the queue IS next". That is wrong: `findNextQueueItemWithSuggestions` returns `queue[0]` **only** when the current item is `null` (`packages/shared/play-view/src/queue-navigation.ts:118-127`); with an orphan current it falls through to `getNextPlaylistClimb` and never consults `queue[0]`. Behaviour is unchanged — index 0 is still the best available slot — but the reason is now the real one, with the two cases spelled out separately. The wrong phrasing is gone from all three.
- **NB-4 — the one untested move branch.** Orphan current with the target already queued is the single path where `planPlayNext` decides "after current" without calling `playNextInsertPosition`. Now covered (`[a,b,c]` + orphan current + target `c` → `move 2→0` → `['q-c','q-a','q-b']`); mutating that arm turns it red. Also added the tightest backward move (`oldIndex === currentIndex - 1`) and a real splice assertion on the current-at-tail fallback.
- **NB-5 — the snackbar copy was untested on both implementations.** Now covered per platform. Verified by mutation: breaking **only** the Liquid Glass call site fails exactly the two Liquid Glass tests and leaves Material green, so the file's "identical public prop API" contract is genuinely enforced rather than incidentally passing.

**Noted, not fixed — these are judgement calls for a human**

- **NB-2 — the move path confirms before the server accepts.** `playNext` shows the "up next" snackbar and returns `'moved'` synchronously, while `reorderQueue` is fire-and-forget. If the backend rejects (`RATE_LIMITED`, or a stale index), the climber sees "Climb is up next" and then ~200 ms later "Action failed", with the queue rolled back. The insert path does not have this problem: its snackbar is inside `commitQueueAdd`, which only runs on a committed add. Worth noting this is new UX surface — a drag reorder shows no success snackbar at all, so there was previously nothing to contradict. Fixing it means either sequencing the snackbar off the reorder's resolution (a spinner-shaped delay on a gesture that is otherwise instant) or deliberately accepting the double message. Left for review to decide rather than picked unilaterally.
- **NB-3 — stacked in-flight reorders roll back too far, and there is no reconcile.** `reorderQueue` snapshots the **whole** queue at dispatch time and restores it on failure. Two Play nexts in quick succession: call 2 lands, call 1 is rejected, call 1's rollback wipes call 2's move that the server did apply — and unlike `addQueueItem`/`removeQueueItem`, the reorder catch does not call `reconcileFailedContentMutation`, so nothing pulls the client back until the next `FullSync`. **Pre-existing in `reorderQueue`**, but Play next raises the trigger rate: a drag needs the queue sheet open and a sustained gesture, whereas this is two taps from any browse list. Deliberately not fixed here — it is a change to shared reorder failure handling that wants its own PR and its own tests, not a rider on a feature branch.

**Also in this commit**

- **Snackbar API renamed to the agreed cross-PR shape:** `showQueueAddedSnackbar({ kind: 'added' | 'playNext', count? })`. `count` is **unused by this PR** — a Play next is always exactly one climb. It exists so the bulk playlist append (#4673) resolves its plural against this same confirmation vocabulary instead of standing up a second, near-identical snackbar. Documented as such at the type definition so it does not read as dead API.
- **`insertQueueItemAfterCurrent` deleted.** Unreferenced repo-wide (definition plus one re-export, zero call sites, zero tests) and it *appended* for an orphan current — a third, contradictory answer to the question `planPlayNext` now owns, sitting one file away from it. Removed rather than left with a comment asking the next reader not to "fix" one to match the other. Called out here so the deletion is reviewable rather than looking like unrelated cleanup. The reducer's `insertAfterCurrent` flag also appends for an orphan and **stays** — correctly, because there the inserted climb *becomes* current, so its slot never decides what plays next. A comment on `planPlayNext` records that distinction.

Unfixed nits from the review that I judged not worth churn: the `is-current` snackbar copy is approximate in a race the UI gate makes unreachable; the fr/de wording preferences fold into the native-speaker pass already flagged above; `queue.next` may want an `iosOpticalCenterRatio`, which needs a device to judge.

## Release Notes

Long-press any climb and hit "Play next" to slot it right behind the one on the wall — no more waiting out the whole queue while the crew rests.

## Test plan

Ran locally on Linux, all green:

| Command | Result |
|---|---|
| `vp test run --reporter=agent --project queue` | **pass** — 157 tests / 7 files |
| `vp test run --reporter=agent --project queue-react` | **pass** — 63 tests / 3 files |
| `vp test run --reporter=agent --project i18n` | **pass** — 101 tests / 2 files (catalog parity for the 8 new values) |
| `vp run typecheck:mobile` | **pass** |
| `vp run test:mobile` | **pass** — 6950 tests / 702 files |
| `vp run check:mobile-bundle` | **pass** — Android Metro bundle OK |
| `vp run check:i18n:orphans` | **pass** — 3855 catalog keys all referenced, 4731 references all resolve |
| `vp check` | **pass** — 0 errors (259 pre-existing repo-wide warnings, none in changed files) |
| pre-commit hook (`vp check --fix`, i18n orphans, board-art network) | **pass** — no `--no-verify` |

**New tests**

- `packages/shared/queue/src/__tests__/play-next.test.ts` — 17 cases covering every row of the semantics table, including the backward-move case asserted by *applying* the double-splice and checking the item lands immediately after the current climb.
- `packages/shared/queue/src/__tests__/reducer.test.ts` — extended `DELTA_ADD_QUEUE_ITEM`: position 0, `position === queue.length`, overshoot, negative, `undefined` (the default-append regression guard), and idempotent echo suppression *with* a position.
- `packages/mobile/src/providers/__tests__/queue-provider-play-next.test.tsx` — 9 cases asserting the local dispatch and the wire mutation carry the **same** number, the default add still sends `(item, undefined)`, the move path calls `reorderQueueItem` and never `addQueueItem`, and the analytics attribution.
- `packages/mobile/src/components/climb-actions/__tests__/use-climb-actions.test.tsx` — the row is hidden for the active climb, and `queueItemUuid` is forwarded only when supplied.

**Not verified (Linux host)**

- `vp run check:mobile-simulator` and `vp run mobile:screenshot` are macOS-only and skip here. No simulator/device run, so the visual placement of the new row and the SF Symbol `text.insert` rendering are **unverified on a real iOS build** (it type-checks against `SymbolViewProps['name']`, which is compile-time validated, and Android falls back to `playlist-play`).
- `vp test run --project backend` did not complete locally within the time budget with other agents contending for the shared docker test infra; backend tests are documented as no-concurrent-runs, and CI's isolated run is the authority. **This PR touches zero backend files** — `git diff --stat` shows nothing under `packages/backend/`.
- The two-device shared-session check in `.boardsesh/qa-notes.md` (step 4: peer B does a Play next; both devices must show the climb move to the slot after current, **not** the bottom) needs real hardware and has **not** been run.

QA notes for a device pass are in `.boardsesh/qa-notes.md`.

https://claude.ai/code/session_01HM7fsZBidhjQ6o8DQrhZDk


marcodejongh:
Checked the review bot's four points against the code. Two of them rest on a false premise.

**1. "playNext bypasses climbToQueueItem" — not accurate, no change made.**

`use-climb-actions.ts` does *not* call `climbToQueueItem` on the add path. The neighbouring `'queue'` action is:

```ts
void addToQueue({ uuid: randomUUID(), climb });   // line 216
```

`climbToQueueItem` appears exactly once in the file, on the `'preview'` action (line 201, for `previewQueueItem`). The new `playNext` entry uses the identical `{ uuid: randomUUID(), climb }` literal, so it is consistent with its neighbour rather than diverging from it. Changing only the new row would be the inconsistency.

**3. "Android icon undifferentiated" — the collision described does not exist in the rendered menu.**

The "Add to queue" row does not use the `queue` icon. It uses `icon: 'add'` → `plus-circle-outline` on Android. "Play next" uses `queue.next` → `playlist-play`. Side by side in the menu they are already distinct glyphs.

`queue` and `queue.next` do share `playlist-play` in the map, but `queue` is the tab-bar / queue-button icon and never renders next to a Play next row. `icon-map.test.ts` explicitly tolerates duplicate values ("Some legitimate duplicates exist"). I left it alone rather than swap in a MaterialCommunityIcons name I cannot render-check on this Linux host — a bad glyph name has no test that would catch it and would ship as an empty box on Android. Happy to change it if a reviewer names a glyph they have seen render.

**2 and 4 are fair and already flagged in the PR body** — the French wording needs a native-speaker pass, and the two-device shared-session check (peer B does a Play next; both devices must show the move to the slot after current, *not* the bottom) needs real hardware. Both are called out under "Two calls for review" and "Not verified"; neither can be closed from here.

For the record on the local backend run mentioned in the test plan: it finished at 3702 passed / 0 assertion failures, with 8 files erroring on worker-database setup (`database "boardsesh_backend_test_w2" does not exist`) — the documented no-concurrent-runs contention, not a real failure. CI's isolated `test-backend` job passed in 2m46s, and this PR touches zero files under `packages/backend/`.

marcodejongh:
## Adversarial review — nothing blocking

Read-only pass over the full diff. Re-ran `--project queue` (155/155) and the touched mobile suites (78/78) locally; CI is 27/27 green.

**The placement maths is correct.** I hand-checked `planPlayNext` against `splice(old,1)` + `splice(new,0,item)` for: adjacent indices in both directions, `oldIndex === currentIndex ± 1`, queue length 1, current at index 0, current at the tail, target *is* current, duplicate climb uuids on both sides of current, and an explicit `queueItemUuid` that no longer exists. No off-by-one. `newIndex` is also provably within `[0, queue.length-1]` for every `move`, so the reducer's bounds guard (`reducer.ts:176`) can never silently swallow a Play-next reorder and the backend's out-of-range throw is unreachable from a fresh local plan. `play-next.test.ts:71-82` and `:113-119` both go red if the direction-aware branch is reverted.

**The divergence from the plan is right, and I'd have asked for it.** The plan's `plan.kind === 'insert' ? plan.position : undefined` is a live bug: a peer queuing the same climb while the cross-board dialog is open flips the re-derived plan to `move`/`already-next` → `undefined` → `insertQueueItemIdempotent` **appends** the fresh item. `playNextInsertPosition` is genuinely total, and `planPlayNext` delegates to it for `insert`, so there is one definition of "where a new item lands".

**Also confirmed:** the same `position` reaches both consumers with no async boundary between them (`queue-provider.tsx:842-846` → `:853` + `:872`); the move path inherits `reorderQueue`'s pre-dispatch snapshot + `UPDATE_QUEUE` rollback intact; the perf checklist is clean (no array `.length`, no inline closures, `playNext` in both the `actionsValue` object and its deps, `useClimbActions` memo deps carry `playNext`/`queueItemUuid`/`activeClimbUuid`); 2 keys × 4 locales with no glossary violations (no `envoyer`, no `senden`, `bloque`/`bloc`/`Boulder` all correct); no `any`, no hardcoded strings, `@boardsesh/queue` still imports nothing but its own types.

**Both rebuttals to the review bot check out.** `use-climb-actions.ts:216` really is `void addToQueue({ uuid: randomUUID(), climb })` and `climbToQueueItem` really does appear once, at `:201` for `preview`. "Add to queue" really is `icon: 'add'` → `plus-circle-outline`, distinct from `queue.next` → `playlist-play`; `icon-map.test.ts` has no android uniqueness assertion. Bonus: `'text.insert'` is a real member of the `SymbolViewProps['name']` union (sf-symbols-typescript 2.2.0), so the iOS glyph is compile-checked rather than a guess.

---

### Non-blocking

**1. "The head of the queue IS next" is false for an orphan current climb** — `packages/shared/queue/src/play-next.ts:70-73`, `docs/websocket-implementation.md:859`, and the PR body's table.

`findNextQueueItemWithSuggestions` returns `queue[0]` **only** when `currentClimbQueueItem === null` (`packages/shared/play-view/src/queue-navigation.ts:127`). With a current item set but not in the queue (`currentIndex === -1`) it falls through to `getNextPlaylistClimb(...)` at `:123-124` and never looks at `queue[0]`. Mobile's forward nav is that function (`queue-provider.tsx:1390`).

Scenario: current is an uncommitted playlist peek, queue `[A,B,C]`, Play next on `X` → `X` lands at index 0 → swipe forward serves the next *playlist* climb (or nothing). `X` does not play next.

Position 0 is still the best available landing slot, so the behaviour is defensible — it's the justification that's wrong, in the comment a future reader will trust when deciding whether to "fix" the branch. Fix: reword to separate `current === null` (where `queue[0]` genuinely is next) from the orphan case.

**2. The move path confirms success before the server accepts it** — `queue-provider.tsx:1030-1032`.

`reorderQueue` is fire-and-forget; `showQueueAddedSnackbar('next')` fires on the next line and `playNext` returns `'moved'`. If the backend rejects (`RATE_LIMITED`, or the stale-index throw at `backend/.../queue/mutations.ts:203-205`), `reorderQueue`'s catch rolls back and toasts — so the climber gets "Climb is up next" followed by "Action failed"/"slow down" with the queue unchanged. The insert path is fine (its snackbar is inside `commitQueueAdd`). New surface: drag reorder shows no success snackbar, so nothing previously contradicted itself. Fix: sequence the snackbar off the reorder's resolution, or accept it deliberately in a comment.

**3. Stacked in-flight reorders roll back too far, with no reconcile** — `queue-provider.tsx:978-999`.

`previousQueue` is a whole-queue snapshot and the failure path dispatches `UPDATE_QUEUE` with it. Tap Play next on `D` then immediately on `E`: call 1 snapshots `Q1`, call 2 snapshots `Q2`. Call 2 lands, call 1 is rejected → `UPDATE_QUEUE(Q1)` wipes `E`'s move that the server *did* apply. Unlike add/remove, the reorder catch never calls `reconcileFailedContentMutation`, so local stays diverged until the next FullSync. Pre-existing, but Play next raises the trigger rate a lot — a drag needs the sheet open and a sustained gesture per move; this is two taps from any browse list, and the QA plan itself expects the rate-limit path at 6+ taps. Fix: `reconcileFailedContentMutation(error)` after the rollback, or refuse a second reorder while one is in flight.

**4. Missing test: the one move branch that doesn't go through `playNextInsertPosition`** — `packages/shared/queue/src/__tests__/play-next.test.ts`.

No case for *current set but not in the queue AND target already queued* → `{ kind: 'move', newIndex: 0 }`. That's the only branch deciding "after current" without calling the helper (`play-next.ts:115`, the `currentIndex === -1 ? 0` arm). Change `playNextInsertPosition`'s orphan branch to append and five tests go red while the move branch silently keeps `0` — two definitions, one green. Concretely: queue `[a,b,c]`, current = orphan, target `c` → `{ move, oldIndex: 2, newIndex: 0 }` → `['q-c','q-a','q-b']`. Lower value, also uncovered: the adjacent-backward move (`oldIndex === currentIndex - 1`) and an `applyMove` assertion on the current-at-tail case.

**5. The snackbar variant is untested in both implementations** — `packages/mobile/src/components/__tests__/queue-added-snackbar.test.tsx`.

All seven existing tests exercise the default; the new `variant` prop has no coverage on Material (`:77`) or Liquid Glass (`:119`). The shared `snackbarMessage` helper (`QueueAddedSnackbar.tsx:31-33`) is the right design and contains the risk today, but the file's contract is "identical public prop API for both", and a later refactor that inlines the key into one variant would ship the wrong copy on one platform with nothing red. Two tests, one per variant.

---

### On the acknowledged gaps

**iOS glyph — low risk.** `text.insert` is compile-validated by the SF Symbols union, so it can't ship as an empty box. Residual risk is purely optical (weight next to `plus.circle`; no `iosOpticalCenterRatio` set, worth an eyeball on the first build).

**Two-device check — medium-low, and nothing here makes a peer append *more* likely.** The opposite: the diff removes the only guaranteed cause (the provider dropping `position`), and `playNextInsertPosition` can only emit `0` or `currentIndex + 1`, bounded by the local queue length. The one remaining append is the server clamp (`mutations.ts:87-91`), which needs this client's queue strictly *longer* than the server's — what the FullSync/sequence gate exists to prevent. What the two-device run would actually earn is coverage of a different, pre-existing hazard: the reorder resolver (`mutations.ts:200-210`) splices whatever is at `oldIndex` **without** checking `queue[oldIndex].uuid === uuid`, while every client reducer does (`reducer.ts:181-183`). A peer's add/remove landing between snapshot and mutation makes the server move the wrong climb, and peers then reject the echo on the uuid mismatch — server diverged from every client until FullSync. Same one-tick window as a drag, so not widened by this PR; worth a follow-up issue, not a blocker here.

Nits (wording alternatives, the `is-current` copy edge, the climb-uuid vs queue-item-uuid definition split) left out of this comment — happy to paste them if useful.

marcodejongh:
## Review: semantics, party races, and the wire

Independent review against a separately-produced plan for #4671 that this branch has not seen. **Overall: this is careful work and the headline claim holds** — `position` genuinely reaches the wire (`mutations.addQueueItem(item, position)`), a test asserts the argument value rather than just the resulting array order, and I could not find a path that silently duplicates a climb. The one thing I'd want decided before merge is a deliberate product choice this PR makes that the plan makes the other way: **Play next pulls a climb *out of the session history* rather than re-queuing it.** Everything else below is medium-or-lower.

---

### 1. Play next drags a climb out of the queue history instead of re-queuing it — CONFIRMED (behaviour), product call

`findTargetIndex` in `packages/shared/queue/src/play-next.ts` falls back to `queue.findIndex(matchesClimb)` over the **whole** queue when there is no match after the current climb. So a Play next on a climb you already sent this session **moves the history row forward** instead of inserting a fresh item.

Why that is user-visible, not just internal:

- `buildQueueListModel` renders `queue.slice(0, currentIndex)` as `history-item` rows under a divider (`packages/shared/play-view/src/queue-list-model.ts:32`, `:35-46`). Moving that row past `currentIndex` makes it disappear from history — the crew's "we already did this" record for that climb is gone.
- The reducer states the opposite model in its own comment: *"the same climb CAN appear multiple times in the queue, e.g., user adds it again after completing it"* (`packages/shared/queue/src/reducer.ts:151-152`).
- The action is on every climb surface, including the history-shaped ones: `LogbookTab.tsx:506`, `SessionTickRow.tsx:127`, `InSessionView.tsx:184`. "I did this earlier, let's do it again" is precisely the tap that hits this branch.

The plan's semantics table resolves this row the other way (ignore occurrences at or before current; insert a fresh item; the history row stays as the record), and so does a second independent implementation of the same issue. Both `play-next.test.ts` cases *"moves a HISTORY climb (before current)…"* and *"falls back to the first match when every copy is before current"* lock this branch in, so it can't be changed later without touching tests — which is why it wants a decision now rather than a follow-up.

Worth noting what it costs: the "direction-aware move index, and that is the load-bearing bit" from the PR description exists **only** to serve this case. If the fallback is restricted, `newIndex` is always `currentIndex + 1` and the whole `oldIndex < currentIndex` arm plus its two tests go away.

Suggested narrowing that keeps the good part: use the anywhere-fallback **only when `target.queueItemUuid` is supplied** — there the climber long-pressed that exact row in the queue sheet and "move this one up" is unambiguous. From the climb list, search, playlist, or a logbook row, restrict to `index > currentIndex` and insert.

### 2. The backend reorder splices by index without checking the uuid — CONFIRMED, pre-existing but newly reachable

`packages/backend/src/graphql/resolvers/queue/mutations.ts:202-207` bounds-checks `oldIndex`/`newIndex` and then splices, but never asserts `queue[oldIndex].uuid === uuid`. The client reducer *does* (`packages/shared/queue/src/reducer.ts:182-185`).

Consequence for the move path: if a peer inserts anything ahead of `oldIndex` between the local plan and the server write, the server moves **whatever now sits at `oldIndex`** — a different climb — and then publishes `QueueReordered{uuid, oldIndex, newIndex}`, which every peer's reducer *rejects* on the uuid check it does have. Server and all clients diverge until the 60 s hash watchdog (`packages/mobile/src/providers/queue/use-session-realtime.ts:906-935` → `packages/shared/queue-runtime/src/sync-gate.ts:267-296`) pulls them back.

This is pre-existing in drag-reorder, and I agree it isn't this PR's job to fix. But NB-3 in the description covers the *rollback* half of `reorderQueue`, not this, and the frequency argument is the same one you already made: a drag needs the queue sheet open and a sustained gesture, this is two taps from any browse list. Worth an issue for a server-side uuid guard, referenced from here.

### 3. "the origin and its peers cannot disagree about where the climb landed" is too strong — CONFIRMED

That sentence is now in `docs/websocket-implementation.md`, so it will be read as a guarantee. It holds for a single writer, not for the concurrent case the PR makes reachable.

Trace, queue `[x, C, y]` with `C` current at index 1, A and B both tap Play next and both send `position: 2`:

- Server serializes on `withQueueVersionRetry`. A lands at 2, publishes `actualPosition: 2`. B retries, lands at 2, publishes `actualPosition: 2`. Server queue: `[x, C, B, A, y]`.
- Client A applies B's event at 2 → `[x, C, B, A, y]`. Correct.
- Client B's **own** echo is a no-op: `insertQueueItemIdempotent` returns the same array for a uuid already present (`packages/shared/queue/src/event-utils.ts:16-18`) and the reducer bails on reference equality (`reducer.ts:152-155`). So B never learns its item actually landed at 3. It then applies A's event at 2 → `[x, C, A, B, y]`. **Diverged from the server and from A.**

Append-only adds never hit this, because the server's `actualPosition` for an append is monotonic — inserting a peer's item at a later index preserves order either way. Positional adds break that. It self-heals on the 60 s watchdog, so it's a UX wobble, not corruption. Ask: soften the doc sentence to say the origin's own optimistic index is *not* corrected by the server's `actualPosition`, and that concurrent placements converge via the ordered-hash watchdog rather than immediately.

### 4. Only the position is re-derived after the cross-board await, not the plan — CONFIRMED, narrow

`commitQueueAdd` recomputes with `playNextInsertPosition(stateRef.current.queue, …)` after the prompt resolves, which is the important half and is right. But the **insert-vs-move** decision was taken before `guardCrossBoardAdd`'s `await`. If the climb enters the queue while the dialog sits open — a peer adds it, or an accepted board switch reseeds the queue — the insert path adds a second copy, which is the exact duplicate this feature exists to avoid. One-line fix: re-run `planPlayNext` after the await and return `'unchanged'` when it no longer says `insert`.

### 5. Android shows the same glyph for "Add to queue" and "Play next" — cosmetic

`icon-map.ts` has `queue: { …, android: 'playlist-play' }` and the new `'queue.next': { ios: 'text.insert', android: 'playlist-play' }`. Two adjacent rows in the same menu, identical icon on Android. iOS is fine — I checked the resolved `sf-symbols-typescript` (2.2.0) and `'text.insert'` **is** in the union, so the compile-time validation the comment relies on is real. A distinct MaterialCommunityIcons glyph for one of the two would help.

### 6. Test coverage note — small gap, the important assertion is present

The assertion the issue is actually about is there and is the right shape: `addQueueItem` called with `(item, 2)`, `(item, 0)`, and `(item, undefined)`. The already-queued-further-down case and the no-current-climb case both have provider tests. The gap: no provider test drives **`playNext`** through the most common path (current in the middle, target not queued) *and* asserts the position — *"sends the SAME position…"* goes through `addToQueue(item, { placement: 'next' })` instead, and *"tags the add event with the placement"* asserts only the event props, not the index or the resulting order. One extra assertion on the existing test closes it.

---

### Checked and clean

- **Semantics table, row by row.** No current climb → insert at 0 (not append) ✓. Current climb is the tail → degenerates to an append ✓. Same climb queued twice after current → the copy after current wins, first-match tiebreak ✓. Target *is* the current climb → hidden in the UI *and* handled by the planner, tested both ways ✓. Already directly after current → no mutation ✓. Orphan current pointer → head, and the doc comment now states the real reason instead of the "queue[0] is next" one, which I verified is indeed false for an orphan current. The `queueItemUuid` threading through `use-queue-sheet-handlers` → `openClimbActions` → `ClimbReactionMenu` → `useClimbActions` is a genuine improvement over the plan for twice-queued climbs.
- **No silent duplicate found** on any path except the narrow race in finding 4.
- **`newIndex` bounds.** Every branch of `planPlayNext` produces a `newIndex` inside the reducer's `newIndex < queue.length` validation — forward moves land in `[currentIndex+1, existingIndex]`, backward at `currentIndex`, orphan at 0.
- **Overshoot/negative clamp parity** between `insertQueueItemIdempotent` and the resolver's `position >= 0 && position <= queue.length` branch, and the positional self-echo stays a no-op — with a test that would catch a regression.
- **The move path correctly skips the cross-board gate** — the item is already in the queue, so the board was accepted when it went in. Easy thing to get wrong.
- **Perf.** The new row reads `useActiveClimbUuid()` (`queue-contexts.ts:326`), whose value is memoized on the uuid string at `queue-provider.tsx:1461-1463` — the menu does **not** subscribe to the queue array. The `useMemo` dep list gains only `queueItemUuid`, `activeClimbUuid` and `playNext`; no array `.length`, no inline closures. `actionsValue` and the snackbar provider's context value both stay `useMemo`'d with the new members in their deps.
- **i18n.** Both keys present in all four locales across both catalogs, literal `t()` keys so the orphan scanner sees them. `es` respects *cola* and the existing *bloque* vocabulary; `fr` avoids « envoyer » and the media-player « lire ensuite » register and reuses *bloc*; `de` is informal, avoids *senden*, and matches the sibling's *Boulder* / *Warteschlange* vocabulary. Agree « Mettre juste après » wants a native-speaker pass, as you flagged — not blocking.
- **No `any`, no hardcoded colours** (`accentColor` off the theme), diff is `.ts/.tsx/.json/.md` only, so the OTA fingerprint is unchanged and the neutral "OTA compatibility" check is the expected result.
- **`insertQueueItemAfterCurrent` deletion** verified genuinely unreferenced — definition plus one re-export, zero call sites.

### For a human to decide

Finding 1 (history move vs. fresh re-queue) is the one that changes what climbers see. NB-2 (snackbar confirms before the server accepts on the move path) and NB-3 (stacked reorder rollbacks with no reconcile) from the description are fair calls to defer, and finding 2 belongs in the same follow-up as NB-3.

Issue #4711, opened 2026-08-24
Offline sync dead-letters RATE_LIMITED mutations instead of retrying them


marcodejongh:
## What happens

The offline mutation drainer **dead-letters** any write the backend rejects with
`RATE_LIMITED`, instead of retrying it. The write is lost permanently — it never
drains, and the climber has no way to make it go through.

## Why

The backend attaches a **string** `code` and no numeric `status`
(`packages/backend/src/graphql/resolvers/shared/helpers.ts:517-519`):

```ts
throw new GraphQLError(error.message, {
  extensions: { code: 'RATE_LIMITED', operation, retryAfterSeconds: error.retryAfterSeconds },
});
```

`statusFromGraphqlErrors` (`packages/shared/offline-sync/src/mutation-queue/error-classification.ts:5-17`)
only reads a **numeric** `extensions.code` or `extensions.status`, so
`getErrorStatus()` returns `null`. `isRetryable()` (same file, `:231-260`) then
takes its "no resolvable status → programmer/validation bug" branch:

```ts
if (status === null) {
  return false;
}
```

`drainer.ts:519` takes the non-retryable path and marks the mutation
`status: 'dead_letter'` (`drainer.ts:551-558`). The `429` branch immediately
below the `status === null` check is never reached, because a GraphQL
`RATE_LIMITED` never resolves to 429 in the first place.

## Reachability (narrow but real)

Two outbox kinds hit rate-limited resolvers:

- **`setter_follows`** create/delete → `applyRateLimit(ctx, 30, 'followSetter')`
  / `'unfollowSetter'`
  (`packages/backend/src/graphql/resolvers/social/setter-follows.ts:849,917`).
  An offline climber who queues more than 30 setter follows loses the overflow.
- **`boardsesh_ticks`** create/update **carrying a beta link** →
  `applyRateLimit(ctx, 30, 'beta-link-validation')`
  (`packages/backend/src/graphql/resolvers/ticks/mutations.ts:383` and `:1263`,
  inside `validateAndEnrichBetaLinkInsert` on the tick write path). A bulk tick
  sync after a session offline can blow past 30 and silently drop ticks.

## The fix already exists — on a branch that never landed

Commit `06e0d0671` ("feat: standings privacy consent…") carries the both-ends
fix, but it is **not on `main`**:

```
$ git merge-base --is-ancestor 06e0d0671 origin/main   →  exit 1 (not an ancestor)
$ git branch -a --contains 06e0d0671
[...]

PR #4713, opened 2026-08-24
fix(mobile): bound and retry the playlist queue-replacement drain

### LINKED ISSUE #4622: Starting a large smart playlist fails: the unpaced page drain trips the smartPlaylist rate limit

marcodejongh:
## Summary

Closes #4622.

**Lead with the honest finding: the `RATE_LIMITED` half of this ticket was a server bug, and it is already fixed.**

`52d9a631f` ("fix(backend): MoonBoard playlist activation left the queue empty") changed how the smart-playlist resolver derives `hasMore`:

```diff
-      hasMore: (recPage + 1) * pageSize < totalCount,
+      hasMore: recPage < maxRecPage && (recPage + 1) * pageSize < totalCount,
```

Before it, the four `RECOMMENDED_*` presets clamped their offset at `MAX_RECOMMENDATION_OFFSET = 500` but computed `hasMore` against the unclamped catalog-wide `totalCount`. From page 6 on the server re-served the same 100 climbs with `hasMore: true` forever, and the client's `while (hasMore && !signal.aborted)` had no exit — it fired `smartPlaylist` requests until the 60/min bucket rejected it. That is the mechanism behind the 23 `RATE_LIMITED` events.

That commit has been on `main` since **2026-07-31 20:48 UTC** (merge `5b08d2029`, PR #4010) and there has been no recurrence in the 24 days since. For calibration: the original bursts ran 07-15, 07-16, 07-17, 07-20, 07-21, 07-28, so the largest gap inside the incident was 7 days. 24 days is 3.4× that. At `pageSize: 100` a recommendation drain is now capped at **6 requests**.

Two caveats stated plainly rather than papered over:

- The ticket's timeline does **not** prove the fix stopped it. The last `RATE_LIMITED` event is 2026-07-28 07:49, which is 3d 13h *before* `52d9a631f` reached `main`. The evidence is the silence *since*, not the stop itself.
- **I could not verify the Sentry gate.** The review asked for a re-run of the Discover query at `statsPeriod=30d` filtered to `source:playlist op:replace-queue` to confirm zero `RATE_LIMITED` events after 2026-07-31 20:48 UTC. Sentry reads need a `sntryu_` user token, and this environment has none (no `sentry-cli`, no Sentry MCP, no token in env). **Please confirm that query before merging.** If any `RATE_LIMITED` appears after that timestamp, the server fix did not hold and this becomes a backend investigation instead of client hardening. Nothing here depends on that answer being "zero" — the still-firing `TypeError: Network request failed` half (21 events / 9 users, last seen 2026-08-17) justifies the change on its own — but the framing above does.

### What this PR actually fixes

**One dropped page no longer kills the whole queue replacement.** `replaceQueueWithPlaylist` awaits the drain and its result is the *only* input to `setQueue`, so any thrown page landed in the catch-all and produced `detail.queueReplace.loadFailed`. With the server fix in, a recommendation drain is 1–6 requests, so this is mundane: one flaky `fetch` out of six. The drain now retries a transport failure once.

**A drain that stops early now reports itself.** `page-cap` and `no-progress` both truncate silently for the climber, so a `reportHandledError` under `op: replace-queue-page-cap` / `replace-queue-no-progress` is the only signal either produces. They are split because their odds differ enormously: `page-cap` needs 20 productive pages (2,000 climbs) and should never fire, while `no-progress` is the count/select drift the guard exists for and *can* fire on live data — a logbook smart playlist whose refs all fail to hydrate (`hydrate-climbs.ts:163` drops refs with no `board_climbs` row) returns `climbs: [], hasMore: true`, which would otherwise be a silent one-item queue. Without this, the PR would have made that shape *less* observable than `main`.

**Abort is decided by the signal, not the error's `name`.** `isAbortError` alone is only correct because every shipped build sets `EXPO_PUBLIC_USE_RN_FETCH=1`. Without it, `expo/fetch` throws a `FetchError` with no `name` and a `fetch failed:` prefix, which `isTransportNetworkError` matches — so a cancellation would be retried as a transport blip and then surface as a red toast plus a Sentry event. Since this PR is what plumbs the signal into `fetch`, both catch sites now check `signal.aborted` first.

**Retries are narrow on purpose.** Only transport-class errors retry, via `isTransportNetworkError` from `@boardsesh/offline-sync/error-classification`. A deterministic server verdict — a GraphQL validation error, a 4xx, the logbook resolver's `throw new Error('User not found')` at `smart-playlists.ts:350` — fails on the first attempt, as it does today. Retrying those would only make the climber wait ~1s longer for the same toast.

**Backing out of a playlist now actually cancels the load.** `signal` was already on the `fetchPage` option type but neither screen destructured it, so `replacementAbortRef.abort()` only stopped the *next* page from being issued while the in-flight one ran to completion. Both screens now pass it into `getHttpClient().request({ document, variables, signal })`.

**The loop is bounded.** `MAX_PLAYLIST_QUEUE_REPLACE_PAGES = 20` plus a no-progress guard (a page adding zero new climb uuids stops the drain). The no-progress guard is the precise defence against the `52d9a631f` defect class — it fires on the first repeated page rather than after 30 — and against any future drift between the logbook branch's `selectSmartClimbRefs` and its separate `countSmartClimbRefs` at `smart-playlists.ts:353-355`. CLAUDE.md's mobile performance checklist bans drain-until-`hasMore` outright.

**The error toast on the confirm path is now visible.** The toast overlay is a root-level JS view that renders *behind* a native `@expo/ui` sheet (`toast-provider.tsx:54-59`, whose own comment says so). `confirmQueueReplacement` runs the drain while `PlaylistQueueReplaceSheet` is still up, so a failure there left the sheet visible with an invisible error behind it. `setPendingReplacement(null)` now runs before `showToast`. This is a latent bug the existing `loadFailed` toast has had all along.

### One behavioural delta worth calling out

`fetchPlaylistSuggestionClimbs` now **deduplicates by uuid**. The old loop pushed `pageResult.climbs` unfiltered; the new path shares `drainPlaylistPages`' `seenClimbUuids` set, which is what the no-progress guard is built on. This is an improvement — duplicates in a swipe track are a bug, and `buildPlaylistQueue` already deduped downstream — but it is a real change for anything that relied on page-order duplicates. Nothing does.

### Deliberately not in this PR

The cap is 20 rather than 30 deliberately: 20 pages x 2 attempts = **40 requests** worst case against the 60-per-60s `smartPlaylist` bucket, which is keyed `${userId}:smartPlaylist` and shared with the list screen's own paging. 30 pages would have put the worst case at exactly 60 — sitting on the trip point the guard exists to avoid. The arithmetic is in a comment next to the constant.

- **No truncation toast and no new i18n keys.** The 30-page cap is a runaway bound, not a product feature. It cannot fire for the four recommendation presets (the server stops at 6 pages), so a toast plus 8 strings across `en-US/es/fr/de` would be permanent translation-parity obligation for a branch nothing should reach. Truncation is silent for the climber and loud for us: hitting the cap fires `reportHandledError` with `op: replace-queue-capped`. If that ever shows in Sentry, we add the toast then, with evidence.
- **No pacing.** Both rate-limit tiers count requests per 60s window, so any delay short enough to be acceptable inside an interactive tap gives zero rate-limit protection — it only adds latency. Terminating the loop is the cap's job.
- **No resume state machine.** At ≤6 pages, "retrying starts from page 0" costs 6 requests out of 60.
- **No backend change.** The 60/min `smartPlaylist` budget stays as it is.

### Triage note

The Sentry tag `use-playlist-activation.ts:414` in the ticket is **stale**. Line 414 today is `setPendingReplacement({` inside the future-queue warning branch; the catch-all moved when the file gained the #3891 empty-fetch canary on 2026-07-30. Triage on the `op: replace-queue` tag, not the line tag. This PR also adds `{ reason, pagesFetched }` to that report so the next occurrence is classifiable without one.

### Filed separately

#4711 — the offline-sync drainer dead-letters `RATE_LIMITED` mutations instead of retrying them. The fix exists on `feat/standings-phase0`/`phase1` (`06e0d0671`) but is **not on `main`**, so `isRetryable` still dead-letters string-coded `RATE_LIMITED`, reachable via `setter_follows` (30/min) and beta-link ticks. Cross-linked to #4458 so it isn't fixed twice. Out of scope here — the playlist drain is a read, not an outbox mutation.

## Release Notes

Starting a big playlist no longer dies on one dropped page: a flaky page retries on its own, and backing out of a playlist actually cancels the load.

## Test plan

Every command below was run in a fresh worktree off `origin/main` (`76cc8f522`) and passed:

| command | result |
| --- | --- |
| `vp run typecheck:web` | pass (`playlists-react` is a web dep) |
| `vp test run --project playlists-react --reporter=agent` | pass — 41 tests, 5 files |
| `vp run test:mobile` | pass — 6938 tests, 701 files |
| `vp run check:mobile-bundle` | pass |
| `vp check` | pass |

**30 new tests** (21 shared, 9 mobile). `packages/shared/playlists-react/src/__tests__/fetch-playlist-suggestion-climbs.test.ts` is new — there was no test for this file at all before. It covers: the page cap; the no-progress guard replaying the `52d9a631f` bug (page 5 onward re-serving page 4 with `hasMore: true`); an empty page with `hasMore: true` vs an empty *final* page (the second must stay `complete`, because the #3891 canary is gated on it); transport retry; **non-transport errors NOT being retried**; `RATE_LIMITED` at 2s waited out vs 46s failing fast, built in the real graphql-request `ClientError` shape; abort during a page; abort during a sleep under fake timers asserting `vi.getTimerCount() === 0` and no late resolve; a pre-aborted signal doing zero fetches; and the preserved `fetchPlaylistSuggestionClimbs` contract (10-page cap, `maxClimbsAfterActivated` counted only after the activated uuid is seen, no double-count on a retried page).

**Three mutation tests were run to prove the new guards are load-bearing**, not just green:

1. Flipping the retry predicate to retry-everything (the policy the original plan proposed) turned exactly one test red: *"does NOT retry a non-transport error"*.
2. Removing the canary stop-reason gate turned red: *"does not fire the #3891 empty-fetch canary when the drain stopped on no-progress"*.
3. Removing the `setPendingReplacement(null)` call from the catch turned red: *"dismisses the confirm sheet when the confirmed replacement fails"*. To be precise about what that test guards: it pins the **presence** of the dismiss, not its position. Moving the call after `showToast` leaves it green, because both are React state updates in the same async continuation and React 19 batches them into one commit — the order cannot change what renders. The presence is the real fix: on `main` the catch never dismissed the sheet at all, so a confirm-path failure left it up indefinitely with an invisible toast behind it. The ordering in the source is documentation of intent, not a guarantee the test enforces.

One deliberate divergence from the review's test list: it asks for *"`page-cap` replaces the queue and does not fire the #3891 canary"*. With the real drain wired in, a `page-cap` stop is **unreachable with zero climbs** — it requires 30 successful pages that each added new climbs, so `climbs.length >= 30` and the canary could never fire on it anyway. The reachable canary-poisoning case is `no-progress` with zero climbs (an empty page that still claims `hasMore`), so the test guards that instead, plus a separate test that `page-cap` replaces the queue and reports `replace-queue-capped`. Both were confirmed to fail before the fix.

The mobile test file previously stubbed the entire `@boardsesh/playlists-react` module, which would have made every page-fetch assertion vacuous once the loop moved into the package. It now pulls the real pure-TS drain through its subpath export and stubs only the React hook.

**Not verified:** the Sentry gate (no read token in this environment — see Summary), and on-device QA. `.boardsesh/qa-notes.md` covers the flows to exercise: happy path unchanged on both screens, airplane-mode toggle mid-drain riding it out, a non-network failure failing immediately rather than after a retry, back-out actually cancelling, and the confirm-sheet toast being visible on failure.

**Risk / rollback.** JS-only, no migrations, no backend change, no native change, no BLE code (no Fable review gate). Behaviour is governed by constants in one shared module: setting `PLAYLIST_PAGE_MAX_ATTEMPTS = 1` reverts to today's timing with the cap intact; `git revert` plus the automatic production OTA on merge to `main` is a full rollback at an unchanged fingerprint.

### DIFF
--- a/packages/mobile/app/(tabs)/discover/[playlist_uuid].tsx
+++ b/packages/mobile/app/(tabs)/discover/[playlist_uuid].tsx
@@ -121,10 +121,12 @@ export default function PlaylistDetail() {
       page,
       pageSize,
       board,
+      signal,
     }: {
       page: number;
       pageSize: number;
       board: { boardName: string; layoutId: number; sizeId: number; setIds: string; angle: number };
+      signal: AbortSignal;
     }) => {
       const input: GetPlaylistClimbsInput = {
         playlistId: playlistUuid,
@@ -136,9 +138,14 @@ export default function PlaylistDetail() {
         page,
         pageSize,
       };
+      // See the smart-playlist screen: passing the signal is what turns
+      // "stop paging" into "cancel the request that is already out".
       const response = await getHttpClient().request<GetPlaylistClimbsQueryResponse, { input: GetPlaylistClimbsInput }>(
-        GET_PLAYLIST_CLIMBS,
-        { input },
+        {
+          document: GET_PLAYLIST_CLIMBS,
+          variables: { input },
+          signal,
+        },
       );
       return {
         climbs: toQueueClimbs(response.playlistClimbs.climbs),
--- a/packages/mobile/app/(tabs)/discover/smart/[type].tsx
+++ b/packages/mobile/app/(tabs)/discover/smart/[type].tsx
@@ -55,7 +55,17 @@ export default function SmartPlaylistDetail() {
   // Suggestion-refresh fetcher pages the smart playlist scoped to the active
   // board name so the play-drawer swipe walks the full computed list.
   const fetchPage = useCallback(
-    async ({ page, pageSize, board }: { page: number; pageSize: number; board: { boardName: string } }) => {
+    async ({
+      page,
+      pageSize,
+      board,
+      signal,
+    }: {
+      page: number;
+      pageSize: number;
+      board: { boardName: string };
+      signal: AbortSignal;
+    }) => {
       const input: GetSmartPlaylistInput = {
         type: smartType,
         userId,
@@ -63,10 +73,14 @@ export default function SmartPlaylistDetail() {
         page,
         pageSize,
       };
-      const response = await getHttpClient().request<GetSmartPlaylistQueryResponse, { input: GetSmartPlaylistInput }>(
-        GET_SMART_PLAYLIST,
-        { input },
-      );
+      // `signal` is what makes backing out of a playlist actually cancel the
+      // load. Without it the drain only skipped the NEXT page; the in-flight
+      // request ran to completion and could still replace the queue late.
+      const response = await getHttpClient().request<GetSmartPlaylistQueryResponse, { input: GetSmartPlaylistInput }>({
+        document: GET_SMART_PLAYLIST,
+        variables: { input },
+        signal,
+      });
       return {
         climbs: toQueueClimbs(response.smartPlaylist.climbs),
         hasMore: response.smartPlaylist.hasMore,
--- a/packages/mobile/src/lib/playlists/__tests__/use-playlist-activation.test.tsx
+++ b/packages/mobile/src/lib/playlists/__tests__/use-playlist-activation.test.tsx
@@ -884,6 +884,7 @@ describe('usePlaylistActivation (mobile wrapper)', () => {
           expect.objectContaining({ tags: { source: 'playlist', op: 'replace-queue-capped' } }),
         );
       });
+      // Exactly one call per page: the mock never rejects, so no retry inflates this.
       expect(fetchPage).toHaveBeenCalledTimes(MAX_PLAYLIST_QUEUE_REPLACE_PAGES);
       // Truncated, not failed: the climber still gets a circuit and no toast.
       expect(mocks.setQueue).toHaveBeenCalled();
--- a/packages/mobile/src/lib/playlists/__tests__/use-playlist-activation.test.tsx
+++ b/packages/mobile/src/lib/playlists/__tests__/use-playlist-activation.test.tsx
@@ -90,6 +90,20 @@ vi.mock('../../climb-to-queue-item', () => ({
   }),
 }));
 
+/** The shape graphql-request 7.4.0 actually throws for a rate-limited operation. */
+function makeRateLimitClientError(retryAfterSeconds: number): Error {
+  return Object.assign(new Error(`Rate limit exceeded. Try again in ${retryAfterSeconds} seconds.`), {
+    response: {
+      errors: [
+        {
+          message: `Rate limit exceeded. Try again in ${retryAfterSeconds} seconds.`,
+          extensions: { code: 'RATE_LIMITED', operation: 'smartPlaylist', retryAfterSeconds },
+        },
+      ],
+    },
+  });
+}
+
 function makeClimb(uuid: string, overrides: Partial<Climb> = {}): Climb {
   return {
     uuid,
@@ -881,7 +895,7 @@ describe('usePlaylistActivation (mobile wrapper)', () => {
       await waitFor(() => {
         expect(mocks.reportHandledError).toHaveBeenCalledWith(
           expect.any(Error),
-          expect.objectContaining({ tags: { source: 'playlist', op: 'replace-queue-capped' } }),
+          expect.objectContaining({ tags: { source: 'playlist', op: 'replace-queue-page-cap' } }),
         );
       });
       // Exactly one call per page: the mock never rejects, so no retry inflates this.
@@ -914,6 +928,81 @@ describe('usePlaylistActivation (mobile wrapper)', () => {
       );
       // And it does not spin: one page is enough to know the server is stuck.
       expect(fetchPage).toHaveBeenCalledTimes(1);
+    });
+
+    // The reachable half of the guard, and the one that used to be silent: a
+    // logbook playlist whose refs all fail to hydrate returns `climbs: []` with
+    // `hasMore: true`. The climber still gets a one-item queue either way, so
+    // this Sentry report is the only signal anyone gets.
+    it('reports a no-progress stop under its own op so triage can tell it from the page cap', async () => {
+      const fetchPage = vi.fn().mockResolvedValue({ climbs: [], hasMore: true });
+      const { result } = renderActivation(fetchPage, {
+        replaceQueueOnActivate: true,
+        sourceId: 'playlist:no-progress-2',
+        allClimbs: [makeClimb('a'), makeClimb('b')],
+      });
+
+      await act(async () => {
+        await result.current.activate(makeClimb('b'));
+      });
+
+      await waitFor(() => {
+        expect(mocks.reportHandledError).toHaveBeenCalledWith(
+          expect.any(Error),
+          expect.objectContaining({
+            tags: { source: 'playlist', op: 'replace-queue-no-progress' },
+            extra: { sourceId: 'playlist:no-progress-2', pagesFetched: 1, climbCount: 0 },
+          }),
+        );
+      });
+      // Still silent for the climber.
+      expect(mocks.showToast).not.toHaveBeenCalled();
+    });
+
+    // The rate-limit path runs through the REAL parseRateLimitError from
+    // @boardsesh/graphql-client (this file does not mock it). Without these two,
+    // renaming `retryAfterSeconds` or breaking the ClientError shape match would
+    // leave every suite green while every rate limit became a hard failure.
+    it('waits out a short rate limit read by the real parser and still fills the queue', async () => {
+      const tapped = makeClimb('b');
+      const fetchPage = vi
+        .fn()
+        .mockRejectedValueOnce(makeRateLimitClientError(1))
+        .mockResolvedValue({ climbs: [makeClimb('a'), tapped], hasMore: false });
+      const { result } = renderActivation(fetchPage, { replaceQueueOnActivate: true });
+
+      await act(async () => {
+        await result.current.activate(tapped);
+      });
+
+      expect(fetchPage).toHaveBeenCalledTimes(2);
+      expect(mocks.showToast).not.toHaveBeenCalled();
+      const lastSetQueue = mocks.setQueue.mock.calls.at(-1);
+      expect(lastSetQueue?.[0].map((item: ClimbQueueItem) => item.climb.uuid)).toEqual(['a', 'b']);
+    }, 10_000);
+
+    it('fails fast on a long rate limit and hands the original error to Sentry', async () => {
+      const fetchPage = vi.fn().mockRejectedValue(makeRateLimitClientError(46));
+      const errorSpy = vi.spyOn(console, 'error').mockImplementation(() => {});
+      const { result } = renderActivation(fetchPage, { replaceQueueOnActivate: true });
+
+      await act(async () => {
+        await result.current.activate(makeClimb('b'));
+      });
+
+      await waitFor(() => {
+        expect(mocks.showToast).toHaveBeenCalledWith('detail.queueReplace.loadFailed', 'error');
+      });
+      // No spinner held for 46 seconds, and no retry spent on it.
+      expect(fetchPage).toHaveBeenCalledTimes(1);
+      // The ORIGINAL error is what reaches reportHandledError, so
+      // isGraphqlRateLimitedError can still read extensions.code and downgrade
+      // the event to warning + rate_limited instead of paging someone.
+      const reported = mocks.reportHandledError.mock.calls.at(-1);
+      expect(reported?.[0]).toMatchObject({
+        response: { errors: [{ extensions: { code: 'RATE_LIMITED', retryAfterSeconds: 46 } }] },
+      });
+      errorSpy.mockRestore();
     });
 
     // The toast overlay renders BEHIND a native @expo/ui sheet, so a failure on
--- a/packages/mobile/src/lib/playlists/__tests__/use-playlist-activation.test.tsx
+++ b/packages/mobile/src/lib/playlists/__tests__/use-playlist-activation.test.tsx
@@ -3,6 +3,7 @@ import { describe, it, expect, vi, beforeEach } from 'vitest';
 import { renderHook, act, waitFor } from '@testing-library/react';
 import type { Climb, ClimbQueueItem } from '@boardsesh/queue';
 import type { UsePlaylistClimbActivationOptions } from '@boardsesh/playlists-react';
+import { MAX_PLAYLIST_QUEUE_REPLACE_PAGES } from '@boardsesh/playlists-react/fetch-playlist-suggestion-climbs';
 import { usePlaylistActivation, _resetEmptyBoardFetchReportsForTests } from '../use-playlist-activation';
 
 // The shared `usePlaylistClimbActivation` is exercised by @boardsesh/playlists-react's
@@ -36,15 +37,25 @@ const mocks = vi.hoisted(() => ({
 
 const VIEW_ONLY_BOARD = { boardName: 'tension', layoutId: 9, sizeId: 5, setIds: '1,2', angle: 35 };
 
-vi.mock('@boardsesh/playlists-react', () => ({
-  usePlaylistClimbActivation: (options: UsePlaylistClimbActivationOptions) => {
-    mocks.captured = options;
-    return mocks.activate;
-  },
-  fetchPlaylistSuggestionClimbs: (args: unknown) => mocks.fetchSuggestion(args),
-  isAbortError: (error: unknown) => error instanceof Error && error.name === 'AbortError',
-  PLAYLIST_SUGGESTION_REFRESH_PAGE_SIZE: 100,
-}));
+// Only the React hook is stubbed. The paged drain is the pure-TS half of the
+// package, and stubbing it would make every assertion below about page fetching
+// vacuous — so it comes from the real module via its subpath export.
+vi.mock('@boardsesh/playlists-react', async () => {
+  const drain = await vi.importActual<typeof import('@boardsesh/playlists-react/fetch-playlist-suggestion-climbs')>(
+    '@boardsesh/playlists-react/fetch-playlist-suggestion-climbs',
+  );
+  return {
+    usePlaylistClimbActivation: (options: UsePlaylistClimbActivationOptions) => {
+      mocks.captured = options;
+      return mocks.activate;
+    },
+    fetchPlaylistSuggestionClimbs: (args: unknown) => mocks.fetchSuggestion(args),
+    isAbortError: drain.isAbortError,
+    drainPlaylistPages: drain.drainPlaylistPages,
+    MAX_PLAYLIST_QUEUE_REPLACE_PAGES: drain.MAX_PLAYLIST_QUEUE_REPLACE_PAGES,
+    PLAYLIST_SUGGESTION_REFRESH_PAGE_SIZE: drain.PLAYLIST_SUGGESTION_REFRESH_PAGE_SIZE,
+  };
+});
 vi.mock('../../../providers/queue-provider', () => ({
   useQueueActions: () => ({
     setCurrentClimb: mocks.setCurrentClimb,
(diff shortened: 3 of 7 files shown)


marcodejongh:
Thanks — addressed 1 and 3 in `f92451842`, both comment-only.

**(1) `stopAfterPage` skips the final page.** Correct, and worth documenting. It is only consulted while `hasMore` is still true, so on the last page there is nothing left to stop. Harmless for `fetchPlaylistSuggestionClimbs` (its counters only decide whether to keep going), but I have spelled out the invariant on the option's doc comment so nobody reaches for it as a see-every-page hook.

**(3) Page-cap test assumes no retries.** Correct — that mock never rejects. Added a one-line comment saying so.

**(2) Multi-paragraph comments — not acting on this one.** I grepped `CLAUDE.md` for the cited "one short line max" comment rule and it does not exist; there is no comment-length rule in the file at all. The surrounding code is also full of multi-paragraph explanatory comments by design — the #3891 empty-fetch canary comment immediately below my changes is eight lines, and it earns them. The blocks flagged here explain *why* the sleep must lose to its abort signal and *why* there is no pacing, which are exactly the things a future reader would otherwise undo. Happy to trim if a human reviewer disagrees.

marcodejongh:
Second round addressed in `3c3105b76`.

**Attempt budget vs rate limits** — accurate reading, and the behaviour is intentional: the per-page budget covers both retry reasons, so a rate limit arriving on attempt 2 throws rather than sleeping again. That is precisely what keeps a page bounded at two requests however the failures interleave, and it is what stops a rate limit becoming an unbounded wait. You were right that nothing pinned it. Added a test that interleaves them (transport failure then rate limit) and verified by mutation that hoisting the rate-limit branch above the attempt check turns that test red and nothing else.

**Dedup delta** — correct, and now called out explicitly in the PR description. `fetchPlaylistSuggestionClimbs` deduplicates by uuid now because it shares the `seenClimbUuids` set the no-progress guard is built on. Strict improvement (duplicates in a swipe track are a bug, and `buildPlaylistQueue` already deduped downstream), and no caller relies on page-order duplicates.

**Sentry gate** — still unverified and still flagged in the description; no read token in the environment I ran in. It is a pre-merge ask for a human, not something I can close out.

marcodejongh:
## Review — no blocking defects, four non-blocking fixes

Read the full diff against `origin/main` in a clean worktree at `3c3105b76`. Locally green: `typecheck:mobile`, `typecheck:web`, the shared suite (19/19) and the mobile activation suite (41/41).

**Verified correct, so I'm not re-litigating any of it:** the retry precedence (abort → budget → rate-limit ≤3 s → transport → rethrow) including `Error('User not found')` failing on attempt 1; the attempt budget, whose check at `fetch-playlist-suggestion-climbs.ts:181` runs *before* classification so alternating error types can't buy extra attempts; the abortable sleep (pre-aborted, `clearTimeout` on abort, `{ once: true }` plus explicit removal, no late resolve, `getTimerCount() === 0` asserted with real timers); the `page-cap`-can't-be-empty arithmetic behind the test-11 reframe, and that the reframed test is non-vacuous (`renderableCount === 3`); atomicity — `setQueue` only after the full drain, never on a throw; and that `parseRateLimitError`'s structural path matches what `graphql-request@7.4.0` actually throws. Divergences (a), (c) and (d) all hold up.

### 1. `no-progress` truncates silently *and* switches off the only signal that would have caught it

`fetch-playlist-suggestion-climbs.ts:217-222` · `use-playlist-activation.ts:413-418, 429-433`

`page-cap` reports to Sentry and the PR body sets out the right evidence loop for it. `no-progress` gets nothing — no toast, no breadcrumb, no handled error — and it's the branch that can actually fire on live data, while `page-cap` needs 30 productive pages (3,000 climbs).

Concretely: a `PROJECTS` smart playlist where `countSmartClimbRefs` returns 400, so `hasMore` is true (`smart-playlists.ts:369`), but `hydrateClimbsByRefs` drops every ref on page 0 (`hydrate-climbs.ts:163` `if (!row) continue;` — stale ticks against removed catalogue rows). That's the exact count/select drift the guard's own comment cites. The drain stops at page 0 with `climbs: []`, `use-playlist-activation.ts:431` suppresses the #3891 canary for that stop reason, `setQueue` builds a one-item queue, and nothing is reported. For `climbs.length === 0 && hasMore === true` this PR makes the failure *less* observable than main.

Worth writing down alongside it: the logbook branch paginates on non-unique `ORDER BY` (`MOST_REPEATED` is `SUM(attempt_count) DESC` with no tiebreaker, `smart-playlists.ts:108-121`), so separate executions can overlap between offsets. Partial overlap survives; a full repeated page truncates a valid playlist. Low probability, and stopping is still right — but only if it's observable.

Fix (~8 lines, no i18n): extend the existing report to `drainStopReason === 'page-cap' || drainStopReason === 'no-progress'`, with `op: 'replace-queue-no-progress'` so the #3891 canary stays clean.

### 2. Abort isn't normalised — the "abort wins over everything" contract depends on which `fetch` shipped

`fetch-playlist-suggestion-climbs.ts:180` · `use-playlist-activation.ts:474-486`

Both places identify an abort structurally; neither checks `signal.aborted`. Fine today, because every shipped build sets `EXPO_PUBLIC_USE_RN_FETCH=1` (`ios-testflight-rn.yml:70`, `android-apk-rn.yml:94`, `mobile-ota-production.yml:99`) and RN's whatwg-fetch rejects with `name: 'AbortError'`.

That flag is an opt-*out* of Expo's default (`expo/src/winter/runtime.native.ts:41-52`). Without it — or in a local dev build with no `packages/mobile/.env` — `expo/fetch` throws `new FetchError('The operation was aborted.')` (`fetch.ts:78,86`), and `FetchError` (`FetchErrors.ts:2-10`) sets no `name` and prefixes the message `fetch failed:`. So `isAbortError` → false, and `isTransportNetworkError` → **true** via the `fetch failed` marker (`error-classification.ts:65`). Attempt 1 survives by luck (the sleep converts it); attempt 2 hits the budget check at `:181` and rethrows the raw `FetchError`, which reaches `use-playlist-activation.ts:475` unrecognised. The climber hits one transport blip, backs out, and gets a red `loadFailed` toast plus an `op:replace-queue` Sentry event for a cancel they asked for. Newly reachable, because this PR is what plumbs the signal into `fetch` at all.

Two-line fix, preferably in the drain so the contract is true for any runtime: `if (signal.aborted || isAbortError(error)) throw createAbortError();` at the top of the catch in `fetchPageWithRetry`.

### 3. The rate-limit path is never exercised against the real `parseRateLimitError`

`__tests__/fetch-playlist-suggestion-climbs.test.ts:515` uses a local stand-in — right call for a package with no `graphql-client` dep — but the mobile suite never throws a rate-limit-shaped error either. So `PLAYLIST_DRAIN_RETRY_POLICY` (`use-playlist-activation.ts:56-60`) and the entire justification for divergence (a) have zero coverage. Rename `retryAfterSeconds`, or start boxing errors as `GraphQLOperationError`, and every rate limit silently becomes a first-attempt hard failure with both suites still green. Two mobile cases fix it: the `ClientError` shape with `retryAfterSeconds: 2` (re-issued) and with `46` (one call, original error reported).

### 4. `maxPages × maxAttempts` lands exactly on the rate-limit budget

30 × 2 = 60 requests, which is precisely `applyRateLimit(ctx, 60, 'smartPlaylist')` (`smart-playlists.ts:274`) on a bucket shared with the list screen's own infinite scroll — zero headroom in the worst case of the guard that exists to prevent a rate limit. Separately, the 3 s ceiling is per page, not per drain, so 30 pages can cumulatively sleep ~97 s behind the `isReplacingQueue` spinner with no overall deadline. Both need a 3,000-climb logbook playlist plus sustained failures, so: design note, not a live bug. Either drop the cap to 20 or add a drain-level deadline.

### On the two known gaps

**Sentry gate:** low-to-medium. Nothing in the diff depends on the answer — every guard is additive and inert on a healthy server. It protects the framing, not the code. The one way it bites is that if `RATE_LIMITED` did recur, finding 1 means the drain absorbs it into a short queue instead of failing loudly. Someone with a `sntryu_` token should run the ticket's Discover query at `statsPeriod=30d`, `source:playlist op:replace-queue`, but I wouldn't hold the diff on it.

**No on-device QA:** medium, and concentrated entirely on the abort plumbing. The pure-TS half carries little residual risk. The two screen changes are the only thing whose behaviour depends on the device runtime's `fetch`, and that's exactly where finding 2 lives. Five minutes covers it: open a smart playlist, tap a climb, then immediately navigate back — and separately tap a different climb — and confirm no error toast and no `op:replace-queue` event. Repeat on the saved-playlist detail screen.

marcodejongh:
All four non-blocking items landed in `25da9285b`, plus the nits. Thanks — N1 and N2 were both real and I would not have found either.

**N1 — `no-progress` was the less observable branch.** Confirmed the whole path you traced. This PR was making that shape *harder* to see than `main`: the drain stops with `climbs: []`, the stop-reason gate suppresses the #3891 canary, `setQueue` builds a one-item queue, nothing reported. Both stops now report, under `op: replace-queue-page-cap` / `replace-queue-no-progress` so triage can separate a 2,000-climb runaway from count/select drift. Pinned by a new mobile test asserting the `climbs: [], hasMore: true` case specifically; reverting the condition to page-cap-only turns exactly that test red.

**N2 — abort now asks the signal.** Took your preferred fix (drain-side, so `fetchPlaylistSuggestionClimbs` is covered too) *and* the outer catch, since they fail independently. New shared test throws `Error('fetch failed: The operation was aborted.')` — no `name`, expo/fetch's exact shape — with `isRetryable: () => true` to make the misclassification maximally likely; it must still reject as `AbortError` with one call and no sleep. Reverting to the shape check turns only that test red.

**N3 — real-parser coverage.** Two mobile cases through the genuine `ClientError` shape: `retryAfterSeconds: 1` re-issues the page and fills the queue; `46` makes exactly one call, and I assert the **original** error object reaches `reportHandledError` so `isGraphqlRateLimitedError` can still read `extensions.code`. That also gives divergence (a) the coverage it was missing.

**N4 — cap 30 → 20.** Chose the cap over a global budget: 20 × 2 = **40** requests worst case against the 60/min bucket, 20 of headroom in a bucket shared with the list screen's paging. 30 sat exactly on the trip point, which as you say is the thing this guard exists to avoid. Arithmetic is in a comment next to the constant, along with a note that the 3 s ceiling is per page and a fully rate-limited drain could still sleep ~60 s total.

**Nits:** dead exports removed; `stopAfterPage` now reports `'caller-stop'` instead of borrowing `'page-cap'`; dedupe marks as it goes so intra-page duplicates are dropped too (the code now does what the PR body claimed).

**Nit 4 — you are right and I have corrected the PR body.** The sheet test pins *presence*, not ordering; moving `setPendingReplacement(null)` after `showToast` leaves it green, because both are state updates in the same continuation and React 19 batches them into one commit. The presence is the fix — on `main` the catch never dismissed the sheet at all. Nit 5 fixed too: the helper is `isGraphqlRateLimitedError`, not `parseRateLimitError`.

Validation: playlists-react 45/45, mobile 6940/6940, `typecheck:mobile` / `typecheck:web` / `typecheck:playlists-react` / `vp check` all clean. Sentry gate still unrun and still flagged in the description.

PR #4718, opened 2026-08-24
fix(mobile): give the Android tick note room for a real beta note

### LINKED ISSUE #4642: 🐞 Bug report: When i send a route and i want to write a comment i cant see the text. it is there but the text window...

## Comment

When i send a route and i want to write a comment i cant see the text. it is there but the text window is to small

## Metadata

| Field | Value |
| --- | --- |
| Platform | android |
| App version | 2.3.1 (2000824) |
| Source | drawer-bug |
| Board | kilter / layout 1 / size 10 / sets [1,20] @ 55° |
| Climb | A Fuori (7a/V6) |
| Session | unknown |
| URL | /climbs |
| User agent | unknown |
| Feedback row | app_feedback #10206 |

## Contact

✅ Reporter agreed to follow-up — resolve **app_feedback #10206** in the DB to reach them.

marcodejongh:
## Summary

Fixes #4642.

The reporter filed one complaint with two halves, and they resolve differently.

**Half 1 — "i cant see the text. it is there".** Already fixed. The note rendered as a ~5pt band of glyph bottoms and a stub caret; that landed on `main` in #4684 / `245f9ca1b` on 2026-08-22 and shipped on the `production` OTA channel on 2026-08-23, two days after this issue was filed against 2.3.1. #4684 was merged without the emulator pass its own test plan asked for, and it linked #4231 (an unrelated gym-claim ticket, closed 2026-08-11) rather than this issue — which is why #4642 is still open. This PR closes that loop: it verifies half 1 on a device, states the mechanism honestly, and guards it.

**Half 2 — "the text window is to small".** Still live, and this PR fixes it.

### What the emulator showed

Android emulator, Pixel AVD at 810x1800 px / density 320 (= 405x900pt, close to the reporter's 411x914pt device), dark, current `main`. Measurements are `uiautomator` node bounds on the `Tick note` node.

| | Measured on `main` |
| --- | --- |
| Note field at rest | 44pt — one line |
| Note field with a long note | **95.5pt** — pinned at the 96pt `maxHeight`, four lines |
| Empty sheet between the Note row and the Attempt/Send bar, keyboard down | **~302pt** |
| Sheet body height above the keyboard | ~293pt |

Past four lines the note scrolls **inside** the field. An Android multiline `TextInput` has no scrollbar prop at all in RN 0.86, so there is nothing on screen to say text exists above the fold — a six-line note opened mid-sentence on *"before the dyn beta is solid now and I wan…"*, with the first lines silently gone. That is exactly "the text window is to small", and the sheet had ~300pt of empty space it was not using.

### The change

| | before | after |
| --- | --- | --- |
| `minHeight` | 44 — one line | **64** — two lines |
| `maxHeight` | 96 — four lines | **160** — seven lines |
| label alignment on the note row | centred against the box | **top-aligned** to the first line of text |
| border at rest | `transparent` | `systemColors.separator` |

**`maxHeight` 96 → 160.** Seven lines (7 × 20pt `subheadline` line-height + 16pt padding + 2pt border = 158, 2pt spare), +67% on the old ceiling. A maxed note spends 116 of the ~302pt of measured dead space.

The binding constraint is not Android but **iOS**, and this ceiling was 180 until review re-derived that bound. On the reference device the detent test pins (window 844 − 44 top inset − 24 iOS-26 card gap = base 776): the `'92%'` keyboard detent gives a 694pt column, `KeyboardAvoidingView behavior="padding"` takes 336 for the keyboard → 358, less the 56pt header and the footer. The footer is where the two readings split. The 180 figure assumed the window bottom inset collapses to 0 while the keyboard is up (footer 106 → body 196pt); it does not — `useWindowBottomInset()` keeps returning the published 34pt inset and nothing in the app zeroes it on keyboard show (`hooks/use-window-bottom-inset.ts:20-24`), so the footer is 140 and the **visible body is 162pt**. 180 overflowed it by 18pt, and a field taller than the visible body can never scroll fully into view — the exact rule the old comment stated and its own value broke.

160 is safe under both readings. The derivation, including which assumption killed 180, is now in the constant's comment, so nobody re-raises the ceiling on the old argument.

**`minHeight` 44 → 64.** Two lines of `subheadline` plus the field's own padding and border, matching `EndSessionSheet`'s recap box. A one-line box to start a beta note in was half the complaint still shipping.

**`alignTop` on `TickFormRow`.** `TickFormRow` centres its label against the control, which is correct for every row on the 56pt beat and wrong for the one control that now grows to 160pt — the label ended up level with the middle of the note rather than its first line. New opt-in prop (row `alignItems: 'flex-start'` + `paddingTop: spacing[1]`, label `paddingTop: spacing[2]`), passed **only** from `QuickTickBar` and `LogbookEditSheet`, so every other row is untouched. It defers to `stacked`, which already puts the label on its own line. Verified on the emulator at a 178pt field (the tallest the pre-review 180pt ceiling allowed): label top and field top land on the same pixel, was ~31pt below the first text line. The alignment is independent of the ceiling, so lowering it to 160 does not change that geometry.

One caveat for the next adopter, deliberately left as-is: the inset is **top-only** (`paddingTop: spacing[1]`, no matching `paddingBottom`). Invisible on the two rows that opt in today — both pass `showSeparator={false}` and are last in their sheet — but the first row that takes `alignTop` *with* a separator gets its hairline flush against the control's bottom edge with 4pt above it, which reads as a rendering bug rather than a missing prop. Mirror the inset, or keep pairing `alignTop` with `showSeparator={false}`.

**Visible edge at rest.** `borderColor: 'transparent'` meant the unfocused field had no boundary at all. Now `systemColors.separator`, with `borderWidth` unchanged so focus recolours rather than resizes. Being straight about it: it clears WCAG 1.4.11's 3:1 in **neither** scheme — **~1.9:1 in dark, ~1.6:1 in light**, so dark is the better of the two rather than the worst case. The design system has no border role that clears 3:1 on both platforms. Tracked in **#4722**, not solved here.

**Deliberately not touched — and two independent emulator passes reached this same conclusion:**

- **The detents.** `CREATE_TICK_SNAP_POINTS[0]` is discarded before it reaches the sheet on Android (`sheet-snap-points.ts:45`, under `androidOpensExpanded`), and `@expo/ui`'s Android sheet never reads a `%` for sizing — so a detent edit changes zero pixels on the reported platform while forcing `log-ascent-sheet.test.tsx` to be re-baselined. The body is scrollable under a pinned footer, so the taller field costs the action bar nothing.
- **`TickFormRow`'s `height` prop.** Applied as `minHeight` (`TickFormRow.tsx:65`), so the row already sizes to a taller child. Inert.
- **Parking the caret at the start on blur.** An overflowing note stays scrolled wherever the caret left it, so blurring can leave you looking at the middle of your own note. The taller box makes that far rarer but does not remove it. Fixing it means `setNativeProps({ selection })` during blur, which touches Android IME composition and needs a device matrix rather than one emulator. Filed as **#4730**.

### On the mechanism, honestly

#4684's commit message credits *"nested-flex vertical centering"* for the sliver. That does not survive the arithmetic: the old outer box was 44 − 16 padding − 2 border = 26pt inner, and a 20pt line fits in 26pt. Nor does the stronger version of that story — this repo's vendored Yoga clamps `availableInnerMainDim` to the min constraint *before* `resolveFlexibleLength` (`CalculateLayout.cpp`), so the old child resolved to ~26pt, not 0.

What **is** established: the ink band measured ~5pt, the first ink sat ~9.7pt below the input's top edge despite `textAlignVertical: 'top'`, and declaring padding on the input fixed it. The **leading hypothesis** for the ~9.7pt is Fabric's — `AndroidTextInputComponentDescriptor.h:98-102` writes the theme's default `EditText` padding into the Yoga style of any `TextInput` that declares none, and declaring padding sets `hasPaddingVertical` and suppresses it. The header comment now says exactly that much and no more. Either way the padding is what changed the rendering, so `tick-note-field.test.tsx` guards it — and a future "simplify" pass that keeps the flat shape but drops the padding now fails a test instead of re-opening the bug.

### Review follow-ups (applied in `f6135165c`)

- **The ceiling: 180 → 160.** Re-derived above. The old number rested on an unstated assumption about the window bottom inset that the code does not honour.
- **Detent derivations un-drifted.** `tick-sheet-metrics.ts` still derived both sheets from a 56pt note row while this PR made the row **68** (field `minHeight` 64 + the 4pt `alignTop` inset). So the create column is 496, not 484 (+20 chrome → 516/769 = 67.1%), and the edit column 686, not 674 (706/769 = 91.8%). **Detent values are unchanged** — `['65%', '92%']` / `['80%', '92%']` still ship, and `log-ascent-sheet.test.tsx` needs no re-baseline. What changed is that the comments now say the truth: at `'65%'` the 480pt column opens ~16pt short of the content, deliberately, because the body scrolls under a pinned footer and erring short beats a clipped footer (#3330). That module exists to stop the field size and the detent arithmetic drifting apart, so it does not get to drift itself.
- **WCAG figures corrected** — both schemes stated, above.
- **Two assertions no longer fail via `NaN`.** `minHeight` / `maxHeight` living on a wrapper instead of the input made every number downstream `NaN`; the tests went red (no false green) but reported *"expected NaN to be >= 40"*. They now check the property is on the input first, and the regression reads *"expected undefined to be type of 'number'"* — mutation-checked by moving both off the input.

### Cross-reference cleanup

Six comments cited `#4231` for the Android tick-sheet work: `sheet-snap-points.ts`, `Sheet.tsx`, `ModalSheet.tsx`, `LogAscentSheet.tsx`, `you/LogbookEditSheet.tsx`, `docs/mobile-sheets-vs-routes.md`. Every one of them describes the **pinned-footer / Send-button-below-the-fold** bug — `androidOpensExpanded`, Android's ~50% partial state stranding the footer — **not** the note field. So they point at **#4723**, the issue filed for that, not at this one. (An earlier revision of this PR pointed them at #4642; that was wrong for the same reason #4231 was, and is corrected in `f38f4ec1`.)

## Release Notes

- Room for real beta when you log a send: the note box opens two lines tall and grows to hold a full paragraph, so what you're writing stays on screen instead of scrolling out of sight.

## Test plan

Emulator: Android, Pixel AVD 810x1800 @ density 320, dark, dev-client against Metro from this branch. Climbs → a climb → log a send → Note.

- [x] **Short note** (`gogogo`) on current `main`: full-height glyphs, full-height caret, light violet on the dark violet fill. Half 1 confirmed fixed by #4684 — the check #4684 never ran.
- [x] **Long note** on current `main`: field pinned at 95.5pt / four lines, note rendering from mid-sentence, ~302pt of empty sheet below it. Half 2 reproduced.
- [x] **This branch, at rest**: field 64pt / two lines, label level with the placeholder's first line, visible border before focus.
- [x] **This branch, long note**: field grows 98 → 118 → 158 → 178pt reading from its first word, clamps at exactly 180pt, label top and field top on the same pixel, Attempt/Send on screen above the keyboard, body scrolls the rows above. Captured at the 180pt revision; **not re-captured at 160**, which caps the same field one 20pt line earlier (158pt rendered) and changes nothing else on screen.
- [x] `vp check` — my files clean (scoped run: 16 files, no warnings or errors); the full-repo run reports 8 errors + 259 warnings, all pre-existing in `packages/backend` and `packages/db` and untouched here
- [x] `vp test run --project mobile tick-note-field tick-form-row log-ascent` — 4 files, 31 tests, all pass
- [x] `vp run typecheck:mobile` — clean
- [x] `vp run test:mobile` — 703 files, 6940 tests, all pass
- [x] `vp run check:mobile-bundle` — android bundle OK

Every guard added here was **mutation-tested** — each fails on its own regression and stays green for the others:

| Mutation | Fails |
| --- | --- |
| drop `paddingVertical` | padding guard only |
| `maxHeight` back to 96 | ceiling guard only |
| `minHeight` back to 44 | two-line guard only |
| `borderColor` back to `transparent` | visible-edge guard only |
| wrap the input in a `View` | flat-shape guard only |
| `alignTop` ignored | opt-in guard only |
| `alignTop` applied unconditionally | default-beat guard only |
| `minHeight` / `maxHeight` moved off the input | both height guards, now naming the moved property instead of `NaN` |

JS-only, so this rides the production OTA — no native rebuild, no fingerprint change.

Two independent emulator passes (this one and a parallel session) reproduced the same 4-line cap and independently rejected the detent bump and the `height` prop.

The reporter's screenshot is on #4642. GitHub image upload is not reachable from this environment, so the numbers above are node bounds rather than pixel-peeping, and the captures are laid out side by side here: **https://claude.ai/code/artifact/1fa50a2c-1e7c-4174-961a-11644245bed4** (Boardsesh team only). The PNGs are at `.boardsesh/shots/` in the working tree — happy to have them dragged into this body by anyone with a browser session.

### DIFF
--- a/docs/mobile-sheets-vs-routes.md
+++ b/docs/mobile-sheets-vs-routes.md
@@ -146,7 +146,7 @@ and which index is requested — it has a fixed ~50% "partial" state and a conte
 "expanded" state, nothing in between (see `androidSafeSnapPoints` in
 `src/components/sheet-snap-points.ts`). A sheet whose detents are tuned against iOS's real first
 fraction (e.g. `65%`/`80%`, sized so a pinned footer fits under the form) can be TALLER than
-Android's ~50% partial state, stranding that footer below the fold (#4642). For a multi-detent
+Android's ~50% partial state, stranding that footer below the fold (#4723). For a multi-detent
 sheet with a pinned footer, pass `androidOpensExpanded` to `Sheet` / `ModalSheet` — `LogAscentSheet`
 and `LogbookEditSheet` opt in. It collapses the sheet to its single LAST detent on Android (via
 `androidSafeSnapPoints`) rather than requesting the last index of a multi-detent config: a single
--- a/docs/mobile-sheets-vs-routes.md
+++ b/docs/mobile-sheets-vs-routes.md
@@ -146,7 +146,7 @@ and which index is requested — it has a fixed ~50% "partial" state and a conte
 "expanded" state, nothing in between (see `androidSafeSnapPoints` in
 `src/components/sheet-snap-points.ts`). A sheet whose detents are tuned against iOS's real first
 fraction (e.g. `65%`/`80%`, sized so a pinned footer fits under the form) can be TALLER than
-Android's ~50% partial state, stranding that footer below the fold (#4231). For a multi-detent
+Android's ~50% partial state, stranding that footer below the fold (#4642). For a multi-detent
 sheet with a pinned footer, pass `androidOpensExpanded` to `Sheet` / `ModalSheet` — `LogAscentSheet`
 and `LogbookEditSheet` opt in. It collapses the sheet to its single LAST detent on Android (via
 `androidSafeSnapPoints`) rather than requesting the last index of a multi-detent config: a single
--- a/packages/mobile/src/components/LogAscentSheet.tsx
+++ b/packages/mobile/src/components/LogAscentSheet.tsx
@@ -144,7 +144,7 @@ export function LogAscentSheet({
       surface="solid"
       footerSurface="flush"
       // Android's ~50% partial state can't fit this form under a pinned
-      // footer (see `androidOpensExpanded` on `ModalSheet`, #4231) — open
+      // footer (see `androidOpensExpanded` on `ModalSheet`, #4642) — open
       // expanded there so the Send button never lands off the fold.
       androidOpensExpanded
       header={
--- a/packages/mobile/src/components/LogAscentSheet.tsx
+++ b/packages/mobile/src/components/LogAscentSheet.tsx
@@ -144,7 +144,7 @@ export function LogAscentSheet({
       surface="solid"
       footerSurface="flush"
       // Android's ~50% partial state can't fit this form under a pinned
-      // footer (see `androidOpensExpanded` on `ModalSheet`, #4642) — open
+      // footer (see `androidOpensExpanded` on `ModalSheet`, #4723) — open
       // expanded there so the Send button never lands off the fold.
       androidOpensExpanded
       header={
--- a/packages/mobile/src/components/ModalSheet.tsx
+++ b/packages/mobile/src/components/ModalSheet.tsx
@@ -71,7 +71,7 @@ type ModalSheetProps = {
   header?: ReactNode;
   /** Collapses to a single, expanded-only detent on Android instead of the usual
    * multi-detent config, so a pinned footer never strands below Android's ~50%
-   * partial state (#4642). See `androidSafeSnapPoints` for the full rationale;
+   * partial state (#4723). See `androidSafeSnapPoints` for the full rationale;
    * no effect on iOS/web, or on a single detent already at/above 75%. */
   androidOpensExpanded?: boolean;
 };
--- a/packages/mobile/src/components/ModalSheet.tsx
+++ b/packages/mobile/src/components/ModalSheet.tsx
@@ -71,7 +71,7 @@ type ModalSheetProps = {
   header?: ReactNode;
   /** Collapses to a single, expanded-only detent on Android instead of the usual
    * multi-detent config, so a pinned footer never strands below Android's ~50%
-   * partial state (#4231). See `androidSafeSnapPoints` for the full rationale;
+   * partial state (#4642). See `androidSafeSnapPoints` for the full rationale;
    * no effect on iOS/web, or on a single detent already at/above 75%. */
   androidOpensExpanded?: boolean;
 };
--- a/packages/mobile/src/components/Sheet.tsx
+++ b/packages/mobile/src/components/Sheet.tsx
@@ -71,7 +71,7 @@ type SheetProps = {
   header?: ReactNode;
   /** Collapses to a single, expanded-only detent on Android instead of the usual
    * multi-detent config, so a pinned footer never strands below Android's ~50%
-   * partial state (#4231). See `androidSafeSnapPoints` for the full rationale;
+   * partial state (#4642). See `androidSafeSnapPoints` for the full rationale;
    * no effect on iOS/web, or on a single detent already at/above 75%. */
   androidOpensExpanded?: boolean;
 };
--- a/packages/mobile/src/components/Sheet.tsx
+++ b/packages/mobile/src/components/Sheet.tsx
@@ -71,7 +71,7 @@ type SheetProps = {
   header?: ReactNode;
   /** Collapses to a single, expanded-only detent on Android instead of the usual
    * multi-detent config, so a pinned footer never strands below Android's ~50%
-   * partial state (#4642). See `androidSafeSnapPoints` for the full rationale;
+   * partial state (#4723). See `androidSafeSnapPoints` for the full rationale;
    * no effect on iOS/web, or on a single detent already at/above 75%. */
   androidOpensExpanded?: boolean;
 };
--- a/packages/mobile/src/components/play-drawer/QuickTickBar.tsx
+++ b/packages/mobile/src/components/play-drawer/QuickTickBar.tsx
@@ -121,7 +121,10 @@ export const QuickTickBar = React.memo(function QuickTickBar({ form }: QuickTick
         />
       </TickFormRow>
 
-      <TickFormRow label={tClimbs('mobile.tick.noteLabel')} showSeparator={false} testID="tick-row-note">
+      {/* `alignTop`: the note is the one control that grows well past the 56pt
+          row beat, and a centred label floats ~31pt below the first line of
+          the climber's own text (#4642). */}
+      <TickFormRow label={tClimbs('mobile.tick.noteLabel')} alignTop showSeparator={false} testID="tick-row-note">
         <TickNoteField
           value={comment}
           onChangeText={onCommentChange}
--- a/packages/mobile/src/components/sheet-snap-points.ts
+++ b/packages/mobile/src/components/sheet-snap-points.ts
@@ -22,7 +22,7 @@ const ANDROID_NEAR_FULL_DETENT_PERCENT = 75;
  * already have a partial state, and non-% (px) detents are left as-is.
  *
  * `androidOpensExpanded` (a multi-detent sheet whose pinned footer only fits at
- * its LAST detent, e.g. the tick sheets, #4231) collapses to that single last
+ * its LAST detent, e.g. the tick sheets, #4642) collapses to that single last
  * detent on Android — not "keep both detents and pick the last index at
  * present-time", which depends on an `@expo/ui` Android `expand()` call the
  * library's own `ModalBottomSheetView.kt` can silently no-op ("Expanded anchor
--- a/packages/mobile/src/components/sheet-snap-points.ts
+++ b/packages/mobile/src/components/sheet-snap-points.ts
@@ -22,7 +22,7 @@ const ANDROID_NEAR_FULL_DETENT_PERCENT = 75;
  * already have a partial state, and non-% (px) detents are left as-is.
  *
  * `androidOpensExpanded` (a multi-detent sheet whose pinned footer only fits at
- * its LAST detent, e.g. the tick sheets, #4642) collapses to that single last
+ * its LAST detent, e.g. the tick sheets, #4723) collapses to that single last
  * detent on Android — not "keep both detents and pick the last index at
  * present-time", which depends on an `@expo/ui` Android `expand()` call the
  * library's own `ModalBottomSheetView.kt` can silently no-op ("Expanded anchor
--- a/packages/mobile/src/components/tick/TickFormRow.tsx
+++ b/packages/mobile/src/components/tick/TickFormRow.tsx
@@ -29,7 +29,7 @@ type TickFormRowProps = {
    *  row's trailing gutter so the rail reads as scrollable rather than clipped. */
   bleed?: boolean;
   /** Top-align the label against a control that is much taller than the 56pt
-   *  beat — the note field, which grows to 180pt. Centred, its label floats
+   *  beat — the note field, which grows to 160pt. Centred, its label floats
    *  ~31pt below the first line of the climber's own text and reads as broken.
    *  Opt-in, so every other row keeps the centred beat. NOT `stacked`: that
    *  flips the row to a column and is a different layout. */
--- a/packages/mobile/src/components/tick/TickFormRow.tsx
+++ b/packages/mobile/src/components/tick/TickFormRow.tsx
@@ -28,6 +28,12 @@ type TickFormRowProps = {
   /** Let the control run to the screen edge (a horizontal rail), dropping the
    *  row's trailing gutter so the rail reads as scrollable rather than clipped. */
   bleed?: boolean;
+  /** Top-align the label against a control that is much taller than the 56pt
+   *  beat — the note field, which grows to 180pt. Centred, its label floats
+   *  ~31pt below the first line of the climber's own text and reads as broken.
+   *  Opt-in, so every other row keeps the centred beat. NOT `stacked`: that
+   *  flips the row to a column and is a different layout. */
+  alignTop?: boolean;
   /** Dim + inert, without unmounting: the edit sheet's Tries row when the tick
    *  is a flash. Keeping the row mounted preserves the form's vertical rhythm
    *  and the control's state. */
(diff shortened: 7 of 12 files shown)


marcodejongh:
Thanks — took both. One landed, one didn't survive checking.

**#2, the assertion side-effect: fixed** in `0b7ece81`. You were right that an `expect` inside a helper three tests share reports the failure at the wrong place. Rather than move the check, I removed the need for it: `totalVerticalPaddingOf` now returns top + bottom instead of one edge, so the arithmetic holds whichever way the padding is spelled and no edge-equality assumption exists to assert. Re-mutation-tested — each of the four guards still fails on its own regression and nothing else.

**#1, the comment blocks: declining, because the rule quoted isn't in CLAUDE.md.**

```
$ grep -rn -i "multi-paragraph|multi-line comment|one short line|docstring" CLAUDE.md
(no matches)
```

Nothing in this repo's CLAUDE.md constrains comment length. The house style in this exact directory is the opposite — leading rationale blocks of 9 lines (`TickFormRow.tsx`, `tick-sheet-metrics.ts`), 12 (`TickActionBar.tsx`) and 13 (`tick-token-discipline.test.ts`).

On the substance: "the test already guards the preconditions, that's the right place for the enforcement" is exactly the trap this PR exists to close. The test guards *that* the padding is there. It cannot say *why*, and the why is the whole bug — on Fabric, `AndroidTextInputComponentDescriptor` writes the theme's default `EditText` padding into the Yoga style of any `TextInput` that declares none. #4684 fixed this for a reason its own comment got wrong (it credits nested-flex centring; the arithmetic is 44 − 16 − 2 = 26pt, and a 20pt line fits in 26pt), and the fix stood undocumented for two days. A PR description is not read by whoever opens this file in six months to tidy the padding away.

Worth flagging that the quote was presented verbatim with a file attribution. Please verify citations before treating them as blocking.

marcodejongh:
## Review — ship with the fixes below. Nothing blocking.

Read-only pass at `f38f4ec1c`. I went after the four things that could actually bite — `alignTop` on a shared primitive, the height arithmetic, the focus-border resize, and whether the new tests go red against the broken revision — and found no correctness defect. Local re-run: `vp test run --project mobile tick log-ascent logbook-edit` → 16 files / 136 tests pass.

### Non-blocking

**1. The iOS `maxHeight` ceiling rests on an unstated assumption, and the other reading of it fails.** `TickNoteField.tsx:95-99`

Re-derived against the repo's own reference device (window 844 / top inset 44 / card gap 24 → base 776, the one `log-ascent-sheet.test.tsx:404` pins): `'92%'` column = `round(776 × 0.92) − 20` = **694**; KAV pad = `(150 + 694) − 508` = **336**; less header 56 → **302**. The footer is where it splits:

- `windowInsetBottom → 0` while the keyboard is up → footer 106 → visible body **196pt**. That reproduces your number exactly, so this is the assumption.
- `windowInsetBottom` stays 34 → footer 140 → visible body **162pt**. Nothing in the repo zeroes it on keyboard show (`use-window-bottom-inset.ts:20-24` just returns the root-published `useSafeAreaInsets().bottom`).

Under the second reading 180 is 18pt taller than the visible body — the comment's own rule broken by its own value. Scenario: iOS, keyboard up, 8+ line note; the line you're typing can sit under the pinned footer with no scroll position that shows the whole field. Cosmetic and rare, but it's the platform this PR couldn't run.

Cheapest fix: `maxHeight: 160` — 7 lines (`7 × 20 + 16 + 2 = 158`), safe under both readings, still +67% on today. Otherwise measure it: `expo:eas-simulator` gets a cloud iOS sim from a Linux host, which is exactly what the derivation is missing.

**2. The detent derivations now under-count the note row by 12pt.** `tick-sheet-metrics.ts:66` and `:78-80`

Both still say `note 56`. The row is **68** at rest now — field 64 plus the new `paddingTop: spacing[1]` from `alignTop` (`TickFormRow.tsx:77`). Create column 484 → 496. Keeping `'65%'` is right and I'm not asking you to move it — but it means the iOS create sheet opens with ~12pt of the note row below the fold, and this module exists precisely so the field size and the detent arithmetic can't drift apart. Update both derivations (create 496, edit 686) and record that the detent is deliberately short because the body scrolls. No value change, so no test re-baseline.

**3. The WCAG caveat is accurate but scoped as if dark were the worst case.** `TickNoteField.tsx:67-69`

Reproduced your number: separator over the fill, against the sheet ground, is **1.87:1** in dark. But light is worse (**~1.6:1** by the same method) and the border's inner edge against the fill is ~1.15:1. Still a strict improvement — `transparent` gave the field a 1.26:1 boundary in dark and no edge at all. Suggest dropping the `on Android dark` qualifier.

**4. Two of the five note-field assertions go red via `NaN`.** `tick-note-field.test.tsx:113-115`, `:148-151`

I mutation-checked all five against the real broken revision (`git show acfa3bbee:…/TickNoteField.tsx`) — all five fail there, so no false green, and the padding guard is on the right element. But `minHeight`/`maxHeight` lived on the wrapper at that revision, so those two fail as *"expected NaN to be >= 40"*, which doesn't tell the next person the property moved off the input. One line each: `expect(style.minHeight).toBeTypeOf('number')`.

**5. `alignTop`'s inset is asymmetric, and it's now public API.** `TickFormRow.tsx:77`

4pt above, none below. Invisible today because both adopters pass `showSeparator={false}` and are last in their sheet. The first row that opts in *with* a separator gets a hairline flush against the control's bottom edge. Mirror the padding, or put the constraint in the JSDoc at `:31-35`.

### Confirmed correct

- **`alignTop` is strictly opt-in.** All 12 call sites enumerated; the ten that don't pass it get `paddingTop: 0` against styles that declare no `paddingTop` — byte-identical to `main`, `alignItems: 'center'` and the 56pt beat intact, and pinned by a test. `alignTop && !stacked` handles the collision, also tested.
- **The label really lands on the first line**: first-line centres at 24 (label) vs 23 (field) on Material, 23 vs 23 on Liquid Glass. Both magic numbers work on both variants.
- **Heights recompute**: `64 − 16 − 2 = 46 ≥ 2 × 20`; `180 − 16 − 2 = 162 → 8` lines. Both variants share `lineHeight: 20`.
- **Focus can't resize the box** — `borderWidth: 1` is in the static `StyleSheet`, only `borderColor` is in the dynamic object.
- **All six references repointed, none missed, none wrong.** Zero `#4231` left in `packages/` or `docs/`; all six `#4723` hits sit on `androidOpensExpanded` / the partial-state footer bug, and #4723 is the right issue. Nothing repointed to #4642 by mistake.
- **Scope held**: `tick-sheet-metrics.ts` and `log-ascent-sheet.test.tsx` are both absent from the diff, no `height` prop passed.
- **The mechanism comment is honestly calibrated** — imperative first, observation and hypothesis separated, refuted alternative named. Neither overclaims nor hedges into uselessness.

### On the earlier bot review

I re-ran the check on the comment-length rule: `grep -rn -i "multi-paragraph|multi-line comment|comment block|docstring" CLAUDE.md .claude/` returns nothing. No such rule exists in this repo — declining it was right, and it shouldn't be re-litigated. The `= 58` vs `minHeight: 64` arithmetic it flagged is still there at `TickNoteField.tsx:82-85` and is worth a one-word fix while you're in the file.

marcodejongh:
Independent check of the "On the mechanism, honestly" section, from a separate pass over #4642. Both of its claims hold in the vendored sources, and I think the padding story can be stated more strongly than "leading hypothesis" — the missing piece is the shape of the **2.3.1** component itself.

**1. The flex-collapse story is refuted.** Confirmed at
`packages/mobile/node_modules/react-native/ReactCommon/yoga/yoga/algorithm/CalculateLayout.cpp:1614-1616`:

```cpp
if (yoga::isDefined(minInnerMainDim) &&
    flexLine.sizeConsumed < minInnerMainDim) {
  availableInnerMainDim = minInnerMainDim;
}
```

That runs before the flex distribution, so the old child could not resolve to 0. Same conclusion as the PR.

**2. The Fabric path is real, and 2.3.1 walked straight into it.** `AndroidTextInputComponentDescriptor.h:101-108`:

```cpp
if (!textInputProps.hasPadding && !textInputProps.hasPaddingTop && !textInputProps.hasPaddingVertical) {
  changedPadding = true;
  style.setPadding(yoga::Edge::Top, yoga::StyleLength::points(theme.top));
}
```

The part that makes this conclusive rather than plausible: in `5554e6522` the padding was on the **wrapper `View`**, and the input declared none at all —

```jsx
<View style={[styles.box, { paddingHorizontal: spacing[3], paddingVertical: spacing[2], ... }]}>
  <BottomSheetTextInput style={[textStyles.subheadline, styles.input, { color: systemColors.label }]} />
```

with `styles.input = { flex: 1, textAlignVertical: 'top' }`. No `padding`, no `paddingTop`, no `paddingVertical` — so the guard above passes and the theme padding gets written into the input's Yoga style. #4684 fixed it by moving `paddingVertical` onto the input, which sets `hasPaddingVertical` and suppresses the injection. That is the same before/after the PR describes, now with the reason the guard fired.

**The arithmetic closes on the reporter's pixels.** Wrapper inner box = `44 − 2 border − 16 padding = 26pt`. Take the ~10pt top + ~10pt bottom the emulator's 9.7pt first-ink offset implies, and the input's content box is ~6pt. Sampling the attachment on #4642 (density 2.625): the ink band is **14px = 5.3pt** including the caret. ~6pt predicted vs 5.3pt measured, and the 9.7pt offset independently pins `theme.top` at ~10pt.

So the three numbers agree: the 9.7pt offset you measured, the 5.3pt band in the reporter's screenshot, and the 26pt box the source gives. I'd be comfortable dropping the "best hypothesis rather than proven" hedge in the header comment — though keeping it costs nothing and the test guards the behaviour either way.

Nothing here asks for a change to the diff. One thing this pass did turn up that the PR does not cover: the caret in the reporter's screenshot samples exactly `#80CBC4` (Material Teal 200) because no RN `TextInput` in the app sets `cursorColor`/`selectionColor` — filed separately as #4744, cosmetic.

PR #4727, opened 2026-08-24
feat(mobile): show who queued each climb in a shared session

### LINKED ISSUE #4672: Queue: show who queued each climb (the data is already there)
`ClimbQueueItem` carries `addedByUser {id, username, avatarUrl}` — stamped on add, synced over GraphQL, preserved through climb re-resolution. Nothing renders it: `QueueItemRow` has no attribution prop, and a grep for `addedByUser` across mobile UI returns only model/wire hits. Session history rows have the same gap.

Proposed: a small avatar (compose `BoardDriverAvatar`) on queue + history rows in a shared session. Makes the queue legible as a crew artifact ("whose climb is next") and grounds the fairness conversations from the #4281 thread.

Refs #4281

marcodejongh:
Closes #4672

## What

Queue rows already carry `addedByUser` end to end — the provider stamps it on every local add, it survives the wire in both directions, and it round-trips through the offline snapshot. Nothing ever drew it. This renders it.

During a party session, each queue row shows a small (20dp) avatar of the climber who put that climb up. It appears on all three row kinds — history, current, and upcoming — because `QueueList` renders `QueueItemRow` from three separate call sites and the ticket calls out history rows explicitly.

## How it's gated

Two conditions, both deliberate:

1. **`sessionId != null`**, not `useIsSharedSession()`. That hook resolves to `sessionActive && distinctUserCount > 1`, so the moment a peer left — or presence flapped — every face would vanish off a queue still full of the crew's climbs.
2. **Per-row self-exclusion.** `attributeNewItem` stamps *solo* adds with this device's identity too, so without excluding `addedByUser.id === viewerUserId` a one-person session would put your own face on every row you queued.

A face therefore means exactly "someone else queued this".

Absent, self, or nameless attribution renders **no element at all** — not an empty slot. `Avatar` with neither a uri nor a name draws a `?` glyph on a primary-coloured disc, which is the broken-looking placeholder the ticket warns about. The avatar is also suppressed in edit mode, following the same rule the sent-at-angle chip already uses (edit mode is the row's widest state).

The avatar is decorative and non-interactive: it carries `pointerEvents="none"` and is hidden from screen readers, because the row's own accessibility label now ends with "Added by {name}". No `userId` is passed to `BoardDriverAvatar`, which keeps it on the non-interactive branch rather than nesting a Pressable inside the row's gesture arena.

## Notes

- Per-row cost is O(1) off the row's own item — no roster read, no Map lookup, no new array. `renderRow`'s deps gain two primitives.
- `resolveQueueRowAttribution` is a pure function in `packages/mobile/src/lib/queue-attribution.ts` so the decision table is unit-testable with no mocks.
- JS/TS/JSON only — no `app.config.ts`, `plugins/**`, `modules/**`, `patches/**` or dependency change, so the runtime fingerprint is unchanged and this ships OTA.

## Tests

- `queue-attribution.test.ts` — 11 cases over the decision table, including the `avatarUrl: undefined` normalisation and an assertion that the render payload never leaks a user id.
- `queue-item-row-memo.test.tsx` — attribution rendering, edit-mode suppression, the accessibility label, and proof the row still bails out of a same-props re-render with the new props set. Also widens the file's `AccessibilityCapture` to record `accessibilityLabel`, which nothing was capturing before.
- `queue-list-added-by.test.tsx` (new) — asserts **all three** `QueueItemRow` call sites receive the props. Verified by mutation: deleting them from just the current-item call site turns this file red. Nothing else in the repo renders `QueueList` unmocked, so without this file a two-of-three wiring would have shipped green.

Validated: `vp check`, `vp run typecheck:mobile`, `vp run test:mobile`, `vp test run --project i18n`, `vp run check:i18n:orphans`, `vp run check:mobile-bundle`.

## Release Notes

- The queue now shows who put each climb up while you're in a session with your crew.
- A small face sits on every row someone else added — upcoming, current, and the ones you've already worked through. Your own adds stay clean.

### DIFF
--- a/packages/mobile/src/components/QueueItemRow.tsx
+++ b/packages/mobile/src/components/QueueItemRow.tsx
@@ -418,12 +418,15 @@ function QueueItemRowComponent({
   const showSentAtAngle =
     isHistoryItem && !isEditMode && typeof climbedAtAngle === 'number' && climbedAtAngle !== board.angle;
 
-  // Edit mode strips secondary chrome (same rule as the sent-at-angle chip) and
-  // is the row's widest state — the checkbox slot plus the delete affordance.
   const addedBy = resolveQueueRowAttribution(item.addedByUser, {
-    showAddedBy: showAddedBy && !isEditMode,
+    showAddedBy,
     viewerUserId,
   });
+  // Edit mode strips secondary chrome (same rule as the sent-at-angle chip) and
+  // is the row's widest state — the checkbox slot plus the delete affordance.
+  // Only the glyph goes: the accessibility label has no width budget, and a
+  // VoiceOver user needs to know whose climb they are about to bulk-delete.
+  const showAddedByAvatar = addedBy != null && !isEditMode;
 
   // A plain const, deliberately NOT a useMemo: `rowAccessibilityActions` above is
   // memoized because a test asserts its identity across re-renders, but a string
@@ -489,7 +492,7 @@ function QueueItemRowComponent({
         )}
 
         {/* Who queued it — decorative; the row's own label carries the name. */}
-        {addedBy && (
+        {showAddedByAvatar && addedBy && (
           <View
             pointerEvents="none"
             accessibilityElementsHidden
--- a/packages/mobile/src/components/QueueItemRow.tsx
+++ b/packages/mobile/src/components/QueueItemRow.tsx
@@ -15,6 +15,8 @@ import { Text } from './Text';
 import { Icon } from './Icon';
 import { ClimbListItemContent } from './ClimbListItemContent';
 import { THUMBNAIL_WIDTH } from './ClimbListThumbnail';
+import { BoardDriverAvatar } from './board-presence/BoardDriverAvatar';
+import { resolveQueueRowAttribution } from '../lib/queue-attribution';
 import { iosSystemColors } from '../theme/ios-colors';
 import { spacing } from '../theme/tokens';
 import { springs } from '../theme/animations';
@@ -69,6 +71,10 @@ type QueueItemRowProps = {
   queueIndex?: number;
   /** Whether this row may be dragged (future rows, not in edit mode). */
   isDraggable?: boolean;
+  /** Show who queued this climb. True only while a party session is active. */
+  showAddedBy?: boolean;
+  /** The viewer's own party-profile id; their own adds render no avatar. */
+  viewerUserId?: string | null;
 };
 
 function PositionIndicator({
@@ -121,6 +127,8 @@ function QueueItemRowComponent({
   rowIndex,
   queueIndex,
   isDraggable = false,
+  showAddedBy = false,
+  viewerUserId = null,
 }: QueueItemRowProps) {
   const { systemColors, brandColors } = useTheme();
   const { t } = useTranslation('session');
@@ -410,6 +418,21 @@ function QueueItemRowComponent({
   const showSentAtAngle =
     isHistoryItem && !isEditMode && typeof climbedAtAngle === 'number' && climbedAtAngle !== board.angle;
 
+  // Edit mode strips secondary chrome (same rule as the sent-at-angle chip) and
+  // is the row's widest state — the checkbox slot plus the delete affordance.
+  const addedBy = resolveQueueRowAttribution(item.addedByUser, {
+    showAddedBy: showAddedBy && !isEditMode,
+    viewerUserId,
+  });
+
+  // A plain const, deliberately NOT a useMemo: `rowAccessibilityActions` above is
+  // memoized because a test asserts its identity across re-renders, but a string
+  // has no such contract and memoizing it would only add a deps array to keep true.
+  const positionLabel = t('mobile.queue.positionLabel', { position });
+  const rowAccessibilityLabel = addedBy
+    ? `${climbName}, ${positionLabel}, ${t('mobile.queue.addedByAria', { name: addedBy.name })}`
+    : `${climbName}, ${positionLabel}`;
+
   const rowContent = (
     // touchAction="pan-y" (web only): RNGH otherwise defaults the row's DOM node
     // to `touch-action: none`, which blocks native touch-scrolling for any drag
@@ -420,7 +443,7 @@ function QueueItemRowComponent({
       <Animated.View
         accessible
         accessibilityRole="button"
-        accessibilityLabel={`${climbName}, ${t('mobile.queue.positionLabel', { position })}`}
+        accessibilityLabel={rowAccessibilityLabel}
         accessibilityState={{ selected: isEditMode ? isSelected : isCurrentClimb }}
         onAccessibilityTap={handlePress}
         accessibilityActions={rowAccessibilityActions}
@@ -463,6 +486,18 @@ function QueueItemRowComponent({
           >
             {t('mobile.queue.sentAtAngle', { angle: climbedAtAngle })}
           </Text>
+        )}
+
+        {/* Who queued it — decorative; the row's own label carries the name. */}
+        {addedBy && (
+          <View
+            pointerEvents="none"
+            accessibilityElementsHidden
+            importantForAccessibility="no-hide-descendants"
+            style={styles.addedByAvatar}
+          >
+            <BoardDriverAvatar uri={addedBy.avatarUrl} name={addedBy.name} size={20} status="none" />
+          </View>
         )}
 
         {/* Trailing action: tick (history) or drag handle (upcoming) */}
@@ -581,6 +616,9 @@ const styles = StyleSheet.create({
   sentAtAngle: {
     flexShrink: 0,
     fontVariant: ['tabular-nums'],
+  },
+  addedByAvatar: {
+    flexShrink: 0,
   },
   deleteAction: {
     position: 'absolute',
--- a/packages/mobile/src/components/__tests__/queue-item-row-memo.test.tsx
+++ b/packages/mobile/src/components/__tests__/queue-item-row-memo.test.tsx
@@ -50,6 +50,7 @@ type AccessibilityCapture = {
   onAccessibilityTap?: () => void;
   accessibilityActions?: { name: string; label?: string }[];
   onAccessibilityAction?: (event: { nativeEvent: { actionName: string } }) => void;
+  accessibilityLabel?: string;
 };
 const a11y = vi.hoisted(() => ({
   row: null as null | AccessibilityCapture,
@@ -112,6 +113,7 @@ vi.mock('react-native-reanimated', () => {
         onAccessibilityTap: rest.onAccessibilityTap,
         accessibilityActions: rest.accessibilityActions,
         onAccessibilityAction: rest.onAccessibilityAction,
+        accessibilityLabel: rest.accessibilityLabel,
       };
     }
     return createElement('div', null, children);
@@ -230,6 +232,14 @@ vi.mock('../ClimbListItemContent', () => ({
 
 vi.mock('../ClimbListThumbnail', () => ({ THUMBNAIL_WIDTH: 96 }));
 
+// Without this mock the new import drags PressableAvatar → expo-router's
+// useRouter and Avatar → Image/PixelRatio into a graph whose `react-native` mock
+// above exports only Platform, PlatformColor, View, Pressable and StyleSheet —
+// which would break every test in this file, not just the attribution ones.
+vi.mock('../board-presence/BoardDriverAvatar', () => ({
+  BoardDriverAvatar: ({ name }: { name?: string | null }) => createElement('span', { 'data-added-by': name ?? '' }),
+}));
+
 vi.mock('../../theme/tokens', () => ({ spacing: { 1: 4, 2: 8, 3: 12 } }));
 
 vi.mock('../../theme/animations', () => ({ springs: { interactive: {} } }));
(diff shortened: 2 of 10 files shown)


marcodejongh:
Thanks — took two of the three, pushed in 1b65df0.

**2. Label assertion on the raw key** — fixed. The `t` mock now resolves `mobile.queue.addedByAria`'s `{{name}}` (and only that key, so the rest of the file keeps its raw-key assertions), and the tests assert `'Added by Mina'`. Verified by mutation: dropping the name argument in `QueueItemRow` now fails with `Received: "Crimp Master, mobile.queue.positionLabel, Added by undefined"`. Good catch — the old assertion was vacuous.

**3. Missing false→true transition at the list level** — added. Note a plain rerender would have proved nothing here, since `QueueList` is `React.memo`'d and the props don't change when a session starts. The mocked hooks now sit on a small external store read via `useSyncExternalStore`, so flipping the session re-renders `QueueList` through the memo the way the real context change does — which also guards against the memo stranding rows on stale attribution.

**1. Narrow-device layout** — not changing this, but the failure mode is bounded rather than an overflow. The three trailing elements are `flexShrink: 0` siblings of `ClimbListItemContent`, whose `centerColumn` is `flex: 1, minWidth: 0` — so the extra ~32dp comes out of the climb name's width via `numberOfLines` truncation, not out of the row. Edit mode, the widest state (checkbox slot + delete affordance), is excluded outright by the `!isEditMode` guard.

The combined three-element state also needs a *history* row, in a session, at an angle different from the wall's — and RN layout isn't measured in this jsdom harness, so a unit test could assert the elements are all present but not that the result looks right. It's on the QA list for the two-device pass, alongside the half-opacity history-row avatar. If it reads badly on an SE, dropping the sent-at-angle chip when an avatar is present is the obvious follow-up.

marcodejongh:
All three checked, no code change needed:

- **Avatar spacing** — the parent does define one. `styles.row` is `flexDirection: 'row'` with `columnGap: spacing[3]` (`QueueItemRow.tsx:593`), the same gap that already separates the position slot, the climb content, the sent-at-angle chip and the trailing button. The avatar inherits it as a sibling; `flexShrink: 0` is there only so it doesn't get squeezed when a long climb name competes for width.
- **French** — agreed, and the referent here is `{{name}}` (the climber), not the climb. Left as is; the glossary's send/`enchaîné` rule doesn't apply to a queue add.
- **Row-kind assertions** — those assertions *are* the order anchor rather than a victim of it: if QueueList reordered its render sites, `props[0].isHistoryItem` would be `false` and the test fails loudly. That's deliberate — the comment above them says as much ("so a future row-order change can't quietly make this assertion vacuous"). The thing they protect against is the props assertion passing over three rows that aren't actually the three kinds.

marcodejongh:
## Code review — PR #4727

**Verdict: APPROVE WITH NITS.** No blocking findings. I went in expecting to find the two failure modes the plan review called out — a two-of-three wiring and a test that asserts `undefined === undefined` — and neither is present. Both are actually closed, and closed in the specific ways that would have caught them.

### What I verified rather than took on trust

**Wiring is complete at all three call sites.** `QueueList.tsx` renders `QueueItemRow` at :277 (history), :297 (current), :315 (future) and all three receive `showAddedBy` + `viewerUserId` (:290/291, :308/309, :330/331). `grep -rn QueueItemRow packages/mobile/src packages/mobile/app` finds no fourth render site anywhere in the app — every other hit is a `QueueItemRowBoard` type import (`QueueSheet.tsx:11`, `PreSessionView.tsx:17`, `drawer-host-provider.tsx:28`, `app/play.tsx:9`). So `QueueList` really is the whole surface.

**The `QueueList` test discriminates.** `queue-list-added-by.test.tsx` renders the real `QueueList` (with `buildQueueListModel` from `@boardsesh/play-view` unmocked, so the row model is real) and loops `expect(props.showAddedBy).toBe(true)` over all three captured rows — dropping the props from any one call site makes `undefined !== true` and the file goes red. The `props[0].isHistoryItem` / `props[1].isCurrentClimb` / `props[2].isDraggable` assertions are the right shape too: they anchor *which* three rows the loop ran over, so a row-order change fails loudly instead of making the loop vacuous.

**The `AccessibilityCapture` hole is closed, and the mock is honest.** `queue-item-row-memo.test.tsx` adds `accessibilityLabel` to both the type (:53) and the reanimated passthrough's captured field set (:116) — without the second edit the assertion would have been the silent `undefined` half-pass. The `react-i18next` mock deliberately resolves `mobile.queue.addedByAria`'s `{{name}}` (:206–211) rather than returning the raw key, so `toContain('Added by Mina')` also proves the name argument reached the catalog call. Good instinct.

**The gate is the intended one, and it holds up better than the PR body claims.** `showAddedBy = sessionId != null` (`QueueList.tsx:105`) off the narrow `useQueueSessionId` context, not `useIsSharedSession` — which I confirmed reduces to `shouldDefaultToBrowse` → `sessionActive && distinctUserCount > 1` (`packages/shared/play-view/src/browse-defaults.ts:42`), i.e. exactly the presence-flap wipe the comment describes. Two things make the chosen gate safer than the write-up admits: `setSessionId(null)` is only ever reached from `leaveSession` (`use-session-commands.ts:246`) — no transport-disconnect path nulls it, so faces cannot flap on a dropped socket — and that same function immediately dispatches `INITIAL_QUEUE_DATA` with an empty queue, so "faces vanish off a queue still full of the crew's climbs" isn't reachable on leave either.

**The comparand is the right id.** `viewerUserId` is `usePartyProfile().profile?.id`, and `addedByUser.id` is stamped from `partyProfile.id` in `queue-provider.tsx:274`. Same identity space. `PartyProfileProvider` sits at `app/_layout.tsx:501`, above everything the play drawer mounts under, so the `usePartyProfile` throw at `party-profile-provider.tsx:165` isn't reachable from here.

**Empty/absent/self all render nothing, and the whitespace case is real not theoretical.** `resolveQueueRowAttribution` bails on `!addedByUser` (covers both `undefined` from `queue-conversion.ts:95`'s `?? undefined` and a server `null`), on self, and on `username.trim() === ''`. Worth noting why that last one matters: `Avatar.tsx:59` is `name ? getInitials(name) : '?'`, and a whitespace name is truthy — so without the trim you'd get `getInitials('   ')` on a `brandColors.primary` disc, which is the exact broken placeholder the ticket warns about. `avatarUrl ?? null` handles all three of `string | null | undefined`.

**No nested pressable in the gesture arena.** No `userId` is passed, so `PressableAvatar` takes the `if (!userId)` branch at `PressableAvatar.tsx:49` and returns a bare `Avatar` — no `PressableSurface`, nothing for the row tap / swipe Pan / drag handle / tick button to arbitrate against. `pointerEvents="none"` on the wrapper belts it.

**Perf checklist passes.** Row still `memo(QueueItemRowComponent)`; `renderRow` deps gain only the two primitives, with no array `.length` and no inline closures; `keyExtractor` hoisted into a `useCallback` with `[]`; the lookup is `resolveQueueRowAttribution(item.addedByUser, …)` off the row's own item — no roster read, no `filter`, no Map. `usePartyProfile`'s context value is `useMemo`'d on primitives plus the stable `profile` object, so a `useProfile` refetch doesn't churn `renderRow`.

**OTA is safe.** Ten files: seven under `packages/mobile/src`, four locale JSONs under `packages/shared/i18n/locales`. That last path looks like the `locales/**` fingerprint input CLAUDE.md warns about but isn't — the InfoPlist sources are `packages/mobile/locales/**`, untouched here. Empirically confirmed by the preview bot publishing `1b65df0` to `pr-4727` for both platforms as a JS-only change. No `any` in any new file; keys added to all four locale dirs (`de`, `en-US`, `es`, `fr` — there is no fifth).

### Nits

1. **Edit mode strips the name out of the accessibility label, not just the glyph** — `QueueItemRow.tsx:421–432`. `addedBy` is resolved once with `showAddedBy && !isEditMode`, and the label is derived from that same `addedBy`, so entering edit mode also drops `Added by {name}` from `rowAccessibilityLabel`. The visual suppression has a reason (width — checkbox slot plus delete affordance); the label has no width budget. Net effect: a VoiceOver user bulk-selecting rows to delete loses the attribution at exactly the moment they're deciding whose climb to remove. Cheap fix if you want it: resolve the label off `showAddedBy` alone and gate only the `{addedBy && …}` JSX on `!isEditMode`.

2. **Self-exclusion is off while the party profile is still loading** — `QueueList.tsx:106`. `profile` starts `null` and is filled asynchronously by `ensureProfile` (`party-profile-provider.tsx:52–69`), while `sessionId` hydrates on its own async path from storage (`use-queue-persistence.ts:104/126`). Whichever lands first is a race, so on a cold launch into a restored session there's a window with `sessionId != null` and `viewerUserId === null` — the viewer's own face flashes on their own rows for a frame or two, then vanishes. `usePartyProfile()` already returns `isLoading`, so `sessionId != null && !isLoading` closes it. I'm flagging it because the test at `queue-list-added-by.test.tsx` ("still shows peers faces to a viewer with no party profile") pins the current behaviour, and I can't tell from the diff whether the own-face flash was the accepted cost of that or just not noticed. Signed-out devices are immune (`selfAttributionRef` needs `partyUsername`, so they never stamp at all).

3. **`BoardDriverAvatar` with `status="none"` is all wrapper and no badge** — `QueueItemRow.tsx:495`. `showBadge` is false, so the component reduces to a `position: relative` View around `PressableAvatar`, which itself calls `useRouter()` and `useTranslation('common')` before falling through to `Avatar` on the `!userId` branch. That's one extra host node plus two hook subscriptions per visible row for machinery that's switched off. `Avatar` directly is the same pixels. The ticket did say "compose `BoardDriverAvatar`", so this is defensible as written — efficiency nit only, and the cost is small at ~10 visible rows.

4. **The one thing no test here can see: width.** The row gains 20dp plus a `columnGap: spacing[3]` (12dp). `ClimbListItemContent`'s text column is `flex: 1` (`ClimbListItemContent.tsx:372`) with the thumbnail and grade both `flexShrink: 0`, so nothing overflows or clips — the climb name just truncates ~32dp earlier. Worth one screenshot of a session queue on the narrowest supported width with a long climb name and a drag handle showing, since that's the state that eats the most trailing chrome (28 position + 96 thumbnail + 20 avatar + 36 handle + gaps).

5. Two of the seven new row tests wouldn't fail against the unfixed code — "still skips a re-render with attribution props set" and "re-renders exactly once when a session starts" both pass trivially when the props don't exist, because `memo`'s shallow compare reaches the same answer either way. Harmless; the four rendering assertions carry the discrimination. No action.

### Not findings, recorded so they're not re-litigated

- `accessibilityElementsHidden` + `importantForAccessibility="no-hide-descendants"` on the wrapper is redundant under the parent's `accessible` on iOS, but it's load-bearing on Android where `accessible` doesn't reliably merge descendants. Keep it.
- The issue's "session history rows have the same gap" reads, in context, as the queue sheet's history rows — which this covers. `InSessionView`'s history list is `SessionDetailTick` (who *sent* it), a different model with no `addedByUser`, so it's correctly out of scope.
- End-to-end wire coverage holds for a joiner, not just for live deltas: `SUBSCRIPTION_QUEUE_ITEM_FIELDS` (`operations.ts:1271`) selects `addedByUser { id username avatarUrl }` and is spread into `FullSync`, both single-item variants, *and* the initial query at :1366–1367. So faces are there on the first paint after a join, not one delta later.
- No new privacy surface: username and avatarUrl already crossed the wire before this PR; it only renders what was already arriving.

Nothing here blocks. Ship it — with a glance at nit 1 (a two-line change, and the only one with a user-visible consequence) and a device shot for nit 4.

marcodejongh:
Thanks — the two nits with a user-visible consequence are fixed in `f0f718985`. Naming the rest as accepted-but-deferred.

**Fixed**

1. **Edit mode stripped attribution from the accessibility label, not just the glyph** (nit 1). You were right that this is the only one that costs a user something. The 20dp avatar does have to go in edit mode — widest state, checkbox slot plus delete affordance — but the label has no width budget, and a VoiceOver user bulk-selecting rows to delete is precisely who needs to know whose climb each row is. The attribution is now resolved independently of `isEditMode` and only the avatar is gated on it. `queue-item-row-memo.test.tsx` now asserts both halves: glyph gone, `Added by Mina` still in the label.
2. **Own face could flash on a cold launch** (nit 2) — this was unnoticed, not accepted, so thank you for flagging the ambiguity. `showAddedBy` now also waits on `usePartyProfile().isLoading`, so no face renders until `viewerUserId` can actually suppress the viewer's own. `queue-list-added-by.test.tsx` gained a case that pins both directions: nothing while loading, everything once it resolves, without a remount. The mocked provider now models `isLoading` too — it previously didn't, which is why no existing test caught this.

**Deferred, with reasons**

- **`BoardDriverAvatar` with `status="none"` carries switched-off badge machinery** (nit 3) — real, but the ticket explicitly asked to compose `BoardDriverAvatar`, and dropping to `Avatar` is the same pixels for a follow-up to decide. Efficiency only.
- **Narrow-width truncation with a long climb name + drag handle** (nit 4) — genuinely untestable in CI, agreed. Wants a device screenshot, which is part of the QA pass rather than something to guess at here.
- **Two of the seven row tests pass trivially against unfixed code** (nit 5) — harmless memo shallow-compare artefacts; not churning the suite for them.

**Also noted:** your two corrections to my PR body are right and better than what I wrote — `setSessionId(null)` is only reachable from `leaveSession` (no transport-disconnect path nulls it) and that same call empties the queue, so the "faces vanish off a full queue" scenario I was defending against isn't actually reachable. The `sessionId != null` gate is still the right choice, but for a simpler reason than I claimed.

**Validation on `f0f718985`:** `typecheck:mobile` clean · `test:mobile` 703 files / 6,955 tests pass · `vp check` clean via the pre-commit hook.

Issue #4721, opened 2026-08-24
Shard the layout snapshot artifact so a failed transfer costs one shard, not 99 MB


marcodejongh:
26-31% of Kilter layout-artifact transfers fail, and every one of them throws away everything it had downloaded — up to ~99 MB. The artifact GET is unresumable and provably cannot be made resumable client-side, so the only lever left is making a failed transfer cost less.

## The evidence

Per-transfer telemetry (`Offline Artifact Transfer`, added in #4394) on the Kilter layout artifact:

- 16 of 51 iOS transfers failed (31%), 10 of 38 Android (26%).
- 21 of those 26 carried `backgroundedDuringTransfer: true`.
- Every failure discards all bytes: `snapshot-source.ts` deletes any file without a valid `.complete` sidecar before re-downloading.
- Successful transfers run at 33,786 kbit/s p50 (iOS) / 42,818 (Android), wall p50 13.2s / 17.0s, `downloadMs` p90 29.5s (iOS) / 125.1s (Android).

## Why not resume instead

Settled, with the measurement in `docs/board-snapshots.md`. Artifacts are served `Content-Encoding: gzip` unconditionally — it is stored-object metadata, not negotiated, so even `Accept-Encoding: identity` gets gzip — and `Range` therefore addresses ENCODED bytes while both platforms write DECODED bytes to disk. expo-file-system's Android resume computes its offset from the destination file's length and sends `Range: bytes=<decoded>-`: off by 2.6x on `kilter:1`, with no JS-side fix. Gunzipping in JS needs a native dependency, which is OTA-forbidden. `Offline Artifact Transfer.resumed` should stay `false`.

## What sharding is

Split the per-layout artifact into N smaller objects so a failed transfer costs one shard instead of the whole file, and a backgrounded app resumes at a shard boundary rather than at zero.

## Why it is not one PR

- backend export change (`export-board-snapshots.ts`)
- manifest format-version bump, with both versions live through the 14-day prune grace plus CDN-cached manifests
- multi-file client import, including the partial-shard-set case
- retention and the superseded-partial sweep, per shard
- a rollback path

And its failure mode is the only one in this workstream that is permanent, silent catalog loss: a client that imports a subset of shards and stamps a checkpoint over the gap never revisits it, because the delta pull is a strict `>` keyset.

## Trigger condition

Do this when either number stops improving on its own:

- `Offline Artifact Transfer` failure rate for the Kilter layout artifact stays above ~20%, or
[...]

Issue #4719, opened 2026-08-24
Kilter grades still crawl for 17s p50 on the 74% of download cycles the grades artifact never reaches


marcodejongh:
The Boardsesh-grades snapshot artifact (#4346/#4347) was supposed to end the paged `board_climb_grades` crawl. On the cycles that import it, it did. On most Kilter cycles it never arrives, and those still crawl grades for tens of seconds while the climbs and stats crawls finish in well under a second.

## The measurement

PostHog project 412845, `Offline Board Download Completed`, `scopeKey` starting `kilter:`, since 2026-08-15 (the first nightly grades export after #4346), split on whether the cycle carried an import (`importMs IS NOT NULL`):

| | cycles | `gradesPullMs` p50 | `gradesPullMs` p90 | `climbsPullMs` p50 | `statsPullMs` p50 |
| --- | ---: | ---: | ---: | ---: | ---: |
| import-carrying | 42 | 261 ms | — | — | — |
| import-free | 119 | **16,940 ms** | **174,068 ms** | 347 ms | 633 ms |

The import-free row is the interesting one. Those same cycles crawl climbs in 347 ms and stats in 633 ms, so it is not a slow link or a slow device — it is real grade-page work, hundreds of serial authenticated GraphQL round trips.

74% of Kilter completions (119 of 161) are in that population.

## Why this was not visible before

`ScopeDownloadPhaseBreakdown` is scoped to the cycle that completed the download (`pull-client.ts:150-160`), and the phase-carrying subset is explicitly flagged in the same file as "biased toward the healthy population" (`pull-client.ts:275-279`). Reading `gradesPullMs` over the phase-carrying events alone says the grades term is dead. It is dead for one cycle in four.

## What to look at

`importGradesForScope` (`packages/shared/offline-sync/src/sync/pull-client.ts`) runs on two paths: straight after a successful whole-layout import, and as a retrofit whenever the grades checkpoint is absent on a scope whose climb catalog is already complete (`bootstrap-done` or `scope-complete`). Candidate reasons a cycle reaches neither:

- The retrofit gate wants a COMPLETE catalog. A scope that is still finishing its delta pull never qualifies, and that is exactly the multi-cycle population above.
- `MAX_GRADES_BOOTSTRAP_ATTEMPTS` is 2 and is charged on a null return as well as a throw. Two pocketed-phone cycles spend it for the life of the install.
[...]

Issue #4393, opened 2026-08-12
Offline Board Download Completed reports gradesRows: 0 when the import happened in an earlier cycle


marcodejongh:
When a scope's Completed fires in a later cycle than its snapshot/grades import (the designed per-scope-lifetime semantics — downloadMs/importMs are correctly absent then), `gradesRows` reports `0` instead of being absent. Verified on-device: grades were fully imported and render in the play view while the event said 0. Cosmetic but it corrupts the #4310 telemetry read — 0 is indistinguishable from a genuinely empty grades import. Follow the absent-when-unknown convention for gradesRows too.

Fleet reference: 4× gradesRows:0 completions on 2026-08-12 evening are this artifact, not missing grades.

Refs #4310 #4316.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01UT6VP1sWqk6bY9mUgyeS2U

marcodejongh:
Shipped in #4399 (merged): gradesRows is now absent-when-unknown (present only when this cycle's crawl started from a cursor no earlier cycle advanced), the new gradesArtifactRows property carries the snapshot-artifact row count, and the engine's bootstrapHealed verdict is now actually emitted — all three feed the #4310 telemetry read.

PR #4728, opened 2026-08-24
perf(sync): batch the snapshot import into short exclusive transactions

### LINKED ISSUE #4310: Kilter offline download is slow on the snapshot fast path (p50 2m55s, Android p50 8m52s)

marcodejongh:
Part of #4310.

The whole-layout snapshot import held **one** `BEGIN EXCLUSIVE` around ~710,000 rows. This splits it into an autocommit preamble, a reconcile transaction, `ceil(rows / 5000)` row transactions, and a final transaction holding both checkpoints plus the deletions rewind — releasing the write lock between every one of them.

## What actually changed

**The import.** `importScope` becomes `importScopeBatched`. The column intersection, the staging table and every keyset probe run in autocommit holding nothing; only the `INSERT OR REPLACE`s take the lock. `board_climbs` is batched on a TEMP staging table's rowid; `board_climb_stats` is batched on its PRIMARY KEY.

**The invariant that replaces all-or-nothing.** A crash or purge mid-import can now leave rows committed. It can never leave a checkpoint: both `setCheckpoint` calls and `rewindDeletionsCheckpoint` live only in the final transaction. That asymmetry is the one `scope-teardown.ts`'s header is written around — rows without markers is benign (the re-import is idempotent, and a teardown still removes them), markers without rows is unrecoverable, because the strict `>` delta pull never revisits anything at or below a stamped cursor. Two tests assert "no checkpoint after an interrupted import", one throw-injected and one purge-injected.

**The purge guard moved inside the lock.** `isSigningOut() || hasPurgeLanded(...)` is now re-read after `BEGIN EXCLUSIVE`, not before it. `beginScopePurge` latches before its delete transaction takes the lock, so the only two orderings are: a batch wins the lock and commits (and the purge's own DELETE removes what it wrote), or the purge wins and the batch reads the latch and bails.

Three things the batching itself needed:

- **`synchronous = NORMAL` on the import connection only**, applied in autocommit. Without it, ~142 commits each pay an fsync and the batched import is slower than the single transaction it replaces. SQLite *rejects* the pragma inside a transaction rather than ignoring it, which is what makes "call it in the preamble" enforceable; there's a test for that. In WAL it cannot corrupt — it can only lose transactions committed just before a power cut, and the WAL replays to its last valid frame, so a later commit can never survive an earlier one being lost. That is exactly what makes "checkpoints stamped last" safe.
- **SQLite row-value keyset syntax for the stats batches.** Measured with `EXPLAIN QUERY PLAN` on SQLite 3.53.3 against `board_climb_stats`' real PK `(board_type, climb_uuid, angle)`:
  - `a > ? OR (a = ? AND b > ?)` → `SEARCH s USING INDEX ... (board_type=?)` — a full partition rescan on **every** batch, ~142 passes over 306k rows.
  - `(a, b) > (?, ?)` → `SEARCH s USING INDEX ... (board_type=? AND (climb_uuid,angle)>(?,?))`.

  Written the first way, the "batched" import would be O(n²) and *slower* than what it replaces, with an identical row set — so a row-set test cannot catch it. A test EXPLAINs the statement the import actually ran and asserts the plan contains `(climb_uuid,angle)>`.
- **A lost lock race gets its own budget.** `classifyBootstrapFailure` returned `structural-artifact` unconditionally for the import stage. One acquisition became ~143, against writers that genuinely exist — every tick and favourite, plus a `removeBoardScopeData` for a *different* layout that this scope's purge guard does not cover and that `scope-teardown.ts` says "runs for seconds on a 40k-climb layout", longer than the import connection's 5s `busy_timeout`. Two lost races would have spent the whole two-strike structural budget and settled a board onto the paged crawl for the life of the install (`remove-offline-board.ts:145-148` already documents that hazard for VACUUM). Now: a lost `BEGIN EXCLUSIVE` retries on a 250/750/2000 ms ladder, a lock-shaped import failure classifies as a new `database-locked` kind with its own 3-strike budget, and the retained artifact is not deleted on that path. Why not the transport budget — which is where the first version of this put it — is the next section.

## The blocker the adversarial review caught, and the fix

The first version of the lock escape classified a lost race as `transport`, on the reasoning that "any successful download clears it, so it can never accumulate". That reasoning is the bug.

`pull-client.ts` runs `clearTransportFailures` after every successful download — including a **retained** one, where `downloadArtifact` short-circuits with `{ reused: true }` having moved **zero bytes**. So on a device with persistent write-lock contention: cycle N charges 1, cycle N+1's zero-byte reuse resets it to 0, charges 1 again, forever. `transportFailures` never reaches `MAX_TRANSPORT_DOWNLOAD_FAILURES`, `isTerminal` is never true, `markBootstrapPagedFallback` never runs — and `shouldSkipPagedPull` returns `true` every cycle for a fresh scope, because the 2-minute cooldown sits well inside the 30-minute grace window. The board becomes unreachable by **both** paths, while the device re-ATTACHes and re-`quick_check`s a 271 MB artifact every couple of minutes indefinitely. Before this PR the same failure spent two structural strikes and settled onto the paged crawl inside ~30 minutes.

The fix is a third budget rather than a conditional clear: `MAX_BOOTSTRAP_LOCK_FAILURES = 3` on `BootstrapRetryState.lockFailures`, on the transport cooldown rungs (2 min / 15 min), which nothing clears. A conditional clear (`skip clearTransportFailures when reused`) fixes the retained-artifact loop but not the one where the OS evicts the file and each cycle re-downloads ~103 MB fresh — that path resets a transport counter legitimately and would loop just as forever, at 103 MB a go.

Why not the structural budget either: it strands the board after **two** strikes and deletes the ~103 MB file, for a fault the artifact did not cause. And `canRearmOnNewArtifact` deliberately does not cover the lock budget — tonight's export cannot win a race the last one lost. The consented escape stays "Try the fast download again", which restores every budget.

The test drives three `pullSync` cycles against a permanently-locked stats batch with `reused: true` every time and asserts the scope goes terminal, `shouldSkipPagedPull` flips to `false`, the paged-fallback marker lands, and the artifact survives. Reverting just the bucket choice makes it fail on `isTerminal` — which is the defect, so it is the first assertion in the case.

## The framing correction — read this before the numbers

**`importMs` is not a lock hold, and never was.** It is stamped around the whole call: ATTACH, `PRAGMA quick_check` over a 271 MB file, `verifySnapshotMeta`'s two full `COUNT(*)` scans, the scoped watermark reads, *then* the write work — and everything before the last item runs in autocommit holding nothing. Its own docblock says so.

Two in-repo comments (`db/pragmas.ts`, `db/write-retry.ts`) sized a retry ladder on the claim that "`importMs` over 60 days is p50 806ms, p90 2.3s, max 3.2s — every observed import fits inside attempt 1's window". That window never existed: `importMs` was first emitted on 2026-08-12 (#4337/#4345) and `write-retry.ts` was written on 2026-08-14, so at most two days of the series were ever readable. Both comments are corrected to say that, rather than substituting today's p50 2,944 / max 253,939 — which would swap one wrong number for another, because it is still the whole import.

**How long the import held the lock has never been measured.** This PR emits `importLockMaxMs` for exactly that. No `OFFLINE_DB_*` constant is touched; re-tuning the ladder is a separate decision that wants that series first.

The same discipline applied one level down, after review: `importReconcileMs` and `importRowsMs` were stamped around `runExclusive`, which `BEGIN`s first — so a contended acquisition (up to `busy_timeout` per attempt plus the ladder's sleeps) was charged to the phase it waited in. A reconcile that executes in 40 ms behind a `removeBoardScopeData` would have reported seconds, and the follow-up trigger below (`importReconcileMs` p90 on `bootstrapHealed = true`) would have fired on contention rather than on reconcile cost. The wait is now accumulated separately as `importLockWaitMs` and subtracted out of both, so those two are work-plus-hold and `importLockMaxMs` stays a hold and nothing else.

One cost does stay inside the holds, and the batch-size rationale now names it: SQLite attempts a WAL autocheckpoint at COMMIT, and nothing in the engine overrides the 1000-page default (only the explicit `wal_checkpoint(TRUNCATE)` calls in `vacuum.ts` and `remove-offline-board.ts` exist). The single pre-change transaction paid that once, at the end; ~142 commits cross the threshold roughly every 4 MB of WAL, so a minority of batches COMMIT with a passive checkpoint inside `heldMs`. So the `importLockMaxMs` tail is set by checkpoint-carrying batches, not by the ~150 ms typical batch `SNAPSHOT_IMPORT_BATCH_ROWS` is sized from — and that same cost is the most likely way `importRowsMs` regresses against the single transaction. `wal_autocheckpoint = 0` for the import would move it out of the holds at the price of a WAL that grows to the size of the whole import on a phone; not a trade worth making before the number says the tail is a problem.

## Telemetry

New on `Offline Board Download Completed`, all **absent-when-unknown**, never zero: `importVerifyMs`, `importReconcileMs`, `importRowsMs`, `importLockMaxMs`, `importLockWaitMs`, `importBatches`, `gradesDownloadMs`, `gradesVerifyMs`, `gradesLockMs`. Plus `lockFailures` on `Offline Snapshot Retry Scheduled`, which is how you tell whether the batched import contends for the lock in the field at all.

They follow `events.ts`'s stated rule ("Read a missing value as UNKNOWN, never as 0", with #4393's `gradesRows: 0` as the precedent) rather than emitting honest zeros. **Two different filters**, and the first version of this PR got the second one wrong in four places:

- The six `import*` props: filter on `importMs IS NOT NULL`. **119 of 161** post-2026-08-15 Kilter completions carry no import at all, so an unfiltered p90 over `importLockMaxMs` would read ~0 whether or not this works.
- The three `grades*` props: filter each on **its own** `IS NOT NULL`, never on `importMs`. `importGradesForScope` has a second call site — the retrofit path for a scope that is already bootstrapped but holds no grades checkpoint — which downloads and imports grades in a cycle where no whole-layout import ran. That is exactly the still-crawling population #4719 is about, so the `importMs` filter would have excluded the cycles `gradesLockMs` was added to measure.

`gradesDownloadMs` / `gradesVerifyMs` / `gradesLockMs` also close a real blind spot: `importGradesForScope` ran its own transfer and its own exclusive transaction, and neither touched `phases.downloadMs` or `phases.importMs` — most of the ~11s p50 gap between `durationMs` and the sum of the phases. The grades import is still **one unbatched** exclusive transaction; `gradesLockMs` is what decides whether it needs the same treatment.

## What this PR does not claim

- **Not that lost sends are fixed.** `Offline Local Write Attempt Failed` over 45 days is 31 events across 26 people, max `elapsedMs` 342ms, most recent 2026-08-22 — no observed user write has ever waited out its 2,500ms foreground window. The claim here is a bounded lock hold and a determinate progress bar, not a recovered send.
- **Not a magnitude for the new hold.** The release note deliberately does not put a number on it. `importLockMaxMs` is unmeasured until this reaches the fleet, the worst import throughput ever observed puts one 5,000-row batch at ~1.8s, and `reconcileScope` is still unbatched inside the same `Math.max`.
- **Not that the grades term is dead.** Split on import presence, post-2026-08-15 Kilter completions give `gradesPullMs` p50 261ms for the 42 import-carrying cycles but **p50 16,940ms / p90 174,068ms** for the 119 import-free ones — whose `climbsPullMs` p50 is 347ms and `statsPullMs` p50 633ms in the same cycle, so it is real crawl work, not a slow link. Filed as #4719.
- **Not that the whole import got faster.** The lock hold gets bounded. `importRowsMs` before/after is what says whether the total moved, and it is measured either way.

## The gate, scoped honestly

The number to judge this on is `importLockMaxMs` p90 for `scopeKey LIKE 'kilter:%'`, and the target is ≤ 2,000ms — **asserted for fresh scopes**. `reconcileScope` is still one unbatched exclusive transaction (two DELETEs with nested correlated `EXISTS`/`NOT EXISTS` and a per-row `json_each`), and it is inside the `Math.max`. Cheap on a fresh scope, not bounded on a heal-over-partial (#4313) or a second size of an already-downloaded layout. That population is a named follow-up rather than a claim: read `importReconcileMs` p90 filtered on `bootstrapHealed = true` — now with the lock wait subtracted out, so the number is reconcile cost — and batch it off the same staging table if it says so.

## Test plan

On the `pr-4728` OTA preview, with a Kilter board — the biggest import:

1. Remove your Kilter board, then re-download it → the import bar moves.
2. Log a send while the import runs → it saves.
3. Force-quit mid-import, reopen → the download resumes and finishes.
4. Search that board with Wi-Fi off → every climb and grade shows.
5. Check `Offline Board Download Completed` → `importLockMaxMs` present, non-zero.

## Risk

Risk: 4/5 — the offline import's write path, shipped over OTA. Rows can now commit without their checkpoint, which is the safe half of the asymmetry (the next import redoes them). The unsafe half — a checkpoint without its rows — loses catalog silently and forever, so checkpoints-last carries two interruption tests and a purge test.

## Validation

`vp check`; `vp run typecheck:mobile`; `vp run typecheck:offline-sync`; `vp run test:offline-sync` (761 tests); `vp run test:mobile`; `vp run check:mobile-bundle`.

## Native fingerprint

**Not moved.** Every changed file is TypeScript under `packages/shared/**` and `packages/mobile/src/**`, plus `docs/`. No `app.config.ts`, plugin, native module, patch or dependency change — this ships over OTA to existing binaries.

## Follow-ups filed

- #4719 — Kilter grades still crawl for 17s p50 on the 74% of download cycles the grades artifact never reaches.
- #4721 — Shard the layout snapshot artifact so a failed transfer costs one shard, not 99 MB (with its trigger condition named).

## Why not `Closes #4310`

#4310's title and numbers describe a state the fleet has already left (#4345/#4346/#4347 moved Kilter Android p50 from 8m52s to 2m34s), and this PR's own result is not readable until the OTA reaches the fleet and `importLockMaxMs` has coverage. What would justify closing it: `importLockMaxMs` p90 at or under 2,000ms on both platforms on fresh Kilter scopes, with `Offline Board Download Failed` for `database-locked` / `artifact-invalid` / `unknown` and `Offline Local Write Attempt Failed` volume both flat. Whether #4310 is then re-scoped onto #4721 or closed with the numbers is Marco's call.

## Release Notes

Downloading a big board no longer takes the app's database hostage. The import now works in short bursts and hands the database back after every one, so logging a send or favouriting a climb while a Kilter or Tension board comes down waits on one burst instead of the whole import.

The download bar moves during the import too, instead of sitting on a spinner with nothing to show for the last stretch.


marcodejongh:
## Review round 2 — what came back and what changed

Two reviews landed on this: an adversarial pass over the whole change, and `claude-review`. Both found something real.

### The blocker: a lost lock race had no bound at all

The adversarial review's one major finding, and it reproduces. Amendment 4a's lock escape put a lost `BEGIN EXCLUSIVE` on the **transport** budget, reasoning that "any successful download clears it, so contention can never accumulate". That reasoning is the bug.

`pull-client.ts` runs `clearTransportFailures` after every successful download, and `snapshot-source.ts` hands a **retained** artifact straight back with `{ reused: true }` having moved **zero bytes**. So with persistent write-lock contention: cycle N charges 1, cycle N+1's zero-byte reuse resets it to 0, charges 1, forever. `transportFailures` never reaches `MAX_TRANSPORT_DOWNLOAD_FAILURES`, `isTerminal` is never true, `markBootstrapPagedFallback` never runs, and `shouldSkipPagedPull` returns `true` every cycle for a fresh scope because the 2-minute cooldown sits inside the 30-minute grace window. The board is unreachable by **both** paths while the device re-ATTACHes and re-`quick_check`s a 271 MB artifact every couple of minutes — strictly worse than the two structural strikes the escape replaced, which settled onto the paged crawl in ~30 minutes. The comment at `snapshot-source.ts` claiming "trusting the sidecar here cannot strand a scope" was false for that path.

Fixed with a third budget rather than a conditional clear: `database-locked` / `MAX_BOOTSTRAP_LOCK_FAILURES = 3` on `BootstrapRetryState.lockFailures`, on the transport cooldown rungs, cleared by nothing but the climber's explicit retry. I took the reviewer's option (b) over option (a) deliberately — "skip `clearTransportFailures` when `reused`" fixes the retained-artifact loop but not the one where the OS evicts the file and each cycle re-downloads ~103 MB fresh, which resets a transport counter legitimately and would loop just as forever at 103 MB a go. `canRearmOnNewArtifact` deliberately does not cover the lock budget: tonight's export cannot win a race the last one lost.

Test: three `pullSync` cycles against a permanently-locked stats batch with `reused: true` every time, asserting the scope goes terminal, `shouldSkipPagedPull` flips to `false`, the paged-fallback marker lands, and the artifact survives. Reverting only the bucket choice makes it fail on `isTerminal` — verified — so that is the case's first assertion. Plus four pure cases in `bootstrap-retry.test.ts`, one of which walks the exact clear-then-charge order the caller uses.

### From `claude-review`: a real error-recovery hole this PR introduced

`bootstrapScopeGradesFromSnapshot` now COMMITs its own transaction (so `gradesLockMs` covers the hold) and hands the wrapper a trailing empty `BEGIN`. That shape is new in this PR, and it opened a window the old one did not have: a throw from that `BEGIN` meets expo's unconditional ROLLBACK with no transaction active, and "cannot rollback - no transaction is active" masks the error the caller dispatches on. The whole-layout import already carries the restore-BEGIN wrapper for exactly this; grades had none. Fixed, with a test that fails the post-COMMIT `BEGIN` once — verified to produce the masked `cannot rollback` message without the fix.

I did not chase its "measure the hold to the ROLLBACK too" suggestion: `importLockMaxMs` sizes the hold a concurrent write survives on the path that COMMITs, and a batch that threw is a failed import the caller settles. Documented in `runExclusive`'s `finally` rather than changed. Same for the missing `applyBulkImportPragmas` on the grades transaction — one transaction means one fsync, so a single durable commit is the cheaper end of the trade; now written down instead of left implicit.

### The four minor findings

- **The `importMs IS NOT NULL` filter was wrong for the three `grades*` props**, in all four places it appears. `importGradesForScope` has a second call site — the retrofit path for a scope that is already bootstrapped but holds no grades checkpoint — which imports grades in a cycle with no whole-layout import. That is exactly the #4719 population, so the filter excluded the cycles `gradesLockMs` was added to measure. Each of the three is now documented as self-filtering.
- **`importReconcileMs`/`importRowsMs` included the lock-acquisition wait**, repeating one level down the elapsed-vs-hold conflation this PR exists to undo. A reconcile that executes in 40 ms behind a `removeBoardScopeData` would have reported seconds, and the PR's own follow-up trigger (`importReconcileMs` p90 on `bootstrapHealed = true`) would have fired on contention rather than cost. The wait is now accumulated as `importLockWaitMs` and subtracted out of both. Tested by injecting a 40 ms wait on the reconcile's acquisition and asserting `importReconcileMs < 40`.
- **The batch-size rationale omitted the WAL autocheckpoint** that batching moves inside the measured hold. Named in the sizing docblock, the telemetry note and the docs rather than overridden: `wal_autocheckpoint = 0` for the import would move the cost out of the holds at the price of a WAL that grows to the size of the whole import on a phone.
- **The docblock above the `bootstrapScopeFromSnapshot` call still asserted the all-or-nothing invariant** this PR removes. Rewritten to the new one.

### The release note

Reworded. "Waits on a hold measured in fractions of a second" was a quantitative claim against a number this PR says has never been measured, and it would have read as a promise if `importLockMaxMs` lands above 1 s on Android. It now claims a bounded hold without a magnitude and leaves the number to the series.

Nothing declined. `vp check`, `vp run typecheck:offline-sync`, `vp run typecheck:mobile`, `vp run test:offline-sync` (760), `vp run test:mobile` (6,932) all green locally; `ota-check` still reports no native change.

marcodejongh:
## Review round 3 — the two reviews that landed after the last push

Rebased onto `main` (50 commits, no conflicts) and re-validated. Two `claude-review` passes arrived after the round-2 push, so nothing in them had been answered yet. Four points applied, two declined with reasons.

### Applied

**No test asserted that the import calls `applyBulkImportPragmas`.** Correct, and worth more than "low priority" — I checked by deleting the call: 17 of the 18 cases in `snapshot-import-batching.test.ts` still passed. The pragma is the only reason ~142 commits are cheaper than one, so dropping it turns this PR from a fix into a regression that no assertion in the repo would have caught — it would have surfaced weeks later as an `importLockMaxMs` shift in the field. Added a case that asserts the pragma is issued *and* that it is issued before the first `BEGIN EXCLUSIVE`, since autocommit is the only place SQLite accepts it. With the call removed it fails; with it restored, 18/18.

**`applyBulkImportPragmas` and pooled connections.** The docblock claimed "per connection" without naming the mechanism, which is why the question came up. It now says it: the only caller applies it inside `withExclusiveTransactionAsync`, which expo-sqlite runs on a connection it opens (`useNewConnection: true`) and tears down when the task returns. There is no pooled slot left loosened — the loosened connection stops existing when the import ends. That was already written down one file over, in `bootstrapScopeFromSnapshot`'s CONNECTION INVARIANT block; it just was not written where the question gets asked.

**The `EXPLAIN QUERY PLAN` assertion was whitespace-brittle.** Now a regex (`/\(\s*climb_uuid\s*,\s*angle\s*\)\s*>/`) rather than a literal `toContain`. Same guarantee, one less way to fail on a rendering change.

**The 134-character comment line** in `bootstrap-retry.ts` is rewrapped, plus one more of the same in `events.ts` I found while I was there.

### Declined, with reasons

**Gating the lock escape on `stage === 'import'`.** The suggested one-word change makes the failure worse, not tighter. `classifyBootstrapFailure`'s fallthrough for an ungated lock error is `structural-device` — the harshest bucket there is: two strikes, never re-armed by a new `builtAt`, and it is what a lock error would land in with the stage check added. Charging transient contention there is strictly worse than the lock budget, which costs zero downloads and retries three times. The bucket is named by *cause*, not by stage, so if a future change ever does put a write on the manifest or download stage, landing in `database-locked` is the outcome to want.

The review was right that the docstring and the code disagreed, though — the docstring said "at the import stage" while the check is unconditional. Fixed on the docstring side, with the stage-independence now stated as a deliberate choice and the `structural-device` fallthrough spelled out.

**Replacing the plan-string assertion with a row-set equivalence check.** That is the one substitution this assertion exists to refuse. Both forms return byte-identical rows — that is the whole hazard, and there are already two row-set equivalence cases in the file that pass under either. Re-measured independently while checking this, on SQLite 3.51.2 (the `node:sqlite` build CI actually runs, not the 3.53.3 the comment cites):

```
ROW-VAL : SEARCH s USING COVERING INDEX ... (board_type=? AND (climb_uuid,angle)>(?,?))
OR-FORM : SEARCH s USING COVERING INDEX ... (board_type=?)
```

Same split across a two-minor-version gap, which is also the answer to the version-drift worry: if SQLite ever stops emitting that shape, a loud CI failure on a Node bump is the correct outcome, not a silent return to a full partition rescan on every one of ~142 batches.

### Verification of the round-1 amendments

I re-checked all seven against the diff rather than trusting the earlier report, including running `EXPLAIN QUERY PLAN` myself (above). Row-value keyset with `board_type = ?` as the leading equality: present. Nine new timings optional and spread behind the `=== undefined` guard, absent from `emptyScopeDownloadPhases()`: present. Lock escape, in-loop `BEGIN EXCLUSIVE` ladder, and no `deleteArtifact` on a lock-classified failure: present. Per-batch emission in its own swallowing try/catch: present. Both `setCheckpoint`s and `rewindDeletionsCheckpoint` in the final transaction only: present. No `singleTransaction` production flag (the pre-change shape is a test-file helper). The concurrent-write case races from an independent loop, not from inside the awaited hook.

`vp check`, `vp run typecheck:offline-sync`, `vp run typecheck:mobile`, `vp test run --project offline-sync` all green locally after the rebase.

marcodejongh:
## Ready for review — 30/30 green

Rebased onto `main` (50 commits, clean), pushed as `e754aed`, and out of draft.

**One red on the way, and it was not this branch.** `test-default (1, 2)` timed out at 5,000 ms on `snapshot-bootstrap.test.ts > spends only the structural budget for a device-side fault`. Checked before re-running rather than after: that case is untouched by this PR, and its download fails with a disk-full error, so it never reaches the import — none of the batching, the pragma or the lock ladder executes in it. It runs in **46 ms** locally and passed 3/3 on a repeat loop; the shard it sits in ran 176 files and 2,225 tests in 102 s on a loaded runner. Re-ran the job once, per the flake rule: green, and so is everything else.

Two new gates arrived from `main` while this sat: `PR Test Plan / check` (added after the branch's base commit) wanted a `## Test plan` and a `## Risk` line, and `board-art-geometry` / `board-render-version` came in with #4833. All satisfied — the test plan runs against the `pr-4728` OTA channel, and the risk is 4/5 with the reason being the checkpoints-last asymmetry rather than the batching itself.

**Local validation after the rebase:** `vp check`, `vp run typecheck:offline-sync`, `vp run typecheck:mobile`, `vp test run --project offline-sync` (761, up one from the new pragma case). `ota-check` still reports no native change, so this ships over OTA.

Nothing left open. The two review points I declined are argued in the round-3 comment above; both are one-line answers if anyone disagrees. Not merging — over to you.

PR #4731, opened 2026-08-24
feat(mobile): add a playlist to the queue without clearing it

### LINKED ISSUE #4673: Playlists: "Add to queue" that appends instead of replacing

marcodejongh:
## Summary

> ### ⚠️ Stacked on #4712 — base is `feat/4671-queue-play-next`, not `main`
>
> This PR consumes the parametric `showQueueAddedSnackbar({ kind, count })` seam
> that #4712 owns, so it is based on that branch rather than `main`.
>
> **The diff shown against `main` will look larger than this PR's own change**
> until #4712 merges — it carries #4712's commits too. Review this against
> `feat/4671-queue-play-next`.
>
> **Once #4712 merges:** rebase this onto `main` and retarget the base back to
> `main`. Do not merge it into `main` before then.

Closes #4673.

Every bulk path from a playlist into the queue was destructive. A crew wanting to
merge two people's playlists into one session had to pick between losing what was
already queued and giving up.

Two additive landings, no new chrome:

**1. An "Add to queue" row in the playlist-detail list header** (above
Discussion). Appends every board-scoped climb behind the live queue. Nothing is
cleared, the current climb never moves, the wall never repaints.

It is a labelled row rather than a second FAB or a menu item because that is the
only surface that reaches all four populations at once: both platform branches
(the existing bulk control is a Material-only `Appbar.Action`), owners *and*
non-owners, and **smart playlists** — Liked Climbs / Five Stars / Projects render
no overflow menu at all, so a menu item could never have reached exactly the
playlists a crew bulk-queues. Direct precedent 60 dp below it: `PlaylistDiscussionRow`
is the same shape.

**2. The destructive confirmation becomes a three-way fork.** "Start playlist?"
is where the climber is *told* they are about to lose their queue, and its only
non-destructive exit was "Keep queue" — a dead end. It now offers
**Start playlist** (destructive, still first: they tapped ▶ and we don't hijack
that tap) · **Add to queue** · **Keep queue**.

`PlaylistQueueReplaceSheet.tsx` (94 lines) is **deleted**. The fork is
`useChoose()` — the same imperative prompt the cross-board queue-add gate already
uses for a queue-add decision one screen over. That buys the 44 dp tap-target
floor (the sheet's buttons were 40 dp, on a destructive confirm), the wrapped
Material action row for `Zur Warteschlange hinzufügen`, and a promise-shaped API
so the confirmation fires after the prompt is gone by construction — all of which
already exist inside `dialog-provider`.

Also: `detail.activateAll` renamed from "Queue every climb" to **"Start playlist"**.
It is the only name a screen-reader user gets for the ▶ control, and it described
the wrong one of the two paths.

### Provider seam

`appendGeneratedSession(items)` → `appendQueueItems(items, { activateFirstWhenIdle? })`:

- **The pointer move is now opt-in and defaults off.** The workout generator opts
  in (`{ activateFirstWhenIdle: true }`) and behaves exactly as before. The
  playlist row leaves it off — "Add to queue" already means "don't take the wall"
  everywhere else in the app (the single-add path's `DELTA_ADD_QUEUE_ITEM` never
  sets `currentClimbQueueItem` either, not even into an empty queue).
- Clamped to a new shared `MAX_SYNCED_QUEUE_ITEMS` and returns what landed.

### How the null current pointer is prevented

Making `setQueue([...queue, ...items], null)` reachable for the first time would
mean the resolver **clears the session's current climb** — `setQueue` in
`packages/backend/src/graphql/resolvers/queue/mutations.ts` defaults an absent
`currentClimbQueueItem` to `null` and hands that straight to
`roomManager.updateQueueState`, which fans it out to every peer. An *additive*
action must never be able to blank a crew's wall.

`appendQueueItems` never constructs that payload:

- With a current climb, it re-asserts **that** item. The pointer is explicitly
  unchanged, never inferred.
- With **nothing** current and `activateFirstWhenIdle` off, it does not send a
  wholesale replace at all — it dispatches the batch locally and fans it out as
  per-item `addQueueItem` mutations, which carry no pointer. The invariant then
  holds *by construction* rather than by trusting that the local mirror of shared
  session state is fresh. `clearQueue` already fans out one `removeQueueItem` per
  item for the same reason, so this is the established shape, not a new one.
- `activateFirstWhenIdle` can only ever **fill** an empty pointer; it can never
  move one that is set.

Covered by `appendQueueItems leaves the current pointer null by default when
nothing is current` (asserts `setQueue` is *not* called) and
`appendQueueItems broadcasts per-item adds (never a pointer-clearing setQueue)
when nothing is current`.

### Known limitation, not fixed here

**The 500-item clamp runs before the concurrent-add merge.** In the resolver,
`parseArrayTolerant(..., 500)` validates the *incoming* payload, but the merge
below it writes `[...queue, ...outcome.survivors]` — so a session can end up
storing more than `MAX_SYNCED_QUEUE_ITEMS`, and the next full `setQueue` from any
client throws. Pre-existing for every `setQueue` caller; an append can reach it
more often. The mobile clamp also measures against the **local** queue length,
which is that same mirror. Follow-up filed: **#4732**.

### Other correctness notes

- **No active board → no row.** `addToQueue.append` is `undefined` when
  `activeBoard` is null, so the detail view's render gate collapses "you haven't
  picked a board" into "no row" instead of a control whose only possible answer
  is "Nothing here for your board yet" (the board-scoped fetch returns `[]`).
- **The empty-fetch canary is now keyed `<op>|<sourceId>`**, not `sourceId`.
  Sharing the key would have let whichever path fired first silently swallow the
  other's Sentry canary for the rest of the session — mutation-tested by
  `an append canary does not suppress the replace canary for the same playlist`,
  which asserts *both* reports.
- **Separate abort controllers**, so a row tap mid-append cannot cancel the append
  (and vice versa).
- **Confirmation fires once, with no double-fire.** The bulk append bypasses
  `commitQueueAdd` (which is what fires the snackbar for a single add), so it
  fires the snackbar itself, once for the batch.
- **Copy trigger boundaries**: `queueFull` (error toast) when remaining capacity is
  0 *or* the provider takes none of the batch; `nothingToAdd` (info toast) when
  the board-scoped fetch returns empty; `failed` (error toast) on a non-abort
  throw; the queue-added snackbar whenever `appended > 0`. All five have tests.
- **No BLE surface is touched, so no Fable review.** The pointer decision is
  BLE-adjacent and deliberate: `currentClimbQueueItem` is the only input the BLE
  auto-sender reads, and the append never moves it, so the wall never repaints.

## Release Notes

- Add a whole playlist to your queue without wiping what's already in it — the new **Add to queue** row on any playlist keeps every climb you and your crew already lined up.
- Starting a playlist now asks properly: keep your queue, add the playlist behind it, or start fresh. No more choosing between losing your queue and giving up.
- Works on smart playlists too — Liked Climbs, Five Stars and Projects all get the same row.

## Test plan

Run from the worktree on Linux, on the current head, **serially** (see the note
below). No emulator — a single shared emulator is held by the orchestrating
session for this batch — so **nothing below is a device check**. A visual pass on
the row, the fork's three actions and the German label lengths is still
outstanding. QA script: `.boardsesh/qa-notes.md`.

| command | result |
| --- | --- |
| `vp run test:mobile` | pass — **701 files, 6966 tests, 0 failures** |
| `vp test run --project i18n` | pass — 2 files, 101 tests (4 locales, key + `{{max}}` placeholder parity) |
| `vp check` | pass — 0 errors, 259 pre-existing warnings across 4477 files, none in the touched files |
| `vp run check:mobile-bundle` | pass — `android bundle: OK` |

**A note on `screenshot-locale.test.ts`.** Three earlier full-suite runs showed it
timing out at its hard 5 s budget, and I nearly reported it as pre-existing. It
is not pre-existing, and it is also not this PR: running the same file in this
worktree and in a clean `origin/main` worktree back to back, idle, gives
identical timings (1.03–1.70 s vs 1.06–1.69 s), and the full suite passes
701/701 here with nothing else on the machine. Every failure happened while a
second suite or `vp check` was running alongside it. The real finding is that the
test spends ~1.2 s of a 5 s budget re-importing the whole i18n module graph after
`vi.resetModules()`, so it is thin under parallel load. Filed as **#4759**.

### CI had to be forced — the `pull_request` event was dropped

`gh pr checks` on this PR read "no failures, nothing pending" while the workflow
carrying lint / typecheck / test-default / mobile-bundle / i18n **had never run on
the head commit**. Root cause is a GitHub event-delivery drop on this branch, not
a filter in the workflow:

- `ci.yml` triggers on bare `pull_request:` (default types include `synchronize`)
  with **no `paths` / `paths-ignore` filter**, so nothing in the repo's config can
  exclude a `packages/mobile/src/providers/**` change.
- `GET /commits/f718810a2/check-suites` shows only the two **push**-triggered
  suites (2 runs each). The `pull_request` suite was never created — so
  `gh run rerun` had nothing to rerun.
- The next push (`fbe6d9504`) produced **no suites at all** — neither `push` nor
  `pull_request` — while another branch got a full run four minutes earlier, and
  `githubstatus.com` reported Actions fully operational.
- Cancellation is ruled out: `cancel-in-progress` leaves a run with
  `conclusion: cancelled`, and one such run is visible on an earlier SHA here.

Filed as **#4758** — it can false-green any PR in this repo.

Recovered with `gh workflow run ci.yml --ref feat/4673-playlist-add-to-queue`.
A dispatch run is strictly *stronger* than the PR run it replaces: `changes`
forces every filter true for non-`pull_request` events, so it ran 3 `test-default`
shards and 2 `test-backend` shards where the PR run had skipped backend entirely.

Two empty commits were spent trying to coax a real `synchronize` out of GitHub;
neither produced one, so the head is `acb2db115` and the dispatch was re-run
against it.

**Verified by SHA, not by colour.** On `acb2db1157d4025f163389de2f3a540b9d9d48d4`,
`[.[] | select(.headSha==HEAD) | .name] | unique` returns exactly `["CI"]`, and
that run concluded `success` with every job green:

```
changes, large-files, lint, i18n, typecheck, codegen-drift, listing-guards,
db-migrations, deploy-config, rest-surface, test-ocr, mobile-bundle,
pg18-artifacts, test-location-sync-integration, commit-lint,
test-default (1,3) (2,3) (3,3), test-backend (1,2) (2,2), ci-status  -> success
release-notes, changelog-owned, test-report                          -> skipped
```

`release-notes` and `changelog-owned` are `pull_request`-only jobs, so they are
skipped on a dispatch run, and no `pull_request` event ever reached this branch
across three pushes. **Their last real result is from `f4b9b94c2`** — worth
knowing before this merges. The `## Release Notes` section below is present and
unchanged in shape since that run; per the known "release-notes reads a stale
body" behaviour it would need a delivered push event to re-evaluate anyway.

## Deviations

Called out rather than done silently.

1. **`added` / `addedPartial` copy is not in this PR.** Design directive 17 gave
   them their own toasts; the design lead's S6 overruled that and routed anything
   that landed climbs to `QueueAddedSnackbar` — whose copy lives in
   `DrawerHostProvider` and whose parametric signature
   (`showQueueAddedSnackbar({ kind, count })`) is being cut by #4712, which has
   not merged. This PR calls today's no-argument `showQueueAddedSnackbar()` with a
   `TODO(#4671)` at the call site. **Until that lands, a 12-climb append confirms
   with the hardcoded singular "Climb added to queue" and no count**, and a
   partial append (only reachable at the 500-item cap) says the same. One line to
   fix once the seam is on main; deliberately not a second snackbar API or a
   parallel toast path.
2. **No `announceForAccessibility` on success** (design §6). The confirmation is
   the snackbar per S6, so the announcement belongs with its copy in #4712 —
   adding one here would double-speak.
3. **Design directives 10, 11, 13 and 14** (restack the sheet buttons, raise them
   to `size="large"`, in-flight button states, fire the toast after dismissal) are
   moot: S7 replaced the sheet with `useChoose()`, which handles all four.
4. **Design §7 blocker 1 not discharged.** The bottom-docked back/play bar at
   748 dp was not reproduced — this batch's rule is not to boot the shared
   emulator. Nothing new is hung off that bar either way, which is what the
   blocker actually required. Worth one look during the batch capture pass.
5. **Scope: the backend constant swap is deferred.** `MAX_SYNCED_QUEUE_ITEMS`
   lands in `@boardsesh/queue` (additive; the mobile clamp consumes it), but
   `mutations.ts` keeps its inline `500`. Swapping it drags a backend test run into
   a mobile PR that the review already called "at the upper edge of one reviewable
   PR", and the user-visible feature does not need it. Filed as **#4732**,
   together with the clamp-before-merge gap above.
6. **`es` `detail.activateAll` is "Iniciar playlist", not the design's
   "Iniciar la lista".** Directive 18's point is that the play control's
   accessible name reads *identically* to the prompt's primary action, and the
   existing Spanish confirm button is "Iniciar playlist". en/fr/de already
   matched; following the design's literal Spanish string would have been the
   one locale where the two controls sound different.
7. **`appendQueueItems` grew a second broadcast shape** (per-item adds when
   nothing is current) that the plan did not have. Forced by the null-pointer
   correctness issue; rationale above.

## Variant and scheme

Every dp and colour in the design review this implements was measured on
**Material dark** (`#221A33` / `#181225`) at 810×1800 px / density 320 on an
Android emulator. Four of the surfaces here also ship to iPhone as **Liquid
Glass**, where `systemColors` and `actionColors` resolve to Apple `PlatformColor`
and the hexes are not knowable at author time — none of those numbers are claimed
for the app as a whole. The success glyph is `theme.actionColors.success`, which
is monochrome on Liquid Glass by design; the row's text label, not its colour,
carries the meaning.


marcodejongh:
## Review — code read against `main` @ `76cc8f5`, plus the backend resolvers

**Verdict: the hard parts are right.** The bulk append is atomic (nothing is written until every page resolves), it cannot construct a payload the 500-item resolver cap rejects, it reaches party peers, and the null-current-pointer branch in `appendQueueItems` is a genuinely better answer than the obvious one — `setQueue([...queue, ...items], null)` really would blank the crew's current climb, and this PR is the only version of this feature I've seen that noticed. The tests are real tests: they assert the things that would break. The one thing I'd block on is **feedback honesty** — as merged today, every append confirms with the singular, count-free "Climb added to queue", including the case where the batch was silently truncated at the cap. Everything else is sequencing and amplification.

---

### 1. Silent truncation, and a confirmation that never says how many — CONFIRMED · needs a human call

Two clamps drop climbs with no signal to the user:

- `use-playlist-activation.ts` (new `appendClimbsToQueue`): `fetchAllClimbsForBoard({ signal, limit: remainingCapacity })` where `remainingCapacity = MAX_SYNCED_QUEUE_ITEMS - queue.length`. The new `limit` slices the tail (`return limit === undefined ? climbs : climbs.slice(0, limit)`).
- `queue-provider.tsx` `appendQueueItems`: `items.slice(0, remainingCapacity)`.

Both are *correct* — the cap has to exist, and `parseArrayTolerant(ClimbQueueItemSchema, rawQueue, 'queue', 500)` at `packages/backend/src/graphql/resolvers/queue/mutations.ts:414` does throw rather than truncate, exactly as the new `packages/shared/queue/src/limits.ts` JSDoc says. The problem is what the user is told. The only success path is `showQueueAddedSnackbar()`, whose copy is hardcoded at `packages/shared/i18n/locales/en-US/session.json:305` — `"added": "Climb added to queue"`.

So: Liked Climbs with 900 entries, empty queue → 500 land, 400 vanish, and the app says *"Climb added to queue"*. Singular. No count. No truncation notice. That is the definition of the defect — a user gets fewer climbs than they asked for with no indication. And the count-free copy applies to the *ordinary* case too: a clean 40-climb append reads as if one climb was added.

The PR body is honest about this and defers to #4712. I checked: **#4712 is still OPEN.** If this merges first, the count-free snackbar is what ships. Options, in my order of preference:

1. Fire a `showToast` with a count alongside (or instead of) the snackbar for the bulk path only, the way the approved plan did. The catalogs would need four new plural keys — `added_one`/`added_other` and a capped variant — in all four locales.
2. Land the truncation case only: when `appendedCount < queueItems.length`, show `detail.addToQueue.*` truncation copy rather than the snackbar. Smaller change, kills the actual dishonesty.
3. Gate the merge on #4712.

Doing nothing is the one option I'd argue against.

### 2. Guaranteed merge conflict with the #4622 fix — CONFIRMED · sequencing

`fetchAllClimbsForBoard` is the unbounded drain that issue #4622 is about, and it has **two open PRs** rewriting it: **#4709** and **#4713**. This PR changes its signature (`{ signal, limit }`), its loop condition, and its return. That much is a normal rebase.

The unavoidable part: this PR **deletes** `packages/mobile/src/components/playlist/PlaylistQueueReplaceSheet.tsx` and `__tests__/playlist-queue-replace-sheet.test.tsx`. **#4709 modifies both of those files.** Whichever lands second gets a delete/modify conflict on two files plus overlapping hunks in `use-playlist-activation.ts`, `[playlist_uuid].tsx` and `smart/[type].tsx`. That's not a "rebase cleanly" situation — someone has to decide whether the sheet survives. Worth an explicit sequencing call with the #4622 owner before either goes green.

(The scope expansion that causes this — replacing the confirm sheet with a three-way `useChoose()` fork — is defensible on its merits and I like the result. It just makes this PR collide with a fix that was already in flight.)

### 3. The per-item fallback amplifies one failure into N — CONFIRMED, narrow trigger

`queue-provider.tsx` `appendQueueItems`, no-current-pointer branch:

```
for (const item of appended) {
  dispatch(...); mutations.addQueueItem(attributedItem).catch((error) => { ...; reconcileFailedContentMutation(error); });
}
```

`reconcileFailedContentMutation` (`queue-provider.tsx:779-785`) shows a rate-limit toast *and* calls `resyncQueueAfterMutationFailure` → a full `resyncQueueFromServer`. Per item. A party session with a dropped socket and a 300-climb append fires 300 rate-limit toasts and 300 full-queue resyncs. It's the same shape `commitQueueAdd` uses for a single add — the bug is only that it's now in a loop. A shared "one reconcile per batch" guard (a boolean captured in the closure) would be a few lines.

Rate-limit budget itself is fine: `addQueueItem` is `RATE_LIMIT_SESSION = 1200`/min (`mutations.ts:42`, `shared/helpers.ts:53`), so one 500-item append is comfortable; three back-to-back would trip it.

### 4. The per-item path gives up playlist order on peers — PLAUSIBLE

`mutations.addQueueItem(item)` is called with **no `position`** (`create-queue-mutations.ts:455-465` — `addQueueItem: async (item, position)`, position `undefined` here), so the resolver appends. The loop is synchronous but each call `await`s `resolveCore` before sending, and the server-side write goes through a CAS retry. I could not verify on a device that N parallel adds land in call order. If they don't, peers see the playlist shuffled, and the local client inherits that order on its next FullSync. The `setQueue` branch has no such exposure because it ships one ordered array.

Cheap fix if it's real: pass the index as `position`, or reuse the deferred-add path that already reads `getQueuePosition` at send time.

### 5. No dedupe against the live queue — CONFIRMED, deliberate divergence, worth a decision

`appendClimbsToQueue` builds `seenClimbUuids` fresh, so it collapses repeats *across pages* (good, and tested) but not against what's already queued. Consequence: press "Add to queue" twice and the whole playlist appears twice; merge two crew members' overlapping playlists and every shared climb is queued twice.

I don't think that's automatically wrong — the single-add path deliberately allows the same climb twice. But it compounds with finding 1: the confirmation says "Climb added to queue" with no count, the row returns to its idle state, so a user who isn't sure it worked presses again and silently doubles their queue. If the count lands, this becomes much less sharp. Flagging it because the approved plan went the other way and the issue's motivating case is exactly the overlapping-playlist merge.

### 6. Double canary on one empty fetch — CONFIRMED, cosmetic

Path: row tap → `replaceQueueWithPlaylist` → empty fetch fires `reportEmptyBoardFetchOnce('replace-queue-empty')` → post-drain re-check finds future items → fork → user picks "Add to queue" → `appendClimbsToQueue({ loadedClimbs: [] })` → `fetchedClimbs.length === 0` → fires `append-queue-empty` too. Two Sentry reports for one fetch. The `<op>|<sourceId>` keying is right and I like the mutation test that proves it; this is just the one case where both ops legitimately see the same empty result.

---

### Checked and clean

- **Failure atomicity (review point 2).** `appendQueueItems` is called only after the page loop resolves; any page rejection propagates out of `fetchAllClimbsForBoard` to the `catch`, which toasts `detail.addToQueue.failed` and writes nothing. No incremental commit anywhere. Queue is byte-identical after a mid-fetch failure. Test: *"surfaces an error toast and reports on a non-abort failure"*.
- **Cap correctness (review point 1).** Clamped in two places against `MAX_SYNCED_QUEUE_ITEMS = 500`; `remainingCapacity <= 0` skips the fetch entirely. I traced `parseArrayTolerant(..., 'queue', 500)` to `mutations.ts:414` — the constant and the JSDoc are accurate. The client cannot construct the rejected payload. The known gap you filed as #4732 is real and correctly characterised: `mutations.ts:523` does `[...queue, ...outcome.survivors]` *after* the parse, so the stored queue can exceed 500 — pre-existing for every `setQueue` caller.
- **The null-pointer branch.** Confirmed against the resolver: `mutations.ts:428-429` initialises `currentClimbQueueItem = null` and only overwrites it `if (rawCurrentClimbQueueItem)`, then `roomManager.updateQueueState` fans that out. An absent pointer really is a crew-wide clear. Fanning out per-item adds instead is the correct call and is better than what the approved plan specified (which would have activated `items[0]` and taken the wall). Both tests here are the right shape — one asserts `setQueue` is *not* called, the other asserts N `addQueueItem` calls.
- **Party correctness (review point 4).** `appendQueueItems` → `setQueue` → `create-queue-mutations.ts` `SET_QUEUE_WITH_BASELINE` → `mutations.ts:396` FullSync. Peer concurrent adds survive: `collectConcurrentAdds` (`set-queue-merge.ts:67-125`) folds the post-baseline window and `mutations.ts:523` re-appends survivors, excluding the caller's own adds. Solo is a true no-op.
- **Smart playlists + the 60/window limit (review point 6).** `applyRateLimit(ctx, 60, 'smartPlaylist')` is at `smart-playlists.ts:274`. The append path is bounded by `limit = remainingCapacity ≤ 500` at `PLAYLIST_SUGGESTION_REFRESH_PAGE_SIZE = 100`, so ≤ 5 requests per press — nowhere near it. The fork's "add instead" branch reuses the replace path's already-drained list rather than re-fetching, so it adds zero requests. Smart screens are wired and there's a test that says why (no overflow menu ⇒ the header row is the only affordance they can ever have).
- **The destructive confirm stays shut on an append (review point 9).** `expect(mocks.choose).not.toHaveBeenCalled()` in *"pages the board-scoped list and appends it without touching setQueue"*. With the sheet gone this is the right equivalent assertion. The `beforeEach` default of `mocks.choose.mockResolvedValue('cancel')` is a nice touch — a test that forgets to answer can't pass by destroying the queue.
- **Tests generally.** These would fail with the bug present. The 499-item paging test, the `appendedCount === 5` clamp test, the `activateFirstWhenIdle` pair, the abort-ref independence test, and the canary-key mutation test all pin behaviour rather than shape. The `PreSessionView` test asserting `{ activateFirstWhenIdle: true }` is exactly the guard the default-flip needed.
- **i18n.** All six new `detail.addToQueue.*` leaves plus `queueReplace.addInstead` are present in `en-US`, `es`, `fr`, `de`; `{{max}}` appears in all four `queueFull` values and nowhere it shouldn't. No plural forms needed because nothing carries a count (see finding 1 — that's the same fact from the other side). Spanish uses **plafón** (`"Aquí no hay nada para tu plafón todavía"`), per `docs/i18n-spanish-glossary.md`. French avoids « envoyer » entirely and stays on **vous**, consistent with the neighbouring `queueReplace.message` — and swapping `activateAll` from "Mettre toutes les voies en file" to "Lancer la playlist" also drops a stray "voies" for board climbs. German is informal **du**, uses **Board** (neuter) and **Warteschlange**, no "senden". No orphans — every leaf has a live `t()` call site.
- **Repo rules.** No `any` in the added lines (the only hits are `expect.any(Error)`). Mobile styling is `StyleSheet.create` + `spacing[2]` + `actionColors.success` (which exists — `variant-tokens.ts:21-32`) with no hardcoded colours. Both variants render the row and both have a test. `PlaylistAddToQueueRow` is `memo`'d, `append` is a stable `useCallback`, and `activatePlaylistClimb` does not depend on `isAppending`, so the spinner flip doesn't churn `renderItem`. JS/TS/JSON/MD only — OTA fingerprint unchanged. No BLE surface, so no Fable review needed; the reasoning in the body about `currentClimbQueueItem` being the BLE auto-sender's only input is correct.
- **Provider availability.** `QueueSnackbarProvider` wraps the tabs at `packages/mobile/app/_layout.tsx:507`, so the new `useQueueSnackbar()` call can't throw at render. `useChoose` supports `destructive` / `cancel` / `cancelValue` as used (`dialog-provider.tsx:11-24`, `:56-58`).
- No dangling references to the deleted sheet anywhere on the branch.

---

**Summary of what I'd want before this leaves draft:** the count/truncation story from finding 1 (product decision), and a sequencing agreement with #4709/#4713 from finding 2. Findings 3 and 5 are small code changes I'd take in this PR; 4 wants a device check; 6 is optional.

marcodejongh:

Conflicting: use-playlist-activation.ts, queue-provider.tsx-adjacent test files, index-keyboard.test.tsx.

Issue #4735, opened 2026-08-24
A 21st board is unreachable: myBoards is capped at 20 on both board surfaces


marcodejongh:
Pre-existing, and more consequential after #4623.

## The cap

Both board surfaces call `useMyBoards(undefined, …)` with no input:

- `packages/mobile/app/boards/index.tsx` (the picker)
- `packages/mobile/app/boards/manage.tsx` (the download console)

With no input the resolver applies `limit ?? 20` (`packages/backend/src/graphql/resolvers/social/boards.ts:1280`). Neither surface pages. So a user with 21+ owned-or-followed boards has one that cannot be switched to, edited, deleted, unfollowed, or downloaded — from anywhere in the app.

## Why it matters more now

#4623 made the picker's horizontally-scrolling carousel the primary place to edit or remove a board, and the drill-in shares the same capped query. Before, the cap was mostly a discovery annoyance; now it also hides destructive and management actions.

It is also silent: nothing tells the user boards are missing.

## Who hits it

Gym staff and anyone following a lot of gym boards. Worth a prod read of the `user_boards` per-user count distribution before picking a fix — if the p99 is 6, a bumped limit is the whole fix; if there is a real tail, the carousel needs paging.

## Options

1. Bump the limit for these two call sites (cheapest; still a cap, just a higher one).
2. Page the carousel on end-reach — one page per reach, per the mobile performance checklist, never a drain-until-`hasMore` loop.
3. Page `/boards/manage`'s `FlashList` and leave the carousel capped, with the drill-in as the escape hatch. Wants a "showing 20 of N" line on the picker so the cap stops being silent.

ES-Alexander:
Does this have any impact on which boards are tracked in a user's logbook?

If not, it seems potentially ok to just have "my latest boards" with a fixed cap, ordered by recent usage, and if a user reaches that cap then adding a new board kicks the least recently used (public?) one off the list. With that sorting it seems unlikely to be a findability issue to have lots of boards though, so we could maybe increase the cap to whatever point might start causing issues with other parts of the application (like, as high as 100 or even 1000)?

Issue #4734, opened 2026-08-24
Gym admins and moderators have no edit entry point for boards they can edit


marcodejongh:
Pre-existing, surfaced while working #4623. Not introduced by that PR, and not closed by it.

## The mismatch

`canEdit` is set by the server for the **owner OR a role-editor OR a gym admin** (`packages/backend/src/graphql/resolvers/social/boards.ts:425-426`). `/boards/edit` gates on exactly that (`edit.tsx:69`).

But every *entry point* to that screen gates on viewer-ownership instead:

- before #4623: `BoardManageRow.tsx:333` gated tap-to-edit on `isOwned`
- after #4623: the board card's pencil gates on `item.isViewerOwner`

So a gym admin or community moderator with genuine edit rights on a board they don't own can reach `/boards/edit` only by typing the route. There is no button anywhere.

## Why #4623 did not fix it by gating on `canEdit`

`myBoards` returns boards you own **or** follow (`boards.ts:1290-1294`). Inside that list, `canEdit === true && isViewerOwner === false` *is* the followed-board case. Gating the card's single action slot on `canEdit` would put a pencil where a gym admin needs "Following", and take away their one-tap unfollow — trading one missing affordance for a worse one. #4623 gates on ownership, which is also what the issue asked for literally and what shipped behaviour already did, so nothing regressed.

Pinned by a test: `packages/mobile/app/boards/__tests__/index-board-actions.test.tsx` → "offers unfollow, never edit, on a board the viewer only follows", whose fixture carries `canEdit: true`.

## What a fix needs

A second affordance, because the 168 pt card has one slot and it is spoken for. Candidates:

- a long-press board-actions menu on the card (the app's canonical per-item secondary-action pattern is `ClimbListRow` → `openClimbActions`; the natural host, `BoardDetailSheet.tsx`, is dead code with zero importers and would need resurrecting)
- an edit row on the gym-admin boards surface
- a row action on `/boards/manage`

Whichever it is, it should not be a third glyph on the thumb.

Issue #4733, opened 2026-08-24
Re-scope /boards/manage into an offline-downloads console


marcodejongh:
Follow-up from #4623 (PR: the "Change board" + "My Boards" merge). Filed as a separate issue because it is a product decision about the download funnel, not part of what #4623 asks for.

## Where #4623 left it

`/boards/manage` kept its route, its title, its owned/followed grouping, its identity resolution, its create CTA, its empty state and the whole offline console. It lost only the destructive/edit affordances, which now live on the board cards in `/boards`. It is reached from a "Manage boards" drill-in under the "Your boards" carousel, plus `OfflineSpotlightCard.tsx:40`.

## What this issue is

The screen's remaining job is the per-board offline-download console: the toggle, the live status caption, the progress bar, the snapshot-retry / paged-fallback notices and "retry fast download" (the shipped surface for #4311 / #4313 / #4316 / #4452). Everything else on it is now vestigial. Re-scoping it to an "Offline downloads" screen means stripping:

- `noUserIdAvailable` → `shouldUseOfflineList` (`manage.tsx:374-375` pre-#4623 numbering)
- the identity loading gate (`:623-635`) and the `!currentUserId` error state (`:640-659`)
- both `offlineItems` branches (`:376-407`)
- the `ManageItem` union with its `keyExtractor` / `getItemType` / `EMPTY_ITEMS` (`:72-75`) and `renderItem`'s header branch
- `onCreate`, the header "Create a board" button, the "No boards yet" empty state and the `myBoards` refresh control

plus:

- retitle `_layout.tsx:38` (the `common:myBoards.title` key, currently "Manage boards", is shared with the drill-in label — decide whether the two split)
- **rewrite `manage-offline.test.tsx` (1,386 lines).** Its header comment says it exists to pin exactly the grouping-survives-offline behaviour this strip removes. #4623 touched it narrowly on purpose.
- amend `docs/offline-reads.md:25`, which cites the `currentUserId` error state as the justification for allowlisting `['profile']` in the persisted cache. Removing that state strands the rationale.

## Open question before any of it

Settings → Storage (`StorageSettingsScreen.tsx:170-180`) already answers "how much disk". This screen answers "is this board ready". Two overlapping surfaces is the reason to think about it, and merging them is on the table — but Storage has no progress bar, no retry-fast-download, no snapshot-retry notice, and labels rows by scope (`storage-board-label.ts`) not board name, so it is not a drop-in.

PR #4736, opened 2026-08-24
feat(mobile): one Boards page instead of Change board + My Boards

### LINKED ISSUE #4623: [UI/UX]: Combine "Change board" and "My Boards" pages
The "Change board" page already shows your boards at the top, and allows creating new ones.
The unique features of the "My Boards" page are indicating which boards you follow vs own, editing the details of a board, and deleting a board from your list.
It is confusing to be looking at the same boards but not be able to do some of the actions just depending on which page you're on.

I propose having a single "Boards" page, which is like the current "Change board" page but
1. adds an edit icon button for all owned boards, and a follow/following icon button for all unowned boards
    - This could be in the top right corner, to complement the offline board icon button that's already in the top left corner, and move the active board tick icon to the bottom right corner instead
2. adds an "Edit"/"Done" button to the right of the "Your boards" header to enable the board deletion mode

marcodejongh:
## Summary

Closes #4623.

"Change board" and "My Boards" were two rows in the same drawer card, 48.5 dp apart, with the **identical icon** and the **identical chevron** — and the same board list behind both. Nothing on screen said which one edited a board and which one switched to it.

This collapses them into one **Boards** row onto `/boards`, and moves the actions that only existed on the manager onto the picker's cards.

**On `/boards`:**

- Each card in "Your boards" gets an ownership slot top-right: a **pencil** on a board you own (opens the edit form), and a **non-interactive person-with-check** on one you follow. The second one is a status indicator, not a button — see B1 below.
- **Edit** beside the "Your boards" header swaps that slot for a labelled **Delete** / **Unfollow** button under each card. Delete confirms. Unfollow confirms *only* when it would clear your active board — the one case with a side effect beyond the list.
- The active tick moves off the top-right corner and becomes an **"Active" pill bottom-left**. The distance pill stays bottom-right, so an active Near-you board can show both without stacking.
- A **Manage boards** row under the carousel drills into `/boards/manage`.

**On `/boards/manage`:** grouping, identity resolution, the create CTA, the empty state and the whole offline-download console are untouched. It loses only the edit/delete/unfollow affordances that now live on the cards.

**Followed boards were never missing from the picker.** `myBoards` returns boards you own *or* follow (`boards.ts:1290-1294`) and the carousel has always been built from all of them — they were just undifferentiated. That is what the `person.check` glyph fixes, and it is why there is no second "Following" carousel here. `sortViewerOwnedFirst` leads with your own boards, because the server's `desc(userBoards.isOwned)` orders by a column meaning "a real wall", not "owned by the viewer".

### Guards on a modal opened mid-task

`/boards` is `presentation: 'modal'`, pushed from **twelve** non-test call sites — onboarding, the in-session chrome, the cross-board queue-add gate, mid playlist render. Hanging a delete off that needs care:

1. Edit mode is route-local `useState`, so it cannot survive a dismiss — every one of the twelve entries gets it off.
2. The Edit control is hidden during onboarding, offline, with no boards, and with no resolvable identity.
3. Delete confirms; unfollow confirms when it would clear the active board.
4. Per-card in-flight state: the glyph becomes a spinner and a second tap does nothing.
5. Auto-exit edit mode after a successful delete — the carousel has just reflowed under a finger still over a red button.

The resting surface has **no** destructive control at all — the followed-board glyph is a `View`, not a `Pressable`, and publishes no custom action either. Reaching a removal costs a deliberate Edit tap plus a labelled 44 dp button plus (for delete) a confirm. That is more friction than today's swipe-to-delete, which is already two taps from a live session via the avatar toolbar action.

Guard 2 covers the **whole action slot**, not just the toggle: during onboarding, offline, with no boards or with no resolvable identity the screen withholds `actionFor`/`onAction` entirely, so no card renders a slot. Switching boards is untouched in every one of those states.

### Trade: swipe-to-delete is gone

`SwipeableRow.tsx` (230 lines) had exactly one importer, `BoardManageRow`. With the destructive affordances moved, it is deleted. **Swipe-to-delete a board no longer exists** — the deletion gesture is the Edit mode above, which is discoverable where a swipe was not. (The two `QueueItemRow` "mirrors SwipeableRow" hits were comments, not imports; both are reworded. Queue swipe-to-remove is untouched.)

### Two live defects fixed on the cards being edited

- The corner badge glyph was `brandColors.primary` on a **hardcoded white** disc: `#A78BFA` on `#FFFFFF` = **2.72:1**, a WCAG failure, visibly washed out. Now `brandColors.primaryFill` = **5.70:1** (light `#6D28D9` = 7.10:1). The disc is a fixed white, so unlike every other number in this batch this ratio holds on both variants and both schemes.
- `hitSlop={8}` on a 26 dp badge is a **42 dp** target. Both corner badges are now `hitSlop={9}` = exactly 44 dp. On a 168 dp thumb the two rects span x −1…43 and x 125…169: 82 dp apart, no overlap.

### Shared seam

The design lead's ruling (S3) assigned this edit to #4671 and told #4623 to consume it. #4712, the #4671 implementation, does not touch `src/lib/row-accessibility-actions.ts`, so the seam landed here instead — verified conflict-free against that branch.

`rowAccessibilityActionsWith` in `src/lib/row-accessibility-actions.ts` is now **variadic** (`...nested: AccessibilityActionInfo[]`). The board card publishes three actions through it — `activate`, `boardAction`, `download` — because the outer `Pressable` absorbs its children on iOS and VoiceOver can otherwise never reach the corner glyphs. The existing single-arg call sites (`ClimbListRow.tsx`, `QueueItemRow.tsx`) compile untouched. #4672 / #4673 may want it too.

### Contrast + variant honesty

Every ratio above except the badge glyph was measured against **Material dark** (`#221A33`). The edit-mode footer button is `#F87171` on `withAlpha(error, 0.12)` over `#221A33` = **5.04:1** dark, `#C81E1E` over white = **4.70:1** light — AA with no margin, which is why the alpha is pinned at 0.12. On iPhone this ships as Liquid Glass, where `systemColors` resolve to Apple `PlatformColor` and the hexes are not knowable at author time. The badge-glyph number is the one that generalises, because its disc is a hardcoded white.

### Telemetry note

`/boards/manage` goes from a top-level drawer row to one tap inside the `/boards` modal. `OfflineBoardToggled` / `OfflineBoardDownloadCancelled` keep `source: 'manage'` (the route is unchanged), which is exactly what makes the move measurable: **take a 14-day `OfflineBoardToggled { source: 'manage' }` baseline before merge and re-read it 14 days after.** If the download funnel drops materially, the fix is to promote the drill-in (or restore a drawer row), not to revert.

### Scope

Split per the adversarial review. This PR is the user-visible merge only. Filed as follow-ups, no code here:

- #4733 — re-scope `/boards/manage` into an offline-downloads console (+ the 1,386-line `manage-offline.test.tsx` rewrite and the `docs/offline-reads.md` amendment).
- #4734 — the `canEdit` vs `isOwned` edit-gate mismatch: a gym admin who can edit a board they don't own has no edit entry point, before or after this PR.
- #4735 — the 20-board page cap, shared by both surfaces, so a 21st board is unreachable from either.

**No BLE code is touched** — the delete/unfollow path never reaches `packages/shared/ble-protocol/` or `packages/mobile/src/lib/ble/`. No Fable review gate.

## Deviations

Four, all from the design review's literal wording, all deliberate:

0. **D-9's shortened Create label is scoped to the tile it was measured for.** D-9 changes `boards:mobile.discovery.create` from "Create board" to "Create", justified by the 84 dp mode tile (68 dp of text) truncating in all four locales. That key has **four** call sites: the tile, the `/boards` empty-state button, and two buttons on `/boards/manage`. The other three are full-width CTAs with no `+` glyph and no surrounding noun, and a bare "Create" / "Crear" / "Créer" / "Erstellen" under a "No boards yet" headline is worse than the truncation D-9 set out to fix. So the tile gets a new `mobile.discovery.createTile` = "Create" and `create` keeps its full value. Same fix on the surface D-9 measured, no collateral on three it did not.

1. **The composed card accessibility label is a filtered join, not "one of" three keys.** §4 specifies `"{title}, {subtitle}"` + one of `activeBadge` / `ownedBadgeAria` / `followingBadgeAria`. An active board you also own would then announce only "Active" and lose the ownership signal that S13 requires, since the glyph is the only owned-vs-followed cue left on this surface. The label appends both: `"Marco garage, Original 12×12, Active, Your board"`.
2. **The edit-mode footer button reaches VoiceOver as the card's `boardAction` custom action**, not as a separately focusable element. QA step 18 asks for "its own element". On iOS the card's `accessible` container is a leaf, so a nested element is unreachable by construction — the custom action is the repo's own answer to this (`ClimbListRow.tsx:310-325`) and the only one that works. On Android TalkBack it *is* its own element, because the button keeps its own role and label.
3. **The Manage boards drill-in also renders on the offline branch.** D-6.19 lists what offline loses (action slot, Edit control) and says "no drill-in state change". With the drawer's second board row gone, omitting it offline would leave `/boards/manage` — which has a working offline list and a working local download toggle — unreachable. The drill-in is navigation, not a mutation, so it stays.

## Release Notes

- One Boards page instead of two. "Change board" and "My Boards" were the same list with different buttons — now it's one place to switch, edit, and clear out boards.
- Every board in "Your boards" says whether it's yours or one you follow, and the board you're on now reads "Active" instead of a bare tick.
- Tap Edit to delete a board you own or stop following someone else's, straight from the card. Swiping a row to delete is gone — the Edit button replaces it.

## Test plan

Run from the worktree, all green:

| Command | Result |
| --- | --- |
| `vp run typecheck:mobile` | clean, 0 errors |
| `vp run test:mobile` | **704 files, 6987 tests passed** |
| `vp test run --project i18n` | 2 files, 101 tests passed (catalog parity across all four locales) |
| `vp run check:i18n:orphans` | OK — 3858 catalog keys, every one referenced |
| `vp check` | clean; 0 warnings in any file this PR touches |
| `vp run check:mobile-bundle` | android bundle: OK |

New / extended cover:

- `board-card-actions.test.ts` (new, 8 cases) — the three-branch action, `readOnly` sweeping every combination to `null`, delete never offered on a non-owned board, a type-level pin that the union has no `'follow'` member, and `sortViewerOwnedFirst` including the no-`ownerId` offline-snapshot fallback.
- `board-discovery-card.test.tsx` (27 cases, mocks rebuilt) — the mock now forwards `style`, `testID` and `disabled`, and the animated Pressable keeps its `onPress`, so "the action fires `onAction` and **not** `onPress`" is finally falsifiable. Corner assertions read *which edge keys are present*, never pixels. Includes the pill-placement pin (Active bottom-left + distance bottom-right on one card), the pending-spinner state, the composed ownership label, and one stable `accessibilityActions` identity across a re-render.
- `index-board-actions.test.tsx` (new, 27 cases) — the gym-admin pin (`canEdit: true` on a non-owned board still resolves `unfollow`, never `edit`), the Edit gate across all four suppressions, both confirm paths and both decline paths, `forgetOfflineBoard` + conditional `clearActiveBoard`, auto-exit after delete, and the **resilience pin**: with `useProfile` erroring and no stored id the screen still renders the carousel and still activates on tap — no "Something went wrong", no identity spinner.
- `row-accessibility-actions.test.ts` (new, 4 cases) — the variadic helper, including the zero-arg and multi-arg shapes.
- `board-items.test.ts`, `index-offline.test.tsx`, `manage-offline.test.tsx`, `board-manage-row.test.tsx`, `user-drawer-provider.test.tsx` — extended / stripped in step.

### Independent review pass — two blockers (third commit, `cbe38fd31`)

**B1 · The resting `person.check` badge was a one-tap, unconfirmed unfollow, and the onboarding guard did not cover it.** It rendered as a `Pressable`, and `canEditBoards` gated only the Edit/Done toggle — so on More → Replay onboarding → `/boards?source=onboarding`, a followed gym board carried a 26 dp white disc, visually identical to the non-interactive download-status badge on the opposite corner, that removed the board in one tap. No confirm, no toast, no undo; re-following needs geolocation range or the gym finder. Live mid-session from `InSessionView.tsx:444` too. This is the exact shape the design review rejected for the edit-mode corner ⊖, and it contradicted the "no destructive control at rest" paragraph this body used to make.

Fixed: the Following glyph is a non-interactive `View` with no press target, no `button` role and **no published custom action** (or VoiceOver would hand back the tap the touch surface just gave up). Unfollow lives only on Edit mode's labelled footer button. The per-card slot now shares the `canEditBoards` predicate with the toggle. Mutation-tested — making the glyph a `Pressable` again turns its pin red.

**B2 · The owner name was being deleted from followed boards on `/boards/manage`.** The pre-existing `isOwned ? boardRowSubtitle(board) : (ownerDisplayName ?? boardRowSubtitle(board))` was already correct — adversarial objection #4 targeted a *proposed* rewrite, not shipped code, so this PR "fixed" a bug that never existed. The justification ("the Following group header carries whose board it is") was false: `buildManageItems` pushes a static "Following" string that names nobody, and the picker's cards have never shown an owner, so the owner would have appeared nowhere in the app. Two followed gym boards with the same truncated title and the same layout became indistinguishable. Restored, with both branches and the no-owner fallback pinned. Also mutation-tested.

Three should-fixes from the same review: `flexShrink: 1` on the footer label (German "Nicht mehr folgen" runs ~158 dp at the 1.2 cap against 148 dp of usable width, and RN will not ellipsize a non-shrinking `Text`); corner badges inset by 9 rather than 8, because **Android clips a child's `hitSlop` to its parent's bounds** and these discs live inside a `thumb` with `overflow: 'hidden'` — the old inset clamped the target to 43 dp, so the 44 dp claim only held on iOS; and the dead `readOnly` input dropped from `boardCardAction`, since the real gate is the caller withholding `actionFor`.

### Adversarial review pass (second commit, `3bd38c1cc`)

The diff went through an adversarial review after the first push. Five real findings, all fixed in `3bd38c1cc`:

- **`useMutation` returns a fresh object literal every render** (`{ ...result, mutate, mutateAsync: result.mutate }`), so depending on the mutation *objects* rebuilt the card action handler on every commit — which churned the carousel's `renderItem` and broke **every** card's memo, not just the mutating one. `MutationObserver.bindMethods()` binds `mutate` once in its constructor, so `mutateAsync` is the stable half; the handler depends on that now. The memo comment on the card claimed the opposite and has been corrected.
- The `create` string scoping above.
- **The card's `onAccessibilityAction` switch had no test**, and it is the only VoiceOver route to either nested glyph — a mis-wired `actionName` would have shipped green. Two tests added, then mutation-checked by swapping the two routes (both go red).
- **One assertion could never fail**: `expect(queryByText('Something went wrong')).toBeNull()` in the resilience pin, where that literal was not in the file's `t` stub. Key mapped, then mutation-checked by adding a fail-closed identity gate to the screen — both resilience pins go red, which is exactly the regression they exist for.
- **Restored a pin lost with the manage-screen strip**: a rejected delete must not drop the board's offline snapshot.

Plus two defensive tightenings the review flagged as unreachable-but-wrong: the ownership-unknown branch now bails explicitly instead of collapsing `undefined` into "followed" via `=== true` (mutation-checked), and the VoiceOver board action routes to the same handler the touch path uses so the destructive button keeps its `hapticHeavy()`.

My own reviewer raised the resting-unfollow risk here first and I left it, on the grounds that D-1.3 specifies a `Pressable` and D8.4 says ordinary unfollow stays confirm-free. That was the wrong call: the design document contradicts itself (`design-4623:213` promises no destructive control at rest, `:298` rejects exactly this shape), and where it does, the safe reading wins. Fixed above as B1.

**Not verified — needs a device pass. This PR stays in draft until it is done.**

- No emulator run: a single Android emulator is held by another session in this batch. Everything above is unit-level.
- **iOS VoiceOver (QA step 18) is declared blocking by both the plan and the design review, and is unrun — this host is Linux, so it cannot be cleared here.** Android nested-button semantics differ from iOS's, where the outer `Pressable` absorbs children. Needs a Mac reviewer to confirm the rotor exposes `boardAction` and `download` on a resting card, and the footer button on an edit-mode card, and that the disabled card body is not announced as a button.
- Dynamic type at fontScale 1.5: both bottom pills and the footer label cap at 1.2 (`CHROME_LABEL_MAX_FONT_SCALE`), but the German "Nicht mehr folgen" in a 168 dp box wants an eyeball.


marcodejongh:
**Owner decision on the delete affordance — this PR ships the option that was turned down.**

Flagging a divergence rather than a defect. This PR was opened in parallel with a separate planning pass on #4623 that put exactly this question to the repo owner. The question asked was where board deletion should live once the two pages merge, given that `deleteBoard` (`packages/backend/.../boards.ts:2372-2424`) is an owner-only soft delete of the board *itself* — every follower loses it, kiosks get a tombstone, and **there is no undelete mutation anywhere in the backend**. It is irreversible from the app.

The owner chose:

> **The merged Boards page gets follow/unfollow and an edit pencil. Real deletion moves to `/boards/edit`, behind the existing destructive confirm.**

This PR instead keeps delete on `/boards` behind an Edit-mode toggle plus a labelled button plus a confirm — the alternative that was explicitly declined, and the one item 2 of the issue text literally asks for.

To be fair to this PR: the guards here are genuinely strong, and stronger than the issue asked for. Edit mode being route-local `useState` so none of the twelve entry points can resume into it, the per-card in-flight lock, the auto-exit after a successful delete, and the confirm are all the right instincts, and the "no destructive control on the resting surface" framing is a fair answer to the modal-opened-mid-task worry. The friction comparison against today's swipe-to-delete is also accurate.

The decision still stands on a different axis, though: not *how much friction* guards an irreversible action, but *which surface hosts it at all*. `/boards` is pushed from twelve non-test call sites including onboarding and the cross-board queue-add gate; `/boards/edit` is reached deliberately. That is the distinction the owner was buying.

Concretely, what that means for this PR:

- Keep the pencil (owned) and the person-with-check unfollow (followed) on the cards — both are wanted.
- Keep unfollow in Edit mode, including the confirm for the clear-your-active-board case.
- Move **delete** to `/boards/edit`, reachable from the card pencil, behind the confirm that already exists there.
- The "Manage boards" drill-in row and everything in the follow-up split (#4733, #4734, #4735) are unaffected.

Two things worth keeping from this PR regardless of where delete lands: the corner-badge contrast fix (`brandColors.primary` on hardcoded white was 2.72:1 — a real WCAG failure) and the `hitSlop` 8→9 correction that takes the 26 dp badge to a true 44 dp target. Those are independent of the delete question.

One flag on the trade this PR makes: deleting `SwipeableRow.tsx` removes swipe-to-delete entirely. If delete moves to `/boards/edit`, re-check whether that deletion still holds — the component may still have a home there.

Not requesting changes unilaterally, since this is the owner's call and it has now been made twice in two places. Tagging @marcodejongh to confirm which way this lands.

marcodejongh:
## Review: destructive-action safety, dangling nav, capability loss, perf, i18n, contrast

Read against the approved plan for #4623 and against a second, independent implementation of the same issue. **I could not find a path where this deletes a board by accident.** Every delete goes through `confirm()`, the mutation is keyed on `board.uuid` (never a list index), and the label the user reads and the mutation that fires are both derived from the same pure `boardCardAction()` call on the same `item` in the same render — there is no window where the button says "Unfollow" and the handler deletes. The contrast arithmetic I recomputed matches to the second decimal. What I did find is one real screen-reader regression, one undeclared information loss, and a double-tap window that sits in a place the PR's own guard doesn't cover. None of them are merge blockers on their own; findings 1 and 2 are cheap to fix and I'd fix them before this goes out.

*(The `/boards` vs `/boards/edit` placement of delete is already flagged separately — not re-litigated here. Everything below is about the delete path's correctness on its own terms.)*

---

### 1. The new composed `accessibilityLabel` silently drops the distance — CONFIRMED

`packages/mobile/src/components/board-discovery/BoardDiscoveryCard.tsx` — before this PR the card passed **no** `accessibilityLabel` (base file line 114 is a bare `accessibilityRole="button"`), so the `accessible` container auto-composed its announcement from its child `<Text>`s, distance pill included. The PR now sets one explicitly:

```
const accessibilityLabel = [item.title, item.subtitle, item.isActive ? activeLabel : null, ownershipLabel]
```

An explicit label replaces the auto-composed one. A Near-you card that used to announce "…, 320m" now announces only title + subtitle. The distance is the *only* thing that distinguishes two same-named gym boards in that carousel, and it is the whole point of the "Near you" section.

The new tests pin the regression rather than catching it — `board-discovery-card.test.tsx`, `announces ownership in the composed label` expects exactly `'Tension 8x10, Original 12x12, Your board'`, and the corner-budget test that *does* set `distanceMeters: 320` never reads the label.

Fix is one array entry: `item.distanceMeters != null ? formatDistance(item.distanceMeters) : null`.

### 2. `ownerDisplayName` is now rendered nowhere in either board list — CONFIRMED, and not in the PR description

`packages/mobile/src/components/board-discovery/BoardManageRow.tsx:129` was:

```ts
const subtitle = isOwned ? boardRowSubtitle(board) : (board.ownerDisplayName ?? boardRowSubtitle(board));
```

and becomes unconditional `boardRowSubtitle(board)`. The new comment justifies it with "whose board a followed one is comes from the *Following* group header instead" — but that header reads "Following" / "Siguiendo" / "Abonné", not the owner's name. After this PR `ownerDisplayName` survives only in `BoardDetailSheet.tsx:82` and `SerialReuseConfirmSheet.tsx:79`, neither of which is reachable from the picker or the manage list, and the picker card's own `followingBadgeAria` is likewise just "Following". So a user following three gym boards has no surface left that says whose they are.

The stated reason (the row no longer receives `isOwned`) is avoidable — `manage.tsx:397` still computes `item.isOwned` for the grouping; it costs one prop to keep the branch. And the replacement test (`board-manage-row.test.tsx`, `never subtitles a board with its owner, even when the wire carries one`) locks the loss in, so a future restore has to delete a test.

The bug the comment describes (owned boards subtitled with *your own* name) is real and worth fixing, but it was already impossible: the old branch only reached `ownerDisplayName` when `isOwned` was false.

### 3. The in-flight lock doesn't cover the confirm — CONFIRMED as a code fact, low practical severity

`app/boards/index.tsx`:

```ts
const pendingActionKey = deleteBoard.isPending ? (deleteBoard.variables ?? null) : …
```

`isPending` only flips *after* `await confirm(...)` resolves and `mutateAsync` is called. So for the entire life of the confirm dialog, `isActionPending` is `false` on every card and both of the card's guards (`disabled={isActionPending}` and `if (isActionPending) return`) are inert. `dialog-provider.tsx` queues concurrent prompts by design rather than coalescing them ("Concurrent prompts queue… the native Alert queues itself"), so two taps that land before the Alert presents raise two confirms, and confirming both calls `deleteBoardAsync(uuid)` twice.

In practice the window is roughly one frame — once the native `Alert` or the Paper scrim is up, taps behind it are blocked — and the second delete just errors into the existing toast, so this is a UX wart rather than data loss. A `useRef<string | null>` latch set *before* `await confirm(...)` and cleared in a `finally` closes it properly.

Related, same line: `pendingActionKey` is a single global scalar. Starting a delete on card B while card A's is in flight moves the spinner off A onto B (`variables` follows the most recent call on the one `MutationObserver`) while both mutations still run. Worth knowing; not worth a second observer.

### 4. Nothing tests that edit mode dies on dismiss

The claim holds by construction — `isEditingBoards` is `useState` in a `presentation: 'modal'` route, and a dismissed native-stack route is removed from state and unmounted, so all twelve entry points do get it off. But `index-board-actions.test.tsx` has no unmount/remount case, so the property that makes the whole "destructive action on a mid-task modal" argument safe is the one property with no pin on it. Also worth a comment: it *does* survive the drill-in to `/boards/manage` and back, which is correct but is the one case a reader will wonder about.

### 5. Minor

- `board-card-actions.ts` — `readOnly` is never `true` at either production call site (`index.tsx` passes a literal `false` in both `myBoardActionFor` and `runMyBoardAction`); the offline branch omits `actionFor` entirely instead. The branch exists only for the tests. Either wire the offline carousel through it or drop the parameter.
- `index.tsx` — both `catch { showToast(...) }` blocks swallow the error with no `reportError()`. A failed *board delete* leaving no Sentry breadcrumb is the one failure here you'd most want to see in production. Carried over from `manage.tsx`, so not a regression, but `packages/mobile/src/lib/error-reporting.ts:21` is right there and the other implementation of this issue calls it.
- `BoardDiscoveryCard.tsx` — `accessibilityActions` still publishes `activate` in edit mode while `handleAccessibilityAction` gates it on `!isEditing`, so TalkBack offers an "Activate" that does nothing.
- The memo comment ("`useMutation` returns a fresh object literal every render… `mutateAsync` is the stable half") is half right. Verified in the installed `@tanstack/react-query@5.101.4`: `useMutation` returns `{ ...result, mutate, mutateAsync: result.mutate }`, `result.mutate` is `this.mutate` bound once in `MutationObserver`'s constructor — *and* the returned `mutate` is a `useCallback(..., [observer])`, so it is stable too. Only the result **object** is fresh. The fix is correct; the stated reason implies `mutate` is unstable, which it isn't.

---

### Checked and clean

- **No accidental-delete path.** Resting state offers only pencil (owned → `/boards/edit`) or one-tap unfollow (followed). `'delete'` is reachable only via `isEditing && isViewerOwner`, and only behind a destructive `confirm()`. `boardCardAction` returning `'delete'` for a non-owned board is pinned by a test, and delete is owner-gated server-side as a second net.
- **Unfollow can never invoke delete.** The carousel picks `actionTitle` from `actionFor(item)` and `runMyBoardAction` re-derives from the same `item.isViewerOwner` + the same `isEditingBoards`, both pure, both from the same render. No divergence window. The `isViewerOwner === undefined` case bails explicitly instead of coercing to "followed".
- **Keyed by uuid, not index** (`myBoards.find(c => c.uuid === item.key)`), so a list reflow mid-gesture can't retarget. Confirm message names `board.name`. Auto-exit after delete, and the edit-mode button appears *below* the subtitle rather than where a finger already is.
- **Failure handling:** a rejected delete does not call `forgetOfflineBoard` or `clearActiveBoard` — pinned by a test.
- **No dangling navigation.** `/boards/manage` survives in this PR, so the two remaining pushes still resolve: `OfflineSpotlightCard.tsx:40` and the new drill-in (`index.tsx`, `onManageBoards`). `_layout.tsx:38` keeps the `Stack.Screen`. `navigateToManageBoards` is removed from all four of its references (`UserDrawerProvider.tsx:22,81,127,142`) plus `user-drawer.tsx:84,241`. Nothing in `app/+native-intent.ts`, nothing in `packages/mobile/.maestro/*.yaml` (no flow opens `com.boardsesh.app://boards` at all), no notification deep link. `docs/offline-reads.md:25` still describes `/boards/manage`'s hard "no user id" gate, which this PR leaves in place, so the doc stays accurate.
- **`SwipeableRow` had exactly one importer.** `BoardManageRow.tsx:6` is the only `import`; the other two hits (`QueueItemRow.tsx:179`, `queue-item-row-memo.test.tsx:770`) are comments and both are reworded. There is no standalone `swipeable-row` test to orphan.
- **Capability inventory on `/boards/manage`:** grouping (`buildManageItems`), identity resolution, create CTA, empty state, pull-to-refresh, the full offline console (`useSyncStatus` + per-scope bootstrap + all the status branches), `OfflineDownloadProgressBar:287`, retry-fast-download `:291`, and the `AccessibilityInfo.announceForAccessibility` live region `:204` are all still there.
- **Perf checklist** (rules verified in `.oxlintrc.json:133` / `vite.config.ts:64` before citing): `FlashList`, `keyExtractor` hoisted at module scope, card is `React.memo`, `renderItem` deps are functions + scalars with no array `.length` and no inline closures, no new context values. The `useSyncStatus()` claim checks out — `/boards` still derives offline state from settings + checkpoints and this PR does not add a live-progress subscription to it; `pendingActionKey` and `isEditing` are scalars, so a mutation starting rebuilds `renderItem` but only the one card whose `isActionPending` changed fails the memo compare.
- **i18n:** every new key (`mobile.manage.unfollowTitle/Message/Confirm`, `mobile.discovery.activeBadge/ownedBadgeAria/followingBadgeAria/createTile`) is present in all four catalogs. `userDrawer.changeBoard` removed from all four — the only key orphaned by the merge. `myBoards.title` still referenced (`_layout.tsx:38` + the new drill-in row). Spanish is on-glossary throughout ("plafón", masculine agreement in "tu plafón activo" / "Activo" / "Tu plafón"). French avoids « envoyer » and keeps board feminine consistently with the existing catalogue ("ta board active", "Active", "Ta board"). German is informal *du*, neuter *Board*, no *senden*.
- **Contrast, recomputed independently:** badge glyph `#A78BFA` on `#FFFFFF` = **2.72:1**; `primaryFill` dark `#7C3AED` = **5.70:1**; light `#6D28D9` = **7.10:1**. Error button `#F87171` on `withAlpha(error, 0.12)` over `#221A33` = **5.04:1**; `#C81E1E` over the same 12% tint on white = **4.70:1**. All five match the PR body exactly. Geometry too: `spacing[2]` is 8, so 26 + 2×9 = 44 dp, rects at x −1…43 and x 125…169, 82 dp apart.
- **`Pressable disabled` does not block children** (verified in the installed `react-native@0.86.2` — `disabled` only feeds the Pressability config, no `pointerEvents` change), so the edit-mode footer button is reachable inside the `disabled` card body as intended.
- **No `any`** anywhere in the diff; the only hardcoded hexes are inside test mocks; everything on the card reads `brandColors` / `systemColors` / `overlays` / `radii`. **JS/TS + JSON only**, so the OTA fingerprint is unchanged.

### Not verified here

iOS VoiceOver behaviour on the `disabled` card in edit mode (whether UIKit still exposes custom actions on an element carrying the not-enabled trait) — Linux host, no simulator, and the PR already flags this as needing a Mac pass. Same for the German "Nicht mehr folgen" in a 168 dp box at fontScale 1.5.

marcodejongh:
Marked ready for review with one gap stated plainly: **the iOS VoiceOver pass (QA step 18) has not been run.** This host is Linux; `check:mobile-simulator` and `mobile:screenshot` are macOS-only.

Ready-for-review, not ready-to-merge. That step gates deviation #2 — the Edit-mode footer button is reachable only as a custom action on a `disabled` container — so it needs a Mac reviewer before merge.

Everything else is closed: both blocking review findings fixed and **mutation-tested** (reverting either fix turns its pin red), CI 27/0, and no BLE surface touched, so no Fable review is required for this one.

PR #4761, opened 2026-08-25
fix(mobile): route playback in the climb creator, and stop the wall garbling routes

### LINKED ISSUE #4634: [Bug: UX] Climb: Create: Route playback doesn't work
There's a play button when creating a climb, but for routes it seems to light up a single set of LEDs that's a weird mix of some of the frames, and there's no speed control.

Currently that means you need to publish a route before you can actually play the moves automatically (which makes it difficult to test the climb before publishing it).

marcodejongh:
## Summary

Closes #4634.

Two separate defects sat behind one report. Both are fixed here.

**The wall was lighting a mix of frames.** Aurora (`aurora.ts`) and MoonBoard
(`moonboard.ts`) both tokenise a frames string with `frames.split('p')` and then
`split('r')`. Neither parses the `,"` frame separator or the `x<id>` off-token
that a multi-frame route is encoded with. Handed a raw route they light holds a
later frame turned **off** and silently drop the hold sitting immediately before
every frame boundary — its role parses as `NaN`.

That is not limited to exotic routes, and it is not limited to the creator.
`Duplicate frame` copies the active frame forward, so a route authored the normal
way encodes as `frame 0 absolute` + `,"` + only the new holds. On
`p100r42p200r43p300r44,"p400r43` the tokeniser drops hold 300 — the finish. One
hold destroyed per frame boundary, every time. A second failure rides along: an
`x<id>` off-token glues onto the preceding token, so the hold before it is
dropped *and* the hold a later frame explicitly turned **off stays lit**.

Aurora's own catalogue stores its **709 multi-frame climbs** in the same grammar.
So this was live data corruption on **every multi-frame climb ever written to a
wall** — from the queue auto-sender, undo, kiosk relight, the mirror toggle and
the iOS lock screen — since the tokenisers were written. The creator is only
where it is unavoidable, because there is no Play Drawer there to correct it.

The fix collapses to the route's **union** at `sendFramesToBoard`, the one choke
point every JS wall write funnels through (queue auto-sender, undo, kiosk
relight, mirror re-send, connect-and-light). That is already what the ESP32 gym
controller, the web kiosk, the web renderer and the share card show for a route
nothing is animating, so a phone-driven wall now agrees with the wall's own
controller. Flat strings return byte-identical, so 99.9% of the catalog is
untouched on every board.

**There was no playback in the creator at all.** The ▶ glyph is
`Set as active climb` — it pushes the WIP into the queue and returns. Nothing
animates, the drawer does not close, and it is disabled at ~1.1:1 contrast with
no explanation until the climb has a start and a finish. The creator now mounts
the same `PlaybackControls` the play drawer uses, driven by the shared playback
engine over the editor's own frames, and the ▶ glyph becomes a queue glyph with
copy that says what it does.

### Two ways a wall could go dark while the app buzzed success — both closed

- **A route can collapse to nothing.** MoonBoard FOOT/AUX have no canonical
  code, an unrecognised role code decodes to the `{id}={code}` sentinel, and an
  unknown board name has no code table at all. An empty result would have
  reached the clear-all path and returned `true`, driving the auto-sender's
  success path over a dark wall. It now refuses with the existing
  `incompatible_climb` shape.
- **Aurora v2's power ladder can zero every colour channel.** `computeV2Scale`
  walks 1.0 → 0.05 against an 18 W budget and `scaledColorV2` is
  `floor(value * scale) >> 6`, so any rung at or below 0.25 turns a saturated
  255 channel into 0 — and the 0.2 rung's computed power is 0, so it "fits" for
  any climb however dense. Past roughly 135 lit holds on a Kilter box (2 LEDs
  per hold) the board receives a full, well-formed packet of colourless LEDs.
  `getAuroraBluetoothPacket` now reports `allLedsDark` and the sender refuses.
  **This was already reachable for dense single-frame climbs**; the union makes
  routes dense enough to reach it too, so it is fixed here rather than left.
  Scoped to v2 on purpose: on v3 the only route to a zero colour byte is a
  caller-supplied near-black `LedColorOverrides` value, which is the caller's own
  choice, not the protocol overruling them. The refusal gets its own copy
  (`ble.errorTooManyHolds`) rather than the generic "wrong board" line, and a
  successful v2 send now reports `ledsWritten` / `ledsWithColour` so the
  partially-dark residual the all-or-nothing detector lets through can be sized
  from field data before anyone picks a threshold. With the default palettes that
  gap should always be zero — every canonical role colour keeps a lit channel at
  every rung above the cliff — so a non-zero reading means custom colour
  overrides.

### iOS lock screen writes the same bytes

`use-live-activity.ts` ships the queue snapshot's raw `frames` to native, and
Swift's `parseFrames` (`BoardBleEncoding.swift`) is the same naive
`split("p")`/`split("r")`. Without flattening there, the same climb on the same
wall would show the union in-app and the garble from a lock-screen Prev/Next or
a reconnect re-light — and the JS dedup would then confirm over a wall it never
wrote. Flattened at the snapshot build site; **no Swift change**, because Swift
parses a flat string byte-correctly (and its `mirroredFrames` stops dropping
pairs as a bonus).

**Author verification**

The gate for this branch, run from the worktree:

- `vp run typecheck` (whole repo) and `vp run typecheck:mobile`
- `vp run test:mobile`
- `vp test run --project i18n` — catalogue parity for the new keys
- `vp check` and `vp run check:i18n:orphans`
- Pre-commit hook on every commit; no `--no-verify`.

New tests:

- `mobile/create-climb/__tests__/create-route-playback-slot.test.tsx` — the
  slot's three states: nothing on a board that cannot hold a route, the strip
  (hint + a11y label + 44 dp Add frame pill that calls back) at one frame, the
  transport under its own gesture root at two.

- `ble-protocol/__tests__/aurora.test.ts` — the regression oracle runs the
  **real** packet builder: a raw additive route yields `skippedRoleCount === 1`,
  the collapsed union yields `0` skips and all four holds; a raw route with an
  `x` token lights the hold it cleared; a single-frame climb is byte-identical.
  Plus `allLedsDark` across v3, a light v2 climb, a 160-hold v2 climb, the
  clear-all packet, a near-black v3 colour override (must stay writable), and the
  full chain with no mocks anywhere: a real 160-hold two-frame route through the
  real `toFlatFrames` into the real builder at API level 2.
- `board-constants/__tests__/hold-states.test.ts` — collapse output carries no
  `,` / `"` / `x`; MoonBoard foot/aux-only route → `''`; preview roles 47/48 →
  43; the frame-boundary hold survives an additive route.
- `mobile/lib/ble/__tests__/use-board-bluetooth.test.ts` — the builder receives
  the union not the raw route; a flat string is byte-identical; `''` still
  clears; mirroring gets the flat union (no `NaN`); a collapsed route and an
  all-dark v2 packet both refuse without writing.
- `mobile/lib/ble/__tests__/use-ble-frame-writer.test.ts` — burst collapse,
  dedup, reset-key re-flush, drain survives a failed write, and an emptied frame
  (`''`) reaches the wall instead of taking the `null` stand-down (mutation-checked).
- `mobile/lib/live-activity/__tests__/use-live-activity.test.tsx` — the native
  snapshot carries the union; single-frame unchanged.
- `mobile/create-climb/__tests__/use-create-climb-playback.test.ts` — animatable
  gate, flat frame string, clock drives the editor cursor while playing, editor
  drives the clock while paused, seek moves both, play on the last frame
  replays, a paint stroke stops the clock.
- `mobile/create-climb/__tests__/use-create-climb-screen-playback.test.tsx` —
  the hand-off contract: Set Active pauses and hands over, a draft save does the
  same, the preview stands down, paint/duplicate/delete/play/undo/redo take the
  wall back, and every creator write invalidates the auto-sender's wall-state
  record.

**Not verified:** no emulator or simulator run (the batch's single Android
emulator is held by another session), and no physical BLE board — see the BLE
review section for the three hardware checks that want a device pass. The
device-QA script also wants: a 667 dp phone with a 5-frame route (the peek must
still leave a strip of the below-fold form visible and dragging up must still
travel), OS-maximum text size (counter must not wrap, pill must not clip), the
Android speed-slider gesture inside the Compose sheet, and TalkBack/VoiceOver
across the five transport stops.

## Release Notes

- Routes now light up properly on the wall — the whole route, not a scrambled mix of moves.
- Play your route in the climb creator before you publish it, with the same transport and speed control the play drawer has.
- The creator's play button was never a play button. It's now a queue glyph that says "Put it on the wall", and it tells you why it's greyed out.
- The creator now shows a route strip under the board from the first frame, so you can see how to build and play a route.

## What changed, by area

**BLE correctness**
- `use-board-bluetooth.ts` — `toFlatFrames` at the `sendFramesToBoard` choke
  point; both `frames === ''` clear tests stay on the caller's original argument
  so a deliberate clear is still a clear; mirroring runs on the collapsed string
  (it is a third `split('p')` tokeniser and used to emit `p<id>rNaN` or throw).
- `aurora.ts` — additive `allLedsDark` diagnostic on `BluetoothPacketResult`.
  No byte changes.
- `connection-error.ts` — `power_budget_dark` added to `BleSendFailureReason`.
- `use-live-activity.ts` — flatten the frames in the Live Activity queue
  snapshot.

**Creator playback**
- New `use-ble-frame-writer.ts` — the latest-wins drain lifted verbatim out of
  `use-mobile-playback.ts`, now shared by both surfaces. The existing
  `use-mobile-playback` suite is the guard and passed unmodified. Standing the
  writer down now also clears any frame queued behind an in-flight write: without
  that, the drain wakes after the hand-off and writes a stale playback frame
  *over* the auto-sender's union — the wall lights the route, then flips back to
  one frame under the "On the wall" chip, after presence was told the climb is
  lit. On slow boxes (v2 with-response, MoonBoard) "write in flight plus a queued
  tick" is the common state at the 750 ms pace, so this was near-deterministic
  rather than a tight race. It heals the same pre-existing hole on the play
  drawer's own stand-downs (close / suppress) for free.
- New `use-create-climb-playback.ts` — composes `usePlaybackEngine` over the
  editor's live frames. Deliberately not `useMobilePlayback`: that one is bound
  to a saved `Climb` and the party-sync seam, and a WIP must never be broadcast
  to the crew.
- `PlaybackControls` + `playback-speed-report` moved from `play-drawer/` to
  `components/playback/` — two surfaces own it now.
- `bluetooth-provider.tsx` — new `invalidateWallState()`. The creator writes
  straight past the auto-sender, so its record of what the wall physically shows
  has to be dropped, or re-selecting the previously-lit queue item hits the
  dedup skip and confirms a climb over a wall showing the creator's frame 3.
- `use-create-climb-screen.ts` — `holdsJson` now derives from **frame 0**, not
  the active frame. It is an autosave dep, and playback moves the active frame
  every tick: at the 500 ms debounce under a 750 ms pace that was one draft
  write per frame. Frame 0 is also what the legacy single-frame restore
  fallback actually wants.

**The transport owns time, the action bar owns editing.** The ⏮ / counter / ⏭
cluster moves out of the horizontally-scrolling action row into the transport
card; Duplicate and Delete frame stay. That also fixes a defect found on device:
the scroller could not show all its controls at once, and right after adding a
second frame only ⏮ was visible with the counter clipped to ~5 px and no scroll
hint.

**The route slot is never empty (after the QA decline).** The transport mounted
only at two frames, and the only way to reach two frames was a bare `copy` glyph
fourth inside the action bar's scroller — so a fresh create sheet showed no play
control and nothing saying routes existed. New `CreateRoutePlaybackSlot.tsx` owns
that place in the layout: nothing on Woods, an inert strip at one frame (greyed
play glyph, a line saying what a frame buys you, an "Add frame" pill in the
transport's own card chrome), and the unchanged transport at two — `PlaybackControls`
takes no new prop, so PlayDrawer's mount is byte-identical. The board budget
follows: `ROUTE_STRIP_RESERVE = 52` for the strip, `PLAYBACK_TRANSPORT_RESERVE`
= 84 + 52 for the transport plus its frame-actions row, neither on a
single-frame board. Frame editing has ONE home: a labelled Add frame / Delete
frame row under the transport, so the third frame is as reachable as the second.
Both glyphs leave the action bar's scroller entirely.

## Design directives

All of D1–D13 from the design review are implemented:

- **D1** Set-Active uses `queue`, not `flash` (`flash` is the flashed-ascent
  glyph) and not `lightbulb` (the header BLE toggle in the same sheet).
- **D2** The route slot's reserve comes out of `boardMaxHeight` only on a board
  that can hold a route — `PLAYBACK_TRANSPORT_RESERVE = 84` at 2+ frames,
  `ROUTE_STRIP_RESERVE = 52` at one, nothing on Woods — plus a
  `0.92 × (windowHeight - insets.top)` clamp on `peekHeight` so the peek and
  `'100%'` snaps can never collapse into one. The slider's extra 36 dp is
  deliberately **not** reserved.
- **D3** `setActive` copy reuses the four `playView.wallState.putOnWall` values
  already shipping in every locale — which also retires the es *vía* / fr *voie*
  error (a board problem is not a sport route). New `setActiveHint` explains the
  `isValid` gate on the disabled button.
- **D4** New `handedOff` state. Set-Active pauses the transport and stands the
  creator's writers down so the queue's union stays on the wall; the transport's
  counter is replaced by an "On the wall" chip, because "2 / 3" over a fully-lit
  wall is worse than no counter. Any paint / duplicate / delete / play / seek
  takes the wall back. One new optional prop (`wallStateLabel`) — the play
  drawer's render is unchanged. Extended one step past the directive: a **draft
  save** also pushes the climb into the queue and leaves the drawer open, so it
  hands the wall over the same way.
- **D5** Speed pill `minHeight: 28` → 44 dp with `hitSlop` in both type scales.
- **D6** The speed slider becomes `accessibilityRole="adjustable"` with
  value/min/max and 0.5× increment/decrement, its children hidden from Android's
  accessibility tree; the pill gains a hint so the long-press reveal is
  announced at all.
- **D7** Slider thumb gains a 2 dp `brandColors.primary` ring — white on the
  track was ≈1.09:1, failing WCAG 1.4.11.
- **D8** Frame counter gains `Frame {{index}} of {{total}}` and
  `numberOfLines={1}`.
- **D9** Transport renders inside the measured `onLayout` View (so it is visible
  at the peek) wrapped in a nested `GestureHandlerRootView` with an explicit
  `alignSelf: 'stretch'` — load-bearing on Android, where this sheet's Compose
  `ModalBottomSheet` is outside the app's single root and RNGH defaults to
  `flex: 1`.
- **D11** Frame stepping moves to the transport chevrons.
- **D12** es "cuadro" → "fotograma" on the two surviving frame buttons, so one
  screen no longer uses two words for "frame".
- **D13** The tuned speed is **preview-only**: `handleSave` writes
  `framesPace: 0` and `useClimbFrames` falls back to `DEFAULT_PACE_MS`, so 1× in
  the creator is honestly the shipped speed. Nothing here authors the pace.
- **D14** (persisting the tuned pace) is **not** done — it needs sign-off and
  overlaps #4633.

## Deviations

- **`failureReason` for the v2 dark-wall refusal is `power_budget_dark`, not
  `incompatible_climb`.** The BLE review asked for the `incompatible_climb`
  *refusal shape* so the field tells us how often it fires. Reusing the existing
  string would have made the two causes indistinguishable in PostHog, defeating
  that purpose, so the refusal shape is identical (warn + the same alert copy +
  `ClimbSentToBoardFailure` + `return false`) but the reason is its own allowlist
  member.
- **Fable's nit 4634-5** (calling `invalidateWallState` from the play drawer's
  frame writer too) is deliberately **not** taken here: the play-drawer
  extraction is guarded by its existing suite on the promise of no behaviour
  change. It is a pre-existing hole and belongs in its own PR.
- **`resolve.dedupe: ['react']` added to `packages/mobile/vite.config.ts`.**
  Several `@boardsesh/*-react` packages carry their own React devDependency
  (`^19.2.8`) while mobile pins `19.2.3`, so bun materialises a nested copy under
  each; without deduping, a mobile suite that renders a hook from one of them
  loads a second React and every hook call throws. Needed for the new no-mock
  suites that render the real `usePlaybackEngine`. This mirrors production rather
  than diverging from it — `metro.config.js` already forces the same singletons
  via `SINGLETON_MODULES` — and the two are now cross-referenced in both files.

## BLE review

**This PR changes the bytes written to physical walls and carries the
`ble-needs-fable-review` label.** Three things for the reviewer, all verified:

1. **Aurora: nonsense → coherent.** Today a multi-frame route lights whatever
   survives `split('p')` mangling, including holds a later frame turned off, and
   counts the rest as `skippedRoleCount`. After: the union of every frame, no
   skips.
2. **MoonBoard: some dark holds will now LIGHT.** `MOONBOARD_ROLE_MAP` knows
   only 42/43/44, so a raw route's codes 45–48 are skipped and never lit today.
   After the collapse, **47 and 48 re-emit as canonical 43** (both are `HAND`)
   and will light. **45 (FOOT) and 46 (AUX) stay dark** — no canonical code, so
   the collapse drops them, matching today's outcome by a different route. This
   is holds appearing, not holds disappearing.
3. **Kilter Tycho (product 6) colour loss.** `toFlatFrames` re-emits through
   `STATE_TO_PRIMARY_CODE`, which maps Kilter `HAND → 43`, so a multi-frame Tycho
   climb comes out one colour instead of six. Bounded: flat strings return
   untouched, and multi-frame Tycho already loses those colours through the
   existing playback path. Strictly better than today's garbage packet.

**Worst case, computed.** The largest LED map in the catalogue is 690 placements
(layout-size `'10-10'`). A circuit lighting every hold is the bound: v3 → 2,124
bytes / 9 protocol messages / 9–107 BLE chunks (~0.6–1.5 s without-response);
v2 bare-name box → 1,416 bytes / 71 with-response chunks (~3.5–7 s), and
guaranteed all-dark until the `allLedsDark` refusal above. MoonBoard bound: 198
holds ≈ 1.2 KB text frame. Message count is unbounded by design
(FIRST/MIDDLE/LAST/ONLY framing) and the builder splits bodies at 255 B, so no
size cap is exceeded.

**Revert unit: the whole PR.** The BLE work spans three commits
(`3b309a567`, `e912a47d4`, `a3b40016a`) rather than sitting in one, so the
bytes-on-walls half cannot be reverted without also taking the creator feature
down. That is a consequence of shipping both halves together; whoever merges
should know the rollback costs the feature too. Next time the wall-bytes half
lands alone first.

**Rollback is clean and fast (OTA).** JS only, `fingerprint` runtime policy. No
persisted state serves the union after a revert: `toFlatFrames` runs per send,
the dedup refs are in-memory, and the draft store and queue persist **raw**
frames. The last Live Activity snapshot holds flat strings until the next queue
update, and flat strings are valid input to every version of the encoder.

**Not verified on hardware.** No physical board was available. Three things want
a real-device pass: a dense union on a **bare-name (v2) Kilter box** — expect the
refusal, not a dark wall; a ~1.2 KB union frame on an **official Moon
controller** (line-buffer tolerance is unverified in-repo; community
ArduinoMoonBoardLED parses incrementally and is fine); and lock-screen Prev/Next
on **iOS** for the Live Activity path.

## Notes for follow-up

- **#4633 is smaller than it reads.** The long-press slider already spans
  0.1×–10×; it is the tap-to-cycle pill that stops at 0.5×, and nothing announced
  the long-press (this PR adds that hint). What remains genuinely unbuilt is the
  seconds-per-frame re-spec and the 0–30 s start delay. Because both surfaces
  now mount the same `PlaybackControls`, #4633 is a one-component change instead
  of a two-surface one.
- The creator still saves `frames_pace: 0`, so a route tuned to 3× publishes at
  the 750 ms default (D14 / #4633).
- Mobile climb thumbnails still render frame 0 where web renders the union
  (#3988). This PR moves the **wall** onto the union, which narrows #3988 to the
  thumbnail surface but does not close it.

## Test plan

1. Needs a Kilter or Tension board over Bluetooth, signed in.
2. Climbs tab → + → see the greyed ▶ strip under the board.
3. Tap Add frame twice, paint holds → transport reads 3 / 3.
4. Tap play → the wall steps through all three frames.
5. Tap the queue icon → whole route lights, chip reads "On the wall".

## Risk

Risk: 5/5 — changes the bytes written to physical walls on every send.


marcodejongh (commented):
## Fable BLE review — bytes-on-walls half of #4761

Reviewed the BLE diff against its base (`fac0aa52f`) and re-read the encoders on `main` rather than the PR text. Re-verified after the rebase at `effcdec5d` (BLE files byte-identical to the reviewed diff apart from the one change noted below).

**Defect is real, fix is at the right seam.** `aurora.ts` parses each token's role with `Number(role)`, so the hold sitting before every `,"` frame boundary gets role `NaN` and is skipped, and an `x<id>` off-token glues onto the previous token so the hold a later frame turned off stays lit. `moonboard.ts` is the same `split('p')`/`split('r')` with a 42/43/44 map. Collapsing to the union with `toFlatFrames` at `sendFramesToBoard` covers every JS writer at once; the fast path (`!includes(',') && !includes('x')`) returns the string untouched, so single-frame climbs are byte-identical on every board.

**Clear semantics survive.** Both `frames === ''` checks still read the caller's original string. The new refusal only fires when a non-empty route collapses to `''` (MoonBoard foot/aux-only, unknown role codes, unknown board) and returns `false` with `incompatible_climb`, so the auto-sender no longer buzzes success over a wall it just cleared. Mirroring now runs on the collapsed string, which is right: `convertToMirroredFramesString` is a third `split('p')` tokeniser.

**`allLedsDark` checks out.** `scaledColorV2` is `floor(value * scale) >> 6`, so any ladder rung at or below 0.25 zeroes a saturated channel, and the 0.2 rung always "fits" the 18 W budget. Scoped to v2 (`isV2 && ledsWithColour === 0`); the `0xfc` mask matches the v2 byte `(r<<6)|(g<<4)|(b<<2)|posHi`, and v3's colour is the third byte of `[...encodePositionV3, encodeColorV3]`. Keeping `power_budget_dark` as its own reason string is the right call — folding it into `incompatible_climb` would make the two indistinguishable in PostHog.

**Frame writer.** The drain is the play drawer's code moved verbatim; standing down now clears the pending frame, which closes the stale-frame-after-hand-off race on slow (v2 with-response / MoonBoard) boxes. One change was required and is in this rebase: the stand-down test is `frame === null`, not falsy — a deliberately empty frame (`''`, a route frame with every hold erased) must write the clear packet rather than leave the previous frame lit.

**Live Activity.** Flattened at the snapshot with `toBoardName` validation; a route that collapses to `''` keeps its raw string so native's own zero-encoded refusal handles it instead of a clear-all. Swift untouched, and it parses a flat string correctly.

**Accepted consequences (say so in release notes if asked):** MoonBoard preview codes 47/48 now light as HAND (43) where they were silently skipped; 45/46 stay dark. Multi-frame Kilter Tycho climbs come out one colour per hold — already true on the playback path.

**Non-blocking nit:** `toFlatFrames(frames, boardName as BoardName)` in `use-board-bluetooth.ts` — prefer `toBoardName(boardName)` with a null guard, matching `use-live-activity.ts`. No functional risk today because flat strings bypass the table entirely.

**Not verified on hardware by me either.** Before merge the QA tester should run, from the PR test plan: a dense (>130-hold) route on a bare-name Kilter box → expect the "too many holds" refusal, not a dark wall; and one lock-screen Prev/Next on a route on iOS. Rollback is a plain OTA revert; nothing persisted depends on the union.

**Verdict: approve the BLE half.**

Issue #952, opened 2026-03-29
Top search bar overflows when too many filters are active


marcodejongh:
<img width="603" height="1311" alt="Image" src="https://github.com/user-attachments/assets/f0fc83f0-0619-4764-b0f8-8b65e0e2cdae" />

We can group into "hidden x,y" instead of "hidden x, hidden y". In addition to that we need to fix the max width on it and just show a "..." when theres too much to show

Make sure the text in the bar and the quick search history stays consistent

![Image](https://github.com/user-attachments/assets/e861c446-c287-477a-84fa-c90cfd59ac2a)

marcodejongh:
Thanks!

PR #4857, opened 2026-08-30
feat(shared): per-image board-art tracing, exact EDT, r-derived trim radii

marcodejongh:
## Summary

The board-art tracer was cutting holds apart at boundaries that do not exist on the wall.

Every board's art ships as a stack of transparent per-set layers, and the tracer flattened
them into one bitmap before deciding where each hold ends. Two holds from different sets
are bolted into different holes and their art overlaps by 0.06% of opaque pixels across the
whole catalogue — but stacked into one image they touch, and "touching" is what drove the
neck trim and the contact pullback. On Kilter Homewall 12x12, **439 of 499 placements are
art-adjacent to a differently-labelled hold on the composite; 64 are when each layer is
measured on its own.** The other 375 were being chopped for nothing, and 78 of them ended
up keeping under 80% of their own art body.

This PR traces each placement against the one layer that draws it, replaces the partition's
chamfer with an exact Euclidean transform, and sizes the neck trim off the placement radius
instead of the board's pixel width. Every shard regenerates; the contract is unchanged.

**No renderer change and no new field.** `outlines`, `silhouetteLightness` and `ledBright`
keep their names, units and meaning — the polygons behind them are just the right shape now.

### The commits, in order

| Commit | What |
|---|---|
| `bfef4d5` | Per-hold trace metrics + a `--report` mode that draws every silhouette on the art. Byte-identical output. |
| `486443b` | Mask-provider seam: layers decoded once, `traceOutlines` takes a `TraceField` and knows nothing about alpha, sharp or files. Byte-identical output. |
| `ec90cf5` | Exact separable Felzenszwalb–Huttenlocher EDT carrying labels, replacing the chamfer. |
| `88613e3` | One trace field per art layer. Aurora inverts `images_to_holds`; MoonBoard routes through `MOONBOARD_CELL_SETS`. |
| `ddb0610` | Trim and clearance radii become `max(2, round(0.078 · r))`. |
| `cf5abb2` | Gates reworked per-image, gate 7 added, all pins re-derived. |
| `8475bfd` | Docs. |
| (this push) | Review fixes: gate-7 denominator, gate-6 probe scaling, re-harvested pins. |

The first two commits are mechanical and were verified byte-for-byte against the drift gate,
so the diff that actually moves data is the three after them.

### Kilter Homewall 12x12 (`kilter/8-25`), before → after

Both columns are produced by the **committed gate code**, run once against the shards at the
PR base and once against the shards this PR ships. Same measurement both sides.

| Metric | Before | After | Target |
|---|---:|---:|---:|
| gate 6 `opaqueMean` (boundary with art behind it — the chop metric) | 17.3% | **0.6%** | < 5% |
| holds losing > 20 board px² to trim + pullback | 376 | **33** | ≤ ~64 |
| pulled back off a neighbour | 369 | **33** | ~64 |
| holds keeping < 0.8 of their own art body | 78 | **0** | — |
| area recovery, mean | 0.897 | **0.981** | — |
| gate 6 `overFivePercent` | 0 | **0** | 0 |

Catalogue-wide, same two runs: `opaqueMean` **fell on 41 of 49 shards and rose on none**;
`chopped` fell on 33 and is now 0 on 45 of 49.

### Determinism and the two invariants

- Full regeneration run **twice** at each data-moving commit; the second run wrote 0 files
  every time. The EDT uses integer squared distances and exact integer cross-multiplication
  in the parabola envelope, so no comparison is ever decided by floating-point rounding;
  ties resolve lower column, then lower row, then lower placement index. Checked against a
  brute-force nearest-seed scan over 60 random boards — 0 distance mismatches.
- **`wall-lightness.cjs` is byte-untouched**, and so is **every `ledBright` entry** in all 49
  shards. Both verified by diff against the PR base. Colour readings still measure the
  composite, because what a mark has to be legible against is the stack a climber sees.

### Two placements honestly lost their outline

`decoy/2-1`'s 935 and 950 — frame-rail T-nuts whose art on their own layer runs into the
search box, so the existing box backstop rejects the trace and the renderer falls back to a
ring at the placement radius. That is the correct answer for them; nothing else in the
catalogue changed traced state. Coverage goes 15,501 → 15,499 of 15,784.

### The one place this is not a strict win

Gate 6's `overFivePercent` rises on **three** shards by one or two outlines each
(`grasshopper/1-4` 0 → 1, `tension/10-6` 0 → 2, `tension/11-10` 0 → 1). Those are boards
whose cut clearance narrowed from 3 board px to 2 under the new radius rule, so the boundary
sits marginally closer to a same-set neighbour.

An earlier draft reported this as nine shards and up to 32 outlines. That was the *measure*:
the gate probed a flat 2.5 board px outward, which asks a shard guaranteeing 2 px of
clearance whether it has art half a pixel **beyond** its own guarantee. Sweeping
`touchstone/1-1` on unchanged geometry gives 6 outlines over 5% at a probe of 2.0, 32 at
2.5 and 69 at 3.0. The probe now scales with each shard's clearance, and the residue is the
three above.

**Clearance revisit deferred** — measured, and it makes things worse: decoupling the
clearance to `0.094 · r` buys back `overFivePercent` but degrades `opaqueMean`, `chopped` and
`recoveryP10`, which are the metrics this rework exists for. The smaller lever if we want it
later is `clearance = max(3, round(0.078 · r))`, now that the gate fixes have landed.

Related: `ddb0610` deleted `CUT_CLEARANCE_AT_REFERENCE`, which carried a real measurement —
"at a clearance of 2, shoulder ink on a neighbour is 731 board px² across the spike's seven
boards against 25 at 3". That number was taken **on the composite**, where cross-set holds
appear to touch, so most of the 731 px² was ink landing on a neighbour that is not a
neighbour once each set is traced alone. It is superseded rather than refuted, and it is
preserved verbatim in the comment on the new coefficient so nobody re-derives the clearance
without re-measuring.

### Gates

Seven now, not six. Gate 6 is rescoped per-image and its `opaqueMean` half became an enforced
ratcheting chop budget. Gate 7 is new and is the only one that catches a hold which simply
lost half of itself — gate 3 clears a chopped silhouette, gate 5's open clears it, and gate 6
positively likes it, because a boundary well inside the hold's own art is exactly what a
pullback is supposed to produce. Its fixture shows precisely that: a silhouette cut back to
half its slab, with gate 6 reading zero on it.

Gate 5's known-failure pin (`tension/11-10` #952) is **empty** — that outline was the
board-width radius rule showing its edge, and the rule is gone. Gate 1's pin drops 5 → 3;
two of the five were the composite's cut running between a bolt and its own hold.

Every measurement still **restates** the generator's constants and its placement→image
routing rather than importing them. A gate that shares its inputs with the code it audits
only checks that the code agrees with itself.

### Eyeballing it

Before/after contact sheets for all 49 configs — the composited board art with every
silhouette stroked on it, amber where the tracer pulled back off a neighbour, red where it
kept under 80% of the hold, dashed grey where it fell back to a ring — are in the worktree at
`.boardsesh/art-report/before/` and `.boardsesh/art-report/after/` (gitignored, ~54 MB each).
`vp run generate:board-art-geometry -- --report=<dir>` regenerates them and touches no
committed file. `summary-8-25-touchstone.md` in the same directory is inlined above.

## Release Notes

none

## Test plan

1. CI green.
2. Board-art drift gate green (byte-deterministic regeneration).
3. Gates pass under `BOARD_ART_GATES=all`.

## Risk

Risk: 2/5 — offline tooling + regenerated data


packages/backend/src/handlers/native-auth.ts
marcodejongh, line 591: Where did this come from? Looks like its from another PR possibly? Maybe we brought this inf rom origin main?

PR #4960, opened 2026-08-31
feat(mobile): ask which board look to use, with live previews

marcodejongh:
## Summary

2.4 makes the **Boardsesh** drawing the default board renderer. Rather than changing how
every climber's board looks without asking, the app now asks once — with a horizontal
carousel of live renders of **their own board**, so the choice is made on what it actually
looks like rather than on a label.

The picker is a step inside the existing `/onboarding` route, **not a new modal**. The app
already has three launch-time prompts (first-run tour, QA tester, offline nudge); a fourth
would be one too many. Climbers who already finished the tour get this step on its own.

The same carousel then replaces the flat preset chip row on **More → Board look**, so the
presets are visible there instead of being four bare words.

### What changed

- **Retired the `board-render-mode-default` flag.** It shipped at 0% and existed to roll the
  mode out; 2.4 makes Boardsesh the app default outright, so `requestedBoardRenderMode` now
  answers `boardsesh` for a stored `mode: 'default'` with no flag in the loop.
  `board-glow-falloff` (a live A/B) stays. An older binary that cannot draw the mode is still
  pinned to Classic by the capability probe — that check is now the only thing standing
  between a 2.3 build and a mode it cannot draw, so it is pinned by a test.
- **Fixed a live accessibility bug.** `applyCvdPalette('monochrome')` turns `roleGlyphs` on,
  because a colour-only cue is meaningless once every role is a shade of grey. Applying any
  preset rewrote all twelve fields verbatim and switched those glyphs back off — a colour-only
  board for the one climber who cannot use colour. Preset application now only ever *raises*
  accessibility-owned fields. Worth reviewing on its own: it is a bug on `main` today.
- **`renderSettingsOverride` on `useNativeClimbRender`**, so a preview card can draw a preset
  the climber is not on. Substituted above the mode/veil/signature chain so it reaches the
  cache key: each preset caches as its own PNG, and a card matching the climber's real
  settings *shares* the real PNG rather than displacing it. Hold colours still come from the
  global override store, so previews are always in the climber's own palette.
- **One shared carousel** (`BoardLookCarousel`) behind both surfaces, over one pure options
  module, so the two can never drift on what the options are or what picking one does.

### Notable decisions

- **Skipping is a real answer.** It marks the step seen, leaves `mode: 'default'` (the new
  default), and never asks again. Read `outcome: 'skipped'` as "accepted the default", not as
  a funnel drop.
- **Cards skeleton rather than lie.** While the capability probe is unanswered, a Boardsesh
  card would resolve to a *classic* render under a Boardsesh label. Those cards render a frame
  with no board image until the probe answers; if it answers `false` the step never shows, and
  the seen flag is deliberately **not** written, so the climber is asked after they update.
- **Fresh installs are asked after they pick a board.** There is no board to preview during
  the first-run tour — the tour's whole job is to get them to pick one. The step is held
  (`wait`, not `none`) until a board is bound, and the route guard keeps it off the picker.

Rejected, worth a line: moving `roleGlyphs` into the `holdColorOverrides` store is the right
long-term fix — preset writes structurally could not reach it — but it needs a two-key storage
migration plus a change to the render signature's token set. The one-directional merge is the
same behaviour with none of the migration risk.

## Release Notes

- Pick how your board looks — Boardsesh, Subtle, High contrast, or the classic markers — with a live preview of each on your own wall.
- Lit holds now glow in their own shape by default. Not for you? Switch back to the classic markers any time under More → Board look.
- Board look now previews every preset on your board, in your own hold colours.

## Test plan

1. Open the app. You see "Pick your board look".
2. Swipe the row. Each card shows your board.
3. Tap "Boardsesh Subtle", then "Use this look".
4. Reopen the app. The question does not return.
5. More → Board look. Same row, applies instantly.

## Risk

Risk: 3/5 — a launch-time interruption plus a change to the default renderer for everyone who
never chose. Bounded: the per-climber setting still wins, the capability probe still forces
Classic on a build that cannot draw the mode, and the step self-suppresses there.

The diff itself is JS/JSON only and touches no fingerprint input, but the branch is based on
`release/next`, whose native renderer artifacts already differ from `main` — so `ota-check`
correctly reports a native change and this ships with the 2.4 store build rather than by OTA.
That is the intent: the default flip, the flag retirement and the step that explains them all
land in the same binary.

Two behaviour changes beyond the step itself, both worth a look at review:

- Applying a preset no longer clears a colour-blind climber's role glyphs (a bug on `main`).
- `ensureBoardseshSupportProbed` now runs for every install rather than the 0% flag cohort —
  two renders once per JS lifetime. It also latches `false` when the native module loader has
  permanently given up, without which `Climb View Opened` would go silent for the session on a
  build with no renderer.


docs/feature-flags.md
```diff
@@ -150,24 +150,29 @@ value)` accepts `boolean | string`, and the Feature Flags screen renders a
...
-### The two board-render flags (issue #2202)
-
-- **`board-render-mode-default`** — `variants: ['classic', 'boardsesh']`.
-  Which drawing a climber who has never chosen a mode themselves gets. A
-  climber's own Settings choice always wins over this (see
-  `resolveEffectiveRenderSettings` in
-  `packages/mobile/src/lib/board-render-settings.ts`). **Unresolved reads as
-  `classic`** — the shipped drawing, so a flag outage or a build that predates
-  the flag draws exactly what it always drew.
+### The board-render flag (issue #2202)
```
marcodejongh: Lets delete all the board render related flags. We'll go live in the 2.4 release and make it user configurable

packages/mobile/app/onboarding.tsx
```diff
@@ -1,13 +1,27 @@
...
 import { markOnboardingSeen } from '../src/lib/onboarding/onboarding-storage';
+import { useBoardPreviewClimb } from '../src/hooks/use-board-preview-climb';
+import { useEffectiveBoardRenderSettings } from '../src/hooks/use-native-climb-render';
 import { reportError } from '../src/lib/error-reporting';
 
 /**
+ * The onboarding route, which hosts two independent steps.
+ *
+ * `/onboarding` is the first-run framing screen. `/onboarding?step=board-look`
+ * is the one-time "pick your board look" question that 2.4 asks when the
```
marcodejongh: We should make sure users always pick a board, and pick a renderer, and that we dont allow them to skip. If their account already has baords configure, we could show their existing boards in a horizontal slider so they can choose

packages/mobile/src/components/board-look/BoardLookPreviewCard.tsx
```diff
@@ -0,0 +1,147 @@
...
+};
+
+/**
+ * One board look, drawn on the climber's own board.
+ *
+ * Memoized, and `onPress` takes the option id, so the carousel's `renderItem`
+ * needs no per-card closure and a selection change re-renders only the two
+ * cards whose `selected` actually flipped.
+ */
+export const BoardLookPreviewCard = React.memo(function BoardLookPreviewCard({
```
marcodejongh, line 44: Make sure we reuse as much as possible from the regular board picker slider

packages/mobile/src/components/board-look/BoardLookStep.tsx
```diff
@@ -0,0 +1,292 @@
...
+  useEffect(() => {
+    // Marked seen HERE, beside the Shown event, rather than when the route
+    // mounts: the flag means "this climber has been asked", and the decision
+    // module is explicit that a climber with nothing to preview — or on a build
+    // that cannot draw the mode — must NOT have it burned. Skipping still
+    // counts, and so does a force-quit mid-step, which is why it is written on
+    // arrival rather than on an answer. Fire-and-forget for the same reason
+    // `markOnboardingSeen` is: a keychain failure must not strand the climber,
+    // but it must be reported, since swallowing it re-asks every cold start.
+    markTipSeen(BOARD_LOOK_STEP_SEEN_KEY).catch((error: unknown) => {
```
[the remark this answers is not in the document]
marcodejongh: Fixed in c12a6435f. The marker now lives in AsyncStorage (`boardLookStepSeen`) beside `boardRenderSettings`, so the two share a lifecycle — an uninstall takes both. Covered by two cases in `replay-board-look-step.test.ts`: the marker is in AsyncStorage, and clearing the sandbox leaves `hasSeenBoardLookStep()` false with the mode back at `default`.

packages/mobile/src/components/board-look/BoardLookCarousel.tsx
```diff
@@ -0,0 +1,142 @@
...
+  const handleViewableItemsChanged = useCallback(({ viewableItems }: { viewableItems: { key: string }[] }) => {
+    const report = onCardSeenRef.current;
+    if (!report) return;
+    for (const entry of viewableItems) report(entry.key as BoardLookOptionId);
+  }, []);
+
+  const extraData = useMemo(
+    () => ({ selectedId, boardseshRendererAvailable }),
+    [selectedId, boardseshRendererAvailable],
+  );
```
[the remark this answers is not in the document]
marcodejongh: Fixed in c12a6435f. `extraData` now carries `previewSettingsById` alongside `selectedId` and the renderer latch, so a change to the live settings invalidates the recycled rows.

docs/feature-flags.md
marcodejongh: Done in ee1d4581a — `board-glow-falloff` is gone too, so no board-render flag remains. A `default` falloff resolves to `soft` and `glowFalloffSource` narrows to `user | default`; `Climb View Opened` drops its `$feature_flag*` exposure properties with the experiment.

Worth flagging: it was the last multivariate flag, so per your call the machinery went with it — `useFeatureFlagVariant`, the variant row rendering, `variantFlagOptions`, the `variants` field on a definition, and the variant branch of the PostHog read. The catalog is boolean-only now, and a future multivariate flag rebuilds that from git history. Larger than the "~20 tests" I estimated when I asked; say the word if you would rather keep the machinery.

The capability probe is now the only thing forcing `classic` on a binary that cannot draw the mode, so it has its own test.

packages/mobile/app/onboarding.tsx
marcodejongh: Captured as #4961 rather than built here, per your "follow-up PR" call — no board and no render mode without an answer, and no exit on either step.

One thing I put in the issue: "no exit" is only safe because `decideBoardLookStep` already refuses to show the step unless there is an active board with a synced climb AND the renderer probe answered `true`. By the time it is on screen its previews are known to render. If that gate is ever relaxed the exit has to come back, or a climber gets stranded. The board step needs the same property — a failed board list should fall through to the full picker, not a dead end.

packages/mobile/src/components/board-look/BoardLookPreviewCard.tsx
marcodejongh: Noted on #4961 as part of the same work — the board-pick step reuses `BoardCarousel` + `BoardDiscoveryCard` rather than growing a second slider.

For this card specifically: it shares the FlashList config with `BoardCarousel` (snap interval, alignment, deceleration, hoisted `keyExtractor`) but not the card body, because a discovery card shows board metadata while this one is a render of a climb under different settings. Happy to extract the shared frame when the two sit side by side in the onboarding flow and it is clear what actually overlaps.

PR #5107, opened 2026-09-02
fix(mobile): stop the create-climb sheet stealing pinch/pan on Android

fallaciousreasoning:
## Summary

- Pinch-zoom and pan on the create-climb board were unreliable on Android — either the sheet slid instead of the board zooming/panning, or the pinch got cancelled outright.
- `InteractiveCreateBoard` renders inside `@expo/ui/community/bottom-sheet`'s native sheet (Jetpack Compose on Android, SwiftUI on iOS). Two **separate** gesture surfaces there compete with the board's pinch/pan, and this PR went through two failed attempts before finding both:
  1. **The sheet's own drag-to-resize/dismiss.** First attempt gated `enableContentPanningGesture`/`enableHandlePanningGesture` — confirmed inert by device testing, then by reading `@expo/ui`'s source: its own `types.ts` says outright *"This prop is accepted for API compatibility but has no effect."* Neither `BottomSheet.android.tsx` nor `BottomSheet.ios.tsx` reads either prop; both drive their real gesture toggle from `enablePanDownToClose` alone. Fixed by deriving `enablePanDownToClose` from `!subSheetOpen && !boardInteractionActive` instead — the one prop both platforms actually wire up. This incidentally fixes the pre-existing `subSheetOpen` gating for the hold-role sub-sheet too, which was equally inert.
  2. **The content scroll.** Still unreliable after (1) — more device testing showed the pinch got cancelled specifically by vertical movement, reliable only when pinching along a careful horizontal axis. `@expo/ui`'s `BottomSheetScrollView` (wrapping all of CreateDrawer's content) is, per its own doc comment, *"a direct re-export of React Native's ScrollView"* — a classic Android View, not an RNGH gesture, so nothing could declare a relation to it. Its `onInterceptTouchEvent` could claim the touch stream on the first vertical-ish move, before the board's pinch (several views deeper, behind a Compose-bridge nested `GestureHandlerRootView`) had a chance to activate. Fixed by swapping in RNGH's own `ScrollView` and threading a `scrollRef`, so `useZoomPanGesture` can declare the same relation PlayDrawer already uses for its own surrounding scroll: pinch simultaneous with the scroll (never cancelled), zoomed-pan blocks the scroll (drags the board, not the sheet).
- Considered patching `@expo/ui` directly for (1) — verified with `expo-updates runtimeversion:resolve` that any edit to that package, even pure JS, bumps the Android OTA fingerprint (autolinking-config embeds pnpm's patch-hash directory suffix per module). That would force a store rebuild before this could reach anyone, including this PR's own OTA preview, so both fixes stayed app-source-only. Fingerprint verified unchanged from `main` after every commit.

## Release Notes

- Pinch-zoom and panning on the board now work smoothly while you're building a climb on Android.

## Test plan

1. Android → Create a new boulder problem → board loads.
2. Two-finger pinch on the board, in any direction → board zooms, sheet stays still.
3. While zoomed, one-finger drag in any direction → board pans, sheet stays still.
4. Tap a hold, then long-press → paints, then opens the role sheet.
5. iOS → repeat steps 1-4 → same result.

## Risk

Risk: 2/5 — the underlying bugs are Android-only (two separate gesture surfaces the board can't declare a relation to by default), both fixes are additive on both platforms, confirmed OTA-compatible (fingerprint verified unchanged after every commit), and mobile typecheck/tests/bundle check are all green.

https://claude.ai/code/session_01ESKRXzVj6mNfhn8Aaz77Uk

### DIFF
--- a/packages/mobile/src/components/create-climb/CreateDrawer.tsx
+++ b/packages/mobile/src/components/create-climb/CreateDrawer.tsx
@@ -87,6 +87,14 @@ export function CreateDrawer({
   const windowInsetBottom = useWindowBottomInset();
   const { width: windowWidth, height: windowHeight } = useWindowDimensions();
   const sheetRef = useRef<BottomSheet>(null);
+  // True while InteractiveCreateBoard is zoomed or mid-pinch. The sheet's own
+  // content-panning gesture is a native Compose gesture (Android), not RNGH —
+  // nothing in the board's gesture tree can block it, so it competes directly
+  // with the board's pinch/pan and can win, sliding the sheet instead of
+  // zooming/panning the board. Disabling it for the duration, the same way
+  // subSheetOpen already does for the role-picker sub-sheet, is the only lever
+  // `@expo/ui`'s bottom sheet exposes for this.
+  const [boardInteractionActive, setBoardInteractionActive] = useState(false);
 
   // Measured above-fold height drives the peek snap-point (the native grabber is
   // a fixed reserve now, not a measured custom handle).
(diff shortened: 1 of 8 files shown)


marcodejongh (approved):
Hmmm thought I had merged this amready?! Feel free to nag me on discord next time I miss a pr

PR #5129, opened 2026-09-05
fix(ci): one deployment row per PR, and get the admin creds off the publish job

marcodejongh:
## Summary

A mobile PR left **three** deployment rows in its timeline every push, two of which are GitHub environments used purely as secret containers. GitHub renders "requested a deployment to `ota-preview` — In progress" for each one, which reads as a pending approval. Nothing in this workflow is gated — both preview environments have `protection_rules: []`, so same-repo PRs have always published automatically.

Confirming that turned up real drift:

- **`ota-preview` held `OTA_ADMIN_PASSWORD` + `OTA_ADMIN_EMAIL`** — creds that can delete *any* branch, `production` included. That is the environment the `publish` job declared: the job that checks out PR head and runs PR-author code (`vp install`, `app.config.ts` `execSync`). The workflow header claimed the opposite in capitals. They were never interpolated, but a same-repo `pull_request` runs the PR's *own* copy of the workflow, so reaching them was one diff away.
- The environment only existed to pick up `GOOGLE_MAPS_API_KEY`, so it's retired. The key becomes a repo secret and `publish` declares **no environment at all**.
- **`EOO_TOKEN` has been a repo secret since July**, not `ota-preview`-scoped as `docs/mobile-ota-updates.md`, `CLAUDE.md`, `AGENTS.md` and `mobile-ota-setup.ts` all claimed.

An empty maps key would have failed silently — `eoas` exits 0 and the Android bundle lands under a runtimeVersion no shipped binary has — so `publish` now asserts the key is present *before* it checks out anything, using a presence flag rather than the value.

The one remaining environment, `ota-preview-unattended`, still scopes the teardown creds to `reset` / `cleanup` / the daily sweep. A new `tidy` job retires and deletes the deployment records it materialises. It depends on `reset` + `cleanup` but deliberately **not** on `publish`, so the row clears in seconds instead of after a 150-minute publish, and it declares no environment or it would recreate the row it deletes. It refuses to touch default-branch records, which belong to the sweep and the fork companion.

Fork behaviour is unchanged. The repo is public, so [fork `pull_request` runs never receive secrets](https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows) — an environment approval cannot change that — and `/ota-preview` from a write/maintain/admin user stays the fork publish path.

### ⚠️ Blocked on one manual step before this can merge

`GOOGLE_MAPS_API_KEY` must exist as a **repo-level** secret first:

```sh
gh secret set GOOGLE_MAPS_API_KEY --repo boardsesh/boardsesh
```

The value must be **byte-identical to the copy in the `Production` environment** — seed it from the same source of truth (GCP console / 1Password), not by copying the `ota-preview` copy, which was set 17 days later and may already have drifted. The key perturbs the resolved `android.config`, so a wrong value shifts the Android fingerprint and previews stop reaching store binaries.

**This PR's own preview publish will fail red on the new guard until that secret exists** — which is the guard doing its job.

### After merge

```sh
gh api -X DELETE repos/boardsesh/boardsesh/environments/ota-preview
```

Do this *after* one preview has published Android successfully, not before: while the workflow still named it, GitHub would silently auto-recreate it empty.

Unrelated but worth an audit: `GOOGLE_MAPS_API_KEY` also sits in the `Native Release` environment, which no workflow references. Four copies of a fingerprint-affecting secret is a standing drift hazard.

## Release Notes

## Test plan

1. CI green.

## Risk

Risk: 4/5 — touches the OTA publish pipeline. The maps key must match `Production` byte for byte or Android previews go invisible; the new presence guard catches an empty key but cannot catch a wrong one. Verify by comparing the announce comment's `fingerprint_android` against the latest `android-apk-rn` build's runtimeVersion.


marcodejongh:
Rebased the five original commits onto current main and published 87e8a80963da5db4a3268d51d3b994d0f8d770e8. GitHub now supports `environment.deployment: false`, so reset/cleanup retain their trusted administrator environment without creating cosmetic deployment records. This replaces the original tidy/deleteDeployment job. The explicit full-SHA pr-preview readiness record, current lifecycle locks/stale-head checks/fork authorization and per-platform publish contracts remain intact. Official behavior: https://docs.github.com/en/actions/how-tos/deploy/configure-and-manage-deployments/control-deployments#using-environments-without-deployments .

Read-only metadata confirms ota-preview-unattended has no custom protection rules, which are incompatible with deployment:false. No environment settings, history, secret values, OTA branches or production state were changed. The environment-free publish boundary and early Android maps presence guard are retained.

All 126 focused OTA environment/publish/compatibility/cleanup tests passed; the new suppression contract failed before the workflow change. Full repository checks, all 53 typecheck tasks, independent paired review and the mandatory hook passed. The PR remains draft: repository maps-key equality and a published Android fingerprint matching a shipped binary are not established by these source checks. The original operator prerequisite remains explicit.

PR #5176, opened 2026-09-06
perf(board-presence): index the recent-senders read and pin its angle contract

marcodejongh:
## Summary

Three review follow-ups on the wall-kiosk "recent senders" byline shipped in #4160.

**Index.** `boardClimbRecentSenders` pins `board_id + climb_uuid + angle + status`, but the widest index it could use was `boardsesh_ticks_board_climbed_at_idx` (board_id, climbed_at) — every other predicate fell to a heap filter over every tick ever logged on that board. Migration `0235_yielding_hellion` adds a four-column equality index for it.

`climbed_at` and `user_id` stay out even though the query groups and sorts on them: an index-only scan would also need `board_type`, and those three text/timestamp columns would roughly double the index on our largest table to save a heap fetch on a few dozen rows. `status` is a key column rather than a partial-index predicate (the shape `flash_send_updated_at_idx` uses) because the resolver sends it as bind parameters, which a generic plan cannot prove implication from. Written `IF NOT EXISTS` with the same header as `0121` — the Drizzle migrator wraps migrations in a transaction, so `CONCURRENTLY` is out and prod wants an out-of-band build first. That pattern is now written down in `docs/db-migrations.md` ("Indexes on a large, write-hot table") rather than living only in one file's header.

**Angle contract.** The args schema accepted `-90..90` while a tick can only ever be written at `-5..90` plus a per-board negative-tilt check, so angles no tick could carry reached the DB and came back empty. It now tracks `SaveTickInputSchema` — same bounds, same `isBoardAngleSupported` refinement, with the board row's own type threaded in (a plain string, not `BoardNameSchema`: it is a server-side value, and reporting it as bad user input would be wrong). The client-side guard in `useBoardClimbRecentSenders` moves to match.

The test that read as if it covered this asserted `[]` for angle 90 — a valid angle nobody had ticked, not a rejection. Split into the three cases it was conflating: a blank UUID rejects, an unwritable angle rejects (91, -6, and -5 on kilter), and a valid unticked angle returns empty. Grasshopper's -5 slab, the one negative angle any board routes, gets its own catalog fixture and an acceptance case.

**Over-fetch coverage.** The +3 cushion that absorbs unparseable timestamps had no test, and could not get one from the integration suite: `climbed_at` is `timestamp NOT NULL`, so Postgres cannot produce the input the cushion exists for. Row-shaping moves into `recent-senders.ts` and both sides of the edge are pinned — three unreadable rows still fill the byline, four leave it short. The comment now says the cushion is a soft guarantee rather than implying it always holds.

### One commit here that is not this PR's subject

**`packages/mobile/.../use-board-route-target.test.tsx`** waits 5s via `waitFor` inside a test bounded by Vitest's 5s default, so the inner allowance could never be spent and the test died first — which is why #4418's raise of that `waitFor` never actually stopped the flaking. `packages/mobile/vite.config.ts` still sets no `testTimeout`, so the fix is still live: the per-test budget is now 15s. Reproduced at 1 failure in 4 local runs before, 6 clean after.

### Rebuilt on current `main`

The renumber bot kept this branch moving while `main` took `0215` through `0229`, then stopped three runs in a row: the rebase conflicted outside the migration folder, first in `docs/db-migrations.md`, then `CLAUDE.md`, then `queries.ts`. So the branch was rebuilt by hand, and the three conflicts resolved rather than papered over:

- **`queries.ts`** — #5462 added `assertSprayBoardIsReadable(board, ctx.userId)` right where this PR rethreads `validateInput`. Both are kept; they are adjacent calls, not alternatives.
- **`docs/db-migrations.md`** — main's new "check() constraint footgun" section and this PR's "Indexes on a large, write-hot table" section both kept.
- **`CLAUDE.md`** — main's new `docs/keychain-namespaces.md` line kept alongside this PR's extended `docs/db-migrations.md` description.

The migration was regenerated rather than renamed: `vp run build:db` then `vp exec drizzle-kit generate` against a freshly built schema, so it chained off main's newest snapshot. The generated `.sql` was one statement — the index, no stray `DROP` — and was then replaced with the hand-written `IF NOT EXISTS` + `CONCURRENTLY` header version, same index name and column list. The renumber bot has since moved it five times more, to `0231`, `0232`, `0233`, `0234` and now `0235`, as `main` took `0230_low_the_renegades`, `0231_round_vector`, `0232_repair_moonboard_snapped_tick_angles`, `0233_kilter_live_history` and `0234_shallow_the_phantom`; after each move the SQL, the journal ordering and the snapshot chain were re-checked.

The branch also briefly carried a ported fix for `spray-visibility-sweep.test.ts`, which failed on `main` after SW-17 (#5498) added `Query.sprayWallReports` for the sweep to enumerate. That is gone: #5516 landed the same allow-list entry on `main` while CI was running, so the branch is rebased onto it and carries `main`'s version rather than a duplicate key.

### Verification

`vp check` (0 errors, 377 warnings — unchanged from main), `vp run typecheck` (clean), `vp run check:db-migrations -- --base origin/main` (236 migrations, journal consistent), the `recent-senders` unit file (6 passed), and `board-presence-react` plus the mobile routing test (143 passed across 6 files). The `0235` snapshot diff against `0234_shallow_the_phantom` touches one table, `boardsesh_ticks`, with one added index and nothing removed; the manually-managed `boardsesh_ticks_quality_range` check constraint is preserved. Docker is unavailable in the authoring environment, so the backend integration suite runs in CI — green there on the current `0235` head, along with the dev-database image build that applies the migration folder end to end.

## Release Notes

## Test plan

1. CI green.

## Risk

Risk: 3/5 — backend resolver plus a migration, but the migration only adds an index (no backfill, no data written) and the resolver change narrows an input bound to what the write path already enforced.

https://claude.ai/code/session_01SbYEtLTcpAH5bAK2cVZtaM


marcodejongh:
Rebuilt on current `main` (`10626c41`) — the branch was `CONFLICTING`, and the renumber bot had stopped on it three runs running.

**Why the bot stopped.** Its last three reports were all "couldn't renumber automatically": conflicts outside the migration folder, in `docs/db-migrations.md`, then `CLAUDE.md`, then `packages/backend/src/graphql/resolvers/board-presence/queries.ts`. That last one is the real one — #5462 (spray wall API) added `assertSprayBoardIsReadable(board, ctx.userId)` at exactly the line this PR rethreads `validateInput`. The bot was right to hand it back; that needed a decision, not a strategy flag.

**The three resolutions**, none of them a "take theirs":

- **`queries.ts`** — kept both. The spray assertion and the `boardType`-threaded `validateInput` are adjacent calls in the same guard block, not competing versions of one line. Dropping either would have been a silent regression: theirs loses the angle bound, mine loses the private-wall read check.
- **`docs/db-migrations.md`** — main's new "check() constraint footgun" section and this PR's "Indexes on a large, write-hot table" section are both kept, in that order.
- **`CLAUDE.md`** — main's new `docs/keychain-namespaces.md` line kept alongside this PR's extended `docs/db-migrations.md` description.

**The migration was regenerated, not renamed.** `main` had moved from `0215` to `0229` across five bot renumbers, and `packages/db/dist` in this environment was ten days stale — generating against that would have diffed the old schema and emitted `DROP`s for everything the build lacked, while exiting 0. So: `vp install`, `vp run build:db`, then `vp exec drizzle-kit generate`. The generated `0230` was a single statement — the index, nothing dropped — and I replaced it with the hand-written `IF NOT EXISTS` + `CONCURRENTLY` header version, identical index name and column list. The snapshot chains off main's `0229`, the diff touches only `boardsesh_ticks` with one added index and nothing removed, and `boardsesh_ticks_quality_range` survived.

Also folded the five renumber commits away — the branch is now two commits, the feature and the unrelated mobile flake fix, which is still live (`packages/mobile/vite.config.ts` still sets no `testTimeout`).

`vp check` 0 errors, `vp run typecheck` clean, `check:db-migrations --base origin/main` 231 migrations journal consistent, 6/6 `recent-senders` unit tests, 143/143 across `board-presence-react` + the mobile routing test. Docker is unavailable here, so `board-presence.test.ts` is on CI.

---

marcodejongh:
Took the third note, `adb0d07`. The first two don't hold — both are checkable against the file rather than a matter of taste, so here's the check.

**Test fragility — fixed.** Right that `'2026-07-08T10:00:00.000Z'` silently encoded both `RECENT_CLIMB_SENDERS_FETCH_LIMIT` and the epoch. It's derived from those two constants now. I didn't use `postgresTimestamp(RECENT_CLIMB_SENDERS_FETCH_LIMIT - 1)` as suggested, though: that helper is what builds the *input*, and its whole job is to strip the zone, so reusing it would have made the assertion restate the conversion under test instead of pinning it. Raw `new Date(WINDOW_EPOCH_MS + (RECENT_CLIMB_SENDERS_FETCH_LIMIT - 1) * ONE_DAY_MS).toISOString()` tracks the constants and still fails if the unzoned-to-UTC parse regresses.

**Raw `angle` vs `validated.angle` — already `validated`.** The WHERE clause reads `eq(dbSchema.boardseshTicks.angle, validated.angle)`, and the canonical lookup above it uses `validated.climbUuid`. The raw args are destructured only to be handed to `validateInput`. This is visible in the diff context around the changed lines, and the note is hedged as "likely still references" — worth reading before filing.

**Client guard is strictly tighter than before, not a behaviour change.** The note says -5 on kilter previously "would skip the fetch entirely". `main`'s guard is `angle >= -90` (`use-board-climb-recent-senders.ts:79`), so -5 always passed the client guard and always reached the server. This PR narrows it to `>= -5`, so the set of requests that go out can only shrink. The remaining gap — a kilter board at -5 passing the hook and being rejected server-side — is the one I described in the comment, and it's unreachable in practice: the kiosk sends the board's own angle, and a kilter board has no -5 setting to send.

`vp check` 0 errors and the 6 `recent-senders` tests green on `adb0d07`.

---

marcodejongh:
`test-backend` went red on `adb0d07`, and it isn't this PR's — **`main` fails identically**. Ported the fix rather than waiting, in `f3c676e`.

**The failure.** One test, in an area this PR doesn't touch:

```
packages/backend/src/__tests__/spray-visibility-sweep.test.ts
  > the spray-wall visibility sweep > covers every enumerated field — exercised, or allow-listed with a reason

AssertionError: expected [ Array(1) ] to deeply equal []
+   "Query.sprayWallReports is swept (its arguments name boardUuid) but the OWNER saw nothing of the
+    private wall through it, so the negative half proves nothing. Either seed what it reads in
+    seedWorld(), or add a NOT_APPLICABLE['Query.sprayWallReports'] entry saying why it can never name a climb."
```

**Why it isn't mine.** I checked rather than asserted it: `main`'s own CI run on `10626c41` — the exact commit this branch is rebased onto — fails on the same file, same test case, same assertion string (1 failed / 2614 passed there; 1 failed / 3590 passed here, the difference being shard size). SW-17 (#5498) added `Query.sprayWallReports` to the schema, and the sweep enumerates fields dynamically, so any branch built on that commit inherits the failure. Nothing in this PR is within reach of it — no spray-wall code, no schema field, no seed.

**The fix already exists**, in #5508 (SW-05c), open since yesterday: it adds the `NOT_APPLICABLE['Query.sprayWallReports']` entry the sweep asks for. I read that change and ported exactly that hunk, verbatim including its comment. It no-ops the moment #5508 merges.

I deliberately did **not** port the other half of #5508's change to that file — it also removes the `Query.sessionSummary` entry, which only holds given the rest of that PR (its resolver and seed changes). Taking that here would have turned a green sweep into a different red one.

`vp check` 0 errors on `f3c676e`. Backend integration needs Docker, which this environment doesn't have, so that one is CI's to confirm.

---

marcodejongh:
Both notes are already answered in the diff — pointers rather than a new commit.

**Migration lock / runbook.** This is what the PR adds, not something it leaves out. `docs/db-migrations.md`, "Indexes on a large, write-hot table", opens by stating the exact hazard the note describes — `migrate` wraps migrations in a transaction, so a plain `CREATE INDEX` takes a `SHARE` lock and blocks writes, and on `boardsesh_ticks` that is "a write outage for as long as the build runs" — then gives the three-step pattern, step 1 being "Build the index by hand against production first, outside the migrator: `CREATE INDEX CONCURRENTLY <name> ON <table> (...)`". The `.sql` header repeats it and points back at that section. That is the runbook; it did not exist before this PR, which is why the PR writes it down rather than leaving it in one file's header.

On "if the migration runs first (the default), writes block": true, and deliberately not papered over. `IF NOT EXISTS` means the ordering is safe either way — pre-created, the migration is a no-op; not pre-created, it builds and the cost is a blocked-write window, not a broken deploy. I can't quote an expected duration honestly: it depends on the production row count at the time, which I have no way to measure from here, and a made-up number in a runbook is worse than the qualitative rule that's there.

**`boardType: z.string()`.** Right that it is never read back off `validated` — but it isn't dead input. It is consumed inside the schema, by the refinement that is the point of this change:

```ts
.refine((args) => isBoardAngleSupported(args.boardType, args.angle), {
  message: BOARD_ANGLE_VALIDATION_MESSAGE,
  path: ['angle'],
});
```

Remove it and the per-board negative-tilt check goes with it. On tightening it to `BoardNameSchema`: declined, same reason as earlier in this thread — `boardType` comes from the bound board row (`board.boardType`), not the client. A board row carrying an unknown type is a server-side data problem, and `validateInput` would report it to the caller as bad user input with a 400, which is the wrong error for the wrong party. `z.string()` is the honest type for a value the server vouches for.

That's this file's third round of notes on the same two decisions. Both are recorded in the code and the PR body; I'd rather not spend another CI cycle restating them.

---

marcodejongh:
Dropped the ported spray-sweep fix and rebased onto `cd54e96b` — `main` now carries it as #5516.

Timing: I ported #5508's `NOT_APPLICABLE['Query.sprayWallReports']` hunk to get this branch off a failure it inherited from `main`, and while CI was running, #5516 landed the same fix on `main` as a standalone PR. Keeping both would have put the same key in that object twice, so the port is gone and `main`'s version — which carries a fuller comment about the admin-only read and why the owner half is vacuous — is the one that survives. `sprayWallReports` now appears exactly once on this branch, and it is `main`'s.

That leaves the three commits this PR is actually about. `0230_yielding_hellion` is untouched and still unclaimed on `main`.

Re-validated after the rebase: `vp check` 0 errors, `vp run typecheck` clean, `check:db-migrations -- --base origin/main` 231 migrations journal consistent, `recent-senders` 6/6.

For the record on the head before this one: CI went green on `f3c676e` — `test-backend`, both `test-default` shards, `db-migrations`, `typecheck`, `lint`, `mobile-bundle`, `docker-web`, `test-postgres-postgis`, `ci-status` and `test-report` all success, 6318 passed / 0 failed / 2 skipped. The rebase changes no source, so I expect the same on `8c81271`.

---

PR #5212, opened 2026-09-06
fix: grade a climb at a new angle from the first send, and refresh the library after a tick

### LINKED ISSUE #4798: 🐞 Bug report: On the woods board, When I log a climb at a new angle from the one it was set at, it does not show up...

## Comment

On the woods board, When I log a climb at a new angle from the one it was set at, it does not show up in the library under the grade given at the new angle. I would expect the climb to populate after being sent at a new angle when searching climbs at that angle.

## Metadata

| Field | Value |
| --- | --- |
| Platform | ios |
| App version | 2.3.1 (2) |
| Source | drawer-bug |
| Board | woods / layout 1 / size 2 / sets [1] @ 40° |
| Climb | Red Rumped Bush Tyrant |
| Session | unknown |
| URL | /climbs |
| User agent | unknown |
| Feedback row | app_feedback #10216 |

## Contact

✅ Reporter agreed to follow-up — resolve **app_feedback #10216** in the DB to reach them.

marcodejongh:
## Summary

Fixes #4798.

A Woods problem is graded only at the angle it was set at. When a climber sent one at a different angle and logged a grade, the climb never appeared in the library at that angle under that grade. Kilter and Tension had the same hole at angles Aurora has no stats for. Four causes, all fixed here:

1. **The recompute never graded a tick-seeded row.** `board_climb_stats` gets a row at the tick's angle, but `display_difficulty` / `difficulty_average` were written only for Boardsesh-owned climbs, so a catalog climb at a new angle stayed ungraded forever and every grade filter dropped it. The recompute now derives the grade from ticks whenever the row has no grade, or the grade it holds is one the recompute wrote itself and no upstream sync has stamped the row since. A new nullable `board_climb_stats.tick_graded_at` column is that marker. Upstream grades (Aurora, kilter-sync, the Woods and MoonBoard importers) are never touched: a graded row with a NULL marker is upstream's and stays verbatim. Measured on prod before choosing the marker: 12 Kilter + 2 Tension non-owned graded rows have no `upstream_synced_at` stamp, and 134k graded Tension rows are unstamped, so "unstamped" alone could not mean "ours".
2. **The grade arrived 2 s after the app refetched.** `saveTick` and `updateTick` now run the single-key recompute inline before the debounced queue, so the row is graded by the time the mutation returns. The debounced path still coalesces bursts and publishes `climbStatsUpdated`.
3. **The mobile library never refetched.** The tick hooks busted `['searchClimbs']`, but the `/climbs` list reads `['infiniteSearchClimbs']` and `['searchClimbsCount']`. Both are now invalidated by the save/edit/delete hooks, the climb-save hooks, and the offline drainer when a queued tick lands.
4. **Downloaded boards read stale SQLite.** After a local tick write the adapter ran a bare outbox drain. It now runs a drain-then-pull sync cycle (single-flight, same cost as an app foreground), so the server-graded stats row lands in the device database and the list refreshes.

**Migrations (approved by the maintainer 2026-09-06):** 0218 adds the nullable column. 0219 backfills the rows already sitting ungraded: 162 candidates on prod (kilter 147, tension 8, woods 3, touchstone 3, grasshopper 1), 88 of which carry a graded tick and get a grade.

Not in scope, filed separately: the web SSR `climb-search-v8` cache still serves Woods list pages for 24 h.

Author-side verification: typecheck for db, backend, shared and mobile; `vp check` (0 errors, formatting clean); `check:db-migrations`; backend real-Postgres suites for the recompute (incl. the new Woods, legacy-unstamped, stamp-ordering, delete-clears and backfill-migration cases), the tick mutations, and board-presence; db package tests; the full mobile suite (8,452 tests); Metro bundle check. The two `snapshot-bootstrap` timeouts in offline-sync reproduce on main's version of the only file this PR changes there.

## Release Notes

- Send a Woods problem at a new angle and it shows up in the library at that angle with the grade you gave it
- Kilter and Tension climbs get a grade at angles nobody had logged before, from the first send
- The climb list refreshes right after you log a send, on downloaded boards too

## Test plan

1. Signed in, Woods board set to an angle the climb was not set at.
2. Open the climb → log a send with a grade → back to Climbs.
3. Grade filter on that grade → the climb is listed at this angle.
4. Switch to the climb's set angle → it still shows its original grade.
5. Download the board for offline → repeat 2–3 → list refreshes by itself.

## Risk

Risk: 5/5 — two migrations (one with a prod backfill), shared stats SQL used by every board, and a new per-tick sync cycle on mobile. The per-tick cycle is one more trigger point for a fresh catalog download of an enabled-but-not-yet-downloaded board, the same class of trigger as every app foreground today; it runs on the offline-sync client with the hard request deadline. Replaying N queued offline ticks now serializes N inline recomputes in the drainer before the debounced pass; each is one short PK-indexed transaction.

https://claude.ai/code/session_012bjLP9V8oC58RB8xFKpo3r


packages/db/src/queries/climb-stats/real-catalog-data.ts
```diff
@@ -81,3 +91,61 @@ export function statsRowCarriesRealCatalogDataSql(tableAlias: StatsTableAlias):
...
+  const alias = sql.raw(tableAlias);
+  return sql`(
+    ${alias}.board_type <> 'moonboard'
+    AND (
+      ${alias}.display_difficulty IS NULL
+      OR (
+        ${alias}.tick_graded_at IS NOT NULL
+        AND (
+          ${alias}.upstream_synced_at IS NULL
+          OR ${alias}.tick_graded_at > ${alias}.upstream_synced_at
```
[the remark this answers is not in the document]
marcodejongh: Taken in 6973baff9 — good catch. The stamp comparison is gone: the recompute now derives when the row has no grade or `tick_graded_at IS NOT NULL`, and the upstream writers own the marker. Aurora shared-sync clears it whenever it writes the grade; kilter-sync catalog-sync and stats-repair share `kilterStatsGradeConflictSet()`, which clears it exactly when the incoming difficulty is non-NULL (mirroring their COALESCE); the Woods importer does the same. Regression tests cover a tick-graded row with a newer stamp still re-deriving and clearing (single + bulk), plus the writer conflict sets.

packages/db/src/queries/climb-stats/recompute.ts
```diff
@@ -56,8 +56,69 @@ type DrizzleDb = PgDatabase<PgQueryResultHKT, Record<string, unknown>>;
  *     next pass. (The one-time 0157 backfill clears the pre-existing
  *     tick-derived crowns this rule used to allow.)
  *
- * difficulty_average / display_difficulty: recomputed from flash/send ticks on
- * OWNED climbs only (upstream owns them on synced climbs).
+ * difficulty_average / display_difficulty / tick_graded_at: recomputed from
```
marcodejongh, line 59: This comment is way too long. Lets make it more concise, nobodys going to read this. But it does seem like this comment is important

packages/db/src/queries/climb-stats/real-catalog-data.ts
```diff
@@ -26,9 +26,19 @@ import { boardClimbStats } from '../../schema/boards/unified';
  * failure direction is protective: an extra TRUE reads as "real catalog data",
  * and the fix responds by declining to move the tick or delete the row.
  *
- * The tick recompute never writes any of the four (see the recompute module doc),
- * so a row where all four are absent is provably an artefact of a tick landing at
- * that angle — a phantom row — rather than a graded angle of the problem.
+ * The tick recompute writes exactly one of the four, and only where there is no
```
marcodejongh, line 29: Same, this comments is way too long

packages/db/drizzle/0219_backfill_tick_graded_climb_stats.sql
```diff
@@ -0,0 +1,81 @@
+-- Backfill the grades of stats rows that only exist because of Boardsesh ticks.
+--
+-- Refs #4798. Approved by maintainer 2026-09-06.
+--
+-- Until this migration the tick recompute wrote difficulty_average /
```
marcodejongh, line 5: Long comments in migrations I have less issues with

docs/ascents-and-attempts.md
```diff
@@ -168,13 +168,28 @@ There are exactly three production code paths that insert into `boardsesh_ticks`
   media id when possible. That keeps one canonical beta row per video and one
```
marcodejongh, line 168: I wonder if we should open a websocket connection on climbs currently in view in the UI to get live updates? that way if 2 users are climbing the same board, if the other ticks it the changes will automatically stream-in? Might be worht a followup issue

PR #5218, opened 2026-09-06
fix(mobile): reach the climb actions menu with a screen reader, and measure the list

marcodejongh:
## Summary

Groundwork for the climb-list IA work. No pixels move. Two accessibility dead ends and three measurement holes, all found while working out whether the row's ⋮ button earns its place.

**The ⋮ is a user setting. The menu's reachability was riding on it.** `rowAccessibilityActions` published the `moreActions` custom action only when the button was visible. RNGH gestures never register with the accessibility bridge (the reasoning is already written down in `lib/row-accessibility-actions.ts`), so the long-press is unperformable by VoiceOver and TalkBack — which means with "Show quick-actions button" off, queue, tick, favourite and playlist were unreachable to a screen reader entirely. The action is now published whenever the row can open the menu at all. The code comment deferring this for "product copy" was stale: `mobile.climbRow.moreActions` already exists and the button already uses it.

**The row read out its name and nothing else.** The container is `accessible` with `accessibilityLabel={climb.name}`, and an explicit label on an `accessible` container overrides its children — so the benchmark and no-match glyphs, the favourite heart, the ascent status, the subtitle and the grade were all silent, every one of which already publishes its own label. Dropping the label lets RN compose them, and the children are already in visual order. The ⋮ gets `accessibilityElementsHidden` on iOS so its own label does not tack "More actions" onto every row VoiceOver reads; TalkBack focuses inside the row, so Android keeps the button as it is.

**A screen-reader menu open no longer claims to be a button tap.** The row's custom action is published whether or not the ⋮ is on screen, so it reports `source: 'accessibility_action'` rather than crediting a button the climber may not have.

### The measurement half

Three events, because the ellipsis question could not actually be answered from what we collect:

- **`surface` on `Climb Actions Opened`** (`climbs_list` | `playlist` | `profile` | `board_sheet`). Only the climbs list passes `showMoreButton`, so every other surface can emit `long_press` and nothing else. Today's 154-⋮-users vs 86-long-press-users split therefore pools four screens against one and reads as a like-for-like comparison it is not.
- **`Climb Row Tapped`** — the denominator. We know how many people open the menu; we have never known how many rows they open first, so "the menu is a minor path" has been unfalsifiable in both directions.
- **`Climb List Paginated`** — `handleEndReached` tracked nothing, so list scroll depth is invisible. Any future row-density or scan-speed change is unevaluable without it. The page index is read through a ref so the callback's deps stay free of a `.length`, which would rebuild `renderItem`'s siblings on every page and re-render every mounted row.

Both new a11y tests were confirmed to fail against the unfixed component (`2 failed | 7 passed`), then pass with it (`9 passed`).

**Worth a reviewer's eye:** the composed label is the part no unit test can prove. The RN mock makes `View` a bare div, so the tests pin that the row no longer *sets* a label — not what VoiceOver and TalkBack actually speak. That needs the device pass in the test plan, on both platforms.

## Release Notes

- VoiceOver and TalkBack can now open a climb's quick-actions menu from the climb list, even with the ⋮ button turned off — before, the menu was unreachable without it
- Screen readers now read a climb row in full: its grade, whether you have sent it, whether you have favourited it, and its badges — not just the name

## Test plan

1. More → Display → turn the ⋮ button off.
2. Turn on VoiceOver (or TalkBack), open Climbs.
3. Swipe to a climb row → it reads name, grade, status.
4. Use the rotor (iOS) or actions menu → "More actions" opens the menu.
5. Turn the ⋮ back on → row still reads once, not twice.

## Risk

Risk: 2/5 — no layout changes and no new user-facing surface; it removes one accessibility prop and adds three analytics events. The exposure is what screen readers speak on every climb row, which unit tests cannot cover, so it carries a real device pass on both platforms.

### DIFF
--- a/packages/mobile/app/(tabs)/climbs/index.tsx
+++ b/packages/mobile/app/(tabs)/climbs/index.tsx
@@ -757,9 +757,19 @@ function ClimbListInner() {
     void refetch();
   }, [refetch]);
 
+  // Read through a ref so the tracking below can name the page being fetched without
+  // putting a `.length` in this callback's deps — that would rebuild `renderItem`'s
+  // sibling handlers on every page and re-render every mounted row (perf checklist §3).
+  const loadedPageCountRef = useRef(0);
+  loadedPageCountRef.current = searchPages?.pages.length ?? 0;
+
   const handleEndReached = useCallback(() => {
     if (hasNextPage && !isClimbsLoading && !isFetchingNextPage && !isRefetching && !isLoadingMoreRef.current) {
       isLoadingMoreRef.current = true;
+      // List scroll depth, which was previously unmeasurable: this handler tracked
+      // nothing, so no row-density or scan-speed change could be evaluated against
+      // how far people actually scroll. 0-based, matching the search pages.
+      track(SHARED_EVENTS.ClimbListPaginated, { page: loadedPageCountRef.current });
       void fetchNextPage().finally(() => {
         isLoadingMoreRef.current = false;
       });
@@ -1384,6 +1394,7 @@ function ClimbListInner() {
         showPlaylistChips
         showFavorite
         showMoreButton={quickActionsButtonEnabled}
+        surface="climbs_list"
       />
     ),
     [
--- a/packages/mobile/src/components/ClimbListRow.tsx
+++ b/packages/mobile/src/components/ClimbListRow.tsx
@@ -216,8 +216,12 @@ type ClimbListRowProps = {
    * `Climb Row Tapped`. Only the climbs list passes `showMoreButton`, so without
    * this every other surface can emit `long_press` and nothing else — and a
    * `source` breakdown silently pools four screens against one.
+   *
+   * REQUIRED on purpose: an optional one would let a new list ship rows whose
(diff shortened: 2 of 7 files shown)


marcodejongh:
Rebased and narrowed to the still-relevant accessibility repair at `fa5217f83b1572ddc27103668cb43d55b03b5002`. Main intentionally retired unread climb-action telemetry in #5249, so the event catalog, list loader and row callers now remain identical to main. The former pagination tracking observations are superseded by this reduced scope.

VoiceOver/TalkBack menu actions remain available when the visible button is hidden. Unsupported rows and removed handlers neither open a menu nor vibrate; iOS and Android retain their distinct nested-button exposure. Two new stale/unavailable regressions failed before the correction.

Validation: 21 focused row tests; all 11,305 mobile tests; mobile and all 53 monorepo typechecks; native iOS/Android and browser Metro exports; lint/format; mandatory hook; independent paired review. Draft remains while exact-head CI and the published preview for the shipped runtime are checked. Device speech still requires the PR test plan.

PR #5221, opened 2026-09-06
feat(mobile): teach the long-press, and that the ⋮ is optional

marcodejongh:
Stacks on `feat/climb-row-a11y-telemetry` — review that one first; this PR's base is that branch, not `main`.

## Why

Two ways into the climb quick-actions menu, and climbers find exactly one of them. PostHog, production, week of 2026-08-30:

- 154 unique users opened the menu with the **⋮**, 86 with a **long-press**, 214 unique users in total — so only ~26 (12%) ever used both. Roughly 128 people have never found the long-press.
- In two months, **11 users** have ever turned the ⋮ off and 6 turned it back on. That reads less like "nobody wants to" and more like "nobody knows they can".

The button, its default, and the setting are untouched. This PR only socialises both.

## What was wrong, and what changed

**1. The tip fired at the worst possible moment.** `mobile.onboarding.quickActionsTip` armed on a `useFocusEffect` the *first* time a climber landed on the Climbs tab — before they had any reason to want a menu — and marked itself seen on render.

It now waits for the **third** landing on Climbs, and steps aside entirely for anyone who has already opened the menu. Both signals are local, through the existing keyed tip store: a counter (`onboarding_tip_quickactions_visits`, which stops writing at the threshold) and a flag (`onboarding_tip_quickactions_used`). It still never stacks with the board-reveal banner, and anyone who saw the tip under the old timing keeps their existing seen flag and is not taught twice.

The "has opened the menu" flag is written at `DrawerHostProvider.openClimbActions` — the one chokepoint every long-press, ⋮ tap and screen-reader action from every surface goes through — so a long-press on a playlist or a profile counts too. It is guarded to one write per launch. `ClimbListRow`, the ⋮ button, the preference and its default are not touched; the row does not move a pixel.

**2. It named only one of the two paths.** New copy names the long-press, the ⋮, and that the button can be hidden, in all four locales.

**3. It was completely uninstrumented.** Three events added to `@boardsesh/analytics`, each with a `tip` prop so one funnel covers every just-in-time tip: `Onboarding Tip Shown` (plus `visitCount`, so the third-visit rule is checkable rather than assumed), `Onboarding Tip Dismissed`, `Onboarding Tip Pressed`.

The **outcome** metric already exists and is not duplicated here: `Climb Actions Opened` carries `source: 'long_press' | 'more_button' | 'accessibility_action'` and, as of the branch this stacks on, a `surface` property. **The number to watch after this ships is the long-press share of `Climb Actions Opened` within `surface = 'climbs_list'`** — never pooled across surfaces, since only the Climbs list renders a ⋮ at all.

## Where the banner press goes, and why

`OnboardingTipBanner`'s `onPress` now pushes `/(tabs)/profile/more`.

Display is a **section inside** that screen, not a route: `app/(tabs)/profile/more.tsx` builds one view-model that a native SwiftUI `Form` (iOS) / Compose `LazyColumn` (Android) renders, with no anchor, scroll-to, or search-param handling. So More is as deep as a link can go without building section-deep-linking into the native form, which is out of scope here. The setting and its explanatory subtitle ("You can always long-press instead") are one screen down.

## Copy, and what got dropped

Measured with a TTF advance-width parser over Liberation Sans Regular (metric-compatible with Arial, which runs a few percent **wider** than SF Pro Text, so these are conservative) at 15pt — the `subheadline` size the banner uses at Dynamic Type 1.0. Available text width on a 393pt phone is **277pt**: 393 − 32 banner margins − 24 surface padding − 8 gap − 28 leading icon+gap − 24 close button.

| Locale | Copy | Width | Lines |
| --- | --- | --- | --- |
| en-US | Quick actions: long-press a climb, or use ⋮. Tap to hide that button. | 448pt | 2 |
| es | Acciones rápidas: mantén pulsado un bloque, o usa ⋮. Toca para ocultarlo. | 501pt | 2 |
| fr | Actions rapides : appui long sur un bloc, ou le bouton ⋮. Touche pour le masquer. | 543pt | 2 |
| de | Schnellaktionen: Boulder lange drücken oder ⋮ nutzen. Tippen zum Ausblenden. | 537pt | 2 |

**The brief's one-line constraint turned out not to exist**, and could not have been met. `OnboardingTipBanner` sets no `numberOfLines`, so its text wraps freely, and one line at 277pt holds about 34 characters — three ideas will not fit in any language. The *existing* copy already wrapped: 2 lines in en/es/fr and **3 in German**. So the real target is two lines, which is what shipped: three of four locales keep the height they already had, and German shrinks from three lines to two.

**Dropped idea, all four locales: the enumeration of what the menu contains** ("queue, tick and playlists" / "cola, encadene y listas" / "file, croix, listes" / "Warteschlange, Eintrag und Playlists"). It was the weakest of the four for this PR's purpose — the tip's job is teaching the *gesture*, not the menu's contents — and it is already spelled out in the settings subtitle one tap away. All three ideas the brief asked for survive in every locale.

Glossaries followed: `plafón` is untouched (the string says *bloque*, the climb, not the board); French avoids *envoyer* (« appui long », « masquer »); German avoids *senden* and keeps informal du.

## Test infrastructure

Importing the tip flags into `DrawerHostProvider` pulls `expo-secure-store` into that module graph, and its `main` points at TS source whose `expo-modules-core` re-export throws in the Vitest worker — it red the existing `drawer-host-provider` suite at import time. Added an inert, always-empty `expo-secure-store` stub behind a vite alias, following the `expo-file-system` / `expo-image` / `expo-haptics` stubs already sitting next to it. `vi.mock` still takes precedence, and the stub deliberately does not remember writes so nothing leaks between suites.

## Release Notes

The Climbs list now shows you both ways to reach a climb's quick actions — long-press a climb, or use the ⋮ button — and tells you the ⋮ is yours to hide. It waits until your third visit instead of your first, and never bothers you if you already found the menu. Tap it to go straight to the setting.

## Test plan

1. Reinstall the app, or More → Replay walkthrough. Onboarding restarts.
2. Open Climbs, leave, return, leave, return. Tip appears third time.
3. Read the tip. It names long-press, ⋮, and hiding it.
4. Tap the tip body. More opens; Display holds the ⋮ setting.
5. Reinstall, long-press any climb, then open Climbs repeatedly. No tip.

## Risk

Risk: 2/5 — JS-only, OTA-compatible. One new screen link and four copy strings; no change to the ⋮ button, the setting, its default, or the climb row.

https://claude.ai/code/session_01ELkpe6R3CCsHmmvNShY9Ta


marcodejongh:
Published repair: `1ab52918e78458d1b7c7ce42adf831f6ad7e8dd5`.

Restacked this PR’s two own commits onto repaired #5218. The third-visit tip, old seen flags, any-surface menu-use detection and four locales remain relevant; main’s retired climb-action telemetry is preserved.

Addressed the prior feedback and paired-review timing finding:
- Removed the test-only production reset export; test isolation now resets modules.
- Seen/menu-use flags become effective before preference writes finish, including failed writes and stale reads.
- A subscribed visibility hook immediately retires both visible tips and tips waiting behind the board-reveal banner; a shown guard prevents repeated events.
- Removed the obsolete claim that retired climb-action events still measure this feature.

Validation: all 11,329 mobile tests, 79 focused storage/visibility/drawer tests, 217 shared storage/analytics/locale tests, native iOS/Android and browser exports, mobile typecheck, independent paired review, and final full lint/typecheck and mandatory hook. The pending menu-write regression failed before correction. Exact publication SHA and fresh CI/OTA status are tracked separately; device navigation and screen-reader behavior still need the PR test plan.

marcodejongh:
Addressed the final test-fixture nit: the visibility suite now explicitly returns Promise.resolve(null) for an absent storage read, matching the real adapter. No application or native inputs changed. All 11,329 mobile tests passed again; final lint/format, all 53 typecheck tasks, the mandatory hook and independent paired review passed.

marcodejongh:
Published conflict repair `71d0515a3b3acee23db66adf49440f8a14473232` onto main `e597969912305a4cf4fbd3545dc361d850632e27`. The sole conflict was additive analytics events; both main's onboarding-gate event and all three tip events remain. Paired range review, all 11,627 mobile tests, native/browser bundles, full 53-task typecheck and lint passed. Draft while exact-head CI and shipped-runtime preview checks run again.

marcodejongh:
Addressed in `9e2e2bf3bfddcb5b51f85a4e2972f745bc5cf196`. Both dismiss and settings press now record the seen flag before analytics or navigation, including interactions before the shown effect runs. Component regressions exercise the immediate-interaction race, all three event payloads, the shown visit count, and the More destination. The concurrent focus comment now describes a possible delayed visit without claiming a fixed one-visit bound. The event contract is recorded in `POSTHOG_INVENTORY.md`.

The More URL follows the existing Expo Router literal-path convention; `route-segments.ts` contains route predicates rather than a shared destination constant for this path. The component test now pins that destination.

Validation: independent paired source review, 60 focused tests, all 12,008 mobile tests across 1,034 files, native iOS/Android and browser bundles, all 53 typecheck tasks, lint/format and the commit hook passed. Physical device QA remains outstanding on the refreshed preview.

marcodejongh:
Reviewed both observations against `9e2e2bf3bfddcb5b51f85a4e2972f745bc5cf196`. Concurrent counter reads can share an increment and delay eligibility, as the source comment records; the visit cap and synchronous seen/used guards still prevent repeated prompts. The child layout-effect interaction is deliberate: it exercises dismissal/press before the parent passive shown effect, and the test checks persistence precedes interaction analytics. It remains a component regression alongside the independent pending-write/refocus helper tests. No further runtime change is needed for these nonblocking observations.

Issue #4838, opened 2026-08-27
🐞 Bug report: Bring back the medium view while scrolling playlists/the climb database, larger wall images with names...


marcodejongh:
Reported from the Boardsesh mobile app.

## Comment

Bring back the medium view while scrolling playlists/the climb database, larger wall images with names above

## Metadata

| Field | Value |
| --- | --- |
| Platform | ios |
| App version | 2.3.1 (2) |
| Source | drawer-bug |
| Board | kilter / layout 1 / size 10 / sets [1,20] @ 50° |
| Climb | double dutch (8a+/V12) |
| Session | unknown |
| URL | /discover |
| User agent | unknown |
| Feedback row | app_feedback #10225 |

## Contact

✅ Reporter agreed to follow-up — resolve **app_feedback #10225** in the DB to reach them.

Issue #4801, opened 2026-08-25
[UI] Climb: On mirrored boards, it's hard to tell whether you’ve sent one or both orientations


marcodejongh:
Reported from the Boardsesh mobile app.

## Comment

On mirrored boards, you can’t tell when looking at the list of climbs if you’ve sent either or both the original and mirror version of climbs. It would be nice if there was a forward and backward check mark or something.

## Metadata

| Field | Value |
| --- | --- |
| Platform | ios |
| App version | 2.3.1 (2) |
| Source | drawer-bug |
| Board | woods / layout 1 / size 2 / sets [1] @ 40° |
| Climb | hip dysplasia (7c/V9) |
| Session | unknown |
| URL | /climbs |
| User agent | unknown |
| Feedback row | app_feedback #10219 |

## Contact

✅ Reporter agreed to follow-up — resolve **app_feedback #10219** in the DB to reach them.

Issue #4635, opened 2026-08-20
[UI] Climbs: Indicate which climbs are routes, and their frame count


ES-Alexander:
Currently if you filter by boulders and routes it's not clear from the list which ones are which. The preview image on routes is also just the first frame, which can make it seem like an undesirable climb until it's clicked on.

I suggest adding a frames icon with a count number (possibly in the middle of it?) in the listing for routes (i.e. any climb with more than one frame. Given the preview image for routes is already not super helpful, that icon could maybe be on top of the preview image, so it's very visible. Otherwise it could be below it, but that might increase the list item height, or result in the indicator being quite small and hard to read.

PR #5222, opened 2026-09-06
feat(mobile): let climbers pick how much rides in each climbs-list row

marcodejongh:
Stacked on `feat/climb-row-a11y-telemetry` — review that one first; this PR's base is that branch, not `main`.

Ships the selectable climbs-list density asked for in #4838 (bring the medium view back) as one control instead of three competing row requests (#4635, #4801, #4533). Set once in More → Climb list; the Climbs tab is the only surface that reads it.

| Tier | Thumbnail | Content | Measured row pitch |
| --- | --- | --- | --- |
| Compact | 56×72 | name + attribute glyphs + grade | **88pt** (72 + 8 + 8) |
| Standard | 76×96 | today exactly | **112pt** (96 + 8 + 8), unchanged |
| Rich | 76×96 | standard + playlist tags under the subtitle | **112pt** |

Plus a hairline separator on every tier, so the on-screen pitch is 88.33 / 112.33pt at 3× DPR.

### No tier renders a thumbnail wider than 76pt

`ClimbListThumbnail` renders at `Math.max(400, cellWidth * 5)`, so the compact 56pt cell resolves to the **same 400px render** as today's 76pt cell — byte-identical, no new cache generation. A larger cell would not be free: 132pt asks for 660px and mints a second generation in both the native render cache and expo-image's disk cache, on the app's largest memory consumer (`docs/react-native-performance.md` §7 — foreground OOM kills on 4 GB iPhones, #3479). **So the rich tier adds a line, not pixels.** That rule is written into `climb-list-thumbnail-metrics.ts` and asserted by a test that walks every tier.

It also means compact had to shrink the cell: the thumbnail is the tallest child in the row, so a tier that only dropped text lines would have saved nothing.

### Measurements, and what the right-hand rail does

The rail is `flexDirection: 'row'` — the grade sits *beside* the ⋮, not stacked under it — so the rail's height is `max(44pt ⋮, grade)`. The grade is `title3`: 25pt line height on the Liquid Glass scale, 28pt on Material. 44 < 72, so the **thumbnail still drives the compact row** and the pitch is a real 88pt, not the taller number a stacked rail would have forced. That holds to Dynamic Type 1.5× too (title3 → 37.5/42pt, ⋮ fixed at 44pt, name → 33pt — all under 72). No hitSlop or negative margin was needed; the ⋮ keeps its full 44pt target.

### FlashList: nothing needed, and here is why

Checked against `@shopify/flash-list@2.3.2` in `node_modules`, not from memory:

- There is **no `estimatedItemSize`** in v2's `FlashListProps.ts` at all — nothing to keep in sync.
- `ViewHolder.tsx` gives every cell `onLayout → onSizeChanged(index, layout)`, and sets an explicit `height` only when `layout.enforcedHeight`. `LinearLayoutManager.ts:88` sets only `enforcedWidth` for a vertical list, so row height stays content-driven and re-measures on any re-render.
- `LayoutManager.ts` (`computeEstimatesAndMinMaxChangedLayout`) compares each reported height against the stored layout and recomputes from the lowest changed index — exactly the relayout a tier switch needs.
- `ViewHolder`'s memo comparator includes `prevProps.renderItem === nextProps.renderItem`, and `density` is in `renderClimbItem`'s deps, so every mounted cell re-renders and re-measures on a switch.
- `getItemType` would be actively **wrong** here. It only partitions the running height-average window (`getEstimatedHeight → heightAverageWindow.getCurrentValue(getItemType(index))`); the tier is global, so every item shares one type at any instant and a density-keyed type buys no estimate accuracy — while throwing away the whole recycling pool on each switch, forcing fresh mounts of every visible row on the app's largest memory consumer. Same reason no `key` on the list (it would also dump scroll position).

### Choices worth a second pair of eyes

- **`climb-list-row-styles.ts` values are untouched.** Its docblock says it exists so `ClimbPreviewCard` renders byte-for-byte like a list row, so the tiers add a style *alongside* (`compactSeparator`) rather than overriding. No style-override prop was needed: `contentRow` is identical across all three tiers (same 8pt padding, same 12pt gap) — only the thumbnail's height moves the row, and only the separator's inset moves sideways. Both separator insets are derived from `separatorInsetForDensity()` in the metrics module, so the compact row carries no second magic number.
- **Rich shows the playlist tags whatever the "Show playlist tags" toggle says.** That toggle defaults OFF, so without this the rich tier would have been a no-op for most climbers — picking rich in More → Climb list *is* the opt-in, and the section footer says so. The toggle still governs the standard tier. Shout if you'd rather rich deferred to it.
- **The control is its own section, not a row inside "Display".** A segmented row publishes no visible label of its own — `MoreForm.ios.tsx` says the section header *is* the label — so folding it under a header reading "Display" would have left three unexplained segments. It sits directly above Display, following the Grade Format pattern (segmented row, description as the section footer).
- Compact also drops `LiveClimbSubtitle`, and with it a per-row `useEffectiveClimbStats` subscription, so a compact row genuinely costs less rather than just looking smaller.
- The initial-load skeleton (`ClimbListRowSkeleton`) still draws at 112pt in every tier. Left alone to keep the diff on the row itself; a compact user sees one settle on first load.

Seams for the follow-ups are left open and empty: no placeholder props for the personal-progress line or the artwork pips.

## Release Notes

The Climbs tab now fits your eyes. Pick Compact for a tighter list that gets you deeper into the results in one scroll, Standard for the row you already know, or Rich to see which playlists a climb is in without opening it. One setting, in More → Climb list.

## Test plan

1. More → Climb list → tap Compact. Rows shrink, subtitles go.
2. Scroll the Climbs tab. Smaller wall images, still sharp.
3. Tap Rich. Rows grow back, playlist tags appear.
4. Tap Standard. Rows look exactly as before.
5. Force-quit, reopen Climbs. Your pick survived.

## Risk

Risk: 2/5 — JS-only row geometry behind a new setting that defaults to today's exact row; no native change, no migration, no BLE.

https://claude.ai/code/session_01ELkpe6R3CCsHmmvNShY9Ta

### DIFF
--- a/packages/mobile/app/(tabs)/climbs/index.tsx
+++ b/packages/mobile/app/(tabs)/climbs/index.tsx
@@ -90,6 +90,7 @@ import {
 } from '../../../src/lib/onboarding/onboarding-storage';
 import { ONBOARDING_TIP_QUICKACTIONS_KEY } from '@boardsesh/key-value-storage';
 import { useClimbQuickActionsButton } from '../../../src/lib/climb-quick-actions-button-preference';
+import { useClimbListDensity } from '../../../src/lib/climb-list-density-preference';
 import { useAuth } from '../../../src/providers/auth-provider';
 import { ensureBackgroundsCached } from '../../../src/lib/background-image-cache';
 import {
@@ -210,6 +211,10 @@ function ClimbListInner() {
   // The ⋮ quick-actions button is a user setting that defaults on (More → Display
   // lets climbers turn it off).
   const { enabled: quickActionsButtonEnabled } = useClimbQuickActionsButton();
+  // Row density is a user setting too (More → Climb list), and ONLY this list reads
+  // it — the other four surfaces that render a ClimbListRow keep the default shape.
+  // A primitive, so it drops straight into renderClimbItem's deps below.
+  const { density: rowDensity } = useClimbListDensity();
   const { addToQueue } = useQueueActions();
   const {
     filters,
@@ -1378,6 +1383,17 @@ function ClimbListInner() {
     [useNativeSearch, t, handleNativeSearchChange, handleSearchFocus, handleSearchBlur, handleNativeSearchCancel],
   );
 
+  // A density change relayouts every row, and FlashList v2 needs nothing extra for
+  // it. `rowDensity` in these deps gives `renderItem` a new identity, which is one
+  // of the props ViewHolder's memo compares (ViewHolder.tsx), so every mounted cell
+  // re-renders; its `onLayout` reports the new height (nothing sets
+  // `enforcedHeight` for a vertical list — LinearLayoutManager only enforces
+  // width), and the layout manager recomputes from the lowest changed index. There
+  // is no `estimatedItemSize` in v2 to keep in sync, and `getItemType` would be
+  // actively wrong here: the tier is global, so every item shares one type at any
+  // instant and a density-keyed type would add no estimate accuracy while throwing
+  // away the whole recycling pool on each switch — fresh mounts of every visible
+  // row on the app's largest memory consumer.
   const renderClimbItem = useCallback(
     ({ item: climb }: { item: Climb }) => (
       <ActiveAwareClimbListRow
@@ -1395,6 +1411,7 @@ function ClimbListInner() {
         showFavorite
         showMoreButton={quickActionsButtonEnabled}
         surface="climbs_list"
+        density={rowDensity}
       />
     ),
     [
@@ -1408,6 +1425,7 @@ function ClimbListInner() {
       openAddToPlaylist,
       handleAddToQueue,
       quickActionsButtonEnabled,
+      rowDensity,
     ],
   );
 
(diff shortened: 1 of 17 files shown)


marcodejongh:
Rebased the density changes onto the repaired accessibility parent and published 6bb86912588fb5ff1a7da8f44be88a0ed5f69e08.

Addressed the review: spacing comes from shared tokens, all thumbnail dimensions share one metrics source, and initial/pagination skeletons follow Compact/Standard/Rich. Preference write errors are reported without disabling the current selection. The app now uses Avatar → Settings → Climb list; the test plan corrects the old More-tab directions from the declined QA report.

Validation: 11,332 mobile tests across 1,002 files; focused regressions reproduced the skeleton issue before its fix; iOS/Android/browser bundles; all 53 typecheck tasks; lint/format; independent source review; mandatory hook. The historical QA decline remains intact and needs a fresh device verdict. This stays draft while the new head completes CI and shipped-runtime preview publication.

PR #5228, opened 2026-09-06
docs: specify Kilter live activity and plan bidirectional integration

marcodejongh:
## Summary

Documents the live board activity contract from the installed Google Play Android 2.10.1 / build 65: authenticated REST reads, debounced display publication, and the separate moderation endpoint. The spec retains one provenance line, confidence markers, binary anchors, and explicit server-side unknowns.

Plans both viewing through each user's own Kilter account and contribution of eligible Boardsesh display events. Covers serial-less imported wall identity, explicit bindings, board merges, actor attribution, consent, traffic budgets, ambiguous writes, and terms/account risks. Records that the existing push-back architecture currently stops at `pushNotWired()` before HTTP calls.

Validation: `vp check` passed with zero errors; `vp run typecheck` passed; local Markdown links and diff whitespace checked. No application binaries are committed and no upstream live activity was written.

## Release Notes

## Test plan

1. CI green.

## Risk

Risk: 1/5 — Documentation only; implementation and production rollout remain gated by protocol verification.

### DIFF
--- a/docs/KILTER_HTTP_API_SPEC.md
+++ b/docs/KILTER_HTTP_API_SPEC.md
@@ -1,7 +1,7 @@
 # Kilter HTTP API Specification
 
 **Covered version**: Kilter Board mobile app, current as of 2026-05-23
-**Sibling docs**: [KILTER_POWERSYNC_SPEC.md](KILTER_POWERSYNC_SPEC.md), [AURORA_BLUETOOTH_PROTOCOL_SPEC.md](AURORA_BLUETOOTH_PROTOCOL_SPEC.md)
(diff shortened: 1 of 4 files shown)


marcodejongh:
When a user with a linked Kilter account connects to the boards history/presence websocket subjetc, we should start periodically polling the live entry of kilter through its rest api, we should then make sure that we merge what kilter returns into our own representation of history, which I believe if I remebmer correctly lives in Redis.
We need to make sure we dont just dump what we get from kilter on top.

We'll then poll the live endpoint periodically, until there are no more users connected to that subject that have a kilter account linked.

We'll only sync one way, since we dont have agreement with kilter on syncing back up

marcodejongh:
Closing as already applied. Both commits are patch-equivalent to changes now on main, and an actual rebase onto current main leaves this branch empty. The Kilter live protocol specification and integration planning are preserved on main.

PR #5232, opened 2026-09-07
feat(mobile): measure the board-account link funnel

marcodejongh:
## Summary

Stacked on #5231.

Nothing in the product emitted an event, a person property, or even a log line when a
climber linked a board account. So "how many people link their Kilter or Tension
account, and how many try and fail?" was unanswerable — which is awkward, because that
is exactly the question behind the Discord report that started this work.

Adds three events to `@boardsesh/analytics` and wires them from the Connected apps
screen:

- `Board Account Link Started` — on submit, not on opening the dialog. Someone who
  opens the form and closes it never tried, and counting that as a start would
  understate the success rate of the people who did.
- `Board Account Linked`
- `Board Account Link Failed` — carries a `reason`, mapped straight from the existing
  `BoardAccountError` codes.

All three carry `boardType` and `source`. The `source` split is the point: a climber who
navigated to Settings on purpose is not comparable to one we interrupted, and the later
surfaces in this stack (the empty logbook, onboarding) report their own.

**These are deliberately not the existing `Integration Connected` event.** That one means
an external platform — Apple Health, Strava — and it feeds the
`integrations_connected_count` person property. Folding board accounts into it would
silently change what that number means.

### Two things worth a look

**`onError` now takes `variables`.** Without it the failure event couldn't say which
board the attempt was for, which makes a funnel split by `boardType` useless. The test
harness's `useMutation` stub was dropping that second argument, so it was invisible —
the stub now matches the real signature, verified against the installed
`@tanstack/query-core` 5.101.4 type rather than assumed.

**No person property in this PR.** The plan called for a `board_accounts_linked_count`
alongside the events. Wiring it into `party-profile-provider` would add a REST call to
every cold start for every user just to populate an analytics field, and setting it
opportunistically from the Connected apps screen would only ever cover people who
visited that screen — biased in exactly the direction that makes it useless. The events
answer the funnel questions; sizing the already-linked population is a database
question, better asked directly. Flagging rather than quietly dropping it.

## Release Notes

## Test plan

1. CI green.

## Risk

Risk: 1/5 — analytics only. No user-visible change and no behaviour change.

### DIFF
--- a/packages/mobile/src/components/integrations/BoardAccountsSection.tsx
+++ b/packages/mobile/src/components/integrations/BoardAccountsSection.tsx
@@ -16,6 +16,12 @@ import { useMutation, useQuery, useQueryClient } from '@tanstack/react-query';
 import { useTranslation } from 'react-i18next';
 import type { TFunction } from 'i18next';
 import { boardTypeLabel } from '@boardsesh/board-constants';
+import {
+  trackLinkFailed,
+  trackLinkStarted,
+  trackLinkSucceeded,
+  type BoardLinkFailureReason,
+} from '../../lib/integrations/board-link-analytics';
 import { DUPLICATE_BOARD_ACCOUNT_CIRCUITS_SYNC_ERROR } from '@boardsesh/shared-schema/sync-error-codes';
 import {
   AURORA_BOARDS,
@@ -101,6 +107,12 @@ const AURORA_UNSYNCED_QUERY_KEY = ['auroraCredentials', 'unsynced'] as const;
 
 // MoonBoard isn't an Aurora board, so it has no credential/sync flow.
 const MOONBOARD_SUPPORT_EMAIL = 'moonboardsupport@moonclimbing.com';
+
+// This screen is the one surface that offers linking today, so its funnel `source`
+// is fixed. Later surfaces (the empty logbook, onboarding) report their own, which
+// is the whole point of splitting the funnel by source — a climber who came here
+// deliberately is not comparable to one we interrupted.
+const LINK_SOURCE = 'integrations' as const;
 
 // Brand names, not a capitalised slug: `charAt(0).toUpperCase()` renders `soill`
 // as "Soill", which shipped a mangled trademark on the So iLL card and into every
@@ -249,6 +261,13 @@ async function saveBoardCredential(input: { boardType: AuroraBoardName; username
   return saveAuroraCredential(input);
 }
 
+// The funnel's `reason`. A non-`BoardAccountError` is a thrown network/parse
+// failure rather than a code the server sent, so it reports as `request_failed` —
+// the same bucket its user-facing copy falls into below.
+function failureReasonFor(error: unknown): BoardLinkFailureReason {
+  return error instanceof BoardAccountError ? error.code : 'request_failed';
+}
+
 function errorMessageFor(error: unknown, t: TFunction<'settings'>): string {
   if (error instanceof BoardAccountError) {
     switch (error.code) {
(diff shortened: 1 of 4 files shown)


marcodejongh:
Restacked onto repaired parent #5231 at `c7dc8e5b9` and addressed the remaining review findings in `7f79b8cb7f7314e537e52d21f47f856bbe2816e6`. The component tests use the real `BoardAccountError` constructor for invalid credentials and already-linked accounts; network errors retain `request_failed`. Removed the never-emitted cancellation reason and shortened the added comments. The parent’s canonical board names, white inputs, scoped So iLL fixture and truthful sync timing remain intact.

Independent paired review approved the five-file delta. All 11,306 mobile tests, 22 focused component tests, full lint, all 53 typecheck tasks, iOS/Android bundles and the commit hook pass. A check with installed React Query 5.101.4 confirms that a failed post-save refetch still produces only Linked, because these invalidation calls do not enable throwOnError. No credential values enter analytics properties, and no native input or production analytics configuration changed.

This PR stays draft while current-head CI and a new preview matching the shipped app are verified.

marcodejongh:
Restacked onto the parent’s final test-title correction at `d970d1b0b`; current head is `5432a2fc74c9e9d5279be30fc12d529066b4dcb8`. Both telemetry commits retain identical stable patch IDs. The only tree change from the prior reviewed head is that one test-group title; product code, fixtures and assertions are unchanged. The focused component suite passes all 22 tests and full 53-task typecheck passes again; the prior full mobile and bundle verification still applies to the unchanged implementation.

For the optional Kilter coverage note: the existing Kilter card test exercises `saveKilterCredentialViaPassword`, and both Kilter and Aurora return through the same `saveCredentialMutation` success/error callbacks. Event properties take the submitted `variables.boardType`, with no separate provider-specific telemetry branch. The typed terminal reasons and no-submit behavior are covered in the new funnel cases.

marcodejongh:
The latest optional alias note does not indicate an omitted failure case. `BoardAccountErrorCode` explicitly includes `request_failed`, `failureReasonFor` maps an ordinary network Error to that value, and the component regression asserts the corresponding failed event. Keeping the shared union also prevents the analytics type from drifting from the credential client’s supported error codes.

marcodejongh:
The current shared union already includes `request_failed`: [aurora-credentials.ts](https://github.com/boardsesh/boardsesh/blob/ca41a2f1fd4437a884820413feba48f49048e231/packages/mobile/src/lib/aurora-credentials.ts#L39) lists it explicitly. Therefore `BoardLinkFailureReason = BoardAccountErrorCode` and the event-catalog comment cover that outcome; adding the same literal again would not widen the type. The real network-error regression asserts `request_failed`, all 22 component cases pass, and full TypeScript checking passes. Both telemetry commits are unchanged after the parent locale restack.

marcodejongh:
Checked these against the current source:

- `BoardLinkFailureReason` is a direct alias of `BoardAccountErrorCode`, rather than a separately maintained enum. It intentionally uses that same vocabulary, including `request_failed`. There is no second analytics code list to drift.
- `failureReasonFor` forwards `error.code` for every `BoardAccountError`, without a whitelist or switch. The existing two typed-code cases exercise that branch; the untyped-error case exercises the fallback. Adding an unrecognized value outside the constructor’s type would not test a separate implementation branch.
- Keeping that classification beside the credential form lets the analytics helper retain a type-only dependency on the credential module. The extraction in #5234 preserves the same classification in the dialog rather than adding a runtime credential dependency to analytics.

The latest parent-only restack preserves both reviewed telemetry commits exactly, with 118 locale/component cases and all 53 typecheck tasks passing.

marcodejongh:
Reviewed the type concern: BoardAccountErrorCode already includes 'request_failed', so BoardLinkFailureReason and the event contract cover the network fallback. The component helper deliberately adapts the component's credential errors to that shared analytics contract; both save paths share the same success/error callbacks. Existing focused tests cover 22 cases, including the link funnel.

The restack onto the reviewed #5231 web-label follow-up preserves both own patches unchanged. Full typecheck and lint passed. Exact-head CI and publisher 35664284202 are green.

PR #5234, opened 2026-09-07
refactor(mobile): extract the board-account link dialog

marcodejongh:
## Summary

Stacked on #5233. Pure refactor — no behaviour change, no new strings.

Lifts the board-account sign-in dialog out of `BoardAccountsSection` (1640 lines) into
`LinkBoardAccountModal`, which owns its own form state, mutation, error mapping and
funnel reporting. A host now only says which board and which surface.

Three surfaces want this dialog: Connected apps today, the empty-logbook prompts from
#5233, and the onboarding step that follows. Leaving it inline would mean the
Kilter-vs-Aurora routing (`kilter` goes through the password grant, everything else posts
to the Aurora endpoint), the `BoardAccountError` → copy mapping, and the Started/Linked/
Failed events all existed in two or three places. The failure mode of that drift is
specific and nasty: a linking flow that works in Settings and silently misroutes
elsewhere.

`errorMessageFor` moves with it and is imported back for the unlink mutation, so there is
exactly one mapping from an error code to what the climber reads.

### Dead code removed along the way

The extraction stranded a fair amount in the section, all deleted: the `username` /
`password` state, `saveBoardCredential`, `failureReasonFor`, the `inputStyle` /
`inputBackground` / `inputBorder` trio and its `iosDarkColors` import, the `TextInput`
import, and the three analytics imports. Net −149 lines there, +229 in a file that three
callers share.

### How this is verified

The existing 29 integration tests pass **unchanged**. They drive the dialog through
`BoardAccountsSection` — opening it, typing, submitting, asserting the Kilter password
grant is used and that the funnel events fire with the right board — so they exercise the
extracted component end to end without knowing it moved. That is the proof the refactor
preserves behaviour; a test file edited alongside a refactor proves much less.

## Release Notes

## Test plan

1. Profile → More → Connected apps.
2. Tap Link on Tension. The form opens.
3. Enter a wrong password. An error shows, form stays open.
4. Tap Cancel, then Link on another board. Fields are empty.

## Risk

Risk: 2/5 — moves a dialog between files with no behaviour change, covered by the
existing suite unchanged.

### DIFF
--- a/packages/mobile/src/components/integrations/BoardAccountsSection.tsx
+++ b/packages/mobile/src/components/integrations/BoardAccountsSection.tsx
@@ -7,7 +7,6 @@ import {
   Platform,
   ScrollView,
   StyleSheet,
-  TextInput,
   View,
 } from 'react-native';
 import { File } from 'expo-file-system';
@@ -16,12 +15,7 @@ import { useMutation, useQuery, useQueryClient } from '@tanstack/react-query';
 import { useTranslation } from 'react-i18next';
 import type { TFunction } from 'i18next';
 import { boardTypeLabel } from '@boardsesh/board-constants';
-import {
-  trackLinkFailed,
-  trackLinkStarted,
-  trackLinkSucceeded,
-  type BoardLinkFailureReason,
-} from '../../lib/integrations/board-link-analytics';
+import { LinkBoardAccountModal, errorMessageFor } from './LinkBoardAccountModal';
 import { DUPLICATE_BOARD_ACCOUNT_CIRCUITS_SYNC_ERROR } from '@boardsesh/shared-schema/sync-error-codes';
 import {
   AURORA_BOARDS,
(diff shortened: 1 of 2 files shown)


marcodejongh:
Rebased this extraction onto repaired #5233 at `7b436be6c`; published head `0f79242a1`. All three original review notes are addressed: both consumers import credential/cache keys from the existing shared credential-query module, the dialog documentation is concise, and `failureReasonFor` is private. The extraction preserves the current white inputs and contrast, canonical board labels, truthful sync copy, typed errors and attempt/result telemetry. New regressions cover closing/reopening, switching boards, and waiting for both cache invalidations before the host callback.

All 11,328 mobile tests and both native Metro bundles passed before the parent-only restack. Exact range/tree proofs preserve the two reviewed own commits; after restacking, 43 combined focused tests, full lint and all 53 typecheck tasks pass. The mandatory hook passed. This remains draft while fresh exact-head CI and a preview matching the shipped runtime complete.

marcodejongh:
The unlink mutation still uses `errorMessageFor` in `BoardAccountsSection.tsx:267`: its `onError` shows the translated typed account error through that shared helper. See [the current head](https://github.com/boardsesh/boardsesh/blob/0f79242a129453cd13990af2b2e3c7310f5bb587/packages/mobile/src/components/integrations/BoardAccountsSection.tsx#L262-L268). The import/export therefore remains necessary after the linking mutation moved into the dialog. `failureReasonFor`, which has no external caller, is already private.

marcodejongh:
The three observations do not require a source change. The submit path closes the modal for the mutation the user explicitly started; the host cannot switch the modal's board selection while that modal is presented, and the awaited invalidations complete before `onLinked` advances the host flow. `errorMessageFor` remains used by `BoardAccountsSection` for unlink errors, while the link error paths continue through the extracted modal and are covered by the existing host/component error regressions. The current exact head is green and has no unresolved review thread.

PR #5235, opened 2026-09-07
feat(mobile): offer the board-account link during first-run

marcodejongh:
## Summary

Stacked on #5234. Behind `board-link-onboarding-step`, off by default.

The last piece: after a new climber picks their board, offer to link their board
account. This is what the Discord report asked for — the user signed in with Google,
assumed his Tension sends would follow, and had no reason to think otherwise.

The card leads with what they get, states the correction as a fact about the system
rather than about them ("Your Tension account lives with the people who make the board.
Boardsesh is separate — same email, but the two don't know about each other"), and puts
the crux in the footnote **before** they hit a field they can't fill: *linking asks for
your Tension username, not your email.*

### It sits before the board-look step, and that is forced rather than chosen

`app/onboarding.tsx` forbids chaining past `BoardLookStepGate` — that gate's refusal to
run without a synced climb and a renderer probe is exactly what makes a no-exit step
safe, and chaining past it would throw the guarantee away. Board-look may also never run
at all. Sitting here also means the board type is warm, so the card can name the wall
they picked ninety seconds ago rather than asking again.

The offline-download dialog is awaited inside `onBound`, so the two interruptions never
overlap.

### The one escapable step in a flow that was just made mandatory

#5024 closed the last onboarding skip five days ago, so this needs justifying rather than
sneaking in. Steps 1–3 guarantee state the app cannot run without — a bound board, a
chosen drawing. This one asks for a password to somebody else's service, and nobody may
be compelled to type that. It also asks a question that is **false for some people**:
there is no lookup-by-email anywhere in Aurora's API, so we cannot detect whether an
account exists. We can only ask.

So it deliberately does **not** call `useBlockBack`. The route sets `gestureEnabled:
false` for the whole file, so skippable had to mean a real visible button; Android
hardware back routes to the same handler as "Not now" and counts as an answer.

### Who never sees it

`shouldOfferLink` is a pure function, tested without a renderer, mirroring
`decideBoardLookStep`. It returns `none` for: the flag off, **MoonBoard** (no credential
flow exists — its only route in is a CSV obtained by emailing Moon Climbing a GDPR
subject access request, so "a few minutes" would be a lie), offline, already answered, and
already linked. The credential read is tri-state: unresolved is `wait`, never `show`, or a
climber who linked months ago gets a first-run card.

Anything other than `show` goes to Climbs, including `wait`. This is a one-shot moment
with no second chance, and stalling first-run on a storage read is a worse trade than
skipping an optional card — the empty-logbook prompt from #5233 catches exactly that case.

### Measurement

`Onboarding Link Prompt Shown` / `Resolved {linked | declined | abandoned}`. Every Shown
resolves to exactly one Resolved, including the nav-aways no button produced — without
the unmount guard this would repeat the hole `onboarding-analytics.ts` documents, where
~a third of tour Starts resolved to nothing and deflated the completion metric. The link
attempt itself is not re-counted here; it reports through `Board Account Link*` with
`source: 'onboarding'`, so its success rate is directly comparable to Settings.

**Guardrail before raising the flag:** `Onboarding Board Activated` per `Onboarding Tour
Started`, split by flag cohort. Roll back on a visible drop, or if the decline rate
exceeds ~85% (which would mean the card is irrelevant to almost everyone).

### A note on the storage

The answered marker is AsyncStorage, not SecureStore, following the rule
`board-look-step-seen.ts` spells out: a marker belongs in the same store, with the same
lifecycle, as the thing it records. On iOS the keychain survives an uninstall while the
sandbox does not, so a SecureStore marker would outlive every other trace of the account —
a climber who reinstalled and still hadn't linked would never be asked again. Re-asking
after a reinstall is the right failure; silently never asking is not. It is written on an
**answer**, never on arrival, so a force-quit mid-step leaves the question live.

## Release Notes

New climbers now get asked, right after picking their board, whether they want to bring
their Kilter or Tension sends across — with a plain explanation that your board account
and your Boardsesh account are two different logins, even on the same email. Skip it and
nothing changes.

## Test plan

1. Sign in as a new climber. Pick a Tension board.
2. A card offers to link your Tension account.
3. Tap Not now → Climbs opens.
4. Repeat with a MoonBoard. No card appears.
5. Relaunch. The card does not come back.

## Risk

Risk: 4/5 — adds a step to first-run behind a flag that is off by default. Onboarding
conversion is the metric at risk; guardrail and rollback trigger above.

### DIFF
--- a/packages/mobile/app/onboarding.tsx
+++ b/packages/mobile/app/onboarding.tsx
@@ -4,6 +4,7 @@ import { router, useLocalSearchParams } from 'expo-router';
 import { useTheme as usePaperTheme } from 'react-native-paper';
 import { OnboardingPrompt } from '../src/components/onboarding/OnboardingPrompt';
 import { OnboardingBoardRoute } from '../src/components/onboarding/OnboardingBoardRoute';
+import { OnboardingLinkRoute } from '../src/components/onboarding/OnboardingLinkRoute';
 import { BoardLookStep } from '../src/components/board-look/BoardLookStep';
 import { useTheme } from '../src/providers/theme-provider';
 import { useVariantValue } from '../src/theme/variants';
@@ -67,6 +68,17 @@ export default function OnboardingScreen() {
 
   if (step === 'board-look') {
     return <BoardLookRoute accentColor={accentColor} bodyColor={bodyColor} backgroundColor={backgroundColor} />;
+  }
+
+  if (step === 'link') {
+    return (
+      <OnboardingLinkRoute
+        accentColor={accentColor}
+        iconColor={iconColor}
+        bodyColor={bodyColor}
+        backgroundColor={backgroundColor}
+      />
+    );
   }
 
   if (step === 'board') {
--- a/packages/mobile/src/components/onboarding/OnboardingBoardRoute.tsx
+++ b/packages/mobile/src/components/onboarding/OnboardingBoardRoute.tsx
@@ -1,4 +1,4 @@
-import { useCallback, useMemo } from 'react';
+import { useCallback, useEffect, useMemo, useRef, useState } from 'react';
 import { router } from 'expo-router';
 import type { UserBoard } from '@boardsesh/shared-schema';
 import { OnboardingBoardStep } from './OnboardingBoardStep';
@@ -14,6 +14,10 @@ import { useBoardOfflineState } from '../board-discovery/use-board-offline-state
 import { trackNudgeAccepted, trackNudgeDismissed, trackNudgeShown } from '../../lib/offline-nudges/nudge-analytics';
 import type { NudgeEventContext } from '../../lib/offline-nudges/nudge-analytics';
 import { offlineBoardKeyForBoard } from '../../settings';
+import { shouldOfferLink } from '../../lib/onboarding/should-offer-link';
+import { hasAnsweredLinkStep } from '../../lib/onboarding/link-step-answered';
+import { useBoardAccountCredentials } from '../../lib/integrations/use-board-account-credentials';
+import { useFeatureFlag } from '../../providers/feature-flags-provider';
 
 // Module-level so an absent board list keeps a stable identity — a fresh `[]`
 // per render would rebuild the carousel's items on every commit.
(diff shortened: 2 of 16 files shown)


marcodejongh:
Restacked onto repaired #5234 and published `94be93b4aa190f4bd1c599e1948baab259e3900c`. Both original review notes are addressed: French onboarding consistently uses informal tu, and prompt events are grouped with their own comment before the credential-attempt events. All four locales now promise only available sends with variable timing/history.

The repair also fixes two reproduced defects: Android Back now records one declined outcome (or closes an open credential dialog first), and a second board tap cannot overwrite an activation already waiting for its download offer. Direct navigation respects the positive rollout flag; a failed answer-marker write is reported without blocking exit.

Independent source and documentation review approved the final change. All 11,365 mobile tests, 72 focused cases, full lint, all 53 typecheck tasks, iOS/Android Metro bundles and the mandatory hook pass. The concrete QA plan includes the real Replay walkthrough entry and an unlinked, never-answered installation prerequisite. This stays draft pending current-head CI and a compatible shipped-runtime preview. No production rollout, data clearing or human QA verdict is claimed.

marcodejongh:
Addressed the concrete documentation and coverage gaps in the final restack:

- The flag description and documentation now describe all supported Aurora boards: Kilter, Tension, Decoy, Touchstone, Grasshopper and So iLL. MoonBoard retains its separate import flow. The QA route uses Avatar → Settings → Replay walkthrough.
- Direct answer-marker tests cover missing/true/false values, read failure, screenshot mode, and successful/failed persistence. The stable-presentation test rerenders the mounted card and verifies one shown event and one terminal outcome.
- I retained the current event lifecycle after checking the installed renderer and analytics client: development effect replay occurs in ReactFabric-dev, while PostHog is disabled by `!__DEV__` and `getPostHogClient()` returns null there. A real remount is a new presentation. A local shown ref would neither survive that remount nor safely preserve the effect cleanup outcome pairing. Existing development-disabled analytics tests were reused.

The full mobile suite passes 11,373 tests; the seven focused files pass 101 cases. No event/storage implementation or production flag was changed. The follow-up has independent source review; current-head CI and a compatible published preview remain the readiness gates.

marcodejongh:
Addressed in f16b3a3eb44559af1ac889a861778a8f82b6b4de. The shared activation hook now offers linking after binding/download for carousel, picker and builder, checks the latest eligibility, and leaves ordinary board selections unchanged. Exiting linking dismisses the underlying picker/builder stack to Climbs. Spanish copy uses encadenes. Independent review approved the source and test isolation changes; 11,380 mobile tests, native/browser bundles, all 53 typecheck tasks and lint passed. All five own commits survived the final restack onto #5234 unchanged. Fresh CI and shipped-runtime preview are pending; physical first-run/keyboard/Back QA is still listed in the PR.

marcodejongh:
Both observations match deliberate bounded contracts. Eligibility is account-level: the documented gate requires a **known empty credential list**, and the answered marker is likewise global to this optional first-run step. It is not a per-board reminder; after any Aurora account is linked, normal Settings account management remains available for another board family. A Kilter credential therefore intentionally suppresses a later Tension onboarding prompt. The activation hook reads the latest full credential list after binding/download, as covered by its shared eligibility tests. Also, `boardType` is fixed for one mounted link route; a different selected board creates a new route presentation. Resetting `resolvedRef` with that route input preserves one terminal analytics outcome per presentation and does not re-arm on ordinary rerenders.

PR #5244, opened 2026-09-07
fix: show Woods climbs' real rules everywhere, not just in search

### LINKED ISSUE #5214: 🐞 Bug report: Since the feet and matching rules were implemented, all the climbs display “matching rule not recorded...

## Comment

Since the feet and matching rules were implemented, all the climbs display “matching rule not recorded” and “foot rule not recorded”. I’m not sure exactly what to make of this. It appears as though even the climbs that do allow matching or any feet display the same thing, so maybe the kinks just need to get worked out. The original all just had every set climb as either matching allowed or no matching, and any feet or tracking. Although for tracking feet, there was nothing displayed

## Metadata

| Field | Value |
| --- | --- |
| Platform | ios |
| App version | 2.4.0 (13) |
| Source | drawer-bug |
| Board | woods / layout 1 / size 2 / sets [1] @ 40° |
| Climb | The very very very very very Hardest Henry Route  (4a/V0) |
| Session | unknown |
| URL | /climbs |
| User agent | unknown |
| Feedback row | app_feedback #10244 |

## Contact

✅ Reporter agreed to follow-up — resolve **app_feedback #10244** in the DB to reach them.

marcodejongh:
## Summary

Woods climbs showed **"Matching rule not recorded · Feet rule not recorded"** under every name in the play drawer. The September 2026 production rule repair did land — the catalogue rows carry their rules — but five "full climb" GraphQL surfaces never selected `board_climbs.characteristics`, and `SimilarClimb` had no such field at all.

Verified against production before the fix, same climb both times:

| probe | `characteristics` |
| --- | --- |
| `climb(woods/1/2/"1"/40, "eebea34a-…")` — "Stipend" | `["no_match"]` ✅ |
| `setterClimbsFull(username:"AstroGrey", woods/1/2/"1"/40)` | `null` ❌ |

The list surfaces are not a lesser view of a climb: mobile opens the play drawer with the payload they hand back, no refetch. So the missing column was a missing feature on screen — and not only on Woods. `Climb.is_no_match` is a derived resolver that falls back to the Aurora description convention whenever the array is absent, which is a no-op on Woods and MoonBoard, so every one of these surfaces also dropped the ⊘ no-match glyph on Kilter and Tension.

Fixed in four resolver projections (`userFavoriteClimbs`, `hydrateClimbsByRefs` → `playlistClimbs` both modes + `smartPlaylist`, `setterClimbsFull` both modes, `userClimbs`), the five client documents, and end-to-end on `SimilarClimb`. That SDL field is a **nullable** list on purpose, unlike `compatibleSizeIds` beside it: `decodeClimbRules` treats `null` (never recorded) and `[]` (set under all the defaults) as different answers.

**Why it recurred.** `resolver-frames-fields.test.ts` guarded this exact bug class one field earlier, and has now been outgrown three times — `framesCount`/`framesPace`, then `compatibleSizeIds` (threaded through with no guard at all), now `characteristics`. It is a table now, one row per entry point and one list of required fields, and there is a matching client-side guard on the documents — the half the user actually sees, since a perfect projection is invisible if nothing asked for it.

**Author-side verification.** Every guard assertion was mutation-tested: dropping any single projection, mapper line, raw-SQL half or document field turns the matching test red. The first pass at the raw-SQL guard stayed green with the column removed (the fixture supplied the value either way), so it now reads the statement back and checks both the `owned_climbs` CTE and the outer SELECT. `setter-follows-integration.test.ts` seeds a real `text[]` and reads it back, proving the column survives against Postgres rather than a mock. Full mobile suite 9,023 passed; web 4,517 passed (one unrelated gym-admin flake that passes in isolation); the nine backend files touching this code path pass in isolation. The local full backend run has widespread pre-existing `users` FK / TRUNCATE-deadlock flakes under this box's IO, in unrelated files (apns, push-tokens, board-merge, board-presence) — CI is the arbiter.

**Not in this PR.** Two more play-drawer entry points build a `Climb` from a slim payload type that has no `characteristics` field at all, and show the same wrong text on Woods: the Logbook / session feed (`tick-to-climb.ts`) and a notification tap (`notification-to-climb.ts`). Both need new schema fields on the tick and notification payloads, which widens this into the ticks and notifications subsystems, and the guard here cannot cover them. Tracked in #5245.

Also documented in `docs/woods-climb-rules.md`: the climb-search cache holds results for 24 hours with no invalidation hook, so a repair that changes stored rules without changing code serves the old ones for a day. That is most likely what the reporter saw on the `/climbs` list itself, whose own query was correct all along.

Closes #5214

## Release Notes

Woods climbs show their real matching and feet rules everywhere you open them — playlists, favourites, a setter's climbs, your own climbs and similar climbs.
No-match climbs get their ⊘ back in those lists on every board.

## Test plan

1. Woods board, 40°. You tab → Climbs → tap a climb.
2. Rules line reads real rules, never "not recorded".
3. Discover → a Woods playlist → tap a climb. Same.
4. In the drawer, Similar Climbs → tap one. Same.
5. Switch to Kilter: no-match climbs show ⊘ in lists.

## Risk

Risk: 3/5 — one new GraphQL schema field plus five resolver projections; no BLE, no migrations, no data writes.

https://claude.ai/code/session_01DQDZf8wLBJybB8KwLrRjqC


packages/shared/graphql/src/generated/graphql.ts
```diff
@@ -10146,6 +10159,8 @@ export type GetSmartPlaylistQuery = {
       boardseshDifficulty?: number | null;
       boardseshConfidence?: string | null;
       compatibleSizeIds?: Array<number> | null;
+      characteristics?: Array<string> | null;
+      is_no_match?: boolean | null;
```
marcodejongh, line 10163: Why do we have a is_no_match flag, when these characteristics also end up in the characteristics array?

PR #5270, opened 2026-09-07
fix: stage verification for requests reaching the web origin

marcodejongh:
## Summary
The public Railway web hostname lets crawlers bypass Cloudflare and trigger rendering directly. Add staged origin verification across all routes, stripping the secret before downstream handlers and external rewrites. Only GET/HEAD `/api/health` is exempt when enabled.

Add a narrowly scoped, manually dispatched Cloudflare request-header rule and authenticated normal/rollback deployment smokes. Redirects cannot carry the secret to another origin. Server and edge Sentry hooks redact the incoming credential before error, transaction, span and structured-log export. The flag stays disabled until the stripping middleware, secret and Cloudflare header are deployed in that order; the runbook includes rollback and rotation.

Validation: full repository typecheck, full lint (0 errors) and 238 focused plus 25 telemetry tests passed, including POST-body preservation, previously excluded routes, missing/wrong secrets, header stripping and Cloudflare ownership/error redaction.

The origin and crawler branches also merge cleanly; all 378 combined tests passed.

## Release Notes
none

## Test plan
1. Open Boardsesh → homepage loads.
2. Open a shared climb → climb details appear.
3. Sign in → your account appears.

## Risk
Risk: 3/5 — incorrect activation could block www; default-disabled staging and rollback are documented.

### DIFF
--- a/.github/workflows/cloudflare-origin-protection.yml
+++ b/.github/workflows/cloudflare-origin-protection.yml
@@ -1,0 +1,34 @@
+name: Cloudflare origin protection
+on:
+  workflow_dispatch:
+    inputs:
+      apply:
+        description: Apply the www request header after the stripping middleware is deployed
+        type: boolean
+        default: false
+permissions:
+  contents: read
+concurrency:
+  group: cloudflare-origin-protection
+  cancel-in-progress: false
+jobs:
+  configure:
+    runs-on: ubuntu-latest
+    environment: Production
+    timeout-minutes: 5
+    steps:
+      - uses: actions/checkout@d23441a48e516b6c34aea4fa41551a30e30af803 # v6
+      - uses: actions/setup-node@v4
+        with:
+          node-version: '22.x'
+      - name: Configure origin header
+        env:
+          CLOUDFLARE_API_TOKEN: ${{ secrets.CLOUDFLARE_API_TOKEN }}
+          WEB_ORIGIN_VERIFY_SECRET: ${{ secrets.WEB_ORIGIN_VERIFY_SECRET }}
+          APPLY: ${{ inputs.apply }}
+        run: |
+          if [ "$APPLY" = true ]; then
+            node scripts/cloudflare-origin-apply.ts --apply
+          else
+            node scripts/cloudflare-origin-apply.ts
+          fi
--- a/packages/web/app/lib/web-origin.ts
+++ b/packages/web/app/lib/web-origin.ts
@@ -1,0 +1,26 @@
+/** Shared secret is server-only; never use a NEXT_PUBLIC_ variable for it. */
+export const WEB_ORIGIN_HEADER = 'x-boardsesh-origin-verify';
+
+export function acceptsWebOrigin(request: Request, pathname: string): boolean {
+  if (process.env.WEB_ORIGIN_VERIFY_ENABLED !== '1') return true;
+  // Railway probes only this cheap endpoint; no prefix or locale exceptions.
+  if (pathname === '/api/health' && (request.method === 'GET' || request.method === 'HEAD')) return true;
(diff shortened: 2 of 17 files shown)


packages/web/app/lib/web-origin.ts
```diff
@@ -0,0 +1,26 @@
+/** Shared secret is server-only; never use a NEXT_PUBLIC_ variable for it. */
```
marcodejongh, line 3: If its server only, we should also import server only

marcodejongh: Added `import 'server-only'` and the existing Vitest mock pattern to affected tests. All 255 affected tests and the full repository typecheck, including the production middleware/edge compilation, passed.

marcodejongh:
Addressed the static-request overhead, secret-format mismatch, documentation link and missing disabled-value coverage. After origin authorization, non-page requests without a verification header now return `NextResponse.next()` without cloning headers or constructing another request. Requests carrying the header still get an explicit sanitized override; protected static routes still reject missing credentials before this fast path. The middleware now requires the same 64 lowercase hexadecimal characters as the existing Cloudflare helper and rollout runbook.

The comparison comment now explicitly says JavaScript offers no constant-time guarantee. A factual correction to the arithmetic note: bitwise operators convert `NaN` to zero, so `NaN ^ 120` is `120`, not `NaN`; the length difference still rejects mismatched lengths. No constant-time security claim is made.

The dot exclusion already exists in base `61f2514c1`'s matcher (`.*\..*`). It remains the established locale/session/CORS scope. New tests prove a mid-path decimal remains outside page middleware while still receiving origin verification and header stripping. Changing which existing routes receive locale/session handling would be a separate routing-policy change.

Thirteen new regressions failed before the fixes; 236 focused web/middleware/redaction tests now pass, together with full lint/format, all 53 typecheck tasks and the commit hook. Independent paired review approved the six-file follow-up. No Cloudflare apply, secret provisioning, deployment or activation occurred.

marcodejongh:
Checked the two version notes against the official upstream sources. [Node 22.18.0](https://nodejs.org/en/blog/release/v22.18.0), released July 31, 2025, enabled type stripping by default, so the workflow’s current `node-version: 22.x` supports executing this erasable TypeScript script without the experimental flag. The script imports only Node’s built-in `node:url`; it does not require an installed `tsx` dependency. No workflow dispatch or Cloudflare operation was needed to verify this.

The official [checkout v6 tag](https://github.com/actions/checkout/tree/v6) resolves to `d23441a48e516b6c34aea4fa41551a30e30af803`, exactly the workflow pin; [v6.0.0](https://github.com/actions/checkout/releases/tag/v6.0.0) is an official release. The `# v6` annotation is correct. Neither factual version note requires a code change.

.github/workflows/cloudflare-origin-protection.yml
```diff
@@ -0,0 +1,34 @@
...
+  group: cloudflare-origin-protection
+  cancel-in-progress: false
+jobs:
+  configure:
+    runs-on: ubuntu-latest
+    environment: Production
+    timeout-minutes: 5
+    steps:
+      - uses: actions/checkout@d23441a48e516b6c34aea4fa41551a30e30af803 # v6
+      - uses: actions/setup-node@v4
```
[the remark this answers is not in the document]
marcodejongh: Pinned in `5a14c26df` to `actions/setup-node@49933ea5288caeca8642d1e84afbd3f7d6820020`, the official v4 tag commit and the same SHA already reviewed in production-deploy.yml. Node22 inputs and credential handling are unchanged. Full lint/format, all53 typecheck tasks and the commit hook passed; independent paired review approved the one-line change. No Cloudflare operation or production workflow dispatch occurred.

PR #5330, opened 2026-09-08
feat(mobile-screenshots): record and replay backend traffic during captures

marcodejongh:
## Summary

Store captures now replay sanitized backend responses from a checksum-pinned dev-bucket snapshot, with a frozen clock. Both capture workflows default to replay, and fail when the app requests an unrecorded response or a fixture no longer matches the current GraphQL schema.

- Publishes a sanitized snapshot from the dedicated production test account: 130 GraphQL responses recorded across Android, iPhone, and both iPads on September 18. Other climbers’ names, handles, emails, and avatars are replaced before saving; credentials are excluded.
- Stores only a small snapshot reference in Git; 131 recorded fixture files stay untracked. Captures, standalone replay, and mobile test setup automatically fetch the immutable 741 KB snapshot into an ignored cache. Downloaded and cached bytes are verified; missing fixtures fail instead of skipping drift checks.
- Integrates replay with the merged iOS probe/fan-out workflow. Recording keeps its artifacts private, and replay does not require production credentials.
- Bounds screenshot pagination and seeds workout selection so capture timing does not change visible content.
- Fixes a board-switch race discovered during local Android replay: workout previews no longer show the previous board’s climbs under the newly selected board. Same-board option changes retain their preview while rebuilding.

Restores the probe/fan-out and baseline files from merged #5326 that the later wireframe merge #5552 removed; its web changes remain intact.

Stacked on #5556 for the shared dev object-store uploader.

Verification: all nine snapshot integrity/cache tests, all 127 capture/backend contract tests, all 31 fixture drift/privacy checks and all 12 workout preview tests pass, including a regression that failed before the board-switch fix. Full workspace typecheck and both native Metro bundle checks pass. The final integrated mobile suite passes all 11,130 tests. Local Android replay captured all eight screens with no missing responses; renderer, frozen-clock, content, and dimension gates pass.

## Release Notes

Switch boards without briefly seeing your previous board’s workout climbs.

## Test plan

1. Signed in with two boards: Record → choose workout; climbs appear.
2. Switch boards; preview climbs update to match your selected board.
3. Change workout options; existing rows remain while the preview updates.

## Risk

Risk: 4/5 — changes capture authentication and replay inputs; strict gates reject missing responses.


marcodejongh:
All three addressed, in `579e5ad0` (code + tests + docs) and `5ac682ce` (the rewritten fixture set).

### Blocker — real user PII in committed fixtures

Fixed at the source rather than by pruning the feed: the recorder now pseudonymises every climber but the recording account **before a response is ever written**, and the committed set was rewritten with the same function.

- `pseudonymiseResponse` — `scripts/lib/screenshot-fixtures.ts:686`, rule table at `scripts/lib/screenshot-fixtures.ts:364`. A field counts as personal only when it sits on an object that also carries the id key naming whose field it is (`userId` / `ownerId` / `ownerUserId` / `creatorId` / `authorId` / `id` / `uuid`, each with its own name, handle, avatar and email keys). Nothing matches on a name alone, so a climb's `name`, a playlist's `name`, a board's `name`, a gym's `name` and a session's `sessionName` are never touched. Groups are ordered, so `displayName` on `{ id, userId, displayName }` is the commenter's, not the comment's.
- Replacements: display name → `<Adjective> <Noun>` from a fixed 32×32 word list, chosen by a stable hash of the person's id, so one climber reads as the same person in every fixture and the feed stays coherent; handle → `climber_<6 hex>`; avatar / profile image URL → `null` (every one is a nullable `String` in `packages/shared-schema/src/schema/`, and the app renders initials); email → `climber_<6 hex>@pseudonymised.invalid`. A beta link's `foreignUsername` has no id in the payload, so the handle itself is the identity it hashes from.
- Applied at record time by default — `scripts/lib/screenshot-backend.ts:510` — logging `PSEUDONYMISED graphql <Op> <hash12> persons=<n>`, which is never a problem line. `--no-pseudonymise` is the record-only escape hatch for a set that will not be committed.
- The committed set was rewritten with `vp run mobile:screenshot-fixtures-pseudonymise` (`scripts/screenshot-fixtures-pseudonymise.ts`, idempotent, `--check` to report only): **11 of 130 fixtures, 56 real climbers, 107 fields** — Discover playlist creators (60), the "Everyone" feed's participants (21), recent-beta Instagram handles (20), 4 avatar URLs, one board owner. 107 insertions / 107 deletions; every changed line is a personal field. A fixture is keyed by its document and variables, never its response, so no hash, filename or manifest entry moved — replay hits exactly the fixtures it hit before. No static avatar remained to drop (all 16 recorded assets are beta-link thumbnails), so `manifest.json` is byte-identical.
- Guard so it cannot regress: `packages/mobile/src/lib/graphql/__tests__/screenshot-fixture-drift.test.ts:513` fails on any committed fixture holding a real person's name, handle, avatar or email, naming the file and the field. Mutation-checked — editing one `displayName` back to `"Xin Wei Chow"` fails it with that exact path; restored, green.
- Deliberately NOT rewritten, documented in `docs/mobile-screenshot-fixtures.md:295`: `setterUsername` / `setter_username` / `faUsername` (Aurora's public attribution on the **climb**, rendered on boardsesh.com's public climb pages, and the object it sits on carries the tick logger's id, not the setter's — there is no id to hash a stable pseudonym from); a beta link's `link` / `thumbnail` (a public post URL carrying no handle, and the static fixture is keyed on its shortcode); bare user ids (opaque uuids, and what the pseudonym hashes from). Our own account and the all-zero "owned by Boardsesh" id stay real.

Proof: grepping every real display name, creator name, owner name and Instagram handle from the pre-rewrite survey, as whole JSON string values, returns nothing across `packages/mobile/screenshot-fixtures/`.

### Minor — inconsistent pager guard

`PlaylistDetailView` now wraps its end-reach handler in `screenshotModeLoadMore` like Discover and the profile beta shelf, instead of the inline `process.env` check — `packages/mobile/src/components/playlist/PlaylistDetailView.tsx:268`. The helper returns one shared no-op in screenshot mode, so the memoized handler keeps a stable identity. `playlist-detail` suite green (40 tests), including both page-cap cases.

### Minor — fixture git history

`docs/mobile-screenshot-fixtures.md:644`, "Refresh cadence and history": re-record only when the drift test demands it or a store screen changes — not for a stale-looking date; the merge clears `graphql/`, `static/` and `manifest.json` before writing, so a refresh is a diff over ~5 MB rather than another 5 MB beside it, and most fixtures come back byte-identical; old blobs stay reachable in history and that is accepted, since the set has to be versioned WITH the code that replays it and nothing in it is a secret (tokens are refused before a fixture is written, and everyone but the recording account is now a stand-in).

Validation: `vp check` 0 errors, all files formatted · `vp run typecheck:scripts` and `typecheck:mobile` clean · `vp test run --project scripts` 2326 passed (112 files) · mobile `screenshot` 96 passed, `screenshot-fixture-drift` 31 passed, `playlist-detail` 40 passed.

marcodejongh:
The "Do not merge" review above repeats the pre-`5ac682ce4` finding; it is resolved on the current head:

- Every third-party person in the committed set is pseudonymised at `579e5ad03` (record-time rule, `PSEUDONYMISED_PERSON_FIELDS` in `scripts/lib/screenshot-fixtures.ts`) and the set was rewritten in place at `5ac682ce4`: 56 real climbers across 11 fixtures → stable `<Adjective> <Noun>` names keyed by user id, handles → `climber_<hash>`, avatars → `null`.
- Verification on `5ac682ce4`: the only `displayName` values left in `GetSessionGroupedFeed` are the pseudonyms and `Test User` (our own screenshots account); the only non-null `avatarUrl` values anywhere in the set belong to that same account (`fbb773f8-…`), which is deliberately kept real. The drift test now fails on any fixture holding a non-pseudonymised person (`screenshot-fixture-drift.test.ts`, mutation-checked).
- The two minors (`PlaylistDetailView` pager guard → `screenshotModeLoadMore`; refresh cadence and history paragraph in `docs/mobile-screenshot-fixtures.md`) landed in the same commit.

PR #5331, opened 2026-09-08
ci(mobile-screenshots): run captures after each native deploy that shipped a binary

marcodejongh:
## Summary

Store screenshots run automatically after a successful native deploy on `release/next` that shipped a binary. iOS captures one probe first, then captures and uploads the complete set only when screenshots changed or the probe cannot safely represent the change. Manual capture remains available from any branch.

- Pins captures to the deployed commit and skips deployments that reused an existing binary.
- Keeps replay free of production credentials; record/live credentials are scoped to the capture step.
- Separates probe and recovery artifact names. If baseline fetching or comparison fails after upload, the full matrix can recapture that pair, and finalization excludes the failed probe’s artifacts.
- Keeps Android commit-back dispatch-only and gates App Store uploads on a complete validated capture.
- Keeps screenshot tooling failures best-effort after creating a store draft; removes obsolete credential fallbacks and resolves documentation conflicts.

Verification: 40 workflow contract tests pass, including late-probe recovery; full workspace typecheck passes. The integrated pipeline captured all eight Android screens locally from sanitized fixtures. Automatic `workflow_run` behavior requires a native deploy after merge for live verification.

Stacked on #5330 until its replay changes merge.

## Release Notes

## Test plan

1. CI green.

## Risk

Risk: 2/5 — automatic store uploads require a shipped binary and complete validated screenshots.


marcodejongh:
Addressed all three points, rebased on the latest `feat/screenshot-probe-fanout` head (`0d3615dd4`), new PR6 head `2ee76af7b`:

**(a) Contract test's exact job-order equality — fixed.** Commit `2ee76af7b` (`test(mobile-screenshots): pin the required jobs without forbidding extra ones`), `scripts/__tests__/mobile-screenshots-workflows.test.ts:174-186`. The test now `toContain`s each of the five required job names individually, then compares only those names (filtered out of the full job list) against the required order — so a later debug/reporting job appended anywhere won't fail it, while a missing or misspelled job still will.

**(b) Empty-candidate-directory test coverage gap — already covered.** Landed on #5326 at `0d3615dd4`, `scripts/__tests__/compare-screenshots.test.ts:285-303` (`'reports changed with every baseline file missing when the candidate directory is empty'`). It exercises exactly the Maestro-crash-before-any-PNG path you flagged: an existing-but-empty candidate directory reports every baseline file `missing` rather than reading as unchanged.

**(c) Credential fallback (`secrets.SCREENSHOT_USER_EMAIL || 'test@boardsesh.com'`) — kept intentionally, not changed.** This isn't a stray test credential; it's the documented PROD screenshots account (`docs/mobile-screenshot-fixtures.md`) the whole capture pipeline is built around — App Store screenshots are captured signed in as that account by design, not as a fallback-of-convenience. #5330 (companion PR, fixture record/replay) additionally scopes these credentials to the capture step only, and only outside replay, so a replay run doesn't touch the account at all. Failing loudly on a misconfigured repo would break every automatic post-deploy capture and every dispatch, since the secret is deliberately unset in most environments — the account being test-only is the whole point, not a risk to fence off.

https://claude.ai/code/session_01VjmyzX7wp1cgksH1zweWWA

marcodejongh:
Addressed the fail-closed note from the latest re-review in commit \`8fd7820a2\`:

Both screenshot workflows' first job (\`setup\` on iOS, \`gate\` on Android) now runs a "Refuse an automatic capture with no screenshot credentials" step, gated on \`github.event_name == 'workflow_run'\`, that fails the whole run with \`::error::\` before either secret's value is ever read if \`SCREENSHOT_USER_EMAIL\`/\`SCREENSHOT_USER_PASSWORD\` is unset. The dispatch fallback (\`|| 'test@boardsesh.com'\` / \`|| 'test'\`) is unchanged — a human dispatching a capture by hand is the documented intent to use the shared test account — and a one-line comment next to each fallback now notes that an automatic run never actually reaches it.

Contract test (\`scripts/__tests__/mobile-screenshots-workflows.test.ts\`, "\$platform fails closed on an automatic run with no screenshot credentials") pins the guard step's \`if:\`, both secret references, and the exact \`::error::\` text for both platforms. Mutation-tested: removing the guard step from either workflow fails that platform's test.

https://claude.ai/code/session_01VjmyzX7wp1cgksH1zweWWA

marcodejongh:
On the latest note (missing `APP_STORE_CONNECT_API_KEY_BASE64` in the attach step): the attach step runs in `mobile-store-draft.yml`'s `ios` job, which already decodes the key in its "Decode App Store Connect API key" step (`APP_STORE_CONNECT_API_KEY_BASE64` → `$ASC_KEY_PATH` via `GITHUB_ENV`, lines 336-348 on the base branch `feat/screenshot-probe-fanout`) before the attach at line 437, and cleans it up at line 474; `ASC_KEY_PATH` reaches fastlane through the job env, so the attach step does not need the base64 secret itself. #5326's next commit adds an explicit `[ -f "$ASC_KEY_PATH" ]` guard in the attach step so a moved or skipped decode fails visibly instead of being swallowed by `continue-on-error`.

marcodejongh:
Addressed both items from the latest bot pass, in two commits on the rebased head (\`b9f247350\`):

**\`e71e15f34\` (on #5326, pulled in by this rebase)** — \`fetchBaseline\` now skips (with a \`::warning::\`) any downloaded release zip that \`parseAssetName\` can't decode, or that the baseline manifest doesn't list, instead of extracting it or failing the whole fetch over it.

**\`b9f247350\`** — closes the documented-but-accepted race from earlier reviews: a dispatch that can write to a shared target (\`upload\`/\`publish_baseline\` on iOS, \`commit_to_main\` on Android) now joins the same serial concurrency group as automatic runs, instead of getting its own per-run group. A capture-only dispatch (the common case) is unaffected — it still gets a per-run group since it never writes. Both workflows' header trade-off paragraphs are updated to say the race is closed, not just tolerated. Also added the missing assertion the bot flagged: a dedicated contract test pins that \`FULL_SET_CAPTURED\` opens on \`needs.ios-capture.result == 'success' &&\`, so it cannot read true on the partial/unchanged path where \`ios-capture\` was skipped.

Both mutation-tested (collapsing the write-condition back to a bare per-run group, and dropping the \`ios-capture\` result clause, each fail the relevant test) and validated: \`vp check\`, YAML parse, \`typecheck:scripts\`, and the full \`scripts\` project (115 files / 2349 tests) all green.

https://claude.ai/code/session_01VjmyzX7wp1cgksH1zweWWA

Issue #5293, opened 2026-09-08
iOS crash: Fabric SchedulerDelegate use-after-free drained on the JS thread (BOARDSESH-8S)


marcodejongh:
## Symptom

EXC_BAD_ACCESS on `com.facebook.react.runtime.JavaScript` inside `facebook::react::RuntimeScheduler_Modern::updateRendering`. The app dies and relaunches fine, so nobody has reported it as feedback.

## Evidence

BOARDSESH-8S. 28 events / 27 distinct users, `environment: production` on 28/28, iOS on 28/28. ~15 device models (iPhone12,8 → iPhone18,4, iPad13,6), iOS 17.6.1 → 26.6.1, ~14 countries.

The shape is an escalation, not a flat rate: roughly 2 events in the first seven weeks after firstSeen 2026-07-03, then 28 between 2026-08-21 and 2026-09-07, with 12 of those in the final four days as `2.4.0+13` rolled out. Rate went from ~0.04/day to ~1.6/day and is rising with adoption of the current build.

The faulting addresses are the recycled block's first 8 bytes, which is why they take three shapes: wild 64-bit garbage (0x3147dda7976b84a2), tiny integers (0x40, 0x4b, 0x69, 0x7a), and one whose bytes decode little-endian to ASCII `}nicode6` — the SQLite FTS5 tokenizer name `unicode61`. A pointer value that is literally a recycled C string is a use-after-free, not a null deref and not memory pressure.

## Root cause

react-native 0.86.3, `ReactCommon/react/renderer/scheduler/Scheduler.cpp:322-336`: `uiManagerDidFinishTransaction` queues a rendering update onto the RuntimeScheduler capturing `delegate = delegate_`, a raw `SchedulerDelegate*` (`Scheduler.h:125`). The same unsafe capture is at :356-374 (`uiManagerDidDispatchCommand`). The third site, :443-449, captures `uiManager` as a shared_ptr and is safe — which is what makes the delegate capture the specific defect rather than a general pattern.

The lambda's only protection is `if (guardEnabled && *invalidated) return;`, gated on `ReactNativeFeatureFlags::enableSchedulerDelegateInvalidation()`, which returns false (`ReactNativeFeatureFlagsDefaults.h:206`) because nothing in the app or in Expo ever calls `ReactNativeFeatureFlags::override()`. With it off, neither `~Scheduler` (:180) nor `setDelegate` (:274) ever sets the invalidation token.

[...]

PR #5354, opened 2026-09-09
fix(mobile): [native-train] promise.reject(code, description) keeps its message

### LINKED ISSUE #5296: iOS: every promise.reject(code, description) reaches JS as "<CODE>: undefined reason (at ExpoModulesCore/Promise.swift:65)"
## Symptom

On iOS, no native module rejection carries its message. The reason is stored and then thrown away by the renderer, so JS gets a fixed string with the wrapper's own file and line. Downstream, every classifier that reads the message falls to its default.

## Evidence

A Sentry aggregate on `message:"*undefined reason*"` over 90 days returns four issues, two of which are not even ours: BOARDSESH-FK `ERR_UPDATES_FETCH` (12 events / 4 users) and BOARDSESH-DJ `ERR_UPDATES_CHECK` (4 / 3) come from expo-updates. This is an app-wide bridge defect, not a BLE bug. `git grep 'promise.reject("' origin/main -- packages/mobile/modules` returns 10 two-arg sites: BoardBleModule (5), HealthWorkoutsModule (3), LiveActivityModule (2).

Highest-cost instance is the iOS connect path, BOARDSESH-CB: 81 events / 67 users over 14 days, all `environment: production`, including 21 events / 19 users on `2.4.0+13`, the newest iOS binary. Not fleet lag. firstSeen 2026-08-04, steady daily through 2026-09-07, so not a burst.

The event proves the chain itself: BOARDSESH-CB carries `failure_category: unknown` and `level: error`, while correctly-classified siblings (7J, CS, D3) sit at `warning` — exactly what `bleConnectReportLevel` (`use-board-bluetooth.ts:74-78`) predicts.

PostHog measures the consequence: over 14 days, iOS `connect_failed` on `Bluetooth Connection Failed` is **1 event**, while iOS `write_failed` on that same connect event is **138 events / 67 users**. iOS connect attribution is corrupt.

## Root cause

expo-modules-core 57.0.14 (pinned in `packages/mobile/package.json`, no patch in `patchedDependencies`, so upstream source is authoritative):

- `ios/Core/Promise.swift:65` — `rejecter(Exception(name: code, description: description, code: code))`.
- `ios/Core/Exceptions/Exception.swift:31-36` — that initializer assigns `self.description`. It does **not** assign `reason`.
- `Exception.swift:55-57` — `debugDescription` renders `"\(name): \(reason) (at \(origin.file):\(origin.line))"`, and `reason` is the base class's `"undefined reason"` (:9-11). `origin` defaults to the wrapper's own `#fileID`/`#line`, i.e. Promise.swift:65, not the caller's.
- `ios/JS/ExpoRuntimeInstaller.swift:86-89` states the trap outright: "The default `JavaScriptThrowable.message` is `String(reflecting: self)`, which on `Exception` resolves to `debugDescription`". Expo works around it there by overriding `message`, which is proof the trap is known upstream.

So the message is stored in `.description` and never rendered.

`BoardBleModule.swift:131`/`:159` do pass `error.localizedDescription`. It just never survives.

Downstream chain: `classifyBleFailure` matches no ladder and returns `unknown` → `bleConnectReportLevel` escalates to Sentry `error` → the alert falls to the default branch `bluetooth.unknownError` ("Couldn't connect to your board. Try again.") instead of `bluetooth.connectFailed` ("…Move closer to the board and try again.") → `use-board-bluetooth.ts:1704` does `failureCategory === 'unknown' ? classifyBleFailureReason(error) : failureCategory`, and that classifier's default (`connection-error.ts:305`) is `write_failed`, which is what PostHog records. The `service_missing` branch (`use-board-bluetooth.ts:1646`) is unreachable on iOS, so #3480's `getLastConnectDiagnostics` is dead there.

## Suggested fix

An `Exception` subclass that overrides `reason` (or `message`) so the description renders. Not "stop relying on the description" — the description is already being passed correctly.

Swift change, so it is a native-fingerprint input: `[native-train]` draft, and per CLAUDE.md it needs a Fable or Astra review before merge.

## Scope note

BOARDSESH-CK is **not** evidence for this. Its `failure_reason: write_failed` is already correct and its `bleWriteDiagnostics` survived intact (`watchdogTripped: true`, withResponse, 5293 ms, moonboardCharacteristic). It is an iOS MoonBoard write stall belonging to #3235/#3926. Including its 9 events here would pad the impact and invite a close as "that's the MoonBoard write thing".

Strip `user_cancelled` from any BLE rate you quote — it is roughly two thirds of every `Bluetooth Connection Failed` row (Sept: 621 of 903). See #4714.

## Related

- #4143 — this blocks its scope item 2. Re-landing another description string will not make iOS diagnosable while the bridge eats it.
- #3480 — its iOS diagnostics are unreachable because of this.
- #3676, #3608 — both assumed the native error string was all JS had.

Filed from the 2026-09-08 Sentry triage sweep (BOARDSESH-CB, BOARDSESH-CK).

## Backlog triage — 2026-09-20

Fresh impact, production native PostHog, UTC [2026-09-13, 2026-09-20): iOS Bluetooth Connection Failed with write_failed recorded 91 events / 40 unique person IDs; [2026-08-21, 2026-09-20): 310 / 140. This is an attribution proxy, not proof all events share the bridge defect. Sentry BOARDSESH-CB independently has 52 events / 40 users in its rolling seven-day window at Sep 20 04:24 UTC; do not add the sources together. P1 is a scheduling judgment for core connection diagnosis/recovery, not evidence every underlying failure is caused by this wrapper. Sources: https://us.posthog.com/project/412845/insights/new and https://boardsesh.sentry.io/issues/BOARDSESH-CB .

Priority: priority:P2 → priority:P1. Parent telemetry review reports 52 recent Sentry events / 40 users in seven days on iOS connect-path promise rejection; prioritize core connection diagnosis. Wrapper signature can span underlying causes; do not conflate raw event count with distinct connection failures.

marcodejongh:
## Summary

Every native `promise.reject(code, description)` on iOS reached JS as `"<CODE>: undefined reason (at ExpoModulesCore/Promise.swift:65)"`. The description was silently dropped. This is an **upstream expo-modules-core bug**, not a call-site mistake on our side.

**Root cause, verified against the installed source** (`node_modules/expo-modules-core@57.0.14/ios/Core/Exceptions/Exception.swift`):
- `Promise.reject(_ code: String, _ description: String)` builds `Exception(name: code, description: description, code: code)`.
- That initializer assigns `self.description = description`, but never touches `reason`.
- `Exception.debugDescription` (which backs the default `JavaScriptThrowable.message`) renders `"\(name): \(reason) (...)"`, and `reason` is a hardcoded `"undefined reason"` unless a subclass overrides it. The two-arg initializer builds a plain `Exception`, so it never does.
- Not fixed by a newer version: the same code is still on `expo/expo`'s default branch. The upstream issue, [expo/expo#49677](https://github.com/expo/expo/issues/49677), was bot-closed for lacking a minimal repro, not resolved.
- Our own `promise.reject(code, description)` call sites (`BoardBleModule.swift` ×5, `HealthWorkoutsModule.swift` ×4, `LiveActivityModule.swift` ×2) use the documented two-arg overload. There is no other overload we should have used.

**Fix**: a `pnpm patch` on `expo-modules-core@57.0.14`'s `Exception.swift`. It adds a private `explicitReason: String?`, set from `init(name:description:code:)`, and `reason` falls back to it before `"undefined reason"`. The SDK's own `ReadOnlyExpoModulesPropertyException` (in `ExpoRuntimeInstaller.swift`) overrides `reason` per subclass for the same reason; this applies it to the base class so every two-arg rejection benefits, including `expo-updates`' own `ERR_UPDATES_FETCH` / `ERR_UPDATES_CHECK`.

### Making the patch reach the binary (round 3 of Fable review)

The first version applied to `node_modules` and iOS CI went green, but **the patch was never compiled into the app**. `expo-modules-core` ships its own prebuilt `ExpoModulesCore` xcframework, and Expo's precompiled modules are on by default. The earlier iOS CI log (run 34297020756) listed `📦 ExpoModulesCore (57.0.14)` as precompiled and contained `Exceptions/Exception.swift` 0 times.

- **`expo-modules-core` added to `expo.autolinking.buildFromSource`** in `packages/mobile/package.json`, following the rule from 97bfd8029 (every patched native pod goes in `buildFromSource`).
- **Cascade, and why it is required.** `resolve_prebuilt_status` in `expo-modules-autolinking/scripts/ios/precompiled_modules.rb` returns `dependency_unavailable` for any precompiled pod whose dependency builds from source. ExpoImage, ExpoFont, ExpoImageManipulator and others declare `expo-modules-core/ExpoModulesCore` in their `spm.config.json` `externalDependencies`, so they fall back to source too. That is what we want: the patch adds a stored property to `open class Exception`, so every subclass in those pods has to compile against the patched layout. In the new run, pod install prints `Precompiled modules enabled but no xcframeworks found. All modules will build from source.`
- **iOS CI now proves it.** "Assert the patched native module really compiled" in `ios-rn-ci.yml` now also requires a `SwiftCompile` line naming `expo-modules-core/ios/Core/Exceptions/Exception.swift` and `expo-image/ios/ImageView.swift`, not just `RNSScrollViewFinder`.
- **Linux guard.** `vp run check:mobile-patches` now fails when a rule edits an `ios/` file of a package that is not built from source. A package counts as built from source if it is listed in `buildFromSource`, or if it depends through its installed `spm.config.json` on a package that is (the same `dependency_unavailable` cascade).

### expo-image patch: dead on iOS before this PR, live after

`patches/expo-image@57.0.3.patch` (#3928, PR #4301) edits `localAssetName(from:)` in `ios/ImageView.swift`. It returns `nil` early for `file://` URLs that have a host or an absolute path. Without it, those URLs get their leading slash stripped and are passed to `UIImage(named:)`, which does a synchronous bundle scan on the main thread and always misses. Relative xcasset names such as `app_icon` and `Images/MyIcon` are unchanged.

ExpoImage was precompiled, so this patch never reached a binary. #4301's own iOS CI runs (31545368262, 31856635728) and this PR's earlier run list `📦 ExpoImage` and have 0 compile lines for `expo-image/ios/ImageView.swift`. Because of the cascade above, it compiles from source starting with this PR (run 34754317177 has the `SwiftCompile … expo-image/ios/ImageView.swift` line).

This is a behaviour change: the #3928 fix ships for the first time. I think it's safe. The only inputs it changes are URLs that `UIImage(named:)` could never resolve: after the slash is stripped, the lookup misses and loading already falls through to SDWebImage. One edge is a root-level `file:///app_icon` URL, which used to be treated as an asset name and no longer is. I found no `file:///` literal in `packages/mobile/src`, `app` or `modules`. No device check has been done on this.

### Fingerprint and BLE

This changes native Swift inside a native dependency and changes `packages/mobile/package.json`'s autolinking config, so it **moves the native fingerprint** on both platforms and triggers TestFlight / Play-internal builds from `main`. It can't reach the current store fleet by OTA.

**BLE-adjacent.** The patch changes the message that every `BoardBleModule.swift` rejection (`BLE_SCAN_FAILED`, `BLE_CONNECT_FAILED`, `BLE_WRITE_FAILED`) carries to JS. The regex classifiers in `packages/shared/ble-protocol/src/connection-error.ts` were written for these `BoardBleError` strings but could never match on iOS while every message read "undefined reason". On iOS this restores the behaviour described in #5296:
- **Connect**: `classifyBleFailure` sorts into `unavailable` / `board_not_found` / `connect_failed` / `service_missing` instead of `unknown`, `bleConnectReportLevel` stops over-escalating to Sentry `error`, and the Alert becomes the specific `bluetooth.connectFailed` copy instead of `bluetooth.unknownError`.
- **Write**: `write_timeout` / `write_recovery_failed` populate, and `isDisconnectionError` can match a CoreBluetooth "disconnected" message. That runs `handleDisconnection` right away, and it is adapter- and generation-guarded, so it is idempotent with the native `disconnected` event.

**Fable review history.** Round 1 found the first patch didn't compile (`explicitReason` assigned after the `open lazy var` setters); fixed in `60ad5f893`. Round 2 compiled it under `swift:6.0.3` and approved. Round 3 (at `60ad5f8`) was **blocking**: the patch never reached the binary. Fixed in `cb5094a8d`. Waiting on a fresh Fable review.

Fixes #5296

## Testing

**iOS CI compiles the patch now** (run 34754317177, job 103716283830, head `cb5094a8d`):
- The assertion step prints `ExpoModulesCore Exception.swift compiled from source — patch exercised.` and `ExpoImage ImageView.swift compiled from source — patch exercised.`
- The build log names `<pnpm>/node_modules/expo-modules-core/ios/Core/Exceptions/Exception.swift` in a batched `SwiftCompile normal arm64` line. The previous run's log had it 0 times.
- Pods with ExpoModulesCore-dependent targets now compile: ExpoModulesCore 85 compile lines, ExpoFileSystem 80, ExpoLocation 66, ExpoImage 62, ExpoModulesWorklets 50, ExpoMaps 47, ExpoFont 44, ExpoImageManipulator 31.
- The ABI check passes: `[mobile-abi] OK — 127 cross-framework symbol(s) resolve across 5 embedded binaries.` Swift unit tests pass.

**Build time** (both runs had a cold Pods cache):

| Run | Build-the-app step | Whole job | CompileC lines | SwiftCompile lines |
|---|---|---|---|---|
| 34297020756 (`60ad5f8`, precompiled) | 16m25s | 25m55s | 1307 | 1004 |
| 34754317177 (`cb5094a8d`, source) | 14m16s | 21m00s | 1577 | 1095 |

The compile volume goes up by 270 CompileC lines and 91 SwiftCompile lines. The wall-clock change is inside runner noise: another PR's run the same morning (#5293, precompiled) took 11m44s. The TestFlight Release archive time was not measured.

**CI assertion, mutation-tested on the real log** from job 102296938664. The step's script was run against that log copied as `build.log`:

```
RNScreens RNSScrollViewFinder.mm compiled from source — patch exercised.
::error::ExpoModulesCore Exception.swift was not compiled — the patch was not exercised. Check whether a prebuilt xcframework replaced the pod (buildFromSource in packages/mobile/package.json).
::error::ExpoImage ImageView.swift was not compiled — the patch was not exercised. Check whether a prebuilt xcframework replaced the pod (buildFromSource in packages/mobile/package.json).
exit=1
```

With the two compile lines appended it printed all three "patch exercised" lines and exit=0. The `@expo/ui` `ImageView.swift` compile lines in that log (2) don't match the expo-image pattern. Run 34754317177 is the real green.

**Linux guard, mutation-tested.** With `expo-modules-core` removed from `buildFromSource`:

```
 FAIL  __tests__/mobile-patches-check.test.ts > the shipped buildFromSource list > builds every ios/-patched pod from source, so each patch reaches the binary
+   "expo-image: patch edits ios/ImageView.swift, but nothing builds expo-image from source. …",
+   "expo-modules-core: patch edits ios/Core/Exceptions/Exception.swift, but nothing builds expo-modules-core from source. …",
 FAIL  … > builds expo-image from source only through the expo-modules-core cascade
      Tests  2 failed | 74 passed (76)

[mobile-patches] FAILED — native patch(es) not applied:
  ✗ expo-image: patch edits ios/ImageView.swift, but nothing builds expo-image from source. …
  ✗ expo-modules-core: patch edits ios/Core/Exceptions/Exception.swift, but nothing builds expo-modules-core from source. …
```

After restoring it, all 76 tests pass, `[mobile-patches] OK — 12 native patch(es) applied, 8 patch file(s) accounted for.`, and the tree is clean.

**Earlier rounds.** The `Exception.swift` source guard was mutation-tested against the literal upstream file (FAILED, then OK after restoring). Fable compiled the patched `Exception.swift` under `swift:6.0.3`, and CI now compiles it under Xcode 26.6.

**Not verified.** Runtime behaviour on a device: the BLE Alert copy, the Sentry message text, and the expo-image `file://` change. Release archive build time.

## Release Notes

iOS: clearer heads-up when your board won't connect — points you at what's actually wrong (move closer, try again) instead of a generic error.

## Test plan

1. CI green.

## Risk

Risk: 5/5 — native dependency patch that moves the native fingerprint, switches ExpoModulesCore and its dependents from prebuilt to source builds, and changes every BLE rejection message; also turns on the dormant expo-image #3928 patch.

https://claude.ai/code/session_013BYuvLtqeNTQQUJqA7zbog


marcodejongh (commented):
**Fable review (BLE / native fingerprint)** — PR #5354 at `60ad5f8`, independent third pass.

**Verdict: BLOCKING.** The Swift fix is correct, but on the build this repo actually ships it never compiles: the iOS app links `ExpoModulesCore` as a prebuilt xcframework, so the patched `Exception.swift` is not in the binary. Merging as-is costs a native-train (both fingerprints move) and leaves Sentry reading `undefined reason` exactly as before.

### Blocking

**1. The patch is applied to `node_modules` but not to the app — ExpoModulesCore is precompiled.**

- `expo-modules-core@57.0.14` ships its own prebuilt: `prebuilds/output/{debug,release}/xcframeworks/ExpoModulesCore.tar.gz` (present in the installed package). `ExpoModulesCore.podspec:118` only sets `source_files = 'ios/**/*.swift'` when `try_link_with_prebuilt_xcframework(s)` returns false; `expo-modules-autolinking/scripts/ios/precompiled_modules.rb:120-129` turns it on whenever `EXPO_USE_PRECOMPILED_MODULES=1`, which the generated Podfile sets by default (`.github/workflows/ios-rn-ci.yml:168`, `scripts/mobile-framework-abi-check.ts:13` both document this).
- This PR's own iOS CI run proves it (run 34297020756, job `build-and-test` 102296938664, head `60ad5f8`). The pod-install summary reads `[Expo-precompiled] Precompiled modules:` … `📦 ExpoModulesCore (57.0.14)`, alongside `RNScreens: building from source (configured by buildFromSource)`. The string `Exceptions/Exception.swift` appears **0 times** in the 93,544-line build log; no `CompileSwift` for ExpoModulesCore ran at all. The build was green because it compiled none of the patch.
- `packages/mobile/package.json:129` `expo.autolinking.buildFromSource` lists `react-native-screens`, `@expo/ui`, `expo-dev-launcher` — the exact rule commit `97bfd8029` wrote down ("pin every patched native pod to `buildFromSource`") — and this PR does not add `expo-modules-core`.

Concrete failure: merge → TestFlight/Play builds from `main` → iOS binary links the unpatched `ExpoModulesCore.xcframework` → `BLE_CONNECT_FAILED: undefined reason (at ExpoModulesCore/Promise.swift:65)` keeps arriving, `classifyBleFailure` keeps returning `unknown`, and the store fleet took a fingerprint bump for nothing. `check:mobile-patches` cannot catch this: it reads the installed source (`scripts/mobile-patches-check.ts:519-534`), which *is* patched; the binary is what isn't.

Fix (three parts, all in this PR):

1. Add `"expo-modules-core"` to `expo.autolinking.buildFromSource` in `packages/mobile/package.json`. Know the cost before you do: `resolve_prebuilt_status` (`precompiled_modules.rb:1944-1985`) propagates `dependency_unavailable`, so every precompiled pod that depends on ExpoModulesCore falls back to source too — from the log that is ExpoFileSystem, ExpoFont, ExpoImage, ExpoImageManipulator, ExpoLocation, ExpoMaps, ExpoModulesWorklets. That cascade is *required*, not a nuisance: the patch adds a stored property to `open class Exception`, so every subclass in those pods must be compiled against the patched layout. Don't fight it. iOS build time goes up; by how much is unverified here. (Setting `EXPO_USE_PRECOMPILED_MODULES=0` in the iOS workflows is the same outcome with one env var — either is fine, but `buildFromSource` is the pattern the repo already documents.)
2. Extend `ios-rn-ci.yml:176` "Assert the patched native module really compiled" so it also greps `build.log` for `Core/Exceptions/Exception.swift` (a `SwiftCompile` line naming the file). Today that step only asserts `RNSScrollViewFinder`; it would have gone green on this PR.
3. Optional but cheap: in `scripts/mobile-patches-check.ts`, for every `RULES` entry whose `file` starts with `ios/`, assert the package is in `buildFromSource` (or that the package ships no `prebuilds/output`). That turns this class of false green into a Linux-runner failure, which is what that guard exists for.

### Adjacent, pre-existing (not caused by this PR; file it separately)

`patches/expo-image@57.0.3.patch` edits `ios/ImageView.swift`, and the same log lists `📦 ExpoImage (57.0.3)` as precompiled. The only `ImageView.swift` compile lines in the log are `@expo/ui`'s. So that patch is very likely dead in the shipping binary today (unverified beyond the log). Note that fix step 1 above will pull ExpoImage to source via the cascade and *revive* that patch — a behaviour change worth a sentence in the PR body.

### Verified, no findings

- **Mechanism, against the installed source.** `Exception.swift:9-11` hardcodes `reason`; `:32-37` sets `description` only; `:55-58` `debugDescription` renders `reason`. `Promise.swift:64-66` builds `Exception(name:description:code:)`. `expo-modules-jsi …/Protocols/JavaScriptThrowable.swift:28-31` defaults `message` to `String(reflecting:)` → `debugDescription`; `JavaScriptError.swift:42-51` puts `message` + `code` on the JS `Error`. With the patch the JS message is `BLE_CONNECT_FAILED: Bluetooth connection timed out (at ExpoModulesCore/Promise.swift:65)` — the wrapper's file:line still rides along (cosmetic; no user-facing surface prints `error.message`, all Alerts use i18n copy).
- **Patch integrity.** `git apply --check` of the PR's patch against this worktree's *unpatched* installed `expo-modules-core@57.0.14` applies cleanly; `sha256` of the patch file = the lockfile `patch_hash` `331eed2d…`; both installed copies (react 19.2.3 / 19.2.8 peer variants) are re-keyed in `pnpm-lock.yaml`. Prior rounds compiled it under `swift:6.0.3`; I did not repeat that.
- **JS string-matchers.** Every regex in `packages/shared/ble-protocol/src/connection-error.ts` is an unanchored substring test, so the `CODE: ` prefix and `(at …)` suffix can't false-match a ladder. Walked each `BoardBleError.errorDescription` (`BoardBleManager.swift:155-187`) through the ladders: `bluetoothUnavailable`→`unavailable`, `deviceNotFound`→`board_not_found`, `connectTimedOut`→`connect_failed`, `uartServiceMissing`/`writeCharacteristicMissing`→`service_missing`; `writeTimedOut`/`writeRecoveryFailed` hit their own buckets; "No board is connected" still does not match `/not connected/`, so the MoonBoard streak logic (`use-board-bluetooth.ts:1276-1284`) is unchanged. Nothing in `packages/` matches `"undefined reason"`. `handleDisconnection` (`use-board-bluetooth.ts:855-870`) is adapter + generation guarded, so the new write-failure path is idempotent with the native `disconnected` event.
- **Old binaries / Android.** No JS change in the PR, so OTA-fleet binaries keep today's behaviour. Nothing Android in the patch and no Kotlin `promise.reject` sites in `packages/mobile/modules`; the Android fingerprint still moves because `fingerprint.config.js:123-127` hashes `../../patches` for both platforms — a Play rebuild for zero Android change, which that config already accepts as the safe direction.
- **Guard quality.** The two new vitest cases are genuinely red without the rule (the second feeds the literal upstream text and asserts `patch NOT applied`), and the PR's mutation test on the installed tree is the right shape. The gap is scope, not quality: it guards `node_modules`, not the binary — see finding 1.

Unverified: Swift compile (relying on rounds 1-2), the build-time cost of the `buildFromSource` cascade, and the expo-image patch's runtime behaviour.

marcodejongh (commented):
**Fable re-review (BLE / native fingerprint) — head 3234e9ae7**

**Verdict: approve to merge.** The blocking finding from round 3 is fixed and the fix is guaranteed by config, not by the runner. Three non-blocking notes below.

### 1. Exception.swift now compiles, and nothing precompiled is left to mix with it

Read both iOS CI logs (34754317177 at `cb5094a8d`, 34755435092 at `3234e9ae7`; the second is a comment-only change and the two logs agree line for line on everything below).

- The patched file is in a real compile line, and the path proves it is the patched copy: `SwiftCompile normal arm64 Compiling\ AnyArgument.swift,\ … /Users/runner/work/boardsesh/boardsesh/node_modules/.pnpm/expo-modules-core@57.0.14_patch_hash=331eed2d…/node_modules/expo-modules-core/ios/Core/Exceptions/Exception.swift …`. Pre-fix run 34297020756 had 0 such lines.
- Every pod the pre-fix run listed as `📦` now compiles from source (arm64 `SwiftCompile` lines per package in run 34755435092): expo-modules-core 10 batches, expo-image 28, expo-file-system 24, expo-location 18, expo-maps 13, expo-font 10, expo-image-manipulator 1; the `ExpoModulesWorklets` target compiles `expo-modules-core/ios/Worklets/*.swift` from the same `patch_hash=` path. `grep -c` for `prebuilds/output`, `artifacts/*.xcframework`, `Pods/Expo*/Expo*.xcframework` in the build phase: 0. So no consumer of `Exception` is linked precompiled against the old layout.
- **Every `[Expo-precompiled]` line still present in 34755435092** (lines 966-977, 1155-1158): `RNReanimated`, `react-native-safe-area-context`, `RNSVG`, `RNWorklets`, `RNCAsyncStorage` — each `building from source (prebuilt tarball not found)` (third-party; identical in the pre-fix run); `RNScreens: building from source (configured by buildFromSource)`; the warning `⚠️ Precompiled modules enabled but no xcframeworks found. All modules will build from source.`; and the summary `Precompiled modules:` with the single entry `⚠️ ExpoModulesCore (57.0.14) (configured by buildFromSource)`. No `📦` line remains. None of the third-party pods link `ExpoModulesCore`.
- `mobile:abi-check` on the Debug .app: `OK — 127 cross-framework symbol(s) resolve across 5 embedded binaries`. Swift unit tests pass.

### 2. Why the "no xcframeworks found" warning — config, not environment (coordinator's question)

Traced in the installed `expo-modules-autolinking@57.0.12/scripts/ios/precompiled_modules.rb` (the version `expo@57.0.18` pins in `pnpm-lock.yaml`):

1. `autolinking_manager.rb:30-33` calls `PrecompiledModules.configure(build_from_source: configuration['buildFromSource'])`, read from `packages/mobile/package.json` by the autolinking `resolve` step. `build_from_source=` (`:136`) compiles each entry to `^pattern$`.
2. `_resolve_prebuilt_status_uncached` (`:1943-1944`) returns `{available: false, reason: :build_from_source}` for a pod whose name **or npm package** matches, **before** it ever looks for a tarball. `ExpoModulesCore` and `ExpoModulesWorklets` both have `npm_package == expo-modules-core`, so both short-circuit here.
3. For every other prebuilt-capable pod, the same function checks its own tarball (`resolve_own_prebuilt_info`, `:1923`) and then walks `pod_info[:prebuilt_dependency_pods]` (`:1950-1966`); a dependency that resolved unavailable returns `:dependency_unavailable`. I read the installed `spm.config.json` of every pod the pre-fix run had precompiled: `expo-image`, `expo-font`, `expo-file-system`, `expo-image-manipulator`, `expo-location`, `expo-maps` all declare `expo-modules-core/ExpoModulesCore` in `externalDependencies`. So each resolves unavailable for the buildFromSource reason, deterministically.
4. `should_auto_patch_spec?` (`:384-387`) is `enabled? && has_prebuilt_xcframework?(name)`, and `has_prebuilt_xcframework?` (`:346-350`) is just `resolve_prebuilt_status(name)[:available]`. It is false for all of them, so `cocoapods/sandbox.rb:35` never calls `patch_spec_for_prebuilt`, which is the only place a cascaded reason gets logged (`:647`). That is why the summary names only `ExpoModulesCore` (logged by the inline `try_link_with_prebuilt_xcframework` path at `:612`) and no `📦` at all — and `print_linking_summary` (`:735-738`) prints the "no xcframeworks found" warning whenever `@linked_pods.none? { found }`. The message is the resolver reporting the cascade, not a missing download: the tarballs are present in the installed packages (`prebuilds/output/release/xcframeworks/ExpoImage.tar.gz` etc. in this worktree), no Expo pod logs `prebuilt tarball not found`, and the tarball check is never reached for `expo-modules-core`.

**Answer: the source build is guaranteed by config.** The only inputs are `packages/mobile/package.json` `expo.autolinking.buildFromSource` and the installed packages' `spm.config.json` files; no env var, cache state or network is consulted before the short-circuit. The pre-fix run (job 102296938664) found the xcframeworks because the resolver reached `resolve_own_prebuilt_info` and the tarballs were there; the new runs never get that far. Removing `expo-modules-core` from the list is the one way back to a mixed tree, and the new Linux guard fails on exactly that (see 4).

### 3. TestFlight and other native paths

- `ios-testflight-rn.yml:237-262` runs the same `vp exec expo prebuild --platform ios --clean --no-install` and a plain `pod install` in `packages/mobile/ios` against the same `package.json`, so the same resolver decision applies to the Release archive. `xcodebuild archive -configuration Release` (`:528-531`) then has no prebuilt Expo pods to link. It has **no** SwiftCompile assertion; it does run `mobile:abi-check` on the archived `.app` (`:562`). That check resolves symbols only, so it would not catch a class-layout mismatch by itself — the config guarantee plus the Linux guard are what protect the Release path. Stating it plainly: the CI assertion covers the Debug simulator build; the Release archive is covered by config, not by a log assertion. I consider that sufficient because the decision is made before any Release/Debug or tarball input.
- EAS preview builds (`vp run mobile:preview-build`) read the same `package.json` on the EAS runner; **unverified** by a run here.
- Android is unaffected (no Kotlin change; fingerprint moves via `patches/` hashing, as before).

### 4. The CI assertion and the Linux guard

- **Fails the job for real:** the step runs under Actions' default `bash -e`, sets `status=1` on any miss and ends with `exit $status` (`ios-rn-ci.yml:29-42` in the diff). `grep -Eq` inside `if` is not affected by `-e`. Both new runs print all three `compiled from source — patch exercised` lines and 0 `::error::`.
- **Cannot pass vacuously:** the regex requires the `SwiftCompile normal <arch> ` prefix plus the package-relative path; only xcodebuild's compile step prints that prefix, and a compile that *failed* still fails the build step first (`set -o pipefail` + `tee`, `:146-158`). If a future toolchain stops naming files in batched lines, the step goes red, not green.
- **Can fail falsely (minor, fails safe, pre-existing):** `build build.log || { …; build build.log; }` (`:165`) re-runs `tee build.log`, which truncates. On the codegen-race retry the incremental DerivedData will skip targets already compiled in attempt one, so a retry log may lack all three lines and red the step. This existed for `RNSScrollViewFinder`; the PR triples the surface. `tee -a` on the retry (or asserting on both logs) fixes it. Not blocking.
- **Guard (`scripts/mobile-patches-check.ts` `sourceBuildChain` / `checkIosPatchesBuildFromSource`):** anchored `^pattern$` matches Ruby's; it matches npm package names only, while Ruby also matches pod names — a config written with pod names would fail the guard falsely (safe direction). It only models `dependency_unavailable`, as documented, also the safe direction. One theoretical **false negative**: the guard follows any `externalDependencies` package, but Ruby only cascades through pods present in `pod_lookup_map` (`:1953`), i.e. packages that ship a `spm.config.json`. `expo-modules-jsi` has none in this install, so a future list of `["expo-modules-jsi"]` with `expo-modules-core` removed would satisfy the guard (`expo-modules-core → expo-modules-jsi (buildFromSource)`) while Ruby keeps `ExpoModulesCore` prebuilt. Nobody has a reason to write that config today, and the iOS CI assertion would still catch it; a `readSpmConfig(dep) !== null` condition on the cascade edge would close it. Not blocking.
- The real-package tests (`the shipped buildFromSource list`) run against the mobile-resolved copies via `createNodeEnv`, so a bump that changes a dependency's `spm.config.json` is caught on Linux. Linux CI on this head is green (lint, typecheck, test-default x2, check x2).

### 5. Revived expo-image patch — no regression found

`localAssetName` (`expo-image/ios/ImageView.swift:861-895`) is consumed only by `localAssetImage` → `maybeRenderLocalAsset` (`:369-374`, `:359-367`); a `nil` falls through to the normal SDWebImage load. Unpatched, an absolute `file://` URL became `UIImage(named: "var/mobile/…/x.png")`, which returns nil and falls through the same way; patched, it returns nil without the bundle probe. Same rendering, less main-thread work. Every `file://` source in the app is absolute: `LayeredClimbImage.backgroundImageUri` (`file://${path}`), the render scheduler's holds PNG in `Paths.cache`, `background-image-cache.ts` `asset.localUri`, avatar picker URIs. `require()`d module assets resolve to absolute `file:///…/Boardsesh.app/…` in release as well. No scheme-less asset names with a leading slash exist in `packages/mobile`. Runtime behaviour on a device: **unverified**.

### 6. Earlier verified-OK points, re-checked on this head

`git diff 21bfbb8ba..origin/main` touches none of `packages/shared/ble-protocol/src/connection-error.ts`, `packages/mobile/modules/**`, `packages/mobile/src/lib/ble/**` or `use-board-bluetooth.ts`, and the PR does not touch them either, so the matcher walk from round 3 (unanchored substring regexes; every `BoardBleError.errorDescription` lands in the intended bucket; "No board is connected" still misses `/not connected/`) stands. The Swift patch itself is byte-identical to the round-3 head (`sha256 331eed2d…` in both lockfile keys). Main's drift in `scripts/mobile-patches-check.ts` (`SERVER_PATCHES`) and `pnpm-workspace.yaml` (postgres patch) is in different hunks; GitHub reports `MERGEABLE`, and the PR's `main()` change does not touch the `checkPatchInventory` call main extended.

Unverified: EAS runner behaviour, the expo-image change on a device, the Release-archive compile volume/time.

PR #5465, opened 2026-09-15
fix(db): MoonBoard grade estimates at every angle, plus a grade-range filtering fix

fallaciousreasoning:
## Summary

Gets a real Boardsesh grade estimate onto MoonBoard, and fixes two bugs found
while proving it out on a local dev DB:

Screenshot_20260916-105559

- **MoonBoard grade coverage.** MoonBoard is back in `CROWD_MEAN_BOARDS` now
  that its `difficulty_average` is sourced from the catalog's real `userGrade`
  field (a prior commit on this branch) rather than duplicating the setter's
  grade — a climb graded at *both* its real angles (25°/40°) now gets a genuine
  `confirmed`/`provisional` grade, on top of the existing same-board angle
  transpose and the wide-angle borrowed-shape estimate (also earlier commits
  here). Together the three jobs cover every angle a MoonBoard climb can be
  played at, including the ones behind the `moonboard-wide-angles` flag.
- **Data-integrity fix (`deleteStaleGrades`).** With three jobs now writing to
  `board_climb_grades` for one board, the nightly refresh's stale-row cleanup
  had no `model_version` filter — re-running it deleted every other job's rows
  for that board, not just its own. Confirmed on the dev DB: one re-run wiped
  ~3M rows the other two jobs had just published. Scoped the DELETE.
- **Fabricated-grade fix.** MoonBoard's catalog import leaves a zero-ascent
  placeholder stats row at a problem's non-graded fixed angle, carrying a
  catalog-implied label instead of NULL. The main model read that label as
  crowd evidence and manufactured a `setter_only` grade there, which won the
  upsert over the transpose job's real estimate for the same angle. Excluded
  MoonBoard's zero-ascent rows from the main model's read.
- **Grade-range filtering fix.** Unrelated to MoonBoard specifically, but
  surfaced by it: filtering by grade only ever compared against the legacy
  `display_difficulty` column, which is `NULL` for any climb+angle with no
  `board_climb_stats` row (a MoonBoard wide angle, or any unclimbed angle with
  a published cross-angle estimate) — and the default sort routed through an
  `INNER JOIN` that dropped those climbs before the filter even ran, so
  filtering came back empty instead of just under-matching. Now falls back to
  the Boardsesh grade, applied consistently across the shared search query,
  the count query, the holds heatmap, and the offline SQLite mirror.
- **UI copy.** Finishes wiring the `moonboard_wide_angle_estimate` confidence
  tier through the play drawer's Boardsesh grade card, with its own copy in
  all four locales (distinct from the same-board transpose's wording, since
  it's a rougher, cross-board estimate).

Author-side verification: re-ran the full nightly pipeline (main model + both
MoonBoard jobs) three times back-to-back on the dev DB and confirmed stable
row counts with zero unexpected deletions each time; confirmed a real climb
(Moon Girl) now shows `confirmed` at both 25° and 40°; confirmed a wide-angle
grade-range filter query returns matching climbs via a direct GraphQL call
against the dev backend.

## Release Notes

- MoonBoard climbs now get a Boardsesh grade estimate at every angle,
  including angles beyond the board's usual 25°/40°.
- Filtering by grade range now works at angles nobody's climbed yet, instead
  of coming back empty.

## Test plan

Needs: **More → Feature Flags** → force `boardsesh-grade` and
`moonboard-wide-angles` ON, and **More → Display** → "Show Boardsesh grades"
ON. Use a MoonBoard board.

1. Open a MoonBoard climb at 25° or 40° → grade badge shows a real number
   (no `≈`).
2. Switch the same climb to a wide angle (e.g. 45°) → badge shows `≈` and
   "Rough estimate" wording, distinct from the 25°/40° estimate wording.
3. On the climb list at a wide angle, drag the grade-range filter → list
   updates with matching climbs (was empty before this PR).
4. Climb list still works normally with grades off (toggle "Show Boardsesh
   grades" off) → legacy grades and filtering unaffected.

## Risk

Risk: 4/5 — re-enables MoonBoard in the production grade pipeline (previously
excluded) and changes the nightly grade-refresh job's delete/upsert behavior;
also touches the shared search/count/heatmap query paths used by every board.
Narrowly scoped (COALESCE fallbacks only fire when the primary source is
NULL) and covered by new unit tests, but the blast radius if wrong is every
board's search results and the nightly grade pipeline.


packages/db/scripts/moonboard-wide-angle-estimate-helpers.ts
```diff
@@ -0,0 +1,154 @@
+/**
+ * EXPERIMENTAL (branch: experiment/moonboard-boardsesh-grade) — planning
+ * helpers for `refresh-moonboard-wide-angle-estimates.ts`. See
```
marcodejongh: Do i need to run this? If so create follow up issue and nag me on discord about it 😆

marcodejongh (dismissed):

marcodejongh:
Should probably fix the last claude feedback too, but can be followuo

fallaciousreasoning:
Addressed the Claude review feedback in 6 follow-up commits:

- Removed the stale `EXPERIMENTAL`/"NOT for merge" branch references (11 occurrences) and the "no frontend copy yet" note — all stale since this shipped.
- Fixed the French copy bug (`plafonds` → `boards`, matching the glossary/Spanish string).
- Fixed a real count/search inconsistency: `count-climbs.ts` was joining `board_climb_grades` at the literal browsed angle while the search list resolves it through `gradeJoinAngleSql` under cross-angle (#5405) — a climb whose grade only exists at its set angle could be found by one and missed by the other. Both now share the same helper.
- Deduped `MOONBOARD_WIDE_ANGLES` (was inlined with a manual keep-in-sync comment; `packages/db` already depends on `@boardsesh/board-config`).
- Documented the disjoint-angle-set invariant the wide-angle job's `ON CONFLICT` relies on.
- Added the missing `moonboard_wide_angle_estimate` tier to the GraphQL schema doc strings (regenerated codegen).
- Added the missing GitHub Actions workflow for `refresh-moonboard-wide-angle-estimates.ts` — it was registered in package.json but had no scheduled job, so it never would have run in production.
- Updated `docs/boardsesh-grade.md`: MoonBoard's `CROWD_MEAN_BOARDS` inclusion, why cross-angle projection still can't fire for it, the angle-transpose section's new userGrade-backed input, a new section for the wide-angle estimate, and the confidence tier table.

Left one item deliberately unaddressed: narrowing `loadShapeCoefficients`'s return type via `Pick<GradeCoefficients, 'angleOffset' | 'coeffVersion'>`. `hasAngleOffset`/`lookupAngleOffset` in `blend.ts` are shared across the whole EB pipeline and take the full `GradeCoefficients` type, so narrowing it properly would mean touching those shared signatures — more risk than the cosmetic win is worth right now. Happy to pick it up as a real follow-up if wanted.

All new commits pass typecheck + the full test suite (added coverage for the count/search fix and the grade-range fallback). CI is running on the latest push.

fallaciousreasoning:
Addressed the second round of feedback:

- **Real bug (#1, silent data-loss path):** confirmed — `loadShapeCoefficients` had no equivalent to the sibling script's abort-before-publish gate. If the angle-surface fit ever came back with zero coverage, the planner would produce 0 upserts and reap the entire ~2.8M-row table. Added the same `report.problems`-style abort the sibling script already has, before targets/existing/plan are ever loaded.
- **#2 (statsDrivenSearch join):** verified this path is actually unreachable under cross-angle today — `chooseSearchPath` routes every cross-angle search to `standard-only` before this INNER JOIN function can run (see its own doc comment). Fixed anyway for defense-in-depth: now goes through the same `gradeJoinAngleSql` helper, so a future routing change can't silently reintroduce the mismatch.
- **#5 (unenforced invariant):** added a hard assertion in `planMoonboardWideAngleEstimates` — throws if it ever plans an upsert at 25°/40°, instead of relying only on the loop-level skip.
- **#6 (stale workflow comment):** fixed.

Left two as-is, both matching existing codebase patterns rather than being new problems:
- **#3 (`getClimbStatsConditions` footgun):** the three current callers are correctly paired (verified by tests) and both the getter and every call site carry a comment explaining the pairing requirement. Didn't see a clean way to enforce it structurally without more invasive changes than the risk warrants.
- **#4 (existing-keys snapshot race):** this is the same pattern the sibling `refresh-moonboard-angle-estimates.ts` already uses (load existing keys outside the transaction). The `concurrency` group prevents scheduled overlap; only a manual `workflow_dispatch` racing the cron boundary could interleave, and the worst case is a stale row surviving one extra week until the next run reaps it — not data loss. Not something I want to change unilaterally on both jobs without know if there's a reason it hasn't come up before.

All green on typecheck + the full test suite (verified the new abort/assert logic against the dev DB directly — normal runs are byte-identical, no false positives).

packages/db/scripts/moonboard-wide-angle-estimate-helpers.ts
fallaciousreasoning: added a github action to do it automatically

marcodejongh (approved):

Issue #5488, opened 2026-09-15
A spray-wall climb has no OG share card and shares a numeric URL


marcodejongh:
Found in the SW-11 (#5444) parity pass. Two halves of the same surface.

**No card.** `GET /og/climb` draws from the bundled catalogue geometry (`packages/backend/src/services/board-render.ts` — no `spray` branch anywhere in it), so it cannot draw a wall: the holds live in `spray_wall_holds` and the background is a photograph in the private bucket. SW-11 stopped the mobile share sheet from PREWARMING that URL, because warming it would cache a blank card under exactly the URL an unfurler is about to ask for. The link still shares; it just unfurls without a picture.

**Numeric URL.** `buildReadableClimbViewPath` resolves the layout and size names from the catalogue tables, so a wall falls back to the numeric form `/spray/<layoutId>/<layoutId>/1/<angle>/view/<uuid>/<name>`. SW-14 (#5447) / SW-16 (#5449) own `/b/{slug}`, which is the real answer.

## Fix

Both belong with the public-wall sharing slice: teach the OG renderer to compose a wall (published version's photo + alive holds, public walls only — a private wall must never render a card), and route the share link through the wall's slug. Then drop the `boardName === 'spray'` early return in `packages/mobile/src/hooks/use-share-climb.ts`.

PR #5499, opened 2026-09-15
feat(web): view spray wall climbs on the web with OG cards (SW-16)

### LINKED ISSUE #5449: SW-16: Web: view a wall climb, OG card, sitemap rules
Epic #5346 (SW-16). The web side: a shareable climb page for wall climbs, the OG card, and the indexing rules.

## Why

Share links from the app open on the web. The climbing UI moved to Expo, but www still serves `/b/{slug}/{angle}/view/{climb}` and the backend serves `GET /og/climb` (`docs/og-climb.md`) from build-time board details, which a wall does not have. Public walls are search surfaces; unlisted and private ones must not be indexed or served.

## Scope

- `packages/web` `/b/{slug}/{angle}/view/{climb}` for spray: async board details from `sprayWallRenderData` (public URL for public walls, presigned for unlisted when the viewer is allowed, 404 for private), server-rendered photo + hold overlay (the web render worker path `board-render.worker.ts` / `board-image-layers.tsx` gets a runtime-geometry branch like SW-07), canonical to `/view/...`, `robots: { index: false, follow: true }` for unlisted, full metadata for public (`createPageMetadata`, title leads with the climb and wall name, JSON-LD `BreadcrumbList`).
- `GET /og/climb` supports spray for **public** walls only (the backend `board-render.ts` / `board-geometry.ts` read the runtime geometry); unlisted / private return the generic card. Never JS-challenge the page (unfurlers run no JS).
- Sitemap: the climb shard registry includes public walls' climbs only; the boards shard stays catalogue-only; `vp run check:i18n` for the new strings.
- Web profile stats: the per-board tick fan-out was re-pointed at the schema-wide board list in #5453 (`/profile-stats` `BOARD_TYPES`), so wall ticks are fetched; what remains here is the presentation rule — group them as "Spray wall · <name>" for the owner, and show them on other viewers' profiles only for public walls (#5447 decision).

## Out of scope

Any web climbing UI (queue, play, create) — Expo-web is the browser path (`docs/expo-web-deployment.md`).

## Acceptance

- A public wall climb link unfurls in Discord and Slack with the photo card; an unlisted one shows the generic card and `noindex`; a private one 404s for a signed-out visitor.
- `curl` of the public page returns an `h1`, the climb description and a crawlable link back to the wall in the first HTML.
- Sitemap tests cover inclusion of public and exclusion of unlisted / private walls.

## Where

`packages/web/app/b/[board_slug]/...`, `packages/web/app/lib/seo/sitemap/shard-registry.ts`, `packages/web/app/components/board-renderer/*`, `packages/backend/src/handlers/og-climb*`, `board-render.ts`, `board-geometry.ts`, `docs/og-climb.md`, `docs/sitemap.md`.

## Blocked by / Blocks

- Blocked by: SW-05 (#5438), SW-14 (#5447).
- Blocks: none.

## Size

M.

marcodejongh:
## Summary

A climb on somebody's spray wall now has a page on www, and a card when you share it.

Stacked on #5489 (SW-11) → #5483 → #5479; #5496 (SW-14) has merged into SW-11 already. Merge the rest in order, then this PR retargets to main.

**The page.** `/b/{slug}/{angle}/view/{climb}` gets a spray branch. It draws the wall's own photograph with the climb's holds on it — inline SVG over an ``, server-rendered into the first HTML byte, because that picture is the page's LCP and a crawler runs no JavaScript. There is no board renderer involved: every other board's art is a bundled photo plus a WASM overlay addressed by the catalogue tuple, and a wall has neither. Hold coordinates are canonical-frame pixels and the stored matrix maps photo → canonical, so the page inverts it once through `@boardsesh/spray-wall-geometry` and pushes every centre, radius and silhouette point back into photo pixels. A traced hold draws its real outline; an untraced one draws a ring, the same fallback a catalogue placement gets.

**Three states, one rule.** A public wall is server-rendered, indexable and self-canonical (the config-tuple tree does not exist for spray, so `/b/{slug}` is the climb's only URL). An unlisted wall renders for whoever followed the link, with `noindex, follow` and no card. A private wall is `notFound()` — never a 403, and 404 for its signed-in owner too: telling a stranger that a URL *is* a wall they may not see is itself the leak, and www has no session-split cache to serve one safely. The decision reads the board row the slug already resolved to, so a private wall costs no round trip and the backend is never asked a question whose answer would confirm the wall exists.

**Which photograph.** A public wall shows the copy SW-14 makes in the world-readable bucket — a stable URL a CDN and an unfurler can hold. An unlisted wall has no such copy, and a 15-minute presigned URL must never go into HTML that carries a 24-hour CDN `s-maxage`, so it gets `/api/v1/spray-walls/{uuid}/photo`, which mints a fresh signature per image fetch and is cached nowhere. That route's gate is the backend's: the read is anonymous, so a private wall is simply "not found".

**The card.** `GET /og/climb?board_name=spray` composes the card from the public copy and the lit holds, with sharp and SVG rather than the WASM overlay. Public walls only: everything else is 404 before anything reads a photo, because an OG URL is guessable from a layout id and layout ids are sequential. It is the one `/og/climb` response that is not `immutable` — a wall reset re-points the photo and the holds under the same `layout_id` + `frames`, so a year at the edge would pin last year's wall.

**The share link.** `buildSprayWallShareUrl` sends people to `/b/{slug}/{angle}/list`, with `?wall=<uuid>` when the wall is unlisted, and that route used to 404 for spray — so the one link the app hands out went nowhere in a browser, the gym page's `/b/{slug}` included. The route now renders the wall (photograph, angle, hold count; not a climb list) and redeems `?wall=` as a capability: the uuid has to be *this* wall's, the one the slug already resolved to, the same pairing `saveClimb` makes against `sprayWallUuid`. A public wall needs no param and renders on its clean URL; a mismatch is the same 404 as no param, so nothing is an oracle for which slugs are unlisted walls; a private wall refuses everyone before anything asks the backend. Always `noindex, follow`, no canonical, and nothing of it in the sitemap.

**No `BreadcrumbList`**, deliberately, though #5449 asked for one. A breadcrumb needs a parent to name, and a wall's only parent is its own page — which is capability-gated and `noindex`. Pointing an indexed climb page at a noindex parent is the same conflicting signal the metadata avoids by withholding a canonical on a noindex page. Written down in `docs/spray-walls.md` so nobody re-adds it.

**Indexing.** Public walls' climbs join the climbs shard, through a new config source that only ever emits a wall that is public, published and slugged; `buildChosenSubquery` carries the by-layout visibility predicate as a belt underneath. The boards shard is unchanged — a wall has no `/list` page to submit, and this PR makes that route an explicit 404 rather than the 500 it was (`getBoardDetailsForBoard` has no catalogue row for a wall). `/play/...` already canonicalised to `/view/...` for spray; there is now a test saying so.

Verified here: `vp run typecheck:web`, `vp run typecheck:backend`, `vp test run --project web`, `vp test run --project backend`, `vp run check:i18n`, `vp run check:i18n:orphans`, `vp check`. Two gates were mutation-tested — dropping the private-wall check on the page and on the OG handler each turned the matching test red.

Follow-ups from review, fixed here: the spray branch now runs under the **shared** render cap (`BOARD_RENDER_CONCURRENCY`, same semaphore as the catalogue path) and answers the same `503` + `Retry-After: 5` when it saturates — the guard wraps only the fetch-and-compose half, so a private wall is refused and a cached card is served without spending a slot. And a wall whose public object has gone missing (bucket non-2xx, fetch timeout, over the byte ceiling, undecodable bytes) now answers the ordinary 404 instead of a 500, which also keeps it from becoming an enumeration oracle. Both mutation-tested.

Left deliberately unfixed and tracked in #5511: `version.photoWidth ?? sourceMetadata.width` never compares the two, so a row that disagrees with the object in the bucket misplaces every hold on a card that still looks plausible. Which source wins is a product call — the homography is solved in the version's frame, so today's behaviour may well be right — and it wants a `warn` log in production before anything changes.

Closes #5449

Half of #5488, so it stays open. The "no OG card" half is fixed here. The "numeric share URL" half is mobile work this PR does not touch: `buildReadableClimbViewPath` still falls back to `/spray/<layoutId>/<layoutId>/1/<angle>/view/<uuid>/<name>` because a wall has no layout, size or set names to slug, and the `boardName === 'spray'` early return in `packages/mobile/src/hooks/use-share-climb.ts` is still there. Both want the app to route a wall's share link through `/b/{slug}`, which is a mobile change with its own review.

## Release Notes

- Share a climb from your spray wall and the link opens a real page: your wall photo with the holds lit up.
- Shared links unfurl with a picture of the wall in chats and feeds, for walls you have made public.
- Private walls stay private. A link to one just doesn't open, for anyone.

## Test plan

1. Open a public wall climb link. Photo and holds show.
2. Open a private wall link signed out. Not found.
3. Share a public climb to a chat. Card shows the wall.
4. Share an unlisted wall, open its link. Wall photo shows.

## Risk

Risk: 3/5 — new server-rendered board type on public pages

https://claude.ai/code/session_01Q7BjHrRALJ1gjtiRWm5rLV


marcodejongh:
## Review items addressed

`fc4d32de4`. Everything actionable from the bot reviews, plus the two coverage gaps.

- **`fetchPhotoBytes` had no deadline and no size cap** — now an 8 s `AbortSignal.timeout` and a 12 MB ceiling. `Content-Length` is checked first as a cheap refusal, the decoded length again afterwards (the one that actually binds). The dep is injected, so no test shape changed.
- **Double sharp decode on the `photoWidth`-null path** — one `sharp` instance answers both the resize and the source metadata. New test asserts a wall with no recorded width renders byte-identically to one that has it, and that the photo is fetched once.
- **Missing test: wall gone, climb row still there** — `deleteSprayWall` keeps the climbs, so `getClimb` answers while the wall read returns null. Pinned: the page 404s.
- **Missing test: `version.photoWidth` null** — covered by the decode test above.
- **502 logged at the wrong level** — `log.error`, and the redundant `console.error` is gone, so the route/method/Railway request id ride on the one line.
- **German dative article order** — `einer*einem unbekannten Routenbauer*in` was both reversed and ugly. The glossary prefers a neutral rewrite where a starred form gets ugly, so it is now `unbekannter Hand`: "geschraubt von unbekannter Hand".
- **Photo redirect route loading hold geometry** — flagged as non-blocking, fixed anyway. It wanted one presigned URL and was reading `sprayWallRenderData`, which carries every alive hold (up to 1,500 rows per image fetch). It now reads `sprayWall`, same view rule, same gate.

Not changed, deliberately:

- **`count()` returning `bigint`** — `Number(row.climbCount)` is already the guard, and it is read only as a `> 0` gate. An extra `::int` cast would be a second spelling of the same guarantee.

Gates: `vp run typecheck:web`, `vp run typecheck:backend`, `vp check` (0 errors in changed code; the 2 repo errors are the pre-existing `@gorhom/bottom-sheet` nested-install pair in `packages/mobile`), `vp run check:i18n`, `check:i18n:orphans`, `--project i18n` 101 pass, spray web tests 51 pass, backend `og-c` 49 pass.

marcodejongh:
## Review items addressed (round 2)

`ac75e1a8b`.

- **Non-finite mapped coordinates in `buildSprayLitHoldMarks`** — real bug, fixed. The report is right that the backend twin already guarded this and the web path did not, and slightly understates it: a browser reads `cx="NaN"` as 0, so the hold did not disappear, it moved to the corner of the photograph and showed a hold the climb does not use. Centres and radii are now dropped when non-finite, and a silhouette with one bad vertex discards the whole ring so the hold falls back to its circle rather than being drawn as a different shape. Mutation-tested: removing the guard turns the new `drops a hold the map sends to infinity` case red (matrix `[1,0,0, 0,1,0, 0, 1/400, 1]`, whose inverse puts the vanishing line through the wall).
- **Duplicated `withTimeout`** — extracted to `packages/web/app/lib/seo/sitemap/with-timeout.ts`, one copy, with the "stops waiting rather than cancelling" reasoning written where the function is.

Not changed, with the reason:

- **`console.error` for a singular homography and Railway log drains** — a Next server component's `console` writes to stdout, which is exactly what the drain reads, so the line does arrive. Importing the web request logger would pull server-only modules into `spray-climb-view.ts`, which is otherwise pure arithmetic with a unit test around it. Left as is.

Gates: `vp run typecheck:web`, `vp check` (0 errors in changed code; the 2 repo errors are the pre-existing `@gorhom/bottom-sheet` nested-install pair under `packages/mobile`), `vp run check:i18n`, and `--project web spray sitemap` at 32 files / 315 tests.

PR #5506, opened 2026-09-15
docs: explain why Kilter's app shows more climbs than Boardsesh

marcodejongh:
## Summary

Climbers ask why Kilter's app shows about 36k climbs on an 8x12 Kilter Homewall while Boardsesh shows about 18k. Both apps have the same climbs. Boardsesh counts each climb once; Kilter's app counts each (climb, angle) pair.

Prod numbers (read-only, 2026-09-15), Kilter Homewall layout 8, `is_listed AND NOT is_draft AND NOT is_hidden`, all frame counts on both sides:

| Size | Distinct climbs | `board_climb_stats` rows | Rows per climb | Kilter's app |
| 8x12 Full Ride (size 23) | 18,335 | 36,000 | 1.96x | ~36k |
| 10x12 Full Ride (size 25) | 28,671 | 49,892 | 1.74x | ~50.6k |

We measured our own data, not Kilter's code: counting our stat rows one per (climb, angle) matches their 8x12 total and lands within about 1.5% on the 10x12. The docs carry that caveat, plus the 114 of 36,000 rows (0.3%) that exist only because of Boardsesh ticks (`upstream_synced_at IS NULL`). The user-facing copy stays unhedged.

What changed:

- **Help page:** new "Why does Kilter's app show more climbs?" section at `/help#climb-counts`. `Section` takes an optional `id`, and `.section` gets `scroll-margin-top` so the anchor lands below the fixed header.
- **Mobile filter sheet:** on Kilter boards only, a one-line note under the Show button: "Each climb counts once, not once per angle."
- **i18n:** `help.climbCounts.*` (marketing) and `mobile.filter.countNote` (climbs) in en-US, es, fr and de. The translations follow the glossaries: es bloque/plafón/encadenes, fr bloc/croix/la board/vous, de Boulder/Begehungen/das Board/du.
- **Docs:** `docs/kilter-sync.md` gains "Why Kilter's app shows more climbs than Boardsesh". It covers the counting unit (`count-climbs.ts` joins stats at one angle), the prod table, the two SQL queries to re-check, and the boulders-only default.

No data, schema or backend changes. JS-only, so it targets `main`.

One for the next translator pass, left alone here: `mobile.filter.showCount_one/_other` says "vías" in es and "voies" in fr, so the Show button says routes while the new note under it says "bloque" / "bloc". The glossaries call a climb a bloque / bloc, so the button strings are the ones that are off.

Author-side verification (local, foreground):

- `vp run check:i18n`: OK (re-run after the review fixes)
- `vp run check:i18n:orphans`: OK (re-run after the review fixes)
- `vp test run --reporter=agent packages/shared/i18n/src/__tests__/catalog-completeness.test.ts`: 96/96 passed (re-run after the review fixes)
- `vp test run --reporter=agent packages/mobile/src/components/__tests__/climb-filter-sheet.test.tsx`: 53/53 passed, including 3 new cases (note shown on Kilter, hidden on Tension, hidden with no board)
- `vp run typecheck:mobile`: passed
- Web typecheck, the full test suite and the manual `/help#climb-counts` scroll check are left to CI and QA.

## Release Notes

- The filter sheet and the Help page now explain why Kilter's app shows more climbs than Boardsesh: it counts each climb once per angle.

## Test plan

1. Kilter board → Climbs → open filters → note under Show button.
2. Switch to a Tension board → filters → no note.
3. Open boardsesh.com/help#climb-counts → section title visible below header.

## Risk

Risk: 2/5 — copy plus a small filter-sheet note, no data changes

https://claude.ai/code/session_01QDDL8BXuC5j6rExWufofo9


marcodejongh:
Thanks — went through all three, no changes needed:

1. **`scroll-margin-top` on `.section`** — deliberate. Every help section is a legitimate anchor target, and the page sits under a fixed header (`--global-header-height`, set in `packages/web/app/components/index.css`), so without it any `#fragment` scrolls the heading under the header. Scoping it to one section would leave the same bug waiting for the next anchor. Calling it out here rather than adding a second class.

2. **`boardName === 'kilter'`** — matches how the rest of the app gates on board (`ClimbActionsSheet.tsx:57` does the same), and `boardName` comes from the typed board config rather than user input, so casing is fixed. The note is Kilter-only on purpose: Kilter's app is the one we measured (36,000 (climb, angle) stat rows vs 18,335 climbs on an 8x12 Homewall). We have no equivalent measurement for Tension's app, so claiming it there would be a guess. Widening it later is one line.

3. **Tests don't assert the key** — they do. The suite's `t` mock returns the key it is given, so `getByText('mobile.filter.countNote')` fails on a wrong or missing key, which is exactly the case this would catch. The other two cases assert the Kilter gate.

PR #5528, opened 2026-09-17
perf(backend): speed up Crew feed loading and pagination

marcodejongh:
## Summary

Follow-up to #5526 and #5527: fix the slow Crew first load and infinite scroll without changing the mobile app or database schema.

- Replace the catalogue-wide OR-of-EXISTS candidate query with three author-index lookups (direct setters, native authors, linked board accounts), deduplicated with UNION.
- Materialize the narrow followed-climb rows before legacy date parsing. Preserve the 30-day window, exact cursor, spray visibility, and enrichment-time membership/visibility rechecks.
- Add result-equivalence, overlapping-follow, cross-board, empty-follow, microsecond-pagination, and executed-plan regression tests.

Read-only production benchmark, one anonymized account following 26 users and two setters:

| Climb candidate query | Before | After |
| --- | --- | --- |
| First page | 10.2 s | 77 ms |
| Pagination | 11.0 s | 56 ms |

The original pagination plan filtered 788,160 rows to return five. The new date-processing stage receives only 1,963 followed climbs; existing author indexes serve the lookup branches. First-page IDs, ordering, and timestamps matched exactly. Pagination used a representative one-day-old cursor. These are candidate-query timings, not end-to-end app timings. The supplied production role cannot SELECT spray_walls, so both benchmarks exclude spray climbs; local tests exercise the complete visibility predicate. No production data, permissions, or indexes were changed. Existing session candidate queries measured 0.35–0.55 s separately.

Verification: all CI checks green, including 145 backend tests across 14 files. All 30 focused local tests pass with one worker (Crew candidates/feed/snapshot, followed authors, and credential concurrency); workspace and backend typechecks plus lint pass. Review comments addressed in 16d8db38d.

Local validation limitation: the unrestricted full-backend run exhausted the local PostgreSQL shared-memory/lock budget during concurrent schema setup and was interrupted; it is not claimed as a passing full-suite run. The credential-concurrency test that failed during that overloaded run passes in the bounded rerun. No unrelated application code or database settings were changed.

## Release Notes

Load your Crew feed and scroll to older climbs faster.

## Test plan

1. Signed in, open Home → Crew loads followed climbs and sessions.
2. Scroll down → older cards load without repeats.
3. Pull to refresh → newest followed activity appears first.
4. Unfollow a setter → their exclusively followed climbs disappear after refreshing.

## Risk

Risk: 3/5 — backend candidate selection changes; no migration or mobile changes.


marcodejongh:
Addressed the review points:

1. Added an explicit link to the membership source and the equivalence tests. The three indexable branches intentionally implement the same predicate differently; reusing OR-ed EXISTS would reintroduce the measured broad scan.
2. The CTE alias now comes from Drizzle's public `getTableName` API, with an inline explanation of how projected columns are reused.
3. Replaced the fixed `9` assertion with an independently queried expected membership count. Kept the executed materialization assertion: preventing publication predicates from being pushed into the catalogue lookup is the regression this test must catch. It makes no assertion about join algorithms or index choice.
4. Confirmed fixture isolation: `src/__tests__/setup.ts` truncates the tables before each file, and `worker-db.ts` gives each parallel worker its own database. Added that context beside the fixtures.
5. Documented that intermediate work scales with the complete followed catalogue, not page size. Capping before date filtering/global ordering would drop valid climbs; this PR deliberately preserves existing results.

No production data or schema changes. The implementation uses existing author indexes; the read-only production sample measured 77 ms / 56 ms for first-page / pagination candidate queries.

Issue #5451, opened 2026-09-14
SW-18: Server-side detector service (our Apache model in a CPU container) + job worker


marcodejongh:
Epic #5346 (SW-18). **Rescoped 2026-09-15:** build this when on-device cannot clear the gesture-savings gate (#5434) within the phone budget. Default shape: serve OUR OWN Apache-2.0 RF-DETR from a CPU-only Python container IN THIS REPO (`packages/hold-detector`), plus the durable job plumbing below. The separate AGPL-3.0 repo variant applies only if an AGPL model is chosen; the AGPL reference benchmark bought ~0.06 F1 over a one-epoch Apache run, so that variant is unlikely.

**Licensing (owner decision 2026-09-14):** this monorepo and the app stay Apache-2.0 and never embed AGPL code or weights. The detector service is a separate repo (suggested name `boardsesh/hold-detector`) under AGPL-3.0, so it may use Ultralytics YOLO or any other AGPL model. Only detection results cross the wire, so the app is not a derivative. The AGPL network clause is satisfied by that repo being public. Nothing from it is vendored here; this repo holds only the HTTP contract, the job table, the worker and the GraphQL surface.

## Why

If a phone-sized model cannot reach the quality bar but a medium model on a CPU can, detection moves to the server. The Node backend cannot host it comfortably (`Dockerfile.backend` is `node:22-alpine` / musl, ONNX Runtime prebuilds target glibc, and inference would compete with the WebSocket-serving replicas), so this is a separate container. Railway offers no GPU and does offer Serverless / app sleeping, which suits a low-volume async job.

## Scope

- Separate repo `boardsesh/hold-detector` (AGPL-3.0): FastAPI on `python:3.11-slim`, CPU inference (ONNX Runtime or the model family's own runtime), the best model from #5434 or an AGPL one if that is what clears the gate, endpoints `POST /detect` (photo → holds in photo px with outlines and colour descriptors) and `GET /health`; shared-secret header; reachable only over Railway private networking; Serverless on; 2 GB memory cap. Its Railway service is configured from that repo, not from `vp run railway:apply` here. This monorepo documents the HTTP contract in `docs/spray-walls.md` and pins the contract with a recorded-response test.
[...]

PR #5536, opened 2026-09-18
feat: trace hold outlines from a segmentation model

marcodejongh:
## Summary

`@boardsesh/hold-detection` handed back circles, and `candidates.ts` documented why: the detector returned boxes, and the only silhouette on the Python side was a classical in-box guess that stopped at a boolean mask with no tracer to port. Every one of those reasons was about the model — and the model changed. `seg-nano-tiled-1024` (`mask_source: "model"`) emits a third tensor of per-query mask logits.

`decodeRfDetr` now takes that tensor and traces each surviving query into a ring, in the tile's own frame where the query index still lines up with a slice of the mask. The ring comes out in units of the hold's equivalent-circle radius — the `@boardsesh/board-art-geometry` contract `spray_wall_holds.outline` and the renderer already speak — so `mergeTiles` and `unLetterbox` never have to touch it.

This matters because `spray_wall_holds.outline` being NULL means the renderer draws the hold as a plain circle. Outlines are the Aura look.

### Three decisions, each a way to get a visibly wrong outline

- **Interpolate the logits, then threshold** — matching rfdetr's own `_postprocess_masks` (`F.interpolate(..., mode="bilinear") > 0.0`). Measured on the spray-wall eval split, the other order drops outlines in the plausible 0.40–0.80 fill band from **96% to 79%**.
- **Take the component under the box's centre**, not the largest one. A mask can carry a second blob where a neighbouring hold bled in.
- **Trace through pixel centres and leave the resulting few-percent area bias uncorrected**, because OpenCV's `findContours` shares that convention and the Python harness these numbers were measured against carries the same bias. Correcting one side would put the two out of agreement.

### Additive by construction

`outlines` defaults off, a detection-only model has no mask tensor to plan from, and a detection whose mask came back empty still returns `undefined` — so the renderer's circle fallback remains the path for every config shipping today. The committed parity fixtures are unchanged and still pass.

9 new unit tests cover the ring being scale-free, centred on the box, picking the right blob, declining empty and degenerate input, surviving a mask that touches its own edge, and actually smoothing rather than stepping.

## Test plan

1. CI green.

## Risk

Risk: 2/5 — new code path is opt-in and unreachable from any shipping config; existing parity fixtures unchanged.

## Release Notes


marcodejongh:
Integrated these outline commits into #5587 with durable service jobs and the homelab worker. Keeping this PR open pending review of the combined rollout so we do not lose its review history or merge overlapping work accidentally.

marcodejongh:
Closing as superseded by merged #5587 (`0c0cd0b6d`). Independent reviews verified that all six files are absorbed: decoder, candidates, exports and types are identical; main additionally improves outline crop margins, simplification and storage-contract safeguards. The original review follow-ups are preserved on #5451: https://github.com/boardsesh/boardsesh/issues/5451#issuecomment-5761817399. The homelab rollout issue remains open; this closure does not claim deployment or device QA is complete. Branch retained.

PR #5537, opened 2026-09-18
feat: group a setter's new climbs and open them in the play drawer

marcodejongh:
## Summary

The Home feed's "New climb" cards had three problems, all visible in the reported screenshot.

**They got stuck on a spinner.** Tapping one pushed the climb redirector, which opened the play drawer and then popped it straight back off: `openPlayDrawer` ends in `router.navigate('/play')`, so the `router.back()` that followed dismissed `/play` instead of the redirector — and `handedOffRef` was already claimed, so the effect could never retry. The climber was left on a headerless spinner with no way out. The existing test mocked `openClimbInPlayDrawer`, so it asserted call order and never saw the pop land on the wrong screen.

The redirector now leaves before it opens, in both branches — `/play` is a root `transparentModal`, not a child of the popped stack, so it survives the pop. The test models a stack instead of call order; reverted against the old ordering it fails with `expected ['/(tabs)/climbs/[climbUuid]'] to deeply equal ['/play']`, which is the bug exactly.

The card never needed that route anyway: the feed already carries `frames`, `renderBoard` and `angle`, and the card drew board art from them. It now builds the climb locally and opens the drawer in place, the way a session tick already does — no round trip, nothing to hang on.

**One card per climb.** `crewFeed` now groups by `(board, author, local day)` in SQL, *before* the page is cut, so a setter's day can never split across two pages. That forces the cursor into a `HAVING` over `max(pub)`: as a `WHERE` it would drop a group's newest members, move its timestamp backwards, and hand the same setter a second card later. A second query ranks each selected group's climbs with one window function and takes the ten newest, so a page costs two queries however many climbs the groups hold. `timeZone` on `CrewFeedInput` draws the day boundary where the climber's clock puts it; an unknown zone degrades to UTC rather than failing the feed.

Grouping is **opt-in** via `CrewFeedInput.groupClimbs`, and off by default. An adversarial review caught that my first cut would have taken the Home tab *down* on every store build that hadn't yet pulled the OTA: an old client has no fragment for `CrewClimbGroupItem`, so the item arrives as a bare `__typename`, its `renderItem` falls through to the session branch, and `item.session.socialEntityType` throws inside the feed list. An unasked client now pages over climbs through its own candidate query — the pre-grouping one, so its feed is byte-for-byte what it gets today.

The same review caught a second live-outage bug: `resolveTimeZone` validated with `Intl`, which accepts every IANA backward link, while Debian's Postgres images moved those to `tzdata-legacy` in 2023. `Asia/Calcutta`, `Europe/Kiev`, `US/Eastern`, `US/Pacific` and `America/Buenos_Aires` all pass `Intl` and then raise `time zone not recognized` inside the query — and Android reports `Asia/Calcutta` on a great many devices, with React Query keying the feed cache on the zone so those climbers could never recover. Zones are now checked against `pg_timezone_names`, read once per process.

**They formatted wrong.** The card was edge-to-edge where every `SessionFeedCard` is inset by `spacing[4]`, dropped the avatar the resolver already sends, and joined grade and angle into one string — so an ungraded climb read as a bare `30°`. It now matches the send card: avatar and setter in the header, grade right-aligned in its grade colour, angle and board as their own slots, hairline divider. An unformattable grade falls back to its raw name instead of vanishing.

A Codex review then caught that my first attempt at that compatibility path was itself broken: it expanded groups *after* the limit and the chronological merge, so a limit-20 request could return 200 cards and a group's 08:00 climb could be emitted ahead of a 09:00 session. Hence the separate legacy candidate query. Codex also found three pieces of climb detail the direct-open path dropped that the redirector used to fetch — the setter's notes, the authored `framesPace` for a multi-frame climb, and a benchmark check that treated `0` as a benchmark when imports use `0` and `NULL` interchangeably — plus a "See all" that resolved its board from the *viewer's* active board, landing a group from another wall on a list filtered to a board that setter never set on.

Two more from the adversarial review: the card's page index is now paired with its group (FlashList recycles one group cell into the next, and the old code left group B showing group A's page — with no dot lit at all), and the feed carries real `ascensionistCount`/`qualityAverage`/`isBenchmark` instead of zeros, because the drawer does **not** refetch a preview whose angle already matches, so opening in place would have shown "0 ascents, no stars" on a climb the old route displayed correctly.

Author-side verification: backend suite green, full mobile suite green, `typecheck:backend` + `typecheck:mobile` clean, `vp check` 0 errors, `check:mobile-bundle` OK, `check:i18n:orphans` OK (it caught `authors.setBy` going orphan when the card dropped its "Set by" footer — removed from all four locales), i18n parity 96/96. Each of the three ordering/recycling fixes has a test that was confirmed to fail against the unfixed code.

No migration. No native change, so this ships by OTA.

## Release Notes

- A setter's new climbs from one day now arrive as a single card you swipe through
- Tapping a new climb opens it straight in the play drawer instead of a stuck loading screen
- New-climb cards line up with the rest of the feed, with the setter's avatar and a colour-coded grade

## Test plan

1. Home → My crew. New climb cards line up with session cards.
2. Find a setter with several climbs. One card, not several.
3. Swipe that card sideways. Next climb, dots follow.
4. Tap any climb. Play drawer opens — no loading screen.
5. Tap a climb on someone else's board. Switch-board prompt, not a dead tap.

## Risk

Risk: 3/5 — new GraphQL union member (opt-in, so old builds are unaffected) and a rewritten crew feed query, plus a hand-off ordering fix shared with deep links.

https://claude.ai/code/session_017PEdfDrdgVziSEQvZf3Kcn


marcodejongh:
All five Codex findings are fixed in f86e771. Each was reproduced before fixing.

**P1 — legacy pagination before expanding groups.** Correct, and the worst of the five. An unasked client now pages over *climbs* through its own candidate query (`buildCrewClimbRowCandidatesQuery`), which is the pre-grouping query — so `limit` and newest-first both hold, and its feed is what it is on `main` today. Note this makes the legacy path **uncapped** rather than ten-per-group: nothing is silently lost off a group's tail. Test walks the cursor and asserts every climb appears exactly once, in order, within the limit.

**Setter notes / `framesPace`.** Both correct — the redirector used to fetch the full climb. `description` and `framesPace` now ride the feed payload and the mapper copies them.

**Zero benchmark difficulty.** Correct. Verified the precedent before changing it: `proposals/enrichment.ts:164,474` and `ticks/queries.ts:44` all test `> 0`. Now `(benchmarkDifficulty ?? 0) > 0`, with a fixture pinning a `0` row as non-benchmark.

**"See all" board.** Correct — `usePlaylistRenderBoard(null)` resolves from the viewer's active board, so a group from another wall landed on a list filtered to a board that setter never set on. The setter route now takes optional `boardType`/`layoutId`, and the card passes the group's.

Backend crew-feed 39/39, full mobile suite 11073/11073, both typechecks and `vp check` clean.

One declined, for the record: the suggestion to add a TTL to the `pg_timezone_names` cache. `pg_timezone_names` only changes when the server's tzdata is upgraded, which restarts the process anyway; a stale-but-successful read costs at most a UTC fallback for a brand-new zone name, while a TTL adds a recurring query and a refetch race. The asymmetry is deliberate — a failed read must retry, a successful one needn't.

PR #5538, opened 2026-09-18
fix(mobile): drop Android elevation shadow on the connected lightbulb halo

alejandrosanchezcabana:
## Summary

Reported: on Android, when you're connected to a board (or holding a ledless wall), the lightbulb "connected" indicator — the bulb at the bottom-left of the current-climb bar / the in-drawer lightbulb — casts a visible **hexagonal** shadow instead of a soft round glow.

Root cause: `BoardControlIndicator` and `BleLightbulbButton` both draw their "connected" halo on a fully-rounded 44×44 `Pressable` (`borderRadius: size / 2`) using `elevation: 2` alongside the iOS `shadow*` props. Android's native elevation shadow on a small perfect circle renders as a low-poly hexagon instead of a smooth circle on real hardware — this is a known Android outline-shadow rendering artifact, not something tunable from the shadow radius/opacity values.

The warm amber halo itself is already carried by `backgroundColor: haloColor` on both platforms (a translucent tint of the connected color), so the native drop shadow is a pure extra depth cue that isn't needed on Android. Made it iOS-only (`elevation: Platform.OS === 'ios' ? 2 : 0`) in both places that share this exact pattern.

No BLE protocol/behavior change — style-only fix to a `StyleSheet` value. Verified with `vp run typecheck:mobile` (clean) and `vp test run --project mobile` (983 files / 11062 tests passing) and `vp check` (0 errors).

## Release Notes

- Fixed a glitchy hexagon-shaped shadow around the lit bulb icon on Android when connected to a board

## Test plan

1. On Android, connect to a board (or take a ledless wall).
2. Look at the bulb icon, bottom-left of the current-climb bar.
3. See a clean round glow — no hexagon-shaped shadow.

## Risk

Risk: 2/5 — isolated style change to two shared components, covered by existing render tests.

https://claude.ai/code/session_016qoEL9uwoBpTHbPTFswy7p

---
_Generated by [Claude Code](https://claude.ai/code/session_016qoEL9uwoBpTHbPTFswy7p)_


alejandrosanchezcabana:
The `gate` check on **iOS CI (React Native)** failed on this PR's head (`543092f`), but it's a CI infra flake, not this PR's:

The job's `setup-vp` step hit an SSL/TLS connection error reaching `registry.npmjs.org` / `viteplus.dev` while installing the Vite+ CLI, after exhausting its 4 built-in retries — unrelated to the diff (a Bluetooth/queue-control style-only change). This is confirmed as transient rather than systemic: the same commit's other jobs (`release-notes`, `commit-lint`, `i18n`) ran the identical `setup-vp` install step successfully around the same time.

I don't have permission to trigger a re-run from here (`rerun_failed_jobs` / `rerun_workflow_run` both return `403 Resource not accessible by integration`). Could someone with Actions access re-run the failed `gate` job on this PR? I'll keep watching for the result.

---

alejandrosanchezcabana:
Both addressed in 9c6f265: trimmed the two comment blocks to one line each, and added `BleLightbulbButton-elevation.test.tsx` covering the Android branch (asserts `elevation: 0`) alongside the existing iOS-pinned render test.

---

alejandrosanchezcabana:
Both addressed in 6de9228:

- Added `board-control-indicator-elevation.test.tsx`, the same iOS/Android pair for `BoardControlIndicator`.
- On the "fragile assertion" point: the test doesn't use React Native's Test Renderer — it fully mocks the `react-native` module (`vi.doMock`) with a custom `Pressable` that resolves the style array/function itself and writes `elevation` into a `data-elevation` attribute, then renders through `@testing-library/react` on jsdom. So it's not reading a real RN style prop off the DOM. To double-check it isn't vacuous, I reverted the fix locally (hardcoded `elevation: 2` in both files) and confirmed both Android tests fail (`expected '2' to be '0'`) while the iOS tests still pass, then restored the fix and confirmed all green again.

---

alejandrosanchezcabana:
Leaving `flattenStyle` duplicated for now — two call sites doesn't clear the bar for a shared test utility yet, but agreed it's worth extracting if a third test needs the same pattern.

---

packages/mobile/src/components/queue-control/__tests__/board-control-indicator-elevation.test.tsx
```diff
@@ -0,0 +1,78 @@
+// @vitest-environment jsdom
+// Regression coverage for the Android hexagon-shadow bug (see
```
marcodejongh, line 2: Lol it made you add a lot of tests for a tiny change

marcodejongh (approved):

alejandrosanchezcabana:
Fair — the automated review flagged the Android path had zero coverage on the first pass, so I closed the gap for both components rather than just one. Thanks for the approval!

---

Issue #5539, opened 2026-09-18
Support Kilter live sync for boards without an exact imported wall match


marcodejongh:
The first Kilter live-history integration supports only exact imported public gym-wall identities, including canonical board-merge redirects.

Follow up on safe mappings for the remaining board types:

1. Design explicit wall selection for manually created gym boards and persist the confirmed binding.
2. Resolve ambiguous imported/merged wall identities without combining distinct physical installations.
3. Verify custom/home-wall selectors, ownership, and visibility with authorized read-only probes.
4. Establish whether configuration-only boards can be mapped safely; configuration or controller serial alone must not identify a physical wall.
5. Define binding validation, revocation, UI, and regression tests before expanding polling coverage.

Keep sync one-way: read Kilter displays into Boardsesh history. No upstream activity writes.

PR #5540, opened 2026-09-18
Merge Kilter live activity into board history

marcodejongh:
## Summary

A linked Kilter viewer now starts one Redis-coordinated live-history poller for an exactly matched public wall. Polling stops when the last linked viewer leaves. Imported occurrences merge into Redis and durable history with source-aware deduplication; they never change the current climb, holder, queue, or native activity statistics. This integration only reads Kilter activity.

The presence sheet now loads its first durable page automatically, pages chronologically independently of Redis, retries failed pages, and refreshes both histories. Kilter entries carry a source label. Token rotation is serialized with the existing sync daemon.

Rollout: apply migration 0233, let the Kilter reference/location sync populate wall selectors, then enable `KILTER_LIVE_SYNC_ENABLED=1` on backend instances. Broader wall matching remains in #5539. Includes the protocol documentation from #5228; the implementation guide at `docs/kilter-live-history.md` documents the approved one-way scope, standalone Redis requirement, and rollout.

Validation: workspace typecheck and format/lint passed; 452 shared tests and 113 focused backend tests passed across the final runs; iOS and Android bundles passed. All 11,062 mobile tests passed. Simulator/screenshot wrappers skipped on Linux. Production database access remained read-only. After rebasing onto main: 157 focused backend tests pass across Kilter sync, board presence, access controls, async iterators, and managed subscription cleanup. This includes 31 Kilter cases covering invalid caches, exact-angle grades, cancellation, Redis failures, socket teardown, remote credential broadcasts, and merge-depth diagnostics. Workspace typechecking, lint, and migration checks pass. The development-image smoke check permits the new wall-mapping table while still rejecting legacy tables. Optional partial-index optimizations are deferred to #5551 at Marco’s request. Latest review fixes add one-second Redis read deadlines, reuse the transaction connection for eligibility, log viewer-removal failures, and validate all cached history fields; 48 focused Kilter/credential tests pass, with workspace typechecking, lint, and migration checks green locally.

## Release Notes

See recent climbs displayed through Kilter alongside your board history.
Older board history loads when you open the presence sheet.

## Test plan

1. With Kilter linked, open a public gym board’s presence sheet.
2. Wait 35 seconds; Kilter entries show their source.
3. Open a quiet board’s presence sheet; older climbs load automatically.
4. Pull history down; recent and older climbs refresh without duplicates.
5. Scroll history; older climbs appear in time order.

## Risk

Risk: 5/5 — Adds a history migration, credential locking, and distributed polling. Disabled until configured.


marcodejongh:
Checked both automated reviews against commit e23e4e084de18700265263ecdbc08ca6a7a4c6d5. Marco approved retaining the safeguards below and slowing unmatched-wall retries.

1. **Location sync does not expose partially de-listed rows.** The reset and upserts are inside [one transaction](https://github.com/boardsesh/boardsesh/blob/e23e4e084de18700265263ecdbc08ca6a7a4c6d5/packages/kilter-sync/src/sync/locations-sync.ts#L192). Other connections cannot read its uncommitted updates under READ COMMITTED; see [PostgreSQL's isolation documentation](https://www.postgresql.org/docs/17/transaction-iso.html#XACT-READ-COMMITTED).

2. **The reported missing guards already exist.** [The duplicate lookup precedes sequence allocation](https://github.com/boardsesh/boardsesh/blob/e23e4e084de18700265263ecdbc08ca6a7a4c6d5/packages/backend/src/services/kilter-live-import.ts#L147). A concurrent unique-conflict race can leave an allowed sequence gap. [Refresh clears the in-flight flag before loading](https://github.com/boardsesh/boardsesh/blob/e23e4e084de18700265263ecdbc08ca6a7a4c6d5/packages/shared/board-presence-react/src/use-board-history-pagination.ts#L179), including the exact line proposed by the review. [Credential notification failures already log a warning](https://github.com/boardsesh/boardsesh/blob/e23e4e084de18700265263ecdbc08ca6a7a4c6d5/packages/backend/src/services/aurora-credentials.ts#L622); the dynamic import is cached.

3. **Mixed history follows a separate, history-only path.** Cold-start and reconnect results from `fetchRecentHistory` use [MERGE_HISTORY](https://github.com/boardsesh/boardsesh/blob/e23e4e084de18700265263ecdbc08ca6a7a4c6d5/packages/shared/board-presence-react/src/use-board-presence.ts#L241), also at line 334. `BACKFILL_HISTORY` receives native recent climbs and can affect current-wall state, so excluding Kilter there is intentional. The legacy `boardHistory` endpoint stays native-only to preserve old-client behavior; the new history endpoints include Kilter. Sharing the mixed results with the legacy path would weaken that compatibility boundary.

4. **Connection cleanup and cancellation checks are intentional.** The [close listener](https://github.com/boardsesh/boardsesh/blob/e23e4e084de18700265263ecdbc08ca6a7a4c6d5/packages/backend/src/websocket/setup.ts#L236) is installed once per socket, which may carry multiple subscription operations. Removing it when one operation ends would lose cleanup for the others; viewer cleanup is idempotent. `isCurrent()` checks ownership and eligible viewers, not the token helper. The helper runs once per poll, plus a forced retry after a 401. Credential row locking coordinates token rotation/unlink. Per-occurrence transactions and fresh eligibility checks cost round trips, but retain cancellation and account-change safeguards; that is a performance tradeoff, not the claimed per-occurrence token-lock bug.

5. **Accepted retry improvement; other minor findings describe hypothetical contract changes.** Unmatched walls already make no token or upstream history request. Their binding retry is now five minutes, with an integration test covering the cooldown and absent credential/REST calls (10 Kilter integration tests pass). Cursor timestamps use an explicitly configured string-mode column, and timezone-less upstream timestamps are intentionally rejected instead of guessed. The observed Kilter response includes a timezone. The PR also updates the WebSocket docs and the Kilter integration/spec docs.

Separately, the real CI access-sweep failure was fixed with seeded history plus owner/stranger/anonymous coverage. The generated migration also preserves account deletion through cascading source-map cleanup. These fixes are already in e23e4e084.

marcodejongh:
Addressed the review findings with regression coverage:

1. The `sentAt` string type guard was already present. The missing check was whether that string represents a valid timestamp. Invalid cached timestamps now produce a warning and fall back to PostgreSQL, preserving recoverable history. Invalid timestamps in merged history are explicitly counted and logged; normal expiry remains silent.
2. The angle fallback never supplied the grade: that already used the exact `(climbUuid, angle)` lookup. The implementation now separates angle-independent name/frames/setter metadata from angle-specific statistics, making that constraint explicit. A test imports the same climb at two angles with statistics at only one and verifies the other grade is null.
3. Added cancellation tests after the first committed occurrence and inside the transaction before insertion. They verify that no further occurrence is committed and no cache update or notification is emitted after cancellation.
4. Added Redis failure tests for control-channel subscription setup, viewer-lease pipeline errors, and an active REST read interrupted by a coordination failure. They cover listener cleanup, recovery, request cancellation, and rejection of late responses.
5. Board/layout and wall-binding validation failures now warn with board/source identifiers and a reason, once per reason per batch. Expected duplicates and routine cancellation remain quiet.

All 23 Kilter integration tests pass (13 new cases), and workspace typechecking, formatting/lint, and commit hooks pass. Pushed as c800de01b. The separate default-shard CI failure on ad80b9f00 was a five-second timeout in the unchanged offline-sync snapshot bootstrap test; that exact test passes locally in 462 ms. The new commit starts a fresh full CI run.

marcodejongh:
Addressed the new findings in the edited review:

1. Socket teardown snapshots the viewer entries before removal. Added tests covering multiple subscription operations on one socket, idempotent disconnect, another viewer staying connected, and a socket watching multiple boards. The original `delete` happened before the first await; the snapshot makes the iteration set explicit.
2. Added a Redis fan-out test: an instance with no local watchers broadcasts `credentialsChanged()`, and the remote owner aborts before its 15-second heartbeat.
3. Generated migration 0233 makes external-occurrence uniqueness partial (`external_occurrence_key IS NOT NULL`) and adds a Kilter-only chronological index. The importer conflict target includes the partial-index predicate. Migration 0232 remains unchanged; rollout now lists both migrations.
4. The development-database image failure was the legacy-table smoke guard classifying `kilter_wall_sources` as a retired Aurora table. The guard now excludes only this new table, and its SQL was checked against both valid and legacy names.
5. Timestamp checks, explicit angle-independent metadata, mid-import cancellation tests, and validation warnings are in c800de01b. The integration document already says one-way reading is implemented and flag-gated; it now documents the index changes too. Production rollout has not happened, so calling it shipped would be premature.

Pushed as 5968d38e7. All 26 Kilter integration tests, full workspace typechecking, formatting/lint, commit hooks, and migration consistency checks pass. The fresh CI run must finish before the PR is marked ready.

marcodejongh:
Scope update: Marco chose to track the two partial-index optimizations separately in #5551. Removed the optional 0233 migration, snapshot/journal entry, and corresponding schema/importer/test-fixture changes from this PR. Rollout requires only 0232 again.

The CI smoke-check fix, socket teardown snapshot, and all cancellation/Redis/remote-notification tests remain. All 26 Kilter integration tests pass against the original indexes, and the migration journal is consistent at 233 migrations. Repeated index suggestions should be tracked in #5551 rather than added back to this PR.

marcodejongh:
Rebased onto current main and resolved the subscription cleanup conflict. The Kilter watcher and Redis listener now share main's managed subscription lifetime; tests verify immediate idle unsubscribe and cancellation before Redis setup finishes.

Added diagnostics for unresolved wall mappings whose merge ancestry exceeds three links, with tests distinguishing a three-link chain from actual truncation. Documented the standalone Redis requirement for the two-key Lua script. Renamed the implemented behavior guide to `docs/kilter-live-history.md` and updated links. Optional index work remains deferred to #5551.

Validation: 157 focused backend tests pass across five files, including 31 Kilter cases. Full workspace typechecking, formatting/lint, and migration checks pass. The two disputed review claims about the existing Redis reset and exact table exclusion are awaiting Marco's decision under the repository review policy.

marcodejongh:
Marco approved retaining the existing code and dismissing these two findings with evidence:

1. **Redis subscription failures do reset the flag.** The [catch block](https://github.com/boardsesh/boardsesh/blob/25a4d3ebac8e1340a84ec5091a445ddc41322626/packages/backend/src/services/kilter-live-sync.ts#L200) sets `controlSubscribed = false`, removes the message listener, and rethrows. The [integration test](https://github.com/boardsesh/boardsesh/blob/25a4d3ebac8e1340a84ec5091a445ddc41322626/packages/backend/src/__tests__/kilter-live.test.ts#L528) verifies that a failed control subscription cleans up its listener and succeeds on retry. The proposed reset was already implemented.
2. **The smoke check allows exactly one table.** Its [predicate](https://github.com/boardsesh/boardsesh/blob/25a4d3ebac8e1340a84ec5091a445ddc41322626/scripts/dev-db-image-smoke.sh#L169) rejects tables matching `kilter_` or `tension_` prefixes unless the full table name equals `kilter_wall_sources`. `kilter_foo` is not equal to `kilter_wall_sources`, so it still fails the guard. This does not allow the rest of the namespace.

The other findings were addressed in 25a4d3eba: merge-depth diagnostics and tests, standalone Redis documentation, and renaming the implemented behavior guide. The rebase also preserves main's managed subscription cleanup. Index optimizations remain deferred to #5551. There is no remaining decision needed for these two findings.

marcodejongh:
Addressed the latest review in cb06cc7ac:

1. **Bounded coordination checks under the board lock.** The REST history fetch happens before importing, so its 15-second timeout does not hold the board lock. Redis reads during eligibility checks now each have a one-second deadline. The account lookup inside the transaction reuses that transaction's connection, avoiding a second pool allocation while holding the lock. Fresh ownership/account validation remains in place. A regression test stalls the in-transaction Redis read indefinitely and verifies that the transaction releases the lock without importing anything.
2. **Viewer-removal failures now warn.** The local viewer was already deleted before Redis I/O; the claimed leftover local-map entry does not occur. The catch now logs the board and error and still aborts the request. The new test verifies logging, cancellation, and idempotent local removal.
3. **Legacy history compatibility is intentional and documented.** The [implementation guide](https://github.com/boardsesh/boardsesh/blob/cb06cc7ac/docs/kilter-live-history.md) now explicitly explains that older clients can update current-wall state through native backfill. New clients use `boardRecentHistory` / `boardHistoryPage` for mixed history. Removing the legacy filter would weaken that compatibility boundary.
4. **The OAuth secret is optional.** [Kilter's configured public PKCE client](https://github.com/boardsesh/boardsesh/blob/cb06cc7ac/docs/kilter-sync.md#L62) intentionally omits it; only confidential deployments set it. The [token helper](https://github.com/boardsesh/boardsesh/blob/cb06cc7ac/packages/kilter-sync/src/api/keycloak.ts#L24) supports both modes. A missing-secret warning for the default public client would incorrectly report valid configuration as an error.
5. **Cache validation covers the complete supported shape.** It checks nullable string fields, integer angles, positive safe-integer sequences, and parseable timestamp strings. Malformed entries fall back to durable history. Added cases for malformed angle, frames, names, grades, attribution, and sequence values. Serial per-occurrence transactions remain intentional for cancellation/validation boundaries; optional index optimizations remain deferred to #5551.

Validation: all 48 focused Kilter/credential tests, full workspace typechecking, lint/hooks, and migration consistency pass. The repository bot rebased and renumbered the live-history migration to **0233_kilter_live_history**; this is the live-sync migration, not the deferred index optimization. Current-head CI is running before the ready-for-review signal.

marcodejongh:
The edited review still describes behavior absent from the merged head `cb06cc7ac`:

- **`frames` is validated**: it appears in [`nullableStringFields`](https://github.com/boardsesh/boardsesh/blob/cb06cc7ac/packages/backend/src/services/board-history.ts#L154), which the guard checks with `.every(...)`. Null is supported by the history representation; non-string, non-null values are rejected. The malformed-cache regression cases include frames.
- **Transaction checks reuse the same DB connection**: [`isCurrent(transaction)`](https://github.com/boardsesh/boardsesh/blob/cb06cc7ac/packages/backend/src/services/kilter-live-import.ts#L166) passes the active transaction to eligibility queries. Redis coordination reads have a [one-second deadline](https://github.com/boardsesh/boardsesh/blob/cb06cc7ac/packages/backend/src/services/kilter-live-sync.ts#L18). The 15-second REST fetch happens before the transaction.
- **Legacy history filtering is documented and intentional**: [compatibility explanation](https://github.com/boardsesh/boardsesh/blob/cb06cc7ac/docs/kilter-live-history.md#L47). Mixed-history clients use the new endpoints.
- **The OAuth secret is optional for the public PKCE client**: [configuration explanation](https://github.com/boardsesh/boardsesh/blob/cb06cc7ac/docs/kilter-live-history.md#L25).

No additional code change is needed for those repeated findings. Marco merged this PR while CI was still running; this comment does not claim all CI has completed. Optional index improvements remain tracked separately in #5551.

Issue #5545, opened 2026-09-18
Move integration exports onto pg-boss, with retries


marcodejongh:
Epic: #5613. Depends on runtime #5614; apply cutover gate #5626.

## Outcome

Session exports to integrations survive transient failures and backend restarts. Preserve the existing export API and claim behavior while adding durable jobs and remote-side-effect reconciliation.

## Current behavior

`autoSyncSessionToIntegrations()` in `packages/backend/src/integrations/export-service.ts` runs detached when a session ends. `integration_exports` records a claim, but transient failures have no durable retry owner.

## Implementation

1. Create the claim and enqueue `integration-export` in the same transaction using the existing Drizzle-aware queue adapter. Payload contains the claim ID only. Rollback leaves neither claim nor job; repeated requests return the existing eligible export.
2. Run the handler in the appropriate homelab maintenance/delivery role. Preserve provider selection, authorization, payload and user-visible API; persist attempts, remote IDs and terminal outcomes.
3. Retry transient 408/429/5xx and network failures with bounded backoff and provider Retry-After. Permanent credential/scope/payload errors become explicit failed outcomes rather than consuming retries indefinitely. Define per-family heartbeat, expiration, retention and exhausted-job reconciliation.
4. Distinguish safe retry from **ambiguous remote creation**. Use provider idempotency keys where supported, otherwise reconcile using a stable export marker/remote lookup before retrying creation. If the provider cannot resolve ambiguity, expose that state for safe reconciliation instead of blindly producing duplicate activities. Do not promise exactly-once remote delivery.
5. Remove the detached invocation after single-owner cutover; retain durable recovery for failures after remote success but before local commit.

## Acceptance

- [ ] Transient Strava 503 eventually succeeds and updates the export row; a revoked-token 401 becomes a clear terminal failure.
- [ ] Claim/enqueue transaction rollback creates neither record nor job.
- [ ] Worker restart resumes the existing claim; duplicate deliveries cannot create competing attempts.
- [ ] Remote success followed by a lost response/local commit failure is reconciled before another create; unresolvable ambiguity is visible.
- [ ] Existing `packages/backend/src/__tests__/integration-auto-sync.test.ts` behavior remains covered, with new retry/backoff and ambiguous-response cases.

## References and risk

Queue foundation PR #5547; `packages/backend/src/services/job-queue.ts`; `docs/partner-workouts-internal.md`. Medium risk: remote activity creation is user-visible and cannot rely on local transaction atomicity alone.

Issue #5544, opened 2026-09-18
Move the refresh-token cleanup onto pg-boss


marcodejongh:
Epic: #5613. Depends on runtime #5614; coordinate scheduler migration #5620 and cutover gate #5626.

## Outcome

Run refresh-token cleanup once daily through the homelab maintenance worker instead of starting timers on every backend replica. Preserve current token retention behavior.

## Current behavior

`startRefreshTokenCleanup()` in `packages/backend/src/handlers/native-auth.ts` starts an initial timeout and recurring interval from `packages/backend/src/server.ts`. `runRefreshTokenCleanup()` already contains the reusable work.

## Implementation

1. Pre-create `refresh-token-cleanup` through owner-controlled queue initialization. Runtime roles cannot create schema/queues or require DDL privileges.
2. Register one cluster-wide daily schedule in the shared registry and consume it on the maintenance role. Configure the appropriate queue/singleton policy as well as any singleton key; a key alone is not a complete concurrency guarantee.
3. Keep `runRefreshTokenCleanup()` and `REVOKED_TOKEN_RETENTION_MS` semantics. Process bounded batches, propagate errors for retries and define explicit expiration/retry/retention.
4. Add last-success/freshness monitoring and retry/reconciliation behavior for homelab outage or exhausted jobs.
5. During cutover, disable the old timeout/interval and remove its server startup/teardown hooks. Demonstrate one owner and rollback without running both triggers.

## Acceptance

- [ ] Expired and long-revoked tokens are removed with unchanged retention; valid tokens remain.
- [ ] Two backend/worker replicas result in one daily scheduled job, not one per process.
- [ ] Worker termination or transient DB failure retries safely; permanent/exhausted failure is observable.
- [ ] No refresh-token cleanup timer or `startRefreshTokenCleanup` server call remains after cutover.
- [ ] Startup works under restricted runtime DB grants, and the migration/queue setup path is covered.

## References and risk

Queue foundation PR #5547; `packages/backend/src/services/job-queue.ts`, `packages/db/src/job-queue-schema.ts`, `docs/partner-workouts-internal.md`. Low risk: daily idempotent cleanup; retain the current data policy.

Issue #5542, opened 2026-09-18
Move the room-manager inactivity sweep onto pg-boss


marcodejongh:
Epic: #5613. Depends on shared runtime #5614; apply cutover gate #5626. Related scheduler migration #5620.

## Outcome

Replace per-replica inactivity timers with one durable scheduled sweep while preserving room/session teardown on every affected backend replica.

## Current behavior

Room-manager uses a one-minute interval and a one-hour inactivity threshold. The sweep is already extracted into a callable handler. It also coordinates live backend state, so this family must remain **backend-affine**, rather than depending on a homelab maintenance consumer.

## Implementation

1. Pre-create `session-inactivity` through the owner-controlled queue migration/initialization path. Register one cluster-wide schedule every minute; configure queue/singleton policy as well as the fixed key so replicas cannot duplicate the same sweep.
2. Consume on the backend with shared runtime registration, not the homelab-only worker. Preserve **one-hour inactivity threshold** and **one-minute cadence** exactly.
3. Keep the existing extracted sweep logic but make persistent closure plus teardown intent recoverable. Publish/consume replica cleanup through the existing distributed mechanisms so each replica drops local room/session state; preserve Live Activity end behavior and kiosk tombstones.
4. Handle failure after persistent closure but before teardown/broadcast. Retain sufficient durable identity/state to retry idempotently and reconcile missed effects; do not assume a second database query will rediscover already-closed sessions.
5. Remove only the inactivity timer and its teardown after single-owner cutover. Keep heartbeat/reconnect/presence timers that serve process-local state. Homelab outage must not disable inactivity cleanup.

## Acceptance

- [ ] An inactive session is closed approximately one minute after crossing the unchanged one-hour threshold.
- [ ] With two replicas, one sweep runs per minute and both replicas release affected local state.
- [ ] Worker/process death after DB closure but before teardown recovers Live Activity termination and kiosk cleanup without duplicate destructive effects.
- [ ] Homelab workers offline does not stop this backend-affine job; backend runtime has no DDL permissions.
- [ ] Existing room-manager tests pass, new cross-replica/retry coverage proves recovery, and no inactivity `setInterval` remains.

## References and risk

Queue foundation PR #5547; `packages/backend/src/services/room-manager`, `packages/backend/src/services/job-queue.ts`, `docs/partner-workouts-internal.md`, `docs/websocket-implementation.md`. Medium risk: incorrect teardown is visible in active sessions; retain threshold and cadence.

PR #5530, opened 2026-09-17
docs: partner workout API spec (Sequence integration)

marcodejongh:
## Summary

Spec for letting a training app (Sequence first) hand a planned board session to Boardsesh and get the results back. Documentation only, no code.

Two files:

- `docs/partner-api.md` is the contract a partner builds against: Boardsesh as the OAuth 2.0 authorization server (PKCE S256 only, opaque `bsa_`/`bsr_` tokens, grant-level revocation), `POST /v1/partner/workouts` with the block schema and a per-block `support: native | freeform` flag so a plan is never rejected, the `/w/{id}` launch link, and the signed `workout.completed` webhook with its retry table and a pull fallback.
- `docs/partner-workouts-internal.md` is the design behind it: the seven tables, why consent is on web and the token endpoint on the backend, why partner tokens never pass through `validateToken`, the mobile `/w/` route and pending-workout store, how each block type maps onto the generator or the rest timer today, the outbox plus in-process sweep for deliveries, the PR order (AASA `NOT /oauth/*` ships first), and what waits on the workouts epic (#5399, #5379, #5380).

Both index bullets added to `CLAUDE.md` and `AGENTS.md`.

Design was checked against the existing Strava export path, the native-auth refresh rotation, the watch-pair flow, the room-manager inactivity sweep, and the AASA/intent-filter config. The Strava-shaped alternative (Boardsesh as OAuth client) is recorded and rejected in the internal doc. Sequence's own docs (logbook fields, RPE/RIR, their Strava and MCP integrations) informed the results payload and the mapping table.

## Release Notes

- [x] No release note needed (internal / technical change)

## Test plan

1. CI green.

## Risk

Risk: 1/5 — docs only

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_012rtbqxUXAMrX8UoaQMT1wU

PR #5547, opened 2026-09-18
feat: add pg-boss as the backend's durable job queue

marcodejongh:
## Summary

The design is `docs/partner-workouts-internal.md` ("The job queue"), which landed as a spec in #5530. This is the queue itself — start, stop, and the one helper that makes an enqueue join a caller's transaction.

**No tenants yet.** The spec wants the three existing sweeps moved onto it so it "arrives already proven", but each is a behaviour change to a live system and all three stood between here and the spray-wall detector (#5451). They are filed instead: **#5544** (refresh-token cleanup — the safest, do it first), **#5542** (room-manager inactivity sweep), **#5545** (integration exports, which also gains the retry it has never had). Two of them also fix a smaller bug on the way: those sweeps run once per replica today, where a scheduled job runs once.

## What the spec asked me to verify, and what I found

It was written against assumptions and says to check them against the installed version (12.33.0). All three mattered:

- **`send(..., { db })` really does run on a caller-supplied transaction.** Verified behaviourally on the real drizzle path in `job-queue-transaction.test.ts`: enqueue inside a transaction, roll it back, the job is gone — and a control asserting the pool path still enqueues, so "the job is gone" cannot pass for the wrong reason. The type signature alone only proves an adapter is *accepted*, not that every statement `send` issues goes through it.
- **pg-boss 12 ships an official `fromDrizzle(tx, sql)` adapter.** I had hand-rolled a `$n`-placeholder parser over `IDatabase.executeSql`; it is deleted. That was the riskiest code in the change and it turns out to be a supported API.
- **`retryBackoff` has no ceiling by default.** It grows `retryDelay * 2^retryCount`, so a 15 s base puts the fourth attempt minutes out. `DEFAULT_JOB_OPTIONS` sets `retryDelayMax: 120`; a retry should not drift past the window a climber is still watching a spinner in. The spec's retry table assumed doubling and never mentions the cap.

Two smaller ones: pg-boss 12 is ESM with a **named** export (`import PgBoss from 'pg-boss'` gives `default is not a constructor`), and it publishes `IDatabase` as `Db`.

## Where it starts, and why not in `startServer()`

In `index.ts`. The queue is a property of the **process**, not of a server instance: it holds its own Postgres pool and polls on a timer, while `startServer()` is called by hundreds of tests that want an HTTP/WS surface and no background work.

That is not a style preference — starting it in `startServer()` put `board-presence.test.ts` over the edge under full-suite load (102/102 in isolation, 1 failure in the suite). Moving it out restores the suite to exactly its pre-existing failures.

## Test plan

1. CI green.

## Risk

Risk: 3/5 — a new dependency that starts with the process and creates its own `pgboss` schema on first boot. Nothing enqueues onto it yet, so the blast radius is startup and shutdown. Noted in `docs/db-migrations.md` follow-up: that schema lives outside `packages/db/drizzle`, so `check:db-migrations` does not see it and `pg_dump` carries it.

## Release Notes


marcodejongh:
Integrated this queue foundation into #5587, including migration-owner initialization, restricted-role tests, and the spray recognition consumer. Keeping this PR open pending review of the combined rollout so we can coordinate the overlapping merge.

marcodejongh:
Closing as superseded by merged #5587 (0c0cd0b6d). Current main already includes pg-boss 12.33.0, the Drizzle transaction adapter and transaction regression tests, and startup/shutdown wiring from this PR. It also moves queue schema migrations to the deployment owner instead of runtime startup. The rebase add/add conflicts were with that adopted implementation. The separately tracked queue consumers #5544, #5542 and #5545 remain open; this closure does not mark those migrations complete.

PR #5549, opened 2026-09-18
feat(web): render one scheme — pin www to dark and delete the switch

marcodejongh:
## Summary

www is a marketing surface now. The climbing UI moved to the Expo app in W-16 (#4435), and the light/dark toggle went out with it in #4467. What stayed behind was the plumbing — a pre-paint script, an IndexedDB preference, a localStorage mirror and two MUI themes — with nothing left to drive it.

Worse, the pre-paint script fell back to `'light'` when no preference was saved and the OS preferred light. So the scheme most visitors landed in was one that was **unreachable by choice and unavoidable by accident** — nobody could pick it, and nobody had designed for it.

This pins the app to dark and deletes the machinery:

- `theme-init-script.ts` and its test go; `layout.tsx` and `global-error.tsx` stop injecting the inline script. `data-theme="dark"` stays on `<html>` as a static, documented attribute (nothing in-repo reads it; it's the contract for anything embedding a page).
- `ColorModeProvider` is **gutted, not deleted** — it also owns `CssBaseline` and the MUI X date adapter — and renamed `AppThemeProvider`.
- `use-color-mode` / `use-is-dark-mode` and their tests go, along with the six test mocks that were forcing light grade colours.
- The four runtime scheme branches collapse: `climb-title`, `similar-climbs-list`, `hold-indicator`, `use-grade-format`. `softenColor()` in `@boardsesh/play-view` keeps its `darkMode` parameter — mobile still renders light.
- The `colorMode` IndexedDB key retires through `ORPHANED_PREFERENCE_KEYS`.
- `viewport.themeColor` and `global-error`'s ground were both `#15101e`, the shared Material anchor that `velvet-tokens-parity` explicitly says web's page base is *not*. Both now read the token (`#110A20`).

**Tokens are deliberately untouched.** `buildTheme('dark')` still produces today's theme byte-for-byte, so the rendered page is unchanged for anyone already in dark. One real pixel delta though: `viewport.themeColor` moves `#15101e` → `#110A20`, so the mobile browser chrome shifts for every visitor on their next load — that is the fix, not a side effect. Collapsing `themeTokens` onto the dark values — which also fixes ~200 direct token reads that currently render light values on dark surfaces — is the next PR, kept separate because that one *is* a visual change.

Author-side verification: `vp run typecheck:web` clean; `vp test run --project web` 405 files / 4658 tests pass; `vp check` clean (the two `@gorhom/bottom-sheet` resolution errors are pre-existing in `packages/mobile` and untouched here).

## Release Notes

## Test plan

1. Open boardsesh.com on a Mac set to Light appearance.
2. The site renders dark, not pale lilac.
3. Tap Gyms, then About. Both stay dark.
4. Reload the page. No white flash before it paints.

## Risk

Risk: 2/5 — deletes the colour-mode plumbing, but leaves every token value alone.


marcodejongh:
Addressed the review:

- **Stale hydration comment** — fixed. The `<body>` comment justified the suppression as "unrelated to the theme swap on `<html>`", and there is no theme swap any more. Both suppressions stay, though: browser extensions stamp attributes onto `<html>` and `<body>` alike, which is now the entire reason, and the comment says so.
- **`themeColor` is a visible delta** — correct, and the summary undersold it. The mobile browser chrome moves `#15101e` → `#110A20` on next visit for everyone, dark-mode users included. That is the point of the change (`#15101e` is `materialSurfaces.dark.background`, the shared anchor `velvet-tokens-parity` explicitly asserts web's page base is *not*), but it is a real pixel change rather than pure cleanup. Summary updated.

PR #5553, opened 2026-09-18
feat(web): collapse the theme onto one dark scheme

marcodejongh:
## Summary

Part three of the dark-only work. **Stacks on #5549 and #5550** — merge those first; the diff shown here includes their commits until they land.

`themeTokens` now holds the dark values and `darkTokens` is gone. That is the point of the change, and it is a correctness fix more than a design one:

**~200 direct token reads across 53 files were rendering LIGHT values on a dark page.** `themeTokens.neutral.*` (107 reads) and `themeTokens.colors.*` (96) were the light set — only the CSS vars flipped with the scheme, so anything reading the constants in TS or `sx` got the wrong colour. `app/gym/[gym_slug]/page.tsx` alone had seven. `docs/ai-design-guidelines.md` documented this as a known trap and told people to route around it. They are all correct now, for free.

### What moved

- **`theme-config.ts`** — dark values under the same keys. `darkTokens`, `type DarkTokens` and `statusBg` deleted (`statusBg` folds into `colors.successBg` / `errorBg` / `warningBg`, so `buildTheme`'s ternary disappears rather than becoming a lookup). The neutral ramp keeps its numeric keys deliberately — renaming 107 call sites on top of a value change would bury the diff — but the comment now leads with the inversion: **50 is the darkest step, 900 the lightest**. `colors.successHover` stays at its light value on purpose: it feeds `palette.success.dark`, where darker is still the right direction. Flagging it rather than silently retuning it.
- **`mui-theme.ts`** — `buildTheme(mode)` de-branched. `lightShadows`, `lightTheme` and `muiTheme` deleted (grep confirms zero consumers; `muiTheme` was aliased to *light*, which was its own latent bug). `palette.mode` stays hardcoded `'dark'` — MUI derives `getContrastText`, its alpha overlays and the Skeleton/Backdrop/Rating defaults from it, so removing it is a silent restyle. `primary.contrastText` is now the **dark** ink: white on the lifted `#A78BFA` is 2.5:1, and un-branching that ternary the obvious way would have shipped it.
- **`index.css`** — the `html[data-theme='dark']` block folds into `:root`, all 52 vars name-for-name (verified: every one had a `:root` counterpart, nothing lost). **`color-scheme: dark` is hoisted with it** — it lived *only* inside that block, and dropping it would have reverted scrollbars, `<select>` popups, date pickers and autofill to light widgets on a near-black page, with no test catching it. Adds `--semantic-surface-elevated`, which `theme-config` has had all along but the CSS layer never exposed.
- **The three remaining `:global(html[data-theme='dark'])` branches flatten.** Board art now picks the dark layer unconditionally; collapsing the *fetch* to one file is a separate perf change, because the `<link rel="preload">` sites still name the light file and a half-done collapse is an LCP regression.
- **Docs** — `docs/ai-design-guidelines.md`'s web section rewritten. Its "a direct `themeTokens` read renders the light value in dark mode" warning is now false and actively misleading. Added a note in `@boardsesh/velvet-tokens` that `brandColors` (light) must not be deleted as dead code just because no web file imports it — mobile still renders light.

### The parity test gets stronger, not shorter

Halving the scheme count halves the coverage unless you add fences that only matter in a one-scheme world. New assertions:

- No `html[data-theme=…]` or `prefers-color-scheme` block can grow back in the CSS.
- `:root` declares `color-scheme: dark`.
- **The foreground violet and the fill violet are different colours**, with the contrast maths spelled out. With two schemes, the light rows (equal) and the dark rows (divergent) together said the split was deliberate. One scheme says nothing — two palette slots holding different violets just reads like an accident waiting to be "simplified".
- `primary.contrastText` is the dark ink and clears AA against the foreground violet.
- The surface ladder's luminance order holds (depth is a lighter violet, not a shadow).
- A new block covering `printSurfaceTokens` on white — the contrast coverage the OG carve-out in #5550 had none of.

Author-side verification: `vp run typecheck:web` clean; `vp test run --project web` 405 files / 4667 tests pass; `vp check` clean (the two `@gorhom/bottom-sheet` resolution errors are pre-existing in `packages/mobile`).

**Not yet done: the by-eye sweep.** The shared dev Postgres currently has a half-applied `0226_spray_wall_tables` from another in-flight worktree (`type "spray_hold_source" already exists`), so `vp run db:up` fails and the dev server won't boot. QA notes are written at `.boardsesh/qa-notes.md`. The manual pass — native form chrome, the gym poster staying black-on-white, the OG unfurls, the admin screens — still needs doing before this leaves draft.

## Release Notes

## Test plan

1. Open any page on a Mac set to Light appearance. It renders dark.
2. Open a dropdown and the page scrollbar. Both are dark, not light grey.
3. Open a gym poster page. It stays black on white.
4. Share a profile link into Discord. The preview card is white with dark text.
5. Open a dialog. It is a lighter violet than the page, with a thin edge.

## Risk

Risk: 3/5 — changes every colour constant the app reads; ~200 call sites shift value.


marcodejongh:
## Verified against a running dev server

Spun up a local dev DB from the pre-built image (the shared Tailscale one had a stuck `0226_spray_wall_tables` — see below) and swept the site with the browser emulating a **Light-appearance machine**, which is the case that used to break.

Every page, measured from the painted pixels rather than eyeballed:

| page | body background | `color-scheme` |
| --- | --- | --- |
| `/`, `/about`, `/support`, `/help`, `/legal`, `/privacy`, `/gyms`, `/gyms/kilter`, `/auth/login`, `/docs` | `rgb(17, 10, 32)` = `#110A20` | `dark` |

All ten. On a light-appearance browser. That is the fix.

**The regression this PR was most likely to cause did not happen.** `/api/og/setter` still renders white card, dark text — the `printSurfaceTokens` carve-out from #5550 holds under the collapse. (Screenshot attached to the session; `profile`/`playlist` return 400/404 on a fresh DB for lack of seed rows, but they share the same code path and the same token import.)

Also confirmed by the sweep: the marketing header bar is dark rather than the old white gradient, the gym search field rides the elevated violet, facet chips keep their borders, and the hero CTA's amber glow survives.

### Two things the scan flagged, neither a regression

- **`leaflet-container` on `/gyms`** — Leaflet's own container plus light map tiles. Untouched by this PR and already called out in the gyms wireframe (#5552) as a dark-tile task for the gyms restyle.
- **`MuiCardActionArea-focusHighlight` on `/`** — MUI's focus overlay span. Transparent in the render; false positive from my detector reading its computed background.

### Not covered

Admin screens (need auth) and the gym poster (this DB has no gym rows). The poster's stylesheet is untouched by this diff and uses raw literals by design, so it cannot have moved. QA notes for a human pass are at `.boardsesh/qa-notes.md`.

### Unrelated: the shared dev DB is still broken

`0226_spray_wall_tables` applied **fully** on the Tailscale dev Postgres — both enum types, all four `spray_*` tables, every index and FK, and `board_climbs.missing_hold_count` — but the row in `drizzle.__drizzle_migrations` is missing, so every run re-attempts it and dies on `CREATE TYPE`. Nothing from 0227+ landed, the spray tables are empty and the column is NULL across all 895,453 rows. The fix is to record 0226 as applied, not to drop and re-run. Not done here; it belongs to whoever owns the spray-wall worktrees.

marcodejongh:
Both fixed.

**The badge tint was a real bug — good catch.** The base `.badge` / `.badgeMirrored` rules carry no `background-color` at all, so the dark-scheme override *was* the tint rather than an adjustment to it. Flattening deleted it instead of folding it in, and the badges would have shipped as pure backdrop blur. `var(--overlay-light)` now sits on the base rule with a comment saying why it is load-bearing: these badges sit on board art, and on bright art blur alone leaves the icon fighting the image.

**Board-art comment** retitled — it still said "theme swap" for a pick that no longer swaps. The deferral note about the fetch and the `<link rel=preload>` hazard stays, since that is still true.

PR #5555, opened 2026-09-18
feat(web): give the long-form pages one shared frame

marcodejongh:
## Summary

**Stacks on #5553** — merge that first; the diff here includes its commits until it lands.

`/about`, `/help` and `/support` each owned a near-duplicate CSS module (`diff about.module.css support.module.css` was identical for lines 1–59), `/legal` imported `/about`'s outright, and `/privacy` inlined its own local `Section` and `Paragraph` helpers. Five pages, five copies of the same `.pageLayout` / `.header` / `.content`.

This adds `app/components/ui/page-shell` — `PageShell`, `PageSection`, `PageCard`, `Prose`, `ProseList` — and puts all five pages on it.

### It kills the double header

`/support`, `/about` and `/legal` each rendered their own bar (back arrow, logo, title) **underneath** the global `MarketingHeader`, then padded for it with `var(--global-header-height)`. Two stacked bars on every one of them. The in-page bars are gone; the global header is the header.

`MarketingHeader` itself is untouched. `TITLE_HEADER_PAGE_PREFIXES` could not absorb them: its `CenteredHeader` renders `component="h1"`, which would put a **second** `<h1>` on `/support` and fail `support-content.test.tsx:89-95`; and that branch renders no nav, while `site-chrome.spec.ts` asserts the four nav anchors **on `/about` itself**. So the bars die instead.

### Server-safe on purpose

No `'use client'`, no hooks anywhere in the shell. `/privacy` and the gym directory are React Server Components — a client shell would drag them over the boundary. Verified: `/privacy/page.tsx` still has zero `'use client'`. Strings arrive as props already resolved by the caller's `t()`, which also keeps `check:i18n` happy because the literals stay in the page files.

### Other notes

- `/about` and `/legal` **gain a real `<h1>`** — neither had one.
- `--global-header-height` now appears in exactly one place.
- The shell uses `min-height`, never a trailing `margin`, because `site-chrome.spec.ts` measures `scrollHeight − footerBottom < 24` and a bottom margin reads as a dead gutter under the footer.
- `PageSection` carries `scroll-margin-top` so `/help#climb-counts` (linked from `docs/kilter-sync.md`) still clears the fixed header.
- Deletes `about.module.css`, `help.module.css`, most of `support.module.css`, the cross-page import in `legal-content.tsx`, and the `support`/`about`/`legal` `headerTitle` keys in all four locales.
- `components/ui/page-container.module.css` is deliberately **not** extended — it is a playlists module wearing a generic name (`.heroSquareEmoji`, `.climbsSection`, a `:global(:root)` declaration, `max-width: 1400px`). Migrating playlists onto this shell is a later PR.

**Not a copy change.** Every string is the one that was there before. The `/support` redesign — real buttons instead of text links, the heading that stops saying "Two ways to chip in" above one way, and the "other ways to help" section — is the next PR.

Author-side verification: `vp run typecheck:web` clean; `vp test run --project web` 406 files / 4677 tests pass (10 new for the primitives, including the `data-testid` passthrough the Stripe rail depends on); `vp run check:i18n` and `vp run check:i18n:orphans` both clean; `vp check` clean bar the two pre-existing `@gorhom/bottom-sheet` errors in `packages/mobile`.

Swept against a running dev server with the browser emulating a **Light**-appearance machine — all five pages: exactly one `<h1>`, exactly one `<main>`, body `rgb(17, 10, 32)`.

## Release Notes

- The About, Help, Support, Legal and Privacy pages read as one site now, with one header bar instead of two.

## Test plan

1. Open Support. One bar at the top, not two.
2. Scroll down. Text sits in a readable column, not a box.
3. Open About, Help, Legal, Privacy. Same frame each time.
4. Tap Help, then a link into a section. It lands below the bar.
5. Narrow the window to phone width. Nothing overlaps.

## Risk

Risk: 2/5 — layout-only across five static pages; no data, no auth.


marcodejongh:
Thanks — this review is reading the stacked diff, so two of the three land on #5553. Handled:

**Stale comment in `index.css` — real, fixed** (on #5553, `a48...`). The block above `--bs-text-brand-*` still pointed at `text.brandPrimaryLight` / `text.brandMutedLight`, which that PR deletes, and still described "the dark-mode block below", which it also deletes. Now points at the surviving pair and says where the `*Light` ones went.

**Board-art double fetch — tracking issue filed: #5558.** It records why it was deferred, which is the part worth not losing: the `<link rel="preload">` sites in `page.tsx:47`, `playlists/page.tsx:39` and the board list/view pages still name the **light** file, so a half-done collapse is an LCP regression rather than a win.

**Parity-test coverage regression — I don't think this one holds.** The row-level TS↔CSS assertions are still there, at `velvet-tokens-parity.test.ts:64-92`: a 26-row table fed to `it.each(rows)('%s matches theme-config')`, and `--bs-text-brand-primary` is row 25 (`themeTokens.text.brandPrimary`). What changed is the tuple shape — `[cssVar, light, dark]` became `[cssVar, value]`, because there is one scheme to compare against instead of two. The scenario described (someone edits `theme-config.ts` and forgets the CSS var) still fails that table.

The structural fences are *additional*, not a replacement, and they cover what the two-scheme version used to imply for free — chiefly that the foreground and fill violets are different colours on purpose, which with one scheme nothing else says. Happy to be shown otherwise if you're looking at a specific row that went missing.

marcodejongh:
Closing as already landed — no rebase needed.

#5557 was stacked **on top of** this branch, so when it was squash-merged (`888d55322`) it carried this PR's commits in with it. Everything here is in `main`:

- `packages/web/app/components/ui/page-shell/` — all six files plus its tests ✅
- `legal-content.tsx`, `privacy/page.tsx`, `help-content.tsx` converted ✅
- `about.module.css` and `help.module.css` deleted ✅
- the `support`/`about`/`legal` `headerTitle` keys removed from all four locales ✅
- the stale `--bs-text-brand-*` comment fixed ✅

I checked rather than assumed: replaying this branch's commits onto current `main` produces an **empty** commit for the comment fix, and the only conflict is the page-shell commit trying to put `/support` back to its pre-#5557 state. There is nothing left here to merge — rebasing it would just be a no-op with a revert hazard attached.

Stack for the record: #5553 → this → #5557, all three now in `main`.

PR #5559, opened 2026-09-18
refactor(web): one definition for the filled violet pill

marcodejongh:
## Summary

`marketing-header.tsx`, `site-footer.tsx`, `help-content.tsx` and `home-page-content.tsx` each carried their own near-identical `*_SX` object for the same button. They differed only in horizontal padding and, in the hero's case, an amber glow — and they had **already drifted**: only the hero had the glow, only the header set `flexShrink`.

`brandCtaSx({ size, glow })` in `app/components/ui/brand-cta.ts` replaces all four. Called at module scope and hoisted, exactly as the constants it replaces were, so nothing allocates per render.

The glow literal `rgba(255, 138, 61, 0.35)` becomes a token — `--shadow-accent-glow` in `index.css` — with a note that amber is fill-only in Velvet Send and this is the one place it appears as *light*, so at most one surface per page should carry it.

The helper documents the trap it sits next to: the fill is `--color-primary-fill` (`#7C3AED`), **not** `--color-primary` (`#A78BFA`), because white on the foreground violet is 2.5:1.

Verification: `vp run typecheck:web` clean, `vp test run --project web` 406 files / 4677 tests, `vp check` clean bar the two pre-existing `@gorhom/bottom-sheet` errors in `packages/mobile`.

## Release Notes

## Test plan

1. CI green.

## Risk

Risk: 2/5 — shared button styling touches the header, footer and home CTA.


marcodejongh:
Both fixed in `4b2…`:

- **The falsy spread** is `{}` now rather than `null`. It worked, but it reads like a mistake.
- **The `as object` casts are gone** — you were right that they discarded the style typing, so a mistyped key in a spread would have sailed through. The fix was to stop annotating the return as `SxProps<Theme>`: that type is a union which also admits an array and a callback, which is precisely why the result couldn't be spread and why the cast existed. The return type is inferred and checked with `satisfies` instead, which keeps the key checking without widening. The one at the `home-page-content.tsx` call site went with it.

Tried `SystemStyleObject` from `@mui/system` first — not resolvable from `packages/web` — and `CSSObject`, which rejects MUI's `px`/`py` shorthands. `satisfies` is the one that gives the checking without a cast.

Issue #4381, opened 2026-08-12
publicGyms enumeration query + gyms sitemap shard (supersedes #3643)


marcodejongh:
Part of #4372. **Blocked by:** #4358 W-22 (sitemap-index infra — a locked cross-epic edge) AND Tier 0 dedup gate exit (read #4373's numbers; link them in the PR description at merge time). Soft dependency on #3666 so the shard's hub page exists when crawlers arrive. Supersedes #3643 — close it with a pointer here.

## Context

Public gym pages are crawl-orphans: the tracked comment in `packages/web/app/sitemap.ts` (lines 20-23) explains gym pages are unlisted because no dedicated public-gyms enumeration exists — that claim is the code comment's, not the dossier's. DOC-08 in fact notes `searchGyms` has **no required field**, so DB-side enumeration for sitemaps is already possible. A dedicated `publicGyms` query is still the right build, for reasons `searchGyms` doesn't cover: `updatedAt` in the payload (real `lastModified` timestamps), the sitemap eligibility predicate (live canonical public gyms only), and rate limiting sized for an unauthenticated enumeration endpoint. Without the shard, 4,740 gym pages (DB-01) rely on crawl-through links from 3 kiosk pages, organic search is ~5-6 pageviews from 2 users in 90 days (PH-07), and hypothesis H5 cannot be tested.

This ships only after the dedup gate exits: indexing a directory with live duplicates is an SEO own-goal, indexed duplicates persist in Google's cache long after a merge (DOC-03), and merged gyms must keep resolving because printed QR codes exist in the field.

## The work

- **Backend (`packages/backend`, GraphQL — not the Next.js app):** a public, rate-limited, paged `publicGyms` enumeration query returning slug + updatedAt for sitemap-eligible gyms only: `is_public`, `deleted_at IS NULL`, `merged_into_gym_id IS NULL`, `slug IS NOT NULL`, and at least 1 linked live board OR a live kiosk (the #3666 gating rule).
- **Web:** a gyms sitemap shard registered in Epic 1's W-22 sitemap index, emitting canonical slugs only, with hreflang alternates for all four locales via the existing `buildLanguageMap` pattern. ~4.7k gyms × 4 locales ≈ 19k URLs — a single shard under the 50k limit; assert the count in a test so growth past the limit fails loudly.
[...]

Issue #4372, opened 2026-08-12
Epic: marketing + gym discovery on www


marcodejongh:
We host 4,740 live gym listings (DB-01) that almost nobody can find: all public gym pages combined drew 78 pageviews from 23 users in the last 90 days, no single gym page reached more than 3 distinct viewers, and true organic search accounted for ~5-6 pageviews from 2 users (PH-05, PH-06, PH-07; window 2026-05-14 → 2026-08-12). Meanwhile #4358 deletes the www climbing surfaces and, with them, the homepage's only gym-discovery affordances (DOC-08) — after that teardown, www has no gym discovery UI unless this epic builds one. This epic makes www earn its keep as the marketing + SEO + gym surface: a public /gyms directory with board-type facets, a claim funnel that doesn't require knowing a human, listing fields worth filling, a printable QR kit with scan attribution baked in, and the instrumentation to know whether any of it works. Grounded in the 2026-08-12 telemetry/DB audit (PH-01..16, DB-01..10, DOC-01..09) and a fact-checked simulated gym-owner panel.

## The honest state

- **Supply without demand.** 4,740 live gyms; 2,896 of the 2,897 SYSTEM-synced ones have at least one linked board and no owner account behind them (DB-01, DB-10). Gym pages drew 78 pv / 23 users in 90 days (PH-05).
- **The funnel is dark.** Zero of 445 App Install Clicks fired on a gym page (PH-13); no event exists for QR scans, claim CTAs, or kiosk-link clicks (PH-15); the claim funnel has zero instrumentation and zero pageviews (PH-09, PH-10); gym search is invisible in telemetry (PH-12).
- **Claiming requires knowing a human.** 26 claims ever, all `method='admin'`, all founder-processed (DB-02). The claim call-out is invisible to signed-out visitors (#3672), and the dialog's domain fast-path is structurally dead: only 5 of 4,740 gyms have a website on file (DB-09).
- **Directory cards would look dead today.** `image_url` filled on 0 of 4,740 gyms, logo 1, description 5, website 5; hours columns don't exist — the planned migration 0187 never landed (DB-09, DOC-09). Name (100%), map pin (63.1%), and board list (98.7%) are the only reliable card fields (DB-06, DB-04).
- **Dedup is stalled and unmeasured.** ~46-60 physical duplicates estimated (DOC-03); 8 merges ever, all in one batch on 2026-07-22, zero since; no pending-queue table exists to track the backlog (DB-07).
[...]

PR #5563, opened 2026-09-18
feat(web): restyle the gym directory and gym page for the dark ground

marcodejongh:
## Summary

The gym surfaces were the last www pages still drawn for the light scheme. Four concrete dark-ground problems, fixed:

1. **The card grid had no grid.** `GymDirectoryCard` was `1px solid var(--neutral-200)` over nothing — on the near-black page ground that is 24 near-invisible outlines. It renders through `PageCard` now: a surface fill plus the one hairline, raising to the elevated surface with a violet border on hover. The fill is what makes a row of cards read as a row of cards.
2. **The filter row vanished.** The facet chips were default outlined MUI `Chip`s, which on dark is a barely-there border and no fill. Both states are spelled out now — surface + separator at rest; elevated surface, violet border, violet label and a soft ring when current, plus `aria-current="page"`. The near-me radius toggles and the pagination controls got the same two states, so every "which one is selected" on the page answers the same way.
3. **The map carried literal white.** `gym-directory-map.tsx` built its Leaflet `divIcon` markup as an HTML string with `#fff` rings and `rgba(0,0,0,0.4)` shadows. Those are CSS custom properties now, as are the Leaflet chrome the stylesheet hardcodes white: the container ground (its default pale grey flashed at every edge the tiles do not reach), the zoom buttons, the attribution bar and the popup. The `divIcon` string is the one place the repo's `style`-prop ban cannot apply — Leaflet takes a string and owns the element — and the file now says so, at length, so nobody "fixes" it. The rings stay light rather than becoming the page ground: OSM's standard raster tiles are a light basemap, and a dark ring on light streets is worse than what ships today. Retuning them is the day the tile source becomes dark (mockup note 7).
4. **Two branches nobody had designed.** Zero results rendered two bare `Typography` lines; a failed fetch rendered three. Both get the same designed panel now — card surface, a ringed glyph (brand for empty, amber for the outage), one line of what happened, and the ways out. The empty panel's actions only render when they would change the query, so a bare `/gyms` with nothing to clear shows neither.

Framing: `/gyms`, `/gyms/kilter`, `/gyms/moonboard`, `/gyms/tension` and `/gym/[slug]` all sit in `PageShell width="wide"` now, both directory branches included. No page file hand-rolls the fixed-header offset any more — the shell owns it. The directory h1 still comes from `facetHeading(t, facet)` and still survives an outage. Each page gained a Home › Gyms breadcrumb, which is also the gym page's only route back into the directory after a poster scan. `PageCard` learned a `component` prop so a card can be the `li` of a real list.

**Indexing is deliberately untouched.** All four directory routes stay `noindex, follow`. That is gated on the duplicate-gym dedup queue and the gyms sitemap shard (#4372 / #4381), not on how the page looks; flipping it belongs in its own one-line PR after the queue drains.

Author-side verification: `vp run typecheck:web`, `vp test run --project web` (4677 tests, 406 files, all green), `vp run check:i18n`, `vp run check:i18n:orphans` and `vp check` all pass. The only `vp check` errors are two pre-existing `no-floating-promises` hits in `packages/db` test scripts this branch never touches.

Two open points for the reviewer, both one-prop changes:

- The wireframe's note 1 puts the **gym detail page** at `width="prose"`; the brief said `width="wide"` for both files. This PR uses `wide` and keeps the long-form blocks (wall shot, description, hours, install copy) at a 68ch/800px reading measure inside it, so a paragraph never stretches to 1200px. Say the word and it becomes `prose`.
- Adopting the shell moves the gym page's h1 above the wall shot instead of beside the logo. The logo now sits with the address. The shell owns the h1, so that ordering is the cost of the frame.

New copy, added to `en-US` / `es` / `fr` / `de`: `gyms.breadcrumb.home`, `gyms.breadcrumb.gyms`, `gyms.results.emptyClearSearch`, `gyms.results.emptyBrowseAll`, `gyms.error.retry`, `gyms.error.home`, `kiosk.gymPage.breadcrumbGyms`. Spanish keeps *rocódromo* / *plafón*, French *salles*, German informal *du* and *Hallen*.

## Release Notes

- The gym finder and gym pages now read properly in the dark: cards you can actually see, a filter row that shows what you picked, and a real answer when a search comes up empty.

## Test plan

1. Open Gyms. Cards show a filled panel, not hairlines.
2. Tap the Kilter filter. That chip fills violet and stands out.
3. Search a nonsense word. A panel offers ways back.
4. Scroll to the map. Pins and zoom buttons read clearly.
5. Open any gym. Breadcrumb shows Boardsesh › Gyms.

## Risk

Risk: 3/5 — presentation only across five public routes, no data or query changes, but it touches the shared PageCard primitive and every gym surface at once.

https://claude.ai/code/session_01XY41pnHH3Trr5YVDEAF3r6


marcodejongh:
On the two things you flagged:

**Gym page width — keep `wide`, don't flip it to `prose`.** The mockup note and my brief disagreed, and what shipped is better than either. The gym detail page carries *both* long-form copy (description, hours, install text) and card grids (boards, spray walls). A blanket `width="prose"` would give the prose the right measure and then crush the grids into a 760px column; `wide` plus the per-block `maxWidth: '68ch'` caps you added gives each content type the width it actually wants. That is the right shape — the mockup's note was written before the shell had a per-block escape hatch. Leaving it.

**Hero reordering — fine.** The h1 moving above the wall shot is the shell doing its job, and it matches every other page on the site now. A gym page whose h1 sits beside a logo mid-column was the odd one out.

**`PageCard` gaining a `component` prop** is a good catch — a card that is the `li` of a real list is exactly what a directory grid should be, and `Box`'s own generic doesn't survive `ComponentProps`. That improves the shared primitive for everyone.

On the three you scoped out (claimed badge, the location-off hint, a dark tile source): agreed, all three are their own change. The dark tile source is the interesting one — it is why the marker rings stay light rather than becoming the page ground, and it needs a tile provider decision rather than a CSS one.

marcodejongh:
Ran the chrome e2e gate against this branch, since it is the one PR here that changes a page frame.

**106 passed. 4 failed — and the same 4 fail identically on clean `main`**, same test names:

- `site-chrome.spec.ts:251` / `:255` — WebSocket counts on kiosk and embed routes
- `embed-headers.spec.ts:203` — raw embed requests after a page load
- `expo-web/smoke.spec.ts:97` — expo-web boot

So nothing in this branch regressed them. The assertions that would actually catch a shell mistake — header and footer present, the four nav anchors, the footer carrying every static-sitemap link, and `scrollHeight − footerBottom < 24` (the dead-gutter check that a trailing margin on the shell would break) — are all in the 106 that passed.

Worth someone knowing separately that those four are red on `main` too. They may be artefacts of a local Docker database rather than genuinely broken in CI, but I did not chase that here.

Also checked by hand, since static tooling will not: the six CSS custom properties your `divIcon` strings reference all exist in `index.css` (`--border-radius-full`, `--shadow-sm`, `--color-primary-fill`, `--color-on-primary`, `--neutral-900`, `--semantic-selected`), the new gym keys are present in all four locales with the right native terms, and both pages render exactly one `h1` — from the shell, with no second one left behind.

PR #5570, opened 2026-09-18
feat(web): add the "Find a board near you" homepage section

marcodejongh:
## Summary

Builds the **"Find a board near you"** gym-search block from the approved marketing-refresh mockup (`docs/design/marketing-refresh/homepage.html`, notes 5-8) as one self-contained section component. **Nothing is wired into the homepage here** — `home-page-content.tsx` and `page.tsx` are untouched, so this lands green and dormant and a follow-up mounts it.

**It is a React Server Component**, and that is the design. The eyebrow, the `<h2>`, the intro line, the board-type links and every gym card's `<a href>` are in the first HTML response, so the site's highest-traffic page finally carries real gym names and real internal links instead of one "find a gym" card pointing at `/gyms`. The search box is a plain `method="get"` form to `/gyms`, so typing a town and pressing enter works before hydration and with JavaScript off. The single client island is the geolocation button, which can't be anything else — it rounds the fix to ~110 m and navigates to `/gyms?lat=…&lng=…&radius=25` rather than running a second near-me implementation the directory already does better.

**Reuses the directory instead of growing a second one.** `fetchDirectoryPage` on the unfiltered first page — byte-for-byte the query `/gyms` itself runs, so the teaser and the directory's page one share one cache entry — plus `fetchFacetCounts` for the chip numbers, `GymDirectorySearchForm` for the form and `GymDirectoryCard` for the cards. Facet chips point at the literal `/gyms/kilter` routes, never `?boardType=`, so nothing self-canonicalises away (note 7). The section mounts its own nested `I18nProvider` for the `gyms` namespace, so it carries its own dependencies and the wiring PR is a one-liner.

**Fails soft.** Either fetch reporting `{ ok: false }` costs the cards or the counts and nothing else: the heading, the copy and the crawlable `/gyms` anchor render no matter what the gym backend is doing, and a dead count query drops the number from the intro rather than printing a confident "0 gyms".

New files:
- `packages/web/app/components/home/home-gym-search.tsx` — the section (RSC, no props)
- `packages/web/app/components/home/home-gym-search-near-me.tsx` — the geolocation button
- `packages/web/app/components/home/__tests__/home-gym-search.test.tsx` — 9 tests

New i18n keys in all four locales under `marketing.json` → `home.gymSearch.*`: `eyebrow`, `title`, `lead_one` / `lead_other`, `leadNoCount`, `geoHint`, `cardsUnavailable`, `browseAll`. Everything else reads existing `gyms` keys.

Verified: `vp run typecheck:web`, `vp test run --project web` (4,686 passing), `vp run check:i18n`, `vp run check:i18n:orphans`, `vp check` (the two remaining errors are a pre-existing `@gorhom/bottom-sheet` resolution failure in the mobile web shim, untouched here).

## Release Notes

The homepage now has a real gym search: type a town or tap your location and find the walls near you that run a Kilter, Tension or MoonBoard.

## Test plan

1. Open the homepage → "Find a board near you" appears.
2. Type a town, press enter → gym results page opens.
3. Tap "Use my location", allow → nearby gyms listed.
4. Tap a board chip → that board's gym list opens.
5. Tap a gym name → that gym's page opens.

## Risk

Risk: 2/5 — a new, unmounted marketing section; it reuses the existing gym directory data path and fails soft when that path is down.


marcodejongh:
Integration follow-up in #5572: added homepage geolocation tests (press-only permission, rounded locale-aware navigation, denied/unavailable/unsupported fallbacks, and loading). Keeping the all-gyms facet as the mockup specifies; the lower directory link is a separate end-of-section CTA. On rendering: app/page.tsx already calls getLocale(), which awaits next/headers, so removing the session read does not restore static rendering. Confirmed with Marco that accurate server-resolved claim analytics should stay; directory query results retain their existing data cache.

Issue #5573, opened 2026-09-18
Prioritize donating gyms in homepage recommendations


marcodejongh:
## Goal

Give gyms that help fund Boardsesh priority in the homepage gym previews, following the claimed-first ordering introduced by the marketing refresh (#5572).

## Follow-up decisions

- Define how a donation is associated with a gym, when eligibility starts and ends, and how refunds or cancellations affect it.
- Choose the order of donating, claimed, and unclaimed gyms, with deterministic ties and unclaimed fallback.
- Decide how sponsored placement is disclosed and reconcile it with the current support-page promise that donating buys nothing extra.
- Keep public directory and proximity results unchanged unless separately approved.

## Acceptance

Implement the ranking in the backend before pagination, preserve public/linkable eligibility, and test donor eligibility changes, stable ordering, and fallback. This issue does not authorize payment or donation-schema changes in the current marketing refresh.

PR #5572, opened 2026-09-18
feat(web): finish the marketing refresh

marcodejongh:
## Summary

Finishes the marketing refresh, including a second design/marketing pass after Marco flagged the missing logo and disjointed composition.

- Genuine logo/wordmark restored across shared chrome; split editorial hero, benefits before discovery, consistent headings/gutters, balanced support layout and readable gym cards.
- Mobile typography refined at 320–430px: full-width feature copy, consistent heading scale, readable board labels, and 44px actions. Gym boards use compact unframed thumbnails beside their details instead of stretched image panels.
- Popular boards are real, named gym installations, ranked by distinct climbers on the physical board. Each homepage gym card previews up to three of its boards, with selected-climb holds when confirmed. App actions and climb navigation preserve the same shared board identity rather than creating configurations.
- New bounded public board discovery query excludes private, unlisted, location-hidden, merged and moderated boards. Read-only Redis snapshots require a matching active writer; anonymous or unverifiable selections show unlit boards. Website snapshots refresh after 30 seconds and expire after 60 seconds; there are no subscriptions or BLE writes.
- Real play-view and queue captures, registered in the immutable asset catalog; the hero now owns image preload priority.
- Shared claimed/unclaimed gym badges, translated location hints, and a visibility-gated OpenFreeMap dark basemap with attributed OSM fallback and matching marker contrast.
- Homepage-only claimed-first recommendations, applied before backend pagination by #5574. The directory keeps its existing order. Donating-gym priority is tracked separately in #5573; board-art double fetching remains #5558.
- Resolves prior review findings: empty catalog is not an outage; desktop renders both stores; near-me permission/fallback coverage; explicit install-platform handling; server-only guards; isolated feature CTA translations.

Verification: full web/locale suites, targeted backend discovery/privacy tests, and mobile/desktop Playwright regressions pass. Web/backend/shared types, formatting/lint, i18n usage/orphans, and static-asset catalog checks pass. Independent visual QA covers desktop/phone composition, all four locales at 320–430px, genuine logo loading, iPhone store selection, map attribution, and named-board navigation. Local missing beta thumbnails now show intentional placeholders, including failures before hydration.

#5570, #5571 and #5574 are merged. The new additive board discovery query ships in this PR; its optional website previews fail soft while backend deployment catches up. Accurate server-side claim analytics are retained with Marco's approval: locale detection already makes the homepage request-rendered, while directory query results remain cached.

## Release Notes

Find your next board night from the refreshed Boardsesh website.
[...]

PR #5574, opened 2026-09-18
feat(backend): support claimed-first gym recommendations

marcodejongh:
## Summary

Adds opt-in claimed-first gym ordering for the homepage marketing refresh (#5572). Ranking happens before pagination and retains the existing newest-first or distance ordering within each claim group. Default callers keep their current ordering; unclaimed and sparse gyms remain eligible.

Validation: 33 focused backend tests passed, including real-database pagination and unchanged default ordering. Repository typecheck and pre-commit checks passed. No migration is needed. Deploy this addition before the homepage opts in. Donor ranking is tracked separately in #5573.

## Release Notes

## Test plan

1. CI green.

## Risk

Risk: 3/5 — optional resolver ordering with unchanged defaults and database coverage.


marcodejongh:
Addressed the minor review findings in 21013c730: joined ORDER BY fragments with sql.join, used the Drizzle owner column, made the tied-ID expectation explicit, and added text + proximity + claimed-priority pagination coverage. All 35 targeted backend tests and backend typecheck pass.

marcodejongh:
Final review follow-up in 5547aecef adds explicit prioritizeClaimed:false compatibility for both query paths and normalizes the text-order placeholder assertion. Local backend query confirms three claimed fixtures precede unclaimed fallback before limit:4. No index migration is needed for this opt-in, cached homepage query at the current catalog size.

PR #5583, opened 2026-09-19
fix(web): give interactive borders a 3:1 token, and close the glow parity gap

marcodejongh:
First PR of the marketing polish pass. Screenshot review of the five top-level pages found that the same thing is drawn several different ways per scroll, and one root cause is that there is no token for "the edge of an interactive thing" — so everything reached for `--separator`.

## The problem

`--separator` is `rgba(185, 170, 215, 0.2)`. Composited, it lands at:

| on | ratio |
|---|---|
| `--semantic-background` | 1.43:1 |
| `--semantic-surface` | 1.50:1 |
| `--semantic-surface-elevated` | 1.49:1 |

WCAG 1.4.11 wants ≥ 3:1 for the visual boundary of a UI component. That would be survivable if the border were decoration on top of a visible fill — but for every control involved here it is the *only* affordance. Their fill is `--semantic-surface` on `--semantic-background`, which is **1.19:1**, or there is no fill at all. `onboarding-card.tsx` already said so in a comment: "#110A20 is nearly invisible".

## The fix

`--control-border: #837A9C`, for the boundary of interactive things. `--separator` keeps the decorative role it was designed for.

**The token is opaque on purpose.** A translucent border's contrast depends on whatever sits behind it, so the same token can pass on the page and fail on a card, and no single assertion can pin it — which is how this drifted in the first place. Opaque makes 1.4.11 one arithmetic fact:

| on | ratio |
|---|---|
| `--semantic-surface-elevated` (worst case) | **3.58:1** |
| `--semantic-surface` | **4.03:1** |
| `--semantic-background` | **4.81:1** |

`#837A9C` is `--neutral-400` lifted a step toward `--neutral-500` and nudged violet, so it sits in the existing family. ~19% headroom on the worst surface, so a future surface retune will not silently drop it under 3.

## What moved, and what deliberately did not

Seven call sites, chosen by one test — *is the border the only thing identifying this control?*

- both facet chips (`home-gym-search`, `directory-page`)
- "Use my location" (`home-gym-search-near-me`)
- the radius toggle (`gym-directory-near-me`)
- pagination pills (`gym-directory-pagination`)
- the onboarding cards (whole card is a `CardActionArea`)
- the outlined install button (`marketing-install-links`)

Keeping `--separator`, because they are decoration or a surface rather than a control boundary:

- dividers, hairlines, rules, the map frame and the pinned-count badge
- the `StatePanel` glyph ring
- `elevatedPaper` (dialog/drawer/menu/popover) — its own comment describes it as "a hairline separator border" making floating surfaces read as one Velvet layer. That is a surface edge, and a deliberate documented choice; not mine to overrule here.
- the hero secondary CTA — it moves when the outlined CTA recipe lands, so the switch happens once, in one place, rather than twice.

This is fewer sites than the original sweep proposed. I narrowed it to the ones where the contrast failure actually costs a user something.

## Glow parity

`--shadow-accent-glow` and `--shadow-accent-glow-hover` existed **only** in `index.css`, despite that file's header claiming it mirrors `theme-config.ts` and the parity test asserting exactly that. The test could not see them. Now mirrored.

**Mirrored, not retuned.** A reviewer flagged the warm-amber glow under the violet hero pill as looking muddy, but `index.css` documents it as intentional — "Amber is fill-only in Velvet Send, and this is the only place it appears as light rather than a fill — it is the spark, so exactly one surface should carry it." That is a design decision with a stated rationale, so this PR only closes the parity hole. Worth a separate look with a designer if the team wants it revisited.

## Tests

Three new fences in `velvet-tokens-parity.test.ts`, because a split like this re-drifts the moment nothing tests it:

1. `--control-border` clears 3:1 on page, card and elevated surface.
2. `--separator` stays **below** 3:1 — not a tolerated bug, the reason the split exists. If someone "fixes" it to pass, it stops being a hairline, and the test says so before review has to.
3. `neutral[400]` is recorded as failing at 2.72:1 on the elevated surface — it is the obvious thing to reach for, so the next person should not have to re-derive why it does not work.

Plus a two-line rule comment above both declarations in `index.css` saying which token is for which role.

## Verification

- `vp check` — 2 errors / 380 warnings, **identical to `origin/main`** (both in `packages/db` migration tests, pre-existing and untouched here).
- `vp test run --project web` — 4752 passed, 418 files.
- Pre-commit hook clean, including `check:i18n` and `check:i18n:orphans`.

No geometry change: every edit is a colour value on an existing border. Nothing gains or loses a border, and no box dimension moves.

## Release Notes

Buttons, filters and page controls now have a visible edge. Previously their outline was faint enough to disappear against the dark background, which made it hard to tell what was tappable — especially outdoors or on a dim phone screen.

## Test plan

1. Open the gyms page. Board filter chips show a clear edge.
2. Tap a chip. The selected one turns violet.
3. Scroll down. Page number buttons are clearly outlined.
4. Open the home page. The bottom cards show visible edges.

## Risk

Risk: 2/5 — one new colour token and seven border values on public marketing pages. No layout, no data paths, no behaviour change.


marcodejongh:
Thanks — two of three actioned in 973243750, and I'm pushing back on the third.

**1. Docs — fixed, but not where you pointed.**

You flagged the table at line 165. That table is the resolved **React Native** theme (Android fallback values / iOS `PlatformColor` roles), and `--control-border` is a web CSS custom property, so a row there would have been wrong — it would imply mobile resolves a token it has no concept of.

The right home was the **"Web (consuming Velvet Send)"** section (line 578), which already documented the surface ladder and "one hairline (`--separator`)" and was exactly the place someone would look before reaching for a border token. Added a bullet there covering the role split, the ratios on all three surfaces, why the token is opaque, and the `neutral[400]` trap.

**2. Brittle negative assertion — agreed, removed.**

You're right. It pinned a current-palette fact as a permanent invariant and would have failed a legitimate lightening of `neutral[400]` for no benefit. The intent was documentation, and that already lives in the `theme-config.ts` comment directly beside the token:

```
// Do not reuse neutral[400] here — it lands at 2.72:1 on surfaceElevated.
```

Which is where you suggested it belongs, and where someone changing the token will actually read it.

I kept the `--separator` below-3:1 assertion, though, because it is a different kind of claim. `neutral[400]`'s value is incidental to this token; `--separator` being low-contrast is *definitional* — it's a hairline, and a hairline that clears 3:1 has stopped being one. If that test ever fails it means the decorative/interactive split is being collapsed, which is precisely the regression worth catching. Happy to drop it too if you read it differently.

**3. Hex case — this matches the existing convention in each file.**

The two files use different cases on purpose, and the values follow suit:

| file | lowercase hexes | containing uppercase |
|---|---:|---:|
| `index.css` | 38 | 3 |
| `theme-config.ts` | 8 | 39 |

So `#837a9c` in CSS and `#837A9C` in TS is consistent with each file's dominant style — `--color-primary: #a78bfa` sits beside it in CSS, `primaryFill: '#7C3AED'` beside it in TS. Normalising them to match each other would make each file internally inconsistent instead. Leaving as-is; the parity test normalises for exactly this reason.

---

CI is green (1164 passed). Still draft — it's the first of a sequenced marketing polish pass, and I'd rather land the foundations together.

PR #5588, opened 2026-09-19
refactor(web): one section-heading declaration, and a fence so it stays one

marcodejongh:
Fourth PR of the marketing polish pass. Follows #5583, #5584, #5585.

The 32px section heading was declared in **four stylesheets and three inline `sx` objects**, and they had drifted three different ways at once:

| Drift | Detail |
|---|---|
| letter-spacing | `-0.02em` in page-shell + feature strip · `-0.025em` in the support block, gym search and beta headings · **none at all** on "Make it yours" |
| colour | `--neutral-900` on "Make it yours", `text.primary` everywhere else — the two-brightest-text split, on one page, three sections apart |
| **the breakpoint** | every CSS copy switched to 28px at **760px**; the inline `sx` copies used MUI's `md` (**900px**) |

That last one is the interesting one: between 760px and 900px the same page rendered headings at **two different sizes** depending on which copy happened to style each.

## The fix

One class, `.sectionHeadingType`. `PageSection` applies it beside `.sectionHeading` — the flex row that already owned the `spacing-3` margin two of the copies were hand-rolling. It's exported as `sectionHeadingTypeClassName` for the one heading that genuinely cannot be a `PageSection` child: "Where to start" spans its grid with `grid-column: 1 / -1`, and a nested `<section>` would break the span.

**The class is doubled (`.x.x`) on purpose.** The rule it replaces was `.section > h2`, whose descendant combinator out-specified MUI's Typography variant *by accident*. A single class ties that variant and loses on injection order — headings silently fall back to h3 (24px/700). I know because my first attempt did exactly that, and the geometry harness caught it: **36 elements moved**. `.x.x` states on purpose what the old selector achieved by luck.

## Measured, not assumed

Across 5 pages × 2 viewports (210 elements), the final diff is **6 changes, every one intended**:

```
~ desktop / h2|Find a board near you
    letterSpacing: -0.8px -> -0.64px
    textWrap: wrap -> balance
~ desktop / h2|Make it yours
    letterSpacing: normal -> -0.64px
    color: rgb(243,239,250) -> rgb(231,226,240)
    textWrap: wrap -> balance
~ desktop / h2|Boardsesh runs on climbers chipping in
    letterSpacing: -0.8px -> -0.64px
  (+ the same three at mobile)
```

**No width or height moved**, so nothing rewrapped despite the letter-spacing and `text-wrap` changes. "Board night, sorted" is byte-identical before and after — it was the control experiment.

## The fence

`mobile typography stays readable…` now asserts **every `main h2`** rather than two named ones, and checks weight and letter-spacing alongside size.

Nine copies didn't appear overnight — they arrived one component at a time, and a named-heading assertion cannot see the tenth one coming. This version would have caught the specificity bug above. Verified passing at 320 / 390 / 430px against a live server.

## Deliberately not migrated

- **The popular-boards rail** (`.section .title`) and **the beta heading's section** both render `null` without data the dev database doesn't have, so neither is verifiable locally. PR 11 already opens the rail for its title clamp, so the rail's heading goes there — along with its `--neutral-900` colour, the one remaining drift. Its size/weight/letter-spacing/breakpoint already match the shared class, so the widened fence passes over it either way.
- **`gym/[gym_slug]/page.tsx`** keeps its two inline copies — a detail page, out of scope for a top-level marketing pass.

## Verification

- `vp test run --project web` — 4756 passed, 419 files.
- `vp check` — 2 errors / 380 warnings, identical to `origin/main`.
- `playwright test --grep "mobile typography"` — 3 passed against a live server.

## Release Notes

Section headings are now the same size, weight and spacing on every page. Previously a few were subtly different, and at some window widths two headings on the same page rendered at different sizes.

## Test plan

1. Open the home page on a phone.
2. Every section heading is the same size.
3. Slowly widen the window past tablet size.
4. All headings grow at the same moment.

## Risk

Risk: 2/5 — CSS and class names on marketing headings. No data paths; measured to 6 intended property changes with no layout movement.


marcodejongh:
Thanks for looking. I've checked both points and I don't think either holds — happy to be shown otherwise.

**1. The "one short line max" comment rule doesn't exist.**

I grepped every CLAUDE.md that applies here (`CLAUDE.md`, `packages/web/CLAUDE.md`, `packages/web/AGENTS.md`, and the user-level one) for `one short line`, `single-line comment`, `multi-line comment` and similar. No match. If it's written down somewhere I missed, point me at it and I'll reformat.

More to the point, `page-shell.module.css` on `main` — before this PR — already contains six multi-line block comments, including a 5-line file header and this, which is the closest analogue to what I wrote:

```css
/* The ONE place the fixed MarketingHeader's height is cleared. If you find
   yourself adding var(--global-header-height) to a page, use this instead. */
```

So multi-line explanation is the established house style in this exact file, not a deviation from it. The two comments I added explain (a) why the class is doubled and (b) why the breakpoint is 760px rather than MUI's `md` — both are non-obvious facts that cost me a real bug to learn, and compressing them to one line each would drop the reason and keep only the assertion.

**2. There is no dark mode to diverge in — the site is dark-only.**

`theme-config.ts:13`: *"There is ONE scheme: dark. www is a marketing surface and the light/dark switch came out with the climbing UI (#4467)."* `index.css` says the same and the parity test fences a second scheme from coming back.

So the measurement wasn't taken in one of two schemes — it was taken in the only one that exists.

And to be clear about what that colour change is: it isn't incidental, it's one of the three drifts this PR set out to remove. `--neutral-900` (`#F3EFFA`) and `text.primary` (`#E7E2F0`) are genuinely different, and before this PR *one* heading on the home page used the first while every other heading used the second. The harness records the unification explicitly:

```
~ desktop / h2|Make it yours
    color: rgb(243,239,250) -> rgb(231,226,240)
```

"Make it yours" now matches "Board night, sorted", "Find a board near you" and "Boardsesh runs on climbers chipping in", which it didn't before. If the preference is that *all four* should be `--neutral-900` rather than `text.primary`, that's a reasonable call — but it's a change to the primitive affecting every `PageSection` heading sitewide, so I'd rather do it deliberately in its own PR than smuggle it in here.

PR #5591, opened 2026-09-19
feat(backend): make playlist discovery mean popular, and let it exclude and band

marcodejongh:
Seventh PR of the marketing polish pass, and the backend half of the `/playlists` rebuild.

## The problem, on production

`sortBy: 'popular'` ordered by `count(climbs) DESC` — **biggest, not popular**. Today's top ten:

| # | name | climbs | climbers who kept it |
|---:|---|---:|---:|
| 1 | favorites | 672 | 3 |
| 2 | Mine @ 40 | 456 | 1 |
| 3 | FL1 | 408 | 1 |
| 4 | to do | 322 | 5 |
| … | | | |
| 7 | FL2 | 268 | 0 |
| 8 | Favorites | 228 | 0 |
| 9 | favorite | 215 | 0 |
| 10 | favorites | 179 | 0 |

Four spellings of "favorites" in one grid, and the list a climber actually cares about is nowhere near the top.

## Three changes

**1. Popular now means kept.** Distinct climbers who pinned or followed the playlist, then climb count as the tiebreak. Both tables carry a unique `(user, playlist)` constraint, so a plain `count(*)` *is* a distinct-user count, and both are indexed on the playlist column.

**The tiebreak is load-bearing, not decoration.** Production has **106 pins and 20 follows across 829** public playlists with climbs — only 25 carry any engagement, and just 9 carry more than one. So most rows tie at zero and the second term is what actually orders the page: this **degrades to exactly the old ordering** instead of going arbitrary, and sharpens on its own as pins accrue.

Redefining the existing value rather than adding a new one is deliberate — every caller wanted `popular` to mean this, mobile included.

**2. A size band, through `HAVING`.** A climb count only exists after the `GROUP BY`, so this cannot be a `WHERE` — a `WHERE` would filter join rows and quietly return the wrong playlists. `minClimbs` drops one-climb scratch lists; `maxClimbs` drops the 600-climb exports that a size sort promoted to the top.

The count query had to learn the same clause or the page reports a total it will never show. `count(DISTINCT id)` has no `GROUP BY` to hang a `HAVING` on, so when a band is in play the rows are grouped first and the groups counted.

**3. Exclude-by-creator, server-side.** The viewer's own playlists were filtered on the **client** — so they were fetched, counted against `pageSize`, and then removed, silently shrinking the grid they were removed from.

## Verified against production, not asserted

The community stream now leads with:

| name | climbs | kept by |
|---|---:|---:|
| Benchmarks - vukylebeta | 115 | **13** |
| worthy | 148 | 4 |
| Classics | 74 | 2 |

Supply is healthy: **505** community playlists inside a 5–150 band, plus the **27** curated cohort lists that `/playlists` will lead with.

No new argument was needed to split curated from community — `generatedRecommendation` already does it.

## Deliberately not here

Junk-name exclusion and same-name dedupe. They need a window function over the whole public set, and the two that survive this change (`DPPTPTTPDPDDPPTPTT`, `TODO`) are a much smaller problem than the one being fixed. The page copy stays honest about the ordering either way.

## Tests

Six new, pinning the things that can silently break:

- popular orders by engagement **and** keeps the size tiebreak (4 order terms vs `recent`'s 3 — the count is what distinguishes "engagement plus tiebreak" from either alone)
- `recent` is left alone
- the band is a `HAVING`, and no `HAVING` is emitted when no band was asked for
- exclusion reaches the `WHERE`
- an out-of-range band is rejected

## Verification

- `vp run typecheck` — clean.
- `vp check` — 0 errors; no warnings in any changed file.
- `vp test run --project backend src/__tests__/discover-playlists.test.ts` — 20 passed.
- Full backend suite: 5360 passed, 7 failed. **None from this change**, verified rather than assumed — the 2 `graphql-resolvers` failures (`socket hang up`) reproduce identically on clean `origin/main`, and the moonboard/gym migration-replay failures pass in isolation on this branch, i.e. they're flaky under full parallel load.

## Release Notes

Playlists are now sorted by how many climbers actually saved them, instead of which one has the most climbs in it. A 600-climb dump of everything someone ever liked no longer outranks a playlist people use.

## Test plan

1. Open the playlists page.
2. The first playlists are ones climbers have saved.
3. A giant "favorites" list is no longer first.
4. Sign in and reload. Your own playlists are not listed.

## Risk

Risk: 3/5 — changes the meaning of an existing sort used by web and mobile. The new ordering degrades to the previous one where engagement is absent, which is most rows today.


marcodejongh:
Both actioned in a0c2d730b.

**Cross-field validation** — you're right, and the failure mode is the bad kind: `{ minClimbs: 100, maxClimbs: 50 }` passed validation, ran, and returned zero rows, which is indistinguishable from "the catalogue has nothing to show". Added the `.refine()` and a test that pins the inversion specifically, separate from the per-field range test.

**Drizzle internals in the test helpers** — fair, and worth being precise about which assertions are actually at risk. The positive ones (`toContain`) fail loudly if the helper stops finding anything, so a Drizzle shape change breaks the build rather than passing vacuously. The one real exposure was the `.not.toContain` on the `recent` sort, which would pass on an empty result. That's now paired with an independent term count — `recent` has three order terms, `popular` four — so the helper breaking can't make it pass silently either way. Comment added saying so.

I kept the helpers rather than dropping the assertions: the alternative is not checking that the new ordering reaches the pin and follow tables at all, and that's the thing most likely to be broken by a future edit.

PR #5594, opened 2026-09-19
fix(web): stop the gym directory asking "Is this your gym?" 24 times

marcodejongh:
## Summary

Five fixes to the `/gyms` directory list, all of them noise the page was adding to itself.

1. **The claim prompt was on every row.** Most gyms are unclaimed, so "Is this your gym?" was the most repeated line on the page — 24 of them under a list of 24. It stopped reading as an offer to an owner and started reading as a defect stamped on each listing. It renders once now, under the list, and is suppressed on an empty result page.
2. **It was not identifiable as a link.** `--neutral-500` with `underline="hover"` is pixel-identical at rest to the secondary sentence beside it (WCAG 1.4.1). Now `--color-primary` with a permanent underline.
3. **Rows had no reserved height.** 109px / 154px / 155px depending on whether a gym has an address, a distance, both or neither, so the board chips stepped down the page. The row is a column with a minimum in `var(--spacing-*)` arithmetic and the chip row pinned to the bottom of that slot — `min-height` alone reserves the space but leaves the chips hugging the text.
4. **Pagination pills were a fixed 40px**, under the 44px target the rest of the site holds, and `height` cannot grow for a longer label. Now `minHeight: 44`.
5. **The visible pagination label was the accessibility string.** The current pill literally read "Page 2, current page" — three words of scaffolding inside a 40px pill, which wrapped the pager onto a second row. The number is what you see; the sentence moved to a visually hidden span, and the numbered links carry it as their accessible name.

Also: the "Unclaimed" badge was a bordered pill you cannot press, so it drew a control. It keeps the label and drops the box. "Claimed" stays a chip — it has a fill, an icon, and reports a state somebody acted to reach.

### ⚠️ This renames an analytics value — do not skim past this

`Gym Claim CTA Clicked` fires from the claim prompt with a `placement` property. The prompt moved from per-row to once-per-page, so the value moves with it: **`directory-card` → `directory-footer`**, and **`gymUuid` is `null`** on that placement.

- **It is the same event and the same funnel step, but not the same denominator.** One impression per page instead of one per unclaimed row. The click rate before and after is not the same measurement — a series that spans this ship must be split at it, not read as a trend.
- **Saved insights filtering on `placement = 'directory-card'` stop matching** the day this lands. Nothing errors; they just go quiet. That is the cost, and it is deliberate — an alias would have silently averaged 24 prompts against one, which is the failure mode `docs/gym-funnel-analytics.md` exists to prevent.
- **`gymUuid` is `null`** because one prompt under 24 gyms is not about a gym. The prompt jumps back to the search box rather than guessing which listing is yours. The key stays present so a PostHog breakdown reads "no gym" instead of dropping the directory out of the denominator.
- `docs/gym-funnel-analytics.md` carries the note, the property table and the call-site list.

`viewerState` on the card and on the near-me list only ever fed the per-row prompt, so both props went with it.

### Verified by the author

`vp run typecheck:web`, `vp check` (0 errors), `vp run check:i18n`, `vp run check:i18n:orphans`, and the full web suite — 421 files / 4766 tests green, including a new `gym-directory-pagination.test.tsx` and three new directory-page tests that count the prompt.

**Not verified visually, and I am not going to imply otherwise:** the local dev database has no gym rows, so `/gyms` renders the outage panel instead of a list. The row height, the chip baseline, the footer prompt and the pager never rendered on localhost. Those four need eyes on a deploy preview.

## Release Notes

- The gym directory stops shouting "Is this your gym?" at you on every single listing — it asks once, at the bottom, where it belongs.
- Gym rows line up: board chips sit on one baseline whether a gym listed its address or not.
- Page buttons are big enough to hit on a phone, and the pager fits on one row again.

## Test plan

1. Open Gyms — rows line up, chips on one baseline.
2. Scroll to the bottom — "Is this your gym?" appears once.
3. Tap "Find your gym" — the page jumps to the search box.
4. Tap page 2 in the pager — pills read 1, 2, 3.
5. Check an unclaimed gym's badge — plain text, no box.

## Risk

Risk: 2/5 — presentation on one marketing page, covered by tests; the only durable cost is the renamed analytics placement.


marcodejongh:
Force-pushed a rebase (`887c10c67` → `33b64ab50`) to correct the diff, not the content.

This branch was cut on top of `7d4282550`, the rails commit from #5593 — a side effect of several agents sharing one worktree while these PRs were being written. The result was that this PR's diff carried four files that belong to #5593:

- `components/beta-videos/boardsesh-beta-list.tsx`
- `components/beta-videos/boardsesh-beta.module.css`
- `components/board-entity/popular-board-rail.module.css`
- `components/board-entity/popular-board-rail.tsx`

Merging both PRs would have applied those changes twice, and reviewing this one would have meant reading someone else's work as part of it.

Rebased with `--onto origin/main` to drop that ancestor. The gym commit itself touches none of those files, so the replay was clean and **not one line of this PR's own content changed** — verified by diffing the commit's stat before and after.

Re-verified after the rebase: `vp run typecheck:web` clean, `vp test run --project web` 4766 passed across 421 files.

#5593 is unaffected and still carries the rail work.

PR #5600, opened 2026-09-20
feat(web): rewrite the help guides and show their gestures as clips

marcodejongh:
## Summary

A writing pass over the seven `/help` guides and the hub that #5599 shipped yesterday, plus short looping clips of the app for the gestures a still cannot show.

**Writing.** The guides are now about 3,200 words, down from 6,400: every section leads with the action, steps are numbered imperatives, and reassurance and rare edge cases are gone. Two features that were buried now have named sections: "Share a reel from Instagram or TikTok" opens the beta-videos guide (`#share`) and "Control the board from the lock screen (Live Activity)" closes the Bluetooth guide (`#live-activity`); both are named on their hub cards.

Before the cut, four writers re-read every page against the mobile source and rewrote all ~360 `help.*` strings, then a separate reviewer went through the result for signs of AI writing (fragment section leads, the "X, not Y" tic, em dashes, colon asides, wry sign-offs) and unified the naming (play drawer, long press, American spelling to match the app's labels). Spanish, French and German were retranslated from the new English with the glossaries; ten older French strings moved from vous to tu so a page no longer mixes registers.

The rewrite also fixed what the pages claimed. The bigger corrections:
- Kilter linking is **Sign in to Kilter** (Tension is **Link**); sync is one-way and runs about every 12 hours; the climb-count comparison now says "appears to" and carries the numbers from `docs/kilter-sync.md`.
- The playlist create sheet takes a name, description and color; emoji and public/private live under Edit details. Hidden Gems, Most Repeated and Projects are defined the way the backend defines them.
- The edit-ascent sheet has no date field. **Unsent** keeps attempted climbs; the five Progress chips are named.
- "Last connected board" is only a fallback label. Auto-disconnect is set in Settings, not in the long-press sheet. The config-mismatch alert is **Board configuration doesn't match** with Cancel / Connect anyway / Switch board setup.
- The Climbs tab has no top-bar lightbulb; the queue-bar board chip is the control there.
- Sessions: **Start a session** buttons on Home and in the board sheet also lead to Record; the live toggle is on by default; the out-of-sync toast follows a failed queue write, not a wifi drop; only the climber who started a session can end it.
- Climb actions: menu order and gates match `use-climb-actions.ts`; Edit climb is a 24-hour window; Open in Aurora app excludes Kilter.
- Beta videos: Instagram and TikTok only, attach is one tap with no undo, the caption match only works with Boardsesh's own quoted caption and not on MoonBoard.

**Clips.** Nine 10–13 s silent recordings of the real app on an iPhone 16 Pro Max simulator (screenshot-mode dev client, prod test account), driven by Maestro flows committed under `packages/mobile/.maestro/help-clips/` so they can be re-shot. `vp run help:convert-clips` turns a raw `simctl recordVideo` into mp4 + webm + a poster; the static-asset catalog now carries video; `HelpClip` autoplays muted and looped, and shows the poster with controls when the OS asks for reduced motion. Contract and runbook in `docs/help-clips.md`.

Author-side verification: `vp run typecheck:web`, `vp check`, `vp run check:i18n`, `check:i18n:orphans`, the i18n parity tests, the scripts tests and the static-asset check are green; every page was opened in Chrome at desktop and phone width, including `/es/help/climb-actions`.

**Found along the way.** Two bare-name LED boxes ("Kilter Board") in one gym collapse into a single picker row because the scan cache keys on the advertised name. Filed as #5601; the guide now states it as a current limitation with the workaround instead of describing it as how things work.

## Release Notes

- The help guides on boardsesh.com now show short clips of the gestures they describe, and the copy was rewritten and fact-checked against the app.

## Test plan

1. Open boardsesh.com/help → seven guide cards, each says what it answers.
2. Open Climb actions → a looping clip shows a long press opening the menu.
3. Open Finding climbs on your phone → clips play inside the phone frame.
4. Turn on Reduce Motion → reload → clips show a poster with play controls.
5. Open /es/help/playlists → Spanish copy, "plafón" for board.

## Risk

Risk: 1/5 — copy, static assets and a docs pipeline; no app or backend code paths change.

https://claude.ai/code/session_01AkKDWRDqy1iF9uiN9Zy6kL


marcodejongh:
Addressed the three Codex notes in b74b85c: clips now play only within 200 px of the viewport and pause when scrolled away; the CI staticAssets filter and the production deploy classifier count public/videos mp4/webm (with a test); the German title and hub card use Board-Bereich like the app.

PR #5634, opened 2026-09-20
fix(mobile): activate session climbs after loading missing details

marcodejongh:
## Summary

A session tick without hold data opened through the climb route as a preview, even after its climb details loaded. The same tap therefore behaved differently depending on the session payload. The fallback now preserves an internal tick tap's activation intent, so a recovered climb can become current when climbing solo.

Activation uses a short-lived, single-use in-memory intent bound to the climb and board configuration. Public URL parameters alone cannot activate a climb. Explicit previews, notification and moderation references, and the shared-session browse guard retain their existing behavior.

Refs #5254. The report does not establish whether this specific fallback caused the reported difficulty; missing climbs and physical Bluetooth lighting remain separate QA limits.

Independent review approved. Focused verification passed 87 helper/route tests and three real drawer solo/crew cases, including delayed frames, external links, intent replay/expiry, and explicit previews. Final validation passed: full typecheck; 990 mobile test files / 11,165 tests; iOS and Android bundle exports; repository format/lint checks (0 errors); cached iOS simulator startup and screenshot. Simulator QA reached a clean sign-in screen on this branch’s Metro; signed-in session flows and physical LEDs remain manual checks.

## Release Notes

Session climbs that need to load their hold details now become current after loading when you're climbing solo. Preview actions and browsing with your crew still leave the current climb alone.

## Test plan

1. You → Sessions → open a past session; climbs appear.
2. Tap a climb; it becomes current when climbing solo.
3. Long-press a climb → Preview; current climb stays unchanged.
4. Open a shared climb link; current climb stays unchanged.
5. Join a crew → tap session climb; current climb stays unchanged.

## Risk

Risk: 3/5 — changes session-climb navigation intent; shared-session browsing stays guarded. No BLE, schema, or native dependency changes.


marcodejongh:
Fixes https://github.com/boardsesh/boardsesh/issues/5254

PR #5635, opened 2026-09-20
fix(web): stop the setter fallback branches republishing hreflang twins

marcodejongh:
## Summary

`generateMetadata` on `/setter/[setter_username]` uses `createBoardContentPageMetadata` on its happy path — canonical pinned to the en-US URL, `alternates: { canonical }` and nothing else. Its two fallback branches did not. Both `createNoIndexMetadata` calls (the setter the OG summary cannot find, and the lookup that throws) delegate to `createPageMetadata`, which emits a **self-canonical plus the full four-locale `alternates.languages` map**.

So the branches a crawler is most likely to hit were republishing exactly the hreflang cluster the board-content carve-out exists to collapse, across a surface of roughly 32,000 setter URLs. `robots: noindex` does not suppress it: Google reads `hreflang` independently, and it wants every member of a cluster to self-canonicalise — a noindexed twin advertising three more twins is the same contradiction the carve-out was measured to remove (~205k climb-view renders/day, Postgres at 183/200 connections).

Both branches now call `createBoardContentPageMetadata({ ..., robots: NOINDEX_FOLLOW })`. That option is already threaded into the same helper on the happy path (`robots ?? (linklessPage ? NOINDEX_FOLLOW : undefined)`), so nothing new was needed — the path existed and was simply not used. `createNoIndexMetadata` is no longer imported here.

`board-content-metadata-guard.test.ts` should have caught this and could not: `BARE_METADATA_HELPER` only matched the bare `createPageMetadata`, and `createNoIndexMetadata` does not contain that name. The regex now rejects both across all five board-content route files. Checked: the setter page was the only one of the five reaching for it.

**Verification (author side):**
- `vp run test:web` — 420 files, 4759 tests, all passing.
- `vp check` — clean on the three touched files (the two pre-existing errors in `packages/mobile/src/web-shims/bottom-sheet.tsx` are on `main` and untouched here).
- Confirmed both new guards actually bite: reverting the first branch to `createNoIndexMetadata` fails the tightened guard **and** the new head test (`2 failed | 39 passed`). Restoring the fix returns all 41 to green.

## Release Notes

## Test plan

1. CI green.

## Risk

Risk: 2/5 — isolated metadata change on two fallback branches, pinned by new tests.

https://claude.ai/code/session_01W2TBCiJ2p7w5Pw9AuC8kUb


marcodejongh:
Checked the indentation call-out: the `catch` block's object literal is 6-space indented on every property, `robots: NOINDEX_FOLLOW,` included — byte-identical to the `if (!summary)` branch. `vp check` (Prettier) is clean on the file and the `check` job passed, so there is nothing to change here.

PR #5636, opened 2026-09-20
fix(web): keep climbers' home walls out of the search index

marcodejongh:
## Summary

**This closes a live privacy exposure, not an SEO chore.**

Measured read-only against production on 2026-09-20: of **4,468 live public gyms, 2,848 carry coordinates** — leaving exactly **1,620 rows that are climbers' personal home walls**, every one with a slug and therefore a live URL. **1,053 of those names begin with the owner's own `users.name`** (joined `gyms.owner_id -> users.id`), and `gymPage.metaTitle` renders them as `"{name} climbing gym"` with full indexable metadata. Google already serves 21 of them: `zhouzhou-s-moonboard-2024-standard` sits at position 3.5 with a click, `/de/gym/clemens-m-ller-s-moonboard-2024-standard` at position 2.5.

The page's index directive was `gym.isPublic ? createPageMetadata : createNoIndexMetadata`, and `isPublic` is true for a garage wall.

**What changed**

1. New `packages/web/app/lib/seo/gym-index-eligibility.ts` exports one predicate, `gymIsIndexableVenue(gym)` — public **and** geocoded. Its docblock carries the 4,468 / 2,848 / 1,620 measurement and the `searchGyms` precedent so nobody "simplifies" the coordinate clause away.
2. `generateMetadata` gates on it. A gym that fails gets `createNoIndexMetadata`.
3. **A failing gym is passed no `path`.** This is the load-bearing half: `createNoIndexMetadata` delegates to `createPageMetadata`, which builds `alternates.canonical` plus the four-locale `hreflang` map only when a path is present (`metadata.ts:82-87`). A noindexed page that still advertises its three locale twins republishes the cluster we are removing. Verified in the source and pinned by a test. The sibling `/gym/[slug]/poster` has taken the same no-`path` route since it landed.
4. The `SportsActivityLocation` JSON-LD rides the same predicate. Describing a home wall as an activity location is the same leak in a second format.
5. **No `robots.txt` `Disallow`, deliberately.** The pages keep serving 200 and stay crawlable. A disallowed URL is never recrawled, so Google would never read the `noindex` and the 21 already-indexed entries would be stranded permanently.

**Why coordinates are the proxy.** The Aurora gym sync validates coordinates before writing, so a synced commercial gym always has a pin and a hand-created garage wall never does. The same proxy is already load-bearing one layer down: `searchGyms` filters `location IS NOT NULL` (`packages/backend/src/graphql/resolvers/social/gyms.ts:913`), which is the only reason `/gyms` renders no home walls today. `location` and `latitude`/`longitude` are the same gate — the `gyms_set_location` trigger derives the geography from the pair on every write.

**Known collateral, stated plainly**

- Roughly **540 pin-less walls have neutral names** and are caught too. Acceptable: they are thin pages with no venue content, and they were never going to rank for anything.
- **About 8 real commercial gyms have never been geocoded** and go `noindex, follow` with them. Self-healing — they return the moment someone drops a pin on the listing. I could not name them from here: the measurement above came from a read-only production query run before this branch, and I have no prod DB access in this worktree to re-run the name lookup. Worth a one-off query before merge if you want them geocoded first.

**My verification:** `vp run test:web` — 422 files / 4,769 tests pass. `vp run typecheck` clean. `vp check` reports no error in any file this PR touches (the two errors it prints are pre-existing `no-floating-promises` hits in `packages/db/scripts/dedupe-gym-locations-args.test.ts`, untouched here). Pre-commit hook passed, including `check:i18n` — no new strings.

`/gyms`, the sitemap (the gyms shard is declared-empty today) and `isGymViewable` are untouched: every gym that served a 200 before still serves one.

## Release Notes

## Test plan

1. Open a gym page with an address. Page loads normally.
2. Open a home-wall gym page. Page still loads normally.
3. CI green.

## Risk

Risk: 3/5 — changes the indexing directive on live ranking pages.

https://claude.ai/code/session_01W2TBCiJ2p7w5Pw9AuC8kUb


marcodejongh:
Both test-coverage points applied in fc41ea0 — the private-gym case now asserts `alternates` is undefined (that behaviour change was indeed silent), and `Infinity` / `-Infinity` are covered.

On the sitemap: nothing to follow up. The gyms shard is declared-empty today — `buildGymEntries()` returns `[]` and the registry marks it `expectsUrls: false`, because there is no public-gyms enumeration query and #4372's duplicate queue gates it. So no home wall is submitted, and none burns crawl budget from a sitemap. When #4381 fills that builder, `gymIsIndexableVenue` is the predicate it should use.

